PR #1512, opened 2016-08-12
gsdx sw: add extrathreads_height to control the quantity of pixels pr…

gregory38:
Value could range from 1 to 9. Default is 4 and it is potentially the
best option.  Feel free to test some values on your system, behavior
might depends on the core number and thread number

Value is exponential so 4 is 2 times more pixels than 3.

Small value increased thread overhead, big value increase wait/sync latency

### DIFF
--- a/plugins/GSdx/GSRasterizer.cpp
+++ b/plugins/GSdx/GSRasterizer.cpp
@@ -24,13 +24,20 @@
 #include "stdafx.h"
 #include "GSRasterizer.h"
 
-// - for more threads screen segments should be smaller to better distribute the pixels
-// - but not too small to keep the threading overhead low
-// - ideal value between 3 and 5, or log2(64 / number of threads)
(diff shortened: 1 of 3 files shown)


ramapcsx2:
Can you just merge this to master, so we can test it at leisure? :)

refractionpcsx2:
> Can you just merge this to master, so we can test it at leisure? :)

just grab the appveyor builds :P

ramapcsx2:
It's inconvenient still :p I'm asking because this is basically a no-op change and could just go onto master directly. We can test it whenever there's time / without going through appveyor or branches.

gregory38:
Guys, you have the power to merge it :)

gregory38:
Edit and by the way, I think 4 is the best value. But it could still be useful in future, if someone (or I) change thread managements.

PR #1513, opened 2016-08-12
GameDB: Fix Grandia III MemCardFilter

trostboot:
MemCardFilter for Grandia III NTSC-U Disc 2 was applied to the wrong
game. The serials for Disc 1 and 2 are SLUS-21334 and SLUS-21345,
respectively - I assume someone fatfingered when issuing the serials.


ramapcsx2:
Thanks, looks like it :)

PR #1514, opened 2016-08-12
Gsdx alpha test improvement

gregory38:
This PR reorganize a bit alpha testing and depth configuration.

First purpose is to ease future investigation on uber shader for alpha test. It also permits to improve some text rendering issue on some games (burnout, kengo). Note improvements are only for openGL but anyone could port the change to Dx.


refractionpcsx2:
if you can outline the actual changes that are needed that might make porting it to GSDX easier.  TBH it would be nice to port all the channel shuffle stuff properly, but it's a heck of a lot of work.

FlatOutPS2:
> if you can outline the actual changes that are needed that might make porting it to GSDX easier. TBH it would be nice to port all the channel shuffle stuff properly, hell, maybe even the depth stuff, but it's a heck of a lot of work.

I already ported it to DX, I'll make a PR soon. :)

gregory38:
2 last commits, one to optimize depth to always. And the last to use it to enable first color processing.

gregory38:
```
TBH it would be nice to port all the channel shuffle stuff properly, hell, maybe even the depth stuff, but it's a heck of a lot of work.
```
Actually I'm not sure it is doable anymore (because I might sample the framebuffer). But it won't be that complicated. All the complex code on the texture cache is the same. You just need to adapt the shader to convert the depth value. Maybe one could start by the shader in convert, there are quite basic. You just need to find how read/write depth texture. And then update strech rectangle to use them properly.

Issue #1488, opened 2016-07-29
Discussion on code formating


gregory38:
I would like to square the code database to uniformize coding style.
- At a bare-minimum, I would like to remove all trailing white space (small impact on history)
- It would be nice to convert leading tab to space everywhere (big impact on history, depends on the part)
- Reformat with an automatic tool such as `astyle` or `clang-format` (huge impact on history)

Potentially we could limit the change to a subset of the database. For example we could start with a couple of plugin onepad/spu2x. I don't know if it possible to add a hook on git to enforce the coding style (aka the hard policy).

edit: hook example can be found in `.git/hooks/`

Or we could keep current situation huge mess of coding style.

For refernce here all the configuration available on clang-format

```


Language:        Cpp
# BasedOnStyle:  LLVM
AccessModifierOffset: -2
AlignAfterOpenBracket: true
AlignConsecutiveAssignments: false
AlignEscapedNewlinesLeft: false
AlignOperands:   true
AlignTrailingComments: true
AllowAllParametersOfDeclarationOnNextLine: true
AllowShortBlocksOnASingleLine: false
AllowShortCaseLabelsOnASingleLine: false
AllowShortFunctionsOnASingleLine: All
AllowShortIfStatementsOnASingleLine: false
AllowShortLoopsOnASingleLine: false
AlwaysBreakAfterDefinitionReturnType: None
AlwaysBreakBeforeMultilineStrings: false
AlwaysBreakTemplateDeclarations: false
BinPackArguments: true
BinPackParameters: true
BreakBeforeBinaryOperators: None
BreakBeforeBraces: Attach
BreakBeforeTernaryOperators: true
BreakConstructorInitializersBeforeComma: false
ColumnLimit:     80
CommentPragmas:  '^ IWYU pragma:'
ConstructorInitializerAllOnOneLineOrOnePerLine: false
ConstructorInitializerIndentWidth: 4
ContinuationIndentWidth: 4
Cpp11BracedListStyle: true
DerivePointerAlignment: false
DisableFormat:   false
ExperimentalAutoDetectBinPacking: false
ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH ]
IndentCaseLabels: false
IndentWidth:     2
IndentWrappedFunctionNames: false
KeepEmptyLinesAtTheStartOfBlocks: true
MacroBlockBegin: ''
MacroBlockEnd:   ''
MaxEmptyLinesToKeep: 1
[...]

PR #1516, opened 2016-08-14
pcsx2: manually cast function pointer to void*

gregory38:
Templace is nicer but give a hard time to compiler.

New version compile in both gcc&clang without hack

=> #1488

### DIFF
--- a/pcsx2/x86/microVU_Branch.inl
+++ b/pcsx2/x86/microVU_Branch.inl
@@ -57,8 +57,8 @@ void mVUDTendProgram(mV, microFlagCycles* mFC, int isEbit) {
 			mVU_XGKICK_DELAY(mVU);
 		}
 		if (doEarlyExit(mVU)) {
-			if (!isVU1) xFastCall(mVU0clearlpStateJIT);
-			else		xFastCall(mVU1clearlpStateJIT);
+			if (!isVU1) xFastCall((void*)mVU0clearlpStateJIT);
+			else		xFastCall((void*)mVU1clearlpStateJIT);
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


turtleli:
Compiles fine with clang 3.8.1. Too bad VS2013 doesn't like it.

gregory38:
Yeah. Maybe I need to add an (uptr) cast too :(

gregory38:
Sigh... Uptr won't do it. Strange.

pcsx2/x86/microVU_Branch.inl
```diff
@@ -192,8 +192,8 @@ void normJumpCompile(mV, microFlagCycles& mFC, bool isEvilJump) {
 		xJMP(mVU.exitFunct);
 	}
 	
-	if (!mVU.index) xFastCall(mVUcompileJIT<0>, gprT2, gprT3); //(u32 startPC, uptr pState)
-	else			xFastCall(mVUcompileJIT<1>, gprT2, gprT3);
+	if (!mVU.index) xFastCall((void*)(uptr)mVUcompileJIT<0>, gprT2, gprT3); //(u32 startPC, uptr pState)
+	else			xFastCall((void*)(uptr)mVUcompileJIT<1>, gprT2, gprT3);
```
turtleli: `(void*)(void(*)())mVU...` works.

gregory38: Thanks you

gregory38:
Ok. Compilers are happy, let's merge it.

danilaml:
Btw, are there any plans to drop VS2013 in the near future? VS2015 is all-around better.

gregory38:
When ms release a new version of visual studio. 2013 supports most of the useful C++11 construct. It isn't blocking as previous version.

Issue #1329, opened 2016-04-30
Grand Theft Auto: Vice City HW text looks wrong (PAL)


AmbientMalice:
-PCSX2 version: v1.5.0-20160429001616

-PCSX2 options: Defaults.

-Plugins used: Defaults.

-Plugin settings: GL HW & GL SW.

-Description: 2D elements like text have odd patterns similar to embossing. Not affected by upscaling, but more noticable.

SW
![gsdx_20160430155027](https://cloud.githubusercontent.com/assets/10704701/14934207/feaaa2f8-0eec-11e6-95a3-0b819aa7fc26.png)
HW
![gsdx_20160430155024](https://cloud.githubusercontent.com/assets/10704701/14934202/b452401c-0eec-11e6-949b-86e94647286c.png)

-How to replicate: Run the game. Issue appears anytime normal HUD is active.

-Last known version to work: Don't know. I think it was always this way.

-PC specifications: CPU: Intel Q9400z GPU: Gigabyte GTX 750Ti OS: Windows 10 64 Bit

pepiczech77:
You are using old version of PCSX2. Try latest master.

AmbientMalice:
@pepiczech77 Sorry, that was a typo.

refractionpcsx2:
can you show us your GSDX settings?

AmbientMalice:
![gta_set](https://cloud.githubusercontent.com/assets/10704701/14935636/b7c1ed54-0f1b-11e6-8a4b-ccb5fbbec282.PNG)

refractionpcsx2:
Okay can you try again with anisotropic filtering off, just to rule that out.  Maybe play with a few other settings to see if they are causing it

AmbientMalice:
Tried various settings, and different backends. No difference.

I did notice the following when the resolution is cranked up:
This is software.
![shadow_1](https://cloud.githubusercontent.com/assets/10704701/14935666/34a77c20-0f1d-11e6-8d67-b0b372759c4a.png)
This is HW.
![shadow_2](https://cloud.githubusercontent.com/assets/10704701/14935667/3bfd7b28-0f1d-11e6-9f52-c7102881828a.png)
The drop shadow+overlay texture isn't being layered properly.
![shadow_3](https://cloud.githubusercontent.com/assets/10704701/14935672/5b84750a-0f1d-11e6-88c0-39acbe2341f2.jpg)

FlatOutPS2:
[...]

PR #1517, opened 2016-08-14
Gsdx alpha test improvement DX

FlatOutPS2:
Port for the DX renderers of the alpha test improvement PR for OGL(#1514) created
by gregory38.

Tested and confirmed to work the same as OGL with both DX9 and DX11, fixing text issues for the Burnout games, Kengo and Ridge Racer V.

### DIFF
--- a/plugins/GSdx/GSRendererDX.cpp
+++ b/plugins/GSdx/GSRendererDX.cpp
@@ -48,21 +48,143 @@ GSRendererDX::~GSRendererDX()
 {
 }
 
+void GSRendererDX::EmulateAtst(const int pass, const GSTextureCache::Source* tex)
+{
+	static const uint32 inverted_atst[] = { ATST_ALWAYS, ATST_NEVER, ATST_GEQUAL, ATST_GREATER, ATST_NOTEQUAL, ATST_LESS, ATST_LEQUAL, ATST_EQUAL };
+	int atst = (pass == 2) ? inverted_atst[context->TEST.ATST] : context->TEST.ATST;
+
+	if (!context->TEST.ATE) return;
+
+	switch (atst) {
+	case ATST_LESS:
+		if (tex && tex->m_spritehack_t) {
+			ps_sel.atst = 0;
+		}
+		else {
+			ps_cb.FogColor_AREF.a = (float)context->TEST.AREF - 0.1f;
+			ps_sel.atst = 1;
+		}
+		break;
+	case ATST_LEQUAL:
+		ps_cb.FogColor_AREF.a = (float)context->TEST.AREF - 0.1f + 1.0f;
+		ps_sel.atst = 1;
+		break;
+	case ATST_GEQUAL:
+		// Maybe a -1 trick multiplication factor could be used to merge with ATST_LEQUAL case
+		ps_cb.FogColor_AREF.a = (float)context->TEST.AREF - 0.1f;
+		ps_sel.atst = 2;
+		break;
+	case ATST_GREATER:
+		// Maybe a -1 trick multiplication factor could be used to merge with ATST_LESS case
+		ps_cb.FogColor_AREF.a = (float)context->TEST.AREF - 0.1f + 1.0f;
+		ps_sel.atst = 2;
+		break;
+	case ATST_EQUAL:
+		ps_cb.FogColor_AREF.a = (float)context->TEST.AREF;
+		ps_sel.atst = 3;
+		break;
+	case ATST_NOTEQUAL:
+		ps_cb.FogColor_AREF.a = (float)context->TEST.AREF;
+		ps_sel.atst = 4;
+		break;
+	case ATST_NEVER:
+	case ATST_ALWAYS:
+	default:
+		ps_sel.atst = 0;
+		break;
+	}
+
+	// Destination alpha pseudo stencil hack: use a stencil operation combined with an alpha test
+	// to only draw pixels which would cause the destination alpha test to fail in the future once.
+	// Unfortunately this also means only drawing those pixels at all, which is why this is a hack.
+	// The interaction with FBA in D3D9 is probably less than ideal.
+	if (UserHacks_AlphaStencil && DATE && dev->HasStencil() && om_bsel.wa && (!context->TEST.ATE || context->TEST.ATST == 1))
+	{
+		if (!context->FBA.FBA)
+		{
+			if (context->TEST.DATM == 0)
+				ps_sel.atst = 2; // >=
+			else {
+				if (tex && tex->m_spritehack_t)
+					ps_sel.atst = 0; // <
+				else
+					ps_sel.atst = 1; // <
+			}
+			ps_cb.FogColor_AREF.a = (float)0x80;
+		}
+		if (!(context->FBA.FBA && context->TEST.DATM == 1))
+			om_dssel.alpha_stencil = 1;
+	}
+}
+
+void GSRendererDX::EmulateZbuffer()
+{
+	if (context->TEST.ZTE)
+	{
+		om_dssel.ztst = context->TEST.ZTST;
+		om_dssel.zwe = !context->ZBUF.ZMSK;
+	}
+	else
+	{
+		om_dssel.ztst = ZTST_ALWAYS;
+	}
+
+	uint32 max_z;
+	if (context->ZBUF.PSM == PSM_PSMZ32) {
+		max_z = 0xFFFFFFFF;
+	}
+	else if (context->ZBUF.PSM == PSM_PSMZ24) {
+		max_z = 0xFFFFFF;
+	}
+	else {
+		max_z = 0xFFFF;
+	}
+
+	// The real GS appears to do no masking based on the Z buffer format and writing larger Z values
+	// than the buffer supports seems to be an error condition on the real GS, causing it to crash.
+	// We are probably receiving bad coordinates from VU1 in these cases.
+
+	if (om_dssel.ztst >= ZTST_ALWAYS && om_dssel.zwe && (context->ZBUF.PSM != PSM_PSMZ32)) {
+		if (m_vt.m_max.p.z > max_z) {
+			ASSERT(m_vt.m_min.p.z > max_z); // sfex capcom logo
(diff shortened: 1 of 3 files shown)


plugins/GSdx/GSRendererDX.cpp
```diff
@@ -48,15 +48,176 @@ GSRendererDX::~GSRendererDX()
...
+	
+	// Minor optimization of a corner case (it allow to better emulate some alpha test effects)
+	if (om_dssel.ztst == ZTST_GEQUAL && m_vt.m_min.p.z >= max_z_flt) {
+#ifdef _DEBUG
+		fprintf("Optimize Z test GEQUAL to ALWAYS (%s)", psm_str(context->ZBUF.PSM));
+#endif
+		om_dssel.ztst = ZTST_ALWAYS;
+	}
+
+	if (m_fba)
```
gregory38: Why did you put FBA & blending emulation too ? The function is called Z buffer emulation aka depth ;)

FlatOutPS2:
> Why did you put FBA & blending emulation too ? The function is called Z buffer emulation aka depth ;)

I think it was because of a regression in an early version of the port. I completely forgot about it being there. :p

plugins/GSdx/GSRendererDX.cpp
FlatOutPS2: > Why did you put FBA & blending emulation too ? The function is called Z buffer emulation aka depth ;)

I think it was because of a regression in an early version of the port. I completely forgot about it being there. :p

MrCK1:
I guess I'll test Vice City later. I'm not expecting anything different than the half-baked results I got in OGL

FlatOutPS2:
If it's not fixed in the OGL version, it won't be any better in this. It's a port, not an upgrade. :D

gregory38:
What a deception :p

Care to share a link to a gs dump. Current implementation is rather limited to a special case. (Depth test can be executed in an out of order). At least I can tell if it is the same bug or another one.

FlatOutPS2:
> Care to share a link to a gs dump. Current implementation is rather limited to a special case. (Depth test can be executed in an out of order). At least I can tell if it is the same bug or another one.

I already asked for one in the issue he linked in your PR: #1329

gregory38:
I have some pending change for ate.
* a case that wasn't correct
* a 2nd optimization for vice city that disable alpha test completely in some cases.
* And I'm looking at the remaining vice city issue (which is very close)

gregory38:
Sorry, you need to update it again.

FlatOutPS2:
That fixes the map in GTA. There's one more remaining issue that was fixed when I hack-fixed Burnout games, but wasn't touched by any of the changes in this PR, the dancing girl in Ridge Racer V's main menu is still split in half, left side is correct, depth of right side is wrong.

gregory38:
```
the dancing girl in Ridge Racer V's main menu is still split in half, left side is correct, depth of right side is wrong.
```
Half screen issue feels like a texture cache issue rather than a wrong alpha test per se (I would say a wrong depth buffer content).

FlatOutPS2:
> Half screen issue feels like a texture cache issue rather than a wrong alpha test per se (I would say a wrong depth buffer content).

It's not. It's not fixed by turning off texture cache and it was fixed by an alpha test modification.

gregory38:
```
It's not. It's not fixed by turning off texture cache and it was fixed by an alpha test modification.
```
I don't think depth buffer is properly read when you disable the texture cache. Anyway, give me a gs link, if it is an alpha test issue it should be easy to find the bad draw call.

FlatOutPS2:
GS link: [Ridge Racer V.zip](https://github.com/PCSX2/pcsx2/files/420601/Ridge.Racer.V.zip)

gregory38:
Did you try to replay your dump? It seems to be fine. However there is a single ATE draw call. It is done to draw the circle. And I can see a half screen delimitation effect. But your screenshot seems to be fine on that side. Remaining alpha test is skipped. 

Hum, however there is a destination alpha test draw call to draw the girl but it is actually a nop. I need to run the game (I have it).

gregory38:
I understand the line. The depth buffer is cleared with line. I need to finish a hack to replace point/line with square/rectangle to avoid such aliasing. I'm trying to look at the half screen now. The depth buffer is wrong so ATE later is broken.

plugins/GSdx/GSRendererDX.cpp
```diff
@@ -48,15 +48,139 @@ GSRendererDX::~GSRendererDX()
 {
 }
 
+void GSRendererDX::EmulateAtst(const int pass, const GSTextureCache::Source* tex)
+{
+	GSDeviceDX* dev = (GSDeviceDX*)m_dev;
```
gregory38: For the record I do that in openGL because it allow the compiler to optimize virtual function call. OpenGL device is the lastest inheritance, so you know the function at compile time. On dx it is useless you still need to dispatch it at runtime to select either dx9 or dx11.

plugins/GSdx/GSRendererDX.cpp
```diff
@@ -48,15 +48,139 @@ GSRendererDX::~GSRendererDX()
...
+		om_dssel.ztst = ZTST_ALWAYS;
+	}
+}
+
 void GSRendererDX::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Source* tex)
 {
-	GSDrawingEnvironment& env = m_env;
-	GSDrawingContext* context = m_context;
+	context = m_context;
+	env = m_env;
```
gregory38: This code suck because it will a (useless) copy of the env object. On openGL I replaced all occurence to use `m_` prefix instead. It doesn't worth it, it just save a couple of chars.
You can only replace it on the newly created function. It would be a better idea to do it in a separate commit if you do it.

FlatOutPS2: > You can only replace it on the newly created function. 

What do you mean by that?

gregory38: I mean the only pupose of `env` is to write `env.foo` instead of `m_env.foo`. In short it is useless. If you use `env...` in the new created function for depth/alpha, you just need to call `m_env....` instead.

FlatOutPS2: Right, but why can I only replace it on the newly created function?

gregory38: What I mean is that you could do
```
depth () {
 m_env....
}
draw() {
  env = m_env;
 env.ff
```
Or you can replace it everywhere. Just more change and a bigger diff but you're done for the next time. It is up to you. Or you could use the env = m_env renaming everywhere. But don't store it in the object, it is a bad idea.

FlatOutPS2: Okay. But what about m_context?

gregory38: Same stuff. Look at the code in the openGL renderer. You can see that I directly use `m_context`

FlatOutPS2: Yes, I already did. But in the original GSRendererDX code both m_context and context are used, so I wasn't sure if it was required or not.

gregory38: Old code uses context, newer code use m_context (imported from openGL)

plugins/GSdx/GSRendererDX.cpp
```diff
@@ -48,21 +48,143 @@ GSRendererDX::~GSRendererDX()
...
+	// than the buffer supports seems to be an error condition on the real GS, causing it to crash.
+	// We are probably receiving bad coordinates from VU1 in these cases.
+
+	if (om_dssel.ztst >= ZTST_ALWAYS && om_dssel.zwe && (context->ZBUF.PSM != PSM_PSMZ32)) {
+		if (m_vt.m_max.p.z > max_z) {
+			ASSERT(m_vt.m_min.p.z > max_z); // sfex capcom logo
+			// Fixme :Following conditional fixes some dialog frame in Wild Arms 3, but may not be what was intended.
+			if (m_vt.m_min.p.z > max_z) {
+#ifdef _DEBUG
+				fprintf("Bad Z size on %s buffers", psm_str(context->ZBUF.PSM));
```
gregory38: Does it compile on debug build ? Don't you need either `stdout` or `stderr` in first paramerter (or use `printf`). And you might want to add a final `\n`

plugins/GSdx/GSRendererDX.cpp
```diff
@@ -48,21 +48,143 @@ GSRendererDX::~GSRendererDX()
...
+				om_dssel.ztst = ZTST_ALWAYS;
+			}
+		}
+	}
+
+	GSVertex* v = &m_vertex.buff[0];
+	// Minor optimization of a corner case (it allow to better emulate some alpha test effects)
+	if (om_dssel.ztst == ZTST_GEQUAL && (m_vt.m_eq.xyzf & 0x8) && v[0].XYZ.Z == max_z) {
+#ifdef _DEBUG
+		fprintf("Optimize Z test GEQUAL to ALWAYS (%s)", psm_str(context->ZBUF.PSM));
```
gregory38: same as above

plugins/GSdx/GSRendererDX.cpp
```diff
@@ -358,30 +437,30 @@ void GSRendererDX::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sourc
...
-				ps_sel.atst = 5; // >=
-			else
-				ps_sel.atst = 2; // <
-			ps_cb.FogColor_AREF.a = (float)0x80;
-		}
-		if (!(context->FBA.FBA && context->TEST.DATM == 1))
-			om_dssel.alpha_stencil = 1;
+	if (ate_skip) {
+#ifdef _DEBUG
+		fprintf("Alternate ATE handling: ate_skip");
```
gregory38: See previous comment for those 2 printfs too.

gregory38:
Is it ready to be merged ? TBH, I don't like the placement of  `{` around if/else but clang-format will take of it.

FlatOutPS2:
In terms of it working as good as the OGL version it was ready to be merged when I created the PR.

Placement of `{` is based on your PR, what's used in nearby inconsistent code and possible VS trickery.

PR #1521, opened 2016-08-16
gsdx: improve Horizonal/Vertical line upscaling

gregory38:
Line width was increased from 0 to 1 pixel. Scaling will keep line proportion and it will avoid aliasing.

Avoid empty line in the screen for lines based rendered (Ridge Racer).
Emulation is done on the CPU hopefully perf impact will be null.

@FlatOutPS2 feel free to toy with it. Ridge Racer is better. But I saw some glitch on letters, didn't test others scaling option.


gregory38:
Sorry I forgot a part of the commit. It would work better this way :)

ramapcsx2:
Do you know of other affected games? This is an interesting change!

gregory38:
I remember at least 1 or 2 games massively affected by this issue.  Maybe a FMV. However I don't remember the name of those games ;)
Massive == screen rendered as lines. However game might use line to draw some menu elements (ace combat) so it is harder to notice it. Lines become thinner with upscaling.

willkuer:
> Lines become thinner with upscaling

With this commit that's not longer the case? Or do lines still disappear?

plugins/GSdx/GSRendererHW.h
```diff
@@ -29,7 +29,7 @@
 
 class GSRendererHW : public GSRenderer
 {
-private:
+protected:
```
willkuer: don't you expose a lot here just to get m_upscale_multiplier?

Also you are not interested in m_upscale_multiplier but you want to know if (m_upscale_multiplier>1). Why not a protected function protected bool IsUpscaled() {return m_upscale_multiplier>1;)?

gregory38:
Purely horizontal and vertical lines will keep their relative size.

plugins/GSdx/GSRendererHW.h
gregory38: I won't say a lot. Only a couple of stuff. A function might be a good idea, it could be possible to use it at difference place.

FlatOutPS2: You can just use GetUpscaleMultiplier(), which returns m_upscale_multiplier.

FlatOutPS2:
Background in RRV now looks like software mode, but I think I saw a regression in an FMV. I'll check up on it tonight.

plugins/GSdx/GSRendererOGL.cpp
```diff
@@ -1329,7 +1329,25 @@ void GSRendererOGL::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 		vs_cb.TextureScale = vs_cb.TextureScale.xyxy(point_size);
 	}
 #endif
-	m_gs_sel.sprite = m_vt.m_primclass == GS_SPRITE_CLASS;
+	m_gs_sel.sprite = m_vt.m_primclass == GS_SPRITE_CLASS || m_vt.m_primclass == GS_LINE_CLASS;
+	if ((m_upscale_multiplier > 1) && (m_vt.m_primclass == GS_LINE_CLASS) && GLLoader::found_geometry_shader) {
```
ssakash: ``(m_upscale_multiplier > 1)`` means you have excluded custom resolutions, is that intended ?

gregory38:
Ok. I just reused the geometry shader to transform line in sprite. But I'm not sure it is 100% exact. I wasn't in the mood to create more shaders.

plugins/GSdx/GSRendererOGL.cpp
gregory38: Yes because I'm not sure how it will react.

FlatOutPS2:
I was checking some demo's and came accross this reqression. The white squares are meant to be lines.

v-rally 3 demo
[VRally3.zip](https://github.com/PCSX2/pcsx2/files/423719/VRally3.zip)

gregory38:
Oh easy I'm just stupid.

gregory38:
Ok after 2 try, I think I have something more correct. I didn't test your dump, but it ought to be fine now.

willkuer:
So there is no solution for diagonal lines? Can't you rotate the square?

gregory38:
There are solutions but there are more complex. I have some pending work for point too. 

And diagonal suffers less of aliasing, if a dev try to render a rectangle with diagonal lines, it ought to be shot at once. Is the rendering that bad with diagonal line ?

FlatOutPS2:
The new commit now reduces the thickness of the horizontal lines in V-Rally 3.

gregory38:
I guess the line aren't 100% horizontal. Or they're mixed with diagonal lines. I guess, I need to think on the math to properly transform those lines in an efficient way.

gregory38:
FWIW, I rebased the commit on master and I added a small hack to handle the point too
`UserHacks_unscale_point_line = 1`. It can help on some FMV that use point renderer (dirge of cerberus).

gregory38:
Got a hard time to remember my math. But I managed a solution for all lines. It might miss an absolute call but it feels a good start.

gregory38:
I rebased and commited the code in master. New code is protected with the `UserHacks_unscale_point_line = 1`. It would be less build to test. Tell me if it render correctly.

MrCK1:
Hopefully this should help the Sly games. I'll take a look if I have time later

willkuer:
Why do you actually think point should be converted to square in upscaling? Maybe it should be circle instead? Probably circle shows more aliasing though?

gregory38:
May I ask you the same question : why circle?

willkuer:
I don't know how it is used. [Halftones](https://en.wikipedia.org/wiki/Halftone) are often circles.

Just when you said you wanted to unscale point as well I thought that this is really simple as you just replace it by a square. However a point has zero dimension. The intention of the game designer is thus lost.

Maybe stars drawn as points/pixels are intended to be circles. If they are now perfectly sharp squares it might look strange.

However square is probably better as a matrix of circles will probably produce something like moiree patterns while points might produce a closed surface... I don't know.

From a mathematical point of view I have the feeling that circle is a better representation due to the isotropy. However screens and pixels are not isotropic so maybe that doesn't matter.

FlatOutPS2:
Why was the closed, and then merged outside of the PR? :S

gregory38:
To reduce number of build to test and the number of pending PR.

For the circle,  I need to look at the sampling. But a Point is 0 or 1 pixels at native resolution. And a pixel is square. So a square feels better otherwise there will be aliasing. And it will be more complex to generate another pattern. Note even at 3x, you only have 3x3 pixels, it will be hard to do a nice circle.

FlatOutPS2:
> To reduce number of build to test and the number of pending PR. Code won't break anything at nothing is enabled by default

But why merge it outside of the PR? This way people will think it's been closed and not merged and less likely to test it. Unless you read the PR and know it's been merged outside of it, nobody would think to use the hidden option. 

And why not add a GUI option? Hidden options are hidden in use as well, unless you look at the ini(which people rarely do), you have no way of knowing if the option is turned on and you run the risk of spending an hour debugging before you realise an issue/bug is caused by a hidden ini option you forgot to turn off weeks ago.

gregory38:
I merged the branch, but I did a rebase. So I guess github didn't match the 2. Last time it was working good but without a rebase.

I don't know how to add option in Windows GUI. I didn't do it on linux because I'm waiting feedback from Windows users.

Issue #1438, opened 2016-06-29
GSDX: Flames are incorrectly rendererd on Ben 10 Alien Force : Vilgax Attacks


ssakash:
**Description of the Issue**:

The flames and sparks are incorrectly rendered on the following game on all of the hardware renderers. However, they're rendered correctly on the software renderer.

**Comparison Pictures**:

* ``OpenGL (Hardware mode)``

![Ben10_vilgax_hardware_mode.png](http://ultraimg.com/images/2016/06/29/Ben10_vilgax_hardware_mode.png)

* ``Software Renderer``

![Ben10_vilgax_software_mode.png](http://ultraimg.com/images/2016/06/29/Ben10_vilgax_software_mode.png)

* The spark issue would be a bit hard to notice just by looking at the pictures, here's a GIF to easily  notice it.

![GIF_Ben10_Vilgax_Attacks.gif](http://ultraimg.com/images/2016/06/29/GIF_Ben10_Vilgax_Attacks.gif)

**Test Case**: [**Master branch**](https://github.com/PCSX2/pcsx2) with head at https://github.com/PCSX2/pcsx2/commit/1f4f55bcc9cf6ecf4b9109e76a06c5e7e3123857

**GS Dump**: [**Download Link**](https://www.dropbox.com/s/oscogtyqko1f759/Ben10_Vilgax_Attacks_Flame_issue.rar?dl=0)

prafullpcsx2:
Most probably it's a regression caused by https://github.com/PCSX2/pcsx2/commit/59ef668fd2d0d599b3667e715a102f756ac05dd3

ssakash:
Confirmed , #1222 is the culprit. Nice work on finding it prafull !

gregory38:
Hum, I would be very lucky but it might worth to check #1434

ssakash:
#1434 doesn't fix the issue.

FWIW, I tried to bisect the issue by checking out the status of detached head:

``git checkout  d922225e692690d7d7bb6fe5e835c06beddb50cd``
**HEAD at https://github.com/PCSX2/pcsx2/pull/1222/commits/d922225e692690d7d7bb6fe5e835c06beddb50cd** : The issue is present.

`` git checkout 9aea76b0e80b87a0362263a1437cd7476e18138a``
**HEAD at https://github.com/PCSX2/pcsx2/pull/1222/commits/9aea76b0e80b87a0362263a1437cd7476e18138a** : The issue doesn't exist.

[...]

PR #1522, opened 2016-08-18
gsdx tc: don't partially invalidate small target

gregory38:
It creates a regression on game that uses a small temporary target to
upload textures of various sizes. Inital code was done to handle direct
frame write (background, FMV) so big target

See #1438


gregory38:
Need to replay my FMV dump.

PR #1523, opened 2016-08-18
Make settings dialog usable with 1024x768 resolution

nE0sIghT:
Hi,

Currently PCSX2 settings dialog is not usable with 1024x768 resolution because Speedhacks and Gamefixes panels exceed 768 pixels height.

While Gamefixes can be split in 2 columns, it's hard to pack Speedhacks panel properly to fit both 1024 pixel width and 768 pixel height.

This PR:
1. Removes redundant vertical line and whitespace of slider messages from Speedhacks panel.
2. Makes pxCheckboxes height smaller to remove redundant whitespace from Speedhacks and Gamefixes panels.
3. Adds scrollbars to Speedhacks and Gamefixes panels.

This issue was originally [reported at Gentoo bugzilla](https://bugs.gentoo.org/show_bug.cgi?id=591316).

### DIFF
--- a/pcsx2/gui/Dialogs/SysConfigDialog.cpp
+++ b/pcsx2/gui/Dialogs/SysConfigDialog.cpp
@@ -229,6 +229,21 @@ Dialogs::SysConfigDialog::SysConfigDialog(wxWindow* parent)
 	AddOkCancel();
 	AddPresetsControl();
 
+	// Add scrollbars to Speedhacks and Gamefixes panels if dialog size
+	// exceed desktop size.
+	// Assume user have panels with 50px size in total.
+	if(GetBestSize().GetWidth() >= (wxSystemSettings::GetMetric(wxSYS_SCREEN_X) - 50) ||
+		GetBestSize().GetHeight() >= (wxSystemSettings::GetMetric(wxSYS_SCREEN_Y) - 50))
+	{
+		for(int i = 4; i <= 5; i++)
+		{
+			wxPanelWithHelpers* panel = (wxPanelWithHelpers*) m_listbook->GetPage(i);
+			panel->InvalidateBestSize();
+			panel->FitInside();
+			panel->SetScrollRate(10, 10);
+		}
+	}
+
 	SetSizerAndFit(GetSizer());
 
 	if( wxGetApp().Overrides.HasCustomHacks() )
--- a/common/include/Utilities/wxGuiTools.h
+++ b/common/include/Utilities/wxGuiTools.h
@@ -553,7 +553,7 @@ protected:
 //    matically adjusts the width based on the sizer type (groupsizers get truncated to
 //    account for borders).
 //
-class wxPanelWithHelpers : public wxPanel
+class wxPanelWithHelpers : public wxScrolled<wxPanel>
 {
 	DECLARE_DYNAMIC_CLASS_NO_COPY(wxPanelWithHelpers)
 
--- a/common/src/Utilities/wxHelpers.cpp
+++ b/common/src/Utilities/wxHelpers.cpp
@@ -434,27 +434,27 @@ pxStaticText& wxPanelWithHelpers::Heading( const wxString& label )
 }
 
 wxPanelWithHelpers::wxPanelWithHelpers( wxWindow* parent, wxOrientation orient, const wxString& staticBoxLabel )
-	: wxPanel( parent )
+	: wxScrolled<wxPanel>( parent )
 {
 	SetSizer( new wxStaticBoxSizer( orient, this, staticBoxLabel ) );
 	Init();
 }
 
 wxPanelWithHelpers::wxPanelWithHelpers( wxWindow* parent, wxOrientation orient )
-	: wxPanel( parent )
+	: wxScrolled<wxPanel>( parent )
 {
 	SetSizer( new wxBoxSizer( orient ) );
 	Init();
 }
 
 wxPanelWithHelpers::wxPanelWithHelpers( wxWindow* parent )
-	: wxPanel( parent )
+	: wxScrolled<wxPanel>( parent )
 {
 	Init();
 }
 
 wxPanelWithHelpers::wxPanelWithHelpers( wxWindow* parent, const wxPoint& pos, const wxSize& size )
-	: wxPanel( parent, wxID_ANY, pos, size )
+	: wxScrolled<wxPanel>( parent, wxID_ANY, pos, size )
 {
 	Init();
 }


gregory38:
Wouldn't point 3 be enough ? I'm afraid that whitespace removal will cause issue with translated string.

nE0sIghT:
I think it will be enought

gregory38:
I give a try to your change (on linux). And 2 things bother me.
* My mouse wheel doesn't act on the scrollbar of the speedhack setting (likely intercepted by the slider widget). I'm sure people will complain.
* On my 24'' monitor, I have a tiny window with scrollbar. It would be nice if scrollbar were only enabled on small resolution display. I don't know if it doable.

nE0sIghT:
>  It would be nice if scrollbar were only enabled on small resolution display. I don't know if it doable.

For that I think we can use ([wxSystemSettings::GetMetric(wxSYS_SCREEN_Y)](http://docs.wxwidgets.org/trunk/classwx_system_settings.html#aa18e3b5794dc4193c4b0668d28d4933a) < 960) condition

turtleli:
> For that I think we can use (wxSystemSettings::GetMetric(wxSYS_SCREEN_Y) < 960) condition

960 might be a bit big, considering the dialog just about fits if only one panel is used on a 1024x768 LXDE environment (I checked with a Fedora VM).

The scrollbars might need to be non-Windows only - on Windows 10 the emulation settings dialog only has a height of about 600 pixels. It seems skipping the call to SetScrollRate() is sufficient for the dialog not to shrink in height and for there to be no scrollbars (the same appears to be true on Linux, but I mention the Windows situation in case you don't have a Windows environment).

nE0sIghT:
> 960 might be a bit big

In GNOME3 [it is 996x911](http://storage6.static.itmages.com/i/16/0819/h_1471590849_5508348_9bc211149d.png).

I think we can compare effective dialog size with desktop size and add scrollbars only if dialog exceeds desktop.
There is several methods to get size of wxWindow. For some reason GetSize() returns 20x20 for me, but GetEffectiveMinSize returns a bit smaller values than real.

nE0sIghT:
Here is new attempt, however it's not perfect.
For some reason dialog window is not resized (but dialog content resized), so I get gray space at the bottom.

Call to Fit() did not helps.

Screenshot

turtleli:
> In GNOME3 it is 996x911.

That's a bit surprising. I know there's some variation with different GTK themes and WM/DEs but I didn't expect that it could get so big.

> Here is new attempt, however it's not perfect.
> For some reason dialog window is not resized (but dialog content resized), so I get gray space at the bottom.
> This behavior happens because I use GetBestSize() prior to SetSizerAndFit().
> 
> Call to Fit() did not helps.

I'm not sure if it would work, but try calling InvalidateBestSize().

nE0sIghT:
> I'm not sure if it would work, but try calling InvalidateBestSize().

This do not works also

pcsx2/gui/Dialogs/SysConfigDialog.cpp
```diff
@@ -229,6 +229,20 @@ Dialogs::SysConfigDialog::SysConfigDialog(wxWindow* parent)
...
 
+	// Add scrollbars to Speedhacks and Gamefixes panels if dialog size
+	// exceed desktop size.
+	// Assume user have panels with 50px size in total.
+	if(GetBestSize().GetWidth() >= (wxSystemSettings::GetMetric(wxSYS_SCREEN_X) - 50) ||
+		GetBestSize().GetHeight() >= (wxSystemSettings::GetMetric(wxSYS_SCREEN_Y) - 50))
+	{
+		for(int i = 4; i <= 5; i++)
+		{
+			wxPanelWithHelpers* panel = (wxPanelWithHelpers*) m_listbook->GetPage(i);
```
turtleli: Maybe try adding `panel->InvalidateBestSize();` here?

nE0sIghT: That is works! Thanks

turtleli:
In general I think this is fine. It would probably be better for the code to apply to all panels (in case we add stuff and it doesn't fit any more), but from my experiments with the code, it doesn't seem like that's a simple thing.

I'll give a few more days for people (and myself) to test/review before merging.

nE0sIghT:
I still investigating mouse wheel event stealing for Speedhack panel

turtleli:
Oops. bad memory ;)

Possibly the scroll events are intercepted by the child windows and you need to pass the event on.

nE0sIghT:
> Possibly the scroll events are intercepted by the child windows and you need to pass the event on.

The problem is that I make all wxPanelWithHelpers objects scrollable.
Thus many PCSX2 controls begins to steal scroll events.

I'm not so good in C++ so I can not figure best solution for this with minimal edits.

willkuer:
maybe derive class `class wxPanelWithHelpersScrollable : public wxScrolled<wxPanelWithHelpers> ` instead and changing the corresponding parent panels to this new class?

nE0sIghT:
With ```wxScrolled<wxPanelWithHelpers>``` I'm getting many compilation errors hard for me to resolve

gregory38:
What is the current status of the PR ?

nE0sIghT:
I still investigating proper way to use ```wxScrolled<wxPanelWithHelpers>```

gregory38:
@turtleli could you help @nE0sIghT to have a nice solution for the next release? Thanks you.

jcdenton2k:
Question:  Why do we even need to support this resolution?  I'm just curious.  Even my old display from 10+ years ago was 1280 x 1024 resolution.  My current one is 1680 x 1050 and a few years old.

I know certain Linux programs allow for the lower-right corner to be dragged around to resize things as needed.  Any possibility of this being done for PCSX2?  Might solve a lot of these really low-res issues.

nE0sIghT:
> Question: Why do we even need to support this resolution?

Some users need this: https://bugs.gentoo.org/show_bug.cgi?id=591316#c5
Also GSDX support such low resolutions.

refractionpcsx2:
I don't see the point in supporting 1024x768, 1.78% of users (according to steam) have that resolution, even 1280x1024 is only 4%, a large majority of users have the resolution of 1366x768 or 1920x1080.  1024x768 was last seen as a resolution to use when CRT monitors were common, they are now obsolete and haven't been sold in stores for at least 7 years :P If your monitor only supports that, I really think it's time you upgraded.

nE0sIghT:
> a large majority of users have the resolution of 1366x768 

This resolution is also affected by this problem as I remember

gregory38:
Yes the issue is the 768 lines...

refractionpcsx2:
Right okay, that's fine then.

jcdenton2k:
I would like a response to the 'resizeable window' bit.  Is it possible to do this in Linux or Windows side?  Or both?

I can get some video of how it works but anyone using Linux Mint can see it easily.  If you open the System Settings Control Center, you can dynamically resize the window any way you like and it will automatically rearrange the items within the window itself.

I'm unaware of whether or not this is possible with the current setup but it might be nice to put on the wishlist if nothing else.

turtleli:
> turtleli could you help nE0sIghT to have a nice solution for the next release? Thanks you.

I'll look into it, but there probably isn't an easy way of fixing the scrolling issue (I tried a few weeks ago, didn't find a solution).

> I would like a response to the 'resizeable window' bit. Is it possible to do this in Linux or Windows side? Or both?

It can be done, but I doubt it'd be easy.

> I honestly believe PCSX2 might even be better off just dropping support for mobile chipsets altogether. As in the program would detect your hardware and warn you that your hardware isn't supported and that you won't receive technical support for it if you have issues. Just a thought.
> 
> This could be applied to low resolution displays as well as lower-end hardware.

Hmm. How about no?

gregory38:
Ok. Good luck :)

gregory38:
Can you tell me what is the status of this PR ? What remain to be fixed ? If I understand correctly, scrollbar will only be added for small screen. Is it correct?

nE0sIghT:
> Can you tell me what is the status of this PR ? What remain to be fixed ?

No properly solution found to make wxPanelWithHelpers scrollable only for some panels and in the same time do not steal scroll events when it should not be scrollable.
I think the fix should be very easy by hooking scroll events/configuration via class constructor/initializer however I stuck here.

> If I understand correctly, scrollbar will only be added for small screen. Is it correct?

It's correct. However AFAIK wxPanelWithHelpers's children will steal scroll events everywhere - don't know if it bad for PCSX2 GUI, but it's definitely wrong.

gregory38:
In this case, would it be possible to reduce the space in the EE Cyclerate/VU Cycle Stealing frame?
(maybe 4 would be enough)
```
diff --git a/pcsx2/gui/Panels/SpeedhacksPanel.cpp b/pcsx2/gui/Panels/SpeedhacksPanel.cpp
index 449b247..ac5fa2d 100644
--- a/pcsx2/gui/Panels/SpeedhacksPanel.cpp
+++ b/pcsx2/gui/Panels/SpeedhacksPanel.cpp
@@ -142,7 +142,7 @@ Panels::SpeedHacksPanel::SpeedHacksPanel( wxWindow* parent )
        wxDefaultPosition, wxDefaultSize, wxHORIZONTAL | wxSL_AUTOTICKS | wxSL_LABELS );
 
    m_msg_eecycle = new pxStaticHeading( m_eeSliderPanel );
-   m_msg_eecycle->SetHeight(5);
+   m_msg_eecycle->SetHeight(3);
 
    const wxChar* ee_tooltip = pxEt( L"Setting lower values on this slider effectively reduces the clock speed of the EmotionEngine's R5900 core cpu, and typically brings big speedups to games that fail to utilize the full potential of the real PS2 hardware. Conversely, higher values effectively increase the clock speed which may bring about an increase in in-game FPS while also making games more demanding and possibly causing glitches."
    );
@@ -159,7 +159,7 @@ Panels::SpeedHacksPanel::SpeedHacksPanel( wxWindow* parent )
        wxHORIZONTAL | wxSL_AUTOTICKS | wxSL_LABELS );
 
    m_msg_vustealer = new pxStaticHeading(m_vuSliderPanel);
-   m_msg_vustealer->SetHeight(5);
+   m_msg_vustealer->SetHeight(3);
 
    const wxChar* vu_tooltip = pxEt( L"This slider controls the amount of cycles the VU unit steals from the EmotionEngine.  Higher values increase the number of cycles stolen from the EE for each VU microprogram the game runs."
    );
```

It would remain the issue of the game fix panel.

gregory38:
I quickly try the pxGetFixedFont and it doesn't work as expected. I just got a bigger space between combo box...

nE0sIghT:
I was unable to get proper window size by reformatting controls and panels last time I tried.
My first attempt was reformatting and second - scrollbars.

gregory38:
There is a big trick. The various object are wrapper around the real object. So ->SetFont don't call the font of the checkbox but instead the font of something else. So yeah I'm not sure it is a good idea.

gregory38:
Maybe it would be better to use a new widget
* http://docs.wxwidgets.org/3.0/classwx_check_list_box.html (inherit for next one but I'm not sure if style can inherited)
* http://docs.wxwidgets.org/3.0/classwx_list_box.html (support scroll bar)

gregory38:
However, we will have a single massive tooltip rather plenty of small tooltip (Very bad for translation)

@turtleli any opinion on wx check list box ? Do you think it worth a try ? The wx code use a wxListBox internally so scrollbar will be supported. And a gtk_tree, so potentially it will be more tightly packed.

turtleli:
It might be worth trying. I'm not sure tooltips in the gamefixes panel are that useful anyway (only some items have tooltips, the tooltips only show some affected games, and users should prefer using automatic gamefixes).

The height on the EE cyclerate and VU stealing can't be reduced though, otherwise the messages might be truncated on Windows (I don't remember whether it affects Linux too).

gregory38:
```
The height on the EE cyclerate and VU stealing can't be reduced though, otherwise the messages might be truncated on Windows (I don't remember whether it affects Linux too).
```
156f573842d730520a490b0d7d83cc50833e8f4e maybe 4 could work. Otherwise maybe reduce 1 size on the font. Hum, Hum. I don't know if you already notice it but Jake rewrote a couple of wx object behavior (px code). For example pxStaticText, maybe it doesn't work well with long line.
```
// This class's purpose is to overcome two fundamental annoyances in wxWidgets design:
//
//   * An inability to wrap text to conform to a fitted window (a limitation imposed by
//     wxWidgets inability to fit individual directions, ie fit widths and then fit heights,
//     which would allow a textbox to wrap text to a sizer-determined width, and then grow
//     the sizers vertically to fit the calculated text-wrapped height).
//
//   * Textbox alignment requires aligning both the textbox contents, and aligning the text
//     control within it's containing sizer.  If both alignment flags do not match the result
//     is typically undesirable.
//
```
Typically it used by 
```
pcsx2/gui/Panels/SpeedhacksPanel.cpp:	m_msg_eecycle = new pxStaticHeading( m_eeSliderPanel );
pcsx2/gui/Panels/SpeedhacksPanel.cpp:	m_msg_vustealer = new pxStaticHeading(m_vuSliderPanel);
pcsx2/gui/Dialogs/FirstTimeWizard.cpp:	m_page_bios		+= new pxStaticHeading( &m_page_bios,
```
And I'm pretty sure those 3 strings suffers of the cut off annoyance.

gregory38:
@turtleli  I tried to hack the ee slider with a standard wxStaticText. It might miss a space between text and slider but potentially it doesn't cut anymore. Could you give it a shot on windows, thanks.
http://pastebin.com/DC8PkWvM

turtleli:
It almost works, though the height doesn't self adjust unless a Layout() call is used.

```
diff --git a/pcsx2/gui/Panels/SpeedhacksPanel.cpp b/pcsx2/gui/Panels/SpeedhacksPanel.cpp
index 449b247..28fdbec 100644
--- a/pcsx2/gui/Panels/SpeedhacksPanel.cpp
+++ b/pcsx2/gui/Panels/SpeedhacksPanel.cpp
@@ -369,6 +370,7 @@ void Panels::SpeedHacksPanel::Defaults_Click( wxCommandEvent& evt )
 void Panels::SpeedHacksPanel::EECycleRate_Scroll(wxScrollEvent &event)
 {
        SetEEcycleSliderMsg();
+       Layout();
        event.Skip();
 }

```

gregory38:
Well, I'm a complete noob on gui stuff.

I was looking at the implementation of wx (to compare it with our). When you call SetLabel(), it invalidates the size and get a new size namely `SetSize( GetBestSize() );`
I try on my side but it doesn't work. Then looking more at the code, something is bugging me.

* DoGetBestSize <= GetBestWrappedSize <= compute an ideal width based on the parent size. Then wrap the string based on the width and get the size.
* `_updateWrapping` will really do the wrapping but based on the current width not the ideal width

With some printf I found ideal width is 284 but the wrappint size is 260... So the area computed for the size isn't based on the real wrapping of the string.

gregory38:
Note for myself:
Need to test New width += padding (instead of -) in updateWrapping.

gregory38:
Ok I fount it. I need to check the string in game fixes then I will do a PR to remove some spaces. Then we could resume the discussion on game fixes check list.

gregory38:
@turtleli 
On the gamefix panel
```
	*this	+= m_check_Enable	| StdExpand();

```
StdExpand will increase the width of the subText (below the checkbox). It will compute a new wrapping. However, the height is the same before/after the wrapping (StdExpand()) call. Hence the useless space. I really don't know what must be done to resize it properly. Any idea of what could be missing/wrong?

turtleli:
I don't have a clue. That space below the text has always confused me.

gregory38:
pcsx2_coresettings_game fixes
Here what happen when you remove the `StdExpand()` As you see it the same height, it seems like expand trigger expand the text but doesn't trigger a new computation of the "box" around the text.

turtleli:
Ok, I see where the space comes from now. I'm completely guessing, but maybe the height of the subtext needs to be invalidated or explicitly set?

gregory38:
I try to call a test procedure after the expand
```
void pxCheckBox::Test()
{
	Console.WriteLn("Old size %d", m_subtext->GetBestSize().x);
	m_subtext->InvalidateBestSize();
	m_subtext->Layout();
	Layout();
	Console.WriteLn("New size %d", m_subtext->GetBestSize().x);
	//m_subtext->SetHeight(1);
	Console.WriteLn("SH size %d", m_subtext->GetBestSize().x);
}
```
SetHeight uses a size in line where I can automatically compute the size in pixels. Maybe I just need to get the number of pixel of the font.

gregory38:
OoO,

I think I understand what happen. If you don't expand the text, you will see that width is the size of the main menu width.

wrap_width


So in pxStaticText::GetBestWrappedSize loop I tried to skip the menu frame with this hack
```
diff --git a/common/src/Utilities/pxStaticText.cpp b/common/src/Utilities/pxStaticText.cpp
index 26dc7e9..046c222 100644
--- a/common/src/Utilities/pxStaticText.cpp
+++ b/common/src/Utilities/pxStaticText.cpp
@@ -173,6 +174,7 @@ wxSize pxStaticText::GetBestWrappedSize( const wxClientDC& dc ) const
        // to play nice and NOT resize according to the new min size.  (wtf stupid)
        // Anyway, this fixes it -- ignore min size specifier on wxWizard!
        if( wxIsKindOf( millrun, wxWizard ) ) break;
+       if( wxIsKindOf( millrun, wxFrame ) && millrun->GetParent() == nullptr ) break;
 
        int min = (int)((millrun->GetMinWidth() - parentalAdjust) * parentalFactor);
 
```

after_hack

Width explode but heigh is smaller (not easy to see).

gregory38:
Based on a list box.
pcsx2_coresettings_game fixes

Need to add  back a tooltip (merge of previous tooltip). And maybe disable the scrollbar on big screen (need this pr code)

shentino:
Anyone alive?  Downstream asked me to poke you guys

PR #1524, opened 2016-08-18
EE: Minor changes to syscall function

ssakash:
**Summary of changes**:

* Move interlace/progressive & Frame/Field detection code to ``SetGsCrt()`` instead of relying on SMODE2 registers.
* Add some nice enums to improve code readability on ``SYSCALL()``.

### DIFF
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -34,7 +34,6 @@
 using namespace Threading;
 
 extern u8 psxhblankgate;
-extern bool gsIsInterlaced;
 static const uint EECNT_FUTURE_TARGET = 0x10000000;
 static int gates = 0;
 
--- a/pcsx2/R5900OpcodeImpl.cpp
+++ b/pcsx2/R5900OpcodeImpl.cpp
@@ -25,6 +25,7 @@
 #include "GS.h"
 
 GS_VideoMode gsVideoMode = GS_VideoMode::Uninitialized;
+bool gsIsInterlaced = false;
 
 static __fi bool _add64_Overflow( s64 x, s64 y, s64 &ret )
 {
@@ -884,8 +885,13 @@ void SYSCALL()
 
 	switch (call) {
 		case 2: {
-					const char* inter = (cpuRegs.GPR.n.a0.UL[0] & 1) ? "Interlaced" : "Progressive";
-					const char* field = (cpuRegs.GPR.n.a2.UL[0] & 1) ? "FRAME" : "FIELD";
+					//Function "SetGsCrt(Interlace, Mode, Field)"
+					//Useful for fetching information of interlace/video/field display parameters of the Graphics Synthesizer
+
+					gsIsInterlaced = cpuRegs.GPR.n.a0.UL[0] & 1;
+					bool gsIsFrameMode = cpuRegs.GPR.n.a2.UL[0] & 1;
+					const char* inter = (gsIsInterlaced) ? "Interlaced" : "Progressive";
+					const char* field = (gsIsFrameMode) ? "FRAME" : "FIELD";
 					std::string mode;
 					// Warning info might be incorrect!
 					switch (cpuRegs.GPR.n.a1.UC[0])
--- a/pcsx2/R5900OpcodeImpl.cpp
+++ b/pcsx2/R5900OpcodeImpl.cpp
@@ -883,8 +883,11 @@ void SYSCALL()
 
 	BIOS_LOG("Bios call: %s (%x)", R5900::bios[call], call);
 
-	switch (call) {
-		case 2: {
+
+	switch (static_cast<Syscall>(call))
+	{
+		case Syscall::SetGsCrt:
+		{
 					//Function "SetGsCrt(Interlace, Mode, Field)"
 					//Useful for fetching information of interlace/video/field display parameters of the Graphics Synthesizer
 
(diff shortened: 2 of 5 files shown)


pcsx2/Counters.cpp
```diff
@@ -34,7 +34,6 @@
 using namespace Threading;
 
 extern u8 psxhblankgate;
-extern bool gsIsInterlaced;
```
refractionpcsx2: I'm assuming that now won't compile. (what did I tell you about testing? :P )

Also below where you removed it from GS.h you will need to move that somewhere relevant as well.

ssakash: > I'm assuming that now won't compile. (what did I tell you about testing? :P )

Appveyor/TravisCI seems to be happy. :)

> Also below where you removed it from GS.h you will need to move that somewhere relevant as well.

I didn't remove anything from ``GS.h``. I assume you mean the ``gsIsInterlaced`` variable  ? It has been moved to get initialized at the SetGsCrt function.

refractionpcsx2: Sorry yes, you moved it, my bad :P

ssakash: > I'm guessing it's okay because GS.h is included in the counters?

Yeah, you're right.

gregory38:
May I suggest to reset the video mode on reset (if not already done).
For the enum, it would be nice to add some numbers (=10, =20 ....) (or bigger step).

willkuer:
Is there actually a reason for 632 enum entries but only five are actually handled in the switch? Does that mean that 99% of operations are not emulated. Or does it mean that these are handled somewhere else in the code where still the numerical flags 0x7F and so on are used and where it might be reasonable to use the same enum?

pcsx2/R5900OpcodeImpl.cpp
```diff
@@ -883,8 +883,10 @@ void SYSCALL()
 
 	BIOS_LOG("Bios call: %s (%x)", R5900::bios[call], call);
```
gregory38: Does it compile in debug mode ? What happen if you enable the bios log trace (in the F10 menu)

gregory38:
The kernel of the PS2 supports 256 syscalls. However, there are handled currently by the kernel code not by "us directly". When an enter syscall function is executed, we take the opportunity to hook it up to do extra jobs. I'm not sure we need all the entry is the enum. Even an HLE implementation will just use an array of function pointer. So we could reduce the enum to the useful value only.

pcsx2/R5900OpcodeImpl.cpp
ssakash: Yes, it compiles fine and logging does work.

ssakash:
> Does that mean that 99% of operations are not emulated. 

As Gregory said already, code is handled by the kernel. We only use this ``SYSCALL()`` to gather useful information (or) do some additional stuffs like in ``Deci2Call()``.

pcsx2/R5900OpcodeImpl.cpp
```diff
@@ -882,10 +883,17 @@ void SYSCALL()
...
-					const char* inter = (cpuRegs.GPR.n.a0.UL[0] & 1) ? "Interlaced" : "Progressive";
-					const char* field = (cpuRegs.GPR.n.a2.UL[0] & 1) ? "FRAME" : "FIELD";
+
+	switch (static_cast<Syscall>(call)) {
+		case Syscall::SetGsCrt:
+		{
+					//Function "SetGsCrt(Interlace, Mode, Field)"
+					//Useful for fetching information of interlace/video/field display parameters of the Graphics Synthesizer
+
+					bool IsInterlaced = gsIsInterlaced = cpuRegs.GPR.n.a0.UL[0] & 1;
```
gregory38: `IsInterlaced` is useless now? You could use gsIsInterface directly, couldn't you?

pcsx2/R5900OpcodeImpl.cpp
```diff
@@ -975,6 +973,17 @@ void SYSCALL()
 			}
 			break;
 
+		case Syscall::Deci2Call:
```
gregory38: move of code was kind of useless with naming enum

gregory38:
Ok with the 2 small comments code is reviewed by me. I let's you update code accordingly.

gregory38:
By the way @ssakash goes easy on tag. Otherwise github will have more tag than code source file. Goal is to make group not to replace title ;)

refractionpcsx2:
> By the way @ssakash goes easy on tag.

yeh "Core" would have done :P

pcsx2/R5900OpcodeImpl.cpp
```diff
@@ -882,10 +883,17 @@ void SYSCALL()
...
 
-	switch (call) {
-		case 2: {
-					const char* inter = (cpuRegs.GPR.n.a0.UL[0] & 1) ? "Interlaced" : "Progressive";
-					const char* field = (cpuRegs.GPR.n.a2.UL[0] & 1) ? "FRAME" : "FIELD";
+
+	switch (static_cast<Syscall>(call)) {
+		case Syscall::SetGsCrt:
+		{
+					//Function "SetGsCrt(Interlace, Mode, Field)"
```
turtleli: Indentation is rather strange for this case (issue existed before PR).

pcsx2/R5900OpcodeImpl.cpp
ssakash: Some IDE's show the enum value as tooltip when hovering over the enum, I prefer having it sorted.

gregory38: Oh, kind of an useless feature. Name are better than number. But fine.

PR #297, opened 2014-09-11
OSX and 64-bit compatibility testing

aktau:
This is just a branch I'm using to try to get as far as possible in compiling things on 64-bit OSX. Up until now, not even one file gets compiled before it errors out. But, the number of errors is diminishing. Currently looking at things like:

**UPDATE 4**: Slowly but surely getting there. Current error listed below

Summary:

| Library | Builds | Works |
| --- | --- | --- |
| libUtilities | Yes | ? |
| libx86emitter | Yes | ? |
| pcsx2 core | In progress | ? |

```
[ 58%] Building CXX object pcsx2/CMakeFiles/pcsx2-dbg.dir/gui/AppMain.cpp.o
In file included from /Users/Kevin/github/aktau/pcsx2/pcsx2/gui/AppMain.cpp:17:
/Users/Kevin/github/aktau/pcsx2/pcsx2/gui/MainFrame.h:133:15: warning: 'MainEmuFrame::DispatchEvent' hides overloaded virtual function [-Woverloaded-virtual]
        virtual void DispatchEvent( const PluginEventType& plugin_evt );
                     ^
/Users/Kevin/github/aktau/pcsx2/pcsx2/gui/AppEventListeners.h:146:15: note: hidden overloaded virtual function 'IEventListener_AppStatus::DispatchEvent' declared here: type mismatch at 1st parameter ('c
onst AppEventInfo &' vs 'const PluginEventType &')
        virtual void DispatchEvent( const AppEventInfo& evtinfo );
                     ^
In file included from /Users/Kevin/github/aktau/pcsx2/pcsx2/gui/AppMain.cpp:17:
/Users/Kevin/github/aktau/pcsx2/pcsx2/gui/MainFrame.h:134:15: warning: 'MainEmuFrame::DispatchEvent' hides overloaded virtual function [-Woverloaded-virtual]
        virtual void DispatchEvent( const CoreThreadStatus& status );
                     ^
/Users/Kevin/github/aktau/pcsx2/pcsx2/gui/AppEventListeners.h:146:15: note: hidden overloaded virtual function 'IEventListener_AppStatus::DispatchEvent' declared here: type mismatch at 1st parameter ('c
onst AppEventInfo &' vs 'const CoreThreadStatus &')
        virtual void DispatchEvent( const AppEventInfo& evtinfo );
                     ^
/Users/Kevin/github/aktau/pcsx2/pcsx2/gui/AppMain.cpp:238:3: error: Unsupported Target Platform.
[...]

PR #1528, opened 2016-08-20
build.sh: simplify and semi-modernize

aktau:
1. All POSIX shells support $(...) syntax [1], including /bin/sh. shellcheck
   warns about it.
2. [[ won't work in /bin/sh [2], so use [ everywhere. I wonder why it worked
   now, perhaps the test was running on a system where /bin/sh -> /bin/bash.
3. In POSIX sh, string indexing is undefined. [SC2039]. Unfortunately, this
   means we require a subprocess: https://wiki.ubuntu.com/DashAsBinSh. Very
   ugly.
4. In POSIX sh, arrays are undefined. We seem to use $flags as an array
   after constructing it by string concatenation. I tried to verify that
   this has the same effect as just passing the quoted string in bash:

      bash-3.2$ flags="-DCMAKE_GOOK"
      bash-3.2$ flags="$flags -DCMAKE_MOARMA"
      bash-3.2$ flags="$flags -DCMAKE_URURURUR"
      bash-3.2$ ./argv $flags
      0: ./argv
      1: -DCMAKE_GOOK
      2: -DCMAKE_MOARMA
      3: -DCMAKE_URURURUR
      bash-3.2$ ./argv "${flags[@]}"
      0: ./argv
      1: -DCMAKE_GOOK -DCMAKE_MOARMA -DCMAKE_URURURUR
      bash-3.2$ ./argv "$flags"
      0: ./argv
      1: -DCMAKE_GOOK -DCMAKE_MOARMA -DCMAKE_URURURUR
      bash-3.2$

5. Enable exit on unknown variable (-u). All variables should be known,
   otherwise we have an error in the script. shellcheck doesn't warn so I
   think it's fine.

Apart from shellcheck(1), I also ran checkbashisms(1). The latter only
reported that "command -v" might not be available in other shells.
Apparently only ash(1) doesn't understand it.

NOTE: Why are we even trying to support pre-Mavericks (Darwin < 13) OSX? We
don't even support the most modern OSX (El Capitan) fully yet. OSX upgrades
are free and generally don't leave old machines behind. Most machines made
after 2009 can upgrade to El Capitan, AFAIK. I also believe that systems
that have all the utilities and libraries necessary to build PCSX2 will have
/bin/bash >= 3.x.

NOTE 2: Does cmake/ninja generate the same type of output in
compile_commands.json?

[1]: http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_03
[2]: http://serverfault.com/a/52050

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/sh -u
 
 # PCSX2 - PS2 Emulator for PCs
 # Copyright (C) 2002-2014  PCSX2 Dev Team
@@ -33,14 +33,15 @@ build="$root/build"
 coverity_dir="cov-int"
 coverity_result=pcsx2-coverity.xz
 
-if [ `command -v ninja` ]; then
+if command -v ninja ; then
     flags="$flags -GNinja"
 fi
 
-if [ `uname -s` = 'Darwin' ]; then
-    ncpu=`sysctl -n hw.ncpu`
-    release=$(uname -r)
-    if [[ ${release:0:2} -lt 13 ]]; then
+if [ "$(uname -s)" = 'Darwin' ]; then
+    ncpu="$(sysctl -n hw.ncpu)"
+
+    # Get the major Darwin/OSX version.
+    if [ "$(sysctl -n kern.osrelease | cut -d . -f 1)" -lt 13 ]; then
         echo "This old OSX version is not supported! Build will fail."
         toolfile=cmake/darwin-compiler-i386-clang.cmake
     else
@@ -48,7 +49,7 @@ if [ `uname -s` = 'Darwin' ]; then
         toolfile=cmake/darwin13-compiler-i386-clang.cmake
     fi
 else
-    ncpu=`grep -w -c processor /proc/cpuinfo`
+    ncpu=$(grep -w -c processor /proc/cpuinfo)
     toolfile=cmake/linux-compiler-i386-multilib.cmake
 fi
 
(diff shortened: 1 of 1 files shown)


mirh:
Are you going to still keep #297 open?

And then I dunno if OSX is more like linux (newer == better) or more like Windows (old can be better), but I guess you'll know if pre-maverick is more like XP or more like 7.

aktau:
> Are you going to still keep #297 open?

I was under the impression the majority of those changes were first imported by @juhalaukkanen and then merged via one of their PRs. There's no real need to keep exactly that one open, although I dream of resuming that work at one point. I remember getting stuck at some assembly that wasn't allowed on x86-64 (but was allowed on x86).

> And then I dunno if OSX is more like linux (newer == better) or more like Windows (old can be better), but I guess you'll know if pre-maverick is more like XP or more like 7 to keep supported.

It's a UNIX, so more like Linux. The point is that OSX was never supported by PCSX2, so it's now easy to say that only versions 10.11+ will be supported.

gregory38:
on 1/ why not? but why is is better ? `$(..)` needs more char and harder to highligh properly in editor.

Script used to be sh. But then I got convinced by bad people that bash was much better. I put back sh recently  because some users got issue to use the file. I didn't use any tool to validate it so some bad stuffs remain (such as 2/ and 4/)

Good question on ninja json format.

build.sh
```diff
@@ -225,17 +225,16 @@ if [ "$clangTidy" -eq 1 ] && [ `command -v clang-tidy` ]; then
         #clang-tidy -p $compile_json $cpp -checks='*' -header-filter='.*'
 
         # Don't check header, don't check google/llvm coding conventions
-        echo "$count/$total"
```
gregory38: It would be nice to fix the code. Or add a progress bar, I don't know if there is any standard stuff.

gregory38:
Did you test command -v for command that don't exist. Just to validate the syntax without `[ ]`

aktau:
Yes I did, in /bin/sh :)

build.sh
gregory38: Or a better solution will be to multithread it a bit. Half an hour to get result is damn slow.

gregory38:
Ok perfect.

mirh:
@aktau Yes, I imagined one day you'd have liked to make it advance, but I was wondering whether that got definitively surpassed (and new work would basically require all new commits) or if on the other hand you wasn't keeping it open because there's still something worth.

build.sh
aktau: I agree. There's an easy solution that doesn't require a lot (or any) synchronization: we can use GNU parallel (xargs would parallelize but also mangle the output). I'll submit a PR.

aktau:
> @aktau Yes, I imagined one day you'd have liked to make it advance, but I was wondering whether that got definitively surpassed (and new work would basically require all new commits) or if on the other hand you wasn't keeping it open because there's still something worth.

I'm not sure if _all_ of the commits I had got merged into @juhalaukkanen's branch, and it's difficult to see at a glance because they squashed those commits into bigger ones. I guess I'll notice whenever I try to rebase my branch. In any case, the most important work was already merged.

PR #1529, opened 2016-08-20
build.sh: parallelize clang-tidy

aktau:
Also suppress spurious output from command -v.

As requested.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -33,7 +33,7 @@ build="$root/build"
 coverity_dir="cov-int"
 coverity_result=pcsx2-coverity.xz
 
-if command -v ninja ; then
+if command -v ninja >/dev/null ; then
     flags="$flags -GNinja"
 fi
 
@@ -170,7 +170,7 @@ fi
 ############################################################
 # CPP check build
 ############################################################
-if [ "$cppcheck" -eq 1 ] && command -v cppcheck ; then
+if [ "$cppcheck" -eq 1 ] && command -v cppcheck >/dev/null ; then
     summary=cpp_check_summary.log
     rm -f $summary
     touch $summary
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -206,35 +206,40 @@ if [ "$clangTidy" -eq 1 ] && command -v clang-tidy ; then
...
-
-        # Check all, likely severals millions of log...
-        #clang-tidy -p $compile_json $cpp -checks='*' -header-filter='.*'
-
-        # Don't check header, don't check google/llvm coding conventions
-        clang-tidy -p $compile_json "$cpp" -checks='*,-llvm-*,-google-*'  >> $summary
-    done < $cpp_list
+    if command -v parallel >/dev/null ; then
+        # Run clang-tidy in parallel with as many jobs as there are CPUs.
+        parallel clang-tidy -p $compile_json "{}" -checks='*,-llvm-*,-google-*' < $cpp_list > $summary
```
gregory38: oh very nice. Did you use the `>` instead of `>>` ? I thinks it won't work if shell defines no clober.

aktau: Indeed, but /bin/sh doesn't define noclobber by default. We only enable `set -u`, which exits if an unset variable is referenced. Anyway, I've made it clearer by moving the `rm/touch` two-step to the `else`-block, where it belongs. Updated.

gregory38: sh maybe. But if you have a bash that load a .bashrc file that uses this option. Or another shell.

aktau: Ah, but the shebang at the top of the file is `#!/bin/sh` (you changed it recently, which is why I finished the POSIXification in an earlier commit). Which means it will not load `.bashrc`. Moreover, when starting `/bin/sh` like this, that's a non-interactive shell, I do not believe that any `rc` file is loaded: https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html

If it were true that user preferences in `.bashrc` or any other `*rc` file could have such an effect on scripts that aren't incredibly defensively written, very few things would work as expected.

gregory38:
```
/bin/sh -x ./build.sh --dbg --clean --no-simd --clang-tidy --clang

+DEBUG+>'[' 1 -eq 1 ']'
+DEBUG+>command -v clang-tidy
+DEBUG+>compile_json=compile_commands.json
+DEBUG+>cpp_list=cpp_file.txt
+DEBUG+>summary=clang_tidy_summary.txt
+DEBUG+>grep '"file"' compile_commands.json
+DEBUG+>sed -e 's/"//g' -e 's/^\s*file\s*:\s*//'
+DEBUG+>command -v parallel
+DEBUG+>parallel clang-tidy -p compile_commands.json '{}' '-checks=*,-llvm-*,-google-*'
zsh:1: no matches found: -checks=*,-llvm-*,-google-*
zsh:1: no matches found: -checks=*,-llvm-*,-google-*
```
For the record, echo $SHELL => /bin/zsh.

aktau:
> For the record, echo $SHELL => /bin/zsh.
> Edit: escaping * to \* seem to do the trick

Yea, my bad :). parallel basically runs the provided command in a shell, but the provided command was already parsed once, so all the quotes are gone. I'll fix it.

aktau:
Done. Please give it another try.

aktau:
I've changed it even more, getting rid of the `>>` indirection in the serial leg. I had forgotten one can indirect if/else statements both ways as well.

gregory38:
Ok cool . It takes 6 minutes to generate clang tidy info instead of more than 30 minutes. I was expected a bigger gain with my 8c/16t CPU. But at least it is usable now.

aktau:
OK, I believe this PR is now in its final form. I added the `-m` flag to the parallel invocation to divides the files over all clang-tidy instances. This basically means that clang-tidy won't be respawned for every file, which avoids a small start-up cost.

I've also resplaced the `while read` loop with a xargs invocation. It's possible to parallelize this as well, but it will mix the output, which might interleave files from different invocations, so I don't enabled parallelizing.

gregory38:
Oups sorry I push the trigger a bit too soon.

aktau:
> I was expected a bigger gain with my 8c/16t CPU. But at least it is usable now.

I suspect it's running 8 jobs then. Perhaps you can try artificially incrementing that with `-j16` or even `-j32`. I suspect it won't help too much, clang-tidy isn't bottlenecked on IO, but on CPU. At least here on my machine.

aktau:
> Oups sorry I push the trigger a bit too soon.

No I think you have the correct one.

gregory38:
Ah yes you're right I have it.

It ran on 16 threads. But HT isn't real core

aktau:
Oh crap. My reasoning was off. The `-m` flag does avoid the clang-tidy startup cost (which isn't large), but it also increases tail latency because it allows a straggler command to run much longer. Suppose that many heavy .cpp files are bundled into one clang-tidy invocation.

Can you try removing `-m` and seeing if it's faster for you?

gregory38:
I can t do it right now. But I will do a test with time. Don't bother too much. The tool is slow nothing we can't do about it. (I won't trim down include files).
Potentially it could be faster with less generated warning (+400K in our code).

gregory38:
I pushed a new commit on your behalf. See the benchmark on the commit message. It saves a minute of my life by analyze :+1:

aktau:
> It saves a minute of my life by analyze 👍

Great to hear ;).

aktau:
By the way, while I was doing this, I thought of another way to massively speed this up. Generally, you don't want to be running clang-tidy on the entire codebase a lot, because the entire codebase doesn't change that often. So why don't you just run it on those that changed?

```make
# Totally untested, might have made some syntax missups. Put *.tidy files in .gitignore. _OR_ do it
# the hard way and put them somewhere in a /build directory, but you'll need some path-mangling
# in Make. You also need some extra trickery if you want header files in there (such as including 
# the _full_ path in the `%.tidy` target).
%.tidy: %.cpp
    clang-tidy $(TIDY_FLAGS) $< > $@

# A "make tidy" target. Make will only re-make the .tidy files that are older than the .cpp/.h files
# they were made from. This list can also be made with a glob.
tidy: something.tidy some-other-cpp-file.tidy ...
```

gregory38:
Honestly I'm fine with current solution. 5 minutes isn't a huge time. And you barely run it anyway. You generate once, fix some warnings and then validate it with a new run. I'm not sure that I will fix those warning as it could be quite time consuming and it is more oriented on robustness rather than perf. But I will use it to modernize a bit the code source after the reformatting mess.

PR #1530, opened 2016-08-20
Clang format plugin v1

gregory38:
First shot to reformat the code base to a common & consistent syntax.

In order to (barely) reduce blame annoyance, I use a Clang Format author.

If you don't like the syntax, it is your last chance to complain.

Note: I don't know what to do about windows rc file.

### DIFF
--- a/plugins/CDVDnull/CDVD.cpp
+++ b/plugins/CDVDnull/CDVD.cpp
@@ -27,102 +27,122 @@ const unsigned char version = PS2E_CDVD_VERSION;
 const unsigned char revision = 0;
 const unsigned char build = 6;
 
-EXPORT_C_(char*) PS2EgetLibName()
+EXPORT_C_(char*)
+PS2EgetLibName()
 {
-	snprintf( libraryName, 255, "CDVDnull Driver %lld%s",SVN_REV,	SVN_MODS ? "m" : "");
-	return libraryName;	
+    snprintf(libraryName, 255, "CDVDnull Driver %lld%s", SVN_REV, SVN_MODS ? "m" : "");
+    return libraryName;
 }
 
-EXPORT_C_(u32) PS2EgetLibType()
+EXPORT_C_(u32)
+PS2EgetLibType()
 {
-	return PS2E_LT_CDVD;
+    return PS2E_LT_CDVD;
 }
 
-EXPORT_C_(u32) CALLBACK PS2EgetLibVersion2(u32 type)
+EXPORT_C_(u32)
+CALLBACK PS2EgetLibVersion2(u32 type)
 {
-	return (version << 16) | (revision << 8) | build;
+    return (version << 16) | (revision << 8) | build;
 }
 
-EXPORT_C_(s32) CDVDinit()
+EXPORT_C_(s32)
+CDVDinit()
 {
-	return 0;
+    return 0;
 }
 
-EXPORT_C_(s32) CDVDopen(const char* pTitle)
+EXPORT_C_(s32)
+CDVDopen(const char* pTitle)
 {
-	return 0;
+    return 0;
 }
 
-EXPORT_C_(void) CDVDclose()
+EXPORT_C_(void)
+CDVDclose()
 {
 }
 
-EXPORT_C_(void) CDVDshutdown()
+EXPORT_C_(void)
+CDVDshutdown()
 {
 }
 
-EXPORT_C_(s32) CDVDreadTrack(u32 lsn, int mode)
+EXPORT_C_(s32)
+CDVDreadTrack(u32 lsn, int mode)
 {
-	return -1;
+    return -1;
 }
 
 // return can be NULL (for async modes)
-EXPORT_C_(u8*) CDVDgetBuffer()
+EXPORT_C_(u8*)
+CDVDgetBuffer()
 {
-	return NULL;
+    return NULL;
 }
 
-EXPORT_C_(s32) CDVDreadSubQ(u32 lsn, cdvdSubQ* subq)
+EXPORT_C_(s32)
+CDVDreadSubQ(u32 lsn, cdvdSubQ* subq)
 {
-	return -1;
+    return -1;
 }
 
-EXPORT_C_(s32) CDVDgetTN(cdvdTN *Buffer)
+EXPORT_C_(s32)
+CDVDgetTN(cdvdTN* Buffer)
 {
-	return -1;
+    return -1;
 }
 
-EXPORT_C_(s32) CDVDgetTD(u8 Track, cdvdTD *Buffer)
+EXPORT_C_(s32)
+CDVDgetTD(u8 Track, cdvdTD* Buffer)
 {
-	return -1;
+    return -1;
 }
 
-EXPORT_C_(s32) CDVDgetTOC(void* toc)
+EXPORT_C_(s32)
+CDVDgetTOC(void* toc)
 {
-	return -1;
+    return -1;
 }
 
-EXPORT_C_(s32) CDVDgetDiskType()
+EXPORT_C_(s32)
+CDVDgetDiskType()
 {
-	return CDVD_TYPE_NODISC;
+    return CDVD_TYPE_NODISC;
 }
 
-EXPORT_C_(s32) CDVDgetTrayStatus()
+EXPORT_C_(s32)
+CDVDgetTrayStatus()
 {
-	return CDVD_TRAY_CLOSE;
+    return CDVD_TRAY_CLOSE;
 }
 
-EXPORT_C_(s32) CDVDctrlTrayOpen()
+EXPORT_C_(s32)
+CDVDctrlTrayOpen()
 {
-	return 0;
+    return 0;
 }
 
-EXPORT_C_(s32) CDVDctrlTrayClose()
+EXPORT_C_(s32)
+CDVDctrlTrayClose()
 {
-	return 0;
+    return 0;
 }
 
-EXPORT_C_(void) CDVDconfigure()
+EXPORT_C_(void)
+CDVDconfigure()
 {
-	SysMessage("Nothing to Configure");
+    SysMessage("Nothing to Configure");
 }
 
-EXPORT_C_(void) CDVDabout()
+EXPORT_C_(void)
+CDVDabout()
 {
-	SysMessage("%s %d.%d", "CDVDnull Driver", revision, build);
+    SysMessage("%s %d.%d", "CDVDnull Driver", revision, build);
 }
 
-EXPORT_C_(s32) CDVDtest()
+EXPORT_C_(s32)
+CDVDtest()
 {
-	return 0;
+    return 0;
 }
--- a/plugins/CDVDnull/CDVD.h
+++ b/plugins/CDVDnull/CDVD.h
@@ -14,7 +14,7 @@
  */
 
 
- #ifndef __CDVD_H__
+#ifndef __CDVD_H__
 #define __CDVD_H__
 
 #define CDVDdefs
@@ -30,7 +30,7 @@
 #ifdef _MSC_VER
 #define EXPORT_C_(type) extern "C" type CALLBACK
 #else
-#define EXPORT_C_(type) extern "C" __attribute__((stdcall,externally_visible,visibility("default"))) type
+#define EXPORT_C_(type) extern "C" __attribute__((stdcall, externally_visible, visibility("default"))) type
 #endif
 
 extern const unsigned char version;
@@ -38,5 +38,5 @@ extern const unsigned char revision;
 extern const unsigned char build;
 extern const unsigned int minor;
 
-extern void SysMessage(const char *fmt, ...);
+extern void SysMessage(const char* fmt, ...);
 #endif /* __CDVD_H__ */
--- a/plugins/FWnull/Config.cpp
+++ b/plugins/FWnull/Config.cpp
@@ -25,55 +25,51 @@ PluginConf Ini;
 
 void setLoggingState()
 {
-	if (conf.Log)
-	{
-		FWLog.WriteToConsole = true;
-		FWLog.WriteToFile = true;
-	}
-	else
-	{
-		FWLog.WriteToConsole = false;
-		FWLog.WriteToFile = false;
-	}
+    if (conf.Log) {
+        FWLog.WriteToConsole = true;
+        FWLog.WriteToFile = true;
+    } else {
+        FWLog.WriteToConsole = false;
+        FWLog.WriteToFile = false;
+    }
 }
 
-EXPORT_C_(void) FWabout()
+EXPORT_C_(void)
+FWabout()
 {
-	SysMessage("FWnull: A simple null plugin.");
+    SysMessage("FWnull: A simple null plugin.");
 }
 
-EXPORT_C_(void) FWconfigure()
+EXPORT_C_(void)
+FWconfigure()
 {
-	LoadConfig();
-	PluginNullConfigure("Since this is a null plugin, all that is really configurable is logging.", conf.Log);
-	SaveConfig();
+    LoadConfig();
+    PluginNullConfigure("Since this is a null plugin, all that is really configurable is logging.", conf.Log);
+    SaveConfig();
 }
 
 void LoadConfig()
 {
-	string IniPath = s_strIniPath + "/FWnull.ini";
-	if (!Ini.Open(IniPath, READ_FILE))
-	{
-		FWLog.WriteLn("Failed to open %s", IniPath.c_str());
-		SaveConfig();
-		return;
-	}
+    string IniPath = s_strIniPath + "/FWnull.ini";
+    if (!Ini.Open(IniPath, READ_FILE)) {
+        FWLog.WriteLn("Failed to open %s", IniPath.c_str());
+        SaveConfig();
+        return;
+    }
 
-	conf.Log = Ini.ReadInt("logging", 0);
-	setLoggingState();
-	Ini.Close();
+    conf.Log = Ini.ReadInt("logging", 0);
+    setLoggingState();
+    Ini.Close();
 }
 
 void SaveConfig()
 {
-	string IniPath = s_strIniPath + "/FWnull.ini";
-	if (!Ini.Open(IniPath, WRITE_FILE))
-	{
-		FWLog.WriteLn("Failed to open %s\n", IniPath.c_str());
-		return;
-	}
+    string IniPath = s_strIniPath + "/FWnull.ini";
+    if (!Ini.Open(IniPath, WRITE_FILE)) {
+        FWLog.WriteLn("Failed to open %s\n", IniPath.c_str());
+        return;
+    }
 
-	Ini.WriteInt("logging", conf.Log);
-	Ini.Close();
+    Ini.WriteInt("logging", conf.Log);
+    Ini.Close();
 }
-
--- a/plugins/FWnull/Config.h
+++ b/plugins/FWnull/Config.h
@@ -15,7 +15,7 @@
 
 void SaveConf();
 void LoadConf();
-void SysMessage(const char *fmt, ...);
+void SysMessage(const char* fmt, ...);
 
 #define is_checked(main_widget, widget_name) (gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON(lookup_widget(main_widget, widget_name))))
-#define set_checked(main_widget,widget_name, state) gtk_toggle_button_set_state(GTK_TOGGLE_BUTTON(lookup_widget(main_widget, widget_name)), state)
+#define set_checked(main_widget, widget_name, state) gtk_toggle_button_set_state(GTK_TOGGLE_BUTTON(lookup_widget(main_widget, widget_name)), state)
--- a/plugins/FWnull/FW.cpp
+++ b/plugins/FWnull/FW.cpp
@@ -23,20 +23,20 @@ using namespace std;
 #include "FW.h"
 #include "svnrev.h"
 
-const u8 version  = PS2E_FW_VERSION;
+const u8 version = PS2E_FW_VERSION;
 const u8 revision = 0;
-const u8 build    = 7;    // increase that with each version
+const u8 build = 7;  // increase that with each version
 
 #ifdef _MSC_VER
 #define snprintf sprintf_s
 #endif
 static char libraryName[256];
 
-string s_strIniPath="inis";
+string s_strIniPath = "inis";
 string s_strLogPath = "logs";
 
 u8 phyregs[16];
-s8 *fwregs;
+s8* fwregs;
 Config conf;
 PluginLog FWLog;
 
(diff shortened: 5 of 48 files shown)


avih:
Is there a summary, in English, of the main things at the new preferred style?

Here's what I recognized from the first few changes:

- Function definition braces are on a new line.
- Code blocks braces start at the same line (like with `if` or `while` etc) [1].
- Uses 4 spaces indents instead of tabs.
- in pointer the `*` is adjacent to the variable: `char *p;` rather than `char* p;` or `char * p;`.
- Stuff before function definitions (types?) goes on its own line.
- Use single space consistently before and after operators.
- Don't use space after `(` or before `)`

That's what I noticed after looking at 8 files or so from this PR.

FWIW, I like all the style preferences above. Where's the full list though?

[1] I notice that some code was changed to use brace at the same line, but some code was not - even though the same block was modified for indents.

avih:
> in pointer the * is adjacent to the variable: char *p; rather than char* p; or char * p;.

Actually, its the other way around, e.g.:
```diff
-s8 *fwregs;
+s8* fwregs;
```

and TBH I prefer the `*` adjacent to the variable name, like `char *p;`, and I also think it's a more common style.

gregory38:
It is a good summary. You can see detail of the options here
http://clang.llvm.org/docs/ClangFormatStyleOptions.html + https://github.com/PCSX2/pcsx2/blob/master/.clang-format

I try to keep style as common as possible. That why I remove space around if. For function/class `{}` will be on next line. Actually it is the style of the linux kernel.

```
[1] I notice that some code was changed to use brace at the same line, but some code was not - even though the same block was modified for indents.
```
Where ? I allow to use a single line syntax.

Init list will also be split in multiple line. It you have more than 3 empty lines they will be merged to 3.

gregory38:
Oh I prefer the `char* p` rather than `char *p`. Need a vote of others.

avih:
> Where ? I allow to use a single line syntax.

This (and on other places) changes from separate line brace to the same line brace:
```diff
-	string IniPath = s_strIniPath + "/FWnull.ini";
-	if (!Ini.Open(IniPath, WRITE_FILE))
-	{
-		FWLog.WriteLn("Failed to open %s\n", IniPath.c_str());
-		return;
-	}
+    string IniPath = s_strIniPath + "/FWnull.ini";
+    if (!Ini.Open(IniPath, WRITE_FILE)) {
+        FWLog.WriteLn("Failed to open %s\n", IniPath.c_str());
+        return;
+    }
```

While this changes the block, but leaves it on a separate line:
```diff
-	PHYACC |= phyregs[reg] | (reg << 8);
-	
-	if(fwRu32(0x8424) & 0x40000000) //RRx interrupt mask
-	{
-		fwRu32(0x8420) |= 0x40000000;
-		FWirq();
-	}	
+    if (fwRu32(0x8424) & 0x40000000)  //RRx interrupt mask
+    {
+        fwRu32(0x8420) |= 0x40000000;
+        FWirq();
+    }
```


> Oh I prefer the char* p rather than char *p. Need a vote of others.

I'm fine with a vote, but really, most (all?) big projects which I'm familiar with use e.g. `char *p;`

gregory38:
Hum, I suspect it is related to the comment. I'm not sure it is an intended behavior. Dunno if there is a conf.

avih:
Yes, it could be related to the comment. I noticed few other places where it's left on a separate line and they had comments too (though I only looked at ~10% of the changes so far).

avih:
TBH though, while I also mostly prefer the new style (and has been using a similar style elsewhere too), I think that for such global change we should limit the changes to white spaces only, and same-line only.

I.e. keep the braces on the same line they were before, keep the function types on the same lines they were before, etc.

This would leave every line of code at exactly the same place it was and will make it easier for git to look past whitespace changes in blame, while still making all the indentation and indent char conform to the new style.

And for new code, sure, we can agree that we only use the new style.

I think it would be a good balance that way.

aktau:
> I think it's safer in general. Consider char* a, b; where it looks like both a and b are pointers, vs char *a, b; where it's clear that a is a pointer and b is not.

I must agree, I also prefer `*` close to the variable name, for this reason. I guess I started doing it after reading K&R a very long time ago.

> Stuff before function definitions (types?) goes on its own line.

If I get a vote, I'd vote against this...

> Function definition braces are on a new line.

...and this. The reason being that I like being able to read as much as possible on my often small terminal windows. I can't see a good reason for having the type and that brace on a separate line, it's a bit of a waste. Types aren't on a different line for variables either, why should they be for functions? The same for braces, apparently clang-format doesn't do that for if/else/for/while/switch/... either, why for functions?

```c++
int
main(int argc, char *argv[])
{
  return 0;
}
```

```c++
int main(int argc, char *argv[]) {
  return 0;
}
```

turtleli:
Hmm. It seems it doesn't actually put return types on a separate line. The parenthesis in `EXPORT_C_(type)` makes clang-format act weird, and the null plugins are full of those.

In general I do like the style, but I need to experiment with some of the clang-format options to see what I like/dislike.

gregory38:
Quoting the guy because I agree with him.
```
Bjarne Stroustrup said:

    The choice between "int* p;" and "int *p;" is not about right and wrong, but about
 style and emphasis. C emphasized expressions; declarations were often considered
 little more than a necessary evil. C++, on the other hand, has a heavy emphasis on types.

    A "typical C programmer" writes "int *p;" and explains it "*p is what is the int"
 emphasizing syntax, and may point to the C (and C++) declaration grammar
 to argue for the correctness of the style. Indeed, the * binds to the name p in the grammar.

    A "typical C++ programmer" writes "int* p;" and explains it "p is a pointer to an int"
 emphasizing type. Indeed the type of p is int*. I clearly prefer that emphasis
 and see it as important for using the more advanced parts of C++ well.
```

```
I think it's safer in general. Consider char* a, b; where it looks like both a and b are pointers, vs char *a, b; where it's clear that a is a pointer and b is not.
```
It is a fine example for C, but this pattern is unlikely in C++.
* Mixing pointer and normal type in confusing for both syntax.
* In modern C++, you often don't declare variable before affection (but there are some case)
* In modern C++, you ought to use smart pointer not raw pointer (when you need pointer)
Even if you find case, you need several raw pointers, rule is to put  a variable by line ;)



```
    Stuff before function definitions (types?) goes on its own line.
```
Ah no sorry. I put template before not type. Potentially attribute will be put before too or maybe the extern qualifier. I think it explain that EXPORT_C(...) is before. TBH, I found it nices because EXPORT_C is kind of special.



```
    Function definition braces are on a new line.

...and this. The reason being that I like being able to read as much as possible on my often small terminal windows. I can't see a good reason for having the type and that brace on a separate line, it's a bit of a waste. Types aren't on a different line for variables either, why should they be for functions? The same for braces, apparently clang-format doesn't do that for if/else/for/while/switch/... either, why for functions?
```
We could do it everywhere. Normally, you don't have lots of class/function by files, few gain so I follow the linux kernel doc here. Honestly I'm fine with both way.

For the record from linux doc
```
However, there is one special case, namely functions: they have the
opening brace at the beginning of the next line, thus:

	int function(int x)
	{
		body of function
	}

Heretic people all over the world have claimed that this inconsistency
is ...  well ...  inconsistent, but all right-thinking people know that
(a) K&R are _right_ and (b) K&R are right.  Besides, functions are
special anyway (you can't nest them in C).
```

```
TBH though, while I also mostly prefer the new style (and has been using a similar style elsewhere too), I think that for such global change we should limit the changes to white spaces only, and same-line only.
```
No. I won't do half backed change.
* Clutter history more (already changing most of the line due to space anyway)
* more difficult to handle as I can't configure the tool this way
* done once, and then subject is closed
* Most of the file will never be changed so it would remain a mess for a century

avih:
> No. I won't do half backed change.

Allow me to quote you from #1488:

> I would like to square the code database to uniformize coding style.
> 
>     At a bare-minimum, I would like to remove all trailing white space (small impact on history)
>     It would be nice to convert leading tab to space everywhere (big impact on history, depends on the part)
>     Reformat with an automatic tool such as astyle or clang-format (huge impact on history)

It's you who insists on the refactor, and it's you who wanted to do the whitespace changes and asked for opinion about the third one with "huge impact on history", and that's exactly the thing I want to avoid - that huge impact on history.

avih:
Just to make it clear, I fully prefer a style similar to this clang-format configuration over the current PCSX2 code style, which is all over the place. But I care more about the ability to track code changes _easily_ than about the style.

gregory38:
Oh. There was a misunderstanding. Either we do only white space change. Either we do all. If we do all, it is much better to do in one shot.

I'm in favor to do all. In my opinion, either we paid an expensive cost once and enjoy it. Or we paid it a small fee every day. So far we've decided the latter. And we're already paid more than the big initial cost.

Asking people to read a spec file when no file follow it will be hard. Machine jobs ought to be done by machine so we can concentrate on clever task.

History will be a bit harder to read. But with a good GUI, it is a matter of a few clicks to select the version before the reformatting.

Note: I think it is possible to have a style configure by directory. But I would prefer a single one for the project.

avih:
> either we paid an expensive cost once and enjoy it

It's far from once and that's exactly my concern. You pay it _every time_ you try to track code changes and bump into the reformat wall. Both tomorrow or a year from now or 5 years from now.

gregory38:
By once I mean, it will be clean for all future work. But yes it would be annoying to track code changes. However,
* you read/write/review code more often than you look at the history
* there are already severals format massive commit (white space, EOL).
* To limit the blame issue, you can use the `-w` option. It will drop the space change which are the biggest change of those commits (compare with `git diff --ignore-all-space`). Hum, need to look at this option for my GUI.

avih:
Anyway, we don't have to agree on everything :) so on this case let's agree to disagree. Both of us would accept whatever most devs prefer even if it's not our personal preference, so that seems to be the case here.

The question becomes: what do most devs prefer? full reformat and harder to track changes? or mostly white spaces reformat but easier to track changes?

Regardless, at least between those who voiced their opinions, baring minor points, there's an agreement that we should use a style similar to this clang-format config, and that it's better than the current/no style which the code has now.

If we end up with mostly white spaces changes, of course, when one touches code with the old style, it should be changed to the new style.

aktau:
It actually seems like a lot of the codebase already has bad blame output. I randomly opened pcsx2/VUmicro.cpp:

```
$ git blame --line-porcelain VUmicro.cpp | awk '/^author /{print hash, $0} {hash = $1}' | cut -f1,3 -d' ' | sort | uniq -c | sort -nr
  94 a05e9c62be64acfe9f2640ef8e68414e9a6a3a03 sudonim1
   7 3c478fed8fcd7450f5a8765b91ebd1379afa0cc2 ramapcsx2
   6 3cfd0c68af15cdda7bb83927e555c071d92d57ce Jake.Stine
   1 70d47bf240e7c3f9630a08880a11f45727d20ee0 Jake.Stine
   1 376e373137bcea992823d8b5f546d69c74b6a2f4 ramapcsx2

$ git blame VUmicro.cpp | head -25
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000   1) /*  PCSX2 - PS2 Emulator for PCs
70d47bf2 (Jake.Stine 2010-05-03 14:08:02 +0000   2)  *  Copyright (C) 2002-2010  PCSX2 Dev Team
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000   3)  *
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000   4)  *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000   5)  *  of the GNU Lesser General Public License as published by the Free Software Found-
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000   6)  *  ation, either version 3 of the License, or (at your option) any later version.
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000   7)  *
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000   8)  *  PCSX2 is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000   9)  *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  10)  *  PURPOSE.  See the GNU General Public License for more details.
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  11)  *
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  12)  *  You should have received a copy of the GNU General Public License along with PCSX2.
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  13)  *  If not, see <http://www.gnu.org/licenses/>.
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  14)  */
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  15)
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  16) #include "PrecompiledHeader.h"
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  17) #include "Common.h"
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  18) #include "VUmicro.h"
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  19)
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  20) #define useDeltaTime 1
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  21)
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  22) #if useDeltaTime
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  23)
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  24) // Executes a Block based on EE delta time
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  25) void BaseVUmicroCPU::ExecuteBlock(bool startUp) {
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  26)    	const u32& stat	= VU0.VI[REG_VPU_STAT].UL;
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  27)    	const int  test = m_Idx ? 0x100 : 1;
376e3731 (ramapcsx2  2012-11-12 20:48:55 +0000  28)    	const int  s = 1024*8; // Kick Start Cycles (Silver Surfer needs this amount)
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  29)    	const int  c = 1024*1; // Continue Cycles
a05e9c62 (sudonim1   2010-04-24 21:37:39 +0000  30)    	if (!(stat & test)) return;
```

sudonim1 Seems to be the main author of this file from the git blame view. Seems plausible, this was in 2010. Let's see:

```
$ git log a05e9c62 VUmicro.cpp
commit a05e9c62be64acfe9f2640ef8e68414e9a6a3a03
Author: sudonim1 <sudonim1@96395faa-99c1-11dd-bbfe-3dabce05a288>
Date:   Sat Apr 24 21:37:39 2010 +0000

    Set svn:eol-style=native on all *.c *.cpp *.h

    git-svn-id: http://pcsx2.googlecode.com/svn/trunk@2894 96395faa-99c1-11dd-bbfe-3dabce05a288
```

Arg, a big formatting change. Which covers most of the file.

Luckily, that's pretty easy to solve in some editors. Vim has [vim-fugitive](https://github.com/tpope/vim-fugitive). With it, you can go to any file and type `:Gblame`. This will open a split screen window similar to the command line `git blame` output, but highlighted. If you don't want to see this specific style commit, you can go to a line which that commit touched, and press `P`. This will re-blame at that commit - 1. Which means that you can see who was responsible for those lines before the style change.

It seem to me this style change is a smaller impediment than first though for blame layering.

gregory38:
Let's be honest we don't have any real history.
Lots of improvements are code drop with massive changeset. Remaining commits are reformatting. That why I feel an extra reformatting can still be handled without too much pain.

I need to look at the vim plugin. I'm using qgit (qt gui for git) currently which do something similar.

turtleli:
I prefer a full reformat. Otherwise the code will still be a multi-style mess (though to a lesser extent) and the code will likely stay that way for years if the code doesn't need changes.

Looking into the style a bit, I'm not sure the `Allow*OnASingleLine: true` is that nice. Code consistency might be better with options set to false.

**Original**
```
void TGE()  { if (cpuRegs.GPR.r[_Rs_].SD[0] >= cpuRegs.GPR.r[_Rt_].SD[0]) trap(_TrapCode_); }
void TGEU() { if (cpuRegs.GPR.r[_Rs_].UD[0] >= cpuRegs.GPR.r[_Rt_].UD[0]) trap(_TrapCode_); }
```
**AllowShortIfStatementsOnASingleLine: true**
```
void TGE()
{
    if (cpuRegs.GPR.r[_Rs_].SD[0] >= cpuRegs.GPR.r[_Rt_].SD[0]) trap(_TrapCode_);
}
void TGEU()
{
    if (cpuRegs.GPR.r[_Rs_].UD[0] >= cpuRegs.GPR.r[_Rt_].UD[0]) trap(_TrapCode_);
}
```
**AllowShortIfStatementsOnASingleLine: false**
```
void TGE()
{
    if (cpuRegs.GPR.r[_Rs_].SD[0] >= cpuRegs.GPR.r[_Rt_].SD[0])
        trap(_TrapCode_);
}
void TGEU()
{
    if (cpuRegs.GPR.r[_Rs_].UD[0] >= cpuRegs.GPR.r[_Rt_].UD[0])
        trap(_TrapCode_);
}
```

**AllowShortFunctionsOnASingleLine** seems to only allow simple single statement functions on one line, above functions don't seem to qualify.

**AllowShortBlocksOnASingleLine: true**  doesn't seem to work? I even used the documentation test case and still ended up with multiple lines (unless another option is interfering).

**Original**
```
void mVUloadReg(const xmm& reg, xAddressVoid ptr, int xyzw)
{
	switch( xyzw ) {
		case 8:		xMOVSSZX(reg, ptr32[ptr]);		break; // X
		case 4:		xMOVSSZX(reg, ptr32[ptr+4]);	break; // Y
		case 2:		xMOVSSZX(reg, ptr32[ptr+8]);	break; // Z
		case 1:		xMOVSSZX(reg, ptr32[ptr+12]);	break; // W
		default:	xMOVAPS (reg, ptr128[ptr]);		break;
	}
}
```
**AllowShortCaseLabelsOnASingleLine: true**
```
void mVUloadReg(const xmm& reg, xAddressVoid ptr, int xyzw)
{
    switch (xyzw) {
        case 8:
            xMOVSSZX(reg, ptr32[ptr]);
            break;  // X
        case 4:
            xMOVSSZX(reg, ptr32[ptr + 4]);
            break;  // Y
        case 2:
            xMOVSSZX(reg, ptr32[ptr + 8]);
            break;  // Z
        case 1:
            xMOVSSZX(reg, ptr32[ptr + 12]);
            break;  // W
        default: xMOVAPS(reg, ptr128[ptr]); break;
    }
}
```
**AllowShortCaseLabelsOnASingleLine: false**
```
void mVUloadReg(const xmm& reg, xAddressVoid ptr, int xyzw)
{
    switch (xyzw) {
        case 8:
            xMOVSSZX(reg, ptr32[ptr]);
            break;  // X
        case 4:
            xMOVSSZX(reg, ptr32[ptr + 4]);
            break;  // Y
        case 2:
            xMOVSSZX(reg, ptr32[ptr + 8]);
            break;  // Z
        case 1:
            xMOVSSZX(reg, ptr32[ptr + 12]);
            break;  // W
        default:
            xMOVAPS(reg, ptr128[ptr]);
            break;
    }
}
```
In the true case it allows two statements per case before reformatting it. It also seems to format differently depending on comments or not.

gregory38:
@turtleli 
For the case did you test without the comment? Do you want that I disable all `Allow*OnASingleLine` ?

Any opinion on the `{}` for functions/classes

ssakash:
Personally I also prefer a full format. It would make tracking older changes via Git blame a bit hard but it's still feasible to track them down.

About ``char* val`` **vs** ``char *val`` - I prefer the former :)

About braces placement, I prefer having the braces placed on a new line.
```
void clear()
{
   active_state = false;
}
```

instead of something like this,

```
void clear(){
    active_state = false;
}
```

turtleli:
**AllowShortCaseLabelsOnASingleLine: true** (comments removed)
```
void mVUloadReg(const xmm& reg, xAddressVoid ptr, int xyzw)
{
    switch (xyzw) {
        case 8: xMOVSSZX(reg, ptr32[ptr]); break;
        case 4: xMOVSSZX(reg, ptr32[ptr + 4]); break;
        case 2: xMOVSSZX(reg, ptr32[ptr + 8]); break;
        case 1: xMOVSSZX(reg, ptr32[ptr + 12]); break;
        default: xMOVAPS(reg, ptr128[ptr]); break;
    }
}
```
So comments do make a difference to the formatting.

> Do you want that I disable all Allow*OnASingleLine ?

IMO clang-format can probably work more consistently with all `AllowShort*OnASingleLine` disabled, so yes. `AllowShortFunctionsOnASingleLine: All` might be ok, but I need to figure out the behaviour (will post about it when I do).

if/switch/for/while braces: Strongly prefer `{` on the same line. More or less the same reason as aktau - it saves space when I'm looking at the code with tmux + vim + other stuff.
Function/class braces: Prefer on separate line (current style), but fine either way. Technically I should prefer on the same line for the same reason as the if/switch stuff, but hey, I'm inconsistent ;)

Oh, it might be better to not reformat resource.h, since that file will get overwritten if someone makes a change using the Visual Studio resource editor. Maybe.

gregory38:
For single line function check on header. Allow to write.
```
auto foo() {return bar; }
```
Instead of
```
auto foo()
{
return bar;
}

**Comment by turtleli** [comment]
Original (from various places)
```
void EnableDepth() { m_depth_enable = true; }
void SetDepth(GLenum func, bool mask) { m_depth_func = func; m_depth_mask = mask; }
bool IsMaskEnable() { return m_depth_mask != GL_FALSE; }
const OPCODE& Class_REGIMM(u32 op)  { return tbl_RegImm[(op >> 16) & 0x1F]; }
void TGE()  { if (cpuRegs.GPR.r[_Rs_].SD[0] >= cpuRegs.GPR.r[_Rt_].SD[0]) trap(_TrapCode_); }
```
**AllowShortFunctionsOnASingleLine: All**
```
void EnableDepth() { m_depth_enable = true; }
void SetDepth(GLenum func, bool mask)
{
    m_depth_func = func;
    m_depth_mask = mask;
}
bool IsMaskEnable() { return m_depth_mask != GL_FALSE; }
const OPCODE& Class_REGIMM(u32 op) { return tbl_RegImm[(op >> 16) & 0x1F]; }
void TGE()
{
    if (cpuRegs.GPR.r[_Rs_].SD[0] >= cpuRegs.GPR.r[_Rt_].SD[0])
        trap(_TrapCode_);
}
```
So 1 simple statement (no if/while etc) - same line, everything else - multiple lines. Comments don't seem to affect the behaviour.

**Comment by gregory38** [comment]
So do we want to allow it or no ?

**Comment by turtleli** [comment]
I think it'd be ok.

**Comment by gregory38** [comment]
Ok. Could you play with namespace indentation `NamespaceIndentation` . I set it to none. Only inner doesn't seem to work nicely. And indentation become big in file with 3-4 nested namespace.

**Comment by aktau** [comment]
> Ok. Could you play with namespace indentation NamespaceIndentation . I set it to none. Only inner doesn't seem to work nicely. And indentation become big in file with 3-4 nested namespace.
 Show all checks

Most big, heavily namespaced projects I've worked on seem to prefer no-indentation (not even for the last level). 80 columns rule is a bit archaic but some form of it is definitely handy for code readability when you're viewing files side-by-side.

**Comment by turtleli** [comment]
Inner seems to indent only the innermost namespace. Reading the documentation, I was expecting it to indent all but the outermost namespace. All and None work as intended.

I don't think it's been mentioned, but struct and class braces are placed differently with the Linux style.
```
struct structname {
};
class classname
{
};
```

In addition to braces being inconsistent when comments are involved, it seems array formatting can be inconsistent too. Note the trailing commas and `{` placement.
```
static const OPCODE tbl_MMI1CUT1[4] =
{
	MMI_Unknown, PABSW, PCEQW, PMINW
};

static const OPCODE tbl_MMI1CUT2[4] =
{
	MMI_Unknown, PABSW, PCEQW, PMINW,
};

static const OPCODE tbl_MMI1CUT3[4] = {
	MMI_Unknown, PABSW, PCEQW, PMINW
};

static const OPCODE tbl_MMI1CUT4[4] = {
	MMI_Unknown, PABSW, PCEQW, PMINW,
};

static const OPCODE tbl_MMI1CUT5[4] = { MMI_Unknown, PABSW,
                                        PCEQW, PMINW };
static const OPCODE tbl_MMI1CUT6[4] = { MMI_Unknown, PABSW,
                                        PCEQW, PMINW,};
```
Reformatted:
```
static const OPCODE tbl_MMI1CUT1[4] =
    {
        MMI_Unknown, PABSW, PCEQW, PMINW};

static const OPCODE tbl_MMI1CUT2[4] =
    {
        MMI_Unknown, PABSW, PCEQW, PMINW,
};

static const OPCODE tbl_MMI1CUT3[4] = {
    MMI_Unknown, PABSW, PCEQW, PMINW};

static const OPCODE tbl_MMI1CUT4[4] = {
    MMI_Unknown, PABSW, PCEQW, PMINW,
};

static const OPCODE tbl_MMI1CUT5[4] = {MMI_Unknown, PABSW,
                                       PCEQW, PMINW};
static const OPCODE tbl_MMI1CUT6[4] = {
    MMI_Unknown, PABSW,
    PCEQW, PMINW,
};
```

**Comment by gregory38** [comment]
Hum, I guess we will need to use custom brace options.
```
    bool AfterClass Wrap class definitions.
    bool AfterControlStatement Wrap control statements (if/for/while/switch/..).
    bool AfterEnum Wrap enum definitions.
    bool AfterFunction Wrap function definitions.
    bool AfterNamespace Wrap namespace definitions.
    bool AfterObjCDeclaration Wrap ObjC definitions (@autoreleasepool, interfaces, ..).
    bool AfterStruct Wrap struct definitions.
    bool AfterUnion Wrap union definitions.
    bool BeforeCatch Wrap before catch.
    bool BeforeElse Wrap before else.
    bool IndentBraces Indent the wrapped braces themselves.
```

For the array, could it be linked to the following option ?
```
Cpp11BracedListStyle (bool)

    If true, format braced lists as best suited for C++11 braced lists.

    Important differences: - No spaces inside the braced list. - No line break before the closing brace. - Indentation with the continuation indent, not with the block indent.

    Fundamentally, C++11 braced lists are formatted exactly like function calls would be formatted in their place. If the braced list follows a name (e.g. a type or variable name), clang-format formats as if the {} were the parentheses of a function call with that name. If there is no name, a zero-length name is assumed.
```
My guess is that *Construtor* are for constructor of object.

I don't know neither if `Penalty*` options have any impact.

**Comment by turtleli** [comment]
Yes, Cpp11BracedListStyle does affect it. Changing the option to false:
```
static const OPCODE tbl_MMI1CUT1[4] =
    {
      MMI_Unknown, PABSW, PCEQW, PMINW
    };

static const OPCODE tbl_MMI1CUT2[4] =
    {
      MMI_Unknown, PABSW, PCEQW, PMINW,
    };

static const OPCODE tbl_MMI1CUT3[4] = {
    MMI_Unknown, PABSW, PCEQW, PMINW
};

static const OPCODE tbl_MMI1CUT4[4] = {
    MMI_Unknown, PABSW, PCEQW, PMINW,
};

static const OPCODE tbl_MMI1CUT5[4] = { MMI_Unknown, PABSW,
                                        PCEQW, PMINW };
static const OPCODE tbl_MMI1CUT6[4] = {
    MMI_Unknown, PABSW,
    PCEQW, PMINW,
};
```


I think the `Penalty*` options are only used if `ColumnLimit` is set to something other than 0.

**Comment by gregory38** [comment]
Hum, might worth to check the behavior on `pcsx2/x86/R5900_Profiler.h`
I'm sure clang won't like it ^^

**Comment by willkuer** [comment]
As this pr is about consistency you should probably also force consistency in the array initialization. I also don't see the point of adding the comma if you have more than one element per line...

**Comment by gregory38** [comment]
Honestly, some array initialization looks much better when done by an human (with proper vertical alignment). TBH, I won't be against to disable clang formatting with a comment for those cases.

But yes, we need to do a manual rules on the trailing comma. The coma helps copy/past when you write code vertically. Otherwise it is useless.
And also on the braces placement. It sucks there isn't a nice automatic way to format it in a cool way.

**Comment by gregory38** [comment]
@turtleli What do we do for the array init stuff ?

Additional question on the brace.
```
    try {
        psxRec.Reserve();
    } catch (Exception::RuntimeError& ex) {
```
vs
```
    try {
        psxRec.Reserve();
    }
    catch (Exception::RuntimeError& ex) {
```

And by the way, did you play with the `ReflowComments` option. I'm not sure what it does exactly.

**Comment by turtleli** [comment]
> What do we do for the array init stuff ?

Dunno. Probably leaving `Cpp11BracedListStyle` as true is fine. I think we might have to reformat some of the arrays manually whether it's true or false, which will be a pain.

I'm not opposed to using `// clang-format off` for some arrays where it makes a difference (the GSdx blend tables come to mind).

> Additional question on the brace.

I prefer the first one.

> And by the way, did you play with the ReflowComments option. I'm not sure what it does exactly.

It splits comments into multiple lines if it's too long. It only works if `ColumnLimit` isn't 0. Should we limit line length?

**Comment by gregory38** [comment]
```
Should we limit line length?
```
So far, I think it is a bad idea. It will make change way bigger, at least we can ignore white space in the diff. Besides I'm afraid it will break line in unexpected place. Potentially we could still do an incremental update of the format later to break line.

**Comment by gregory38** [comment]
I rebased with latest format update. Update is rather small on those null plugins. A couple of if was split in 2 lines. And a break before `{` for struct. Next step will be to integrate properly check_format/git hook. This way we can test the script with the null plugin.

Oh crap I set the wrong author on the commit message. Damn. I will redo it tomorrow.

**Comment by gregory38** [comment]
Ok let's merge this first round of update so we can setup travis and git properly.

**Comment by gregory38** [comment]
```
it's a misunderstanding of the language specification
```
You're overreacting, aren't you. One of the (the?) C++ creator said `int* ` was better for him. So it is a matter of opinion.

**Comment by willkuer** [comment]
I guess the question is not how it suppose to be in modern c++ but maybe rather how it is implemented.

Avih's example of `char* a, b;` vs `char *a, b;` is really illustrative. Sure that's not modern c++ but nevertheless this is how it is implemented.

I totally agree with you that `char* a;` looks and feels better than `char *a;` and indeed in the "modern" c# implementation [it is exactly used like that](https://msdn.microsoft.com/en-us/library/y31yhkeb.aspx) and the asterisk is part of the type.

Still for some reason c++ tries to stick with violence to c backwards compatibility. And I would rather say you should use the language how it is implemented and not how you want it to be.


(maybe just avoid stupid pointers at all ;) )

**Comment by avih** [comment]
@gregory38 at this point is seems that by far more people prefer the `*` to be adjacent to the variable rather than to the type. Don't force your own preference on something which other care about just because it's you who pushed that commit.

**Comment by willkuer** [comment]
I think it is currently you and pseudo against ssakash and gregory. Maybe @turtleli, @ramapcsx2 and @refractionpcsx2 want to add their opinion and you get a nice democratic decision.

**Comment by gregory38** [comment]
The major goal is to attract new contributors. New contributor want modern C++. And soon C++11 won't even be the legacy.

I know some code that use this pattern: C-GTK stuff for the linux GUI. Otherwise in C++ code, raw pointer is close to be deprecated. I'm playing with clang-tidy which a linter tool and it reports plenty of basic pointer operation as bad practice (based on cpp core guideline). So previous pattern is 100% relevant in C but not that much in C++.

```
Still for some reason c++ tries to stick with violence to c backwards compatibility. And I would rather say you should use the language how it is implemented and not how you want it to be.
```
I don't think it is true anymore. C+11 and future version don't share C paradigm. I mean C++ is oriented toward robustness.
```
pointers are unsafe => reference
macros are unsafe => template
union are unsafe => boost::variant (likely for C++17)
variadic arguments are unsafe => variadic template
array are unsafe => STL with bound check
memory allocation (mem leak) => smart pointer
C cast => dynamic_cast
....
```

Anyway, I will love to have others opinions :)

**Comment by avih** [comment]
> The major goal ...

People stated their preference and arguments, and others are still welcome to do so too. But they will not keep the argument going forever, and unless they said so explicitly, it means their mind has not changed.

So far no one changed his/her mind over this. So it's down to a majority preference.

**Comment by gregory38** [comment]
To clear any misurdestanding. It was a reply to
```
I guess the question is not how it suppose to be in modern c++ but maybe rather how it is implemented.

Avih's example of char* a, b; vs char *a, b; is really illustrative. Sure that's not modern c++ but nevertheless this is how it is implemented.
```
I should have quoted him (I'm a slow typer)

Otherwise, for the core of the debate. Yes I'm fine with the majority.

By the way, plugins/USBqemu/usb-eyetoy/usb-eyetoy.cpp
Before
```
-	unsigned char *pIn, *pOut, *pOutLine;
```
Ater
```
+    unsigned char *pIn, *pOut, *pOutLine;
```

**Comment by gregory38** [comment]
Ouch, I didn't expect this behavior
```
-   if (vendor == VEND_OMNIVISION
-    && product != PROD_OV511
-    && product != PROD_OV518
-    && product != PROD_OV511PLUS
-    && product != PROD_OV518PLUS
-    && product != PROD_OV519
-    && product != PROD_OV1519
-    && product != PROD_OV2519
-    && product != PROD_OV3519
-    && product != PROD_OV4519
-    && product != PROD_OV5519
-    && product != PROD_OV6519
-    && product != PROD_OV7519
-    && product != PROD_OV8519
-    && product != PROD_OV9519
-    && product != PROD_OVA519
-    && product != PROD_OVB519
-    && product != PROD_OVC519
-    && product != PROD_OVD519
-    && product != PROD_OVE519
-    && product != PROD_OVF519
-    && product != PROD_OV530)
+    if (vendor == VEND_OMNIVISION && product != PROD_OV511 && product != PROD_OV518 && product != PROD_OV511PLUS && product != PROD_OV518PLUS && product != PROD_OV519 && product != PROD_OV1519 && product != PROD_OV2519 && product != PROD_OV3519 && product != PROD_OV4519 && product != PROD_OV5519 && product != PROD_OV6519 && product != PROD_OV7519 && product != PROD_OV8519 && product != PROD_OV9519 && product != PROD_OVA519 && product != PROD_OVB519 && product != PROD_OVC519 && product != PROD_OVD519 && product != PROD_OVE519 && product != PROD_OVF519 && product != PROD_OV530)
         return NULL;
```

**Comment by willkuer** [comment]
I guess a switch wouldn't hurt...

**Comment by sudonim1** [comment]
* is part of the declarator no matter how "modern" the C++ is.  You're just trying to pretend C++ is a different language than it actually is.

**Comment by sudonim1** [comment]
```
+    unsigned char *pIn, *pOut, *pOutLine;
```

This just shows that this clang format rule is not only nonsensical, it's inconsistent.

**Comment by gregory38** [comment]
My remarks on modern C++11 was "you shall not use raw pointer in C++"

How do you write reference ? `const int& f = ... ` or `const int &f = ` Or do set the return type of a function `void *get_pointer()` or `void* get_pointer()`

**Comment by gregory38** [comment]
```
I guess a switch wouldn't hurt...
```
Maybe. The issue is that I didn't expect the tool to merge line. It is annoying because we will need to add an automatic line breaker to avoid those kind of horrors. And then a lots of tunning will be required. Bigger diff.

**Comment by sudonim1** [comment]
References are the same of course, part of the declarator.  Functions are a funny case, but (this will and should horrify you) this is legal C and C++:

int fortytwo = 42, main(int argc, const char *argv[]);

...so yes, void *get_pointer()

**Comment by sudonim1** [comment]
Or maybe more illustrative:

```
const char slash='/', hi[] = "Hello world", *get_program_name();
```

**Comment by sudonim1** [comment]
And to be clear I also think that this is a terrible aspect of the language that shouldn't've been designed this way, but here we are.  Misleading whitespace won't fix C++.

**Comment by gregory38** [comment]
The fact is `void*` is more used than `void *` on the code. And you could check any type. Reference is the same  `foo&`  rather than `foo &`.
```
 gr "void\* " pcsx2 common | wc -l
249
gr "void \*" pcsx2 common | wc -l 
79
```

Honestly I will survive with any type. I'm working on the HW industry, whitespace are generated based on the output of /dev/random (and I'm not kidding, unfortunately oh god!). Just saying that neither of the choice are so obvious, neither a standard.

**Comment by gregory38** [comment]
@turtleli 
I'm trying to plug my script into travis. But I got a strange git error. Do you know the shell and the git version on travis ? 
```
+git diff master...HEAD --name-only --diff-filter=ACMRT

++grep -v plugins/GSdx_legacy/

grep+ -v plugins/LilyPad/

+grep -v plugins/GSdx/

grep -v+ plugins/onepad/

++grep -v plugins/xpad/

+grep -v+grep -v plugins/zerogs/

 plugins/PeopsSPU2/+grep -v plugins/zerospu2/

grep

+ -v plugins/zzogl-pg/

+grep -v /resource.h

grep+ -v 3rdparty/

++++grepgrepgrepgrepgrep -v -v -v bin/ -v unfree/ tools/

 tests/ -v

+grep cmake/+

 -v plugins/spu2-x/

grep -v+grep plugins/dev9ghzdrk/

fatal: ambiguous argument 'master...HEAD': unknown revision or path not in the working tree.

Use '--' to separate paths from revisions, like this:

'git <command> [<revision>...] -- [<file>...]'
```

**Comment by turtleli** [comment]
git is 1.9.1 on Travis CI. I don't know what shell is being used, though I'd assume bash.

**Comment by sudonim1** [comment]
What's more common currently really doesn't matter if we're running a huge reformatting pass over the repository.  Only one actually reflects the structure of the language, another is pretending to be C#.

**Comment by gregory38** [comment]
Hum travis only populate a single branch so I can't do a diff with master. I don't know if I can get a smart commit of the branch.

@sudonim1 fair point. But you can't say one is real C++. Especially when the one that wrote C++, don't agree with you ;)

**Comment by sudonim1** [comment]
Stroustrup says a lot of stupid things, as evidenced by C++.

**Comment by gregory38** [comment]
Maybe but I still think it is  a matter of opinion ;)  It would be best if everyone can share his point of view. Then I will do the update accordingly.

Here a couple of extract. Left vs right. As you can see Clang move the `&` too. I didn't checked function pointer.
```
-void JoystickInfo::EnumerateJoysticks(vector<GamePad*>& vjoysticks)
+void JoystickInfo::EnumerateJoysticks(vector<GamePad *> &vjoysticks)

-    vector<GamePad*>::iterator itjoy = s_vgamePad.begin();
+    vector<GamePad *>::iterator itjoy = s_vgamePad.begin();

-    void OnButtonClicked(wxCommandEvent&);
+    void OnButtonClicked(wxCommandEvent &);

-    wxButton* bt_tmp = (wxButton*)event.GetEventObject();  // get the button object
+    wxButton *bt_tmp = (wxButton *)event.GetEventObject();  // get the button object
```

**Comment by gregory38** [comment]
So status on left/right pointer troll

turtleli,me => nicer on the left for c++
ssakash => prefer on the left

avih, aktau => right (big project & variable name)
sudo => right (language construct)

ref => middle camp

**Comment by mirh** [comment]
Not that I'm really all that into programming, but the other day I was thinking exactly on this highlighting type rather the variable. 
You'd do the second to highlight that's the only thing for which the pointer is valid.. 
But why in the hell don't you just write

    char* a;
    char b; 

??

**Comment by gregory38** [comment]
Sometime people want to write
```
char *a, *b; // you need to put 2 *
char* a, * b; // could work but feel strange
char* a; // of course 2 lines is good for everyone
char* b;
```

**Comment by avih** [comment]
When there's no variable (like in a cast or template type), IMO it's fine to put the `*` near the type ("left"). I only prefer it adjacent to the variable when there actually is a variable.

**Comment by gregory38** [comment]
Well giga does the same. Left everywhere, except for variable. However I can't do that with clang format. And it isn't consistent. I guess I will do it on the right.

**Comment by avih** [comment]
Can't you just not enforce it and we'll just know when to put it right and when left?

**Comment by gregory38** [comment]
Dunno. Honestly I would prefer having right everywhere than a random mess as currently.

**Comment by avih** [comment]
There's quite a distance between agreeing how it should look, or even allowing some slack, to "random mess". We're humans, each of us have slightly different preferences sometimes, we might want to align some stuff between lines to make it look cleaner, etc.

We don't have to go from one extreme side (the mess it was) to the other extreme (100% controlled by fixed style using some external formatter).

**Comment by sudonim1** [comment]
It's very silly that it's still adding the space when there's no declarator, that seems like something that needs to be fixed in clang-format itself.

**Comment by sudonim1** [comment]
Nobody's seriously suggesting that mixed type declarations are a good idea, @mirh.  It's just to demonstrate why grouping it with the type specifier is actually a lie.

**Comment by gregory38** [comment]
I'm not sure the tool allow relax rule.
```
We're humans, each of us have slightly different preferences sometimes,
```
It is the summary of the issue. Boys will be boys. Code will diverge and it will become mess again. An external format tool isn't perfect but so far it is the best solution. It comes with some constraints but it worth it (IMHO).

**Comment by willkuer** [comment]
Have you considered `DerivePointerAlignment: true`? This would mean you can adjust pointer in the files you touch as you like and other can do it differently. It's consistent by file.

Sure that still doesn't solve the problem with the space during casts...
But actually I don't get the point of sudo and avih. Why don't you like the space? To me it looks rather correct and for sure is consistent... If you'd like to stick it to the type I would also do it in the declaration of variables...

**Comment by sudonim1** [comment]
The examples here are different grammar (abstract-declarator instead of declarator), and ideally we would not have the space in these, but clang-format doesn't currently support this.  It does seem that they are aware of the issue, there's a fixme comment in their repository.

```
-void JoystickInfo::EnumerateJoysticks(vector<GamePad*>& vjoysticks)
+void JoystickInfo::EnumerateJoysticks(vector<GamePad *> &vjoysticks)

-    vector<GamePad*>::iterator itjoy = s_vgamePad.begin();
+    vector<GamePad *>::iterator itjoy = s_vgamePad.begin();

-    void OnButtonClicked(wxCommandEvent&);
+    void OnButtonClicked(wxCommandEvent &);

-    wxButton* bt_tmp = (wxButton*)event.GetEventObject();  // get the button object
+    wxButton *bt_tmp = (wxButton *)event.GetEventObject();  // get the button object
```

**Comment by gregory38** [comment]
A per file solution will be awful. Besides current style is mostly on the left. We can still use dedicated format by plugins.

Which fixme. For me right syntax is consistent. What you're trying to say is that you only want the right position for variable declaration? 
```
wxButton *bt_tmp = ...; // note could use: auto bt_tmp = ....
```

**Comment by sudonim1** [comment]
```
  // FIXME: Currently we don't handle properly the PointerAlignment: Right
  // The * and & are not aligned and are left dangling. Something has to be done
  // about it, but it raises the question of alignment of code like:
  //   const char* const* v1;
  //   float const* v2;
  //   SomeVeryLongType const& v3;
```

The comment's not really explicit about what they consider handling it "properly", but at least they're aware that it's not ideal.

sudonim1:
And no, anything with an actual declarator should have the right alignment.  An abstract-declarator is a declarator with no identifier.

gregory38:
Example would be nice for layman guy ;) So what are those declarators stuff?

sudonim1:
abstract-declarators can be used where a type is required but not an identifier, e.g. in function prototypes, casts and template arguments.  The full grammar is pretty complex, but the only actual difference is the lack of an identifier.

gregory38:
Well function prototypes are likely better with a name. However I don't understand the logic. I mean you want to put a space after the type when you declare a variable because `*&` aren't related to the type but related to the identifier. If there is no identifier, you want no space but `*&` still don't belong to the type.

sudonim1:
Well where an abstract-declarator is called for in the grammar it is actually part of a string of tokens that specify a type and nothing else.

gregory38:
IMHO, I don't think that following the grammar `stricto sensu` is a good goal. Goal is to find something that people will write easily and read as easily.

So in the end, it could be a better idea to use the left setting
* Multiple pointer declarations will remain on right. (It impacts few line of code, anyway, and likely less in the future)
* If someone adds a new option `only right for variable declaration`, we could still change it with less impact.

gregory38:
I created the PR #1551 with the right declaration. I converted the pad plugin too. If you don't like the right syntax anymore. Please tell me, we can still go back to the left one.

jobs-git:
If you need some man power to update syntax and formatting manually on some files, just ping me.

PR #1531, opened 2016-08-21
pcsx2|common|gsdx: use empty() instead of .size() ==/!= 0 check

gregory38:
Enhance readability reported by clang tidy

### DIFF
--- a/common/src/Utilities/wxAppWithHelpers.cpp
+++ b/common/src/Utilities/wxAppWithHelpers.cpp
@@ -482,7 +482,7 @@ void wxAppWithHelpers::OnSynchronousCommand( pxSynchronousCommandEvent& evt )
 void wxAppWithHelpers::AddIdleEvent( const wxEvent& evt )
 {
 	ScopedLock lock( m_IdleEventMutex );
-	if( m_IdleEventQueue.size() == 0 )
+	if( m_IdleEventQueue.empty() )
 		PostEvent(wxCommandEvent(pxEvt_StartIdleEventTimer));
 
 	m_IdleEventQueue.push_back( evt.Clone() );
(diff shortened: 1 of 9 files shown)


danilaml:
Dunno about readability. `.empty()` is a bad name since it's a verb. One might think it empties the container. The should've gone with `is_empty` or something.

ssakash:
``empty()`` seems a bit confusing for the reasons stated by danilami. Personally ``size()`` seems to make more sense to me.

Is there any benefit on using ``empty()`` instead of ``size()`` ?

gregory38:
Well inside a if context, I won't think about something that empty the container. By definition if check a state.
Besides, it is a rather standard name for the c++ library.

turtleli:
I like `empty()` over `size()` ;)

gregory38:
Some people know better who is the boss :stuck_out_tongue: 

Seriously, http://llvm.org/releases/3.8.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability-container-size-empty.html and it feels easier to read, less symbol to decipher for limited brain like mine.

willkuer:
If there is a function in the interface to be used for exactly one purpose and you don't use it when facing exactly this purpose... Well that's not the smartest idea.

Although it's interesting that they say that the size might not be retrievable in constant time for all containers. Would be interesting for which container the size is not counted upon adding and removing elements.

danilaml:
@willkuer IIRC `std::list` could have linear complexity for `size()` in some implementations until c++11 and forward_list doesn't have size method at all.

ssakash:
> Some people know better who is the boss 😛

Hmm. Thinking again,  ``empty()`` does sound better! 

Though honesty,  ``empty()`` is dedicated for these scenarios as willkuer pointed out and it returns a bool value which is more suitable for conditional checks.

So yeah,  I love ``empty()`` over ``size()``!

gregory38:
Yes `std::list.size()` which is still linear on GCC 4.9. One GSdx function could be made 5 times faster with a proper size implementation.

PR #1532, opened 2016-08-23
PCSX2: Fix a bunch of coverity defects

ssakash:
**Summary of changes**:

* EE-Counters : the defect coverity found was not actually an issue but apparently someone marked it as ``fix required`` so FWIW I made the code behavior more explicit by moving the unknown video mode handling to default case.

* LilyPad : Remove useless self assignment.

* PCSX2-Core : Initialize class members on GIF/VIF/SuperVU

Initially I was planning on only fixing the defect which I had introduced (counters code) though that would make the PR too small and pointless so made a bunch of random defect fixes along with it.

### DIFF
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -335,14 +335,13 @@ u32 UpdateVSyncRate()
 		scanlines = SCANLINES_TOTAL_NTSC;
 		break;
 
-	// Falls through to unknown when unidentified mode parameter of SetGsCrt is detected.
 	case GS_VideoMode::Unknown:
+	default:
+		// Falls through to default when unidentified mode parameter of SetGsCrt is detected.
 		// For Release builds, keep using the NTSC timing values when unknown video mode is detected.
 		// Assert will be triggered for debug/dev builds.
 		scanlines = SCANLINES_TOTAL_NTSC;
 		Console.Error("PCSX2-Counters: Unknown video mode detected");
-
-	default:
 		pxAssertDev(false , "Unknown video mode detected via SetGsCrt");
 	}
 
--- a/plugins/LilyPad/InputManager.cpp
+++ b/plugins/LilyPad/InputManager.cpp
@@ -433,7 +433,6 @@ Device *InputDeviceManager::GetActiveDevice(InitInfo *info, unsigned int *uid, i
 							  (devices[i]->oldVirtualControlState[j] > 31*FULLY_DOWN/32 && devices[i]->virtualControlState[j] < 7*FULLY_DOWN/8))) {
 									continue;
 						}
-						devices[i]->virtualControls[j].uid = devices[i]->virtualControls[j].uid;
 					}
 					else if ((((devices[i]->virtualControls[j].uid>>16)&0xFF) == ABSAXIS)) {
 						if (devices[i]->oldVirtualControlState[j] > 15*FULLY_DOWN/16)
--- a/pcsx2/x86/sVU_zerorec.cpp
+++ b/pcsx2/x86/sVU_zerorec.cpp
@@ -304,13 +304,17 @@ VuBaseBlock::VuBaseBlock()
 	type = 0;
 	endpc = 0;
 	cycles = 0;
-	pcode = NULL;
 	id = 0;
 	memzero(pChildJumps);
 	memzero(startregs);
 	memzero(endregs);
 	allocX86Regs = nStartx86 = nEndx86 = -1;
 	prevFlagsOutOfBlock = 0;
+	startpc = 0;
+	vuxy = 0;
+	vuxyz = 0;
+	pendcode = nullptr;
+	pcode = nullptr;
 }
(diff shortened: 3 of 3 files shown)


ssakash:
I hope Gregory is fine with the new coverity label :P

gregory38:
I'm not happy with the tag. There are gcc warning/clang tidy/cppcheck and countless of formal/qa tools.

You can use
* Q&A tools
* Formal analysis
* Quality assurance
Search google for better naming. But get a generic word

ssakash:
@gregory38 

Feel free to edit the [label](https://github.com/PCSX2/pcsx2/labels) to a more generic word. (I suck at picking words :P)

pcsx2/Counters.cpp
```diff
@@ -335,14 +335,12 @@ u32 UpdateVSyncRate()
 		scanlines = SCANLINES_TOTAL_NTSC;
 		break;
 
-	// Falls through to unknown when unidentified mode parameter of SetGsCrt is detected.
-	case GS_VideoMode::Unknown:
+	default:
```
gregory38: I would have done
```
case GS_VideoMode::Unknown:
default:
....
```
This way we clearly understand the intent of your code.

plugins/LilyPad/InputManager.cpp
```diff
@@ -433,7 +433,6 @@ Device *InputDeviceManager::GetActiveDevice(InitInfo *info, unsigned int *uid, i
 							  (devices[i]->oldVirtualControlState[j] > 31*FULLY_DOWN/32 && devices[i]->virtualControlState[j] < 7*FULLY_DOWN/8))) {
 									continue;
 						}
-						devices[i]->virtualControls[j].uid = devices[i]->virtualControls[j].uid;
```
gregory38: Are you the line is useless. It could be mistype of a wrong variable. At work I got a similar bug, I use the wrong variable on the right. Dunno if there any info on git history

pcsx2/x86/newVif_Unpack.cpp
```diff
@@ -73,8 +73,21 @@ static const __aligned16 Fnptr_VifUnpackLoop UnpackLoopTable[2][2][2] = {
...
-	numBlocks	=  0;
+	vifBlocks = NULL;
+	block.startPtr = NULL;
+	recReserve = NULL;
+	recWritePtr = NULL;
+	idx = 0;
+	bSize = 0;
+	bPtr = 0;
+	numBlocks = 0;
+	block.num = 0;
```
gregory38: I think there are better alternative to init blocks struct
* could be memset
* could be initializer list: `= {}`
* Others?

pcsx2/Gif.h
```diff
@@ -26,7 +26,8 @@
 #define GUNIT_WARN(...) do {} while(0)
 
 enum GIF_PATH {
-	GIF_PATH_1 = 0,
+	GIF_PATH_Unknown = -1,
```
gregory38: I would prefer to avoid to add a value to this enum. Because the value isn't checked.

pcsx2/Gif_Unit.h
```diff
@@ -166,8 +166,18 @@ struct Gif_Path {
...
-	Gif_Path()  { Reset(); }
-	~Gif_Path() { _aligned_free(buffer); }
+	Gif_Path() {
+		buffer = nullptr;
+		buffSize = 0;
+		buffLimit = 0;
+		dmaRewind = 0;
+		idx = GIF_PATH_Unknown;
+		Reset();
+	}
```
gregory38: Do you really need all those init. As far as I know, the object is only used in Gif_Unit object which call the `Init` function in the constuctor. Maybe it could be better to merge the `constructor` and the `Init` function.

gregory38:
Ping @ssakash

ssakash:
Too busy with semester exam preparations right now. I could probably only update this PR after September 4. Sorry!

gregory38:
I'm fine with it. Good luck with your exam :+1:

ssakash:
> I'm fine with it. Good luck with your exam :+1:

Thanks. I did it sort of decent I guess :P 

I have removed those GIF/VIF changes from the PR (free time is too limited to handle them right now). Updated the other stuffs as per your suggestion :)

plugins/LilyPad/InputManager.cpp
ssakash: I tried looking into the code a bit but couldn't find any other scenarios where other variables would be useful. pinging @turtleli for feedback in case he has any ideas.

turtleli: Nope, no ideas here.

I assume it's code left over from someone rearranging the code, considering the last two parameters of GetActiveDevice aren't used for anything at all (it's only called in one place, and that doesn't use the variables).

gregory38: I gave a look too. I'm fine with the removal.

PR #1535, opened 2016-08-24
clang-tidy: use parenthesis around macro parameters

gregory38:
Could avoid misused in the future

Clang Tidy warning misc-macro-parentheses 157 => 66

### DIFF
--- a/pcsx2/FPU.cpp
+++ b/pcsx2/FPU.cpp
@@ -74,23 +74,28 @@
 //****************************************************************
 
 // If we have an infinity value, then Overflow has occured.
-#define checkOverflow(xReg, cFlagsToSet, shouldReturn) {  \
-	if ( ( xReg & ~0x80000000 ) == PosInfinity ) {  \
-		/*Console.Warning( "FPU OVERFLOW!: Changing to +/-Fmax!!!!!!!!!!!!\n" );*/  \
-		xReg = ( xReg & 0x80000000 ) | posFmax;  \
-		_ContVal_ |= cFlagsToSet;  \
-		if ( shouldReturn ) { return; }  \
-	}  \
+bool checkOverflow(u32& xReg, u32 cFlagsToSet)
+{
+	if ( (xReg & ~0x80000000) == PosInfinity ) {
+		/*Console.Warning( "FPU OVERFLOW!: Changing to +/-Fmax!!!!!!!!!!!!\n" );*/
+		xReg = (xReg & 0x80000000) | posFmax;
+		_ContVal_ |= (cFlagsToSet);
+		return true;
+	}
+
+	return false;
 }
 
 // If we have a denormal value, then Underflow has occured.
-#define checkUnderflow(xReg, cFlagsToSet, shouldReturn) {  \
-	if ( ( ( xReg & 0x7F800000 ) == 0 ) && ( ( xReg & 0x007FFFFF ) != 0 ) ) {  \
-		/*Console.Warning( "FPU UNDERFLOW!: Changing to +/-0!!!!!!!!!!!!\n" );*/  \
-		xReg &= 0x80000000;  \
-		_ContVal_ |= cFlagsToSet;  \
-		if ( shouldReturn ) { return; }  \
-	}  \
+bool checkUnderflow(u32& xReg, u32 cFlagsToSet) {
+	if ( ( (xReg & 0x7F800000) == 0 ) && ( (xReg & 0x007FFFFF) != 0 ) ) {
+		/*Console.Warning( "FPU UNDERFLOW!: Changing to +/-0!!!!!!!!!!!!\n" );*/
+		xReg &= 0x80000000;
+		_ContVal_ |= (cFlagsToSet);
+		return true;
+	}
+
+	return false;
 }
 
 /*	Checks if Divide by Zero will occur. (z/y = x)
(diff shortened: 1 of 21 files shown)


refractionpcsx2:
That looks horrible.

gregory38:
What is horrible? Macro are horrible ;) It is safer with parenthesis.

refractionpcsx2:
the parenthesis around them all looks horrible, it would be better to prefix the macro with something like m_ or equivalent than sticking brackets around everything, it's just messy.

ssakash:
I agree with ref. Best to avoid extra parenthesis whenever they're not needed.  ( the code looks messy with all those parenthesis )

sudonim1:
Uh, prefixes? I guess you missed the point.
It's to avoided unexpected precedence when you use an expression as a macro argument.  Macros just do token substitution.

sudonim1:
To emphasise, this is not in any way a style decision.  It enforces correct logic.

sudonim1:
However the ones on macros which just expand to constant expressions confuse me.

sudonim1:
And the brackets on macro invocations are weird too.  This should all be taken care of in the macro

ssakash:
> However the ones on macros which just expand to constant expressions confuse me.

Yes.  That's why I said 

```
Best to avoid extra parenthesis whenever they're not needed
```

I feel some redundant paranthesis' are also added along with the needed ones which enforce precedence (looking at the code)

gregory38:
Which parenthesis are redundant? Most of them are mandatory to guarantee the behavior of the macro. So far it is fine but someday, someone will call them and will get a hard to debug issue.

Otherwise, a better alternative will be to replace some macros with some templates whenever possible.

pcsx2/IPU/mpeg2lib/Idct.cpp
```diff
@@ -38,8 +38,8 @@
 #define W5 1609 /* 2048*sqrt (2)*cos (5*pi/16) */
 #define W6 1108 /* 2048*sqrt (2)*cos (6*pi/16) */
 #define W7 565  /* 2048*sqrt (2)*cos (7*pi/16) */
-#define clp(val,res)	res = (val < 0) ? 0 : ((val > 255) ? 255 : val);
-#define clp2(val,res)	res = (val < -255) ? -255 : ((val > 255) ? 255 : val);
+#define clp(val,res)	res = ((val) < 0) ? 0 : (((val) > 255) ? 255 : (val));
+#define clp2(val,res)	res = ((val) < -255) ? -255 : (((val) > 255) ? 255 : (val));
```
turtleli: These 2 macros are unused.

sudonim1:
'''
+#define IPU_Y_COEFF   (0x95)	//  1.1640625
'''

This for instance.  I can't think of any reason that would be necessary

gregory38:
I think it was reported by the tool. I need to double check. Tool likely doesn't make any difference between a number or a complex macro (that need extra parenthesis).

gregory38:
Actually the tool report issue only with negative number.

gregory38:
I updated the branch to use nice function instead of macro (for easy case). No parenthesis clutter and type safe.

gregory38:
@refractionpcsx2 it the updated version better for you ?

refractionpcsx2:
It doesn't look any different to me, still far too many parenthesis for my liking

sudonim1:
@refractionpcsx2 you do understand that they are actually necessary, don't you? This isn't a style issue.  For example HWADDR(x & 0xFFFFFFF0U ) would expand to (hwLUT[x & 0xFFFFFFF0U >> 16] + (x & ~15)) which is (hwLUT[x & 0xFFF0U] + (x & ~15)).  Not what you were expecting, right?

refractionpcsx2:
Sudo, I know THOSE  are necessary, butiI mean like the (xReg) =  whatever  etc,   pointless and ugly parentheses

gregory38:
Hum, I would replace FPU stuff by real function it would be nicer.

willkuer:
Wouldn't it be straightforward to replace these macro's by inline functions?
What is stopping you from replacing all macros by functions?

gregory38:
I did replace some macro by some functions. It was faster to add parenthesis. And some macros aren't straightforward. For example of the FPU file, you have some return on the macro. So you need also to replace macro call everywhere. But yeah, I will replace more macro with functions.

willkuer:
Ah I missed the return...  then it makes sense.

sudonim1:
@refractionpcsx2 , "ridiculous amount of parentheses" isn't an argument, it is entirely possible to write code that will silently and mysteriously break with almost every one of these macros, and in many cases the code wouldn't even be that unusual.  Macros do dumb token replacement, so you turn every argument into a primary-expression this way, it's the only safe thing to do.

sudonim1:
The answer to macros being ugly or convoluted is use less and simpler macros, not use unsafe macros.

avih:
Macros should be safe.

Arguably, macros which are very local where the usage is only in close proximity to the definition may use less safe form as long as it's highly unlikely to use them incorrectly (with complex arguments). Though that's arguable. If in doubt - safety is preferred.

Also, a macro is defined once, so as long as its function and usage are clear (use comments where necessary), the clarity of the macro itself is less important, and using seemingly extra parenthesis with macros is not unusual - as indeed it's the safe way to write macros.

As for replacing macros with inline/static/template functions, I think it's OK, but whether or not `inline` is useful should be proven first. In general I think today's compilers know to inline very small functions on their own without the devs telling them to.

IIRC rama and myself tried it once where we replaced the inline macro with nothing (i.e. just don't inline anything), and there wasn't any measurable speed difference in our quick tests. Though the executable became quite a bit smaller (IIRC ~30% smaller).

gregory38:
I replaced a couple of more macro with functions and removed one unused macro.

Remaining macro will be harder (impossible) to convert to function. The extra parenthis is rather limited now.
* potentially ALLOC stuff could be replaced by function but must add info parameter everywhere. And code would be redone for a 64 bits port.
* _LIMX  would love a template but again need to update all callers.

Safety is always better IMHO (when it doesn't impact perf). Besides it make the code more `"pro"`.

gregory38:
Could I merge it ?

pcsx2/FPU.cpp
```diff
@@ -180,14 +188,14 @@ void ABS_S() {
 
 void ADD_S() {
 	_FdValf_  = fpuDouble( _FsValUl_ ) + fpuDouble( _FtValUl_ );
-	checkOverflow( _FdValUl_, FPUflagO | FPUflagSO, 1 );
-	checkUnderflow( _FdValUl_, FPUflagU | FPUflagSU, 1 );
+	if (checkOverflow( _FdValUl_, FPUflagO | FPUflagSO)) return;
+	if (checkUnderflow( _FdValUl_, FPUflagU | FPUflagSU)) return;
```
turtleli: `if (something) return;` is kind of unnecessary at the end of a function.

gregory38: Yes, you're right. I did a search and replace. I like the symetry but I would remove them if you prefer.

turtleli: I'd prefer if they were removed.

gregory38: Done

gregory38:
@turtleli are you fine with it ?

turtleli:
Yeah, it looks fine.

gregory38:
Good. Let's merge it then.

Issue #1499, opened 2016-08-02
GSdx: 2 passes alpha test isn't accurate (wrong order)


gregory38:
Game: Kengo
Letters rendering is awful, seem to miss pixels (first screen if you have no memcard save)

Game will render 2 sprites by letters with Alpha Test enabled (RGBA only, i.e. depth will be discarded if alpha failed the test). Both blending and depth test are enabled too. (note: on this particular case, depth test is useless)

Current algorithm does:
1. render all pixels that pass alpha test (color + depth)
2. render all pixels that fail alpha test (only color)

The 2 passes doesn't conserve the order of primitive for blending operation. First primitive could be discarded because it failed the alpha test. And it will only be executed in FS in the 2nd phase of the algorithm.

The game uses a constant MAX depth for primitives in GREATER_OR_EQUAL mode. So in this case, we don't need to guarantee the order of the depth buffer (equivalent to ALWAYS). And it would be a solution to compute first color (of all fragments), and then the depth (of fragment that pass the alpha test). However it would be nice to find a generic solution, if possible, as HW is quite limited here.

FlatOutPS2:
It's the same issue as Driv3r, Ridge Racer V, and Burnout games text.

gregory38:
Look similar or proven to be the same? It might worth to check the depth value of the primitive. In kengo depth test will never failed (due to the usage of max value). If this property holds, I can fix it.

FlatOutPS2:
Not sure if it's exactly the same, but they're text issues caused by an inaccurate alpha test.

I'm not 100% about Driv3r, but when I recently hacked a fix for the Burnout games, it improved Driv3r as well.

gregory38:
could you give me a gs dump of burnout letter so I can quickly compare the alpha testing.

FlatOutPS2:
There's one in the issue about it: https://github.com/PCSX2/pcsx2/issues/935

gregory38:
Thanks you. Yes it is the same issue. And they also use 0xFF_FFFF (aka the max) as depth value. It is strange because depth buffer already contains the max value. So as Kengo, Alpha test is useless. Or maybe I failed to see the real effect due to depth rounding. (or likely game designer were high again...)

FlatOutPS2:
Did you look at Ridge Racer V as well? It's not exactly the same as Kengo and Burnout, but it might be insightful.

DPersonalized:
[...]

PR #1538, opened 2016-08-27
gsdx ogl: extend ATE to better support additional corner case (issue …

gregory38:
…#1499)

When depth primitive is constant and depth test is greater or equal, we can
execute the depth write after color (depth status will only depends on the initial
value)

New case for RGB_ONLY ate:
If the blending equation uses a fixed alpha or a source alpha. We can postpone the alpha write
in a 2nd pass.
If depth can also be postponed, we can guarantee the order of correctness of the value.
1st pass => do RGB
2nd pass => do Alpha & Depth

It fixed Stuntman letter rendering :) Remaining of the game is still broken :(


FlatOutPS2:
> It fixed Stuntman letter rendering :) Remaining of the game is still broken :(

Well, at least it fixes the most important part of the game. :)

Can't you add the DX version as well? It only requires minimal changes.

gregory38:
```
Can't you add the DX version as well? It only requires minimal changes.
```
I can't even compile the file. I'm sure you can do it as fast as me ;)

Issue #551, opened 2015-05-22
[Loss layer] Lets Make a J-League Pro Soccer Club [NTSC-J] 


kennybaby:
# PCSX2 version: Tested with both latest SVN 1.3.1.xxxx, and with PCSX2 1.2.1
# Emulation options: All default
# Plugins Used: GSdx SSE2、3、4. Both software and hardware mode on direct3d9、direct3d10、direct3d11、OPGL.
# Most polygon‘s player eye is white，just like missing layer。The game interface also missing some menu，I think it’s the same trouble。

In the game some PIC is missing，display the white……

I know the game is unwelcome in Europe and America ,people do not like to play，But in Asia It is a classic.
I do not know any English or just a little bit，I use google translate read and type,This cost my whole afternoon,Hope you can read this and my [Bug Report] is correct format.(copy from other thread)……
So PLZ……
How it looks:
http://forums.pcsx2.net/Thread-Loss-layer-Lets-Make-a-J-League-Pro-Soccer-Club-NTSC-J
![qq 20140208164328](https://cloud.githubusercontent.com/assets/12553071/7768486/01dc1536-00ae-11e5-9cad-6748d2d76c78.jpg)
![qq 20140208163625](https://cloud.githubusercontent.com/assets/12553071/7768488/07f63a14-00ae-11e5-9cc8-1a07298d13df.jpg)

kksxf:
Just hope somebody could see this issue,i really like this game.

ssakash:
does ZeroGS have this issue ?

kksxf:
yes

ssakash:
then it's more likely to be a core issue, I don't think it would be dealt with fast then. could you check out some older version of PCSX2 and check whether this bug existed ? like 0.9.6, 0.9.8 and other older versions.

kksxf:
in version 0.9.6,the game cannot start normally.
in version 0.9.8,the same issue still exists.

ssakash:
could you try different rounding/ clamping modes and Interpreters and SuperVU recompiler and check out whether it works on the latest development build ?

kksxf:
while vu0: Interpreters or superVU recompiler
         vu1:microVU recompiler
rounding/ clamping modes:negative direction

it does make some changes,but the picture is still UNnormal

![image](https://cloud.githubusercontent.com/assets/13393823/8764805/f96cca98-2e21-11e5-9a10-e4c045c2a5a6.png)

[...]

Issue #451, opened 2015-02-12
Driver Parallel Lines GSdx DX11/OpenGL SW Cars randomly changing color - HW garbage textures


joder666:
I am using **pcsx2-v1.3.1-8-gf88bea5-windows-x86** from orphis buildbot.
Default settings (2-safe) except for:
GSdx10 software +2 rendering threads
SPU-X set to async mix and interpolation to cubic.

To reproduce just start a new game. While driving around you'll notice how your vehicle changes color randomly (red,white,black,blue) as well as other cars at sight, except for Taxis and Cops.

There are not known revisions of pcsx2 that previously worked OK with this game iirc. Last time i tested it was with svn r5559, unplayable using GSdx hardware mode due to graphics errors, while too slow with software mode.

BTW, i am using the igpu from processor HD Graphics 2000 with latest drivers installed 9.17.10.4101.  

**Log:**

PCSX2 1.3.1-20150201194910 - compiled on Feb  1 2015
Savestate version: 0x9a0a0000

Host Machine Init:
    Operating System =  Microsoft Windows 7 Enterprise Edition Service Pack 1 (build 
7601), 64-bit
    Physical RAM     =  7882 MB
    CPU name         =  Intel(R) Core(TM) i5-2500 CPU @ 3.30GHz
    Vendor/Model     =  GenuineIntel (stepping 07)
    CPU speed        =  3.299 ghz (4 logical threads)
    x86PType         =  Standard OEM
    x86Flags         =  bfebfbff 1fbae3ff
    x86EFlags        =  28100000

x86 Features Detected:
    SSE2.. SSE3.. SSSE3.. SSE4.1.. SSE4.2.. AVX

Reserving memory for recompilers...

Loading plugins...
    Binding   GS: D:\BC\Juegos\Emus\pcsx2-v1.3.1-8-gf88bea5-windows-x86\plugins\GSdx32-
AVX.dll 
Windows 6.1.7601 (Service Pack 1 1.0)
    Binding  PAD: D:\BC\Juegos\Emus\pcsx2-v1.3.1-8-gf88bea5-windows-x86\plugins
\LilyPad.dll 
    Binding SPU2: D:\BC\Juegos\Emus\pcsx2-v1.3.1-8-gf88bea5-windows-x86\plugins\SPU2-
X.dll 
    Binding CDVD: D:\BC\Juegos\Emus\pcsx2-v1.3.1-8-gf88bea5-windows-x86\plugins
[...]

Issue #287, opened 2014-09-05
GSdx software bug: Network Profile Wizard (the one built-in with some games), Most text/boxes stuck in top left + Other Glitches


TheLastRar:
-PCSX2 version: 1.3.0 GIT 05-08-2014

-PCSX2 options: Defaults, no speed-hacks or game-fixes used.

-Plugins used: GSdx, SPU2-X, Lilypad, Linuz ISO/Gigaherz CDVD,rest null plugins

-Plugin settings: Tested GSdx DX11 software and Hardware renderer.

-Description: Text/pop-up boxes (i.e.do you want to quit) appears in the top left, making the wizard virtually unusable (see images below)

-How to replicate: The game these images are taken from is Ratchet Gladiator, however, Other games network wizards seem to also share the issue.

-Last known version to work: Unknown, probably never worked.
-PC specifications: CPU: Intel Core i5 2430M @ 2.4Ghz GPU: Geforce GT 520M OS: Windows 7

I think this issue is already known, but I didn't see a issue for it created in Github

image you are greeted with when you start the wizard
![Image of main screen](http://i.imgur.com/PnzDC4x.png)

image for entering a manual IP
![Image of IP screen](http://i.imgur.com/Qdxf3Ys.png)

Image you get when you give up
![Image of Cancel screen](http://i.imgur.com/ABPc67b.png)

Edit: Log (note that the start of this includes the Ratchet Gladiators, which contained the Net Wizard I got the images from)

```
PCSX2 1.3.0-20140902212146  - compiled on Sep  2 2014
Savestate version: 0x9a0a0000

Host Machine Init:
    Operating System =  Microsoft Windows 7 Home Premium Edition Service Pack 1 (build 7601), 64-bit
    Physical RAM     =  5995 MB
    CPU name         =  Intel(R) Core(TM) i5-2430M CPU @ 2.40GHz
    Vendor/Model     =  GenuineIntel (stepping 07)
    CPU speed        =  2.393 ghz (4 logical threads)
    x86PType         =  Standard OEM
    x86Flags         =  bfebfbff 1fbae3bf
    x86EFlags        =  28100000

x86 Features Detected:
    SSE2.. SSE3.. SSSE3.. SSE4.1.. SSE4.2.. AVX

Reserving memory for recompilers...

Loading plugins...
    Binding   GS: C:\Games\PCSX2 1.2.1\Plugins\GSdx32-AVX.dll 
[...]

Issue #168, opened 2014-07-26
Graphics issue: Jak 3 (USA) - SCUS97330 - Corrupted textures and excessive player shadow


Byrkoet:
![1](https://cloud.githubusercontent.com/assets/1343700/3710933/ff644f0c-14b4-11e4-8e95-806dbb0e0ccb.png)
![2](https://cloud.githubusercontent.com/assets/1343700/3710935/ff64bd2a-14b4-11e4-8709-525334ca6e92.png)
![3](https://cloud.githubusercontent.com/assets/1343700/3710934/ff6484fe-14b4-11e4-9f02-6c5cde61cf41.png)

**-PCSX2 version:**
1.2.1.r5873 - compiled on Feb  3 2014

**-PCSX2 options:**
Defaults, no settings modifications affect the issue.

**-Plugins used:**
- gsdx32-sse4-r5875
- lilypad-r5875
- spu2-x-r5875
- cdvdGigaherz

**-Plugin settings:**
Tested GSdx DX9/11 hardware;
DX11 seems to make things worse by having more of an impact.
SkipDraw 1 used to remove player shadow altogether.

**-Description:**
Basically, most textures are corrupted. One great example is the right eye of most characters (also an issue in previous installments of the game).

**-How to replicate:**
Boot the game and enjoy the black-textured landscape and corrupted eyes and miscellaneous objects.

**-Last known version to work:**
Unknown, probably never worked.

**-PC specifications: CPU:**
- Intel Core i5 650 @ 4.1GHz 54 °C
- 4.00GB Dual-Channel DDR3 @ 819MHz (9-9-9-24)
- Gigabyte Technology Co., Ltd. P55-UD6 (Socket 1156) 50 °C
- 1023MB NVIDIA GeForce GTX 560 Ti (EVGA) 69 °C
- 465GB Western Digital WDC WD5000AAKX-00ERMA0 ATA Device (SATA) 45 °C

**-Other comments:**
The graphical corruption, including the missing eye(s), is also present in previous installments, including, but not limited to Jak and Daxter (SCUS97558 & SCUS97124).

gregory38:
Could you test the shadows issue on the openGL renderer?

charliebruce:
Latest version from git, using OpenGL, all default settings except 4x scaling, i7-4702HQ @ 2.2GHz / nVidia 765M / 8GB RAM.

![how the transporter should look](http://i.imgur.com/vATmGDM.png)
[...]

PR #1542, opened 2016-08-29
gsdx: save the blit buffer register when a transfer is started

### LINKED ISSUE #1319: GSDX: Motocross Mania - Bad/Missing Textures + Text
Found this one going through my games checking one of the other issues.  Looks like this has been outstanding for a long time.

Motocross Mania in GSDX, any mode Software or Hardware, OGL/DX.  Fine in ZeroGS.

Here is how it looks in GSDX
motocrossmania-textures-text

and here is how it looks in ZeroGS
mcm-zerogs

**GS Dump**: [**Here**](https://www.dropbox.com/s/mvlwkygwhld6g0p/MotoCrossMania.rar?dl=0)

gregory38:
Fix motocross mania missing texture. Close #1319

As far as I understand, transfer is initialized in DIR. But the real
write only occured later so the blit buffer could have been overwritten
by a new value.

BLIT 0 13700
TREG 40 40
DIR 0 0
BLIT 0 13f00 <=== the bad guy
Write! ...  => 0x3f00 W:1 F:C_32 (DIR 00), dPos(0 0) size(64 64)

### DIFF
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -1343,10 +1343,10 @@ void GSState::GIFRegHandlerTRXDIR(const GIFReg* RESTRICT r)
 	switch(m_env.TRXDIR.XDIR)
 	{
 	case 0: // host -> local
-		m_tr.Init(m_env.TRXPOS.DSAX, m_env.TRXPOS.DSAY);
+		m_tr.Init(m_env.TRXPOS.DSAX, m_env.TRXPOS.DSAY, m_env.BITBLTBUF);
 		break;
 	case 1: // local -> host
-		m_tr.Init(m_env.TRXPOS.SSAX, m_env.TRXPOS.SSAY);
+		m_tr.Init(m_env.TRXPOS.SSAX, m_env.TRXPOS.SSAY, m_env.BITBLTBUF);
 		break;
 	case 2: // local -> local
 		Move();
@@ -1524,7 +1524,8 @@ void GSState::Write(const uint8* mem, int len)
 	int w = m_env.TRXREG.RRW;
 	int h = m_env.TRXREG.RRH;
 
-	const GSLocalMemory::psm_t& psm = GSLocalMemory::m_psm[m_env.BITBLTBUF.DPSM];
+	GIFRegBITBLTBUF& blit = m_tr.m_blit;
+	const GSLocalMemory::psm_t& psm = GSLocalMemory::m_psm[blit.DPSM];
 
 	/*
 	 *  The game uses a resolution of 512x244. RT is located at 0x700 and depth at 0x0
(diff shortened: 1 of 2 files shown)


gregory38:
@refractionpcsx2 @FlatOutPS2 @ramapcsx2 This pull request will love some tests. If you have game that have issue with the SW renderer, it might worth a try. I hope it won't create any regression.

FlatOutPS2:
Itadaki Street Special might be a good case to test.

gregory38:
I think my dump is still broken but I'm not sure GS dump are 100% accurate to reproduce the issue.

By the way, potentially save states are broken. I dunno if there are done at vsync or potentially anywhere.

ramapcsx2:
Shadow Man 2nd Coming? :)

refractionpcsx2:
> Shadow Man 2nd Coming? :)

I doubt it, I don't think that looks right on ZeroGS either :P

gregory38:
It would be nice to have a list of games that are still broken in sw renderer. The difficulty is to be sure it isn't a core bug.

gregory38:
`I doubt it, I don't think that looks right on ZeroGS either`
Note, it could be others bug of ZeroGS too.

Here a list of SW bug from the tracker
#83 #451 #551

pepiczech77:
What about #168 ?

gregory38:
This one is a different (known) issue.

FlatOutPS2:
Driver was among the first games I tried with this PR, but it doesn't fix it. It's a completely different issue, so I didn't expect it to either.

Are you going to create a thread/issue about list of SW broken games?

gregory38:
My internet connection is dead (and my 3G smartphone quota is done too). Feel free to create a GSdx SW meta tracker with a list of issue reported for the SW renderer.

refractionpcsx2:
Driv3r has been looking fine for a while in OGL HW as far as I know? The texture flicker was surely gone a while ago, unless it's back.

FlatOutPS2:
> Driv3r has been looking fine for a while in OGL HW as far as I know? The texture flicker was surely gone a while ago, unless it's back.

I wouldn't call it fine. The car textures are still flickering and the text is still completely scrambled. And it's still very slow running in HW modes.

refractionpcsx2:
> I wouldn't call it fine. The car textures are still flickering and the text is still completely scrambled. And it's still very slow running in HW modes.

hmm im sure the text was fine when i last looked :/ I'll have to go through what builds i have to see if i can see if i was just going mad. Unless as I say it's broken again.

FlatOutPS2:
I don't recall the text ever being fixed in HW mode.

gregory38:
Fwiw, I'm currently looking at Harley Davidson. Rendering is still bad, but zzogl is fine. Maybe it is a close issue.

TheLastRar:
Since someone greg was listing SW bugs,

You seem to have missed https://github.com/PCSX2/pcsx2/issues/287

gregory38:
I saw this one but it is broken on zzogl/zerogs too, isn't it? It will be hard to debug. It is already hard enough when you have the expected output...

TheLastRar:
So you where only listing issues that where broken in GSDX but not ZeroGS?

Also what do you mean by "It is already hard enough when you have the expected output"
are you comparing GSDX output to ZeroGS output?

gregory38:
Yes I compare all states/memory of zzogl versus GSdx. Later was good for motocross mania (and for davidson). I have major log but I can find the bad draw call. Then I can dump all registers memory read/write of the draw call to find the bad state. Then I can try to understand what goes wrong in GSdx. Current bugs are really details of the spec/hardware.

gregory38:
I need to look further but I think there are some issues to handle mutiple contexts. I need to double check but I think the issue is
* write tex0_0 (load clut)
* write tex0_1 (reload clut)
* draw with contex 0 (but clut contains data of context 1)

I'm not sure how it is handled on the hardware

refractionpcsx2:
hack it until something works :P It might give you some idea of what it wants to happen

gregory38:
For the record
```
[09:07:51] <gregory38> I can confirm the issue is due to multiple context
[09:08:49] <gregory38> T:Error    ID:0    S:Mid    => Start a draw with CBP 33ed on context 0
[09:08:49] <gregory38> T:Error    ID:0    S:Mid    => TEX0 context 0
[09:08:49] <gregory38> T:Error    ID:0    S:Mid    => Apply TEX0 3671 IN ctx 0
[09:08:49] <gregory38> T:Error    ID:0    S:Mid    => TEX0 context 1
[09:08:49] <gregory38> T:Error    ID:0    S:Mid    => Apply TEX0 368a IN ctx 1
[09:08:49] <gregory38> WriteCLUT32_I8_CSM1 from f300aa00 to f777f100 (from 368a/context 1)
[09:08:49] <gregory38> T:Error    ID:0    S:Mid    => Start a draw with CBP 3671 on context 0
[09:09:15] <gregory38> So basically a write to TEX0 context 1 will trigger a load of the clut but the draw will use context 0 parameter
[09:09:18] <gregory38> so the clut is wrong
[09:09:50] <gregory38> Now  the real question
[09:09:56] <gregory38> is how hardware handle it
[09:11:01] <gregory38> 2 cluts buffer (quite costly for the hardware)
[09:11:09] <gregory38> or clut loading is defered at draw time
[09:15:01] <gregory38> Or maybe only load clut when tex0 is written if the context is the good one
```
Note that potentially core emulation is wrong too and generate out-of-order register.

@gabest11 your opinion is more than welcome :)

gregory38:
I've done something on #1550 I'm not 100% sure the new behavior is correct. Feel free to give me any feedback

refractionpcsx2:
Register information is processed in the order it is sent  in the core, so I'm not sure how that could be wrong. 
 The only scenario where it might be different is if path3  masking is used.

gregory38:
EE thread executed in wrong order. But it is unlikely to be the issue. Still behavior is quite strange.

Did someone got any issue with this PR and save state? If it is good, I propose to merge it.

Issue #652, opened 2015-07-12
IOCTL_DVD_START_SESSION is used even for CD disks


mirh:
While I was investigating for #648, I noticed in my cd-rom only drive that this request failed. 
_0x335000 (Device:0x33 Function:1024 Method: 0)_

It's not like this seemed to have any particular consequence.. still I guess using a simplest, more essential call would be more elegant.

turtleli:
Well, it can be reduced to once per disc/pause and resume/other situations that I can't currently think of instead of n times per second which it currently does (at least for my drive).

The code only requires IOCTL_DVD_START_SESSION to recognise that a DVD is in the drive and to obtain the DVD's physical format information. A DVD session doesn't have to be open to read sectors from the DVD.

mirh:
I see, but my suggestion was more about using another control code for CD-ROMs.

turtleli:
The call is needed for DVD detection, the alternatives are to
 * use SPTI with a READ DISC STRUCTURE command;
 * read the disc contents to determine whether it's a CD or DVD.

Both methods are probably more complicated than the current disk detection method.

mirh:
Mhh.. Well, SPTI support is already there after all.. 
So I guess this (and the second option ofc) should just apply to the fallback path?

Your time to spend anyway, your choice d:

turtleli:
I don't think it's worth doing. In the SPTI case. it'd only change the call that fails.

My only issue with IOCTL_DVD_START_SESSION in the plugin is that it's called unnecessarily (which I'll take care of at some point).

turtleli:
Closing - I've minimised the use of the ioctl, and I don't think there's anything more that can really be done.

mirh:
I wouldn't really know if that's what I was "complaining"

..anyway, too bad I haven't any cd-rom-only reader available anymore to test the calls.

turtleli:
I quote myself:

>The call is needed for DVD detection,

And the ioctl fails on DVD drives too if the DVD isn't inserted. And on the alternatives:

> I don't think it's worth doing. In the SPTI case. it'd only change the call that fails.

So there really is nothing that can be done.

mirh:
> the alternatives are to [...]: read the disc contents to determine whether it's a CD or DVD.

I thought this was what this issue was about. How does the ps2 behave in this case?
Or is this just one of the many mysteries of the player firmware?

turtleli:
[...]

PR #1545, opened 2016-09-02
cdvdgigaherz: Use SPTI to read raw CD sectors

### LINKED ISSUE #648: IOCTL_CDROM_RAW_READ fails for CD-ROM games in many drives
I wouldn't know if "many" is really the right determiner. 
My narrow experience though [tell](http://web.archive.org/web/20021206183205/http://www.btceurope.nl/product_home.html) me [that](https://web.archive.org/web/20030213004916/http://www.lge.com/c_product/catalog/product_information.jsp?mdl_pfx=4CDWS&mdl_cod=GCE-8400B&mdl_mak=1) out [of](https://web.archive.org/web/20090411041856/http://www.liteonit.com/global/index.php?option=com_content&task=view&id=251&Itemid=67) 9 [readers](https://web.archive.org/web/20120604031008/http://www.samsung.com/my/consumer/pc-peripherals-printer/optical-disc-drive/dvd-rom/SH-D162D/RSBP) I [tried](https://web.archive.org/web/20120613182714/http://www.samsung.com/ca/consumer/office/optical-disk-drives/optical-disk-drives/SH-S223C/BEBE) only [four](http://web.archive.org/web/20070402110200/http://www.sonynec-optiarc.eu/en/exhibits/hh-drives/ad-7170a.html) [could](http://web.archive.org/web/20150906164550/http://hlds.co.kr/product/e_dvd_series.php#GT34N) [handle](http://club.myce.com/f44/dh-16a6s-crossflashing-304838/#post2469234) [CDs](http://www.lg.com/us/burners-drives/lg-GUB0N). 

When pcsx2 enters "2340" mode (which has to be read with either XAForm2 or YellowMode2) I can see in Process Monitor that device says *Control: 0x2403e (Device:0x2 Function:15 Method: 2)* is an invalid request. 

Emulogs (if required) are [here](http://forums.pcsx2.net/Thread-Bug-Report-Tom-Jerry-s-War-of-the-Wiskers-PAL). 

One way to workaround this would be using SPTI as ripper softwares do (as gigaherz suggested), but there must be a simpler way to make stupid drive understand the normal command. 

Given I can workaround this with ISOs I won't offend if somebody put a *Low Priority* or *longterm* label.

turtleli:
IOCTL_CDROM_RAW_READ apparently does not work for some read modes on some optical drives, which makes some CD-ROM games unplayable from the disc.

Work around the issue by using SPTI to retrieve the raw sector data. The old reading method has been retained in case SPTI cannot be used (if the device could not be opened with write access).

Note: While I have tested the SPTI code, I haven't tested it within cdvdGigaherz (I don't own any PS2 CD-ROM games).

Should fix #648

### DIFF
--- a/plugins/cdvdGigaherz/src/CDVD.h
+++ b/plugins/cdvdGigaherz/src/CDVD.h
@@ -90,7 +90,7 @@ class IOCtlSrc: public Source
 	IOCtlSrc(IOCtlSrc&);
 
 	HANDLE device;
-
+	bool m_can_use_spti;
 	bool OpenOK;
 
 	s32 last_read_mode;
--- a/plugins/cdvdGigaherz/src/Windows/IOCtlSrc.cpp
+++ b/plugins/cdvdGigaherz/src/Windows/IOCtlSrc.cpp
@@ -18,6 +18,12 @@
 #include <winioctl.h>
 #include <ntddcdvd.h>
 #include <ntddcdrm.h>
+// "typedef ignored" warning will disappear once we move to the Windows 10 SDK.
+#pragma warning (push)
+#pragma warning (disable: 4091)
+#include <ntddscsi.h>
+#pragma warning (pop)
+
 #include <cstddef>
 #include <cstdlib>
 
@@ -284,13 +290,15 @@ s32 IOCtlSrc::Reopen()
 	DWORD size;
 
 	OpenOK = false;
-
+	// SPTI only works if the device is opened with GENERIC_WRITE access.
+	m_can_use_spti = true;
 	device = CreateFile(fName, GENERIC_READ|GENERIC_WRITE|FILE_READ_ATTRIBUTES, share, NULL, OPEN_EXISTING, flags, 0);
 	if(device==INVALID_HANDLE_VALUE)
 	{
 		device = CreateFile(fName, GENERIC_READ|FILE_READ_ATTRIBUTES, share, NULL, OPEN_EXISTING, flags, 0);
 		if(device==INVALID_HANDLE_VALUE)
 			return -1;
+		m_can_use_spti = false;
 	}
 	// Dual layer DVDs cannot read from layer 1 without this ioctl
 	DeviceIoControl(device, FSCTL_ALLOW_EXTENDED_DASD_IO, nullptr, 0, nullptr, 0, &size, nullptr);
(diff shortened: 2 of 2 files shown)


FlatOutPS2:
I can't be 100% sure, as it doesn't always fail, but I had no crashes with this PR. and 2 with the master. So I assume it works. :D

mirh:
Oh god, it really does work, tested with Tom and Jerry War of The Whiskers and Donald Duck - Phantomias - Platyrhynchos Kineticus. 
My biggest thanks.

mirh:
Oh, also, while you are at it.. Could you check #652 ?

turtleli:
Well, guess it does work, so I can merge. :)

MrCK1:
@turtleli I'm not sure if we need to look at this again in the future, but SPTI does not seem to work for some PS1 games when reading from a physical drive. I'll rip the game to a .bin file in the meantime.

Console Snippet:
```
 * CDVD: CD-ROM read mode change
 * CDVD: Trying XAForm2
 * CDVD: Trying YellowMode2
 * CDVD: Trying CDDA
 * CDVD: Failed to read this CD-ROM with error code: 1
 * CDVD: SPTI failed reading sectors 31520-31535
```

turtleli:
I probably need to see the CD TOC. I added a comment in the SPTI commit:
```
// FIXME: Probably doesn't work if the sectors to be read are from two tracks
// of different types.
```
If that's what the issue is then it shouldn't be too hard to fix.

MrCK1:
Do you want me to post the emulog or anything else?

turtleli:
The emulog would probably help.

MrCK1:
[Here you go](http://pastebin.com/Z50sXpwP) :)

turtleli:
Yeah, it's reading from a data track and audio track in the same read - the MMC specs say that will fail.

mirh:
MultiMediaCard?

turtleli:
SCSI Multi-Media Commands

mirh:
Lmao. 

Could [this](http://forum.osdev.org/viewtopic.php?f=13&t=29352) help anyway?

turtleli:
Well, I know what needs to be done. I can either:
 * split the read so it's done sector by sector (I have to do this for raw sector reads in Linux anyway unless I use the SG_IO ioctl); or
 * use the TOC to split the read up into multiple parts when necessary

I just like to take my time (over)thinking about things.

PR #1550, opened 2016-09-04
gsdx state: TEX0 write behavior might depends on the context

gregory38:
Game: harley davidson
* write tex0 ctx0
* write tex0 ctx1
* draw  ctx 0

Previous GSdx behavior will load the clut every write of TEX0. In the
above case the draw will take the wrong clut.

To be honest, it could be a wrong emulation on the EE core emulation.
The hardware likely got a single clut (1KB cache is quite expensive)
So clut loading must be skipped if the context is wrong.

Next draw will use the ctx1 clut so I apply TEX0 when the context is switched

Please test harley davidson :)

### DIFF
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -821,7 +821,8 @@ void GSState::GIFRegHandlerXYZ2(const GIFReg* RESTRICT r)
 
 template<int i> void GSState::ApplyTEX0(GIFRegTEX0& TEX0)
 {
-	// even if TEX0 did not change, a new palette may have been uploaded and will overwrite the currently queued for drawing
+	// Mask invalid bit before any comparison
+	TEX0.CPSM &= 0xa; // 1010b
 
 	// Handle invalid PSM here
 	switch (TEX0.PSM) {
@@ -831,6 +832,20 @@ template<int i> void GSState::ApplyTEX0(GIFRegTEX0& TEX0)
 		default:
 			break;
 	}
+
+	if (PRIM->CTXT != i) {
+		// Context isn't current. Only update the register value.
+		// Handling (real write of the clut buffer) will be done when ctxt will be switched
+		if((TEX0.u32[0] ^ m_env.CTXT[i].TEX0.u32[0]) & 0x3ffffff) // TBP0 TBW PSM
+		{
+			m_env.CTXT[i].offset.tex = m_mem.GetOffset(TEX0.TBP0, TEX0.TBW, TEX0.PSM);
+		}
+
+		m_env.CTXT[i].TEX0 = (GSVector4i)TEX0;
+		return;
+	}
+
+	// even if TEX0 did not change, a new palette may have been uploaded and will overwrite the currently queued for drawing
 
 	bool wt = m_mem.m_clut.WriteTest(TEX0, m_env.TEXCLUT);
 
@@ -843,7 +858,6 @@ template<int i> void GSState::ApplyTEX0(GIFRegTEX0& TEX0)
 		Flush();
 	}
 
-	TEX0.CPSM &= 0xa; // 1010b
 
 	if((TEX0.u32[0] ^ m_env.CTXT[i].TEX0.u32[0]) & 0x3ffffff) // TBP0 TBW PSM
 	{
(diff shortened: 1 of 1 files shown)


Nobbs66:
I only tested Kingdom Hearts and Burnout 3, but they both crash. Burnout has graphics issues on the loading screen before a race, but then crashes once it goes in game.

Burnout 3 loading screen http://i.imgur.com/M4tXFqn.png

plugins/GSdx/GSState.cpp
```diff
@@ -809,6 +821,16 @@ void GSState::GIFRegHandlerXYZ2(const GIFReg* RESTRICT r)
 
 template<int i> void GSState::ApplyTEX0(GIFRegTEX0& TEX0)
 {
+	// Mask invalid bit before any comparison
+	TEX0.CPSM &= 0xa; // 1010b
+
+	if (PRIM->CTXT != i) {
+		// Context isn't current. Only update the register value.
+		// Handling (real write of the clut buffer) will be done when ctxt will be switched
+		m_env.CTXT[i].TEX0 = (GSVector4i)TEX0;
```
FlatOutPS2: This line causes crash in Burnout 3 mentioned by @Nobbs66

gregory38: ????
Do you have more info  ? the callstack? I don't get it, the line was done below.

FlatOutPS2: If you add the following line (which is also done below) above line 830 it doesn't crash anymore.
```
		if ((TEX0.u32[0] ^ m_env.CTXT[i].TEX0.u32[0]) & 0x3ffffff) // TBP0 TBW PSM
		{
			m_env.CTXT[i].offset.tex = m_mem.GetOffset(TEX0.TBP0, TEX0.TBW, TEX0.PSM);
		}
```

gregory38: You mean if we set a value to offset.tex ?

gregory38: Maybe those game toys with PRMODE register. And I didn't get all the context switch. Let's me try to do something.

FlatOutPS2: Yes, apparently. The same if statement is above the `m_env.CTXT[i].TEX0 = (GSVector4i)TEX0;` at line 859.

gregory38: Hum, I pushed another commits that re apply tex0 for all context switches. The goal is to not use data of context 1 when you do a draw of context 0. OTOH, this line is dedicated to a single context so we could add it.

gregory38: Ok. I get it, and yes you're right. We need to do it due to the if condition. Otherwise next call will bypass it. Good catch :+1:

FlatOutPS2:
Tried a few games with the latest changes and noticed no further regressions.

Nobbs66:
Yep, just tried a few myself, and everything is good.

gregory38:
Oh cool. I squased all commits and rebased on master (so it also include the texture transfer bug fix). This way it would be easier to test others games that have issue with the SW renderer (even if I have the feeling that it is 1 bug per game)

FlatOutPS2:
Do you have a screenshot of the Harley Davidson issue? It might be useful to recognise a game with a similar problem.

gregory38:
http://imgur.com/a/ceRlV

2nd picture. (3rd was a hack tentative). First one is reference. Basically the issue is that you have the wrong clut for the texture. So you have random colors.

MrCK1:
GT4 had some discoloration in the transitons while viewing the dive replay; but it's probably another one of my pointless observations :P

refractionpcsx2:
> GT4 had some discoloration in the transitons while viewing the dive replay; but it's probably another one of my pointless observations :P

hasn't GT4 had that as far back as it ran? :P

sudonim1:
fwiw the documentation quite explicitly says this doesn't happen.  Of
course docs are just docs, but two temporary clut buffers seems more likely
(and more useful to games programmers).  Failing that I expect it's
actually something stranger than we are thinking of currently.

gregory38:
I won't say more likely
* a 8 Kilo Bits memory is huge from a silicon point of view.
* doc doesn't say anywhere that there is 2nd caches whereas it is a crucial information for game dev to extract as most performance as possible.

So far nothing make sense.

refractionpcsx2:
It's possible it doesn't make sense, devs regularly took advantage of "quirks" in the hardware, this could be one of them.

gregory38:
```
 took advantage 
```
You mean code garbage that still render a nice image and don't look further.

From an HW point of view, I suspect the context mask the write on the alternate TEX0 register. Then the context switch will likely trigger the TEX0 update.

Here, the way I saw the HW is potentially coded.
```
current_tex0 = (ctx == 1) ? tex0_1 : tex0_0;

@posedge clk {
  previous_tex0 <= current_tex0; // register that will delay current_tex0 of 1 cycle
}

if (previous_tex0 != current_tex0) {
  // TEX0 update detected, please update CLUT mess.
}
```

HW test will be nice :+1:

refractionpcsx2:
> You mean, devs code garbages that still render a nice image and don't look further.

precisely :P

gregory38:
I rebased the PR on master. So you can also play with the new hack too.

gregory38:
@sudonim1 
Would you do some hardware test ? Please & thanks :)

gregory38:
Let's merge the code to increase test coverage on it.

FlatOutPS2:
This PR has triggered a regression in Gran Turismo 4: http://forums.pcsx2.net/Thread-Gran-Turismo-4-regression?pid=536948

pepiczech77:
Probably there were 2 different regressions, see updated forum topic.

PR #1554, opened 2016-09-08
gsdx ogl: only use geometry shader to convert big enough draw call

gregory38:
The purpose of geometry shader is to reduce bandwidth (72 bytes by sprite)
and CPU load.

Unfortunately it increases CPU load due to extra shader validations.

So geometry shader will only be enabled for draw call with more than
16 sprites (arbitrarily, smallest number before shadow hearts plummet)

I need performances testing @FlatOutPS2

### DIFF
--- a/plugins/GSdx/GSRendererOGL.cpp
+++ b/plugins/GSdx/GSRendererOGL.cpp
@@ -175,7 +175,9 @@ void GSRendererOGL::SetupIA(const float& sx, const float& sy)
 
 			// If the draw calls contains few primitives. Geometry Shader gain with be rather small versus
 			// the extra validation cost of the extra stage.
-			if (GLLoader::found_geometry_shader &&  m_vertex.next > 32) { // <=> 16 sprites (based on Shadow Hearts)
+			//
+			// Note: keep Geometry Shader in the replayer to ease debug.
+			if (GLLoader::found_geometry_shader && (m_vertex.next > 32 || GLLoader::in_replayer)) { // <=> 16 sprites (based on Shadow Hearts)
 				m_gs_sel.sprite = 1;
 
 				t = GL_LINES;
--- a/plugins/GSdx/GSRendererOGL.cpp
+++ b/plugins/GSdx/GSRendererOGL.cpp
@@ -162,7 +162,20 @@ void GSRendererOGL::SetupIA(const float& sx, const float& sy)
 			break;
 
 		case GS_SPRITE_CLASS:
-			if (GLLoader::found_geometry_shader) {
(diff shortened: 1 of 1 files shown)


gregory38:
Games that are potentially impacted in good way: `shadow hearts` & `guilty gear`

FlatOutPS2:
No change with Guilty Gear. Tested OpenGL HW native and 4x, performance remains the same between master and PR.

gregory38:
Could you try the Guilty Gear without geometry shader ?

FlatOutPS2:
Tried it without geometry shader and then also tried the PR again and I did now notice an fps increase of 5-10% with both gs off and PR. Something must've been holding it back earlier. :S

gregory38:
Are you sure you didn't test the PR with gs off ? Maybe 32 ceil is a bit low. I would love to have performance test on the linux driver too.

Hum, I need to disable this perf optimization in the replayer. It would be too annoying for debug otherwise.

plugins/GSdx/GSRendererOGL.cpp
```diff
@@ -162,7 +162,20 @@ void GSRendererOGL::SetupIA(const float& sx, const float& sy)
...
+			// Note: severals openGL operation does draw call under the wood like texture upload. So even if
+			// you do 10 consecutive draw with the geometry shader, you will still pay extra validation if new
+			// texture are uploaded. (game Shadow Hearts)
+			//
+			// Note2: Due to MultiThreaded driver, Nvidia suffers less of the previous issue. Still it isn't free
+			// Shadow Heart is 90 fps (gs) vs 113 fps (no gs)
+
+			// If the draw calls contains few primitives. Geometry Shader gain with be rather small versus
+			// the extra validation cost of the extra stage.
+			if (GLLoader::found_geometry_shader &&  m_vertex.next > 32) { // <=> 16 sprites (based on Shadow Hearts)
```
gregory38: Keep standard line in replayer to ease debug.

FlatOutPS2:
> Maybe 32 ceil is a bit low.

I think 32 is about the correct value.

gregory38:
```
I think 32 is about the right value.
```
It is hard to judge. On openGL the gs is really a performance trade off.

It saves bandwidth. But PCIe3 ought to be fine, ofc, APU ought to good too. It could help to reduce the CPU load but the driver requires extra work to support the gs stage. It is quite hard to quantify the later. Beside, driver and gsdx don't run on the same thread on Nvidia. So it is even harder to benchmark it. Geometry shader is quite slow on the GPU. Not easy to find a good trade-off.

PR #1431, opened 2016-06-23
Osd manager

JasonBrownDeveloper:
This is no where near ready to merge, but I wanted to get eyes on what I have.  Most of the base functionality is done.  I'm still going to add some gettext support and add a settings page.

PR #1557, opened 2016-09-10
3rdparty: Add Freetype and link to GSdx

turtleli:
Necessary for #1431. Possibly could be trimmed slightly (would save 2-3MB).


gregory38:
I'm not against to trim it a bit. What is the size currently ?

turtleli:
11MB.

gregory38:
Yes trim it :)

turtleli:
Trimmed - size is now 8.2MB.

gregory38:
So much to draw 10 letters on the screen. Anyway, we can't do better. You can merge it.

Actually won't it be possible to only upload a binary file ? (the .a equivalent on windows). Or does it depends on the windows version ?

turtleli:
Well, we'd still need the include files (1.8MB), and the compiled static .lib file is 5.7MB (VS2015). So we won't be saving much space.

PR #1566, opened 2016-09-15
Link Unconditionally

Nobbs66:
Link instructions used to store the return address if the branch was
taken, but the correct behavior is to store the return address whether
or not the branch is taken.

### DIFF
--- a/pcsx2/Interpreter.cpp
+++ b/pcsx2/Interpreter.cpp
@@ -299,10 +299,10 @@ void BGEZ()    // Branch if Rs >= 0
 
 void BGEZAL() // Branch if Rs >= 0 and link
 {
-
+	_SetLink(31);
 	if (cpuRegs.GPR.r[_Rs_].SD[0] >= 0)
 	{
-		_SetLink(31);
+		
 		doBranch(_BranchTarget_);
 	}
 }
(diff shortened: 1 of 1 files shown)


sudonim1:
Yup, that's correct behaviour and the recompiler already does this.   I'm surprised the interpreter has such a blatant bug.  The diff looks good although you should squash your commits into one.

After squashing I don't see any reason this shouldn't be merged.

refractionpcsx2:
We can squash in the merge, it's fine.

FlatOutPS2:
Great. You just set a precedent for us to never squash our commits again. :p

refractionpcsx2:
Well you don't need to if Github gives us the option to do it for you :P

avih:
> Great. You just set a precedent for us to never squash our commits again. :p

Well, it depends. Currently github interface only allows squashing all the commits into one (and it rebases while at it - which is a good thing).

But if the PR consists of several distinct functionalities which should remain as such, then the GH interface doesn't allow custom squash (like in `rebase -i`), and also, it's the author's job to make sure the commits are clean, as s/he knows best how to split them in a meaningful way.

Rebasing would also be appreciated especially if the commits parent is a non-recent commit - it makes the git history after the merge much less cluttered (and PCSX2's history with merges is super ugly due to this).

PR #1569, opened 2016-09-16
cdvdgigaherz: Remove ISO file reader

turtleli:
It doesn't support dual layer DVD ISOs, and the ini has to be edited manually so it loads an ISO image ("$" has to be prepended to the ISO path as well). The PCSX2 internal ISO file reader is probably better in
most/all aspects and I don't think it's worth copying the logic from PCSX2 into the plugin.

Also, it makes refactoring the code a pain, which I want to do before I port the plugin to Linux (I have stuff working, but I need to do config + gui stuff + misc stuff I can't remember)


gregory38:
```
Also, it makes refactoring the code a pain, which I want to do before I port the plugin to Linux (I have stuff working, but I need to do config + gui stuff + misc stuff I can't remember)
```
:cupid:

ramapcsx2:
Would be very nice if you ported this to Linux. We will soon need a cdvd plugin that can stream cd audio. It'd be best to make it a multi platform one ;p

gregory38:
actually potentially code was imported from the code.

Do you know if the plugin support block dump ?

turtleli:
It can't create or read block dumps.

I don't think it's well known, but PCSX2 can create block dumps without an external plugin (change CdvdDumpBlocks in PCSX2_vm.ini to enabled).

gregory38:
You can read dump with the null reader. Need to check if the internal dump works well. It will be a good opportunity to move/delete the others cdvd plugins.

turtleli:
> You can read dump with the null reader.

? Do you mean the internal ISO reader? Pretty sure the null reader does nothing.

gregory38:
Yeah sorry I meant the internal iso reader.

gregory38:
```
[16:46:24] <gigaherz> mostly
[16:46:33] <gigaherz> the only use it would have would be for devices with very slow HDDs
[16:46:53] <gigaherz> since the cachhing/readahead may be a bit more effective than the internal iso reader
[16:46:57] <gigaherz> but it doesn't really matter
[16:47:10] <gigaherz> you can mount the iso in like, ImDisk or any other virtual cd
[16:47:21] <gigaherz> and t hen use that with the device reader
[16:47:44] <gregory38> ok
```
So it can be removed with any 2nd thought

PR #1331, opened 2016-04-30
PCSX2: Implement GIF FIFO

refractionpcsx2:
Spent the last couple of days playing with this to see if I could get it to work. so far it seems to be working.

What this PR brings:
- Fixes DJ Hero (so it now boots and with right colours)
- Removes an always on hack for Wallace & Gromit which could have interfered with other games
- Removes a hack from the Gamefixes section for Hot Wheels World Race, this is no longer required
- Added the appropriate entries in the GameDB for the games mentined.
- Savestate version bump - Your old savestates will not work from here on. Sorry, not sorry :P 

This FIFO is not always on, it is enabled in the Gamefixes tab as people will moan about the speed loss (anywhere from 5-10fps) so it can be used for certain games which refuse to boot, there might be others out there which need this and that's why they don't boot, but I can't remember what games don't boot :P If you find one that still doesn't boot with it on, please don't report it, that was just a passing thought.

What needs testing? Well, everything you can, with the option on and off to make sure compatibility hasn't been compromised.

PR #1577, opened 2016-09-20
(Minor) Mainframe: Fixed shutdown UI function to properly disable itself afte…

MrCK1:
- The shutdown function no longer stays enabled after the VM has been wiped, ect. Previously, the button would not disable itself after a shutdown and could confuse the user
- I also made a small changes to a the top line in the EE/VU panels (pet-peeve :P) and rewrote the language dialog to sound more fluent.

### DIFF
--- a/pcsx2/gui/Dialogs/SysConfigDialog.cpp
+++ b/pcsx2/gui/Dialogs/SysConfigDialog.cpp
@@ -275,13 +275,13 @@ Dialogs::InterfaceConfigDialog::InterfaceConfigDialog(wxWindow *parent)
 }
 
 Dialogs::InterfaceLanguageDialog::InterfaceLanguageDialog(wxWindow* parent)
-	: BaseConfigurationDialog(parent, _("Language selector"), 400)
+	: BaseConfigurationDialog(parent, _("Language Selector"), 400)
 {
 	*this += 5;
 
 	// Keep this in English - same as the menu item.
-	*this += Heading(L"New language will affect newly opened windows.\n");
-	*this += Heading(L"Full change will happen after restarting PCSX2.");
+	*this += Heading(L"Some menu items may not change until PCSX2 is restarted\n");
+	*this += 2;
 	*this += new Panels::LanguageSelectionPanel(this, false) | StdCenter();
 
 	AddOkCancel();
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -665,7 +665,7 @@ void MainEmuFrame::ApplyCoreStatus()
 		}
 	}
 
-	menubar.Enable( MenuId_Sys_Shutdown, SysHasValidState() || CorePlugins.AreAnyInitialized() );
+	menubar.Check( MenuId_Sys_Shutdown, SysHasValidState() || CorePlugins.AreAnyInitialized() );
 }
 
 //Apply a config to the menu such that the menu reflects it properly
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -665,7 +665,7 @@ void MainEmuFrame::ApplyCoreStatus()
 		}
 	}
 
-	menubar.Check( MenuId_Sys_Shutdown, SysHasValidState() || CorePlugins.AreAnyInitialized() );
+	menubar.Enable( MenuId_Sys_Shutdown, SysHasValidState());
 }
 
 //Apply a config to the menu such that the menu reflects it properly
--- a/pcsx2/gui/Panels/CpuPanel.cpp
+++ b/pcsx2/gui/Panels/CpuPanel.cpp
@@ -109,7 +109,7 @@ Panels::AdvancedOptionsVU::AdvancedOptionsVU( wxWindow* parent )
 Panels::CpuPanelEE::CpuPanelEE( wxWindow* parent )
 	: BaseApplicableConfigPanel_SpecificConfig( parent )
 {
-	*this	+= Text( pxE( L"Notice: Most games are fine with the default options. ")
+	*this	+= Text( pxE( L"Notice: Most games are fine with the default options.")
 	) | StdExpand();
 
 	const RadioPanelItem tbl_CpuTypes_EE[] =
--- a/nsis/nsis_instructions.txt
+++ b/nsis/nsis_instructions.txt
@@ -2,12 +2,12 @@
   PCSX2 NSIS Installer Instructions
 -------------------------------------
 
- * Install NSIS (tested with 3.00), make sure to include the "Modern User Interface"
+ * Install NSIS (tested with 2.46), make sure to include the "Modern User Interface"
  * Install NSIS Script: Advanced Uninstall Log (you can find it on NSIS wiki)
    ( Currently at http://nsis.sourceforge.net/Advanced_Uninstall_Log_NSIS_Header )
 
   * Download the Visual C++ 2015 Redistributable and save it to this nsis/ folder.
-   ( https://download.microsoft.com/download/9/3/F/93FCF1E7-E6A4-478B-96E7-D4B285925B00/vc_redist.x86.exe )
+   ( https://www.microsoft.com/en-us/download/details.aspx?id=49984 )
    Rename the download to "vcredist_2015_Update_1_x86".
    
  * Download the DirectX Web Installer (dxwebsetup.exe) from Microsoft's website and save it to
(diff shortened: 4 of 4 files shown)


pcsx2/gui/MainFrame.cpp
```diff
@@ -665,7 +665,7 @@ void MainEmuFrame::ApplyCoreStatus()
 		}
 	}
 
-	menubar.Enable( MenuId_Sys_Shutdown, SysHasValidState() || CorePlugins.AreAnyInitialized() );
+	menubar.Check( MenuId_Sys_Shutdown, SysHasValidState() || CorePlugins.AreAnyInitialized() );
```
turtleli: I don't think this is correct. The shutdown option isn't a checkable item, nor would it make sense to make it a checkable item.

MrCK1: I guess it's possible, (looking at the [Wx Reference](http://docs.wxwidgets.org/trunk/classwx_menu_bar.html#ab53b19f2e8b0e7b39ae551d51f2f6f6b) ) but then why doesn't this function correctly with enable used. I feel like this worked at one point, but I can't recall

pcsx2/gui/Panels/CpuPanel.cpp
```diff
@@ -109,7 +109,7 @@ Panels::AdvancedOptionsVU::AdvancedOptionsVU( wxWindow* parent )
 Panels::CpuPanelEE::CpuPanelEE( wxWindow* parent )
 	: BaseApplicableConfigPanel_SpecificConfig( parent )
 {
-	*this	+= Text( pxE( L"Notice: Most games are fine with the default options. ")
+	*this	+= Text( pxE( L"Notice: Most games are fine with the default options.")
```
gregory38: Is it really an improvement ? Is it worth it ? I mean all people will need to revalidate the translation.

pcsx2/gui/MainFrame.cpp
turtleli: > I guess it's possible, (looking at the Wx Reference ) but then why doesn't this function correctly with menubar.Enable ? I though this worked at one point, but I can't recall

Well, ` SysHasValidState() || CorePlugins.AreAnyInitialized()` must be returning `true`.
 
> Edit: maybe there is a bug in Wx? People here seem to note that menubar.Check behaves differently on certain platforms. It's not defined in the documentation specifically (IMO) but isn't this just checking if it was clicked? (not a physical checkbox?)

`Check()` changes whether a checkmark is present.. The documentation is pretty clear IMO.

ssakash: I just took a quick look into the issue and apparently this is caused due to GSdx being initialized even after a shutdown of the VM.

ssakash: Well, thinking again I don't think the issue is specific to GSdx since ZeroGS also exhibits the same behavior.

So a quick fix would be to probably change,

```
menubar.Check( MenuId_Sys_Shutdown, SysHasValidState() || CorePlugins.AreAnyInitialized() )
```

**to**

```
menubar.Check( MenuId_Sys_Shutdown, SysHasValidState())
```

ssakash: @turtleli - what's your opinion ? Is there some sort of bug on ``CorePlugins`` which indicates that the GS plugin is initialized even after a shutdown ? If yes, this needs to be fixed instead of omitting `` || CorePlugins.AreAnyInitialized()`` on the ``Enable()`` function. That's my opinion on the matter.

turtleli: Are you sure it's only the GS plugin that still reports it's initialised? It could be that all the plugins have not been shutdown yet.

ssakash: > Are you sure it's only the GS plugin that still reports it's initialised? It could be that all the plugins have not been shutdown yet.

Derp, it seems you're right. All the plugins seem to be initialized even after a shutdown event.

 I'll make a blog post later explaining on how I screwed it up thinking only the GS plugin was initialized. Jokes aside, shouldn't we then fix the issue by properly shutting down the plugins when ``Shutdown`` is pressed ?

MrCK1:
@turtleli FYI, I did notice this change between the original function and my changes while testing.

The normal function is 'menubar.Enable' - the Shutdown option is greyed out when you first launch the exe; but it doesn't get disabled when Shutdown is clicked .

When I changed the function to 'menubar.Check' the Shutdown isn't greyed out when the exe is first launch; but it does get greyed out when shutdown is clicked

pcsx2/gui/MainFrame.cpp
turtleli: Well, I don't know if it's even useful to shutdown a plugin.

MrCK1: @turtleli FYI, I did notice this change between the original function and my changes while testing.

The normal function is ```menubar.Enable``` - the Shutdown option is greyed out when you first launch the exe; but it doesn't get disabled when Shutdown is clicked .

When I changed the function to ```menubar.Check``` the Shutdown isn't greyed out when the exe is first launch; but it does get greyed out when Shutdown is clicked

turtleli:
> The normal function is menubar.Enable - the Shutdown option is greyed out when you first launch the exe; but it doesn't get disabled when Shutdown is clicked .
> 
> When I changed the function to menubar.Check the Shutdown isn't greyed out when the exe is first launch; but it does get greyed out when Shutdown is clicked

The menu option is being disabled elsewhere then. Perhaps the line isn't even necessary. `Check()` will likely throw an assertion in a debug build for a non-checkable item.

On changing existing strings - it'd probably be better to open an issue on what people want changed and for what reason. If there's agreement that it should be done, then it'd be nicer to do it all at once (it saves having to update the translation files n times over).

ssakash:
> The menu option is being disabled elsewhere then. Perhaps the line isn't even necessary.

Yes, there's a ``Bind()`` event which grays out the menu option. However the line is indeed necessary. If you remove it, the ``Shutdown`` menu item will look active upon launch of PCSX2 where it ought to be grayed out.

So maybe it would be better to follow the [quick fix](https://github.com/PCSX2/pcsx2/pull/1577#discussion_r79802209) which I had mentioned earlier?

MrCK1:
@ssakash Just compiled your fix in a new build - you were right on the bulls-eye ;)

Should we wait until #1331 is done so we only need to update translation files once?

turtleli:
> Just compiled your fix in a new build - you were right on the bulls-eye ;)

You may have missed reading the previous posts on an alternative?

> Should we wait to merge until #1331 is done so we only need to update translation files once?

My suggestion was:
1. Open an issue to collect changes that people want on existing strings. (I'm sure people wanted Iso to be changed to ISO at some point)
2. Discuss.
3. Update all at once if it's agreed that it's worth changing.

Changing existing strings slightly requires manual intervention to preserve the translations. And there are currently 23 localisations.

pcsx2/gui/Dialogs/SysConfigDialog.cpp
```diff
@@ -275,13 +275,13 @@ Dialogs::InterfaceConfigDialog::InterfaceConfigDialog(wxWindow *parent)
...
 Dialogs::InterfaceLanguageDialog::InterfaceLanguageDialog(wxWindow* parent)
-	: BaseConfigurationDialog(parent, _("Language selector"), 400)
+	: BaseConfigurationDialog(parent, _("Language Selector"), 400)
 {
 	*this += 5;
 
 	// Keep this in English - same as the menu item.
-	*this += Heading(L"New language will affect newly opened windows.\n");
-	*this += Heading(L"Full change will happen after restarting PCSX2.");
+	*this += Heading(L"Some menu items may not change until PCSX2 is restarted\n");
```
turtleli: Previous text is more correct.

MrCK1:
Oh sorry. It's easy to miss stuff as ForkHub doesn't let you view comments on diffs; so I have to check the website for that :P

gregory38:
Could I have a quick status/summary ?

MrCK1:
Right now we're discussing other strings that could be changed in issue #1620 as you probably saw already. Until then this sits :p

pcsx2/gui/Dialogs/SysConfigDialog.cpp
piorrro33: The previous text was more correct but the PR is clearer and more concise IMO.

FlatOutPS2: Maybe change it to:
```
*this += Heading(L"Language switch will only affect newly opened windows.\n");
*this += Heading(L"Full change will not apply until PCSX2 is restarted.\n"); 
```

piorrro33: This one sounds better.

jcdenton2k:
I'm one of those that wanted to see Iso changed to ISO just for the sake of keeping things clean/uniform and easier visibility for the end-user.

turtleli:
Resolved by #1651 (well, maybe except for the space after the period). Closing.

PR #1579, opened 2016-09-20
Gsdx generic double half clear

gregory38:
Implement a generic half clear

Remove CRC of those games.
    * SpidermanWoS
    * TyTasmanianTiger
    * TyTasmanianTiger2
    * DigimonRumbleArena2
    * XmenOriginsWolverine
    * CallofDutyFinalFronts
    * GoW2 (only in openGL)

Note: it seems tale of legendia is fixed too. Could someone test the fifa games ?


FlatOutPS2:
FIFA games are not affected. But FIFA doesn't have the half clear issue, so it was to be expected.

gregory38:
You mean they only suffer of an half screen bug. But not the clear one. Well it is possible, my memory is just wrong.

FlatOutPS2:
Yes, the games just miss the bottom half of the the green and dark green patches on the field.
You can see an example here: http://forums.pcsx2.net/Thread-GSDX-Hardware-mode-Bug-Report-FIFA-14-PAL

Issue #882, opened 2015-10-07
Recommended CRC hack level GSDX HW


willkuer:
Current implementation of CRC hack levels has the problem that the default/recommended CRC hack levels are different for different renderers. I would suggest to add another level called 'recommended' or 'automatic' which selects the appropriate CRC hack level for OGL ('Partial') and Dx9/11 ('Full') on configload.

I think this should increase usability and is more or less cheap to implement. If you consider this to be a reasonable extension I could try to do that by myself. (Probably one of the devs only needs 10 minutes so you can do it as well by your own.)

willkuer:
My idea would be to add another element into the first position of the crc hack combobox. The saved value would be the itemposition-1. Like that all settings are compatible to old versions and 'automatic' would get the magic number -1, which seems to me appropriate. On getconfig(CRC level) the recommended level would be selected.

What do you think?

willkuer:
Finished it in #62ba3fc

gregory38:
Is it done? Could it be closed?

willkuer:
It's not done and requires #1018 to fulfill the original idea.

willkuer:
I close it for the moment. There seems not too much interest in the forum for such a function. It would be good though. Maybe I will find time later to do it.

PR #1582, opened 2016-09-22
GSdx: Automatically set default CRC Hack Level

FlatOutPS2:
Automatically switches the default CRC Hack Level from Full in D3D to
Partial in OGL and vice versa.

Does not change CRC Hack Level when it is set to a non-default
setting.

### DIFF
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -23,7 +23,7 @@
 #include "GSState.h"
 #include "GSdx.h"
 
-int s_crc_hack_level = 3;
+int s_crc_hack_level = -1;
 
 // hacks
 #define Aggresive (s_crc_hack_level > 3)
@@ -2435,7 +2435,19 @@ void GSState::SetupCrcHack()
 {
 	GetSkipCount lut[CRC::TitleCount];
 
-	s_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+	GSRendererType renderer = static_cast<GSRendererType>(theApp.GetConfigI("Renderer"));
+	int selected_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+	if (selected_crc_hack_level == -1) {
+		if (renderer == GSRendererType::OGL_HW) {
+			printf("Automatic crc hack level set to Partial\n");
+			s_crc_hack_level = 2;
+		} else if (renderer == GSRendererType::DX9_HW || renderer == GSRendererType::DX1011_HW) {
+			printf("Automatic crc hack level set to Full\n");
+			s_crc_hack_level = 3;
+		}
+	} else {
+		s_crc_hack_level = selected_crc_hack_level;
+	}
 
 	memset(lut, 0, sizeof(lut));
 
--- a/plugins/GSdx/GSSetting.cpp
+++ b/plugins/GSdx/GSSetting.cpp
@@ -42,6 +42,7 @@ const char* dialog_message(int ID, bool* updateText) {
 				"Partial:\nEnable most of the CRC hacks.\nRecommended OpenGL setting (Accurate/depth options may be required).\n\n"
 				"Full:\nEnable all CRC hacks.\nRecommended Direct3D setting.\n\n"
 				"Aggressive:\nUse more aggressive CRC hacks. Only affects a few games, removing some effects which might make the image sharper/clearer.\n"
+				"Automatic:\nAutomatically select Partial or Full level based on wether you're using an OpenGL or Direct3D renderer.\n"
 				"Affected games: FFX, FFX2, FFXII, GOW2, ICO, SoTC, SSX3, SMT3, SMTDDS1, SMTDDS2.\n"
 				"Works as a speedhack for: Steambot Chronicles.";
 		case IDC_SKIPDRAWHACK:
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -23,8 +23,6 @@
 #include "GSState.h"
 #include "GSdx.h"
 
-extern int g_crc_hack_level;
-
 //#define Offset_ST  // Fixes Persona3 mini map alignment which is off even in software rendering
 
 int GSState::s_n = 0;
@@ -77,7 +75,18 @@ GSState::GSState()
 	//s_savel = 0;
 
 	UserHacks_WildHack = theApp.GetConfigB("UserHacks") ? theApp.GetConfigI("UserHacks_WildHack") : 0;
-	m_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+
+	GSRendererType renderer = static_cast<GSRendererType>(theApp.GetConfigI("Renderer"));
+	int selected_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+	if (selected_crc_hack_level == -1) {
+		if (renderer == GSRendererType::OGL_HW) {
+			m_crc_hack_level = 2;
+		} else if (renderer == GSRendererType::DX9_HW || renderer == GSRendererType::DX1011_HW) {
+			m_crc_hack_level = 3;
+		}
+	} else {
+		m_crc_hack_level = selected_crc_hack_level;
+	}
 
 	memset(&m_v, 0, sizeof(m_v));
 	memset(&m_vertex, 0, sizeof(m_vertex));
--- a/plugins/GSdx/GSTextureCache.cpp
+++ b/plugins/GSdx/GSTextureCache.cpp
@@ -48,7 +48,18 @@ GSTextureCache::GSTextureCache(GSRenderer* r)
 
 	m_paltex = theApp.GetConfigB("paltex");
 	m_can_convert_depth &= s_IS_OPENGL; // only supported by openGL so far
-	m_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+
+	GSRendererType renderer = static_cast<GSRendererType>(theApp.GetConfigI("Renderer"));
+	int selected_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+	if (selected_crc_hack_level == -1) {
+		if (renderer == GSRendererType::OGL_HW) {
+			m_crc_hack_level = 2;
+		} else if (renderer == GSRendererType::DX9_HW || renderer == GSRendererType::DX1011_HW) {
+			m_crc_hack_level = 3;
+		}
+	} else {
+		m_crc_hack_level = selected_crc_hack_level;
+	}
 
 	// In theory 4MB is enough but 9MB is safer for overflow (8MB
 	// isn't enough in custom resolution)
--- a/plugins/GSdx/GSdx.cpp
+++ b/plugins/GSdx/GSdx.cpp
@@ -211,6 +211,7 @@ void GSdxApp::Init()
 	m_gs_hack.push_back(GSSetting(1,  "Half", ""));
 	m_gs_hack.push_back(GSSetting(2,  "Full", ""));
 
+	m_gs_crc_level.push_back(GSSetting(-1, "Automatic", ""));
 	m_gs_crc_level.push_back(GSSetting(0 , "None", "Debug"));
 	m_gs_crc_level.push_back(GSSetting(1 , "Minimum", "Debug"));
 	m_gs_crc_level.push_back(GSSetting(2 , "Partial", "OpenGL Recommended"));
@@ -285,7 +286,7 @@ void GSdxApp::Init()
 	m_default_configuration["CaptureHeight"]                              = "480";
 	m_default_configuration["CaptureWidth"]                               = "640";
 	m_default_configuration["clut_load_before_draw"]                      = "0";
-	m_default_configuration["crc_hack_level"]                             = "3";
+	m_default_configuration["crc_hack_level"]                             = "-1";
 	m_default_configuration["CrcHacksExclusions"]                         = "";
 	m_default_configuration["debug_glsl_shader"]                          = "0";
 	m_default_configuration["debug_opengl"]                               = "0";


plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -320,6 +322,13 @@ void GSSettingsDlg::UpdateRenderers()
 	{
 		GSRendererType ini_renderer = GSRendererType(theApp.GetConfigI("Renderer"));
 		renderer_setting = (ini_renderer == GSRendererType::Undefined) ? GSUtil::GetBestRenderer() : ini_renderer;
+
```
FlatOutPS2: There's usually a space between other code and an if statement.

FlatOutPS2: I checked for whitespaces, so I really need to figure out what keeps adding 'em. :/

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -320,6 +322,13 @@ void GSSettingsDlg::UpdateRenderers()
 	{
 		GSRendererType ini_renderer = GSRendererType(theApp.GetConfigI("Renderer"));
 		renderer_setting = (ini_renderer == GSRendererType::Undefined) ? GSUtil::GetBestRenderer() : ini_renderer;
+		
+		if (ini_renderer == GSRendererType::Undefined) {
+			if (renderer_setting == GSRendererType::OGL_HW)
+				ComboBoxInit(IDC_CRC_LEVEL, theApp.m_gs_crc_level, 2);
```
FlatOutPS2: I had used SendMessge in an early draft of the commit, but it didn't work correctly. It must've been some mistake I made, because it seems to works perfectly now. :+1:

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -371,6 +380,30 @@ void GSSettingsDlg::UpdateControls()
...
 		bool ocl = renderer == GSRendererType::DX9_OpenCL || renderer == GSRendererType::DX1011_OpenCL || renderer == GSRendererType::OGL_OpenCL;
 		bool null = renderer == GSRendererType::Null;
+		
+		bool prev_dx_hw = prev_renderer == GSRendererType::DX9_HW || prev_renderer == GSRendererType::DX1011_HW;
+		bool prev_undef = prev_renderer == GSRendererType::Undefined;
+		int selected_CRC = SendMessage(GetDlgItem(m_hWnd, IDC_CRC_LEVEL), CB_GETCURSEL, 0, 0);
+
+		if (prev_undef)
+			prev_renderer = static_cast<GSRendererType>(i);
+		else if (ogl && hw && renderer != prev_renderer && prev_renderer != GSRendererType::OGL_HW) {
```
FlatOutPS2: Just leftovers from debugging. I've cleaned it up.

FlatOutPS2:
I think there's a bug in the Github app. It didn't pick up on a change I made. :/

plugins/GSdx/GSSettingsDlg.h
```diff
@@ -23,6 +23,7 @@
 
 #include "GSDialog.h"
 #include "GSSetting.h"
+#include "GSUtil.h"
```
turtleli: GS.h ? Since that's where the enum actually is.

plugins/GSdx/GSSettingsDlg.h
```diff
@@ -80,6 +81,9 @@ class GSSettingsDlg : public GSDialog
 	vector<GSSetting> m_ocl_devs;
 	uint32 m_lastValidMsaa; // used to revert to previous dialog value if the user changed to invalid one, or lesser one and canceled
 
+	GSRendererType prev_renderer = GSRendererType::Undefined;
+	int non_default_CRC = 0;
```
turtleli: Member variables should use the `m_` prefix.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -28,7 +28,6 @@
 #include "resource.h"
 #include "GSSetting.h"
 
-
```
turtleli: Maybe re-add the line back in so it looks like you haven't touched this area.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -372,6 +378,28 @@ void GSSettingsDlg::UpdateControls()
 		bool ocl = renderer == GSRendererType::DX9_OpenCL || renderer == GSRendererType::DX1011_OpenCL || renderer == GSRendererType::OGL_OpenCL;
 		bool null = renderer == GSRendererType::Null;
 
+		int selected_CRC = SendMessage(GetDlgItem(m_hWnd, IDC_CRC_LEVEL), CB_GETCURSEL, 0, 0);
+
+		if (prev_renderer == GSRendererType::Undefined)
+			prev_renderer = static_cast<GSRendererType>(i);
+		else if (renderer == GSRendererType::OGL_HW && prev_renderer != GSRendererType::OGL_HW) {
+			if (selected_CRC == 3 && non_default_CRC != 3) {
```
turtleli: Would making `non_default_CRC` a bool work? (I don't know whether it could be made to work, the current logic does make me think quite hard though.)

Side note: Just my personal opinion - if an `if` or `else` condition uses `{}`, then I think it looks nicer if the matching `else`/`if` also uses `{}`.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -372,6 +378,28 @@ void GSSettingsDlg::UpdateControls()
...
+				SendMessage(GetDlgItem(m_hWnd, IDC_CRC_LEVEL), CB_SETCURSEL, 2, 0);
+				non_default_CRC = 0;
+			} else if (selected_CRC == 2)
+				non_default_CRC = 2;
+
+			prev_renderer = static_cast<GSRendererType>(i);
+		} else if ((renderer == GSRendererType::DX9_HW || renderer == GSRendererType::DX1011_HW) && prev_renderer != GSRendererType::DX9_HW && prev_renderer != GSRendererType::DX1011_HW) {
+			if (selected_CRC == 2 && non_default_CRC != 2) {
```
ssakash: Most of the ``if()`` checks at this file have braces placed on a new line instead of placing them on the same line as the conditional checks, It would be better to have the placement of braces consistent on the file.

So better change,
```
if(condition){
}
```

**to**

```
if(condition)
{
}
```

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -320,6 +319,13 @@ void GSSettingsDlg::UpdateRenderers()
 	{
 		GSRendererType ini_renderer = GSRendererType(theApp.GetConfigI("Renderer"));
 		renderer_setting = (ini_renderer == GSRendererType::Undefined) ? GSUtil::GetBestRenderer() : ini_renderer;
+
+		if (ini_renderer == GSRendererType::Undefined) {
+			if (renderer_setting == GSRendererType::OGL_HW)
+				SendMessage(GetDlgItem(m_hWnd, IDC_CRC_LEVEL), CB_SETCURSEL, 2, 0);
+			else if (renderer_setting == GSRendererType::DX9_HW || renderer_setting == GSRendererType::DX1011_HW)
```
ssakash: Couldn't this just be ``else`` instead ? CRC hack level is only visible for the three hardware renderers (OpenGL, D3D9 , D3D11) .

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -320,6 +319,13 @@ void GSSettingsDlg::UpdateRenderers()
 	{
 		GSRendererType ini_renderer = GSRendererType(theApp.GetConfigI("Renderer"));
 		renderer_setting = (ini_renderer == GSRendererType::Undefined) ? GSUtil::GetBestRenderer() : ini_renderer;
```
ssakash: The else branch could probably be simplified into 

```
else
	{
		renderer_setting = (GSRendererType)theApp.GetConfigI("Renderer");

		if (renderer_setting == GSRendererType::Undefined)
		{
			renderer_setting = GSUtil::GetBestRenderer();

			if (renderer_setting == GSRendererType::OGL_HW)
				SendMessage(GetDlgItem(m_hWnd, IDC_CRC_LEVEL), CB_SETCURSEL, 2, 0);
			else
				SendMessage(GetDlgItem(m_hWnd, IDC_CRC_LEVEL), CB_SETCURSEL, 3, 0);
		}
	}
```

This avoids the need of an extra local variable ``ini_renderer``

plugins/GSdx/GSSettingsDlg.cpp
FlatOutPS2: renderer_setting is not only available for HW modes...

plugins/GSdx/GSSettingsDlg.h
FlatOutPS2: You're correct. I just copied the include that was used in GSSettingsDlg.cpp.

FlatOutPS2:
Had to move prev_renderer and non_default_CRC back to the cpp otherwise they would get reset each time you close the GSdx window.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -318,8 +320,17 @@ void GSSettingsDlg::UpdateRenderers()
...
-		GSRendererType ini_renderer = GSRendererType(theApp.GetConfigI("Renderer"));
-		renderer_setting = (ini_renderer == GSRendererType::Undefined) ? GSUtil::GetBestRenderer() : ini_renderer;
+		renderer_setting = GSRendererType(theApp.GetConfigI("Renderer"));
+
+		if (renderer_setting == GSRendererType::Undefined)
+		{
+			renderer_setting = GSUtil::GetBestRenderer();
+
+			if (renderer_setting == GSRendererType::OGL_HW)
+			{
```
ssakash: I would prefer if the redundant braces are removed at the ``if (renderer_setting == GSRendererType::OGL_HW)`` check. too much braces are not always nice :)

ssakash:
> Had to move prev_renderer and non_default_CRC back to the cpp otherwise they would get reset each time you close the GSdx window.

Maybe static variables ?

FlatOutPS2:
> Maybe static variables ?

Yes, but then they'd need to be declared and defined seperately. It would just add more code. I'm not sure if it's beneficial.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -372,6 +383,47 @@ void GSSettingsDlg::UpdateControls()
...
+				}
+				m_non_default_CRC = false;
+			}
+			else if (selected_CRC == 3)
+			{
+				m_non_default_CRC = true;
+			}
+
+			m_prev_renderer = static_cast<GSRendererType>(i);
+		}
```
ssakash: Maybe it would be worth having a dedicated function for these CRC Hack level management stuffs? (@turtleli - any opinion?)

I prefer separating them into a separate function because the code looks rather complicated in comparison to the other simpler and straightforward codes in `UpdateControls()``.

ssakash:
> Yes, but then they'd need to be declared and defined seperately. It would just add more code. I'm not sure if it's beneficial.

I don't mind the global variables. I had just mentioned an alternative since turtleli is quite picky on avoiding global variables ;)

turtleli:
Would a default/recommended CRC level be a better idea (#882)?

Current code is a bit hard for me to understand, and I'm not sure what the intended behaviour is.

Case 1: Starting at OpenGL HW -> Partial
 * Switch to D3D11 HW -> switches to full
 * Manually switch full to partial
 * Switch to OpenGL HW -> stays at partial (expected)
 * Manually switch partial to full
 * Manually switch full to partial
 * Switch to D3D11 HW -> stays at partial

Case 2: Starting at OpenGL HW -> Partial
 * Switch to D3D11 HW -> switches to full
 * Manually switch full to minimum
 * Switch to OpenGL HW -> stays at minimum (expected)
 * Manually switch minimum to partial
 * Switch to D3D11 HW -> switches to full

FlatOutPS2:
> Switch to D3D11 HW -> switches to full
    Manually switch full to partial
    Switch to OpenGL HW -> stays at partial (expected)
    Manually switch partial to full
    Manually switch full to partial
    Switch to D3D11 HW -> stays at partial

The intended behaviour is that it switches to Partial or Full if the default CRC levels(Partial for OGL, Full for D3D) are selected. If a custom change is made(Full on OGL), it shouldn't switch back to Partial if you go to D3D11 and back to OGL HW without making changes to the CRC level again. But if you set it to Partial on OGL HW manually it should go back to full on D3D11 HW, so there's a bug in the code. I know what it is, so I'll go and fix it. ;)

My PR is better than the manual behaviour because it automatically switches between the default levels for OGL and D3D, which most users won't do out of themselves, even though it's often needed for CRC fixes.

willkuer:
I actually wanted to finish the old pr soon. Currently writing my phd thesis and don't have time though.

For obvious reasons I also rather like the idea of an explicit automatic setting. Just changing settings in the background without notification sounds not so good. But in the end you guys decide that wether is a better implementation.

FlatOutPS2:
I've updated my PR to add an Automatic option instead. Tell me if you like this better.

willkuer:
I'd like that much more. However there are two problems I also tried to tackle myself following this approach.

First the renderer selection should be refractored into a single function that is called by twice in the appropriate place. However when I was looking at the problem these two places only shared gs.h . That's why I was putting the function into the header which is a bit sloppy. IIRC Avih however agreed that it is ok in this situation.

Another problem is that the hack level screams for an enum. Especially if you include -1 (automatic). However simply changing to enum doesn't work as the application of the level uses comparison operators. So either you cast to int at that place which will break if somebody changes the ordering in the enum or you need to reimplement the crc hack application. I started that by using a flag enum:
0x1 => implemented in ogl (iogl)
0x2 => implemented in dx (idx)
iogl & idx => fully implemented (crc none)
iogl  => not implemented in dx (partial)
0x0 => not implemented at all (full)
...

However this approach needs a lot of refractoring with almost no further improvements. It would be easier to extend to new renderers but as gregory said he'll only support one and drop all others. I originally wanted to design it lkke that in case we have ogl3.3 and ogl4.5 renderer. Maybe if vulcan or dx12 will be implemented. Then the linear hack evel doesn't work anymore. But maybe one tackles this problems when they occur.

I'd suggest to use an enum and mention in the enum that ordering is important and then cast to int in the crc hack application code.

willkuer:
If you use -1 for automatic I'd also suggest to assert(crc_level>=0) in the crc consumer code just to be sure that all branches are handled.

gregory38:
Otherwise just add a new value. Partial in opengl_and_full_dx. Just check the renderer when the hack are initialized. However it would be a good idea to check the crc hack level with a common function.

plugins/GSdx/GSState.cpp
```diff
@@ -77,7 +75,18 @@ GSState::GSState()
...
+	GSRendererType renderer = static_cast<GSRendererType>(theApp.GetConfigI("Renderer"));
+	int selected_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+	if (selected_crc_hack_level == -1) {
+		if (renderer == GSRendererType::OGL_HW) {
+			m_crc_hack_level = 2;
+		} else if (renderer == GSRendererType::DX9_HW || renderer == GSRendererType::DX1011_HW) {
+			m_crc_hack_level = 3;
+		}
+	} else {
+		m_crc_hack_level = selected_crc_hack_level;
```
ssakash: Maybe 
```
if (selected_crc_hack_level == -1)
{
    if (renderer == GSRendererType::OGL_HW) {
			m_crc_hack_level = 2;
	} else if (renderer == GSRendererType::DX9_HW || renderer == GSRendererType::DX1011_HW) {
		m_crc_hack_level = 3;
  	}
} else {
		m_crc_hack_level = selected_crc_hack_level;
}
```

**could probably be changed to**

```
if (selected_crc_hack_level == -1)
{
    if (renderer == GSRendererType::OGL_HW)
			m_crc_hack_level = 2;
	else if (renderer == GSRendererType::DX9_HW || renderer == GSRendererType::DX1011_HW)
		    m_crc_hack_level = 3;
 }
else
{
		m_crc_hack_level = selected_crc_hack_level;
}
```

I might be biased but the latter looks much nicer and easier to read, any opinion? If you want to change it to the latter then make sure to also do it for the other files where you had used the previous format.

plugins/GSdx/GSSetting.cpp
```diff
@@ -42,6 +42,7 @@ const char* dialog_message(int ID, bool* updateText) {
 				"Partial:\nEnable most of the CRC hacks.\nRecommended OpenGL setting (Accurate/depth options may be required).\n\n"
 				"Full:\nEnable all CRC hacks.\nRecommended Direct3D setting.\n\n"
 				"Aggressive:\nUse more aggressive CRC hacks. Only affects a few games, removing some effects which might make the image sharper/clearer.\n"
+				"Automatic:\nAutomatically select Partial or Full level based on wether you're using an OpenGL or Direct3D renderer.\n"
```
ssakash: ``wether`` ***->*** ``whether``

plugins/GSdx/GSHwHack.cpp
```diff
@@ -2435,7 +2435,19 @@ void GSState::SetupCrcHack()
 {
 	GetSkipCount lut[CRC::TitleCount];
 
-	s_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+	GSRendererType renderer = static_cast<GSRendererType>(theApp.GetConfigI("Renderer"));
+	int selected_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+	if (selected_crc_hack_level == -1) {
+		if (renderer == GSRendererType::OGL_HW) {
+			printf("Automatic crc hack level set to Partial\n");
```
ssakash: ``crc`` ***->*** ``CRC``. Same for the ``printf()`` below.

jcdenton2k:
Kudos for maintaining code legibility.  I enjoy the updated/cleaner look you presented.

gregory38:

FlatOutPS2:
It's too much of a clusterfuck for just a simple UI experience enhancement. I can't really see this working out soon with how I'd like it to be and also adhering to all the requested changes to the code, so I'll close this.

ssakash:
>  and also  adhering to all the requested changes to the code

Well, the only changes I explicitly requested were the typo fixes. The code beautify stuff was optional, sorry if my nitpicks were the reason on closure of the PR.

FlatOutPS2:
I didn't mean your requests, I meant the ones before that.

PR #1587, opened 2016-09-24
gsdx OGL:  add CRC hack for Ratchet & Clank to use Jak shadow shader

Dreadmoth:
* 0x76F724A3, RatchetAndClank, EU
* 0x17125698, RatchetAndClank3, EU
* 0xD697D204, RatchetGladiator, EU

Ratchet & Clank games use the same shadow technique as the Jak and Daxter series, and benefit from the alternate shader added for them in c2b67cc .
(Ratchet & Clank 2 not added as my PC can't read the disc currently - can't test it)

No hacks: before
Auto Flush hack: autoflush
Shader: after

(sorry for broken texture screenshots gregory)

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -521,6 +521,9 @@ CRC::Game CRC::m_games[] =
 	{0x12804727, Jak3, EU, 0},
 	{0x644CFD03, Jak3, US, 0},
 	{0xDF659E77, JakX, EU, 0},
+	{0x76F724A3, RatchetAndClank, EU, 0},
+	{0x17125698, RatchetAndClank3, EU, 0},
+	{0xD697D204, RatchetGladiator, EU, 0},
 };
 
 map<uint32, CRC::Game*> CRC::m_map;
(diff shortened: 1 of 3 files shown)


gregory38:
Oh I didn't know that it impacts others game. Let's me check if we can make somethings more automatic instead.

FlatOutPS2:
J&D and R&C engine share common features, they are both Sony family series. :)

gregory38:
Hum. I need to check but maybe checking that fbmask is 0x00FFFFFF could be enough. I need to think about it.

gregory38:
@Dreadmoth could you give a test on #1588 hopefully it will work without any CRC.

Dreadmoth:
>@Dreadmoth could you give a test on #1588 hopefully it will work without any CRC.

Works well in Ratchet & Clank 1 - 3, Ratchet: Gladiator, Jak 3, Jak X :+1: 
Tried a few other games (BG: Dark Alliance, Drakan, Metal Arms, Transformers, Tribes) and didn't notice any new problems.
(off-topic: I tried my scratched-up R&C2 disc in another PC and ImgBurn created an image successfully, guess newer DVD drives have better error correction)

gregory38:
perfect. Thanks you :+1:

PR #1590, opened 2016-09-25
cdvdgigaherz: Read raw CD sectors one by one

turtleli:
A multi sector raw disk sector read that reads data from two tracks of different types will not complete successfully. Reading the sectors one at a time should fix the issue.

@MrCK1 Could you test the PS1 game(s) that previously had read errors?

### DIFF
--- a/plugins/cdvdGigaherz/src/Windows/IOCtlSrc.cpp
+++ b/plugins/cdvdGigaherz/src/Windows/IOCtlSrc.cpp
@@ -352,12 +352,8 @@ s32 IOCtlSrc::ReadSectors2048(u32 sector, u32 count, char *buffer)
 }
 
 
-// FIXME: Probably doesn't work if the sectors to be read are from two tracks
-// of different types.
 s32 IOCtlSrc::ReadSectors2352(u32 sector, u32 count, char *buffer)
 {
-    RAW_READ_INFO rri;
-
     DWORD size = 0;
 
     if (!OpenOK)
@@ -374,13 +370,10 @@ s32 IOCtlSrc::ReadSectors2352(u32 sector, u32 count, char *buffer)
         sptd.info.Cdb[0] = 0xBE;
         // Don't care about sector type.
         sptd.info.Cdb[1] = 0;
-        sptd.info.Cdb[2] = (sector >> 24) & 0xFF;
-        sptd.info.Cdb[3] = (sector >> 16) & 0xFF;
-        sptd.info.Cdb[4] = (sector >> 8) & 0xFF;
-        sptd.info.Cdb[5] = sector & 0xFF;
-        sptd.info.Cdb[6] = (count >> 16) & 0xFF;
-        sptd.info.Cdb[7] = (count >> 8) & 0xFF;
-        sptd.info.Cdb[8] = count & 0xFF;
+        // Number of sectors to read
+        sptd.info.Cdb[6] = 0;
+        sptd.info.Cdb[7] = 0;
+        sptd.info.Cdb[8] = 1;
         // Sync + all headers + user data + EDC/ECC. Excludes C2 + subchannel
         sptd.info.Cdb[9] = 0xF8;
         sptd.info.Cdb[10] = 0;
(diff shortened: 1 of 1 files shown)


MrCK1:
Will do when I get a chance

zuzma:
I'm probably wrong since I was just looking at the redump wiki and wikipedia, but it looks like it needs channel p for PlayStation games. Dunno if that would be the reason, unless i'm looking in the wrong spot. It also doesn't look like libcrypt games would work either

turtleli:
Hmm. Code should be less likely to have read issues, though I might have misdiagnosed the problem. I probably need to add in some debug code, though I'm not sure I'll find anything out from doing so.

I do note that redump says that the first track has 31528 sectors but the log says it has 31678 sectors.

zuzma:
Well it sounds like it's working properly then if it's adding 150 sectors. It was just a shoot in the dark really

turtleli:
I added debug stuff to the commit. It won't fix the issue, but it might give me a better idea of what's failing (it may also completely confuse me too).

@MrCK1 Could you provide another log of the same game? Thanks.

MrCK1:
@turtleli [Here's another one](http://pastebin.com/62iCHZjH)

turtleli:
The log confuses me since the sense buffer is all 0. I don't know whether the read is succeeding but just not returning 2352 bytes. Commit has been updated again, so it needs another test.

ramapcsx2:
This pr works better than master with my Battle Arena Toshinden 2 (PAL) CD.
With master, it just crashes shortly after the PSX logo. With this version, it gets ingame but can't load battles.

FlatOutPS2:
> This pr works better than master with my Battle Arena Toshinden 2 (PAL) CD.
With master, it just crashes shortly after the PSX logo. With this version, it gets ingame but can't load battles.

Does the ISO load the battles?

ramapcsx2:
Yep, the PCSX2 internal iso reader works fine with an image of this disc.

ramapcsx2:
Hm, the battles not loading could be a core timing thing.
When I read data off a real disc, there's additional delays (just like on the real console).
Those could affect compatibility.
It'd be great to test this more reliably with the delays added to the internal reader.
Would adding forced spinwaits work?

willkuer:
But if the internal loader works why should delay be a problem? Less delay than by using iso's barely should be possible..

ramapcsx2:
I meant that the extra delays could have a positive effect. That's why I also want to try adding them to the internal loader.
In effect, this would produce more "sector not ready yet" situations when the core asks for new data.
The PSX is as sensitive to this as is the PS2, maybe even more so.
This would purely be for testing of course. It doesn't have to be fancy as it wouldn't stay for long.

FlatOutPS2:
The latest master with the extra delay improves a random demo disc I tried. With the previous commit it only loaded an intro video when booting from the disc, now it loads the main menu just like the ISO.

ramapcsx2:
Nice to know. The other PSX DMA channels require timing as well.

turtleli:
So are there any disc read errors with this PR (from the actual disc reading itself, not those related to emulation issues)?

Oh, and if you see a ` * CDVD: SPTI short transfer of x bytes` message then I'd be interested in seeing the emulog too.

ramapcsx2:
I didn't have time to test this in-depth. But from what I saw, it is an improvement for compatibility.
Performance might drop though? Not sure.

turtleli:
@MrCK1 Retest IRHA Drag Racing?

MrCK1:
@turtleli I'll test IHRA later tonight, I don't have my USB CD/DVD drive with me

MrCK1:
So I've actually discovered another game that still has this issue. I didn't spot it before because the game used to crash until SPU2 was fixed.

The game is Rascal (absolutely horrible game :p)

I'll post the log when I get to my laptop later

lightningterror:
https://github.com/PCSX2/pcsx2/projects/2
Can these Projects in progress be closed since the issues are resolved ?

MrCK1:
@lightningterror PSX mode is moving at a slow progress. The projects allow us to keep track of what changes have been made in the past.

lightningterror:
Ah okay , I thought In Progress topics are closed once they are resolved so nvm.

FlatOutPS2:
> Ah okay , I thought In Progress topics are closed once they are resolved so nvm.

The topics could be moved to a "Finished" column. Keeping 'em in the "In Progress" section does indicate they're still in development.

Issue #1025, opened 2015-12-08
D3D11's "Bilinear (PS2)" Texture Filtering option is not filtering some textures that the PS2 filters


karasuhebi:
The "Bilinear (PS2)" Texture Filtering option in the D3D11 renderer is not filtering certain textures that the PS2 is, while the D3D9 and OGL renderers do. If you set the D3D11 renderer to "Bilinear (Forced)" though it does filter the textures, so it's obviously capable of it, it's just not doing it on "Bilinear (PS2)" for some reason. Here's some screenshot examples, look at the bottom-left text and icon:
D3D11, Bilinear (PS2):
![D3D11_PS2](http://i.imgur.com/etZBbTh.png)

OGL, Bilinear (PS2):
![OGL_PS2](http://i.imgur.com/SvKorUz.png)

D3D9, Bilinear (PS2):
![D3D9_PS2](http://i.imgur.com/no1YnYj.png)

D3D11, Bilinear (Forced):
![D3D11_Forced](http://i.imgur.com/5KQ2CZS.png)

This issue can be seen as far back as at least 1.2.1.

gregory38:
Hum, it could be related to an ATI hack. Otherwise code is very close between renderer

karasuhebi:
I have an AMD GPU so that seems like a reasonable assumption to make. Is there anything I can do to help confirm that this is the case?

karasuhebi:
bump.

DPersonalized:
I too have an AMD GPU and this is also happening to me, but in the game I tested, **almost nothing** is being filtered!:

Hardware
![hardware_b3_filter](https://cloud.githubusercontent.com/assets/8225831/12517507/ed22d51a-c12a-11e5-9485-ca2aa4bd6bbb.png)

Software
![software_b3_filter](https://cloud.githubusercontent.com/assets/8225831/12517535/0c4bb34e-c12b-11e5-8b9a-f35bbbdcc35e.png)

The most obvious difference here is in the boxes outlined in blue to the right, which feature sharp steps in color instead of having a smooth transition between the different shades of blue.

willkuer:
For you the sane regarding dx9/11/ogl?

DPersonalized:
Yup. DX9 and OGL are unaffected.

sudonim1:
Could you please provide proof that the PS2 is filtering those same textures? The (PS2) option means that it filters when the GS is set to do bilinear filtering and not otherwise.

[...]

PR #1593, opened 2016-09-27
gsdx hw ate

gregory38:
I shuffle a bit the ATE code in order to improve performance. It will avoid various depth look up when the buffer isn't used.

An old TEXA hack was removed too due to various bad interactions.
```
It would be nice to test those games
* Ar Tonelico 2 (line in sprite regression?)
* breath of fire dragon quarter (overlayed user interface in the game)
```

### DIFF
--- a/plugins/GSdx/GSRendererDX.cpp
+++ b/plugins/GSdx/GSRendererDX.cpp
@@ -474,8 +474,12 @@ void GSRendererDX::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sourc
 	{
 		const GSLocalMemory::psm_t &psm = GSLocalMemory::m_psm[m_context->TEX0.PSM];
 		const GSLocalMemory::psm_t &cpsm = psm.pal > 0 ? GSLocalMemory::m_psm[m_context->TEX0.CPSM] : psm;
+		// The texture cache will handle various format conversion internally for non-target texture
+		// After the conversion the texture will be RGBA8 (aka 32 bits) hence the 0 below
+		int gpu_tex_fmt = (tex->m_target) ? cpsm.fmt : 0;
+
 		bool bilinear = m_filter == 2 ? m_vt.IsLinear() : m_filter != 0;
-		bool simple_sample = !tex->m_palette && cpsm.fmt == 0 && m_context->CLAMP.WMS < 3 && m_context->CLAMP.WMT < 3;
+		bool simple_sample = !tex->m_palette && gpu_tex_fmt == 0 && m_context->CLAMP.WMS < 3 && m_context->CLAMP.WMT < 3;
 		// Don't force extra filtering on sprite (it creates various upscaling issue)
 		bilinear &= !((m_vt.m_primclass == GS_SPRITE_CLASS) && m_userhacks_round_sprite_offset && !m_vt.IsLinear());
 
@@ -484,7 +488,7 @@ void GSRendererDX::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sourc
 		if (ps_sel.shuffle) {
 			ps_sel.fmt = 0;
 		} else {
-			ps_sel.fmt = tex->m_palette ? cpsm.fmt | 4 : cpsm.fmt;
+			ps_sel.fmt = tex->m_palette ? gpu_tex_fmt | 4 : gpu_tex_fmt;
 		}
 		ps_sel.aem = m_env.TEXA.AEM;
 		ps_sel.tfx = m_context->TEX0.TFX;
--- a/plugins/GSdx/GSRendererDX.cpp
+++ b/plugins/GSdx/GSRendererDX.cpp
@@ -479,7 +479,7 @@ void GSRendererDX::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sourc
 		int gpu_tex_fmt = (tex->m_target) ? cpsm.fmt : 0;
 
 		bool bilinear = m_filter == 2 ? m_vt.IsLinear() : m_filter != 0;
-		bool simple_sample = !tex->m_palette && gpu_tex_fmt == 0 && m_context->CLAMP.WMS < 3 && m_context->CLAMP.WMT < 3;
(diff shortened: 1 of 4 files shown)


refractionpcsx2:
Game tests should be done on DX also, as OGL behaviour hasn't changed much.

gregory38:
Oh yes, you're right. Sorry I forgot this information.

FlatOutPS2:
Menu text disappears in Breath Of Fire DQ with D3D renderers.

gregory38:
Is it fine with openGL ?

FlatOutPS2:
Yes, it's still fine with OGL.

gregory38:
So I did really fix some stuff for openGL. I was hoping it was limited to the texture cache code, but not enabled because it wasn't tested.... I guess I need to redo my PR to add back the hack for Dx.

FlatOutPS2:
Can't the Dx side be improved as well?

gregory38:
```
Can't the Dx side be improved as well?
```
In theory yes, if you're volunteer. I would need to check the git history but here the history.

GS textures are converted to 32 bits RGBA by the CPU. Palette is resolved and alpha is extended (with TEXA info). Everything is fine.
When texture is an old target (m_target) rendered by the GPU, the conversion isn't done, it must be done by the GPU on the fly. Both palette and TEXA. I think Dx doesn't set the shader flag to do TEXA the conversion. Or TEXA is wrongly done on the shader.

gregory38:
Dx
```
ps_sel.aem = m_env.TEXA.AEM;
```
OpenGL => much complicated :stuck_out_tongue:

gregory38:
I pushed the 2 firsts commits in master. I kept here the bad TEXA stuff. 

@FlatOutPS2 
I think I see the Dx bug

Here the shader code that will do the TEXA translation
```	
[unroll]
	for (uint i = 0; i < 4; i++)
	{
		if((PS_FMT & ~FMT_PAL) == FMT_32)
		{
			#if SHADER_MODEL <= 0x300
			if(PS_RT) c[i].a *= 128.0f / 255;
			#endif
		}
		else if((PS_FMT & ~FMT_PAL) == FMT_24)
		{
			c[i].a = !PS_AEM || any(c[i].rgb) ? TA.x : 0;
		}
		else if((PS_FMT & ~FMT_PAL) == FMT_16)
		{
			c[i].a = c[i].a >= 0.5 ? TA.y : !PS_AEM || any(c[i].rgb) ? TA.x : 0; 
		}
	}
```
As you can see it will be done if color format is 24/16 bits.

Looking at the Dx renderer,  format is set as
```
const GSLocalMemory::psm_t &psm = GSLocalMemory::m_psm[m_context->TEX0.PSM];
const GSLocalMemory::psm_t &cpsm = psm.pal > 0 ? GSLocalMemory::m_psm[m_context->TEX0.CPSM] : psm;
	
	if (ps_sel.shuffle) {
			ps_sel.fmt = 0;
		} else {
			ps_sel.fmt = tex->m_palette ? cpsm.fmt | 4 : cpsm.fmt;
		}
```
So Dx always sets the texture format from the context.

However the texture cache will convert 24/16 bit texture to 32 bits format. Please note that texture cache can only do it for textures that aren't older frame buffer.
So, you must do something like that
```
if (tex->m_target) {
  real_tex_fmt = cpsm.fmt; // we need to handle alpha extension in the GPU
} else {
 real_tex_fmt = 0; // The texture cache already did the alpha extension. 
}
```
And then use real_tex_fmt instead of cpsm.fmt.

Do you have any questions ?

FlatOutPS2:
No questions needed. The example fix you made works perfectly as is. :p

Declaring `real_tex_fmt` and using real_tex_fmt instead of cpsm.fmt is all you need to add.

gregory38:
@FlatOutPS2 ok. I updated the commit. I also updated the filtering stuff, bilinear will be enabled more often. Could be related to some pending issue. Could you confirm me that I didn't screw it up.

FlatOutPS2:
There is a small regression in the letters. The lines under some letters are several pixels thicker than they're supposed to be with the  DX renderers. This doesn't happen with my DX fix custom build of the previous version of the PR.

gregory38:
It is due to this change
```
-		bool simple_sample = !tex->m_palette && cpsm.fmt == 0 && m_context->CLAMP.WMS < 3 && m_context->CLAMP.WMT < 3;
+		bool simple_sample = !tex->m_palette && gpu_tex_fmt == 0 && m_context->CLAMP.WMS < 3 && m_context->CLAMP.WMT < 3;
```
What does openGL ? Which filter option did you set ? SW ?

FlatOutPS2:
It's DX and Hardware only. I used default filter option.

gregory38:
Hum, could you do a quick check with this line instead.
```
bool simple_sample = !tex->m_palette && gpu_tex_fmt == 0 && m_context->CLAMP.WMS < 2 && m_context->CLAMP.WMT < 2;
```

FlatOutPS2:
That fixed it. :)

gregory38:
However it would create others regressions from user point of view.

The full story is that you can sample texture with the HW unit or emulate it with shader.
`simple_sample` is true when HW unit can be supported. 

Technically current code is wrong (when WMS/WMT are 2) but it is good most of time. Mode 3 is a clamp to a sub region of the texture.
When `simple_sample` is false.
* no aniso (well actually maybe Dx would still wrongly activate it)
* no forced bilinear

However maybe it could fix this issue #1025 ? Any preference ? Your fix, or the complete fix ?

FlatOutPS2:
Yes, WMS/WMT = 2 fixes the FFXII issue as well. Wouldn't the correct code that also fixes two issues be better than keeping using 3 for DX?

I also know of two games(Samurai Shodown V, The Mark Of Kri) that have no PS2 level bilinear filtering with DX11 when you boot, and it's only enabled after you pause and resume the emulator. DX9 and OGL are also fine in those games. Could it be related?

gregory38:
It could be. Well it is strange that it is fine for dx9. But I guess the region clamping is emulated in a different way. I will update the pr.

gregory38:
ok, could we merge the code then.

PR #1595, opened 2016-09-27
GSDX: Add a minor optimization to software renderer

ssakash:
**Summary of changes**:

* Add an optimization which disables depth test whenever it's not necessary. (suggested by @gregory38)

@FlatOutPS2 

Could you check the performance impact on some games? I don't have any games which trigger the condition so couldn't properly check it.

### DIFF
--- a/plugins/GSdx/GSRendererSW.cpp
+++ b/plugins/GSdx/GSRendererSW.cpp
@@ -1001,6 +1001,28 @@ bool GSRendererSW::CheckSourcePages(SharedData* sd)
 
 #include "GSTextureSW.h"
 
+// The idea is to disable Z test at scenarios whenever it's not necessary.
+// There is a certain fast path for the code to take when all the pixel tests are disabled
+// at such cases, disabling the Z test should be helpful on providing a bit of increase in performance.
+// the function below identifies such cases where it would be useful to omit the Z test.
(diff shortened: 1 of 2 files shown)


FlatOutPS2:
The code looks much slicker as my implementation, but the result is the same; there's no performance impact on the Burnout games.

gregory38:
Hum, maybe it is bad optimization. If it is only used for letter rendering, it would be only few draw call on small screen portion. Could you test ICO too ?

FlatOutPS2:
I don't think ICO triggers it. I tried it with my own implementation yesterday.

gregory38:
I think I saw it in your sea testcase,

ssakash:
> there's no performance impact on the Burnout games.

Does it have any other pixel tests enabled? I think Persona 3 triggered the condition.

FlatOutPS2:
> I think I saw it in your sea testcase,

No difference at that location either.

@ssakash I believe so, but I didn't use a debug build so I'm not sure.

gregory38:
Sorry. In the end, I'm afraid it isn't a good idea.

ssakash:
FWIW, I also tested a dump on burnout where the optimization helps to enter the fast path but still there was barely any difference. So yeah, closing this. ;)

PR #1600, opened 2016-09-30
Update GSCrc.cpp

puzzo888:


ssakash:
A more descriptive commit message might be useful, other than that LGTM 👍

FlatOutPS2:
Do we add iso hacked (translation)CRC's to the database?

danilaml:
Now it just changes tabs to spaces on one line and nothing else.

gregory38:
So I disagree to add tons of hacked iso in the code. Otherwise, the PR is now empty so I will close it.

PR #1601, opened 2016-09-30
Greg/gsdx mipmap

gregory38:
A step to bring full mipmap. All layers are uploaded to the GPU. Remain to compute lod for the filtering on the GPU. But first I want to validate the speed impact of all those textures transfer.

To enable new code
```
UserHacks_mipmap = 2
mipmap = 2
```

There will be no graphical improvement with previous code only a perf impact that shall be benchmarked.

Edit: openGL only


FlatOutPS2:
`mipmap = 2` ??

This will automatically get reset to 1 when you enter and exit the GSdx plugin settings menu.

gregory38:
Oh, don't use the gui then :)

FlatOutPS2:
Yes, I know. But it's annoying. :p

gregory38:
Yeah. Fwiw, there are some bugs that seemd to generate some bad texture parameter.

gregory38:
Edit: maybe we can add a new mipmap combo box.

gregory38:
I need an nvidia tester too.

MrCK1:
@FlatOutPS2 I'm assuming you were testing the 1st R&C? You have an AMD gpu too right? 😛 

My testing in R&C: Going Commando saw no speed impact. Previous testing from a few days ago with the new mipmap implementation saw no issues either.

Dreadmoth:
>I need an nvidia tester too.

Unless I'm doing something wrong, there seems to be no performance impact in the areas I'm testing in R&C (Blackwater City landing pad), R&C3 (Starship Phoenix landing pad) and Jak 3 (Spargus City near the ocean).

gregory38:
Be sure both values are 2 in the ini. Maybe check memory usage. More textures require more vrams.

FlatOutPS2:
You need to look for locations with many mipmapping textures. In Ace Combat the speed dropped most when you look at the horizon where the flickering textures where before.

Dreadmoth:
On the Metropolis overlook in R&C (50Hz, 5x Native, High blending accuracy):
mipmap=1 -> 58fps, 437MB VRAM usage
mipmap=2 -> 56fps, 477MB VRAM usage
Disabling mipmapping causes a ~10fps drop in the same position.

gregory38:
Hum I need to optimize layers upload based on max lod. I think it could reduce the speed impact.

gregory38:
@FlatOutPS2 @Dreadmoth could you test the last commit. I try to reduce the number of texture upload on the GPU.

Dreadmoth:
Metropolis overlook in R&C:
mipmap=1 -> 58fps, 409MB VRAM usage
mipmap=2 -> 58fps, 413MB VRAM usage

Maktar Resort landing pad in R&C2:
mipmap=1 -> 63fps, 415MB VRAM usage
mipmap=2 -> 62fps, 422MB VRAM usage

gregory38:
Seem to be better. I'm curious if it could work to generate pseudo mipmap from the base layer. Anyway I guess now I need to compute the lod on shader

gregory38:
Sorry my code was wrong. Need a test again.

FlatOutPS2:
> Metropolis overlook in R&C:
mipmap=1 -> 58fps, 409MB VRAM usage
mipmap=2 -> 58fps, 413MB VRAM usage

Metropolis overlook in R&C (with latest bad typo fix commit):
mipmap=1 -> 63%, 391MB VRAM usage
mipmap=2 -> 44%, 397MB VRAM usage

Yes, OGL speeds are that bad in this game. :p And that's still a 30% drop betwen mipmap 1 and 2.

gregory38:
Of course there is another typo. Anyway,
jak_sea

Here, lod mipmap value is computed automatically. I only load the PS2 mipmap not all the mipmap of the texture hence the black stuff. However, there are some way to generate all those mipmap automatically. Maybe it could be faster (or slower), however it would be less accurate.

gregory38:
Nice progress :stuck_out_tongue: 
dark_jak

gregory38:
Arg, so I finally implemented a full mipmap correctly (with minor limitation)
full_mipmap

Now, you will tell me that it isn't really better. The front of the sea is still bad. However it isn't a mipmap issue per se. What happen, is that GS will do various rendering to draw the various layers of the texture. Current (full) mipmap implementation expects to read texel data from the GS memory. Here, it would require to search data on the GPU too (and to convert it too).

Dreadmoth:
Looks like nice progress in Jak and Daxter :+1: 

-
Setting mipmap2 now seems to cause broken distortion effects in Jak 3:

Jak 3 broken distortion effect
[GS dump](https://github.com/PCSX2/pcsx2/files/504655/Jak3-Distortion-Mipmap2.zip)

-
Something is causing aliasing when using upscaled resolutions in Ratchet & Clank:
[1862 2x Native](http://i.imgur.com/aQEE1Mz.png) | [GS dump](https://github.com/PCSX2/pcsx2/files/504693/Ratchet-Aliasing.zip)
[1856 2x Native](http://i.imgur.com/Ytv7KId.png)
This doesn't seem to be affected by the mipmapping settings.

-
@FlatOutPS2 [this](http://i.imgur.com/966Lgwq.jpg) is the place I was using. After landing I just moved forward (without moving the camera) until Ratchet's foot was on the centre of the platform. If I go all the way forward to the railing, it drops to 53fps - but performance is still the same with each setting. 
If I turn the camera around immediately after landing I get ~103fps. (performance numbers from 1856, need to re-test with current)

FlatOutPS2:
> @FlatOutPS2 this is the place I was using. After landing I just moved forward (without moving the camera) until Ratchet's foot was on the centre of the platform. If I go all the way forward to the railing, it drops to 53fps - but performance is still the same with each setting.
If I turn the camera around immediately after landing I get ~103fps. (performance numbers from 1856, need to re-test with current)

That's where I was as well, though I was all the way forward at the railing. I think I got something like 90 fps after turning around after landing, I didn't pay much attention to the actual fps, I just noticed it went from running really slow to really fast. :)

gregory38:
I'm not sure that mipmap 2 will ever work. Anyway, I pushed a tons of change. You have now a new extended filtering option. Mipmap must be either 1 or 0. Tell me if it helps to reduce shimmering.

FlatOutPS2:
I tried FIFA with the trilinear option and it works great for the textures in the distance :+1: 

Note: why is it called `Trilinear `instead of `Trilinear (PS2)`?

gregory38:
Because `PS2` is different. This "trilinear" mode will generate the mipmap layer based on the first layer. And then it will use the automatic lod computed by the GPU.
1/ GS can have different layer texture. For example APE that use them to emulate a light effect
2/ GS Lod is computed based on the texture Q + constant parameter L/K. Whereas GPU will use derivative of texture coordinate/position.

If the game uses a `normal` trilinear mipmap to smooth texture in distance, the trilinear tricks will be close of it. By the way, is the perf good or bad ?

Dreadmoth:
Nice results in Tribes and Drakan - effectively eliminated texture aliasing / shimmering on nearby terrain and structures. Can combine with anisotropic filtering to improve detail.
Some improvements in Jak 3.
No effect on Snowblind Engine games (aside from large performance hit) - but they don't seem to suffer much texture aliasing anyway.

I'll check on performance a bit more later - one scene in Tribes (4x Native, full blending):
Bilinear (forced): ~260fps
Trilinear: ~230fps
Trilinear (forced): ~150fps
Trilinear (forced) + 16x AF: ~150fps

-
Is anyone else seeing the upscaling issue I mentioned in the Ratchet & Clank games?

Possibly related (and making the issue worse in motion) - OpenGL seems to be scaling PCSX2's output to the display/window using nearest-neighbour interpolation. D3D9 and D3D11 (and OpenGL on earlier commits for this PR and master) seem to use bilinear or bicubic.

plugins/GSdx/GSTextureOGL.cpp
```diff
@@ -249,6 +249,8 @@ GSTextureOGL::GSTextureOGL(int type, int w, int h, int format, GLuint fbo_read)
 		throw GSDXErrorOOM();
 	}
 
+	m_max_layer = (m_type == GSTexture::Texture) && mipmap ? (int)log2(max(w,h)) : 1;
```
gregory38: Can only be supported with RGBA8

gregory38:
@Dreadmoth Thanks for the test. Jak is still suffering of various texture cache issue. Is trilinear forced really better ? So far trilinear forced, just force bilinear + trilinear on all draws. Looking at the speed hit, maybe I could limit it to force bilinear on all draws + trilinear on the requested draws.

```
Possibly related (and making the issue worse in motion) - OpenGL seems to be scaling PCSX2's output to the display/window using nearest-neighbour interpolation. D3D9 and D3D11 (and OpenGL on earlier commits for this PR and master) seem to use bilinear or bicubic.
```
Just to be sure, delete your GSdx.ini file and retry.

FlatOutPS2:
There seems to be a 20% drop with both Trilinear and Trilinear Forced in Ace Combat.

Does Trilinear work in DX? As the option isn't limited to OGL.

gregory38:
On Dx, it will behave as bilinear.

Dreadmoth:
>Just to be sure, delete your GSdx.ini file and retry.

I have tried deleting it, both issues still occur.

It seems the aliased upscaling in Ratchet & Clank is limited to OpenGL as well. Guess I should have checked the other renderers earlier, I tried every other setting...

On-screen output on the current commit looks like this (same settings):
[OpenGL](http://i.imgur.com/dbYiEWI.png)
vs
[Direct3D 11](http://i.imgur.com/h4Ibzgg.png)

refractionpcsx2:
> On-screen output on the current commit looks like this (same settings):
> OpenGL
> vs
> Direct3D 11

OpenGL looks like a lower resolution, strange :/

gregory38:
I think I get it. There are 2 linear modes. It depends if texture is expanded or reduced. I need to merge them.

gregory38:
I updated the code. But looking at your picture, I'm not sure which solution is better. The house on the right feels much better with nearest. You can see the details.

Anyway please test the new code.

gregory38:
@Dreadmoth @FlatOutPS2 is it better on RC ?

Is trilinear forced give you a better results or barely ?

Dreadmoth:
Nice, the last update restored both the smoother screen output and the quality of upscaling in Ratchet & Clank. :+1: 
>looking at your picture, I'm not sure which solution is better. The house on the right feels much better with nearest. You can see the details.

Nearest did look a bit sharper in places, but at the cost of aliasing and scaling artifacts.
Sharpening using external shaders works well to achieve decent clarity on low resolutions without the jaggies:
current OGL output with sharpening

-
>Is trilinear forced give you a better results or barely ? 

Trilinear (Forced) provides slightly improved results in Ratchet & Clank. 
Using the regular Trilinear option, most texture aliasing is eliminated, but some bands of differing detail levels are visible on some textures - hard to spot if standing still, more noticeable when moving.
Trilinear (Forced) seems to blend the bands together, creating a smoother gradient of detail.
If anisotropic filtering is used, Trilinear and Trilinear (Forced) look identical (as far as I can tell), but Trilinear still has a lower performance hit.
Differences between the two options in other games I've tried seem pretty subtle.

gregory38:
@Dreadmoth 
I created a new filter option. Could you compare the 3 trilinear options. (I added trilinear bi)

Dreadmoth:
Trilinear: Seems to apply bilinear filtering the same as Bilinear (PS2) - useful to avoid filtering 2D / sprite-based games or to prevent full-screen blur that may occur in some 3D games with forced filtering.

Trilinear (Forced Bi): Useful for filtering UI/HUDs in games that don't do so by default (Tribes: AA, Drakan) and for filtering textures in certain games (Spider-Man 2 distant buildings). Performance in the games I tried was at worst 1fps slower compared to the base Trilinear option.

Trilinear (Forced Tri): Increases trilinear coverage compared to the other options. Some games seem to need this option for trilinear filtering to be applied at all (Predator: Concrete Jungle). Improvements in games with low-resolution and/or low contrast textures seem quite subtle. Noticeably reduces texture aliasing / shimmering in Darkwatch and Metal Arms: Glitch in the System compared to the base Trilinear option. Performance impact varies significantly. In CPU-limited games like Ratchet & Clank, the setting had no impact on performance. In Tribes: Aerial Assault, performance decreased by up to 40%. Typical performance impact seems to be 15 - 25%.

> If anisotropic filtering is used, Trilinear and Trilinear (Forced) look identical (as far as I can tell)

^I was wrong - with AF applied to both, Trilinear (Forced Tri) still looks better.

I noticed no new graphical glitches with any of the Trilinear settings enabled.

gregory38:
@Dreadmoth Thanks for your test. Do you think Forced tri is worth it ? I'm afraid that user will set it to the max and then complain on the speed impact.

Dreadmoth:
I think it's worth having available - if you think a note about performance in the texture filtering tooltip wouldn't be enough, would it be possible to have Forced Tri only available through GSdx.ini?

FlatOutPS2:
Just add a note `Forced Tri (very slow)`. It should be clear enough. :p

gregory38:
@turtleli could you give a quick review of this pr.

FlatOutPS2:
We either need to add DX support or remove the Trilinear options from the list when using the Direct3D renderers. Leaving it as it is now will confuse many users who will complain it doesn't work.

Dreadmoth:
Recent changes seem to have reintroduced some bad textures and texture flickering in Ratchet & Clank and Jak 3.
[Ratchet & Clank 3 GS dump](https://github.com/PCSX2/pcsx2/files/514695/Ratchet.Clank.3.-.mipmap.-.dev-1432.zip)

gregory38:
ah interesting so it does create issue.

jcdenton2k:
Having 'more accurate' options at the expense of performance is good to do IF the user is informed through the tooltip or other notification.

Having the options not work under DirectX rendering (but still appear in the options) will confuse some people.  Should possibly do some code to detect current renderer and adjust available options as needed.

plugins/GSdx/GSDeviceOGL.cpp
```diff
@@ -227,7 +230,7 @@ GSTexture* GSDeviceOGL::CreateSurface(int type, int w, int h, bool msaa, int fmt
 
 	// A wrapper to call GSTextureOGL, with the different kind of parameter
 	GSTextureOGL* t = NULL;
-	t = new GSTextureOGL(type, w, h, fmt, m_fbo_read);
+	t = new GSTextureOGL(type, w, h, fmt, m_fbo_read, m_mipmap > 1 || m_filter > 2);
 	if (t == NULL) {
```
turtleli: Hmm. Comment replies don't appear in the feed?

Anyway, I think nullptr can only be returned if `new(std::nothrow)` is used.

gregory38:
I merged the branch, the gui/trilinear can be updated later (by someone else ;) ).

plugins/GSdx/GSSetting.cpp
```diff
@@ -33,8 +33,11 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_FILTER:
 			return "Control the texture bilinear filtering of the emulation.\n\n"
 				"Nearest:\nAlways disable interpolation, rendering will be blocky.\n\n"
-				"PS2:\nUse same mode as the PS2. It is the more accurate option.\n\n"
-				"Forced:\nAlways enable interpolation. Rendering is smoother but it could generate some glitches.";
+				"Bilinear PS2:\nUse same mode as the PS2. It is the more accurate option.\n\n"
+				"Bilinear Forced:\nAlways enable interpolation. Rendering is smoother but it could generate some glitches.\n\n"
+				"Trilinear:\nUse openGL trilinear interpolation when PS2 uses mipmaps.\n\n"
+				"Trilinear Forced Bilinear:\nSame as above but always enable bilinear interpolation.\n\n"
+				"Trilinear Ultra:\nAlways enable full trilinear interpolation. !WARNING SLOW!.\n\n";
```
jcdenton2k: It is important to specify bilinear filtering since it differs from trilinear filtering.

Issue #1603, opened 2016-10-03
GSDX: JAK2 USA(CRC: 9184AAF1) is still bugs


arcade1977:
Jak‘s hair and face are still exist strange bugs，eyes are black

ssakash:
Please use the [forums](http://forums.pcsx2.net/index.php) instead for support. Your report lacks a lot of necessary information to be considered as a [**_valid_ bug report**](https://github.com/PCSX2/pcsx2/wiki/How-to-create-useful-and-valid-issues).

PR #1604, opened 2016-10-03
GameDB: Add Tiger Woods PGA Tour series VU Clamping Mode fixes

FlatOutPS2:
Fixes black textures on characters in Tiger Woods PGA Tour 2004, 2005,
and 06 through 10.

Also add Jak 2 US CRC to GSCrc list. Fixes issue: #1603

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -10553,6 +10553,7 @@ Region = PAL-E
 Serial = SLES-51887
 Name   = Tiger Woods PGA Tour 2004
 Region = PAL-E
+vuClampMode = 2 // fix black textures on characters
 ---------------------------------------------
 Serial = SLES-51888
 Name   = RTL Skiijumping 2004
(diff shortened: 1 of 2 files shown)


avih:
Do you know for a fact that the DB fix works on all the versions which you modified?

FlatOutPS2:
All games use the same engine and have the exact same issues. I can't confirm 2008 and 2009, but as it has been confirmed they have the exact same issue and engine, the fix should be the same too.

The name changes are the correct ones. The previous names weren't in sync with the online list, some aren't now because the years on the compatibility list don't use the abbreviation (2006-2006 should be 06-10), otherwise they're in sync now.

avih:
OK, please add to the DB commit message which serials are known to fix the issue and which are presumed to fix it.

gregory38:
Thanks for the jak2 crc.

avih:
Thanks. For some reason I didn't get a notification when you updated the commit message, and only noticed it now.

Issue #1562, opened 2016-09-12
Meta: GSdx SW issue


gregory38:
Single issue report to centralize all current issue/limitation with the SW renderer.  The trick is to avoid core bug.

FlatOutPS2:
Might be core issue, but potentially GSdx issue:
http://forums.pcsx2.net/Thread-Smackdown-4-Shut-Your-Mouth-special-meter-doesn-t-display-properly

Alpine Racer 3. GSdx issue that also affects SW, there's a CRC hack for HW mode. Without the CRC hack DX11 HW is fine, but DX9 HW, OGL HW and SW modes are affected.

TheLastRar:
Network Profile Wizard (the one built-in with some games), Most text/boxes stuck in top left.
I believe this is a GS emulation issue rather than a core issue. 
https://github.com/PCSX2/pcsx2/issues/287

FlatOutPS2:
Colin McRae Rally 2005 SW issue has been resolved thanks to https://github.com/PCSX2/pcsx2/commit/3994141d4085c6365876efc0c561893cda9aba96

gregory38:
We need to add the jak issue and the explanation in this report.

prafullpcsx2:
Although it wasn't mentioned in this thread but I would like to report that software mode ghosting in Catwoman is also fixed by auto flush hack. HW mode ghosting is same as before.

gregory38:
Good, I wanted to ask you :+1:

monsterjamp:
@gregory38 The cutscenes in Thrillville are cut off even in SW mode. Can you confirm if this is a core bug? There's a GS dump in this forum thread: http://forums.pcsx2.net/Thread-Bug-Report-Thrillville-NTSC-U

gregory38:
@monsterjamp did you test zerogs? Honestly a partial texture upload is likely a core bug.

monsterjamp:
ZeroGS keeps crashing when it gets to the cutscenes so I can't test it.

There's also another issue but this one is only present is SW mode. In the pause menu, the bottom right of the screen is missing.

Hardware mode:
![2016-10-06 13_11_43-slot_ 0 _ speed_ 101 60 36 _ limiter_ normal _ gsdx d3d11 hw _ 640x445 _ bob](https://cloud.githubusercontent.com/assets/7116269/19165020/d3e7f540-8bc7-11e6-95a5-1e1d4458c342.png)


Software mode:
[...]

Issue #1269, opened 2016-03-29
[GSdx hardware regression] Distorted icon texture in War of the Monsters' character select menu.


Aced14:
**PCSX2 version:**
* v1.5.0-dev-504-gf6ee945 (2016-03-28; f6ee94589ffe86ff61d72cf33a425fdf3bc93d4d).
* The first automated build containing this regression was v1.3.1-641-gde644c5 (2015-07-01; de644c54376449381a23f97620bdd402a7352fbf). The regression itself was introduced in one of [July 1, 2015's 51 first commits](https://github.com/PCSX2/pcsx2/compare/35081f922a08d5069e9e1f4c66a2b9f140f5ef97...de644c54376449381a23f97620bdd402a7352fbf).

**PCSX2 options:**
Defaults. No options modifications affect this issue.

**Plugins used:**
Defaults (GSdx32-AVX, LyliPad, SPU2-X, cdvdGigaherz, USBnull, FWnull and DEV9ghzdrk). No particular plugins affect this issue.

**Plugin settings:**
GSdx (Direct3D11/OpenGL hardware modes).

**Description:**
While doing research in the lead-up to issue #1267, I noticed a distorted icon texture in my "good" screenshot of War of the Monster's character select menu. Specifically, the third character select icon looks black/distorted. It's supposed to look like a praying mantis. I determined it wasn't related to #1267 and set out to create a separate issue for it.

It should be noted that this issue can only be observed by enabling one of GSdx's hardware renderers before booting War of the Monsters. Toggling between renders at runtime (even re-initializing the same hardware renderer the game was booted with) will fix the distorted texture.

The commit that introduced this regression is from July 1, 2015. The automated build for that day's first commit is fine. At the end of the day (51 commits later), this regression was in place. Unfortunately, there are only 3 automated builds available from that day. One is for that day's first commit. The other 2 are for that day's final 2 commits. [Here's a list of all 51 relevant commits and what they changed](https://github.com/PCSX2/pcsx2/compare/35081f922a08d5069e9e1f4c66a2b9f140f5ef97...de644c54376449381a23f97620bdd402a7352fbf).

To resolve this, a developer will need to pinpoint the specific commit or recognize a change that could've caused it amongst the aforementioned commits. It's probably something minor, given that all the other character select menu's icons appear properly and since re-initializing the same renderer at runtime fixes it.

**Screenshots:**
* OpenGL
[...]

PR #1605, opened 2016-10-03
Greg/gsdx berserk 1526

gregory38:
First commit might require some tests.

### DIFF
--- a/plugins/GSdx/GSRendererHW.cpp
+++ b/plugins/GSdx/GSRendererHW.cpp
@@ -493,7 +493,7 @@ void GSRendererHW::Draw()
 	GSTextureCache::Target* rt = NULL;
 	GSTexture* rt_tex = NULL;
 	if (!no_rt) {
-		rt = m_tc->LookupTarget(TEX0, m_width, m_height, GSTextureCache::RenderTarget, true);
+		rt = m_tc->LookupTarget(TEX0, m_width, m_height, GSTextureCache::RenderTarget, true, fm);
 		rt_tex = rt->m_texture;
 	}
 
--- a/plugins/GSdx/GSTextureCache.cpp
+++ b/plugins/GSdx/GSTextureCache.cpp
@@ -403,7 +403,7 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 	return src;
 }
 
-GSTextureCache::Target* GSTextureCache::LookupTarget(const GIFRegTEX0& TEX0, int w, int h, int type, bool used)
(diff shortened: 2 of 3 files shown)


FlatOutPS2:
Does it require testing because it could trigger a regression in some games(if so, any likely candidates)?

gregory38:
Try those games
* https://github.com/PCSX2/pcsx2/issues/1269
* Metal gears solid 3
Honestly texture cache updates are :scream: :gun:

FlatOutPS2:
I had a hunch and was right, this also fixes the HUD on WWE Smackdown Shut Your Mouth! :+1:

gregory38:
HW renderer ? Nice :smiley:

FlatOutPS2:
It was broken on all renderers, now it's fixed on all renderers. ;)

gregory38:
Oh even the SW renderer. Could you tell me if it is the 2nd or the 3rd commit that fix it.

FlatOutPS2:
Sure, I'll check which one fixes it later.

I linked this HUD issue in https://github.com/PCSX2/pcsx2/issues/1562

So you can mark that one as solved when it's been merged.

FlatOutPS2:
The Smackdown HUD issue is also solved by the 8 bits commit, so no 4 bits case yet. :p

gregory38:
Ok. Honestly I'm already surprised that more 1 game uses this trick. I added the 4 bits because pseudonym told me that we ought to emulate all the cases. In doubt, it can't hurt, can it?

FlatOutPS2:
I'm sure that it'll fix some game. If someone is crazy enough to use the illegal 8 bits, it's very likely there's someone crazy enough to use 4 bits as well.

FlatOutPS2:
Test Drive Pause menu HUD is fixed by the first commit. :)

gregory38:
Very good. 3 games fixed in one shot.

FlatOutPS2:
The PR also seems to show a new effect that occurs when taking a picture of a ghost in Project Zero. Actual ghost still a no show in HW though. :p

gregory38:
new effect ?

FlatOutPS2:
It's new to PCSX2, as in as far as I know it has never been visible in earlier versions. ;)

gregory38:
sw or hw ?

FlatOutPS2:
Both I believe, but it's very noticeable in HW because the ghost itself is invisible(the effect occurs around the outline of the ghost).

gregory38:
Maybe it would be easier to generate a nice dump so I can properly reproduce the ghost issue on my side.

gregory38:
I pushed the 8/4 bits frame format to the master. Only remain the texture cache change that is unsafe.

FlatOutPS2:
Played a bit in MGS3 but I didn't notice any regressions.

PR #1606, opened 2016-10-04
FPU has only 2 control registers.

Huud:
I assumed it was made 32 for a recompiler hack or something, but I tested it with a couple of games, played fine.

### DIFF
--- a/pcsx2/R5900.h
+++ b/pcsx2/R5900.h
@@ -194,7 +194,7 @@ union FPRreg {
 
 struct fpuRegisters {
 	FPRreg fpr[32];		// 32bit floating point registers
-	u32 fprc[32];		// 32bit floating point control registers
+	u32 fprc[2];		// 32bit floating point control registers
 	FPRreg ACC;			// 32 bit accumulator
 	u32 ACCflag;        // an internal accumulator overflow flag
 };


sudonim1:
a) No, it has 32 but 1-31 aren't used.
b) This makes code access whatever the hell is outside of this structure, corrupting memory.
c) This breaks save states without incrementing the save state version.
d) Please don't assume that code is correct because "a couple of games play fine".

Huud:
a) Unless these are some of the undocumented registers, the manual flat out says the FPU has two 32-bit floating-point controller registers, so I guess someone misread it as 32 floating-point controller registers.
B) Assuming there is no insane code that accesses non-existent/unused registers, and unless the recompiler does not access it by address, this should not be a problem, but nothing guarantees the code is sane, so fair enough.
C) Then delay it til the next increment.
D) Really? Do I have to test the whole library before I consider submitting any commit? Couple of games is more than enough, especially for a very essential part that the emulator should not be able to boot a single game if it was broken.

I guess this is not an important correction, it should not break or fix anything other than a small wasted memory, but being too scared of correcting the code because you don't know "whatever the hell is outside of this structure" might do is something I don't understand, not unless you found some bugs that you are not sharing?

gregory38:
a) There are 32 logical registers. But 0-15 are aliased to 0 and 16-31 are aliased to 31.
b) Of course, all PCSX2 will try to access either 0 or 31. Both interpreters/recompilers are now broken. Code is sane it behaves like the hardware. Besides you potentially broke some SSE2 alignments. Honestly I'm even surprised that you manage to compile it.

```
 but being too scared of correcting the code because you don't know "whatever the hell
 is outside of this structure" might do is something I don't understand, not unless
 you found some bugs that you are not sharing?
```
https://github.com/PCSX2/pcsx2/search?utf8=%E2%9C%93&q=fprc%5B31%5D&type=Code
```
fpuRegs.fprc[31] = 0x01000001; // fpu Status/Control
cpuRegs.GPR.r[_Rt_].SD[0] = (s32)fpuRegs.fprc[31]; 
```

Huud:
That's a more informative and convincing response, thanks.

sudonim1:
Broken SSE alignments was something else I checked for but didn't find to be applicable, which is why I didn't mention it.  (And by 1-31 I of course meant 1-30 but I was in a rush to close this in case somebody thought merging it was a bright idea.)

sudonim1:
Also the FPU is not an essential part of the emulator used in every game.  In fact it is used very little because VU is generally superior.

sudonim1:
And memory corruption issues like the one you introduced can affect anything or nothing frequently or rarely depending on the data layout.

Issue #1576, opened 2016-09-19
Lilypad: Allow for separate bindings between controller modes (ie. Dualshock 2 and Guitar)


MrCK1:
Would it be possible to have separate bindings for each controller mode?

For example, say I have everything bound to their matching button on my DS3. Unfortunately, Band Hero forces me to rebind a few buttons due to its' guitar detection. The problem is that when you switch to and from "Guitar" mode, I have to change the bindings everytime I play another game because every mode shares the same configuration.

vsub:
Not sure if this is what you are asking for but did you tried "Allow binding multiple PS2 controls to one PC control"

willkuer:
Until it is resolved you could always save the lilypad ini file. With a frontend or the command line you could load the ini game specific in an automatic fashion. However what you suggest sounds like a solid feature.

vsub:
> Sorry, that won't work for this scenario- I think that's meant for keyboards so you can press L1 and R1 with 1 keypress.

No that's not what it means...it means that you can set more than one key/button to execute for example Square
![2016-09-19_191109](https://cloud.githubusercontent.com/assets/5032540/18639604/34684f4c-7e9d-11e6-86ed-ff4e0a0dcc47.jpg)
Setting it like this,I can execute Pad1 square by using my usb gamepad,my logitech gamepad or by pressing J

Setting for example L1 and square to a single button/key will execute a combo but that's not because of that option

PR #1609, opened 2016-10-05
LilyPad: Add separate bindings

FlatOutPS2:
**LilyPad: Add separate bindings for each pad type**

Adds separate bindings for each of the pad types (DualShock2, Guitar,
Pop'n Music). This allows the user to change the button configuration to
better suit the Guitar and Pop'n Music pads without messing up the
bindings already setup for the DS2.

Fullfills request: #1576


FlatOutPS2:
What do to with the Linux side? It fails the build because it uses some of the same files, which are now incompatible.

willkuer:
Virtual machine is your friends...

FlatOutPS2:
I don't have the the disc space for a Linux build enviroment. It would be easier to move old versions of the modified LilyPad files to the Linux folder.

gregory38:
sorry for the extra work. maybe one day someone will finish the job.

MrCK1:
Separate bindings work great. Nice job!

FlatOutPS2:
> Edit: although bindings are separated properly now, the indicator showing the # of bindings is still constant between both modes. I don't know if we want to do anything with that?

Do you mean the # of bindings in the pad selection area of the Pads section on the General tab? Because that works perfectly for me.

MrCK1:
@FlatOutPS2 Yes I did. Never mind I guess 😛

FlatOutPS2:
@MrCK1 So was this a mistake or does it still not behave correctly?

MrCK1:
What I meant was that Lilypad could say that there are 39 bindings, but number is the same in Dualshock 2 and Guitar Since the bindings are separated now by mode, the number of displayed bindings should change with the selected mode.

FlatOutPS2:
Yes, I know. And like I said it works on my side. But if it doesn't work for you, we need to look into it. Can you share your LilyPad.ini file and can you try if it works correctly with a fresh ini file?

MrCK1:
Deleting the .ini didn't work. I also noticed that any deleted binding is removed in all modes

[Here's the .ini](http://pastebin.com/gPmcQJ0H) - I also left the deleted binding (R1) removed

FlatOutPS2:
I can't reproduce your issue, with or without your ini. I even tried it on a different PC and it worked without issue too. 

Is that the ini before you used it with my PR? It only contains one profile for pad 1 DualShock2 in the old format that hasn't been saved by the LilyPad from this PR.

FlatOutPS2:
The Travis-ci linux build seem to fail because a few files aren't in clang format, but it doesn't specify anything. :/

gregory38:
I will improve the logging. You can search `did not match`  in the build log
```
file plugins/LilyPad/Config.cpp did not match clang-format
file plugins/LilyPad/DirectInput.cpp did not match clang-format
file plugins/LilyPad/Linux/Config.cpp did not match clang-format
file plugins/LilyPad/LilyPad.cpp did not match clang-format
```
You need to install clang 3.8 plugin to MSVC. And then use the default PCSX2 formatting.
http://llvm.org/builds/
```
Go to Tools->Options->LLVM/Clang->ClangFormat and put file in the Style option field. (could be optional)
```

gregory38:
@turtleli do you have a link to the version 3.8 of clang-format for windows. My above link is latest svn.

FlatOutPS2:
@gregory38 Thanks. I found the 3.8 version here: http://llvm.org/releases/download.html

I did the see the files mentioned, but what I meant was it didn't specify what's wrong inside those files. It should only be a few minor issues.

gregory38:
Ok. I would change the script to give more details (hopefully it isn't a clang/linux/msvc version issue)

FlatOutPS2:
> Edit: did you try to run clang-format on those files ?

I tried, but nothings happens when I apply it. :/

gregory38:
I was afraid of this behavior. Which version did you take ? 3.8 or 3.8.1 ?

FlatOutPS2:
Yeah, turns out I selected a slightly newer version of the plugin by mistake.

gregory38:
Which one exactly? I need to check your latest commit too. Maybe we need to pack the plugin in git. I need to think about it.

FlatOutPS2:
I'd have to look it. The VS plugin wasn't included in the 3.8.0 build, so I took the plugin version from the same date as the 3.8.0 release. The first time I downloaded a version from the wrong date (slightly newer).

gregory38:
It seems like 3.8 got some bugs that were fixed a bit later. However the ubuntu/debian version is 3.8.0. I need to check the behavior of 3.9.

gregory38:
I try 3.9. But behavior is still bad...
The only different is this one
```
diff --git a/plugins/LilyPad/WndProcEater.cpp b/plugins/LilyPad/WndProcEater.cpp
index 0379005..a4a9f5d 100644
--- a/plugins/LilyPad/WndProcEater.cpp
+++ b/plugins/LilyPad/WndProcEater.cpp
@@ -127,7 +127,7 @@ bool WndProcEater::SetWndHandle(HWND hWnd)
     //printf( "(Lilypad) (Re)-Setting window handle! -> this=0x%08x, hWnd=0x%08x\n", this, hWnd );
 
     Release();
-    SetProp(hWnd, L"LilyHaxxor", (HANDLE) this);
+    SetProp(hWnd, L"LilyHaxxor", (HANDLE)this);
 
     eatenWndProc = (WNDPROC)SetWindowLongPtr(hWnd, GWLP_WNDPROC, (LONG_PTR)OverrideWndProc);
     hWndEaten = (eatenWndProc) ? hWnd : 0;
```
I didn't review whitespace and tabulation. I guess I need to create a script to check the behavior.

gregory38:
@FlatOutPS2 
Could you create a github wiki page with the link to the clang-format plugin. And how to install/configure it on windows. So I could forward it to @ramapcsx2

FlatOutPS2:
> Could you create a github wiki page with the link to the clang-format plugin. And how to install/configure it on windows. So I could forward it to @ramapcsx2 

I would, but the problem is I have't found a stable plugin. The older plugin I found worked for LilyPad, but might still differ from 3.8 when applying it to other parts of the PCSX2 code.

FlatOutPS2:
@gregory38 Added it here: https://github.com/PCSX2/pcsx2/wiki/Clang-Format-Visual-Studio-plug-in

FlatOutPS2:
I've decided to add the other features as a separate PR after this has been merged.

plugins/LilyPad/Config.cpp
```diff
@@ -933,9 +945,8 @@ int LoadSettings(int force, wchar_t *file)
         }
     }
 
-    for (int i = 0; i < sizeof(BoolOptionsInfo) / sizeof(BoolOptionsInfo[0]); i++) {
+    for (int i = 0; i < sizeof(BoolOptionsInfo) / sizeof(BoolOptionsInfo[0]); i++)
         config.bools[i] = GetPrivateProfileBool(L"General Settings", BoolOptionsInfo[i].name, BoolOptionsInfo[i].defaultValue, file);
-    }
```
turtleli: While I personally prefer not using braces for single statement if/else/for/while, this is kind of a pointless change since you're not changing anything else and should be avoided in large commits.

plugins/LilyPad/Config.cpp
```diff
@@ -946,9 +957,8 @@ int LoadSettings(int force, wchar_t *file)
         config.keyboardApi = WM;
     config.mouseApi = (DeviceAPI)GetPrivateProfileIntW(L"General Settings", L"Mouse Mode", 0, file);
 
-    if (config.debug) {
+    if (config.debug)
```
turtleli: Same here.

plugins/LilyPad/Linux/Config.cpp
```diff
@@ -374,21 +382,30 @@ int LoadSettings(int force, wchar_t *file)
             }
             last = 1;
             unsigned int uid;
-            int port, command, sensitivity, turbo, slot = 0, deadZone = 0;
+            int port, command, sensitivity, turbo, slot = 0, deadZone = 0, padtype = 0;
+            ;
```
turtleli: Hmm.

plugins/LilyPad/LilyPad.cpp
```diff
@@ -815,9 +817,8 @@ int saveStateIndex = 0;
 s32 CALLBACK PADinit(u32 flags)
 {
     // Note:  Won't load settings if already loaded.
-    if (LoadSettings() < 0) {
+    if (LoadSettings() < 0)
         return -1;
-    }
```
turtleli: Unnecessary changes here and below too.

plugins/LilyPad/Linux/Config.cpp
FlatOutPS2: Hmm indeed. Don't know how that got there.

plugins/LilyPad/Linux/Config.cpp
```diff
@@ -374,21 +382,30 @@ int LoadSettings(int force, wchar_t *file)
...
             string[w] = 0;
-            int len = sscanf(string, " %u , %i , %i , %i , %i , %i , %i", &uid, &port, &command, &sensitivity, &turbo, &slot, &deadZone);
+            int len = sscanf(string, " %u , %i , %i , %i , %i , %i , %i , %i", &uid, &port, &command, &sensitivity, &turbo, &slot, &deadZone, &padtype);
             if (len >= 5 && type) {
                 VirtualControl *c = dev->GetVirtualControl(uid);
                 if (!c)
                     c = dev->AddVirtualControl(uid, -1);
                 if (c) {
-                    BindCommand(dev, uid, port, slot, command, sensitivity, turbo, deadZone);
+                    if (padtype == NULL) { // If ini file is imported from older version, make sure bindings aren't applied to "Unplugged" padtype.
```
FlatOutPS2: Good point. :p

plugins/LilyPad/LilyPad.cpp
```diff
@@ -511,7 +512,7 @@ void Update(unsigned int port, unsigned int slot)
 
     LastCheck = t;
 
-    int i;
+    int i, padtype;
```
FlatOutPS2: I moved some stuff around. Must've missed this one.

plugins/LilyPad/Config.h
```diff
@@ -21,13 +21,13 @@
 extern u8 ps2e;
 
 enum PadType {
+    // Changing the number of PadTypes requires numPadTypes in InputManager.h to be modified as well.
```
FlatOutPS2: And how would you count it?

ssakash: Maybe turtleli meant something like this?

Currently
> DisabledPad = 0,
     Dualshock2Pad = 1,
     GuitarPad = 2,
     PopnPad = 3

If you add a new enum called ``Pad Count`` then,

>> DisabledPad = 0,
     Dualshock2Pad = 1,
     GuitarPad = 2,
     PopnPad = 3,
     PadCount = 4 // Counts the total number of pads and is flexible as it adjusts itself when more enums are added

FlatOutPS2: Yes, I thought of that as well. But the use of fake enums is usually frowned upon. :p

turtleli: Well, if you can think of a better way, go for it? ;)

plugins/LilyPad/XInputEnum.cpp
```diff
@@ -113,31 +114,31 @@ class XInputDevice : public Device
...
+                L"B",
+                L"X",
+                L"Y",
+                L"Left Trigger",
+                L"Right Trigger",
+                L"Left Thumb X",
+                L"Left Thumb Y",
+                L"Right Thumb X",
+                L"Right Thumb Y",
+            };
```
turtleli: Still don't know why this was moved. IMO it would actually be better left where it originally was and `i < 21` changed into `i < (sizeof(names) / sizeof(names[0]))` or similar (I didn't check if it works).

plugins/LilyPad/DirectInput.cpp
```diff
@@ -220,7 +222,7 @@ class DirectInputDevice : public Device
             format.dwFlags = 0;
             format.dwNumObjs = numPhysicalControls;
             format.rgodf = formats;
-            int res = did->SetDataFormat(&format);
+            // int res = did->SetDataFormat(&format); // is not used
```
turtleli: Are you sure? `res` doesn't look like it's used anywhere, but `did->SetDataFormat()` seems important (I don't know DirectInput stuff). If this line is commented out, then setting the members of `format` then becomes useless.

plugins/LilyPad/DualShock3.cpp
```diff
@@ -208,13 +208,13 @@ class DualShock3Device : public Device
     int index;
     HANDLE hFile;
     DS3Command sendState;
-    unsigned char getState[49];
+    unsigned char getState[49] = {};
     OVERLAPPED readop;
     OVERLAPPED writeop;
     int writeCount;
-    int lastWrite;
+    int lastWrite = NULL;
```
turtleli: IMO it's more readable if you use 0 instead of NULL when the type is not a pointer.

At the very least, I associate `NULL` with pointers.

plugins/LilyPad/DirectInput.cpp
FlatOutPS2: Yeah, `did->SetDataFormat()` is required. I should've only removed the `int rest = `.

plugins/LilyPad/Linux/Config.cpp
```diff
@@ -81,29 +81,31 @@ struct GeneralSettingsBool
...
 {
     fprintf(stderr, "delete binding %d:%d\n", port, slot);
-    Binding *bindings = dev->pads[port][slot].bindings;
+    int active_padtype = config.padConfigs[port][slot].type;
+    Binding *bindings = dev->pads[port][slot][padtype].bindings;
     int i = b - bindings;
-    memmove(bindings + i, bindings + i + 1, sizeof(Binding) * (dev->pads[port][slot].numBindings - i - 1));
-    dev->pads[port][slot].numBindings--;
+    if (padtype == active_padtype) // Only do the memmove when it is the active padtype
+        memmove(bindings + i, bindings + i + 1, sizeof(Binding) * (dev->pads[port][slot][padtype].numBindings - i - 1));
```
turtleli: This still needs fixed.

Apart from that it seems fine and I see no obvious errors (I haven't tested the code myself though).

FlatOutPS2:
I've made a change to Config.cpp to prevent to warn the user of conflicting/deleted bindings when swapping profiles.

turtleli:
Would it be better if the swap bindings code swapped all bindings on one pad with the respective bindings on the other pad (DS2 with DS2, guitar with guitar, and so on)?

I'm not sure what your current swap bindings code is supposed to do (Lilypad code isn't too readable for me, hence why I take donkeys to review the PR).
Test case:
1. Start with no Lilypad.ini
2. Open Lilypad config - Both pads set to DS2 and have 27 bindings
3. Swap bindings, confirm change -> Now both pads have 11 bindings -> Makes no sense to me.

FlatOutPS2:
> Swap bindings, confirm change -> Now both pads have 11 bindings -> Makes no sense to me.

I changed part of the code last minute to streamline it and that caused the issue. It should normally only ever delete from one of the pads.

I've changed the behaviour. Swapping bindings will no longer let you delete any bindings. 
It now works like this:
Default behaviour: Swap pad1 with pad2 swaps all bindings from all pad types for both pads (like you said).
New "Swap Inidivdual pad types" option allows swapping and clearing inidivual pad types.

Example to get delete warning (which should rarely happen):
1. Both pads set to DS2, change pad1 pad type to Guitar.
2. Code checks if pad1 guitar has conflicting bindings with pad2 DS2 (if multiple bindings option is disabled).
3. If so , show a warning message explaining the issue and ask if you want to delete the conflicting binding from the newly selected pad type.
4. You choose :
- Yes delete the conflicting bindings and switch to new pad type.
- No do not delete anything and switch to new padtype.
- Cancel and return to the previously selected padtype without doing anything.

turtleli:
I tested before you made your latest edits and it seemed to work ok.

Would it be better to have both "all swap/clear" and "individual swap/clear" in the pop up menu at the same time instead of having to check/uncheck the checkbox to change the behaviour?

P.S. If you make further changes, do it in a separate commit. It means I have less to review and we can get this merged quicker. I can always squash merge on my end.

plugins/LilyPad/Config.cpp
```diff
@@ -2079,26 +2200,50 @@ INT_PTR CALLBACK GeneralDialogProc(HWND hWnd, unsigned int msg, WPARAM wParam, L
...
+                            config.padConfigs[port2][slot2].type = config.padConfigs[port1][slot1].type;
+                            config.padConfigs[port1][slot1].type = padCfgTemp.type;
+                        }
                         for (int i = 0; i < dm->numDevices; i++) {
                             if (dm->devices[i]->type == IGNORE)
                                 continue;
-                            PadBindings bindings = dm->devices[i]->pads[port1][slot1];
-                            dm->devices[i]->pads[port1][slot1] = dm->devices[i]->pads[port2][slot2];
-                            dm->devices[i]->pads[port2][slot2] = bindings;
+                            if (config.swapPadtypes) {
```
turtleli: `swapPadTypes` isn't great naming if it's changing between individual binding swap and all bindings swap IMO.

FlatOutPS2:
> Would it be better to have both "all swap/clear" and "individual swap/clear" in the pop up menu at the same time instead of having to check/uncheck the checkbox to change the behaviour?

Yeah, I had some issues with it before, but I've managed to make it work.

turtleli:
Something's not quite right.

Test case (starting with default settings, both DS2):
1. Click port 1 multitap.
2. Right click PAD 1B, choose "Swap with Pad 1A, DualShock 2 bindings only" => Pad 1B DS2 bindings swap with Pad 2 DS2 bindings.

If you want, you can always keep just the "Swap all bindings options" for this PR and then add individual swap in a later PR (entirely up to you). I think some of the other LilyPad changes you wanted to make probably could have been done in parallel if this PR wasn't holding it up.

FlatOutPS2:
> If you want, you can always keep just the "Swap all bindings options" for this PR and then add individual swap in a later PR (entirely up to you). I think some of the other LilyPad changes you wanted to make probably could have been done in parallel if this PR wasn't holding it up.

I fixed the issue you described. It was a very small placement error that only negatively affected this one option. :p I'd rather have everything that's now in the PR function correctly, then postponing it. The other changes will come later as I will make a tooltips PR next.

FlatOutPS2:
> If you want, you can always keep just the "Swap all bindings options" for this PR and then add individual swap in a later PR (entirely up to you). I think some of the other LilyPad changes you wanted to make probably could have been done in parallel if this PR wasn't holding it up.

I fixed the issue you described. It was a very small placement error that only negatively affected this one option. :p I'd prefer to have everything that's now in the PR functioning correctly, rather then postponing it. The other changes will come later as I will make a tooltips PR next.

plugins/LilyPad/Linux/Config.cpp
```diff
@@ -167,20 +167,22 @@ int BindCommand(Device *dev, unsigned int uid, unsigned int port, unsigned int s
...
-                PadBindings *p = dev->pads[port2] + slot2;
-                for (int i = 0; i < p->numBindings; i++) {
-                    Binding *b = p->bindings + i;
-                    int uid2 = dev->virtualControls[b->controlIndex].uid;
-                    if (b->controlIndex == controlIndex || (!((uid2 ^ uid) & 0xFFFFFF) && ((uid | uid2) & (UID_POV | UID_AXIS)))) {
-                        DeleteBinding(port2, slot2, dev, b);
-                        i--;
+                for (int padtype2 = 0; padtype2 < numPadTypes; padtype2++) {
+                    if (port2 == (int)port && slot2 == (int)slot)
+                        continue;
```
turtleli: `if` should be moved outside the loop.

plugins/LilyPad/Default.ini
```diff
@@ -10,6 +10,7 @@ XInput=1
 DualShock 3=0
 Multitap 1=0
 Multitap 2=0
+Swap Pad Types=0
```
turtleli: Needs removed.

PR #1612, opened 2016-10-08
Added missing CRC for Grand Theft Auto San Andreas.

zapling:
Could not get this to work [[SOLVED] GTA SA skip draw hack without losing orange sky effect](http://forums.pcsx2.net/Thread-SOLVED-GTA-SA-skip-draw-hack-without-losing-orange-sky-effect). Did some digging and realized my DVD CRC for Grand Theft Auto was not added in this file.

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -459,6 +459,7 @@ CRC::Game CRC::m_games[] =
 	{0xFF9C0E93, TouristTrophy, US, 0},
 	{0xCA9AA903, TouristTrophy, EU, 0}, //crc hack not fully working on PAL, still needs brightness =0
 	{0xA1B3F232, GTASanAndreas, EU, 0}, // cutie comment
+	{0xB440A8FE, GTASanAndreas, EU, 0},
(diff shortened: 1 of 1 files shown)


avih:
Thanks.

gregory38:
You can now emulate the game with the auto flush option. This crc hack is useless now.

FlatOutPS2:
> You can now emulate the game with the auto flush option. This crc hack is useless now.

The fix does still work though. And it's probably a better experience than the auto flush option if you're on a low spec system.

gregory38:
is it that bad on this game. I think it just a matter of a couple of draw calls.

FlatOutPS2:
It's probably not that bad overall, but if you're on a system that can barely run the game at full speed, it could have a noticeable impact. We could add the CRC's to the legeacy plugin and remove the hack from the modern plugin.

jcdenton2k:
Request:
Leave the CRC in there but disabled by default.  Players who need to use it could manually enable it under the 'GameFixes' section.  Maybe add in information in the official PCSX2 wiki to inform others easier.

For those who aren't aware of the auto-flush option, add this info to the PCSX2 wiki as well.

PR #1615, opened 2016-10-09
psxmode: SPU2-X backwards compatibility with the original SPU

ramapcsx2:
psxmode: SPU2-X backwards compatibility

### DIFF
--- a/pcsx2/Sio.h
+++ b/pcsx2/Sio.h
@@ -17,7 +17,7 @@
 
 // Let's enable this to free the IOP event handler of some considerable load.
 // Games are highly unlikely to need timed IRQ's for PAD and MemoryCard handling anyway (rama).
-//#define SIO_INLINE_IRQS
+#define SIO_INLINE_IRQS
 
 #include "MemoryCardFile.h"
 
--- a/pcsx2/Sio.h
+++ b/pcsx2/Sio.h
@@ -17,7 +17,7 @@
 
 // Let's enable this to free the IOP event handler of some considerable load.
 // Games are highly unlikely to need timed IRQ's for PAD and MemoryCard handling anyway (rama).
-#define SIO_INLINE_IRQS
+//#define SIO_INLINE_IRQS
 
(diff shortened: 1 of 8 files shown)


FlatOutPS2:
Is the sound mostly correct now?

ramapcsx2:
Not sure about correct but it's working. There'll be details like pitch mod nuances and the whole core still runs at 48khz but overall, it's good :)

ramapcsx2:
Oh yea, and reverb is dangerous. Due to the nature of the beast, it can go off in some situations. I've tested a few games and it's fine in those. But don't be surprised if it runs into a feedback loop (loud!).
Getting reverb right is very hard.

MrCK1:
I can hear the new reverb during the PSX intro, but there's no music on the title screen for C12:Final Resistance

FlatOutPS2:
Just tried it with Driver and it sounds really good. During gameplay all sounds and music seem to work. FMV's don't have any sound yet though.

ramapcsx2:
No CD music yet. It will require CDVD plugin support.

FlatOutPS2:
From what I recall, in Driver the FMV sounds aren't CD tracks.

I also noticed a few notes in the music on some levels in Driver are wrong. One of the notes was just an empty beep.

ramapcsx2:
Okay, I'll check Driver out.
I've also noticed some garbage data (bleeps, adpcm error logs in console) randomly in Chrono Cross.

FlatOutPS2:
Driver isn't very crashy. I've only had crashes when trying to restart a level. Otherwise I haven't had any crashes.

ramapcsx2:
Okay, I have Driver in mind. There's still lots to do anyway. This pr is to get things going :)

sudonim1:
Don't ever expect full compat with psx games, this is basically an experiment and it will be a miracle if most playstation games end up being only moderately buggy, let alone pcsx2 becoming a playstation emulator replacement.

Issue #694, opened 2015-08-02
Devil May Cry scaling issue


gab:
As entered in the old bugtracker with screenshots: https://code.google.com/p/pcsx2/issues/detail?id=1319

I reproduced the bug with pcsx2 1.2.1 r5875 and gsdx32-avx2-r5875 under Windows 7.
Here is the GS dump: https://dl.dropboxusercontent.com/u/22764021/Devil%20May%20Cry%20scaling%20bug%20-%20gsdx%20r5875.7z

Whenever you set a resolution higher than native (multiplier or custom res) in Devil May Cry 1:
1- A small box containing part of the screen appears at the upper left
2- **Text renders in this small box instead of appearing at its normal place on the screen** (wasn't mentioned in the original bug report)
3- None of the HW hacks fix it (only thing I haven't tried is the layers hack)
4- The bug is present in all modes, HW or SW, DirectX 9/11 or OpenGL

It's a pity since the game assets are actually very nice, and look beautiful in high resolution.

The reason I'm adding this back here is that I tested with an **older Gsdx** (r890, 0.1.14), scaled it up to 1920x1200, and **the problem is not present**. Of course running that very old plugin has many inconvenients including worse speed and little configurability. But if it scales fine with that old code, there's gotta be a way to bring back the same behavior in current Gsdx trunk with a hack!

ssakash:
please also provide a GS dump along with the report for the developers to look into it.

gab:
GS Dump uploading to Dropbox, it's gonna be available in 1 min - link already edited into the description.

Jozman:
I know how to solve that issue. You need to press "select", than go to "Options", after that select controller settings and stay in it for 5-7 seconds, at the end return to game bu pressing ▲ triangle at the medium speed and wallah, you have the health and DT box recovered, but it's still low resolution which is not that of a problem. :)

MrCK1:
Issue still active as of latest git

FlatOutPS2:
The box should be stretched to full screen, but is instead stuck in the top left corner. The box is for text messages that should be placed in the bottom middle of the screen.

gab:
Unfortunately it seems that the old issue disappeared from Google Code Archive along with the screenshots. Have those issues been archived somewhere else?
My description should be accurate enough to identify the bug without screenshots, though.

FlatOutPS2:
Here's a clear screenshot:
[...]

PR #1617, opened 2016-10-13
GSDX: Generalize the scaling parameters on SetScale() function calls

ssakash:
**Summary of changes**:
- Generalize the calculation of multipliers passed on the two `SetScale()` functions. Might help fixing #694 on custom resolutions. Previous code only considered cases for native scaling resolutions and also potentially had a bug as it didn't check for `m_renderer->CanUpscale()` but probably not critical.
- Add a constructor to dispatch same values on x/y for GSVector2. Might help to save some time in the future :)

@FlatOutPS2 

Could you check whether custom resolution on DMC is fine with this PR? When I tried the dump provided on #694, it only gave the top part of the screen as output. Would be better to check the game directly instead of relying on these dumps :P

### DIFF
--- a/plugins/GSdx/GSTextureCache.cpp
+++ b/plugins/GSdx/GSTextureCache.cpp
@@ -403,6 +403,33 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 	return src;
 }
 
+void GSTextureCache::ScaleTexture(GSTexture* texture)
+{
+	if (!m_renderer->CanUpscale())
+		return;
+
+	float multiplier = static_cast<float>(m_renderer->GetUpscaleMultiplier());
+	bool custom_resolution = (multiplier == 0);
+	GSVector2 scale_factor(multiplier);
+
+	if (custom_resolution)
+	{
+		int width = m_renderer->GetDisplayRect().width();
+		int height = m_renderer->GetDisplayRect().height();
+		int real_height = static_cast<int>(round(m_renderer->GetInternalResolution().y / texture->GetScale().y));
+
+		// Fixes offset issues on Persona 3 (512x511) where real value of height is 512
+		if (real_height % height == 1)
+			height = real_height;
+
+		GSVector2i requested_resolution = m_renderer->GetCustomResolution();
+		scale_factor.x = static_cast<float>(requested_resolution.x) / width;
+		scale_factor.y = static_cast<float>(requested_resolution.y) / height;
+	}
+
+	texture->SetScale(scale_factor);
+}
+
 GSTextureCache::Target* GSTextureCache::LookupTarget(const GIFRegTEX0& TEX0, int w, int h, int type, bool used, uint32 fbmask)
 {
 	const GSLocalMemory::psm_t& psm_s = GSLocalMemory::m_psm[TEX0.PSM];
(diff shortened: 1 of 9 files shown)


plugins/GSdx/GSTextureCache.cpp
```diff
@@ -403,6 +403,33 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 	return src;
 }
 
+void GSTextureCache::ScaleTexture(GSTexture* texture)
+{
+	if (m_renderer->CanUpscale())
```
ssakash: Potentially it would be nice to replace this gigantic scoped if
```
if (m_renderer->CanUpscale())
{


*stuffs*

}
```

with 

```
if (!m_renderer->CanUpscale()) return;
```

Any opinions?

FlatOutPS2:
I tried DMC with several different custom resolutions and it no longer has the box in the top left, it's placed at the correct location thanks to the PR. :+1:

ssakash:
I pushed some more commits, testing for regressions would be nice. :)

plugins/GSdx/GSTextureCache.cpp
ssakash: Done 👍

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -153,7 +153,9 @@ bool GSRendererHW::CanUpscale()
 		return false;
 	}
 
-	return m_upscale_multiplier!=1 && m_regs->PMODE.EN != 0; // upscale ratio depends on the display size, with no output it may not be set correctly (ps2 logo to game transition)
+	// Upscale ratio depends on the display size, with no output it may not be set correctly (ps2 logo to game transition)
+	// Only allow upscaling when either one of the output circuit is enabled.
+	return m_upscale_multiplier!=1 && m_regs->PMODE.EN != 0 && (m_regs->PMODE.EN1 || m_regs->PMODE.EN2);
```
gregory38: Just found the existence of the `EN` field. So it is a 2 bits concatenation of EN1 and EN2. So if either EN1 or EN2 is 1 then EN != 0. And if both EN1 and EN2 == 0 then EN == 0.

FlatOutPS2:
DMC is still fine wit the latest change. As far the few other games I tried with it. I did notice an unrelated regression, which I need to investigate.

turtleli:
> Add a constructor to dispatch same values on x/y for GSVector2. Might help to save some time in the future :)

IMO the new constructor isn't useful. The time saved is pretty minimal.

ssakash:
> IMO the new constructor isn't useful. The time saved is pretty minimal.

The time saved isn't the only factor on why I added it, it also seems better from a code readability perspective at some cases where variables with longer names are passed as actual arguments. Here's an example : 

```
float multiplier = static_cast<float>(m_renderer->GetUpscaleMultiplier());
GSVector2 scale_factor(multiplier, multiplier); // the double multiplier looks awkward
```

**VS**

```
float multiplier = static_cast<float>(m_renderer->GetUpscaleMultiplier());
GSVector2 scale_factor(multiplier); 
```

It's not frequently used at most places so it's not much useful but I still feel it would be convenient to have a constructor which passes same values on both fields.

gregory38:
My 2 cents, it doesn't hurt to have this extra constructor.

gregory38:
Is it ready to be merged ?

ssakash:
> Is it ready to be merged ?

The code is complete, I have no more changes left to do.

PR #1586, opened 2016-09-24
Gsdx refresh gui options

gregory38:
I removed the safe accurate blending option. Not sure it plays well on windows.

PR #1621, opened 2016-10-14
GSDX-Windows: Changes to Texture filtering option on GUI

ssakash:
**Summary of changes**:

* Add a nice enum for texture filtering options.
* Show Trilinear combobox items only when OpenGL (hardware mode) is used.

### DIFF
--- a/plugins/GSdx/GS.h
+++ b/plugins/GSdx/GS.h
@@ -1414,3 +1414,13 @@ enum class GSVideoMode : uint8
 	DTV_720P,
 	DTV_1080I
 };
+
+enum class Filtering : uint8
+{
+	Nearest,
+	Bilinear_Forced,
+	Bilinear_PS2,
+	Trilinear,
+	Trilinear_Bilinear_Forced,
+	Trilinear_Always
+};
--- a/plugins/GSdx/GSDeviceOGL.cpp
+++ b/plugins/GSdx/GSDeviceOGL.cpp
@@ -75,7 +75,7 @@ GSDeviceOGL::GSDeviceOGL()
 	GLState::Clear();
 
 	m_mipmap = theApp.GetConfigI("mipmap");
-	m_filter = theApp.GetConfigI("filter");
+	m_filter = static_cast<Filtering>(theApp.GetConfigI("filter"));
 
 	// Reset the debug file
 	#ifdef ENABLE_OGL_DEBUG
@@ -228,8 +228,9 @@ GSTexture* GSDeviceOGL::CreateSurface(int type, int w, int h, bool msaa, int fmt
 {
 	GL_PUSH("Create surface");
 
+	bool trilinear = m_filter == Filtering::Trilinear || m_filter == Filtering::Trilinear_Bilinear_Forced || m_filter == Filtering::Trilinear_Always;
 	// A wrapper to call GSTextureOGL, with the different kind of parameter
-	GSTextureOGL* t = new GSTextureOGL(type, w, h, fmt, m_fbo_read, m_mipmap > 1 || m_filter > 2);
+	GSTextureOGL* t = new GSTextureOGL(type, w, h, fmt, m_fbo_read, m_mipmap > 1 || trilinear);
 
 	// NOTE: I'm not sure RenderTarget always need to be cleared. It could be costly for big upscale.
 	// FIXME: it will be more logical to do it in FetchSurface. This code is only called at first creation
--- a/plugins/GSdx/GSSettingsDlg.cpp
+++ b/plugins/GSdx/GSSettingsDlg.cpp
@@ -188,7 +188,31 @@ void GSSettingsDlg::OnInit()
 	AddTooltip(IDC_LOGZ);
 	AddTooltip(IDC_LARGE_FB);
 
+	UpdateFilteringCombobox();
 	UpdateControls();
+}
+
+void GSSettingsDlg::UpdateFilteringCombobox()
+{
+	INT_PTR i;
+	ComboBoxGetSelData(IDC_RENDERER, i);
+	bool opengl = static_cast<GSRendererType>(i) == GSRendererType::OGL_HW;
+	bool hw_mode = opengl || static_cast<GSRendererType>(i) == GSRendererType::DX1011_HW || static_cast<GSRendererType>(i) == GSRendererType::DX9_HW;
+	if (!hw_mode)
+		return;
+
+	uint8 filter = (ComboBoxGetSelData(IDC_FILTER, i)) ? static_cast<uint8>(i) : static_cast<uint8>(theApp.GetConfigI("filter"));
+	if (!opengl) //Currently Trilinear is only exclusive to OpenGL, remove those combobox items when any other renderer is used
+	{
+		auto head = theApp.m_gs_filter.begin();
+		auto tail = head + static_cast<uint8>(Filtering::Trilinear);
+		vector<GSSetting> list(head, tail);
+		ComboBoxInit(IDC_FILTER, list, std::max(uint8(Filtering::Nearest), std::min(filter, uint8(Filtering::Bilinear_PS2))));
+	}
+	else
+	{
+		ComboBoxInit(IDC_FILTER, theApp.m_gs_filter, filter);
+	}
 }
 
 bool GSSettingsDlg::OnCommand(HWND hWnd, UINT id, UINT code)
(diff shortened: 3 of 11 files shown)


FlatOutPS2:
Where's the mipmapping option? :p

plugins/GSdx/GS.h
```diff
@@ -1414,3 +1414,13 @@ enum class GSVideoMode : uint8
 	DTV_720P,
 	DTV_1080I
 };
+
```
willkuer: maybe use flags?

```
enum class Filtering
{
	Nearest = 0x00,
	Bilinear = 0x01,
	Trilinear = 0x02,
	
	PS2 = 0x10,
	Forced = 0x20,
	Slow = 0x40,
	

	Bilinear_Forced = Bilinear | Forced,
	Bilinear_PS2  = Bilinear | PS2,
//	Trilinear_PS2 = Trilinear | PS2, //? is this the idea of Trilinear?
	Trilinear_Forced = Trilinear | Forced,
	Trilinear_Slow = Trilinear | Slow
};
```

Then the trilinear request in GSDeviceOGL::CreateSurface would be more simple (`(m_filter&Trilinear)==Trilinear)` or `(m_filter&Trilinear)>0`)

gregory38:
```
+	Trilinear_Forced,
+	Trilinear_Slow
```
Hum, Triliner_Slow => Trilinear_Always ? Trilinear_Forced => Triliner_Bilinear_Forced ? I'm sure we can find better name.

plugins/GSdx/GSDeviceOGL.cpp
```diff
@@ -224,12 +224,20 @@ void GSDeviceOGL::GenerateProfilerData()
 	}
 }
 
+ static bool IsTrilinearFiltering(Filtering filter)
+{
+	if (filter == Filtering::Trilinear || filter == Filtering::Trilinear_Forced || filter == Filtering::Trilinear_Slow)
+		return true;
+
+	return false;
```
gregory38: You need to stop with this kind of style.
```
return (filter == Filtering::Trilinear || filter == Filtering::Trilinear_Forced || filter == Filtering::Trilinear_Slow);
```
is enough.

gregory38: Actually don't create a function. Just a local boolean variable is enough.

ssakash:
> Where's the mipmapping option? :p

Honestly, I have no clue on how to handle it as it needs to be a three state checkbox for hardware mode and two state checkbox for software mode. FWIW I don't really care about the Mipmap option. [**I let you handle it as stated over here :P**](https://github.com/PCSX2/pcsx2/pull/1586#issuecomment-250981670).

gregory38:
Well it could still be a 3 state options in SW mode. Or maybe we can keep it a 2 state options. The 3rd state is a proof of concept but I don't think it will be usable.

ssakash:
> Well it could still be a 3 state options in SW mode. Or maybe we can keep it a 2 state options. The 3rd state is a proof of concept but I don't think it will be usable.

That's not the only concern, currently it's placed on software mode settings groupbox and I'm not really sure on which would be the ideal place to move it. Honestly it would be much better if someone who's actually competent on GUI coding/Design expertise handles it instead of handling it myself :P

FlatOutPS2:
I did make the suggestion elsewhere to add it as a dropdown list below `Interlacing (F5):`.

gregory38:
I will add the linux part. I will let you adapt it on windows.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -188,9 +188,30 @@ void GSSettingsDlg::OnInit()
...
+	INT_PTR i;
+	ComboBoxGetSelData(IDC_RENDERER, i);
+	bool OpenGL = static_cast<GSRendererType>(i) == GSRendererType::OGL_HW;
+	uint8 filter = (ComboBoxGetSelData(IDC_FILTER, i)) ? static_cast<uint8>(i) : static_cast<uint8>(theApp.GetConfigI("filter"));
+
+	vector<GSSetting> list = theApp.m_gs_filter;
+	if (!OpenGL) //Currently Trilinear is only exclusive to OpenGL, remove those combobox items when any other renderer is used
+	{
+		for (uint8 option = uint8(Filtering::Trilinear_Always); option != uint8(Filtering::Bilinear_PS2); option--)
+			list.erase(list.begin() + option);
```
turtleli: `erase()` can remove multiple items, so the `for` loop is unnecessary.

Alternatively, instead of copying `theApp.m_gs_filter` and then erasing the trilinear options, you could simply copy only the necessary vector elements using `insert()`.

plugins/GSdx/GSDeviceOGL.cpp
```diff
@@ -228,8 +228,9 @@ GSTexture* GSDeviceOGL::CreateSurface(int type, int w, int h, bool msaa, int fmt
 {
 	GL_PUSH("Create surface");
 
+	bool Trilinear = m_filter == Filtering::Trilinear || m_filter == Filtering::Trilinear_Bilinear_Forced || m_filter == Filtering::Trilinear_Always;
```
turtleli: `trilinear`? Variables in this file all seem to be lowercase.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -188,9 +188,30 @@ void GSSettingsDlg::OnInit()
...
 	AddTooltip(IDC_LARGE_FB);
 
+	UpdateFilteringCombobox();
 	UpdateControls();
 }
 
+void GSSettingsDlg::UpdateFilteringCombobox()
+{
+	INT_PTR i;
+	ComboBoxGetSelData(IDC_RENDERER, i);
```
turtleli: Maybe nicer to early return unless it's a hardware renderer?

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -188,9 +188,30 @@ void GSSettingsDlg::OnInit()
...
+	bool OpenGL = static_cast<GSRendererType>(i) == GSRendererType::OGL_HW;
+	uint8 filter = (ComboBoxGetSelData(IDC_FILTER, i)) ? static_cast<uint8>(i) : static_cast<uint8>(theApp.GetConfigI("filter"));
+
+	vector<GSSetting> list = theApp.m_gs_filter;
```
turtleli: It's unnecessary to copy `theApp.m_gs_filter` in the OpenGL case.

gregory38:
@turtleli I let you merge this PR when it is good.

plugins/GSdx/GSdx.cpp
```diff
@@ -199,12 +199,12 @@ void GSdxApp::Init()
...
 
-	m_gs_filter.push_back(GSSetting(0, "Nearest", ""));
-	m_gs_filter.push_back(GSSetting(1, "Bilinear", "Forced"));
-	m_gs_filter.push_back(GSSetting(2, "Bilinear", "PS2"));
-	m_gs_filter.push_back(GSSetting(3, "Trilinear", "Ultra/Slow"));
-	m_gs_filter.push_back(GSSetting(4, "Trilinear", ""));
-	m_gs_filter.push_back(GSSetting(5, "Trilinear", "Forced Bilinear"));
+	m_gs_filter.push_back(GSSetting(static_cast<uint32>(Filtering::Nearest), "Nearest", ""));
+	m_gs_filter.push_back(GSSetting(static_cast<uint32>(Filtering::Bilinear_Forced), "Bilinear", "Forced"));
+	m_gs_filter.push_back(GSSetting(static_cast<uint32>(Filtering::Bilinear_PS2), "Bilinear", "PS2"));
```
turtleli: Bilinear PS2 comes before Bilinear Forced in the tooltip.

plugins/GSdx/GS.h
```diff
@@ -1414,3 +1414,13 @@ enum class GSVideoMode : uint8
...
+enum class Filtering : uint8
+{
+	Nearest,
+	Bilinear_Forced,
+	Bilinear_PS2,
```
turtleli: Bilinear PS2 comes before Bilinear Forced in the tooltip.

plugins/GSdx/GSdx.cpp
turtleli: Oh, just realised it might mess users' config up if it's changed. Hmm.

ssakash: Well, there is also the other option of modifying the tooltip instead.

PR #1625, opened 2016-10-15
GSDX-Windows: Add mipmap combobox on GUI

ssakash:
**Summary of changes**:
- Add a combobox for mipmap option. (Following the same naming format as on Linux GUI)

mipmap


gregory38:
Could you remove `UserHacks_mipmap` logic too.

ssakash:
Maybe the space bellow ``"Rendering Threads"`` and ``Edge Anti-aliasing`` on the software mode group box could be removed?

> Could you remove UserHacks_mipmap logic too.

Yeah, sure. :)

gregory38:
Oh add a nice commit message saying that HW mipmap is now supported for the worst and for the best.

ssakash:
>  add a nice commit message saying that HW mipmap is now supported for the worst and for the best.

I assume it's intended for people reading the changelog on buildbot? If yes, I think I'll pass that offer. It would mislead people into thinking that I was the author of the mipmapping change. ;)

ssakash:
Nevermind, added it to the commit message. I love hogging extra credits 😈

plugins/GSdx/GSdx.rc
```diff
@@ -236,6 +236,8 @@ BEGIN
     COMBOBOX        IDC_ADAPTER,71,55,166,118,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
     LTEXT           "Renderer:",IDC_STATIC,6,72,34,8
     COMBOBOX        IDC_RENDERER,71,70,166,118,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
+    COMBOBOX        IDC_MIPMAP,71,100,166,118,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
+    LTEXT           "Mipmap (Insert):",IDC_MIPMAP_TEXT,6,102,53,8
```
ssakash: It was taken from [gregory's code](https://github.com/PCSX2/pcsx2/commit/3c30db965dfcc41f101e43ede37980506630e17d#diff-748d76fbcaa43637145d131d37054679R397), so it's all in the hands of @gregory38 if he wants to change it :)

Personally I just want the Linux-Windows GUI's to have similar names.

ssakash: > And mipmapping is much better than just mipmap.

Yeah, I also like ``Mipmapping`` over ``Mipmap`` but I used the latter for consistency on Linux-Windows GUI.

@gregory38 

Any opinion on ``Mipmapping`` **VS** ``Mipmap`` and ``(Ins)`` **VS** ``(Insert)``?

gregory38: Honestly I don't care, you're big boy, I let you judge what it is the best.

gregory38:
```
I assume it's intended for people reading the changelog on buildbot? If yes, I think I'll pass that offer. It would mislead people into thinking that I was the author of the mipmapping change. ;)
```
Oh, you crack my plan :p I want people be aware that now the mipmap option will have an impact on the HW renderer too.

plugins/GSdx/GSdx.rc
ssakash: Thanks, I have also made the string changes suggested by @FlatOutPS2 on Linux GUI.

FlatOutPS2:
I just noticed outdated code comments are finally shrunk again. :D

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -380,6 +384,9 @@ void GSSettingsDlg::UpdateControls()
 #ifndef ENABLE_OPENCL
 		ShowWindow(GetDlgItem(m_hWnd, IDC_OPENCL_DEVICE), SW_HIDE);
 		ShowWindow(GetDlgItem(m_hWnd, IDC_OPENCL_TEXT), SW_HIDE);
+#else
+		ShowWindow(GetDlgItem(m_hWnd, IDC_MIPMAP), SW_HIDE);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_MIPMAP_TEXT), SW_HIDE);
```
turtleli: Move all the GUI stuff down a bit instead?

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -33,18 +33,17 @@ GSRendererHW::GSRendererHW(GSTextureCache* tc)
 	, m_channel_shuffle(false)
 	, m_lod(GSVector2i(0,0))
 {
+	m_mipmap = theApp.GetConfigI("mipmap");
```
gregory38: you don't need this line. (check parent class)

plugins/GSdx/GSSettingsDlg.cpp
ssakash: Sorry, I was too lazy to do it.

gregory38:
LGTM :+1: Do we need to merge trilinear filtering first?

turtleli:
Windows side seems fine. "Half" and "Full" mipmapping aren't self explanatory though - would be nice to fix that later.

> Do we need to merge trilinear filtering first?

I don't think the PRs will cause merge conflicts with each other.

gregory38:
I just don't know how to explain it nicely. However, something is bothering me. Toggling F9 won't change the mipmapping option. I would be nice to keep the option always on for SW renderer but still keep the option for debug purpose.

turtleli:
Maybe "Off", "Software/OpenCL", "All"? Not sure either.

> Toggling F9 won't change the mipmapping option.

Do you mean Insert? I think that only toggles between 0 and 1 right now, even if it starts at 2.

gregory38:
Off => easy
half => fully emulated in SW (OpenCL ?). On the HW renderer, it only computes the valid layers of the textures. It doesn't do mipmapping per se but it is enough to fix texture corruption.
All => really try to emulate mipmapping on the HW renderer. Slow and various texture cache issue.

My issue is that recommended setting for the HW renderer is `OFF` expect `HALF` on some games. The recommended setting for the SW renderer is `HALF`/`FULL`. If people use F9 to avoid some game issue, they potentially won't have mipmaping emulated.

```
I think that only toggles between 0 and 1 right now, even if it starts at 2.
```
I fixed it normally.

turtleli:
Ok, I see now. Sounds fun ;)

> I fixed it normally.

Oops. I looked at the wrong commit.

FlatOutPS2:
Maybe we could change it to:
```
Off
Full (Software) / Minimum (Hardware, fast)
Full (Software) / Full (Hardware, slow)

```
And add the explanation to the tooltip.

refractionpcsx2:
Personally i think the dropdown for software mip mapping should be separate, as we're separating all the other options.

willkuer:
So handling hw mipmap and sw mipmap completely independent?

FlatOutPS2:
> Personally i think the dropdown for software mip mapping should be separate, as we're separating all the other options.

We could keep the existing SW mipmapping toggle, but that would also require changing the insert key behaviour(switch between off and on in SW mode, and between off, minimum and full in HW mode).

ssakash:
> Off
Full (Software) / Minimum (Hardware, fast)
Full (Software) / Full (Hardware, slow)

I don't like the ``Full (Software)`` redundancy on combobox.

**Note**: Removing one of the items dynamically from the combobox is also not a valid solution as the console message for ``Insert`` toggle uses the text from the combobox vector and it would cause some conflicts.

> Personally i think the dropdown for software mip mapping should be separate, as we're separating all the other options. 

Unless I'm wrong, only the unique/exclusive options are separated. Have we separated any option before which is common to both hardware/software mode?

> So handling hw mipmap and sw mipmap completely independent?

I personally don't think it would be a nice choice. (Waste of GUI space)

gregory38:
Otherwise I propose you an easy solution. We keep only OFF/HALF (whatever you name them). And the full option will be for dbg/dev build, if one day it really works we could resume this discussion.

refractionpcsx2:
What if you want mipmapping enabled on software but not hardware? This is why I think we should keep the two separate, the number of options could get silly.  

 TBF it should probably be permanently on for Software mode and not have an option to disable it as that is designed for emulation accuracy and mipmapping is required to achieve that, then just have a dropdown for hardware mode.

gregory38:
I'm in favor of always on but
1/ keep a way to disable it for debug (so it is easy to test if the hw issue is related to mipmapping)
2/ I think it would be possible to optimize SW renderer. Current code load all layers. Potentially we could skip some top layer (actually only the first one could be enough).  That being said I guess people will use HW renderer for the speed now.

gregory38:
Here an others solutions
* add m_mipmap = true in the SW renderer constructor
  * SW will always start with mipmap on
  * but it will still be possible to toggle it with `INS`
  * As F9 recreate the object (I think), mipmapping on SW will still be on.
* Then move back the combo box in HW section

ssakash:
After some discussions on FB, we had eventually decided to go with [**gregory's suggestion**](https://github.com/PCSX2/pcsx2/pull/1625#issuecomment-254451727).

> add m_mipmap = true in the SW renderer constructor 

Simply ingenious :) **(or)** I'm _probably_ dumb af for not realizing this nice idea.

gregory38:
```
Simply ingenious :) (or) I'm probably dumb af for not realizing this nice idea.Simply ingenious :) (or) I'm probably dumb af for not realizing this nice idea.
```
Easy kid, I'm an engineer :) (which is `ingénieur` in French, and ingenious is `ingénieux` same latin root). 
Seriously I'm glad we found a good compromise. I let you do the update then :+1:

FlatOutPS2:
> Easy kid, I'm an engineer :) (which is ingénieur in French, and ingenious is ingénieux same latin root). 

Uncanny though how close that is to the French word for inferior: `inférieur`... :p

gregory38:
It doesn't work, it isn't the same root. You're only allowed to change the prefix or the suffix.

FlatOutPS2:
> It doesn't work, it isn't the same root. You're only allowed to change the prefix or the suffix.

This is GS territory, anything is allowed.

gregory38:
I'm the GS King. Mouah ah ah :p

turtleli:
@ssakash Is this PR finished?

ssakash:
> @ssakash Is this PR finished?

Yes, the code was completed five days ago :P

woj1993:
@ssakash Can you marge it?

gregory38:
Add a comment to explain why mipmap is forced in sw renderer. And the impact to do it.

refractionpcsx2:
> Add a comment to explain why mipmap is forced in sw renderer. And the impact to do it.

Hmm I dunno about that, doing it for that makes as much sense as doing it for any other GS feature and we don't need a comment about them.

Maybe change "Enable HW Hacks" to be "Enable Hacks" then add an option in the hacks panel to disable mipmapping on software.  Just a thought.

ssakash:
> Maybe change "Enable HW Hacks" to be "Enable Hacks" then add an option in the hacks panel to disable mipmapping on software. Just a thought.

Well, you can still disable mipmapping on software mode by pressing ``Insert``. Though I do agree that a comment about it seems pointless.

woj1993:
@gregory38 @ssakash @refractionpcsx2 Sorry that I am bothering you. I think that no one uses software mode without mipmaps. Having to press "Insert" to turn it off is good. What are we doing with this branch?

refractionpcsx2:
That is what we're doing

willkuer:
> What are we doing with this branch?

We? We users let the devs do their thing.

refractionpcsx2:
You users also get to give your input,  this is a  community project :)

FlatOutPS2:
Yeah, looks like it's just an unfortunate choice of words. 

I do prefer a UI option to turn off mipmapping in SW mode, as some games just look a little better without mipmapping, and having to turn it off manually everytime you start a game can get annoying.

lightningterror:
How about something like this?

v1
mpv1

v2
mpv2

I recommend v2 if some games need switching between software and hardware render and mipmapping is not needed on one render.

FlatOutPS2:
We already discussed those options...

refractionpcsx2:
I suggest reading this issue from top to bottom before making suggestions so you don't look silly :P

lightningterror:
Ah , I thought we were still discussing it since @FlatOutPS2 mentioned about turning off mipmap on sw render so I thought I'd put my suggestions here as well.

ssakash:
> I do prefer a UI option to turn off mipmapping in SW mode, as some games just look a little better without mipmapping, and having to turn it off manually everytime you start a game can get annoying.

Then how would we handle the cases mentioned by gregory? It seems rather important as most people rely on F9 software switch for accurate rendering.

>My issue is that recommended setting for the HW renderer is OFF expect HALF on some games. The recommended setting for the SW renderer is HALF/FULL. **If people use F9 to avoid some game issue, they potentially won't have mipmaping emulated.**

FlatOutPS2:
Just also keep the current SW mipmapping toggle.

i30817:
I still hate that header with a burning passion. You realize that for people with 1366x768 monitors on gnome 3 the current non resizable box doesn't show everything and hides the bottom ok and cancel buttons right?

MrCK1:
> I still hate that header banner with a burning passion. You realize that for people with 1366x768 monitors on gnome 3 the current non resizable box doesn't show everything and hides the bottom ok and cancel buttons right? 

> It's not like you're the only one, QMC2 does the same, but that is always painful. Same thing for the 'Emulation Settings' actually. Learn to love 'tab tab tab enter' (or how many tabs it is, it varies).


@i30817 That isn't relative to this discussion. If you take another look at the repository; you'll see that PR #1523 is already addressing this issue

ssakash:
> Just also keep the current SW mipmapping toggle.

You mean the ``m_mipmap = 1`` force mipmapping toggle for software renderer (or) the ``Ins`` toggle? The latter is already there and the former is impossible to keep with GUI code as it implicitly enables mipmap whenever the constructor is created.

Though honestly, I prefer gregory's current suggestion. Software mode emulates lots of features like mipmapping and why would you need to specifically disable mipmapping at certain cases for software mode? It seems to barely impact performance at most cases. _(I might be wrong)_

However I do agree that a UI option would be nice for debugging mipmapping issues on GSDumpGUI but you can still use older versions of GSdx for that.

FlatOutPS2:
> You mean the m_mipmap = 1 force mipmapping toggle for software renderer (or) the Ins toggle? The latter is already there and the former is impossible to keep with GUI code as it implicitly enables mipmap whenever the constructor is created.

I mean the GUI option. You can keep the toggle, and change the underlying code so that it is only checked in SW mode.

> Though honestly, I prefer gregory's current suggestion. Software mode emulates lots of features like mipmapping and why would you need to specifically disable mipmapping at certain cases for software mode? It seems to barely impact performance at most cases. (I might be wrong)

Like I said earlier some games look better without mipmapping, and having to turn it off and on again everytime you play a game isn't desirable behaviour.

> However I do agree that a UI option would be nice for debugging mipmapping issues on GSDumpGUI but you can still use older versions of GSdx for that.

So you want all mipmapping debugging to happen in what will eventually be ancient GSdx versions?

gregory38:
I don't like my idea. Hum, I would need to bypass the force in the replayer. 

For the gui size, no I'm not aware. My 1920x1200 monitor is 8 years old. And previous monitor was likely 1280x1024. Even my 5 years old smartphone got a slightly better resolution (1200x800). So open an issue with a screenshot so we can understand which panel is too big.

FlatOutPS2:
@gregory38 Is there a difference in height between the GSdx panel on Windows and Linux? On Windows it's 644 pixels tall, so there should be enough room for it to function on a x768 screen.

turtleli:
>Is there a difference in height between the GSdx panel on Windows and Linux? On Windows it's 644 pixels tall, so there should be enough room for it to function on a x768 screen.

On Linux it's dependent on the GTK theme and maybe the DE/WM too. With a GTK2 build, Adwaita theme, Fluxbox WM, the GSdx Linux dialog has a height of about 778px. It might be greater when other configurations are used. Personally I think the logos aren't useful and should be removed from both the Windows and Linux dialogs.

MarcoEstevez:
Oh no, lovely logos, can not just make it wider, or scrollable?

ssakash:
I pushed an update to move the mipmapping combobox back to the general place and also added a code to dynamically change the combobox elements for hardware/software mode.

**Note**: The console message when mipmapping is toggled won't be consistent with the name off the combobox items.

FlatOutPS2:
I noticed a regression. If you set Hardware mipmapping to Full, then go to the Software renderer and back to the Hardware renderer, the Hardware mipmapping is set back top Half.

I personally think the Half option should be called Minimum or even Basic rather than Half. It's clearer to the user(half what? only half the screen gets mipmapping?) and Minimum or Basic would fit with other options already in place like the CRC option Minimum or BUA's Basic.

ssakash:
> I noticed an issue. If you set Hardware mipmapping to Full, then go to the Software renderer and back to the Hardware renderer, the Hardware mipmapping is set back top Half.

It's intentional behavior, the texture filtering combobox also has the same behavior.

* Select ``Trilinear`` at Texture Filtering combobox.
* Change back to D3D9/11 renderer and it'll get changed to ``Bilinear (PS2)``.
* Now change back to OpenGL renderer and you'll still have ``Bilinear (PS2)``.

Maybe this behavior could be fixed to your desired result but probably not worth it to mess with the code.

> I personally think the Half option should be called Minimum or even Basic rather than Half. It's clearer to the user("half what? only half the screen gets mipmapping?" a user might think) and Minimum or Basic would fit with other options already in place like the CRC option Minimum or BUA's Basic.

Maybe ``Medium``? Honestly I have no clue about the accuracy of the level 1 mipmapping option on hardware mode. (@gregory38 : any opinion ?)

FlatOutPS2:
> It's intentional behavior, the texture filtering combobox also has the same behavior.

Then who intended it and why?

jcdenton2k:
This is another reason that the 'UI revamp' should probably start being organized alongside making it all dynamically resizeable when necessary.  Let us expand or shrink any of the windows as we like and dynamically add or remove scroll bars when doing so ends up hiding elements of that window.

I realize this isn't common convention on Windows (it is all too rare) but it is expected UI behavior on Linux-based OS's.

MrCK1:
@jcdenton2k I'm not sure if you can even dynamically resize windows in WxWidgets. (I'm 99% wrong probably) Most window sizes are a hard coded height and width in the .cpp files IIRC.

gregory38:
You can. It just awful to get something nice. Anyway if GSdx panel is too big. It means I added too much options ;) Worst case I can still move post processing shader in a separate tab. It depends on the final shape of the mipmap option

refractionpcsx2:
Why don't we just make the dialog landscape? it would make life so much easier :P

ssakash:
> Then who intended it and why?

Well, that's how the combobox selection works. When you select ``Full`` for mipmapping on hardware mode, the value of the item is 2, eventually after you change it back to software mode, the value gets clamped and it gets changed to ``Enabled`` which has a value of 1. If you don't clamp the value then the combobox will be in an indeterminate state.

So when you switch back to hardware mode it will be ``Half`` (which has a value of 1) as the previous value for software mode ``Enabled`` is 1. I could probably correct this behavior by recording the previous selection state of the combobox with some static variables but that would probably make the code a bit messy and might compromise code readability which I personally don't prefer.

> I'm not sure if you can even dynamically resize windows in WxWidgets. (I'm 99% wrong probably) Most window sizes are a hard coded height and width in the .cpp files IIRC.

It is feasible on Windows GUI but a PITA if anyone wants to add UI elements at the future as the resize parameters (width, height) have to be manually adjusted for each addition. Tabs would be a better alternative.

> Why don't we just make the dialog landscape? it would make life so much easier :P

I suggested it before but you replied saying that you preferred portrait more :P

willkuer:
Well if you allow dynamic window creation you don't specify width, height and position element-wise.

FlatOutPS2:
> I could probably correct this behavior by recording the previous selection state of the combobox with some static variables but that would probably make the code a bit messy and might compromise code readability which I personally don't prefer.

Well, I assume most people would prefer user friendly UI behaviour over code readability. If you add a few clear comments to the code it shouldn't compromise the readability that much.

refractionpcsx2:
> I suggested it before but you replied saying that you preferred portrait more :P

I do, but if vertical height is such an issue for people with laptops.. :P

ssakash:
>  If you add a few clear comments to the code it shouldn't compromise the readability that much.

A good code is self-explanatory, no need of superfluous comments :P

FlatOutPS2:
> A good code is self-explanatory, no need of superfluous comments :P 

So you can't write good code?

ssakash:
> So you can't write good code?

Did you miss the commit which I had pushed earlier? https://github.com/PCSX2/pcsx2/pull/1625/commits/112e0e50703c4178303499d19e5a54acb4493fe2

FlatOutPS2:
Yes. Usually it says new commit(s) since your last visit, but it didn't so I missed it. :S

jcdenton2k:
> A good code is self-explanatory, no need of superfluous comments :P
I'm really anxious/worried about any programmer that thinks this way.
Yes, code legibility is a huge necessity to manage a large codebase like this PCSX2 emulator.

However, in the absence of clear accessible well-written explanatory documentation for everything, code comments are the next best alternative.  Even if documentation is available, a reference to said documentation in the code itself (as a comment) would be significantly more useful.

The idea is that if the coder poofs, others can go in and know exactly what something is and how it works by reading the documentation.  If not included as code comments then at least a comment referencing the (existing) documentation is preferable to nothing at all.

**Any programmer worth their code knows that it is easy for code to be 'self explanatory' to one coder while another person trying to interpret it has no clue what is going on.**

Even among programmers that return after a haitus, they can sometimes not know what code THEY wrote so long ago or how it works.

Lack of documentation and/or comments is just lazy.

I am not here to say anyone is incapable or not incredibly intelligent.  I merely say the lack of documentation and/or code commenting should be remedied rather than dismissed.

FlatOutPS2:
We still need some input on what the "half" option should be called. The best candidate as far as I'm concerned is "Basic". The word itself fits what the option is about and is also less likely to be wrongly interpreted. And it coheres nicely with the Blending Unit Accuracy options which also feature Basic and Full options.

And the current behaviour is mipmapping off in Hardware mode = mipmapping off in Software mode. Is everyone happy with that?

jcdenton2k:
@FlatOutPS2 
I would agree with the two options being 'Basic' and 'Full' for mipmapping.  Half might be a bit confusing unless it is something like 'half pixel offset' which makes more sense.

Also please be sure to include the really awesome/useful detailed tooltips with more info when you hover over the option.  I love those!  It makes it more useful instead of having to cross-reference with a CHM/HTML manual or help files.

lightningterror:
@FlatOutPS2 Basic sounds good. Partial can be a good candidate as well.

FlatOutPS2:
> Basic sounds good. Partial can be a good candidate as well.

I've considered Partial as well, but it could make people think it only partially works(and thus breaks several of the games it's used on), and they could turn it off because of it.

willkuer:
voting for 'Basic'

turtleli:
> And the current behaviour is mipmapping off in Hardware mode = mipmapping off in Software mode. Is everyone happy with that?

Hmm. Maybe the hardware and software mipmapping settings should be independent of each other? Just throwing out suggestions.

FlatOutPS2:
> Hmm. Maybe the hardware and software mipmapping settings should be independent of each other? Just throwing out suggestions.

That's what I prefer. As a user could now notice an issue unknowingly caused by turning off mipmapping in Hardware, then switch to Software to check if it affects both modes, notice the same issue and think it's a bug instead of merely a setting mistake.

ssakash:
Well, I think separating the mipmapping option isn't really a good idea for the following reasons : 

* It would mean two options to control mipmapping on the GUI, one would probably be a checkbox for software mode and the other one would be a combobox. The repetition of the UI elements text wouldn't look nice.

* Two INI variables instead of one.

* GSState handles checks for mipmap draw along with the INI variable and passes the information to HW/SW renderers, this should also get extended to deal with multiple INI variables instead of one.

However I'm still open to other suggestions, I have no problem in having the options independent to software/hardware if that's the consensus.

> As a user could now notice an issue unknowingly caused by turning off mipmapping in Hardware, then switch to Software to check if it affects both modes, notice the same issue and think it's a bug instead of merely a setting mistake.

Isn't this the same case for all the other options which influence HW/SW mode? Interlacing/FXAA?


On the other hand, changed the combobox text to ``Basic`` instead of ``Half`` as most people agreed with it.

FlatOutPS2:
> Isn't this the same case for all the other options which influence HW/SW mode? Interlacing/FXAA?

While those options can have a negative effect on the look of a game, they don't affect the graphics of a game as directly as mipmapping, and have far less serious negative effects that could convince people it's a bug.

refractionpcsx2:
I still think the two should be separate, making an extra variable is really a minor almost non-reason to do it. 

As for the GUI, we have a separate HW and SW section for settings, so I see no reason why they can't be added as separate options.

My biggest hate would be having it as a single option then having god knows how many options in the dropdown to handle the difference scenarios for hardware and software (like, 6 options or something?), that would be even more confusing IMO, it is better to have a single box for  hardware mode containing its 3 states and a separate 2 state option for software.

gregory38:
Yeah. It would be easier to have both option.

gregory38:
I mean 2 separate options

jcdenton2k:
Having separate options that appear/disappear (with the disappeared ones being disabled from modification) when you change rendering is something that I feel would be the ideal solution here.

ssakash:
Went with the two separate options suggestion, hopefully people are fine with it.

ssakash:
@gregory38 : Could you provide a patch for linux GUI? **(or)** if you prefer, I could just ifdef out these macros for now and you could make your update commit later.

```
 error: ‘IDC_MIPMAP_SW’ was not declared in this scope
```

gregory38:
Update the enum in GSSetting.h

gregory38:
@ssakash could you just update the enum and I will take care of the linux gui. I would like to merge this code as soon as possible.

ssakash:
Sorry for the delay, updated the commit. Also linux compilation error is not my fault, it seems to be caused by sudonim's [commit](https://github.com/PCSX2/pcsx2/commit/8259b29896ddfe3d8446d3983ff4c926de11e27d).

gregory38:
OoO I pushed a quick fix. If it is green a quick rebase will be nice so we can bug free here  :+1:

ssakash:
> If it is green a quick rebase will be nice

Travis CI seems green, awaiting turtleli's green light now. :)

plugins/GSdx/GSState.cpp
```diff
@@ -44,11 +44,11 @@ GSState::GSState()
 	, m_crc(0)
 	, m_options(0)
 	, m_frameskip(0)
+	, m_mipmap(1)
 {
 	// m_nativeres seems to be a hack. Unfortunately it impacts draw call number which make debug painful in the replayer.
 	// Let's keep it disabled to ease debug.
 	m_nativeres = theApp.GetConfigI("upscale_multiplier") == 1 || GLLoader::in_replayer;
-	m_mipmap = theApp.GetConfigI("mipmap");
```
turtleli: I don't think this file and GSRendererSW.cpp actually need to be changed. It'll also make the OpenCL renderer always start with mipmapping enabled.

Apart from that, the code seems fine.

plugins/GSdx/GSdx.cpp
```diff
@@ -303,6 +307,7 @@ void GSdxApp::Init()
 	m_default_configuration["linear_present"]                             = "1";
 	m_default_configuration["MaxAnisotropy"]                              = "0";
 	m_default_configuration["mipmap"]                                     = "1";
+	m_default_configuration["mipmap_hw"]                                  = "1";
```
turtleli: Actually, what's the default meant to be?

ssakash: @gregory38 : 0 or 1?

gregory38:
0 for hw renderer

gregory38:
ok fine. I will do the linux part later

PR #1626, opened 2016-10-16
gsdx ogl: tekken5: only trigger channel effect shader when frame buffer is a single page

gregory38:
Fix missing model regression.

Please test tekken5

### DIFF
--- a/plugins/GSdx/GSRendererOGL.cpp
+++ b/plugins/GSdx/GSRendererOGL.cpp
@@ -483,13 +483,18 @@ void GSRendererOGL::EmulateChannelShuffle(GSTexture** rt, const GSTextureCache::
 			m_context->TEX0.TFX = TFX_DECAL;
 			*rt = tex->m_from_target;
 		} else if (m_game.title == CRC::Tekken5) {
-			GL_INS("Tekken5 RGB Channel");
-			m_ps_sel.channel = 7;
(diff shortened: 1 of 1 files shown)


bositman:
Seems to be working fine now even on CRC None/minimum/partial

gregory38:
You need at least minimum for some effects in some stages (...moon...).

bositman:
I tested all this with CRC minimum
Oh if you mean all stages issues, some are still there, like the Burning Temple issue (horizontal line on character's bodies) and additional issues if using scaling.
pcsx2 2016-10-16 13-18-42-18

You need both align sprite for vertical black lines to get fixed and half pixel offset when upscaling to fix ghosting, but you also get this then:
pcsx2 2016-10-16 13-21-37-64

gregory38:
Good. Those are known issues. The first one is the typical half bottom issue. For the 2nd one, I need to find a way to merge properly sprite for processing effect. Or maybe I could just round them.

FlatOutPS2:
The characters are back, but now there are all kinds of shadowy things moving accross the screen.

gregory38:
which scene ? I didn't notice it on the missing model dump. However, I'm not sure the effect is correctly rendered

FlatOutPS2:
Never mind, the issue only appears with the default Full CRC hack level.

gregory38:
Cool :+1:

PR #1630, opened 2016-10-17
Fix PCSX2 on Wayland

frantisekz:
Require XWayland in GTK3 build by setting GDK_BACKEND=x11


gregory38:
Thanks.

DmitryHetman:
Can it work under wayland natively?

gregory38:
nope. Wayland isn't ready for prime time anyway. And honestly Wayland is nice for desktop but it will only bring limitation for OpenGL application (as PCSX2).

PR #1633, opened 2016-10-20
PSX-mode: Detect video mode via colorburst

ssakash:
**Summary of changes**:
- Improve PSX video mode detection by using info from colorburst at certain cases. `( Might help some scenarios on PSX games where PAL/NTSC video mode was improperly set to a wrong limit instead of it's actual vertical frequency limit.)`

@FlatOutPS2 

IIRC you had mentioned some PSX games didn't have a proper limit of 50fps when using PAL video mode, could you check the behavior of this patch with such games? Thank you :)

### DIFF
--- a/pcsx2/GS.cpp
+++ b/pcsx2/GS.cpp
@@ -25,6 +25,7 @@
 using namespace Threading;
 using namespace R5900;
 
+static GS_VideoMode s_ColorBurst = GS_VideoMode::Uninitialized;
 __aligned16 u8 g_RealGSMem[Ps2MemSize::GSregs];
 
 void gsOnModeChanged( Fixed100 framerate, u32 newTickrate )
@@ -35,9 +36,12 @@ void gsOnModeChanged( Fixed100 framerate, u32 newTickrate )
 
 void gsSetVideoMode(GS_VideoMode mode )
 {
-	if( gsVideoMode == mode ) return;
+	if( gsVideoMode == mode )
+		return;
 
-	gsVideoMode = mode;
+	// SetGsCrt doesn't seem to work for PSX games and they're left with the initial bios mode
+	// At such cases, let's use the colorburst to detect the video mode.
+	gsVideoMode = (mode == GS_VideoMode::BIOS) ? s_ColorBurst : mode;
 	UpdateVSyncRate();
 }
 
@@ -138,6 +142,14 @@ __fi void gsWrite8(u32 mem, u8 value)
 	GIF_LOG("GS write 8 at %8.8lx with data %8.8lx", mem, value);
 }
 
+static void GetColorBurst(u64 value)
+{
+	GSRegSMODE1 Register;
+	Register.SMODE1 = value;
+	if(Register.CMOD)
+		s_ColorBurst = (Register.CMOD == 3) ? GS_VideoMode::PAL : GS_VideoMode::NTSC;
+}
+
 //////////////////////////////////////////////////////////////////////////
 // GS Write 16 bit
 
@@ -201,6 +213,8 @@ void __fastcall gsWrite64_generic( u32 mem, const mem64_t* value )
 
 void __fastcall gsWrite64_page_00( u32 mem, const mem64_t* value )
 {
+	if (mem == GS_SMODE1)
+		GetColorBurst(*value);
 	gsWrite64_generic( mem, value );
 }
 
--- a/pcsx2/GS.h
+++ b/pcsx2/GS.h
@@ -199,6 +199,41 @@ union tGS_IMR
 };
 
 // --------------------------------------------------------------------------------------
+//  GSRegSMODE1
+// --------------------------------------------------------------------------------------
+// Currently it's only used to get the CMOD bit from the SMODE1 register value.
+union GSRegSMODE1
+{
+	struct
+	{
+		u32 RC : 3;
+		u32 LC : 7;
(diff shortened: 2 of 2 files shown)


FlatOutPS2:
As far as I know all PAL PSX games don't have the proper limit of 50fps. ;)

ssakash:
@FlatOutPS2 

I pushed some bug fixes, could you do a re-test on the PSX games?

gregory38:
Hum it would be nice to understand the ps2 kernel code. Gs register can only be written by the kernel. Maybe they use another syscall. Note we might have wrong reset value too. I would suggest to put breakpoint on those gs write. Find the EE pc and check behavior with internal debugger.

ramapcsx2:
As discussed elsewhere, we found out that a PAL BIOS currently wrongly runs in 60Hz. Fixing this has priority and even with no other change, this will put PAL PSX games into 50Hz when run through a PAL BIOS. This was the only supported way on the real machine as well.

gregory38:
So I guess a pal (ntsc) bios will write the register pal (ntsc) value during the init. What happen if you run a pal game on a ntsc bios (or the contrary) ? does it work?

pcsx2/GS.cpp
```diff
@@ -138,13 +139,24 @@ __fi void gsWrite8(u32 mem, u8 value)
...
 }
 
+static void GetColorBurst(u32 mem, u32 value)
+{
+	if (mem != GS_SMODE1)
+		return;
+
+	GSRegSMODE1 Register;
+	Register.SMODE1 = value;
+	if(u32 colorburst = Register.CMOD)
```
gregory38: I don't like the mix of condition and affection. By the way, you can easily write
if (CMOD == 3)
 ... PAL
else if (CMOD != 0)
... NTSC

pcsx2/GS.cpp
```diff
@@ -138,13 +139,24 @@ __fi void gsWrite8(u32 mem, u8 value)
...
+}
+
 //////////////////////////////////////////////////////////////////////////
 // GS Write 16 bit
 
 __fi void gsWrite16(u32 mem, u16 value)
 {
 	GIF_LOG("GS write 16 at %8.8lx with data %8.8lx", mem, value);
-
+	GetColorBurst(mem, value);
```
gregory38: add a case on the switch

pcsx2/GS.cpp
```diff
@@ -173,7 +185,7 @@ __fi void gsWrite32(u32 mem, u32 value)
 {
 	pxAssume( (mem & 3) == 0 );
 	GIF_LOG("GS write 32 at %8.8lx with data %8.8lx", mem, value);
-
+	GetColorBurst(mem, value);
```
gregory38: add case on the switch

FlatOutPS2:
> What happen if you run a pal game on a ntsc bios (or the contrary) ? does it work?

Region lock has not been passed, so you currently can't boot PSX games from other regions.

gregory38:
ok. So basically it won't work.

pcsx2/GS.cpp
ssakash: How is it better compared to just calling ``GetColorBurst()``?

gregory38: First you won't need to add the mem parameter. Then you understand that GetColorBurst is only for handling a single register and not a generic function that could impact everything.

gregory38: I didn't say to duplicate code. But instead to call the function from the switch case. It is far more logic this way. Instead to always check the address in the function.

FlatOutPS2:
Tried the latest change, but I still get 3 video mode changes in a row.

gregory38:
Well potentially 16/32 bits access are wrong as the code only intercept the lsb. And by the way, 128 bits writes are unlikely as it will 2 registers at the same time.

refractionpcsx2:
> Well potentially 16/32 bits access are wrong as the code only intercept the lsb

I did mention that to him, he said he was aware, dunno why he didn't fix it :P

ssakash:
> Tried the latest change, but I still get 3 video mode changes in a row.

Pushed a new change to only use the 64bit write, could you check the status of the frequent video mode changes? I'm afraid I can't make it any better.

FlatOutPS2:
> Pushed a new change to only use the 64bit write, could you check the status of the frequent video mode changes? I'm afraid I can't make it any better.

Good news, you don't have to. It's now back to one change. :+1:

pcsx2/GS.cpp
```diff
@@ -202,6 +217,7 @@ void __fastcall gsWrite64_generic( u32 mem, const mem64_t* value )
 void __fastcall gsWrite64_page_00( u32 mem, const mem64_t* value )
 {
 	gsWrite64_generic( mem, value );
+	GetColorBurst(mem, static_cast<u32>(value[0]));
```
gregory38: I would prefer to have the mem check outside of the function
```
if (mem == GS_SMODE1)
    GetColorBurst(.......)
```
It avoid a parameter.

And I don't like the cast here. `[0]` kinds of imply an array where it is a plain pointer. So a `*` might be better. That being said, maybe you could just use a 64 bits value for the gs register.

ssakash: > That being said, maybe you could just use a 64 bits value for the gs register.

Nice catch 👍

gregory38:
I think we are good now, aren't we ?

PR #1635, opened 2016-10-21
SPU2-X: Reverb rewrite

sudonim1:
This is based on http://problemkaputt.de/psx-spx.htm and http://drhell.web.fc2.com/ps1/ by Martin Korth and "Dr. Hell" respectively, who mostly agree on the reverb engine's inner workings.  The whole thing, especially the old implementation, is too tangled for me to say how it differs, I just rewrote it to the new spec.

This is definitely still not correct, and testing makes me suspect the cross-channel reflection ("DIFF") code.  Actual test programs need to be written, which is very unpleasant because all the SPU code has to be done on the IOP and just designing good investigatory tests for this is difficult.  Nonetheless, the improvement is marked.

I split the work into a functional commit and a register renaming commit for ease of review while rebasing.


FlatOutPS2:
inux builds will also fail on this PR due to clang-format issue.

sudonim1:
I used the clang format plugin.  What do you want from me, blood?

sudonim1:
Actually the FIR might be there, it turns out the implementation I ripped out has a signedness error, doing an unsigned division on samples with predictably awful results.  I'm getting close results now I'm experimenting with it, but I haven't found a FIR length that's right yet (if one exists).

sudonim1:
Correction: I have no idea any more.  I had a version which came very close to PS2 output but I lost it and don't know what I did any more.

ramapcsx2:
This is still a tremendous improvement. Any title making good use of reverb will sound better now, but it especially benefits PSX :)

sudonim1:
Since opening this I think I've worked out that all the significant difference from the old code is the two all pass filters.  I have no idea how a fast digital all pass filter is meant to work so both versions are greek to me though.  The remaining issues are probably either in a missing input filter or in the uh whatever you call the simple filter used for same/diff.

ramapcsx2:
It would be best if we do another round of hardware reverb samples. This time with different waveforms.
I imagine a square wave of left channel only, right channel only and both channels.

sudonim1:
I'm personally inclined to merge it soon if there are no objections since I don't know when I'm going to get around to figuring out the remaining differences from real hardware.

ramapcsx2:
No objections. Go ahead please :)

sudonim1:
Okay I guess nothing's broken or got noticeably worse.

PR #1640, opened 2016-10-27
Updates locales/templates readme

AniLeo:
Replaces the broken wiki link for the correct one
(Latest line wasn't changed, newline wasn't added/removed)


refractionpcsx2:
Thanks for that :)

Issue #1570, opened 2016-09-17
French Translation Missing for Boot CDVD (Full)


Catarax:
**Latest build (No French Translation)**
![screenshot_1](https://cloud.githubusercontent.com/assets/3964507/18607330/2c9a63ce-7cca-11e6-828e-7158817e085e.png)

**PCSX2 1.4 (correct French Translation)**
![screenshot_2](https://cloud.githubusercontent.com/assets/3964507/18607337/3fecf7a2-7cca-11e6-88b6-ff53da57e5a1.png)

mirh:
It's not just French. 
And.. I dunno, if you have the time you could try to bisect this. 26 February build was still fine for example.

gregory38:
just likely need to update the translation file. git build aren't official release so translation can be a bit off.

mirh:
Bisected, this was introtuced in #1388.

FlatOutPS2:
It's because this was expected to be merged: https://github.com/PCSX2/pcsx2/pull/1381/commits
But it never happened and I guess the translations weren't updated again afterwards.

gregory38:
Translation must be unfuzzed but I think we are good now.The initial string is value so I'm closing this one.

Issue #1620, opened 2016-10-14
WX/GUI: Discuss possible string changes


MrCK1:
This issue is intended as a continuation of PR #1577 as turtleli suggested. If you'd like to see any strings changed, you can suggest them here

So far we have:
- [x]  Fix "Shutdown" menu state if necessary
- [x]  Change `Iso` to `ISO` (turtleli's suggestion)
- [x]  Resolve boot medium GUI behavior (#1381)
- [x]  Remove `HostFs` from GUI on release builds (user can still edit value via pcsx2_vm)
- [ ]  Reduce string size of gamefix panel help text.
- [x]  ~~Remove tooltip of ``Restore Defaults`` in speedhacks panel.~~ **(done by turtleli)**

FlatOutPS2:
We should also still add part of this PR: https://github.com/PCSX2/pcsx2/pull/1381

For this behaviour: Properly display the boot medium on menu item Example: ISO , CDVD , BIOS

piorrro33:
I personally am perfectly fine with all these changes, except that I would personally think that the Debug tab should be removed by default and may be enabled manually through an edit in a config file, or maybe simply shift it to the Misc. menu?

gregory38:
```
Remove Host Filesystem option from menu - It's rarely used besides homebrew/demos and any user who needs to use it can just change the HostFs value in PCSX2_vm
```
Keep it on dbg/dev build.

MrCK1:
Right now we're discussing other strings that could be changed in issue #1620 as you probably saw already. Until then this sits

FlatOutPS2:
Why are you referencing this issue from within this issue? :S

piorrro33:
I think he meant #1381.

jcdenton2k:
I would also like to see Iso changed to ISO :)

gregory38:
What is the status? If you want to update string it is now

piorrro33:
Seems fine to me, but I'd add my comment above too, remove the Debug tab by default, since it's useless for most.

ssakash:
@MrCK1 

Are you planning on doing these [suggested changes](https://github.com/PCSX2/pcsx2/issues/1620#issue-183052252) at #1577? I could handle it if you're busy. Sorry for the rush, as these changes should be handled before the translation update and @gregory38 seems to be in a bit of hurry for the 1.6 release :P

piorrro33:
> Rather unlikely to happen.
[...]

PR #1644, opened 2016-10-30
Fixed French translation

TotalCaesar659:
Pull request for solving https://github.com/PCSX2/pcsx2/issues/1570

### DIFF
--- a/locales/fr_FR/pcsx2_Main.po
+++ b/locales/fr_FR/pcsx2_Main.po
@@ -1645,7 +1645,7 @@ msgid "Hard reset of the active VM."
 msgstr "Redémarrage intégral du système d'émulation."
 
 #: pcsx2/gui/MainFrame.cpp:655
-msgid "Boot CDVD (full)"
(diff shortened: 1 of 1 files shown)


gregory38:
The solution is to run the script to update all translations.

FlatOutPS2:
> The solution is to run the script to update all translations.

Would be nice if we could do these updates first: https://github.com/PCSX2/pcsx2/issues/1620

gregory38:
Please do a PR for all string changes. And please do it soon, otherwise it will be postponed after the release.

gregory38:
I will integrate the PR if automatic translation update doesn't fix it.

TotalCaesar659:
Sorry, I was busy and didn't read it. Do I have to check and fix this string in every language translation?

TotalCaesar659:
OK, thank you very much.

gregory38:
I updated all po file with latest string. It might need to fix a couple of fuzzy strings. Otherwise I think we can close this one.

PR #1646, opened 2016-11-01
3rd Party: Update glext header files to the current revision 

MrCK1:
Our files are a little over a year old; so let's update ours to the newest revision that was released a few days ago. 

The new revision number changes from 31811 (08-10-2015) to 33248 (10-24-2016)


gregory38:
Fwiw, the update is completely useless as I need to keep compatibility with older header provided in older linux distribution.

PR #1647, opened 2016-11-04
GSDX: Clean up warnings on MSVC

ssakash:
**Summary of changes**:
* Explicitly cast w_pages and h_pages into uint32.
* Prevent signed/unsigned comparison by converting lod into unsigned integer, honestly how coud a mipmapping level be negative?

On the other hand, don't really care much about the warnings. They were just annoying as they kept popping up all the time I rebuild GSdx.


gregory38:
Hum, need to check the warning on the linux build bot ;)
```
 honestly how coud a mipmapping level be negative?
```
If by mipmap level you mean the lod, yes It could be negative. It mean that texture will be expanded (bilinear filtering) instead of reduced (mipmap filtering).

PR #1649, opened 2016-11-06
IOP: PSX bios console output filtering of repeated and empty lines.

sudonim1:
Pretty ghastly code imo, also clearly in the wrong place, this should be a generic per source thing in the console, but @ramapcsx2 wants this badly apparently?


ramapcsx2:
Appears to work very well. Thanks :)

PR #1651, opened 2016-11-06
PCSX2-WX: Display boot medium, fix shutdown menu behavior and some string updates

### LINKED ISSUE #1620: WX/GUI: Discuss possible string changes

ssakash:
**Summary of changes**:

* Fix shutdown menu item behavior (Thanks to @MrCK1 for looking into it)
* Disable Host Filesystem menu item for release builds as it isn't generally useful for endusers.
* Display proper boot medium on the menu items.
* Some string updates, here and there.

~~``[WIP]`` I still need to update https://github.com/PCSX2/pcsx2/commit/80ed0c367f8d46c30aa89fdefd8c93fd0c130bfd at a later time, other than that code is complete.~~ **Done**

Gregory suggested refraction to reduce the gamefix panel help string size, so #1620 could probably be closed after that I guess.

### DIFF
--- a/common/include/Utilities/IniInterface.h
+++ b/common/include/Utilities/IniInterface.h
@@ -74,10 +74,11 @@ public:
     void EnumEntry(const wxString &var, T &value, const wxChar *const *enumArray = NULL, const T defvalue = (T)0)
     {
         int tstore = (int)value;
+        auto defaultvalue = enum_cast(defvalue);
         if (enumArray == NULL)
-            Entry(var, tstore, defvalue);
+            Entry(var, tstore, defaultvalue);
         else
-            _EnumEntry(var, tstore, enumArray, defvalue);
+            _EnumEntry(var, tstore, enumArray, defaultvalue);
         value = (T)tstore;
     }
 
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -461,17 +461,17 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
 	// FIXME: VS2015 thinks there's a memory issue here and there probably is one.
 	// The submenu is owned by a unique_ptr, but any menu that is attached to a
 	// menubar or another menu will be deleted by its parent.
-	m_menuCDVD.Append( MenuId_IsoSelector,	_("Iso &Selector"), &isoRecents );
+	m_menuCDVD.Append( MenuId_IsoSelector,	_("ISO &Selector"), &isoRecents );
 	m_menuCDVD.Append( GetPluginMenuId_Settings(PluginId_CDVD), _("Plugin &Menu"), m_PluginMenuPacks[PluginId_CDVD] );
 
 	m_menuCDVD.AppendSeparator();
-	m_menuCDVD.Append( MenuId_Src_Iso,		_("&Iso"),		_("Makes the specified ISO image the CDVD source."), wxITEM_RADIO );
+	m_menuCDVD.Append( MenuId_Src_Iso,		_("&ISO"),		_("Makes the specified ISO image the CDVD source."), wxITEM_RADIO );
 	m_menuCDVD.Append( MenuId_Src_Plugin,	_("&Plugin"),	_("Uses an external plugin as the CDVD source."), wxITEM_RADIO );
 	m_menuCDVD.Append( MenuId_Src_NoDisc,	_("&No disc"),	_("Use this to boot into your virtual PS2's BIOS configuration."), wxITEM_RADIO );
 
 	//m_menuCDVD.AppendSeparator();
 	//m_menuCDVD.Append( MenuId_SkipBiosToggle,_("Enable BOOT2 injection"),
-	//	_("Skips PS2 splash screens when booting from Iso or DVD media"), wxITEM_CHECK );
+	//	_("Skips PS2 splash screens when booting from ISO or DVD media"), wxITEM_CHECK );
 
 	// ------------------------------------------------------------------------
 
@@ -598,14 +598,12 @@ void MainEmuFrame::ApplyCoreStatus()
 	wxMenuBar& menubar( *GetMenuBar() );
 
 	wxMenuItem* susres	= menubar.FindItem( MenuId_Sys_SuspendResume );
-	wxMenuItem* cdvd	= menubar.FindItem( MenuId_Boot_CDVD );
-	wxMenuItem* cdvd2	= menubar.FindItem( MenuId_Boot_CDVD2 );
 	wxMenuItem* restart	= menubar.FindItem( MenuId_Sys_Restart );
 
 	// [TODO] : Ideally each of these items would bind a listener instance to the AppCoreThread
 	// dispatcher, and modify their states accordingly.  This is just a hack (for now) -- air
 
-	bool vm = SysHasValidState();
+	bool ActiveVM = SysHasValidState();
 
 	if( susres )
 	{
@@ -617,8 +615,8 @@ void MainEmuFrame::ApplyCoreStatus()
 		}
 		else
 		{
-			susres->Enable(vm);
-			if( vm )
+			susres->Enable(ActiveVM);
+			if( ActiveVM )
 			{
 				susres->SetItemLabel(_("R&esume"));
 				susres->SetHelp(_("Resumes the suspended emulation state."));
@@ -633,7 +631,7 @@ void MainEmuFrame::ApplyCoreStatus()
 
 	if( restart )
 	{
-		if( vm )	
+		if( ActiveVM )
 		{
 			restart->SetItemLabel(_("Res&tart"));
 			restart->SetHelp(_("Simulates hardware reset of the PS2 virtual machine."));
@@ -645,31 +643,64 @@ void MainEmuFrame::ApplyCoreStatus()
 		}
 	}
 
-	if( cdvd )
+	const CDVD_SourceType Source = g_Conf->CdvdSource;
+	const MenuIdentifiers fullboot_id = MenuId_Boot_CDVD;
+	const MenuIdentifiers fastboot_id = MenuId_Boot_CDVD2;
+
+	wxMenuItem *cdvd_fast = menubar.FindItem(fastboot_id);
+	if (Source == CDVD_SourceType::NoDisc)
 	{
-		if( vm )
+		if(cdvd_fast)
+			m_menuSys.Destroy(cdvd_fast);
+	}
+	else
+	{
+		wxString label;
+		wxString help_text = _("Use fast boot to skip PS2 startup and splash screens");
+
+		switch (Source)
 		{
-			cdvd->SetItemLabel(_("Reboot CDVD (f&ull)"));
-			cdvd->SetHelp(_("Hard reset of the active VM."));
+		case CDVD_SourceType::Iso:
+			label = _("Boot ISO (&fast)");
+			break;
+		case CDVD_SourceType::Plugin:
+			label = _("Boot CDVD (&fast)");
+			break;
+		//case CDVD_SourceType::NoDisc: (Fast boot menu item is destroyed when no disc is selected)
+		default:
+			pxAssert(false);
+		}
+
+		if (cdvd_fast)
+		{
+			cdvd_fast->SetItemLabel(label);
+			cdvd_fast->SetHelp(help_text);
 		}
 		else
 		{
-			cdvd->SetItemLabel(_("Boot CDVD (f&ull)"));
-			cdvd->SetHelp(_("Boot the VM using the current DVD or Iso source media"));
-		}	
+			m_menuSys.Insert(1, fastboot_id, label, help_text);
+		}
 	}
 
-	if( cdvd2 )
+	if (wxMenuItem *cdvd_full = menubar.FindItem(fullboot_id))
 	{
-		if( vm )
+
+		switch (Source)
 		{
-			cdvd2->SetItemLabel(_("Reboot CDVD (&fast)"));
-			cdvd2->SetHelp(_("Reboot using fast BOOT (skips splash screens)"));
-		}
-		else
-		{
-			cdvd2->SetItemLabel(_("Boot CDVD (&fast)"));
-			cdvd2->SetHelp(_("Use fast boot to skip PS2 startup and splash screens"));
+		case CDVD_SourceType::Iso:
+			cdvd_full->SetItemLabel(_("Boo&t ISO (full)"));
+			cdvd_full->SetHelp(_("Boot the VM using the current ISO source media"));
+			break;
+		case CDVD_SourceType::Plugin:
+			cdvd_full->SetItemLabel(_("Boo&t CDVD (full)"));
+			cdvd_full->SetHelp(_("Boot the VM using the current CD/DVD source media"));
+			break;
+		case CDVD_SourceType::NoDisc:
+			cdvd_full->SetItemLabel(_("Boo&t BIOS"));
+			cdvd_full->SetHelp(_("Boot the VM without any source media"));
+			break;
+		default:
+			pxAssert(false);
 		}
 	}
 }
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -448,6 +448,7 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
 
 	m_menuSys.Append(MenuId_Sys_Shutdown,	_("Shut&down"),
 		_("Wipes all internal VM states and shuts down plugins."));
+	m_menuSys.FindItem(MenuId_Sys_Shutdown)->Enable(false);
 
 	m_menuSys.Append(MenuId_Exit,			_("E&xit"),
 		AddAppName(_("Closing %s may be hazardous to your health")));
@@ -671,8 +672,6 @@ void MainEmuFrame::ApplyCoreStatus()
 			cdvd2->SetHelp(_("Use fast boot to skip PS2 startup and splash screens"));
 		}
 	}
-
-	menubar.Enable( MenuId_Sys_Shutdown, SysHasValidState() || CorePlugins.AreAnyInitialized() );
 }
 
 //Apply a config to the menu such that the menu reflects it properly
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -57,14 +57,14 @@ void MainEmuFrame::UpdateIsoSrcSelection()
 
 	switch( g_Conf->CdvdSource )
 	{
-		case CDVDsrc_Iso:		cdsrc = MenuId_Src_Iso;		break;
-		case CDVDsrc_Plugin:	cdsrc = MenuId_Src_Plugin;	break;
-		case CDVDsrc_NoDisc:	cdsrc = MenuId_Src_NoDisc;	break;
+		case CDVD_SourceType::Iso:		cdsrc = MenuId_Src_Iso;		break;
+		case CDVD_SourceType::Plugin:	cdsrc = MenuId_Src_Plugin;	break;
+		case CDVD_SourceType::NoDisc:	cdsrc = MenuId_Src_NoDisc;	break;
 
 		jNO_DEFAULT
 	}
 	sMenuBar.Check( cdsrc, true );
-	m_statusbar.SetStatusText( CDVD_SourceLabels[g_Conf->CdvdSource], 1 );
+	m_statusbar.SetStatusText( CDVD_SourceLabels[enum_cast(g_Conf->CdvdSource)], 1 );
 
 	EnableCdvdPluginSubmenu( cdsrc == MenuId_Src_Plugin );
 
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -440,8 +440,9 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
 	m_menuSys.Append(MenuId_EnableWideScreenPatches,	_("Enable &Widescreen Patches"),
 		wxEmptyString, wxITEM_CHECK);
 
-	m_menuSys.Append(MenuId_EnableHostFs,	_("Enable &Host Filesystem"),
-		wxEmptyString, wxITEM_CHECK);
+	if(IsDebugBuild || IsDevBuild)
+		m_menuSys.Append(MenuId_EnableHostFs,	_("Enable &Host Filesystem"),
+			wxEmptyString, wxITEM_CHECK);
 
 	m_menuSys.AppendSeparator();
 
--- a/pcsx2/CDVD/CDVDaccess.cpp
+++ b/pcsx2/CDVD/CDVDaccess.cpp
@@ -38,7 +38,7 @@
 
 const wxChar* CDVD_SourceLabels[] =
 {
-	L"Iso",
+	L"ISO",
 	L"Plugin",
 	L"NoDisc",
 	NULL
--- a/pcsx2/CDVD/CDVDaccess.cpp
+++ b/pcsx2/CDVD/CDVDaccess.cpp
@@ -285,11 +285,11 @@ static void DetectDiskType()
 }
 
 static wxString			m_SourceFilename[3];
-static CDVD_SourceType	m_CurrentSourceType = CDVDsrc_NoDisc;
+static CDVD_SourceType	m_CurrentSourceType = CDVD_SourceType::NoDisc;
 
 void CDVDsys_SetFile( CDVD_SourceType srctype, const wxString& newfile )
 {
-	m_SourceFilename[srctype] = newfile;
+	m_SourceFilename[enum_cast(srctype)] = newfile;
 
 	// look for symbol file
 	if (symbolMap.IsEmpty())
@@ -309,7 +309,7 @@ void CDVDsys_SetFile( CDVD_SourceType srctype, const wxString& newfile )
 
 const wxString& CDVDsys_GetFile( CDVD_SourceType srctype )
 {
-	return m_SourceFilename[srctype];
+	return m_SourceFilename[enum_cast(srctype)];
 }
 
 CDVD_SourceType CDVDsys_GetSourceType()
@@ -323,15 +323,15 @@ void CDVDsys_ChangeSource( CDVD_SourceType type )
 	
 	switch( m_CurrentSourceType = type )
 	{
-		case CDVDsrc_Iso:
+		case CDVD_SourceType::Iso:
 			CDVD = &CDVDapi_Iso;
 		break;
 
-		case CDVDsrc_NoDisc:
+		case CDVD_SourceType::NoDisc:
 			CDVD = &CDVDapi_NoDisc;
 		break;
 
-		case CDVDsrc_Plugin:
+		case CDVD_SourceType::Plugin:
 			CDVD = &CDVDapi_Plugin;
 		break;
 
@@ -354,8 +354,9 @@ bool DoCDVDopen()
 	// question marks if the filename is another language.
 	// Likely Fix: Force new versions of CDVD plugins to expect UTF8 instead.
 
-	int ret = CDVD->open( !m_SourceFilename[m_CurrentSourceType].IsEmpty() ?
-		static_cast<const char*>(m_SourceFilename[m_CurrentSourceType].ToUTF8()) : (char*)NULL
+	auto CurrentSourceType = enum_cast(m_CurrentSourceType);
+	int ret = CDVD->open( !m_SourceFilename[CurrentSourceType].IsEmpty() ?
+		static_cast<const char*>(m_SourceFilename[CurrentSourceType].ToUTF8()) : (char*)NULL
 	);
 
 	if( ret == -1 ) return false;	// error! (handled by caller)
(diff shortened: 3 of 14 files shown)


sudonim1:
Hmm, not sure I agree on host:, it's for homebrew in general and I was just now looking at finishing the work on it so getstat(), dopen() and dread() can be used (catch: the structures for these vary with ioman versions, but this is simpler to deal with than I remembered), which are what break most homebrew.  And besides we use it for running test programs and we kind of need to run tests on release builds....

FlatOutPS2:
You can still enable it through the ini, so you can still use it in release builds.

If we do keep it, we could consider greying out the option when preset is enabled.

sudonim1:
Well it's basically harmless, it doesn't do anything unless a program tries to open a host: path

sudonim1:
Uh I SAY it's harmless but upon reviewing it seems that file writing was added without proper safety in path translation.  That's not harmless at all, for all we know thousands of users could've been rooted.

common/include/Utilities/IniInterface.h
```diff
@@ -70,10 +70,11 @@ class IniInterface
 	void EnumEntry( const wxString& var, T& value, const wxChar* const* enumArray=NULL, const T defvalue=(T)0 )
 	{
 		int tstore = (int)value;
+		int defaultvalue = (int)defvalue;
```
gregory38: why? To support the enum class? But why did you convert cdvd source to an enum class, actually ?

ssakash: Well, I remember a previous discussion where you had mentioned to consider using``enum class`` over normal enums at all scenarios to prevent scope/name classes. Also personally, I prefer ``enum class`` over ``enum``. ``( Yes, I noticed the pseudo scope but I still feel enum classes are better)``

Though honestly what's wrong with the modification on ``IniInterface.h``? It should also work for other enums as they have an underlying type of ``int``.

gregory38: Well, I don't like all those verbose cast, beside it is error-prone on the type.

First, we (you ;) ) need to add a function to ease the cast. Here some example from stack overflow. For the naming, I took `as_int` but there are various possibillities.
```
template <typename Enumeration>
auto as_int(Enumeration const value)
    -> typename std::underlying_type<Enumeration>::type
{
    return static_cast<typename std::underlying_type<Enumeration>::type>(value);
}

template <typename E>
constexpr typename std::underlying_type<E>::type as_int(E e) {
    return static_cast<typename std::underlying_type<E>::type>(e);
}

template <typename E>
constexpr auto as_int(E e) noexcept
{
    return static_cast<std::underlying_type_t<E>>(e);
}
```

Then EnumEntry uses a template type. However you will always cast to an `int` which defeat the purpose of the template. Boolean variable will be stored as int. I'm pretty sure we can find a better solution (SFINAE), basically a template that call EnumEntry(..., as_int(value), ....)

pcsx2/gui/MainFrame.cpp
```diff
@@ -637,35 +639,51 @@ void MainEmuFrame::ApplyCoreStatus()
...
+	SourceLabel[static_cast<uint8_t>(CDVD_SourceType::Iso)] = _("ISO");
+	SourceLabel[static_cast<uint8_t>(CDVD_SourceType::Plugin)] = _("CDVD");
+	SourceLabel[static_cast<uint8_t>(CDVD_SourceType::NoDisc)] = _("BIOS");
+
+	BootIndicator[0][0] = _("Boot");
+	BootIndicator[0][1] = _("Reboot");
+	BootIndicator[1][0] = _(" (f&ull)");
+	BootIndicator[1][1] = _(" (&fast)");
+
+	wxString MenuItemLabel = static_cast<wxString>(BootIndicator[0][ActiveVM]) + " " + static_cast<wxString>(SourceLabel[static_cast<uint8_t>(Source)]);
```
turtleli: Some languages don't really use spaces (Chinese), so the `" "` shouldn't be placed here. Actually, breaking up a translatable string into multiple parts isn't a great idea - different languages have different word orders.

Also, those casts are unnecessary when you could just do `wxString(whatever)`.

common/include/Utilities/IniInterface.h
ssakash: I'm not sure I fully understand your idea, so basically you want to replace the static casts with a template function call? ``as_int(enum)`` instead of ``static_cast<uint8>(enum)`` to directly cast to the underlying type of the enum?

pcsx2/gui/MainFrame.cpp
ssakash: > Some languages don't really use spaces (Chinese), so the " " shouldn't be placed here. 

I see, didn't know you were from China. I thought you were from UK based on your wordings on replies but guess I was wrong :P 

> Actually, breaking up a translatable string into multiple parts isn't a great idea - different languages have different word orders.

Fair enough, I'll fix it ASAP along with the ``" " `` thing on few translations.

ssakash:
FWIW, I'm not around a build environment right now. I'll probably handle these changes on Sunday, if anyone wants to get this merged before that time, feel free to add commits to add the changes suggested by turtleli and Gregory.

 I have opened this branch on the PCSX2 repo so you guys should be able to push changes on it.

pcsx2/gui/MainFrame.cpp
turtleli: > I see, didn't know you were from China. I thought you were from UK based on your wordings on replies but guess I was wrong :P

Last I checked I was from the UK. XD

My second language does happen to be Chinese though (Cantonese/Traditional Chinese).

pcsx2/gui/MainFrame.cpp
```diff
@@ -437,13 +437,15 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
...
-		wxEmptyString, wxITEM_CHECK);
+	if(IsDebugBuild || IsDevBuild)
+		m_menuSys.Append(MenuId_EnableHostFs,	_("Enable &Host Filesystem"),
+			wxEmptyString, wxITEM_CHECK);
 
 	m_menuSys.AppendSeparator();
 
 	m_menuSys.Append(MenuId_Sys_Shutdown,	_("Shut&down"),
 		_("Wipes all internal VM states and shuts down plugins."));
+	m_menuSys.FindItem(MenuId_Sys_Shutdown)->Enable(false);
```
MrCK1: So I know it's not great to run external programs with PCSX2; but I've ran FRAPS with PCSX2 all the time (as an FPS meter) and I've never had any issues until now.  SuperVU fails to allocate memory if the First Time Wizard is run while a FRAPS process is active. After running the First Time Wizard, the plugins are shutdown right before virtual memory mapping occurs (if you reopen the .exe, this doesn't happen)
	```Shutdown DEV9 and so on``` ...
```Plugins shutdown successfully```
```Reserving memory for recompilers...```
```SuperVU0 Recompiler Cache: host memory @ 0x0E800000 -> 0x0F000000 is unavailable; attempting to map elsewhere...```
```SuperVU1 Recompiler Cache: host memory @ 0x0F000000 -> 0x0F800000 is unavailable; attempting to map elsewhere...```
	```CPU provider error:```
	```Virtual memory map failed while reserving 'SuperVU1 Recompiler Cache':```
	```SuperVU failed to allocate virtual memory below 256MB.```

Do you think this is related to any changes in this PR? I ran the FTW with an older build/FRAPS running; and had no problems with the sVU rec. It would be good to find out if other programs cause this as well, so we don't have any issues in the upcoming stable release.

Comment (MrCK1) [inline pcsx2/gui/MainFrame.cpp]: OK, sounds good ;)

**Comment by ssakash** [comment]
Updated the commit to add a template function for enumeration casts and integrated the strings into a single part as suggested by turtleli.

**Comment by FlatOutPS2** [comment]
It seems like VS2013 compilation is failing. :P

**Comment by ssakash** [comment]
Oh, I think I understood why VS2013 compilation is failing now. It seems VS2013 doesn't support ``constexpr`` https://msdn.microsoft.com/en-us/library/hh567368.aspx :/

**Comment by ssakash** [comment]
So, there are two options - let's remove support for VS2013 **(or)** drop the template idea casting of enumeration classes? ``constexpr`` is mandatory at this case as we need compile time constants for the array subscripts.

Honestly I'm in favor of dropping VS2013 if it prevents us from using new features like ``constexpr`` , ``noexcept`` and * **insert bunch of other cool features** *.

**Comment by refractionpcsx2** [comment]
Well 2013 is still current, so I'm against dropping support for that until VS 2016/2017 is out of beta. So maybe comment it out for now with a comment about using that once 2013 is dropped and for now drop the template idea.

**Comment by gregory38** [comment]
@refractionpcsx2 I'm not sure what you mean by beta but I think it is RC (and VS2017).
https://www.visualstudio.com/en-us/news/releasenotes/vs2017-relnotes

However, I will prefer to wait an official release.

**Comment by avih** [comment]
I'm also with 2013. Can't get myself to register with MS only to be able to compile with 2015...

**Comment by refractionpcsx2** [comment]
> However, I will prefer to wait an official release.

That's what I mean, I want to wait for the completely stable release before we jump in to it.

> I'm also with 2013. Can't get myself to register with MS only to be able to compile with 2015...

Well, you might have to soon :P

**Comment by gregory38** [comment]
@ssakash Where do you need compile time constant exactly ?

**Comment by ssakash** [comment]
> @ssakash Where do you need compile time constant exactly ?

As I said earlier,
> constexpr is mandatory at this case as **we need compile time constants for the array subscripts**.

Here's a list of the errors produced when ``constexpr`` is removed from the function.
```
.\..\gui\MainFrame.cpp(643): error C2131: expression did not evaluate to a constant
1>  ..\..\gui\MainFrame.cpp(643): note: failure was caused by non-constant arguments or reference to a non-constant symbol
1>  ..\..\gui\MainFrame.cpp(643): note: see usage of 'TotalSources'
1>..\..\gui\MainFrame.cpp(645): error C2131: expression did not evaluate to a constant
1>  ..\..\gui\MainFrame.cpp(645): note: failure was caused by non-constant arguments or reference to a non-constant symbol
1>  ..\..\gui\MainFrame.cpp(645): note: see usage of 'TotalSources'
1>..\..\gui\MainFrame.cpp(647): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(648): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(649): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(651): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(652): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(653): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(654): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(655): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(656): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(657): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(658): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(659): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(660): error C3863: array type 'const wxChar *[TotalSources]' is not assignable
1>..\..\gui\MainFrame.cpp(665): error C2664: 'void wxMenuItem::SetItemLabel(const wxString &)': cannot convert argument 1 from 'const wxChar *[TotalSources]' to 'const wxString &'
1>  ..\..\gui\MainFrame.cpp(665): note: Reason: cannot convert from 'const wxChar *[TotalSources]' to 'const wxString'
1>  ..\..\gui\MainFrame.cpp(665): note: No constructor could take the source type, or constructor overload resolution was ambiguous
1>..\..\gui\MainFrame.cpp(669): error C2664: 'void wxMenuItemBase::SetHelp(const wxString &)': cannot convert argument 1 from 'const wxChar *[TotalSources]' to 'const wxString &'
1>  ..\..\gui\MainFrame.cpp(669): note: Reason: cannot convert from 'const wxChar *[TotalSources]' to 'const wxString'
1>  ..\..\gui\MainFrame.cpp(669): note: No constructor could take the source type, or constructor overload resolution was ambiguous
1>..\..\gui\MainFrame.cpp(673): error C2664: 'void wxMenuBase::Insert(size_t,int,const wxString &,const wxString &,bool)': cannot convert argument 3 from 'const wxChar *[TotalSources]' to 'const wxString &'
1>  ..\..\gui\MainFrame.cpp(673): note: Reason: cannot convert from 'const wxChar *[TotalSources]' to 'const wxString'
1>  ..\..\gui\MainFrame.cpp(673): note: No constructor could take the source type, or constructor overload resolution was ambiguous
1>..\..\gui\MainFrame.cpp(675): error C2664: 'void wxMenuItem::SetItemLabel(const wxString &)': cannot convert argument 1 from 'const wxChar *[TotalSources]' to 'const wxString &'
1>  ..\..\gui\MainFrame.cpp(675): note: Reason: cannot convert from 'const wxChar *[TotalSources]' to 'const wxString'
1>  ..\..\gui\MainFrame.cpp(675): note: No constructor could take the source type, or constructor overload resolution was ambiguous
```

Only compile time constants can be used for declaring an array's size.

**Comment by ssakash** [comment]
Wait, _**what the heck?**_ So all I need to do is just replace a single instance of ``enum_cast`` with ``static_cast`` and there should be no problem as the array is the only thing which needs ``constexpr``. why the heck didn't I do just do that then? I'm an idiot! -_-

File: pcsx2/gui/MainFrame.cpp [hunk 4]
```diff
@@ -636,35 +638,49 @@ void MainEmuFrame::ApplyCoreStatus()
...
+	MenuItemLabels[0][0][enum_cast(CDVD_SourceType::Iso)]    = _("Boot ISO (f&ull)");
+	MenuItemLabels[0][1][enum_cast(CDVD_SourceType::Iso)]    = _("Boot ISO (&fast)");
+	MenuItemLabels[1][0][enum_cast(CDVD_SourceType::Iso)]    = _("Reboot ISO (f&ull)");
+	MenuItemLabels[1][1][enum_cast(CDVD_SourceType::Iso)]    = _("Reboot ISO (&fast)");
+	MenuItemLabels[0][0][enum_cast(CDVD_SourceType::Plugin)] = _("Boot CDVD (f&ull)");
+	MenuItemLabels[0][1][enum_cast(CDVD_SourceType::Plugin)] = _("Boot CDVD (&fast)");
+	MenuItemLabels[1][0][enum_cast(CDVD_SourceType::Plugin)] = _("Reboot CDVD (f&ull)");
+	MenuItemLabels[1][1][enum_cast(CDVD_SourceType::Plugin)] = _("Reboot CDVD (&fast)");
+	MenuItemLabels[0][0][enum_cast(CDVD_SourceType::NoDisc)] = _("Boot BIOS");
+	MenuItemLabels[1][0][enum_cast(CDVD_SourceType::NoDisc)] = _("Reboot BIOS");
```
refractionpcsx2: Does it matter too much? This isn't time critical code, afterall.

turtleli: Well, it's not optimal, and I think using switches and ifs would be nicer and have the additional benefit of not having to enum cast every single thing.

refractionpcsx2: I agree a switch would be nicer.

ssakash: > Maybe it'd be better not to use an array? This code does 13 translation lookups, but only uses up to 3 of the strings.

Maybe an array with static storage class so that the string lookup is done only once for each of them?

Don't really love the switch-case idea as I'd have to later duplicate that string's value on the insert function's argument.

pcsx2/gui/MainFrame.cpp
```diff
@@ -636,35 +638,49 @@ void MainEmuFrame::ApplyCoreStatus()
...
-			cdvd2->SetHelp(_("Reboot using fast BOOT (skips splash screens)"));
-		}
-		else
-		{
-			cdvd2->SetItemLabel(_("Boot CDVD (&fast)"));
-			cdvd2->SetHelp(_("Use fast boot to skip PS2 startup and splash screens"));
-		}
-	}
+	if(!cdvd2)
+		cdvd2 = m_menuSys.Insert(1, MenuId_Boot_CDVD2, MenuItemLabels[ActiveVM][1][enum_cast(Source)]);
```
ssakash: > It'd be better to not create the menu item in the no disc case if it doesn't already exist since it'll just get destroyed later.

IIRC, I added a ``Source!= CDVD_SourceType::Nodisc`` check on that ``if()`` before which ended up with the whole GUI freezing so didn't really bother with that and was too lazy to look into it.

>Also, assertions will fail in debug mode in the no disc case since a menu item is being created with no label.

Passing a string with blank spaces should be good then, I assume the array had it as a null character since it was left out of initialization.

turtleli: It really isn't so difficult. This:
```
wxMenuItem *cdvd_fast = menubar.FindItem(MenuId_Boot_CDVD2);
if (g_Conf->CdvdSource == CDVD_SourceType::NoDisc) {
	if (cdvd_fast)
		m_menuSys.Destroy(cdvd_fast);
} else {
	wxString label;
	wxString help_text;
	if (ActiveVM) {
		if (g_Conf->CdvdSource == CDVD_SourceType::Iso)
			label = _("Reboot ISO (&fast)");
		else
			label = _("Reboot CDVD (&fast)");
		help_text = _("Reboot using fast boot (skips splash screens)");
	} else {
		if (g_Conf->CdvdSource == CDVD_SourceType::Iso)
			label = _("Boot ISO (&fast)");
		else
			label = _("Boot CDVD (&fast)");
		help_text = _("Use fast boot to skip PS2 startup and splash screens");
	}

	if (cdvd_fast == nullptr) {
		m_menuSys.Insert(1, MenuId_Boot_CDVD2, label, help_text);
	} else {
		cdvd_fast->SetItemLabel(label);
		cdvd_fast->SetHelp(help_text);
	}
}
```
works for me. I didn't need to create and destroy an item that doesn't get used. I also didn't need to use an array and spam casts.

turtleli: I prefer my version - it's more obvious to me what labels and help messages are being used in each case.

Your array code is dependent on enum order. And I don't think checking for an empty label is a great way of expressing "We don't want a fast boot menu item when No Disc is selected". I'm not sure the static will work if the language is changed BTW - I don't know whether the pointed to memory will go out of scope or not, I haven't checked. (The menu language also won't change after changing the language, but that applies to quite a bit of the other translated parts of the GUI as well).

> (Though it should probably be changed for variable naming consistency)

Change if necessary - some of the GUI files use N different naming conventions which is a PITA, so I just used that convention (FWIW the naming isn't all that inconsistent when you only look at the rest of the function).

refractionpcsx2: @turtleli your version looks better, however you have absolutely no "boot fast" options.

turtleli: Sure, my version is longer, but it's also safer. It doesn't assume specific enum values so you avoid ever having to bugfix like a415d59e2f9e36c92c218073709d540c03a57a0e. The safety is worth the extra lines.

refractionpcsx2: But that commit adds the enums where they were missing? I don't get your point there.

I understand your issue with Enums, but unless that stuff needs to change at some point to add new options etc, it shouldn't be an issue.

refractionpcsx2: oh yes, sorry I see it now.

turtleli: BTW, there may be other ways of writing the code that we haven't considered - `std::map` could possibly work too.

pcsx2/gui/MainFrame.cpp
```diff
@@ -584,19 +587,66 @@ void MainEmuFrame::OnActivate( wxActivateEvent& evt )
 }
 // ----------------------------------------------------------------------------
 
+static void GetMenuItemLabel(MenuIdentifiers Id, wxString* label)
```
gregory38: It would be so much better to directly return the string. Beside it would remove the various break and `{}` in the case.

pcsx2/gui/MainFrame.cpp
```diff
@@ -636,35 +686,45 @@ void MainEmuFrame::ApplyCoreStatus()
...
+	{
+		wxString label;
+		std::map<CDVD_SourceType, wxString> help_text;
+		help_text.insert(std::pair<CDVD_SourceType, wxString>(CDVD_SourceType::Iso, _("Boot the VM using the current ISO source media")));
+		help_text.insert(std::pair<CDVD_SourceType, wxString>(CDVD_SourceType::Plugin, _("Boot the VM using the current CD/DVD source media")));
+		help_text.insert(std::pair<CDVD_SourceType, wxString>(CDVD_SourceType::NoDisc, _("Boot the VM without any source media")));
+		GetMenuItemLabel(fullboot_id, &label);
+
+		cdvd_full->SetItemLabel(label);
+		cdvd_full->SetHelp(help_text.at(Source));
```
gregory38: Why not a switch case ? Or a bunch of if/elseif.

pcsx2/gui/MainFrame.cpp
```diff
@@ -584,19 +587,66 @@ void MainEmuFrame::OnActivate( wxActivateEvent& evt )
 }
 // ----------------------------------------------------------------------------
 
+static void GetMenuItemLabel(MenuIdentifiers Id, wxString* label)
+{
+	const CDVD_SourceType Source = g_Conf->CdvdSource;
+	const bool ActiveVM = SysHasValidState();
```
ssakash: ``ActiveVM`` is no longer needed as we removed the ``reboot`` option. Just a left over from my older code where the function was also initializing strings for the reboot menu item labels.

pcsx2/gui/MainFrame.cpp
ssakash: Fine, IIRC I made it call by reference for some reason but no clue why.

pcsx2/gui/MainFrame.cpp
ssakash: Probably because the code will take up more lines?

pcsx2/gui/MainFrame.cpp
ssakash: > It would probably be better to just use the switch directly instead of placing the switch in a function.

I'm not keen on separating it from a dedicated function, it looks much cleaner to handle the menu item labels naming separately and helps to avoid congesting up the code in ``MainEmuFrame::ApplyCoreStatus()``.

turtleli: Well, I think less time is used to determine what label and help text is used in:
```
if (wxMenuItem *cdvd_full = menubar.FindItem(fullboot_id)) {
	switch (Source) {
	case CDVD_SourceType::Iso:
		cdvd_full->SetItemLabel("Boot ISO (f&ull)");
		cdvd_full->SetHelp(_("Boot the VM using the current ISO source media"));
		break;
	case CDVD_SourceType::Plugin:
		cdvd_full->SetItemLabel("Boot CDVD (f&ull)");
		cdvd_full->SetHelp(_("Boot the VM using the current CD/DVD source media"));
		break;
	case CDVD_SourceType::NoDisc:
		cdvd_full->SetItemLabel("Boot BIOS");
		cdvd_full->SetHelp(_("Boot the VM without any source media"));
		break;
	default:
		pxAssert(false);
	}
}
```
than in:
```
if (wxMenuItem *cdvd_full = menubar.FindItem(fullboot_id))
{
	wxString label = GetMenuItemLabel(fullboot_id);
	std::map<CDVD_SourceType, wxString> help_text;
	help_text.insert(std::make_pair(CDVD_SourceType::Iso, _("Boot the VM using the current ISO source media")));
	help_text.insert(std::make_pair(CDVD_SourceType::Plugin, _("Boot the VM using the current CD/DVD source media")));
	help_text.insert(std::make_pair(CDVD_SourceType::NoDisc, _("Boot the VM without any source media")));

	cdvd_full->SetItemLabel(label);
	cdvd_full->SetHelp(help_text.at(Source));
}
```
where the help text is more or less immediately known, but you have then navigate to SetMenuItemLabel to determine what label is being used:

```
if (Id == MenuId_Boot_CDVD)
{
	switch (Source)
	{
	case CDVD_SourceType::Iso:    return _("Boot ISO (f&ull)");
	case CDVD_SourceType::Plugin: return _("Boot CDVD (f&ull)");
	case CDVD_SourceType::NoDisc: return _("Boot BIOS");
	default:
		pxAssert(false);
	}
}
```
YMMV

ssakash: > but you have then navigate to SetMenuItemLabel to determine what label is being used:

I think you mean ``GetMenuItemLabel()``? It's literally just one line before ``MainEmuFrame::ApplyCoreStatus()`` and I don't see how navigating to a function to view it's content is a hard thing. I would prefer to separate these label string initialization to a separate function to reduce the mess in  ``MainEmuFrame::ApplyCoreStatus()`` as stated earlier.

turtleli: It requires less time (as I said before) and less thinking to understand the switch version IMO. But let's just disagree on this.

turtleli:
Oops, didn't realise I created a merge conflict. Anyway, reviewed, I still think just using local switches is preferable instead of map + switch in separate function, but I'm not going to stop a merge if others think it's okay.

ssakash:
> Oops, didn't realise I created a merge conflict.

 "**What turtleli actually thinks**"

``Haha, you deserve that merge conflict for not inclining to my request on using ifs/switch. Mwahahaha!``

:P

turtleli:
**"What turtleli actually thinks"**

> I do not comprehend. ApplyCoreStatus() only enables/disables menu items and sets their labels and help text. Is a helper function needed for getting the labels when the function is simple enough? I do not comprehend.

;)

turtleli:
@gregory38 @refractionpcsx2 I'll leave the decision to you ;)

gregory38:
@turtleli which decision ?

gregory38:
A switch case is better than a map. @ssakash could you update it please.

By the way what do you think of `s/CDVD_SourceType/CDVD_Source/` ? Enum name feels rather long.

turtleli:
Yeah, it was the decision to merge or not - I did test the code and it did work fine, but I disagreed on the separate function and map.

ssakash:
> A switch case is better than a map. @ssakash could you update it please.

Guess I'm outnumbered here, changed it to ``switch-case`` as turtleli suggested.

> By the way what do you think of s/CDVD_SourceType/CDVD_Source/ ? Enum name feels rather long.

Well, the original name of the enum was ``CDVD_SourceType`` and personally I have no plans to change it unless it's necessary to be changed for some sort of reason, also we use these sort of naming in other places at GSdx. ``(GPURendererType, GSRendererType, stateType)``

Maybe it could be reduced for typing convenience but honestly the current name seems rather fine to me.

pcsx2/gui/MainFrame.cpp
```diff
@@ -645,31 +643,65 @@ void MainEmuFrame::ApplyCoreStatus()
...
-			cdvd->SetItemLabel(_("Reboot CDVD (f&ull)"));
-			cdvd->SetHelp(_("Hard reset of the active VM."));
+		case CDVD_SourceType::Iso:
+			label = _("Boot ISO (&fast)");
+			break;
+		case CDVD_SourceType::Plugin:
+			label = _("Boot CDVD (&fast)");
+			break;
+		case CDVD_SourceType::NoDisc:
+			break; // Fast boot menu item is destroyed when no disc is selected.
```
turtleli: Hmm. Coverity might complain about this case since it can't actually be reached.

pcsx2/gui/MainFrame.cpp
```diff
@@ -643,35 +643,66 @@ void MainEmuFrame::ApplyCoreStatus()
...
+		case CDVD_SourceType::Iso:
+			cdvd_full->SetItemLabel(_("Boot ISO (f&ull)"));
+			cdvd_full->SetHelp(_("Boot the VM using the current ISO source media"));
+			break;
+		case CDVD_SourceType::Plugin:
+			cdvd_full->SetItemLabel(_("Boot CDVD (f&ull)"));
+			cdvd_full->SetHelp(_("Boot the VM using the current CD/DVD source media"));
+			break;
+		case CDVD_SourceType::NoDisc:
+			cdvd_full->SetItemLabel(_("Boot BIOS"));
```
turtleli: One last thing: Perhaps make the shortcut key `t` for the full/bios boot cases?

pcsx2/gui/MainFrame.cpp
```diff
@@ -465,13 +465,13 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
 	m_menuCDVD.Append( GetPluginMenuId_Settings(PluginId_CDVD), _("Plugin &Menu"), m_PluginMenuPacks[PluginId_CDVD] );
 
 	m_menuCDVD.AppendSeparator();
-	m_menuCDVD.Append( MenuId_Src_Iso,		_("&Iso"),		_("Makes the specified ISO image the CDVD source."), wxITEM_RADIO );
```
ssakash: *sigh* I missed the instance of ``Iso`` in line [464](https://github.com/PCSX2/pcsx2/pull/1651/files#diff-1d6e5cd2c2b8d912c67d623d8e73d31dR464).

turtleli: There's Iso in pcsx2/gui/AppRes.cpp line 66 as well.

PR #1654, opened 2016-11-08
IOP: Convert most IOP memory access in the IRX HLE and debugging modu…

sudonim1:
…le to safe access through iopMem* functions.

Also made a minor change to allow passing module version numbers to our function implementations, which I'll make part of the extended ioman functionality pull request since it needs that.

Kprintf requires special attention so I didn't tackle it here, I don't even think it has the same features as the actual Sony IOP printf (e.g. no * for field width).

### DIFF
--- a/pcsx2/IopMem.cpp
+++ b/pcsx2/IopMem.cpp
@@ -490,3 +490,14 @@ void __fastcall iopMemWrite32(u32 mem, u32 value)
 		}
 	}
 }
+
+std::string iopMemReadString(u32 mem, int maxlen)
+{
+    std::string ret;
+    char c;
+
+    while ((c = iopMemRead8(mem++)) && maxlen--)
+        ret.push_back(c);
+
+	return ret;
+}
--- a/pcsx2/IopBios.cpp
+++ b/pcsx2/IopBios.cpp
@@ -91,10 +91,42 @@ namespace R3000A {
 #define ra (psxRegs.GPR.n.ra)
 #define pc (psxRegs.pc)
 
-#define Ra0 (iopVirtMemR<char>(a0))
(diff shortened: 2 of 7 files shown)


sudonim1:
And yes that's a lot of copying.  Yes it's necessary to use IOP memory safely without unwarranted assumptions.

pcsx2/IopMem.cpp
```diff
@@ -490,3 +490,14 @@ void __fastcall iopMemWrite32(u32 mem, u32 value)
 		}
 	}
 }
+
+std::string iopMemReadString(u32 mem, int maxlen)
```
gregory38: In my humble opinion, you ought to compute `maxlen` based on mem and the IOP memory size. This way you could remove the while loop below with only a string constructor that will use the safe  `maxlen` limit.

sudonim1: That's actually intended specifically for strings which aren't NUL terminated, it's not just for sanity.

sudonim1: To be clear, it is in fact entirely permissible for module names, the first application of this, to not be NUL terminated in both export and import tables.

sudonim1: Well maybe the default's too high, given that these will be going on the stack often.

sudonim1: Ah never mind, what am I saying? It allocates on the heap of course.

sudonim1: I really don't see why we should be concerned about the performance of this at all, it's only going to be invoked a few times per boot during recompilation and when logging.  An absolute guarantee of safety and correctness seems more important.

sudonim1:
'k well disagreement aside nothing preventing merging came up.

PR #1655, opened 2016-11-08
pcsx2 gui: use git desribe as version name

### LINKED ISSUE #1043: Keep Main Window Title Bar Name Consistent with Root Folder Name
Currently, the name used for PCSX2's title bar is **much** different to the name of the root folder. This is what I mean:

Folder: pcsx2-v1.3.1-1538-gd3ec74b-windows-x86
Title: PCSX2 1.3.1-20151211085228

I would be very grateful if the name for the title bar could be changed to that of the root folder, since it's basically impossible to tell which version I'm currently using. 

Something like this would be preferable - at least for me, anyway:

PCSX2 1.3.1-1538-gd3ec74b (capitalized and no OS name)

gregory38:
Shorter than date and give all the useful info

Require windows test

close #1043

### DIFF
--- a/cmake/Pcsx2Utils.cmake
+++ b/cmake/Pcsx2Utils.cmake
@@ -37,6 +37,7 @@ endfunction()
 
 function(write_svnrev_h)
     set(PCSX2_WC_TIME 0)
+    set(PCSX2_GIT_REV "")
     if (GIT_FOUND AND EXISTS ${PROJECT_SOURCE_DIR}/.git)
         EXECUTE_PROCESS(WORKING_DIRECTORY ${PROJECT_SOURCE_DIR} COMMAND ${GIT_EXECUTABLE} show -s --format=%ci HEAD
             OUTPUT_VARIABLE PCSX2_WC_TIME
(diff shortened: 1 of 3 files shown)


FlatOutPS2:
Old:
ps2

New:
ps2-2

As you can see it works, but I do miss the date (the `Nov 8 2016` bit).

OT: Is it my imagination or are Appveyor builds slower lately.

gregory38:
Well I got my hand heavy tonight :stuck_out_tongue: 

You mean the compilation date, or the date of the commit. The former isn't useful and we have it at the top of the log.

FlatOutPS2:
But it looks so impersonal now. :cry: :p

If you want to keep to keep yourself busy you can always go ahead and fix this regression:
http://forums.pcsx2.net/Thread-ps1-runtime-error

MrCK1:
@gregory38 Just had a quick afterthought - I can open a separate issue if you want

GSdx screenshots are currently saved in this format: `gsdx_year|month|day|hour:min:sec` but would it be possible to simplify this: `gsdx_20161109230019` to something more readable like 

`gsdx_1579_dx11hw_nov182016_230019`
 `gsdx_1.6.0_dx11hw_nov182016_230019`
or
`gsdx_git1579_dx11hw_20161109_230019`
`gsdx_stable1.6.0_dx11hw_20161109_230019`

Issue #1386, opened 2016-06-05
Emulation Settings dialog page labels may be truncated


turtleli:
Seems to affect Windows only.

With certain languages (Arabic, Finnish, Korean, Russian, Swedish) you cannot see the full label text (EE/IOP, VU, etc.) unless you click on/hover over the label. I only checked the languages at 100% dpi. It might also affect the Plugin/BIOS selector dialog, but I didn't check.

The issue was introduced by #801, which can't be reverted because it fixed greater usability issues, like not being able to change some settings at higher dpi settings.

It can also be triggered on Windows 7 (and I assume Windows Vista as well) for other languages if some of the system appearance settings are changed i.e. decreasing the icon size (which seems to affect the value returned by `ListView_ApproximateViewRect`, I didn't investigate further).

FlatOutPS2:
> It can also be triggered on Windows 7 (and I assume Windows Vista as well) 

I assume that means it was discovered on Windows 8 or 10? I definitely see it in Windows 7, not sure if it's because any non-default settings. It doesn't look that bad though.

turtleli:
It affects all Windows, but can affect Windows 7 more depending on how you customise your system appearance settings (vsub noticed the Windows 7 stuff in #801, I was just too lazy to investigate and open an issue for it).

It's not a major issue, but it would still be nice to fix it.

Issue #1032, opened 2015-12-11
Git Version - Text Cut Off In Intro GUI


ghost:
I just compiled the latest git version from source in Slackware.

I compiled wxWidgets-3.0.2 against GTK2, and when I start PCSX2 for the first time and get the introduction GUI, I noticed text is being cut off. Please see the screen shots...

![pcsx2_1](https://cloud.githubusercontent.com/assets/13072581/11736278/f6ce80ba-9fc3-11e5-97c4-f6dd3265e422.png)

Next screen

![pcsx2_2](https://cloud.githubusercontent.com/assets/13072581/11736311/4dc42ed8-9fc4-11e5-984b-87eb6c706f46.jpg)

turtleli:
What GTK2 theme are you using?

Btw, you need to specify the actual git revision - "latest" is useless once new commits are pushed.

Geyup:
Sorry;  ```PCSX2 1.3.1-20151207111040```

This text cutting off happens with any GTK2 theme I'm using...

gregory38:
Are you sure that PCSX2 is linked with WX3? Could you do a `ldd path_to/PCSX2`
Could you test GTK3. There is an option to enable it at build time (like GTK3_API)

Geyup:
Ok I'll try with gtk3 and below is ldd.

ldd /usr/bin/32/PCSX2

```


	linux-gate.so.1 (0xf7755000)
	libwx_baseu-3.0.so.0 => /usr/lib/libwx_baseu-3.0.so.0 (0xf74a6000)
	libwx_gtk2u_core-3.0.so.0 => /usr/lib/libwx_gtk2u_core-3.0.so.0 (0xf6fde000)
	libwx_gtk2u_adv-3.0.so.0 => /usr/lib/libwx_gtk2u_adv-3.0.so.0 (0xf6e58000)
	libfreetype.so.6 => /usr/lib/libfreetype.so.6 (0xf6dc8000)
	libgtk-x11-2.0.so.0 => /usr/lib/libgtk-x11-2.0.so.0 (0xf695e000)
	libgdk-x11-2.0.so.0 => /usr/lib/libgdk-x11-2.0.so.0 (0xf68af000)
	libcairo.so.2 => /usr/lib/libcairo.so.2 (0xf6775000)
	libpango-1.0.so.0 => /usr/lib/libpango-1.0.so.0 (0xf672b000)
[...]

PR #1656, opened 2016-11-11
Greg/wx font

gregory38:
Related to PR #1523 , issue #1032 and issue #1386 

Need to test the behavion on windows. EE/VU cycle rate + debugger.


turtleli:
Doesn't affect issue #1386.

There are issues with German, Hungarian, and Japanese having text cut off when the EE cyclerate slider is set to extreme overclock on Windows.
japanese

The EE slider panel is actually larger in this case (11px with Japanese, didn't check the rest) due to the dynamic resizing. The vertical padding also seems a bit strange for the extreme overclock setting only - compare the space above and below the EE and VU slider texts.

gregory38:
yes space is a bit weird. I think we can improve it. First we can upgrade min width to 350. Then we can change the packing to use a 2x2 grid.

However the height annoying me, even in english I lost the cdvd option (I reduced gamefix panel size). It seems like the panel size isn't impacted by the layout call.

gregory38:
@turtleli I pushed some improvements for the sizer. However I'm not sure what is the best way to better pack EE/VU hack.

gregory38:
@nE0sIghT could you tell if this PR is enough to handle properly 1024x768 resolution. Try to move the EE slider around.

FlatOutPS2:
Holy ****, this looks horrible. :/ The game fixes tab is a nightmare.

gregory38:
Thanks for the encouragement ;) Screenshot please.

FlatOutPS2:
crap

So much empty space and I won't be able to tell if half of the fixes are turned on. Tooltips also seem to freeze in place and won't disappear until you move the mouse out of the fixes box.

gregory38:
Tooltip doesn't freeze but is attached to the widget (which is the full zone now). On linux I have normal space on the right/left. You seem to have a wider window. Maybe we can add some expands.

gregory38:
Maybe you can add a minimum size for the height. (and expand instead of center).
```
diff --git a/pcsx2/gui/Panels/GameFixesPanel.cpp b/pcsx2/gui/Panels/GameFixesPanel.cpp
index 645adc9..d7936b1 100644
--- a/pcsx2/gui/Panels/GameFixesPanel.cpp
+++ b/pcsx2/gui/Panels/GameFixesPanel.cpp
@@ -62,11 +62,13 @@ Panels::GameFixesPanel::GameFixesPanel( wxWindow* parent )
        )
    );
 
+   m_listbox->SetMinSize(wxSize(-1, 350));
+
    m_check_Enable->SetToolTip(pxE(L"Gamefixes can work around wrong emulation in some titles. \nThey may also cause compatibility or performance issues.\n\nThe safest way is to make sure that all game fixes are completely disabled.")).SetSubPadding( 1 );
    m_check_Enable->SetValue( g_Conf->EnableGameFixes );
 
    *this   += m_check_Enable   | StdExpand();
-   *this   += m_listbox        | pxCenter;
+   *this   += m_listbox        | StdExpand();//pxCenter;
 
    Bind(wxEVT_CHECKBOX, &GameFixesPanel::OnEnable_Toggled, this, m_check_Enable->GetId());
 
```

However it is a bit tricky. EE/VU will change the size of the window, so it would require to recompute the layout of the GameFix panel too.

gregory38:
Hum, potentially we can do
```
	if(GetBestSize().GetHeight() < (wxSystemSettings::GetMetric(wxSYS_SCREEN_Y) - 50))
		m_listbox->SetMinSize(wxSize(-1, 430));
```
or
```
       // use full size for real screen
	if(wxSystemSettings::GetMetric(wxSYS_SCREEN_Y) > 900)
		m_listbox->SetMinSize(wxSize(-1, 430));
```

gregory38:
@FlatOutPS2 Could you redo a screenshot with latest change.

nE0sIghT:
> @nE0sIghT could you tell if this PR is enough to handle properly 1024x768 resolution.

Will try today

FlatOutPS2:
@FlatOutPS2 Could you redo a screenshot with latest change.

window latest

The window height is back to that of the master, is that what's supposed to happen on screen with higher vertical resolutions?

gregory38:
```
The window height is back to that of the master, is that what's supposed to happen on screen with higher vertical resolutions?
```
Close. It is smaller on my case, I don't have the white part. Could you check the speedhack panel, do you have a space at the bottom ?

FlatOutPS2:
windowspeed

Yeah, there's a gap.

gregory38:
pcsx2/gui/Panels/GameFixesPanel.cpp:68
Could you try 350 instead of 420

FlatOutPS2:
window 300


> pcsx2/gui/Panels/GameFixesPanel.cpp:68
Could you try 350 instead of 420

That helped reducing it slightly, but setting it to 300 made it the same height as the original PR.

gregory38:
Ideally I need to check the wx code, maybe we can query/compute the minimum size to avoid the scrollbar.

gregory38:
@FlatOutPS2 I found another solution. On windows, I can disable the scrollbar, so hopefully it will fit better.

nE0sIghT:
> @nE0sIghT could you tell if this PR is enough to handle properly 1024x768 resolution

No, window's height is too large.

gregory38:
@nE0sIghT Thanks for the info.Hum microVU hack are really verbose. We really need to fix this white space issue for enable sub-string.

FlatOutPS2:
> @FlatOutPS2 I found another solution. On windows, I can disable the scrollbar, so hopefully it will fit better.

It now looks like the original screenshot I posted again, except without the scrollbar, so you can't access the out of view fixes at all anymore...

gregory38:
Great.I love gui coding. So we will go with 400 as a fixed size.

gregory38:
@nE0sIghT @FlatOutPS2 I need  a new test again. I have updated the main background so everything will be wider (and hopefully smaller)

nE0sIghT:
It's better now - buttons are visible, but flattened

nE0sIghT:
> By the way nE0sIghT what is your local ?

Russian

nE0sIghT:
Just retested with LANG=C and all looks good.

gregory38:
But buttons are still broken in Russian ?

nE0sIghT:
> But buttons are still broken in Russian ?

Yes. Visible, but flattened

gregory38:
What happen if you move the EE/VU slider ?

nE0sIghT:
Here is screenshots

gregory38:
Are buttons better, move the windows with alt-left click.

nE0sIghT:
> Are buttons better, move the windows with alt-left click.

Could you explain, please?
Moving window do not fix buttons. Also alt+left mouse click do nothing in GNOME 3 if I understand correctly (it's same as left mouse click without alt).
I can resize window in GNOME 3 (make it larger than screen resolution) - this action fixes buttons. However user that reported 1024x768px issue in Gentoo's Bugzilla claimed that in LXDE windows is not resizable.

gregory38:
EE cyclerate change relayouts the window. I wanted to know, if they also relayout the button correctly. But answer seems to be no. Anyway, on my side the window is 1125x641 (with the line removal but I'm not sure it impacts the global height).

gregory38:
Hum, I think the buttons are flattened because the windows is too big versus the desktop.

nE0sIghT:
Definitely it's true

gregory38:
Hum, potentially we can pack the text in the top-left corner (height will be the same). Then we can move the restore default button on the top right. It would make the "microVU column" smaller. But it doesn't solve the "EE column" case.

gregory38:
@nE0sIghT I guess you compile from source. What happen if you increase further the size of pcsx2/gui/Resources/BackgroundLogo.png ? Let's say to 500 pixels wide. Do you have issue on the X axis ? Or does wx manage to fit the window properly.

gregory38:
@nE0sIghT I've maybe found a solution. I reduced the padding, it seems to save 30/40 pixels of height. I hope it will be enough.

FlatOutPS2:
Latest change has increased width from 837 to 1003(which makes the whole menu look horrible) and height has increased again from 504 to 563. All Game Fixes are visible again though.

nE0sIghT:
Looks perfect now with Russian locale

gregory38:
Ok. I will push to master the commit that fix stuff properly
* Default spacing of 3 for small screen
* Wrong text size
* Properly relayout after string change in speedhack

I will keep here for further discussion
* background update
* the new check list box

gregory38:
@nE0sIghT I pushed some change on the master git. And I reduce further the spacing to 2 (instead of 3). Maybe it will be enough. Could you regenerate me 2 screenshot of the speedhack/game fixes panels from the master git. Thanks you very much.

nE0sIghT:
Master branch don't looks good for me

gregory38:
Hum, I just noticied that previous good screen got the text cut-off on the microVU speedhack sub panel. Display already fill all the width.

gregory38:
@nE0sIghT I reduced further the padding. As it mostly miss the button, it could be enough for the standard case. But I'm afraid it won't be enough for the speed hack panel when you move the cycle rate on the right.

nE0sIghT:
Buttons are visible now, but flattened. However it's usable now

nE0sIghT:
> But I'm afraid it won't be enough for the speed hack panel when you move the cycle rate on the right.

It's true. Moving sliders makes buttons invisible

gregory38:
@nE0sIghT I guess the issue is mostly the speed hack panel. As you can see there is a default button, do you think it will help if we remove it. On your screenshot EE/VU side have same height. So this extra (useless) button is quite bad.

willkuer:
But isn't the main problem the large margin of the top text?

FlatOutPS2:
> As you can see there is a default button, do you think it will help if we remove it. On your screenshot EE/VU side have same height. So this extra (useless) button is quite bad.

It's not useless. Instead of removing it, it should be added to the tabs that don't currently have it.

gregory38:
```
But isn't the main problem the large margin of the top text?
```
To compute the text width the width of the parents are checked. The last parent is the small startup GUI (which size depends on the background image). So virtually the GUI is sized on 2 columns, each with the size of the background. Unlike others texts, the top text is wider than a single column. In order to achieve it, the code will compute the size of the area for 1 column (W,H) and then later it will expand the width to fill the gap (2*W, H). Please note that I said expand not shrink. So far I didn't find a way to relayout all the widget to reduce the height. If we remove the last parent constraint, the GUI becomes wider. At least on my PC, I don't know how it will behave for small display.

```
It's not useless. Instead of removing it, it should be added to the tabs that don't currently have it.
```
It would be better to add it at the bottom (with preset). But by removing, I can still move it at the top right, and put the text in the top left.

gregory38:
```
It would be better to add it at the bottom (with ok/cancel/apply)
```
Note: if anybody ( @turtleli , @ssakash , @FlatOutPS2 , others ) want to do it, please tell me it would help me ;)
I suspect we could add a wxID_DEFAULT button in `pcsx2/gui/Dialogs/ConfirmationDialogs.cpp` or `common/src/Utilities/wxHelpers.cpp` (I'm not sure which file control the button). Then it remains to route the signal to either the good panel or to all panels.

turtleli:
I think the `wxEVT_CHECKLISTBOX` event needs to be binded in the BaseConfigurationDialog constructor so that the apply button will enable properly when changes occur in the `wxCheckListBox`.

As for the default button, it should probably be added to the `m_extraButtonSizer` in `pcsx2/gui/Dialogs/SysConfigDialog.cpp` unless the plan is to add a default button to the other dialogs as well.

Side note: I'm making an attempt to rewrite the dialogs from scratch without using most of the helper classes - I'm hoping it'll eventually make the GUI code simpler and less buggy. It'll probably take a really long time, though what I've done so far seems to be working okay (I seem to have word wrapping with resizable dialogs working).

Master:
gamefixes_master

My branch:
gamefixes

gregory38:
To be honest, I'm not sure a single list check box widget is a good idea. If you can fix the top space, it will have enough place to fit properly. The list box isn't nice for the tooltip.

ssakash:
Personally, I prefer turtleli's GUI. The panels are neatly organized with a list box control similar to the disassembly dialog (I assume ``wxNotebook?``) and the weird logos are no longer there and potentially we get a nice initiative to rework the GUI with better placement of the dialog elements, so I'm all in support for turtleli's idea. 👍

MarcoEstevez:
+1 turtleli's fashion

gregory38:
@ssakash 
I don't find the logo weird. Otherwise it is a bit austere.

@turtleli 
tbh, I didn't understand yet/fully how event bind is working. I prefer toying with vtune ;)

turtleli:
> I assume wxNotebook?

Yes - it seems to fix #1386, so I'm using that in my branch.

> To be honest, I'm not sure a single list check box widget is a good idea. If you can fix the top space, it will have enough place to fit properly. The list box isn't nice for the tooltip.

Yeah, it isn't. I'll probably switch back to individual checkboxes on my branch, though the reason for that is because I want to see if I can design an enum friendly helper class for it.

> tbh, I didn't understand yet/fully how event bind is working. I prefer toying with vtune ;)

That's best understood by reading the documentation ;) Well, I think I understand how it works, but maybe I'm just in denial ;)

lightningterror:
I also prefer turtleli's gui , perhaps small icons can be added next to the letters but that's optional.

lightningterror:
What are we gonna do with this ? It hasn't been touched for over a year now.

gregory38:
Maybe drop it ? I'm sorry but I think people should really buy laptop with bigger resolution.

nE0sIghT:
> Maybe drop it ? I'm sorry but I think people should really buy laptop with bigger resolution.

Looks reasonable.

lightningterror:
Ok I'll close it, maybe turtleli or someone else can revisit this in the future.

PR #1659, opened 2016-11-12
reformat main common directory

gregory38:
Disable some reformat around emitter registers declaration and a massive cases but otherwise it is good.

The core will be converted later do avoid too much conflict with current PR.


gregory38:
No complain so I merge it to avoid conflict later.

PR #1661, opened 2016-11-13
Replace __threadlocal with thread_local.

KousukeItsagame:
Requires GCC 4.8.x See cxx11.

### DIFF
--- a/common/include/Utilities/TlsVariable.inl
+++ b/common/include/Utilities/TlsVariable.inl
@@ -22,6 +22,7 @@
 #		define DeclareTls(x) __threadlocal x
 #	else
 #		define DeclareTls(x) thread_local x
+#	endif
 #else
 #	define DeclareTls(x) Threading::TlsVariable<x>
 #endif
--- a/common/include/Utilities/TlsVariable.inl
+++ b/common/include/Utilities/TlsVariable.inl
@@ -18,7 +18,10 @@
 #include "Threading.h"
 
 #if PCSX2_THREAD_LOCAL
-#	define DeclareTls(x) __threadlocal x
+#	if _MSC_VER = 1800
+#		define DeclareTls(x) __threadlocal x
+#	else
+#		define DeclareTls(x) thread_local x
 #else
 #	define DeclareTls(x) Threading::TlsVariable<x>
 #endif
--- a/common/include/Pcsx2Defs.h
+++ b/common/include/Pcsx2Defs.h
@@ -168,7 +168,11 @@ static const int __pagesize	= PCSX2_PAGESIZE;
 #	define PCSX2_ALIGNED16_EXTERN(x)	extern __declspec(align(16)) x
 
 #	define __noinline		__declspec(noinline)
-#	define __threadlocal	__declspec(thread)
+#	if _MSC_VER == 1800
+#		define __threadlocal	__declspec(thread)
+#   else
+#		define local_thread 	__declspec(thread)
+#	endif
 
 // Don't know if there are Visual C++ equivalents of these.
 #	define likely(x)		(!!(x))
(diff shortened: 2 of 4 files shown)


sudonim1:
From what I read this absolutely murders performance on every reference to a thread local variable.  Unsure how it affects MSVC.

KousukeItsagame:
It compiles on visual studio 2015.
I do not know how to check te performance of it in windows
but I added this as it seems to do the same but is newer.

KousukeItsagame:
I accidently mad another pull request.
Can you delete it?

uyjulian:
Just close it.

refractionpcsx2:
Closed the other one.

KousukeItsagame:
thanks 👍

sudonim1:
So we've gone from doesn't compile in msvc 2013 to doesn't compile anywhere?

refractionpcsx2:
..\..\src\Utilities\Console.cpp(26): error C2144: syntax error: 'int' should be preceded by ';' [C:\projects\pcsx2\common\build\Utilities\utilities.vcxproj]
         ..\..\src\Utilities\Console.cpp(26): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int [C:\projects\pcsx2\common\build\Utilities\utilities.vcxproj]
         ..\..\src\Utilities\Console.cpp(29): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int [C:\projects\pcsx2\common\build\Utilities\utilities.vcxproj]
         ..\..\src\Utilities\Console.cpp(29): error C2086: 'int __threadlocal': redefinition [C:\projects\pcsx2\common\build\Utilities\utilities.vcxproj]
         ..\..\src\Utilities\Console.cpp(29): error C2146: syntax error: missing ';' before identifier 'ConsoleColors' [C:\projects\pcsx2\common\build\Utilities\utilities.vcxproj]
         ..\..\src\Utilities\Exceptions.cpp(53): error C2144: syntax error: 'int' should be preceded by ';' [C:\projects\pcsx2\common\build\Utilities\utilities.vcxproj]
         ..\..\src\Utilities\Exceptions.cpp(53): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int [C:\projects\pcsx2\common\build\Utilities\utilities.vcxproj]

refractionpcsx2:
In future please compile your code before submitting it please :)

sudonim1:
What exactly is the point of using the C++11 specifier if we have to hide it behind a lot of macros and precompiler conditions anyway?

KousukeItsagame:
Maybe not now, But in the future?
the macro's are only for visual studio 2013
it's the only build target that doesn't work with local_thread

I'm finishing the work anyway :)
I'm aware it might not be used though

@refractionpcsx2 ok

common/include/Utilities/TlsVariable.inl
```diff
@@ -18,7 +18,11 @@
 #include "Threading.h"
 
 #if PCSX2_THREAD_LOCAL
-#	define DeclareTls(x) __threadlocal x
+#	if _MSC_VER = 1800
```
willkuer: = or ==?

gregory38:
```
Maybe not now, But in the future?
```
I don't think compiler will stop supporting their builtin. Is just a different keyword to parse. In the enterprise world, people just began to use C++11.

On your PR, if I understand correctly, `__threadlocal` is local PCSX2 macro. So you only need to update `Pcsx2Defs.h`.

Honestly, I think it will best to wait VS2017 release (or whatever it is called). Then we could drop VS2013 and uses fully the new syntax sugar of C++11.

gregory38:
Sorry, we are in the middle of automatic reformating the code with clang-format (the common subdirectory). You will need to rebase, and you will need to format the code with clang-format.

KousukeItsagame:
I'll agree to postpone it.
I'll start again when it's more feasable in the future :)

gregory38:
For the record, I plan to use clang-tidy to upgrade the db to C++11 syntax sugar. However it will be done after the complete reformating. And likely after the next release.

PR #1663, opened 2016-11-16
debugger: memory view improvements

mogaika:
Changes:
1. Highlight memory location referenced by current register
2. Use radio and checks in menu
3. Allow memory window not be aligned by row size


gregory38:
Is it a screenshot from Windows or Linux ?

And what do you mean by 3? Which window ? Which row ? Ideally a screenshot with option off/on would be nice.

refractionpcsx2:
> Is it a screenshot from Windows or Linux ?

looks like a windows screenshot to me.

ssakash:
> looks like a windows screenshot to me.

From the looks of it, he has probably used MS paint to add those red arrows in the picture so he might have probably captured it from Windows.

**(or)**

He could have taken the screenshot in Linux and stored it in a partition common to both the Linux/Windows operating systems and eventually he could have switched the OS to Windows to edit the picture using MS paint.

However, the latter method sounds rather troublesome, so I assume it's the former case.

FlatOutPS2:
I'm sure MS Paint can work on Linux through Wine. :p

mogaika:
@gregory38 
aligned window like your left byte address cannot be non 0x??????00 byte (aligned like by dword)
in my previous commit 7e91eb661691829f96a82e995612040caac557cf I specially allow this behaviour, but in f2a0f9f61ae014120d3b34ba459575b51247929a  it specially fixed by Kingcom. But in different cases different methods is preferably.
I don't think why Kingcom made align by rowSize, not by 16 byte, but I think there is some logic

@ssakash win! I use mspaint. Because I break my ubuntu when trying to build pcsx long time ago.

@gregory38 Collage for you with feauture switched off
collage

gregory38:
ok it is fine for me. Rendering is the same on Linux.

Issue #796, opened 2015-08-27
[Feature request] GSdx: detect supported instruction sets on runtime.


willkuer:
Current implementation creates five different forks of the same code using sse2, ssse3, sse4.1, avx, avx2 which are selected at build time. The user then can choose one of these plugins during runtime. Unfortunately most user don't know anything about instruction sets and are puzzled which of these five plugins is the best.

Would it be possible to merge all forks into one and outsource only the instruction set dependent part interfacing necessary functions which then gets dynamically loaded at runtime?

I don't know how the structure of gsdx is and where the instruction codes make the difference but I would believe oop should enable such a possibility. I would even believe due to maintainability reasons you should already have every implemented...

The advantages are:
Simplification of plugin selection.
Possibly easier introduction of new/other promising instruction sets.

Disadvantages are most probably:
Increased code complexicity
Stealing time of developers for more important stuff.

refractionpcsx2:
Yes and no i think.  

As it's currently set up, the profiles take advantage of allowing the compiler to use the enhanced instruction sets on the standard code as well as having the segmented code for the specific software renderers, if it was to detect on the fly, the best gsdx could be compiled with from the compiler is sse2, essentially losing some of the benefits of the higher instruction sets.

micove:
As explained in:
https://software.intel.com/en-us/articles/avoiding-avx-sse-transition-penalties
http://stackoverflow.com/questions/7839925/using-avx-cpu-instructions-poor-performance-without-archavx

It's not fun. It would be a bit easier if there was no AVX/AVX2 code. AVX/AVX2 uses 256 bit registers but modern cpu's only read/write 128bit at a time so the difference between SSE4 and AVX should not be much. I don't have a AVX cpu so I can't test it for PCSX2 but for some applications SSE2 code runs faster than the AVX equivalent,

tsunami2311:
I could be mistake but dont some instructions set work better then others in different setups?

sse4.1 works better then AVX  and vis versa in certain setups? maybe instead of have different plugins for each instruction set, just have 1plugin and let the user pick which set?, which guess wouldn't help with the confusion. but atelst it all be in one plugin?

willkuer:
@tsunami2311
This was more or less the idea. Better to have a hidden advanced option where you select the instruction set than a forced selection choice before you even see one ingame scene.

@micove
[...]

PR #357, opened 2014-11-23
64 bit : interpreter & GSdx Hardware working

3kinox:
Fixes up things so as to get 64 bit building & "working".
Interpreter runs, so does GSdx though only in Hardware( some fixes are actually done for software but someone who knows it better than me will have to finish it as it seems clearly unfinished. It needs to be builld in avx too, as sse2 one is not considered as valid(many methods are empty, it builds but linking seems to fail, don't know why, it was already like that).
SuperVU is removed as it is, according to your own discussions in dev part of forum, not portable and I could not even build with it. As a bonus, build is faster after it's suppressed :)
Fixes mostly include typing, I don't know if I got every last stray "L" before strings going to wxwidget, some of them are even nescessary, the others cause crashes, I probably got almost all reachables, however, if you ever get a crash in wxclen, just backtrace until you have some write to wxwidget that include a "L" prefixed chain.
I will use that as a base to continue the port. I will see if I just port the current frontend/backend(my head ache a little as I try to understand the code though for now) or if I just make a new one(potentially using qemu as a base then).

PR #1664, opened 2016-11-19
Greg/gsdx 64b

gregory38:
AVX/x64 implementation of GSdx JIT compiler. So far, perf sucks. -20% to -10% versus 32 bits.

With hard work, it might be possible to reach the 32 bits performance level (or close enough) but it is really a low priority. At least, we can run, test and dev on a 64 bits GSdx.


willkuer:
Do you run the core in interpreter? How do you benchmark? In GS replayer?

gregory38:
Yes I benchmarked without the core through the replayer. Interpreter will be too slow and I have additional timing info on the replayer.

Here an example on HW renderer (but it is the same).
```
Performance Profile for 53 frames:
Min  1.40 ms	(711.80 fps)
Mean 7.50 ms	(133.31 fps)
Max  13.54 ms	(73.88 fps)
SD   3.72 ms

Frame Repartition
  0 ms =>   2 ms	   1
  2 ms =>   4 ms	  16
  4 ms =>   6 ms	   9
  6 ms =>   8 ms	   0
  8 ms =>  10 ms	   7
 10 ms =>  12 ms	  15
 12 ms =>  14 ms	   5
 14 ms =>  16 ms	   0
```

willkuer:
Interesting that the histogram shows two maxima. Maybe one can improve the bottleneck of the slower one.

Have you tried with different compiler? I guess compiler can be important here.

Also do you know the bottleneck is in gsdx and not in the gpu driver? Maybe calling the driver from 64bit results in differences?

gregory38:
Above was an example on the hw renderer (32 bits). I didn't test the hw renderer but it could be interesting. It is less critical on the hw renderer (often gpu limited)

Some frames are nearly empty (internal 30 fps). So they are quick to render.

gregory38:
By the way, if someone can test the SW renderer on 32 bits with all ISA. I want to avoid regression due to code factorization.

gregory38:
I pushed additional change to select the ISA (SSE2/SSSE3/SSE41/AVX1) at runtime.

It only impact the self-generated code (AKA the SW renderer). Nevertheless (if I manage to compile it on VS), it make 
* SSSE3 build useless
* AVX1 build mostly useless

So in the future, we could limit it to SSE2/SSE4.1/AVX2.

gregory38:
@turtleli 
Could you help me to fix the compilation issue on Windows? Are AVX files compiled on windows ?

turtleli:
Yeah, it's the exclude build stuff.

Fix for 32-bit Windows build (64 bits doesn't compile because `__x86_64__` in stdafx.h isn't defined on Windows) - https://gist.github.com/turtleli/a05906730b9b885d5dde2e862440238f

gregory38:
Thanks for the patch.

gregory38:
Let's go. The best way to have some test coverages :+1:

willkuer:
Did I understood correctly that this pr solved #796?

gregory38:
No. It is only a step forward. You have the C code, and you have the code generated by the C itself which is then executed. The latter is used by the SW renderer.
In case of SSE2 up AVX1 build, the generated code will be based on your CPU capabilities. However all C code will depend on the compilation flag.

However, I think SSSE3 and AVX1 build are useless now (I hope I won't have AVX penalty).

mirh:
Mhh.. is this affecting #357 then?

gregory38:
Well it was only a stub (on gsdx sw) to allow the compilation (used to crash). New code is working.
The limitation
* mipmap isn't implemented
* no AVX2
* no SSE but any CPU that support AVX will pick up the AVX version (~70% of the users)

Tbh, AVX2 will be done. But I don't think pure SSE will be done. It is too slow to worth it.

PR #1665, opened 2016-11-20
cdvdgigaherz: Fix race condition and other stuff

turtleli:
Changes:
 * Fixes a race condition where the wrong data could be sent to PCSX2.
 * Fixes a data race.
 * Fixes a sign mismatch in the CDVD plugin API.
 * Change variable/parameter types to reduce casting.
 * Fixes the cdvdgigaherz Coverity issue.

The plugin will probably have slightly worse performance in some cases, but it might be unavoidable (and it's better than random crashing).

### DIFF
--- a/common/include/PS2Edefs.h
+++ b/common/include/PS2Edefs.h
@@ -437,7 +437,7 @@ void CALLBACK CDVDnewDiskCB(void (*callback)());
 // new funcs
 
 // read a track directly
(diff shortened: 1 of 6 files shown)


gregory38:
LGTM. Do you want better test coverage, otherwise can I merge it?

turtleli:
It's probably fine to merge it.

PR #1672, opened 2016-11-22
Fix LTO flags.

np511:
Fixes LTO - removes the "--plugin" option that only worked with GCC 4.9 and only on x86_64 and wasn't needed there either. Uses "gcc-ar", "gcc-ranlib", and "gcc-nm" instead of the usual commands as those don't work with LTO. Removes the "-fuse-ld=gold" argument as a few plugins fail to compile with "gold" and "gold" is not used when LTO is disabled anyway.


gregory38:
Hum, is LTO really working without the gold linker ?

np511:
Yes, I just compiled it 100% successfully, and didn't have to change anything.

There were only a few plugins that didn't work even when gold was in use.

gregory38:
the question was "is the compiler able to do any real optimization without the plugin" or is the compiled program mostly the same as a standard build.

np511:
I'm not sure I understand - the LTO plugin *is* being used. That's what the -fuse-linker-plugin option does. I know that LTO is being used because the regular "ar" and "ranlib" both say that the plugin is needed. "gcc-ar" is a wrapper around ar that uses the plugin, and mostly the same with ranlib and nm. If LTO wasn't being used, there wouldn't be a need for the gcc- versions.

If the question is really, "Does LTO result in significant speedups?" I can't answer for certain.

With the stable version in the Debian repo, PCSX2 is rather slow no matter which preset is used and with minimal graphics options on the GSDX plugin, and Kingdom Hearts is barely playable both with and without frame skip. With this version and LTO, I can run with 2x Native resolution with otherwise normal options on the fourth preset (Aggressive) at a faster speed without frame skip, and I'm assuming LTO contributes at least partly to that. The LTO version is when using the Devel build type, not the Release, which I assume the repo uses (and the wiki says is faster).

It's likely that some speedup is a result of the newer compiler, and the fact that the GSDX plugin I compiled has SSE4.2 support (unlike the one installed with stable).

gregory38:
Yes debian stable is SSE2 whereas self build is optimized for your CPU. So you can't compare them. Besides I packs severals GSdx optimization since latest release.
If LTO has any impact it will be small but I will test it.

np511:
I am very confused - I though nothing worked on x86_64? The 64-bit test says it compiled without error!

np511:
With DesMuMe (A DS emulator), LTO increased average framerate in some games from ~25 FPS (In case you couldn't tell from the issue, my CPU is garbage and AFAIK my GPU isn't much better) to ~40.

Actually, DesMuMe's default renderer is SW, not HW, so that probably makes a big difference.

Would using a profile increase the speed on a local computer? (-fprofile-generate, use it for a bit, then -fuse-profile)?

gregory38:
```
I am very confused - I though nothing worked on x86_64? The 64-bit test says it compiled without error!
```
Who said nothing ? It misses the recompiler so it is slow as hell.


```
With DesMuMe (A DS emulator), LTO increased average framerate in some games from ~25 FPS (In case you couldn't tell from the issue, my CPU is garbage and AFAIK my GPU isn't much better) to ~40.

Actually, DesMuMe's default renderer is SW, not HW, so that probably makes a big difference.
```
GSdx SW renderer is hand-written in ASM. It allows to set all parameters as constant and to remove all branches (and jump). It means the impact of the compiler optimization will be small for us. It would be interesting to compare the same build with/without LTO.

```
Would using a profile increase the speed on a local computer? (-fprofile-generate, use it for a bit, then -fuse-profile)?
```
It depends on which part of the emulation you are limited. But it can help, the best is to do it by game.

gregory38:
I did a quick test on the SW interpreter (GSdx only). I didn't notice any real speed improvement.

Anyway, I'm fine to merge it.

np511:
No problem.

I'd forgotten that you mentioned that the SW renderer was hand-written. From what I can tell, LTO only has a real impact on C[++].

Where can I find an up-to-date TO-DO list for Linux? I'd like to help out more, but I'm not sure what needs to be done and what's already being worked on.

gregory38:
Honestly I was hoping an impact on all layers around the rasterizer. Anyway maybe core is more inpacted but I don't have a way to measure it properly.

Well there is an outdated todo list https://github.com/PCSX2/pcsx2/wiki/Contributing-%28TODO-List%29

* You can look at some Q&A stuff. Like gcc warning / coverity https://scan.coverity.com/projects/pcsx2-pcsx2 / clang-tidy (./build.sh --clang-tidy)
* Bigger task, support game.iso.xz compressed file. Lastest version support a builtin index so it can be interesting.
* Look at github issue
If you found something that you want to work, please open a forum thread on (linux) section. I can surely give you some useful inputs.

np511:
Just a quick point, after running "git pull" to the latest master and executing "./build.sh --release --extra --cross-multilib --lto" I got an error message - a few minutes later, I realized that it was caused by the upgrade from GCC 6.1 to 6.3 - apparently, combining the two changes caused an error.

I added --clean and removed --lto, and the error is gone - I'm going to find a menu/scene/etc where I can achieve >60 FPS, record the min/max/mean, then recompile with lto and repeat in the same scene (using a savestate).

TBH I'm not sure if it was the combined changes because I haven't used LTO yet, but it's most likely that LTO still works fine. 

Seeing that I haven't found any issues (on GitHub I mean, not that I personally haven't found any, though that is true as well) with LTO, I'm going to create a pull request to remove the warnings today or tomorrow (PGO as well).

PR #1673, opened 2016-11-23
GSdx Merge Circuit: Fix regressions

FlatOutPS2:
Fixes screen shaking in Tenchu: Wrath Of Heaven, and graphical issues in NASCAR 09.

### DIFF
--- a/plugins/GSdx/GSRenderer.cpp
+++ b/plugins/GSdx/GSRenderer.cpp
@@ -237,17 +237,18 @@ bool GSRenderer::Merge(int field)
 
 		// Time Crisis 2/3 uses two side by side images when in split screen mode.
 		// Though ignore cases where baseline and display rectangle offsets only differ by 1 pixel, causes blurring and wrong resolution output on FFXII
-		if(display_diff.x > 2)
+		// And cap the offset range to avoid cases with extreme values (NASCAR 09) causing graphical issues.
+		if(display_diff.x > 2 && display_diff.x < 10)
 		{
 			off.x = tex[i]->GetScale().x * display_diff.x;
 		}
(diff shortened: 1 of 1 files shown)


ssakash:
Well, did this seriously get merged? The code is definitely wrong as it doesn't allow any offset values above 10, haven't checked it yet but I'm 100% sure that it'll break split screen rendering in Time Crisis 2.

ssakash:
Also it's not really your fault, FlatOut. I blame the awful code, the merge circuit code in GSdx is full of hacks and it obscures the actual intention of the code in a lot of ways. so would be nice to properly clean up the code someday.

ssakash:
Also in case someone asks me for a GS dump - I can't provide it right now as I'm nowhere remotely close to my desktop/laptop, but IIRC I did provide a GS dump of ``Time Crisis 2`` to turtleli in some pull request, so try looking for the dump if you want to fix it. :P

FlatOutPS2:
No need, I found the Time Crisis dump in my dump archive and you're correct I does break the split screen. What number should be the max then, 640?

ssakash:
> What number should be the max then, 640?

I don't think having a guess based constant is a nice idea as it isn't generic and it would just obfuscate the already hacked up code in ``GSRenderer::Merge()``.

The question falls down to,

* Will having a max limit of ``640`` fix the regression we're facing with ``NASCAR 09``? **Probably yes**.

* Though will it be accurate to PS2 behavior? **It won't**.

I suggest you do what you think is right, I'm obviously no expert in GSdx. However for future PCRTC related PR's, would be nice if you could ask my feedback before it gets merged :P 

If you're highly insistent on going with a max limit idea then please do the checks only for the frame memory, not for the output circuit dimension.

PR #1675, opened 2016-11-24
GSdx Merge Circuit: Fix regression and issue

FlatOutPS2:
EDIT: Fixes graphical issues in NASCAR 09 and reverses changes that caused a regression in Time Crisis 2/3 split screen mode.

https://github.com/PCSX2/pcsx2/pull/1683 should prevent the Software mode crashes in the EA NASCAR games.

### DIFF
--- a/plugins/GSdx/GSRenderer.cpp
+++ b/plugins/GSdx/GSRenderer.cpp
@@ -139,42 +139,31 @@ bool GSRenderer::Merge(int field)
 		m_regs->DISP[0].DISPFB.FBW == m_regs->DISP[1].DISPFB.FBW &&
 		m_regs->DISP[0].DISPFB.PSM == m_regs->DISP[1].DISPFB.PSM;
 
-	// bool blurdetected = false;
-
 	if(samesrc /*&& m_regs->PMODE.SLBG == 0 && m_regs->PMODE.MMOD == 1 && m_regs->PMODE.ALP == 0x80*/)
 	{
-		if(fr[0].eq(fr[1] + GSVector4i(0, -1, 0, 0)) && dr[0].eq(dr[1] + GSVector4i(0, 0, 0, 1))
-		|| fr[1].eq(fr[0] + GSVector4i(0, -1, 0, 0)) && dr[1].eq(dr[0] + GSVector4i(0, 0, 0, 1)))
-		{
-			// persona 4:
-			//
-			// fr[0] = 0 0 640 448
-			// fr[1] = 0 1 640 448
-			// dr[0] = 159 50 779 498
-			// dr[1] = 159 50 779 497
-			//
-			// second image shifted up by 1 pixel and blended over itself
-			//
-			// god of war:
-			//
-			// fr[0] = 0 1 512 448
-			// fr[1] = 0 0 512 448
-			// dr[0] = 127 50 639 497
-			// dr[1] = 127 50 639 498
-			//
-			// same just the first image shifted
+		// persona 4:
+		//
+		// fr[0] = 0 0 640 448
+		// fr[1] = 0 1 640 448
+		// dr[0] = 159 50 779 498
+		// dr[1] = 159 50 779 497
+		//
+		// second image shifted up by 1 pixel and blended over itself
+		//
+		// god of war:
+		//
+		// fr[0] = 0 1 512 448
+		// fr[1] = 0 0 512 448
+		// dr[0] = 127 50 639 497
+		// dr[1] = 127 50 639 498
+		//
+		// same just the first image shifted
+		//
+		// These kinds of cases are now fixed by the more generic frame_diff code below, as the code here was too specific and has become obsolete.
+		// NOTE: Persona 4 and God Of War are not rare exceptions, many games have the same(or very similar) offsets.
 
-			int top = min(fr[0].top, fr[1].top);
-			int bottom = max(dr[0].bottom, dr[1].bottom);
-
-			fr[0].top = top;
-			fr[1].top = top;
-			dr[0].bottom = bottom;
-			dr[1].bottom = bottom;
-
-			// blurdetected = true;
-		}
-		else if(dr[0].eq(dr[1]) && (fr[0].eq(fr[1] + GSVector4i(0, 1, 0, 1)) || fr[1].eq(fr[0] + GSVector4i(0, 1, 0, 1))))
+		int topDiff = fr[0].top - fr[1].top;
+		if (dr[0].eq(dr[1]) && (fr[0].eq(fr[1] + GSVector4i(0, topDiff, 0, topDiff)) || fr[1].eq(fr[0] + GSVector4i(0, topDiff, 0, topDiff))))
 		{
 			// dq5:
 			//
@@ -188,10 +177,7 @@ bool GSRenderer::Merge(int field)
 
 			fr[0].top = fr[1].top = top;
 			fr[0].bottom = fr[1].bottom = bottom;
-
-			// blurdetected = true;
 		}
-		//printf("samesrc = %d blurdetected = %d\n",samesrc,blurdetected);
 	}
 
 	GSVector2i fs(0, 0);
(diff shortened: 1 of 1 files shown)


willkuer:
Was it finally an uint/int issue?

I think the code deserves comments regarding timesplitters and nascar. Especially the line //shouldn't it be >2 that you changed from 4 to 40 deserves an update (i.e. why 40?)

FlatOutPS2:
> Was it finally an uint/int issue?

Yeah. The value (2047) made no sense to me otherwise, so I gave this a try and it immediately resolved the issue.

> I think the code deserves comments regarding timesplitters and nascar. Especially the line //shouldn't it be >2 that you changed from 4 to 40 deserves an update (i.e. why 40?)

That was a mistake. :p The github app doesn't show changes as clear as the website does.

ssakash:
Code doesn't make sense to me,  DBX/DBY is a bit field with 11 bit width and no sign bit. So not sure why you converted it to ``int``?

gregory38:
What is DBX/DBY again ? Position offset.

GS memory wrap after 4MB, you must restart to 0. If code is wrong, it will overflow the memory and then crash. (as a trick GS memory is bigger to avoid small overflow but too big overflow will explode).

gregory38:
Is there any gs dump that can reproduce the crash ?

ssakash:
Yes, a GS dump of the crash would help. The crash is most likely caused due to either a bug in software mode (or) an insane register value init by the developer on ``DISPFB``, I hope it's the former.

FlatOutPS2:
> The crash might most likely be caused due to either a bug in software mode (or) an insane register value init by the developer on DISPFB, I hope it's the former.

Do you consider setting DBY to -1 an insane value?

ssakash:
> Do you consider setting DBY to -1 an insane value?

Insane? No, it's an impossible value. IIRC the data type format of DBX/DBY on the Graphics synthesizer is ``int 0:11:0`` which means no negative values as there's no sign bit.

FlatOutPS2:
> Insane? No, it's an impossible value. IIRC the data type format of DBX/DBY on the Graphics synthesizer is int 0:11:0 which means no negative values as there's no sign bit.

And because it's impossible it gets converted to 2047.

ssakash:
> And because it's impossible it gets converted to 2047.

I'm not sure, if the developer did indeed pass ``-1`` then it would most likely be stored as ``10000000001`` which would then be converted into``1025`` with respect to the integer format of DBY.

FlatOutPS2:
> I'm not sure, if the developer did indeed pass -1 then it would most likely be stored as 10000000001 which would then be converted into1025 with respect to the integer format of DBY.

When I set the value to -1 manually it got converted to 2047 as well.

ssakash:
> When I set the value to -1 manually it got converted to 2047 as well.

Oh, then it probably stores the 2's complement value. However, I still don't think any developer would be stupid enough to assign an illegal value to the bit-fields.

gregory38:
Framebuffer size is 2048x2048, so 2047 seem valid. And the memory wraps after 4MB. There are nothing illegal on the gs.

gregory38:
Any backtrace on the crash ? I'm pretty sure that we miss some memory wrapping.

plugins/GSdx/GSRenderer.cpp
```diff
@@ -139,59 +139,27 @@ bool GSRenderer::Merge(int field)
...
-			// dr[1] = 127 50 639 494
-
-			int top = min(fr[0].top, fr[1].top);
-			int bottom = min(fr[0].bottom, fr[1].bottom);
-
-			fr[0].top = fr[1].top = top;
-			fr[0].bottom = fr[1].bottom = bottom;
-
-			// blurdetected = true;
+			if (dr[0].eq(dr[1]) && (fr[0].eq(fr[1] + GSVector4i(0, j, 0, j)) || fr[1].eq(fr[0] + GSVector4i(0, j, 0, j))))
```
turtleli: There's got to be a better way than using a for loop. Maybe something along the lines of:
```
diff = fr[0] - fr[1];
if (dr[0].eq(dr[1]) && diff.x == diff.z && diff.x == 0 && diff.y == diff.w)
{
    // do whatever
```
Well, I think that's what your change is equivalent to.

FlatOutPS2: > There's got to be a better way than using a for loop.

I think my thinking got stuck inside of it...

turtleli: You need to fix the indentation of your new commit BTW.

FlatOutPS2: Late night text editor changes; not recommended...

plugins/GSdx/GSRenderer.cpp
```diff
@@ -237,18 +202,17 @@ bool GSRenderer::Merge(int field)
...
 		}
 		// If the DX offset is too small then consider the status of frame memory offsets, prevents blurring on Tenchu: Fatal Shadows, Worms 3D
-		else if((display_diff.x != frame_diff.x) && display_diff.x < 10 && frame_diff.x < 10)
+		else if(display_diff.x != frame_diff.x)
 		{
 			off.x = tex[i]->GetScale().x * frame_diff.x;
 		}
 
-		if(display_diff.y >= 4 && display_diff.y < 10) // Shouldn't this be >= 2?
+		if(display_diff.y >= 4) // Shouldn't this be >= 2?
```
ssakash: > // Shouldn't this be >= 2?

Maybe you could update this comment explaining why it's ``>=4`` instead?

plugins/GSdx/GSRenderer.cpp
```diff
@@ -139,42 +139,10 @@ bool GSRenderer::Merge(int field)
...
-			fr[0].top = top;
-			fr[1].top = top;
-			dr[0].bottom = bottom;
-			dr[1].bottom = bottom;
-
-			// blurdetected = true;
-		}
-		else if(dr[0].eq(dr[1]) && (fr[0].eq(fr[1] + GSVector4i(0, 1, 0, 1)) || fr[1].eq(fr[0] + GSVector4i(0, 1, 0, 1))))
+		int topDiff = fr[0].top - fr[1].top;
+		if (dr[0].eq(dr[1]) && (fr[0].eq(fr[1] + GSVector4i(0, topDiff, 0, topDiff)) || fr[1].eq(fr[0] + GSVector4i(0, topDiff, 0, topDiff))))
```
ssakash: > dr[0].eq(dr[1])

What about games which differ on the output circuit? God of war and Persona 4 for example as pointed out in the code comments.

```
            // persona 4:
			//
			// fr[0] = 0 0 640 448
			// fr[1] = 0 1 640 448
			// dr[0] = 159 50 779 498
			// dr[1] = 159 50 779 497
			//
			// second image shifted up by 1 pixel and blended over itself
			//
			// god of war:
			//
		    // fr[0] = 0 1 512 448
		    // fr[1] = 0 0 512 448
		    // dr[0] = 127 50 639 497
     	    // dr[1] = 127 50 639 498
		    //

FlatOutPS2: These cases are fixed by the display_diff/frame_diff code below, so it's become obsolete.

plugins/GSdx/GSRenderer.cpp
FlatOutPS2: And why is it `>=4` ? :P

ssakash: No clue, you were the one who added it :P

ssakash: Nevermind, it seems it has been 4 even before the 1.4 times.

plugins/GSdx/GSRenderer.cpp
gregory38: Could you keep the comment for history? Saying that is obsolete and that is fixed with code blabla..

FlatOutPS2: > Could you keep the comment for history? Saying that is obsolete and that is fixed with code blabla..

The entire code or just the Persona and GOW offset details?

Note that many games actually hit this code, not just those two games.

refractionpcsx2:
Well I've thrown some games at this and all seems well as normal :)

You okay with it @gregory38 as I see you weren't sure on something?

gregory38:
Fine PR was updated. You can merge it

willkuer:
This pr seems to introduce a regression in Warship Gunner 2 (#1723). Seems like depth handling broke.

refractionpcsx2:
Are you sure it's this commit? Those depth issues are usually issues involved with VIF and VU timing.

Nice timing btw :P Might have been more useful before the merge lol

disastranagant:
The recent VIF commits were why I was testing it but all recent builds but this one work.

refractionpcsx2:
Fair enough, didn't think this commit would cause that sort of problem.  If you can do a GS Dump, maybe @FlatOutPS2 would like to look in to it.

disastranagant:
dump https://www.dropbox.com/s/z8leoubpc3hrksr/warship%20gunner%202%20regression.7z?dl=0

FlatOutPS2:
Thanks for the dump. It has nothing to do with this PR. It's depth issue, not a PCRTC issue, so it would've been very odd for this to trigger it. See https://github.com/PCSX2/pcsx2/issues/1723 for details.

MrCK1:
Nascar 09 is all good - still not sure why it leaks memory though

Issue #1670, opened 2016-11-21
The Simpsons: Hit & Run Graphical Glitches


MarioSonic2987:
***PCSX2 version:***
1.5.0-dev-1637-g268b4d9

***PCSX2 options:***
Safest preset

***Plugins used:***
Default, except GSdx (AVX2). 

***Settings:*** 
- Renderer: HW/SW OGL
- Resolution: 3x Native
- Large Framebuffer
- Blending Unit: High
- CRC Hack level: Full
- Accurate Date

***Description of the issue:***
Lens flare are buggy and the textures surrounding it have a noticeable square. In SW, only have the lens flare issue.

![](http://i.imgur.com/PqH7T3s.jpg)
![](http://i.imgur.com/NcpD1Ah.jpg)
![](http://i.imgur.com/32fj3Tb.png)
The last picture is in SW mode.

[GS Dump HW](http://www.mediafire.com/file/a8d32aj1ml5zta9/gsdx_20161121150921.gs)
[GS Dump SW](http://www.mediafire.com/file/es7sj984iwkblyg/gsdx_20161121151322.rar)

***How to reproduce the issue:***
Focus the camera to the sun.

***Last known version to work:***
Probably never.

***PC specifications:***
 CPU: Intel Core i7 6700k GPU: GeForce GTX 970 OS: Windows 10 64 Bit

FlatOutPS2:
Set Blending Unit Accuracy to "None" will resolve the squares. And go to Config > Emulation Settings > Games Fixes tab, click "Enable manual game fixes" and enable "FPU Negative Div Hack" to resolve the lens flare issue.

MarioSonic2987:
These solved the issues. Thanks.

willkuer:
> "FPU Negative Div Hack" to resolve the lens flare issue.

Probably something for the gamedb?

MarioSonic2987:
Yeah.

PR #1676, opened 2016-11-25
GameDB: Add fixes for EA Sports team games and more

FlatOutPS2:
Add fixes for missing geometry issues in EA Sports team games(FIFA, NCAA Football, Madden NFL, NBA Live).

All NTSC-J and the Madden NFL PAL versions unconfirmed, but very unlikely to act any different.

Adds The Simpsons: Hit & Run lens flare fix (see issue https://github.com/PCSX2/pcsx2/issues/1670)

Fixed title for NCAA College Football 2K3 (game is not in compatibility list).

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -6537,6 +6537,7 @@ Region = PAL-G
 Serial = SLES-50021
 Name   = Madden NFL 2001
 Region = PAL-E
+vuClampMode = 3			// missing geometry with microVU
 ---------------------------------------------
 Serial = SLES-50022
 Name   = NBA Live 2001
@@ -7430,6 +7431,7 @@ Region = PAL-G
 Serial = SLES-50422
 Name   = Madden NFL 2002
 Region = PAL-E
+vuClampMode = 3			// missing geometry with microVU
 ---------------------------------------------
 Serial = SLES-50423
 Name   = F1 2001
@@ -9041,6 +9043,7 @@ Region = PAL-M8
 Serial = SLES-51154
 Name   = Madden NFL 2003
 Region = PAL-E
+vuClampMode = 3			// missing geometry with microVU
 ---------------------------------------------
 Serial = SLES-51156
 Name   = Silent Hill 2 - Director's Cut
@@ -9122,6 +9125,7 @@ vuClampMode = 2			//missing geometry with microVU
 Serial = SLES-51198
 Name   = NBA Live 2003
 Region = PAL-M5
+vuClampMode = 2			//missing geometry with microVU
 ---------------------------------------------
 Serial = SLES-51199
 Name   = 4x4 Evolution II
@@ -10285,6 +10289,7 @@ Compat = 5
 Serial = SLES-51797
 Name   = Madden NFL 2004
 Region = PAL-E
+vuClampMode = 3			// missing geometry with microVU
 ---------------------------------------------
 Serial = SLES-51798
 Name   = NHL 2004
@@ -10578,6 +10583,7 @@ Serial = SLES-51897
 Name   = Simpsons, The - Hit & Run
 Region = PAL-M4
 Compat = 5
+FpuNegDivHack = 1			// Lens flare appears even when behind objects
 ---------------------------------------------
 Serial = SLES-51903
 Name   = AFL Live 2004 - Aussie Rules Football
@@ -10702,6 +10708,7 @@ EETimingHack = 1 // Else serious slowdown due to VIF timing. Can be disabled onc
 Serial = SLES-51953
 Name   = FIFA 2004
 Region = PAL-M4
+vuClampMode = 2			//missing geometry with microVU
 ---------------------------------------------
 Serial = SLES-51954
 Name   = Max Payne 2 - The Fall of Max Payne
@@ -10728,10 +10735,12 @@ Region = PAL-M5
 Serial = SLES-51963
 Name   = FIFA 2004
 Region = PAL-F-G
+vuClampMode = 2			//missing geometry with microVU
 ---------------------------------------------
 Serial = SLES-51964
 Name   = FIFA 2004
 Region = PAL-P-S
+vuClampMode = 2			//missing geometry with microVU
 ---------------------------------------------
 Serial = SLES-51966
 Name   = Bombastic
@@ -10835,6 +10844,7 @@ Compat = 5
 Serial = SLES-52008
 Name   = NBA Live 2004
 Region = PAL-M5
+vuClampMode = 2			//missing geometry with microVU
 ---------------------------------------------
 Serial = SLES-52011
 Name   = Tak and The Power of Juju
@@ -10876,6 +10886,7 @@ Region = PAL-M5
 Serial = SLES-52025
 Name   = NFL Street
 Region = PAL-E
+vuClampMode = 3			// missing geometry with microVU
 ---------------------------------------------
 Serial = SLES-52026
 Name   = Wallace & Gromit in Project Zoo
@@ -11971,6 +11982,7 @@ Region = PAL-E
 Serial = SLES-52581
 Name   = Madden NFL 2005
 Region = PAL-E
+vuClampMode = 3			// missing geometry with microVU
 ---------------------------------------------
 Serial = SLES-52584
 Name   = Burnout 3 - Takedown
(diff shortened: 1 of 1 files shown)


avih:
Mandatory question: are all these fixes mentioned at the compatibility forum for each of the affected serials?

FlatOutPS2:
> Mandatory question: are all these fixes mentioned at the compatibility forum for each of the affected serials?

No, like I mentioned not the NTSC-J serials and the Madden NFL PAL serials. The others are confirmed.

> Assuming the "compatibility list" is the compatibility forum, please add it before we merge it.

I meant the list on the PCSX2 website, and those lead to the compatibility the forum, so yeah. There are so many games missing on the forum, no sense in adding just this one. :S

avih:
The list on the website is updated (automatically - I think daily) from the compatibility forum.

The forum is our master database because it includes the discussions and relevant info (such as who tested it, with which pcsx2 revision, etc). Therefore we must first update the forum, and only then update the DB which is distributed with pcsx2.

We have tools to import the compatibility level (and new entries) from the website data, and we use PRs, like this one, to update info which does not appear on the website - like game fixes.

So please make sure that every serial this PR modifies already has an entry at the compatibility forum, and it includes the gamefix info.

Alternatively, and that's up to @bositman, maybe we could have aggregate forum posts like "All tested EA sport games need ABC gamefix, we'd be adding them to the games db, but note that not all of them were tested". Though I still think it's be better to have entry per serial.

FlatOutPS2:
I don't know if you've noticed but many of the previous GameDB updates over the past few years contain games/serials which do not have correspondings threads in the compatibility forum. If you insist on this it means adding many new empty threads in the compatibility forum or not fixing many games at all...

avih:
> I don't know if you've noticed but many of the previous GameDB updates over the past few years contain games/serials which do not have correspondings threads in the compatibility forum. If you insist on this it means adding many new empty threads in the compatibility forum or not fixing many games at all...

I'm well aware of it and it needs to stop. It's going to be hard work to back-update the forum from the DB info, and therefore we don't plan to do that (though it would be nice if someone took it as a mini project).

However, we can make sure that every update to the DB is first documented at the forum, and if we don't have a forum entry for that serial, then it'd be a good time to add one.

refractionpcsx2:
There is a problem with adding games that have no entry on the forum.

Firstly the gameDB is compiled from sonyindex (I believe) so when you boot the game you at least get a game name etc, it hasn't just been random games.

Secondly, we have no "Unknown" status for forum compatibility and the reason it isn't on the forum is because nobody has it or hasn't tested it, so we can't actually put it on there unless somebody can test the game.

avih:
> Firstly the gameDB is compiled from sonyindex (I believe)

Initially yes, but that index hasn't updated in some years. Over the past few years the majority of the compatibility levels (1-5) were updated from the forum using `tools/GameIndex.dbf-tool/pcsx2-GameIndex.dbf-updater.html` and the games fixes etc were updated via manual PRs.

avih:
Why did you close it?

avih:
> Secondly, we have no "Unknown" status for forum compatibility and the reason it isn't on the forum is because nobody has it or nobody has tested it, so we can't actually put it on there unless somebody can test the game.

We can still add entries, state that no one tested it, but add $whatever reasons why we think it should have a compatibility value and/or game fixes ("e.g. known to use the same engine as game ABC - which was tested").

The point is to have a record of why the DB value is what it is. A guess is fine, but let's just have it documented.

refractionpcsx2:
I'm not sure what affect that would have on the website, that could balls it up completely.  You don't get to see any comments underneath, it just takes in to account the status= part, which we can't assign a value to.

avih:
Would you prefer people running that game in PCSX2, seeing it has a status 5, then go to the website or forum and not finding it there? not even a mention why it was declared playable (or any other status)?

The compatibility list on the website does have links back to the compatibility forum (expand the entry, then `Notes-Settings Used Click here!`). Better to state there whatever we want to state there, over stating nothing IMO.

refractionpcsx2:
> Would you prefer people running that game in PCSX2, seeing it has a status 5, then go to the website or forum and not finding it there? not even a mention why it was declared playable (or any other status)?

why would they go to the forum to find a compatibility report if the game was playable? I doubt anybody would care.  It's only when they don't work properly they might go and look.  But these will report as "unknown" on PCSX2 as they don't have a compatibility status, in this case it will just add a most likely required gamefix, but it doesn't promise any sort of status.

avih:
> why would they go to the forum to find a compatibility report if the game was playable?

Because it might not actually be playable. It just says so at the PCSX2 console dir to the DB entry, but it doesn't appear at the compatibility site or forum. Where does the user go from there?

FlatOutPS2:
> Why did you close it?

- Because it would require creating 15-20 new threads on forum, and updating the titles of several others.
- All that for a few empty threads.
- Only a handful of people have the rights to make/edit threads in the compatibility forum, not including myself so I can't do the work.

> Where does the user go from there?

The wiki.

avih:
> The wiki.

If you think the wiki is a better place than the compatibility forum, why do we even keep the forum? And who vouches for the info at the wiki? or guarantees it's there to begin with? recall, you're adding compatibility info based on some whatever guesses (which might be great), but you refuse to document it anywhere.

FlatOutPS2:
> And who vouches for the info at the wiki? or guarantees it's there to begin with? 

Who vouches for the info on the forum. Anyone can make compatibility reports just like anyone can add to the wiki.

> you're adding compatibility info based on some whatever guesses (which might be great), but you refuse to document it anywhere.

It's documented in the database file, here on github and on the wiki.

refractionpcsx2:
> Because it might not actually be playable - because no one has actually tested it. It just says so at the PCSX2 console due to the DB entry, but it doesn't appear at the compatibility site or forum. Where does the user go from there?

You completely ignored the bit where I said none of these games display a compatibility status and will say "Unknown".  These debate is completely irrelevant to this PR.

avih:
> It's documented in the database file, here on github and on the wiki.

The DB doesn't document that it's an assessment rather that the result of someone actually testing it. The wiki doesn't either, and without me asking here, neither had this githib issue.

avih:
> You completely ignored the bit ...

Not intentionally.

Regardless, the forum contains the most detailed compatibility info about games, and it's our de-facto master list. By making changes to the DB without mentioning the serials at the forum you're de-valuing the forum, and makes it impossible to have a single master list of compatibility info.

And for no real good reason either. it's hard to find evidence for past entries, but for new enteries we know 100% why we're changing them, like here. That's fine, but let's just try to keep a single master list for it, for the sake of useful records.

FlatOutPS2:
> The DB doesn't document that it's an assessment rather that the result of someone actually testing it. The wiki doesn't either, and without me asking here, neither had this githib issue.

You can't rely on (especially older) compatibility reports either. They often have EE/VU clamping or rounding mode to non-default values without mentioning why, say no bugs encountered but even testing with the same PCSX2 revision they used plenty of very obvious bugs appear they can't have missed. I believe some are just copy-paste reports and we can't seriously assess whether the game was actually tested or just a copied report of the previous entry in the series.

avih:
> You can't rely on ...

Let's assume that's correct (and I have no reason to assume otherwise). Do you think this implies that it's useless? that we shouldn't document _current_ changes while we know 100% why we're making them? Shall we just drop the compatibility forum and that's it?

Or, alternatively, realize that past info might be unreliable, but document the best we can every new info we add?

If you say "past info is sh*tty, not worth documenting new info either", then that's fine. And let's drop the compatibility forum and website which draws from it.

FlatOutPS2:
It's worth documenting credible reports, not making 20 new threads without a full test report for just a single fix. 

I'd rather see a new single thread in the compatibility forum were we keep a list of games that aren't fully tested, but do have some compatibility info(like these fixes) on. And add a request for users to make full test reports for these games. By directly requesting these, we might even see more reports posted.

avih:
But I'm not gonna fight over this. If you guys think it's fine to update DB entries without mentioning them at the forum, go for it. I don't think it's great, but if others do, I'll live fine with it.

avih:
> I'd rather see a new single thread in the compatibility forum were we keep a list of games that aren't fully tested, but do have some compatibility info(like these fixes) on. And add a request for users to make full test reports for these games. By directly requesting these, we might even see more reports posted.

This is fine and I even suggested it earlier myself.

PR #1679, opened 2016-11-30
LilyPad: GUI update and add new functions

FlatOutPS2:
Updates the UI by reducing the height of the plugin window. This has
been achieved by removing some buttons below the diagnostics and
bindings list and incorporating those functions into the
lists(accessible by right-clicking in the list). The binding
configurations on the Pad tabs have been moved to a separate page, like
the Forcefeedback bindings, to separate the configuration from the
bindings.

And adds three new options:
- **Adds a skip deadzone option to the Pad tabs**

With the normal deadzone, if the control input value is below the
deadzone threshold, the input is ignored.
However, some controllers also benefit from shortening the input range
by skipping a deadzone.

- **Add Configure on bind option**

Part of the GUI update, this function switches to the configuration page
immediately after binding an input instead of staying on the bindings
page.

- **LilyPad: Add Device Select option**

Adds a device select option that hides bindings and disables binding new
inputs from all non-selected devices on the bindings list. This also
avoids input conflict issues when one controller is recognized as
several devices through different APIs.

Note: Some of the Skip deadzone user interface code is part of the GUI update commit.


FlatOutPS2:
Build failed because it didn't include the latest build fix commit from the master. :/

plugins/LilyPad/LilyPad.cpp
```diff
@@ -384,8 +384,15 @@ void AddForce(ButtonSum *sum, u8 cmd, int delta = 255)
...
 void ProcessButtonBinding(Binding *b, ButtonSum *sum, int value)
 {
-    if (value < b->deadZone || !value)
+    if (value < b->deadZone || !value) {
         return;
+    }
+    if (b->skipDeadZone > b->deadZone) {
+        value = (int)(((__int64)value * (FULLY_DOWN - (__int64)b->skipDeadZone)) / FULLY_DOWN) + b->skipDeadZone;
+        if (value > 0x10000) {
+            value = 0x10000;
```
ssakash: Maybe ``std::max(value,0x10000)`` ?

plugins/LilyPad/Tooltips.cpp
```diff
@@ -57,10 +57,23 @@ LPWSTR dialog_message(int ID, bool *updateText)
...
-            return L"Displays a continuously updated list of the state of all objects on the selected device.\n\n"
-                   L"Use this option to check if all the inputs on a controller function properly.";
+        case IDC_DIAG_LIST:
+            return L"Shows a list of currently available input devices.\n\n"
+                   L"Double-click a device in the list or right-click it and select \"Test Device\" to display a continuously updated list of the state of all inputs on the selected device.\n"
+                   L"Use this option to check if all the inputs on a controller function properly.\n\n"
+                   L"Right-click and select \"Refresh\" to update the list of devices in case a recently connected device has not shown up yet.";
         // Pad tabs
+        case IDC_BINDINGS_LIST:
+            return L"Shows a list ofcurrently bound inputs of the selected Pad.\n\n"
```
ssakash: Space after ``of`` in ``ofcurrently``?

plugins/LilyPad/Tooltips.cpp
```diff
@@ -57,10 +57,23 @@ LPWSTR dialog_message(int ID, bool *updateText)
...
+                   L"Left-click on one the bindings in the list to configure it.\n\n"
```
ssakash: ``one``?

plugins/LilyPad/resource.h
```diff
@@ -91,18 +96,17 @@
...
+#define IDC_CONFIGURE_ON_BIND           0x111F
+#define IDC_DEVICE_SELECT               0x1120
 #define ID_SAVE                         0x1204
 #define ID_LOAD                         0x1205
-#define IDC_LIST                        0x1207
-#define IDC_FORCEFEEDBACK               0x1208
-#define IDC_FORCEFEEDBACK_FUNCTION      0x1209
-#define ID_BIG_MOTOR                    0x120A
-#define ID_SMALL_MOTOR                  0x120B
+#define IDC_BINDINGS_LIST               0x1207
```
ssakash: Might be nice to follow arithmetic progression in the macro order but honestly I don't care.

plugins/LilyPad/LilyPad.cpp
```diff
@@ -41,8 +41,8 @@
 #define FORCE_UPDATE_WPARAM ((WPARAM)0x74328943)
 #define FORCE_UPDATE_LPARAM ((LPARAM)0x89437437)
 
-// LilyPad version.
-#define VERSION ((0 << 8) | 11 | (0 << 24))
+// LilyPad version. Last updated on: November 29th, 2016
```
ssakash: > Last updated on: November 29th, 2016
Maybe this part isn't needed? It would be a PITA for future contributors if they want to keep updating the comment with the date, also timezone problems/commit merge date conflicts?

plugins/LilyPad/LilyPad.cpp
```diff
@@ -384,8 +384,15 @@ void AddForce(ButtonSum *sum, u8 cmd, int delta = 255)
 
 void ProcessButtonBinding(Binding *b, ButtonSum *sum, int value)
 {
-    if (value < b->deadZone || !value)
+    if (value < b->deadZone || !value) {
```
ssakash: Well, it's an ``int``, so ``value == 0`` might be more suitable but just my opinion. :)

plugins/LilyPad/Tooltips.cpp
ssakash: Was the word ``one`` intentional in the statement? It certainly doesn't make sense to me.

plugins/LilyPad/LilyPad.cpp
FlatOutPS2: Shouldn't it be std:min(value, 0x10000) then?

ssakash: Yeah, I screwed it up there. Nice catch :P

plugins/LilyPad/Tooltips.cpp
FlatOutPS2: Yes, it was intentional. Why does it not make sense to you?

MrCK1: Also forgot a space between ```of``` and ```currently``` in the line above if you didn't see :p

ssakash: > Also forgot a space between of and currently in the line above if you didn't see :p

I reported that already : https://github.com/PCSX2/pcsx2/pull/1679#discussion_r90237235

FlatOutPS2: I fixed that already: https://github.com/PCSX2/pcsx2/pull/1679/commits/4b29064cd7af228ff5396a1a6b27f49bb0a42d7d#diff-412bcec8b699ca8df71fcf9e117c9d18R68

Though the VS2013 build seems to have failed without stating what went wrong.

plugins/LilyPad/LilyPad.cpp
```diff
@@ -428,7 +432,7 @@ void CapSum(ButtonSum *sum)
 #ifdef __linux__
         int div = std::max(abs(sum->sticks[i].horiz), abs(sum->sticks[i].vert));
 #else
-        int div = max(abs(sum->sticks[i].horiz), abs(sum->sticks[i].vert));
+        int div = std::max(abs(sum->sticks[i].horiz), abs(sum->sticks[i].vert));
```
turtleli: The if and else are the same.

plugins/LilyPad/Config.cpp
```diff
@@ -1181,7 +1241,7 @@ void AddTooltip(UINT id, HWND hWnd)
         return;
 
     TOOLINFO toolInfo = {0};
-    toolInfo.cbSize = sizeof(toolInfo);
+    toolInfo.cbSize = tooltipStructSize;
```
turtleli: I assume tooltips weren't working on other emus?

plugins/LilyPad/Tooltips.cpp
```diff
@@ -57,10 +57,23 @@ LPWSTR dialog_message(int ID, bool *updateText)
...
+            return L"Shows a list of currently bound inputs of the selected Pad.\n\n"
+                   L"Left-click on one of the bindings in the list to configure it.\n\n"
+                   L"Right-click and select \"Delete Selected\" to remove the selected input from the list.\n\n"
+                   L"Right-click and select \"Clear All\" to remove all the inputs from the list.\n\n"
+                   L"Note: Use Shift/Ctrl + Left-click to select multiple bindings. Changing the displayed configuration will now copy it to all selected bindings.";
+        case IDC_DEVICE_SELECT:
+            return L"Select a single device to hide the bindings from other devices in the bindings list, and to only be able to add new bindings for the selected device.\n\n"
+                   L"This can also avoid input conflict issues when one controller is recognized as several devices through different APIs.";
+        case IDC_CONFIGURE_ON_BIND:
+            return L"Immediately go to the configuration setup when you create a new binding.\n\n";
```
turtleli: `\n\n` is pointless at the end of a string.

plugins/LilyPad/Config.cpp
FlatOutPS2: Yes. :P

turtleli:
What situations would "Configure on Bind" be useful for? I can't imagine it being used much.

FlatOutPS2:
> What situations would "Configure on Bind" be useful for? I can't imagine it being used much.

It isn't meant to be used much, that's why it's off by default. Binding keyboard or X360 controller inputs usually won't require any configuration at all. It's useful when setting up several pressure sensitive inputs or analog axis that require independent configurations.

MrCK1:
The border above "Pads" seems a bit too close together now that the bottom lip is missing. Can we get rid/combine some of the grey borders?

lilypadgui

FlatOutPS2:
> Edit: Why not relocate the Port Multitap checkboxes to the other side of the text box (under the Analog Only options) It looks to be a better fit than its current location IMO :P

Because the multitap checkboxes affect the pad list(they add more pads), and the options on the right side are individual options for each pad you select in the list.

> The border area above "Pads" seems a bit too close together now that some of the white space is removed. I think the same applies to the area around the miscellaneous box. Can we get rid/combine some of the grey borders?

What about this:

lp

MrCK1:
@FlatOutPS2 That looks better. Thanks for the reply

turtleli:
> It isn't meant to be used much, that's why it's off by default. Binding keyboard or X360 controller inputs usually won't require any configuration at all. It's useful when setting up several pressure sensitive inputs or analog axis that require independent configurations.

Well, I think it's more likely a user will set up all bindings, play a bit, change sensitivity, play a bit, change sensitivity a bit more than use Configure to Bind. Though perhaps I'm just unable to think of a scenario (I don't spend much time tinkering pad settings myself so perhaps I'm not the right person to judge how useful this is). Keep it if you think it's useful.

As for GUI design, another approach could be input API + Advanced on the left, everything else on the right (this is a completely random suggestion, so if you think it'll be horrible, just ignore it).

FlatOutPS2:
> Well, I think it's more likely a user will set up all bindings, play a bit, change sensitivity, play a bit, change sensitivity a bit more than use Configure to Bind. Though perhaps I'm just unable to think of a scenario (I don't spend much time tinkering pad settings myself so perhaps I'm not the right person to judge how useful this is). Keep it if you think it's useful.

I'll keep it. Even if it only helps a few users, that might be more than some of the other functions that are already in place. :)

> As for GUI design, another approach could be input API + Advanced on the left, everything else on the right (this is a completely random suggestion, so if you think it'll be horrible, just ignore it).

I didn't really like how that looked, but what about this:

l

turtleli:
I like the Advanced stuff being moved out of the Input API groupbox - I'm not sure any of the options belong there. But I don't like how the Input API radioboxes and checkboxes aren't 1 item per row - it feels less neat to me.

But yeah, my GUI suggestions are more just to provide thought - I don't have any design in mind, and there's always the option of another PR if you wish to change the layout (so that we can have the additional  functionality merged sooner).

FlatOutPS2:
> I like the Advanced stuff being moved out of the Input API groupbox - I'm not sure any of the options belong there. But I don't like how the Input API radioboxes and checkboxes aren't 1 item per row - it feels less neat to me.

I did consider replacing the KB and Mouse radioboxes with a combobox, and keeping the Gamedevices checkboxes 1 per row.

> But yeah, my GUI suggestions are more just to provide thought - I don't have any design in mind, and there's always the option of another PR if you wish to change the layout (so that we can have the additional functionality merged sooner).

I pushed the commit that adds the fixes I made in responce to @MrCK1 remarks. I'll leave further updates for the next PR.

willkuer:
> As for GUI design, another approach could be input API + Advanced on the left, everything else on the right (this is a completely random suggestion, so if you think it'll be horrible, just ignore it).

> I didn't really like how that looked

Just that we are talking about the same?

vertical alignment

turtleli:
> Just that we are talking about the same?

Not quite sure who you're addressing, but yeah, my suggestion was along those lines.

turtleli:
@FlatOutPS2 Code seems ok, so you should split and squash the fixup commits into the appropriate commits.

lightningterror:
I recommend removing the border on the right side of the "Bindings" column.

FlatOutPS2:
I did some very minor offset improvements based on the results I got from using the plugin on other emu's, and what I assume @lightningterror meant. :p

lightningterror:
I was talking about this one in the image :)

disbd

FlatOutPS2:
> I was talking about this one in the image :)

Great, because that's the one i mended. :D

turtleli:
>> I was talking about this one in the image :)

> Great, because that's the one I mended. :D

Great. Now you get additional horizontal scrollbars if the multitap options are enabled ;)

FlatOutPS2:
> Great. Now you get additional horizontal scrollbars if the multitap options are enabled ;)

I fixed it just in case there's someone who won't think it's great. :)

turtleli:
Ok. Could you squash the bindings commit into the GUI update? It'll probably be ready for merge after that.

Side note: The list view column widths aren't great at higher DPI scaling settings. Just in case you want to work on that in a future PR ;)

PR #1681, opened 2016-12-01
PCSX2-GS: Use interrupt mask register bitfields

ssakash:
**Summary of changes**:

* Use the IMR register union to get mask interrupt value of specific bit-fields of the register instead of using multiple bitwise operations all over the place to get the required field.

 I assume this was actually needed to be done but never got done for some reason? As the union was left there without any purpose before this commit.

### DIFF
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -450,10 +450,8 @@ static __fi void VSyncStart(u32 sCycle)
 	if (!CSRreg.VSINT)
 	{
 		CSRreg.VSINT = true;
-		if (!(GSIMR&0x800))
-		{
+		if (!GSIMR.VSMSK)
 			gsIrq();
-		}
 	}
 
 	hwIntcIrq(INTC_VBLANK_S);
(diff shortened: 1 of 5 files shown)


ssakash:
Oh, I forgot that refraction is at Switzerland right now. So this would probably get merged some time later then.

If there's anyone with more knowledge on the interrupt masks , feel free to review the commit.

FlatOutPS2:
> Oh, I forgot that refraction is at Switzerland Sweden right now. 

Are you sure he's not in Swaziland?

ssakash:
> Are you sure he's not in Swaziland?

Well, I have no clue what's Swaziland to be honest. Though I'm highly sure that he's in Sweden as he earlier posted a picture with a ``hotell`` label which stands for ``hotel`` in Swedish.

FlatOutPS2:
> Well, I have no clue what's Swaziland to be honest. Though I'm highly sure that he's in Sweden as he earlier posted a picture with a hotell label which stands for hotel in Swedish.

Or he might be in a hotel in Swindon that's so bad it isn't allowed to be called a hotel and they had to add an l for legal reasons.

ssakash:
> Or he might be in a hotel in Swindon that's so bad it isn't allowed to be called a hotel and they had to add an l for legal reasons.

You have an enigmatic imagination skill :P  Though it seems like @refractionpcsx2 is indeed present in Sweden right now, he said he'll come back home on Saturday.

ref

refractionpcsx2:
I was indeed in Sweden :P

But yeh, this looks fine to me. It probably just never got added to this code, it was probably just not considered when it was written.

gregory38:
I'm fine too.

PR #1683, opened 2016-12-02
GSDX: Fix output texture height calculation

ssakash:
**Summary of changes**:

* Properly calculate the texture height on ``GetOutput()``, the bug was shared by all the renderers.

Previously, the value of the frame offset was also considered for the total height of the texture which was obviously wrong as the portion before the offset value isn't part of the frame memory. _Might_ help on fixing some GSdx bugs.

### DIFF
--- a/plugins/GSdx/GSRendererCL.cpp
+++ b/plugins/GSdx/GSRendererCL.cpp
@@ -169,7 +169,7 @@ GSTexture* GSRendererCL::GetOutput(int i, int& y_offset)
 	const GSRegDISPFB& DISPFB = m_regs->DISP[i].DISPFB;
 
 	int w = DISPFB.FBW * 64;
-	int h = GetFrameRect(i).bottom;
+	int h = GetFrameRect(i).height();
 
 	// TODO: round up bottom
 
--- a/plugins/GSdx/GSRendererCS.cpp
+++ b/plugins/GSdx/GSRendererCS.cpp
@@ -345,7 +345,7 @@ GSTexture* GSRendererCS::GetOutput(int i, int& y_offset)
 	const GSRegDISPFB& DISPFB = m_regs->DISP[i].DISPFB;
 
 	int w = DISPFB.FBW * 64;
-	int h = GetFrameRect(i).bottom;
+	int h = GetFrameRect(i).height();
 
 	// TODO: round up bottom
 
--- a/plugins/GSdx/GSRendererHW.cpp
+++ b/plugins/GSdx/GSRendererHW.cpp
@@ -219,7 +219,7 @@ GSTexture* GSRendererHW::GetOutput(int i, int& y_offset)
 
 	GSTexture* t = NULL;
 
-	if(GSTextureCache::Target* rt = m_tc->LookupTarget(TEX0, m_width, m_height, GetFrameRect(i).bottom))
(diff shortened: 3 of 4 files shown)


gregory38:
I don't think it is obvious. Beside nobody report any issue on this code ;) Potentially it only wastes memory.

FlatOutPS2:
You might as well replace it with GetDisplayRect().height as there won't be a difference.

ssakash:
> I don't think it is obvious. Beside nobody report any issue on this code ;) Potentially it only wastes memory.

Potentially this might fix the crash issues in NASCAR series games as we no longer consider those massive offset values.

> You might as well replace it with GetDisplayRect().height as there won't be a difference.

Well, the surrounding code uses a reference to FBW (framebuffer width) which is a reference to the frame memory, so for consistency I personally don't mind using ``GetFrameRect().height()`` instead.

FlatOutPS2:
> Potentially this might fix the crash issues in NASCAR series games as we no longer consider those massive offset values.

It will.

> Well, the surrounding code uses a reference to FBW (framebuffer width) which is a reference to the frame memory, so for consistency I personally don't mind using GetFrameRect().height() instead.

Actually, the entire GetFrameRect() is actually useless as it's values aren't used anywhere but the MergeCircuit anymore. It can just be merged into the MergeCircuit code.

ssakash:
> Actually, the entire GetFrameRect() is actually useless as it's values aren't used anywhere but the MergeCircuit anymore. It can just be merged into the MergeCircuit code.

Actually, I missed an important detail in my previous post on how ``GetFrameRect()`` differs from ``GetDisplayRect()``, we do special handling for games with double scan flag enabled on ``GetFrameRect()`` . For example, check games like ``Disgaea 2`` which have double scan flag enabled. Haven't checked it yet, but I'm 100% sure that replacing ``GetFrameRect().height()`` with ``GetDisplayRect().height`` would cause issues on such games.

https://github.com/PCSX2/pcsx2/blob/master/plugins/GSdx/GSState.cpp#L451
```
if (isinterlaced() && m_regs->SMODE2.FFMD && h > 1)
		h >>= 1;
```

FlatOutPS2:
> Actually, I missed an important detail in my previous post on how GetFrameRect() differs from GetDisplayRect(), we do special handling for games with double scan flag enabled on GetFrameRect() . For example, check games like Disgaea 2 which have double scan flag enabled. Haven't checked it yet, but I'm 100% sure that replacing GetFrameRect().height() with GetDisplayRect().height would cause issues on such games.

You could replace h = GetFrameRect().height with `h = isinterlaced() && m_regs->SMODE2.FFMD && h > 1 ? GetDisplayRect(i).height >>=1 : GetDisplayRect(i).height` or just reduce `GetFrameRect()` to `GetFrameHeight()`.

ssakash:
Fine, I have some future plans to swap around the merge circuit code to deal with some special cases on double scan games, I'll probably do it then. Not planning on delaying the merge of this PR. :)

gregory38:
Hum, FMV that requires a large framebuffer need to be tested.

FlatOutPS2:
> Hum, FMV that requires a large framebuffer need to be tested.

Do we have a list of those? :p

gregory38:
I'm too busy right now but maybe I could push all my gsdump into my google drive (2GB). There are xz compressed but support can be added on Windows too. Or maybe we have some room in PCSX2 server. This way we could merge all gs dump into a single repository.

PR #904, opened 2015-10-19
Autoselect CRC Hack by renderer

willkuer:
Inline function to select the appropriate crc level by renderer.
Accordingly new Combobox Element 'Automatic'

Added as well a (at least to me) nice renderer enum and replaced all
renderer int's by the enum.

Defined default renderer device according to OS.

PR #911, opened 2015-10-21
Gsdx: Add Mipmap (Software mode) and preload data frame to GUI.

ssakash:
~~Windows build for testing : https://www.dropbox.com/s/o4xb7n2z1fo6a1s/GSdx32-SSE2.dll?dl=0~~ outdated build of the commit.

PR #1686, opened 2016-12-05
Automatic CRC hack level

willkuer:
Adds an automatic CRC Hack level to the GUI. The value will be interpreted as either 'Partial' or 'Full' depending on the renderer. Was already discussed in #1582, #904, and #882. It's basically the same as FlatOut's approach.

Additionally added an enum class for the CRCHackLevel. Although the usage is not completely consequently done in consumer code. The CRC hack section largely depends on the ordering of the enum as CRC hacks on the partial level are as well applied on the full level. Reordering of the enum values would break the code. But that would as well break the backwards compatibility of the ini-File. A more strict approach to use an enum (maybe as flag) would need a larger non-functional rewrite.

Additionally added a template-version of GetConfig (GetConfigT) to avoid the casts in the consumer code and possibly improve readability.

### DIFF
--- a/plugins/GSdx/GS.cpp
+++ b/plugins/GSdx/GS.cpp
@@ -506,6 +506,8 @@ EXPORT_C_(int) GSopen2(void** dsp, uint32 flags)
 	static bool stored_toggle_state = false;
 	bool toggle_state = !!(flags & 4);
 
+	theApp.ClearTempConfig();
+
 	GSRendererType renderer = s_renderer;
 	// Fresh start up or config file changed
 	if (renderer == GSRendererType::Undefined)
--- a/plugins/GSdx/GS.cpp
+++ b/plugins/GSdx/GS.cpp
@@ -212,7 +212,7 @@ static int _GSopen(void** dsp, const char* title, GSRendererType renderer, int t
 
 	if(renderer == GSRendererType::Undefined)
 	{
-		renderer = static_cast<GSRendererType>(theApp.GetConfigI("Renderer"));
+		renderer = theApp.GetTempConfig<GSRendererType>();
 	}
 
 	if(threads == -1)
@@ -512,11 +512,7 @@ EXPORT_C_(int) GSopen2(void** dsp, uint32 flags)
 	// Fresh start up or config file changed
 	if (renderer == GSRendererType::Undefined)
 	{
-		renderer = static_cast<GSRendererType>(theApp.GetConfigI("Renderer"));
-#ifdef _WIN32
-		if (renderer == GSRendererType::Default)
-			renderer = GSUtil::GetBestRenderer();
-#endif
+		renderer = theApp.GetTempConfig<GSRendererType>();
 	}
 	else if (stored_toggle_state != toggle_state)
 	{
@@ -545,6 +541,7 @@ EXPORT_C_(int) GSopen2(void** dsp, uint32 flags)
 		default: renderer = GSRendererType::OGL_SW; break; // fallback to OGL SW
 		}
 #endif
+		theApp.SetTempConfig<GSRendererType>(renderer);
 	}
 	stored_toggle_state = toggle_state;
 
@@ -590,7 +587,7 @@ EXPORT_C_(int) GSopen(void** dsp, const char* title, int mt)
 	{
 		// normal init
 
-		renderer = static_cast<GSRendererType>(theApp.GetConfigI("Renderer"));
+		renderer = theApp.GetTempConfig<GSRendererType>();
 	}
 
 	*dsp = NULL;
@@ -1566,7 +1563,7 @@ EXPORT_C GSReplay(char* lpszCmdLine, int renderer)
 
 	GSRendererType m_renderer;
 	// Allow to easyly switch between SW/HW renderer -> this effectively removes the ability to select the renderer by function args
-	m_renderer = static_cast<GSRendererType>(theApp.GetConfigI("Renderer"));
+	m_renderer = theApp.GetTempConfig<GSRendererType>();
 
 	if (m_renderer != GSRendererType::OGL_HW && m_renderer != GSRendererType::OGL_SW)
 	{
--- a/plugins/GSdx/GS.h
+++ b/plugins/GSdx/GS.h
@@ -1429,6 +1429,7 @@ enum class Filtering : uint8
 //Enum order matters! Don't change it!
 enum class CRCHackLevel : int8
 {
+	Automatic = -1,
 	None,
 	Minimum,
 	Partial,
--- a/plugins/GSdx/GS.h
+++ b/plugins/GSdx/GS.h
@@ -1425,3 +1425,13 @@ enum class Filtering : uint8
 	Trilinear_Bilinear_Forced,
 	Trilinear_Always
 };
+
+//Enum order matters! Don't change it!
+enum class CRCHackLevel : int8
+{
+	None,
+	Minimum,
+	Partial,
+	Full,
+	Aggressive
+};
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -2433,7 +2433,7 @@ void GSState::SetupCrcHack()
 {
 	GetSkipCount lut[CRC::TitleCount];
 
-	s_crc_hack_level = theApp.GetConfigT<CRCHackLevel>("crc_hack_level");
+	s_crc_hack_level = theApp.GetTempConfig<CRCHackLevel>();
 
 	memset(lut, 0, sizeof(lut));
 
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -23,11 +23,12 @@
 #include "GSState.h"
 #include "GSdx.h"
 
-int s_crc_hack_level = 3;
+CRCHackLevel s_crc_hack_level = CRCHackLevel::Full;
 
 // hacks
-#define Aggresive (s_crc_hack_level > 3)
-#define Dx_only   (s_crc_hack_level > 2)
+#define Aggresive  (s_crc_hack_level >= CRCHackLevel::Aggressive)
+#define Dx_only    (s_crc_hack_level >= CRCHackLevel::Full)
+#define Dx_and_OGL (s_crc_hack_level >= CRCHackLevel::Partial)
 
 CRC::Region g_crc_region = CRC::NoRegion;
 
@@ -2432,11 +2433,11 @@ void GSState::SetupCrcHack()
 {
 	GetSkipCount lut[CRC::TitleCount];
 
-	s_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+	s_crc_hack_level = theApp.GetConfigT<CRCHackLevel>("crc_hack_level");
 
 	memset(lut, 0, sizeof(lut));
 
-	if (s_crc_hack_level > 1) {
+	if (Dx_and_OGL) {
 		lut[CRC::AceCombat4] = GSC_AceCombat4;
 		lut[CRC::AlpineRacer3] = GSC_AlpineRacer3;
 		lut[CRC::BlackHawkDown] = GSC_BlackHawkDown;
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -23,8 +23,6 @@
 #include "GSState.h"
 #include "GSdx.h"
 
-extern CRCHackLevel g_crc_hack_level;
-
 //#define Offset_ST  // Fixes Persona3 mini map alignment which is off even in software rendering
 
 int GSState::s_n = 0;
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -23,7 +23,7 @@
 #include "GSState.h"
 #include "GSdx.h"
 
-extern int g_crc_hack_level;
+extern CRCHackLevel g_crc_hack_level;
 
 //#define Offset_ST  // Fixes Persona3 mini map alignment which is off even in software rendering
 
@@ -79,7 +79,7 @@ GSState::GSState()
 	//s_savel = 0;
 
 	UserHacks_WildHack = theApp.GetConfigB("UserHacks") ? theApp.GetConfigI("UserHacks_WildHack") : 0;
-	m_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+	m_crc_hack_level = theApp.GetTempConfig<CRCHackLevel>();
 
 	memset(&m_v, 0, sizeof(m_v));
 	memset(&m_vertex, 0, sizeof(m_vertex));
@@ -2494,7 +2494,7 @@ void GSState::SetGameCRC(uint32 crc, int options)
 {
 	m_crc = crc;
 	m_options = options;
-	m_game = CRC::Lookup(m_crc_hack_level ? crc : 0);
+	m_game = CRC::Lookup(m_crc_hack_level > CRCHackLevel::None ? crc : 0);
 	SetupCrcHack();
 
 	// Until we find a solution that work for all games.
--- a/plugins/GSdx/GSTextureCache.cpp
+++ b/plugins/GSdx/GSTextureCache.cpp
@@ -48,7 +48,7 @@ GSTextureCache::GSTextureCache(GSRenderer* r)
 
 	m_paltex = theApp.GetConfigB("paltex");
 	m_can_convert_depth &= s_IS_OPENGL; // only supported by openGL so far
-	m_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+	m_crc_hack_level = theApp.GetConfigT<CRCHackLevel>("crc_hack_level");
 
 	// In theory 4MB is enough but 9MB is safer for overflow (8MB
 	// isn't enough in custom resolution)
@@ -271,7 +271,7 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 
 						// Gregory: to avoid a massive slow down for nothing, let's only enable
 						// this code when CRC is below the FULL level
-						if (m_crc_hack_level < 3)
+						if (m_crc_hack_level < CRCHackLevel::Full)
 							Read(t, t->m_valid);
 						else
 							dst = t;
--- a/plugins/GSdx/GSTextureCache.cpp
+++ b/plugins/GSdx/GSTextureCache.cpp
@@ -28,7 +28,7 @@ bool GSTextureCache::m_disable_partial_invalidation = false;
 GSTextureCache::GSTextureCache(GSRenderer* r)
 	: m_renderer(r)
 {
-	s_IS_OPENGL = (static_cast<GSRendererType>(theApp.GetConfigI("Renderer")) == GSRendererType::OGL_HW);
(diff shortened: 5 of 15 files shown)


plugins/GSdx/GSdx.cpp
```diff
@@ -218,11 +218,11 @@ void GSdxApp::Init()
 	m_gs_hw_mipmapping.push_back(GSSetting(1, "Basic", "Fast"));
 	m_gs_hw_mipmapping.push_back(GSSetting(2, "Full", "Slow"));
 
-	m_gs_crc_level.push_back(GSSetting(0 , "None", "Debug"));
-	m_gs_crc_level.push_back(GSSetting(1 , "Minimum", "Debug"));
-	m_gs_crc_level.push_back(GSSetting(2 , "Partial", "OpenGL Recommended"));
-	m_gs_crc_level.push_back(GSSetting(3 , "Full", "Safest"));
-	m_gs_crc_level.push_back(GSSetting(4 , "Aggressive", ""));
+	m_gs_crc_level.push_back(GSSetting(static_cast<int8>(CRCHackLevel::None) , "None", "Debug"));
```
willkuer: static_cast can be removed. GSSetting already supports direct casting via template.

willkuer:
I almost forgot how much fun it is to work with git...

Well obviously nobody needs the build bot atm.

plugins/GSdx/GSHwHack.cpp
```diff
@@ -23,11 +23,12 @@
...
 
-int s_crc_hack_level = 3;
+CRCHackLevel s_crc_hack_level = CRCHackLevel::Full;
 
 // hacks
-#define Aggresive (s_crc_hack_level > 3)
-#define Dx_only   (s_crc_hack_level > 2)
+#define Aggresive  (static_cast<int8>(s_crc_hack_level) >= static_cast<int8>(CRCHackLevel::Aggressive))
+#define Dx_only    (static_cast<int8>(s_crc_hack_level) >= static_cast<int8>(CRCHackLevel::Full))
+#define Dx_and_OGL (static_cast<int8>(s_crc_hack_level) >= static_cast<int8>(CRCHackLevel::Partial))
```
turtleli: I don't think it's necessary to cast to int when you're comparing enums of the same type. `s_crc_hack_level >= CRCHackLevel::Whatever` should work.

plugins/GSdx/GS.h
```diff
@@ -1425,3 +1425,25 @@ enum class Filtering : uint8
...
+
+static inline CRCHackLevel AutoSelectCRCHackLevel(GSRendererType renderer)
+{
+	switch (renderer)
+	{
+	case GSRendererType::OGL_HW:
+		return CRCHackLevel::Partial;
+	default:
+		return CRCHackLevel::Full;
+	}
```
ssakash: ``switch-case`` doesn't seem ideal here, a ternary operator should suffice. `` return (renderer == GSRendererType::OGL_HW) ? CRCHackLevel::Partial : return CRCHackLevel::Full``. **(or)** maybe a simple ``if()`` check might be enough.

plugins/GSdx/GS.h
```diff
@@ -1425,3 +1425,25 @@ enum class Filtering : uint8
...
+
+//Enum order matters! Don't change it!
+enum class CRCHackLevel : int8
+{
+	Automatic = -1,
+	None = 0,
+	Minimum = 1,
+	Partial = 2,
+	Full = 3,
+	Aggressive = 4
```
ssakash: FYI, initializing the first enum element is enough and the succeeding elements value will follow arithmetic progression with a difference of 1.

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -271,7 +273,7 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 
 						// Gregory: to avoid a massive slow down for nothing, let's only enable
 						// this code when CRC is below the FULL level
-						if (m_crc_hack_level < 3)
+						if (static_cast<int8>(m_crc_hack_level) < static_cast<int8>(CRCHackLevel::Full))
```
ssakash: static casts aren't necessary here.

plugins/GSdx/GS.h
willkuer: Well, I like to combine enum's with switches. Makes it easy to extend if necessary. However I agree that the if/then/else would be much shorter. Maybe a third opinion switch vs. if would be helpful?

plugins/GSdx/GS.h
willkuer: I did that on purpose to reflect the values that are written in the ini-File. This is important if you insert new elements which would break ini-compatibility. But I am open to remove these manual assignments. Maybe here a third opinion ( @FlatOutPS2 @turtleli )?

plugins/GSdx/GSTextureCache.cpp
willkuer: I didn't know that. I'll update.

plugins/GSdx/GSHwHack.cpp
willkuer: I didn't know that. I'll update.

plugins/GSdx/GS.h
turtleli: Huh... well, ideally I think all the values should be removed, but the ini should use strings instead of numbers to represent crc levels. Of course that'd completely break ini compatibility. You may want opinion from someone other than me. ;)

willkuer: Not sure if you were addressing me or turtleli.

If you insert something (not at the end) you would shift all following values by one. E.g. somebody who selected Full would get Partial instead.

For turtleli's enum parse it would behave like you described and shouldn't be a big issue.

FlatOutPS2: Yes, I was addressing turtleli. Though applying turtleli's suggestion could potentially break compatibility again if the strings were ever to be updated.

turtleli: Well, it's unlikely that strings would be changed. And if they are changed, then the user would end up on the default setting if the string doesn't match. My suggestion is out of scope for this PR though - I'd like a string not only for the CRC hack level, but also some of the other settings too (renderer comes to mind but there's probably others). It isn't truly related to this PR - we can have the enum and/or the suggestion.

So basically I have no opinion on whether to explicitly assign values to each enum or not (I also have no opinion on the switch/if/ternary) but well, I was sort of asked for an opinion and so I provided one ;)

willkuer:
hm.... somehow my fixes don't manage to show up here... interesting.

plugins/GSdx/GSState.cpp
```diff
@@ -23,7 +23,7 @@
 #include "GSState.h"
 #include "GSdx.h"
 
-extern int g_crc_hack_level;
+extern CRCHackLevel g_crc_hack_level;
```
willkuer: this is never used anywhere.... I would remove it?

FlatOutPS2: I removed it in my PR. :p

FlatOutPS2:
> It is also associated with this pr.... maybe github has some problems?

I had similar issues with a PR after a DDOS attack caused server issues, and github was offline earlier today, so I'd say so.

willkuer:
Ok I added the suggested fixes and removed one unused variable.

Only the question with ternary/ifthen/switch remains. I like the switch but again I would follow the majority here. (Performance wise it shouldn't make a difference so the question is what you consider more readable)

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -48,7 +48,9 @@ GSTextureCache::GSTextureCache(GSRenderer* r)
 
 	m_paltex = theApp.GetConfigB("paltex");
 	m_can_convert_depth &= s_IS_OPENGL; // only supported by openGL so far
-	m_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+	m_crc_hack_level = theApp.GetConfigT<CRCHackLevel>("crc_hack_level");
+	if (m_crc_hack_level == CRCHackLevel::Automatic)
+		m_crc_hack_level = AutoSelectCRCHackLevel(theApp.GetConfigT<GSRendererType>("Renderer"));
```
turtleli: This might not work properly on Windows. If the renderer type is `GSRendererType::Default`, then you may end up using an OpenGL renderer but have CRC hacks at the full level due to the renderer autoselection code. The same applies for the other uses of `AutoSelectCRCHackLevel()`.

One solution would be to check if `m_crc_hack_level` is set to `CRCHackLevel::Automatic` in the most derived constructor (GSTextureOGL, GSRendererDX11, etc.) and adjust the level there if it is. There may be other/better solutions.

(Unrelated to PR): I note that s_IS_OPENGL might not work properly on Windows for the same reasons.

plugins/GSdx/GSHwHack.cpp
```diff
@@ -2432,11 +2433,13 @@ void GSState::SetupCrcHack()
 {
 	GetSkipCount lut[CRC::TitleCount];
 
-	s_crc_hack_level = theApp.GetConfigI("crc_hack_level");
+	s_crc_hack_level = theApp.GetConfigT<CRCHackLevel>("crc_hack_level");
```
turtleli: Is `s_crc_hack_level` necessary? Unless I'm missing something, `m_crc_hack_level` could just be used.

plugins/GSdx/GSTextureCache.cpp
willkuer: As the source of the problem is the same for the gsrenderer I would solve both in this pr if that is ok with you. That would solve the s_IS_OPENGL issue as well.

FlatOutPS2: IIRC depth emulation can malfunction if you set the renderer to OGL Software in the plugin settings window, then switch to OGL Hardware with the F9 key. This issue with `s_IS_OPENGL` might be at the source of it.

turtleli: > As the source of the problem is the same for the gsrenderer I would solve both in this pr if that is ok with you. That would solve the s_IS_OPENGL issue as well.

Sure, go for it.

> IIRC depth emulation can malfunction if you set the renderer to OGL Software in the plugin settings window, then switch to OGL Hardware with the F9 key. This issue with s_IS_OPENGL might be at the source of it.

Seems like it.

willkuer:
I put a generic/template autoselection for the case we want to add more automatic enum's. It is used on getconfig.

For the crc hack level it yields a possible duplicate request of the GetConfig("Renderer") and thus might be "slower" than a direct approach. However code looks clean to me and GetConfig shouldn't be used in performance critical code at all.

I still need to check what happens with the static initializations of GSRendererType when Default is used in consumer code. Normally he should always check the ini independent on initialization.

turtleli:
Well, I don't think your approach will work for what FlatOutPS2 described: using GSdx OGL SW, then pressing F9 to switch to GSdx OGL HW. That'll set you up with a full CRC hack level and s_IS_OPENGL will be false, unless I've made a mistake following the code.

willkuer:
You are right. I didn't think about that problem.

Maybe using a lookup table/dictionary like GSdx->m_default_configuration for the current configuration and ReadConfig only on cache miss? Consumer code however would always need to access/set this configuration and thus one would need to remove all member variables with corresponding values in the ini. It would be a huge change to the code...

Somehow I don't like the idea to do something general in most derived classes.

turtleli:
Well, I think that is perhaps overcomplicating things. ;) I think something like this:
```
GSTextureCacheOGL::GSTextureCacheOGL(GSRenderer* r)
	: GSTextureCache(r)
{
    if (m_crc_hack_level == CRCHackLevel::Automatic)
        m_crc_hack_level = CRCHackLevel::Partial;
    m_renderer_type = GSRendererType::OGL_HW;
}
```
a few times is simple enough, but yeah, feel free to think up alternative solutions.

willkuer:
I see that it produces less code lines... the problem I just see is that this kind of implementation produces errors like the one flatout observed. At the moment the code might work but just add another derived class or another toggle/automatic enum field and it might break again.

I know my solution is really huge in comparison to the small problem. As always I am just suggesting and finally somebody else has to decide which implementation is the best. I'll follow your advice.

turtleli:
Yeah, I agree my approach was fragile. I'll do some thinking on the possible solutions.

willkuer:
I provided one ;)

You don't like the cache idea?

I just see that I messed up the order... the autoselection isn't cached :(

turtleli:
Your idea is included in the thinking. ;)

willkuer:
So auto selected value will be cached. It's not perfect performance wise. But it should work for the moment.

Maybe everything will be eventually done differently. Probably not the right time for micro-optimization.

turtleli:
Well, I think perhaps only the current renderer type needs to be stored and used?

Performance isn't really a concern cause GetConfig shouldn't be called in code that needs to be fast anyway. However, unlike Windows, the ini settings are actually already cached on Linux (kinda caused by the fact that we just mimic the GetPrivateProfile*/SetPrivateProfile* Windows functions on non-Windows), so if the current settings cache idea is used you end up with 3 sets of settings - the default settings, the ini settings, and the current settings. I think the only difference between the ini and current settings will be the renderer type if F9 is pressed and also any unread settings, so I feel it's perhaps unnecessary to store every setting.

willkuer:
I saw that linux is already doing that but as you said it is hidden behind the windows alias. I thought I should better not touch it. Possibly windows is doing something similar under the hood.

Using the current code it would be simple to only store GSRendererType or maybe those all who are autoconverted... Do you dislike the amount ram used? Just generally not nice to store the same data twice?

There was once a discussion about the other hotkey settings (like <Ins> for mipmap level). But I don't know whether the final solution was to permanently switch them or only for the current emulation session. One could revisit this idea again. (I think @ssakash was involved in the decision?)

There were also some requests for hotkeys for the internal resolution (switch between native and selected) and such. I am not sure where the hotkeys are heading and if it would be better to store all or just the ones mentioned in this pr.

ssakash:
About hotkeys - Yes, there was a discussion before on doing a permanent toggle but [**Bositman** didn't seem to like it](https://github.com/PCSX2/pcsx2/pull/911#issuecomment-152977990), hence it was never done.

Though from my personal opinion, I still think permanent toggle is better and I don't see how temporary toggle is better in any sort of way.

On the other hand - Don't really like the settings cache idea, I'd personally prefer a more simple solution but feel free to ignore me. :P

turtleli:
> Just generally not nice to store the same data twice?

Yeah, I don't like storing the same data twice.

gregory38:
```
However, unlike Windows, the ini settings are actually already cached on Linux (kinda caused by the fact that we just mimic the GetPrivateProfile*/SetPrivateProfile* Windows functions on non-Windows), 
```
I'm not sure cache is efficient. I didn't bother to make it fast. I wanted a quick solution that was good enough.

willkuer:
So what is the general idea?

I'd say either you transfer the settings from local member variables to global ones and read config only once or you cache the result of get config.

As I am a c#ler I don't like global variables. So I'd prefer some cache solution.

If cache then there are three possibilities:
1. ignore the linux implementation and build the cache on top (as currently done)
2. reuse the linux cache on windows
3. only use a small cache for selected fields which can be toggled via hotkeys as for the renderer

Or just completely remove hot toggling as ssakash suggested.

I think reusing linux cache is the preferred version that I'd use as well. One could also merge the default config map and current config map and use only one map for both windows and linux.

turtleli:
Hmm, I think with hot toggling you'd need to use 2 (reuse linux cache) and 3 (small cache) together? Otherwise the settings in the configure dialog might not actually reflect the ini settings unless you reload the settings from the file.

willkuer:
My idea with 2 was that I store all configs in the linux cache - even on windows - and manipulate this cache for temporary toggles. Then I don't need another one.

3 would mean that I keep the linux cache untouched and thus use the windows native function on windows and the linux cache on linux and add a second cache on top that only saves temporarily changing settings.

turtleli:
> My idea with 2 was that I store all configs in the linux cache - even on windows - and manipulate this cache for temporary toggles. Then I don't need another one.

I'm thinking more about some of the specific details.

Basically - if I boot a game, press F9 to switch renderer and then open the GSdx dialog - the renderer setting stored in the ini should be shown (since we're currently using the temporary toggle system), but the cache will have the temporary setting instead. So I think you'd either need a small temp settings cache or to reload the settings from file in that situation. I suppose this is an implementation detail and can be worked out later.

willkuer:
You will always find a flaw, or?

As alternative to what you suggested one could also allow to bypass the cache on GetConfig by overloading with an additional boolean parameter. GetConfig*(entry, reloadFromFile) and only call this new function from the plugin selector. Not sure which solution is better.

turtleli:
> You will always find a flaw, or?

Well, I'm just being thorough. ;) I'm not sure what would work best either, but I do think having the same ini handling on Windows and Linux would be a good thing.

turtleli:
Well, I think the ini cache + small temp cache approach might be better. With the additional boolean parameter, you'd have to individually override every value that might be temporary, whereas with the temp cache approach, it'd probably be easier to enable/disable/clear all the temporary values at once.

willkuer:
I was on a short trip and away from web. Will implement this either today or in the following days.

If I understood correctly you'd like to generalize the ini cache towards windows as well? Thus, GetConfig of linux remains rather untouched but will be adjusted on win? Would you like to have a static variable as cache or some hashset?

willkuer:
The last question was regarding the short temp cache for gsrenderer and crchacklevel. Rather two static members or some map?

turtleli:
> If I understood correctly you'd like to generalize the ini cache towards windows as well? Thus, GetConfig of linux remains rather untouched but will be adjusted on win?

Difficult question - I guess so. Since the Linux ini stuff was built be to compatible with the pre-existing Windows stuff, it perhaps could be improved if we're removing the Windows stuff. But that's probably a whole load of work.

Personally I wonder whether I could extend and adapt the cdvdGigaherz ini stuff for GSdx, but that's probably a bit much. ;)

> The last question was regarding the short temp cache for gsrenderer and crchacklevel. Rather two static members or some map?

I'm not sure the crchacklevel needs stored - the crchacklevel problem seems to be a side effect of the renderer being wrong or on default, but if we're caching the renderer type that issue disappears. You can choose either approach - The map is more future proof, but I'm not sure there'll ever be any additional values added to it.

plugins/GSdx/GS.cpp
```diff
@@ -506,15 +506,13 @@ EXPORT_C_(int) GSopen2(void** dsp, uint32 flags)
...
 	GSRendererType renderer = s_renderer;
 	// Fresh start up or config file changed
 	if (renderer == GSRendererType::Undefined)
 	{
-		renderer = static_cast<GSRendererType>(theApp.GetConfigI("Renderer"));
-#ifdef _WIN32
-		if (renderer == GSRendererType::Default)
-			renderer = GSUtil::GetBestRenderer();
-#endif
+		renderer = theApp.GetTempConfig<GSRendererType>();
```
turtleli: This statement might be redundant since `_GSOpen()` seems to handle the undefined case anyway. Leaving it in might be better for readability though.

willkuer:
No not finished. I just don't have much time and try to put some stuff whenever I find a free moment. Still thanks for the remarks.

The linux cache produces a non-conform ini file (the ini section is not written). And on each setconfig the whole ini file is written. I'd rather flush the config cache to disk on plugin close and plugin config close and only update the cache on set config but I still need to check the usage of setconfig.

Otherwise I think the code is nicer than the windows equivalent.

willkuer:
Sorry guys. Shit hit the fan. I can not continue working on this in the near future.

It is an important feature that should be added. Somebody else has to do it.

PR #1687, opened 2016-12-06
GSDX: Apply saturation only to interlaced video mode

ssakash:
**Summary of changes**:

* Apply saturation only to interlaced video modes of NTSC, progressive scan variant of NTSC seems to allow usage of higher value.

JMMT uses a bigger display height on NTSC progressive scan mode, which is not really unusual hence the saturation hack has been adjusted to only take effect on interlaced NTSC mode.

However, the whole double screen issue on FMV still exists. As a bit of information, this game has the second output disabled but seems to have some valid data inside of it, maybe the second output data is leaked into the first one? most likely a bug in the frambuffer data management rather than a CRTC issue (needs to be investigated)

### DIFF
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -423,7 +423,7 @@ GSVector4i GSState::GetDisplayRect(int i)
 	//  Limit games to standard NTSC resolutions. games with 512X512 (PAL resolution) on NTSC video mode produces black border on the bottom.
 	//  512 X 448 is the resolution generally used by NTSC, saturating the height value seems to get rid of the black borders.
 	//  Though it's quite a bad hack as it affects binaries which are patched to run on a non-native video mode.
-	if (videomode == GSVideoMode::NTSC && height > 448 && width < 640 && m_NTSC_Saturation)
+	if (m_NTSC_Saturation && isinterlaced() && videomode == GSVideoMode::NTSC && height > 448 && width < 640)
 		height = 448;
 
 	// Set up the display rectangle based on the values obtained from DISPLAY registers


refractionpcsx2:
Id personally put the isInterlaced check as either the first or second option, so it has less checks to do before it bottles out, but then I don't know how time dependent that function is :P

FlatOutPS2:
JMMT is Jonny Moseley Mad Trix. Just in case anyone ever wonders which game this is refers to. :p

ssakash:
> Id personally put the isInterlaced check as either the first or second option (the second probably preferred), so it has less checks to do before it bottles out, but then I don't know how time dependent that function is :P

Hmm, how does it work again in assembly for ``if()`` checks with ``&&``? I assume the following below, too lazy to look at disassembly to micro-optimize a stupid conditional check.

```
CMP   Videomode, NTSC
JNE   _next_statement
CMP   Height, 448
JLE   _next_statement
CMP   Width,640
JGE   _next_statement
CMP   Interlace_mode,1
JNE   _next_statement
CMP   NTSC_Saturation,1
JNE   _next_statement
MOV   Height,448

_next_statement:
// other stuffs
```

So the condition which is most likely to be false should be had first for efficiency on faster jump, so it doesn't make sense to me on why we should have ``isinterlaced()`` check before other checks which are more likely to be false, ``crtc_size`` checks for example as they're more likely to be false as they have a higher range of values which means odds of 1/(2047 - 448) (11 bits for DH) and 1/(4095 - 640) (12 bits for DW) on becoming true, hence atleast 99.4% chances on becoming false whereas ``isinterlaced`` only has two valid choices, returns true (or) false which means it has 50% chances on becoming false, hence the current placement seems rather ideal to me.

Aside from my broken math, I really think this doesn't matter. You're just over analyzing this single line of conditional check :P

turtleli:
IMO `m_NTSC_Saturation` should be first - if the setting is turned off, there's no point in doing the rest of the comparisons in the first place. It also feels more readable that way.

ssakash:
> IMO m_NTSC_Saturation should be first

I didn't place it first because only a few number of people would most likely disable it for debugging and at most of the other scenarios, it would always be enabled by default.

> **if the setting is turned off, there's no point in doing the rest of the comparisons in the first place**

Isn't that the same for all the other checks? For example, If the video mode is not ``NTSC`` then there's no point in doing the other checks?

> It also feels more readable that way.

I agree with code readability, but honestly we're just wasting our precious time over one ultra super negligible conditional check overhead.

turtleli:
> Isn't that the same for all the other checks? For example, If the video mode is not NTSC then there's no point in doing the other checks?

> I agree with code readability, but honestly we're just wasting our precious time over one ultra super negligible conditional check overhead.

It's about the overall meaning - if the user has disabled NTSC saturation, why do checks related to NTSC saturation in the first place?

ssakash:
Fine, I'm changing it only because ref made an inspirational quote which struck right at my conscience. :P

>We're supposed to be working as a team, if we aren't helping and suggesting things to each other, we aren't working as a team - **_Refraction_**

refractionpcsx2:
Doesn't have the same impact without the picture of a sunrise behind or a black and white picture of me next to it :P

willkuer:
We're supposed to be working as a team, if we aren't helping and suggesting things to each other, we aren't working as a team - Refraction

refractionpcsx2:
Thank you @willkuer , that's better xD

danilaml:
addtext_com_mtezotiymjkwnte

refractionpcsx2:
We should start a group called Emu-Meme's lol

gregory38:
@ssakash 
Yes IsInterlaced will be better in 2nd position (because it is easier to understand than other magic condition).

Otherwise (don't take any offense), you need to be more humble and to listen valuable feedback (even if you don't fully agree, just say thanks you). You can learn a lot from others :)

ssakash:
> Otherwise (don't take any offense), you need to be more **humble** and to listen valuable feedback

Yes, I need to learn on how to be [**humble like you**](https://github.com/PCSX2/pcsx2/pull/1625#issuecomment-254924160). ;)

gregory38:
I didn't say that I was an example neither a good target ;)

It was a general comment not limited to this PR. I'm sure you can do better.

FlatOutPS2:
> As a bit of information, this game has the second output disabled but seems to have some valid data inside of it, maybe the second output data is leaked into the first one? most likely a bug in the frambuffer data management rather than a CRTC issue (needs to be investigated)

I just noticed this comment. I've been checking some CRTC related dumps the past few days and a dump of a loading screen black bar issue appears to have the correct height in the disabled output. So there definitely does seem to be something to this.

ssakash:
> I just noticed this comment. I've been checking some CRTC related dumps the past few days and a dump of a loading screen black bar issue appears to have the correct height in the disabled output. So there definitely does seem to be something to this.

Interesting, could you provide any GS dump of such cases? would be nice to look into them.

PR #1688, opened 2016-12-08
gsdx: Use std::thread and std::function for GSJobQueue

turtleli:
Changes:
 * Use std::thread on all platforms - I removed the GSThread class because I feel it's now unnecessary.
 * Makes GSJobQueue non-inheritable - std::function is used instead to specify the worker function - it's probably slightly safer this way (see commit message).


gregory38:
Nice jobs. Did you run some benchmark ?

turtleli:
I didn't notice any performance difference on my system.

willkuer:
what would be a good benchmark? sw mode with many threads?

turtleli:
Well, that's more or less what I did - I ran GS dumps in sw mode with a few threads.

gregory38:
Perfect, it is the same on my PC too.

PR #1690, opened 2016-12-09
Greg/vtune

gregory38:
Improve the support of vtune.

The biggest change is the removal of pre-reservation mechanism for recompiler buffers. It increases the physical memory requirements by ~200MB. But it avoids issues with external tools. Code is also much easier (and portable)


refractionpcsx2:
oh good, maybe now I can use the nvidia tools in the recompiler without it crashing :P

gregory38:
There is still the memory protect feature to handle write in the EE memory segment code (based on segmentation fault too). We need it for performance but maybe it is possible to implement an unsafe option for debug (because I'm pretty sure that most games don't write in the elf code).

Was the crash with openGL ? Persistent buffer requires a lots of memory in their debugger (you can quickly reach the 4GB limit). I need to test the new release maybe it is better now.

refractionpcsx2:
The crash was regardless of renderer

gregory38:
Ok. To be honest I never tested any GPU tool with PCSX2. Only standalone gs dump. It is already a mess with standard debugger.

gregory38:
Any opinion on the extra memory requirements?

refractionpcsx2:
Yeah. PCSX2 uses hardly any memory (relatively speaking), so stop splitting hairs over a few hundred kilobytes or a few meg, just do it if it's better :P

lightningterror:
Indeed. pcsx2 requires higher platforms anyway so pretty much all systems have 4gb ram as a starting point these few years. Heck even ddr2 systems have good amount of ram so don't worry about it.

Do what you gotta do :)

FlatOutPS2:
> Yeah. PCSX2 uses hardly any memory (relatively speaking), so stop splitting hairs over a few hundred kilobytes or a few meg, just do it if it's better :P

It's not a few hundred KB or a few MB's, it's 200MB, which could be significant for some lower end users who are still on 2 or 4 GB's of RAM.

Though a quick test I did shows only a 40MB increase, which isn't that bad.

gregory38:
The extra 180MB was on startup. If you run a game, it will fill the recompiler buffers so extra memory will be smaller. It also depends on the OS memory management. OS is allowed to not allocate physical memory if nothing is written (aka copy-on-write) (and there is the swap too). It is quite hard to quantify the real increase of memory requirement.

I'm glad to know that we are closer of 50MB rather than 200MB.

lightningterror:
Btw curious. Why was this merged if 2 checks didn't pass.

gregory38:
Short answer, because I'm the boss.

Lon g answer, In my case, compilation is done twice (PR + branch), status of 1 is enough. Sometime bot failt to fetch the toolchain package. On linux, I know the compilation status ;)

lightningterror:
https://media.tenor.co/images/3d9c3220a9c0bfab4f6af53229b29a8a/raw

Ah okay then. Thought there might be some issues but if it's ok fair enough.

FlatOutPS2:
Also note that the commits merged after this compiled without issue, so you can tell there's nothings wrong with this one. ;)

lightningterror:
Hmm , I'm not sure if it's the result of this merge but there seems to be a fps increase in GT4 , didn't test others.

I like the results.

PR #1693, opened 2016-12-12
GSDX: Pass total height of framebuffer on GetOutput()

ssakash:
**Summary of changes**:

* Pass the full height of the frame memory in ``GetOutput()`` ignoring the baseline offsets as usual. Fixes flickering issues in ``Crash Bash``.

@FlatOutPS2 - I trust you on checking for regressions. 👍

### DIFF
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -473,6 +473,20 @@ GSVector4i GSState::GetFrameRect(int i)
 	return rectangle;
 }
 
+int GSState::GetFramebufferHeight()
+{
+	const GSVector4i output[2] = { GetFrameRect(0), GetFrameRect(1) };
+	// Framebuffer height is 11 bits max according to GS user manual
+	const int height_limit = (1 << 11);
+	int max_height = std::max(output[0].height(), output[1].height());
+	int frame_memory_height = std::max(max_height, output[0].runion_ordered(output[1]).height() % height_limit);
+
+	if (frame_memory_height > 1024)
+		GL_PERF("Massive framebuffer height detected! (height:%d)", frame_memory_height);
+
+	return frame_memory_height;
+}
+
 bool GSState::IsEnabled(int i)
 {
 	ASSERT(i >= 0 && i < 2);
--- a/plugins/GSdx/GSRendererCL.cpp
+++ b/plugins/GSdx/GSRendererCL.cpp
@@ -169,7 +169,7 @@ GSTexture* GSRendererCL::GetOutput(int i, int& y_offset)
 	const GSRegDISPFB& DISPFB = m_regs->DISP[i].DISPFB;
 
 	int w = DISPFB.FBW * 64;
-	int h = GetFrameRect(i).height();
+	int h = GetFramebufferHeight();
 
 	// TODO: round up bottom
 
--- a/plugins/GSdx/GSRendererCS.cpp
+++ b/plugins/GSdx/GSRendererCS.cpp
@@ -345,7 +345,7 @@ GSTexture* GSRendererCS::GetOutput(int i, int& y_offset)
 	const GSRegDISPFB& DISPFB = m_regs->DISP[i].DISPFB;
 
 	int w = DISPFB.FBW * 64;
-	int h = GetFrameRect(i).height();
+	int h = GetFramebufferHeight();
 
 	// TODO: round up bottom
 
--- a/plugins/GSdx/GSRendererHW.cpp
+++ b/plugins/GSdx/GSRendererHW.cpp
@@ -219,7 +219,7 @@ GSTexture* GSRendererHW::GetOutput(int i, int& y_offset)
 
 	GSTexture* t = NULL;
 
-	if(GSTextureCache::Target* rt = m_tc->LookupTarget(TEX0, m_width, m_height, GetFrameRect(i).height()))
+	if(GSTextureCache::Target* rt = m_tc->LookupTarget(TEX0, m_width, m_height, GetFramebufferHeight()))
 	{
 		t = rt->m_texture;
 
--- a/plugins/GSdx/GSRendererSW.cpp
+++ b/plugins/GSdx/GSRendererSW.cpp
@@ -165,7 +165,7 @@ GSTexture* GSRendererSW::GetOutput(int i, int& y_offset)
 	const GSRegDISPFB& DISPFB = m_regs->DISP[i].DISPFB;
 
 	int w = DISPFB.FBW * 64;
-	int h = GetFrameRect(i).height();
+	int h = GetFramebufferHeight();
 
 	// TODO: round up bottom
 
--- a/plugins/GSdx/GSState.h
+++ b/plugins/GSdx/GSState.h
@@ -242,6 +242,7 @@ public:
 
 	void ResetHandlers();
 
+	int GetFramebufferHeight();
 	GSVector4i GetDisplayRect(int i = -1, bool merged_rect = false);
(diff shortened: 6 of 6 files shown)


plugins/GSdx/GSState.cpp
```diff
@@ -465,6 +465,21 @@ GSVector4i GSState::GetFrameRect(int i)
 	return rectangle;
 }
 
+int GSState::GetFramebufferHeight()
+{
+	const GSVector4i output[2] = { GetFrameRect(0), GetFrameRect(1) };
+	// Framebuffer height is 11 bits max according to GS user manual
+	const int height_limit = (1 << 11) - 1;
```
gregory38: It is true that rendering is limited to a square of 2048x2048. However I'm not sure that reading back the memory is limited by the same limit. I guess the limit depends on the window offset and the maximum pixels that can be read in x/y direction.

plugins/GSdx/GSState.cpp
```diff
@@ -465,6 +465,21 @@ GSVector4i GSState::GetFrameRect(int i)
...
+	int top = std::min(output[0].top, output[1].top);
+	int bottom = std::max(output[0].bottom, output[1].bottom);
+	// According to Gregory, GS seems to wrap when values are passed over limit.
```
gregory38: Sorry, you misunderstood me.

GS memory is a flat array. The address wrap after the 4MB.
```
u8 memory[4MB]

address = compute_address_based(base pointer, buffer width, X, Y) % 4MB
```
So X, Y can range from 0 to UINT_MAX. Address will always remain in the memory.

ssakash:
@gregory38 

> It is true that rendering is limited to a square of 2048x2048. However I'm not sure that reading back the memory is limited by the same limit. I guess the limit depends on the window offset and the maximum pixels that can be read in x/y direction.

Scissoring is max 11 bits so that already gives a good idea on it and FWIW effective value of FBW is between 0-32 too (where no.of pixels = 0x40 * FBW). hence it's highly likely that reading back also has the same limit _I guess_.

> Sorry, you misunderstood me.
GS memory is a flat array. The address wrap after the 4MB.
>```
>u8 memory[4MB]
>
>address = compute_address_based(base pointer, buffer width, X, Y) % 4MB
>```
>So X, Y can range from 0 to UINT_MAX. Address will always remain in the memory.

Actually, the frame memory also uses a similar concept. I'll explain. :)

frame_memory

There are two context of rectangular areas on the framebuffer, the above schematic diagram shows a scenario where a developer set one of the rectangle's DBY to 2047(Nascar) in DISPFB. As you can see, the rectangle's content is then wrapped to the start Y co-ordinate of the Framebuffer when it goes out of bound.

However on ``GSVector4``, we calculate the rectangle's height by subtracting it's bottom value with the top value. In that case, we calculate bottom value by ``top + height``. We all know that top value is 2047, bottom value would be ``2047+x``. When we compare rectangle A with rectangle B to find the merged rectangle's height we'll need to compute it by calculating the lowest top value and highest bottom value of both the rectangles, here's where it gets trickier! rectangle A has a top of 0 and B has a top of 2047, rectangle A has a bottom of 448 and B has a bottom of 2495, so without wrapping the total height will be calculated as 2495 whereas in reality, both of the rectangles actually overlap each other with a total height of 448, I wrap it with respect to the max limit of framebuffer to prevent such issues. Any opinions? Aside from my shabby drawing of the frame memory :P

gregory38:
@ssakash 
Scissoring is only when you use the rasterizer. You can't infer anything from it, except that rendering will be limited to the scissor box. Write/read from the EE/CRTC engine aren't impacted by the scissor.

You need to the check the value of DISP register not FRAME register. But yes, BW likely limit the W around 2048.

Could you post all the info of Nascar case? The base pointer, DBX/Y, format. Hum, it is actually possible that x, y values are coded on 11 bits.

ssakash:
> Could you post all the info of Nascar case? The base pointer, DBX/Y, format. Hum, it is actually possible that x, y values are coded on 11 bits.

Here's a log of NASCAR which I had stored earlier on my PCRTC log archive.
```
PCRTC Log Dump #1
Video Mode : 480P 
Double Scan Flag : Disabled 

 SYNCV Registers: 
 VFP :9 , VFPE : 0  VBP : 30  VBPE : 0 VDP : 1144 VS : 1 

 PMODE Registers:
 ALP: 40 AMOD: 1 , CRTMD: 1 , EN1: 1 , EN2: 1 , MMOD: 1 , SLBG: 0
 
 SMODE1 Registers: 
CLKSEL : 1 , CMOD : 0 , EX: 0 . GCONT: 0 , LC: 32 , NVCK: 1 , PCK2: 0
 PEHS: 0 , PEVS: 0 , PHS: 0 , PRST: 1 , PVS:0 , RC:4 , SINT:0 
SLCK : 0 , SLCK2: 1 , SPML : 2 , T1248: 1 , VCKSEL : 1 , VHP: 1 , XPCK:0

SMODE2 Registers:
 INT: 0 , FFMD : 1 , DPMS : 0 

DISP Output1 : Width: 640  , Height : 448 , Left: 156 , Top: 51 
DISP Output1 Registers: DX:312  , DY:51  , DW:1279  , DH:447 , MAGH: 1 , MAGV: 0
DISP Output2 : Width: 640  , Height : 448 , Left: 156 , Top: 51 
DISP Output2 Registers: DX:312  , DY:51  , DW:1279  , DH:447 , MAGH: 1 , MAGV: 0
FRAME Output1 : PSM: PSMCT32, FBP: 0x8c, Width: 640 , Height : 448 , Left: 0 , Top: 2047 
FRAME Output2 : PSM: PSMCT32, FBP: 0x8c, Width: 640 , Height : 448 , Left: 0 , Top: 0 

Scissor registers:
 SCAX0: 0 , SCAX1:639 , SCAY0:0 , SCAY1:447

Real Size of Image : Width: 640 , Height: 448 
```

plugins/GSdx/GSState.cpp
```diff
@@ -465,6 +465,21 @@ GSVector4i GSState::GetFrameRect(int i)
...
+	int frame_memory_height = (bottom - top) % height_limit;
```
turtleli: Is this supposed to have a range of 0-2047? You have `const int height_limit = (1 << 11) - 1;` or 2047, so the current range is 0-2046.

plugins/GSdx/GSState.cpp
```diff
@@ -465,6 +465,21 @@ GSVector4i GSState::GetFrameRect(int i)
...
+	const GSVector4i output[2] = { GetFrameRect(0), GetFrameRect(1) };
+	// Framebuffer height is 11 bits max according to GS user manual
+	const int height_limit = (1 << 11);
+	int top = std::min(output[0].top, output[1].top);
+	int bottom = std::max(output[0].bottom, output[1].bottom);
+	int max_height = std::max(output[0].height(), output[1].height());
+	int frame_memory_height = std::max(max_height, (bottom - top) % height_limit);
+
+	if (frame_memory_height > 1024)
+		GL_PERF("Massive framebuffer height detected! (height:%d)", frame_memory_height);
```
gregory38: add an empty line after the GL_PERF

gregory38:
@FlatOutPS2 did you test this PR ? What is the status ? (just to be on the safe side before the merge)

FlatOutPS2:
@gregory38 Yes, but I didn't test the latest commit yet. I've only used that on the GSDump from @MrCK1 posted in the SW meta issue to check if it fixed the regression he found(it did), As MrCK1 said this PR fixes the other regressions he noticed it should be fine now.

PR #1694, opened 2016-12-12
GSDX: Cleanup warnings on MSVC

ssakash:
**Summary of changes**:

* Explicitly cast some bitfields/local loop variables to uint8 as certain functions seem to have uint8 as the parameter datatype.

Though to be honest, I don't really care about these warnings. They were just annoying as they keep popping up all the time I rebuild GSdx.

### DIFF
--- a/plugins/GSdx/GSDrawScanlineCodeGenerator.x86.cpp
+++ b/plugins/GSdx/GSDrawScanlineCodeGenerator.x86.cpp
@@ -2889,7 +2889,7 @@ void GSDrawScanlineCodeGenerator::ReadTexel_SSE(int pixels, int mip_offset)
 				movdqa(ptr[&m_local.temp.test], xmm7);
 			}
 
-			for(int j = 0; j < 4; j++)
+			for(uint8 j = 0; j < 4; j++)
 			{
 				mov(ebx, ptr[&lod_i->u32[j]]);
 				mov(ebx, ptr[ebp + ebx * sizeof(void*) + mip_offset]);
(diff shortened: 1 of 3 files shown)


plugins/GSdx/GSDrawScanlineCodeGenerator.x86.cpp
```diff
@@ -2889,7 +2889,7 @@ void GSDrawScanlineCodeGenerator::ReadTexel_SSE(int pixels, int mip_offset)
 				movdqa(ptr[&m_local.temp.test], xmm7);
 			}
 
-			for(int j = 0; j < 4; j++)
+			for(uint8 j = 0; j < 4; j++)
```
CookiePLMonster: You might be bringing a bit of extra overhead by indexing via 8-bit variable, see here:

http://www.viva64.com/en/a/0050/

Do note I have **not** compared assembly outputs for this specific case and it's probably a negligible difference, but depending on how performance critical this code is, it might be a good practice to let go and opt for a natural word size variable instead.

CookiePLMonster: In theory indexing by natural word sized variables should be fastest but yeah, the difference is most likely negligible or subject to a compiler optimization. This loop may even end up being unrolled.

There is also uint8_fast_t if you want to make sure!

danilaml: This loop must be unrolled. It's only 4 iterations. If compiler fails to do so, when it can be unrolled by hand.

CookiePLMonster: I have seen cases where even a `for ( i = 0; i < 1; i++ )` loop wouldn't be removed in Release mode though... so it's probably best to ensure it's the best code possible first.

Or just unroll it manually and get it over with.

gregory38: Guys, as I said, this loop is never (barely) executed. So let's stop the talk on optimizations.

ssakash:
I also made the changes on the x64 version, since @gregory38 is on a vacation for 2 weeks right now, anyone else wants to review it? (_stares at turtleli_)

Issue #1692, opened 2016-12-11
PSX mode: savestate errors


FlatOutPS2:
***PCSX2 version:***
pcsx2-v1.5.0-dev-1705

***PCSX2 options:***
Default / (IOP to interpreter)

***Plugins used:***
Default

***Description of the issue:***
Several PSX games are not identified by just their serial number, but as `cdrom:SLES_....`. This causes issues with loading/saving savestates. An error` "A file could not be created. Path \\cdrom\pcsx2-v1.5.0...."` pops up when saving a savestate. And I did not place pcsx2 on a cdrom. :p

***How to reproduce the issue:***
Boot a game and try creating a savestate for an affected game(check the console title to see if a game serial number is preceded by `cdrom:`).

***Last known version to work:***
v1.5.0-dev-1339-gef9a144
Issue was introduced by the following PR. The save states worked when the games were still identified as `running BIOS`.
https://github.com/PCSX2/pcsx2/commit/5e460c57f86ddd1001b7857340a5100b143aed32

PR #1696, opened 2016-12-13
psxmode: Correct exe name for several PSX titles

FlatOutPS2:
Several PSX titles lack a backslash in the elf path, which made the disc
serial contain 'cdrom:'. This caused savestate issues in those games.

Solves: https://github.com/PCSX2/pcsx2/issues/1692


ramapcsx2:
Thanks, fix looks good :)

PR #1239, opened 2016-03-18
Remove cheats file from redundant folder

FlatOutPS2:
Persona 4 cheats file and the cheats folder a redundant, as the
installer doesn't use this folder/file, but creates a seperate cheats folder
in My Documents/PCSX2 and the Binary release creates a cheats folder the
first time you start up PCSX2.

This can also confuse users of the installer package, as shown here: http://forums.pcsx2.net/thread-54286.html

PR #1699, opened 2016-12-15
NSIS: Update installer scripts, bug fixes, added some new goodies

MrCK1:
Summary of changes:

Main Installer Changes:
- Conducted compatibility testing for installer and build scripts in NSIS 3.01

- Implemented uninstall script for future version detection. If an older or equivalent PCSX2 version is detected, the user is prompted to silently uninstall/overwrite or abort the installer. (The function is not compatible with any current stable releases because they lack the new registry string required for version detection.) 

- We can launch PCSX2 directly from the installer now 👍 

You can try the new installer(s) with version detection [here](https://drive.google.com/open?id=0Bw4EU0WRGcbbZGN2cmhyd2RTMFU)

MakeNSIS: 
- Suppressed errors related to unused plugins
- Replaced some functions 
- Removed an leftover parameter.

Uninstaller: 
- Added section that allows users to keep just memory cards/savestates. 
- Fixed an issue with a registry key left behind by the uninstaller. 
- Fixed an issue with .ini files being deleted unintentionally 
- Removed obsolete function

Misc:
- Updated copyright, added icon to installer
- Reformatted installer instructions and added additional sections, fixed broken VC++ URL.


nsis/SectionCoreReqs.nsh
```diff
@@ -31,11 +31,11 @@
...
 
 	!insertmacro UNINSTALL.LOG_CLOSE_INSTALL
-
-    SetOutPath "$INSTDIR\Cheats"
+	
+	;/bin/Cheats needs CreateDirectory to prevent MakeNSIS from failing due to the folder being empty
     !insertmacro UNINSTALL.LOG_OPEN_INSTALL
-    File                                            ..\bin\Cheats\*
-    !insertmacro UNINSTALL.LOG_CLOSE_INSTALL
+    CreateDirectory                                 "$INSTDIR\Cheats"
```
turtleli: The cheats directory in the install folder isn't used anymore, so there's no need to create it.

nsis/pcsx2_full_install.nsi
```diff
@@ -45,6 +53,69 @@ ShowUninstDetails nevershow
...
+	${If} $R1 S== ${APP_VERSION}
+	MessageBox MB_ICONEXCLAMATION|MB_OKCANCEL "This version of PCSX2 is already installed. Do you want to continue?" IDOK true IDCANCEL false
+	true:
+	DetailPrint "Overwriting current install"
+		Goto Done
+		false:
+		Quit
+		${Else}
+		DetailPrint "Not the same version"
+	${EndIf}
```
turtleli: I don't do NSIS stuff , but it's probably better if it's indented like this:
```
${If} $R1 S== ${APP_VERSION}
	MessageBox MB_ICONEXCLAMATION|MB_OKCANCEL "This version of PCSX2 is already installed. Do you want to continue?" IDOK true IDCANCEL false
	true:
		DetailPrint "Overwriting current install"
		Goto Done
	false:
		Quit
${Else}
	DetailPrint "Not the same version"
${EndIf}
```

From reading the documentation I think you can actually omit `IDOK true` and the true label since you don't actually need to jump in that case.

nsis/pcsx2_full_install.nsi
```diff
@@ -85,45 +156,17 @@ SectionEnd
...
-  ExecWait '"$TEMP\vcredist_2013_x86.exe" /qb'
-  DetailPrint "Finished Visual C++ 2013 Redistributable Setup"
-  
-  Delete "$TEMP\vcredist_2013_x86.exe"
-
-done:
-SectionEnd
 !endif
-
+	SectionEnd
```
turtleli: `SectionEnd` shouldn't be indented here.

ramapcsx2:
Thanks a lot for these improvements!

nsis/SectionCoreReqs.nsh
```diff
@@ -53,20 +53,13 @@
     ; in order to run (including CDVD!) -- and really there should be more but we don't have working
     ; SPU2 null plugins right now.
  
-    File ..\bin\Plugins\GSnull.dll
+    ;File ..\bin\Plugins\GSnull.dll
```
ramapcsx2: Last I checked, null plugins should get shipped.

nsis/SectionCoreReqs.nsh
```diff
@@ -53,20 +53,13 @@
...
     ;File ..\bin\Plugins\SPU2null.dll            
     File ..\bin\Plugins\USBnull.dll
     File ..\bin\Plugins\DEV9null.dll
     File ..\bin\Plugins\FWnull.dll
-    File ..\bin\Plugins\CDVDnull.dll
+    ;File ..\bin\Plugins\CDVDnull.dll
```
ramapcsx2: same

nsis/SectionCoreReqs.nsh
```diff
@@ -18,8 +18,8 @@
   
   SetOutPath "$INSTDIR"
   !insertmacro UNINSTALL.LOG_OPEN_INSTALL
-    File           /oname=${APP_EXE}                ..\bin\pcsx2.exe
-    ;File /nonfatal /oname=pcsx2-dev.exe  					..\bin\pcsx2-dev.exe
+    File			..\bin\pcsx2.exe
```
FlatOutPS2: Probably better to use spaces instead of tabs. Same goes for several of the other changes as well.

nsis/SectionUninstaller.nsh
```diff
@@ -30,24 +30,28 @@ Section "Un.Program and Plugins ${APP_NAME}"
 SectionEnd
 
 ; /o for optional and unticked by default
-Section /o "Un.Configuration files (Program and Plugins)"
+Section /o "Un.Configuration files (PCSX2, GSdx, SPU2-X, ect.)"
```
FlatOutPS2: `Program and Plugins` seems clearer to me.

nsis/SectionCoreReqs.nsh
gregory38: IMHO, GSnull is rather useless. It is better to keep real GS plugin.

nsis/SectionUninstaller.nsh
```diff
@@ -30,24 +30,28 @@ Section "Un.Program and Plugins ${APP_NAME}"
...
+Section /o "Un.Memory cards/savestates"
 
+  SetShellVarContext current  
+  RMDir /r "$DOCUMENTS\PCSX2\memcards"
+  RMDir /r "$DOCUMENTS\PCSX2\sstates"
 SectionEnd
 
 ; /o for optional and unticked by default
-Section /o "Un.User files (Memory Cards, Savestates, etc)"
+Section /o "Un.User files (Cheats, logs, snaps, etc)"
```
FlatOutPS2: `Cheats, Logs, Snapshots` ?

nsis/SectionCoreReqs.nsh
ramapcsx2: Honestly, I can't think of a use either. I'd still ship it though, as clearly someone thought it would be useful..

FlatOutPS2: It might be of some use for debugging, but it doesn't seem like something the end user would need. I don't recall anyone complaining that they were removed from the v1.5.0 builds. Someone who would need it would likely be able to build it from source. It could also be made available in the plugins section of the website for the three people who do think they need it. :)

nsis/pcsx2_full_install.nsi
```diff
@@ -45,6 +53,69 @@ ShowUninstDetails nevershow
...
+	${If} $R0 == ""
+		Goto Done
+	${EndIf}
+	
+	;Check if any other version is installed
+	ReadRegStr $R1 HKLM "${INSTDIR_REG_KEY}" "InstalledVersion"
+	
+	;This check for older versions (pre 1.6.0) without InstalledVersion string will bypass this section
+	${If} $R1 == ""
+	DetailPrint "InstalledVersion string not found, skipping version check"
```
danilaml: Can't it be assumed that if `InstalledVersion` is not in the registry, then the older version is installed? This installer would be used only for the newer builds, wouldn't it?

nsis/SectionCoreReqs.nsh
MrCK1: I only removed the plugins that we got rid of and weren't required for functionality. 

I could also keep the plugins in the installer by simply adding another section (unchecked by default) for null plugins and experimental plugins like the new release of USBqemu?

nsis/SectionCoreReqs.nsh
refractionpcsx2: Depends if you're using a portable install

nsis/SectionCoreReqs.nsh
refractionpcsx2: is it really needed? You can just set "No Disc" in the menu

nsis/SectionCoreReqs.nsh
turtleli: I'm not sure what you're talking about?

PCSX2 creates the cheats folder if it's missing, regardless of mode. The cheats folder in the install directory used to contain a Persona 4 pnach, but that was removed (#1239), so creating a cheats folder in the install directory now has no purpose.

refractionpcsx2: Oh i see what you mean, I thought we were still providing a Final Fantasy X patch

nsis/pcsx2_full_install.nsi
MrCK1: @danilaml I knew I had a good reason for including this check. 

To answer your question, we can't detect the uninstaller executable on older versions is because the uninstaller filename is currently named like so: ```Uninst-pcsx2 1.4.0.exe```

The ```${APP_VERSION}``` string had to be removed to allow for the uninstaller script to function because we can't predict what version the user will have installed. Detecting those older versions wouldn't work unless we defined each registry string individually - which isn't worth the extra installer code.

MrCK1:
I went ahead and updated the Uninstaller labels with FlatOut's suggestions. I also merged the remaining code from SharedRedtape into SectionUninstaller; and the deleted the unused SVN templates

@FlatOutPS2 @ramapcsx2
Do we have any thoughts on what we want to do with the null plugins? If we include them in the installer, it would look something like this: 

installlersample

ramapcsx2:
"Optional Plugins" looks like a good solution to me. As well as having them disabled by default.

FlatOutPS2:
I'm in favour of not including 'em at all. There are always users who click everything and then complain that they have no graphics, no sound or whatever issues Jackun's plugin can cause. The few people who really want to use these plugins can easily find them on the website.

refractionpcsx2:
I'm inclined to agree with @FlatOutPS2 at least as far as the null plugins go.  The useful USB plugins should be included along with any additional pad plugins like pokopom IMO.

FlatOutPS2:
> I'm inclined to agree with @FlatOutPS2 at least as far as the null plugins go. The useful USB plugins should be included along with any additional pad plugins like pokopom IMO.

If you add 'em, you need to support 'em too. And the USB plugins are functional at best. They're not very user friendly and still suffer many issues. And we already have three PAD plugins, no sense in adding anymore.

refractionpcsx2:
> If you add 'em, you need to support 'em too. And the USB plugins are functional at best. They're not very user friendly and still suffer many issues.

True but they do add a lot of functionality people like.

> And we already have three PAD plugins, no sense in adding anymore.

Do we? In my git repo I have xpad and LilyPad

FlatOutPS2:
> True but they do add a lot of functionality people like.

And they can download those plugins if they wish. It's not like those plugins see a lot of maintenance. I posted an issue including a possible solution for the USB mic plugin, but the dev never even responded to any of the posts in the thread @Bositman made.

> Do we? In my git repo I have xpad and LilyPad

And OnePad for Linux. I don't see the use of more PAD plugins. Just gives more room for new issues. We could possibly replace xpad with Pokopom. I'm not sure anyone actually uses xpad regularly. :p

refractionpcsx2:
It depends which of the two is going to be more useful :P

MrCK1:
I guess you guys can figure out which PAD plugins we want to include and whether or not to include Jackun's USB plugin.

For the future, I was also thinking about replacing the FAQ .pdf with a link to a configuration on the wiki. The current Windows guide was written for 1.2.1 and contains obsolete information. The readme could possibly be simplified to a .txt file containing the patch notes.

turtleli:
I don't think xpad should be included - it's not as compatible as LilyPad. If xpad is used and you try to play Xenosaga 2, you get the following message:

```
The DUALSHOCK®2 Analog controller has been removed.

Please insert a DUALSHOCK®2 analog controller into controller port 1 to continue.
```

lightningterror:
Hmm is it a good idea to include generic usb gamepad drivers ?

In "Other" category.(not installed by default) The latest version is 3.70a if I'm not mistaken.

What do you guys think about it ?

MrCK1:
Generic USB drivers are probably best left to be handled by the user's operating system. It's also much easier to bundle DirectX packages with the installer than to host them separately and introduce additional user error to the equation ;)

toehead2001:
It would be a good idea to make it scale properly on HiDPI.

Add this to the pcsx2_full_install.nsi file:
`ManifestDPIAware true`

MrCK1:
I'll add that in when I get a chance ;)

MrCK1:
@turtleli Looking at your previous comment, maybe you should also remove xpad from the VS solution?

turtleli:
> Looking at your previous comment, maybe you should also remove xpad from the VS solution?

I suppose it could be removed from the solution. Or moved to legacy. Dunno.

nsis/SharedBase.nsh
```diff
@@ -38,7 +38,8 @@ SetCompressorDictSize 24
 ; The name of the installer
 Name "${APP_NAME}"
 
-OutFile "output\pcsx2-${APP_VERSION}-${OUTFILE_POSTFIX}.exe"
+; Output the installer to the nsis folder (in git repo)
+OutFile "pcsx2-${APP_VERSION}-${OUTFILE_POSTFIX}.exe"
```
toehead2001: You'll need to update the gitignore to reflect this change.

Why not have NSIS automatically create the output directory if it doesn't exist?

MrCK1: The only way to create a directory at compile time in MakeNSIS (remember we can't use functions) is to run some code through command line; so it's rather pointless in that matter.

Not sure about the gitignore though, maybe it'll make more sense after I get some rest ;)

MrCK1:
I just checked the current gitignore - I see what I forgot to do when I removed the svn templates earlier. I'll remove the directories tomorrow

MrCK1:
Here's a summary of the additional changes in the past few weeks:
- Fixed a regression I introduced in the registry that caused random FTW configuration issues to occur after a reinstall/upgrade.
- Reduced version detection into a single function and added a message reminding users to back up files before a silent uninstall.
- Removed unnecessary description text and parameters.
- Installation is now blocked on XP and support for high DPI screens has been added.

ramapcsx2:
Hey @MrCK1,
is the link to a test installer in the first post still valid? :)

MrCK1:
@ramapcsx2 The link is still valid but it's very outdated at this point (it doesn't include any of the reviewed changes). I'll update the .zip file in 2 hours if you don't mind waiting ;)

ramapcsx2:
Thanks for the update!
The installer works very well and didn't cause any issues with my previous 1.4.0 install. A good sign!
It installs to it's own folder (PCSX2 1.6.0) and ships with the most necessary plugins.

One odd thing: It does ship with the 5 GSdx.dll variations, even though GSdx recently got unified, if I remember correctly.

Installer overall is very nice!

MrCK1:
I actually never updated the link... [here's the new one](https://drive.google.com/open?id=0Bw4EU0WRGcbbTF9yLWtEUU54M0k)

I ended up spending 2 days rewriting the version detection script, and testing. The new script alone is 1/2 of the main installer and will automatically detect previous installations of 0.9.8, 1.0.0, 1.2.1 and 1.4.0 as well as any future builds 😃 - I think I've officially gone overboard with this PR 😈

gregory38:
```
One odd thing: It does ship with the 5 GSdx.dll variations, even though GSdx recently got unified, if I remember correctly.
```
I partially unified the SW renderer but not all the code.

Anyway, is the code ready to be merged ? Future enhancement can still be done in a later PR.

MrCK1:
The PR is ready to be merged. The only thing that isn't here is the CPU check mentioned in #796 but it can be added later if needed. A final review before 1.6 will also be a good idea.

mirh:
Iirc dx redist isn't required from W8.1 onwards.

MrCK1:
Not sure. If it's true I can remove the check box for it and skip installation on newer OSes.

danilaml:
@mirh are you sure about the older dx redist though? I think that I remember that there were some possible issues with some dlls missing from dx9 redist on clean (not upgrade) install of w8.1+.

mirh:
XAudio is an "issue" only on w7- (wouldn't know on 8.0)
@turtleli

gregory38:
Let's just drop Win7.

FlatOutPS2:
Let's just drop Win10 first.

gregory38:
Fine for me. We can drop the support of both. ;)

mirh:
Also x86

lightningterror:
Linux master race ?

turtleli:
I copy from the progress report:

> For a long time it has been necessary for the DirectX redistributables to be installed for PCSX2 to work. However, by moving to a newer Windows SDK which isn't XP compatible and adjusting the code to load libraries that are already present on these operating systems, the need for the DirectX redistributables to be installed on Windows 8.1 and Windows 10 has been eliminated. (Caveat: SCP Xinput Wrapper users that need pressure sensitivity will still require the DirectX redists to be installed since the XInput wrapper wraps around XInput 1.3 and not XInput 1.4)

So the answer is it's unnecessary for some users.

MrCK1:
So what is the DirectX SDK used for?

mirh:
Only if you are <8.1 or you want lillypad_scp. 
Speaking of which, should we mention that somewhere for newbs?

Like: "remember some games requires a ds2/ds3 with special drivers to work properly".

MrCK1:
Hmm. I think I will remove the option to deselect and force install both packages. 

There's too much OS overlap right now. 8.1 is still rather new and many users still run Win 7. I'd rather not create additonal problems with missing packages. Easiet solution is to force installation of said packages until 8.1 becomes the norm in a few more years.

mirh:
Force installing sucks. 
At most make it "default yes". 

And.. Can't you like have a check if redists are already installed?

MrCK1:
Sorry I didn't see your comment earlier. The redists already check for that. The installer skips them if that's the case. 

The installers are enabled by default, but I think users don't listen to instructions well :P

MrCK1:
@mirh when you mentioned "lilypad_scp" what are you referring to? Is it related to the pressure sensitive functions that turtleli mentioned before?

I'm making a change so that installation will not be "forced" per say. Instead, installation of the DirectX redistributables will _not be optional for the user_ but will be skipped/installed based on the user's OS and their response to a prompt following it.

mirh:
> when you mentioned "lilypad_scp" what are you referring to? Is it related to the pressure sensitive functions that turtleli mentioned before?

[Yes](https://github.com/nefarius/ScpToolkit/wiki/Pressure-sensitivity-in-PCSX2). 
But the thing is that at least now, it's not *really* our problem. 
We are not shipping the right dll. 

I don't think we should ask user and all. 
Just check if dx redist is installed, otherwise if (<8.1) default enabled.

MrCK1:
I would agree that the message box was useless in this case. Oh well, your wish has been granted ;)

gregory38:
status ? more work/tests or could it be merged?

MrCK1:
There were a few small fixes I was going to merge. It seems an Intel driver is now crashing VS2015 on my laptop, so I have to fix that first.

MrCK1:
Sorry for the wait. I've been busy recently, but I'll merge some additional work in a few days hopefully.

gregory38:
Any status ?

ramapcsx2:
Even if the directx issue stays unresolved, the installer was already far improved with the old version I tested ~2 months ago. Don't let details hold you up. They can always be addressed in a new PR later :)

MrCK1:
I'm doing a review tonight (now that I'm done with midterms) and looking for issues regarding the new user-level implementation on 7 and Vista specifically. I'm a procrastinator no doubt about it 😛

MrCK1:
So I stopped being lazy and finished my work 😛 There shouldn't be any outstanding issues, and my lazy formatting in some places can always be fixed later. Here's quick summary:

- The installer not runs at user level. If the original account isn't admin, the user can only install PCSX2 in portable mode and vice-versa.

- The DirectX redistributables are now downloaded instead of loaded from the installer. I wanted to reduce the size and avoid downloading extra stuff that isn't needed. In this case, only users running a legacy system that's older than 8.1 will have to download these packages. The installer size is now reduced to ~10MB compressed.

bositman:
Hm I don't like PCSX2 needing internet to get installed though, I thought that was the reason we embedded those the first time

mirh:
Dx web setup would require internet regardless. 
Otherwise the alternative is shipping 90MB of redist in our installer :)

ramapcsx2:
So what will happen if a user runs the installer with no net access? Suppose he already has all the requirements installed. Will the installer give up with the download and finish regardless? That would be preferred.

mirh:
Why should it even care for requirements [and download] if you already have them?

MrCK1:
@ramapcsx2 I'll add a timeout to the URL to account for that. Thanks for the quick feedback

mirh:
Isn't the download after the selection?

danilaml:
@MrCK1 wait. What timeout? Are you downloading the WEB installer to download dx? Web installer should be packaged with pcsx2 installer (it's less than 300KB) and it should handle connection/dx detection automatically.

ramapcsx2:
This could be solved by simply testing the behavior xD

MrCK1:
No, I was talking about adding a timeout with NSISdl, to prevent a possible hang when trying to download the redistributable package.

danilaml:
@ramapcsx2 testing is for the weak!

mirh:
@MrCK1 Uh, that makes additional sense.. But I'm not sure it was what rama had asked.

gregory38:
Can we merge it ?

gregory38:
OMG 34 commits ! It would be nice to squash the commit that can be squashed (such as the revert one)

MrCK1:
Yes, but I don't know how to squash commits  ;)

gregory38:
You need to do an `interactive rebase` => aka `git rebase -i master`. Then you can use `squash` keyword => google it for more details.

MrCK1:
@gregory38 Thanks, will do when I get a chance

MrCK1:
I didn't mean to merge the branch, sorry 😭

MrCK1:
Here's a RC for the final installer if anybody wants a peek, and this time it's small enough for Github 😛
[pcsx2-1.6.0-setup.pdf](https://github.com/PCSX2/pcsx2/files/901328/pcsx2-1.6.0-setup.pdf) (rename to .exe)

lightningterror:
Hmm is this normal on the demo , I get an error at the beginning.
image

MrCK1:
Hmm, well I never got that error during my testing, maybe it's a Win 7 issue. I'll check on my old laptop later.

FlatOutPS2:
I tried the portable setup and it only installs the SSE2 GSdx plugin.

It also marks the SCP toolkit as install by default. That's more an option that should be manually enabled. Last time I tried the SCP toolkit installer was quite intrusive too,

MrCK1:
@FlatOutPS2 I only included the GSdx SSE2  plugin for this test build. I'll make SCP toolkit optional and check out the DX installer later.

turtleli:
Personally I think the SCP tookit shouldn't be part of the installation at all - I don't think it's a good idea to include drivers.

MrCK1:
@FlatOutPS2 Which DX package were you talking about, and what OS did you test on?

FlatOutPS2:
> Which DX package were you talking about, and what OS did you test on?

afbeelding
It shows the DX package and overlapping progress bars. I did the test on Windows 7x64,

MrCK1:
@FlatOutPS2 that behavior was intentional for OSes older than 8.1. We can't check whether or not that package is installed, so there are two options I can do here:

- Force install to be safe but inconvenient for users depending on internet speed.

- Have a message box asking user if they'd like to install the package - like previous stable releases, but the user may be missing the package if they decline.

mirh:
> We can't check whether or not that package is installed

Wait, wat? Why?

Can't you check if XAudio2_7.dll, D3DCompiler_43.dll or something like that are present or not?

MrCK1:
You probably could to some degree, but I don't think that's a reliable clear-cut solution. The Visual C++ Redistributables have a regkey to check if they're installed or not; but the DXSDK has nothing like that.

danilaml:
@MrCK1 so wait a minue. I asked if you were downloading dxwebinstaller [here](https://github.com/danilaml) and you replied no. But now, looking at these screenshots it seems that this is exactly what you are doing! You DON'T need to download dxWEBInstaller. Because it's downloader in itself, i.e. it will check whether dx9 redist is already installed and f not it will ask user if they want to download it.

lightningterror:
I recommend not using the dx web installer as it's got some bloatware.

MrCK1:
@danilaml I'm sorry about that. Not sure how I went 4 months without noticing what's right in front of me😛 I'll address this in a few hours now that it makes more sense.

@lightningterror The Bing Bar included in the installer isn't an issue, running the setup with the ```/Q``` paremeter (silent mode) doesn't install it.

mirh:
> Searching in system folders would probably require elevation as well.

I don't know where you have your mind, but that's what EVERY program EVER do when starting. 

> but the DXSDK has nothing like that unfortunately

If your nuisance is with file access instead of just registry, [here](https://pastebin.com/g1n49rcE)'s a list of entries created after first dxredist run.

MrCK1:
@mirh I needed to push some changes to work on my other PR; but are there any more files besides the two you mentioned (XAudio2_7.dll and D3DCompiler_43.dll) that we need to look for on Win 8.0 and older? I'm asking because (if this is the case) I can skip the DXWebSetup if both are present on the user's system. 😃

mirh:
Well, that would be the point. 
Anyway, those were just two examples. I could have said XInput1_3 too. One alone is good as the others.

gregory38:
Could we stop adding more goodies to this PR ? It would be nice to validate the already huge work, merge it. And then do another PR with more goodies.

Note that I'm also fine if people prefer to do it in one shot.

MrCK1:
OK, there's really nothing else to be added besides the new code for skipping DxWebSetup, but I'll leave this as-is for now. The changes in the last commit should address most of the remaining concerns.

danilaml:
@MrCK1 you don't really need to skip DxWebSetup manually since it will do all the necessary detection by itself (try running it yourself after you already installed dx9).

MrCK1:
Well that's good then, not sure how I got so worried about skipping the package in the first place. 😛

mirh:
> you don't really need to skip DxWebSetup manually since it will do all the necessary detection by itself

Doesn't it still need internet?

danilaml:
@mirh to install dx, yes. It's Web Installer after all. But you'll need internet either way if you don't want to bundle dx9 with pcsx2 installer.

mirh:
> But you'll need internet either way if you don't want to bundle dx9 with pcsx2 installer.

You also don't need internet if we detect in the first place the thing is not needed.

danilaml:
@mirh it want burn your computer and kill your dog if there is no internet. It just won't install anything. So if you have no internet connection the result will be the same, whether you have dx9 installed or not.

mirh:
Assuming we can detect return codes, having no internet from dx setup should abort pcsx2 installation.

danilaml:
@mirh should it? Can't pcsx2 work without dx9 redist?

MrCK1:
Don't forget this is only here for legacy systems. 

@mirh I doubt a noticable speed improvement will be seen by searching the registry for .Dlls versus letting the package abort itself.

mirh:
> should it? Can't pcsx2 work without dx9 redist?

I'm not sure if we crash or just fallback to whatever, but  at least no XAudio and no XInput is a pretty big deal imo. 
Don't know about gsdx requirements then. 

> I don't think that a noticable speed improvement will be seen by searching the registry for .Dlls versus just letting the package abort itself.

It wasn't about speed. It was more about: "no ignorant left behind". 
If this details bothers you, we can even discuss it separately sometime else.

MrCK1:
@gregory38 does this PR still need a review/anything before we merge it?

I'll resolve/add the registry detection to skip dxwebsetup in another PR later. Other than that, the scripts are fully functional as-is.

gregory38:
@MrCK1 lol don't ask me. 

@turtleli @refractionpcsx2 @ramapcsx2 
Can we merge this branch ?

nsis/SectionVersionCheck.nsh
```diff
@@ -0,0 +1,133 @@
...
+    Goto Exec0.9.8
+${Else}
+        Goto +1
+${EndIf}
+
+ExecNormal:
+SetOutPath "$DOCUMENTS"
+CopyFiles /SILENT "$DOCUMENTS\PCSX2" "$DOCUMENTS\PCSX2_backup"
+RMDir /r "$DOCUMENTS\PCSX2"
+    SetOutPath "$TEMP"
```
turtleli: These 4 lines are common in each uninstall case. It'd be nice to avoid duplication.

nsis/SectionVersionCheck.nsh
```diff
@@ -0,0 +1,133 @@
...
+    Goto Exec1.2.1
+${Else}
+        Goto +1
+${EndIf}
+
+${If} $R5 != ""
+    ${AndIf} $5 == "0.9.8"
+    Goto Exec0.9.8
+${Else}
+        Goto +1
```
willkuer: Zapdos?

nsis/pcsx2_shared_init.nsi
```diff
@@ -0,0 +1,224 @@
...
+SetOutPath "$TEMP"
+SendMessage $hwnd ${PBM_SETPOS} 60 0
+${NSD_CreateLabel} 0 45 100% 10u "Installing DXWebSetup package"
+SendMessage $hwnd ${PBM_SETPOS} 80 0
+
+File "dxwebsetup.exe"
+ExecWait '"$TEMP\dxwebsetup.exe" /Q' $DirectXSetupError
+
+${NSD_CreateLabel} 0 45 100% 20u "Finished installing Visual C++ Redistributables. Click Next to continue."
+${NSD_CreateLabel} 0 45 100% 20u "Finished installing Visual C++ Redistributables. Click Next to continue."
```
MrCK1: I should have added a comment here. The last label in the sequence doesn't display in the window due to some sort of bug. The workaround was to duplicate the line.

nsis/SectionVersionCheck.nsh
MrCK1: I can move those lines up somewhere, shouldn't be a big issue.

nsis/SectionVersionCheck.nsh
```diff
@@ -0,0 +1,133 @@
...
+
+Exec0.9.8:
+SetOutPath "$DOCUMENTS"
+CopyFiles /SILENT "$DOCUMENTS\PCSX2" "$DOCUMENTS\PCSX2_backup"
+RMDir /r "$DOCUMENTS\PCSX2"
+    SetOutPath "$TEMP"
+    CopyFiles /SILENT /FILESONLY "$R5\Uninst-pcsx2-r4600.exe" "$TEMP"
+        ExecWait '"$TEMP\Uninst-pcsx2-r4600.exe" /S _?=$R5'
+        Delete "$TEMP\Uninst-pcsx2-r4600.exe"
+        Goto Done
```
MrCK1: Not sure what I was thinking 4 months ago, this code is horrific 😱

nsis/SectionVersionCheck.nsh
```diff
@@ -0,0 +1,133 @@
...
+; $result_var: This will store our result if the "needle" is found. Otherwise it will return null ("")
+; $needle: String to search for
+; $Haystack: String to look in
+
+; Search for 1.0.0
+ReadRegStr $R7 HKLM "${INSTDIR_REG_KEY}" "Uninst-pcsx2Directory"
+${StrContains} "$4" "1.0.0" "$R7"
+StrCmp $4 "" +1
+${If} $R7 == ""
+${EndIf}
```
MrCK1: You're right, I falsely assumed StrCmp needed it to work.

nsis/pcsx2_shared_init.nsi
turtleli: It'd be better to figure out what the bug is.

BTW, a brief look at the documentation suggests that you need to `Pop` the window handles (HWND) off the stack whenever you use `NSD_CreateLabel` (or any `NSD_Create*` function).

nsis/SectionVersionCheck.nsh
```diff
@@ -0,0 +1,95 @@
...
+${If} $R2 == ""
+${AndIf} $R3 == ""
+${AndIf} $R5 == ""
+${AndIf} $R7 == ""
+Return
+${EndIf}
+
+; Installing another version
+MessageBox MB_ICONEXCLAMATION|MB_OKCANCEL "Another version of PCSX2 is already installed. The current configuration folder in Documents will be duplicated and renamed as PCSX2_backup. Click OK to uninstall PCSX2 or Cancel to abort the setup." IDOK SetUninstPath
+
```
turtleli: Err... you removed the `Quit` for some reason.

nsis/pcsx2_shared_init.nsi
MrCK1: @turtleli This doesn't really matter anymore. I forgot to add SendMessage at the end of the function (the same command I used earlier to skip the page on new OSes)

MrCK1:
@turtleli I'm ready for another review :)

nsis/SectionVersionCheck.nsh
```diff
@@ -0,0 +1,98 @@
...
+${AndIf} $R3 == ""
+${AndIf} $R5 == ""
+${AndIf} $R7 == ""
+Return
+${EndIf}
+
+; Installing another version
+MessageBox MB_ICONEXCLAMATION|MB_OKCANCEL "Another version of PCSX2 is already installed. The current configuration folder in Documents will be duplicated and renamed as PCSX2_backup. Click OK to uninstall PCSX2 or Cancel to abort the setup." IDOK SetUninstPath IDCANCEL false
+
+false:
```
turtleli: Not sure you understood me. If the returned MessageBox button ID doesn't match any of the return checks then execution will resume at the next line. So the `IDCANCEL false` part of the MessageBox line and the `false:` label are unnecessary.

Note: This isn't really a big issue. You can leave it if you want. And it's also based on my reading of the NSIS documentation - I haven't actually used NSIS before (not sure if I mentioned it before), so I could be misunderstanding something.

turtleli:
I've reviewed what I can (well, I'm completely ignoring whitespace issues, otherwise it may never end). I doubt I'll be creating the final installer package when it's time to do so, so I'm not the right person to ask to merge this (I also didn't test this at all).

gregory38:
Let's merge the code. It won't create regression anyway and PR is already huge

ramapcsx2:
> @turtleli @refractionpcsx2 @ramapcsx2
Can we merge this branch ?

This PR is so huge, it can't be managed anymore anyway.
Merge it and we'll go from there, making references back. Only if absolutely necessary :p

PR #1702, opened 2016-12-15
SPU2-X: Prevent volume check from altering volume

FlatOutPS2:
Fixes issue in Tokyo Xtreme Racer Zero where checking the left master volume would alter it.

### DIFF
--- a/plugins/spu2-x/src/spu2sys.cpp
+++ b/plugins/spu2-x/src/spu2sys.cpp
@@ -1021,18 +1021,21 @@ static void __fastcall RegWrite_VoiceParams(u16 value)
             thisvoice.ADSR.regADSR2 = value;
             break;
 
+        // REG_VP_ENVX, REG_VP_VOLXL and REG_VP_VOLXR have been confirmed to not be allowed to be written to, so code has been commented out.
+        // Colin McRae Rally 2005 triggers case 5 (ADSR), but it doesn't produce issues enabled or disabled.
+
         case 5:
             // [Air] : Mysterious ADSR set code.  Too bad none of my games ever use it.
             //      (as usual... )
-            thisvoice.ADSR.Value = (value << 16) | value;
-            ConLog("* SPU2: Mysterious ADSR Volume Set to 0x%x\n", value);
+            //thisvoice.ADSR.Value = (value << 16) | value;
+            //ConLog("* SPU2: Mysterious ADSR Volume Set to 0x%x\n", value);
             break;
 
         case 6:
-            thisvoice.Volume.Left.RegSet(value);
+            //thisvoice.Volume.Left.RegSet(value);
             break;
         case 7:
-            thisvoice.Volume.Right.RegSet(value);
+            //thisvoice.Volume.Right.RegSet(value);
             break;
 
             jNO_DEFAULT;
@@ -1075,14 +1078,17 @@ static void __fastcall RegWrite_VoiceAddr(u16 value)
         // for example).  Tests should be done to find games that write to this
         // reg, and see if they're buggy or not. --air
 
+        // NextA has been confirmed to not be allowed to be written to, so code has been commented out.
+        // FlatOut & Soul Reaver 2 trigger these cases, but don't produce issues enabled or disabled.
+
         case 4:
-            thisvoice.NextA = ((value & 0x0F) << 16) | (thisvoice.NextA & 0xFFF8) | 1;
-            thisvoice.SCurrent = 28;
+            //thisvoice.NextA = ((value & 0x0F) << 16) | (thisvoice.NextA & 0xFFF8) | 1;
+            //thisvoice.SCurrent = 28;
             break;
 
         case 5:
-            thisvoice.NextA = (thisvoice.NextA & 0x0F0000) | (value & 0xFFF8) | 1;
-            thisvoice.SCurrent = 28;
+            //thisvoice.NextA = (thisvoice.NextA & 0x0F0000) | (value & 0xFFF8) | 1;
+            //thisvoice.SCurrent = 28;
             break;
     }
 }
@@ -1428,6 +1434,13 @@ static void __fastcall RegWrite_CoreExt(u16 value)
 
         case REG_P_BVOLR:
             thiscore.InpVol.Right = GetVol32(value);
+            break;
+
+        // MVOLX has been confirmed to not be allowed to be written to, so cases have been added as a no-op.
+        // Tokyo Xtreme Racer Zero triggers this code, caused left side volume to be reduced.
+
+        case REG_P_MVOLXL:
+        case REG_P_MVOLXR:
             break;
 
         default: {


sudonim1:
No, wait, what? That's a register write, it SHOULD change the register.  It's not a "volume check", you're turning master volume register writes into a no-op.

sudonim1:
Wait I'm being hasty, this is MVOLX, not MVOL.  That should indeed be a no-op... probably? It's curious that it's written at all.

mirh:
OT, but whatever: would it be possible to lock spu2-x volume with Windows mixer volume?

sudonim1:
@mirh  That is really offtopic and I think you have no idea what you're saying because it makes no sense.

refractionpcsx2:
"SD_VP_ENVX, SD_VP_VOLXL, SD_VP_VOLXR, and SD_P_MVOLX are read only, so they cannot be set"

So this is correct then, have we got similar "NOPs" for the EVNX and VOLX registers?

FlatOutPS2:
> So this is correct then, have we got similar "NOPs" for the ENVX and VOLX registers?

Those aren't defined outside of `regs.h`, so no. :P

refractionpcsx2:
#define VP_ENVX(cr, vc)  (0x400 * cr + 0x00A + (vc << 4))    // voice envelope (current value)
#define VP_VOLXL(cr, vc) (0x400 * cr + 0x00C + (vc << 4))    // voice volume (current value left)
#define VP_VOLXR(cr, vc) (0x400 * cr + 0x00E + (vc << 4))    // voice volume (current value right)

according to the old peops register file :P

sudonim1:
Those go through RegWrite_VoiceParams and RegWrite_VoiceAddr.  Cases 5, 6 and 7 in params are ENVX, VOLXL and VOLXR, and cases 4 and 5 in addr are NAX high and low.  All are treated as writable, and some or all of them MIGHT be, we don't know.

And yes, since this is likely going to come up from looking at this code, the high words are before the low words for addresses.  That's just how it is, the proof is a) software almost all works and b) libsd.irx does that in sceSdSetAddr.

refractionpcsx2:
They are not writable, I checked the sdk manuals

FlatOutPS2:
Air did add notes to those cases stating he didn't know any games that write to it or if it was allowed.

FlatOutPS2:
FlatOut writes to RegWrite_VoiceAddr cases 4 & 5 at the start of a race. Turning it off doesn't seem to cause any issues, but I'll do some more testing to make sure.

refractionpcsx2:
> Air did add notes to those cases stating he didn't know any games that write to it or if it was allowed.

I bet he thought the same for MVOLX too :P There is always a game out there that breaks the rules. it's stupid it does something that purposefully breaks it, maybe it was some of their ways of getting around it working on emulators, or doing that with some other stuff is a sneaky optimisation because they know it will be ignored :)

ramapcsx2:
It could be a bug in the games but we still have to do what the hardware does.

http://problemkaputt.de/psx-spx.htm#spuvoiceflags
There's a note on writing to the ENDX register there, which should be read-only. Apparently it allows setting values, which can be read back for a while, but never affect the actual mixer.
I believe this is the general rule for "read-only" registers. You can write to them but it won't have any actual effect on the behavior.

refractionpcsx2:
Why did you close this PR? There's nothing wrong with it

willkuer:
Probably he is afraid you steal his fame again...

refractionpcsx2:
fame? I have no choice when merging :P

FlatOutPS2:
Your tongue emoticon says it all. You had to manually remove links/mention of me and my PR from the merge commit for this to happen, so keep your BS to yourself. Bye.

refractionpcsx2:
Right.

1.  I just did a standard merge, I didn't touch the commit message.
2.  Github did that, I just left the options for the merge alone because using the "squash" option caused problems before.
3. I'm not trying to do ANYTHING to take your glory, if it's such a big deal I'll put a comment on the commit to say you did it. There done.
4. Sort your f##king shit out dude, you've been a really miserable s##t the last couple of days, you're convinced we're all ganging up against you (or I am) to prove how shit you are and take all your glory, I hadn't even CONSIDERED the idea before you mentioned it, in fact, I had very little to do with any of your commits.  In an effort to make ammends for something I didn't know I done, I tried to push forward some of your PR's you had on here, even merged one as everybody was happy with it. But because github changed it to my name, that's not even good enough for you!  What do you want? Power? Glory? a f#cking medal?

We appreciate every and all PR people put forward and we do our best to get them in a state where everybody is happy with them.  If you don't respond to questions, WE CAN'T COMPLETE IT!  I'm not purposefully ignoring ANYTHING you do, I'm waiting on YOU to make sure everybody is happy with your work before it gets merged and if it's GSDX or something, I don't touch it because I don't know how it works!

Well thanks FlatOut, your extreme childishness has put me in a piss awful mood.

gregory38:
Oh I see now, I think there is a misunderstanding. There is an extra commit for a merge called a merge-commit. The merge commit is ref however the main author remain (but you need to check the older history as date of commit != date of merge)

refractionpcsx2:
Maybe thats what happened then, is that the first (default) option?

gregory38:
Yes it is the normal behavior if your merge (used to be the only option on the web version)

ramapcsx2:
The PR is still important to discuss and eventually fix ;)

mirh:
MFW for the second year in a row unwarranted drama happens the week after christmas.

FlatOutPS2:
> Oh I see now, I think there is a misunderstanding. There is an extra commit for a merge called a merge-commit. The merge commit is ref however the main author remain (but you need to check the older history as date of commit != date of merge)

No, these are normal merge commits. The title always mentions the original creator and links to the PR:

https://github.com/PCSX2/pcsx2/commit/c88cd1b065fdba5bc5a131d45ea1ebbbb60d1f60
https://github.com/PCSX2/pcsx2/commit/29d229264dbf1df85e77c63bff6ac394c5b333f6

Unless you alter it(and as added bonus thereby also hide the original commit(normally added as a no build for reference) from the buildbot):

https://github.com/PCSX2/pcsx2/commit/8fecd3512ccc86bcc4c23f5f8431a01c61b0b34f

gregory38:
I don't know what ref did but somehow a line of log was removed. Tbh I didn't even know it was possible. So yes it misses a GitHub reference to your commit. But you still own it as written in the git history. You can use git blame to convince yourself. Is this issue appear once or is it a regular issue ?

refractionpcsx2:
@FlatOutPS2 your commit is still listed in the list of commits https://github.com/PCSX2/pcsx2/commit/9b6c3bd1061f73335070741b42208eb4ef50771f

sudonim1:
@FlatOutPS2 mate, you don't understand git.  Notice that 8fecd35 says it has 2 parents, c88cd1b which is the last change in master before then and 9b6c3bd, which is your commit merged into master.  8fecd35 is a merge commit which shows that @refractionpcsx2 is the one who performed the merge.  git history isn't a line, it's a tree and nobody is stealing credit.

refractionpcsx2:
Can we reopen this PR? I would really like to see this get merged

refractionpcsx2:
@FlatOutPS2 can we put this one back now please? I would really like to see this merged, it's a good improvement.

PR #1706, opened 2016-12-16
Greg/vif hash

gregory38:
I reimplemented the block management of the VIF recompiler. The goal was to achieve a better speed.

Based on vtune report, it seems to be a success. Overhead goes from 14.5% to 3.8%. Hopefully it will translate to better fps. Benchmark are welcome.

### DIFF
--- a/pcsx2/x86/newVif_Dynarec.cpp
+++ b/pcsx2/x86/newVif_Dynarec.cpp
@@ -326,10 +326,11 @@ _vifT __fi void dVifUnpack(const u8* data, bool isFill) {
 	__aligned16 nVifBlock   block;
 	block.upkType = upkType;
 	block.num     = (u8&)vifRegs.num;
+	block.length  = 0;
 	block.mode    = (u8&)vifRegs.mode;
+	block.aligned = vif.start_aligned;  //MTVU doesn't have a packet size!
 	block.cl      = vifRegs.cycle.cl;
 	block.wl      = vifRegs.cycle.wl;
-	block.aligned = vif.start_aligned;  //MTVU doesn't have a packet size!
 	block.startPtr = 0; // Ease the detection of the end of the hash bucket
 
 	if ((upkType & 0xf) != 9)
--- a/pcsx2/x86/newVif_Dynarec.cpp
+++ b/pcsx2/x86/newVif_Dynarec.cpp
@@ -308,20 +308,26 @@ _vifT __fi void dVifUnpack(const u8* data, bool isFill) {
 	const int	doMask    = isFill? 1 : (vif.cmd & 0x10);
 
 	__aligned16 nVifBlock   block;
-	block.upkType = upkType;
-	block.num     = (u8&)vifRegs.num;
-	block.mode    = (u8&)vifRegs.mode;
-	block.aligned = vif.start_aligned;  //MTVU doesn't have a packet size!
-	block.cl      = vifRegs.cycle.cl;
-	block.wl      = vifRegs.cycle.wl;
 
+	// Performance note: initial code was using u8/u16 field of the struct
+	// directly. However reading back the data (as u32) in HashBucket.find
+	// leads to various memory stalls. So it is way faster to manually build the data
+	// in u32 (aka x86 register).
+	//
+	// Warning the order of data in hash_key/key0/key1 depends on the nVifBlock struct
+	u32 hash_key = (u32)(upkType & 0xFF) << 8 | (vifRegs.num & 0xFF);
+
+	u32 key1 = ((u32)vifRegs.cycle.wl << 24) | ((u32)vifRegs.cycle.cl << 16) | ((u32)(vif.start_aligned & 0xFF) << 8) | ((u32)vifRegs.mode & 0xFF);
 	if ((upkType & 0xf) != 9)
-		block.aligned &= 0x1;
+		key1 &= 0xFFFF01FF;
 
-	//DevCon.Warning("Alignment %d", block.aligned);
 	// Zero out the mask parameter if it's unused -- games leave random junk
 	// values here which cause false recblock cache misses.
-	block.mask	= doMask ? vifRegs.mask : 0;
+	u32 key0 = doMask ? vifRegs.mask : 0;
+
+	block.hash_key = hash_key;
+	block.key0 = key0;
+	block.key1 = key1;
 
 	//DevCon.WriteLn("nVif%d: Recompiled Block!", idx);
 	//DevCon.WriteLn(L"[num=% 3d][upkType=0x%02x][scl=%d][cl=%d][wl=%d][mode=%d][m=%d][mask=%s]",
--- a/pcsx2/x86/newVif_Dynarec.cpp
+++ b/pcsx2/x86/newVif_Dynarec.cpp
@@ -307,7 +307,7 @@ _vifT __fi void dVifUnpack(const u8* data, bool isFill) {
 	const u8	upkType   = (vif.cmd & 0x1f) | (vif.usn << 5);
 	const int	doMask    = isFill? 1 : (vif.cmd & 0x10);
 
-	__aligned16 nVifBlock   block;
+	nVifBlock   block;
 
 	// Performance note: initial code was using u8/u16 field of the struct
 	// directly. However reading back the data (as u32) in HashBucket.find
--- a/pcsx2/x86/newVif_Dynarec.cpp
+++ b/pcsx2/x86/newVif_Dynarec.cpp
@@ -272,16 +272,8 @@ static u16 dVifComputeLength(uint cl, uint wl, u8 num, bool isFill) {
 	return std::min(length, 0xFFFFu);
 }
 
-_vifT __fi void dVifCompile(nVifBlock& block, bool isFill) {
+_vifT __fi nVifBlock* dVifCompile(nVifBlock& block, bool isFill) {
 	nVifStruct& v = nVif[idx];
-	nVifBlock*  b = v.vifBlocks.find(block);
-
-	//  Cache hit
-	if (likely(b != nullptr)) {
-		block.startPtr = b->startPtr;
-		block.length = b->length;
-		return;
-	}
 
 	// Check size before the compilation
 	if (v.recWritePtr > (v.recReserve->GetPtrEnd() - _256kb)) {
@@ -303,7 +295,7 @@ _vifT __fi void dVifCompile(nVifBlock& block, bool isFill) {
 	Perf::vif.map((uptr)v.recWritePtr, xGetPtr() - v.recWritePtr, block.upkType /* FIXME ideally a key*/);
 	v.recWritePtr = xGetPtr();
 
-	return;
+	return &block;
 }
 
 _vifT __fi void dVifUnpack(const u8* data, bool isFill) {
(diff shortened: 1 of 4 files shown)


gregory38:
@refractionpcsx2 @sudonim1 please review my change. I tried to do small commits to ease review.

@FlatOutPS2 please test and benchmark it.

refractionpcsx2:
I'll give it a test later, I know a good few awkward unpack games :)

gregory38:
By the way, I did my measurement with all recommended speedhack + MTVU.

refractionpcsx2:
> By the way, I did my measurement with all recommended speedhack + MTVU.

We shall see :) I'll make up some numbers Vs latest Git

refractionpcsx2:
Here's some benchmarks, looks like some mild improvement :)

tekken tag
vifhash: 82-85fps
master: 79-83fps

Spyro - Heros Tale
vifhash: 166-170fps
master: 163-168fps

MGS3
vifhash: 198-207fps
master: 194-204fps

Enthusia Professional Racing (maybe not the best test :P)
vifhash: 217-224fps
master: 217-223fps

Crash N Burn - Menu
vifhash: 51-53fps
master: 49-52fps

Gran Turismo 4 - Stopping after race start
vifhash: 173-177fps
master: 173-176fps

Grandia III
vifhash: 173-174fps
master: 168-169fps

FlatOutPS2:
I noticed about the same improvements of 1-3 FPS at most in the games I tried, but I didn't have MTVU enabled. I'll give it another shot with that enabled.

refractionpcsx2:
I had MTVU enabled, GSDX set to "native" with basic blending. Usual recommended speedhacks, but everything else default

FlatOutPS2:
> I had MTVU enabled, GSDX set to "native" with basic blending. Usual recommended speedhacks, but everything else default

I used basically the same default settings, the only thing I changed was setting Renderer to OGL HW and CRC hack to Partial.

Tourist Trophy on master without MTVU was 61 FPS on average, and with MTVU 59 FPS (so slower with), but using the PR it's 61 FPS with or without the hack, so eventhough it's not faster, the slowdown seems to be gone.

gregory38:
I'm quite happy with the fps bonus. Code was already optimized by both Cotton & Jake. I only did some micro optimizations.

Here MTVU thread vtune report
screenshot - 12162016 - 10 47 16 pm

Potentially some atomic/locking can be improved. The real meat is the recompiler but the ring buffer overhead isn't small.

refractionpcsx2:
I think we can expect the utilisation from the recompiler to be high, that is a given (we can maybe optimise things within that, but it's a big task, means sifting through recompiled code compared to the original, not a fun job) but yeh, the ring buffers shouldn't be consuming lots of time, the less smaller tasks like that take up, the better.

gregory38:
Yeah, we're both on the same line. Improving the recompiler is a daunting task and it might not even be possible. That why I spent some time on the vif unpack. Which is still high (it used to be slower than the ring bug)

As you can see we have a huge lock time. But potentially it is only a sync between VU/GS and EE (there are warning on debug build). However we spend quite some time on the std::dequeue (push_back). And potentially we do too much (redundant) lock/atomic. IMHO, the dequeue could be replaced with an (no lock/atomic?) ring buffer.

refractionpcsx2:
if you think it might provide better performance without impact on compatibility, then it's certainly worth a go :)

gregory38:
Honestly for the perf, I don't know ;) There is another thing that bug me. There is a boolean `GS_Packet` structure. It increase the size from 16B to 20B. And we do various copy/memset of the structure. This boolean is basically only used as return value of `path.ExecuteGSPacket();` Maybe we can just a reference to a boolean. And potentially, if the struct is aligned properly compiler might use some SSE op. Anyway, a lock free ring buffer will be nice but I will first take some vacations ;)

pcsx2/x86/newVif_Dynarec.cpp
```diff
@@ -261,60 +260,42 @@ void VifUnpackSSE_Dynarec::CompileRoutine() {
...
+	block.startPtr = (uptr)xGetAlignedCallTarget();
+	block.length = dVifComputeLength(block.cl, block.wl, block.num, isFill);
+	v.vifBlocks.add(block);
+
+	VifUnpackSSE_Dynarec(v, block).CompileRoutine();
+
+	Perf::vif.map((uptr)v.recWritePtr, xGetPtr() - v.recWritePtr, block.upkType /* FIXME ideally a key*/);
+	nVif[idx].recWritePtr = xGetPtr();
+
+	return v.vifBlocks.find(block);
```
turtleli: I think returning `&block` is pretty much the same thing?

The `nVif[idx].recWritePtr` slightly above is a bit odd since you've used `v` instead of `nVif[idx]` everywhere else in the function.

pcsx2/x86/newVif_HashBucket.h
```diff
@@ -13,87 +13,121 @@
...
 public:
 	HashBucket() {
-		for (int i = 0; i < hSize; i++) {
-			mBucket[i].Chain	= NULL;
-			mBucket[i].Size		= 0;
-		}
+		m_bucket.fill(nullptr);
 	}
+
 	virtual ~HashBucket() throw() { clear(); }
```
turtleli: This could probably be made non-virtual.

pcsx2/x86/newVif_HashBucket.h
```diff
@@ -13,87 +13,121 @@
...
-
-template< typename T >
-struct SizeChain
-{
-	int Size;
-	T*  Chain;
-};
+#include <array>
+
+// nVifBlock - Ordered for Hashing; the 'num' field and the lower 6 bits of upkType are
```
turtleli: Comment needs fixed since it's no longer just the lower 6 bits of upkType.

pcsx2/x86/newVif_HashBucket.h
```diff
@@ -13,87 +13,121 @@
...
-		SizeChain<T>& bucket( mBucket[d % hSize] );
 
-		if( (bucket.Chain = (T*)pcsx2_aligned_realloc( bucket.Chain, sizeof(T)*(bucket.Size+1), 16, sizeof(T)*bucket.Size)) == NULL ) {
+	void add(const nVifBlock& dataPtr) {
+		u32 b = dataPtr.hash_key;
+
+		u32 size = bucket_size( dataPtr );
+
+		// Warning there is an extra +1 due to the empty cell
+		if( (m_bucket[b] = (nVifBlock*)pcsx2_aligned_realloc( m_bucket[b], sizeof(nVifBlock)*(size+2), 16, sizeof(nVifBlock)*(size+1) )) == NULL ) {
```
turtleli: Since SSE isn't used anymore, maybe aligned realloc/malloc/free don't need to be used? Though it won't hurt to leave it in.

refractionpcsx2:
Good plan :) 
BTW I was looking at your vu check code you committed, in not sure how much that will help as it only checks compilation time, not the time vu blocks take to run. 

Site compilation time is slow, but it happens rarely, so it isn't really the side we need to focus on

refractionpcsx2:
Sure not site, damn I wish git mobile let you edit posts!

pcsx2/x86/newVif_HashBucket.h
```diff
@@ -13,87 +13,121 @@
...
+		}
+
+		return size;
+	}
+
 	void clear() {
-		for (int i = 0; i < hSize; i++) {
-			safe_aligned_free(mBucket[i].Chain);
-			mBucket[i].Size = 0;
+		for (size_t i = 0; i < m_bucket.size(); i++)
```
turtleli: Ranged for? Same below.

gregory38:
You mean my recent commit. It doesn't measure the compilation time. It tells the profiler that data between 0x... and 0x.... are code named VU (I could also have a block view, so you can see if the recompiler speed is mostly dependent on a single block)

refractionpcsx2:
Oh right, I understand now

gregory38:
@refractionpcsx2 by the way, if you look at my previous screenshot. You can see both `VU` and `VIF` entries. As you can see the time spent in VIF recompiler is really smaller. The overhead to find/jump to the block is 2 time bigger (used to be 4x-5x).

On the ring buffer topic, there are various possibilities
* use a lock free mutex instead of std mutex
* use a ring buffer instead of dequeue (to avoid memory allocation stuff). Need to find the best size.
* combine 2 previous idea into a lock free ring buffer (i.e. use atomic for read/write pointer)
Hum, that being said, there is also a semaphore `semaXGkick` that (at least partially) protect the structure too. Potentially a ring buffer without lock neither atomic will be enough.

refractionpcsx2:
Hmm now you mention it, I believe the gs thread was a ring buffer (not sure if it still is?), I have a smelt suspicion that the ring buffer size varied depending on the game and how heavy it was on that unit. 

I could be wrong, just a niggle in the back of my head

pcsx2/x86/newVif_HashBucket.h
gregory38: hum, the annoying part is the realloc (I'm a very lazy guy). Potentially some std::vector could be used too. However I'm afraid for the perf, I need to check the generated code.

And the code is the `find` is wrong. I do the `++` too soon !

pcsx2/x86/newVif_Dynarec.cpp
gregory38: yes interface is a bit awkward. Another possibility will be to use a `void dVifCompile (const nVifBlock& block ...)` interface and move startPtr/length update outside of the compilation

gregory38:
GS.h
```
// Size of the ringbuffer as a power of 2 -- size is a multiple of simd128s.
// (actual size is 1<<m_RingBufferSizeFactor simd vectors [128-bit values])
// A value of 19 is a 8meg ring buffer.  18 would be 4 megs, and 20 would be 16 megs.
// Default was 2mb, but some games with lots of MTGS activity want 8mb to run fast (rama)
static const uint RingBufferSizeFactor = 19;

// size of the ringbuffer in simd128's.
static const uint RingBufferSize = 1<<RingBufferSizeFactor;

// Mask to apply to ring buffer indices to wrap the pointer from end to
// start (the wrapping is what makes it a ringbuffer, yo!)
static const uint RingBufferMask = RingBufferSize - 1;

struct MTGS_BufferedData
{
	u128		m_Ring[RingBufferSize];
	u8			Regs[Ps2MemSize::GSregs];

	MTGS_BufferedData() {}

	u128& operator[]( uint idx )
	{
		pxAssert( idx < RingBufferSize );
		return m_Ring[idx];
	}
};

extern __aligned(32) MTGS_BufferedData RingBuffer;
```

refractionpcsx2:
Okay so 8meg was taken as the sweet spot in that scenario, fair enough :)

gregory38:
However I'm not 100% sure of the connection between all threads. My feeling is that
* EE will send the command to MTVU and will send a WAIT_MTVU command to MTGS.
* MTVU will receive the command, and post the result in dequeue
* MTGS will wait the result in dequeue

MTVU/MTGS interconnection is quite complex.

refractionpcsx2:
I can imagine it isn't fun lol. passing data from one thread back to another thread to pass it to another thread can't be easy.

pcsx2/x86/newVif_HashBucket.h
gregory38: Hum, `std::vector` is maybe not a good solution. The idea of storing only a pointer (innside m_bucket`) was to reduce cache miss probability. A `std::vector` contains 3 pointers on LinuxSTL.

gregory38:
@turtleli I updated the code based on your feedback. And I fix my silly mistake. I will take 2 weeks of vacations. You're in charge of the merge, merry Christmas :stuck_out_tongue_closed_eyes:

turtleli:
> I will take 2 weeks of vacations.

No problem, have fu

> You're in charge of the merge

....N noooooooooooooooooooooooooooooooooooo! XD

Yeah sure, I'll check over the code. Have fun :)

refractionpcsx2:
Oh yeh, forgot to test this more lol.  Oh well, I'm sure we'll see if any issues pop up.  Great job @gregory38  :)

gregory38:
Thanks you. I will look at the ring buffer when I come back. I think the overhead is related to the huge number of fake mtvu packets. Hopefully it will improve mtvu efficiency

refractionpcsx2:
That would be excellent

PR #1709, opened 2016-12-17
cdvd: Street Fighter EX3 (NTSC-J) black screen fix

turtleli:
No, I don't know what I'm doing. This might break everything.

@prafullpcsx2 sent me a block dump and found that Street Fighter EX3 (NTSC-J) was broken by this commit - https://sourceforge.net/p/pcsx2/code/347/.

I found an odd change while searching through the history (bc9e0b08adbe4922433a980ee82166ef03d987c2), where `cdvd.PwOff` was somehow replaced with `cdvd.Status` in the `cdvdRead()` code.

Undoing both those changes lets me get past the black screen.

Related forum issue: http://forums.pcsx2.net/Thread-Street-Fighter-EX3-black-screen-only-on-NTSC-J-version

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -1048,9 +1048,9 @@ u8 cdvdRead(u8 key)
 			CDVD_LOG("cdvdRead07(Break) %x", 0);
 			return 0;
 
-		case 0x08:  // STATUS
-			CDVD_LOG("cdvdRead08(Status) %x", cdvd.Status);
-			return cdvd.Status;
+		case 0x08:  // INTR_STAT
+			CDVD_LOG("cdvdRead08(IntrReason) %x", cdvd.PwOff);
+			return cdvd.PwOff;
 
 		case 0x0A:  // STATUS
 			CDVD_LOG("cdvdRead0A(Status) %x", cdvd.Status);
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -890,7 +890,9 @@ __fi void cdvdReadInterrupt()
 
 	if (--cdvd.nSectors <= 0)
 	{
-		cdvd.PwOff |= 1<<Irq_CommandComplete;
+		// Setting the data ready flag fixes a black screen loading issue in
+		// Street Fighter Ex3 (NTSC-J version).
+		cdvd.PwOff |= (1 << Irq_DataReady) | (1 << Irq_CommandComplete);
 		psxHu32(0x1070)|= 0x4;
 
 		HW_DMA3_CHCR &= ~0x01000000;


prafullpcsx2:
Thanks @turtleli . I can confirm that Street Fighter EX3 (NTSC-J) works fine now, I will check other games for regressions now.

ramapcsx2:
Oh, this could be a rename, based on newer findings (back then). I'd certainly get @refractionpcsx2 to look at it. Maybe he remembers.
In any case, CDVD status flags are touchy. I thought we had some hardware tests for them but dunno now..

refractionpcsx2:
@prafullpcsx2 can you remember that other game we fixed (think it was an F1 game) Which needed the status to be a certain status. Does that one still work?

This will need testing on games we can think of which were very touchy about this to make sure they are still okay.

> Oh, this could be a rename, based on newer findings (back then). I'd certainly get @refractionpcsx2 to look at it. Maybe he remembers.

he barely remembers what happened yesterday :P I think we based a lot of information off the open PS2 SDK, possibly the No$PSX documents aswell, I also had this site in my bookmarks which may have played a role http://lukasz.dk/mirror/cdvdmania/docsps2.html

FlatOutPS2:
> @prafullpcsx2 can you remember that other game we fixed (think it was an F1 game) Which needed the status to be a certain status. Does that one still work?

Assuming you mean Formula One 2001 I can confirm that it still works with this PR.

I've used this PR on a decent number of games and I've seen no regression.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -1222,6 +1222,7 @@ static void cdvdWrite04(u8 rt) { // NCOMMAND
 				Console.WriteLn( Color_Gray, L"CdRead: Reading Sector %07d (%03d Blocks of Size %d) at Speed=%dx",
 					cdvd.SeekToSector, cdvd.nSectors,cdvd.BlockSize,cdvd.Speed);
 
+			cdvd.PwOff = 1 << Irq_DataReady;
```
turtleli: If I use `cdvdSetIrq(1 << Irq_DataReady);` in the N_DVD_READ or N_CD_READ cases it seems to break things.
```
(EE pc:80005560) TLB Miss, addr=0x0 [load]
(EE pc:80005560) TLB Miss, addr=0x0 [load]
(EE pc:80005560) TLB Miss, addr=0x0 [load]
(EE pc:80005560) TLB Miss, addr=0x0 [load]
(EE pc:80005560) TLB Miss, addr=0x0 [load]
(EE pc:80005560) TLB Miss, addr=0x0 [load]
(EE pc:80005560) TLB Miss, addr=0x0 [load]
Too many execution errors.  VM execution has been suspended!
```
I guess the same will apply to the N_CD_READ_CDDA and N_CD_READ_XCDDA case. Any ideas on what to try?

About the N_CD_CHG_SPDL_CTRL case - I'm not sure the change I made even makes sense now that I think about it, but yeah, I do know pretty much nothing about CDVD emulation stuff.

refractionpcsx2: Well if it makes you feel any better, I know very little :P I suspect that command isn't really supposed to be there, but I can't prove it, nor do I know where exactly it should be.  With DVD and CD read, I suspect the data is not classed as "ready" until the data has transferred, but I couldn't guarantee that.

turtleli: Possibly the change could go in `cdvdReadInterrupt()`. The Street Fighter block dump still works, and  I think it makes more sense that way (the data actually is ready if I'm reading the code correctly).
```
diff --git a/pcsx2/CDVD/CDVD.cpp b/pcsx2/CDVD/CDVD.cpp
index 9b0da9a..8ecd9fe 100644
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -890,7 +890,7 @@ __fi void cdvdReadInterrupt()
 
 	if (--cdvd.nSectors <= 0)
 	{
-		cdvd.PwOff |= 1<<Irq_CommandComplete;
+		cdvd.PwOff |= (1 << Irq_CommandComplete) + (1 << Irq_DataReady);
 		psxHu32(0x1070)|= 0x4;
 
 		HW_DMA3_CHCR &= ~0x01000000;
```

turtleli:
Commits updated - it probably makes more sense? Dunno.

pcsx2/CDVD/CDVD.cpp
refractionpcsx2: Yeah that makes much more sense, if it's right however, who knows? :p

refractionpcsx2:
Makes more sense :) Still confused why they called that reg (or why we called it) PwOff, it has nothing to do with turning the power off, unless there's an interrupt there to say "im turning off"

refractionpcsx2:
So, I tried this on the only games I know which are fussy about the interrupt status stuff (the Formula One games) and they all worked fine. Not sure what else I can test with lol

ramapcsx2:
The Time Crysis games (2 and 3) are very picky about flags. If they still work, well, this wouldn't break them! ( :p )
I'll try them now.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -890,7 +890,9 @@ __fi void cdvdReadInterrupt()
 
 	if (--cdvd.nSectors <= 0)
 	{
-		cdvd.PwOff |= 1<<Irq_CommandComplete;
```
ramapcsx2: This line is used twice in this file. Have you checked if both need the change?

refractionpcsx2: it looks like they have both been changed, not sure why we aren't just calling `cdvdSetIrq` though.

There is also another place where we are manually clearing the pwOff, I'm not sure we should be doing that, the BIOS handling for clearing IRQ's should be doing that.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -1048,9 +1048,9 @@ u8 cdvdRead(u8 key)
 			CDVD_LOG("cdvdRead07(Break) %x", 0);
 			return 0;
 
-		case 0x08:  // STATUS
-			CDVD_LOG("cdvdRead08(Status) %x", cdvd.Status);
-			return cdvd.Status;
```
ramapcsx2: You're right. This looks very copypasta buggy with 2 cases both returning Status.

refractionpcsx2: It does, seems the wrong way around tho xD You'd expect 0x0A to be the copy pasta of 0x08 lol

ramapcsx2:
Time Crysis 2 and 3 still work fine, as they used to. Judging by how right this change looks, I'll say merging is okay.

pcsx2/CDVD/CDVD.cpp
ramapcsx2: Both methods probably work the same but cdvdSetIrq also prepares IOP for a branch test, so the flag change is potentially seen earlier then. I don't think it will make it worse..

refractionpcsx2:
Agreed

ramapcsx2:
There is still one occurrence of "cdvd.PwOff |= 1<<Irq_CommandComplete;" that might need changed and one other occurrence of this fix, but commented out!
"cdvdSetIrq( (1<<Irq_CommandComplete) ); //| (1<<Irq_DataReady) );"

@turtleli any ideas on those?
Otherwise we can still merge this known good fix and look at the other questions later.
Not dealing with them now will at least not break anything.

refractionpcsx2:
hmm, maybe. I'm not sure though as that just returns a result, same as most of the SCOMMAND's, but none of them even set an IRQ! (not sure if that's wrong or not)

ramapcsx2:
The flags are being set and surely picked up later by the routine branch checks. All the cdvdSetIrq function does is put in an early branch test so the flag change is noticed immediately.

refractionpcsx2:
Yeah but I don't know if any games are waiting for an IRQ, that's my thought process :P

turtleli:
> There is still one occurrence of "cdvd.PwOff |= 1<<Irq_CommandComplete;" that might need changed

I don't think the data ready flag quite makes sense there since none of the actions involve data transfers.

> "cdvdSetIrq( (1<<Irq_CommandComplete) ); //| (1<<Irq_DataReady) );"

No clue. It depends whether the TOC is treated as data?

> it looks like they have both been changed, not sure why we aren't just calling cdvdSetIrq though.

Just tried that - seems to work(barely tested). Should I add the change to this PR, or would merging first be preferable? I don't mind either way.

ramapcsx2:
If I may vote: Merge this PR as is and leave the new questions to a new PR.

refractionpcsx2:
Well the setIrq function sets up the interrupt (like you were doing) and sets the branch time accordingly to make sure the IOP picks it up as soon as possible and doesn't wait for another DMA irq to trigger it, so it could be a plan to use that in the PR :)

turtleli:
I'll... let you two come to a decision ;)

refractionpcsx2:
Eeny meeny miney mo.. :P

ramapcsx2:
:)
Okay, truth is, I'm very scared of the cdvd flags. There seems to be a timing element to this, on top of total guesswork + reversing what we could. Maybe a good look at current Open PS2 Loader cdvd emulation could help!

refractionpcsx2:
I was thinking more for the consistency throughout cdvd, seems silly to make it different to everything else :P

ramapcsx2:
Sure, we just need to test it. Didn't someone earlier in a discussion in the PR had TLB misses?

refractionpcsx2:
Not that my F3 key can find :P

ramapcsx2:
You'll have to expand one post of yours further up with "show outdated" :p

PR #1710, opened 2016-12-18
cdvdgigaherz: Improve sector prefetch, fix past end of buffer read

turtleli:
First 2 commits are refactoring.

3rd commit should fix the read past the end of the disk prefetch issue. It also improves the prefetch logic a bit (see commit message).

4th commit fixes a read past the end of the buffer.


FlatOutPS2:
The ReadFile failed warnings are gone. 👍

PR #1714, opened 2016-12-22
cdvdgigaherz: Reduce stuttering

turtleli:
This should reduce stuttering.

Bootup might be slightly slower, mainly because cdvdReadSector doesn't send a request to the cache if the data isn't present and it's the first thing called (it doesn't really get called after that). I should perhaps make it do so...

### DIFF
--- a/plugins/cdvdGigaherz/src/ReadThread.cpp
+++ b/plugins/cdvdGigaherz/src/ReadThread.cpp
@@ -29,7 +29,7 @@ struct SectorInfo
 {
     u32 lsn;
     s32 mode;
-    // Sectors are read in blocks, not individually
+    // Sectors are cached in blocks, not individually
     u8 data[2352 * sectors_per_read];
 };
 
@@ -70,9 +70,10 @@ u32 cdvdSectorHash(u32 lsn, s32 mode)
 void cdvdCacheUpdate(u32 lsn, s32 mode, u8 *data)
 {
     std::lock_guard<std::mutex> guard(s_cache_lock);
-    u32 entry = cdvdSectorHash(lsn, mode);
+    const u32 entry = cdvdSectorHash(lsn, mode);
 
-    memcpy(Cache[entry].data, data, 2352 * sectors_per_read);
+    const u32 block_size = mode == CDVD_MODE_2048 ? 2048 : 2352;
+    memcpy(Cache[entry].data, data, block_size * sectors_per_read);
     Cache[entry].lsn = lsn;
     Cache[entry].mode = mode;
 }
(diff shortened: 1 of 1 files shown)


FlatOutPS2:
Yes, bootup is even slower with this PR. It's another 2 seconds slower(adding to the 3 seconds it already slowed down a while ago).

turtleli:
Ok, so I need to see if I can reduce bootup time.

What about stuttering?

FlatOutPS2:
> What about stuttering?

I don't have a game that really suffers from it lying here right now, so I'll need to test that later.

turtleli:
No problem, it might actually be better to wait till I update the PR.

turtleli:
Updated. Bootup time should be a little faster than master.

ramapcsx2:
Best if you always test with a DVD and a CD (PS2) game.
Eject the disc between runs. That should bypass the OS cache.

turtleli:
> Best if you always test with a DVD and a CD (PS2) game

I can't test PS2 CD games because I don't have any (I do have some PSX games though).

Stuttering seems worse on Linux with this PR, so I need to think about this a bit more.

FlatOutPS2:
Overall it feels like stuttering was reduced very slightly on both CD and DVD, but I noticed sometimes a stuttering sequence was faster on the PR on first pass, but on second pass the master seemed slightly faster. :/

Guilty Gear X(one of the games that didn't boot from CD until a while ago).
Master CD boot up: 35+ sec.
PR CD boot up: 40 sec. (confirming boot up is slower)
ISO boot up: barely 1 sec. :P

I also got two readfile errors on boot with both master and PR:

```
 * CDVD Disk Open: CD, 1 tracks (1 to 1):
 * * Track 1: Data (Mode 1) (311128 sectors)

 * CDVD ReadFile failed: sectors 2512-2527: error 23
 * CDVD ReadFile failed: sectors 2512-2527: error 23
```

turtleli:
> Guilty Gear X(one of the games that didn't boot from CD until a while ago).
> Master CD boot up: 35+ sec.
> PR CD boot up: 40 sec. (confirming boot up is slower)
> ISO boot up: barely 1 sec. :P

Those are long boot up times. Are those boot up times specific to that game, some games or all games?

> I also got two readfile errors on boot with both master and PR:

CD reading probably still needs a bit of work. I get a similar error with Suikoden, but I haven't spent time looking at the issue yet.

FlatOutPS2:
> Those are long boot up times. Are those boot up times specific to that game, some games or all games?

Specific to that game. Most other games load reasonably fast(or load slowly on disc and in ISO format).

turtleli:
I'm going to abandon this PR. I'm not sure I can improve it without significantly changing how the threading and caching currently works.

ramapcsx2:
Okay, it appears there isn't a lot of demand for this anyway. It would have been nice but priorities :)

PR #1720, opened 2016-12-24
PCSX2-Counters: Fix RTC counting at certain cases

ssakash:
**Summary of changes**:

* Fix a minor bug in counting time for real time clock. (potential improvements _might_ only be exclusive to progressive mode games)

According to some of refraction's older comments, getting the timer wrong might cause some games to not boot, so look out for progressive mode games which refuse to boot! Also needs extensive testing to make sure that no regressions pop up.


refractionpcsx2:
Metal gear solid 3 was the game that didn't boot until the clock was fixed, there might be others, but u never knew any :p

refractionpcsx2:
Seems fine to me, MGS3 still works (which was the big offender with RTC) and i accidentally left the emulator running in the BIOS for about 10-15 minutes and it didn't even lose a second :)

PR #1727, opened 2016-12-30
Greg/gsdx sem

gregory38:
Implement GSdx ring buffer with ring buffer. Note the wait isn't fully implemented but it is enough for early test/benchmark.


gregory38:
Of course it doesn't compile on Windows. @turtleli could you help me to add the missing include path on VS (pthread on GSdx).

refractionpcsx2:
I just added it (to try and save some time) but new errors have arisen 

7>GSRasterizer.obj : error LNK2001: unresolved external symbol __imp__sem_post
7>GSRasterizer.obj : error LNK2001: unresolved external symbol __imp__sem_destroy
7>GSRendererSW.obj : error LNK2001: unresolved external symbol __imp__sem_init
7>GSRendererSW.obj : error LNK2001: unresolved external symbol __imp__sem_wait

gregory38:
Did you add the pthread link option ?

refractionpcsx2:
I've added everything I can find, looks like it references the lib as an additional dependency, but can't figure out beyond linking it in the command options how the hell to link it in lol.

So I can hack it in :P

gregory38:
Those functions are used in common/Utilities

refractionpcsx2:
Yeah, I can find no logical way to link these in. I tried checking how baseclasses.lib is referenced, but that seems to be done with magic as it isn't referenced in any file. I did have it working, now it won't >.<

refractionpcsx2:
That seems to work on my end, lets hope it does for appveyor :P

refractionpcsx2:
\o/

lightningterror:
Hey Greg , is this one of the performance optimizations you were talking about?

gregory38:
@refractionpcsx2 is it working on your side or do you have a crash ?

@lightningterror no. This one is a test to make code easier, and potentially a bit faster. I had to kill 2 hours yesterday after a day of train. So here the result. I'm not even sure that it will work.

refractionpcsx2:
I only tried it briefly to make sure it worked and it seemed to go fine, I haven't done any benchmarking yet though

refractionpcsx2:
Oh I only tested hardware mode though..

gregory38:
Could you give a shot to the sw mode? It seem to be bad on windows.

refractionpcsx2:
As i just said in my edit, it does lock up but doesn't crash if i boot in sw mode, looks like GSDX gets stuck in a thread loop.

gregory38:
Oh I missed the edit. I get it. It is my flaw. The init of element in the constructor, I often forgive them. And count is random ;)

gregory38:
Actually could you test it on your side. I have some pending update to replace the spin loop by a proper wait.
```
diff --git a/plugins/GSdx/GSThread_CXX11.h b/plugins/GSdx/GSThread_CXX11.h
index c1ba4ae..a9a265e 100644
--- a/plugins/GSdx/GSThread_CXX11.h
+++ b/plugins/GSdx/GSThread_CXX11.h
@@ -69,6 +69,7 @@ public:
    GSJobQueue(std::function<void(T&)> func) :
        m_func(func),
        m_exit(false)
+       m_count(0)
    {
        m_thread = std::thread(&GSJobQueue::ThreadProc, this);
        sem_init(&m_sem_work, false, 0);
```

refractionpcsx2:
Oh so you can reproduce it? Good, because I'd have had to pass it off to turteli otherwise, because I don't have a clue :P

refractionpcsx2:
Well 2 now works, but 3 crashes (didn't try that before).

gregory38:
No I can't reproduce it but I know me enough to spot my usual mistake ;)

Geez, what did I do again, maybe 4 ?

plugins/GSdx/GSdx.vcxproj
```diff
@@ -116,11 +117,14 @@
...
     <ClCompile>
       <PrecompiledHeader>Use</PrecompiledHeader>
+      <AdditionalIncludeDirectories Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">$(SolutionDir)3rdparty\pthreads4w\include;$(VTUNE_AMPLIFIER_XE_2015_DIR)include;$(ProjectDir);$(SolutionDir)unfree;$(SolutionDir)3rdparty;$(SolutionDir)3rdparty\freetype\include;$(SolutionDir)3rdparty\libpng;$(SolutionDir)3rdparty\opencl;$(SolutionDir)3rdparty\zlib;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
     </ClCompile>
     <Link>
       <ModuleDefinitionFile>.\GSdx.def</ModuleDefinitionFile>
       <TargetMachine Condition="'$(Platform)'=='Win32'">MachineX86</TargetMachine>
       <TargetMachine Condition="'$(Platform)'=='x64'">MachineX64</TargetMachine>
+      <AdditionalLibraryDirectories Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">$(SvnRootDir)\deps\$(Platform)\$(Configuration);$(ProjectDir)vtune\x86;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <AdditionalDependencies Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">%(AdditionalDependencies)</AdditionalDependencies>
```
turtleli: All 3 Additional* lines are unnecessary.

plugins/GSdx/GSThread_CXX11.h
```diff
@@ -24,6 +24,87 @@
...
+		}
+	}
+
+public:
+	GSJobQueue(std::function<void(T&)> func) :
+		m_func(func),
+		m_exit(false)
+	{
+		m_thread = std::thread(&GSJobQueue::ThreadProc, this);
+		sem_init(&m_sem_work, false, 0);
```
turtleli: sem_init should take place before the thread is created, otherwise there's a chance you use an uninitialised semaphore.

refractionpcsx2:
4 works, 5 doesn't have you got some weird code dividing by 2? :P

gregory38:
:upside_down_face: Do you have some graphical glitches for 2/4 ?

refractionpcsx2:
btw something that consistently crashes, if you change the number of render threads after you start a game, as soon as you hit okay on the settings dialogue, it says "opening GS plugins" and instantly crashes.

I'll check for glitches

gregory38:
Well I need to check the reset behavior. So far, I try to make it run at least once.

gregory38:
It means that sync doesn't work !

refractionpcsx2:
boo :(

gregory38:
I pushed my latest update (+ rebase). Code is more complex but unlikely to fix the strange windows behavior ! I don't get it. (I need to check the boss comment)

plugins/GSdx/GSThread_CXX11.h
gregory38: why ?

gregory38:
@refractionpcsx2 could you repopulate and retry

plugins/GSdx/GSThread_CXX11.h
turtleli: Well, you're calling sem_wait immediately in ThreadProc. It'll fix the crashing.

refractionpcsx2:
slow as all hell, but it's stable (since the sem_init before thread commit)

plugins/GSdx/GSThread_CXX11.h
gregory38: Ah yes, good catch.

gregory38:
By slow, you mean slower than master. Or slow as SW renderer. I need to profile it on linux on full PCSX2. Quick test show that gs dump are in the same performance range.

plugins/GSdx/GSdx.vcxproj
refractionpcsx2: Yeah well it was a hack job :P easy enough to remove once we're happy with it.

gregory38:
Hum is 3rdparty/pthreads4w the latest version ?

refractionpcsx2:
No it's second to latest, by coincidence I was just looking at that :P

refractionpcsx2:
Just checked the differences, it's so minor it's not worth bothering with.

refractionpcsx2:
> By slow, you mean slower than master. Or slow as SW renderer. I need to profile it on linux on full PCSX2. Quick test show that gs dump are in the same performance range.

really? Hopefully then this one you're in for a treat, because I get about 4fps with 3 extra threads (master does ~50fps)

http://www.filedropper.com/mgs3swsemi

gregory38:
Master 3ET
```
Performance Profile for 189 frames:
Min  26.89 ms	(37.19 fps)
Mean 29.99 ms	(33.34 fps)
Max  48.76 ms	(20.51 fps)
SD   3.67 ms
```

Branch 3ET
```
Performance Profile for 189 frames:
Min  27.55 ms	(36.30 fps)
Mean 34.77 ms	(28.76 fps)
Max  58.92 ms	(16.97 fps)
SD   5.95 ms
```

Branch with a spin lock Wait (instead of a blocking mutex)
```
Performance Profile for 189 frames:
Min  26.15 ms	(38.23 fps)
Mean 28.01 ms	(35.70 fps)
Max  46.94 ms	(21.31 fps)
SD   2.98 ms
```

gregory38:
Note the spin lock is quite heavy on the mtgs thread. But replayer doesn't contain EE/VU thread. Besides my CPU is 8 cores.

refractionpcsx2:
I wish I got 16fps on that scene :P

gregory38:
Performance is really strange. Linux semaphore got various optimization (and even 64 bit opt). They map directly to futex. However the windows implementation seem to use pthread mutex. Well condition variable are based on pthread semaphore. Potentially there is still a bug. Did you restart the renderer or is it in the first start ? Do you need N cpu half running or 1 core is heavy used ?

gregory38:
Hum based on this paper (figure 30.12), my cond var implementation isn't too bad.  
http://pages.cs.wisc.edu/~remzi/OSTEP/threads-cv.pdf

I was afraid that a single mutex can't be used for 2 cond vars. It seems to be legit. Looking at the CV implementation, there is one access to m_count that isn't properly protected by the mutex. Maybe it is the Skylake issue.

gregory38:
Hum maybe I understand the perf difference between the 2 implementations.

In ThreadProc:
* CV: It will empty the full queue for a CV wakeup
* Semaphore: it will call wait for all elements in the queue. 

Linux semaphore implementation is optimized to a couple of atomic operations when wait will only decrement the semaphore counter without blocking the thread. I'm not sure the windows behavior is equivalent.

gregory38:
Some doc to implement interesting stuff http://preshing.com/20150316/semaphores-are-surprisingly-versatile/

refractionpcsx2:
> Did you restart the renderer or is it in the first start ? 

from first start

>Do you need N cpu half running or 1 core is heavy used ?

No idea, I didn't check task manager to be honest

gregory38:
Hum in the Wait function, could you change the `#if 1` to `#if 0`

refractionpcsx2:
with yield: 23fps
with unique lock: 24fps

barely a difference :P

avih:
> with yield: 23fps
> with unique lock: 24fps
> 
> barely a difference :P

I wouldn't call it barely a difference. First, they're too close to actually tell, but assuming you measured accurately and consistently, and one is 23.0 fps and the other is 24.0 fps, it's almost 5% difference. Quite meaningful in general.

refractionpcsx2:
> I wouldn't call it barely a difference. First, they're too close to actually tell, but assuming you measured accurately and consistently, and one is 23.0 fps and the other is 24.0 fps, it's almost 5% difference. Quite meaningful in general.

well yes, but compared to 39fps on master with the saem settings, it doesn't help much, that was my point.

FlatOutPS2:
I get 18 FPS in the MGS intro scene(parachute and cockpit scene) with 3 ET with the latest commit, which is also double what I got with yesterday's PR version(9 FPS), but I only get 30 FPS on master...

gregory38:
I likely did my previous bench at 3GHZ (turbo off).

For me there is 2 potential speed issue with Semaphore
* I call sem_trywait for all jobs. Condition variable stuff is called only once for N jobs. Semaphore is lighter but called more often. Speed will depends on the load of the work queue.
* Pthread semaphore implementation suck

On linux the generic implementation is
```
spin_lock (based on atomic => few ASM instructions)
value--
spin_unlock
```
The fast implementation (no lock at all)
```
compare_and_swap value with value -1
```
I don't know which one I have. But I'm around 10-15 asm instructions (from top of my head) to execute the trywait

Then you have the w32 pthread implementation. The idea is close of linux generic implementation but with full mutex. So you have a function call, 3/4 branch to check the mutex validity. If you're unlucky it will block, so context switch and so on... 
Now, I better understand why the core try to optimize the call to avoid semaphore functions.

So I don't know what I will do. Either forget about it. Or move the atomic operation outside of the semaphore (see above example).
* https://github.com/preshing/cpp11-on-multicore/blob/master/common/autoresetevent.h

refractionpcsx2:
Well maybe the best option for now is to ifdef the whole block for Linux only, so we can revisit it for windows if need be, but Linux users can reap the benefits

gregory38:
I'm not sure the code is better on Linux. I would need to benchmark it correctly with a recent pthread version (mine is 2 years old). And also check the 64 bits behavior (as recent code use 64 bit atomic to improve the performance). It could worth to check the behavior of the AutoReset event object but it is a low priority

Initial goal was to make the code easier and I hoped to avoid the strange perf issue on Skylake CPU.

refractionpcsx2:
> Initial goal was to make the code easier and I hoped to avoid the strange perf issue on Skylake CPU.

I thought that was the only real goal lol. It would be nice, but if it's not possible, then i guess we will have to look at alternatives.  as for the pthreads version, the windows one we have is only 0.0.1 versions behind and the addition is literally very minor, I think it specifies a new class on a function, that's it.

turtleli:
The performance of master and this branch seems to be about the same on ArchLinux.

Side note: The SW renderer seems so much faster on Linux for me - On master, using the MGS3 dump and 3ET, I get ~30fps on Windows and ~44fps on Linux.

gregory38:
@turtleli what is your CPU ? Initial GSdx code was based on std::queue + pthread/win stuff. It was slow on Linux. My guess was shared pointer requires too much atomic operation. I replaced the code with boost/C+11. It was faster on Linux but similar speed on Windows. So I conclude that old code was only slow on Linux. Most atomic operation are done without any barrier. You might need to check the generated ASM. Otherwise maybe it could help to separate m_count in its own cache line.

turtleli:
It's an i5-4590 (Haswell refresh). One thing I forgot to test was the 0ET case in case it's simply better generated code (I compiled with clang 3.9.1, should probably do the GCC test too)

> Otherwise maybe it could help to separate m_count in its own cache line.

Actually I tried removing m_count altogether - performance didn't change. The best improvement I managed on my own branches used custom mutexes based on the Windows SRW lock stuff (VS2015 does this anyway for std::mutex, but it's slower because the C++ runtimes has to cater for Windows XP and Vista) - on the MGS3 dump I get a 0.5fps increase.

gregory38:
On my Haswell-E at 3Ghz, I'm already around 30 fps (old GCC 4.9, yes Debian users). 44 is a bit high if you didn't overclock your CPU, but you can expect around 40 fps. I don't think GCC/Clang code really matter (ofc, granted that you enable minimal optimization). The slow path is inside the JIT code anyway.

I don't know how you implemented the SRW. I wanted to do something to avoid the multiple wait/spurious wake up of MTGS. The idea was to create a busy lock RW. ThreadProc will get a read-lock so you can run all queues. However the wait would take a write lock. It means that write lock will be unlocked only when all queues are done.

turtleli:
Hmm. I just checked - with 0ET, I get ~17fps on Windows and ~20.5fps on Linux.

> I don't know how you implemented the SRW

I used Windows specific stuff:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa904937(v=vs.85).aspx
https://msdn.microsoft.com/en-us/library/windows/desktop/ms682052(v=vs.85).aspx

I just made drop in replacements for the mutex/etc stuff (well, I implemented the bare minimum). I was just testing the std mutex speed.

gregory38:
Hopefully someone will post those results on reddit. And then all users will switch to Linux ;)

The difference is huge. I can't explain it. Could be
* better STL container (unlikely)
* memcpy/memset that uses AVX instead of SSE
* Linux locks your CPU to a higher turbo frequency (or at least longer)

turtleli:
> Hopefully someone will post those results on reddit. And then all users will switch to Linux ;)

About that - FreeBSD 11.0 does 48fps with 3ET ;)

Yeah, I've no idea why Windows is noticeably slower. I don't think it's the turbo though - both Windows and Linux claim I'm slightly above 3.6x Ghz on a 0ET thread run.

gregory38:
30 FPS => VSync ?

turtleli:
I don't think it's vsync.

PR #1728, opened 2016-12-31
GSDX-TextureCache: Fix corner cases on custom resolution scaling

ssakash:
**Summary of changes**:

* Scale the textures based on the total size of both the output circuits combined, fixes a regression on ``Time Crisis 2``.
* Remove an older hack as it's no longer necessary.

Potentially there are still some bugs in the code as I'm not too confident with it for some reason,

### DIFF
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -405,7 +405,7 @@ GSVideoMode GSState::GetVideoMode()
 	return videomode;
 }
 
-GSVector4i GSState::GetDisplayRect(int i)
+GSVector4i GSState::GetDisplayRect(int i, bool merged_rect)
 {
 	if(i < 0) i = IsEnabled(1) ? 1 : 0;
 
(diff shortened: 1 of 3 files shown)


refractionpcsx2:
That was bad timing, you pushed this when gregory had broken the master so it didn't compile :p

ssakash:
I'll rebase it later when I get free time, however the code still has some bugs. I probably shouldn't calculate the union of rect when it's intersecting too much. That's not ideal and might potentially cause some scaling errors due to some small pixel offsets on DISP by few games.

plugins/GSdx/GSState.cpp
```diff
@@ -433,8 +433,16 @@ GSVector4i GSState::GetDisplayRect(int i)
 	rectangle.right = rectangle.left + width;
 	rectangle.bottom = rectangle.top + height;
 
-	// Useful for debugging games:
-	//printf("DW: %d , DH: %d , left: %d , right: %d , top: %d , down: %d , MAGH: %d , MAGV: %d\n", m_regs->DISP[i].DISPLAY.DW, m_regs->DISP[i].DISPLAY.DH, r.left, r.right, r.top, r.bottom , m_regs->DISP[i].DISPLAY.MAGH,m_regs->DISP[i].DISPLAY.MAGV);
```
ssakash: Before someone comments "Oh, it's rather rude to remove comments of other developers without any reason", I was the one who added this comment and I think it's useless, so removing it.

ssakash:
I've updated the code to only pass the merged output circuit dimensions when the intersect and union offsets are similar **(or)** when either the width/height are not intersecting, seems much better now as we don't need to worry about some games which do funny things with the alternating display offsets.

plugins/GSdx/GSState.cpp
FlatOutPS2: Uncommenting it only isn't enough to make it work anyway. I tried to use it ages ago for a quick test, but then the `r.` had already been replaced by `rectangle` and it has down instead of bottom which was a little confusing. :p

refractionpcsx2: I think he means the comments explaining code.  Not that anybody has ever said not to do that, to my knowledge

FlatOutPS2:
I've tested this on several GS dumps and games. I haven't noticed any regressions so far.

gregory38:
I'm good with it. It can be merged

PR #1729, opened 2017-01-01
Extend LTO support and remove warnings

np511:
Adds option to change number of cores (so default of four can be changed) to CMake and build.sh and removes LTO warnings.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -87,6 +87,7 @@ for ARG in "$@"; do
         -D*                 ) flags="$flags $ARG" ;;
 
         *)
+        echo $ARG
             # Unknown option
             echo "** User options **"
             echo "--dev / --devel : Build PCSX2 as a Development build."
--- a/cmake/BuildParameters.cmake
+++ b/cmake/BuildParameters.cmake
@@ -295,7 +295,7 @@ endif()
 #-------------------------------------------------------------------------------
 # Set some default compiler flags
 #-------------------------------------------------------------------------------
-option(USE_LTO "Enable LTO optimization (will likely break the build)")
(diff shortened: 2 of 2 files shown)


gregory38:
in the build script there already a variable that contains the number of core. I don't see the point to add another option. And potentially Cmake got an handy variable too. Did you check it ?

np511:
> in the build script there already a variable that contains the number of

> core. I don't see the point to add another option. And potentially Cmake

> got an handy variable too. Did you check it ?

I just took a look in the build script - let me clarify: the number doesn't need to be the number of cores on the system - usually, setting them equal makes compilation faster, but you can set the number lower or higher if you want.

gregory38:
Hum. Let's be honest, the best setting is likely the number of core. IMHO, it doesn't worth to add another option. Nobody will check LTO scaling on multiple core system ;)

np511:
> Hum. Let's be honest, the best setting is likely the number of core. IMHO,
> it doesn't worth to add another option. Nobody will check LTO scaling on
> multiple core system ;)
*Sighs* Alright, I'll change it later.

np511:
Oops. I've been updating PCSX2 every day (sometimes more). When I pushed this commit, I changed the origin remote url in the config file - and then "git pull" started updating from my fork.

I'll fix this later, I now have two separate folders cloned - one of here, one of the fork.

How do I make my changes part of the first commit (instead of creating a second one?)

gregory38:
`git commit --amend` and then force push. Otherwise, I will use the github option to do it.

np511:
Fixed this

PR #1735, opened 2017-01-03
GSdx: Prevent FMV crash

FlatOutPS2:
Fixes FMV crashing PCSX2 in The Simpsons: Road Rage.


gregory38:
Geez, I always forget about this case. Thanks.

PR #1741, opened 2017-01-04
GSdx: Improve GSJobQueue behaviour

### LINKED ISSUE #998: [Needs further testing on Skylake] GSdx Software mode is slower(or has issues with multithreading) on Skylake CPUs after switch to VS2015
Because of the rarity of users with Skylake CPUs, for the first report of this we told the guy to wait for another user with Skylake to confirm it.

Now and independent unrelated confirmation seems to have happened.

Relevant threads:
http://forums.pcsx2.net/Thread-Is-anyone-using-a-Skylake-cpu -this one is basically a dupe of:
http://forums.pcsx2.net/Thread-Software-mode-runs-a-lot-slower-after-1-3-1-1415

and now:

http://forums.pcsx2.net/Thread-Bug-Report-Final-Fantasy-Final-Fantasy-XII-International-Zodiac-Job-System

The first two are by Dr_Hycodan and he was the one told to wait for confirmation, the latter is that confirmation.

The basic gist of it is that in software mode, Skylake CPU performance dropped massively after we switched to Visual Studio 2015 compilation (pcsx2-v1.3.1-1416-gb6f78d8-windows-x86 and forward). In the original thread Dr_Hycodan thought that PCSX2 was just ignoring "Extra rendering threads" but I had him monitor his CPU usage with different numbers of threads and it does change as expected so it's not that. 

In the later thread, we can see that the GS% jumped massively after the switch, from 27% to over 80%. The why is beyond me.

The only relevant thing seems to be skylake CPU as users with non skylake(like me) haven't been able to reproduce it thus far. It's also still possible it's something REALLY esoteric and the fact that only Skylake users have reproduced it so far is a coincidence, thus why I believe we need more Skylake testing.

Summary:
Affected versions: pcsx2-v1.3.1-1416-gb6f78d8-windows-x86 and forward
Hardware: Intel Skylake CPUs
Games: Apparently most/all

@gregory38 You said in Dr_Hycodan's thread it would be interesting to have another example from Skylake, here you go!

turtleli:
Changes:
 * Don't use a separate counter to count queue items. Seems to fix #998.
 * Use a separate mutex for waiting - should reduce context switches.
 * Change the exit flag within a lock and make it non-atomic.

### DIFF
--- a/plugins/GSdx/GSThread_CXX11.h
+++ b/plugins/GSdx/GSThread_CXX11.h
@@ -44,10 +44,9 @@ private:
 		while (true) {
 
 			while (m_count == 0) {
-				if (m_exit.load(memory_order_relaxed)) {
-					m_exit = false;
+				if (m_exit.load(memory_order_relaxed))
 					return;
-				}
+
 				m_notempty.wait(l);
 			}
 
@@ -87,10 +86,11 @@ public:
 
 	~GSJobQueue()
 	{
-		m_exit = true;
-		do {
-			m_notempty.notify_one();
-		} while (m_exit);
+		{
+			std::lock_guard<std::mutex> l(m_lock);
+			m_exit = true;
+		}
+		m_notempty.notify_one();
 
 		m_thread.join();
 	}
--- a/plugins/GSdx/GSThread_CXX11.h
+++ b/plugins/GSdx/GSThread_CXX11.h
@@ -34,6 +34,7 @@ private:
 	ringbuffer_base<T, CAPACITY> m_queue;
 
 	std::mutex m_lock;
+	std::mutex m_wait_lock;
 	std::condition_variable m_empty;
 	std::condition_variable m_notempty;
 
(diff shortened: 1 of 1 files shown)


turtleli:
> empty() is potentially inaccurate because of the relaxed atomic.

I've taken that into account:
 * the worker thread will not report not empty when it's not - the read index is updated on this thread.
 * the worker thread may report empty when it's not - the m_lock mutex locking deals with that.
 * the main thread will not report empty when it's not - the write index is updated on this thread.
 * the main thread may report not empty when it's not - the m_wait_lock mutex locking deals with that.

So it should be thread-safe (or so I think - can never be too sure).

wareya:
Works for me, and general performance the same as your previous test builds.

gregory38:
As a side note, potentially relaxing the fetch_add/sub atomic could have been enough. It might worth to do a review of fetch in MTGS/MTVU code path.

PR #1742, opened 2017-01-05
Greg/mtvu mtgs lockless

gregory38:
Remove one mutex on the MTVU/MTGS channel. Test are welcome.

### DIFF
--- a/pcsx2/Gif_Unit.h
+++ b/pcsx2/Gif_Unit.h
@@ -209,7 +209,7 @@ struct Gif_Path {
 	}
 
 	bool isMTVU() const           { return !idx && THREAD_VU1; }
-	s32 getReadAmount()           { return readAmount.load() + gsPack.readAmount; }
+	s32 getReadAmount()           { return readAmount.load(std::memory_order_acquire) + gsPack.readAmount; }
 	bool hasDataRemaining() const { return curOffset < curSize; }
 	bool isDone() const           { return isMTVU() ? !mtvu.fakePackets : (!hasDataRemaining() && (state == GIF_PATH_IDLE || state == GIF_PATH_WAIT)); }
 
@@ -389,7 +389,9 @@ struct Gif_Path {
 
 	// MTVU: Gets called after VU1 execution on MTVU thread
 	void FinishGSPacketMTVU() {
-		readAmount.fetch_add(gsPack.size + gsPack.readAmount);
+		// Performance note: fetch_add atomic operation might create some stall for atomic
+		// operation in gsPack.push
+		readAmount.fetch_add(gsPack.size + gsPack.readAmount, std::memory_order_acq_rel);
 		while (!mtvu.gsPackQueue.push(gsPack))
 			;
 
--- a/pcsx2/Gif_Unit.h
+++ b/pcsx2/Gif_Unit.h
@@ -17,6 +17,11 @@
 #include <deque>
 #include "System/SysThreads.h"
 #include "Gif.h"
+#include "GS.h"
+
+// FIXME common path ?
+#include "../plugins/GSdx/boost_spsc_queue.hpp"
+
 struct GS_Packet;
 extern void Gif_MTGS_Wait(bool isMTVU);
 extern void Gif_FinishIRQ();
@@ -147,10 +152,13 @@ static __fi void incTag(u32& offset, u32& size, u32 incAmount) {
 struct Gif_Path_MTVU {
 	u32   fakePackets; // Fake packets pending to be sent to MTGS
 	GS_Packet fakePacket;
-	Mutex gsPackMutex; // Used for atomic access to gsPackQueue
-	std::deque<GS_Packet> gsPackQueue; // VU1 programs' XGkick(s)
+	// Set a size based on MTGS but keep a factor 2 to avoid too waste to much
+	// memory overhead. Note the struct is instantied 3 times (for each gif
+	// path)
+	ringbuffer_base<GS_Packet, RingBufferSize / 2> gsPackQueue;
 	Gif_Path_MTVU() { Reset(); }
-	void Reset()    { fakePackets = 0; gsPackQueue.clear();
+	void Reset()    { fakePackets = 0;
+		gsPackQueue.reset();
 		fakePacket.Reset();
 		fakePacket.done =  1; // Fake packets don't get processed by pcsx2
 		fakePacket.size =~0u; // Used to indicate that its a fake packet
@@ -380,22 +388,21 @@ struct Gif_Path {
 
 	// MTVU: Gets called after VU1 execution on MTVU thread
 	void FinishGSPacketMTVU() {
-		if (1) {
-			ScopedLock lock(mtvu.gsPackMutex);
-			readAmount.fetch_add(gsPack.size + gsPack.readAmount);
-			mtvu.gsPackQueue.push_back(gsPack);
-		}
+		readAmount.fetch_add(gsPack.size + gsPack.readAmount);
+		while (!mtvu.gsPackQueue.push(gsPack))
+			;
+
 		gsPack.Reset();
 		gsPack.offset = curOffset;
 	}
 
 	// MTVU: Gets called by MTGS thread
 	GS_Packet GetGSPacketMTVU() {
-		ScopedLock lock(mtvu.gsPackMutex);
-		if (mtvu.gsPackQueue.size()) {
-			GS_Packet t = mtvu.gsPackQueue[0];
-			return t; // XGkick GS packet(s)
+		// FIXME is the error path useful ?
+		if (!mtvu.gsPackQueue.empty()) {
+			return mtvu.gsPackQueue.front();
 		}
+
 		Console.Error("MTVU: Expected gsPackQueue to have elements!");
 		pxAssert(0);
 		return GS_Packet(); // gsPack.size will be 0
@@ -403,18 +410,13 @@ struct Gif_Path {
 
 	// MTVU: Gets called by MTGS thread
 	void PopGSPacketMTVU() {
-		ScopedLock lock(mtvu.gsPackMutex);
-		if (mtvu.gsPackQueue.size()) {
-			mtvu.gsPackQueue.pop_front();
-		}
+		mtvu.gsPackQueue.pop();
 	}
 
 	// MTVU: Returns the amount of pending
 	// GS Packets that MTGS hasn't yet processed
 	u32 GetPendingGSPackets() {
-		ScopedLock lock(mtvu.gsPackMutex);
-		u32 t = mtvu.gsPackQueue.size();
-		return t;
+		return mtvu.gsPackQueue.size();
 	}
 };
 
--- a/pcsx2/Gif_Unit.h
+++ b/pcsx2/Gif_Unit.h
@@ -123,11 +123,13 @@ struct Gif_Tag {
 };
 
 struct GS_Packet {
+	// PERF note: this struct is copied various time in hot path. Don't add
+	// new field
+
 	u32  offset;     // Path buffer offset for start of packet
 	u32  size;	     // Full size of GS-Packet
 	s32  cycles;     // EE Cycles taken to process this GS packet
 	s32  readAmount; // Dummy read-amount data needed for proper buffer calculations
-	bool done;	     // 0 = Incomplete, 1 = Complete
 	GS_Packet()  { Reset(); }
 	void Reset() { memzero(*this); }
 };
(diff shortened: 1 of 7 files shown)


gregory38:
@refractionpcsx2 what do you think about it ? By the way, there is a define in Git_Unit.h to only test the mutex/lockless change

refractionpcsx2:
I'll give it a try, will see what MGS3 is like as that does a lot of MTGS wait's so it's likely to show up any issues.

if it's stable and faster, then win :)

refractionpcsx2:
Sounds far too complicated for me lol.  No idea why I have a wait, all I can assume is the GS plugin isn't processing things fast enough for some reason.

gregory38:
I'm curious if we could increase the size of the git path buffer. So far, they are around 8MB.
```
	Gif_Unit() : gsSIGNAL(), gsFINISH(), stat(gifRegs.stat), lastTranType(GIF_TRANS_INVALID) {
		gifPath[0].Init(GIF_PATH_1, _1mb*9, _1mb  + _1kb);
		gifPath[1].Init(GIF_PATH_2, _1mb*9, _1mb  + _1kb);
		gifPath[2].Init(GIF_PATH_3, _1mb*9, _1mb  + _1kb);
}
```
Maybe one path is quickly filled.

I won't be surprised that it is linked to the channel effect (in codec for example). It requires too send lots of vertex data to the GS. That being said, if we have only a couple of wait by frame, it might not kill the perf.

refractionpcsx2:
I could try that, I did try increasing the size of the MTGS buffer, that did nothing

gregory38:
Hum, you're right there is also a dependency with the MTGS buffer. As far as I understand, there is at least a command send to it. However, the (VU?) data are sent to the gif path

refractionpcsx2:
3 ring buffers? are you sure? that seems awfully wasteful

gregory38:
I'm not sure. But I suspect there is a MTVU/MTGS ring buffer by gif path. I used the same size as the MTGS ring buffer. But I guess it could be 4 times smaller.

pcsx2/Gif_Unit.h
```diff
@@ -258,9 +278,10 @@ struct Gif_Path {
 
 	// If completed a GS packet (with EOP) then returned GS_Packet.done = 1
 	// MTVU: This function only should be called called on EE thread
-	GS_Packet ExecuteGSPacket() {
+	GS_Packet ExecuteGSPacket(bool &done) {
```
ssakash: Actually, it might be nice to have the parameter as a pointer instead. That way we could easily understand that the variable's original value might be modified when we take a look at the function call. ``path.ExecuteGSPacket(&done)`` **instead of** ``path.ExecuteGSPacket(done)``. Any opinion?

plugins/GSdx/boost_spsc_queue.hpp
```diff
@@ -86,10 +93,10 @@ class ringbuffer_base
         size_t ret = arg + 1;
 #if 0
         while (unlikely(ret >= max_size))
+            ret -= max_size;
 #else
-        while (ret >= max_size)
+        ret = (ret % max_size);
```
ssakash: Could be ``ret %= max_size;`` instead.

plugins/GSdx/boost_spsc_queue.hpp
```diff
@@ -169,6 +195,17 @@ class ringbuffer_base
...
     }
 
+    bool size() const
+    {
+        const size_t write_index =  write_index_.load(memory_order_relaxed);
+        const size_t read_index = read_index_.load(memory_order_relaxed);
+        if (read_index > write_index) {
+            return (write_index + max_size) - read_index;
+        } else {
+            return write_index - read_index;
```
ssakash: I wonder if the ``else`` is necessary over here.

pcsx2/Gif_Unit.h
```diff
@@ -380,41 +400,63 @@ struct Gif_Path {
 
 	// MTVU: Gets called after VU1 execution on MTVU thread
 	void FinishGSPacketMTVU() {
+#ifdef LOCKLESS
+		// FIXME order ?
+		readAmount.fetch_add(gsPack.size + gsPack.readAmount);
+		while (!mtvu.gsPackQueue.push(gsPack))
+			;
+#else
 		if (1) {
```
ssakash: Might be nice to remove this useless ``if()`` check.

plugins/GSdx/boost_spsc_queue.hpp
refractionpcsx2: Looks necessary to me.

gregory38: Well better spend time to redo the MTVU atomic management. My initial port to std::atomic was completely crazy ! And architecture of the code isn't good for atomic.

Code ought to be
```
T1
wp = load write position (memory relaxed)
do write 1 on wp
do write 2 on wp + x
commit write position (memory acquire); // act as a memory store barrier
```
So far we reload the atomic every write with random memory constrains (and same for the read part)

refractionpcsx2:
Seems to work fine on MGS3 :) SW mode no longer causes MTGS waits as a nice side effect.

You will need to bump the savestate version though, older states now crash

gregory38:
@refractionpcsx2 did you change the gif path size ? Do you know which change require a savestate bump ?

refractionpcsx2:
I didn't touch anything, I just used the artefact from this PR :)

I'll have a look through the code and see if i can find it, all I know is i tried to load a couple of savestates and it caused PCSX2 to crash, my other recent master executable doesn't, unless it's a change since i last checked out.... I'll do a fetch and rebase to see if that's it.

refractionpcsx2:
Hmm just thought, is it worth testing this in games broken with mtvu incase it was a threading issue?

gregory38:
I don't know. I pushed this morning some changes to improve the correctness of atomic. I barely tested it, and I didn't benchmarked it. I need to review my code but I think it is more correct. However if MT was not safe, it would break all games. I've the feeling that games are broken due to the way MTVU is done.

gregory38:
By the way, you're right the save state issue is due to done. The current packet `GS_Packet gsPack` in Gif_Path is saved.

refractionpcsx2:
Thought it might be.

The symptoms vary, some games don't work, others seem to just execute completely different VU code lol, it's very strange.

refractionpcsx2:
Thinking about it, I wonder if the issue there is the unpacked stuff from VIF isn't getting to the VU across the thread before the VU runs the VU program, that would possibly have that effect. Maybe a wait for MTVU before starting a VU program would fix that, unless we can stop it executing MTVU commands out of order (which i don't know if it's doing)

gregory38:
http://wiki.pcsx2.net/index.php/PCSX2_Documentation/Threading_VU1
Could just be the limitation of the MTVU. It seems VIF is duplicated and VU/VIF aren't sync with the EE thread on purpose.

Yes I know for the VS2013 error. I used `alignas` instead of `__aligned`. I was in a rush to push something so it can be tested.

refractionpcsx2:
hmm. In that case, I wonder if we can pause the VU until VIF catches up, it will get rid of the speed boost, but it will be more compatible, we could always have it as a gamefix then put it in to the gameDB for the games we know are broken.

gregory38:
Or a gamefix to disable MTVU when broken ;) Honestly I don't know. IMHO, we need to debug one game understand why MTVU broke it. And what could be done or not.

refractionpcsx2:
Well yes that is an option ;) But we have that meta for MTVU bugs, it wold be nice so it worked on as many games as possible, even if compromises for those games have to be made as it will still bring a performance boost. 

It's a shame we can't have cases where VU0 can be threaded, stupid COP2...

gregory38:
The EE recompiler could be better. 64 bits will bring us better sign extension of 32 bits operation, better pure 64 bits operation, better TLB and potentially allow better register allocation. I'm pretty confident that we can reduce the EE thread overhead (maybe 5%-10% if we are lucky). It means VU/GS thread will be more often the limited factor (except for the few game that aren't happy with MTVU). It would be complex to extract more threading jobs that could improve further the speed.

By the way, is this PR faster/slower/same than master ?

refractionpcsx2:
from my initial tests, it was very much the same, maybe slightly faster with the omission of the mtgs wait problem, I've not tried since your recent changes.

pcsx2/Gif_Unit.h
gregory38: Pointer is error prone. You need to check if it is null, otherwise someone might use it badly on the future.

plugins/GSdx/boost_spsc_queue.hpp
```diff
@@ -169,6 +195,17 @@ class ringbuffer_base
         return write_index_.is_lock_free() && read_index_.is_lock_free();
     }
 
+    bool size() const
```
turtleli: size_t?

gregory38: OMG !

refractionpcsx2: lol

gregory38:
@refractionpcsx2 , @turtleli 
I'm not sure that WaitVU is thread safe. What is the expected behavior ? Flusing the entire VU ring buffer.
* It will be safer to forceKick (even if I think it is working)
   * Note: I have the feeling that part `&& GetReadPos() != m_ato_write_pos.load(std::memory_order_relaxed)` is mostly always true.
* Anyway, even if post the semaphore, we have no guarantee that VU thread will run right away. So `ScopedLock lock(mtxBusy);` will be useless

If the goal is to purge the ring as I understand it. We need to retry until the ring is actually empty
```
while (GetReadPos() != GetWritePos()) {
   std::this_thread::yield();
   ScopedLock lock(mtxBusy);
}
```

Any opinion ?

refractionpcsx2:
My opinion is you must the for loop and the IsDone. 

:p

refractionpcsx2:
Missed * goddamnit,  I wish you could edit on mobile

gregory38:
Yeah you missed, as you said, the edit ;)

refractionpcsx2:
I didn't miss it, I was just being cheeky :p

refractionpcsx2:
Just tested MGS3 again, this branch is about 2-3 fps slower than master with MTVU off, about the same speed with it on.

gregory38:
HW or SW renderer ?

refractionpcsx2:
Hardware, I don't think SW is much different.

gregory38:
Could you find which commit introduce the speed regression ? Thanks you

refractionpcsx2:
well it's something in this PR, i'm terrible at doing blame stuff without manually reverting xD

gregory38:
You can git checkout sha1_version and compile. No need to blame or revert

refractionpcsx2:
Oh, I didn't know you could do that haha

refractionpcsx2:
Worked out how to do it with tortoisegit :P Will have a play

refractionpcsx2:
Okay went through all the commits, the only one which seemed to negatively affect the speed was the last one 201493981e52bce7380266440d6c4c22d853b638 
Although it's negligible tbh, maybe it just had a bad run, it does seem to vary depending on how it feels :P

gregory38:
I'm surprised the mtvu ring buffer is used without mtvu. Anyway, as we saw with Gsdx Skylake issue, atomic can generate some bad behavior. For example the Wait VU could spin à long time before the thread is really started.

gregory38:
I pushed some small change. Unfortunately I don't think I can do better. I'm pretty sure, we still have a couple of CPU stalls that are bad for the perf. But we have also tons of register write that must transit from the EE thread (gif/vif) to VU thread / GS thread (my testcase is sotc start).

So far the 2 potential stalls are on `FinishGSPacketMTVU` and `VU_Thread::ReserveSpace`

refractionpcsx2:
I'll test it and make sure it works and we will call it there for now then :) Good job!

gregory38:
@refractionpcsx2  Any status ? I need to remove some commented code and do a final review of the code.

refractionpcsx2:
sorry, totally forgot, ill have a quick play now

Edit: Seems to be fine :) You can merge.

gregory38:
There is a deadlock issue in SotC intro... It seem to be trigger because of the `
std::this_thread::yield()` in `WaitOnSize`. If I put back the mutex wait (ie `ScopedLock lock(mtxBusy)`) it is fine. 

I don't understand it. Potentially the code is broken but not enough called to trigger an error. (as the lock will flush the full ring buffer).

gregory38:
Hum, I think we miss some transactions. VU buffer is empty when we have the deadlock. I try to save more space on WaitOnSize (+10000) and it is fine.

Issue #1743, opened 2017-01-06
Ecco the Dolphin: Defender of the Future [PAL-E] - hangs on first loading screen


trostboot:
***PCSX2 version:***
1.5.0-dev-1764

***PCSX2 options:***
defaults, see workaround down below

***Plugins used:***
GSdx11 SW, SPU2-X

***Description of the issue:***
Ecco the Dolphin [PAL-E] [SCES-50499] hangs on the very first loading screen after booting. 

***How to reproduce the issue:***
Boot the game with default settings

***Last known version to work:***
Never, as far as I can tell.

***PC specifications:***
i5-4690k, R9 270X, 16GB RAM, Win10 Pro x64 14393

***Workaround:***
Setting EE cyclerate to -3 while also disabling Wait Loop Detection will allow the game to load properly. Once the game has loaded, these settings can be changed again. The game seems to run fine.

Making this report mostly so it is on Github and can be properly tagged and tackled someday. It's a known issue after all. The workaround maybe less so. Maybe it can help track down this particular gremlin

refractionpcsx2:
That's handy, previously we've used a hack in the code to get around the boot problem :)

refractionpcsx2:
Hmm, that workaround doesn't work here, what about other settings?

refractionpcsx2:
Oh wait! It does after Turteli's CDVD changes! awesome! :)

trostboot:
That's odd. I've tried a number of speedhack combinations and those two were the only ones that mattered. Rest should be defaults. I'll double-check when I get back from walking the dog.

refractionpcsx2:
No ignore me, it does work, and only those settings too, nice job! :)

trostboot:
Sweet. Nice to have it narrowed down some more with the CDVD changes, too.

refractionpcsx2:
I think some more are in order :)

ramapcsx2:
Hm yea, it works intermittently with the workaround. I needed to retry a few times.
Anyway, fiddling with timing always made this game work. The previous 2 "fixes" where a different SIF delay and an IOP underclock, back when we used to have that.
It would be nice though if the flag change affected it. Is it related?

refractionpcsx2:
Well my master didn't work when i posted before, because I hadn't updated to the new CDVD changes, now it does :)

trostboot:
[...]

PR #1744, opened 2017-01-06
cdvd: Use cdvdSetIrq() instead of directly setting cdvd.pwOff

turtleli:
Follow-up of #1709, where there was a suggestion to use cdvdSetIrq() instead.


ramapcsx2:
These flags and their timing might be related to Ecco the Dolphin booting. Info in https://github.com/PCSX2/pcsx2/issues/1743

refractionpcsx2:
Ecco works as well as it did in the previous PR commit :)

FlatOutPS2:
http://forums.pcsx2.net/Thread-Is-there-any-workaround-for-Klonoa-2-crashing-door?page=3

The previous PR(#1709) is mentioned as causing @prafullpcsx2 patches for Klonoa 2 to stop working and causing PCSX2 to crash.

turtleli:
> The previous PR(#1709) is mentioned as causing prafullpcsx2 patches for Klonoa 2 to stop working and causing PCSX2 to crash.

That thread now says it works again. So maybe something else is responsible.

atomic83GitHub:
This PR make Ecco to boot without workaround?

turtleli:
No.

FlatOutPS2:
@turtleli Does this still need testing?

refractionpcsx2:
I think there was a problem with some games doing it this way, I can't remember, it might have been the original change

ramapcsx2:
@refractionpcsx2 do you know whether these commands go through the interrupt controller on real hardware?
If they do, then this commit is more correct and should be used.

refractionpcsx2:
I couldn't tell you to be honest, it's not something I've checked. When I suggested it, it seemed logical

refractionpcsx2:
Can't edit on a phone, damnit lol

I forgot to say it was more for the timing of it which made it make sense, it's probably unlikely the interrupt controls it on real hardware

turtleli:
> I think there was a problem with some games doing it this way, I can't remember, it might have been the original change

I think it was just a problem in the original PR before I found a better place to make the pwOff changes.

> turtleli Does this still need testing?

Dunno. I only briefly tested some of my games. It didn't seem to break/fix anything.

FlatOutPS2:
So, should this still get merged for possible timing benefits or left open just in case someone who runs into a cdvd bug can easily test this?

ramapcsx2:
Well, it's part of a bigger thing with the timing @refractionpcsx2 mentioned.
We have one vote on "it's probably not right" though, so I'd say we leave it open and mark for testing.

arcum42:
Did anyone ever test this?

GovanifY:
As this is a potential hot change and aims to make the codebase one line of code nicer I'll close it for now. We can always look into it later during a formatting change which will have in depth testing.

Issue #1719, opened 2016-12-24
Wallace & Grommit - Curse of the Wererabbit black box on videos


refractionpcsx2:
Strange issue with the videos in this game, there appears to be a black box in the bottom right corner of all the videos.

After some testing it seems the problem lies in GSDX, after using the r1888 beta with an older gsdx it is fine, the one that came with r1888 has the issue, so the problem happened somewhere between the 0.9.6 release and SVN r1873 (the version stamped on GSDX).  I can also confirm it happens in software and hardware mode.

I am trying to get hold of @prafullpcsx2 to help narrow it down to see if the changelog (sorry, no code :( ) will help give any clues.

I have however provided a GSDump which you can download from here: http://forums.pcsx2.net/Thread-PCSX2-FMV-playback-issue-Wallace-And-Gromit-Curse-Of-The-Were-Rabbit?pid=546694#pid546694

Screenshots can also be seen in that threads first post showing the issue

gregory38:
Could be the vertex tracing due to NaN/ infinite float(I don't know when it was added). File GSVertexTrace*
Anyway the old code can be found in a branch (check mine)

gregory38:
Actually it is playground code

refractionpcsx2:
It is old yeah, Praful went ahead and searched overnight and narrowed it down to svn 1285-1310

refractionpcsx2:
Suspect 1 of 2
https://github.com/PCSX2/pcsx2/commit/ffda8ead30584973a6a4cbd406960d1f4de675b5

refractionpcsx2:
The other suspect was in pcsx2 so it wasn't that. I'll look through the list of commits (tortoisegit makes it super easy to just check gsdx commits within a date range!) and post and others which could cause issues

refractionpcsx2:
Ok other potentials: 
4/8bit transfers: https://github.com/PCSX2/pcsx2/commit/65b90bc
Shader stuff (unlikely as sw is affected): https://github.com/PCSX2/pcsx2/commit/02167e9

and the one I posted above, that's all the relevant GSDX changes, there was another for dx9 only, but I doubt it's that, if you want to see it it's here: https://github.com/PCSX2/pcsx2/commit/db8feac

ssakash:
FWIW I'm also at a vacation right now similar to Gregory,  I could probably only look into this next week. 

However kudos to ref and prafull, they're on a mission here!

refractionpcsx2:
[...]

PR #1748, opened 2017-01-07
gsdx: wrap the block number

gregory38:
Please test the performance impact

Related to issue #1719

I think SW is good but small glitches might remain in the HW renderer

### DIFF
--- a/plugins/GSdx/GSLocalMemory.h
+++ b/plugins/GSdx/GSLocalMemory.h
@@ -188,46 +188,46 @@ public:
 
 	static uint32 BlockNumber32(int x, int y, uint32 bp, uint32 bw)
 	{
-		return bp + (y & ~0x1f) * bw + ((x >> 1) & ~0x1f) + blockTable32[(y >> 3) & 3][(x >> 3) & 7];
+		return (bp + (y & ~0x1f) * bw + ((x >> 1) & ~0x1f) + blockTable32[(y >> 3) & 3][(x >> 3) & 7]) % MAX_BLOCKS;
 	}
 
 	static uint32 BlockNumber16(int x, int y, uint32 bp, uint32 bw)
 	{
-		return bp + ((y >> 1) & ~0x1f) * bw + ((x >> 1) & ~0x1f) + blockTable16[(y >> 3) & 7][(x >> 4) & 3];
+		return (bp + ((y >> 1) & ~0x1f) * bw + ((x >> 1) & ~0x1f) + blockTable16[(y >> 3) & 7][(x >> 4) & 3]) % MAX_BLOCKS;
 	}
 
 	static uint32 BlockNumber16S(int x, int y, uint32 bp, uint32 bw)
 	{
-		return bp + ((y >> 1) & ~0x1f) * bw + ((x >> 1) & ~0x1f) + blockTable16S[(y >> 3) & 7][(x >> 4) & 3];
+		return (bp + ((y >> 1) & ~0x1f) * bw + ((x >> 1) & ~0x1f) + blockTable16S[(y >> 3) & 7][(x >> 4) & 3]) % MAX_BLOCKS;
 	}
 
 	static uint32 BlockNumber8(int x, int y, uint32 bp, uint32 bw)
 	{
 		// ASSERT((bw & 1) == 0); // allowed for mipmap levels
 
-		return bp + ((y >> 1) & ~0x1f) * (bw >> 1) + ((x >> 2) & ~0x1f) + blockTable8[(y >> 4) & 3][(x >> 4) & 7];
+		return (bp + ((y >> 1) & ~0x1f) * (bw >> 1) + ((x >> 2) & ~0x1f) + blockTable8[(y >> 4) & 3][(x >> 4) & 7]) % MAX_BLOCKS;
 	}
 
 	static uint32 BlockNumber4(int x, int y, uint32 bp, uint32 bw)
 	{
 		// ASSERT((bw & 1) == 0); // allowed for mipmap levels
 
-		return bp + ((y >> 2) & ~0x1f) * (bw >> 1) + ((x >> 2) & ~0x1f) + blockTable4[(y >> 4) & 7][(x >> 5) & 3];
+		return (bp + ((y >> 2) & ~0x1f) * (bw >> 1) + ((x >> 2) & ~0x1f) + blockTable4[(y >> 4) & 7][(x >> 5) & 3]) % MAX_BLOCKS;
 	}
 
 	static uint32 BlockNumber32Z(int x, int y, uint32 bp, uint32 bw)
 	{
-		return bp + (y & ~0x1f) * bw + ((x >> 1) & ~0x1f) + blockTable32Z[(y >> 3) & 3][(x >> 3) & 7];
+		return (bp + (y & ~0x1f) * bw + ((x >> 1) & ~0x1f) + blockTable32Z[(y >> 3) & 3][(x >> 3) & 7]) % MAX_BLOCKS;
 	}
 
 	static uint32 BlockNumber16Z(int x, int y, uint32 bp, uint32 bw)
 	{
-		return bp + ((y >> 1) & ~0x1f) * bw + ((x >> 1) & ~0x1f) + blockTable16Z[(y >> 3) & 7][(x >> 4) & 3];
+		return (bp + ((y >> 1) & ~0x1f) * bw + ((x >> 1) & ~0x1f) + blockTable16Z[(y >> 3) & 7][(x >> 4) & 3]) % MAX_BLOCKS;
 	}
 
 	static uint32 BlockNumber16SZ(int x, int y, uint32 bp, uint32 bw)
 	{
-		return bp + ((y >> 1) & ~0x1f) * bw + ((x >> 1) & ~0x1f) + blockTable16SZ[(y >> 3) & 7][(x >> 4) & 3];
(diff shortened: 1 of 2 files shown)


prafullpcsx2:
Yes SW mode is fine in windows too but there is a single small black box in HW mode.

FlatOutPS2:
> Yes SW mode is fine in windows too but there is a single small black box in HW mode.

Yes, and the same goes for Thrillville.

gregory38:
Yes I saw it too on a dump. I didn't debugged it yet. But I'm not sure it is doable. Was initial HW code fully correct ?

prafullpcsx2:
Thrillville: off the rails is also crashing the emu after the second fmv now.

gregory38:
SW or HW or both ?

gregory38:
Ok I read the end :)

gregory38:
I guess you don't have a backtrace

refractionpcsx2:
God of war video is now fixed in SW mode also, in HW mode it instantly crashes as soon as it hits the FMV now.

gregory38:
The one ? Which video ? Yes I would love a backtrace.

refractionpcsx2:
The intro video as soon as you start the game.

Well that's kind of annoying, it doesn't crash in debug >.< I will run it with an SSE2 build and see if i get any sense out of it, might have to fiddle with it and get it to build with symbols.

refractionpcsx2:
Ok here you go https://s30.postimg.org/sejof9l67/Go_WCrash.png

gregory38:
Oh crap I need to think about it. The texture cache need the texture area. However it is lost with the wrapping.

refractionpcsx2:
Cool, you've got what you need then :P

gregory38:
It will be complex to fix :( Besides I need to be very careful to avoid some memory leaks

refractionpcsx2:
as a temp measure could you truncate the memory address the texture cache so it acts like it did before the PR but still works in SW mode?

gregory38:
If you look at GetPagesCoverage you can see the Insert is more or less clamped. It think the issue is the new line might wrap.
Potentially we need to wrap page. And update the condition is the delete function. Either compute a wrapped end or exit the loop when we don't find the texture. The global structure is an array (entry by page) of list (that contains the pointer to the texture). When a texture cover page1 and page2, we add it to the list m_map[page1] and m_map[page2]. When we need to delete the texture we need to remove it in all the lists

refractionpcsx2:
I'll leave it to you :P

gregory38:
You're not my friend anymore :p

I forget to said that looping over all lists of the cache isn't free for the CPU. Vulkan or not ;)

I think the safest implementation will be to get back the texture coverage and to iterate on all covered page. This way it will be symmetrical to the insert. (And potentially fix the coverage to wrap properly)

refractionpcsx2:
I guess if you can detect the wrappnig, or save it in the texture cache that it's wrapped, we can reduce the overhead when it isn't wrapping (which I guess is most of the time)

gregory38:
I did the safest implementation. It need to be benchmarked to see if it is too slow.

refractionpcsx2:
Videos seem fast enough, GS usage sits at 4%.  There is still a line missing in the GoW intro video in HW mode, just not all the way to the bottom. Same deal in Wallace & Grommit

gregory38:
We need to look at perf impact in heavy texture game.


I didn't look yet at the missing part in HW mode. I'm afraid it will be complex to fix.

refractionpcsx2:
I thought SotC would be considered a heavy game xD You got any suggestions? Baldurs Gate maybe?

gregory38:
Zone of ender. Baldur Gate. Sotc uses cell shading. So it uses a huge quantity of vertices but few texture.

refractionpcsx2:
Okay, so any clues to games which are texture heavy? I thought Baldurs Gate would be because it uses lots of small textures.

gregory38:
Sorry misunderstanding I mean yes for BG or ZoE

refractionpcsx2:
right okay, you confused me :P I thought you were saying they used high number of verticies but very few textures, so it wouldn't be a good test lol.

Trying BG2 now

refractionpcsx2:
BG2 has roughly a 15fps loss at the point you first enter the game.  ~365fps vs ~380fps

refractionpcsx2:
so is m_target NULL if the texture wraps?

FlatOutPS2:
Based on a couple of GS dump tests I noticed a 5 to 10 % drop in performance. Not the best of benchmarks, but it's about the same drop in performance as @refractionpcsx2 's tests. So it probably is about that bad.

gregory38:
Ok. I will change it.

gregory38:
So far there is 2 ideas. I can detect if we go above max page in coverage (hence wrapping). However I'm not sure it will be enough. I'm afraid some game uses 1024x1024 texture (because size doesn't mean anything on GS world). Potentially it is much slower because we do handle the wrapping properly. (Could you quickly check the perf without last commit). So we insert/remove in more list. First it might worth to check SH FMV. 

2nd idea is to limit the coverage page between base page and max page. It would restore the initial behavior.

refractionpcsx2:
Do we want the initial behaviour? I thought that's what we wanted to get away from, or am I missing the point?

Well, if games did use 1024x1024 textures, this problem could have cropped up a lot more than we imagine, causing alsorts of glitches.  But IMO i think the number of textures that huge will be limited and only to really fancy games, wrapping isn't something on the emulator side that crops up overly often, but if we can avoid the slow code 80-90% of the time, I'd be happy with that.

FlatOutPS2:
> First it might worth to check SH FMV.

No change on any level. :P

gregory38:
Ref,
The texture size is the default size for the automatic clamping. You can configure the clamping manually. So you can create a game that only uses 1024x1024 textures. But actually uses chunk of 8x8. You must see it as the max buffer. It doesn't mean that it will contains useful data.

I will try the fast path detection but I'm afraid it won't be enough. (Framebuffer is often at the start of the memory so textures are at the end). Maybe we can create a faster list container. I have an idea but potentially I can find something in the EA STL (a vector that is handled as a list)

I also need to understand what is the issue with the missing page.

refractionpcsx2:
You are the master of solutions, you've got this! :P

gregory38:
Ok boy new version available. Could you tell me both the HW and the SW performance impact.

FlatOutPS2:
Great, the annoying single small block In HW mode is gone. 👍

FlatOutPS2:
Performance seems slightly worse(a few FPS less) than in the previous version. :(

refractionpcsx2:
God of War video also fixed :)

gregory38:
@FlatOutPS2 performance of HW or SW renderer. I would like to know if the perf impact is from the 1st or the 2nd commit.

gregory38:
OpenGL or directX by the way ?

FlatOutPS2:
It seems my previous test was compromised, because I could not reproduce the results. It now looks like it is actually better than the previous PR, but still a little worse than master with all renderers.

gregory38:
Just to be sure by `all` you really mean both SW/HW renderer and also both dx/gl.

FlatOutPS2:
Yes. Although D3D11 did give some odd results in Baldur's Gate, that indicate it might be slower at times. But I couldn't get a reliable result on that, and other games didn't seem to be slower with D3D11.

gregory38:
ok. Any number for little worse ? Maybe we can add back the virtual fifo implementation. But I want to avoid extra complexity for no gain.

FlatOutPS2:
>ok. Any number for little worse ? 

I'd say about 1-5%. But it was hard to get a solid read, it fluxuated too much. Baldur's Gate especially seems to get faster by time. If you let it run at a single location without moving, it creeps up the FPS little by little for a while.

gregory38:
@FlatOutPS2 could you test the first commit only ? 4c4c7ba35cd07cce95c9d980408850b24d2b6a90

gregory38:
So we have 2 possibilities
* wrap block pointer computing with modulo
   * always enabled
   * cross platform
   * speed impact
* Use a V2P mirror
    * can be enabled as hack
    * per platform implementation (I have Linux, @FlatOutPS2 do you still have the window code)
    * similar speed impact

I think the speed impact come from the extra transfer. Currently if you have a 1024x1024 texture at the end of the memory, only a couple of block will be loaded. Most games won't use them so it isn't a big deal. With the new implementation the full texture will get data. Potentially another issue could be that memory transfer isn't contiguous so transfer is more complicated.

gregory38:
@FlatOutPS2 do you remember where you posted the Windows virtual fifo equivalent ?

It doesn't worth to slowdown all game for only a couple of fixes. So I propose to use yet another option. The option will enable the new texture cache behavior (already in master). And it will handle the wrapping with V2P mirroring. It would be too annoying to implement it as an option on all textures access. This way perf impact won't be null when the option isn't enabled.

FlatOutPS2:
I believe my Windows code didn't fix the Wallace & Grommit issue. So it's either not the latest version I had made, it's missing some updates you did for the Linux equivalent, or I just need to improve it beyond only fixing SH3. :p

gregory38:
The HW misses some updates that I did recently. The SW ought to be fine (I'm not 100% sure). Try to rebase on latest master.

FlatOutPS2:
I rebased the code and it definitely does fix W&G now(HW and SW). :)

So the plan is to add the hidden option to the GUI?

gregory38:
Cool. Yes I'm waiting the 100th commit ;) Don't know the status of the other PR.

gregory38:
Could you PR your code ? Or do we prefer to wait until GUI is done ?

FlatOutPS2:
I'd prefer to wait until GUI and your Linux update is done.

PR #1751, opened 2017-01-08
LilyPad: Another GUI update and add more new functions

FlatOutPS2:
Updates the UI by moving some functions on their respective tabs, removes "Ignore Key" and "Send escape on window close" hacks, and removes "L3 toggles Turbo" hack and replaces it with a bindable Turbo Button on the Pad tabs. The "Turbo" checkbox in the input configuration has been renamed "Rapid Fire".

Adds a new option called "Exclude Input" to the Pad tabs. This option disables an input so it will be ignored when trying to bind another input. This is helpful when binding controls for a device with an input that's difficult to center or faulty and always registered as active.

And adds neGcon as a pad type. This controller can be used in several PS1 and PS2 titles such as the Ridge Racer and Ace Combat series.


turtleli:
...Note to self: consider refactoring LilyPad.

FlatOutPS2:
...Note to self: setup a Linux build environment.

gregory38:
Note to not-myself add a Linux GUI ;)

plugins/LilyPad/Config.cpp
```diff
@@ -48,7 +48,7 @@ void PCSX2_overrideConfig(GeneralConfig &config_in_out)
     config_in_out.disableScreenSaver = 0;   // Not required - handled internally by PCSX2
     config_in_out.escapeFullscreenHack = 0; // Not required - handled internally by PCSX2
     config_in_out.saveStateTitle = 0;       // Not required - handled internally by PCSX2
-    config_in_out.closeHacks = 0;           // Cannot function when used by PCSX2
+    config_in_out.closeHack = 0;            // Cannot function when used by PCSX2
```
turtleli: Is the remaining close hack useful? I'm not sure ExitProcess() allows for proper cleanup to happen.

plugins/LilyPad/LilyPad.cpp
```diff
@@ -1209,256 +1212,291 @@ u8 CALLBACK PADpoll(u8 value)
...
-
-                        query.response[3] = 0xFF;
-                        query.response[4] = b1;
-                        query.response[5] = sum->sticks[1].horiz / 2;
-                        query.response[6] = sum->sticks[1].vert / 2;
-                        query.numBytes = 7;
-                        query.lastByte = 1;
-                        DEBUG_OUT(pad->mode);
-                        return pad->mode;
+        if (pad->config || value == 0x42 || value == 0x43) {
```
FlatOutPS2: > don't know whether you need to have query.numBytes = 0; and query.queryDone = 1; too.

Those are necessary.

I used the early return in another build. Not sure why I didn't change it before pushing this PR.

plugins/LilyPad/Config.cpp
FlatOutPS2: It can be useful if you want a quick exit out of an emulator. Yes, it does say it might not do a proper cleanup, but that's been the case ever since it was introduced. I haven't found anyone ever complaining about it, but that might also be because hardly anyone has ever used it. :p

plugins/LilyPad/Config.cpp
```diff
@@ -981,9 +983,12 @@ int LoadSettings(int force, wchar_t *file)
...
     }
-    config.closeHacks = (u8)GetPrivateProfileIntW(L"General Settings", L"Close Hacks", 0, file);
-    if (config.closeHacks & 1)
-        config.closeHacks &= ~2;
+    config.closeHack = (u8)GetPrivateProfileIntW(L"General Settings", L"Close Hacks", 255, file);
+    if (config.closeHack == 1) {
+        config.closeHack = 0;
+    } else if (config.closeHack == 255) {
+        config.closeHack = (u8)GetPrivateProfileIntW(L"General Settings", L"Close Hack", 0, file);
+    }
```
turtleli: Might be better to not add old version compatibility code for an easily adjustable option - it may be more inconvenient from the user POV, but if you throw in compatibility code, who knows how long it'll stay there and that'll just make things less maintainable if this stuff builds up.

plugins/LilyPad/Linux/Config.cpp
```diff
@@ -320,9 +319,9 @@ int LoadSettings(int force, wchar_t *file)
     }
 
 
-    config.closeHacks = (u8)cfg.ReadInt(L"General Settings", L"Close Hacks");
-    if (config.closeHacks & 1)
-        config.closeHacks &= ~2;
+    config.closeHack = (u8)cfg.ReadInt(L"General Settings", L"Close Hacks");
+    if (config.closeHack == 0)
+        config.closeHack = (u8)cfg.ReadInt(L"General Settings", L"Close Hack");
```
turtleli: Ditto.

turtleli:
The commits need to be split up properly, but otherwise the code seems fine.

PR #1564, opened 2016-09-15
GSdx: Add Windows UI options for HW hacks

FlatOutPS2:
Add Windows UI options for Auto Flush and Unscale Point Line HW hacks.

And add a cancel button to the Hacks dialog to make it easier for the user to exit the window without making changes.

PR #1758, opened 2017-01-10
gsdx windows: update gui to add a HPO v2 checkbox

woj1993:
Tested.


lightningterror:
Or maybe @ssakash since he does windows gui. Gregory does linux. :)

woj1993:
I don't have any problem with it.
@ssakash Can you review it?

FlatOutPS2:
You just copied two parts from the original HPO hack. It won't even show up right now, did you even compile it to check if it works?

And IIRC it only works for OGL, so you need to grey it out for D3D renderers.

FlatOutPS2:
Check this PR I made a few months ago when I added a few HW hacks to the GUI: https://github.com/PCSX2/pcsx2/pull/1564/files

ssakash:
Are there any testcases where this performs better than the original ``Half-pixel offset hack``? If no, I don't think adding it might be an useful idea. (we already have too much hacks...)

woj1993:
I know at least one: "Captain Tsubasa". Ask Gregory if it is usable by anything else, because he marged this hack. I'm only doing windows UI for it.

gregory38:
Well it could be better for some depth effects. If you want to reduce hack you need to remove upscaling ;)

woj1993:
Hi @gregory38 to what variable I need to connect it?

gregory38:
You didn't see me. (Check Linux code and others options)

woj1993:
@ssakash Can you help me? Do you think it is better to have 3 way checkbox or ComboBox? And how to create 3 way checkbox?

FlatOutPS2:
> Are there any testcases where this performs better than the original Half-pixel offset hack? If no, I don't think adding it might be an useful idea. (we already have too much hacks...)

In most cases where the original HPO works fairly well, the new HPO hack is basically useless. Another name would be good, because people shouldn't use it instead of the original HPO.

woj1993:
Any suggestions about name?

plugins/GSdx/resource.h
```diff
@@ -107,6 +107,7 @@
 #define IDC_LARGE_FB                    2103
 #define IDC_AUTO_FLUSH                  2104
 #define IDC_UNSCALE_POINT_LINE          2105
+#define IDC_OFFSETHACK_NEW                  2106
```
FlatOutPS2: You need to fix the spacing here. '2106' should be straight below '2105'.

plugins/GSdx/GSdx.rc
```diff
@@ -143,6 +143,7 @@ BEGIN
     COMBOBOX        IDC_SPRITEHACK,88,158,58,63,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
     CONTROL         "Fast Texture Invalidation",IDC_FAST_TC_INV,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,14,102,92,8
     CONTROL         "Auto Flush", IDC_AUTO_FLUSH,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,14,128,70,8
+    CONTROL         "Half-pixel Offset New", IDC_OFFSETHACK_NEW,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,14,138,92,8
```
FlatOutPS2: There shouldn't be a space between `"Half-pixel Offset New",` and `IDC_OFFSETHACK_NEW`. Same goes for the Auto Flush line above. You can fix it that too it you feel like it. ;)

plugins/GSdx/resource.h
woj1993: Ok.

woj1993:
@FlatOutPS2 Any suggestion about name?

FlatOutPS2:
> @FlatOutPS2 Any suggestion about name?

Not yet. HPO v2 won't fix the same issues as the original HPO, and I'm not sure how many games do benefit from it. It definitely needs to be clear it's not a v2/successor of HPO. We've already seen with the mipmapping that not warning users clearly enough about the risks of using the advanced option( `Full` mipmapping), they will use it instead of the more compatible (`Basic`) and then complain about issues.

woj1993:
Maybe Half Pixel Offset alternative or in short HPO alt? If it would cause such trouble I can write in help text that is in different approach of Half Pixel Offset hack and it can work worse/better for some games.

woj1993:
@FlatOutPS2 @turtleli What with this? And can you help me with git? I don't think I properly connected main branch into mine.

FlatOutPS2:
Maybe a name like "Half-Pixel Offset Special" would work and explain in the tooltip that it is for special cases like Captain Tsubasa only, and can't be used as an alternative to the original Half-Pixel Offset.

@gregory38 Do you know a few games this definitely works better for than the normal HPO hack?

FlatOutPS2:
> I don't think 3 state checkboxes should be used - they cause a lot more confusion than comboboxes do.

In this case I think just having two seperate checkboxes is the best.

turtleli:
> In this case I think just having two seperate checkboxes is the best.

Not sure if we're on the same page. I think half pixel offset has two states, the OpenGL only half pixel offset seems to have 3 states, and I was referring only to the OpenGL only hack.

gregory38:
It could also be a good idea to merge the 2 variables option in the code. (And realign the Linux GUI)

woj1993:
@gregory38 @FlatOutPS2 @turtleli So can we agree what I have to do?

woj1993:
@gregory38 Will you do code and Linux GUI corrections?

woj1993:
@turtleli I know it is much, but can you tell me why it is not compiling?

ssakash:
> @ssakash Hi. can you help me with GUI look on windows? How to expand window and add 4 option to ComboBox?

You need to define a vector over [here](https://github.com/PCSX2/pcsx2/blob/master/plugins/GSdx/GSdx.cpp#L216) for storing the combobox items text and then use that vector on the ``ComboBoxInit()`` function.

For example, currently you have ``ComboBoxInit(IDC_OFFSETHACK, theApp.m_gs_hack, theApp.GetConfigI("UserHacks_HalfPixelOffset"));``, you need to change ``theApp.m_gs_hack`` to the name of your newly created vector. By vector, I mean the vector STL container of C++.

woj1993:
@FlatOutPS2 @turtleli @ssakash @gregory38 Is it good? gregory38 Will you do code and Linux GUI changes?

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -712,7 +711,8 @@ void GSHacksDlg::OnInit()
 	CheckDlgButton(m_hWnd, IDC_FAST_TC_INV, theApp.GetConfigB("UserHacks_DisablePartialInvalidation"));
 	CheckDlgButton(m_hWnd, IDC_AUTO_FLUSH, theApp.GetConfigB("UserHacks_AutoFlush"));
 	CheckDlgButton(m_hWnd, IDC_UNSCALE_POINT_LINE, theApp.GetConfigB("UserHacks_unscale_point_line"));
-
+
```
FlatOutPS2: You need to remove the space from this empty line.

FlatOutPS2:
Original, new var 1 and new var 2 aren't very clear descriptions. It should be something like `Normal (Recommended)`, `Special Cases 1` and `Special Cases 2` and add a note to the tooltip that the special cases are only useful in a couple of games like Captain Tsubasa.

And the new options need to be disabled for the Direct3D renderers.

woj1993:
@FlatOutPS2 Do you have contact with gregory38?

gregory38:
@woj1993 
Geez, Let the guy rest in peace. First finish the Window side. I will do the Linux when this PR is clean and ready to be merged.

woj1993:
@FlatOutPS2 What do you think of new names? I didn't wrote "Normal (Recommended)" because I think this would mislead many people into believing that it should be used all the time. P.S. When I change OpenGl to DirectX variable stays the same. Where in code I should change it.

ssakash:
>  When I change OpenGl to DirectX variable stays the same. Where in code I should change it?

You need to get the value of the renderer combobox from the parent dialog, not from the INI file.

On the other hand, the naming seems really awful. ``Special case 1`` & ``Special case 2`` makes the hack look really _hacky_, would be nice if someone could come up with a more technical name for the combobox options.

FlatOutPS2:
> On the other hand, the naming seems really awful. Special case 1 & Special case 2 makes the hack look really hacky, would be nice if someone could come up with a more technical name for the combobox options.

It would be nice if @gregory38 could tell us the difference between those two options.

woj1993:
@ssakash We probably misunderstand each other? I got value of renderer by this "bool ogl = renderer == GSRendererType::OGL_HW" but problem is that special cases 1 and 2 aren't supposed to be accessible with DirectX. I don't display them if you have DirectX selected but if you had selected them in OpenGl and switch to DirectX variable of this ComboBox stays the same and ask for this special cases. I don't know if it should have it own function or only be added to function that change renderer window?

gregory38:
Basically, 
* first implementation will offset the vertex position on **some** old frame buffer (note: the offset will impact the texture sampling too but I won't explain all the theory)
* second implementation will offset the texture sampling on **all** old frame buffer

gregory38:
Oups. previous description is old hack vs new hack.
SC1 and SC2 are very close. SC2 is more aggressive, I'm not sure it is useful actually (tbb I completely forgot that I did the SC2 case)

woj1993:
@gregory38 Any proposition about names?

gregory38:
I don't know something like that
```
OFF
Half Pixel Offset / Vertex
Half Pixel Offset / Texture
Half Pixel Offset / Texture (aggressive)
```

FlatOutPS2:
> OFF
Half Pixel Offset / Vertex
Half Pixel Offset / Texture
Half Pixel Offset / Texture (aggressive)

That would be too confusing because the new (Texture) versions won't work for the majority of games and this needs to be clear to the user.

gregory38:
The important words are the first 3 ;) Anyway add a recommended keyword.

FlatOutPS2:
Maybe a mix of new and old:
Normal (Vertex)
Special (Texture)
Special (Texture - aggressive)

plugins/GSdx/GSSetting.cpp
```diff
@@ -55,7 +55,9 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_ALPHAHACK:
 			return "Different alpha handling. Can work around some shadow problems.";
 		case IDC_OFFSETHACK:
-			return "Might fix some misaligned fog, bloom, or blend effect.";
+			return "Might fix some misaligned fog, bloom, or blend effect."
+				"Special cases are only uselful in couple of games like Captain Tsubasa."
+				"Special cases don't work for now.";
```
ssakash: Please add a single space after full stop. Also, ``uselful`` **->** ``useful``.

Also the tooltip text seems rather confusing in my opinion, first you've said ``"Special cases are only uselful in couple of games like Captain Tsubasa"`` and then, ``"Special cases don't work for now."``, a contradiction?

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -738,6 +746,7 @@ void GSHacksDlg::OnInit()
 	EnableWindow(GetDlgItem(m_hWnd, IDC_ROUND_SPRITE), !native);
 	EnableWindow(GetDlgItem(m_hWnd, IDC_SPRITEHACK_TEXT), !native);
 	EnableWindow(GetDlgItem(m_hWnd, IDC_ROUND_SPRITE_TEXT), !native);
+	EnableWindow(GetDlgItem(m_hWnd, IDC_OFFSETHACK_TEXT), !native);
```
ssakash: Please add this directly below ``EnableWindow(GetDlgItem(m_hWnd, IDC_OFFSETHACK), !native);``

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -712,7 +711,16 @@ void GSHacksDlg::OnInit()
...
+	if(ogl && theApp.m_gs_offset_hack.size()==2)
+	{
+		theApp.m_gs_offset_hack.push_back(GSSetting(2,  "Special cases 1", ""));
+		theApp.m_gs_offset_hack.push_back(GSSetting(3,  "Special cases 2", ""));
+	}
+	else if (!ogl && theApp.m_gs_offset_hack.size()==4)
+	{
+		theApp.m_gs_offset_hack.erase(theApp.m_gs_offset_hack.begin() + 2,theApp.m_gs_offset_hack.begin() + 4); //We need to remove element 4 because function deletes elements between first and second including first but excluding second.
+	}
+	ComboBoxInit(IDC_OFFSETHACK, theApp.m_gs_offset_hack, theApp.GetConfigI("UserHacks_HalfPixelOffset"));
```
ssakash: I think you need to clamp here with the index of the last pushed value in the vector, else you'll reach an indeterminate state in the combobobx when you select options exclusive to OpenGL and switch to the Direct3D renderers.

I mean something like,
```
ComboBoxInit(IDC_OFFSETHACK, theApp.m_gs_offset_hack, std::min(max_index_of_combobox, theApp.GetConfigI("UserHacks_HalfPixelOffset")));

File: plugins/GSdx/GSdx.cpp [hunk 7]
```diff
@@ -213,6 +213,11 @@ void GSdxApp::Init()
 	m_gs_hack.push_back(GSSetting(0,  "Off", ""));
 	m_gs_hack.push_back(GSSetting(1,  "Half", ""));
 	m_gs_hack.push_back(GSSetting(2,  "Full", ""));
+
```
ssakash: Is the trailing white space necessary here?

plugins/GSdx/GSSetting.cpp
woj1993: Special cases aren't working because Gregory originally used additional variable and ComboBox on Linux but want old and new offset hack merged now.

plugins/GSdx/GSSettingsDlg.cpp
woj1993: Thanks I was working on this problem. Fixing this here lead to one possible problem if user change renderer from OpenGl to DirectX and don't go into hack window. I was asking where to check this possibility?

plugins/GSdx/GSdx.cpp
woj1993: No. Sorry about it, NetBeans add it by default.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -276,6 +276,10 @@ bool GSSettingsDlg::OnCommand(HWND hWnd, UINT id, UINT code)
 			if(ComboBoxGetSelData(IDC_RENDERER, data))
 			{
 				theApp.SetConfig("Renderer", (int)data);
+				if ((data == GSRendererType::OGL_HW || data == GSRendererType::OGL_SW || data == GSRendererType::OGL_OpenCL) && (theApp.GetConfigI("UserHacks_HalfPixelOffset")>1))
+				{
+					theApp.SetConfig("UserHacks_HalfPixelOffset", 1);
```
ssakash: I think you only need to check for ``OGL_HW`` and by the way, I don't think you need this in the first place. Even if you have a value above 1 for ``UserHacks_HalfPixelOffset`` in INI file, nothing unexpected will happen when running the D3D renderer, the normal HPO hack will take place as ``UserHacks_HalfPixelOffset`` is a not a zero value. (Refer the code in TextureCache for more information)

plugins/GSdx/GSdx.cpp
```diff
@@ -214,6 +214,11 @@ void GSdxApp::Init()
 	m_gs_hack.push_back(GSSetting(1,  "Half", ""));
 	m_gs_hack.push_back(GSSetting(2,  "Full", ""));
 
+	m_gs_offset_hack.push_back(GSSetting(0,  "Off", ""));
+	m_gs_offset_hack.push_back(GSSetting(1,  "Normal (Vertex)", ""));
+	m_gs_offset_hack.push_back(GSSetting(2,  "Special (Texture)", ""));
+	m_gs_offset_hack.push_back(GSSetting(3,  "Special (Texture - aggressive)", ""));
```
woj1993: Ok.

plugins/GSdx/GSSettingsDlg.cpp
woj1993: Ok. Thx.

plugins/GSdx/GSSettingsDlg.cpp
woj1993: Added.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -712,7 +711,16 @@ void GSHacksDlg::OnInit()
...
+	if(ogl && theApp.m_gs_offset_hack.size()==2)
+	{
+		theApp.m_gs_offset_hack.push_back(GSSetting(2,  "Special (Texture)", ""));
+		theApp.m_gs_offset_hack.push_back(GSSetting(3,  "Special (Texture - aggressive)", ""));
+	}
+	else if (!ogl && theApp.m_gs_offset_hack.size()==4)
+	{
+		theApp.m_gs_offset_hack.erase(theApp.m_gs_offset_hack.begin() + 2,theApp.m_gs_offset_hack.begin() + 4); //We need to remove element 4 because function deletes elements between first and second including first but excluding second.
+	}
+	ComboBoxInit(IDC_OFFSETHACK, theApp.m_gs_offset_hack, std::min((theApp.m_gs_offset_hack.size()-1),(theApp.GetConfigI("UserHacks_HalfPixelOffset"))));
```
ssakash: I don't think this will compile, ``std::min`` only accepts parameters of the same type.

ssakash: I suggest going with something like this,

```
if(!ogl)
	{
		std::vector<GSSetting>::iterator head = theApp.m_gs_offset_hack.begin();
		std::vector<GSSetting> options(head, head + 2);
		const int combobox_limit = options.size() - 1;
		ComboBoxInit(IDC_OFFSETHACK, options, std::min(combobox_limit, theApp.GetConfigI("UserHacks_HalfPixelOffset")));
	}
	else
	{
		ComboBoxInit(IDC_OFFSETHACK, theApp.m_gs_offset_hack, theApp.GetConfigI("UserHacks_HalfPixelOffset"));
	}
```

It looks more readable in my humble opinion.

woj1993: I was thinking both size() and GetConfigI will return me int. Any suggestion how to fix this?

ssakash: > Any suggestion how to fix this?

[I gave you one ;)](https://github.com/PCSX2/pcsx2/pull/1758#discussion_r98257498)

ssakash:
By the way, the combobox width seems rather small and the text is not properly visible. Maybe the width could be increased? make sure to also increase it on the other elements for consistency.

text

plugins/GSdx/GSSettingsDlg.cpp
FlatOutPS2: > I don't think this will compile, std::min only accepts parameters of the same type.

It would compile by simply changing it to `std::min<int>`.

woj1993:
@ssakash Is there any program that I can make changes in Gui with preview in this file?

FlatOutPS2:
> @ssakash Is there any program that I can make changes in Gui with preview in this file?

Visual Studio. That's why there's a visual studio solution in main folder of the project source. :p

woj1993:
@ssakash It should be okay now.
P.S. I don't hit 100 commit barrier. :(

gregory38:
@woj1993 
```
 P.S. I don't hit 100 commit barrier. :(
```
No worry. Coding isn't easy. I did hundreds of commit that was never released.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -712,7 +711,16 @@ void GSHacksDlg::OnInit()
...
-
+	if(ogl && theApp.m_gs_offset_hack.size()==2)
+	{
+		theApp.m_gs_offset_hack.push_back(GSSetting(2,  "Special (Texture)", ""));
+		theApp.m_gs_offset_hack.push_back(GSSetting(3,  "Special (Texture - aggressive)", ""));
+	}
+	else if (!ogl && theApp.m_gs_offset_hack.size()==4)
+	{
+		theApp.m_gs_offset_hack.erase(theApp.m_gs_offset_hack.begin() + 2,theApp.m_gs_offset_hack.begin() + 4); //We need to remove element 4 because function deletes elements between first and second including first but excluding second.
+	}
```
ssakash: I don't like the checks for vector size here (``theApp.m_gs_offset_hack.size()==4`` & ``theApp.m_gs_offset_hack.size()==2`` ). There are other alternatives to do this without checking for the vector size, I'd prefer such implementation.

plugins/GSdx/GSSetting.cpp
```diff
@@ -55,7 +55,9 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_ALPHAHACK:
 			return "Different alpha handling. Can work around some shadow problems.";
 		case IDC_OFFSETHACK:
-			return "Might fix some misaligned fog, bloom, or blend effect.";
+			return "Might fix some misaligned fog, bloom, or blend effect.\n"
+				"Special cases are only useful in couple of games like Captain Tsubasa.\n"
+				"\n Special cases are disabled from code side for now.";
```
ssakash: Maybe add some note stating that ``Normal (Vertex)`` option is recommended to be used? I know you've added ``Special cases are only useful in couple of games like Captain Tsubasa.`` which insinuates that ``Normal (Vertex)`` should be preferred but best to be explicit here.

plugins/GSdx/resource.h
```diff
@@ -121,6 +121,7 @@
 #define IDC_OSDBUTTON                   2117
 #define IDC_OSD_MAX_LOG                 2118
 #define IDC_OSD_MAX_LOG_EDIT            2119
+#define IDC_OFFSETHACK_TEXT             2120
```
ssakash: In my humble opinion, this should be placed just below ``#define IDC_OFFSETHACK                  2043``.

ssakash:
By the way, here's how the hack dialog looks in this PR: (For OpenGL renderer)

hack_dialog

Some suggestions,
* You could probably have two checkboxes at a single X-axis to save some height for the dialog.
* ``OK`` and ``Cancel`` buttons are off-center, would be nice to fix them.

plugins/GSdx/resource.h
woj1993: But with this number?

plugins/GSdx/GSSettingsDlg.cpp
woj1993: What implementation we are talking about?

lightningterror:
Hmm how about adding a "save snapshot" button on the bottom left as well?

FlatOutPS2:
>Hmm how about adding a "save snapshot" button on the bottom left as well?

Those aren't in any plugins yet AFAIK, so let's not complicate this PR any further. ;)

plugins/GSdx/GSSettingsDlg.cpp
ssakash: Maybe something like this?

```
  std::vector<GSSetting> hpo_combobox = theApp.m_gs_offset_hack;
    if (!ogl)
        hpo_combobox.erase(hpo_combobox.begin() + 2, hpo_combobox.begin() + 4); //TODO: Enums would be more nicer than magic numbers.

    ComboBoxInit(IDC_OFFSETHACK, hpo_combobox, std::min<int>(theApp.GetConfigI("UserHacks_HalfPixelOffset"), hpo_combobox.size() - 1));
```

plugins/GSdx/resource.h
ssakash: I know, it should be troublesome to increment the value of all the macro constants, but I feel it's worth it. ;)

plugins/GSdx/GSSetting.cpp
```diff
@@ -55,7 +55,9 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_ALPHAHACK:
 			return "Different alpha handling. Can work around some shadow problems.";
 		case IDC_OFFSETHACK:
-			return "Might fix some misaligned fog, bloom, or blend effect.";
+			return "Might fix some misaligned fog, bloom, or blend effect.\n"
+				"Preferred option should be Normal (Vertex). Special cases are only useful in couple of games like Captain Tsubasa.\n"
```
FlatOutPS2: It'd probably be better to expend it to:
"The preferred option is Normal (Vertex) as it is most likely to resolve misalignment issues. The special...

FlatOutPS2:
You should also try and fix the "OK and Cancel buttons are off-center" issue @ssakash mentioned.

You could also do the other suggestion he mentioned, but you could leave that as I could fix that in an upcoming PR. 

And squash your commits when your done. ;)

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -712,7 +711,12 @@ void GSHacksDlg::OnInit()
 	CheckDlgButton(m_hWnd, IDC_FAST_TC_INV, theApp.GetConfigB("UserHacks_DisablePartialInvalidation"));
 	CheckDlgButton(m_hWnd, IDC_AUTO_FLUSH, theApp.GetConfigB("UserHacks_AutoFlush"));
 	CheckDlgButton(m_hWnd, IDC_UNSCALE_POINT_LINE, theApp.GetConfigB("UserHacks_unscale_point_line"));
-
+	std::vector<GSSetting> hpo_combobox = theApp.m_gs_offset_hack;
+	if (!ogl)
+	{
+		hpo_combobox.erase(hpo_combobox.begin() + 2, hpo_combobox.begin() + 4); //TODO: Enums would be more nicer than magic numbers.
```
ssakash: >  //TODO: Enums would be more nicer than magic numbers.

Why not add them then? ;)

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -712,7 +711,12 @@ void GSHacksDlg::OnInit()
...
+	}
+	ComboBoxInit(IDC_OFFSETHACK, hpo_combobox, std::min(static_cast<int>(theApp.m_gs_offset_hack.size()-1),static_cast<int>(theApp.GetConfigI("UserHacks_HalfPixelOffset"))));
```
ssakash: ``std::min(static_cast<int>(theApp.m_gs_offset_hack.size()-1),static_cast<int>(theApp.GetConfigI("UserHacks_HalfPixelOffset")``

Is the clamp necessary anymore?

ssakash:
Not your fault but seems like someone overlooked to update the ``_APS_NEXT_CONTROL_VALUE`` [macro's value](https://github.com/woj1993/pcsx2/blob/7f883ab2b11401f83a017cabb8780f3714f1803d/plugins/GSdx/resource.h#L144) in resource.h, could you update it's value in this PR?

By the way, providing screenshot of the current state of HW hacks dialog would be nice.

woj1993:
What value is should have?

plugins/GSdx/GSSettingsDlg.cpp
FlatOutPS2: Don't bother. ;)

ssakash:
> What value is should have?

Currently, the latest GUI element ``(#define IDC_OSD_MAX_LOG_EDIT)`` seems to have a value of 2120, so ``_APS_NEXT_CONTROL_VALUE`` should be 2121.

woj1993:
@FlatOutPS2 How to squash commits on Github? Can it be this way?

FlatOutPS2:
@woj1993 One way of squashing the commits is by `git rebase -i HEAD~38` (38 being the number of commits), then, in the file that opens, replacing all the `pick` 's below to original commit with `fix`. This way, you can also easily edit the original commit message if you need to update it.

The OK and cancel buttons in the screenshot still seem slightly offcenter and it looks like you can reduce the width of the dialog window if you move the checkboxes to the right a bit.

gregory38:
`git rebase -i master` to avoid to count commit

FlatOutPS2:
> git rebase -i master to avoid to count commit

I'm not sure that works here as one of the commits merged the master into the PR.

plugins/GSdx/GSdx.rc
```diff
@@ -111,39 +111,40 @@ IDB_PSX_NULL            BITMAP                  "res\\psx-logo-null.bmp"
 // Dialog
 //
 
-IDD_HACKS DIALOGEX 0, 0, 161, 235
+IDD_HACKS DIALOGEX 0, 0, 219, 225
```
FlatOutPS2: The width should be reduced by 1 to `IDD_HACKS DIALOGEX 0, 0, 218, 225`

plugins/GSdx/GSdx.rc
```diff
@@ -111,39 +111,40 @@ IDB_PSX_NULL            BITMAP                  "res\\psx-logo-null.bmp"
...
 STYLE DS_SETFONT | DS_MODALFRAME | DS_FIXEDSYS | WS_POPUP | WS_CAPTION | WS_SYSMENU
 CAPTION "Hacks Configuration"
 FONT 8, "MS Shell Dlg", 400, 0, 0x1
 BEGIN
-    DEFPUSHBUTTON   "OK",IDOK,27,214,50,14
-    DEFPUSHBUTTON   "Cancel",IDCANCEL,84,214,50,14
```
FlatOutPS2: And change the X-offset of the OK and Cancel to:
```
    DEFPUSHBUTTON   "OK",IDOK,56,199,50,14
    DEFPUSHBUTTON   "Cancel",IDCANCEL,112,199,50,14
```

And then the PR can be squashed.

woj1993:
@FlatOutPS2 Sorry about long break. Can you look why it isn't passing all check?

FlatOutPS2:
@woj1993 It looks like a random failure, nothing to do with your changes. If you squash the commits it should build again without issue.

woj1993:
@FlatOutPS2 It says that "error: could not apply da57014eb... GSdx OSD: Add Windows UI and new option" what I am supposed to do now. Resolve it in IDE?

gregory38:
`git rebase --abort`

woj1993:
@gregory38 And next?

gregory38:
I don't know. I only have my phone. Maybe you have a merge conflict. I don't know what you did. Above command is handy to reset the mess.

FlatOutPS2:
You could probably remove the early commits before the merge with master commit, because all those changes have likely been replaced by the later commits.

woj1993:
@FlatOutPS2 But the error is after marge.

plugins/GSdx/resource.h
```diff
@@ -45,82 +45,160 @@
...
+#define IDC_OSD_SIZE_EDIT               2112
+#define IDC_OSD_TRANSPARENCY_SLIDER     2113
+#define IDC_OSD_TRANSPARENCY_TEXT       2114
+#define IDC_OSD_LOG                     2115
+#define IDC_OSD_MONITOR                 2116
+#define IDC_OSD_INDICATOR               2117
+#define IDC_OSDBUTTON                   2118
+#define IDC_OSD_MAX_LOG                 2119
+#define IDC_OSD_MAX_LOG_EDIT            2120
+#define IDC_OFFSETHACK_TEXT             2044
```
gregory38: Is it me, or resource define was duplicated in your merge?

woj1993:
@gregory38 You are correct about 2 defines. Squashing didn't go how it should. But it should be ok now.
@FlatOutPS2 I squashed it to 3 commit. I can't squash it anymore. It need review because of this squashing. But everything should be ok.

woj1993:
@FlatOutPS2 Is everything okay with this pull request?

FlatOutPS2:
@woj1993 The resource entries are still duplicated, but the rest is okay.

woj1993:
@FlatOutPS2 Apparently there was weird 3 duplicates. Can I ammend last commit with resource.h proper fix?

FlatOutPS2:
> Apparently there was weird 3 duplicates. Can I amend last commit with resource.h proper fix?

Yes, that or making a new commit and squashing it into the last commit will both work.

woj1993:
@FlatOutPS2 This time it should really be fixed.

FlatOutPS2:
Yes, it's all back to normal now. 👍

plugins/GSdx/GSSettingsDlg.cpp
ssakash: As of https://github.com/PCSX2/pcsx2/commit/eda22c241d64e50702fbdde95d53a3fcb244c426, this clamp is no longer necessary. Do you agree?

woj1993:
@ssakash @gregory38 @turtleli Everyone agree that it can be marged?

gregory38:
@woj1993 
What do you I need to do ? Did I miss something ?
* Linux GUI
* plug the various level of the option instead of the 2 old options

@ssakash 
Are we fine ? I want to merge the code soon to add the other options.

FlatOutPS2:
@woj1993: @ssakash requested this change: https://github.com/PCSX2/pcsx2/pull/1758#discussion_r102946388

woj1993:
@FlatOutPS2 Help me in resolving conflict problem.  Git is telling me that GSSettingsDlg.cpp is conflicting with master but when I click on resolve conflict it tells me all my file is in conflict to master. an you look what is happening?

woj1993:
@FlatOutPS2 Ok. ssakash sugestion was taken into acount. Redy to be merged.
P.S. Github is drunk. He had truble mergin GSDialog.cpp which I must marge by hand. Now he thinks I write all lines of the file. :) I will make diff of file on master branch and my and will post it here if you want?

FlatOutPS2:
I don't think it's a good idea to merge it like this. It will mess with git blame, making it seem like you changed the entire file. It might help if you amend the `Repairing resource.h and deleting unneded clamp.` commit and remove the GSSettingsDlg.cpp changes from it. Then ad the changes back in a new commit after the second master merge.

woj1993:
@FlatOutPS2 Any specific instruction? Github always think I changed almost every line of GSSettingsDlg.cpp. It doesn't matter if I use old file version before margin with master or if I use after file after the merge without this change one change that started this hell.

gregory38:
I think it is due to the `end of line`. You need to edit the commit to restore the original `eol` special char.

woj1993:
@gregory38 How can I do that? I don't even see EOL character.

FlatOutPS2:
Try adding an empty line at the end of the file.

gregory38:
@woj1993 just do the ssakash requested change in a separate commit. I will handle the squash and the end of line.

woj1993:
@gregory38 Ready. Good luck.

FlatOutPS2:
You haven't removed the GSSettingsDlg changes from the previous commit though.

woj1993:
?

woj1993:
@gregory38 This is what you wanted or did I do something wrong again?

gregory38:
Give me a couple of minutes ;)

gregory38:
Actually I would more time ;) I don't know what you did, but you kill the poort git.

woj1993:
Poor git? He is killing me! Squashing commits? Problem. Merging commit? Conflict. Changing one line of a file? Conflict in every line of this file.
I literally changed one line in a file. Apparently this is enought to kill. Even in commits before this change.

gregory38:
I'm busy with real life. I will try to fix it in coming day. Don't touch anything until then.

gregory38:
For the record, you change the end of line with CRLF. Maybe something is wrongly set in your text editor.

gregory38:
@woj1993 @FlatOutPS2 can you tell me how much I screw up in the the PR #1840

FlatOutPS2:
Why not just push it to this PR ? :S

gregory38:
@FlatOutPS2 due to git conflict I redid a branch from scratch

FlatOutPS2:
I cloned the branch and it wasn't that hard to fix without starting from scratch. :D

gregory38:
I try to rebase and got tons of commit from master. But maybe I did something wrong (I'm not a git expert). It is still the work of woj1993. I just renamed the branch.

FlatOutPS2:
> I try to rebase and got tons of commit from master. 

Yes, that's due to the merge with master commit(it shows all of the commits contained in that single commit when you do a rebase). You can just remove the merge with master commit, it just causes a minor conflict in resource.h to resolve.

gregory38:
I wasn't sure because first commit was kind of merged with others commit too. I mean the commmit before the merge commit.

woj1993:
Yes. It was weird because when I rebased for the first time git didn't show me commit margin main branch into my.

FlatOutPS2:
> I wasn't sure because first commit was kind of merged with others commit too. I mean the commmit before the merge commit.

Yes, but if you merge with master before the commits of this PR those changes from other commits automatically disappear from that first commit.

The PR looked to be in a worse state than it was. ;)

woj1993:
You mean it was it worse state that it looked?

FlatOutPS2:
> You mean it was it worse state that it looked?

No, it was in a good state. It just looked really bad because github tries to mess with you.

woj1993:
It is good to know that GitHub hates me. :)

FlatOutPS2:
> It is good to know that GitHub hates me. :)

You can take comfort in the fact that it hates us all equally. :P

ssakash:
> No, it was in a good state. It just looked really bad because github tries to mess with your **head**.

toucan

np511:
For the record, you change the end of line with CRLF. Maybe something is wrongly set in your text editor.


Can't you just run `dos2unix` on the file then?

gregory38:
@np511 I can see the CRLF character in vim so for me it is easy to change. I don't know if `dos2unix` exist in windows. The easiest solution is to use correct editor config.

gregory38:
The branch was merged witht the other PR. So I'm closing this one.

PR #1760, opened 2017-01-11
gsdx tc: allow a hit on older depth buffer

### LINKED ISSUE #1757: Full Spectrum Warrior - GSdx HW vertical lines regression
***PCSX2 version:***
v1.5.0-dev-1779-g83b5bdb

***PCSX2 options:***
Default + VU1 set to SuperVU

***Plugins used:***
Default

Vertical lines flicker onscreen.

Use Hardware renderer and load a mission.

Regression caused by:
https://github.com/PCSX2/pcsx2/pull/1185

GS Dump: 
[Full Spectrum Warrior - vertical lines regression.zip](https://github.com/PCSX2/pcsx2/files/696708/Full.Spectrum.Warrior.-.vertical.lines.regression.zip)

gregory38:
Avoid an issue when rendering take some shortcut (Full Spectrum Warrior)

Close #1757

### DIFF
--- a/plugins/GSdx/GSTextureCache.cpp
+++ b/plugins/GSdx/GSTextureCache.cpp
@@ -174,6 +174,12 @@ GSTextureCache::Source* GSTextureCache::LookupDepthSource(const GIFRegTEX0& TEX0
 		// Note: might worth to check previous frame
 		// Note: otherwise return NULL and skip the draw
 
+		// Full Spectrum Warrior: first draw call of cut-scene rendering
+		// The game tries to emulate a texture shuffle with an old depth buffer
+		// (don't exists yet for us due to the cache)
+		// Rendering is nicer (less garbage) if we skip the draw call.
+		throw GSDXRecoverableError();
+
 		//ASSERT(0);
 		return LookupSource(TEX0, TEXA, r);
 	}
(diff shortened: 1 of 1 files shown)


FlatOutPS2:
The main issue is gone, but I noticed there's still a flash of vertical lines before a cutscene.

gregory38:
Hum, could you try to do the same pattern on the FIXME line (check the diff you will understand)

FlatOutPS2:
> Hum, could you try to do the same pattern on the FIXME line (check the diff you will understand)

I was already building it. :)

gregory38:
I propose that we test a couple of game to avoid extra regression. I will edit the commit message to remove the close. Then we merge it.

Is the flash really annoying?

FlatOutPS2:
> Is the flash really annoying?

Yes, but I'm not sure how often it occurs. I'll need to get a bit further in the game to figure that out. I'll try to trap it in a GS dump if it returns regularly.

FlatOutPS2:
It seems to happen before almost all cutscenes. I think I captured a decent example in this GS Dump:

http://forums.pcsx2.net/attachment.php?aid=61832

gregory38:
So adding the extra condition give `0x7FFF mask + 16 bit depth` give better result. But the rendering is still wrong, you just see it less. The texture channel effect is now trigger but input texture is garbage. Due to the depth miss. So instead, I will just skip the draw call. Rendering will still be bad but you might not see it.

FlatOutPS2:
It seems to work, I haven't see the garbage lines anymore. 👍 

I'll go check a few more cutscenes and do some regression checking.

gregory38:
Ok. I will merge it to improve the test coverage. I really hope it won't create any regression.

ssakash:
>  I really hope it won't create any regression.

Sorry for the bad news but it seems to cause regression in Jackie Chan Adventures. ( Refer https://github.com/PCSX2/pcsx2/issues/1838)

PR #1763, opened 2017-01-12
Translate Simple Chinese in .desktop file

guoyunhe:
Add Simple Chinese translation in .desktop file
Change "Playstation" to "PlayStation". (Official name)


gregory38:
Thanks

Issue #1698, opened 2016-12-15
Tales of Legendia - Regressions in CRC level


orbea:
OS: `Slackware64-current`
`pcsx2-2016.12.12_61669d1_master-x86_64-1_git`

After commit ccd9ce1728c1d25e74319862a00d5683aeca9635 Tales of Legendia suffered several regressions unless CRC level is set to full where before these issues did not occur with partial. They still occured with the CRC level set to minimum or none.
```
ccd9ce1728c1d25e74319862a00d5683aeca9635 is the first bad commit
commit ccd9ce1728c1d25e74319862a00d5683aeca9635
Author: Gregory Hainaut <gregory.hainaut@gmail.com>
Date:   Sun Sep 25 10:17:08 2016 +0200

    gsdx hack: move tales of legendia hack to full move
    
    Also remove the OI hack which is handled by the generic double half clear

:040000 040000 edb1a260fdb1a0421fb30cc237598754b9debf49 cb437c7114d7a0bafd1c9c73d78367cc0d0f9e6a M	plugins
```

1. In the title screen there will be a blue box in the top left corner.

GS dump - http://ks392457.kimsufi.com/orbea/stuff/games/pcsx2/snaps/gsdx_20161214175119_TOL_title.gs.xz
Apitrace - http://ks392457.kimsufi.com/orbea/stuff/trace/PCSX2_TOL_title.trace.xz

![1](http://ks392457.kimsufi.com/orbea/stuff/pics/scrots/pcsx2/2016-12-14-174253_1680x1050_scrot.png)

2. Certain areas will have a lot of slow down associated with them as will battles.

CRC level partial:
Field - 45 fps
Battle - 13 fps

CRC level full:
Field - 110+ fps
Battle - 60+ fps

GS dump - http://ks392457.kimsufi.com/orbea/stuff/games/pcsx2/snaps/gsdx_20161214175854_TOL_slow_field.gs.xz
Apitrace - http://ks392457.kimsufi.com/orbea/stuff/trace/PCSX2_TOL_slow_field.trace.xz

3. Shadows will display incorrectly in battles.

GS dump - http://ks392457.kimsufi.com/orbea/stuff/games/pcsx2/snaps/gsdx_20161214182214_TOL_battle.gs.xz
Apitrace - http://ks392457.kimsufi.com/orbea/stuff/trace/PCSX2_TOL_battle.trace.xz

[...]

PR #1764, opened 2017-01-12
GSdx: Remove/move CRC hacks

FlatOutPS2:
Removes Alpine Racer 3 hack. Issue has been resolved.

Moves NanoBreaker hack. Issue has been resolved for OpenGL and hack has
been moved to DX only.

Moves Tri-Ace games hacks. Hacks are also necessary for OpenGL with "Partial" CRC Hack Level to prevent massive slowdown.

Move Tales Of Legendia hack back as it's also necessary for OpenGL with "Partial" CRC Hack Level to prevent graphical issues.
Close: #1698

Added PAL and NTSC-U CRC's for Ar tonelico II.


gregory38:
Could you move back tales of legendia in gl level. There is an open issue for it. Thanks.

PR #1766, opened 2017-01-12
Greg/gsdx boost texturecache

gregory38:
Greatly reduce the texture removal operation overhead. It might help to improve GSdx speed a little.

### DIFF
--- a/plugins/GSdx/GSTextureCache.cpp
+++ b/plugins/GSdx/GSTextureCache.cpp
@@ -1597,6 +1597,9 @@ GSTextureCache::Source::Source(GSRenderer* r, const GIFRegTEX0& TEX0, const GIFR
 		{
 			m_p2t = r->m_mem.GetPage2TileMap(m_TEX0);
 		}
+
+		// will be useless with texture page coverage
+		m_erase_it.fill(list<Source*>::iterator(0));
 	}
 }
 
@@ -1969,8 +1972,10 @@ void GSTextureCache::SourceMap::Add(Source* s, const GIFRegTEX0& TEX0, GSOffset*
 		// TODO
 
 		// GH: I don't know why but it seems we only consider the first page for a render target
+		size_t page = TEX0.TBP0 >> 5;
 
-		m_map[TEX0.TBP0 >> 5].push_front(s);
+		m_map[page].push_front(s);
+		s->m_erase_it[page] = m_map[page].begin();
 
 		return;
 	}
@@ -1984,6 +1989,7 @@ void GSTextureCache::SourceMap::Add(Source* s, const GIFRegTEX0& TEX0, GSOffset*
 		if(uint32 p = pages[i])
 		{
 			list<Source*>* m = &m_map[i << 5];
+			auto* e = &s->m_erase_it[i << 5];
 
 			unsigned long j;
 
@@ -1995,6 +2001,7 @@ void GSTextureCache::SourceMap::Add(Source* s, const GIFRegTEX0& TEX0, GSOffset*
 				p ^= 1 << j;
 
 				m[j].push_front(s);
+				e[j] = m[j].begin();
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


refractionpcsx2:
error C2079: 'GSTextureCache::Source::m_erase_it' uses undefined class 'std::array<std::_List_iterator<std::_List_val<std::_List_simple_types<GSTextureCache::Source *>>>,512>'

gregory38:
:sob: I don't know what is the issue (except the VS suck).

refractionpcsx2:
I'm on it don't worry :P I'll commit it when I fix it

refractionpcsx2:
array<list<Source*>::iterator, MAX_PAGES> m_erase_it;  

saying "incomplete type not allowed"

refractionpcsx2:
also 

"cannot deduce 'auto' type" on line 1989 and 2091 in texturecache

gregory38:
auto will work when m_erase_it is properly detected. What happen if you remove the `::iterator` does it manage to detect the m_erase_it type (obviously it will fail later).

refractionpcsx2:
Nope but if i make it a pointer it works

array<list<Source*>::iterator, MAX_PAGES> *m_erase_it;

gregory38:
Otherwise try to forward declare Source before m_erase_it.
```class Source
blabla m_erase_it
```

refractionpcsx2:
> Otherwise try to forward declare Source before m_erase_it.

What? lol

gregory38:
See example above. Just telling the compiler that Source is a type which might not be yet true inside Source.

refractionpcsx2:
I don't understand your example... what exactly do you want me to do?

gregory38:
http://stackoverflow.com/questions/553682/when-can-i-use-a-forward-declaration

turtleli:
? Not sure how this even compiles on Linux. I think the `<array>` include is missing?

gregory38:
Ah it must be it. I guess it depends on the order of the include (or issue is include from Dx).

gregory38:
let's see if it is better.

refractionpcsx2:
it compiles here now :)

gregory38:
Great. Conclusion, Linux is much better for dev. You give garbage, and it replies with a working program. VS choose the reverse way :stuck_out_tongue: 

Now, let's hope perf is better !

refractionpcsx2:
It's not like Linux not to be anal about something... :P Very out of character!

Going to test a few games now Vs Master

gregory38:
(I hope it isn't slower). The speed impact might be bigger on game that are heavy on texture such as Zone of Ender, baldur gate.

refractionpcsx2:
With MTVU off, Baldurs gate seems possibly 2-5fps faster. so I'm hitting 335fps roughly, instead of 330fps

lightningterror:
2-5 fps is a good improvement.

gregory38:
Faster is still better than slower. Besides it is more robust and might allow to fix some bugs.
It would have been nice to test the Anubis boss on ZoE. Maybe it will be less slow.

refractionpcsx2:
Sorry I don't have ZoE

gregory38:
No worry. Maybe someone will shine.

gregory38:
Could this PR be merged ? So I can continue the work on the GS memory wrapping (#1748)

refractionpcsx2:
It seems to work, so sure :)

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -2070,16 +2074,34 @@ void GSTextureCache::SourceMap::RemoveAt(Source* s)
...
+
+				unsigned long j;
 
-			if(*j == s) {m.erase(j); break;}
+				while(_BitScanForward(&j, p))
+				{
+					// FIXME: this statement could be optimized to a single ASM instruction (instead of 4)
+					// Either BTR (AKA bit test and reset). Depends on the previous instruction.
+					// Or BLSR (AKA Reset Lowest Set Bit). No dependency but require BMI1 (basically a recent CPU)
+					p ^= 1 << j;
```
turtleli: I suppose I wonder whether this should actually be `p ^= 1U << j;` (I think 1 << 31 is undefined behaviour, but 1U << 31 is not).

As for merging -  You can go ahead, though I kinda want `std::` to be prefixed onto array and list ;)

MarcoEstevez:
Good job

Issue #1552, opened 2016-09-06
Meta: AMD Issues/Workarounds


mirh:
Follows [#1508](https://github.com/PCSX2/pcsx2/pull/1508#issuecomment-243181270) and hrydgard/ppsspp#8698
Gs dump is [here](https://github.com/PCSX2/pcsx2/pull/1508#issuecomment-243237225). 

Bissected up to either 16c2baa0df2d7859619d51d3995b78f057a8e965 or 29c97a9bf21a985e1524e0b428ff97aa678adcc4
Happens in Ace Combat 5 after "press start" screen _only_ after Blending Unit Accuracy has been set to none in OGL hw. 

I'd just complain over at AMD but I'd like to get a more straightforward testcase for them. 
Wouldn't be bad if somebody added "Upstream | External" label

List of AMD issues:
- ~SSO is broken. SSO is partially disabled in GSdx due to this issue.~
- ~Disabling blending causes a BSOD. This is related to SSO as the issue does not happen on v1.4 where SSO isn't mandatory and is disabled on AMD GPUs.~
- Dreadful performance. OpenGL can be slower by about 80% compared to Direct3D with same effects emulated.

Links to AMD forum issue threads:
https://community.amd.com/message/2748362
https://community.amd.com/message/2756964

**Possible BSOD Citra Workaround**
Merged from issue #2362 
As Gregory requested so we don't forget about it.

Currently Citra added a workaround for the amdfail driver that fixes the crashing caused by SSO.
The commit is located here https://github.com/citra-emu/citra/pull/3499/commits/0cf6793622b01f3941fbc77fe04c3b68476004ca

Reddit post:
https://www.reddit.com/r/emulation/comments/88vva4/citra_on_twitter_new_update_to_the_hardware/

Idea would be for this to be checked out and maybe implemented.

Some useful info

> You unbind everything so you pay extra invalidation
> What we need to do is
> Create a pipeline by shader combination
> So you only bind once stage to a pipeline
> And then we bind and rebind pipeline
> But I'm not sure it will fix the crash
> Potentially citra workaround might not work on ours side.

gregory38:
There is a way to dump blending setup
```
https://github.com/PCSX2/pcsx2/blob/master/plugins/GSdx/GSRendererOGL.cpp#L675
```
Replace this line with `if 1`

Normally tracing can be enabled with the `debug_opengl = 1` on dev/dbg build. As the bonus it will validate openGL functions call.

[...]

Issue #1121, opened 2016-01-16
AMD OGL issues


mirh:
Gregory used to mention [GL_ARB_separate_shader_objects](https://community.amd.com/thread/194895)
Probably there are more (e.g. multithreaed optimization)

@bositman could you retweet that link to amd?
https://twitter.com/grahamsellers

gregory38:
Please tweet it at graham sellers
Ask it if him or an AMD engineer can give it a look at the testcase.
Thanks you.

bositman:
I don't understand what I'm supposed to do? :P

ssakash:
he's asking you to tweet the [bug report](https://community.amd.com/thread/194895) to [graham sellers](https://twitter.com/grahamsellers) on twitter, not sure why he's specifically asking you,,,,

gregory38:
Dont do it. Delroth fordward the bug report to his amd contact.
(and I want to keep my tweeter virginity ;) )

gregory38:
@mirh 
So, did they add the fix in their Lara's driver ?

mirh:
Uh, no, they didn't. 
I thought I already mentioned that on IRC.

gregory38:
Let's close the issue. AMD will release a fix in the next major driver release.

mirh:
Next™

avih:
Not sure if that's how it "supposed" to be right now, but for me on AMD gpu it's not only slow, but also broken visually on many (most) games. That's what I see at my log:
```
3.x GL context successfully created
OpenGL information. GPU: AMD Radeon R9 M200X Series. Vendor: ATI Technologies Inc.. Driver: Core Profile Forward-Compatible Context 15.200.1046.0
INFO: GL_EXT_texture_filter_anisotropic is available
INFO: GL_ARB_gpu_shader5 is available
INFO: GL_ARB_draw_buffers_blend is available
INFO: GL_ARB_viewport_array is available
INFO: GL_ARB_shader_image_load_store is available
INFO: GL_ARB_clear_texture is available
OpenGL renderer is slow on AMD GPU due to inefficient driver. Sorry.
```

I didn't try many GSdx configurations, but currently, for OpenGL HW I have:
- CRC Hack level: partial
- Enabled HW depth
- Enabled Accurate date
- Blending accuracy: medium

gregory38:
[...]

PR #1768, opened 2017-01-13
GSdx: Windows gui updates

lightningterror:
Rename CRC Hack Level "Full (Safest)" to "Full (Direct3D Recommended)"

Add tooltip info for Blending Accuracy None that it doesn't work with
AMD cards.

### DIFF
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -367,7 +367,7 @@ namespace GLLoader {
 #ifdef _WIN32
 		if (status) {
 			if (fglrx_buggy_driver) {
-				fprintf(stderr, "The \"OpenGL (Hardware)\" renderer is slow on AMD GPU's due to an inefficient driver. Check out the links below for further information.\n"
+				fprintf(stderr, "The \"OpenGL Hardware\" renderer is slow on AMD GPUs due to an inefficient driver. Check out the links below for further information.\n"
 					"https://community.amd.com/message/2756964\n"
 					"https://community.amd.com/thread/205702\n"
 					"Note: Due to an AMD GPU OpenGL driver issue, setting Blending Unit Accuracy to \"None\" can cause an application or system crash.\n"
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -367,11 +367,11 @@ namespace GLLoader {
 #ifdef _WIN32
 		if (status) {
 			if (fglrx_buggy_driver) {
-				fprintf(stderr, "The \"OpenGL Hardware\" renderer is slow on AMD GPUs due to an inefficient driver. Check out the links below for further information.\n"
+				fprintf(stderr, "The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver. Check out the links below for further information.\n"
 					"https://community.amd.com/message/2756964\n"
 					"https://community.amd.com/thread/205702\n"
-					"Note: Due to an AMD GPU OpenGL driver issue, setting Blending Unit Accuracy to \"None\" can cause an application or system crash.\n"
-					"Keep Blending Unit Accuracy set to at least the default \"Basic\".\n"
+					"Note: Due to an AMD OpenGL driver issue, setting Blending Unit Accuracy to \"None\" can cause an application or system crash.\n"
+					"Keep Blending Unit Accuracy set to at least the default \"Basic\" level.\n"
 					"AMD has a fix for the issue that will be released in the coming months. The issue does not affect AMD GPUs on legacy drivers.\n");
 			}
 		}
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -370,8 +370,8 @@ namespace GLLoader {
 				fprintf(stderr, "OpenGL renderer is slow on AMD GPU due to inefficient driver.Please check the links below.\n"
 					"https://community.amd.com/message/2756964\n"
 					"https://community.amd.com/thread/205702\n"
-					"Note: Due to an AMD GPU OpenGL driver issue, setting Blending Unit Accuracy to "None" can cause a application or system crash.\n"
-					"Keep Blending Unit Accuracy set to at least the default "Basic".\n"
+					"Note: Due to an AMD GPU OpenGL driver issue, setting Blending Unit Accuracy to \"None\" can cause a application or system crash.\n"
+					"Keep Blending Unit Accuracy set to at least the default \"Basic\".\n"
 					"AMD have a fix for the issue that will be released in the coming months. The issue does not affect AMD GPU's on legacy drivers.\n");
 			}
 		}
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -367,7 +367,11 @@ namespace GLLoader {
 #ifdef _WIN32
 		if (status) {
 			if (fglrx_buggy_driver) {
-				fprintf(stderr, "OpenGL renderer is slow on AMD GPU due to inefficient driver. Please check the links below.\nhttps://community.amd.com/thread/205702\nhttps://community.amd.com/message/2756964\n");
+				fprintf(stderr, "OpenGL renderer is slow on AMD GPU due to inefficient driver.Please check the links below.\n"
+					"https://community.amd.com/message/2756964\n"
+					"https://community.amd.com/thread/205702\n"
+					"Note: When running OpenGL on AMD cards make sure Blending Unit Accuracy is set to at least Basic.\n"
+					"Having it set to None will cause a BSOD and crash the system.\n");
 			}
 		}
 #endif
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -367,12 +367,12 @@ namespace GLLoader {
 #ifdef _WIN32
 		if (status) {
 			if (fglrx_buggy_driver) {
-				fprintf(stderr, "OpenGL renderer is slow on AMD GPU due to inefficient driver.Please check the links below.\n"
+				fprintf(stderr, "The \"OpenGL (Hardware)\" renderer is slow on AMD GPU's due to an inefficient driver. Check out the links below for further information.\n"
 					"https://community.amd.com/message/2756964\n"
 					"https://community.amd.com/thread/205702\n"
-					"Note: Due to an AMD GPU OpenGL driver issue, setting Blending Unit Accuracy to \"None\" can cause a application or system crash.\n"
+					"Note: Due to an AMD GPU OpenGL driver issue, setting Blending Unit Accuracy to \"None\" can cause an application or system crash.\n"
 					"Keep Blending Unit Accuracy set to at least the default \"Basic\".\n"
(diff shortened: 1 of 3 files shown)


refractionpcsx2:
This is the first i've heard of Blending Accuracy causing BSOD.  I thought they'd fixed most of the bugs in the driver?

gregory38:
@refractionpcsx2 it is the contrary. No BA which use standard hardware unit generates a BSOD (various games/emulator are impacted)

However I don't think it must be added to the tooltip. AMD said they fixed it last year. They will likely release the fix in 3 months.

FlatOutPS2:
> However I don't think it must be added to the tooltip. AMD said they fixed it last year. They will likely release the fix in 3 months.

And it works fine if you use older AMD drivers.

gregory38:
Besides, they will likely add new regression in the process.

That being said, maybe we can maybe add an extra line https://github.com/PCSX2/pcsx2/blob/master/plugins/GSdx/GLLoader.cpp#L370 with a link to https://community.amd.com/thread/205702
It would be so much easier to have working AMD driver.

lightningterror:
Hmm my idea was to add the tooltip temporary so people know they should't set blending to none.
Greg's idea works as well , I think we had a link before for the other bug as well but was removed once they released the fix.

mirh:
> This is the first i've heard of Blending Accuracy causing BSOD. I thought they'd fixed most of the bugs in the driver?

#1552 is still there. 
Besides, BSOD seems to "reside" somewhere behind blending itself, given iirc there were problems also in vulkan or something. 

> And it works fine if you use older AMD drivers.

Which are instead broken for [other reasons](https://github.com/PCSX2/pcsx2/issues/1121) ( ͡° ͜ʖ ͡°)

gregory38:
For me the issue is the setup of the GPU. They likely reserve a single shader color. But dual blending will output 2 colors. Therefore you have a buffer overflow which translates to no pilot in the air plane ;)

lightningterror:
@gregory38 Something like this?

FlatOutPS2:
Better add a short explanation about the BSOD issue too if you add a link to it. ;)

lightningterror:
Something like this?

FlatOutPS2:
Yes. But you forgot to mention the issue only affects recent driver releases. Older AMD GPU's on legacy drivers and users who just haven't updated their drivers in the past year won't have that issue.

lightningterror:
Isn't GL_ARB_separate_shader_objects broken on older drivers? If so it's a trade off.

Imo I don't think it's needed to add that it works on legacy , especially since we are expecting a fix "soon™"

FlatOutPS2:
If expecting a fix soon is a reason not to add something, we might as well not add those links at all. :/

Legacy/old driver users can use "None" Blending Unit Accuracy without issue, so there's no reason to tell them not to.

mirh:
That's why there is `legacy_fglrx_buggy_driver`

gregory38:
So is it ok to merge or not ?

FlatOutPS2:
The note should be updated:

"Note: Due to an AMD GPU OpenGL driver issue, setting Blending Unit Accuracy to "None" can cause a application or system crash.\n");
"Keep Blending Unit Accuracy set to at least the default "Basic".\n"
"AMD have a fix for the issue that will be released in the coming months. The issue does not affect AMD GPU's on legacy drivers.\n"`

mirh:
Very very nice wording there. 
Do we have something similar for legacy_fglrx?

lightningterror:
Is it a good idea to add quotation marks to Basic and None?

FlatOutPS2:
You need to change `"Basic"` to `\"Basic\"`, and the same for "None".

lightningterror:
All good ? :)

plugins/GSdx/GLLoader.cpp
```diff
@@ -367,7 +367,12 @@ namespace GLLoader {
 #ifdef _WIN32
 		if (status) {
 			if (fglrx_buggy_driver) {
-				fprintf(stderr, "OpenGL renderer is slow on AMD GPU due to inefficient driver. Sorry.\n");
+				fprintf(stderr, "OpenGL renderer is slow on AMD GPU due to inefficient driver.Please check the links below.\n"
+					"https://community.amd.com/message/2756964\n"
+					"https://community.amd.com/thread/205702\n"
+					"Note: Due to an AMD GPU OpenGL driver issue, setting Blending Unit Accuracy to \"None\" can cause a application or system crash.\n"
```
ssakash: ``a application`` **->** ``an application``

plugins/GSdx/GLLoader.cpp
```diff
@@ -367,7 +367,12 @@ namespace GLLoader {
...
+					"Keep Blending Unit Accuracy set to at least the default \"Basic\".\n"
+					"AMD have a fix for the issue that will be released in the coming months. The issue does not affect AMD GPU's on legacy drivers.\n");
```
ssakash: ``AMD have`` **->** ``AMD has``

plugins/GSdx/GLLoader.cpp
```diff
@@ -367,7 +367,12 @@ namespace GLLoader {
 #ifdef _WIN32
 		if (status) {
 			if (fglrx_buggy_driver) {
-				fprintf(stderr, "OpenGL renderer is slow on AMD GPU due to inefficient driver. Sorry.\n");
+				fprintf(stderr, "OpenGL renderer is slow on AMD GPU due to inefficient driver.Please check the links below.\n"
```
ssakash: I would prefer if ``Please check the links below`` is changed to ``Please check out the links listed below``. Also, ``OpenGL renderer`` **->** ``OpenGL hardware renderer``.

ssakash: Also a single space after the first full stop would be nice.

FlatOutPS2: Better improve the whole line:
`"The \"OpenGL (Hardware)\" renderer is slow on AMD GPU's due to an inefficient driver. Check out the links below for further information.\n"`

plugins/GSdx/GLLoader.cpp
```diff
@@ -367,7 +367,12 @@ namespace GLLoader {
 #ifdef _WIN32
 		if (status) {
 			if (fglrx_buggy_driver) {
-				fprintf(stderr, "OpenGL renderer is slow on AMD GPU due to inefficient driver. Sorry.\n");
+				fprintf(stderr, "The \"OpenGL (Hardware)\" renderer is slow on AMD GPU's due to an inefficient driver. Check out the links below for further information.\n"
```
turtleli: `GPU's` -> `GPUs`.

plugins/GSdx/GLLoader.cpp
```diff
@@ -367,7 +367,12 @@ namespace GLLoader {
...
+					"https://community.amd.com/message/2756964\n"
+					"https://community.amd.com/thread/205702\n"
+					"Note: Due to an AMD GPU OpenGL driver issue, setting Blending Unit Accuracy to \"None\" can cause an application or system crash.\n"
+					"Keep Blending Unit Accuracy set to at least the default \"Basic\".\n"
+					"AMD has a fix for the issue that will be released in the coming months. The issue does not affect AMD GPU's on legacy drivers.\n");
```
turtleli: Ditto.

gregory38:
@lightningterror could you fix the GPUs spelling ? @ssakash is it ok for you ?

lightningterror:
I already did , https://github.com/PCSX2/pcsx2/pull/1768/commits/0a7c60325aa7560648b2327005af5e9b35bbb764

plugins/GSdx/GLLoader.cpp
```diff
@@ -367,7 +367,12 @@ namespace GLLoader {
 #ifdef _WIN32
 		if (status) {
 			if (fglrx_buggy_driver) {
-				fprintf(stderr, "OpenGL renderer is slow on AMD GPU due to inefficient driver. Sorry.\n");
+				fprintf(stderr, "The \"OpenGL Hardware\" renderer is slow on AMD GPUs due to an inefficient driver. Check out the links below for further information.\n"
+					"https://community.amd.com/message/2756964\n"
+					"https://community.amd.com/thread/205702\n"
+					"Note: Due to an AMD GPU OpenGL driver issue, setting Blending Unit Accuracy to \"None\" can cause an application or system crash.\n"
+					"Keep Blending Unit Accuracy set to at least the default \"Basic\".\n"
+					"AMD has a fix for the issue that will be released in the coming months. The issue does not affect AMD GPU's on legacy drivers.\n");
```
ssakash: You missed ``GPU's`` at this line.

lightningterror: ups , didn't see the second one xD

plugins/GSdx/GLLoader.cpp
```diff
@@ -367,7 +367,12 @@ namespace GLLoader {
 #ifdef _WIN32
 		if (status) {
 			if (fglrx_buggy_driver) {
-				fprintf(stderr, "OpenGL renderer is slow on AMD GPU due to inefficient driver. Sorry.\n");
+				fprintf(stderr, "The \"OpenGL Hardware\" renderer is slow on AMD GPUs due to an inefficient driver. Check out the links below for further information.\n"
```
ssakash: ``Hardware`` **->** ``hardware``. Also I think the quotes surrounding ``OpenGL Hardware`` is unnecessary here, better to remove them?

plugins/GSdx/GLLoader.cpp
```diff
@@ -367,7 +367,12 @@ namespace GLLoader {
...
+					"https://community.amd.com/message/2756964\n"
+					"https://community.amd.com/thread/205702\n"
+					"Note: Due to an AMD GPU OpenGL driver issue, setting Blending Unit Accuracy to \"None\" can cause an application or system crash.\n"
+					"Keep Blending Unit Accuracy set to at least the default \"Basic\".\n"
```
ssakash: ``to at least the default \"Basic\"`` **->** ``to at least the default \"Basic\" level.``

plugins/GSdx/GLLoader.cpp
```diff
@@ -367,7 +367,12 @@ namespace GLLoader {
 #ifdef _WIN32
 		if (status) {
 			if (fglrx_buggy_driver) {
-				fprintf(stderr, "OpenGL renderer is slow on AMD GPU due to inefficient driver. Sorry.\n");
+				fprintf(stderr, "The \"OpenGL Hardware\" renderer is slow on AMD GPUs due to an inefficient driver. Check out the links below for further information.\n"
+					"https://community.amd.com/message/2756964\n"
+					"https://community.amd.com/thread/205702\n"
+					"Note: Due to an AMD GPU OpenGL driver issue, setting Blending Unit Accuracy to \"None\" can cause an application or system crash.\n"
```
ssakash: ``Due to an AMD GPU OpenGL driver issue`` could probably be simplified to ``Due to an AMD driver issue``, any opinion?

FlatOutPS2: Better not, because it's an OpenGL driver issue only, and we don't need to confuse users. Removing GPU should be fine though.

lightningterror:
I guess now it can be merged.

gregory38:
Ok squashed and merged. Next time, it would be nice to squash it on your side. This way
* you write your own message
* I don't fight with the 2 line text editor of github (seriously github 2 lines)

PR #1770, opened 2017-01-13
Cleanup GCC warnings - still needs some work

np511:
Cleaned up some of the GCC warnings. Not done yet, but a start.


np511:
Just noticed - because I don't compile all the plugins, I hadn't noticed the problems - should be done in ~10 minutes.

gregory38:
Honestly I'm not sure it worth to fix extra plugins. There are quite deprecated.

gregory38:
And please split your commit.

np511:
Split?

gregory38:
As cut is small commit.
You currently have a single commit that contains all the changes. It would be better to have a commit by class of issue. It allows easy review and easy regression test.

np511:
I'll split it momentarily, just fixing one file first.

np511:
Quick question: how do I undo the commit? I accidentally ran `git commit --amend` and now the last merge is showing up and I don't know how to get rid of it.

gregory38:
You need to do a rebase `git rebase master` to avoid the merge commit.

gregory38:
And here my aliases that could be helpful (but don't uncommit the merge commit)
```
uncommit = reset --soft HEAD~1
unstage = reset HEAD --
```

gregory38:
@np511 most code of the project (not yet all) is formatted with clang-format 3.8. Hence the build failure with clang (which check format + compilation). Cdvd giga + all null plugins must be reformated properly.

np511:
I know - I had it fixed before the mistake when I was splitting, I'll fix it later.

I would have it done, but I just switched to Xubuntu, and it took some time to get the 32-bit libs working without screwing up the system.

gregory38:
Ok no worry. Take your time.

pcsx2/Gif.cpp
```diff
@@ -486,7 +486,6 @@ static u16 QWCinGIFMFIFO(u32 DrainADDR)
 		gifstate |= GIF_STATE_EMPTY;
 
 	GIF_LOG("%x Available of the %x requested", ret, gifch.qwc);
-	pxAssert(ret >= 0);
```
gregory38: @refractionpcsx2 could you double check the function behaves correctly. So we're sure that the assertion is indeed useless.

np511: I was a bit wary of removing it - I didn't see why it was added, and GCC flagged it as always true, but I didn't think it would be added without a reason. I figured it would be simpler to add it now, and rebase later if needed.

np511:
Finished splitting them - I'm going to build it again here and see how many I can remove in the next ~30 minutes before I have *real* work to do (unfortunately).

pcsx2/Gif.cpp
gregory38: It is fine to remove it. However this warning is due to recent Ref change. I want to be sure that it is expected.

refractionpcsx2: Yes it works :) Already been testing it recently

np511: If it *didn't* work, it would be indicative of something wrong with the compiler.

gregory38:
```
I'm going to build it again here and see how many I can remove in the next ~30 minutes
```
??? remove of what ?

np511:
There are still some GCC warnings - I'm trying to remove as many as I can with my limited understanding of the emulator (if I don't know why it's there, I'm hesitant to change it before taking a more thorough look at the code so I know what I'm doing).

gregory38:
Oh. Just to be sure only fix warning in normal (not extra) plugins.

np511:
I need to take a look at GDK  - `/home/noam/Documents/dev/emulators/pcsx2/plugins/GSdx/GSLinuxDialog.cpp:503:27: warning: ‘GdkPixbuf* gdk_pixbuf_from_pixdata(const GdkPixdata*, gboolean, GError**)’ is deprecated [-Wdeprecated-declarations]` - might take some time. I've worked with it in the past, but not to a large extent.

np511:
The only "extra" plugin I'm building right now is the GSdx Legacy plugin.

np511:
Just to be clear, what is the reason for this:

```
/home/noam/Documents/dev/emulators/pcsx2/plugins/GSdx/GSRendererSW.cpp:1115:57: warning: left shift of negative value [-Wshift-negative-value]
      gd.l = GSVector4((float)(-0x10000 << context->TEX1.L));
```
There are other instances, but is there any specific reason for this?

np511:
I'm almost out of time - I fixed the GDK deprecated method warning (I think) but I'm going to test it first.

gregory38:
```
      gd.l = GSVector4((float)(-0x10000 << context->TEX1.L));
```
So basically the formula is `LOD = log2(1/|Q|) << L + K` when LCM is 0. So basically `LOD = gd.l * log2(|Q|) + K`. `TEX1.L` is 2 bits so between 0 and 3. Note the math uses fixed integer (hence 0x10000 instead of 1). gd.l is always negative (to optimize the reciprocal operation)

np511:
Somewhat complex explanation (probably my brain saying, "OW! You're hurting me! I need sleep!" and my responding "Screw you, brain!" isn't helping), I'll look at the code later and see if I can figure it out.

For now, working on GDK.

np511:
I'm done for the day. If I try to force myself to keep going, I'll lose interest. Right now, I'm going to start another build and remember to log it this time so I can use grep on a file instead of a pipe.

gregory38:
Normally spu2null and gsnull are not built. So I'm not sure how you got warnings on them.
If you use build.sh it will create a build log file.

plugins/GSdx/GSLinuxLogo.h
```diff
@@ -1,1617 +1,2257 @@
...
- *  GNU General Public License for more details.
- *
- *  You should have received a copy of the GNU General Public License
- *  along with GNU Make; see the file COPYING.  If not, write to
- *  the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA USA.
- *  http://www.gnu.org/copyleft/gpl.html
- *
- */
-
 /* GdkPixbuf RGB C-Source image dump 1-byte-run-length-encoded */
```
gregory38: Add a copyright header. And explain how you generate the file, please.

plugins/GSdx/GSLinuxDialog.cpp
```diff
@@ -500,8 +500,9 @@ bool RunLinuxDialog()
 	GtkWidget* osd_box      = gtk_vbox_new(false, 5);
 
 	// Grab a logo, to make things look nice.
-	GdkPixbuf* logo_pixmap = gsdx_ogl_logo;
-	GtkWidget* logo_image  = gtk_image_new_from_pixbuf(logo_pixmap);
+	GResource * resources = GSdx_res_get_resource();
+	GdkPixbuf * ogl_logo = gdk_pixbuf_new_from_stream(g_resource_open_stream(resources,"/GSdx/res/logo-ogl.bmp",G_RESOURCE_LOOKUP_FLAGS_NONE,NULL),NULL,NULL);
+	GtkWidget* logo_image  = gtk_image_new_from_pixbuf(ogl_logo);
```
gregory38: One question. Does it requires to install a bmp file ? The old way (even if deprecated) allow to include it in the binary.

plugins/GSnull/GS.cpp
```diff
@@ -66,6 +65,7 @@ EXPORT_C_(const char *)
 PS2EgetLibName()
 {
 #ifdef _MSC_VER
+    char libraryName[64];
```
gregory38: It is legal to return a pointer of a temporary object on the stack ?

willkuer:
Does the `char*` to `const char*` transition in the plugin interface break plugin backwards compatibility?

gregory38:
I think I already did one. I think it will be fine because we export C symbol (i.e. not mangled) and we lookup the function name manually/directly. So potentially ABI will be updated but it will remain compatible.

np511:
On Tuesday, January 17, 2017 3:09:11 AM EST willkuer wrote:

> Does the `char*` to `const char*` transition in the plugin interface break

> plugin backwards compatibility?

Technically, the value returned is exactly the same - a pointer to a character. Only the compilation should be affected, as the compiler enforces `const` during compilation time, not run time.

np511:
On Tuesday, January 17, 2017 1:36:02 AM EST Gregory Hainaut wrote:

> gregory38 commented on this pull request.

>

> > @@ -66,6 +65,7 @@ EXPORT_C_(const char *)

>

> PS2EgetLibName()

> {

> #ifdef _MSC_VER

> + char libraryName[64];

>

> It is legal to return a pointer of a temporary object on the stack ?

I really wasn't as careful as I should have been. I'll take a look in a moment.

gregory38:
```
Technically, the value returned is exactly the same - a pointer to a character. Only the compilation should be affected, as the compiler enforces `const` during compilation time, not run time.
```
And the link as you could have both `void foo(const char*)` and `void foo(char*)`

np511:
On Tuesday, January 17, 2017 1:31:10 AM EST Gregory Hainaut wrote:

> gregory38 commented on this pull request.

>

> > - * it under the terms of the GNU General Public License as published by

>

> - * the Free Software Foundation; either version 2, or (at your option)

> - * any later version.

> - *

> - * This Program is distributed in the hope that it will be useful,

> - * but WITHOUT ANY WARRANTY; without even the implied warranty of

> - * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the

> - * GNU General Public License for more details.

> - *

> - * You should have received a copy of the GNU General Public License

> - * along with GNU Make; see the file COPYING. If not, write to

> - * the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,

> Boston, MA 02110-1301, USA USA. - * http://www.gnu.org/copyleft/gpl.html

> - *

> - */

> -

> /* GdkPixbuf RGB C-Source image dump 1-byte-run-length-encoded */

>

> Add a copyright header. And explain how you generate the file, please.

I ran `rm GSLinuxLogo.h`. The whole file is gone, not just the header.



I modified CMakeLists.txt in the GSdx folder to automatically generate two files before compiling - GSdxResources.c and GSdxResources.h. These are generated from a file in the GSdx/res folder called gsdx-res.xml. GSdxResources.c contains the bundle and the functions to load it, and GSdxResources.h contains the header for the function. They aren't included with the commit because they are automatically generated.



In short, no, the bmp doesn't need to be included separately.

gregory38:
Ok. Yes my comment was on older commit.

np511:
On Tuesday, January 17, 2017 1:34:39 AM EST Gregory Hainaut wrote:

> gregory38 commented on this pull request.

>

> > @@ -500,8 +500,9 @@ bool RunLinuxDialog()

>

> GtkWidget* osd_box = gtk_vbox_new(false, 5);

>

> // Grab a logo, to make things look nice.

> - GdkPixbuf* logo_pixmap = gsdx_ogl_logo;

> - GtkWidget* logo_image = gtk_image_new_from_pixbuf(logo_pixmap);

> + GResource * resources = GSdx_res_get_resource();

> + GdkPixbuf * ogl_logo =

> gdk_pixbuf_new_from_stream(g_resource_open_stream(resources,"/GSdx/res/logo

> -ogl.bmp",G_RESOURCE_LOOKUP_FLAGS_NONE,NULL),NULL,NULL); + GtkWidget*

> logo_image = gtk_image_new_from_pixbuf(ogl_logo);

>

> One question. Does it requires to install a bmp file ? The old way (even if

> deprecated) allow to include it in the binary.

No. GSdx_res_get_resource() returns a pointer to a resource bundle in memory containing the bmp. g_resource_open_stream(resource,"/GSdx/res/logo-ogl.nmp"...) opens a GInputStream to the location of the bmp data in memory. gdk_pixbuf_new_from_stream creates the pixbuf instead of gdx_pixbuf_new_from_inline, but the rest isn't too different.

np511:
On Monday, January 16, 2017 10:33:24 PM EST Gregory Hainaut wrote:

> Normally spu2null and gsnull are not built. So I'm not sure how you got

> warnings on them. If you use build.sh it will create a build log file.

I'd already built with the extra plugins, and while I won't go out of my way to fix them, something so simple and quick that I already knew about didn't seem like a problem.



Question: if the goal is to remove all warnings, what should be done about the bit-shifting of negative numbers?

plugins/GSnull/GS.cpp
```diff
@@ -61,11 +61,14 @@ PS2EgetLibType()
     return PS2E_LT_GS;
 }
 
+#ifdef _MSC_VER
+char libraryName[64];
```
gregory38: or as static in the function could work too ? What do you think ?

gregory38:
```
I'd already built with the extra plugins, and while I won't go out of my way to fix them, something so simple and quick that I already knew about didn't seem like a problem. 
```
Sure. Just saying it doesn't worth it.

```
Question: if the goal is to remove all warnings, what should be done about the bit-shifting of negative numbers?
```
All is an idea. It depends how many warning remain. Standard need to be read too because behavior of C++/C++11/C++14 aren't always the same.

gregory38:
Note: AFAIU the behavior of `-1 << N` depends on the signed representation. It isn't a big deal because PCSX2 only runs on arch that uses standard 2 complement.

turtleli:
The standard states that left shifting a negative number is undefined behaviour. The fix is usually to cast the negative number to an unsigned type, and cast the shift result back to a signed type.

gregory38:
I have the feeling the various casts just hide the warning ;)

np511:
If porting to other architechtures is intended (even in the very distant future - say, after PCSX3 - so two decades, give or take a century), hiding the warning might be a bad idea.



Should I use the cast solution mentioned by @turtleli, leave the warning, or something else? Being rather new around here, I realize this decision is above my metaphorical pay grade,

turtleli:
> I have the feeling the various casts just hide the warning ;)

Well, `-0x10000 << context->TEX1.L` has undefined behaviour, whereas `int32(uint32(-0x10000) << context->TEX1.L)`  has implementation defined behaviour due to the unsigned to signed cast. ;)

gregory38:
@np511 
do you plan to fix more warnings ? Or could we review the commits (squash some) and merge them ?

np511:
> do you plan to fix more warnings ? Or could we review the commits (squash
> some) and merge them ?
I do plan to fix some more. I was having some problems with my installed PCSX2 
and wanted to make sure it wasn't because of this. Apparently, using the build 
script with "--extra" and then without the "--extra" flag without running a 
clean build can lead to weird results.

np511:
On Sunday, January 22, 2017 3:10:58 AM EST Gregory Hainaut wrote:

> @np511

> do you plan to fix more warnings ? Or could we review the commits (squash

> some) and merge them ?

Done now. There are a few more, but none I can fix. Most are shifts of negative values, one involves the GifTag struct and a potentially unintialized value, and one is a signed/unsigned value comparison.

gregory38:
Ok. It would be nice to
* squash commits that fix the same warning. Typically when you when you tried it 2/3 times.
* add a small tag on the commit message (such as `gsnull:` or `gsdx:`...)

np511:
On Tuesday, January 24, 2017 2:37:54 AM EST Gregory Hainaut wrote:

> Ok. It would be nice to

> * squash commits that fix the same warning. Typically when you when you

> tried it 2/3 times. * add a small tag on the commit message (such as

> `gsnull:` or `gsdx:`...)

I don't have time right now, I'll have this done within two hours at most though (now that I'm more familiar with git).

gregory38:
Don't worry. I can understand that you have a real life ;) Just ping me when it is done.

gregory38:
@np511 
I'm currently trying to understand a bit the glib resource loading. I think it could potentially be used for `*.glsl` files and `tfx.cl`. I don't know if it will fit well with `GSdxApp::LoadResource`. Anyway, as far as I understand there are 2 APIS:
* a binary bundle (.gresources file)
* Or a `.c` + `.h` file (as you've done)

why did you chose the 2nd ?

plugins/GSdx/GSLinuxDialog.cpp
```diff
@@ -500,8 +500,9 @@ bool RunLinuxDialog()
 	GtkWidget* osd_box      = gtk_vbox_new(false, 5);
 
 	// Grab a logo, to make things look nice.
-	GdkPixbuf* logo_pixmap = gsdx_ogl_logo;
-	GtkWidget* logo_image  = gtk_image_new_from_pixbuf(logo_pixmap);
+	GResource * resources = GSdx_res_get_resource();
+	GdkPixbuf * ogl_logo = gdk_pixbuf_new_from_stream(g_resource_open_stream(resources,"/GSdx/res/logo-ogl.bmp",G_RESOURCE_LOOKUP_FLAGS_NONE,NULL),NULL,NULL);
```
gregory38: There is a potential memory leak.

https://developer.gnome.org/gio/stable/GResource.html#g-resource-open-stream
```
Returns

GInputStream or NULL on error. Free the returned object with g_object_unref(). 
```

np511:
I chose the second because it compiled directly into the program - the .gresources method would require far more work and install an additional file.

np511:
On Tuesday, January 24, 2017 12:01:50 PM EST Gregory Hainaut wrote:

> gregory38 commented on this pull request.

>

> > @@ -500,8 +500,9 @@ bool RunLinuxDialog()

>

> GtkWidget* osd_box = gtk_vbox_new(false, 5);

>

> // Grab a logo, to make things look nice.

> - GdkPixbuf* logo_pixmap = gsdx_ogl_logo;

> - GtkWidget* logo_image = gtk_image_new_from_pixbuf(logo_pixmap);

> + GResource * resources = GSdx_res_get_resource();

> + GdkPixbuf * ogl_logo =

> gdk_pixbuf_new_from_stream(g_resource_open_stream(resources,"/GSdx/res/logo

> -ogl.bmp",G_RESOURCE_LOOKUP_FLAGS_NONE,NULL),NULL,NULL);

>

> There is a potential memory leak.

>

> https://developer.gnome.org/gio/stable/GResource.html#g-resource-open-stream

> ```

> Returns

>

> GInputStream or NULL on error. Free the returned object with

> g_object_unref(). ```

Nice catch - I'll fix it immediately.

gregory38:
My understanding was gresource is a binary section added to ELF. Anyway I'm fine, I trust you.

By squashing, I didn't mean to squash everything. But to squash a commit that fix a previous commit.

np511:
On Tuesday, January 24, 2017 12:35:07 PM EST Gregory Hainaut wrote:

> My understanding was gresource is a binary section added to ELF. Anyway I'm

> fine, I trust you.

>

You can then use glib-compile-resources to compile the XML to a binary bundle that you can load with g_resource_load()<https://developer.gnome.org/gio/unstable/GResource.html#g-resource-load>. However, its more common to use the --generate-source and --generate-header arguments to create a source file and header to link directly into your application. This will generate get_resource(), register_resource() and unregister_resource() functions, prefixed by the --c-name argument passed to glib-compile-resources. get_resource() returns the generated GResource<https://developer.gnome.org/gio/unstable/GResource.html> object. The register and unregister functions register the resource so its files can be accessed using g_resources_lookup_data()<https://developer.gnome.org/gio/unstable/GResource.html#g-resources-lookup-data>.

> By squashing, I didn't mean to squash everything. But to squash a commit

> that fix a previous commit.

... Oops? Not sure if I can undo it.

gregory38:
You can split the commit (then you will become a git expert ;) )

gregory38:
(if it can help) In your branch
```
git uncommit => git reset --soft HEAD~1
git unstage => reset HEAD --
```
then loop to select the part that you want to commit
```
git add -p
git commit
```
When done,
```
git push -f
```

np511:
On Wednesday, January 25, 2017 7:35:35 AM EST Gregory Hainaut wrote:
> (if it can help) In your branch
> ```
> git uncommit => git reset --soft HEAD~1
> git unstage => reset HEAD --
> ```
> then loop to select the part that you want to commit
> ```
> git add -p
> git commit
> ```
> When done,
> ```
> git push -f
> ```

What is unstage? How is it different from uncommit?

gregory38:
You have several levels in git.
* uncommitted, unstaged => basically you updated a file in your favorite editor
* uncommited, staged => you did git add -p (note -p is interactive selection of your change)
* commited => after git commit, it will commit the updates that are staged. (unstaged update remains uncommit, unstaged)

The above command are alias because git syntax is awful. So unstage/uncommit is easier to understand for me. Here a book that can be useful https://git-scm.com/book/en/v2

gregory38:
Once it is merged (i.e. after you split the commits ;) ), I have a job for you (if you want). I don't know if you have a Win env to test the alternate OS. Anyway, here the improvement

Current situation:
* windows can load resources with this code https://github.com/PCSX2/pcsx2/blob/master/plugins/GSdx/GSdx.cpp#L46
* For openGL I run a PERL script https://github.com/PCSX2/pcsx2/blob/master/linux_various/glsl2h.pl to convert the shader file to an include file. Awful, but very efficient.

Future:
* Implement the load resources in GSdx.cpp. Honestly might be better to change the interface, to generate a std::string instead of a vector. Or maybe create a load text resources function.
* Then udpate the PERL script to avoid the generation of shader that are automatically loaded. (Note: don't remove the script as it is useful for the legacy zzogl plugin)
* Perf wise, all glsl files are read once except TFX related shader (if we could avoid a huge memcpy every time we compile the shader it would be nice)
* Note: potentially it would be enough to support openCL on Linux ;)

What do you think?

np511:
On Friday, January 27, 2017 2:24:43 AM EST Gregory Hainaut wrote:
> Once it is merged (i.e. after you split the commits ;) ), I have a job for
> you (if you want). I don't know if you have a Win env to test the alternate
> OS. Anyway, here the improvement
> 
> Current situation:
> * windows can load resources with this code
> https://github.com/PCSX2/pcsx2/blob/master/plugins/GSdx/GSdx.cpp#L46 * For
> openGL I run a PERL script
> https://github.com/PCSX2/pcsx2/blob/master/linux_various/glsl2h.pl to
> convert the shader file to an include file. Awful, but very efficient.
> 
> Future:
> * Implement the load resources in GSdx.cpp. Honestly might be better to
> change the interface, to generate a std::string instead of a vector. Or
> maybe create a load text resources function. * Then udpate the PERL script
> to avoid the generation of shader that are automatically loaded. (Note:
> don't remove the script as it is useful for the legacy zzogl plugin) * Perf
> wise, all glsl files are read once except TFX related shader (if we could
> avoid a huge memcpy every time we compile the shader it would be nice) *
> Note: potentially it would be enough to support openCL on Linux ;)
> 
> What do you think?
I think I'm going to start splitting the comments immediately.

I've done some reading on OpenCL (not much, mainly when I was having problems 
with OpenGL on my old computer and found OpenCL by accident), but don't know 
much about it. Why is OpenCL important?

I finally crashed from my insomnia, which prevented me from finishing this 
sooner.

gregory38:
The main goal is to avoid a Perl script to convert the shader.Resource seems better, with a compression it could reduce the binary size.
OpenCL is a nice to have. So I can play with it on Linux.

np511:
On Friday, January 27, 2017 11:43:40 PM EST Gregory Hainaut wrote:

> The main goal is to avoid a Perl script to convert the shader.Resource seems

> better, with a compression it could reduce the binary size. OpenCL is a

> nice to have. So I can play with it on Linux.

My question: what is OpenCL? I don't have time to research it.



I'm splitting the commit now, should be done tomorrow at the latest.

np511:
Finished with splitting. Thanks for your explanations on uncommit/unstage - they helped a lot.



I really need to get some sleep. Only did this now because I knew if I didn't I wouldn't get it done for probably close to a week.



Now that I know how to use the GResources API, I should be able to add it in easily enough. There's built-in compression support (I think, I'll have to double-check later), though when running, memory would be used for both the compressed and the uncompressed versions, but it shouldn't be a big deal.



I think the GL threading optimizations for mesa3d currently only supports gallium-based drivers (even in the fork). I'm pretty sure I did it right with no visible effects. I had to install mesa-utils:i386 into a local folder to test this, but I'm pretty sure I did it right.

gregory38:
I'm glad that my explanation was useful. 

OpenCL is yet another API to access the GPU coprocessor. The interesting part is that renderer is implemented as a SW renderer on the GPU. So it doesn't use the fixed HW unit (perf--), it doesn't support upscaling neither. However it doesn't bother with a texture cache (kind of HSA/HMM/unified memory). I would like to study it. Unfortunately I never got the time. Anyway it would be nice to enable it again.

On the resource topic, it depends how you implement the loading. It might already require an extra copy on vector/string. So RAM impact will be null.

np511:
> OpenCL is yet another API to access the GPU coprocessor. The interesting
> part is that renderer is implemented as a SW renderer on the GPU. So it
> doesn't use the fixed HW unit (perf--), it doesn't support upscaling
> neither. However it doesn't bother with a texture cache (kind of
> HSA/HMM/unified memory). I would like to study it. Unfortunately I never
> got the time. Anyway it would be nice to enable it again.
> 
> On the resource topic, it depends how you implement the loading. It might
> already require an extra copy on vector/string. So RAM impact will be null.
Actually, if it already requires an extra copy, the RAM impact will exist - it 
will be good, as there will be one compressed copy + one uncompressed copy 
instead of two uncompressed copies.

Thanks for the info on OpenCL. I would definitely like to see more renderer 
options on Linux.

Is this good now? No more work needed? I'll start looking into doing what you 
said (brain hurting now. I've been teaching myself x86 (and x86_64) assembly 
and my head is hurting from a bug in the ASM calculator I spontaneously 
decided to write. I don't mean that the computer bug is physically affecting - 
you know what I mean. I'm going to stop rambling now, and will most likely 
look at this reply later and palmface, but that's nothing new.

gregory38:
Stupid github, I only see the discussion on my phone.

Static, reduce the scope to a file. And it will be even better to put the variable in the function directly (if it used only the function). 64 is likely fine, but don't be cheap, if someone change it one day, it reduces the probability of an overflow (if the string function aren't safe which is bad).

Anyway, the best is to fix your commit in a separate commit. Then you can do a `git rebase -i master` (the `-i` stand for interactive). A text editor will open, and you can change the order of commits (1 commit == 1 line). You can also squash (use `s` instead of `pickup`) commit into the previous one. If you squash commit `a` and commit `b`, it will fuse both commits into a single commit `c`. You can then update the commit message accordingly. Note: if you use `r` on a commit, you can update the commit message.

np511:
On Sunday, January 29, 2017 6:11:59 AM EST Gregory Hainaut wrote:

> Stupid github, I only see the discussion on my phone.

>

> Static, reduce the scope to a file. And it will be even better to put the

> variable in the function directly (if it used only the function). 64 is

> likely fine, but don't be cheap, if someone change it one day, it reduces

> the probability of an overflow (if the string function aren't safe which is

> bad).

>

> Anyway, the best is to fix your commit in a separate commit. Then you can do

> a `git rebase -i master` (the `-i` stand for interactive). A text editor

> will open, and you can change the order of commits (1 commit == 1 line).

> You can also squash (use `s` instead of `pickup`) commit into the previous

> one. If you squash commit `a` and commit `b`, it will fuse both commits

> into a single commit `c`. You can then update the commit message

> accordingly. Note: if you use `r` on a commit, you can update the commit

> message.

I found an easier way - I edited the file (without running any git commands) with mcedit (the Midnight Commander editor - super useful for Linux programming). After looking through ```git log```, I found the last commit before the one I wanted to edit, and ran ```git rebase -i commitid``` Went to the line "pick... Remove Warnings..." and changed "pick" to "edit". A simple "git commit --amend", save file, "git rebase --continue", and done.



Could have been done better, but I'm not very good with rebasing.

gregory38:
Yes that was the 2nd method. Note, you can just use `git rebase -i master`

gregory38:
I'm fine with the code but I need someone to double check the compilation of old plugins on Windows.
@turtleli @refractionpcsx2 could you try a compilation of the old_project (the one that included all deprecated plugins).

np511:
On Sunday, January 29, 2017 6:26:01 AM EST Gregory Hainaut wrote:
> Yes that was the 2nd method. Note, you can just use `git rebase -i master`
I figured that out - you mentioned it in your other post. So - done?

np511:
Is there a problem if I do the work on the `master` branch? I don't think I 
need to create a branch per pull request.

gregory38:
Yes everything will get messy. 1 branch == 1 PR. Otherwise if you add more commits they will be included in this PR too. We don't want to do that.

np511:
On Sunday, January 29, 2017 9:54:21 AM EST Gregory Hainaut wrote:
> Yes everything will get messy. 1 branch == 1 PR. Otherwise if you add more
> commits they will be included in this PR too. We don't want to do that.
OK, got it

turtleli:
> could you try a compilation of the old_project (the one that included all deprecated plugins).

I don't have the CG toolkit installed (and don't want to install it) so I can't actually successfully compile zzogl on Windows. There's a new "conflicting definitions" error though. The following should fix it (but I can't confirm it will do so):
```
diff --git a/plugins/zzogl-pg/opengl/Util.h b/plugins/zzogl-pg/opengl/Util.h
index 27fb53765..a62348235 100644
--- a/plugins/zzogl-pg/opengl/Util.h
+++ b/plugins/zzogl-pg/opengl/Util.h
@@ -64,13 +64,6 @@
 #include "CRC.h"
 #include "ZZLog.h"
 
-#ifdef _WIN32
-// need C definitions -- no mangling please!
-extern "C" u32   CALLBACK PS2EgetLibType(void);
-extern "C" u32   CALLBACK PS2EgetLibVersion2(u32 type);
-extern "C" char* CALLBACK PS2EgetLibName(void);
-#endif
-
 // Allow easy copy/past between GSdx and zzogl
 typedef unsigned char uint8;
 typedef signed char int8;
```
All the other old plugins compile fine.

gregory38:
ok good.

Issue #444, opened 2015-01-31
[Feature Request] - Display of internal(in-game) FPS


Sarania:
I think this would be very nice to have. Currently we only have the GSdx FPS indicator which of course is not true in game FPS. But there should be a relatively easy way to count the internal FPS too, I think. It would be useful in several ways I can think of. One is helping novice users to understand the effect of speedhacks better and how their game can feel slow even at 60 FPS. Another is testing the effects of the proposed EE overclock.

ramapcsx2:
I'll second the request.
Maybe one can get the (VFPS, is that what other emus call this?) by measuring time between hardware Vsync events? Dunno.

Sarania:
As I understand it, VFPS is what we have now, and what we are requesting here is RFPS.

For instance in nullDC, 60VFPS is full speed for NTSC, but RFPS can vary internally(like e.g. its 30 in Grandia II).

And yeah, measuring vsync events would be one way to do it, assuming all games on PS2 use vsync.

Asmodean-:
I was going to have a bash at this last night. Untill I realised that I can't even compile the latest gsdx, because of the opencl implementation ;p anyone recommend an opencl sdk?. I briefly tried the intel one, but it didn't work.

Sarania:
I have the AMD one and GSdx compiles fine for me.

karasuhebi:
@ramapcsx2 Dolphin calls it VPS. According to one of their developers (@RachelBryk), VPS standard for 'vertical interrupts per second'. So as long as the emulator is running at 100% speed, the VPS reading should be 60Hz. From what I understand then, the FPS reading PCSX2 uses would be considered VPS.

https://forums.dolphin-emu.org/Thread-what-does-vps-stand-for

Sarania:
So then you are saying my description is right... VPS is what we have currently, right?

karasuhebi:
Yeah that's what I'm saying.

RachelBryk:
Yes, that's right. Both measurements are useful to have.

gregory38:
I think we currently using the internal vsync counter (ps2 hw) to measure fps. What others time do you want?

Sarania:
@gregory38 That can't be - else that means every single PS2 game runs at 60(or 50) FPS which we know is not true(e.g. many games run at 30). Else I'm misunderstanding you.

Either way what we want to display is the actual in game FPS itself. Like how some games run at 60, some at 30, some are variable and have drops, etc.

[...]

PR #1771, opened 2017-01-14
GSdx: Improve captured screenshots naming

### LINKED ISSUE #1700: [Improvement] Faster sequential snapshots
***PCSX2 version:***
pcsx2-v1.5.0-dev-1711-g29d2292-windows-x86

I like to take a lot of screenshots of games, usually sequential snaps of a certain animation or cutscene. Problem is no matter how fast I press the snapshot key, there are a few seconds worth of delay between each generated screenshot. Meaning, that I can never get a quick sequential screenshot result.
I would be great if that delay was removed, so that if I press F8 four times in one second, I would get four screenshots of that specific time frame.

By the way, all settings are at default. As in, native resolution, no effects, no AA, etc.

Just press F8 a bunch of times. 

***PC specifications:***
``(For example: CPU: Intel core i5 3470 GPU: none OS: Windows 7 64 Bit)``

ssakash:
**Summary of changes**:

* Prevent a conflict where it wasn't possible to capture multiple screenshots at a single second. (Refer commit message for further details)

Fixes #1700

### DIFF
--- a/plugins/GSdx/GSRenderer.cpp
+++ b/plugins/GSdx/GSRenderer.cpp
@@ -556,13 +556,27 @@ bool GSRenderer::MakeSnapshot(const string& path)
 {
 	if(m_snapshot.empty())
 	{
-		time_t t = time(NULL);
+		time_t cur_time = time(nullptr);
+		static time_t prev_snap;
+		// The variable 'n' is used for labelling the screenshots when multiple screenshots are taken in
+		// a single second, we'll start using this variable for naming when a second screenshot request is detected
+		// at the same time as the first one. Hence, we're initially setting this counter to 2 to imply that
+		// the captured image is the 2nd image captured at this specific time.
+		static int n = 2;
+		char local_time[16];
 
-		char buff[16];
-
-		if(strftime(buff, sizeof(buff), "%Y%m%d%H%M%S", localtime(&t)))
+		if (strftime(local_time, sizeof(local_time), "%Y%m%d%H%M%S", localtime(&cur_time)))
 		{
-			m_snapshot = format("%s_%s", path.c_str(), buff);
+			if (cur_time == prev_snap)
+			{
+				m_snapshot = format("%s_%s_(%d)", path.c_str(), local_time, n++);
+			}
+			else
+			{
+				n = 2;
+				m_snapshot = format("%s_%s", path.c_str(), local_time);
+			}
+			prev_snap = cur_time;
 		}
 	}
 


FlatOutPS2:
Woulnd't it be better/simpler to just add miliseconds?

turtleli:
What's the actual goal? Multiple snapshots per second, or frame by frame snapshots?

ssakash:
> Woulnd't it be better/simpler to just add milliseconds?

I'm not sure if adding milliseconds is the optimal approach, it would give you 3 extra digits at each captured file's name which seems a bit large instead of the current approach of adding just a single number alongside the file's name.

I could get the milliseconds, but I only know how to do it in WIN32 API, I'm not aware of any cross platform alternatives. Feel free to suggest one if you have any in your mind. Also do note that the current function for getting string of the local time (``strftime()``) is closely linked with the naming of the captured file, I'd prefer to not replace it with any other library function to save some trouble.

> What's the actual goal? Multiple snapshots per second, or frame by frame snapshots?

The goal is to capture multiple snapshots per second, I'm not sure what you mean by ``frame by frame snapshots``? Sounds confusing to me on what it actually means.

turtleli:
Frame by frame snapshots - consecutive snapshots without missing frames?

mirh:
Wouldn't that require "defining" what a *frame* is in the first place, which was one of the great problems in #444 ?

turtleli:
No? I'm thinking of video capture but where the output is to separate image files without the resizing that the GS video recording does.

plugins/GSdx/GSRenderer.cpp
```diff
@@ -552,17 +552,41 @@ void GSRenderer::VSync(int field)
 	}
 }
 
+static bool IsTimeEqual(tm t1, tm t2)
```
danilaml: any reason for pass by value?

ssakash: No specific reasons. Why should I pass by reference instead?

ssakash:
> Frame by frame snapshots - consecutive snapshots without missing frames?

That would probably produce over 25-30 image captures per second, not sure if it would be feasible even with lowest compression factor. Also from my opinion, I don't think frame by frame capture would be useful. Even if you're making a GIF, merging 30+ images is quite time consuming, capturing something like 4 frames per second would be rather enough at most cases for making a decently smooth GIF. (Or, I guess so?)

FlatOutPS2:
> I'm not sure if adding milliseconds is the optimal approach, it would give you 3 extra digits at each captured file's name which seems a bit large instead of the current approach of adding just a single number alongside the file's name.

True, but you are adding a space and (), so you're adding more characters anyway. :P

I do suggest changing it from `currentime (number)` to 'currenttime-number'. This way the sequential number snapsshots will added after the first entry in the snaps folder if the files are sorted by name or date.

And wouldn't a solution like this be simpler:

```
		static char prev_buff[16];
		static int n = 2;
		char buff[16];

		if(strftime(buff, sizeof(buff), "%Y%m%d%H%M%S", localtime(&t)))
		{
			if (!memcmp(buff, prev_buff, 16)) {
				m_snapshot = format("%s_%s (%d)", path.c_str(), buff, n++);
			}
			else {
				n = 2;
				m_snapshot = format("%s_%s", path.c_str(), buff);
			}
			memcpy(prev_buff, buff, sizeof(buff));
		}
	}
```

plugins/GSdx/GSRenderer.cpp
danilaml: Well it kinda big (9 ints).

ssakash: I could remove this useless bool function if I follow [Flatout's idea](https://github.com/PCSX2/pcsx2/pull/1771#issuecomment-272634750). So yeah, I'll be doing that. ;)

ssakash:
> I do suggest changing it from currentime (number) to 'currenttime-number'. This way the sequential number snapsshots will added after the first entry in the snaps folder if the files are sorted by name or date.

Sounds reasonable enough, I'll be using ``-number`` instead of ``(number)`` then. However, the latter seems more nice to be honest for the reasons I stated [earlier](https://github.com/PCSX2/pcsx2/issues/1700#issuecomment-272625104).

_Quoting my post_,
> it would be better to just add a (2),(3) next to the image's name to indicate that multiple images have been captured in a single second. **(Yeah, following the same format which Windows does when it wants to add a new file which has a similar name of an old file present in the current directory)** 

> And wouldn't a solution like this be simpler:

I would prefer to avoid the K&R coding style and the mem functions. :P

turtleli:
> That would probably produce over 25-30 image captures per second, not sure if it would be feasible even with lowest compression factor. Also from my opinion, I don't think frame by frame capture would be useful. Even if you're making a GIF, merging 30+ images is quite time consuming, capturing something like 4 frames per second would be rather enough at most cases for making a decently smooth GIF. (Or, I guess so?)

Well, the idea is that you get to choose the images you want afterwards instead of mashing F8 and hoping you've captured the desired image. ;) But yeah, it's not mutually exclusive with mashing F8 to get multiple snapshots per second.

plugins/GSdx/GSRenderer.cpp
```diff
@@ -557,12 +557,22 @@ bool GSRenderer::MakeSnapshot(const string& path)
 	if(m_snapshot.empty())
 	{
 		time_t t = time(NULL);
```
turtleli: Why not just compare time_t values?

plugins/GSdx/GSRenderer.cpp
```diff
@@ -557,12 +557,22 @@ bool GSRenderer::MakeSnapshot(const string& path)
...
+		static int n = 2;
+		static char prev_snap[16];
+		char cur_time[16];
```
turtleli: Buffer overflow?

ReyVGM:
Guys, don't overthink this. 

Just do what the Dolphin guys did with this very same issue was just add a (1), (2) after each image. For example:
Game 07:59.png
Game 08:00.png
Game 08:00 (1).png
Game 08:00 (2).png
Game 08:01.png
I believe that should be simple enough and would solve the issue.

I would advise against changing the naming from date/hour to sequential numbers because if you take a bunch of screenshots, and then delete an early screen, then when you take a new screen instead of continuing with the last number it will try to fill that slot you deleted and then you'll have non-chronological screenshots. With date/hour naming, that will never happen.

Hirato:
I think it might be better if a _00 was tacked onto the end of the name, and then incremented to _01, _02, etc, in the event of collision.
it should account for people making any ridiculous number of screenshots per second, and it'll still be sorted.


That aside, isn't webm the go to format nowadays for people who want to record and share a little bit of footage?
You could just capture a few seconds of PCSX2 via obs and call it a day, and if you still want to make gifs from it, it might be a lot easier to step through that footage.

ReyVGM:
You can't take pixel-perfect images from videos, much less make pixel perfect animated GIFs.

MrCK1:
Isn't M-JPEG a bunch of individual images stitched together to make a video?

mirh:
Pixel perfect means lossless. 
100 quality m-jpeg may possibly be visually indistinguishable from that, but if you have to convert and stuff on it, I guess at that point is just better to shoot uncompressed.

gregory38:
On Linux, the GS video recording feature is implemented as consecutive frame dump (and people need to convert it to video). Maybe it can be extended. The PNG compression is done on separate threadS.

ReyVGM:
Oh please don't mess with the screenshot button. Pressing F8 and getting a bunch of frame dump images is fine, but make it an option for people that want to make a sequence or animated GIF.

But for now, what needs to be improved is the naming of the screenshots that only allows one screen per second.

gregory38:
IMHO, you need a gif option, that will transform the F8 as save N frames of data. It is much more clever than smashing F8. And there is already a linear naming.
```
     	m_out_dir = theApp.GetConfigS("capture_out_dir");
	std::string out_file = m_out_dir + format("/frame.%010d.png", m_frame);
```
However, I potentially broke the frame counter. Anyway, it can be improved without changing the standard snapshot behavior.

ssakash:
So what's the conclusion on naming format of the screenshots? Should it be ``time (n)`` **(or)** ``time-n``
**(or)** ``time_n``?

FlatOutPS2:
> So what's the conclusion on naming format of the screenshots? Should it be time (n) (or) time-n
(or) time_n?

It can also be `time_(n)` (but not `time-(n)`).

plugins/GSdx/GSRenderer.cpp
```diff
@@ -556,13 +556,23 @@ bool GSRenderer::MakeSnapshot(const string& path)
 {
 	if(m_snapshot.empty())
 	{
-		time_t t = time(NULL);
+		time_t cur_time = time(nullptr);
+		static time_t prev_snap;
+		static int n = 2;
```
gregory38: Just a quick comment why the reset value is 2 and not 0 (or 1)

turtleli:
> So what's the conclusion on naming format of the screenshots? Should it be `time (n)` (or) `time-n`
> (or) `time_n`?

I'd prefer one of the versions without spaces.

plugins/GSdx/GSRenderer.cpp
ssakash: I'm not sure if you're asking me to add a comment on the source code (or) asking me to explain it to you right now?

gregory38: In the code. I guess to keep compatibility, there is no .1 image hence the .2 start.

ssakash:
I went with ``time_(n)``, thanks for all the feedback!

PR #1472, opened 2016-07-22
Gsdx gta depth issue 1457

gregory38:
Need to test game on Dx (openGL support depth correctly)

PR #1776, opened 2017-01-16
Revert "gsdx tc: partial support of pseudo depth for Dx"

FlatOutPS2:
This reverts commit
https://github.com/PCSX2/pcsx2/commit/d6383e6c211e03a9bb036c6fc8655cd6bba8b9ff
Link to original pull request:
https://github.com/PCSX2/pcsx2/pull/1472

It created a regression in Everybody's Golf 4/Hot Shots Golf 4, breaking the renderering when depth emulation is disabled/when using a Direct3D Hardware renderer.


gregory38:
I'm pretty sure it was also the source of others regression in the Dx renderer. Thanks you.

PR #1783, opened 2017-01-18
PCSX2: Fix stall on branch .. in delay slot

FlatOutPS2:
Fixes stall when loading a stage in WRC 3.

Removes patch that served as workaround.

### DIFF
--- a/pcsx2/x86/ix86-32/iR5900-32.cpp
+++ b/pcsx2/x86/ix86-32/iR5900-32.cpp
@@ -1315,17 +1315,29 @@ void recompileNextInstruction(int delayslot)
 			case 1:
 				switch(_Rt_) {
 					case 0: case 1: case 2: case 3: case 0x10: case 0x11: case 0x12: case 0x13:
-						Console.Warning("branch %x in delay slot!", cpuRegs.code);
+						DevCon.Warning("branch %x in delay slot!", cpuRegs.code);
 						_clearNeededX86regs();
 						_clearNeededXMMregs();
+						pc += 4;
+						g_cpuFlushedPC = false;
+						g_cpuFlushedCode = false;
+						if (g_maySignalException)
+							xAND(ptr32[&cpuRegs.CP0.n.Cause], ~(1 << 31)); // BD
+						g_recompilingDelaySlot = false;
 						return;
 				}
 				break;
 
 			case 2: case 3: case 4: case 5: case 6: case 7: case 0x14: case 0x15: case 0x16: case 0x17:
-				Console.Warning("branch %x in delay slot!", cpuRegs.code);
+				DevCon.Warning("branch %x in delay slot!", cpuRegs.code);
 				_clearNeededX86regs();
 				_clearNeededXMMregs();
+				pc += 4;
+				g_cpuFlushedPC = false;
+				g_cpuFlushedCode = false;
+				if (g_maySignalException)
+					xAND(ptr32[&cpuRegs.CP0.n.Cause], ~(1 << 31)); // BD
+				g_recompilingDelaySlot = false;
 				return;
 		}
 	}
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -1731,12 +1731,6 @@ Name   = WRC 3
 Region = PAL-M8
 Compat = 5
 XgKickHack = 1
-[patches = 80802EA9]
-comment= Patch By Prafull
-// fix hang at loading screen
-patch=1,EE,002aa934,word,00000000
-patch=1,EE,002aa574,word,00000000
-[/patches]
 ---------------------------------------------
 Serial = SCES-51685
 Name   = Primal


refractionpcsx2:
That was quick! I assumed branches in the delay slot would be handled like the vu ones, which would be a pig to fix. 

Trying to think if there's any other games that do this too. 

Good job

refractionpcsx2:
Apparently wrc really evolved and possibly midnight club 2 suffered from these too

MonJamp:
@FlatOutPS2 Do you know if this fixes enabling MTVU for Thrillville (you have the game right)?

refractionpcsx2:
This won't fix games that break when hacks are enabled. The mtvu issue is completely separate

willkuer:
May I ask how you came to this solution? Were you checking what the hack did? I guess there is no forum in the web where people discussed 'well, if you have a stall in branch … delay slot, just put these lines and the errors are gone…' and ps2 documentation also rather seems poor...

refractionpcsx2:
PS2 documentation is poor and I'm not sure how the EE handles it.

On the VU side, what happens is the VU will take the first branch, read the first instruction, then take the second branch (if it succeeded).  Of course if the first branch wasn't taken, as long as it isn't a Likely branch (speaking from the EE side, the VU doesn't have likely's), it would take the second branch only.

however it seems there is an exception for when there is a branch in the branch delay slot, so it's possible the EE does handle it completely differently and just hands it off to the exception manager to deal with.

gregory38:
I would suggest to run the EE PS2 testsuite. Initial code got an infinite function call loop.
The global variables to handle the flush are not really used. (All registers allocation is completely broken, the big part to redo on x64)

ramapcsx2:
Woho, a EE recompiler fix. Awesome!
Would be great to test this on a few more games.
The positive (buggy) case will be very clear: It writes "branch xxxx in delay slot!" :)

prafullpcsx2:
This looks interesting. Will test it thoroughly after my tournament ends in two days time.

pcsx2/x86/ix86-32/iR5900-32.cpp
```diff
@@ -1315,17 +1315,29 @@ void recompileNextInstruction(int delayslot)
...
 					case 0: case 1: case 2: case 3: case 0x10: case 0x11: case 0x12: case 0x13:
-						Console.Warning("branch %x in delay slot!", cpuRegs.code);
+						DevCon.Warning("branch %x in delay slot!", cpuRegs.code);
 						_clearNeededX86regs();
 						_clearNeededXMMregs();
+						pc += 4;
+						g_cpuFlushedPC = false;
+						g_cpuFlushedCode = false;
+						if (g_maySignalException)
+							xAND(ptr32[&cpuRegs.CP0.n.Cause], ~(1 << 31)); // BD
```
ssakash: According to the documents, BD consists of two bits (BD & BD2) and both of them seem to have the same functionality, shouldn't you also clear the second bit if you're doing it for the first one? Any specific reason to only clear the first one?

ssakash:
> That said we aren't actually setting the BD bit, we're just clearing it?

Yeah, doesn't make sense to me either. According to the sources, we only need to set BD when encountering exceptions other than the reset, NMI, performance counter and debug on the delay slot. I don't seem to find any information about clearing the bit at any cases.

Also seems like @sudonim1 was the one who introduced it in https://github.com/PCSX2/pcsx2/commit/d4bc1a62755b6d4fcaf873d85bb03ef61b3d5e0a, so assigning this to him as he probably knows how this works.

gregory38:
@ssakash here the theory
When you encounter an exception/interrupt, you need to stop the program to handle the exception/interrupt. When it is done you must resume the execution. Therefore you must save the PC before jumping to the exception handler (saved in the EPC register)
For example:
```
0x400:load v0, 0;
0x404:load v1, v0; // may generate a TLB miss
0x408:load v0, v0;
```
2nd load will generate a TLB miss. So you save the PC (here 0x404) in the EPC register. You execute the TLB miss handler. Then you return to the program with a `jump $EPC`.

Now imagine, that exception occurred in a delay slot.
```
0x400:load v0, 0;
0x404:jump 0x410;
0x408:load v1, v0; // may generate a TLB miss
0x40C:load v0, 1;
0x410: print $v0;
```
Again the CPU will save the PC of the instruction that raise the exception. Here 0x408. Handler will be executed. And finally jump $EPC (reminder 0x408). So next instruction will be 0x40C and 0x410. As you can see you didn't resume the execution correctly (will print 1). An exception/interrupt must never change the program behavior. So instead you save $EPC + the BD flag. This way, you will jump back to $EPC-4 (so 0x404) so program will resume correctly (will print 0).

There is 2 level of exceptions, hence BD/BD2. It is easier to handle in the CPU (and potentially it allow lvl2 exception in lvl1 exception). LVL2 exception must never happen in the emulation.

Now remain to answer the behavior of  what must happen if there is a branch in the delay slot case. I don't remember the exact behavior. Either 2nd jump is a NOP or final address is the one of the 2nd jump. Behavior can be checked with the testsuite.
```
load v0, imm;
jump 0x1000;
jump 0x2000; // jump in delay slot (could trigger a tlb miss)
jump 0x3000;
print PC; <= likely 0x1000 (or 0x2000)
```

BD will be cleared when you resume program execution (check ERET opcode). Potentially the behavior is equivalent to (for every cycle)
```
if (cpu is in std mode)
  bd = jump/branch opcode ? 1 : 0;
else if (cpu is in exception 1 mode)
  bd2 = jump/branch opcode ? 1 : 0;
else
   nop
```

gregory38:
@FlatOutPS2 
I forgot to say in my wall of text, could you create an handy function to avoid copy/past.

ssakash:
@gregory38 

Thanks for the write up, really appreciated. :)

> I forgot to say in my wall of text, could you create an handy function to avoid copy/past.

Agreed, I was also planning on mentioning it but forgot about it for some reasons.

sudonim1:
I think that commit of mine was just preparation for future work that has only kind of happened because at least at the time it was impossible for any recompiler instruction to raise an exception.

gregory38:
Please add a common function. And I need to run the test suite.

gregory38:
do you have the instructions around the branch in branch delay slot.

From the HW testsuite:
```	int result = -1;

	// This is technically not valid MIPS anyway.
	// But it seems to always take target 2?
	asm volatile (
		".set    noreorder\n"

		"li      $t0, 0\n"
		"beq     $t0, $t0, target1_%=\n"
		"beq     $t0, $t0, target2_%=\n"
		"beq     $t0, $t0, target3_%=\n"
		"beq     $t0, $t0, target4_%=\n"
		"nop\n"

		"target1_%=:\n"
		"li      %0, 1\n"
		"j       skip_%=\n"
		"nop\n"

		"target2_%=:\n"
		"li      %0, 2\n"
		"j       skip_%=\n"
		"nop\n"

		"target3_%=:\n"
		"li      %0, 3\n"
		"j       skip_%=\n"
		"nop\n"

		"target4_%=:\n"
		"li      %0, 4\n"
		"j       skip_%=\n"
		"nop\n"

		"skip_%=:\n"

		: "+r"(result)
	);

	printf("beq: delay branch: %08x\n", result);
```
So far we return 1 (even with this PR), and the HW 2.

sudonim1:
So basically branch in branch delay slot on real hardware works like you'd
expect.  The delay slot is fetched/the first branch executed, the first
branch target fetched/the second branch executed, the first branch target
executed/the second branch target fetched, execution continues from the
second branch target.

gregory38:
@FlatOutPS2 could you create a generic function ?

FlatOutPS2:
> could you create a generic function ?

I already did, but I was still to look into some possibly related issues before updating the PR. I just haven't gotten around to it yet.

atomic83GitHub:
Any status?

MrCK1:
I don't think this will be merged anytime soon. Flat out was in charge and we'll, he's gone for now.

atomic83GitHub:
Apparently Midnight club games are fixed with this PR.

atomic83GitHub:
I just seen that on a forum post (about the 2nd if a remember).

FlatOutPS2:
Someone once said != actually true.

I don't see any post on the forum about testing this PR with that game.

atomic83GitHub:
This wasn't in the PCSX2 forum, I unfortunately don't remember it because It's been a long time I seen that :). Anyway is there any other games affected by this?

FlatOutPS2:
> Anyway is there any other games affected by this?

No. The only other mentions of this issue are due to cheats/hacks.

refractionpcsx2:
The only other game I know I mentioned earlier in the thread is WRC 4 i think it is

FlatOutPS2:
WRC Evolved. ;) Though it should also WRC 4, but that has other issues not related to this PR.

atomic83GitHub:
The Action Replay MAX throw this in the console when applying Ratchet and Clank 2 cheats (which instantly close PCSX2):

capture

But even with your PR, it won't work.

atomic83GitHub:
I tested this PR with a lot of games and I have 2 improvement:

-Action replay MAX
-WRC 2005 avec Sébastien Loeb

I think it's safe to merge as I did not notice any regression with it.

lightningterror:
@refractionpcsx2 , @ramapcsx2 what are your thoughts, what are we gonna do with this?
@atomic83GitHub seems to have tested a bunch of games with no new issues popping up.

refractionpcsx2:
If it works, merge it. Better than what we've got. however due tot he gamedb conflicts you might have to open a new PR

ramapcsx2:
@lightningterror 
If you're doing a new PR, could you add a comment somewhere that links to the earlier discussion here?
It seems this code change works in favor of the games, but is still not entirely correct.
@gregory38 tested this in 2017 using the ps2autotests suite and remarked "So far we return 1 (even with this PR), and the HW 2."

I feel this bit shouldn't be lost. A comment could either link here, or summarize that there's still problems with the test suite.
Thanks :)

lightningterror:
Closing in favor of https://github.com/PCSX2/pcsx2/pull/2919

PR #1786, opened 2017-01-23
GSdx-PCRTC: Improve output circuit auto selection

ssakash:
**Summary of changes**:

* Improve auto selection of the output circuit when ``GetDisplayRect()`` is called without any parameter.
  **(refer the commit message for more details)**
* Update a debug message on ``GetFrameRect()``.
* Remove some redundant conditional checks.

### DIFF
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -405,9 +405,27 @@ GSVideoMode GSState::GetVideoMode()
 	return videomode;
 }
 
-GSVector4i GSState::GetDisplayRect(int i, bool merged_rect)
+GSVector4i GSState::GetDisplayRect(int i)
 {
-	if(i < 0) i = IsEnabled(1) ? 1 : 0;
+	if (!IsEnabled(0) && !IsEnabled(1))
+		return GSVector4i(0);
+
+	// If no specific context is requested then pass the merged rectangle as return value
+	if (i == -1)
+	{
+		if (m_regs->PMODE.EN1 & m_regs->PMODE.EN2)
+		{
+			GSVector4i r[2] = { GetDisplayRect(0), GetDisplayRect(1) };
+			GSVector4i r_intersect = r[0].rintersect(r[1]);
+			GSVector4i r_union = r[0].runion_ordered(r[1]);
+
+			// If the conditions for passing the merged rectangle is unsatisfied, then
+			// pass the rectangle with the bigger size.
+			bool can_be_merged = !r_intersect.width() || !r_intersect.height() || r_intersect.xyxy().eq(r_union.xyxy());
+			return (can_be_merged) ? r_union : r[r[1].rarea() > r[0].rarea()];
+		}
+		i = m_regs->PMODE.EN2;
+	}
 
 	GSVideoMode videomode = GetVideoMode();
 	GSVector2i magnification (m_regs->DISP[i].DISPLAY.MAGH + 1, m_regs->DISP[i].DISPLAY.MAGV + 1);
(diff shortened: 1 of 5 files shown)


plugins/GSdx/GSState.cpp
```diff
@@ -405,9 +405,30 @@ GSVideoMode GSState::GetVideoMode()
 	return videomode;
 }
 
-GSVector4i GSState::GetDisplayRect(int i, bool merged_rect)
+GSVector4i GSState::GetDisplayRect(int i)
 {
-	if(i < 0) i = IsEnabled(1) ? 1 : 0;
+	if (!(IsEnabled(0) || IsEnabled(1)))
```
gregory38: IMHO, `!A && !B` is easier to read than `!(A || B)`

plugins/GSdx/GSState.cpp
```diff
@@ -405,9 +405,30 @@ GSVideoMode GSState::GetVideoMode()
...
+	// If no specific context is requested then pass the merged rectangle as return value
+	if (i == -1)
+	{
+		if (m_regs->PMODE.EN1 & m_regs->PMODE.EN2)
+		{
+			GSVector4i r[2] = { GetDisplayRect(0), GetDisplayRect(1) };
+			GSVector4i r_intersect = r[0].rintersect(r[1]);
+			GSVector4i r_union = r[0].runion_ordered(r[1]);
+
+			if (!(r_intersect.width() && r_intersect.height()) ||
```
gregory38: It is fine. But potentially it could be easier to use `?` This way it is easier to see that code will always return. And it avoid the strange indentation on the if.
```
/* the comment here */
bool can_be_merged = bla ..intersect... bla
return can_be_merged ? r_union : r[...area...];
```

plugins/GSdx/GSState.cpp
```diff
@@ -405,9 +405,27 @@ GSVideoMode GSState::GetVideoMode()
...
+			GSVector4i r[2] = { GetDisplayRect(0), GetDisplayRect(1) };
+			GSVector4i r_intersect = r[0].rintersect(r[1]);
+			GSVector4i r_union = r[0].runion_ordered(r[1]);
+
+			// If the conditions for passing the merged rectangle is unsatisfied, then
+			// pass the rectangle with the bigger size.
+			bool can_be_merged = !(r_intersect.width() && r_intersect.height()) || r_intersect.xyxy().eq(r_union.xyxy());
+			return (can_be_merged) ? r_union : r[r[1].rarea() > r[0].rarea()];
+		}
+		GetDisplayRect(m_regs->PMODE.EN2);
```
turtleli: Should be ` i = IsEnabled(1) ? 1 : 0;`? Otherwise you'll access m_regs->DISP[] with a negative index.

ssakash: As far as I can see, I'm only accessing PMODE here, not DISP. PMODE doesn't have multiple contexts.

turtleli: You're missing a return though. And instead of a recursive call you can just change `i`.

ssakash: > You're missing a return though

Oops, you're right. I guess I'll just change the value of ``i`` instead of ``return GetDisplayRect(...)``

gregory38:
Is it good ?

gregory38:
@ssakash was it tested?

ssakash:
I think it's good, didn't do much testing though. If you think it needs more testing, you could leave it open for 2-3 days.

@FlatOutPS2 : Could you do some checks for regression? Thank you.

plugins/GSdx/GSState.cpp
```diff
@@ -405,9 +405,27 @@ GSVideoMode GSState::GetVideoMode()
...
+	{
+		if (m_regs->PMODE.EN1 & m_regs->PMODE.EN2)
+		{
+			GSVector4i r[2] = { GetDisplayRect(0), GetDisplayRect(1) };
+			GSVector4i r_intersect = r[0].rintersect(r[1]);
+			GSVector4i r_union = r[0].runion_ordered(r[1]);
+
+			// If the conditions for passing the merged rectangle is unsatisfied, then
+			// pass the rectangle with the bigger size.
+			bool can_be_merged = !(r_intersect.width() && r_intersect.height()) || r_intersect.xyxy().eq(r_union.xyxy());
```
gregory38: Tell me if is the same as (or if I'm wrong)
```
!r_intersect.width() || !r_intersect.height() || r_intersect.xyxy().eq(r_union.xyxy())
```
If I'm right, don't you think it is easier this way ?

ssakash: Yeah, it seems more readable.

MrCK1:
@ssakash The intro in JMMT is still broken. However, I noticed that the emulator changes the output to Progressive when the FMV starts playing.

PR #1788, opened 2017-01-25
Plugin Selector: Make plugins easier to read (remove Git version string).

MrCK1:
This PR simply shortens the displayed plugin name by removing the version/build time from the displayed string to make it easier to read in release builds.

For instance,
Current
```Lilypad (2015121311175709) 0.11.0 [lilypad]```

Proposed
 ```Lilypad 0.11.0 [lilypad]```

### DIFF
--- a/plugins/FWnull/FW.cpp
+++ b/plugins/FWnull/FW.cpp
@@ -69,7 +69,7 @@ PS2EgetLibType()
 EXPORT_C_(char *)
 PS2EgetLibName()
 {
-    snprintf(libraryName, 255, "FWnull Driver");
+    snprintf(libraryName, 255, "FWnull Driver %lld%s", SVN_REV, SVN_MODS ? "m" : "");
     return libraryName;
 }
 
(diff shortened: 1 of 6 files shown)


FlatOutPS2:
The build time can be useful when a user accidentally uses an outdated plugin in a newer PCSX2 build(is known to happen after copying a newer build over an older one). The actual version numbers are rarely updated.

FYI, you need to use clang format(probably just replace the tabs with spaces) in the plugin files or they won't compile on Linux. ;)

mirh:
I also agree with flatout. 
When i mix/mess/hack/compile myself stuff, it starts to get really difficult to distinguish what's what. 

On an "official release" though, I agree that shouldn't be there. 
...
Besides, why are you removing the version from dlls, instead of the "check" inside UI?

MrCK1:
Maybe a better solution would be to shorten the build time string to ```YYYY/MM/DD``` instead of removing it entirely?

MrCK1:
I'm lazy, not really interested in doing this anytime soon :P

PR #1790, opened 2017-01-26
gsdx:windows: Remove SSSE3 and AVX configurations

turtleli:
See discussion at #796.


gregory38:
Nobody complain (yet) let's go. Thanks for the update.

CeruleanSky:
>Nobody complain (yet) let's go. Thanks for the update.

Considering the audience, those who know about orphis buildbot and what SSE, AVX vs AVX2 is and would notice it missing, etc, I am not surprised at the lack of people speaking up.

The last build I used was Jan 13th, 2017 , 14 days before AVX was removed, I went to update today and noticed it was missing and saw this commit on orphis's page.

I'd like to request a partial revert and a continuation of building the AVX plugin.

AVX CPU

PR #1791, opened 2017-01-26
Remove old cdvd plugins

turtleli:
This removes cdvdlinuz, cdvdiso, cdvdisoEFP and cdvdolio.

The internal ISO reader and cdvdgigaherz cover most of the ISO/disk reading functionality these plugins have. The only feature which I think will be lost is the custom bzip2/zlib compressed ISO functionality in cdvdiso/cdvdisoEFP.

I made a project a while back to compare the plugins (I haven't finished looking at cdvdpeops yet): https://github.com/PCSX2/pcsx2/projects/3


refractionpcsx2:
Can we get a GUI button for creating blockdumps with the internal iso reader then please? I still use linuziso (or cdvdlinuz or whatever it's called) for creating blockdumps because I don't know how to use the built in one :P

MrCK1:
These two lines still need to be removed from .gitignore as well.
```!/plugins/CDVDpeops/i386.asm```
```/plugins/CDVDolio/Template```

gregory38:
It could be nice to add a git tag before the removal, in case someone want to dig back the code.

turtleli:
> Can we get a GUI button for creating blockdumps with the internal iso reader then please? I still use linuziso (or cdvdlinuz or whatever it's called) for creating blockdumps because I don't know how to use the built in one :P

Can't see how it's hard. ;)
But sure, I'll add some sort of GUI item somewhere.

> This line still needs to be removed from .gitignore as well.
> `/plugins/CDVDolio/Template`

That line hasn't been necessary for quite a while. There are other lines in the .gitignore that need removed/updated.

> It could be nice to add a git tag before the removal, in case someone want to dig back the code.

I think going back to the 1.4.0 tag would be enough if someone wants to look at the code? And if a tag is added, we might have people claiming we've released a new version. ;)

FlatOutPS2:
> Can't see how it's hard. ;)
But sure, I'll add some sort of GUI item somewhere.

Because it's not properly explained anywhere AFAIK.

gregory38:
Fine for the 1.4 tag.

mirh:
Dumb and OT question (as always..)
Why do we have older versions in tags *and* in branches, if *anyway* it's not like fixes has ever been backported or somebody ever cared about maintaining old features?

gregory38:
RIP

PR #1794, opened 2017-01-26
pcsx2:gui: Add menu item for creating block dumps

turtleli:
This adds a menu item for creating blockdumps. I put it in the Debug menu, though maybe it belongs in the CDVD menu.

I didn't think too hard on the strings, so if you have suggestions, go for it.


FlatOutPS2:
The debug menu seems like the right location to me.

gregory38:
Yes it is fine.

Issue #1171, opened 2016-02-09
GSDX: Drop (or) keep D3D9 renderer ?


ssakash:
It was recently stated that the last stable version (1.4) will be the final version which will have support for windows XP, so having D3D9 renderer isn't as necessary as before since we don't have to worry about people with windows XP.

**Advantages on removing D3D9 renderer:**
- Initial step on removing DX SDK dependency.
- prevents inexperienced users from using D3D9 since it misses lots of recent changes like Snowblind fix and other stuffs.
- D3D9 renderer also has lots of other issues which aren't present on D3D11/OpenGL. less better choices makes it more user friendly for the users.
- No need to have a logo for D3D9 and also saves space at combo box.

**Disadvantages on removing D3D9 renderer:**
- People with graphics cards lower than Fermi / AMD equivalent can't use any renderers on GSDX.

refractionpcsx2:
I think I'm for it. It serves little purpose these days with OpenGL being so good, and as you say, we aren't supporting Windows XP after 1.4, so that's a non-issue.

turtleli:
> Initial step on removing DX SDK dependency.

It's possible to remove the DX SDK dependency from GSdx without removing the D3D9 renderer.

Nobbs66:
Like ref said, if it's not getting in the way, then just change the default to OGL or DX11, but still allow those on old hardware to use Dx9

refractionpcsx2:
Yeh i think the big issue was the windows sdk not supporting windows xp, so it should be fine.

Nobbs66:
Maybe for 1.6 DX9 can be removed, but the minimum requirements will have to be updated to fit.

refractionpcsx2:
That's the idea, that's why we are in the development period for 1.6 :P

Nobbs66:
Yeah, I guess I really don't care if it's there or not. The default does need changing though.

FlatOutPS2:
> Like ref said, if it's not getting in the way, then just change the default to OGL or DX11, but still allow those on old hardware to use Dx9

The default is already DX11. If you use a fresh PCSX2 build and boot into a game without changing any VIdeo Plugin settings DX11 is used, not DX9(even though it does show DX9 as default in the VIdeo Plugin settings menu).

vsub:
> The default is already DX11

[...]

PR #1284, opened 2016-04-08
GSDX: Improvements to the GSDX dialog

ssakash:
**Summary of changes:**

* Add new logos for the renderers. **( logos were made by [bmate](http://forums.pcsx2.net/User-bmate) )**

_**D3D9**_
D3D9

_**D3D11**_
D3D11

_**OpenGL**_
ogl

_**Null**_
null

* Gray out upscaling hacks on native resolution.

* Properly display usage of D3D10/D3D11 on the combobox. Also, Sort all the renderers in ascending order.

* Use highest available version of DX by default when GSDX dialog is used the first time (or) when GSDX.ini is not found.

* Remove all the extra NULL renderers with different rendering backends. ( There should be no difference between them and it makes maintaining renderer related code a bit easier)

**Build download link : [here](https://ci.appveyor.com/api/buildjobs/8m9je19co68kox92/artifacts/pcsx2-v1.5.0-dev-576-g51b413f-master-816-vs2015-Win32-AppVeyor.7z)**

PR #1384, opened 2016-06-03
GSDX-Windows: Better detection of default renderer

ssakash:
**Summary of Changes**:

*  Better detection of default renderer based on the vendor ID ( Nvidia, AMD , Intel) [(Thanks to mirh)](https://github.com/PCSX2/pcsx2/pull/1370#issuecomment-222302060)
*  Gray out ``Disable Depth Emulation`` for renderers other than OpenGL.

Closes #1039 

**Testing Required**:
- [x] Nvidia (Tested myself , works fine)
- [x] AMD (**Flatout** tested it and seems to be fine)
- [x] Intel IGPU (**Bositman** and **Flatout** tested it, seems to be fine)

PR #1795, opened 2017-01-27
GSdx: Integrate software renderers

ssakash:
**Summary of changes**:

* Integrate all the software renderers into a single item in renderer combobox.
* Slightly modify the F9 toggle behavior to make it suitable with the new integration of software renderers.
* I've update the enumeration values in GSRendererType, I prefer to keep the code clean in cost of temporarily breaking the INI compatibility.

**TODO**:

* Missing logo for software renderer?
* Maybe ``DX1011_HW``, ``DX9_HW``... could be converted to ``DX1011``, ``DX9`` respectively as there's only one software renderer right now? (I mean the enumeration in code, _nothing to see here for end users_)

I remember having a discussion with Gregory on IRC before and the consensus was to use D3D9 for SW rendering on Windows and OpenGL for Linux if there was a plan to integrate the software renderers. I did that approach on this PR, actually I was planning on making this change [**last year**](https://github.com/PCSX2/pcsx2/pull/1284#issuecomment-207447845) but I only got the time to do it now. :P

_**Master branch**_:

Before

_**Pull request branch**_:

Now

### DIFF
--- a/plugins/GSdx/GS.h
+++ b/plugins/GSdx/GS.h
@@ -231,13 +231,11 @@ enum class GS_MIN_FILTER : uint8_t
 enum class GSRendererType : int8_t
 {
 	Undefined = -1,
-	DX9_HW,
-	DX9_SW,
-	DX1011_HW = 3,
-	DX1011_SW,
+	DX9,
+	DX1011 = 3,
 	Null = 11,
-	OGL_HW,
-	OGL_SW,
+	OGL,
+	Software,
 	DX9_OpenCL,
 	DX1011_OpenCL,
 	OGL_OpenCL = 17,
@@ -247,7 +245,7 @@ enum class GSRendererType : int8_t
 #else
 	// Use ogl renderer as default otherwise it crash at startup
 	// GSRenderOGL only GSDeviceOGL (not GSDeviceNULL)
-	Default = OGL_HW
+	Default = OGL
 #endif
 
 };
--- a/plugins/GSdx/GS.cpp
+++ b/plugins/GSdx/GS.cpp
@@ -236,41 +236,18 @@ static int _GSopen(void** dsp, const char* title, GSRendererType renderer, int t
 		}
 
 		const char* renderer_fullname = "";
-		const char* renderer_mode = "";
-
-		switch (renderer)
-		{		
-		case GSRendererType::DX9_SW:
-		case GSRendererType::DX1011_SW:
-		case GSRendererType::OGL_SW:
-			renderer_mode = "(Software mode)";
-			break;
-		case GSRendererType::Null:
-			renderer_mode = "(Null mode)";
-			break;
-		case GSRendererType::DX9_OpenCL:
-		case GSRendererType::DX1011_OpenCL:
-		case GSRendererType::OGL_OpenCL:
-			renderer_mode = "(OpenCL)";
-			break;
-		default:
-			renderer_mode = "(Hardware mode)";
-			break;
-		}
 
 		switch (renderer)
 		{
 		default:
 #ifdef _WIN32
-		case GSRendererType::DX9_HW:
-		case GSRendererType::DX9_SW:
+		case GSRendererType::DX9:
 		case GSRendererType::DX9_OpenCL:
 			dev = new GSDevice9();
 			s_renderer_name = " D3D9";
 			renderer_fullname = "Direct3D9";
 			break;
-		case GSRendererType::DX1011_HW:
-		case GSRendererType::DX1011_SW:
+		case GSRendererType::DX1011:
 		case GSRendererType::DX1011_OpenCL:
 			dev = new GSDevice11();
 			s_renderer_name = " D3D11";
@@ -282,16 +259,24 @@ static int _GSopen(void** dsp, const char* title, GSRendererType renderer, int t
 			s_renderer_name = " Null";
 			renderer_fullname = "Null";
 			break;
-		case GSRendererType::OGL_HW:
-		case GSRendererType::OGL_SW:
+		case GSRendererType::OGL:
 		case GSRendererType::OGL_OpenCL:
 			dev = new GSDeviceOGL();
 			s_renderer_name = " OGL";
 			renderer_fullname = "OpenGL";
 			break;
+		case GSRendererType::Software:
+#ifdef _WIN32
+			dev = new GSDevice9();
+#else
+			dev = new GSDeviceOGL();
+#endif
+			s_renderer_name = " SW";
+			renderer_fullname = "Software mode";
+			break;
 		}
 
-		printf("Current Renderer: %s %s\n", renderer_fullname, renderer_mode);
+		printf("Current Renderer: %s\n", renderer_fullname);
 
 		if (dev == NULL)
 		{
@@ -304,24 +289,22 @@ static int _GSopen(void** dsp, const char* title, GSRendererType renderer, int t
 			{
 			default:
 #ifdef _WIN32
-			case GSRendererType::DX9_HW:
+			case GSRendererType::DX9:
 				s_gs = (GSRenderer*)new GSRendererDX9();
-				s_renderer_type = " HW";
+				s_renderer_type = "";
 				break;
-			case GSRendererType::DX1011_HW:
+			case GSRendererType::DX1011:
 				s_gs = (GSRenderer*)new GSRendererDX11();
-				s_renderer_type = " HW";
+				s_renderer_type = "";
 				break;
 #endif
-			case GSRendererType::OGL_HW:
+			case GSRendererType::OGL:
 				s_gs = (GSRenderer*)new GSRendererOGL();
-				s_renderer_type = " HW";
+				s_renderer_type = "";
 				break;
-			case GSRendererType::DX9_SW:
-			case GSRendererType::DX1011_SW:
-			case GSRendererType::OGL_SW:
+			case GSRendererType::Software:
 				s_gs = new GSRendererSW(threads);
-				s_renderer_type = " SW";
+				s_renderer_type = "";
 				break;
 			case GSRendererType::Null:
 				s_gs = new GSRendererNull();
@@ -349,8 +332,7 @@ static int _GSopen(void** dsp, const char* title, GSRendererType renderer, int t
 #ifdef _WIN32
 			switch (renderer)
 			{
-			case GSRendererType::OGL_HW:
-			case GSRendererType::OGL_SW:
+			case GSRendererType::OGL:
 			case GSRendererType::OGL_OpenCL:
 				s_gs->m_wnd = new GSWndWGL();
 				break;
@@ -491,7 +473,7 @@ static int _GSopen(void** dsp, const char* title, GSRendererType renderer, int t
 		return -1;
 	}
 
-	if (renderer == GSRendererType::OGL_HW && theApp.GetConfigI("debug_glsl_shader") == 2) {
+	if (renderer == GSRendererType::OGL && theApp.GetConfigI("debug_glsl_shader") == 2) {
 		printf("GSdx: test OpenGL shader. Please wait...\n\n");
 		static_cast<GSDeviceOGL*>(s_gs->m_dev)->SelfShaderTest();
 		printf("\nGSdx: test OpenGL shader done. It will now exit\n");
@@ -513,10 +495,14 @@ EXPORT_C_(void) GSosdMonitor(const char *key, const char *value, uint32 color)
 
 EXPORT_C_(int) GSopen2(void** dsp, uint32 flags)
 {
+	static GSRendererType recent_hw_renderer = GSRendererType::Undefined;
 	static bool stored_toggle_state = false;
 	bool toggle_state = !!(flags & 4);
 
 	GSRendererType renderer = s_renderer;
+	if (renderer == GSRendererType::DX1011 || renderer == GSRendererType::DX9 || renderer == GSRendererType::OGL)
+		recent_hw_renderer = renderer;
+
 	// Fresh start up or config file changed
 	if (renderer == GSRendererType::Undefined)
 	{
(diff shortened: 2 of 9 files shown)


FlatOutPS2:
> I remember having a discussion with Gregory on IRC before and the consensus was to use D3D9 for SW rendering on Windows and OpenGL for Linux if there was a plan to integrate the software renderers. I did that approach on this PR

from https://github.com/PCSX2/pcsx2/issues/83
"Is this intended to still be broken on the D3D software renderers? It works on OGL HW and SW, but why is there a difference in SW renderers between D3D and OGL?"

There might be other games with similar issues, and it would cause software renderer regressions for everyone on Windows.

ssakash:
Oh, that one was due to Gregory's recent feedback circuit improvement. Maybe we could have OpenGL renderer as the base device for software rendering among Windows and Linux?

On the other hand, seems like Linux compilation is failing. I'd probably fix it tomorrow, way past bed time right now. 💤

gregory38:
Yeah, all SW renderers are different.
* openGL is top notch (and potentially barely faster) but requires a working driver (look at Intel, AMD)
* DX11 can use advance effect of post processing shader
* DX9 is compatible with everyone.

plugins/GSdx/GS.h
```diff
@@ -232,15 +232,13 @@ enum class GSRendererType : int8_t
 {
 	Undefined = -1,
 	DX9_HW,
-	DX9_SW,
-	DX1011_HW = 3,
-	DX1011_SW,
-	Null = 11,
+	DX1011_HW,
```
gregory38: please don't change the value here. If would be painful for me, potentially some value are still hardcoded in the code. And I work only with the ini file.

prafullpcsx2:
Apart from reducing the size of renderer menu box what are the other benefits of this PR?

gregory38:
@prafullpcsx2 my guess is to please the anti-configuration user. We ought to propose them to use the Linux version instead (1 plugin, 1 renderer they will be happy).

Technically the DX9 SW renderer is mostly useless now. Unfortunately DX9 is still a requirements for WINE (OSX). In my opinion we should drop DX9 after the 1.6 release. If people want speed, they will stay on the future 1.6 release.

lightningterror:
This is a bad idea since opengl is slow on amd cards and let's not talk about intel which refuses to launch on windows 7 and 8. Don't know the state of windows 10.

Unless dx11 is the default software render.

gregory38:
@lightningterror 
Don't tell me that openGL SW renderer is slow on the AMD card. It is a matter of 1-10 draw calls by frame. AMD can do it :)

mirh:
Wait what?
Since when opengl software rendering has anything to do with graphics card driver?

I clearly remember I had no problem there, back with #1121

lightningterror:
@gregory38 Okay maybe around 1-2 fps difference in dbz bt3 , 0 in gt4 , fatal frame 2 4 fps , fatal frame 3 1-2 fps

Seriously ... I was expecting same results since the load is barely minimum but come on  amd ...

What worries me is intels drivers.

gregory38:
well with AMD, you never know, it can burn your house to use their driver ;) Seriously AMD is fine, except on older (DX10/9) HW because they didn't implemented some basic extension (Note same mess for Nvidia, Intel is out-of-scope).

@lightningterror seriously 2-4 fps ! It is crazy. Do you have extra shader enabled ?

Anyway here a trick from an old dev, read the PR comment before posting ;)
```
the consensus was to use D3D9 for SW rendering on Windows and OpenGL for Linux if there was a plan to integrate the software renderers. I did that approach on this PR,
```

mirh:
Shouldn't we then close #1171?

gregory38:
@mirh no.
I don't know what @ssakash want to do. But my feeling is that
* After 1.6 we drop DX9 (either people upgrade their GPU, either they use the 1.6 version). 1.8/2.0 won't be available soon anyway.
* We keep both remaining SW renderers but we could set DX11 as default (actually DX10).

mirh:
> either people upgrade their GPU

I thought our point never was that, but more something with performance, "OS compatibility" and simply "it doesn't hurt". 
But better I stop going OT.

gregory38:
`"OS compatibility"`, XP is a dead OS. And it will be more than dead for the 1.8 release. Modern application aren't compatible with old OS and old hardware. DX11 class GPU were released 7 years ago and we still don't use some new hardware capabilities. (both DX11 and OpenGL are limited to DX10 hardware class feature).

DX9 is faster because it is less accurate. Therefore you don't want to use a newer plugin that is more accurate and therefore slower. And I'm in favor of accuracy.

mirh:
> "OS compatibility", XP is a dead OS.

I was actually talking about wine in osx and whatever special linux case there may be.

gregory38:
Wine devs are working on dx11 support. Potentially GSdx dx11 might already work. I'm confident it will work for the 1.8 release. Honestly I don't care. But I think dx9 is old enough to be dropped after 1.6 release.

FlatOutPS2:
> But I think dx9 is old enough to be dropped after 1.6 release.

I don't think anyone on Linux and/or using an NVIDIA GPU should be allowed to make that decision. :p

gregory38:
Yeah, it was too painful to write from my phone (personally I don't care). Anyway, I propose to postpone the dx9 discussion after 1.6 release.

lightningterror:
@gregory38 Agreed.

ssakash:
> well with AMD, you never know, it can burn your house to use their driver ;) Seriously AMD is fine, except on older (DX10/9) HW because they didn't implemented some basic extension (Note same mess for Nvidia, Intel is out-of-scope).

I've added an automatic selection mechanism for selecting the best rendering API for software renderer based on your feedback.

So here's the pseudo code for the current selection pattern:

```
if( DX11 support && GPU != Intel)
{
return OpenGL
}
else
{
return D3D9/11;
}
```
Any opinion?

FlatOutPS2:
I'd still like an option to manually select the other Software renderers. Maybe something similar to the Automatic CRC Hack Level. An default automatic option, but also be able to manually change to another renderer.

mirh:
> I've added an automatic selection mechanism for selecting the best rendering API for software renderer based on your feedback.

Which would be #1384 minus AMD if I understand well?
It's still better on dx11 given #1552 -generally- abysmal performance.

gregory38:
So what do we do ? Honestly, if people want less configuration option, they need to install the Linux version. I'm afraid the multiple renderer will stay.

mirh:
I still can't understand if there's any actual difference between APIs in software renderer.

ssakash:
> I still can't understand if there's any actual difference between APIs in software renderer (because my impression was no, in which case even d3d9 would be fine)

Well, currently there's one difference - OpenGL GSDevice backend has support for feedback write whereas it's unimplemented on GSdx for the other APIs.

> I'd still like an option to manually select the other Software renderers. Maybe something similar to the Automatic CRC Hack Level. An default automatic option, but also be able to manually change to another renderer.

Adding an automatic setting would just make the combobox look even more messy with more options. AFAIK, no other project has multiple different APIs for software renderer - why should we then have all these options unnecessarily when software renderer is pretty much similar across all the different backends?

Though if you really want it then I could add it for the Dev/Debug builds of GSdx but I honestly think it would be pointless. There was a time when we were even having multiple different API's for Null renderer, the GSdx dialog was way more horrible back then. Removing all these unnecessary manual options would be a good step in reaching user friendly interface in my humble opinion.

> I'm afraid the multiple renderer will stay.

I don't mind closing the PR if more people want the multiple renderers to stay.

mirh:
> Well, currently there's one difference - OpenGL GSDevice backend has support for feedback write whereas it's unimplemented on GSdx for the other APIs.

-.- You had even mentioned it in second comment. 

Anyway.. Is that broken in intel/amd drivers? If not we could use GL everywhere, profit.

gregory38:
Maybe someone could port the grey scale shader to dx10 (merge function in the device).

turtleli:
> Well, currently there's one difference - OpenGL GSDevice backend has support for feedback write whereas it's unimplemented on GSdx for the other APIs.

OSD and external shaders are affected too.

gregory38:
```
Anyway.. Is that broken in intel/amd drivers? If not we could use GL everywhere, profit.
EDIT: Which GL version would that require?
```
Technically it requires GL3.3 hardware + some GL4 extensions.

Basically,
```
Ivy => 4.1 => KO

Sandy => 3.1 => KO
Nvidia/AMD DX10 GPU => 4.x => likely KO
```
Based on steam, you have 4-5% for Ivy and 10-11 for remaining DX10 GPU. It is annoying because SB/IB aren't that old. I really hope AMD will release cheap and powerful APU to reduce the market share of those limited GPUs.

mirh:
Amd dx10 hardware should range between [this](http://opengl.gpuinfo.org/gl_generatereport.php?reportID=1326) and [this](http://opengl.gpuinfo.org/gl_generatereport.php?reportID=886), while for nvidia [this](http://opengl.gpuinfo.org/gl_generatereport.php?reportID=1714) and [this](http://opengl.gpuinfo.org/gl_generatereport.php?reportID=1543).. 
If that's not enough.. Perhaps we could just put a check for GL [<](http://opengl.gpuinfo.org/gl_generatereport.php?reportID=1336)4.1, in which case switch to dx9 (is there a difference with respect to dx9)?

Possibly with a warning in emulog that tells you about the subtle switch.

gregory38:
@mirh 
https://github.com/PCSX2/pcsx2/blob/master/plugins/GSdx/GLLoader.cpp#L420
Mandatory extensions has `true` as 2nd parameters.

Maybe it would be possible to detect the GL support in `_GSopen`. Otherwise if the GPU/driver support DX11 => OpenGL will be supported (the only exception is IvyBridge).

mirh:
Well, they miss SSO, KHR_debug, buffer_storage and clip_control (which I guess make sense considering they are all written against ≥4 specs, sometimes even requiring it). 

So.. d3d makes more sense. Is there any downside in 9 compared to 11 then?

> Otherwise if the GPU/driver support DX11 => OpenGL will be supported (the only exception is IvyBridge).

Well, that's why I proposed to start from gl (where for some reason IB is a hell behind what it should be supposed to be if only looking at directx support)

gregory38:
```
Well, they miss SSO, KHR_debug, buffer_storage and clip_control (which I guess make sense considering they are all written against ≥4 specs, sometimes even requiring it).
```
Buffer storage is a must. It isn't an hardware limitation as the Linux driver exports those extensions.

```
So.. d3d makes more sense. Is there any downside in 9 compared to 11 then?
```
FXAA/Post-processing shader are better/supported on DX10. 

http://store.steampowered.com/hwsurvey/videocard/
Unfortunately Steam has a DX8 section which could be the not-detected GPU. Otherwise DX9 seems to be 0.5% of the market. The real question is what is the 3.0% otherwise I'm not sure it worth to bother for the remaining of 0.5%

mirh:
So....

````
if( GL support > 4.0)
{
return OpenGL
}
else
{
    if(dx10)
    {
    return D3D11;
    }
    else
    {
    return D3D9 
    }
}
````
?

**Comment by gregory38** [comment]
Yeah something like that. I will move the extension/gl check in the GSWnd setup. The trick will be to shuffle the init of GSDevice/GSRender/GSWnd to GSWnd/GSDevice/GSRender

**Comment by gregory38** [comment]
@mirh 
What would be the expected behavior when you toggle between HW and SW. Do we need to try openGL for the SW renderer even if HW was Dx ?

**Comment by mirh** [comment]
As far as I had intended this PR, we are just merging the sw renderers. Offering extra choices is redundant when the renderer is one and only one and there's no difference (as I had kind of hoped in #1637)

...

Ok this was the theory🙃. In practice we have aforementioned problems, so that's why we have the "checklist" (thank god the gl>dx11>dx9 hierarchy is clear cut). 
In other news, I think that it would be more elegant.. If before checking for GL>4 we checked for GL in the first place :s It's not elegant for "what's even this api" errors to appear.

**Comment by gregory38** [comment]
See PR #1836 for a better init of the window. Hopefully it will help to detect the openGL support.

**Comment by gregory38** [comment]
@mirh 
Yes theory, everybody got a modern OS with a modern GPU and modern Driver. 

Otherwise, may we can do something similar as `GSUtil::GetBestRenderer()` (just found the function) but for SW renderer
* level >= D3D_FEATURE_LEVEL_11_0 && (AMD || Nvidia) => OGL_SW
* level >= D3D_FEATURE_LEVEL_12_0 && Intel => OGL_SW
It won't be a perfect match but it will be close of the reality.

**Comment by mirh** [comment]
> Yes theory, everybody got a modern OS with a modern GPU and modern Driver.

Really, even an older OS with an old GPU would be enough.. *With a working driver*

>  It won't be a perfect match but it will be close of the reality.

I'm not sure why we are still talking about this, when the scheme above actually manages to cover all cases.

**Comment by mirh** [comment]
So.. #1836 got merged, which of the pseudo-code I suggested there should be put here?

**Comment by gregory38** [comment]
Life is more complicated. People uses tons of external shaders. Only Asmo's shaders are working.  I'm not sure we are ready to keep a single device for SW renderer.

**Comment by mirh** [comment]
> thank god the gl>dx11>dx9 hierarchy is clear cut

Ok, so this premise goes to hell (if dx10 would be better than ogl in some use-cases), and.. 
I guess now, if any, an attempt to remove clutter from menu.. Could only put a checkbox to enable/disable software renderer?

**Comment by FlatOutPS2** [comment]
My idea of leaving a manual option (in hacks/advanced settings) would still allow for an automatic default software renderer. If someone needs DX Software for something like shaders, they can switch to it.

**Comment by mirh** [comment]
1. It seems so random (and not-intuitive) to be placed there
2. At this rate we are gonna need a sub-menu for the hacks sub-menu.

**Comment by gregory38** [comment]
Do we have a tooltip to select the renderer ? Maybe it could help people to select. 

Otherwise I don't understand what you try to optimize. Yes we have lots of options but it fit 99.9% of the usage. If you want less option, you need to ditch some use cases.

**Comment by mirh** [comment]
I just feel like 6x1 settings instead of 3x2 (like I suggested above) to be redundant and unoptimal. 
In this latest idea moreover, I don't see which cases would be cut. 

Of course it's by no means a big deal, and I hope you didn't get the impression we were like crazy for this.

**Comment by gregory38** [comment]
```
besides, how's OGL initialization now? Is ivy bridge "warranted"?
```
Not, yet. I need to change the code so it can fallback on DX (11?/9?) when OpenGL isn't supported.

gregory38:
On recent GIT, IvyBridge is enough for OGL SW renderer.

mirh:
Ok so... 
This PR would be golden then now, were it not for the fact DX still has the lead with regards to shaders. 
So, either we fix that (\**giggles*\*) or.. I think the SW mode toggle checkbox idea (that allows one to choose between accuracy and.. "beauty") is better.

turtleli:
#2117 is relevant to this PR.

lightningterror:
The best optimal solution would be to just merge dx9,10,11 without ogl.
If dx11 isn't supported use 10, if 10 isn't supported use 9.

mirh:
Mhh, I dunno, I think there were people explicitly using d3d9 because it was quite faster (of course, in the games where it works). And 10-11 are already the same thing, anyway. 

Said this, we still have the "software opengl isn't yet totally superior to directx" problem. 
Until that's fixed (shaders) the best solution to this is a "sw toggle" checkbox imo.

lightningterror:
I'll be closing this to keep the number of open prs realistic.

We'll be dropping dx9 right after 1.6 is released anyway so there will be less renders anyway.
Being able to select between dx10/11 and ogl sw render is preferred, otherwise there are a couple of new issues that appear.

lightningterror:
I'll reopen this after a discussion with Akash.

PR #1798, opened 2017-01-30
GSdx: Improve robustness of ComboBoxInit()

ssakash:
**Summary of changes**:

* Prevent comboboxes from reaching an indeterminate state **(Refer commit message for more details)**

### DIFF
--- a/plugins/GSdx/GSDialog.cpp
+++ b/plugins/GSdx/GSDialog.cpp
@@ -166,9 +166,16 @@ void GSDialog::SetTextAsInt(UINT id, int i)
 
 void GSDialog::ComboBoxInit(UINT id, const vector<GSSetting>& settings, int32_t selectionValue, int32_t maxValue)
 {
+	if (settings.empty())
+		return;
+
 	HWND hWnd = GetDlgItem(m_hWnd, id);
 
 	SendMessage(hWnd, CB_RESETCONTENT, 0, 0);
+
+	const auto is_present = [=](const GSSetting& x) { return selectionValue == x.value; };
+	if (std::none_of(settings.begin(), settings.end(), is_present))
+		selectionValue = settings.front().value;
 
 	for(size_t i = 0; i < settings.size(); i++)
 	{


plugins/GSdx/GSDialog.h
```diff
@@ -51,7 +51,7 @@ class GSDialog
 	void SetText(UINT id, const char* str);
 	void SetTextAsInt(UINT id, int i);
 
-	void ComboBoxInit(UINT id, const vector<GSSetting>& settings, int32_t selectionValue, int32_t maxValue = INT32_MAX);
+	void ComboBoxInit(UINT id, const vector<GSSetting> settings, int32_t selectionValue, int32_t maxValue = INT32_MAX);
```
gregory38: why did you remove the reference?

plugins/GSdx/GSDialog.cpp
```diff
@@ -164,12 +164,27 @@ void GSDialog::SetTextAsInt(UINT id, int i)
...
+void GSDialog::ComboBoxInit(UINT id, const vector<GSSetting> settings, int32_t selectionValue, int32_t maxValue)
 {
+	if (settings.empty()) return;
+
 	HWND hWnd = GetDlgItem(m_hWnd, id);
+	bool currentsel_found = false;
 
 	SendMessage(hWnd, CB_RESETCONTENT, 0, 0);
 
+	for (const GSSetting i : settings)
```
gregory38: better `const auto& i` or if you want the type `const GSSetting& i`

plugins/GSdx/GSDialog.h
ssakash: I probably overlooked the const reference I guess.

plugins/GSdx/GSDialog.cpp
turtleli: `std::find_if` could be used instead of a for loop.

ssakash: > std::find_if could be used instead of a for loop.

Unless I'm wrong, I'd also have to define an unary function for the third parameter of ``find_if()`` in that case, is that fine?

gregory38: Here some examples (because I played with the texture cache, I don't know if VS2013 support it).

```
plugins/GSdx/GSDevice.cpp-	auto compatible_texture = [&type, &format, &size, &msaa](const GSTexture* t) {
plugins/GSdx/GSDevice.cpp-		return (t->GetType() == type && t->GetFormat() == format && t->GetSize() == size && t->IsMSAA() == msaa);
plugins/GSdx/GSDevice.cpp-	};
plugins/GSdx/GSDevice.cpp-
plugins/GSdx/GSDevice.cpp:	const auto hit = std::find_if(m_pool.begin(), m_pool.end(), compatible_texture);
plugins/GSdx/GSDevice.cpp-
plugins/GSdx/GSDevice.cpp-	if (hit != m_pool.end()) {
plugins/GSdx/GSDevice.cpp-		GSTexture* t = *hit;
plugins/GSdx/GSDevice.cpp-		m_pool.erase(hit);
plugins/GSdx/GSDevice.cpp-		return t;
plugins/GSdx/GSDevice.cpp-	}
```
```
plugins/GSdx/GSTextureCache.cpp-	// Seach first in the same type cache
plugins/GSdx/GSTextureCache.cpp-	auto is_bp = [&bp](const Target* t) { return t->m_TEX0.TBP0 == bp; };
plugins/GSdx/GSTextureCache.cpp:	const auto hit = std::find_if(m_dst[type].begin(), m_dst[type].end(), is_bp);
plugins/GSdx/GSTextureCache.cpp-
plugins/GSdx/GSTextureCache.cpp-	if (hit != m_dst[type].end()) {
plugins/GSdx/GSTextureCache.cpp-		dst = *hit;
```

ssakash: > Here some examples (because I played with the texture cache, I don't know if VS2013 support it).

I think lambda expressions were supported even at the VS2010 versions, so I _think_ it should compile.

plugins/GSdx/GSDialog.cpp
```diff
@@ -166,10 +166,18 @@ void GSDialog::SetTextAsInt(UINT id, int i)
 
 void GSDialog::ComboBoxInit(UINT id, const vector<GSSetting>& settings, int32_t selectionValue, int32_t maxValue)
 {
+	if (settings.empty()) return;
+
 	HWND hWnd = GetDlgItem(m_hWnd, id);
 
 	SendMessage(hWnd, CB_RESETCONTENT, 0, 0);
 
+	auto is_present = [&selectionValue](const GSSetting x) { return (selectionValue == x.value); };
```
turtleli: GSSetting should be passed by reference.
The parenthesis around `selectionValue == x.value` is unnecessary.
(`selectionValue` could be passed by value `[=]`.)

plugins/GSdx/GSDialog.cpp
```diff
@@ -166,10 +166,18 @@ void GSDialog::SetTextAsInt(UINT id, int i)
...
+	bool currentsel_found = std::find_if(settings.begin(), settings.end(), is_present) != settings.end();
```
turtleli: (The bool isn't necessary if the condition is just used directly in the if statement condition.)

ssakash:
> (The bool isn't necessary if the condition is just used directly in the if statement condition.)

I think the bool variable's name makes the code self explanatory, I would prefer to have it.

```
const bool currentsel_found = std::find_if(settings.begin(), settings.end(), is_present) != settings.end();
if (!currentsel_found)
    selectionValue = settings.front().value;
```

**VS**

```
if(std::find_if(settings.begin(), settings.end(), is_present) == settings.end())
     selectionValue = settings.front().value;
```

gregory38:
Well if you read the line you have `...find_if .... is_present... == ... end...`, it seems enough to understand that you check an entry isn't present in the container.

plugins/GSdx/GSDialog.cpp
gregory38: Hum, potentially a reference is easier for the compiler to optimize/inline the function.

ssakash:
> Well if you read the line you have ...find_if .... is_present... == ... end..., it seems enough to understand that you check an entry isn't present in the container.

To be honest, I still think it's confusing without the variable name but _I don't feel like arguing right now_.

turtleli:
Actually, now that I think about it, `std::none_of` might be a better fit and should be more readable.

ssakash:
>Hum, potentially a reference is easier for the compiler to optimize/inline the function.

Oops, didn't notice your comment earlier. So what's the conclusion - pass by _reference_ **(or)** _value_?

gregory38:
Well we don't really care. Do by value. (A const ref is better than a ref)

plugins/GSdx/GSDialog.cpp
```diff
@@ -166,10 +166,17 @@ void GSDialog::SetTextAsInt(UINT id, int i)
...
 void GSDialog::ComboBoxInit(UINT id, const vector<GSSetting>& settings, int32_t selectionValue, int32_t maxValue)
 {
+	if (settings.empty())
+		return;
+
 	HWND hWnd = GetDlgItem(m_hWnd, id);
 
 	SendMessage(hWnd, CB_RESETCONTENT, 0, 0);
 
+	const auto is_present = [=](const GSSetting& x) { return selectionValue == x.value; };
```
gregory38: Oh I didn't understand that `[=]` was enough to catch all variables by value. It is much nicer this way.

gregory38:
For me it is fine.

PR #1799, opened 2017-01-31
PCSX2: Remove Unused And Unneeded Argument.

Huud:
rdd is neither used, nor needed. It appears it was there to pass the \_Rd_ word to write to, but the writing was moved to PHMSBH() to have one "if (\_Rd_)".

Also I'm not sure the "cpuRegs.(LO/Hi).UL[dd+1] = ~firsttemp;" lined are needed, they shouldn't be changed according to the manual. Is there a reason for it, or should it be removed?


gregory38:
The recompiler uses the code in pcsx2/x86/iMMI.cpp
```
////////////////////////////////////////////////////
//upper word of each doubleword in LO and HI is undocumented/undefined
//it contains the NOT of the upper multiplication result (before the substraction of the lower multiplication result)
void recPHMSBH()
```

Huud:
God, this undocumented stuff is everywhere, and there is no way to know if something is an error, or if it is just an undocumented step. 

There is no way an unused argument is not an error, right? ... RIGHT?

gregory38:
Thanks you. I merged your code.

PR #1804, opened 2017-02-05
Make fps2bios compilable with latest PS2 toolchain

uyjulian:
Make fps2bios compilable with latest PS2 toolchain.  
It does not work, currently. PCSX2 crashes when trying to use this replacement PS2 ROM.


gregory38:
I need to check the init (or sif DMA). I don't remember well but I spot a bug in the past. It could explain why nothing work.

gregory38:
I didn't have time to look at the init. But you're code can still be merged. Thanks.

PR #1807, opened 2017-02-08
gsdx tc: avoid calling splice if not needed

AlessandroVetere:
Micro optimization of the MRU algorithm in 2 places: avoid a method call (begin()) for each hit if the splice should be done and 2 method calls (splice() and begin()) if the splice should not be done, at the cost of an extra branch and an extra local variable (the saved iterator).

### DIFF
--- a/plugins/GSdx/GSTextureCache.cpp
+++ b/plugins/GSdx/GSTextureCache.cpp
@@ -199,8 +199,9 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 
 	list<Source*>& m = m_src.m_map[TEX0.TBP0 >> 5];
 
-
-	for(list<Source*>::iterator i = m.begin(); i != m.end(); ++i)
+	list<Source*>::iterator m_begin = m.begin();
+	
+	for(list<Source*>::iterator i = m_begin; i != m.end(); ++i)
 	{
 		Source* s = *i;
 
@@ -222,7 +223,10 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 				continue;
 		}
 
-		m.splice(m.begin(), m, i);
+		// Move pointer in front of the list if not already first
+		if (m_begin != i) {
+			m.splice(m_begin, m, i);
+		}
 
 		src = s;
 
@@ -442,13 +446,18 @@ GSTextureCache::Target* GSTextureCache::LookupTarget(const GIFRegTEX0& TEX0, int
 
 	Target* dst = NULL;
 
-	for(list<Target*>::iterator i = m_dst[type].begin(); i != m_dst[type].end(); ++i)
+	list<Target*>::iterator m_dst_type_begin = m_dst[type].begin();
+	
+	for(list<Target*>::iterator i = m_dst_type_begin; i != m_dst[type].end(); ++i)
 	{
 		Target* t = *i;
 
 		if(bp == t->m_TEX0.TBP0)
 		{
-			m_dst[type].splice(m_dst[type].begin(), m_dst[type], i);
+			// Move pointer in front of the list if not already first
+			if (m_dst_type_begin != i) {
+				m_dst[type].splice(m_dst_type_begin, m_dst[type], i);
+			}
 
 			dst = t;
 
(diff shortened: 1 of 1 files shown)


plugins/GSdx/GSTextureCache.cpp
```diff
@@ -199,8 +199,9 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 
 	list<Source*>& m = m_src.m_map[TEX0.TBP0 >> 5];
 
-
-	for(list<Source*>::iterator i = m.begin(); i != m.end(); ++i)
+	list<Source*>::iterator m_begin = m.begin();
```
ssakash: ``m_`` is generally used to name class members, maybe not a good name for a local variable?

AlessandroVetere: My bad, you are right of course. I'll rename the two local variable now.

gregory38:
I'm rather mitigated. begin() is likely implemented as a move. I'm pretty sure GCC optimize it. And here the code of slice
```
   │1365          void                                                                                                                                                                                                   │
   │1366          splice(iterator __position, list& __x, iterator __i)                                                                                                                                                   │
   │1367    #endif                                                                                                                                                                                                       │
   │1368          {                                                                                                                                                                                                      │
   │1369            iterator __j = __i._M_const_cast();                                                                                                                                                                  │
   │1370            ++__j;                                                                                                                                                                                               │
  >│1371            if (__position == __i || __position == __j)                                                                                                                                                          │
   │1372              return;                                                                                                                                                                                            │
   │1373                                                                                                                                                                                                                 │
   │1374            if (this != &__x)                                                                                                                                                                                    │
   │1375              _M_check_equal_allocators(__x);                                                                                                                                                                    │
   │1376                                                                                                                                                                                                                 │
   │1377            this->_M_transfer(__position._M_const_cast(),                                                                                                                                                        │
   │1378                              __i._M_const_cast(), __j);                                                                                                                                                         │
   │1379          }        
```

Code is inlined. So potentially you save 2/3 moves and 1 cmp (affectation and the ++ iterator). IMHO, I'm not sure it worth the increase of code complexity.

AlessandroVetere:
I don't see that much increase in code complexity. Also removing unnecessary method calls reduces jumps and pressure on the registers.

gregory38:
Well it isn't a huge increase. But it is still an increase. And actually if you're not in first position you get extra jobs (the comparison is done twice).
```
Also removing unnecessary method calls reduces jumps and pressure on the registers.
```
But there is no method call. Sure in C++ you have function but in the ASM nothing is called, the code is inlined. So the code (ASM) already does what you're trying to do.

AlessandroVetere:
You are right, but I wouldn't make choices based on the fact that I'm inlining a library function implementation that I don't possess full control on, if it isn't costly to do so, but that's just my opinion :)
Given that, I'll look into some benching, it's also interesting to know which is the distribution of the number of the iterations of the for loops involved, for further investigations.

gregory38:
IMHO, you need to trust the standard library and the compiler.

AlessandroVetere:
I seriously do, that's the reason why I'm not asking the library nor the compiler to do nothing more than they should.

AlessandroVetere:
I did some statistical checks in the opening scene (after the cutscene) of ZoE2 as it is full of particles and textures: I reported each time whether branch was taken or not, and I also reported the wasted for loop iterations in LookupSource and LookupTarget (the situation is quite bad in some cases, I would say when dealing with particles, much better in other cases).
I'll attach two txt files containing the console output, it's about 1 frame in game, and the modified GSTextureCache file.
[Stats.txt](https://github.com/PCSX2/pcsx2/files/762476/Stats.txt)
[Stats2.txt](https://github.com/PCSX2/pcsx2/files/762477/Stats2.txt)
[GSTextureCache.txt](https://github.com/PCSX2/pcsx2/files/762479/GSTextureCache.txt)
Long story short: in average the branch is taken (so the splice is called) 1 times out of 6, in this particular case.

gregory38:
I will post you tonight a screenshot of Vtune. Maybe you will understand that you're trying to micro-optimize a nop.

AlessandroVetere:
It's fine, I'll take a look at Vtune by myself, but at this point optimizing 1,2 or 3 ASM instructions means little to nothing as the majority of the time in the code is passed looping.

gregory38:
I will close this PR as for me it doesn't worth it.

PR #1808, opened 2017-02-08
GSdx-PCRTC: Minor changes to display rectangle setup

ssakash:
**Summary of changes**:

* Isolate all the display rectangle setup hacks to a separate function to make ``GetDisplayRect()`` more readable.
* Allow height saturation only for Interlaced (Field) mode, we shouldn't be doing it for double scan mode. (**_Refer _commit message_ for more details_**)

### DIFF
--- a/plugins/GSdx/GSState.h
+++ b/plugins/GSdx/GSState.h
@@ -243,6 +243,7 @@ public:
 	void ResetHandlers();
 
 	int GetFramebufferHeight();
+	void SaturateOutputSize(GSVector4i& r);
 	GSVector4i GetDisplayRect(int i = -1);
 	GSVector4i GetFrameRect(int i = -1);
 	GSVideoMode GetVideoMode();
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -405,6 +405,30 @@ GSVideoMode GSState::GetVideoMode()
 	return videomode;
 }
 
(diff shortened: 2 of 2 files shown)


gregory38:
I'm fine with the code.

plugins/GSdx/GSState.h
```diff
@@ -243,6 +243,7 @@ class GSState : public GSAlignedClass<32>
 	void ResetHandlers();
 
 	int GetFramebufferHeight();
+	void SaturateOutputSize(GSVector4i*);
```
turtleli: Maybe it'd be better to pass by reference?

ssakash: Any reason to pass by reference? As far as I can see, I only see a demerit that the function call wouldn't have ``&`` which was necessary to show that the variable's value might be modified by the function.

gregory38: Otherwise pass by const ref or value. And return the object. And please put a parameter name.

Pointers are error prone. You have no guarantee that `*r` is a `GSVector4i` object it could be anything.

turtleli: Returning the value sounds like a good idea.

> As far as I can see, I only see a demerit that the function call wouldn't have & which was necessary to show that the variable's value might be modified by the function.

For pointer and reference parameters, the presence or absence of `const` is a better indicator of whether a value might be modified. References/pointers are also used in cases where passing by value is inefficient.

ssakash: I went with reference.

Issue #1749, opened 2017-01-07
GSdx: VS2013 generates bad code for _xgetbv(0)


turtleli:
AVX2 builds of GSdx compiled with VS2013 crash on my system when I use the SW renderer. I bisected the crash and found it was introduced by commit 0f5529be18eb0e93cb951925d4e793f5192b8f73, but that made no sense to me.

I looked at the disassembly and found that the generated code was incorrect - it didn't touch the eax register at all after the xgetbv instruction. However, changing the xbyak code from:
```
uint64 bv = getXfeature();
if ((bv & 6) == 6) {
```
to
```
if ((_xgetbv(0) & 6) == 6) {
```
so that _xgetbv(0) is directly used fixes the code generation issue. Fun.

I'm making a note of this instead of fixing it. One other thing that we've used that VS2013 doesn't support is %z length modifiers in printfs. I'll leave fixing this stuff to someone who wants to keep VS2013 support.

mirh:
[Update 5](https://bugs.chromium.org/p/libyuv/issues/detail?id=529#c4)?

turtleli:
I was using update 5.

mirh:
Mhh.. and seems like microsoft [stopped](https://connect.microsoft.com/VisualStudio/Feedback/Details/981479) to care about it long time ago.

gregory38:
Tbh I was surprised that %z compile. Which is dumb for a runtime feature. Anyway what is the status of VS2017 ? They released the RC recently but I didn't find any info on the final version.

mirh:
https://connect.microsoft.com/VisualStudio/feedback/details/806338/vc-printf-and-scanf-should-support-z
https://connect.microsoft.com/VisualStudio/feedback/details/2083820/printf-format-specifier-zu-is-supported-but-not-documented

If %zu is comparable, then they kind of support this already in VS15.

gregory38:
AFAIU, the issue is VS2013. We said in the past that we will try to support at least the 2 latest versions of VS.  It seems VS2017 will be released soon. I'm not sure, it worth to bother with VS2013. Besides based on their own developers, it is likely far of the C++ standard
* https://www.reddit.com/r/cpp/comments/5dai1i/vs_2017_rc_is_now_available/da30hxd/

@refractionpcsx2 any opinion on the matter ?

avih:
As this is easy to fix (and the `%z` is probably less critical), why not just fix it?

gregory38:
[...]

PR #1660, opened 2016-11-13
LilyPad: Add PS mouse support

FlatOutPS2:
Adds Playstation Mouse as a new pad type.

For use with PS1 games only.

Issue #384, opened 2014-12-08
[wx] dlopen: cannot load any more object with static TLS


micove:
PCSX2 version: 1.3.0 GIT 9e8b0b5

Okay last issue.

Issue:
If I remove the configure folder to setup a clean 1st time configuration on in general change plugins.

I tend to get a lot of
[wx] dlopen: cannot load any more object with static TLS
(p) (L) The configured SPU2 plugin file is not a valid dynamic library(LoadCorePlugins) 

Checking the plugins I see
$ readelf -W -a libzzogl-0.4.0.so | grep STATIC_TLS
 0x0000001e (FLAGS)                      TEXTREL STATIC_TLS
$ readelf -W -a libspu2x-2.0.0.so | grep STATIC_TLS
 0x0000001e (FLAGS)                      TEXTREL STATIC_TLS

Since we are using a global resource it gets exhausted and fails to load the plugins with dlopen. 

More info in
[1]https://bugzilla.redhat.com/show_bug.cgi?id=1124987
[2]https://www.cygwin.com/ml/libc-help/2013-11/msg00033.html

In [2] there are some suggestions to solve this and I'm not sure asking to bump DTV_SURPLUS in glibc is going to work. Not sure how feasible is to dynamically link to them instead of dlopen or to just pack the plugins inside the binary like dolphin-emu does now.

This is kind of minor issue since with a little fiddling the plugins get loaded.

gregory38:
zzogl support is dropped so no way to toggle the GS plugin anymore. Therefore no more issue :)

I will close the bug report. If you aren't happy with current situation, please reopen it.

PR #1812, opened 2017-02-09
Replace __threadlocal with local_thread

KousukeItsagame:
Resubmitting


KousukeItsagame:
apparently failes because of a different HEAD?

KousukeItsagame:
Done, but only 4 files are changed

KousukeItsagame:
yay ;d

gregory38:
Tbh, I don't see the point. When we drop VS2013, we could remove some lines. But here, there is no gain currently. Or did I miss an improvement ?

KousukeItsagame:
No, There hasn't changed anything since last time
I completed the patch after dropping it a while ago.
That's why I posted it :)
It's not going to be valid for long I know :)
But I"m learning stuff along the way :)

It's no gain, it's an alternative.

gregory38:
IMHO, it would be much better to keep the __threadlocal define instead to put ifdef everywhere. Then you can define  __threadlocal with the new C++11 keyword if you want.

KousukeItsagame:
Ok, fine for me
thanks for the good talk :)

mirh:
I don't even know what TLS is, but.. are you sure about that _MSC_VER *equal* 1800? Why not >=?
If it's just about differentiating between windows and linux (gcc right?) shouldn't you just check for _WIN32?

And if not, and you truly need a specific vs version.. you sure to want to check for **IDE** version instead of toolset one?

gregory38:
TLS stands for "thread local storage". Which allow to have a global variable **by** thread instead of a global variable for **all** threds.

All compilers have extension to support TLS. C++11 decides to give a standard keyword for all implementations (nothing change in the implementation).

gregory38:
I propose to coordinate with #1749

In short we might drop support of VS2013 in a couple of weeks.

gregory38:
@KousukeItsagame 
VS2013 support was dropped. You can edit your PR to avoid the various ifdef. And directly use the `C++11` keyword.

KousukeItsagame:
I'm setting up my development environment again
After that all change the patch :)

KousukeItsagame:
"PR" yeah XD

KousukeItsagame:
can I do the clang formatting in cygwin or mingw?
I have problems installing vs studio 2015 with visual studio 2017

turtleli:
Probably? I installed LLVM from http://releases.llvm.org/ , which includes clang-format (since the plugin doesn't work on VS2017).

gregory38:
IMHO, it could be easier to redo the PR on a recent master. You need to redo it anyway to drop VS2013 support and remove the thread local related define

KousukeItsagame:
it's almost done,
forgot to delete some spaces and an the .git file

common/src/Utilities/wxAppWithHelpers.cpp
```diff
@@ -678,7 +678,7 @@ wxAppWithHelpers::wxAppWithHelpers()
     // FS segment register won't have been initialized by the main exe, due to tls_insurance
     // being optimized away >_<  --air
 
-    static __threadlocal int tls_insurance = 0;
+    static local_thread int tls_insurance = 0;
```
gregory38: Do you mean thread_local ?

common/include/Pcsx2Defs.h
```diff
@@ -168,7 +168,7 @@ static const int __pagesize = PCSX2_PAGESIZE;
 #define PCSX2_ALIGNED16_EXTERN(x) extern __declspec(align(16)) x
 
 #define __noinline __declspec(noinline)
-#define __threadlocal __declspec(thread)
+#define local_thread __declspec(thread)
```
gregory38: IIRC, this define is useless

common/include/Pcsx2Defs.h
```diff
@@ -216,7 +216,7 @@ static const int __pagesize = PCSX2_PAGESIZE;
 #ifndef __noinline
 #define __noinline __attribute__((noinline))
 #endif
-#define __threadlocal __thread
+#define thread_local __thread
```
gregory38: And likely this define too

gregory38:
@KousukeItsagame Please find my review comments above.

@turtleli 
Any opinion on removing `PCSX2_THREAD_LOCAL` which is always `1` and the related `Threading::TlsVariable`, `Threading::BaseTlsVariable`.

common/include/Pcsx2Defs.h
KousukeItsagame: yeah, I checking it again, I probably mixed up some

KousukeItsagame:
sorry, I did mix it up. it's confusing

KousukeItsagame:
there is a define local_thread from pcsx2 I think and the c++11 version is backwards

common/include/Pcsx2Defs.h
KousukeItsagame: i should learn to be more thorough

KousukeItsagame: I  don't know why it's useless?

turtleli:
> Any opinion on removing PCSX2_THREAD_LOCAL which is always 1 and the related Threading::TlsVariable, Threading::BaseTlsVariable.

I'm for removing it.

common/include/Pcsx2Defs.h
```diff
@@ -168,7 +168,7 @@ static const int __pagesize = PCSX2_PAGESIZE;
 #define PCSX2_ALIGNED16_EXTERN(x) extern __declspec(align(16)) x
 
 #define __noinline __declspec(noinline)
-#define __threadlocal __declspec(thread)
+#define thread_local __declspec(thread)
```
turtleli: Why are you redefining a keyword?

common/include/Pcsx2Defs.h
turtleli: Ditto.

common/include/Pcsx2Defs.h
KousukeItsagame: I searched for all instances, and changed them basically

KousukeItsagame: or do you mean why i'm changing it?

turtleli: You shouldn't redefine a keyword unless there's a need to. You should have just removed the line.

gregory38: @KousukeItsagame 
In case you don't know how define work. Otherwise you can skip this comment.
```
#define word1 word2
```
Mean to the compile to replace occurance of word1 by word2. In the previous code the virtual keyword `__threadlocal` was replaced by the compiler extension `__declspec(thread)`. So it allow to use
```
#define __threadlocal __declspec(thread)
__threadlocal my_var;
```
and the compiler see it as
```
__declspec(thread) my_var;
```
 As you notice C++11 uses a standard name `thread_local` so we don't need anymore the intermediate keyword. You can now use it directly. Hence
```
thread_local my_var;
```

gregory38: @KousukeItsagame 
So could you remove the `__threadlocal`/`thread_local` from this file.

gregory38: @blazekousuke @KousukeItsagame 
Any update ? If you're too busy, I can still amend your PR.

gregory38:
Ok it is much better. Ideally squashing all commits together will be nice but I'm afraid of various conflicts.

@FlatOutPS2 
I will be glad if you can explain me how merge/rebase/squash properly.

turtleli:
In this case I think you can squash locally with
```
git fetch upstream pull/1812/head:branch_to_merge
git merge --squash branch_to_merge
git commit --author="youknowthispartalready"
```

danilaml:
Or you can use squash merge feature of GitHub. Though there will be some garbage (repeated commit messages) in the commit description.

turtleli:
> Or you can use squash merge feature of GitHub.

I had issues with that once (#1660). So I'm personally a bit wary of using it.

gregory38:
I know the button ;) But there are conflicts to resolve, so no magical button for me ;)

@turtleli do you approve all changes by the way ?

turtleli:
Yeah, seems ok (if I ignore the mass of commits).

gregory38:
@turtleli Actually I set PCSX2_THREAD_LOCAL to 0 when I build the library for spu2x. It removes some requirements for limited TLS variable slot. I don't remember well.

gregory38:
Branch was squashed and merged to master https://github.com/PCSX2/pcsx2/commit/11aebe465f4a879490b98e7211ac6e061e2139cb

gregory38:
It was isssue #384

Issue #1813, opened 2017-02-09
Support llvmpipe on windows. Works around #345 and #1716


pal1000:
***PCSX2 options:***
``All default, Gsdx OpenGL (mainly software, hardware would be too slow)``

***Plugins used:***
``All default. Gsdx SSE4.1. AVX2 not supported by CPU.``

***Description of the issue:***
``Gsdx fails to initialize. It cannot create GL 3.x context. It appears GSdx cannot detect Core profile at all:``

```
PCSX2 1.5.0-20170208194254 - compiled on Feb  9 2017
Savestate version: 0x9a0d0000

Host Machine Init:
	Operating System =  Microsoft Windows 10 Pro (build 14393), 64-bit
	Physical RAM     =  3975 MB
	CPU name         =  Intel(R) Core(TM) i3-2375M CPU @ 1.50GHz
	Vendor/Model     =  GenuineIntel (stepping 07)
	CPU speed        =  1.496 ghz (4 logical threads)
	x86PType         =  Standard OEM
	x86Flags         =  bfebfbff 1dbae3bf
	x86EFlags        =  28100000

x86 Features Detected:
	SSE2.. SSE3.. SSSE3.. SSE4.1.. SSE4.2.. AVX

Reserving memory for recompilers...

Loading plugins...
	Binding   GS: C:\Software\Development\projects\pcsx2\bin\plugins\GSdx32-SSE4.dll 
	Binding  PAD: C:\Software\Development\projects\pcsx2\bin\plugins\LilyPad.dll 
	Binding SPU2: C:\Software\Development\projects\pcsx2\bin\plugins\Spu2-X.dll 
	Binding CDVD: C:\Software\Development\projects\pcsx2\bin\plugins\cdvdGigaherz.dll 
	Binding  USB: C:\Software\Development\projects\pcsx2\bin\plugins\USBnull.dll 
	Binding   FW: C:\Software\Development\projects\pcsx2\bin\plugins\FWnull.dll 
	Binding DEV9: C:\Software\Development\projects\pcsx2\bin\plugins\DEV9null.dll 
Plugins loaded successfully.

(GameDB) 9709 games on record (loaded in 455ms)
HLE Notice: ELF does not have a path.


Initializing plugins...
	Init GS
	Init PAD
	Init SPU2
	Init CDVD
	Init USB
	Init FW
	Init DEV9
Plugins initialized successfully.

Patches: No CRC found, using 00000000 instead.
Opening plugins...
[...]

PR #1814, opened 2017-02-10
gsdx wgl: retry to create a context without advance flag

gregory38:
Hopefully it will support more drivers #1813 

Need to check compilation and the above issue.

### DIFF
--- a/plugins/GSdx/GSWndWGL.cpp
+++ b/plugins/GSdx/GSWndWGL.cpp
@@ -45,7 +45,7 @@ LRESULT CALLBACK GSWndWGL::WndProc(HWND hWnd, UINT message, WPARAM wParam, LPARA
 }
 
 
-bool GSWndWGL::CreateContext(int major, int minor)
+void GSWndWGL::CreateContext(int major, int minor)
 {
 	if ( !m_NativeDisplay || !m_NativeWindow )
 	{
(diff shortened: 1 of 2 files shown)


plugins/GSdx/GSWndWGL.cpp
```diff
@@ -88,13 +88,20 @@ bool GSWndWGL::CreateContext(int major, int minor)
...
+			throw GSDXRecoverableError();
 		}
 
 		HGLRC context30 = wglCreateContextAttribsARB(m_NativeDisplay, NULL, context_attribs);
 		if (!context30) {
-			fprintf(stderr, "Failed to create a 3.x context\n");
-			return false;
+			fprintf(stderr, "Failed to create a 3.x context with standard flags\n");
+			// retry with more compatible option for (Mesa on Windows, OpenGL on WINE)
+			context_attribs[2*2+1] = 0;
```
np511: Any specific reason you use `2*2+1` and not `5`?

gregory38: yes because there aren't the same. 5 is 6th postision. 2*2 + 1 is the 3rd row, 2nd column.

np511:
> context_attribs[2*2+1] = 0;

>

> yes because there aren't the same. 5 is 6th postision. 2*2 + 1 is the 3rd

> row, 2nd column.



How? Isn't `2*2+1` evaluated before [], and 3rd row 2nd column = `[4][1]`?

PR #1821, opened 2017-02-13
GSDX-Windows: Automatic initial CRC hack level selection

### LINKED ISSUE #1722: GSDX Setup Tuning: Initial setup uses sub-optimal CRC level in GSDX
After doing some experimenting, i noticed that when the initial setup of the plugins is done, GSDX defaults to OpenGL Hardware (good) and Full (Safest) CRC level (not so good). 

It would be better in this scenario to default to Partial (OpenGL Recommended).

I could do it but I refuse to touch GSDX and I know how much @ssakash loves it :P

ssakash:
**Summary of changes**:

* Add an enum class for ``CRC hack level``.
* Default to ``Partial`` level on a fresh startup if the best renderer is detected as OpenGL. Closes #1722.


gregory38:
I'm fine with the code.

PR #1825, opened 2017-02-14
pcsx2: Fix gzipped ISO reader

### LINKED ISSUE #1820: PCSX2 hangs when trying to boot compressed isos.
***PCSX2 version:***
``1.5.0.r1885.f86ea9f19``

``Trying to run isos compressed as gz, makes PCSX2 hang during boot with high cpu usage. Trying to run uncompressed isos or mounting them using cdemu before running them, works as intended. With the same PCSX2 version on Windows, everything is working fine with compressed isos.``

``I believe it is an issue with the latest zlib, as it was working as intended with 1.2.8. Current version of zlib in Archlinux is 1.2.11 which was updated in January 17``

***PC specifications:***
``Intel i5-4690k, Nvidia GeForce GTX 560 Ti, Archlinux 4.9.8``

turtleli:
Changes:
 * Updates zlib to 1.2.11 on Windows.
 * Fixes the freeze issue on zlib 1.2.9 onwards when using a gzipped ISO.

Fixes #1820.

### DIFF
--- a/3rdparty/zlib/ChangeLog
+++ b/3rdparty/zlib/ChangeLog
@@ -1,10 +1,53 @@
 
                 ChangeLog file for zlib
 
+Changes in 1.2.11 (15 Jan 2017)
+- Fix deflate stored bug when pulling last block from window
+- Permit immediate deflateParams changes before any deflate input
+
+Changes in 1.2.10 (2 Jan 2017)
+- Avoid warnings on snprintf() return value
+- Fix bug in deflate_stored() for zero-length input
+- Fix bug in gzwrite.c that produced corrupt gzip files
+- Remove files to be installed before copying them in Makefile.in
+- Add warnings when compiling with assembler code
+
+Changes in 1.2.9 (31 Dec 2016)
+- Fix contrib/minizip to permit unzipping with desktop API [Zouzou]
+- Improve contrib/blast to return unused bytes
+- Assure that gzoffset() is correct when appending
+- Improve compress() and uncompress() to support large lengths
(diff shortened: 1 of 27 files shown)


mirh:
What do ya think of submodules?

avih:
Hmm.. that was quick. Good job in wrapping your head around the code. I still didn't refresh myself with the original code, so I find it hard to assess the changes by looking at the diff now.

Can you explain the fix in words please?

One thing though, do we keep internal data structures on disk? if yes, then I think we should probably add the zlib version to the index such that we can invalidate it when zlib changes again. There's a version data at the index file, and the code already knows to invalidate it if the index format version changes, I just didn't take into account the zlib version as well.

turtleli:
> Can you explain the fix in words please? Was there an actual bug?

The code was shallow copying the stream structures, but copying streams should only be done with `inflateCopy()` since the streams have internal state to manage (internally, the zlib stream internal data contains a pointer to the stream itself). So the fix is to avoid shallow copying and also use `inflateCopy()` as necessary.

> One thing though, do we keep zlib internal data structures on disk?

Not sure. I don't think it does.

gregory38:
@mirh I hope you don't feel like hijacking a PR ;) May I suggest to open a separate issue/question.

@turtleli 
Great job. It is nice example why C++ is safer with copy-ctor and such.

gregory38:
I'm in favor to merge the code to increase test coverage.

LoathingKernel:
I have been testing it for the past few days and so far it works. I have tested with existing and freshly made indexes.

Issue #1797, opened 2017-01-29
Pressing Alt+Tab disables binded save loading.


ReyVGM:
***PCSX2 version:***
pcsx2-v1.5.0-dev-1749-g9d1b27c-windows-x86

***PCSX2 options:***
Everything is default, except the "PCSX2_keys.ini"

***Description of the issue:***
If you press Alt+Tab to switch to another screen, then loading a save state through a bound key gets disabled while the emulator is running.

***How to reproduce the issue:***
This has been happening ever since I started using the emulator, but I could never figure out why it happened. But I finally found out today.
I have save state saving/loading to F2/F4. But whenever you press Alt+Tab to switch to another window, and then come back to the emulator, you can no longer load the save state through the bound key if the emulator is active. If the emulator is _paused_, then the bound save load works. (It also works if you load through the menu too.) 

I attached my ini just in case that is causing the issue.

[PCSX2_keys.zip](https://github.com/PCSX2/pcsx2/files/737877/PCSX2_keys.zip)

FlatOutPS2:
It's not your ini. The same thing happens to the F4 key with default settings. After you alt-tab the F4 key no longer functions for its assigned purpose.

Hirato:
Just tap that same alt button again.
It gets stuck in the PRESSED state because it doesn't get the RELEASE event because of the focus switch.

PR #1831, opened 2017-02-16
LilyPad: Updates

FlatOutPS2:
Adds 4 new commits:

- Improve analog/pressure sensitive detection:

Improves detection of analog or pressure sensitive support, which previously made some games unable to detect the correct mode.

- Add Quick Setup:

Adds quick setup that allows for quickly setting up all the default pad
buttons without having to click on each one separately.

- Display a warning when all controls are inactive:

Displays a console warning if no controls/controllers are active and the
emulation cannot be controlled.

- General fixes and UI improvements

Some general fixes and UI improvements. And adds Restore Defaults option.


plugins/LilyPad/LilyPad.cpp
```diff
@@ -1580,6 +1580,9 @@ keyEvent *CALLBACK PADkeyEvent()
 #ifdef _MSC_VER
     static char shiftDown = 0;
     static char altDown = 0;
+    if (!activeWindow)
+        altDown = shiftDown = 0;
```
turtleli: Is this related to #1797?

FlatOutPS2: Yes, I forgot to add that to the commit message. It fixes the issue.

plugins/LilyPad/LilyPad.cpp
```diff
@@ -1201,14 +1206,19 @@ u8 CALLBACK PADpoll(u8 value)
         return query.response[++query.lastByte];
     }
 
-    int i;
     Pad *pad = &pads[query.port][query.slot];
+    int padtype = config.padConfigs[query.port][query.slot].type;
+
+    if (padtype != pad->previousType) {
```
FlatOutPS2: There are no other situations this currently occurs under, and any new situation would be handled elsewhere, so I've moved it to PADopen.

plugins/LilyPad/LilyPad.cpp
```diff
@@ -531,6 +531,8 @@ void Update(unsigned int port, unsigned int slot)
     dm->Update(&info);
     static int rapidFire = 0;
     rapidFire++;
+    static bool anyDevActiveAndBound = true;
+    bool currentDevActiveAndBound = false;
```
FlatOutPS2: No, it shouldn't be moved as a new pad can (dis)connect at other times than just when PADopen is triggered.

I actually used dev because of the code below `Device *dev`, but I've changed it to avoid the confusion.

FlatOutPS2:
> I'm not sure it's a good idea having an additional "page" for special inputs. The dialog could always be made wider.

I don't want to make the dialog wider. As far as I'm concerned it's as wide as it should be. And there will be more special inputs added later that would further increase the required space. It also avoids confusion between the default pad inputs and any extra inessential ones.

plugins/LilyPad/LilyPad.cpp
turtleli: This change should probably be in a separate commit.

plugins/LilyPad/LilyPad.cpp
turtleli: What about loading a savestate?

turtleli:
Another thing - the "General fixes and UI improvement" commit message is lacking a bit of detail. It doesn't mention the addition of the "restore defaults" button, nor does it mention the double click change.

> I don't want to make the dialog wider. As far as I'm concerned it's as wide as it should be. And there will be more special inputs added later that would further increase the required space. It also avoids confusion between the default pad inputs and any extra inessential ones.

Personally I still think it could it be wider. But anyway, I think it would probably be nicer to have a tab or something instead of a "Go to Special Inputs button" and "Back to Controls" button.

FlatOutPS2:
@turtleli I've changed the Special Inputs, added the messing details to the commit message and made a seperate commit for the alt-tab issue fix as requested.

plugins/LilyPad/LilyPad.cpp
```diff
@@ -1217,23 +1227,14 @@ u8 CALLBACK PADpoll(u8 value)
         switch (value) {
             // CONFIG_MODE
             case 0x43:
-                if (pad->config) {
```
turtleli: I don't get why you moved the code here into an if/else in the following switch case. Couldn't it just be:
```
case 0x43:
    if (pad->config && padtype != neGconPad) {
        // blah blah
        return whatever;
    }
    // Fallthrough
case 0x42:
```

plugins/LilyPad/Config.cpp
```diff
@@ -1393,8 +1394,13 @@ int BindCommand(Device *dev, unsigned int uid, unsigned int port, unsigned int s
     if (port > 1 || slot > 3 || padtype >= numPadTypes)
         return -1;
 
-    if (!sensitivity)
-        sensitivity = BASE_SENSITIVITY;
+    if (!sensitivity) {
+        if (((uid >> 16) & 0xFF) == ABSAXIS) {
+            sensitivity = 87183;
```
FlatOutPS2: 87183 is the default sensitivity for the analog sticks in default.ini. But if you create a new binding for analog sticks you always got BASE_SENSITIVITY(which results in 65536) in previous LilyPad versions.

plugins/LilyPad/Config.cpp
```diff
@@ -2355,6 +2412,21 @@ INT_PTR CALLBACK GeneralDialogProc(HWND hWnd, unsigned int msg, WPARAM wParam, L
...
                 PropSheet_Changed(hWndProp, hWnd);
+            } else if (HIWORD(wParam) == BN_CLICKED && LOWORD(wParam) == ID_RESTORE_DEFAULTS) {
+                int msgboxID = MessageBoxA(hWndProp, "This will delete all current settings and revert back to the default settings of LilyPad. Continue?",
+                                           "Restore Defaults Confirmation", MB_YESNO | MB_DEFBUTTON2 | MB_ICONEXCLAMATION);
+                switch (msgboxID) {
+                    case IDNO:
+                        break;
+                    case IDYES:
+                        static char iniLocation[MAX_PATH * 2] = "inis/LilyPad.ini";
+                        remove(iniLocation);
```
FlatOutPS2: No, it didn't convert properly.

plugins/LilyPad/Config.cpp
turtleli: You should probably use a constant to represent this (maybe DEFAULT_ANALOG_SENSITIVITY or any other suitable name).

plugins/LilyPad/Config.cpp
turtleli: Ok. Does iniLocation need to be static?

plugins/LilyPad/LilyPad.cpp
```diff
@@ -1342,7 +1350,7 @@ u8 CALLBACK PADpoll(u8 value)
             // QUERY_DS2_ANALOG_MODE
             case 0x41:
                 // Right?  Wrong?  No clue.
-                if (pad->mode == MODE_DIGITAL || pad->mode == MODE_PS1_MOUSE) {
+                if (pad->mode == MODE_PS1_MOUSE || pad->mode == MODE_NEGCON) {
```
turtleli: `pad->mode == MODE_DIGITAL` has been removed. Is this intentional?

Apart from that, the code seems fine.

FlatOutPS2: Yes, that is very much intentional. It was one of the reasons why some game didn't enable analog/pressure sensitive mode.

pgert:
One of these commits caused "Network Access Disc" to become unresponsive to DS2 input.

Issue #954, opened 2015-11-06
[VU0] Street Fighter EX3 - Broken cloth physics


rz5:
**_PCSX2 versions used:**_
Pcsx2 0.9.6
Pcsx2 Beta 976
pcsx2-v1.3.1-1378-g8ab9cea-windows-x86

**_Emulation settings:**_ "Safest" preset, no hacks.

**_Plugins used:**_ 
GSdx-AVX2 
GSdx-SSE2 version 1.15
GSdx-SSE2 version 1.14
ZeroGS 97.1

**_Plugin settings:**_ CRC Hack Level "None"
#### Hardware Specifications

**_CPU:**_ 
AMD Athlon II 64 X4 @ 3.20GHz
Intel® Core™ i5-4690K @ 3.50GHz

**_GPU:**_ 
GeForce 9600GT
GeForce GTX 970 (Driver: NVIDIA 358.50)
#### Description

The dynamic cloth physics of the game are broken. Cloth items that are supposed to flow and instead
they are static and messed up.

This bug is not affected by EE/VU clamp/round mode, superVU and microVU are the same.
HW and SW renderers don't influence it either.
None of the gamefixes affect the bug.

Setting VU0 to "Interpreter" changes the bug slightly (the hakama isn't a disc anymore, but
it is still static and slanted to the left).
Due to this, I prefixed this issue with [VU0].
#### Screenshots

Character on the left incorrectly has the shirt bunched up and the neck scarf to the left.
Character on the right is supposed to be wearing some hakama pants; instead they appear as flat static discs attached to the character's waist.
![image:1](http://forums.pcsx2.net/attachment.php?aid=17335)

Another screenshot of the character with bugged hakama pants.
![image:1](http://forums.pcsx2.net/attachment.php?aid=17336)

refractionpcsx2:
This is a long standing bug, I don't think it's ever been solved, not sure why lol

gregory38:
Fixed with the above PR.

PR #1835, opened 2017-02-23
VU0: fix for Street Fighter EX3 and R: Racing Evolution

volodymyrkutsenko:
Hi guys!

Remember the *Street Fighter EX3* clothing issue #954 and this thread *http://forums.pcsx2.net/Thread-R-Racing-Evolution-invisible-cars* describing the lack of the cars on the race track in R Racing Evolution?
Well, the dark days for these games seem finally to be over as there is a fix for them!

sfex3
rri

In short, there is something peculiar about the CFC2 instruction (VU0 Macro) when it copies the value from the TPC register.
These two mentioned games use the code like below for running the VU0 micro subroutine:

...
cfc2	t4, TPC
ctc2	t4, CMSAR0
callmsr
...

The official VU coding manual requires the value in the CMSAR0 to be divided by 8 before running the VCALLMSR instruction.
Apparently CFC2 divides the value by 8 itself because all the other games I have found also using the VCALLMSR instruction (24: The Game, GTA LCS, GTA VCS and FFXII) indeed divide the value by 8 in code before running the micro subroutine:

(From the FFXII game)

...
lw		v0, 0x14(v1)
sra		v0, 0x03
ctc2	v0, CMSAR0
...
callmsr	
...

The only difference is that they never fetch the value from the TPC directly.

I have also run some homebrew tests on my PS2 and discovered that the CFC2 instruction really ends up with the value already divided by 8.
There are basically two possibilities: either CFC2 also divides by 8 while fetching from the TPC register or the TPC register always works with the values already divided by 8.
The latter seems unlikely because TPC is the Program Counter register that should hold the memory address of the instruction.
Sadly there seems to be no way to see what is really inside the TPC register before fetching from it by CFC2 because CFC2 is the only way to retrieve its value (so says the manual).
In any case PCSX2 already implements TPC as the instruction address pointer so sticking to CFC2 dividing by 8 looks like the correct way for the fix.

(I left a descriptive comment directly in code describing the same, fixed for both the EE Interpreter and the Recompiler)

It was possible for me to test some games apart from all of the mentioned above and there _seem_ to be no regression.
However this needs to be tested on a bigger amount of games as obviously the impact may be pretty significant.
Who knows, it might even fix some issues related to VU0 in other games! :)

Please take a look at your earliest convenience.

### DIFF
--- a/pcsx2/VU0.cpp
+++ b/pcsx2/VU0.cpp
@@ -122,7 +122,15 @@ void CFC2() {
 		_vu0WaitMicro();
 	}
 	if (_Rt_ == 0) return;
-	cpuRegs.GPR.r[_Rt_].UL[0] = VU0.VI[_Fs_].UL;
+	
+	if (_Fs_ == REG_TPC) {
+		// For explanation why this is needed here please refer to 
+		// recCFC2() definded in microVU_Macro.inl
+		cpuRegs.GPR.r[_Rt_].UL[0] = VU0.VI[_Fs_].UL >> 3;
+	} else {
+		cpuRegs.GPR.r[_Rt_].UL[0] = VU0.VI[_Fs_].UL;
+	}
+
 	if(VU0.VI[_Fs_].UL & 0x80000000)
 		cpuRegs.GPR.r[_Rt_].UL[1] = 0xffffffff;
 	else
--- a/pcsx2/x86/microVU_Macro.inl
+++ b/pcsx2/x86/microVU_Macro.inl
@@ -276,6 +276,39 @@ static void recCFC2() {
 	}
 	else xMOV(eax, ptr32[&vu0Regs.VI[_Rd_].UL]);
 
+	if (_Rd_ == REG_TPC) { // Divide TPC register value by 8 during copying
+		// Ok, this deserves an explanation.
+		// Accoring to the official PS2 VU0 coding manual there are 3 ways to execute a micro subroutine on VU0
+		// one of which is using the VCALLMSR intruction.
+		// The manual requires putting the address of the micro subroutine
+		// into the CMSAR0 register divided by 8 using the CTC2 command before executing VCALLMSR.
+		// Many games (for instance, 24: The Game, GTA LCS, GTA VCS and FFXII) do in fact use this way, 
+		// they diligently put the address of the micro subroutine into a separate register (v0, v1 etc), divide it by 8
+		// and move it to CMSAR0 by calling the CTC2 command.
+	
+		// However, there are also at least 2 confirmed games (R Racing Evolution, Street Fighter EX3) 
+		// that execute a piece of code to run a micro subroutine on VU0 like this:
+		// 
+		// ...
+		// cfc2	t4, TPC
+		// ctc2	t4, CMSAR0
+		// callmsr
+		// ...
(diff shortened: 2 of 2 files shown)


MarcoEstevez:
I cant believe it just happen today, Thanks

gregory38:
Good job.

Disclaimer I don't know anything on VU. But from an hardware point of view, if the TPC is always aligned on 8B, you don't need register for the 3 lsb. Otherwise a division by 8 is free in the HW. So it makes sense :+1:

tabnk:
Waiting a long time for this fix. Street Fighter EX3 is near perfect :)

FlatOutPS2:
I can confirm it fixes R: Racing, but it does add slowdown. Not caused by the fix, it's just that the cars slow down OpenGL Hardware(required for best results) with AMD drivers. :p

refractionpcsx2:
Nice job! That does all make sense, certainly something simple that got completely overlooked.  I wonder if it fixes the Superman game with the insane SPS as well.  But I'm happy with this :)

FlatOutPS2:
It doesn't fix the missing characters in My Street though.

refractionpcsx2:
No I know what causes that, it is the T bit/macro interlocking.  The same thing with Mike Tysons Boxing.  I did managed to hack it to a point My Street looked mostly OK and the animation worked in Mike Tyson (although textures were a black mess, but it was messy and did cause some other issues.

Here's some screenshots :P

mystreet
miketyson

sudonim1:
Fun fact: the cfc2, ctc2, vcallmsr sequence is precisely what the manual says to do, although it doesn't outright provide an assembly listing for it.

Nobbs66:
Wonder if this will fix Sega superstars tennis.

refractionpcsx2:
> Wonder if this will fix Sega superstars tennis.

It might do

FlatOutPS2:
The Sega Superstarts Tennis issue reminded me of the Amplitude character issue. But the Amplitude issue is unaffected by this PR.

Nobbs66:
I'll test in a few min.

refractionpcsx2:
Damn, didn't fix Superman Returns, oh well nevermind! It fixed 2 games which needed fixing so good job :)

volodymyrkutsenko:
Surprisingly enough I looked at Superman Returns too while searching for the games that use VCALLMSR. But the game uses VCALLMS instead to run the vu0 micro subroutines so sadly it's unrelated to this particular fix.
Maybe some other time I'll get to it too :)

refractionpcsx2:
Indeed, not sure what causes the SPS on that, it is the same across all renderers, which is kinda good i guess??? But it's a bit of a mystery for now, could be a COP2/FPU problem, but I'm just guessing.

If you do work it out, you get bonus internet points for being the first to solve it :P

awichen:
I got a problem...

When Guile use "Sonic Boom Typhoon", there's still something wrong
If i choose Renderer "Hardware", i can't see the tornado.
sonic2


If i choose Renderer 'Software", i can see it...
sonic1


Does anyone know how to fix it??

gregory38:
It isn't the subject of this PR. Otherwise it won't work neither in SW. No clue what happen here but HW renderer is far from perfect.

PR #1370, opened 2016-05-24
Gsdx default option

gregory38:
Remaining work:
* Finish to port GPU GetConfig
* Default Renderer on Windows
* Test Windows

PR #1836, opened 2017-02-23
Greg/gsdx init

gregory38:
I redo a bit the init code to move the window creation first. Main goal is to allow openGL detection before creating the renderer which is not yet done. First I want to be sure that it is still working. And if I can use dynamic_cast on GSdx.

### DIFF
--- a/plugins/GSdx/GS.cpp
+++ b/plugins/GSdx/GS.cpp
@@ -220,7 +220,7 @@ static int _GSopen(void** dsp, const char* title, GSRendererType renderer, int t
 		threads = theApp.GetConfigI("extrathreads");
 	}
 
-	GSWnd* wnd[2] = { NULL, NULL };
+	std::shared_ptr<GSWnd> wnd[2];
 
 	try
 	{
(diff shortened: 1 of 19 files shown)


mirh:
Cool. 
Would this also allow that "proper" detection of GL driver/features you had dismissed [last year](https://github.com/PCSX2/pcsx2/pull/1370#issuecomment-221850542)?

gregory38:
@turtleli could you review my code, it would love your skilled eyes :)

@mirh 
Yeah something like that. So far the idea was more or less a priority list. Put openGL first, if supported keep it, otherwise go for Dx. It really depends of what we want to do, I said `we` but I mean `you`.

mirh:
Well, if we are at it, I guess like you could put some detection code where if an api isn't available you don't show it in the menu?

I guess we could have a initialization like:

````
if (dx10){
     show dx11 & show dx9
     set sw render on dx11
} else{
        if(dx9){
                show dx9
                set sw render on dx9
        }
}

if (GL){
     show GL
     if (GL support > 4.0){
           set sw render on GL
}
````

**Comment by ssakash** [comment]
> Well, if we are at it, I guess like you could put some detection code where if an api isn't available you don't show it in the menu?

Unless I'm wrong, this has already ben done for the Windows GUI.

**Comment by gregory38** [comment]
It is quite complex (and slow) to test GL version. However we can guess GL support from the DX level. On the GUI side (or first level check), you can requires DX11 to show openGL. Technically DX10 hardware is fine but all vendors decide to not-implement basic GL feature. So all DX10 GPU are unsupported. The only DX11 GPU that won't work even with latest driver is Ivy Bridge (market share will only goes down).

**Comment by mirh** [comment]
> It is quite complex (and slow) to test GL version.

Oh, so that's (still) the point. 
So, I guess you could check for whatever you have to check at first "attempt to even use" it?
Like:

````
if (dx10){
     show dx11 & show dx9
     set sw render on dx11
     if (dx11){
               show GL
     }
} else{
        if(dx9){
                show dx9
                set sw render on dx9
        }
}
````
Then whenever you try hw GL, do the if (GL support > 4.0) check to warn about shit ivy. 
When you try sw renderer, do the if (GL support > 4.0) to set sw render on GL, otherwise warn about shit ivy.

**Comment by turtleli** [comment]
I think the code is fine.

> And if I can use dynamic_cast on GSdx.

RTTI is currently disabled on Windows. It can't be used unless it's enabled.

**Comment by gregory38** [comment]
@turtleli 
```
RTTI is currently disabled on Windows. It can't be used unless it's enabled.
```
Ok. Only GSdx ? Is RTTI that costly ? I can easily workaround it with a virtual function.

turtleli:
It's disabled globally. I don't know how costly it is.

PR #1839, opened 2017-02-24
Plugin-Function: Add OSD for incomplete/corrupt savestates

ssakash:
**Summary of changes**:

* Add OSD log messages for cases where a savestate load fails. (_**Refer commit message for more details**_)


gregory38:
Did you test it ? (for example by creating a dummy save state file and try to load it)

FlatOutPS2:
@gregory38 It works just fine.

gregory38:
Thanks both of you :)

Issue #1684, opened 2016-12-02
Galerians Ash (NTSC-U) - Several screens and texts missing.


piau9000:
***PCSX2 version:*** 1.5.0 build 1581.
***PCSX2 options:*** Tried all combinations of Clamp and Round. Also tried Software/Hardware mode and OpenGL/DirectX. Other settings are default.
***Plugins used:*** All default plugins.
***Last known version to work:*** I've not tested but, by reading some posts in the forum, I think it never worked.
***PC specifications:*** Intel Core I5 4690, EVGA Geforce GTX 750Ti, Windows 7 64 Bit.

***Description of the issue:*** Some texts and screens does not appear.
 The first example is right at the intro. The company logos in the intro does not appear. The screen keeps fading-in to white and fading-out to black. But the logos does not appear. You can only see the fade-in and fade-out. Here is how it's supposed to be:
https://www.youtube.com/watch?v=XJbII-_DGvE (right at the beggining)
Here is a gsdump:
http://www15.zippyshare.com/v/r5HlsZbg/file.html

Also, right at the beggining of the game, when the main character checks the body a text at the bottom says "Cain". Then the screens cut out to a door and then to a background with his face and the info about a item. Here's a video explaining it better:
https://www.youtube.com/watch?v=CZRys2oGYsk (go to min 8:47)
In pcsx2 it's really glitchy. The text "Cain" does not appear. When the screen cuts out to his face it is just a black background. The info about the item does not appear too. The only thing that appears is the 3D object floating.
Here's a gs dump of the specific part:
http://www15.zippyshare.com/v/gjRYRCaN/file.html
This bug appears during the entire game. Also some parts (puzzles) are unbeatable because of it.

piau9000:
Okay. Effing zippyshare nuked all my files. :(
Sorry gregory. Here, I have reuploaded them just in case you need:

Intro:
http://www.mediafire.com/file/1mqggcocm3717ky/Galerians_2_-_Intro_Logos.zip
Ingame:
http://www.mediafire.com/file/82co4er6u8ha6ox/Galerian_2_-_Text_And_2D_Sceens.zip

gregory38:
@piau9000 instead could you test the pr #1843 and tell me if it is working as expected.

piau9000:
It seems that #1843 is not merged yet. I don't know how to build it myself.
But I tested rev 1910 and 1911 (the ones that make change to the grandient division) and the bug still persists.

[...]

Issue #1769, opened 2017-01-13
[GSdx SW] Xenosaga Episode I [NTSC-U]: glitchy event slot visuals


trostboot:
***PCSX2 version:***
1.5.0-dev-1789, 1.4.0

***PCSX2 options:***
defaults, no speedhacks

***Plugins used:***
GSdx SW, SW settings have no effect on it

***Description of the issue:***
Some pixels in the event slot graphic (bottom right of the screen) during battles are shifted to the right in SW rendering modes. It looks correct in HW modes, regardless of API.

SW mode (glitched):
![gsdx_20170113105621](https://cloud.githubusercontent.com/assets/17594477/21926101/a48209ec-d980-11e6-8011-7034531b6e0c.png)

HW mode (correct):
![gsdx_20170113105619](https://cloud.githubusercontent.com/assets/17594477/21926115/b45aec1c-d980-11e6-8431-5422817acfca.png)

GSdump: 
[gsdx_20170113105443.zip](https://github.com/PCSX2/pcsx2/files/704011/gsdx_20170113105443.zip)

***How to reproduce the issue:***
Enter any battle and look at the event slot. It's most noticeable in the empty slot.

***Last known version to work:***
Unknown, it's present in current git builds as well as 1.4.0. I can look at older builds later today if need be.

***PC specifications:***
i5-4690k, R9 270X, 16GB RAM, Win10 Pro x64 14393

gregory38:
I'm trying to debug the dump. The bad draw call is the number 936

The game renders the area the inner rectangle in 3 parts (top/middle/bottom). Each part is composed of 2 triangles. Delta(position  X) == Delta(texture X) == 288.

However when I dump the `dscan` variable (delta). Note, position is scaled with a factor `1.0f` and texture is scaled with a factor `65536.0f`
```
Prim dscan p:1.000000 t:65536.007812
Prim dscan p:1.000000 t:65536.007812

Prim dscan p:1.000000 t:65535.996094
Prim dscan p:1.000000 t:65535.996094

Prim dscan p:1.000000 t:65536.007812
Prim dscan p:1.000000 t:65536.007812
```

As you can see the top/bottom part got t slighly above 1. But the middle is smaller than 1 which could "enlarge" the texture. I don't have any proof that this small rounding can actually provoke a wrong sampling. 

[...]

PR #1843, opened 2017-02-27
gsdx sw: rcp introduce bad rounding on the grandiant

gregory38:
Division is slower but more accurate. Fix rendering issue on Xenosaga (batte slot)/Jak3 (skin color)

Geez need to edit the commit to link the issue.
Issue #1769 

TODO
* [x] add comment
* [x] use accurate div on sw ref implementation
* [ ] add option for sw fragment shader
* [x] add template parameter for VertexTrace
* [x] automatically enable previous options when Q is big enough

### DIFF
--- a/plugins/GSdx/GSRendererSW.cpp
+++ b/plugins/GSdx/GSRendererSW.cpp
@@ -60,11 +60,15 @@ GSRendererSW::GSRendererSW(int threads)
 		m_tex_pages[i] = 0;
 	}
 
+	#define InitCVB2(P, Q) \
+		m_cvb[P][0][0][Q] = &GSRendererSW::ConvertVertexBuffer<P, 0, 0, Q>; \
+		m_cvb[P][0][1][Q] = &GSRendererSW::ConvertVertexBuffer<P, 0, 1, Q>; \
+		m_cvb[P][1][0][Q] = &GSRendererSW::ConvertVertexBuffer<P, 1, 0, Q>; \
+		m_cvb[P][1][1][Q] = &GSRendererSW::ConvertVertexBuffer<P, 1, 1, Q>;
+
 	#define InitCVB(P) \
-		m_cvb[P][0][0] = &GSRendererSW::ConvertVertexBuffer<P, 0, 0>; \
-		m_cvb[P][0][1] = &GSRendererSW::ConvertVertexBuffer<P, 0, 1>; \
-		m_cvb[P][1][0] = &GSRendererSW::ConvertVertexBuffer<P, 1, 0>; \
-		m_cvb[P][1][1] = &GSRendererSW::ConvertVertexBuffer<P, 1, 1>; \
+		InitCVB2(P, 0) \
+		InitCVB2(P, 1)
 
 	InitCVB(GS_POINT_CLASS);
 	InitCVB(GS_LINE_CLASS);
@@ -207,9 +211,11 @@ GSTexture* GSRendererSW::GetFeedbackOutput()
 }
 
 
-template<uint32 primclass, uint32 tme, uint32 fst>
+template<uint32 primclass, uint32 tme, uint32 fst, uint32 q_div>
 void GSRendererSW::ConvertVertexBuffer(GSVertexSW* RESTRICT dst, const GSVertex* RESTRICT src, size_t count)
 {
+	// FIXME q_div wasn't added to AVX2 code path.
+
 	#if 0//_M_SSE >= 0x501
 
 	// TODO: something isn't right here, this makes other functions slower (split load/store? old sse code in 3rd party lib?)
@@ -313,12 +319,27 @@ void GSRendererSW::ConvertVertexBuffer(GSVertexSW* RESTRICT dst, const GSVertex*
 				#if _M_SSE >= 0x401
 
 				t = GSVector4(xyzuvf.uph16() << (16 - 4));
-					
+
 				#else
 
 				t = GSVector4(GSVector4i::load(src->UV).upl16() << (16 - 4));
 
 				#endif
+			}
+			else if(q_div)
+			{
+				// Division is required if number are huge (Pro Soccer Club)
+				if(primclass == GS_SPRITE_CLASS && (i & 1) == 0)
+				{
+					// q(n) isn't valid, you need to take q(n+1)
+					const GSVertex* next = src + 1;
+					GSVector4 stcq1 = GSVector4::load<true>(&next->m[0]); // s t rgba q
+					t = (stcq / stcq1.wwww()) * tsize;
+				}
+				else
+				{
+					t = (stcq / stcq.wwww()) * tsize;
+				}
 			}
 			else
 			{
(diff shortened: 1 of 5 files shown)


plugins/GSdx/GSRasterizer.cpp
```diff
@@ -464,7 +462,7 @@ void GSRasterizer::DrawTriangle(const GSVertexSW* vertex, const uint32* index)
 	ddx[1] = ddx[0].yxzw();
 	ddx[2] = ddx[0].xzyw();
 
-	GSVector8 _dxy01c(dxy01 * cross);
+	GSVector8 _dxy01c(dxy01 / cross);
```
gregory38: Need to add comment of rcp vs div

gregory38:
@sudonim1 

The issue #551 is potentially related to inaccurate rcp operation on the VertexTrace (and likely in the fragment shader too)
```
But Gsdx isn't good neither.

    S/T = 3.40282347e+38 (biggest float)
    Q = 3.40282347e+38

GSdx requires to compute S/Q (and T/Q)

    1/Q gives 0 due to float error
    S * (1/Q) gives you 0 instead of 1
```

I looked this morning at Agner manual. And `div` is really slow on AMD Jaguar CPU `~40 cycles`. I need to check recent AMD CPU. Intel's Sandy is 28. Recent Intel (at least Haswell and maybe IB) is 14. So rcp isn't a so good idea nowadays. It would be nice to find a good benchmark (high number of polygons + texture + ST coordinate)

sudonim1:
If it comes down to it, it can be an option.  The benchmark would just be any high poly textured 3D scene.

gregory38:
I think for the JIT I will try to detect the CPU type. (Maybe based on AVX 1 or 2 support). Per fragment operation is much more costly than per primitive operation.

Initially I wanted to look SotC but it uses cell shading not texture.

gregory38:
Question remain how to handle this case ? 3 divisions or do a 1.0f/q that hopefully won't be optimized to rpc. I'm afraid that 3 div will stall the pipeline. However it might handle better huge/small value.
```
GSVector4 q = stq0.wwww(stq1).xzww(stq2).rcpnr();

stq0 = (stq0.xyww() * q.xxxx()).xyww(stq0);
stq1 = (stq1.xyww() * q.yyyy()).xyww(stq1);
stq2 = (stq2.xyww() * q.zzzz()).xyww(stq2);
```

mirh:
Just for the records, Jaguar might even be last AMD architecture (til tomorrow at least :s), but it was meant to be a low budget, low power one. I'm not saying it's an atom, but compromises might have been made.

gregory38:
Current commit is fine. But if we replace more code. We gonna replace an reciprocal+3 mult by 3 div. Which is slower on all arch.Now I don't know the real impact of game fps.

woj1993:
@gregory38 If you got ready code, then why don't test speed impact?
PS. Gregory I don't know is it the same error I suspect this to be Z fighting but Megaman X Command Mission have issues in SW and HW modes. In SW mode it looks like small miscalculation. Can you check is it related to this?

gregory38:
First I'm busy with real life. Then I'm still looking at all the implication. For example the fragment shader uses low accuracy operation.
```
rcpps(xmm0, xmm4); // 1/Q
mulps(xmm2, xmm0); // S * (1 / Q)
mulps(xmm3, xmm0);  // T * (1 / Q)
```
2 division will be 30-50 times slower to execute than the above code. So I think I will need to add an option.

Third we need to find a good benchmark. If the game doesn't hit the code path => no speed impact. If the game mostly hit the code path => speed impact.

```
PS. Gregory I don't know is it the same error I suspect this to be Z fighting but Megaman X Command Mission have issues in SW and HW modes. In SW mode it looks like small miscalculation. Can you check is it related to this?
```
It is much more complicated. The data are coming from the VU. The data are already rounded by VU, GSdx will round them further. Let's first fixes the issue related to `rcp` vs `div`.

gregory38:
@sudonim1 @turtleli Could you review the code.

Not implemented
* SW renderer: option to select div operation. So far my testcase got a constant `Q` so the div is done on the vertex level.
* HW renderer: Optimization of constant Q on the CPU to avoid float overflow on GPU. Bigger code updates, so far it is required by 1 game which isn't fully fixed on the SW renderer. So I propose to postpone it.

plugins/GSdx/GSRendererSW.cpp
```diff
@@ -1153,46 +1167,9 @@ bool GSRendererSW::GetScanlineGlobalData(SharedData* data)
...
-							GSVector4 w = t1.zzzz().rcpnr();
-
-							v[i + 0].t = (t0 * w).xyzw(t0);
-							v[i + 1].t = (t1 * w).xyzw(t1);
-						}
-					}
-				}
+				// skip per pixel division if q is constant
+				// Note: the Q division was done GSRendererSW::ConvertVertexBuffer
+				gd.sel.fst |= (m_vt.m_eq.q || primclass == GS_SPRITE_CLASS);
```
sudonim1: Doesn't check for 1.0 Q unlike GSRendererSW::Draw? It should be a no-op so I guess it won't affect emulation other than speed.

gregory38: Because in `GSRendererSW::Draw` the division by 1 could replaced by a nop. So you don't actually need to do it.
So whatever the Q value, if Q is constant you could avoid the division in the Fragment Shader.

gregory38:
I need to benchmark the speed impact of the pre Q division. Code is likely slower but I don't think we will do the pre division often.

I will push the first commit in master because it fixes at least 2 games. And there is no way to have a faster solution.

plugins/GSdx/GSRendererSW.cpp
sudonim1: I was just having trouble following the slightly different logic for pre-division and flat ST for texture mapping.  The code's correct.  Flat ST handling if not mipmapping and (Q is constant or drawing sprites), pre-division with the same conditions but skips q=1 for efficiency, got it.

gregory38:
@turtleli 
https://en.wikipedia.org/wiki/Texture_mapping#Perspective_correctness

My understanding is that `S == u/z`, `T == v/z`, `Q == 1/z`. In 2d rendering, `u` and `v` are interpolated from the vertex position. For correct 3d rendering, you need to take into account the depth value. So the GS will compute `interpolation of S / interpolation of Q`. If Q is uniform on the primitive, the `interpolation of Q == Q`. And `alpha * interpolation of x == interpolation of alpha * x`.

Remain a strange case, the sprite case. Sprites are 2d rectangle without depth. So it doesn't make sense to do a perpective correction. However sudonim1 is likely right when he said that Q can be set to any values.

My latest commit only moves code around. ST are normalized coordinate (without wrapping range from 0 to 1), so you need to multiply the coordinate by the size of the texture. And then do the division by Q. However if ST are huge. `ST * size` is infinite and `ST * size / Q` is bad. My commit tries to do the Q division first. So `(ST / Q) * size` gives a standard number. My commit also use `/ x` instead of `* 1/x` to avoid accuracy issue but the instruction is slower.

prafullpcsx2:
This seems to fix https://github.com/PCSX2/pcsx2/issues/1684

gregory38:
@prafullpcsx2 thanks for the info. I pushed the first commit of this PR to master, could you tell me if it is enough. If not, could you share with us some gsdump, the one from the issue report are expired. Long story short, I found some bug in my PR on the sprite handling (but less buggy than master by pure luck). I want to be sure that I understand the expected behavior correctly.

prafullpcsx2:
@gregory38 First commit of this PR doesn't fix it. I will upload the gsdump later. I do have a block dump ready though incase you want to use it. http://www.filedropper.com/ga_1

gregory38:
@prafullpcsx2 ok. I think I have a quick fix that could explain some bugs. I will push it soon.

gregory38:
@prafullpcsx2 yes, you're right. The game uses huge float. 
```
 Vertex Trace: float overflow detected ! min 3.402823e+38 max 3.402823e+38
```
Now, I know that at least 2 games are affected. Unfortunately I need to redo/fix my code to handle the Q properly for sprite.

gregory38:
Please don't test the branch now, I break ATV again.

PR #1844, opened 2017-03-03
gsdx: split texture filtering option

### LINKED ISSUE #1837: GSdx: linear filtering option
Currently we have a combo box to force the texture filtering mode
* always nearest
* PS2
* always bilinear
* trilinear
* trilinear (+ bilinear)
* trilinear forced

Some people want similar (near/bilinear) option for all renderers (namely the SW renderer).

Instead to add more code in the sw renderer. I want to change the code that does linear detection (in GSVertexTrace). And split the option in 2.

We will have a common option to select nearest/bilinear mode.
* always nearest
* PS2
* always bilinear but SPRITE (upscaling issue).
* always bilinear

Then we will have a dedicated option for trilinear for openGL.
* off (recommanded)
* trilinear
* trilinear always

I think it would make the configuration easier. And code easier too. But I want some feedbacks.

gregory38:
Bilinear applies to all renderer
* Common code done in GSVertexTrace
* Extend it with forced but sprite (trade-off between linear/upscale glithes)
* Linux GUI option was moved at the top with the renderer selection

Trilinear is moved to OGL hack

TODO: update Windows GUI

close #1837 

Please review.

### DIFF
--- a/plugins/GSdx/GSSetting.cpp
+++ b/plugins/GSdx/GSSetting.cpp
@@ -33,11 +33,17 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_FILTER:
 			return "Control the texture filtering of the emulation.\n\n"
 				"Nearest:\nAlways disable interpolation, rendering will be blocky.\n\n"
+				"Bilinear Forced (excluding sprite):\nAlways enable interpolation except for sprites (FMV/Text/2D elements)."
+				" Rendering is smoother but it could generate a few glitches. If upscaling is enabled, this setting is recommended over 'Bilinear Forced'\n\n"
 				"Bilinear Forced:\nAlways enable interpolation. Rendering is smoother but it could generate some glitches.\n\n"
-				"Bilinear PS2:\nUse same mode as the PS2. It is the more accurate option.\n\n"
+				"Bilinear PS2:\nUse same mode as the PS2. It is the more accurate option.\n\n";
+#ifdef __unix__
+		case IDC_TRI_FILTER:
+			return "Control the texture tri-filtering of the emulation.\n\n"
+				"None:\nNo extra trilinear filtering.\n\n"
 				"Trilinear:\nUse OpenGL trilinear interpolation when PS2 uses mipmaps.\n\n"
-				"Trilinear Forced Bilinear:\nSame as above but always enable bilinear interpolation.\n\n"
-				"Trilinear Ultra:\nAlways enable full trilinear interpolation. Warning Slow!\n\n";
+				"Trilinear Forced:\nAlways enable full trilinear interpolation. Warning Slow!\n\n";
+#endif
 		case IDC_CRC_LEVEL:
 			return "Control the number of Auto-CRC hacks applied to games.\n\n"
 				"None:\nRemove nearly all CRC hacks (debug only).\n\n"
--- a/plugins/GSdx/GS.h
+++ b/plugins/GSdx/GS.h
@@ -1416,14 +1416,19 @@ enum class GSVideoMode : uint8
 	DTV_1080I
 };
 
-enum class Filtering : uint8
+enum class BiFiltering : uint8
 {
 	Nearest,
-	Bilinear_Forced,
-	Bilinear_PS2,
-	Trilinear,
-	Trilinear_Bilinear_Forced,
-	Trilinear_Always
+	PS2,
+	Forced_But_Sprite,
+	Forced,
+};
+
+enum class TriFiltering : uint8
+{
+	None,
+	PS2,
+	Forced,
 };
 
 enum class CRCHackLevel : uint8
--- a/plugins/GSdx/GSDeviceOGL.cpp
+++ b/plugins/GSdx/GSDeviceOGL.cpp
@@ -76,7 +76,9 @@ GSDeviceOGL::GSDeviceOGL()
 	GLState::Clear();
 
 	m_mipmap = theApp.GetConfigI("mipmap");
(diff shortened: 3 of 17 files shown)


MrCK1:
A note/tooltip should be added telling users that Trilinear can cause artifacts.

FlatOutPS2:
I'll add the Windows UI to my upcoming GSdx PR.

plugins/GSdx/GSSetting.cpp
```diff
@@ -33,11 +33,16 @@ const char* dialog_message(int ID, bool* updateText) {
...
 				"Bilinear Forced:\nAlways enable interpolation. Rendering is smoother but it could generate some glitches.\n\n"
-				"Bilinear PS2:\nUse same mode as the PS2. It is the more accurate option.\n\n"
+				"Bilinear PS2:\nUse same mode as the PS2. It is the more accurate option.\n\n";
+#ifdef __unix__
+		case IDC_TRI_FILTER:
+			return "Control the texture tri-filtering of the emulation.\n\n"
+				"None:\nNo extra trilinear filtering.\n\n"
 				"Trilinear:\nUse OpenGL trilinear interpolation when PS2 uses mipmaps.\n\n"
-				"Trilinear Forced Bilinear:\nSame as above but always enable bilinear interpolation.\n\n"
 				"Trilinear Ultra:\nAlways enable full trilinear interpolation. Warning Slow!\n\n";
```
FlatOutPS2: It's called `Trilinear Forced` in code, but it's still `Trilinear Ultra` here. Either one needs to change. :)

plugins/GSdx/GSSetting.cpp
```diff
@@ -33,11 +33,16 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_FILTER:
 			return "Control the texture filtering of the emulation.\n\n"
 				"Nearest:\nAlways disable interpolation, rendering will be blocky.\n\n"
+				"Bilinear Forced/Upscale:\nAlways enable interpolation except for sprite. Rendering is smoother but it could generate few glitches.\n\n"
```
FlatOutPS2: `Forced/Upscale` sounds confusing to me. I'd prefer something like `Forced (excluding sprite)` , which is closer to what it's called in the code.

`Rendering is smoother but it could generate few glitches.`
There should be an`a` between generate and few:
`Rendering is smoother but it could generate a few glitches.`

FlatOutPS2:
This PR causes graphical regressions in both Hardware and Software modes in several games.

gregory38:
Which configuration of texturing ?

FlatOutPS2:
> Which configuration of texturing ?

Nearest and Bilinear (PS2).

Both Forced options seem fine in HW and SW.

FlatOutPS2:
Ace Combat Squadron Leader has a regression in Direct3D11 HW only. SW mode and Direct3D9 HW and OpenGL HW are fine. Any reason why that would happen?

gregory38:
Not tested but reading the code, I forgot to remove a line. It could explain some bad behaviors.

gregory38:
Hum, it might not explain the DX discrepancy. Could be another typo.

FlatOutPS2:
> Hum, it might not explain the DX discrepancy. Could be another typo.

Removing the line seems to remove all the issues. I'm still confused why one of the issues was D3D11 HW only though. Especially as I ran into another D3D11 only issue with the vertex shader a few days ago. It seems D3D11 HW does something different to D3D9 and OGL somewhere.

FlatOutPS2:
@gregory38 Do you know any good games for testing the new "forced but sprite" option?

gregory38:
2D JRPG games + upscale (as the goal is to avoid texture sampling leak when upscaling is enabled)
* mana khemia
* ar tolenico

FlatOutPS2:
> as the goal is to avoid texture sampling leak when upscaling is enabled

You should add something like that to the tooltip, maybe with an example game it fixes.

FlatOutPS2:
@gregory38 @MrCK1 The new Bilinear Forced/Upscale option seems to avoid the "Bilinear (Forced)" texture filtering from messing up the car textures. Although that's not an upscaling issue.

MrCK1:
Thanks for the heads up. As for the Trilinear (Ultra) option, what is the status on that?/ - do we still need to worry about corrupt textures?

MrCK1:
@FlatOutPS2 I just checked Nascar and you are correct about the textures. A captured frame also shows that the corrupt texture still remain in the draw call should an issue pop up in the future.

gregory38:
@FlatOutPS2 I push a new commit based on your feedback. Tell me if you're ok with it.

@MrCK1 Trilinear is a hack so yes you can expect texture corruption (that why I prefer to put it in the hack section).

FlatOutPS2:
@gregory38 As I mentioned in a later comment it would be nice if the Forced (excluding sprite) tooltip also has a note about the upscaling.

gregory38:
@FlatOutPS2 does this one sound better ?
```
+               "Bilinear Forced (excluding sprite):\nAlways enable interpolation except for sprites (FMV/Text/2D elements)."
+               " Rendering is smoother but it could generate a few glitches. If upscaling is enabled, this setting is recommanded over 'Bilinear Forced'\n\n"
```

FlatOutPS2:
`recommanded` should be `recommended`, but other than that it looks fine. 👍

gregory38:
It is late don't ask me too much :p

gregory38:
@FlatOutPS2 I let you handle the Windows GUI

PR #1853, opened 2017-03-11
GSdx: Windows and Direct3D updates

FlatOutPS2:
- Adds GSdx memory wrapping option to the Windows hacks dialog

- Adds Direct3D port of new Half-pixel Offset Special options to the Windows hacks dialog

- Updates all GS windows to more closely resemble Linux GUI
- Adds OpenGL Very Advanced Custom Settings to the Windows hacks dialog

- Adds Texture Filtering Of Display option to the Windows Shader dialog

- Adds split Texture Filtering Windows GUI options (see commit for further details)
Close: https://github.com/PCSX2/pcsx2/issues/1837

- Adds Unscale Point&Line Direct3D11 port to the Windows hacks dialog


gregory38:
Oh my! Big PR for windows users :) Next time try to do smaller commits.

FlatOutPS2:
> Oh my! Big PR for windows users :) Next time try to do smaller commits.

I would, but I had so much time waiting for other PR's to merge. :P And I needed several of the functions together for some games too, so it made more sense to keep 'em together. I did leave one commit out of the PR though, so there's that.

lightningterror:
`Oh my! Big PR for windows users :) Next time try to do smaller commits.`
It's cool like this as well.

lightningterror:
> Oh my! Big PR for windows users :) Next time try to do smaller commits.

It's cool like this as well.

gregory38:
@lightningterror easy for you, you're not the one to review the code ;)

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,48 @@ void vmfree(void* ptr, size_t size)
...
+	if (fh != 0)
+		_unlink(file_name);
+	else
+		fprintf(stderr, "Failed to reserve memory due to %s\n", strerror(errno));
+
+	void* fifo = MapViewOfFile(fh, FILE_MAP_ALL_ACCESS, 0, 0, size);
+	
+	for (size_t i = 1; i < repeat; i++) {
+		void* base = (uint8*)fifo + size * i;
+		Next[i] = (uint8*)MapViewOfFileEx(fh, FILE_MAP_ALL_ACCESS, 0, 0, size, base);
```
gregory38: Return an error when `Next` will overflow (i.e. `repeat >= countof(Next)`)

plugins/GSdx/GSSetting.cpp
```diff
@@ -143,6 +143,8 @@ const char* dialog_message(int ID, bool* updateText) {
 			return "Increases the width of lines at higher than native resolutions. This ensures that the lines will keep the correct proportions and prevents aliasing. "
 				"Avoids empty lines on the screen in games such as Ridge Racer V, and clears FMVs obscured by a grid in games like the Silent Hill series and Dirge of Cerberus.";
 #ifdef _WIN32
+		case IDC_MEMORY_WRAPPING:
```
gregory38: You could add the enum in the `GSSetting.h` file and avoid the ifdef. (yes less work for me)

plugins/GSdx/res/tfx.fx
```diff
@@ -92,6 +92,7 @@ cbuffer cb0
 	float4 VertexScale;
 	float4 VertexOffset;
 	float2 TextureScale;
+	float4 TextureOffset;
```
gregory38: Won't be possible to pack `TextureOffset` and `TextureScale` together. The GPU unit is `float4`. Beside it will allow to avoid a GSVector4 check when you udpate the buffer cache.

plugins/GSdx/GSDeviceDX.h
```diff
@@ -162,7 +154,7 @@ class GSDeviceDX : public GSDevice
 				uint32 fst:1;
 				uint32 wms:2;
 				uint32 wmt:2;
-				uint32 fmt:3;
+				uint32 fmt:4;
```
gregory38: You might need to update the `uint32()` operator. For openGL I added the missing bits as free and static assert on the structure size. The init will set 0 to all bits. And if I forget to reduce the free bits, I will get an assert.

plugins/GSdx/GSTextureFX11.cpp
```diff
@@ -139,11 +153,12 @@ void GSDevice11::SetupVS(VSSelector sel, const VSConstantBuffer* cb)
 	IASetInputLayout(i->second.il);
 }
 
-void GSDevice11::SetupGS(GSSelector sel)
+void GSDevice11::SetupGS(GSSelector sel, const GSConstantBuffer* cb)
 {
 	CComPtr<ID3D11GeometryShader> gs;
 
-	if(sel.prim > 0 && (sel.iip == 0 || sel.prim == 3)) // geometry shader works in every case, but not needed
+	bool Unscale_GSShader = (sel.point == 1 || sel.line == 1) && theApp.GetConfigB("UserHacks") && theApp.GetConfigB("UserHacks_unscale_point_line");
```
gregory38: Perf wise, GetConfig* aren't a good idea.

mirh:
God, it feels like christmas. 
Tested half-pixel and unscale. Looks lovely. 

What's dx still missing then? Merge_pp_sprite and depth?

gregory38:
Merge pp Sprite must be finish first. It is maybe not a good hack anyways. Otherwise you will miss accurate blending, channel effects and depths effects. But yes it is a huge improvement for users with a limited driver.

FlatOutPS2:
DX also has an issue with Full mipmapping(in Ape Escape 2 at least) and doesn't have Trilinear Filtering.

mirh:
> Merge pp Sprite must be finish first. It is maybe not a good hack anyways. 

Indeed, even after having tinkered the [hell](https://www.reddit.com/r/emulation/comments/56wj30/mipmapping_makes_the_ace_combat_games_run_way/d8osrry/) out of every fix.. In the end I still have some UI element slight amiss in AC5. 

> Perf wise, GetConfig* aren't a good idea.

And don't know if that's an issue just at loading time or even when running, but unscale seems indeed to have a 5% performance hit.

plugins/GSdx/res/tfx.fx
FlatOutPS2: >Won't be possible

If you say so. :P It would require `vs_cb.TextureScale = GSVector4(1.0f / 16) / WH.xyxy();` to be replaced by something like:
```
		vs_cb.Texture_ScOff.x = (1.0f / 16) / WH.x;
		vs_cb.Texture_ScOff.y = (1.0f / 16) / WH.y;
```

tabnk:
Nice. Good for AMD GPU. Just missing accurate blending :)

lightningterror:
> And don't know if that's an issue just at loading time or even when running, but unscale seems indeed to have a 5% performance hit.

Faster or slower ? 

> @lightningterror easy for you, you're not the one to review the code ;)

Ups I forgot that you were assigned to review the code xD

gregory38:
@FlatOutPS2 
Due to latest change, I updated the behavior value of option.
```
enum class BiFiltering : uint8
{
	Nearest,
	PS2,
	Forced_But_Sprite,
	Forced,
};
```
1 used to be Forced, 2 used to be PS2. Can I put back PS2 on 2? Or will it impact your code/gui ?
enum class BiFiltering : uint8
{
	Nearest,
	PS2 = 2,
	Forced_But_Sprite = 1,
	Forced = 3,
};
```

File: plugins/GSdx/res/tfx.fx (hunk 3 above)
Comment (gregory38) [inline plugins/GSdx/res/tfx.fx]: Don't ask me how. But it is possible to shuffle thing around to only do a single division.

**Comment by FlatOutPS2** [comment]
@gregory38 It shouldn't be an issue. Even if somehow there is, it should be an easy fix.

My earlier responce to your code comment: A look at a Wallace and Gromit dump seems to indicate it does need the hack for SW too. How should it be included in the GUI? Do we want to add hacks for the Software renderer to the GUI?

**Comment by gregory38** [comment]
Ah yes, I searched some review comments that I need to reply but I didn't found it. Github is quite buggy sometimes :(

I think it depends on the speed impact. I think it is free for the SW renderer. So it could be potentially be enabled by default. However code need to be stable.

File: plugins/GSdx/stdafx.cpp [hunk 6]
```diff
@@ -71,16 +71,48 @@ void vmfree(void* ptr, size_t size)
...
+uint8* Next[4];
+
 void* fifo_alloc(size_t size, size_t repeat)
 {
-	// FIXME check linux code
-	return vmalloc(size * repeat, false);
+	ASSERT(fh == 0);
+
+	const char *file_name = "/GSDX.mem";
+	fh = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE, 0, repeat * size, file_name);
```
turtleli: Is `file_name` necessary? Couldn't nullptr just be passed in so the mapping is created without a name?

gregory38:
@FlatOutPS2 note SW memory wrapping can be postponed to another PR.

plugins/GSdx/res/tfx.fx
```diff
@@ -652,27 +637,112 @@ VS_OUTPUT vs_main(VS_INPUT input)
...
+	// Transform a point to a NxN sprite
+	VS_OUTPUT Point = input[0];
+
+	// Get new position
+	float4 lt_p = input[0].p;
+	float4 rb_p = input[0].p + float4(PointSize.x, PointSize.y, 0.0f, 0.0f);
+	float4 lb_p = rb_p;
+	float4 rt_p = rb_p;
+    lb_p.x    = lt_p.x;
+    rt_p.y    = lt_p.y;
```
turtleli: Whitespace issues.

plugins/GSdx/res/tfx.fx
```diff
@@ -652,27 +637,112 @@ VS_OUTPUT vs_main(VS_INPUT input)
...
+[maxvertexcount(6)]
+void gs_main(line VS_OUTPUT input[2], inout TriangleStream<VS_OUTPUT> stream)
+{
+	// Transform a line to a thick line-sprite
+	VS_OUTPUT left = input[0];
+	VS_OUTPUT right = input[1];
+	float2 lt_p = input[0].p.xy;
+	float2 rt_p = input[1].p.xy;
+
+    // Potentially there is faster math
```
turtleli: Whitespace issues.

plugins/GSdx/GSRendererDX.cpp
```diff
@@ -145,6 +224,194 @@ void GSRendererDX::EmulateZbuffer()
...
+
+
+	if (m_context->TEX0.TFX == TFX_MODULATE && m_vt.m_eq.rgba == 0xFFFF && m_vt.m_min.c.eq(GSVector4i(128))) {
+		// Micro optimization that reduces GPU load (removes 5 instructions on the FS program)
+		m_ps_sel.tfx = TFX_DECAL;
+	}
+	else {
+		m_ps_sel.tfx = m_context->TEX0.TFX;
+	}
+
```
turtleli: Trailing whitespace.

plugins/GSdx/res/tfx.fx
```diff
@@ -652,27 +637,112 @@ VS_OUTPUT vs_main(VS_INPUT input)
...
+	stream.Append(right);
+	stream.RestartStrip();
+
+	// Triangle 2
+	left.p.xy = lb_p;
+	stream.Append(left);
+
+	right.p.xy = rt_p;
+	stream.Append(right);
+
```
turtleli: Trailing whitespace.

plugins/GSdx/GSRendererDX11.cpp
```diff
@@ -60,25 +110,6 @@ void GSRendererDX11::SetupIA()
...
-		t = D3D11_PRIMITIVE_TOPOLOGY_LINELIST;
-		break;
-	case GS_TRIANGLE_CLASS:
-		t = D3D11_PRIMITIVE_TOPOLOGY_TRIANGLELIST;
-		break;
-	default:
-		__assume(0);
-	}
-	
+	dev->IASetIndexBuffer(m_index.buff, m_index.tail);
```
turtleli: Trailing whitespace.

lightningterror:
I'm loving the new Hacks GUI. Now I have a few suggestions.

Advanced Settings & Hacks
 - Add toltips to Geometry Shader , Image Load Store , Trilinear Filtering 

OSD
 - Change usage display. I'd suggest putting FPS first , EE second , VU third and GS third. You can also add percentages to them.
//Bug: When you disable MTVU while the monitor is running VU is still shown with it's last value.\\
 - Change the default font size to 25. I think this is a perfect balance between size and visibility.
 - I would also recommend changing the default font to something else and adding a black outline to the text. Making it look similar to MSI Afterburner would be nice but that's just my opinion.

MrCK1:
Having all that junk in the OSD would be polling the info twice, and that's too much. Alot of games user the corners for their HUDs so must users probably wouldn't want them to be covered up.

FlatOutPS2:
>    Add toltips to Geometry Shader , Image Load Store

@lightningterror Gregory hasn't made tooltips for these, so I can't add 'em.

>    Change the default font size to 25. I think this is a perfect balance between size and visibility.

I fear 25 would be considered too small by some, especially in windowed mode.

The other changes are out of the scope of this PR. I already knew about the VU bug(a workaround is pressing the F9 key twice), and the order of the OSD monitor is currently automatically done in alphabetical order from bottom to top. It could be less than ideal to detect and modify the order of each element.

gregory38:
```
Gregory hasn't made tooltips for these, so I can't add 'em.
```
You can add tooltips without waiting for me ;) However I think I added one for trilinear.

FlatOutPS2:
> You can add tooltips without waiting for me ;) However I think I added one for trilinear.

Tell me what to put in the tooltip and I'll add'em. The Memory wrapping tooltip needs some good text too.

I had already fixed the trilinear tooltip locally, the linux only restrictions in GSSettings.cpp haven't been removed in the current state of the PR. ;)

plugins/GSdx/GSRendererDX11.cpp
```diff
@@ -37,10 +43,54 @@ bool GSRendererDX11::CreateDevice(GSDevice* dev)
...
 }
 
-void GSRendererDX11::SetupIA()
+void GSRendererDX11::SetupIA(const float& sx, const float& sy)
 {
 	GSDevice11* dev = (GSDevice11*)m_dev;
 
+	D3D11_PRIMITIVE_TOPOLOGY t;
+
+	bool unscale_hack = UserHacks_unscale_pt_ln  & (GetUpscaleMultiplier() > 1);// &GLLoader::found_geometry_shader;
```
ssakash: > (GetUpscaleMultiplier() > 1)

Won't work for custom resolutions in that case, intended?

Also FWIW, it would be nice to limit the usage of bitwise AND ``(&)`` only when comparing bitfields or for bitwise manipulations, logical AND ``(&&)`` makes more sense here.

FlatOutPS2: I just copied that line from Gregory's OGL code. As there's no custom resolution on Linux anymore, he probably didn't even consider it.

I thought I'd removed the comment and double space though. :S

ssakash: > I just copied that line from @gregory38 's OGL code. As there's no custom resolution on Linux anymore, he probably didn't even consider it.

You're planning on adding this hack for custom resolutions?

gregory38: Ultimately it would be nice to have a common function for all renderers. But we need to handle `Q division / line to sprite / wild arm hack` in a single function. Besides maybe it would be nicer to directly copy the result in the gl/dx buffer (map/unmap). Rather than to update a local buffer and then copy it in gl/dx buffer.

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,53 @@ void vmfree(void* ptr, size_t size)
...
-	// FIXME check linux code
-	return vmfree(ptr, size * repeat);
+	ASSERT(s_fh != NULL);
+
+	if (s_fh == NULL)
+		return;
+
+	UnmapViewOfFile(ptr);
+
+	for (size_t i = 1; i < 8; i++) {
```
gregory38: I don't like hardcoded 8. There is a `countof` macro

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,53 @@ void vmfree(void* ptr, size_t size)
...
+	void* fifo = MapViewOfFile(s_fh, FILE_MAP_ALL_ACCESS, 0, 0, size);
+
+	for (size_t i = 1; i < repeat; i++) {
+		void* base = (uint8*)fifo + size * i;
+		s_Next[i] = (uint8*)MapViewOfFileEx(s_fh, FILE_MAP_ALL_ACCESS, 0, 0, size, base);
+		errorID = ::GetLastError();
+		if (s_Next[i] != base) {
+			fprintf(stderr, "Failed to memory map contiguous segment, WIN API ERROR:%u\n", errorID);
+		}
+		else if (repeat >= countof(s_Next))
```
gregory38: It would be better to put this check at the start of the function.

plugins/GSdx/GSDeviceDX.h
```diff
@@ -180,12 +170,14 @@ class GSDeviceDX : public GSDevice
 				uint32 point_sampler:1;
 				uint32 shuffle:1;
 				uint32 read_ba:1;
+
+				uint32 _free:32;
```
gregory38: Oh my bad you got 32 bits exactly. It would still be useful for future addition anyway :+1:

plugins/GSdx/GSDevice11.cpp
```diff
@@ -30,13 +30,21 @@ GSDevice11::GSDevice11()
...
 	FXAA_Compiled = false;
 	ExShader_Compiled = false;
 	
 	m_state.topology = D3D11_PRIMITIVE_TOPOLOGY_UNDEFINED;
 	m_state.bf = -1;
+
+	if (theApp.GetConfigB("UserHacks")) {
+		UserHacks_unscale_pt_ln = theApp.GetConfigB("UserHacks_unscale_point_line");
+	}
+	else {
```
gregory38: Please don't introduce yet another style in GSdx.

plugins/GSdx/GSDeviceDX.h
```diff
@@ -141,6 +158,7 @@ class GSDeviceDX : public GSDevice
 		operator uint32() {return key & 0x7;}
```
gregory38: The mask value is wrong.

plugins/GSdx/GSRendererDX11.cpp
```diff
@@ -27,6 +27,12 @@
 GSRendererDX11::GSRendererDX11()
 	: GSRendererDX(new GSTextureCache11(this), GSVector2(-0.5f))
 {
+	if (theApp.GetConfigB("UserHacks")) {
+		UserHacks_unscale_pt_ln = theApp.GetConfigB("UserHacks_unscale_point_line");
+	}
+	else {
```
gregory38: same comment as above on the style

plugins/GSdx/GSRendererDX11.cpp
```diff
@@ -37,10 +43,54 @@ bool GSRendererDX11::CreateDevice(GSDevice* dev)
...
+
+		t = D3D11_PRIMITIVE_TOPOLOGY_LINELIST;
+
+		break;
+	case GS_SPRITE_CLASS:
+		if (!m_vt.m_accurate_stq && m_vertex.next > 32) { // <=> 16 sprites (based on Shadow Hearts)
+
+			t = D3D11_PRIMITIVE_TOPOLOGY_LINELIST;
+		}
+		else {
```
gregory38: :)

plugins/GSdx/GSDevice11.cpp
FlatOutPS2: Please introduce clang in GSdx. :P

FlatOutPS2: Actually. I didn't introduce it. The other else's in this file are in the same style. :-)

gregory38: :cry: 

I was close to use clang. But I need to import 2 patches from the old Gabest branch (openCL). And there is also the issue with the VR stuff.

gregory38: You can't forget my comments on the style

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,53 @@ void vmfree(void* ptr, size_t size)
...
+
+	void* fifo = MapViewOfFile(s_fh, FILE_MAP_ALL_ACCESS, 0, 0, size);
+
+	for (size_t i = 1; i < repeat; i++) {
+		void* base = (uint8*)fifo + size * i;
+		s_Next[i] = (uint8*)MapViewOfFileEx(s_fh, FILE_MAP_ALL_ACCESS, 0, 0, size, base);
+		errorID = ::GetLastError();
+		if (s_Next[i] != base)
+			fprintf(stderr, "Failed to memory map contiguous segment, WIN API ERROR:%u\n", errorID);
+		break;
```
turtleli: I think you forgot to remove the `break`.

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,53 @@ void vmfree(void* ptr, size_t size)
...
-	// FIXME check linux code
-	return vmalloc(size * repeat, false);
+	ASSERT(fh == NULL);
+
+	if (repeat >= countof(s_Next))
+		fprintf(stderr, "Memory mapping overflow (%zu >= %u)\n", repeat, countof(s_Next));
+
+	s_fh = CreateFileMapping(INVALID_HANDLE_VALUE, nullptr, PAGE_READWRITE, 0, size * repeat, nullptr);
+	DWORD errorID = ::GetLastError();
+	if (s_fh == NULL)
```
turtleli: If it fails at this stage, the rest of the code in this function won't work as well (though I note that the non-Windows version isn't all that robust either).

gregory38: Well in this case, it would be better to make both versions robust.

plugins/GSdx/GSRendererDX.cpp
```diff
@@ -444,6 +446,54 @@ void GSRendererDX::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sourc
 		}
 	}
 
+	// Upscaling hack to avoid various line/grid issues
```
gregory38: Would it be be possible to move the code in a common function in the HW renderer ?

FlatOutPS2:
@gregory38 We still need tooltips for memory wrapping/OGL Geometry Shader/OGL Image Load Store/Merge Sprite. Are you going to add those later along with the Linux memory wrapping GUI?

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,54 @@ void vmfree(void* ptr, size_t size)
...
+static uint8* s_Next[8];
+
 void* fifo_alloc(size_t size, size_t repeat)
 {
-	// FIXME check linux code
-	return vmalloc(size * repeat, false);
+	ASSERT(s_fh == NULL);
+
+	if (repeat >= countof(s_Next))
+		fprintf(stderr, "Memory mapping overflow (%zu >= %u)\n", repeat, countof(s_Next));
```
gregory38: I think it would be better to return a `nullptr` in case of error. I know that it will still crash later as the `nullptr` isn't handled by the `fifo_alloc` caller.

gregory38:
@FlatOutPS2 
I'm fine with the code. Just a minor comment on error handling.

```
We still need tooltips for memory wrapping/OGL Geometry Shader/OGL Image Load Store/Merge Sprite. Are you going to add those later along with the Linux memory wrapping GUI?
```
English isn't my best skill. I let's you improve it. You will do a better job than me.
* memory wrapping
   * emulate GS memory wrapping accurately. It typically help games that miss some image block such as .... . It have a small impact on performance.
* OGL Geometry Shader
   * Allow to use the the GPU to transform line into sprite. Reduce CPU load and bandwidth requirement but it is heavier on the GPU. Automatic detection is recommended. (Note DX10+ GPU only)
* OGL Image Load Store
   * Allow to use advance atomic operation to speed up accurate date. Only disable it if you spot driver issue (Note DX11+ GPU only)
* Merge Sprite
   * Not yet working hack :) Replace post-processing multiple paving sprites by a single fat sprite. It typically help to reduce various upscaling lines.

mirh:
> Not yet working hack :) Replace post-processing multiple paving sprites by a single fat sprite. It typically help to reduce various upscaling lines.

Speak for yourself 😎
Anyway, seriously, it's already now something totally useful imo. 

Besides, is it normal that it can supplant align sprite hack altogether?

gregory38:
`Merge sprite` is kind of a big bertha version of `align sprite`. Potentially it would be just be better to extend `align sprite`.

plugins/GSdx/stdafx.cpp
```diff
@@ -124,10 +164,12 @@ void* fifo_alloc(size_t size, size_t repeat)
 
 	const char* file_name = "/GSDX.mem";
 	s_shm_fd = shm_open(file_name, O_RDWR | O_CREAT | O_EXCL, 0600);
-	if (s_shm_fd != -1)
+	if (s_shm_fd != -1) {
 		shm_unlink(file_name); // file is deleted but descriptor is still open
-	else
+	} else {
 		fprintf(stderr, "Failed to open %s due to %s\n", file_name, strerror(errno));
+		return nullptr;
```
gregory38: Thanks 👍

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,56 @@ void vmfree(void* ptr, size_t size)
...
+	if (s_fh == NULL) {
+		fprintf(stderr, "Failed to reserve memory. WIN API ERROR:%u\n", errorID);
+		return nullptr;
+	}
+
+	void* fifo = MapViewOfFile(s_fh, FILE_MAP_ALL_ACCESS, 0, 0, size);
+
+	for (size_t i = 1; i < repeat; i++) {
+		void* base = (uint8*)fifo + size * i;
+		s_Next[i] = (uint8*)MapViewOfFileEx(s_fh, FILE_MAP_ALL_ACCESS, 0, 0, size, base);
```
turtleli: From the documentation:

> If the lpBaseAddress parameter specifies a base offset, the function succeeds if the specified memory region is not already in use by the calling process.

So if this call fails some random memory that is already in use might get overwritten.

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,56 @@ void vmfree(void* ptr, size_t size)
...
-	// FIXME check linux code
-	return vmalloc(size * repeat, false);
+	ASSERT(s_fh == NULL);
+
+	if (repeat >= countof(s_Next)) {
+		fprintf(stderr, "Memory mapping overflow (%zu >= %u)\n", repeat, countof(s_Next));
+		return nullptr;
+	}
+
+	s_fh = CreateFileMapping(INVALID_HANDLE_VALUE, nullptr, PAGE_READWRITE, 0, size * repeat, nullptr);
```
turtleli: I think this only needs to be have a size of `size`?

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,73 @@ void vmfree(void* ptr, size_t size)
...
+	if (s_fh == NULL) {
+		fprintf(stderr, "Failed to reserve memory. WIN API ERROR:%u\n", errorID);
+		return vmalloc(size * repeat, false); // Fallback to default vmalloc
+	}
+
+	int mmap_segment_failed = 0;
+	void* fifo = MapViewOfFile(s_fh, FILE_MAP_ALL_ACCESS, 0, 0, size);
+	for (size_t i = 1; i < repeat; i++) {
+		void* base = (uint8*)fifo + size * i;
+		s_Next[i] = (uint8*)MapViewOfFileEx(s_fh, FILE_MAP_ALL_ACCESS, 0, 0, size, base);
```
turtleli: `s_Next` is kinda unnecessary - if everything succeeds, the memory allocations will be evenly spaced anyway, so the memory addresses for the repeating blocks can just be calculated from the initial block.

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,73 @@ void vmfree(void* ptr, size_t size)
...
+		errorID = ::GetLastError();
+		if (s_Next[i] != base) {
+			for (i; i >= 1; i--) {
+				UnmapViewOfFile(s_Next[i]);
+				s_Next[i] = 0;
+			}
+			mmap_segment_failed++;
+			i = 0;
```
turtleli: i is already 0 here?

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,73 @@ void vmfree(void* ptr, size_t size)
...
+	}
+
+	int mmap_segment_failed = 0;
+	void* fifo = MapViewOfFile(s_fh, FILE_MAP_ALL_ACCESS, 0, 0, size);
+	for (size_t i = 1; i < repeat; i++) {
+		void* base = (uint8*)fifo + size * i;
+		s_Next[i] = (uint8*)MapViewOfFileEx(s_fh, FILE_MAP_ALL_ACCESS, 0, 0, size, base);
+		errorID = ::GetLastError();
+		if (s_Next[i] != base) {
+			for (i; i >= 1; i--) {
```
turtleli: `i` at the init-statement condition does nothing. I would have done
```
do {
// blah
} while (--i > 0);
```
but that's just personal preference.

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,73 @@ void vmfree(void* ptr, size_t size)
...
+				UnmapViewOfFile(s_Next[i]);
+				s_Next[i] = 0;
+			}
+			mmap_segment_failed++;
+			i = 0;
+			fifo = MapViewOfFile(s_fh, FILE_MAP_ALL_ACCESS, 0, 0, size);
+		}
+
+		if (mmap_segment_failed > 4) {
```
turtleli: This could be moved to just after the variable is incremented.

plugins/GSdx/stdafx.cpp
```diff
@@ -71,16 +71,73 @@ void vmfree(void* ptr, size_t size)
...
+			fprintf(stderr, "Memory mapping failed after %d attempts, aborting. WIN API ERROR:%u\n", mmap_segment_failed, errorID);
+			fifo_free(fifo, size, repeat);
+			fifo = vmalloc(size * repeat, false); // Fallback to default vmalloc
+			break;
```
turtleli: `return vmalloc(size * repeat, false);`

plugins/GSdx/GSSetting.cpp
```diff
@@ -142,6 +140,18 @@ const char* dialog_message(int ID, bool* updateText) {
...
+				"Note: This hack can have a small impact on performance.";
+		case IDC_MERGE_PP_SPRITE:
+			return "Replaces post-processing multiple paving sprites by a single fat sprite. It reduces various upscaling lines.\n"
+				"Note: This hack is a work in progress.";
+		case IDC_GEOMETRY_SHADER_OVERRIDE:
+			return "Allows the GPU instead of just the CPU to transform lines into sprites. This reduces CPU load and bandwidth requirement, but it is heavier on the GPU.\n"
+				"Automatic detection is recommended.\nNote: This option is only supported by GPU's which support at least DirectX10.";
+		case IDC_IMAGE_LOAD_STORE:
+			return "Allows advanced atomic operations to speed up Accurate Date. Only disable this if using Accurate Date causes (GPU driver) issues.\n"
+				"Note: This option is only supported by GPU's which support at least DirectX11.";
```
turtleli: `GPUs`
`Direct3D11`

plugins/GSdx/GSSetting.cpp
```diff
@@ -142,6 +140,18 @@ const char* dialog_message(int ID, bool* updateText) {
...
 				"Avoids empty lines on the screen in games such as Ridge Racer V, and clears FMVs obscured by a grid in games like the Silent Hill series and Dirge of Cerberus.";
+		case IDC_MEMORY_WRAPPING:
+			return "Emulates GS memory wrapping accurately. This fixes issues where part of the image is cut-off by block shaped sections such as the FMVs in Wallace & Gromit: The Curse of the Were-Rabbit and Thrillville.\n"
+				"Note: This hack can have a small impact on performance.";
+		case IDC_MERGE_PP_SPRITE:
+			return "Replaces post-processing multiple paving sprites by a single fat sprite. It reduces various upscaling lines.\n"
+				"Note: This hack is a work in progress.";
+		case IDC_GEOMETRY_SHADER_OVERRIDE:
+			return "Allows the GPU instead of just the CPU to transform lines into sprites. This reduces CPU load and bandwidth requirement, but it is heavier on the GPU.\n"
+				"Automatic detection is recommended.\nNote: This option is only supported by GPU's which support at least DirectX10.";
```
turtleli: `GPUs`

And `Direct3D10` would probably be better (I remember there are people who will complain about this).

FlatOutPS2: The only thing they could complain about is that I didn't add a space between DirectX and 10. As far as I'm concerned DirectX makes more sense here. If you look at GPU spec lists they list support for the DirectX version, not Direct3D.

mirh: *May* have something to do with GPUs technically being embroiled in DirectWrite/Direct2D/DirectCompute too. 
If PCSX2 has nothing to do with [that](https://en.wikipedia.org/wiki/DirectX#Components), though, I guess Direct3D should be the proper naming.

gregory38:
Ok good. Yes we can improve later the robustness of the code.

gregory38:
@FlatOutPS2 good jobs :+1:

plugins/GSdx/GSdx.cpp
```diff
@@ -225,7 +225,7 @@ void GSdxApp::Init()
 
 	m_gs_hw_mipmapping.push_back(GSSetting(0, "Off", ""));
 	m_gs_hw_mipmapping.push_back(GSSetting(1, "Basic", "Fast"));
-	m_gs_hw_mipmapping.push_back(GSSetting(2, "Full", "Slow"));
+	m_gs_hw_mipmapping.push_back(GSSetting(2, "Full", "Very Slow - Not Recommended)"));
```
MrCK1: The bracket inside the quotes needs to be removed after Recommended, it's doubled in the dialog :)

PR #1854, opened 2017-03-11
gsdx ogl: add an HLE shader to avoid upscaling line in ICO

### LINKED ISSUE #1816: Ico white lines
***PCSX2 version:***

1.5.0-dev-1882-g213fa1c

***PCSX2 options:***
-Using OpenGL (Hardware) "Tested too with DX11 hardware, but the problem is worse"
-Tested with/without widescreen patch
-Tested with trilinear ultra and bilinear ps2 filtering
-Using mipmaping basic
-Half pixel offset hack
-3x Internal resolution
-Accurate date
-Blending unit accuracy set to High
-CRC level: Partial
-Interlacing: Auto

***Plugins used:***
AVX2

White lines all of time you are near a wall.

***PC specifications:***
I7 4790k 4.4 GHZ
GTX 970
16 GB DDR3

Note: 2 Images and 1 link with two videos attached (The link of the videos expire in 1 week)
ico bug 2
ico bug 1

Link with the videos: https://we.tl/P7nIh8FNzK

gregory38:
ICO uses a depth of field effect for the fog. Depth is extracted
into the alpha channel of a texture. And then used as blending factor.
You need a 1:1 texture/pixel mapping otherwise you will line at boundaries.

In order to extract the DoF, ICO moves the depth buffer around the GS
memory. Memory moves are implemented in the not-scaled world. It means
that we can't have the above 1:1 ratio. And we don't know anymore that
data are coming from the current depth buffer.

The solution: I added a dedicated shader to read the depth buffer directly.
This way I have the guarantee that pixel/depth are aligned.

Close #1816

### DIFF
--- a/plugins/GSdx/GSRendererOGL.cpp
+++ b/plugins/GSdx/GSRendererOGL.cpp
@@ -1494,6 +1494,40 @@ void GSRendererOGL::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 	// Always bind the RT. This way special effect can use it.
 	dev->PSSetShaderResource(3, rt);
 
+	if (m_game.title == CRC::ICO) {
+		GSVertex* v = &m_vertex.buff[0];
+		if (tex && m_vt.m_primclass == GS_SPRITE_CLASS && m_vertex.next == 2 && PRIM->ABE && // Blend texture
+				v[1].U == 8200 && v[1].V == 7176 && // at display resolution 512x448
+				tex->m_TEX0.PSM == PSM_PSMT8H) {  // i.e. read the alpha channel of a 32 bits texture
+			// Note potentially we can limit to TBP0:0x2800
+
+			// Depth buffer was moved so GSdx will invalide it which means a
+			// downscale. ICO uses the MSB depth bits as the texture alpha
+			// channel.  However this depth of field effect requires
+			// texel:pixel mapping accuraccy.
+			//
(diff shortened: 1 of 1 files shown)


gregory38:
@prafullpcsx2 a :gift: for you :smile:

gregory38:
I forget to say you need at least CRC hack on minimum.

FlatOutPS2:
> I forget to say you need at least CRC hack on minimum.

And below Full, otherwise the effect will be disabled.

gregory38:
Actually maybe I can reuse others part of shader
* PS_DEPTH_FMT == 1 or 2
* PS_CHANNEL_FETCH == 3

It would reduce the size of the PR.

prafullpcsx2:
Thanks @gregory38 .

This also fixes similar problem with Kya - Dark Lineage. I believe many other games suffer from this problem. I think instead of game specific crc it will be better to keep this in HW hacks.

gregory38:
Honestly I would be surprised it works well for other games. Tbh I even surprised that it didn't break ICO. So far, when we have a display texture that is blended, the hack will sample from the current depth instead of the texture. It "works" because the texture is the depth that was moved. However I don't have any guarantee that it isn't a color buffer instead.

Note for myself​: check that current texture isn't already a known color/depth buffers. Might avoid some bad cases.

lightningterror:
> Honestly I would be surprised it works well for other games.

Which other games do you think it can work for ?

PR #1858, opened 2017-03-14
GSdx ogl: SSO Workaround for AMD buggy drivers

### LINKED ISSUE #1343: [AMD GPU] Serious graphics problems all games opengl-hw recent builds
There's a bug that was only recent in newer builds. i Used to have blending accuracy on high and everything worked fine but now sometime since build pcsx2-v1.5.0-dev-511-gf1b3d7d. i cant play any games really unless i put it on ultra and that makes it painfully slow So i know for sure there is a error there because it used to work just fine until recently. I have a 8350fx and a 7970 xfx card all drivers up to date as well. ill post images as well of what all games do now on the newer builds unless i go ultra.
blending unit error
ultra

lightningterror:
Added Greg's SSO Workaround for AMD buggy drivers with some additions. Fixes #1343 
All 2017 drivers are now blacklisted.
The BSOD/crash issue is still there so don't set Blending Accuracy to None! Check #1552

### DIFF
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -335,12 +335,12 @@ namespace GLLoader {
 		fprintf(stdout, "OpenGL information. GPU: %s. Vendor: %s. Driver: %s\n", glGetString(GL_RENDERER), vendor, &s[v]);
 
 		// Name changed but driver is still bad!
-		if (strstr(vendor, "ATI") || strstr(vendor, "Advanced Micro Devices"))
+		if (strstr(vendor, "ATI") || strstr(vendor, "Advanced Micro Devices") || strstr(vendor, "ATI Technologies Inc."))
 			fglrx_buggy_driver = true;
 		if (fglrx_buggy_driver && (
-					strstr((const char*)&s[v], " 15.") || // blacklist all 2015 drivers
-					strstr((const char*)&s[v], " 16.") || // And all 2016 drivers. Good jobs AMD !
-					strstr((const char*)&s[v], " 17.1")   // In doubt take also first 2017 driver
+					strstr((const char*)&s[v], " 15.") || // Blacklist all 2015 AMD drivers.
+					strstr((const char*)&s[v], " 16.") || // Blacklist all 2016 AMD drivers.
+					strstr((const char*)&s[v], " 17.")   // Blacklist all 2017 AMD drivers for now.
 					))
 			legacy_fglrx_buggy_driver = true;
 
@@ -355,7 +355,7 @@ namespace GLLoader {
 		mesa_buggy_driver = !nvidia_buggy_driver && !fglrx_buggy_driver;
 #endif
 
-		buggy_sso_dual_src = intel_buggy_driver || legacy_fglrx_buggy_driver;
+		buggy_sso_dual_src = intel_buggy_driver || fglrx_buggy_driver || legacy_fglrx_buggy_driver;
 
 		if (theApp.GetConfigI("override_geometry_shader") != -1) {
 			found_geometry_shader = theApp.GetConfigB("override_geometry_shader");
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -437,8 +437,9 @@ namespace GLLoader {
 		status &= status_and_override(found_GL_ARB_get_texture_sub_image, "GL_ARB_get_texture_sub_image");
 
 		if (fglrx_buggy_driver) {
-			fprintf(stderr, "The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver. Check out the link below for further information.\n"
-					"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know\n");
+			fprintf(stderr, "The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver.\n"
+			"Check out the link below for further information.\n"
+			"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know\n");
 		}
 
 		if (!found_GL_ARB_viewport_array) {
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -335,12 +335,12 @@ namespace GLLoader {
 		fprintf(stdout, "OpenGL information. GPU: %s. Vendor: %s. Driver: %s\n", glGetString(GL_RENDERER), vendor, &s[v]);
 
 		// Name changed but driver is still bad!
-		if (strstr(vendor, "ATI") || strstr(vendor, "Advanced Micro Devices") || strstr(vendor, "ATI Technologies Inc."))
+		if (strstr(vendor, "Advanced Micro Devices") || strstr(vendor, "ATI Technologies Inc.") || strstr(vendor, "ATI"))
 			fglrx_buggy_driver = true;
 		if (fglrx_buggy_driver && (
 					strstr((const char*)&s[v], " 15.") || // Blacklist all 2015 AMD drivers.
 					strstr((const char*)&s[v], " 16.") || // Blacklist all 2016 AMD drivers.
-					strstr((const char*)&s[v], " 17.")   // Blacklist all 2017 AMD drivers for now.
+					strstr((const char*)&s[v], " 17.") // Blacklist all 2017 AMD drivers for now.
 					))
 			legacy_fglrx_buggy_driver = true;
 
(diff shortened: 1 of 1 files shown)


mirh:
A better naming for what you are doing is "expand blacklist of AMD driver". 
As I was kind of saying in #1552 though.. What's the point in separate fglrx and legacy_fglrx ?

gregory38:
The initial plan was to have a half broken legacy driver and a good driver. I didn't expect that supposed driver will be a completely broken driver.... Seriously it would be faster and honest for them to say that they don't support openGL

mirh:
Oh, ok, so there's no difference at least atm. 

You could always put 16.2 and 16.3 to fglrx_buggy, which in turns only just disable "none" accuracy though.

gregory38:
```
which in turns only just disable "none" accuracy though.
```
I don't remember we have some code that force BA for anyone.

lightningterror:
Even with that there are still issues , when using OpenGL the emulator crashes a lot.

mirh:
*stares at coders*

lightningterror:
On GT4 I seem to get them at certain parts depending on the maps.

plugins/GSdx/GLLoader.cpp
```diff
@@ -335,12 +335,12 @@ namespace GLLoader {
 		fprintf(stdout, "OpenGL information. GPU: %s. Vendor: %s. Driver: %s\n", glGetString(GL_RENDERER), vendor, &s[v]);
 
 		// Name changed but driver is still bad!
-		if (strstr(vendor, "ATI") || strstr(vendor, "Advanced Micro Devices"))
+		if (strstr(vendor, "ATI") || strstr(vendor, "Advanced Micro Devices") || strstr(vendor, "ATI Technologies Inc."))
```
FlatOutPS2: Very minor note: It would be nicer if `"ATI"` and `"ATI Technologies Inc."` were next to each other. :P

lightningterror: Done however I won't push it yet , I want to make some other changes.

I noticed the gs version is 1.1.0 shown in the emulator , however checking the dll file version and product version it's 1.0.1.9  
Should this be updated ?

lightningterror:
@FlatOutPS2 Done.
@mirh Here is the error log.
[errorlog.txt](https://github.com/PCSX2/pcsx2/files/845788/errorlog.txt)

mirh:
Tbh, windgb report of the memory dump would have been way more helpful. 
And I just tried very briefly GT4 (slideshows are hard to bear) and it seemed fine

lightningterror:
You mean this? 
`Unhandled` exception at 0x3000348D in pcsx2.exe: 0xC0000005: Access violation reading location 0x0034D203

If there is a handler for this exception, the program may be safely continued..

mirh:
I mean, attach windbg to pcsx2, then when it crashes it *should* prompt you.

lightningterror:
With or without Heap Information ? With heap it's about 1gb which will take a while to upload.

gregory38:
You're kidding me. I want touch AMD/win stuff even      with a 10 foot pole.

lightningterror:
Greg right now xD
http://i.imgur.com/ExBLAI3.jpg

Anyway the PR is ready for merge.

mirh:
> Greg right now xD

You miss the Windows logo. 

> Anyway the PR is ready for merge.

*had suggested improvements*

gregory38:
Yes Seriously ! Where is the TROLL and the click bait.

gregory38:
Is it fine for everybody ?

mirh:
In the end I guess those that will keep <17.2 drivers just for this will be on the same order of magnitude of those knowing and using `override_GL_ARB_separate_shader_objects` for testing regardless.. So, whatever.

FlatOutPS2:
We should do something about the messages printed for OpenGL. Every time you switch renderer(especially annoying when switching between OGL SW and HW with the F9 key), it prints more than 15 lines of OpenGL information in the log on Windows. Whereas on Linux, the driver information is only printed once when first booting PCSX2.

MrCK1:
Maybe don't print GL extensions unless they're detected as unsupported?

gregory38:
What are the 15 lines printed?

FlatOutPS2:
> What are the 15 lines printed?

All the extensions and other normal driver information are printed every time, and then there's this:

```
The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver. Check out the links below for further information.
https://community.amd.com/message/2756964
https://community.amd.com/thread/205702
Note: Due to an AMD OpenGL driver issue, setting Blending Unit Accuracy to "None" can cause an application or system crash.
Keep Blending Unit Accuracy set to at least the default "Basic" level.
AMD has a fix for the issue that will be released in the coming months. The issue does not affect AMD GPUs on legacy drivers.
```

mirh:
I'd say that *even now*, chances are most people still wouldn't notice it. 
Also, tbh.. Is that really all of that problem if you aren't a *compulsive tester*?

FlatOutPS2:
> Also, tbh.. Is that really all of a problem if you aren't a compulsive tester?

Yes, because it's so much information it can actually hide important messages printed in the log.

lightningterror:
I would also recommend shortening the message to something like this.
```
The OpenGL hardware render is slow on AMD GPUs due to an inefficient driver. Check out the link for further information.
https://github.com/PCSX2/pcsx2/issues/1552
```

And also editing the first post of the git issue with all the needed information required.

mirh:
> Yes, because it's so much information it can actually hide important messages printed in the log.

Ok, then it's only a problem for testers 😛 

> And also editing the first post of the git issue with all the needed information required.

A [wiki](https://github.com/PCSX2/pcsx2/wiki) page would probably fit the purpose better.

lightningterror:
> A wiki page would probably fit the purpose better.

That's even better.

FlatOutPS2:
>Ok, then it's only a problem for testers 😛

It could also make regular users miss some useful log information too. In any case it working like the Linux version would be ideal.

> Check out the link for further information.

I'd leave the `below` in there.

lightningterror:
Okay I created the wiki page. Feel free to edit the info since I know there is a lot of stuff I missed probably.

Anyway moving back on topic how does this sound?
```
Detected inefficient AMD OpenGL driver. Check the link below for further information.
https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know
```

Or the old one
```
The OpenGL hardware render is slow on AMD GPUs due to an inefficient driver. Check out the link for further information.
https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know
```

FlatOutPS2:
The old one is better. The new one is very ambiguous.

mirh:
For the moment (EVERY amd driver sucks), I'd keep the later. 
When they'll finally have a working one.. "detected bad version" will be better. 

Also, change inefficient to buggy.

gregory38:
A working driver might never happen (quite likely)

FlatOutPS2:
> A working driver might never happen (quite likely)

It will happen, but likely not the driver made by AMD.

Besides, once they fix the current BSOD issue and the recent regression, and GSdx gets multi-threading it should already be much better.

gregory38:
Not sure what you mean by "not by AMD". The opensource driver is fine. It will even get MT support once I got back from my vacation to finish some patches.

mirh:
If you are so fervent on mesa, the HD 6310 on my laptop is always ready to be ssh'd. 😄

gregory38:
@mirh I can run `nouveau` on my Nvidia GPU. I alternate between the 2 drivers when I reboot but I barely reboot my computer those days.

So are we fine with this PR ?

mirh:
I was actually more referring to your goodwillingness to work on mesa, rather than pcsx2 specifically 
nvc0 is almost too boring to develop now, while r600 has still a long road ahead :p

Anyway, yes fine for me. 
Pretty nice you don't have to make a PR to edit the wiki entry.

gregory38:
I think r600 is fine with PCSX2. My goal is to have a good experience with PCSX2 on Mesa not to provide a nice driver for everyone ;)

gregory38:
@FlatOutPS2 is it ok for you ?

plugins/GSdx/GLLoader.cpp
```diff
@@ -437,12 +437,8 @@ namespace GLLoader {
 		status &= status_and_override(found_GL_ARB_get_texture_sub_image, "GL_ARB_get_texture_sub_image");
 
 		if (fglrx_buggy_driver) {
-			fprintf(stderr, "The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver. Check out the links below for further information.\n"
-					"https://community.amd.com/message/2756964\n"
-					"https://community.amd.com/thread/205702\n"
-					"Note: Due to an AMD OpenGL driver issue, setting Blending Unit Accuracy to \"None\" can cause an application or system crash.\n"
-					"Keep Blending Unit Accuracy set to at least the default \"Basic\" level.\n"
-					"AMD has a fix for the issue that will be released in the coming months. The issue does not affect AMD GPUs on legacy drivers.\n");
+			fprintf(stderr, "The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver. Check out the link below for further information.\n"
```
FlatOutPS2: It looks better if you replace the space between the sentences with `\n`.

lightningterror:
All good now?

FlatOutPS2:
As far as I'm concerned it can be merged now.

PR #1867, opened 2017-03-17
gsdx: disable GCC optimization when checking SSE/AVX support

### LINKED ISSUE #1859: Linux: Crash caused by presence of libGsdx-AVX2.so plugin on non-AVX CPU
Hey all,

I'm running Ubuntu 16.10 x64 (w/ GTX 780 & 378.13 drivers) on an i7 930 CPU. My system is running the latest version of PCSX2 from the official PPA.

If I open the Plugin/BIOS selector, PCSX2 hangs for a few seconds and crashes. The Stack Trace, from the Ubuntu error report, indicates an issue between the **libGsdx-AVX2.so** plugin and **/lib/ld-linux.so.2** (_/lib/i386-linux/gnu/ld-2.24.so_). When running PCSX2 via a terminal, the last entry recorded after crashing is **Illegal instruction (core dumped)**.

This has occurred over the builds i've tested during the last month or so. So far, deleting the plugin stops the crash from happening.

If there's any other specific information anyone would like, let me know.

gregory38:
Fix #1859 (not tested)

### DIFF
--- a/plugins/GSdx/xbyak/xbyak_util.h
+++ b/plugins/GSdx/xbyak/xbyak_util.h
@@ -82,11 +82,28 @@ extern "C" unsigned __int64 __xgetbv(int);
 
 namespace Xbyak { namespace util {
 
+/* GCC uses AVX/SSE4 operation to handle the uint64 type.
+ *
+ * It is quite annoying because the purpose of the code is to test the support
+ * of AVX/SSEn
+ *
+ * So far, we don't need other ISA on i386 so I hacked the code to limit the
+ * type to 32 bits. If we want to support AVX512 we might need to shuffle the
+ * code a bit.
+ *
+ * Extra note: it would be waste to use AVX512 on 32 bits, registers are
+ * limited to 8 instead of 32.
+ */
+
 /**
 	CPU detection class
 */
 class Cpu {
+#ifdef XBYAK64
 	uint64 type_;
+#else
+	uint32 type_;
+#endif
 	unsigned int get32bitAsBE(const char *x) const
 	{
 		return x[0] | (x[1] << 8) | (x[2] << 16) | (x[3] << 24);
@@ -155,7 +172,11 @@ public:
 		return ((uint64)edx << 32) | eax;
 #endif
 	}
+#ifdef XBYAK64
 	typedef uint64 Type;
+#else
+	typedef uint32 Type;
+#endif
 	static const Type NONE = 0;
 	static const Type tMMX = 1 << 0;
 	static const Type tMMX2 = 1 << 1;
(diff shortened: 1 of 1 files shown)


turtleli:
Doesn't work on GCC 6.3.

My previous analysis was wrong, it's not the initialisation which uses vector instructions, its actually `if(!g_cpu.has(checks[i].type))` (it's still related to the array size though).

gregory38:
Oh. I'm curious what does GCC generate ?

turtleli:
See https://godbolt.org/g/WRhm6h

gregory38:
Thanks you. Ok it is related to the 64 bits type.

turtleli:
Latest commit doesn't fix the issue. If you add the pragmas to the previous godbolt link and choose the "compile to binary and disassemble the output" option you'll still see VEX encoded instructions (I also checked the disassembly on Fedora).

Side note: pragmas need to be ifdef'd - AppVeyor shows 300+ warnings ;)

gregory38:
where is this option `"compile to binary and disassemble the output"` On the right panel I can see the asm. Code is SSE4 (which doesn't solve fully the issue). But AVX is gone. 
I think I will remove the uint64 type.

turtleli:
> where is this option "compile to binary and disassemble the output"

It's the 11010 near the top of the asm panel.

> Code is SSE4 (which doesn't solve fully the issue). But AVX is gone.

It seems that if -mavx is in the command line then it'll use the VEX encoded instructions even if the pragma is there.

gregory38:
Hum... I will just use an uint32 type.

gregory38:
So I limit the check to 32 bits. It would be annoying to support AVX512 on i386 but it won't happen soon. There is an assert to avoid the pitfall. Otherwise the other solution will be to put intrinsics to do the check ourself with SSE2.

turtleli:
Seems to work for CheckSSE. However, I managed to reproduce the crash on accessing the plugin/BIOS selector panel (I think no-simd and native avx2 compilation produces different code for me, so I didn't see the issue the first time). It seems the initialisation of `g_cpu` can use avx instructions.

BTW, the GSdx GUI height is too big for a x768px screen - I can't see the ok button on my VM (set to 1024x768), and that's with the logo missing too.

The logo actually doesn't show up on some distros (I think it's because the pixdata format loader has been removed from recent gdk-pixbuf versions), though I can fix that (well, I still think all the GSdx logos should just be removed).

gregory38:
Of course, there are plenty of uint64 in the constructor... Geez... 

Which tab is too big ? I guess the HW hack. It is strange that it doesn't fit in a 768p screen.
For the logo, I guess it misses a dependency on cmake. It would be nice to fix it. Not for the logo, but for the embedded shaders (current PERL solution is quite dirty)

turtleli:
> Which tab is too big ? I guess the HW hack.

Postprocessing/OSD, but the HW hack tab might also be too big.

gregory38:
@turtleli could you check the new commit. I will do a separate PR to reduce the GUI size, potentially with a resync on latest flatout addition.

turtleli:
Works on my end.

Issue #1862, opened 2017-03-15
SVC Chaos - Main menu invisible in Hardware mode


FlatOutPS2:
***PCSX2 version:***
v1.5.0 dev-1927

***PCSX2 options:***
Default

***Plugins used:***
Default

***Description of the issue:***
The main menu is completely black in Hardware mode. This is a regression following this commit:
https://github.com/PCSX2/pcsx2/commit/de38963904140259ca1b398234e04931f545bbf8
Specifically this line: `tex = tex_psm.depth ? m_tc->LookupDepthSource(context->TEX0, env.TEXA, r) : m_tc->LookupSource(context->TEX0, env.TEXA, r);`

***How to reproduce the issue:***
Set Renderer to any of the Hardware renderers, boot the game and advance to the main menu.

***Last known version to work:***
v1.5.0-dev-640

gregory38:
Could it be linked to #1838 ? Does `LookupDepthSource` throw or return a bad texture ?

MrCK1:
@FlatOutPS2 Care to retest this?

FlatOutPS2:
@MrCK1 see https://github.com/PCSX2/pcsx2/pull/1868

lightningterror:
Perhaps CanConvertDepth() check should be added to that line and also reuse the old code when depth isn't supported to fix the regression on D3D ?

Immersion95:
This is still happening in the latest build with any HW renderer in the menu of those games : 
-SVC Chaos
-KOF2002

Should I open another issue ?

refractionpcsx2:
No I'll just reopen this one.

Immersion95:
> No I'll just reopen this one.
> 
> Can you post your CRC for SVC?

Here it is : 
- SVC Chaos 94834BD3
- KOF 2002 ABD16263

refractionpcsx2:
I'm not sure if KOF 2002 is the same thing (we've never had a CRC entry for that) but your version of SVC Chaos is missing.

Can you give me the Serial/Region of your version of SVC Chaos so we can add it please

Immersion95:
It is this one http://redump.org/disc/37370/

refractionpcsx2:
Okay, your serial differs from the normal serial, is your version modified?

Immersion95:
> Okay, your CRC differs from the normal CRC, is your version modified?

Mine is not and matches exactly the one from redump. I converted my CHD to BIN/CUE and I have the same CRC.

[...]

Issue #1838, opened 2017-02-24
[GSDX-TC] Regression in Jackie Chan Adventures


ssakash:
Jackie Chan Adventures doesn't seem to like the skip of drawcall (https://github.com/PCSX2/pcsx2/commit/87fc4c1e44cfa7215cf1b327e0c0f93cadfb5bda) in #1760 , reverting that and adding back ``return LookupSource(TEX0, TEXA, r);`` seems to fix the half screen flickering issue in Jackie Chan Adventures. FYI, this game also enters the half right path which is there for the Snowblind games.

Before #1760:

![before](http://i.imgur.com/OqW64D8.png)

After #1760:

![after](http://i.imgur.com/4jcpoRe.png)

_**GSDump**_: [_**here**_](https://www.dropbox.com/s/8k47jq8h9qvbykp/JackieRegression.7z?dl=0)

FlatOutPS2:
> FYI, this game also enters the half right path which is there for the Snowblind games.

What happens if you block it?

ssakash:
> What happens if you block it?

Issue still exists even if you block the special case for Snowblind games. (I just mentioned about the ``half_right`` code path in case it might be useful)

gregory38:
Hum, it could just require to handle the half case when we lookup the depth. I won't touch it now as I would like a more generic solution.

gregory38:
@ssakash @FlatOutPS2 What about a crc hack with a big comment message to explain the situation.

lightningterror:
There still seems to be some issues to iron out on D3D. The language select background is black instead of yellow.

Regression PR #1760

lightningterror:
Fixed on d3d11.

PR #1868, opened 2017-03-17
gsdx tc: revert 87fc4c1e44cfa7215cf1b327e0c0f93cadfb5bda

gregory38:
At least 2 regressions
* SVC Chaos #1862
* Jackie Chan #1838

I don't see any difference on Full Spectrum Warrior gsdump (#1757). Both return
and throw flicker. And none has the original vertical line issue. I suspect
it is linked to the behavior of depth miss.

The correct fix would be to trace all textures writes to be sure of the source.
But it is a much bigger work.

@FlatOutPS2 @ssakash 
Could you test those 3 games
* [ ] SVC Chaos
* [ ] Jackie Chan
* [ ] Full Spectrum Warrior

### DIFF
--- a/plugins/GSdx/GSTextureCache.cpp
+++ b/plugins/GSdx/GSTextureCache.cpp
@@ -175,10 +175,10 @@ GSTextureCache::Source* GSTextureCache::LookupDepthSource(const GIFRegTEX0& TEX0
 		// The game tries to emulate a texture shuffle with an old depth buffer
 		// (don't exists yet for us due to the cache)
 		// Rendering is nicer (less garbage) if we skip the draw call.
-		throw GSDXRecoverableError();
+		//throw GSDXRecoverableError();
 
 		//ASSERT(0);
-		//return LookupSource(TEX0, TEXA, r);
+		return LookupSource(TEX0, TEXA, r);
 	}
 
 	return src;


FlatOutPS2:
PR does the following:
SVC Chaos: It fixes the black screen on OpenGL HW, the regression is still there with the Direct3D HW renderers.
Jackie Chan: Half-screen flickering is gone, half-black screen issue is still there.
FSW: Reintroduced the bugs that were fixed.

gregory38:
SVC: D3D will kill me. The only solution will be to port the depth shader.
JC: yes. The real half bug is still there.
FSW: strangely I don't see it on the gsdump. There is a green flickering in both mater and this branch.

MrCK1:
What are we doing about this PR? It would be a good idea to decide soon if we need to postpone this before the 1.6 release.

lightningterror:
> `What about a crc hack with a big comment message to explain the situation.`

I was about to suggest the same thing, but then I saw your comment :+1: 
It's been a while but what are we gonna do with this ?

gregory38:
The solution would be to implement a better search of depth buffer. With the help of the texture offset, we should be able to fix it properly (along with half bottom issue).

FlatOutPS2:
The solution would be to implement a better search of depth buffer. With the help of the texture offset, we should be able to fix it properly (along with half bottom issue).

@gregory38 Are you still going to work on that solution? Otherwise this should probably get merged.

gregory38:
```
@gregory38 Are you still going to work on that solution? Otherwise this should probably get merged.
```
I'm way too busy with real life thing. I'm not sure this PR improve rendering (fix some games, broke others). If you feel that it is better, it could be merged.

FlatOutPS2:
> I'm way too busy with real life thing. I'm not sure this PR improve rendering (fix some games, broke others). If you feel that it is better, it could be merged.

AFAIK it only breaks the rendering on Full Spectrum Warrior again. And that regression isn't nearly as bad as the current Jackie Chan regression, or even the one in SVC. We could always add a crc based workaround for FSW if a better solution doesn't come around in the near future.

refractionpcsx2:
So, what are we going to do with this?

lightningterror:
Sure, which code should be better to be on by default if crc is not detected ?

Currently I have it like this, need to include svn chaos.

> 		if (m_renderer->m_game.title == CRC::JackieChanAdv)
> 			return LookupSource(TEX0, TEXA, r);
> 		else
> 			throw GSDXRecoverableError();

MrCK1:
Did this PR ever fix anything major in the first place? I'll have to do some testing when I get a chance I guess.

lightningterror:
Close in favor of #2382

Issue #1762, opened 2017-01-11
Ar Tonelico 2 - World Map renders incorrectly


orbea:
OS: `Slackware64-current`
`mesa-compat32-2016.12.29_b7699ce07c_master-x86_64-1_gitcompat32`
`pcsx2-2017.01.10_62c21fd58_master-x86_64-1_git`

In the game Ar Tonelico 2 (Patched with the Project Metafalica retranslation) the world map will render incorrectly. I've tried various options such as the CRC level and hacks which failed to make a difference. I also tried as far back as commit e329b6f and also found the same problem.

Gsdx dump - http://ks392457.kimsufi.com/orbea/stuff/games/pcsx2/snaps/gsdx_20170111150438_Ar_Tonelico_2_project_metafalica.gs.xz
Apitrace - http://ks392457.kimsufi.com/orbea/stuff/trace/PCSX2_Ar_Tonelico_2_project_matafalica.trace.xz

![1](http://ks392457.kimsufi.com/orbea/stuff/pics/scrots/pcsx2/2017-01-11-151210_1680x1050_scrot.png)

FlatOutPS2:
Try it without applying the patch.

orbea:
I just did, same result. Someone else was also able to reproduce with the apitrace using intel, so its probably either the mesa-core or pcsx2 itself.

gregory38:
GSdx uses a CRC hack for this game. We need to check if it is enabled. What is your game CRC ?

orbea:
Here is the crc32 for both isos I tried.
```
027a757f Ar_tonelico_2_Project_Metafalica.iso
49205494 ar_tonelico_2_us.iso
```
The latter (Unpatched iso) should be a redump as the md5sum, sha1sum and crc32 matches the following link.
http://redump2.rssing.com/chan-20787209/all_p337.html

md5sum - `1b83937656fd8fede5c0fbc8dbc11f8f`
sha1sum - `df94f6a165cfc829821a5a18c1e3bf650624de52`

To be clear I tried CRC hack levels none, partial, full and aggressive without any difference. Additionally loading the redump iso will print.
`Patches: No CRC found, using 00000000 instead.`

gregory38:
GSdx CRC are different. Anyway they don't match.
{0xF46142D3, ArTonelico2, NoRegion, 0},

gregory38:
File is GScrc.cpp, you can edit it and recompile it ;)

orbea:
[...]

PR #1872, opened 2017-03-21
Add crc hack for project metafalica

orbea:
Closes https://github.com/PCSX2/pcsx2/issues/1762

Given that all three versions of the patch should result in the same iso this should be `NoRegion`, this CRC was tested with an US patch applied to an iso matching the US redump.

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -27,6 +27,7 @@ CRC::Game CRC::m_games[] =
 {
 	{0x00000000, NoTitle, NoRegion, 0},
 	{0xF46142D3, ArTonelico2, NoRegion, 0},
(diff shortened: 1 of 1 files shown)


gregory38:
Thanks you.

gregory38:
Note. I know that we want to avoid patched version on the CRC hack section. But I already got several bug report for this game. So I grant this game an exception.

FlatOutPS2:
From the changelog of the project it seems like it contains (content from) copyrighted files, which would make it piracy.

gregory38:
Please, let avoid a witch hunting. I just checked the website. As far as I'm concerned, they're just fan redoing the translation. They provide only the patch to an original ISO not the ISO itself (I didn't check the patch content before you ask). We are far from users that don't want to pay theirs games or trying to make money with hacked version. Not my role to judge them and we don't support them. Just added a CRC to fix an initial game issue.

PR #1874, opened 2017-03-24
gsdx: Add an automatic CRC hack level

turtleli:
Changes:
* Fixes a bug where depth emulation is disabled if the user starts with the OpenGL software renderer and F9s to the hardware renderer.
* Adds an automatic CRC hack level, which will autoselect the recommended CRC hack level for the selected renderer. I used some stuff from willkuer's PR (#1686).

### DIFF
--- a/plugins/GSdx/GSSetting.cpp
+++ b/plugins/GSdx/GSSetting.cpp
@@ -46,10 +46,14 @@ const char* dialog_message(int ID, bool* updateText) {
 #endif
 		case IDC_CRC_LEVEL:
 			return "Control the number of Auto-CRC hacks applied to games.\n\n"
+				"Automatic:\nAutomatically sets the recommended CRC hack level based on the selected renderer.\n"
+				"This is the recommended setting.\n"
+				"Partial will be selected for OpenGL.\nFull will be selected for Direct3D.\n\n"
 				"None:\nRemove nearly all CRC hacks (debug only).\n\n"
 				"Minimum:\nEnable a couple of CRC hacks (23).\n\n"
-				"Partial:\nEnable most of the CRC hacks.\nRecommended OpenGL setting (Accurate/depth options may be required).\n\n"
-				"Full:\nEnable all CRC hacks.\nRecommended Direct3D setting.\n\n"
+				"Partial:\nEnable most of the CRC hacks.\n"
+				"For an optimal experience with OpenGL, Blending Unit Accuracy/Depth Emulation may need to be enabled.\n\n"
+				"Full:\nEnable all CRC hacks.\n\n"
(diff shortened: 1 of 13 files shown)


FlatOutPS2:
Is this basically still the same as @ssakash's implementation of @willkuer's PR? As in it does nothing unless you have an NVIDIA GPU and boot PCSX2 for the very first time/without existing inis?

turtleli:
No?

There's a new "Automatic" CRC hack level (which has been set to the new default) - users with an existing ini will have to manually select it.

plugins/GSdx/GSSetting.cpp
```diff
@@ -46,6 +46,7 @@ const char* dialog_message(int ID, bool* updateText) {
 #endif
 		case IDC_CRC_LEVEL:
 			return "Control the number of Auto-CRC hacks applied to games.\n\n"
+				"Automatic:\nAutomatically select 'Partial' or 'Full' level based on the selected renderer.\n\n"
```
FlatOutPS2: It should clearly state which option is automatically selected for each renderer. It's a bit vague now.

turtleli: Well, I like the vagueness. The string doesn't need to be updated as much that way. How about:

> Automatically select the recommended CRC hack level based on the selected renderer. Recommended setting.
Partial will be chosen for OpenGL.
Full will be chosen for Direct3D.

FlatOutPS2: I'd prefer:
```
Automatically sets the recommended CRC hack level based on the selected renderer. 
This is the recommended setting.
Partial will be selected for OpenGL.
Full will be selected for Direct3D.
```

You can also remove the recommended at the Full/Partial levels of the tooltip, as it is now printed in the combobox for both options. And replace `Accurate/depth options may be required` with `Blending Unit Accuracy/Depth Emulation may be need to be enabled` at OpenGL.

turtleli: > And replace `Accurate/depth options may be required with Blending Unit Accuracy/Depth Emulation may be need to be enabled` at OpenGL.

I'd need to keep the recommended info in the tooltip if I'm to make this change. Otherwise I think it wouldn't make as much sense.

FlatOutPS2: Adding `(only available for the OpenGL renderer)` to that line would resolve that.

gregory38:
I'm fine with the code. 👍

FlatOutPS2:
Has anyone actually tested it to be sure it always works properly? :p

gregory38:
I trust turtleli :)

PR #1875, opened 2017-03-24
PCSX2: Console log update

FlatOutPS2:
Remove full path from every plugin binding and only add it to the preceding "Loading Plugins..." line.
Instead the revision date and version number are printed for each plugin to make it easier to identify issues with plugin versions.

And improve the widescreen patch message when loading from the cheats_ws folder instead of the archive.

And add a very minor compatibility info update in the GameIndex file for 2 games bassed on compatibility reports I made earlier.

EDIT: And adds fix for Sled Storm(confirmed to be required for both the PAL(SLES-50683) and NTSC-U(SLUS-20363) versions).

### DIFF
--- a/pcsx2/PluginManager.cpp
+++ b/pcsx2/PluginManager.cpp
@@ -1126,8 +1126,11 @@ void SysCorePlugins::Load( PluginsEnum_t pid, const wxString& srcfile )
 {
 	ScopedLock lock( m_mtx_PluginStatus );
 	pxAssert( (uint)pid < PluginId_Count );
-	Console.Indent().WriteLn( L"Binding %4s: %s ", WX_STR(tbl_PluginInfo[pid].GetShortname()), WX_STR(srcfile) );
+
 	m_info[pid] = std::unique_ptr<PluginStatus_t>(new PluginStatus_t(pid, srcfile));
+
+	Console.Indent().WriteLn(L"Bound %4s: %s [%s %s]", WX_STR(tbl_PluginInfo[pid].GetShortname()), 
+		WX_STR(wxFileName(srcfile).GetFullName()), WX_STR(m_info[pid]->Name), WX_STR(m_info[pid]->Version));
 }
 
 void SysCorePlugins::Load( const wxString (&folders)[PluginId_Count] )
@@ -1135,8 +1138,8 @@ void SysCorePlugins::Load( const wxString (&folders)[PluginId_Count] )
 	if( !NeedsLoad() ) return;
 
 	wxDoNotLogInThisScope please;
-	
-	Console.WriteLn( Color_StrongBlue, "\nLoading plugins..." );
+
+	Console.WriteLn(Color_StrongBlue, L"\nLoading plugins from %s...", WX_STR(g_Conf->Folders[FolderId_Plugins].ToString()));
 
 	ConsoleIndentScope indent;
 	const PluginInfo* pi = tbl_PluginInfo; do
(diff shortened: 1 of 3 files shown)


mirh:
>PluginManager.cpp: In member function ‘virtual void SysCorePlugins::Load(PluginsEnum_t, const wxString&)’:

> PluginManager.cpp:1135:302: error: conditional expression between distinct pointer types ‘const char*’ and ‘const wchar_t*’ lacks a cast [-fpermissive]

  >  Console.Indent().WriteLn(L"Binding %4s: %s (%s %d.%d.%d) ", WX_STR(tbl_PluginInfo[pid].GetShortname()), WX_STR(srcfile.AfterLast('\\')), lib_name.Find("20") != -1 ? WX_STR(lib_name.Mid(lib_name.Find("20"), 14)) : L"Unknown", (lib_version >> 8) & 0xff, lib_version & 0xff, (lib_version >> 24) & 0xff);

pcsx2/PluginManager.cpp
```diff
@@ -1126,20 +1126,25 @@ void SysCorePlugins::Load( PluginsEnum_t pid, const wxString& srcfile )
...
 {
 	if( !NeedsLoad() ) return;
 
 	wxDoNotLogInThisScope please;
-	
-	Console.WriteLn( Color_StrongBlue, "\nLoading plugins..." );
+
+	const PluginInfo* pi = tbl_PluginInfo;
+
+	Console.WriteLn(Color_StrongBlue, "\nLoading plugins from %S...", WX_STR(wxFileName(folders[pi->id]).GetPath()));
```
turtleli: `L"\nLoading plugins from %s..."` (or at least that works on Linux).

The `\n` at the start doesn't really need to be there IMO.

gregory38: Yes I confirm, it is better to use a wide-char string.

FlatOutPS2: > L"\nLoading plugins from %s..." (or at least that works on Linux).

%s does not work on Windows, otherwise I would've used it.

turtleli: > %s does not work on Windows, otherwise I would've used it.

Did you try what I suggested? The `L` makes a lot of difference.

pcsx2/gui/AppCoreThread.cpp
```diff
@@ -446,7 +446,7 @@ static void _ApplySettings( const Pcsx2Config& src, Pcsx2Config& fixup )
 		if (int numberLoadedWideScreenPatches = LoadPatchesFromDir(gameCRC, GetCheatsWsFolder(), L"Widescreen hacks"))
 		{
 			gameWsHacks.Printf(L" [%d widescreen hacks]", numberLoadedWideScreenPatches);
-			Console.WriteLn(Color_Gray, "Found ws patches at cheats_ws --> skipping cheats_ws.zip");
+			Console.WriteLn(Color_Gray, "Found widescreen patches in the cheats_ws folder --> skipping cheats_ws.zip");
```
turtleli: Personally would prefer this change as a separate commit, so that there's an "improve plugin console messages" commit and an "improve cryptic widescreen patch message" commit.

pcsx2/PluginManager.cpp
```diff
@@ -1126,20 +1126,25 @@ void SysCorePlugins::Load( PluginsEnum_t pid, const wxString& srcfile )
...
 
 void SysCorePlugins::Load( const wxString (&folders)[PluginId_Count] )
 {
 	if( !NeedsLoad() ) return;
 
 	wxDoNotLogInThisScope please;
-	
-	Console.WriteLn( Color_StrongBlue, "\nLoading plugins..." );
+
+	const PluginInfo* pi = tbl_PluginInfo;
```
turtleli: This line should be moved back to its original location.

Apart from that, the code seems fine to me.

PR #1876, opened 2017-03-27
gsdx: workaround AVX2 generated code by GCC

gregory38:
See commit for the full details.

I will likely open a GCC bug.

Edit: actually it is a regression of GCC4.9. GCC4.8 is fine (unoptimal code is generated)

### DIFF
--- a/plugins/GSdx/GSDrawScanline.cpp
+++ b/plugins/GSdx/GSDrawScanline.cpp
@@ -45,7 +45,19 @@ void GSDrawScanline::BeginDraw(const GSRasterizerData* data)
 
 	if(m_global.sel.mmin && m_global.sel.lcm)
 	{
+#if defined(__GNUC__) && _M_SSE >= 0x501
+		// GCC 4.9/5/6 doesn't generate correct AVX2 code for extract32<0>. GCC7 and upward are unknown.
+		// Intrinsic code is _mm_cvtsi128_si32(_mm256_castsi256_si128(m))
+		// It seems recent Clang got _mm256_cvtsi256_si32(m) instead. I don't know about GCC.
+		//
+		// Generated code keep the integer in an XMM register but bit [64:32] aren't cleared.
+		// So the srl16 shift will be huge and v will be 0.
+		//
+		int lod_x = m_global.lod.i.x0;
+		GSVector4i v = m_global.t.minmax.srl16(lod_x);
+#else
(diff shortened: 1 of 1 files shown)


plugins/GSdx/GSDrawScanline.cpp
```diff
@@ -45,7 +45,20 @@ void GSDrawScanline::BeginDraw(const GSRasterizerData* data)
 
 	if(m_global.sel.mmin && m_global.sel.lcm)
 	{
+#if defined(__GNUC__) && _M_SSE >= 0x500
```
turtleli: \_\_GNUC__ is also true for clang.

plugins/GSdx/GSDrawScanline.cpp
```diff
@@ -45,7 +45,20 @@ void GSDrawScanline::BeginDraw(const GSRasterizerData* data)
...
+		// GCC 4.9/5/6 doesn't generate correct AVX2 code for extract32<0>. GCC7 and upward are unknown.
+		// Intrinsic code is _mm_cvtsi128_si32(_mm256_castsi256_si128(m))
+		// It seems recent Clang got _mm256_cvtsi256_si32(m) instead. I don't know about GCC.
+		//
+		// Generated code keep the integer in an XMM register but bit [64:32] aren't cleared.
+		// So the srl16 shift will be huge and v will be 0.
+		//
+		// The volatile forbid GCC to generate bad optimization and invalid code.
+		volatile int lod_x = m_global.lod.i.extract32<0>();
+		GSVector4i v = m_global.t.minmax.srl16(lod_x);
```
turtleli: An alternative could be
```
GSVector4i whatever = m_global.lod.i.extract<0>().blend16<0xfc>(GSVector4i::zero());
GSVector4i v = m_global.t.minmax.srl16(whatever);
```
or something along those lines.

gregory38: Yes. I was going to search the zero extend function. Thanks.

```
__GNUC__ is also true for clang.
```
Code is barely called. And impact might be null with `u32to64`.

turtleli:
GCC 6.3 seems ok:
```
vmovdqa xmm3,XMMWORD PTR [ebx+0xb0]
vmovd  xmm4,DWORD PTR [ebx+0x160]
vpsrlw xmm0,xmm3,xmm4
```

PR #1877, opened 2017-03-29
GSdx: Putting the G and S back in CRC Hack Level

FlatOutPS2:
Ace Combat 4 CRC hack removes clouds for a good speed boost, which removes both 3D clouds(invisible with Hardware renderers, but cause slowdown) and 2D background clouds.
Removes blur from player airplane.
This hack also removes rockets, shows explosions(invisible without CRC hack) as garbage data, causes flickering issues with the HUD, and in some (night) missions removes the HUD altogether.
CRC hack has been moved to the aggressive level.

Aggressive is misspeled several times in the file, this has been
adressed.


gregory38:
Thanks

Issue #1878, opened 2017-03-29
Gran Turismo 3 - GT Racing Pack [PBPX-95503] Memory Card folder invalid filter


alondero:
***PCSX2 version:***
Development build v1.5.1961

***PCSX2 options:***
Memory Card (Folder), everything else default

***Plugins used:***
Default settings

***Description of the issue:***
With memory card folders selected, the GT3 save data is filtered out from this game as whilst the internal serial code of this game is PBPX-95503, the game still writes to ''BASCUS-97102GAMEDATA'' as per normal GT3 discs.
Suspect line in the log:
``McdSlot 0: [Folder] *******\Memory Cards
(FolderMcd) Indexing slot 0 with filter "PBPX-95503"``

***How to reproduce the issue:***
0. Ensure Memory Card Folders are in use in slot 1 and no memory card inserted in slot 2
0. Boot and Save Game
0. Exit PCSX2 and load again
0. Note that the auto-load screen didn't pop up nor any progress has been noted

***Last known version to work:***
Suspect this has always been the case since Memory Card folders were introduced.

***PC specifications:***
Windows 10, 4790k, GTX 980, 32GB Ram

alondero:
Not really sure how this can be solved *cleanly*, but thought I'd log it anyway. Not sure if there are any other special edition discs that this affects?

turtleli:
PR #1879 was merged, so this should be fixed. Closing.

PR #1879, opened 2017-03-30
GameDB: Add GT3 memcard filter

trostboot:
Adds base Gran Turismo 3 serial to other releases as per #1878

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -124,6 +124,7 @@ Region = PAL-Unk
 Serial = PBPX-95503
 Name   = Gran Turismo 3 - A-Spec [PS2 Bundle]
 Region = NTSC-U
+MemCardFilter = PBPX-95503/SCUS-97102
 ---------------------------------------------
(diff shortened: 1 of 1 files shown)


bin/GameIndex.dbf
```diff
@@ -124,6 +124,7 @@ Region = PAL-Unk
 Serial = PBPX-95503
 Name   = Gran Turismo 3 - A-Spec [PS2 Bundle]
 Region = NTSC-U
+MemCardFilter = PBPX-95503/SCUS-97102
```
turtleli: From reading the issue, I don't think PBPX-95503 needs to be added here.

trostboot: I left it in as a precaution, just in case it does at some point write something to a save corresponding to its own serial. Can take it again if you think it's unnecessary. Personally I'd err on the side of caution.

turtleli: I kinda doubt it's necessary, but I suppose it can't hurt to have the extra entry.

Issue #1830, opened 2017-02-16
GSdx missing GUI option 


gregory38:
Self remainder

* W : linear present
* WL: wrap mem
* WL: SW texture filtering

gregory38:
And #1758

FlatOutPS2:
Where is linear present located on the Linux GUI?

gregory38:
With post-processing shader. It was the only common section for all renderers. Besides it is really a bilinear filtering that is apply to the final image.

gregory38:
As a side note, I'm sure we can find a better name than `linear present`

FlatOutPS2:
> As a side note, I'm sure we can find a better name than linear present

What about Texture Filtering of Display? :P

gregory38:
It is up to you.

FlatOutPS2:
@gregory38 Did you add the wrap mem Linux GUI yet?

gregory38:
Maybe. It could be in my forest ;)

gregory38:
@FlatOutPS2 Actually it seems I didn't.

FlatOutPS2:
@gregory38 I have a PR ready that includes the Windows GUI for the memory wrapping and split texture filtering, I'm just waiting for the TF PR to merge so I can rebase my PR. Are you going to add the wrap mem Linux GUI soon, or can it wait till after my PR(my PR still needs a memory wrapping tooltip though)?

gregory38:
Wrap mem Linux can wait after your PR. I will address the issue of TF so we can go on.

gregory38:
12d2c6be569ed4316d2f32320c836523dedbc088

PR #1880, opened 2017-04-01
Greg/gsdx smaller gui

gregory38:
@FlatOutPS2 did I forget something ?

### DIFF
--- a/plugins/GSdx/GSLinuxDialog.cpp
+++ b/plugins/GSdx/GSLinuxDialog.cpp
@@ -28,6 +28,11 @@
 
 static GtkWidget* s_hack_frame;
 
+bool BigEnough()
+{
+	return (gdk_screen_get_height(gdk_screen_get_default()) > 1000);
+}
+
 void AddTooltip(GtkWidget* w, int idc)
 {
 	gtk_widget_set_tooltip_text(w, dialog_message(idc));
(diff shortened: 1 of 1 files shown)


FlatOutPS2:
Only that it closes https://github.com/PCSX2/pcsx2/issues/1830 ;)

gregory38:
Thank. I completely forgot it...

gregory38:
@turtleli is is better on your VM ? It should be small enough now.

turtleli:
GTK2 - it's small enough.
GTK3 - it's small enough, but for GTK 3.22 and later, it's better to add this line as well:
```
gtk_scrolled_window_set_propagate_natural_height(GTK_SCROLLED_WINDOW(scrollbar), true);
```
It'll avoid having a scrollbar when it isn't necessary, and the dialog resize behaviour is better (without it, the tab contents won't resize if the user resizes the dialog height)

Side note: there's quite a few deprecated warnings with the GTK3 compile.

i30817:
Thanks for this. Depending on the GTK3 ability to zoom out to show which widget is selected wasn't fun.

gregory38:
@turtleli 
Done. Thanks for the tip.

```
Side note: there's quite a few deprecated warnings with the GTK3 compile.
```
I guees the warnings are related to GTK4. It would be so much easier if all distributions used the same GTK version for wxWidget.

PR #1882, opened 2017-04-02
GSdx: OpenGL console messages

lightningterror:
Disable/remove show available extensions messages.
Only unavailable extensions will show messages.

### DIFF
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -306,8 +306,6 @@ namespace GLLoader {
 
 		if (!found) {
 			fprintf(stdout, "INFO: %s is NOT SUPPORTED\n", name.c_str());
-		} else {
-			fprintf(stdout, "INFO: %s is available\n", name.c_str());
 		}
 
 		std::string opt("override_");


FlatOutPS2:
I don't like hiding the available extensions. The correct solution is to only show the extension information on first boot.

lightningterror:
Hmmm , Is it really needed tho , what are the benefits ?
All it does is show available extensions.

Unavailable extensions serves a better purpose imo , shows which extensions aren't available.
And people will know what isn't supported.

MrCK1:
Curren behavior only shows both available and unavailable extensions that are required by PCSX3. 

Think of this as a diagnostic check, it's better to have more information printed then to be missing some when it comes time to solve user issues that could possibly be related to GSdx.

lightningterror:
Hmm true , I thought that with this it will make the console a bit more cleaner and since unavailable are shown as well then it will make a nice addition to remove available to make it more tidy.

Shall I close it then ?

FlatOutPS2:
The console will be much cleaner if the driver info is only printed once like on Linux. You could look into making it work the same way with Windows, and updating the PR instead of closing it.

turtleli:
> The console will be much cleaner if the driver info is only printed once like on Linux.

? The Windows and Linux behaviour is the same.

FlatOutPS2:
> ? The Windows and Linux behaviour is the same.

Switching renderers(like with F9) always reprints all the available driver extensions on Windows(which is more than 10 lines each time), while on Linux it just prints the current renderer when switching renderer.

turtleli:
It always reprints on Linux as well.
```
Suspending single plugin: GS
	Saving GS
	Closing GS
Delete 0 Shaders, 39 Programs, 28 Pipelines
Recovering single plugin: GS
	Opening GS
glX-Version 1.4 with Direct Rendering
OpenGL information. GPU: GeForce GTX 970/PCIe/SSE2. Vendor: NVIDIA Corporation. Driver: NVIDIA 378.13
INFO: GL_EXT_texture_filter_anisotropic is available
INFO: GL_ARB_gpu_shader5 is available
INFO: GL_ARB_viewport_array is available
INFO: GL_ARB_shader_image_load_store is available
INFO: GL_ARB_clear_texture is available
INFO: GL_ARB_direct_state_access is available
INFO: GL_ARB_texture_barrier is available
INFO: GL_ARB_get_texture_sub_image is available

Current Renderer: OpenGL (Software renderer)
Available VRAM/RAM:3840MB for textures
GSdx Lookup CRC:3E205556
	Loading GS
```

FlatOutPS2:
Then either something changed or different drivers/distro's have different behaviour.

gregory38:
@MrCK1 don't talk about the secret project 😛 

Honestly, the extension info isn't useful. I'm sure nobody miss the various D3D level capabilities.

This one could be removed on release build. I wanted to know if we generate lots of shaders (potentially multiple billions)
```
Delete 0 Shaders, 39 Programs, 28 Pipelines
```

Dokman:
@gregory38 are u kidding me? : O PCSX3?
: O : O

gregory38:
Are never serious ;)

FlatOutPS2:
Yes, definitely just a joke. Move along, nothing to see here...

lightningterror:
I actually read that as pcsx2 untill Dokman pointed it out xD

Dokman:
Sorry hahahaha i didn't get it hahahah

gregory38:
@FlatOutPS2 @MrCK1 
Do you guy really need to know the available extension to support users ?

MrCK1:
No, I don't have any issues modifying the behavior.

FlatOutPS2:
> Do you guy really need to know the available extension to support users ?

It can be informative and what's the harm if it's only shown once? The main issue with out without showing the available extensions is still that all the driver information is shown every time the renderer changes or the renderer settings are updated.

MrCK1:
Enabling memory wrapping causes the VRAM check to fail IIRC.

gregory38:
@lightningterror could you add a `static bool print_me_once` variable to avoid print when F9 is toggled.

lightningterror:
Where and what should I put to be precise , don't wanna mess something up.

gregory38:
I trust you 👍 
Hum what I will do is
1/ create a single function that call both check_gl_version and check_gl_extensions
2/ at start a static var print = true
3/ at the end var = false
4/ Then create an info function (I'm also fine with a macro) that will check the variable and print to stdout if true

As a bonus, I think we can only print missing function pointer in debug build. It would avoid tons of scary message for Intel users

gregory38:
@lightningterror don't hesitate to tell me if it is too complicated for you.

lightningterror:
For now it is xD
I'm still a beginner in coding. I mostly edit and tweak for now xD

gregory38:
I will do it.

gregory38:
@lightningterror thanks for you contribution. Next time you will provide the full patch :)

lightningterror:
> gsdx: only print once OpenGL status message All credits go to lightningterror

But I didn't do anything ... on this pr xD 
You did most of the stuff.

Issue #1716, opened 2016-12-23
OpenGL Support Intel IGPU'S (windows)


lightningterror:
Last known working version: v1.5.0-dev-561-g1208dbe
Version with issue: v1.5.0-dev-569-gb9e8384 till recent git

***PCSX2 options:***
Default options

***Plugins used:***
Default plugins

***Description of the issue:***
OpenGL Is not available on Intel Haswell IGPUs. The issue could be present on both Sandy and Ivy Bridge. Broadwell and Skylake have not been tested.
I'm not sure if the issue is present on Windows 10. So far I've tested windows 7.

When a game is loaded the emulator shows the plugin configuration error and promotes the user to change some settings.

The following error is also displayed in the console:
(pxActionEvent) GS plugin failed to open!(thread:MTGS)(thread:EE Core)

Other errors are shown in the console as well as to what features and plugins are supported by the igpu and driver but I think they are not relevant.

***How to reproduce the issue:***
If you have an ivy , sandy , haswell chip run the emulator with opengl render using the igpu. (Also make sure you disable your dedicated gpu if you have one when running the emulator in opengl , if you don't it will try to use the dedicated gpu instead of the integrated even tho the integrated is selected)

***Last known version to work:***
v1.5.0-dev-561-g1208dbe

***PC specifications:***
i3 4160  , Intel HD 4400 , 16gb ram , Windows 7 64 bit

MrCK1:
It's actually not a regression, it's a problem on Intel's end with their OpenGL driver support.

gregory38:
It would me much better to ask public review site to highlight the fact that Intel supports iGPU for 1-2 years (likely closer of 1). Note sandy bridge is working fine on Linux.... It misses real competition here.

lightningterror:
Hmm my guess is haswell works on linux as well. 
Heck for windows the last update we got was around April maybe.

gregory38:
Yes, I mean from Sandy Bridge up to latest Intel iGPU.

mirh:
Is this the same of #345?

Also: it [seems](http://leshcatlabs.net/forums/showthread.php?tid=1256#pid17075) there are [kind of] newer drivers. 
I dunno then if they are w10 only AND if they even differs from latest official ones in the first place.

lightningterror:
There are beta drivers that are only for windows 10 except for skylake which supports 7 and 8.

[...]

Issue #345, opened 2014-11-11
Gsdx-ogl not compatible with Intel MS driver (glsl compilation error)


pal1000:
-PCSX2 version used: 1.3.0-20141110154305
-Plugins used: GSdx, SPU2-X, Lilypad, Gigaherz CDVD,rest null plugins
-PCSX2 options: no settings modifications affect the issue except for switching GSDX to an OpenGL mode (hardware or software)
-Plugin settings: Tested GSdx DX9/11 hardware and software renderers. The issue doesn't happen with  DX9/11 hardware or software renderer of GSdx.
-Description: The bios or games graphics appear either black or white and they alternate depending what the game / bios does. Sound is playing normally.
-How to replicate: Boot any game or just the bios
-Last known version to work: Unknown, probably never worked
PC specifications: CPU: Intel Core i3-3220 @ 3.3Ghz GPU: Intel HD Graphics 2500, RAM: 4 GB OS: Windows 7 SP1 x64
Other comments: 
Issue in action: https://www.youtube.com/watch?v=VN3v_g67juc
This issue affects all driver releases for Intel HD Graphics 2500 that I ever used since got this PC which were in chronological order:
r2867 - it was crashing PCSX2 on game / bios boot when using OpenGL hardware or software;
r3071 -  it was crashing PCSX2 on game / bios boot when using OpenGL hardware or software;
r3347 - it was causing the same white and black alternation that affects GSDX OpenGL hardware and software modes now;
r3621 - it was causing the same white and black alternation that affects GSDX OpenGL hardware and software modes now;
r3945  - it was causing the same white and black alternation that affects GSDX OpenGL hardware and software modes now;
r3958 - this is the current driver.
Driver updates for Intel HD Graphics 2500 are published here: https://downloadcenter.intel.com/SearchResult.aspx?lang=eng&ProdId=3711
I also reported a similar problem with PPSSPP when usage of dual source alpha blending was added to it. See: https://github.com/hrydgard/ppsspp/issues/4848#issuecomment-30673657 .
I also opened a thread on Intel community forums regarding dual source alpha blending: https://communities.intel.com/thread/55691

gregory38:
I added a hack to compile shader on Intel Window driver. Could you test it ?

maximu:
OGL HW working load now under windows.

[...]

PR #1885, opened 2017-04-05
GSdx ogl sw: Move GL_ARB_copy_image to optional

lightningterror:
Follow up to https://github.com/PCSX2/pcsx2/commit/5b91dc45e0d51647ee45e95161b7e92dcbd53932
Move GL_ARB_copy_image to optional for OpenGL SW render.
It will allow Ivy Bridge to work with OpenGL SW as it's not required.

Sandy Bridge is not yet tested , would be nice if someone could test.

Improves #1716

### DIFF
--- a/plugins/GSdx/GLLoader.h
+++ b/plugins/GSdx/GLLoader.h
@@ -370,6 +370,7 @@ namespace GLLoader {
 
 	// GL
 	extern bool found_geometry_shader;
+	extern bool found_GL_ARB_copy_image;
 	extern bool found_GL_ARB_clip_control;
 	extern bool found_GL_ARB_gpu_shader5;
 	extern bool found_GL_ARB_shader_image_load_store;
--- a/plugins/GSdx/GSDeviceOGL.cpp
+++ b/plugins/GSdx/GSDeviceOGL.cpp
@@ -1243,6 +1243,7 @@ void GSDeviceOGL::CopyRect(GSTexture* sTex, GSTexture* dTex, const GSVector4i& r
 	PSSetShaderResource(6, sTex);
 #endif
 
+	ASSERT(GLLoader::found_GL_ARB_copy_image && glCopyImageSubData);
 	glCopyImageSubData( sid, GL_TEXTURE_2D,
 			0, r.x, r.y, 0,
 			did, GL_TEXTURE_2D,
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -424,7 +424,7 @@ namespace GLLoader {
 		status &= status_and_override(found_GL_ARB_shading_language_420pack, "GL_ARB_shading_language_420pack", true);
 		status &= status_and_override(found_GL_ARB_texture_storage, "GL_ARB_texture_storage", true);
 		// GL4.3
-		status &= status_and_override(found_GL_ARB_copy_image, "GL_ARB_copy_image", true);
+		status &= status_and_override(found_GL_ARB_copy_image, "GL_ARB_copy_image", required_for_hw);
 		status &= status_and_override(found_GL_KHR_debug, "GL_KHR_debug", true);
 		// GL4.4
 		status &= status_and_override(found_GL_ARB_buffer_storage,"GL_ARB_buffer_storage", true);
--- a/plugins/GSdx/GSWnd.cpp
+++ b/plugins/GSdx/GSWnd.cpp
@@ -121,7 +121,7 @@ void GSWndGL::PopulateGlFunction()
 	GL_EXT_LOAD_OPT(glBindImageTexture);
 	GL_EXT_LOAD_OPT(glMemoryBarrier);
 	// GL4.3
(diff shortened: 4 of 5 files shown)


plugins/GSdx/GSDeviceOGL.cpp
```diff
@@ -1242,12 +1242,12 @@ void GSDeviceOGL::CopyRect(GSTexture* sTex, GSTexture* dTex, const GSVector4i& r
 #ifdef ENABLE_OGL_DEBUG
 	PSSetShaderResource(6, sTex);
 #endif
-
-	glCopyImageSubData( sid, GL_TEXTURE_2D,
-			0, r.x, r.y, 0,
-			did, GL_TEXTURE_2D,
-			0, 0, 0, 0,
-			r.width(), r.height(), 1);
+	if (GLLoader::found_GL_ARB_copy_image && glCopyImageSubData)
```
gregory38: The function is never executed on the SW renderer. I think an assert will be be better at the top of the function instead.
```
ASSERT(GLLoader::found_GL_ARB_copy_image && glCopyImageSubData);
```

lightningterror: Hmm , should the semicolon(;) be added at the end?

gregory38: you mean after the assert ? Likely but maybe not. Why ?

lightningterror: yea because like it is now it doesn't have , just making sure If I should add or not when changing.

lightningterror:
Ready for merge I guess :)

plugins/GSdx/GLLoader.h
```diff
@@ -370,6 +370,7 @@ namespace GLLoader {
 
 	// GL
 	extern bool found_geometry_shader;
+	extern bool found_GL_ARB_copy_image;
```
turtleli: Trailing whitespace.

lightningterror: Ups didn't see that. Fixed.

plugins/GSdx/GSDeviceOGL.cpp
```diff
@@ -1242,7 +1242,7 @@ void GSDeviceOGL::CopyRect(GSTexture* sTex, GSTexture* dTex, const GSVector4i& r
 #ifdef ENABLE_OGL_DEBUG
 	PSSetShaderResource(6, sTex);
 #endif
-
```
FlatOutPS2: It would look better with the empty line in between.

lightningterror: Added.

pal1000:
Here is an experiment with SandyBridge:
https://github.com/PCSX2/pcsx2/compare/master...pal1000:work
Requirements for SandyBridge to work:
- Windows 10 manifest removal,  done, easy fix, it wasn't working anyway for some reason, only caused OpenGL to refuse to load;
- lower GL context to 3.1. On windows GL 3.3 context is not supported, done in an aggressive way, I assume it should try 3.3 and fallback to 3.1 if 3.3 is not supported.  Tested with Mesa the impact of the aggressive context downgrade, it broke context flags, but not fatally, could get it working by setting MESA_GL_VERSION_OVERRIDE=3.1COMPAT
Only 1 extension is missing per [this log](https://github.com/PCSX2/pcsx2/files/903800/log.txt) - glGetDebugMessageLogARB.

lightningterror:
I knew it ... Sandy is quite behind. Well for the windows 10 manifest maybe a quick check can be written if Sandy is detected not to load Windows 10 manifest

Btw can you check these drivers out , I think ms tweaked some things so maybe w10 manifest works
http://www.catalog.update.microsoft.com/Search.aspx?q=9.17.10.4459

Lowering the gl requirement to 3.1 can be done as well

As for glGetDebugMessageLogARB well I dunno what @gregory38 will say about it since it's mandatory.

gregory38:
I'm in favor to drop SB. D3D will never be removed anyway. So you don't gain anything. SB is dx10 only.   Now that dx12 is out, we can expect dx11 will become a minimal requirement which will trigger the upgrade process of those old iGPU.

pal1000:
> Btw can you check these drivers out , I think ms tweaked some things so maybe w10 manifest works
> http://www.catalog.update.microsoft.com/Search.aspx?q=9.17.10.4459
I already have this driver, and since early March is pushed by default by Windows Update.
Anyway, this is not a big deal, This system was dated when it was released in early 2013, now is mostly obsolete.

gregory38:
Yes it's time to upgrade. Dx11 is 8 years old already. I'm pretty sure there are plenty games that requires dx11. SB is 6 years old.

lightningterror:
Hmm I didn't know SB is dx10 only , Intel could have done so much with their igpus , they just wasted a lot of their hardware potential ..rip

Anyway shall we close this then? https://github.com/PCSX2/pcsx2/issues/345

gregory38:
Hence only 3.1 ;) To be fair, iGPUs aren't easy to do.

PR #1886, opened 2017-04-06
Don't check for patches when NoDisc is the CDVD_SourceType.

MrCK1:
Skip checking for cheats, WS hacks, ect. and save a few lines from being printed in the console; as we already know it will return with nothing if NoDisc is the source. I also noticed a very small perf increase when booting with NoDisc.

### DIFF
--- a/pcsx2/gui/AppCoreThread.cpp
+++ b/pcsx2/gui/AppCoreThread.cpp
@@ -388,7 +388,8 @@ static void _ApplySettings( const Pcsx2Config& src, Pcsx2Config& fixup )
 
 	ForgetLoadedPatches();
 
-	if (!curGameKey.IsEmpty())
+	bool loadFromDisc = g_Conf->CdvdSource != CDVD_SourceType::NoDisc;
+	if (!curGameKey.IsEmpty() && loadFromDisc)
 	{
 		if (IGameDatabase* GameDB = AppHost_GetGameDatabase() )
 		{
@@ -430,18 +431,20 @@ static void _ApplySettings( const Pcsx2Config& src, Pcsx2Config& fixup )
 	}
 
 	//Till the end of this function, entry CRC will be 00000000
-	if (!gameCRC.Length())
+	if (gameCRC.IsEmpty() && loadFromDisc)
 	{
 		Console.WriteLn(Color_Gray, "Patches: No CRC found, using 00000000 instead.");
 		gameCRC = L"00000000";
 	}
 
 	// regular cheat patches
-	if (fixup.EnableCheats)
+	if (fixup.EnableCheats && loadFromDisc)
+	{
 		gameCheats.Printf(L" [%d Cheats]", LoadPatchesFromDir(gameCRC, GetCheatsFolder(), L"Cheats"));
+	}
 
 	// wide screen patches
-	if (fixup.EnableWideScreenPatches)
+	if (fixup.EnableWideScreenPatches && loadFromDisc)
 	{
 		if (int numberLoadedWideScreenPatches = LoadPatchesFromDir(gameCRC, GetCheatsWsFolder(), L"Widescreen hacks"))
 		{


pcsx2/gui/AppCoreThread.cpp
```diff
@@ -427,23 +427,29 @@ static void _ApplySettings( const Pcsx2Config& src, Pcsx2Config& fixup )
...
+	}
 
 	// regular cheat patches
-	if (fixup.EnableCheats)
+	CDVD_SourceType cdvdsrc(g_Conf->CdvdSource);
+	if ((fixup.EnableCheats) && (cdvdsrc != CDVD_SourceType::NoDisc))
+	{
 		gameCheats.Printf(L" [%d Cheats]", LoadPatchesFromDir(gameCRC, GetCheatsFolder(), L"Cheats"));
+	}
+		else if ((fixup.EnableCheats) && (cdvdsrc == CDVD_SourceType::NoDisc))
```
woj1993: Is it needed?

FlatOutPS2:
There are widescreen patches for a few BIOS versions. AFAIK they're not in the archive, so I'm not sure if anyone cares about it or being able to cheat in the BIOS.

lightningterror:
Well it's a nice addition , anything that can help improve pcsx2 is welcome I guess.

pcsx2/gui/AppCoreThread.cpp
MrCK1: IIRC, it won't compile without it. I'll recheck later.

ssakash:
@MrCK1 - Code still needs to be improved, it has obvious mistakes everywhere. Too lazy to point them out right now on my mobile, assigning someone else who recently edited this file for the review. (Nice feature Github)

By the way, personally I don't think it's worth to do this, it's just restrictive to others who might potentially use cheats/patches for some stuff on the BIOS, but if it shows a benefit in performance/boot up time as you had said earlier, then maybe it's worth it.

MrCK1:
I figured there was something I had to be doing wrong :P. Maybe perf with the patching system can still be improved further in another PR later.

pcsx2/gui/AppCoreThread.cpp
FlatOutPS2: It'll either do nothing or block widescreen patches when cheats are enabled. Plus you need to look at the spacing of several lines, there's a tab too much at the beginning of this line for example(though the whole line should be removed).

pcsx2/gui/AppCoreThread.cpp
```diff
@@ -427,23 +427,29 @@ static void _ApplySettings( const Pcsx2Config& src, Pcsx2Config& fixup )
...
 
 	// wide screen patches
-	if (fixup.EnableWideScreenPatches)
-	{
-		if (int numberLoadedWideScreenPatches = LoadPatchesFromDir(gameCRC, GetCheatsWsFolder(), L"Widescreen hacks"))
+	if ((fixup.EnableWideScreenPatches) && (cdvdsrc == CDVD_SourceType::NoDisc));
+		else if (int numberLoadedWideScreenPatches = LoadPatchesFromDir(gameCRC, GetCheatsWsFolder(), L"Widescreen hacks"))
```
FlatOutPS2: Either make it one if:
```
if ((fixup.EnableWideScreenPatches) && !(cdvdsrc == CDVD_SourceType::NoDisc &&
    int numberLoadedWideScreenPatches = LoadPatchesFromDir(gameCRC, GetCheatsWsFolder(), L"Widescreen hacks"))
```
or keep the old format and just add the nodisc check to the first if. Adding an empty if check with an else is just confusing. ;)

MrCK1: Yeah I knew there was a better way than doing an empty check, could t think of it at the time.

pcsx2/gui/AppCoreThread.cpp
```diff
@@ -427,23 +427,29 @@ static void _ApplySettings( const Pcsx2Config& src, Pcsx2Config& fixup )
 		// the BIos isn't loaded until after initial calls to ApplySettings.
 
 		gameName = L"Booting PS2 BIOS... ";
-	}
 
 	//Till the end of this function, entry CRC will be 00000000
-	if (!gameCRC.Length())
+	CDVD_SourceType cdvdsrc(g_Conf->CdvdSource);
```
FlatOutPS2: You could convert this line to a bool for checking `cdvdsrc != CDVD_SourceType::NoDisc`.

turtleli:
So... is it worth doing this? I don't think a user will really care about a few extra messages when they're booting the BIOS.

The only reason why there might be a small performance increase while booting with this PR is because the widescreen patch archive doesn't need to be searched if widescreen patches are enabled.

MrCK1:
@FlatOutPS2 @turtleli I'm not at my main PC, so I can't test any games. Can either of you make sure patches still load correctly? Thanks.

pcsx2/gui/AppCoreThread.cpp
```diff
@@ -427,21 +427,23 @@ static void _ApplySettings( const Pcsx2Config& src, Pcsx2Config& fixup )
...
-	if (fixup.EnableCheats)
+	if ((fixup.EnableCheats) && (cdvdsrc != CDVD_SourceType::NoDisc))
+	{
 		gameCheats.Printf(L" [%d Cheats]", LoadPatchesFromDir(gameCRC, GetCheatsFolder(), L"Cheats"));
+	}
 
-	// wide screen patches
-	if (fixup.EnableWideScreenPatches)
+		// wide screen patches
+	if ((fixup.EnableWideScreenPatches) && (cdvdsrc != CDVD_SourceType::NoDisc))
```
FlatOutPS2: You can just use one set of brackets: `if (fixup.EnableWideScreenPatches && cdvdsrc != CDVD_SourceType::NoDisc)`
The same goes for the other lines.

And there's an unnecessary tab before the comment above this line.

turtleli:
> I'm not at my main PC, so I can't test any games. Can either of you make sure patches still load correctly? Thanks.

I haven't tested it, but it's definitely broken. The code is also incorrectly indented - if it was properly indented you'd notice where the major bug is quite easily.

pcsx2/gui/AppCoreThread.cpp
```diff
@@ -457,6 +459,7 @@ static void _ApplySettings( const Pcsx2Config& src, Pcsx2Config& fixup )
 			gameWsHacks.Printf(L" [%d widescreen hacks]", numberDbfCheatsLoaded);
 		}
 	}
+}
```
FlatOutPS2: This needs to be at line 430 where you removed it.

MrCK1:
I implemented the suggested changes. A quick test shows that WS patches load properly again.

pcsx2/gui/AppCoreThread.cpp
```diff
@@ -430,18 +430,21 @@ static void _ApplySettings( const Pcsx2Config& src, Pcsx2Config& fixup )
 	}
 
 	//Till the end of this function, entry CRC will be 00000000
-	if (!gameCRC.Length())
+	CDVD_SourceType cdvdsrc(g_Conf->CdvdSource);
```
willkuer: > FlatOutPS2 20 hours ago 
> You could convert this line to a bool for checking `cdvdsrc != CDVD_SourceType::NoDisc`.

`bool loadFromDisc = g_Conf->CdvdSource != CDVD_SourceType::NoDisc`

L434:
`if (!gameCRC.Length() && loadFromDisc )`

MrCK1: Done, thanks 👍

pcsx2/gui/AppCoreThread.cpp
```diff
@@ -430,21 +430,21 @@ static void _ApplySettings( const Pcsx2Config& src, Pcsx2Config& fixup )
 	}
 
 	//Till the end of this function, entry CRC will be 00000000
-	CDVD_SourceType cdvdsrc(g_Conf->CdvdSource);
-	if (!gameCRC.Length() && cdvdsrc != CDVD_SourceType::NoDisc)
+	bool loadFromDisc = g_Conf->CdvdSource != CDVD_SourceType::NoDisc;
+		if (!gameCRC.Length() && loadFromDisc)
```
FlatOutPS2: The indentation tab strikes again. ;)

MrCK1:
Ah crap, I actually thought about that for a sec and figured it should be indented.

pcsx2/gui/AppCoreThread.cpp
```diff
@@ -430,18 +430,21 @@ static void _ApplySettings( const Pcsx2Config& src, Pcsx2Config& fixup )
 	}
 
 	//Till the end of this function, entry CRC will be 00000000
-	if (!gameCRC.Length())
+	bool loadFromDisc = g_Conf->CdvdSource != CDVD_SourceType::NoDisc;
+	if (!gameCRC.Length() && loadFromDisc)
```
turtleli: `gameCRC.empty()` would be more readable IMO.

MrCK1: I suppose we could reuse ```gameCRC.IsEmpty()``` on line 423. That would function the same in this case correct?

turtleli: I don't understand what you mean. Explain?

MrCK1: Never mind, I actually figured it out.

MrCK1:
@turtleli Is this ready to be merged? I added the suggested changes and suppressed another console message.

pcsx2/CDVD/IsoFS/IsoFS.cpp
```diff
@@ -49,8 +50,9 @@ IsoDirectory::IsoDirectory(SectorSource& r)
 	uint i = 16;
 
 	m_fstype = FStype_ISO9660;
-
-	while( !done )
+	
+	bool loadFromDisc = g_Conf->CdvdSource != CDVD_SourceType::NoDisc;
+	while( !done && loadFromDisc)
```
turtleli: This isn't the right solution for suppressing the "invalid partition descriptor" message. It doesn't make sense to me.

turtleli:
@avih - Any opinion on this PR? I figure you care a lot more than I do about widescreen patches ;)

pcsx2/CDVD/IsoFS/IsoFS.cpp
MrCK1: Huh didn't realize I left the else in there. I'm pretty sure I had this working at some point, but I'll have to look into it or remove this completely.

turtleli: Not sure what you mean by else. The code might suppress the message, but it doesn't make sense for the code to be in the ISO file system code at all.

MrCK1: The error only appears on the first boot when NoDisc is the source. If you boot with another source and switch to NoDisc later, you won't see the message. I guess some other data related to disk sectors is getting left behind. Hope that clears it up.

turtleli: I'm not talking about when the error message appears.

I'm saying this fix is in the wrong place. There is likely a better place to add code that prevents the error message from showing up.

MrCK1: Ah ok. I'll remove this portion then.

gregory38:
@avih Could you look at this PR ? Thanks you.

avih:
> @avih - Any opinion on this PR? I figure you care a lot more than I do about widescreen patches ;)

Sorry, must have missed that.

I do, I don't see how that improves the code/experience. If users don't want the patches messages or the patches, then just disable them at the menu IMO. We already have an option for that.

> I haven't tested it, but it's definitely broken. The code is also incorrectly indented

I'm guessing this relates to the PR code rather than to PCSX2's, and it's been fixed since?

turtleli:
> I'm guessing this relates to the PR code rather than to PCSX2's, and it's been fixed since?

Yes.

gregory38:
Any status on this PR ?

turtleli:
I'm guessing this should probably be closed without merging? The overall opinion of this PR doesn't seem positive.

MrCK1:
That's fine with me, didn't really seem like it was going anywhere. I have follow-ups planned for my other stuff, but right now it's difficult for me to find time between work and school.

PR #1253, opened 2016-03-20
spu2-x:windows:Use XAudio2.8+ for Windows 8 and later

turtleli:
Changes:
 * Only use the DirectX SDK for XAudio 2.7 support - everything else will use the Windows SDK.
 * Use XAudio 2.8+ for Windows 8 and later, use XAudio 2.7 for Windows Vista and 7

Combined with my GSdx (#1238) and XInput (#1237) PRs, Windows 8.1 and 10 users should be able to run PCSX2 without requiring the DirectX redists.

PR #1238, opened 2016-03-18
gsdx: Remove DirectX SDK dependency

turtleli:
Changes:
 * Removes DirectX SDK dependency from GSdx by using D3DCompile to compile the shaders instead of the D3DX equivalents
 * Bundles d3dcompiler_47.dll with PCSX2 - this is only of use to Windows Vista and 7 (EDIT: and 8) users - Windows 8.1+ users can delete the file to use the version that comes with their OS.

PR #1237, opened 2016-03-18
lilypad, xpad: Remove DirectX SDK dependency, support both XInput 1.3 and 1.4

turtleli:
Changes:
 * Removes DirectX SDK dependency from Lilypad and xpad
 * Allow Windows 8 and above to use XInput 1.4 if the DirectX redists aren't installed. If the XInput 1.3 dll is present, it'll choose to use that first - I think SCP only has an XInput 1.3 wrapper at the moment.

Issue #529, opened 2015-05-06
Support vsync in opengl windows


gregory38:
It must be easy to do.

http://stackoverflow.com/questions/589064/how-to-enable-vertical-sync-in-opengl

https://www.opengl.org/wiki/Swap_Interval

```
    wglSwapIntervalEXT = GetProcAddress("wglSwapIntervalEXT");
```

ssakash:
wglSwapIntervalExt function is Windows only.

gregory38:
??? Did you read tag and title?

ssakash:
sorry, I overlooked both of those things. :p 

by the way, should we swap interval with glFinish (or) without glFinish?

gregory38:
For future

```
[22:30:33] <Armada> MrKB: if you specify the WS_POPUPWINDOW style you can bypass the window manager in OpenGL basically for free
[22:31:19] <Armada> relies on driver heuristics, so it's undocumented, but it's pretty consistent
[22:31:30] <Armada> it's what I use for exclusive mode at Dolphin
```

karasuhebi:
Vsync isn't working when using OpenGL on Windows. I see tearing. Can anyone confirm?

avih:
> Vsync isn't working when using OpenGL on Windows. I see tearing. Can anyone confirm?

I didn't try, but there are few factors in play here:

TL;DR, OpenGL in Windows in general:
- in a window, no tearing should happen but vsync timing is not likely to work.
- In full screen, tearing may or may not happen, and vsync timing might or might not work (respectively, if it can tear then vsync timing CAN work) - depending on the driver and the the program (PCSX2 here).
- With PCSX2, even in full screen, it's unlikely to have tearing, and vsync timing is not likely to not work.

When displaying content in a window (i.e. not full screen), Windows should not be able to display tearing even if you try really hard. Or at least it shouldn't when the compositor ("Aero") is being used, which is always on Windows 8 onward, and when using the default theme with windows Vista and 7.

Because the desktop and all windows are vsync'ed implicitly via the compositor, there's not really a monitor vsync signal to sync to, but rather a compositing software which can pick the output buffer from the OpenGL driver at potentially irregular times.

Whether or not the OpenGL driver can sync to this potentially irregular timing for a window when used with `wglSwapIntervalEXT(1)` ("vsync") is up to driver implementation.

[...]

PR #1887, opened 2017-04-06
pcsx2:windows: Remove DWM code

turtleli:
I don't think DwmEnableMMCSS actually makes a difference.

DwmSetPresentParameters is unused and is also not implemented from Windows 8.1 onwards.


sudonim1:
Uh okay, is that just a hunch or have you tested the overall latency somehow? And do you mean doesn't make a difference on windows 10? Windows 8? Even if it only helps on Windows 7, isn't that worth keeping?

turtleli:
It's a guess, hence the PR. No, I haven't tested the latency.

There's barely any info I can find on the DwmEnableMMCSS function and I think it has to be used with the  AvSetMmThreadCharacteristics function for it to do anything (https://msdn.microsoft.com/en-us/library/windows/desktop/ms684247(v=vs.85).aspx).

mirh:
Some kind of documentation can be found [here](http://www.oszone.net/4364/Windows_Vista). 
And according to people [here](http://sites.music.columbia.edu/pipermail/portaudio/2013-September/015317.html) and [here](https://github.com/filoe/cscore/issues/73) it does make a difference. 

The only question if any, is whether the use-cases where this would be observed applies to pcsx2 (for example, it's no mp3 player expected to still run flawlessly minimized). 

> DwmSetPresentParameters is unused and is also not implemented from Windows 8.1 onwards.

Reading what it does, my educate guess is that we could compare this to W10 game mode.

turtleli:
> And according to people here and here it does make a difference.

First one says "might be worth investigating". The reply says "makes no difference".
Second one says increasing thread priority makes a difference. It doesn't say whether DWMEnableMMCSS makes a difference.

I already saw both - my conclusion was that no conclusion can be reached. There's simply too much uncertainty in any information that can be found. The commit that introduced the call also says "I will leave it to the world to figure out if it really makes a difference".

> Reading what it does, my educate guess is that we could compare this to W10 game mode.

The code doesn't ever call the function that calls DwmSetPresentParameters. And https://www.opengl.org/discussion_boards/showthread.php/164119-Making-composite-managers-go-along-with-OpenGL/page2 says it makes things much worse on OpenGL while only improving D3D slightly. I couldn't find any additional info which agrees/disagrees with the finding.

mirh:
Ok, for the former you are right, but the latter issue was actually closed with the mmcss commit. 

> The code doesn't ever call the function that calls DwmSetPresentParameters.

True. @gigaherz you shouldn't commit on late nights 😛 
I found [this](https://www.forum.rme-audio.de/viewtopic.php?id=17575#p100562) though for plain MMCSS scheduling though. 
Actual [benchmarking](http://www.pcworld.com/article/3187171/windows/tested-windows-10s-game-mode-makes-unplayable-games-playable-sometimes.html) with respect to pcsx2 shouldn't be impossible then 

> I couldn't find any additional info which agrees/disagrees with the finding.

Imo, poster was missing something, since 1) fQueue false should be basically equal to normal 2) fUseSourceRate true or false is quite of an opposite thing
Anyway, I wonder if this couldn't play some role in #1437 ?

---
I lost quite a lot of time following [this code](https://github.com/libretro/fbalpha2012/blob/a8a918e8b65a4a18544504b47a3ca4f361441769/src-new/burner/win32/dwmapi_core.cpp) then, which ultimately lead me on [these](http://neosource.1emulation.com/forums/index.php?topic=1482.0) [threads](http://neosource.1emulation.com/forums/index.php?topic=1260.45). Hope you may figure something out of them.

turtleli:
> Ok, for the former you are right, but the latter issue was actually closed with the mmcss commit.

I read this part:
```
Yes. Increasing the thread priority allows using lower latencies. I'm able to use a latency
of 200ms now, without playback interruptions when browsing through the UI.
Conclusion: it is helpful.
```
IOW, issue was fixed before the MMCSS commit even happened. So there's no useful data there.

> I found this though for plain MMCSS scheduling though.
> Actual benchmarking with respect to pcsx2 shouldn't be impossible then

Well, that shows some actual data. But since it talks about tasks, it also somewhat hints to what I wrote earlier:
```
I think it has to be used with the AvSetMmThreadCharacteristics function for it to do anything
(https://msdn.microsoft.com/en-us/library/windows/desktop/ms684247(v=vs.85).aspx).
```
Which we don't call. And note that function has to be called by every thread that needs it.

> Imo, poster was missing something, since 1) fQueue false should be basically equal to normal 2) fUseSourceRate true or false is quite of an opposite thing

I dunno what you're reading. I read:
No messing about with DwmSetPresentParameters - OpenGL is faster than D3D for test application.
Messing about with DwmSetPresentParameters - OpenGL crawls, D3D slightly faster for test application.

> Anyway, I wonder if this couldn't play some role in #1437 ?

User is using Windows 10 + function is not implemented on Windows 8.1 onwards = No.

> I lost quite a lot of time following this code then, which ultimately lead me on these threads. Hope you may figure something out of them.

Doesn't seem useful.

mirh:
>  So there's no useful data there.

I feel almost like criminally dumb to skim over those parts? Sorry. 

> Which we don't call. And note that function has to be called by every thread that needs it.

Well, my point was that we could then. 
Of course it would hardly make too much of a difference. But similarly to leaving or removing the code, I guess the deal is (in **every** case) pretty slim. 

Even though, why not, you could take advantage of this *impasse* to add support for w10 game mode now (or better, whenever the hell microsoft decides to release docs). 

>  Messing about with DwmSetPresentParameters - OpenGL crawls, D3D slightly faster for test application.

And I'm inferring that when he says it, he must be generally referring to just some specific setting, not "every possible combination". 

> User is using Windows 10 + function is not implemented on Windows 8.1 onwards = No.

The alleged **fix** function not existing doesn't entail the problem isn't still that one in the first place. 
Moreover, this could just be due to a rewrite of DWM that put in charge of the same responsibility [other](https://msdn.microsoft.com/en-us/library/windows/apps/dn448914.aspx) new [functions](https://msdn.microsoft.com/en-us/magazine/dn745861).

turtleli:
> add support for w10 game mode now (or better, whenever the hell microsoft decides to release docs).

...What. In nice words - I think you've misunderstood what game mode is.

> Well, my point was that we could then.
> Of course it would hardly make too much of a difference. But similarly to leaving or removing the code, I guess the deal is (in every case) pretty slim.

I'd rather not introduce "performance" code that "might" make a difference without evidence that it actually makes a difference. And apparently AvSetMmThreadCharacteristics is likely independent of DWMEnableMMCSS, seeing that PortAudio calls the former but not the latter. Which leads us back to the question of what DWMEnableMMCSS even does since there's now no useful information for it. Which is why I opened this PR in the first place.

> And I'm inferring that when he says it, he must be generally referring to just some specific setting, not "every possible combination".

I don't really care. There's insufficient data.

> The alleged fix function not existing doesn't entail the problem isn't still that one in the first place.
> Moreover, this could just be due to a rewrite of DWM that put in charge of the same responsibility other  new functions.

Answer is I don't know. I interpreted your question as "Couldn't DwmSetPresentParameter have an effect" (seriously, I can barely understand you most of the time, and I don't believe that's my fault either - remember #1237 and #1238, and how the 8.1 SDK "doesn't" work on Windows 7?).

gigaherz:
Uhm... The MMCSS (Multimedia Class Scheduler Service), allows processes that require high-precision timing to have shorter processor quanta, allowing threaded operations to switch faster, and updating the tick counters more often, which results in more accurate timers. The MMCSS is primarily designed for multimedia tasks, such as video and audio playback, where having a smooth framerate and audio-video sync is important.

It does however come at a cost: there's more context-switching overhead due to the faster scheduling frequency.

So far as I know, Windows enables the MMCSS if there's any process that requests MMCSS.

That said, I don't even remember writing that code, so if it has never been used then... *shrug* ;P

turtleli:
DwmEnableMMCSS is used. I don't know whether it actually makes a difference though (it doesn't seem to on mine).

The comment in the code that calls DwmEnableMMCSS says the timer resolution is set to 10ms, but I also notice we call timeBeginPeriod to set a minimum timer resolution of 1ms. Can't say I know what to make of that though.

mirh:
> ...What. In nice words - I think you've misunderstood what game mode is.

You prioritize the game compared to other processes?
I mean.. I guess in this case you don't pass through DWM, but it's not like I can create a new feature request by myself every time I have a showerthought. (:
I think I already put too many irons in the fire for this month. 

> I'd rather not introduce "performance" code that "might" make a difference without evidence that it actually makes a difference

Granted. I see especially how developers should always ask thrice before.. guess what? Their application just so coincidentally happened to be The most important one on Earth. 
This was why I asked for how much we are sensitive and caring to latency (we are no DAW, true, but we aren't Word either). 

> I don't really care. There's insufficient data.

Indeed, the vagueness applies in both situations though. 

>  I interpreted your question as "Couldn't DwmSetPresentParameter have an effect" 

You interpreted my question as W10 related, ok, I can see I should have specified better. 

> remember 

Yes, I always remember when I made a total fool of myself (aka you can use 8.1 SDK to target 7)
By the time of #1253 iirc I had finally even grasped that whatever microsoft's headers say about loading this dll in this situation, we can always not care! Did I already say sorry?

Anyway, you can even remove the function if you want, it's just it seemed quite [a lot](https://blogs.msdn.microsoft.com/larryosterman/2007/08/28/windows-vista-sound-causes-network-throughput-slowdowns/) like we were fitting all *the theory* for MMCSS to be justly deserved and helping. 
Plus, recommendations [here](https://msdn.microsoft.com/en-us/library/windows/desktop/ee890072(v=vs.85).aspx#Summary_For_Frame_Sync) seemed to help with the *other* problem, too. 

> Can't say I know what to make of that though.

The big guy [here](http://web.archive.org/web/20091216063227/http://blogs.msdn.com/larryosterman/archive/2009/09/02/what-s-the-difference-between-gettickcount-and-timegettime.aspx#comments) says that's actually bad, MMCSS funnily being one of its replacements.

turtleli:
> You prioritize the game compared to other processes?

Yes, but it's a user setting, not something a dev adds support for.

> Anyway, you can even remove the function if you want, it's just it seemed quite a lot like we were fitting all the theory for MMCSS to be justly deserved and helping.

The point is to figure out whether making the DWM opt into MMCSS makes a difference. (I grepped some of the programs on my PC and only found a trace of DWMEnableMMCSS in a Silverlight dll). The other MMCSS stuff at least seems to make an actual difference. If there is a difference, we'll know it makes a difference and can document it and revert the change if it's been merged. Personally I'm thinking of waiting till after 1.6 is released before I consider a merge.

mirh:
> Yes, but it's a user setting, not something a dev adds support for.

~~Uuuh..~~
Actually, they had [planned](https://arstechnica.com/gaming/2017/01/windows-10-game-mode-a-free-fps-boost-of-a-few-percent-with-more-to-come/) to offer devs this capability too. 
And indeed, it seems like they are [already](http://www.businessinsider.com/windows-10-game-mode-explained-2017-3) shipping this with some of *their* games. 

> The other MMCSS stuff at least seems to make an actual difference. 

What? You mean.. AvSetMmThreadCharacteristics?

turtleli:
> Actually, they had planned to offer devs this capability too.
> And indeed, it seems like they are already shipping this with some of their games.

Okay, sure, I'm wrong. I'm not interested in some on by default functionality. Sounds like a waste of time.

> What? You mean.. AvSetMmThreadCharacteristics?

Yes? Did you not find evidence that it matters in some cases in one of your links? (BTW, can't you just leave the URLs visible? It's hard to find things.)

mirh:
Noted. 
https://www.forum.rme-audio.de/viewtopic.php?id=17575
But if you are talking about this thread, they mention *additional tweaks to profiles* **in addition** to "plain" MMCSS scheduling enabling.. But afaiu that's not what was used in the benchmarks you see. 

https://blogs.technet.microsoft.com/markrussinovich/2007/08/26/vista-multimedia-playback-and-network-throughput/
Btw, just knowing MMCSS keeps DPC latency under control really, really would make me hope that every game used it.

turtleli:
> But if you are talking about this thread, they just *mention* additional tweaks to profiles in addition to "plain" MMCSS scheduling enabling..
> But afaiu that's not what was used in the benchmarks you see.

Not sure what you mean? My assumption is "MMCSS on" in the graph means AvSetMmThreadCharacteristics was called with the task set to Audio or Pro Audio.

mirh:
I guess you are right, sorry. 
Back to us, you could even merge this. Even assuming DwmEnableMMCSS is worth it I think in a perfect world that should be placed in gsdx (and possibly only be enabled with emulation running). 
Which, hoping somebody will find the insipration, is probably better to merge with the stuttering-related work.

gregory38:
so what is the status/summary ?

mirh:
That you might want to check if https://github.com/pcsx2/pcsx2/issues/529#issuecomment-293110693 is sound 😛 
Then imo, prioritizing everything the OS allows you to prioritize as if we ere The Most Important App On Earth seems bad.. 
Though having *at least* a single one of these tricks (say, on the most timing sensible thread/element of the calculating/rendering chain of pcsx2) would feel nice.

turtleli:
> so what is the status/summary ?

Merge after 1.6 is released to see if the code actually did anything?

gregory38:
ok. Fine let's postpone it.

mirh:
I'm [getting](https://social.technet.microsoft.com/Forums/windows/en-US/d4f155d8-6ff6-4741-bb2d-8ed184b7a45c/dual-monitors-windows-7-problems) reports that DWM going dumb -for some reasons- might actually be the cause for the need of [DWMFlush](https://github.com/PCSX2/pcsx2/issues/1437). 
(possibly also [DVR](https://www.reddit.com/r/Brawlhalla/comments/5shti9/plea_to_everyone_who_uses_a_nvidia_card_diakou/ddflptd/) on W10)

@Nicholas-Steel @belmont1990 opinions?

Nicholas-Steel:
Yes DWMFlush is important to solving erroneous visual stutter under Windows Vista and newer (I mean frame pacing issues, which is not the same as performance related stutter). VerdugaGreeneyes wrote some great information on how the DWM works when he was helping me with BSNES, he modified the emulator in a way that fixed the frame pacing issues I was getting while the DWM was active and his changes worked brilliantly! The changes however have never been implemented in official releases of the emulator for various reasons (boo!).

I still have copies of his excellent information and you can read it all in this image: http://imgur.com/a/Imuhu

mirh:
> can't disable DWM under Windows 10

You can't disable it (at least officially) but you can certainly bypass it. 
In similar cases for example [firefox](https://github.com/mozilla/gecko-dev/blob/master/gfx/thebes/gfxWindowsPlatform.cpp) (and [Verduga](https://www.vogons.org/viewtopic.php?f=32&t=30205#p386624) himself) uses WaitForVBlank

Though, we usually "stay" behind Wx comfort most of times. 

> alarmingly this seems to be very slowly getting replaced by Borderless Fullscreen/pretend Fullscreen in the last 2~ years

Funnily we are getting that backwards 😄 

---
Anyway one would wonder the exact reason for this to happen. 
Because, I mean.. 7-8 years ago when cpu were mostly dual core, I see how your heavy program/game stealing cycles from dwm.exe could be quite frequent. 
But today? With these powerhouses? And with pcsx2 that I'd say would barely make full use of an i3?
Sounds fishy. 

Though afaiu we are already using [DwmEnableMMCSS](https://blogs.msdn.microsoft.com/greg_schechter/2006/09/14/apis-in-the-desktop-window-manager/) and that's as far as you can go today. 
...
@WizLizard could you test this build, just for the sake of doing anything one can?

Nicholas-Steel:
Firefox doesn't have good frame pacing on my system, especially when using Adobe Flash software. Internet Explorer has excellent frame pacing even when Adobe Flash software is running (Easy test is Bubbletanks 3 game) but for some reason no other Microsoft software seems to render the same as IE like Microsoft Edge, & Windows Media Player for example frequently exhibit anomalous stutter.

You'll notice that second quote in the Verduga Greeneyes post starts with a code snippet that effectively accomplishes what I summarized in my previous post here.

WizLizard:
@mirh Sure where can i download it?

mirh:
> You'll notice that second Verduga Greeneyes quote in the Franpa post you linked to mirh, starts with a code snippet that effectively accomplishes what I summarized in my previous post here

Yes, #2000 PR is getting that into the program. 
Unfortunately it's also locks your framerate down, which is kind problematic for us. 

@WizLizard _**as always**_, just click "show all checks" > appveyor details > artifacts

WizLizard:
Ahhh ok sorry I'm not that familiar withh appveyor.

arcum42:
You know, if this isn't implemented in Windows 8.1+, and we're dropping Windows 7, maybe we should go ahead and commit this?

Nicholas-Steel:
I dunno if edits to comments trigger notifications but I edited my previous comments to make them read better.

turtleli:
It's only DwmSetPresentParameters that is not implemented in Win8.1+, not DwmEnableMMCSS.

But let's merge and find out if DwmEnableMMCSS actually makes a difference.

PR #1888, opened 2017-04-08
Exclude .vs folder

pal1000:
Visual Studio uses it to cache many things in there


gregory38:
Note that it will conflict with PR #1812

pal1000:
I'll close for now. I'll wait and see what happens with #1812.

PR #1890, opened 2017-04-08
Greg/glsl res loader

gregory38:
Implement a better shader resource loader for GSdx.

It avoids the intermediate glsl_source.h file and the related PERL script (which still exist for the deprecated zzogl).

Windows behavior must be checked. I didn't know, if `GSdx.rc` needs some `include` statements


lightningterror:
What needs testing ? I might be able to test some things.

gregory38:
First it needs to compile but I need help of @turtleli or any nice guys

turtleli:
Quick compile fix is to include resource.h in GSDeviceOGL.cpp and GSShaderOGL.cpp.

I get a black screen when I try to run it though, yet there's no shader compilation error messages.

gregory38:
Just check that shader vectors are properly populated. I'm not sure rc file is correct.

turtleli:
I think I figured it out - LoadResource on Windows doesn't null terminate. It works for DX because that also requires the size to be specified.

BTW, you might want to check the contents of plugins/res/glsl/fxaa.fx ;)

gregory38:
```
LoadResource on Windows doesn't null terminate
```
Arg. Could you try to push back a `\0`

```
BTW, you might want to check the contents of plugins/res/GSdx/glsl/fxaa.fx ;)
```
What do you mean ? It is the link to the fx file.

turtleli:
> Arg. Could you try to push back a \0

Yeah, that works.

> What do you mean ? It is the link to the fx file.

Oh, I didn't realise it was a symlink. I don't think symlinks really work on Windows (the file only contains `../fxaa.fx` here).

gregory38:
Well I'm not sure we need the link anymore. I need to check it.

gregory38:
Ok I pushed the compilation/black screen fix. Thanks you very much. Now let's look at fxaa.

turtleli:
Well, I guess remove the symlink and just use IDR_FXAA_FX (res/fxaa.fx)?

gregory38:
Yes I was on it :+1:

plugins/GSdx/GSRendererCS.cpp
```diff
@@ -532,7 +532,7 @@ void GSRendererCS::Draw()
 				{"COLOR", 1, DXGI_FORMAT_R8G8B8A8_UNORM, 0, 28, D3D11_INPUT_PER_VERTEX_DATA, 0},
 			};
 
-			vector<unsigned char> shader;
+			vector<char> shader;
 			theApp.LoadResource(IDR_CS_FX, shader);
 			dev->CompileShader((const char *)shader.data(), shader.size(), "cs.fx", nullptr, "vs_main", macro, &vs.vs, layout, countof(layout), &vs.il);
```
turtleli: Casts could be removed since the vector doesn't use unsigned char anymore (though I could always do it (and std:: prefixing) later since it's in a Windows file).

plugins/GSdx/GSdx.cpp
```diff
@@ -54,16 +54,72 @@ bool GSdxApp::LoadResource(int id, vector<unsigned char>& buff, const char* type
...
+			path = "/GSdx/res/glsl/shadeboost.glsl";
+			break;
+		case IDR_TFX_VGS_GLSL:
+			path = "/GSdx/res/glsl/tfx_vgs.glsl";
+			break;
+		case IDR_TFX_FS_GLSL:
+			path = "/GSdx/res/glsl/tfx_fs.glsl";
+			break;
+		default:
+			ret = false;
```
turtleli: I think an early return would be better here.

plugins/GSdx/GSdx.cpp
```diff
@@ -54,16 +54,72 @@ bool GSdxApp::LoadResource(int id, vector<unsigned char>& buff, const char* type
...
+		}
+
+		buff.clear();
+		buff.resize(size + 1);
+		memcpy(buff.data(), data, size + 1);
+
+		g_bytes_unref(bytes);
+	}
+
+	ASSERT(ret);
```
turtleli: `assert`?

gregory38:
@turtleli 
Thanks you for the review. Hopefully it is better now.

plugins/GSdx/GSTextureFX9.cpp
```diff
@@ -183,7 +183,7 @@ void GSDevice9::SetupPS(PSSelector sel, const PSConstantBuffer* cb, PSSamplerSel
 
 		CComPtr<IDirect3DPixelShader9> ps;
 
-		vector<unsigned char> shader;
+		vector<char> shader;
 		theApp.LoadResource(IDR_TFX_FX, shader);
 		CompileShader((const char *)shader.data(), shader.size(), "tfx.fx", "ps_main", macro, &ps);
```
turtleli: You missed this file.

plugins/GSdx/GSTextureFX11.cpp
```diff
@@ -261,7 +261,7 @@ void GSDevice11::SetupPS(PSSelector sel, const PSConstantBuffer* cb, PSSamplerSe
 
 		CComPtr<ID3D11PixelShader> ps;
 
-		vector<unsigned char> shader;
+		vector<char> shader;
 		theApp.LoadResource(IDR_TFX_FX, shader);
 		CompileShader((const char *)shader.data(), shader.size(), "tfx.fx", nullptr, "ps_main", macro, &ps);
```
turtleli: This one too.

plugins/GSdx/GSdx.h
```diff
@@ -53,7 +53,7 @@ class GSdxApp
 	int GetPrivateProfileInt(const char* lpAppName, const char* lpKeyName, int nDefault, const char* lpFileName);
 #endif
 
-	bool LoadResource(int id, vector<unsigned char>& buff, const char* type = NULL);
+	bool LoadResource(int id, vector<char>& buff, const char* type = NULL);
```
turtleli: `std::`, `nullptr`

plugins/GSdx/GSdx.cpp
```diff
@@ -43,7 +43,7 @@ BOOL APIENTRY DllMain(HMODULE hModule, DWORD ul_reason_for_call, LPVOID lpReserv
 	return TRUE;
 }
 
-bool GSdxApp::LoadResource(int id, vector<unsigned char>& buff, const char* type)
+bool GSdxApp::LoadResource(int id, vector<char>& buff, const char* type)
```
turtleli: `std::`

plugins/GSdx/GSdx.cpp
```diff
@@ -52,18 +52,70 @@ bool GSdxApp::LoadResource(int id, vector<unsigned char>& buff, const char* type
...
 	memcpy(buff.data(), LockResource(hGlobal), size);
 	return true;
 }
 
 #else
 
-bool GSdxApp::LoadResource(int id, vector<unsigned char>& buff, const char* type)
+#include "GSdxResources.h"
+
+bool GSdxApp::LoadResource(int id, vector<char>& buff, const char* type)
```
turtleli: `std::`

plugins/GSdx/GSdx.cpp
```diff
@@ -52,18 +52,70 @@ bool GSdxApp::LoadResource(int id, vector<unsigned char>& buff, const char* type
...
+			printf("LoadResource not implemented for id %d\n", id);
+			return false;
+	}
+
+	GBytes *bytes = g_resource_lookup_data(GSdx_res_get_resource(), path.c_str(), G_RESOURCE_LOOKUP_FLAGS_NONE, NULL);
+
+	size_t size = 0;
+	const void* data = g_bytes_get_data(bytes, &size);
+
+	if (data == NULL || size == 0) {
```
turtleli: `nullptr`

plugins/GSdx/GSdx.cpp
```diff
@@ -52,18 +52,70 @@ bool GSdxApp::LoadResource(int id, vector<unsigned char>& buff, const char* type
...
+			break;
+		case IDR_TFX_CL:
+			path = "/GSdx/res/tfx.cl";
+			break;
+		default:
+			printf("LoadResource not implemented for id %d\n", id);
+			return false;
+	}
+
+	GBytes *bytes = g_resource_lookup_data(GSdx_res_get_resource(), path.c_str(), G_RESOURCE_LOOKUP_FLAGS_NONE, NULL);
```
turtleli: `nullptr`

plugins/GSdx/CMakeLists.txt
```diff
@@ -289,23 +288,30 @@ if(OPENCL_FOUND)
...
     COMMAND glib-compile-resources --sourcedir "${RESOURCE_DIR}" --generate-header
         --c-name GSdx_res "${RESOURCE_DIR}/gsdx-res.xml" --target=GSdxResources.h
-    DEPENDS res/gsdx-res.xml res/logo-ogl.bmp)
+        DEPENDS res/gsdx-res.xml ${RESOURCE_FILES})
 add_custom_command(
     OUTPUT GSdxResources.cpp
     COMMAND glib-compile-resources --sourcedir "${RESOURCE_DIR}" --generate-source
         --c-name GSdx_res "${RESOURCE_DIR}/gsdx-res.xml" --target=GSdxResources.cpp
-    DEPENDS res/gsdx-res.xml res/logo-ogl.bmp)
+        DEPENDS res/gsdx-res.xml ${RESOURCE_FILES})
```
turtleli: I don't think this should be further indented since DEPENDS is one of the cmake options like OUTPUT and COMMAND.

gregory38:
Once upon a time, I planed to run clang-tidy modernize to remove NULL and others nicety such as
```
	modernize-loop-convert => 141
	modernize-make-unique => 22
	modernize-pass-by-value => 8
	modernize-redundant-void-arg => 43
	modernize-use-auto => 122
	modernize-use-default => 104
	modernize-use-nullptr => 1789
	modernize-use-override => 261
```

Anyway, I hope I didn't miss any of your feedbacks :)

plugins/GSdx/CMakeLists.txt
```diff
@@ -289,23 +288,30 @@ if(OPENCL_FOUND)
...
     COMMAND glib-compile-resources --sourcedir "${RESOURCE_DIR}" --generate-header
         --c-name GSdx_res "${RESOURCE_DIR}/gsdx-res.xml" --target=GSdxResources.h
-    DEPENDS res/gsdx-res.xml res/logo-ogl.bmp)
+	DEPENDS res/gsdx-res.xml ${RESOURCE_FILES})
 add_custom_command(
     OUTPUT GSdxResources.cpp
     COMMAND glib-compile-resources --sourcedir "${RESOURCE_DIR}" --generate-source
         --c-name GSdx_res "${RESOURCE_DIR}/gsdx-res.xml" --target=GSdxResources.cpp
-    DEPENDS res/gsdx-res.xml res/logo-ogl.bmp)
+	DEPENDS res/gsdx-res.xml ${RESOURCE_FILES})
```
turtleli: Use spaces instead? Same above.

Apart from that, it seems fine.

turtleli:
> Once upon a time, I planed to run clang-tidy modernize to remove NULL and others nicety

I see make_unique is on the list - I think that requires c++14 and gcc 4.9 minimum. Apparently rhel based distros use 4.8.x, and I thought Debian was slow.

plugins/GSdx/CMakeLists.txt
gregory38: I really need to reconfigure my text editor to use space by default :)

gregory38:
```
I see make_unique is on the list - I think that requires c++14 and gcc 4.9 minimum. Apparently rhel based distros use 4.8.x, and I thought Debian was slow.
```
I think it could be done after the 1.6 release. Debian current stable is 4.9. And old stable end of life is May 2018. You're right, RHEL is 4.8. I better understand why we just upgraded to 4.8 at work. It seem that RH provides a special version for dev `Red Hat Developer Toolset (DTS)`.

Anyway, my main concern is that we need to enable C++14 to enable it. I would like to keep pure C++11. But we can do a local implementation, just copy the code from the libstd. And ifdef it.

PR #1892, opened 2017-04-09
Greg/gsdx cross libc

gregory38:
The PR replaces windows code by small wrappers.

With #1890 it should remove all the remaining windows ifdef in OpenCL code.


plugins/GSdx/GSUtil.cpp
```diff
@@ -442,16 +428,34 @@ GSRendererType GSUtil::GetBestRenderer()
 	return GSRendererType::DX9_HW;
 }
 
-#else
+#endif
 
 void GSmkdir(const char* dir)
 {
+#ifdef _WIN32
+	if (!CreateDirectory(dir, NULL)) {
```
turtleli: `nullptr` ;)

Otherwise, I think it's fine.

gregory38: Damn, I should create a git hook to check NULL in the commit diff.

Issue #1418, opened 2016-06-18
All games: xbox controller and OnePad have sticky directions


PaulBGD:
Whenever I press up, down, left, or right, it will keep moving in that direction until I press a different direction.

Ubuntu 16.04

```
PCSX2 1.4.0-0- compiled on Jan  6 2016
Savestate version: 0x9a0b0000

Host Machine Init:
    Operating System =  Linux 4.4.0-24-generic x86_64
    Physical RAM     =  16008 MB
    CPU name         =  Intel(R) Core(TM) i7-5820K CPU @ 3.30GHz
    Vendor/Model     =  GenuineIntel (stepping 02)
    CPU speed        =  3.298 ghz (12 logical threads)
    x86PType         =  Standard OEM
    x86Flags         =  bfebfbff 7ffefbbf
    x86EFlags        =  2c100000

Loading plugins...
    Binding   GS: /usr/lib/games/PCSX2/libGSdx-1.0.0.so 
    Binding  PAD: /usr/lib/games/PCSX2/libonepad-1.1.0.so 
    Binding SPU2: /usr/lib/games/PCSX2/libspu2x-2.0.0.so 
    Binding CDVD: /usr/lib/games/PCSX2/libCDVDnull.so 
    Binding  USB: /usr/lib/games/PCSX2/libUSBnull-0.7.0.so 
    Binding   FW: /usr/lib/games/PCSX2/libFWnull-0.7.0.so 
    Binding DEV9: /usr/lib/games/PCSX2/libdev9null-0.5.0.so 
Plugins loaded successfully.
```

Catarax:
When you map a key with pcsx2 on a controller, at the first try, it may map a weird direction control and the second time, it'll map something right. 
And then you have to delete the wrong key you set at start and check if the game don't move anymore.
A bit messy.

I got some of theses mistakes, when i tried to map my 360 guitar to pcsx2 recently.

PaulBGD:
I remapped every button and still no luck.

Catarax:
Even a double time the key until you get the correct value? Possible that's more a linux issue about detecting here.

Catarax:
Found a fix, only check "XINPUT" controller in Lilypad settings before going to map a 360 controller.

PaulBGD:
That'd be nice, but sadly linux only has onepad.

gregory38:
Could you test the PR #1895 ? If you pad is correctly detected, the mapping will be done automatically and hopefully right.

[...]

Issue #1385, opened 2016-06-04
[OnePad] [Linux] DualShock 4 D-Pad key presses are not being recognised (via Bluetooth)


hazrpg:
- PCSX2 version: 1.4.0
- PCSX2 options: Defaults, no settings modifications affect the issue.
- Plugins used: GSdx [1.0.0], SPU2-X [2.0.0], OnePad [1.1.0], the rest are null plugins.
- Plugin settings: Defaults, no settings modifications affect the issue.
- Description: All regular keys map successfully (and correctly) however the D-Pad always maps to the sixaxis tilt positions (despite not tilting the controller). Sometimes the keys map to `JPOVR-0` or `JPOVL-0`, with repeated succession of trying to map them - but I can never got the opposite mappings. So JPOVR-0 seems to be right, and JPOVL-0 seems to be left - both seem to set randomly (but possibly due to the fact that I need to mash the keys in succession for it to be received). Mapping on any other software or emulators seem to be fine, the issue only seems to lie in PCSX2 and OnePad on Linux (Ubuntu 16.04).
- How to replicate: Pair the DS4 controller via Bluetooth (PS Button + Share button together to enter pair mode), open PCSX2 and attempt to map the d-pad buttons.
- Last known version to work: Never worked, tried by compiling the master branch, and the prepackaged PPA version.
- PC specifications: CPU: Intel(R) Core(TM) i7-2670QM CPU @ 2.20GHz :: GPU: Nvidia Optimus GeForce 540M (and built-in Intel) :: OS: Ubuntu 16.04 (amd64 build - Linux Kernel 4.4.0-22).

hazrpg:
I have managed to resolve this by manually setting the the hex numbers, however the numbers used when it tries to map manually are different to the ones that get "found". I ended up using these values:

    [0][12] = 0x30100
    [0][13] = 0x30200
    [0][14] = 0x30400
    [0][15] = 0x30800

The values that were being picked up were:

`0x30400` (working left), `0x2000a` (sixaxis tilt up), `0x30200` (working right)

I have attached the (now) working OnePAD.ini file (found in ~/.config/PCSX2/inis_1.4.0/) for others who may need it:
[OnePAD.ini.txt](https://github.com/PCSX2/pcsx2/files/298960/OnePAD.ini.txt)

gregory38:
Could you post the game-pad name detected by onepd. (on 1.4 look at the top with button/axis number) ?

Which version of SDL2 do you have ?

[...]

Issue #1291, opened 2016-04-13
[onepad][linux] DS3 clone not recognized at all


clbr:
I have a DS3 clone controller connected via USB. The system sees it fine, other emulators see it fine, but onepad doesn't see it at all - the configuration dialog only offers keyboard/mouse in the dropdown.

I'm using 1.4.0 from the official ppa.

$ ./sdl-jstest --list
Found 1 joystick(s)

Joystick Name:     'Gasia Co.,Ltd PS(R) Gamepad'
Joystick Number:    0
Number of Axes:    27
Number of Buttons: 19
Number of Hats:     0
Number of Balls:    0

clbr:
I see the official build was using SDL2, so I built 1.4.0 myself with SDL 1.2. And what do you know, my controller is seen and working.

I suggest to move the official package back to SDL 1.

johnathana:
I can confirm clbr. 

I am using ubuntu:14.04 and it seems SDL2 does not detect the gamepad. On the other hand libsdl1.2debian:i386 does the job but you have to recompile onepad against that library.

Probably it's distro specific because under fedora SDL2 works fine.

gregory38:
There are several versions of SDL2. Could you give us the exact fedora/ubuntu version?

clbr:
Mine is 2.0.2+dfsg1-3ubuntu1.

gregory38:
And let's me guess fedora is 2.0.4?

johnathana:
Sorry for my late reply.

For Fedora: SDL2-2.0.3-7
For Ubuntu: 2.0.2+dfsg1-3ubuntu1.1 

Seems that SDL-1.2 detects gamepads throught /dev/input/js*, that works fine. I had to use the onepad plugin from pcsx2-1.2.2-r5896-lnx.

clbr:
> Seems that SDL-1.2 detects gamepads throught /dev/input/js*

My kernel does not have JS support at all, so SDL 1.2 also works via the evdev node under /dev/input/event*.

gregory38:
Honestly it seems to be an SDL 2.0.2 bug (SDL 2.0.4 is out)

hazrpg:
[...]

Issue #1096, opened 2016-01-09
[Feature request] [OnePAD] Gamepad hot-plugging on Linux


zaps166:
- run pcsx2,
- connect a gamepad,
- go to OnePAD settings,
- the gamepad is not visible.

gregory38:
Just keep your pad plugged ;)

FWIW, detection of gamepad is done during the init of the plugin. Potentially we can relaunch an auto detection of them in the GUI. But there are more important priority.

nhydock:
Suggestion for this: since PCSX2 has been moving over to SDL2, could OnePad make use of the [SDL_GameControllerDB](https://github.com/gabomdq/SDL_GameControllerDB) when compiled with SDL2 support?

gregory38:
SDL1 is still supported for older distribution. It is also required if wx is built against sdl1. Dunno maybe we can drop it now.

ac3d912:
This could (possibly?) address an issue where a wireless gamepad is disconnected, and then reconnected.
e.g. PS3 DualAxis controller using sixad: battery dies; charge it for a bit, plug it in...

Right now: I hope you saved.
OnePad periodically re-polls: Success!

noammor:
This enhancement will enable using PCSX2 with a Steam Controller with Steam Home Streaming, see [here](http://forums.pcsx2.net/Thread-Attempting-to-use-Steam-Home-Streaming-with-PCSX2).

FlatOutPS2:
> This enhancement will enable using PCSX2 with a Steam Controller with Steam Home Streaming

PCSX2 and other PAD plugins already support hot-plugging, so it should work for most people.

gregory38:
Onepad doesn't support hot-plug. If we can drop SDL1, we could upgrade to the game controller API. The new API would help us to support auto-configuration and hot-plugging.

gregory38:
I gave a quick look at the new game controller API which is quite nice
* Allow to automatically setup joystick (like xinput)
* Support new event for hot plugging
* Likely better support of various joystick.

However it would be more difficult to setup the mapping. Potentially it can be done with external tools. (see the list below)
* https://github.com/AntiMicro/antimicro/wiki/Adding-A-New-Game-Controller-Mapping-under-AntiMicro
```
Besides AntiMicro, Steam Big Picture Mode and the Game Controller Map test program
 bundled in the SDL source code (test/controllermap.c) are the only GUI applications for
[...]

PR #1895, opened 2017-04-13
Onepad2 : Greg/sdl2

gregory38:
Upgrade onepad to leverage new SDL2 capabilities
* hot plug
* automatic mapping a la xinput so plug and play
* better support of not standard pad (dualshock....)

It should take care of the remaining onepad issue or at least greatly improve the user experience (#1418 #1385 #1291 #1235 #1096 #414 )

But it brings interesting discussion/limitation
* Do we want to keep SDL1.2 support ?
* Mapping is based on a database (SDL built-in, env variable, resources). What do we do if gamepad isn't supported ?
   * Use old joystick code path
   * Ask the user to use an external tool to configure their pad mapping. (Stream/antimicro/export SDL_GAMECONTROLLERCONFIG=). And drop tons of code.
   * Provide a GUI to generate the mapping. And drop some codes.

There are some limitations
* Pressure emulation won't be possible
   * I think we can still get access to raw data. But it will only work if the "button" is mapped as an axis.
* We might need a `GUID <=> player number` mapping to properly support multiplayer

The PR is already big enough so I think it is a good opportunity to discuss it now.

TODO
* [x] Remove DS3 hack in the GUI
* [x] Behavior of the GUI when joystick is auto-configured
* [x] player mapping is based on index whereas it ought to use guid


gregory38:
FWIW, I'm leading toward to drop completely the support of SDL1 joystick API and to rely only on automapping.  It will remove a lots of cruft.

gregory38:
So I was in the mood to remove everything, 600 lines of code !

I took the opportunity to shuffle the code a bit. And I fixed a regression for the analog pad, in case someone did test the branch ;)

gregory38:
So I added a GUI combo box to select the pad player mapping. I need to test it further, an entry of the list used to disappear from time to time.

Now we have a huge mess of updates  ...

Hirato:
I would definitely say drop SDL 1.2, there's no point keeping it.
I presume every linux user of PCSX2 already pulls SDL2 anyway because PortAudio requires it for PulseAudio output.

I think the SDL2 gamecontroller api also needs you to activate the joystick subsystem, so it's probably not that big a deal if you keep code around to leverage it as a fallback.

SDL_Joystick does contain enumerations for things like arcade sticks, which I'm sure a lot of people would like to use for fighting games, emulated or not - no idea if these things are supported by SDL_gamecontroller.
If it's easy to generate working mappings for any such arbitrary controller which makes them compatible with SDL_gamecontroller, then I say go ahead and dump the joystick path.
Otherwise you really should keep it, and obviously enhance it to listen for `SDL_JOYDEVICEADDED` and `SDL_JOYDEVICEREMOVED` in the event loop (if you haven't already).


Also, I'm pretty sure SDL reports between -32767 and 32767 for the trigger buttons (whereas the actual hardware will report between 0-255)

gregory38:
my concern is distribution that still link wxWidget against SDL1.2 but yeah it is is time to upgrade anyway.
```
PortAudio requires it for PulseAudio output.
```
Oh, I didn't know that PortAudio supports PulseAudio.

Actually the gamecontroller API is basically a wrapper around the joystick API. Gamecontroller is a virtual joystick + a  virtual to physical mapping. On Linux various games require that you configure (v2p mapping) your PAD (if not detected) with Steam Big Picture. There are also 3rd party app such as AntiMicro. 

Onepad currently got nearly 150 builtin profiles. We just need one guy to post the profile for a missing controller and it will work for all users of the same controller. Currently we have around 600 line of code and quirk to handle the mapping. It is half-broken for some pads... So it is quite appealing :)

```
Also, I'm pretty sure SDL reports between -32767 and 32767 for the trigger buttons (whereas the actual hardware will report between 0-255), same goes for all 6 of the axis
```
Hum, the wiki said 0-32k for triggers, I will check the code.

Hirato:
Yeah, when I played with SDL_gamecontroller back in 2013, I handled the triggers with `(event.caxis.value + 32767) >> 1` to normalise it within the 0 - 32767 range.
Could legitimately just be a bug that was fixed in the interim.

gregory38:
@Hirato 
Here an extract of SDL2.0.2 code
```
Sint16
SDL_GameControllerGetAxis(SDL_GameController * gamecontroller, SDL_GameControllerAxis axis)
{
   ...

    if (gamecontroller->mapping.axes[axis] >= 0 )
    {
        Sint16 value = ( SDL_JoystickGetAxis( gamecontroller->joystick, gamecontroller->mapping.axes[axis]) );
        switch (axis)
        {
            case SDL_CONTROLLER_AXIS_TRIGGERLEFT:
            case SDL_CONTROLLER_AXIS_TRIGGERRIGHT:
                /* Shift it to be 0 - 32767. */
                value = value / 2 + 16384;
            default:
                break;
        }
        return value;

```

gregory38:
Hum, latest HG is more complicated. I can see this kind of code so I think it is still valid.
```
   468         if (axis == SDL_CONTROLLER_AXIS_TRIGGERLEFT || axis == SDL_CONTROLLER_AXIS_TRIGGERRIGHT) {
   469             bind.output.axis.axis_min = 0;
   470             bind.output.axis.axis_max = SDL_JOYSTICK_AXIS_MAX;
```

turtleli:
> my concern is distribution that still link wxWidget against SDL1.2

I think Fedora and OpenSUSE do this. For Fedora I think frantisekz has a COPR repo with an alternate wx package, but I'm not sure it's well known. I don't know if OpenSUSE has an alternate wx package that can be used with SDL2.

gregory38:
@frantisekz 
Is there any reason why Fedora still use SDL 1.2 for wxWidget ?

gregory38:
@jobermayr I think you know OpenSUSE. Do you know if wxWidget could be linked againt SDL2 (or no SDL) instead of SDL1.2 ?

gregory38:
@turtleli 
Otherwise, we can duplicate `onepad-1.3` to `onepad-legacy-1.3` and create a SDL2 only `onepad-1.4`
This way it gives people a smoother transition path. 

On the PR itself, I think we have all the features (except pressure emulation). I might need to squash some commits.

frantisekz:
@gregory38 I've contacted Fedora package maintainer of wxGTK3. It's probably just leftover IMO, but we'll see. I'll post post here when I get response.

frantisekz:
@gregory38 I received reply from package maintainer. It looks like that there is not any SDL2 support in wxGTK3. My repository apparently just disabled SDL support completely in wxGTK3 (it doesn't know --with-sdl2 nor it needs SDL2-devel to compile).

So, I tried to hack around PCSX2 (cmake/ApiValidation.cmake) and lower the FATAL_ERROR to WARNING on line 21. Compilation went well, but
```
ldd /home/fanys/pcsx2/bin/PCSX2 | grep "SDL"
	libSDL-1.2.so.0 => /lib/libSDL-1.2.so.0 (0xf58df000)
```

it looks like that it switched to SDL1.2 somewhere else. Do you have any idea how to try to "force" SDL2?

(I am compiling it with build.sh --gtk3 --release)

gregory38:
you mean for wx? Did you try to set the variable `SDL_CONFIG` to `/usr/bin/sdl2-config` (not sure of the path neither name). Wx code has a `SDL_MAJOR_VERSION` define so I think the code was ported (tbh I'm not even sure that someone really use the sdl audio plugin of wxwidget)

frantisekz:
(Check my updated post, idk if github updates it automatically)

Doing export SDL_CONFIG=/usr/bin/sdl2-config doesn't change anything. How can you tell if it's compiled against SDL2? Cause if I have wxGTK3 without SDL support (which PCSX2 detects as SDL2), it doesn't load any SDL so file.

gregory38:
```
Doing export SDL_CONFIG=/usr/bin/sdl2-config doesn't change anything. How can you tell if it's compiled against SDL2? 
```
Who ? WxWidget ? You can do a ldd on the wx lib (I don't know which one). Just to be sure, you need to export the variable before the configure step of wx. Otherwise you got a free ride to debug the configure script to understand how to set properly the lib flags (i.e. -lsdl2)

gregory38:
fwiw,
```
export SDL_CONFIG=/usr/bin/sdl2-config
./configure --with-sdl
```
Based on Makefile it seems work.
```
% g sdl2 Makefile
CFLAGS = -pthread -Wall -Wundef -O2 -fno-strict-aliasing -pthread ..... -I/usr/include/SDL2 -D_REENTRANT -D_GNU_SOURCE -fvisibility=hidden
CXXFLAGS = -DWX_PRECOMP -pthread -O2 -fno-strict-aliasing -pthread ..... -I/usr/include/SDL2 -D_REENTRANT -fvisibility=hidden -fvisibility-inlines-hidden
EXTRALIBS_SDL = -L/usr/lib/i386-linux-gnu -lSDL2
```

frantisekz:
> Who ? WxWidget ?
No, PCSX2. According to wxGTK3 package maintainer in Fedora, wxGTK currently does not support SDL2 at all, so it can be either SDL1 or noSDL.

So, if I use wxGTK3 compiled with SDL(1.2) support, PCSX2 binary is linked to SDL(1.2), but plugins  (SPU2, onepad,...) get linked to both SDL1.2 and SDL2. IDK why both versions, it doesn't make sense.

I tried to compile your (Greg/sdl2) branch with default wxGTK3 in Fedora. It crashes on opening onepad config:
Thread 1 "PCSX2" received signal SIGSEGV, Segmentation fault.
0xf614c5ed in mem_write () from /lib/libSDL-1.2.so.0

I'll try to compile wxGTK3 (and not PCSX2) with export SDL_CONFIG=/usr/bin/sdl2-config and let you know.

gregory38:
Yes SDL_CONFIG is for wx autotool build.

PCSX2 binary is linked with (from build_prof/build.ninja)
```  
LINK_LIBRARIES = common/src/Utilities/libUtilities.a common/src/x86emitter/libx86emitter.a -L/usr/lib/i386-linux-gnu -pthread -lwx_baseu-3.0 -lwx_gtk2u_core-3.0 -lwx_gtk2u_adv-3.0 -lglib-2.0 -lgobject-2.0 -latk-1.0 -lgio-2.0 -lgthread-2.0 -lgmodule-2.0 -lgdk_pixbuf-2.0 -lcairo -lpango-1.0 -lpangocairo-1.0 -lpangoft2-1.0 -lpangoxft-1.0 -lgdk-x11-2.0 -lgtk-x11-2.0 -lz -laio -lrt -ldl -lm -lwx_baseu-3.0 -lwx_gtk2u_core-3.0 -lwx_gtk2u_adv-3.0
```
In my opinion the usual suspects are
* wx_baseu-3.0 
* wx_gtk2u_core-3.0
* wx_gtk2u_adv-3.0

However spu2x and onepad will also got SDL from our build system. SDL2 by default and 1.2 if you force it with `-DSDL2_API=FALSE`.
So you have this kind of link hierarchy.
```
onepad <= wxwdiget <= SDL1.2
onepad <= SDL2
```

frantisekz:
Ok, I compiled wxGTK3 with export SDL_CONFIG=/usr/bin/sdl2-config . I checked few lines of compilation log and it actualy included SDL2.

Configuring PCSX2 went fine (no warning about wxGTK3 against SDL1.2) but your Greg/SDL2 branch is crashing on same place (after opening settings of onepad plugin).

0xf614c5ed in mem_write () from /lib/libSDL-1.2.so.0

frantisekz:
So, it's working all fine, PCSX2 and all plugins linked only to SDL2 when wxGTK3 compiled with export SDL_CONFIG=/usr/bin/sdl2-config.

I'll ask package maintainer if it's acceptable solution. Thanks!

turtleli:
> Otherwise, we can duplicate onepad-1.3 to onepad-legacy-1.3 and create a SDL2 only onepad-1.4
> This way it gives people a smoother transition path.

Well, question - is this PR for 1.6 or not? If it is, then I guess it'd be an okay short term solution.

If it's not, then another solution would be to add wxGTK to 3rdparty - that'll also solve the problem, and we won't have to wait for wxWidgets upstream to release a new version to get GUI fixes on Linux (3.0.2 was released about 2.5 years ago and newer GTK versions don't work too well with it, especially GTK3).

gregory38:
```
Well, question - is this PR for 1.6 or not? If it is, then I guess it'd be an okay short term solution.
```
Tbh, I didn't have any plan. Now with the possibility of a smooth upgrade path, I think we can do it for 1.6. This way people can still fallback on current well tested version.

The biggest limitation of this new onepad version are
* no pressure emulation
* Only SDL2
* Can only add a single unsupported controller (based on env variable so you're screw if you need 2+ pads)
* need tests on various pads (typically all DS models) on various systems (USB, bluetooth...). At least we don't have to test SDL1.2 with the old js device interface.

And we win
* plug and play
* hotplug
* better support of fancy pads
* smaller code base

```
If it's not, then another solution would be to add wxGTK to 3rdparty - that'll also solve the problem, and we won't have to wait for wxWidgets upstream to release a new version to get GUI fixes on Linux (3.0.2 was released about 2.5 years ago and newer GTK versions don't work too well with it, especially GTK3).
```
I'm afraid that all distributions will remove the 3rd party version to use an older version. Is there any distribution that provide wx 3.1 ? I've the feeling that migration will be painful if they don't release the 3.2 soon.

gregory38:
@frantisekz 
Thanks you for your help :+1: 
```
I'll ask package maintainer if it's acceptable solution. Thanks!
```
I don't know if there is any application that uses both SDL and wxWidget (except us ofc). You might want to review all packages that depends on wx to avoid the multiple so version link. If you find any, please tell me so we can transfer the info to others distributions.

I guess I would need to drop/improve the WX vs SDL check.

frantisekz:
> If you find any, please tell me so we can transfer the info to others distributions.

I've put together list of such packages that depends on both wxGTK3 and SDL1( [sorted.txt](https://github.com/PCSX2/pcsx2/files/924204/sorted.txt) ). I can make such list even for compat-wxGTK3-gtk2 (wxWidgets3 and GTK2) if it helps?

> Is there any distribution that provide wx 3.1 ?

Fedora 26+ is including wx 3.0.3(git). IDK if it helps.

Apart from that, wxGTK3 package maintainer said he'll work on proper SDL2 support for wxGTK3.

gregory38:
Are you sure that those packages depends on SDL too. I don't have any SDL dependency on Debian for `boinc-manager`. We must be sure that all those packages can be upgraded from SDL1 to SDL2... 

```
Fedora 26+ is including wx 3.0.3(git). IDK if it helps.
```
Yeah, I guess distributions will pick up only the stable version 3.2. We can't go faster than wx here. However we could prepare the transition after the 1.6 release. 
@turtleli did you enable WX28 compatibility in the 3.0 Windows build ? Did you already try the 3.1 release ?

frantisekz:
> Are you sure that those packages depends on SDL too. I don't have any SDL dependency on Debian for boinc-manager. 

I've run repoquery --recursive --whatrequires SDL to find out. It's returning boinc-manager, but it's probably a mistake - I can remove SDL and boinc-manager will still have satisfied dependencies. List of SDL dependencies looks weir anyway - 3000 thousand packages, so manual checking will probably bee necessary :/

turtleli:
> did you enable WX28 compatibility in the 3.0 Windows build ? Did you already try the 3.1 release ?

wx2.8 compatibility is disabled on Windows. I haven't tried PCSX2 with 3.1 on Windows (though the files in 3rdparty do have a few changes from 3.1), but it compiled fine on Linux the last time I checked (about a year ago), though I didn't do proper tests.

> I'm afraid that all distributions will remove the 3rd party version to use an older version. Is there any distribution that provide wx 3.1 ? I've the feeling that migration will be painful if they don't release the 3.2 soon.

I'm not aware of any distros providing 3.1, and I doubt 3.2 will be out any time soon. Dolphin seems to build wx from their externals directory, and unless I'm misunderstanding something in the distro build scripts I've been looking at (Debian/Fedora/Slackware/ArchLinux/OpenSUSE). they all use the Dolphin provided wx files.

> I've run repoquery --recursive --whatrequires SDL to find out. It's returning boinc-manager, but it's probably a mistake - I can remove SDL and boinc-manager will still have satisfied dependencies. List of SDL dependencies looks weir anyway - 3000 thousand packages, so manual checking will probably bee necessary :/

I'm guessing the issue is that anything that requires wxGTK3 on Fedora will also pull in SDL1.2 whether it actually needs it or not.

frantisekz:
> I'm guessing the issue is that anything that requires wxGTK3 on Fedora will also pull in SDL1.2 whether it actually needs it or not.

This probably happened in repoquery, but it's not the case in real use (wxGTK3 depends on SDL-devel only for build, not runtime).

> they all use the Dolphin provided wx files.

Fedora removes all bundled libraries but xxhash, look at lines 117-119:
[dolphin-emu.spec](https://src.fedoraproject.org/cgit/rpms/dolphin-emu.git/tree/dolphin-emu.spec)

turtleli:
> Fedora removes all bundled libraries but xxhash, look at lines 117-119:

Yeah, you're right, I completely missed that. I might have to look at the other ones again...

gregory38:
I did a massive rebase. Commits are much better now. I duplicate onepad to keep SDL1.2 compatibility.

@turtleli 
Thanks for the info. Unlike GTK, WX APIs are quite stable. Hopefully all distributions will switch to GTK3 when the WX3.2 is out.

Based on https://blogs.gnome.org/desrt/2016/06/13/gtk-4-0-is-not-gtk-4/ GTK3 API will stabilize soon. So I hope we could drop GTK2 and WX30 support together. But it won't be soon

frantisekz:
wxGTK3 is now built with SDL2 for Fedora 26+ ( https://src.fedoraproject.org/cgit/rpms/wxGTK3.git/commit/?id=e31717383b8086768df19a87f179a82d38a135d8 )

> GTK3 API will stabilize soon.

It is stable already AFAIK (no new major GTK3 release is planned (latest is 3.22 from Q3/2016), GTK 3.90 is already GTK4 unstable API). 

Isn't it possible to drop GTK2 but stay on wx30 (=this is the way I build PCSX2 on Fedora for a few years already)?

gregory38:
```
wxGTK3 is now built with SDL2 for Fedora 26+ 
```
Good news 👍
```
It is stable already AFAIK (no new major GTK3 release is planned (latest is 3.22 from Q3/2016), GTK 3.90 is already GTK4 unstable API).
```
Yeah it seems so.

```
Isn't it possible to drop GTK2 but stay on wx30 (=this is the way I build PCSX2 on Fedora for a few years already)?
```
Some distributions (Debian/Ubuntu/...) still build wx30 with GTK2. It isn't bad idea, as Turtleli said, wx30 got plenty of issue with GTK3.

mirh:
> Some distributions (Debian/Ubuntu/...) still build wx30 with GTK2

I'm not sure about which in particular or the possible problems.. Though new debian stable is coming fairly soon and 12.04 LTS is being dropped in.. I think a matter of months. That might change something.

danilaml:
>and 12.04 LTS is being dropped in.. I think a matter of months.

>Ubuntu 12.04.5 LTS
End of life: April 28, 2017

gregory38:
```
I'm not sure about which in particular or the possible problems.. Though new debian stable is coming fairly soon and 12.04 LTS is being dropped in.. I think a matter of months. That might change something.
```
Did you check that the future Debian release will use GTK3 or did you assume it ?
=> https://packages.debian.org/sid/libwxgtk3.0-0v5

So I'm afraid that both WX3.2 and GTK3 won't be available soon.

mirh:
> Did you check that the future Debian release will use GTK3 or did you assume it ?

Definitively the latter :)
I found [this](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=786986#41) possibly interesting backstory meanwhile. 
TL;DR they are [waiting](https://github.com/wxWidgets/wxWidgets/commit/a309157a666e2f7582a9134699a69f2548d99949) them for 3.2 too.

gregory38:
@turtleli if you don't have any objections. I will merge the branches (I put onepad duplication in a separate branch).

turtleli:
Can't say I've actually reviewed it yet so I need more time, I was looking at the progress report stuff instead since we're behind schedule (as usual).

plugins/onepad/GamePad.h
```diff
@@ -56,8 +56,14 @@ class GamePad
...
     }
 
+    bool IsProperlyInitialized()
+    {
+        return m_no_error;
+    }
+
 protected:
     int m_deadzone;
+    bool m_no_error;
```
turtleli: This isn't initialised to false, so the check might not work properly.

plugins/onepad/SDL/joystick.h
```diff
@@ -56,4 +56,5 @@ class JoystickInfo : GamePad
     SDL_GameController *m_controller;
     SDL_Haptic *m_haptic;
     std::array<int, NB_EFFECT> m_effects_id;
+    bool m_no_error;
```
turtleli: Is this one necessary?

gregory38: Oh I might scew up a bit in my rebase. I will check it at home.

You can take more time.

gregory38:
The m_no_error smells fishy.

However I suggest that we merge the first commit. It would reduce the noise/size of this PR.

turtleli:
> However I suggest that we merge the first commit. It would reduce the noise/size of this PR.

Perhaps. It might be better not to append version numbers to the plugin filenames though - it'll eventually stop users from accidentally using older versions of the plugin (I'm sure we've seen a few cases before).

gregory38:
```
Perhaps. It might be better not to append version numbers to the plugin filenames though - it'll eventually stop users from accidentally using older versions of the plugin (I'm sure we've seen a few cases before).
```
If the user does a fresh install, it will be good because I renamed the old plugin. But yeah we could go to
```
bin/plugins/libonepad-legacy.so
bin/plugins/libonepad.so
```

turtleli:
> bin/plugins/libonepad-legacy.so
> bin/plugins/libonepad.so

Yes, that's what I want - it'd be simpler for users who compile their own stuff and don't notice that the plugin version has changed.

gregory38:
Ok. Done.

plugins/onepad/SDL/joystick.h
```diff
@@ -32,59 +32,31 @@
...
 
-    void Destroy();
     // opens handles to all possible joysticks
-    static void EnumerateJoysticks(vector<GamePad *> &vjoysticks);
+    static void EnumerateJoysticks(std::vector<std::unique_ptr<GamePad>> &vjoysticks);
 
-    void Rumble(int type, int pad);
-
-    bool Init(int id); // opens a handle and gets information
+    void Rumble(unsigned type, unsigned pad);
```
gregory38: @turtleli 
one question, in your opinion, is it better to keep virtual in "not-base" class? Or is final/override enough ?

turtleli: I think override/final is enough, virtual is a bit redundant if either is specified.

gregory38:
Note: first commit is pending work on my master branch.

gregory38:
@nE0sIghT Hello, I think you still use SDL1 on Gentoo
*  http://forums.pcsx2.net/Thread-Gentoo-Linux-Fail-to-emerge-9999-version-of-PCSX2

New onepad version requires SDL2. But it would also requires to either drop SDL from WX (nobody use it, for example, Debian/Ubuntu doesn't enable it and nobody complains) or upgrade it to SDL2.

@turtleli I pushed all your change. Tell me if it is ok for you.

nE0sIghT:
@gregory38 
Thanks for notice. I will look into it, but after 15 May only.

turtleli:
> I pushed all your change. Tell me if it is ok for you.

Yeah, I think it's fine.

Side note: wx3.0.3 might be released soon - https://github.com/wxWidgets/wxWidgets/commit/b6ad01d2a99b7df98a8e876cd9df819c3c5c07be

gregory38:
Ok. Code merged. Thanks for the review.

Do we need something from Wx3.0.3 ?

turtleli:
Not really, but it should fix some of the GTK2/GTK3 issues that wx3.0.2 had.

gregory38:
Oh. It won't hurt :) Hopefully distribution will pick this new version soon.

danilaml:
Btw, wxWidgets 3.0.3 [is now out](http://www.wxwidgets.org/news/2017/05/wxwidgets-3.0.3-released/).

PR #1896, opened 2017-04-14
PCSX2-Counters: Detect DVD variant videomodes

ssakash:
**Summary of changes**:

* Removed the older generic BIOS video mode in the enumeration by expanding it with the technically accurate DVD variants of NTSC/PAL video mode to the video mode list. These new variant video mode related timings need to be corrected to get proper timing in PSX mode games, personally not interested on that quest though. _(Please look at the commit message for the whole story)_

* Remove the color burst code (hack to get the PSX video mode) and SMODE1 register union as they're no longer needed.

As a side note, might be useful to add OSD messages for the video mode changes. (Maybe?)

### DIFF
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -253,16 +253,17 @@ static const char* ReportVideoMode()
 {
 	switch (gsVideoMode)
 	{
-	case GS_VideoMode::PAL:        return "PAL";
-	case GS_VideoMode::NTSC:       return "NTSC";
-	case GS_VideoMode::VESA:       return "VESA";
-	case GS_VideoMode::BIOS:       return "BIOS";
-	case GS_VideoMode::HDTV_480P:  return "HDTV 480p";
-	case GS_VideoMode::HDTV_576P:  return "HDTV 576p";
-	case GS_VideoMode::HDTV_720P:  return "HDTV 720p";
-	case GS_VideoMode::HDTV_1080I: return "HDTV 1080i";
-	case GS_VideoMode::HDTV_1080P: return "HDTV 1080p";
-	default:                       return "Unknown";
+	case GS_VideoMode::PAL:          return "PAL";
+	case GS_VideoMode::NTSC:         return "NTSC";
+	case GS_VideoMode::DVD_NTSC:     return "DVD NTSC";
+	case GS_VideoMode::DVD_PAL:      return "DVD PAL";
+	case GS_VideoMode::VESA:         return "VESA";
+	case GS_VideoMode::HDTV_480P:    return "HDTV 480p";
+	case GS_VideoMode::HDTV_576P:    return "HDTV 576p";
+	case GS_VideoMode::HDTV_720P:    return "HDTV 720p";
+	case GS_VideoMode::HDTV_1080I:   return "HDTV 1080i";
+	case GS_VideoMode::HDTV_1080P:   return "HDTV 1080p";
+	default:                         return "Unknown";
 	}
 }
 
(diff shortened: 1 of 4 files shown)


gregory38:
@ssakash 
```
These new variant video mode related timings need to be corrected to get proper timing in PSX mode games, personally not interested on that quest though. 
```
Is it a "regression" versus the current behavior ? Or is current timing inaccurate in the first place ?

refractionpcsx2:
Current timing code is inaccurate, but it's the best we have at the moment. I am slightly hesitant messing with timers so close to release. 

Also considering how little information there is on it, I wouldn't remove the smode1 union either, that provides the structure which is difficult to find documentation on otherwise.

willkuer:
`#if 0` could be an option

gregory38:
In this case, add a comment to explain why there is an `#if 0`

ssakash:
> Is it a "regression" versus the current behavior ? Or is current timing inaccurate in the first place ?

Current timing code **_(master branch)_** is not accurate, I just simplified the variant modes timing code by using timing values similar to their base video mode (NTSC/PAL), I know the new version isn't totally accurate either but still better than the older one as we don't need to rely on colorburst anymore and we can directly apply the timing values based on the parameter of SetGsCrt. (I always worried about the possibility of 0 in CMOD bit when we needed to get a colorburst signal, I wouldn't have to worry about such corner cases in this new implementation)

>  I am slightly hesitant messing with timers so close to release.

The change in code only influences timing in DVD variant modes, so only impacts the timing logic in BIOS (BIOS isn't technically accurate to define it, but you know what I mean, right?) and PSX mode games when they're run through backward compatibility unless there are any other testcases. It would be easier to test these changes to DVD variant video modes as there are only few limited number of testcases which make use of this mode.

To be honest, I'd prefer to include this before the release as in the older code, there was always the possibility of spam in the console log due to multiple video mode inits when we used colorburst for video mode detection.

> Also considering how little information there is on it, I wouldn't remove the smode1 union either, that provides the structure which is difficult to find documentation on otherwise.

If you don't know the bitfield structure of SMODE1, you probably shouldn't be coding for a PS2 emulator. :P 

All jokes aside, I was the one who added the SMODE1 info and I don't think it's hard to get it in the first place. If I managed to get it, literally anyone in the team can easily get it.

ssakash:
Okay, fair enough. Seems like the official GS documentation doesn't cover the SMODE1 register, I'll comment out the union then.

ssakash:
Commented out the union, I have no further changes to make in this pull request.

For those who are interested in fixing the timing code of these new variant NTSC/PAL video modes:

 please look into ``vSyncTimingInfo``. You need to initialize the structure with proper values in ``vSyncInfoCalc()``, you probably need to take some considerations for the DVD NTSC/PAL mode as they might not have similar timings to the regular NTSC/PAL videomodes. (Similar to the way we deal with timings of progressive video modes in the function)

You need a good testcase to experiment with it (A specific scene with bad timing in PSX mode games for example maybe?), I don't have any PSX games so can't help there, interested people can do it.

ssakash:
> If I wanted to nitpick, I would have asked an #if 0..#endif pattern to comment the union. But I'm fine.

I'm glad that you're fine with the single line comments, saves me a bit of work. :P 

FWIW, I was discussing about this with @refractionpcsx2 in Facebook earlier:

relatable

**Note**: Had to photoshop his real life avatar to the fake one to protect his privacy. Yes, I'm really good with MS paint.

gregory38:
```
I'm glad that you're fine with the single line comments, saves me a bit of work. :P
```
Any real text editor can edit line in column ;) So it is quite easy to remove and to use a better style ;)

Anyway, I hope you don't say bad stuff on me on those social network :stuck_out_tongue:

danilaml:
In "real" text editor it's just easy to `Ctrl+/` (or `Ctrl+K, Ctrl+C` in VS) to automatically comment (or uncomment) out highlighted block of code.

gregory38:
@refractionpcsx2 Could you merge it if it is fine for you ?

refractionpcsx2:
OK I'll merge, but if this breaks stuff it's on your head :P

gregory38:
Damn it, I need to review the code OoO :p

PR #1897, opened 2017-04-14
GSdx: Move/add GT4 CRC

lightningterror:
Removed a crc hw hack that caused vram spikes on GT4 when using the Full
CRC hack level. Vram spikes should be greatly reduced now when using
Full hack level (Recomended for Direct3D).
Testing is welcomed for the vram spikes.

Added missing CRC game version for GT4 pal.

Note: The issue might be the same on GT3 and GTConcept , the code might
need to be removed for those games as well. Needs to be tested.

### DIFF
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -1370,13 +1370,18 @@ bool GSC_GT4(const GSFrameInfo& fi, int& skip)
 	// Game requires to extract source from RT (block boundary) (texture cache limitation)
 	if(skip == 0)
 	{
-		if(fi.TME && fi.FBP >= 0x02f00 && fi.FPSM == PSM_PSMCT32 && (fi.TBP0 == 0x00000 || fi.TBP0 == 0x01180 /*|| fi.TBP0 == 0x01a40*/) && fi.TPSM == PSM_PSMT8) //TBP0 0x1a40 progressive
+		if(Aggressive)
 		{
-			skip = 770;	//ntsc, progressive 1540
-		}
-		if(g_crc_region == CRC::EU && fi.TME && fi.FBP >= 0x03400 && fi.FPSM == PSM_PSMCT32 && (fi.TBP0 == 0x00000 || fi.TBP0 == 0x01400 ) && fi.TPSM == PSM_PSMT8)
-		{
-			skip = 880;	//pal
+			// Removes layer obscuring the screen when the in-game brightness or contrast setting is set to any value but 0.
+			// The hack can cause VRAM/RAM spikes when both brightness and contrast settings are set to 0.
+			if(fi.TME && fi.FBP >= 0x02f00 && fi.FPSM == PSM_PSMCT32 && (fi.TBP0 == 0x00000 || fi.TBP0 == 0x01180 /*|| fi.TBP0 == 0x01a40*/) && fi.TPSM == PSM_PSMT8) //TBP0 0x1a40 progressive
+			{
+			skip = 770;	// ntsc, progressive 1540
+			}
+			if(g_crc_region == CRC::EU && fi.TME && fi.FBP >= 0x03400 && fi.FPSM == PSM_PSMCT32 && (fi.TBP0 == 0x00000 || fi.TBP0 == 0x01400 ) && fi.TPSM == PSM_PSMT8)
+			{
+			skip = 880;	// pal
+			}
 		}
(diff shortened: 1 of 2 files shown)


MrCK1:
Doesn't this just make the CRC hack empty now?

lightningterror:
CRC hack level is still being used for the brightness/contrast on OpenGL and also on Direct3D that fixes the top left corner/garbage screen I think.

gregory38:
Yes there is another "hack" in the OGL renderer. However there is no hack for Dx. Only a minimal implementation of the channel effect. I did the minimal to reduce the memory spike. Full implementation is more complicated (well it is doable).

lightningterror:
Thanks to my game version crc not being on the list I was able to find this issue with the hack :)

FlatOutPS2:
I get the opposite behaviour. The CRC hack prevents GSDX out of memory warnings and keeps a relative stable RAM level. Tested with Direct3D 11 at 5x native with Large Framebuffer enabled and Full CRC Hack Level.

lightningterror:
I did some test with large framebuffer and I also get vram spikes but a bit less than with the code.

Can you test without large framebuffer ?

FlatOutPS2:
Removing the hack has no effect on the memory levels with or without large framebuffer.

The hack only seems useful for Direct3D.

lightningterror:
Ok I have some additional info how to reproduce vram spikes on git builds.

Brightness contrast both set to 0 (if the value is something else then there are no vram spikes)
Direct3D , crc full

Vram maxes out at 2gb , ram goes up from 2 to 6gb usage. Usage is just all over the place for me.


Removing the hack solves that, I don't see any vram spikes at all.
It's really odd how our results are quite different.

FlatOutPS2:
> It's really odd how our results are quite different.

The results will likely the same, because I did not know I had to set brightness/contrast levels to 0 to trigger to spikes, I left those on the default levels.

MrCK1:
When was the CRC first added again - I'm just trying to come to grips with this? Would the AMD driver being more or less functional at the time affect the issue with or without the CRC hack on the newer driver?

lightningterror:
> Edit: oh wait this is D3D only right?

Yes D3D only. Tho you can get the spikes on OpenGL if you crank up crc to full but that isn't recommended. That also bugs the brightness/contrast crc not to work on full/ aggressive on OpenGL.

Removing the hack solves that issue as well.


This might help you as well.
D3D with 0 values.
https://snag.gy/Kh68FB.jpg

D3D with values changed.
https://snag.gy/NxhGwI.jpg

FlatOutPS2:
@FlatOutPS2 which version do you have pal or ntcs?

I have the PAL version.

BTW, if you want to add images on github please upload 'em directly to your post.

lightningterror:
Hmm do you have any suggestions ?
What do you think about moving the hack to aggressive ?

That way people can switch between if they get vram spikes on either config.

gregory38:
@FlatOutPS2 @lightningterror could I merge the code  ?

plugins/GSdx/GSHwHack.cpp
```diff
@@ -1370,13 +1370,15 @@ bool GSC_GT4(const GSFrameInfo& fi, int& skip)
...
 	if(skip == 0)
 	{
-		if(fi.TME && fi.FBP >= 0x02f00 && fi.FPSM == PSM_PSMCT32 && (fi.TBP0 == 0x00000 || fi.TBP0 == 0x01180 /*|| fi.TBP0 == 0x01a40*/) && fi.TPSM == PSM_PSMT8) //TBP0 0x1a40 progressive
-		{
-			skip = 770;	//ntsc, progressive 1540
-		}
-		if(g_crc_region == CRC::EU && fi.TME && fi.FBP >= 0x03400 && fi.FPSM == PSM_PSMCT32 && (fi.TBP0 == 0x00000 || fi.TBP0 == 0x01400 ) && fi.TPSM == PSM_PSMT8)
-		{
-			skip = 880;	//pal
+		if(Aggressive) { // Can fix or cause VRAM and RAM spikes. Depends on userconfig , game version and maybe hardware.
```
FlatOutPS2: `// Removes layer obscuring the screen when the in-game brightness or contrast setting is set to any value but 0. The hack can cause VRAM/RAM spikes when both brightness and contrast settings are set to 0.`

lightningterror:
@gregory38 This one is ready. Did a couple of tests again and everything seemed ok.

I'll update/test dbz bt3 tomorrow.

gregory38:
@FlatOutPS2 is it ok for you ?

FlatOutPS2:
Yes, it's fine. 👍

PR #1900, opened 2017-04-14
GSdx: Remove/move DBZ:BT3 CRC

lightningterror:
Enables character outlines to partially work on Full CRC. More info in commit message.
Great for Direct3D users :)

Images for comparison.
OpenGLHW
https://snag.gy/rGYpeg.jpg

Direct3DHW
https://snag.gy/pLyhZD.jpg

### DIFF
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -78,21 +78,15 @@ bool GSC_DBZBT3(const GSFrameInfo& fi, int& skip)
 {
 	if(skip == 0)
 	{
-		if(fi.TME && fi.FBP == 0x01c00 && fi.FPSM == PSM_PSMCT32 && (fi.TBP0 == 0x00000 || fi.TBP0 == 0x00e00 || fi.TBP0 == 0x01000) && fi.TPSM == PSM_PSMT8H)
+		if(fi.TME && (fi.FBP == 0x00000 || fi.FBP == 0x00e00 || fi.FBP == 0x01000) && fi.FPSM == PSM_PSMCT32 && fi.TPSM == PSM_PSMT8H)
 		{
-			//not needed anymore?
-			//skip = 24; // blur
-		}
-		else if(fi.TME && (fi.FBP == 0x00000 || fi.FBP == 0x00e00 || fi.FBP == 0x01000) && fi.FPSM == PSM_PSMCT32 && fi.TPSM == PSM_PSMT8H)
-		{
-			if (Dx_only) { // Ought to be fine with blending accuracy (fbmask?)
+			if (Aggressive) { // Partially works on D3D. OpenGL has no such issues.
 				if(fi.FBMSK == 0x00000)
 				{
-					skip = 28; // outline
-				}
-				if(fi.FBMSK == 0x00FFFFFF)
-				{
-					skip = 1;
+					// Hack removes character outlines as well as black borders on the sides of the screen.
+					// Without the hack one of the borders at the bottom of the screen can start shaking when using the Direct3D 9 renderer. This is caused by resolution upscale.
+					// This can be fixed by using the TC Offsets x and y hardware hacks or using Native resolution.
+					skip = 28;
 				}
 			}
 		}
@@ -100,17 +94,18 @@ bool GSC_DBZBT3(const GSFrameInfo& fi, int& skip)
 		{
 			// Texture shuffling must work on openGL
 			if (Dx_only)
-				skip = 5;
+				skip = 5; // Sky texture (depth related). On Direct3D the blue sky texture is shown on the whole screen in front of the player if hack is disabled.
 		}
 		else if(fi.TME && fi.FPSM == fi.TPSM && fi.TBP0 == 0x03f00 && fi.TPSM == PSM_PSMCT32)
 		{
+			// Ghosting (glow) effect. Ghosting appears when resolution is upscaled. Doesn't appear on native resolution.
 			if (fi.FBP == 0x03400)
 			{
-				skip = 1;	//PAL
+				skip = 1; // PAL
 			}
 			if(fi.FBP == 0x02e00)
 			{
-				skip = 3;	//NTSC
+				skip = 3; // NTSC
 			}
 		}
 	}


lightningterror:
Also I have a question regarding this particular code

```
				if(fi.FBMSK == 0x00FFFFFF)
				{
					skip = 1;
				}
```
What did it do on dbz bt3 , I wasn't able to spot any difference disabling it.

Also I wasn't able to spot any vertical lines on both direct3d and opengl with crc set to none which is a really good thing :)

gregory38:
If you understand why you don't need it anymore, you can remove it. Otherwise you can comment it temporally. One day it will be garbage collected ;)

Tbh, it would be better to implement depth support on D3D before purging all those hacks.

lightningterror:
I'll just disable it instead of removing it since outlines partially work , it can be purged at a later date like you said when depth support is added.

As for the hack code I mentioned above I still have no idea what it does , I'm switching between on and off and can't find a difference.

gregory38:
Could you test openGL without accurate blending (BSOD...) ? Without depth support? My guess is that it is an effect that require accurate blending.

lightningterror:
I'm getting a weird issue , disabling the two codes , even just the outlines gives a weird issue on object shadows when I move (moving shadows blurry ) but if I remove the code or move it to Aggressive the issue is gone.

Don't merge this yet , I want to do some testing , don't want to add a regression on my hand.

gregory38:
```
don't want to add a regression on my hand.
```
You mean the head with a nice `Wanted Reward` :laughing:

lightningterror:
Okay so after further investigation the second code I mentioned above has something to do with shadows on stationary objects , right now when enabled when a player moves the shadows move on an object as well in a weird motion blur way. The issue is present on the latest git as well so it seems disabling it fixes the problem , seems I stumbled upon a fix instead of a regression.

Now the question is to purge or to comment the code. I'm tempted to purge it.

plugins/GSdx/GSHwHack.cpp
```diff
@@ -85,22 +85,18 @@ bool GSC_DBZBT3(const GSFrameInfo& fi, int& skip)
 		}
 		else if(fi.TME && (fi.FBP == 0x00000 || fi.FBP == 0x00e00 || fi.FBP == 0x01000) && fi.FPSM == PSM_PSMCT32 && fi.TPSM == PSM_PSMT8H)
 		{
-			if (Dx_only) { // Ought to be fine with blending accuracy (fbmask?)
+			if (Aggressive) { // Partially works on Direct3D check git #1900
```
FlatOutPS2: You should explain the issue in the comment, not mention some git PR future coders might not be able to find. ;)

lightningterror:
Is it better now? Also can you do some tests to see if everything is alright?
Or maybe @ssakash can if he has it.

FlatOutPS2:
> Is it better now?

The line from the your commit messages seems better for the `aggressive `hack.  `outlines` and `sky texture` need details, which outlines(character outlines IIRC) / sky textures and what's the issue?

gregory38:
@FlatOutPS2 @lightningterror could I merge the code  ?

FlatOutPS2:
@gregory38 No, some of the details are still missing in the comments. @lightningterror also asked for a test, but no one has tested it yet(I don't have the game).

lightningterror:
I updated the PR , did you check the changes ?

plugins/GSdx/GSHwHack.cpp
```diff
@@ -85,22 +85,18 @@ bool GSC_DBZBT3(const GSFrameInfo& fi, int& skip)
...
-			if (Dx_only) { // Ought to be fine with blending accuracy (fbmask?)
+			if (Aggressive) { // Partially works. A small black line shakes at the bottom on dx9. Can be solved with x y offset or removing outlines with Aggressive crc.
 				if(fi.FBMSK == 0x00000)
 				{
-					skip = 28; // outline
-				}
-				if(fi.FBMSK == 0x00FFFFFF)
-				{
-					skip = 1;
+					skip = 28; // Character outlines
```
FlatOutPS2: It needs a note on why it removes the character outlines.

lightningterror: I mentioned that in the above comment. The black line shaking at the bottom is the result of not full implementation yet. Aside from that the code can be purged but I left it there just in case people don't want to have the 50% outlines working and the shaking on dx9.
There are no other issues , maybe in the past it was connected with the vertical red lines but that issue has been fixed.

plugins/GSdx/GSHwHack.cpp
```diff
@@ -85,22 +85,18 @@ bool GSC_DBZBT3(const GSFrameInfo& fi, int& skip)
 		}
 		else if(fi.TME && (fi.FBP == 0x00000 || fi.FBP == 0x00e00 || fi.FBP == 0x01000) && fi.FPSM == PSM_PSMCT32 && fi.TPSM == PSM_PSMT8H)
 		{
-			if (Dx_only) { // Ought to be fine with blending accuracy (fbmask?)
+			if (Aggressive) { // Partially works. A small black line shakes at the bottom on dx9. Can be solved with x y offset or removing outlines with Aggressive crc.
```
FlatOutPS2: `dx9` -> Direct3D 9

`x y offset` -> TC Offset x & y (does it need the x value though?)

lightningterror: From the way I tested it seems so.

lightningterror:
I updated the comments @FlatOutPS2 thoughts ?

plugins/GSdx/GSHwHack.cpp
```diff
@@ -85,22 +85,22 @@ bool GSC_DBZBT3(const GSFrameInfo& fi, int& skip)
...
 				if(fi.FBMSK == 0x00000)
 				{
-					skip = 28; // outline
-				}
-				if(fi.FBMSK == 0x00FFFFFF)
-				{
-					skip = 1;
+					// Hack removes character outlines as well as black borders on the sides of the screen.
+					// In the past the hack was used to remove red vertical lines on the screen I believe but now that issue has been fixed.
+					// The outlines partially work so on D3D9 can cause one of the borders at the bottom to shake, this can be fixed by removing the outlines with Aggressive crc or both x and y offsets.
```
FlatOutPS2: `Without the hack one of the borders at the bottom of the screen can start shaking when using the Direct3D 9 renderer. This can be fixed by using the TC Offsets x and y hardware hacks.`

The line below this can be removed as far as I'm concerned.

plugins/GSdx/GSHwHack.cpp
```diff
@@ -85,22 +85,22 @@ bool GSC_DBZBT3(const GSFrameInfo& fi, int& skip)
...
+			if (Aggressive) { // Partially works on D3D. OpenGL has no such issues.
 				if(fi.FBMSK == 0x00000)
 				{
-					skip = 28; // outline
-				}
-				if(fi.FBMSK == 0x00FFFFFF)
-				{
-					skip = 1;
+					// Hack removes character outlines as well as black borders on the sides of the screen.
+					// In the past the hack was used to remove red vertical lines on the screen I believe but now that issue has been fixed.
```
FlatOutPS2: This line can be removed, as I'm pretty sure the `skip = 1` you removed did that, it shouldn't apply to the remaining hack.

If you want to confirm it you should build the GSdx from v1.4.0 from source after removing the outline hack `skip = 28` and test the CRC hack.

plugins/GSdx/GSHwHack.cpp
```diff
@@ -85,22 +85,22 @@ bool GSC_DBZBT3(const GSFrameInfo& fi, int& skip)
 		}
 		else if(fi.TME && (fi.FBP == 0x00000 || fi.FBP == 0x00e00 || fi.FBP == 0x01000) && fi.FPSM == PSM_PSMCT32 && fi.TPSM == PSM_PSMT8H)
 		{
-			if (Dx_only) { // Ought to be fine with blending accuracy (fbmask?)
+			if (Aggressive) { // Partially works on D3D. OpenGL has no such issues.
```
FlatOutPS2: Does it not require the blending unit accuracy as mentioned in the removed comment?

lightningterror: I can't check that since I have an amd gpu. Don't wanna get a driver crash ...

plugins/GSdx/GSHwHack.cpp
lightningterror: You're right , skip 1 did that , I wasn't able to find any render problems with the outlines on 1.4 and compiling 1.2 seems a bit of a pain since it's for vs 13

Honestly I have no idea why outlines were removed unless they caused some specific issues pre 1.4

lightningterror:
Updated with some additional info. Also removed the blur hack which was commented out since it wasn't even used on 1.4

gregory38:
Is it ready ?

lightningterror:
Depends if @FlatOutPS2 approves. If not then back to the editing.

gregory38:
@FlatOutPS2 is it fine for you ?

bositman:
Could you attach the modified gsdx so i could take a look? Cant compile atm.

gregory38:
@bositman 
All PR are automatically compiled by appveyor.

bositman:
This should be fine after some chatting with @lightningterror and some tests of my own ;)

plugins/GSdx/GSHwHack.cpp
```diff
@@ -78,39 +78,34 @@ bool GSC_DBZBT3(const GSFrameInfo& fi, int& skip)
...
+			// Ghosting (glow) effect. Ghosting appears when resolution is upscaled. Doesn't appear on native resolution.
 			if (fi.FBP == 0x03400)
 			{
-				skip = 1;	//PAL
+				skip = 1; // PAL
 			}
 			if(fi.FBP == 0x02e00)
 			{
-				skip = 3;	//NTSC
+				skip = 3; // NTSC
```
ssakash: Maybe check for NTSC/PAL videomode before applying the hack? (necessary function is ``GetVideoMode()``)

lightningterror:
@bositman which version did you test? I have the ntcs disc.
Here are some ss's for comparison

PS2
img_20170609_032807
img_20170609_033129

PCSX2 - Direct3D Native - no hacks
pcsx2 2017-06-09 03-46-14-64
pcsx2 2017-06-09 03-32-15-17

PCSX2 - Direct3D 3x native
pcsx2 2017-06-09 03-47-37-75
pcsx2 2017-06-09 03-47-55-93

woj1993:
@lightningterror This bug is on OpenGl too.
@gregory38 I think it can be interesting to you? In this game there is more than one error on OpenGl. Do you want GsDump?
How it looks:
https://www.dropbox.com/s/u9128vwup8b7yl7/gsdx_20170612124229.png?dl=0
How it should:
https://www.dropbox.com/s/jcqzevakveh4pod/gsdx_20170612124245.png?dl=0

plugins/GSdx/GSHwHack.cpp
lightningterror: I rather not touch anything else and leave things as they were :)

ssakash: I don't mind if you don't want to do it (hence a normal comment and not a comment requesting a change via Git review system), I was just more or less concerned about the PAL version using a similar FBP value for doing another effect. (We probably wouldn't want to skip it at such cases)

lightningterror:
I know it is , I'm just putting comparison between my pr and the ps2 version. Showing that there are no regressions.

Anyway ready for merge.

PR #1901, opened 2017-04-15
Debugger: Make memory view row size configurable

Kingcom:
This makes the row size of the memory view configurable. Previously it was hard coded to 32 by #1663, which required you to have a very wide debug window if you wanted to actually see the ASCII part of the memory view. The value is read from the ini file. Another setting added by that pull request is also permanently remembered now.


gregory38:
Welcome back :+1:

PR #1902, opened 2017-04-16
Debugger: Use armips as submodule

Kingcom:
This isn't ready for merging yet (the Visual Studio configurations need to be handled properly and it needs to be added to cmake), but it likely needs some discussions so I'm opening this early.

What this does is remove the crippled assembler I put into the debugger a few years ago and instead replace it with an armips submodule, similar to what was done in PPSSPP. The assemble feature so far is greatly limited and doesn't even support basic macros like li. Using armips directly, almost its entire feature set can be used with little extra work.

The consequences are that you need to run "git submodule update --init --recursive" after cloning.


gregory38:
How does it interact with cmake ? How does the versioning work ? For example, what happen if you commit stuff in your submodule ? Do we get latest version after a `submodule update` ?

Kingcom:
A submodule is a reference to a separate repository. It references a specific commit of that repository, so it doesn't automatically update to the latest version when commits happen over there. ```submodule update``` does a checkout of the referenced submodule versions. 
As for building, you just need to reference the files like with any internal folders.

gregory38:
Ok. It is not clear where is stored the reference but it isn't important.

However I would prefer to separate submodule from 3rdparty directory. So it is easier to know which directory is handled by submodule. The other reason is that 3rdparty is optional on Linux, it is allowed to remove it on Linux to save disk space.

@turtleli @gigaherz @refractionpcsx2 
I know that giga doesn't like submodule ^^ But if we start to use submodule for armips, we might end up use it for others 3rdparty. So let's have a discussion, please.

gigaherz:
Don't mind me. I'm not developing pcsx2 these days, so do whatever makes you most comfortable.

willkuer:
Licenses are compatible?

Kingcom:
armips uses the MIT license, and tinyformat the Boost license. So yes, compatible.

gregory38:
You need to add include dir here
https://github.com/PCSX2/pcsx2/blob/master/cmake/SearchForStuff.cmake

And it would be nice to reduce a bit the gcc warning spamming :)

gregory38:
Appveyvor info, I'm not sure we need it as your repo is public.
* https://www.appveyor.com/docs/how-to/private-git-sub-modules/

turtleli:
Yeah, AppVeyor doesn't init submodules automatically, unlike Travis CI. It's easy to fix (I can do it if you want me to).

I'm not sure what submodule discussion you want though? If it's about using submodules, I'm fine with it.

I'm guessing for this we might need to maintain a separate VS project file from the one in the repo? (I looked at PPSSPP and that's what they seem to do.)

Kingcom:
I think that was only needed because they wanted to link the Visual Studio runtime library statically. That doesn't seem to be an issue here. I'm purging clang warnings at the moment, sure are a lot indeed...

turtleli:
Well, if we don't use a separate VS project file then I think we'll always have to use the same SDKs and toolkits. I don't know if this'll be a problem.

gregory38:
@Kingcom I hope you're no trying to fix all clang warnings :)

gregory38:
@Kingcom Could you rebase on master so we can see if bots are happy with submodule.

Kingcom:
Sorry, it's been a busy time lately. I'll try to purge those warnings this weekend. I had about half of the files cleaned up locally, so it shouldn't take that long anymore.

ramapcsx2:
Is the submodule necessary though? Do they allow someone just taking a copy of the current snapshot and using it? We could put it in 3rd party then. It doesn't seem like a module that requires a lot of updates.

Kingcom:
When you copy the code, there's always the chance that local changes will be made in ways that would make any future updates harder. A submodule would prevent both that and have less redundancy in general. 
I'm not sure when I'll get to finish this, so I'll close it for now. I'll get back when it's working.

mirh:
> so it doesn't automatically update to the latest version when commits happen over there

IIRC if you are brave enough, you can point submodule directly to a specific branch (such as master)

> Do they allow someone just taking a copy of the current snapshot and using it? We could put it in 3rd party then. It doesn't seem like a module that requires a lot of updates.

It's certainly less clutter in our .git history though.

PR #1905, opened 2017-04-18
GSdx: Move CRC hacks

FlatOutPS2:
Moves Resident Evil 4 hack to Aggressive level as it is no longer required to fix any issues, but does offer a decent speed boost.

Moves The Getaway CRC hack to Full level, as the issue can be resolved when using the OpenGL Hardware renderer.

### DIFF
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -629,23 +629,6 @@ bool GSC_TombRaiderUnderWorld(const GSFrameInfo& fi, int& skip)
 	return true;
 }
 
-bool GSC_FFVIIDoC(const GSFrameInfo& fi, int& skip)
-{
-	if(skip == 0)
-	{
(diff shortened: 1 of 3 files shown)


lightningterror:
You beat me to it , wanted to make the RE 4 pr as well but was kinda skeptical since I wanted to test more :)

plugins/GSdx/GSHwHack.cpp
```diff
@@ -629,23 +629,6 @@ bool GSC_TombRaiderUnderWorld(const GSFrameInfo& fi, int& skip)
 	return true;
 }
 
-bool GSC_FFVIIDoC(const GSFrameInfo& fi, int& skip)
```
gregory38: Might worth to add in the commit message that it was related/is handled by OI_DoubleHalfClear.

FlatOutPS2: Are you sure OI_DoubleHalfClear also removed the garbage data that was shown when objects (dis)appear in the world?

gregory38: I'm not 100% sure. But it took me 3 gsdump to fix DHC so it could look great. And the fact that it is also working on Dx mean that it can't be a depth effect. Hum, it could be a texture cache bug that was fixed too. But DHC is likely main culprit.

PR #1906, opened 2017-04-18
gsdx: port X11 GLX/EGL to XCB/X11 GLX/EGL

gregory38:
Unfortunately there is a strong connection between GLX and X11
so only half the code is ported to XCB

At least the ported stuff is thread safe (which is better for MT driver)

After the 1.6 release, we will drop GLX in favor of EGL.
Then it will be easier to port the remaining code.

Edit: oups I forget to remove the old XInitThread commented code

### DIFF
--- a/plugins/GSdx/GSWnd.cpp
+++ b/plugins/GSdx/GSWnd.cpp
@@ -23,6 +23,11 @@
 #include "stdafx.h"
 #include "GSWnd.h"
 
+#if defined(__unix__)
+#include <X11/Xlib-xcb.h>
+#include <X11/Xlib.h>
+#endif
+
 void GSWndGL::PopulateGlFunction()
 {
 	*(void**)&(gl_BlendColor) = GetProcAddress("glBlendColor");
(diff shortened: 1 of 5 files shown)


gregory38:
@turtleli 
I need your opinion, initially I wanted to use common code for GLX/EGL. But (display/window) types aren't the same so it mean template and stuff. In the end, I plan to drop GLX after the 1.6 release (I hope AMD supports EGL properly). So I'm not sure it worth to bother it.

turtleli:
If you plan to drop GLX then yeah, I wouldn't bother.

Though for some reason EGL doesn't work for me when I use FreeBSD (and my internet is down right now which doesn't help).

gregory38:
What is your driver on FreeBSD ? Wayland will require EGL, actually I don't know if they will use it on *BSD.

turtleli:
NVIDIA 375.39. Could be a driver thing (didn't have time to debug, but IIRC the trace showed it crashing at glGetString).

I don't think Wayland is ready on FreeBSD, though they're working on it.

gregory38:
They enabled full openGL/EGL recently on Linux https://github.com/PCSX2/pcsx2/commit/99c43881df351474df05d7bc5f0c79443adc831b
I will ask you to test EGL again after the 1.6 release :)

Main goal of this PR was to avoid some crashes with Mesa+glthread. Unfortunately it doesn't really solve the crash. I'm waiting a clear status of Mesa direction to take a decision. But I will likely close the PR.

turtleli:
> They enabled full openGL/EGL recently on Linux

Yeah, I didn't test much, but it did seem to work fine on ArchLinux.

BTW, I completely forgot to mention this, but master doesn't compile with EGL enabled since CheckContext doesn't exist anymore ;)

gregory38:
Ah yes, I nearly forget it (I removed the bad line in this commit)

gregory38:
I have another branch with a better solution for EGL. I think Xlib will remain for GLX.

PR #1907, opened 2017-04-18
pcsx2: Fix commandline options

### LINKED ISSUE #1099: PCSX2 Commandline options --usecd & *.IsoFile don't override CDVD iso/Plugin selection.
PCSX2 version: any I've tried up to latest git release: pcsx2-v1.5.0-dev-27

-Description: 
If you want to use a commandline argument to load an ISO or a CD/DVD and the opposing option is selected in PCSX2 the commandline option will fail.
If you use the commandline option "--usecd" (boots from the CDVD plugin (overrides IsoFile parameter) and have iso selected in the CDVD tab it doesn't load the CD and gives two error messages in the log.
http://oi66.tinypic.com/2hg7hi1.jpg
These are the errors.
"CDVDiso Error: No filename specified."
and
"(pxActionEvent) CDVD plugin failed to open!(thread:EE Core)"
The disc will load correcty if you select Plugin in the tab and then launch PCSX2 with the --usecd commandline option.
With Plugin selected though, loading an iso with a commandline option( PCSX2 "C:\game.iso") will fail and load a disc instead. This however will not give an error message in the log. The problem seems to be that they don't override the parameter set it in PCSX2.

-How to replicate: Select iso in the CDVD tab, close PCSX2 and launch PCSX2 with the commandline option --usecd
Or:
Select Plugin in the CDVD tab, close PCSX2 and launch PCSX2 with an iso gamefile as a commandline option.

turtleli:
Changes:
 * Fixes #1099
 * Fixes the --nodisc commandline option
 * EDIT: Prevent the current ISO from being cleared if not autorunning an ISO from the command line

### DIFF
--- a/pcsx2/gui/AppInit.cpp
+++ b/pcsx2/gui/AppInit.cpp
@@ -495,11 +495,8 @@ bool Pcsx2App::OnInit()
 
 		if( Startup.SysAutoRun )
 		{
-			// Notes: Saving/remembering the Iso file is probably fine and desired, so using
-			// SysUpdateIsoSrcFile is good(ish).
-			// Saving the cdvd plugin override isn't desirable, so we don't assign it into g_Conf.
-
 			g_Conf->EmuOptions.UseBOOT2Injection = !Startup.NoFastBoot;
+			g_Conf->CdvdSource = Startup.CdvdSource;
 			SysUpdateIsoSrcFile( Startup.IsoFile );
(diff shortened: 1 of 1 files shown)


lightningterror:
Hmm this disables the ability to have shortcuts made for each game that launch each game automatically.

turtleli:
Try latest master. I pushed a fix.

lightningterror:
Great , it works again.

PR #1908, opened 2017-04-18
spu2-x:xaudio2: Improve exception handling

turtleli:
Changes:
 * Combine the XAudio 2.7/2.8+ code into an inline file with ifdefs to avoid some code duplication - most of the code does the same thing and there's not that much which is 2.7/2.8+ specific (Note the XAudio 2.7/2.8+ interfaces are slightly incompatible and use different xaudio2.h headers, which is why separate modules are needed).
 * Improve exception handling to avoid resource leaks.
 * Add comments for stuff that might need investigating...


plugins/spu2-x/src/Windows/SndOut_XAudio2.inl
```diff
@@ -86,14 +86,18 @@ private:
...
         BaseStreamingVoice(uint numChannels)
-            : m_nBuffers(Config_XAudio2.NumBuffers)
+            : pSourceVoice(nullptr)
+            , qbuffer(nullptr)
+            , m_nBuffers(Config_XAudio2.NumBuffers)
             , m_nChannels(numChannels)
             , m_BufferSize(SndOutPacketSize * m_nChannels * PacketsPerBuffer)
             , m_BufferSizeBytes(m_BufferSize * sizeof(s16))
         {
+            InitializeCriticalSection(&cs);
```
gregory38: maybe use std::mutex or atomic_flag (wrapped into a class) instead of critical section ?

plugins/spu2-x/src/Windows/SndOut_XAudio2.inl
```diff
@@ -132,14 +132,13 @@ private:
             pSourceVoice->FlushSourceBuffers();
             pSourceVoice->Start(0, 0);
 
-            qbuffer = new s16[m_nBuffers * m_BufferSize];
-            ZeroMemory(qbuffer, m_BufferSizeBytes * m_nBuffers);
+            m_buffer = std::make_unique<s16[]>(m_nBuffers * m_BufferSize);
```
gregory38: did you remove the ZeroMemory on purpose ?

plugins/spu2-x/src/Windows/SndOut_XAudio2.inl
```diff
@@ -182,7 +181,7 @@ private:
             }
 
             EnterCriticalSection(&cs);
-            safe_delete_array(qbuffer);
+            m_buffer = nullptr;
```
gregory38: it should be useless (generally speaking). However still using an object in a thread and executing the destructor in another thread isn't nice.

plugins/spu2-x/src/Windows/SndOut_XAudio2.inl
turtleli: Yes. make_unique value initialises each element to 0.

plugins/spu2-x/src/Windows/SndOut_XAudio2.inl
turtleli: I don't think the buffer is used by any other thread when this code is reached. That's what the documentation suggested to me but I didn't remove the code just in case I'm wrong.

I didn't change the critical section to a std::mutex since I was wondering whether it was necessary in the first place (FWIW I looked at Dolphin's XAudio2 code and they don't use any mutexes).

plugins/spu2-x/src/Windows/SndOut_XAudio2.inl
```diff
@@ -16,42 +16,37 @@
...
+#include <stdexcept>
+#include <string>
 
 namespace Exception
 {
 class XAudio2Error : public std::runtime_error
 {
-protected:
-    static const char *SomeKindaErrorString(HRESULT hr)
+private:
```
turtleli: I suppose so. I could probably remove it (or swap the order and have the public stuff first).

plugins/spu2-x/src/Windows/SndOut_XAudio2.inl
gregory38:
```
Yes. make_unique value initialises each element to 0.
```
Is it a windows behavior ? On linux, it is only a wrapper to `unique_ptr(new...)`. I don't see any think related on it on the GCC header. Note I didn't look at the standard.

plugins/spu2-x/src/Windows/SndOut_XAudio2.inl
gregory38: ok

plugins/spu2-x/src/Windows/SndOut_XAudio2.inl
turtleli: No, it's not Windows specific. See http://stackoverflow.com/questions/42140212/does-make-unique-value-initializes-char-array

gregory38: Oh. Ok. It calls the default constructor.

gregory38:
Code is fine for me. Could I merge it ?

turtleli:
Yeah, it's probably fine to merge.

gregory38:
good jobs :+1:

PR #1909, opened 2017-04-19
gsdx sw: always enable the gs memory wrapping emulation

gregory38:
I think it is better this way.


FlatOutPS2:
Code seems fine to me. I also tested the PR with all the HW and SW renderers and it works without issue.

PR #1910, opened 2017-04-19
Greg/common make unique

gregory38:
Question do we also do it on CDVDgiga and GSdx ?

### DIFF
--- a/common/include/Utilities/MakeUnique.h
+++ b/common/include/Utilities/MakeUnique.h
@@ -1,0 +1,73 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2017-2017  PCSX2 Dev Team
+ *
+ *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
+ *  of the GNU Lesser General Public License as published by the Free Software Found-
+ *  ation, either version 3 of the License, or (at your option) any later version.
+ *
+ *  PCSX2 is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+ *  PURPOSE.  See the GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along with PCSX2.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <memory>
+
+// make_unique is quite handy but it requires to enable C++14 support (of
+// course on C++14 compliant compiler)
+//
+// Instead just provide the std++ implementation when only C++11 is enabled
+// File could be dropped when we switch to C++14
+
+#if __cplusplus <= 201103L && !defined(_MSC_VER)
+
+namespace std
+{
+
+template <typename _Tp>
+struct _MakeUniq
+{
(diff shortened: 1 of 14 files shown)


gregory38:
Hum is C++14 enabled on Visual Studio ? Or does it export std::make_unique even on C++11 build ?

turtleli:
make_unique has been available since VS2013.

I think both VS2015 and VS2017 use C++14.

gregory38:
I'm not sure it is a good idea to enable C++14 construct. I mean that C++11 code will compile all compliers. C++14 code might depends on the compiler.

turtleli:
There isn't an option to compile as C++11 with VS. Right now it's either C++14, or C++ latest (follows latest drafts).

common/include/Utilities/MakeUnique.h
```diff
@@ -23,7 +23,7 @@
 // Instead just provide the std++ implementation when only C++11 is enabled
 // File could be dropped when we switch to C++14
 
-#if __cplusplus <= 201103L
+#if __cplusplus <= 201103L && !defined(_WIN32)
```
mirh: `_MSC_VER` feels more orthodox 👅

gregory38: You're right but I fail to find the define name. I will update it.

gregory38:
```
There isn't an option to compile as C++11 with VS. Right now it's either C++14, or C++ latest (follows latest drafts).
```
Oh. Ok. It isn't nice but we will manage.

gregory38:
https://blogs.msdn.microsoft.com/vcblog/2016/06/07/standards-version-switches-in-the-compiler/
```
Note that there is no plan to add a C++11 switch.

There are some post-C++14 features already implemented in previous releases of the Visual C++ compiler. To avoid breaking changes for users who have already taken dependencies on these features we’ll leave them under the /std:c++14 switch. They include the following features, all of which have already shipped in Visual Studio 2015 Update 2:

    New rules for auto with braced-init-lists
    typename in template template-parameters
    Removing trigraphs
    Attributes for namespaces and enumerators
    u8 character literals
```
So the switch enables some C++17 construct. And I don't think they support all C++14... 

Nevertheless, it would be nice to enable the C++14 switch.

mirh:
> So the switch enables some C++17 construct. And I don't think they support all C++14 (in VS2015)...

https://docs.microsoft.com/en-us/cpp/visual-cpp-language-conformance
Afaiu they are just missing a couple of C++14 features in VS2015. 

[In other news](https://blogs.msdn.microsoft.com/vcblog/2017/04/17/cmake-support-in-visual-studio-whats-new-in-2017-15-2-update/), seems like they are pretty quick with updating their CMake support.

gregory38:
Well C++14 isn't that big, so a couple is nearly half ;) The point is to have the same feature set on all compilers. I found it easier this way but I'm maybe oldschool.

Hopefully someone will plug CMake/VS properly on PCSX2.

common/include/Utilities/MakeUnique.h
```diff
@@ -0,0 +1,73 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2010  PCSX2 Dev Team
```
turtleli: Wrong copyright date?

gregory38: Done. thank you.

PR #1914, opened 2017-04-23
Greg/gsdx gtk3

gregory38:
Replace GTK  deprecated code by modern GTK3. Nothing was tested yet. And I suspect that I misses some paddings for the table. Anyway it is a start.

(PS: yeaah I upgraded my debian to stretch)

### DIFF
--- a/plugins/GSdx/GSLinuxDialog.cpp
+++ b/plugins/GSdx/GSLinuxDialog.cpp
@@ -26,11 +26,24 @@
 #include "GSdxResources.h"
 #include "GSSetting.h"
 
+// Port of deprecated GTK2 API to recent GTK3. Those defines
+// could prove handy for testing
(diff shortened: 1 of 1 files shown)


gregory38:
@turtleli feels free to give it a look. I'm really not confident on the monitor API to get the screen size.

turtleli:
The GtkGrid stuff doesn't seem to expand to use the full dialog width.

I haven't looked at the monitor stuff yet.

gregory38:
Oh. It seems it is done with properties. I need to look at it. And I need to find a way to run the gui in standalone mode.

turtleli:
Latest commit doesn't have any effect.

gregory38:
Ok. I propose to merge this PR with both GTK3 define disabled. I need to rebuild wx with GTK3 support to implement the code properly. At least the PR factorize a bit the code so it would reduce the number of warning and will help future fixes.

PR #1915, opened 2017-04-23
zh_CN translation update

KazushiMe:
Well, couldn't find po files of plugins.


gregory38:
thanks you for the translation

PR #1917, opened 2017-04-24
Greg/gcc6 warning

gregory38:
A bunch of GCC6 compiler fix.

### DIFF
--- a/pcsx2/GS.cpp
+++ b/pcsx2/GS.cpp
@@ -25,7 +25,6 @@
 using namespace Threading;
 using namespace R5900;
 
-static GS_VideoMode s_ColorBurst = GS_VideoMode::Uninitialized;
 __aligned16 u8 g_RealGSMem[Ps2MemSize::GSregs];
 
 void gsOnModeChanged( Fixed100 framerate, u32 newTickrate )
--- a/pcsx2/Gif_Unit.h
+++ b/pcsx2/Gif_Unit.h
@@ -79,6 +79,7 @@ struct Gif_Tag {
 		hasAD = false;
 		nRegIdx = 0;
 		isValid = 1;
+		len = 0; // avoid uninitialized compiler warning
 		switch(tag.FLG) {
 			case GIF_FLG_PACKED:
 				nRegs  = ((tag.NREG-1)&0xf) + 1;
--- a/pcsx2/IopGte.cpp
+++ b/pcsx2/IopGte.cpp
@@ -468,9 +468,9 @@ __inline double limG2(double x) {
 	return (x);
 }
 
-__inline s32 F12limA1S(s64 x) { _LIMX(-32768 << 12, 32767 << 12, 24); }
-__inline s32 F12limA2S(s64 x) { _LIMX(-32768 << 12, 32767 << 12, 23); }
-__inline s32 F12limA3S(s64 x) { _LIMX(-32768 << 12, 32767 << 12, 22); }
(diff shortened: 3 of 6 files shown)


woj1993:
@gregory38 I don't think _LIMX(-(32768 << 12) gives the same result as _LIMX(-32768 << 12). This probably need testing.

gregory38:
Gdb output
```
(gdb) p/x (-32768<<12)
$2 = 0xf8000000
(gdb) p/x -(32768<<12)
$3 = 0xf8000000
```
So I think it should work.

woj1993:
Ok. I was worrying because on:http://stackoverflow.com/questions/8415895/is-left-and-right-shifting-negative-integers-defined-behavior. In supercat answer was this line: "if x is type int, x<<y is equivalent to (int)((unsigned)x << y)." So I was worrying that x sign will get lost. But if it gives the same result then it is ok.

gregory38:
You right it is quite tricky. The issue is the overflow case for example
```
(gdb) p/x (-32768) << 16
$1 = 0x80000000
```
```
(gdb) p/x (32768) << 16
$1 = 0x80000000
```
Which is -2 147 483 648 in signed logic. However the biggest int value is 2 147 483 64**7**

woj1993:
What about other cases?

gregory38:
It is the only case that I'm not sure about the behavior. Others are fine because we don't overflow.

woj1993:
@gregory38 And doesn't sign change on other cases?
for example old code gd.l = GSVector4((float)(**-0x10000** << context->TEX1.L))); 
if supercat is right about:
> if x is type int, x<<y is equivalent to (int)((unsigned)x << y).

Wouldn't be compiled as gd.l = GSVector4((float)(**0x10000** << context->TEX1.L))); and new version is 
gd.l = GSVector4((float)(**-(0x10000 << context->TEX1.L)**));

gregory38:
hypothesis `a << b` is same as `a*2^b`

so `(-a) << b`=> `(-a)*2^b` => `-1*(a*2^b)`=> `-1 * (a << b)`

Note `unsigned(-1)` is `0xFFFF_FFFF`not 1

woj1993:
> Note unsigned(-1) is 0xFFFF_FFFFnot 1
I totally forgot about that. :(
What are we doing in this line __inline s32 F12limA3S(s64 x) { _LIMX(-(32768 << 12), 32767 << 12, 22); } ?
Are we casting s64 x to s32 x?

gregory38:
Hard-coded constant are 32 bits. You need to add the `ll` suffix to get a 64 bits. However here, macro will upcast the value to 64 bits. But it is awful, here template will be nicer.

Yes s32 is int. Minimum value is –2 147 483 648 but maximum value is 2 147 483 647. So -INT_MIN is bigger than INT_MAX.

gregory38:
@turtleli could you review the change. Thanks you.

turtleli:
It's probably fine? I think none of the bit shifts are large enough to invoke undefined behaviour. (If you like reading standards, it's section 5.8 [expr.shift], the C++11 and C++14 standards differ slightly on what undefined behaviour is in this case).

gregory38:
Could you provide an example of assert?

turtleli:
I think it'd only apply for the spu2-x clamp_mix functions. Something like `assert(bitshift <= 15);` in `clamp_mix(s32 x, u8 bitshift)`?

gregory38:
Yes it is a good idea. I did it :+1:

Issue #1916, opened 2017-04-23
[Feature Request] Integer scaling/Fit to height toggle


rz5:
This is about how the image is displayed in PCSX2.

Currently, the fit of the video is always set to "fit to height". This stretches the image while keeping the user-chosen aspect ratio.

This feature request is to have a checkbox on the GS Window options labelled, for example, "Integer scaling" where the final image size is NOT stretched e.g. if the frame size is 640x480, then display a centered 640x480 frame while leaving the rest black.

The reason for this feature request is to have a setting to display a pixel-perfect (i.e. unstretched) image, because stretching the image to a non-integer multiple resolution results in blurriness because we don't get a 1:1 pixel correspondence. 

Back of the envelope implementation would be:
```
if (gs_window_options.integer_scaling == true) {
    /* Old behavior in case frame overflows from window */
    if (frame_size > gs_window.size) {
        fit_to_height();
    } else {
        present_as_is();
    }
}        

...
```

gregory38:
```
 because stretching the image to a non-integer multiple resolution results in blurriness because we don't get a 1:1 pixel correspondence
```
Potentially the option linear present could help to avoid the blur

rz5:
@gregory38 Using nearest-neighbor vs bilinear filtering would help combat the blur, yes. It doesn't fix the lack of 1:1 pixel ratio.

Also, while on that topic, I suggest refactoring occurrences of "Texture Filtering of Display" and "linear_present" to "Bilinear filtering". I had to trudge around the source code and in the GUI to understand what linear present meant.

James-F2:
PCSX2 painfully lacks Integer-Scaling feature.

Another good way to scale the image is multiply the image by X times and only then stretch it.
This option is also not available in PCSX2.

Boviligg:
Has there been any work or updates regarding this issue? It would really be worth it. Thanks.

Muhlawor:
Can it be added?

DukeSkinny:
Indeed, it looks like much of the work has been done already (https://github.com/PCSX2/pcsx2/pull/1918), but that's an old unfinished PR that seems wholly dropped.

I too would love the option to enable integer scaling, which would tremendously with running PCSX2 artifact-free to my VGA monitor. Currently, the way to do it is to lock the window size to the game's resolution & stretch the screen using the monitor buttons.

refractionpcsx2:
Qt will have integer scaling (it already has)

PR #1312, opened 2016-04-26
Automatic aspect ratio switch during FMV playback

FlatOutPS2:
Add option to Emulation Settings - GS Window to for automatically switching to 4:3 aspect ratio when a FMV plays, and back to widescreen/stretch when it's finished.

aspectswitch

Code probably requires optimising to follow PCSX2 coding guidelines, as I'm not a PCSX2 code or C++ expert.

Feature-request by trostboot as posted in issues: #1212

Update: latest commit Appveyor build link(April 29th): https://ci.appveyor.com/api/buildjobs/9lu8c562g7cuy52a/artifacts/pcsx2-v1.5.0-dev-678-gf6c11a1-master-933-vs2015-Win32-AppVeyor.7z

PR #1918, opened 2017-04-26
PCSX2: Scaling Types / 16:9 FMV aspect ratio switch

FlatOutPS2:
- Adds Scaling Types

And adds a combo box(in Config > Emulation Settings > GS Window tab) to
select the scaling type.

Adds new "Centered" option to scale the image down to a centered image
at the native(1x) resolution used by the game.

And adds "Integer" option to scale to the maximum integer magnification
of the native image set by the game which fits within the window/screen.

Adds new aspect ratio "Frame" which sets the aspect ratio to that of the
native image used by the game.
Closes https://github.com/PCSX2/pcsx2/issues/1916

- Add 16:9 option for FMV aspect ratio switch

Adds combobox option to switch to 16:9 aspect ratio while an FMVs plays to fix
16:9 widescreen FMV's when using custom widescreen patches to play a
game at other aspect ratios than 4:3 or 16:9.

Close: #2114

### DIFF
--- a/pcsx2/PluginManager.cpp
+++ b/pcsx2/PluginManager.cpp
@@ -1,5 +1,5 @@
 /*  PCSX2 - PS2 Emulator for PCs
- *  Copyright (C) 2002-2010  PCSX2 Dev Team
+ *  Copyright (C) 2002-2018  PCSX2 Dev Team
  *
  *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
  *  of the GNU Lesser General Public License as published by the Free Software Found-
@@ -171,6 +171,7 @@ _GSinitReadFIFO    GSinitReadFIFO;
 _GSinitReadFIFO2   GSinitReadFIFO2;
 _GSchangeSaveState GSchangeSaveState;
 _GSgetTitleInfo2   GSgetTitleInfo2;
+_GSgetImageSize    GSgetImageSize;
 _GSmakeSnapshot	   GSmakeSnapshot;
 _GSmakeSnapshot2   GSmakeSnapshot2;
 _GSirqCallback 	   GSirqCallback;
@@ -423,6 +424,7 @@ static const LegacyApi_OptMethod s_MethMessOpt_GS[] =
 	{	"GSinitReadFIFO",	(vMeth**)&GSinitReadFIFO	},
 	{	"GSinitReadFIFO2",	(vMeth**)&GSinitReadFIFO2	},
 	{	"GSgifTransfer1",	(vMeth**)&GSgifTransfer1	},
+	{	"GSgetImageSize",	(vMeth**)&GSgetImageSize	},
 	{ NULL }
 };
 
--- a/pcsx2/gui/FrameForGS.cpp
+++ b/pcsx2/gui/FrameForGS.cpp
@@ -641,6 +641,10 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
 	out << std::fixed << std::setprecision(2) << fps;
 	OSDmonitor(Color_StrongGreen, "FPS:", out.str());
 
+	const u64 &smode2 = *(u64 *)PS2GS_BASE(GS_SMODE2);
+	int interlacing = smode2 & 1;      // INT: 0 = Progressive, 1 = Interlaced
+	int interlacing_mode = smode2 & 2; // FFMD: 0 = Field mode, 2 = Frame mode
+
 	AspectRatioType Aspect_Ratio = g_Conf->GSWindow.AspectRatio;
 
 	bool supported_GSWindow_Settings = (Aspect_Ratio == AspectRatio_Frame || (Aspect_Ratio != AspectRatio_Stretch &&
@@ -648,8 +652,13 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
 
 	if ( GSgetImageSize != nullptr && supported_GSWindow_Settings )
 	{
+		bool framemode_compensation = !g_Conf->GSWindow.DisableScalingCompensation && interlacing == 1 && interlacing_mode == 2;
+
 		int new_width = 0, new_height = 0;
 		GSgetImageSize(&new_width, &new_height);
+		if ( framemode_compensation && new_height > 1 )
+			new_height *=2;
+
 		if ( new_width > 1 && new_height > 1 && (s_image_width != new_width || s_image_height != new_height) )
 		{
 			if ( GSPanel* gsPanel = GetViewport() )
@@ -688,9 +697,8 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
 		}
 	}
 
-	const u64& smode2 = *(u64*)PS2GS_BASE(GS_SMODE2);
-	wxString omodef = (smode2 & 2) ? templates.OutputFrame : templates.OutputField;
-	wxString omodei = (smode2 & 1) ? templates.OutputInterlaced : templates.OutputProgressive;
+	wxString omodef = interlacing_mode == 2 ? templates.OutputFrame : templates.OutputField;
+	wxString omodei = interlacing == 1 ? templates.OutputInterlaced : templates.OutputProgressive;
 
 	wxString title = templates.TitleTemplate;
 	title.Replace(L"${slot}",		pxsFmt(L"%d", States_GetCurrentSlot()));
--- a/pcsx2/gui/FrameForGS.cpp
+++ b/pcsx2/gui/FrameForGS.cpp
@@ -1,5 +1,5 @@
 /*  PCSX2 - PS2 Emulator for PCs
- *  Copyright (C) 2002-2010  PCSX2 Dev Team
+ *  Copyright (C) 2002-2018  PCSX2 Dev Team
  *
  *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
  *  of the GNU Lesser General Public License as published by the Free Software Found-
@@ -29,6 +29,9 @@
 #include <sstream>
 #include <iomanip>
 
+static int s_image_width;
+static int s_image_height;
+
 static const KeyAcceleratorCode FULLSCREEN_TOGGLE_ACCELERATOR_GSPANEL=KeyAcceleratorCode( WXK_RETURN ).Alt();
 
 //#define GSWindowScaleDebug
@@ -56,6 +59,7 @@ void GSPanel::InitDefaultAccelerators()
 	m_Accels->Map( AAC( WXK_TAB ).Shift(),		"Framelimiter_SlomoToggle" );
 
 	m_Accels->Map( AAC( WXK_F6 ),				"GSwindow_CycleAspectRatio" );
+	m_Accels->Map( AAC( WXK_F6 ).Shift(),		"GSwindow_CycleScalingType" );
 
 	m_Accels->Map( AAC( WXK_NUMPAD_ADD ).Cmd(),			"GSwindow_ZoomIn" );	//CTRL on Windows/linux, CMD on OSX
 	m_Accels->Map( AAC( WXK_NUMPAD_SUBTRACT ).Cmd(),	"GSwindow_ZoomOut" );
@@ -162,37 +166,48 @@ void GSPanel::DoResize()
 	extern AspectRatioType iniAR;
 	extern bool switchAR;
 	double targetAr = clientAr;
+	bool acceptedImageSize = s_image_width > 0 && s_image_width <= client.GetWidth() && s_image_height > 0 && s_image_height <= client.GetHeight();
+	AspectRatioType Aspect_Ratio = g_Conf->GSWindow.AspectRatio;
 
 	if (g_Conf->GSWindow.AspectRatio != iniAR) {
 		switchAR = false;
 	}
 
-	if (!switchAR) {
-		if (g_Conf->GSWindow.AspectRatio == AspectRatio_4_3)
+	if (!switchAR && Aspect_Ratio != AspectRatio_Stretch) {
+		if (Aspect_Ratio == AspectRatio_4_3)
 			targetAr = 4.0 / 3.0;
-		else if (g_Conf->GSWindow.AspectRatio == AspectRatio_16_9)
+		else if (Aspect_Ratio == AspectRatio_16_9)
 			targetAr = 16.0 / 9.0;
+		else if (Aspect_Ratio == AspectRatio_Frame)
+			targetAr = acceptedImageSize ? (double)s_image_width / s_image_height : 4.0 / 3.0;
 	} else {
 		targetAr = 4.0 / 3.0;
 	}
 
 	double arr = targetAr / clientAr;
 
-	if( arr < 1 )
+	if (arr < 1)
 		viewport.x = (int)( (double)viewport.x*arr + 0.5);
-	else if( arr > 1 )
+	else if (arr > 1)
 		viewport.y = (int)( (double)viewport.y/arr + 0.5);
 
+	if (Aspect_Ratio != AspectRatio_Stretch && g_Conf->GSWindow.ScalingType != ScalingType_Fit && acceptedImageSize) {
+		int image_width_AR = s_image_height * targetAr;
+		int scaling = g_Conf->GSWindow.ScalingType == ScalingType_Integer ? std::min(client.GetWidth() / image_width_AR, client.GetHeight() / s_image_height) : 1;
+		viewport.x = image_width_AR * scaling;
+		viewport.y = s_image_height * scaling;
+	}
+
 	float zoom = g_Conf->GSWindow.Zoom.ToFloat()/100.0;
-	if( zoom == 0 )//auto zoom in untill black-bars are gone (while keeping the aspect ratio).
+	if (zoom == 0) // auto zoom in untill black-bars are gone (while keeping the aspect ratio).
 		zoom = std::max( (float)arr, (float)(1.0/arr) );
 
 	viewport.Scale(zoom, zoom*g_Conf->GSWindow.StretchY.ToFloat()/100.0 );
 	if (viewport == client && EmuConfig.Gamefixes.FMVinSoftwareHack && g_Conf->GSWindow.IsFullscreen)
 		viewport.x += 1; //avoids crash on some systems switching HW><SW in fullscreen aspect ratio's with FMV Software switch.
-	SetSize( viewport );
+	SetSize(viewport);
 	CenterOnParent();
-	
+
 	int cx, cy;
 	GetPosition(&cx, &cy);
 	float unit = .01*(float)std::min(viewport.x, viewport.y);
@@ -585,7 +600,6 @@ GSPanel* GSFrame::GetViewport()
 	return (GSPanel*)FindWindowById( m_id_gspanel );
 }
 
-
 void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
 {
 	// Update the title only after the completion of at least a single Vsync, it's pointless to display the fps
@@ -622,6 +636,27 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
 	std::ostringstream out;
 	out << std::fixed << std::setprecision(2) << fps;
 	OSDmonitor(Color_StrongGreen, "FPS:", out.str());
+
+	AspectRatioType Aspect_Ratio = g_Conf->GSWindow.AspectRatio;
+
+	bool supported_GSWindow_Settings = (Aspect_Ratio == AspectRatio_Frame || (Aspect_Ratio != AspectRatio_Stretch &&
+																				g_Conf->GSWindow.ScalingType != ScalingType_Fit));
+
+	if ( GSgetImageSize != nullptr && supported_GSWindow_Settings )
+	{
+		int new_width = 0, new_height = 0;
+		GSgetImageSize(&new_width, &new_height);
+		if ( new_width > 1 && new_height > 1 && (s_image_width != new_width || s_image_height != new_height) )
+		{
+			if ( GSPanel* gsPanel = GetViewport() )
+			{
+				s_image_width = new_width;
+				s_image_height = new_height;
+
+				gsPanel->DoResize();
+			}
+		}
+	}
 
 #ifdef __linux__
 	// Important Linux note: When the title is set in fullscreen the window is redrawn. Unfortunately
--- a/pcsx2/gui/FrameForGS.cpp
+++ b/pcsx2/gui/FrameForGS.cpp
@@ -180,8 +180,12 @@ void GSPanel::DoResize()
 			targetAr = 16.0 / 9.0;
 		else if (Aspect_Ratio == AspectRatio_Frame)
 			targetAr = acceptedImageSize ? (double)s_image_width / s_image_height : 4.0 / 3.0;
-	} else {
-		targetAr = 4.0 / 3.0;
+	} else if (switchAR) {
+		if (g_Conf->GSWindow.FMVAspectRatioSwitch == FMV_AspectRatio_Switch_16_9 && Aspect_Ratio != AspectRatio_4_3) {
+			targetAr = 16.0 / 9.0;
+		} else {
+			targetAr = 4.0 / 3.0;
+		}
 	}
 
 	double arr = targetAr / clientAr;
--- a/pcsx2/gui/GSFrame.h
+++ b/pcsx2/gui/GSFrame.h
@@ -1,6 +1,5 @@
-
 /*  PCSX2 - PS2 Emulator for PCs
- *  Copyright (C) 2002-2010  PCSX2 Dev Team
+ *  Copyright (C) 2002-2018  PCSX2 Dev Team
  *
  *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
  *  of the GNU Lesser General Public License as published by the Free Software Found-
--- a/pcsx2/gui/GlobalCommands.cpp
+++ b/pcsx2/gui/GlobalCommands.cpp
@@ -1,5 +1,5 @@
 /*  PCSX2 - PS2 Emulator for PCs
- *  Copyright (C) 2002-2010  PCSX2 Dev Team
+ *  Copyright (C) 2002-2018  PCSX2 Dev Team
  *
  *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
  *  of the GNU Lesser General Public License as published by the Free Software Found-
@@ -177,17 +177,39 @@ namespace Implementations
 	{
 		AspectRatioType& art = g_Conf->GSWindow.AspectRatio;
 		const char *arts = "Not modified";
-		if (art == AspectRatio_Stretch && switchAR) //avoids a double 4:3 when coming from FMV aspect ratio switch
+		if (art == AspectRatio_Stretch && switchAR) // avoids a double 4:3 when coming from FMV aspect ratio switch
 			art = AspectRatio_4_3;
 		switch( art )
 		{
-			case AspectRatio_Stretch:	art = AspectRatio_4_3; arts = "4:3"; break;
-			case AspectRatio_4_3:		art = AspectRatio_16_9; arts = "16:9"; break;
-			case AspectRatio_16_9:		art = AspectRatio_Stretch; arts = "Stretch"; break;
+			case AspectRatio_Stretch:	art = AspectRatio_4_3;		arts = "4:3"; break;
+			case AspectRatio_4_3:		art = AspectRatio_16_9;		arts = "16:9"; break;
+			case AspectRatio_16_9:		art = AspectRatio_Frame;	arts = "Frame"; break;
+			case AspectRatio_Frame:		art = AspectRatio_Stretch;	arts = "Stretch"; break;
 			default: break;
 		}
 
 		OSDlog(Color_StrongBlue, true, "(GSwindow) Aspect ratio: %s", arts);
+		UpdateImagePosition();
+	}
+
+	void GSwindow_CycleScalingType()
+	{
+		if (g_Conf->GSWindow.AspectRatio == AspectRatio_Stretch) {
+			OSDlog(Color_StrongBlue, true, "(GSwindow) Scaling Types are locked under \"Stretch\" aspect ratio");
+			return;
+		}
+
+		ScalingTypes& st = g_Conf->GSWindow.ScalingType;
+		const char *sts = "Not modified";
+		switch (st)
+		{
+			case ScalingType_Fit:		st = ScalingType_Integer;	sts = "Integer"; break;
+			case ScalingType_Integer:	st = ScalingType_Centered;	sts = "Centered"; break;
+			case ScalingType_Centered:	st = ScalingType_Fit;		sts = "Fit"; break;
+			default: break;
+		}
+
+		OSDlog(Color_StrongBlue, true, "(GSwindow) Scaling Type: %s", sts);
 		UpdateImagePosition();
 	}
 
@@ -537,6 +559,13 @@ static const GlobalCommandDescriptor CommandDeclarations[] =
 
 	{	"GSwindow_CycleAspectRatio",
 		Implementations::GSwindow_CycleAspectRatio,
+		NULL,
+		NULL,
+		true,
+	},
+
+	{	"GSwindow_CycleScalingType",
+		Implementations::GSwindow_CycleScalingType,
 		NULL,
 		NULL,
 		true,
--- a/pcsx2/gui/Panels/GSWindowPanel.cpp
+++ b/pcsx2/gui/Panels/GSWindowPanel.cpp
@@ -77,6 +77,7 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 
 	m_check_SizeLock	= new pxCheckBox( this, _("Disable window resize border") );
 	m_check_HideMouse	= new pxCheckBox( this, _("Always hide mouse cursor") );
+	m_check_ScalingCompensation = new pxCheckBox(this, _("Disable Scaling Type height compensation"));
 	m_check_CloseGS		= new pxCheckBox( this, _("Hide window when paused") );
 	m_check_Fullscreen	= new pxCheckBox( this, _("Default to fullscreen mode on open") );
 	m_check_DclickFullscreen = new pxCheckBox( this, _("Double-click toggles fullscreen mode") );
@@ -112,6 +113,10 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 	m_check_HideMouse->SetToolTip(pxEt(L"Check this to force the mouse cursor invisible inside the GS window; "
 										L"useful if using the mouse as a primary control device for gaming. "
 										L"By default the mouse auto-hides after 2 seconds of inactivity."));
+
+	m_check_ScalingCompensation->SetToolTip(pxEt(L"Don't let Scaling Types compensate(double) the height for games that put out an image "
+												L"that is half as high as displayed. This will half the height for some games. Only for purists.\n\n"
+												L"NOTE: Only available when \"Frame\" Aspect Ratio or \"Integer\" or \"Centered\" Scaling Type is selected."));
 
 	m_check_Fullscreen->SetToolTip(pxEt(L"Enables automatic mode switch to fullscreen when starting or resuming emulation. "
 										L"You can still toggle fullscreen display at any time using alt-enter."));
@@ -157,6 +162,7 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 	*this += m_check_SizeLock;
 	*this += m_check_HideMouse;
 	*this += m_check_CloseGS;
+	*this += m_check_ScalingCompensation;
 	*this += new wxStaticLine( this )	| StdExpand();
 
 	*this += m_check_Fullscreen;
@@ -198,6 +204,10 @@ void Panels::GSWindowSettingsPanel::ApplyConfigToGui( AppConfig& configToApply, 
 		m_check_SizeLock->SetValue(conf.DisableResizeBorders);
 		m_check_HideMouse->SetValue(conf.AlwaysHideMouse);
 		m_check_CloseGS->SetValue(conf.CloseOnEsc);
+		m_check_ScalingCompensation->SetValue(conf.DisableScalingCompensation);
+		m_check_ScalingCompensation->Enable((conf.AspectRatio != AspectRatio_Stretch &&
+											conf.ScalingType != ScalingType_Fit) ||
+											conf.AspectRatio == AspectRatio_Frame);
 
 		m_check_Fullscreen->SetValue(conf.DefaultToFullscreen);
 		m_check_DclickFullscreen->SetValue(conf.IsToggleFullscreenOnDoubleClick);
@@ -210,6 +220,8 @@ void Panels::GSWindowSettingsPanel::ApplyConfigToGui( AppConfig& configToApply, 
 void Panels::GSWindowSettingsPanel::ScalingTypeChanged(wxCommandEvent &event)
 {
 	m_combo_ScalingType->Enable(m_combo_AspectRatio->GetSelection() != 0);
+	m_check_ScalingCompensation->Enable((m_combo_AspectRatio->GetSelection() != 0 &&
+		m_combo_ScalingType->GetSelection() != 0) || m_combo_AspectRatio->GetSelection() == 3);
 	m_combo_FMVAspectRatioSwitch->Enable(m_combo_AspectRatio->GetSelection() != 1);
 
 	event.Skip();
@@ -228,6 +240,7 @@ void Panels::GSWindowSettingsPanel::Apply()
 	appconf.DisableResizeBorders = m_check_SizeLock->GetValue();
 	appconf.AlwaysHideMouse = m_check_HideMouse->GetValue();
 	appconf.CloseOnEsc = m_check_CloseGS->GetValue();
+	appconf.DisableScalingCompensation = m_check_ScalingCompensation->GetValue();
 
 	appconf.DefaultToFullscreen = m_check_Fullscreen->GetValue();
 	appconf.IsToggleFullscreenOnDoubleClick = m_check_DclickFullscreen->GetValue();
--- a/pcsx2/gui/Panels/GSWindowPanel.cpp
+++ b/pcsx2/gui/Panels/GSWindowPanel.cpp
@@ -33,6 +33,13 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 		_("Frame")
 	};
 
+	const wxString fmv_aspect_ratio_switch_labels[] =
+	{
+		_("Standard (4:3)"),
+		_("Widescreen (16:9)"),
+		_("Off")
+	};
+
 	const wxString scaling_type_labels[] =
 	{
 		_("Fit to Window/Screen"),
@@ -56,6 +63,9 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 	m_combo_ScalingType = new wxComboBox(this, wxID_ANY, wxEmptyString, wxDefaultPosition, wxDefaultSize,
 		ArraySize(scaling_type_labels), scaling_type_labels, wxCB_READONLY);
 
+	m_combo_FMVAspectRatioSwitch = new wxComboBox(this, wxID_ANY, wxEmptyString, wxDefaultPosition, wxDefaultSize,
+		ArraySize(fmv_aspect_ratio_switch_labels), fmv_aspect_ratio_switch_labels, wxCB_READONLY);
+
 	m_combo_vsync = new wxComboBox( this, wxID_ANY, wxEmptyString, wxDefaultPosition, wxDefaultSize,
 		ArraySize(vsync_label), vsync_label, wxCB_READONLY );
 
@@ -70,7 +80,6 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 	m_check_CloseGS		= new pxCheckBox( this, _("Hide window when paused") );
 	m_check_Fullscreen	= new pxCheckBox( this, _("Default to fullscreen mode on open") );
 	m_check_DclickFullscreen = new pxCheckBox( this, _("Double-click toggles fullscreen mode") );
-	m_check_AspectRatioSwitch = new pxCheckBox(this, _("Switch to 4:3 aspect ratio when an FMV plays"));
 	//m_check_ExclusiveFS = new pxCheckBox( this, _("Use exclusive fullscreen mode (if available)") );
 
 	m_combo_AspectRatio->SetToolTip(pxEt(L"Stretch: Stretches the image to the window/screen size.\n\n"
@@ -81,6 +90,11 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 										L"NOTE: Using the widescreen 16:9 aspect ratio will result in a stretched image unless "
 										L"widescreen is enabled in-game or through a widescreen patch. "
 										L"Widescreen is not available for all games."));
+
+	m_combo_FMVAspectRatioSwitch->SetToolTip(pxEt(L"4:3: Temporarily switch back to 4:3 aspect ratio while an FMV plays to correctly display an 4:3 FMV. "
+												L"This option works when the normal aspect ratio is set to anything but 4:3.\n\n"
+												L"16:9: Temporarily switch back to a 16:9 aspect ratio while an FMV plays to correctly display a widescreen 16:9 FMV. "
+												L"This option only works when the normal aspect ratio is set to Frame or Stretch."));
 
 	m_combo_ScalingType->SetToolTip(pxEt(L"Fit: Scales the image width and/or height to the window/screen size depending on the selected aspect ratio.\n\n"
 										L"Integer Scaling: Scales the image to the highest integer magnification of the selected aspect ratio that "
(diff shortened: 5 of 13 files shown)


gregory38:
I'm not sure I love the pixel name.

However, I'm not sure it is a good idea to query the size on the GS plugin. I think you can read the CRTC register on the EE side directly. However it might duplicate some code for the height. I'm not sure what is the best solution.

FlatOutPS2:
> I'm not sure I love the pixel name.

It's called pixel aspect ratio, so I'm not sure what would be better.

> However, I'm not sure it is a good idea to query the size on the GS plugin. I think you can read the CRTC register on the EE side directly. However it might duplicate some code for the height. I'm not sure what is the best solution.

Querying it seemed like the best way to me. Duplicating the code would also mean having to modify the code in both places when improving the CRTC detection.

ssakash:
> However, I'm not sure it is a good idea to query the size on the GS plugin. I think you can read the CRTC register on the EE side directly. 

Please don't confuse people by adding new names for registers. :P 

Though yes, DISP and DISPFB are privileged registers. You can get the values directly from EE and I'd personally prefer if it's done that way to be honest, would remove the unnecessary plugin dependency. It's always ideal to keep most of the code in the core and rely on other modules _only_ when necessary,

> Querying it seemed like the best way to me. Duplicating the code would also mean having to modify the code in both places when improving the CRTC detection.

Unless I'm wrong, your plugin side implementation won't work when using a newer version of PCSX2 along with an older version of GSdx plugin. Implementing the function in core instead would not introduce any such problem.

ssakash:
> However, I'm not sure it is a good idea to query the size on the GS plugin. I think you can read the CRTC register on the EE side directly.

Okay, apologies. I see what you mean now, you mean all the registers which modify the settings of the PCRTC circuit, right? I was just confused thinking that you meant "_one_ single register" as the CRTC register, my bad.

gregory38:
Yes I mean all privileged registers that are used to set the size.

I'm afraid there isn't an ideal solution.  I let @turtleli decide here 👍 

By the way, I'm not sure 1:1 ratio is a good thing. A 512x448 game will be very very small on a  4K display well there is upscaling. Dunno @rz5 do you prefer 1:1 or a multiple that can still fit in the screen (or both option)?

gregory38:
Actually for upscaling you need to query the size on the GS plugin.

MrCK1:
Isn't it possible that this would also break savestates?

ssakash:
> Actually for upscaling you need to query the size on the GS plugin.

Looking at the code (not on my IDE, at Github online editor and I could be potentially wrong), @FlatOutPS2 is only sending the native resolution size to core, not the upscaled resolution size.

willkuer:
I think the original idea of #1916 is to round the scaling to an integer value such that one pixel of the ps2 output covers an integer number of pixels on the final screen. (Like instead of  512x448 you send a twice larger output 1024x896 where each ps2 pixel covers 4 LCD pixels)

FlatOutPS2:
> By the way, I'm not sure 1:1 ratio is a good thing. A 512x448 game will be very very small on a 4K display well there is upscaling. 

You can use zoom to increase the size when using the native size, but the pixel aspect ratio will work like a regular aspect ratio(using the full width or height of the screen) too.

mirh:
By the way.. perhaps it's just i'm not a native speaker, I dunno.. But "aspect ratio" tag for something that is a *point* feels wrong.

turtleli:
I'm not sure whether #1916 is requesting integer scaling or unscaled. Personally I see integer scaling (as willkuer described) to be the more useful option - I don't think unscaled will be nice on high res monitor (as gregory already mentioned).

> I let turtleli decide here

Can I decide to not decide? ;) I think querying GSdx is probably the better option since it avoids code duplication. The other way would be to merge GSdx into PCSX2... XD

pcsx2/gui/FrameForGS.cpp
```diff
@@ -183,13 +185,19 @@ void GSPanel::DoResize()
 	else if( arr > 1 )
 		viewport.y = (int)( (double)viewport.y/arr + 0.5);
 
+	if (g_Conf->GSWindow.ScaleToImageSize && image_width > 0 && image_width < client.GetWidth() &&
+		image_height > 0 && image_height < client.GetHeight() && g_Conf->GSWindow.AspectRatio != AspectRatio_Stretch) {
+			viewport.x = image_width;
+			viewport.y = (int)((double)image_width / targetAr);
```
turtleli: So, I don't get how this is scaled to image size. Wouldn't scaled to image size mean
`viewport.y = image_height;`?

pcsx2/gui/FrameForGS.cpp
```diff
@@ -615,6 +623,22 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
 	out << std::fixed << std::setprecision(2) << fps;
 	OSDmonitor(Color_StrongGreen, "FPS:", out.str());
 
+	int new_width, new_height;
+	GSgetImageSize(&new_width, &new_height);
+	if (new_width > 0 && new_height > 0)
```
turtleli: With an older GSdx version, `new_width` and `new_height` will be uninitialised, so you might have random dimensions.

pcsx2/gui/FrameForGS.cpp
```diff
@@ -615,6 +623,22 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
...
+	{
+		if (GSPanel* gsPanel = GetViewport())
+		{
+			if (gsPanel->image_width != new_width || gsPanel->image_height != new_height)
+			{
+				gsPanel->image_width = (int)new_width;
+				gsPanel->image_height = (int)new_height;
```
turtleli: Are the int casts even necessary?

pcsx2/gui/FrameForGS.cpp
```diff
@@ -183,13 +185,19 @@ void GSPanel::DoResize()
...
+
 	float zoom = g_Conf->GSWindow.Zoom.ToFloat()/100.0;
-	if( zoom == 0 )//auto zoom in untill black-bars are gone (while keeping the aspect ratio).
+	if( zoom == 0 ) // auto zoom in until black bars are gone (while keeping the aspect ratio).
 		zoom = std::max( (float)arr, (float)(1.0/arr) );
 
 	viewport.Scale(zoom, zoom*g_Conf->GSWindow.StretchY.ToFloat()/100.0 );
 	if (viewport == client && EmuConfig.Gamefixes.FMVinSoftwareHack && g_Conf->GSWindow.IsFullscreen)
-		viewport.x += 1; //avoids crash on some systems switching HW><SW in fullscreen aspect ratio's with FMV Software switch.
+		viewport.x += 1; // avoids crash on some systems switching HW><SW in fullscreen aspect ratio's with FMV Software switch.
```
turtleli: The apos'trophe s'eems' unneces's'ary.

Conjecture: some systems -> systems with AMD GPUs on Windows.

pcsx2/gui/GSFrame.h
```diff
@@ -48,6 +48,9 @@ class GSPanel : public wxWindow
 	bool					m_coreRunning;
 
 public:
+	int						image_width;
+	int						image_height;
```
turtleli: Why are these public?

plugins/GSdx/GS.cpp
```diff
@@ -947,6 +947,15 @@ EXPORT_C GSgetTitleInfo2(char* dest, size_t length)
 	strcpy(dest, s.c_str());
 }
 
+EXPORT_C GSgetImageSize(int *image_width = 0, int *image_height = 0)
+{
+	if (gsopen_done && s_gs != NULL)
```
turtleli: `nullptr`

plugins/GSdx/GS.cpp
```diff
@@ -947,6 +947,15 @@ EXPORT_C GSgetTitleInfo2(char* dest, size_t length)
 	strcpy(dest, s.c_str());
 }
 
+EXPORT_C GSgetImageSize(int *image_width = 0, int *image_height = 0)
```
turtleli: Why have default arguments been used? I don't think they're necessary.

FlatOutPS2:
> I'm not sure whether #1916 is requesting integer scaling or unscaled. Personally I see integer scaling (as willkuer described) to be the more useful option - I don't think unscaled will be nice on high res monitor (as gregory already mentioned).

As I said integer scaling is possible with my implementation by using zoom. With integer scaling setting Internal Resolution in GSdx to 2x native (of 512x448) in GSdx will only result in 1024x896. However, with my implementation you can set the IR to 4x native and zoom to 200%. This will also result in 1024x896, but with at a higher resolution.

pcsx2/gui/FrameForGS.cpp
FlatOutPS2: No, as you can change the aspect ratio.

rz5:
> By the way, I'm not sure 1:1 ratio is a good thing. A 512x448 game will be very very small on a 4K display well there is upscaling. Dunno @rz5 do you prefer 1:1 or a multiple that can still fit in the screen (or both option)?

@gregory38 On second thought, yeah, I want both! I want to be able to see the real size, no matter how tiny it may be and I want PCSX2 to use the largest integer multiple resolution that can fit inside the window.

I think that would require two checkboxes to accomplish it. "Stretch to fit" and one underneath it called "Integer scaling"

Right now it's as if PCSX2 always has "Stretch to fit" on and "Integer scaling" off. If I disable "Stretch to fit", I should get the real size, tiny as it is. If "Stretch to fit" is enabled along with "Integer scaling", then the picture should be stretched to the largest integer multiple that can still fit the window.

And I guess down the line you should have a "Keep aspect ratio" checkbox too, so you can disable it and allow a "Stretch to fill" mode where the picture is stretched to fill the whole screen. I wouldn't use it but there are savages out there, I dunno.

pcsx2/gui/FrameForGS.cpp
turtleli: I don't understand. I interpret this:
```
if the frame size is 640x480, then display a centered 640x480 frame
while leaving the rest black.
```
as meaning "avoid stretching the image and applying aspect ratios".

FlatOutPS2:
@rz5 This PR currently offers everything you mention...

rz5:
No offense, but I don't want to play around with the Zoom option, as versatile as it is.

tony971:
Sounds like mednafen's <system>.stretch setting

https://mednafen.github.io/documentation/#Global+Settings+Reference

rz5:
> Sounds like mednafen's .stretch setting
@tony971 Yup, like that and like RetroArch's "Integer scaling" video option. I'm not sure what's the reasoning behind that aspect_mult2 option tho...

tony971:
http://forum.fobby.net/index.php?t=msg&&th=938&goto=3141#msg_3151

> The default stretch of "aspect_mult2"(specifically the mult2 bit) is intended to allow for interlaced video and/or scanline effects to be displayed properly.

plugins/GSdx/GS.cpp
gregory38: Besides I'm not sure they do what we expect. For me it sets them to null pointer (but I'm not 100% sure)

pcsx2/gui/FrameForGS.cpp
mirh: Seems like you are a seer https://github.com/PCSX2/pcsx2/pull/1312#issuecomment-217485166
I wonder if it's still actual though.

gregory38:
So can someone told me the current/expected behavior for those cases. Let's take a 640x480 game on a 1920x1080 display.
* native resolution
* 2x resolution

Note on the name, "centered" could be an alternative instead of pixel.

FlatOutPS2:
> So can someone told me the current/expected behavior for those cases. Let's take a 640x480 game on a 1920x1080 display.

Native -> Scales to native frame size.
Integer scaling -> Scales to largest integer magnification of the frame size that still fits the entire image on screen.

The GSdx internal resolution has no influence on it.

Centered is wrong on all levels, All options relating to the image/screen size are centered.

rz5:
@FlatOutPS2 - I think the description of native is a bit confusing due it saying it actually "scales". There should be no scaling done, period. You present the frame as-is.

If you had a combo-box toggling the picture scaling, here's the choices and behaviors I'd expect:

For a 640x480 game on a 1920x1080 display:
. Native/Don't scale/Original size would be 640x480;
. "Fill"/"Full stretch" would stretch the original pic to 1920x1080;
. "Fit to height" would zoom the original pic to 1440x1080;
. "Integer scale"/"mult2" would scale the pic to 1280x960 (x2 scale)

mirh:
> Centered is wrong on all levels, All options relating to the image/screen size are centered.

It's actually a pretty [well understood](http://pcgamingwiki.com/wiki/Glossary:Graphics_card#GPU_scaling) term tbh. 

To be even more honest, imo, allowing "Fit to Window/Screen" (or more in general to screw aspect ratio) is immoral. I can't see a single situation in which squished images would be preferable (worse, in the past I always got confused on whether 4:3/16:9 from GS windows settings actually made the game render differently). 

Thus, I just see the following cases: 
- no scaling/centered/native
- integer scaling
- float/"free" scaling

FlatOutPS2:
> It's actually a pretty well understood term tbh.

That relates to scaling, not aspect ratio. Centered could be used instead of native/no scaling, but centered isn't a very clear term either.

rz5:
@mirh - Perhaps a good feature to include would be to do away with the manual aspect ratio option and have PCSX2 deduce it from the video mode and/or usage of widescreen hacks.

mirh:
> That relates to scaling, not aspect ratio.

In my ideal world that's not something that is ever touched [after rendering] at all ^^

> manual aspect ratio option and have PCSX2 deduce it from the video mode and/or usage of widescreen hacks.

...
It's funny I always thought there to be some mechanism already in place.. Instead it was just a hardcoded value and what not. 

Speaking of which, unless "free scaling" is enabled.. What's even the point in presenting a window filled in black? I mean, perhaps in fullscreen if any, but not elsewhere.

FlatOutPS2:
@mirh - Kind of off topic but perhaps a good feature to include would be to do away with the manual aspect ratio option and have PCSX2 deduce it from the video mode and/or usage of widescreen hacks.

That's not possible as neither will tell you the correct aspect ratio.

rz5:
@FlatOutPS2 - This is your PR thread and I am again taking this off-topic, feel free to say 'stop'.

> That's not possible as neither will tell you the correct aspect ratio.

Consider the following:
0) Users can only pick 2 aspect ratios, 4:3 and 16:9;
1) By default, PCSX2 sets Emulation Settings > GS Window > Aspect Ratio to Standard (4:3);
2) It must be manually changed to Widescreen (16:9) when playing several games with widescreen hacks or a few games with HDTV mode;
3) https://github.com/PCSX2/pcsx2/blob/cca6e1dcae33f8d9956767190bba3ea2436f7c03/pcsx2/R5900OpcodeImpl.cpp#L900#L941 - At a glance, we can tell which modes are definitely 16:9 and which are 4:3;
4) If the user has "Enable Cheats" and "Enable Widescreen Patches" on, if one of the widescreen patches match the current game, odds are the user wants the AR set to 16:9;

Based on all of the above, would you not say it's possible to programatically set the "correct" AR? I put it in air quotes because I did some googling on the AR of 640x448 and it led me to some obscure wiki on [dot clock rates](https://pineight.com/mw/index.php?title=Dot_clock_rates)...but PCSX2 uses 4:3 on it and it looks inoffensive to my eyes.

FlatOutPS2:
0. There are more aspect ratio's than just 4:3 and 16:9(set by using fill/stretch)
3. It doesn't work like that. Video modes (resolutions) do not equal a game's aspect ratio. Changing aspect ratio doesn't change video modes either. ;)
4. Widescreen patches can be made for others aspect ratio's too. And some widescreen patches only affect certain parts of a game, like a 2D game that has a widescreen patch which only applies to battle scenes which are in 3D. you can't accurately check when such a scene is active. And there's the games with in-game widescreen options...

rz5:
@FlatOutPS2 - I'll open a new issue to continue discussing this topic.

FlatOutPS2:
@rz5 There is no discussion. It just won't work.

mirh:
> Video modes (resolutions) do not equal a game's aspect ratio.

Wtf? Why?
Even if part of the output of a game was letterboxed.. Wouldn't that still be displayed even in our current implementation?

FlatOutPS2:
The video mode can be `NTSC 640x448` (aspect ratio 10:7)
The frame size can be 512x448 (aspect ratio 8:7)
But the aspect ratio to correctly show the contents can be 4:3, or 16:9 (or anything else, although the PS2 was obviously aimed towards showing an unstretched image on 4:3 and widescreen 16:9 TV's).

rz5:
@FlatOutPS2 - I just built the ui branch of your fork. I'm testing it now. Here's my observations:
Here's my setup:
My monitor's resolution is 1366x768 and I am testing Disc 1 of MGS3:Subsistence (US version). I am in-game and am looking at the pause menu;

When PCSX2 is windowed:
. Integer scaling seems to kind of work. The picture is no longer stretched to fit so that is different from the default behavior. https://i.snag.gy/Mj75Wx.jpg
It seems that the way you calculate the max integer multiple resolution is working - the reported frame size is 512x448, doing a x2 scale would definitely exceed my monitor's resolution.

When PCSX2 is fullscreen:
. Integer scaling doesn't seem to be working. The old behavior of "Fit to height" is in effect.
https://i.snag.gy/aPL6Yw.jpg

rz5:
Also, about the "Frame" AR option - it's a nice curiosity to have I guess, since you get to see the picture in the weird ass resolutions that the PS2 internally uses before it gets stretched to the usual 4:3 aspect ratio.

Since I don't see a point of having that feature exposed to the user, I suggest you keep it off release builds. I dunno, maybe it's useful for debugging so keep it behind a debug ifdef?

pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -131,7 +150,9 @@ void Panels::GSWindowSettingsPanel::ApplyConfigToGui( AppConfig& configToApply,
 		m_check_SizeLock	->SetValue( conf.DisableResizeBorders );
 
 		m_combo_AspectRatio	->SetSelection( (int)conf.AspectRatio );
+		m_combo_AspectRatio	->Enable( conf.ScalingType != ScalingType_Fill );
```
ssakash: Missing [bind event handler](http://docs.wxwidgets.org/3.1.0/classwx_evt_handler.html#a0f30c8fa5583b4a5f661897d63de3b62) for the Aspect ratio combobox grayout?

pcsx2/gui/FrameForGS.cpp
```diff
@@ -615,6 +629,22 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
 	out << std::fixed << std::setprecision(2) << fps;
 	OSDmonitor(Color_StrongGreen, "FPS:", out.str());
 
+	int new_width, new_height = 0;
```
ssakash: Why is ``new_width`` uninitialized here?

pcsx2/gui/GlobalCommands.cpp
```diff
@@ -174,22 +174,46 @@ namespace Implementations
...
+		const char *sts = "Not modified";
+		//if (art == AspectRatio_Frame && switchAR) //avoids a double 4:3 when coming from FMV aspect ratio switch
+		//	art = AspectRatio_4_3;
+		switch (st)
+		{
+			case ScalingType_Fill:		st = ScalingType_Fit;		sts = "Fit"; break;
+			case ScalingType_Fit:		st = ScalingType_Integer;	sts = "Integer"; break;
+			case ScalingType_Integer:	st = ScalingType_Centered;	sts = "Centered"; break;
+			case ScalingType_Centered:	st = ScalingType_Fill;		sts = "Fill"; break;
+			default: break;
```
ssakash: An assert would be nice here.

pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -49,16 +60,22 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
...
+	m_check_AspectRatioSwitch = new pxCheckBox( this, _("Switch to 4:3 aspect ratio when an FMV plays") );
 	//m_check_ExclusiveFS = new pxCheckBox( this, _("Use exclusive fullscreen mode (if available)") );
 
-	m_text_Zoom->SetToolTip( pxEt( L"Zoom = 100: Fit the entire image to the window without any cropping.\nAbove/Below 100: Zoom In/Out\n0: Automatic-Zoom-In untill the black-bars are gone (Aspect ratio is kept, some of the image goes out of screen).\n  NOTE: Some games draw their own black-bars, which will not be removed with '0'.\n\nKeyboard: CTRL + NUMPAD-PLUS: Zoom-In, CTRL + NUMPAD-MINUS: Zoom-Out, CTRL + NUMPAD-*: Toggle 100/0"
+	m_combo_AspectRatio->SetToolTip(pxEt(L"4:3: Keeps the aspect ratio to 4:3, the default ratio used by the PS2.\n16:9: Sets the aspect ratio to widescreen 16:9, the ratio used by games when a widescreen option or normal widescreen patch is enabled.\nFrame:Uses the aspect ratio of the frame size of the game, to keep a 1:1 pixel ratio.\n\nNOTE: Using the widescreen 16:9 aspect ratio will result in a stretched image unless widescreen is enabled in-game or through a widescreen patch. Widescreen is not available for all games."
```
ssakash: Some space after ``Frame:`` please.

gregory38:
I'm fine with your new improvements.

pcsx2/gui/GlobalCommands.cpp
gregory38: there is jno_default or something close of it macro.

pcsx2/PluginManager.cpp
```diff
@@ -420,6 +422,7 @@ static const LegacyApi_ReqMethod s_MethMessReq_GS[] =
 	{	"GSsetExclusive",	(vMeth**)&GSsetExclusive,	(vMeth*)GS_setExclusive	},
 	{	"GSchangeSaveState",(vMeth**)&GSchangeSaveState,(vMeth*)GS_changeSaveState },
 	{	"GSgetTitleInfo2",	(vMeth**)&GSgetTitleInfo2,	(vMeth*)GS_getTitleInfo2 },
+	{	"GSgetImageSize",	(vMeth**)&GSgetImageSize,	(vMeth*)GS_getImageSize },
```
gregory38: It should be moved to `s_MethMessOpt_GS` instead. And a check must be done if the pointer is nullptr before using it.

pcsx2/gui/FrameForGS.cpp
```diff
@@ -615,6 +629,22 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
...
+	GSgetImageSize(&new_width, &new_height);
```
gregory38: Make `GSgetImageSize` optional and check null ptr here.

pcsx2/gui/GlobalCommands.cpp
```diff
@@ -174,22 +174,46 @@ namespace Implementations
...
 
 		OSDlog(Color_StrongBlue, true, "(GSwindow) Aspect ratio: %s", arts);
 		UpdateImagePosition();
 	}
 
+	void GSwindow_CycleScalingType()
+	{
+		ScalingTypes& st = g_Conf->GSWindow.ScalingType;
+		const char *sts = "Not modified";
+		//if (art == AspectRatio_Frame && switchAR) //avoids a double 4:3 when coming from FMV aspect ratio switch
```
gregory38: delete left-overs

pcsx2/gui/FrameForGS.cpp
```diff
@@ -183,14 +190,21 @@ void GSPanel::DoResize()
 	else if( arr > 1 )
 		viewport.y = (int)( (double)viewport.y/arr + 0.5);
 
+	if (g_Conf->GSWindow.ScalingType != ScalingType_Fit && g_Conf->GSWindow.ScalingType != ScalingType_Fill && acceptedImageSize) {
```
gregory38: I think it would be easier to avoid the negation
`..ST ==ScalingType_Integer || ..ST == ScalingType_Centered`

gregory38:
```
When PCSX2 is fullscreen:
. Integer scaling doesn't seem to be working. The old behavior of "Fit to height" is in effect.
```
It is strange as fullscreen is just a big border-less window.

James-F2:
Why is it still not implemented??
PCSX2 looks a mess with its current scaling.

MrCK1:
The person who was working on it has disappeared unfortunately. It wouldn't be right for us to modify/merge his work without him.

James-F2:
Are you serious? This is crucial that you sort it out.

Integer Scaling is very important in emulation for proper true to source look.
Simply stretching with bilinear filtering can fix that,, but by the look of it PCSX2 doesn't do even that.
No matter what I do, the image doesn't look 'coherent'...

Look at this X from God of War for example:
Unfiltered: https://s17.postimg.org/p41bqqq9r/X_Go_W.png
Filtered: https://s17.postimg.org/4jwhshfq7/X_Go_W_filtered.png
It appears that PCSX2 doesn't use bilinear filtering properly... very peculiar.

I think you would do the community a great service implementing this code (or similar), as said before, PCSX2 painfully lacking proper Integer Scaling, or a functional non-integer stretching with bilinear filtering for that matter.

mirh:
I'd be surprised if git hadn't a command to cherry-pick, rebase and whatever, while still retaining authorship. 
Of course somebody has to have the time to tackle on the problem then.

James-F2:
It is in the "Pull Request" section after all, so the author FlatOutPS2 theoretically already gave you the permission to implement this.
FlatOutPS2 could have given up on this because the long pull time, or simply more urgent business like life itself.
In any case, PCSX2 visuals suffer greatly without this, or similar.

MrCK1:
@James-F2 Are you also aware that you can download the preliminary build through the Appveyor link below?

James-F2:
No, I'm not aware of that. 
Where is that link?

MrCK1:
https://ci.appveyor.com/api/buildjobs/dqx2fti06oijuy7y/artifacts/pcsx2-v1.5.0-dev-2050-g111089ebb-2969-vs2015-Win32-AppVeyor.7z

James-F2:
Thank you!

I have tried it and unfortunately was disappointing.
For some odd reason the integer scaling is attached to the horizontal plane and not the vertical as it should.
Maybe some day someone will implement proper integer scaling like in Retroarch, but until then I'll get by with what PCSX2 has to offer now.

Thanks.

mirh:
Little reminder that our roads departed in the middle of a [discussion](https://github.com/PCSX2/pcsx2/issues/1923#issuecomment-298395079) (don't want to clutter that thread, where this discussion if any would be OT)
@refractionpcsx2 if you wanted to add something, I dunno. 

Best wishes for your return.

James-F2:
FlatOutPS2, are you really back?
Hopefully the integer scaling patch will be pulled to the main trunk soon, PCSX2 still lacks a good integer scaling.

FlatOutPS2:
@James-F2 There was a mistake in the old code. Did you try the new updated one?

James-F2:
Now indeed it works with the vertical scale, so this is fixed.

Base game resolution 512x448, resulting 4:3 windows:
298x224
596x448
894x671
1192x896
1490x1120

I understand the base vertical resolution is 224 (one frame) half resolution, and not 448 two frames full resolution.
This creates uneven scanlines of the full 448 resolution, in 671 (x1,5) and 1120 (x2.5) modes, but correct integer in 448 (x1) and 896 (x2).

When using Aspect ratio: Frame, it calculates based on half the resolution 224.
This resulting in a very uneven scanlines in any integer scaling x factor.

May I suggest using the full resolution (two frame) 448 as base for the integer scaling, because half resolution creates uneven scalines which is as bad as no integer scaling at all.
This means that only 1192x896 (x2) will be display properly on a 1080p screen.

Great to see you back FlatOutPS2.

pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -56,22 +77,52 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
...
 	m_check_DclickFullscreen = new pxCheckBox( this, _("Double-click toggles fullscreen mode") );
-	m_check_AspectRatioSwitch = new pxCheckBox(this, _("Switch to 4:3 aspect ratio when an FMV plays"));
 	//m_check_ExclusiveFS = new pxCheckBox( this, _("Use exclusive fullscreen mode (if available)") );
 
-	m_text_Zoom->SetToolTip( pxEt( L"Zoom = 100: Fit the entire image to the window without any cropping.\nAbove/Below 100: Zoom In/Out\n0: Automatic-Zoom-In untill the black-bars are gone (Aspect ratio is kept, some of the image goes out of screen).\n  NOTE: Some games draw their own black-bars, which will not be removed with '0'.\n\nKeyboard: CTRL + NUMPAD-PLUS: Zoom-In, CTRL + NUMPAD-MINUS: Zoom-Out, CTRL + NUMPAD-*: Toggle 100/0"
+	m_combo_AspectRatio->SetToolTip(pxEt(L"4:3: Keeps the aspect ratio to 4:3, the default aspect ratio used by the PS2.\n"
+		L"16:9: Sets the aspect ratio to widescreen 16:9, the ratio used by games when a widescreen option or a normal widescreen patch is enabled.\n"
+		L"Frame: Uses the aspect ratio of the frame size of the game, to keep a 1:1 pixel ratio.\n"
+		L"Stretch: Stretches the image to the window / screen size.\n\n"
```
lightningterror: **Stretch:** tooltip should be first.

lightningterror:
I think it's a better to separate the tooltips with an empty row for each option that explains what it does. Would look much more tidy and easier to notice.

James-F2:
Tested the AppVeyor build. 
The "Disable Interlacing Height Compensation" checkbox works perfectly.
Fantastic job, hope to see it in PCSX2 nightly soon.

pcsx2/gui/FrameForGS.cpp
```diff
@@ -642,21 +641,19 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
 	out << std::fixed << std::setprecision(2) << fps;
 	OSDmonitor(Color_StrongGreen, "FPS:", out.str());
 
-	int new_width = 0, new_height = 0;
+	const u64& smode2 = *(u64*)PS2GS_BASE(GS_SMODE2);
+	int new_width = 0, new_height = 0, interlaced = !g_Conf->GSWindow.DisableScalingInterlaceCompensation && (smode2 & 2) && (smode2 & 2);
```
ssakash: I think ``interlaced`` variable should be a bool.

```
(smode2 & 2) && (smode2 & 2)
```

I guess this is some sort of mistake? I'd prefer if you could just get the INT/FFMD bit-field instead of doing logical operations on the whole register, definitely helps in the aspect of code readability.

pcsx2/gui/FrameForGS.cpp
```diff
@@ -642,21 +641,19 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
...
-		if (new_width > 0 && new_height > 0)
+		if (new_width > 0 && new_height > 0 && (image_width != new_width || (image_height != new_height && !interlaced) || (image_height == new_height && interlaced)))
 		{
 			if (GSPanel* gsPanel = GetViewport())
 			{
-				if (image_width != new_width || image_height != new_height)
-				{
 					image_width = new_width;
-					image_height = new_height;
+					image_height = interlaced ? new_height * 2 : new_height;
```
ssakash: ``new_height * ( 1 + interlaced)`` could have been nice, but you can't do it if you change it to bool though.

pcsx2/gui/FrameForGS.cpp
FlatOutPS2: Yes, it could be a bool. I used it differently at first and forgot to change it. The mistake is likely a result of the same change in use.

The smode2 code was already there. I assumed it was the easiest way to get the INT/FFMD data, defenitely the easiest way for me. 😛

pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -140,16 +196,21 @@ void Panels::GSWindowSettingsPanel::ApplyConfigToGui( AppConfig& configToApply,
...
+		m_check_ScalingInterlaced		->SetValue( conf.DisableScalingInterlaceCompensation);
+		m_check_ScalingInterlaced		->Enable( (conf.AspectRatio != AspectRatio_Stretch && conf.ScalingType != ScalingType_Fit) || conf.AspectRatio == AspectRatio_Frame );
+		m_check_SizeLock				->SetValue( conf.DisableResizeBorders );
+
+		m_combo_AspectRatio				->SetSelection( (int)conf.AspectRatio );
+		m_text_Zoom						->ChangeValue( conf.Zoom.ToString() );
+		m_combo_ScalingType				->SetSelection( (int)conf.ScalingType );
+		m_combo_ScalingType				->Enable( conf.AspectRatio != AspectRatio_Stretch );
+		m_combo_FMVAspectRatioSwitch	->SetSelection( (int)conf.FMVAspectRatioSwitch );
+		m_combo_FMVAspectRatioSwitch	->Enable( conf.AspectRatio != AspectRatio_4_3 );
```
ssakash: Doing all the grayouts only on applying the settings isn't enough, you also need to bind them to event handlers to make the grayouts immediate on customization of the combo box.

pcsx2/gui/FrameForGS.cpp
FlatOutPS2: > but you can't do it if you change it to bool though.

Yes, you can. :)

It could also have been reduced further to image_height *= 1 + interlaced;

pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -140,16 +196,21 @@ void Panels::GSWindowSettingsPanel::ApplyConfigToGui( AppConfig& configToApply,
...
-		m_combo_AspectRatio	->SetSelection( (int)conf.AspectRatio );
-		m_text_Zoom			->ChangeValue( conf.Zoom.ToString() );
+		m_check_CloseGS					->SetValue( conf.CloseOnEsc );
+		m_check_Fullscreen				->SetValue( conf.DefaultToFullscreen );
+		m_check_HideMouse				->SetValue( conf.AlwaysHideMouse );
+		m_check_ScalingInterlaced		->SetValue( conf.DisableScalingInterlaceCompensation);
+		m_check_ScalingInterlaced		->Enable( (conf.AspectRatio != AspectRatio_Stretch && conf.ScalingType != ScalingType_Fit) || conf.AspectRatio == AspectRatio_Frame );
+		m_check_SizeLock				->SetValue( conf.DisableResizeBorders );
+
+		m_combo_AspectRatio				->SetSelection( (int)conf.AspectRatio );
```
ssakash: I'd prefer if you could use the generic ``enum_cast()`` function instead of all these manual casts on the enumeration values. This might also help avoid any potential unsigned/signed integer cast conflicts with enumerations.

pcsx2/gui/FrameForGS.cpp
ssakash: > The smode2 code was already there. I assumed it was the easiest way to get the INT/FFMD data, definitely the easiest way for me.

Easy for you, but not everyone knows the position of INT and FFMD fields in SMODE2. 😛 

Bonus points for also changing the older code into proper bitfield name of the register. I hate codes like that without any proper comments and doing bitwise operations on the whole register, have to look at the documents all the time to know which field they're messing up with. -_-

pcsx2/gui/FrameForGS.cpp
ssakash: Cool, didn't know about it. I thought it might require an integer cast, but casting a boolean to integer might look awkward. (Not even sure if upcasting like that is even possible)

Or is the bool just taken as an integer without any sort of cast?

FlatOutPS2: > Or is the bool just taken as an integer without any sort of cast?

Yes, as far as I know it still is. Though I agree that it can look awkward.

James-F2:
@MrCK1 
I don't think the "missing lines problem" has anything to do with Scaling or deinterlacing for that matter.
The problem lies before both of them, since they already receive an image with missing lines.

FlatOutPS2:
Testing if git messed anything up. It did mess with my memory card PR yesterday.

pcsx2/gui/FrameForGS.cpp
```diff
@@ -623,6 +641,28 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
 	out << std::fixed << std::setprecision(2) << fps;
 	OSDmonitor(Color_StrongGreen, "FPS:", out.str());
 
+	const u64 &smode2 = *(u64 *)PS2GS_BASE(GS_SMODE2);
+	int interlacing = (smode2 & 1) ? 1 : 0;      // INT: 0 = Progressive, 1 = Interlaced
+	int interlacing_mode = (smode2 & 2) ? 1 : 0; // FFMD: 0 = Field mode, 1 = Frame mode
```
ssakash: The ternary operator ``? 1 : 0`` is redundant here, coverity will complain about it.

FlatOutPS2: I only added it as a placeholder because VS was complaining about it(don't know why, building it now without ? 1 : 0 VS no longer mentions any issue :/). I pushed it because git had messed with my other PR and I wanted to know everything was fine with this one.

pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -136,27 +189,46 @@ void Panels::GSWindowSettingsPanel::AppStatusEvent_OnSettingsApplied()
...
+		m_check_ScalingCompensation		->SetValue(conf.DisableScalingCompensation);
+		m_check_ScalingCompensation		->Enable((conf.AspectRatio != AspectRatio_Stretch &&
+			conf.ScalingType != ScalingType_Fit) || conf.AspectRatio == AspectRatio_Frame);
 
-		m_text_WindowWidth	->ChangeValue( wxsFormat( L"%d", conf.WindowSize.GetWidth() ) );
-		m_text_WindowHeight	->ChangeValue( wxsFormat( L"%d", conf.WindowSize.GetHeight() ) );
+		m_check_DclickFullscreen 		->SetValue(conf.IsToggleFullscreenOnDoubleClick);
+
+		m_text_WindowWidth				->ChangeValue(wxsFormat(L"%d", conf.WindowSize.GetWidth()));
+		m_text_WindowHeight				->ChangeValue(wxsFormat(L"%d", conf.WindowSize.GetHeight()));
```
turtleli: IMO all that whitespace in between hurts readability.

pcsx2/gui/FrameForGS.cpp
```diff
@@ -623,6 +641,28 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
...
+	int interlacing_mode = (smode2 & 2); // FFMD: 0 = Field mode, 1 = Frame mode
+
+	int new_width = 0, new_height = 0;
+	bool framemode_compensation = !g_Conf->GSWindow.DisableScalingCompensation && interlacing_mode == 1;
+	AspectRatioType Aspect_Ratio = g_Conf->GSWindow.AspectRatio;
+	if (GSgetImageSize != nullptr && (Aspect_Ratio == AspectRatio_Frame || (Aspect_Ratio != AspectRatio_Stretch &&
+																			g_Conf->GSWindow.ScalingType != ScalingType_Fit))) {
+		GSgetImageSize(&new_width, &new_height);
+		if (new_width > 1 && new_height > 1 && (image_width != new_width || (image_height != new_height && !framemode_compensation) ||
+												(image_height == new_height && framemode_compensation)))
```
turtleli: As a note, I think `(image_height != new_height) != framemode_comparison` would also work.

pcsx2/gui/FrameForGS.cpp
```diff
@@ -29,6 +29,9 @@
 #include <sstream>
 #include <iomanip>
 
+int	image_width;
+int	image_height;
```
turtleli: Probably should make this static and use the `s_` prefix.

I'd prefer a space instead of a tab too.

pcsx2/PluginManager.cpp
```diff
@@ -171,6 +171,7 @@ _GSinitReadFIFO    GSinitReadFIFO;
 _GSinitReadFIFO2   GSinitReadFIFO2;
 _GSchangeSaveState GSchangeSaveState;
 _GSgetTitleInfo2   GSgetTitleInfo2;
+_GSgetImageSize    GSgetImageSize;
 _GSmakeSnapshot	   GSmakeSnapshot;
```
lightningterror: Just a small ocd nitpick I saw in the file. Some of them have weird spacing. Do you wanna update them as well since they kinda look out of place since ?

You don't have to do it if you don't want to :stuck_out_tongue:

pcsx2/gui/Panels/GSWindowPanel.cpp
lightningterror: The entire file could actually use an update. The code is a complete mess.

FlatOutPS2: > The entire file could actually use an update. The code is a complete mess.

Pretty much all files in the PCSX2 core are a mess. ;)

lightningterror: Ye but this is special, messy on a new level xD

FlatOutPS2: Not really. This is quite average. What we really need is to clang-format the entire emulator code.

lightningterror: Are there any side effects doing that ?

FlatOutPS2: It would require an extra click when looking at the code blame history.
And it would require all future commiters to clang-format (pretty much) all the files they change, or the automatic (travis) build will fail.

pcsx2/gui/FrameForGS.cpp
```diff
@@ -623,6 +637,26 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
...
+	int new_width = 0, new_height = 0;
+	AspectRatioType Aspect_Ratio = g_Conf->GSWindow.AspectRatio;
+	if (GSgetImageSize != nullptr && (Aspect_Ratio == AspectRatio_Frame || (Aspect_Ratio != AspectRatio_Stretch &&
+																			g_Conf->GSWindow.ScalingType != ScalingType_Fit))) {
+		GSgetImageSize(&new_width, &new_height);
+		if (new_width > 0 && new_height > 0)
+		{
+			if (GSPanel* gsPanel = GetViewport())
+			{
+				if (image_width != new_width || image_height != new_height)
```
turtleli: This commit doesn't compile. Should be `s_image_width`, `s_image_height`.

FlatOutPS2: Fixed.

pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -136,46 +189,63 @@ void Panels::GSWindowSettingsPanel::AppStatusEvent_OnSettingsApplied()
...
+	appconf.FMVAspectRatioSwitch = (FMVAspectRatioSwitchType)m_combo_FMVAspectRatioSwitch->GetSelection();
+	appconf.Zoom = Fixed100::FromString(m_text_Zoom->GetValue());
 
-	appconf.AspectRatio		= (AspectRatioType)m_combo_AspectRatio->GetSelection();
-	appconf.Zoom			= Fixed100::FromString( m_text_Zoom->GetValue() );
-
-	gsconf.VsyncEnable = static_cast<VsyncMode>(m_combo_vsync->GetSelection());
+	appconf.DisableResizeBorders = m_check_SizeLock	->GetValue();
+	appconf.AlwaysHideMouse = m_check_HideMouse	->GetValue();
+	appconf.CloseOnEsc = m_check_CloseGS	->GetValue();
```
turtleli: Trailing whitespace. Also, you should probably remove the whitespace before `->` on these lines.

turtleli:
Also, first commit still doesn't compile and needs fixed.

James-F2:
The last change broke the correct behavior, it (again) treats the horizontal plane as vertical....
Also, the Frame option is not working, it remains 4:3.

FlatOutPS2:
> The last change broke the correct behavior, it (again) treats the horizontal plane as vertical....
Also, the Frame option is not working, it remains 4:3.

Did you copy the PCSX2.exe to an older(master) PCSX2 version? If so it won't work because you're using an out of date GSdx.

James-F2:
Seems the problem is only in SSX3.
It scales to 640x512 or 1280x1024 from 512x448... hmm.

ssx3

FlatOutPS2:
You seem to be using an out of date GSdx version. An up to date version would report the videomode (region) on the game window title bar.

ssx3

James-F2:
pardon my ignorance, but I use the one that comes in AppVeyor or the latest SVN, they all report the framebuffer resolution.
Any other place to get the latest GSdx?

lightningterror:
> You seem to be using an out of date GSdx version. An up to date version would report the videomode (region) on the game window title bar.

Not if his **PCSX2_ui.ini** is outdated.

@James-F2 delete the PCSX2_ui.ini and try again.

FlatOutPS2:
> Not if his PCSX2_ui.ini is outdated.

Yeah, that's probably it. The GSdx version doesn't matter for the region display. :p
Still, there seems to be more wrong than just the PCSX2_ui.ini. It's probably multiple things copied from an old version.

James-F2:
I completely removed the folder and extracted the newest AppVeoyr build.
Same result, 640x512 from 512x448.
I use the SSE2/AVX GSdx.

ssx3 2

FlatOutPS2:
This is what I get with the same GSWindow settings:
ssx3

James-F2:
Found the problem.
Using "Boot ISO (full)" will result in an incorrect detection, as I reported.

FlatOutPS2:
@James-F2 Try again. It should work now.

James-F2:
Yep, fixed. 
Thank you.

turtleli:
Hmm. Question. Should you be doubling by default on Interlaced Framed games? Basically I tried Gran Turismo 4, PAL, SCES 51719, Centered, Framed. It looks like:
gt4

FlatOutPS2:
> Question. Should you be doubling by default on Interlaced Framed games

Yes. This was a side effect of using a different function that didn't support it. Everything should work as expected now.

plugins/GSdx/GS.cpp
```diff
@@ -960,6 +961,23 @@ EXPORT_C GSgetTitleInfo2(char* dest, size_t length)
...
+	if (!gsopen_done || s_gs == nullptr)
+		return;
+
+	int en0 = s_gs->IsEnabled(0);
+	int en1 = s_gs->IsEnabled(1);
+	if (!en0 && !en1)
+		return;
+
+	// Disregard the disabled circuit. With full boot the BIOS will fill the disabled circuit with its own data.
+	int i = en0 ? 0 : en1 && !en0 ? 1 : -1;
```
turtleli: I think this might cause issues with Time Crisis 2/3 (2 player mode, side by side half width frames IIRC)? Might be better to work with GetInternalResolution() (or whatever it was) that you used in your previous revision.

Side note: `&& !en0` is unnecessary, it's already on the en0 == false path..

plugins/GSdx/GS.cpp
```diff
@@ -960,6 +961,23 @@ EXPORT_C GSgetTitleInfo2(char* dest, size_t length)
...
+	if (!gsopen_done || s_gs == nullptr)
+		return;
+
+	int en0 = s_gs->IsEnabled(0);
+	int en1 = s_gs->IsEnabled(1);
+	if (!en0 && !en1)
+		return;
+
+	// Disregard the disabled circuit. With full boot the BIOS will fill the disabled circuit with its own data.
+	int i = en0 && !en1 ? 0 : en1 && !en0 ? 1 : -1;
```
ssakash: This could be simplified into ``int i = (en0 ^ en1) ? en1 : -1;``. Current one might require a bit more effort to be put to understand due to the redundancy. (A rather negligible amount really)

A comment indicating '-1` is for merged contexts might also be nice.

lightningterror:
What's the status on this, is it ready for a merge ?

FlatOutPS2:
> What's the status on this, is it ready for a merge ?

Yeah, I updated with the suggestion by @ssakash. It should be ready.

James-F2:
May I suggest an important feature before it gets merged?

This is not related to integer scaling but to pixel accurate blur-free 4:3 scaling.
512x480 is quite low resolution and it's not 4:3, so when it gets upscaled to 4:3 (on a 1920x1080 screen) with bilinear filtering the result is quite blurry.
I often use in other emulators a simple Scale x3,4,5 filter BEFORE bilinear filtering, thus the filtering is doing down-scaling and the resulting image is much sharper than upscaling as the interpolated pixels only where they need to be.
This method is called "super sampling" which is simple yet very effective for uneven scaling with the sharpest result, yet very close to being pixel perfect.

@FlatOutPS2, 
Do you think a scaling option like this is relevant to PCSX2?

FlatOutPS2:
>  Do you think a scaling option like this is relevant to PCSX2?

It's relevant to PCSX2, but it's out of scope for this PR. ;) Did you try disabling "Texture Filtering of Display" (in the GSdx Shader Configuration window) to reduce the blur?

James-F2:
Understood, I though you might kill 2 birds with one blow since you already worked on more than a few scaling features on this PR.
Yes I am aware of the "texture filtering" feature and I was talking exactly about that.
It is important to have some kind of Interpolation when doing non integer scaling because without it (nearest neighbor) the image is distorted especially with low resolutions like 320x240 or 640x480.
I hoped to reduce some of that interpolation blur of the filtering by upscaling the image prior to it.

FlatOutPS2:
> Understood, I though you might kill 2 birds with one blow since you already worked on more than a few scaling features on this PR.

Yes, but if I keep adding features it will never get merged. ;)

I'll have a look at this later, but there are a few other projects that I need to finish first.

lightningterror:
> Yes, but if I keep adding features it will never get merged. ;)

I agree, it's been delayed for almost a year now, other features can be added in separate PRs, let's not delay this any longer and merge it :stuck_out_tongue:

MrCK1:
That's nowhere near as bad as the "endless mountain of code" in #1699 😛

mirh:
Nah, I'd say macos PR is even more exploded. 
Anyway, why you don't just commit this
And @James-F2 create an issue to further discuss this?

turtleli:
I still haven't really reviewed the strings. I'll try and review tomorrow.

lightningterror:
@FlatOutPS2 status on this ? It was reviewed 2 weeks ago.

lightningterror:
The **16:9 FMV aspect ratio switch** commit has been merged to master.

lightningterror:
I'll close this in favor of https://github.com/PCSX2/pcsx2/tree/flatout_aspectratio
Someone needs to finish the code.

PR #1922, opened 2017-04-30
gsdx: dump gsdump in xz format directly

gregory38:
Reduce disk space. Easy to share.

@turtleli @FlatOutPS2 
It would be nice to port this code to Windows. (and the readback too)
libzma code was taken from https://git.tukaani.org/?p=xz.git


turtleli:
Hmm. Wouldn't it takes ages to compress long multi-frame GS dumps?

Personally I think it's fine to just compress afterwards in xz format so multiple threads can be used and compression can be maximised. I think the GS dump instructions could be updated so it mentions compressing the dump in xz format - I don't think most users actually know that that's what is preferred.

gregory38:
Well I didn't tested the speed on long dump. Most of the dumps are short. At least I'm in favour to automatically compress the short dump. 

I'm not sure that  MT is a good idea. It will badly impact the compression ratio.

I will factorize the code to keep both compressed and not compressed

turtleli:
> Well I didn't tested the speed on long dump. Most of the dumps are short. At least I'm in favour to automatically compress the short dump.

Well I think the main issue with compressing a long dump with the current implementation is if the size passes 200MB - emulation will freeze and the user will have to keep holding shift for a while waiting for it to unfreeze if they still need more frames for whatever reason.

> I'm not sure that MT is a good idea. It will badly impact the compression ratio.

Oh you're right, it does seem to affect it quite a bit.

gregory38:
Yes I know. Maybe we can push it further. 500MB/1000MB. I don't think we need more. Besides it would just crash when we read back the dump.(well I think we can use a 64 bits gsdx now)

But honestly I would be fine to keep compression for only short dump.

Data on the dump is quite repetitive. So yes having a single dictionary is way better. Hum I never tested but maybe ratio can remain high enough with lower compression level.

gregory38:
In order to have some numbers, I did a quick test on my biggest dump (1.2 GB).
```
xz -4 -k GT3_R11_Slowdown_2.gs --stdout > gt.4.xz  172.06s user 0.32s system 99% cpu 2:52.38 total
275M May  7 13:20 gt.4.xz
xz -5 -k GT3_R11_Slowdown_2.gs --stdout > gt.5.xz  217.20s user 0.29s system 99% cpu 3:37.49 total
113M May  7 13:38 gt.5.xz
xz -6 -k GT3_R11_Slowdown_2.gs --stdout > gt.6.xz  290.68s user 0.28s system 99% cpu 4:50.97 total
110M May  7 13:27 gt.6.xz

-7/-8/-9 are slower for no gain.
```
I propose that we keep the compressed version for the short dump to ease sharing of them.

gregory38:
@turtleli 
Any opinion on the interface ? Is is better to use inheritance or add a boolean check ? (or other ?)

turtleli:
In this case I'd prefer inheritance, I think it'd be more readable.

The interface - I'm not keen on some parts. Personally I'd do this for the base class:
```
class GSDumpBase
{
public:
	GSDumpBase();
	virtual ~GSDumpBase() = default;

	void ReadFIFO(uint32 size);
	void Transfer(int index, const uint8* mem, size_t size);
	bool VSync(int field, bool last, const GSPrivRegSet* regs);

protected:
	void AddDumpHeader(uint32 crc, const GSFreezeData& fd, const GSPrivRegSet* regs);

private:
	virtual void AppendRawData(const void *data, size_t size) = 0;
	virtual void AppendRawData(uint8 c) = 0;

	int m_frames;
	int m_extra_frames;
};
```
so opening, writing, and closing the file is entirely up to the derived classes,, but everything else is taken care of by the base class. Well, opening and closing could perhaps be done by the base class too. For the derived classes, I'd do something like this:
```
class GSDump final: public GSDumpBase
{
public:
	GSDump(const std::string& fn, uint32 crc, const GSFreezeData& fd, const GSPrivRegSet* regs);
	~GSDump() override;

private:
	void AppendRawData(const void *data, size_t size) override;
	void AppendRawData(uint8 c) override;

	FILE* m_gs;
};

class GSDumpXz final: public GSDumpBase
{
public:
	GSDumpXz(const std::string& fn, uint32 crc, const GSFreezeData& fd, const GSPrivRegSet* regs);
	~GSDumpXz() override;

private:
	void AppendRawData(const void *data, size_t size) override;
	void AppendRawData(uint8 c) override;

	void Flush(bool close);

	FILE* m_gs;
	lzma_stream m_strm;
	std::vector<uint8> m_in_buff;
};
```
That way, long dumps can be written straight to file without devouring memory, and short dumps can be buffered, xz compressed, and then written to file.

gregory38:
@turtleli is this way better ? Note: I need to test the code.

plugins/GSdx/GSDump.cpp
```diff
@@ -22,74 +22,165 @@
...
+void GSDumpXz::AppendRawData(const void *data, size_t size)
+{
+	size_t old_size = m_in_buff.size();
+	m_in_buff.resize(old_size + size);
+	memcpy(&m_in_buff[old_size], data, size);
+
+	// Enough data was accumulated, time to write/compress it.  If compression
+	// is enabled, it will freeze PCSX2. 1GB should be enough for long dump.
+	if (m_in_buff.size() > 1024*1024*1024)
+		Flush(false);
```
turtleli: I guess the if won't ever be true anymore.

Actually, unless I'm misreading something, I think there's a small risk that `lzma_code(&m_strm, LZMA_FINISH)` won't be called if this path is ever taken (VSync -> flush(false) -> empty buffer -> destructor -> flush(true) -> early return -> `lzma_code(&m_strm, LZMA_FINISH)` not called + memleak because `lzma_end()` isn't called).

plugins/GSdx/GSDump.cpp
```diff
@@ -22,74 +22,165 @@
...
+
+	return ((++m_frames & 1) == 0 && last && (m_extra_frames < 0));
 }
 
-void GSDump::ReadFIFO(uint32 size)
+//////////////////////////////////////////////////////////////////////
+// GSDump implementation
+//////////////////////////////////////////////////////////////////////
+
+GSDump::GSDump(const string& fn, uint32 crc, const GSFreezeData& fd, const GSPrivRegSet* regs) : GSDumpBase(fn + ".gs")
```
turtleli: Maybe put the base class constructor on a new line? IMO it's a bit hard to notice. Ditto GSDumpXz.

plugins/GSdx/GSDump.cpp
gregory38: With the smart_ptr, there is no reason to close the stream in the flush. So I will move it in the destructor.

However, I'm not sure of what will happen if `lzma_code(&m_strm, LZMA_FINISH)` isn't called. Neither if I always call it in the destructor with a zero sized buffer.

plugins/GSdx/GSRenderer.cpp
```diff
@@ -463,28 +463,21 @@ void GSRenderer::VSync(int field)
...
 
-			m_dump.Open(m_snapshot, m_crc, fd, m_regs);
+#ifdef LZMA_SUPPORTED
+			if (m_control_key)
+				m_dump = std::unique_ptr<GSDumpBase>(new GSDump(m_snapshot, m_crc, fd, m_regs));
+			else
+				m_dump = std::unique_ptr<GSDumpBase>(new GSDumpXz(m_snapshot, m_crc, fd, m_regs));
+#else
+			m_dump = std::unique_ptr<GSDumpBase>(new GSDump(m_snapshot, m_crc, fd, m_regs));
+#endif
```
gregory38: I don't think it worth the extra code. However I did an early abort on Vsync. It should be enough.

plugins/GSdx/GSDump.cpp
gregory38: actually it seems to crash :(

gregory38: Yeah I missed with an output buffer to contain the end of stream. I tested short and long dump and it works as expected now.

plugins/GSdx/GSRenderer.cpp
turtleli: Is Vsync the first thing that is added to the dump after the header?

I think you might need to add an `if (m_gs == nullptr)` check in the GSDump constructor to prevent some error message spam if the file can't be opened.

gregory38: First Vsync is the next frame, so yes could be verbose. We need to do an initial Vsync when replaying.

I created a Write function to check size and m_gs properly. It should be better now. Is it ok for you ?

turtleli: Well, personally I would have added an init check function (there's less thinking required about what would happen if things go wrong), but yeah, it looks better.

plugins/GSdx/GSDump.cpp
```diff
@@ -22,74 +22,177 @@
...
+
+GSDumpXz::GSDumpXz(const std::string& fn, uint32 crc, const GSFreezeData& fd, const GSPrivRegSet* regs)
+	: GSDumpBase(fn + ".gs.xz")
 {
-	if(m_gs && size > 0)
-	{
-		fputc(2, m_gs);
-		fwrite(&size, 4, 1, m_gs);
+	m_strm = LZMA_STREAM_INIT;
+	lzma_ret ret = lzma_easy_encoder(&m_strm, 6 /*level*/, LZMA_CHECK_CRC64);
```
turtleli: Hmm. In the unlikely (I think?) case that it fails, it'll leave an empty file behind.

plugins/GSdx/GSDump.cpp
```diff
@@ -22,74 +22,177 @@
...
+	m_strm.avail_in = m_in_buff.size();
+
+	Compress(LZMA_RUN, LZMA_OK);
+
+	m_in_buff.clear();
+}
+
+void GSDumpXz::Compress(lzma_action action, lzma_ret expected_status)
+{
+	std::vector<uint8> out_buff(1024*1024);
```
turtleli: Would it be better to use an std::array here?

plugins/GSdx/GSDump.cpp
```diff
@@ -22,74 +22,177 @@
...
+	do {
+		m_strm.next_out = &out_buff[0];
```
turtleli: As an alternative, `out_buff.data()` could be used instead.

plugins/GSdx/GSDump.cpp
gregory38: yes, I will change it with the `.data()`

gregory38:
@turtleli is it good for you ?

turtleli:
Well, it looks ok, though the empty file creation if the xz init fails isn't ideal.

plugins/GSdx/GSDump.cpp
```diff
@@ -22,74 +22,177 @@
...
+	AppendRawData(3);
+	AppendRawData(regs, sizeof(*regs));
+
+	AppendRawData(1);
+	AppendRawData(field);
+
+	if (last)
+		m_extra_frames--;
+
+	return ((++m_frames & 1) == 0 && last && (m_extra_frames < 0));
```
turtleli: Some unnecessary parentheses have been used.

plugins/GSdx/GSDump.h
```diff
@@ -43,20 +43,57 @@ Regs data (id == 3)
...
-class GSDump
+class GSDumpBase
 {
-	FILE* m_gs;
 	int m_frames;
 	int m_extra_frames;
+	FILE* m_gs;
+
+protected:
+
```
turtleli: Unnecessary blank line.

plugins/GSdx/GSRenderer.cpp
```diff
@@ -496,24 +489,10 @@ void GSRenderer::VSync(int field)
...
-
-			#else
-
-			control = m_control_key;
-
-            #endif
-
-	    	m_dump.VSync(field, !control, m_regs);
-		}
+		if (m_dump->VSync(field, !m_control_key, m_regs))
```
turtleli: Would be nice to clang-format GSdx and PCSX2 some day. ;)

gregory38: It is the first thing that I will do after the release. And I will likely do the core too.

gregory38: Update done. Thanks for the review.

gregory38:
@turtleli Could the code be merged ? Could you take care of the Win side :)

turtleli:
Yeah, you can merge it.

As for Windows side, it might take a while - I might experiment with submodules a bit.

gregory38:
How long is a "while" for you ? Weeks/Months ?

turtleli:
Maybe a week or two? It's mostly to figure out how best to handle VS project files when building stuff in submodules.

gregory38:
No pressure, it is fine ;) I'm looking forward for the submodule support.

PR #1925, opened 2017-05-02
GSDX-Windows: Minor fixes to the GUI

ssakash:
**Summary of changes**:

* Removed a doubled bracket from one of the options texts in mipmapping combo box.
* Converted the group box text from ``OpenGL Very Advanced Custom Settings`` to ``OpenGL Advanced Settings``. I felt the older name was unnecessarily verbose and long, changed it into something more simpler.
* Added proper grayouts to the shadeboost option texts, making it similar to the grayout behavior of external shader option.


plugins/GSdx/resource.h
```diff
@@ -76,70 +76,73 @@
...
+#define IDC_PRELOAD_GS                  2117
+#define IDC_FAST_TC_INV                 2118
+#define IDC_AUTO_FLUSH                  2119
+#define IDC_MEMORY_WRAPPING             2120
+#define IDC_TRI_FILTER                  2121
+#define IDC_TRI_FILTER_TEXT             2122
+#define IDC_UNSCALE_POINT_LINE          2123
+#define IDC_MERGE_PP_SPRITE             2124
 // Shader:
-#define IDC_SHADEBOOST                  2140
```
turtleli: I'm guessing the number gap was to avoid having to change the whole RC file every time something is added/removed. Dunno.

turtleli:
Didn't test it, but it looks fine to me.

Small things:
1. paranthesis -> parenthesis
2. It might be nice to manually wrap your commit messages if the editor doesn't automatically do it for you (it makes the commit message easier to view)

ssakash:
> paranthesis -> parenthesis

Damn, I got the spelling of "parenthesis" wrong my whole life. 😨 

> It might be nice to manually wrap your commit messages if the editor doesn't automatically do it for you (it makes the commit message easier to view)

To be honest, I didn't like it. That's why I had manually removed the line wrapping done by the editor, wasting all those horizontal space doesn't make sense to me. I can make an exception for this PR, but in the future - it would depend upon my preference.

turtleli:
> To be honest, I didn't like it. That's why I had manually removed the line wrapping done by the editor, wasting all those horizontal space doesn't make sense to me. I can make an exception for this PR, but in the future - it would depend upon my preference.

The line breaks can make things easier to read. Here's an example (`git log` on command line, should have cropped the image more but I already uploaded it):

gitlog

PR #1926, opened 2017-05-02
PCSX2-Core: Update videomode enumerations and fix a SMODE2 delay bug in title update

ssakash:
**Summary of changes**:

* Update the enumeration of videomodes with proper prefix of the television system. (Thanks to @colepcsx2 for informing about it)

* Fix a bug where an interlaced mode game will be displayed as progressive for a short moment of time due to the delay in SMODE2 value initialization. (We overcome the issue by waiting for the frame rendering to take place - SMODE2 init will be taken place before that time)


gregory38:
Thanks you.

PR #1934, opened 2017-05-08
PCSX2-WX: Add Manual ISO selection option

ssakash:
**Summary of changes**:

* Add a ``Manual ISO selection`` option which allows you to select your ISO via the file explorer each time whenever you ``full/fast boot``. This one is for you @vsub ❤️ 

newoption

PCSX2 by default only allows you to hold 12 ISOs in the recent ISO list and when you try to add anymore, the previous ones will replaced by your newer ISO selection. The only way to modify the total limit of ISOs capable of being stored in the recent ISO list is to change the value of an INI variable, obviously no end user would be aware of it.

Hence for people with large amount of ISOs in their storage devices, it would be ideal if they could just directly look for their ISOs instead of looking through the recent ISO list which is not capable of storing all of their games.

Another alternative would be to increase the default maximum no.of ISOs allowed in the recent ISO list but that would just make the list a bit larger and at one point it would even grow big enough to touch your taskbar! there's no possible way that the recent ISO list would be able to accommodate all the ISOs of say someone who has a collection of over 50+ games. Also it's not desirable behavior to go to ``CDVD`` -> ``ISO Selector`` -> ``Browse..`` to add individual ISOs each and ever single time. -_-

To avoid such issues, I'm proposing to add this option which would directly take you to the file explorer when you're doing a full/fast boot. Any opinion, guys?

### DIFF
--- a/pcsx2/gui/RecentIsoList.cpp
+++ b/pcsx2/gui/RecentIsoList.cpp
@@ -171,13 +171,16 @@ void RecentIsoManager::InsertIntoMenu( int id )
 		wxid = this->m_firstIdForMenuItems_or_wxID_ANY + id;
 
 	curitem.ItemPtr = m_Menu->AppendRadioItem( wxid, Path::GetFilename(curitem.Filename), curitem.Filename );
-	bool exists = wxFileExists( curitem.Filename );
+	curitem.ItemPtr->Enable(wxFileExists(curitem.Filename) && !g_Conf->AskOnBoot);
+}
 
-	if( m_cursel == id && exists )
-		curitem.ItemPtr->Check();
-
-	if ( !exists )
-		curitem.ItemPtr->Enable( false );
+void RecentIsoManager::EnableItems(bool display)
+{
+	for (const auto& list : m_Items)
+	{
+		// Files which don't exist still need to be grayed out.
+		list.ItemPtr->Enable(display && wxFileExists(list.Filename));
+	}
 }
 
 void RecentIsoManager::LoadListFrom( IniInterface& ini )
--- a/pcsx2/gui/App.h
+++ b/pcsx2/gui/App.h
@@ -86,6 +86,7 @@ enum MenuIdentifiers
 	MenuId_Boot_Iso,			// Opens submenu with Iso browser, and recent isos.
 	MenuId_RecentIsos_reservedStart,
 	MenuId_IsoBrowse = MenuId_RecentIsos_reservedStart + 100,			// Open dialog, runs selected iso.
+	MenuId_Ask_On_Booting,
 	MenuId_Boot_CDVD,
 	MenuId_Boot_CDVD2,
 	MenuId_Boot_ELF,
--- a/pcsx2/gui/AppConfig.cpp
+++ b/pcsx2/gui/AppConfig.cpp
@@ -670,6 +670,7 @@ void AppConfig::LoadSaveRootItems( IniInterface& ini )
 
 	IniEntry( EnablePresets );
 	IniEntry( PresetIndex );
+	IniEntry( AskOnBoot );
 	
 	#ifdef __WXMSW__
 	IniEntry( McdCompressNTFS );
--- a/pcsx2/gui/AppConfig.cpp
+++ b/pcsx2/gui/AppConfig.cpp
@@ -535,7 +535,7 @@ AppConfig::AppConfig()
 {
 	LanguageId			= wxLANGUAGE_DEFAULT;
 	LanguageCode		= L"default";
-	RecentIsoCount		= 12;
+	RecentIsoCount		= 20;
 	Listbook_ImageSize	= 32;
 	Toolbar_ImageSize	= 24;
 	Toolbar_ShowLabels	= true;
--- a/pcsx2/gui/AppConfig.h
+++ b/pcsx2/gui/AppConfig.h
@@ -315,6 +315,8 @@ public:
 	bool		EnablePresets;
 	int			PresetIndex;
 
+	bool		AskOnBoot;
+
 	wxString				CurrentIso;
 	wxString				CurrentELF;
 	wxString				CurrentIRX;
--- a/pcsx2/gui/AppRes.cpp
+++ b/pcsx2/gui/AppRes.cpp
@@ -63,7 +63,10 @@ const wxImage& LoadImageAny(
 RecentIsoList::RecentIsoList(int firstIdForMenuItems_or_wxID_ANY)
 {
 	Menu = std::unique_ptr<wxMenu>(new wxMenu());
-	Menu->Append( MenuId_IsoBrowse, _("Browse..."), _("Browse for an ISO that is not in your recent history.") );
+	Menu->AppendCheckItem( MenuId_Ask_On_Booting, _("Always ask when booting"), _("Manually select an ISO upon boot ignoring the selection from recent ISO list.") );
+	Menu->AppendSeparator();
+	Menu->Append( MenuId_IsoBrowse, _("Browse..."), _("Browse for an ISO that is not in your recent history."))->Enable(!g_Conf->AskOnBoot);
+	Menu->Check( MenuId_Ask_On_Booting, g_Conf->AskOnBoot );
 	Manager = std::unique_ptr<RecentIsoManager>(new RecentIsoManager( Menu.get(), firstIdForMenuItems_or_wxID_ANY ));
 }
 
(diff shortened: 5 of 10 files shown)


willkuer:
Very good feature! I always wanted ti have something like that. However somehow I got used to the inconvinient Browse button.

Maybe 'Browse upon boot' or 'Iso selection upon boot' or something like that? Without your commit explanation I wouldn't understand what 'Manual ISO selection' means.

Also I didn't check the implementation but according to your screenshot you should probably additionally disable the 'Browse' menu entry whenever manual iso is enabled. I think it should then also be placed on top of Browse as the first element. Maybe seperated by another menu seperator as the browse button is then a slave to the checkbox.

Or you can add an 'Iso selection' menu entry with two real slave entries (only appear on mouse hover on the right) which are radio buttons with label 'Manual ISO selection' or whatever variant and 'Remember ISO selection' or 'Reload last booted game'

Also manual should be probably lower case or selection uppercase if you prefer title case style over normal case style.

gregory38:
Tbh 12 iso entries is really small. I'm also in favor to increase it. Maybe 20-25 would be better, dunno if it fit on small display.

avih:
I also think we could increase the maximum to 20-25. However, I haven't tested how it behaves when the list becomes taller than the screen. Though I doubt 20 will be too much on any screen.

Another thing we could do, is improve the MRU behavior (i.e. which files get to the top of the list and when). At the time when I implemented the MRU I thought that keeping the ISO's place at the list is more important than moving the last used ISO to the top, but maybe we can change that too. Should be trivial, and I think it would feel some natural to users.

ssakash:
> disable the 'Browse' menu entry whenever manual iso is enabled. I think it should then also be placed on top of Browse as the first element. Maybe seperated by another menu seperator as the browse button is then a slave to the checkbox.

> Also manual should be probably lower case or selection uppercase if you prefer title case style over normal case style.

Done, thanks a lot for the suggestions. 👍 

> Maybe 'Browse upon boot' or 'Iso selection upon boot' or something like that? Without your commit explanation I wouldn't understand what 'Manual ISO selection' means.

There's a help text which will be displayed on the console log window when the user hovers over the menu item, maybe that might help? Though to be honest, I think the text ``Manual ISO Selection`` in itself is rather easy to understand for users. (Hopefully I'm not the only one who feels so?)

> Tbh 12 iso entries is really small. I'm also in favor to increase it. Maybe 20-25 would be better, dunno if it fit on small display.

25 and the list is reaching my taskbar. :P 

omg

gregory38:
Why not 20 ? What is your screen resolution ?

ssakash:
> Why not 20 ? 

I'm not sure about people with even lower screen size so I left it at 15 just to be on the safe side. I don't mind updating it to 20, it's just troublesome when the list gets larger to a point where it takes extra effort to select the bottom most entry.

> What is your screen resolution ?

1366 x 768 (It's the most common resolution for 15.6” laptops)

gregory38:
So 20 should fit on all displays (or at least 18). I don't understand why manufacturers don't upgrade display to full-hd to boost sell.

avih:
I think just changing the default list size is enough (I'd choose 20 but I'm fine with other values too).

I don't really like the original/rest of this PR though.

Before:
- To boot a different game: choose one from the MRU (or open the ISO browser from the same menu) and then boot.
- To boot the same game: just boot.

After (assuming the user has the new option enabled):
- To boot a different game: Boot, then find your ISO in the browser.
- To boot the same game: same as booting a different game.

So the only scenario where it saves clicks (two click) is if you're changing games all the time and your list is too short to hold the all the games you're switching between.

To me, it's not worth "9 files modified" of this PR, as small as the changes might be.

willkuer:
@avih I would like to have this feature. Seems like vsub as well. The nice thing about this PR is that you can always have the old behavior if you like it more.

@ssakash Adding entries to the list might also be related to the DPI settings. Possibly 20 is fine on default settings but not if the DPI are enhanced.

Hirato:
Usually when Menus get too big to fix on screen, they either get scrolling arrows at the top and bottom, or they overflow into a second column (or third, or fourth).

turtleli:
> There's a help text which will be displayed on the console log window when the user hovers over the menu item, maybe that might help? Though to be honest, I think the text Manual ISO Selection in itself is rather easy to understand for users. (Hopefully I'm not the only one who feels so?)

I find it hard to understand. ;) It's probably best if a user can understand what a menu item does without having to look at the help text. I dunno what would be better though (maybe "Always ask when booting" or one of willkuer's suggestions or some other random thought).

I'm not sure it's a great idea to disable the browse button (at least after booting) when the feature is enabled. What if the user has to swap discs?

gregory38:
It isn't expected that menu got out of screen.
http://wxwidgets.10942.n7.nabble.com/Getting-wxMenu-size-td88159.html

ssakash:
> I find it hard to understand. ;) It's probably best if a user can understand what a menu item does without having to look at the help text. I dunno what would be better though (maybe "Always ask when booting" or one of willkuer's suggestions or some other random thought).

Fair enough, changed it to ``Always ask when booting`` just to be on the safe side. I figured you might nitpick if I chose a name of my own. :P

pcsx2/gui/RecentIsoList.cpp
```diff
@@ -178,6 +178,23 @@ void RecentIsoManager::InsertIntoMenu( int id )
...
+	if (m_Items.empty())
+		return;
+
+	for (const RecentIsoManager::RecentItem& list : m_Items)
+	{
+		list.ItemPtr->Enable(display);
+
+		// Files which don't exist still need to be grayed out.
+		if(!wxFileExists(list.Filename))
+			list.ItemPtr->Enable(false);
```
turtleli: I think this could just be
```
list.ItemPtr->Enable(display && wxFileExists(list.Filename));
```

pcsx2/gui/RecentIsoList.cpp
```diff
@@ -178,6 +178,23 @@ void RecentIsoManager::InsertIntoMenu( int id )
...
 	if ( !exists )
 		curitem.ItemPtr->Enable( false );
+	else
+		curitem.ItemPtr->Enable( !g_Conf->AskOnBoot );
+}
+
+void RecentIsoManager::EnableItems(bool display)
+{
+	if (m_Items.empty())
+		return;
```
turtleli: The early return isn't really necessary in this case - the for loop will just be skipped if `m_Items` is empty.

ssakash:
@turtleli 

Should this be merged before 1.6 (or) postponed after 1.6 release? I'll let you handle the decision, :)

turtleli:
Well, I'm not sure about the browse button behaviour when the option is enabled. I think the behaviour should be:
 * Game/BIOS not booted/shutdown -> Browse button disabled
 * Otherwise -> Browse button enabled

ssakash:
> Well, I'm not sure about the browse button behaviour when the option is enabled. I think the behaviour should be:

Added the behavior which you have described, anything else left to do?

turtleli:
The browse button stuff doesn't work properly if I boot an ISO from the command line - the Browse button remains disabled.

I also wonder whether the browse button behaviour should be extended to the recent ISO list too (I think it's a good idea, but I'm unsure).

pcsx2/gui/RecentIsoList.cpp
```diff
@@ -178,6 +178,17 @@ void RecentIsoManager::InsertIntoMenu( int id )
 
 	if ( !exists )
 		curitem.ItemPtr->Enable( false );
+	else
+		curitem.ItemPtr->Enable( !g_Conf->AskOnBoot );
+}
+
+void RecentIsoManager::EnableItems(bool display)
+{
+	for (const RecentIsoManager::RecentItem& list : m_Items)
```
turtleli: `const auto&`? It's up to you.

ssakash:
> The browse button stuff doesn't work properly if I boot an ISO from the command line - the Browse button remains disabled.

Nice catch, I pushed a more generic solution to handle it.

> I also wonder whether the browse button behaviour should be extended to the recent ISO list too (I think it's a good idea, but I'm unsure).

I also think it's a good idea, implemented it.

turtleli:
Something isn't right since the latest update - it potentially crashes when closing PCSX2.

ssakash:
> Something isn't right since the latest update - it potentially crashes when closing PCSX2.

Should be sorted out now I guess.

pcsx2/gui/RecentIsoList.cpp
```diff
@@ -178,6 +178,17 @@ void RecentIsoManager::InsertIntoMenu( int id )
 
 	if ( !exists )
 		curitem.ItemPtr->Enable( false );
+	else
+		curitem.ItemPtr->Enable( !g_Conf->AskOnBoot );
```
turtleli: Combine the if and else? `curitem.ItemPtr->Enable(exists && !g_Conf->AskOnBoot);`

Apart from that, I think it's fine.

ssakash: I think the local variable ``exists`` is also pointless. The function's name is descriptive on it's own.

vsub:
Not that I can try this feature right now but thanks.

vsub:
@ssakash 
Sorry for the late responce

Seems to be working fine but there is one small problem(not really a big deal(I think)because I can just restart pcsx2 after that)

If the option is enabled and the first things you use is anything other that ISO(for exaple No Disk to access the bios),when you switch to ISO,you are asked what to do,swap the disk or reset.

Both options lead to the "CDVD plugin failed to open" message and then it ask you if you want to check the plugins or not.

With "Swap Disk" you can just hit cancel and then fast\full boot to get asked what game to load
With "Reset",you will again get the same question but there is nothing you can do there because pcsx2 freezes and you have to kill the process

PR #1940, opened 2017-05-11
Greg/modernize use default

gregory38:
Initial goal was to replace constructor/destructor code reported by clang-tidy.

But I learn in the process that destructors are noexcept by default. So `throw()` => /dev/null

And I found plenty of case of trivial destructor not reported by clang-tidy but that I feel could be `= default` too.

Finally, I replaced the few remaining `throw()` by `noexcept`.


gregory38:
Thanks you.

PR #1942, opened 2017-05-12
GSdx-HW: Optimize RT buffer size calculation for custom resolutions

ssakash:
**Summary of changes**:

* Revamped custom resolution buffer scaling code making it more efficient, should provide decent performance boost when large framebuffer is enabled at GS limited scenarios.

* Allow future resizing of the buffer size for custom resolution when the buffer required is increased due to the changes in CRTC/Framebuffer dimensions.

Honestly, string formatting is awful in C++ without any external libraries like Boost, It's so simple in Python. They should really consider providing any easier alternative instead of using ``sprintf/snprintf``, using the streams for formatting is even more worse. I've made the most stupid use of smart pointers ever. :P

Here are some benchmarks of the improvement, (Mostly done at GS limited scenarios)

**Master Branch**:

Game | Resolution | Large Framebuffer disabled | Large framebuffer enabled
------------ | ------------- | ------------ | -------------
ICO | 1024x1024 | 28 fps | 28 fps
ICO | 3840x2160 | 15 fps | 12 fps
Digimon Rumble Arena | 1024x1024 | 46 fps | 46 fps
Digimon Rumble Arena | 3840x2160 | 34 fps | 32 fps
Ben10 Alien Force: Vilgax Attacks | 1024x1024 | 95 fps | 95 fps
Ben10 Alien Force: Vilgax Attacks | 3840x2160 | 26 fps | 2 fps

**Pull Request Branch**:

Game | Resolution | Large Framebuffer disabled | Large framebuffer enabled
------------ | ------------- | ------------ | -------------
ICO | 1024x1024 | 28 fps | 28 fps
ICO | 3840x2160 | 15 fps | 12 fps
Digimon Rumble Arena | 1024x1024 | 46 fps | 46 fps
Digimon Rumble Arena | 3840x2160 | 34 fps | 34 fps
Ben10 Alien Force: Vilgax Attacks | 1024x1024 | 95 fps | 95 fps
Ben10 Alien Force: Vilgax Attacks | 3840x2160 | 26 fps | 26 fps

To be less verbose, the new buffer size calculation algorithm only increases the buffer size when necessary (Like at ICO) and doesn't increase them unnecessarily. (giving performance boost in GS limited games like Ben10 Alien Force: Vilgax Attacks)

### DIFF
--- a/plugins/GSdx/GSRendererHW.cpp
+++ b/plugins/GSdx/GSRendererHW.cpp
@@ -65,6 +65,12 @@ GSRendererHW::GSRendererHW(GSTextureCache* tc)
 
 void GSRendererHW::SetScaling()
 {
+	if (!m_upscale_multiplier)
+	{
+		CustomResolutionScaling();
+		return;
+	}
+
 	GSVector2i crtc_size(GetDisplayRect().width(), GetDisplayRect().height());
 
 	// Details of (potential) perf impact of a big framebuffer
@@ -111,52 +117,6 @@ void GSRendererHW::SetScaling()
 	int upscaled_fb_w = fb_width * m_upscale_multiplier;
 	int upscaled_fb_h = fb_height * m_upscale_multiplier;
 	bool good_rt_size = m_width >= upscaled_fb_w && m_height >= upscaled_fb_h;
-	bool initialized_register_state = (m_context->FRAME.FBW > 1) && (crtc_size.y > 1);
-
-	if (!m_upscale_multiplier && initialized_register_state)
-	{
-		if (m_height == m_custom_height)
-		{
-			const float scaling_ratio = ceil(static_cast<float>(m_custom_height) / crtc_size.y);
-			// Avoid using a scissor value which is too high, developers can even leave the scissor to max (2047)
-			// at some cases when they don't want to limit the rendering size. Our assumption is that developers
-			// set the scissor to the actual data in the buffer. Let's use the scissoring value only at such cases
-			const int scissor_height = std::min(640u, (m_context->SCISSOR.SCAY1 - m_context->SCISSOR.SCAY0) + 1);
-			const int single_buffer_size = std::max(GetDisplayRect().height(), scissor_height);
-
-			// We have two contexts of framebuffer height -
-			// One for lower memory consumption and low accuracy
-			// Another one for higher memory consumption at necessary scenarios for higher accuracy
-			std::array<int, 2> framebuffer_height;
-			// When Large Framebuffer is disabled - Let's only consider the height of the display rectangle
-			// as the base (This is wrong implementation when we consider it theoretically as CRTC has no relation
-			// to the Framebuffer size)
-			framebuffer_height[0] = static_cast<int>(round((crtc_size.y * scaling_ratio)));
-			// When Large Framebuffer is enabled - We also consider for potential scissor sizes which are around
-			// the size of the actual image data stored. (Helps ICO to properly scale to right size by help of the
-			// scissoring values) Display rectangle has a height of 256 but scissor has a height of 512 which seems to
-			// be the real buffer size.
-			framebuffer_height[1] = static_cast<int>(round(single_buffer_size * scaling_ratio));
-
-			m_tc->RemovePartial();
-			m_width = std::max(m_width, native_buffer.x);
-			m_height = std::max(framebuffer_height[m_large_framebuffer], native_buffer.y);
-
-			std::string overhead = to_string(framebuffer_height[1] - framebuffer_height[0]);
-			std::string message = "(Custom resolution) Framebuffer size set to " + to_string(crtc_size.x) + "x" + to_string(crtc_size.y);
-			message += " (" + to_string(m_width) + "x" + to_string(m_height) + ")\n";
-
-			if (m_large_framebuffer)
-			{
-				message += "Additional " + overhead + " pixels overhead by enabling Large Framebuffer\n";
-			}
-			else
-			{
-				message += "Saved " + overhead + " pixels overhead by disabling Large Framebuffer\n";
-			}
-			printf("%s", message);
-		}
-	}
 
 	// No need to resize for native/custom resolutions as default size will be enough for native and we manually get RT Buffer size for custom.
 	// don't resize until the display rectangle and register states are stabilized.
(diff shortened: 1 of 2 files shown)


plugins/GSdx/GSRendererHW.cpp
```diff
@@ -138,6 +129,52 @@ void GSRendererHW::SetScaling()
...
+	if (m_width >= m_custom_width && m_height >= framebuffer_height[m_large_framebuffer])
+		return;
+
+	m_tc->RemovePartial();
+	m_width = max(m_width, native_buffer.x);
+	m_height = max(framebuffer_height[m_large_framebuffer], native_buffer.y);
+
+	const int overhead = framebuffer_height[1] - framebuffer_height[0];
+	std::unique_ptr<char[]> buffer = std::make_unique<char[]>(150);
+	int size = snprintf(buffer.get(), 150, "(Custom resolution) Framebuffer size set to %dx%d (%dx%d)\n", crtc_width, crtc_height, m_width, m_height);
```
gregory38: Oh please don't do this kind of code. You can use `std::string foo = "my message " + std::to_string(integer_var) + " x " .. ...`

ssakash: Initially I wanted to do it in C++ style, something like,

``` cpp
std::stringstream s;
s >> "(Custom resolution) Framebuffer size set to" << crtc_width << "x" ...;
```

The thing is that on both this and your solution, the code will be too long horizontally, so I decided to just format it using snprintf. What's wrong with the current way? This is the traditional C way to format strings. (Though aside from using smart pointer instead of bare one)

gregory38: you know you can hit enter ;)

Or you can do
```
message = "foo";
message += to_string(2);
message += "esfsfs";
```

C is awful. And you're code might not compile on Linux.

gregory38:
Anyway, I'm not sure it is a good solution (but I need to understand what you did first). Maybe it would be possible to dynamically allocate some buffers.

The constrain is that size of the rendering on the GPU is intersection of all buffers (color + depth).

We can monitor the real size of the GS draw call, Here a kind of algo (might need some tuning)
* if depth enabled
  * if cache hit, increase the size of the buffer is needed (copy)
  * else allocate both (color depth) buffers based on max FBW/Display BW values ever reported
* else
  * if cache hit, increase the size of the buffer is needed (copy)
  * else if sprite rendering and draw is small (let's say 128x128) allocate a RT of 128x128
  * else allocate a buffer based on max FBW/Display BW values ever reported

plugins/GSdx/GSRendererHW.cpp
ssakash: In python, it's as simple as,

``` Python
s = "(Custom resolution) Framebuffer size set to {}x{} ({}x{})".format(crtc_width..)
```

Standard string library of C++ Sucks :/

Though fair enough, I'll convert it to ``std::string`` as you suggested.

gregory38: Otherwise just use
```
const char* first_word = (large) ? "large" : "save";
fprintf(.., "%s ... %d", first_word... 14);
```
All but the above code ;)

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -138,6 +129,52 @@ void GSRendererHW::SetScaling()
...
+	// scissoring values) Display rectangle has a height of 256 but scissor has a height of 512 which seems to
+	// be the real buffer size.
+	framebuffer_height[1] = static_cast<int>(round(single_buffer_size * scaling_ratio));
+
+	if (m_width >= m_custom_width && m_height >= framebuffer_height[m_large_framebuffer])
+		return;
+
+	m_tc->RemovePartial();
+	m_width = max(m_width, native_buffer.x);
+	m_height = max(framebuffer_height[m_large_framebuffer], native_buffer.y);
```
turtleli: `std::`

plugins/GSdx/GSRendererHW.cpp
danilaml: @ssakash might want to take a look: http://fmtlib.net/latest/index.html
The author is currently working on a std proposal: http://fmtlib.net/Text%20Formatting.html

gregory38: Oh, I didn't know the std proposal. Interesting, however best case is 2020/2021 so it won't available in the standard soon (if it is ever accepted)

ssakash:
> Anyway, I'm not sure it is a good solution (but I need to understand what you did first). Maybe it would be possible to dynamically allocate some buffers.
The constrain is that size of the rendering on the GPU is intersection of all buffers (color + depth).

>We can monitor the real size of the GS draw call, Here a kind of algo (might need some tuning)

>    if depth enabled
        if cache hit, increase the size of the buffer is needed (copy)
        else allocate both (color depth) buffers based on max FBW/Display BW values ever reported
    else
        if cache hit, increase the size of the buffer is needed (copy)
        else if sprite rendering and draw is small (let's say 128x128) allocate a RT of 128x128
        else allocate a buffer based on max FBW/Display BW values ever reported

I'm not sure if your idea will be reliable, there are too much assumptions on it which makes the algorithm seem a bit volatile, though I haven't really looked much into it ``(lack of free time)`` and plus not too keen on adding a new algorithm just before a release, don't want too much regressions on my head. ;)

I'd prefer to just keep the current reliable algorithm for now.

lightningterror:
> I'm not sure if your idea will be reliable, there are too much assumptions on it which makes the algorithm seem a bit volatile, though I haven't really looked much into it (lack of free time) and plus not too keen on adding a new algorithm just before a release, don't want too much regressions on my head. ;)

There's gonna be plenty of other releases for that, save it for 1.7 xD

gregory38:
Sorry, I didn't mean to change it now.

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -117,13 +117,44 @@ void GSRendererHW::SetScaling()
...
 		if (m_height == m_custom_height)
 		{
-			float ratio = ceil(static_cast<float>(m_height) / crtc_size.y);
-			float buffer_scale_offset = (m_large_framebuffer) ? ratio : 0.5f;
-			ratio = round(ratio + buffer_scale_offset);
+			const float scaling_ratio = ceil(static_cast<float>(m_custom_height) / crtc_size.y);
+			// Avoid using a scissor value which is too high, developers can even leave the scissor to max (2047)
+			// at some cases when they don't want to limit the rendering size. Our assumption is that developers
+			// set the scissor to the actual data in the buffer. Let's use the scissoring value only at such cases
+			const int scissor_height = std::min(640u, (m_context->SCISSOR.SCAY1 - m_context->SCISSOR.SCAY0) + 1);
```
gregory38: I don't like to rely on scissor data here. Scissor is a per-draw call parameter. You can't infer anything with it. At least not like that. You can do it, if you check the scissor value of all draw calls.
```
Union on all draw calls of (scissor inter r_draw)
```

gregory38: But I guess most of your gain come from it. What happen if you keep `640` instead.

ssakash:
> But I guess most of your gain come from it. What happen if you keep 640 instead.

It'll work but will consume more memory than necessary, like cases at ICO where scissor is 512 and the output circuit height is 256. And also similar situations at VP2/SH2 where output circuit is 448 but scissor 512 (needed for proper scaling).

I think using the scissor should be safe here, we have a safe lower limit (output circuit height) and a safe upper limit (640), nothing could possibly go wrong here. We're only using the scissor when it has a sensible value.

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -138,6 +129,53 @@ void GSRendererHW::SetScaling()
...
+	if (m_width >= m_custom_width && m_height >= framebuffer_height[m_large_framebuffer])
+		return;
+
+	m_tc->RemovePartial();
+	m_width = std::max(m_width, native_buffer.x);
+	m_height = std::max(framebuffer_height[m_large_framebuffer], native_buffer.y);
+
+	std::string overhead = to_string(framebuffer_height[1] - framebuffer_height[0]);
+	std::string message = "(Custom resolution) Framebuffer size set to " + to_string(crtc_width) + "x" + to_string(crtc_height);
+	message += " (" + to_string(m_width) + "x" + to_string(m_height) + ")\n";
```
turtleli: `std::to_string`

gregory38:
On Mon, 22 May 2017 09:28:40 -0700
Akash <notifications@github.com> wrote:

> > But I guess most of your gain come from it. What happen if you keep 640 instead.  
> 
> It'll work but will consume more memory than necessary, like cases at ICO where scissor is 512 and the output circuit height is 256. And also similar situations at VP2/SH2 where output circuit is 448 but scissor 512 (needed for proper scaling).
> 
> I think using the scissor should be safe here, we have a safe lower limit (output circuit height) and a safe upper limit (640), nothing could possibly go wrong here. We're only using the scissor when it has a sensible value.
> 
> 

It will work most of the time because most games will use latest frame buffer as image source and will use a scissor at least on the framebuffer size.

But you don't have any guarantee.

However if I understand it right, the size is only increased. So we can suspect the value of scissor will be big enough at least once.

gregory38:
@ssakash I let you handle the tutleli feedback and then we could merge it.

ssakash:
A bit busy due to exams, It'll take me some time to update the code. (Preferably at 1st of June)

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -138,6 +129,53 @@ void GSRendererHW::SetScaling()
 	printf("Frame buffer size set to  %dx%d (%dx%d)\n", fb_width, fb_height , m_width, m_height);
 }
 
+void GSRendererHW::CustomResolutionScaling()
+{
+	const int crtc_width = GetDisplayRect().width();
+	const int crtc_height = GetDisplayRect().height();
+	const float scaling_ratio = ceil(static_cast<float>(m_custom_height) / crtc_height);
```
ssakash: ``std::ceil``

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -138,6 +129,53 @@ void GSRendererHW::SetScaling()
...
+	const int single_buffer_size = std::max(GetDisplayRect().height(), scissor_height);
+
+	// We have two contexts of framebuffer height -
+	// One for lower memory consumption and low accuracy
+	// Another one for higher memory consumption at necessary scenarios for higher accuracy
+	std::array<int, 2> framebuffer_height;
+	// When Large Framebuffer is disabled - Let's only consider the height of the display rectangle
+	// as the base (This is wrong implementation when we consider it theoretically as CRTC has no relation
+	// to the Framebuffer size)
+	framebuffer_height[0] = static_cast<int>(round((crtc_height * scaling_ratio)));
```
ssakash: ``std::round``

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -138,6 +129,53 @@ void GSRendererHW::SetScaling()
...
+	// When Large Framebuffer is enabled - We also consider for potential scissor sizes which are around
+	// the size of the actual image data stored. (Helps ICO to properly scale to right size by help of the
+	// scissoring values) Display rectangle has a height of 256 but scissor has a height of 512 which seems to
+	// be the real buffer size.
+	framebuffer_height[1] = static_cast<int>(round(single_buffer_size * scaling_ratio));
```
ssakash: ``std::round``

gregory38:
@ssakash 
Maybe removes the `use std` statement and fixes all missing `std::` in a future commit :)

ssakash:
> Maybe removes the use std statement and fixes all missing std:: in a future commit :)

I figure @turtleli would be more interested, I don't really care much about it.

turtleli:
Yeah, I've already removed `using namespace std;` in one of my branches ;)

gregory38:
Are we good ?

ssakash:
> Are we good ?

I think we're good to go.

lightningterror:
1-2 fps difference in GT4. More fps is always welcome.

Issue #1756, opened 2017-01-10
Gsdx dicussion: improvement of texture cache list container


gregory38:
GSdx need to manage a pool of texture pointer to properly handle the texture cache. So far the main object
is an array of list : `list<Source*> m_map[MAX_PAGES]`. One list by GS memory page (512). It is quite costly on the CPU side. So let do a discussion how we can improve the list container (if possible).

Here a brief explanation of the operation done on the list.
* Texture lookup will search a valid texture on a single list. It is costly to validate a texture. So a `MRU` (most recently used) algorithm is done. I.E when you have a hit, the texture pointer is moved at the front of the list. So far it is done with `splice`. MRU potentially avoid pitfall such as reusing wrongly an old texture.
* Texture insertion in case of cache miss. Technically it is less called than a texture lookup but there is a trick. You need to insert the texture pointer in **all** lists that is impacted by the texture. (I.E, there is texture data in the GS page). Therefore huge texture will do MAX_PAGES list insertions.
* Texture removal. As the texture pointer were inserted in multiple lists. You need to find them **all**, and delete the texture pointer of **all** lists. Again for huge texture, it will require to traverse MAX_PAGES lists (hopefully early exit on first erase) + to erase 1 element of MAX_PAGES lists.

So in short, we need a container that is ordered, got fast lookup, fast insertion, fast deletion. (AKA the ultimate container ;) ) Here an idea of a list container that try its best to fit the above requirements.

Hypothesis: the list size is rather small and likely smaller than 64K. 
```
template T
struct element {
      T  data;
      u16 next_index;
      u16 prev_index;
};
```

```
template T
class fast_list {
    array<element<T>, 64K> m_buffer; // the element 0 will be the head of the list. So there is always 1 element
    vector<u16> m_free_list; // contain a list of index that are free in m_buffer. Or a static array to avoid indirect memory lookup and extra memory overhead

   // Iterator note will be a u16 index. Iterator is never invalidated in insert/erase operation (except on operation on itself)

   void list_insert(Iterator it) {
         update the prev/next index to add it in the chain
   }
   void list_remove(Iterator it) {
         update the prev/next index to remove it of the chain
   }

   // Return the iterator so we can easily delete it without a traversal of the list.
   Iterator insert(T) {
       // get a free index
       Iterator free_index =m_free_list.back();
       m_free_list.pop_back();
      
       m_buffer[free_index].data = T;
       list_insert(free_index)
      
        return  free_index;
    }
 
[...]

PR #1944, opened 2017-05-13
GSdx TC: Peformance improvement by using custom container. Ported erase iterator trick to SW mode.

AlessandroVetere:
This follows #1756.

06/06/2017: Major edit reflecting current status of PR. I will fix errors and clean the PR as soon as possible.
12/06/2017: PR updated.

Some improvements to the existing texture caching system are done:
- I implemented the template class GSFastList as replacement for std::list in GSdx TC, both HW and SW. In GSdx TC HW (similar applies to SW Texture*), the std::list (double linked list) used for Source* and Target* lists is replaced by the custom GSFastList, double linked list built on top of a compact array, which uses a stack for free indexes and a custom struct as list node (with data, index of previous element and index of next element) to manage list traversal, node insert and node deletion.
The data content of the node is generic T and also the initial size of the list can be declared via template.
The list has max size equal to 2^16-1 = 65535, but the head of the list is always present thus only 2^16-2 elements can be added.
The list automatically resizes by doubling it's size (both the array of nodes and the stack, array, of free indexes are realloc'd) when an insert occurs and the list is full.
No resizing on deletion of nodes, thus the size in memory of the list only grows with time.
This list by being compact is not heavily fragmented into memory and thus much more cache friendly than a std::list.
It should be almost as fast as an std::vector if not faster, and it is optimized for the TC tasks.
Beware there is no bounds checking for indexes passed to GSFastList functions.
- I ported the infamous "erase iterator" trick from GSdx TC hardware to software, with serious FPS improvements in the same games that benefit from the trick in hardware mode.
- I became paranoid and hunted down std::list in other parts of the code, replacing it with the data structure that seemed most stuited for the task.

The performance improvement are measured using standard GSdx-SSE4 32 bit D3D11 without any modification to any config.
My machine is i5-2500k@4.3GHz, GTX770 OC, 16GB DDR3-1600, Win7 64bit

|   HW Mode     | Before | After |
|--------|--------|-------|
| ZoE2 - Anubis | 40     | 42.5    |
| BG:DA - Inn| 25     | 29.5    |

|   SW Mode     | Before | After |
|--------|--------|-------|
| ZoE2 - Anubis | 37 | 38    |
| BG:DA - Inn| 55     | 65    |

OGL performances on my machine seemed to be bottlenecked by other factors in ZoE2, while in BG:DA they showed almost the same results of D3D11, but I didn't include those though.
I would be very happy to receive comments and results of other benchmarks if someone is brave enough.

### DIFF
--- a/plugins/GSdx/CMakeLists.txt
+++ b/plugins/GSdx/CMakeLists.txt
@@ -142,6 +142,7 @@ set(GSdxHeaders
     GSDump.h
     GSdx.h
     GSdxResources.h
+    GSFastList.h
     GSFunctionMap.h
     GS.h
     GSLocalMemory.h
--- a/plugins/GSdx/GSCodeBuffer.cpp
+++ b/plugins/GSdx/GSCodeBuffer.cpp
@@ -32,9 +32,9 @@ GSCodeBuffer::GSCodeBuffer(size_t blocksize)
 
 GSCodeBuffer::~GSCodeBuffer()
 {
-	for(list<void*>::iterator i = m_buffers.begin(); i != m_buffers.end(); ++i)
+	for(auto buffer : m_buffers)
 	{
-		vmfree(*i, m_blocksize);
+		vmfree(buffer, m_blocksize);
 	}
 }
 
--- a/plugins/GSdx/GSDevice.cpp
+++ b/plugins/GSdx/GSDevice.cpp
@@ -45,7 +45,7 @@ GSDevice::GSDevice()
 
 GSDevice::~GSDevice()
 {
-	for(auto &t : m_pool) delete t;
+	for(auto t : m_pool) delete t;
 
 	delete m_backbuffer;
 	delete m_merge;
@@ -66,7 +66,7 @@ bool GSDevice::Create(const std::shared_ptr<GSWnd>& wnd)
 
 bool GSDevice::Reset(int w, int h)
 {
-	for(auto &t : m_pool) delete t;
+	for(auto t : m_pool) delete t;
 
 	m_pool.clear();
 
@@ -133,9 +133,9 @@ void GSDevice::Present(GSTexture* sTex, GSTexture* dTex, const GSVector4& dRect,
 
 GSTexture* GSDevice::FetchSurface(int type, int w, int h, bool msaa, int format)
 {
-	GSVector2i size(w, h);
+	const GSVector2i size(w, h);
 
-	for(list<GSTexture*>::iterator i = m_pool.begin(); i != m_pool.end(); ++i)
+	for(auto i = m_pool.begin(); i != m_pool.end(); ++i)
 	{
 		GSTexture* t = *i;
 
@@ -154,9 +154,8 @@ void GSDevice::PrintMemoryUsage()
 {
 #ifdef ENABLE_OGL_DEBUG
 	uint32 pool = 0;
-	for(list<GSTexture*>::iterator i = m_pool.begin(); i != m_pool.end(); i++)
+	for(auto t : m_pool)
 	{
-		GSTexture* t = *i;
 		if (t)
 			pool += t->GetMemUsage();
 	}
--- a/plugins/GSdx/GSFastList.h
+++ b/plugins/GSdx/GSFastList.h
@@ -1,0 +1,277 @@
+/*
+*	Copyright (C) 2017-2017 Alessandro Vetere
+*	Copyright (C) 2007-2009 Gabest
+*	http://www.gabest.org
+*
+*  This Program is free software; you can redistribute it and/or modify
+*  it under the terms of the GNU General Public License as published by
+*  the Free Software Foundation; either version 2, or (at your option)
+*  any later version.
+*
+*  This Program is distributed in the hope that it will be useful,
+*  but WITHOUT ANY WARRANTY; without even the implied warranty of
+*  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+*  GNU General Public License for more details.
+*
+*  You should have received a copy of the GNU General Public License
+*  along with GNU Make; see the file COPYING.  If not, write to
+*  the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA USA.
+*  http://www.gnu.org/copyleft/gpl.html
+*
+*/
+
+#pragma once
+
+template <class T>
+struct Element {
+	T  data;
+	uint16 next_index;
+	uint16 prev_index;
+};
+
+template <class T>
+class FastListIterator;
+
+template <class T>
+class FastList {
+	friend class FastListIterator<T>;
+private:
+	// The index of the first element of the list is m_buffer[0].next_index
+	//     The first Element<T> of the list has prev_index equal to 0
+	// The index of the last element of the list is m_buffer[0].prev_index
+	//     The last Element<T> of the list has next_index equal to 0
+	// All the other Element<T> of the list are chained by next_index and prev_index
+	// m_buffer has dynamic size m_capacity
+	// Due to m_buffer reallocation, the pointers to Element<T> stored into the array
+	//     are invalidated every time Grow() is executed. But FastListIterator<T> is
+	//     index based, not pointer based, and the elements are copied in order on Grow(),
+	//     so there is no iterator invalidation (which is an index invalidation) until 
+	//     the relevant iterator (or the index alone) are erased from the list.
+	// m_buffer[0] is always present as auxiliary Element<T> of the list
+	Element<T>* m_buffer;
+	uint16 m_capacity;
+	uint16 m_free_indexes_stack_top;
+	// m_free_indexes_stack has dynamic size (m_capacity - 1)
+	// m_buffer indexes that are free to be used are stacked here
+	uint16* m_free_indexes_stack;
+
+public:
+	__forceinline FastList() {
+		m_buffer = nullptr;
+		clear();
+	}
+
+	__forceinline ~FastList() {
+		_aligned_free(m_buffer);
+	}
+
+	void clear() {
+		// Initialize m_capacity to 4 so we avoid to Grow() on initial insertions
+		// The code doesn't break if this value is changed with anything from 1 to USHRT_MAX
+		m_capacity = 4;
+
+		// Initialize m_buffer and m_free_indexes_stack as a contiguous block of memory starting at m_buffer
+		// This should increase cache locality and reduce memory fragmentation
+		_aligned_free(m_buffer);
+		m_buffer = (Element<T>*)_aligned_malloc(m_capacity * sizeof(Element<T>) + (m_capacity - 1) * sizeof(uint16), 64);
+		m_free_indexes_stack = (uint16*)&m_buffer[m_capacity];
+
+		// Initialize m_buffer[0], data field is unused but initialized using default T constructor
+		m_buffer[0] = { T(), 0, 0 };
+
+		// m_free_indexes_stack top index is 0, bottom index is m_capacity - 2
+		m_free_indexes_stack_top = 0;
+
+		// m_buffer index 0 is reserved for auxiliary element
+		for (uint16 i = 0; i < m_capacity - 1; i++) {
+			m_free_indexes_stack[i] = i + 1;
+		}
+	}
+
+	// Insert the element in front of the list and return its position in m_buffer
+	__forceinline uint16 InsertFront(const T& data) {
+		if (Full()) {
+			Grow();
+		}
+
+		// Pop a free index from the stack
+		const uint16 free_index = m_free_indexes_stack[m_free_indexes_stack_top++];
+		m_buffer[free_index].data = data;
+		ListInsertFront(free_index);
+		return free_index;
+	}
+
+	__forceinline void push_front(const T& data) {
+		InsertFront(data);
+	}
+
+	__forceinline const T& back() const {
+		return m_buffer[LastIndex()].data;
+	}
+
+	__forceinline void pop_back() {
+		EraseIndex(LastIndex());
+	}
+
+	__forceinline uint16 size() const {
+		return m_free_indexes_stack_top;
+	}
+
+	__forceinline bool empty() const {
+		return size() == 0;
+	}	
+
+	__forceinline void EraseIndex(const uint16 index) {
+		ListRemove(index);
+		m_free_indexes_stack[--m_free_indexes_stack_top] = index;
+	}
+
+	__forceinline void MoveFront(const uint16 index) {
+		if (FirstIndex() != index) {
+			ListRemove(index);
+			ListInsertFront(index);
+		}
+	}
+
+	__forceinline const FastListIterator<T> begin() const {
+		return FastListIterator<T>(this, FirstIndex());
+	}
+
+	__forceinline const FastListIterator<T> end() const {
+		return FastListIterator<T>(this, 0);
+	}
+
+	__forceinline FastListIterator<T> erase(FastListIterator<T> i) {
+		EraseIndex(i.Index());
+		return ++i;
+	}
+
+private:
+	// Accessed by FastListIterator<T> using class friendship
+	__forceinline const T& Data(const uint16 index) const {
+		return m_buffer[index].data;
+	}
+
+	// Accessed by FastListIterator<T> using class friendship
+	__forceinline uint16 NextIndex(const uint16 index) const {
+		return m_buffer[index].next_index;
+	}
+
+	// Accessed by FastListIterator<T> using class friendship
+	__forceinline uint16 PrevIndex(const uint16 index) const {
+		return m_buffer[index].prev_index;
+	}
+
+	__forceinline uint16 FirstIndex() const {
+		return m_buffer[0].next_index;
+	}
+
+	__forceinline uint16 LastIndex() const {
+		return m_buffer[0].prev_index;
+	}
+
+	__forceinline bool Full() const {
+		// The minus one is due to the presence of the auxiliary element
+		return size() == m_capacity - 1;
+	}
+
+	__forceinline void ListInsertFront(const uint16 index) {
+		// Update prev / next indexes to add m_buffer[index] to the chain
+		Element<T>& head = m_buffer[0];
+		m_buffer[index].prev_index = 0;
+		m_buffer[index].next_index = head.next_index;
+		m_buffer[head.next_index].prev_index = index;
+		head.next_index = index;
+	}
+
+	__forceinline void ListRemove(const uint16 index) {
+		// Update prev / next indexes to remove m_buffer[index] from the chain
+		const Element<T>& to_remove = m_buffer[index];
+		m_buffer[to_remove.prev_index].next_index = to_remove.next_index;
+		m_buffer[to_remove.next_index].prev_index = to_remove.prev_index;
+	}
+
+	void Grow() {
+		if (m_capacity == USHRT_MAX) {
+			throw std::runtime_error("FastList size maxed out at USHRT_MAX (65535) elements, cannot grow futhermore.");
+		}
+
+		const uint16 new_capacity = m_capacity <= (USHRT_MAX / 2) ? (m_capacity * 2) : USHRT_MAX;
+
+		Element<T>* new_buffer = (Element<T>*)_aligned_malloc(new_capacity * sizeof(Element<T>) + (new_capacity - 1) * sizeof(uint16), 64);
+		uint16* new_free_indexes_stack = (uint16*)&new_buffer[new_capacity];
+
+		memcpy(new_buffer, m_buffer, m_capacity * sizeof(Element<T>));
+		memcpy(new_free_indexes_stack, m_free_indexes_stack, (m_capacity - 1) * sizeof(uint16));
+		
+		_aligned_free(m_buffer);
+		
+		m_buffer = new_buffer;
+		m_free_indexes_stack = new_free_indexes_stack;
+
+		// Initialize the additional space in the stack
+		for (uint16 i = m_capacity - 1; i < new_capacity - 1; i++) {
+			m_free_indexes_stack[i] = i + 1;
+		}
+
+		m_capacity = new_capacity;
+	}
+};
+
+
+template <class T>
+// This iterator is const_iterator
+class FastListIterator
+{
+private:
+	const FastList<T>* m_fastlist;
+	uint16 m_index;
+
+public:
+	__forceinline FastListIterator(const FastList<T>* fastlist, const uint16 index) {
+		m_fastlist = fastlist;
+		m_index = index;
+	}
+
+	__forceinline bool operator!=(const FastListIterator<T>& other) const {
+		return (m_index != other.m_index);
+	}
+
+	__forceinline bool operator==(const FastListIterator<T>& other) const {
+		return (m_index == other.m_index);
+	}
+
+	// Prefix increment
+	__forceinline const FastListIterator<T>& operator++() {
+		m_index = m_fastlist->NextIndex(m_index);
+		return *this;
+	}
+
+	// Postfix increment
+	__forceinline const FastListIterator<T> operator++(int) {
+		FastListIterator<T> copy(*this);
+		++(*this);
+		return copy;
+	}
+
+	// Prefix decrement
+	__forceinline const FastListIterator<T>& operator--() {
+		m_index = m_fastlist->PrevIndex(m_index);
+		return *this;
+	}
+
+	// Postfix decrement
+	__forceinline const FastListIterator<T> operator--(int) {
+		FastListIterator<T> copy(*this);
+		--(*this);
+		return copy;
+	}
+
+	__forceinline const T& operator*() const {
+		return m_fastlist->Data(m_index);
+	}
+
+	__forceinline uint16 Index() const {
+		return m_index;
+	}
+};
--- a/plugins/GSdx/GSTextureCache.cpp
+++ b/plugins/GSdx/GSTextureCache.cpp
@@ -75,7 +75,7 @@ void GSTextureCache::RemovePartial()
 
 	for (int type = 0; type < 2; type++)
 	{
-		for (auto &t : m_dst[type]) delete t;
+		for (auto t : m_dst[type]) delete t;
 
 		m_dst[type].clear();
 	}
@@ -87,7 +87,7 @@ void GSTextureCache::RemoveAll()
 
 	for(int type = 0; type < 2; type++)
 	{
-		for (auto &t : m_dst[type]) delete t;
+		for (auto t : m_dst[type]) delete t;
 
 		m_dst[type].clear();
 	}
@@ -200,10 +200,9 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 
 	Source* src = NULL;
 
-	list<Source*>& m = m_src.m_map[TEX0.TBP0 >> 5];
+	auto& m = m_src.m_map[TEX0.TBP0 >> 5];
 
-
-	for(list<Source*>::iterator i = m.begin(); i != m.end(); ++i)
+	for(auto i = m.begin(); i != m.end(); ++i)
 	{
 		Source* s = *i;
 
@@ -225,7 +224,7 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 				continue;
 		}
 
-		m.splice(m.begin(), m, i);
+		m.MoveFront(i.Index());
 
 		src = s;
 
@@ -257,10 +256,7 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 		// (Simply not doing this code at all makes a lot of previsouly missing stuff show (but breaks pretty much everything
 		// else.)
 
-		for(list<Target*>::iterator i = m_dst[RenderTarget].begin(); i != m_dst[RenderTarget].end(); ++i)
-		{
-			Target* t = *i;
-
+		for(auto t : m_dst[RenderTarget]) {
 			if(t->m_used && t->m_dirty.empty()) {
 				// Typical bug (MGS3 blue cloud):
 				// 1/ RT used as 32 bits => alpha channel written
@@ -357,9 +353,7 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 			// Unfortunately, I don't have any Arc the Lad testcase
 			//
 			// 1/ Check only current frame, I guess it is only used as a postprocessing effect
-			for(list<Target*>::iterator i = m_dst[DepthStencil].begin(); i != m_dst[DepthStencil].end(); ++i) {
-				Target* t = *i;
-
+			for(auto t : m_dst[DepthStencil]) {
 				if(!t->m_age && t->m_used && t->m_dirty.empty() && GSUtil::HasSharedBits(bp, psm, t->m_TEX0.TBP0, t->m_TEX0.PSM))
 				{
 					GL_INS("TC: Warning depth format read as color format. Pixels will be scrambled");
@@ -445,13 +439,13 @@ GSTextureCache::Target* GSTextureCache::LookupTarget(const GIFRegTEX0& TEX0, int
 
 	Target* dst = NULL;
 
-	for(list<Target*>::iterator i = m_dst[type].begin(); i != m_dst[type].end(); ++i)
-	{
+	auto& list = m_dst[type];
+	for(auto i = list.begin(); i != list.end(); ++i) {
 		Target* t = *i;
 
 		if(bp == t->m_TEX0.TBP0)
 		{
-			m_dst[type].splice(m_dst[type].begin(), m_dst[type], i);
+			list.MoveFront(i.Index());
 
 			dst = t;
 
@@ -476,9 +470,7 @@ GSTextureCache::Target* GSTextureCache::LookupTarget(const GIFRegTEX0& TEX0, int
 		// Depth stencil/RT can be an older RT/DS but only check recent RT/DS to avoid to pick
 		// some bad data.
 		Target* dst_match = nullptr;
-		for(list<Target*>::iterator i = m_dst[rev_type].begin(); i != m_dst[rev_type].end(); ++i) {
-			Target* t = *i;
-
+		for(auto t : m_dst[rev_type]) {
 			if (bp == t->m_TEX0.TBP0) {
 				if (t->m_age == 0) {
 					dst_match = t;
@@ -610,10 +602,8 @@ GSTextureCache::Target* GSTextureCache::LookupTarget(const GIFRegTEX0& TEX0, int
 
 
 #if 0
-	for(list<Target*>::iterator i = m_dst[RenderTarget].begin(); i != m_dst[RenderTarget].end(); i++)
+	for(auto t : m_dst[RenderTarget])
 	{
-		Target* t = *i;
-
 		if(bp == t->m_TEX0.TBP0)
 		{
 			dst = t;
@@ -674,7 +664,8 @@ void GSTextureCache::InvalidateVideoMemType(int type, uint32 bp)
 	if (!CanConvertDepth())
 		return;
 
-	for(list<Target*>::iterator i = m_dst[type].begin(); i != m_dst[type].end(); ++i)
+	auto& list = m_dst[type];
+	for(auto i = list.begin(); i != list.end(); ++i)
 	{
 		Target* t = *i;
 
@@ -684,7 +675,7 @@ void GSTextureCache::InvalidateVideoMemType(int type, uint32 bp)
 					t->m_texture ? t->m_texture->GetID() : 0,
 					t->m_TEX0.TBP0);
 
-			m_dst[type].erase(i);
+			list.erase(i);
 			delete t;
 
 			break;
@@ -707,13 +698,11 @@ void GSTextureCache::InvalidateVideoMem(GSOffset* off, const GSVector4i& rect, b
 	{
 		// Remove Source that have same BP as the render target (color&dss)
 		// rendering will dirty the copy
-		const list<Source*>& m = m_src.m_map[bp >> 5];
-
-		for(list<Source*>::const_iterator i = m.begin(); i != m.end(); )
+		auto& list = m_src.m_map[bp >> 5];
+		for(auto i = list.begin(); i != list.end(); )
 		{
-			list<Source*>::const_iterator j = i++;
-
-			Source* s = *j;
+			Source* s = *i;
+			++i;
 
 			if(GSUtil::HasSharedBits(bp, psm, s->m_TEX0.TBP0, s->m_TEX0.PSM))
 			{
@@ -726,14 +715,11 @@ void GSTextureCache::InvalidateVideoMem(GSOffset* off, const GSVector4i& rect, b
 			// Detect half of the render target (fix snow engine game)
 			// Target Page (8KB) have always a width of 64 pixels
 			// Half of the Target is TBW/2 pages * 8KB / (1 block * 256B) = 0x10
-
-			const list<Source*>& m = m_src.m_map[bbp >> 5];
-
-			for(list<Source*>::const_iterator i = m.begin(); i != m.end(); )
+			auto& list = m_src.m_map[bbp >> 5];
+			for(auto i = list.begin(); i != list.end(); )
 			{
-				list<Source*>::const_iterator j = i++;
-
-				Source* s = *j;
+				Source* s = *i;
+				++i;
 
 				if(GSUtil::HasSharedBits(bbp, psm, s->m_TEX0.TBP0, s->m_TEX0.PSM))
 				{
@@ -749,11 +735,8 @@ void GSTextureCache::InvalidateVideoMem(GSOffset* off, const GSVector4i& rect, b
 			uint32 end_block = GSLocalMemory::m_psm[psm].bn(rect.width(), rect.height(), bp, bw);
 			auto type = RenderTarget;
 
-			for(auto i = m_dst[type].begin(); i != m_dst[type].end(); )
+			for(auto t : m_dst[type])
 			{
-				auto j = i++;
-
-				Target* t = *j;
 				if (t->m_TEX0.TBP0 > bp && t->m_end_block < end_block) {
 					// Haunting ground expect to clean buffer B with a rendering into buffer A.
 					// Situation is quite messy as it would require to extract the data from the buffer A
@@ -784,34 +767,33 @@ void GSTextureCache::InvalidateVideoMem(GSOffset* off, const GSVector4i& rect, b
 	{
 		uint32 page = *p;
 
-		const list<Source*>& m = m_src.m_map[page];
-
-		for(list<Source*>::const_iterator i = m.begin(); i != m.end(); )
+		auto& list = m_src.m_map[page];
+		for(auto i = list.begin(); i != list.end(); )
 		{
-			list<Source*>::const_iterator j = i++;
-
-			Source* s = *j;
+			Source* s = *i;
+			++i;
 
 			if(GSUtil::HasSharedBits(psm, s->m_TEX0.PSM))
 			{
-				uint32* RESTRICT valid = s->m_valid;
-
 				bool b = bp == s->m_TEX0.TBP0;
 
 				if(!s->m_target)
 				{
-					if (m_disable_partial_invalidation && s->m_repeating) {
+					if(m_disable_partial_invalidation && s->m_repeating)
+					{
 						m_src.RemoveAt(s);
-					} else {
+					}
+					else
+					{
+						uint32* RESTRICT valid = s->m_valid;
+
 						// Invalidate data of input texture
 						if(s->m_repeating)
 						{
 							// Note: very hot path on snowbling engine game
-							vector<GSVector2i>& l = s->m_p2t[page];
-
-							for(vector<GSVector2i>::iterator k = l.begin(); k != l.end(); ++k)
+							for(const GSVector2i& k : s->m_p2t[page])
 							{
-								valid[k->x] &= k->y;
+								valid[k.x] &= k.y;
 							}
 						}
 						else
@@ -842,10 +824,10 @@ void GSTextureCache::InvalidateVideoMem(GSOffset* off, const GSVector4i& rect, b
 
 	for(int type = 0; type < 2; type++)
 	{
-		for(list<Target*>::iterator i = m_dst[type].begin(); i != m_dst[type].end(); )
+		auto& list = m_dst[type];
+		for(auto i = list.begin(); i != list.end(); )
 		{
-			list<Target*>::iterator j = i++;
-
+			auto j = i++;
 			Target* t = *j;
 
 			// GH: (I think) this code is completely broken. Typical issue:
@@ -867,7 +849,7 @@ void GSTextureCache::InvalidateVideoMem(GSOffset* off, const GSVector4i& rect, b
 				}
 				else
 				{
-					m_dst[type].erase(j);
+					list.erase(j);
 					GL_CACHE("TC: Remove Target(%s) %d (0x%x)", to_string(type),
 								t->m_texture ? t->m_texture->GetID() : 0,
 								t->m_TEX0.TBP0);
@@ -962,12 +944,8 @@ void GSTextureCache::InvalidateLocalMem(GSOffset* off, const GSVector4i& r)
 	// It works for all the games mentioned below and fixes a couple of other ones as well
 	// (Busen0: Wizardry and Chaos Legion).
 	// Also in a few games the below code ran the Grandia3 case when it shouldn't :p
-	for(list<Target*>::iterator i = m_dst[RenderTarget].begin(); i != m_dst[RenderTarget].end(); )
+	for(auto t : m_dst[RenderTarget])
 	{
-		list<Target*>::iterator j = i++;
-
-		Target* t = *j;
-
 		if (t->m_TEX0.PSM != PSM_PSMZ32 && t->m_TEX0.PSM != PSM_PSMZ24 && t->m_TEX0.PSM != PSM_PSMZ16 && t->m_TEX0.PSM != PSM_PSMZ16S)
 		{
 			if(GSUtil::HasSharedBits(bp, psm, t->m_TEX0.TBP0, t->m_TEX0.PSM))
@@ -1075,17 +1053,20 @@ void GSTextureCache::InvalidateVideoMemSubTarget(GSTextureCache::Target* rt)
 	if (!rt)
 		return;
 
-	for(list<Target*>::iterator i = m_dst[RenderTarget].begin(); i != m_dst[RenderTarget].end(); ) {
-		list<Target*>::iterator j = i++;
-		Target* t = *j;
+	auto& list = m_dst[RenderTarget];
 
-		if ((t->m_TEX0.TBP0 > rt->m_TEX0.TBP0) && (t->m_end_block < rt->m_end_block) && (t->m_TEX0.TBW == rt->m_TEX0.TBW)
+	for(auto i = list.begin(); i != list.end(); ) {
+		Target* t = *i;
+
+		if((t->m_TEX0.TBP0 > rt->m_TEX0.TBP0) && (t->m_end_block < rt->m_end_block) && (t->m_TEX0.TBW == rt->m_TEX0.TBW)
 				&& (t->m_TEX0.TBP0 < t->m_end_block)) {
 			GL_INS("InvalidateVideoMemSubTarget: rt 0x%x -> 0x%x, sub rt 0x%x -> 0x%x",
 					rt->m_TEX0.TBP0, rt->m_end_block, t->m_TEX0.TBP0, t->m_end_block);
 
-			m_dst[RenderTarget].erase(j);
+			i = list.erase(i);
 			delete t;
+		} else {
+			++i;
 		}
 	}
 }
@@ -1095,19 +1076,20 @@ void GSTextureCache::IncAge()
 	int maxage = m_src.m_used ? 3 : 30;
 
 	// You can't use m_map[page] because Source* are duplicated on several pages.
-	for(hash_set<Source*>::iterator i = m_src.m_surfaces.begin(); i != m_src.m_surfaces.end(); )
+	for(auto i = m_src.m_surfaces.begin(); i != m_src.m_surfaces.end(); )
 	{
-		hash_set<Source*>::iterator j = i++;
-
-		Source* s = *j;
+		Source* s = *i;
 
 		if(s->m_shared_texture) {
 			// Shared textures are temporary only added in the hash set but not in the texture
 			// cache list therefore you can't use RemoveAt
-			m_src.m_surfaces.erase(s);
+			i = m_src.m_surfaces.erase(i);
 			delete s;
-		} else if(++s->m_age > maxage) {
-			m_src.RemoveAt(s);
+		} else {
+			++i;
+			if (++s->m_age > maxage) {
+				m_src.RemoveAt(s);
+			}
 		}
 	}
 
@@ -1121,11 +1103,10 @@ void GSTextureCache::IncAge()
 
 	for(int type = 0; type < 2; type++)
 	{
-		for(list<Target*>::iterator i = m_dst[type].begin(); i != m_dst[type].end(); )
+		auto& list = m_dst[type];
+		for(auto i = list.begin(); i != list.end(); )
 		{
-			list<Target*>::iterator j = i++;
-
-			Target* t = *j;
+			Target* t = *i;
 
 			// This variable is used to detect the texture shuffle effect. There is a high
 			// probability that game will do it on the current RT.
@@ -1138,12 +1119,14 @@ void GSTextureCache::IncAge()
 
 			if(++t->m_age > maxage)
 			{
-				m_dst[type].erase(j);
+				i = list.erase(i);
 				GL_CACHE("TC: Remove Target(%s): %d (0x%x) due to age", to_string(type),
 							t->m_texture ? t->m_texture->GetID() : 0,
 							t->m_TEX0.TBP0);
 
 				delete t;
+			} else {
+				++i;
 			}
 		}
 	}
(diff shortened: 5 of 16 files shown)


ramapcsx2:
I'm totally out of league with this impressive work but I noticed this one thing:

> Beware there is no bounds checking for indexes passed to FastList functions, nor is checked whether the list is full before adding an element, but the list size should be sufficient for every game out in the wild.

Let me advise you to never assume the boundaries of things going wrong in GSdx.
If at all possible, try to have an indicator of this list overflowing. Maybe a special end node? Maybe a counter?
If it really can't be done, maybe make a debug version of this and then test the hell out of it.

It is nice work though!

AlessandroVetere:
> If at all possible, try to have an indicator of this list overflowing. Maybe a special end node? Maybe a counter?

It is possible indeed by checking before InsertFront if the stack of the free indexes is empty and before Erase if it is full.
Maybe this could be added as an ifdef to be enabled only when debug mode is enabled, and print error messages in case an exception happens.

plugins/GSdx/GSTextureCache.h
```diff
@@ -152,7 +152,7 @@ class GSTextureCache
 		GSTexture* m_from_target;
 		GIFRegTEX0 m_layer_TEX0[7]; // Detect already loaded value
 		// Keep a GSTextureCache::m_map iterator to allow fast erase
-		std::array<uint16, MAX_PAGES> m_erase_it;
+		uint16 m_erase_it[MAX_PAGES];
```
gregory38: Why ? It shouldn't have any overhead.

gregory38:
You need to clean this PR first.
* you need to rebase on master to avoid the merge commit.
* you need to remove properly the pool stuff. I don't want a commit that adds stuff and a later commit that remove it.
* you need to use the same style as the file. (bonus: no braces for single line is better)
```
if ()
{
}
else
{
}
```
or
```
if () {
} else{
}
```

AlessandroVetere:
Will do.
Also forgot the fast path for InsertFront my bad.
I tried using FastList for Target* too, without pooling system as Target objects are quite small, but it didn't lead to visible improvements. At least in the few games I tried. Would you prefer to have FastList also for Target(s), possibly with custom size?

plugins/GSdx/GSTextureCache.h
AlessandroVetere: To uniform usage of arrays vs std::array in the class and also because the functionalities of std::array aren't used (bounds checking and static init).

gregory38: Well, personally I like the std::array. Features could still be useful in the future. It would be better to migrate `C` array to `C++` array instead. In order to avoid massive change for small gain, I prefer that new code uses them.

gregory38:
Number of target is rather small for an obvious reason VRAM requirement. Most games are likely max 10 targets. Some games create lots of target (100-1000) but they often crash with out-of-memory.
A target is around `5MB * upscaling^2`

plugins/GSdx/GSTextureCache.h
```diff
@@ -29,6 +29,82 @@ class GSTextureCache
...
+
+	template <class T>
+	class FastList {
+	public:
+		// Seems like 512 is a reasonable upper bound for the size (ZoE2 goes just below 500), maybe 1024 better
+		static const uint16 SIZE = 512;
+		// The last element of the list has next index equal to END
+		static const uint16 END = SIZE;
+		// m_buffer[0] is the head of the list
+		Element<T> m_buffer[SIZE];
```
turtleli: Does this have to be a fixed size? Would a vector work?

AlessandroVetere: A vector does invalidate all the iterators when the underlying array is resized thus would invalidate the iterators saved in the Source objects during list insertion (for future fast deletion).
So yes I cannot imagine a compact storage with iterators that do not invalidate on list operations that has not fixed size.
Maybe a hierarchy std::list -> m_buffer[SIZE] dynamically managed could work in principle, using a cache friendly SIZE and ensuring that the number or nodes of the std::list are minimized.
Logic would become unfriendly though.

AlessandroVetere: Nevermind I was joking.
I implemented a automatically growing FastList which does not invalidate iterator simply using realloc() and increasing the size of the two underlying arrays.
Benchmarks shows that there is not much decrease in performance wrt the statically allocated / fixed size one, so the latter is for sure a better choice. I need to clean this PR.

AlessandroVetere:
gregory38 I will fix the mess don't worry.
Basically I did the following:
- Thanks to turtleli I implemented the resizable FastList, and moved it in a separate header file named GSFastList.h
- I ported the "erase iterator" trick also to software mode texture cache system.
- I used FastList also in software mode for Texture*
- I used FastList also for Target* in hardware mode
- Hunted down std::list in other parts of the code (maybe only useful for GSDevice and GSDirtyRect)

Performance improvement in software mode is real, BG:DA shows +10 fps, ZoE2 +0,5~1 fps, all other games benefit a bit.
Hardware mode didn't like passing to dynamic memory instead of stack memory to save texture pointers but after some optimization it didn't left much of a impact.

gregory38:
Please ping me when you have a clean PR. Thanks you. (don't hesitate to ask any question)

AlessandroVetere:
I basically didn't have time to fix neither the code nor the PR, but I should be able to do both this weekend.
I think I will rebase on master and then redo the commits in a coherent way.
Also I have a question: is _aligned_realloc available under clang?
It seems the only compiler to throw error: "/home/travis/build/PCSX2/pcsx2/plugins/GSdx/./GSFastList.h:57:27: error: use of undeclared identifier '_aligned_realloc'".
The other errors under g++ or clang will be fixed easily.

AlessandroVetere:
PR cleaned.
I think _aligned_realloc will still make build under clang fail.

woj1993:
@AlessandroVetere You have marge conflict.

AlessandroVetere:
@woj1993 Should I resolve the conflict? It would result in a "merge commit" which Gregory38 told me to avoid. I'm new to github PR system though, so I don't know what to do exactly.
@Gregory38 I cleaned up the PR, almost.

mirh:
I _think_ that rebasing to over 11ed4de0aa04c52c6ab0c3f4d2d67fcbd1e790a5 should be enough?

AlessandroVetere:
@mirh I'll look into that tonight thanks for the tip. I think I can also directly rebase on top of the latest commit on master branch, just to be sure.

AlessandroVetere:
@gregory38 PR cleaned, AppVeyor build succeeded. Travis doesn't show up though.

AlessandroVetere:
MoveFront(index) could be implemented in a much more efficient way by swapping m_buffer[m_buffer[0].next_index].data with m_buffer[index].data, instead of doing the ListRemove(index) followed by ListInsertFront(index).
I wait until the work done so far is validated to do further modifications though.

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,155 @@
...
+// Class size in bytes:
+// 2 * sizeof(uint16) + sizeof(Element<T>*) + sizeof(uint16*),
+// The Element<T> array m_buffer has dynamic size: CURRENT_SIZE,
+// The uint16 array m_free_indexes_stack has dynamic size: CURRENT_SIZE - 1,
+// Where CURRENT_SIZE starts from INITIAL_SIZE and grows as needed
+template <class T, uint16 INITAL_SIZE = 4>
+class FastList {
+public:
+	// The last Element<T> of the list has next_index equal to END
+	uint16 END;
```
gregory38: Call the variable `m_end` instead.

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,155 @@
...
+	// m_buffer[0] is the head of the list, and m_buffer has END elements
+	Element<T>* m_buffer;
+
+	FastList() {
+		m_buffer = NULL;
```
gregory38: use `nullptr`. Same for all occurences

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	__forceinline void MoveFront(const uint16 index) {
+		if (m_buffer[0].next_index != index) {
+			ListRemove(index);
+			ListInsertFront(index);
+		}
+	}
+private:
+	uint16 m_free_indexes_stack_top;
```
gregory38: move private `m_*` variable at the top. This way, it is easier to find them.

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,155 @@
...
+
+template <class T>
+struct Element {
+	T  data;
+	uint16 next_index;
+	uint16 prev_index;
+};
+
+// Class size in bytes:
+// 2 * sizeof(uint16) + sizeof(Element<T>*) + sizeof(uint16*),
```
gregory38:
```
// Class size in bytes:
// 2 * sizeof(uint16) + sizeof(Element<T>*) + sizeof(uint16*),
```
I don't see the purpose of this comment.

plugins/GSdx/GSFastList.h
gregory38: Actually, is `END` really useful ? Couldn't you use the head (i.e. 0) as special index?

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -75,9 +75,13 @@ void GSTextureCache::RemovePartial()
 
 	for (int type = 0; type < 2; type++)
 	{
-		for (auto &t : m_dst[type]) delete t;
-
-		m_dst[type].clear();
+		FastTargetPtrList& list = m_dst[type];
```
gregory38: Why not provide an iterator for your new container ?

plugins/GSdx/GSFastList.h
AlessandroVetere: It's kind of a leftover from design phase: I can remove it.

plugins/GSdx/GSFastList.h
AlessandroVetere: Ok for the renaming.
The variable I think it's useful also to keep track of the size of the list without adding another class member.
Otherwise it could be used a constant value like 0 (unused "next_index") for the next_index of the last element of the list, and keep a "m_size" uint16 variable for the size.

AlessandroVetere: Ok it's much much better to use constant value 0 for next_index of last element and rename END to m_size, using it only to manage Full() and Grow().

plugins/GSdx/GSTextureCache.cpp
AlessandroVetere: I thought of that, but being paranoid I didn't do so: would it introduce overhead? (I think not).

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,155 @@
...
+		memcpy(new_free_indexes_stack, (uint16*)&m_buffer[END], (END - 1) * sizeof(uint16));
+		m_free_indexes_stack = new_free_indexes_stack;
+
+		// Replace the END marker with the updated one. This is the easiest way and does not require additional class members
+		for (uint16 i = m_buffer[0].next_index; i != END; i = m_buffer[i].next_index) {
+			if (m_buffer[i].next_index == END) {
+				m_buffer[i].next_index = NEW_END;
+				break;
+			}
+		}
```
AlessandroVetere: By using constant value for last next_index, the above piece of code (140-146) could be removed.

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,155 @@
...
+		// Stack top is 0, bottom is END - 2
+		m_free_indexes_stack_top = 0;
+		// Index 0 reserved for head
+		for (uint16 i = 0; i < END - 1; i++) {
+			m_free_indexes_stack[i] = i + 1;
+		}
+	}
+
+	__forceinline bool Empty() {
+		return m_free_indexes_stack == 0;
```
AlessandroVetere: Obviously it's `m_free_indexes_stack_top` and not `m_free_indexes_stack` that should be 0.

AlessandroVetere:
@gregory38 I managed to fix what you requested, leaving the iterator implementation aside (it would require a bit of work). What isn't working now is `_aligned_realloc` under Linux (both clang and gcc). Can something be done to fix the problem? As I can see `_aligned_malloc` and `_aligned_free` are fully functional, and they are of the same family.

turtleli:
`_aligned_malloc` and `_aligned_free` don't actually exist on Linux - they work in this case due to macros/compatibility functions added to GSdx.

Does `_aligned_*` actually provide any performance benefit in this case? I think it should be possible to  just use two vectors - one to hold `Element<T>*`, the other to hold the indices of any removed elements.

gregory38:
There are available on Linux (C11) but they weren't compatible with GCC 4.9 ASAN hence the macro. However reallloc doesn't exists in the standard. Anyway, I think vector should be as efficient.

I'm leaving for 10 days of vacation :) It gives you time this way to look at iterator. Iterators are quite light so I don't think there is any overhead.

AlessandroVetere:
@turtleli Didn't test if `_aligned_*` provides significant benefits.

I'll try using only `_aligned_malloc`, `memcpy`, and `_aligned_free` to replace `_aligned_realloc` in a dirty way.
The only problem would be a slightly average execution time for the `Grow()` function, which isn't critical as it's both amortized and never called after the first warmup of the texture cache, and more importantly a slightly higher memory fragmentation.

I'll try also the plain `realloc` and `free` which should be available under any platform (but makes you lose alignment), and maybe it's better to lose alignment than continuous reallocation.

I would avoid using `vector` as I don't need their advanced capabilities over plain arrays, and I prefer to get the minimum overhead possible (which for standard library containers I'm not able to precisely evaluate in advance).


@gregory38 Have a nice holiday! 
I'll look forward to implement the `iterator`

turtleli:
> I would avoid using vector as I don't need their advanced capabilities over plain arrays, and I prefer to get the minimum overhead possible (which for standard library containers I'm not able to precisely evaluate in advance).

Well, main advantage is that it would likely be simpler to implement, and there probably isn't much additional performance overhead (if there is any) compared to your current approach.

AlessandroVetere:
**Updates**
- Now I have a building branch, which have some commits to be squashed though.
- To replace `_aligned_realloc` I decided to use the `_aligned_malloc`, `memcpy` and `_aligned_free` approach to keep memory aligned, as `FastList` linear scan starts by accessing `m_buffer[0]` so alignment can minimize cache misses (the enemy we are fighting using this custom container).
- Instead of implementing `iterator`, I revamped the way the list is accessed and managed.
Here's a sample of the new usage:
    ```c++
    for (uint16 i = list.First(); i != list.End(); i = list.Next(i)) {
        T data = list.Data(i);   
    }
    ```

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,162 @@
...
+		// Update prev / next indexes to remove m_buffer[index] from the chain
+		const Element<T>& to_remove = m_buffer[index];
+		m_buffer[to_remove.prev_index].next_index = to_remove.next_index;
+		if (to_remove.next_index != End()) {
+			m_buffer[to_remove.next_index].prev_index = to_remove.prev_index;
+		}
+	}
+
+	void Grow() {
+		if (m_size == 65535) {
```
ssakash: Maybe `` _UI16_MAX``/ ``USHRT_MAX`` macro could be used here to make the intention more clear? ``65535`` might just seem like a magic number at first sight.

plugins/GSdx/GSTextureCache.h
```diff
@@ -104,11 +105,12 @@ class GSTextureCache
 		void Update();
 	};
 
+	using FastSourcePtrList = FastList<Source*, 4>;
```
ssakash: Maybe a type definition might be more suitable? (I'm unsure, I've never really seen anyone exploiting the ``using`` keyword this way)

```cpp
typedef FastList<Source*, 4> FastSourcePtrList;
```

plugins/GSdx/GSTextureCache.h
```diff
@@ -120,10 +122,11 @@ class GSTextureCache
 		void RemoveAt(Source* s);
 	};
 
+	using FastTargetPtrList = FastList<Target*, 4>;
```
ssakash: Ditto.

plugins/GSdx/GSTextureCacheSW.h
```diff
@@ -54,10 +56,11 @@ class GSTextureCacheSW
 		bool Save(const string& fn, bool dds = false) const;
 	};
 
+	using FastTexturePtrList = FastList<Texture*, 4>;
```
ssakash: Ditto.

plugins/GSdx/GSFastList.h
AlessandroVetere: Sure, I didn't search for the correct constant. I'll use one of the two.

plugins/GSdx/GSTextureCache.h
AlessandroVetere: They are equivalent in this case, but if `using` is never used elsewhere in the code, I'll also use `typedef`.

danilaml: @ssakash what do you mean by exploiting? It's one of the intended usages of C++11 `using` - better replacement for `typedef` with saner semantics (see typedefing function ptrs) and better support for templates. http://en.cppreference.com/w/cpp/language/type_alias

ssakash: > what do you mean by exploiting?

I just felt awkward writing ``I have never seen anyone using using this way`` and hence I went with ``exploiting`` instead. (FWIW, exploiting doesn't specifically mean to unfairly use something, it could also be used to depict efficient usage of a certain thing)

> They are equivalent in this case

> It's one of the intended usages of C++11 using - better replacement for typedef with saner semantics (see typedefing function ptrs) and better support for templates. http://en.cppreference.com/w/cpp/language/type_alias

Oh cool, then I'm totally fine with exploiting ``using`` this way. (Time to find another synonym for using 😕)

danilaml: >FWIW, exploiting doesn't specifically mean to unfairly use something, it could also be used to depict efficient usage of a certain thing

Well, in the context of C++ it could easily mean using some clever hack to do something unexpected (like what the whole template meta-programming turned out to be) ;P
>Time to find another synonym for using
😄

plugins/GSdx/GSFastList.h
danilaml: `_UI16_MAX` is non-standard AFAIK. Also http://en.cppreference.com/w/cpp/types/numeric_limits could be used, though if `USHRT_MAX` is already included, might be easier to use that.

plugins/GSdx/GSCodeBuffer.cpp
```diff
@@ -32,7 +32,7 @@ GSCodeBuffer::GSCodeBuffer(size_t blocksize)
 
 GSCodeBuffer::~GSCodeBuffer()
 {
-	for(list<void*>::iterator i = m_buffers.begin(); i != m_buffers.end(); ++i)
+	for(std::vector<void*>::iterator i = m_buffers.begin(); i != m_buffers.end(); ++i)
```
gregory38: a C++11 range loop will be better as it will work for all containers.

plugins/GSdx/GSDevice.cpp
```diff
@@ -135,7 +135,7 @@ GSTexture* GSDevice::FetchSurface(int type, int w, int h, bool msaa, int format)
 {
 	GSVector2i size(w, h);
 
-	for(list<GSTexture*>::iterator i = m_pool.begin(); i != m_pool.end(); ++i)
+	for(std::deque<GSTexture*>::iterator i = m_pool.begin(); i != m_pool.end(); ++i)
```
gregory38: range loop and same below

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,162 @@
...
+#pragma once
+
+template <class T>
+struct Element {
+	T  data;
+	uint16 next_index;
+	uint16 prev_index;
+};
+
+template <class T, uint16 INITIAL_SIZE = 4>
```
gregory38: INITIAL_SIZE could be a constructor parameter instead. I don't understand why you use a template here.

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,162 @@
...
+		m_buffer[0].prev_index = 0;
+		// Stack top is 0, bottom is END - 2
+		m_free_indexes_stack_top = 0;
+		// Index 0 reserved for head
+		for (uint16 i = 0; i < m_size - 1; i++) {
+			m_free_indexes_stack[i] = i + 1;
+		}
+	}
+
+	__forceinline uint16 First() const {
```
gregory38: A real iterator will be mucher nicer.

Here an example of the GNU STL list iterator.
```
  /**
   *  @brief A list::iterator.
   *
   *  All the functions are op overloads.
  */
  template<typename _Tp>
    struct _List_iterator
    {
      typedef _List_iterator<_Tp>                _Self;
      typedef _List_node<_Tp>                    _Node;

      typedef ptrdiff_t                          difference_type;
      typedef std::bidirectional_iterator_tag    iterator_category;
      typedef _Tp                                value_type;
      typedef _Tp*                               pointer;
      typedef _Tp&                               reference;

      _List_iterator() _GLIBCXX_NOEXCEPT
      : _M_node() { }

      explicit
      _List_iterator(__detail::_List_node_base* __x) _GLIBCXX_NOEXCEPT
      : _M_node(__x) { }

      _Self
      _M_const_cast() const _GLIBCXX_NOEXCEPT
      { return *this; }

      // Must downcast from _List_node_base to _List_node to get to value.
      reference
      operator*() const _GLIBCXX_NOEXCEPT
      { return *static_cast<_Node*>(_M_node)->_M_valptr(); }

      pointer
      operator->() const _GLIBCXX_NOEXCEPT
      { return static_cast<_Node*>(_M_node)->_M_valptr(); }

      _Self&
      operator++() _GLIBCXX_NOEXCEPT
      {
	_M_node = _M_node->_M_next;
	return *this;
      }

      _Self
      operator++(int) _GLIBCXX_NOEXCEPT
      {
	_Self __tmp = *this;
	_M_node = _M_node->_M_next;
	return __tmp;
      }

      _Self&
      operator--() _GLIBCXX_NOEXCEPT
      {
	_M_node = _M_node->_M_prev;
	return *this;
      }

      _Self
      operator--(int) _GLIBCXX_NOEXCEPT
      {
	_Self __tmp = *this;
	_M_node = _M_node->_M_prev;
	return __tmp;
      }

      bool
      operator==(const _Self& __x) const _GLIBCXX_NOEXCEPT
      { return _M_node == __x._M_node; }

      bool
      operator!=(const _Self& __x) const _GLIBCXX_NOEXCEPT
      { return _M_node != __x._M_node; }

      // The only member points to the %list element.
      __detail::_List_node_base* _M_node;
    };
```

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,162 @@
...
+		_aligned_free(m_buffer);
+	}
+
+	void Clear() {
+		// Initialize END
+		m_size = INITIAL_SIZE;
+		// Initialize arrays
+		_aligned_free(m_buffer);
+		m_buffer = (Element<T>*)_aligned_malloc(m_size * sizeof(Element<T>) + (m_size - 1) * sizeof(uint16), 64);
+		m_free_indexes_stack = (uint16*)&m_buffer[m_size];
```
gregory38: If I understand right you do a single allocation but cut the allocated memory in 2 parts. I don't understand why. It would be so much easier to do 2 separates allocations here.

gregory38:
I did  a quick review.

```
Instead of implementing iterator, I revamped the way the list is accessed and managed.
Here's a sample of the new usage:

for (uint16 i = list.First(); i != list.End(); i = list.Next(i)) {
    T data = list.Data(i);   
}
```
I'm still prefering the iterator way. So you could write directly.
```
for (const auto &data : list)
   ...
```

Note: you need both a const and a non-const iterator

plugins/GSdx/GSFastList.h
AlessandroVetere: It's just because FastList is only used in an array, so the default constructor would be the only one used. Using template to pass parameter you can kind of override the default constructor behaviour.

plugins/GSdx/GSFastList.h
AlessandroVetere: Just to keep memory as compact as possible and exploit cache blocks alignment at max.

plugins/GSdx/GSFastList.h
AlessandroVetere: The fact is that the C++11 range loops would be only used in the part of the code where they are already used now (which is only when looping on the lists to destroy Source objects if I remember well), in all the other parts of the code looping using iterators is needed because you need to have the iterator to the element to perform operations (like MoveFront or Erase).
So introducing iterators would only shorten one usage of the list, while leaving unchanged (ok, instead of using the helper functions you can do ++ operator) most of the code.

AlessandroVetere:
The problem of that kind of looping is that is would be used only once in the code, because most of the time you still need the iterator to the element to perform operations like MoveFront or Erase.
So introducing iterator wouldn't give much benefit to the code, IMHO, just the shortening of one loop (the one which is using already range loop in current master branch) and the usage of ++ operator to advance to next element.

plugins/GSdx/GSFastList.h
gregory38: Ok. My initial feeling was that list will be too big to fit in a single cache line. But we have room for 6 entries in 32 bits. And 4 in 64 bits. Maybe it could explain why you set 4 as an initial value.

plugins/GSdx/GSFastList.h
gregory38: Tbh, I think you shouldn't bother with an initial size parameter. If we need bigger list, you will have severals reallocation the first frame and then you're done. 0 impacts on speed.

plugins/GSdx/GSFastList.h
gregory38: Range loop was only an example. The idea is to have the same API for all containers. If later we want to use STL stuff to erase element from the list, it is still possible. As bare minimum uses the same semantics as STL (`begin()`, `end()`...)

plugins/GSdx/GSFastList.h
AlessandroVetere: In fact this is a leftover from when the list had fixed size. I agree that initial size management could be removed.

plugins/GSdx/GSFastList.h
AlessandroVetere: I'll then adapt the GNU list::iterator implementation to FastList, but I wouldn't use STL stuff to operate on FastList anyway 😀
There may be hidden contracts or something obscure that could break the behaviour.
There is indeed a problem: in a linked list the iterator having a list node (value and pointer to next node) alone is enough to iterate on the list.
On FastList you need the index of the array and the reference to the underlying array.
Easily you can make a struct containing both, but that would be extremely inefficient.
Do you foresee any better solution?

gregory38: Oh, I just understood what you mean about the start of the array. I think you iterator class just need to contain the start pointer and the index of the current node (extend constructor). Iterator is only a wrapper that overload basic operator.

AlessandroVetere: std::vector<>::iterator is basically a pointer to the position in the array and the ++ operator is implemented as ++ on the pointer itself given the memory contiguity of the elements stored in the vector.
About FastList, what I'm trying to say is that now I have 2 bytes index to the element and use those to cycle and store in Source.
If I make an iterator I cannot go below 6 bytes of memory because the struct should contain both the index (2 bytes) and the start pointer (4 bytes or 8 bytes in 64 bits).
It's a huge difference, which I mean would destroy one of the best thing of this container (2 bytes to save the """"iterator"""" to the node, but usable only as parameter to the list itself).

gregory38: Sorry, I fail to understand you. Without the iterator, you get an index right (2 B) and then you lookup the buffer (4B pointer). With iterator, you should have the same. The pointer to the buffer + an index.

As a side note, maybe it should be possible to replace the absolute index by a relative offset.

AlessandroVetere: What you said it's correct, but if I'm using an iterator and incrementing, I have to create e new iterator that copies 2B of index and 4B of buffer ptr. While without iterator I only update 2B of index.

gregory38: Ok. I'm not sure it would be an issue as the value must be stored in a register anyway. Let's keep this way. Could you squash the relevant commit. And also add a comment to explain the single allocation for both stack and data.

gregory38: @turtleli what is your opinion ?

AlessandroVetere: I'll fix the PR by tomorrow. Though maybe some comments would be lost so it's better to move the discussion in the main section.

AlessandroVetere:
I updated the PR, squashing the commits and applying all the suggested changes (FastList iterator excluded).

woj1993:
@gregory38 What do you think?

plugins/GSdx/GSFastList.h
gregory38: Hello, 

Sorry to get back to the iterator discussion. But the iterator is a local variable so saved in stack, the stack won't cache miss. Besides the iterators is only constructed once in the loop init. Or did I miss something.

AlessandroVetere: Let's say we have an iterator like this:

```C++
struct iterator {
	Element<T>* m_buffer;
	uint16 index;
}
```

then of course if you update `index` in place, only 2B are updated each iteration.
But if you need to save the `iterator` for future usage, like it is the case for `GSTextureCache`, you need to copy the whole struct (2B + 4B) becase the one used for looping is going to be updated at the next iteration.
We can use an hybrid technique though, use `iterator` for looping to enable `C++11 range loop syntax` and save only the `uint16 index` as kind of `iterator` for future removal.

gregory38: I like your hybrid proposal 👍 A standard iterator for looping so you don't need to update all loop code in the texture cache. But keeping only an index for fast removal code.

lightningterror:
Also @ssakash , @turtleli can you take a look at it as well again.

AlessandroVetere:
Working on the iterator, I'm using a conservative approach like:
```C++
FastList<Source*>& m = m_src.m_map[TEX0.TBP0 >> 5];
for (auto i = m.cbegin(); i != m.cend(); ++i)
{
	Source* s = *i;
        //...
        m.MoveFront(i.Index());
}
```
So now `FastList` exposes an iterator `FastListIterator` (which enables range loops, here not used as iterator, really only the `uint16` index, is necessary) but the `FastList` functions that operates on the list (both public and private) still accept only `uint16` indexes as parameters.
So the `FastListIterator` exposes it's `m_index` with the method `Index()`, which returns it.
This approach has minimal added costs to the bare metal solution.
Is the const iterator enough? The non const iterator would never be used anyway, at least in the current usage of `FastList`.
It would be interesting though to try using `FastList` whenever in the `pcsx2` code there are similar needs as those of `GSTextureCache`, as I think it's speed and simplicity (thanks also to gregory for #1756) are quite impressive for being an handcrafted solution.

plugins/GSdx/GSTextureCacheSW.cpp
```diff
@@ -121,21 +115,21 @@ void GSTextureCacheSW::InvalidatePages(const uint32* pages, uint32 psm)
...
 	{
-		l.clear();
+		l.Clear();
 	}
 }
 
 void GSTextureCacheSW::IncAge()
 {
-	for(hash_set<Texture*>::iterator i = m_textures.begin(); i != m_textures.end(); )
+	for(hash_set<Texture*>::const_iterator i = m_textures.cbegin(); i != m_textures.cend(); )
```
gregory38: I hate to be that guy. But could use `const auto` instead. And others iterators

plugins/GSdx/GSTextureCacheSW.cpp
```diff
@@ -89,23 +83,23 @@ void GSTextureCacheSW::InvalidatePages(const uint32* pages, uint32 psm)
...
 			{
 				uint32* RESTRICT valid = t->m_valid;
 
 				if(t->m_repeating)
 				{
-					vector<GSVector2i>& l = t->m_p2t[page];
+					const vector<GSVector2i>& l = t->m_p2t[page];
 
-					for(vector<GSVector2i>::iterator j = l.begin(); j != l.end(); ++j)
+					for(const auto& j : l)
```
gregory38: Maybe the intermediate variable `l` is useless now ?

plugins/GSdx/GSdx.vcxproj
```diff
@@ -210,6 +210,7 @@
     <ClInclude Include="GSDrawScanlineCodeGenerator.h" />
     <ClInclude Include="GSDump.h" />
     <ClInclude Include="GSdx.h" />
+    <ClInclude Include="GSFastList.h" />
```
gregory38: Could you add it to CMakeList.txt file too (or did I miss it).

plugins/GSdx/GSDevice.cpp
```diff
@@ -133,9 +133,9 @@ void GSDevice::Present(GSTexture* sTex, GSTexture* dTex, const GSVector4& dRect,
 
 GSTexture* GSDevice::FetchSurface(int type, int w, int h, bool msaa, int format)
 {
-	GSVector2i size(w, h);
+	const GSVector2i size(w, h);
 
-	for(list<GSTexture*>::iterator i = m_pool.begin(); i != m_pool.end(); ++i)
+	for(std::deque<GSTexture*>::const_iterator i = m_pool.cbegin(); i != m_pool.cend(); ++i)
```
gregory38: I don't remember the performance impact of this code. But I think ideally, m_pool shall be split into an array of pool. The rational is that type/format/msaa take few value. Beside I think it would be nice to have a different profile based on m_type.

RenderTarget are often quite expensive in memory, so we don't want too many of them. Whereas texture are smaller so it make sense too keep more pool for them.

Finally, a smaller pool would allow to use a basic vector.

I'm fine with current code. It was just a general comment.

plugins/GSdx/GSTextureCacheSW.cpp
AlessandroVetere: Right

plugins/GSdx/GSTextureCacheSW.cpp
AlessandroVetere: Will update

plugins/GSdx/GSdx.vcxproj
AlessandroVetere: No I missed CMakeList.txt! I will update it.

plugins/GSdx/GSDevice.cpp
AlessandroVetere: Ideally we can have only lookup for `size` if we keep the superset of `type`, `format` and `msaa` values as different pools, as current lookup code clearly says:
```C++
		if(t->GetType() == type && t->GetFormat() == format && t->GetSize() == size && t->IsMSAA() == msaa)
		// ...
```
Also we can rework `AgePool()` to remove all the old textures in a forward fashion,  so that the textures can be `push_back` instead of being `insert_front`: in this way we keep a (multidimensional) array of `vector` in which to store the textures to be pooled.
This would improve the speed of all texture pooling operations.

gregory38: AgePool can be slow. It is only called once by frame.

FetchSurface/Recycle are called more often. And I would say the same number of time. I don't think the performance is that critical for the code. However, if you don't find a `surface` in `FetchSurface`, you will call `CreateSurface` which is a tad slower (and requires a driver thread synchronization). Having too many texture in the pool and you will get an out-of-memory.

A standard PS2 texure is around `256x256 texels` (worst case is `1024x1024`). Smallest framebuffer is `1024 x 1024 x upscaling^2`

AlessandroVetere: `deque` itself instead of `list` is for sure an improvement in linear scan speed (though erase at random position can be slower depending on the implementation).
Maybe it could be worth it to use array lookup + `vector` scan to squeeze additional speed in `FetchSurface` and `Recycle`.
I don't get why can't we keep track of the VRAM available instead of hardcoding maximum sizes though both in `Recycle` and `AgePool`.

gregory38: Sure you can do fancy stuff. But
* You have diminish return quickly
* How do you know the memory of allocated data by the driver for a texture
* there is also RAM vs VRAM limit.

So basically, you want to avoid too many framebuffer which are quite costly. Even if you have memory available, it could still be used for better purpose.

AlessandroVetere: We can do a """worst case""" analysis: width * height * 4 * 32bit (RGBA) for each texture (no double precision right?), then make sure the total textures (pool + what is outside) doesn't exceed a given size.
When aging the pool or recycling, start by removing bigger textures, then fallback, until the target amount of free memory is reached.
We can trade a big framebuffer for a higher number of smaller textures in this way!
It wouldn't be perfect but better than nothing.

gregory38: Honestly it does't need to be clever. You don't need thousand of textures. 300 (unallocated) textures is already quite huge.

gregory38: @AlessandroVetere 
Those improvements are only idea for the future. So far, I would like to have/use a nice iterator for the new container :) This way we could merge this PR.

AlessandroVetere:
I have code ready but messed up rebase..Will fix tonight

plugins/GSdx/CMakeLists.txt
```diff
@@ -142,6 +142,7 @@ set(GSdxHeaders
     GSDump.h
     GSdx.h
     GSdxResources.h
+	GSFastList.h
```
danilaml: tab

AlessandroVetere: My bad, I'll collect more comments before fixing

plugins/GSdx/GSdx.vcxproj
AlessandroVetere: CMakeList.txt is updated now.

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,221 @@
...
+// This iterator is const_iterator by design as T is intended to be a primitive type or pointer
+class FastListIterator
+{
+private:
+	// Should be a * instead of &, because otherwise it cannot be uninitialized (see end())
+	const FastList<T>* m_fastlist;
+	uint16 m_index;
+
+public:
+	// Used for end() iterator, which does not need m_fastlist ptr.
```
gregory38: I don't agree with you. This kind of code should be legal.
```
auto it = fl.end();
it--;
fl.erase(it);
```

So yes it isn't used currently. But it is a bad idea to drop it. The compiler is capable to optimize a mov instruction.

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -75,9 +75,9 @@ void GSTextureCache::RemovePartial()
 
 	for (int type = 0; type < 2; type++)
 	{
-		for (auto &t : m_dst[type]) delete t;
-
-		m_dst[type].clear();
+		for (auto t : m_dst[type]) delete t;
+
```
gregory38: Useless space. And please rename `Clear` to `clear` to keep a standard API.

gregory38:
If possible, it would be nice to split the last commit in smaller one. For example
* erase trick on sw side
* use range loop
* use the new list iterator

I'm well aware that it will be difficult now to have a nice split. So it isn't mandatory but if you could split the part that is possible to split it would be nice. In case of regression, it would be easier to bisect it.

plugins/GSdx/GSFastList.h
AlessandroVetere: By design of `FastList` it isn't possible from the end of the list (which happens to be the head) to come back to the last element as the prev_index of the end element (the head) isn't managed.
Do I have to implement this behaviour?

gregory38: Oh. What do you miss ? Only the value of prev_index for the head ? Is it a big change ? I would prefer to have it.

AlessandroVetere: It is not a big change there is the need to manage the value both in insert and erase of an element from the list. It's definetely viable to add the feature and update the iterator.

gregory38: Ok. So you can do it :)

AlessandroVetere: It was much easier than expected :)

AlessandroVetere:
After managing the index to the last element, I decided that `FastList` could take the place of the `deque` I put in `GSDevice` as `m_pool`. By adding some API friendly methods to the class, this has been possible without changing the actual `GSDevice.cpp` code, only by updating `GSDevice.h`. It runs fine, and IMO the custom container is a better choice overall.

gregory38:
I would be fine if you separate the erase trick.

AlessandroVetere:
Expect the updated PR coming by tonight or tomorrow then!

danilaml:
@AlessandroVetere 
> It runs fine, and IMO the custom container is a better choice overall.

IMHO, custom container is never a better choice, unless it's reliably outperforms the standard one in various benchmarks.  It's harder to maintain and can always contain sneaky bugs in some weird corner cases, while the std implementation should conform to the standard and receive incomparably more testing. It would also function uniformly on all supported platforms, while custom container might have some unobvious  performance/behavior quirks on some OS/arch.
TL/DR: it's almost always better to use std containers unless you can reliably prove the custom one is better in some way.

Sorry, for the rant. Just expressing my opinion. I'm just a lurker here, not a contributor, so it's probably doesn't weigh a lot.

mirh:
> It's harder to maintain and can always contain sneaky bugs in some weird corner cases

I thought hitting upstream bugs was our second job? 😄 

> It would also function uniformly on all supported platforms, while custom container might have some unobvious performance/behavior quirks on some OS/arch.

*Something something something [vs 2015 bug](https://github.com/PCSX2/pcsx2/issues/998).*

danilaml:
@mirh that was a compiler bug, not an STL one, wasn't it? MSVC STL was in a pretty good shape since latest updates to 2015 and even better in 2017. AFAIK, they also test their implementation on the same testing suite as clang too. Anyway, custom solutions tend to bit rot, while upstream bugs are eventually fixed. When it's not the case it could be the argument for using custom container, but IMHO it should be proven first that it solves this problem and not follow logic "I replaced std container which worked fine with custom one which may or may not be faster, it compiled so it should be OK".

mirh:
Benchmarks are lied down in the OP. 
Plus there's all the technical blathering I can't comprehend in #1756, that I'm pretty certain has to be cool stuff if made enthusiast our C veterans here.

danilaml:
@mirh OP has benchmarks only for the initial GSdx TC change. I was referring to the one in `GSDevice.h` (not yet committed) ,since the only rationale given was that ` It runs fine`.

gregory38:
```
IMHO, custom container is never a better choice, unless it's reliably outperforms the standard one in various benchmarks. It's harder to maintain and can always contain sneaky bugs in some weird corner cases, while the std implementation should conform to the standard and receive incomparably more testing.
```
Most of the time I will agree with you. And for the GSdevice.h case it might be useless. The container code is generic anyway so it won't hurt. And code isn't too complex (unlike a map/hash implementation).

But I think you highlight the std container limitation `various benchmarks` and `should conform to the standard`. You want speed for you're use case not `various benchmarks`. It is often a speed trade-off between some operations. Typically here, the size of the list will be small whereas the std container allow infinite list. It allow us to avoid memory allocation, improve cache locality and reduce node size by ~33% (on 64 bits it would be ~50%).

AlessandroVetere:
@danilaml I agree with you in general, but that said, in the worst case you can just revert the line to go back to the standard container as I made sure to be not invasive. As @gregory38 pointed out, there are many benefits in our case, and the code being his initial idea is also non coming out of the blue.
What could be done in this direction though is assing a test suite for pcsx2. And maybe a benchmark suite for the code. But that's OT.

AlessandroVetere:
This is it, possibly the last version of the code, further fixes and tweaks aside.
For what concerns `FastList`, I added API standard methods to make it able to replace the `std::list` in `GSDevice` without updating any call, and with some `auto` magic this enables the swapping of container easily.
In the other parts of the code I updated syntax or patterns (like the erasure of an iterator while looping), and whenever `FastList` was involved, I minimized the difference from a `std` container only to the methods which works on `uint16` index instead of fully fledged iterators.
The thing mentioned above ensure that if one changes back to a `std` container, it is only required to rewrite the hot parts which now are `FastList` only code. I avoided using `FastList` capabilities like the one of erasing an element in a range loop (this would have caused major pain in the future).
There may be still room for improvement in the code, it would be great if someone else could do some benchmarks.

plugins/GSdx/GSDevice.cpp
```diff
@@ -45,7 +45,7 @@ GSDevice::GSDevice()
 
 GSDevice::~GSDevice()
 {
-	for(auto &t : m_pool) delete t;
+	for(auto t : m_pool) delete t;
```
gregory38: why did you remove the reference ?

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	__forceinline ~FastList() {
+		_aligned_free(m_buffer);
+	}
+
+	void clear() {
+		// Initialize m_capacity to 4 so we avoid to Grow() on initial insertions
+		// The code doesn't break if this value is changed with anything from 1 to USHRT_MAX
+		m_capacity = 4;
```
gregory38: Could you add some empty lines to make the code more readable. One here and likely a couple of one in the block below.

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,252 @@
...
+		// Initialize m_buffer and m_free_indexes_stack as a contiguous block of memory starting at m_buffer
+		// This should increase cache locality and reduce memory fragmentation
+		_aligned_free(m_buffer);
+		m_buffer = (Element<T>*)_aligned_malloc(m_capacity * sizeof(Element<T>) + (m_capacity - 1) * sizeof(uint16), 64);
+		m_free_indexes_stack = (uint16*)&m_buffer[m_capacity];
+		// Initialize m_buffer[0]
+		m_buffer[0].data = nullptr; // Unused field
```
gregory38: maybe could be written as
`m_buffer[0].data = {nullptr, 0, 0};`

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,252 @@
...
+		for (uint16 i = 0; i < m_capacity - 1; i++) {
+			m_free_indexes_stack[i] = i + 1;
+		}
+	}
+
+	// Insert the element in front of the list and return it's position in m_buffer
+	__forceinline uint16 InsertFront(const T data) {
+		if (Full()) {
+			Grow();
+		}
```
gregory38: again add an empty line

plugins/GSdx/GSTextureCacheSW.cpp
```diff
@@ -133,30 +121,26 @@ void GSTextureCacheSW::RemoveAll()
...
 	{
-		hash_set<Texture*>::iterator j = i++;
-
-		Texture* t = *j;
+		Texture* t = *i;
 
 		if(++t->m_age > 10)
 		{
-			m_textures.erase(j);
+			i = m_textures.erase(i);
```
gregory38: I'm not sure that it will work. Erase should return 1.

AlessandroVetere: The return value of erase is the iterator following the last removed element, according to the standard.

plugins/GSdx/GSFastList.h
AlessandroVetere: It is `m_buffer[0] = {nullptr, 0, 0};`, but I agree it is nicer. Will change...maybe, read below please.
That's the occasion to discuss a problem here: `m_buffer[0].data` is meant to be unused, but initializing it with `nullptr` makes the code non compiling for every `T` which is not a pointer. Should I drop it's initialization, so to have compilation for every type? Beware that some design choice of `FastList`, that is returning `T` instead of `T&` when accessing the data, makes it unusable for every non primitive-non pointer type.

plugins/GSdx/GSDevice.cpp
AlessandroVetere: Two reasons:
1) begin `t` of a pointer type (that is necessary given it's usage in `delete t`, so it's without loss of generality), it's useless to "suggest" to access the data in the list remotely by reference. (which the compiler could/would optimize of course). All the other access to the list element infact are by copying the values (returning actual values from iterators instead of keeping the reference to the pointer, which is useless). So it's to keep the code coherent.
2) `FastList` doesn't expose the methods to access by reference the data (which could or should maybe be fixed, if it's non intended).

plugins/GSdx/GSTextureCacheSW.cpp
gregory38: Sorry, I read the wrong line on the man. Anyway, I still think previous pattern is better. It avoid the else branch and it is more generic. Tbh, a better solution could be something with algortihm/lambda (speed is useless here)

plugins/GSdx/GSFastList.h
gregory38: Use T default constructor.

plugins/GSdx/GSDevice.cpp
gregory38: Ok. Yes people (i.e. I) often forgot to add the `&` (which ought to be the default). But I agree it is the same here.

For completeness, could you extend fast list to support access by reference (if it is doable).

plugins/GSdx/GSTextureCacheSW.cpp
AlessandroVetere: Just for reference, previous pattern is:

```C++
for (auto i = list.begin(); i != list.end(); ) {
    auto j = i++;
    if (condition) {
        list.erase(j);
    }
}
```
Well this solution has one more variable but no else branch and discards `erase` return value. Still convinced that it's better? I just thought that the return value of `erase` should be spoiled, as it let room for optimization, but I let you decide.

plugins/GSdx/GSFastList.h
AlessandroVetere: Neat

plugins/GSdx/GSDevice.cpp
AlessandroVetere: Yes I read also that `auto` discards by default the `&`, even if (in our case) the iterator `*` operator returns a reference to the element...what a mess.
To keep coherency though, removing it here it's ok.
I think it's no effort to add access by reference (which will replace the current one by value), but, I wonder what changes could lead this to the compiled code...I hope none.

plugins/GSdx/GSTextureCacheSW.cpp
AlessandroVetere: By the way I applied the new pattern in all the erase while looping code of the PR, not only here.

plugins/GSdx/GSDevice.cpp
danilaml: You can just use `GSTexture*` instead of `auto`. This type is not that hard to read/type, to be worth replacing with `auto`, especially since the other issues.

AlessandroVetere: Done, no changes in performance nor code other than adding some `&`, so far so good. 
Everything is managed by the compiler.
Now this opens the question: should I also now implement a non const iterator? It's completely useless IMO and without meta-programming tricks it would consist in a copy paste without `const` keyword of the current one.

plugins/GSdx/GSTextureCacheSW.cpp
gregory38: This code is only called once by frame so execution time is 0 even without optimization. Honestly micro-optimization is good for hot-path otherwise it is better to ensure readability of code.

AlessandroVetere: Sure but given that I was trying to keep the same approach in the code, also in the other loops the same pattern has been updated. What do you prefer to do?

gregory38: Fine, keep the new pattern.

AlessandroVetere:
Updated `FastList` to support access by `T&`, which is the default for `std` containers instead of `T`, and also updated code formatting.
Kept `auto` for range loop on pointers.
Kept new erase while looping pattern.

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -842,11 +824,11 @@ void GSTextureCache::InvalidateVideoMem(GSOffset* off, const GSVector4i& rect, b
 
 	for(int type = 0; type < 2; type++)
 	{
-		for(list<Target*>::iterator i = m_dst[type].begin(); i != m_dst[type].end(); )
+		// AV: This code should be rewritten to avoid using this much "continue"
```
gregory38: IMHO continue doesn't hurt.

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -867,7 +849,7 @@ void GSTextureCache::InvalidateVideoMem(GSOffset* off, const GSVector4i& rect, b
 				}
 				else
 				{
-					m_dst[type].erase(j);
+					i = list.erase(i);
```
gregory38: honestly I found your pattern very error-prone. For short loop it is fine but here code is quite complete and you need to read it carefully (if future branch are created/delete) to properly update the iterator. Speed wise the saving is 0, so for me it isn't a good idea here

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,253 @@
...
+class FastList {
+	friend class FastListIterator<T>;
+private:
+	// The index of the first element of the list is m_buffer[0].next_index
+	//     The first Element<T> of the list has prev_index equal to 0
+	// The index of the last element of the list is m_buffer[0].prev_index
+	//     The last Element<T> of the list has next_index equal to 0
+	// All the other Element<T> of the list are chained by next_index and prev_index
+	// m_buffer has dynamic size m_capacity
+	// m_buffer[0] is always present as auxiliary Element<T> of the list
```
gregory38: Could you add some explanation that due to array re-allocation, pointer to element are invalidated. But fortunately for us the iterator is based on index which protect us from invalidation. (I let you translate in correct English)

plugins/GSdx/GSFastList.h
```diff
@@ -0,0 +1,253 @@
...
+		m_buffer[to_remove.next_index].prev_index = to_remove.prev_index;
+	}
+
+	void Grow() {
+		if (m_capacity == USHRT_MAX) {
+			throw std::runtime_error("FastList size maxed out at USHRT_MAX (65535) elements, cannot grow futhermore.");
+		}
+
+		// 32768 is the first number that when doubled overflows uint16
+		const uint16 new_capacity = m_capacity < 32768 ? m_capacity * 2 : USHRT_MAX;
```
gregory38: better use `m_capacity < USHRT_MAX / 2` and remove the comment

plugins/GSdx/GSTextureCache.cpp
AlessandroVetere: As for this loop I agree with you, the code here is branch-wise so complex that the standard pattern is not suited at all. The "continue" comment was because I already noticed that and I thought that refactoring the code a bit would help. Easiest solution though is to use old erase pattern here.

plugins/GSdx/GSFastList.h
AlessandroVetere: I would then say: `m_capacity <= (uint16)(USHRT_MAX / 2)`, as `32768` is `(USHRT_MAX / 2) + 1`. Do you think the cast is necessary? (not that is makes any difference)

plugins/GSdx/GSFastList.h
AlessandroVetere: I wrote:
```C++
// Due to m_buffer reallocation, the pointers to Element<T> stored into the array
//     are invalidated every time Grow() is executed. But FastListIterator<T> is
//     index based, not pointer based, and the elements are copied in order on Grow(),
//     so there is no iterator invalidation (which is an index invalidation) until 
//     the relevant iterator (or the index alone) are erased from the list.
```
It's ok?

AlessandroVetere:
Rebased with the suggested changes:
- Old erase pattern in `GSTextureCache::InvalidateVideoMem(...)`, also removed comment
- Invalidation explanation comment added in `FastList<T>`
- Removed the magic number in `FastList<T>::Grow()`

plugins/GSdx/GSDevice.cpp
mirh: > How do you know the memory of allocated data by the driver for a texture

👀 👀 👀 👀 👀
http://phoronix.com/scan.php?page=news_item&px=NVIDIA-NV_query_resource

lightningterror:
@gregory38 what's the status on this ?

plugins/GSdx/GSFastList.h
gregory38: ok

gregory38:
Review is fine. I want to run a couple of dump on it. And then we can merge it.

gregory38:
Ok let's go. I don't have time to test it more. And I don't want to block PRs.

AlessandroVetere:
Thanks for the approval and the review.

AlessandroVetere:
@gregory38 There's some problem going on with the master branch Travis CI which seems to be unrelated to the merged commits.

```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:
The following packages have unmet dependencies:
 libgl1-mesa-dev:i386 : Depends: libgl1-mesa-glx:i386 (= 10.1.3-0ubuntu0.6) or
                                 libgl1-mesa-glx-lts-utopic:i386 but it is not going to be installed or
                                 libgl1-mesa-glx-lts-vivid:i386 but it is not going to be installed or
                                 libgl1-mesa-glx-lts-wily:i386 but it is not going to be installed or
                                 libgl1-mesa-glx-lts-xenial:i386 but it is not going to be installed
 libglu1-mesa-dev:i386 : Depends: libglu1-mesa:i386 (= 9.0.0-2) but it is not going to be installed
 libsdl2-dev:i386 : Depends: libegl1-mesa-dev:i386
                    Depends: libgles2-mesa-dev:i386
 libwxgtk3.0-dev:i386 : Depends: libwxgtk3.0-0:i386 (= 3.0.0-2) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
The command "./travis.sh before_install" failed and exited with 100 during .
Your build has been stopped.
```

gregory38:
Thanks for the copy/past. I didn't manage to access the log from work. I will try to check it tomorrow.

AlessandroVetere:
I know the PR is merged but ZoE2 on the most recent build is hitting 43.8 fps here in the bench case, which is +1.3 fps from my local branch (which was behind 3 months of GSdx updates) and ultimately +3.8 fps from the starting point.
Keep up the good work guys!

Issue #961, opened 2015-11-08
Meta: GSdx PCRTC issues


rz5:
The documentation of the GS states that frame buffer width can be obtained by multiplying the value of FBW (a field of the FRAME register) by 64.
Frame buffer height isn't specified in registers; the documentation says the height on the PS2 is to be controlled with scissoring tests.

Right now pcsx2 relies on hacks and guesses, including:
0) Display height is good for every game except for Pool Paradise, which sets DH to 1151
when the proper value should be 576;
1) Frame buffer height is roughly the same height as display (DH + 1 / MAGH +1),
unless it's DMC3 or the US/JP version of SkyGunner - then the height gets set to 448 and it
may be halved if the game is interlaced + frame mode
2) "Device" height is the same as display height, unless it's interlaced + frame mode - then
the current height gets halved.

There are still some games where the hacks/guesses in the codebase don't work, where the height is still not correct and we get black bars on the bottom of the screen or the picture.

Ideally there would be a way to detect and set the correct height solely based the state of registers that the GS is able to see.
- **List of _known_ PCRTC issues**
- [ ] [KOF 2000](http://forums.pcsx2.net/Thread-GSDX-PCRTC-Bug-Report-The-King-of-Fighters-2000-SLPS-251-56) `(improper height)`
- [ ] Metal Slug 3 `(improper height)` (NTSC 60hz only)
- [ ] Twisted Metal FMV `(improper width)`
- [ ] ICO FMV `(improper width)`

Possible
- [ ] Virtua Figher 4 - https://forums.pcsx2.net/Thread-Virtua-Fighter-4-SCES-50759-E--30477?pid=612849#pid612849

mirhl:
Improved by #962

gregory38:
To complete the information:
* Framebuffer size is independent of the real display size.
* Scissor only impact the rendering in the Framebuffer
* Pool paradise uses 16 bits formats, height are 2 times bigger with 16 bits format.

rz5:
@gregory38 had an idea about NTSC mode resolutions having a maximum height of 448. #962 changes GSState.cpp::GetFrameRect() to test this idea and it seems to work.

gregory38:
Thanks for the credits ^^ However I don't know if it a real physical limitation (speed of the PLL), potentially some NTSC games can use a bigger resolution.

rz5:
I'm quoting @ramapcsx2 on this: 
[...]

PR #1948, opened 2017-05-20
GSdx: Implement SYNCV register bitfields

ssakash:
**Summary of changes**:

* Add bitfield structure of SYNCV register to get differential phase and vertical front/back porch interval timings, most likely the timings are done in microseconds (highly likely) but I'm still not sure, so left it unknown for now in the comment. The last one (VS) is a total mystery at the moment.

I had to go through the [technical specifications of NTSC television system](http://www.tek.com/document/primer/television-measurements-ntsc-systems) to even get a bare idea on what most of bitfields do and I see no direct connection of these values to the height determination of output circuit at the moment, but there _might_ be one.

### DIFF
--- a/plugins/GSdx/GS.h
+++ b/plugins/GSdx/GS.h
@@ -511,8 +511,12 @@ REG64_(GSReg, SYNCH2)
 REG_END
 
 REG64_(GSReg, SYNCV)
-	uint64 _DUMMY;
-	// TODO
+	uint32 VFP:10; // Vertical Front Porchinterval (?s)
+	uint32 VFPE:10; // Vertical Front Porchinterval End (?s)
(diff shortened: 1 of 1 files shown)


lightningterror:
I'm interested to know what will this do.

ssakash:
> I'm interested to know what will this do.

I don't think these values will be of much use, it's really low-level synchronization timing values used by the Sync Generator in PCRTC circuit.

I have only added them hoping that they might somehow help with https://github.com/PCSX2/pcsx2/issues/961 but haven't found any sort of connection between them yet.

gregory38:
It could be also cycle timing. Because you can wait N cycles in the hardware but not some ms. And don't ask me the size of a cycle ;)

Issue #1941, opened 2017-05-11
[Q] Is input lag has low as we can?


mirh:
Follows the initially v-sync focused only [brain-storm session post](https://github.com/PCSX2/pcsx2/issues/1437#issuecomment-293114109). 
Now, putting aside that (which thankfully seems quite much on the road of getting fixed), I'm wondering if those techniques couldn't be used.. to lower input lag on the other hand?

The availability situation of the various swap chains.. Is terrible. 
Though, I guess conditionals for DXGI_SWAP_EFFECT_FLIP_SEQUENTIAL or D3DSWAPEFFECT_FLIPEX not to crash on different OSs could do miracles :p
[sample*1*](https://github.com/mpv-player/mpv/blob/master/video/out/opengl/context_angle.c#L375)

What about poor GL, you'll ask me then? Microsoft is too selfish to reserve to it a fast track too (even though for some reason [this](http://www.neogaf.com/forum/showpost.php?p=198239900&postcount=34) not-so-seemingly-dumb guy claims otherwise)
Well, surprise: say hello to `NV_`***DX***`_interop` [extension](https://sites.google.com/site/snippetsanddriblits/OpenglDxInterop), which leads us back to the previous case. 
[sample*2*](https://github.com/mpv-player/mpv/blob/master/video/out/opengl/context_dxinterop.c)
Aside of helping with avoiding BitBlt, could this also use as a stop-gap measure to use DX shaders with GL renderer #1795?

Last but not least, I changed [PFD_DOUBLEBUFFER](https://github.com/PCSX2/pcsx2/blob/b2a2a3ace196e9d7afe27e62622b2b8292b9ecb5/plugins/GSdx/GSWndWGL.cpp#L192) to  PFD_DOUBLEBUFFER_DONTCARE and W-T-F, borderless fullscreen is not exclusive anymore on AMD (https://github.com/PCSX2/pcsx2/issues/1437#issuecomment-300770824)

@turtleli we are all lovingly waiting from you 😘

turtleli:
I have no idea about what you're trying to say. I don't think I could help even if I did.

mirh:
![](https://i-msdn.sec.s-msft.com/dynimg/IC554633.png)

willkuer:
I think he rather asks for an avih-style presentation of the problem/solution:

1. The following issue is present in pcsx2
2. I can trigger the issue by doing...
3. I expect instead that and that to happen.

[...]

PR #1950, opened 2017-05-21
Greg/vsync

gregory38:
To be discussed because OpenGL tearing/compositor/vsync is a nightmare.

Related to issue #1437

### DIFF
--- a/pcsx2/Pcsx2Config.cpp
+++ b/pcsx2/Pcsx2Config.cpp
@@ -20,6 +20,7 @@
 #include "Utilities/IniInterface.h"
 #include "Config.h"
 #include "GS.h"
+#include "gui/GSFrame.h"
 
 void TraceLogFilters::LoadSave( IniInterface& ini )
 {
@@ -200,7 +201,7 @@ Pcsx2Config::GSOptions::GSOptions()
 {
 	FrameLimitEnable		= true;
 	FrameSkipEnable			= false;
-	VsyncEnable				= false;
+	VsyncEnable				= VSYNC_NO;
 
 	SynchronousMTGS			= false;
 	DisableOutput			= false;
@@ -232,6 +233,25 @@ void Pcsx2Config::GSOptions::LoadSave( IniInterface& ini )
 
 	IniEntry( FramesToDraw );
 	IniEntry( FramesToSkip );
+}
+
+int Pcsx2Config::GSOptions::GetVsync() const
+{
+	// If we have a compositor (AKA not exclusive fullscreen), we need to
+	// wait the compositor signal. 0x10000 is a flag for the GS plugin
+	if (VsyncEnable == VSYNC_COMPOSITOR)
+		return 0x10000;
+
+	if (!FrameLimitEnable || g_LimiterMode == Limit_Turbo)
+		return 0;
+
+	switch (VsyncEnable) {
+		case VSYNC_DYNAMIC: return -1;
+		case VSYNC_ON: return 1;
+		default: return 0;
+	}
+
+	return 0;
 }
 
 const wxChar *const tbl_GamefixNames[] =
--- a/pcsx2/gui/Panels/GSWindowPanel.cpp
+++ b/pcsx2/gui/Panels/GSWindowPanel.cpp
@@ -32,10 +32,21 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 		_("Widescreen (16:9)")
 	};
 
+	const wxString vsync_label[] =
+	{
+		_("Disabled"),
+		_("Standard"),
+		_("Adaptative"),
+		_("Window Compositor")
+	};
+
 	m_text_Zoom = CreateNumericalTextCtrl( this, 5 );
 
 	m_combo_AspectRatio	= new wxComboBox( this, wxID_ANY, wxEmptyString, wxDefaultPosition, wxDefaultSize,
 		ArraySize(aspect_ratio_labels), aspect_ratio_labels, wxCB_READONLY );
+
+	m_combo_vsync = new wxComboBox( this, wxID_ANY, wxEmptyString, wxDefaultPosition, wxDefaultSize,
+		ArraySize(vsync_label), vsync_label, wxCB_READONLY );
 
 	m_text_WindowWidth	= CreateNumericalTextCtrl( this, 5 );
 	m_text_WindowHeight	= CreateNumericalTextCtrl( this, 5 );
@@ -47,7 +58,6 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 	m_check_HideMouse	= new pxCheckBox( this, _("Always hide mouse cursor") );
 	m_check_CloseGS		= new pxCheckBox( this, _("Hide window when paused") );
 	m_check_Fullscreen	= new pxCheckBox( this, _("Default to fullscreen mode on open") );
-	m_check_VsyncEnable	= new pxCheckBox( this, _("Wait for Vsync on refresh") );
 	m_check_DclickFullscreen = new pxCheckBox( this, _("Double-click toggles fullscreen mode") );
 	m_check_AspectRatioSwitch = new pxCheckBox(this, _("Switch to 4:3 aspect ratio when an FMV plays"));
 	//m_check_ExclusiveFS = new pxCheckBox( this, _("Use exclusive fullscreen mode (if available)") );
@@ -55,7 +65,7 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 	m_text_Zoom->SetToolTip( pxEt( L"Zoom = 100: Fit the entire image to the window without any cropping.\nAbove/Below 100: Zoom In/Out\n0: Automatic-Zoom-In untill the black-bars are gone (Aspect ratio is kept, some of the image goes out of screen).\n  NOTE: Some games draw their own black-bars, which will not be removed with '0'.\n\nKeyboard: CTRL + NUMPAD-PLUS: Zoom-In, CTRL + NUMPAD-MINUS: Zoom-Out, CTRL + NUMPAD-*: Toggle 100/0"
 	) );
 
-	m_check_VsyncEnable->SetToolTip( pxEt( L"Vsync eliminates screen tearing but typically has a big performance hit. It usually only applies to fullscreen mode, and may not work with all GS plugins."
+	m_combo_vsync->SetToolTip( pxEt( L"Vsync eliminates screen tearing but typically has a big performance hit. It usually only applies to fullscreen mode, and may not work with all GS plugins."
 	) );
 
 	m_check_HideMouse->SetToolTip( pxEt( L"Check this to force the mouse cursor invisible inside the GS window; useful if using the mouse as a primary control device for gaming.  By default the mouse auto-hides after 2 seconds of inactivity."
@@ -91,6 +101,11 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 	s_AspectRatio	+= Label(_("Zoom:"))			| StdExpand();
 	s_AspectRatio	+= m_text_Zoom;
 
+	wxFlexGridSizer& s_vsync( *new wxFlexGridSizer( 2, StdPadding, StdPadding ) );
+	s_vsync.AddGrowableCol( 1 );
+
+	s_vsync += Label(_("Vsync Mode:"))		| pxMiddle;
+	s_vsync += m_combo_vsync      			| pxExpand;
 
 	*this += s_AspectRatio				| StdExpand();
 	*this += m_check_SizeLock;
@@ -105,7 +120,7 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 	//*this += m_check_ExclusiveFS;
 	*this += new wxStaticLine( this )	| StdExpand();
 
-	*this += m_check_VsyncEnable;
+	*this += s_vsync;
 
 	wxBoxSizer* centerSizer = new wxBoxSizer( wxVERTICAL );
 	*centerSizer += GetSizer()	| pxCenter;
@@ -138,10 +153,9 @@ void Panels::GSWindowSettingsPanel::ApplyConfigToGui( AppConfig& configToApply, 
 
 		m_text_WindowWidth	->ChangeValue( wxsFormat( L"%d", conf.WindowSize.GetWidth() ) );
 		m_text_WindowHeight	->ChangeValue( wxsFormat( L"%d", conf.WindowSize.GetHeight() ) );
-	}
 
-	m_check_VsyncEnable->SetValue( configToApply.EmuOptions.GS.VsyncEnable );
-	m_check_VsyncEnable->Enable  ( !configToApply.EnablePresets );//grayed-out when presets are enabled
+		m_combo_vsync->SetSelection(configToApply.EmuOptions.GS.VsyncEnable);
+	}
 }
 
 void Panels::GSWindowSettingsPanel::Apply()
@@ -157,7 +171,7 @@ void Panels::GSWindowSettingsPanel::Apply()
 	appconf.AspectRatio		= (AspectRatioType)m_combo_AspectRatio->GetSelection();
 	appconf.Zoom			= Fixed100::FromString( m_text_Zoom->GetValue() );
 
-	gsconf.VsyncEnable		= m_check_VsyncEnable->GetValue();
+	gsconf.VsyncEnable = m_combo_vsync->GetSelection();
 
 	appconf.IsToggleFullscreenOnDoubleClick = m_check_DclickFullscreen->GetValue();
 	appconf.IsToggleAspectRatioSwitch = m_check_AspectRatioSwitch->GetValue();
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -63,6 +63,14 @@ enum GamefixId
 	GamefixId_COUNT
 };
 
+enum VsyncMode
+{
+	VSYNC_NO,
+	VSYNC_ON,
+	VSYNC_DYNAMIC,
+	VSYNC_COMPOSITOR,
+};
+
 // Template function for casting enumerations to their underlying type
 template <typename Enumeration>
 typename std::underlying_type<Enumeration>::type enum_cast(Enumeration E)
@@ -293,7 +301,7 @@ struct Pcsx2Config
 
 		bool	FrameLimitEnable;
 		bool	FrameSkipEnable;
-		bool	VsyncEnable;
+		int		VsyncEnable;
 
 		int		FramesToDraw;	// number of consecutive frames (fields) to render
 		int		FramesToSkip;	// number of consecutive frames (fields) to skip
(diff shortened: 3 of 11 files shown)


mirh:
Seriously, it's just about nvidia driver giving us problems. 
I'm not saying we should flip the bird (I already posted on their dev forums, hoping they can clarify.. just anything)... 
but in the meantime for as much I know, you might as well rollout the perfect implementation in the ideal® world (what I think you/I/we had already in mind), *plus*:
* just wait to reintroduce WS_POPUP
* to the DwmIsCompositionEnabled && window!=monitor check, add && vendorId==nvidia

tabnk:
Interesting :). Going to test soon ...

image

tabnk:
illegal else without matching if	

D:\torrent\pcsx2\plugins\GSdx\GSWndWGL.cpp	332	

void GSWndWGL::SetVSync(int vsync)
{
	m_vsync = vsync;
	// m_swapinterval uses an integer as parameter
	// 0 -> disable vsync
	// n -> wait n frame
	**else if (m_swapinterval && vsync != 0x10000) m_swapinterval(vsync);**
}

tabnk:
Tested on Windows 10 with Nvidia GPU. Full screen and GDSX in OpenGL. Nvidia vsync setting handle by pcsx2.

Disabled: No tearing but frame pacing.
Adaptive: No tearing and no frame pacing. 
Windows Compositor: No tearing but minor/some frame pacing.

mirh:
>  Windows Compositor: No tearing but minor/some frame pacing.

Isn't that the same of what was [reportedly](https://github.com/PCSX2/pcsx2/issues/1437#issuecomment-300638364) perfect?

> Btw, both adaptive and Windows compositor increase input lag

Mhhh, so my paranoia in #1941 wasn't totally tinfoil 😝 
http://www.blitzbasic.com/Community/posts.php?topic=94676#1088686
Might this make some sense?

tabnk:
Game like GT4 with input lag will be easily noticeable.

pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -47,15 +58,14 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
...
 	m_check_DclickFullscreen = new pxCheckBox( this, _("Double-click toggles fullscreen mode") );
 	m_check_AspectRatioSwitch = new pxCheckBox(this, _("Switch to 4:3 aspect ratio when an FMV plays"));
 	//m_check_ExclusiveFS = new pxCheckBox( this, _("Use exclusive fullscreen mode (if available)") );
 
 	m_text_Zoom->SetToolTip( pxEt( L"Zoom = 100: Fit the entire image to the window without any cropping.\nAbove/Below 100: Zoom In/Out\n0: Automatic-Zoom-In untill the black-bars are gone (Aspect ratio is kept, some of the image goes out of screen).\n  NOTE: Some games draw their own black-bars, which will not be removed with '0'.\n\nKeyboard: CTRL + NUMPAD-PLUS: Zoom-In, CTRL + NUMPAD-MINUS: Zoom-Out, CTRL + NUMPAD-*: Toggle 100/0"
 	) );
 
-	m_check_VsyncEnable->SetToolTip( pxEt( L"Vsync eliminates screen tearing but typically has a big performance hit. It usually only applies to fullscreen mode, and may not work with all GS plugins."
+	m_combo_vsync->SetToolTip( pxEt( L"Vsync eliminates screen tearing but typically has a big performance hit. It usually only applies to fullscreen mode, and may not work with all GS plugins."
```
ssakash: Maybe add some explanation for the different Vsync modes?

plugins/GSdx/GS.cpp
```diff
@@ -967,7 +965,7 @@ EXPORT_C GSsetFrameSkip(int frameskip)
 
 EXPORT_C GSsetVsync(int enabled)
 {
-	s_vsync = !!enabled;
+	s_vsync = enabled;
```
ssakash: Parameter name should most likely be updated (Similar to the way you've done on the GSDevice* files)

pcsx2/Pcsx2Config.cpp
```diff
@@ -234,6 +235,28 @@ void Pcsx2Config::GSOptions::LoadSave( IniInterface& ini )
...
+	// If we have a compositor (AKA not exclusive fullscreen), we need to
+	// wait the compositor signal. 0x10000 is a flag for the GS plugin
+	if (VsyncEnable == VSYNC_COMPOSITOR)
+		return 0x10000;
+
+	if (!FrameLimitEnable)
+		return 0;
+
+	if (g_LimiterMode == Limit_Turbo)
+		return 0;
```
ssakash: I wonder if these ``if()`` statements could be combined? (Avoids redundant return statement)

pcsx2/gui/Panels/GSWindowPanel.cpp
mirh: Which indeed I still haven't understood. 
I can see the point in letting user select between adaptive and normal, albeit (50/60hz mumblings and [quite](https://www.hardocp.com/article/2012/04/16/nvidia_adaptive_vsync_technology_review/3) [random](http://techreport.com/review/22735/a-closer-look-at-some-geforce-gtx-680-features/2) [confusion](https://hardforum.com/threads/nvidia-adaptive-vsync-vs-vsync-triple-buffering.1688239/#post-1038642454) [aside](https://forums.geforce.com/default/topic/517715/do-you-need-tripple-buffering-with-the-new-adaptive-vsync-option-enabled-/#3683196)) I still feel like one of them has to be 100% better than the other ultimately. 

But wtf is compositor vsync? It's not like being subject to dwm is in program's control.

pcsx2/Pcsx2Config.cpp
gregory38: Yes we could.

pcsx2/Pcsx2Config.cpp
```diff
@@ -234,6 +235,28 @@ void Pcsx2Config::GSOptions::LoadSave( IniInterface& ini )
...
+
+	switch (VsyncEnable) {
+		case VSYNC_DYNAMIC: return -1;
+		case VSYNC_ON: return 1;
+		default: return 0;
+	}
+
+	return 0;
```
gregory38: Yes I don't like the code neither ;) I will add an intermediate variable.

plugins/GSdx/GS.cpp
gregory38: ooups I miss this one.

pcsx2/gui/Panels/GSWindowPanel.cpp
gregory38: You're free to propose me a description and better naming :)

* Normal
  * tear free
   * but 30fps drop when you can't hit the target. Bad for the SW renderer, and bad for slow GS post-processing effect.
* Adaptive 
   * may tear 
    * Avoid most of the performance hit when you can't reach the target.
* Compositor
  * vsync is handled by compositor not us.

By the way, could someone generate me a backtrace to the code that set `WS_POPUP`. Maybe there is an option/argument somewhere to control the flag.

mirh: > Avoid most of the performance hit when you can't reach the target.

That's a pretty old story. ~~Has frame rendering and presentation really to be locked?~~

Ok yes, I see you were referring to the former frames. I thought they invented triple buffering exactly for that?

> Avoid most of the performance hit when you can't reach the target.

Sorry, now that I think.. Isn't that working with actually frames displayed on screen?
Because I'd guess PAL is always out for that then. 

> vsync is handled by compositor not us.

That's not a choice you have. Compositor (if it "owns" your window) is always syncing you. 
You can just cooperate if any, or do the infant. 

> By the way, could someone generate me a backtrace to the code that set WS_POPUP

For the love of god, let's just leave nvidia alone for now. It's already a good mess without.

gregory38: Triple buffering gives you plenty of issues (read your link ;) ). Tbh, the only good solution is gsync/freesync.

Dunno how PAL behaves. Likely badly. I'm curious if we can emulate faster to render at 60 fps. I'm afraid there isn't any SW solution.

Yes it would be better to automatically detect compositor but I'm not sure how to do it. Besides option is nice.

Market GPU is Nvidia (even more on the GL renderer). Having a way to enable/disable this flag will greatly help the testing. It does't mean that we need a public option

mirh: > read your link ;)

I did. But I got lost once everybody in the fucking world started to have a **very** convinced opposite opinion 😭😭
At least in the small little theory I have in mind though, it sounds as best as [v-sync] can: render as fast as possible, only present most recent frame. 
Of course (I think?) this would only make a difference whenever rendering (and not only presentation) is in the slim sub-60-yet-still-full-speed limbo. 

> Tbh, the only good solution is gsync/freesync.

Oh, absolutely, that's 100% perfection. 
But in that case we really wouldn't have bother with anything in the first place 😛 

> Dunno how PAL behaves. Likely badly. I'm curious if we can emulate faster to render at 60 fps. 

That's pretty defeatist on your side
Aside of nobody actually even ever complaining about this.. We could have an option that (when in fullscreen - when we'll have it) switches screen to 50Hz (or even 59.94 ?) - which at least in my monitor are natively supported refresh rates. 

> Yes it would be better to automatically detect compositor but I'm not sure how to do it. 

.-.
Didn't we had like 50 posts just to discuss that?
Putting aside the nvidia case (which I already suggested a dumb workaround, to be used just until hopefully they can answer me) it seems to just be about DwmIsCompositionEnabled && window!=monitor

... Or were you referring specifically to the actual windows functions to call? :S

 > Besides option is nice.

Only if your users are smart enough! (and option makes sense separated in the first place)

> Having a way to enable/disable this flag will greatly help the testing

Ok, makes sense, for as much at the moment I'm thinking to either some style we still aren't using or driver bug.

gregory38: > I did. But I got lost once everybody in the fucking world started to have a very convinced opposite opinion

Every 1/60 s ~~ 16 ms, you have a ready interrupt from your display
* if a frame is ready, pick it up
* otherwise wait
The difference between vsync and adaptive vsync is that the former will wait the next display interrupt. Whereas the latter will wait the frame ready signal from your GPU. Adaptive vsync avoid to wait a long time when you barely miss the target. However if you're screen is busy displaying something, you will get some tearing. In short adaptive allow value between 30fps and 60fps. Whereas standard vsync is either 30fps or 60fps.

```
it seems to just be about DwmIsCompositionEnabled && window!=monitor
```
We can do something when we properly understand the situation because it is far from clear for me. And how do we test if we are in exclusive fullscreen or not.


Anyway, so far goal is to have a testbench. And then tune it. I won't merge this PR in current state.

mirh: > We can do something when we properly understand the situation because it is far from clear for me.

The only mystery here is nvidia driver internal functioning. Yes, they are the majority of GL users, but for the moment there's simply already a lot of variables to "understand properly", even leaving aside bugs. 

> Whereas the latter will wait the frame ready signal from your GPU. 

I see, if that's the case though, I don't see why we should even ever prefer normal v-sync anymore? 
Yes, we may say that on the down side, adaptive may still get tearing then. 
But what would happen otherwise? Stuttering, and that's simply it. 
Is there really some case where somebody would prefer it to tearing?

> Whereas standard vsync is either 30fps or 60fps.

Yes, but what about triple buffering?

> And how do we test if we are in exclusive fullscreen or not.

It's exactly the `window!=monitor` part. 
Nvidia exclusive fullscreen behavior for the moment is unknown, thus the additional && vendorid!=nvidia thing I had recommended above, so we can just think to that in another issue, otherwise it will never end.

gregory38: Well I guess it depends on the user. But yeah, potentially we can keep only adaptive. I need to add code to check the proper driver support.

> Yes, but what about triple buffering?
3B is another story. Instead to block the GPU when you miss the display interrupt. You can start the rendering into another 3rd buffer. Then you have 2 possibilities if you're still suck. Either you drop a frame (GL) or your stall the GPU (DX).

It is more complicated to implement and it still have various limitation. Potentially it could work well enough for us when game uses interlacing because we have a light frame after the heavy frame. So we have 2 vsync to render a frame. Adaptive is much easier to implement, just need to call SwapInterval with -1.

> `window!=monitor`
You mean size of the window different of size of monitor. I don't know how to query the size of the monitor. Besides what will happen in the next driver upgrade. What must happen for a 4:3 surface on a 16:9 monitor.

mirh: > I need to add code to check the proper driver support.

I see you are always a step ahead :*

> Either you drop a frame (GL) or your stall the GPU (DX)

I believe that \*especially\* for our use-case, GL behavior is actually golden. 

> You mean size of the window different of size of monitor. 

Yep. 

> I don't know how to query the size of the monitor. Besides what will happen in the next driver upgrade. 

Afaiu, for as much all of these are drivers quirks, they are not there at random. 
AFAIK, in Windows there's no way to set opengl (or I guess anything but DX) to fullscreen (or perhaps it was that *the application* has no way to know when it's what), so this is why they do this. 
Which I guess now with variable refresh rate monitors is more important then ever. 
I'm pretty confident this is expected and desired behavior on their side. 

To check, I had written [here](https://github.com/PCSX2/pcsx2/issues/1437#issuecomment-300828554). Though really, I understand if you prefer not to think about this. 

> What must happen for a 4:3 surface on a 16:9 monitor.

Who cares? Stretch/fit/fill/keep user preference is what controls this affair.

gregory38: > Who cares? Stretch/fit/fill/keep user preference is what controls this affair.

There are 3 sizes. Surface size, Window size, Screen size. I'm not sure how to get them in the difference place (GSdx or Core). I think the easier solution will be in the core. So what is the magic formulae to detect exclusive fullscreen ?

> Afaiu, for as much all of these are drivers quirks, they are not there at random.

It is random as it isn't a standard behavior. Adaptive vsync is standard, you set a flag and you're done. Exclusive fullscreen is just a hack from the driver under your back. Anything could change the behavior of the driver. And the current old situation of WS_POPUP is a direct proof.

mirh: > There are 3 sizes. Surface size, Window size, Screen size.

I'm not sure why anything between surface (which is gsdx final rendering I guess) and "above levels" should change or matter now?

> So what is the magic formulae to detect exclusive fullscreen ?

They aren't supposed to, indeed as I was saying. 
We just assume under X, Y, Z conditions that to be the case. 

> And the current old situation of WS_POPUP is a direct proof.

What? I'm not sure how you can deduce the "degree of ABI stability" from breakage cases alone. 
For as much as I [know](https://devtalk.nvidia.com/default/topic/1009367/opengl/v-sync-is-always-forced-in-fullscreen-even-with-swapinterval-0-and-driver-forced-off/), they may have vsync (is that standard enough?) broken in the first place.

gregory38:
* https://github.com/PCSX2/pcsx2/blob/master/pcsx2/gui/FrameForGS.cpp#L494
This function uses an extra parameter for flags. Technically it ought to be wx flags but I think MS flags will be forwarded.
* [Wx Flag](http://docs.wxwidgets.org/trunk/classwx_top_level_window.html#ab4089f1274bcb74e5f7763d1fb84ee28)

gregory38:
@tabnk 
I added the hidden option `EnableVsyncWindowFlag` that must be equivalent to `WS_POPUP`. Could you test if it is working as expecting.

tabnk:
For this build, whether EnableVsyncWindowFlag disabled or enabled cannot go full screen.

gregory38:
@tabnk I fixed my code. Could you retry it thanks you.

tabnk:
With EnableVsyncWindowFlag=enabled at PCSX2_ui.ini. Nvidia setting handle by pcsx2.

vsync disabled: No tearing but frame pacing
vsync standard: No tearing but frame pacing
vsync adaptive: No tearing but less frame pacing
vsync window compositor: No tearing but little frame pacing

mirh:
Coooooould you test <381 drivers?

tabnk:
Can't tested in driver 381 below. Already customise per game setting on Nvidia control panel so if install old driver than all my setting will deleted.

gregory38:
@tabnk 
Do you feel that new option have any impact ? In the end what is the best configuration ?

tabnk:
Actually I'm felt almost no different whether EnableVsyncWindowFlag is enabled or disabled. 

Depends on individual preference, both Adaptive and Windows Compositor are good.

mirh:
https://forums.geforce.com/default/topic/914268/export-import-nvidia-control-panel-settings-and-color-profiles/#4796732
Also, last time I checked settings were kept unless checking "clean install" option.

gregory38:
Well I'm not sure the flag propagate well to the windows API. Need a windows dev to confirm it.

tabnk:
Comparing to this https://github.com/PCSX2/pcsx2/blob/d8e0b9f5412431ab89bc14b9b8c30b0bef46a560/3rdparty/wxwidgets3.0/src/msw/toplevel.cpp#L1114 and your new code.

In full screen mode and disable all vsync, earlier one got tearing but your no tearing. Maybe your doesn't propagate well.

gregory38:
New version that should propagate the flag.

tabnk:
On new version, screen tearing work on full screen.

mirh:
Does any of the v-sync options fix it then?

tabnk:
With tearing enable on full screen.

vsync disabled: Tearing and frame pacing
vsync standard: Tearing and some frame pacing
vsync adaptive: No tearing and no noticeable frame pacing. Very good for my GT4.
vsync window compositor: Tearing but maybe some frame pacing

gregory38:
So, I don't know exactly what to do. Here my current proposal
* Keep the combo box but with 3 states (OFF / ON / adaptive). Please share you opinion on the name. Adaptive is also called "Late Vsync". Therefore keep integer setup on GS side. I need to add a fallback for drivers that don't support it.
* I removed Vsync from the preset as "Late vsync" should be better but it isn't supported by all drivers. Opinion is welcome.
* Keep the hidden WS_POPUP option
* Drop vsync window compositor (or at least don't merge it until we understand fully the behavior and potentially auto detect the fullscreen mode).

mirh:
Drivers that don't support adaptive should just not have it presented/available, I guess. 
For the remainder, I'm fine. 
(assuming this DwmIsCompositionEnabled && window!=monitor check gets added)
Then, nvidia users can even screw for all that I care, considering I have been asking about trying previous drivers and nobody seems to listen.

gregory38:
```
Drivers that don't support adaptive should just not have it presented/available, I guess.
```
The thing is that you know about the support when you call the GL swap interval function. And the option is on the core side. Instead we could just fallback to vsync.

```
 (assuming this DwmIsCompositionEnabled && window!=monitor check gets added)
```
I propose that we merge the others parts. And then we try to find how to write stuff for DWM.

mirh:
> The thing is that you know about the support when you call the GL swap interval function. And the option is on the core side. Instead we could just fallback to vsync.

Oh right. Then I guess whatever you said make sense. 

> I propose that we merge the others parts. And then we try to find how to write stuff for DWM.

Again, I don't think the problem is dwm, and that's also simply what's already used by glfw guys too. 
Besides, wasn't all this fuss basically only about the discovery of dwmflush?

gregory38:
Note: I pushed a rebased version. So only the important bits remain.

lightningterror:
Can this be closed since there is a new improved PR ?

gregory38:
Yes let's close it.

PR #1952, opened 2017-05-26
Add xz submodule, port GSdx xz dump save/readback to Windows

turtleli:
Changes:
* Adds xz as a submodule - It uses a mirror I created in my own account (it'd probably be better to have the mirror in the PCSX2 account, I don't think there's an official Github xz mirror). I left the repo untouched and just added a project file and config file outside the directory. The submodule commit corresponds to the v5.2.3 tag (which I really should add to the commit message).
* Use std::vector for the xz dump output buffer  - I completely forgot that Windows by default uses a 1MB stack, so I didn't realise the 1mb std::array would break things. Oh well.
 * Port the xz readback/save to Windows - note that I haven't updated the GSDumpGUI, so I don't think the xz readback will work when using that (I tested with rundll32.exe).


plugins/GSdx/GS.cpp
```diff
@@ -1086,233 +1087,124 @@ EXPORT_C GSReplay(HWND hwnd, HINSTANCE hinst, LPSTR lpszCmdLine, int nCmdShow)
...
 
-		uint8 regs[0x2000];
-		GSsetBaseMem(regs);
+	Console console("GSdx", true);
 
-		s_vsync = theApp.GetConfigB("vsync");
+	GSinit();
 
-		HWND hWnd = NULL;
+	uint8 regs[0x2000];
```
turtleli: array

plugins/GSdx/GS.cpp
```diff
@@ -1086,233 +1087,124 @@ EXPORT_C GSReplay(HWND hwnd, HINSTANCE hinst, LPSTR lpszCmdLine, int nCmdShow)
...
 
-		fread(regs, 0x2000, 1, fp);
+	uint32 crc;
+	file->Read(&crc, 4);
+	GSsetGameCRC(crc, 0);
 
-		long start = ftell(fp);
+	GSFreezeData fd;
+	file->Read(&fd.size, 4);
+	fd.data = new uint8[fd.size];
```
turtleli: unique_ptr/vector

plugins/GSdx/GS.cpp
```diff
@@ -1086,233 +1087,124 @@ EXPORT_C GSReplay(HWND hwnd, HINSTANCE hinst, LPSTR lpszCmdLine, int nCmdShow)
...
 
-		list<Packet*> packets;
-		vector<uint8> buff;
-		int type;
+	struct Packet {uint8 type, param; uint32 size, addr; std::vector<uint8> buff;};
 
-		while((type = fgetc(fp)) != EOF)
-		{
-			Packet* p = new Packet();
+	std::list<std::unique_ptr<Packet>> packets;
```
turtleli: unique_ptr seems unnecessary.

mirh:
https://git.tukaani.org/xz.git
https://github.com/xz-mirror/xz.git

turtleli:
> https://git.tukaani.org/xz.git

I'm avoiding the official repo because the buildbots will devour bandwidth.

> https://github.com/xz-mirror/xz.git

Seem unrelated to the xz dev.

mirh:
> I'm avoiding the official repo because the buildbots will devour bandwidth.

Is the buildbot so dumb to have to clone every repo every time?

> Seem unrelated to the xz dev.

Seems trustworthy enough.

turtleli:
> Is the buildbot so dumb to have to clone every repo every time?

I don't know what you are expecting of generic buildbots that don't preserve state.

> Seems trustworthy enough.

I prefer either an official repo/mirror or maintaining a mirror ourselves. Less what-ifs to think of.

gregory38:
I'm looking at github permission. You might receive plenty of useless mail :p

gregory38:
So I think I manage to put you in the core team with full access. I didn't find how to remove from the committer team. Enjoy :)

danilaml:
>I don't know what you are expecting of generic buildbots that don't preserve state.

Appveyor has between build caching as well as travis (to some extent).
Also, traffic can be greatly minimized if you specify git clone depth at around 5 (for both submodules and initial pull)
https://github.com/PCSX2/pcsx2/blob/master/appveyor.yml#L20
(and in travis somewhere too)

turtleli:
> So I think I manage to put you in the core team with full access. I didn't find how to remove from the committer team. Enjoy :)

Uh oh. More dangerous buttons. ;)

@danilaml 
> Appveyor has between build caching as well as travis (to some extent).

I can't think of a way to use the AppVeyor cache stuff. Caching has been already been set up on Travis CI.

> Also, traffic can be greatly minimized if you specify git clone depth at around 5 (for both submodules and initial pull)

Are there any potential issues with doing so (at least when it's used on the buildbots)?

danilaml:
>Are there any potential issues with doing so (at least when it's used on the buildbots)?
Well, if, while the build is queued, someone pushes more than depth num commits to the branch, then that build would fail (it will fail to check-out), but it can also be viewed as a positive change (since, it won't clog the buildbot so much, if someone goes mad with pushing/rebasing/squashing commits every 5 seconds in his pr). So yeah, if the clone depth is bigger than the potential buildbot queue there shouldn't be any problems at all, at if it isn't there shouldn't be much problems either.

Btw, judging by appveyor log, there is no need to limit depth of submodules (since currently it takes less than 5 seconds), but checking out whole pcsx2 repo takes ~1min.

You can control clone depth in appveyor.yml by adding `clone_depth: Num`.

>I can't think of a way to use the AppVeyor cache stuff. Caching has been already been set up on Travis CI.
Well, I'm not talking about compiler cache. In travis you can just cache whole folder/files. But yeah, I don't see you downloading huge dependencies and I'm not sure that msbuild would be smart enough to no build some project twice, even if it was built before.
Anyway, PCSX2 current appveyor build is only 8 minutes (and would be even less with clone depth) so I don't think it's worth the effort,

turtleli:
Hmm. Seems there's actually a small issue with shallow cloning: the window title on dev builds changes from
`PCSX2 v1.5.0-dev-xxx-gsha`
to
`PCSX2 1.5.0-date (git)`

turtleli:
Updated:
 * Added and used xz mirror in PCSX2 account
 * Fixed an EOF handling bug

danilaml:
@turtleli Ah, yeah, totally forgot about that. If your application versioning depends on `git describe` it might be a problem since it looks for the latest reachable tag while when you specify depth, you only check out certain number of commits that may not include any tags.

gregory38:
Travis just downloads the .so files from the package. Maybe we can precompile the dependency on Windows too if it is too slow

plugins/GSdx/GSLzma.cpp
```diff
@@ -111,14 +111,14 @@ void GSDumpLzma::Decompress() {
 }
 
 bool GSDumpLzma::IsEof() {
-	return feof(m_fp) && (m_avail == 0);
+	return feof(m_fp) && m_avail == 0 && m_strm.avail_in == 0;
 }
 
-void GSDumpLzma::Read(void* ptr, size_t size) {
+bool GSDumpLzma::Read(void* ptr, size_t size) {
```
gregory38: The pointer could be a `const void *` instead

plugins/GSdx/GS.cpp
```diff
@@ -1086,233 +1087,123 @@ EXPORT_C GSReplay(HWND hwnd, HINSTANCE hinst, LPSTR lpszCmdLine, int nCmdShow)
...
-		int type;
+	GSvsync(1);
 
-		while((type = fgetc(fp)) != EOF)
-		{
-			Packet* p = new Packet();
+	struct Packet {uint8 type, param; uint32 size, addr; std::vector<uint8> buff;};
 
-			p->type = (uint8)type;
+	auto read_packet = [&file](uint8 type) {
```
gregory38: Maybe we can use common OS code to read data.

plugins/GSdx/GSLzma.cpp
turtleli: Err... the memory needs to be modified, so it can't be const.

plugins/GSdx/GS.cpp
turtleli: Yeah, the replay stuff could probably be combined to some extent.

Personally I think it'd be nice to have a cross platform GS dump viewer at some point in the future, something that'd allow some level of automated testing.

plugins/GSdx/GSLzma.cpp
gregory38: Oups. Forget what I said :)

plugins/GSdx/GS.cpp
gregory38:
```
cross platform GS dump viewer at some point in the future, something that'd allow some level of automated testing.
```
Yeah. It is already possible to dump the frame `savef`. Maybe we should keep frame data in memory. Then we mostly need to run both renderer and to compare the frame output.

gregory38:
```
EDIT: I'll see what I can do.
```
It can also be done later. This PR will already be a huge improvement. As you wish.

turtleli: Yeah, I think I'll leave it alone for now.

plugins/GSdx/GSLzma.cpp
```diff
@@ -162,19 +162,17 @@ bool GSDumpRaw::IsEof() {
...
-		memcpy(ptr, &v, 1);
-	} else {
-		size_t ret = fread(ptr, 1, size, m_fp);
-		if (ret != size) {
-			fprintf(stderr, "GSDumpRaw:: Read error (%zu/%zu)\n", ret, size);
-			throw "BAD"; // Just exit the program
-		}
+bool GSDumpRaw::Read(void* ptr, size_t size) {
+	size_t ret = fread(ptr, 1, size, m_fp);
+	if (ret != size && ferror(m_fp)) {
```
ssakash: It makes more sense to use ``||`` over here. What about cases when only one of the condition is false? you won't be throwing the error.

I'm not sure at which scenarios ``ferror()`` returns 1, [reference at some sites](http://www.cplusplus.com/reference/cstdio/ferror/) just mentions that 1 will be set when the previous file operation has failed but it's still not fully clear on how they define a failed operation. Plus [C library code for ferror()](https://sourceware.org/git/?p=glibc.git;a=blob;f=libio/ferror.c;h=90a3b6949625e3b6e913c8705434872a9a8fb94b;hb=9d46370ca338054cb6ea7ebeddcf06c7ac7ad1a9) doesn't make sense to me.

turtleli: > It makes more sense to use || over here. What about cases when only one of the condition is false? you won't be throwing the error.

That's intended behaviour. If an n-byte `fread` reads less than n bytes, then the EOF might have been reached, which is not an error.

> I'm not sure at which scenarios `ferror()` returns 1, reference at some sites just mentions that 1 will be set when the previous file operation has failed but it's still not fully clear on how they define a failed operation. Plus C library code for ferror() doesn't make sense to me.

You need to look at `fread`, `fwrite`, `fseek`, etc., not at `ferror`. I'm not sure about what doesn't make sense to you in the glibc ferror code.

Side note: cppreference is the better reference site IMO.

PR #1954, opened 2017-05-27
GSdx-ogl: Printing console messages v2

lightningterror:
Follow up to
https://github.com/PCSX2/pcsx2/commit/ec63b04719fd9c05a6aeeacb55dc1c54f5ef145b

Add intel broken driver wiki link message , print intel and amd buggy
driver message once , pring texture barrier and viewpoint array info
once.
Some other minor edits.

### DIFF
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -445,14 +445,14 @@ namespace GLLoader {
 		if (s_first_load) {
 			if (fglrx_buggy_driver) {
 				fprintf(stderr, "The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver.\n"
-				"Check out the link below for further information.\n"
-				"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know\n");
+					"Check out the link below for further information.\n"
+					"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know\n");
 			}
 
 			if (intel_buggy_driver) {
 				fprintf(stderr, "The OpenGL renderer is inefficient on Intel GPUs due to an inefficient driver.\n"
-				"Check out the link below for further information.\n"
-				"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-Intel-GPUs-All-you-need-to-know\n");
+					"Check out the link below for further information.\n"
+					"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-Intel-GPUs-All-you-need-to-know\n");
 			}
 
 			if (!found_GL_ARB_viewport_array) {
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -442,25 +442,33 @@ namespace GLLoader {
 		status &= status_and_override(found_GL_ARB_texture_barrier, "GL_ARB_texture_barrier");
 		status &= status_and_override(found_GL_ARB_get_texture_sub_image, "GL_ARB_get_texture_sub_image");
 
-		if (fglrx_buggy_driver) {
-			fprintf(stderr, "The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver.\n"
-			"Check out the link below for further information.\n"
-			"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know\n");
-		}
+		if (s_first_load) {
+			if (fglrx_buggy_driver) {
+				fprintf(stderr, "The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver.\n"
+				"Check out the link below for further information.\n"
+				"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know\n");
+			}
 
-		if (!found_GL_ARB_viewport_array) {
-			fprintf(stderr, "GL_ARB_viewport_array: not supported ! function pointer will be replaced\n");
-			glScissorIndexed   = ReplaceGL::ScissorIndexed;
-			glViewportIndexedf = ReplaceGL::ViewportIndexedf;
-		}
+			if (intel_buggy_driver) {
+				fprintf(stderr, "The OpenGL renderer is inefficient on Intel GPUs due to an inefficient driver.\n"
+				"Check out the link below for further information.\n"
+				"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-Intel-GPUs-All-you-need-to-know\n");
+			}
 
-		if (!found_GL_ARB_texture_barrier) {
-			fprintf(stderr, "GL_ARB_texture_barrier: not supported ! Rendering will be corrupted\n");
-			glTextureBarrier = ReplaceGL::TextureBarrier;
+			if (!found_GL_ARB_viewport_array) {
+				fprintf(stderr, "GL_ARB_viewport_array is not supported! Function pointer will be replaced\n");
+				glScissorIndexed   = ReplaceGL::ScissorIndexed;
+				glViewportIndexedf = ReplaceGL::ViewportIndexedf;
+			}
+
+			if (!found_GL_ARB_texture_barrier) {
+				fprintf(stderr, "GL_ARB_texture_barrier is not supported! Blending emulation will not be supported\n");
+				glTextureBarrier = ReplaceGL::TextureBarrier;
+			}
 		}
 
 #ifdef _WIN32
-		// Thanks you Intel to not provide support of basic feature on your iGPU
+		// Thank you Intel for not providing support of basic features on your IGPUs.
 		if (!found_GL_ARB_direct_state_access) {
 			Emulate_DSA::Init();
 		}


lightningterror:
Hope I didn't make too many mistakes this time around.

plugins/GSdx/GLLoader.cpp
```diff
@@ -247,7 +247,7 @@ namespace Emulate_DSA {
 
 	// Replace function pointer to emulate DSA behavior
 	void Init() {
-		fprintf(stderr, "DSA is not supported. Expect slower performance\n");
+		fprintf(stderr, "Direct State Access: is not supported! Expect slower performance\n");
```
ssakash: Is the colon here necessary?

gregory38:
Maybe it will be better to add a macro. I should have done it the first time

Something like that. If it doesn't work with the stream parameter, create 2 macro for stdout/stderr.
```
#define PRINT_ONCE(stream, ...)	do { \
    if (s_print_once) printf(stderr, format(__VA_ARGS__).c_str()); \
} while(0);
```

Fwiw there are similar kinds of macro in stdafx.h

plugins/GSdx/GLLoader.cpp
```diff
@@ -442,25 +442,36 @@ namespace GLLoader {
...
+				fprintf(stderr, "GL_ARB_viewport_array: is not supported! Function pointer will be replaced\n");
+			}
 			glScissorIndexed   = ReplaceGL::ScissorIndexed;
 			glViewportIndexedf = ReplaceGL::ViewportIndexedf;
 		}
 
 		if (!found_GL_ARB_texture_barrier) {
-			fprintf(stderr, "GL_ARB_texture_barrier: not supported ! Rendering will be corrupted\n");
+			if (s_first_load) {
+				fprintf(stderr, "GL_ARB_texture_barrier: is not supported! Blending emulation will not be supported\n");
```
ssakash: Ditto

plugins/GSdx/GLLoader.cpp
```diff
@@ -442,25 +442,36 @@ namespace GLLoader {
...
+				fprintf(stderr, "The OpenGL renderer is inefficient on INTEL GPUs due to an inefficient driver.\n"
+					"Check out the link below for further information.\n"
+					"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-INTEL-GPUs-All-you-need-to-know\n");
+			}
 		}
 
 		if (!found_GL_ARB_viewport_array) {
-			fprintf(stderr, "GL_ARB_viewport_array: not supported ! function pointer will be replaced\n");
+			if (s_first_load) {
+				fprintf(stderr, "GL_ARB_viewport_array: is not supported! Function pointer will be replaced\n");
```
ssakash: Ditto

plugins/GSdx/GLLoader.cpp
lightningterror: Not really , I just thought It would be better to match how the other messages look (texture barrier and viewpoint array.

lightningterror:
@gregory38 I placed the macro right under row 388
`// Note: GL messages are present in common code, so in all renderers.`
Is that correct?
Also what about the other changes I and you made ?

gregory38:
Put the macro in the beginning of the namespace

MrCK1:
IMO I believe the commit message should be a little more clear as to what the changes are for future reference.

lightningterror:
@MrCK1 I'll make the changes.

@gregory38 In stdafx.h or GLLoader.cpp?

gregory38:
GLLoader.cpp. If you put the macro in a .h it will be available everywhere whereas we want to limit it to the current namespace.  Note it must be put after the declaration of `s_first_load`

plugins/GSdx/GLLoader.cpp
lightningterror: I can put the intel/amd buggy driver + texture barrier + viewpoint array in one but is that ok?
```
glScissorIndexed   = ReplaceGL::ScissorIndexed;
glViewportIndexedf = ReplaceGL::ViewportIndexedf;
```
Is it ok if these are under the same statement ?

MrCK1: What's the point? I would at least leave it as is until after the 1.6 release.

plugins/GSdx/GLLoader.cpp
```diff
@@ -482,4 +490,9 @@ namespace GLLoader {
 
 		s_first_load = false;
 	}
+	
+	#define PRINT_ONCE(stream, ...)	do { \
```
gregory38: The goal is to use the macro.
So you can write `PRINT_ONCE(stderr, "The OpenGL renderer .....");` instead. But macro must come first.

gregory38: @lightningterror do you understand what I mean ?

lightningterror: Something like this? 

```
		#define PRINT_ONCE(stream, ...)	do { \
		if (s_print_once) printf(stderr, format(__VA_ARGS__).c_str()); \
		} while(0);

		if (fglrx_buggy_driver) {
			PRINT_ONCE(stderr, "The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver.\n"
				"Check out the link below for further information.\n"
				"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know\n");
			}
```

ssakash: Yes, that's what he meant.

lightningterror: I'm getting a compile error tho , I'll need some help.
`Error C2664 'int printf(const char *const ,...)': cannot convert argument 1 from 'FILE *' to 'const char *const `'

gregory38: Google `fprintf` and `printf` to understand the differences.

plugins/GSdx/GLLoader.cpp
```diff
@@ -442,25 +442,33 @@ namespace GLLoader {
...
+				}
+					
+			if (intel_buggy_driver) {
+				fprintf(stderr, "The OpenGL renderer is inefficient on INTEL GPUs due to an inefficient driver.\n"
+					"Check out the link below for further information.\n"
+					"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-INTEL-GPUs-All-you-need-to-know\n");
+				}
+
+			if (!found_GL_ARB_viewport_array) {
+				fprintf(stderr, "GL_ARB_viewport_array: is not supported! Function pointer will be replaced\n");
```
turtleli: ssakash asked before, but I'll ask again - is the colon necessary? I don't think it makes much sense.

lightningterror: No it isn't , I wasn't the one that put it there. I can make some edits and remove them.

lightningterror:
Sigh I'm having issues with adding the macro , I think I'll just leave that for another pr in the future.

@turtleli any other suggestions ?

plugins/GSdx/GLLoader.cpp
```diff
@@ -442,25 +442,33 @@ namespace GLLoader {
...
+				fprintf(stderr, "The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver.\n"
+					"Check out the link below for further information.\n"
+					"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know\n");
+				}
+
+			if (intel_buggy_driver) {
+				fprintf(stderr, "The OpenGL renderer is inefficient on INTEL GPUs due to an inefficient driver.\n"
+					"Check out the link below for further information.\n"
+					"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-INTEL-GPUs-All-you-need-to-know\n");
+				}
```
turtleli: `}` has been indented too much. It should have 1 level less indentation. Same applies to the other `if` statements.

Also, INTEL -> Intel

plugins/GSdx/GLLoader.cpp
```diff
@@ -442,25 +442,33 @@ namespace GLLoader {
...
+				fprintf(stderr, "The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver.\n"
+				"Check out the link below for further information.\n"
+				"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know\n");
+				}
+
+			if (intel_buggy_driver) {
+				fprintf(stderr, "The OpenGL renderer is inefficient on Intel GPUs due to an inefficient driver.\n"
+				"Check out the link below for further information.\n"
+				"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-Intel-GPUs-All-you-need-to-know\n");
+				}
```
turtleli: You misunderstood me. I meant this:
```
if (intel_buggy_driver) {
	fprintf(stderr, "The OpenGL renderer is inefficient on Intel GPUs due to an inefficient driver.\n"
		"Check out the link below for further information.\n"
		"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-Intel-GPUs-All-you-need-to-know\n");
}
```
The `if` and `}` should have the same level of indentation.

lightningterror: Ah yea I did, what about the spaces , should I revert them back or?

lightningterror:
@turtleli , @ssakash all good ?

gregory38:
So could it be merged ?

plugins/GSdx/GLLoader.cpp
```diff
@@ -442,25 +442,33 @@ namespace GLLoader {
...
 
-		if (!found_GL_ARB_viewport_array) {
-			fprintf(stderr, "GL_ARB_viewport_array: not supported ! function pointer will be replaced\n");
-			glScissorIndexed   = ReplaceGL::ScissorIndexed;
-			glViewportIndexedf = ReplaceGL::ViewportIndexedf;
-		}
+			if (intel_buggy_driver) {
+				fprintf(stderr, "The OpenGL renderer is inefficient on Intel GPUs due to an inefficient driver.\n"
+				"Check out the link below for further information.\n"
+				"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-Intel-GPUs-All-you-need-to-know\n");
```
turtleli: Indentation isn't quite right. Should be something like:
```
fprintf(stderr, "blah blah\n"
	"continue on\n"
	"blah blah\n");
```

lightningterror:
Fixed.
I'll let you guys do the squashing since I'm away from my pc :)

PR #1966, opened 2017-06-08
Add Danish translation

rffontenelle:


Jakob5566:
Thanks for the help!

Jakob5566:
I've now testet the new build and something is wrong. I cannot choose Danish in Language Selection. What could possibly be wrong?

gregory38:
I think it is normal. So far only the source code was updated. I will push the precompiled mo file on my next compilation.

Jakob5566:
Okay - didn't know that :)

Issue #1930, opened 2017-05-04
IOP: TLB miss in my video mode test homebrew


ssakash:
I was planning on looking into the undocumented privileged GS registers related to vertical and horizontal timings by monitoring the register changes in my [homebrew](https://www.dropbox.com/s/i5aeb9ele6roqg4/VModeTest.elf?dl=0) but seems like the master branch is failing to properly load it, the log spews out some TLB miss errors which weren't the case before when I tried with it on an older version of PCSX2.

Hence I decided to start looking for the commit which introduced the issue and ta-duh, it's this one! https://github.com/PCSX2/pcsx2/commit/b3b1f3ac685da28030993e70ae6433175060d57e

@sudonim1 : I'd prefer if you could fix it ASAP, thank you.

sudonim1:
I'll look into this soon, but I'll quickly mention that the last guy who had a similar issue was loading his homebrew from a network share and it went away when he tried a local path.

gregory38:
@sudonim1 any news ?

sudonim1:
You'd think so, wouldn't you?

ramapcsx2:
A workaround idea:
Put the elf onto a CD, run ulaunch.elf, navigate to the elf on the CD, run the elf.
It's not a lot of work, might do the job.

seapancake:
This is working for me on 2148, has this been fixed or am I missing something?

gregory38:
@ssakash is it still an issue ?

PR #1967, opened 2017-06-09
GSdx-PCRTC: Add delay before usage of new DH value

ssakash:
**Summary of changes**

* Add some delay before usage of newly initialized DH value from DISP register. Fixes vertical shaking issue in Mizuiro. (http://forums.pcsx2.net/Thread-GSDX-PCRTC-Bug-Report-Mizuiro-NTSC-J)

Code needs to be tested on games which alter the height in-game, tested this on ICO which switches to different height when changed to either PAL/NTSC videomode and the delay seems to be very short and the height transition on the video mode change seems to be fine. Though I still need to test this on my ``VideomodeTest.ELF`` to make sure that the code works fine but that can't be done until #1930 is resolved.

@MrCK1 @FlatOutPS2 @lightningterror - Please test the code for any potential regressions/fixes, counting on all of you. 👍

### DIFF
--- a/plugins/GSdx/GSRenderer.cpp
+++ b/plugins/GSdx/GSRenderer.cpp
@@ -83,6 +83,28 @@ void GSRenderer::ResetDevice()
     if(m_dev) m_dev->Reset(1, 1);
 }
 
+// Mizuiro constantly alters it's DH value on the DISP register from 464 to 480
+// This causes the screen to shake vertically due to the frequent change in DISP register value
+// This doesn't happen in the PS2, the assumption is that the GS has a delay before applying
+// a new DH value to the output circuit. The following function handles such delay.
+GSVector4i DelayDisplayRectangleChange(const GSVector4i& current, int context)
+{
+	static std::array<GSVector4i, 2> cached_rectangle = { GSVector4i(0), GSVector4i(0) };
+	static std::array<bool, 2> initialized_delay = { false, false };
+	const GSVector4i previous = cached_rectangle[context];
+
+	if (!initialized_delay[context] && current.x == previous.x && current.y == previous.y && current.w != previous.w && current.z == previous.z)
+	{
+		GL_INS("Delay on display rectangle height transistion (Requesting change from %d to %d) ", previous.w, current.w);
+		initialized_delay[context] = true;
+		return previous;
+	}
+
+	cached_rectangle[context] = current;
+	initialized_delay[context] = false;
+	return current;
+}
+
 bool GSRenderer::Merge(int field)
 {
 	bool en[2];
@@ -200,7 +222,7 @@ bool GSRenderer::Merge(int field)
 	{
 		if(!en[i] || !tex[i]) continue;
 
-		GSVector4i r = fr[i];
+		GSVector4i r = DelayDisplayRectangleChange(fr[i], i);
 		GSVector4 scale = GSVector4(tex[i]->GetScale()).xyxy();
 
 		src[i] = GSVector4(r) * scale / GSVector4(tex[i]->GetSize()).xyxy();


plugins/GSdx/GSRenderer.cpp
```diff
@@ -83,6 +83,27 @@ void GSRenderer::ResetDevice()
...
+// a new DH value to the output circuit. The following function handles such delay.
+GSVector4i DelayDisplayRectangleChange(const GSVector4i& current, const GSVector4i& previous, int context)
+{
+	if (previous.eq(GSVector4i(0)))
+		return current;
+
+	static std::array<bool, 2> initialized_delay = { false, false };
+	if (!initialized_delay[context] && current.x == previous.x && current.y == previous.y && current.w != previous.w && current.z == previous.z)
+	{
+		GL_PUSH("Delay on display rectangle height transistion (Requesting change from %d to %d) ", previous.w, current.w);
```
gregory38: `GL_PUSH` allow to group log. So instead to have a flat log, you can have a tree. So `GL_PUSH` is nice for big operation that will generate various API calls. In your case, you just need a standard log (for example `GL_INS` if my memory is right). Check `stdafx.h`

plugins/GSdx/GSRenderer.cpp
```diff
@@ -200,11 +221,12 @@ bool GSRenderer::Merge(int field)
 	{
 		if(!en[i] || !tex[i]) continue;
 
-		GSVector4i r = fr[i];
+		static std::array<GSVector4i, 2> r = { GSVector4i(0), GSVector4i(0) };
+		r[i] = DelayDisplayRectangleChange(fr[i], r[i], i);
```
gregory38: Hum, I think it would be possible to keep this code mostly unchanged. In my opinion, you should be able to do `GSVector4i r =  DelayDisplayRectangleChange(fr[i], i);`.

plugins/GSdx/GSRenderer.cpp
```diff
@@ -83,6 +83,27 @@ void GSRenderer::ResetDevice()
     if(m_dev) m_dev->Reset(1, 1);
 }
 
+// Mizuiro constantly alters it's DH value on the DISP register from 464 to 480
+// This causes the screen to shake vertically due to the frequent change in DISP register value
+// This doesn't happen in the PS2, the assumption is that the GS has a delay before applying
+// a new DH value to the output circuit. The following function handles such delay.
```
gregory38: You mean that DH value is overwritten before it is use, isn't it ?

What is "frequent" change for you ? Before Vsync ?

MrCK1:
Will do, but it may be few days before I get a chance to hop on my PC.

plugins/GSdx/GSRenderer.cpp
ssakash: I'll fix it.

plugins/GSdx/GSRenderer.cpp
ssakash: Not sure I understand your idea, this variable needs to be static as I need the previous values of the output for comparison on the function.

gregory38: The idea is that all static stuff should be in DelayDisplayRectangleChange.

plugins/GSdx/GSRenderer.cpp
ssakash: > You mean that DH value is overwritten before it is use, isn't it ?

Yes.

> What is "frequent" change for you ? Before Vsync ?

It's like this,

```
Vsync#1: 480
Vsync#2: 464
Vsync#3: 480
Vsync#4: 464
Vsync#5: 480
Vsync#6: 464
```
In my code, the issue is resolved by delaying the usage of new DH value usage by 1 Vsync. I think something like the following happens with the original hardware,

```
Vsync#1: 480
* First DH value used on the PCRTC output circuits initialization.

Vsync#2: 464
* New DH value detected, wait till next Vsync before applying it

Vsync#3: 480
* DH was updated once again in this Vsync back to the current value used by the PCRTC Circuit.
* PCRTC ignores the previous DH value (from Vsync#2) as it would eventually need to switch to the current DH value anyway.
```

willkuer:
You don't think that you are just masking a core bug?

What happen if you start delayed by one vsync (maybe using a savestate)? Then the whole image will be shifted, or?

ssakash:
> You don't think that you are just masking a core bug?

I don't see any possibilities for a core bug. DISP is a privileged register, it's mapped to the EE's memory address space. We're just basically mirroring the specific EE's memory space to get the value of the register. I think it's highly unlikely for a bug to be present at that. (I _might_ be wrong though as I haven't really looked much into it)

> What happen if you start delayed by one vsync (maybe using a savestate)? Then the whole image will be shifted, or?

In that case, I think it will go with the other height value which is alternating back and forth (464), however vertical shaking would still be avoided. (which was the whole point of the PR)

Here are some comparison pictures on how it looks with 464 and 480 for output circuit height,

**480**


**464**

ssakash:
I would've loved to get my hands on this game to do more tests on my PS2 but unfortunately, I can't seem to find this game anywhere near my local store or even on my regional Amazon site!

The game seems to be only available on Japan's portal of Amazon : https://www.amazon.co.jp/Mizuiro-Best-Japan-NEC-%E4%B8%A6%E8%A1%8C%E8%BC%B8%E5%85%A5%E5%93%81/dp/B01J8TMIXA

MrCK1:
I will probably have some time late tonight to do some testing, so I'll try to report back then.

Off the top of my head, I know Iridium Runners has the same vertical bouncing described here.

- The Suffering: Ties That Bind should be interesting. There's enough PCRTC issues with that game as it is, but I'll have to see how this will factor with the alpha blending, ect.

- Jonny Moseley is an OI_BlitFMV case, not expecting any changes, but I'll test that as well.

gregory38:
> I don't see any possibilities for a core bug. DISP is a privileged register, it's mapped to the EE's memory address space. We're just basically mirroring the specific EE's memory space to get the value of the register. I think it's highly unlikely for a bug to be present at that. (I might be wrong though as I haven't really looked much into it)

I won't be so sure. It could be bad timing on the EE emulation code and a EE/VU thread which execute too soon or too late.

lightningterror:
I tested a coupe of games and everything seems fine. 

> Code needs to be tested on games which alter the height in-game, tested this on ICO which switches to different height when changed to either PAL/NTSC videomode and the delay seems to be very short and the height transition on the video mode change seems to be fine.

Which games do that ?

ssakash:
> Which games do that ?

I don't know, I only had ICO in my collection which did it. Maybe you could look out for the games which have an option to change their videomode in-game, they're more likely to alter their DH at most cases.

By the way, thanks for testing! (Also to you @MrCK1) Really appreciated 👍

gregory38:
Hum, @ssakash could you check it isn't related to PR #1988 too. Both seem to be related to CRTC height.

ssakash:
> Hum, @ssakash could you check it isn't related to PR #1988 too

No, they're not related.

ramapcsx2:
How much is a value of 16 in that DH register on a CRT with interlacing?
Because that's what this sounds like to me :p

gregory38:
The unit is "pixel" (on the display). Tbh, it could be related to feature not fully emulated such as GS register status (interrupt, ... ). It could also be a core bug. Or maybe ssakash is right.

gregory38:
@ramapcsx2 do you think it is the intended purpose of the game ?

gregory38:
@ramapcsx2 ping ?

gregory38:
@ssakash 
I propose that you add clearly that behavior of GS is totally different and that is could be a hacking solution for the game. And let's call it a day.

gregory38:
@ssakash could you add the comment so we can merge this pr.

ramapcsx2:
I'd add this as a configurable option (hack). The game is trying to do something here and we're not supporting some step of the process.
If this is available as an option, it can be tested on other titles or homebrew.
It has the potential to be useful on other titles as well, by the way. Anything that does tricks for flicker reduction might benefit.

gregory38:
I'm not sure an option will be useful. Most games likely keep the same setting for all frames. Note potentially there are some issues with context handling. I know there is something strange on the texture context too.

ssakash:
I'll postpone this one for now, I don't think this is the accurate way to fix it. On the other hand, I'm not sure if the register change 464 -> 480 is intended by the developers or some sort of bug in our code.

And apparently this game is only available in Japan's portal of Amazon so can't even get it to test it. -_-

MrCK1:
@ssakash @ssakash Is it possible this could still affect PSX mode in any way after #2239 is merged?

PR #1968, opened 2017-06-09
GSdx: Cleanup MSVC warnings

ssakash:
**Summary of changes**

* Explicitly cast the variables to the required data type to fix some warnings.
* Omit warnings from liblzma (Too much warnings and not on our side anyway - Any opinion? @turtleli @gregory38 )

Don't really care about the warnings to be honest, they were just annoying as they kept popping up all the time I rebuilt GSdx. -_-

Personally I'd prefer if these warnings from liblzma is ignored for the reasons stated on the commit message. (Same for all the other 3rd party code but I decided to only do for this subproject at the moment as I noticed lots of warnings from this specific one)

### DIFF
--- a/plugins/GSdx/GSDump.cpp
+++ b/plugins/GSdx/GSDump.cpp
@@ -51,7 +51,7 @@ void GSDumpBase::Transfer(int index, const uint8* mem, size_t size)
 		return;
 
 	AppendRawData(0);
-	AppendRawData(index);
+	AppendRawData(static_cast<uint8>(index));
 	AppendRawData(&size, 4);
 	AppendRawData(mem, size);
 }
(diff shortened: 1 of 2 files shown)


plugins/GSdx/GSDump.cpp
```diff
@@ -51,7 +51,7 @@ void GSDumpBase::Transfer(int index, const uint8* mem, size_t size)
 		return;
 
 	AppendRawData(0);
-	AppendRawData(index);
+	AppendRawData(static_cast<uint8>(index));
```
gregory38: Dunno, if it wouldn't be better to use `int` in the interface. And add an assert if the value is > 255. But I'm fine with any way.

mirh:
If you are good-willing and lovely enough, you can join [this](https://sourceforge.net/p/lzmautils/discussion/708858/thread/5ea0391b/#f1fc) discussion and post patches for xz.

turtleli:
> Omit warnings from liblzma (Too much warnings and not on our side anyway - Any opinion?

Dunno. I'm more of the opinion that things should be fixed upstream if possible. Or we disable specific warnings, not everything.

ssakash:
> Dunno. I'm more of the opinion that things should be fixed upstream if possible. Or we disable specific warnings, not everything.

Fair enough, I've removed the second commit, I trust you on fixing it upstream. 👍

PR #1974, opened 2017-06-21
PCSX2-WX: Add more granularity to EE cycle rate slider

ssakash:
**Summary of changes**:

* Added more granularity to the slider, 251 different options to select from! (ranging from 50% to 300% clockspeed)

slider
gifslider
Some other things to discuss about,
* Maybe ``EE Cyclerate`` in the slider should be renamed to something different?

 It's something technical, we increment the cycles before an event by a higher value in underclock and increment it to a lower value in overclock, so it seems sort of confusing to me as people might think that right side increments the cycle rate even higher? Well, normal users might not know much about it, so maybe not worth worrying about and just keep the name as it is?

For those who don't understand, in order to manage the number of instructions executed per second, in most emulators, the following method of cycle clock timing is used:

```c++
cycle_clock += instructions_executed;
while(cycle_clock > next_event_time)
{
  execute_event();
  setup_next_event();
}
```

The bigger the value of ``instructions_executed``, the fewer the instructions which can be executed by the game before the next vsync event occurs. In case of underclock, we're having the value of ``instructions_executed`` as a very high value which causes the EE to actually do less amount of work, it's the opposite for overclock. The manipulation is done by modifying the IPC rate, the system still counts in the units of EE's 294MHz clock. I think we're modifying the scale of cycles per instruction instead of directly modifying the clock speed due to potential problems in timing of the event scheduler which is closely linked to the clock speed of EE. (Maybe?)

* Added colors for different levels of selection in the slider.

I've added various text colors for different levels in the slider ranging from Red, Orange Red, Forest Green, Dark Green, Green. Please test it out and provide your opinion on them - Is it too much/unnecessary (or) actually something useful which might hint the compatibility of the slider level selection to the end user?

* Need tons of testing!

Previously, there was a safe limit on the code which prevented the scaling of blocks when the current amount of blocks recompiling is too low. I removed that limit for overclock values to provide proper scaling of blocks in overclock values. (While making the overclock a bit more effective, it might also introduce some new issues, so please do tons of tests for glitches/crashes/explosions on your games compared to previous overclock behavior)

**Build download link**: [[**Appveyor**]](https://ci.appveyor.com/api/buildjobs/5j0e8pabdr76y52t/artifacts/pcsx2-v1.5.0-dev-2122-g98dacdde3-3100-vs2015-Win32-AppVeyor.7z)

### DIFF
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -390,7 +390,7 @@ struct Pcsx2Config
 				vuThread        :1;		// Enable Threaded VU1
 		BITFIELD_END
 
-		s8	EECycleRate;		// EE cycle rate selector (1.0, 1.5, 2.0)
+		u16	EECycleRate;
 		u8	VUCycleSteal;		// VU Cycle Stealer factor (0, 1, 2, or 3)
 
 		SpeedhackOptions();
@@ -509,7 +509,7 @@ TraceLogFilters&				SetTraceConfig();
 
 
 /////////////////////////////////////////////////////////////////////////////////////////
-// Helper Macros for Reading Emu Configurations.
+// Helper Macros/Functions for Reading Emu Configurations.
 //
 
 // ------------ CPU / Recompiler Options ---------------
@@ -520,6 +520,12 @@ TraceLogFilters&				SetTraceConfig();
 #define CHECK_EEREC					(EmuConfig.Cpu.Recompiler.EnableEE && GetCpuProviders().IsRecAvailable_EE())
 #define CHECK_CACHE					(EmuConfig.Cpu.Recompiler.EnableEECache)
 #define CHECK_IOPREC				(EmuConfig.Cpu.Recompiler.EnableIOP && GetCpuProviders().IsRecAvailable_IOP())
+
+// ------------ Speed Hacks!!! ---------------
+#define EESlider_MaxValue (300)
+#define EESlider_MinValue (50)
+#define EESlider_DefaultValue (100)
+const auto EESliderValueCheck = [](int x) { if (x > EESlider_MaxValue || x < EESlider_MinValue) return EESlider_DefaultValue; else return x; };
 
 //------------ SPECIAL GAME FIXES!!! ---------------
 #define CHECK_VUADDSUBHACK			(EmuConfig.Gamefixes.VuAddSubHack)	 // Special Fix for Tri-ace games, they use an encryption algorithm that requires VU addi opcode to be bit-accurate.
--- a/pcsx2/Pcsx2Config.cpp
+++ b/pcsx2/Pcsx2Config.cpp
@@ -47,7 +47,7 @@ Pcsx2Config::SpeedhackOptions::SpeedhackOptions()
 Pcsx2Config::SpeedhackOptions& Pcsx2Config::SpeedhackOptions::DisableAll()
 {
 	bitset			= 0;
-	EECycleRate		= 0;
+	EECycleRate		= 100;
 	VUCycleSteal	= 0;
 	
 	return *this;
--- a/pcsx2/gui/AppConfig.cpp
+++ b/pcsx2/gui/AppConfig.cpp
@@ -1056,8 +1056,8 @@ bool AppConfig::IsOkApplyPreset(int n)
 		case 5 :	//Set VU cycle steal to 2 clicks (maximum-1)
 					vuUsed?0:(vuUsed=true, EmuOptions.Speedhacks.VUCycleSteal = 2);
 		
-		case 4 :	//set EE cyclerate to 2 clicks (maximum)
-					eeUsed?0:(eeUsed=true, EmuOptions.Speedhacks.EECycleRate = -2);
+		case 4 :
+					eeUsed?0:(eeUsed=true, EmuOptions.Speedhacks.EECycleRate = 60);
 
 		case 3 :	//Set VU cycle steal to 1 click, set VU clamp mode to 'none'
 					vuUsed?0:(vuUsed=true, EmuOptions.Speedhacks.VUCycleSteal = 1);
@@ -1066,8 +1066,8 @@ bool AppConfig::IsOkApplyPreset(int n)
 					EmuOptions.Cpu.Recompiler.vuSignOverflow = false; //VU Clamp mode to 'none'
 
 		//best balanced hacks combo?
-		case 2 :	//set EE cyclerate to 1 click.
-					eeUsed?0:(eeUsed=true, EmuOptions.Speedhacks.EECycleRate = -1);
+		case 2 :
+					eeUsed?0:(eeUsed=true, EmuOptions.Speedhacks.EECycleRate = 80);
 					// EE timing hack appears to break the BIOS text and cause slowdowns in a few titles.
 					//EnableGameFixes = true;
 					//EmuOptions.Gamefixes.EETimingHack = true;
--- a/pcsx2/gui/Panels/ConfigurationPanels.h
+++ b/pcsx2/gui/Panels/ConfigurationPanels.h
@@ -351,7 +351,7 @@ namespace Panels
 		void ApplyConfigToGui( AppConfig& configToApply, int flags=0 );
 
 	protected:
-		const wxChar* GetEEcycleSliderMsg( int val );
+		std::string GetEEcycleSliderMsg( int val );
 		const wxChar* GetVUcycleSliderMsg( int val );
 		void SetEEcycleSliderMsg();
 		void SetVUcycleSliderMsg();
@@ -359,7 +359,7 @@ namespace Panels
 
 		void OnEnable_Toggled( wxCommandEvent& evt );
 		void Defaults_Click( wxCommandEvent& evt );
-		void EECycleRate_Scroll(wxScrollEvent &event);
+		void EECycleRate_Scroll(wxCommandEvent &event);
 		void VUCycleRate_Scroll(wxScrollEvent &event);
 	};
 
--- a/pcsx2/gui/Panels/SpeedhacksPanel.cpp
+++ b/pcsx2/gui/Panels/SpeedhacksPanel.cpp
@@ -19,51 +19,43 @@
 
 using namespace pxSizerFlags;
 
-const wxChar* Panels::SpeedHacksPanel::GetEEcycleSliderMsg( int val )
+std::string Panels::SpeedHacksPanel::GetEEcycleSliderMsg( int val )
 {
-	switch( val )
+	int clockspeed = static_cast<float>(294) * val / 100;
+	wxColour color = wxColour(0, 0, 0);
+
+	switch (val / 10)
 	{
-		case -3:
-		{
-			m_msg_eecycle->SetForegroundColour(wxColour(L"Red"));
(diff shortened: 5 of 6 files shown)


MrCK1:
Trying this out should be fun... 😈 Let the destruction begin!

avih:
I wouldn't have modified it myself, and I think a continuous slider is uncomfortable - UI wise and conceptually - to set, restore, observe its effect, etc. Recall - this is not about fixing issues but rather to change the game internal speed on too-weak or powerful-enough systems. 

Some points:
- If going for a generic scaler, then it would make more sense to me that both sides would be of the same magnitude, but currently it's x2 on one side (100 to 50) while x3 at the other side (100 to 300). I'm unable to estimate how useful 30% is (or 300%), but even if it's not, IMHO it's better to let the user go beyond the useful range and realize by himself that it needs to go back a bit, than going to the edge and wondering how it would work if the slider had a bit more range.
- Re colors, personally yes, I do think it's an overkill. There's value, IMHO, in distinct and clear granularity, and too many colors make it harder to grasp in a brief look where it stands. Again, this is not unlike the effectively infinite options for slider position. I'd prefer less colors, even much less.
- The original EE slider positions (which the presets use) were hand tuned to work well with hundred of games, and it was done before my time.  Especially presets 3/4 were good speedup with very high compatibility. It's hard for me to estimate how the presets behavior would change, and I'm not seeing any summary of tests you performed before and after the change for at least several games. Please test at least several popular games, including some you know are sensitive, in all the preset values, and post some summary of how it compares, including which preset levels don't work before and after, fps in places you choose, fluctuations, or any other observations you think would help others evaluate this change.

MrCK1:
I agree with the points above. In addition, the games I tested showed no glaring issues. 

The only major issue I had was that Vice City froze when I enabled the EE Timing Hack while using the new cycle rate slider. ( I don't know if this was always the case before this PR.)

I still have yet to check Test Drive to make sure that it still passes the loading screen. (FYI, the proper solution requires EE macro mode.)

willkuer:
The problem is that these new settings probably should be checked on weak (cpu bottlenecked) machines and ideally it should be tested against number of cores and for a large amount of games.

Or you just push it (maybe after 1.6) to the devel version and let the community test it. If there are bugs then complains will appear in the forum at some point. I guess that's part of the idea of having a dev build.

avih:
There's also another, more hidden issue. While the "official" goal of this slider is underclock and overclock, i.e. to only affect speeds, it also has a side effect of also affecting compatibility in various ways, more or less subtle, because it changes internal relations.

And while previously there were only few discrete positions, which, especially the "compatible" ones, were tuned manually to break the least number of games, the new "continuous" slider also introduces a very big number of different compatibility options. It's entirely possible, IMO, that e.g. slider values of 150 and 152 work as expected with barely noticeable behavior/speed difference between them, while with 151 all hell breaks loose.

If the slider sole effect was speed, I'd still think we don't need continuous values, but at least its behavior would have been expected. But where each tick could also affect compatibility beyond a minor speed diff, this could make it extremely hard to use, and, IMHO, could do more harm than good.

But then again, without some extensive testing of before and after the change with as many slider positions as possible, it's impossible to assess how it would affect things.

ssakash:
> I still have yet to check Test Drive to make sure that it still passes the loading screen. (FYI, the proper solution requires EE macro mode.)

What do you mean by ``it requires EE macro mode``? As far as I'm aware, all the VU macro mode instructions are implemented in the code. What else needs to be added? (or) is there some sort of special hidden option in the code for enabling some mode?

> The problem is that these new settings probably should be checked on weak (cpu bottlenecked) machines and ideally it should be tested against number of cores and for a large amount of games.

> Please test at least several popular games, including some you know are sensitive, in all the preset values, and post some summary of how it compares, including which preset levels don't work before and after, fps in places you choose, fluctuations, or any other observations you think would help others evaluate this change.

Well, for now I've asked Fezzer to test it, he has lots of games which are influenced by the EE cyclerate and I might probably open a thread soon on the forums for getting more test cases from the forum users.

> Or you just push it (maybe after 1.6) to the devel version and let the community test it. If there are bugs then complains will appear in the forum at some point. I guess that's part of the idea of having a dev build.

I'm planning to get this merged before 1.6, there's a testing phase before the release where the testers on the team do extensive testing for bugs, I could utilize those resources for getting more tests on this change.

I don't like the idea of limiting it to devel build configuration, it's more work on my side (I need to commit it to devel and then eventually move it back to release), better to just properly test it and add it to release version when things are fine.

avih:
> there's a testing phase before the release where the testers on the team do extensive testing for bugs, I could utilize those resources for getting *more* tests on this change.

More tests is great, but where are the first tests results? This is a meaningful change, and _you_ must provide at least some meaningful test results of your own before you expect others to test further.

ssakash:
> you _**must**_ provide at least some meaningful test results of your own before you expect others to test further.

*Must*? -_-

FYI I'm under no obligation to mention about my testing results to you and it's up to me on if I expect my code to be tested by others or not.

Regardless, the objective is to make sure that things work fine compared to before by testing the change on various games and getting broader testcases which could not be done without the help of other testers and I don't see how my limited game collection (which may/may not be influenced much by the EE Cyclerate) can give a meaningful test result, so I trust all the interested testers to handle it for me. I don't care if what I'm doing is morally wrong or right, the main objective still remains - Proper testing is crucial and necessary before merging this pull request.

willkuer:
With devel version I meant the development version (1.7) and not the stable release (1.6). For sure you should enable it for the release build target otherwise it wouldn't get tested.

ssakash:
> How accurate is actually the MHz reading? This would be a cool feature. Like that people might understand what the ee cyclerate hack really does. Accordingly I would call it Emulated Clockspeed or similar and switch MHz and % -> 294 MHz (100%).

The thing is that there is no real theory behind it, the cycle rate is mostly arbitrary on the PS2 EE as it's superscalar. It makes use of [out of order execution](https://en.wikipedia.org/wiki/Out-of-order_execution#Out-of-order_processors) and the cycle stalls of the current instructions also depend upon the stalls of the previous instructions and sometimes those stalls may depend upon the status of data caches and [I don't think PCSX2 handles it accurately](https://github.com/PCSX2/pcsx2/blob/master/pcsx2/R5900OpcodeTables.cpp#L97).

The current calculation entirely depends on the accuracy of the default cycle rate (100%), if that's correct then the other values in the slider would also be right. (which I highly doubt due to the explanation provided above)

Also I'm not sure if ``Emulated clock speed`` is a good word to use? I'm not touching the clockspeed in any way, I'm just modifying the IPC rate. The internal counters still count in the units of EE's 294MHz clock, the only thing which differs is on how fast it counts toward that 294MHz per instruction.

Here's some rough code on how we handle it at the moment,
``` python
# Clock Speed - Number of Clock Cycles per second
# IPC         - Average number of Instructions per Clock Cycle.

# Let's just consider the "MHz" as "Hz" for simpler calculation.
ee_clockspeed = 294
# Let's assume default IPC of EE as 2.
ee_IPC = 2

# With the default values, 
# the number of instructions executed per second would result in 588
instruction_per_second = ee_clockspeed * ee_IPC

#-------------PCSX2 Scenario-----------------
# Users will request for a certain multiplier range between 50-300%
requested_clockspeed = ee_clockspeed * requested_multiplier
requested_instruction_per_second = requested_clockspeed * ee_IPC

# We'll tune our IPC to a certain value to gain the requested clockspeed's 
# instruction throughput. We'll modify the IPC like this,
ee_IPC = requested_instruction_per_second / ee_clockspeed
```

So finally we'll get an instruction throughput equivalent to the clockspeed requested by the user, this is primarily the reason on why I have the clockspeed enclosed in parenthesis instead of the percentage value. Maybe ``EE Clock Modifier`` might be a better word to use?

>  Is there some homebrew that displays the clock speed?

Sorry, I'm not aware of any such homebrews.

Also apologies @avih - I was just in a bad mood yesterday (Too much ego isn't good for my age) and I plan on updating this pull request with my own test results, thanks for the concern.

ssakash:
@FaithLV just noticed a bug on the slider. The slider's text only gets updated when releasing the slider marker and not when dragging it. Pushed a fix for it (I'll be updating the build download link in the first post after appveyor builds the new change)

lightningterror:
Can you add a mark or something to indicate where the default value is ?

ssakash:
I've added marks for the default, positive and negative ends. (The + and - symbols look awkward to me and I don't think we could add custom labels at the end of the slider, we could only display the min/max values over there using the ``wxSL_LABELS`` style)

Also 2:00AM over here, later. 💤

gregory38:
Hum, I think it is a bit artificial.
```
#define DEFAULT_SCALED_BLOCKS() (s_nBlockCycles >> 3)
```
must be big enough to take a percentage value.

gregory38:
```
The thing is that there is no real theory behind it, the cycle rate is mostly arbitrary on the PS2 EE as it's superscalar. It makes use of out of order execution and the cycle stalls of the current instructions also depend upon the stalls of the previous instructions and sometimes those stalls may depend upon the status of data caches and I don't think PCSX2 handles it accurately.
```
It is 2 way superscalar but instructions are executed in order.

masterotaku:
I have tried the build with two games, Kingdom Hearts 2 Final Mix (120fps codes) and Ratchet & Clank 3, with G-Sync + ULMB in DX11 hardware renderer mode to have the best possible smoothness. MHz granularity seems to be working fine. I tried lots of settings and situations with internal fps drops were less and less, until I reached perfection.

It's hard or outright impossible to say specific fps numbers that correspond to specific MHz because the games I tried drop to the next vsync divisor, instead of staying at things like 53fps.

ssakash:
> It is 2 way superscalar but instructions are executed in order.

Oops. you're right. ;)

EE doesn't seem to make use of OoOE, however EE does have it's own method of alleviating concerns of memory latency via the scratchpad ([SPRAM](https://en.wikipedia.org/wiki/Scratchpad_memory)) and the write back buffer. It's probably much cheaper this way but no clue, you're the hardware expert here.

> Edit: any stats will be nice.

Status (or) Statistics? I'd prefer if you could be more clear. 😛 

Statistically, code seems to be fine as I monitored the cycle increment compared to the default cycle rate and the slider options seem to be producing proper results.

About status - I've tested the code on various games and they seem fine, however the interesting testcase is Batman Begins which enters black screen when -2/-3 is set on the slider at the master branch (According to the description, -2 is 60% of EE's insn throughput) and I can reproduce that issue now whenever I go below or equivalent to 60% of clockspeed in the current implementation. So yeah, I'm surprised on how accurate it is as well. 😱

avih:
> Status (or) Statistics

stats is pretty much always statistics, i.e. data, behavior on various use cases, summaries, etc.

ssakash:
> stats is pretty much always statistics, i.e. data, behavior on various use cases, summaries, etc.

Oh, I wondered if he might be asking about the status of the pull request (WIP/Completed..etc) like he usually does. (Also hard to tell if he missed the 'u' or not, best to use the full form to avoid confusion)


-----

lightningterror:
Maybe a checkbox that can be toggled to switch between the old style and the new style will be a good addition.

MrCK1:
@ssakash Not sure what you want to do with this PR, for the time being I think we can all agree that it will be postponed.

ivan89el:
It works well for me. It is desirable to make a range of 25%, some games are missing and performance is increased on weak CPUs.

lightningterror:
I will close this as it's not being worked on for quite some time.

PR #1975, opened 2017-06-22
GSdx-Capture Dialog: Grayout configure button when there's no configuration dialog

ssakash:
**Summary of changes**:

* Grayout the configure button when there is no configuration dialog available for the selected codec.

I think I've had enough of this, I was recently trying out new codecs for making gameplay videos of my favorite games and whenever I wanted to try a new codec I usually check out it's configuration dialog (like most people do) and sometimes the configuration dialog for the codec refuses to open for some reason.

Enduserdilemma

Eventually after looking into it, I found out that there are no configuration dialogs available for such codecs. 😠 Here's where this pull request comes into play - Whenever there's a codec without an available configuration dialog, let's gray out the configure button to avoid problems depicted in the above GIF file.

@bositman: I know that you use lots of codecs for recording videos, could you check if this commit works fine? Thank you.

**Build download link**: [**[Appveyor]**](https://ci.appveyor.com/api/buildjobs/tpjih9459issnfxn/artifacts/pcsx2-v1.5.0-dev-2121-g59de81fa2-3098-vs2015-Win32-AppVeyor.7z)

### DIFF
--- a/plugins/GSdx/GSCaptureDlg.cpp
+++ b/plugins/GSdx/GSCaptureDlg.cpp
@@ -54,6 +54,30 @@ int GSCaptureDlg::GetSelCodec(Codec& c)
 	return 0;
 }
 
+void GSCaptureDlg::UpdateConfigureButton()
+{
+	Codec c;
+	bool enable = false;
+
+	if (GetSelCodec(c) != 1)
+	{
+		EnableWindow(GetDlgItem(m_hWnd, IDC_CONFIGURE), false);
+		return;
+	}
+
+	if (CComQIPtr<ISpecifyPropertyPages> pSPP = c.filter)
+	{
+		CAUUID caGUID;
+		memset(&caGUID, 0, sizeof(caGUID));
+		enable = SUCCEEDED(pSPP->GetPages(&caGUID));
+	}
+	else if (CComQIPtr<IAMVfwCompressDialogs> pAMVfWCD = c.filter)
+	{
+		enable = pAMVfWCD->ShowDialog(VfwCompressDialog_QueryConfig, nullptr) == S_OK;
+	}
+	EnableWindow(GetDlgItem(m_hWnd, IDC_CONFIGURE), enable);
+}
+
 void GSCaptureDlg::OnInit()
 {
 	__super::OnInit();
@@ -112,6 +136,7 @@ void GSCaptureDlg::OnInit()
 		ComboBoxAppend(IDC_CODECS, s.c_str(), (LPARAM)&m_codecs.back(), c.DisplayName == selected);
 	}
 	EndEnumSysDev
+	UpdateConfigureButton();
 }
 
 bool GSCaptureDlg::OnCommand(HWND hWnd, UINT id, UINT code)
@@ -180,6 +205,7 @@ bool GSCaptureDlg::OnCommand(HWND hWnd, UINT id, UINT code)
 	}
 	case IDC_CODECS:
 	{
+		UpdateConfigureButton();
 		break;
 	}
 	case IDOK:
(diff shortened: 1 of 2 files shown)


avih:
I like the improved behavior, but surely you could check if a dialog is available and disable a button accordingly with less than 50 modified LOC plus 30 new ones?!

avih:
Another option would be to test if opening the config dialog succeeded, and if not, pop a message box with an explanation. This would be like 2 LOC.

MrCK1:
I had always dreamed of adding preset resolutions in a drop-down instead of manually entering height and width values.

Integrating newer codecs would be a nice bonus as well, but it doesn't really affect me since I use Shadowplay.

avih:
> I had always dreamed of adding preset resolutions in a drop-down instead of manually entering height and width values.

IIRC at some stage I added a print to the console of the native resolution and/or some other potentially useful info. It didn't make it into a preset though. However, that's unrelated tot his PR.

ssakash:
> Another option would be to test if opening the config dialog succeeded, and if not, pop a message box with an explanation. This would be like 2 LOC.

I don't think it's better compared to just graying it. If we gray the configure button, the user wouldn't even bother clicking it but on your suggestion, the user would click it and get disappointed while we also need to generate an extra popup message box to annoy them further. Maybe not a good idea?

On the other hand, I don't really understand why the number of lines modified is a concern. (I think most of those come from changing the ``OnCommand`` event entries of the dialog into a switch-case for code readability and for easier event additions in the future )

> IIRC at some stage I added a print to the console of the native resolution

You can't just get the standard native resolution of the game, you can only get the current size of the output circuit. So if users start recording before the output circuit is initialized then your message would probably print 0x0.

lightningterror:
Hmm I have a couple of suggestions for the recorder tho they are not really related to this pr.

Record using the mp4 format instead of avi since it provides many benefits over avi.
Already assigned default save location "Videos/Recordings" or something in Documents.
Default capture resolution should be the same as what the game is rendered at. (example 3x native).People ofc can pick a custom resolution.

avih:
> Record using the mp4 format instead of avi since it provides many benefits over avi

It's not a problem converting it to any format you like. Also, the current recording code uses VFW AFAIK, which outputs avi. If anything, I'd improve it by outputting the correct aspect ratio, but not sure if avi supports non square pixels - which we need.

And using a different encoding backend is at the very least considerable work for not much gain. However, if someone wants to use some HW encoding, then this could speed things up nicely. But again, research and work.

gregory38:
LOC isn't an issue. Besides there are formatting in the middle.

@bositman did you have time to give this PR a shot ?

plugins/GSdx/GSCaptureDlg.cpp
```diff
@@ -113,83 +136,89 @@ void GSCaptureDlg::OnInit()
 		ComboBoxAppend(IDC_CODECS, s.c_str(), (LPARAM)&m_codecs.back(), c.DisplayName == selected);
 	}
 	EndEnumSysDev
+	UpdateConfigureButton();
 }
 
 bool GSCaptureDlg::OnCommand(HWND hWnd, UINT id, UINT code)
 {
-	if(id == IDC_BROWSE && code == BN_CLICKED)
+	switch (id)
```
turtleli: I'm not sure this change to switch + misc formatting is really worth it. If you have to do it then I'd prefer you put the change in a separate commit.

bositman:
Tested the button behavior, seems fine and does what is described. I agree with @turtleli on separating commits as this could be easily mistaken as a commit that could not break something other than the capture dialog (in a worst case scenario it introduces some kind of regression)

gregory38:
@ssakash could you please the commit properly please.

ssakash:
> @ssakash could you please the commit properly please.

Done.

Issue #1986, opened 2017-06-29
Beyond good and evil : broken graphics


atomic83GitHub:
***PCSX2 version:***
Last 1.5 git

***PCSX2 options:***
All default (Clamping and Round Mode don't do anything) , Only MTVU is enebeled (bug also without).

***Plugins used:***
All default (GSDX)
![image](https://user-images.githubusercontent.com/26351275/27700618-6f08328a-5cfe-11e7-9cc7-7a2f6d45a223.png)
![image](https://user-images.githubusercontent.com/26351275/27700640-7b46c228-5cfe-11e7-85d0-77d85d2965ac.png)

***Description of the issue:***
The game Beyond good and evil have some graphical issues

1.SPS 
![sps](https://user-images.githubusercontent.com/26351275/27700144-f8a329ca-5cfc-11e7-9f4a-66d10a90acd5.JPG)
2. DirectX hardware issues (due to the lack of the death shader)
![probleme dx](https://user-images.githubusercontent.com/26351275/27700207-29e23b20-5cfd-11e7-82ff-8f860da547fb.JPG)
3.Broken water effect (heavy performance impact)
![eau](https://user-images.githubusercontent.com/26351275/27700262-5c8ce5ac-5cfd-11e7-9fd4-36efdefdd5d3.JPG)
4.The shadows are broken in hardware mode
![ombres](https://user-images.githubusercontent.com/26351275/27700298-75dda7f8-5cfd-11e7-8cf6-aaa202f53c7e.JPG)
5.The software render have a bug at the bottom of the screen
![software bug](https://user-images.githubusercontent.com/26351275/27700480-0789d730-5cfe-11e7-962f-2ba0bd35f48e.JPG)
6.Cinematics pictures are missing in hardware mode
Hardware:
![probleme cinematique](https://user-images.githubusercontent.com/26351275/27700530-26ce6e76-5cfe-11e7-9a9a-07571559aa27.JPG)
Software:
![software cinematique](https://user-images.githubusercontent.com/26351275/27700543-312fc982-5cfe-11e7-90df-2453c282ec3b.JPG)
7.Acurate date cause a crash in water areas.

***How to reproduce the issue:***
Set OpenGL or DX 9/11 in GSDX and start Beyond good and evil.

***Last known version to work:***
Don't exist,The game was also broken on 1.4 and other versions

***PC specifications:***
Intel Core 2 Duo E4600 , GTX 780Ti , OS: Windows 10 64 Bit

[...]

PR #1988, opened 2017-06-30
GSdx: Rework framebuffer height calculation

ssakash:
**Summary of changes**

* Improve framebuffer height calculation by also consider the DBY value of the DISPFB register. (For more information, check the commit message) **Fixes software mode issue mentioned on #1986**

Tested GS Dumps of Nascar, Beyond Good and Evil, [FMV bottom issue](https://github.com/PCSX2/pcsx2/issues/1562#issuecomment-270256331) reported by @MrCK1 and Crash Bash. All of them are fine with this, more tests would also be nice.

Also I added this to the 1.6 milestone list as this needs to be merged before the 1.6 release, not because it's a big progress or anything. :P

### DIFF
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -474,6 +474,10 @@ GSVector4i GSState::GetDisplayRect(int i)
 
 GSVector4i GSState::GetFrameRect(int i)
 {
+	// If no specific context is requested then pass the merged rectangle as return value
+	if (i == -1)
+		return GetFrameRect(0).runion(GetFrameRect(1));
+
 	GSVector4i rectangle = GetDisplayRect(i);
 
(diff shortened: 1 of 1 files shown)


plugins/GSdx/GSState.cpp
```diff
@@ -474,6 +474,10 @@ GSVector4i GSState::GetDisplayRect(int i)
 
 GSVector4i GSState::GetFrameRect(int i)
 {
+	// If no specific context is requested then pass the merged rectangle as return value
+	if (i == -1)
+		return GetFrameRect(0).runion_ordered(GetFrameRect(1));
```
gregory38: Use safer `runion` instead and same below too.

lightningterror:
Any update on this? The changes have been made for a while now.

Issue #1979, opened 2017-06-25
Finding Nemo (The Game) Graphical regression with OpenGL


atomic83GitHub:
***PCSX2 version:***
Last 1.5 dev build
***PCSX2 options:***

(The options are the same between the 2 versions)

Core:
Default settings 
GSDX
![image](https://user-images.githubusercontent.com/26351275/27515509-fe2f4afa-59a4-11e7-9508-b937763bc042.png)
![image](https://user-images.githubusercontent.com/26351275/27515514-172694be-59a5-11e7-98db-aa22b2b26f86.png)
***Plugins used:***
The last GSDX version of the 2 versions
***Description of the issue:***
Fiding Nemo don't look great with OpenGL in newer versions , look:
1.4.0
![image](https://user-images.githubusercontent.com/26351275/27515529-97158cf2-59a5-11e7-8909-7e04827f6b91.png)
1.5 dev
![image](https://user-images.githubusercontent.com/26351275/27515552-ec8e2748-59a5-11e7-8e67-05df0606318b.png)
***How to reproduce the issue:***
Load PCSX2 set OpenGL Hardware in GSDX and load Fiding Nemo
***Last known version to work:***
1.4.0 Stable version
***PC specifications:***
Intel Core 2 duo E8600 , GTX 780ti

ssakash:
Just to get things straight,

* 1.4 + custom resolution (1920x1080) is good ✔️ 
* 1.5 (Latest development builds) + custom resolution is not good 🚫 

That's the case, right? Is the issue reproducible in native/ native scaling resolutions (2x,3x,4x...etc)? By the way, finding the first working version which caused the issue would always help make the developer's life easier. ;)

atomic83GitHub:
Native resolution do the same error , and it wil be monger to found but i will made it :)

atomic83GitHub:
Longer sorry

atomic83GitHub:
Sorry but I am uncapable to found a working 1.5 version beacause the first 1.5 build on build orphis have the issue!

atomic83GitHub:
I have found a working version wich is the 1.5 dev 31 
The bug is introduced between the 1.5 dev 31 and 1.5 dev 66

atomic83GitHub:
I can confirm the death shader not working as it should look:
1.4
Without the death shader:
![image](https://user-images.githubusercontent.com/26351275/27516827-8fb80cac-59c1-11e7-827c-3f1182938410.png)
With the death shader:
[...]

PR #1994, opened 2017-07-02
gsdx ogl: handle float integral coordinate for depth sampling

gregory38:
Fix depth effect regression on Nemo (issue #1979)


lightningterror:
https://github.com/PCSX2/pcsx2/pull/1995#issuecomment-314830995

Seems it's ready for merge.

PR #1995, opened 2017-07-02
glsl: reduce geometry shader complexity

gregory38:
output 1 strip of 2 triangles instead of 2 strips of 1 triangle.

Potentially it would reduce the geometry shader overhead. And it
might avoid a middle line in sprite in some AMD GPU/driver/OS
bad combination

Need a quick test on various driver, AMD, Nvidia, Intel.

### DIFF
--- a/plugins/GSdx/res/glsl/tfx_vgs.glsl
+++ b/plugins/GSdx/res/glsl/tfx_vgs.glsl
@@ -93,7 +93,7 @@ struct vertex
     vec4 c;
 };
 
-void out_vertex(in vertex v)
+void out_vertex(in vec4 position, in vertex v)
 {
     GSout.t_float  = v.t_float;
     GSout.t_int    = v.t_int;
@@ -104,6 +104,7 @@ void out_vertex(in vertex v)
 #else
     GSout.fc       = GSin[1].fc;
 #endif
+    gl_Position = position;
     gl_PrimitiveID = gl_PrimitiveIDIn;
     EmitVertex();
 }
@@ -113,7 +114,7 @@ layout(points) in;
 #else
 layout(lines) in;
 #endif
-layout(triangle_strip, max_vertices = 6) out;
+layout(triangle_strip, max_vertices = 4) out;
 
 #if GS_POINT == 1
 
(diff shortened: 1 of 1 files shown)


lightningterror:
Wish I could test this now but i can't on both intel and amd. eta 2 weeks maybe.
I'd like this to be added before 1.6 release. More fps is always welcome :)

@Sarania can you do some tests?

MrCK1:
Works fine for me on Nvidia.

lightningterror:
@MrCK1 can you do some benchmarks?

MrCK1:
I can, but not until tonight.

gregory38:
Don't expect too much. First of all, Opengl barely uses geometry shader. And compiler might already do the job correctly. Hum I will check the shader size with Nouveau.

gregory38:
Here some stats on Nouveau compiler. `GPR` is the number of used register (less register means more shader run in //). `Inst` is the number of instruction (less instruction means faster shader run). `bytes` is the size of the shader. First shader is to convert sprite. And the next 2 are used in the unscale line/point hack.

Before
```
T:Oth		S:	=> type: 2, local: 0, gpr: 34, inst: 187, bytes: 1712
T:Oth		S:	=> type: 2, local: 0, gpr: 29, inst: 197, bytes: 1808
T:Oth		S:	=> type: 2, local: 0, gpr: 44, inst: 211, bytes: 1936
```

After
```
T:Oth		S:	=> type: 2, local: 0, gpr: 37, inst: 133, bytes: 1216
T:Oth		S:	=> type: 2, local: 0, gpr: 30, inst: 142, bytes: 1304
T:Oth		S:	=> type: 2, local: 0, gpr: 33, inst: 147, bytes: 1344
```

mirh:
> avoid a middle line in sprite

Wait, like in the actual finished rendered image I might see some improvement?
I could give a check to it tomorrow then

gregory38:
I remembered someone posted some screenshot on an AMG GPU that show sprite diagonal (border of 2 triangles). Normally rasterization rules are obvious. Only a single fragment must be generated at the border. I don't know if it was due to the 2 sprites or a random bug in the driver. Anyway, code is better this way. And maybe someone could do the same for latest added shader.

gregory38:
I have a new branch with all black magic. I will review my code tomorrow and make a new PR. The branch also implements correctly all the logic for late vsync but no GUI. I don't how to export it properly on the GUI. (Either combo box or a gsdx option, other?)

mirh:
> I have a new branch with all black magic. I will review my code tomorrow and make a new PR. The branch also implements correctly all the logic for late vsync but no GUI. I don't how to export it properly on the GUI. 

I think you posted in the wrong issue 😛

gregory38:
Anyone to test Nemo ? @prafullpcsx2 I think you have the game, isn't it.

prafullpcsx2:
This doesn't show any difference with Nemo. I guess you meant https://github.com/PCSX2/pcsx2/pull/1994 ? That one surely fixes the Nemo regression.

lightningterror:
Looks like greg posted in the wrong thead again :D

gregory38:
Ah yes. I'm useless

PR #1996, opened 2017-07-02
Gamedb lightningterror

lightningterror:


lightningterror:
Ups. Treat this as a misclick.

Issue #1971, opened 2017-06-11
Bug/Feature request: Disable vsync temporarily while frame limiting is disabled


arifogel:
***PCSX2 version:***
1.4.0

***PCSX2 options:***
vsync, HW OpenGL renderer, vsync, frame limiting

***Description of the issue:***
Frame limiting Turbo/Slow mode has no effect while vsync is enabled. This makes technical sense due to the nature of vsync. However, it is dumb to make turbo unavailable. Instead, vsync should be temporarily disabled while turbo mode is enabled.
This is especially cumbersome due to the fact that:
  1.  Vsync setting on linux has no effect until application restart. (Also needs to be fixed. Let me know if I should file another issue).
  2.  There is no keyboard shortcut to toggle vsync.

***How to reproduce the issue:***
Enable vsync. Then frame limiting setting does nothing.

***PC specifications:***
OS: Ubuntu 16.04 64-bit
Video Card: Nvidia GTX 970
Video Driver: NVIDIA-Linux-x86_64-381.22
CPU: Intel(R) Core(TM) i7-5930K CPU

arifogel:
Follow-up of #842

mirh:
It is dumb to make turbo unavailable, but just because you are making dumb assumption over which refresh rates the monitor of your users are gonna support. 
Proper behavior, imo, should just be to allow this, end. 

And if you want to actually have turbo on your I guess 60hz screen, disable v-sync.

arifogel:
@mirh I am unclear on what point you are trying to make here. I am not a maintainer of this project, and thus I do not have users. I am not making any assumptions on refresh rates of other users; I personally use 60 hz, which I think is the most common refresh rate of users today.
For me, on my 60hz monitor, vsync gets rid of screen tearing, which is very visible and jarring otherwise. I would like to be able to switch between nice-looking vsync and time-saving turbo without:

  1.  saving state
  2.  leaving full screen
  3.  pulling another window to the foreground
  4.  opening graphics settings
  5.  disabling a checkbox
  6.  clicking ok
  7.  exiting the application
  8.  restarting the application
  9.  reloading my savestate
  10.  hitting the shortcut to enable turbo

Instead, it would be great if either one of two workflows were available instead:
   1. Press shortcut to toggle turbo mode, and vsync is automatically temporarily disabled.

OR
   1. Presss shortcut key to toggle vsync.
   2. Press shortcut to toggle turbo mode.

[...]

Issue #1642, opened 2016-10-30
Enabling Speedhack presets disables (grayed out) Vsync option


psennermann:
**_PCSX2 version:**_
1.5.0-1498

**_Description of the issue:**_
`(What's the problem? - Screenshots showing the issue if applicable)`

Enabling the preset option (from 1 to 6) in the pcsx2 emulation/configuration options, grayes out the Vsync option (which is funny, since Vsync is not a speedhack), so the only way to enable Vsync in PCSX2 is to disable the speedhack preset.

**_How to reproduce the issue:**_
Enable the speedhack preset option, set it from 1 to 6 and see that the Vsync option is grayed out

**_Last known version to work:**_
I think Vsync has always been managed this way

**_PC specifications:**_
CPU: Intel Core 2 Duo E8600 GPU: GTX 950 OS: Windows 10 64 Bit

FlatOutPS2:
Preset is not just for speedhacks, it is a general emulation options preset and keeps unexperienced users from ruining their experience by enabling the wrong options. If you want to enable advanced features like vsync, which could cause emulation issues, you'll just have to disable the preset.

trostboot:
Care to elaborate on issues caused by vsync? It's been my experience that, if anything, leaving vsync off is what causes issues in newer Windows versions where vsync is always enforced through the DWM for non-exclusive applications like PCSX2, resulting in poor framepacing (although even that might just happen under certain conditions, I've never tried to figure it out in more detail once I've noticed enabling vsync in PCSX2 fixes it).

jcdenton2k:
Windows Desktop Composition (Desktop Windows Manager) is always on in Windows 8 and later:
https://msdn.microsoft.com/en-us/library/hh848042.aspx

The Display Driver Must Support Vsync-Control:
https://msdn.microsoft.com/en-us/library/ff569520(VS.85).aspx

DWM Info:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa969540(v=vs.85).aspx

On Windows 7 it can be disabled but enabling Vsync without Triple Buffering can increase input latency in certain scenarios.  Having both Vsync AND Triple Buffering mitigates this issue entirely BUT it requires significantly more RAM consumption (and specifically on-board GPU ram).

DWM doesn't give me any issues on Win7 x64 personally, but others have experienced issues depending on their setups.

More details on vsync + triple buffering in detail (article from 2009 but still relevant):
http://www.anandtech.com/show/2794/2

[...]

Issue #1075, opened 2015-12-30
Issue with Vsync and OpenGL Fullscreen tearing


tsunami2311:
http://forums.pcsx2.net/Thread-Slowdown-when-using-V-sync-and-OpenGL

Sagnip:
Hello, I'm the original poster on the forum. Just to make things simpler I will explain the problem here:

Basically, If you enable Vsync while using the OpenGL backend it causes slowdown.
This occurs when using the PCSX2 setting, and also occurs if you activate Vsync through driver software (with setting off in PCSX2).
This occurs in all games, regardless of any other settings (or at least any I've tested).
Slowdown only occurs in fullscreen when Vsync is on (because Vsync only takes effect in fullscreen).

tsunami2311:
forgot the part about opengl in fullscreen is tearing, before one didnt need to check vsync in pcsx2 keep fullscreen from tearing, my original coment about it could be something nvidia mucked up in 361.43 drivers, and i to lazy to down grade to find outm if it drivers or pcsx2 builds.

I know there was some change in reguards to vsync a while back, not sure if that has anything to do with it

Sagnip:
Yep, that is probably it, that is why I was getting confused in on the forum because I'm almost certain OGL is in exclusive fullscreen. There are quite a few things that point towards this that I mentioned there, I had just assumed all the backends were that way but after looking OGL seems to be the only one using exclusive.

So, basically we need 2 fixes:

1. Make OGL's fullsceen borderless fullscreen (This will eliminate the need for Vsync for those who have Aero running.)
2. fix Vsync in OGL for people who don't have Aero.

Blackbird88:
Can confirm. With D3D11 there is no tearing at all in fullscreen (I don't even need Vsync) 
While with OpenGL in fullscreen there is tearing everywhere.

vsub:
There is something else too...while vsync is enabled and using OpenGL,you can't go above your screen refresh rate when you disable the fps limit...you can't also use the Turbo(TAB)

You can while using DX and you still will not see screen tearing even if you are running at above full speed

You see screen tearing even while using slow motion while vsync disabled in OGL mode

Sagnip:
seems about right OGL in 16:9 is exclusive, needs to be changed.

willkuer:
All of you know that we have like one thread per week requesting exclusive fullscreen instead of borderless windowed because borderless is soooo bad?

Sagnip:
[...]

PR #2000, opened 2017-07-04
Greg/vsync take2

gregory38:
Replace PR #1950 

This PR factorize some code in the OpenGL window API to allow adaptive Vsync. And automatically detect DWM crazyness. It misses a GUI update to add adaptive Vsync. Plan is to adapt commit of the previous PR. Meanwhile people can test the various driver behavior.

### DIFF
--- a/pcsx2/gui/GlobalCommands.cpp
+++ b/pcsx2/gui/GlobalCommands.cpp
@@ -91,7 +91,6 @@ namespace Implementations
 		}
 		else if( g_LimiterMode == Limit_Turbo )
 		{
-			GSsetVsync( g_Conf->EmuOptions.GS.VsyncEnable );
 			g_LimiterMode = Limit_Nominal;
 
 			if ( g_Conf->Framerate.SkipOnLimit)
@@ -107,7 +106,6 @@ namespace Implementations
 		}
 		else
 		{
-			GSsetVsync( false );
 			g_LimiterMode = Limit_Turbo;
 
 			if ( g_Conf->Framerate.SkipOnTurbo)
@@ -121,7 +119,11 @@ namespace Implementations
 				g_Conf->EmuOptions.GS.FrameSkipEnable = false;
 			}
 		}
+
 		gsUpdateFrequency(g_Conf->EmuOptions);
+
+		GSsetVsync(g_Conf->EmuOptions.GS.GetVsync());
+
 		pauser.AllowResume();
 	}
 
@@ -134,7 +136,6 @@ namespace Implementations
 		// out a better consistency approach... -air
 
 		ScopedCoreThreadPause pauser;
-		GSsetVsync( g_Conf->EmuOptions.GS.VsyncEnable );
 		if( g_LimiterMode == Limit_Slomo )
 		{
 			g_LimiterMode = Limit_Nominal;
@@ -146,7 +147,11 @@ namespace Implementations
 			OSDlog( Color_StrongRed, true, "(FrameLimiter) SlowMotion ENABLED." );
 			g_Conf->EmuOptions.GS.FrameLimitEnable = true;
 		}
+
 		gsUpdateFrequency(g_Conf->EmuOptions);
+
+		GSsetVsync(g_Conf->EmuOptions.GS.GetVsync());
+
 		pauser.AllowResume();
 	}
 
@@ -154,8 +159,17 @@ namespace Implementations
 	{
 		ScopedCoreThreadPause pauser;
 		g_Conf->EmuOptions.GS.FrameLimitEnable = !g_Conf->EmuOptions.GS.FrameLimitEnable;
-		GSsetVsync( g_Conf->EmuOptions.GS.FrameLimitEnable && g_Conf->EmuOptions.GS.VsyncEnable );
 		OSDlog( Color_StrongRed, true, "(FrameLimiter) %s.", g_Conf->EmuOptions.GS.FrameLimitEnable ? "ENABLED" : "DISABLED" );
+
+		// Turbo/Slowmo don't make sense when framelimiter is toggled
+		g_LimiterMode = Limit_Nominal;
+
+		// Disable Vsync when frame limited is disabled
+		if( g_Conf->EmuOptions.GS.FrameLimitEnable )
+			GSsetVsync(g_Conf->EmuOptions.GS.GetVsync());
+		else
+			GSsetVsync(0);
+
 		pauser.AllowResume();
 	}
 
--- a/plugins/GSdx/GSWndWGL.cpp
+++ b/plugins/GSdx/GSWndWGL.cpp
@@ -139,6 +139,11 @@ void GSWndWGL::DetachContext()
 	}
 }
 
+void GSWndWGL::PopulateWndGlFunction()
+{
+	m_swapinterval = (PFNWGLSWAPINTERVALEXTPROC)wglGetProcAddress("wglSwapIntervalEXT");
+}
+
 bool GSWndWGL::Attach(void* handle, bool managed)
 {
 	m_NativeWindow = (HWND)handle;
@@ -146,13 +151,7 @@ bool GSWndWGL::Attach(void* handle, bool managed)
 
 	OpenWGLDisplay();
 
-	CreateContext(3, 3);
-
-	AttachContext();
-
-	m_swapinterval = (PFNWGLSWAPINTERVALEXTPROC)wglGetProcAddress("wglSwapIntervalEXT");
-
-	PopulateGlFunction();
+	FullContextInit();
 
 	UpdateWindow(m_NativeWindow);
 
@@ -288,16 +287,9 @@ bool GSWndWGL::Create(const string& title, int w, int h)
 
 	OpenWGLDisplay();
 
-	CreateContext(3, 3);
-
-	AttachContext();
-
-	m_swapinterval = (PFNWGLSWAPINTERVALEXTPROC)wglGetProcAddress("wglSwapIntervalEXT");
-
-	PopulateGlFunction();
+	FullContextInit();
 
 	return true;
-
 }
 
 //Same as DX
--- a/plugins/GSdx/GSWndWGL.cpp
+++ b/plugins/GSdx/GSWndWGL.cpp
@@ -40,7 +40,7 @@ static void win_error(const char* msg, bool fatal = true)
 
 
 GSWndWGL::GSWndWGL()
-	: m_NativeWindow(NULL), m_NativeDisplay(NULL), m_context(NULL)
+	: m_NativeWindow(nullptr), m_NativeDisplay(nullptr), m_context(nullptr), m_has_late_vsync(false)
 {
 }
 
@@ -142,6 +142,15 @@ void GSWndWGL::DetachContext()
 void GSWndWGL::PopulateWndGlFunction()
 {
 	m_swapinterval = (PFNWGLSWAPINTERVALEXTPROC)wglGetProcAddress("wglSwapIntervalEXT");
+
+	// To ease the process, extension management is itself an extension. Clever isn't it!
+	PFNWGLGETEXTENSIONSSTRINGARBPROC wglGetExtensionsStringARB = (PFNWGLGETEXTENSIONSSTRINGARBPROC)wglGetProcAddress("wglGetExtensionsStringARB");
+	if (wglGetExtensionsStringARB) {
+		const char* ext = wglGetExtensionsStringARB(m_NativeDisplay);
+		m_has_late_vsync = m_swapinterval && ext && strstr(ext, "WGL_EXT_swap_control_tear");
+	} else {
+		m_has_late_vsync = false;
+	}
 }
 
 bool GSWndWGL::Attach(void* handle, bool managed)
--- a/plugins/GSdx/GSWndWGL.cpp
+++ b/plugins/GSdx/GSWndWGL.cpp
@@ -315,7 +315,7 @@ void* GSWndWGL::GetProcAddress(const char* name, bool opt)
 
 //TODO: check extensions supported or not
 //FIXME : extension allocation
-void GSWndWGL::SetVSync(int vsync)
+void GSWndWGL::SetSwapInterval(int vsync)
 {
 	// m_swapinterval uses an integer as parameter
 	// 0 -> disable vsync
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -63,6 +63,13 @@ enum GamefixId
 	GamefixId_COUNT
 };
 
+enum class VsyncMode
+{
+	Off,
+	On,
+	Adaptive,
+};
+
 // Template function for casting enumerations to their underlying type
 template <typename Enumeration>
 typename std::underlying_type<Enumeration>::type enum_cast(Enumeration E)
@@ -291,9 +298,9 @@ struct Pcsx2Config
 		bool	DisableOutput;
 		int		VsyncQueueSize;
 
-		bool	FrameLimitEnable;
-		bool	FrameSkipEnable;
-		bool	VsyncEnable;
+		bool		FrameLimitEnable;
+		bool		FrameSkipEnable;
+		VsyncMode	VsyncEnable;
 
 		int		FramesToDraw;	// number of consecutive frames (fields) to render
 		int		FramesToSkip;	// number of consecutive frames (fields) to skip
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -305,6 +305,8 @@ struct Pcsx2Config
 		GSOptions();
 		void LoadSave( IniInterface& conf );
 
+		int GetVsync() const;
+
 		bool operator ==( const GSOptions& right ) const
 		{
 			return
--- a/pcsx2/MTGS.cpp
+++ b/pcsx2/MTGS.cpp
@@ -203,7 +203,7 @@ void SysMtgsThread::OpenPlugin()
 		result = GSopen( (void*)pDsp, "PCSX2", renderswitch ? 2 : 1 );
 
 
-	GSsetVsync(EmuConfig.GS.FrameLimitEnable && EmuConfig.GS.VsyncEnable);
+	GSsetVsync(EmuConfig.GS.GetVsync());
 
 	if( result != 0 )
 	{
--- a/pcsx2/Pcsx2Config.cpp
+++ b/pcsx2/Pcsx2Config.cpp
@@ -20,6 +20,7 @@
 #include "Utilities/IniInterface.h"
 #include "Config.h"
 #include "GS.h"
+#include "gui/GSFrame.h"
 
 void TraceLogFilters::LoadSave( IniInterface& ini )
 {
@@ -232,6 +233,14 @@ void Pcsx2Config::GSOptions::LoadSave( IniInterface& ini )
 
 	IniEntry( FramesToDraw );
 	IniEntry( FramesToSkip );
+}
+
+int Pcsx2Config::GSOptions::GetVsync() const
+{
+	if (g_LimiterMode == Limit_Turbo)
+		return 0;
+
+	return VsyncEnable;
 }
 
 const wxChar *const tbl_GamefixNames[] =
--- a/pcsx2/Pcsx2Config.cpp
+++ b/pcsx2/Pcsx2Config.cpp
@@ -201,7 +201,7 @@ Pcsx2Config::GSOptions::GSOptions()
 {
 	FrameLimitEnable		= true;
 	FrameSkipEnable			= false;
-	VsyncEnable				= false;
+	VsyncEnable				= VsyncMode::Off;
 
 	SynchronousMTGS			= false;
 	DisableOutput			= false;
@@ -225,7 +225,7 @@ void Pcsx2Config::GSOptions::LoadSave( IniInterface& ini )
 
 	IniEntry( FrameLimitEnable );
 	IniEntry( FrameSkipEnable );
-	IniEntry( VsyncEnable );
+	ini.EnumEntry( L"VsyncEnable", VsyncEnable, NULL, VsyncEnable );
 
 	IniEntry( LimitScalar );
 	IniEntry( FramerateNTSC );
@@ -240,7 +240,15 @@ int Pcsx2Config::GSOptions::GetVsync() const
 	if (g_LimiterMode == Limit_Turbo)
 		return 0;
 
-	return VsyncEnable;
+	// D3D only support a boolean state. OpenGL waits a number of vsync
+	// interrupt (negative value for late vsync).
+	switch (VsyncEnable) {
+		case VsyncMode::Adaptive: return -1;
+		case VsyncMode::Off: return 0;
+		case VsyncMode::On: return 1;
+
+		default: return 0;
+	}
 }
 
 const wxChar *const tbl_GamefixNames[] =
--- a/pcsx2/gui/AppCorePlugins.cpp
+++ b/pcsx2/gui/AppCorePlugins.cpp
@@ -359,9 +359,8 @@ bool AppCorePlugins::OpenPlugin_GS()
 
 	bool retval = _parent::OpenPlugin_GS();
 
-	if( g_LimiterMode == Limit_Turbo )
-		GSsetVsync( false );
-		
+	GSsetVsync(EmuConfig.GS.GetVsync());
+
 	return retval;
 }
 
--- a/pcsx2/gui/AppCoreThread.cpp
+++ b/pcsx2/gui/AppCoreThread.cpp
@@ -356,7 +356,7 @@ static void _ApplySettings( const Pcsx2Config& src, Pcsx2Config& fixup )
 	if( overrides.ProfilingMode )
 	{
 		fixup.GS.FrameLimitEnable = false;
-		fixup.GS.VsyncEnable = false;
+		fixup.GS.VsyncEnable = VsyncMode::Off;
 	}
 
 	wxString gameCRC;
--- a/pcsx2/gui/Panels/ConfigurationPanels.h
+++ b/pcsx2/gui/Panels/ConfigurationPanels.h
@@ -267,6 +267,7 @@ namespace Panels
 	{
 	protected:
 		wxComboBox*		m_combo_AspectRatio;
+		wxComboBox*		m_combo_vsync;
 
 		wxTextCtrl*		m_text_Zoom;
 
--- a/pcsx2/gui/Panels/GSWindowPanel.cpp
+++ b/pcsx2/gui/Panels/GSWindowPanel.cpp
@@ -32,10 +32,21 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 		_("Widescreen (16:9)")
 	};
 
+	// Warning must match the order of the VsyncMode Enum
+	const wxString vsync_label[] =
+	{
+		_("Disabled"),
+		_("Standard"),
+		_("Adaptive"),
+	};
+
 	m_text_Zoom = CreateNumericalTextCtrl( this, 5 );
 
 	m_combo_AspectRatio	= new wxComboBox( this, wxID_ANY, wxEmptyString, wxDefaultPosition, wxDefaultSize,
 		ArraySize(aspect_ratio_labels), aspect_ratio_labels, wxCB_READONLY );
+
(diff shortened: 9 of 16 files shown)


lightningterror:
Maybe @ssakash can do something about the GUI.

ssakash:
Nah, no worries. @gregory38 is way more competent than me when it comes to coding at all aspects. ;)

lightningterror:
He doesn't use Windows tho.

Embrace it.
https://www.youtube.com/watch?v=G6b5ubvfa3I

ssakash:
@gregory38 

Don't hesitate to ping me if you need any help on Windows side.

mirh:
> GSWndWGL::CompositorEnabled

I guess that.. in light of the remainder checks, the function could be named CompositorAffected, CompositorHandled, UnderCompositor or something?
I mean, I know there are other emus that are lurking at these PRs.. And perhaps the whole rationale is not entirely clear as in a 100 comments discussion. 

> if (!dwm)

A XP fallback 😆

> if (GLLoader::nvidia_buggy_driver)

Cool naming standardization there. 
I would suggest to add a `TODO: figure out specifically affected circumstances.` maybe?
`Is v-sync still enforced by driver even on W7 with dwm off?`
`Are driver versions below 380.xx branch affected?`

> GSVector4i surface = GetClientRect();

You could mention above this line that this should make for the amd/intel case instead. 

---
Anyway, I tested the thing
And something fishy is going on, because everytime I toggle fullscreen mode (be it exclusive or non exclusive!) unlimited framerate get clamped to 60FPS. 
Even if I have v-sync disabled. 

Maybe you should check surface**!=**window for truth?
Funnily, I was now thinking that since this comparison implicitly also takes into account the window border (which is 0 only when in fullscreen afaik?) we might as well not need the monitor size check.

gregory38:
@lightningterror @ssakash 
Well as I said, there is already some code for wx on the other PR. I just need to tune it. And to take a decision on "string" translation and preset behavior. However I need to some help to test & debug this code. I didn't test anything on it. My current driver doesn't support adaptive Vsync and ofc I'm on Linux.

@mirh 
Hum yes you're right condition is bad. And yes I will add some comments.

mirh:
> My current driver doesn't support adaptive Vsync

Ehrm.. What is GLX_MESA_swap_control for ?

mirh:
> BOOL dwm;

Wonder: isn't bool initialized to a random value this way?

That in the event of DwmIsCompositionEnabled failing, would make for !dwm check (aka *is not 0*?) meaningless?

> Recent driver (+380.xx) seem to never bypass the compositor.

Actually, as I was saying I'm unsure. 
`Recent driver (+380.xx ??) seem to always get v-sync enforced someway.` would seem more accurate. 

> At least not with current wxWindow setup.

Mhh, dunno. My educate guess over nvidia dev forums, judge this more spread than just wx, or any special case we may be falling under. 

> if (CompositorEnabled() && m_vsync != 0)

I'm not sure we can drop dwmflush from the "I don't want v-sync" use-case - if we still want to avoid stuttering. 
But since I'm no eagle, I'll let other judge on this. 

Anyway, for some reason I'm never getting v-sync to begin with (albeit it was kind of emotional to *finally* see tearing). 

...

> if (CompositorEnabled()) SetSwapInterval(0)

What happens to the CompositorEnabled function when none of the paths you placed are hit?
[remember me why we are not just using good old *if 1 OR 2 OR 3 plus elsing*?]

ssakash:
> Wonder: isn't bool initialized to a random value this way?
That in the event of DwmIsCompositionEnabled failing, would make for !dwm check (aka is not 0?) meaningless?

No, ``dwm`` will be initialized after the call to ``DwmIsCompositionEnabled()``.

mirh:
Oh, I see thank you. Now, if that's good then:
> // Only possible on Win7

Should maybe have a ***≤***. 

Then, I was thinking: if we are always performing the check, we really don't need WM_DWMCOMPOSITIONCHANGED. 
But.. Perhaps isn't a bit too much rude/heavy to indeed ALWAYS check the thing at every Flip?

gregory38:
@mirh it would be nice to comment in the code directly
> I'm not sure we can drop dwmflush from the "I don't want v-sync" use-case - if we still want to avoid stuttering.

The question is what is the behavior of no-vsync + no dwm wait. Are you locked to 60fps but with stuttering or can you go lower/higher ?

gregory38:
> Then, I was thinking: if we are always performing the check, we really don't need WM_DWMCOMPOSITIONCHANGED.
But then.. Perhaps isn't a bit too much rude/heavy to indeed ALWAYS check the thing at every Flip?

I'm not sure the event will be generated if we resize the surface. 60 check by seconds seems to be free vs MFLOPS/GFLOPS of a CPU.

mirh:
> The question is what is the behavior of no-vsync + no dwm wait. 

It's not even a question, because if you think to it.. That's literally whatever we always have now by default. 

> 60 check by seconds seems to be free vs MFLOPS/GFLOPS of a CPU.

Not sure I have understood?

gregory38:
The question was in case the dwm compositor is still enabled. For example in not-full screen mode. Are we still locked at 60 fps with stuttering. Or can we go higher ? lower? 

CPUs are capable of millions of operations per seconds. A couple of instructions every vsync is peanut

gregory38:
@ssakash @mirh I need help for the option description. What do you think of 
```
	m_combo_vsync->SetToolTip( pxEt( L"Vsync eliminates screen tearing but typically has a big performance hit. Adaptive Vsync is a trade-off between screen tearing and performance. Vsync usually only applies to fullscreen mode, and may not work with all GS/renderer plugins."
```

mirh:
> The question was in case the dwm compositor is still enabled.

Of course. I'm never assuming it to be disabled. 

> Are we still locked at 60 fps with stuttering. Or can we go higher ? lower?

As I was saying this is basically what we have now

> For example in not-full screen mode. 

And given what's the majority of our good-willing testers.. Basically in every test not-of-mine you can 100% expect this not to matter. 

> CPUs are capable of millions of operations per seconds. A couple of instructions every vsync is peanut

Mhh.. ok, but still why not somewhere else?

Also, v-sync being enabled or not.. If you are _affected_ by compositor, it's not like you are not still in need of dwmflush. 

> Vsync usually only applies to fullscreen mode, and may not work with all GS/renderer plugins.

Mhh, what about: 
`V-sync is usually being already performed by OS compositor and shouldn't require additional explicit enforcement. `

Then, I guess we could even mention which use-cases trigger what.. But I'm not sure the lengthy explanation would be worth the scariness of the wall of text.

gregory38:
> V-sync is usually being already performed by OS compositor and shouldn't require additional explicit enforcement.

The string is intended for the user. And I'm pretty sure people won't understand it correctly. Because it depends on the OS/driver/...

gregory38:
Besides, we need to explain the adaptive vsync .... However I think we can remove the part about the "GS supported stuff". We have only a single GS plugin nowadays.

mirh:
> And I'm pretty sure people won't understand it correctly. Because it depends on the OS/driver/

Yees. 
So we either describe everything in detail in the tooltip, or we just keep vague and tell user: trust us, we passed through all it. 
From his side he should just have to ask himself: do I have the kind of mind/monitor combo that I don't care about tearing? Or not?
IMO. 

> we need to explain the adaptive vsync

Fine point, but first - I'm not even sure the downsides are real to begin with. 
Too bad there's nobody on this planet capable to test that /irony-not-irony

gregory38:
Ok. But in this case, I think it would be better to directly state that option is only available when the OS compositor doesn't do vsync. In you statement, some people might think the option override the OS/driver setting. What do you think ?

mirh:
Mhh
I think that you should look at it from user side, not dev "whatever it takes to do it" one. 
In other words: if ticking that option doesn't do anything, but user still is cool, that's the most important thing. 

*Then*... I guess like we should provide some concise explanation (to avoid unpleasant misunderstandings like "why v-sync is unticked but I still have it"?)

> Vertical synchronization is a [method/protocol/technique/something] that locks frames displaying to monitor pixels refresh rate in order to avoid tearing. This may lead to stuttering when framerate is subpar.
> Adaptive (half-tick) trades this absolute clamp to allow any range of performance. 

or 
> Adaptive trades this absolute clamp to allow any range of performance. and will be used if supported. 

Then here stuff gets tricky and I dunno if whatever I say should be considered GL only or not. 

> Most of times* though applications don't need to [do anything for/explicitly enable any of] this, since OS compositor will already handle it (at the cost of just a frame of latency). 
> The setting here tries to be smart enough to recognize such cases and will avoid double-applying wrecks. 
> 
> *SOME exceptions being [at least] amd and intel drivers in opengl fullscreen mode

gregory38:
The things is lot of people report me that
* I disabled vsync but fps is still 60 fps
* I enabled vsync but I can still see tearing (hopefully this one will be fixed)

Thinking about it again, I think people can google vsync and such. So let's do something easy. 

Most of times, window compositor already handle Vsync. If you suffer of screen tearing, you can still enable it here at performance/stuttering cost.

mirh:
> I disabled vsync but fps is still 60 fps

Unless we are talking about frame limiting, I think to have been the only one reporting that. 
And it was about dwmflush, if any, not something else. 

In this case, I think triple buffering would be a godsend. 

> I enabled vsync but I can still see tearing

Where ever, except [this](https://github.com/pcsx2/pcsx2/issues/1075)?

> Thinking about it again, I think people can google vsync and such.

Yes, I think they could - but I don't think they should. Tooltips are exactly for that ultimately. 

Then, I don't like giving this half-assed impression "we don't have any clue about problems, experiment". 
Not because I value reputation or something, but because actually we do know the great part of the situation. 
In the end as I was saying, user should just have to ask himself what side of the *personal trade-off* he is (that's what user settings are for after all, right?)

Ideally the perfect tooltip should manage to suggest this dilemma.  

> Vertical synchronization is a GPU ↔ monitor frames locking mechanism used to avoid tearing. 
> When not already performed by the OS (e.g. window compositor is disabled or graphics drivers is forcing a different fullscreen behavior) pcsx2 will use this setting. 
-Triple buffering note- ?
> [adaptive note]

Plus, figure out what variable refresh rate entails I guess.

gregory38:
Please let's keep 3 buffering outside of the topic. 3B creates others issues.

 > In the end as I was saying, user should just have to ask himself what side of the personal trade-off he is (that's what user settings are for after all, right?)

Yes. We still need something easy enough to understand. I'm afraid the translator might change the sense a bit. It is quite a complex topic. Anyway I like you latest proposal.

What do you think for adaptive.
Adaptive -> Adaptive Vsync is a relaxed Vsync to reduce, but not remove, tearing with a smaller speed penalty.

mirh:
> I'm afraid the translator might change the sense a bit. It is quite a complex topic

Good point. Guess like we could put a "dev-only" comment (explaining long and wide) aside of the string?

> Adaptive Vsync is a relaxed Vsync to reduce, but not remove, tearing with a smaller speed penalty

Yes and no. Since wording would all depend on the previous or not 3B note. 
Which, I see what you are saying, but [waiting for a fixed built to test the conjecture] I'm afraid it might be needed to avoid hard-locking on dwmflush.

gregory38:
If you don't want to wait on dwmflush, the only good solution is to remove the compositor. 3B will bring us various issue such as extra latency. People should invest in 120Hz/144Hz display.

mirh:
> If you don't want to wait on dwmflush, the only good solution is to remove the compositor. 3B will bring us various issue such as extra latency

Putting aside that I'm still not sure tbh (a [fixed](https://github.com/PCSX2/pcsx2/pull/2000#issuecomment-313376726) build would be welcome).. how would 3B affect latency?
I wasn't saying to do triple-buffered _v-sync_. Just the former thing to decouple frame rendering from presentation. 

> People should invest in 120Hz/144Hz display.

Yes, but we want perfection 🙃

willkuer:
Can't you just write:

1. Adaptive: Might reduce tearing, might introduce stuttering
2. Full: Might further reduce tearing, might further increase stuttering
3. No: Might reduce stuttering, might introduce tearing

And let the do the experimenting, reasoning, complaining, guessing and frustrating? The shorter the better.

mirh:
> And let the do the experimenting, reasoning, complaining, guessing and frustrating?

Who do you think we are? Microsoft?

willkuer:
So we are the guys that save the world by adding essays into tooltips?
Write an explanation in the official configuration guide if you want to.....

mirh:
> So we are the guys that save the world by adding essays into tooltips?

When did 4 lines become an essay?
Though, I just checked [what](https://github.com/PCSX2/pcsx2/blob/6589feaacdff52b0dfcca57bb2107c3fcf70d051/pcsx2/gui/Panels/GSWindowPanel.cpp#L58) we have now and I kind of see some point. 

So....... could we just postpone this thing after we have sorted the technical stuff?

gregory38:
Let's keep current string. It is way too late to change it now. I added the "Changes .. strings" label as a remainder to check that I didn't change any strings.

mirh:
Cool, back to technical stuff. 

Now there'd be those "I [cannot](https://github.com/PCSX2/pcsx2/pull/2000#issuecomment-313376726) get v-sync to work" and "dwmflush shouldn't depend on v-sync" thingies.

gregory38:
So the emergency fix  leads to an epic failure as always... I will fix it after work.

mirh:
Mhh.. somebody mistyped some variable 🙄

gregory38:
Yeah, yeah. Note that it compiles fine on my system :)

plugins/GSdx/GSWndWGL.cpp
```diff
@@ -323,16 +331,45 @@ void* GSWndWGL::GetProcAddress(const char* name, bool opt)
...
+	GSVector4i surface = GetClientRect();
+	GSVector4i window = GetWindowRect();
+	return !(surface == window).alltrue();
+}
+
 void GSWndWGL::Flip()
 {
+	// The compositor manages its own kind of Vsync/double-buffering. You need to wait the availabitiliy of the buffer
+	// before the swap to avoid stuttering.
+	if (CompositorEnabled() && vsync)
```
mirh: mhh no, like - I meant seriously. 
Putting aside I'm skeptic about it, until two days ago I remember that read m_vsync or something.

gregory38:
So on the string side, I added ":" on the label to be uniform with others combo boxes. There are also new strings (entries of combox)

mirh:
Little showerthoughts: 

* what happens to CompositorEnabled function assuming none of the "checks" get triggered?

* are you sure [GetWindowRect](https://msdn.microsoft.com/en-us/library/windows/desktop/ms633519.aspx) and eglQuerySurface are returning the same "kind" of RECT array?

* would it have been all that different to just put normal if/and/or/else instead of your "point-per-point" checklist? Like
````
if (!GLLoader::nvidia_buggy_driver && DwmIsCompositionEnabled(&dwm) == S_OK && surface != monitor){ 
	return true
} else return false

File: pcsx2/gui/GlobalCommands.cpp [hunk 2]
```diff
@@ -150,15 +150,20 @@ namespace Implementations
...
 	}
 
 	void Framelimiter_MasterToggle()
 	{
 		ScopedCoreThreadPause pauser;
 		g_Conf->EmuOptions.GS.FrameLimitEnable = !g_Conf->EmuOptions.GS.FrameLimitEnable;
-		GSsetVsync( g_Conf->EmuOptions.GS.FrameLimitEnable && g_Conf->EmuOptions.GS.VsyncEnable );
 		OSDlog( Color_StrongRed, true, "(FrameLimiter) %s.", g_Conf->EmuOptions.GS.FrameLimitEnable ? "ENABLED" : "DISABLED" );
+
+		GSsetVsync(EmuConfig.GS.GetVsync());
```
turtleli: This breaks the ability to disable the frame limiter when Vsync is enabled. `g_Conf->EmuOptions.GS.FrameLimitEnable` is toggled, but `EmuConfig.GS.FrameLimitEnable` is checked instead in GetVsync().

mirh: Especially with the coming of >60Hz screen I can see very legitimate situations where `v-sync ON` and `frame limiter off` could happily coexist, we shouldn't assume that. 

Also wtf - I always used frame limiting off "effects" to test whether v-sync was working or not in pcsx2.

gregory38: Hum, I mostly shuffle the code to make it easier. It wasn't the goal to break it. I will check the logic. Maybe we could just separate framelimiter and vsync.

MrCK1: Separation of the framelimiter and vsync sounds like a good plan ;)

MrCK1: Separation of the framelimiter and vsync sounds like a good plan ;)

gregory38: Anyone with a 120Hz/144Hz display that know how to code ?

It is possible to query the monitor mode with wx. `wxVideoMode wxDisplay::GetCurrentMode` or `wxArrayVideoModes wxDisplay::GetModes`.  Then `mode.GetRefresh()` could give us the refresh rate. It seems to be implemented only on windows. But I think it would make sense to not disable Vsync when turbo is enabled on high speed monitor.

mirh: Conditionally decouple framelimiter from vsync only for certain refresh rates sounds really bad. 
(also, with variable refresh rates monitors, situations might \*_really_\* become awkward)

Btw, I think "Buffer Object Streaming" is the better naming/technique for the buffering thing I had in my mind yesterday. 
Which certainly wouldn't lead us back to square one if what I'm supposing in #1887 is right.

gregory38: Actually linux is supported but it crashes (due to glx thread unsafety...)

gregory38: @mirh I wrote turbo not frame limiter.

Currently when the turbo is enabled, vsync is force disabled. Otherwise the turbo will be limited to refresh rate, which is 60 fps most of the time (so not a turbo). However, on high speed monitor it is doable. So you will be able to run turbo in high quality (i.e. no tearing). I don't know what will happen if the user can't manage the turbo fps target (stuttering maybe if not adaptive vsync).

On variable refresh rate monitor, the vsync parameter must remain off (or always on I don't know) so current behavior is broken.

Then on the framelimiter topic. I think all combinations of framelimiter/vsync are possibles. So it doesn't make sense to interlock them.

mirh: Oh, I see. 
Anyway I don't think that's something we should assume on our side either.

> On variable refresh rate monitor, the vsync parameter must remain off

I think this. @masterotaku

frantisekz:
Do I understand Adaptive V-sync correctly as FreeSync/G-Sync?

I can see in my display OSD if the video is using FreeSync or not and in case of PCSX2, it doesn´t indicate active FreeSync.

mirh:
That's variable refresh rate. 
Adaptive v-sync can be seen as a kind of dynamic v-sync (aka late vsync, aka swap tearing, aka smart vsync)

AFAIK applications shouldn't require anything special to enable FreeSync/G-Sync (except exclusive fullscreen I  guess)

---
In other news, I just [checked](http://www.anandtech.com/show/10325/the-nvidia-geforce-gtx-1080-and-1070-founders-edition-review/13) nvidia's fast sync is as expected whatever I was referring to under the name of triple buffering - and when benchmarked increase in latency has been minor (with [no](https://forums.blurbusters.com/viewtopic.php?f=10&t=2857&start=40) reports of stuttering)

gregory38:
@frantisekz 
Adaptive Vsync is basically an accurate frame limiter. The standard frame limiter isn't constant, instead it is 60 fps +/- big delta. Adaptive Vsync gives you exactly 60 fps for the maximum speed.

@ramapcsx2 
By the way, this PR may interest you, the adaptive vsync part. I think you've tried to implement something like adaptive vsync with the frame limiter.

Nicholas-Steel:
Adaptive Vsync:
- Traditional Vsync enabled when frame rate meets or exceeds monitor refresh rate
- Traditional Vsync disabled when frame rate falls below the monitor refresh rate (and you get screen tearing).

G-Sync/Freesync is completely different technology to Adaptive Vsync.

plugins/GSdx/GSWndWGL.cpp
```diff
@@ -323,16 +331,45 @@ void* GSWndWGL::GetProcAddress(const char* name, bool opt)
...
+	// Note: DWM is always true on Win8 and later
+	if (!dwm)
+		return false;
+
+	// Easy case are done. It remains driver black magic
+
+	// Recent driver (+380.xx ??) seem to always get v-sync enforced someway.
+	// XXX it would be nice to understand what happen to fix it properly
+	if (GLLoader::nvidia_buggy_driver)
+		return true;
```
turtleli: Hmm. I think this means that the standard and adaptive options for Windows 8+ NVIDIA users is always the same?

mirh: It means the mindblowing mystery that I tried to figure out in #1437
First I thought it was about exclusive mode not triggering. Than it seemed to be about v-sync always being enforced. 
Now.. Hell I'd knew. 

I totally trust you'll be able to determine whenever it should be the case we applied DWMFlush (and set swap_interval to 0) and when not.

gregory38: Yeah. Even without this code, any Vsync won't work if exclusive fullscreen isn't triggered properly. Plan is to fix frame limiter option regression and to merge all but this commit.

pcsx2/gui/GlobalCommands.cpp
gregory38: @turtleli 
Do you know the difference between `g_Conf->EmuOptions` and `EmuConfig` ?

turtleli: Not sure I do.

My interpretation: `EmuConfig` is used (or should only be used) by the core emulation code and write access to it is supposed to be restricted to prevent accidental modifications (I note that it being const is likely undefined behaviour). `g_Conf->EmuOptions` is for changes that are supposed to be applied to `EmuConfig`, though it's not always obvious if and when the changes are applied.

I think you have to use `g_Conf->EmuOptions` in this case.

gregory38: Hum, interesting, behavior wasn't the same between GSopen and GSopen2.

Still something annoye me, hitting turbo/slowmo will enable the framelimiter. But it will never be disabled if you left the turbo/slowmo mode. I understand that turbo/slowmo don't make sense without a frame limiter. But I'm not sure toggling framelimiter is a good idea.

So is there any case that we want to disable Vsync. We have the turbo mode (but it is likely useless for high fps display). Do we want to disable Vsync when the user hit F4 to disable the frame limiter ? My feeling is no. And then there is the magical combination.
* no frame limiter + vsync
* hit turbo -> enable frame limiter / disable vsync
* hit f4 -> no frame limiter / vsync still off / turbo still on

Nicholas-Steel:
Note that the DWM Compositor can be disabled under Windows 8.0 & 8.1 however it is a slightly involved process and the same procedure won't work under Windows 10 (will cause a system deadlock under Windows 10).

See here for steps on turning the DWM Compositor off for Windows 8 and 8.1: https://pastebin.com/X3XgVAva

pcsx2/gui/GlobalCommands.cpp
turtleli: > So is there any case that we want to disable Vsync. We have the turbo mode (but it is likely useless for high fps display). Do we want to disable Vsync when the user hit F4 to disable the frame limiter ? My feeling is no.

I think vsync should be disabled if the user hits F4. It'd be more convenient if a user only has to press F4 to disable/enable the framelimiter instead of having to also disable/enable vsync as well (I'm assuming vsync and the framelimiter are coupled together, haven't thought much about what the logic would be if they're decoupled).

mirh:
Mhh.. guess like kudos for the "one thing at time" thought. 

And unbelievably, I do am getting sync now. No BS. 
It's a bit of a pity you have to close and re-open the rendering window for the setting to apply though?
Also, are "safe presets" expected to grey out v-sync?

lightningterror:
> It's a bit of a pity you have to close and re-open the rendering window for the setting to apply though?

It probably requires the gs plugin to be restarted. Changing core gs /emulation settings doesn't do that if I'm correct.

> Also, are "safe presets" expected to gray out v-sync?

It shouldn't. Any kind of preset actually grays it out. It's the same on master.

gregory38:
```
It probably requires the gs plugin to be restarted. Changing core gs /emulation settings doesn't do that if I'm correct.
```
It is the issue detected by turtleli. OpenGL is per-context, so you can't update vsync from the main thread. Once fixed, it will be easier to test Vsync.

```
It shouldn't. Any kind of preset actually grays it out. It's the same on master.
```
Sorry. What do you mean ? It should grays it out or not ?

lightningterror:
image

This , I don't see why it should gray it out , since it's vsync and not really a hack or anything.

gregory38:
I think it is as master. We could discuss about it. Search on github there is a related (maybe closed) issue on it.

lightningterror:
Found it #1642
Anyway it's not really an issue , more like for safety and doesn't really matter. Forget what I said :nerd_face: 

Seems that travis build didn't compile properly since git had issues earlier today.

turtleli:
Actually the "close/resume emulation before vsync" is a different issue from what I detected. It was also reported in #1971 - "Vsync setting on linux has no effect until application restart" - which if you think about it is probably the same thing. I don't think we detect if the vsync setting has been changed while emulation is active, so if someone only changes the vsync setting but doesn't touch the hotkeys (let's pretend it's fixed or they're using d3dwhatever) then there'll be no change to the vsync setting until emulation is paused/resumed. So that's one more thing that probably needs fixed at some point.

gregory38:
You're right. Actually I got this behavior in my quick test of this PR (with a dirty hotkey hack fixes). But does it work on Windows ? Vsync setting is only set during opening and hotkey. Maybe we should just call GSsetVsync after any setting change.

turtleli:
> But does it work on Windows ?

No, I don't think it does, I was just quoting from the linked issue.

gregory38:
Ok.

So how do we proceed ? Merge this PR and let's you do the fix of the threading sync issue ? Or the reverse way ?

mirh:
> This , I don't see why it should gray it out , since it's vsync and not really a hack or anything.It's the same on the other presets as well.

Thinking better, v-sync might lead to some kind of problem depending on whatever weird monitor refresh rate you may have, so I guess it may have sense to gray it out. 

So, there's just the "window has to be restarted" problem imo. 
Which.. I dunno, if it isn't all that complex could be merged here, given the PR is pretty lightweight at this point.

turtleli:
> So how do we proceed ? Merge this PR and let's you do the fix of the threading sync issue ? Or the reverse way ?

Dunno. I think either way is fine. I'm not sure the compositor stubs should be in the commits since the DWM stuff has been removed for now.

gregory38:
Ok. I merged the first part. I let you handle the set vsyc part.

mirh:
Cool. 
Now, aside of that, 2 other things to tackle:
* can we get any objective "preference"/*precedence* between adaptive and standard vsync?
* how to neatly integrate dwmflush?
* adjust v-sync label

Interestingly here, even on a 60Hz screen, PAL framerate has no problems to be maintained with v-sync.

gregory38:
@mirh 
It depends on the machine. And it is a tradeoff between frame latency / tearing / stuttering / frame drop / frame pacing.
* Adaptive + no frame limiter will give you better compromise (except maybe frame latency)
* If you can't stand tearing => full vsync but you will get frame drop

mirh:
> Well that's lame, disablng vsync with the frame limiter.

Is it?

> It depends on the machine. 

Yes, and that's why any assumption with arbitrary values (like the alleged one above) should not ship. 
Especially with no mention to the end-user. It just makes for extreme confusion.

DonelBueno:
I've tested the latest build and enabling vsync (full or adaptative) still caps the framerate to my monitor refresh rate even when I hit F4 during a game.

gregory38:
@DonelBueno yeah. It isn't the purpose of this PR. It will be fixed soon.

DonelBueno:
Oh, got it. Is it part of the "set vsyc part" you mentioned?

gregory38:
yep

mirh:
I'm starting to go crazy. 
I thought the set vsync part was about not having to close and reopen the window for vsync setting to apply?

PR #2003, opened 2017-07-05
PCSX2-WX: Add restore defaults button to GS panel

ssakash:
**Summary of changes**:

* Added ``Restore Defaults`` button to the GS panel of Emulation settings dialog.

Initially, I was planning on adding it for all the dialogs but it seems pointless to add it to Gamefixes panel as you can just uncheck the ``Enable Manual Gamefixes`` and be done with it and the GS Window panel seems to be under a bit of a work right now ( #1950 & #1918), wouldn't want to cause any conflicts with them, so I'm doing it only for the GS panel at the moment.

### DIFF
--- a/pcsx2/gui/Panels/ConfigurationPanels.h
+++ b/pcsx2/gui/Panels/ConfigurationPanels.h
@@ -303,6 +303,7 @@ namespace Panels
 	{
 	protected:
 		pxCheckBox*			m_check_SynchronousGS;
+		wxButton*			m_restore_defaults;
 		pxCheckBox*			m_check_DisableOutput;
 		FrameSkipPanel*		m_span;
 		FramelimiterPanel*	m_fpan;
@@ -312,6 +313,7 @@ namespace Panels
 		virtual ~VideoPanel() = default;
 		void Apply();
 		void AppStatusEvent_OnSettingsApplied();
+		void Defaults_Click(wxCommandEvent& evt);
 		void ApplyConfigToGui( AppConfig& configToApply, int flags=0 );
 
 	protected:
--- a/pcsx2/gui/Panels/VideoPanel.cpp
+++ b/pcsx2/gui/Panels/VideoPanel.cpp
@@ -291,6 +291,8 @@ Panels::VideoPanel::VideoPanel( wxWindow* parent ) :
 		_t("For troubleshooting potential bugs in the MTGS only, as it is potentially very slow.")
 	);
 
+	m_restore_defaults = new wxButton(right, wxID_DEFAULT, _("Restore Defaults"));
+
 	m_check_DisableOutput = new pxCheckBox( left, _("Disable all GS output"),
 		_t("Completely disables all GS plugin activity; ideal for benchmarking EEcore components.")
 	);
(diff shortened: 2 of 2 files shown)


gregory38:
I love screenshot when you update the GUI. Where did you put the button ?

ssakash:
Sorry, you know how much I love to post screenshots. ;)

picture

I've placed it in the bottom right of the right most panel. (similar to the placement of the other panels for consistency).

lightningterror:
Looks good, behaves properly as well.
Is it also added on linux btw?

ssakash:
> Is it also added on linux btw?

Yes, wxWidgets code is common for all platforms.

gregory38:
It would be nice to have a single button in the bottom bar (on the same level as ok/cancel/apply). Instead to have a per panel button. Besides it will save room for small screen. I don't know if it is difficult to code.

ssakash:
> It would be nice to have a single button in the bottom bar (on the same level as ok/cancel/apply).

I don't see how it's better, it looks sort of awkward to me to have it at the same level as ``OK/Cancel/Apply``. The bottom is already crowded with those buttons plus the preset slider and screenshot icon.

I just placed the ``Restore Defaults`` similar to it's position in other tabs. (``EE/IOP``, ``VUs``, ``Speedhacks``)

gregory38:
> I just placed the Restore Defaults similar to it's position in other tabs. (EE/IOP, VUs, Speedhacks)

Yes I know that it is the same issue for others tabs. It wasn't directly related to this PR.

It is strange to have a default button at various location which depends on the current tab. Here it is fine but it takes too much room on the speedhack panel (people still have low height display). Besides, maybe it would be better to have a single restore default button that restore all default instead of the current tab option (except if the use case is to only "default" a single tab).

> it looks sort of awkward to me to have it at the same level as OK/Cancel/Apply.

Why ? I mean all 4 buttons control the settings. You can validate the settings, you can discard the settings. It feels logical to be able to reset the settings too. The fact that you need to duplicate it in all tabs is a proof that it doesn't belong to a tab.

ssakash:
> Besides, maybe it would be better to have a single restore default button that restore all default instead of the current tab option

> **(except if the use case is to only "default" a single tab)**.

Yeah, thanks for repudiating your own suggestions by providing a counter-argument along with it. :P 

Let's assume a certain scenario with your global restore defaults idea,

* **Person A** has his manually tuned speedhacks necessary for his game but now he encounters a new rounding/clamping bug, he tries playing with the rounding/clamping modes and now wants to switch back to the defaults. **What would he do?** He can't possibly use the global restore defaults option as it would also ruin his personal combination of speedhacks.


> Why ?  I mean all 4 buttons control the settings. You can validate the settings, you can discard the settings. It feels logical to be able to reset the settings too.

Probably because ``OK/Cancel/Apply`` applies to all the tabs and not just for the current tab like ``Restore Defaults``? It doesn't make much sense to have them at the same level.

Also it's not used enough to place it near the global ``OK/Cancel/Apply`` buttons which are frequently used, that's my reasoning on why I'd prefer to not have together but your mileage may vary, it's a rather subjective topic.

>  The fact that you need to duplicate it in all tabs is a proof that it doesn't belong to a tab.

Well, there's no other way, It needs to be duplicated in every single tab if we need a button which restores only the current tab's options to defaults.

gregory38:
Not really a counter argument. Even if you put the button in the bar, you can still code it to reset only the current tab.

The fact is that this button consume too much space in the speedhack panel. So far the solutions are
* keeps the button in the middle and render crap in small screen
* puts the button in a different location

Technically it would be the same story for the gamefix panel... But I guess it won't have it.

Anyway, I'm not against your PR.

PR #2005, opened 2017-07-09
PCSX2-WX: Revert titlebar update behavior on Windows

### LINKED ISSUE #2002: Emulation window title doesn't get updated when in fullscreen
The one with FPS, gs utilization and all. 
They always keep the same until you get back to windowed.

ssakash:
**Current behavior** - The tilebar isn't updated when the user enters full screen mode and when the user returns back to windowed mode they have the older title bar values for a brief second, this sort of behavior is undesirable just in the cost of saving some overhead for updating the title. (which is really negligent)

Hence reverting the code to always update the title like it has been for the past 7 years (Yes, I did my research), I'm doing it only for the windows side at the moment as a code comment describes of some sort of Linux specific issue on wxWidgets side

Closes #2002 (Only on Windows side, the title was always static on fullscreen for Linux anyway)

### DIFF
--- a/pcsx2/gui/FrameForGS.cpp
+++ b/pcsx2/gui/FrameForGS.cpp
@@ -618,9 +618,11 @@ void GSFrame::OnUpdateTitle( wxTimerEvent& evt )
 	out << std::fixed << std::setprecision(2) << fps;
 	OSDmonitor(Color_StrongGreen, "FPS:", out.str());
 
+#ifdef __linux__
 	// Important Linux note: When the title is set in fullscreen the window is redrawn. Unfortunately
 	// an intermediate white screen appears too which leads to a very annoying flickering.
 	if (IsFullScreen()) return;
+#endif
 
 	AppConfig::UiTemplateOptions& templates = g_Conf->Templates;
 


gregory38:
I think the issue was on ZZogl. I don't think GSdx was already ported to Linux.

Anyway,``` back to windowed mode they have the older title bar values for a brief second, ```
Is it really a that big issue ?

ssakash:
> Anyway,back to windowed mode they have the older title bar values for a brief second,
Is it really a that big issue ?

I wouldn't say it's a *big* issue but it might be a very *small* one.

I just don't think it's worth skipping the ``SetTitle()`` call in full screen when it introduces issues like this one and when the function is not even remotely demanding. (Unless I'm wrong, this seems to be the consensus regarding this matter at #2002)

gregory38:
I agree with you. For completeness, a clever driver can do black magic when the title of the window is updated. Look at Linux, which goes wrong when you update the title of the no-title bar window...

I will relaunch appveyor tonight and I will merge this pr.

gregory38:
Well I don't know what is going-on with appveyor. Let's merge it.

PR #1977, opened 2017-06-23
GameDB: Add fixes to various games

lightningterror:
- Add fix for R&C2 Going Commando - Fixes #964
fixes cinematic and crash  
- Add fix for Metal Gear Solid 3: Subsistence  
 fixes screen artifacts
- Add multiple fixes for Max Payne  
- Add / correct fix for Mercenaries 2  
Corrected incorrect title and added timing fix

Add ScarfaceIbitHack in Gameindex since now it will be used

Thanks to @atomic83GitHub for testing out the games.
Close #1976

Code tidying up

PR #2014, opened 2017-07-17
PCSX2-Counters: Fix tracking of scalar limit

ssakash:
**Summary of changes**

* Properly track the scalar limit value. Fixes the second issue mentioned over here (http://forums.pcsx2.net/Thread-Change-the-normal-speed?pid=566120#pid566120).

**Procedure to reproduce the issue**:
* Toggle to turbo mode
* Enter emulation settings dialog and change the ``Turbo Adjust`` to a different percentage value from the current one.
* Apply the settings

_**Master branch behaviour**_
> The framelimit won't be updated, it will only be updated after you toggle back to normal mode and once again come back to turbo mode.

_**Pull request behaviour**_
> It's updated instantaneously after you press either the ``OK/Apply`` button in the emulation settings dialog.

### DIFF
--- a/pcsx2/gui/GlobalCommands.cpp
+++ b/pcsx2/gui/GlobalCommands.cpp
@@ -86,7 +86,6 @@ namespace Implementations
 		{
 			g_Conf->EmuOptions.GS.FrameLimitEnable = true;
 			g_LimiterMode = Limit_Turbo;
-			g_Conf->EmuOptions.GS.LimitScalar = g_Conf->Framerate.TurboScalar;
 			OSDlog( Color_StrongRed, true, "(FrameLimiter) Turbo + FrameLimit ENABLED." );
 			g_Conf->EmuOptions.GS.FrameSkipEnable = !!g_Conf->Framerate.SkipOnTurbo;
 		}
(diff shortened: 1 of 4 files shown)


ssakash:
@turtleli 

Do you know what's wrong with Appveyor? The same issue also happened in https://github.com/PCSX2/pcsx2/pull/2005#issuecomment-314881845.

turtleli:
No idea. Try rebasing to the latest master?

ssakash:
> No idea. Try rebasing to the latest master?

Works now but I'm not sure about the issue. I usually have my branch trailing by 4-10 commits from master whenever I make a pull request but I've never noticed this issue before, seems to only occur recently.

lightningterror:
It happened on mine as well.
https://github.com/PCSX2/pcsx2/pull/1977

pcsx2/gui/GlobalCommands.cpp
```diff
@@ -86,15 +86,15 @@ namespace Implementations
 		{
 			g_Conf->EmuOptions.GS.FrameLimitEnable = true;
 			g_LimiterMode = Limit_Turbo;
-			g_Conf->EmuOptions.GS.LimitScalar = g_Conf->Framerate.TurboScalar;
+			gsUpdateFrequency(g_Conf->EmuOptions);
```
gregory38: It will conflict with my VSYNC PR. But would it be possible to move this new line outside of all the if/else branches.

ssakash: > But would it be possible to move this new line outside of all the if/else branches.

Nice catch 👍

Issue #1976, opened 2017-06-23
GameDB improvement and missing patch files


atomic83GitHub:
![image](https://user-images.githubusercontent.com/26351275/27491972-bac6987e-5844-11e7-8b7f-0d2d705d8b2b.png)

MrCK1:
Is there a valid reason for these improvements? (Ie. Are they documented in the forums, ect.)

lightningterror:
Apparently there is for Max Payne.
http://forums.pcsx2.net/Thread-Max-Payne-SLES-50325-E--29333

As for R&C is he talking about Going Commando?

Anyway here is the modified GameDB. Anyone care to test this out ? Don't wanna make a pr yet.
https://drive.google.com/open?id=0B8cQAq-Ko9-ueWpZOFp0TEVIWWM

atomic83GitHub:
Yes its Ratchet and Clank Going Commando (Ratchet and Clank 2 in my country).

atomic83GitHub:
Your GameDB work and solved these games but missing patch are not also added.

lightningterror:
@MrCK1 I didn't add the VU I bit hack since i don't see it in the dbf file mentioning that it can be added.
Can it still be added with just adding "ScarfaceIbit = 1" ? Or other changes need to be done first before it can be used here?

atomic83GitHub:
Oups , i have forget to added Metal gear solid 3 (all versions): VUs :Clamping mode , extra + preserv sign

lightningterror:
Metal Gear Solid 3 Subsistence or Metal Gear Solid 3 Snake Eater ?

http://forums.pcsx2.net/Thread-Metal-Gear-Solid-3-Snake-Eater-SLES-82013-E--30514
It seems for Snake Eater the values shouldn't be changed since it was tested.

Also are you using the latest dev build ?

atomic83GitHub:
I have used the 1.4.0 and the latest build but in this game , its strange because , if Clamping mode is set to Normal or never , it will cause many artefacts on screen in both versions.

ssakash:
> Can it still be added with just adding "ScarfaceIbit = 1" ? Or other changes need to be done first before it can be used here?

It's actually ``ScarfaceIbitHack`` and yes, it'll work without doing any other internal change in the codebase.

atomic83GitHub:
For MGS3 Screen artefacts is reduce on latest build but still here.

lightningterror:
[...]

Issue #1894, opened 2017-04-09
Automatic game hacks


mirh:
Yes, I know, people wisdom and will is crap. 
And yes, I realize in some cases we are in situations comparable to a triangle, with speed, accuracy and "neatness" on the opposite vertices.
But it's also undeniable that in others, enabling them (for as much "dirty as mud") it's a no brainer. And really, it starts to feel like a joke for people to wander forums and do the same [guesswork](https://www.reddit.com/r/emulation/comments/56wj30/mipmapping_makes_the_ace_combat_games_run_way/d8osrry/) over and over again separately. 

And looking at how nicely #1874 has been executed, I was dreaming a similar "automatic preset" for hacks too (handled in [gameindex](https://github.com/PCSX2/pcsx2/issues/1893) together with gamefixes perhaps?)
Or at least some way, link, note, whatever for ANY average joe to at least take notice he can do something. 
It's depressing to see how many people just crank up resolution, and then complains even it the issue had been worked around ages before. 

EDIT: supposedly Dolphin calls this "Game INIs"

gregory38:
Gameindex is already a huge file to handle in GIT. If we really put a database, it would be better to store it in a more efficient format. But eventually a solution to handle per game setting would be nice.

mirh:
> Gameindex is already a huge file to handle in GIT.

Indeed. But either we drop the "editable/readble by people" criteria, or otherwise I don't see how in the world a 5000 entries X 10 lines each X let's say 20 characters could fit in less than the actual ~1MB. 

Reorganizing it as I mentioned in #1893 could quite potentially fix some redundancy (who knows, perhaps in the same order of magnitude of the additions for hacks we'd have to make) then, but with time I'm sure it will still grow. 
With these assumptions, and if the problem is *just* with git being a bit of a PITA for large files.. Perhaps LFS and/or a separate repo just for the database might be the answer?

gregory38:
Yeah, I know. But the fact is that last time I listed the big entry in `.git` (I don't remember how). This file got easily 15 entries on the top 20. I'm not sure that I want to handle a GB history for it.

mirh:
Oh, in that sense, yeah I see it being a massive problem. 
But.. Shouldn't git be supposed to just store differences [at least for plain text files]?

gregory38:
Top 100 in reverse (bigger file at the bottom)
```
1219134 bin/GameIndex.dbf
1219293 bin/GameIndex.dbf
[...]

PR #2015, opened 2017-07-18
This should make replay loading work in Virtua fighter 4 after enabling fast cdvd speedhack

prafullpcsx2:


ramapcsx2:
@prafullpcsx2 it will break Tales of the Abyss. 
You cannot easily make this hack compatible with all games. In fact, the current version is making it better for most of the otherwise incompatible games.
If you need a list of them, check out an old HDLoader compatibility list and sort by those that require a certain DMA mode.

willkuer:
I guess adding an automatic/manual gamefix 'Force Fast CDVD Read' or something similar would be reasonable then. I think you are also not able to add speedhacks to the gamedbf, or?

mirh:
Currently supported entries are linked [here](https://github.com/PCSX2/pcsx2/issues/1976#issuecomment-310780221). 
Kind of relates to #1894

lightningterror:
Adding patches comments in the gamedb is a somewhat okay alternative.

I wouldn't want this to be merged if it breaks other games.

woj1993:
What about having few methods of fast cdvd and list witch allows choosing metod or entry in gamedb that chose right metod for game?

lightningterror:
Maybe something like this?

```
	if (EmuConfig.Speedhacks.fastCDVD) {
		if(gamecrc || (eCycle < Cdvd_FullSeek_Cycles))
			eCycle = 3000;
	}
```

MrCK1:
That wouldn't work. The '''||''' operator means that both statements on each side have to be true. Even if the crc was specific, it doesn't make sense here.

lightningterror:
Hmm I thought it was an or statement , well then using else statement would work no?

mirh:
>  The || operator means that both statements on each side have to be true

I thought that was what && operator was about.

refractionpcsx2:
|| is logical OR, so either statement has to be true for it to pass. && would be logical AND, meaning both statements would have to be true

atomic83GitHub:
So, this PR solve Virtua fighter 4, Roland Garros 2003 but break Tales of the Abyss?

lightningterror:
I think it breaks other games as well.

prafullpcsx2:
@lightningterror : Virtua Fighter 4 fix comes into action only when fast cdvd speedhack is enabled, so how can it break other games in normal mode? Yes, the fix may make some games (like tales of the abyss) incompatible with fast cdvd speedhack but that's about it.

FlatOutPS2:
You could add a second layer of fast CDVD support, the original functionality and a forced option.

lightningterror:
@prafullpcsx2 are you gonna continue working on this or can I close it for now and reopen it when you decide to resume ?

prafullpcsx2:
I am reopening this pull request. I tested Tales of the Abyss today and it seems to work fine with this even after enabling cdvd speedhack. @ramapcsx2 Can you please explain which part of the game breaks with these changes? 

@atomic83GitHub reported that this also reduces loading times of Oni and helps in beyond good and evil.

Regarding changes related to Roland Garros 2003 @refractionpcsx2 earlier stated that most probably these changes are correct. I haven't faced any regressions so far with both these changes. So may be it's time to commit them?

ramapcsx2:
Abyss breaks semi-randomly in certain cut sequences and when roaming the overworld and fighting monsters. There is a cut scene with 2 people in front of one big tree early in the game, it breaks reliably on that with the hack enabled.
I forgot what other games broke, but there were a few. Check the HDLoader compatibility list for entries that require a different HDD mode. Those games are most likely to be affected.

prafullpcsx2:
I played Abyss for about 20 minutes today but didn't encounter any problems. Will probably play more tomorrow to find out the problem areas if any.

GovanifY:
I will close this PR for now, as master seems to have checks that make more sense to me and the PR seemed to have issues.

```
<<<<<<< HEAD
	if( cpuRegs.PERF.n.pccr.val & ((1 << (cpuRegs.CP0.n.Status.b.KSU + 2)) | (cpuRegs.CP0.n.Status.b.EXL << 1)))
=======
	if( cpuRegs.PERF.n.pccr.b.U0 )
>>>>>>> be07348b5 (Trying to fix Roland Garros 2003)
```

with ` 	if( cpuRegs.PERF.n.pccr.val & 0xf )` being used before the PR change for reference

PR #2018, opened 2017-07-21
Update GSCrc.cpp

puzzo888:


MrCK1:
A better title would be "Add missing Front Mission 5 CRC to GsCRC"

gregory38:
THanks

PR #2029, opened 2017-07-27
Translations update: tr_TR and cs_CZ

lightningterror:
Credits to [PyramidHead.](http://forums.pcsx2.net/Thread-Translation-files-for-Turkish-tr-TR-Updated-07-22-2017?pid=566545#pid566545)

On a side note why are .mo files inside "pcsx2\locales\tr_TR" shouldn't they be present only in "pcsx2\bin\Langs\tr_TR" ?


MrCK1:
*/locales is used by the PCSX2 app itself, but */langs is used by NSIS when compiling the installer.

lightningterror:
That is understandable but what I'm doing is comparing other translations
pcsx2\locales have only .po files
pcsx2\bin\Langs have only .mo files.

pgert:
The ".mo"-files are not to be included in */locales (they have no effect there).
Have the "locales/tr_TR/pcsx2_Iconized.mo" & "locales/tr_TR/pcsx2_Main.mo" files deleted.

gregory38:
It isn't mandatory to provide mo file. I will automatically build them on Linux and I will push them. Otherwise yes
```
pcsx2\locales have only .po files
pcsx2\bin\Langs have only .mo files.
```

Issue #2027, opened 2017-07-26
Tomb Raider Underworld NTSC-U & PAL (SLUS-21858) (SLES-55442) - Graphic glitches and very slow


LeCharlatan:
Hi (again),

First, here are basics informations :

**PCSX2 :**

**Build :** Dev PCSX 1.5.0~git201707251955 (PPA PCSX2 Team Daily Builds)

**PCSX2 options :**

**GSDX :** OGL Backend
**Speed hacks :** OFF

**GSDX Options :**

**Texture Filtering :** Bilinear (PS2)
**Large Framebuffer :** ON (Checked)
**Internal Resolution :** 2x Native
**Anisotropic Filtering :** OFF
**Blending Unit Accuracy :** High (Recommended high-end PC)
**Automatic CRC level :** Partial (OpenGL Recommended)
**Mipmapping (Insert) :** Basic (Fast)
**Mipmapping :** ON (Checked)

**Description of the issue:**

**NTSC-U Version**

TR - Underworld is showing a bad glitches that makes the NTSC-U unplayable using OGL HW
![screenshot_20170726_215159](https://user-images.githubusercontent.com/22820335/28642712-a2146634-7253-11e7-9271-fde04daf7501.png)

It seems that OGL SW is working fine tho
![screenshot_20170726_215010](https://user-images.githubusercontent.com/22820335/28642752-c0b3b11c-7253-11e7-98e6-3b3705f6de96.png)

**PAL Version**

Unlike the NTSC-U version, PAL looks fine using OGL HW (maybe it's missing some glow effect) but, it misses the blue blinking square on choice menu on OGL HW

Missing blue blinking cursor (OGL HW)
![screenshot_20170726_223604](https://user-images.githubusercontent.com/22820335/28642894-39a9ae96-7254-11e7-9908-5b2b4b5afaa0.png)

Here it is ! (OGL SW)
![screenshot_20170726_223636](https://user-images.githubusercontent.com/22820335/28642967-87c8f956-7254-11e7-8dd1-4977b386e983.png)

**Note :** this blinking blue cursors does not appear on PAL version using OGL HW, however, the ntsc-u shows it, even on OGL HW

A last point about both versions, they're using 100% of one CPU's core and the game is extremely slow on some points. This only happens in-game, not on menus
![screenshot_20170726_220031](https://user-images.githubusercontent.com/22820335/28644104-77e489a2-7258-11e7-9943-3e57907b2b5f.png)

[...]

PR #2031, opened 2017-07-27
GSdx: Add missing CRC for Tomb Raider Underworld

lightningterror:
NTSC-U version.

Issue #2027

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -357,6 +357,7 @@ CRC::Game CRC::m_games[] =
 	{0xF21EE6E0, CrashNburn, US, 0},
 	{0x694A998E, TombRaiderUnderworld, JP, 0}, // cutie comment
 	{0x8E214549, TombRaiderUnderworld, EU, 0},
+	{0x618769D6, TombRaiderUnderworld, US, 0},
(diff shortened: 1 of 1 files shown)


mirh:
Would it be possible to squash the two commits? 🙃

lightningterror:
I never was able to do on those type of commits , that update the forked branch , dunno why , maybe I'm missing something.

ssakash:
> I never was able to do on those type of commits 

I've cherry picked the valid commit to master branch, I didn't do squash-merge as I was unsure if it would properly squash the remote tracking commit.

gregory38:
it is the merge commit, isn't it ? A rebase should fix it. A squash merge should work too. Because it does a squash + rebase.

lightningterror:
That's weird , it doesn't work for me , It doesn't even appear when squash / rebasing , all others do.
Maybe it's because I'm using the GitHub desktop that comes with it's git shell and not the other Git application that provides more options.

ssakash:
> it is the merge commit, isn't it ? A rebase should fix it. A squash merge should work too. Because it does a squash + rebase.

Good to know. Thanks for the information. :)

mirh:
> not the other Git application that provides more options

That's quite cooler.

Issue #1669, opened 2016-11-20
Meta: MTVU compatibility issues


refractionpcsx2:
As with previous meta's. This doesn't mean we are working on this currently or that it will ever get fixed, this is more for information.  I just saw gregory's post on the forum about it so figured i'd make a start :P

Note: This is ONLY for broken games, right now, I consider slowdown and glitches a "tough" kind of situation. If you want a comprehensive list of those games, please visit the relevant forum thread here: http://forums.pcsx2.net/Thread-LIST-Games-that-may-not-work-well-with-MTVU

Fixed (As of Dec 2020):
- [x] 007: Agent Under Fire - (Internal slowdowns. Reported by Nobbs66) - Fixed in #4035 
- [x] 10,000 Bullets - (Black screen with MTVU. Reported by jackyyn) - Fixed in #4035 
- [x] Backyard Basketball - (Black screen with MTVU. Reported by junknaut) - Fixed in #4035 
- [x] Baroque - (Black screen on boot. Reported by Topken) - Fixed in #4035 
- [x] Brothers in Arms: Road to Hill 30 - (Broken. Reported by refraction) - Fixed at some point before 1.6
- [x] Crash Nitro Kart - (Broken. Reported by Coornio) - Fixed in #4035 
- [x] Def Jam: Fight for New York - (Hangs before title screen. Reported by indianchan) Fixed in #3971
- [x] DT Carnage - (Hangs on menu. Reported by CK1)
- [x] DT Racer - (Reports infinite loop Reported by refraction) Fixed at some point
- [x] Dog's Life - (Hangs before menu/gameplay, Reported by FlatOut) - Fixed in #3971 - Thrillville is probably fixed as well
- [x] FIFA series - (multiple issues, Reported by lightningterror, jesalvein, Ryudo, FlatOut) - Fixed in #3971
- [x] Fight Night Round 3 - (Crashes. Reported by shogunreaper) - Partially fixed in #3971, remaining issues fixed in https://github.com/PCSX2/pcsx2/commit/7b7278bc85fa3becd39c47168fe6c912523b4d43
- [x] Forever Kingdom (Crashes going in-game. Reported by CK1) - Fixed at some point.
- [x] Gradius V - (Black screen. Reported by billyash) - Fixed in #4035 
- [x] Grandia Xtreme - (Graphical glitches. Reported by jesalvein) - Fixed at some point.
[...]

PR #2032, opened 2017-07-29
PCSX2-WX: Force disable MTVU hack on safest preset

ssakash:
**Summary of changes**:

* Force disable MTVU hack on safest preset. (Reasons given at commit message)
* Minor modifications to improve the issue template.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,14 +1,14 @@
 ***PCSX2 version:***
 ``(For example, stable build v1.4.0 (or) development builds v1.5.xxx)``
 
-- **Note**: It's preferred to only report bugs which are present on the [latest development builds](http://pcsx2.net/download/development.html).
-
-- **Note2**: The new backwards compatibility (psx) mode is in active development. Issue reports are too early at this point.
+- **Notes**:
+  - It's preferred to only report bugs which are present on the [latest development builds](http://pcsx2.net/download/development.html).
+  - The PSX backward compatibility mode is still in its early stage of development. Issue reports are too early at this point.
 
 ***PCSX2 options:***
 ``(For example, Clamping/Rounding, Speedhacks, GameFixes)``
 
-- **Note**: Generally it's recommended to have settings as default as possible and use the ``Safest`` preset along with MTVU speed hack disabled. If there's no issue with these settings but there is an issue with some other non-default option, then please state it explicitly.
+- **Note**: Generally it's recommended to have settings as default as possible and use the ``Safest`` preset. If there's no issue with these settings but there is an issue with some other non-default option, then please state it explicitly.
 
 
 ***Plugins used:***
--- a/pcsx2/gui/Panels/SpeedhacksPanel.cpp
+++ b/pcsx2/gui/Panels/SpeedhacksPanel.cpp
@@ -281,7 +281,8 @@ void Panels::SpeedHacksPanel::EnableStuff( AppConfig* configToUse )
 	m_check_waitloop->Enable(HacksEnabledAndNoPreset);
 	m_check_fastCDVD->Enable(HacksEnabledAndNoPreset);
 
-	m_check_vuThread->Enable(hacksEnabled); // MTVU is unaffected by presets
+	// Grayout MTVU on safest preset
+	m_check_vuThread->Enable(hacksEnabled && (!hasPreset || configToUse->PresetIndex != 0));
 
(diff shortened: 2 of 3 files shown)


lightningterror:
Sounds good considering there are games that won't boot with mtvu.

gregory38:
@ramapcsx2 what do you think of this MTVU/preset commit ?

@ssakash what happen for others preset levels ?

ssakash:
> @ssakash what happen for others preset levels ?

The MTVU hack won't be grayed out on other preset levels, the end user would still be able to enable or disable it.

pcsx2/gui/Panels/SpeedhacksPanel.cpp
```diff
@@ -281,7 +281,8 @@ void Panels::SpeedHacksPanel::EnableStuff( AppConfig* configToUse )
 	m_check_waitloop->Enable(HacksEnabledAndNoPreset);
 	m_check_fastCDVD->Enable(HacksEnabledAndNoPreset);
 
-	m_check_vuThread->Enable(hacksEnabled); // MTVU is unaffected by presets
+	// Grayout MTVU on safest preset
+	m_check_vuThread->Enable(hacksEnabled && ((configToUse->PresetIndex != 0 && hasPreset) || !hasPreset));
```
turtleli: I think this could be simplified to:
`m_check_vuThread->Enable(hacksEnabled && (!hasPreset || configToUse->PresetIndex != 0));`

pcsx2/gui/Panels/SpeedhacksPanel.cpp
```diff
@@ -309,12 +310,14 @@ void Panels::SpeedHacksPanel::ApplyConfigToGui( AppConfig& configToApply, int fl
...
 
 	m_check_vuFlagHack->SetValue(opts.vuFlagHack);
-	if( !(flags & AppConfig::APPLY_FLAG_FROM_PRESET) )
-		m_check_vuThread	->SetValue(opts.vuThread);
 	m_check_intc->SetValue(opts.IntcStat);
 	m_check_waitloop->SetValue(opts.WaitLoop);
 	m_check_fastCDVD->SetValue(opts.fastCDVD);
 
+	bool preset_request = flags & AppConfig::APPLY_FLAG_FROM_PRESET;
+	if (!preset_request || (preset_request && configToApply.PresetIndex == 0))
```
turtleli: Could be simplified to:
`if (!preset_request || configToApply.PresetIndex == 0)`

Not sure you need the `preset_request` variable anymore. Probably should be made const if you do keep it.

ISSUE_TEMPLATE.md
```diff
@@ -1,14 +1,14 @@
 ***PCSX2 version:***
 ``(For example, stable build v1.4.0 (or) development builds v1.5.xxx)``
 
-- **Note**: It's preferred to only report bugs which are present on the [latest development builds](http://pcsx2.net/download/development.html).
-
-- **Note2**: The new backwards compatibility (psx) mode is in active development. Issue reports are too early at this point.
+- **Notes**:
+  - It's preferred to only report bugs which are present on the [latest development builds](http://pcsx2.net/download/development.html).
+  - The new PSX backward compatibility mode is still in its nascent state. Issue reports are too early at this point.
```
turtleli: Side note: I'm not sure that non-native English speakers or indeed, native English speakers, will be familiar with the meaning of "nascent". Personally I understood it via contextual clues at first, and then I looked it up to confirm the meaning.

ramapcsx2:
Yea, this is fine. MTVU is not 100% stable. 
I remember there were some issues with some games, don't ask me which they were though :p

lightningterror:
@ramapcsx2 https://github.com/PCSX2/pcsx2/issues/1669 here they are.

ISSUE_TEMPLATE.md
ssakash: In my opinion, ``"Nascent"`` just seems like a perfect word to be used over here. Also the intention is to prevent users from reporting PSX mode bug reports which is made clear by the statement following it  (``Issue reports are too early at this point``).

gregory38: I agree with turtleli, it is too complex. You could also use `early state` but something low level

gregory38:
Except the wiki wording code is fine.

ssakash:
Updated the issue template to remove the nascent part, I'll probably merge this in 2-3 days. If anyone has any objections with this PR, do mention about it before I do so.

PR #2033, opened 2017-08-03
GSdx-ogl: Add AMD SSO workaround for older drivers/gpus

lightningterror:
Add older drivers to amd's buggy driver list. Fixes the broken driver on
older gpus using older drivers.
http://forums.pcsx2.net/Thread-OpenGL-GS-plugin-failed-to-open?pid=567330#pid567330

fglrx_buggy_driver and legacy_fglrx_buggy_driver renamed to amd_buggy_driver and legacy_amd_buggy_driver.

### DIFF
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -266,13 +266,13 @@ namespace GLLoader {
 
 	bool s_first_load = true;
 
-	bool legacy_fglrx_buggy_driver = false;
-	bool fglrx_buggy_driver    = false;
-	bool mesa_buggy_driver     = false;
-	bool nvidia_buggy_driver   = false;
-	bool intel_buggy_driver    = false;
-	bool in_replayer           = false;
-	bool buggy_sso_dual_src    = false;
+	bool amd_legacy_buggy_driver = false;
+	bool vendor_id_amd      = false;
+	bool vendor_id_nvidia   = false;
+	bool vendor_id_intel    = false;
+	bool mesa_driver        = false;
+	bool in_replayer        = false;
+	bool buggy_sso_dual_src = false;
 
 
 	bool found_geometry_shader = true; // we require GL3.3 so geometry must be supported by default
@@ -341,26 +341,31 @@ namespace GLLoader {
 
 		// Name changed but driver is still bad!
 		if (strstr(vendor, "Advanced Micro Devices") || strstr(vendor, "ATI Technologies Inc.") || strstr(vendor, "ATI"))
-			fglrx_buggy_driver = true;
-		if (fglrx_buggy_driver && (
-					strstr((const char*)&s[v], " 15.") || // Blacklist all 2015 AMD drivers.
-					strstr((const char*)&s[v], " 16.") || // Blacklist all 2016 AMD drivers.
-					strstr((const char*)&s[v], " 17.") // Blacklist all 2017 AMD drivers for now.
-					))
-			legacy_fglrx_buggy_driver = true;
-
+			vendor_id_amd = true;
+		/*if (vendor_id_amd && (
+				strstr((const char*)&s[v], " 10.") || // Blacklist all 2010 AMD drivers.
+				strstr((const char*)&s[v], " 11.") || // Blacklist all 2011 AMD drivers.
+				strstr((const char*)&s[v], " 12.") || // Blacklist all 2012 AMD drivers.
+				strstr((const char*)&s[v], " 13.") || // Blacklist all 2013 AMD drivers.
+				strstr((const char*)&s[v], " 14.") || // Blacklist all 2014 AMD drivers.
+				strstr((const char*)&s[v], " 15.") || // Blacklist all 2015 AMD drivers.
+				strstr((const char*)&s[v], " 16.") || // Blacklist all 2016 AMD drivers.
+				strstr((const char*)&s[v], " 17.") // Blacklist all 2017 AMD drivers for now.
+				))
+			amd_legacy_buggy_driver = true;
+		*/
 		if (strstr(vendor, "NVIDIA Corporation"))
-			nvidia_buggy_driver = true;
+			vendor_id_nvidia = true;
 
 #ifdef _WIN32
 		if (strstr(vendor, "Intel"))
-			intel_buggy_driver = true;
+			vendor_id_intel = true;
 #else
 		// On linux assumes the free driver if it isn't nvidia or amd pro driver
-		mesa_buggy_driver = !nvidia_buggy_driver && !fglrx_buggy_driver;
+		mesa_driver = !vendor_id_nvidia && !vendor_id_amd;
 #endif
 
-		buggy_sso_dual_src = intel_buggy_driver || fglrx_buggy_driver || legacy_fglrx_buggy_driver;
+		buggy_sso_dual_src = vendor_id_intel || vendor_id_amd /*|| amd_legacy_buggy_driver*/;
 
 		if (theApp.GetConfigI("override_geometry_shader") != -1) {
 			found_geometry_shader = theApp.GetConfigB("override_geometry_shader");
@@ -443,13 +448,13 @@ namespace GLLoader {
 		status &= status_and_override(found_GL_ARB_get_texture_sub_image, "GL_ARB_get_texture_sub_image");
 
 		if (s_first_load) {
-			if (fglrx_buggy_driver) {
+			if (vendor_id_amd) {
 				fprintf(stderr, "The OpenGL hardware renderer is slow on AMD GPUs due to an inefficient driver.\n"
 					"Check out the link below for further information.\n"
 					"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know\n");
 			}
 
-			if (intel_buggy_driver) {
+			if (vendor_id_intel) {
 				fprintf(stderr, "The OpenGL renderer is inefficient on Intel GPUs due to an inefficient driver.\n"
 					"Check out the link below for further information.\n"
 					"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-Intel-GPUs-All-you-need-to-know\n");
--- a/plugins/GSdx/GLLoader.h
+++ b/plugins/GSdx/GLLoader.h
@@ -358,11 +358,11 @@ extern PFNGLGENERATEMIPMAPPROC                  glGenerateMipmap;
 namespace GLLoader {
 	void check_gl_requirements();
 
-	extern bool fglrx_buggy_driver;
-	extern bool legacy_fglrx_buggy_driver;
-	extern bool mesa_buggy_driver;
-	extern bool nvidia_buggy_driver;
-	extern bool intel_buggy_driver;
+	extern bool vendor_id_amd;
+	extern bool vendor_id_nvidia;
+	extern bool vendor_id_intel;
+	extern bool amd_legacy_buggy_driver;
+	extern bool mesa_driver;
 	extern bool buggy_sso_dual_src;
 	extern bool in_replayer;
 
--- a/plugins/GSdx/GSDeviceOGL.cpp
+++ b/plugins/GSdx/GSDeviceOGL.cpp
@@ -528,7 +528,7 @@ bool GSDeviceOGL::Create(const std::shared_ptr<GSWnd> &wnd)
 	// Get Available Memory
 	// ****************************************************************
 	GLint vram[4] = {0};
-	if (GLLoader::fglrx_buggy_driver) {
+	if (GLLoader::vendor_id_amd) {
 		// Full vram, remove a small margin for others buffer
 		glGetIntegerv(GL_TEXTURE_FREE_MEMORY_ATI, vram);
 	} else if (GLLoader::found_GL_NVX_gpu_memory_info) {
@@ -998,7 +998,7 @@ void GSDeviceOGL::SelfShaderTestRun(const string& dir, const string& file, const
 
 #ifdef __linux__
 	// Nouveau actually
-	if (GLLoader::mesa_buggy_driver) {
+	if (GLLoader::mesa_driver) {
 		if (freopen(out.c_str(), "w", stderr) == NULL)
 			fprintf(stderr, "Failed to redirect stderr\n");
 	}
@@ -1010,7 +1010,7 @@ void GSDeviceOGL::SelfShaderTestRun(const string& dir, const string& file, const
 
 #ifdef __linux__
 	// Nouveau actually
-	if (GLLoader::mesa_buggy_driver) {
+	if (GLLoader::mesa_driver) {
 		if (freopen("/dev/tty", "w", stderr) == NULL)
 			fprintf(stderr, "Failed to restore stderr\n");
 	}
--- a/plugins/GSdx/GSShaderOGL.cpp
+++ b/plugins/GSdx/GSShaderOGL.cpp
@@ -224,7 +224,7 @@ std::string GSShaderOGL::GenGlslHeader(const std::string& entry, GLenum type, co
 	} else {
 		header += "#define DISABLE_GL42_image\n";
 	}
-	if (GLLoader::fglrx_buggy_driver || GLLoader::intel_buggy_driver)
+	if (GLLoader::vendor_id_amd || GLLoader::vendor_id_intel)
 		header += "#define BROKEN_DRIVER as_usual\n";
 
 	// Stupid GL implementation (can't use GL_ES)
@@ -331,7 +331,7 @@ GLuint GSShaderOGL::CompileShader(const std::string& glsl_file, const std::strin
 // GLSL improvement (unfortunately).
 int GSShaderOGL::DumpAsm(const std::string& file, GLuint p)
 {
-	if (!GLLoader::nvidia_buggy_driver) return 0;
+	if (!GLLoader::vendor_id_nvidia) return 0;
 
 	GLint   binaryLength;
 	glGetProgramiv(p, GL_PROGRAM_BINARY_LENGTH, &binaryLength);


mirh:
> fglrx_buggy_driver and legacy_fglrx_buggy_driver renamed to legacy_amd_buggy_driver and amd_buggy_driver.

The other way around if any. 
Code is correct though. 

In other news, what's the point in blacklisting specific driver releases, when _anyway_ just ANY is bugged?

MrCK1:
So isn't this still the user's fault for using older driver? Like @mirh said, it seems redundant to try and blacklist every possible driver - AMD's drivers are such a clusterfuck.

lightningterror:
> ATI Technologies Inc.

That was renamed from AMD to ATI Technologies Inc. this year I think. I added that in the code myself in the previous pr.

> In other news, what's the point in blacklisting specific driver releases, when anyway just ANY is bugged?

I asked Gregory about it , I wanted to remove the driver version check and just detect vendor ID but he said it might be useful in the future. Perhaps if they fix the drivers.

mirh:
> So isn't this still the user's fault for using older driver?

Yes and no. 
First, older cards aren't getting newer/fixed drivers in the first place. 
Second, even if newer drivers were available.. User should at least be warned. 

> Like @mirh said, it seems redundant to try and blacklist every possible driver - AMD's drivers are such a clusterfuck.

Then you hear what dolphin folks are banging their head on.. And oh boy. At least AMD shines on linux. 

> In other news, what's the point in blacklisting specific driver releases, when anyway just ANY is bugged?

I'll answer to myself: there is a point if we wanted to let SSO work with those older releases where bug was partially fixed. 

> I wanted to remove the driver version check and just detect vendor ID but he said it might be useful in the future. Perhaps if they fix the drivers.

He has a point.

lightningterror:
> He has a point.

It might actually be a bad idea. Especially if the driver is not fixed till 2018/19/never and people start getting the same issue on pcsx2 1.6  on newer drivers because newer aren't blacklisted.

My original idea was to re introduce it for older drivers once the drivers got fixed.

mirh:
> It might actually be a bad idea. Especially if the driver is not fixed till 2018/19/never and people start getting the same issue on pcsx2 1.6 on newer drivers because newer aren't blacklisted.

Well, everything depends on your expectation of the bug getting fixed.. 
Anyway what I was talking about was just blacklisting the vendor name. 
Or even better.. If we could convert the initial part of the driver revision string to an integer, it would be extremely handful for comparisons.

gregory38:
I would prefer pro_amd name. Because, on linux the free amd driver is working fine. But not the pro amd driver. Fglrx was the inital name of the OpenGL implementation before the N marketing rename (which actually work).

mirh:
Ehrm.. can we decide which use-cases AT THE MOMENT we'd have to support first of all?
Iirc the only meaningful difference to this point (unless I or the other guy get in the mood to even test older releases) is the kind of fixed state we had last year.

gregory38:
Note: I will also be fine to `#if 0 ... #endif` the driver detection code and to always enable it. I want to keep the code for the future AMD bugs/fixes.

gregory38:
By the way, the initial intent of the code is completely broken but it isn't important. You will also notice that all drivers got sso disabled.

lightningterror:
> By the way, the initial intent of the code is completely broken but it isn't important. You will also notice that all drivers got sso disabled.

Oops , I didn't actually notice that :) Thought the code was working for only added drivers in the blacklist. So what now?

gregory38:
I'm fine with new code. It works as bad as before ;) Is everybody happy ?

mirh:
I still haven't clear why *now* we are making this amd and legacy_amd distinction
And what's the point into blacklisting every release ever one by one, instead of just basing on vendor name.

gregory38:
Goal should be to distinguish between legacy and non-legacy amd driver. You can't do it with the vendor name. So you shouldn't need to blacklisting every release.

lightningterror:
Well 1.4 works on vendor detection but the user on the forum still experienced the issue.
I think we are missing another vendor id that is missing. Oh how amd loves changing them.

mirh:
> So you shouldn't need to blacklisting every release.

Well, at the moment that's what's enough. 

And, if both of you are right (1. all newest drivers are ATI Technologies Inc. && we shouldn't get to admit any older one) in the future we could just as well check if `(stoi((const char*)&s[v]) > *int*)`

Then, now that I check.. I'm not sure why you are using "nvidia_buggy" and "fglrx_buggy" namings just to inverse-check for mesa driver.

gregory38:
```
Then, now that I check.. I'm not sure why you are using "nvidia_buggy" and "fglrx_buggy" namings just to inverse-check for mesa driver.
```
Maybe need a linux only, otherwise it is true.

mirh:
Yes, yes. But then: that's the whole point for the existence of both amd_buggy and nvidia_buggy. 
Do you really need the variables, instead of just throwing everything in that one single instance?

Also, cannot you just check.. whatever glxinfo seems to have hooked to "client glx vendor string" ?
It tells you directly and precisely about Mesa.

gregory38:
Honestly what the point to discuss those stuff ? Mesa contains multiple drivers. So why bother ? There are 4 possibilities. In short, Mesa is the default on Linux.
* Nvidia pro
* AMD pro
* Mesa
* no driver...

lightningterror:
I'm currently testing "atioglxx.dll" versions , right now testing Catalyst 15.11 and there are some weird issues , games flicker with a white screen so I can see  why Greg wanted legacy as well. Hmm got any idea which extension could be causing this issue ?

mirh:
> Honestly what the point to discuss those stuff ? 

Removing overcomplicated stuff?

> I'm currently testing "atioglxx.dll" versions , right now testing Catalyst 15.11 and there are some weird issues , games flicker with a white screen so I can see why Greg wanted legacy as well. 

I'm not \*so\* sure you can that much peacefully swap dlls between distant drivers version tbh.

lightningterror:
> I'm not *so* sure you can that much peacefully swap dlls between distant drivers version tbh.

I wasn't very lucky to get older ones to work , maybe because the gpu is newer , anyway I just place the dll in the pcsx2 directory and that way I know if it works or not.

mirh:
> I wasn't very lucky to get older ones to work , maybe because the gpu is newer 

I was going to tell you that "come on that should not matter"... Though a R7 360 would perhaps be lucky to be supported in 15.6. Then 15.7 was a pretty messy release for various reason and.. whatever. 
Anyway, one of these day, I could test older driver releases on my pre-gcn laptop. 

> anyway I just place the dll in the pcsx2 directory and that way I know if it works or not.

Yes, sure, that's also how I can "happily" use the SSO-fixed atioglxx.dll even with latest SSO-broken driver, though being just a single branch afar is pretty different than two years.

lightningterror:
> Anyway, one of these day, I could test older driver releases on my pre-gcn laptop.

The Catalyst Preview driver is what I'm interested in. Since the OP had issues with SSO on 1.4
http://forums.pcsx2.net/Thread-OpenGL-GS-plugin-failed-to-open?pid=567330#pid567330

mirh:
> Since the OP had issues with SSO on 1.4

OP was having issues with a 2011 driver for the records. 

> The Catalyst Preview driver is what I'm interested in

Let's say that they messed up hard during that period of times with non-GCN drivers.

lightningterror:
@mirh How does this look ? Feel like testing ?

mirh:
You can find a tl;dr [here](https://github.com/PCSX2/pcsx2/issues/1552#issuecomment-279744823). 
If any though, I'm not sure the user whether user is "protected by BSOD" if he uses just the dll of older drivers. 

> Feel like testing ?

I would on my GCN computer.. But I'm in the middle of a messy 3 months of uptime and I cannot tinker with stuff that makes me reboot.

gregory38:
Tbh, I'm rather against it. Whitelist few old drivers isn't a good idea. In particular when SSO speed isn't that better on AMD. More code and no gain. 

I think we should go back to
* detect AMD driver by name
* separate legacy / recent driver based on the version
* blacklist SSO on both legacy / recent driver

Playing nice with very old drivers (from the OP) isn't the goal of the project.

gregory38:
The goal of this code is to easily detect driver. So far there is no difference, but code is ready if one day we need to add a difference. It would be costlier to add back the code than to remove the 3 lines. And maybe one day AMD will fix SSO.

```
I see. Though, would you know why a driver wouldn't pass your testcase.. While instead working nicely in pcsx2?
```
Because there are plenty of bugs. If you tune the testcase differently you might hit another bug.

mirh:
> So far there is no difference, but code is ready if one day we need to add a difference. It would be costlier to add back the code than to remove the 3 lines.

Fine. 
But THEN, in that case, you'd still just have \*A\* broken driver ["series"]. 

So I'm not sure how the whitelisting here, would have any different coding than whatever we are gonna do. 
Or why those amd_buggy and nvidia_buggy when in fact the variables have nothing to do with a bug in the driver.

gregory38:
It used to be to avoid various driver bugs. Eventually bug got fixed. Anyway, you could rename them if you want. Otherwise current name helps to start silly discussion on reddit ;)

mirh:
> It used to be to avoid various driver bugs. 

*Oooohh*
@lightningterror axe eeem

> Otherwise current name helps to start silly discussion on reddit ;)

Argh, I'm having second thoughts now. 

---
In other news.. I have spent playing with the dlls. 
And putting aside there's no way in this world for me to get `override_GL_ARB_separate_shader_objects` to work, swapping dlls seems to have no effect at all on the goddamn "Compatibility Profile Context" version.

lightningterror:
> And putting aside there's no way in this world for me to get override_GL_ARB_separate_shader_objects to work, swapping dlls seems to have no effect at all on the goddamn "Compatibility Profile Context" version.

Did you remove the code from pcsx2 to get sso to work?

mirh:
? I was talking about the gsdx.ini setting I also mentioned [here](https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-AMD-GPUs---All-you-need-to-know). 
For some reason, it doesn't work. Console isn't mentioning the override.

lightningterror:
I'm a bit confused . you can't get sso to work with swapping dlls because it detects the driver and not the dll version. And yes I agree that downgrading drivers is a bad idea. The best solution is to detect the OpenGL build version / file version , I kinda dunno how to implement that xD
I'll need a hand.

mirh:
Autodetection is good if it "avoids configuring" yourself. 
If dll detection doesn't work (mfw I always thought about the contrary).. Really no point into whitelisting drivers ("as a whole") that nevertheless one shouldn't stick to in general because for those same cards quite newer ones exist. 

So just two things imo are needed here:
* tell _every_ AMD user to GTFO (even though, it wouldn't be bad if you could implement a stub of that string to int thingy)
* check if override_SSO works for you because screw myself, it didn't here

Then if gregory feels in the mood he could also check whether there aren't neater ways for the Mesa check (or drop it? is it still needed today?)

gregory38:
Maybe it will be nice to say what you want to achieve ?
* renaming variable => OK
* restore legacy / latest driver difference => OK

```
Then if gregory feels in the mood he could also check whether there aren't neater ways for the Mesa check (or drop it? is it still needed today?)
```
Mesa check is neat. Only a single line. And so far, no feature are disabled.

mirh:
> Maybe it will be nice to say what you want to achieve ?

TL;DR for the moment.. just blacklist every "Advanced Micro Devices"  || "ATI", profit. 

And rather than this "strstr" feast, I think there could be neater way for the future to check for driver version. But I guess we may only know/see in the future. 

> Mesa check is neat. Only a single line. And so far, no feature are disabled.

So why at all? xD

ameenross:
Just out of curiosity if I may, can't Gallium3D be used? To prevent HW specific code altogether?

gregory38:
> So why at all? xD

It is just handy to have a way to detect current brand when you create new code. Not all GL implementations are equivalent. So instead to wait 1 year that all drivers fix their bugs. I prefer to put a workaround for a couple of months. This file is 0% of GSdx complexity.

> Just out of curiosity if I may, can't Gallium3D be used? To prevent HW specific code altogether?

??? Goal is to workaround driver bug. Or to use the API in an efficient way. Besides there is no way that common code perform the same on different hardware.

mirh:
> It is just handy to have a way to detect current brand when you create new code. 

Oh.. so it's a just stub. 
Make sense then (even though.. I dunno, I'm not seeing in which way \*all\* mesa could regress for \*all\* drivers symoultaneously in 2017 - and in a way that wouldn't just be better fixed in a day upstream)

It feels a tad odd that we use vendor_buggy to detect the vendor, and vendor_buggy_something to detect actual bugs tho.

ameenross:
> ???

I apologize if that was a stupid question, I don't have any experience with graphics programming.

> Besides there is no way that common code perform the same on different hardware.

I assumed that there would be workarounds in Gallium3D for broken/cripple hardware drivers, so when developing for Gallium you don't have to worry about that? Maybe my assumption was wrong.

mirh:
Gallium3D \*is\* **THE** (well, part of at least) graphics driver.

ameenross:
> Gallium3D is *THE* (well, part of at least) graphics driver.

Hehe, sorry, I kind of misunderstood the position of Gallium. But there's no graphics utopia yet then :/

lightningterror:
> It feels a tad odd that we use vendor_buggy to detect the vendor, and vendor_buggy_something to detect actual bugs tho.

vendor_id_nvidia/amd/intel/ or vendor_detect_nvidia ..etc would have been a better naming :)

gregory38:
> Make sense then (even though.. I dunno, I'm not seeing in which way *all* mesa could regress for *all* drivers symoultaneously in 2017 - and in a way that wouldn't just be better fixed in a day upstream)

Yet, it already happened in the past. The detection is coarse to not bother. There are only **temporary workaround** (cough, cough, AMD). And something behavior is different. Swap extension are different on mesa versus Nvidia/AMD. Sometime Mesa follow the standard differently etc... The cost fo those lines of code is 0, they days you need it you are happy. It is like a seat belt on a car, it is useless yet it saves life.

@ameenross 
By driver, I mean the full GPU stack (including OpenGL/Vulkan/Xorg). There are bugs in all layers of code. In gallium3D and even in the hardware itself. I can tell you that all serious applications that use GL have code dedicated for each vendor.

@lightningterror 
Yes but you aren't the poor guy that spent months to workaround those pesky bugs ;)

mirh:
> Sometime Mesa follow the standard differently

Ok, legit then, but it hurts my head to get to read the code, see `buggy`, assume X.. while instead it's not-X. 

>  Yes but you aren't the poor guy that spent months to workaround those pesky bugs ;)

😃

gregory38:
> Ok, legit then, but it hurts my head to get to read the code, see buggy, assume X.. while instead it's not-X.

Sigh...
```
renaming variable => OK
```

lightningterror:
How about commenting out the driver version like this , that way it won't do the check since it didn't work properly and you can still use the code in the future , oh and also renamed the buggy driver names.

mirh:
Rename amd_pro_buggy to amd_legacy_buggy
Then also include that into the commenting

Rename mesa_buggy_driver to mesa_driver
And I guess I could call it a day

lightningterror:
Still waiting for those pre gcn tests xD

mirh:
WTF TEST CASE IS GREEN ON 11.7
(pcsx2 still crash then, for lack of glDebugMessageCallback, glBindImageTexture, glMemoryBarrier, glTexStorage2D but whatever)

lightningterror:
> Btw I'm still also getting ATI Technologies Inc. vendor here..

Odd , did you use DDU to remove the drivers ?

mirh:
Only the first time (when going from 2016 modded driver back to 2011), but still.

gregory38:
Good for me. Let's merge it.

mirh:
Cool. 
Now I'd be curios to know where does Advanced Micro Devices thingy is coming from.

gregory38:
The linux proprietary driver.

mirh:
:\
Just checked. 
Glxinfo indeed shows OpenGL vendor string as such. 
OpenGL information inside pcsx2 is still the usual ATI Technologies Inc. though

(and in other news I can only get a totally blank window in hardware mode, even without overriding any setting)

lightningterror:
I tried the llvmpipe driver and the vendor change works which is VMware.

On a side note override_GL_ARB_separate_shader_objects doesn't work because the override was removed. It's still there in the legacy code tho.

mirh:
Yeah, guessed like. 

Anyway, I'm pressing over this petty issue, because if it was the case that string is ATI Technologies Inc. since the dawn of times (or well, at least the last 3-4 years we might be caring about).. It'd be a cakewalk to pass a `substr` to `stoi` and get a very handy integer to evaluate against. 
If I can explain.

mirh:
How this is done in rpcs3
https://github.com/RPCS3/rpcs3/pull/3618

mirh:
As if I weren't full enough of Windows's problems, I just finished some linux testing with fglrx (ya know, opencl doesn't work on open driver). 
Seems like latest ever 15.11/15.12 version (aka 15.30.3) fixed an "everything white" issue, present [at least!] on boot and bios menu. 

I guess this should at least warrant "printing a warning" should you not be up-to-date, but the thing is that unless we want to pass via dmesg/libgl_debug/xorg logs grepping, detection here has to (funnily?) rely on the \*first\* version number of the GL_VERSION string instead. 

---
In other news, after banging my head over why in the hell the same GL_VENDOR string in glxinfo was different than in pcsx2, I have damn finally realized the solution: it's AMD for 64-bit applications, and ATI for 32-bit ones. 
Confirmed by diffing glxinfo32 and glxinfo verbose output, diverging only in that. 

We could have
`if (atof(glGetString(GL_VENDOR).substr(2,7)) < 5,13416) do warn`

PR #2035, opened 2017-08-08
PCSX2-GUI Logo Update

Ryudo300:
Logo update for PCSX2 GUI.


ssakash:
Before someone asks on why there is a ``wxWidgets`` label - Yes, there are no wxWidgets related changes and we're only updating an image but we still use the ``wxImage`` class reference for handling the image, so it makes sense to also add this label for technical correctness I guess.

gregory38:
Would it be possible to have a slightly wider logo (10-20px). It could help to reduce height requirement for low-res screen. The size of the logo is used to compute all GUI size.

Ryudo300:
I used the same size as the old logo... but I can try to add 10 pixels. Are you sure it won't look blurry though, because it'll be a completely different size from the old logo, and if it then has to be forcefully shrunk down after that you'll definitely be looking at blurriness.

gregory38:
Normally it should use the new size. Size is read from the png file directly.

ssakash:
Superseded by #2036. (This can't be updated anymore as he seems to have deleted this branch, look at ``unknown repository`` mention at the source branch name)

PR #2036, opened 2017-08-08
PCSX2-GUI: Logo Update (Wider image)

Ryudo300:
New logo for the main interface of PCSX2. (400px instead of 380)


Ryudo300:
Sorry for the mess with the PRs.... still new with this whole merge request thing and branches, etc.

lightningterror:
Some weird stuff is going on with his fork , managed to fix it somehow .. for now at least. I still see other issues if we try to rebase.

Ryudo300:
Welp, with a lot of help from LT it finally seems to have gone through ok now. :D

gregory38:
Perfect

turtleli:
Perhaps the filesize could be reduced? I ran zopflipng on the image and it reduced the filesize from 102KB to 70KB.

Ryudo300:
I can bring it down to 77KB if necessary. But going further below that will result in reduced quality. Not sure if that additional 7KB is worth the quality reduction afterwards, but yeah. This logo is bigger in resolution than the older one, which automatically results in a bit bigger file size. So yeah. 77KB is doable at least. Alternatively if I go with PNG8 it would be at 36KB, but that's a major quality loss lol. Best to keep it at PNG24 for the best quality per byte.

turtleli:
> I can bring it down to 77KB if necessary. But going further below that will result in reduced quality. Not sure if that additional 7KB is worth the quality reduction afterwards

Just to clarify - zopflipng uses lossless compression (assuming you don't use the lossy options). It does a better job at reducing the filesize, but it does take much longer to do so.

gregory38:
Are we good on the PNG size ?

turtleli:
backgroundlogozopfli

Attached image is 70KB. I used imagemagick to make sure the image is identical to the source image.

MrCK1:
If the image is wider, don't we need to widen the GUI window as well?

turtleli:
No. Sizing is automatic.

Ryudo300:
Alright. Updated with the 70kb one.

gregory38:
@MrCK1 actually the goal is to make the GUI wider because all size are based on the image size. Hopefully it is better for 720p screen.

PR #2041, opened 2017-08-15
GSdx - dx9: Edit Alpha Correction (FBA) tooltip. 

lightningterror:
Edit tooltip to FBA. Causes graphical glitches in some games.

Affected games are DBZ BT3.
It makes some textures look cut off or adds a blue texture in front of the screen when enabled but it also fixes the health bars. Otherwise the health bars do not work properly but there are no glitched textures.
BT2 might also be affected as they share the same engine.
I don't know about others , maybe.

### DIFF
--- a/plugins/GSdx/GSSetting.cpp
+++ b/plugins/GSdx/GSSetting.cpp
@@ -159,7 +159,7 @@ const char* dialog_message(int ID, bool* updateText) {
 #ifdef _WIN32
 		// DX9 only
 		case IDC_FBA:
-			return "Makes textures partially or fully transparent as required by emulation. May cause unusual slowdowns for some games.";
+			return "Makes textures partially or fully transparent as required by emulation. May cause unusual slowdowns or graphical glitches for some games.";
 		case IDC_LOGZ:
 			return "Treat depth as logarithmic instead of linear. Recommended setting is on unless it causes graphical glitches.";
 #endif


MrCK1:
I don't think we really need to bother with this. The DX9 renderer will be removed soon once we get rid of some other dependencies.

lightningterror:
Maybe 1 last commit before we see dx9 off xD
Anyway Isn't it getting removed after 1.6 release , if so then It's not a bad idea for a stable release.

willkuer:
If it would only break graphics accuracy and introduces slowdown there would be no point to add this feature. Why was it added in the first place? Which games were running better using that option? How are these games running nowadays while disabling this options?

There are probably regressions due to this PR.

lightningterror:
Looking at the Gsdx wiki I managed to find "Super Robot Taisen Original Generations".

gregory38:
FBA is the normal behavior of the GS. IMHO, the option was added as a speed hack. But I suspect alpha emulation is already broken so sometimes it could hurt. I'm not sure what is the best default setting.

lightningterror:
It probably can be fixed but this is dx9 we are talking about.

gregory38:
No I don't think it can be fixed. Anyway, I'm afraid it is a bit too late to update dx9 default behavior. It would be better to add an info in the wiki saying that DBZ game is better with the option off.

lightningterror:
All this does is just uncheck it by default. Not a major change.

ssakash:
> All this does is just uncheck it by default. Not a major change.

The GS does alpha correction on the alpha bits before writing data to the framebuffer (Similar to dithering and clamping on the RGB values). FBA enabled is the accurate behaviour of GS, it doesn't make sense to have a non-accurate value as the default one. (or) are you going for the approach that DX9 is already too hacky so let's compromise some accuracy for even more speed?

gregory38:
The thing is that you can expect some guides that told you to disable this option for some games. But none (so far) will tell you that you need this option.

lightningterror:
Alpha correction on
pcsx2 2017-08-17 16-07-29-58

There's also another issue where textures from distance look a bit weird or cut off.

Alpha correction off
pcsx2 2017-08-17 16-08-01-84

The bugs present are fixed but another issue appears , health bars do not work with the off setting.

mirh:
🤔
Strange, I remembered the motto was "it's not like we despise old stuff - it's just we ain't caring (unless it's a burden)". 

Anyway, if any I think the thing is that you cannot just act on the basis of just a single data point (iirc it was also raised in another PR recently). 
I'm not saying "make a poll, ask the world whatever" (even though that would help) but at least doing some searches to try to estimate whether it causes more problems than it solves.

MrCK1:
Let's be realistic here, I doubt anyone is even using the DX9 renderer for anything and XP support is already out the window for 1.6. I just don't see the point in this PR.

mirh:
If you don't care, I'm not sure why you are commenting.

lightningterror:
Maybe something like this then? Just a tooltip update.

mirh:
Even the original PR could have been fine.. Just we don't know if there are more games where this improves or worsen things.

gregory38:
Honestly I don't care. Technically I think that enabling FBA doesn't hide the graphical glitch (which isn't the same as generate glitch). But who care. So let's merge this update.

PR #2049, opened 2017-08-22
gsdx: Fix OpenGL vsync, pcsx2: Apply vsync changes immediately

### LINKED ISSUE #842: VSync is limiting the disabled frame limit  in OGL HW mode
I can go above full speed with DX9\11 HW\SW and OGL SW when vsync is enabled but I can't with OGL HW mode(which is I think the fastest mode on any game I tried)

I'm using  PCSX2  1.3.1-20150903224941 and I can't try the latest beta(something is wrong with the buildbot page)

### LINKED ISSUE #1971: Bug/Feature request: Disable vsync temporarily while frame limiting is disabled

turtleli:
Changes:
 * OpenGL renderer - calls the WGL/EGL/GLX swap interval function on the rendering thread so the turbo/framelimiter hotkeys function properly. Fixes #842.
 * Applies vsync change immediately when the vsync settings are changed. Fixes #1971.

### DIFF
--- a/pcsx2/Pcsx2Config.cpp
+++ b/pcsx2/Pcsx2Config.cpp
@@ -237,7 +237,7 @@ void Pcsx2Config::GSOptions::LoadSave( IniInterface& ini )
 
 int Pcsx2Config::GSOptions::GetVsync() const
 {
-	if (g_LimiterMode == Limit_Turbo)
+	if (g_LimiterMode == Limit_Turbo || !FrameLimitEnable)
 		return 0;
 
 	// D3D only support a boolean state. OpenGL waits a number of vsync
--- a/pcsx2/gui/AppCoreThread.cpp
+++ b/pcsx2/gui/AppCoreThread.cpp
@@ -512,7 +512,7 @@ void AppCoreThread::ApplySettings( const Pcsx2Config& src )
 	}
 
 	if (m_ExecMode >= ExecMode_Paused)
-		GSsetVsync(EmuConfig.GS.FrameLimitEnable ? EmuConfig.GS.GetVsync() : 0);
+		GSsetVsync(EmuConfig.GS.GetVsync());
 }
 
 // --------------------------------------------------------------------------------------
--- a/pcsx2/gui/AppCoreThread.cpp
+++ b/pcsx2/gui/AppCoreThread.cpp
@@ -510,6 +510,9 @@ void AppCoreThread::ApplySettings( const Pcsx2Config& src )
 	{
 		_parent::ApplySettings( fixup );
 	}
+
+	if (m_ExecMode >= ExecMode_Paused)
+		GSsetVsync(EmuConfig.GS.FrameLimitEnable ? EmuConfig.GS.GetVsync() : 0);
 }
 
 // --------------------------------------------------------------------------------------
--- a/plugins/GSdx/GSWnd.h
+++ b/plugins/GSdx/GSWnd.h
@@ -60,7 +60,8 @@ class GSWndGL : public GSWnd
 {
 protected:
 	bool m_ctx_attached;
-	int m_vsync;
+	std::atomic<bool> m_vsync_change_requested;
+	std::atomic<int> m_vsync;
 
 	bool IsContextAttached() const { return m_ctx_attached; }
 	void PopulateGlFunction();
@@ -68,11 +69,11 @@ protected:
 	void FullContextInit();
 	virtual void CreateContext(int major, int minor) = 0;
 
-	virtual void SetSwapInterval(int vsync) = 0;
+	virtual void SetSwapInterval() = 0;
 	virtual bool HasLateVsyncSupport() = 0;
 
 public:
-	GSWndGL() : m_ctx_attached(false), m_vsync(0) {};
+	GSWndGL() : m_ctx_attached(false), m_vsync_change_requested(false), m_vsync(0) {};
 	virtual ~GSWndGL() {};
 
 	virtual bool Create(const string& title, int w, int h) = 0;
--- a/plugins/GSdx/GSWndEGL.cpp
+++ b/plugins/GSdx/GSWndEGL.cpp
@@ -235,15 +235,18 @@ GSVector4i GSWndEGL::GetClientRect()
 	return GSVector4i(0, 0, w, h);
 }
 
-void GSWndEGL::SetSwapInterval(int vsync)
+void GSWndEGL::SetSwapInterval()
 {
 	// 0 -> disable vsync
 	// n -> wait n frame
-	eglSwapInterval(m_eglDisplay, vsync);
+	eglSwapInterval(m_eglDisplay, m_vsync);
 }
 
 void GSWndEGL::Flip()
 {
+	if (m_vsync_change_requested.exchange(false))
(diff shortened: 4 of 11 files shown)


mirh:
> The swap interval function must be called on the same thread that
rendering takes place on.

Putting aside.. I wonder if this mightn't have also made a difference when testing the dwmflush affair, I still think any link between "host" and "guest" framerate is wrong. 
If it's really a big deal for you, just make for the .ini supporting the same hotkey for more functions (in this case vsync and framelimiter on/off toggling), but really not by default. 

Then v-sync off >v-sync on (while FL is off) doesn't to anything. 
If I start and stop emulation again though, it finally engages. I'm not sure what was expected behavior, but still it doesn't feel right. 
The opposite (v-sync on > v-sync off) instead seems to work consistently.

pcsx2/gui/AppCoreThread.cpp
```diff
@@ -510,6 +510,15 @@ void AppCoreThread::ApplySettings( const Pcsx2Config& src )
 	{
 		_parent::ApplySettings( fixup );
 	}
+
+	if (m_ExecMode >= ExecMode_Paused)
+	{
+		// Disable Vsync when frame limiter is disabled
```
ssakash: It makes more sense to place the comment near the else statement (or) change the check into ``if (!g_Conf->EmuOptions.GS.FrameLimitEnable)`` but I think I'm being too nitpicky here, your choice.

turtleli: I think it's fine where it is. I kinda just copied the comment over from the old code. I'm not sure the comment is actually necessary though since it simply describes the code (which I think is self explanatory, but maybe it isn't) instead of explaining why.

turtleli:
> Putting aside.. I wonder if this mightn't have also made a difference when testing the dwmflush affair, I still think any link between "host" and "guest" framerate is wrong.
> If it's really a big deal for you, just make for the .ini supporting the same hotkey for more functions (in this case vsync and framelimiter on/off toggling), but really not by default.

Not sure I understand you. PR simply fixes some bugs, it doesn't uncouple the vsync and frame limiter, which is what I assume you are talking about.

> Then v-sync off >v-sync on (while FL is off) doesn't to anything.
> If I start and stop emulation again though, it finally engages. I'm not sure what was expected behavior, but still it doesn't feel right.

I think the pause/resume behaviour should be changed so the frame limiter/vsync behaviour is consistent.

turtleli:
Updated to make the pause/resume behaviour more consistent.

plugins/GSdx/GSWnd.h
```diff
@@ -60,19 +60,20 @@ class GSWndGL : public GSWnd
 {
 protected:
 	bool m_ctx_attached;
-	int m_vsync;
+	std::atomic_bool m_vsync_change_requested;
```
gregory38: Why not use `std::atomic<bool>` ?

plugins/GSdx/GSWndEGL.cpp
```diff
@@ -235,15 +235,19 @@ GSVector4i GSWndEGL::GetClientRect()
...
 {
 	// 0 -> disable vsync
 	// n -> wait n frame
-	eglSwapInterval(m_eglDisplay, vsync);
+	eglSwapInterval(m_eglDisplay, m_vsync);
 }
 
 void GSWndEGL::Flip()
 {
+	if (m_vsync_change_requested) {
```
gregory38: What do you think about using an atomic swap ? I think, it shows better the intended behavior.
`if (m_vsync_change_requested.exchange(false))`

plugins/GSdx/GSWnd.h
ssakash: ``std::atomic_bool`` is just a type definition of ``std::atomic<bool>``. He has probably used this for the same reason on why we're using ``GSVector2i`` over ``GSVector2<int>`` at most cases.

Though I think ``std::atomic<bool>`` might be more readable at this case as not everyone might be aware of the atomic type definitions, not sure though.

turtleli: I don't mind either way. I'll change it.

plugins/GSdx/GSWndEGL.cpp
turtleli: Ok, will do.

pcsx2/MTGS.cpp
```diff
@@ -202,8 +202,7 @@ void SysMtgsThread::OpenPlugin()
 	else
 		result = GSopen( (void*)pDsp, "PCSX2", renderswitch ? 2 : 1 );
 
-
-	GSsetVsync(EmuConfig.GS.GetVsync());
+	GSsetVsync(EmuConfig.GS.FrameLimitEnable ? EmuConfig.GS.GetVsync() : 0);
```
ssakash: I would prefer if the ``FrameLimitEnable`` check is done inside ``EmuConfig.GS.GetVsync()``. (Inside the function, we're already returning 0 for turbo mode, doesn't hurt to also extend it to framelimit)

turtleli:
Updated to address review comments.

plugins/GSdx/GSWnd.h
gregory38: @ssakash yeah I checked google before asking the reason ? I don't know why they add it, maybe for raw C purpose.

pcsx2/Pcsx2Config.cpp
```diff
@@ -237,7 +237,7 @@ void Pcsx2Config::GSOptions::LoadSave( IniInterface& ini )
 
 int Pcsx2Config::GSOptions::GetVsync() const
 {
-	if (g_LimiterMode == Limit_Turbo)
+	if (g_LimiterMode == Limit_Turbo || !FrameLimitEnable)
```
gregory38: Is it still possible to use Vsync without framelimiter ?

mirh: Every time you disable FL, vsync gets nuked now. 
Which is [bad](2053) - albeit certainly at least consistent now.

gregory38: @turtleli you can merge it. And we can see later how to handle framelimiter/vsync case (if there is really an issue).

turtleli: Ok, I'll merge after Travis CI is fixed.

PR #2051, opened 2017-08-23
GSdx-D3D11: Avoid undefined behavior in swapchain

### LINKED ISSUE #2050: GSDX DX11: black screen with vsync set to adaptive
***PCSX2 version:***
Last dev version.

***PCSX2 options:***
Vsync to adaptive and GSDX to DX11 (software or hardware)

***Plugins used:***
All default

GSDX show a black screen under DX11 if vsync is set to adaptive.

Set GSDX to DX11 mode and vsync to adaptive.

Don't exist: the introduction of vsync adaptive introduced the issue

***PC specifications:***
Intel Core 2 Duo E8600 Gigabyte GTX 780Ti  Windows 10 64 Bit

ssakash:
Closes #2050

### DIFF
--- a/plugins/GSdx/GSDevice11.cpp
+++ b/plugins/GSdx/GSDevice11.cpp
@@ -432,6 +432,11 @@ void GSDevice11::SetExclusive(bool isExcl)
 	}
 }
 
+void GSDevice11::SetVSync(int vsync)
+{
+	m_vsync = vsync ? 1 : 0;
+}
+
 void GSDevice11::Flip()
 {
 	m_swapchain->Present(m_vsync, 0);
--- a/plugins/GSdx/GSDevice11.h
+++ b/plugins/GSdx/GSDevice11.h
@@ -170,6 +170,7 @@ public:
 	bool Create(const std::shared_ptr<GSWnd> &wnd);
 	bool Reset(int w, int h);
 	void Flip();
+	void SetVSync(int vsync) final;
 
 	void SetExclusive(bool isExcl);
 


plugins/GSdx/GSDevice11.cpp
```diff
@@ -434,7 +434,12 @@ void GSDevice11::SetExclusive(bool isExcl)
 
 void GSDevice11::Flip()
 {
-	m_swapchain->Present(m_vsync, 0);
+	// PCSX2 sends a negative value (-1) to GSdx when adaptive mode is specified for Vsync
+	// This mode is exclusive to OpenGL at the moment and passing a negative value to the swapchain
+	// present function is undefined as it only accepts unsigned integers as a parameter, so let's
```
turtleli: I think the `Present()` documentation suggests it only accepts 0-4.

Otherwise it seems fine (personally I would have overridden SetVSync to change the m_vsync value).

gregory38:
Is DX9 fine ?

ssakash:
> Is DX9 fine ?

D3D9 is unaffected. It disables Vsync when zero is passed from the core and it enables normal Vsync for non zero values, so in this case -1 will cause D3D9 to have standard Vsync.

plugins/GSdx/GSDevice11.cpp
```diff
@@ -432,9 +433,18 @@ void GSDevice11::SetExclusive(bool isExcl)
 	}
 }
 
+void GSDevice11::SetVSync(int vsync)
+{
+	if (m_vsync == vsync)
+		return;
+
+	__super::SetVSync(vsync);
+	m_vsyncinterval = vsync ? 1 : 0;
```
turtleli: Maybe just `GSDevice::SetVSync(std::abs(vsync));` or even `m_vsync = std::abs(vsync);`?

Also, your commit message is a bit inaccurate since the Present function only accepts 0-4 (I did a quick check).

ssakash: > Also, your commit message is a bit inaccurate since the Present function only accepts 0-4 (I did a quick check).

You mean this part ``"the present function
of swapchain only accepts unsigned integers as parameter"``? I just wrote what's given in [Microsoft documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/bb174576(v=vs.85).aspx).

``` cpp
HRESULT Present(
   UINT SyncInterval,
   UINT Flags
);
```
Also I said ``"only accepts unsigned integers as parameter"`` which implies that values exclusive to signed integers are not accepted. Doesn't necessarily mean that all the unsigned integers are accepted as parameters.

> Maybe just GSDevice::SetVSync(std::abs(vsync)); or even m_vsync = std::abs(vsync);?

``std::abs`` isn't a good idea, if Gregory adds a new -2 mode to the VSync combobox then we'll pass 2 as the Vsync interval for swapchain present function and it won't be easy to find this altered behavior as it's just a delay in presentation instead of something as noticeable as #2050.

I guess what you're actually against is the addition of ``m_vsyncinterval``, I think it's useful for certain reasons but I can certainly remove it if you want it to be removed, don't really feel like arguing about it right now.

turtleli: 0-4 is more specific and is also in the documentation.

I doubt another vsync mode will be added, but I guess you're right on `std::abs`. And yes, I don't get why `m_vsyncinterval` was used instead of `m_vsync`.

mirh: > if Gregory adds a new -2 mode

Mhh.. Why should he? Adaptive is supposed to be -1.

gregory38: -2 is legal in GL as 2. Besides, it could also be interesting to keep some special value for strange mode (such as dwm).

mirh: .-.
Why should we always make this assumption for the user?

PR #2061, opened 2017-08-31
GSdx: Grey out OSD and Shader Configuration

lightningterror:
Grey out OSD Configuration button for Direct3D since only OpenGL supports the feature atm.
Grey out Shader Configuration button in Core Benchmark.

### DIFF
--- a/plugins/GSdx/GSSettingsDlg.cpp
+++ b/plugins/GSdx/GSSettingsDlg.cpp
@@ -399,7 +399,7 @@ void GSSettingsDlg::UpdateControls()
 		ShowWindow(GetDlgItem(m_hWnd, IDC_OPENCL_TEXT), SW_HIDE);
 #endif
 
-		ShowWindow(GetDlgItem(m_hWnd, IDC_LOGZ), dx9? SW_SHOW: SW_HIDE);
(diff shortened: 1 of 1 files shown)


lightningterror:
Curious why the "Shader Configuration" isn't greyed out on Core Benchmark. Is there any particular use for it ? If not then I can update the pr adding that as well.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -443,6 +443,12 @@ void GSSettingsDlg::UpdateControls()
 		// Hacks
 		EnableWindow(GetDlgItem(m_hWnd, IDC_HACKS_ENABLED), hw);
 		EnableWindow(GetDlgItem(m_hWnd, IDC_HACKSBUTTON), hw && IsDlgButtonChecked(m_hWnd, IDC_HACKS_ENABLED));
+
+		// OSD Configuration
+		EnableWindow(GetDlgItem(m_hWnd, IDC_OSDBUTTON), ogl);
+
+		// Shader Configuration
+		EnableWindow(GetDlgItem(m_hWnd, IDC_SHADEBUTTON), hw || sw || ocl);
```
ssakash: `` hw || sw || ocl`` could be simplified to ``!null``

PR #2062, opened 2017-09-01
gsdx: remove "using namespace std;"

turtleli:
Changes:
 * Removes use of `using namespace std;` by prefixing `std::` or using `auto` and similar.
 * Removes hash_{set,map} macros.

### DIFF
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -385,7 +385,7 @@ namespace GLLoader {
 
 		if (glGetStringi && max_ext) {
 			for (GLint i = 0; i < max_ext; i++) {
-				string ext((const char*)glGetStringi(GL_EXTENSIONS, i));
+				std::string ext{(const char*)glGetStringi(GL_EXTENSIONS, i)};
 				// Bonus
 				if (ext.compare("GL_EXT_texture_filter_anisotropic") == 0) found_GL_EXT_texture_filter_anisotropic = true;
 				if (ext.compare("GL_NVX_gpu_memory_info") == 0) found_GL_NVX_gpu_memory_info = true;
--- a/plugins/GSdx/GSRendererHW.h
+++ b/plugins/GSdx/GSRendererHW.h
@@ -96,18 +96,18 @@ private:
 
 		template<class T> class FunctionMap : public GSFunctionMap<uint32, T>
 		{
-			list<HackEntry<T> >& m_tbl;
+			std::list<HackEntry<T> >& m_tbl;
 
 			T GetDefaultFunction(uint32 key)
 			{
 				CRC::Title title = (CRC::Title)(key & 0xffffff);
 				CRC::Region region = (CRC::Region)(key >> 24);
 
-				for(typename list<HackEntry<T> >::iterator i = m_tbl.begin(); i != m_tbl.end(); i++)
+				for(const auto &entry : m_tbl)
 				{
-					if(i->title == title && (i->region == CRC::RegionCount || i->region == region))
+					if(entry.title == title && (entry.region == CRC::RegionCount || entry.region == region))
 					{
-						return i->func;
+						return entry.func;
 					}
 				}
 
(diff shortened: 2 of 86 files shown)


AlessandroVetere:
It seems all good.
Where you use decltype is it to enforce const_iterator instead of iterator? Would auto be enough?

turtleli:
Yes, I used `decltype` instead of `auto` to keep the use of `const_iterator`. `auto` might be enough in some cases, but I didn't want to drop the const.

plugins/GSdx/GSRendererHW.h
```diff
@@ -96,31 +96,31 @@ class GSRendererHW : public GSRenderer
...
-			list<HackEntry<T> >& m_tbl;
+			std::list<HackEntry<T> >& m_tbl;
 
 			T GetDefaultFunction(uint32 key)
 			{
 				CRC::Title title = (CRC::Title)(key & 0xffffff);
 				CRC::Region region = (CRC::Region)(key >> 24);
 
-				for(typename list<HackEntry<T> >::iterator i = m_tbl.begin(); i != m_tbl.end(); i++)
+				for(auto i : m_tbl)
```
turtleli: Could probably make this `const auto &i`.

AlessandroVetere: Yes, it's probably better to access the list by reference given that `HackEntry<T>` is not primitive nor pointer.

AlessandroVetere:
For the `const_iterator` "issue", here there's an alternative `C++17` solution: https://stackoverflow.com/a/45057578/2627697

plugins/GSdx/GSRendererHW.h
```diff
@@ -96,31 +96,31 @@ class GSRendererHW : public GSRenderer
...
-			list<HackEntry<T> >& m_tbl;
+			std::list<HackEntry<T> >& m_tbl;
 
 			T GetDefaultFunction(uint32 key)
 			{
 				CRC::Title title = (CRC::Title)(key & 0xffffff);
 				CRC::Region region = (CRC::Region)(key >> 24);
 
-				for(typename list<HackEntry<T> >::iterator i = m_tbl.begin(); i != m_tbl.end(); i++)
+				for(const auto &i : m_tbl)
```
AlessandroVetere: I know it's being pedant, but maybe renaming `i` to something more related to the list element type is better, like `entry`, to remark that the variable is not an `iterator` but a `HackEntry<T>&`. Here maybe it's useless, but this applies in general where the iterator has been substitued by a entry reference or entry copy in the new foreach loops.

turtleli: I made the change in this case (and one other I think). The rest I left alone since I don't think it would improve readability (think they mostly/all involve map/unordered_map).

turtleli:
Used `std::as_const` instead of decltype - Although it is C++17, the affected code is Windows only and VS2015+ supports it.

plugins/GSdx/GLLoader.cpp
```diff
@@ -385,7 +385,7 @@ namespace GLLoader {
 
 		if (glGetStringi && max_ext) {
 			for (GLint i = 0; i < max_ext; i++) {
-				string ext((const char*)glGetStringi(GL_EXTENSIONS, i));
```
gregory38: What is the advantage to use `{...}` instead of `( ... )` for the constructor ? (I'm just curious)

turtleli: It avoids ambiguity in some cases - what might look like a variable declaration might actually be a function declaration. Some further reading at https://herbsutter.com/2013/05/09/gotw-1-solution/ (or you can look up "most vexing parse")

Though in this case, I felt like changing it, got part of the way through, and then abandoned the effort...

gregory38: Thanks for the info.

gregory38:
@turtleli @AlessandroVetere @ssakash 
Any pending GSdx work on your side ? I will clang-format the plugin this week-end (I hope). Tell me if you have any pending work.

turtleli:
Nope.

Small note: I don't think the VS clang-format plugin that corresponds (roughly) to 3.8 works on VS2017.

AlessandroVetere:
No pending work on my side.

ssakash:
There are [5 active pull requests](https://github.com/PCSX2/pcsx2/pulls?q=is%3Aopen+is%3Apr+label%3A%22Plugin%3A+GSdx%22) at the moment involving GSdx, out of which, three of them are mine. I don't mind if you break #1795 but I'd prefer if conflict is avoided on my other two PRs. (_I'll try to wrap them up when I get free time_)

ssakash:
@turtleli 

Orphis buildbot doesn't seem to be generating any builds after the merge of your pull request. Not sure if it's a problem with the buildbot or  something related to usage of ``std::as_const``.

turtleli:
I've no idea why the latest commit hasn't been built on the Orphis buildbot. If it's not a problem with the buildbot then someone will have to tell me what's failing.

gregory38:
Yes it is `as_const`. I'm asking orphis to upgrade the bot.

PR #2064, opened 2017-09-05
GSdx-ogl: Fix GSdx crashing on intel igpus

lightningterror:
Regression was introduced in #1954
GSdx caused the emulator to crash when the renderer was restarted on OpenGL.

edit: It may have affected older gpus from nvidia and amd with older OpenGL support but I'm not sure.

### DIFF
--- a/plugins/GSdx/GLLoader.cpp
+++ b/plugins/GSdx/GLLoader.cpp
@@ -454,17 +454,19 @@ namespace GLLoader {
 					"Check out the link below for further information.\n"
 					"https://github.com/PCSX2/pcsx2/wiki/OpenGL-and-Intel-GPUs-All-you-need-to-know\n");
 			}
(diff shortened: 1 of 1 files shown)


mirh:
.... I wonder if my testings might not have had something to do with this..
Well, whatever.

lightningterror:
Sorry , didn't want to introduce a major regression.

mirh:
Putting aside that even if driver officially supported something, it doesn't mean practically it would be working fine.. Now that you mention that, no I wasn't restarting anything. Cold boots every time. 
So, nothing to see here. Thanks.

lightningterror:
I did encounter crashes testing my old geforce 9 gpu just for lolz with the llvmpipe driver by just starting pcsx2.
With the fix from this PR it didn't crash.

PR #2067, opened 2017-09-09
Add PSX memory card support

RedPanda4552:
The majority of code for PSX memory cards existed with a few exceptions:
- PSX memory card creation was not possible through the emulator
- Games could not see a PSX memory card was inserted

This PR addresses these two issues by doing the following:
- Adds a checkbox to the memory card creation screen allowing PSX cards to be made, as well as the logic for writing the empty card. There is no regard for custom sizes or folder format cards, only the standard 128 KiB size, and file based cards.
- Modifying the existing preprocessor based SIO interrupts to instead work based on the running game's type (PSX/PS2). This allows PS2 games to skip the delays as they were before, but for PSX games to use the delays, which are necessary for memory card interaction.

Edit: I should apologize in advance for the eyesore that is the commit history, learned about the evil that is merging from remote, and had issues that involved forking multiple times and finessing Git locally.

### DIFF
--- a/pcsx2/R3000A.cpp
+++ b/pcsx2/R3000A.cpp
@@ -172,9 +172,8 @@ static __fi void _psxTestInterrupts()
 	IopTestEvent(IopEvt_SIF0,		sif0Interrupt);	// SIF0
 	IopTestEvent(IopEvt_SIF1,		sif1Interrupt);	// SIF1
 	IopTestEvent(IopEvt_SIF2,		sif2Interrupt);	// SIF2
-#ifndef SIO_INLINE_IRQS
-	IopTestEvent(IopEvt_SIO,		sioInterrupt);
-#endif
+	// Originally controlled by a preprocessor define, now PSX dependent.
+	if (psxHu32(HW_ICFG) & (1 << 3)) IopTestEvent(IopEvt_SIO, sioInterruptR);
 	IopTestEvent(IopEvt_CdvdRead,	cdvdReadInterrupt);
 
 	// Profile-guided Optimization (sorta)
--- a/pcsx2/Sio.cpp
+++ b/pcsx2/Sio.cpp
@@ -26,6 +26,7 @@ _mcd *mcd;
 SIO_MODE siomode = SIO_START;
 static void sioWrite8inl(u8 data);
 #define SIO_WRITE void inline
+#define SIO_FORCEINLINE __fi
 
 // Magic psx values from nocash info
 static const u8 memcard_psx[] = {0x5A, 0x5D, 0x5C, 0x5D, 0x04, 0x00, 0x00, 0x80};
@@ -79,21 +80,6 @@ void ClearMcdEjectTimeoutNow()
 	}
 }
 
-// SIO Inline'd IRQs : Calls the SIO interrupt handlers directly instead of
-// feeding them through the IOP's branch test. (see SIO.H for details)
-
-#ifdef SIO_INLINE_IRQS
-#define SIO_INT() sioInterrupt()
-#define SIO_FORCEINLINE __fi
-#else
-__fi void SIO_INT()
-{
-	if( !(psxRegs.interrupt & (1<<IopEvt_SIO)) )
-		PSX_INT(IopEvt_SIO, 64 ); // PSXCLK/250000);
-}
-#define SIO_FORCEINLINE __fi
-#endif
-
 // Currently only check if pad wants mtap to be active.
 // Could lets PCSX2 have its own options, if anyone ever
 // wants to add support for using the extra memcard slots.
@@ -132,12 +118,32 @@ void sioInit()
 	sio.packetsize = 0;
 }
 
-void SIO_FORCEINLINE sioInterrupt()
-{
-	PAD_LOG("Sio Interrupt");
-	sio.StatReg|= IRQ;
-	iopIntcIrq(7); //Should this be used instead of the one below?
-	//psxHu32(0x1070)|=0x80;
+bool isR3000ATest = false;
+
+// Check the active game's type, and fire the matching interrupt.
+// The 3rd bit of the HW_IFCG register lets us know if PSX mode is active. 1 = PSX, 0 = PS2
+// Note that the R3000A's call to interrupts only calls the PS2 based (lack of) delays.
+SIO_FORCEINLINE void sioInterrupt() {
+	if ((psxHu32(HW_ICFG) & (1 << 3)) && !isR3000ATest) {
+		if (!(psxRegs.interrupt & (1 << IopEvt_SIO)))
+			PSX_INT(IopEvt_SIO, 64); // PSXCLK/250000);
+	} else {
+		PAD_LOG("Sio Interrupt");
+		sio.StatReg |= IRQ;
+		iopIntcIrq(7); //Should this be used instead of the one below?
+		//psxHu32(0x1070)|=0x80;
+	}
+
+	isR3000ATest = false;
+}
+
+// An offhand way for the R3000A to access the sioInterrupt function.
+// Following the design of the old preprocessor system, the R3000A should
+// never call the PSX delays (oddly enough), and only the PS2. So we need
+// an extra layer here to help control that.
+__fi void sioInterruptR() {
+	isR3000ATest = true;
+	sioInterrupt();
 }
 
 SIO_WRITE sioWriteStart(u8 data)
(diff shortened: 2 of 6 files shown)


lightningterror:
My eyes ... xD
Anyway never use the Git gui for  updating your fork , use Git shell.
Here is a guide how to do so.
https://gist.github.com/CristinaSolana/1885435
**DON'T DO THIS ATM**! Because you made changes to the master branch and not a separate branch created from master.

Next you should never make the changes on your master but branches created from master so that way you can always have pcsx2 master and your master the same without updates and be able to create separate branches that are up do date with no changes from start.

You need to squash the commits in to 1
You can do that when you have a branch selected open up git shell and type
git rebase -i master     (this will open up all the commits on the branch ) but since you're working on the master branch you'd go with) git rebase -i HEAD~19 
A doc will popup , all the instructions are there followed by a second doc popup.
After you're done you will see the commits merge in to 1.
Once all of that is done you type in  git push -f     to push the changes on github

I advice you to make a copy from the entire code changes before doing this just in case.

RedPanda4552:
The rebase is claiming merge conflicts... How critical is it I squash this?

lightningterror:
It keeps git history clean . otherwise it will get quite messy. It will also show up on the buildbot website and pcsx2.net

Anyway I was able to rebase the last 5 so you can start with that but first you need to abort the previous rebase
git rebase --abort
git rebase -i HEAD~5

lightningterror:
You can also drop 4b1402a/3088a48 , b2877d2 / 7622e58 since you reverted the changes.

lightningterror:
My guess is the conflict was caused from the reverts. It's all good now. :+1:

RedPanda4552:
Alright that was a final rebase there. Left out the memory card creation bit which is kind of important.

pcsx2/R3000A.cpp
```diff
@@ -172,9 +172,10 @@ static __fi void _psxTestInterrupts()
 	IopTestEvent(IopEvt_SIF0,		sif0Interrupt);	// SIF0
 	IopTestEvent(IopEvt_SIF1,		sif1Interrupt);	// SIF1
 	IopTestEvent(IopEvt_SIF2,		sif2Interrupt);	// SIF2
-#ifndef SIO_INLINE_IRQS
-	IopTestEvent(IopEvt_SIO,		sioInterrupt);
-#endif
+	// Originally controlled by a preprocessor define, now PSX dependent.
+	if (psxMode) {
+		IopTestEvent(IopEvt_SIO, sioInterruptR);
```
ssakash: All the conditional checks at this file use this format:

```
if(condition)
{
}
```

not 

```
if(condition) {
}
```

I'd prefer to keep the style which is more common. Though I think it would be better if braces are not used for single line statements.

pcsx2/R5900.cpp
```diff
@@ -548,6 +549,7 @@ void __fastcall eeloadHook()
 
 	wxString discelf;
 	int disctype = GetPS2ElfName(discelf);
+	psxMode = disctype == 1 ? true : false;
```
ssakash: ``? true : false`` is not necessary here. (Coverity complains about such redundant operations)

pcsx2/Sio.cpp
```diff
@@ -132,12 +119,31 @@ void sioInit()
...
-{
-	PAD_LOG("Sio Interrupt");
-	sio.StatReg|= IRQ;
-	iopIntcIrq(7); //Should this be used instead of the one below?
-	//psxHu32(0x1070)|=0x80;
+bool isR3000ATest = false;
+
+// Check the active game's type, and fire the matching interrupt.
+// Note that the R3000A's call to interrupts only calls the PS2 based (lack of) delays.
+__fi void sioInterrupt() {
```
ssakash: Well why not use ``SIO_FORCEINLINE``? If its uses are limted, best to remove that macro in my opinion. I'm suggesting to use it as I assume there might be certain cases where it would take different values based on other ``ifdef`` checks.

pcsx2/R5900.cpp
```diff
@@ -548,6 +549,7 @@ void __fastcall eeloadHook()
 
 	wxString discelf;
 	int disctype = GetPS2ElfName(discelf);
+	psxMode = disctype == 1;
```
ramapcsx2: This isn't the most elegant PSX mode detector but for now it's good enough. (It might misdetect modes when starting a homebrew .elf with a PSX disc mounted.)

mirh: *TODO: elgant-ify detection*?

RedPanda4552: Do you guys have any thoughts as to a better place to hook in? So far this is the best thing I've found; `GetPS2ElfName` returns 2 for a PS2 disc, 1 for a PSX disc, 0 for all others. `mcd->IsPSX()` isn't any good because PS2 games still probe PSX cards.

mirh: If rama said that's good, I have nothing else to add. 
If it's still not 'perfect' though, a note just for the records would be cool.

lightningterror:
@AdmiralCurtiss can you also take a look at the code ?

pcsx2/R5900.cpp
ramapcsx2: There's some hardware register that indicates PSX mode with a flag. Once set, it can never be unset, unless doing a hardware reset. It's somewhere in my mess of reversing notes :p

RedPanda4552:
I found `HW_ICFG = 0x1f801450` in `IopHw.cpp`. Searching for `0xBF801450` yields no results in the project. Dumping `psxHu32(HW_ICFG)` to console outputs 1 for PS2 games, and 8 for PSX games. Change that to binary and we get 0001 for PS2, 1000 for PSX. So when you said 3rd bit, if you were starting with the rightmost bit being 0th bit, then that's spot on. Though it is mildly interesting the 0th bit is changing as well from 1 to 0.

gregory38:
0xA00....0xBFF address space is a kernel segment of uncached memory. Bottom part of the segment is the RAM, upper part is the ROM and middle the register. The segment are 512MB.

pcsx2/Sio.cpp
```diff
@@ -122,9 +121,12 @@ void sioInit()
 bool isR3000ATest = false;
 
 // Check the active game's type, and fire the matching interrupt.
+// The HW_IFCG register lets us know if PSX mode is active.
+// The "3rd bit" is supposed to be the flag, but the actual value evaluates to 8 for PSX.
+// Conversely, the value is 1 for PS2. 
 // Note that the R3000A's call to interrupts only calls the PS2 based (lack of) delays.
 SIO_FORCEINLINE void sioInterrupt() {
-	if (psxMode && !isR3000ATest) {
+	if (psxHu32(HW_ICFG) == 8 && !isR3000ATest) {
```
ramapcsx2: // check if bit 3 is set
if ((psxHu32(HW_ICFG) & (1<<3)) && !isR3000ATest) {

ramapcsx2: This way we don't care about the other bits, so the register can be something other than evaluating exactly to 8.
Nice that you found it though! :)

ramapcsx2:
I'm fine with merging. What's everyone else's opinion?

MrCK1:
I haven't tried it, but the code ooks fine to me.

ramapcsx2:
I've tested it briefly:
- It never triggers the PSX mode stuff in PS2 games.
- It works for PSX titles. Compatibility is up, but a few games (mostly Squaresoft titles) don't work with it. They didn't work before either, so that's fine.
- Card creation works fine. The BIOS can recognize and format a new PSX card and games accept it as well.

lightningterror:
> It never triggers the PSX mode stuff in PS2 games.

Did you also test behavior with compressed and uncompressed memory cards ?

RedPanda4552:
Tested with and without NTFS compression, memory card creation, formatting, saving and loading all work without issue on Digimon World 2 and 3.

Some things that may be arbitrary but I still want to point out:
- I'm not sure at what level NTFS compression operates, but I did not touch it, and cannot confirm or deny if the setting is having a tangible effect on the PSX cards because...
- The PSX card sizes on my file system are both 128 KiB, and the save files within are identically sized for both games as well.

mirh:
NTFS compression should be totally 100% transparent to applications. 

And its size, and files inside one, being always like fixed is normal (and I guess makes for the enabling of compression in the first place)

ramapcsx2:
Nice work, thanks! :)

Issue #1637, opened 2016-10-22
GSDumpGUI issues


mirh:
- [ ] Crash on start when another instance is already open
- [ ] You have to click two times start/configure buttons before anything happens

refractionpcsx2:
It doesn't support opengl,  that's an issue :p

FlatOutPS2:
> It doesn't support opengl, that's an issue :p

Huh?

refractionpcsx2:
It doesn't support any renderer except dx

mirh:
What? 
It does man. 
I mean, you have a point about renderer override options if any. 

But you have configure button for a reason?

refractionpcsx2:
Really? It must have been updated since I last used it then as it didn't support opengl then

refractionpcsx2:
Really? It must have been updated since I last used it then as it didn't support opengl then

FlatOutPS2:
> Really? It must have been updated since I last used it then as it didn't support opengl then

That must've been more than a year ago then. :p

refractionpcsx2:
It was :p  and apologies for the double post, github timed out, I'll sort that shortly

FlatOutPS2:
I hope it was just a github fluke, and not a sign of another DDOS attack...

mirh:
There [has been](https://github.com/PCSX2/pcsx2/commits/master/tools/GSDumpGUI) no relevant change in the last half decade.

turtleli:
#656 fixed OpenGL.

refractionpcsx2:
> I hope it was just a github fluke, and not a sign of another DDOS attack...

I think it was a mixture of github and my phone, didn't want to play ball :P

MrCK1:
@mirh IIRC, GSDumpGUI doesn't always terminate after the GUI window is closed - similar to the issue we solved with the PCSX2 executable

ssakash:
> Renderer Override could be better structured with dx9/dx11/ogl radio buttons and a SW (or hw) checkbox

I vote for removing it, seems rather useless in my opinion.

mirh:
> IIRC, GSDumpGUI doesn't always terminate after the GUI window is closed - similar to the issue we solved with the PCSX2 executable

Afaik that happens due to having to press 2 times start button. 
That in turn fires two children process (a "dead" one and an actually working one) and only the later is terminated in the end. 

> I vote for removing it, seems rather useless in my opinion.

[...]

PR #2078, opened 2017-09-25
GSDumpGUI: Update the GUI interface

lightningterror:
Rearranged/updated many  GUI elements.
Reduce height of window to make it more appealing for laptop users.
Fix Render Overrides (Render overrides will now work properly for each render and will not conflict when GSdx settings are changed).
Add OpenGL HW/SW Options to Render Override.

Here's the updated version how it looks.

image

And the program itself for anyone wanting to test it.
[GSDumpGUI.zip](https://github.com/PCSX2/pcsx2/files/1331110/GSDumpGUI.zip)

### DIFF
--- a/tools/GSDumpGUI/Forms/frmMain.cs
+++ b/tools/GSDumpGUI/Forms/frmMain.cs
@@ -64,7 +64,7 @@ namespace GSDumpGUI
 
         public void ReloadGSDXs()
         {
-            txtIntLog.Text += "Starting GSDX Loading Procedures" + Environment.NewLine + Environment.NewLine;
+            txtIntLog.Text += "Starting GSdx Loading Procedures" + Environment.NewLine + Environment.NewLine;
 
             txtGSDXDirectory.Text = Properties.Settings.Default.GSDXDir;
             txtDumpsDirectory.Text = Properties.Settings.Default.DumpDir;
@@ -90,14 +90,14 @@ namespace GSDumpGUI
                     }
                     catch (InvalidGSPlugin)
                     {
-                        txtIntLog.Text += "Failed to load \"" + itm + "\". Is it really a GSDX DLL?" + Environment.NewLine;
+                        txtIntLog.Text += "Failed to load \"" + itm + "\". Is it really a GSdx DLL?" + Environment.NewLine;
                     }
                 }
             }
 
-            txtIntLog.Text += Environment.NewLine + "Completed GSDX Loading Procedures" + Environment.NewLine + Environment.NewLine;
+            txtIntLog.Text += Environment.NewLine + "Completed GSdx Loading Procedures" + Environment.NewLine + Environment.NewLine;
 
-            txtIntLog.Text += "Starting GSDX Dumps Loading Procedures : " + Environment.NewLine + Environment.NewLine;
+            txtIntLog.Text += "Starting GSdx Dumps Loading Procedures : " + Environment.NewLine + Environment.NewLine;
             if (Directory.Exists(txtDumpsDirectory.Text))
             {
                 String[] Dumps = Directory.GetFiles(txtDumpsDirectory.Text, "*.gs", SearchOption.TopDirectoryOnly);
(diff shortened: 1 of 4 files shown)


tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -199,12 +211,12 @@ private void ExecuteFunction(String Function)
                     {
                         String newini = ini.Substring(0, pos + 9);
                         newini += SelectedRenderer;
-                        newini += ini.Substring(pos + 10, ini.Length - pos - 10);
+                        newini += ini.Substring(pos + 11, ini.Length - pos - 11);
```
turtleli: I don't think padding the renderer number with a space so that it's always 2 characters is a good idea. Maybe a better idea would be to look for the following newline and cut everythng out between `Renderer=` and the newline.

Also, I don't think the second `Substring` argument is necessary in this case. Without the second argument, the substring ends at the end of the original string anyway.

Side note: I'm not familiar with C#.

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -199,12 +211,12 @@ private void ExecuteFunction(String Function)
...
                         File.WriteAllText(AppDomain.CurrentDomain.BaseDirectory + "GSDumpGSDXConfigs\\" + GSDXName + "\\inis\\gsdx.ini", newini);
                     }
                     else
                     {
-                        File.WriteAllText(AppDomain.CurrentDomain.BaseDirectory + "GSDumpGSDXConfigs\\" + GSDXName + "\\inis\\gsdx.ini", ini + Environment.NewLine + "Renderer=" + SelectedRenderer);
+                        File.WriteAllText(AppDomain.CurrentDomain.BaseDirectory + "GSDumpGSDXConfigs\\" + GSDXName + "\\inis\\gsdx.ini", ini + "[Settings]" + Environment.NewLine + "Adapter=default" + Environment.NewLine + "Renderer=" + SelectedRenderer + Environment.NewLine);
```
turtleli: The `Adapter=default` part is probably unnecessary.

lightningterror: It isn't. Added it just in case. [Settings] is tho , as without it it will conflict when changing Gsdx settings.

tools/GSDumpGUI/Forms/frmMain.cs
lightningterror: I'll leave the padding as it is since it's working as it is unless someone else wants to change it , plus also not that familiar with C# as well :laughing: 

As for the second argument on the Substring I removed it and it's working properly.

lightningterror:
Made a few more adjustments.

Added N/A to packet sizes when they are inactive before running a gs dump.
image

turtleli:
The renderer number padding doesn't work well in some cases. Test case, starting with no ini:
1. No renderer override -> Configure GSdx -> Select D3D9/11 HW/SW renderer -> Click Ok.
2. Select a renderer override (i.e. OGL HW) -> Configure GSdx -> Click Cancel.
3. Open GSdx.ini -> The renderer line will look like `Renderer=12interlace=7`.

So it would be better to use the approach I suggested. It should be fairly simple if you look at the surrounding code (IndexOf, Environment.NewLine) and also look at the corresponding documentation.

Side note: Maybe add `[skip ci]` to the commit message when making further updates to this PR. Neither AppVeyor nor Travis CI will build the GSDumpGUI.

MrCK1:
This also addresses the issues mentioned in #1687

lightningterror:
There is some space/row when moving the cursor between the letters but it didn't show up properly.
Adding + "" seems to work, removing the substring with 11 causes the rest of the config after it to be removed/reset when changing overrides.

turtleli:
Any progress? It should be quite simple.

lightningterror:
So far I'm stuck.

turtleli:
Well, I think you need something along the lines of
```
int newlinepos = ini.IndexOf(Environment.NewLine, pos);
if (newlinepos != -1)
{
   // handle newline found scenario
}
else
{
    // handle newline not found scenario
}

```

ramapcsx2:
@turtleli Do you want to try finishing this PR?

turtleli:
I can do so if necessary. TBH I could have provided the code ages ago, but I didn't because I thought it would help lightningterror's coding improve.

@lightningterror Do you wish to complete this yourself or would you rather I handle it?

lightningterror:
> TBH I could have provided the code ages ago, but I didn't because I thought it would help lightningterror's coding improve.

Yeah I know :stuck_out_tongue: 
Anyway you can handle the rest.

turtleli:
PR updated.

FWIW I was looking for something like:
```
int newlinepos = ini.IndexOf(Environment.NewLine, pos);
if (newlinepos != -1)
{
   newini += ini.Substring(newlinepos);
}
else
{
    newini += Environment.NewLine;
}
```
though I think I've replaced it with something that's simpler and more robust (see the relevant commit).

lightningterror:
Works like a charm now, btw do you also wanna look at some of the issues mentioned in #1637 since we are updating the code atm or leave it for another time ?

turtleli:
> btw do you also wanna look at

Nope! ;)

I'm not really interested in the GSDumpGUI.

PR #2080, opened 2017-09-28
PCSX2-WX: Add clear list option to ISO Selector

### LINKED ISSUE #2071: Feature Request: Clear ISO list cache
Would be nice to have a neat feature like this since not many people wanna bother clearing the iso cache or they don't know how to do it.

iso clear

Something like this for example. @ssakash  what do you think ?

RedPanda4552:
Adds "Clear ISO List" option to ISO Selector (Issue #2071).

A look at the UI before clearing, with all my iso and bin files: https://i.imgur.com/YnPCAXQ.png
I was kind enough to give users an opportunity to change their mind: https://i.imgur.com/VUVo5l0.png
And after clearning: https://i.imgur.com/Wvsdzfn.png

Closes #2071

### DIFF
--- a/pcsx2/gui/MainMenuClicks.cpp
+++ b/pcsx2/gui/MainMenuClicks.cpp
@@ -412,6 +412,23 @@ void MainEmuFrame::Menu_IsoBrowse_Click( wxCommandEvent &event )
 	AppSaveSettings();		// save the new iso selection; update menus!
 }
 
+void MainEmuFrame::Menu_IsoClear_Click(wxCommandEvent &event)
+{
+	wxDialogWithHelpers dialog(this, _("Confirm clearing ISO list"));
+	dialog += dialog.Heading(_("This will clear the ISO list. If an ISO is running it will remain in the list. Continue?"));
+
+	bool confirmed = pxIssueConfirmation(dialog, MsgButtons().YesNo()) == wxID_YES;
+
+	if (confirmed)
+	{
+		// If the CDVD mode is not ISO, or the system isn't running, wipe the CurrentIso field in INI file
+		if (g_Conf->CdvdSource != CDVD_SourceType::Iso || !SysHasValidState())
+			SysUpdateIsoSrcFile("");
+		wxGetApp().GetRecentIsoManager().Clear();
+		AppSaveSettings();
+	}
+}
+
 void MainEmuFrame::Menu_Ask_On_Boot_Click(wxCommandEvent &event)
 {
 	g_Conf->AskOnBoot = event.IsChecked();
--- a/pcsx2/gui/RecentIsoList.cpp
+++ b/pcsx2/gui/RecentIsoList.cpp
@@ -14,6 +14,7 @@
  */
 
 #include "PrecompiledHeader.h"
+#include "App.h"
 #include "AppCoreThread.h"
 #include "RecentIsoList.h"
 #include "IsoDropTarget.h"
@@ -34,7 +35,9 @@ RecentIsoManager::RecentIsoManager( wxMenu* menu, int firstIdForMenuItems_or_wxI
 	, m_firstIdForMenuItems_or_wxID_ANY ( firstIdForMenuItems_or_wxID_ANY )
 {
 	m_cursel	= 0;
-	m_Separator	= NULL;
+	m_Separator	= nullptr;
+	m_ClearSeparator = nullptr;
+	m_Clear = nullptr;
 
 	IniLoader loader;
 	LoadListFrom(loader);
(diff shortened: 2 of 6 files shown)


lightningterror:
Looks like you beat @ssakash to it :laughing: 

Tested. Works properly.

ssakash:
I'm wondering if it might be a good idea to remove/gray out the menu item when there are no ISOs present in the list. Any opinion? (Personally I'd prefer destroying and appending the item over graying it out using ``Enable()``)

Turtleli was a bit faster than me on adding that ``Changes translatable strings`` label. :/

lightningterror:
I like your suggestion. If there are no isos then the button should disappear. Having it just grey out will make the list bigger for no reason when it's not needed.

avih:
The original intent with the grayed out items was that you could put your ISOs on an external drive and plug it in or out, and you'll know when they're available. They should get out of the list anyway when you open new ISOs. But feel free to remove it.

RedPanda4552:
Latest commit makes the list item and separator generate in time with the ISO list itself instead of the entire sub menu. 

One thing I would like to draw attention to is that doing so required `MainFrame.h` to be included in `RecentIsoList.cpp`, for the `MenuIdentifiers` enum to be visible. It feels dirty to include an entire header file for one enum. Is there a better way to go about this or is this an acceptable practice?

MrCK1:
#RedPanda4552 The recent ISO list is considered part of the mainframe. That's what all the Wx menu systems are based off of. Basically, it's necessary for compilation.

RedPanda4552:
That helps but also raises another question. Is there a known reason `RecentIsoList.cpp` **didn't** include `MainFrame.h` before these changes?

turtleli:
I think you need to include App.h instead of MainFrame.h since that's where the enum actually is.

Anyway, the current behaviour isn't that nice if you clear the ISO list while a game is running. Test case:
1. Boot game.
2. Clear ISO list.
3. Pause.
4. Resume -> Cannot resume.

RedPanda4552:
Latest commit should resolve. The CurrentIso field in the ini was being wiped regardless of running state, which when the CDVD plugin is pausing and resuming, would understandably cause problems.

lightningterror:
I think you can squash it if it's ready for merge.

RedPanda4552:
And squashed.

pcsx2/gui/MainMenuClicks.cpp
```diff
@@ -412,6 +412,26 @@ void MainEmuFrame::Menu_IsoBrowse_Click( wxCommandEvent &event )
 	AppSaveSettings();		// save the new iso selection; update menus!
 }
 
+void MainEmuFrame::Menu_IsoClear_Click(wxCommandEvent &event)
+{
+	wxDialogWithHelpers dialog(this, _("Confirm clearing ISO list"));
+	dialog += dialog.Heading(_("This will clear your entire ISO list. Continue?"));
+
+	if (SysHasValidState())
+		dialog += dialog.Heading(_("(Your running ISO will remain in the list)"));
```
turtleli: Hmm. Instead of conditionally appending this message, maybe it could be combined with the above message and always be shown e.g. 
```
This will clear the ISO list. If an ISO is running it will remain in the list. Continue?
```
It might be better for translators this way (more flexibility), but I dunno.

pcsx2/gui/MainMenuClicks.cpp
```diff
@@ -412,6 +412,26 @@ void MainEmuFrame::Menu_IsoBrowse_Click( wxCommandEvent &event )
...
+
+	bool confirmed = (pxIssueConfirmation(dialog, MsgButtons().Yes().Cancel()) != wxID_CANCEL);
+
+	if (confirmed)
+	{
+		if (!SysHasValidState())
```
turtleli: If you boot with the `Plugin` or `No disc` option selected then there'll still be one entry left in the recent ISO list even though an ISO isn't running.

RedPanda4552:
Review changes made and squashed.

pcsx2/gui/RecentIsoList.cpp
```diff
@@ -99,6 +102,18 @@ void RecentIsoManager::RemoveAllFromMenu()
 		m_Menu->Destroy( m_Separator );
 		m_Separator = NULL;
 	}
+
+	if ( m_ClearSeparator != NULL )
```
turtleli: `nullptr` can also be used here.

turtleli:
Code seems fine to me.

The buttons being Yes+Cancel feels slightly odd (I generally expect Yes+No or Ok+Cancel) but maybe that's just me.

pcsx2/gui/MainMenuClicks.cpp
```diff
@@ -412,6 +412,23 @@ void MainEmuFrame::Menu_IsoBrowse_Click( wxCommandEvent &event )
 	AppSaveSettings();		// save the new iso selection; update menus!
 }
 
+void MainEmuFrame::Menu_IsoClear_Click(wxCommandEvent &event)
+{
+	wxDialogWithHelpers dialog(this, _("Confirm clearing ISO list"));
+	dialog += dialog.Heading(_("This will clear the ISO list. If an ISO is running it will remain in the list. Continue?"));
+
+	bool confirmed = pxIssueConfirmation(dialog, MsgButtons().YesNo()) != wxID_CANCEL;
```
turtleli: Should probably be `== wxID_YES`? Otherwise the no button will also wipe the ISO list.

Issue #569, opened 2015-06-01
FIFA Street 2 not rendering correctly.


Shackkill:
Issue title: Graphical bug / Rendering not working well

-PCSX2 version: 1.2.1

-PCSX2 options: Defaults, plugin settings: http://gyazo.com/5d536816216460396c475f1e73248846

-Plugins used: GSd10 : http://gyazo.com/5d536816216460396c475f1e73248846

-Plugin settings: http://gyazo.com/5d536816216460396c475f1e73248846

-Description: Graphics not rendered correctly, like the terrain, the flags in the menu and the players and their clothing.

-How to replicate: Play Fifa Street 2

-Last known version to work: Never worked

-PC specifications: CPU: Intel Core i7 4770 @ 4.3Ghz GPU: Gigabyte Geforce GTX660 2 GB OS: Windows 8 64-bit

karasuhebi:
Relevant:
http://www.reddit.com/r/PCSX2/comments/37yhiy/fifa_street_2_graphics_clothesflags_are_bugged/

bositman:
You didn't follow the guidelines so this report is more or less useless.
https://github.com/PCSX2/pcsx2/wiki/How-to-create-useful-and-valid-issues

bositman:
Eh just noticed you made a wiki page of it...? I meant reading that and editing this issue with those guidelines.
Anyway, did it for you

gregory38:
Try the version from here http://forums.pcsx2.net/Thread-GSdx-2-0
With the following setup
```
accurate_fbmask = 1
```
And use normal filtering: i.e.
```
filter = 2
```

Normally it must be fixed.

Shackkill:
Where do I have to put the downloaded file? And where do I change the settings of fbmask and filter

Date: Tue, 30 Jun 2015 02:34:09 -0700

gregory38:
In the plugins directory. You need to edit GSdx.ini file.

Shackkill:
I can't find a GSdx.ini file...

Date: Tue, 30 Jun 2015 10:00:06 -0700

gregory38:
the ini file is in the ini directory. The new plugin must be copied into the plugin dir.

Shackkill:
Where can I find all this?

Date: Tue, 30 Jun 2015 10:11:30 -0700

gregory38:
In your PCSX2 installation. If it is too complicated for you, you will have to wait a newest release.

[...]

PR #2086, opened 2017-10-05
GSdx: Add CPU Frame buffer Conversion Hack

lightningterror:
Add HwW Hack that enables 4-bit and 8-bit Frame buffer Conversion on the CPU.
Can fix broken textures on games but at the cost of slower performance.
List of games: Harry Potter games, Fifa Street games, MGS3 games.
Games like Call of Duty, Kung fu Panda might also be affected as well as others especially on Direct3D.
Add HW Hack GUI option on Windows/Linux for Framebuffer Convertion hack named "CPU FB Conversion."

Should fix
https://forums.pcsx2.net/Thread-Harry-Potter-and-The-Chamber-of-Secrets-SLES-51192-E-artifacting

Should Improve
https://github.com/PCSX2/pcsx2/issues/569


plugins/GSdx/GSSetting.cpp
```diff
@@ -122,6 +122,13 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_TC_DEPTH:
 			return "Disable the support of Depth buffer in the texture cache.\n"
 				"It can help to increase speed but it will likely create various glitches.";
+		case IDC_8BITCPUFBHACK:
```
MrCK1: A better label would be `IDC_8BIT_CPU_FBHACK`

It's a bit of an eyesore in its' current state.

plugins/GSdx/GSSetting.h
```diff
@@ -63,6 +63,7 @@ enum {
 	IDC_ACCURATE_BLEND_UNIT,
 	IDC_ACCURATE_DATE,
 	IDC_TC_DEPTH,
+	IDC_8BITCPUFBHACK,
```
MrCK1: Ditto.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -698,6 +698,7 @@ void GSHacksDlg::OnInit()
 	CheckDlgButton(m_hWnd, IDC_PRELOAD_GS, theApp.GetConfigB("preload_frame_with_gs_data"));
 	CheckDlgButton(m_hWnd, IDC_ALIGN_SPRITE, theApp.GetConfigB("UserHacks_align_sprite_X"));
 	CheckDlgButton(m_hWnd, IDC_TC_DEPTH, theApp.GetConfigB("UserHacks_DisableDepthSupport"));
+	CheckDlgButton(m_hWnd, IDC_8BITCPUFBHACK, theApp.GetConfigB("UserHack_cpu_8bis_fb_conversion"));
```
MrCK1: Ditto. 

Also, the label should read `"UserHacks_CPU_8Bit_FB_Conversion"` or something similar to match the common format here.

plugins/GSdx/GSdx.rc
```diff
@@ -142,6 +142,7 @@ BEGIN
     CONTROL         "Alpha",IDC_ALPHAHACK,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,14,71,34,8
     CONTROL         "Alpha Stencil",IDC_ALPHASTENCIL,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,108,71,57,8
     CONTROL         "Disable Depth Emulation",IDC_TC_DEPTH,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,108,84,92,8
+    CONTROL         "CPU FB Conversion", IDC_8BITCPUFBHACK, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 14, 84, 91, 8
```
MrCK1: The label needs to include 8-bits somewhere. `"CPU FB Conversion"` isn't really clear on the purpose for the end-user.

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -268,25 +270,17 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
...
-							Read(t, t->m_valid);
-						else
-							dst = t;
-					} else {
+					// It is a complex to convert the code in shader. As a reference, let's do it on the CPU, it will be slow but
+					// 1/ it just works :)
+					// 2/ even with upscaling
+					// 3/ for both DX and OpenGL
+					if (m_cpu_8bit_fb_coversion && psm == PSM_PSMT8)
+						// Hack can fix broken textures on games such as: Harry Potter games (Direct3D and OpenGL), Fifa Street games (Direct3D only), MGS3 (Direct3D only).
```
gregory38: It would be nice to split the comment in 2 parts.
* Dx doesn't have a shader conversion so you must do it but it cost perf
* OpenGL has a shader but it doesn't handle some corner case

This way, comments can be easily updated after the glsl/hlsl port.

gregory38:
[HLSL <=> GLSL](https://anteru.net/blog/2016/mapping-between-hlsl-and-glsl/index.html)

```
vec2 => float2
uvec2 (vector of 2 unsigned int) => uint2 
ivec2 (vector 2 signed int) => int2
gl_FragCoord => SV_Position (I hope you won't have issue with the Y axis order)
texelFetch => Load
```
Copy/past GLSL code, replace the above. Update header/return based on other HLSL equivalent (ps_main0). 

Oh I just saw that you need also the ScalingFactor constant buffer parameter. My code is awful without comment...

plugins/GSdx/GSSetting.cpp
```diff
@@ -122,6 +122,13 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_TC_DEPTH:
 			return "Disable the support of Depth buffer in the texture cache.\n"
 				"It can help to increase speed but it will likely create various glitches.";
+		case IDC_8BIT_CPU_FBHACK:
+			return "Forces 8-bit shader conversion to be done on the CPU instead of the GPU but performance will be slower.\n"
+				"Direct3D doesn't support shader conversion , OpenGL does support shader conversion but it doesn't render some corners properly.\n"
```
turtleli: Space before comma should be removed. Also `corners` -> `corner cases`

plugins/GSdx/GSSetting.cpp
```diff
@@ -122,6 +122,13 @@ const char* dialog_message(int ID, bool* updateText) {
...
+				"The hack can fix glitches or broken graphics in some games.\n"
+				"Harry Potter games (Direct3D and OpenGL).\n"
+				"Fifa Street games (Direct3D).\n"
```
turtleli: Fifa -> FIFA

plugins/GSdx/GSdx.rc
```diff
@@ -142,6 +142,7 @@ BEGIN
     CONTROL         "Alpha",IDC_ALPHAHACK,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,14,71,34,8
     CONTROL         "Alpha Stencil",IDC_ALPHASTENCIL,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,108,71,57,8
     CONTROL         "Disable Depth Emulation",IDC_TC_DEPTH,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,108,84,92,8
+    CONTROL         "CPU 8-bit FB Conversion", IDC_8BIT_CPU_FBHACK, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 14, 84, 92, 8
```
turtleli: There shouldn't be spaces after commas in rc files.

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -268,25 +270,16 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
...
-						else
-							dst = t;
-					} else {
+					// It is a complex to convert the code in shader. As a reference, let's do it on the CPU, it will be slow but
+					// 1/ it just works :)
+					// 2/ even with upscaling
+					// 3/ for both DX and OpenGL
+					if (m_cpu_8bit_fb_coversion && psm == PSM_PSMT8)
+						// Direct3D doesn't have a shader conversion so it needs to be done on the CPU side which causes slowdowns (Fifa Street).
+						// OpenGL has a shader but it doesn't handle some corners properly (Harry Potter, COD ?, Kung fu Panda ?).
```
turtleli: corner cases

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -268,25 +270,16 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
...
-						if (m_crc_hack_level < CRCHackLevel::Full)
-							Read(t, t->m_valid);
-						else
-							dst = t;
-					} else {
+					// It is a complex to convert the code in shader. As a reference, let's do it on the CPU, it will be slow but
+					// 1/ it just works :)
+					// 2/ even with upscaling
+					// 3/ for both DX and OpenGL
+					if (m_cpu_8bit_fb_coversion && psm == PSM_PSMT8)
```
turtleli: I wonder whether it would be better to have the option work the other way around (slower but better rendering by default, with the hack option to enable faster but worse rendering).

plugins/GSdx/GSTextureCache.h
```diff
@@ -130,6 +130,7 @@ class GSTextureCache
 	bool m_preload_frame;
 	uint8* m_temp;
 	bool m_can_convert_depth;
+	bool m_cpu_8bit_fb_coversion;
```
turtleli: Conversion is spelt incorrectly.

plugins/GSdx/GSTextureCache.cpp
lightningterror: It's probably a bad idea if it causes slowdowns and people dunno what's causing the issue.

plugins/GSdx/GSTextureCache.cpp
lightningterror: I updated the comment to match what's written on the tooltip. I think it's better.

lightningterror:
Updated.

p.s. the shader port will be in a separate pr. Don't wanna delay this a month (maybe) or so.

plugins/GSdx/GSTextureCache.cpp
gregory38: Unfortunately the impact on the perf could be huge. Read will trigger a synchronous read from the GPU memory. Which mean
1/ flush the driver thread
2/ flush all GPU commands
3/ finally read GPU memory

If the GPU is late on the rendering, the delay could be 1 or 2 renderer frames.

plugins/GSdx/GSdx.rc
```diff
@@ -142,6 +142,7 @@ BEGIN
     CONTROL         "Alpha",IDC_ALPHAHACK,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,14,71,34,8
     CONTROL         "Alpha Stencil",IDC_ALPHASTENCIL,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,108,71,57,8
     CONTROL         "Disable Depth Emulation",IDC_TC_DEPTH,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,108,84,92,8
+    CONTROL         "CPU 8-bit FB Conversion",IDC_8BIT_CPU_FBHACK,"Button",BS_AUTOCHECKBOX | WS_TABSTOP, 14, 84, 92, 8
```
turtleli: There are still spaces after some of the commas.

lightningterror: Oh, I completely missed those.

lightningterror:
You can merge it if the code is ok @turtleli.

plugins/GSdx/GSSetting.cpp
```diff
@@ -122,6 +122,13 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_TC_DEPTH:
 			return "Disable the support of Depth buffer in the texture cache.\n"
 				"It can help to increase speed but it will likely create various glitches.";
+		case IDC_8BIT_CPU_FBHACK:
+			return "Forces 8-bit shader conversion to be done on the CPU instead of the GPU but performance will be slower.\n"
+				"Direct3D doesn't support shader conversion, OpenGL does support shader conversion but it doesn't render some corner cases properly.\n"
```
turtleli: Is the text correct? `8-bit shader conversion` - I'm guessing it's supposed to be texture conversion? I'm not sure the shader situation needs to be explained in the tooltip. So something like:

> Uses the CPU instead of the GPU to convert 8-bit textures. May significantly reduce performance.
Fixes glitches or broken graphics in games such as blahdeblahdeblah

turtleli:
> You can merge it if the code is ok turtleli.

I think you should ask Gregory in this case ;)

plugins/GSdx/GSSetting.cpp
gregory38: Yes it should "8 bits frame buffer"

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -268,25 +270,20 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
...
-						if (m_crc_hack_level < CRCHackLevel::Full)
-							Read(t, t->m_valid);
-						else
-							dst = t;
-					} else {
+					// It is a complex to convert the code in shader. As a reference, let's do it on the CPU, it will be slow but
+					// 1/ it just works :)
+					// 2/ even with upscaling
+					// 3/ for both DX and OpenGL
+					if (m_cpu_8bit_fb_conversion && psm == PSM_PSMT8)
```
gregory38: Add also the 4 bits format `PSM_PSMT4`. I don't know if some game use it but there is no shader to convert this format. This way it would help to test bad graphics.

lightningterror: Seems weird to have the hack named 8bit after this change. CPU framebuffer conversion ?

gregory38: well it isn't important. Otherwise rename 8bits to index or palette format. Note the other 8 bit option handles also both format.

lightningterror: For the option in the gui maybe
CPU Texture conversion
To keep things simple.

lightningterror:
I made some changes / renaming. How does it look now ?

plugins/GSdx/GSSetting.cpp
```diff
@@ -122,6 +122,13 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_TC_DEPTH:
 			return "Disable the support of Depth buffer in the texture cache.\n"
 				"It can help to increase speed but it will likely create various glitches.";
+		case IDC_CPU_FB_CONVERSION:
+			return "Enables 4-bit and 8-bit frame buffer conversion to be done on the CPU instead of the GPU.\n\n"
```
gregory38: IMHO, the English construction is way too complicated.

You can directly use 
`Convert 4-bit and 8-bit frame buffer on the CPU ....`

lightningterror:
Updated.

atomic83GitHub:
This hack cause massive slowdown and a driver crash (nvidia) in Mercenaries 2.It also cause graphics glitch and slowdown in MGS3 under OpenGL (DX is beter with it).
This hack don't solve Kung fu panda and cause a GSDX out of memory issue.

lightningterror:
It's a hack. It should be enabled if only a game needs it. Otherwise no.

willkuer:
I guess it shouldn't produce crashes, or? Slowdowns and glitches are obviously a different story.

ssakash:
Why is this actually described as a hack? It's more accurate than the shader alternative,  right?

atomic83GitHub:
#gregory38 I use DX mode and I can confirm that the driver crash is a out of memory issue.

gregory38:
@atomic83GitHub thanks for the info.Potentially we could reduce the memory usage.

So far we do
* create an offscreen texture which contains a data buffer
* read the data
* put the texture in GSdx garbage collector

All calls are synchronous (hence the slowness). So even if we have multiple offscreen texture, only one data buffer will be used. So instead to duplicate all data buffers, we could use a big common one. Anyway, my feeling is that this hack isn't a good solution most of the time. I.e. better port the glsl shader that can handle 90% of conversion at fast speed.

lightningterror:
> Anyway, my feeling is that this hack isn't a good solution most of the time. I.e. better port the glsl shader that can handle 90% of conversion at fast speed.

Agreed. So far it seems only Harry Potter and TCoS will need it even with the ported shader. And I guess games that use 4-bit as well.

FlatOutPS2:
> All calls are synchronous (hence the slowness). So even if we have multiple offscreen texture, only one data buffer will be used. So instead to duplicate all data buffers, we could use a big common one. Anyway, my feeling is that this hack isn't a good solution most of the time. I.e. better port the glsl shader that can handle 90% of conversion at fast speed.

@gregory38 Which shader is it? I don't get different results on OGL compared to DX with this hack.

lightningterror:
https://github.com/PCSX2/pcsx2/blob/51c64fcbe6ce40d5f344195701175959aaa87533/plugins/GSdx/res/glsl/convert.glsl#L199

[FIFA-Street-8bits and Blending.zip](https://github.com/PCSX2/pcsx2/files/1652325/FIFA-Street-8bits.and.Blending.zip)
You can use this as a testcase.

FlatOutPS2:
@lightningterror Thanks for the info. 

I tried a quick port to DX for easier testing, but speed was terrible. I hope I did something wrong. :p Will look more into it later to hopefully improve/expand the shader.

lightningterror:
> From Gregory: Note you might need to add some dummy/stub shader for 11-16

This might help :stuck_out_tongue:

FlatOutPS2:
> This might help 😛

You don't need to add those. If anything, that would make things slower.

lightningterror:
> EDIT: Turns out I did make a mistake unrelated to the shader that made things slower. :)

I sense a PR Soon™.

Btw the shader on ogl isn't perfect either.
[Harry Potter COS.zip](https://github.com/PCSX2/pcsx2/files/1654030/Harry.Potter.COS.zip)
Harry Potter COS doesn't like the shader so it could be tweaked a bit more, do you wanna try to debug/solve the issue ?

FlatOutPS2:
> do you wanna try to debug/solve the issue ?

That was the point of my interest. I noticed your hack helps Stuntman, but I suspect the current shader doesn't work on that game because the game uses 4-bits.

Have you found any more games that need the hack/improved shader?

lightningterror:
> doesn't work on that game because the game uses 4-bits

The shader is only for 8-bits. A new shader probably needs to be written for 4-bit.

MGS3 maybe ? I'm not sure. 
OGL seems to work good with the shader, enabling FBC causes a whole bunch of other graphical issues.
I get no results on D3D with the FBC hack. The only result I get is enabling texture shuffle on D3D so it could be related to why I don't get any result in the first place. On top of that MGS3 also uses channel shuffle I think. Not sure tho.

gregory38:
The shader have 2 limitations
* 4 bits format isn't handled.
* page offset isn't supported either. This offset changes the pixel position.

In my opinion we should build a lookup table (texture) with the pixel position based on the various format offset.

8 bits texture is often uses with depth effect / channel shuffling / texture shuffling. It doesn't help to reduce the complexity... 

For the story I create the version of this shader for the blue fog of MGS3. However I don't remember if it was related to depth. I vaguely remember a (box?) fighting game (with night in the name) that also use 8 bits format (but it could be blending issue too).

I really should upload all my gsdump somewhere...

FlatOutPS2:
>  vaguely remember a (box?) fighting game (with night in the name) that also use 8 bits format (but it could be blending issue too).

Yeah, I found it. It's Fight Night Round 2. Another EA title(like FIFA Street). And like FIFA Street it also needs blending to properly fix the issue.
And my suspicion about Stuntman was proven correct. It does need a 4 bits handler.

lightningterror:
I'll just add this info here, it may be important. The FBC hack breaks channel shuffle effect on MGS3, dunno about other games.
All the more reason to replace the hack with a shader.

lightningterror:
How is progress going ?

FlatOutPS2:
> How is progress going ?

There's hardly been any progress. I've been busy with other PCSX2 projects. I'll have to add some basic blending too for this work properly in Direct3D, so it'll take longer than just a simple port plus improvements.

PR #2096, opened 2017-10-12
cdvdgigaherz: Fix Q subchannel relative offset calculation

turtleli:
The wrong comparison was used, so all the relative offsets were completely wrong. Fixes the wrong track issue in the CD player.

Regression introduced in f314c2a4d9a0ab349c0357b692312c137e2e6ac6.


MrCK1:
@turtleli You can merge this, looks fine on my end :)

PR #2099, opened 2017-10-16
GSdx: Automatic HW Mipmapping

lightningterror:
Adds automatic HW mipmapping support.
It relies on crc ids so if a game does
not have their crc id but needs mipmapping
it will not work until the id is added.

Add GUI menu and tooltip for Automatic mipmap
"Automatic (Default)"
This option will be default option from now on.

Rename "Very Slow" text option to "Slow" for full mipmap
as it caused the text not to fit properly in
the menu.

Credits also go to @RedPanda4552 and @ssakash for helping with the code.

Note: Not all games are added. So it needs a bit of work.


plugins/GSdx/GSdx.cpp
```diff
@@ -373,7 +374,7 @@ void GSdxApp::Init()
 	m_default_configuration["linear_present"]                             = "1";
 	m_default_configuration["MaxAnisotropy"]                              = "0";
 	m_default_configuration["mipmap"]                                     = "1";
-	m_default_configuration["mipmap_hw"]                                  = "0";
+	m_default_configuration["mipmap_hw"]                                  = "-1";
```
MrCK1: You still need to change the value in GSdx.rc, the dialog window is always initialized with Off as the default value.

Also, I don't think it's necessary (probably not good practice either) to start at `-1` in this case. I would just change it to `0` and adjust the other values accordingly.

plugins/GSdx/GSdx.cpp
```diff
@@ -276,9 +276,10 @@ void GSdxApp::Init()
 	m_gs_offset_hack.push_back(GSSetting(2,  "Special", "Texture"));
 	m_gs_offset_hack.push_back(GSSetting(3,  "Special", "Texture - aggressive"));
 
+	m_gs_hw_mipmapping.push_back(GSSetting(-1, "Automatic", "Default"));
 	m_gs_hw_mipmapping.push_back(GSSetting(0, "Off", ""));
 	m_gs_hw_mipmapping.push_back(GSSetting(1, "Basic", "Fast"));
-	m_gs_hw_mipmapping.push_back(GSSetting(2, "Full", "Very Slow - Not Recommended"));
+	m_gs_hw_mipmapping.push_back(GSSetting(2, "Full", "Very Slow"));
```
MrCK1: I think it's better to leave the original line and just change it to `Full (Slow - Not Recommended)`

plugins/GSdx/GSRendererOGL.cpp
```diff
@@ -698,11 +698,11 @@ void GSRendererOGL::EmulateTextureSampler(const GSTextureCache::Source* tex)
 	{
 		case TriFiltering::Forced:
 			trilinear = static_cast<uint8>(GS_MIN_FILTER::Linear_Mipmap_Linear);
-			trilinear_auto = m_mipmap != 2;
+			trilinear_auto = m_mipmap != 2 || !auto_mipmap_full;
 			break;
 
 		case TriFiltering::PS2:
-			if (need_mipmap && m_mipmap != 2) {
+			if (need_mipmap && (m_mipmap != 2 || !!auto_mipmap_full)) {
```
MrCK1: Typo, formatting seems inconsistent from the others.

plugins/GSdx/GSSetting.cpp
```diff
@@ -174,6 +174,9 @@ const char* dialog_message(int ID, bool* updateText) {
 			return "Uploads GS data when rendering a new frame to reproduce some effects accurately. Fixes black screen issues in games like Armored Core: Last Raven.";
 		case IDC_MIPMAP_HW:
 			return	"Control the accuracy level of the mipmapping emulation\n\n"
+				"Automatic:\nAutomatically sets the recommended mipmapping level based on games that need mipmapping.\n"
+				"This is the recommended setting.\n"
+				"Off will be selected for games that don't need mipmapping support.\nBaisc will be selected for games that need Basic mipmapping support.\nFull will be selected for games that need Full mipmapping support.\n\n"
```
MrCK1: Typo again.

plugins/GSdx/GSDeviceOGL.h
```diff
@@ -412,6 +412,7 @@ class GSDeviceOGL final : public GSDevice
 	uint32 m_msaa;				// Level of Msaa
 	int m_force_texture_clear;
 	int m_mipmap;
+	bool auto_mipmap_full;
```
MrCK1: Still missing `bool auto_mipmap_basic;`

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -841,6 +841,25 @@ void GSRendererHW::Draw()
...
+		case CRC::RatchetAndClank3:
+		case CRC::RatchetAndClank4:
+		case CRC::RatchetAndClank5:
+			auto_mipmap_basic = m_mipmap == -1;
+				break;
+			// Games requiring full mipmapping
+		case CRC::ApeEscape2:
+			auto_mipmap_full = m_mipmap == -1;
+				break;
+		}
```
MrCK1: I'm not sure if this really fits in this section; and I'm sure there's a more efficient way to do this without a bunch of case labels.

plugins/GSdx/GSCrc.h
```diff
@@ -109,6 +111,11 @@ class CRC
...
 		PiaCarroteYoukosoGPGakuenPrincess,
 		RadiataStories,
+		RatchetAndClank,
+		RatchetAndClank2,
+		RatchetAndClank3,
+		RatchetAndClank4,
+		RatchetAndClank5,
 		RedDeadRevolver,
 		ResidentEvil4,
 		RozenMaidenGebetGarden,
```
MrCK1: Ditto. (See comment above)

plugins/GSdx/GSCrc.cpp
```diff
@@ -251,7 +253,15 @@ CRC::Game CRC::m_games[] =
...
+	{0xC70FC973, ValkyrieProfile2, IT, 0},
+	{0xCE4933D0, RatchetAndClank, US, 0},
+	{0x76F724A3, RatchetAndClank, EU, 0},
+	{0xB3A71D10, RatchetAndClank2, US, 0}, // Going Commando
+	{0x38996035, RatchetAndClank2, US, 0},
+	{0x2F486E6F, RatchetAndClank2, EU, 0},
+	{0x45FE0CC4, RatchetAndClank3, US, 0}, //  Up Your Arsenal
+	{0x9BFBCD42, RatchetAndClank4, US, 0}, // Deadlocked
+	{0x8661F7BA, RatchetAndClank5, US, 0}, // Size Matters
 	{0x47B9B2FD, RadiataStories, US, 0},
```
MrCK1: These new CRCs should be in their own section that denotes them from CRCs required for CRC hacks.

```
//Auto Mipmapping CRCs
{0xC70FC973, ValkyrieProfile2, IT, 0},
{0xCE4933D0, RatchetAndClank, US, 0},
...
```

plugins/GSdx/GSRendererHW.cpp
RedPanda4552: Shot in the dark; if there is a register that changes value when a game uses mipmaps, that could be very useful. If such an obscure thing exists.

gregory38: A register exists but I can tell you that nearly all decent 3D games use mipmap. However you don't need to emulate it.

plugins/GSdx/GSDeviceOGL.cpp
```diff
@@ -241,7 +241,7 @@ GSTexture* GSDeviceOGL::CreateSurface(int type, int w, int h, bool msaa, int fmt
 	GL_PUSH("Create surface");
 
 	// A wrapper to call GSTextureOGL, with the different kind of parameter
-	GSTextureOGL* t = new GSTextureOGL(type, w, h, fmt, m_fbo_read, m_mipmap > 1 || m_filter != TriFiltering::None);
+	GSTextureOGL* t = new GSTextureOGL(type, w, h, fmt, m_fbo_read, (m_mipmap > 1 || auto_mipmap_full) || m_filter != TriFiltering::None);
```
gregory38: create an intermediate variable for `(m_mipmap > 1 || auto_mipmap_full) || m_filter != TriFiltering::None`. And you don't need the parenthesis

plugins/GSdx/GSRendererHW.cpp
gregory38: Yes it is bad to update this code here. It should be done in the CRC init of GSState. And IsMipMapActive should return true/false.

plugins/GSdx/GSDeviceOGL.h
lightningterror: No need since basic is not called in GSDeviceOGL.cpp

plugins/GSdx/GSCrc.cpp
lightningterror: I didn't because some games were already added in the list but sure I can move the ones that weren't.

plugins/GSdx/GSdx.cpp
lightningterror: The code in GSdx.rc controls where the button is placed.

Did you delete your GSdx.ini before doing the testing ?

atomic83GitHub:
Don't forget Harry poter games, Gran tourismo 4 and Half life.

lightningterror:
Gran Turismo 4 does not need mipmapping. I'm not sure about Harry Potter games.

atomic83GitHub:
Chamber of secrets and the Half blood prince massively use it.

MrCK1:
@lightningterror You're wrong about GT4. While there is no corruption if mipmapping is off, it's required to reduce texture flickering (which is what mipmapping is intended for anyways)

lightningterror:
Which version pal or ntsc? 
For me on pal it actually causes texture flickering.

atomic83GitHub:
I am with the Pal fr version, seeting these 2 Harry poter games with Mipmapping basic solve some incorect distant texture but set mipmapping to full cause texture flickering.

lightningterror:
This is what mipmapping causes on the ground textures. OpenGL and Direct3D so I won't add it if it fixes but breaks something else.

https://drive.google.com/open?id=0B8cQAq-Ko9-uMFV5T1hxVW9QRUU

MrCK1:
That shouldn't be happening, IIRC Full breaks this game. Try half instead.

lightningterror:
The video is from Half/Basic mipmap. Full does the same.

mirh:
> This is what mipmapping causes on the ground textures on GT4. OpenGL and Direct3D so I won't add it if it fixes but breaks something else.

The determinant should be whether you broke more than what you fix, tbh. 

Anyway, this should be expanded to include just about everything else too imo #1894

MrCK1:
I'm looking at the code again and I keep thinking that this might be breaking more games. If a CRC is already in the list for CRC hacks, mipmapping may be wrongly applied automatically and break more.

lightningterror:
All the code does is read the crc id of the game and applies mipmapping if it's on the list. Nothing else.
If a game already has crc hacks then they will function the same.

gregory38:
Looking at the code, it should be better to do something like that (in GSRendererHW.cpp)
```
void GSRendererHW::SetGameCRC(uint32 crc, int options)
{
	GSRenderer::SetGameCRC(crc, options);

	m_hacks.SetGameCRC(m_game);

      if (m_mipmap == AUTO_MODE)  {
                switch (m_game.title.title) 
                       case game1:
                       case game2:
                                  m_mipmap = ON;
                        break;
                        default:
                                 m_mipmap = OFF:
     }
}
```

It avoids to create new variable and to put more code in the rendering loop. (Note, you can extend the above with FULL mode)

plugins/GSdx/GSCrc.h
lightningterror: I prefer to have this list under alphabetical order. Looks clean.

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -186,6 +186,42 @@ void GSRendererHW::SetGameCRC(uint32 crc, int options)
 	GSRenderer::SetGameCRC(crc, options);
 
 	m_hacks.SetGameCRC(m_game);
+
+	// Code for Automatic Mipmapping. Relies on game CRCs.
+	if (theApp.GetConfigI("mipmap_hw") == -1) {
+		switch (m_game.title) {
+			// Games requiring basic mipmapping
```
ssakash: Instead of having comments for explaining each of the values of ``mipmap_hw``, I'd prefer if an ``enum class`` is added for it similar to CRC hack level.

plugins/GSdx/GSSetting.cpp
```diff
@@ -174,6 +174,9 @@ const char* dialog_message(int ID, bool* updateText) {
 			return "Uploads GS data when rendering a new frame to reproduce some effects accurately. Fixes black screen issues in games like Armored Core: Last Raven.";
 		case IDC_MIPMAP_HW:
 			return	"Control the accuracy level of the mipmapping emulation\n\n"
+				"Automatic:\nAutomatically sets the recommended mipmapping level based on games that need mipmapping.\n"
+				"This is the recommended setting.\n"
+				"Off will be selected for games that don't need mipmapping support.\nBasic will be selected for games that need Basic mipmapping support.\nFull will be selected for games that need Full mipmapping support.\n\n"
```
ssakash: Line 178 and 179 seems sort of redundant to me.

You've already hinted at 177 that this option sets the recommended mipmapping level, so it'd obviously be the recommended setting.

And Line 179, I'm not too fussed about this but it's just saying the same thing as the combobox texts, ``Basic will be selected for games that need Basic mipmapping support`` and ``Full will be selected for games that need Full mipmapping support``. Though if you do want to keep it, please remove capitalization in the second instance of ``Basic`` and ``Full`` you had used in your sentences.

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -186,6 +186,42 @@ void GSRendererHW::SetGameCRC(uint32 crc, int options)
 	GSRenderer::SetGameCRC(crc, options);
 
 	m_hacks.SetGameCRC(m_game);
+
+	// Code for Automatic Mipmapping. Relies on game CRCs.
+	if (theApp.GetConfigI("mipmap_hw") == -1) {
+		switch (m_game.title) {
```
ssakash: I was wondering if a dictionary might be better compared to switch-case, but maybe not.

plugins/GSdx/GSdx.cpp
```diff
@@ -276,9 +276,10 @@ void GSdxApp::Init()
 	m_gs_offset_hack.push_back(GSSetting(2,  "Special", "Texture"));
 	m_gs_offset_hack.push_back(GSSetting(3,  "Special", "Texture - aggressive"));
 
+	m_gs_hw_mipmapping.push_back(GSSetting(-1, "Automatic", "Default"));
 	m_gs_hw_mipmapping.push_back(GSSetting(0, "Off", ""));
 	m_gs_hw_mipmapping.push_back(GSSetting(1, "Basic", "Fast"));
-	m_gs_hw_mipmapping.push_back(GSSetting(2, "Full", "Very Slow - Not Recommended"));
+	m_gs_hw_mipmapping.push_back(GSSetting(2, "Full", "Slow - Not Recommended"));
```
ssakash: I think you can also use the enum values here instead of integer constants.

lightningterror: I think it's fine now. That will make the code just bigger. Plus there are other options that can be converted with enums as well if we look at it xD

ssakash: I'm just suggesting to use the already defined ``HWMipmap::Off`` instead of ``0`` at this place and similar for all other values. I'm not sure why you think it'll make the code bigger, you'll still have the same line of changes.

lightningterror: Ah I thought you meant as in also creating a new enum as well. I'll update it in a min.

plugins/GSdx/GSdx.cpp
```diff
@@ -373,7 +376,7 @@ void GSdxApp::Init()
 	m_default_configuration["linear_present"]                             = "1";
 	m_default_configuration["MaxAnisotropy"]                              = "0";
 	m_default_configuration["mipmap"]                                     = "1";
-	m_default_configuration["mipmap_hw"]                                  = "0";
+	m_default_configuration["mipmap_hw"]                                  = "-1";
```
ssakash: enum here over constant please.

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -186,6 +186,40 @@ void GSRendererHW::SetGameCRC(uint32 crc, int options)
 	GSRenderer::SetGameCRC(crc, options);
 
 	m_hacks.SetGameCRC(m_game);
+
+	// Code for Automatic Mipmapping. Relies on game CRCs.
+	if (theApp.GetConfigI("mipmap_hw") == -1) {
```
ssakash: Enum over here too, Either cast ``theApp.GetConfigI("mipmap_hw")`` to ``HWMipmap`` and do the check with ``HWMipmapLevel::Automatic``.

```
static_cast<HWMipmapLevel>(theApp.GetConfigI("mipmap_hw")) == HWMipmapLevel::Automatic
```

**(or)**

Do it the opposite way by casting the enum to integer. Your wish but the idea is to avoid magic numbers and use defined constants.

plugins/GSdx/GSState.h
```diff
@@ -218,6 +218,10 @@ class GSState : public GSAlignedClass<32>
 	GSPerfMon m_perfmon;
 	uint32 m_crc;
 	CRC::Game m_game;
+	// Needed for Automatic mipmapping.
+	// The original m_game variable is not used because then Automatic mipmapping will not work if CRC Hack Level is set to None.
+	// Using a separate variable addresses the mentioned issue.
+	CRC::Game m_game_mipmap;
```
turtleli: Maybe just use a local variable in GSRendererHW::SetGameCRC? It's not used anywhere else.

plugins/GSdx/GSSetting.cpp
```diff
@@ -181,6 +181,8 @@ const char* dialog_message(int ID, bool* updateText) {
 			return "Uploads GS data when rendering a new frame to reproduce some effects accurately. Fixes black screen issues in games like Armored Core: Last Raven.";
 		case IDC_MIPMAP_HW:
 			return	"Control the accuracy level of the mipmapping emulation\n\n"
+				"Automatic:\nAutomatically sets the recommended mipmapping level based on games that need mipmapping.\n"
```
turtleli: Maybe just "Automatically sets the mipmapping level based on the game." would do for this line? The repetition feels a bit awkward to me, but I dunno.

lightningterror: I wanted for it to be a little more consistent with how the CRC Hack level tooltip is.

turtleli: Well, it's the "games that require mipmapping" part that I find awkward - I think it's somewhat implied already. IMO "recommended" is somewhat implied too - I don't think we're deliberately going to choose what is suboptimal.

Though perhaps it should be "based on whether the game is known to require mipmapping".

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -186,6 +186,43 @@ void GSRendererHW::SetGameCRC(uint32 crc, int options)
 	GSRenderer::SetGameCRC(crc, options);
 
 	m_hacks.SetGameCRC(m_game);
+
+	// Code for Automatic Mipmapping. Relies on game CRCs.
+	CRC::Game m_game_mipmap;
+	m_game_mipmap = CRC::Lookup(crc);
```
ssakash: This could be had in a single line, also ``m_`` prefix is only used for member variables, not for local ones.

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -186,6 +186,42 @@ void GSRendererHW::SetGameCRC(uint32 crc, int options)
 	GSRenderer::SetGameCRC(crc, options);
 
 	m_hacks.SetGameCRC(m_game);
+
+	// Code for Automatic Mipmapping. Relies on game CRCs.
+	CRC::Game game_mipmap; game_mipmap = CRC::Lookup(crc);
```
ssakash: Actually, do we even need a local variable here? I think just using ``switch (CRC::Lookup(crc).title) {`` will be enough. Sorry for not mentioning about it earlier! It just struck my mind right now and I suck at properly reviewing things.

lightningterror:
I just found this out. Suggestions ?
https://github.com/PCSX2/pcsx2/pull/1853#issuecomment-285961816

gregory38:
Code is fine for me. Is everybody happy ? Otherwise code could be merged

RedPanda4552:
I did notice that there's no longer a condition for Full mipmap games in the switch. Was a decision made to scrap that, that I haven't seen in here?

lightningterror:
There are no games on the Full list atm so there is no reason to have it. If a game pops up it will be added ofc.
Ape Escape was removed since it seems there are issues with DX and trilinear filtering not supported with dx. (Ape Escape requires Mipmap Full + Trilinear(ogl only).

Temporus:
The Incredible Hulk: Ultimate Destruction need Basic Mipmapping for texture issues on buildings

PR #2104, opened 2017-10-25
pcsx2|common|gsnull: Remove GSprintf API

turtleli:
It's not really used, and the OSD uses a different API.

The specified calling convention (stdcall) is also incorrect since variadic functions are caller-clean, not callee-clean. The compilers ignore the stdcall and just use cdecl (I think), though it does trigger a -Wcast-calling-convention on clang.


MrCK1:
@turtleli Are you forgetting [GSApi.h](https://github.com/PCSX2/pcsx2/blob/master/common/include/api/GSApi.h#L81) or does that need to be kept?

turtleli:
I didn't touch that file because it seems outdated and unused.

gregory38:
Fwiw, once upon a time, there was a plan to replace current (PCSX1 inherited) API with a more adapted one. I think there was several trials but none succeed (likely due to plugin incompatibility and effort to port the plugin).

PR #2122, opened 2017-11-08
GSdx-TC: Fix load size calculation in target update

### LINKED ISSUE #1972: Many games crash with custom resolution.
***PCSX2 version:***
v1.5.0-dev-2117-g4a56240

***PCSX2 options:***
EE-IOP
Round Mode : chop zero 
Clamping mode : Normal

VUs
Micro-vu
Same as th EE-IOP

Hack:
None used

***Plugins used:***
GSDX dx9 (OpenGL have the same issue)

Many games randomly crash with custom resolution , native resolution (x3 for exemple) works.

In Silent hill 3, Load the first level of the game and walk to the librairy (where we see Claudia the first time) and you will see crash (before sometimes).
In Top Spin, just load the game and it will never past the FMV.

1.5dev 1435

***PC specifications:***
Amd Phenom 2 x4 955 be
Nvidia gtx 560ti

### LINKED ISSUE #2110: PCSX2 application crash on Rogue Galaxy when charge attack is used by Steve
***PCSX2 version:***
1.5.0-dev-2133-g45ac69a

- **Notes**:
No notes or logs, as Emu crashes and can't generate any

***PCSX2 options:***
Defaults, no speed hacks

***Plugins used:***
Any combination of plugins cause this issue
GSDX settings seem to have no effect.
Hardware/Software, DX9/11/OGL
Native/scaled
etc etc

When performing a Power attack as Steve PCSX2 crashes.

While playing as Steve(once you load the game push L3 and select Steve), enter a battle and hold X to charge your power attack and release, PCSX2 will crash.

Memory card file here: https://drive.google.com/file/d/0B1_PuJEKWxWlR1Bnek00MzI4Mms/view?usp=sharing

None, PCSX2 wiki has had this issue listed since i first tried to play in PCSX2 1.2/1.3
https://wiki.pcsx2.net/Rogue_Galaxy

***PC specifications:***
CPU: Core i5-4690K @4.5ghz
GPU: GTX 1080
OS: Windows 10 x64

### LINKED ISSUE #2138: [Regression] Beyond good and evil fmv's flickering
***PCSX2 version:***
Last 1.5 dev

***PCSX2 options:***
All default , speedhacks enabled (bug also without)

***Plugins used:***
All default

The beyond good and evil fmv's tends to flicker with the last 1.5git version of PCSX2, the screenshots don't render the issue accurately

Before:
capture

After:
capture32

Load PCSX2 and Beyond good and evil, go to the extra menu and load the BGE E3 trailer.

PCSX2 1.5dev 721

***PC specifications:***
AMD fx8350 Gainward 780ti Windows 10 64 Bit

***GSdump and Blockdump***
https://drive.google.com/file/d/1hpJTCFyuK9qxMIqVH1xTPQifyeoJ1Ev0/view?usp=sharing

ssakash:
**Summary of changes**:

* Fix the load size calculation in target update.
* Add an optional macro (``ENABLE_ACCURATE_BUFFER_EMULATION``) for enabling more accurate emulation of the buffer size. Haven't discovered any cases where it's necessary yet, but according to my old logs, it might be necessary on some superman game which uses 2048x2048 on scissoring.

Might help with #1972, also the code impacts every resolutions not just custom. So potentially there might also be some improvements apart from custom resolutions. Also please look out for the performance impact of this change.

Fixes #1972, Fixes #2110  (Custom resolution crashing regression)
Fixes #2138 (FMV flickering Beyond good and evil regression)

Anyone wants to help me test this? @atomic83GitHub @MrCK1 @lightningterror


atomic83GitHub:
I can help for this, can you post a custom PCSX2 version with the changes?

lightningterror:
https://ci.appveyor.com/project/gregory38/pcsx2/build/1.3425-master/job/tjqw0wv1ehxbcj1n/artifacts

MrCK1:
I'll test it, but for the record I never had crashing on custom res.

atomic83GitHub:
So, I have test Sillent Hill 3, Top spin and Kung fu panda, and none of them crash with custom resolution!
I have also noticed that the ingame texture seems more precise with any resolution wich improve the overal render, good!
Also: it seems that Sillent hill 3 did not display the up and left part of the inventory screen before this PR, it is now perfect.
capture
And for finish, I don't have any performance impact on my games.

ssakash:
> Also: it seems that Sillent hill 3 did not display the up and left part of the inventory screen before this PR

Was the issue exclusive to custom resolution before (or) was it also present in native scaling resolutions (2x, 3x, 4x...)?

Also thanks for testing, really appreciated. :)

atomic83GitHub:
It was present in any Hardware resolution (even with native).

lightningterror:
Can you mark it using paint so I can reproduce it ? :stuck_out_tongue_closed_eyes:

atomic83GitHub:
Before, in some places if you goes into inventory, you have this corruption, now it is gone for me.
capture

MrCK1:
self-reminder for myself to check Timesplitters later. :)

MrCK1:
Timesplitters is still busted unfortunately.

lightningterror:
Looks like it also solves MGS 2 and 3 crashing.
https://forums.pcsx2.net/Thread-MGS2-3-randomly-crash-after-1-3-minutes

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -1847,13 +1847,12 @@ void GSTextureCache::Target::Update()
...
-	GSVector2 t_scale = m_texture->GetScale();
+	GSVector2i t_size = default_rt_size;
 
-	//Avoids division by zero when calculating texture size.
-	t_scale = GSVector2(std::max(1.0f, t_scale.x), std::max(1.0f, t_scale.y));
-	t_size.x = lround(static_cast<float>(t_size.x) / t_scale.x);
-	t_size.y = lround(static_cast<float>(t_size.y) / t_scale.y);
+	// Ensure buffer width is at least of the minimum required value.
+	// Probably not necessary but doesn't hurt to be on the safe side.
+	int buffer_width = static_cast<int>(m_TEX0.TBW << 6);
```
gregory38: Do you really need the cast here ?

hellbringer616:
This PR fixes #2110

gregory38:
So if I understand it right, before we limit the upload based on GS size (t_size/t_scale). Now we limit it to the maximum size. However following code will stretch the data based on the scale factor.
```
	m_renderer->m_dev->StretchRect(t, m_texture, GSVector4(r) * GSVector4(m_texture->GetScale()).xyxy());
```
Strangely we avoid the 0 value in scale. But in this case, it means that uploading code is useless. So maybe it can be completely skipped. I'm not sure to understand where is the code fix.

Then it would be better to limit the stretch to the texture size. 
```
upscaled_size = GSVector4(r) * GSVector4(m_texture->GetScale()).xyxy()
```
```
min(upscaled_size, m_texture->GetSize());
```

Finally, perf wise, the impact is likely small. t_size is just a way to clamp the value of `r`. So it might only impact the preload hack (which is quite dumb)

lightningterror:
What's the status on this ?

lightningterror:
He was on discord 2 days ago.

hellbringer616:
No regressions on the games i've played (can provide a list if needed) solves a crashing issue on Rogue Galaxy when performing a power attack by Steve when using custom resolution.

FlatOutPS2:
The PR now fixes Beyond Good and Evil FMV's as long as the Memory Wrapping hack is enabled.

ssakash:
> Then it would be better to limit the stretch to the texture size.

That would change the behaviour compared to before and I'm not sure why we'd need to do it, the limit might only work on higher resolutions though (3x and above I guess), though is there any specific reason for suggesting to change the older stretch logic?

> Strangely we avoid the 0 value in scale. But in this case, it means that uploading code is useless. So maybe it can be completely skipped. I'm not sure to understand where is the code fix.

I'm not sure how a 0 value scenario for scale would be possible, the rescaling logic created chaos on custom resolutions where the buffer width and height values are unusual, removing that rendering size dependency was the fix for such cases.

> The PR now fixes Beyond Good and Evil FMV's as long as the Memory Wrapping hack is enabled.

Hmm, you mean the issue which was caused by one of gregory's older commits? (https://github.com/PCSX2/pcsx2/commit/265ea82780da1191dd6bbf69de4653c504c735af)

FlatOutPS2:
> Hmm, you mean the issue which was caused by one of gregory's older commits? (265ea82)

Yes, the issue from #2138 was fixed by your PR. The game requires memory wrapping because of a later change.

lightningterror:
So that's 2 regressions fixed. Pretty neat.

Gave Silent Hill2/3 a quick test again just to be sure it doesn't crash again with custom resolution and it's all ok.

FlatOutPS2:
> Gave Silent Hill2/3 a quick test again just to be sure it doesn't crash again with custom resolution and it's all ok.

Did you check the top left corner issue too?

lightningterror:
Well there is a way to corrupt the mini window, when you switch between hardware and software renders you can cause the mini window to be completely broken, either corrupted garbage graphics or just black window instead, as a matter of fact when you switch between hardware and software you can trigger an entire screen garbage but it lasts less than 1s.

Through normal gameplay I wasn't able to reproduce it, tho I have the NTSC version, or it's hard to reproduce I dunno.

FlatOutPS2:
Switching to the SW renderer and back didn't corrupt the mini window when I tried it. But I can confirm the less than 1s garbage when switching renderers.

I found a reliable way of reproducing the issue. Go to the menu, examine the pendant(didn't try other objects, probably works for other objects as well), then go to the options menu and straight back to the inventory. The mini window will now be corrupted. This PR nor  the Preload Frame Data hack, seems to affect the issue.

lightningterror:
That explains why I only reproduced it by chance.

Well this is the result I expected honestly.

atomic83GitHub:
I can confirm this PR solve Beyond Good and Evil FMV regression when the memory wrapping option is enable.

MrCK1:
No change in Syphon Filter unfortunately.

refractionpcsx2:
Is this ready to merge?

FlatOutPS2:
> Is this ready to merge?

We're waiting on a test of Rogue Galaxy.

refractionpcsx2:
Then I'll mark it as needs testing.

lightningterror:
This can be merged.

https://github.com/PCSX2/pcsx2/issues/2110#issuecomment-370030465

FlatOutPS2:
Not yet. @hellbringer616 mentioned no crashes so far, which suggest it's still being tested. ;)

hellbringer616:
After 3 hours of game play i can't get it to crash; I'd call that good

refractionpcsx2:
Before I merge this, do we want any "Closes" comments in the commit message?

lightningterror:
I edited the PR message, you can add them in the commit messages if it's needed.

> Fixes #1972, Fixes #2110 (Custom resolution crashing regression)
> Fixes #2138 (FMV flickering Beyond good and evil regression)

lightningterror:
Ah I've been waiting so long for this to get merged.

refractionpcsx2:
I bet ssakash has too :p

My only question, is this going to have any effect on the large frame buffer option? I feel like it's rendered it obsolete (although now it always uses a large frame buffer which is bad for perf)

lightningterror:
I think it will still help in games with fmv flickering just like before.

PR #2123, opened 2017-11-11
GSdx: Add/move CRC hacks for Street Fighter EX3

lightningterror:
Add missing CRC ids.
Move CRC hacks to OpenGL since the effects aren't emulated properly atm.

https://forums.pcsx2.net/Thread-Pushed-Bug-Report-Street-Fighter-EX3-NTSC-J

No hacks/option seem to work to workaround the issue even on OpenGL so I moved the hacks to Partial.


ssakash:
I think I have a block dump of this game, will look into it.

PR #2133, opened 2017-11-16
GSdx: Remove James Bond Hack

MrCK1:
It was originally used to prevent a VRAM spike in the weapons menu, but it isn't needed anymore.


ssakash:
There's a small typo in the commit message but probably not worth wasting time on fixing it.

Issue #1026, opened 2015-12-08
[Regression] D3D11 renderer adds black lines to bottom and right of the screen


karasuhebi:
While I was taking screenshots for #1025 to showcase the bug, I noticed that the D3D11 renderer adds very thin black lines on the bottom and right side of the screen, cutting off a bit of the image. This did not happen in 1.2.1 and it does not happen in the D3D9 or OGL renderers.

![6x](http://i.imgur.com/etZBbTh.png)

The lines are actually thicker in native resolution:
![native](http://i.imgur.com/YZUXqAc.png).

willkuer:
So you know the game...
Where did it broke?

gregory38:
Nvidia workaroud?

refractionpcsx2:
possibly? ugh... Can we just drop DX11?

ssakash:
@karasuhebi 

could you try using Git bisect and find the commit which caused this issue ? @refractionpcsx2  is a bit busy lately, It'll definitely make the fix a bit faster. :)

refractionpcsx2:
It probably is the nvidia workaround as I move the viewport over a little which seems to eleviate the stretching issue, but i am moving the top left to -0.001,-0.001 which would potentially make a black bar  in the bottom and right edges.

ssakash:
that makes sense , I think the fix should be then some kind of proper handling for the view port changes. maybe when it only satisfies a certain condition the Top left co-ordinate could be changed ?

refractionpcsx2:
Thats the problem, it's a bug in the Nvidia drivers and we don't know exactly the situation that causes it.  I did have a hack that made it always go through the geometry shader and that seemed to fix it for windows 7 and 8 but then it completely broke it in windows 10, which is frustrating.

gregory38:
Is still an issue? I didn't follow all hack iteration. (that being said, I see no reason to use D3D on Nvidia for most games).

willkuer:
A workaround is to use the gs window zoom function of the emulator. Settings zoom to something like 101% removes the outermost pixels. As it provides almost no negative side effects it can be constantly set to 101% independent on renderer choice.

FlatOutPS2:
> Is still an issue? I didn't follow all hack iteration. (that being said, I see no reason to use D3D on Nvidia for most games).

Yes, it's still an issue. DX11 is still widely used, not just by AMD gamers, but also by gamers with a slightly less powerful system.

gregory38:
[...]

PR #2145, opened 2017-11-25
GSdx: Add hidden option to disable Nvidia hack

lightningterror:
Add hidden option "UserHacks_DisableNVhack" to disable
the Nvidia hack on Direct3D which added black lines on the right
and bottom of the screen. Could be useful for Intel and AMD GPUs.

A better solution would be to add Vendor Id detection instead,
but this will do for now.

To disable the Nvidia hack add UserHacks_DisableNVhack=1 in GSdx.ini

Issue #1026

### DIFF
--- a/plugins/GSdx/GSDevice11.h
+++ b/plugins/GSdx/GSDevice11.h
@@ -58,6 +58,7 @@ class GSDevice11 : public GSDeviceDX
 	bool isNative;
 
 	bool UserHacks_unscale_pt_ln;
+	bool UserHacks_disable_NV_hack;
 
 	struct
 	{
--- a/plugins/GSdx/GSDevice11.cpp
+++ b/plugins/GSdx/GSDevice11.cpp
@@ -41,8 +41,10 @@ GSDevice11::GSDevice11()
 
 	if (theApp.GetConfigB("UserHacks")) {
 		UserHacks_unscale_pt_ln = theApp.GetConfigB("UserHacks_unscale_point_line");
+		UserHacks_disable_NV_hack = theApp.GetConfigB("UserHacks_DisableNVhack");
 	} else {
 		UserHacks_unscale_pt_ln = false;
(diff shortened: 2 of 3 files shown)


mirh:
[Mhhh](https://github.com/lightningterror/pcsx2/commit/52802371e47ee333f623d45bb67a313184d042ae)..
Reminds me of [this](https://github.com/PCSX2/pcsx2/commit/546adf52c9e6b9d4a93f1a8d31aeb634b93295c7).

lightningterror:
Just your typical driver bugs everywhere xD
Anyway I know that the sprite hack disables the nvidia hack  but enabling it may cause other issues where games don't need that specific hack.

mirh:
Well, idk. 
I think it would be a good practice to keep a unified list of all the driver hacks though.

ssakash:
Looks good to me, always handy to have a hidden option to check if the hack is needed or not instead of manually changing the code and compiling it every time a new driver comes up.

By the way, your vendor ID suggestion also seems good. You can also add it to this PR if you want.

lightningterror:
> By the way, your vendor ID suggestion also seems good. You can add it to this PR if you want.

I'll need your help with that :smile:

ssakash:
Look into #1795 for reference.

lightningterror:
You can merge it as it is now, I'll look in to improving the conditions how it works another time in the future :)

plugins/GSdx/GSDevice11.h
```diff
@@ -58,6 +58,7 @@ class GSDevice11 : public GSDeviceDX
 	bool isNative;
 
 	bool UserHacks_unscale_pt_ln;
+	bool UserHacks_disable_NV_hack;
```
ssakash: I'm not sure how this style of not using ``m_`` prefix for member variables got introduced, but I'm merging this for now as this style seems to be present for most of the variables on this class.

PR #2131, opened 2017-11-15
GameDB : critical changes for the upcoming 1.6 stable version.

atomic83GitHub:
A short version of the closed pull request with some new adds, this was made in order to save the time of the devs and for avoid users to manualy add these folowing hacks.

I had test them but I would like to have some other tests on these games before do anything else or merge the PR.

PR #2151, opened 2017-11-27
[WIP] Onepad 4 Controller Support

tsukasaminioncraft:
Unfinished modification to enable 4 controllers to work in pcsx2 using Onepad (not legacy)


tsukasaminioncraft:
If you could add the wip flag I'd appreciate it

MrCK1:
What do you mean by "unfinished"? Pull requests are generally supposed to be fully functional for the end-user.

Also, you're going to have to squash your commits before it can be merged in the future.

tsukasaminioncraft:
sorry, first time contributing back, I'm learning, usually I fork old projects and keep them updated.

I was asked by gregory to submit early so it would be easier for him to read what I've done so far

plugins/onepad/onepad.cpp
```diff
@@ -258,8 +259,9 @@ PADfreeze(int mode, freezeData *data)
 
         // Tales of the Abyss - pad fix
         // - restore data for both ports
-        for (int port = 0; port < 2; port++) {
-            for (int slot = 0; slot < 4; slot++) {
+		//more port < 2; to GAMEPAD_NUMBER, change slot for < 4 to < 6?
+        for (int port = 0; port < GAMEPAD_NUMBER; port++) {
+            for (int slot = 0; slot < 6; slot++) {
```
turtleli: I think `port` and `slot` should remain as 2 and 4 everywhere - A PS2 has 2 pad ports and you can plug in a multitap with 4 slots in each port.

tsukasaminioncraft: Very nice catch, I'll add it and compile and test, thanks, I believe I fixed the clang build issue also.

I will double check all port/slot vars

turtleli: The Travis CI clang build also checks the file formatting (which is why the build failed). If you run clang-format on the file (I think `clang-format -i /path/to/file`) it should auto-format it for you.

However, I do spot some weird formatting in the file (the indentation is wrong on some of the lines that you haven't touched), so if it still fails after that then don't worry about it for now.

tsukasaminioncraft: I usually fix programs in java, C isn't my strong suit.  But I'm very good with Linux, so I apologize in advance if I make some rookie mistakes :)

If you have advise on a decent/free IDE to use to make things easier to spot errors I'm more than willing to use it.  I've been direct editing my fork which in hind sight is not the best approach lol.

tsukasaminioncraft:
alright, done with the rapid changes, builds correctly now.
Current issues from this point
starting with deleted Onepad2.ini

1 if multitap 1 is checked, controller one controls all 4 players

2 if multitap 2 is checked (not 1), controller 2 controls players 2 3 and 4

3 deleted ini then open pcsx2 and load game without multitap = 2 working controllers fully like standard onepad, the ini will be saved without any uid's for the controllers, and only pad 0's controls written.

4 if you try to manually assign controllers to ports in GUI, game will crash on load, but the ini will contain UID for all controllers, and still only have the keymap for pad 0, the INI will need to be removed before normal function returns.

Thanks for any advise

gregory38:
@tsukasaminioncraft thanks you for the PR. As turtleli said, don't bother with clang.

I think you should look at `controller.h`. The field `packed_options` is likely wrong.

plugins/onepad/state_management.h
```diff
@@ -113,7 +113,8 @@ struct PadPluginFreezeData
 
 extern QueryInfo query;
 extern Pad pads[2][4];
-extern int slots[2];
+//fix for cpp and h mismatch
```
gregory38: You can fix it without a comment ;)

plugins/onepad/state_management.cpp
```diff
@@ -39,7 +39,8 @@ static const u8 queryAct[2][7] = {
 
 QueryInfo query;
 Pad pads[2][4];
-int slots[2] = {0, 0};
+//2 ports 4 slots with multitap
+int slots[4] = {0, 0};
```
gregory38: Add the 2 missing 0 => `{0, 0, 0, 0}` to avoid compilation warning.

plugins/onepad/onepad.h
```diff
@@ -21,8 +21,8 @@
 
 #ifndef __PAD_H__
 #define __PAD_H__
-
-#define GAMEPAD_NUMBER 2 // numbers of gamepad
+//this is where GAMEPAD_NUMBER max is assigned, possibly rewrite later for dynamic based on sdlgamepads detected
```
gregory38: If I'm right this value is the maximum number of PS2 gamepad. So it should be a constant. You could always use a keyboard to emulate a PS2 gamepad.

plugins/onepad/Linux/ini.cpp
```diff
@@ -106,12 +112,19 @@ void LoadConfig()
     if (fscanf(f, "ff_intensity = %u\n", &value) == 1)
         g_conf.set_ff_intensity(value);
 
-    size_t uid;
-    if (fscanf(f, "uid[0] = %zu\n", &uid) == 1)
-        g_conf.set_joy_uid(0, uid);
-    if (fscanf(f, "uid[1] = %zu\n", &uid) == 1)
-        g_conf.set_joy_uid(1, uid);
-
+    //size_t uid;
```
gregory38: Just delete old code.

plugins/onepad/Linux/ini.cpp
```diff
@@ -61,13 +61,19 @@ void SaveConfig()
     fprintf(f, "options = %d\n", g_conf.packed_options);
     fprintf(f, "mouse_sensibility = %d\n", g_conf.get_sensibility());
     fprintf(f, "ff_intensity = %d\n", g_conf.get_ff_intensity());
-    fprintf(f, "uid[0] = %zu\n", g_conf.get_joy_uid(0));
-    fprintf(f, "uid[1] = %zu\n", g_conf.get_joy_uid(1));
-
+//    fprintf(f, "uid[0] = %zu\n", g_conf.get_joy_uid(0));
```
gregory38: same as below, just delete old code.

mirh:
Maybe all the "4 controller support" commits could be squashed?
Or edited with a proper message?

tsukasaminioncraft:
making all the proposed changes now, thanks for all the help/advise/grief counselling.
Mirh:  I'm new to git pull requests, I'll try looking up "squashing commits" thanks for the pointer.

MrCK1:
@tsukasaminioncraft You can follow my instructions here, the only difference is that you'll be using bash/VIM or genie, ect. instead of Powershell and Notepad.

https://github.com/PCSX2/pcsx2/pull/2131#issuecomment-345154674

tsukasaminioncraft:
I see now that I should be using a local git repository and uploading my changes to github in 1 request, instead of editting github pages, I'll clone my fork and do that now, Thanks MrCK1

tsukasaminioncraft:
Question regarding multitap, is it implemented in the ps2, or in the multitap itself?  briefly browsed what I thought were the relevent files and couldn't find anything, so I was curious if the ps2 expects for example port0,a,b,c,d port0,1,2,3,4 when the checkbox is marked. I'll keep looking, just thought a heads up would save some time :)

gregory38:
To be honest, I don't know how multitap work. However it is working on Windows side so common code should be good (i.e. core emulator). I think you should fix the setting in controller.h to support multiple controller (code need to depends on GAMEPAD_NUMBER)

tsukasaminioncraft:
Looks like everything is already set to depend on gamepad_number in controller.h to me.  I'll keep reading the lilypad sources and see if I can find the exact methods each pad/port is fed to the emulator.  Shouldn't be too big a deal, just alot to read through for something I've probably overlooked already :)  Holiday season is gonna be effecting my free time off and on for awhile.  Thanks for the feedback.

gregory38:
```
    union
    {
        struct
        {
            u16 forcefeedback : 1;
            u16 reverse_lx : 1;
            u16 reverse_ly : 1;
            u16 reverse_rx : 1;
            u16 reverse_ry : 1;
            u16 mouse_l : 1;
            u16 mouse_r : 1;
            u16 _free : 9;             // The 9 remaining bits are unused, do what you wish with them ;)
        } pad_options[GAMEPAD_NUMBER]; // One for each pads
        u32 packed_options;            // Only first 8 bits of each 16 bits series are really used, rest is padding
};
```
packed_options is always 32 bits. However pad_options[GAMEPAD_NUMBER] is 32 bits for 2 pads and 64 bits for 4 pads.

plugins/onepad/onepad.cpp
```diff
@@ -162,7 +162,7 @@ PADinit(u32 flags)
     Pad::reset_all();
 
     query.reset();
-
+
```
lightningterror: Why the whitespace here?

plugins/onepad/Linux/ini.cpp
```diff
@@ -61,13 +61,16 @@ void SaveConfig()
...
+	for (int uidpad = 0; uidpad < GAMEPAD_NUMBER; uidpad++)
+		if (g_conf.get_joy_uid(uidpad) != NULL) {
+			fprintf(f, "uid(%d) = %zu\n", uidpad, g_conf.get_joy_uid(uidpad));
+		}
+	//something here isn't working, placeholder while investigating, possibly works with non-identical controllers
     for (int pad = 0; pad < GAMEPAD_NUMBER; pad++)
         for (auto const &it : g_conf.keysym_map[pad])
             fprintf(f, "PAD %d:KEYSYM 0x%x = %d\n", pad, it.first, it.second);
-
+
```
lightningterror: Why the whitespace here?

plugins/onepad/onepad.cpp
```diff
@@ -286,7 +286,6 @@ PADfreeze(int mode, freezeData *data)
         strncpy(pdata->format, "OnePad", sizeof(pdata->format));
         pdata->version = PAD_SAVE_STATE_VERSION;
         pdata->query = query;
-
```
lightningterror: I think it was better with the extra line here but maybe that's just me and my preference.

tsukasaminioncraft:
Hope everyone is having a great holiday season.
@gregory38  I've tried a few things in regard to "packed_options is always 32 bits. However pad_options[GAMEPAD_NUMBER] is 32 bits for 2 pads and 64 bits for 4 pads."

I've learned alot about struct and union in C++ lately, been a good experience.
trying to figure out how exactly to strip the padding and merge the results. I assume the trailing bits are what is being set, example 0000000010101010.  if indeed that gets read correctly from the emulator, if the full u32 is used without padding. setting packed_options to u64 wasn't a good move.  

@lightningterror I'll commit the whitespace changes after I get something else to commit with it to test.  Thanks for reading through it, formatting for readability is a good thing.

I'm not giving up on fixing onepad whatsoever, but I think in the future I'll make a branch and code a new plugin from scratch to learn C++ more completely.  Thanks for all the tips.

tsukasaminioncraft:
So the missing piece has to be outside of packed_options, they haven't made any noticeable difference at all changing it around in a few different ways.   Still 2 controllers work without multitap checked.  If multi1 is on, controller one controllers all 4 players, if multi2 is on, player 1 is good, and player 2 controls 2-4 like before.  I believe the whitespaces are all fixed now :)

plugins/onepad/Linux/ini.cpp
```diff
@@ -106,12 +109,12 @@ void LoadConfig()
...
-        g_conf.set_joy_uid(0, uid);
-    if (fscanf(f, "uid[1] = %zu\n", &uid) == 1)
-        g_conf.set_joy_uid(1, uid);
-
+	size_t uid;
+	for (int uidpadload = 0; uidpadload < GAMEPAD_NUMBER; uidpadload++)
+		if (fscanf(f, "uid(%d) = %zu\n", uidpadload, &uid) == 1) {
+			g_conf.set_joy_uid(uidpadload, uid);
+		}
+
```
lightningterror: Whitespace.

plugins/onepad/onepad.cpp
```diff
@@ -286,7 +286,7 @@ PADfreeze(int mode, freezeData *data)
...
+
```
lightningterror: Ditto.

gregory38:
@tsukasaminioncraft 
If you look at state_management.cpp, you have those kind of line to select the pad
```
                uint16_t buttons = g_key_status.get(query.port);
```
Maybe it should also depends on the `query.slot` variable.

plugins/onepad/Linux/ini.cpp
tsukasaminioncraft: Lol, nice catch, fixed, will send with next set of tests.

plugins/onepad/state_management.cpp
```diff
@@ -248,28 +248,27 @@ u8 pad_poll(u8 value)
 
                 if (pad->mode != MODE_DIGITAL) { // ANALOG || DS2 native
                     query.numBytes = 9;
-
-                    query.response[5] = g_key_status.get(query.port, PAD_R_RIGHT);
-                    query.response[6] = g_key_status.get(query.port, PAD_R_UP);
-                    query.response[7] = g_key_status.get(query.port, PAD_L_RIGHT);
-                    query.response[8] = g_key_status.get(query.port, PAD_L_UP);
+                    buttons, PAD_R_RIGHT;
```
gregory38: I guess you should redo the code here. I would advice to test your change locally before the push.

So current situation is rather bad. Technically we should update all `KeyStatus` code to use an array of array (port, slot). As a temporary solution we could do
```
// Build a linear pad number (range from 0 to 3) from the slot and the port. Technically there are 4 slots but KeyStatus code support only 4 gamepads (2 ports * 2 slots).
int pad_number = query.port + 2 * query.slot;
```
Then you could use this kind of code
```
query.response[5] = g_key_status.get(pad_number, PAD_R_RIGHT);
...
```

tsukasaminioncraft: Done, still no changes in functionality.  players 1 and 2 work like normal, if multitap is set, that controller controls that slot and the slots after it.  leads me to believe that setting multitap (either) is the source of the biggest blocker.  I'll keeping reading through and learning, Thanks for the issues you've found, you've been a big help.

gregory38: No you misunderstood me.
```
int pad_number = query.port + 2 * query.slot;
```
pad number must be updated after query update. Here the value is set once. Basically you need to compute the pad_number after
```
Pad *pad = &pads[query.port][query.slot];
```

By the way why did you remove test_bit test ?

tsukasaminioncraft: ahhh, still learning, C++ is quite different from java in a sickening number of ways.

Your example of 
"query.response[5] = g_key_status.get(pad_number, PAD_R_RIGHT);"
didn't have the !testbit part in it, so I mistakenly assumed that it should have been coded in that manner instead.  

I'll fix it up sir :)

The part that makes me wonder is nothing has functionally changed after any iteration of changes.  Could it be that this is legacy code from Onepad original that is superceded by the SDL2 code?

lightningterror:
What's the status on this ?

tsukasaminioncraft:
Sorry, got sidetracked with life.  Moved houses, extra hours, normal stuff.  I would love to get back into it as I do use pcsx2 often when I get free time to relax.  It wouldn't bother me if the pull was cancelled till I have time to work on it more and resubmit later, currently aside from all the commits, there's been zero functionality changed.

no multitap = 2 working controllers
multi1 enabled = controller 1 takes over all 4 players
multi2 enabled = controller 1 works, controller 2 takes players 2-4

I currently believe it has to do with the sdl code, I have read up on alot of tutorials, but haven't been able to find the exact spot that needs fixing.

lightningterror:
I'll close this for now, ping us when you resume working on it.

PR #2152, opened 2017-11-28
WinMisc: Update detection to flag XP as an unsupported OS.

MrCK1:
This is just a small update to WinMisc to bring it in line with previous changes relating to XP being dropped since 1.4.0. 

If anybody is still on XP for whatever reason, the console will graciously let them know that their OS is officially unsupported by us.

### DIFF
--- a/common/src/Utilities/Windows/WinMisc.cpp
+++ b/common/src/Utilities/Windows/WinMisc.cpp
@@ -24,58 +24,58 @@ static __aligned16 LARGE_INTEGER lfreq;
 
 void InitCPUTicks()
 {
-	QueryPerformanceFrequency( &lfreq );
+    QueryPerformanceFrequency(&lfreq);
 }
 
 u64 GetTickFrequency()
 {
-	return lfreq.QuadPart;
+    return lfreq.QuadPart;
 }
 
 u64 GetCPUTicks()
 {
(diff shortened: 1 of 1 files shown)


atomic83GitHub:
Windows server 2003 , XP pro 64bit and ReactOS are also concerned by this PR?

MrCK1:
ReactOS would appear as an unsupported OS anyways, because it's not a Microsoft Product. Windows Server 2003 and Win2K are the same "Major Version" category as XP.

Just so you know, this _doesn't actually stop those OSes from running PCSX2,_ it just flags them as `Unsupported Operating System` in the console. WxWidgets requires Windows NT as a minimum anyways, so that's the real limiting factor for us supporting older OSes.

common/src/Utilities/Windows/WinMisc.cpp
```diff
@@ -239,32 +227,7 @@ wxString GetOSVersionString()
 					retval += L"Standard Edition";
 			}
 		}
```
turtleli: So... why did you decide to keep this code?

MrCK1:
I tried to fix it the easy way by running clang-format on the file, but I guess I can't do that...

I'll try to do a proper rebase, but this might take a while until I remember the steps...

PR #2153, opened 2017-11-29
Flag XP as an unsupported OS.

MrCK1:
Roses are red, violets are blue, if I can't rebase properly; why should you 😛


turtleli:
So... why have you moved from working with a year old branch to a half a year old branch? While it won't cause a merge issue, it just doesn't make sense.

MrCK1:
Yeah, I know. I was being lazy but I need to stop forking old branches to avoid future problems.

PR #2208, opened 2017-12-08
PSX memcard to Pocketstation hack

RedPanda4552:
### Context
PSX titles made to be compatible with the Pocketstation peripheral, namely Final Fantasy VIII, send a different hex to SIO, 0x58, that was previously ignored. As a result, these games could not see PSX cards.
### Changes
This hack ~~when enabled~~ makes SIO execute the same payload for 0x58 as for normal PSX calls. ~~A checkbox is added to the main memcards screen to enable/disable. Defaults to enabled.~~
### The effect
The entire system sees PSX cards as Pocketstation devices. Though not "correct", this has no negative consequences, as Pocketstations are supposed to behave as memory cards. Additionally, games that use this Pocketstation value (and do not have other issues elsewhere in SIO) can now read and write the cards. It is confirmed that this works for Final Fantasy VIII (but only if there is not a PS2 card in slot 1, the starvation issue mentioned below) and I suspect this may help others as well, such as Saga Frontier 2, which was reported to completely stop working after the PSX SIO interrupts were introduced.
### A warning about memcard safety
In an attempt to also fix an issue where a PS2 card in slot 1 would starve out a PSX card in slot 2, I somehow managed to make it unformat any PS2 cards in the entire memcard directory. I rolled back these changes and have not been able to reproduce the issue, but it feels wrong to not mention it for safety.

### DIFF
--- a/pcsx2/Sio.cpp
+++ b/pcsx2/Sio.cpp
@@ -685,6 +685,7 @@ SIO_WRITE sioWriteMemcard(u8 data)
 		case 0x52: // PSX 'R'ead
 		case 0x53: // PSX 'S'tate
 		case 0x57: // PSX 'W'rite
+		case 0x58: // PSX Pocketstation
 			siomode = SIO_DUMMY;
 			break;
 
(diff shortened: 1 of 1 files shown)


lightningterror:
I suggest to also add a short summary what the PR does in the commit description.

ramapcsx2:
Oh, that's a good find!
This hack could be avoided if we just correctly replied with "Sorry, I'm just a regular Memory Card". It's probably a packet delivered right in the next reply ;)

RedPanda4552:
Yeah I've realized there has to be some sort of flag or trigger somewhere to tell the game "calm down it's just a memcard" but that's an ongoing effort as i keep tearing apart PCSXR and what PSX docs are available. Hopefully this hack won't be around for too long but in the meantime, it may help.

MrCK1:
If you need any more docs, let me know! I can give you the URLs if necessary.

ramapcsx2:
Okay, I understand your one step at a time approach.
If I understand this correctly, it will be good enough to just permanently reply to 0x58 for now.
PSX mode is declared as wip and Memory Card support is incomplete anyway.
I don't see any harm in reporting all PSX Memory Cards as Pocketstations for now.

So, how about just using the SIO part of this PR and not doing the GUI change?
That would reduce the impact of this temporary hack and make it easy to clean up, once the proper reply has been found.

ramapcsx2:
@gregory38 Is this okay with you? 
It is isolated to PSX mode and the PS2 Browser, since the Browser supports PSX peripherals and knows about Pocket Stations.
It's a minor hack and will help remind us to figure out the proper way to handle it.
In the mean time, it allows more progress in some of the nicer PSX titles :)

MrCK1:
@ramapcsx2 BTW, me and Pandubz changed the original thread that was used for memory card testing into an unofficial compatibility list with our results so far: https://forums.pcsx2.net/Thread-New-feature-Needs-testing-PSX-Mode-Unofficial-Compatibility-List

It's not looking too bad right now ;)

ramapcsx2:
Yep, looking good!
It's fortunate that this particular problem was so simple. :)

PR #2212, opened 2017-12-11
GameDB: Beyond good and evil, fix for wrong typo

atomic83GitHub:
I put a better version of the fix, round mode to nearest fix some issues with ingame typo as well as SPS.

Can be merge.

[skip ci]


lightningterror:
A little tip on adding skip ci

image

No need to add that many, just 1 is enough.

MrCK1:
You have a translation file in the commit, you better remove that first.

PR #2223, opened 2017-12-17
Game DB: fix for Star Wars Bounty Hunter and Petz-horsez 2

atomic83GitHub:
Add some fix for these 2 games along with a comment for Nascar 09

[skip ci]


MrCK1:
The comments on Nascar 09 aren't really necessary. While you're at it though, you can change the compatibility from 1 to 5. I'm not sure why that didn't get updated.

atomic83GitHub:
Done, I prefer keeping the comments because these games don't boot in a normal way.

atomic83GitHub:
I have done a squash for my 2 open PR, and I finaly be able to complete Star Wars so you can safely merge it.

PR #2237, opened 2017-12-28
PCSX2-WX: Display the video mode on titlebar

ssakash:
**Summary of changes**:

* Display current video mode at the title bar. (Useful for debugging purposes and for user feedback at cases when video mode is being changed)

### DIFF
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -249,7 +249,7 @@ static void vSyncInfoCalc(vSyncTimingInfo* info, Fixed100 framesPerSecond, u32 s
 	// is thus not worth the effort at this time.
 }
 
-static const char* ReportVideoMode()
+const char* ReportVideoMode()
 {
 	switch (gsVideoMode)
 	{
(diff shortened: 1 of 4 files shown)


pcsx2/gui/AppConfig.h
```diff
@@ -20,6 +20,8 @@
 #include "CDVD/CDVDaccess.h"
 #include <memory>
 
+const wxString DefaultTitleTemplate = L"Slot: ${slot} | Speed: ${speed} (${vfps}) | ${videomode} | Limiter: ${limiter} | ${gsdx} | ${omodei} | ${cpuusage}";
```
turtleli: Is this variable necessary? It's only used once.

Note: by default const has internal linkage, so every file that directly or indirectly includes AppConfig.h will have its own copy of the variable. The compiler might be able to optimise it away in a release build, but that probably won't happen in a debug build.

ssakash: Yeah, I'll be removing it. I initially had it for a different intention but not needed any more.

FlatOutPS2:
Good addition, but the location on the titlebar between Speed and Limiter is not as good.

karasuhebi:
Why is this on by default? I understand it can probably be useful for debugging purposes (so it's nice to have the option to turn it on if needed) but I don't think users care all that much when PCSX2 is on NTSC or PAL. And if they know enough to care about that, they can certainly figure out how to turn the option on.

PR #1368, opened 2016-05-24
PCSX2-Counters: Some Changes to Video mode counter stuffs

ssakash:
**Summary of Changes**:

* The detected values using the SMODE registers were for the video modes and not for the region of the game, changed the variable naming to video modes to prevent any confusions.

* Fixed a bug where 1080I was reported as progressive. ( I need more tests from other games where 1080I mode was used, @FlatOutPS2 - could you give GT4 a test ? )

~~* Future Change - I can only do it after #1234 is merged. **Don't merge yet**.~~

PR #2239, opened 2017-12-29
Counters: Fix Hblank calculation for DVD videmodes

ssakash:
**Summary of changes**:

- Made the DVD variant video modes to use the timing algorithm used by all the analog video modes, previously it was using the one we had for HD/SDTV modes which was weird. (Refer commit message for more information)

 @PCSX2/bug-squad 
This change would influence PSX games mostly, I don't have any PSX games with me so tests are appreciated.

### DIFF
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -59,7 +59,7 @@ static void rcntWhold(int index, u32 value);
 
 static bool IsAnalogVideoMode()
 {
-	return (gsVideoMode == GS_VideoMode::PAL || gsVideoMode == GS_VideoMode::NTSC);
+	return (gsVideoMode == GS_VideoMode::PAL || gsVideoMode == GS_VideoMode::NTSC || gsVideoMode == GS_VideoMode::DVD_NTSC || gsVideoMode == GS_VideoMode::DVD_PAL);
 }
 
 void rcntReset(int index) {


MrCK1:
Can't recall if any of my PSX/PS2 games even trigger the DVD modes, I'll test this once it compiles anyway.

ssakash:
> Can't recall if any of my PSX/PS2 games even trigger the DVD modes, I'll test this once it compiles anyway.

I thought PSX games always use NTSC/PAL DVD modes?

MrCK1:
I did a quick check with #2237 to verify. It looks like you're right, the BIOS also uses it.

seapancake:
Tested Rapid Reload (One the better games compatibility-wise and no difference, used #2237 build as well to confirm game uses DVD PAL.

lightningterror:
@RedPanda4552 can do some tests, he's got a huge collection.

RedPanda4552:
*Growing. But yes, I can. Is this change for the CDVD Plugin mode, ISO mode or both?

RedPanda4552:
Using out of box settings, except PSX titles marked as needing IOP interpreter and OGL SW renderer. Used ISO mode. For reference, here was my test pool by system:
**PS2**
- Gran Turismo 4
- Lego Star Wars The Original Trilogy
- Ratchet and Clank Going Commando
**PSX**
- Digimon World 2
- Digimon World 3
- Final Fantasy 8 (first disc only)
- Hot Wheels Turbo Racing (IOP Interpreter)
- Lego Rock Raiders (IOP Interpreter)
- Toy Story 2 (IOP Interpreter)

#2237 Results
All PS2 gameplay was NTSC. The PS2 BIOS splash always ran as DVD NTSC.
All PSX gameplay was DVD NTSC.

#2239 Results
PS2 games _appear_ unaffected. Lots of console output complaining about microVU1 and IPU, but I think that was due to SW renderer and not tweaking recompilers to match the games.

PSX Games:
- FMV corruption unchanged.
- EE hyperactivity unchanged.
- SIO quirks and memcard speeds unchanged.
- DMA3 Not Ready on PSX FMV's with -3 EE underclock (and sometimes on normal EE clock) unchanged.
- Game load speeds (not to be confused with memcard load) seem unchanged.
- Digimon World 3's frame repeating/strobing problem in HW renderers is unchanged.
-Lego Rock Raiders' game speed/synchronization discrepancies unchanged. Though, ingame graphics corruption seems to have disappeared. Loading screen graphic corruption is unchanged, however.

I just sort of threw out every "glaring" issue I've seen in PSX mode. Is there anything in particular I should be looking for?

ramapcsx2:
I'm out of the loop on what we consider "digital video mode".
All the PS2 video modes are analog, just with different timings.
Is the question here whether rounding should be applied?

ramapcsx2:
I'm going to merge this, on the assumption that this patch represents the currently best accepted theory on how this timing mode works.

ssakash:
> I just sort of threw out every "glaring" issue I've seen in PSX mode. Is there anything in particular I should be looking for?

The last time we got this wrong, it caused skippy FMV and out of sync issues in few games (https://github.com/PCSX2/pcsx2/pull/1368#issuecomment-224757278)

Thanks for all the tests by the way. :)

> I'm out of the loop on what we consider "digital video mode".

Not probably the accurate term but I just consider modes other than NTSC/PAL standard as digital video modes. I just use the classification between them to manage the timing code more accurately.

> Is the question here whether rounding should be applied?

Not just rounding, there's also halving of the hblank and hrender in case of video modes other than NTSC/PAL, previously that was being applied while it shouldn't have been.

> I'm going to merge this, on the assumption that this patch represents the currently best accepted theory on how this timing mode works.

Sure, I'm happy with the patch for now, seems like it fixed an issue (mentioned in pandubz's post) and hasn't caused any regressions so far.

ramapcsx2:
> halving of the hblank and hrender   

vSyncInfoCalc() is just so full of speculation, it's hard to tell if any later timing changes do the right thing.
It should definitely be rewritten.
> // Important!  The hRender/hBlank timers should be 50/50 for best results.
//  (this appears to be what the real EE's timing crystal does anyway)
u64 hBlank = Scanline / 2; 
u64 hRender = Scanline - hBlank;

Yea, I find that hard to believe. Even if the comment says that it's true. I just don't know how we can test for it.
It'd be best to integrate a test for this with the video mode test homebrew.

FlatOutPS2:
I did a quick test with a number of PSX and PS2 games and noticed no issues or other changes. I specificially checked for chopy/out of sync FMVs as mentioned, but there seems to be no change whatsover.

willkuer:
Don't you thinks it's kind of interesting that you changed something as important as timing by a non-negligible change as dividing or multiplying some part of the calculation by two and there is no major change? Just assume you divide your FSB's or one of your RAM modules' clockrate by two...

ssakash:
@willkuer 

Same here, I was expecting more games to be impacted by this but most of the games don't seem to care about the change as long as the blanking interval duration is similar. ¯\_(ツ)_/¯

PR #2244, opened 2018-01-02
Enhance "Assemble Opcode" to be pre-populated by existing instruction.

rkubes:
A primary use case for the "Assemble Opcode" function from the context menu is to modify an existing instruction. This commit populates the text entry with the existing opcode. This is quicker than trying to do a copy/paste, or rewrite the entire opcode if just one parameter is changing.
For the use cases where an entirely new opcode is written, it is just as easy to delete the populated opcode from the text entry.
One downside is this potentially shines more light on existing discrepancies between the assembler and disassembler. For example, when the disassembler displays "b 0x00001234" (short-hand of "beq zero, zero, 0x00001234"), if that short-hand is re-entered without change to "Assemble Opcode," then it is actually assembled as "j 0x00001234" - essentially changing it from a relative branch to an absolute jump. The user must know to write out the long-hand opcode for it to assemble as expected. This is also seen in some "load immediate" opcodes, where the disassembler shows it as the "li" short-hand, rather than the "ori" long-hand. However, the assembler cannot understand the "li" short-hand - the user must know to write out the long-hand.

### DIFF
--- a/pcsx2/gui/Debugger/CtrlDisassemblyView.cpp
+++ b/pcsx2/gui/Debugger/CtrlDisassemblyView.cpp
@@ -779,7 +779,7 @@ void CtrlDisassemblyView::onPopupClick(wxCommandEvent& evt)
 			break;
 		}
 	case ID_DISASM_ASSEMBLE:
-		assembleOpcode(curAddress,"");
+		assembleOpcode(curAddress, disassembleCurAddress());
 		break;
 	default:
 		wxMessageBox( L"Unimplemented.",  L"Unimplemented.", wxICON_INFORMATION);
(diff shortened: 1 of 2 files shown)


rkubes:
Regarding the "downside" I mentioned, I've since learned that the short-hand (simplified) display of disassembled instructions is driven by whether or not "Show Symbols" is enabled (toggled with the "Tab" key).

Therefore, I would no longer consider that a downside. It should be seen as expected behavior, since the user can toggle Show Symbols if needed. With "Show Symbols" toggled off, the "Assemble Opcode" in this patch will be pre-populated with the non-simplified (long-hand) instruction.

Kingcom:
Both of the downsides would be fixed if the assembler was updated to use the latest version of armips (currently it's based on an ancient stripped down version). I did some work on it last year (https://github.com/PCSX2/pcsx2/pull/1902), but never got around to finishing it... maybe it would be something you may be interested in picking up. It mostly needs proper integration into the various project files for the different platforms.

ramapcsx2:
So this is fine and should be merged?

PR #2247, opened 2018-01-03
Corrected MipsOpscodes table for unconditional branch.

rkubes:
The assembler was previously assembling instruction "b i" as an alias to "j I".
This caused unexpected behavior when attempting to assemble an unconditional branch. The previous behavior would cause position-independent code to fail due to the distinction between the absolute address of a jump vs. the offset of a branch.

Instruction "b i" is actually a simplification of "beq zero,zero, i" - an unconditional branch with an offset.
Prior to this patch, the assembler was treating "b i" as "j I" - a jump with an absolute address target.

### DIFF
--- a/pcsx2/DebugTools/MipsAssemblerTables.cpp
+++ b/pcsx2/DebugTools/MipsAssemblerTables.cpp
@@ -52,10 +52,10 @@ const tMipsOpcode MipsOpcodes[] = {
 //  hi |-------|-------|-------|-------|-------|-------|-------|-------|
 //		*1 = SPECIAL	*2 = REGIMM		*3 = COP0		*4 = COP1
 	{ "j",		"I",			MIPS_OP(0x02), 			MA_MIPS1,	MO_IPCA|MO_DELAY|MO_NODELAYSLOT },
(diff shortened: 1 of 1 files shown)


ramapcsx2:
Interesting, I don't suppose you found a game or some software that triggered the problem?

rkubes:
Yes, this is for the debugger only - specifically for the assembler portion of the debugger.

The debugger view properly disassembles unconditional branches in the format of "b i". However, if you try to assemble an unconditional branch in the format of "b i", the assembler changes it to the "j I" format and uses a completely different opcode.

In most cases this probably won't cause an issue. However, the assembled code now becomes position dependent. If the hex values of the assembled instructions are copied to other areas for patching (for example, if the modified code is not always loaded in the same area of RAM all the time), then the patch will fail due to jumping to the incorrect location.

Currently, the workaround is to always write out the long-hand "beq zero,zero, i" when needing to assemble an unconditional branch. This patch updates the MipsOpcodes table so that the unconditional branch is assembled correctly from the short-hand.

ramapcsx2:
Got it!

Issue #760, opened 2015-08-13
Enemies and characters in Final Fantasy X will sometimes face away from eachother in battle


myownfriend:
In the fight that Tidus has with the large underwater monster at the beginning of the game, he and the monster will be facing away from each other and they will turn around to attack. The same thing happens in the fight with the creature he fights near the campfire, though in that case, it's only the monster and only the portion of the fight with Rikku in it.

karasuhebi:
Known issue:
http://wiki.pcsx2.net/index.php/Final_Fantasy_X#Characters_face_the_wrong_direction_during_battles

refractionpcsx2:
This is a well known issue which requires clamping to fix, this isn't really something we can deal with. I guess we could change the rounding by default in the gamedb, however some users might complain about the speed drop.

myownfriend:
Ah, sorry about that. Should I close this issue or no?

refractionpcsx2:
Leave it open for now, i've marked it as a potential change to the game database, but ill leave this open for discussion about pro's and con's of doing it.

willkuer:
Default options should almost every time be accurate. Users can specify speedhacks if wanted.

Also is rounding really changing performance? Clamping is clear but rounding?

refractionpcsx2:
Sorry, I meant clamping, I wil edit my post

willkuer:
Well ee round to negative works as well.

refractionpcsx2:
probably because it rounds weird and avoids the need for the clamp, but i prefer changing clamping than rounding as rounding tends to have more unintended behaviour

PR #2231, opened 2017-12-25
GameDB: Max Payne, Klonoa 2, Evolution Snowboarding, Mana Khemia and Fifa Street 2 fixes

atomic83GitHub:
EE clamping mode to full fix Max Payne from crashing and freezing.
VU clamping mode to extra fix some reflexion issue in Klonoa 2.

Can be merge, I have done a lot of testing.

Close #2141

PR #2137, opened 2017-11-17
GameDB: Vampire Night fixes and code formatting.

atomic83GitHub:
A short version of the closed pull request with some new adds, this was made in order to save the time of the devs and for avoid users to manualy add these folowing hacks.

All games have been tested, can be merge.

- [x] Rebase to fix merge conflict.
- [x] Explanation in commit message why Vampire Night EETiminghack was removed.
- [x] Clean up all new comments by starting with a capital letter and ending with a point.

PR #2251, opened 2018-01-05
GameDB: Final Fantasy X fix

atomic83GitHub:
A long time that I want to post these fix for this major game.

It will need a least 1 test per version (PAL,NTSC,NTSC-J), I have tested the Pal F version and Round mode to negative fix some reverse boss and control issue. The International version is not supported in this PR.

Close https://github.com/PCSX2/pcsx2/issues/760

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -1109,6 +1109,7 @@ Serial = SCED-50642
 Name   = Final Fantasy X [Demo] [Final Fantasy VI PS1 - Bonus Disc]
 Region = PAL-E
 IPUWaitHack = 1
+eeRoundMode = 1 // Fix reverse control and boss in some places
 ---------------------------------------------
 Serial = SCED-50748
 Name   = Official PlayStation 2 Magazine Demo 26
@@ -1118,6 +1119,7 @@ Serial = SCED-50907
 Name   = Final Fantasy X [Bonus Disc - Beyond Final Fantasy]
 Region = PAL-Unk
 IPUWaitHack = 1
+eeRoundMode = 1 // Fix reverse control and boss in some places
 ---------------------------------------------
 Serial = SCED-51531
 Name   = Official PlayStation 2 Magazine Demo 33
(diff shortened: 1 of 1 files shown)


MrCK1:
Shawn Murray's Wakeboarding Unleashed hangs at the loading screen, it doesn't go in-gane at all.

atomic83GitHub:
Realy? What is this? https://forums.pcsx2.net/Thread-Core-Bug-Report-Shaun-Murray-s-Pro-Wakeboarder-PAL

FlatOutPS2:
The compatibility surely isn't 5, as what MrCK1 said is true. Even getting past the loading screen the status is more in-game than fully playable.
And though vuClampMode = 2 does reduce some further spikes/flashes, it has not been tested thoroughly, and might as well cause regressions in some parts of the game.

I suggest you don't add gameDB changes for games that you haven't thorougly tested yourself, let alone ones that you haven't played at all. And even when you're sure some fix is necessary, you need to have some backup sources showing the other regions of the game require the same changes. There's no guarantee the NTSC version requires any of these fixes.

atomic83GitHub:
Ok so let's forget about Wakeboarding, I will replace it with a another game that I have and this will need external test: Final Fantasy X : EE round mode to negative for reverse control and boss issue (and some minor compat update).

MrCK1:
By the way, I think we should slow down with the GameDB fixes. Opening more pull requests will only create more merge conflicts for you/us once we approve them ;).

atomic83GitHub:
@MrCK1, this is my last GameDB PR :). I pushed all my fixes and everything is ready for the upcoming 1.6 stable version on my side.

FlatOutPS2:
Are these compatibility ratings you add corroborate by the status on the website? They're not supposed to be changed just because you or some other contributor played the game. Each version/region has its own rating, so you shouldn't change it for all versions when only one is tested either.

I had a quick look at the other recent gameDB changes, and I noticed you changed the name of "Star Wars - Jango Fett" to "Star wars - Bounty Hunter", even though Jango Fett is what the Japanese versions of Bounty Hunter are called. You should probably revert that as that's not something that is done for other games across regions either.

lightningterror:
There won't be any merge conflicts if he doesn't modify the same portion of the code.

Anyway I do agree here, would have been better to have 1 pr with all the changes instead.
As for Star Wars a comment in the DB entity should be enough mentioning they are the same.

mirh:
TBH, the point of all his PR seems "this worked better here - let's extend this universally". 
Not saying it could never actually be the case.. Or that you should complete 100% the game.. But we are on a level that I could just google for "fixes" and call it a day. 

Also wtf flatout are you back?

atomic83GitHub:
@lightningterror , I separated the PR's because some games were not tested at the time. But now I think it's not anymore the case, basicly this PR is ready https://github.com/PCSX2/pcsx2/pull/2137 and this one just need more Fifa test on my side https://github.com/PCSX2/pcsx2/pull/2231 (And by the way, did you finish your test?).

FlatOutPS2:
None of these PRs are ready. You need to remove the compat=5's and check if there's any need to actually add any one these fixes. Some are very likely wrong.

bin/GameIndex.dbf
```diff
@@ -1109,6 +1110,7 @@ Serial = SCED-50642
 Name   = Final Fantasy X [Demo] [Final Fantasy VI PS1 - Bonus Disc]
 Region = PAL-E
 IPUWaitHack = 1
+eeRoundMode = 1 // Fix reverse control and boss in some places
```
FlatOutPS2: There was a discussion about rounding vs clamping for this issue here #760.

MrCK1:
@atomic83GitHub Are you going to update this PR soon to address the changes that @FlatOutPS2 mentioned?

FlatOutPS2:
@atomic83GitHub There's still a compatibility status change for the Korean version of Ratchet & Clank 3 in the commit. And you need to update the commit message to reflect the current status.

atomic83GitHub:
Fixed :).

FlatOutPS2:
One last thing. You should add that it closes this issue: https://github.com/PCSX2/pcsx2/issues/760

lightningterror:
> One last thing. You should add that it closes this issue: #760

That can be done in the PR, and I prefer if it's done in the PR as well, if it's done in the commit then every time you update the commits the issue gets spammed with mentions.

refractionpcsx2:
Done

Issue #895, opened 2015-10-17
[GSDX] Soul Calibur III [NTSC-U]


ssakash:
**_PCSX2 versions used:**_ pcsx2-v1.3.1-1256-gbbd74e5 (latest dev build)
**_Emulation settings:**_ "Safest" preset, no hacks.
**_Plugins used:**_ GSdx AVX2 using the OpenGL renderer and software mode.
**_Plugin settings:**_ CRC Hack Level "None"
#### Description:

Using the OpenGL renderer, character shadows seem to be missing. Also missing is the subtle scanline overlay. OpenGL Software mode works as expected.
## Screenshots:

OpenGL renderer. No character shadows and no scanline overlay.
![image:1](http://i.imgur.com/4of7qnI.jpg)

Software mode output, for reference:
![image:2](http://i.imgur.com/CzelVJ4.jpg)

**_GS dump**_ : http://forums.pcsx2.net/attachment.php?aid=57584

gregory38:
Please test Dx, various accurate option

prafullpcsx2:
If I tested correctly these shadows were fixed in software mode when @gabest11  recently fixed Lupin game. https://github.com/PCSX2/pcsx2/commit/42f51591df66c3b82d353dc6bb6aad7e14d1127d

So may be kinda hard to fix it in HW mode anytime soon.

ssakash:
These shadows are present in all non-software renderers according to my tests. Problem still present on Ultra level of Blending unit accuracy.

gregory38:
@prafullpcsx2 actually lupin fix isn't hard to implement (gabest did 99% of the job). The remaining issue is the impact on perf.

lightningterror:
This has been fixed on OpenGL with Depth Emulation. Issue remains on Direct3D since it lacks Depth support.

FlatOutPS2:
* Currently still needs CRC hacks to be disabled to fix it on OpenGL. ;)

lightningterror:
Soon™

PR #2256, opened 2018-01-08
GSdx: Move CRC hacks for Soul Calibur 2&3 to DX level

lightningterror:
Add missing CRC ids for Soul Calibur 2 and 3.
Move CRC hacks to DX level. They are not needed anymore on OpenGL since Depth Emulation fixes depth issues (shadows).
  
Improves #895

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -491,13 +491,17 @@ CRC::Game CRC::m_games[] =
 	{0xC4D0FACC, SDGundamGGeneration, JP, 0}, // cutie comment
 	{0xBBDE6926, SDGundamGGeneration, JP, 0}, // cutie comment
 	{0x49D60A00, SDGundamGGeneration, JP, 0}, //NEO
+	{0x86C57952, SoulCalibur2, JP, 0},
 	{0x83AFB38A, SoulCalibur2, KO, 0},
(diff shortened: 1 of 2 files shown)


FlatOutPS2:
Seems like you can merge GSC_SoulCalibur2 and 3 into one. They do exactly the same thing under the same conditions.

lightningterror:
Yeah I know, me and @RedPanda4552 had a discussion about it but he was a bit skeptical about it if a new was to be added in the future.

FlatOutPS2:
Unless you know someone is making a hack or there is a pressing issue that will likely see a new hack, I don't see why they can't be merged. If someone in the faraway future needs to make a hack, it would could be for OpenGL as well so it would need a new GSC anyway, it could affect both games in the same way like this or the current CRC hack might have been removed/replaced.

lightningterror:
Both games need **VU clamping mode Extra** as well, but I'll save that for another PR.
Maybe @atomic83GitHub will do it :stuck_out_tongue:

atomic83GitHub:
@lightningterror, as I said in the FFX topic, I will not update anymore the GameDB :). And I saw that Soul Calibur 2 already have the fix.

FlatOutPS2:
@atomic83GitHub You don't need to make a new PR anymore. You can add the fix @lightningterror mentions here and then merge all the GameDB PR's you currently have open into one with only the confirmed fixes left.

PR #2260, opened 2018-01-10
cmake: Don't print a git describe error with shallow clones.

orbea:
My pcsx2 build script will use shallow clones with `git clone --depth=1` to save bandwidth and time when cloning, but at some point this started printing the following error message because `git describe` does not work with shallow clones.
```
fatal: No names found, cannot describe anything.
```
To work around this `git describe --always` can be used instead. For a full clone the behavior will remain unchanged and for a shallow clone it result in a less verbose description instead of an error.

For example.

Full clone
```
$ git describe --always
v1.5.0-dev-2247-g29ef586f9
```
Shallow clone
```
$ git describe --always
29ef586
```


gregory38:
Thanks

PR #2262, opened 2018-01-11
GSdx: Add Dark Cloud CRCs to automatic mipmapping

FlatOutPS2:
Needed for some background scenery textures.

Seems VS deleted spaces at the end of a few lines, and it didn't mess anything else up. That's a first, so I left it intact.


ssakash:
> Seems VS deleted spaces at the end of a few lines, and it didn't mess anything else up. 

It's fine, always nice to get rid of trailing white spaces.

PR #2264, opened 2018-01-12
SPU2-X: Prevent volume check from altering volume

FlatOutPS2:
Fixes issue in Tokyo Xtreme Racer Zero where checking the left master volume would alter it.

Replaces closed PR(to push from dedicated branch).


refractionpcsx2:
Yay! I was hoping you'd come back to this :) Thank you.

PR #2265, opened 2018-01-12
GSdx: Disable Merge Sprite on native resolution

lightningterror:
Previously if HW hacks were enabled Merge Sprite was active(if checked) on
native resolution even if the GUI option was disabled, which in result
caused glitches in games on native resolution.

This should address that issue.

### DIFF
--- a/plugins/GSdx/GSRendererHW.cpp
+++ b/plugins/GSdx/GSRendererHW.cpp
@@ -56,8 +56,9 @@ GSRendererHW::GSRendererHW(GSTextureCache* tc)
 	}
 
 	if (m_upscale_multiplier == 1) { // hacks are only needed for upscaling issues.
-		m_userhacks_round_sprite_offset = 0;
-		m_userhacks_align_sprite_X      = 0;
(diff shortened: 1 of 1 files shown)


mirh:
*Uhh*
F* knows how many times I made wrong comparisons given this for granted 😓
How about `unscale_point_line`, `SpriteHack`, `merge_pp_sprite` and `HalfPixelOffset` instead?

lightningterror:
I didn't see any issues with the others if they are enabled so I think it should be fine.

ssakash:
I think it'd be fine to also disable those other options, we don't need them in native resolution anyway.

lightningterror:
HPO https://github.com/PCSX2/pcsx2/blob/fe27d3b25a6ec4ad0f4186960fea17bf2b01f6d0/plugins/GSdx/GSRendererHW.cpp#L421

Unscale (dx , ogl has similar)
https://github.com/PCSX2/pcsx2/blob/4a562401dcb0c501424f74af26618c119956f709/plugins/GSdx/GSRendererDX11.cpp#L51

PR #2272, opened 2018-01-23
PCSX2: Add configurable Cheats folder to Components Selectors

### LINKED ISSUE #1584: [Feature Request] Open cheat folder in explorer window button.
This is a quality of life feature I'd like to see, similar to how you have for memory card management ie. "Open in Explorer" I'd like to see the same but for cheats, since if you want to add, disable, or edit your cheats you'll need to go into that folder to edit the pnach file in notepad or whatever, and where I keep my PCSX2 folder is quite a few levels deep on my HD, so it gets annoying I can't do this in a few clicks.

Not sure where you would put this option to open in explorer, maybe right underneath the "Enable Cheats" under System.

lightningterror:
Add GUI option in the Plugins Component Selector to change the default "Cheats" folder location.
Similar to savestates/logs/..etc

This has been bugging me. I have no idea how many times I deleted my cheats folder by mistake when updating to newer dev builds. This should solve all those issues for many people.

image

Close #1584

### DIFF
--- a/pcsx2/gui/Panels/PathsPanel.cpp
+++ b/pcsx2/gui/Panels/PathsPanel.cpp
@@ -34,6 +34,14 @@ Panels::StandardPathsPanel::StandardPathsPanel( wxWindow* parent )
 	: BasePathsPanel( parent )
 {
 	*this += BetweenFolderSpace;
+	*this += (new DirPickerPanel( this, FolderId_Cheats,
+		_("Cheats:"),
+		_("Select folder for Cheats") ))->
+		SetToolTip( pxEt( L"This folder is where PCSX2 loads cheats from."
+		)
+	) | SubGroup();
+
+	*this += BetweenFolderSpace;
 	*this += (new DirPickerPanel( this, FolderId_Savestates,
 		_("Savestates:"),
 		_("Select folder for Savestates") ))->


MrCK1:
Just wondering, how do you accidentally delete your cheats folder when upgrading dev builds? If it was backed up in a safe place, we wouldn't have to go through the trouble of changing all of the translatable strings again.

lightningterror:
Well since I have per game configs all of the files (bios, memory cards,  .. etc) are located in documents, so when I purge the old version I delete everything and keep forgetting about the Cheats.

As for changing the translatable strings again well they need to be changed anyway when FlatOut's PR gets merged :stuck_out_tongue:

mirh:
I wonder if all those options couldn't be moved under a single "stuff" path?

FlatOutPS2:
> I wonder if all those options couldn't be moved under a single "stuff" path?

That would not be a good idea. Someone could have, say, a global cheats folder somewhere outside of the PCSX2 folder, while using the default paths for all other folders. Although adding a global option that affects all folders instead of having to modify all of 'em(folder options for bios/mc/plugins are all located elsewhere in the UI) when you do have all folders in the same location could be helpful.

pcsx2/gui/Panels/PathsPanel.cpp
```diff
@@ -33,6 +33,14 @@ Panels::BasePathsPanel::BasePathsPanel( wxWindow* parent )
 Panels::StandardPathsPanel::StandardPathsPanel( wxWindow* parent )
 	: BasePathsPanel( parent )
 {
+	*this += BetweenFolderSpace;
+	*this += (new DirPickerPanel( this, FolderId_Cheats,
+		_("Cheats:"),
+		_("Select folder for Cheats") ))->
+		SetToolTip( pxEt( L"This folder is where PCSX2 custom cheats can be stored."
```
turtleli: Maybe
```
This folder is where PCSX2 searches for cheats.
```
or
```
This folder is where PCSX2 loads cheats from.
```
would be better?

lightningterror: I'll go with the second one.

lightningterror:
@turtleli do you wanna merge this ?

turtleli:
Well, you assigned ssakash so I figured you wanted him to review your PR?

lightningterror:
Ye since he's usually the one that does gui updates but seems he's a bit busy these days.
Tho you also reviewed :P

PR #2273, opened 2018-01-23
cdvd: Fix blockdump writing issues

### LINKED ISSUE #2266: CDVD: Blockdump writing issues
Blockdump reading seems to work fine with PCSX2. However, there seems to be a few issues on the writing side:
 * It uses the old format dump, which creates much larger dumps than necessary (the files compress well, but still).
 * There is code to use the new format dumps, but there is a bug that causes it to crash (easily fixed)
 * CD (.bin) dumps do not work at all (not so easily fixed - I haven't figured out all the issues here yet) - The workaround is to use cdvdlinuz (CDVDiso.dll) to create CD blockdumps until it gets fixed (unfortunately I removed it from the repo a while back, though it should be easy enough to checkout a older git revision and build it or find the dll from an older PCSX2 build).

turtleli:
Changes:
 * Fixes writing to V2 block dumps.
 * Switches to writing V2 block dumps.
 * Fixes CD blockdumps.

Fixes #2266.

### DIFF
--- a/pcsx2/CDVD/CDVDaccess.cpp
+++ b/pcsx2/CDVD/CDVDaccess.cpp
@@ -394,7 +394,7 @@ bool DoCDVDopen()
 	cdvdTD td;
 	CDVD->getTD(0, &td);
 
-	blockDumpFile.Create(temp, 3);
+	blockDumpFile.Create(temp, 2);
 
 	if( blockDumpFile.IsOpened() )
 	{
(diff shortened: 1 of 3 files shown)


FlatOutPS2:
I used this dump to get a smaller blockdump(of a compressed iso that's not supported by older PCSX2 versions) and it worked nicely. Went from a massive 2 GB to under 40 MB.

bositman:
Tested a CD game blockdump, worked fine and was really small compared to before. Great job as usual @turtleli  :)

Issue #1712, opened 2016-12-22
Jak series, unsual object speedup 


FerrazaDazzle:
***PCSX2 LATEST DEV

***PCSX2 options:***
EE, VU default, Fast loading off, everything else under speedhacks are on

***Plugins used:***
SSE4.1, everything else is default

***Description of the issue:***
Jak series have some issues where there is excessive speedups on player (best example is on Jak 2, at the first fortress mission avoiding the KG machine which shoots you) volume and everything else runs at default 60 including voices but the player moves so fast.

***How to reproduce the issue:***
attempt Jak 2, get up to 5% through the game (SAVE FILE LINKED BELOW) where you have to go into the fortress and play there.

https://mega.nz/#!TsR0iZha!HFJVM0xp5Xyct6vgTo-aTxtljFPFYW7cmPjeiS23dOY   -Jak 2 save file, PLAY SAVE 2. (save slot 2 with 5% complete)

***Last known version to work:***
never worked in any version

***PC specifications:***
i7- 5960x, GTX 980Ti G1 Gaming, Windows 10 Enterprise 64bit

FlatOutPS2:
I believe one of the EE/VU slider settings on the speedhacks tab removes this speedup. Try them all one by one and see if it helps.

FerrazaDazzle:
@FlatOutPS2 when EE is set to -33% it helps FEW parts, VU doesnt help really. Regarding this, there is also a FPS issue with the Jak series, some parts even lag on native.
in terms of FPS, Jak 2 has the most issues, Jak 1 is near perfect, Jak 3 has very few issues.

MrCK1:
Is this really a valid issue?

SpiffyJUNIOR:
Yeah, this is still an issue even on the latest version. Increasing the EE rate to 180% or higher helps remove some of these slowdowns, although this will cut your FPS in half or more.
The reason this happens is because Jak automatically speeds up and slows down the game based on the framerate. J_P on the fourms managed to pin point the commit that caused this regression though:
(https://github.com/PCSX2/pcsx2/commit/c8a7a709dcc33172bc252fb6ac7f2fb75da65605)
This most likely won't ever be fixed unless a custom branch is made just for this game :(

Issue #360, opened 2014-11-26
Shadow of Rome - Audio cuts in and out during FMVs


Kolrath:
Game: Shadow of Rome
Region: NTSC-U.  Unsure if, but assume, it occurs in other region releases.

Issue:  During FMVs the audio will start to cut in and out.  Each video starts off fine, but quickly get progressively worse.  In some long FMVs the audio will actually cut out entirely for the last portions of the FMV.

The quickest way to replicate the issue is to simply watch the opening intro that plays after the CAPCOM logo when you start the game.  By the time the video reaches the third line of dialogue "and the rule of the senate elders had begun to fray at the edges",  and is starting to show a temple on screen, the audio will start cutting in and out.  This will continue throughout the rest of the FMV.

Was there ever a working commit?  Yes

The last build with working audio was:
pcsx2-5903-windows-x86
With the notes:
"Vsync: Put it back to half slices, there seems to be games which don't like the vsync timing more accurate either way. Until i can work out why we will go back to the old half n half timing which all known games seem happy with."

The build that the audio stuttering first occurs is:
pcsx2-5904-windows-x86
With the notes:
"If you're gonna do a revert, you might as well do it properly :P If you don't want it doing properly, give it to me."

Not sure what changed between those two builds that caused the audio stuttering but 5904 seems to be the culprit.

During testing I used the "Preset 1 - Safest" settings and also tried changing the SPU-X plugin to async.  Didn't have an effect on the audio issue.  Regardless of setting changes I made build 5903 and anything before it worked fine while build 5904 and anything after it had the audio cutting in and out.

FlatOutPS2:
http://forums.pcsx2.net/Thread-Pushed-Bug-Report-Shadow-of-Rome-NTSC-U

Bug report on the forum mentions this has been a on again, off again bug:

> extra note: I have found the build which breaks the audio after hours of testing. The version that breaks the audio is v1.3.1-1539-gc8a7a70. From then on the cut scene audio don't work properly even on the newest builds. Before that version it was working perfectly. 

So it has been fixed and regressed at least twice.

ssakash:
> It seems like it has been fixed and regressed at least twice.

That's what happens if the timer code keeps getting reverted :P 

http://forums.pcsx2.net/Thread-vSyncInfoCalc-timing-change

DaanDeMeyer:
I have the same problem. Tried playing around with different settings but so far nothing seems to fix the issue.

[...]

PR #2275, opened 2018-01-24
PCSX2: Vsync Timing gamefix

FlatOutPS2:
Adds Vsync timing gamefix to Gamefixes panel of the UI. Fixes timing issues in Jak II (randomly runs too fast), Fatal Fury Battle Archives Volume 1 (runs at half-speed) and Shadow Of Rome (FMV timing issues).

Close: #360
Close: #1712

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -70,6 +70,7 @@
 -- VIF1StallHack     = 1 // SOCOM II HUD and Spy Hunter loading hang.
 -- FMVinSoftwareHack = 1 // Silent Hill 2-3. Fixes FMVs that are obscured when using hardware rendering by switching to software rendering while an FMV plays.
 -- ScarfaceIbitHack  = 1 // VU I bit Hack avoid constant recompilation. ( Scarface The World Is Yours)
+-- VsyncTimingHack   = 1 // Alternate Vsync timing. Fixing timing issues in Jak II, Fatal Fury Battle Archives 1, Shadow Of Rome.
 
 ---------------------------------------------
 -- Speed Hacks (SpeedHackName = <value>)
@@ -485,6 +486,7 @@ Region = NTSC-Unk
 Serial = SCAJ-20073
 Name   = Jak and Daxter II
 Region = NTSC-Unk
+VsyncTimingHack = 1
 ---------------------------------------------
 Serial = SCAJ-20074
 Name   = King of Fighters 2002, The
@@ -1697,6 +1699,7 @@ Serial = SCES-51608
 Name   = Jak II: Renegade
 Region = PAL-M7
 Compat = 5
+VsyncTimingHack = 1
 ---------------------------------------------
 Serial = SCES-51610
 Name   = This is Football 2004 (Red Devils 2004)
@@ -2800,6 +2803,7 @@ EETimingHack = 1 // Else serious slowdown due to VIF timing. Can be disabled onc
 Serial = SCKA-20010
 Name   = Jak II
 Region = NTSC-K
+VsyncTimingHack = 1
 ---------------------------------------------
 Serial = SCKA-20011
 Name   = Ratchet & Clank 2
@@ -3377,6 +3381,7 @@ Region = NTSC-J
 Serial = SCPS-15021
 Name   = Jak & Daxter 2
 Region = NTSC-J
+VsyncTimingHack = 1
 ---------------------------------------------
 Serial = SCPS-15022
 Name   = Dual Hearts
@@ -3542,6 +3547,7 @@ MemCardFilter = SCPS-15056/SCPS-15037
 Serial = SCPS-15057
 Name   = Jak & Daxter 2
 Region = NTSC-J
+VsyncTimingHack = 1
 ---------------------------------------------
 Serial = SCPS-15058
 Name   = Arc the Lad - Generation
@@ -5055,6 +5061,7 @@ Serial = SCUS-97265
 Name   = Jak II
 Region = NTSC-U
 Compat = 5
+VsyncTimingHack = 1
 ---------------------------------------------
 Serial = SCUS-97266
 Name   = Final Fantasy XI [Disc1of2]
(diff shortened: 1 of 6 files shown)


ramapcsx2:
Well, if anything, this confirms that the timings are indeed off.

mirh:
*Changes translatable strings*

MrCK1:
Any ideas of what else this could possibly fix?

FlatOutPS2:
>Any ideas of what else this could possibly fix?

Similar issues in other games. :p

atomic83GitHub:
Great job! I have just a question on this, why integrate this as a hack? It can break other games?

Also I have this error in the console with your PR:

capture

FlatOutPS2:
> Great job! I have just a question on this, why integrate this as a hack? It can break other games?

It has been changed a lot true the history of PCSX2, constantly breaking one game or another. ;)

> Also I have this error in the console with your PR:

Good catch. There was a `=1` missing.

mirh:
> Great job! I have just a question on this, why integrate this as a hack? It can break other games?

Nope, because it's being selectively enabled (thanks to gameindex)

atomic83GitHub:
Hum, Jak 2 seems not affected by this PR, I still have speed issue :(.

FlatOutPS2:
> Edit, I found why, When the game is in the 60 hertz mode, it is fixed but when it is set in the 50 hertz mode, it still have the speed issue.

I couldn't reproduce the issue on 50 Hz earlier. I retested and did encounter it this time. I've included the fix for 50Hz.

ramapcsx2:
> Blank = (Frame / scansPerFrame) * (gsVideoMode == GS_VideoMode::NTSC ? 26 : 22);
This sounds better than what we originally do (the same 22 hblanks in both, PAL and NTSC modes) but I would have assumed that PAL has the longer vblank.
In any case, someone once mentioned that the video timings would all be the same, regardless of the output mode.
Does someone still have that test homebrew that is mentioned in the source?

atomic83GitHub:
It is available here: 
https://www.dropbox.com/s/i5aeb9ele6roqg4/VModeTest.elf?dl=0

refractionpcsx2:
I wonder if this will have an affect on the french game Gift too, that goes really jerky, you have to mess around with the framerate and the turbo to get it to work smoothly

FlatOutPS2:
> I wonder if this will have an affect on the french game Gift too, that goes really jerky, you have to mess around with the framerate and the turbo to get it to work smoothly

Did it ever work before? Cause that sounds quite similar to the behaviour of Fatal Fury BA1.

MrCK1:
Might have to try Shark Tale again but I guarantee it' still borked.

ramapcsx2:
The test looks like this:
https://i.imgur.com/HK8fWBq.png

So I still have this elf but no source for it. We don't know what it's doing exactly.
By the way, check out those numbers ><

refractionpcsx2:
If you wanna try it, here's a blockdump of Gift (done with linuz iso because the internal one is broken :P) https://drive.google.com/open?id=1yMxLzbP18yaiE4tT75q02wYJGzEWK--E

That'll take you up to the menu, you'll be able to see from there the game runs at about 15fps.

MrCK1:
I only tested it over Teamviewer so there's a bit of delay, but it *feels* like R&C2 runs better. I can see the "speed" difference in The Suffering as well.

FlatOutPS2:
@refractionpcsx2 What framerate is Gift supposed to run at?

refractionpcsx2:
30 I think.  If you set the PAL framerate to 25 then normal speed to 200% (I think that's right) it runs pretty smooth, but as it is, it's unplayably jerky in the game

ramapcsx2:
Okay, so I don't know what's up with the EE and GS hsyncs but it does look like the EE one does the proper 525 hsync per NTSC frame.
And we do 22. Go figure ><
Maybe it works because we also wrongly fire gate events?

lightningterror:
Can you update this PR ? I see conflicting files.

FlatOutPS2:
@lightningterror I'll do that when I update the PR. There's no reason to do it now.

lightningterror:
@refractionpcsx2 , @ramapcsx2 so what are your suggestions to do with this pr ?

refractionpcsx2:
The only thing you can do if this definitely an improvement. Rip it off and fix the problems then merge it.

ramapcsx2:
I agree on merging this.

This topic needs new hardware tests, with some verification that the results are correct.
I don't know what the old test used, but maybe it should be more than just an EE timer or whatever.
Also, it definitely must not rely on the GS taking render time, since in PCSX2 the GS is kind of infinitely fast :p

Silanda:
I'd just like to add that the Phantasy Star Collection needs this fix too. Phantasy Star 4 runs with massive slowdown at times unless either this fix is used or the EE cycle rate is increased to 180% (to be safe).

ramapcsx2:
Oh, that's good to know! 
I only knew about overclocking the EE as a fix, but the overclock necessary didn't seem plausible.
Also, the EE isn't the fastest general purpose CPU, but it shouldn't struggle with Mega Drive emulation.
So a sync timing fix is probably what the game needs, and this change will make it easier to test.

ramapcsx2:
So I spent some time investigating once more. 
The gamefix proposed here should be correct, though we don't know for sure if it differentiated between PAL and NTSC (or the other video modes!). 
That distinction is insignificant compared to the current code though.

With the gamefix code, the earlier timing test .elf runs correctly for the vblank counter and duration. It looks exactly like the original hardware screenshot, too.
The gamefix works for Phantasy Star 4 (Mega Drive emulation).

I haven't found a game that broke yet, and I tested everything that I had that was mentioned in this PR, plus some others.
Are there any known titles that break with this change?

MrCK1:
Thanks for the hardware test! I was wondering if this was more accurate to hardware behavior. I will test the games I have and look for regressions.

ramapcsx2:
Oh, I didn't do a new hw test, just check with the existing .elf. Compare this new result to the earlier original hw screenshot that ref posted:
https://i.imgur.com/pY1v6Wi.png

MrCK1:
That's what I meant to say actually ;)

ramapcsx2:
Kk ;)
The new results look entirely plausible, and the before result obviously bad.
So if possible, the fix should be the default, unless it breaks stuff..

lightningterror:
Closing in favor of https://github.com/PCSX2/pcsx2/pull/2835

TaRan7uLa:
I don't understand what will I do for a new github user, Could you please explain for new PCSX2 user?  I do not have "Vsync timing gamefix on Gamefixes panel of the UI". I have PCSX2 1.4.0 and 1.5.0 dev version. How will I make this fix and what I download?

Issue #2022, opened 2017-07-22
GSdx: Incomplete texture shuffle/ or depth emulation on texture shuffle


lightningterror:
***PCSX2 version:***
v1.5.0-dev-2143-g1d983a681

***PCSX2 options:***
Default options , no fixes , native resolution

***Description of the issue:***
PAL games that require Depth Emulation are rendered incorrectly compared to their NTSC counterparts.
Only part of the textures are rendered in pal versions leaving the rest untouched or making it worse.
Related issue #1339

***How to reproduce the issue:***
Test pal games that make use of depth emulation and compare them to ntsc counterparts.

***Last known version to work:***
None

***PC specifications:***
i3 4160 , r7 360 amd , windows 7 64 bit

***Additional info:***
Affected games: DBZ BT3 #2001 (pal), Area 51 (pal)
Other games are very likely to be affected.

_Games that also had their crc hacks (depth related) moved to dx only should also be checked as they might experience glitches on opengl pal versions._

For comparison Area 51
Depth off
![gsdx_20170722200837](https://user-images.githubusercontent.com/18107717/28493739-fc376520-6f1b-11e7-90db-c8523d1f14d5.png)

Depth on
![gsdx_20170722200811](https://user-images.githubusercontent.com/18107717/28493742-08129a22-6f1c-11e7-811e-929a2fa75bd2.png)

Area51 GS Dumps
https://drive.google.com/file/d/0B8cQAq-Ko9-uZE1GcWFUbmNlTVU/view?usp=sharing

atomic83GitHub:
Yes , same issue on Rayman revolution and Half life (pal).

MrCK1:
Somebody who owns a PS2 with PAL bios should test these games to make sure it's not just an error related to playing PAL games with NTSC bios. Can't hurt to try I guess.

atomic83GitHub:
I have a pal bios with these games :) .

Hirato:
I wonder if the workaround gregory put in place for FFX-2 ( 247d33a62705cc809321408fa5bd93e577a851c8 )is related to this, since my FFX-2 was also PAL region.

lightningterror:
Nah the issue is present on 1.4 as well so it's not that. That code works for all regions btw so it's not just for pal.

Hirato:
I was more just wondering if that issue shared a common cause,
I didn't mean to imply "I wonder if this caused it."

Much like the showcase here, the game had all manner of rendering oddities
[...]

Issue #1339, opened 2016-05-03
Meta: GSdx HW half-bottom screen issue


gregory38:
Let's list all games impacted. This way it would be easier to check any potential implementation.

- [ ] mercenaries
- [ ] destruction derby
- [ ] MGS3 codec (some versions)
- [ ] get away

gregory38:
@refractionpcsx2 @ssakash @FlatOutPS2 @prafullpcsx2
I let's you complete the list :+1:

FlatOutPS2:
The Getaway doesn't have a half-screen bug as far as I know, it did have a fog wall, but it's been fixed.
Black Monday does have a depth issue with sun rayes, not sure if it's a half-bottom screen issue.

Games with HW half-bottom screen issue:
Need For Speed Hot Pursuit 2
Killzone
Total Immersion Racing(fog in wet races)
TimeSplitters (Future Perfect)
WRC II Extrme (fog in distance)

gregory38:
I need to check my get away dump (or maybe I mix it with another game)

prafullpcsx2:
Oops. Sorry I directly added few games to the first list. i hope gregory wasn't editing the post at the same time.

refractionpcsx2:
I added one :P Cant remember many off hand

pepiczech77:
NFS Underground could be probably added too.

refractionpcsx2:
Does that have the half screen issue? I've never seen it on that game

FlatOutPS2:
I believe most, if not all, PS2 NFS games have the issue. NFS Underground 2 and Most Wanted definitely.

refractionpcsx2:
Alrighty, added all those to the list.

pepiczech77:
I think so. Bottom half of the screen is blurred or something and it misses fog. 
Probably even NFS Carbon has it, but it is not the only issue there, because the game looks too green in HW mode. I will check those games on real PS2 soon.

refractionpcsx2:
Yeh carbon had a new engine I believe, so it might be different, need to test it, I have it but haven't touched it in a while

pepiczech77:
I tested it, bottom half of the screen is rendered differently than the upper half.

refractionpcsx2:
OK I'll add that one as well then

refractionpcsx2:
I've also had a scan through the forums to try and find some other games with these issues and added them to the list.

MarcoEstevez:
[...]

PR #2278, opened 2018-01-26
GSdx: Adjust DBZ BT2 CRC hacks and id, add R&C3 id, change skipdraw GUI limit

lightningterror:
Adjust region id for BT2.
Move the sky texture(depth) hack back to Partial level only for the EU regions.  Effect is still not rendered correctly and causes a half screen bottom issue (#2022)(#1339).

GSdx: Change Skipdraw Maximum limit value.
Previously the limit was 1000, now 10000 in the GUI. It should help in
some rare cases where a higher number is needed without the need of ini
editing and value reset issues caused by the GUI.

### DIFF
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -58,11 +58,17 @@ bool GSC_DBZBT2(const GSFrameInfo& fi, int& skip)
 	{
 		if(fi.TME && /*fi.FBP == 0x00000 && fi.FPSM == PSM_PSMCT16 &&*/ (fi.TBP0 == 0x01c00 || fi.TBP0 == 0x02000) && fi.TPSM == PSM_PSMZ16)
 		{
-			if (Dx_only) // Feel like texture shuffle but not sure
-				skip = 26; //27
+			// Alpha channel (red line issue has been fixed).
+			// Sky texture (Depth) is properly rendered on OpenGL only for the NTSC version. The PAL version still has some issues (half screen bottom issue).
+			if (g_crc_region == CRC::EU || Dx_only)
+			{
+				skip = 26;
+			}
 		}
 		else if(!fi.TME && (fi.FBP == 0x02a00 || fi.FBP == 0x03000) && fi.FPSM == PSM_PSMCT16)
 		{
+			// Character outlines.
+			// Glow/blur effect. Causes ghosting on upscaled resolution.
 			skip = 10;
 		}
 	}
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -92,7 +92,7 @@ CRC::Game CRC::m_games[] =
 	{0xB0D195EF, MetalGearSolid3, KO, 0}, //Metal Gear Solid 3 Subsistence disc1
 	{0x3EBABC9C, MetalGearSolid3, KO, 0}, //Metal Gear Solid 3 Subsistence disc2
 	{0x8A5C25A7, MetalGearSolid3, ES, 0}, //Metal Gear Solid 3 Subsistence Spanish version
-	{0x278722BF, DBZBT2, US, 0},
+	{0x278722BF, DBZBT2, EU, 0},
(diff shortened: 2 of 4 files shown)


lightningterror:
I just realized that I created this PR master fork ... oh well.

FlatOutPS2:
If you based the GitS changes solely on the gsdump, did you ask someone to confirm it does not cause any issues elsewhere in the game?

lightningterror:
I actually have the game so I did my testing playing but not much.
D3D should be the same since the code wasn't changed for it.
OpenGL has the hack adjusted so that can be further tested.

@ivan89el wanna do a bit more testing?
You can download the build here
https://ci.appveyor.com/project/gregory38/pcsx2/build/1.3702-master/job/mdrxexya1bdfu8s3/artifacts

FlatOutPS2:
Then why ask for a gsdump? :S

You should also be clearer that you change the skipdraw limit with this PR. Even looking at the code I might've missed that one. Is the hyper extreme skipdraw setting still necessary with the updated CRC hack?

lightningterror:
Was lazy to search for the disc :stuck_out_tongue: 
Yeah the high skipdraw is needed because it uses a channel shuffle effect which is not implemented properly yet.

turtleli:
I think you should also make the same change to the skipdraw range on the Linux GUI as well. I'd also prefer if you split the GUI change into a separate commit.

lightningterror:
@bositman did some more testing and everything is good. Aside from shadows not working with the gamefix patch but the patch is needed to go through a door in the beginning. After that gamefixes can be disabled.

turtleli:
Did you comment on the wrong PR?

lightningterror:
No, he was testing GITS Stand Alone Complex so I thought I'd share his test results here as well, oh seems that I forgot to mention which game earlier.

FlatOutPS2:
It kind of looks like you were telling bositman you did the testing. 😛

bositman:
He told me to test it and so I did :P

lightningterror:
> It kind of looks like you were telling bositman you did the testing.

It kinda does, but if you look closely I never used **I** in the sentence :)

lightningterror:
I decided to remove the GITS update, I want to do some revisiting on that but in the future.

FlatOutPS2:
Looks good to merge then. I might have to rebase my skipdraw offset PR because of 10000 change.

plugins/GSdx/GSHwHack.cpp
```diff
@@ -58,11 +58,17 @@ bool GSC_DBZBT2(const GSFrameInfo& fi, int& skip)
 	{
 		if(fi.TME && /*fi.FBP == 0x00000 && fi.FPSM == PSM_PSMCT16 &&*/ (fi.TBP0 == 0x01c00 || fi.TBP0 == 0x02000) && fi.TPSM == PSM_PSMZ16)
 		{
-			if (Dx_only) // Feel like texture shuffle but not sure
-				skip = 26; //27
+			// Alpha channel (red line issue has been fixed).
+			// Sky texture (Depth) is properly rendered on OpenGL only for the NTSC version. The PAL version still has some issues (half screen bottom issue).
+			if (g_crc_region == CRC::EU || Dx_only)
+			{
+				skip = 26; // 27
```
turtleli: Maybe remove the `// 27`?

turtleli: Test both, see what's better, and keep only that value then. Comment was introduced in 357030330a587aa51e1461a0b8009919b40be594.

FlatOutPS2: Based on the commit message the 27 value was used when it only worked on the PAL version. Either it was just a general correction, or using the higher value removed something it shouldn't on the NTSC versions.

lightningterror: > or using the higher value removed something it shouldn't on the NTSC versions

Yeah it was that. Feast your eyes on this, looks lovely doesn't it with 27 :smile: 

pcsx2 2018-02-05 15-30-10-92

lightningterror:
You can merge it now.

PR #2283, opened 2018-01-28
[WIP] Fix SIO interrupt timings and Baud register

RedPanda4552:
From what I can tell, the Baud register is supposed to be a constantly updating 16 bit timer that influences timing on SIO interrupts. Looking at PCSXR, since it's source is readily viewable on Github, they seem to use the Baud register multiplied by 8, whereas PCSX2 has been using a constant of 64. Since enabling the PSX SIO interrupts fixed memory cards for many games **but completely broke Silent Hill**, I question if this constant is the problem and we need to be using the Baud register as well. Which is where part two of the problem comes in.

As is, the Baud register in PCSX2 does not update. Ever. Thinking this was a SIO or IOP problem, I compared to PCSXR again and found that both projects only assign to Baud in one place, and in PCSX2's case it is _always_ assigning 0x88, which is the SIO command that is sent in. Here's the section of IopHwWrite.cpp that handles all the SIO registers:

image

The nocash docs suggest there is a formula for computing the baud rate, but the registers involved seem to never change their values, which goes against the idea of the Baud register being a timer. And just looking at these assignments, I question how exactly these registers are supposed to update, since the command value is (essentially) just getting slung into them every time.

What I have done here is applied the same BaudReg * 8 that PCSXR uses for the SIO interrupts. I also am printing out all the parts that nocash says are factors in the Baud rate, such that they only print if they change from the last interrupt that ran. Nothing crazy, and it doesn't break anything (in my collection) that already worked.

The entire point here is to try and get Silent Hill working again, which I do not own but CK and Nobbs pointed out to me as broken after the SIO interrupt fix. The tricky thing is, this may not be a Baud problem but some other area. As a start though, I would like to know what happens under the metal in Silent Hill when it hangs. If it hangs on an interrupt, or if the registers are changing (which as is, does not happen at all in any working game I have) when it hangs would be two notable reactions.

### DIFF
--- a/pcsx2/Sio.cpp
+++ b/pcsx2/Sio.cpp
@@ -120,13 +120,26 @@ void sioInit()
 
 bool isR3000ATest = false;
 
+u16 currentStat = 0, lastStat = 0, currentMode = 0, lastMode = 0, currentBaud = 0, lastBaud = 0;
+
 // Check the active game's type, and fire the matching interrupt.
 // The 3rd bit of the HW_IFCG register lets us know if PSX mode is active. 1 = PSX, 0 = PS2
 // Note that the R3000A's call to interrupts only calls the PS2 based (lack of) delays.
 SIO_FORCEINLINE void sioInterrupt() {
 	if ((psxHu32(HW_ICFG) & (1 << 3)) && !isR3000ATest) {
-		if (!(psxRegs.interrupt & (1 << IopEvt_SIO)))
-			PSX_INT(IopEvt_SIO, 64); // PSXCLK/250000);
+		if (!(psxRegs.interrupt & (1 << IopEvt_SIO))) {
+			currentStat = sio.StatReg & 67106816; // Mask out bits 11-25
+			currentMode = sio.ModeReg & 3; // Mask out bits 0-1
+			currentBaud = sio.BaudReg; // The entire baud reg
+			// If anything changes, print.
+			if (currentStat != lastStat || currentMode != lastMode || currentBaud != lastBaud)
+				DevCon.WriteLn("%s // StatReg (bits 11-25) = %08X // ModeReg = %08X // BaudReg = %08X", __FUNCTION__, currentStat, currentMode, currentBaud);
+			lastStat = currentStat;
+			lastMode = currentMode;
+			lastBaud = currentBaud;
+			
+			PSX_INT(IopEvt_SIO, sio.BaudReg * 8); // PSXCLK/250000);
+		}
 	} else {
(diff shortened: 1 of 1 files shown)


atomic83GitHub:
This PR can affect ps2 games or it's psx only?

RedPanda4552:
Any changes will be targeted at PSX only, but we should keep an eye out for any bleeding effects.

pcsx2/Sio.cpp
```diff
@@ -120,13 +120,26 @@ void sioInit()
...
+
 // Check the active game's type, and fire the matching interrupt.
 // The 3rd bit of the HW_IFCG register lets us know if PSX mode is active. 1 = PSX, 0 = PS2
 // Note that the R3000A's call to interrupts only calls the PS2 based (lack of) delays.
 SIO_FORCEINLINE void sioInterrupt() {
 	if ((psxHu32(HW_ICFG) & (1 << 3)) && !isR3000ATest) {
-		if (!(psxRegs.interrupt & (1 << IopEvt_SIO)))
-			PSX_INT(IopEvt_SIO, 64); // PSXCLK/250000);
+		if (!(psxRegs.interrupt & (1 << IopEvt_SIO))) {
+			currentStat = sio.StatReg & 67106816; // Mask out bits 11-25
```
ssakash: **Note:** Personally I prefer hexadecimal for bitwise operations like this, seems more readable.  This could be ``0x3FFF800``which stands out as ``0011 1111 1111 1111 1000 0000 0000`` showing which bits you fetch easily without the need of comments. Whereas, ``67106816`` makes us pop up a calculator or just hope the author's comment is correct.

pcsx2/Sio.cpp
```diff
@@ -120,13 +120,26 @@ void sioInit()
...
+			currentMode = sio.ModeReg & 3; // Mask out bits 0-1
+			currentBaud = sio.BaudReg; // The entire baud reg
+			// If anything changes, print.
+			if (currentStat != lastStat || currentMode != lastMode || currentBaud != lastBaud)
```
ssakash: I'm not sure how likely it is to get 0 value on all of these bitfields, if such a scenario arises then the message won't be printed. Maybe ``-1`` could be a better default value (of course also changing it to signed in that case).

atomic83GitHub:
It seems that "Fear effect" goes a little further but is still stuck at a black screen.

pcsx2/Sio.cpp
ramapcsx2: Yeah, definitely use something readable here please.

PR #2284, opened 2018-01-29
pcsx2: Fix copy byte patch command

turtleli:
This adjusts the copy bytes command format from
5aaaaaaa nnnnnnnn
00000000 bbbbbbbb
to
5aaaaaaa nnnnnnnn
bbbbbbbb 00000000
so that it matches the copy bytes command format used by PS2 cheat
devices (GS/CB/XP/AR2).

**Note**: This isn't really my PR. A user emailed me this, and I think it makes sense. A portion of the email is below:
> I was trying to create a copy code for a ps2 game on pcsx2 emulator and it wasn't working, so I tested it on my old PS2 console with codebreaker and my code worked just as I intended.
That made me curious what pcsx2 was doing differently, so I looked into the source and found the "problem". later I found a topic where the same problem was addressed some years ago and nicely summarized: https://forums.pcsx2.net/Thread-About-the-implementation-of-type-5-cheat-code?highlight=5aaaaaaa
> pcsx2 src@Jan2018   vs   Codemasters Project Description (  
http://www.codemasters-project.net/portal-english/apportal/cmp_plugins/content/content.php?content.23  
)
(commit:ab44ebd)    vs   (Standard Formats (CB/GS/XP/AR2) / CB7+)
5aaaaaaa nnnnnnnn   vs   5aaaaaaa nnnnnnnn
00000000 bbbbbbbb   vs   bbbbbbbb 00000000
a = Address to copy from
b = Address to copy to
n = Number of bytes to copy

### DIFF
--- a/pcsx2/Patch_Memory.cpp
+++ b/pcsx2/Patch_Memory.cpp
@@ -74,11 +74,11 @@ void handle_extended_t(IniPatch *p)
 		PrevCheatType = 0;
 		break;
 
-	case 0x5000: // dddddddd iiiiiiii
+	case 0x5000: // bbbbbbbb 00000000
 		for (u32 i = 0; i < IterationCount; i++)
 		{
 			u8 mem = memRead8(PrevCheatAddr + i);
-			memWrite8(((u32)p->data) + i, mem);
+			memWrite8((p->addr + i) & 0x0FFFFFFF, mem);
 		}
 		PrevCheatType = 0;
 		break;


bositman:
Will implementing this break all existing patches?

turtleli:
No. The widescreen patches don't use it, and from what I can find in the forums, anyone who has attempted to use this type of patch (almost noone) has tried to use:
```
patch=1,EE,5aaaaaaa,extended,nnnnnnnn
patch=1,EE,bbbbbbbb,extended,00000000
```
just like the creator of this patch did.

pcsx2/Patch_Memory.cpp
```diff
@@ -74,11 +74,11 @@ void handle_extended_t(IniPatch *p)
 		PrevCheatType = 0;
 		break;
 
-	case 0x5000: // dddddddd iiiiiiii
+	case 0x5000: // bbbbbbbb iiiiiiii
 		for (u32 i = 0; i < IterationCount; i++)
 		{
 			u8 mem = memRead8(PrevCheatAddr + i);
-			memWrite8(((u32)p->data) + i, mem);
+			memWrite8(((u32)p->addr & 0x0FFFFFFF) + i, mem);
```
turtleli: Actually, I wonder whether it should be `memWrite8(((u32)p->addr + i) & 0x0FFFFFFF, mem);`

FlatOutPS2:
> anyone who has attempted to use this type of patch (almost noone) has tried to use...

Well, of course they did. It's the correct order. There are quite a few gameshark/AR enthusiasts, so it's surprising nobody changed this before.

Did a quick test and it works great. 👍

yxmline:
This does not need to convert GS/CB/XP/AR2 to XP/RAW ？？

FlatOutPS2:
> Does not need to convert GS/CB/XP/AR2 to XP/RAW??

You still need to convert the actual codes. This just corrects an issue with the order of an address and value.

Issue #1981, opened 2017-06-25
[Request improvement] DVD Movies Compatibility


atomic83GitHub:
It would be nice to brings DVD Movies on pcsx2 , actualy , play a DVD Movie on pcsx2 cause a black screen (but no crash)

Farmerjoe12:
Why would you watch movies through pcsx2, when you're on a computer, likely with a CD drive, and the ability to download any other software (VLC)

lightningterror:
The feature would have been nice if pcs didn't have the ability to watch movies xD

Anyway it's not even worth putting a git issue on this.

atomic83GitHub:
I assume its a litle useless but it should be fun to play DVD Movies like the real ps2 :)

Blackbird88:
It's not out of the question since even Dolphin implemented unused IOS feature, but the thing is they have the manpower to do anything they want lol
If someone was determined enough it could probably be done.

atomic83GitHub:
After some reflections , I close this request and will reopened it when the most important improvement/issues are merged.

mirh:
PS2 could do it, so it's absolutely within scope of the project. 
Then, depending on where CSS keys were stored on original hardware I guess things might be easier or harder. 

Also depends on #432

Phantaztique:
Would implementing this improve performance on FMV playback?

MrCK1:
FMV playback is handled by the IPU. What performance is there to gain lol?

willkuer:
It could be interesting for framerates and output resolutions. Also it might help regarsing general disk management (i.e. might help resolving disk read and swap issues). Also who knows. Maybe some non-bootable games use a bios routine that is typically only used by dvd's. In general improving the accuracy of the emulator helps improving the accuracy of the emulator. Emulation of the psx on the emulated ps2 might also solve some unresolved issues.

lightningterror:
I'll close this since it's really low priority. I don't think anyone will be looking in to this for a long time.

mirh:
It's a feature missing of a PS2?

You could close even psx and optical audio issues then with this attitude

lightningterror:
You could improve the emulator in a lot of different ways, the problem is we lack the manpower to do so, which is why I closed it. Improving accuracy on games for both psx and ps2 takes higher priority. 

Plus the percentage of this being used by people is very low which is why it's on the lowest priority.
If someone wants they can work on it if they want to.

turtleli:
[...]

PR #2285, opened 2018-01-30
 cdvdgigaherz:windows: Avoid unneeded DVD ioctls

turtleli:
Avoid using the IOCTL_DVD_START_SESSION and IOCTL_DVD_END_SESSION ioctls - it's not necessary to obtain an Authentication Grant ID (AGID) before requesting the DVD physical format layer descriptor.

### DIFF
--- a/plugins/cdvdGigaherz/src/Windows/IOCtlSrc.cpp
+++ b/plugins/cdvdGigaherz/src/Windows/IOCtlSrc.cpp
@@ -197,13 +197,6 @@ bool IOCtlSrc::ReadSectors2352(u32 sector, u32 count, u8 *buffer) const
 bool IOCtlSrc::ReadDVDInfo()
 {
     DWORD unused;
-    DVD_SESSION_ID session_id;
-
-    BOOL ret = DeviceIoControl(m_device, IOCTL_DVD_START_SESSION, nullptr, 0,
-                               &session_id, sizeof(session_id), &unused, nullptr);
-    if (!ret)
-        return false;
-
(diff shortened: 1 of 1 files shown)


willkuer:
Does it hurt to obtain it? Is this a performance improvement or do these calls arbitrarily fail and you want to increase compatibility?

turtleli:
Well, I originally thought those ioctls were some strange Windows DVD requirement, but when I checked the specs I found out that they're used for dealing with content protection and that it is not necessary to use them in this case.

There won't be any performance improvement (called once per disc). It might increase compatibility with virtual drives if they aren't able to handle these ioctls but can still handle IOCTL_DVD_READ_STRUCTURE.

mirh:
Kinda remembers me of #652

> they're used for dealing with content protection

Anyway, couldn't this be counterproductive to #1981?

turtleli:
No.

FlatOutPS2:
Some reference for someone reading this PR in the future: https://forums.pcsx2.net/Thread-Bug-Report-cdvdGigaherz-plugin-error

The PR has been confirmed to work in the bug report.

ssakash:
> Some reference for someone reading this PR in the future: https://forums.pcsx2.net/Thread-Bug-Report-cdvdGigaherz-plugin-error

Nice reference you got there.

bositman:
You should post the link after you move the thread with the moderator tool, the redirect expires after 1 day. Fixed link: https://forums.pcsx2.net/Thread-Resolved-Bug-Report-cdvdGigaherz-plugin-error

PR #2286, opened 2018-01-30
PCSX2: Avoid SYSTEM.CNF error message in PSX mode

FlatOutPS2:
Avoid error on non-empty last line of file.


ramapcsx2:
Yep, that's good!

Issue #2219, opened 2017-12-13
[GSdx HW] Dragon Ball Z Budokai Tenkaichi 2 and 3 - Character reflection textures


Stickend:
***PCSX2 version:***
v.1.5.0-dev-2237-g41051f3cf

***PCSX2 options:***
Everything by default.

***Plugins used:***
GSDX, Lilypad, SPU2-X, Rest Default null plugins.

***GSDX configuration:***
Resolution 5x Native
OpenGL Hardware Mode
Accurate Date
Blending Full
CRCHack None
Mipmapping Automatic

***HW Hacks:***
Trilinear
TC Offset X 440
TC Offset Y 420

***Description of the issue:***
With current settings, emulation is almost identical to PS2. But there is a bug with textures which are supposed to reflect light.

![cell mal](https://user-images.githubusercontent.com/34516283/33950323-64e10faa-e02c-11e7-8c1a-7f2e310f94b5.jpg)

If I go into HW Hacks and disable Depth Emulation, these textures get corrected, but ghost effect appears. I try to correct it with Skipdraw at 5. It partially works, but either way disabling Depth Emulation destroys the natural lighting of maps. (It also affects to the "heat" effect of the Desert-Day map, which it's missing)

Without Skipdraw

![cell bien](https://user-images.githubusercontent.com/34516283/33950337-6db2b1ce-e02c-11e7-9f07-9b0b540ccc52.jpg)

With Skipdraw 5

![cell bien 2](https://user-images.githubusercontent.com/34516283/33950349-72abfcda-e02c-11e7-9125-881729faf532.jpg)

Is there any way of correcting these textures without disabling Depth Emulation or sacrificing anything as the natural lighting? Realize that dinamic lighting isn't affected, as it works fine with all of these settings.

***How it should be (SW Mode):***

![cell mal sw](https://user-images.githubusercontent.com/34516283/33950357-78ee7208-e02c-11e7-93cd-9d6b01119ee2.jpg)

***How to reproduce the issue:***
Textures: choosing any character with reflexive textures (Cell, Freezer)
Lighting: Disable Depth Emulation. It affects to every map.

***PC specifications:***
CPU: Intel Core i7-4790 3.60Ghz
GPU: Nvidia GTX 970
OS: Windows 10
RAM: 16GB

I use the ***NTSC*** version of DBZBT3

MrCK1:
Why are you using hardware hacis to begin with? Those are probably causing more issues.

[...]

PR #2290, opened 2018-02-04
GSdx HW: Add Skipdraw Offset option for Skipdraw hack

FlatOutPS2:
Adds an option called "Skipdraw Offset" to the GSdx Advanced Settings and Hacks submenu.

For example: When the broken effects are at frames 91-100, the default skipdraw always skips 0-100, possibly skipping several functioning effects as well. By adding this new option, it is now possible to skip just frames 91 - 100.

For the example given above use Skipdraw Offset = 90, and Skipdraw = 10.

See: https://forums.pcsx2.net/Thread-Extended-skipdraw


lightningterror:
> Nice. I always wondered whether this is possible. This could potentially fix a huge amount if glitches without much drawbacks.

Even less of a reason to add game specific crc hacks.

willkuer:
Btw this would make mirh's idea of game specific gsdx hacks almost mandatory (except for you consider 3rd party frontends to be a solution).

FlatOutPS2:
> Even less of a reason to add game specific crc hacks.

Users can now pinpoint broken effects easier, which makes it easier to add a new CRC hack. We don't want users to have to manually add skipdraw for each game.

mirh:
>  (except for you consider 3rd party frontends to be a solution).

In the most theoretically pure form, my idea is by no means exclusive to them. 
You should just name hacks (in game database) around "concepts to be fixed" rather than "actual implementation-in-plugins name". 

Besides, CRC hacks still have plenty of room to exist. I mean, for one I'm not sure how many game-specific options you'd want in gsdx or whatever else. 
Also, at least until we don't figure out some "extension of logic", conditionals and stuff like that still requires to sits in some .cpp

willkuer:
If I saw correctly the offset is possibly slightly awkward with old settings. Somehow if I set skipdraw to 5 to fix an issue and now see this nice offset I would check what happens with offset 1, 2, ... expecting that I would use skipdraw 1..5, 2..5, 3..5 but instead it will be 1..6, 2..7, 3..8. (Or did I misinterpret the code changes?)

bositman:
Skipdraw 5 with offset 1 would be 5-6, Skipdraw 5 with offset 2 would be 5-7 etc AFAIK

lightningterror:
> Skipdraw 5 with offset 1 would be 5-6, Skipdraw 5 with offset 2 would be 5-7 etc AFAIK

The way i see it skipdraw 5 with offset 1 would be 2-5
Skipdraw 5 with offset 2 would be 3-5

refractionpcsx2:
that's confusing then, if offset was 6 and skipdraw was 5, I'd expect it to skip 6-11 as in my mind you are skipping 5 frames from frame 6 (the offset)

FlatOutPS2:
> that's confusing then, if offset was 6 and skipdraw was 5, I'd expect it to skip 6-11 as in my mind you are skipping 5 frames from frame 6 (the offset)

That is what the PR currently does.

refractionpcsx2:
Now I am confused lol, you just said lightningterror's example of doing that would do nothing xD

FlatOutPS2:
> Now I am confused lol, you just said lightningterror's example of doing that would do nothing xD

I misread. The current PR with Offset = 6 does the first **6** draws as usual. Your example would only draw the first **5** draws as usual with offset 6.

refractionpcsx2:
You're totally losing me now.
So when you say the offset is 6, it does the first 6 draws as usual, but when I say the offset is 6 it only does the first 5 draws as usual?

OK lets get this clear, say in the current PR (I'm not talking about anything else btw) I set the offset to 6 and skipdraw to 5, what EXACTLY will be drawn and what won't be?

mirh:
I hope somebody will make a blog post about this whole affair (who knows, perhaps together with automatic hacks 🙃)

FlatOutPS2:
> OK lets get this clear, say in the current PR (I'm not talking about anything else btw) I set the offset to 6 and skipdraw to 5, what EXACTLY will be drawn and what won't be?

The first 6 will be drawn, the next 5 (7-11) will be skipped.

refractionpcsx2:
And in the original incarnation, it skipped from "skipdraw" up to 100?

I still think it would be better to have a range rather than a skip and offset mess, so you can type in from 7 to 11 rather than having to work out the offset-1 and how many you need to skip. Remember idiots have to use this :P

willkuer:
I think the workflow of adjusting skipdraw is:

1. I see a glitch in game xyz.
2. I enable hw and increase the last skipped draw step by step until the glitch is gone.
3. I increase the threshold for the first skipped draw just before the issue appears again

Step 2 requires that skipdraw is increased stepwise.

Step 3 requires currently that skipdraw is decreased by one and skip offset is increased by one. That's kind of not so nice from a usability pov.

lightningterror:
I totally got the wrong impression how this works.

FlatOutPS2:
> I think the workflow of adjusting skipdraw is:

The workflow should be.

1. Set skipdraw to 1.
2. Increase offset is necessary till you find the bad draw(s).
3. If more than 1 draw needs to skipped, increase skipdraw.

Stickend:
I'm not completely sure if this is gonna work right in the #2219 issue which I posted long time ago. In the mere moment I'd apply the Disable Depth Emulation code, it would correct reflective textures, but it will "destroy" the sky textures and dynamic lighting. How am I supposed to use this new code instead? Skipdraw code can't be used in Dragon Ball Budokai Tenkaichi 2 and 3, there is no way right now.

willkuer:
But what if bad draws happen at draw 4 and 6? In your suggestion I'd use skipdraw 1 and offset 3 or 5 (5 iterations) which both kind of fix a bit and then need to test skipdraw 2 with 2,3,4, or 5 (another 5 iterations) which all kind of fix a bit and then I finally set skipdraw to 3 and offset to 3 (another 3 iterations) to get the perfect setting.

In my suggestion you could set skipdraw to 5 to fix the glitch (5 iterations) and lower limit to 3 (3 iterations).

So it will be 13 against 8 iterations. Not a huge gain but you also kind of need less thinking imo.

But do as you like. Having the feature is already a huge improvement.

FlatOutPS2:
In my suggestion you set skipdraw to 1(first iteration), then increase offset till it gets 3(another 3 iterations). Then because the rendering is still broken, but 1 bad draw is already fixed, increase skipdraw to 3(last 2 iterations) to get the second bad draw. and you would be done with 6 iterations...

Stickend:
Okay then, I'll test it when the code is ready. I'm still not sure if I completely understand how this work, but I'll test it anyways. I hope this finally solve the half bottom screen issue in Budokai Tenkaichi 2/3 when Disable Depth Emulation is active. Otherwise there wouldn't be any other solution to this.

willkuer:
You win 👻

FlatOutPS2:
Changed it to a range as mentioned by @refractionpcsx2 .

Needs UI behaviour update for Linux.

mirh:
Also, strings to translate I guess.

FlatOutPS2:
Also, strings to translate I guess.

No, plugin strings don't get translated...

refractionpcsx2:
Is this ready to be merged now?

FlatOutPS2:
> Is this ready to be merged now?

No. I need to make a few changes.

refractionpcsx2:
OK no problem :)

refractionpcsx2:
Are we good to go now?

FlatOutPS2:
It needs to be reviewd first.

plugins/GSdx/GSSetting.cpp
```diff
@@ -57,9 +58,16 @@ const char* dialog_message(int ID, bool* updateText) {
...
-			return "Skips drawing n surfaces completely. "
-				"Use it, for example, to try and get rid of bad post processing effects."
-				" Try values between 1 and 100.";
+			return "Completely skips drawing up to the surface specified here from the surface specified in the box to the left.\n\n"
+				"Use it, for example, to try and get rid of bad post processing effects.\n"
+				"Increase this value and keep the value in the box next to this set to \"1\" to keep the same skipdraw behaviour as before Skipdraw Range was introduced.\n"
+				"Try values between 1 and 100.";
+		case IDC_SKIPDRAWOFFSET:
+		case IDC_SKIPDRAWOFFSETEDIT:
+			return "Completely skips drawing surfaces from the surface specified here up to the surface specified in the box to the right.\n\n"
```
turtleli: Is there any way to use the same message for both boxes?

FlatOutPS2: Done.

lightningterror:
What's the status on this ?

FlatOutPS2:
> What's the status on this ?

@turtleli needs to review it (again). I'm not sure the first comment was part of complete review.

lightningterror:
What about changes to the linux gui, nothing needs changing there ?
Just making sure.

FlatOutPS2:
> What about changes to the linux gui, nothing needs changing there ?
Just making sure.

I didn't add the Linux GUI changes here. I don't have a Linux build environment to check if it behaves correctly.

hellbringer616:
@FlatOutPS2 
I am currently using the latest version of Ubuntu; if you or someone else can post a binary i should be able to test it.

Quick GUI tests are a hall of a lot easier than trying to find time to game for a few hours :P

lightningterror:
Something like this maybe ? I can't test either so not sure.
https://github.com/lightningterror/pcsx2/commit/b621a94ec153ad567f2b43cd7b703771765f66b1

FlatOutPS2:
> I am currently using the latest version of Ubuntu; if you or someone else can post a binary i should be able to test it.

I'd really need to have a built environment to create a proper Linux version. So there's no use in just posting a binary. If the Travis CI(linux) build came with a binary, I'd be able to test it myself as well.

@lightningterror: No.

hellbringer616:
I'm guessing since you haven't set one up it's painful to do so?
if we have any documentation i don't mind setting one up; though i guarantee it won't me timely. 
But it's better than nothing.

FlatOutPS2:
> I'm guessing since you haven't set one up it's painful to do so?

It's usually not that painful. There's a guide at the wiki: https://github.com/PCSX2/pcsx2/wiki/Installing-on-Linux
I just don't have the harddrive space to add a Linux VM/dual boot.

plugins/GSdx/GSHwHack.cpp
```diff
@@ -2469,25 +2470,31 @@ bool GSState::IsBadFrame()
...
 			{
-				m_skip = m_userhacks_skipdraw;
+				m_skip_offset = std::max(m_userhacks_skipdraw_offset, 1);
+				m_skip = std::max(m_userhacks_skipdraw, m_skip_offset);
 			}
 		}
 	}
 
-	if(m_skip > 0)
+	if(m_skip_offset == 1 && m_skip > 0)
```
turtleli: Breaks CRC hacks?

plugins/GSdx/GSSetting.cpp
```diff
@@ -1,4 +1,5 @@
 /*
+ *	Copyright (C) 2016-2018 PCSX2 Development Team
```
turtleli: Copyright dates probably not correct.

gregory38:
Can I have a screenshot of the GUI change ? So we can do the same on Linux.

lightningterror:
image

mirh:
Ssso.. Are we just missing a linux GUI to land this?

turtleli:
> Are we just missing a linux GUI to land this?

Well, that's my main reason for not merging this.

turtleli:
Closing since the other PR was merged.

PR #2292, opened 2018-02-04
pcsx2: Fix trace logging behaviour and dialog

turtleli:
Fixes the following issues in the Trace Logging dialog:
 * The apply button does not enable when something has changed.
 * The checkboxes may appear to be incorrectly enabled.
 * The misc section is not disabled if the master trace logging checkbox is unchecked.

Fixes the following trace logging behaviour:
 * SIF logging is not disabled if the master trace logging switch is disabled.

Note: Only testable with devel/debug builds.


refractionpcsx2:
It's about time somebody fixed this, I keep noticing it and then pretending nothing was wrong :P

FlatOutPS2:
I remember running into the wrong apply button behaviour several times. Good to see it fixed.

lightningterror:
Same here.

bositman:
Anything else to check here? Else let's merge it

PR #2298, opened 2018-02-06
PCSX2: Remove obsolete "DisableOutput" option from core/GUI

MrCK1:
It's been broken/unused for some time.

### DIFF
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -295,7 +295,7 @@ struct Pcsx2Config
 		// forces the MTGS to execute tags/tasks in fully blocking/synchronous
 		// style.  Useful for debugging potential bugs in the MTGS pipeline.
 		bool	SynchronousMTGS;
-		bool	DisableOutput;
+
 		int		VsyncQueueSize;
 
 		bool		FrameLimitEnable;
@@ -318,7 +318,6 @@ struct Pcsx2Config
 		{
 			return
 				OpEqu( SynchronousMTGS )		&&
-				OpEqu( DisableOutput )			&&
 				OpEqu( VsyncQueueSize )			&&
 				
 				OpEqu( FrameSkipEnable )		&&
--- a/pcsx2/MTGS.cpp
+++ b/pcsx2/MTGS.cpp
@@ -297,11 +297,6 @@ void SysMtgsThread::ExecuteTaskInThread()
 		// ever be modified by this thread.
 		while( m_ReadPos.load(std::memory_order_relaxed) != m_WritePos.load(std::memory_order_acquire))
 		{
-			if (EmuConfig.GS.DisableOutput) {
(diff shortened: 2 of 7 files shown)


gregory38:
```
It's been broken/unused for some time and is superseeded by the "Core Benchmark" option that exists internally, inside GSdx.
```
There aren't the same option.

This option removes all traffics between GSdx and the core (so basically MTGS is idling).
The GSdx option is a null renderer, so you still emulate GS register, texture transfers. To be fair, remaining activity should be light.

MrCK1:
Ah OK, didn't realize they were actually different options, they seemed like the same thing. Guess I'll close this then :P

gregory38:
Honestly if the core option is broken. I'm not sure it is worth to keep it.

FlatOutPS2:
Is it really broken though?

gregory38:
Maybe we could do a test to check if there is any real difference between the options.

MrCK1:
I tried it earlier, the comment around the GUI option in Video panel says the same thing. 

Another caveat is that this option doesn't work on the fly like GS Null does, it reauires a restart of something (can't quite recall right now)

I'll have to try again in a few hours, but I didn't find anything of use personally.

turtleli:
Seems to infinite loop in CopyGSPacketData() when the disable output option is used.

MrCK1:
@turtleli Do you still want to keep this open? I can change the commit message if necessary.

refractionpcsx2:
@turtleli

turtleli:
Yeah, change the commit message so it doesn't mention the superseded stuff.

FlatOutPS2:
These travis build failures not caused by coding mistakes have been quite common lately. I had one yesterday too. This PR even has 2 travis builds fail without finishing.

lightningterror:
I'd still do a rebase until it compiles property.
Having big red X marks on merged PRs doesn't look good.

FlatOutPS2:
You could add "It's been broken/unused for some time." to the commit message instead of the title to make the push worthwhile. :)

turtleli:
I think you both have the correct permissions to restart the affected Travis CI jobs if they fail?

lightningterror:
Yeah we can, that did the trick.

turtleli:
It looks like you restarted the entire build. Only the individual jobs that failed need to be restarted. It consumes less resources that way.

lightningterror:
Yeah, I noticed that I can do that after I restarted the entire build.

pcsx2/Config.h
```diff
@@ -318,7 +318,6 @@ struct Pcsx2Config
 		{
 			return
 				OpEqu( SynchronousMTGS )		&&
-				OpEqu( DisableOutput )			&&
```
turtleli: I think you still need to remove the DisableOutput variable.

PR #2299, opened 2018-02-06
GameDB: Update Silent Hill 3

FlatOutPS2:
Adds vuClampMode = 2

This fixes some lighting/shadow artefacts. Most light sources like the Flashlight are unaffected and don't cause any artefacts.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -9668,8 +9668,8 @@ Compat = 4
 Serial = SLES-51434
 Name   = Silent Hill 3
 Region = PAL-M5
-Compat = 5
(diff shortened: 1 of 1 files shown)


atomic83GitHub:
@FlatOutPS2, is there a place especially where the issue happen? I can test this if you want :) (my version is NTSC-J)

FlatOutPS2:
@atomic83GitHub A clear example is in the alley right at the beginning of the game (the issue will trigger depending on which direction the character faces):

sh3

atomic83GitHub:
Yes, I am able to reproduce this, nice find!

refractionpcsx2:
Good job :)

PR #2302, opened 2018-02-11
GameDB: VU rounding fixes for Guitar Hero series and Hitman Contracts.

MrCK1:
Guitar Hero series: Fix cut-off numbers and restore missing whitespace in combo meter

Hitman Contracts: Fix missing light shafts under lamps

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -15496,6 +15496,7 @@ Serial = SLES-54132
 Name   = Guitar Hero
 Region = PAL-M5
 Compat = 5
+vuRoundMode = 2 // Fixes cut-off numbers and restores missing whitespace inside combo meter
 ---------------------------------------------
 Serial = SLES-54135
 Name   = Grand Theft Auto - Liberty City Stories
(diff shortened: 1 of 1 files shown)


atomic83GitHub:
The original Guitar Hero also need VU rounding to positive as you can see.

With Chop/zero:
capture

With positive:
capture2

MrCK1:
@atomic83GitHub What region are you using?

atomic83GitHub:
The region is Pal-M5

MrCK1:
Just making sure, is it `SLES-54093`?

atomic83GitHub:
No, it's SLES-54132

FlatOutPS2:
If someone can confirm this is also needed for a PAL version of GH2/GHEncore and NTSC version of GH, then it should be safe to include it for all regions instead of just the one.

MrCK1:
I don't own the PAL version, but it's highly likely to be needed in all regions :P

ssakash:
So let's merge it (or) wait for confirmation on the PAL version?

FlatOutPS2:
Wait for confirmation. Like @MrCK1 said it's most likely to be required for all regions.

FlatOutPS2:
@MrCK1 Did you check if someone reported this issue for other regions on the forum or wiki (or maybe even youtube video)? Then we could add the other regions with a mention of which region was confirmed.

MrCK1:
I just checked the compatibility list. The European versions will need the hack as well - I'm looking for other regions as we speak ;)

MrCK1:
I added the additional regions for the various Guitar Hero games. Hitman Contracts is untouched as there aren't any reliable videos on YouTube.

PR #2303, opened 2018-02-11
GSdx: Adjust Ikkitousen CRC hacks

lightningterror:
Move hacks that disabled shadows to DX level since OpenGL renders
shadows properly with Depth Emulation.

Some other upscaling issues appear with the disabled hack like a small
black border on the bottom of the screen or some ui elements but those
can be fixed with TC X,Y Offset hack.

GS Dump for testing comparison (rename to .xz if needed).
[IkkiTousen.gs.zip](https://github.com/PCSX2/pcsx2/files/1714439/IkkiTousen.gs.zip)

### DIFF
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -217,16 +217,16 @@ bool GSC_Spartan(const GSFrameInfo& fi, int& skip)
 		}
 		else
 		{
-				if(fi.TME)
+			if(fi.TME)
+			{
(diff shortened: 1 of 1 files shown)


ssakash:
For future pull requests, I'd prefer if the code formatting changes are separated into a different commit (or) be mentioned about on the original commit message.

lightningterror:
Sure, I thought small code formatting wasn't important since it doesn't change anything in the code so I didn't mention it.

PR #2304, opened 2018-02-11
GSdx: Remove GT4/Tourist Trophy CRC hacks

FlatOutPS2:
Removes GT4/Tourist Trophy CRC hacks. The hack had already been moved to
aggressive due to VRAM spike issues, but is no longer necessary at all
due to the in-game brightness/contrast setting issue being moved to
behind the frame buffer conversion hack for Direct3D and being resolved for OpenGL.

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -147,14 +147,14 @@ CRC::Game CRC::m_games[] =
 	{0x7ABDBB5E, GT3, CH, 0}, // cutie comment
 	{0x3E9D448A, GT3, CH, 0}, // cutie comment
 	{0xAD66643C, GT3, CH, 0}, // cutie comment
-	{0x6810C3BC, GT3, CH, 0}, //GRAN TURISMO Concept 2002 Tokyo-Geneva
+	{0x6810C3BC, GT3, CH, 0}, // Gran Turismo Concept 2002 Tokyo-Geneva
 	{0x85AE91B3, GT3, US, 0},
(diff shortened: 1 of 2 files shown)


lightningterror:
I wanted to do this a long time ago for GT4.

refractionpcsx2:
The vram spikes still happen as far as I know, once you hit x4 native it murders your pc

lightningterror:
I just tried 4x native on ogl and dx. max vram usage around 1500MB, average ~1100MB.
Maybe there is a small spike(I'm not sure) but the crc hack makes things worse, when you start races with aggressive crc and brightness/contrast 0 you can see it freezing up and stuttering with maxing out my 2GB fram.

FlatOutPS2:
> The vram spikes still happen as far as I know, once you hit x4 native it murders your pc

The vram spikes weren't caused by the patch, but they were exacerbated by it. You can still crash PCSX2 with the wrong GSdx settings though.

@lightningterror You must've switched off Large Framebuffer otherwise it should've had a bigger spike.

lightningterror:
Yeah I had it switched off, they are much bigger with it.

lightningterror:
GT3 and GT Concept might also be worth checking.

refractionpcsx2:
GT3 is a different engine, I believe it doesn't suffer from VRAM spikes.

lightningterror:
Game sequels are usually made on the same(improved) engine.
The code is basically the same comparing it to GT4.

I'll look in to it maybe next week.

Issue #268, opened 2014-08-21
Unreal Tournament (NTSC-U) [SLUS-20034] Game Freezes


shandawgy:
In ether hardware or software modes when u begin in game or first mission the game will freeze and hang.

My pcsx2 version is:
pcsx2-v1.2.1-511-g02b19b7

Plugins:
GSDx 20140818023333(MSVC 18.00,AVX) 0.1.16 [GSdx32-AVX]
XPad 0.2.0
SPU2-X 2014081802333 2.0.0
cdvdGigaherz (20140818023333ll) 0.9.0

My specs are:
Intel core i5-3470T CPU @ 2.90GHz
2x NVidia 560 GForce graphics card SLI
8gb Ram
Windows 8 64bit

refractionpcsx2:
Does this still happen on new GIT builds?

yoshijin:
On my setting the game boots fine and also able to play the first deathmatch. The game hangs after the match ends and there is onscreen prompt to press the "Fire" button. Hangs immediately on pressing fire button OR on its own shortly after. Have been trying latest git builds and also latest v1.4.0
Tried switching HW/SW modes on gsdx on openGL and DX11 with MTVU off. Lastly also tried messing with the VU legacy modes.

joder666:
This is a SPU-X issue. Depending on the module chosen the game will either freeze before/after a match (Xaudio) or have no music (portaudio).
DirectSound and waveout are woking OK.

yoshijin:
Okay then I'll report back after checking.

yoshijin:
Tried DirectSound and waveout modules but the game still froze. Even tried with No Audio (Emulate SPU2 only).
If it is working for you what exact spu settings you used.

joder666:
Ummm, i got v1.5.0-dev-61-ga6eb871 from orphis buildbot and the only changes i did where to renderer to opengl hardware (crcr hack level partial) and the output module of spux, nothing else.

joder666:
I only change the output module of spux after encounter the problem, everything else for spux i left untouched.
Also i used build v1.5.0-dev-61-ga6eb871 from http://buildbot.orphis.net/pcsx2/

yoshijin:
Didn't work. Still freezes after the match is over.

yoshijin:
Can you please confirm it doesn't happen with you after the end of the match and that you are able to proceed to next match.

joder666:
[...]

Issue #2233, opened 2017-12-26
Meta: SPU2-X Delay Cycle


atomic83GitHub:
***This is a meta topic***

***Description of the issue:***
The SPU2-x plugin have the Delay Cycle option set to 4, this solve some games, but break some other:

-Higurashi no Naku Koro ni Matsuri Kakera Asobi (#918)

-Deus Ex: The Conspiracy (#615)

-Armored Core 2

-Unreal Tournament (#268)

They need the Delay Cycle option set to 1.

***Way to fix this***

-Make a gui option like this:

[Delay Cycle hack: set the Delay Cycle value to 1. Fix some missing/broken voices (Note: Unreal tournament will hang without this hack!)]

***Make this before the 1.6 release would be very nice :)***

MrCK1:
We've already discussed a GUI option for DelayCycles. It's a hack and won't be added to the SPU2-X GUI as such. You'll have to resort to editing the .ini for the small amount of games that are affected by this.

MrCK1:
I'm going to use this because all of the games listed have already their own issues open. There's not enough cases here to warrant a meta issue either.

PR #2315, opened 2018-02-18
SPU2-X: Updates

FlatOutPS2:
Fixes 2 regressions:
- Audio cutting out in Wallace And Gromit: Curse of the Were-Rabbit
- AudioExpansion Mode not available for XAudio2 when using a Windows version before 8.

New features:
- Adds automatic fix for SPU2-X delay cycles
- Windows UI options for advanced volume controls and Dolby Pro Logic settings.

### DIFF
--- a/plugins/spu2-x/src/DplIIdecoder.cpp
+++ b/plugins/spu2-x/src/DplIIdecoder.cpp
@@ -101,10 +101,10 @@ void ProcessDplIISample32(const StereoOut32 &src, Stereo51Out32DplII *s)
     float VL = L / (1 - B); // if B>0, it means R>L, so increase L, else decrease L
     float VR = R / (1 + B); // vice-versa
 
-    // 1.73+1.22 = 2.94; 2.94 = 0.34 = 0.9996; Close enough.
+    // 1.73+1.22 = 2.94; 2.94 * 0.34 = 0.9996; Close enough.
     // The range for VL/VR is approximately 0..1,
     // But in the cases where VL/VR are > 0.5, Rearness is 0 so it should never overflow.
-    const float RearScale = 0.34f * 2;
+    const float RearScale = 0.34f; // *2; Let's remove the times 2 and enjoy the rear audo sans the horrible static noise.
 
     float SL = (VR * 1.73f - VL * 1.22f) * RearScale * Rearness;
     float SR = (VR * 1.22f - VL * 1.73f) * RearScale * Rearness;
--- a/plugins/spu2-x/src/Windows/ConfigAdvancedVolume.cpp
+++ b/plugins/spu2-x/src/Windows/ConfigAdvancedVolume.cpp
@@ -1,0 +1,232 @@
+/* SPU2-X, A plugin for Emulating the Sound Processing Unit of the Playstation 2
+ * Developed and maintained by the Pcsx2 Development Team.
+ *
+ * SPU2-X is free software: you can redistribute it and/or modify it under the terms
+ * of the GNU Lesser General Public License as published by the Free Software Found-
+ * ation, either version 3 of the License, or (at your option) any later version.
+ *
+ * SPU2-X is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+ * PURPOSE.  See the GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with SPU2-X.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "Global.h"
+#include "Dialogs.h"
+
+bool AdvancedVolumeControl = false;
+
+float VolumeAdjustFLdb; // Decibels settings, cos audiophiles love that.
+float VolumeAdjustCdb;
+float VolumeAdjustFRdb;
+float VolumeAdjustBLdb;
+float VolumeAdjustBRdb;
+float VolumeAdjustSLdb;
+float VolumeAdjustSRdb;
+float VolumeAdjustLFEdb;
+float VolumeAdjustFL; // Linear coefs calculated from decibels.
+float VolumeAdjustC;
+float VolumeAdjustFR;
+float VolumeAdjustBL;
+float VolumeAdjustBR;
+float VolumeAdjustSL;
+float VolumeAdjustSR;
+float VolumeAdjustLFE;
+
+void AdvancedVolumeConfig::ReadSettings()
+{
+    AdvancedVolumeControl = CfgReadBool(L"MIXING", L"AdvancedVolumeControl", false);
+
+    VolumeAdjustCdb = CfgReadFloat(L"MIXING", L"VolumeAdjustC(dB)", 0);
+    VolumeAdjustFLdb = CfgReadFloat(L"MIXING", L"VolumeAdjustFL(dB)", 0);
+    VolumeAdjustFRdb = CfgReadFloat(L"MIXING", L"VolumeAdjustFR(dB)", 0);
+    VolumeAdjustBLdb = CfgReadFloat(L"MIXING", L"VolumeAdjustBL(dB)", 0);
+    VolumeAdjustBRdb = CfgReadFloat(L"MIXING", L"VolumeAdjustBR(dB)", 0);
+    VolumeAdjustSLdb = CfgReadFloat(L"MIXING", L"VolumeAdjustSL(dB)", 0);
+    VolumeAdjustSRdb = CfgReadFloat(L"MIXING", L"VolumeAdjustSR(dB)", 0);
+    VolumeAdjustLFEdb = CfgReadFloat(L"MIXING", L"VolumeAdjustLFE(dB)", 0);
+
+    VolumeAdjustC = powf(10, VolumeAdjustCdb / 10);
+    VolumeAdjustFL = powf(10, VolumeAdjustFLdb / 10);
+    VolumeAdjustFR = powf(10, VolumeAdjustFRdb / 10);
+    VolumeAdjustBL = powf(10, VolumeAdjustBLdb / 10);
+    VolumeAdjustBR = powf(10, VolumeAdjustBRdb / 10);
+    VolumeAdjustSL = powf(10, VolumeAdjustSLdb / 10);
+    VolumeAdjustSR = powf(10, VolumeAdjustSRdb / 10);
+    VolumeAdjustLFE = powf(10, VolumeAdjustLFEdb / 10);
+}
+
+void AdvancedVolumeConfig::WriteSettings()
+{
+    CfgWriteBool(L"MIXING", L"AdvancedVolumeControl", AdvancedVolumeControl);
+
+    CfgWriteFloat(L"MIXING", L"VolumeAdjustC(dB)", VolumeAdjustCdb);
+    CfgWriteFloat(L"MIXING", L"VolumeAdjustFL(dB)", VolumeAdjustFLdb);
+    CfgWriteFloat(L"MIXING", L"VolumeAdjustFR(dB)", VolumeAdjustFRdb);
+    CfgWriteFloat(L"MIXING", L"VolumeAdjustBL(dB)", VolumeAdjustBLdb);
+    CfgWriteFloat(L"MIXING", L"VolumeAdjustBR(dB)", VolumeAdjustBRdb);
+    CfgWriteFloat(L"MIXING", L"VolumeAdjustSL(dB)", VolumeAdjustSLdb);
+    CfgWriteFloat(L"MIXING", L"VolumeAdjustSR(dB)", VolumeAdjustSRdb);
+    CfgWriteFloat(L"MIXING", L"VolumeAdjustLFE(dB)", VolumeAdjustLFEdb);
+}
+
+static BOOL CALLBACK DialogProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam)
+{
+    int wmId;
+    wchar_t temp[384] = {0};
+
+    switch (uMsg) {
+        case WM_PAINT:
+            return FALSE;
+
+        case WM_INITDIALOG: {
+            INIT_SLIDER(IDC_VOLADJ_CDB, 0, 50, 5, 5, 1);
+            SendDialogMsg(hWnd, IDC_VOLADJ_CDB, TBM_SETPOS, TRUE, VolumeAdjustCdb + 25.0f);
+            swprintf_s(temp, L"%d dB", (int)VolumeAdjustCdb);
+            SetWindowText(GetDlgItem(hWnd, IDC_ADV_VOL_C_LABEL), temp);
+
+            INIT_SLIDER(IDC_VOLADJ_FLDB, 0, 50, 5, 5, 1);
+            SendDialogMsg(hWnd, IDC_VOLADJ_FLDB, TBM_SETPOS, TRUE, VolumeAdjustFLdb + 25.0f);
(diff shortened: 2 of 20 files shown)


MrCK1:
Good job.

Dialog text is cut-off here:

image

FlatOutPS2:
> Some additional descriptive text should be added to the DPL level for users who are unfamiliar with it.

This goes for many SPU2-X UI elements. It's better to add tooltips for those at some point.

> Edit: If the sliders are set to 0%, they appear as -100% when the dialog is reinitialized.

I think something got deleted when cleaning up the code. :p

willkuer:
Do you know whether the change of the plugin api/abi regarding SPU2 delaycycles break backwards compatibility? Another great addition btw.

FlatOutPS2:
> Do you know whether the change of the plugin api/abi regarding SPU2 delaycycles break backwards compatibility?

Backwards compatibility hasn't changed. Only the fix will not automatically activate when using it on older builds.

refractionpcsx2:
Did you not treat those registers individually for Wallace and Gromit to see which one it was? It could be we just got one of them wrong

prafullpcsx2:
Did you had a look at end of dialogue looping issue with Gregory Horror Show? I wonder if it is a SPU-2X bug or Core issue.

FlatOutPS2:
> Did you not treat those registers individually for Wallace and Gromit to see which one it was? It could be we just got one of them wrong

I didn't, I don't have the game. :p The fact that at least one of these registers causes issues in W&G, also opens the possibility that the others could cause issues in other games. It's safer to reverse them all for now and do more testing.

refractionpcsx2:
Well if you'd asked around you would have found I do have it and could have tested it :P

ssakash:
Good work but personally I'm against the inclusion of the commit adding cycle delays on the GameDB, honestly seems too hacky to me. I'd prefer a broken behaviour more than hiding a real issue like this.

Though on the other hand, this is good for documentation purpose on knowing which games need no delays on the Key on/off events. I'd suggest adding those information near the voice start function as comments instead of adding more per game fixes. (Per game hacks are really annoying to look at when you're trying to understand a code properly)

FlatOutPS2:
> Good work but personally I'm against the inclusion of the commit adding cycle delays on the GameDB, honestly seems too hacky to me. I'd prefer a broken behaviour more than hiding a real issue like this.

You might have missed that there are all kinds of actual hacks in the GameIndex, some regular gamefixes also only apply to one or two games only, as do all GSdx CRC hacks...

ssakash:
> You might have missed that there are all kinds of actual hacks in the GameIndex, some regular gamefixes also only apply to one or two games only, as do all GSdx CRC hacks. They're all hacky and hide the real issue...

Yes, but this is the first time I'm seeing one for SPU2-X being added. I don't want another module to get filled with hacks like GSdx. If you feel it's a good idea then go for it, I just provided my opinion regarding the matter

In my opinion, the better way to handle this would be to do some investigation on the need for a kick start in the voice queuing subroutine for these games. (You can discuss with Rama/Sudonim in IRC if you're looking for information on the proper fix)

FlatOutPS2:
> In my opinion, the better way to handle this would be to do some investigation on the need for a kick start in the voice queuing subroutine for these games.

In my opinion it's better to do both. Adding this fix while investigating the original issue and hopefully find the true solution at some point(likely not very soon).

atomic83GitHub:
@FlatOutPS2 , it seems you forgot to add the hack to Unreal Tournament :). I will test this PR on Splinter Cell.

mirh:
> I don't want another module to get filled with hacks like GSdx.

I think what you may be wanting to say, is that gsdx.. Ok, fine, it's a lost cause to get in place for everybody to be happy. 
While sound on the other hand, spu2, has nothing attached saying it cannot be perfect - and keeping the issue "evident" might give some sense of hurry.

FlatOutPS2:
> it seems you forgot to add the hack to Unreal Tournament

Weird, I did add it originally. It must have gone missing when I copied it to the git commit. :/

It's in the file now.

willkuer:
As a dev you can always disable automatic gamefixes. But as user it is much harder to know the why how and where to fix such a problem. I highly would request to add as many gamefixes as possible as long as the compatibility or performance is improved by them.

MrCK1:
The GUI updates are nice, but I would rather have an investigation done on the remaining problematic games rather than add a hack. It's easy to add a hack and then "forget" to look for a proper fix later.

FlatOutPS2:
> The GUI updates are nice, but I would rather have an investigation done on the remaining problematic games rather than add a hack. It's easy to add a hack and then "forget" to look for a proper fix later.

I'll do it, but only if you agree on removing all other GameIndex/CRC hacks from PCSX2 as well. Let's stick it to the users so we have it easier! :roll_eyes: 
I'd say it's actually easier to look into the issue when you have an automatic gamefix. It even prints the fix is loaded in the log window, so instead of noticing a random error, you have something clear to look into if you want to.

mirh:
>  I highly would request to add as many gamefixes as possible as long as the compatibility or performance is improved by them.

Then why nobody is giving some love to #1894 ? :\

> It even prints the fix is loaded in the log window, so instead of noticing a random error, you have something clear to look into if you want to.

Lol. I guess like this beats every other "forgetting" concern. 
Just make sure to at least open a tracking issue though first!

lightningterror:
Perhaps Winamp DSP can be purged as well if it's not needed ?
I forgot what that does as well, I think @MrCK1 told me once if I'm not mistaken but I forgot again xD

FlatOutPS2:
> Perhaps Winamp DSP can be purged as well if it's not needed ?
I forgot what that does as well,

You can add modify/enhance the audio with those. I've never used it, but it seems like a good addition that should not be removed.

MrCK1:
> Is there a place where we can found Winamp DSP plugins? I found this website but I am not sure it's safe. https://winampheritage.com/plugins/DSP-Effect-5

I doubt the Winamp DSP plugin will work, it's not "standardized" AFAIK.

Anyways, somebody needs to test Unreal Tournament on one of these PRs so that #268 can be closed if necessary.

refractionpcsx2:
Is this ready to be merged?

mirh:
Maybe close #268, and reopen #2233 to track "proper fix" of this bug.

MrCK1:
*I have a complaint. 100dB scale is too much - that's going to blow out everybody's headphones 📦 

A 10dB range on both sides is plenty.

plugins/spu2-x/src/Windows/ConfigAdvancedVolume.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+        default:
+            return FALSE;
+    }
+    return TRUE;
+}
+
+void AdvancedVolumeConfig::OpenDialog()
+{
+    INT_PTR ret = DialogBoxParam(hInstance, MAKEINTRESOURCE(IDD_CONFIG_ADVANCEDVOLUME), GetActiveWindow(), (DLGPROC)DialogProc, 1);
```
turtleli: The 5th parameter seems unused. I think you could omit it and just use `DialogBox` instead.

plugins/spu2-x/src/Windows/ConfigAdvancedVolume.cpp
```diff
@@ -0,0 +1,209 @@
...
+* of the GNU Lesser General Public License as published by the Free Software Found-
+* ation, either version 3 of the License, or (at your option) any later version.
+*
+* SPU2-X is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+* without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+* PURPOSE.  See the GNU Lesser General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with SPU2-X.  If not, see <http://www.gnu.org/licenses/>.
+*/
```
turtleli: `*` isn't aligned.

ssakash:
@sudonim1 @ramapcsx2 @refractionpcsx2 

Is the first commit (Linking SPU2-X hack to PCSX2 module via GameIndex) fine with you guys?

bositman:
I think we shouldn't add yet another level of hacks. We can properly document them, also document the 'hackfix' and leave it at that. If people want these fixes so much we could even make some special build with the fixes and a forum thread for people to get it from.

turtleli:
Is audio clipping/distortion likely to occur if any of the advanced volume control sliders are set to a positive db level?

MrCK1:
Last time I tried it, there definitely was at a certain point. I'll check the resulting waveform and see where/if it still starts clipping.

plugins/spu2-x/src/Windows/ConfigAdvancedVolume.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    switch (uMsg) {
+        case WM_PAINT:
+            return FALSE;
+
+        case WM_INITDIALOG: {
+            INIT_SLIDER(IDC_VOLADJ_CDB, 0, 50, 5, 5, 1);
+            SendDialogMsg(hWnd, IDC_VOLADJ_CDB, TBM_SETPOS, TRUE, VolumeAdjustCdb + 25.0f);
+            swprintf_s(temp, L"%d dB", (int)VolumeAdjustCdb);
+            SetWindowText(GetDlgItem(hWnd, IDC_ADV_VOL_C_LABEL), temp);
```
turtleli: Maybe use a function/lambda expression here? i.e.
```
auto init_db_slider = [&](int slider_id, int label_id, float value) {
    INIT_SLIDER(slider_id, 0, 50, 5, 5, 1);
    SendDialogMsg(hWnd, slider_id, TBM_SETPOS, TRUE, value + 25.0f);
    swprintf_s(temp, L"%d dB", (int)value);
    SetWindowText(GetDlgItem(hWnd, label_id), temp);
};

init_db_slider(IDC_VOLADJ_CDB, IDC_ADV_VOL_C_LABEL, VolumeAdjustCdb);
```

plugins/spu2-x/src/DplIIdecoder.cpp
```diff
@@ -101,10 +101,10 @@ void ProcessDplIISample32(const StereoOut32 &src, Stereo51Out32DplII *s)
     float VL = L / (1 - B); // if B>0, it means R>L, so increase L, else decrease L
     float VR = R / (1 + B); // vice-versa
 
-    // 1.73+1.22 = 2.94; 2.94 = 0.34 = 0.9996; Close enough.
+    // 1.73+1.22 = 2.94; 2.94 * 0.34 = 0.9996; Close enough.
     // The range for VL/VR is approximately 0..1,
     // But in the cases where VL/VR are > 0.5, Rearness is 0 so it should never overflow.
-    const float RearScale = 0.34f * 2;
+    const float RearScale = 0.34f; // *2; Let's remove the times 2 and enjoy the rear audo sans the horrible static noise.
```
turtleli: Is it supposed to be correct with the x2 or not? If the x2 was a coding mistake, then the comment isn't really useful.

Also typo in commit title.

turtleli:
> Edit: I adjusted the Front Left/Right (since I only use headphones, and that's what they're recognized as)
> 
> Playing some music through the CD player, it clips instantly at +1dB. Unless there's more adjustments we can make, +25dB is going to be too much for the average user. +5dB would be OK I guess.

IMO the max should be 0db.

refractionpcsx2:
I agree the max should be 0dB

ramapcsx2:
Agreed on the 0DB thing. Audio software should make it impossible to overdrive in the digital realm.
If people want louder, they will have to turn up their amplifier.

> Is the first commit (Linking SPU2-X hack to PCSX2 module via GameIndex) fine with you guys?   

I don't like this hack but if you think this can't be dealt with any nice way for now, then so be it.
The thing is a specification / timing issue that's very narrowly pointing to SPU2 "voice on" behaviour.
Maybe one of the affected games has debug symbols available. Seeing the failing code would probably jumpstart a correct fix.
http://www.retroreversing.com/ps2-unstripped/
But for now, okay, fine.

hellbringer616:
Two cents from a guy who set up PA equipment for a few years; Never ever overdrive, best case you get louder but distorted audio, worst case you blow your speakers.

Question in regards to the pull request; will this be true PL2 processing? Or just enable faux- surround sound by adjusting channels? I cannot find a modern sound card with PL2 processing; I should just cave and buy a true amp and surround system..

ramapcsx2:
At one point in the past, it enabled proper positioning of audio sources onto the back speakers. The sound was meant to come from rear left, for example, and this decoding isolated it on rear left pretty well.
So I can't say if it was "true", but it was certainly no fake "3D audio".

That said, while I hope it still works this way today, I don't have the setup to test it anymore.

ramapcsx2:
@bositman I fundamentally agree with you on the points you listed, but here we have another case where a fat plaster helps the problem for now while also placatively remembering people what needs fixing.
I think it's okay to allow these.

turtleli:
I cherry picked the audio mode expansion regression fix.

lightningterror:
I'll mention this now before I forget again.

> Good work but personally I'm against the inclusion of the commit adding cycle delays on the GameDB, honestly seems too hacky to me. I'd prefer a broken behaviour more than hiding a real issue like this.Yes, but this is the first time I'm seeing one for SPU2-X being added. I don't want another module to get filled with hacks like GSdx. 

I share the same opinion as Akash. Rest of the features are cool.
Anyway I'm removing this from 1.6 milestone as well as High Priority since the regression fix has been added to master and FlatOut still hasn't came back.

Perhaps if someone is willing they can continue working on the PR.

lightningterror:
I included the fix with the proper register enabled for Wallace And Gromit: Curse Of The Were-Rabbit.

plugins/spu2-x/src/spu2sys.cpp
```diff
@@ -1080,15 +1083,16 @@ static void __fastcall RegWrite_VoiceAddr(u16 value)
 
         // NextA has been confirmed to not be allowed to be written to, so code has been commented out.
         // FlatOut & Soul Reaver 2 trigger these cases, but don't produce issues enabled or disabled.
+        // Enabled cases again due a regression in Wallace And Gromit: Curse Of The Were-Rabbit.
```
ramapcsx2: Can you leave a note that mentions which cases are now uncommented (4 and 5 :p)?

(Also, pretty interesting that this works to fix the game..)

lightningterror: I pretty much did in the last two commits
https://github.com/PCSX2/pcsx2/commit/eb955592ee0c473d55aa4164f6873d1ca7923a3a#diff-e5255967576cda2bc7f8be5c3efb5fdbR1083

MrCK1:
I can't remember anymore but I recall that there was a regression related to this fix a long time ago? Maybe my memory is foggy.

plugins/spu2-x/src/spu2sys.cpp
lightningterror: Maybe I can adjust the comment a bit tho.

lightningterror: Should be better now.

ramapcsx2: Thanks :)

lightningterror:
- Two of the regressions have been added to master.
- Most of us are against adding automatic fix for SPU2-X delay cycles.
- Maybe @MrCK1 can finish up the https://github.com/PCSX2/pcsx2/pull/2315/commits/8e9312ac2d7ac3cbc30747288a712231d36bf684
one day when he's free.

Anyway, I'll be closing this since there's been no word from @FlatOutPS2 and I don't want to see ghost prs open.

mirh:
> Most of us are against adding automatic fix for SPU2-X delay cycles.

Yes, but do we have an issue tracking that at least?
Otherwise there's no point in saying "we can do better than hacks"

Issue #2314, opened 2018-02-17
SPU2-X: Project Zero II audio stuttering


FlatOutPS2:
***PCSX2 version:***
v1.5.0 2286

***PCSX2 options:***
Default

***Plugins used:***
Default

***Description of the issue:***
Audio stutters or plays too fast when characters speak during gameplay (does not seem to affect cutscenes/FMV). 

***How to reproduce the issue:***
Play the game. In certain locations the voices will stutter and/or play too fast. Also seems to affect Project Zero III/Fatal Frame III based on mention of bug [here](https://wiki.pcsx2.net/PCSX2_Documentation/Google_Code_svn_repository_comments_archive_5000_to_5499).

***Last known version to work:***
PCSX2 v1.1.0 r5403

This is a regression caused by [this](https://github.com/PCSX2/pcsx2/commit/7dc3b25ddf5ae789bc0c897866eaf6af0c35cdef) commit by @sudonim1 

EDIT: Link to SPU2-X plugin build with regression removed: [Custom SPU2-X plugin](https://forums.pcsx2.net/attachment.php?aid=65729)

pgert:
Thanks for you special build of Spu2-X.
It sure simplifies things for PZ2, and PZ3 runs much better with recent versions of pcsx2\GSdx.

May I suggest that you put up a download-link for the plug-in here for the time being?
If nothing else it's useful for comparisons.

FlatOutPS2:
> May I suggest that you put up a download-link for the plug-in here for the time being?
If nothing else it's useful for comparisons.

I've added the link. Hopefully it won't be necessary for long.

atomic83GitHub:
Is it still an issue now?

refractionpcsx2:
yes

pgert:
Yes it is, furthermore g551847411 is the last version of pcsx2 that supports the custom version of the SPU2-X plugin by FlatOutPS2 posted above.

Mrlinkwii:
is this still an issue on lastest 1.7  ?

refractionpcsx2:
yes

PR #2318, opened 2018-02-20
SPU2-X: Fix Project Zero 2 regression

FlatOutPS2:
Fixes regression that caused audio stuttering when playing voices at some locations in Project Zero 2.

Chaos Legion seems unaffected.

Closes: https://github.com/PCSX2/pcsx2/issues/2314

@PCSX2/bug-squad Please test this on other games. Check for issues with voices/music stuttering or repeating.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -1110,6 +1110,7 @@ Serial = SCED-50642
 Name   = Final Fantasy X [Demo] [Final Fantasy VI PS1 - Bonus Disc]
 Region = PAL-E
 IPUWaitHack = 1
+eeRoundMode = 1 // Fix reverse control and boss in some places
 ---------------------------------------------
 Serial = SCED-50748
 Name   = Official PlayStation 2 Magazine Demo 26
@@ -1119,6 +1120,7 @@ Serial = SCED-50907
 Name   = Final Fantasy X [Bonus Disc - Beyond Final Fantasy]
 Region = PAL-Unk
 IPUWaitHack = 1
+eeRoundMode = 1 // Fix reverse control and boss in some places
 ---------------------------------------------
 Serial = SCED-51531
 Name   = Official PlayStation 2 Magazine Demo 33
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Just out of curiosity, if you put back in the >= TSA, does that break Chaos Legion? it would make more sense if it triggers when it was equal too, unless it only triggers once it passes it.

FlatOutPS2:
> Just out of curiosity, if you put back in the >= TSA, does that break Chaos Legion? it would make more sense if it triggers when it was equal too, unless it only triggers once it passes it.

Yes, TSA breaks Chaos Legion. TDA breaks PZ2.

refractionpcsx2:
interesting, I wonder why that is.  I guess it has to be inside it then :P

lightningterror:
> interesting, I wonder why that is. I guess it has to be inside it then :P

Voodoo ?

refractionpcsx2:
> Voodoo ?

Could be  :P it's more likely something like the memory read has to have passed the IRQ Address for it to trigger, rather than it firing when it's directly on it.

plugins/spu2-x/src/Dma.cpp
```diff
@@ -291,14 +291,14 @@ void V_Core::PlainDMAWrite(u16 *pMem, u32 size)
         // Buffer doesn't wrap/overflow!
         // Just set the TDA and check for an IRQ...
 
-        TDA = (buff1end + 1) & 0xfffff;
+        TDA = buff1end;
```
turtleli: Is this change necessary too? It makes the TDA handling inconsistent (see code above).

refractionpcsx2:
Why did you close this?

FlatOutPS2:
The code was meant for testing not reviewing. No one seemed interested in testing this, and there is something else wrong in the code. So I'll close this while I investigate if that is at the root of this issue as well.

I actually didn't manually close it myself. I send an empty commit, to see if it would automatically close, and it did. :smile:

refractionpcsx2:
Fair enough :) Yeah im sure we would have tested it but of course we wanna make sure there's no problems with the code too so we're not just bug finding and actually testing the changes intended :)

Looking forward to PR Mk 2 :P

PR #2320, opened 2018-02-21
Console: Reset the titlebar when the VM is shutdown.

MrCK1:
This is sort of a nitpick of mine that's bugged me for a bit. After the VM is shutdown, the game that was last played remains in the console. The new behavior will revert the console text back to `PCSX2 Program Log` after being shutdown.

### DIFF
--- a/pcsx2/gui/MainMenuClicks.cpp
+++ b/pcsx2/gui/MainMenuClicks.cpp
@@ -577,6 +577,7 @@ void MainEmuFrame::Menu_SysShutdown_Click(wxCommandEvent &event)
 	//if( !SysHasValidState() && !CorePlugins.AreAnyInitialized() ) return;
 
 	UI_DisableSysShutdown();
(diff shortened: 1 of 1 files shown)


pcsx2/gui/MainMenuClicks.cpp
```diff
@@ -577,6 +577,7 @@ void MainEmuFrame::Menu_SysShutdown_Click(wxCommandEvent &event)
 	//if( !SysHasValidState() && !CorePlugins.AreAnyInitialized() ) return;
 
 	UI_DisableSysShutdown();
+	Console.SetTitle(pxsFmt(L"PCSX2 Program Log"));
```
turtleli: pxsFmt and L aren't necessary I think.

MrCK1: You were correct, it's all set now.

MrCK1:
@turtleli This is ready to be merged as well.

PR #2326, opened 2018-02-27
GameDB: Update Crash Tag Team Racing

### LINKED ISSUE #1785: Crash Tag Team Racing - microvu1: Program cache limit reached
***PCSX2 version:***
1.4.0 and 1.5.0 dev 1803 g253057a

***PCSX2 options:***
default setup

***Plugins used:***
default plugins

Memory leak during gameplay causing the emulator to slowdown, halt unexpectedly for a few seconds, resume normally for a few minutes, then rinse, lather, and repeat. Gameplay is fine during races; problem only occurs while in third person adventure mode.

Walk around in third person through the various zones, jumping on platforms, etc.

N/A

***PC specifications:***
Intel 2600K CPU, nVidia GTX 770, Windows 8.1 x64

FlatOutPS2:
Fixes performance drops and short freezes in adventure mode.

Alas, there is no Crash Tag Team Wrestling, so I replaced it with the
correct title.

Updates Gamefixes Panel entry of VU I bit Hack.

Closes #1785

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -14042,6 +14042,7 @@ Compat = 5
 Serial = SLES-53439
 Name   = Crash Tag Team Racing
 Region = PAL-M6
+ScarfaceIbitHack = 1
 ---------------------------------------------
 Serial = SLES-53441
 Name   = Heroes of the Pacific
@@ -25262,8 +25263,9 @@ Name   = 3-Nen B-Gumi Kinpachi Sensei - Densetsu no Kyoudan ni Tate! [Best]
 Region = NTSC-J
 ---------------------------------------------
 Serial = SLPM-66090
-Name   = Crash Tag Team Wrestling
+Name   = Crash Bandicoot: Gacchanko World
 Region = NTSC-J
+ScarfaceIbitHack = 1
 ---------------------------------------------
 Serial = SLPM-66091
 Name   = Shinobido Imashime
@@ -39488,6 +39490,7 @@ Serial = SLUS-21191
 Name   = Crash Tag Team Racing
 Region = NTSC-U
 Compat = 5
+ScarfaceIbitHack = 1
 ---------------------------------------------
 Serial = SLUS-21192
 Name   = Cabela's Outdoor Adventures 2006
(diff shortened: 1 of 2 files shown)


lightningterror:
Maybe renaming the hack from ScarfaceIbitHack to something else might also be a good idea ?
Or maybe in the future if more games are found that can benefit from the hack.

FlatOutPS2:
> Maybe renaming the hack from ScarfaceIbitHack to something else might also be a good idea ?

Could be done, but it's not unusual to name a fix after the first game that was fixed by it. We even have the Wild Arms Offset hack in GSdx that also helps many other games.

FlatOutPS2:
> I'm surprised it fixes something else, i thought the hack for scarface was quite address specific, but if it helps this too, then cool.

I thought it was worth trying based on the data of the Scarface hack, and it works perfectly. No more ram build up, slowdowns or freezes. It doesn't help Killzone though, which also prints the same message and has the same (but less severe) issue.

> Can somebody check some translations to see if they actually translate this? if not, I don't think we need worry about that.

It is translated: https://github.com/PCSX2/pcsx2/blob/master/locales/de_DE/pcsx2_Main.po#L2169

refractionpcsx2:
What do you suggest? We just put it through and worry about that later or get the translations fixed (or at least modify the string it translates from) first?

lightningterror:
> What do you suggest? We just put it through and worry about that later or get the translations fixed (or at least modify the string it translates from) first?

Leave it to the translators ? :stuck_out_tongue:

FlatOutPS2:
Usually it's just left for later.

refractionpcsx2:
Fair enough :) I'll merge this later when I'm in my pc, not doing it in my phone after last time :p

refractionpcsx2:
On my pc* that's another pitfall of being in my phone, stupid lack of edit button.

lightningterror:
> On my pc* that's another pitfall of being in my phone, stupid lack of edit button.

You can check the "Request Desktop Site" on chrome mobile and Voilà.

FlatOutPS2:
> You can check the "Request Desktop Site" on chrome mobile and Voilà.

Yeah, most mobile browsers have this feature.

PR #2329, opened 2018-02-28
GSdx: Rename HW hacks drop list options

lightningterror:
Add "Default" nametag for default list options for the hacks: Sprite,
Round Sprite, GL Advanced settings, HPO.

Should help users know which options are the default ones with less
confusion.

OCD goodness.
image

### DIFF
--- a/plugins/GSdx/GSdx.cpp
+++ b/plugins/GSdx/GSdx.cpp
@@ -259,19 +259,19 @@ void GSdxApp::Init()
 	m_gs_bifilter.push_back(GSSetting(static_cast<uint32>(BiFiltering::Forced), "Bilinear", "Forced"));
 	m_gs_bifilter.push_back(GSSetting(static_cast<uint32>(BiFiltering::PS2), "Bilinear", "PS2"));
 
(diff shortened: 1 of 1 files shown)


MrCK1:
Do the sprite and HPO options really need `Default` next to them? It seems fairly obvious IMO.

lightningterror:
> Do the sprite and HPO options really need Default next to them? It seems fairly obvious IMO.

You can never be too sure, even I got confused a long time ago.

FlatOutPS2:
> You can never be too sure, even I got confused a long time ago.

People might be confused by Skipdraw and TC Offset and the checkboxes as well. That's why a Reset to Default really needs to be added.

lightningterror:
> People might be confused by Skipdraw and TC Offset and the checkboxes as well. That's why a Reset to Default really needs to be added.

You're right about that as well, this should still help for single option toggling if people don't want to reset other stuff.
I'll leave that to someone else to implement :stuck_out_tongue:

ssakash:
> I'll leave that to someone else to implement 😛

It's not really hard to implement it, I don't mind waiting if you want to work on it. Even if we know the default value, it might be cubersome to manually change it to default value for all the options, whereas just a single click is needed for restore defaults.

PR #2330, opened 2018-03-01
GSdx: GSCrc.cpp Add/adjust CRC ids and code reformatting

lightningterror:
Add CRC id for FFXII US, SFEX3 EU,
GT4 CH and GT4 Online Beta US.
Adjust BullyCC region from US to EU.

Add Missing regions for Ratchet & Clank,
Ace Combat, Destroy All Humans and Soul Reaver
series to Automatic Mipmapping.

GSdx: GSCrc.cpp reformat comments and whitespaces.
Since I was working there I also did a bit of cleaning up to satisfy my and anyone else's ocd :stuck_out_tongue:

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -66,6 +66,7 @@ CRC::Game CRC::m_games[] =
 	{0xCA284668, FFXII, EU, 0},
 	{0xC52B466E, FFXII, EU, 0}, //ES
 	{0xE5E71BF9, FFXII, FR, 0},
+	{0x0779FBDB, FFXII, US, 0},
 	{0x280AD120, FFXII, JP, 0},
 	{0x08C1ED4D, HauntingGround, EU, 0},
 	{0x2CD5794C, HauntingGround, EU, 0},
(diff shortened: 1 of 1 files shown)


lightningterror:
Note for self/future:
`{0xB590CE04, GTConcept, EU, 0},` is GT3 pal. I guess it got moved there so it can utilize the crc hack with the skip = 880 and it shares the code with GTConcept without the need to duplicate the code.
Kinda how like GT4 had separate code skip value targeted for pal 880 and ntsc 770

lightningterror:
I'll do a follow up in the future for more but first I wanna test some games with crc hacks so I'll stop here :stuck_out_tongue:

plugins/GSdx/GSCrc.cpp
```diff
@@ -137,21 +139,23 @@ CRC::Game CRC::m_games[] =
 	// and as "Twisted Metal - Black" (PAL).
 	{0x788D8B4F, ICO, EU, 0},
 	{0x29C28734, ICO, CH, 0},
+	{0xC02C653E, GT4, CH, 0}, // China Speed!
```
ssakash: Sneaky little comment there xD

lightningterror:
Now if someone is willing to add all these crc ids to their respective wiki pages that would be nice.

plugins/GSdx/GSCrc.cpp
FlatOutPS2: You should make PR just to remove this line.

PR #2335, opened 2018-03-08
GSdx: Add/adjust more CRC ids, remove GT3/Concept CRC hacks, cleanup hacks in GSRendererHW

lightningterror:
GSdx: Remove GT3/Concept CRC Hacks.
The CRC hacks are no longer needed due to the partial port of channel
shuffle on D3D.
They cause higher vram spikes  now.

GSdx: Add/adjust CRC ids.
Add some missing regions: Tomb Raider Legend JP,
StarWars Force Unleashed EU, SuperMan Returns EU, Valkyrie Profile 2 FR.
Rename GT3/Concept titles, they were incorrect.
Adjust Harley Davidson region id from NoRegion to US.
Add some missing regions to automatic mipmapping:
FIFA 03 US, FIFA 04 EU, FIFA 05 EU.
Reformat a few comments.

GSdx: Cleanup hacks in GSRendererHW.
Merge identical code for Spyro Games in to one to avoid duplicate code.
Rename hacks name for Jak series from OO_Jak to OO_JakGames since there
are multiple games added.

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -151,13 +151,13 @@ CRC::Game CRC::m_games[] =
 	{0x7ABDBB5E, GT3, CH, 0}, // cutie comment
 	{0x3E9D448A, GT3, CH, 0}, // cutie comment
 	{0xAD66643C, GT3, CH, 0}, // cutie comment
-	{0x6810C3BC, GT3, CH, 0}, // Gran Turismo Concept 2002 Tokyo-Geneva
 	{0x85AE91B3, GT3, US, 0},
 	{0x8AA991B0, GT3, US, 0},
 	{0xC220951A, GT3, JP, 0},
-	{0x9DE5CF65, GT3, JP, 0}, // Gran Turismo 3: A-spec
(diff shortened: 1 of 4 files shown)


ssakash:
@turtleli 

Is it fine to separate minor format changes into a separate commit? (https://github.com/PCSX2/pcsx2/pull/2335/commits/1f3eef55b15d60e7092033097549f4a18be455e9)

Personally I think it's good when there are lots of formatting changes being done but I'm not sure if it's optimal to have a separate commit specifically for it when the change is rather small. I recall you having a strong stance on favouring this, so pinging you for your opinion regarding the matter.

FlatOutPS2:
>  Add missing regions for FIFA03,04,05 series to Automatic Mipmapping.

You might want to word this more specifically as `Add some missing regions...`, as there are still regions missing. The same likely goes for some of the other games as well.

turtleli:
@ssakash I don't recall having a strong stance for small non-functional changes? I guess the same reasoning for having larger non-functional changes in a separate commit still applies to smaller changes, though it's not as important. In this case though it's only changing the comments in very minor ways.

lightningterror:
I'll merge the two commits then.

Issue #232, opened 2014-08-14
Arch macros not correctly defined for all projects when using VS solutions


sudonim1:
The arch macros introduced for preliminary x64 support are not being defined for all projects because of the property sheet mess, blocking pull requests depending on these macros.  The property sheets need to be looked at, multiple "common" property sheets collapsed into a single one, stuff that's not really common to all projects moved out of "common" etc.

PR #2342, opened 2018-03-18
windows: Improve VS property sheet stuff

turtleli:
This is an attempt to improve and simplify the VS property sheet stuff.

The idea so far:
* base.props - For all projects.
* nonthirdparty.props - For all non third party projects.
* vtune.props - For VTune users. Could probably be integrated into nonthirdparty.props. (Untested, don't have it).
* Third-party related property sheets - For projects with third party dependencies. Importing the relevant property sheet will deal with all the necessary stuff (include paths, preprocessor macros, project references, etc.). It feels easier for me so far, but I'll have to work on it more before I figure out whether I want to go forward with this part or not - It does introduce quite a few property sheets, and I'm not sure if it'll end up a quirky mess.

Some other things it does:
* Deals with the arch macros (nonthirdparty.props, #232). This is the main thing I want to sort out.
* Moves most intermediate build files to the `build` directory - I think it's nicer to have it all in one place (when I use grep it's always returning unwanted results from the build files instead of only the source code).

Things to do (notes for myself):
Check build settings (especially release build)
Reduce warnings (fix some, disable others)
Adjust old plugins too
Delete old property sheets
Do commits properly
Reintroduce async exception handling (I'm just experimenting with it off)


mirh:
@MrCK1 might be interested to check VTune stuff

Issue #1318, opened 2016-04-28
Meta: GSdx top left corner issue


gregory38:
I'm trying to implement a solution to handle properly the top left corner post-processing effect. Let's centralize information here.

Game impacted (from the top of my head, likely more).
* Ratchet & Clank
* MGS 2&3
* Prince of Persia
* GT4 ?
* Urban Chaos

refractionpcsx2:
Destroy All Humans has it, but it's not overly noticable (due to the shadow mess on the screen), having a scan through my games to see if i can find any others that do it.

gregory38:
I have a dump on this game, but I don't see it. Maybe not the good scene. However there are various postprocessing effect.

refractionpcsx2:
ill try and get another one if you'd like, it even does it on the menu, but im not totally sure it's right.

I'm updating that post as i find games btw

gregory38:
Ok. So far I have enough gs dump to implement somethings

FlatOutPS2:
Stuntman also crashes with a RAM spike with OpenGL(this happens more often than on 1.4.0).

gregory38:
Ok. I'm nearly sure  it is a variation of a similar effect. Currently my detection is based on FrameBuffer address == Texture address.
However Prince Of Persia. Copy the Framebuffer to another place. So my detection is dull. (and move is badly emulated in texture cache).
However detection work nicely for MGS and killzone.

refractionpcsx2:
That's progress :)

refractionpcsx2:
OK finished going through my DVD ISO's, there wasn't really many that weren't already mentioned that I could find. There might be more but I'm totally not going through all my discs ;)

gregory38:
Ok. Let's me first try to correct MGS/KillZone. For PoP, I need to find a way to detect the mess. Or maybe we could rely on crc/option.

refractionpcsx2:
> For PoP, I need to find a way to detect the mess. Or maybe we could rely on crc/option.

That all depends how often that effect is used and what drain on performance it is. If this is the only game to use it, or there is a huge performance drop, then yeh, crc it.

gregory38:
Without CRC ;)

![sub_rt](https://cloud.githubusercontent.com/assets/1569559/14900773/86cff5ae-0d92-11e6-9f19-2b94cd38fa5e.png)

[...]

PR #2347, opened 2018-03-18
GSdx-D3D: Texture and channel shuffle improvements

lightningterror:
Texture Shuffle changes:
Always Enable Texture shuffle on D3D10/11.
Previously Texture shuffle was enabled if CRC hack
level was below Full, this was kinda not good since
D3D also relies on CRC hacks on Full so you could either
stick with texture shuffle or crc hacks.

Texture shuffle is not supported on D3D9, however we can do a partial
port where instead of vertical lines with the effect we get the effect
on the entire screen. Better than nothing I suppose.

Ported some of the code from OpenGL to D3D
( just a copy - paste job :) ),
part of the code misses a dedicated shader but we can still
use it to fix various issues on many games.

List of affected games tested so far:
The Godfather, Final Fight Streetwise, The Suffering Ties that Bind,
Urban Chaos have their vertical lines issues fixed
(highly possible for other games as well), MGS and Stolen see an improvement
but they are still broken without crc hacks. Other games that suffered
similar issues are probably affected as well.

Channel Shuffle changes:
Update Channel Shuffle detection. A lot of games should see an improvement,
MGS, Urban Chaos, Stolen have their top left corner issues resolved.
Other games should be affected as well that use similar logic.
They still miss a shader so some effects are still broken/show glitches
but it's a nice improvement for D3D users.
Improves #1318

Shared changes:
Texture Shuffle and Channel shuffle have been moved to their
own dedicated functions. Should make things a bit cleaner.

Move part of the code for Texture Shuffle to GSRendererHW to be shared
across all HW renderers, should aboid copy paste/duplicate code.

GS Dumps for testing:
[Final Fight Streetwise.gs.zip](https://github.com/PCSX2/pcsx2/files/1823164/Final.Fight.Streetwise.gs.zip)
[The.Godfather.gs.zip](https://github.com/PCSX2/pcsx2/files/1823165/The.Godfather.gs.zip)

### DIFF
--- a/plugins/GSdx/GSRendererDX.cpp
+++ b/plugins/GSdx/GSRendererDX.cpp
@@ -147,6 +147,53 @@ void GSRendererDX::EmulateZbuffer()
 	}
 }
 
+void GSRendererDX::EmulateChannelShuffle(GSTexture** rt, const GSTextureCache::Source* tex)
+{
+	// Channel shuffle effect not supported on DX. Let's keep the logic because it help to
+	// reduce memory requirement (and why not a partial port)
+
+	// Uncomment to disable (allow to trace the draw call)
+	// m_channel_shuffle = false;
+
+	// First let's check we really have a channel shuffle effect
+	if (m_channel_shuffle) {
+		if (m_game.title == CRC::Tekken5) {
+			if (m_context->FRAME.FBW == 1) {
+				// Used in stages: Secret Garden, Acid Rain, Moonlit Wilderness
+				// Skip channel effect, it misses a shader for proper screen effect but at least the top left screen issue isn't appearing anymore 
+				// 12 pages: 2 calls by channel, 3 channels, 1 blit
+				// Minus current draw call
+				m_skip = 12 * (3 + 3 + 1) - 1;
+			} else {
+				// Could skip model drawing if wrongly detected
+				m_channel_shuffle = false;
+			}
+		} else if ((tex->m_texture->GetType() == GSTexture::DepthStencil) && !(tex->m_32_bits_fmt)) {
+			// So far 2 games hit this code path. Urban Chaos and Tales of Abyss.
+			// Lacks shader like usual but maybe we can still use it to skip some bad draw calls.
+			throw GSDXRecoverableError();
+		} else if (m_index.tail <= 64 && m_context->CLAMP.WMT == 3) {
+			// Blood will tell. I think it is channel effect too but again
+			// implemented in a different way. I don't want to add more CRC stuff. So
+			// let's disable channel when the signature is different.
+			//
+			// Note: Tales Of Abyss and Tekken5 could hit this path too. Those games are
+			// handled above.
+			m_channel_shuffle = false;
+		} else if (m_context->CLAMP.WMS == 3 && ((m_context->CLAMP.MAXU & 0x8) == 8)) {
+			// Read either blue or Alpha. Let's go for Blue ;)
+			// MGS3/Kill Zone
+			throw GSDXRecoverableError();
+		} else if (m_context->CLAMP.WMS == 3 && ((m_context->CLAMP.MINU & 0x8) == 0)) {
+			// Read either Red or Green. Let's check the V coordinate. 0-1 is likely top so
+			// red. 2-3 is likely bottom so green (actually depends on texture base pointer offset)
+			throw GSDXRecoverableError();
+		} else {
+			m_channel_shuffle = false;
+		}
+	}
+}
+
 void GSRendererDX::EmulateTextureSampler(const GSTextureCache::Source* tex)
 {
 	const GSLocalMemory::psm_t &psm = GSLocalMemory::m_psm[m_context->TEX0.PSM];
@@ -302,32 +349,7 @@ void GSRendererDX::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sourc
 
 	dev = (GSDeviceDX*)m_dev;
 
-	// Channel shuffle effect not supported on DX. Let's keep the logic because it help to
-	// reduce memory requirement (and why not a partial port)
-
-	// Uncomment to disable (allow to trace the draw call)
-	// m_channel_shuffle = false;
-
-	if (m_channel_shuffle) {
-		if (m_game.title == CRC::Tekken5) {
-			if (m_context->FRAME.FBW == 1) {
-				// Used in stages: Secret Garden, Acid Rain, Moonlit Wilderness
-				// Skip channel effect, it misses a shader for proper screen effect but at least the top left screen issue isn't appearing anymore 
-				// 12 pages: 2 calls by channel, 3 channels, 1 blit
-				// Minus current draw call
-				m_skip = 12 * (3 + 3 + 1) - 1;
-			} else {
-				// Could skip model drawing if wrongly detected
-				m_channel_shuffle = false;
-			}
-		} else if (m_context->CLAMP.WMS == 3 && ((m_context->CLAMP.MAXU & 0x8) == 8)) {
-			;
-		} else if (m_context->CLAMP.WMS == 3 && ((m_context->CLAMP.MINU & 0x8) == 0)) {
-			;
-		} else {
-			m_channel_shuffle = false;
-		}
-	}
+	EmulateChannelShuffle(&rt, tex);
 
 	// Upscaling hack to avoid various line/grid issues
 	MergeSprite(tex);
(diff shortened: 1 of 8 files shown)


refractionpcsx2:
Nice job dude :)

gregory38:
Which shader do you miss ?

lightningterror:
https://github.com/PCSX2/pcsx2/blob/38010432ee6193503a5a415b875cc5aec7140d92/plugins/GSdx/res/glsl/tfx_fs.glsl#L696

https://github.com/PCSX2/pcsx2/blob/38010432ee6193503a5a415b875cc5aec7140d92/plugins/GSdx/res/glsl/tfx_fs.glsl#L603

Well blending too since it's connected but nvm those.

gregory38:
OK, in this case you should add a note that they can't be supported on DX11. Unlike GL, you can't read the current framebuffer.

And did you check the behavior of shader for those parameters
```
PS_SHUFFLE and PS_READ_BA
```

By the way, did you check DX9 behavior ? DX9 GPUs don't support integer operation... (note I don't give a damn about dx9, it is only to be aware of the situation)

lightningterror:
> OK, in this case you should add a note that they can't be supported on DX11. Unlike GL, you can't read the current framebuffer.

Done

> And did you check the behavior of shader for those parameters
> `PS_SHUFFLE and PS_READ_BA`

Yeah SHADER_MODEL >= 0x400

> By the way, did you check DX9 behavior ? DX9 GPUs don't support integer operation... (note I don't give a damn about dx9, it is only to be aware of the situation)

It doesn't work since the shader is for dx10+

MrCK1:
Forgot to mention this earlier. 

It night be helpful to add comments to the code that doesn't work yet in DX11 to avoid confusion. (such as the Tomb Raider case)

lightningterror:
I'll also isolate the code or just make a skip value for the mask code on dx9 since it sometimes makes the render not show any image, don't want to break dx9 any further.

lightningterror:
I think it's fine as it is since it's a partial port. Plus it also works with blending on gl.

plugins/GSdx/GSRendererDX.cpp
```diff
@@ -23,10 +23,14 @@
 #include "GSRendererDX.h"
 #include "GSDeviceDX.h"
 
+bool s_IS_DX1011 = false;
```
gregory38: You should make the variable local to the class and use the `m_` prefix instead.

plugins/GSdx/GSRendererDX.cpp
```diff
@@ -147,6 +151,189 @@ void GSRendererDX::EmulateZbuffer()
 	}
 }
 
+void GSRendererDX::EmulateTextureShuffleAndFbmask()
+{
+	size_t count = m_vertex.next;
+	GSVertex* v = &m_vertex.buff[0];
+
+	// Gregory: code is not yet ready but it fixes vertical lines issues(partially on some) (MGS, The Godfather, Final Fight Streetwise) so let's enable it.
```
gregory38: maybe remove my old comment.

plugins/GSdx/GSRendererDX.cpp
```diff
@@ -147,6 +151,189 @@ void GSRendererDX::EmulateZbuffer()
...
+	} else {
+		//m_ps_sel.fmt = GSLocalMemory::m_psm[m_context->FRAME.PSM].fmt;
+
+		om_bsel.wrgba = ~GSVector4i::load((int)m_context->FRAME.FBMSK).eq8(GSVector4i::xffffffff()).mask();
+	}
+}
+
+void GSRendererDX::EmulateChannelShuffle(GSTexture** rt, const GSTextureCache::Source* tex)
+{
+	dev = (GSDeviceDX*)m_dev;
```
gregory38: This variable doesn't seem to be used in this function or did I miss it.

lightningterror: No, it's only used for setting shader resource and cb. Thought it might come in handy in the future when(if) the shaders get ported. I can remove it.

gregory38: Note that I did those cast in OpenGL because I only have a single device. This way the compiler can replace the dynamic selection of the device by a static call.

Here, it might not help the compiler because it would need to chose Dx9 device or Dx11 device.

plugins/GSdx/GSRendererDX.cpp
```diff
@@ -147,6 +151,189 @@ void GSRendererDX::EmulateZbuffer()
...
+				// 12 pages: 2 calls by channel, 3 channels, 1 blit
+				// Minus current draw call
+				m_skip = 12 * (3 + 3 + 1) - 1;
+			} else {
+				// Could skip model drawing if wrongly detected
+				m_channel_shuffle = false;
+			}
+		} else if ((tex->m_texture->GetType() == GSTexture::DepthStencil) && !(tex->m_32_bits_fmt)) {
+			// So far 2 games hit this code path. Urban Chaos and Tales of Abyss.
+			// Lacks shader like usual but maybe we can still use it to skip some bad draw calls.
```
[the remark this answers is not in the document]
lightningterror: GSDXRecoverableError seems like an even better solution, I'm noticing a difference between this pr and my local changes. Top left corner issue seems to be resolved.

plugins/GSdx/GSRendererDX.cpp
```diff
@@ -147,6 +147,193 @@ void GSRendererDX::EmulateZbuffer()
 	}
 }
 
+void GSRendererDX::EmulateTextureShuffleAndFbmask()
+{
+	size_t count = m_vertex.next;
+	GSVertex* v = &m_vertex.buff[0];
+
+	bool m_IS_DX1011 = false;
```
gregory38: No, you need to set (cache) the value during the init of the object.

Otherwise a much much nicer implementation would be to create a virtual function with 2 implementations. 1 for Dx9 and 1 for Dx10. You can look at SetupIA if you want an example.

lightningterror:
> Otherwise a much much nicer implementation would be to create a virtual function with 2 implementations. 1 for Dx9 and 1 for Dx10. You can look at SetupIA if you want an example.

Something like this? Let me know so I can squash.

plugins/GSdx/GSRendererDX9.cpp
```diff
@@ -56,6 +56,31 @@ bool GSRendererDX9::CreateDevice(GSDevice* dev)
...
+		// The game will take RG => BA and then the BA => RG of next pixels. 
+		// However, only RG => BA needs to be emulated because RG isn't used.
+		m_texture_shuffle = ((v[1].U - v[0].U) < 256) || m_context->FRAME.Block() == m_context->TEX0.TBP0;
+	}
+
+	if (m_texture_shuffle) {
+		// We can do a partial port for D3D9 that skips the draw call to give it a slight improvement.
+		// It's still broken but more bearable. Broken effect is on the screen but fully instead of vertical lines.
+		throw GSDXRecoverableError();
+	}
```
gregory38: Somehow, it feels like it miss an else part.

gregory38:
Yes it is better this way. Just need to ensure fbmask is properly (well not too wrongly) emulated on both dx9/dx11.

plugins/GSdx/GSRendererDX9.cpp
lightningterror: `else {
		//m_ps_sel.fmt = GSLocalMemory::m_psm[m_context->FRAME.PSM].fmt;

		om_bsel.wrgba = ~GSVector4i::load((int)m_context->FRAME.FBMSK).eq8(GSVector4i::xffffffff()).mask();
	}` Tho that;s part of dx11 only

gregory38: why it is dx11 only ?

lightningterror: Oh nvm, I thought it was part of the shader that's for texture shuffle.

plugins/GSdx/GSRendererDX9.cpp
```diff
@@ -56,6 +56,36 @@ bool GSRendererDX9::CreateDevice(GSDevice* dev)
 	return true;
 }
 
+void GSRendererDX9::EmulateTextureShuffleAndFbmask()
+{
+	size_t count = m_vertex.next;
+	GSVertex* v = &m_vertex.buff[0];
+
+	// Shadow_of_memories_Shadow_Flickering (Okami mustn't call this code)
```
gregory38: Hum this code should be moved where we compute the value of m_texture_shuffle (likely in GSRendererHW). I don't like the copy-past for each renderers

lightningterror: Which line ?
https://github.com/PCSX2/pcsx2/blob/master/plugins/GSdx/GSRendererHW.cpp#L882

gregory38: Yes below this line in RendererHW. Add the below block, because this block correct the detection of m_texture_shuffle. It should be done as soon as we detect the effect.

plugins/GSdx/GSRendererHW.cpp
```diff
@@ -881,6 +881,23 @@ void GSRendererHW::Draw()
 		m_texture_shuffle = (GSLocalMemory::m_psm[context->FRAME.PSM].bpp == 16) && (tex_psm.bpp == 16)
 			&& draw_sprite_tex && tex->m_32_bits_fmt;
 
+		size_t count = m_vertex.next;
+		GSVertex* v = &m_vertex.buff[0];
+
+		// Shadow_of_memories_Shadow_Flickering (Okami mustn't call this code)
+		if (m_texture_shuffle && count < 3 && PRIM->FST && (m_context->FRAME.FBMSK == 0)) {
```
gregory38: 2 small nitpicks
* Replace `count` by `m_vertex.next` (remove count declaration).
* Move `GSVertex* v = &m_vertex.buff[0];` below i.e just after the `GL_INS("WARNING: Possible misdetection of a texture shuffle effect");`

lightningterror:
Don't merge it yet, I'm noticing a small issue.

lightningterror:
Never mind, the issue is also present on master branch, maybe something to do with sw blending ? I'm not sure, anyway you can merge it if it looks good.

PR #2348, opened 2018-03-20
pcsx2: minor fixes

turtleli:
Just 2 small changes:
* Use enums instead of hardcoded integer.
* Avoid potential return of uninitialised value.


gregory38:
Can this be merged ? Maybe just check generated ASM is the same (i.e. 1 conditional)

turtleli:
Yeah, just being lazy. I'll merge the other two after I re-review them.

PR #2351, opened 2018-03-26
pcsx2 > Speedhacks & Game Fixes - GUI Textstring Alterations

pgert:
Changes translatable strings.

### DIFF
--- a/pcsx2/gui/Panels/SpeedhacksPanel.cpp
+++ b/pcsx2/gui/Panels/SpeedhacksPanel.cpp
@@ -115,10 +115,10 @@ Panels::SpeedHacksPanel::SpeedHacksPanel( wxWindow* parent )
 	const wxSizerFlags sliderFlags( wxSizerFlags().Border( wxLEFT | wxRIGHT, 8 ).Expand() );
 
 	m_check_Enable = new pxCheckBox( this, _("Enable speedhacks"),
-		pxE( L"Speedhacks usually improve emulation speed, but can cause glitches, broken audio, and false FPS readings.  When having emulation problems, disable this panel first."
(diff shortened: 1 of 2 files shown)


pgert:
Any support for this?

lightningterror:
Are these changes really necessary ?
Imo the original look better.

pcsx2/gui/Panels/SpeedhacksPanel.cpp
```diff
@@ -115,10 +115,10 @@ Panels::SpeedHacksPanel::SpeedHacksPanel( wxWindow* parent )
 	const wxSizerFlags sliderFlags( wxSizerFlags().Border( wxLEFT | wxRIGHT, 8 ).Expand() );
 
 	m_check_Enable = new pxCheckBox( this, _("Enable speedhacks"),
-		pxE( L"Speedhacks usually improve emulation speed, but can cause glitches, broken audio, and false FPS readings.  When having emulation problems, disable this panel first."
+		pxE( L"Speedhacks overall increases emulation speed, but can cause glitches, broken audio, and/or false FPS readings.\nWhen having emulation problems, disable (or alter) this panel first, in troubleshooting."
```
lightningterror: *increase

bositman:
Original has easier english, I don't think you will improve user experience with more complex wording

ssakash:
I don't see how this is better than the current description.

pcsx2/gui/Panels/SpeedhacksPanel.cpp
refractionpcsx2: "overall increase emulation speed" isn't always true, hence the original wording.
Also I'd rather people disabled speedhacks when having problems, if you just alter them, you may notice no difference and then assume speedhacks aren't the problem.

MrCK1:
These changes would also require a revision of other translations, I don't see the point.

pgert:
Well thanx for the feedback.
I'll close the PR in a couple of days then.
I seemed better to me ...

PR #2353, opened 2018-03-27
pcsx2: Remove theming support

turtleli:
It's incomplete, and there's not much that is actually themeable (basically only the background logo and icons can be changed).


bositman:
Well the background image being changable was an often requested feature, although IDK if it was even possible before this PR

turtleli:
It's possible on current master. If DeskTheme in PCSX2_ui.ini is changed to something other than Default, then PCSX2 will look in themes/[DeskTheme value] for replacement images.

I don't think supporting themes is useful. There's not much that's currently themeable and until I fixed some of it quite a while back it was in a more broken state.

lightningterror:
Well if the ui was a bit different, something like ppsspp then I'd see it being more useful but right now is a bit meh.

Well an alternative would be to add a gui option to change the background.
Misc > Change Background under Change Language sounds good.

turtleli:
I don't see how a change background feature would really improve things? Also, right now the GUI dialog sizes are dependent on the background image size, so letting a user change the logo can cause undesirable side effects.

gregory38:
It can be removed. If people don't like the GUI they will use an external frontend anyway. So let's no clutter the code with it.

PR #2357, opened 2018-03-29
GSdx: Remove/adjust crc hacks, add some crc ids

lightningterror:
GSdx: Adjust CRC hacks.
Purge Sengoku Basara crc hacks. Texture shuffle is emulated correctly on
d3d/ogl. d3d9 skips the bad draw call.

Move Eternal Poison crc hack to d3d level. Not needed on ogl since it's
emulated properly.

Move hack that removes texture shuffle for Demon Stoneback back to ogl level,
half screen bottom issue remains.

Extent CRC hack for The Getaway and The Getaway Black Monday to work on
EU regions.

Gsdx: Add some missing crc ids.
Tenchu Fatal Shadows US,
God Hand EU,
The Getaway EU,
The Getaway Black Monday EU.

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -283,6 +283,7 @@ CRC::Game CRC::m_games[] =
 	{0x1CC39DBD, SuikodenTactics, US, 0},
 	{0x3E205556, SuikodenTactics, EU, 0},
 	{0xB808413B, SuikodenTactics, JP, 0},
+	{0xA33AF77A, TenchuFS, US, 0},
 	{0x64C58FB4, TenchuFS, US, 0},
 	{0xE7CCCB1E, TenchuFS, EU, 0},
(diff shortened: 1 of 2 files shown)


willkuer:
I think having crc's in combination with titles and regions is quite nice. So maybe instead of removing them I'd rather comment or move them to the bottom and `#if 0` them to mark them as unused. But maybe they will never be needed again because of the recent accuracy gain of gsdx. ;-)

lightningterror:
@refractionpcsx2 wanna take care of this ?

lightningterror:
Yep

lightningterror:
I lied! Made some more changes, now it's ready.

refractionpcsx2:
Glad I asked and went to bed now :p

Are you sure this time? Lol

lightningterror:
Yes I'm sure, for now at least before I start testing other games.

Issue #2246, opened 2018-01-02
[Question] API or Com object, library, that can be called to retrieve currently running game name?


JoshWobbles:
***PCSX2 version:***
1.4.0

Is there any available way to call for the name of the currently loaded game?

Would like to incorporate the ability for may application (InputMapper) to load profile automatically based on the current game loaded.

Mem search for string yielded only strings that in the game itself so the address is un-trustworthy.

atomic83GitHub:
If I understand, the console don't show up the Game's name?

atomic83GitHub:
The games are listed by there serials, a PCSX2 dev team can better help you than me on this :).

nullripper:
Not a dev on the project, just a fan of PCSX2 and InputMapper, but I can give you some leads. PCSX2 exposes the name of the currently-running game in the title of the console window. For a hacky solution you could pop the console and read its title whenever a new emulation window appears. As for how it determines the title, I know it involves reading the game's serial number from the CD/DVD/ISO and looking it up in the database file "GameIndex.dbf". I haven't touched C code in nearly a decade so I can't tell how you'd actually _retrieve_ that information, but it's definitely in there somewhere if you poke around for references to serials and that specific database file. Good luck!

mirh:
You could hook the window title. 
I know for example spotifylyrics does something along that.

JoshWobbles:
@mirh Unfortunately the only window that has the game title is the console, and that window is optional.

@nullripper YES! GameIndex may be the missing link, not I have to see if I can find the game serial in the applications memory!

JoshWobbles:
So, just so it is documented for future use;

Memory address 0AF5D2EA-0AF5D2F5 contains the string of the game serial. This needs a tad of text conversion then GameIndex.dbf can be used to get the name.

Thanks guys!

mirh:
Was gonna say solved then, but I wouldn't know about linux best practices then.

i30817:
I'm doing this for ps1 in order not to care about dump format (read with libmirage, part of cdemu and connect the first data track to a iso parser):
https://gist.github.com/i30817/f58c7f904fe44c4599a6c9d48a60bce0

the same approach should be viable for ps2 if it follows ps1 practice. One thing i'm not sure of is if pycdlib can parse dvd data from libmirage the same as iso9660 data.

JoshWobbles:
I decided not to go the memory route, too susceptible to change. Instead I parse the log file using the following:

`
[...]

PR #2359, opened 2018-04-01
Discord RPC Implementation

RedPanda4552:
### The TL;DR
Adds some basic support for DRPC (Discord Rich Presence) to PCSX2. This allows game information to be shown on a Discord user's profile card in addition to the existing "Playing PCSX2" string.

### Some examples with dud images:
[In the menus](https://i.imgur.com/EJIBuDt.png)
[PS2 BIOS (this includes the PS2 splashes at the beginning)](https://i.imgur.com/HSodIbs.png)
[Gran Turismo 4](https://i.imgur.com/i9wHYSw.png)

### A note about social features
Currently game invite and spectate features are not enabled. Not only do these require approval by Discord themselves, we have no use for these, unless netplay makes it to master.

### Some gotchas to consider:
- As is, **this PR is not ready to merge. We must first agree who is going to host the Discord App and on what Discord account.** I personally think it would make more sense for a PCSX2 team member to do so, but I am also willing to modify my existing test App if that is the popular opinion instead. For those unfamiliar, Discord uses "Apps" to control who and what interfaces with their platform. Bots, DRPC, any kind of tool interacting needs an App to authorize against. For DRPC, this simply means someone makes the App (which is stupid easy, you just plug in a name and you're basically done), adds some artwork to it, and we are set. The important thing for DRPC is that the **Client Id** must be a string literal in the source. This tells DRPC what App to authenticate with. Additionally, the App should have artwork added so that the image slots you see in my examples are populated. The literals I used for the images are simply "pcsx2" and "pcsx2_small".
- I think, _think_, I got the lib file linked properly. But I'd like someone to try cloning this branch and attempting a compile on their own.
- ~~DRPC has a version for both Windows and Linux, but I only included the Windows lib and header file. Not sure how different they are, but I would like someone to verify Linux still builds and runs properly. If there is demand, I will attempt to work the Linux version in as well.~~ CK has already shown me the Travis build result, and it looks like Linux can't see the header file, I assume because VS is doing VS things to handle it and gcc isn't finding it right.
- PSX games show up as "PS2 BIOS", because in some way or another gameName and DiscSerial are both coming back as empty. Even though the console title bar is set to DiscSerial, which is clearly not empty considering it in fact throws the serial into the console title, moments before the empty check. If some sort of game db is implemented for PSX, this should, in theory, self resolve.
- Super long game titles will get cut off. The game title is currently being printed in the details line, but the status line underneath is unused. It will be tricky without monospace fonts and especially considering display scaling and resolution differences, but if someone else can think of a decent way to go about it, splitting the game title across the two lines may be an option.

### Personal remarks from RedPanda4552:
- WX is a complete pain. While I would have preferred to just pass wxStrings in to DRPC to avoid using sprintf over and over and waterfalling down to the DRPC call, this ends up completely crashing WX and after several hours of breakpointing and attempted debugging I have yet to find any logical reason.


MrCK1:
It would be best to remove the image for now, the default image allows for quite a bit more text to fit in the box currently.

pr

pcsx2/gui/Discord/DiscordRPC.cpp
```diff
@@ -0,0 +1,101 @@
+/*  PCSX2 - PS2 Emulator for PCs
+*  Copyright (C) 2002-2018  PCSX2 Dev Team
```
lightningterror: You can also include your name here.
```
Copyright (C) 2018 RedPanda4552
Copyright (C) 2002-2018  PCSX2 Dev Team
```

pcsx2/gui/Discord/DiscordRPC.cpp
```diff
@@ -0,0 +1,101 @@
...
+char *drpcGameTitle;
+u64 drpcStartTime;
+
+// Set the display name for the active game for the next update.
+// Pass dprcMenuStr or nullptr to reset to "In the menus".
+// * Does not immediately update Discord RPC, use drpcUpdate to do so
+// ** Side note, wx is a complete dumpster fire. Apparently passing a wxString in to a function crashes
+// the entire library, so we're just going to do wxString to char* conversions in core code and accept
+// that it adds a little more bulk.
+void drpcSetGame(char* newTitle) {
```
lightningterror: I prefer the { begins on a new line to keep the code consistent on void functions.

```
void drpcSetGame(char* newTitle)
{
}
```

gregory38:
Fwiw, there is a macro WX_STR to print wx string `L"%s", WX_STR(my_str)`

avih:
Personally I don't think PCSX2 should integrate with 3rd party presence platforms, and especially not with the closed source discord.

However, you can instead just write the state info to some `pcsx2.state.json` file (or INI file or whatever format you think is useful) whenever the state changes, and then use some external dircord app (not part of PCSX2) which does the discord integration according to the pcsx2 state file content.

lightningterror:
It's actually a great idea to integrate it, a lot of people use discord and it also improves the social aspect.
Plus it's a modern feature that goes well in today's standards.

avih:
We don't integrate twitter, facebook, irc, etc libraries, and IMO discord is no different. You can achieve exactly the same goal by following my suggestion and without cluttering pcsx2 code.

mirh:
I don't think that's the point. 
As I was saying (on discord.. duh), pcsx2 should expose X or Y information neatly and gracefully to the system, *if really any* (in fact #2246 was exactly about this)
But nothing specific to programs, let alone if being "the standard" of nothing.

ramapcsx2:
Considering what's going to happen if we integrate this, I agree with @avih's suggestion here.
Just externalize the whole thing once, and later there can be as many social integrations as people like :)

avih:
Why did you close it? Just change it to write a status file instead of using discord API... then you can integrate with anything which knows to read this status file... (which I suggest to be a UTF8 JSON).

RedPanda4552:
I figure if I'm going to do so, I should start over from scratch without any Discord API included on a new branch. So, leaving this one behind.

mirh:
I know shit nothing about interfaces, but writing to a file seems so 1990. 
Isn't there really some API/interface/bus/whatever for stuff like this?

avih:
A non developer cannot add this feature, as it needs support in PCSX2 itself. It seems the original author lost interest to add generic support for third party presence platforms.

AnonGuy:
If anyone's interested, I've made a quick proof-of-concept with @joshwobbles' idea of parsing log files:
https://github.com/AnonGuy/discordrp-pcsx2

(Excuse the necromancy, I've only just seen this PR when looking for an RPC solution myself)

mirh:
Well, that's a step in the right direction I guess. 
Maybe there's some python lib that could "automate everything" by checking pcsx2 process (or windows children names) title?

PR #2375, opened 2018-04-08
man: Minor fixes

orbea:
This adds two commits.

1. This moves `PCSX2.1` to `PCSX2.6` and install its to man section `6` instead of `1`. The reason for this is according the man manual page games should be in section `6` where section `1` is used for general tools. I think we can agree PCSX2 is more a game than a general tool?

2. This silences some warnings for the PCSX2 man page found with the lint feature from `mandoc(1)`.

Any suggestions are of course welcome.
```
Only select manuals from the specified section.  The currently available sections are:

      1         General commands (tools and utilities).
      2         System calls and error numbers.
      3         Library functions.
      3p        perl(1) programmer's reference guide.
      4         Device drivers.
      5         File formats.
      6         Games.
      7         Miscellaneous information.
      8         System maintenance and operation commands.
      9         Kernel internals.
```
```
man -Tlint PCSX2
```
```
no blank before trailing delimiter
  (mdoc) The last argument of a macro that supports trailing
  delimiter arguments is longer than one byte and ends with a
  trailing delimiter. Consider inserting a blank such that the
  delimiter becomes a separate argument, thus moving it out of
  the scope of the macro.
```
```
skipping no-space macro
  (mdoc) An input line begins with an Ns macro, or the next argument
  after an Ns macro is an isolated closing delimiter. The macro is
  ignored.
```
https://man.openbsd.org/mandoc.1
```
man: PCSX2.6:65:9: STYLE: no blank before trailing delimiter: Fl h,
man: PCSX2.6:119:9: WARNING: skipping no-space macro
man: PCSX2.6:180:72: WARNING: skipping no-space macro
```
1. This is fixed by adding a zero-width space (\&) so that the trailing delimiter character is no longer at the end.
2. Removes the ignored no-space macros. This has the side effect of removing a lone period.

See the following diff of the rendered man page.
```
--- PCSX2.1.orig        2018-04-08 16:02:09.300643308 -0700
+++ PCSX2.6     2018-04-08 16:03:47.677587114 -0700
@@ -1,4 +1,4 @@
-PCSX2(1)                    General Commands Manual                   PCSX2(1)
+PCSX2(6)                         Games Manual                         PCSX2(6)
 
 NAME
      PCSX2 – PlayStation(R)2 Console Emulator
@@ -121,7 +121,6 @@
      Bugs can be reported by submitting an issue at the PCSX2 Github issue
      tracker:
            https://github.com/PCSX2/pcsx2/issues
-     .
 
 PCSX2 RELATED WEBSITES
      •   PCSX2 git repository:
```

### DIFF
--- a/bin/docs/PCSX2.1
+++ b/bin/docs/PCSX2.1
@@ -62,7 +62,7 @@ This allows you to play PS2 games on your PC, with many additional features and
 benefits.
 .Sh GENERAL OPTIONS
 .Bl -tag -width Ds
-.It Fl h, Fl \-help
+.It Fl h,\& Fl \-help
 Displays the list of available command line options.
 .It Fl \-cfg Ns = Ns Ar file
 Uses
@@ -116,7 +116,7 @@ Boots with an empty DVD tray.
 Use this to boot into the PS2 system menu.
 .It Fl \-usecd
 Boots using the configured CDVD plugin instead of booting
-.Ar iso Ns .
+.Ar iso .
 .It Fl \-fullscreen
 Runs the game in fullscreen mode.
 .It Fl \-windowed
@@ -177,7 +177,7 @@ User configuration and data directory.
 .An PCSX2 Dev Team and many other contributors
 .Sh BUGS
 Bugs can be reported by submitting an issue at the
-.Lk https://github.com/PCSX2/pcsx2/issues "PCSX2 Github issue tracker" Ns .
+.Lk https://github.com/PCSX2/pcsx2/issues "PCSX2 Github issue tracker"
 .Sh PCSX2 RELATED WEBSITES
 .Bl -bullet
 .It


turtleli:
I'll look at it soonish. I had no access to a dev machine for the last week and a half.

refractionpcsx2:
I don't wanna touch anything that messes with cmake's, I don't have a clue how they work :P

orbea:
The cmake change is pretty braindead simple, it just changes the filename and install path accordingly for moving it to man section `6` (games) instead of `1` (general commands). The main question is if you guys agree with moving it to section 6 or not. 

The two commits are not dependent on each other and are only in one PR because they are small and both affecting the same man page.

refractionpcsx2:
Maybe so, but I refuse to touch any linux'y stuff, I don't want to be responsible if something breaks as I have no way to test it :P

orbea:
> Well technically PCSX2 is a virtual machine such as qemu. It doesn't work but it should be fixable to run a Linux kernel inside PCSX2.

So potentially someone could run Linux on windows through pcsx2? :)

MrCK1:
Possibly yes. I think he's referring to the "Linux for Playstation 2" disk - I tried it myself and it doesn't even boot.

willkuer:
So can I run pcsx2 on windows via wine on linux on pcsx2 on windows?

turtleli:
> This has the side effect of removing a lone period.

I think the period shouldn't be removed?

As for section I think leaving it in 1 is fine, though I understand why it might also be classed as 6. I don't have a strong preference either way.

gregory38:
@MrCK1 
```
Possibly yes. I think he's referring to the "Linux for Playstation 2" disk - I tried it myself and it doesn't even boot.
```
Yes I know, likely because MMU/COP0 isn't fully implemented. It should be doable to fix it with the interpreter. But it is the lowest priority on Earth.

Let's keep the manual of first section. Avoid to break any package. Besides it isn't important, nobody should care where is located the man page. `man PCSX2` should just work. If some distribution doesn't like it, they can still install the file in the section 6.

orbea:
I dropped the first commit. However I still think dropping that period is a good idea. Here is the difference in the rendered man page here.
```
--- old      2018-04-25 22:23:33.802755655 -0700
+++ new      2018-04-25 22:22:57.464400935 -0700
@@ -121,7 +121,6 @@
      Bugs can be reported by submitting an issue at the PCSX2 Github issue
      tracker:
            https://github.com/PCSX2/pcsx2/issues
-     .
 
 PCSX2 RELATED WEBSITES
      •   PCSX2 git repository:
```

turtleli:
On my end the period is on the same line as the URL and doesn't look out of place.

Maybe an alternative would be to move the link to the PCSX2 related websites section?

lightningterror:
Status on this ?

orbea:
Sorry, I've been distracted... I'll try to get back to this soon.

lightningterror:
It's been a few months, any update ?

lightningterror:
@orbea Status ?

lightningterror:
I will close this, there's been no progress and it's really low priority.

Ping us when you want to resume working on it so we can re open it.

PR #2368, opened 2018-04-05
GameDB: Revert Crash Tag Team Racing gamefix

lightningterror:
Revert ScarfaceIbitHack Gamefix.
Regression was introduced in #2326.
Caused hard freezes making the game unplayable and required an emulator
restart without gamefixes enabled.
Without the fix the game experiences short freezes but the game is still
playable.

Close #2366

PR #2378, opened 2018-04-11
PCSX2: Update VU I bit gamefix tooltip

lightningterror:
Remove mention of Crash Tag Team Racing from VU I bit gamefix tooltip.
Not needed since the gamefix has been removed for the game as it causes
more issues.

Related #2368

### DIFF
--- a/pcsx2/gui/Panels/GameFixesPanel.cpp
+++ b/pcsx2/gui/Panels/GameFixesPanel.cpp
@@ -105,9 +105,8 @@ Panels::GameFixesPanel::GameFixesPanel( wxWindow* parent )
 			wxEmptyString
 		},
 		{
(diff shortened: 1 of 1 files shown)


pgert:
"Known to affect following games:\n * Scarface The World Is Yours" should be
 "Known to affect following game: Scarface The World Is Yours.", yes?

lightningterror:
It's a list (can be one or multiple) so it's fine.

refractionpcsx2:
I would say as it only is one game that it should be "Known to affect Scarface The World is Yours"
We can make it list-like later if we find another game that it does affect.

refractionpcsx2:
Cool :) We can always change it later. Although tbh I'd be surprised if any other games use it, it's pretty specific.

PR #2384, opened 2018-04-14
GameDB: Add vuClampMode hack/remove eeClampMode hack for Pac-Man World Rally

MrCK1:
Fixes bad geometry. This also replaces the eeClampMode hack in the NTSC-U release which didn't appear to do anything (it was probably a mistake and I can't find any references to it)

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -16473,6 +16473,7 @@ Region = PAL-M5
 Serial = SLES-54584
 Name   = Pac-Man Rally
 Region = PAL-M5
+vuClampMode=2 // Fixes bad geometry.
 ---------------------------------------------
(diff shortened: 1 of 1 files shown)


MrCK1:
@refractionpcsx2 I only tested the NTSC-U release but based on the issue, I'm sure the PAL release needs it as well.

refractionpcsx2:
It should probably be tested just for completeness and to be 100% sure (although I'm sure you are right)

lightningterror:
So can this be merged ?

refractionpcsx2:
I guess, I was kinda hoping the PAL version could be tested just to be sure. But, yeah, guess that's not gonna happen, so might as well merge it :P

PR #2403, opened 2018-04-23
GSdx: Update D3D fbmask emulation on texture shuffle

lightningterror:
Improves accuracy on Sly series.

Rendering is still bad and needs a lot more work but
it's a little bit better.

### DIFF
--- a/plugins/GSdx/GSRendererDX11.cpp
+++ b/plugins/GSdx/GSRendererDX11.cpp
@@ -119,17 +119,20 @@ void GSRendererDX11::EmulateTextureShuffleAndFbmask()
 
 		// Please bang my head against the wall!
 		// 1/ Reduce the frame mask to a 16 bit format
-		// FIXME GSVector will be nice here
 		const uint32& m = m_context->FRAME.FBMSK;
 		uint32 fbmask = ((m >> 3) & 0x1F) | ((m >> 6) & 0x3E0) | ((m >> 9) & 0x7C00) | ((m >> 16) & 0x8000);
+		// FIXME GSVector will be nice here
+		uint8 rg_mask = fbmask & 0xFF;
+		uint8 ba_mask = (fbmask >> 8) & 0xFF;
 		om_bsel.wrgba = 0;
 
 		// 2 Select the new mask (Please someone put SSE here)
-		if ((fbmask & 0xFF) == 0) {
-			if (write_ba)
+		if (rg_mask != 0xFF) {
+			if (write_ba) {
 				om_bsel.wb = 1;
-			else
+			} else {
 				om_bsel.wr = 1;
+			}
 		} else if ((fbmask & 0xFF) != 0xFF) {
 #ifdef _DEBUG
 			fprintf(stderr, "Please fix me! wb %u wr %u\n", om_bsel.wb, om_bsel.wr);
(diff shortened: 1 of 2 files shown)


refractionpcsx2:
Is this ready to go or do you need to do more work?

gregory38:
It would be better to use the gl macro for debug. Maybe rename them. And in a separate commit

gregory38:
Or add a dx macro

lightningterror:
> Is this ready to go or do you need to do more work?

If Gregory is ok with it yea.

gregory38:
Please put a macro for those debug message. 

Potentially, it could be a good idea to keep ps_sel.fbmask logic to print a warning in debug build. You should add that rendering is still bad in the commit message.
The game writes only a couple of bits of the colors.
* Old code was write nothing
* New code is write all bits
Both solutions are wrong. Even if the new one is potentially "better".

gregory38:
There are a bunch of macro at the bottom of `stdafx.h`. Gl main print function is `GSDeviceOGL::DebugOutputToFile`

What we need to do
* Create a global logger object with inheritance so it could dispatch the log to the good API (GL/Dx)
* Instantiate an object pointer in GS.cpp
* Then add handy macro that will call the logger object. Macro will save some typing and allow to remove code in release build without adding `ifdef` everywhere.

For the interface of the logger object it is a bit complicated. I propose that we use the [gl_spec](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glDebugMessageInsert.xhtml) as a base. I don't know others APIs but it is unlikely that we need more. We need a `type`, a `severity` and finally a `message` (that might need to be formatted).

```
enum class LogSeverity {
    notification,
    warning,
    error,
    fatal,
};

enum class LogType {
    texture_cache, 
    gs_register,
    performance,
    not_supported,  
    special_effect,  // typically hle stuff
    info,
};


class GSLog {
   // Depends on the API
  virtual insert_message(enum type, enum severity, const std::string& msg);

public:
   // There are various example of printf formatting. Potentially you can use the format function
   void Insert(enum type, enum severity, const char* fmt, ... );
   void Flush(); // should call  GSDeviceOGL::DebugOutputToFile for gl

  // if enable is false, all message related to type should be discarded. Otherwise we keep them
  // maybe add a similar function for severity. I don't know.
  void Control(enum type, bool enable);
};

class GSLogOGL : public GSLog {
   ...
   do inheritance
   ...
};
```

With those kinds of macros. Potentially, we can hide severity inside the macro, for example gs_register info are always a `notification`
```
LOG_CACHE(LogSeverity::Error, "my_message with integer %i", tex_id);
=> call: g_log->Insert(LogType::texture_cache, ...)
```


I'm well aware that it might sound a complicated project. But it is an interesting project, and you will learn plenty of stuff. Don't hesitate to google a lot :)

lightningterror:
> Edit: potentially we could a separate PR for this one. And things aren't well defined in my brain, feel free to tune it. We just need something generic enough so we can control the message information easily and that will work for both API.

Alright will remove the debug changes and plan them for a future PR, who knows how long it will take :laughing:

lightningterror:
Okay so no regressions spotted, Sly series sees an improvement - no more darker stripes.

@refractionpcsx2 it's all yours, merge it.

refractionpcsx2:
Your wish is my command :)

PR #2405, opened 2018-04-25
GSdx-TC: JackieChanAdv run invalidate video memory

### LINKED ISSUE #1838: [GSDX-TC] Regression in Jackie Chan Adventures

lightningterror:
Run invalidate video memory on JackieChanAdv when depth is disabled to
avoid regressions, fixes black/grey background on D3D/OGL and textures
flickering ingame.

The game also doesn't play well with depth emulation on OGL as it causes
a half right screen issue so this is the best workaround we have atm.

Regression PR: #1776
Related issue: #1838

D3D test comparisons:
Master (first boot)
pcsx2 2018-04-25 03-37-50-74

Master (render paused/resumed/restarted)
pcsx2 2018-04-25 03-37-24-64

PR
pcsx2 2018-04-25 03-38-19-82

Close #1838

### DIFF
--- a/plugins/GSdx/GSTextureCache.cpp
+++ b/plugins/GSdx/GSTextureCache.cpp
@@ -669,7 +669,10 @@ GSTextureCache::Target* GSTextureCache::LookupTarget(const GIFRegTEX0& TEX0, int
 // must invalidate the Target/Depth respectively
 void GSTextureCache::InvalidateVideoMemType(int type, uint32 bp)
 {
-	if (!CanConvertDepth())
+	if (!CanConvertDepth() && m_renderer->m_game.title != CRC::JackieChanAdv)
+		// JackieChanAdv shouldn't hit this code path even if depth is not supported,
+		// otherwise it causes a regression on D3D/OGL on the main menu where the background image is back instead of yellow, and it also fixes texture flickering ingame.
+		// Note: The game also doesn't like Depth Emulation as it causes half right screen issue so this is the best workaround we have atm.
 		return;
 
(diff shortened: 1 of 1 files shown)


gregory38:
Tbh, I don't like it. It is pure luck that adding extra invalidation of target is kinds of working. Instead to add more hack for depth stuff, it would be **much much much** better to port the depth shader.

lightningterror:
Yes it would be ideal to port the depth shader however I forgot to mention that the game needs Depth disabled even on OpenGL to work properly, on the main menu if depth is disabled then half right background screen is black/grey instead of yellow.

gregory38:
I don't want to be the bad guy, but I prefer a broken game rather than an awful hack in the middle of the code.

Sure a proper fix is more work. But it is the only solution to fix all those deph lookup issues.

lightningterror:
Oke I'll close this then.

gregory38:
Sorry

PR #2406, opened 2018-04-25
Made changes to ps1 mode - also retargeted pcsx2 

kenshen112:
Vs is dumb and won't let me compile with out retargeting bleh 😣 

changes made to ps1 mode, set PSXCLK to 33mhz, adjusted ee to iop clock although I'm sure there's a missing if statement there.


mirh:
Mhh.. It's a bit I don't compile myself, but for starters the retargeting problem really sounds like a problem on your side. 

Second.. I don't seem to see \*any\* other change in the PR.

PR #2407, opened 2018-04-25
PSX: Mode adjustments, fixed PSXCLK and adjusted EE clock speed 

kenshen112:
The point / goal of this pr is to adjust pcsx2's clock speed in ps1 mode to avoid ps1 game's running faster then intended, most of this commit is focused on adjusting the PSXCLK EE and IOP multipliers and CD speed all centered around IOP speed and it's dependencies.

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -22,6 +22,7 @@
 #include <wx/datetime.h>
 
 #include "CDVD.h"
+#include "CdRom.h"
 #include "CDVD_internal.h"
 #include "CDVDisoReader.h"
 
@@ -1416,6 +1417,10 @@ static __fi void cdvdWrite0F(u8 rt) { // TYPE
 static __fi void cdvdWrite14(u8 rt) { // PS1 MODE?? // This should be done in the SBUS_F240 bit 19 write in HwWrite.cpp
 	u32 cycle = psxRegs.cycle;
 
+	PSXCLK = 33868800;      //33.8688 Mhz Playstation 1 speed.
+    psxEEmultiplier = 8.71; //proper EE to iop cycle freaquency.
+    setPsxSpeed();          //this sets the cd speed for ps1 mode.
+	
 	if (rt == 0xFE)
 		Console.Warning("*PCSX2*: go PS1 mode DISC SPEED = FAST");
 	else
--- a/pcsx2/Common.h
+++ b/pcsx2/Common.h
@@ -19,6 +19,14 @@
 
 static const u32 BIAS = 2;				// Bus is half of the actual ps2 speed
 static const u32 PS2CLK = 294912000;	//hz	/* 294.912 mhz */
+extern s64 PSXCLK; // The Iop cycle clock or the PS1's main processor this variable determins and set's the freaquency of the processor. 	
+
+
+// This variable controls and set's the Multiplier value for the EE to Iop cycle clock,                     
+// This is the thing that controls the freaquency of Iop to EE interaction and vice versa                       
+// This changes between PS2 and PS1 mode's to account for the clock difference in                            
+// The Iop or ps1 processor's speed between the two modes.
+extern int psxEEmultiplier; 
 
 #include "System.h"
 #include "Memory.h"
--- a/pcsx2/Counters.h
+++ b/pcsx2/Counters.h
@@ -93,8 +93,8 @@ struct SyncCounter
 //------------------------------------------------------------------
 #define FRAMERATE_NTSC			29.97 // frames per second
 
-#define SCANLINES_TOTAL_NTSC	525 // total number of scanlines
-#define SCANLINES_VSYNC_NTSC	3   // scanlines that are used for syncing every half-frame
+#define SCANLINES_TOTAL_NTSC 525    // total number of scanlines
+#define SCANLINES_VSYNC_NTSC 3     // scanlines that are used for syncing every half-frame
 #define SCANLINES_VRENDER_NTSC	240 // scanlines in a half-frame (because of interlacing)
 #define SCANLINES_VBLANK1_NTSC	19  // scanlines used for vblank1 (even interlace)
 #define SCANLINES_VBLANK2_NTSC	20  // scanlines used for vblank2 (odd interlace)
--- a/pcsx2/IopCommon.h
+++ b/pcsx2/IopCommon.h
@@ -28,7 +28,6 @@
 #include "IopCounters.h"
 #include "IopSio2.h"
 #include "IopGte.h"
-static const s64 PSXCLK = 36864000;	/* 36.864 Mhz */
 //#define PSXCLK	 9216000	/* 36.864 Mhz */
 //#define PSXCLK	186864000	/* 36.864 Mhz */
 
--- a/pcsx2/R3000A.cpp
+++ b/pcsx2/R3000A.cpp
@@ -63,7 +63,8 @@ void psxReset()
 	g_iopNextEventCycle = psxRegs.cycle + 4;
 
 	psxHwReset();
-
+	psxEEmultiplier = 8; /*Base EE to Iop Clock speed for the Playstation 2*/
+	PSXCLK = 36864000; /* 36.864 Mhz Playstation 2 speed*/
 	ioman::reset();
 	psxBiosReset();
 }
@@ -149,7 +150,7 @@ __fi void PSX_INT( IopEventId n, s32 ecycle )
 		// The EE called this int, so inform it to branch as needed:
 		// fixme - this doesn't take into account EE/IOP sync (the IOP may be running
 		// ahead or behind the EE as per the EEsCycles value)
-		s32 iopDelta = (g_iopNextEventCycle-psxRegs.cycle)*8;
+		s32 iopDelta = (s32)((g_iopNextEventCycle-psxRegs.cycle)*psxEEmultiplier);
 		cpuSetNextEventDelta( iopDelta );
 	}
 }
--- a/pcsx2/R5900.cpp
+++ b/pcsx2/R5900.cpp
@@ -59,6 +59,9 @@ u32 eeloadMain = 0;
 
 extern SysMainMemory& GetVmMemory();
 
+int psxEEmultiplier; //to fixed the unresolved external symbol. 
+
+
 void cpuReset()
 {
 	vu1Thread.WaitVU();
@@ -438,7 +441,10 @@ __fi void _cpuEventTest_Shared()
 
 	// The IOP could be running ahead/behind of us, so adjust the iop's next branch by its
 	// relative position to the EE (via EEsCycle)
-	cpuSetNextEventDelta( ((g_iopNextEventCycle-psxRegs.cycle)*8) - EEsCycle );
+
+	// One of two parts of an equation that determines the EE to Iop clock speed, note the value is trunacated.		                                                                                           
+	// due to the freaquency in PS1 mode being 8.71 that get's rounded up to 9. 
+	cpuSetNextEventDelta((s32)((g_iopNextEventCycle - psxRegs.cycle) * psxEEmultiplier) - EEsCycle); 
 
 	// Apply the hsync counter's nextCycle
 	cpuSetNextEvent( hsyncCounter.sCycle, hsyncCounter.CycleT );
--- a/pcsx2/x86/iR3000A.cpp
+++ b/pcsx2/x86/iR3000A.cpp
@@ -942,7 +942,10 @@ static void iPsxBranchTest(u32 newpc, u32 cpuBranch)
 		xMOV(ptr32[&psxRegs.cycle], eax); // update cycles
 
 		// jump if iopCycleEE <= 0  (iop's timeslice timed out, so time to return control to the EE)
-		xSUB(ptr32[&iopCycleEE], blockCycles*8);
+		// One of two parts of an equation that determins the EE to Iop clock speed, note the value is trunacated.		                                                                      
+		// due to the freaquency in PS1 mode being 8.71 that get's rounded up to 9. 
+		xSUB(ptr32[&iopCycleEE], (s32)(round(blockCycles * psxEEmultiplier)));
+
 		xJLE(iopExitRecompiledCode);
 
 		// check if an event is pending
@@ -992,7 +995,7 @@ void rpsxSYSCALL()
 	j8Ptr[0] = JE8(0);
 
 	xADD(ptr32[&psxRegs.cycle], psxScaleBlockCycles() );
-	xSUB(ptr32[&iopCycleEE], psxScaleBlockCycles()*8 );
+    xSUB(ptr32[&iopCycleEE], (s32)(psxScaleBlockCycles() * psxEEmultiplier));
 	JMP32((uptr)iopDispatcherReg - ( (uptr)x86Ptr + 5 ));
 
 	// jump target for skipping blockCycle updates
@@ -1014,7 +1017,7 @@ void rpsxBREAK()
 	xCMP(ptr32[&psxRegs.pc], psxpc-4);
 	j8Ptr[0] = JE8(0);
 	xADD(ptr32[&psxRegs.cycle], psxScaleBlockCycles() );
-	xSUB(ptr32[&iopCycleEE], psxScaleBlockCycles()*8 );
+	xSUB(ptr32[&iopCycleEE], (s32)(psxScaleBlockCycles()*psxEEmultiplier));
 	JMP32((uptr)iopDispatcherReg - ( (uptr)x86Ptr + 5 ));
 	x86SetJ8(j8Ptr[0]);
 
@@ -1284,7 +1287,7 @@ StartRecomp:
 		else
 		{
 			xADD(ptr32[&psxRegs.cycle], psxScaleBlockCycles() );
-			xSUB(ptr32[&iopCycleEE], psxScaleBlockCycles()*8 );
+            xSUB(ptr32[&iopCycleEE], (s32)(psxScaleBlockCycles() * psxEEmultiplier));
 		}
 
 		if (willbranch3 || !psxbranch) {
--- a/pcsx2/CDVD/CdRom.cpp
+++ b/pcsx2/CDVD/CdRom.cpp
@@ -82,15 +82,10 @@ u8 Test20[] = { 0x98, 0x06, 0x10, 0xC3 };
 u8 Test22[] = { 0x66, 0x6F, 0x72, 0x20, 0x45, 0x75, 0x72, 0x6F };
 u8 Test23[] = { 0x43, 0x58, 0x44, 0x32, 0x39 ,0x34, 0x30, 0x51 };
 
-// 1x = 75 sectors per second
-// PSXCLK = 1 sec in the ps
-// so (PSXCLK / 75) / BIAS = cdr read time (linuzappz)
-//#define cdReadTime ((PSXCLK / 75) / BIAS)
-u32 cdReadTime;// = ((PSXCLK / 75) / BIAS);
+u32 cdReadTime;
 
 #define CDR_INT(eCycle)    PSX_INT(IopEvt_Cdrom, eCycle)
 #define CDREAD_INT(eCycle) PSX_INT(IopEvt_CdromRead, eCycle)
-
 
 static void AddIrqQueue(u8 irq, u32 ecycle);
 
@@ -938,6 +933,15 @@ void psxDma3(u32 madr, u32 bcr, u32 chcr) {
 	HW_DMA3_CHCR &= ~0x01000000;
 	psxDmaInterrupt(3);
 }
+
+void setPsxSpeed()
+{
+    // 1x = 75 sectors per second
+    // PSXCLK = 1 sec in the ps
+    // so (PSXCLK / 75) / BIAS = cdr read time (linuzappz)
+    cdReadTime = ((PSXCLK / 75) / BIAS);
+}
+
 
 #ifdef ENABLE_NEW_IOPDMA
 s32 CALLBACK cdvdDmaRead(s32 channel, u32* data, u32 bytesLeft, u32* bytesProcessed)
--- a/pcsx2/CDVD/CdRom.h
+++ b/pcsx2/CDVD/CdRom.h
@@ -89,6 +89,7 @@ extern cdrStruct cdr;
 void cdrReset();
 void  cdrInterrupt();
 void  cdrReadInterrupt();
+void setPsxSpeed();
 u8   cdrRead0(void);
 u8   cdrRead1(void);
 u8   cdrRead2(void);
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -305,9 +305,9 @@ struct Pcsx2Config
 		int		FramesToDraw;	// number of consecutive frames (fields) to render
 		int		FramesToSkip;	// number of consecutive frames (fields) to skip
 
-		Fixed100	LimitScalar;
-		Fixed100	FramerateNTSC;
-		Fixed100	FrameratePAL;
+		Fixed100 LimitScalar;
+		Fixed100 FramerateNTSC;
+		Fixed100 FrameratePAL;
 
 		GSOptions();
 		void LoadSave( IniInterface& conf );
--- a/pcsx2/IopCounters.cpp
+++ b/pcsx2/IopCounters.cpp
@@ -32,9 +32,11 @@
 	 VBlank non-interlaced	59.82 Hz
 	 HBlank					15.73426573 KHz */
 
+s64 PSXCLK; //This is the definition of the Processor Clock for the PS1 externally declared in IopCommon.h
+
 // Misc IOP Clocks
-#define PSXPIXEL        ((int)(PSXCLK / 13500000))
-#define PSXSOUNDCLK		((int)(48000))
+#define PSXPIXEL          ((int)(PSXCLK / 13500000))
+#define PSXSOUNDCLK		  ((int)(48000))
 
 psxCounter psxCounters[NUM_COUNTERS];
 s32 psxNextCounter;
--- a/pcsx2/R3000AInterpreter.cpp
+++ b/pcsx2/R3000AInterpreter.cpp
@@ -144,7 +144,15 @@ static __fi void execI()
 	psxRegs.cycle++;
 	iopCycleEE-=8;
 
-	psxBSC[psxRegs.code >> 26]();
+    if ((psxHu32(HW_ICFG) & (1 << 3)))    
+	{
+        //this get's set to 9 because in the EE's half of this equation	                 
+		//Were trunacating the value 8.71 to a 9 instead due to casting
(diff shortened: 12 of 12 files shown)


kenshen112:
It's because of the floating point conversion I tried 8.71 with out a cast what so ever and experience similar effects.

pcsx2/x86/iR3000A.cpp
```diff
@@ -941,10 +941,16 @@ static void iPsxBranchTest(u32 newpc, u32 cpuBranch)
...
 		xMOV(ptr32[&psxRegs.cycle], eax); // update cycles
 
+		if ((psxHu32(HW_ICFG) & (1 << 3))) // if were in PS1 mode adjust ee accordingly.
+		{
+			psxEEmultiplier = 8.71;
+		}
+
 		// jump if iopCycleEE <= 0  (iop's timeslice timed out, so time to return control to the EE)
-		xSUB(ptr32[&iopCycleEE], blockCycles*8);
+		xSUB(ptr32[&iopCycleEE], blockCycles*static_cast<int>(psxEEmultiplier));
```
gregory38: You should explain why you "truncate" the multiplier instead of rounding. And why you do it before the multiplication.
i.e.
why not ?
```
static_cast<int>(blockCycles* psxEEmultiplier);
```
or
```
static_cast<int>(round(blockCycles* psxEEmultiplier));
```

ramapcsx2:
Reducing IOPs clock speed is not solved by this alone. It has to go in hand with slowing down all the components that are attached to it, mainly SPU2, but there could be more.
If we reduce just a part of the system, we will create more issues than we solve.
Yes, 36Mhz is too fast, but it is in correct relation to all the other system units.

kenshen112:
@ramapcsx2 with the above changes I've noticed an improvement in several pe1 games I've tested although I am aware that there is much more to be adjusted I thought this a good preliminary commit as this does slow games in good ways with as far as I've tested no real negatives aside from a slight desync which as you saying mainly due to sp2u if you'd like i could make further adjustments to the other clocks and timers.

pcsx2/x86/ix86-32/iR5900-32.cpp
```diff
@@ -1186,6 +1186,11 @@ void recMemcheck(u32 op, u32 bits, bool store)
 	xMOV(edx,eax);
 	xADD(edx,bits/8);
 
+	if ((psxHu32(HW_ICFG) & (1 << 3))) // if we're in PS1 mode adjust ee accordingly.
+	{
+		xADD(edx, (s32)(round(bits / 8.71)));
+	}
+
```
refractionpcsx2: This is just doubling it up, you're adding bits /8 then adding another bits/8.71, you need an else condition here.

pcsx2/x86/iR3000A.cpp
```diff
@@ -941,13 +941,13 @@ static void iPsxBranchTest(u32 newpc, u32 cpuBranch)
...
 
-		if ((psxHu32(HW_ICFG) & (1 << 3))) // if were in PS1 mode adjust ee accordingly.
+		if ((psxHu32(HW_ICFG) & (1 << 3))) // if we're in PS1 mode adjust ee accordingly.
 		{
 			psxEEmultiplier = 8.71;
 		}
 
 		// jump if iopCycleEE <= 0  (iop's timeslice timed out, so time to return control to the EE)
-		xSUB(ptr32[&iopCycleEE], blockCycles*static_cast<int>(psxEEmultiplier));
+		xSUB(ptr32[&iopCycleEE], (s32)(round(blockCycles* psxEEmultiplier)));
```
refractionpcsx2: does psxEEmultiplier default to 8? if it's uninitialised (as it looks to only be set above) it's going to crash when in PS2 mode.

pcsx2/R5900.cpp
```diff
@@ -440,7 +440,7 @@ __fi void _cpuEventTest_Shared()
 
 	if ((psxHu32(HW_ICFG) & (1 << 3))) // if were in PS1 mode adjust ee accordingly.
 	{
-		cpuSetNextEventDelta((static_cast<float>((g_iopNextEventCycle - psxRegs.cycle))*8.71) - EEsCycle);
+		cpuSetNextEventDelta((static_cast<int>((g_iopNextEventCycle - psxRegs.cycle)*8.71)) - EEsCycle);
```
refractionpcsx2: You need an else statement here to put the one in above that you didn't touch. No point in doing it twice.

pcsx2/x86/iR3000A.cpp
refractionpcsx2: Yeah, I'd like it to be set to 8 in psxReset() also, so if they start a new game (a PS2 one) the multiplier isn't off.

pcsx2/R3000A.cpp
```diff
@@ -63,8 +63,8 @@ void psxReset()
 	g_iopNextEventCycle = psxRegs.cycle + 4;
 
 	psxHwReset();
-	psxEEmultiplier = 8;
-	PSXCLK = 36864000;// to reset to default ps2 mode speed if someone decides to switch from ps1 to 2.
+	psxEEmultiplier = 8; // to reset to default ps2 mode speed if someone decides to switch from ps1 to 2.
+	PSXCLK = 36864000; // to reset to default ps2 mode speed if someone decides to switch from ps1 to 2.
 	ioman::reset();
```
refractionpcsx2: Just a note for when you next look at this, if you could move these 2 lines above the reset functions (maybe below the memzero(psxRegs) line), just for neatness, that'd be cool :)

kenshen112: Will do.

kenshen112:
Latest commit breaks recompiler entirely like ps2 mode doesn't work due to issues with cdReadTime, oddly interpreter works?

refractionpcsx2:
Yeah, don't move it to common.h either, there's no reason for the cd read speed to be in the common file, otheriwise it'll end up with loads in there, it's supposed to stay as little as posssible.

pcsx2/x86/ix86-32/iR5900-32.cpp
```diff
@@ -1184,7 +1184,15 @@ void recMemcheck(u32 op, u32 bits, bool store)
 	xFastCall((void*)standardizeBreakpointAddress, ecx);
 	xMOV(ecx,eax);
 	xMOV(edx,eax);
-	xADD(edx,bits/8);
+
+	if ((psxHu32(HW_ICFG) & (1 << 3))) // if we're in PS1 mode adjust ee accordingly.
```
gregory38: Hum I far as I understand, bits is the number of bits that will be read for the memory access. So it will be 8, 16, 32, 64. The division by 8 give you a size in byte. I don't understand why you need to change this stuff for PSX.

gregory38:
Hum, I don't follow anything but for the timing, we should have a virtual clock cycle number such as
```
EE= p * virtual_clk
IOP = n * virtual_clk
p and n are both integers and depends on the speed of IOP/EE.
```
If we really need accuracy we will need float, but it will potentially impact performance.

pcsx2/x86/ix86-32/iR5900-32.cpp
refractionpcsx2: That was my bad, mistook it for a timing function, it wasn't. My bad :P

kenshen112: It has to do with the frequency of IOP to EE even if it only returns size, before I made the change in that file and had only edited the interpreter multiplier recompiler was foobar by adjusting that it kinda started to work albeit with pgif errors

gregory38: I'm not even sure this code is used. It is related to debugger breakpoint.

pcsx2/IopCommon.h
```diff
@@ -28,7 +28,7 @@
 #include "IopCounters.h"
 #include "IopSio2.h"
 #include "IopGte.h"
-static const s64 PSXCLK = 36864000;	/* 36.864 Mhz */
+static s64 PSXCLK = 36864000;	/* 36.864 Mhz */
```
turtleli: `PSXCLK ` should not be `static`. There'll be multiple copies of the variable since it's in a header file, and as a result of that any value modifications will not be global (it'll only affect some uses of the variable), which I don't think is intended in this case.

kenshen112: That's how I found it, I can amend that though.

pcsx2/R5900.cpp
```diff
@@ -435,11 +438,11 @@ __fi void _cpuEventTest_Shared()
 		cpuSetNextEventDelta( 48 );
 		//Console.Warning( "EE ahead of the IOP -- Rapid Event!  %d", EEsCycle );
 	}
-
+
```
lightningterror: Tab

pcsx2/R5900.cpp
```diff
@@ -435,11 +438,11 @@ __fi void _cpuEventTest_Shared()
...
 		//Console.Warning( "EE ahead of the IOP -- Rapid Event!  %d", EEsCycle );
 	}
-
+	
 	// The IOP could be running ahead/behind of us, so adjust the iop's next branch by its
 	// relative position to the EE (via EEsCycle)
-	cpuSetNextEventDelta( ((g_iopNextEventCycle-psxRegs.cycle)*8) - EEsCycle );
-
+	cpuSetNextEventDelta(((s32)((g_iopNextEventCycle - psxRegs.cycle)*psxEEmultiplier)) - EEsCycle);
+
```
lightningterror: Ditto

lightningterror:
The commit descriptions could be worded a bit better to explain what was changed.

Also the commit title should begin with PSX: if the changes were done on psx or similar.

pcsx2/R5900.cpp
refractionpcsx2: Plus you've gone a bit overboard with the brackets

pcsx2/Common.h
```diff
@@ -20,6 +20,10 @@
 static const u32 BIAS = 2;				// Bus is half of the actual ps2 speed
 static const u32 PS2CLK = 294912000;	//hz	/* 294.912 mhz */
 
+extern int psxEEmultiplier; // This variable controls and set's the Multiplier value for the EE to Iop cycle clock,
```
lightningterror: I prefer if the comments are above the psxEEmultiplier, will look much more cleaner.

pcsx2/x86/iR3000A.cpp
```diff
@@ -942,7 +944,8 @@ static void iPsxBranchTest(u32 newpc, u32 cpuBranch)
 		xMOV(ptr32[&psxRegs.cycle], eax); // update cycles
 
 		// jump if iopCycleEE <= 0  (iop's timeslice timed out, so time to return control to the EE)
-		xSUB(ptr32[&iopCycleEE], blockCycles*8);
+		xSUB(ptr32[&iopCycleEE], (s32)(round(blockCycles* psxEEmultiplier))); // One of two parts of an equation that determins the EE to Iop clock speed, note the value is trunacated.
+		                                                                      // due to the freaquency in PS1 mode being 8.71 that get's rounded up to 9.
```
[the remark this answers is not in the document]
lightningterror: You forgot these as well.

kenshen112:
Latest build runs PSX games faster than normal pcsx2 this is due to traunacation with the EE multiplier causing EE usage to go up massively eliciting a speed up rather than the intended slowing of ps1 mode to regular speeds. Ways to solve, find a butter zone between IOP clock speeds and the aforementioned traunacation, change EE to run on floats or find a different close enough value with the EE multiplier

refractionpcsx2:
if the multiplier is truncated then it should be 8, which should be the same as it was before.  If it truncated to 9 then it would run slower.  So I don't think that's your issue.

pcsx2/Counters.h
```diff
@@ -85,17 +85,17 @@ struct SyncCounter
...
+#define HBLANK_COUNTER_SPEED	2 //Set to '3' to double the speed of games like KHII
 //#define HBLANK_TIMER_SLOWDOWN	1 //Set to '2' to increase the speed of games like God of War (FPS will be less, but game will be faster)
 
 //------------------------------------------------------------------
 // NTSC Timing Information!!! (some scanline info is guessed)
 //------------------------------------------------------------------
 #define FRAMERATE_NTSC			29.97 // frames per second
 
-#define SCANLINES_TOTAL_NTSC	525 // total number of scanlines
+#define SCANLINES_TOTAL_NTSC	263 // total number of scanlines
```
ramapcsx2: Did you mean to account for the PSX mostly running in "240p" mode?
If so, it can and does use full NTSC resolution as well.

pcsx2/Counters.h
```diff
@@ -85,17 +85,17 @@ struct SyncCounter
...
 #define SCANLINES_VSYNC_NTSC	3   // scanlines that are used for syncing every half-frame
-#define SCANLINES_VRENDER_NTSC	240 // scanlines in a half-frame (because of interlacing)
+#define SCANLINES_VRENDER_NTSC	120 // scanlines in a half-frame (because of interlacing)
```
ramapcsx2: This is wrong though, as a "half-frame" of active video in NTSC is definitely 240 vlines.

pcsx2/Pcsx2Config.cpp
```diff
@@ -210,8 +210,8 @@ Pcsx2Config::GSOptions::GSOptions()
 	FramesToSkip			= 2;
 
 	LimitScalar				= 1.0;
-	FramerateNTSC			= 59.94;
-	FrameratePAL			= 50.0;
+	FramerateNTSC			= 59.29;
```
ramapcsx2: I have never seen this number mentioned in a PSX doc (or in my own experiments).

pcsx2/Pcsx2Config.cpp
```diff
@@ -210,8 +210,8 @@ Pcsx2Config::GSOptions::GSOptions()
...
+	FrameratePAL			= 49.76;
```
ramapcsx2: This one might be right though, for PSX only. The change would touch / break PS2 mode, as do others in this PR currently.

pcsx2/Counters.h
```diff
@@ -85,17 +85,17 @@ struct SyncCounter
 //------------------------------------------------------------------
 // SPEED HACKS!!! (1 is normal) (They have inverse affects, only set 1 at a time)
 //------------------------------------------------------------------
-#define HBLANK_COUNTER_SPEED	1 //Set to '3' to double the speed of games like KHII
+#define HBLANK_COUNTER_SPEED	2 //Set to '3' to double the speed of games like KHII
```
ramapcsx2: Definitely not.

pcsx2/Counters.h
kenshen112: Yeah pretty much, I was trying to account 240p so would the default pcsx2 value be correct since it can use full NTSC resolution?

pcsx2/Counters.h
kenshen112: forgot to flip that back whoops was just messing around

pcsx2/Pcsx2Config.cpp
ssakash: I'm not sure why these values were changed. Explanation about the new values would be nice.

RedPanda4552: Something to keep in mind is that NTSC and PAL rates are *standards* that *directly interface with external devices*, e.g. TVs and monitors. Changing these doesn't really steer into a timing system that is correct and precise, but more *bends* the current implementation into a state where it *works* but isn't necessarily the most precise, correct or efficient. At least that's how I see it.

plugins/USBnull/USB.h
```diff
@@ -21,7 +21,6 @@
 #define USBdefs
 #include "PS2Edefs.h"
 #include "PS2Eext.h"
-
```
lightningterror: This was fine as it was, not sure why you removed it, and changing the code in USB plugin.

kenshen112:
A big development just happened. By adjusting the timings with Iop Several issues with sio particularly games like medievil and klonoa that explosively wouldn't load or save are working now.

kenshen112:
If you guys want to start giving this project a hefty look through with a fine tooth comb and a half before merging, with the medievil development I'd say it's ready for you guys

pcsx2/Counters.h
```diff
@@ -93,8 +93,8 @@ struct SyncCounter
 //------------------------------------------------------------------
 #define FRAMERATE_NTSC			29.97 // frames per second
 
-#define SCANLINES_TOTAL_NTSC	525 // total number of scanlines
-#define SCANLINES_VSYNC_NTSC	3   // scanlines that are used for syncing every half-frame
+#define SCANLINES_TOTAL_NTSC	525   // total number of scanlines
+#define SCANLINES_VSYNC_NTSC	3    // scanlines that are used for syncing every half-frame
 #define SCANLINES_VRENDER_NTSC	240 // scanlines in a half-frame (because of interlacing)
```
lightningterror: The spacing on the comments could be adjusted. -2 on the first row, -1 on the second.

kenshen112:
This is due to GitHub

pcsx2/Counters.h
kenshen112: @lightningterror this is due to GitHub

ramapcsx2:
So, basically this PR now only changes the EE multiplier if the system is in PSX mode, right?
It's a bit hard to review, due to all those white space differences.

Anyway, I saw nothing that would affect PS2 mode. That's good.
The PSX mode detection is still silly (doing it in that CDVD command), but it isn't different from what we have now.

All in all, the white space issue needs to be solved, so that people can easily review the changes.
Other than that, it looks okay to me.

kenshen112:
@ramapcsx2 should have fixed the whitespace issues

ramapcsx2:
@kenshen112 
Normalizing whitespaces with clang is great, but in this particular case I really meant reverting to existing code. This is to make reviewing easier, since then the diff will only show actual changes.
See https://github.com/PCSX2/pcsx2/pull/2407/files#diff-fbff03a1583eb7e3e96e215bf0d9fb5cR96
or https://github.com/PCSX2/pcsx2/pull/2407/files#diff-9d1716cecfcd928906c3f0be6b2859cdR38
for examples.

kenshen112:
@ramapcsx2 I've never used tortise git before, is there a good guide I could look at?

kenshen112:
@ramapcsx2 just got it done

ramapcsx2:
It's a bit odd that in real machines, the IOP clock is exactly 1/8th of the EE clock in PS2 mode, but when it goes to PSX mode, it becomes an odd multiple. 
Is anything about that documented somewhere?
33.868 * 8 is about 270.95Mhz.
Depending on how the actual machine is set up, either the EE is clocked down to that, or it stays at 294Mhz.
The difference matters in one place: The EE clock (of 294Mhz) is used to calculate the frame time in vSyncInfoCalc(). That frame time then goes on to dictate the intervals in which EE side stuff happens, so this would be really helpful to know.

Regarding this pr, there's still a some cleanup to do.
Everything that isn't actual code changes should be the same as in master.
This is necessary so that in the future, revisions of the work here can be easily done.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -1416,6 +1417,10 @@ static __fi void cdvdWrite0F(u8 rt) { // TYPE
 static __fi void cdvdWrite14(u8 rt) { // PS1 MODE?? // This should be done in the SBUS_F240 bit 19 write in HwWrite.cpp
 	u32 cycle = psxRegs.cycle;
 
+	PSXCLK = 33868800;      //33.8688 Mhz Playstation 1 speed.
+    psxEEmultiplier = 8.71; //proper EE to iop cycle freaquency.
```
ramapcsx2: What is the data type of psxEEmultiplier here? Integer, correct?
It is an external integer in the other files, so unless some C standard addition added some automatic promotion, this line says psxEEmultiplier = 8; ;p

kenshen112: It's a float value to hold the ps1's 8.71

ramapcsx2: Print out that value to console and see what it says.

kenshen112: It trunacates to a 9

ramapcsx2: Truncates is right but the value is 8. That's the default behaviour for when you fit a float into an integer.
Anyway, this part is obviously not working right, but it's still plausible that this pr fixes PSX games.
Must be the cycle correction then?

ramapcsx2: Well, it must be either the PSXCLK variable or the iopCycleEE -= 9. 
Those are the effective changes compared to master.

It looks like psxEEmultiplier used to be a float before?
In that case, this pr definitely needs to be re-tested, to see if it still does the thing ;)

refractionpcsx2: yeah the divider should be 8.71 but there's no mechanism in there currently to account for the remainder to accurately (or semi accurately) control the cycles consumed. So it gets rounded to 9 which is still closer than it was when it was fixed to 8 xD

ramapcsx2:
@refractionpcsx2 Do you have any notes of what the EE clock in PSX mode is? I don't think Sony ever mentioned this, simply because there's no need for anyone except maybe an emudev to know.

refractionpcsx2:
As far as I'm aware, the EE clock doesn't change, only the IOP switches.  If the EE downclocked too that would make life a LOT easier, but I've never seen any documentation to support that theory.

pcsx2/CDVD/CDVD.cpp
ramapcsx2: It's just that the code as is right now has psxEEmultiplier as an integer. The assignment of "8.71" to this is the same as psxEEmultiplier = 8. But from the earlier discussion, it looks like psxEEmultiplier used to be a float, so I'm asking :p

ramapcsx2:
Well, I guess someone is going to have to open a PS2 and hold a probe to the EE xD

refractionpcsx2:
Isn't the EE a BGA chip? It could be a bit of a pain that xD Plus I don't have a sillyscope :P

As I just said to PSI. Another way would be to force the PS2 in to PSX mode and start a timer on the EE at 1x speed, let it run for 50/60 VSyncs then read the value of the counter and multiply it by 2, that would get us a "rough" value.

But I don't know how to write that program or if it'll even work xD

pcsx2/CDVD/CDVD.cpp
kenshen112: It was, but it doesn't make a difference on the performance of this pro the effect a float vs an int has in this case from my tests have been the same

ramapcsx2: Yeah, I'm sure that changing timings will fix some games. I've seen this earlier with CDVD timing especially when I slowed that down (far too much, just as a test).
The task now is to make it correct, and at least the idea of this pr is the right direction.

We should start by figuring out what change in this pr is fixing the games.
Other than that, we already know that setting the value of PSXCLK dynamically is correct.
(And it doesn't require anything fancy.)

kenshen112: The interesting thing is that dma3 can't actually handle that speed in ff viii there's dma3 related hung timeouts that occur exclusively to this build and metal gear solid which previously worked I iop's interpreter no longer starts flashing bad opcodes in the console

refractionpcsx2: That could be because of this

cdReadTime = ((PSXCLK / 75) / BIAS);

That's a fixed read time, we don't know if that's right, plus it can vary depending on seek times, if the disc is spinning etc.

kenshen112: It is, they happen less if that's not set

refractionpcsx2: What happens less if it's not set?

kenshen112: Hung timeouts in ff viii

ramapcsx2:
I took the frequency counter to every trace I could find on the B side of an early PS2.
I never found an actual 294Mhz signal, so I think this clock is generated within the EE, or it's very well hidden.
There were several 392Mhz signals, must be the memory clock.
I got some connections that always carried a 33.868Mhz clock, some with 36.864, and one line did switch from 36.864 to 33.868 when loading a PSX title. I even found the near doubled PSX CPU clock at 67Mhz that drives the graphics RAM in PSX consoles :p

So... Too bad there EE clock couldn't be monitored, but nothing else suggested that it changes speed.
My best guess therefore: It runs at 294Mhz always and we should implement something like this PR.

pcsx2/CDVD/CDVD.cpp
kenshen112: so it turns out it's a combo of PSXCLK cdreadspeed and iopCycleEE -= 9 that breaks metal gear solid but fixes kolona, medievil was fixed before my tampering with it.

ramapcsx2:
Further tests regarding EE usage and "slowness" of the sound or video aren't necessary right now.
The biggest issue is an incomplete PGIF implementation that only works because of a huge hack with a DMA buffer. By tweaking some values, you can change how that hack plays out, but you can't really fix anything before this is solved.

To solve this, we need Wisi. He spent a lot of time reversing the low level stuff down to electrical signal levels. Only by using that knowledge, he could even come up with the current PGIF, wrong as it may be.
It's not much more than a prototype right now and until it is majorly improved, I would simply let this rest.

This PR tackles another issue that we could fix though. PSX mode timing is definitely wrong as well, but fixing it will only improve things ever so slightly.

kenshen112:
@ramapcsx2 Then let me ask you this, is there a specific aspect you'd like to see this pr work on? or do you have any qualms about this getting merged?

ramapcsx2:
I've logged this change and let it print out the new calculated cycles to console:
```
s32 iopDelta = (s32)((g_iopNextEventCycle-psxRegs.cycle)*psxEEmultiplier);
DevCon.WriteLn("new code's iopDelta: %d", iopDelta);
DevCon.WriteLn("old code's iopDelta: %d", (g_iopNextEventCycle-psxRegs.cycle)*8);
DevCon.WriteLn("---");
cpuSetNextEventDelta( iopDelta );
```
So it logs what the old and the new code calculates. This is the result:
```
new code's iopDelta: 23216
old code's iopDelta: 23216
---
new code's iopDelta: 304
old code's iopDelta: 304
---
new code's iopDelta: 25496
old code's iopDelta: 25496
---
new code's iopDelta: 5376
old code's iopDelta: 5376
---
new code's iopDelta: 58968
old code's iopDelta: 58968
```

This is obviously the result of psxEEmultiplier being just 8, since the fractional part got dropped instantly at initialization.
So this would have to be fixed first, along with the rest of formatting issues.
But this begs the question: What is actually fixing games in this pr?
I would recommend singling this out and creating a new pr / patch with just this change.
This pr as it is though? @sudonim1 would go mad if it got merged ;p

kenshen112:
@ramapcsx2 so it turns out it's a combo of PSXCLK cdreadspeed and iopCycleEE -= 9 that breaks metal gear solid but fixes kolona, medievil was fixed before my tampering with it

PR #2418, opened 2018-05-04
Gsdx: Adjust crc ids and purge crc hacks for Captain Tsubasa

lightningterror:
GSdx: Adjust crc ids.
Add a missing crc id for Captain Tsubasa-JP.
Rename WildArms5 to WildArms4 for crc 565B6170 Jap,
it was incorrect.

Gsdx: Remove Captain Tsubasa crc hacks.
The crc hack broke graphics ingame, causing flickering/transparent
textures and other similar issues.

On a side note the game experiences upscaling issues that can be fixed
with Half Pixel offset hack.

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -163,7 +163,7 @@ CRC::Game CRC::m_games[] =
 	{0xC1640D2C, WildArms5, US, 0},
 	{0x0FCF8FE4, WildArms5, EU, 0},
 	{0x2294D322, WildArms5, JP, 0},
-	{0x565B6170, WildArms5, JP, 0},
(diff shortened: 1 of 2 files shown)


turtleli:
So why we do keep game CRCs when they're not used anywhere else in the code? I don't understand it.

turtleli:
"Could be useful one day" means its useless right now and may not ever be used in the future. Having additional CRCs will add to memory usage and initialisation time. If the argument is that it could be useful some day, then why aren't all the game CRCs for every game being added?

lightningterror:
I'm fine with removing them, if everyone agrees with them getting purged then sure.

I guess I'll PR it and we can discuss about it.

lightningterror:
We could keep them away to avoid clutter.

PR #2421, opened 2018-05-05
gsdx: remove DirectCompute backend

turtleli:
It hasn't been accessible since OpenCL was added, and it was barely updated since it was introduced.


gregory38:
Yes we could remove it

PR #2422, opened 2018-05-05
GSdx: Purge crc ids/games no longer needed

lightningterror:
Following the discussion from #2418

The following games in the crc list are not used anywhere so
we can clean this list up. If some are needed in the future then
they can be re added.

List of removed games:
CaptainTsubasa,
Dororo,
HarvestMoon,
Jak1,
JamesBondEverythingOrNothing,
NamcoXCapcom,
SDGundamGGeneration,
SeintoSeiya,
SengokuBasara,
SilentHill2,
SilentHill3,
Siren,
TalesofDestiny,
VF4,
VF4EVO.


gregory38:
It would be nice if we can have an alternate solution than crc. Something based on gamedb.

gregory38:
Jak1 is strange. Are shadow ok ?

lightningterror:
I didn't test it, all I did was purge crc ids that aren't used.
Should I leave Jak1 ?

gregory38:
R&C and jak used similar engine that require an hack on gl for shadows. Or maybe I did something generic enough

lightningterror:
Yes but you updated the code so it doesn't require crc lookup anymore.

https://github.com/PCSX2/pcsx2/blob/c8c97312627aee6d88b913d6afff81ee2c0bb9a5/plugins/GSdx/GSRendererOGL.cpp#L1091

gregory38:
Ok.

lightningterror:
I did a quick test and Jak shadows seem to be fine on ogl.

PR #2424, opened 2018-05-05
Greg/xz : Add reading of game.iso.xz file

gregory38:
I'm sorry, code is in poor state but I don't have time to do better.

Code is based on xz 5.3. And I think some commit can be squashed/rebased (lzma useless option/define).

So far there is enough code to read an xz file. Uncompress a chunk into a temporary file. Maybe one would like to replace it with a cache. I barely tested it on Linux


gregory38:
@lightningterror @turtleli feel free to stole the code :)

gregory38:
@turtleli what are you plan for xz, if you have any ? Submodule was updated but main git still reference the stable version.

gregory38:
Idea is to use the xz version provided by PCSX2 git. So you will able yo use it without a release of xz in your distribution

gregory38:
Note: compression can be done with older version.However  You need to use the block mode

turtleli:
> what are you plan for xz, if you have any ? Submodule was updated but main git still reference the stable version.

No plan.

If you want to use a dev version then it might be better to force the use of the xz submodule on Linux if the xz reader is enabled just in case the API changes before 5.4.x is released (it's probably unlikely to happen, but it might prevent a compile error if it does).

turtleli:
I fixed the Windows build. Didn't make any other changes. Didn't test it either.

mirh:
Pcsx2 already support CSO? #504 
It has its fair deal of limitations though https://github.com/PCSX2/pcsx2/issues/2584#issuecomment-419927212
And I'm not seeing why deflate would be better than lzma anyhow.

lightningterror:
I'll close this as it's not currently being worked on, when work resumes it can be reopen.

Issue #1634, opened 2016-10-21
Valkyrie Profile 2: Silmeria [PAL-E] - A/V desyncs,  hard lock during FMV playback


trostboot:
**_PCSX2 version:**_
PCSX2 1.5.0-20161020041703

**_PCSX2 options:**_
Defaults, no speedhacks

**_Plugins used:**_
GSdx AVX2 D3D11 SW, Pokopom, SPU2-X (also tested with ZeroSPU2)

**_Description of the issue:**_
PAL versions of Valkyrie Profile 2 (version tested is PAL-E, SLES-54644, CRC 04CCB600, image md5 bb2e4fe43daae8b85bcbe4e69e11a9e2) suffer from severe audio/video desync during FMV playback. Audio is played back at normal speeds, video is played back faster and suffers from additional heavy stuttering.
Furthermore, if the intro FMV cinematic isn't skipped, the game will lock up at the end. 

**_How to reproduce the issue:**_
1. Boot up a PAL version of Valkyrie Profile 2

**_Last known version to work:**_
None. As far as I can tell, this never worked. There's the odd thread on the forums mentioning it going back many years, but usually alongside other issues so it got sidelined or classified as PEBKAC.

**_PC specifications:**_
Intel i5-4690k, AMD R9 270X, 16GB RAM, Win10 x64 14393)

[emuLog.txt](https://github.com/PCSX2/pcsx2/files/544998/emuLog.txt)

FlatOutPS2:
Go to Config > Audio (SPU2) > Plugin Settings, and set **Synchronizing Mode** to "Async Mix". Then try the FMVs again.

trostboot:
Already did. Sorry, should have been more clear. Changing sync modes does nothing. And as I stated, it also happens with ZeroSPU2.

willkuer:
Maybe try different delaycycle in spu2-x.ini. Something like 0-10. But probably unrelated if sound is working correctly.

trostboot:
Yeah, that doesn't change anything. I tried a few other settings just to be sure (EE/IOP interpreters, EE timing hack), but it doesn't make any difference.

jcdenton2k:
Try regular memory card files instead of memory card folders.  Also make sure you are using Safe or Safest presets and see what happens.  I'll see about getting someone with the game to help me test something on my end here.

trostboot:
The issue predates memory card folders by years and years. I tried it anyway to be sure, it has no effect.

[...]

PR #2426, opened 2018-05-06
Add a game fix to set EE Cyclerate at 100% while FMVs are playing

Siphonay:
This fixes FMVs being slowed down when EE Cyclerate is set at high values

(Edit on 05/12)
This fixes FMV cutscenes being played at a slower speed than usual when the EE cyclerate is set to high values on some games, notably the Ratchet & Clank series.

Another option would be to merge this functionality in the "FMV in Software Mode" fix and rename it to "FMV Safe Mode". I opened a pull request for that: #2435

### DIFF
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -59,6 +59,7 @@ enum GamefixId
 	Fix_FMVinSoftware,
 	Fix_GoemonTlbMiss,
 	Fix_ScarfaceIbit,
+	Fix_FMVEECycle,
 
 	GamefixId_COUNT
 };
@@ -361,7 +362,8 @@ struct Pcsx2Config
 				GIFFIFOHack		:1,		// Enabled the GIF FIFO (more correct but slower)
 				FMVinSoftwareHack:1,	// Toggle in and out of software rendering when an FMV runs.
 				GoemonTlbHack	:1,		// Gomeon tlb miss hack. The game need to access unmapped virtual address. Instead to handle it as exception, tlb are preloaded at startup
-				ScarfaceIbit 	:1;		// Scarface I bit hack. Needed to stop constant VU recompilation
+				ScarfaceIbit 	:1,		// Scarface I bit hack. Needed to stop constant VU recompilation
+				FMVEECycleHack  :1;     // Set the EE Cycle to 100% while playing a FMV cutscene
 		BITFIELD_END
 
 		GamefixOptions();
@@ -545,6 +547,7 @@ TraceLogFilters&				SetTraceConfig();
 #define CHECK_VIF1STALLHACK			(EmuConfig.Gamefixes.VIF1StallHack)  // Like above, processes FIFO data before the stall is allowed (to make sure data goes over).
 #define CHECK_GIFFIFOHACK			(EmuConfig.Gamefixes.GIFFIFOHack)	 // Enabled the GIF FIFO (more correct but slower)
 #define CHECK_FMVINSOFTWAREHACK	 	(EmuConfig.Gamefixes.FMVinSoftwareHack) // Toggle in and out of software rendering when an FMV runs.
+#define CHECK_FMVEECYCCLEHACK       (EmuConfig.Gamefixes.FMVEECycleHack) // Set the EE Cycle to 100% while playing a FMV cutscene
 //------------ Advanced Options!!! ---------------
 #define CHECK_VU_OVERFLOW			(EmuConfig.Cpu.Recompiler.vuOverflow)
 #define CHECK_VU_EXTRA_OVERFLOW		(EmuConfig.Cpu.Recompiler.vuExtraOverflow) // If enabled, Operands are clamped before being used in the VU recs
--- a/pcsx2/Pcsx2Config.cpp
+++ b/pcsx2/Pcsx2Config.cpp
@@ -267,7 +267,8 @@ const wxChar *const tbl_GamefixNames[] =
 	L"GIFFIFO",
 	L"FMVinSoftware",
 	L"GoemonTlb",
-	L"ScarfaceIbit"
+	L"ScarfaceIbit",
+	L"FMVEECycle"
 };
 
 const __fi wxChar* EnumToString( GamefixId id )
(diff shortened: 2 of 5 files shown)


pgert:
Can you give examples of games that makes use of this?
Can the hack be applied through GameIndex, and if so please add info about the hack in the Game Fixes info section.
"Set EE Cycle rate to 100% when a FMV plays" ought to be "Set the EE Cyclerate to 100% when an FMV plays".

refractionpcsx2:
I would kind of like to see this hack being merged with the software mode FMV hack so we don't have yet ANOTHER gamefix :P Both are related to FMV's and I don't see any harm in doing both just for them IMO.

Of course that depends what other people think..

Siphonay:
@pgert I've experienced this with the Ratchet & Clank and the Sly games. I guess this applies to all MPEG cutscenes but I haven't tested more games to back this claim up.

@refractionpcsx2 this is something I considered, but I haven't been able to figure out what would be best out of the two options, so I decided to put this pull request as is and see people's feedback, so thanks!
I also considered making this a speedhack rather than a gamefix, I feel like it would make more sense if how much cutscenes were slowed down was related with how powerful the machines the emulator is running on was, but since I only have two computers with similarly powerful CPUs, I obtained pretty much the same results on both so I can't say this is the case

refractionpcsx2:
Hmm no it would make less sense as a speedhack as you aren't increasing the speed in any way :P But I guess a gamefix on its own wouldn't make much sense either as you'd need to have the speedhack enabled in the first place.  That's a tough one.

I guess that's one reason to have a merged gamefix :P

Siphonay:
I can get behind the idea of a global FMV safe mode gamefix. Should I wait for more feedback before doing that?

refractionpcsx2:
Yes, I'm going to annoyingly tag some people to get some more feedback on the idea

@gregory38 @ssakash @KrossX @avih @sudonim1 @ramapcsx2 @lightningterror @MrCK1 @turtleli @FlatOutPS2 @gigaherz 

My dudes, what do you think? See my first comment in this thread.

MrCK1:
I don't want another gamefix, but I'd be fine if a compromise was found to merge it with an existing hack, ect.

Siphonay:
@MrCK1 
Well maybe this is something I should have detailed in my initial post but I've done full playthrough of Ratchet & Clank games on PCSX2, and the games all ran better with the EE Cyclerate all the way up to 300%, but this had the effect to slow down FMV cutscenes to around 80%, which made them pretty unenjoyable, so while they were playing I had to minimize the emulator window, and go reset the EE cyclerate to 100%. This fix is a way to automatize this process.

ssakash:
I like ref's idea of integrating it into a "FMV safe mode", seems more cleaner than adding various scenario specific hacks.

gregory38:
Could this new hack be always enabled ? Or is it unsafe ?

Siphonay:
If you mean unsafe as "emulation instability", I don't think setting the EE cyclerate to the default value is unsafe as it is the setting that (according to the UI) has the least compatibility problems.

It term of bugs, as of now: if you stop the emulation while a cutscene is playing, the cyclerate won't reset to what it was before the cutscene started, and if you change the cyclerate while a cutscene is playing, it will be reset to the one that was in effect before the cutscene started.

mirh:
> The we can call it something like "FMV Safe Mode" or something.

That's a smart idea. 
Though, correct me if I'm wrong, there's no "hard reason" for hardware mode to create problems with FMVs (and so that's kinda of a workaround, just like all other hacks AFAIU)
Instead touching timing.. I wonder if there are even FMVs that wouldn't have a different speed?

lightningterror:
Yes I like ref's idea as well, merge the two hacks in to one.

+1 for FMV Safe Mode

pcsx2/Config.h
```diff
@@ -326,7 +327,7 @@ struct Pcsx2Config
 
 				OpEqu( LimitScalar )			&&
 				OpEqu( FramerateNTSC )			&&
-				OpEqu( FrameratePAL )			&&
+		 		OpEqu( FrameratePAL )			&&
```
lightningterror: Please undo this change, the whitespace was fine.

pcsx2/Pcsx2Config.cpp
```diff
@@ -331,6 +332,7 @@ void Pcsx2Config::GamefixOptions::Set( GamefixId id, bool enabled )
 		case Fix_FMVinSoftware:	FMVinSoftwareHack	= enabled;  break;
 		case Fix_GoemonTlbMiss: GoemonTlbHack		= enabled;  break;
 		case Fix_ScarfaceIbit:  ScarfaceIbit        = enabled;  break;
+		case Fix_FMVEECycle:    FMVEECycleHack      = enabled; break;
```
lightningterror: Add an extra space here, it looks out of place compared to the rest above.

RedPanda4552:
+1 for merge. My issue with having it always on, is there may be some (although we're probably scraping the edge case barrel here) situations where someone has to play a game with the EE cyclerate anywhere from level -1 to -3 to hold full speed, and it kicking up to level 0 on a FMV may actually slow it down again, having the inverse effect as the hack intends. At least, if the trend of "More EE cycles = less FMV FPS" holds up.

refractionpcsx2:
I don't know if any cases that need a - ee cycle rate to hold full fps on a video, the ps2 was more than capable of video playback :p

lightningterror:
Well we can also change the gui option to a drop down list how vsync works, if there is an issue if the two are merged.

FMV mode: Switch to SW mode during FMV, Switch to SW + Cycle Rate, 100 Cyclerate.

pcsx2/gui/Panels/GameFixesPanel.cpp
```diff
@@ -107,6 +107,10 @@ Panels::GameFixesPanel::GameFixesPanel( wxWindow* parent )
 		{
 			_("VU I bit Hack avoid constant recompilation (Scarface The World Is Yours)"),
 			wxEmptyString
+		},
+		{
+			_("Set EE Cycle rate to 100% when a FMV plays"),
```
MrCK1: Clarification should be added here that the EE cyclerate will reset to its normal speed (100% isn't very clear)

The notable cases where it is needed should be noted as well, such as R&C.

KrossX:
Yet another +1 for FMV Safe Mode. Also, thread gets updated and people stay notified. =P

turtleli:
> I would kind of like to see this hack being merged with the software mode FMV hack so we don't have yet ANOTHER gamefix :P Both are related to FMV's and I don't see any harm in doing both just for them IMO. The we can call it something like "FMV Safe Mode" or something.

Well, the "software mode" hack switches renderer regardless of whether the current renderer is actually software or hardware. The behaviour of that hack might need to be changed first before merging the 100% cyclerate hack with the "software mode" hack together into an FMV safe mode hack.

> Could this new hack be always enabled ?

Might not be a good idea - see #1634 (Valkyrie Profile 2 PAL FMV locks up at the end unless EE cyclerate is -1 or lower).

Siphonay:
Since the consensus seems to be merging this with the Software FMV hack I think it would be better to close this pull request and do a new one (while also correcting the behavior of the renderer switch if I can figure out how to do that)

What do you think? (I'll try not to mess up indentation this time)

lightningterror:
Well if you want to start over sure, if it were me tho I'd drop the commits during rebase so I would be starting clean anyway without closing and creating a new PR.

refractionpcsx2:
> Well, the "software mode" hack switches renderer regardless of whether the current renderer is actually software or hardware. The behaviour of that hack might need to be changed first before merging the 100% cyclerate hack with the "software mode" hack together into an FMV safe mode hack.

TBH if you're already in software mode, you shouldn't be using the hack anyway, so there would currently be an issue with it being on.  So IMO that's outside the scope of this PR.

gregory38:
I hate to be that guy. But I don't think merging GSdx workaround + this EE timing is a good idea. In the end, it will always be enabled by user, for 1 issue or another.

Except few complex corner case, GSdx should be improved.

turtleli:
> TBH if you're already in software mode, you shouldn't be using the hack anyway, so there would currently be an issue with it being on. So IMO that's outside the scope of this PR, but should probably be looked in to regardless of the result of this.

The hack is in the GameDB for Silent Hill 2 and 3, so it's automatically switched on for any Software mode users playing those games. But yeah, the hack behaviour might be out of scope for this PR.

willkuer:
I'd suggest to enable it by default and disable it via ini setting. Accordingly no merge.

refractionpcsx2:
> Is there any reason to keep this open now? We do have #2435 which is an updated version of this PR unless I'm wrong.

Technically there is no reason, but suddenly people seem to be against the idea in the other version for no apparent reason :P

Siphonay:
@ssakash this one adds a separate gamefix and the other one merges it with an existing one, I kept both open so the PCSX2 team can decide which one to merge (if they decide to merge any)

ssakash:
To be honest, I prefer the other version more than this one. This doesn't seem like something worthy enough to be added as a separate gamefix and we do have lots of gamefixes now. :/

refractionpcsx2:
I completely agree with that.

Siphonay:
The consensus seems to be that we're going with the other PR, so I'm closing this one. Thank you all for your consideration! :)

PR #2431, opened 2018-05-10
Added the ScarfaceIbitHack to the Scarface NTSC-U discs.

pgert:
Added "same CRC" info to SLES-51192 & SLUS-20576.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -9223,6 +9223,7 @@ Compat = 3
 	patch=1,EE,001f70b4,word,1000000f
 	// Fixes flickering graphics (Thanks to Kian Stevens for finding this).
 	patch=1,EE,80000000,word,00000000
+	comment=This disc has the same CRC as SLUS-20576, the NTSC-U disc.
 [/patches]
 ---------------------------------------------
 Serial = SLES-51194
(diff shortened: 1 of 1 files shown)


pgert:
According to this post the NTSC-U disc SLUS-21111 makes use of ScarfaceIbitHack:
https://forums.pcsx2.net/Thread-Scarface-The-World-is-Yours-SLUS-21111-U--28713?pid=585856#pid585856
 and thus it's resonable to assume that the other NTSC-U disc SLUS-21492 also does it.

The Compat change from 4 to 5 for SLUS-21111 has been approved in the Compatibility list.
Since SLUS-21492 is a Collector's Edition, it is resonable to assume that the Compat for this disc also can be set to 5.

bin/GameIndex.dbf
```diff
@@ -9223,6 +9223,7 @@ Compat = 3
 	patch=1,EE,001f70b4,word,1000000f
 	// Fixes flickering graphics (Thanks to Kian Stevens for finding this).
 	patch=1,EE,80000000,word,00000000
+	comment=This disc has the same CRC as SLUS-20576, the NTSC-U disc.
```
lightningterror: I don't think the comments are necessary, the info is not really important for normal users.
Maybe a code comment under [patches = C5473413] would be better.

lightningterror:
@pgert can you make the changes so I can merge this ?

pgert:
The problem seems to have fixed itself.

Issue #2425, opened 2018-05-06
Segfault on Fedora 28


lucag73:
***PCSX2 version:***
Development build 8f6a3d9fe868190e4efbc62cc6860a156fc1f86d

***PCSX2 options:***
PCSX2 built with --debug on Fedora 28

***Description of the issue:***
Segfault on start; this is the output on the console

 Interface is initializing. Entering Pcsx2App::OnInit!
Applying operating system default language...
Command line parsing...
Command line parsed!
Segmentation fault (core dumped)

***How to reproduce the issue:***
compile and try to run PCSX2

***Last known version to work:***
This is related to  patch 15efe69e46969cb56ba2c4d5c66865a46e1bd317.
Using the library version of _xgetbv gives the problem; forcing the usage of the version in the source
file works properly.

***PC specifications:***
Intel(R) Core(TM) i7-6700HQ

I put some extra details as comments in https://github.com/PCSX2/pcsx2/pull/2419 but I suppose it is better to open a new issue (as this is quite different from that)

A quick dump of the core file gave this:

Core was generated by `./PCSX2'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0 x86capabilities::Identify (this=0x1) at /home/lucag/Src/EMU/tst/common/src/x86emitter/cpudetect.cpp:189
189 if (isIdentified)
with the following backtrace

Thread 1 (Thread 0xf5328e80 (LWP 4453)):
#0 0x0836ed2a in x86capabilities::Identify() (this=0x1) at /home/lucag/Src/EMU/tst/common/src/x86emitter/cpudetect.cpp:189
#1 0x0836f0bb in x86capabilities::CountCores() (this=0x1) at /home/lucag/Src/EMU/tst/common/src/x86emitter/cpudetect.cpp:156
#2 0x00000000 in ()

actually, the last instructions executed are
│0x836ed1f x86capabilities::Identify()+1 mov %esp,%ebp
│0x836ed21 x86capabilities::Identify()+3 push %edi
│0x836ed22 x86capabilities::Identify()+4 push %esi
│0x836ed23 x86capabilities::Identify()+5 push %ebx
│0x836ed24 x86capabilities::Identify()+6 sub $0x2c,%esp
│0x836ed27 x86capabilities::Identify()+9 mov 0x8(%ebp),%esi
│0x836ed2a x86capabilities::Identify()+12 cmpb $0x0,(%esi)

but here esi=0x1

gregory38:
[...]

PR #2434, opened 2018-05-11
common: Work around GCC8 _xgetbv bug

turtleli:
Work around the GCC _xgetbv bug.

See #2425.

EDIT: Assigned to myself since I need to add more info to the commit description.


turtleli:
I added the comment. I'll wait till it's fixed before adding code to use the built-in.

PR #2435, opened 2018-05-12
Rename the "FMV in software mode" game fix to "FMV Fix" and make it set the EE cyclerate to 100% in addition to it switching the rendering mode

Siphonay:
This is an alternative of #2426 with its functionality merged with the FMV in Software Mode fix, as suggested by various contributors. I'm leaving both up so the contributors can decide which one to merge if that happens.

This fixes FMV cutscenes being played at a slower speed than usual when the EE cyclerate is set to high values on some games, notably the Ratchet & Clank series.

### DIFF
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -56,7 +56,7 @@ enum GamefixId
 	Fix_VIFFIFO,
 	Fix_VIF1Stall,
 	Fix_GIFFIFO,
-	Fix_FMVinSoftware,
+	Fix_FMVFix,
 	Fix_GoemonTlbMiss,
 	Fix_ScarfaceIbit,
 
@@ -353,15 +353,15 @@ struct Pcsx2Config
 				XgKickHack		:1,		// Erementar Gerad, adds more delay to VU XGkick instructions. Corrects the color of some graphics, but breaks Tri-ace games and others.
 				IPUWaitHack     :1,		// FFX FMV, makes GIF flush before doing IPU work. Fixes bad graphics overlay.
 				EETimingHack	:1,		// General purpose timing hack.
-				SkipMPEGHack	:1,		// Skips MPEG videos (Katamari and other games need this)
+				SkipMPEGHack	:1,		// Skips MPEG videos (Katamari and other games need this).
 				OPHFlagHack		:1,		// Bleach Blade Battlers
 				DMABusyHack		:1,		// Denies writes to the DMAC when it's busy. This is correct behaviour but bad timing can cause problems.
 				VIFFIFOHack		:1,     // Pretends to fill the non-existant VIF FIFO Buffer.
 				VIF1StallHack   :1,     // Like above, processes FIFO data before the stall is allowed (to make sure data goes over).
-				GIFFIFOHack		:1,		// Enabled the GIF FIFO (more correct but slower)
-				FMVinSoftwareHack:1,	// Toggle in and out of software rendering when an FMV runs.
-				GoemonTlbHack	:1,		// Gomeon tlb miss hack. The game need to access unmapped virtual address. Instead to handle it as exception, tlb are preloaded at startup
-				ScarfaceIbit 	:1;		// Scarface I bit hack. Needed to stop constant VU recompilation
+				GIFFIFOHack		:1,		// Enabled the GIF FIFO (more correct but slower).
+				FMVFix			:1,		// Use GSdx software rendering & regular EE cyclerate for FMVs.
+				GoemonTlbHack	:1,		// Gomeon tlb miss hack. The game need to access unmapped virtual address. Instead to handle it as exception, tlb are preloaded at startup.
+				ScarfaceIbit 	:1;		// Scarface I bit hack. Needed to stop constant VU recompilation.
 		BITFIELD_END
 
 		GamefixOptions();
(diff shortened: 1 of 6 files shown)


mirh:
Ehrm.. I'm no dev as always, but squashing some commit wouldn't hurt I guess.

refractionpcsx2:
"Secure" sounds like you're saving it from security issue, or trying to hide something personal.  "Safe" implies it's being executed safely in a careful manner, which makes much more sense to me.

lightningterror:
I agree with Gregory.

refractionpcsx2:
Well the only situations given as examples where it may be unsafe, you don't need either hack anyway.

Why would it not be safe to enable it? The software mode hack has been fine for a long time and the EE Cycle speedhack can easily be switched without causing issues.

ssakash:
So this does seem like a suitable function for safe FMV mode, as it sets the renderer and cycle rate to the most safe value possible. (the default ones)

However @turtleli brought up a [point](https://github.com/PCSX2/pcsx2/pull/2426#issuecomment-387578834) that ``Valkyrie Profile 2`` needs a different cycle rate value to prevent a spinlock in the FMV. However this seems more like a really rare internal issue somewhere in the timings, this gamefix still does what it says and for exceptional cases like what turtleli mentioned, maybe the gamefix could be switched off instead?

If this is getting added, I'd suggest to look into the games which have this game fix entry in the GameDB and look for any conflicting case with the new behavior like what turtleli mentioned and handle those entries based on the test.

> My concern is that it isn't always safe to enable it.

I guess your concern is about the default cycle rate value and not software mode? This is the most safest timing for the default cyclerate we have for now, can't go any safer than this at the moment :P 

For now we have one case mentioned by turtleli which doesn't go well with this gamefix, if there are other games which produce similar results then we should fix the default cycle rate timing first before adding this, but this'll probably do fine for almost 90% of the games. (Totally assumption but you get the idea)

pgert:
"Use GSdx software rendering & full EE cyclerate when an FMV plays" sounds better to me.

refractionpcsx2:
In the case turteli mentioned, you don't need the software mode switch for fmvs, so that point is moot

gregory38:
Maybe call it "generic fmv" gamefix. 

An alternative is to always enable it (EE timing) except for few game that require a different settings. Just a random idea.

refractionpcsx2:
generic fmv fix sounds dull though :P  Might as well just rename it to "FMV Fix" if we're gonna do that, "Generic" makes it sound boring.

gregory38:
Yes fmv fix is nice too :)

pgert:
"Switch to GSdx software rendering and set the EE cyclerate to 100% when an FMV plays" is a long and confusing menuoption-textstring.

How about "Use GSdx software rendering & regular EE cyclerate for FMVs"?

refractionpcsx2:
meh, bit waffley for a title, that could be explained in the description.

lightningterror:
@refractionpcsx2 I didn't want the _fix_ to be always on, I wanted the two fixes to be merged together.

pcsx2/Config.h
```diff
@@ -359,7 +359,7 @@ struct Pcsx2Config
 				VIFFIFOHack		:1,     // Pretends to fill the non-existant VIF FIFO Buffer.
 				VIF1StallHack   :1,     // Like above, processes FIFO data before the stall is allowed (to make sure data goes over).
 				GIFFIFOHack		:1,		// Enabled the GIF FIFO (more correct but slower)
-				FMVinSoftwareHack:1,	// Toggle in and out of software rendering when an FMV runs.
+				FMVFix			:1,		// Use GSdx software rendering & regular EE cyclerate for FMVs
```
lightningterror: You forgot the period at the end.

turtleli:
I need time to review.

pcsx2/gui/Panels/GameFixesPanel.cpp
```diff
@@ -97,7 +97,7 @@ Panels::GameFixesPanel::GameFixesPanel( wxWindow* parent )
 			wxEmptyString
 		},
 		{
-			_("Switch to GSdx software rendering when an FMV plays"),
+			_("Use GSdx software rendering & regular EE cyclerate for FMVs"),
```
turtleli: The `&` doesn't display in the GUI. Also, I strongly prefer `and` instead of `&`.

pcsx2/gui/AppMain.cpp
```diff
@@ -529,8 +530,9 @@ void DoFmvSwitch(bool on)
 				viewport->DoResize();
 	}
 
-	if (EmuConfig.Gamefixes.FMVinSoftwareHack) {
+	if (EmuConfig.Gamefixes.FMVFix) {
 		ScopedCoreThreadPause paused_core(new SysExecEvent_SaveSinglePlugin(PluginId_GS));
+		g_Conf->EmuOptions.Speedhacks.EECycleRate = 0;
```
turtleli: If I close PCSX2 in the middle of an FMV with the hack enabled, the next time I reopen it the EE cyclerate will be set to 0 instead of whatever it was originally at. It might be better to avoid changing the ini value so unexpected changes do not occur.

pcsx2/gui/AppMain.cpp
```diff
@@ -559,6 +562,9 @@ void Pcsx2App::LogicalVsync()
 				DevCon.Warning("FMV off");
 				DoFmvSwitch(false);
 				FMVstarted = false;
+				ScopedCoreThreadPause paused_core(new SysExecEvent_SaveSinglePlugin(PluginId_GS));
+				g_Conf->EmuOptions.Speedhacks.EECycleRate = eecycle_backup;
+				paused_core.AllowResume();
```
turtleli: This causes the GS plugin to be suspended and resumed twice when an FMV ends - once in `DoFmvSwitch()`, once here. It should only be suspended and resumed once.

I think restoring the original EE cyclerate value can be done in `DoFmvSwitch()`.

pcsx2/gui/AppMain.cpp
Siphonay: What value do I need to access to change the EE cyclerate without touching the settings themselves?

Siphonay: Another potentially undesirable behavior of the rendering switch is that if the plugin is set to use software rendering by default, it will switch to hardware rendering.

Is there a value I can check  to avoid the switch when it is already in software mode?

turtleli: > What value do I need to access to change the EE cyclerate without touching the settings themselves?

Probably EmuConfig.Speedhacks.EECycleRate. Maybe adjusting `fixup` in `_ApplySettings` or related code would work (I haven't put much thought into this btw).

> Is there a value I can check to avoid the switch when it is already in software mode?

No. Doesn't currently exist.

gregory38:
A suggestion will be to extend the gs api to send an enter/exit fmv state. This way gsdx could handle the toggle properly. And potentially without a full switch, I'm curious if it would be possible to start both renderer, and select dynamically one of them.

lightningterror:
Status on this PR ?

Siphonay:
Sorry for the inactivity, I don't have much time to work on it and figure everything out yet :/ If anybody wants to help though, they would be welcome to

lightningterror:
@turtleli do you wanna help and finish this PR ?

Siphonay:
I'm sorry for not working further on it, I don't really feel familiar enough with PCSX2's codebase to implement this PR in a sensible way

lightningterror:
I'll close this for now since no work is done, it can be reopened if the OP wishes to continue.
Maybe @RedPanda4552 will continue and finish this up as well one day.

PR #2437, opened 2018-05-15
GSdx: Remove Hummer Badlands CRC hack

MrCK1:
This hack was used to remove shadow artifacts when upscaling. However, Large Framebuffer now does the same operation without removing the shadows entirely.

### DIFF
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -993,19 +993,6 @@ bool GSC_UltramanFightingEvolution(const GSFrameInfo& fi, int& skip)
 	return true;
 }
 
-bool GSC_HummerBadlands(const GSFrameInfo& fi, int& skip)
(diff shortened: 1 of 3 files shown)


plugins/GSdx/GSHwHack.cpp
```diff
@@ -993,18 +993,6 @@ bool GSC_UltramanFightingEvolution(const GSFrameInfo& fi, int& skip)
...
-	if(skip == 0)
-	{
-		if(fi.TME && (fi.FBP ==0x0a00) && (fi.TBP0 ==0x03200 || fi.TBP0==0x3700) && fi.FPSM == fi.TPSM && fi.TPSM == PSM_PSMCT32)
-		{
-			skip = 1;
-		}
-	}
-
-	return true;
-}
```
lightningterror: Extra row, should be removed.

PR #2438, opened 2018-05-16
Update widescreen patches/cheats.

lightningterror:
This syncs the cheats_ws.zip widescreen patches archive with Devina's
archive from 03 May 2018.

Compared to the previous content of cheats_ws.zip:
- 29 files modified
- 55 new files
- 3171 files identical to before
- 28 files that were removed from the forum archive that contained
hunting/fishing games have been re added. Separate archive provied by
FlatOut.

The zip now has a total of 3255 patch files, compared to 3200 before.

7zip with the Ultra compression method was used to compress the archive.


lightningterror:
6 months sounds good and an update before release to finalize things.
Otherwise skipping the 6 months update isn't ideal because it could take years to release a new version (example 1.4 was released a few years ago).

Anyway for compression xz sounds like a better alternative.

avih:
> Anyway for compression xz sounds like a better alternative.

You look at the problem through a keyhole. There are many factors to this issue, and better compression is not necessarily the main one.

The file needs to be reasonably easy to create, preferably such that users can also see what's inside, and PCSX2 obviously also has to be able to open it and extract patches from it.

But maybe more importantly, the fact that it's a big binary blob makes it very uncomfortable to update it frequently - and that's an issue.

It's not a big deal even if it updates once a year, as one could always get the latest archive from the forums and put it at their bin dir, or extract it to their `cheats_ws` dir.

IMO either we change the format such that the updates become incremental and then we can update it as frequently as we want, or we try to update it mostly towards a release.

Another option is to maintain the latest archive at the wiki on github, though Devina maintains a latest archive anyway at the forum, so maybe that's redundant.

gregory38:
Using a separate git/submodule for the cheats will be a so much better solution. Then we need just need to compress it in the install.

Submodule doesn't cost anything as we already use them for xz 3rdparty.

PR #2442, opened 2018-05-16
Null plugins: add/adjust GUI, remove unused/unnecessary code, add and use common config

turtleli:
Removes unused Windows specific stuff from FWnull and USBnull.


lightningterror:
Shouldn't the files you removed also need to be removed from the windows build itself ?

refractionpcsx2:
Sooo, how do we log stuff with them on windows now?

turtleli:
> Shouldn't the files you removed also need to be removed from the windows build itself ?

They're not part of the Windows build.

> Sooo, how do we log stuff with them on windows now?

On master you have to do it via the ini. I didn't notice this before I opened the PR, I assumed PluginNullConfigure was functional when it wasn't. I could perhaps add a messagebox to take care of enabling/disabling logging via the GUI (this would also cover Dev9null).

refractionpcsx2:
That would be handy, thanks :)

turtleli:
I changed my mind. A messagebox just doesn't feel right to me. Dialog it is.

turtleli:
Updated:
* Adds a Windows config dialog for Dev9null, FWnull and USBnull. Options are to log to console and log to file.
* Adjusts the Linux GUI to match the Windows GUI.
* Remove {DEV9,FW,USB}About() since they're unused.
* Remove PluginNullAbout() since that's unused.
* Removes the snprintf define since that's unnecessary now.
* Uses a common config system for DEV9null, FWnull and USBnull to avoid code duplication.

lightningterror:
A much nicer approach, good job.

bositman:
@ssakash Maybe write your issue with it instead of playing the thumbs down fairy? :P

turtleli:
Thumbs down but changes approved.

Maybe just me, but I'm finding that just a little bit absolutely impossible to understand.

PR #2443, opened 2018-05-19
Update README.md

killerjn:
please updated readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Development builds are also available from [our website](http://pcsx2.net/downlo
 ## Minimum
 * OS: Windows Vista SP2 or newer or GNU/Linux (32-bit or 64-bit)
 * CPU: Any that supports SSE2 (Pentium 4 and up, Athlon64 and up)
-* GPU: DirectX 10 GPU or better
+* GPU: DirectX 10 GPU or better (on linux Required Opengl 3.3)
 * RAM: 2GB or more
 
 ## Recommended


mirh:
If just it was as easy as a simple number with GL /s

killerjn:
I do not understand ?

lightningterror:
4.0 Software, 4.3 Hardware.

killerjn:
**I thought that the minimum required was opengl 3.3
according to this warning**
screenshot_20180519_141021

lightningterror:
Check https://github.com/PCSX2/pcsx2/blob/master/plugins/GSdx/GLLoader.cpp

there you will find which extensions are required/mandatory.

TheLastRar:
PCSX2 requires a number of extensions that aren't core to GL 3.3 (A driver can support it, but dosn't have to for compliance)
You can poke though GLLoader.cpp to see what extensions GSDX uses (and if they are mandatory).

If you where to recommend a GL version based on only core support, that that would be GL4.5 (GL_ARB_clip_control, GL_ARB_get_texture_sub_image & GL_ARB_texture_barrier)

mirh:
The thing is, some extensions might have been mandatorily included in some newer version, but for as much it, they could be supported by OpenGL 2 cards (anisotropic landing officially in 4.6 is an example)

So, I mean.. Maybe just mentioning the direct3d feature level \*isn't that bad idea\*

README.md
```diff
@@ -18,7 +18,7 @@ Development builds are also available from [our website](http://pcsx2.net/downlo
 ## Minimum
 * OS: Windows Vista SP2 or newer or GNU/Linux (32-bit or 64-bit)
 * CPU: Any that supports SSE2 (Pentium 4 and up, Athlon64 and up)
-* GPU: DirectX 10 GPU or better
+* GPU: DirectX 10 GPU or better (on linux Required Opengl 3.3)
```
baka0815: Its "Linux" with an uppercase "L", lowercase "r" in "required" and "OpenGL".

You might just rephrase it as "(or OpenGL 3.3 or better on Linux)"

mirh: "I'd just like to interject for a moment...."

baka0815: Sorry, I didn't follow the conversation. My bad.

lightningterror:
`GPU: DirectX 10 GPU or better`

Was enough imo.

mirh:
Isn't minimum D3D9 on Windows?

lightningterror:
D3D10 capable GPUs are able to run GL if the driver supports it, at least I think.

lightningterror:
I'll close this. There's no need to update the readme anyway at this time and the info provided is already there.
It will be updated when v1.6 is released.

PR #2447, opened 2018-05-21
PS1 mode: Adjustments to processor clock speed and CD read speed. 

kenshen112:
This is a simple PR that focused on making the core changes to pcsx2's ps1 mode to make games run at proper speeds. By adjusting the PSXCLK or IOP that represents the main ps1 processor to proper ps1 clock speeds and setting the cdreadspeed and EE to IOP delta timers. This PR fixes games like klonoa that had issues with disk and game speed causing glitches with memory cards or other various issues whilst this commit also add's issues with games like metal gear solid that won't get past the main logo's and do do do do da do part anymore

### DIFF
--- a/pcsx2/CDVD/CdRom.cpp
+++ b/pcsx2/CDVD/CdRom.cpp
@@ -82,7 +82,10 @@ u8 Test20[] = { 0x98, 0x06, 0x10, 0xC3 };
 u8 Test22[] = { 0x66, 0x6F, 0x72, 0x20, 0x45, 0x75, 0x72, 0x6F };
 u8 Test23[] = { 0x43, 0x58, 0x44, 0x32, 0x39 ,0x34, 0x30, 0x51 };
 
-u32 cdReadTime;
+// 1x = 75 sectors per second
+// PSXCLK = 1 sec in the ps
+// so (PSXCLK / 75) / BIAS = cdr read time (linuzappz)
+u32 cdReadTime;// = ((PSXCLK / 75) / BIAS);
 
 #define CDR_INT(eCycle)    PSX_INT(IopEvt_Cdrom, eCycle)
 #define CDREAD_INT(eCycle) PSX_INT(IopEvt_CdromRead, eCycle)
@@ -602,9 +605,6 @@ void cdrWrite0(u8 rt) {
 
 void setPsxSpeed()
 {
-	// 1x = 75 sectors per second
-	// PSXCLK = 1 sec in the ps
-	// so (PSXCLK / 75) / BIAS = cdr read time (linuzappz)
 	 cdReadTime = ((PSXCLK / 75) / BIAS);
 }
 
--- a/pcsx2/CDVD/CdRom.cpp
+++ b/pcsx2/CDVD/CdRom.cpp
@@ -82,11 +82,7 @@ u8 Test20[] = { 0x98, 0x06, 0x10, 0xC3 };
 u8 Test22[] = { 0x66, 0x6F, 0x72, 0x20, 0x45, 0x75, 0x72, 0x6F };
 u8 Test23[] = { 0x43, 0x58, 0x44, 0x32, 0x39 ,0x34, 0x30, 0x51 };
 
-// 1x = 75 sectors per second
-// PSXCLK = 1 sec in the ps
-// so (PSXCLK / 75) / BIAS = cdr read time (linuzappz)
-//#define cdReadTime ((PSXCLK / 75) / BIAS)
-u32 cdReadTime;// = ((PSXCLK / 75) / BIAS);
+u32 cdReadTime;
 
 #define CDR_INT(eCycle)    PSX_INT(IopEvt_Cdrom, eCycle)
 #define CDREAD_INT(eCycle) PSX_INT(IopEvt_CdromRead, eCycle)
@@ -602,6 +598,14 @@ void cdrWrite0(u8 rt) {
 		cdr.ParamC = 0;
 		cdr.ResultReady = 0;
 	}
+}
+
+void setPsxSpeed()
+{
+	// 1x = 75 sectors per second
+	// PSXCLK = 1 sec in the ps
+	// so (PSXCLK / 75) / BIAS = cdr read time (linuzappz)
+	 cdReadTime = ((PSXCLK / 75) / BIAS);
 }
 
 u8 cdrRead1(void) {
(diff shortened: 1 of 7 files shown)


pcsx2/CDVD/CdRom.cpp
```diff
@@ -82,11 +82,7 @@ u8 Test20[] = { 0x98, 0x06, 0x10, 0xC3 };
 u8 Test22[] = { 0x66, 0x6F, 0x72, 0x20, 0x45, 0x75, 0x72, 0x6F };
 u8 Test23[] = { 0x43, 0x58, 0x44, 0x32, 0x39 ,0x34, 0x30, 0x51 };
 
-// 1x = 75 sectors per second
-// PSXCLK = 1 sec in the ps
-// so (PSXCLK / 75) / BIAS = cdr read time (linuzappz)
-//#define cdReadTime ((PSXCLK / 75) / BIAS)
-u32 cdReadTime;// = ((PSXCLK / 75) / BIAS);
+u32 cdReadTime;
```
ssakash: The comment should still be here, VS IDE looks for comments around the variable declaration for variable tooltip when hovering over its instances on the code.

pcsx2/Common.h
```diff
@@ -19,6 +19,8 @@
 
 static const u32 BIAS = 2;				// Bus is half of the actual ps2 speed
 static const u32 PS2CLK = 294912000;	//hz	/* 294.912 mhz */
+static s64 PSXCLK = 36864000;	/* 36.864 Mhz */
```
ramapcsx2: @refractionpcsx2:
Should this be `extern s64 PSXCLK;` and, for example, in CDVD.cpp globally defined `s64 PSXCLK = 33868800;`?
I'm not sure.

refractionpcsx2: Hmm, not sure, I mean theoretically this is correct.  The alternative would be to have it in psxCommon.h and then have an extern reference in common.h.  I don't think it should be in CDVD.cpp though, since it is the global PSX clock, not a CDVD specific thing.

turtleli: Yeah, it must not be static, otherwise any value modifications will not be global.

ramapcsx2: I guess it's a minor thing but when I see such a global, I never know where to initialize it. Ref is right, it doesn't belong into CDVD.cpp either. But where should it go? :p

kenshen112: The issue is that I'm not getting an undefined external that clearly makes it obvious where to redefine it I'm getting a multiply defined symbol error

refractionpcsx2: Well it needs to be in either common.h or psxCommon.h, since it is the IOP clock speed, it only makes sense in those locations really, especially since the EE clock speed is in common.h :P

kenshen112: Exactly the issue is where to redefine it see my previous comment I edited it

lightningterror:
I don't like the idea of fixing 1 game but breaking another.
Which game is worse when broken ?

kenshen112:
@lightningterror Metal gear solid is because it goes from fully playable to nope in one commit, that being said While this "only" fixes kolona - that's as far as my testing has shown so their could be other games that get the fix so to speak.

pcsx2/Common.h
turtleli: Did you do as rama said? `extern s64 PSXCLK;` in Common.h and `s64 PSXCLK = 36864000;` in CDVD.cpp? (Let's just place it there for now even if it doesn't belong there, it can be moved later.)

The error sounds like you didn't add the `extern` in the header file.

pcsx2/R3000AInterpreter.cpp
```diff
@@ -144,6 +144,12 @@ static __fi void execI()
 	psxRegs.cycle++;
 	iopCycleEE-=8;
 
+	if ((psxHu32(HW_ICFG) & (1 << 3)))
+	{
+		//One of the Iop to EE delta clocks to be set in PS1 mode.
+		iopCycleEE-=9;
+	}
```
turtleli: Is the total of -17 intended?

kenshen112: Yes

pcsx2/Common.h
kenshen112: Done.

ramapcsx2:
I'm fine with the PR now and would merge it, unless there is opposition ;p
The fact that it breaks a game, fixes another is regrettable but expected.
The PR is useful, as it adds the more accurate PSX CPU clocks and thus fixes derived PSX counter frequencies.
Overall, we still need a better PGIF before really caring about game compatibility.
(But hey, I managed to talk to wisi so fingers crossed :) )

lightningterror:
Ok merge it when you can.

pcsx2/R3000AInterpreter.cpp
turtleli: Can you explain why?

kenshen112: @turteli because that adjusts the freaquency of IOP to EE communication ie. How many times the IOP converses with the EE and back. The normal ps2 setting is about 8 MHz whilst the ps1 is around 8.71 MHz but I set it to 9 because everything runs off of ints instead of floats.

turtleli: Immediately above the code you added is `iopCycleEE-=8;` You're subtracting another 9 here for a total of -17. So in PS2 mode it subtracts 8 and in PSX mode it subtracts 17. Are you sure that's right?

refractionpcsx2: I would say that isn't right. You've got the IOP running at about half the speed it should be.

kenshen112: The if statement is for ps1 mode meaning it'll only run in ps1 mode, I can ad an else real quick if you'd like though.

refractionpcsx2: Yes but the PS2 mode one is running as well.  You should have an else statement for the PS2 one if that's the case.

kenshen112: done

ramapcsx2: Ehr, sorry, that's confusing me now.
The `iopCycleEE-=8` was always there and should be correct for PS2 mode.
The new change was to add another 9 cycles for when it is in PSX mode, for a total of 17 cycles, right?

However, it makes more sense now:
8 cycles for PS2 mode, 9 cycles for PSX mode.
@kenshen112 Can you confirm this is what it should be?

kenshen112: @ramapcsx2 it is, infact by seperating the two the audio issues that were experienced with this build vanish. This PR is complete unless people have other issues?

PR #2454, opened 2018-05-24
cmake: Fix GSdx EGL build libX11-xcb linkage issue

### LINKED ISSUE #2360: GSdx with EGL does not link with libX11-xcb
***PCSX2 version:*** HEAD (b6a73f6db333d4edf98252990bde2b6141d57a79)

The GSdx plugin does not link with libX11-xcb when building with EGL_API=TRUE, where it uses XGetXCBConnection(3).

```sh-session
$ ./build.sh --release --egl --lto
$ bin/PCSX2
[...snip...]
Loading plugins from ~/code/pcsx2/bin/plugins...
[wx] ~/code/pcsx2/bin/plugins/libGSdx.so: undefined symbol: XGetXCBConnection
(p) (L) The configured GS plugin file is not a valid dynamic library(LoadCorePlugins) 

~/code/pcsx2/bin/plugins/libGSdx.so
```

Looks like d9e43b5d8d7870e81e5fe6e904007b1043a53baa introduced using XGetXCBConnection for EGL. Not that that's actually helpful in this case…

***PC specifications:***
* Arch Linux (x86-64)

***Notes:***
I did manage to convince GSdx to link with `-lX11-xcb` by manually editing the generated build.ninja. But of course, that's a terrible solution to force upon packagers. I don't know enough CMake-foo to figure out how to convince it to output the right flags. It doesn't look like CMake's FindX11.cmake actually looks for libX11-xcb…

turtleli:
Prevents a XGetXCBConnection undefined symbol error.

Fixes #2360.


auscompgeek:
This patch seems sensible, and does indeed fix the issue. Thanks @turtleli!

lightningterror:
@turtleli is this ready for merge ?

PR #2457, opened 2018-05-27
GameDB: additions and alterations of "same CRC" infos

pgert:


pgert:
@lightningterror
Doesn't this look better?

pgert:
Case closed.

PR #2465, opened 2018-05-31
GSdx: Partial port of format destination (dfmt)

lightningterror:
Port dfmt partially from GL to D3D10/11.
Use dfmt instead of fmt on d3d texture shuffle just
like on gl. It should be more accurate.

Update dfmt shader to recent gl code.

GSdx: Reformat PSselector shader list in GSDeviceDX.h.
Match it with gl code, easier to read and to know
what the shaders do.

### DIFF
--- a/plugins/GSdx/GSDeviceDX.h
+++ b/plugins/GSdx/GSDeviceDX.h
@@ -190,8 +190,9 @@ public:
 				uint32 point_sampler:1;
 				uint32 shuffle:1;
 				uint32 read_ba:1;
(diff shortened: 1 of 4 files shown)


atomic83GitHub:
Why did you close this?

gregory38:
Code is wrong. Don't worry though, gl implementation is different but dx shoud handle most cases.

lightningterror:
On a side note maybe I should have left the GSDeviceDX.h reformatting open for merge.

PR #2482, opened 2018-06-18
Added the Jak X Beta Trial Code and a patch that makes it able to boot.

PlasmaLight:


lightningterror:
For the future I recommend making changes and PRs using separate branches instead of master, easier for you to manage and keep separate PRs open.

PlasmaLight:
Thanks for the advice, though I doubt I will be making changes too often. If I find anything useful for the Jak games I will add them for sure.

PR #2483, opened 2018-06-23
Merge pull request #1 from PCSX2/master

ivan89el:
Update

PR #2484, opened 2018-06-23
update

ivan89el:
#2290 #2483


MrCK1:
I think you're trying to update your local branch right? If so, a PR is the wrong way to do that ;)

There are plenty of turorials on how to do this with the GUI or Git Shell, ect.

PR #2487, opened 2018-06-25
GameDB: Various patches update.

atomic83GitHub:
This PR brings some Patches to fix various unplayable games (made by Prafull). It also add some comment for Savage skies to warn users to use the interpreter + cache and missing rounding fix for some version of Shin Megami Tensei 3.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -278,6 +278,7 @@ Region = NTSC-Unk
 Serial = SCAJ-20015
 Name   = Shin Megami Tensei III: Nocturne
 Region = NTSC-Unk
+eeRoundMode = 0 // Ladder glitch in "Assembly of Nihilo B11" level.
 ---------------------------------------------
 Serial = SCAJ-20016
 Name   = Warrior of Argus
@@ -4861,6 +4862,7 @@ Region = NTSC-U
 Serial = SCUS-97217
 Name   = NBA Shootout 2003
 Region = NTSC-U
+EETimingHack = 1 // Fixes an issue with loading screens which cause the game to return on the Bios.
 ---------------------------------------------
 Serial = SCUS-97218
 Name   = Kiosk Demo Disc 2.06
@@ -8036,6 +8038,12 @@ Region = PAL-M5
 Serial = SLES-50713
 Name   = Freaky Flyers
 Region = PAL-E
+vuRoundMode = 1 // Fixes some bad effects.
+[patches = 6E62DE7B]
+	// Fixes hanging problem.
+	patch=1,EE,00132bd0,word,00000000
+	patch=1,EE,00132c88,word,00000000
+[/patches]
 ---------------------------------------------
 Serial = SLES-50714
 Name   = Defender
@@ -9421,6 +9429,11 @@ Serial = SLES-51290
 Name   = Sword of the Samurai
 Region = PAL-E
 Compat = 5
+---------------------------------------------
+Serial = SLES-51292
+Name   = Savage Skies
+Region = PAL-M5
+Compat = 4 // This game needs the EE cache to work which is only supported in "interpreter" mode.
 ---------------------------------------------
 Serial = SLES-51294
 Name   = XIII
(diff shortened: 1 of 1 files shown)


MrCK1:
I think it would be best to ask Prafull first. Some of these patches aren't 100% ready to be added to the codebase yet iirc. I also don't think the comment about the interpreter is necessary, we already have it documented elsewhere ;)

atomic83GitHub:
About the comments, I add them so the users can be aware of that problem without creating a post thread in the forum. And about the Patches, yeah I will send a message to Prafull the next time :).

avih:
> And about the Patches, yeah I will send a message to Prafull the next time :).

Why next time and not now?

prafullpcsx2:
I am OK with these patches as these are pretty stable ones although Harry Potter POA also requires m-bit build to run with mvu.

atomic83GitHub:
@avih Because it's already done.

bin/GameIndex.dbf
```diff
@@ -42018,6 +42057,11 @@ Serial = SLUS-21697
 Name   = Iridium Runners
 Region = NTSC-U
 Compat = 3
+EETimingHack = 1
```
lightningterror: A comment explaining why the hack is needed.

bin/GameIndex.dbf
```diff
@@ -15343,6 +15366,11 @@ Region = PAL-M5
 Serial = SLES-54021
 Name   = Ruff Trigger
 Region = PAL-M5
+EETimingHack = 1
+[patches = b02c81e5]
+	// Fixes random hangs when going ingame.
+	patch=1,EE,001128e0,word,00000000
+[/Patches]
```
lightningterror: I prefer [patches], [/patches] not to have capital letters, check the other changes below as well.

bin/GameIndex.dbf
```diff
@@ -27489,6 +27518,7 @@ Serial = SLPM-66681
 Name   = Shin Megami Tensei - Nocturne Maniax
 Region = NTSC-J
 Compat = 5
+eeRoundMode = 0 // Ladder glitch in "Assembly of Nihilo B11" level
```
lightningterror: Period.

bin/GameIndex.dbf
```diff
@@ -278,6 +278,7 @@ Region = NTSC-Unk
 Serial = SCAJ-20015
 Name   = Shin Megami Tensei III: Nocturne
 Region = NTSC-Unk
+eeRoundMode = 0 // Ladder glitch in "Assembly of Nihilo B11" level
```
lightningterror: Period.

lightningterror:
Maybe you can also check out 
https://forums.pcsx2.net/Thread-Solved-klonoa-2-lunatea-s-veil-Keeps-crashing

Those patches might need to be added if they are missing.

pgert:
"This game need the EE cache"... should be "This game needs the EE cache".
"needs", not "need".

bin/GameIndex.dbf
ssakash: Mate, this is just a comment in a db file. Not a statement in an essay or something,  this is too nitpicky. :/

lightningterror: For the past few months we've been trying to keep the DB a bit cleaner and proper, let's keep it that way, besides there are already pending changes so why not do these as well while at it.

ssakash: I'm not sure how the lack of a period in a single statement is not clean for you, I'd agree with you in case it's a multiple line comment without punctuation but that's not the case here. Even with or without the period, the readability is not affected in any way.

If you really want to keep the GameDB clean, maybe notice the grammatical mistakes like ``This game need the EE cache `` in the commit which are more annoying than the lack of a period without significant meaning. Also there's no standard rule saying that a comment must have a full stop, let's not give extra work for the contributor based on personal opinion when they haven't done anything wrong.

And also I think we're fussing too much about keeping GameDB clean, I'd be concerned if it's a state like this,

```
Serial = SLES-50713
 Name   =              Freaky Flyers
         Region =         PAL-E
vuRoundMode = 1       // Fixes some bad effects.
```

rather than something like this which you changed in your recent GameDB clean up commit
```patch
-// allows import of Xenosaga I, Xenosaga I Reloaded, and Xenosaga Freaks data
+// Allows import of Xenosaga I, Xenosaga I Reloaded, and Xenosaga Freaks data.
```

Overall I feel like it's a really minor thing which doesn't need attention on, focusing on the functional change over these meaningless syntaxes would be much better in my opinion.

bositman: Agreed, I don't think anyone without high levels of OCD cares if the comment in the gamedb file misses a period. Don't get me wrong I'm all for improving and cleaning it up, but as said above there are various other stuff that need fixing and are certainly much more important than some periods :P

atomic83GitHub:
@lightningterror The GameDB already have the Klonoa 2 patch.

bin/GameIndex.dbf
```diff
@@ -11692,6 +11705,17 @@ Serial = SLES-52440
 Name   = Harry Potter and the Prisoner of Azkaban
 Region = PAL-M7
 Compat = 3
+EETimingHack = 1
```
lightningterror: Same as before, explaining why the hack is needed.

bin/GameIndex.dbf
```diff
@@ -15343,6 +15367,11 @@ Region = PAL-M5
 Serial = SLES-54021
 Name   = Ruff Trigger
 Region = PAL-M5
+EETimingHack = 1
```
lightningterror: Ditto.

PR #2490, opened 2018-06-30
GameDB: Drakengard 2 fixes, Mvu flag fix for Full Spectrum Warrior and missing serials.

atomic83GitHub:
This PR add an Emotion Engine clamping fix for Drakengard 2 (tested by Shadow Lady), an Mvu flag fix for Full Spectrum Warrior (Tested by CK1) along with missing serials entries.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -718,6 +718,7 @@ Region = NTSC-Unk
 Serial = SCAJ-20132
 Name   = Drag-On Dragon 2 - Fuuin no Kurenai (Love Red, Ambivalence Black)
 Region = NTSC-J
+eeClampMode = 2 // Fixes wrong color on some characters and breakable objects.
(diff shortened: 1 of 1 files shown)


bositman:
I don't have either game so I can't speak about the fixes working or not.
What I can say is 1) you added a new memcard filter for folder memory cards, is it needed and if yes you should mention it in the PR description, 2) you shouldn't change the compatibility status of games manually, it will be added automatically when we run the script that gets that info from the compatibility threads of the forum (which is the proper source)

atomic83GitHub:
Everything should be fine now.

lightningterror:
Is it fine now @bositman ?

lightningterror:
You don't need to use `[skip ci]` when doing future PRs, it's best used for changes that are not part of the pcsx2 solution that don't get automatically compiled.

PR #2492, opened 2018-07-02
GSdx: Adjust Ghost In The Shell SOC crc hack

ivan89el:
Ease the crc hack and skip less effects.

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -226,6 +226,7 @@ CRC::Game CRC::m_games[] =
 	{0x5D482F18, JackieChanAdv, EU, 0},
 	{0xAC4DFD5A, JackieChanAdv, EU, 0},
 	{0x95CC86EF, GiTS, US, 0}, // same CRC also reported as EU
+	{0x2C5BF134, GiTS, US, 0}, // Demo
 	{0xA5768F53, GiTS, JP, 0},
 	{0xA3643EB1, GiTS, KO, 0},
 	{0xBF6F101F, GiTS, EU, 0}, // same CRC as another US disc
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -139,7 +139,7 @@ bool GSC_GiTS(const GSFrameInfo& fi, int& skip)
 {
 	if(skip == 0)
 	{
-		if(fi.TME && fi.FBP == 0x03000 && fi.FPSM == PSM_PSMCT32 && fi.TPSM == PSM_PSMT8)
+		if(fi.TME && fi.FBP == 0x3000 && fi.FPSM == PSM_PSMCT32 && fi.TPSM == PSM_PSMT8)
 		{
 			skip = 9;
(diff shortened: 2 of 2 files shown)


ivan89el:
@lightningterror What should be done?

lightningterror:
pcsx2 likes to mess with me it seems and didn't like to execute the crc hack properly, anyway the code is fine.

A small nitpick for `FBP == 0x3000` maybe change it to `FBP == 0x03000` so it's consistent with other crc hacks since they are like that.

You can also add this crc id while you're at it.
`{0x2C5BF134, GiTS, US, 0}, // Demo`
Just below this line 
https://github.com/PCSX2/pcsx2/blob/1fefb4897e575dae0ffe31dd0e723aa01ea2493b/plugins/GSdx/GSCrc.cpp#L228


Can you add a proper commit title and description btw?

> title: GSdx: Adjust Ghost In The Shell SOC crc hack.
> description: Ease the crc hack and skip less effects.

lightningterror:
I just noticed that the game looks a bit worse on dx9 now (slightly darker) because shadows aren't working there. It's probably fine tho since it isn't an issue on dx10+. and nobody cares about dx9 anyway.

ivan89el:
Yes, unfortunately, dx9 looks bad (there are still problems). Thanks to one of your commits, dx11 became better in this game. It seems that this commit: https://github.com/PCSX2/pcsx2/commit/0e83b89dbddee367bdf9ee42cbce779a2f60ea98

lightningterror:
That means it's a texture shuffle effect, the darker image you see on dx9.

I've been porting fixes from gl here and there but there are still a lot remaining.

mkrupczak3:
Playing Ghost in the Shell SAC tonight, had issues with hardware mode still on 1.6.0. 

Was this broken again by a regression? @ivan89el @lightningterror ?

lightningterror:
It did made it in 1.6, what is the issue? Can you upload a gsdump?

mkrupczak3:
Screenshot from 2020-12-22 05-09-40
Screenshot from 2020-12-22 05-10-06

Hmm. Must be an issue with the default configuration of PCSX2, sorry. Doesn't quite work out of the box with OpenGL.

I did these settings changes, and it appears to be working okay for now. 

Maybe at some point, PCSX2 could do like dolphin emulator and auto-select all the relevant hacks/config for certain games? One can dream. Everytime I load up PCSX2, I wonder if spaceship controls are this complicated.

mkrupczak3:
Nice work btw. Thanks for the fixes. I changed the CRC hack level from automatic to full, even though I'm using OpenGL.

I was experiencing issues before I increased the CRC hack level, but on brief testing they've gone away. Is this intended behavior or an oversight with the CRC and OpenGL?

lightningterror:
Automatic should work.

mkrupczak3:
Ah. For me on 1.6.0 with OpenGL, automatic settings weren't working. 

Also, I had to adjust the Half-Pixel offset to fix some lighting effects, but none fixed the alignment quite properly. Vertex was the closest. May need to create this as an issue.

PR #2494, opened 2018-07-02
GSdx: Add crc id for Ghost In The Shell SOC Demo

ivan89el:


lightningterror:
I meant in the same PR not create a separate one.

PR #2502, opened 2018-07-09
GSdx: Adjust Kunoichi crc hack

ivan89el:
Removes only depth effects(shadows) not rendered correctly on all renders, without touching burning air effect. Move it to an aggressive hack, and added description of the problem.

### DIFF
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -750,13 +750,13 @@ bool GSC_TalesOfLegendia(const GSFrameInfo& fi, int& skip)
 	return true;
 }
 
-// Removes depth effects(shadows) not rendered correctly on all renders.
 bool GSC_Kunoichi(const GSFrameInfo& fi, int& skip)
 {
 	if(skip == 0)
 	{
 		if(!fi.TME && (fi.FBP == 0x0 || fi.FBP == 0x00700 || fi.FBP == 0x00800) && fi.FPSM == PSM_PSMCT32 && fi.FBMSK == 0x00FFFFFF)
 		{
+			// Removes depth effects(shadows) not rendered correctly on all renders.
 			skip = 3;
 		}
 	}
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -750,6 +750,7 @@ bool GSC_TalesOfLegendia(const GSFrameInfo& fi, int& skip)
 	return true;
 }
 
+//Shadows don't appear or cause darkness issues in different levels.
 bool GSC_Kunoichi(const GSFrameInfo& fi, int& skip)
 {
 	if(skip == 0)
@@ -757,27 +758,6 @@ bool GSC_Kunoichi(const GSFrameInfo& fi, int& skip)
 		if(!fi.TME && (fi.FBP == 0x0 || fi.FBP == 0x00700 || fi.FBP == 0x00800) && fi.FPSM == PSM_PSMCT32 && fi.FBMSK == 0x00FFFFFF)
 		{
 			skip = 3;
-		}
-		if(fi.TME && (fi.FBP ==0x0700 || fi.FBP==0) && fi.TBP0==0x0e00 && fi.TPSM ==0  && fi.FBMSK == 0)
-		{
-			skip = 1;
-		}
-		if(fi.TME)
-		{
-			// depth textures (bully, mgs3s1 intro, Front Mission 5)
-			if( (fi.TPSM == PSM_PSMZ32 || fi.TPSM == PSM_PSMZ24 || fi.TPSM == PSM_PSMZ16 || fi.TPSM == PSM_PSMZ16S) ||
-				// General, often problematic post processing
-				(GSUtil::HasSharedBits(fi.FBP, fi.FPSM, fi.TBP0, fi.TPSM)) )
-			{
-				skip = 1;
-			}
-		}
-	}
-	else
-	{
-		if(fi.TME && (fi.FBP == 0x0e00) && fi.FPSM == PSM_PSMCT32 && fi.FBMSK == 0xFF000000)
-		{
-			skip = 0;
 		}
 	}
 
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -766,7 +766,7 @@ bool GSC_Kunoichi(const GSFrameInfo& fi, int& skip)
 		if(fi.TME)
 		{
 			// Removes burning air effect, the effect causes major slowdowns.
-			if( (Aggressive && (fi.TPSM == PSM_PSMZ32 || fi.TPSM == PSM_PSMZ24 || fi.TPSM == PSM_PSMZ16 || fi.TPSM == PSM_PSMZ16S) ||
+			if(Aggressive && (fi.TPSM == PSM_PSMZ32 || fi.TPSM == PSM_PSMZ24 || fi.TPSM == PSM_PSMZ16 || fi.TPSM == PSM_PSMZ16S) ||
 				// General, often problematic post processing
 				(GSUtil::HasSharedBits(fi.FBP, fi.FPSM, fi.TBP0, fi.TPSM)) )
 			{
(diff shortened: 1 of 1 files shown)


plugins/GSdx/GSHwHack.cpp
```diff
@@ -750,6 +750,7 @@ bool GSC_TalesOfLegendia(const GSFrameInfo& fi, int& skip)
 	return true;
 }
 
+//Shadows don't appear or cause darkness issues in different levels.
```
MrCK1: This comment is better placed inside the CRC hack.

lightningterror:
I'll need to test it tomorrow or the day after to make sure.

plugins/GSdx/GSHwHack.cpp
pgert: "causes" or "may cause".

lightningterror: The wording is fine.

lightningterror:
I tested the first level and briefly the second and it works fine.

How many levels did you test ? Were there no issues at all ?

plugins/GSdx/GSHwHack.cpp
lightningterror: Well if we are gonna change the wording it should be.

"// Removes depth effects not rendered correctly on all renders."

ivan89el:
( (fi.TPSM == PSM_PSMZ32 || fi.TPSM == PSM_PSMZ24 || fi.TPSM == PSM_PSMZ16 || fi.TPSM == PSM_PSMZ16S) ||
				// General, often problematic post processing
				(GSUtil::HasSharedBits(fi.FBP, fi.FPSM, fi.TBP0, fi.TPSM)) )
			{
				skip = 1;
deleted post-processing, which is perfectly emulated

ivan89el:
if(fi.TME && (fi.FBP ==0x0700 || fi.FBP==0) && fi.TBP0==0x0e00 && fi.TPSM ==0  && fi.FBMSK == 0)
		{
			skip = 1;
Probably old code and superfluous. I checked it separately and he did not change anything.

ivan89el:
I checked 3 levels and they all worked well

ivan89el:
I almost forgot
if(fi.TME && (fi.FBP == 0x0e00) && fi.FPSM == PSM_PSMCT32 && fi.FBMSK == 0xFF000000)	
{	
skip = 0;
Probably does nothing (skip = 0)

lightningterror:
skip = 0 is used not to skip a certain effect because crc hacks can skip multiple effects so skip 0 is used to isolate and not skip a certain effect that shouldn't.

Maybe you should leave that in just in case.

ivan89el:
FBP == 0x0e00 not used. I left it just in case

lightningterror:
How's the speed of the game with the new changes ?

ivan89el:
The load on the GPU slightly increased due to post-processing (the load on the GPU slightly increased due to post-processing (it can be disabled to any user skipdraw = 1)

ivan89el:
Just in case, I returned everything except post-processing.

lightningterror:
I'll check it Soon™.

Perhaps the post processing effect was skipped because it caused major slowdowns.
60% for me.

ivan89el:
Perhaps, but if you have good characteristics then why should you play with stripped-down graphics. There are many games in which post-processing causes slowdowns (MK SM, Def Jam: Fight for NY), but it works and the desire can be turned off (skipdraw 1)

bositman:
If it was removed for that, it could be set to the aggressive group like others have been in the past (seems best fit there)

lightningterror:
I tested this game a few months back but didn't touch the crc hack for that same reason.

At this point it's better to move it to aggressive rather than removing it.

plugins/GSdx/GSHwHack.cpp
```diff
@@ -770,7 +770,19 @@ bool GSC_Kunoichi(const GSFrameInfo& fi, int& skip)
 			skip = 0;
 		}
 	}
-
+	else(Aggressive && skip == 0)
+	{
+		if(fi.TME)
+		{
+			// depth textures (bully, mgs3s1 intro, Front Mission 5)
+			if( (fi.TPSM == PSM_PSMZ32 || fi.TPSM == PSM_PSMZ24 || fi.TPSM == PSM_PSMZ16 || fi.TPSM == PSM_PSMZ16S) ||
```
lightningterror: I'd put Aggressive here.

`if(Aggressive && (fi.TPSM == PSM_PSMZ32`

plugins/GSdx/GSHwHack.cpp
```diff
@@ -765,7 +765,7 @@ bool GSC_Kunoichi(const GSFrameInfo& fi, int& skip)
 		if(fi.TME)
 		{
 			// depth textures (bully, mgs3s1 intro, Front Mission 5)
-			if( (fi.TPSM == PSM_PSMZ32 || fi.TPSM == PSM_PSMZ24 || fi.TPSM == PSM_PSMZ16 || fi.TPSM == PSM_PSMZ16S) ||
+			if( (Aggressive && fi.TPSM == PSM_PSMZ32 || fi.TPSM == PSM_PSMZ24 || fi.TPSM == PSM_PSMZ16 || fi.TPSM == PSM_PSMZ16S) ||
```
lightningterror: It's still incorrect, should be

`if(Aggressive && (fi.TPSM == PSM_PSMZ32`

lightningterror:
Also can you put back
`// Removes depth effects(shadows) not rendered correctly on all renders.`

And add this for the post processing effect.
`// Removes burning air effect, the effect causes major slowdowns.`

ivan89el:
Does not work.

ivan89el:
It works now.

plugins/GSdx/GSHwHack.cpp
```diff
@@ -762,10 +763,10 @@ bool GSC_Kunoichi(const GSFrameInfo& fi, int& skip)
 		{
 			skip = 1;
 		}
-		if(fi.TME)
+		// Removes burning air effect, the effect causes major slowdowns.
```
lightningterror: Put this above the skip value.

 			{
 				// Removes burning air effect, the effect causes major slowdowns.
 				skip = 1;
 			}

plugins/GSdx/GSHwHack.cpp
```diff
@@ -752,6 +752,7 @@ bool GSC_TalesOfLegendia(const GSFrameInfo& fi, int& skip)
 
 bool GSC_Kunoichi(const GSFrameInfo& fi, int& skip)
 {
+	// Removes depth effects(shadows) not rendered correctly on all renders.
```
lightningterror: This one as well in it's own skip value.

ivan89el: I can not understand you. Unfortunately I do not know English well.

lightningterror: Move the comment where skip line is, above it.
 		{
 			// Removes depth effects(shadows) not rendered correctly on all renders.
 			skip = 1;
 		}

ivan89el:
Look like that's it.

plugins/GSdx/GSHwHack.cpp
```diff
@@ -756,19 +756,22 @@ bool GSC_Kunoichi(const GSFrameInfo& fi, int& skip)
 	{
 		if(!fi.TME && (fi.FBP == 0x0 || fi.FBP == 0x00700 || fi.FBP == 0x00800) && fi.FPSM == PSM_PSMCT32 && fi.FBMSK == 0x00FFFFFF)
 		{
+			// Removes depth effects(shadows) not rendered correctly on all renders.
 			skip = 3;
 		}
 		if(fi.TME && (fi.FBP ==0x0700 || fi.FBP==0) && fi.TBP0==0x0e00 && fi.TPSM ==0  && fi.FBMSK == 0)
 		{
+			// Removes depth effects(shadows) not rendered correctly on all renders.
```
lightningterror: This comment isn't needed here, it should be only on skip = 3.

lightningterror:
Can you also squash the commits in to 1 with git rebase ?

ivan89el:
How to do it? (Can you also squash the commits in to 1 with git rebase ?)

lightningterror:
Make sure you're on the correct branch and open Git shell.
Type git rebase -i master or git rebase -i HEAD~n (n being the number of commits.)
You follow the instructions from the text editor that pops up -> save 
Then you type git push -f to force push the changes.

ivan89el:
Unfortunately for me it's too complicated.

PR #2512, opened 2018-07-17
GSdx: Add missing CRC id.

Al0ner:
GSdx: Add missing CRC id.
Add TombRaiderUnderworld RU.

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -371,6 +371,7 @@ CRC::Game CRC::m_games[] =
 	{0x694A998E, TombRaiderUnderworld, JP, 0}, // cutie comment
 	{0x8E214549, TombRaiderUnderworld, EU, 0},
 	{0x618769D6, TombRaiderUnderworld, US, 0},
+	{0x8E265148, TombRaiderUnderworld, RU, 0}, // Unofficial RU-version
 	{0xB639EB17, TombRaiderAnniversary, US, 0}, // Also needed for automatic mipmapping
 	{0xB05805B6, TombRaiderAnniversary, JP, 0}, // cutie comment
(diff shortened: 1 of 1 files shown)


plugins/GSdx/GSCrc.cpp
```diff
@@ -136,8 +136,7 @@ CRC::Game CRC::m_games[] =
 	{0xB01A4C95, ICO, JP, 0},
 	{0x2DF2C1EA, ICO, KO, 0},
 	{0x5C991F4E, ICO, EU, 0},
-	{0x7ACF7E03, ICO, NoRegion, 0}, // same CRC as {SpyroNewBeginning, NoRegion}
```
lightningterror: Why did you remove this ?

Al0ner: It does'nt make any sense for this game.
There is no one original PS2 game with this ID.
But almost all unofficial RU-versions of games with a multiloader - have exactly this CRC-id.
7ACF7E03 - is the ID of the multiloader, not the executable file of the game.
Therefore, if you repacked such an image, delete the multiloader and leave only the executable file of the game - the CRC-id will change to the original one, or at least to the unique one.
You know that if the same CRC-hacks and WD-pnaches be applied to different games - it can break them. Therefore, there is no sense in this ID.

lightningterror: Ah yes you're right, I did a bit of research and see multiple games with the same crc.

In that case you could also remove line 414 and 415.
These to be precise.
```
	// {0x7ACF7E03, SpyroNewBeginning, NoRegion, 0}, // same CRC as {ICO, NoRegion}
	// and as "Twisted Metal - Black" (PAL).
```

lightningterror:
Next time please squash your commits properly with Git Shell.

PR #2513, opened 2018-07-17
GSdx: Add Harry Potter Order of the Phoenix to automatic mipmapping.

MrCK1:
This PR Adds Harry Potter Order of the Phoenix to the CRC list for games that require mipmapping.

### DIFF
--- a/plugins/GSdx/GSCrc.h
+++ b/plugins/GSdx/GSCrc.h
@@ -79,6 +79,7 @@ public:
 		GTASanAndreas,
 		GTConcept,
 		HarleyDavidson,
+		HarryPotterOOTP,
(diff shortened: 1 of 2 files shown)


plugins/GSdx/GSCrc.h
```diff
@@ -78,6 +78,7 @@ class CRC
 		GT4,
 		GTASanAndreas,
 		GTConcept,
+		HarryPotterOOTP,
```
turtleli: `r` comes before `l` in the alphabet? You learn something new everyday.

lightningterror:
I just noticed, you didn't actually add it to Automatic mipmapping, you just added the crc ids.

PR #2527, opened 2018-07-29
Debugger: Display ACC register

Souzooka:
Displays/allows ACC VU0f register to be set.

### DIFF
--- a/pcsx2/DebugTools/DebugInterface.cpp
+++ b/pcsx2/DebugTools/DebugInterface.cpp
@@ -348,9 +348,10 @@ int R5900DebugInterface::getRegisterCount(int cat)
 	case EECAT_CP0:
 	case EECAT_FPR:
 	case EECAT_FCR:
-	case EECAT_VU0F:
 	case EECAT_VU0I:
 		return 32;
+	case EECAT_VU0F:
+		return 33;  // 32 + ACC
 	default:
 		return 0;
 	}
@@ -395,7 +396,13 @@ const char* R5900DebugInterface::getRegisterName(int cat, int num)
 	case EECAT_FCR:
 		return R5900::COP1_REG_FCR[num];
 	case EECAT_VU0F:
-		return R5900::COP2_REG_FP[num];
+		switch (num)
+		{
+		case 32:	// ACC
+			return "ACC";
+		default:
+			return R5900::COP2_REG_FP[num];
+		}
 	case EECAT_VU0I:
 		return R5900::COP2_REG_CTL[num];
 	default:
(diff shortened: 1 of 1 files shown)


pcsx2/DebugTools/DebugInterface.cpp
```diff
@@ -348,9 +348,10 @@ int R5900DebugInterface::getRegisterCount(int cat)
 	case EECAT_CP0:
 	case EECAT_FPR:
 	case EECAT_FCR:
-	case EECAT_VU0F:
 	case EECAT_VU0I:
 		return 32;
+	case EECAT_VU0F:
+		return 33;  // 32 + ACC
```
lightningterror: I'd prefer if EECAT_VU0F was not moved. To keep things consistent with the other switch cases below.

Souzooka: If it wasn't moved, it'd be changed to something like
```
case EECAT_CP0:
case EECAT_FPR:
case EECAT_FCR:
  return 32;
case EECAT_VU0F:
  return 33;
case EECAT_VU0I:
  return 32;
```
as the top cases rely on fallthrough. Restoring the order is OK if that's fine though.

lightningterror: I'm fine with the change if it works properly.

ramapcsx2:
Code looks fine.
If that register is in the correct location, this is good :)

pcsx2/DebugTools/DebugInterface.cpp
```diff
@@ -348,7 +348,9 @@ int R5900DebugInterface::getRegisterCount(int cat)
 	case EECAT_CP0:
 	case EECAT_FPR:
 	case EECAT_FCR:
+		return 32;
 	case EECAT_VU0F:
+		return 33;  // 32 + ACC
```
turtleli: IMO the way you originally had it was better.

Souzooka:
Any thoughts on the previous commit? It'd be nice to see ACC in master because a lot of VU ops use it.
Unless it behaves differently and needs to be displayed differently because of that, I couldn't really find documentation on the PS2 ACC register.

pcsx2/DebugTools/DebugInterface.cpp
lightningterror: Will it cause any issues if it's like this ? I still prefer this way since it's consistent but if functionality wise the other was better then the second commit can be dropped.

turtleli: The other functions return different values for each case, so in that case I prefer enum ordering.

This function however returns the same value for multiple cases. In this case I prefer having the cases that return the same values grouped together. I can determine EECAT_GPR has 35 rows, EECAT_VU0F has 33 rows and everything else has 32 rows that little bit faster. `getRegisterSize()` and `getRegisterType()` also do the same thing.

lightningterror:
@Souzooka can you drop the second(newer) commit ?
The original change was better.

lightningterror:
Note: you could have used git shell to drop the second commit.
With either git rebase or reset and after doing a force push.

Anyway it's fine we'll squash it upon merger.

lightningterror:
Can I merge this ?

ramapcsx2:
Still looks fine to me, so please go ahead :)

PR #2531, opened 2018-07-31
GameDB: Add patches for Harry Potter (Chamber of Secrets, Prisoner of Azkaban)

lightningterror:
Chamber of Secrets/Prisoner of Azkaban: Add gamefix patches to make the
games playable/bootable. Some regions that don't have the patches added
still need testing for verification.

Add some missing regions to the DB for the above two games.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -9190,7 +9190,23 @@ Name   = Harry Potter and the Chamber of Secrets
 Region = PAL-E
 Compat = 3
 [patches = C5473413]
-	// This disc has the same CRC as SLUS-20576, the NTSC-U disc.
+	// This disc has the same CRC as other PAL and NTSC discs.
+	comment=Patch by Prafull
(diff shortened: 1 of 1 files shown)


turtleli:
> we can assume that the patches will work on regions that
> don't have the patches added

Are you sure you can assume that? The PCSX2 wiki shows a different CRC for the NTSC-J version of Chamber of Secrets.

lightningterror:
Ye my mistake, SLPM-64528 is Korean.

turtleli:
That's not what I meant. The CRCs for the NTSC-J versions are different to the NTSC-U and PAL versions, so I do not see how it can be assumed that the same patch will work for these versions as well.

lightningterror:
Ah yes you're right, anyway I'll just add tested-ish games.
SLES-51217, SLES-51218 haven't been tested but since the other pal regions use the same crc id and are tested to work those could work the same for the game.

PR #2535, opened 2018-08-02
Reduce clang compilation warnings

turtleli:
Fixes some clang compilation warnings.


willkuer:
First time I heard about trigraphs. Somehow amazing how long such relatively bad (in comparison to unicode or html special chars) design decisions last.

PR #2536, opened 2018-08-02
GSdx: GTK changes

turtleli:
Changes:
 * Switches to non-deprecated GTK3 API - The problems I encountered in #1914 (stuff not expanding) seems to have resolved itself on my end. Reduces the number of deprecation warnings.
 *  I've adjusted the layout slightly so stuff doesn't expand vertically if there's space. IMO it looks better, but GUI changes are rather subjective things...

Before second commit:
mmm

After second commit:
lll

### DIFF
--- a/plugins/GSdx/GSLinuxDialog.cpp
+++ b/plugins/GSdx/GSLinuxDialog.cpp
@@ -29,7 +29,7 @@
 // Port of deprecated GTK2 API to recent GTK3. Those defines
 // could prove handy for testing
 #define GTK3_MONITOR_API GTK_CHECK_VERSION(3, 22, 0)
-#define GTK3_GRID_API GTK_CHECK_VERSION(3, 10, 0)
+#define GTK3_GRID_API GTK_CHECK_VERSION(3, 12, 0)
 
 static GtkWidget* s_hack_frame;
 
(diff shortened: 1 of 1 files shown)


MrCK1:
It looks like "Hardware Mode Settings" and "Software Mode Settings" are overlapping the border a bit, maybe it should be moved over a bit?

turtleli:
It's not overlapping, though it might look like it is. It probably can be adjusted, but I'm not really keen on doing so.

hellbringer616:
This would fix an issue i am having with my brother and his small screen size (768 vertical resolution)

turtleli:
Hmm. Screen size detection doesn't work with the new code on my VM (logo shows on small resolutions, it doesn't show on master branch). Will need to investigate.

turtleli:
Updated:
 * Improve dialog on small screen resolutions - if the screen is small enough the OK button will not be visible on the screen. This fixes that issue (on GTK3 at least, GTK2 is untested because I don't have a VM for it set up right now) by making all the tabs scrollable.
 * (GTK3 only) Fix fetching of monitor resolution - there may not be a primary monitor configured, so take the resolution of the monitor that the mouse cursor is currently on instead.

PR #2537, opened 2018-08-02
Generate cheats_ws.zip from a submodule

turtleli:
This is one approach to generating cheats_ws.zip instead of having to update the zip file in the repo every few months. Whether this is a good idea or not is another question...

Changes:
 * Adds a widescreen patch submodule - Currently a repo in my account is used. A repo under the PCSX2 account will need to be used if this PR goes ahead. I just unzipped the current cheats_ws.zip (and lowercased 3 file extensions), I didn't bother updating it to whatever is most recent.
 * Increases minimum required CMake version to 3.2 - This is required to use the CMake tar command.
 * Adds a PowerShell 5+ script to create the zip file on Windows.
 * Removes the cheats_ws.zip from the repo.

Notes:
 * Powershell 5+ is not installed by default on Windows 7 and 8.1, so this introduces an extra build dependency on those systems. This might be a minor issue? I also don't know if the Orphis buildbot has it installed.
 * Increasing the CMake minimum version requirement to 3.2 shouldn't be an issue. Most Linux distros use a later version. The policies which currently use old behaviour need to be checked though (not really an issue, it'd have to be done sooner or later even without this PR).
 * I normally do not use widescreen patches, so I doubt I'll be active in maintaining the widescreen repo (IOW, someone else should be willing to update the repo if this goes ahead).


mirh:
Mhh.. I wonder if we couldn't replace the script by exploiting somehow github "Download ZIP" feature?
(separately, tbh I have never been a fan of powershell)

Anyway, for as much as I don't mind either way this PR, something I'd really see well in a detached repository is GameIndex.dbf. 
Especially if we had to also [include ps1 titles](https://github.com/PCSX2/pcsx2/issues/2246#issuecomment-355855846), and hacks for [more than just](https://github.com/PCSX2/pcsx2/issues/1894) core.

avih:
> This is one approach to generating cheats_ws.zip instead of having to update the zip file in the repo

Good goal indeed.


> Adds a widescreen patch submodule

What's the reason for this, instead of just putting the files e.g. at `3rdparty/cheats_ws` ?


> Increases minimum required CMake version to 3.2 - This is required to use the CMake tar command

Wouldn't a requirement to have `zip` installed be a much lower bar?


> Adds a PowerShell 5+ script to create the zip file on Windows

I _think_ that including some tiny `zip.exe` from a third party wouldn't be a bad solution either.


@mirh `GameIndex.dbf` is off topic here, lease file a different issue if you think it shouldn't be part of the main repo (fwiw, i think it should stay).

lightningterror:
> Adds a PowerShell 5+ script to create the zip file on Windows

IIRC Git Shell requires PowerShell 5+ so it's fine to add that requirement. People can download the update from MS website. I think it was around 60MB.

avih:
> I think it was around 60MB.

And a `zip.exe` is likely to be less than 100k.

mirh:
> Wouldn't a requirement to have zip installed be a much lower bar?

AFAIU tar is an actual command [inside cmake](https://cmake.org/cmake/help/latest/manual/cmake.1.html#command-line-tool-mode). 

> I think not. I have git for windows installed and working, and my powershell version is 4.

Yep.

turtleli:
> I wonder if we couldn't replace the script by exploiting somehow github "Download ZIP" feature?

I somewhat prefer not having internet connection stability determine whether a build succeeds or not.

> What's the reason for this, instead of just putting the files e.g. at 3rdparty/cheats_ws ?

There's no real reason. I more or less just thought "lots of files, put them in a submodule". I suppose whether it's a good idea or not will really depend on if it's helpful to whoever will be updating the widescreen patches.

> Wouldn't a requirement to have zip installed be a much lower bar?

It depends on how up to date the Linux distro is. Most distros use more recent versions of CMake (see https://repology.org/metapackage/cmake/badges, it covers a fair number of distros). I actually haven't installed zip on my Linux distro.

> I think that including some tiny zip.exe from a third party wouldn't be a bad solution either.

Yeah, that'd also work. It'd add a binary blob, but it's a one-off.

I tried the PowerShell approach because I wanted to attempt a no new binary blob approach first. Obviously the disadvantage is the additional requirement on Windows 7 and 8.1. It might be possible to lower the requirement to Powershell 3+ by changing the script, though Windows 7 devs would still need to update (default Powershell version seems to be 2.0).

mirh:
> I somewhat prefer not having internet connection stability determine whether a build succeeds or not.

I see.. Good point (for as much it's not like ws_hacks are that much important for anything.. "anybody that build its own pcsx2" does)
Anyway, what about the *[!command](https://git-scm.com/docs/git-submodule#git-submodule-customcommand)* variable inside submodule.xxx.update file?

> I suppose whether it's a good idea or not will really depend on if it's helpful to whoever will be updating the widescreen patches.

Well, Devina really don't mind.. As long as he can keep doing whatever he has always been doing. 

If any though, I'm not liking the idea of more than a thousand file in a single folder. 
Would there be any way to further divide/categorize them?

> Yeah, that'd also work. It'd add a binary blob, but it's a one-off.

Fun fact, it seems like `tar`, `gzip` and `bzip2` is a supported command inside git-scm bash cli. 
They may not be actually enough to make a standard zip file on their own.. But to all uses and purposes, for somebody using 7-zip to open them, nothing should change afaik.

orbea:
Please don't use submodules, they are not fun with shallow clones using `--depth=1`. You could alternatively use git sub-trees to avoid this issue.

mirh:
Submodules should [perfectly work](https://stackoverflow.com/questions/2144406/how-to-make-shallow-git-submodules) with shallow. 
Subtrees.. I don't understand what would basically be different than what we already do?

If any I wonder (whether we'll end up doing) if it wouldn't make sense to point to the head of the separate repository, rather than some specific commit.

orbea:
The first problem is depth, particularly how deep referenced submodule commit is. If you can assure that the submodule will always pull the latest commit then I don't think its much of a problem, but if the latest commit is not the referenced commit then there is no really good way to know how deep to make the clone which will fail if its not deep enough...

The second problem is that with my connection when cloning from github I get around 2 mb/s, but sometimes the connection is throttled to 10 kb/s instead and even if the main clone works okay each submodule may or may not be throttled. Additionally the only way I found to know if this is the case is with `git clone --recursive`.

Using git subtrees entirely avoids both problems and prevents the extra build step of updating the submodules which could confuse less experienced users. Though honestly, I've been traumatized by ffmpeg or llvm submodules in other projects...

avih:
I still don't see how a submodule helps here. The best we got so far is:

> There's no real reason. I more or less just thought "lots of files, put them in a submodule". I suppose whether it's a good idea or not will really depend on if it's helpful to whoever will be updating the widescreen patches.

However, whoever is updating them has to deal with the thousands of them regardless of where they live. If anything, the question is whether or not it would bother those who _don't_ handle them but now have them at the repo.

I don't _think_ it should bother anyone as part of the main repo.

turtleli:
It can be changed. The decision to put all the patches in a submodule was rather arbitrary and was done quickly so I could see if I could actually get the rest of the PR to work.

I think there might be another issue with the PR. I don't think it'll be nice for reproducible builds if a zip file is generated. Zip files store modification times, so zip files generated on two PCs with the same build environment won't be identical unless the files were checked out at the exact same time.

mirh:
Very very smart remark. 
So.. Either we drop dates from the archive, or we make them the same. 
I just checked though, and.. I think some additional command-line-fu would be required for the former (\*if any could even\*, at least with regard to windows)

As for the later, it seems pretty doable but `zip` utility doesn't seem to support specifying arbitrary dates.
*ON THE OTHER HAND* I just figured out `git archive` can produce zips. And all timestamps (even when using HEAD tag) are set to the one of the specified commit. 
So.. using it would be a double win I guess?


..But for as much as I agree not having zips into sources feels *neater*
I still loathe the idea of "*thousands of files in a single common folder*"

i30817:
This isn't any of my business since i'm not working on it, but if i was making pcsx2 patches, i'd vastly prefer 'thousands of files on a subdirectory' simply because PRs could be done by editing directly on github. The forum *does* have the advantage of a dedicated population of testers saying yea or nay, but if most of the hardworking ones can get into github the hacks would be better documented.

mirh:
The two things aren't exclusive or at fault. 
It takes seconds to unpack the files and push them. 

And my criticism to the "thousands of files" together, didn't seem to be a flat out no either. 
In fact.. To be honest it could be already made of the archive we have (though, not existing any "real time editing" need for the files, I guess like it never came to the mind of anyone). 

***If**\* git archive is deemed an acceptable archiver, and  \***if**\* the submodule update custom command works.. I think we are pretty much done here.

GovanifY:
I'm fine with the idea of a submodule, which I find much better than putting it in the 3rdparty folder (which also it isn't, fwiw).

You can easily get reproducible zip by changing the timestamp, here is some random gist that does just that: https://gist.github.com/bwindsor/eab6477135e9b372ea508d0da299dc2b
Although i have to say I'm now curious how reproducible our github builds are...

Regardless, that PR looks pretty great to me tbqh. Would you mind updating it to be reproducible/rebase @turtleli ?
I could always do it later if not enough interest is shown but that looks like a great step to get a more maintainable codebase to me.

lightningterror:
Closing as this isn't currently being worked on.
Last time WS patches were updated was 2 years ago so it's not really that of a big deal, it's usually done before a release or yearly I guess if we want it more up to date?

mirh:
Follows in #6777

PR #2541, opened 2018-08-05
GSdx: Direct3D Updates

lightningterror:
GSdx-d3d: Drop an useless min in FS - port from GL.
Commit: PCSX2@a1957a6
GPU will clamp color anyway. It reduces the number of instruction of 25% (4->3).

GSdx-d3d: Reduce state change - port from GL.
Commit: PCSX2@37f9bcf
Don't dirty aref when a fog color is uploaded.
Only set clamp mode in clamp mode (region clamp is handled in shader).

GSdx-d3d: s/sample_4a/sample_4_index/


pgert:
It reduces the number of instructions by 25% ...

"GSdx: Direct3D Updates" -> "GSdx: Direct3D Improvements" ?

gregory38:
It is fine for me :+1:

PR #2545, opened 2018-08-07
GSdx-gui: Update gui interface

lightningterror:
Move SW render options to the same location as Hw render. HW Render
options will be hidden from SW render(when SW is active) and vice versa. Core Benchmark is
blank

The height of the window has been reduced, nice for low resolution
screens.

The gui looks much more cleaner and less confusing for newcomers.

Screenshot goodies.

HW and SW
image


plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -389,6 +389,7 @@ void GSSettingsDlg::UpdateControls()
 		bool ocl = renderer == GSRendererType::DX9_OpenCL || renderer == GSRendererType::DX1011_OpenCL || renderer == GSRendererType::OGL_OpenCL;
 		bool null = renderer == GSRendererType::Null;
 
+		// GSdx logos
```
turtleli: Isn't this kinda obvious from looking at the IDC identifiers?

Same applies to quite a few of the other comments.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -398,57 +399,78 @@ void GSSettingsDlg::UpdateControls()
...
+		ShowWindow(GetDlgItem(m_hWnd, IDC_MIPMAP_HW), hw ? SW_SHOW : SW_HIDE);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_MIPMAP_HW_TEXT), hw ? SW_SHOW : SW_HIDE);
+		//CRC hack
+		ShowWindow(GetDlgItem(m_hWnd, IDC_CRC_LEVEL), hw ? SW_SHOW : SW_HIDE);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_CRC_LEVEL_TEXT), hw ? SW_SHOW : SW_HIDE);
+		// Large framebuffer
+		EnableWindow(GetDlgItem(m_hWnd, IDC_LARGE_FB), hw && integer_scaling != 1);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_LARGE_FB), hw ? SW_SHOW : SW_HIDE);
+		// OpenCL
 		EnableWindow(GetDlgItem(m_hWnd, IDC_OPENCL_DEVICE), ocl);
```
turtleli: What is OpenCL doing in the middle of hardware settings?

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -398,57 +399,78 @@ void GSSettingsDlg::UpdateControls()
...
+		ShowWindow(GetDlgItem(m_hWnd, IDC_LOGZ), dx9 && hw ? SW_SHOW : SW_HIDE);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_FBA), dx9 && hw ? SW_SHOW : SW_HIDE);
+		// Accurate Blending
+		ShowWindow(GetDlgItem(m_hWnd, IDC_ACCURATE_DATE), ogl && hw ? SW_SHOW : SW_HIDE);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_ACCURATE_BLEND_UNIT), ogl && hw ? SW_SHOW : SW_HIDE);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_ACCURATE_BLEND_UNIT_TEXT), ogl && hw ? SW_SHOW : SW_HIDE);
+		// HW mipmap
+		ShowWindow(GetDlgItem(m_hWnd, IDC_MIPMAP_HW), hw ? SW_SHOW : SW_HIDE);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_MIPMAP_HW_TEXT), hw ? SW_SHOW : SW_HIDE);
+		//CRC hack
```
turtleli: Missing space after `//`?

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -398,57 +399,78 @@ void GSSettingsDlg::UpdateControls()
...
+
+		// Software mode settings
+		//
+		ShowWindow(GetDlgItem(m_hWnd, IDC_SW_RENDER_TEXT), sw ? SW_SHOW : SW_HIDE);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_MIPMAP_SW), sw ? SW_SHOW : SW_HIDE);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_AA1), sw ? SW_SHOW : SW_HIDE);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_SWTHREADS_TEXT), sw ? SW_SHOW : SW_HIDE);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_SWTHREADS_EDIT), sw ? SW_SHOW : SW_HIDE);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_SWTHREADS), sw ? SW_SHOW : SW_HIDE);
+		//
```
turtleli: I don't really like how `//` has been used as section dividers. If I see an empty comment it makes me think "did they forget to type the comment".

MrCK1:
I'd like to shrink the dialog in SW mode, maybe I can look at it. That huge space is horrible imo

turtleli:
Personally I prefer current master because I can see both my hardware and software settings at the same time. I don't think the GUI size is really an issue on WIndows for users with small screens (I could be wrong).

lightningterror:
Well it's only a minor inconvenience. SW render has only a few options that are rarely toggled and most people will be using HW render so it should be fine.

turtleli:
But is it really confusing to new users to have both the hardware and software mode options present at the same time? There are only three options for the software renderer, and both the hardware and software mode settings are within groupboxes that indicate when the settings will be used.

lightningterror:
@turtleli are you fine with the PR so I can merge it ?

turtleli:
I still prefer the old dialog.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -398,59 +398,70 @@ void GSSettingsDlg::UpdateControls()
...
-		EnableWindow(GetDlgItem(m_hWnd, IDC_CRC_LEVEL_TEXT), hw);
 		EnableWindow(GetDlgItem(m_hWnd, IDC_OPENCL_DEVICE), ocl);
+		// Hardware Renderer Settings:
+		ShowWindow(GetDlgItem(m_hWnd, IDC_HW_RENDER_TEXT), hw ? SW_SHOW : SW_HIDE);
+		// Allow 8-bit textures
+		ShowWindow(GetDlgItem(m_hWnd, IDC_PALTEX), hw ? SW_SHOW : SW_HIDE);
+		// Large framebuffer
+		EnableWindow(GetDlgItem(m_hWnd, IDC_LARGE_FB), hw && integer_scaling != 1);
+		ShowWindow(GetDlgItem(m_hWnd, IDC_LARGE_FB), hw ? SW_SHOW : SW_HIDE);
+		// Upscale multiplier
```
turtleli: I'm not sure the comments have much value. I think it's obvious in most cases. Dunno.

lightningterror:
Updated.

RedPanda4552:
I do sit in favor of keeping the SW/HW renderer settings separate; it makes more sense to me to display hardware settings when a hardware renderer is up and software settings when a software renderer is up, without mudding up the two.

And, I'm going to be that guy, but lets use Dolphin as an example here.
[Software Renderer](https://i.imgur.com/a2Ysdo4.png)
[Vulkan Renderer](https://i.imgur.com/NYrgKX2.png)

Personally, I think this is a lot more intuitive, and it has my vote. I do wonder if something could be done to better distinguish the software renderers, perhaps not having three of them and just calling it "Software Renderer", rather than naming it after the backend used and making "Software" an afterthought parenthetical.

lightningterror:
You've got to find balance between functionality, appeal, looks ...etc
We don't need to bombard people with a bunch of options that shouldn't actually be there, and a lot of people actually criticized us for having too many options.
Software settings for software render, hardware settings for hardware render.
cr hacks and mipmapping is automatic so most of the time people shouldn't change those settings, Large FB usually left on, Blending basic is ok most of the time as well.

willkuer:
I don’t think my opinion matters much but I do agree with turtleli here. I don’t think that keeping these few sw settings disturbs that much for the newbie. If you would decrease the size by half it would be different.

And since hotswapping via f9 is an option I’d also say both should be adjustable at the same time.

You guys once had the idea to copy the tabcontrol layout from the linux port. What is wrong about that? Then you could have tabs with general, hw, sw, hw advanced...

Or what about a simplified gui where you can just select renderer, internal resolution for hw and number of threads for sw which are the most important settings for newbies and an advanced button where you can see the current gui?

turtleli:
> You've got to find balance between functionality, appeal, looks ...etc

Then we disagree on what the correct balance is. I prefer the little bit extra convenience of master (and prefer even more having both adjustable at the same time, if I want to change a software mode setting but still want to start in hardware mode then both master and this PR are not the most convenient), whereas you prefer the little bit extra newbie friendliness that this PR has.

I suggest waiting a bit so people can provide their opinions if they wish to do so. We can see what the overall opinion is and go from there.

> You guys once had the idea to copy the tabcontrol layout from the linux port. What is wrong about that? Then you could have tabs with general, hw, sw, hw advanced...

I'd prefer that too.

avih:
> a lot of people actually criticized us for having too many options

Citation needed. Also, There will always be comments/criticism regardless of what you choose. The real question is how much objection there would be if we change it (I'd guess a lot more).

> but I do agree with turtleli here. I don’t think that keeping these few sw settings disturbs that much for the newbie.
> 
> And since hotswapping via f9 is an option I’d also say both should be adjustable at the same time.

Agreed here.

Bottom line, IMO they should be exposed together due to the switching, and SW adds very little clutter IMO.

Also consider that with two dialogs some options become duplicate in both dialogs, which could be a source of confusion in itself.

lightningterror:
> You guys once had the idea to copy the tabcontrol layout from the linux port. What is wrong about that? Then you could have tabs with general, hw, sw, hw advanced...

Yes the tabcontrol is pretty slick on linux, especially to get rid of the extra windows for osd/shader/hw hacks is great.

ssakash:
> The gui looks much more cleaner and less confusing for newcomers.

Much cleaner maybe. but I'm not sure about the confusion part. We do gray out the hardware renderer exclusive options when software renderer is selected, how does it cause confusion?

lightningterror:
On a side note what do you guys think about swapping Accurate Date and Blending Unit Accuracy positions ?

willkuer:
You could even move it two levels up.

lightningterror:
Nah then it will look weird on D3D, unless you duplicate the options and position them differently for each render.

willkuer:
Ah yeah you‘re right. Forgot that this is ogl only.

lightningterror:
Checkbox -> Drop down list -> checkbox .. kinda looks weird as well. What do you think ?

image

Ryudo300:
As I mentioned in Discord. Not a fan of any of these changes. 

It looks fine to me as it is right now. All the other suggested changes make it look messier imo.

If anything I'd make use of a wider width and add a column there with some options instead of trying to push everything together. But even that seems a bit unnecessary at this point.

lightningterror:
Alright I'll be closing this.

PR #2550, opened 2018-08-12
pcsx2:gui: Remove Game Database dialog and panel

turtleli:
I'd like to remove the Game Database editor (present in Debug builds).

It has quite a few issues:
 - It's in an incomplete state. The ListView isn't connected up and I'm not sure what it was supposed to do.
 - Comments are stripped from the database when changes are saved.
 - Some key value pairs do not show (e.g. EE/VU clamp/rounding and MemCardFilter).
 - It doesn't work well on Linux (though this one seems easy to fix).


lightningterror:
Is this a good idea ? We're stripping features which may provide useful.
I never used the features so I don't know much about how it works and it's usability so I'm not too sure.

turtleli:
IMO a text editor provides better functionality than what's currently present:
 * Can insert new GameDB entries into the correct place rather than at the end
 * Isn't restricted to searching by serial
 * Can add/edit/remove EE/VU clamp/round modes, MemCardFilter and whatever other settings
 * Doesn't strip comments that are not present in the header
 * Doesn't require any of us to maintain or complete it

It can't sort entries. The tools/GameIndex.dbf-tool html file can take care of that though and will also update the compatibility status at the same time (and it also fixes a few trailing whitespace issues as well, some text editors can do this too).

The only thing that hasn't been covered by a text editor or the html file is the lack of spaces on either side of the `=` added in 011cd90fbe834114955efb1e9ae72e2bf32bca7c.

So I find it rather unnecessary.

ssakash:
Well it's only exclusive to debug builds for a reason, as it's not complete yet. 

Personally I don't see any merit in removing this at the moment, while there's a chance for anyone interested to potentially update it.

turtleli:
It's not been complete for many years now though and nobody is really interested in fixing it. It's only a maintenance burden (and again, I see no advantages compared to a text editor).

mirh:
Well, just about nobody knowing it existed, certainly explains a lot of it. 
But even then I really couldn't see what would be its "selling point" compared to a normal editor. 

Why was it added in the first place?

lightningterror:
> Why was it added in the first place?

I have no idea, maybe it was useful for very old versions.

Anyway valid points for removal.

ssakash:
> and again, I see no advantages compared to a text editor

I guess we should also remove the GSdx plugin and core GUI, it offers no advantage over the conventional INI files.

The point of a GUI has always been about user friendliness, and I'm not sure if your comparison to a text editor and an external GameDB modification tool is reasonable in this aspect.

turtleli:
> I guess we should also remove the GSdx plugin and core GUI, it offers no advantage over the conventional INI files.

It offers a way for users to change settings at runtime, which you might need to do fairly often. Users do not need to frequently adjust the GameDB.

> The point of a GUI has always been about user friendliness, and I'm not sure if your comparison to a text editor and an external GameDB modification tool is reasonable in this aspect.

The GameDB editor is not user friendly in its current state. Also, users generally do not need to change the GameDB. Only devs and contributors really need to. The potential user base is small. Its not worth the effort to maintain the editor, especially in its current state.

mirh:
> It offers a way for users to change settings at runtime, which you might need to do fairly often. Users do not need to frequently adjust the GameDB.

More simply, even "normal noob users" touch gsdx settings all day and night. 
GameDB is already a mess for developers (and I'm not even sure if all of its hacks work happily if applied after boot)

> I have no idea, maybe it was useful for very old versions.

> [If you have a game loaded the editor will automatically load the game's info](https://github.com/PCSX2/pcsx2/commit/149d3290dbdf5ae9a81a833bea010cf2522aecfa)

I mean... I guess seeing "what hacks are getting applied to the game" in-app is not a bad feature.

lightningterror:
> I mean... I guess seeing "what hacks are getting applied to the game" in a nice in-app window is not a bad feature.

For some reason it doesn't show any of that, I guess it's broken as well ?

Anyway how about something like this? First example is just a viewer with the option to open the GameDB file like you would normally would in windows.
game database viewer


Second option with the ability to view and edit but a bit more different I guess ? Where the info is displayed you could also edit it.
game database editor

bositman:
What do we gain from the removal? I mean what is the maintenance cost turtleli refers to?

mirh:
I mean, if you check its history.. It's not like he hasn't to fix it (for as much as slightly) every couple of years or so. 
If it's total long forgotten useless code (for as much as we aren't dolphin /s /s)... It makes sense to drop it. 

On the other hand I'm now getting scruples over it. Ie. if it couldn't be recycled someway else.

turtleli:
Another thing to note -  If it's an installed version of PCSX2, then the GameDB file will be installed into a read-only location. The user will not be able to edit the file unless they run PCSX2 with elevated privileges.

> What do we gain from the removal? I mean what is the maintenance cost turtleli refers to?

It'd be less code to maintain. The dialog is rather incomplete and functions poorly as an editor (it removes all comments that are not in the header, so right now, it destroys data). I'd rather remove it than try to fix it up since I don't see how it's useful in its current state.

The maintenance cost is dev time. wxWidgets work eats time - in my case I have to either dual boot or run a VM a lot of the time to check that both Windows and Linux is fine. Hardly anyone else will do that. For example, the most recent About box dialog change looks fine on Windows, but on Linux it now look likes:

aaaa

The line breaks weren't a great idea.

Basically there's plenty of wxWidgets issues that need solving (dialogs being too big for smaller screens on Linux, Arabic working really badly on Windows, FreeBSD not compiling due to wxString issues, etc.). Removing unnecessary stuff does actually make some of these issues slightly easier to solve. Some of the incomplete, half done stuff actually shares code with stuff that actually is in use, so if someone needs to dig deep to fix things, it's easier if they don't actually have to fix some random unused or incomplete stuff along with it.

PR #2556, opened 2018-08-17
Silence -Wclass-memaccess warnings.

orbea:
gcc-8.2.0 prints this warning many times over making it hard to read the compile output.
```
[ 99%] Building CXX object pcsx2/CMakeFiles/PCSX2.dir/x86/newVif_UnpackSSE.cpp.o
In file included from /tmp/SBo/pcsx2/pcsx2/./PrecompiledHeader.h:80,
                 from /tmp/SBo/pcsx2/pcsx2/x86/ix86-32/iR5900Shift.cpp:17:
/tmp/SBo/pcsx2/common/include/Utilities/MemcpyFast.h: In instantiation of ‘void memzero(T&) [with T = LegacyPluginAPI_Common]’:
/tmp/SBo/pcsx2/pcsx2/./Plugins.h:212:18:   required from here
/tmp/SBo/pcsx2/common/include/Utilities/MemcpyFast.h:32:11: warning: ‘void* memset(void*, int, size_t)’ clearing an object of non-trivial type ‘struct LegacyPluginAPI_Common’; use assignment or value-initialization instead [-Wclass-memaccess]
     memset(&object, 0, sizeof(T));
     ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
In file included from /tmp/SBo/pcsx2/pcsx2/./CDVD/CDVDaccess.h:18,
                 from /tmp/SBo/pcsx2/pcsx2/./System.h:22,
                 from /tmp/SBo/pcsx2/pcsx2/./Common.h:25,
                 from /tmp/SBo/pcsx2/pcsx2/x86/ix86-32/iR5900Shift.cpp:19:
/tmp/SBo/pcsx2/pcsx2/./Plugins.h:194:8: note: ‘struct LegacyPluginAPI_Common’ declared here
 struct LegacyPluginAPI_Common
        ^~~~~~~~~~~~~~~~~~~~~~
```
I tested this locally and it seems to work, but I'm not sure this is the best way to solve this so please review to make sure its right, thanks!

For reference here is the build log.

Before - http://0x0.st/sJgp.log
After - http://0x0.st/sJgf.log

To get a good idea of how verbose these warnings are...
```
$ du -h pcsx2.*
1.1M	pcsx2.1.log
372K	pcsx2.2.log
```

### DIFF
--- a/plugins/spu2-x/src/SndOut.cpp
+++ b/plugins/spu2-x/src/SndOut.cpp
@@ -282,7 +282,7 @@ void SndBuffer::ReadSamples(T *bData)
     // If quietSamples != 0 it means we have an underrun...
     // Let's just dull out some silence, because that's usually the least
     // painful way of dealing with underruns:
-    memset(bData, 0, quietSamples * sizeof(T));
+    memset(static_cast<void *>(bData), 0, quietSamples * sizeof(T));
 }
 
 template void SndBuffer::ReadSamples(StereoOut16 *);
@@ -390,8 +390,8 @@ void SndBuffer::Init()
 
     // clear buffers!
     // Fixes loopy sounds on emu resets.
-    memset(sndTempBuffer, 0, sizeof(StereoOut32) * SndOutPacketSize);
-    memset(sndTempBuffer16, 0, sizeof(StereoOut16) * SndOutPacketSize);
+    memset(static_cast<void *>(sndTempBuffer), 0, sizeof(StereoOut32) * SndOutPacketSize);
+    memset(static_cast<void *>(sndTempBuffer16), 0, sizeof(StereoOut16) * SndOutPacketSize);
 
     sndTempProgress = 0;
 
@@ -445,7 +445,7 @@ void SndBuffer::Write(const StereoOut32 &Sample)
     //Don't play anything directly after loading a savestate, avoids static killing your speakers.
     if (ssFreeze > 0) {
         ssFreeze--;
-        memset(sndTempBuffer, 0, sizeof(StereoOut32) * SndOutPacketSize); // Play silence
+        memset(static_cast<void *>(sndTempBuffer), 0, sizeof(StereoOut32) * SndOutPacketSize); // Play silence
     }
 #ifndef __POSIX__
     if (dspPluginEnabled) {
--- a/common/include/Utilities/MemcpyFast.h
+++ b/common/include/Utilities/MemcpyFast.h
@@ -29,7 +29,7 @@ extern u8 memcmp_mmx(const void *src1, const void *src2, int cmpsize);
 template <typename T>
 static __fi void memzero(T &object)
 {
-    memset(&object, 0, sizeof(T));
+    memset(static_cast<void *>(&object), 0, sizeof(T));
 }
 
 // This method clears an object with the given 8 bit value.
--- a/pcsx2/x86/sVU_zerorec.cpp
+++ b/pcsx2/x86/sVU_zerorec.cpp
@@ -1346,7 +1346,7 @@ static VuBaseBlock* SuperVUBuildBlocks(VuBaseBlock* parent, u32 startpc, const V
 		{
 			pinst->pqcycles = QWaitTimes[pinst->info.pqinst] + 1;
 
-			memset(&w, 0, sizeof(w));
+			memset(static_cast<void *>(&w), 0, sizeof(w));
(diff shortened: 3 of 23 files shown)


orbea:
Not related, but here is a fix for those noisy gtk+2 warnings.

https://gitlab.gnome.org/GNOME/gtk/commit/889a63dffc72c048502d0f7d2b26bfc8532462eb

lightningterror:
You don't need to touch GSdx_legacy. It's basically GSdx from v1.4

turtleli:
I'm not sure it's a good idea to suppress the warnings by static casting everything it's complaining about. That's essentially saying to the compiler that you know better than it does and that there won't be any issues doing so.

It's probably better to work on the warnings on a case by case basis. The LegacyPluginAPI_Common for example could be fixed by removing the memzero and explicitly initialising all the member variables to nullptr. The memzero there was probably because someone didn't want to type a few more lines of code (I see no performance implications at least).

plugins/spu2-x/src/SndOut.cpp
```diff
@@ -390,8 +390,8 @@ void SndBuffer::Init()
 
     // clear buffers!
     // Fixes loopy sounds on emu resets.
-    memset(sndTempBuffer, 0, sizeof(StereoOut32) * SndOutPacketSize);
-    memset(sndTempBuffer16, 0, sizeof(StereoOut16) * SndOutPacketSize);
+    memset(static_cast<void *>(sndTempBuffer), 0, sizeof(StereoOut32) * SndOutPacketSize);
+    memset(static_cast<void *>(sndTempBuffer16), 0, sizeof(StereoOut16) * SndOutPacketSize);
```
turtleli: I'm not sure these lines are even needed. The constructors for StereoOut32 and StereoOut16 should already have zeroed things out.

orbea:
@turtleli Thanks for the feedback, I'll try to do this better next.

> You don't need to touch GSdx_legacy. It's basically GSdx from v1.4

I still build it in my slackbuild because some people have old cards that can't run the newer GSdx and if the warnings are going to be fixed for the nwer GSdx might as well shut them up for the legacy one too since its not much extra work.

arcum42:
This is still something that could use doing, though with nullptrs as mentioned. I switched over to compiling with gcc 8 briefly, and was amazed at how many warnings it generated.

And yes, warnings on legacy plugins should still be addressed. If you build pcsx2 with the "extra" flag, it builds all plugins, including GSdx_legacy, and I build that way often.

orbea:
Sorry for the lack of attention on this and my other PR. I have a lot of unrelated issues in real life which are not going to go away in the foreseeable future and have left me with very little energy and motivation...

lightningterror:
@orbea are you fine if I merge some of this work ? Maybe all excluding the SndOut.cpp changes or at the very least GSdx only for now ?

turtleli:
I'm not fine with it.

orbea:
If anyone wants to or can move on without me, please do.

arcum42:
@lightningterror 
The trouble is that the changes in this pr were casts to cause the compiler not to give the warnings, but don't actually fix the warning. The issue still needs to be fixed properly, by changing the initialization code.

I've put compiler flags on pcsx2 and gsdx that suppress the warnings there for the time being, as they massively spam when compiling on gcc 8.x. I've largely left this open because the issue still needs to be dealt with, but it should be dealt with by rewriting the memset, memzero, and memcpy's instead of recasting.

lightningterror:
Thanks, that was the info I was looking for.

arcum42:
No problem. I was thinking of picking off some of them gradually, but I'm not sure I'll end up getting all of them.

arcum42:
While I've left this open because it's an issue that needs to be taken care of, it occurs to me that I'm basically not actually using any of the code here to do so. 

Right now, I've muted the warnings in all but spu2-x, and I have a pr to take care of that, and I'll be gradually be addressing the other ones and removing the compiler flags that get rid of the warnings.

Any issues with my closing this pr? Unfortunately, there isn't a way for me to convert a pr into an issue...

lightningterror:
Adding `Close #2556` (or other similar commands)  on your pr will close this one, it doesn't matter if it's an issue or a pr.

orbea:
I have no issue with this PR being closed, I'll leave this open for a little while more and close it unless someone objects. Feel free to close it sooner.

PR #2559, opened 2018-08-19
Debugger: Make EE kernel region accessible to debugger

PSI-Rockin:
This PR is _not_ tested! I have not looked into how PCSX2 handles memory accesses, so reading from kernel memory (`0x80000000` and above) may cause issues.

Currently, none of the EE kernel is viewable from the debugger, which poses issues for me when debugging code that makes use of syscalls. This PR should allow the memory region `0x80000000 - 0x800FFFFF` to have readable disassembly.

### DIFF
--- a/pcsx2/DebugTools/DebugInterface.cpp
+++ b/pcsx2/DebugTools/DebugInterface.cpp
@@ -605,7 +605,7 @@ bool R5900DebugInterface::isValidAddress(u32 addr)
 	case 0xB:
 		// [ 8000_0000 - BFFF_FFFF ] kernel
 		// We only need to access the EE kernel (which is 1 MB large)
-		if (lopart <= 0x100000)
+		if (lopart < 0x100000)
 			return true;
 		break;
 	case 0xF:
--- a/pcsx2/DebugTools/DebugInterface.cpp
+++ b/pcsx2/DebugTools/DebugInterface.cpp
@@ -604,7 +604,9 @@ bool R5900DebugInterface::isValidAddress(u32 addr)
 	case 0xA:
 	case 0xB:
 		// [ 8000_0000 - BFFF_FFFF ] kernel
-		// return true;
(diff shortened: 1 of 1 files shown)


ramapcsx2:
I wonder why it was ignored before. Any idea?

gregory38:
Kernel segment (0xA, 0xB, 0x8,...) are physical access of the RAM. So it is mapped to [0 - 32MB]. But 1MB should be enough.

ramapcsx2:
If there's no reason given for excluding these regions, I would assume that it simply wasn't implemented yet. In that case, the PR will be fine. (Especially if this change succeeded in giving the expected results ;p )

lightningterror:
@ramapcsx2 did you test if it works ?

PSI-Rockin:
Actually, there's an issue I noticed. The check is done for cases 0x8, 0x9, 0xA, and 0xB. In reality, the check should only be done for 0x8, as that's the only region where the kernel resides.

I can fix this when I get home, or someone else can before then.

gregory38:
Kernel is also in 0xA region. Regions are 512 MB. 0x8/0x9 region is cached. 0xA/0xB is uncached

lightningterror:
@ramapcsx2 what's the status ?

ramapcsx2:
No, I didn't test this. I wouldn't know what to look for.
The change exposes more of the memory map to the debugger.
As such, the failure mode can be misinformation or at worst, crashes when editing memory that shouldn't be edited.
There should be no problem if the operator knows what they're doing.
If @PSI-Rockin could debug kernel code with this, it is fine.

I suppose cached/uncached could be handled in some way, maybe cloning a change made in uncached to cached (or backwards?), but this goes beyond this pr.

lightningterror:
@PSI-Rockin Status ?

lightningterror:
We did a bit a testing on discord earlier and PSI says it works fine so I'll go ahead and merge this.

PR #2562, opened 2018-08-22
Gsdx: Point Sampler changes

lightningterror:
Re add point sampler to OpenGL. Fixes graphical issues when
Allow 8-bit textures is enabled on AMD gpus.
Issue: https://forums.pcsx2.net/Thread-GSDX-Hardware-mode-Bug-Report-Allow-8-bit-Texture

Adjust the code to be easier to read, and execute the gl code only on amd - suggested by Gregory.

Remove useless ATI_SUCKS define in tfx shader.It wasn't used anywhere outside of the shader.


plugins/GSdx/res/glsl/tfx_fs.glsl
```diff
@@ -82,6 +82,15 @@ vec4 sample_c(vec2 uv)
     return texelFetch(RtSampler, ivec2(gl_FragCoord.xy), 0);
 #else
 
+#if PS_POINT_SAMPLER
+	// Weird issue with ATI cards (happens on at least HD 4xxx and 5xxx),
+	// it looks like they add 127/128 of a texel to sampling coordinates
+	// occasionally causing point sampling to erroneously round up.
+	// I'm manually adjusting coordinates to the centre of texels here,
+	// though the centre is just paranoia, the top left corner works fine.
+	uv = (trunc(uv * WH.zw) + vec2(0.5, 0.5)) / WH.zw;
```
turtleli: This file uses spaces for indentation.

lightningterror: Actually it's 50-50, some parts use spaces while others tabs. I'll go with spaces here since the code next to it also uses them.

mirh:
It would have been lovelier if ATI_SUCKS was dropped when they fixed performance /s

Issue #2558, opened 2018-08-18
[Idea] Add option to print SYSCALL_print (117) strings to console.


Souzooka:
This is something I did myself, though nothing along the lines of it being fully integrated into the UI. The concept is simple, just catch SYSCALL 117 and, if sysout messages are enabled for the console, print them to the console. This gives a helpful option for those who want to capture any sort of stubbed logging in games, for game exploration/hacking and the like. It's something that I could likely myself, I just wanted to gauge an opinion on if something like this would be wanted.

As an example:
Using eeConLog and a bit of rewriting of game code, I can print the debug stream for this game:
![img](https://i.imgur.com/p7XKxdf.png)

ramapcsx2:
Oh yeah, this is useful!
It doesn't have to be a UI option. Simply tie it to dev / debug builds.

Souzooka:
Yeah, I can take a stab at it. Probably want to add it to UI anyways since if games don't have a stubbed printf it can be very verbose and would spam logs anyways, the only other thing is that strings have to be formatted by the game itself, but I'll write more about that limitation and potential solutions when I get back to a computer.

Souzooka:
Alright, to elaborate, basically what I was going to touch on before is that this is probably a feature best used for homebrew development, as this feature should hypothetically capture the resulting strings from `(v)printf` calls and print them to the console, making it a great feature for debugging homebrew on PCSX2 (though I haven't explicitly tested that, but I don't see why it wouldn't work).

That being said though, I did want to add it as a UI element for two reasons:
* I did want to be able to use this feature in future release builds,
* it would be useful to shut up printf calls even on debug if it's not stubbed in a retail game, as it can potentially be very verbose and spammy.

And to touch on the string formatting thing, to recover printf calls fully in a retail game would require the user to rewrite printf to jump to another function, have that function set up all potential arguments correctly, then call sprintf with an allocated buffer and then print the result (and is actually what is being done in the example picture). It would be nice if PCSX2 could format the string with the provided arguments itself, but I don't believe there is a very clean way to do this and even if such a feature was implemented, it would assume that all print syscalls went through a formatting function like printf and thus might generate inaccurate results otherwise. Ultimately, I don't think something like that is in the scope of this feature itself though I might open up an issue about that later on down the line.

[...]

PR #2563, opened 2018-08-22
Add System Out as a logging source.

Souzooka:
Implements system out logging in PCSX2 console. The basic concept is explained in [this issue](https://github.com/PCSX2/pcsx2/issues/2558).

Example video:
https://gfycat.com/ThornyUnfinishedAplomadofalcon

In addition, this solution supports `printf` formatting, forgoing the need for the user to format the string in-game. By calling SYSCALL 0x75 with an unformatted string and arguments, PCSX2 will format the string.

This PR still needs a bit of polish. PCSX2 doesn't pull arguments from the stack to print them in this current implementation, meaning only up to 7 format arguments are used. If there is already a function that aggregates PS2 arguments it might be better utilized here as well. Additionally, another concern is that ConsoleLogSource always appends a newline character when writing text, and it's not necessarily something that may be preferable, though adding a method to write with ConsoleLogSource without a newline would be a lot of work and not necessarily within the scope of this PR. However, if you're pulling development/homebrew messages, these tend to already have newlines at the end of the message and thus you end up with having a lot of empty lines.

EDIT: To clarify on the newline thing, I think we can assume that if the user is a game developer and printing their own strings, or is a user that is restoring console functionality in a PS2 game that those strings already properly have a newline if necessary anyways. Thus I believe that the optimal solution probably would be a IConsoleWriter implementation that does not force newlines, and also provides formatting capabilities, though I would like to hear the opinions of others.

### DIFF
--- a/pcsx2/R5900OpcodeImpl.cpp
+++ b/pcsx2/R5900OpcodeImpl.cpp
@@ -982,6 +982,8 @@ void SYSCALL()
 			}
 			else
 				__Deci2Call(cpuRegs.GPR.n.a0.UL[0], (u32*)PSM(cpuRegs.GPR.n.a1.UL[0]));
+
+			break;
 		}
 		case Syscall::sysPrintOut:
 		{
--- a/pcsx2/R5900OpcodeImpl.cpp
+++ b/pcsx2/R5900OpcodeImpl.cpp
@@ -983,14 +983,27 @@ void SYSCALL()
 			else
 				__Deci2Call(cpuRegs.GPR.n.a0.UL[0], (u32*)PSM(cpuRegs.GPR.n.a1.UL[0]));
 		}
-		break;
 		case Syscall::sysPrintOut:
 		{
-			if (cpuRegs.GPR.n.a0.UL[0] != 0)
+			if (memRead32(cpuRegs.GPR.n.a0.UL[0]) != 0)
 			{
-				sysConLog(ShiftJIS_ConvertString((char*)PSM(cpuRegs.GPR.n.a0.UL[0])));
+				// TODO: Only supports 7 format arguments. Need to read from the stack for more.
+				// Is there a function which collects PS2 arguments?
+				sysConLog(
+					ShiftJIS_ConvertString((char*)PSM(cpuRegs.GPR.n.a0.UL[0])),
+					cpuRegs.GPR.n.a1.UL[0],
+					cpuRegs.GPR.n.a2.UL[0],
+					cpuRegs.GPR.n.a3.UL[0],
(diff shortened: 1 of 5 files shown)


pcsx2/R5900OpcodeImpl.cpp
```diff
@@ -983,7 +983,26 @@ void SYSCALL()
 			else
 				__Deci2Call(cpuRegs.GPR.n.a0.UL[0], (u32*)PSM(cpuRegs.GPR.n.a1.UL[0]));
 		}
-		break;
```
turtleli: I think `break;` shouldn't have been removed?

Souzooka: You're right, that's what I get for writing this late at night.

ramapcsx2:
So I tested a few commercial games. None printed anything more than usual.
I was hoping it would :p 

Is this expected, or did I forget something?

ramapcsx2:
Okay, this makes this basically a no op change for most people, but still useful to game modders and the like.
That's fine by me :)

Code looks fine as well, so if you want to leave it as is, I can merge it.

Souzooka:
I did want to make it so messages don't have a newline, but maybe I'll add a ConsoleLogFromVM/ConsoleLogSource hybrid in a separate PR. If you want to merge this it should be a good first step and it stages this feature for a few tweaks later on.

willkuer:
Actually what happens if this is called with more than 7 arguments? Will it crash, show the first seven arguments, show none of the arguments, not show the message at all?

Just curious.

Souzooka:
IIRC, ConsoleLogSource will replace the arguments with defaults (e.g. "0" or "(null)"), like for example if you have the message
`"SET_POINT_CAMERA_NOW( %8.2f, %8.2f, %8.2f, %8.2f, %8.2f, %8.2f, %8.2f, %8.2f, %8.2f, %8.2f, %8.2f, %8.2f, 900, 0, POINT_CUBIC )"`, it might get printed out like `"SET_POINT_CAMERA_NOW( 9.675, 2, 10, 18, -20, 8.5, 2.125, 0, 0, 0, 0, 0, 900, 0, POINT_CUBIC)"`.

ramapcsx2:
Just a quick question:
Is there an identifiable pattern to these no-op'd printf functions in games?
It'd be nice if PCSX2 could attempt to restore them automatically.

Souzooka:
To get back to you on that, unfortunately I don't really know the answer to this since I'm really only interested in poking at Raw Danger, and other games may have different ways of doing debug, you would have to look at other games and see if there are certain patterns to stubbed printing functions.

PR #2564, opened 2018-08-23
GSdx:Windows: Show different adapters for each renderer

turtleli:
Main change:
 * Show different adapters for each renderer. An adapter cannot be selected on OpenGL, and the available adapters available for D3D11 and D3D9 may be slightly different.

Other changes (that make the above change easier):
 * Don't refer to D3D10 - D3D11 is being used even if the user's GPU only supports D3D feature level 10.0.
 * Move D3D11 and D3D9 adapter enumeration code into separate functions.
 * Open sub dialogs using temporary objects - it simplifies the logic slightly as the dialog state won't be partially preserved if it's closed and reopened.
 * Initialise hack dialog variables in constructor - not exactly part of the PR, but it does kinda follow on from the previous change.

adapt


lightningterror:
Maybe also grey out the _Adapter:_ text in gl ? Since that's how the rest of the gsdx options work.

plugins/GSdx/GSdx.rc
```diff
@@ -286,14 +284,13 @@ FONT 8, "MS Shell Dlg", 400, 0, 0x1
...
     CONTROL         IDB_LOGO9,IDC_LOGO9,"Static",SS_BITMAP | SS_REALSIZECONTROL,6,6,230,44
-    CONTROL         IDB_LOGO10,IDC_LOGO10,"Static",SS_BITMAP | SS_REALSIZECONTROL,6,6,230,44
     CONTROL         IDB_LOGO11,IDC_LOGO11,"Static",SS_BITMAP | SS_REALSIZECONTROL,6,6,230,44
     CONTROL         IDB_LOGOGL,IDC_LOGOGL,"Static",SS_BITMAP | SS_REALSIZECONTROL,6,6,230,44
     CONTROL         IDB_NULL,IDC_NULL,"Static",SS_BITMAP | SS_REALSIZECONTROL,6,6,230,44
-    LTEXT           "Adapter:",IDC_STATIC,6,57,30,8
-    COMBOBOX        IDC_ADAPTER,71,55,166,118,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
-    LTEXT           "Renderer:",IDC_STATIC,6,72,34,8
-    COMBOBOX        IDC_RENDERER,71,70,166,118,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
+    LTEXT           "Renderer:",IDC_STATIC,6,57,30,8
```
turtleli: Label length needs to be increased - `:` is not visible in the images.

turtleli:
Updated to disable `Adapter:` text when OpenGL or Null is selected and to fix the label truncation issue.

lightningterror:
While you're at it you could also disable/grey out Texture Filtering if you like.

turtleli:
No thanks. ;)

I usually prefer to leave GSdx Windows GUI changes to other people.

turtleli:
> What's the status ?

Not sure. Might make your PC implode.

lightningterror:
> Not sure. Might make your PC implode.

http://www.nooooooooooooooo.com/

Issue #2441, opened 2018-05-16
[GSDX] Metal Gear Solid 3: "The boss" battle Hardware mode crash on OpenGL


atomic83GitHub:
***PCSX2 version:***
1.5dev 2352

***PCSX2 options:***
All defaults

***Plugins used:***
GSDX, all defaults

***Description of the issue:***

I does a complete run of this game and I encounter 2 big hardware mode issues nearly at the end of the game:

-When you failed at "the boss" battle, a cutsceen will be trigered, at a certain point during it, PCSX2 will crash if GSDX is set to Hardware mode (https://github.com/PCSX2/pcsx2/issues/2436). There is no way to avoid this appart of using the software render or DX11. This issue will also affect the cutsceen you get if you won the game. About the Skip FMV patch, it will not work here.

-Channel shuffle effect issue.

![capture](https://user-images.githubusercontent.com/26351275/40139857-7fb1bbb0-5951-11e8-84b2-e54f977a844c.PNG)

***How to reproduce the issue:***

Loss the "The Boss" battle or finish the game

***Regression:***

PCSX2 v1.3.1-204-ga0c17ae0a introduce the issue.

***PC specifications:***
AMD fx8350  GTX 1060 Windows 10 64 Bit

***Special notes:***

-They are the problematics cutscene for those who have finished the game and can acces to the "Demo" mode.

![capture2](https://user-images.githubusercontent.com/26351275/40140045-1fd7aa1e-5952-11e8-8d12-29f15fc80b85.PNG)

![capture3](https://user-images.githubusercontent.com/26351275/40140153-6fd10da8-5952-11e8-9cea-9fc9ffd486a9.PNG)

MrCK1:
Sounds unlikely, I'll need a savestate please.

atomic83GitHub:
Here is a savestate of the "Boss" bad ending.

[MGS3 crash issue.zip](https://github.com/PCSX2/pcsx2/files/2010623/MGS3.crash.issue.zip)

Helium-4:
Doesn't affect DX11 only OGL.

lightningterror:
What if you disable blending and depth ?

Also can you upload a block dump so I can test ?

atomic83GitHub:
So here is a Blockdump + savestate of that issue (the file is too big for github).

https://drive.google.com/open?id=15n8eK_6e2PpPhNEXXxi2EocXk1sGCO-a

[...]

PR #2566, opened 2018-08-25
gsdx ogl; support dynamic growing of Vertices/Indexes buffer

gregory38:
It should take care of issue #2441 

Not tested


lightningterror:
Still an issue, `I get Reallocate the VERTICES buffer to 0 MB` in the console.

Triggers assert
`ASSERT(current_chunk > 0 && current_chunk < 5);`

Both chunk values are 0.

gregory38:
Try again :) Condition on the while was bad.

plugins/GSdx/GSVertexArrayOGL.h
```diff
@@ -40,29 +40,41 @@ class GSBufferOGL {
...
-		, m_target(target)
+	void grow()
 	{
-		glGenBuffers(1, &m_buffer_name);
-		// Opengl works best with 1-4MB buffer.
-		// Warning m_limit is the number of object (not the size in Bytes)
-		m_limit = 8 * 1024 * 1024 / STRIDE;
+		unallocate_gl_buffer();
+
+		while (m_count => m_limit) {
```
lightningterror: I get these here

```
Error	C2143	syntax error: missing ';' before '{'
Error	C2059	syntax error: '>'
```

lightningterror:
I don't get a crash anymore but the picture is completely black after the part where the crash should have happened. Sound still plays so it's some progress.

gregory38:
Ok. It probably failed to get the new buffer on the gl layer. Do you have any GL errors on a dbg build ?

lightningterror:
Here's a portion of the log
https://pastebin.com/R1S0k5UE

And the entire log:
[GSdx_opengl_debug_hw.zip](https://github.com/PCSX2/pcsx2/files/2320902/GSdx_opengl_debug_hw.zip)

gregory38:
No clue. Do you have a realloc on both buffers ?

lightningterror:
All I see is
Reallocate the VERTICES buffer to 0 MB

turtleli:
It crashes on my PC after the realloc (null pointer within NVIDIA OpenGL driver). Maybe it's because glBufferStorage allocates immutable storage that cannot be reallocated?

turtleli:
I don't think the crash can be explained in any other way though. Nothing else looks suspicious to me.

If the vertex buffer is initially 16MB then it can get through the blockdump without issues.

gregory38:
It smells like a driver bug. I was afraid of a space issue because I think the buffer should only be deleted once it is unused (should be ref counted in the driver). However, in order to test any dump, I used a very small buffer. 

From the man,
https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glBufferStorage.xhtml
=> GL_INVALID_OPERATION is generated by glBufferStorage if the GL_BUFFER_IMMUTABLE_STORAGE flag of the buffer bound to target is GL_TRUE.

An alternate solution is to
* always have a big buffer
* use a CRC (I'm really surprised that a single game used so much vertices)

turtleli:
I guess they're not expecting the buffer to be deleted so it can be resized?

> An alternate solution is to
> 
> always have a big buffer
> or use a CRC (I'm really surprised that a single game used so much vertices)

Not sure how the CRC would work. I think the buffers will have already been allocated by the time GSdx receives the CRC.

gregory38:
glGen*/glDelete are constructor/desctructor of gl state objects. When the object is created you can allocate a buffer once within the object with glBufferStorage.

I don't have a crash without the delete. So I will try to understand the black screen issue first. Maybe it would help to avoid the crash.

lightningterror:
It also crashes with Intel so it's only on amd it doesn't crash.

gregory38:
I'm curious on Mesa. Anyway, if I don't delete the buffer, I got a black screen too. I don't understand why but it feels like it still uses the old buffer.

gregory38:
but black screen ?

gregory38:
maybe I should recreate the full VAO

gregory38:
New version that recreate the full VAO. It seems to work on my poor testcase.

gregory38:
By the way, I think the issue is that vertex format is somehow linked to the gl buffer.

turtleli:
> Reduce the initial size from 256MB to 128MB

? Master branch can hold 256k vertices and 4M indices. This PR can hold 4M vertices and 4M indices. So the initial size of the vertex buffer is increasing from 8 MB to 128MB.

lightningterror:
It works here as well on all 3. Mesa, amd and intel.

gregory38:
Obviously my math was wrong ! In this case, we can do an hardcoded factor 2. 16 extra MB should be fine. Or I tune this PR with better initial value and support of separate count for vertex/index. I don't know.

gregory38:
I will do another PR

PR #2332, opened 2018-03-02
GameDB: Patch update and EETimingHack fix for Drakan 2 demo and Ruff Trigger

atomic83GitHub:
I update some patch file and add missing EETimingHack to Drakan 2 demo and Ruff Trigger.
Tekken 4 patches will be remove when the @FlatOutPS2 IPU PR will be available.

Close: #1496

PR #2570, opened 2018-08-29
GameDB: Add 136 games!!!!!

turtleli:
Just kidding. Sort of.

Essentially, trailing whitespace causes issues because the GameDB multiline section parser sucks.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -3152,7 +3152,7 @@ Name   = Boku to Mao
 Region = NTSC-J
 Compat = 5
 [patches = EE838B5C]
-	comment=patches by Nachbrenner 
+	comment=patches by Nachbrenner
 	// Skip OP.PSS.
 	patch=0,EE,001d6258,word,24020001
 [/patches]
@@ -9913,7 +9913,7 @@ Compat = 5
 	comment=patches by Nachbrenner
 	// Fix IPU busy!
 	patch=0,EE,003300b0,word,03e00008
-	patch=0,EE,003300b4,word,00000000 
+	patch=0,EE,003300b4,word,00000000
 [/patches]
 ---------------------------------------------
 Serial = SLES-51496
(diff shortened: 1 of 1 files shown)


avih:
Interesting. Does this mean these serials were never matched till now? What were the implications of this issue, i.e. what does it actually fix in terms of what the user sees?

Also, at least for a while, the games DB file was auto updated using https://github.com/PCSX2/pcsx2/blob/master/tools/GameIndex.dbf-tool/pcsx2-GameIndex.dbf-updater.html  (it takes the compatibility db from the forum and updates the compatibility info at the DB accordingly).

I don't recall if it trimmed the serials, but I think it did. Does this mean these entries were added manually?

turtleli:
> I now see "Fixes an issue where 136 games are devoured by the parser (9632 -> 9768)."
> So they were completely not recognized and not parsed into the runtime DB?

Yes. I noticed because I was looking on the forums and saw a higher GameDB count on 1.4.0, which didn't make sense.

>who added the patches and didn't verify that they're actually doing something?!

Regression was introduced in #2332.

> I think a better solution would be to fix the parser TBH.

Agreed. I decided this would do while I figure out a fix.

avih:
> I decided this would do while I figure out a fix.

Dunno. Supposedly it never worked till now, so IMO there's no urgency to fix the DB and just hide the problem.

turtleli:
> IMO there's no urgency to fix the DB

Well, some of the serials that were eaten by the parser have automatic gamefixes or MemCardFilter entries present, so I think it's better to fix the DB now.

avih:
Supposedly it never worked... so there's no regression here, hence IMO no urgency.

Also, leaving it unfixed for now will allow better testing of the parser fix with real-world cases.

I'm just saying that if you have vague plans to fix the parser, then wait with this PR until after fixing the parser.

turtleli:
The entries that were eaten by the parser worked before #2332 was merged. I'd call that a regression.

avih:
I still think the parser should be fixed before fixing the DB, but it's not worth arguing over. Do what you think is right.

lightningterror:
I guess I should have cleaned the entire DB when I was doing it in
https://github.com/PCSX2/pcsx2/commit/a6db486afe306d9ff90d939305c14efd3204b34a
https://github.com/PCSX2/pcsx2/commit/810bec2d65a8e1f3b23a225325ce9433aabcce0d

PR #2571, opened 2018-08-29
pcsx2|utilities: Remove potentially ambiguous wxString-related overloads

turtleli:
Avoid some function overloads that can accept char*, wxChar* and wxString as a parameter. wxString can be constructed with either of those parameters, so the function overloads may actually result in ambiguity errors. Keep the wxString parameter versions and remove the rest.

Fixes some compile errors on FreeBSD.


lightningterror:
This breaks Debug Spu2-x compile on Windows.

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	C2712	Cannot use __try in functions that require object unwinding	utilities	D:\My Documents\GitHub\lightningterror-pcsx2\pcsx2\common\src\Utilities\ThreadTools.cpp	684	
```

PR #2575, opened 2018-09-03
utilities: Split thread internal callback function

turtleli:
__try is used in pthread_cleanup_push when CLEANUP_SEH is used as the pthread cleanup model. That can't be used in functions with objects that have destructors, so move it into a separate function.

Fixes a non-release build compile error on Windows. Regression was introduced in 93d5b52df34c03a2ed859bba6e06fead3b875e70.


lightningterror:
Works now :+1:

turtleli:
Updated to fix `--enable-stl` builds too.

PR #2576, opened 2018-09-04
PCSX2: Add ability to pass launch arguments to games with CLI option '--gameargs="-arg1 -arg2 -etc

Iritscen:
Okay, so I'm sure you're going to have some questions about this :-)  I will try to answer some in advance with a FAQ-style introduction:

**Can't PCSX2 already do this?**
From 2002 to 2009, PCSX2 had the ability to pass launch arguments to the game, though only in the developer build.  This code was contributed by "Florin" (not sure if he's still around).  Various major revisions to the code, such as implementing proper BIOS-based booting and moving the GUI to wxWidgets caused his contribution to disappear from the codebase.  The approach he used will no longer work with the current state of PCSX2's code, but I used it as an inspiration for my work.

There was also an "#if 0" block in the current code (no longer present after my commit) which lamented the fact that supplying launch args through $a0 and $a1 to EELOAD's main() did not work.  In my experience, the actual solution is a good bit trickier.

**How does this work? (short version)**
On full boot, I hook the call "EELOAD rom0:PS2LOGO [game]" and add the args after the game by simply supplementing $a0 and $a1.  PS2LOGO then passes all the args to the game itself when it is finished running.

On fast boot, I overwrite some more of EELOAD's memory to insert the game args alongside the game's ELF name that PCSX2 is already injecting.

**How does this work? (long version)**
A new hook is added to EELOAD besides the one that hooks the start of main() to enable fast boot.  The second hook is placed at the jump to a function that has been called ExecExecutable(), which in turn calls ExecPS2().  My second hook places the user's desired arguments into memory (using a block that seems to be free within EELOAD), then places argc and argv into $a0 and $a1 just as ExecExecutable() loads.

I endeavored to leave the code for the main() hook as unaltered as possible.  I only made one variable global, and simplified the code which replaces "rom0:OSDSYS" with the game ELF name.  Now that code simply clobbers the existing string in EELOAD with the game name instead of writing a new string into "PSM(cpuRegs.GPR.n.a1.UD[0] + 0x40)".

I did this for two reasons: (1) I'll be honest, I didn't understand the current approach, and more importantly, (2) I needed space to build a vargs block*, which I am centering around the string "rom0:OSDSYS".  This takes advantage of at least 16 int32s of free space which come before the string, which is where I place the argv pointers.  The arg strings supplied by the user get written right after the ELF name that is written over top of the OSDSYS string.  This clobbers some other strings in EELOAD, but I don't think it can possibly matter because it happens moments before EELOAD calls ExecPS2() and exits.

To illustrate the flow of the code that I am hooking, the attached decompilation "EELOAD.c" is taken from [this forum post](http://psx-scene.com/forums/f19/documentation-what-loadexecps2-can-do-inner-workings-rom0-eeload-102537/) by SP193.  I have added comments to a couple ExecExecutable() calls to aid understanding.  Note that this is an early version of EELOAD (pre-moduleload2), but the logic is similar in later versions of EELOAD, according to my disassembly work.

*I am referring to a block of memory that contains the argv pointers, immediately followed by the arg strings themselves.  Sometimes a block like this places argc before the start of argv[].  I learned about this block from Florin's comments on his code, so a couple of my console messages have named the block in his honor.  The block I construct in memory does not start with argc, however.

**This isn't some kind of nasty kludge, is it?**
Um, next question :0)  No, I actually feel pretty confident about the solidity of this code.  I don't know if there was a better way to do it, but I tried 50 other things before I settled on this approach.  Admittedly I was learning as I went, and don't have the familiarity with PS2 internals that the PCSX2 team does.  As far as I can tell, though, EELOAD is unwilling to accept arguments upon its first call (made by the kernel?).  My method works around that mysterious limitation.

The one catch with this approach is that the hook point for ExecExecutable() varies by BIOS.  I have identified four different types of EELOAD across all BIOSs which I've lettered A-D, but there might be others.  The BIOSs I have documented are listed in the code's comments.  If the code does not recognize the EELOAD type then it simply won't try to hook ExecExecutable() and will advise the user to report the issue to the dev team so you can figure out what BIOS they're using.  I personally tested the code in full boot and fast boot modes with BIOS versions 1.0, 1.5, 1.6, and 1.7, which covers all four types of EELOAD that I have identified.

**Where is this feature useful?**
Well, homebrew games could use it.  But in my case, I had two actual PS2 games that I used as a testbed for the feature.  See the attached .bat file for ready-made test commands utilizing these games.
- Jak & Daxter (USA): This game supports launch arguments if one supplies the attached patch.  I have not published this patch anywhere, and am not even sure where one would publish such a thing since I am new to the community.
- Oni 2 (unreleased): This development build leaked onto the Internet last year (documentation [here](https://wiki.oni2.net/Oni_2_(Angel_Studios)/Running#Launch_arguments)).  I don't know what your policy is on something like this, so I'm hesitant to link to it in this initial comment.  I don't think it would be considered piracy since the game never went on sale.  In any case, the launch args are still functional since the game was in development.  The initial reason I embarked on this whole task was to find a simpler way to pass arguments to the game than the uLaunchELF-based toolkit which I created before this.

I have not explored any other PS2 titles for launch args, but perhaps various players will be interested in hunting for arguments in the binaries of their favorite games.

**Conclusion**
Okay, that's all I can think to mention for starters.  I also put a lot of documentation into the comments.  Sorry if it's overkill; I could be accused of over-commenting at times :-)  I am new to PCSX2 and am open to any criticism/suggestions that you have for me.

By the way, I did not attempt to add this feature to the GUI because it seemed like it would be presumptuous.  First I want to see if this code meets your approval.  I would assume that the main dev team prefers to implement GUI changes anyway, but I do know how one could fit the feature into the current UI and won't mind proposing something in a PR if you want me to.

P.S.: Enabling DEBUG_LAUNCHARG may prove helpful in following the flow of the code.

[launchargs_pr.zip](https://github.com/PCSX2/pcsx2/files/2349286/launchargs_pr.zip)

### DIFF
--- a/pcsx2/Interpreter.cpp
+++ b/pcsx2/Interpreter.cpp
@@ -522,15 +522,35 @@ static void intExecute()
 				case RESET:
 					do
 						execI();
-					while (cpuRegs.pc != (eeloadMain ? eeloadMain : EELOAD_START));
+					while (cpuRegs.pc != (g_eeloadMain ? g_eeloadMain : EELOAD_START));
 					if (cpuRegs.pc == EELOAD_START)
 					{
 						// The EELOAD _start function is the same across all BIOS versions afaik
 						u32 mainjump = memRead32(EELOAD_START + 0x9c);
 						if (mainjump >> 26 == 3) // JAL
-							eeloadMain = ((EELOAD_START + 0xa0) & 0xf0000000U) | (mainjump << 2 & 0x0fffffffU);
-					} else if (cpuRegs.pc == eeloadMain)
+							g_eeloadMain = ((EELOAD_START + 0xa0) & 0xf0000000U) | (mainjump << 2 & 0x0fffffffU);
+					}
+					else if (cpuRegs.pc == g_eeloadMain)
+					{
 						eeloadHook();
+						if (g_SkipBiosHack)
+						{
+							// See comments on this code in iR5900-32.cpp's recRecompile()
+							u32 typeAexecjump = memRead32(EELOAD_START + 0x470);
+							u32 typeBexecjump = memRead32(EELOAD_START + 0x5B0);
+							u32 typeCexecjump = memRead32(EELOAD_START + 0x618);
+							u32 typeDexecjump = memRead32(EELOAD_START + 0x600);
+							if ((typeBexecjump >> 26 == 3) || (typeCexecjump >> 26 == 3) || (typeDexecjump >> 26 == 3)) // JAL to 0x822B8
+								g_eeloadExec = EELOAD_START + 0x2B8;
+							else if (typeAexecjump >> 26 == 3) // JAL to 0x82170
+								g_eeloadExec = EELOAD_START + 0x170;
+							else
+								Console.WriteLn("recRecompile: Could not enable launch arguments for fast boot mode; unidentified BIOS version! Please report this to the PCSX2 developers.");
+						}
+					}
+					else if (cpuRegs.pc == g_eeloadExec)
+						eeloadHook2();
+
 					if (g_GameLoading)
 						state = GAME_LOADING;
 					else
--- a/pcsx2/Interpreter.cpp
+++ b/pcsx2/Interpreter.cpp
@@ -545,7 +545,7 @@ static void intExecute()
 							else if (typeAexecjump >> 26 == 3) // JAL to 0x82170
 								g_eeloadExec = EELOAD_START + 0x170;
 							else
-								Console.WriteLn("recRecompile: Could not enable launch arguments for fast boot mode; unidentified BIOS version! Please report this to the PCSX2 developers.");
+								Console.WriteLn("intExecute: Could not enable launch arguments for fast boot mode; unidentified BIOS version! Please report this to the PCSX2 developers.");
 						}
 					}
 					else if (cpuRegs.pc == g_eeloadExec)
--- a/pcsx2/R5900.cpp
+++ b/pcsx2/R5900.cpp
@@ -562,11 +562,11 @@ int ParseArgumentString(u32 arg_block)
 	for (int i = 0; i < args_len; i++)
 	{
 		char curchar = *(char *)PSM(arg_block + i);
-		bool isSpace = curchar == ' ' ? true : false;
-		bool isNull = curchar == '\0' ? true : false;
-		if (isNull)
+		if (curchar == '\0')
 			break; // should never reach this
-		else if (isSpace)
+
+		bool isSpace = (curchar == ' ');
+		if (isSpace)
 			memset(PSM(arg_block + i), 0, 1);
 		else if (wasSpace) // then we're at a new arg
 		{
--- a/pcsx2/R5900.cpp
+++ b/pcsx2/R5900.cpp
@@ -626,10 +626,9 @@ void __fastcall eeloadHook()
 		// then we add the desired launch arguments. PS2LOGO passes those on to the game itself as it calls EELOAD a third time.
 		if (!g_Conf->CurrentGameArgs.empty() && !strcmp(elfname.c_str(), "rom0:PS2LOGO"))
 		{
-#if DEBUG_LAUNCHARG
 			const char *argString = g_Conf->CurrentGameArgs.c_str();
-			Console.WriteLn("eeloadHook: Inserting launch argument(s) '%s'...", argString);
-#endif
+			Console.WriteLn("eeloadHook: Supplying launch argument(s) '%s' to module '%s'...", argString, elfname.c_str());
+
 			// Join all arguments by space characters so they can be processed as one string by ParseArgumentString(), then add the
 			// user's launch arguments onto the end
 			u32 arg_ptr = 0;
@@ -721,10 +720,9 @@ void __fastcall eeloadHook2()
 		return;
 	}
 
-#if DEBUG_LAUNCHARG
 	const char *argString = g_Conf->CurrentGameArgs.c_str();
-	Console.WriteLn("eeloadHook2: Attempting to supply launch argument(s) '%s' to ELF '%s'.", argString, (char *)PSM(g_osdsys_str));
-#endif
+	Console.WriteLn("eeloadHook2: Supplying launch argument(s) '%s' to ELF '%s'.", argString, (char *)PSM(g_osdsys_str));
+
 	// Add args string after game's ELF name that was written over "rom0:OSDSYS" by eeloadHook(). In between the ELF name and args
 	// string we insert a space character so that ParseArgumentString() has one continuous string to process.
 	int game_len = strlen((char *)PSM(g_osdsys_str));
--- a/pcsx2/R5900.cpp
+++ b/pcsx2/R5900.cpp
@@ -57,11 +57,13 @@ static const uint eeWaitCycles = 3072;
 bool eeEventTestIsActive = false;
 
 u32 g_eeloadMain = 0, g_eeloadExec = 0, g_osdsys_str = 0;
-const int kMaxArgs = 16; /* I don't know how much space for args there is in the memory block used for args in full boot mode,
-							but in fast boot mode, the block we use can fit at least 16 argv pointers (varies with BIOS version).
-							The second EELOAD call during full boot has three built-in arguments ("EELOAD rom0:PS2LOGO <ELF>"),
-							meaning that only the first 13 game arguments supplied by the user can be added on and passed through.
-							In fast boot mode, 15 arguments can fit because the only call to EELOAD is "<ELF> <<args>>". */
+
+/* I don't know how much space for args there is in the memory block used for args in full boot mode,
+but in fast boot mode, the block we use can fit at least 16 argv pointers (varies with BIOS version).
+The second EELOAD call during full boot has three built-in arguments ("EELOAD rom0:PS2LOGO <ELF>"),
+meaning that only the first 13 game arguments supplied by the user can be added on and passed through.
+In fast boot mode, 15 arguments can fit because the only call to EELOAD is "<ELF> <<args>>". */
+const int kMaxArgs = 16;
 uptr g_argPtrs[kMaxArgs];
 #define DEBUG_LAUNCHARG 0 // show lots of helpful console messages as the launch arguments are passed to the game
 
@@ -555,12 +557,12 @@ int ParseArgumentString(u32 arg_block)
 
 	int argc = 1; // one arg is guaranteed at least
 	g_argPtrs[0] = arg_block; // first arg is right here
-	bool isSpace = false, wasSpace = false; // status of current char. and last char. scanned
+	bool wasSpace = false; // status of last char. scanned
 	int args_len = strlen((char *)PSM(arg_block));
 	for (int i = 0; i < args_len; i++)
 	{
 		char curchar = *(char *)PSM(arg_block + i);
-		isSpace = curchar == ' ' ? true : false;
+		bool isSpace = curchar == ' ' ? true : false;
 		bool isNull = curchar == '\0' ? true : false;
 		if (isNull)
 			break; // should never reach this
@@ -608,7 +610,7 @@ void __fastcall eeloadHook()
 	if (argc) // calls to EELOAD *after* the first one during the startup process will come here
 	{
 #if DEBUG_LAUNCHARG
-		Console.WriteLn("eeloadHook: EELOAD was called with %d arguments according to $a0 and %d according to Florin's block:",
+		Console.WriteLn("eeloadHook: EELOAD was called with %d arguments according to $a0 and %d according to vargs block:",
 			argc, memRead32(cpuRegs.GPR.n.a1.UD[0] - 4));
 		for (int a = 0; a < argc; a++)
 			Console.WriteLn("argv[%d]: %p -> %p -> '%s'", a, cpuRegs.GPR.n.a1.UL[0] + (a * 4),
--- a/pcsx2/R5900.cpp
+++ b/pcsx2/R5900.cpp
@@ -56,7 +56,14 @@ static const uint eeWaitCycles = 3072;
 
 bool eeEventTestIsActive = false;
 
-u32 eeloadMain = 0;
+u32 g_eeloadMain = 0, g_eeloadExec = 0, g_osdsys_str = 0;
+const int kMaxArgs = 16; /* I don't know how much space for args there is in the memory block used for args in full boot mode,
+							but in fast boot mode, the block we use can fit at least 16 argv pointers (varies with BIOS version).
+							The second EELOAD call during full boot has three built-in arguments ("EELOAD rom0:PS2LOGO <ELF>"),
+							meaning that only the first 13 game arguments supplied by the user can be added on and passed through.
+							In fast boot mode, 15 arguments can fit because the only call to EELOAD is "<ELF> <<args>>". */
+uptr g_argPtrs[kMaxArgs];
+#define DEBUG_LAUNCHARG 0 // show lots of helpful console messages as the launch arguments are passed to the game
 
 extern SysMainMemory& GetVmMemory();
 
@@ -110,7 +117,7 @@ void cpuReset()
 	// is why I found out about this) --air
 	LastELF = L"";
 
-	eeloadMain = 0;
+	g_eeloadMain = 0, g_eeloadExec = 0, g_osdsys_str = 0;
 }
 
 void cpuShutdown()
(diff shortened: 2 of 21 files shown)


pcsx2/R5900.cpp
```diff
@@ -56,7 +56,14 @@ static const uint eeWaitCycles = 3072;
 
 bool eeEventTestIsActive = false;
 
-u32 eeloadMain = 0;
+u32 g_eeloadMain = 0, g_eeloadExec = 0, g_osdsys_str = 0;
+const int kMaxArgs = 16; /* I don't know how much space for args there is in the memory block used for args in full boot mode,
```
lightningterror: I prefer is the entire comment is above kMaxArgs.

Iritscen: Sure, makes sense to me; I will change that.  (Going away for some hours, but I will reply to other comments when I'm back.)

Iritscen:
@refractionpcsx2:
1. Admittedly it's an esoteric feature.  The very nature of a console prevents a player from normally supplying launch args to a game, so there won't be any CLI options that the developers build in for players to use.  The potential usefulness is to aid homebrew developers, and for fans who like hunting for hidden arguments in games.

One such fan reached out to me about the Jak & Daxter games, because he is currently exploring and documenting them on a wiki.  He had used my ps2launchargs toolkit (the one based on uLE+pcsx2 which I mentioned in the PR) and was asking for some help since the games were not responding to the arguments.  Jak1 required the attached .pnach to unlock launch args, but I also decided that if other gamers were interested in this feature, and PCSX2 itself had the ability to supply launch args to games, then my toolkit's complex workaround would not be needed, and maybe other players might start searching game binaries for launch args.  Who knows what might be out there in some popular titles?  So it's a feature which would benefit hackers more than casual players, but you never know how that might trickle down, in terms of hackers revealing goodies to players ("Use this launch arg to do X and Y").

2. The style of PCSX2's CLI seems to be "pcsx2.exe [iso] --all-other-options", where only the ISO can be supplied without a switch.  I suppose it could work like this: "pcsx2.exe [iso] [arg1] [arg2] --all-other-options", if that's what you're suggesting.

Honestly, I wrote the code to expect a single string of args because I was thinking of the launch arg dialog box that I would like to see incorporated into the GUI.  Florin's dialog in the old days took all the arguments as a single string from the text field in the window and then converted it into a vargs block.  So I was picturing that GUI in my head, and simply modeled the CLI option on the same approach for maximum code reuse :-)

refractionpcsx2:
I just don't like how the args are being passed that's all, it seems nasty :P If it used to do that, well, it shouldn't anymore IMO! it would be better having it straight in the command line, as long as other functionality isn't hindered in the process.

Iritscen:
I totally understand being concerned about the changes to the boot process.  In the case where there are _no launch args_ (so, an average player's usage of PCSX2), the only functional changes to the code are:

- Full and fast boot: A second hook is being installed -- however the hook function, eeloadHook2(), returns immediately because there are no arguments to supply to the game.

- Fast boot only: "rom0:OSDSYS" is being overwritten with the game's ELF name instead of changing the argv pointer which points there to point to the game ELF name instead (for the reason given in my OP).  I definitely welcome feedback from any specialists on this part of the code as to whether this change could be problematic.  I saw no issues myself in testing.

As far as the launch arg syntax, if it's the general consensus that I should change it to "pcsx2.exe iso_name -arg1 -arg2 -etc --all-other-options", I will do so.  In that case I will probably just rewrite all of my code that handles the argument string(s), since it will make more sense to store the arguments in an array of char * or wxString rather than a single continuous string.  That being said, the launch dialog will work seamlessly with the current code, assuming that a new argument dialog looks like the old one....
sample

Hirato:
Most Linux utilities separate their arguments and the arguments of whatever they're running with either `--`, or a `--args` switch.
eg: `gdb prog.bin --gdb-opt1 --gdb-opt2 --args --prog-opt1 --prog-opt2`

Maybe do that instead - should be intuitive for anyone who is likely to invoke a game via CLI.

pcsx2/R5900.cpp
```diff
@@ -540,6 +549,49 @@ void __fastcall eeGameStarting()
...
+
+	int argc = 1; // one arg is guaranteed at least
+	g_argPtrs[0] = arg_block; // first arg is right here
+	bool wasSpace = false; // status of last char. scanned
+	int args_len = strlen((char *)PSM(arg_block));
+	for (int i = 0; i < args_len; i++)
+	{
+		char curchar = *(char *)PSM(arg_block + i);
+		bool isSpace = curchar == ' ' ? true : false;
+		bool isNull = curchar == '\0' ? true : false;
```
ssakash: redundant ternary operators, coverity will complain about this.

Iritscen: Ah, okay.  I don't know why I did it that way.  I should just change it to "bool isSpace = (curchar == ' '); bool isNull = (curchar == '\0');", right?

pcsx2/R5900.cpp
```diff
@@ -571,44 +686,73 @@ void __fastcall eeloadHook()
...
+
+	if (!g_osdsys_str)
+	{
+		Console.WriteLn("eeloadHook2: Called before \"rom0:OSDSYS\" was found by eeloadHook()!");
+		return;
+	}
+
+#if DEBUG_LAUNCHARG
+	const char *argString = g_Conf->CurrentGameArgs.c_str();
+	Console.WriteLn("eeloadHook2: Attempting to supply launch argument(s) '%s' to ELF '%s'.", argString, (char *)PSM(g_osdsys_str));
```
ramapcsx2: At least one of these DEBUG_LAUNCHARG only messages should always show, so users see that it's being attempted.
If it's not this one, please choose the one that's most likely to cause issues.

Thanks! :)

pcsx2/x86/ix86-32/iR5900-32.cpp
```diff
@@ -1651,15 +1651,33 @@ static void __fastcall recRecompile( const u32 startpc )
...
+			u32 typeBexecjump = memRead32(EELOAD_START + 0x5B0); // v1.20, v1.50, v1.60 (3000x models)
+			u32 typeCexecjump = memRead32(EELOAD_START + 0x618); // v1.60 (3900x models)
+			u32 typeDexecjump = memRead32(EELOAD_START + 0x600); // v1.70, v1.90, v2.00, v2.20, v2.30
+			if ((typeBexecjump >> 26 == 3) || (typeCexecjump >> 26 == 3) || (typeDexecjump >> 26 == 3)) // JAL to 0x822B8
+				g_eeloadExec = EELOAD_START + 0x2B8;
+			else if (typeAexecjump >> 26 == 3) // JAL to 0x82170
+				g_eeloadExec = EELOAD_START + 0x170;
+			else // There might be other types of EELOAD, because these models' BIOSs have not been examined: 18000, 3500x, 3700x,
+				 // 5500x, and 7900x. However, all BIOS versions have been examined except for v1.01 and v1.10.
+				Console.WriteLn("recRecompile: Could not enable launch arguments for fast boot mode; unidentified BIOS version! Please report this to the PCSX2 developers.");
```
ramapcsx2: Could you explain what happens in this case? 
It would be great if it just continued normally with fast boot.

pcsx2/x86/ix86-32/iR5900-32.cpp
```diff
@@ -1669,6 +1687,9 @@ static void __fastcall recRecompile( const u32 startpc )
 			doPlace0Patches();
 		g_patchesNeedRedo = 0;
 	}
+	
+	if (g_eeloadExec && HWADDR(startpc) == HWADDR(g_eeloadExec))
```
ramapcsx2: Maybe this location could be used to check for arguments. If none are given, then this could skip the new hook altogether, right?

pcsx2/Interpreter.cpp
```diff
@@ -522,15 +522,35 @@ static void intExecute()
...
+							u32 typeAexecjump = memRead32(EELOAD_START + 0x470);
+							u32 typeBexecjump = memRead32(EELOAD_START + 0x5B0);
+							u32 typeCexecjump = memRead32(EELOAD_START + 0x618);
+							u32 typeDexecjump = memRead32(EELOAD_START + 0x600);
+							if ((typeBexecjump >> 26 == 3) || (typeCexecjump >> 26 == 3) || (typeDexecjump >> 26 == 3)) // JAL to 0x822B8
+								g_eeloadExec = EELOAD_START + 0x2B8;
+							else if (typeAexecjump >> 26 == 3) // JAL to 0x82170
+								g_eeloadExec = EELOAD_START + 0x170;
+							else
+								Console.WriteLn("recRecompile: Could not enable launch arguments for fast boot mode; unidentified BIOS version! Please report this to the PCSX2 developers.");
```
Iritscen: Oops. Note to self: fix this console message tomorrow.

pcsx2/R5900.cpp
Iritscen: Thanks for the encouragement. You have some good points and I will make commits soon.  I'll enable one message in each hook function -- one for the full boot arg injection and one for the fast boot injection (which would be the message above).

pcsx2/x86/ix86-32/iR5900-32.cpp
Iritscen: It will.  In this `else` case, g_eeloadExec will remain initialized at zero, so the `if (g_eeloadExec …)` check below will fail and eeloadHook2() will not be called.  I wanted the above code, however, to run on all fast boots as this will allow us to more quickly discover any EELOAD versions which have the ExecExecutable() call in a different place, at which points those EELOADs can be added to the list above.  Admittedly, most users will not notice or report the message in their console.

pcsx2/x86/ix86-32/iR5900-32.cpp
Iritscen: That would be more logical, right?  I was originally going to place the check here instead of in R5900.cpp's eeloadHook2(), but AppConfig.h is not included in this source file and so g_Conf is not known.  I was hesitant to change any includes in the project.  What would you suggest?

pcsx2/R5900.cpp
ramapcsx2: Great!

pcsx2/x86/ix86-32/iR5900-32.cpp
ramapcsx2: That sounds good to me.
I don't expect many more official versions by the way. I just worried about customized or obscure development images that someone may try out.
In those cases, it's okay for this function to bail out and continue normally :)

pcsx2/x86/ix86-32/iR5900-32.cpp
ramapcsx2: Ah yeah. I would have decided against including AppConfig.h as well.
If you could add a comment that explains it, that'd be great :)

@refractionpcsx2 Your opinion on this? :)

willkuer: Why is that so? You want to request a config and accordingly add appconfig to this file then... sounds totally reasonable to me.

lightningterror:
What's the status on this, is it ready ?

Iritscen:
Sorry, I'm not sure if you're addressing the dev team or me :-)  My opinion is that the code works fine in my tests, and should have no undesired side effects.

The only unresolved questions that came up are (1) whether my syntax for the CLI argument "--gameargs" is acceptable and (2) the discussion with ramapcsx2 on whether I should include AppConfig.h in iR5900-32.cpp, which would allow me to avoid calling the new hook function if it's not needed (instead of it immediately returning when it is called and sees that it's not needed).

Personally I'm happy to answer any questions/concerns from reviewers, or perform whatever work I have to do in order to increase the chances of this PR being accepted.

pcsx2/x86/ix86-32/iR5900-32.cpp
ramapcsx2: If I was someone external, proposing my additions, I would try to avoid spreading headers. This is just how I'd do it though, not a technical reason.
I'm fine with the code as is ;)

ramapcsx2:
Hey,
the appconfig header thing:
I'm fine with how it is, doesn't have to be changed. And if someone wants it the other way, that's of course always possible after merging as well.

Please do revert the changes to .gitignore though. I missed that :p 

You don't have to continuously keep your PR updated.
As long as there are no merge conflicts (Github will tell you), it can be squashed and merged easily.

That leaves the argument passing format, which I'm undecided on.
Please just keep in mind that there are 3rd party tools that rely on the args feature.
They should of course stay compatible :)
Example1: https://forums.pcsx2.net/Thread-Spectabis-Advanced-Game-Launcher-Frontend
Example2: https://forums.pcsx2.net/Thread-PCSX2Bonus-A-PCSX2-launcher-frontend

refractionpcsx2:
I'd prefer
`pcsx2.exe ISO --emuarg1 --emuarg2="options" --gameargs -arg1 -arg2`
:P

Iritscen:
Heh, okay.  Well, that's doable, and maybe with minimal change to the existing CLI-parsing code.  I'll probably commit something in the next day.

Iritscen:
Okay, apologies @refractionpcsx2, but after thinking about this more, and looking at PCSX2's code for the CLI again, both of the alternate approaches are going to introduce complications.

For one thing, accepting the arguments right after the ISO name implies that any text that is not understood by PCSX2, up until the first argument that *is* recognized by the emulator, should be assumed to be intended for the game and thus passed to it (keep in mind that we have no idea what syntax a game's arguments might use, so we can't assume it's going to be "-arg1 -arg2"; we can only assume whitespace-separated terms of some kind).  Is that really what you want to happen?  Don't you want the emu to warn the user or abort launch if unknown arguments are being passed to it?  Because passing all unknown text to the game will prevent PCSX2 from recognizing bad arguments intended for the emulator and advising the user about them.

More importantly, PCSX2's argument parsing does not take place in a left-to-right manner; you are simply running tests one-by-one for the various recognized arguments by using Found().  I don't see a way to recognize the concept of "all arguments to the left of the first known argument" (or to the right of `--gameargs`, in the alternate approach that was suggested) without refactoring the whole CLI parsing process.  It would probably require creating a table of all CLI arguments matched with function pointers, and then running lookups against it while iterating through the results of wxCmdLineParser::ConvertStringToArgs() — but this will require rewriting much of AppInit.cpp.

This is the only way to recognize the kind of multi-term linear structure being suggested for the arguments.  Whereas my current approach just works because it is all a single argument from PCSX2's perspective :-)  So I don't mean to argue over this too much, but I want to be sure that we're on the same page as far as the implications of your request.  Or maybe there's a simpler way to work within the existing code that I'm not seeing?

ramapcsx2:
I feared as much, @Iritscen.
If @refractionpcsx2 is okay with it, I'd recommend just going with your current implementation.
Unless you feel like coding up that lookup approach?
But then again, imho this is overkill for a somewhat rarely used feature, isn't it.

Iritscen:
Heh, okay.  Hopefully you don't feel like I've been difficult about this.  In all honesty, I was desperate enough to get this PR accepted that I was planning to just start rewriting the argument-parsing code in PCSX2 and submit that as a PR too.  Just haven't had the time.  But if you're okay with the current --gameargs syntax, which I really think is the best approach, then I guess everything should be ready to go.

pcsx2/gui/AppInit.cpp
```diff
@@ -243,6 +243,7 @@ void Pcsx2App::OnInitCmdLine( wxCmdLineParser& parser )
 	parser.AddSwitch( wxEmptyString,L"nohacks",		_("disables all speedhacks") );
 	parser.AddOption( wxEmptyString,L"gamefixes",	_("use the specified comma or pipe-delimited list of gamefixes.") + fixlist, wxCMD_LINE_VAL_STRING );
 	parser.AddSwitch( wxEmptyString,L"fullboot",	_("disables fast booting") );
+	parser.AddOption( wxEmptyString,L"gameargs",    _("passes the specified space-delimited string of launch arguments to the game"), wxCMD_LINE_VAL_STRING);
```
lightningterror: Tabs instead of spaces after `gameargs",` the other lines use tabs.

Iritscen: Ah, indeed.  Fixed it.

lightningterror:
I'll merge this later since it's good to go.

Iritscen:
Excellent, thanks.  I hope it works well for people who want to explore their favorite games for launch arguments and then try them out.  Just wanted to leave a little reminder here that there is currently no GUI access for this feature.  Maybe this kind of niche feature only needs to be in the CLI, but if a dialog box for launch arguments (like the old one posted in my screenshot above) is something that you guys might want to have, I could make a feature request ticket for it and there can be a discussion... or simply propose a GUI change in a PR... or maybe it's better to leave GUI decisions for the regular dev team.  Just let me know if there's interest in that.

lightningterror:
I think it's fine as it is like this since it's a niche feature, if there is a stronger interest from people then a GUI option can be added in the future.

willkuer:
A gui option can trigger more interest

PR #2577, opened 2018-09-06
GSdx-d3d: Port/change shader code to match opengl

lightningterror:
Details in the commit titles/description.

I'm not sure if the second commit is needed, if not then I could drop it.

### DIFF
--- a/plugins/GSdx/res/tfx.fx
+++ b/plugins/GSdx/res/tfx.fx
@@ -402,7 +402,7 @@ float4 sample(float2 st, float q)
 	{
 		float4 uv;
 
-		if(PS_LTF)
+		if(PS_LTF != 0)
 		{
 			uv = st.xyxy + HalfTexel;
 			dd = frac(uv.xy * WH.zw);
@@ -440,7 +440,7 @@ float4 sample(float2 st, float q)
 		}
 	}
 
-	if(PS_LTF)
+	if(PS_LTF != 0)
 	{	
 		t = lerp(lerp(c[0], c[1], dd.x), lerp(c[2], c[3], dd.x), dd.y);
 	}
--- a/plugins/GSdx/res/tfx.fx
+++ b/plugins/GSdx/res/tfx.fx
@@ -254,7 +254,7 @@ float4 clamp_wrap_uv(float4 uv)
 		else if(PS_WMS == 3)
 		{
 			#if SHADER_MODEL >= 0x400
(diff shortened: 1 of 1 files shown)


plugins/GSdx/res/tfx.fx
```diff
@@ -607,11 +610,15 @@ VS_OUTPUT vs_main(VS_INPUT input)
...
+		input.z = input.z & uint(0xffffff); 
 	}
 	else if(VS_BPPZ == 2) // 16
 	{
-		input.z = input.z & 0xffff;
+		input.z = input.z & uint(0xffff);
+	}
+	else
+	{
+		input.z;
```
turtleli: This does nothing.

plugins/GSdx/res/tfx.fx
```diff
@@ -607,11 +610,15 @@ VS_OUTPUT vs_main(VS_INPUT input)
 {
 	if(VS_BPPZ == 1) // 24
 	{
-		input.z = input.z & 0xffffff; 
+		input.z = input.z & uint(0xffffff); 
 	}
 	else if(VS_BPPZ == 2) // 16
 	{
-		input.z = input.z & 0xffff;
+		input.z = input.z & uint(0xffff);
```
turtleli: HLSL does implicit type casting.

plugins/GSdx/res/tfx.fx
```diff
@@ -381,7 +381,10 @@ float4x4 sample_4p(float4 u)
 
 float4 sample(float2 st, float q)
 {
-	if(!PS_FST) st /= q;
+	if(PS_FST == 0)
```
turtleli: Expression before and after are the same.

lightningterror: I wanted it to match gl code, guess it's a bit of an useless change.

plugins/GSdx/res/tfx.fx
lightningterror: What about gl code, isn't it the same as well ?

turtleli: No. The commit you referenced fixed an uninitialised variable issue.

plugins/GSdx/res/tfx.fx
turtleli: AFAICT this PR doesn't actually change anything.

plugins/GSdx/res/tfx.fx
lightningterror: Ah ok, my main reason for opening was this change.

PR #2585, opened 2018-09-09
pcsx2: Improve GameDB parser

turtleli:
Changes:
 * Abort multiline section parsing if the entry separator is found - it prevents bad formatting from eating subsequent GameDB entries.
 * Warn if the GameDB multiline section start tag has incorrect whitespace.

I'm not sure on whether to make the parser more whitespace tolerant on the closing tag. I think the errors/warnings will hint to whoever is editing the GameDB that something isn't quite right.


willkuer:
Couldn‘t the whole read operation be just one regex?

PR #2586, opened 2018-09-09
Add networking capability for Linux

lowlyocean:
The existing dev9ghzdrk plugin was designed to run on Windows only. I've implemented the Linux equivalent. This includes configuration dialog in GTK+, configuration storage and retrieval using LibXML2, and network adapter enumeration using LibPCAP.

I've moved some files from the Win32 sub-directory up one level and made slight capitalization changes to their #include statements so that a Linux build can also locate the headers.

I've updated the appropriate CMake files to conditionally include the plugin on Linux if necessary libraries are detected. I've also embedded post-install commands to copy necessary UI layout and device configuration files to user's .config directory, as well as to grant necessary privileges to PCSX2 executable. The exact commands are printed to install_log.txt toward completion of the build.

The networking code has been slightly modified; the MAC address of the virtual PS2 network adapter contains real Sony MAC address prefix followed by last two bytes of real adapter's MAC address. This allows detection of virtual PS2 console by the XLink Kai LAN-over-internet service while also allowing each user to appear unique on that service.


lightningterror:
The formatting is all over the place in some files between spaces and tabs, please keep it consistent, also there are many lines that shouldn't have been changed.

radical-equanimity:
Also, tap-win32.cpp fails compilation in Visual Studio build environment, needs someone with proper environment to iron out.

TheLastRar:
I don't know specifically why you decided to get rid of switched mode, I seem to remember that being the more reliable option, although I do recall that what worked varied user to user for some reason.

A lot of the changes made seems to be specifically for bridged mode, which mostly seems to be removing the code to change the PS2's mac to the host machine on sent packets (and host mac to PS2 mac on received). I don't know why this code was here in the first place, presumable there was an issue that this code was working around.

I don't know if WinPcap defines PCAP_NETMASK_UNKNOW, someone would need to check that.

XLink Kai will accept packets from the MAC address that the plugin already has, CLR_DEV9 uses the same prefix and changes just the last two bytes (on both WinPcap and tap).

On linux testing I did in the past, I didn't need to run any specific commands to allow pcap to work, either they aren't needed for switched mode, or mono/wine runs them on install.

lowlyocean:
Bridged mode allows IP addressing to be configurable by user with PS2 software. This address needs to be user configurable because Xlink Kai specifies an IP address user must set based on PS2 MAC address. Switched mode doesn't make much sense for this plugin; plugin only needs to provide Layer 2 processing and let the PS2 CPU process Layer 3 natively at the OS and application layers. Performing layer 3 operations in the plugin is intrusive and may lead to unintended consequences.

Wifi devices in promiscuous mode with Pcap can only intercept frames that have their exact MAC address as the destination. Bridged mode allows us to get around this... we receive packets at the real adapter's MAC address, translate to PS2 mac before sending to game. 

I suspect Mono runtime must have been installed with permission to access network devices in promiscuous mode.

I take care of WinPCap absence of PCAP_NETMASK_UNKNOWN by simply defining it if it isn't already defined.

arcum42:
Hmm... First impression is that it could use some work.

I tried to build it, throwing on the gtk 3 flag to make things interesting, noted that I needed dev versions of pcap and libxml2 installed, installed them,and built it a second time. Seems like a fair number of warnings on building, though a fair amount were similar and probably won't be too difficult to clean up.

On going in the bin directory, I remembered that the inis folder that the *.ui file was being copied to doesn't exist until you run pcsx2 (maybe the plugin directory would work better?). I ran it and looked to see if the plugin was there anyways. Didn't show up in the list, but it was in the plugins directory.

I ran the build again, and it didn't copy it the second time, either. I manually copied it from the source. Still not showing.

I'm now doing a clean gtk 2 build, since I probably should've tried the normal version of gtk first. I do notice that while it found PCAP, it didn't find pcap_get_pfring_id, so there might also be a dependency issue it didn't catch...

lowlyocean:
Regarding copying of ui and cfg files, the INSTALL( CODE commands in CMakelists.txt are conditioned on the OS being reported as Linux. This is the desired behavior, but I've noticed the condition is not a reliable indicator of the OS actually being Linux. On some machines, the condition evaluates false when it should evaluates true. Can someone take a look and suggest a solution?

TheLastRar:
Ah, I had gotten confused with the relation between switch/bridged and promiscuous/non-promiscuous 
Your earlier code was removing bridged mode, not switched mode as I previously stated, apologies for the confusion.
(your latest comment is true of bridged mode (non-promiscuous) and not switched (promiscuous))

I don't know the reason why there is an option for non-promiscuous mode, given that it looks like it would be rather disruptive to unrelated network processes (possibly why I never got it to work, although it seemed to work for others). I can only assume it was for compatibility with drivers that didn't work with promiscuous mode, given the plugin also supports using tap this probably is less of a concern.

The PS2 will read the mac address from eeprom, if you set that then you should be able to avoid replacing the mac address for every packet in and out. 

I note that you removed the code for re-sending broadcast packets to the host machine, was this causing duplicate responses for you?

You've included a memory card file, assuming this contains a network config save, I was under the impression that network configs are tied to the BIOS of the system they where created on, is this not the case?

lowlyocean:
The IP address, netmask, gateway are configured by the Network Disc DVD that shipped with the network adapter hardware.  Or could be configured within certain games like SOCOM. Layer 3 addresses. However, layer 2 MAC addresses were presumably tied to what the network adapter reported to the BIOS. No need to re-send broadcast packets because the host machine gets them anyway. Memory card contains network config but this will be overwritten by users trying to connect to XLink.

arcum42:
> Regarding copying of ui and cfg files, the destination directories are created if not found. The INSTALL( CODE commands in CMakelists.txt are conditioned on the OS being reported as Linux. This is the desired behavior, but I've noticed the condition is not a reliable indicator of the OS actually being Linux. On some machines, the condition evaluates false when it should evaluate true. Can someone take a look and suggest a reliable solution?
> 
> To understand correctly- "Didn't show up in the list" means PCSX2 doesn't recognize the library even though it's in the plugins directory?

Yes. It isn't in the list, and prints that the file is not a PCSX2 plugin on the console.

I will note that the CMake files are not used in windows, so any tests would mostly be if it was Linux, OS X, BSD, etc. I modified the CMakelists.txt file to print out the variable you were checking, and it was listed as Linux for me, though.

However, looking closer, it's copying to ~/.config/pcsx2/inis, and I'm actually running directly from bin, with the inis and memcard folders directly in there. I also didn't note that it had a cfg file and memcard that needed to be copied in. Copying them in directly doesn't seem to have helped with it not being a valid plugin, though...

lowlyocean:
> > To understand correctly- "Didn't show up in the list" means PCSX2 doesn't recognize the library even though it's in the plugins directory?
> 
> Yes. It isn't in the list, and prints that the file is not a PCSX2 plugin on the console.
> 
Pushed changes to address this, please confirm that you can see the plugin and configure it to point to bridge: <your eth device> from the dialog's Ethernet drop-down.
> I will note that the CMake files are not used in windows, so any tests would mostly be if it was Linux, OS X, BSD, etc. I modified the CMakelists.txt file to print out the variable you were checking, and it was listed as Linux for me, though.
I have removed the conditions in my latest upload.
> However, looking closer, it's copying to ~/.config/pcsx2/inis, 
I did development against v1.3.1, where ~/.config/pcsx2 seemed to be the right path to place config files. Nowadays it seems the config directory is ~/PCSX2 so I've pushed changes to reflect this. Seems to work OK now.
> > Memory card contains network config
> 
> I wasn't able to use the memory card file you provided (it showed up as 7MB, but did not show up in the BIOS browser menu after it was inserted),
> 
Not entirely sure why this is the case. Then again, the memory card file is a convenience that probably doesn't serve anyone except me. Might be worth removing altogether, as long as someone can verify that they can setup network onto their own memcards


> > However, layer 2 MAC addresses were presumably tied to what the network adapter reported to the BIOS

> Correct, this information is retrieved from eeprom and was updated in pcap_io_init(), however, you seem to have removed that code.
I would like to restore that logic. Will revisit down the road

lowlyocean:
Can someone please help move this forward?

arcum42:
It's also possible for pcsx2 to be running in portable mode, in which case the ini and plugin folders and such are directly in the bin directory. In cmake, there's a variable named "PLUGINS_DIRECTORY" that points to the plugins directory. I didn't see one of the ini directory, but presumably it's in the same folder as the PLUGINS directory.

I'll also note that if you happen to do sudo sh build.sh instead of sh build.sh, the way you've got it written, no one will be able to run pcsx2 except root. You probably *shouldn't* be doing that, but it could happen...

It did create the two directories in question and seemed to read out of them, showed up in the list, and had all the network devices available to configure, and let me pick out my ethernet port, pcap bridge: enp3s0. 

On then trying to run an iso, it gave me the following:
"Opening adapter 'enp3s0'...enp3s0: You don't have permission to capture on that device (socket: Operation not permitted)
Unable to open the adapter. enp3s0 is not supported by WinPcap"
Then crashed a moment later with:
"(PCSX2:2527): GLib-CRITICAL **: 20:40:53.221: Source ID 2137 was not found when attempting to remove it"

Though, oddly, which the first error came up every time, the second crash didn't always happen.

Just to note, I'm not sure if I actually have any games that use the network side of things. I'm just trying to make sure it compiles and doesn't crash.

--arcum42

TheLastRar:
I'll add my test findings

The plugin will crash is the configuration file is not found, when trying to enter the plugin's config menu (showing a popup that can't be closed), or when starting emulation (just crashes), Ideally we should be creating a new config file in such a situation.

The config file are loaded from `~/PCSX2/inis`, while it should be loading from `~/.config/PCSX2/inis` the correct directory should be passed to the plugin via a call `DEV9setSettingsDir()`. I'm unsure how the .ui file should be handled.

It fails to open the selected adapter for me, giving the same error message.
It then crashes with a similar GLib-CRITICAL message, although sometimes a popup box is shown before the crash.

On windows the plugin will show a popup informing you if it can't open the selected adapter, presumably your linux port is trying to do the same thing and runs into issues.

plugins/dev9ghzdrk/pcap_io.cpp
```diff
@@ -43,427 +48,340 @@ char errbuf[PCAP_ERRBUF_SIZE];
...
+                  65536,			// portion of the packet to capture.
+                  // 65536 grants that the whole packet will be captured on all the MACs.
+                  //                                  pcap_mode==switched?1:0,				// promiscuous mode (nonzero means promiscuous)
+                  1, // promiscuous for Xlink usage
+                  1,			// read timeout
+                  errbuf			// error buffer
+                  )) == NULL)
+    {
+        fprintf(stderr, errbuf);
+    fprintf(stderr,"\nUnable to open the adapter. %s is not supported by WinPcap\n", adapter);
```
TheLastRar: This probably should be changed to reference libpcap (or just pcap) on Linux.

TheLastRar:
Shows nothing for me when I run it on the plugin file.

When building I don't see a message that states "Enabling networking capability on Linux..." but I do see the message about copying the MC and ui files.

Also, something I forgot to mention in my last post, The config panel still lists bridged mode adapters, if you are removing bridged mode you should also remove these entries.

lowlyocean:
Agreed, the config panel should not list switched or loopback. I'll add the
filter.

I'm not sure why you don't see "Enabling networking capability on Linux..."
during the install, but it would certainly explain the error opening the
adapter.

lowlyocean:
I've pushed changes:

- Use Dev9setSettingsDir() to determine where to write load & save cfg file
- Install ui file to same directory as PCSX2 executable, no need to provide directory
- Install should set proper networking capabilities for pcap to open adapter even if PACKAGE_MODE=False
- Remove "switch" entries from the dropdown menu in the config dialog
- Add include directories for LibXML2 and LibPcap in CMakelists.txt... on some systems the headers aren't implicitly available

Could you please try now? Thanks for continued feedback

arcum42:
np. I figure I might not be actually using the plugin, and I'm not sure how far I can test it, but I can at least try to get it to the point where it installs and doesn't crash or give me error messages. ^_^

This time I basically wiped the directory and redownloaded from git. It copied the .ui file into the bin directory. The .cfg file didn't seem like it copied anywhere, and since it crashes without it, I had to copy that myself. Having the plugin write that file if it isn't there instead of crashing would be nice, since it looks pretty minimal. (I'm also curious if that ui file works in gtk3, but I just compiled using gtk2 this time.)

On the good side, while I had to futz around copying things to get it not to crash, this time running an iso (Dark Cloud) didn't crash (and I opened it 2-3 times), and presumably it initialised correctly. I saw the following (ignore the ALSA messages).

```
Opening adapter 'enp3s0'...ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
ALSA lib pcm.c:8306:(snd_pcm_recover) underrun occurred
Device uses DLT 1: EN10MB
Ok.
```

So we definitely have some progress, anyways.

--arcum42

lowlyocean:
I had meant to add a check for existence of the cfg file before trying to load it. I've added the check now, so that if there is an existing config it's loaded.

TheLastRar:
I thought we where keeping switch and removing bridge?

The plugin currently looks for the .ui file in the working directory
When using the PCSX2 Daily Builds PPA, this seems to be the home directory (and not where PCSX2 executable is located) I'm assuming this is unintentional.

The error message is now miss-leading on windows.

For some reason, the plugin crashes when used with the PCSX2 Daily Builds PPA (note, I've manually run the set networking caps so that shouldn't be the issue), When using a build of PCSX2 created from your branch the plugin works fine, unsure what is going on there.

The plugin doesn't seem to be able to read the settings it has saved, closing and reopening the emulator appears to reset the settings.

If I open, close, then reopen the dev9 config panel, I get duplicate Expansion and PC Card entries in the Type dropdown box

I'm able to connect Xlink Kai, I was also able to connect to both the RE:Outbreak resurrection server or the .Hack//fragment Netslum server (after some difficulty), so it seems to be working correctly.

---------------------------------------

I've also tested this on windows. I was aiming to check if the filtering works on the windows side, however, I ran into an issue specific to your branch,

It builds correctly but will crash PCSX2 when it searches for plugins. I'm not entirely sure what change you did that is causing the issue.

Attaching a debugger on a debug build gives me the following message
>Run-Time Check Failure #0 - The value of ESP was not properly saved across a function call.  This is usually a result of calling a function declared with one calling convention with a function pointer declared with a different calling convention.

From here; https://github.com/PCSX2/pcsx2/blob/3fe9ff7ac835c2dbdf696cc4aaa476edab549f57/pcsx2/gui/Panels/PluginSelectorPanel.cpp#L117

lowlyocean:
> this seems to be the home directory (and
 > not where PCSX2 executable is located) 
> I'm assuming this is unintentional. 
>
I can probably re-use the cfg directory to know where to read UI at runtime but that info isn't available at compile time, when installing the file, is it? I am trying my best to integrate with the project but could use a hand with details like variables specifying directory structures. Don't want to waste time with the guess-and-check approach. Please feel free to commit if you see easily fixable directory-related issues.

I will just make the message say pcap instead of WinPCap or libpcap.

The failure to load on Windows must be due to the extern "C" EXPORT that I placed in front of the functions that need to be visible from the main executable. I added an #ifdef to mimic the normal Windows mechanism or Linux mechanism. I'll need to fix that for Windows case. Apparently there's a need to change calling convention.

I know nothing about that Daily Build PPA or how it modifies the build environment. Why does the PPA not build the way a developer does? The results should be identical, except perhaps recent commits from master that have not been pulled into the branch. It would be a freak coincidence if the plugin API changed since this pull request was created, but it's possible.

Is failure to save plugin cfg related to file permissions? Do you know if the cfg file is being written? What do the contents look like when you see doubles?

turtleli:
Would you mind if I removed the whitespace changes? It'd make it easier to review the code online.

plugins/dev9ghzdrk/PS2Edefs.h
```diff
@@ -48,12 +48,25 @@
...
-#define CALLBACK
-#else
+#ifndef __linux__
 #include <windows.h>
 #endif
 
+#if defined(_MSC_VER)
+    //  Microsoft
+    #define EXPORT __declspec(dllexport)
+    #define IMPORT __declspec(dllimport)
```
turtleli: I don't think `IMPORT` is used anywhere?

plugins/dev9ghzdrk/PS2Edefs.h
```diff
@@ -48,12 +48,25 @@
...
+#elif defined(__GNUC__)
+    //  GCC
+    #define EXPORT __attribute__((visibility("default")))
```
turtleli: Probably should be `__attribute__((stdcall, externally_visible, visibility("default")))`. The plugin interface uses the stdcall calling convention.

It might be nice to include `extern "C"` as well in EXPORT.

plugins/dev9ghzdrk/Linux/Linux.cpp
```diff
@@ -0,0 +1,149 @@
...
+    OnOk();
+    break;
+    case -6: //IDCANCEL
+    break;
+    }
+    gtk_widget_hide (GTK_WIDGET(dlg));
+
+}
+
+extern "C" EXPORT void DEV9about() {
```
turtleli: An about box isn't really necessary. It won't be accessible from PCSX2 anyway.

plugins/dev9ghzdrk/Win32/ata.cpp
```diff
@@ -2957,7 +2957,7 @@ void ata_term()
 }
 #define LOG_IRS (sz!=2)
 template<int sz>
-u8 CALLBACK ata_read(u32 addr)
+extern "C" EXPORT u8 ata_read(u32 addr)
```
turtleli: I don't think EXPORT needs to be added to this. Only the DEV9* and PS2E* functions really need to be visible. The CALLBACK was probably unnecessary in the first place.

Also, this file isn't even compiled on Windows.

plugins/dev9ghzdrk/PS2Edefs.h
lowlyocean: You are correct, it's defined but not necessary

plugins/dev9ghzdrk/Win32/Win32.cpp
```diff
@@ -111,7 +111,7 @@ void OnOk(HWND hW) {
 	EndDialog(hW, TRUE);
 }
 
-BOOL CALLBACK ConfigureDlgProc(HWND hW, UINT uMsg, WPARAM wParam, LPARAM lParam) {
+extern "C" EXPORT BOOL ConfigureDlgProc(HWND hW, UINT uMsg, WPARAM wParam, LPARAM lParam) {
```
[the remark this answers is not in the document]
lowlyocean: I used a regex search/replace and caught this when apparently it isn't necessary, and may result in issues.

plugins/dev9ghzdrk/Win32/ata.cpp
lowlyocean: Should we remove this entire file? What was it ever used for?

plugins/dev9ghzdrk/pcap_io.cpp
```diff
@@ -18,23 +18,28 @@
 #include "pcap.h"
 #include "pcap_io.h"
 
-#include "dev9.h"
+#include "DEV9.h"
 #include "net.h"
 
+#ifdef _Win32
```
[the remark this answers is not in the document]
lowlyocean: thanks

plugins/dev9ghzdrk/pcap_io.cpp
```diff
@@ -75,53 +80,74 @@ int GetMACAddress(char *adapter, mac_address* addr)
...
-		pcap_mode==switched?1:0,				// promiscuous mode (nonzero means promiscuous)
+		//pcap_mode==switched?1:0,				// promiscuous mode (nonzero means promiscuous)
+							 1,				// promiscuous for Xlink usage
 							 1,			// read timeout
 							 errbuf			// error buffer
 							 )) == NULL)
 	{
 		fprintf(stderr, errbuf);
-		fprintf(stderr,"\nUnable to open the adapter. %s is not supported by WinPcap\n", adapter);
+		fprintf(stderr,"\nUnable to open the adapter. %s is not supported by libpcap\n", adapter);
```
lowlyocean: change to pcap

plugins/dev9ghzdrk/DEV9.cpp
```diff
@@ -40,6 +48,26 @@ HINSTANCE hInst=NULL;
 
 //#define HDD_48BIT
 
+#if defined(__i386__) && !defined(_WIN32)
+
+static __inline__ unsigned long long GetTickCount(void)
+{
+	unsigned long long int x;
+	__asm__ volatile (".byte 0x0f, 0x31" : "=A" (x));
```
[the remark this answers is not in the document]
lowlyocean: Don't know. http://www.mcs.anl.gov/~kazutomo/rdtsc.html

plugins/dev9ghzdrk/PS2Edefs.h
turtleli: cdecl is caller clean and stdcall is callee clean, so if there's a mismatch the stack cleanup will either happen twice or not at all, and that's not going to work out well.

plugins/dev9ghzdrk/DEV9.cpp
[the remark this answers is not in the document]
turtleli: I'd guess it was compiler/assembler related. I do see a comment in GSdx that says
```
// gcc 4.8 define __rdtsc but unfortunately the compiler crash...
```

It's probably fine to use `__asm__ volatile ("rdtsc" : "=A" (x));` since that's used in at least two other places in the codebase.

plugins/dev9ghzdrk/Win32/ata.cpp
turtleli: Dunno. I've barely looked at the file. I wasn't involved with PCSX2 at that point.

plugins/dev9ghzdrk/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+)
+
+set(dev9ghzdrkFinalLibs
+    ${GTK2_LIBRARIES} ${PCAP_LIBRARY} ${LIBXML2_LIBRARIES}
+)
+
+add_pcsx2_plugin(${Output} "${dev9ghzdrkFinalSources}" "${dev9ghzdrkFinalLibs}" "${dev9ghzdrkFinalFlags}")
+
+if(PACKAGE_MODE)
+    install(FILES dev9ghzdrk.ui DESTINATION bin )
```
lowlyocean: Please correct the DESTINATION for PACKAGE_MODE and not

lowlyocean: Someone please set the DESTINATION for both lines here to a reasonable value

plugins/dev9ghzdrk/Linux/Linux.cpp
```diff
@@ -0,0 +1,142 @@
...
+    gtk_widget_hide (GTK_WIDGET(dlg));
+
+}
+
+void __attribute__((constructor)) DllMain() {
+
+    gtk_init (NULL, NULL);
+
+    builder = gtk_builder_new();
+    gtk_builder_add_from_file(builder, "dev9ghzdrk.ui", NULL);
```
lowlyocean: And, if necessary, perform a Copy from whatever that reasonable DESTINATION directory was to the directory stored in s_strIniPath

plugins/dev9ghzdrk/Win32/ata.cpp
arcum42: Looks like hard drive code adapted from QEMU. The plugin probably was trying to do hd emulation as well as networking at some point.

lowlyocean: https://github.com/jackun/USBqemu-wheel seems to have that abiliy (as well as Linux logitech headset capability I needed for SOCOM II online) . Would love to see that integrated at some point

TheLastRar:
> I can probably re-use the cfg directory to know where to read UI at runtime but that info isn't available at compile time, when installing the file, is it?

I'm unfamiliar with PCSX2's build system on Linux so I would not know

> The failure to load on Windows must be due to the extern "C" EXPORT that I placed in front of the functions that need to be visible from the main executable. I added an #ifdef to mimic the normal Windows mechanism or Linux mechanism. I'll need to fix that for Windows case. Apparently there's a need to change calling convention. Or maybe I just need to use _Win32 on line 55 of PS2Edefs.h. Or, remove lines 57 and 58 since I just realized there's a .def file

Looking though the other plugins, they use the define 'EXPORT_C_(type)' (which is defined in PS2Eext.h), can you change all functions that currently use EXPORT to use EXPORT_C_, see below;

https://github.com/PCSX2/pcsx2/blob/2af05a92f8046a5e9151bf014a9e0a9a8375032b/plugins/GSdx/PSX/GPU.cpp#L42

> Is failure to save plugin cfg related to file permissions? Do you know if the cfg file is being written? What do the contents look like when you see doubles?

The Config file was being written to, just seeming not read.

The double entries in the config window Type dropdown where

```
Expansion
PC Card
Expansion
PC Card
```

I've yet to test your latest commit, so I don't know if you've fixed the above already.

lowlyocean:
>they use the define 'EXPORT_C_(type)' 
>
Updated to reflect this

>double entries
>
I meant what the file looks like when this condition happens

plugins/dev9ghzdrk/DEV9.cpp
```diff
@@ -135,6 +172,7 @@ s32 CALLBACK DEV9init()
 
 	FLASHinit();
 
+#ifdef _WIN32
```
turtleli: This mapping part is rather confusing because of the repeated ifdefs. I suggest having just one ifdef cover it.

plugins/dev9ghzdrk/DEV9.cpp
```diff
@@ -65,18 +93,26 @@ static char *libraryName     = "GiGaHeRz's DEV9 Driver"
...
 
-u32 CALLBACK PS2EgetLibType() {
+EXPORT_C_(u32)
+PS2EgetLibType() {
 	return PS2E_LT_DEV9;
 }
 
-const char* CALLBACK PS2EgetLibName() {
+EXPORT_C_(const)
+char* PS2EgetLibName() {
```
turtleli: Compile error. Should be `EXPORT_C_(const char*)`, not `EXPORT_C_(const) char*`.

plugins/dev9ghzdrk/dev9ghzdrk.ui
```diff
@@ -0,0 +1,281 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<interface>
+  <requires lib="gtk+" version="2.24"/>
+  <!-- interface-naming-policy project-wide -->
+  <object class="GtkAboutDialog" id="IDD_ABOUT">
```
turtleli: I don't think this is needed anymore since you removed the about box code?

plugins/dev9ghzdrk/DEV9.cpp
lowlyocean: Corrected

plugins/dev9ghzdrk/dev9ghzdrk.ui
lowlyocean: Removed

plugins/dev9ghzdrk/DEV9.cpp
lowlyocean: updated

plugins/dev9ghzdrk/pcap_io.cpp
```diff
@@ -75,53 +80,76 @@ int GetMACAddress(char *adapter, mac_address* addr)
...
-	virtual_mac.bytes[2] = host_mac.bytes[2];
-	virtual_mac.bytes[3] = host_mac.bytes[2];
-	virtual_mac.bytes[4] = host_mac.bytes[5];
-	virtual_mac.bytes[5] = host_mac.bytes[4];
+	//Lets take the hosts last 2 bytes to make it unique on Xlink
+	virtual_mac.bytes[4] = host_mac.bytes[4];
+	virtual_mac.bytes[5] = host_mac.bytes[5];
+
+	for(int ii=0; ii<6; ii++)
+		eeprom[ii] = virtual_mac.bytes[ii];
```
lowlyocean: @TheLastRar could you please test ability to connect to server again? I've updated the bit to copy MAC into the BIOS so that there's no need for address translation during each send and receive (send and receive have been simplified accordingly)

TheLastRar:
> What do the contents look like when you see doubles?
I just noticed that the issue of doubles also occurs with the adapter dropdown as well.

the cfg file Looks like this;
```
<?xml version="1.0" encoding="UTF-8"?>
<dev9>
  <Eth>pcap bridge:enp0s3</Eth>
  <Hdd/>
  <HddSize>0</HddSize>
  <ethEnable>1</ethEnable>
  <hddEnable>0</hddEnable>
</dev9>
```

The Windows build no longer crashes and seems to work fine.
The Linux build didn't immediately crash the Daily build PPA during emulation, so whatever was causing this seems to be fixed.

lowlyocean:
>the cfg file Looks like this;
>
Interesting that you see doubles in the dropdowns, because the cfg appears fine.

lowlyocean:
> the cfg file Looks like this;
>
That file appears as expected. I have performed some cleanup to consolidate type of adapters supported (this means supporting PCAP only and removing TAP on Windows). Also cleaned up code in LoadConf() that iterates twice, once for bridged and once for switched.
```	modified:   plugins/dev9ghzdrk/Linux/Linux.cpp
	modified:   plugins/dev9ghzdrk/Win32/DEV9ghzdrk.vcxproj
	modified:   plugins/dev9ghzdrk/Win32/Win32.cpp
	deleted:    plugins/dev9ghzdrk/Win32/icmp.cpp
	deleted:    plugins/dev9ghzdrk/Win32/socket_io.cpp
	deleted:    plugins/dev9ghzdrk/Win32/tap-win32.cpp
	modified:   plugins/dev9ghzdrk/pcap_io.cpp
	modified:   plugins/dev9ghzdrk/pcap_io.h
	deleted:    plugins/dev9ghzdrk/tap.h
```

TheLastRar:
Don't remove TAP on windows

lowlyocean:
> Don't remove TAP on windows

reverted

lowlyocean:
Still seeing double @TheLastRar?

TheLastRar:
Still get the doubles after opening, closing, and opening the config menu.

Also, now I get this (before doubling);

monoclrdev9x86builder2 running - oracle vm virtualbox 24_09_2018 07_37_44 pm

Which seems to stack with the previous doubling issue.

monoclrdev9x86builder2 running - oracle vm virtualbox 24_09_2018 07_46_19 pm

lowlyocean:
Interface names should no longer concatenate and I'm seeing only the same list of interfaces when I close the config and reopen.

lowlyocean:
What's still needed to close this out and merge?

TheLastRar:
The doubling issue still occurs for me.
After some testing I observe that.
If you change the selected plugin in the plugin config panel and _don't_ hit apply apply before configuring the plugin, the issue will not occur.

If you change the selected plugin in the plugin config panel and _do_ hit apply before configuring the plugin, then the issue will occur if you close and reopen the config panel.

If you open the plugin config via Config->Dev9->Plugin Settings, closing and reopening will also cause the issue.

lowlyocean:
OnInitDialog() was getting called multiple times. The Windows implementation apparently relies on the fact that Windows sends an "init" signal to the dialog ONLY the first time it's displayed. I needed to either add state to remember if OnInitDialog's already been called, or find an equivalent signal mechanism with GTK+. There doesn't appear to be an equivalent "init" signal to install a callback for, so I've add a persistent boolean to OnInitDialog() that's working. Can you confirm on your end?

TheLastRar:
That has fixed the issue for me

lowlyocean:
@arcum42 wondering if you can help push this issue toward closure?

MojoJojoDojo:
Is it possible to do something similar to what the Dreamcast emulator Reicast did with modem multiplayer?
zero-configuration multiplayer will be so nice.

MrCK1:
I'd say that's out of the scope for this PR.

lowlyocean:
>zero-configuration multiplayer will be so nice.
>
@MojoJojoDojo Have you used Xlink Kai?

MojoJojoDojo:
I never tried to use it :S.

TheLastRar:
@MojoJojoDojo While I don't know specificity how Reicast's zero-configuration multiplayer works, for PCSX2 it hasn't been particularly easy to do, and is part of the reason why CLR_DEV9 exists.

I suppose logic could be added to auto select an adapter (i.e. pick the 1st adapter with a valid DNS address), CLR_DEV9 does have such code, but isn't used in its WinPcap backend.

A network profile is still required, which I believe is tied to specific BIOS versions. Your best bet is to create an Auto everything profile and hope for the best (it dosn't always work).

Side note: I would recommend using the network access disc rather than the utility built into most network games because of issue https://github.com/PCSX2/pcsx2/issues/287

MojoJojoDojo:
> @MojoJojoDojo While I don't know specificity how Reicast's zero-configuration multiplayer works, for PCSX2 it hasn't been particularly easy to do, and is part of the reason why CLR_DEV9 exists.
> 
> I suppose logic could be added to auto select an adapter (i.e. pick the 1st adapter with a valid DNS address), CLR_DEV9 does have such code, but isn't used in its WinPcap backend.
> 
> A network profile is still required, which I believe is tied to specific BIOS versions. Your best bet is to create an Auto everything profile and hope for the best (it dosn't always work).
> 
> Side note: I would recommend using the network access disc rather than the utility built into most network games because of issue #287
> 
> Edit: I agree with it being out of scope for this PR.

OK let me try it when i can.
i wasn't talking about the PR i was just talking in general.

arcum42:
I've replaced the motherboard on my test pc, and re-set it up as a 64-bit copy of Mint, and am now building in a 32-bit chroot, but running it in the 64 bit version (whereas before I was running it in a straight 32 bit environment). I'm still ironing out the issues on this at the moment, which is part of the delay.

At the moment, the plugin crashes with the following errors for me: 
```
"Opening adapter 'eno1'...eno1: You don't have permission to capture on that device (socket: Operation not permitted)
Unable to open the adapter. eno1 is not supported by pcap
[xcb] Unknown request in queue while dequeuing
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
PCSX2: ../../src/xcb_io.c:165: dequeue_pending_request: Assertion `!xcb_xlib_unknown_req_in_deq' failed."
```

It does, however, appear to run properly if I run it as root. I'm assuming building it in a chroot is complicating things. 

I'm also going to note that I get a lot of warnings about __forceinline__ being redefined when compiling in *other* plugins, due to the change to common/include/PS2ext.h.
```
In file included from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/plugins/PadNull/./Pad.h:23:0,
                 from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/plugins/PadNull/Linux/PadLinux.h:22,
                 from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/plugins/PadNull/Linux/PadLinux.cpp:19:
/home/arcum42/pcsx2-dev/lowlyocean/pcsx2/common/include/PS2Eext.h:50:0: warning: "__forceinline" redefined
 #define __forceinline __attribute__((always_inline))
 
In file included from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/common/include/PS2Edefs.h:61:0,
                 from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/plugins/PadNull/./Pad.h:22,
                 from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/plugins/PadNull/Linux/PadLinux.h:22,
                 from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/plugins/PadNull/Linux/PadLinux.cpp:19:
/home/arcum42/pcsx2-dev/lowlyocean/pcsx2/common/include/Pcsx2Defs.h:211:0: note: this is the location of the previous definition
 #define __forceinline __attribute__((always_inline, unused))
```

lowlyocean:
> I've replaced the motherboard on my test pc, and re-set it up as a 64-bit copy of Mint, and am now building in a 32-bit chroot, but running it in the 64 bit version (whereas before I was running it in a straight 32 bit environment). I'm still ironing out the issues on this at the moment, which is part of the delay.
> 
I'm not sure what issues chroot will introduce, if any
> At the moment, the plugin crashes with the following errors for me:
> 
> ```
> "Opening adapter 'eno1'...eno1: You don't have permission to capture on that device (socket: Operation not permitted)
When running cmake, you should be prompted for your password in order to run a ```sudo setcap``` command on the PCSX2 binary. Did that happen?
> Unable to open the adapter. eno1 is not supported by pcap
> [xcb] Unknown request in queue while dequeuing
> [xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
> [xcb] Aborting, sorry about that.
> PCSX2: ../../src/xcb_io.c:165: dequeue_pending_request: Assertion `!xcb_xlib_unknown_req_in_deq' failed."
> ```
> It does, however, appear to run properly if I run it as root. I'm assuming building it in a chroot is complicating things.
Running as root will overcome the missing net capabilities on the executable. Can you run ```getcap PCSX2``` and confirm that it's not showing "cap_net_admin,cap_net_raw+eip"?
> 
> I'm also going to note that I get a lot of warnings about **forceinline** being redefined when compiling in _other_ plugins, due to the change to common/include/PS2ext.h.
> 
> ```
> In file included from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/plugins/PadNull/./Pad.h:23:0,
>                  from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/plugins/PadNull/Linux/PadLinux.h:22,
>                  from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/plugins/PadNull/Linux/PadLinux.cpp:19:
> /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/common/include/PS2Eext.h:50:0: warning: "__forceinline" redefined
>  #define __forceinline __attribute__((always_inline))
>  
> In file included from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/common/include/PS2Edefs.h:61:0,
>                  from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/plugins/PadNull/./Pad.h:22,
>                  from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/plugins/PadNull/Linux/PadLinux.h:22,
>                  from /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/plugins/PadNull/Linux/PadLinux.cpp:19:
> /home/arcum42/pcsx2-dev/lowlyocean/pcsx2/common/include/Pcsx2Defs.h:211:0: note: this is the location of the previous definition
>  #define __forceinline __attribute__((always_inline, unused))
> ```
This can be solved with an #ifndef...#endif around both statements. But as you can see, they intend to do the same thing so only one should be kept. I'll look closer in a few

arcum42:
I remember it coming up prompting me for admin rights when I originally ran it, but getcap wasn't returning anything outside of the chroot and wasn't installed inside of it. I installed a few things and it was there but still not giving any results.

Bizarrely, though, I ran a clean compile, I saw it run that command (though I'd done sudo recently enough that I didn't have to enter my password), and then both inside and outside, getcap showed rights and it ran without crashing.

It would be nice to have the plugin gracefully abort if you don't have rights, rather than crashing, though.

I *suspect* the right answer on the the include problem is to delete the definition out of PS2ext.h, as that was added in your fork, and then include "Pcsx2Defs.h" in the plugin instead, if it isn't already.

lowlyocean:
> I remember it coming up prompting me for admin rights when I originally ran it, but getcap wasn't returning anything outside of the chroot and wasn't installed inside of it. I installed a few things and it was there but still not giving any results.
> 
> Bizarrely, though, I ran a clean compile, I saw it run that command (though I'd done sudo recently enough that I didn't have to enter my password), and then both inside and outside, getcap showed rights and it ran without crashing.
What few things did you install? Maybe the setcap utility was missing beforehand, so the capabilities were not applied during the first build
> 
> It would be nice to have the plugin gracefully abort if you don't have rights, rather than crashing, though.
Yes, I've now mimicked the error handling of the Win32 version with respect to failing to initialize
> 
> I _suspect_ the right answer on the the include problem is to delete the definition out of PS2ext.h, as that was added in your fork, and then include "Pcsx2Defs.h" in the plugin instead, if it isn't already.
This should be correct now. I restored common/include/PS2ext.h

arcum42:
I believe libcap2-bin was the one that actually took care of the executable being missing. And if that provides the command it was using to set it as well, that would make sense. I could easily have entered my password but not noticed an error immediately after that. With a chroot, it does sometimes have packages missing that would normally already be there...

Sounds good on the error handling, though once it's working, it's always hard to test on these things. 

Looks like those compile warnings are gone. There are a number of others in the individual plugin, but most are probably just a matter of casting to the right type. 

At this point, we should probably verify that it still works in Windows...

plugins/dev9ghzdrk/CMakeLists.txt
lowlyocean: still needs attention

plugins/dev9ghzdrk/Linux/Linux.cpp
lowlyocean: still needs attention

TheLastRar: Would it be possible to embed the ui file into the plugin? Either using objcopy or GResource

I found an article discussing using GResouce with gtk2 here https://metzlog.srcbox.net/2014/11/embedding-gtkbuilder-files-in-gtk2/

lowlyocean:
per @TheLastRar's suggestion I now embed the UI file into the library during build:
```
	modified:   cmake/SelectPcsx2Plugins.cmake
	new file:   cmake/macros/BuildTargetScript.cmake
	new file:   cmake/macros/CompileGResources.cmake
	new file:   cmake/macros/GenerateGXML.cmake
	new file:   cmake/macros/GlibCompileResourcesSupport.cmake
	modified:   plugins/dev9ghzdrk/CMakeLists.txt
	modified:   plugins/dev9ghzdrk/Linux/Linux.cpp

```

lowlyocean:
@arcum42 I was able to get into a game of SOCOM II with the Linux build. I just built on Windows, loaded the plugin, configured it, but couldn't verify network connectivity only because there seem to not be any game lobbies active right now. @TheLastRar I believe you'd said you've verified this on Windows?

plugins/dev9ghzdrk/pcap_io.cpp
```diff
@@ -75,53 +74,75 @@ int GetMACAddress(char *adapter, mac_address* addr)
...
+
 	//The checksum seems to be all the values of the mac added up in 16bit chunks
 	checksum = dev9.eeprom[0] + dev9.eeprom[1] + dev9.eeprom[2] & 0xffff;
```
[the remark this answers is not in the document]
lowlyocean: Actually, I'm not even sure that line is doing what the author intended it to do? It's using values from dev9.eeprom but I believe it's not until later that the eeprom[] arraly is populated and dev9.eeprom is assigned to it. So the checksum isn't operating on the right bytes... but it still works anyway? Idk would need to go over with a comb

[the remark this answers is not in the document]
TheLastRar: dev9.eeprom is set `DEV9init()` (which occurs before `pcap_io_init()` is called), the assignment is just casting a pointer, so both eeprom and dev9.eeprom refer to the same block of memory (as u8 and u16 respectively)

The exception to this is if a eeprom.dat file is loaded. The contents of eeprom and dev9.eeprom would then differ as the file handle is only assigned to dev9.eeprom.

https://github.com/PCSX2/pcsx2/blob/2af05a92f8046a5e9151bf014a9e0a9a8375032b/plugins/dev9ghzdrk/DEV9.cpp#L148-L151

lowlyocean:
Confirmed working on Windows
capture

arcum42:
Looking good. We're down to two compiler warnings I'm not really worried about (and one is just because a compiler option is c++ only and is being run on a .c file).

My only concern at this point, really, is whether that setpcap during compiling and having to enter a root password will cause any issues, for packaging for distributions, for example. I suppose if people haven't complained yet about it, they can complain after it's merged, though.

lightningterror:
> We're down to two compiler warnings 

Master branch had more warnings before this was merged ? If so that's good.

arcum42:
Well, we're down to two compilation warnings *from this plugin*. There were more than a dozen. This plugin didn't compile on linux at all before this was merged, so it's still two more than there were before. One or two is reasonable, though. Pcsx2 itself gives far more warnings. (Sio.cpp has quite a number, in fact, IIRC.)

The warnings from this plugin were that a variable (which is used in the next line) isn't used, and that a compiler flag is being ignored on one file because it's a .c file, and the flag is for .cpp files. Nothing too major.

RedPanda4552:
Some users have been reporting a message box saying "Can't open eth0" when launching PCSX2. I don't have more details to offer since I have not experienced this myself (to this day the DEV9ghzdrk plugin has been broken for me and never selectable in the plugin manager), but I want to put it in writing just so it's out there.

lowlyocean:
We need to know what those users get when they run ```sudo getcap $(which PCSX2)```

If they're just builing and not getting to the ```make install``` part of the build then this error won't be surprising.

mirh:
I am on windows.

lowlyocean:
@mirh you get a message about eth0 on windows?

RedPanda4552:
The other user I heard from was also a Windows user.

lowlyocean:
@mirh If you're getting a message about eth0 on Windows then we need information about your build environment. Also please try to build https://github.com/PCSX2/pcsx2/commit/67d8c8f8b2c9b0a948fdd9550743be8064c55092

mirh:
I'm not sure what my "build" environment would have to be. 
Windows 7.. winpcap is installed (but I didn't even try to use this new feature).. then what?

And that commit didn't seem to fix the problem.

lowlyocean:
I meant you're using Visual Studio? As opposed to MinGW or Cygwin or WSL. 

I don't understand where adapter named "eth0" could be coming from in a Windows environment. The only thing I can think of is if you are building in Cygwin or WSL. Do you see where the eth0 is being sourced from?

RedPanda4552:
The other user that was getting it was using one of the precompiled Windows builds from the build bot. Not building on their own. It therefore would have been with VS 2015.

mirh:
Yes, vs2015. 
And I really **only** downloaded last build from orphis bot, and started it.

lowlyocean:
Can you answer the questions above? I cannot recreate the error on my end so I need information to do anything about it. Were you using the networking plugin before?

mirh:
**No at all**. 
I'm not sure how further I can underline it was default pcsx2. 
The only thing I touched was literally setting sse4 gsdx.

lowlyocean:
So you are not using the networking feature but seeing that error when you launch the game. That wasn't made clear before. I know it sounds like you're saying something simple but the details matter. I'll look at the code but without more information about what you have installed or configured I can only help so much

willkuer:
Well when opening the plugin selector all dlls in the ini are loaded at least once (that’s why we get those ‘some kinda plugin error’ for higher gsdx instruction sets when the supported set is lower. I assume during load it is just checked what kind of plugin this dll generates to put it into the appropriate section.

So you could look into the plugin loader from pcsx2 core to check what happens upon loading all plugins from the plugins folder. However dll loading is a nasty thing (and I am not sure how it behaves in the unmanaged world) but typically all linked assemblies and their dependencies are loaded as well (even if they are not explicitly referenced during the load process). So your code might not even be the issue but it could be. And the issue could basically occur everywhere. So I guess there is no easy way around than asking mirh to attach a debugger.

You could possibly run dependency walker on the dll to check whether that one shows some linking errors.

mirh:
```
EE Core	DEV9ghzdrk (2).dll!SysMessage	
 	user32.dll!_NtUserWaitMessage@0()	 
 	user32.dll!_DialogBox2@16()	 
 	user32.dll!_InternalDialogBox@24()	 
 	 user32.dll!_SoftModalMessageBox@4()	 
 	 user32.dll!_MessageBoxWorker@4()	 
 	user32.dll!_MessageBoxTimeoutW@24()	 
 	user32.dll!_MessageBoxTimeoutA@24()	 
 	user32.dll!_MessageBoxExA@20()	 
 	user32.dll!_MessageBoxA@16()	 
 	 DEV9ghzdrk (2).dll!SysMessage(const char * fmt, ...) line 313	 
 	DEV9ghzdrk (2).dll!GetNetAdapter() line 178	 
 	DEV9ghzdrk (2).dll!DEV9open(void * pDsp) line 266	 
 	pcsx2.exe!014d62f5()
```

lowlyocean:
@mirh So after a bit of digging, it turns out @TheLastRar added the logic to copy "eth0" into config.eth by default in commit dcd477f2 , which explains the mystery of how on Windows you get a message about a Linux adapter name. It looks like the changes to Win32/Config.ccp that he made circa 2015 were attempting to use INI files for storing settings instead of the registry.

I know that before that commit 5f068ec4, this plugin was working without error on Windows. I know this because I started developing the Linux version before then.

I was considering just backing out Win32/Config.cpp to commit 86895cbf to get it working again ASAP. Instead I've just made it the Windows equivalent of Linux/Config.cpp

mirh:
Yes, fixed. Thank you. 
(I'm not sure what automagically made my network devices start to appear in the plugin drop down menu, even with the original dev9)

TheLastRar:
I believe that eth0 was the default setting before, although I can't remember at this point.
The expectation was that the user would configure the plugin before tying to use it. admittedly it isn't the most sane default for windows.

While annoying, you can dismiss the popup and emulation should continue, was it crashing for you?  @mirh

That being said, making it not be pop up if you've specificity disable Ethernet emulation is a good idea.

The plugin will only be able to detect your network devices if the winpcap service is running (by default it should be), you may also need to restart for winpcap pickup any changes you made to them (i.e. installing drivers, enabling/disabling adapters). That might be what your issue was.

mirh:
No it was just a popup. And there's no problem now for people that don't care. 
And the other thing yes, I just found out it happens when "NetGroup Packet Filter Driver" (not service, ftr) is not running. 
That should be taken into account or something I guess, if you are interested.

lowlyocean:
The only thing I would do for that case is pop up "have you enabled NetGroup Packet Filter Driver" or something similar in the plugin configuration dialog when there are no devices detected

mirh:
I mean.. it shouldn't be rocket science to distinguish between "driver on, no devices" and "driver off, end"

orbea:
@lowlyocean Not all users have `bash` or `sudo`. Can this be fixed?

orbea:
Also, it seems to be physically changes `/usr/bin/pcsx2` and not `$DESTDIR/usr/bin/pcsx2`....

prafullpcsx2:
@lowlyocean @TheLastRar @arcum42 

Sorry to say but we need both switch and bridge mode to get games properly working online. From my tests LAN mode games work properly only with bridge mode while games with actual Online mode work with Switch mode. I request you to please fix this and bring back option to choose between bridge and switch mode. Also CLR Dev9 isn't compatible enough yet to completely replace dev9ghz so this definitely needs to be fixed.

TheLastRar:
Care to elaborate what issues LAN games encountered with switched mode that bridged mode resolves? 

Regarding CLR DEV9, I assume you are referring to the forum post you made yesterday (I don't check the thread as often as I should) I think I know the cause of that and should have a fix soon(tm)

lowlyocean:
> games with actual Online mode work with Switch mode.

Does this mean there are games that have original servers still running? I was under the impression that most PS2 games are played "online" nowadays are really just using LAN mode over a VPN.

Can someone explain why Switched mode is actually needed? When using bridged mode, the PS2 BIOS thinks your actual PC's network hardware is its own. It just calls it by a different MAC address than your PC does. Everything that happens at Layer 2 on your actual PC's hardware can be seen by the PS2 BIOS if you use bridged mode. This allows Link Layer broadcasting that's used for discovery in LAN. The BIOS/game performs any Layer 3 / IP configuration natively. If you use switched mode, you introduce another level of Layer 3 management by an external tool (the plugin), you lose ability to handle MAC broadcasts for LAN, and it's not clear what benefit you can get.

prafullpcsx2:
@lowlyocean There are private servers running for most of the games now and online play has been revived. See this list. 
https://docs.google.com/spreadsheets/d/1bbxOGm4dPxZ4Vbzyu3XxBnZmuPx3Ue-cPqBeTxtnvkQ/edit

@TheLastRar The main issue happens when trying to join a game. The games can be discovered/created but you are unable to join. 

Believe me I have thoroughly tested this, both modes are needed.

Also if someone is trying to fix clrdev9 then try lan mode in Splinter Cell games.

lowlyocean:
> Believe me I have thoroughly tested this, both modes are needed.

Would be helpful if you can provide details of the testing you've done because it's still not clear *why* switched mode is needed.

prafullpcsx2:
Tested this again today using Socom 2 game lan mode. The created game is never visible using the new plugin (which has only one option after this commit). It works perfectly fine if using bridged mode in old plugin.

lowlyocean:
This plugin uses bridged mode. Switched mode is considered obsolete and was removed to simplify the code. We'd need to capture traffic of the old plugin in Switched mode using Wireshark to understand why those "online" games work with it alone. Bridged mode works with Socom 2 LAN in the new plugin over XLink. Are you using XLink to host/join LAN games?

prafullpcsx2:
No, I am not using Xlink. I am using real lan network at my home. Also used hamachi. Infact even used two pcsx2 running on same pc. They all work with old plugin bridge mode. As I said earlier this plugin has problem in LAN mode. I have been playing PCSX2 online/lan since the first day this plugin was created and I know what I am about talking about. Even @TheLastRar  was confused earlier whether you deleted bridge mode or switch mode. I think some code of bridge mode has definitely been compromised here.

lowlyocean:
`I think some code of bridge mode has definitely been compromised here.`
This is possible, of course - but again, if you can provide a packet capture we can investigate it would help. Otherwise, it's unclear where the bug is.

PR #2589, opened 2018-09-12
GSdx-gui: Minor osd gui adjustments.

lightningterror:
Grey out scrolling speed and maximum onscreen messages when
Enable Log is not checked.


lightningterror:
Note to self > some tooltips are missing for the buttons so maybe add them one day.

PR #2590, opened 2018-09-12
GSdx: GPU accelerate 8 bits texture conversion port from OpenGL to Direct3D11

lightningterror:
Commit:
https://github.com/PCSX2/pcsx2/commit/d29e375f72c94fb206812e589cf6b8c5814249f9

Only native res is supported currently, but it's still great progress.
Someone needs to port the ScalingFactor to D3D from commit:
https://github.com/PCSX2/pcsx2/commit/6121677aa1064269db333f4d5db0b3ead65270c1

Credits to @KrossX for porting the shader.

### DIFF
--- a/plugins/GSdx/res/convert.fx
+++ b/plugins/GSdx/res/convert.fx
@@ -3,7 +3,7 @@
 
 struct VS_INPUT
 {
-	float4 p : POSITION; 
+	float4 p : POSITION;
 	float2 t : TEXCOORD0;
 };
 
@@ -36,7 +36,7 @@ struct PS_OUTPUT
 
 struct VS_INPUT
 {
-	float4 p : POSITION; 
+	float4 p : POSITION;
 	float2 t : TEXCOORD0;
 };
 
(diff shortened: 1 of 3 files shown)


lightningterror:
Note for self > maybe this can be added after
https://github.com/PCSX2/pcsx2/blob/2af05a92f8046a5e9151bf014a9e0a9a8375032b/plugins/GSdx/GSDevice11.cpp#L189

```	// Scalingfactor
	ConvertConstantBuffer cb;

	cb.ScalingFactor = GSVector4i(theApp.GetConfigI("upscale_multiplier"));

	memset(&bd, 0, sizeof(bd));

	bd.ByteWidth = sizeof(ConvertConstantBuffer);
	bd.Usage = D3D11_USAGE_DEFAULT;
	bd.BindFlags = D3D11_BIND_CONSTANT_BUFFER;

	hr = m_dev->CreateBuffer(&bd, NULL, &m_convert.cb);
	// End Scalingfactor```

File: plugins/GSdx/res/convert.fx [hunk 1]
```diff
@@ -219,6 +219,162 @@ PS_OUTPUT ps_main9(PS_INPUT input) // triangular
...
+	// Input column: 8x2 RGBA pixels
+	// 0: 8 RGBA
+	// 1: 8 RGBA
+	// Output column: 16x4 Index pixels
+	// 0: 8 R | 8 B
+	// 1: 8 R | 8 B
+	// 2: 8 G | 8 A
+	// 3: 8 G | 8 A
+	float c;
+
```
turtleli: Trailing whitespace here and a few other places too.

lightningterror:
@turtleli what about the dummy shaders ?  Are they fine like that ?

lightningterror:
Updated the dummy shaders, now they work properly with the help of refraction.

Anyway the results are quite good, there's basically 0 speed penalty with the shader.
Blood Will Tell looks awesome on native.

beforeafter_8bitshader_2

I'll be merging this since it works properly.

PR #2592, opened 2018-09-14
GSdx: TC offset hack related changes

### LINKED ISSUE #2568: GTK-GSDX: Userhack TC-Offsets issue
The issue was reported on discord a few days ago so I'll post it here before all of us forget.
Version v1.5.0-dev-2512 as a reference

The way TC Offsets works on Windows is that it uses bitmasking to store two values in one hexadecimal value. Explained by @RedPanda4552 

Windows:
```
	SendMessage(GetDlgItem(m_hWnd, IDC_TCOFFSETX), UDM_SETRANGE, 0, MAKELPARAM(10000, 0));
	SendMessage(GetDlgItem(m_hWnd, IDC_TCOFFSETX), UDM_SETPOS, 0, MAKELPARAM(theApp.GetConfigI("UserHacks_TCOffset") & 0xFFFF, 0));

	SendMessage(GetDlgItem(m_hWnd, IDC_TCOFFSETY), UDM_SETRANGE, 0, MAKELPARAM(10000, 0));
	SendMessage(GetDlgItem(m_hWnd, IDC_TCOFFSETY), UDM_SETPOS, 0, MAKELPARAM((theApp.GetConfigI("UserHacks_TCOffset") >> 16) & 0xFFFF, 0));
```

On linux it doesn't do that so you end up not knowing what the hexadecimal value is so you don't know what to put there and you end up with a not properly working hack.

Example: TC offsets 350 X 350 Y => UserHacks_TCOffset=22937950

turtleli:
Changes:
 * Move TC offset variables from GSRendererOGL and GSRendererDX to GSRendererHW to avoid some code duplication. Also renames the variables for better consistency.
 * Splits TC offset ini setting into X and Y components - it simplifies loading/saving the TC offset value and also makes it easier to handle in the Linux GSdx GUI (which has been adjusted to use two spin controls instead of one text box). Fixes #2568.

### DIFF
--- a/plugins/GSdx/GSLinuxDialog.cpp
+++ b/plugins/GSdx/GSLinuxDialog.cpp
@@ -436,8 +436,9 @@ void populate_hack_table(GtkWidget* hack_table)
 	GtkWidget* hack_skipdraw_label = left_label("Skipdraw Range:");
 	CreateSkipdrawSpinButtons(0, 10000);
 	GtkWidget* hack_wild_check     = CreateCheckBox("Wild Arms Hack", "UserHacks_WildHack");
-	GtkWidget* hack_tco_label      = left_label("Texture Offset: 0x");
-	GtkWidget* hack_tco_entry      = CreateTextBox("UserHacks_TCOffset");
+	GtkWidget* hack_tco_label      = left_label("Texture Offset:");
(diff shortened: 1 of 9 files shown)


lightningterror:
Tested it just in case and it works properly on Windows.

PR #555, opened 2015-05-26
GSdx: better framebuffer format

gregory38:
It requires a lots of tests ! I'm really afraid of regression.

It improves Castlevania rendering (accurate_colclip = 1 & CRC disabled)

PR #2593, opened 2018-09-14
GSdx-d3d: Mask alpha channel on 24bit format

lightningterror:
Follow up/ish from #2465 maybe

Port from #555

dfmt is not supported but we can use the variable to select the
frame buffer format and later disable writting to the alpha channel.

MGS3 sees an improvement, and possibly other games as well.
So far didn't spot any regressions.

d3d-alpha24bit

I had to skip the channel shuffle draw calls for the image to show a proper comparison since d3d doesn't support it.

Anyway looks like d3d misses channel shuffle now to show a proper picture (minus blending).

### DIFF
--- a/plugins/GSdx/GSRendererDX11.cpp
+++ b/plugins/GSdx/GSRendererDX11.cpp
@@ -50,7 +50,7 @@ void GSRendererDX11::EmulateTextureShuffleAndFbmask()
 	// Note: D3D1011 is limited and can't read the current framebuffer so we can't have PS_FBMASK and PS_WRITE_RG shaders ported and working.
 	if (m_texture_shuffle) {
 		m_ps_sel.shuffle = 1;
-		m_ps_sel.fmt = 0;
+		m_ps_sel.dfmt = 0;
 
 		const GIFRegXYOFFSET& o = m_context->XYOFFSET;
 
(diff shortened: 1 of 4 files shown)


plugins/GSdx/GSRendererDX11.cpp
```diff
@@ -154,7 +154,7 @@ void GSRendererDX11::EmulateTextureShuffleAndFbmask()
 		}
 
 	} else {
-		//m_ps_sel.fmt = GSLocalMemory::m_psm[m_context->FRAME.PSM].fmt;
+		m_ps_sel.dfmt = GSLocalMemory::m_psm[m_context->FRAME.PSM].fmt;
```
ssakash: ``GSLocalMemory::m_psm[m_context->FRAME.PSM].fmt`` could return a potential 16 bit unsigned integer while dfmt is just 2 bits. would it be good to have some sort of checks or asserts here?

plugins/GSdx/GSRendererDX9.cpp
```diff
@@ -58,11 +58,17 @@ bool GSRendererDX9::CreateDevice(GSDevice* dev)
 
 void GSRendererDX9::EmulateTextureShuffleAndFbmask()
 {
-	if (m_texture_shuffle) {
+	if (m_texture_shuffle)
+	{
+		m_ps_sel.dfmt = 0;
```
turtleli: Is this needed if the draw is going to be skipped?

turtleli: So why do you think the alpha mask code might run?

lightningterror: Yeah was useless like you said so updated.

lightningterror:
I'll merge this since it's more accurate and it matches gl behavior.

Maybe the texture shuffle function call can be moved below merge sprite in the future if there is a need but not sure as of now.

PR #2596, opened 2018-09-15
pcsx2-gui: Adjust main menu tooltips

lightningterror:
Small corrections (periods).
Automatic Gamefixes, and No Disc tooltips have been changed to fit the
main window properly without clipping.
About  tooltip has been removed since it's pretty much useless.

### DIFF
--- a/pcsx2/gui/App.h
+++ b/pcsx2/gui/App.h
@@ -73,8 +73,8 @@ enum MenuIdentifiers
 	MenuId_Audio,				// audio options filled in by SPU2 plugin
 	MenuId_Misc,				// Misc options and help!
 
+	MenuId_About,
 	MenuId_Exit = wxID_EXIT,
-	MenuId_About = wxID_ABOUT,
 
 	MenuId_EndTopLevel = 20,
 
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -411,7 +411,7 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
 	m_menuSys.Append(MenuId_Boot_CDVD2,		_("Initializing..."));
 
 	m_menuSys.Append(MenuId_Boot_ELF,		_("&Run ELF..."),
-		_("For running raw PS2 binaries directly"));
+		_("For running raw PS2 binaries directly."));
 
 	m_menuSys.AppendSeparator();
 	m_menuSys.Append(MenuId_Sys_SuspendResume,	_("Initializing..."));
@@ -429,7 +429,7 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
 	m_menuSys.AppendSeparator();
 
 	m_menuSys.Append(MenuId_EnablePatches,	_("Automatic &Gamefixes"),
-		_("Automatically applies needed Gamefixes to known problematic games"), wxITEM_CHECK);
+		_("Automatically applies needed Gamefixes."), wxITEM_CHECK);
 
 	m_menuSys.Append(MenuId_EnableCheats,	_("Enable &Cheats"),
 		wxEmptyString, wxITEM_CHECK);
@@ -448,7 +448,7 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
 	m_menuSys.FindItem(MenuId_Sys_Shutdown)->Enable(false);
 
 	m_menuSys.Append(MenuId_Exit,			_("E&xit"),
-		AddAppName(_("Closing %s may be hazardous to your health")));
+		AddAppName(_("Closing %s may be hazardous to your health.")));
 
 
 	// ------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


pcsx2/gui/MainFrame.cpp
```diff
@@ -411,7 +411,7 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
 	m_menuSys.Append(MenuId_Boot_CDVD2,		_("Initializing..."));
 
 	m_menuSys.Append(MenuId_Boot_ELF,		_("&Run ELF..."),
-		_("For running raw PS2 binaries directly"));
+		_("For running raw PS2 binaries directly."));
```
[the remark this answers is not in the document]
lightningterror: Ah I didn't know that, I thought the translations looked up the lines only. Well I prefer to have periods because it looks a little bit more proper. I'll wait to see what the others have to say about it.

pcsx2/gui/MainFrame.cpp
```diff
@@ -448,7 +448,7 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
 	m_menuSys.FindItem(MenuId_Sys_Shutdown)->Enable(false);
 
 	m_menuSys.Append(MenuId_Exit,			_("E&xit"),
-		AddAppName(_("Closing %s may be hazardous to your health")));
+		AddAppName(_("Closing %s may be hazardous to your health.")));
```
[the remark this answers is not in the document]
lightningterror: Deleting it may be even more hazardous.

pcsx2/gui/MainFrame.cpp
```diff
@@ -429,7 +429,7 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
 	m_menuSys.AppendSeparator();
 
 	m_menuSys.Append(MenuId_EnablePatches,	_("Automatic &Gamefixes"),
-		_("Automatically applies needed Gamefixes to known problematic games"), wxITEM_CHECK);
+		_("Automatically applies needed Gamefixes."), wxITEM_CHECK);
```
turtleli: So, I just remembered that you can't see some of the help texts in the main frame. Is it helpful to make this string shorter if you can't see it anyway in the main window?

pgert: Furthermore, "Automatically applies needed Gamefixes." doesn't tell the whole truth.
If you incist on making it shorter, "Automatically applies Gamefixes." would be better.

lightningterror: At the very least it doesn't clip anymore even if the text is not visible from the menu, so it's still a very slight improvement.

pcsx2/gui/App.h
```diff
@@ -73,8 +73,8 @@ enum MenuIdentifiers
 	MenuId_Audio,				// audio options filled in by SPU2 plugin
 	MenuId_Misc,				// Misc options and help!
 
+	MenuId_About,
 	MenuId_Exit = wxID_EXIT,
-	MenuId_About = wxID_ABOUT,
```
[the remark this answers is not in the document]
lightningterror: Isn't it fine to remove the stock help string like I did ? Functionality wise should be the same.

pcsx2/gui/MainFrame.cpp
turtleli: Before -  Can't see anything.
After - Can't see anything.

Not sure you can call it an improvement.

lightningterror: I can purge it if you're fine with that.

pcsx2/gui/MainFrame.cpp
```diff
@@ -461,7 +461,7 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
 	m_menuCDVD.AppendSeparator();
 	m_menuCDVD.Append( MenuId_Src_Iso,		_("&ISO"),		_("Makes the specified ISO image the CDVD source."), wxITEM_RADIO );
 	m_menuCDVD.Append( MenuId_Src_Plugin,	_("&Plugin"),	_("Uses an external plugin as the CDVD source."), wxITEM_RADIO );
-	m_menuCDVD.Append( MenuId_Src_NoDisc,	_("&No disc"),	_("Use this to boot into your virtual PS2's BIOS configuration."), wxITEM_RADIO );
+	m_menuCDVD.Append( MenuId_Src_NoDisc,	_("&No disc"),	_("Boots into your virtual PS2's BIOS configuration."), wxITEM_RADIO );
```
turtleli: These are clipped on Linux. Also, the strings might actually be longer when it's translated into another language. So I don't think it's worth shortening the string.

pcsx2/gui/App.h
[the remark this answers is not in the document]
turtleli: The wxWidgets documentation kinda suggests to reuse the stock ids where possible.

In their hello world example (https://docs.wxwidgets.org/3.0.4/overview_helloworld.html):
> Notice that we don't need to specify the labels for the standard menu items wxID_ABOUT and wxID_EXIT, they will be given standard (even correctly translated) labels and also standard accelerators correct for the current platform making your program behaviour more native. For this reason you should prefer reusing the standard ids (see Stock Items) if possible.

Though we don't actually use the wx provided translations.

pcsx2/gui/MainFrame.cpp
turtleli: Well, problem is I don't know if they're useful or not.

lightningterror: I'll just close this then, nothing noteworthy to change.

PR #2597, opened 2018-09-16
GSdx-gui: Add debug configuration next to osd configuration only on debug mode

lightningterror:
Still needs a bit of work, will probably match the names with Linux gui.

image

Will leave this as a reference for future use so I have something to look at.
image


lightningterror:
I'm probably forgetting something, the spin buttons work in reverse and go in to a negative number.

turtleli:
So why are there 6 spin controls?

As for the spin controls not working properly, that's probably because the value you put in is too large to fit in a 16-bit integer. The documentation at https://docs.microsoft.com/en-us/windows/desktop/controls/bumper-up-down-control-reference-messages should point you to the solution for that.

lightningterror:
Cool , thanks. Didn't know that was the issue.

I didn't have a ss when I started making this, will probably drop those extra spin buttons.

lightningterror:
Updated. Should be better now.

image

lightningterror:
Ups, messed up the ss, anyway updated.

Re upload 
image

turtleli:
I think it'd be better if you use the same layout that is used on Linux.

I also think it'd be better if the spin control ranges are changed so they match what Linux uses.

lightningterror:
image

> I also think it'd be better if the spin control ranges are changed so they match what Linux uses.

How to do that ?

turtleli:
> How to do that ?

Did you look at the documentation link I provided above? I think it'd be better for your coding if you try and work out how to do it from that instead of having me tell you how to do it.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -1015,5 +1030,81 @@ bool GSOSDDlg::OnMessage(UINT message, WPARAM wParam, LPARAM lParam)
...
+	{
+	case WM_COMMAND:
+	{
+		int id = LOWORD(wParam);
+
+		switch (id)
+		{
+		case IDOK:
+		{
+			INT_PTR data;
```
turtleli: This is unused.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -1015,5 +1030,81 @@ bool GSOSDDlg::OnMessage(UINT message, WPARAM wParam, LPARAM lParam)
...
+			theApp.SetConfig("debug_opengl", (int)IsDlgButtonChecked(m_hWnd, IDC_GL_DEBUGLOG));
+			theApp.SetConfig("debug_glsl_shader", (int)IsDlgButtonChecked(m_hWnd, IDC_GL_DEBUGGLSL));
+			theApp.SetConfig("dump", (int)IsDlgButtonChecked(m_hWnd, IDC_DUMP_GS));
+			theApp.SetConfig("save", (int)IsDlgButtonChecked(m_hWnd, IDC_DUMP_SAVE));
+			theApp.SetConfig("savef", (int)IsDlgButtonChecked(m_hWnd, IDC_DUMP_SAVEF));
```
turtleli: Are the casts necessary?

plugins/GSdx/GSSettingsDlg.h
```diff
@@ -40,6 +40,18 @@ class GSShaderDlg : public GSDialog
 	GSShaderDlg();
 };
 
+class GSDebugDlg : public GSDialog
+{
+	void UpdateControls();
```
turtleli: I don't think this function is necessary if it does nothing.

plugins/GSdx/GSSettingsDlg.cpp
```diff
@@ -1015,5 +1030,81 @@ bool GSOSDDlg::OnMessage(UINT message, WPARAM wParam, LPARAM lParam)
...
+			theApp.SetConfig("savet", (int)IsDlgButtonChecked(m_hWnd, IDC_DUMP_SAVET));
+			theApp.SetConfig("savez", (int)IsDlgButtonChecked(m_hWnd, IDC_DUMP_SAVEZ));
+			theApp.SetConfig("savel", (int)(SendMessage(GetDlgItem(m_hWnd, IDC_DUMP_SAVEL), UDM_GETPOS, 0, 0)));
+			theApp.SetConfig("saven", (int)(SendMessage(GetDlgItem(m_hWnd, IDC_DUMP_SAVEN), UDM_GETPOS, 0, 0)));
+
+			EndDialog(m_hWnd, id);
+		}
```
turtleli: Missing break.

plugins/GSdx/GSSettingsDlg.cpp
lightningterror: I followed the other windows as an example.

ssakash: I'd prefer if redundant casts are removed.

lightningterror:
I'll finish this up another time, I spotted some other issues as well.

PR #2598, opened 2018-09-17
GSdx-d3d: tfx.fx correct texture shuffle shaders

lightningterror:
Some values were incorrect, now I'm no expert but looking at the code
and behavior seemed a bit wrong compared to gl.

Improves d3d rendering on Sonic Unleashed, the purple screen is gone
matching it with gl visuals.

Didn't spot any regressions on games that use texture shuffle that I tested.


lightningterror:
@KrossX confirmed that the original code was wrong so now it should be fine. Will merge this once the checks finish.

PR #2602, opened 2018-09-22
GSdx-d3d: Add macro to support ScalingFactor in shaders

lightningterror:


plugins/GSdx/GSDevice11.cpp
```diff
@@ -189,9 +189,19 @@ bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
 	theApp.LoadResource(IDR_CONVERT_FX, shader);
 	CompileShader(shader.data(), shader.size(), "convert.fx", nullptr, "vs_main", nullptr, &m_convert.vs, il_convert, countof(il_convert), &m_convert.il);
 
+	std::string convert_mstr[1];
```
ssakash: Does this have to be an array for some reason?

plugins/GSdx/GSDevice9.cpp
```diff
@@ -294,9 +294,19 @@ bool GSDevice9::Create(const std::shared_ptr<GSWnd> &wnd)
...
+	std::string convert_mstr[1];
+
+	convert_mstr[0] = format("%d", m_upscale_multiplier);
+
+	D3D_SHADER_MACRO convert_macro[] =
+	{
+		{"PS_SCALE_FACTOR", convert_mstr[0].c_str()},
+		{NULL, NULL},
+	};
+
```
turtleli: Trailing whitespace.

plugins/GSdx/GSDevice11.cpp
```diff
@@ -189,9 +189,19 @@ bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
...
+
+	convert_mstr[0] = format("%d", m_upscale_multiplier);
```
turtleli: Maybe use `std::to_string` instead?

plugins/GSdx/GSDevice11.cpp
KrossX: Doesn't have to, simply keeps it in line with the other macros.

plugins/GSdx/GSDevice11.cpp
KrossX: It's what's done on the other macros and I'd rather keep it the same.

ssakash: Personally I agree with Turtleli, just because someone used an inefficient way to do something before doesn't mean we still have to follow that inefficient approach in the name of consistency.

lightningterror:
Note to self -> remove dx9 stuff since ScalingFactor won't be used on dx9 and it will be a pointless change.

plugins/GSdx/GSDeviceDX.cpp
```diff
@@ -30,6 +30,8 @@ bool GSDeviceDX::s_old_d3d_compiler_dll;
 
 GSDeviceDX::GSDeviceDX()
 {
+	m_upscale_multiplier = theApp.GetConfigI("upscale_multiplier");
```
lightningterror: Maybe tie m_upscale_multiplier with nvidia hack so we use 1 check for both.

lightningterror:
So what are we gonna do with this ? I prefer to merge it since shader 17 needs it and for future depth support.

KrossX:
I like it the way it is, so I have no intention in making the suggested changes.

lightningterror:
I'm fine with the changes even if there is a bit better way to do it, at least this way we keep things consistent, and the code can be improved in the future in one go with all macros.

And I rather have this merged than have a half working feature.

PR #2605, opened 2018-09-24
GameDB: Add VU Clamp Mode and GIF FIFO Hack to SOCOM 1

MrCK1:
This PR resolves bad shadows and random sprite/HUD corruption in SOCOM 1.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -1480,6 +1480,8 @@ Region = PAL-M5
 Serial = SCES-50928
 Name   = SOCOM - U.S. Navy SEALs [Beta & Full Release]
 Region = PAL-M6
+vuClampMode = 2 // Fixes bad shadows.
+GIFFIFOHack = 1 // Resolves random sprite/HUD corruption.
 ---------------------------------------------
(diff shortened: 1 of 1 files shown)


ssakash:
What about the performance impact with GIF FIFO hack in the parts of the game? I'm pretty sure @refractionpcsx2 told me it'll be more demanding. (Just wondering if the performance-accuracy trade is worthy enough to leave it on automatically)

refractionpcsx2:
Probably more demanding, but if it needs it, it needs it

MrCK1:
Well the other option is to use the EE Timing Hack, but we can assume this option is at least a bit safer (and supposedly more correct?) :P

In this instance, I didn't see any speed penalties during my testing.

refractionpcsx2:
Yeah it's more correct. I guess it's going to depend on the performance hit to if the users would use one hack over the other. Is there issues in using the ee timing fix?

MrCK1:
I didn't do any thorough testing using the EE Timing hack but I can check later before we merge this.

lightningterror:
> CK: There's no perf difference with EE timing hack, you can merge it.

PR #2608, opened 2018-09-28
GSdx: Wrap negative UV on region repeat wrap mode.

lightningterror:
@KrossX showing GSdx some more love.

Makes UV coord be on the range [0, SIZE) before applying mask and
offset. This fixes Xenosaga's hair for example, which has negative UV
coords.

Xenosaga Example

Master HW:
image

PR:
image

SW:
image

Needs to be tested for any regressions.
@atomic83GitHub I leave the testing to you.


gregory38:
I know I'm a bit late. It would be nice a comment in the code.

Potentially the fract could be limited on `FST == 0` (float coordinate) mode. Because I don't know how we manage to get negative value with `FST == 1` (uint coordinate)

atomic83GitHub:
Sorry for the very late response, I was very much busy this time, anyway there is no regressions in games I tested, although it improve a bit some post process effect in Beyond good and evil which had the same black lines as Xenosaga.

PR #2613, opened 2018-10-02
GSdx: Limit Wrap negative UV when PS_FST is 0 only

lightningterror:
Follow up to https://github.com/PCSX2/pcsx2/pull/2608#issuecomment-426311409


KrossX:
Thanks! And yup, it still works.

PR #2619, opened 2018-10-05
Get host mac correctly on Windows

lowlyocean:
Follow on to #2586 which ensures that MAC address is unique on Xlink Kai when using Windows. Apparently the old GetAdapterInfo() API is deprecated after Windows XP. Previous build was returning last two bytes as 00:00 rather than last two bytes of host adapter's MAC as highlighted below
capture

### DIFF
--- a/plugins/dev9ghzdrk/Linux/dev9ghzdrk.ui
+++ b/plugins/dev9ghzdrk/Linux/dev9ghzdrk.ui
@@ -69,6 +69,7 @@
         </child>
         <child>
           <object class="GtkComboBoxText" id="IDC_BAYTYPE">
+         <property name="sensitive">False</property>
         <property name="visible">True</property>
         <property name="can_focus">False</property>
         <property name="has_entry">True</property>
@@ -129,7 +130,7 @@
               <object class="GtkComboBoxText" id="IDC_ETHDEV">
             <property name="visible">True</property>
             <property name="can_focus">False</property>
-            <property name="has_entry">True</property>
+            <property name="has_entry">False</property>
             <property name="entry_text_column">0</property>
               </object>
               <packing>
@@ -172,6 +173,7 @@
         <child>
           <object class="GtkCheckButton" id="IDC_HDDENABLED">
             <property name="label" translatable="yes">ENABLED</property>
+            <property name="sensitive">False</property>
             <property name="visible">True</property>
             <property name="can_focus">True</property>
             <property name="receives_default">False</property>
@@ -201,6 +203,7 @@
             </child>
             <child>
               <object class="GtkEntry" id="IDC_HDDFILE">
+            <property name="sensitive">False</property>
             <property name="visible">True</property>
             <property name="can_focus">False</property>
               </object>
@@ -223,7 +226,7 @@
           <object class="GtkLabel" id="label2">
         <property name="visible">True</property>
         <property name="can_focus">False</property>
-        <property name="label" translatable="yes">Hard Disk Drive (not yet properly implemented</property>
+        <property name="label" translatable="yes">Hard Disk Drive (not yet properly implemented)</property>
           </object>
         </child>
       </object>
--- a/plugins/dev9ghzdrk/Win32/Win32.cpp
+++ b/plugins/dev9ghzdrk/Win32/Win32.cpp
@@ -51,16 +51,15 @@ void OnInitDialog(HWND hW) {
 	for (int i=0; i<pcap_io_get_dev_num(); i++) {
 		dev = pcap_io_get_dev_desc(i);
 		int itm=ComboBox_AddString(GetDlgItem(hW, IDC_ETHDEV), dev);
-		ComboBox_SetItemData(GetDlgItem(hW, IDC_ETHDEV),itm,strdup(pcap_io_get_dev_name(i)));
+		ComboBox_SetItemData(GetDlgItem(hW, IDC_ETHDEV),itm,_strdup(pcap_io_get_dev_name(i)));
 		if (strcmp(pcap_io_get_dev_name(i), config.Eth) == 0) {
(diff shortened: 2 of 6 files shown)


lightningterror:
> Apparently the old GetAdapterInfo() API is deprecated after Windows XP

I believe there were also some warnings indicating that on windows side so this should fix some of them as well.

arcum42:
@lightningterror I believe "on Windows" is the key phrase here...

plugins/dev9ghzdrk/pcap_io.cpp
```diff
@@ -51,29 +51,43 @@ int GetMACAddress(char *adapter, mac_address* addr)
...
+		return 0;
+
+	pAdapterInfo = AdapterInfo;
+
+	char adapter_desc[128] = "\0";
+
+	// Must get friendly description from the cryptic adapter name
+	for (int ii = 0; ii < pcap_io_get_dev_num(); ii++)
+		if (0 == strcmp(pcap_io_get_dev_name(ii), adapter) )
+			strcpy(adapter_desc, pcap_io_get_dev_desc(ii));
```
turtleli: Shouldn't there be a `break;` after a successful match?

plugins/dev9ghzdrk/pcap_io.cpp
```diff
@@ -51,29 +51,43 @@ int GetMACAddress(char *adapter, mac_address* addr)
...
+
```
turtleli: Trailing whitespace.

plugins/dev9ghzdrk/pcap_io.cpp
```diff
@@ -51,29 +51,43 @@ int GetMACAddress(char *adapter, mac_address* addr)
...
+		NULL,
+		AdapterInfo,
+		&dwBufLen
+	);
+	if(dwStatus != ERROR_SUCCESS)
+		return 0;
+
+	pAdapterInfo = AdapterInfo;
+
+	char adapter_desc[128] = "\0";
```
turtleli: `""` should be null terminated.

plugins/dev9ghzdrk/pcap_io.cpp
lowlyocean: Optional here. There can only be one match because Windows doesn't allow duplicate descriptions

turtleli: It's more about making the intention of the code clearer. If someone reads this code they might be confused about why the search hasn't ended after a match has occurred.

turtleli:
> Mind amending the commit with your proposed changes?

Maybe later today if no one else has done it.

plugins/dev9ghzdrk/Linux/dev9ghzdrk.ui
```diff
@@ -69,6 +69,7 @@
         </child>
         <child>
           <object class="GtkComboBoxText" id="IDC_BAYTYPE">
+		 <property name="sensitive">False</property>
```
lightningterror: Tabs. The file uses normal spaces. Same goes for the other changes in this file.

plugins/dev9ghzdrk/Linux/dev9ghzdrk.ui
```diff
@@ -69,6 +69,7 @@
         </child>
         <child>
           <object class="GtkComboBoxText" id="IDC_BAYTYPE">
+         <property name="sensitive">False</property>
```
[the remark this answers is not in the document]
lowlyocean: This file was machine generated, I'm going to leave it as is with the exception of the lines I'm adding, unless there's strong objection

plugins/dev9ghzdrk/smap.cpp
```diff
@@ -416,10 +415,11 @@ smap_read16(u32 addr)
 		}
 		*/
 	}
-
+	else
+		return rv;
```
TheLastRar: This will prevent the logging code below from running even when DEV9_LOG_ENABLE is defined.

lowlyocean:
Anything left to do before merge? Thanks all for reviewing

lightningterror:
@TheLastRar , @turtleli  you fine with the changes ?

plugins/dev9ghzdrk/Win32/Win32.cpp
```diff
@@ -77,7 +77,7 @@ void OnOk(HWND hW) {
 	if (i == -1)
 	{
 		//adapter not selected
-		if (Button_GetCheck(GetDlgItem(hW, IDC_ETHENABLED)))
+		if ( config.ethEnable = Button_GetCheck(GetDlgItem(hW, IDC_ETHENABLED)))
```
TheLastRar: This if statement will only run if no adapter is selected, As this is now where the config value is saved, you won't be able to enable, or disable the pulgin's ethernet emulation if you've selected an adapter.

TheLastRar:
After a quick test, the mac address is correctly read from my adapter and set as required (checked with Kai and the Network access disc), The issue with saving the enable ethernet option was the only one I found.

I'm not concerned with the removal of reg import code, It's probably lived its purpose by this point.

plugins/dev9ghzdrk/Win32/Win32.cpp
lowlyocean: Good point. If you can squash the change in (revert back to two separate calls), great. If not I'll do it in a few hours when I'm able

lowlyocean:
Removed trailing whitespace. Moved assignment of ethEnable back where it was

lowlyocean:
A user in SOCOM 2's Xlink lobby last night is anticipating this merge because his mac appears to be changing each time he runs pcsx2 (no idea if he's using CLR-DEV9). Let's close this out

lowlyocean:
Yes continue with merge, thank you

lightningterror:
Just gave it a quick build and looks like some warnings have indeed been fixed.
Arcum also approves of it so I'm merging this.

PR #2630, opened 2018-10-09
GSdx: Remove ancient ATI point sampler hack for 8-bit textures

MrCK1:
Removes an ancient hack for ATI cards that caused texture alignment/minor warping issues in DX11 HW when 8-bit textures were enabled/used. This was first noticed in Shox, but likely affected other games as well. 

Before and after comparison for those who are curious. Notice the line in the bottom left corner of the ground texture.

ati_hack


MrCK1:
Apparently it's still needed for AMD cards.

PR #2631, opened 2018-10-10
PCSX2: GUI textstring corrections

pgert:


lightningterror:
I guess it's a little bit better.

willkuer:
Does this punctuation change invalidate all translations? Of this sentence?

pgert:
> 
> 
> Does this punctuation change invalidate all translations? Of this sentence?

Oops, yes it does.
Should this be reverted?

lightningterror:
Well it does but this is a bit better, will see what others say.

willkuer:
I just wonder whether it makes sense to run some script to revalidate those translated entries by just replacing the old string by the new string.

From my point of view a real ID that never changes and a po/mo for english would be much better. But as long as this is not available I’d suggest to consider some migration logic to enable changing english string without affecting all other localizations.

lightningterror:
@turtleli can you work your magic and create a script like that :)

PR #2641, opened 2018-10-15
GSdx: Change how some crc hacks work on native resolution

lightningterror:
Change the code so that some crc hacks on some games that remove
ghosting/blur issues on upscaled resolutions are disabled on native
resolution if crc hack level is below Aggressive state (most of the time).
They aren't needed then so why don't we aim for better accuracy.

Edit: @mirh I believe you'd like this change :smile:

### DIFF
--- a/plugins/GSdx/GSHwHack.cpp
+++ b/plugins/GSdx/GSHwHack.cpp
@@ -23,6 +23,7 @@
 #include "GSState.h"
 #include "GSdx.h"
 
+bool s_nativeres;
 static CRCHackLevel s_crc_hack_level = CRCHackLevel::Full;
 
 // hacks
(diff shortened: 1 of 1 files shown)


mirh:
I never was all that much into CRC hacks tbh. 
Anyway, why were (and still are on aggressive) upscaling ones enabled on native to begin with?

lightningterror:
For people with potato pcs or if they don't like blur effects so I left them there.

lightningterror:
You know what I might just disable those crc hacks altogether like you said.

mirh:
I'm not sure how performance relates to them.

avih:
From the commit message:

> They aren't needed then so why don't we aim for better accuracy.

Did you check each of those games, verified that the CRC hack fixes/does-something-useful when scaled, but does nothing useful (other than removing a perfectly good blur) at native resolution?

Some of those fixes, IIRC, are for performance reasons to begin with and not necessarily to fix scaling issues.

In general, the whole CRC hacks is hacky to begin with. I'm not entirely sure that adding automation based on resolution is great. Except maybe for hand-picked hacks which were verified to really fix a broken thing when scaled, but do absolutely nothing useful when unscaled.

Is this such a hand-picked list with these conditions?

mirh:
*thinks all hacks should have a thorough explanation for reproducibility and all* (not sure if that wasn't any hinting at my automatic hacks issue)

lightningterror:
I checked all except RedDeadRevolver and UltramanFightingEvolution throughout the year, anyway I just remembered that Street fighter EX is not an upscaling but a different kind of issue also present on native.

lightningterror:
> thinks all hacks should have a thorough explanation for reproducibility and all 

A lot of hacks were added by Cutie with kanji comments, but the comments later on got removed leaving all of us in the dark what they do. Unless you actually do some digging in commit history if you know what you're looking for. Some basic google translations could've been added.
https://github.com/PCSX2/pcsx2/commit/21522e71b2267f879b1c5cef0049959ac42adcf4

PR #2642, opened 2018-10-15
Switch to using the current cmake policies, not depreciated ones.

arcum42:
In the comments in the cmake file, it said that it had already been changed to support the new policies, but it hadn't been tested. Eventually, those depreciated policies will no longer be there, so switching is something that should be done at some point.

I switched both on, and built with all the extra plugins and with gtk 2 and gtk 3. It seemed to work both times. Doing as a PR request do it can get a bit of testing before being turned on in the main system. (And so I can see a travis cl build before committing.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,44 +2,9 @@
 project(Pcsx2)
 
 # Debian-based distributions require at least 2.8.5 due to multiarch.
-cmake_minimum_required(VERSION 2.8.5)
-
-# Keep OLD policy. Code was updated to support the new one but wasn't tested
-# Well, no time like the present to test it!
-#
-# cmake version 2.8.9
-#   CMP0018
-#        Ignore CMAKE_SHARED_LIBRARY_<Lang>_FLAGS variable.
-# 
-#        CMake 2.8.8 and lower compiled sources in SHARED and MODULE libraries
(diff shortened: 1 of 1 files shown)


gregory38:
Hum, we need to check 64 bits build. If I remember correctly -fpic should be mandatory for 64 bits.

gregory38:
I think it will be safer to bump cmake requirement to 2.8.12 (which introduce the policy 22).

And maybe check if we can replace FindGTK3.cmake by  pkg-config (i.e. CheckLib.cmake)

orbea:
Maybe it would be a good idea to add this to the `1.6` milestone? It might be annoying if this was not fixed before the next release and then newer cmake versions remove the deprecated policies.

arcum42:
64 and 32 bit versions have been tested, and I think we're set on this. Best way to get more testing at this point will be merging. Once travis and appveyor give a green light, anyways.

orbea:
I tested this PR here and it worked fine. I would suggest squashing the commits into one and not to be too pedantic, but iit's spelled "Deprecated" without the "i". :)

arcum42:
As long as people understood what I meant. I didn't say depreciated in the comments, at least.

And I'm just basically waiting on the automated builds finishing at this point...

PR #2643, opened 2018-10-16
GSdx-gui: Change internal resolution drop down list names

lightningterror:
Suggested by @MrCK1 

This should make things a bit easier for users when picking resolutions.

image

### DIFF
--- a/plugins/GSdx/GSdx.cpp
+++ b/plugins/GSdx/GSdx.cpp
@@ -237,15 +237,15 @@ void GSdxApp::Init()
 	m_gs_aspectratio.push_back(GSSetting(1, "4:3", ""));
 	m_gs_aspectratio.push_back(GSSetting(2, "16:9", ""));
 
-	m_gs_upscale_multiplier.push_back(GSSetting(1, "Native", ""));
-	m_gs_upscale_multiplier.push_back(GSSetting(2, "2x Native", ""));
-	m_gs_upscale_multiplier.push_back(GSSetting(3, "3x Native", ""));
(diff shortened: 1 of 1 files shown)


lightningterror:
> I think there was once a discussion that this is not entirely correct. Especially regarding p/i this map looks dangerous.

It should be ok because we are telling users what kind of results they can expect when upscaling. We don't need to bother them with the details how upscaling works regarding p/i.

ssakash:
It's technically wrong though, the native resolution differs per game. Some games can have a much more different internal resolution which makes the scaled values you gave to be not true when it's being used.

But I guess user understanding is more preferred here over technicality,  so don't know which one is better.

bositman:
True but I bet the average user doesn't care if it's 200 pixels or so off, this gives them a general idea and IMO is even better than not having anything at all. We've seen many cases of people just maxing it out without realizing they are using something close to 5K and thus should expect huge slowdowns

lightningterror:
Yes i know but overall it should be close most of the time. We don't need the exact value just something to give an idea what each option does.

lightningterror:
Merging this since it's a good idea/improvement.

bositman:
Why was there any warning before that those games actually scaled x2 to what you set? It's the same thing, with the added benefit that for -most- of the games, the approximate is giving you a pretty good idea of what you're gonna expect.

willkuer:
I am not even sure but I guess snowblind is around 720i30 already at native? I did not extensively follow the fixing discussion on the half screen issue...

No objections from my point. I think it’s good that people will use 3x native more often that before this commit.

PR #2653, opened 2018-10-20
Update Visual Studio project filters

tadanokojin:
VS filter fixes for you arcum.


arcum42:
Thanks! Was going to wait for the automatic builds to complete, then commit it, but @lightningterror beat me to it.

Just rebased and did some checking and all the changes I checked that were done to files that were moved are still there.

The real question's going to be what other changes we want to make to this branch before merging, and if there are any pr's we want to merge first and rebase after...

PR #2412, opened 2018-04-29
Add Input Recording/Playback Functionality

xTVaser:
First off, the vast majority of the credit for this work goes out to @pocokhc https://github.com/pocokhc/pcsx2-1.4.0-rr and @DocSkellington https://github.com/DocSkellington/pcsx2-1.4.0-rr I've just simply merged with upstream pcsx2 and made a few small bug fixes and improvements **to the recording features, not pcsx2 itself**.  

These changes are not without [issues](https://github.com/xTVaser/pcsx2-rr/issues) but the core recording functionality is better than anything I've previously seen or worked with on pcsx2 and in my experience is quite reliable (I've never lost a recording yet, desyncs are comparable to Dolphin's recording, etc).  

I will continue to work on polishing up these new features but **I would only want to address major-blocking issues before getting this merged** (otherwise, it never will lol).  In addition, we also have Lua scripting tools and features to contribute, but as this PR is already getting fairly large, I will PR those separately after getting this through.   Said code is currently in this branch https://github.com/xTVaser/pcsx2-rr/tree/add-lua-scripting

### PCSX2 Unimplemented / "Hidden" Features
* We have implemented the `Save/LoadStateOther_Click` which lets you arbitrarily pick a file to save or load to.  These have been refactored from using `Other` -> To/From`File` https://github.com/PCSX2/pcsx2/blob/f6e4a3081dbf49cb349adb1d8375c67bbcff517c/pcsx2/gui/MainMenuClicks.cpp#L531-L539
* We have exposed the `F12` video/audio recording capabilities of GSdx through a menu option
* We have also added a menu option to save a screenshot to a particular path rather than always the default `snaps/` folder 

### Preprocessor Defintions
* In order to facilitate the development story of testing whether or not these additions are causing problems there is a new preprocessor definition that removes all modifications within the pcsx2 core code (anything outside of the `Recording/` directory) - `DISABLE_RECORDING`

### GUI Modifications
* Two new top level menu's - 
  * `Capture`
    * This menu's submenu contains the options to capture video or take a screenshot, it is always displayed
  * `Recording` 
    * This menu's submenu contains everything related to starting a new input recording, opening the virtual controller GUI panels, or the recording editor panel.
    * This menu is not displayed by default and must be toggled on through the new menu option in the `System` submenu called `Enable Recording Tools`.  This toggle is also preceded by a prompt warning the user about the potential performance / instability / work-in-progress nature.
* A simplified window title when recording:
[...]

PR #2657, opened 2018-10-22
GSdx: Folder Reorganization

arcum42:
This pull request is for the pending reorganization of the folder structure on GSdx, making it better organized and easier to work with.

This does not have to be merged immediately, and might be better merged after any other pending GSdx changed, so it can be based on them. I also thought I'd give a chance for any suggestions for files to be moved or additional folders to be created before it gets merged.

Renaming the files as well is still a possibility, but I think it might be better to just do the merge and folder creation at the moment and minimize confusion.


lightningterror:
I prefer if we have proper commit naming starting with **GSdx:**, it's pretty much a naming standard at this point. Same goes for other type of changes (plugins, core, db).

lightningterror:
Last I checked windows worked so it should be ok ? Needs a rebase now tho, also I prefer to merge this after https://github.com/PCSX2/pcsx2/pull/2344

mirh:
Could also wait for TAS pr imho...
It's already yuge and tiresome enough I wouldn't want even the slightest of avoidable problems to hit it

lightningterror:
I forgot about that one.

willkuer:
TAS isn’t affecting gsdx, is it?

mirh:
It does [a little](https://github.com/PCSX2/pcsx2/pull/2412/commits/494a98392bec57b4819b8197ef70675e2f1ceb0a), for recording
Iirc that may get decoupled, for another time, but I wouldn't want to cause unneeded complications

lightningterror:
We can merge this now since TAS needs updating anyway.

arcum42:
It's a busy monday and I've got a few hours to go on work, but my last rebase was after TC. And GSDx was working at that point. 

You did one commit since then, but that didn't touch any of the files involved. We're probably fine on merging it.

lightningterror:
What about the old boost_spsc_queue.hpp file ? Shouldn't it be removed ?

arcum42:
Right, I did forget to go back and delete that, didn't I? Taken care of. Not tested in Windows, though, so we'll see if it compiles there.

plugins/GSdx/CMakeLists.txt
```diff
@@ -198,64 +159,103 @@ set(GSdxHeaders
...
+        Renderers/DX11/GSRendererDX11.cpp
+        Renderers/DX9/GSRendererDX9.cpp
+        Renderers/DXCommon/GSRendererDX.cpp
+        Renderers/DX11/GSTexture11.cpp
+        Renderers/DX9/GSTexture9.cpp
+        Renderers/DX11/GSTextureCache11.cpp
+        Renderers/DX9/GSTextureCache9.cpp
+        Renderers/DX11/GSTextureFX11.cpp
+        Renderers/DX9/GSTextureFX9.cpp
+        Renderers/DXCommon/GSTextureFX.cpp
```
turtleli: This file seems to be unused.

turtleli:
I don't see how that affects the placeholder being unnecessary?

arcum42:
I don't see any reason to delete it right now.

plugins/GSdx/CMakeLists.txt
arcum42: Yes. The cmake files do include the Windows files, even though they are not used, and I updated them to reflect the new paths.

lightningterror: The file GSTextureFX.cpp isn't included in the VS build files so it's only on CMakeList. I believe it can be purged.

lightningterror:
Should the Scanline files be in SW render folder ?

lightningterror:
Let's merge it then.

PR #2659, opened 2018-10-23
GSdx: Add an enum for channel fetch

lightningterror:
Maybe both ShaderConvert and ChannelFetch enums can be moved to GS.h.

@tadanokojin I believe you were looking for something like this ?

### DIFF
--- a/plugins/GSdx/GSDevice.h
+++ b/plugins/GSdx/GSDevice.h
@@ -28,7 +28,8 @@
 #include "GSAlignedClass.h"
 #include "GSOsdManager.h"
 
-enum ShaderConvert {
+enum ShaderConvert
+{
 	ShaderConvert_COPY = 0,
 	ShaderConvert_RGBA8_TO_16_BITS,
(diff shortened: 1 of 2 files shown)


tadanokojin:
That's what I was looking for thank you.

plugins/GSdx/GSDevice.h
```diff
@@ -52,6 +53,17 @@ enum ShaderConvert {
 	ShaderConvert_Count
 };
 
+enum ChannelFetch
+{
+	ChannelFetch_NONE  = 0,
```
willkuer: Thats camel case, snake case and uppercase all in one line. Enum classes are imo a better way to enable namespacing on enums than prefixes.

ssakash:
I don't know why you didn't change it into an enum class as willkuer said.

ssakash:
1/ You don't need the pseduo namespace when using enum class, that's the exact point on using it.
2/ I think there's a generic enum cast function available instead of doing ``static_cast<int>()``

PR #2661, opened 2018-10-24
Adjusting space between words

rafaelsandroni:


locales/ar_SA/pcsx2_Iconized.po
```diff
@@ -86,7 +86,7 @@ msgid ""
 "Please ensure that these folders are created and that your user account is "
 "granted write permissions to them -- or re-run PCSX2 with elevated "
 "(administrator) rights, which should grant PCSX2 the ability to create the "
-"necessary folders itself.  If you do not have elevated rights on this "
+"necessary folders itself. If you do not have elevated rights on this "
```
lightningterror: This will break the translation string since the gui string has 2 spaces as well.

lightningterror:
Sorry but I'll close this one since it doesn't actually change anything, other than introducing a regression.

Issue #1207, opened 2016-02-27
[App][l10n] Quick wizard language setting bug


piorrro33:
EDIT: Jump to the 4th message ;)

Hello,
I just noticed that this string is not present in the template and is therefore not localizable. http://a.pomf.cat/tsyavi.png
Hope this gets fixed, since this is quite minor.

mirh:
It's there to be honest. Both the [french](https://github.com/PCSX2/pcsx2/blob/master/locales/fr_FR/pcsx2_Main.po#L583) translation and [template](https://github.com/PCSX2/pcsx2/blob/master/locales/templates/pcsx2_Main.pot#L480). And I just tested and it works fine in Italian here.

piorrro33:
It seems like it is, but it's not loading for whatever reason. I'll compile my version and try if it works.

piorrro33:
Okay, I actually figured out the bug. So, when I pick System Default or French on the very first screen, and then I import my settings, it will turn back into English U.S. since it was the value that was set in my imported settings. So whether it is a bug or not, I don't know.

mirh:
I'd suggest to reword the issue title then

piorrro33:
Done.

gregory38:
So basically, import settings also gets the old language setting. Some people could see it as a feature rather than a bug.

mirh:
Mhh.. I'm not sure. 
Why Back button should be translated will Finish shouldn't?

willkuer:
What happens if you revisit this window after finalizing the initialization? Is this button always in the wrong language?

piorrro33:
I think it's localized if you revisit the window.

willkuer:
I guess gregory is right. If there is language information written in the ini it should be imported. However it should happen in a consistent way.

CookiePLMonster:
Turns out you still can confuse first time wizard so much so you end up seeing it in three different languages... I'll post repro steps later today, but as of the latest development build I got a 100% repro rate on this anomaly.

It looks like different elements of windows pick their language basing on different rules (I also was able to get Back and Cancel in one language, and Next in another).

lightningterror:
I guess the solution would be to first popup the window for importing settings, and not after the language is selected.

CookiePLMonster:
I would just ignore language when importing - it's the kinda setting which
you may not want to migrate.

And even then half of the buttons being in wrong language implies a
different kind of bug somewhere. Had the settings been overwritten, it
[...]

PR #2664, opened 2018-10-29
Cache strings for "Next>" or "Finish" buttons in wxWizard

CookiePLMonster:
This is a partial fix for #1207, as described in "Bug no. 1".

Back/Cancel buttons were initialized only once when wizard was constructed, so translations were also applied only once at that time. However, Next/Finish buttons had its string replaced when switching pages (so Next can become Finish when appropriate), and that triggered retranslation every time. If translation changed between those two events, you'd end up with a frankenstein navigation bar in two languages:

lang-mix

This PR caches translated strings for Next and Finish at the time of initial creation of those buttons. This has two added benefits:
- Not triggering retranslation every time string changes.
- "Next>" and "Finish" strings are not duplicated in the code anymore - and they could be considered "magic values".

#1207 also triages an issue of selected language being ignored at a point - however, correct behaviour has not been fully agreed upon yet, so for now I am only including a fix for the part everyone can agree on that it's a bug.

### DIFF
--- a/3rdparty/wxwidgets3.0/include/wx/generic/wizard.h
+++ b/3rdparty/wxwidgets3.0/include/wx/generic/wizard.h
@@ -131,6 +131,10 @@ protected:
                 *m_btnNext;     // the "Next>" or "Finish" button
     wxStaticBitmap *m_statbmp;  // the control for the bitmap
 
+    // cached labels so their locale stays consistent
+    wxString m_nextLabel,
+             m_finishLabel;
+
     // Border around page area sizer requested using SetBorder()
     int m_border;
 
--- a/3rdparty/wxwidgets3.0/src/generic/wizard.cpp
+++ b/3rdparty/wxwidgets3.0/src/generic/wizard.cpp
@@ -428,7 +428,10 @@ void wxWizard::AddButtonRow(wxBoxSizer *mainColumn)
         btnHelp=new wxButton(this, wxID_HELP, wxEmptyString, wxDefaultPosition, wxDefaultSize, buttonStyle);
 #endif
 
-    m_btnNext = new wxButton(this, wxID_FORWARD, _("&Next >"));
+    m_nextLabel = _("&Next >");
+    m_finishLabel = _("&Finish");
+
+    m_btnNext = new wxButton(this, wxID_FORWARD, m_nextLabel);
     wxButton *btnCancel=new wxButton(this, wxID_CANCEL, _("&Cancel"), wxDefaultPosition, wxDefaultSize, buttonStyle);
 #ifndef __WXMAC__
     if (GetExtraStyle() & wxWIZARD_EX_HELPBUTTON)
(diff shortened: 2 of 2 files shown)


lightningterror:
I'm not sure it's a good idea to modify WX libraries directly, @MrCK1 also mentioned this.

Can't this be done modifying the wx code in pcsx2 instead ?

CookiePLMonster:
I'm not sure - but judging by the nature of this issue I would guess not.

lightningterror:
If not then it's better to make a PR on the wxWidgets repo instead since we don't maintain the libraries but the providers do.
https://github.com/wxWidgets/wxWidgets

CookiePLMonster:
> Edit: We only update the libraries to newer versions that are provided.

This commit https://github.com/PCSX2/pcsx2/commit/43f5b5e2a655ec3ef40bf99f83dac0c894cff1e0 is dated 2016, Does this mean there's been no wxWidgets updates placed in PCSX2 since then?

MrCK1:
What @lightningterror means is that these changes will be overwritten next time we update the WxWidgets library because it's in the `3rdparty` directory. A more permanent change would be to make a ticket in the Wx repostory (and report it there) or see if it's a bug with the First Time Wizard.

CookiePLMonster:
Assuming PCSX2's widgets is unmodified, I can imagine it should be possible to trigger an identical issue independently of PCSX2. However, there may be a way to prevent this and PCSX2 may not be accounting for this properly. I'll try to investigate.

arcum42:
The Linux version uses the copy of wxwidgets on the system, not the copy in the trunk, which is there for the benefit of Windows. Any changes you make to wxWidgets will not carry over to Linux.

The latest *stable* version of wxwidgets is 3.0.4, so we're within a few minor versions. The latest development version is 3.1.1, but the stable version is the version most distributions have.

lightningterror:
I don't know how the update handled otherwise I could do an update to the latest stable release.

CookiePLMonster:
Fix pushed upstream: https://github.com/wxWidgets/wxWidgets/pull/1000

CookiePLMonster:
The change is now merged in https://github.com/wxWidgets/wxWidgets. However, it cannot be backported to 3.0.x so until PCSX2 migrates to 3.1.x it will remain an issue.

lightningterror:
There have been some commits made to the wx library so I'll post this here so I can keep track when bumping the version.

https://github.com/PCSX2/pcsx2/commits/master/3rdparty/wxwidgets3.0

https://github.com/PCSX2/pcsx2/blob/1279112ac0b0233ff9241a1ee8f64bcc440ee9fc/3rdparty/wxwidgets3.0/src/msw/toplevel.cpp#L1109

CookiePLMonster:
Sure. This file changed a bit in 3.1.x so when upgrading a merge will be
required either way.

Gathering information on pcsx2 wx tweaks sounds like a job for a wiki page,
actually. For the future, merges will be a lot easier if any modified code
is kept in uncommented preprocessor define like PCSX2_WX, with custom code
going in an else branch for this (so defining this preprocessor value
didables all custom tweaks and reverts wx to a pristine state).

PR #2665, opened 2018-10-29
LilyPad: Call XInputGetState only when XInputGetExtended is not

CookiePLMonster:
This PR contains two changes:

1. Removed a nasty hack in form of `pXInputEnable = (_XInputEnable)-1;` which was supposed to ensure XInput DLLs won't be attempted to be loaded multiple times if they failed once already. What's wrong with just using a dedicated boolean for that condition, so we don't need to resort to a dirty hack like this?
2. Moved `XInputGetState` in a way so it does **not** get called if SCP's `XInputGetExtended` is supported and succeeds.
Before that change, `XInputGetState` output values ended up being discarded fully. This could be considered a minor optimization, since rumor has it that `XInputGetState` can be relatively slow. Now there is no need to worry whether it is slow or not, since it's not called needlessly anymore.


lightningterror:
I'll keep this open for a while (2 weeks before merging) so I can test if everything works alright on w7.

plugins/LilyPad/XInputEnum.cpp
```diff
@@ -268,9 +270,9 @@ void EnumXInputDevices()
 {
     wchar_t temp[30];
     if (!pXInputSetState) {
-        // Also used as flag to indicute XInput not installed, so
+        // XInput not installed, so
         // don't repeatedly try to load it.
```
turtleli: Maybe have the comment on a single line instead of two.

Personally I would split the commit into 2 (bool change, call order change).

lightningterror:
@MrCK1 had some concerns if this breaks SCPToolkit compatibility.
https://github.com/PCSX2/pcsx2/commit/27f6fb1ff1bb604086620e94cd237f8ab9febf48#diff-6538b732ea2a9e456db585e800ebec39

turtleli:
I confirm it works fine both with and without the SCP XInput wrapper.

plugins/LilyPad/XInputEnum.cpp
CookiePLMonster: Do you want me to split it and force push to this PR?

turtleli: I think it'd be better to.

CookiePLMonster:
Applied all suggested changes.

lightningterror:
It would've been nice if the commit names started with `Lilypad:` but I guess it's a bit late for that.

CookiePLMonster:
Yeah - something to keep in mind for the future though!

PR #2676, opened 2018-11-05
GameDB: Remove CRC check for Forbidden Siren 2 patch

MrCK1:
Removing this CRC check allows the patch to work with another version containing the same serial but a different CRC (as noted in https://forums.pcsx2.net/Thread-bug-report-Forbidden-Siren-2-Pal?page=2)

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -2305,7 +2305,7 @@ Name   = Forbidden Siren 2
 Region = PAL-M5
 Compat = 5
 XgKickHack = 1 // SPS.
-[patches = C040B6AB]
+[patches]
 	comment=swapping a COP2 op into place to make the flags work without delays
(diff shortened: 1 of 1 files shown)


pgert:
A comment here would be in order, yes?
// This disc is known to have (at least) two CRCs - C040B6AB & ????????.

My guess is that this other CRC comes from an altered version of C040B6AB.
Someone have incorporated some cheat and/or widescreen patch.
Another good reason for removing the CRC check from the GameDB patches in general.

lightningterror:
> Another good reason for removing the CRC check from the GameDB patches in general.

Yeah I'm considering doing that as well, at least for games where patches fix hangs.
You can't make things worse, it's either take the chance to fix untested versions of the game or leave them in a not working state until someone actually reports the issue.

PR #2677, opened 2018-11-07
Removed crc id 0x7D4EA48F (multiloader)

Al0ner:
Removed crc id 0x7D4EA48F and mentions of that id and it's duplicates. The id is from a multiloader when packing images, and not an actual crc id from a game(s).

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -25,7 +25,7 @@
 
 CRC::Game CRC::m_games[] =
 {
-	// Note: Id 0x7ACF7E03 shouldn't be added as it's from the multiloader when packing games.
+	// Note: Id's 0x7ACF7E03, 0x7D4EA48F, 0x37C53760 - shouldn't be added as it's from the multiloaders when packing games.
 	{0x00000000, NoTitle, NoRegion, 0},
 	{0xF46142D3, ArTonelico2, NoRegion, 0},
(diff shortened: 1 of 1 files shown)


plugins/GSdx/GSCrc.cpp
```diff
@@ -25,7 +25,7 @@
 
 CRC::Game CRC::m_games[] =
 {
-	// Note: Id 0x7ACF7E03 shouldn't be added as it's from the multiloader when packing games.
+	// Note: Id's 0x7ACF7E03, 0x7D4EA48F, 0x37C53760 - shouldn't be added as it's from the multiloaders when packing games.
```
lightningterror: IDs

Al0ner:
Yes, I personally checked it. These are just different modifications of the multiloader.

avih:
> Yes, I personally checked it. These are just different modifications of the multiloader.

And is "this" multiloader used for many multi-discs? or is it more like one variation/CRC per multi-disc? if it's the latter, then possibly it'll be better to keep them for the games which need fixing IMO.

Al0ner:
0x7ACF7E03 - occurs VERY often
0x7D4EA48F, 0x37C53760 - less often, but I saw them more than once

avih:
> but I saw them more than once

Other than fixing an issue with one(?) game, do you know if it hurts the other games in some way?

If it doesn't then maybe it's still worth keeping it, and adding a comment that this CRC is for a multi-disc loader and intended for some game X at this disc, and for now there are no known issues with the other games at the disc?

PR #2678, opened 2018-11-09
Update Dutch Translations

mourix:
Added all strings that were missing from the template.


pgert:
@mourix:
The dutch translation could now use an update (if you're up for it).

PR #2681, opened 2018-11-10
GSdx-TC: Use PaletteMap also when 8-bit texture is disabled

AlessandroVetere:
Enabled caching of clut copies with `PaletteMap` also in the case 8-bit texture is disabled, which is the default (on #2344 the caching of clut copies _and palette textures_ was done only when 8-bit texture was enabled).
Brings moderate speedups ~10% in the most concitated parts of the ZoE2 Anubis benchmark, but may improve performances in all the cases when there are many `Source` objects created with clut copies to be stored.
The quality of the comments has been improved to better highlight the mechanics of the caching system.

### DIFF
--- a/plugins/GSdx/GSTextureCache.cpp
+++ b/plugins/GSdx/GSTextureCache.cpp
@@ -168,7 +168,8 @@ GSTextureCache::Source* GSTextureCache::LookupDepthSource(const GIFRegTEX0& TEX0
 		// If it is too expensive, one could cut memory allocation in Source constructor for this
 		// use case.
 		if (palette) {
-			AttachPaletteToSource(src, psm_s.pal);
+			// Palette is attached (Palette object, clut copy and palette texture) directly because the texture is not going in the texture cache list
+			AttachPaletteToSource(src, psm_s.pal, true);
 		}
 
 		m_src.m_surfaces.insert(src);
@@ -402,7 +403,7 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 	}
 
 	if (src->m_should_have_tex_palette && (!src->m_clut || !GSVector4i::compare64(src->m_clut, clut, psm_s.pal * sizeof(uint32)))) {
-		AttachPaletteToSource(src, psm_s.pal);
+		AttachPaletteToSource(src, psm_s.pal, true);
 	}
 
 	src->Update(r);
@@ -1339,6 +1340,7 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
 		// Because the texture is already on the GPU, CPU can't convert it.
 		if (psm.pal > 0) {
 			src->m_should_have_tex_palette = true;
+			// Palette (Palette object, clut copy and palette texture) is going to be attached in Source lookup code exploiting initial condition !src->m_clut
 		}
 		// Disable linear filtering for various GS post-processing effect
 		// 1/ Palette is used to interpret the alpha channel of the RT as an index.
@@ -1453,15 +1455,12 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
 		{
 			src->m_texture = m_renderer->m_dev->CreateTexture(tw, th, Get8bitFormat());
 			src->m_should_have_tex_palette = true;
+			// Palette (Palette object, clut copy and palette texture) is going to be attached in Source lookup code exploiting initial condition !src->m_clut
 		}
 		else {
 			src->m_texture = m_renderer->m_dev->CreateTexture(tw, th);
-
-			if (psm.pal > 0) {
-				uint16 palette_size = psm.pal * sizeof(uint32);
-				src->m_clut = (uint32*)_aligned_malloc(palette_size * sizeof(uint32), 64);
-				memcpy(src->m_clut, (const uint32*)m_renderer->m_mem.m_clut, palette_size);
-			}
+			// by default: src->m_should_have_tex_palette = false;
+			AttachPaletteToSource(src, psm.pal, false); // Attach only Palette object and clut copy
 		}
 	}
 
@@ -1599,8 +1598,8 @@ GSTextureCache::Source::Source(GSRenderer* r, const GIFRegTEX0& TEX0, const GIFR
 
 GSTextureCache::Source::~Source()
 {
-	if (!m_palette_obj) {
-		_aligned_free(m_clut);
+	if (!m_should_have_tex_palette) {
+		// Eventually valid reference for m_palette is not managed by PaletteMap, so it has to be recycled
 		m_renderer->m_dev->Recycle(m_palette);
 	}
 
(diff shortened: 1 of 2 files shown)


lightningterror:
Another early Christmas present :)

lightningterror:
I'll go ahead and merge it. @ssakash checked out the code and said it's fine. @tadanokojin tested the pr on a couple of games and didn't spot any regressions.

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -2046,30 +2045,39 @@ void GSTextureCache::SourceMap::RemoveAt(Source* s)
 	delete s;
 }
 
-// Query the PaletteMap for a valid Palette, then assign both palette texture pointer and clut copy pointer to the Source object
-void GSTextureCache::AttachPaletteToSource(Source* s, uint16 pal)
+// Query the PaletteMap for a valid Palette, then both its reference, clut copy pointer and eventually palette texture pointer to the Source object
```
turtleli: So I'm having trouble parsing the comment. Did you mean to drop the assign? And what does "eventually palette texture pointer" mean?

AlessandroVetere: The "assign" was removed by error in fact, now the statements lacks a verb, sorry for that!
The "palette texture pointer" is the pointer GSTexture* (m_tex_palette in Palette) to the (in the 8-bit texture enabled case) allocated texture with as content the copy of the clut in the relevant Palette object.
This texture pointer is then assigned in the 8-bit texture enabled case to the Source object, whereas in the other case it is null and not assigned to the Source.

AlessandroVetere: Yes I misused the term "eventually" (false friend!); What I meant was "optionally", with the positive case being when the `need_gs_texture` parameter is true, and viceversa.

lightningterror: So what should be the sentence changed to ? I can make a quick push.

Eurose209: From the code and what the original comment was, I'd suggest:
```suggestion
// Query the PaletteMap for a valid Palette, then assign its reference, CLUT copy pointer and optionally its palette texture pointer to the Source object.
```

AlessandroVetere: The proposed change looks fine

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -2046,30 +2045,39 @@ void GSTextureCache::SourceMap::RemoveAt(Source* s)
...
+		m_tex_palette = m_renderer->m_dev->CreateTexture(256, 1);
+		m_tex_palette->Update(GSVector4i(0, 0, pal, 1), m_clut, palette_size);
+	}
+	else {
+		m_tex_palette = nullptr;
+	}
 }
 
-// Default destructor, recycles palette texture and frees clut copy
+// Default destructor, frees clut copy and recycles eventual palette texture
```
turtleli: Would have been nice to fix this too...

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -2046,30 +2045,39 @@ void GSTextureCache::SourceMap::RemoveAt(Source* s)
...
+		s->m_palette = p->GetPaletteGSTexture();
+	}
 	s->m_clut = p->GetClut();
 }
 
 // GSTextureCache::Palette
 
-// Creates a new palette texture with current clut content, keeping a reference to its copy
-GSTextureCache::Palette::Palette(const GSRenderer* renderer, uint16 pal) {
+// Creates a copy of the current clut with eventually a new palette texture with its content
```
turtleli: Ditto,

plugins/GSdx/GSTextureCache.h
```diff
@@ -55,12 +55,12 @@ class GSTextureCache
...
-		GSTexture* m_tex_palette; // Pointer to valid texture with relevant clut as content
-		const GSRenderer* m_renderer; // Pointer to the current renderer, needed to recycle the referenced GSTexture on destruction
+		GSTexture* m_tex_palette; // Pointer to valid texture with relevant clut as content, if instantiated by the constructor
+		const GSRenderer* m_renderer; // Pointer to the current renderer, needed to recycle the eventually referenced GSTexture on destruction
 
 	public:
-		Palette(const GSRenderer* renderer, uint16 pal); // Creates a copy of the current clut and a texture with its content
-		~Palette(); // Default destructor, recycles palette texture and frees clut copy
+		Palette(const GSRenderer* renderer, uint16 pal, bool need_gs_texture); // Creates a copy of the current clut and, if needed (need_gs_texture == true), a texture with its content
+		~Palette(); // Default destructor, frees clut copy and eventually recycles palette texture
```
turtleli: Ditto.

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -1599,8 +1598,8 @@ GSTextureCache::Source::Source(GSRenderer* r, const GIFRegTEX0& TEX0, const GIFR
 
 GSTextureCache::Source::~Source()
 {
-	if (!m_palette_obj) {
-		_aligned_free(m_clut);
+	if (!m_should_have_tex_palette) {
+		// Eventually valid reference for m_palette is not managed by PaletteMap, so it has to be recycled
```
turtleli: Ditto.

plugins/GSdx/GSTextureCache.cpp
```diff
@@ -2046,30 +2045,39 @@ void GSTextureCache::SourceMap::RemoveAt(Source* s)
...
 }
 
-// Query the PaletteMap for a valid Palette, then assign both palette texture pointer and clut copy pointer to the Source object
-void GSTextureCache::AttachPaletteToSource(Source* s, uint16 pal)
+// Query the PaletteMap for a valid Palette, then both its reference, clut copy pointer and eventually palette texture pointer to the Source object
+void GSTextureCache::AttachPaletteToSource(Source* s, uint16 pal, bool need_gs_texture)
 {
-	std::shared_ptr<Palette> p = m_palette_map.LookupPalette(pal);
+	std::shared_ptr<Palette> p = m_palette_map.LookupPalette(pal, need_gs_texture);
 	s->m_palette_obj = p;
```
turtleli: I just noticed the shared pointer usage here is a bit inefficient.
```
{
	std::shared_ptr<Palette> p = m_palette_map.LookupPalette(pal, need_gs_texture);  // copy, +1 ref
	s->m_palette_obj = p; // copy, +1 ref
	// Rest of code in function
	// p destroyed, -1 ref
}
```
It could be
```
{
	s->m_palette_obj = m_palette_map.LookupPalette(pal, need_gs_texture);  // copy, +1 ref
	// Rest of code in function
}
```

plugins/GSdx/GSTextureCache.cpp
lightningterror: Didn't know, I guess because nobody pointed them out. I'll make another quick push along with the shared pointer suggestion soon-ish.

turtleli: So did no one reviewing the code actually read the comments before it was merged? The comment issue should have been spotted.

Also, maybe you should PR the changes to potentially avoid quick push after quick push after quick push.

lightningterror: Nope, didn't have time. I believe kojin and beard tested it when I asked on discord for tests.

I don't have time to do benchmarks so I'll pass.

As for the comments themselves I'm fine with the purge.

AlessandroVetere: @lightningterror please, just go ahead and purge the comments; otherwise I'll open PR this WE :P
As for the benchmarks, I would suggest as positive test cases those that were improved by paltex optimization, e.g. ZoE 2 bosses, MGS 3, Jak 2 (especially "underport level" IIRC), maybe R&C.

lightningterror: I think I'll leave the purge to you, I have my hands full with other stuff as it is.

orbea:
This PR did create some regressions, see issue https://github.com/PCSX2/pcsx2/issues/2702.

PR #2685, opened 2018-11-13
Gamedb: Ease Ghost in the Shell COP2 patch

lightningterror:
Makes shadows work properly on progressive mode.
Also remove crc check to make the Russian version work.

Testing was done by @ivan89el.

Perhaps the patch for the PAL release should be tested and verified as well.


pgert:
@ivan89el:
Does this mean that the RU release has the same serial-number as the NTSC-U release?
Becase otherwise, removing the CRC check won't make any difference.
A new entry for the RU release has to be made, yes?

ivan89el:
@pgert 
Yes, RU has the same number as NTSC-U.

pgert:
@lightningterror:
If the NTSC-U and PAL-RU discs have the same serial-number,
 then may I suggest that you change the Region entry into
 "Region = NTSC-U & PAL-RU	// two discs with the same Serial, but with different CRC's".

lightningterror:
@ivan89el what's the crc id for the Russian version ?

ivan89el:
@lightningterror 
Cyclic redundancy check.
@pgert 
Russian version is based on NTSC-U

pgert:
@ivan89el:
Ha-ha-ha ...
You were asked (by lightningterror) to state the CRC id (number),
 not what CRC stands for (abbreviation).
The CRC for the NTSC-U disc is C040B6AB.
What is the CRC for the PAL-RU disc, the disc you have?
Extract it from the log-window.

lightningterror:
I tested the game and it works fine on both modes.

ivan89el:
Meine id: SLUS_210.06;1 crc=0x28557423

tdiep06:
hey do you think we should add an eyetoy plugin to the pcsx2 in the next dev build.
because i am desperately waiting to play it

lightningterror:
I only removed the patch that caused issues with the shadows on both EU and US regions. The rest will stay because we don't know what they did and they might be needed.

ivan89el:
@lightningterror 
Good job. A regression was found in my version. The effect of blood in the screen that happens fleetingly so it’s impossible to take a screenshot. Your version is fine.

lightningterror:
Yea that's what I was afraid of, so it looks like I made the right choice.

PR #2690, opened 2018-11-17
PCSX2: "FMV Aspect Ratio switch" feature

lightningterror:
Cleaned up the commit a bit, also adjusted that the option works on 4:3 Window Aspect ratio as well.

Change "Switch to 4:3 aspect ratio when an FMV plays" to
"FMV Aspect Ratio switch" drop down menu.

The option temporarily changes the Aspect Ratio to either
16:9 or 4:3 depending on the option that is selected,
and depending on the Window Aspect ratio that is selected.

Original commit:
https://github.com/PCSX2/pcsx2/pull/1918/commits/8d471e1f1683af1d23a274a165ee1865633f9e82

Co author: lightningterror.

Anyone is free to make any desired changes so we can get this up and running on master branch.


pcsx2/gui/AppConfig.cpp
```diff
@@ -810,26 +810,26 @@ void AppConfig::FilenameOptions::LoadSave( IniInterface& ini )
...
+	Zoom                            = 100;
+	StretchY                        = 100;
+	OffsetX                         = 0;
+	OffsetY                         = 0;
+
+	WindowSize                      = wxSize( 640, 480 );
+	WindowPos                       = wxDefaultPosition;
+	IsMaximized                     = false;
+	IsFullscreen                    = false;
+	EnableVsyncWindowFlag           = false;
```
turtleli: IMO in most cases it's not a good idea to align things.

It may look pretty but I think it makes it harder to read the code because when you read from left to right (which is what actually matters in most cases) there's now a large whitespace gap that you have to skip to read the next part of the line. This:
```
WindowSize = wxSize( 640, 480 );
```
is IMO easier to read than
```
WindowSize                                                                              = wxSize( 640, 480 );
```
Also, it unnecessarily increases the diff, Anyone making future changes might have to adjust a lots of lines when they otherwise might only want to add/adjust/remove one line because it'll ruin the existing formatting. And the more lines that are modified, the greater the likelihood of a merge conflict if multiple people are working on nearby areas of code. It also makes it harder for a reviewer to see what's really changed (I don't really want to use the ignore whitespace option all the time).

lightningterror: Yeah those are good points, I'll make some adjustments.

pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -86,41 +99,42 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
...
 	//s_AspectRatio.AddGrowableCol( 0 );
 	s_AspectRatio.AddGrowableCol( 1 );
 
-	s_AspectRatio += Label(_("Aspect Ratio:"))        | pxMiddle;
-	s_AspectRatio += m_combo_AspectRatio              | pxAlignRight;
-	s_AspectRatio += Label(_("Custom Window Size:"))  | pxMiddle;
-	s_AspectRatio += s_customsize                     | pxAlignRight;
+	s_AspectRatio += Label(_("Aspect Ratio:")) | pxMiddle;
+	s_AspectRatio += m_combo_AspectRatio | pxAlignRight;
+	s_AspectRatio += Label(_("FMV Aspect Ratio switch:")) | pxMiddle;
```
lightningterror: I'm actually gonna rename the label to **FMV Aspect Ratio Override:**. It's been bugging me.

pcsx2/gui/AppConfig.cpp
```diff
@@ -880,6 +879,18 @@ void AppConfig::GSWindowOptions::LoadSave( IniInterface& ini )
 	};
 
 	ini.EnumEntry( L"AspectRatio", AspectRatio, AspectRatioNames, AspectRatio );
+
+	static const wxChar* FMVAspectRatioSwitchNames[] =
+	{
+		// FIXME: fmv_aspect_ratio_switch_labels[] is inverted
+		L"Off", // Standard (4:3)
+		L"4:3", // Widescreen (16:9)
+		L"16:9", // Off
```
turtleli: Why wasn't this fixed before it was merged?

lightningterror: Perhaps I should've changed fmv_aspect_ratio_switch_labels instead so the Off option remains at the top instead, will think about it.

PR #2691, opened 2018-11-18
 GSdx-osd: Remove/purge OSD Indicator code

lightningterror:
The code was never working properly ( tested by @tadanokojin ) so let's
clean up and remove it. If someone actually wishes to maintain it then
they can re add it and update the code to a functional stage.


ssakash:
Maybe just remove it from the GUI and still have the necessary code for it available? I don't see any burden in having it, as there's literally no overhead on having it. (Just makes it a pain if someone wants to look at history and add it back at some point)

lightningterror:
> I don't see any burden in having it.

Except when it comes to maintaining it. It will probably be left alone in the _unfinished features void_ that nobody touches.

arcum42:
IIRC, we aren't even totally sure what that code was intended to do. And we're probably already hanging on to too much legacy non-working code.

PR #2693, opened 2018-11-19
pcsx2: Fix some issues on FMV Aspect Ratio Override

lightningterror:
gui: Move the Off option to the top instead of bottom for FMV Aspect Ratio Override.

It's not ideal for it to be at the bottom.

Adjust the logic how the button works, there was an issue if both aspect ratios
were the same the aspect ratio chosen was selected to Fill.
Also adjust the code so the button better responds to realtime toggling.

If disabling the button realtime during a fmv then the normal aspect ratio
should be chosen.

There's still an issue left where if the fmv button is changed first and then
the aspect ratio button the button would would not work in realtime. Switching it off
seems to reset and make it work again.

Note: even the old 4:3 option had these same issues with realtime toggling.

### DIFF
--- a/pcsx2/gui/FrameForGS.cpp
+++ b/pcsx2/gui/FrameForGS.cpp
@@ -167,19 +167,21 @@ void GSPanel::DoResize()
 		switchAR = false;
 	}
 
-	if (!switchAR) {
-		if (g_Conf->GSWindow.AspectRatio == AspectRatio_4_3)
+	if (switchAR) {
+		if (g_Conf->GSWindow.FMVAspectRatioSwitch == FMV_AspectRatio_Switch_4_3) {
 			targetAr = 4.0 / 3.0;
-		else if (g_Conf->GSWindow.AspectRatio == AspectRatio_16_9)
+		} else if (g_Conf->GSWindow.FMVAspectRatioSwitch == FMV_AspectRatio_Switch_16_9) {
 			targetAr = 16.0 / 9.0;
-	} else if (switchAR) {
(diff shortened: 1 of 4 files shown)


pcsx2/gui/FrameForGS.cpp
```diff
@@ -167,19 +167,21 @@ void GSPanel::DoResize()
...
-		else if (g_Conf->GSWindow.AspectRatio == AspectRatio_16_9)
+		} else if (g_Conf->GSWindow.FMVAspectRatioSwitch == FMV_AspectRatio_Switch_16_9) {
 			targetAr = 16.0 / 9.0;
-	} else if (switchAR) {
-		if (g_Conf->GSWindow.FMVAspectRatioSwitch == FMV_AspectRatio_Switch_4_3 && g_Conf->GSWindow.AspectRatio != AspectRatio_4_3) {
+		} else {
+			// Allows for better real time toggling, returns to the non fmv override aspect ratio.
+			switchAR = false;
+		}
+	} else if (!switchAR) {
```
turtleli: @lightningterror Do you notice what is wrong with this else if condition?

lightningterror: switchAR is disabled and could just be changed to a normal else since if (switchAR) checks if it's enabled ?

turtleli: Yes.

I kinda prefer pointing things out before they're merged, not afterwards. It'd be nice if you could actually leave PRs open for longer to give people time to actually review things.

lightningterror: I admit I've been kinda rushing to get things merged faster, just so we could focus on other work without maintaining multiple prs conflict free.
I'll keep them open a bit longer from now on.

I'll update that check on the next wx pr I open.

PR #2695, opened 2018-11-19
pcsx2-gui: Adjust Recompiler, clamp mode gui names/labels

lightningterror:
Add default tags, making them a little bit more user friendly.

image

image

### DIFF
--- a/pcsx2/gui/Panels/CpuPanel.cpp
+++ b/pcsx2/gui/Panels/CpuPanel.cpp
@@ -29,13 +29,13 @@ Panels::BaseAdvancedCpuOptions::BaseAdvancedCpuOptions( wxWindow* parent )
 		RadioPanelItem(_("Nearest")),
 		RadioPanelItem(_("Negative")),
 		RadioPanelItem(_("Positive")),
-		RadioPanelItem(_("Chop / Zero"))
+		RadioPanelItem(_("Chop / Zero (Default)"))
 	};
 
 	const RadioPanelItem ClampModeChoices[] =
 	{
 		RadioPanelItem(_("None")),
-		RadioPanelItem(_("Normal")),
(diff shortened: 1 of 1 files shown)


MrCK1:
Personally, I think labeling defaults are enough. Interpreters don't need a debug label as the tooltip already says the same thing.

Users will be quick to switch back when everything runs at 2FPS :P

turtleli:
Is the bold font not enough to highlight what option is the default? There's also the Restore Defaults button that the user can use.

lightningterror:
I bet there's gonna be someone that doesn't know that bold means default. And it is slightly better when adjusting one setting without resetting all of them.

lightningterror:
> Doesn‘t that again mean you need to retranslate all those labels in all languages? 

Something to keep the translations busy :smile:

turtleli:
Problem is some translations haven't been updated in years. Is it really worth changing the string?

Personally I think the user will be able to figure out what the defaults are without too much difficulty,

Also, by changing the square brackets to round brackets the panel style actually becomes inconsistent with the other panels that also use square brackets.

> If at all I‘d append the localized `(default)` tag after localization so that only ‚default‘ needs to be translated.

I think it's better to keep strings whole to give translators more flexibility to deal with whatever language they're translating to.

lightningterror:
`Also, by changing the square brackets to round brackets the panel style actually becomes inconsistent with the other panels that also use square brackets.`

It should be fine since these are in separate tabs, plus EE Cache also uses round so it fits better in these tabs.

lightningterror:
So what are we gonna do with this ?

gregory38:
Imho, you should postpone all string change after 1.6 release (which should really be done). Otherwise, users will just get untranslated string which isn't an improvement

lightningterror:
Alright we can postponed it then.

PR #2703, opened 2018-11-22
Remove gigaherz old wip 'new' iopdma code.

arcum42:
Alright, so this was Gigaherz '"Improved DMA Handling" Engine WIP'. 

It was a pretty big rewrite back in the day. It had a lot of potential, but it never got finished, and gigaherz is no longer around. And it is a lot of code.

This PR removes it. It's possible I might have missed bits related to it, but this is all the #ifdef'd code.

My though on this is that we may as well get rid of it so the surrounding code is easier to read and we don't have legacy code sitting around. If someone wants to do the rewrite, it may be easier to start from scratch at this point, and you'd still have git history to look at for an example.

The pity is that some of the code it was replacing was pretty ugly. It would've been nice to see it actually finished.

I'm mainly putting it up as a pr to make sure we have a good consensus before deleting it, and also to make sure I didn't accidentally delete anything I shouldn't have, or leave bits of it behind.


arcum42:
@gigaherz Yeah, this is one of those pr's that I'd be perfectly happy to just close if someone said they wanted to work on it, but I don't see that happening. I'm also thinking creating a branch before committing this would be a good idea, so that if anyone *does* want to look at it or work on it, the branch is there.

plugins/spu2-x/src/PS2E-spu2.cpp
```diff
@@ -245,21 +245,13 @@ SPU2dmaInterrupt(s32 channel)
         return Cores[1].NewDmaInterrupt();
 }
```
turtleli: I think SPU2dmaRead, SPU2dmaWrite, and SPU2dmaInterrupt (just above) needs to be removed as well?

plugins/spu2-x/src/Dma.cpp
```diff
@@ -420,12 +413,10 @@ void V_Core::DoDMAwrite(u16 *pMem, u32 size)
     }
     Regs.STATX &= ~0x80;
 //Regs.ATTR |= 0x30;
-#endif
 }
 
 s32 V_Core::NewDmaRead(u32 *data, u32 bytesLeft, u32 *bytesProcessed)
 {
-#ifdef ENABLE_NEW_IOPDMA_SPU2
```
turtleli: This is ifdef, not ifndef.

Actually, you could probably remove NewDmaRead, NewDmaWrite, and NewDmaInterrupt.

ramapcsx2:
>> Sad to see it go, but as you said, no one touched it in 9 years, and even back then, no one was interested enough to help me finish this.

Hm? :p 

I think we ran into many issues here, mainly with the "timing" (cadence of operations).
The rewrite was supposed to work more like the real machine, but even when the new code worked, games weren't too happy about it. The timing made working games run slower and didn't fix bugs we hoped it would.

It stopped working at all later on, and has been in this state for ages now.
It's a pity that it didn't work out, but yeah, it won't be finished anymore.
Might as well delete it then.

If a new attempt comes around, it should take into account that PCSX2 doesn't like when parts of the core are replaced and others not. It needs a broader approach. In this case, probably all the EE + IOP DMA systems, SIF and VIF and whatnot included.
The effort is probably not worth it at this stage.

plugins/spu2-x/src/Dma.cpp
arcum42: I was worried I'd miss one of those, since it kept varying between #ifdef and #ifndef. And yeah, I should probably remove the functions for those three and the ones above as well. I'll look at it when I have a chance.

arcum42:
@turtleli The last commit removed the bits you mentioned, plus code I somehow missed in common/include/PS2Edefs.h and removing the definitions in plugins/spu2-x/src/Windows/Spu2-X.def.

The latter should probably be checked in Windows.

arcum42:
Any other issues with this, or is it ready to be merged?

Also, could someone fire it up in Windows and just make sure that last commit didn't cause any problems loading spu2-x?

lightningterror:
Fired up a couple of games and they have sound.

arcum42:
@lightningterror 
Cool. Thought everything was fine with that, but when editing Windows files, it's hard to tell. If that had broken it, the plugin probably wouldn't have been recognized at all...

arcum42:
Think I'll go ahead and merge this today, unless anyone sees anything.

PR #2705, opened 2018-11-22
pcsx2: Change EE Cyclerate speedhacks gui level colors

lightningterror:
0 - Green - no change
-1/+1 - Orange
-2/+2 - Light Red
-3/+3 - Red - no change.

https://media.discordapp.net/attachments/350645867876712449/515191395972284426/unknown.png

https://media.discordapp.net/attachments/350645867876712449/515191178543628290/unknown.png

https://media.discordapp.net/attachments/350645867876712449/515191268389945344/unknown.png


turtleli:
IMO there's hardly any difference between red and light red. If I wasn't looking for a colour difference I wouldn't have noticed it at all.

lightningterror:
I'll try and see if I can make the orange more on the yellow side without it being hard to read and then I can adjust the light red as well.

pgert:
Please don't make it more yellowish - it would make it too diffuse.

KrossX:
Pick from this color palette perhaps? [Tango Guidelines](http://tango.freedesktop.org/Tango_Icon_Theme_Guidelines)
Already chosen to be rather similar per column.

lightningterror:
Akash mentioned that he already has variable colors planned in his PR so we can avoid duplicate work
https://github.com/PCSX2/pcsx2/pull/1974

I kinda forgot about that, but I'd still say we can finish up this one until that one is finalized and merged.

lightningterror:
I think this is a good balance between functionality and readability, unless you go for other colors that are readable but will seem out of place.

image

willkuer:
Looks good 👍🏻

PR #2710, opened 2018-11-25
GSdx: Always enable Unscale Point and Line

lightningterror:
Always enable unscale point and line hack on upscaled resolutions, and
purge the gui option.

It was discussed in the past with @gregory38 that it can be enabled without
causing any issues which will be beneficial to getting rid of 1 less
option and fixing a bunch of fmv issues without relying on a gui toggle.


gregory38:
It is quite an old discussion. I vaguely remember that flatout found an issue (but it could have been on initial dx implementation)

On one hand, having an hidden (ini only) option could be safer and could ease test. On the other hand, we have already tons of option.

lightningterror:
> On one hand, having an hidden (ini only) option could be safer and could ease test. On the other hand, we have already tons of option.

I was thinking the same when working on it.
We could also purge clut_load_before_draw ini option as well since it's a crc hack for 1 game.

gregory38:
Please Keep this clut option. Behaviour isn't fully understood here.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -26,16 +26,7 @@
...
-	if (theApp.GetConfigB("UserHacks"))
-	{
-		UserHacks_unscale_pt_ln = theApp.GetConfigB("UserHacks_unscale_point_line");
-	}
-	else
-	{
-		UserHacks_unscale_pt_ln = false;
-	}
-}
+{ }
```
turtleli: I prefer
```
{
}
```
I don't think there are many that would use `{ }` with a space in between.

plugins/GSdx/Renderers/DX11/GSDevice11.cpp
```diff
@@ -40,20 +40,9 @@ GSDevice11::GSDevice11()
...
-	else
-	{
-		UserHacks_unscale_pt_ln = false;
-	}
 }
 
-GSDevice11::~GSDevice11()
-{
-}
+GSDevice11::~GSDevice11() { }
```
turtleli: Don't see why the destructor has been reformatted.

lightningterror: Thought it was better since there's no code inside it but sure I'll change both of them back.

lightningterror:
I went through the PR that did the port and didn't spot anything relating to having an issue there, I did see mentions for Trilinear filtering and mipmapping tho.
On a side note performance between P&L on and off seems the same, and I didn't spot any issues on the games I tested.

Ok let's merge it, I have more pending work on the gui that's waiting :)
More early Christmas goodies.

PR #2711, opened 2018-11-25
GSdx-d3d: Port write_rg from Opengl

lightningterror:
Rare corner case when both texture shuffle and date are enabled.

Commit:
https://github.com/PCSX2/pcsx2/commit/744f9ebc097982e5dd6387e8e02826ecf9335452

### DIFF
--- a/plugins/GSdx/res/tfx.fx
+++ b/plugins/GSdx/res/tfx.fx
@@ -39,6 +39,7 @@
 #define PS_POINT_SAMPLER 0
 #define PS_SHUFFLE 0
 #define PS_READ_BA 0
+#define PS_WRITE_RG 0
 #define PS_PAL_FMT 0
 #endif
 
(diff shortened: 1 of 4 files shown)


plugins/GSdx/res/tfx.fx
```diff
@@ -516,10 +517,15 @@ float4 tfx(float4 t, float4 c)
 void datst(PS_INPUT input)
 {
 #if PS_DATE > 0
-	float alpha = sample_rt(input.tp.xy).a;
 #if SHADER_MODEL >= 0x400
+#if PS_WRITE_RG == 1
```
lightningterror: PS_WRITE_RG check doesn't need to fall under SHADER_MODEL since we are using `#if`, the code can be reduced.

lightningterror:
I'll probably fix this up and merge it once the remaining depth/channel prs are merged.

lightningterror:
I'll close it, it's hard finding a game that hits this code patch so we can't test it.

Issue #2706, opened 2018-11-22
TextureCache: DBZ BT3 regression when Allow 8-bit Textures is enabled


lightningterror:
@AlessandroVetere looks like the Palette management rework https://github.com/PCSX2/pcsx2/pull/2344
caused a regression in DBZBT3, perhaps other games are affected as well.

In this example character faces appear as blocks when Allow 8-bit Textures is enabled. Note there are other graphical issues as well. I don't know if other games are affected but it's quite possible.

Image:
![gsdx_20181123003210](https://user-images.githubusercontent.com/18107717/48924431-6888f780-eeb8-11e8-8745-2f125b1f6841.png)

GS Dump:
[DBZ BT3 - TC 8-bit.gs.zip](https://github.com/PCSX2/pcsx2/files/2609410/DBZ.BT3.-.TC.8-bit.gs.zip)

Last working build:
v1.5.0-dev-2653-g77a924ee7

Regression build:
v1.5.0-dev-2654-g9fa1b290b

lightningterror:
I compiled the first commit only https://github.com/PCSX2/pcsx2/pull/2344/commits/44726afb780f06d3356c2b3f625bdc67f0d28b2f and that too has the regression. That should narrow things down.

AlessandroVetere:
I have so the game so I can debug the issue.
Fortunately, the entry points in the existing code of the PR are very few, so in the worst case reverting the code would not be a huge work; but I think this can be solved.
My fear is that moving these lines:

https://github.com/PCSX2/pcsx2/commit/44726afb780f06d3356c2b3f625bdc67f0d28b2f#diff-7436adf7406199c56fd673a16e6086b2L1460-L1464

Did change the 8-bit texture enabled behaviour, as now the first clut lookup is done at:

https://github.com/PCSX2/pcsx2/commit/44726afb780f06d3356c2b3f625bdc67f0d28b2f#diff-7436adf7406199c56fd673a16e6086b2R391-R392

So if in between the first creation of the Source and the first Lookup the CLUT changes, now we pick the latter CLUT values instead of the former.
In small words, the `m_initpalette` mechanism has been dropped, and maybe needs to return.
If this is the case, I can design a solution.

Could you confirm that with 8-bit texture DISABLED, the game doesn't have regressions? In this case the codepath has remained the same, given that the PaletteMap works.

lightningterror:
Yea having the option disabled doesn't have the regression.

[...]

Issue #2702, opened 2018-11-21
Tales of Legendia - Segfaults with CRC level 'None' or 'Minimum'


orbea:
***PCSX2 version:***
https://github.com/PCSX2/pcsx2/commit/3d6921d9ca05ac0b5a637a6c5163110e3133d0a8

***PCSX2 options:***
`Automatic CRC Level: None or Minimum`

***Plugins used:***
`GSdx (GCC 8.2.0 SSE2/AVX) 1.1.0 [libGSdx]`

The legacy GSdx is unaffected.
`GSdx (GCC 8.2.0, SSE2) 1.0.0 [libGSdx-legacy-1.0.0]`

***Description of the issue:***

When playing `Tales of Legendia` with CRC level `none` or `minimum` pcsx2 will segfault consistently in certain parts of the games. For example:

* When loading a save to the world map.
* When opening `Config > Video (GS) > Plugin Settings...` during emulation of a field map.
* When stopping emulation.
* When just running the game for some period of time.

With the first method sometimes this assert will be shown before it proceeds to crash.
```
/tmp/SBo/pcsx2/common/src/Utilities/Linux/LnxHostSys.cpp(70) : assertion failed:
    Function:  void SysPageFaultSignalFilter(int, siginfo_t*, void*)
    Thread:    MTGS
    Condition: false
    Message:   Unhandled page fault @ 0x000000a7

Stacktrace:
[00] wxSizer::Add(wxWindow*, wxSizerFlags const&)
[01] __kernel_rt_sigreturn                       
[02] 0x0xd79ac67d     
```
I tried to use GDB, but I am having difficulty actually proceeding into the game and then capturing the crash. Please see issue https://github.com/PCSX2/pcsx2/issues/1806 for further discussion.

***How to reproduce the issue:***

Here are two methods which seem to reproduce it consistently.

1. Set the `Automatic CRC Level` to `None`.
2. Start `Tales of Legendia`.
3. Proceed to a field map such as the start of the game before the first town.
4. Open `Config > Video (GS) > Plugin Settings...`
5. pcsx2 should segfault.

Or

1. Set the `Automatic CRC Level` to `None`.
2. Start `Tales of Legendia`.
3. Load the a save game to the world map.
4. It should crash shortly after loading and presenting the world map with very wrong colors.

Here is a memory card containing a save file.
http://0x0.st/sU54.xz

***Last known version to work:***

I bisected this.
```
$ git bisect bad
[...]

PR #2713, opened 2018-11-25
GSVector4i: Fix compare64 function

AlessandroVetere:
Fix `GSVector4i::compare64` function by adjusting the number of iterations needed to cover the comparison of all the elements of the input arrays, where the number of bytes to compare is indicated in the `size` parameter of the function. 
The function is used only in GSdx-TC to compare CLUT arrays.
On one hand this fix closes the recent #2706 and possibly the color regressions indicated in #2702 (which were both introduced with the new `PaletteMap` mechanism, which partially replaced the `GSVector4i::update` logic with the bugged `GSVector4i::compare64`).
On the other hand this might impact long standing wrong texture cache lookup when palette was involved with 8-bit palette disabled (to be verified which ones).

### DIFF
--- a/plugins/GSdx/GSVector4i.h
+++ b/plugins/GSdx/GSVector4i.h
@@ -1945,7 +1945,7 @@ public:
 	{
 		ASSERT((size & 63) == 0);
 
-		size >>= 6;
+		size >>= 4;
 
 		GSVector4i* s = (GSVector4i*)src;
 		GSVector4i* d = (GSVector4i*)dst;


lightningterror:
There's still an issue tho, DBZ BT3 will either crash the application at the beginning or the GS percentage will keep on increasing until it reaches 100 and the fps will continue to drop down to 0 before crashing again.

orbea:
@lightningterror Maybe the same issue as https://github.com/PCSX2/pcsx2/issues/2702 which regressed in a different commit?

lightningterror:
> I think the correct fix is to change the i += 4 in the for loop to ++i.

That did the trick.

Issue #2687, opened 2018-11-15
Destroy all Humans HD Crawl


RinMaru:
PCSX2 1.5.0-Dev-2653
GsDxAVX2

Windows 10 x64
CPU- i5 6402p 8GB
GPU- AMD RX 480 4GB

Destory all humans will go to a crawl in HD going from 100% to 7% 2 Frame per 3 seconds
The Ram will fill up as windows starts to close other tasks and the CPU goes into 88% Use as the rest of the system becomes nearly unresponsive.
ways to reproduce is set a HD res and start the game then go to load game and watch the game slow down. switching to native res fixes the issue and once your ingame you can go back to HD but if another blue window pops up it will slow down again

lightningterror:
Probably related to https://github.com/PCSX2/pcsx2/issues/1318

Do you have Large Framebuffer enabled ?

RinMaru:
yea and... that fixed it. last time I tried that it didn't should this be left open or just close it?

lightningterror:
Yea disable it if it causes issues.

lightningterror:
See https://github.com/PCSX2/pcsx2/commit/dce6d3f4517615eabc84078e0bf448c29e56e6ee for more info

RinMaru:
shame Street racing syndicate still locks up the entire system due to a similar issue not even large frame buffer disabled keeps it from eating all ram and forcing me to reset the desktop

MrCK1:
Did you turn down your resolution?

RinMaru:
anything higher then native locks up SRS I think I have an issue open on it and it reguires very specific settings to run

RinMaru:
Issue is caused by Frame Buffer conversion in HW Hacks

Issue #1313, opened 2016-04-26
Gran Turismo series PCSX2 Crashes


MarcoEstevez:
PCSX2 Crashes when going in races, no issues while on menus.

PCSX2 1.5.0-20160426190932 - compiled on Apr 26 2016
    Operating System =  Microsoft Windows 10 Home (build 10586), 64-bit
    Physical RAM     =  7986 MB
    CPU name         =  Intel(R) Core(TM) i7-4700MQ CPU @ 2.40GHz

OpenGL information. GPU: GeForce GT 755M/PCIe/SSE2. Vendor: NVIDIA Corporation. Driver: NVIDIA 364.72

---

I did several tests, and I managed to reproduce only on IR greater than Native.

![image](https://cloud.githubusercontent.com/assets/3447089/14836241/f0a23c38-0c0e-11e6-9b35-1d59047d66ae.png)

If I start the game with Native res. then it works fine, while on game If I change IR to 2x game works during few seconds at full speed until it crashes after few seconds.

---

Im not sure if that is a general issue or its only me due a weak GPU, but I remember being able to play it a 3x on previous builds.

---

Im not sure if that related or not but PCSX2 Console get spammed with thousands of messages.

SummerTime=0
Timezone=60
SummerTime=0
Timezone=60
SummerTime=0
Timezone=60
SummerTime=0
Timezone=60

willkuer:
Can you please post the full emulog after the crash? (can be found in logs/emulog.txt in the directory of pcsx2.exe)

FlatOutPS2:
Related to this: http://forums.pcsx2.net/Thread-Gran-Turismo-4-crash-in-OpenGL-Mode?page=2
And #1314: There was a regression in git builds 650/651, that made it more likely to crash than usual.

The log messages are because you probably haven't been to the bios(boot with no disc selected) to set the time(zone).

gregory38:
Yes, it is expected. And it will remain like that until I (or someone that beat me) properly fix the handling of the bad effect.

MarcoEstevez:
Great, Ill try to set up bios as you said, I just tried different ones yesterday, so most probably is that.

I will also re-test with @gregory38 fix as soon as I get home.

MarcoEstevez:
Same with latest build:

LOG: https://drive.google.com/file/d/0ByJrCj2X0ZpAQXhjTHNDczBXb3c/view?usp=sharing

[...]

PR #2717, opened 2018-11-26
GSdx: Adjust Large Framebuffer

lightningterror:
Make "Large Framebuffer" disabled by default, we are getting a lot of
people reporting the issue of system or application hangs caused by
large ram/vram spikes. It's not safe so better disable it by default,
also move it to HW hacks gui as well.

Windows: move "Accurate Date" where "Large Framebuffer" used to be, also
move Blending Accuracy a little bit higher.

And a bit of eyecandy for windows:
image

Linux:
https://cdn.discordapp.com/attachments/350645867876712449/516835449126387723/unknown.png
https://cdn.discordapp.com/attachments/350645867876712449/516835832271732736/unknown.png

Related:
https://github.com/PCSX2/pcsx2/issues/2687
https://github.com/PCSX2/pcsx2/issues/1313


plugins/GSdx/Window/GSLinuxDialog.cpp
```diff
@@ -304,7 +302,7 @@ void populate_hw_table(GtkWidget* hw_table)
 
 	s_table_line = 0;
 	InsertWidgetInTable(hw_table , paltex_check  , acc_date_check);
-	InsertWidgetInTable(hw_table , large_fb_check, hack_enable_check);
+	InsertWidgetInTable(hw_table , hack_enable_check);
```
lightningterror: I forgot how linux gui looks so this might need adjusting. A screenshot will be nice to see.

plugins/GSdx/GSdx.rc
```diff
@@ -317,11 +317,11 @@ BEGIN
     LTEXT           "CRC Hack Level:",IDC_CRC_LEVEL_TEXT,22,224,79,8
     COMBOBOX        IDC_CRC_LEVEL,105,222,127,118,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
     CONTROL         "Enable HW Hacks",IDC_HACKS_ENABLED,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,11,240,71,10
-    PUSHBUTTON      "Advanced Settings and Hacks",IDC_HACKSBUTTON,105,238,127,14
+    PUSHBUTTON      "Advanced Settings and Hacks",IDC_HACKSBUTTON,105,237,127,14
     // OpenGL (Hardware) Settings:
-    CONTROL         "Accurate Date",IDC_ACCURATE_DATE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,11,256,72,8
-    LTEXT           "Blending Unit Accuracy:",IDC_ACCURATE_BLEND_UNIT_TEXT,22,272,79,10
-    COMBOBOX        IDC_ACCURATE_BLEND_UNIT,105,270,127,118,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
+    CONTROL         "Accurate Date",IDC_ACCURATE_DATE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,105,144,72,8
```
lightningterror: Maybe move it below Allow 8-bit textures to better match ?

lightningterror:
Updated the gui a bit more, shrank it a bit as well and now it should be good to go.

ssakash:
But large framebuffer does fix something right? What about such scenarios? And did you try porting the base buffer size used in custom resolution and check if it fixes the issue while also having the benefits of large framebuffer?

lightningterror:
It was discussed with Akash and Gregory that it would be better to fix the feature properly instead so I'll close this one, will update the tooltip later.

PR #2718, opened 2018-11-27
GSdx-ogl: Implement a fast Accurate Date algorithm

### LINKED ISSUE #2068: Image_Load_Store graphical glitches (GT3 and GT4)
***PCSX2 version:***
v1.5.0-dev-2194-gada0d5745, but also happens on earlier GIT builds, stable 1.4.0, and earlier.

- **Notes**:
  Got worse with build v1.5.0-dev-2184-ga0aa585af, but existed before that. If necessary, I can submit a GS Dump, but it happens throughout gameplay, rather than when something suddenly draws into view.

PCSX2 options:
EE Round: Chop/Zero
EE Clamping: Full
VU Round: Chop/Zero
VU Clamping: Extra + Preserve Sign
Frame limiting and Frame skipping disabled
Vsync setting: Forced on thru Nvidia Control Panel, set to Adaptive in PCSX2
EE Cyclerate: Default
VU Cycle Stealing: None
INTC Spin Detection: Enabled
Wait Loop Detection: Enabled
Fast CDVD: Enabled
mVU Flag Hack: Enabled
MTVU: Enabled
Manual Game Fixes: Disabled

Plugins used:

GSdx Settings:
Adapter: NVIDIA GeForce GTX 1080
Renderer: OpenGL (Hardware)
Interlacing: Bob tff
Texture Filtering: Bilinear (PS2)
Allow 8-bit textures: Enabled
Large Framebuffer: Enabled (issue persists regardless of setting)
Internal Resolution: 3x Native
Mipmapping (Ins): Basic
CRC Hack Level: Partial (OpenGL Recommended)
HW Hacks: Disabled
Accurate Date: Enabled
Blending Unit Accuracy (High)

SPU-2X Settings:
Interpolation: Catmull-Rom
Disable Effects Processing: Disabled
Use de-alias filter: Disabled
Debug Options: Disabled
Module: XAudio2
Volume: 100%
Latency: 30ms
Synchronizing Mode: Async Mix
Audio Expansion Mode: Stereo (none, default)
Use a Winamp DSP Plugin: Disabled

Car bodies shimmer/flicker during races (some track textures in GT4 as well, like the roads in Opera Paris). This issue has existed ever since I started using PCSX2, but got worse with the build mentioned above in the notes. Disabling Image_Load_Store in the HW Hacks fixes the problem, but makes the game run noticeably slower in places that would already slow down, and it would slow down in places it wouldn't normally do so. I know that this is because Image_Load_Store speeds things up, but it also causes graphical glitches.

Use settings above and make sure Image_Load_Store, Accurate Date, and High-Accuracy Blending are enabled. Then load any race in GT3 or GT4. Most noticeable when using chase cam or roof/hood cam.

None

***PC specifications:***
CPU: Intel Core i7 4790K (4.4 GHz)
RAM: 32 GB DDR3 1600 MHz
GPU: Nvidia GeForce GTX 1080
Boot SSD: Samsung 950 Pro 256 GB (PCSX2 loads off of this drive)
Mass Storage 1: WD Blue 1 TB 7200 RPM (games, inis, and memory cards load off of this).
Windows 10 Pro 64-bit, Creators Update

lightningterror:
Suggested by @gregory38 
Change the checkbox to a drop down list Off, Fast, Full.
Off and Full do the same as before, the Fast option is new. Most of the
time users don't actually need full Accurate Date so why not reduce that
cost and use a fast mode instead, plus it can also be used on a gpu with
a limited driver. Also change the default value to Fast instead of Off.

image

Fixes #2068


plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
```diff
@@ -1131,15 +1131,28 @@ void GSRendererOGL::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
...
-					DATE_GL42 = true;
-				} else {
-					m_require_full_barrier = true;
-					DATE_GL45 = true;
-					DATE = false;
+			} else {
+				switch (m_accurate_date)
+				{
+					case ACC_DATE_FULL:
+						GL_PERF("Accurate DATE with alpha %d-%d", m_vt.m_alpha.min, m_vt.m_alpha.max);
```
lightningterror: Perhaps change this to Full Accurate DATE.

lightningterror:
I moved around some stuff on linux to better match windows gui.

plugins/GSdx/Window/GSLinuxDialog.cpp
```diff
@@ -279,40 +283,35 @@ void populate_hw_table(GtkWidget* hw_table)
...
 	AddTooltip(mipmap_label, IDC_MIPMAP_HW);
 	AddTooltip(mipmap_combo_box, IDC_MIPMAP_HW);
+	AddTooltip(acc_date_label, acc_date_combo_box, IDC_ACCURATE_DATE);
+	AddTooltip(acc_bld_label, acc_bld_combo_box, IDC_ACCURATE_BLEND_UNIT);
 
 	s_table_line = 0;
-	InsertWidgetInTable(hw_table , paltex_check);
-	InsertWidgetInTable(hw_table , large_fb_check, hack_enable_check);
+	InsertWidgetInTable(hw_table , paltex_check , large_fb_check);
+	InsertWidgetInTable(hw_table , hack_enable_check);
```
lightningterror: Maybe it can be moved at the bottom instead, because there's empty space of the missing checkbox now. Maybe at the bottom will look less weird but on the other hand all checkboxes are here.

lightningterror:
I made some final adjustments so it should be ready to go.
@gregory38 I think you'll be happy with `Full Accurate DATE`, `Fast Accurate DATE`, and `Inaccurate DATE` logs :)

lightningterror:
I did a final check and the option works fine, tooltips seem good as well, as for adding a tooltop on the label I won't do that because if we do one we gotta do all and I'd rather if we keep things consistent.

On a final note this is how the gui looks under linux:
image

And Windows:
image


Let's merge it and let others play around with the setting. :)
I'm done nitpicking small details.

orbea:
@lightningterror Just out of curiosity, what would be some cases where `Fast` is not enough and `Full` is required?

orbea:
Okay, that helps explain, thanks!

PR #2719, opened 2018-11-27
SPU2-X: GUI - Logo update

Ryudo300:
Replaced the old logo to match up with the ones we use in GSdx, etc.


MrCK1:
+9KB? I don't know if I have enough space for this...

lightningterror:
I believe @turtleli will have ideas what to do with the image and to reduce the size.
dll is quite a bit larger now.

PR #2723, opened 2018-11-28
GSdx-d3d: Partial port for DATE one (Fast Date) and Fast Accurate Date

lightningterror:
It fixes a bunch of shadow/transparency issues.
Fast DATE doesn't rely on the gui option and is always on.
Confirmed fixed issues: Persona 3 shadows on d3d11,
Digital Devil Saga Transparency d3d9/11.

Fast accurate date works the same/similar to OpenGL.
Confirmed fixed issues: DBZ BT3 ground shadows, Fifa Street 1 shadows
on all d3d renders as well.

Also this option doesn't cause other transparency issues like the
Alpha Stencil hack.

Note: If Alpha Stencil is enabled Fast Date and Fast Accurate Date will
be disabled.

Note2: Full Accurate Date is not implemented so the code fallbacks to
Fast mode instead.

Commits:
https://github.com/PCSX2/pcsx2/commit/3ab12cef2f2010372e9625759b5bfb8004662e37
https://github.com/PCSX2/pcsx2/commit/584397a3fda9ea8afa39885b37b89c0dbe2047b6

This will probably be the last feature d3d9 gets before getting purged.


lightningterror:
Ok let's merge this. The games I tested and the dumps from Gregory indicate that everything seems to be working peachy. Games that utilize fast DATE or fast accurate DATE look the same as gl.

Time for the rest of the users to dig up on the early Christmas presents.

MrCK1:
It's not exactly a Christmas present when you already have a GPU that actually functions 😏

PR #2730, opened 2018-12-01
GSdx: OSD Enhancements

### LINKED ISSUE #1585: [Feature Request] Savestate "save successful/load successful" dialogs.
Optional of course as I can see some people wanting to have a clean, no message, no bother experience, but I would prefer to have a little message in the corner of my screen that says something along the lines of "save slot 1 saved" or "save slot 1 loaded" display when you save/loaded a savestate.

There has been a few time in my experience when I hit the save hotkey to save only to have it not actually save for some reason, so when I went to load it, what I thought was my last save, well it was much further back then I had thought it was and lost quite a bit of progress, oh well, but a message here would just let you know everything went okay.

tadanokojin:
An OSD implementation for DX11 and some QoL updates for OGL.

DX11 implementation is D2D and DirectWrite.
Tested working on Windows 7 (works with platform update or logs a message without) and Windows 10. 

Here's DX11:
dx11

And OGL:
ogl

Updated Settings:
dialog

Users can select from a predefined list of colors, or manually edit the config:
```
osd_color_b=0
osd_color_g=0
osd_color_r=255
osd_color_opacity=100
```

Color samples:
sample

Obviously there's still work to be done on this but I think it's time I got eyes on it. 
I look forward to the review.

Fixes #1585


plugins/GSdx/Renderers/DX11/GSDevice11.h
```diff
@@ -231,5 +231,7 @@ class GSDevice11 : public GSDeviceDX
 	void CompileShader(const char* source, size_t size, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11GeometryShader** gs);
 	void CompileShader(const char* source, size_t size, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11GeometryShader** gs, D3D11_SO_DECLARATION_ENTRY* layout, int count);
 	void CompileShader(const char* source, size_t size, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11PixelShader** ps);
+
+	void GSDevice11::RenderOsd();
```
tadanokojin: This is unused, I'll remove it.

plugins/GSdx/Renderers/DX11/GSDevice11.cpp
```diff
@@ -44,6 +45,7 @@ GSDevice11::GSDevice11()
 
 GSDevice11::~GSDevice11()
 {
+
```
lightningterror: Useless row.

gregory38:
All commits should compile (both linux and windows / cmake and VS). So you will need to shuffle a bit the commit order.

plugins/GSdx/Renderers/DX11/GSOsdManagerDX11.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "Renderers/Common/GSOsdManager.h"
+
+class GSOsdManagerDX11 final : public GSOsdManager {
+
+	struct log_info {
```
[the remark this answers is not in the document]
tadanokojin: u32string ogl vs wstring for dx11

plugins/GSdx/Renderers/DX11/GSDevice11.cpp
```diff
@@ -431,6 +486,30 @@ bool GSDevice11::Reset(int w, int h)
 			return false;
 		}
 
+		D2D1_BITMAP_PROPERTIES1 bitmap_props;
+
+		memset( &bitmap_props, 0, sizeof(bitmap_props) );
+
+		bitmap_props.pixelFormat.format = scd.BufferDesc.Format;
+		bitmap_props.pixelFormat.alphaMode = D2D1_ALPHA_MODE_IGNORE;
+		bitmap_props.dpiX = 96.0f;
```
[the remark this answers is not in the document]
tadanokojin: Yes though perhaps hi-dpi might be blurry.

plugins/GSdx/Window/GSSettingsDlg.cpp
```diff
@@ -926,18 +940,32 @@ void GSOSDDlg::OnInit()
...
+
+	// Font
+	EnableWindow(GetDlgItem(m_hWnd, IDC_OSD_COLOR),               monitor_enabled || log_enabled);
+	EnableWindow(GetDlgItem(m_hWnd, IDC_OSD_SIZE_EDIT),           monitor_enabled || log_enabled);
+	EnableWindow(GetDlgItem(m_hWnd, IDC_OSD_SIZE),                monitor_enabled || log_enabled);
+	EnableWindow(GetDlgItem(m_hWnd, IDC_OSD_OPACITY_SLIDER),      monitor_enabled || log_enabled);
+	EnableWindow(GetDlgItem(m_hWnd, IDC_OSD_OPACITY_AMOUNT_TEXT), monitor_enabled || log_enabled);
+	EnableWindow(GetDlgItem(m_hWnd, IDC_OSD_SIZE_TEXT),           monitor_enabled || log_enabled);
+	EnableWindow(GetDlgItem(m_hWnd, IDC_OSD_COLOR_TEXT),          monitor_enabled || log_enabled);
+	EnableWindow(GetDlgItem(m_hWnd, IDC_OSD_OPACITY_TEXT),        monitor_enabled || log_enabled);
```
turtleli: Not a fan of aligning stuff.

plugins/GSdx/Renderers/DX11/GSOsdManagerDX11.cpp
```diff
@@ -0,0 +1,266 @@
...
+		L"en-US", // TODO
+		&m_osd_text_format
+	);
+}
+
+void GSOsdManagerDX11::Log(const char *utf8, uint32 color)
+{
+	std::wostringstream s;
+
+	s << utf8;
```
turtleli: I don't think this will convert UTF-8 characters properly?

plugins/GSdx/Renderers/DX11/GSOsdManagerDX11.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+	hr = dev->QueryInterface(IID_PPV_ARGS(&dxgid));
+
+	hr = DWriteCreateFactory(
+		DWRITE_FACTORY_TYPE_SHARED,
+		__uuidof(IDWriteFactory),
+		reinterpret_cast<IUnknown **>(&m_dwrite_factory)
+	);
+
+	hr = d2d_factory->CreateDevice(dxgid, &m_dev_2d);
```
turtleli: `hr` is assigned but not used multiple times.

plugins/GSdx/Renderers/DX11/GSOsdManagerDX11.cpp
```diff
@@ -0,0 +1,266 @@
...
+	if (!m_log_enabled && !m_monitor_enabled)
+		return;
+
+	HRESULT hr = E_FAIL;
+
+	D2D1_FACTORY_OPTIONS d2d_factory_options = {};
+
+	#ifdef DEBUG
+		d2d_factory_options.debugLevel = D2D1_DEBUG_LEVEL_INFORMATION;
+	#endif
```
turtleli: I feel this should be indented less.

plugins/GSdx/Renderers/Common/GSOsdManager.h
```diff
@@ -1,98 +1,39 @@
...
- *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
- *  GNU General Public License for more details.
- *
- *  You should have received a copy of the GNU General Public License
- *  along with GNU Make; see the file COPYING.  If not, write to
- *  the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA USA.
- *  http://www.gnu.org/copyleft/gpl.html
- *
- */
-
```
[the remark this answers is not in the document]
tadanokojin: My concern this this code has moved from OsdManager to OsdManagerOGL. I think the code should stay with the author.

That said, if you think otherwise I'll restore it.

plugins/GSdx/Window/GSSettingsDlg.cpp
```diff
@@ -902,20 +902,34 @@ void GSOSDDlg::OnInit()
...
-	SendMessage(GetDlgItem(m_hWnd, IDC_OSD_SPEED), UDM_SETRANGE, 0, MAKELPARAM(10, 2));
-	SendMessage(GetDlgItem(m_hWnd, IDC_OSD_SPEED), UDM_SETPOS, 0, MAKELPARAM(theApp.GetConfigI("osd_log_speed"), 0));
+	SendMessage(GetDlgItem(m_hWnd, IDC_OSD_TIMEOUT), UDM_SETRANGE, 0, MAKELPARAM(10, 2));
+	SendMessage(GetDlgItem(m_hWnd, IDC_OSD_TIMEOUT), UDM_SETPOS, 0, MAKELPARAM(theApp.GetConfigI("osd_log_timeout"), 0));
 
 	SendMessage(GetDlgItem(m_hWnd, IDC_OSD_MAX_LOG), UDM_SETRANGE, 0, MAKELPARAM(20, 1));
 	SendMessage(GetDlgItem(m_hWnd, IDC_OSD_MAX_LOG), UDM_SETPOS, 0, MAKELPARAM(theApp.GetConfigI("osd_max_log_messages"), 0));
 
-	SendMessage(GetDlgItem(m_hWnd, IDC_OSD_FONT_EDIT), WM_SETTEXT, 0, (LPARAM)theApp.GetConfigS("osd_fontname").c_str());
+	std::vector<GSSetting> color;
```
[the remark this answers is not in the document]
lightningterror: I like what arcum did with the linux gui for the colors, maybe it's better to get rid of that drop down list along with the custom button and just use rgb sliders here as well.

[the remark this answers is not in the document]
tadanokojin: That's a possible solution as well.

plugins/GSdx/Renderers/DX11/GSOsdManagerDX11.cpp
tadanokojin: I did an early return on failure for those.

plugins/GSdx/Renderers/Common/GSOsdManager.h
```diff
@@ -1,98 +1,39 @@
...
+	virtual void Monitor(const char *key, const char *value, uint32 color) { };
 
-	bool m_texture_dirty;
-	void upload_texture_atlas(GSTexture* t);
+	void SetSize(int x, int y);
 
-	void Log(const char *utf8, uint32 color);
-	void Monitor(const char *key, const char *value, uint32 color);
+	virtual void Release() { };
+#ifdef WIN32
```
lightningterror: Shouldn't it be `_WIN32 `instead ? Or it's the same.

plugins/GSdx/Renderers/Common/GSOsdManager.h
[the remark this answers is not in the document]
lightningterror: Since the code was moved the copyright header author can be changed to.

` *	Copyright (C) 2018 PCSX2 Dev Team`

plugins/GSdx/Renderers/DX11/GSOsdManagerDX11.cpp
```diff
@@ -0,0 +1,294 @@
+/*
+ *	Copyright (C) 2014-2018 PCSX2 Dev Team
```
lightningterror: I think just 2018 is fine since the file is created now.

plugins/GSdx/Renderers/DX11/GSOsdManagerDX11.cpp
```diff
@@ -0,0 +1,294 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with GNU Make; see the file COPYING.  If not, write to
+ *  the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA USA.
+ *  http://www.gnu.org/copyleft/gpl.html
+ *
+ */
```
lightningterror: I prefer adding an extra space between the header and the includes.

plugins/GSdx/Renderers/DX11/GSOsdManagerDX11.h
```diff
@@ -0,0 +1,52 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with GNU Make; see the file COPYING.  If not, write to
+ *  the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA USA.
+ *  http://www.gnu.org/copyleft/gpl.html
+ *
+ */
```
lightningterror: Ditto.

plugins/GSdx/Renderers/DX11/GSOsdManagerDX11.cpp
tadanokojin: A quick search and it seems `mbstowcs` is what needs to be done here.

If it's sounds correct to you I can write an implementation that uses `mbstowcs` with a `wchar` buffer of 256 in similar fashion to OGL.

willkuer:
Optional but I think you should expose the hex color code if you allow font color to span the whole space.

If I was unhappy with the OSD color I would look in the web for eye friendly colors that yield higher contrast. Based on the color selection I’d get a color hexcode (as most color pickers in the web are for intended to be used for ui design). To put this hex code via slider into this dialog would be annoying.

plugins/GSdx/Renderers/DX11/GSOsdManagerDX11.cpp
turtleli: Maybe use `convert_utf8_to_utf16` from common/include/PluginCompatibility.h (header file should already be included). I don't think mbstowcs will work for UTF-8 conversion on Windows .

tadanokojin:
> Was the way the OSD was implemented on OpenGL unsuitable for some reason?

As I've found with channel shuffle and depth, porting things between GL and DX is a painful experience and OSD has been no exception as I couldn't get the GL implementation to do much at all on DX. Based on the fact that it's been 2 years and no one has done so (either out of disinterest or because it was too difficult) I would say I'm not alone.

That left me with a few options:
1. Muck in the GL OSD code, possibly modifying it for one renderer and breaking it for another.
2. Use a third party library that was specific to DX (this was out of the question in my view)
3. D2D/DX10 interoperability with DWrite.

Yes, both OSD approaches have issues. For DX, the Fraps issue is a bit odd because it varies by some factors that are unknown at this time. For example it works fine on mine and KrossX' PC but doesn't on LT's. Considering that I think it can be fixed in time.

This solution means the GL implementation can largely be left alone (minus a few improvements) and the DX version can benefit from cleaner and clearer code as well as the benefits that come with D2D/DWrite.

Issues or not, this is better than the nothing has existed for years, at least in my view.
And of course it can always change later down the road.

tadanokojin:
Speaking with @gregory38 he believes it should be possible to port OGL OSD, I'll give it another shot with his suggestions and if I succeed I'll do a new PR with the UI changes ported.

gregory38:
We managed to find the root cause of the issue. So back to openGL way.

PR #2731, opened 2018-12-02
GSdx: Palette management refactor and fix (v2)

### LINKED ISSUE #2702: Tales of Legendia - Segfaults with CRC level 'None' or 'Minimum'

### LINKED ISSUE #2736: OpenGL regression in Burnout 3 (draw distance is black)
***PCSX2 version:***
v1.5.0-dev-2714

***PCSX2 options:***
All defaults.

***Plugins used:***
All defaults. 

In the OpenGL Hardware plugin, there is a new issue where geometry in the distance is completely black. 

Here is a screenshot: 
image

It is pretty evident in any race. That race in the screenshot is Silver Lake. 

Seems to work fine in dev-2484.

***PC specifications:***
Tested on Windows 7/i7-3770/GTX 1050Ti and Windows 10/i5-7200U/GT 960M, both affected

AlessandroVetere:
Refactor and fix palette management.

1. Improve ICO hack by using palette cache for depth sampling, removing only point of creation of palette textures outside TC (removes conditional deletion of `Source::m_palette` in `~Source()`),
2. Purge comments,
3. Reformat parts of the code,
4. Centralize CLUT comparison logic in `PaletteKey` equality operator,
5. Fix and rationalize palette texture or clut copy attachment and update with respect to pre-`PaletteMap` behavior,
6. Make `PaletteMap` capable of managing at the same time Palette with and without `GSTexture`,
7. Handle possible case of `Source` lookup hit with `psm.pal` value changed.

Closes #2702 , Closes #2736 and prevents possible similar regressions.

### DIFF
--- a/plugins/GSdx/Renderers/Common/GSTextureCache.cpp
+++ b/plugins/GSdx/Renderers/Common/GSTextureCache.cpp
@@ -224,10 +224,9 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 		// Target are converted (AEM & palette) on the fly by the GPU. They don't need extra check
 		if (!s->m_target) {
 			// We request a palette texture (psm_s.pal). If the texture was
-			// converted by the CPU (!s->m_should_have_tex_palette), we need to ensure
+			// converted by the CPU (!s->m_palette), we need to ensure
 			// palette content is the same.
-			// Note: content of the palette will be uploaded at the end of the function
-			if (psm_s.pal > 0 && !s->m_should_have_tex_palette && !s->ClutMatch({ clut, psm_s.pal }))
+			if (psm_s.pal > 0 && !s->m_palette && !s->ClutMatch({ clut, psm_s.pal }))
 				continue;
 
 			// We request a 24/16 bit RGBA texture. Alpha expansion was done by
@@ -378,6 +377,8 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 		}
 	}
 
+	bool new_source = false;
+
 	if(src == NULL)
 	{
 #ifdef ENABLE_OGL_DEBUG
@@ -390,6 +391,7 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 		}
 #endif
 		src = CreateSource(TEX0, TEXA, dst, half_right, x_offset, y_offset);
+		new_source = true;
 
 	} else {
 		GL_CACHE("TC: src hit: %d (0x%x, 0x%x, %s)",
@@ -398,7 +400,7 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 					psm_str(TEX0.PSM));
 	}
 
-	if (src->m_should_have_tex_palette && (!src->m_clut || !src->ClutMatch({ clut, psm_s.pal }))) {
+	if (src->m_palette && !new_source && !src->ClutMatch({ clut, psm_s.pal })) {
 		AttachPaletteToSource(src, psm_s.pal, true);
 	}
 
@@ -1335,7 +1337,7 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
 		// However it is different here. We want to reuse a Render Target as a texture.
 		// Because the texture is already on the GPU, CPU can't convert it.
 		if (psm.pal > 0) {
-			src->m_should_have_tex_palette = true;
+			AttachPaletteToSource(src, psm.pal, true);
 		}
 		// Disable linear filtering for various GS post-processing effect
 		// 1/ Palette is used to interpret the alpha channel of the RT as an index.
@@ -1449,11 +1451,13 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
 		if (m_paltex && psm.pal > 0)
 		{
 			src->m_texture = m_renderer->m_dev->CreateTexture(tw, th, Get8bitFormat());
-			src->m_should_have_tex_palette = true;
+			AttachPaletteToSource(src, psm.pal, true);
 		}
 		else {
 			src->m_texture = m_renderer->m_dev->CreateTexture(tw, th);
-			AttachPaletteToSource(src, psm.pal, false); // Attach only Palette object and clut copy
+			if (psm.pal > 0) {
+				AttachPaletteToSource(src, psm.pal, false);
+			}
 		}
 	}
 
@@ -1549,7 +1553,6 @@ void GSTextureCache::Surface::UpdateAge()
 GSTextureCache::Source::Source(GSRenderer* r, const GIFRegTEX0& TEX0, const GIFRegTEXA& TEXA, uint8* temp, bool dummy_container)
 	: Surface(r, temp)
 	, m_palette_obj(nullptr)
-	, m_should_have_tex_palette(false)
 	, m_palette(nullptr)
 	, m_target(false)
 	, m_complete(false)
--- a/plugins/GSdx/Renderers/Common/GSTextureCache.cpp
+++ b/plugins/GSdx/Renderers/Common/GSTextureCache.cpp
@@ -1173,6 +1173,11 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
 		src->m_target = true;
 		src->m_from_target = dst->m_texture;
 		src->m_texture->SetScale(scale);
+
+		if (psm.pal > 0) {
+			// Attach palette for GPU texture conversion
+			AttachPaletteToSource(src, psm.pal, true);
+		}
 	}
 	else if (dst)
 	{
--- a/plugins/GSdx/Renderers/Common/GSTextureCache.cpp
+++ b/plugins/GSdx/Renderers/Common/GSTextureCache.cpp
@@ -224,10 +224,10 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 		// Target are converted (AEM & palette) on the fly by the GPU. They don't need extra check
 		if (!s->m_target) {
 			// We request a palette texture (psm_s.pal). If the texture was
-			// converted by the CPU (s->m_palette == NULL), we need to ensure
+			// converted by the CPU (!s->m_should_have_tex_palette), we need to ensure
 			// palette content is the same.
 			// Note: content of the palette will be uploaded at the end of the function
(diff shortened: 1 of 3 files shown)


plugins/GSdx/Renderers/Common/GSTextureCache.cpp
```diff
@@ -1451,13 +1454,13 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
...
 			src->m_texture = m_renderer->m_dev->CreateTexture(tw, th, Get8bitFormat());
-			src->m_should_have_tex_palette = true;
-			// Palette (Palette object, clut copy and palette texture) is going to be attached in Source lookup code exploiting initial condition !src->m_clut
+			AttachPaletteToSource(src, psm.pal, true);
 		}
 		else {
 			src->m_texture = m_renderer->m_dev->CreateTexture(tw, th);
-			// by default: src->m_should_have_tex_palette = false;
-			AttachPaletteToSource(src, psm.pal, false); // Attach only Palette object and clut copy
+			if (psm.pal > 0) {
```
AlessandroVetere: This change is most likely what fixed #2702, as we were malloc'in 0 bytes clut arrays without the `psm.pal > 0` check.

orbea: No, that change doesn't seem to affect `Tales of Legendia`.

AlessandroVetere: Oh strange. Could you please try again the last version of PR plugin to see if the regression is still fixed? I applied some modifications to make things easier to manage.

orbea: Yes, I applied the whole PR as a patch to master and both `Tales of Legendia` and `Tales of the Abyss` still seem fixed with and without `8 bits textures`.

plugins/GSdx/Renderers/Common/GSTextureCache.cpp
```diff
@@ -2116,11 +2116,13 @@ std::size_t GSTextureCache::PaletteKeyHash::operator()(const PaletteKey &key) co
...
-	ASSERT(lhs.pal == rhs.pal); // By design, each map SHOULD contain only PaletteKey with the same pal value
-	
-	uint16 pal = lhs.pal;
-	uint16 palette_size = pal * sizeof(uint32);
-	return GSVector4i::compare64(lhs.clut, rhs.clut, palette_size);
+	if (lhs.pal != rhs.pal) {
+		return false;
+	}
+
+	ASSERT((lhs.pal & 15) == 0);
```
gregory38: Why the assert ? It doesn't make any sense here

plugins/GSdx/Renderers/Common/GSTextureCache.cpp
```diff
@@ -1451,6 +1452,7 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
 		{
 			src->m_texture = m_renderer->m_dev->CreateTexture(tw, th, Get8bitFormat());
 			src->m_should_have_tex_palette = true;
+			AttachPaletteToSource(src, psm.pal, true);
```
gregory38: Not related to this line of code in particular.
Is there a possibilities that we attach the palette twice ?
We attach in Create Src and in LookUp Src.

Maybe it could be possible to avoid the compare64 check in lookup ? It isn't critical if current code is safe but potentially a bit "slower".

plugins/GSdx/Renderers/Common/GSTextureCache.cpp
```diff
@@ -2077,6 +2077,13 @@ GSTexture* GSTextureCache::Palette::GetPaletteGSTexture() {
 	return m_tex_palette;
 }
 
+void GSTextureCache::Palette::UploadClut() {
+	if (!m_tex_palette) {
+		m_tex_palette = m_renderer->m_dev->CreateTexture(256, 1);
+		m_tex_palette->Update(GSVector4i(0, 0, m_pal, 1), m_clut, m_pal * sizeof(m_clut[0]));
```
gregory38: Is it expected that UploadClut will upload only once ?

plugins/GSdx/Renderers/Common/GSTextureCache.cpp
gregory38: it feels like the last commit, isn't it ?

gregory38:
It is much better with small commits. Small comments but it is fine for me. It requires a couple of final tests. And it should be fine.

plugins/GSdx/Renderers/Common/GSTextureCache.cpp
AlessandroVetere: You are right, this commit alone introduces double palette attachment (as it invalidates the current fast path), but with the last commit the fast path is managed in an easier way and it always avoid double attachment in Create Src and LookUp Src.

plugins/GSdx/Renderers/Common/GSTextureCache.cpp
AlessandroVetere: Yes because the clut content for a given `Palette` is copied at object creation and never changed, so also the palette need only to be uploaded once (possibly lazy upload)

plugins/GSdx/Renderers/Common/GSTextureCache.cpp
AlessandroVetere: Yeah maybe you are right there is already the ASSERT in the `GSVector4i::compare64`, I can remove it if needed

gregory38: yes because this assert belong to compare64 not this function.

plugins/GSdx/Renderers/Common/GSTextureCache.cpp
gregory38: I think the best will be to do a single commit
* that attach palette in all cases / as soon as possible
* the bool new_source of latest commit (which I think is the fix to double attachment)

This way we have an atomic (and logical) commit.

plugins/GSdx/Renderers/Common/GSTextureCache.cpp
gregory38: In this case, UploadClut could have a better name. Create/Initialize (once) vs Upload (can be repeated). 

You can also reduce the size of the texture. 256 is too big for small m_pal.

AlessandroVetere: Ok for name. As for Tex size, IIRC I tried and creating smaller textures lead to various errors. Will test again.

plugins/GSdx/Renderers/Common/GSTextureCache.cpp
AlessandroVetere: I agree with that, the intermediate commit breaks fast path so it's better to squash them in one.

plugins/GSdx/Renderers/Common/GSTextureCache.cpp
AlessandroVetere: Yeah creating texture with 16 as width breaks the plugin (both dx and ogl), while for big pal (256) it is good (as hardcoded in there). However it would be an interesting enhancement to create smaller textures for this use case.

lightningterror:
The crash mentioned in https://github.com/PCSX2/pcsx2/issues/2702 is reproducible in variety of dumps/games tested, oddly enough the crash is reproducible if a game utilizes depth, now since dx11 got depth convert as well it's also reproducible there.

@gregory38 if the PR has small nitpicks to be taken care of I suggest we postpone them for another PR, this one is rather really important and I'd like to see it merged as fast as possible.

gregory38:
No need to rush man.

Code is fine for me. It can be merged.

lightningterror:
> No need to rush man.

I don't like crash regressions :smile:

PR #2735, opened 2018-12-05
pcsx2-git: Add gsdumpgui .vs folder to ignored files

tadanokojin:
VS generates this folder and it's ignored elsewhere so I figure we might as well ignore it in this case as well.


MrCK1:
Could you add the GSdx debug .txt files as well? I don't remember what they're called, I'll check later.

mirh:
`.vs/` [Might be](https://git-scm.com/docs/gitignore) it?

tadanokojin:
@mirh beat you to it with that forced push.

I suppose I could add the `/` at the end but I don't think it will matter (if a regular file is named `.vs` it will be ignored as well).

gregory38:
Why it isn't merged ?

PR #2745, opened 2018-12-11
GSdx-d3d: Add support for Channel Shuffle on Direct3D11

lightningterror:
Note the feature is still not yet complete, copy function needs to be
implemented (suggested by Gregory) but it can be done at a later date,
this still fixes a bunch of issues on various games even in it's current state.

Related: https://github.com/PCSX2/pcsx2/issues/1318

### DIFF
--- a/plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
+++ b/plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
@@ -174,6 +174,145 @@ void GSRendererDX11::EmulateTextureShuffleAndFbmask()
 	}
 }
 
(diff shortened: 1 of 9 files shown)


gregory38:
Please implement properly the blue/green effect. You just need to extend the ps constant buffer. It should be trivial

lightningterror:
I checked the code a couple of times and everything seems a ok, also Greg reviewed it several times so it should be good as well.

I think that's as much as we can do for the scope of this pr, time for users to get another early Christmas present.

I think it's also possible to include Urban Chaos and Tales of Abyss hle without full depth fmt implementation, that's what I'm working on now.

PR #2752, opened 2018-12-14
GSdx: OSD Enhancements v2

### LINKED ISSUE #1585: [Feature Request] Savestate "save successful/load successful" dialogs.

tadanokojin:
Continued from #2730 

OGL OSD ported to DX11 along with the other features present in the previous PR.

DX11

OGL

Config:

Fixes #1585


plugins/GSdx/res/convert.fx
```diff
@@ -389,6 +395,21 @@ PS_OUTPUT ps_main17(PS_INPUT input)
...
+	PS_OUTPUT output;
+	output.c = input.p;
+	return output;
+}
+
+PS_OUTPUT ps_main19(PS_INPUT input)
+{
+    PS_OUTPUT output;
+    output.c = input.c * float4(1.0, 1.0, 1.0, sample_c(input.t).r);
+    return output;
```
lightningterror: Tabs instead of spaces, also small nitpick but I prefer if there is an empty space between each line, same goes for above to keep things consistent.
```
PS_OUTPUT ps_main18(PS_INPUT input)
{
	PS_OUTPUT output;

	output.c = input.p;

	return output;
}
```

plugins/GSdx/Renderers/DX11/GSDevice11.h
```diff
@@ -93,7 +93,7 @@ class GSDevice11 : public GSDeviceDX
 	{
 		CComPtr<ID3D11InputLayout> il;
 		CComPtr<ID3D11VertexShader> vs;
-		CComPtr<ID3D11PixelShader> ps[18];
+		CComPtr<ID3D11PixelShader> ps[20];
```
lightningterror: You can use ShaderConvert_Count instead.

Blackbird88:
How are you even supposed to change the font? It never worked for me.

plugins/GSdx/Renderers/Common/GSOsdManager.cpp
```diff
@@ -229,32 +235,34 @@ void GSOsdManager::Monitor(const char *key, const char *value, uint32 color) {
 }
 
 void GSOsdManager::RenderGlyph(GSVertexPT1* dst, const glyph_info g, float x, float y, uint32 color) {
+
+
```
lightningterror: Useless extra rows ?

JasonBrownDeveloper:
Was there a reason you removed the color argument from the API?  The colors were supposed to match the messages in the log.

tadanokojin:
@JasonBrownDeveloper it seemed better to give people the option of color. For me personally, I felt the colors were dark and it's not really a common case you get an error or warning. It might be possible though, perhaps in the future we can override the user color when warning/error is displayed.

PR #2755, opened 2018-12-17
GSdx-d3d11: Frame Buffer Masking

lightningterror:


lightningterror:
Ops, wrong repo, this was supposed to be on my fork.

PR #2759, opened 2018-12-18
GSdx-d3d11: Redo destination format and alpha output

lightningterror:
Dfmt in texture shuffle function already picks the 16 format on slot 2
so it's better to let it call the shader instead of aout.

We can keep old aout code until FbMask emulation is added on d3d11. We
can purge aout along with alpha hack then.

### DIFF
--- a/plugins/GSdx/Renderers/DXCommon/GSRendererDX.cpp
+++ b/plugins/GSdx/Renderers/DXCommon/GSRendererDX.cpp
@@ -583,14 +583,16 @@ void GSRendererDX::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sourc
 	m_ps_sel.clr1 = m_om_bsel.IsCLR1();
 	m_ps_sel.fba = m_context->FBA.FBA;
 
+	// FIXME: Purge aout with AlphaHack when FbMask emulation is added.
 	if (m_ps_sel.shuffle)
 	{
 		m_ps_sel.aout = 0;
 	}
 	else
 	{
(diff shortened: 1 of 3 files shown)


plugins/GSdx/res/tfx.fx
```diff
@@ -1002,13 +1003,13 @@ PS_OUTPUT ps_main(PS_INPUT input)
...
 
-	if(PS_AOUT) // 16 bit output
+	if((PS_DFMT == FMT_16) || PS_AOUT) // 16 bit output
 	{
 		float a = 128.0f / 255; // alpha output will be 0x80
 
 		c.a = PS_FBA ? a : step(0.5, c.a) * a;
 	}
-	else if(PS_FBA)
+	else if((PS_DFMT == FMT_32) && PS_FBA)
```
lightningterror: I wasn't sure about this one but I added it just in case following how gl code works.

plugins/GSdx/Renderers/DXCommon/GSRendererDX.cpp
```diff
@@ -583,14 +583,16 @@ void GSRendererDX::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sourc
...
 
+	// FIXME: Purge aout with AlphaHack when FbMask emulation is added.
 	if (m_ps_sel.shuffle)
 	{
 		m_ps_sel.aout = 0;
 	}
 	else
 	{
-		m_ps_sel.aout = UserHacks_AlphaHack || m_context->FRAME.PSM == PSM_PSMCT16 || m_context->FRAME.PSM == PSM_PSMCT16S || (m_context->FRAME.FBMSK & 0xff000000) == 0x7f000000 ? 1 : 0;
+		m_ps_sel.aout = UserHacks_AlphaHack || (m_context->FRAME.FBMSK & 0xff000000) == 0x7f000000 ? 1 : 0;
```
ssakash: Redundant ternary operator

PR #2760, opened 2018-12-19
gsdx-d3d11: Enable HLSL debugging on development builds

tadanokojin:
Enables live debugging for D3D11 shader code.

### DIFF
--- a/plugins/GSdx/Renderers/DX11/GSDevice11.h
+++ b/plugins/GSdx/Renderers/DX11/GSDevice11.h
@@ -232,9 +232,11 @@ public:
 	operator ID3D11Device*() {return m_dev;}
 	operator ID3D11DeviceContext*() {return m_ctx;}
 
-	void CompileShader(const char* source, size_t size, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11VertexShader** vs, D3D11_INPUT_ELEMENT_DESC* layout, int count, ID3D11InputLayout** il);
-	void CompileShader(const char* source, size_t size, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11GeometryShader** gs);
-	void CompileShader(const char* source, size_t size, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11GeometryShader** gs, D3D11_SO_DECLARATION_ENTRY* layout, int count);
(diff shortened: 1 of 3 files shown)


gregory38:
Maybe create a get_compiler_flags ?

gregory38:
Yes it can surely be better. Maybe a void* can help too

plugins/GSdx/Renderers/DX11/GSDevice11.h
```diff
@@ -232,9 +232,11 @@ class GSDevice11 : public GSDeviceDX
...
-	void CompileShader(const char* source, size_t size, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11GeometryShader** gs, D3D11_SO_DECLARATION_ENTRY* layout, int count);
-	void CompileShader(const char* source, size_t size, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11PixelShader** ps);
+	void CreateShader(const char* source, size_t size, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11VertexShader** vs, D3D11_INPUT_ELEMENT_DESC* layout, int count, ID3D11InputLayout** il);
+	void CreateShader(const char* source, size_t size, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11GeometryShader** gs);
+	void CreateShader(const char* source, size_t size, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11GeometryShader** gs, D3D11_SO_DECLARATION_ENTRY* layout, int count);
+	void CreateShader(const char* source, size_t size, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11PixelShader** ps);
+
```
lightningterror: Empty row or intended ?

tadanokojin: Intended to separate the CreateShaders from the CompileShader.

tadanokojin:
@gregory38 how's this for a solution? I abstracted the compile code from the functions so it's only used in one place.

gregory38:
Do we use geometry shader with stream ?

It is better. Note on ogl. You have shader which are compiled and then linked to a program (or ProgramShader).

tadanokojin:
> Do we use geometry shader with stream ?

I'm no shader guy but do you mean like this?
https://github.com/PCSX2/pcsx2/blob/4a0697fda255fe3ed9643082726f23cc283539c3/plugins/GSdx/res/tfx.fx#L716

If so, yes.

PR #2762, opened 2018-12-19
GSdx-d3d: Purge Direct3D9 Render entirely

### LINKED ISSUE #1171: GSDX: Drop (or) keep D3D9 renderer ?

lightningterror:
Nothing else to say.

Close #1171


tadanokojin:
Press F to pay respect.

mirh:
Did 1.6 come earlier?

arcum42:
Unless there are going to be other DirectX renderers in the future, all the common DirectX files could probably be merged in with the DX11 ones.

lightningterror:
Ye I know. Well it's a job that can wait a bit :)

turtleli:
I can still see
https://github.com/PCSX2/pcsx2/blob/fce3918562fac3730619a5d8c89809fdf4ebedb1/plugins/GSdx/GSdx.cpp#L219
and
https://github.com/PCSX2/pcsx2/blob/fce3918562fac3730619a5d8c89809fdf4ebedb1/plugins/GSdx/GSdx.cpp#L310
in the master branch...

PR #2764, opened 2018-12-20
GSdx-d3d: Update EmulateZbuffer code to use bppz for depth masking

lightningterror:
This removes the old code for VS_BPPZ which was used in the older
EmulateZbuffer code and replace it with a VS constant buffer bit
DepthMask. This matches the gl code and behavior.

### DIFF
--- a/plugins/GSdx/Renderers/DXCommon/GSRendererDX.cpp
+++ b/plugins/GSdx/Renderers/DXCommon/GSRendererDX.cpp
@@ -116,10 +116,12 @@ void GSRendererDX::EmulateZbuffer()
 	else if (m_context->ZBUF.PSM == PSM_PSMZ24)
 	{
 		max_z = 0xFFFFFF;
+		m_vs_sel.bppz = 1;
 	}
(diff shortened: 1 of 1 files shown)


plugins/GSdx/Renderers/DXCommon/GSDeviceDX.h
```diff
@@ -39,11 +39,15 @@ class GSDeviceDX : public GSDevice
...
 
+		GSVector2i DepthMask;
+
 		struct VSConstantBuffer()
 		{
 			VertexScale = GSVector4::zero();
 			VertexOffset = GSVector4::zero();
 			Texture_Scale_Offset = GSVector4::zero();
+
+			DepthMask = GSVector2i(0);
```
lightningterror: I'm not sure if I should update the cb Update below.

AlessandroVetere: I think so, otherwise the update does not take into account the new member of the struct

PR #2767, opened 2018-12-21
GSdx: Get rid of remaining d3d9 code and some other updates

lightningterror:
There's a lot more to be done, ps1 compatibility gui needs to be updated as well.
Then we can drop the requirement for d3d9.dll in the common.props at the very least.


plugins/GSdx/vsprops/common.props
```diff
@@ -19,7 +19,7 @@
     </ClCompile>
     <Link>
       <AdditionalDependencies>d3d11.lib;d3d9.lib;dxgi.lib;dxguid.lib;winmm.lib;strmiids.lib;opengl32.lib;comsuppw.lib;comctl32.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
turtleli: You missed this line.

turtleli:
The header include should be removed as well.

lightningterror:
> The header include should be removed as well.

Yes I know, I wanted to make sure nothing else is using them as well before adding those changes.

> Unless there are going to be other DirectX renderers in the future, all the common DirectX files could probably be merged in with the DX11 ones.

I prefer if we do those later on.

gregory38:
Blending enum should be done right.

Dx11 (as ogl, unlike dx9) allow to have a 2nd source for blending source. There are some awful hack in TextureFX11 to replace SRC by SRC1 value. Instead the table should use the correct enum and hack must be removed.

gregory38:
It is fine for me. I don't know if @turtleli I have other comment.

DxCommon/Dx11 merge can be done on a separate pr (as it would be a mess)

plugins/GSdx/PSX/GPU.cpp
```diff
@@ -143,7 +143,7 @@ EXPORT_C_(int32) GPUopen(void* hWnd)
 	case GPURendererType::D3D11_SW: s_gpu = new GPURendererSW(new GSDevice11(), threads); break;
 #endif
 	case GPURendererType::NULL_Renderer: s_gpu = new GPURendererSW(new GSDeviceNull(), threads); break;
-	default: // Fall back to D3D9/null mode if unknown value is read. No one could possibly enter here anyway.
+	default: // Fall back to null mode if unknown value is read. No one could possibly enter here anyway.
```
turtleli: Comment and code don't match.

plugins/GSdx/PSX/GPUSettingsDlg.cpp
```diff
@@ -35,36 +35,8 @@ void GPUSettingsDlg::OnInit()
...
-
-					std::string str = format("%dx%d %dHz", mode.Width, mode.Height, mode.RefreshRate);
-
-					ComboBoxAppend(IDC_RESOLUTION, str.c_str(), (LPARAM)&m_modes.back(), w == mode.Width && h == mode.Height && hz == mode.RefreshRate);
-				}
-			}
-		}
-	}
+	// FIXME: Someone can fix this to use D3D11 or OGL Device, D3D9 is no longer supported.
+	ComboBoxAppend(IDC_RESOLUTION, "Please select...");
```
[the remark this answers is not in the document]
lightningterror: I'm not sure, I couldn't get any result, in the past maybe it did but right now ps1 compatibility is a bit broken. Entire mode barely works.

plugins/GSdx/PSX/GPU.cpp
lightningterror: Looks like I didn't check the below code.

PR #2769, opened 2018-12-22
gsdx-d3d11: Fix Depth FST

tadanokojin:
Texture scaling for x/y values wasn't being set in all cases.
This ensures that it is to fix a bug in finding nemo (and possibly others).

Current master:

After patch:

EDIT: Almost forgot, thank you greg for pointing out the bug.


gregory38:
Cool. Can be merged and need test of SMT

PR #2770, opened 2018-12-22
gsdx-d3d11: enable splitting of alpha in dumps

tadanokojin:
This allows d3d11 to dump textures in a similar fashion to OGL.

Previous behavior was to output the texture with alpha.
Instead, let's split it as outputting the texture with alpha can produce a completely transparent texture which is useless for debugging.

### DIFF
--- a/plugins/GSdx/Renderers/Common/GSTexture.h
+++ b/plugins/GSdx/Renderers/Common/GSTexture.h
@@ -47,7 +47,7 @@ public:
 	virtual bool Map(GSMap& m, const GSVector4i* r = NULL, int layer = 0) = 0;
 	virtual void Unmap() = 0;
 	virtual void GenerateMipmap() {}
-	virtual bool Save(const std::string& fn, bool dds = false) = 0;
+	virtual bool Save(const std::string& fn) = 0;
 	virtual uint32 GetID() { return 0; }
 
 	GSVector2 GetScale() const {return m_scale;}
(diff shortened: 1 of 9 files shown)


tadanokojin:
I updated it to use the OGL define.
With regards to `dds` I noticed TextureSW uses it

https://github.com/PCSX2/pcsx2/blob/dc2eed6ca1e4811048b1def95eec8f32fcbb96a1/plugins/GSdx/Renderers/SW/GSTextureSW.cpp#L88-L90

If I remove the check, TextureSW break if it's passed a depth texture?

gregory38:
Hum, is dds = true used somewhere. If yes does it impact the sw renderer.

tadanokojin:
This is the only place I see it:
https://github.com/PCSX2/pcsx2/blob/a020c98421698e5293f81419cd91089e8eede25c/plugins/GSdx/Renderers/HW/GSRendererHW.cpp#L942-L958

Doesn't appear SW uses it, I guess the check is just redundant.

gregory38:
Remove it. The only place this variable make sense is in sw renderer (I don't think sw has an internal format)  On the hw you can check the internal format of texture to know if it is a depth buffer.

tadanokojin:
Consider it done.

gregory38:
Fine for me. Don't hesitate to ping me, I easily forget stuff.

lightningterror:
Dump works fine both debug and release test so merging this.

PR #2779, opened 2018-12-26
DebugInterface: Display Vector Unit registers as floats

Souzooka:
This PR displays the values of the packed floats in the VU as floats in the debugger (using `%7.2f`), rather than the raw hexadecimal values. I'd like to provide a more precise display, but there is very limited space for text on the UI. That said, these values are likely more parse-able than just the hexadecimal values.

### DIFF
--- a/pcsx2/DebugTools/DebugInterface.cpp
+++ b/pcsx2/DebugTools/DebugInterface.cpp
@@ -37,7 +37,8 @@ R3000DebugInterface r3000Debug;
 #define strcasecmp stricmp
 #endif
 
-enum ReferenceIndexType {
+enum ReferenceIndexType
+{
 	REF_INDEX_PC = 32,
 	REF_INDEX_HI = 33,
 	REF_INDEX_LO = 34,
@@ -49,10 +50,11 @@ enum ReferenceIndexType {
 };
 
 
-class MipsExpressionFunctions: public IExpressionFunctions
+class MipsExpressionFunctions : public IExpressionFunctions
 {
 public:
-	MipsExpressionFunctions(DebugInterface* cpu): cpu(cpu) { };
+	MipsExpressionFunctions(DebugInterface* cpu)
+		: cpu(cpu){};
 
 	virtual bool parseReference(char* str, u64& referenceIndex)
 	{
@@ -92,7 +94,7 @@ public:
 	virtual bool parseSymbol(char* str, u64& symbolValue)
 	{
 		u32 value;
-		bool result = symbolMap.GetLabelValue(str,value);
+		bool result = symbolMap.GetLabelValue(str, value);
 		symbolValue = value;
 		return result;
 	}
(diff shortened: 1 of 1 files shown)


hch12907:
Should it handle the PS2 floats? (those with exponent >=128)

Souzooka:
I'm not familiar with that standard (and the FPU display doesn't seem to do any special handling of a different float standard).

ramapcsx2:
I like the idea. It's probably a good help for debugging game logic.
However, this will hide emulation bugs on, for example, direct floating point comparisons.
A HEX display would make precision issues more clear then.

Could you do this as a toggle switch option? :)

refractionpcsx2:
Also to note, values in the VF registers are not always floats, just something to be aware of.

lightningterror:
So what is the status ?

PSI-Rockin:
It makes sense to have a universal option saying "Display floats in hexadecimal" that acts for both COP1 and COP2. I'm thinking it would be best to place it in the context menu available via right-clicking and also bind a hotkey to it.

ramapcsx2:
Toggle for special formatting or not, I think we can start by merging this PR first.
It's apparently popular enough that 2 people implemented it separately :p 
https://github.com/PCSX2/pcsx2/pull/3010

lightningterror:
How about adding a Main Menu to the Debug Window, kinda like how the console has with a drop down list. Can be named Debug Options or something, this option can be in that list as a checkbox option.

monster860:
This one gets a bit fucky if you have numbers bigger than 9999.

ramapcsx2:
New menus need to be coded of course. It's not a problem of having them, but that someone needs to do the wxWidget stuff.

@monster860 Is it screwing up the display of the values, or the values itself? Is it still legible?

monster860:
If I recall correctly, the printf %7.2f doesn't limit the field to 7 characters it makes the field a minimum of 7 characters. This means if, say, you have a value of -1000000000 (this exact value was actually used in a game I was debugging, believe it or not) it would show up as:

```
vf01    12.34   23.45  532.24 1234.43
vf02   -1000000000.00   23.45  532.24 1234.43
vf03    12.34   23.45  532.24 1234.43
```

In this case, vf02.x would be completely invisible.

ramapcsx2:
Hm, maybe something can be done with the width parameter?
http://www.cplusplus.com/reference/cstdio/printf/
Otherwise, maybe make the GUI fields large enough?

lightningterror:
Formatting needs to be adjusted.

pcsx2/DebugTools/DebugInterface.cpp
```diff
@@ -477,6 +477,12 @@ wxString R5900DebugInterface::getRegisterString(int cat, int num)
 			sprintf(str,"%f",fpuRegs.fpr[num].f);
 			return wxString(str,wxConvUTF8);
 		}
+    case EECAT_VU0F:
+		{
+            char str[256];
+			sprintf(str, "%7.2f %7.2f %7.2f %7.2f", VU0.VF[num].F[0], VU0.VF[num].F[1], VU0.VF[num].F[2], VU0.VF[num].F[3]);
+            return wxString(str, wxConvUTF8);
```
lightningterror, line 484: Tabs instead of spaces, above changes too.

GovanifY:
I went ahead and rebased this to master, do we want to add anything to this PR (eg change the %7.2f, add a hex view and if so any preference?) or is it good to go?

VelocityRa:
Can this get merged please?

VelocityRa:
Actually, apparently you can't changed between hex & the new float view with this.
Not sure if this is acceptable considering these registers don't always have to contain floats.

PR #2784, opened 2018-12-28
GSdx: Adjust Aggressive state crc level.

lightningterror:
Create a Aggressive state function instead of adjusting the hacks
separate, move the aggressive only there, also use "else if" checks
instead. It's more efficient this way.

### DIFF
--- a/plugins/GSdx/Renderers/HW/GSHwHack.cpp
+++ b/plugins/GSdx/Renderers/HW/GSHwHack.cpp
@@ -27,9 +27,9 @@ bool s_nativeres;
 static CRCHackLevel s_crc_hack_level = CRCHackLevel::Full;
 
 // hacks
-#define Aggressive (s_crc_hack_level >= CRCHackLevel::Aggressive)
-#define Dx_only (s_crc_hack_level >= CRCHackLevel::Full)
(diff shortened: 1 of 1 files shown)


turtleli:
> also use "else if" checks instead

I don't think this is correct?

avih:
I think this patch hurts readability because now when you look at one of those functions you can't tell it's an aggressive hack, especially when aggressive is still used at other functions which have also non-aggressive hacks.

Performance wise, absolutely negligible.

I don't think it's worth it.

lightningterror:
> I think this patch hurts readability because now when you look at one of those functions you can't tell it's an aggressive hack, especially when aggressive is still used at other functions which have also non-aggressive hacks.

You could say the same for partial and full level then, anyway when I first open the file I check the states in the bottom to find out which group of category they fall in to, also the hack themselves are still separated and grouped.

> I don't think this is correct?

The description ? I could put statements instead.

avih:
> You could say the same for ...

I guess we see things differently.

lightningterror:
Ah you're right about that, I guess it slipped my mind.

lightningterror:
I meant as in this method is more efficient(better), since it doesn't call the games on aggressive state anymore on lower level when it does nothing.

PR #2789, opened 2018-12-30
3rdparty: Update pthreads to version 3.0.0

MrCK1:
This PR updates pthreads to the latest stable release that was release a few months ago. I did a full compile in my system and everything seems to check out OK.

I wasn't sure what files to keep/remove so please let me know as well.


3rdparty/pthreads4w/implement.h
```diff
@@ -38,9 +38,9 @@
 #if !defined(_IMPLEMENT_H)
 #define _IMPLEMENT_H
 
-#if !defined(PTW32_CONFIG_H)
-# error "config.h was not #included"
-#endif
+//#if !defined(PTW32_CONFIG_H)
+//# error "config.h was not #included"
+//#endif
```
tadanokojin: You should understand why this doesn't work instead of just commenting out code.

RedPanda4552:
UI does not launch on Devel configuration. Also throws an exception on exit on Release configuration.

tadanokojin:
I'd be interested in what everyone's thoughts are on this, but given the state of the repo/build system I don't think we should be updating libraries in giant PRs unless we know what we are doing and there is some benefit to doing so. This commit is massive and a large portion of it appears dedicated to renaming files or updating licenses.

To me, the solution is to update the windows build to use cmake as the linux build uses and allow devs to use a local version of pthreads.

3rdparty/pthreads4w/include/pthread.h
```diff
@@ -395,13 +395,26 @@ enum
...
+typedef struct ptw32_handle_t
 { void * p;                   /* Pointer to actual object */
 #if PTW32_VERSION_MAJOR > 2
   size_t x;                   /* Extra information - reuse count etc */
 #else
   unsigned int x;             /* Extra information - reuse count etc */
 #endif
+#ifdef __cplusplus
+// Added support for various operators so that the struct is
+// more pthreads-compliant in behavior. (air)
```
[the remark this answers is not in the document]
MrCK1: Done

3rdparty/pthreads4w/implement.h
MrCK1: I removed it to see if it would compile on here, same error unfortunately.

MrCK1: The issue is fixed now, forgot to update an preprocessor definition in the VS project file.

JohnHolmesII:
>To me, the solution is to update the windows build to use cmake

Yes!

>allow devs to use a local version of pthreads

No!

Recipe for disaster. Threads are a very sensitive system, and it makes more sense to me to have something that important be unified and stable across all systems. Don't need devs getting different results.

tadanokojin:
> As gregory mentioned yesterday, rewriting to use std::thread would probably ultimately be better.

Just adding that this is my view now as well.

With regards to locally installed libraries, as Arcum says, that train has already left the proverbial linux station.

gregory38:
For the record, semaphore and condition variable are rather close. 1 can be used to emulate the other. Issue is the speed impact.

At heart, I think that Linux uses semaphore and Win cond variable.

Gsdx jobs queue can be used as an example how to use cond variable. (I'm not sure it is the best way). But mtgs/mtvu uses ring buffers to communicate (atomic + semaphore) such as gsdx.

At the end of the day, even if it cost a bit of perf (hopefully very small), cross platform code (without 3rd party lib) is way nicer.

JohnHolmesII:
Indeed, it is fairly simple to write a sempahore based on std atomics. RPCS3 uses one that wraps an atomic int. Using std::thread makes a ton of sense.

MrCK1:
For now I guess this can wait until after 1.6, if we even deem these change to be anything worthwhile. Closing for now.

PR #2791, opened 2018-12-30
GSdx: Adjust dynamic crc hack

lightningterror:
Currently dynamic crc hack is only supported on windows so better move
the defines to windows only, also enable the dll path define only when
dynamic crc hack is enabled.
This also adjusts the dll location for dynamic crc hack to be the pcsx2
install dir instead of a different location that seemed a bit odd.
"tools" folder needs to be created to place the dll in.

I wasn't sure if I should use the tools folder or not.

### DIFF
--- a/plugins/GSdx/config.h
+++ b/plugins/GSdx/config.h
@@ -28,9 +28,6 @@
 
 #define EXTERNAL_SHADER_LOADING 1
 
-//#define ENABLE_DYNAMIC_CRC_HACK
-#define DYNA_DLL_PATH "c:/dev/pcsx2/trunk/tools/dynacrchack/DynaCrcHack.dll"
-
 //#define DISABLE_HW_TEXTURE_CACHE // Slow but fixes a lot of bugs
 
 //#define DISABLE_BITMASKING
@@ -54,4 +51,9 @@
 
 #ifdef _WIN32
 //#define ENABLE_OPENCL
+
+//#define ENABLE_DYNAMIC_CRC_HACK
+#ifdef ENABLE_DYNAMIC_CRC_HACK
+#define DYNA_DLL_PATH "tools/DynaCrcHack.dll"
+#endif
 #endif


avih:
It's not windows only, though it was only tested on windows. Besides, this PR adds exactly nothing.

The path is absolute intentionally, because it should point to an existing dir within the repo which includes its stuff. The way you modified the path makes it non obvious that stuff should be moved around, which is not good even if it was made obvious (because stuff has to be moved around, while with the original abs path to a repo dir it doesn't).

I say don't touch it. This PR really adds nothing and saves nothing.

mirh:
Ehrm... I don't want to be that guy
But hoping for #2412 to land soonish, wouldn't it be better if we moved dynacrchack on there?

I'm not saying the idea was wrong.. But execution (with the external compiler and all) seems lacking

avih:
How is #2412 even remotely related? besides, the whole point is that recompile of the dll is instant and it's loaded on the fly. You could of course use gcc or visual studio and it would work equally well, but tcc fits way better for the task at hand IMO, GCC comes second, and VS probably the least.

mirh:
Of course. 
I mean, maybe as a dev tool this was/is even used and appreciated (I dunno?)

I was referring that at least on the user side this didn't seem to get much traction.

gregory38:
Won't it be easier to use a txt file rather than a complex dll system. Well, it might limit the possibility of the hack. But honestly it is enough to skip few draw calls. If it isn't enough. Maybe people should improve gsdx in the first place.

lightningterror:
Are you fine with just adjusting DYNA_DLL_PATH to run when dynamicrc is enabled only ?

avih:
> Won't it be easier to use a txt file rather than a complex dll system

It's enough for some tasks, sure, but when one develops a CRC hack, it's way way nicer to use actual code and make it affect the game in real/run time.

If, at some point, CRC hacks would become config values, then sure, reloading the config would be more appropriate, but as long as the hacks are actual code, I think depevoping a hack should be done in code and not in config values.

avih:
> Are you fine with just adjusting DYNA_DLL_PATH to run when dynamicrc is enabled only ?

It doesn't add anything useful, therefore I'm against changes which don't help anything.

Really, what's the point of these useless changes? just do real stuff if you want to touch the code. Don't move bits around which affects nothing.

gregory38:
Ofc I'm a slow typer. Honestly on dev side, you can find bad draw call with a GPU debugger easily. Then dump GS data of the said draw call. This feature feel like a nuclear weapon to avoid few recompilations (dev guys) of the plugin. (I'm glad I never bother to port this code).

However I agree this PR is kinds of useless

avih:
> I was gonna suggest purging it

It's in its own dir, with very few LOC in actual PCSX2/GSdx which are `#ifdef` out anyway when it's disabled. With so very little impact on actual PCSX2's code, I'd purge it only if it was completely and utterly useless - which I think it's not.

gregory38:
Well I think we all have our way to solve issues. For hack, I use a gsdump for easy replay. And I skip draw call in GPU debugger to see the effect. Anyway it isn't important dyna crc can stay.

Nowadays gsdx is able to emulate most stuff correctly. You don't need tons of skip. So a basic text file (even for current hacks) would be user friendly.

avih:
> So a basic text file (even for current hacks) would be user friendly

I think this would be great. Being able to config hacks rather than coding them would be very nice IMHO. And also more friendly to reload-in-runtime.

However, I don't _think_ it's a simple thing to do. But maybe it is, definitely worth a shot IMO.

gregory38:
Note: don't have time to code it, and better spend time to fix rendering instead. But for the record.

It will be more limited than current code (it should be enough for 90% hack). Hypothesis is that we need few hack and few draw call need to be skipped (whereas older gsdx was just broken for most post processing effects)

But something like that. You store bad frame info in txt and skip condition (not 0, or 0 to triggers) +number of skip and why not skip_offset too.

You store them inside a vector. Then
For (auto hack : hack_txt) 
   If (hack == current_frame) {
       Skip = hack.skip;
       Return true;
  }

This way if will be equivalent to this pattern
If ....
Else if ..
Else if ..

PR #2792, opened 2018-12-31
Remove CDVDpeops and PeopsSPU2.

arcum42:
On checking, CDVDpeops and PeopsSPU2 were not in either the main or old plugin solution on the windows side, and neither were in the cmake build system on the Linux side, either.

As such, both seem safe candidates for deletion.

I also removed a few references to plugins that aren't actually in the trunk any more, as well.


arcum42:
No time like the present, then. We can always look back on the branches for previous releases is we need to reference them. ^_^

PR #2793, opened 2018-12-31
PCSX2: Update speed hack presets to be more friendly and usable

RedPanda4552:
Old presets touched parts such as VU clamping for what seems to be no good reason at all, and applied some rather dodgy combinations of EE cyclerate and cycle skipping. New presets are less destructive, but I am looking for feedback as to whether these remain helpful at all or should be revised.

~~Per Lightning's request, also made the clamping/rounding mode options unlock on presets 3-6. If this is unpopular (e.g. going against the purpose of having presets at all), this should be an easy reversion so a quick opinion poll will settle that.~~

### DIFF
--- a/pcsx2/gui/Dialogs/SysConfigDialog.cpp
+++ b/pcsx2/gui/Dialogs/SysConfigDialog.cpp
@@ -75,7 +75,7 @@ void Dialogs::SysConfigDialog::UpdateGuiForPreset ( int presetIndex, bool preset
 	//Console.WriteLn("Applying config to Gui: preset #%d, presets enabled: %s", presetIndex, presetsEnabled?"true":"false");
 
 	AppConfig preset = *g_Conf;
-	preset.IsOkApplyPreset( presetIndex );	//apply a preset to a copy of g_Conf.
+	preset.IsOkApplyPreset( presetIndex, false );	//apply a preset to a copy of g_Conf.
 	preset.EnablePresets = presetsEnabled;	//override IsOkApplyPreset (which always applies/enabled) to actual required state
 	
 	//update the config panels of SysConfigDialog to reflect the preset.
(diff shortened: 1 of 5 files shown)


pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -165,7 +165,7 @@ void Panels::GSWindowSettingsPanel::ApplyConfigToGui( AppConfig& configToApply,
 	}
 
 	m_combo_vsync->SetSelection(enum_cast(configToApply.EmuOptions.GS.VsyncEnable));
-	m_combo_vsync->Enable( !configToApply.EnablePresets );//grayed-out when presets are enabled
+	m_combo_vsync->Enable(true); // Always allow, regardless of preset
```
ramapcsx2: Good change.

ramapcsx2:
This seems like a good change set.
It updates the preset system to limit itself to the more sensible options and it seems to reduce clutter a little.
Do I read this right, and VU cycle steal isn't included anymore?
If so, this may be the one option that should still play a role (due to the drastic effect in some titles).

lightningterror:
My suggestion was not to include EE Cycle skipping since we are mostly focused on Aggressive speed as the presets are turned up.
Including it kinda defeats the purpose of what the states should be, otherwise people can still uncheck the presets for the compatibility with specific games that require it.

willkuer:
Sotc improves speed under vu cycle stealing, doesn’t it?
Also gs limited games should work better, or? I thought vu cycle stealing is something similar to a crazy deep sitting frameskipper...

RedPanda4552:
VU cycle stealing/EE cycle skipping, whichever verbage you want to use, gives the VUs more execution time relative to the EE. It helps SotC when on lower levels but on higher levels and the vast majority of games, it is destructive and only slows things down. I do support throwing it back on to one or two presets, but only at level 1. Combining even level 1 with an EE underclock, which several presets used to do, is as destructive to performance as you can get, next to enabling the EE interpreter and cache, which is why I felt the need to do this in the first place. Does anyone know if SotC also benefits from an EE overclock? If so what level?

lightningterror:
It's fine, you can rebase.

pcsx2/gui/Dialogs/SysConfigDialog.cpp
```diff
@@ -125,16 +125,16 @@ void Dialogs::SysConfigDialog::AddPresetsControl()
 		wxDefaultPosition, wxDefaultSize, wxHORIZONTAL /*| wxSL_AUTOTICKS | wxSL_LABELS */);
 	m_slider_presets->SetMinSize(wxSize(100,25));
 
+	const wchar_t* presetTooltip = L"Presets apply some speed hacks that may boost speed on underpowered systems, or speed up games that have unusual performance requirements. Uncheck this box to apply settings manually.\n\n1) Safest - No speed hacks. Most reliable, but possibly slow setting.\n2) Safe - Default. A few speed hacks known to provide boosts, with minimal to no side effects.\n3) Balanced - May help quad core CPUs.\n4) Aggressive - May help underpowered CPUs on less demanding games, but risks causing problems in other cases.\n5) Very Aggressive - May help underpowered CPUs on less demanding games, but is likely to cause problems in other cases.\n6) Mostly Harmful - Harsh application of speed hacks. May help a very small set of games that have unusual performance requirements, but have adverse effects on most others. Not recommended for underpowered PCs.";
+
```
pgert: Translation of this string is no longer supported, assumably because of this change.
Care to fix it?

PR #2797, opened 2019-01-03
PCSX2: Remove the arbitrary limit on patches by converting the patch list to a vector

GovanifY:
Pnach had a limit that, while increased at some point to 2048, is still not enough for everyone. This uses a vector to avoid that limit, as there is no reason to keep it and people loading pnach with over 2048 patches most likely know what they are doing.
Inipatch group member was also unused in the whole codebase so I did some cleanup and removed it.

### DIFF
--- a/pcsx2/Patch.cpp
+++ b/pcsx2/Patch.cpp
@@ -36,7 +36,6 @@ extern void _ApplyPatch(IniPatch *p);
 
 std::vector<IniPatch> Patch;
 
-
 wxString strgametitle;
 
 struct PatchTextTable
--- a/pcsx2/Patch.cpp
+++ b/pcsx2/Patch.cpp
@@ -34,10 +34,8 @@
 extern void _ApplyPatch(IniPatch *p);
 
 
-
 std::vector<IniPatch> Patch;
 
-int patchnumber = 0;
 
 wxString strgametitle;
 
@@ -154,7 +152,7 @@ int LoadPatchesFromGamesDB(const wxString& crc, const Game_Data& game)
 
 	if (patchFound) TrimPatches(patch);
 
-	return patchnumber;
+	return Patch.size();
 }
 
 void inifile_processString(const wxString& inStr)
@@ -176,7 +174,7 @@ void inifile_process(wxTextFile &f1 )
 
 void ForgetLoadedPatches()
 {
-  patchnumber = 0;
+    Patch.clear();
 }
 
 static int _LoadPatchFiles(const wxDirName& folderName, wxString& fileSpec, const wxString& friendlyName, int& numberFoundPatchFiles)
@@ -189,18 +187,18 @@ static int _LoadPatchFiles(const wxDirName& folderName, wxString& fileSpec, cons
 	}
 	wxDir dir(folderName.ToString());
 
-	int before = patchnumber;
+	int before = Patch.size();
 	wxString buffer;
 	wxTextFile f;
 	bool found = dir.GetFirst(&buffer, L"*", wxDIR_FILES);
 	while (found) {
 		if (buffer.Upper().Matches(fileSpec.Upper())) {
 			PatchesCon->WriteLn(Color_Green, L"Found %s file: '%s'", WX_STR(friendlyName), WX_STR(buffer));
-			int before = patchnumber;
+			int before = Patch.size();
 			f.Open(Path::Combine(dir.GetName(), buffer));
 			inifile_process(f);
 			f.Close();
-			int loaded = patchnumber - before;
+			int loaded = Patch.size() - before;
 			PatchesCon->WriteLn((loaded ? Color_Green : Color_Gray), L"Loaded %d %s from '%s' at '%s'",
 			                    loaded, WX_STR(friendlyName), WX_STR(buffer), WX_STR(folderName.ToString()));
 			numberFoundPatchFiles++;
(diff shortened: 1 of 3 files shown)


willkuer:
I wonder how one generates more than 2000 patches for a single game.

pcsx2/Patch.cpp
```diff
@@ -33,7 +34,8 @@
 extern void _ApplyPatch(IniPatch *p);
 
 
-IniPatch Patch[ MAX_PATCH ];
+
+std::vector<IniPatch> Patch;
```
tich: You should watch out for all uses of the 'patchnumber' variable. I admit that I don't fully understand its uses (for example I have no idea why 'patchnumber' is used at all in _LoadPatchFiles() since that function doesn't call any other function that modifies this variable). Perhaps all uses of 'patchnumber' can be cleaned up and the variable may either become unnecessary, or can be replaced with 'Patch.size()' where needed.

But one place where you definitely missed something is ForgetLoadedPatches(). You should be clearing the vector there

gregory38:
It is good idea. Besides it will reduces memory requirements whn few patches are used.

pcsx2/Patch.cpp
GovanifY: iniprocess->initable->PatchTableExecute which executes a different function in a table called commands_patch depending if this is a comment or a patch. In our case it is a patch so PatchFunc::patch is called, which does this:
`	void patch(const wxString& cmd, const wxString& param) { patchHelper(cmd, param); }
`
which calls our patchHelper, trying to load patches :D
Altho darn, indeed missed it! Going to fix it now!

tich: Ah! That clears it up for me, thanks!

pcsx2/Patch.cpp
```diff
@@ -349,9 +349,8 @@ namespace PatchFunc
 // This is for applying patches directly to memory
 void ApplyLoadedPatches(patch_place_type place)
 {
-	for (unsigned int i = 0; i < Patch.size(); i++)
-	{
-	    if (Patch[i].placetopatch == place)
-            _ApplyPatch(&Patch[i]);
+    for (auto i : Patch){
```
tich: this will create a copy of each element. You can use "auto& i" to specify that you want a reference to the existing element

lightningterror:
Formatting is a mess, old code uses tabs for new rows mostly.

GovanifY:
Old formatting was already a mess for that matter, ie line 118, either way just fixed it

GovanifY:
It's been a week since @turtleli has been asked for a review and seeing his GitHub activity profile it looks like he might still be away for a while, do we really need another review on that PR? At it's core it's really just changing the type of a variable and has already been tested+reviewed once

pcsx2/Patch.cpp
```diff
@@ -227,14 +227,14 @@ int LoadPatchesFromZip(wxString gameCRC, const wxString& patchesArchiveFilename)
 		name.MakeUpper();
 		if (name.Find(gameCRC) == 0 && name.Find(L".PNACH")+6u == name.Length()) {
 			PatchesCon->WriteLn(Color_Green, L"Loading patch '%s' from archive '%s'",
-			                    WX_STR(entry->GetName()), WX_STR(patchesArchiveFilename));
+								WX_STR(entry->GetName()), WX_STR(patchesArchiveFilename));
```
lightningterror: Maybe 4 tabs less ?

pcsx2/Patch.cpp
```diff
@@ -187,26 +187,26 @@ static int _LoadPatchFiles(const wxDirName& folderName, wxString& fileSpec, cons
...
-			int before = patchnumber;
+			int before = Patch.size();
 			f.Open(Path::Combine(dir.GetName(), buffer));
 			inifile_process(f);
 			f.Close();
-			int loaded = patchnumber - before;
+			int loaded = Patch.size() - before;
 			PatchesCon->WriteLn((loaded ? Color_Green : Color_Gray), L"Loaded %d %s from '%s' at '%s'",
-			                    loaded, WX_STR(friendlyName), WX_STR(buffer), WX_STR(folderName.ToString()));
+								loaded, WX_STR(friendlyName), WX_STR(buffer), WX_STR(folderName.ToString()));
```
lightningterror: Here as well.

GovanifY:
So both @avih and @turtleli seems to be away and so far there has been no real blockers in this PR, any reason why we _must_ have another reviewer before merging this, like, at all?

avih:
I'm not away, I just don't have anything to add beyond what others said. I don't object this feature even if personally I don't see a need for it.

pcsx2/Patch.h
```diff
@@ -38,7 +38,6 @@
 #include "Pcsx2Defs.h"
 #include "SysForwardDefs.h"
 
-#define MAX_PATCH 2048
```
lightningterror: Maybe remove an extra row as well ?

GovanifY:
@lightningterror I understand but we already had 3 reviewers: greg, you and tich which seemed mostly happy, and I really have trouble seeing why we need up to 5 reviewers for such a small change tbh

pcsx2/Patch.cpp
```diff
@@ -33,9 +34,8 @@
 extern void _ApplyPatch(IniPatch *p);
 
 
-IniPatch Patch[ MAX_PATCH ];
+std::vector<IniPatch> Patch;
 
-int patchnumber = 0;
```
lightningterror: Forgot to mention this one, another row that could be removed.

PR #2805, opened 2019-01-08
pcsx2/spu2-x: Cleanup some unused variables/functions

lightningterror:
pcsx2: Remove unused variables.
MC2_SIZE in MemoryCardFile.cpp,
length in microVU_Log.inl
VU_Neg_Infinity in sVU_Upper.cpp.

Commented out:
mc_sizeinfo_8mb in Sio.cpp

spu2-x: Adjust some Dolby Pro Logic II code.
Comment out unused variable sLogTable in DplIIdecoder.cpp.
Remove DPLII.h from cmake list, it's unused so don't list it.
Some people suggested to keep the source files even if they are unused
so we can do that. Maybe someone will get the code to work properly.


common/include/Utilities/gtkGuiTools.h
```diff
@@ -89,45 +89,3 @@ static GtkWidget *ps_gtk_vseparator_new()
     return gtk_separator_new(GTK_ORIENTATION_VERTICAL);
 #endif
 }
-
-// These two routines have been rewritten over and over. May as well include a copy.
-// Renaming so as not to interfere with existing versions.
```
arcum42: This file was actually written by me within the last couple months. Anything not used in it is because I haven't used it yet.  In particular, as the comment says, these functions have been written over and over in the plugins. They just haven't been *replaced* in those locations yet.

pcsx2/Sio.cpp
```diff
@@ -31,10 +31,6 @@ static void sioWrite8inl(u8 data);
 // Magic psx values from nocash info
 static const u8 memcard_psx[] = {0x5A, 0x5D, 0x5C, 0x5D, 0x04, 0x00, 0x00, 0x80};
 
-// Memory Card Specs for standard Sony 8mb carts:
-//    Flags (magic sio '+' thingie!), Sector size, eraseBlockSize (in pages), card size (in pages), xor checksum (superblock?), terminator (unused?).
-static const mc_command_0x26_tag mc_sizeinfo_8mb= {'+', 512, 16, 0x4000, 0x52, 0x5A};
```
arcum42: One thing about removing this one is that it's documenting specs I could potentially see someone working on the code using...

pcsx2/x86/sVU_Upper.cpp
```diff
@@ -129,7 +129,6 @@ static const __aligned16 u32 VU_Zero_Mask[4]				= {0x00000000, 0x00000000, 0x000
 static const __aligned16 u32 VU_Zero_Helper_Mask[4]			= {0x7fffffff, 0x7fffffff, 0x7fffffff, 0x7fffffff};
 static const __aligned16 u32 VU_Signed_Zero_Mask[4]			= {0x80000000, 0x80000000, 0x80000000, 0x80000000};
 static const __aligned16 u32 VU_Pos_Infinity[4]				= {0x7f800000, 0x7f800000, 0x7f800000, 0x7f800000};
-static const __aligned16 u32 VU_Neg_Infinity[4]				= {0xff800000, 0xff800000, 0xff800000, 0xff800000};
```
arcum42: Shouldn't we be deleting the whole file, along with the rest of superVU?

lightningterror: Someday.

pcsx2/Sio.cpp
lightningterror: What about just commenting out the code then ?

common/include/Utilities/gtkGuiTools.h
lightningterror: I'll exclude the changes from this file then.

plugins/spu2-x/src/DecodeDPLII.cpp
```diff
@@ -1,269 +0,0 @@
...
- * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License
- * for more details.
- *
- * You should have received a copy of the GNU Lesser General Public License along
- * with this library; if not, write to the Free Software
- * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
- */
-
-#include "Spu2.h"
-#include "DPLII.h"
```
arcum42: Not totally sure what DPLII is, but this is the only file that includes DPLII.h, so if it's going, the header probably should too.

Also, any files removed should be removed in CMakeLists.txt as well.

lightningterror: Dolby Pro Logic II, anyway I only chose to remove this since it wasn't actually included neither in VS project files nor in cmake, didn't check the header but that can be removed as well since it's only called in this file.

pcsx2/Sio.cpp
arcum42: Commenting it out is probably fine. I've left commented code in plenty of times when it looks useful for reference or logging/debugging.

PR #2807, opened 2019-01-11
GSdx-d3d11: Remove old colclip algo and add support for HDR colclip

lightningterror:
Remove colclip and negative blend algo.
Code was old and inefficient. HDR colclip will do a better job.

GSdx-d3d11: Add/port HDR colclip support.
It helps render shadows a lot better compared to the old code.

convert.fx: Update main4 shader to better support hdr colclip.

Credits to @gregory38 and  @tadanokojin for helping with the code.


gregory38:
A test of sotc would be nice. Probably only game where old algo was fine

gregory38:
+ the minor comment it could be merged

PR #2814, opened 2019-01-13
GameDB: UEFA Euro 2004 black box fix

atomic83GitHub:
The PR fixes black box issues when going ingame with UEFA Euro 2004.

I only apply it on the fr-g version as it's the only I can test.

Without the fix:
image

With it:
image


MrCK1:
Looks like you're using an old branch. There's going to be conflicts

atomic83GitHub:
Conflicts fixed, ready to merge :).

PR #2815, opened 2019-01-14
GSdx: Purge MSAA entirely

tadanokojin:
This PR completely removes MSAA support for GSdx.

**Backstory**
DX continues to move toward feature parity with OGL.

As such, MSAA was broken in DX following recent accuracy improvements and was not implemented on OGL. This prompted us to first disable the option from the UI and prevent it from being enabled (f6df6b350b3c72e4fd32ea3fb3fb76cb8b325697) followed shortly by removing the disabled option altogether (3a7f2a5c0b12362feb958dfbb207f4a5a15d2dfe).

**Why Remove MSAA?**
1. In it's current state, MSAA is broken.
2. Fixing MSAA to reflect the DX renderer's new direction would mean it would perform about as well as FSAA, which would defeat the entire purpose of having it.
3. It's only supported by the DX renderer and by extension, Windows.
4. Reduction in code complexity

Users who wish to continue using super sampling as a form of AA should use FSAA by increasing the resolution multiplier.

As with the DX9 removal, we kindly ask that you press F to pay respects.


plugins/GSdx/GSdx.rc
```diff
@@ -138,8 +138,6 @@ BEGIN
     CONTROL         "Wild Arms Offset",IDC_WILDHACK,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,108,84,70,8
     // Columns end
     //
-    RTEXT           "MSAA:",IDC_MSAA_TEXT,54,99,22,8
-    COMBOBOX        IDC_MSAACB,80,96,116,63,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
```
lightningterror: The options might need to be moved one slot as well as the height adjusted to compensate for the removed option, it can be done at a later commit/pr tho.

gregory38:
Cool. Add the why in the commit msg. And good for me

tadanokojin:
@gregory38 done

Maxkatsur:
I miss MSAA... Dolphin and DuckStation STILL have MSAA, along with DX12 renderer, ironically speaking. [One video by Threat Interactive](https://youtu.be/SxCMaTEoBoI) made me think about MSAA and its implementations. We can fix MSAA, and I find this decision to purge it entirely very cynical and pointless. And, any ideas about adding CMAA2 as implemented in _Counter-Strike 2_? There is still a room for improving MSAA implementation anyway.

PR #2818, opened 2019-01-18
GSdx-ogl: Small colclip improvements

lightningterror:
1. Allow hrd colclip to also run on the unsupported
cases which should be pretty rare.

2. Remove colclip from high sw blending.

3. Disable the hdr colclip shader which may not be
actually required to run.

Note: I didn't run any tests to compare raw dump, only performance comparison which was within margin.

Idea by @gregory38


lightningterror:
Forgot to comment out PS_HDR for the commented out shader..

PR #2823, opened 2019-01-19
gsdx-d3d11: ICO depth HLE

tadanokojin:
Port of the ICO HLE fix and update to fb copy on slot 4


tadanokojin:
@gregory38 I updated with the following
- cache rt depth stencil ptr on `m_state` and update detection
- ternary for `CopyRect` box (and comment about DX api)
- throw error if copy isn't a RenderTarget/Depth
- remove curly braces
- rename `CopyRenderTarget` to `CopyTexture`

I also cleaned up a cast and the if statement for `DrawIndexPrimitive`

gregory38:
Nice now you can split it in smaller commit ;)

tadanokojin:
Split it up.

I also boldly added the assert ;)
Let me know if the changes are good.

tadanokojin:
Also let me know if the pass by reference is okay.

tadanokojin:
Updated

PR #2824, opened 2019-01-19
gsdx-debug: Add secret ini option for dump directory

tadanokojin:
Add a secret ini option to change the raw dump directory.

Also the following changes:
- Windows creates the directory now
- Directory is now `GSDX_[32|64]_[SW|HW]`

#SaveTheSSDs

### DIFF
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -44,6 +44,7 @@ GSState::GSState()
 	, m_crc(0)
 	, m_options(0)
 	, m_frameskip(0)
+	, m_dump_root("")
 {
 	// m_nativeres seems to be a hack. Unfortunately it impacts draw call number which make debug painful in the replayer.
 	// Let's keep it disabled to ease debug.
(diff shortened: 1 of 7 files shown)


plugins/GSdx/GSState.cpp
```diff
@@ -44,6 +44,7 @@ GSState::GSState()
 	, m_crc(0)
 	, m_options(0)
 	, m_frameskip(0)
+	, m_dump_root("")
```
tadanokojin: my mistake, I don't need to initialize this to an empty string.
will fix tomorrow.

gregory38:
But why not this way too

gregory38:
So. It would be much nicer to have a virtual function that get the root dir.

plugins/GSdx/GSdx.cpp
```diff
@@ -366,6 +366,11 @@ void GSdxApp::Init()
 	m_default_configuration["debug_opengl"]                               = "0";
 	m_default_configuration["disable_hw_gl_draw"]                         = "0";
 	m_default_configuration["dump"]                                       = "0";
+#ifdef _WIN32
+	m_default_configuration["dump_dir"]                                   = "C:\\tmp\\";
```
lightningterror: Maybe dumping in install dir is better ? So we go less back and forth.

PR #2829, opened 2019-01-26
NSIS: Fix regkey removal regression, add comments, misc. tweaks

MrCK1:
Not much is changed since the rewrite I did 2 years ago. This maintenance release does just a few things:

- Fixes a regression with a regkey that was not being removed by the installer.
- Adds a bunch of comments so somebody other then myself can understand what's going on under the hood ;)
- Remove some leftover code and add a UAC shield to the dialog when selecting the standard installer.

Just incase anybody's forgotten what the installer looks like, here's a screenshot and a [test build](https://github.com/PCSX2/pcsx2/files/2798597/pcsx2_1.6.0_RC.zip). Tested it myself and all the functions (including the ini backup) seem to work without a hitch. Let me know if you manage to encounter any bugs ;)

image


ramapcsx2:
Second this. "Regular" or "Normal" sound pretty neutral to me :)
Thanks by the way, those NSIS scripts are ancient aliens stuff imo :p

lightningterror:
Normal does sound better yes.

lightningterror:
Do we need that logo tho ? It looks awful now without it being stretched properly.

PR #2830, opened 2019-01-26
gsdx-d3d11: Framebuffer copy improvements

tadanokojin:
Improved handling of frame buffer copy by accounting for fb read on all slots.

- implement before/after draw functions
- defer setting of resources on the api to just before the draw
- use a bitfield to keep track of set/copied resources
- recycle copies after draw

Fixes: RE4 and Tales of Abyss


tadanokojin:
Updated and rebased.

tadanokojin:
Updated

PR #2842, opened 2019-02-06
GSdx-Legacy: Purge Gsdx Legacy and all traces from master branch

lightningterror:
1.4 branch already does the same so there's no need to have a copy of gsdx.


mirh:
Agreed with the "rationale". 
But.. are we not any more compatible with GL3.3? It's dx10 after all.

orbea:
I think its useful to keep for testing against the newer gsdx in case of regressions. Also it might break compatibility for some users with legacy hardware.

ssakash:
```
I think its useful to keep for testing against the newer gsdx in case of regressions.
```

I agree with this point, would be easier to compile the legacy one to check if an issue was present before or not.

lightningterror:
If people want legacy to remain then we won't purge it, tho once we release 1.6 people can still use the legacy plugin from 1.6 release on linux. Also compile it from the branch as well.
Personally I don't see the point keeping it for 1.7/1.8.

lightningterror:
I'll close this for now, we can leave it up for debate in the future.

PR #2837, opened 2019-02-01
Greg/gsdx gl

gregory38:
Misc PR that mostly clean the loading of GL function. It should allow to code new stuff without bothering with the GL boilerplate.

I also remove 1 layer for texture creation.

Please test GL on windows (intel too).

Issue #2813, opened 2019-01-12
Armored Core Last Raven goes to half speed in menu after missions and AC Test


disastranagant:
***PCSX2 version:***
Several 1.5 builds tested, going back to at least 640, with https://buildbot.orphis.net/pcsx2/index.php?m=dl&rev=v1.5.0-dev-2867-gc58acd110&platform=windows-x86 being the latest tested

***PCSX2 options:***
All settings default except for the required Preload Frame Data hack selected in GSDX

***Description of the issue:***
The menus drop to 30 fps when returning from any battle screen.  Sometimes it reverts to full speed after a while or after closing the GS window and resuming.

***How to reproduce the issue:***
Start a new game.  Go to garage > VR AC Test and choose any option.   After the battle loads hit start and end the mission.  Enjoy your new 30-32 FPS.  Close GS window and resume to maybe go back to 60+.

***Last known version to work:***
Not sure.  I didn't have this issue when I played a couple years ago but I have no idea version that was on.

***PC specifications:***
CPU: Intel I5-4570, 8 GB DDR3 RAM, GPU: Geforce GTX760, Windows 8.1 64bit

GS Dump: https://mega.nz/#!tBkFUQpa!aaWwwfnUiWi3Mv9BhL4UKYdkaKaC3CanLJ2lm87UNB4

MrCK1:
Was this testing done at native resolution ect? Can you post the entire GSdx settings dialog?

disastranagant:
Yes, the most recent version used was a fresh install with just that one hardware hack enabled that is required to not black screen the game.

![image](https://user-images.githubusercontent.com/3312331/51450301-d48bae80-1cf5-11e9-8d78-afa7598adea2.png)
![image](https://user-images.githubusercontent.com/3312331/51450310-e4a38e00-1cf5-11e9-966c-a2d405e1fa3e.png)

lightningterror:
Large Framebuffer should be disabled when upscaling for this game else the vram spikes will be insanely high.

Helium-4:
I did test with it off.

MrCK1:
@Helium-4 @disastranagant Can you test with a GSdx plugin from #2843 and report whether or not VRAM spikes/performance issues still occur? Thanks.

disastranagant:
I've never built gsdx before but I'll give it a go.

lightningterror:
No need to build it, you can download one of the appveyor artifacts
[...]

PR #2843, opened 2019-02-07
Greg/gsdx sparse

gregory38:
Please find some code to create/manage sparse render target&depth buffer.

GL backend is mostly done. Likely need to fix function that rely on texture size.

Dx will gracefully fallback to standard mode.

Nothing is done on renderer neither the texture cache to really use those sparse texture.

### DIFF
--- a/plugins/GSdx/Renderers/OpenGL/GSDeviceOGL.h
+++ b/plugins/GSdx/Renderers/OpenGL/GSDeviceOGL.h
@@ -563,6 +563,8 @@ public:
 	void OMSetRenderTargets(GSTexture* rt, GSTexture* ds, const GSVector4i* scissor = NULL) final;
 	void OMSetColorMaskState(OMColorMaskSelector sel = OMColorMaskSelector());
 
+	virtual bool HasColorSparse() { return GLLoader::found_compatible_GL_ARB_sparse_texture2; }
+	virtual bool HasDepthSparse() { return GLLoader::found_compatible_sparse_depth; }
 
 	void CreateTextureFX();
 	GLuint CompileVS(VSSelector sel);
--- a/plugins/GSdx/Renderers/Common/GSDevice.cpp
+++ b/plugins/GSdx/Renderers/Common/GSDevice.cpp
@@ -163,6 +163,7 @@ void GSDevice::Recycle(GSTexture* t)
 {
 	if(t)
 	{
+		t->Uncommit();
 		t->last_frame_used = m_frame;
 
 		m_pool.push_front(t);
--- a/plugins/GSdx/Renderers/Common/GSDevice.cpp
+++ b/plugins/GSdx/Renderers/Common/GSDevice.cpp
@@ -163,7 +163,12 @@ void GSDevice::Recycle(GSTexture* t)
 {
 	if(t)
 	{
+#ifdef _DEBUG
+		// Uncommit saves memory but it means a futur allocation when we want to reuse the texture.
+		// Which is slow and defeat the purpose of the m_pool cache.
+		// However, it can help to spot part of texture that we forgot to commit
 		t->Uncommit();
+#endif
 		t->last_frame_used = m_frame;
 
 		m_pool.push_front(t);
--- a/plugins/GSdx/Renderers/Common/GSDevice.cpp
+++ b/plugins/GSdx/Renderers/Common/GSDevice.cpp
@@ -201,6 +201,16 @@ void GSDevice::PurgePool()
 	}
 }
 
+GSTexture* GSDevice::CreateSparseRenderTarget(int w, int h, int format)
+{
+	return FetchSurface(HasColorSparse() ? GSTexture::SparseRenderTarget : GSTexture::RenderTarget, w, h, format);
+}
+
+GSTexture* GSDevice::CreateSparseDepthStencil(int w, int h, int format)
+{
+	return FetchSurface(HasDepthSparse() ? GSTexture::SparseDepthStencil : GSTexture::DepthStencil, w, h, format);
+}
+
 GSTexture* GSDevice::CreateRenderTarget(int w, int h, int format)
 {
 	return FetchSurface(GSTexture::RenderTarget, w, h, format);
--- a/plugins/GSdx/Renderers/Common/GSDevice.h
+++ b/plugins/GSdx/Renderers/Common/GSDevice.h
@@ -167,11 +167,16 @@ public:
 	virtual void DrawIndexedPrimitive(int offset, int count) {}
 	virtual void EndScene();
 
+	virtual bool HasDepthSparse() { return false; }
+	virtual bool HasColorSparse() { return false; }
+
 	virtual void ClearRenderTarget(GSTexture* t, const GSVector4& c) {}
 	virtual void ClearRenderTarget(GSTexture* t, uint32 c) {}
 	virtual void ClearDepth(GSTexture* t) {}
 	virtual void ClearStencil(GSTexture* t, uint8 c) {}
 
+	GSTexture* CreateSparseRenderTarget(int w, int h, int format = 0);
+	GSTexture* CreateSparseDepthStencil(int w, int h, int format = 0);
 	GSTexture* CreateRenderTarget(int w, int h, int format = 0);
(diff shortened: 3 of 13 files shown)


plugins/GSdx/Renderers/OpenGL/GSDeviceOGL.h
```diff
@@ -562,6 +562,8 @@ class GSDeviceOGL final : public GSDevice
 	void OMSetRenderTargets(GSTexture* rt, GSTexture* ds, const GSVector4i* scissor = NULL) final;
 	void OMSetColorMaskState(OMColorMaskSelector sel = OMColorMaskSelector());
 
+	virtual bool HasDepthSparse() { return GLLoader::found_compatible_GL_ARB_sparse_texture2; }
+	virtual bool HasColorSparse() { return GLLoader::found_compatible_sparse_depth; }
```
tadanokojin: It looks backward

gregory38:
@tadanokojin @lightningterror I added some small improvements for sparse. In order to test it, you must force enable the sparse extension. Do you have any pending stuff on your side for sparse ?

lightningterror:
I got none, what about https://github.com/PCSX2/pcsx2/pull/2837#issuecomment-467450952 ?

gregory38:
I posted in the wrong PR. It was some tests for AMD if it doesn't help, no need to keep it.

gregory38:
So far code is disabled for everyone. But I'm afraid that it will be limited to Nvidia :(

tadanokojin:
I don't have anything that I can think of, I'll do some testing later today.

tadanokojin:
Major issues with Burnout 3

No Sparse:


Sparse:

gregory38:
Issue is perf ? Or the rendering ?

tadanokojin:
Perf, going from pin 60 to lower ends of 10.
Black sky is unrelated.

gregory38:
Hum need to find the culprit.
* No Sparse depth ?
* Always commit with no uncommit ?

tadanokojin:
I tested SoTC, DC2, GT4 and Jak 3 didn't notice any issues there.

Burnout Revenge has the same issue and I think it's safe to say Burnout Dominator probably has a similar issue (don't own it).

I tested with sparse depth enabled. No luck there.

note: not noticing any high/abnormal gpu usage or anything.

tadanokojin:
Burnout 3 is fine @ 2074597ab07e25e76576a4290d46e52196ff2670 as I remembered

Same commit revenge is spamming the log about sparse not being a multiple of the page size and still has severe perf issues.

gregory38:
@tadanokojin 
I won't have time to finish it. If you have time, could you like at the following idea/code. So far it is only a hack.
Currently, TC might rescale a previous target to use it as an input texture. Rescaling sucks, it is likely slower and it reduces quality. I want to use the committed info, to only copy the useful data to the real sized GS texture.

For example, GS will render into a frame buffer emulated by a (2x upscaled) 1536x768 target on GPU but only 1280x512 is committed. Then GS will read it back as a 1024x512 texture. We want to create a sparse target of 2048x1024 (scaling factor * size of GS). Then copy only the 1280x512 data from old target to new source.

Game that need to be tested 
* tri-ace game (radiata story/star ocean 3/forgot the 3rd one) with blending off, they use it to compute shadow. They will hit this code path 100th a time by frame !
* snowblend engine game (baldur gate/...), because of the half-screen (`half_right`) special case

Here the current hacking patch
```
--- a/plugins/GSdx/Renderers/HW/GSTextureCache.cpp
+++ b/plugins/GSdx/Renderers/HW/GSTextureCache.cpp
@@ -1190,6 +1190,7 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
            GL_INS("ERROR: Reading RT as a packed-indexed 4 bits format is not supported");
        }
 #endif
+       bool is_copy = shader == ShaderConvert_COPY;
 
        if (GSLocalMemory::m_psm[TEX0.PSM].bpp > 8) {
            src->m_32_bits_fmt = dst->m_32_bits_fmt;
@@ -1300,7 +1301,11 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
        // 2/ It doesn't support rescaling of the RT (tw = 1024)
        // Maybe it will be more easy to just round the UV value in the Vertex Shader
 
-       if (!is_8bits) {
+       if (is_copy && m_renderer->m_dev->HasColorSparse()) {
+           GSVector2i committed_size = dst->m_texture->GetCommittedSize();
+           dRect.z = std::min(committed_size.x, w);
+           dRect.w = std::min(committed_size.y, h);
+       } else if (!is_8bits) {
            // 8 bits handling is special due to unscaling. It is better to not execute this code
            if (w > dstsize.x)
            {
@@ -1318,6 +1323,9 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
        }
 
        GSVector4 sRect(0, 0, w, h);
+       if (is_copy && m_renderer->m_dev->HasColorSparse()) {
+           sRect = dRect;
+       }
 
        // Don't be fooled by the name. 'dst' is the old target (hence the input)
        // 'src' is the new texture cache entry (hence the output)
@@ -1357,7 +1365,7 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
        // copy. Likely a speed boost and memory usage reduction.
        bool linear = (TEX0.PSM == PSM_PSMCT32 || TEX0.PSM == PSM_PSMCT24);
 
-       if ((sRect == dRect).alltrue() && !shader)
+       if ((sRect == dRect).alltrue() && is_copy)
        {
            if (half_right) {
                // You typically hit this code in snow engine game. Dstsize is the size of of Dx/GL RT
@@ -1366,7 +1374,7 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
                float real_width = dst->m_TEX0.TBW * 64u * dst->m_texture->GetScale().x;
                m_renderer->m_dev->CopyRect(sTex, dTex, GSVector4i((int)(real_width/2.0f), 0, (int)real_width, h));
            } else {
-               m_renderer->m_dev->CopyRect(sTex, dTex, GSVector4i(0, 0, w, h)); // <= likely wrong dstsize.x could be bigger than w
+               m_renderer->m_dev->CopyRect(sTex, dTex, GSVector4i(0, 0, sRect.z, sRect.w)); // <= likely wrong dstsize.x could be bigger than w
            }
        }
        else
```

lightningterror:
Just so we don't forget https://github.com/PCSX2/pcsx2/issues/2813 needs to be looked over. Seems sparse makes things worse for now.

gregory38:
Strange. Long dump of spike would be nice.

gregory38:
Thing to do
* add commit in the 2 copy function (device)
* only uncommit in recycle in debug/dev build. Too much uncommit is bad for performance.

lightningterror:
I pushed the sparse extension detection, amd sparse depth and tc dead code commits on to master with some minor tweaks followed by a rebase on current branch.

Freestate2nd:
> 
> 
> @tadanokojin

> Game that need to be tested
> 
>     * tri-ace game (radiata story/star ocean 3/forgot the 3rd one) with blending off, they use it to compute shadow. They will hit this code path 100th a time by frame ! Don't forget to disable CRC hack that will skip those shadows

> ```

The third TRI-ACE GAME for ps2 is Valkyrie Profile Silmeria (AKA Valkyrie Profile 2)

Freestate2nd:
Third Tri-ace game for testing is Valkyrie Profile Silmeria

lightningterror:
@gregory38 with the new test code for Star Ocean 3 I'm noticing around 7 fps increase in the game compared to current sparse branch.

gregory38:
@lightningterror what test code ?

lightningterror:
Ye that one that you asked kojin to test.
https://github.com/PCSX2/pcsx2/pull/2843#issuecomment-470906904

gregory38:
Oh. What upscaled resolution. Accurate blending ?

Then test the new PR.

lightningterror:
Blending off and native res.

lightningterror:
I will merge this now but we shouldn't forget to address the mentioned issues in the future.

PR #2844, opened 2019-02-07
Gsdx-gui: Add option for Sparse Texture in Advanced options

lightningterror:


lightningterror:
Tooltlip is missing btw.

gregory38:
Add tooltip ;)

PR #2856, opened 2019-02-19
fxaa.fx: Purge d3d9 shader code

lightningterror:
Looks like I forgot this one.


gregory38:
Did you check external shader too ?

lightningterror:
Yeah I already purged it in the first pr.
https://github.com/PCSX2/pcsx2/pull/2762/files#diff-af3f7d4af20ebd169a87459cd23d36a1L24

gregory38:
Check gl too as it is the same compiled file.

lightningterror:
> Check gl too as it is the same compiled file.

I know, I did the changes with that in mind.

gregory38:
Perfect

PR #2861, opened 2019-02-21
gsdx hw:move vertices fixup for texture shuffle in RendererHw

gregory38:
Bonus: fix m_vt value

Miss Dx and some tests ;)


lightningterror:
Good job with the bbox fix :)
I did the dx code with some minor cleanups as well, test show both dx and gl working as expected on multiple games.

PR #2869, opened 2019-03-01
cdvd: Get rid of code that disabled loading psx discs

### LINKED ISSUE #2868: CDVD PSX Check Obsolete
CDVD have checks for PSX disc inserted. (that haves translation for it.)
but that is obsolete since we have PSX support in PCSX2.

Might be wise to purge that.

lightningterror:
Not used anymore.

Close #2868


ssakash:
Yeah, that macro is pointless and doesn't need to stay anymore.

PR #2882, opened 2019-03-09
Add percentage to satisfy Gettext

rffontenelle:
With only one percentage sign in the c-format string, Gettext will treat "% c" as a placeholder, which is not the case of these strings. The translation of "% c" would be considered an error by Gettext, unless another percentage sign is added.

An example of error displayed by Gettext due to the missing second percentage sign is:
```
$ LC_ALL=C sudo msgfmt -co /usr/share/locale/pt_BR/LC_MESSAGES/pcsx2_Iconized.mo locales/pt_BR/pcsx2_Iconized.po 
locales/pt_BR/pcsx2_Iconized.po:601: format specifications in 'msgid' and 'msgstr' for argument 1 are not the same
locales/pt_BR/pcsx2_Iconized.po:611: format specifications in 'msgid' and 'msgstr' for argument 1 are not the same
locales/pt_BR/pcsx2_Iconized.po:620: format specifications in 'msgid' and 'msgstr' for argument 1 are not the same
locales/pt_BR/pcsx2_Iconized.po:636: format specifications in 'msgid' and 'msgstr' for argument 1 are not the same
locales/pt_BR/pcsx2_Iconized.po:646: format specifications in 'msgid' and 'msgstr' for argument 1 are not the same
locales/pt_BR/pcsx2_Iconized.po:657: format specifications in 'msgid' and 'msgstr' for argument 1 are not the same
msgfmt: found 6 fatal errors
```

### DIFF
--- a/pcsx2/gui/Panels/SpeedhacksPanel.cpp
+++ b/pcsx2/gui/Panels/SpeedhacksPanel.cpp
@@ -26,19 +26,19 @@ const wxChar* Panels::SpeedHacksPanel::GetEECycleRateSliderMsg( int val )
 		case -3:
 		{
 			m_msg_eeRate->SetForegroundColour(wxColour(L"Red"));
-			return pxEt(L"50% cyclerate. Significant reduction of CPU requirements. Speedup for very lightweight games, slows down others. FMVs and audio may stutter or skip.");
+			return pxEt(L"50%% cyclerate. Significant reduction of CPU requirements. Speedup for very lightweight games, slows down others. FMVs and audio may stutter or skip.");
 		}
 		case -2:
 		{
 			const wxColour LightRed = wxColour(255, 80, 0);
 			m_msg_eeRate->SetForegroundColour(LightRed);
-			return pxEt(L"60% cyclerate. Moderate reduction of CPU requirements. Speedup for lightweight games, slows down others. FMVs and audio may stutter or skip.");
+			return pxEt(L"60%% cyclerate. Moderate reduction of CPU requirements. Speedup for lightweight games, slows down others. FMVs and audio may stutter or skip.");
 		}
 		case -1:
 		{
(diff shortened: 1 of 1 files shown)


lightningterror:
Double percentage tho looks weird in the gui now, maybe @avih has a good idea what to do here.

avih:
The PR looks good to me, but I'll leave the merge to whomever changed the texts to their current state in the first place.

lightningterror:
I would actually recommend replacing the percentage value with MHz clock rate instead. Based of the under/overclock percentage, I would also include the default as well.
Percentage values can be left in the switch case with code comments as well.

pgert:
"x% cyclerate" causes problems with POedit too.
4x when translated into norwegian, it becomes "x% syklusfrekvens".
POedit will then state that this is in error,
 insisting on that it should be translated into "% c" insted of "% s".

rffontenelle:
> "x% cyclerate" causes problems with POedit too.

@pgert Poedit relies on the GNU Gettext library to handle translation of .po files, so it is the exactly same scenario. This also applies to other specialized translation editors, like Virtaal, Gtranslators, etc.

rffontenelle:
> I would actually recommend replacing the percentage value with MHz clock rate instead. Based of the under/overclock percentage, I would also include the default as well.
> Percentage values can be left in the switch case with code comments as well.

@lightningterror: I don't know the MHz clock rate, but if someone tell me or point me to the information I could propose this change. Alternatively, this PR is set to allow commits by the maintainers of PCSX2, so anyone is feel free to edit it (I wouldn't mind).

lightningterror:
I'll close this in favor of https://github.com/PCSX2/pcsx2/issues/2886 to properly fix the issue.

Valid suggestions were made by @tadanokojin and @knight-ryu12 for a better and more proper fix.

PR #2885, opened 2019-03-11
Plugins: Remove PADSSSPSX and xpad.

MrCK1:
Given their inactive state in the project, I am proposing to fully remove the two pad plugins; PADSSSPSX and xpad. 

In my opinion, both plugins have reduced functionality and are currently superseded by Lilypad and Onepad respectively. (Prior to this PR, the two plugins are already disabled in CMake prior to this PR but were still included in the old_plugins solution.)

In addition, I removed zeropad from CMakeLists which was leftover from its' removal in 2011. CDVDiso was also lurking in one of the Linux files so I took care of that as well

I understand that this may be a controversial subject for some; so please let me know if you have comments, suggestions, ect.


lightningterror:
Smash that purge button.

MrCK1:
Requested change has been pushed. This can be merged in a few hours incase anyone else has comments.

lightningterror:
Maybe we could also purge ZeroGS in the future.

MrCK1:
@lightningterror I have plans for those, but I'm taking baby steps ;)

Issue #909, opened 2015-10-20
[Bug Report] ops2m demo 33 - Primal [PAL]


rz5:
**_PCSX2 versions used:**_ pcsx2-v1.3.1-1256-gbbd74e5 (latest dev build)
**_Emulation settings:**_ "Safest" preset, MTVU hack not used.
**_Plugins used:**_ GSdx AVX2, cddvdGigaherz
**_Plugin settings:**_ CRC Hack Level "None"
#### Hardware Specifications

**_CPU:**_ Intel(R) Core(TM) i5-4690K CPU @ 3.50GHz
**_GPU:**_ NVIDIA GeForce GTX 970 (Driver: NVIDIA 358.50)
#### Description

D3D9, D3D11 and OGL **HARDWARE** renderers are not able to draw anything on the screen. Instead it's all black.
Only the software renderer is able to draw ~~something~~ the expected scene.
OpenCL renderer remains untested.
#### Screenshots

Software renderer output
![image:1](http://forums.pcsx2.net/attachment.php?aid=57680)
#### GS dump

http://forums.pcsx2.net/attachment.php?aid=57681

gregory38:
Is it working in the SW renderer of the 1.2 release?

vsub:
As far as I know,that game never worked in any hardware mode and in software mode it works but it's extremely slow and you can get a little more speed from VU Stealing but it's like the game notest this and..."ohh you are using hacks to get more speed ehh,lets slow the speed even more until you disable them and then I'll restore it again to the normal slow speed" :D

gregory38:
Ok. It was to be sure it wasn't related to the SW fixes done recently, which aren't backported yet into the HW renderer.

ssakash:
> Is it working in the SW renderer of the 1.2 release?

just tested it out and software mode does work fine on the 1.2.1 build/

vsub:
> just tested it out and software mode does work fine on the 1.2.1 build.

By "work" you mean it works but slow or?
Depending on the place and what you are doing,you can get from 60 to I think 5(not sure but for example when you using the vu stealing and try to switch forms,the speed goes down by a lot)

ssakash:
Speed is awful as usual :p , the black screen issue is resolved though.

bositman:
The description is misleading:
Description

D3D9, D3D11 and OGL renderers are not able to draw anything on the screen. Instead it's all black.
Only the software renderer is able to draw something.

[...]

PR #2888, opened 2019-03-13
Update GSTextureCache.cpp

SpannerSwuys:


refractionpcsx2:
What does this fix? This will probably break the FMV's on the Kingdom Hearts/Grandia 3 games.

lightningterror:
It is for Primal https://github.com/PCSX2/pcsx2/issues/909 but this isn't a proper fix.

Issue #2889, opened 2019-03-16
Wrong conditional to select tex_is_fb


gregory38:
OpenGL HW renderer
Game: Jak/R&C/tri ace shadows

Current condition check if primitives don't overlap. However in case of a single triangle (no overlap), we won't use the tex_is_fb path. Technically input texture is correct but the shader also use a more accurate texture coordinate. Therefore it can cause some glitches in some angles.

Proposed fix: check if primitives are triangles rather than checking overlapping.

lightningterror:
I'll close this now since the pr is merged.

PR #2892, opened 2019-03-18
GSdx-ogl: Check if primitives are triangles instead of overlapping

lightningterror:
Single triangle draw can now hit the tex_is_fb path.
Primitive Overlap will return no overlap  for single primitive (triangle).
Ratchet and Clank, Jak, tri-Ace games.

It should improve shadow rendering.

Idea by Gregory.

Issue #2889


gregory38:
I updated the comment a little

PR #2899, opened 2019-03-24
GameDB: Fix PlayStation Underground 4.4 entries

Aced14:
PlayStation Underground 4.4 is a set of 2 discs:
* SCUS-97120: PS1 CD (disc 1)
* SCUS-97121: PS2 DVD (disc 2)

Specific changes:
* Remove disc 1 since it's out of scope
* Change 4.3 to 4.4 in disc 2's name (4.3 doesn't include any PS2 discs)

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -4506,12 +4506,8 @@ Serial = SCUS-97118
 Name   = NFL GameDay 2001 [Demo]
 Region = NTSC-U
 ---------------------------------------------
-Serial = SCUS-97120
-Name   = PlayStation Underground Demo Disc 4.4 [Disc1&2]
-Region = NTSC-U
----------------------------------------------
 Serial = SCUS-97121
-Name   = PlayStation Underground 4.3 [Disc2]
+Name   = PlayStation Underground 4.4 [Disc2]
 Region = NTSC-U
 ---------------------------------------------
 Serial = SCUS-97122


Aced14:
Btw the PCSX2 wiki already takes this PR's changes into account. Its [PlayStation Underground 4.4](https://wiki.pcsx2.net/PlayStation_Underground_4.4) page clearly states that it's for disc 2, that disc 2's serial number is SCUS-97121 and that disc 1 is for PS1.

pgert:
The GameDB does (to some extent) support PS1 discs.
Wouldn't this look better?

Serial = SCUS-97120
Name   = PlayStation Underground 4.4 [Disc 1 of 2]
Region = NTSC-U
// This is a PS1 CD disc.

Serial = SCUS-97121
Name   = PlayStation Underground 4.4 [Disc 2 of 2]
Region = NTSC-U
// This is a PS2 DVD disc.

Aced14:
@pgert While prepping this commit, I tried searching for some popular PS1 games (like Final Fantasy VII-IX), but didn't find any entries for them. So I had assumed the GameDB was only meant to cover PS2 games.

Are you aware of any PS1 games that have intentionally been included in the GameDB?

I don't mind modifying my commit to retain the PS1 disc, but just want to be sure in advance that it's not breaking an unwritten rule as to which platform's games are supposed to be in the GameDB.

lightningterror:
The gamedb is for ps2 discs only.

mirh:
Depending on what we end up doing in https://github.com/PCSX2/pcsx2/pull/2537#issuecomment-410225989, I guess having a ps1 one shouldn't be that hard..

lightningterror:
I don't mind having a db for ps1 games, as long as they are separated from ps2.

PR #2900, opened 2019-03-24
(GSDumpGUI) Make sure all child processes are closed when closing main form

Souzooka:
This change adds a function to handle the main form being closed, in which all processes are forcibly killed if not already killed.

### DIFF
--- a/tools/GSDumpGUI/Forms/frmMain.Designer.cs
+++ b/tools/GSDumpGUI/Forms/frmMain.Designer.cs
@@ -219,7 +219,7 @@
             this.lblLog.AutoSize = true;
             this.lblLog.Location = new System.Drawing.Point(12, 209);
             this.lblLog.Name = "lblLog";
-            this.lblLog.Size = new System.Drawing.Size(58, 13);
+            this.lblLog.Size = new System.Drawing.Size(54, 13);
             this.lblLog.TabIndex = 14;
             this.lblLog.Text = "Log GSdx";
             // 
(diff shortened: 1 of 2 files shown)


lightningterror:
Kojin will be happy.

Souzooka:
See the new commit @willkuer

willkuer:
This is a good improvement and most probably sufficient. I however meant that you could consider to put it in the Form.Dispose() method. E.g. if I would put an event in front of Form_closing that set event handled to true or I find a way to close the window without executing the form closing event (ie. task kill, application.Exit(), ...).

For me Form_Closing is rather to display a Closing animation or to stop a timer.

Form.Dispose() is for releasing resources as killing and disposing processes or disposing timers and so on.

Also you could consider that a closed window could still be kept in memory (not sure whether WinForms calls Form.Dispose() on Form.Close() but I wouldn‘t expect it - I didn‘t work with WinForms since a few years). And the question then would arise whether the processes should be disposed or not.

So my suggestion: `Form_Closing => Processes.ForEach(p=> p?.Cancel());` (or however that is done on processes) and public ```void Dispose(bool disposing = true)
{
    // default disposing pattern according to MSDN documentation regarding finalizer and such
    /// and whatever WinForms put there like
    // button1.Dispose();

   Processes.Foreach(p=> p?.Dispose());
}
```

But as I said your solution might suffice...

lightningterror:
What is the status, are we waiting on suggested changes @willkuer or it can be merged like this ?

willkuer:
PR is already a thousand times better than master and this is a dev tool. Just merge it. Nobody really cares if you follow dispose pattern in dev tools.

ssakash:
@willkuer  https://github.com/PCSX2/pcsx2/issues/1637

PR #2902, opened 2019-03-25
GameIndex: Add PSI's fix for Panzer Dragoon

knight-ryu12:
JayBee requested the pnach file and confirmed it worked.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -21477,6 +21477,11 @@ Compat = 5
 Serial = SLPM-62718
 Name   = Sega Ages 2500 Series Vol.27 - Panzer Dragoon
 Region = NTSC-J
+[patches = 8D984276]
+	// By PSI
+	comment=Stop screen shaking (game switches to Progressive)
+	patch=1,EE,20158558,extended,30420000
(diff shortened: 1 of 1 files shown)


bin/GameIndex.dbf
```diff
@@ -21477,6 +21477,11 @@ Compat = 5
 Serial = SLPM-62718
 Name   = Sega Ages 2500 Series Vol.27 - Panzer Dragoon
 Region = NTSC-J
+[patches = 8D984276]
+	// By PSI
+	comment=Stop screen shaking (game switches to Progressive)
```
lightningterror: I prefer
```
comment=Patch by PSI
// Stop screen shaking (game switches to Progressive).
```

knight-ryu12: I'd retain what patch actually is; not author

lightningterror: All the other patches (most) are done in the format I suggested.

knight-ryu12: I dislike that idea; Consistency is important, but I think it'd better to show what patch actually does.
Does it show both?

PR #2905, opened 2019-03-26
Readme.md: Update system requirements info

lightningterror:
Replace GHz with STR requirements,
Mention that 1.4 is the last version to support D3D9.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,27 +17,30 @@ Development builds are also available from [our website](https://pcsx2.net/downl
 
 ## Minimum
 * OS: Windows Vista SP2 or newer or GNU/Linux (32-bit or 64-bit)
-* CPU: Any that supports SSE2 (Pentium 4 and up, Athlon64 and up)
+* CPU: Any that supports SSE2 (Pentium 4 and up, Athlon64 and up) @ 1600 STR or better
 * GPU: DirectX 10 GPU or better
 * RAM: 2GB or more
 
 ## Recommended
 * OS: Windows 7/8/8.1/10 (64-bit) or GNU/Linux (64-bit)
-* CPU: Intel Haswell (or AMD equivalent) @ 3.2GHz or better
+* CPU: Intel Haswell (or AMD equivalent) @ 2000 STR or better
 * GPU: DirectX 11 GPU or greater
 * RAM: 4GB or more
 
 ## Notes
 
-- You need the [Visual C++ 2015 x86 Redistributables](https://www.microsoft.com/en-us/download/details.aspx?id=48145) for this version to work.
+- You need the [Visual C++ 2015 x86 Redistributables](https://www.microsoft.com/en-us/download/details.aspx?id=48145) for this version to work.   
+Note: Visual C++ 2017 is directly compatible with Visual C++ 2015. While the project is built with Visual C++ 2015, either version will work.
 
-- PCSX2 1.4.0 is the last version to support Windows XP. Windows XP is no longer getting updates (including security-related updates), and graphics drivers for Windows XP are older and no longer maintained.
+- PCSX2 1.4.0 is the last stable version to support Windows XP and Direct3D9. Windows XP is no longer getting updates (including security-related updates), and graphics drivers for Windows XP are older and no longer maintained.
 
 - Make sure to update your operating system, drivers, and DirectX (if applicable) to ensure you have the best experience possible. Having a newer GPU is also recommended so you have the latest supported drivers.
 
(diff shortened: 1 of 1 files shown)


pgert:
"Visual C++ 2015" -> "Visual C++ 2017"?

https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads

MrCK1:
VS2015 is still supported for building so that doesn't mean much

lightningterror:
> Visual C++ 2015 is used for the build, but both Visual C++ 2015 and 2017 runtimes are compatible. I do think it is worth mentioning the requirement for either of the runtimes, though.

The requirements are based on the buildbot since most people will be using those builds.

lightningterror:
We have separate wiki pages for compiling guides.

RedPanda4552:
My point is, since Orphis currently uses VC++ 2015, it'd be nice to echo that here. Additionally, it's be nice to point out here that VC++ 2017 is directly compatible, for anyone who doesn't know. Lastly, for stable builds, including 1.6 when it does eventually happen, it'd be nice to have a note here about the VC++ versions. Overall seems kind of silly to, on a technical readme for a project, casually ignore dependency versions because someone else's build bot which isn't linked here either, mentions it in passing.

lightningterror:
Good point, I'll update it Soon™.

README.md
```diff
@@ -17,28 +17,30 @@ Development builds are also available from [our website](https://pcsx2.net/downl
 
 ## Minimum
 * OS: Windows Vista SP2 or newer or GNU/Linux (32-bit or 64-bit)
-* CPU: Any that supports SSE2 (Pentium 4 and up, Athlon64 and up)
+* CPU: Any that supports SSE2 (Pentium 4 and up, Athlon64 and up) @ 1600 STR or better
```
ssakash: Most of the common users won't even be aware of the term ``STR`` and the value you're referencing here is from testcases on passmark single thread benchmarks and you don't seem to have any mentions of that either. I feel this would just look cryptic for any random user.

README.md
```diff
@@ -17,28 +17,30 @@ Development builds are also available from [our website](https://pcsx2.net/downl
...
+* CPU: Intel Haswell (or AMD equivalent) @ 2000 STR or better
 * GPU: DirectX 11 GPU or greater
 * RAM: 4GB or more
 
 ## Notes
 
 - You need the [Visual C++ 2015 x86 Redistributables](https://www.microsoft.com/en-us/download/details.aspx?id=48145) for this version to work.
 
-- PCSX2 1.4.0 is the last version to support Windows XP. Windows XP is no longer getting updates (including security-related updates), and graphics drivers for Windows XP are older and no longer maintained.
+- PCSX2 1.4.0 is the last version to support Windows XP and Direct3D9. Windows XP is no longer getting updates (including security-related updates), and graphics drivers for Windows XP are older and no longer maintained.
```
ssakash: ``last stable version`` maybe?

README.md
```diff
@@ -17,28 +17,30 @@ Development builds are also available from [our website](https://pcsx2.net/downl
...
 
 - Make sure to update your operating system, drivers, and DirectX (if applicable) to ensure you have the best experience possible. Having a newer GPU is also recommended so you have the latest supported drivers.
 
 - Because of copyright issues, and the complexity of trying to work around it, you need a BIOS dump extracted from a legitimately-owned PS2 console to use the emulator. For more information about the BIOS and how to get it from your console, visit [this page](https://pcsx2.net/config-guide/official-english-pcsx2-configuration-guide.html#Bios).
 
 - PCSX2 mainly takes advantage of 2 CPU cores. As of [this commit](https://github.com/PCSX2/pcsx2/commit/ac9bf45) PCSX2 can now take advantage of more than 2 cores using the MTVU speedhack. This can be a significant speedup on CPUs with 3+ cores, but it may be a slowdown on GS-limited games (or on CPUs with fewer than 2 cores).
 
+- You can visit [this page](https://www.cpubenchmark.net/singleThread.html) to find out your CPU's STR (Single Threaded Rating).
```
ssakash: Oh fuck, I didn't look at this.

ssakash: What about a hyperlink to the original text in the first place or mention about a reference on it to this content?

ssakash: Since there's the possibility that people might overlook this part.

RedPanda4552: Agreed. I would _open up_ the requirements section with a note in italics something like

> _Requirements benchmarks are based on a statistic from the Passmark CPU bench marking software. When we say "STR", we are referring to Passmark's "Single Thread Rating" statistic. You can look up your CPU on https://cpubenchmark.net to see how it compares to PCSX2's requirements._

More or less. Probably want to be a bit less wordy.

lightningterror: Italic isn't needed but I'll go with this suggestion instead.

lightningterror:
Is everyone fine with the changes ?

ramapcsx2:
Sure, it's an all around improvement.
The use of acronyms isn't perfect, but the old GHz rating is arguably worse. With "STR", you can at least use that for a realistic comparison.

PR #2909, opened 2019-03-29
GSdx: Log private GS registers when doing raw dump

Souzooka:
This adds private GS registers to the context logs during dumps.


gregory38:
Do you really need those priv register value every draw call ?
Or is it enough before the frame presentation ? (I.e. once per frame)

ssakash:
I don't think we need it per draw call, once per frame is enough.

PR #2911, opened 2019-03-31
Pnach: Add Author line

knight-ryu12:
(actually ref to @lightningterror  and my conversation)
You can use it to add Author to indicate who made this patch.

### DIFF
--- a/pcsx2/Patch.cpp
+++ b/pcsx2/Patch.cpp
@@ -285,7 +285,7 @@ namespace PatchFunc
 
 	void author(const wxString& text1, const wxString& text2)
 	{
-        PatchesCon->WriteLn(L"Author: " + text2);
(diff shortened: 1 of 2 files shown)


pcsx2/Patch.cpp
```diff
@@ -282,6 +283,11 @@ namespace PatchFunc
 		PatchesCon->WriteLn(L"comment: " + text2);
 	}
 
+	void author(const wxString& text1, const wxString& text2)
+	{
+        PatchesCon->WriteLn(L"Author: " + text2);
```
lightningterror: Tabs instead of spaces.

knight-ryu12: Fixed

knight-ryu12:
I think it's might be wise to trail "Fix GameDB to use those author line" PR after this get merged.

Issue #2894, opened 2019-03-19
GSdx: reduce HW texture cache overhead when creating a source from a target


gregory38:
When the texture cache hit an older target/depth, it will create a new input source
```GSTextureCache::Source* GSTextureCache::CreateSource```

In most case, it will either copy/rescale/convert the old target. However in most case it is either slow/reduce quality/useless.

Here the code that we want to avoid
```
		if ((sRect == dRect).alltrue() && is_copy)
		{
			if (half_right) {
				// You typically hit this code in snow engine game. Dstsize is the size of of Dx/GL RT
				// which is arbitrary set to 1280 (biggest RT used by GS). h/w are based on the input texture
				// so the only reliable way to find the real size of the target is to use the TBW value.
				float real_width = dst->m_TEX0.TBW * 64u * dst->m_texture->GetScale().x;
				m_renderer->m_dev->CopyRect(sTex, dTex, GSVector4i((int)(real_width/2.0f), 0, (int)real_width, h));
			} else {
				m_renderer->m_dev->CopyRect(sTex, dTex, GSVector4i(0, 0, sRect.z, sRect.w)); // <= likely wrong dstsize.x could be bigger than w
			}
		}
		else
		{
			// Different size or not the same format
			sRect.z /= sTex->GetWidth();
			sRect.w /= sTex->GetHeight();

			if (half_right) {
				sRect.x = sRect.z/2.0f;
			}

			m_renderer->m_dev->StretchRect(sTex, sRect, dTex, dRect, shader, linear);
		}
```

The idea is to defer this code later in the renderer in order to make better choice. Note StretchRect can't be called after `BeginScene`

Case1: current texture and framebuffer overlap. Because of the overlapping, the source will be deleted at the end of the draw...
* OpenGL can sample directly from frame buffer. So copy/rescaling is useless, but it would be slow if there are too many primitives (see Dx below for an extra optimization)
* DirectX can't sample the frame buffer. However we could limit the copy based on the draw call value. Note: it could help OpenGL too if there is too many primitives. For example, on Jak or Star Ocean, the copy will be limited to the shadow area rather than the full framebuffer. 

Case2: current texture and framebuffer don't overlap. Therefore it is possible to bind both during rendering. However remain conversion/scaling issue.
* If conversion is needed -> no choice it must be done
* If scaling is needed -> it can potentially be done on the coordinate level. It might depends on current clamping mode.
[...]

PR #2921, opened 2019-04-10
gsdx ogl: bypass the texture cache when the frame buffer will be sampled

gregory38:
Proof of concept. It should provide a huge speedup when accurate
blending is enabled for tri-ace / Jak / R&C series (shadow rendering)

See #2894
Need PR #2892

### DIFF
--- a/plugins/GSdx/Renderers/HW/GSHwHack.cpp
+++ b/plugins/GSdx/Renderers/HW/GSHwHack.cpp
@@ -1690,11 +1690,6 @@ void GSState::SetupCrcHack()
 		lut[CRC::DeathByDegreesTekkenNinaWilliams] = GSC_DeathByDegreesTekkenNinaWilliams; // + Upscaling issues
 		lut[CRC::SonicUnleashed] = GSC_SonicUnleashed; // + Half screen on texture shuffle
 
-		// These games emulate a stencil buffer with the alpha channel of the RT (too slow to move to Aggressive)
-		lut[CRC::RadiataStories] = GSC_RadiataStories;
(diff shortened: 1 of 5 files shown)


lightningterror:
Master barely reaching 20 fps on native.
PR reaching 94 fps on native.

Well this is a drastic change xD

gregory38:
I did say huge speedup. Good. It is a bit hackish. I will let kojin improve it ;)

Note: maybe a dynamic cast would be better.

lightningterror:
We could also move the crc hacks to dx only now.

gregory38:
Need to test multiple shadow rendering to be sure. And others game too. And it is only a proof of concept. Not sure it is the best way to implement it.

MrCK1:
I got a 20FPS improvement in Hitman Contracts (32->52 at 3x Native w/ Basic Blending), the average framerate still fluctuates a bit standing still

gregory38:
Interesting to know. I should add a comment in the code. Does the area contains complex shadows.

gregory38:
I added a const and put some comments. @lightningterror feel free to add CRC change. And decide to merge it or not ;)

gregory38:
It is far from idea. But I think the gains are big enough to have it for 1.6. Until someone does a better job.

MrCK1:
I'll run some more tests ;)

lightningterror:
> Until someone does a better job.

Greg from the future :P
Btw what about https://github.com/PCSX2/pcsx2/pull/2843#issuecomment-470906904 when no blending is used?

gregory38:
Can't do everything. It is a part of not an ideal solution.

PR #2923, opened 2019-04-13
GameDB: WRC avec sebastien loeb performance fixes

atomic83GitHub:
This PR add a modified patch to "WRC avec sebastien loeb", this fixes a very bad performance problem with a delay slot instruction which for an unknow reason (Timing?) tanks the performances when using physics in the playground mode.

With the delay slot instruction:

Capture

Without the delay slot instruction:

Capture2

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -2281,6 +2281,12 @@ Serial = SCES-53680
 Name   = WRC avec Sebastien Loeb
 Region = PAL-M5
 XgKickHack = 1 // SPS.
+[patches = 79BAD675]
+	author=By Prafull
+	// Nopping a branch in delay slot instruction causing very bad performances
+	// in the playground mode when using physics (such as moving barrels etc..).
+	patch=1,EE,003d2394,word,00000000
+[/patches]
 ---------------------------------------------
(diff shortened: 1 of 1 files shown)


lightningterror:
Curious, how does the game behave on older builds before the delay branch pr was merged with that patch line commented out.

atomic83GitHub:
It is exactly the same with the older build, if this address is commented the game will have the same slowdowns.

bin/GameIndex.dbf
```diff
@@ -2281,6 +2281,12 @@ Serial = SCES-53680
 Name   = WRC avec Sebastien Loeb
 Region = PAL-M5
 XgKickHack = 1 // SPS.
+[patches = 79BAD675]
+	author=By Prafull and atomic83github
```
lightningterror: It will be better to have Prafull as author only since there aren't any actual modifications made to the patch.

lightningterror:
> It is exactly the same with the older build, if this address is commented the game will have the same slowdowns.

Thanks, just the info I needed to make sure the branch delay pr didn't cause any new issues.

aphelionosu:
This works well on WRC: Rally Evolved too. Big thanks

PR #2925, opened 2019-04-13
GSDumpGui: Automatically find the next free TCP port by setting requested port to 0

willkuer:
Fixes multiinstance crash as now every instance has its own port

### DIFF
--- a/tools/GSDumpGUI/Core/Program.cs
+++ b/tools/GSDumpGUI/Core/Program.cs
@@ -52,16 +52,17 @@ namespace GSDumpGUI
         [STAThread]
         static void Main(String[] args)
         {
-            if (args.Length == 4)
+            if (args.Length == 5)
             {
                 // do this first, else racy mess ;)
                 wrap = new GSDXWrapper();
+                var port = Convert.ToInt32(args[4]);
 
                 try
                 {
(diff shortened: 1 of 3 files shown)


willkuer:
fixes multi instance crash reported in #1637

willkuer:
Ok now it works. GsDumpGui handles multiple windows by via launching itsself several times and communicating between each instance via TCP instead of launching multiple windows in the same process. Honestly this is in some sense amazing and I wonder what the intention was.

It used a static port (9999) before but now will use the next free port the OS yields. This needs to be communicated to the clients and hence command line args needed to be touched.

lightningterror:
Gave it a test and works as expected, both crashes when launching multiple instances or copy pasting a new exe have been resolved.

Are there any other pending changes or it's ready for merge?

willkuer:
Done from my point of view.

Except you’d like to see the whole TCP/interprocess communication part removed. I really don’t get why it was done like that. Maybe to save the main gui in case the gs dump in the client crashes? Still it’s a questionable design. I don’t see it in the scope of this pr though.

lightningterror:
Yeah higher priority are the crashes, improving the functionality can be done later.

PR #2926, opened 2019-04-13
GSDumpGui: Only reload dumps and dlls separately if their respective settings change, make config portable, autodetect dump/dll folder

willkuer:
- Add simple logging infrastructure and further improve the existing one
- Delegate loading of dlls and dumps into dedicated loader classes
- Adjust user interaction to restrict updates to only relevant parts
- Gsdx dlls (dumps) are only reloaded when a new path for gsdx dlls (dumps) was given. That means dlls are not reloaded just because the dumps path was changed or another dump was selected
- When GSDumpGui can not find the specified settings folder or there is no setting saved so far it will look into the directory of the currently running application (however as before only exactly in the directory and not in subdirectories)
- Further decouple model, view and 'controller' logic.
- Restrict directory checks to application basepath instead of current directory as current directly is changed at too man places
- Avoid some minor memory leaks by disposing some disposable elements


willkuer:
Pull requests fixes two initialization "bugs" mentioned in #1637 

1. Gsdx dlls (dumps) are only reloaded when a new path for gsdx dlls (dumps) was given. That means dlls are not reloaded just because the dumps path was changed or another dump was selected
2. When GSDumpGui can not find the specified settings folder or there is no setting saved so far it will look into the directory of the currently running application (however as before only exactly in the directory and not in subdirectories)

tools/GSDumpGUI/Forms/Helper/RichTextBoxLogger.cs
```diff
@@ -0,0 +1,59 @@
...
+        private readonly RichTextBox  _richTextBox;
+        public RichTextBoxLogger(RichTextBox  richTextBox)
+        {
+            _richTextBox = richTextBox;
+            _richTextBox.BackColor = Color.White;
+        }
+
+        private void WriteLine(Color color, string line = null)
+        {
+            if (line is null)
```
lightningterror: I'm getting a compiler warning here expecting a type.

willkuer: I am using vs2017 most probably you have an older version. Probably this is c#6/7. I can rewrite it without using modern language features.

lightningterror: Yes I'm using VS2015. It's a good idea to rewrite it since the project supports 2015.

willkuer:
According to https://www.guru99.com/c-sharp-dot-net-version-history.html vs2015 supports c#6. I can not check as I don't have it installed. I restricted the language setting to c#6 to not produce such incompatible commits in the future. Can to be adjusted once pcsx2 moves forward to more recent versions..

tools/GSDumpGUI/Forms/Helper/RichTextBoxLogger.cs
```diff
@@ -0,0 +1,59 @@
...
+        private readonly RichTextBox  _richTextBox;
+        public RichTextBoxLogger(RichTextBox  richTextBox)
+        {
+            _richTextBox = richTextBox;
+            _richTextBox.BackColor = Color.White;
+        }
+
+        private void WriteLine(Color color, string line = null)
+        {
+            if (line == null)
```
lightningterror: Yeah now it compiles fine.

lightningterror:
> When GSDumpGui can not find the specified settings folder or there is no setting saved so far it will look into the directory of the currently running application (however as before only exactly in the directory and not in subdirectories)

It works for dlls but not for dumps, also a suggestion maybe save the path config folder in gsdumpgui directory instead of AppData ?

willkuer:
There was a bug in the autoselection of files in case an gsdx dll was found. That was because the gsdxwrapper was changing the working directory of the current application. It should work now under all cases.

willkuer:
I need to add copyright header 😠

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -177,34 +181,35 @@ private void cmdBrowseDumps_Click(object sender, EventArgs e)
             fbd.SelectedPath = AppDomain.CurrentDomain.BaseDirectory;
             if (fbd.ShowDialog() == DialogResult.OK)
                 txtDumpsDirectory.Text = fbd.SelectedPath;
```
willkuer: missed this one.

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -91,72 +125,42 @@ public GSDumpGUI()
...
         {
-            txtIntLog.Text += "Starting GSdx Loading Procedures" + Environment.NewLine + Environment.NewLine;
-
-            txtGSDXDirectory.Text = Properties.Settings.Default.GSDXDir;
-            txtDumpsDirectory.Text = Properties.Settings.Default.DumpDir;
+            _internalLogger.Information("Starting GSdx Loading Procedures");
 
-            lstGSDX.Items.Clear();
-            lstDumps.Items.Clear();
+            var gsdxFolder = _folderWithFallBackFinder.GetViaPatternWithFallback(Settings.GSDXDir, "*.dll", "", "plugins", "dll", "dlls");
```
willkuer: here gsdx dll fallback folders relative to gsdxgui.exe are given

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -91,72 +125,42 @@ public GSDumpGUI()
...
-                        txtIntLog.Text += "Failed to load \"" + itm + "\". Is it really a GSdx DLL?" + Environment.NewLine;
-                    }
-                }
-            }
+        private void ReloadGsdxDumps()
+        {
+            _internalLogger.Information("Starting GSdx Dump Loading Procedures...");
 
-            txtIntLog.Text += Environment.NewLine + "Completed GSdx Loading Procedures" + Environment.NewLine + Environment.NewLine;
+            var dumpFolder = _folderWithFallBackFinder.GetViaPatternWithFallback(Settings.DumpDir, "*.gs", "", "dumps", "gsdumps");
```
willkuer: here gs dump fallback folders relative to gsdxgui.exe are given

tools/GSDumpGUI/GSDumpGUI.csproj
```diff
@@ -10,14 +10,15 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>GSDumpGUI</RootNamespace>
     <AssemblyName>GSDumpGUI</AssemblyName>
-    <TargetFrameworkVersion>v2.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
```
lightningterror: Is there a need to up the framework version, if it's not required then we can skip it.

lightningterror:
There seems to be a new issue with the image preview not being shown on first bootup of the application, selecting a different dump makes it work.

tools/GSDumpGUI/GSDumpGUI.csproj
willkuer: I originally wanted to use a nuget package but was lost in licence hell and wrote the code myself. I can put it back.

willkuer:
Yes I found the same. Should be fixed now.

There were potentially also other bugs with autodetection once you open the gsdx config. Also these were fixed. Code is a bit chaotic and highly stateful so maybe I oversaw something.

tools/GSDumpGUI/GSDumpGUI.csproj
willkuer: Got the same result. 4.0 is minimum now.

I think most people have at least 4.7 on their machines so I don't think anyone is negatively affected here.

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -315,34 +327,30 @@ private void cmdOpenIni_Click(object sender, EventArgs e)
...
+            {
+                pctBox.Load(_availableGsDumps.Selected.PreviewFile.FullName);
+                pctBox.Cursor = Cursors.Hand;
+            }
+
+            pctBox.Tag = _availableGsDumps.Selected?.PreviewFile?.FullName;
         }
 
-        private void pctBox_Click(object sender, EventArgs e)
+        private static void PreviewImageClick(object sender, EventArgs e)
```
lightningterror: Maybe just private void? Otherwise for me it fails to compile currently like this.

willkuer:
And I think I forgot to mention: configs are now local and not stored in appdata anymore

tools/GSDumpGUI/Forms/frmMain.cs
willkuer: No the `this` in the designer needs to be removed. I think git pull should contain the fix.

lightningterror: Ah yea that did it.

tools/GSDumpGUI/GSDumpGUI.csproj
lightningterror: > 4.0 is compatible to XP in case anyone cares.

I don't think anyone cares at this point :)

Btw I'm on 4.6.1 so not really on latest.

lightningterror:
Another neat feature might be to save the selected dump and dll (renderer too maybe) upon closing/reopening the application since it's very helpful when recompiling/testing the same dll on a specific dump.

willkuer:
The list of dumps/dlls however is created on runtime. What is the expected behavior in case the dump/dll is removed? Reset to default (i.e. first element in the list for each) or select none?

lightningterror:
Select first on the list if dump or dll that was saved is removed.

Btw if you plan on adding that we can do it on another pr and merge this one now if it's ready.

willkuer:
I’d say this one is ready. At least I have no known additions of bugs I need to fix.

Thanks for testing btw.

lightningterror:
Okay let's merge it, this is a huge improvement over current master branch, I didn't spot any other new bugs so far but we can mention/tackle them on the issue thread once we find anything.

lightningterror:
I'm sure everyone will be happy with these changes, I know @tadanokojin was losing his mind with the mentioned issues :)

tools/GSDumpGUI/GSDumpGUI.csproj
```diff
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>GSDumpGUI</RootNamespace>
     <AssemblyName>GSDumpGUI</AssemblyName>
-    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
```
pgert: Just some info:
From what I have understood, 4.6.0 is the latest (or last) version supported by Vista & Windows Server 2008.
https://docs.microsoft.com/en-us/dotnet/framework/install/on-windows-vista

lightningterror: Well if by some miracle someone still uses Vista they don't have to worry since the requirement is 4.0 so it's all good.

willkuer: Actually I consider to port it to wpf/netcore3.0 once it is available. Then it would as well run natively under linux. I wonder how gregory is using it tbh.

gregory38:
I created a basic loader to launch the .so
Ini is open in text editor to tube option.
Then I can dump any state either from gsdx or any GPU debugger.

gregory38:
If you want a bigger project, you could code a kinds of GS debugger.
Step by step draw.
See gs register states.
Allow to read anywhere from gs mem with any fmt.
With default based on tex0/frame/depth register value
...

tadanokojin:
@willkuer Thank you for your hard work. It's much less arse to use now. My hero.

willkuer:
> If you want a bigger project, you could code a kinds of GS debugger.
> Step by step draw.
> See gs register states.
> Allow to read anywhere from gs mem with any fmt.
> With default based on tex0/frame/depth register value
> ...

The problem is that I have no idea how gpu's and even less how the gs works. Most probably I couldn't write the gsdumpgui deserialization logic myself. Also while I worked with unmanaged code called from managed code via marshalling I am not proficient in it. Maybe apart from threading my low-level skills are really bad - there’s typically no need to know such stuff as a c# dev. Debugging issues could make me invest a lot of time without helping you at all. Especially since apparently almost everything is allowed and debugging gs problems can be really PITA.

If you tell me that there is this public gsdx api that you want to feed with this and that I can make it happen. But I am basically restricted to the gsdx api.

willkuer:
> @willkuer Thank you for your hard work. It's much less arse to use now. My hero.

I could further improve usability if you see issues/improvements. I actually did not really use the tool and hence have no idea what kind of steps you do over and over again that could be improved. I.e. since I don‘t know the product I can not design UX.

I would look into the compatibility stuff with intel gpu and x64 dlls next but I guess they are rather low prio.

PR #2928, opened 2019-04-18
GSdx-d3d11:  Add partial support for frame buffer masking

lightningterror:
It works on games such as Fifa Street 1 and 2 (character and stage
rendering), mission impossible operation surma (shadow rendering).

Note: Frame buffer masking is only doable on alpha channel on d3d11
so the feature is not complete unlike on gl but it's a nice improvement overall.

Purge the old aout (Alpha hack) code since fbmask emulation will do a better accurate job.
Enable blending option on d3d to toggle fbmask emulation just like on opengl.
At least basic blending option is required to enable it.


lightningterror:
Ok will merge the initial pr now, it's a lot better than what we currently have and it's a nice Easter present for D3D users :)
More improvements can be done afterwards, I have ideas for multiple blending levels for fbmask.

PR #2931, opened 2019-04-20
Fix 2 minor typos

avindra:
* fixes typos on the tooltip for disabling safe features


lightningterror:
I don't know how I never noticed these ...

PR #2932, opened 2019-04-21
Onepad (legacy) sensibility implementation for linux

y-salnikov:
Sensibility slider now can be used to set sensibility of gamepad axis in linux version of OnePad plugin. Perhaps it is done in wrong way at wrong place.


arcum42:
Ran into a bug while testing this, which was not so much with the new sensibility code as the existing code, so I fixed that and made it so you can actually tell what values are set in the dialog while I was at it.

arcum42:
Think I'll go ahead and merge this in. It *is* possible to set the sensibility low enough that it doesn't register joystick movement, but if someone moves it most of the way to the left, it seems reasonable to assume that'd be a possibility. If I can figure out some way to adjust it to not let you set it in a dead zone, I will.

PR #2933, opened 2019-04-21
Fixing stiks state monitor in onepad (legacy) configuration dialog.

y-salnikov:
There was assumption in code, that SDL enumerate left horizontal axes as 0, vertical as 1,  right horizontal as 2 and right vertical as 3. In general case it is not true.


arcum42:
I don't have any test cases where it isn't true, but not making assumptions seems reasonable, and it doesn't break anything for me. I'm sure if it does for other people, they'll let me know.

PR #2947, opened 2019-04-30
GSdx-sw: Add an option to toggle auto flush on sw renderer

lightningterror:
Windows + Linux.
@arcum42 can you take a look at the gui option for linux, I can't test if it's placed/working properly.


plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -141,7 +141,8 @@ const char* dialog_message(int ID, bool* updateText) {
 			return "Enables external shader for additional post-processing effects.";
 		case IDC_FXAA:
 			return "Enables fast approximate anti-aliasing. Small performance impact.";
-		case IDC_AUTO_FLUSH:
+		case IDC_AUTO_FLUSH_HW:
+		case IDC_AUTO_FLUSH_SW:
```
lightningterror: Maybe we can use a slightly different tooltip for sw option.

Issue #2555, opened 2018-08-15
[Debugger] Editing read-only memory pages causes access violation


Souzooka:
***PCSX2 version:***
PCSX2 v1.5.0-dev-2457

- **Notes**:
The access violation occurs at line 237 of vtlb.cpp, which causes an instant hard crash of PCSX2.

***PCSX2 options:***
N/A

***Plugins used:***
N/A

***Description of the issue:***
Changing the value of an address at a memory page marked as read-only using the debug window causes an access violation.

***How to reproduce the issue:***
Open the debug window, then attempt to edit memory in a read-only memory page (e.g. in a code section). Editing address 0x100000 should cause this crash for any game.

***Last known version to work:***
N/A

***PC specifications:***
N/A

ramapcsx2:
What do you expect to happen instead?

refractionpcsx2:
The only sensible solution I can think of is make it not editable in the debugger

gregory38:
There is no such thing as read only page for RAM. However lower part of memory isn't mapped in the virtual user space because it is reserved for the kernel.

F0bes:
As of today, writing to code memory and unmapped memory doesn't seem to cause PCSX2 to crash.
I do know that if you end up writing to MMIO FIFOs it may cause serious issues. This is a side effect of the debugger accessing memory through the vtlb without any way to prevent the vtlb of deferring the read/write to something like the IPU or VIF. I think this would be a user issue anyways, if you don't know where you are writing to, don't write there.

PR #2960, opened 2019-05-14
debugger: Muti-byte hex strings can now be pasted into the memory view

chaoticgd:
You can now paste a hex string into the memory view (with Ctrl+V). Needs testing on Windows.

Peek 2019-05-14 22-49

### DIFF
--- a/bin/docs/debugger.txt
+++ b/bin/docs/debugger.txt
@@ -32,6 +32,7 @@ memory view:
     right click     open context menu
     ctrl+wheel      zoom memory view
     esc             return to previous goto address
+    ctrl+v          paste a hex string into memory
 
 breakpoint list:
 
--- a/pcsx2/gui/Debugger/CtrlMemView.cpp
+++ b/pcsx2/gui/Debugger/CtrlMemView.cpp
@@ -489,6 +489,10 @@ void CtrlMemView::keydownEvent(wxKeyEvent& evt)
 				}
 			}
 			break;
+		case 'v':
+		case 'V':
+			pasteHex();
+			break;
 		default:
 			evt.Skip();
 			break;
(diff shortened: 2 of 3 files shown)


ramapcsx2:
It looks like it should work on Windows but I'll check it out later.
I guess there's no extensive checking for malicious code injection or anything like that?

chaoticgd:
> It looks like it should work on Windows but I'll check it out later.
> I guess there's no extensive checking for malicious code injection or anything like that?

Malicious code injection was already possible by just sending multiple key events to the window, one after the other, as with any debugger. This just makes it easier for humans to use.

chaoticgd:
I suppose social engineering could be another plausible attack vector, if there was some arbitrary code executable vulnerability in the emulator. If that's enough of a concern, then maybe add a dialog when first opening the debugger warning users not to paste random stuff into the memory view, like how Facebook mitigates XSS.

ramapcsx2:
I realize that there hasn't been any protection before and this won't change it.
It's just something that always comes to mind with features like these.

ramapcsx2:
Editing memory is super crashy here on Windows.
I got it to work a little occasionally. When it didn't crash, the paste function appeared to work fine.
So I guess I'm either doing it wrong, or direct memory editing just isn't stable under Windows.
The pr seems okay.

ramapcsx2:
No change to crash behaviour with the CPU pause commit.

lightningterror:
I'd like any crash related issues to be solved before this is merged.

tadanokojin:
You're talking about #2555 I presume.
Don't see what it's this guy's problem

lightningterror:
Oh if it's a different issue forget what I said then, I assumed it was something new to this pr.

tadanokojin:
Just want to make sure we don't have a goose chase.
I didn't notice any crashes on my end for the record.

lightningterror:
@ramapcsx2 are you fine with the pr?

ramapcsx2:
The pr is okay still, yep.

PR #2973, opened 2019-05-27
GSDX-DX11: Partial port of OGL SW blend to DX11 / changes to blend map of both OGL/DX11.

hibye8313:
Changes:
GSDevice11.h:
- class PSConstantBuffer: Removed variable (MinF) that was not being used and structured to be closer to OGL side.
- class PSSelector: Added variable to for selecting accu blend in the shader.

GSRenderer11.cpp:
- EmulateTextureSampler(): Removed variable (MinF) that was not being used and structured to be similar to OGL.
- DrawPrims(): Added checking/enabling accu blend. Small restructure to alpha test code so it is clearer.

GSRendererOGL.cpp:
- DrawPrims(): Same restructure as above for alpha test clarity.

GSDeviceOGL.h:
- class PSConstantBuffer: Some things seem incorrect here, I'm not sure. Seems as if Update() does not fully update the buffer as it is. I put in a FIXME comment.

GSTextureFX11.cpp:
- SetupPS(): Added defines for shader to enable accu blend.
- SetupOM(): Added changes to the HW blend state when accu blend is enabled.

tfx.fx:
- cbuffer cb1: Changed structure to match OGL side and the changes made above to PSConstantBuffer in 
GSDevice11.h.
- added PS_ACCU_BLEND define
- ps_main(): Added code for doing the accu blend when enabled. Also added formatting to 16 bit output when accu blend is enabled since needed for fixing Tag Team Racing shadows.

GSVertexTrace.cpp:
- CorrectDepthTrace(): seems like this function is unnecessarily complex. Added a FIXME as a comment.

Maybe I made too many changes and it will be hard to review them all? Let me know if I should remove some of structural changes that were not 100% necessary.

### DIFF
--- a/plugins/GSdx/Renderers/Common/GSDevice.h
+++ b/plugins/GSdx/Renderers/Common/GSDevice.h
@@ -114,11 +114,40 @@ public:
 
 #pragma pack(pop)
 
+enum HWBlendFlags
+{
+	// A couple of flag to determine the blending behavior
+	BLEND_A_MAX  = 0x100, // Impossible blending uses coeff bigger than 1
+	BLEND_C_CLR  = 0x200, // Clear color blending (use directly the destination color as blending factor)
+	BLEND_NO_BAR = 0x400, // Doesn't require sampling of the RT as a texture
+	BLEND_ACCU   = 0x800, // Allow to use a mix of SW and HW blending to keep the best of the 2 worlds
+};
+
+// Determines the HW blend function for DX11/OGL
+struct HWBlend { uint16 flags, op, src, dst; };
+
 class GSDevice : public GSAlignedClass<32>
 {
+private:
 	FastList<GSTexture*> m_pool;
+	static std::array<HWBlend, 3*3*3*3 + 1> m_blendMap;
 
 protected:
+	enum : uint16
+	{
+		// HW blend factors
+		SRC_COLOR,   INV_SRC_COLOR,    DST_COLOR,  INV_DST_COLOR,
+		SRC1_COLOR,  INV_SRC1_COLOR,   SRC_ALPHA,  INV_SRC_ALPHA,
+		DST_ALPHA,   INV_DST_ALPHA,    SRC1_ALPHA, INV_SRC1_ALPHA,
+		CONST_COLOR, INV_CONST_COLOR,  CONST_ONE,  CONST_ZERO,
+
+		// HW blend operations
+		OP_ADD, OP_SUBTRACT, OP_REV_SUBTRACT
+	};
+
+	static const int m_NO_BLEND     = 0;
+	static const int m_MERGE_BLEND  = m_blendMap.size() - 1;
+
 	std::shared_ptr<GSWnd> m_wnd;
 	int m_vsync;
 	bool m_rbswapped;
@@ -141,6 +170,7 @@ protected:
 	virtual void DoFXAA(GSTexture* sTex, GSTexture* dTex) {}
 	virtual void DoShadeBoost(GSTexture* sTex, GSTexture* dTex) {}
 	virtual void DoExternalFX(GSTexture* sTex, GSTexture* dTex) {}
+	virtual uint16 ConvertBlendEnum(uint16 generic) = 0; // Convert blend factors/ops from the generic enum to DX11/OGl specific.
 
 public:
 	GSOsdManager m_osd;
@@ -213,6 +243,11 @@ public:
 	void PurgePool();
 
 	virtual void PrintMemoryUsage();
+
+	// Convert the GS blend equations to HW specific blend factors/ops
+	// Index is computed as ((((A * 3 + B) * 3) + C) * 3) + D. A, B, C, D taken from ALPHA register.
+	HWBlend GetBlend(size_t index);
+	uint16 GetBlendFlags(size_t index);
 };
 
 struct GSAdapter
--- a/plugins/GSdx/Renderers/DX11/GSDevice11.h
+++ b/plugins/GSdx/Renderers/DX11/GSDevice11.h
@@ -102,6 +102,7 @@ public:
 		GSVector4i FbMask;
 
 		GSVector4 TC_OffsetHack;
+		GSVector4 Af;
 
 		PSConstantBuffer()
 		{
@@ -113,6 +114,7 @@ public:
 			MskFix = GSVector4i::zero();
 			ChannelShuffle = GSVector4i::zero();
 			FbMask = GSVector4i::zero();
+			Af = GSVector4::zero();
 		}
 
 		__forceinline bool Update(const PSConstantBuffer* cb)
@@ -120,7 +122,8 @@ public:
 			GSVector4i* a = (GSVector4i*)this;
 			GSVector4i* b = (GSVector4i*)cb;
 
-			if(!((a[0] == b[0]) /*& (a[1] == b1)*/ & (a[2] == b[2]) & (a[3] == b[3]) & (a[4] == b[4]) & (a[5] == b[5]) & (a[6] == b[6]) & (a[7] == b[7])).alltrue()) // if WH matches HalfTexel does too
+			if(!((a[0] == b[0]) /*& (a[1] == b1)*/ & (a[2] == b[2]) & (a[3] == b[3]) & (a[4] == b[4]) & (a[5] == b[5]) &
+				(a[6] == b[6]) & (a[7] == b[7]) & (a[9] == b[9])).alltrue()) // if WH matches HalfTexel does too
 			{
 				a[0] = b[0];
 				a[1] = b[1];
@@ -130,6 +133,7 @@ public:
 				a[5] = b[5];
 				a[6] = b[6];
 				a[7] = b[7];
+				a[9] = b[9];
 
 				return true;
 			}
@@ -207,7 +211,12 @@ public:
 				uint32 fbmask:1;
 
 				// Blend and Colclip
+				uint32 blend_a:2;
+				uint32 blend_b:2;
+				uint32 blend_c:2;
+				uint32 blend_d:2;
 				uint32 clr1:1;
+				uint32 hdr:1;
 
 				// Others ways to fetch the texture
 				uint32 channel:3;
@@ -220,7 +229,7 @@ public:
 				uint32 point_sampler:1;
 				uint32 invalid_tex0:1; // Lupin the 3rd
 
-				uint32 _free:27;
+				uint32 _free:18;
 			};
 
 			uint64 key;
@@ -286,6 +295,7 @@ public:
 				uint32 wg:1;
 				uint32 wb:1;
 				uint32 wa:1;
+				uint32 accu_blend:1;
 			};
 
 			struct
@@ -298,7 +308,7 @@ public:
 			uint32 key;
 		};
 
-		operator uint32() {return key & 0x1fff;}
+		operator uint32() {return key & 0x3fff;}
 
 		OMBlendSelector() : key(0) {}
 
--- a/plugins/GSdx/Renderers/DX11/GSDevice11.h
+++ b/plugins/GSdx/Renderers/DX11/GSDevice11.h
@@ -206,7 +206,6 @@ public:
 				uint32 read_ba:1;
 				uint32 fbmask:1;
 
-				// *** Word 2
 				// Blend and Colclip
 				uint32 clr1:1;
 
@@ -309,14 +308,6 @@ public:
 		}
 	};
 
-	struct D3D11Blend
-	{
-		int bogus;
-		D3D11_BLEND_OP op;
-		D3D11_BLEND src, dst;
-	};
-	static const D3D11Blend m_blendMapD3D11[3*3*3*3];
-
 	#pragma pack(pop)
 
 private:
@@ -338,7 +329,7 @@ private:
 	void BeforeDraw();
 	void AfterDraw();
 	
-	//
+	uint16 ConvertBlendEnum(uint16 generic) final;
 
 	CComPtr<ID3D11Device> m_dev;
 	CComPtr<ID3D11DeviceContext> m_ctx;
--- a/plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
+++ b/plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
@@ -501,6 +501,99 @@ void GSRendererDX11::EmulateChannelShuffle(GSTexture** rt, const GSTextureCache:
 	}
 }
 
+void GSRendererDX11::EmulateBlending()
+{
+	// Partial port of OGL SW blending. Currently only works for accumulation blend.
+	const GIFRegALPHA& ALPHA = m_context->ALPHA;
+	bool sw_blending         = false;
+
+	// No blending so early exit
+	if (!(PRIM->ABE || (PRIM->AA1 && m_vt.m_primclass == GS_LINE_CLASS)))
+		return;
+
+	m_om_bsel.abe = 1;
+	m_om_bsel.a = ALPHA.A;
+	m_om_bsel.b = ALPHA.B;
+	m_om_bsel.c = ALPHA.C;
+	m_om_bsel.d = ALPHA.D;
+
+	if (m_env.PABE.PABE)
+	{
+		if (m_om_bsel.a == 0 && m_om_bsel.b == 1 && m_om_bsel.c == 0 && m_om_bsel.d == 1)
+		{
+			// this works because with PABE alpha blending is on when alpha >= 0x80, but since the pixel shader
+			// cannot output anything over 0x80 (== 1.0) blending with 0x80 or turning it off gives the same result
+
+			m_om_bsel.abe = 0;
+		}
+		else
+		{
+			//Breath of Fire Dragon Quarter triggers this in battles. Graphics are fine though.
+			//ASSERT(0);
+		}
+	}
+
+	uint8 blend_index  = uint8(((ALPHA.A * 3 + ALPHA.B) * 3 + ALPHA.C) * 3 + ALPHA.D);
+	int blend_flag = m_dev->GetBlendFlags(blend_index);
+
+	bool accumulation_blend = !!(blend_flag & BLEND_ACCU);
+
+	switch (m_sw_blending)
+	{
+	case ACC_BLEND_HIGH_D3D11:
+	case ACC_BLEND_MEDIUM_D3D11:
+	case ACC_BLEND_BASIC_D3D11:
+		sw_blending |= accumulation_blend;
+	default: break;
+	}
+
+	if (m_env.COLCLAMP.CLAMP == 0)
+	{
+		if (accumulation_blend)
+			sw_blending = true;
+		m_ps_sel.hdr = 1;
+	}
+
+	if (sw_blending)
+	{
+		m_ps_sel.blend_a = ALPHA.A;
+		m_ps_sel.blend_b = ALPHA.B;
+		m_ps_sel.blend_c = ALPHA.C;
+		m_ps_sel.blend_d = ALPHA.D;
+
+		if (accumulation_blend)
+		{
+			m_om_bsel.accu_blend = 1;
+
+			if (ALPHA.A == 2) {
+				// The blend unit does a reverse subtraction so it means
+				// the shader must output a positive value.
+				// Replace 0 - Cs by Cs - 0
+				m_ps_sel.blend_a = ALPHA.B;
+				m_ps_sel.blend_b = 2;
+			}
+			// Remove the addition/substraction from the SW blending
+			m_ps_sel.blend_d = 2;
+		}
+		else
+		{
+			// We shouldn't hit this path as currently only accumulation blend is implemented
+			ASSERT(0);
+		}
+
+		// Require the fix alpha vlaue
+		if (ALPHA.C == 2)
+			ps_cb.Af.x = (float)ALPHA.FIX / 128.0f;
+	}
+	else
+	{
+		m_ps_sel.clr1 = !!(blend_flag & BLEND_C_CLR);
+		// FIXME: When doing HW blending with a 24 bit frambuffer and ALPHA.C == 1 (Ad) it should be handled
+		// as if Ad = 1.0f. As with OGL side it is probably best to set m_om_bsel.c = 1 (Af) and use
+		// AFIX = 0x80 (Af = 1.0f).
+	}
+}
+
 void GSRendererDX11::EmulateTextureSampler(const GSTextureCache::Source* tex)
 {
 	const GSLocalMemory::psm_t &psm = GSLocalMemory::m_psm[m_context->TEX0.PSM];
@@ -776,43 +869,20 @@ void GSRendererDX11::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sou
 	}
 
 	// Blend
-
-	if (!IsOpaque())
+	if (!IsOpaque() && rt)
 	{
-		m_om_bsel.abe = PRIM->ABE || PRIM->AA1 && m_vt.m_primclass == GS_LINE_CLASS;
+		EmulateBlending();
+	}
 
-		m_om_bsel.a = m_context->ALPHA.A;
-		m_om_bsel.b = m_context->ALPHA.B;
-		m_om_bsel.c = m_context->ALPHA.C;
-		m_om_bsel.d = m_context->ALPHA.D;
-
-		if (m_env.PABE.PABE)
-		{
-			if (m_om_bsel.a == 0 && m_om_bsel.b == 1 && m_om_bsel.c == 0 && m_om_bsel.d == 1)
-			{
-				// this works because with PABE alpha blending is on when alpha >= 0x80, but since the pixel shader
-				// cannot output anything over 0x80 (== 1.0) blending with 0x80 or turning it off gives the same result
-
-				m_om_bsel.abe = 0;
-			}
-			else
-			{
-				//Breath of Fire Dragon Quarter triggers this in battles. Graphics are fine though.
-				//ASSERT(0);
-			}
-		}
-
-		// Color clip
-		if (m_env.COLCLAMP.CLAMP == 0 && rt)
-		{
-			// fprintf(stderr, "COLCLIP HDR mode ENABLED\n");
-			GSVector4 dRect(ComputeBoundingBox(rtscale, rtsize));
-			GSVector4 sRect = dRect / GSVector4(rtsize.x, rtsize.y).xyxy();
-			hdr_rt = dev->CreateRenderTarget(rtsize.x, rtsize.y, DXGI_FORMAT_R32G32B32A32_FLOAT);
-			// Warning: StretchRect must be called before BeginScene otherwise
-			// vertices will be overwritten. Trust me you don't want to do that.
-			dev->StretchRect(rt, sRect, hdr_rt, dRect, ShaderConvert_COPY, false);
-		}
+	if (m_ps_sel.hdr)
+	{
+		// fprintf(stderr, "COLCLIP HDR mode ENABLED\n");
+		GSVector4 dRect(ComputeBoundingBox(rtscale, rtsize));
+		GSVector4 sRect = dRect / GSVector4(rtsize.x, rtsize.y).xyxy();
+		hdr_rt = dev->CreateRenderTarget(rtsize.x, rtsize.y, DXGI_FORMAT_R32G32B32A32_FLOAT);
+		// Warning: StretchRect must be called before BeginScene otherwise
+		// vertices will be overwritten. Trust me you don't want to do that.
+		dev->StretchRect(rt, sRect, hdr_rt, dRect, ShaderConvert_COPY, false);
 	}
 
 	if (m_ps_sel.dfmt == 1)
@@ -891,7 +961,6 @@ void GSRendererDX11::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sou
 		}
 	}
 
-	m_ps_sel.clr1 = m_om_bsel.IsCLR1();
 	m_ps_sel.fba = m_context->FBA.FBA;
 
 	if (PRIM->FGE)
--- a/plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
+++ b/plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
@@ -482,7 +482,7 @@ void GSRendererOGL::EmulateBlending(bool DATE_GL42)
 
 	// Compute the blending equation to detect special case
 	uint8 blend_index  = uint8(((ALPHA.A * 3 + ALPHA.B) * 3 + ALPHA.C) * 3 + ALPHA.D);
-	int blend_flag = GSDeviceOGL::m_blendMapOGL[blend_index].bogus;
+	int blend_flag = dev->GetBlendFlags(blend_index);
 
 	// SW Blend is (nearly) free. Let's use it.
 	bool impossible_or_free_blend = (blend_flag & (BLEND_NO_BAR|BLEND_A_MAX|BLEND_ACCU)) // Blend doesn't requires the costly barrier
--- a/plugins/GSdx/res/tfx.fx
+++ b/plugins/GSdx/res/tfx.fx
@@ -44,7 +44,14 @@
 #define PS_URBAN_CHAOS_HLE 0
 #define PS_INVALID_TEX0 0
 #define PS_SCALE_FACTOR 1
+#define PS_HDR 0
+#define PS_BLEND_A 0
+#define PS_BLEND_B 0
+#define PS_BLEND_C 0
+#define PS_BLEND_D 0
 #endif
+
+#define SW_BLEND (PS_BLEND_A || PS_BLEND_B || PS_BLEND_D)
 
 struct VS_INPUT
 {
@@ -106,6 +113,8 @@ cbuffer cb1
 	int4 ChannelShuffle;
 	uint4 FbMask;
 	float4 TC_OffsetHack;
+	float Af;
+	float3 _pad;
 };
 
 cbuffer cb2
--- a/plugins/GSdx/res/tfx.fx
+++ b/plugins/GSdx/res/tfx.fx
@@ -880,6 +880,37 @@ void gs_main(line VS_OUTPUT input[2], inout TriangleStream<VS_OUTPUT> stream)
 
 #endif
 
+void ps_blend(inout float4 Color, float As, float2 pos_xy)
+{
+	if (SW_BLEND)
+	{
+		float4 RT = RtSampler.Load(int3(pos_xy, 0));
+
+		float3 Cs = trunc(Color.rgb * 255.0f + 0.1f);
+		float3 Cd = trunc(RT.rgb * 255.0f + 0.1f);
+		float3 Cv;
+
+		float Ad = (PS_DFMT == FMT_24) ? 1.0f : (RT.a * 255.0f / 128.0f);
+
+		float3 A = (PS_BLEND_A == 0) ? Cs : ((PS_BLEND_A == 1) ? Cd : (float3)0.0f);
+		float3 B = (PS_BLEND_B == 0) ? Cs : ((PS_BLEND_B == 1) ? Cd : (float3)0.0f);
+		float3 C = (PS_BLEND_C == 0) ? As : ((PS_BLEND_C == 1) ? Ad : Af);
+		float3 D = (PS_BLEND_D == 0) ? Cs : ((PS_BLEND_D == 1) ? Cd : (float3)0.0f);
+
+		Cv = (PS_BLEND_A == PS_BLEND_B) ? D : trunc(((A - B) * C) + D);
+
+		// Standard Clamp
+		if (PS_HDR == 0)
+			Cv = clamp(Cv, (float3)0.0f, (float3)255.0f);
+
+		// In 16 bits format, only 5 bits of color are used. It impacts shadows computation of Castlevania
+		if (PS_DFMT == FMT_16)
+			Cv = (float3)((int3)Cv & (int3)0xF8);
+
+		Color.rgb = Cv / 255.0f;
+	}
+}
+
 PS_OUTPUT ps_main(PS_INPUT input)
 {
 	float4 c = ps_color(input);
@@ -916,6 +947,10 @@ PS_OUTPUT ps_main(PS_INPUT input)
 	// Must be done before alpha correction
 	float alpha_blend = c.a * 255.0f / 128.0f;
 
+	// Blending
+	ps_blend(c, alpha_blend, input.p.xy);
+
+	// Alpha correction
 	if (PS_DFMT == FMT_16) // 16 bit output
 	{
 		float a = 128.0f / 255; // alpha output will be 0x80
--- a/plugins/GSdx/Renderers/Common/GSDevice.cpp
+++ b/plugins/GSdx/Renderers/Common/GSDevice.cpp
@@ -423,3 +423,100 @@ GSAdapter::GSAdapter(const DXGI_ADAPTER_DESC1 &desc_dxgi)
 #ifdef __linux__
 // TODO
 #endif
+
+HWBlend GSDevice::GetBlend(size_t index)
+{
+	HWBlend blend = m_blendMap[index];
+	blend.op  = ConvertBlendEnum(blend.op);
+	blend.src = ConvertBlendEnum(blend.src);
+	blend.dst = ConvertBlendEnum(blend.dst);
+	return blend;
+}
+
+uint16 GSDevice::GetBlendFlags(size_t index) { return m_blendMap[index].flags; }
+
+std::array<HWBlend, 3*3*3*3 + 1> GSDevice::m_blendMap =
+{{
+	{ BLEND_NO_BAR               , OP_ADD          , CONST_ONE       , CONST_ZERO}      , // 0000: (Cs - Cs)*As + Cs ==> Cs
+	{ 0                          , OP_ADD          , CONST_ZERO      , CONST_ONE}       , // 0001: (Cs - Cs)*As + Cd ==> Cd
+	{ BLEND_NO_BAR               , OP_ADD          , CONST_ZERO      , CONST_ZERO}      , // 0002: (Cs - Cs)*As +  0 ==> 0
+	{ BLEND_NO_BAR               , OP_ADD          , CONST_ONE       , CONST_ZERO}      , // 0010: (Cs - Cs)*Ad + Cs ==> Cs
+	{ 0                          , OP_ADD          , CONST_ZERO      , CONST_ONE}       , // 0011: (Cs - Cs)*Ad + Cd ==> Cd
+	{ BLEND_NO_BAR               , OP_ADD          , CONST_ZERO      , CONST_ZERO}      , // 0012: (Cs - Cs)*Ad +  0 ==> 0
+	{ BLEND_NO_BAR               , OP_ADD          , CONST_ONE       , CONST_ZERO}      , // 0020: (Cs - Cs)*F  + Cs ==> Cs
+	{ 0                          , OP_ADD          , CONST_ZERO      , CONST_ONE}       , // 0021: (Cs - Cs)*F  + Cd ==> Cd
+	{ BLEND_NO_BAR               , OP_ADD          , CONST_ZERO      , CONST_ZERO}      , // 0022: (Cs - Cs)*F  +  0 ==> 0
+	{ BLEND_A_MAX                , OP_SUBTRACT     , CONST_ONE       , SRC1_ALPHA}      , //*0100: (Cs - Cd)*As + Cs ==> Cs*(As + 1) - Cd*As
+	{ 0                          , OP_ADD          , SRC1_ALPHA      , INV_SRC1_ALPHA}  , // 0101: (Cs - Cd)*As + Cd ==> Cs*As + Cd*(1 - As)
+	{ 0                          , OP_SUBTRACT     , SRC1_ALPHA      , SRC1_ALPHA}      , // 0102: (Cs - Cd)*As +  0 ==> Cs*As - Cd*As
+	{ BLEND_A_MAX                , OP_SUBTRACT     , CONST_ONE       , DST_ALPHA}       , //*0110: (Cs - Cd)*Ad + Cs ==> Cs*(Ad + 1) - Cd*Ad
+	{ 0                          , OP_ADD          , DST_ALPHA       , INV_DST_ALPHA}   , // 0111: (Cs - Cd)*Ad + Cd ==> Cs*Ad + Cd*(1 - Ad)
+	{ 0                          , OP_SUBTRACT     , DST_ALPHA       , DST_ALPHA}       , // 0112: (Cs - Cd)*Ad +  0 ==> Cs*Ad - Cd*Ad
+	{ BLEND_A_MAX                , OP_SUBTRACT     , CONST_ONE       , CONST_COLOR}     , //*0120: (Cs - Cd)*F  + Cs ==> Cs*(F + 1) - Cd*F
+	{ 0                          , OP_ADD          , CONST_COLOR     , INV_CONST_COLOR} , // 0121: (Cs - Cd)*F  + Cd ==> Cs*F + Cd*(1 - F)
+	{ 0                          , OP_SUBTRACT     , CONST_COLOR     , CONST_COLOR}     , // 0122: (Cs - Cd)*F  +  0 ==> Cs*F - Cd*F
+	{ BLEND_NO_BAR | BLEND_A_MAX , OP_ADD          , CONST_ONE       , CONST_ZERO}      , //*0200: (Cs -  0)*As + Cs ==> Cs*(As + 1)
+	{ BLEND_ACCU                 , OP_ADD          , SRC1_ALPHA      , CONST_ONE}       , //?0201: (Cs -  0)*As + Cd ==> Cs*As + Cd
+	{ BLEND_NO_BAR               , OP_ADD          , SRC1_ALPHA      , CONST_ZERO}      , // 0202: (Cs -  0)*As +  0 ==> Cs*As
+	{ BLEND_A_MAX                , OP_ADD          , CONST_ONE       , CONST_ZERO}      , //*0210: (Cs -  0)*Ad + Cs ==> Cs*(Ad + 1)
+	{ 0                          , OP_ADD          , DST_ALPHA       , CONST_ONE}       , // 0211: (Cs -  0)*Ad + Cd ==> Cs*Ad + Cd
+	{ 0                          , OP_ADD          , DST_ALPHA       , CONST_ZERO}      , // 0212: (Cs -  0)*Ad +  0 ==> Cs*Ad
+	{ BLEND_NO_BAR | BLEND_A_MAX , OP_ADD          , CONST_ONE       , CONST_ZERO}      , //*0220: (Cs -  0)*F  + Cs ==> Cs*(F + 1)
+	{ BLEND_ACCU                 , OP_ADD          , CONST_COLOR     , CONST_ONE}       , //?0221: (Cs -  0)*F  + Cd ==> Cs*F + Cd
+	{ BLEND_NO_BAR               , OP_ADD          , CONST_COLOR     , CONST_ZERO}      , // 0222: (Cs -  0)*F  +  0 ==> Cs*F
+	{ 0                          , OP_ADD          , INV_SRC1_ALPHA  , SRC1_ALPHA}      , // 1000: (Cd - Cs)*As + Cs ==> Cd*As + Cs*(1 - As)
+	{ BLEND_A_MAX                , OP_REV_SUBTRACT , SRC1_ALPHA      , CONST_ONE}       , //*1001: (Cd - Cs)*As + Cd ==> Cd*(As + 1) - Cs*As
+	{ 0                          , OP_REV_SUBTRACT , SRC1_ALPHA      , SRC1_ALPHA}      , // 1002: (Cd - Cs)*As +  0 ==> Cd*As - Cs*As
+	{ 0                          , OP_ADD          , INV_DST_ALPHA   , DST_ALPHA}       , // 1010: (Cd - Cs)*Ad + Cs ==> Cd*Ad + Cs*(1 - Ad)
+	{ BLEND_A_MAX                , OP_REV_SUBTRACT , DST_ALPHA       , CONST_ONE}       , //*1011: (Cd - Cs)*Ad + Cd ==> Cd*(Ad + 1) - Cs*Ad
+	{ 0                          , OP_REV_SUBTRACT , DST_ALPHA       , DST_ALPHA}       , // 1012: (Cd - Cs)*Ad +  0 ==> Cd*Ad - Cs*Ad
+	{ 0                          , OP_ADD          , INV_CONST_COLOR , CONST_COLOR}     , // 1020: (Cd - Cs)*F  + Cs ==> Cd*F + Cs*(1 - F)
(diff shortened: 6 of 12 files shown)


lightningterror:
There seems to be an issue with the shadows, they appear fully black.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -1066,20 +1092,7 @@ void GSRendererDX11::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sou
...
-
-		switch(m_context->TEST.AFAIL)
-		{
-			case AFAIL_KEEP: z = r = g = b = a = false; break; // none
-			case AFAIL_FB_ONLY: z = false; break; // rgba
-			case AFAIL_ZB_ONLY: r = g = b = a = false; break; // z
-			case AFAIL_RGB_ONLY: z = a = false; break; // rgb
-			default: __assume(0);
-		}
+		bool z, r, g, b, a;
```
lightningterror: The alpha test changes aren't really needed for this pr, it makes reviewing the pr harder, cleanups can be done at a later time.

plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
```diff
@@ -1369,20 +1370,7 @@ void GSRendererOGL::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
...
-
-		switch(m_context->TEST.AFAIL)
-		{
-			case AFAIL_KEEP: z = r = g = b = a = false; break; // none
-			case AFAIL_FB_ONLY: z = false; break; // rgba
-			case AFAIL_ZB_ONLY: r = g = b = a = false; break; // z
-			case AFAIL_RGB_ONLY: z = a = false; break; // rgb
-			default: __assume(0);
-		}
+		bool z, r, g, b, a;
```
lightningterror: Ditto.

plugins/GSdx/Renderers/Common/GSVertexTrace.cpp
```diff
@@ -601,4 +601,19 @@ void GSVertexTrace::CorrectDepthTrace(const void* vertex, int count)
...
+	for (int i = 1; i < count; i++) {
+		if (z !=  v[i].XYZ.Z) {
+			m_eq.z = 0;
+			return;
+		}
+	}
+	m_eq.z = 1;
+
+
+	*/
```
lightningterror: Fixme note/code also isn't needed, there are many improvements that can be made ofc but let's keep the focus of the main goal of the pr.

plugins/GSdx/Renderers/DX11/GSDevice11.h
```diff
@@ -299,10 +308,10 @@ class GSDevice11 final : public GSDevice
 			uint32 key;
 		};
 
-		operator uint32() {return key & 0x1fff;}
+		operator uint32() {return key & 0x3fff;}
 
 		OMBlendSelector() : key(0) {}
-
+
```
lightningterror: Trailing whitespaces.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -939,7 +965,7 @@ void GSRendererDX11::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sou
 		m_om_bsel.wa = false;
 	}
 	else
-	{
+	{
```
lightningterror: Ditto.

plugins/GSdx/Renderers/DX11/GSDevice11.h
```diff
@@ -221,7 +229,7 @@ class GSDevice11 final : public GSDevice
 				uint32 point_sampler:1;
 				uint32 invalid_tex0:1; // Lupin the 3rd
 
-				uint32 _free:27;
```
lightningterror: Just changing it to 25 would be enough.

lightningterror:
I recommend only doing the accumulation blend changes and everything else reverted.
Anything else can be done in a future pr. It's quite a bit messy atm.

hibye8313:
OK, sounds good I'll split the commits into several smaller changes and remove the unnecessary changes like you said. Might be till tomorrow that I can fully revise, have to go to work soon =(.

gregory38:
Take your time. A good and tidy PR is 100 times faster to review than a bad one.

Besides it would be easier to spot bug and discuss on your code

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -801,10 +807,30 @@ void GSRendererDX11::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sou
...
+				{
+					m_ps_sel.accu_blend = 1;
+				}
+				else
+				{
+					m_ps_sel.accu_blend = 2;
+					ps_cb.TA_Af.a = (float)ALPHA.FIX / 128.0f;
+				}
+			}
+
```
lightningterror: In it's current state accumulation blend won't be usable when clamp isn't 0.

gregory38:
By the way the depth test code purpose was to avoid the branch (but as it is never hit probably not important). Early exit seems a good idea but truth is that depth will be equal nearly all times.

It would be nicer to benchmark code because I don't know which code is faster.

gregory38:
Otherwise much better this way

hibye8313:
@gregory38 Hmm, didn't think of that but yeah makes sense the old way if most of the time we have to loop the whole way anyway.

@lightningterror It seems to be working from my end with default DX11 setting and compiling for debug build. Is it showing black in the full shadow volume or just the ground?
Also I thought accu blend was only useful when CLAMP=0 and using HDR, or am I wrong? Otherwise the blend equation can be done fully in HW and since there is no wrapping there is no problem for HW to clamp values right? Or perhaps the reason for using accu blend is also so that HW will not clamp Af/As values? Just to clarify, does HW blending on DX11/OGL clamp the blend factors (source alpha/color, fixed blend factor) to range [0, 1] before the blending or just clamp the final color/alpha when writing to the framebuffer? i.e if we used As = 192/128 = 1.5 will HW use 1.5 when blending or 1.0? I know I should probably test this myself but its easier to ask : )

I revised the commit to only contain the changes needed for the accu blend, hope that is better.

I think it is still important to make the changes to constant buffer so that will be more maintainable in the future. Is it ok to do this as a separate commit on top of the current one or should that be a separate pull request?

Thanks for feedback both.

lightningterror:
Fifa Street and Genji should be a good testcase for accumulation blend.

[Fifa Street fbmask.zip](https://github.com/PCSX2/pcsx2/files/3228085/Fifa.Street.fbmask.zip)
[GENJI [SCPS 15095] (J) Misaligned Bloom - Leaves.zip](https://github.com/PCSX2/pcsx2/files/3228086/GENJI.SCPS.15095.J.Misaligned.Bloom.-.Leaves.zip)

Keep in mind both of them use fbmask, basic for fifa, genji high on d3d.

hibye8313:
Fifa seems to be working correclty on DX11 as compared with SW renderer. However on Genji I am getting problems on both OGL side and DX11. On DX11 the colors are not bright enough in the leaves so probably some alpha multiplier is wrong, on OGL the colors in the leaves are good but the door on the right side it not showing correctly (tried highest blending accuracy). Looking into the issues now.
DX11:
image
OGL:
image
SW:
image

gregory38:
Feel like a missing effect on dx (and a texture cache issue on gl)

gregory38:
Btw, GS uses integer blending. So it doesn't round but trunc. Even with standard clamping it could have an impact. Need to check god of war, I think it uses it at the start of the game but don't remember for what.

hibye8313:
@gregory38 Right, but what I was referring to was on the GS, alpha = 0x80 means multiply by 1.0. But if alpha > 0x80, say 0xC0 then the multiply will be by 1.5 so colors get brighter than normal. Is this true or does the GS clamp alpha to [0, 0x80] before blending? I'm 95% sure on GPU's alpha/color output from the pixel shader are clamped to [0,1] before blending so I'm guessing that could lead to certain blending inaccuracy when HW blending.

gregory38:
Ah yes. GPU will clamp to 1 whereas GS can go up to 255/128. Maybe the reason I did it also for clamping and 32 bits. I don't remember but you need to do it for all cases.

gregory38:
Actually GPU it depends on the format. With float you can have any alpha but also any color. as far I know It isn't possible to clamp only color (which would solve most of our issue).

lightningterror:
The issue with Genji fbmask breaking is it because it also needs sw blending (not sure if accumulation or not) for fbmask to work correctly.

Anyway what I said earlier about these two being good test cases was because they use accumulation blend when clamp is 1. Current dx pr code doesn't allow using accumulation when clamp is 1.

hibye8313:
@gregory38 I see, thank you.

@lightningterror Working on it now, I think I might have broken something. Ok I looked at the OGL code in GSRendererOGL::EmulateBlending() again and you are right accumulation blend is used in cases where COLCLAMP=1, I will fix this.

plugins/GSdx/res/tfx.fx
```diff
@@ -918,6 +922,27 @@ PS_OUTPUT ps_main(PS_INPUT input)
...
+	}
+	else if (PS_ACCU_BLEND == 2) // Cs*F + Cd or Cd - Cs*F
+	{
+		c.rgb = c.rgb * Af;
+	}
+
+	// Format to 16 bit color. This works for accu blend but may not in all cases.
+	// Needed for accurate shadows in Crash Tag Team Racing/Crash of the Titans
+	// FIXME: Maybe better to just enable for all 16 bit cases?
+	if (PS_ACCU_BLEND != 0 && PS_DFMT == FMT_16)
```
lightningterror: I think you need to enable this when you have clamp 0 and accumulation blend only.

hibye8313: Ok that works, it should still fix Tag Team Racing.

lightningterror: Maybe it is not actually needed since we check 16bit format anyway.

hibye8313:
@lightningterror  Just general question to make sure I understand the purpose of accumulation blend without the HDR algo for COLCLAMP=0. I think we want to compute the Value of Cs*Af/Cs*As in the pixel shader to prevent early clamping of Af/As since it is better to have Cs*Af/Cs*As clamped incorrectly than As/Af (which are probably > 1 more often)?

lightningterror:
Gregory can answer that :p

Anyway wouldn't it be better for the accumulation detection and when it is on to do something like this?
accumulation blend when clamp is 1 is only active if blending option is basic or higher.

```
	if (!IsOpaque() && rt)
	{
		m_om_bsel.abe = PRIM->ABE || PRIM->AA1 && m_vt.m_primclass == GS_LINE_CLASS;

		m_om_bsel.a = m_context->ALPHA.A;
		m_om_bsel.b = m_context->ALPHA.B;
		m_om_bsel.c = m_context->ALPHA.C;
		m_om_bsel.d = m_context->ALPHA.D;

		if (m_env.PABE.PABE)
		{
			if (m_om_bsel.a == 0 && m_om_bsel.b == 1 && m_om_bsel.c == 0 && m_om_bsel.d == 1)
			{
				// this works because with PABE alpha blending is on when alpha >= 0x80, but since the pixel shader
				// cannot output anything over 0x80 (== 1.0) blending with 0x80 or turning it off gives the same result

				m_om_bsel.abe = 0;
			}
			else
			{
				//Breath of Fire Dragon Quarter triggers this in battles. Graphics are fine though.
				//ASSERT(0);
			}
		}

		// Accumulation blend
		const GIFRegALPHA& ALPHA = m_context->ALPHA;
		// Cs*As + Cd, Cs*F + Cd, Cd - Cs*As, Cd - Cs*F
		bool accumulation_blend = 
			((ALPHA.A == 0 && ALPHA.B == 2 && ALPHA.C == 0 && ALPHA.D == 1) ||
			(ALPHA.A == 0 && ALPHA.B == 2 && ALPHA.C == 2 && ALPHA.D == 1) ||
			(ALPHA.A == 2 && ALPHA.B == 0 && ALPHA.C == 0 && ALPHA.D == 1) ||
			(ALPHA.A == 2 && ALPHA.B == 0 && ALPHA.C == 2 && ALPHA.D == 1));

		// Color clip
		if (m_env.COLCLAMP.CLAMP == 0)
		{
			if (accumulation_blend)
				m_ps_sel.hdr = 1;

			// fprintf(stderr, "COLCLIP HDR mode ENABLED\n");
			GSVector4 dRect(ComputeBoundingBox(rtscale, rtsize));
			GSVector4 sRect = dRect / GSVector4(rtsize.x, rtsize.y).xyxy();
			hdr_rt = dev->CreateRenderTarget(rtsize.x, rtsize.y, DXGI_FORMAT_R32G32B32A32_FLOAT);
			// Warning: StretchRect must be called before BeginScene otherwise
			// vertices will be overwritten. Trust me you don't want to do that.
			dev->StretchRect(rt, sRect, hdr_rt, dRect, ShaderConvert_COPY, false);
		}

		if (accumulation_blend && (m_ps_sel.hdr || m_sw_blending))
		{

			m_om_bsel.accu_blend = 1;
			if (ALPHA.C == 0)
			{
				m_ps_sel.accu_blend = 1;
			}
			else // C == 2
			{
				m_ps_sel.accu_blend = 2;
				ps_cb.Af.x = (float)ALPHA.FIX / 128.0f;
			}
		}
	}
```

hibye8313:
Yes I think that would be better since it is closer to OGL. For some reason it didn't register for me that DX11 also had the m_sw_blending setting but I see that is inherited from the super class and enabled in the GUI just like OGL. Ok, I'll make the changes and test. Gotta hit the road now so it may be a while...

lightningterror:
So far we have a couple of games confirmed fixed Crash Tag Team Racing, Crash of the Titans, Castlevania, Nano Breaker. I think it also helps Sly, GOW.

hibye8313:
Yay, I'm glad that worked out : ). Do you prefer if I do the constant buffer changes in a new PR or do on top of these? There is also a small bug in the GSreplay() (GS.cpp) function also which I am using for debugging, is it ok to roll these fixes into one?

lightningterror:
I prefer separate because I don't want to delay the pr with the new goodies :), some crc hack purging will follow once this is merged too.

hibye8313:
@gregory38 I see, thanks for that explanation. I ported over the blend map from OGL to DX11, it was a bit painful but its done. I think it might be better to have an API-independent template map in GSDevice.cpp and then use it create maps for both OGL/DX11. As it is the DX11 map is just a copy/paste job of the OGL map. But this might be a minor issue IDK.

gregory38:
Sorry I wasn't very clear. I said that you could use same define for both dx and gl to reduce duplication. Or at least better see the difference between lookup table.

Actually a non-const generic table might be a good idea. So we can fix-up for special case in 1 api.

Do as you wish.

hibye8313:
No, you were clear, I just didn't do what I said I was going to :P. Ok I'm going to just make a generic blend map in GSDevice then make specific for OGL/DX11 so its less of a headache for anyone who needs to modify it down the road.

hibye8313:
Ok so I put the main blending map in GSDevice.cpp and put API-specific copies in GSDeviceOGL.cpp/GSDeviceDX11.cpp. I spent much longer than I thought I would on this, and TBH I don't  know if its really a good change : (. It makes things more complex and I'm not sure how much the benefit is.

lightningterror:
Sorry got kinda lost but why are we removing the hdr check again? IIRC Star Wars BF2 will hit that code path now on clam 1. Here is a dump if it needs testing.

[SWBF2-GS-BadPostProcessing.zip](https://github.com/PCSX2/pcsx2/files/3250008/SWBF2-GS-BadPostProcessing.zip)

hibye8313:
@lightningterror Sorry, I was really sure after some of the discussion with gregory that 16 bits formatting worked correctly in all cases with accu blend...but I could be (and probably am) wrong. I'll debug the dump and see if maybe something else is causing the issue. If not I'll change back.

lightningterror:
Here is a log, I didn't actually compare rendering to see if there are any visual changes, just checked the log to be aware that we may hit the path.

```
T:Oth	ID:3760	S:Cache	=> TC: Lookup Target(Color) 1280x1024, hit: 22 (0x3000, C_16S)
T:Oth	ID:3760	S:Cache	=> TC: Lookup Target(Depth) 1280x1024, hit: 7 (0x0, Z_32)
T:Error	ID:3760	S:Mid	=> FBMASK Unsafe SW emulated fb_mask:80000000 on 16 bits format
T:Error	ID:3760	S:Mid	=> Accumulation Blend on 16bit format with clamp 1
```

hibye8313:
As far I can see visually the difference between DX11 and OGL is in DX11 the background behind the menu is too dark so something is probably wrong with the alpha. But this is happening with/without the HDR check in the shader so it may be unrelated. I think it might just be alpha mask is not being emulated correctly (I think a similar problem was happening with Genji, but it looks correct accidently). BTW do you know how the "channel shuffle effect" works? There is a lot of code in the renderers for emulating this and I don't understand at all what it does and there are no comments. SWBF is using this and for some reason it causes it to skip a lot of draw calls. Not sure it is related to the problem though.

gregory38:
Accu blend is orthogonal of clamp/format.
Accu blend means that we do A*Cs + Cd (or minus). Goal of accu blend is to do the A*Cs on the shader. And the addition/minus on blending unit.
It allows to support A > 1, mask bit, clamp on shader...
Blending unit will keep thing in order, so no barrier or whatever.

hibye8313:
@gregory38 Yes, DX11 will allow the use of ints instead on enums, but I though it was clearer to just use enums so the code would be more self documenting. I think the code could be simpler, the reason I made the copies m_blendMapD3D11/m_blendMapOGL for both DX11/OGL was as an optimization so that we didn't have to convert from GSDevice::SRC_COLOR => GL_SRC_COLOR etc., on every call. But maybe this is useless since it will be need to be called only 3-4 times in a draw call, so really I only saved like 100 CPU cycles... :P.

What I could do is like you said have a function for OGL/DX11 getBlendMapOGL(int index)/getBlendMapDX11(int index) which just converts the enums from GSDevice  on the fly. This would be a lot cleaner (removes the whole "BlendMap" class) and easier to make changes to if needed.

lightningterror:
Can some of the commits be squashed in to one, it is a bit messy atm, you can use `git rebase -i master` in git shell to rebase the branch and squash them in to one, 
1 commit d3d accu blend
2 commit the table changes for gl/dx

plugins/GSdx/Renderers/DX11/GSDevice11.h
```diff
@@ -206,9 +217,10 @@ class GSDevice11 final : public GSDevice
 				uint32 read_ba:1;
 				uint32 fbmask:1;
 
-				// *** Word 2
 				// Blend and Colclip
+				uint32 accu_blend:2;
 				uint32 clr1:1;
+				uint32 hdr:1;
```
lightningterror: If we aren't using the hdr part for the shader then we might as well revert/purge it.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -802,9 +801,18 @@ void GSRendererDX11::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sou
 			}
 		}
 
+		// Accumulation blend
+		const GIFRegALPHA& ALPHA = m_context->ALPHA;
+		size_t blend_index  = ((ALPHA.A * 3 + ALPHA.B) * 3 + ALPHA.C) * 3 + ALPHA.D;
+		int blend_flags = GSDevice11::m_blendMapD3D11[blend_index].flags;
+		bool accumulation_blend = blend_flags & BLEND_ACCU;
```
lightningterror: You should do !!(blend_flags & BLEND_ACCU) like on gl.

gregory38:
You do as you wish. Just some ideas.

Use raw C enum. A single one that mix, opcode, factors... Don't bother with various enum type, GPU API will print error if it is the wrong enum/value.

Add a virtual int func(enum gs) {
const int table[] = { gl_one.....}; // More compact than a map
Return table[gs];
}

Then you can still build the gl/dx table at init.
Note: I'm not 100% sure that both dx and gl table are/must be exactly the same.

plugins/GSdx/Renderers/DX11/GSDevice11.h
hibye8313: The HDR part might still be still be necessary for deciding when to clamp the color in  SW blending since I recently ported over a little bit more of the SW blend from OGL. I actually removed the PS_ACCU_BLEND flag instead in the new commit to be closr to the OGL side. I will leave PS_HDR for now it in for now and see after I have cleaned things up if it can be removed.

lightningterror: Ah yes with the recent changes it is useful now.

hibye8313:
@lightningterror Sounds good, I will squash the commits etc. like you said. Since I'm the only one working on this branch it shouldn't cause any problems.

@gregory38 Ok, that actually sounds like a good idea. It will probably be a lot clearer, putting all the enums into one enums then have a simple table for looking up the OGL/DX11 values in a single function. That way we can also have all the translating code in the same place as where the enums are declared.

The DX11 and GL tables are pretty much identical as far as the src factor, the dest factor, and blend function are concerned. The OGL just had one extra entry at the end for "merge operation", and it had more information in the flags (DX11 did not use the flags at all in its own table).

I ported over a few more of the features of SW blend from OGL to DX11 so that they are now more uniform. This should fix a few post processing effects with sprites like on the Genji/SWBF2 dumps that lightning posted. However, DX11 still lacks barriers/prim-overlap checking like on OGL so it will only work in the non-overlapping sprites case. I'll try and tidy up/squash these commits a bit.

Side note: There is bug in GSDumpGUI for some reason it shows only black in the SWBF2 dump. I tested this on the main branch as well.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -229,11 +229,13 @@ void GSRendererDX11::EmulateTextureShuffleAndFbmask()
 		case ACC_BLEND_MEDIUM_D3D11:
 			// Enable Fbmask emulation excluding triangle class because it is quite slow.
 			// Exclude 0x80000000 because Genji needs sw blending, otherwise it breaks some effects.
+			// (FIXME: It should (hopefully) be safe to remove the check for 0x80000000 now)
```
lightningterror: I never had much love for it when I first added the check :p, you can purge it, Genji seems fine now.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -696,6 +698,116 @@ void GSRendererDX11::ResetStates()
...
+	}
+
+
+	// Accumulation blend
+	uint8 blend_index  = uint8(((ALPHA.A * 3 + ALPHA.B) * 3 + ALPHA.C) * 3 + ALPHA.D);
+	int blend_flag = GSDevice11::m_blendMapD3D11[blend_index].flags;
+
+	bool impossible_or_free_blend =
+		(blend_flag & (BLEND_A_MAX|BLEND_NO_BAR|BLEND_ACCU))
+		|| m_ps_sel.fbmask;
```
lightningterror: I think adding m_bind_rtsample instead of m_ps_sel.fbmask would work too since fbmask calls the bool anyway.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -696,6 +698,116 @@ void GSRendererDX11::ResetStates()
 	m_om_dssel.key = 0;
 }
 
+// Partial port of OGL SW blending. Currently only works for accumulation blend or when m_sw_blending
+// m_sw_blending >= ACC_BLEND_HIGH_D3D11 since it has not been fully tested. There is no check
+// currently for primitive overlap like in OGL but it should give better results for post-processing effects
+// that are done by non-overlapping sprites.
```
lightningterror: Maybe move the comment inside the function instead, seems cleaner maybe.
Also maybe move the blend function below channel shuffle like on gl, kinda got used to search at the upper half.

Well these two are just minor nitpicks.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -696,6 +698,116 @@ void GSRendererDX11::ResetStates()
...
+			// Disable HW blending
+			m_om_bsel.abe = false;
+
+			m_bind_rtsample |= !(blend_flag & BLEND_NO_BAR);
+		}
+
+		// Require the fix alpha vlaue
+		if (ALPHA.C == 2) {
+			ps_cb.Af.x = (float)ALPHA.FIX / 128.0f;
+		}
```
lightningterror: No need for extra brackets for 1 line code.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -696,6 +698,116 @@ void GSRendererDX11::ResetStates()
...
+		m_ps_sel.blend_a = ALPHA.A;
+		m_ps_sel.blend_b = ALPHA.B;
+		m_ps_sel.blend_c = ALPHA.C;
+		m_ps_sel.blend_d = ALPHA.D;
+
+		if (accumulation_blend)
+		{
+			m_om_bsel.accu_blend = 1;
+
+			if (ALPHA.A == 2) {
```
lightningterror: Another nitpick, keep the brackets consistent with the others.

lightningterror:
Some eyecandy

mgs3
pcsx2 2019-06-06 15-58-51-71

Sly 2
pcsx2 2019-06-06 15-59-20-06

hibye8313:
Great, good to see the changes are actually making a difference in the games : ). Yes I will go ahead and make the fixes you recommended. I'm currently still reorganizing the commits so that the blend map changes are cleaner and come before the other changes, so don't be surprised if the history suddenly dramatically changes...the end result should be the same, however.

That bug with GSDumpGUI is really bugging me though, I'll try and fix that one later.

gregory38:
Mgs "fix" is related to impossible blend.

However current behaviour is incorrect if there are overlapping primitives. Blending will always be done with the initial pixel.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
lightningterror: Maybe we don't need to include Unsafe FB mask in impossible/free blend like on gl.

hibye8313:
@gregory38 Yes that's true, this will mainly be useful for those specific games that use post processing effects with non-overlapping sprites. For example Genji and SWBF2 do a Cs*Ad + Cd blend (or something close to it) with sprites which cannot be emulated by HW because of Ad.  I though about porting the prim_overlap() function from OGL to DX11 (actually it could be moved to GSRenderer since its GPU independent) but DX11 doesn't have texture barriers so I'm guessing it would be too slow to do the same thing as OGL when prims do overlap.

Ok so I restructured the commits into 2 commits, the first is for the blend map changes, and the second for the all the SW blending changes.

General question: How do you guys usually debug graphical issues with pcsx2? What I have been doing is compiling GSDX as an .exe and running the GSReplay() function. Then in GSState::VertexKick() I flush the drawing queue every full primitive and dump the RT to a file in the GSRendererHW::Draw() function so I can see each primitive being draw and which is causing the problem. Also that way I can set break points in GSDX code when replaying a dump. This works ok but I have to keep the debugging on a different branch from the branch I'm working on since I'm changing the VS project files, etc.

I see there is already some very thorough dumping functionality in the code for dumping all of the GS registers/RT's/textures/logging/etc. on every frame but I haven't tried this yet. Any suggestions?

plugins/GSdx/Renderers/Common/GSDevice.h
```diff
@@ -116,6 +122,78 @@ class ShadeBoostConstantBuffer
...
+			GL_SRC1_COLOR,     GL_ONE_MINUS_SRC1_COLOR,     GL_SRC_ALPHA,  GL_ONE_MINUS_SRC_ALPHA,
+			GL_DST_ALPHA,      GL_ONE_MINUS_DST_ALPHA,      GL_SRC1_ALPHA, GL_ONE_MINUS_SRC1_ALPHA,
+			GL_CONSTANT_COLOR, GL_ONE_MINUS_CONSTANT_COLOR, GL_ONE,        GL_ZERO,
+
+			GL_FUNC_ADD, GL_FUNC_SUBTRACT, GL_FUNC_REVERSE_SUBTRACT
+		};
+		return map[hw];
+	}
+
+	static uint16 ConvertBlendEnumDX11(uint16 hw)
```
lightningterror: I think you will need to add #ifdef _WIN32 for dx code.

lightningterror:
> General question: How do you guys usually debug graphical issues with pcsx2?

I usually do on gsdumpgui, opengl provides all the log files needed for debugging and you can dump draw calls to compare between renderers.

Anyway as for the pr maybe commit titles and descriptions could be better.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -501,6 +499,114 @@ void GSRendererDX11::EmulateChannelShuffle(GSTexture** rt, const GSTextureCache:
...
+		}
+	}
+
+	// Accumulation blend
+	uint8 blend_index  = uint8(((ALPHA.A * 3 + ALPHA.B) * 3 + ALPHA.C) * 3 + ALPHA.D);
+	int blend_flag = GSDevice11::m_blendMapD3D11[blend_index].flags;
+
+	bool impossible_or_free_blend =
+		(blend_flag & (BLEND_A_MAX|BLEND_NO_BAR|BLEND_ACCU))
+		|| m_bind_rtsample;
```
lightningterror: Maybe we can split m_bind_rtsample in two. m_bind_rtsample_one, and m_bind_rtsample_full.

m_bind_rtsample_one for unsafe fbmask without including it in blending.
m_bind_rtsample_full for sw emulated fbmask will also be checked here.

plugins/GSdx/res/tfx.fx
```diff
@@ -871,6 +881,42 @@ void gs_main(line VS_OUTPUT input[2], inout TriangleStream<VS_OUTPUT> stream)
 
 #endif
 
+
```
lightningterror: Small nitpick, remove 1 extra line, same goes for bottom remove 1.

lightningterror:
Hmm God of War seems to be a bit broken now with blending enabled.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -228,14 +228,12 @@ void GSRendererDX11::EmulateTextureShuffleAndFbmask()
...
-			// Exclude 0x80000000 because Genji needs sw blending, otherwise it breaks some effects.
-			enable_fbmask_emulation = ((m_vt.m_primclass != GS_TRIANGLE_CLASS) && (m_context->FRAME.FBMSK != 0x80000000));
+			enable_fbmask_emulation = (m_vt.m_primclass != GS_TRIANGLE_CLASS);
 			break;
 		case ACC_BLEND_BASIC_D3D11:
 			// Enable Fbmask emulation excluding triangle class because it is quite slow.
-			// Exclude 0x80000000 because Genji needs sw blending, otherwise it breaks some effects.
 			// Also exclude fbmask emulation on texture shuffle just in case, it is probably safe tho.
-			enable_fbmask_emulation = (!m_texture_shuffle && (m_vt.m_primclass != GS_TRIANGLE_CLASS) && (m_context->FRAME.FBMSK != 0x80000000));
+			enable_fbmask_emulation = (!m_texture_shuffle && (m_vt.m_primclass != GS_TRIANGLE_CLASS));
```
lightningterror: Actually I did a bit of testing and it looks like Genji is fixed but only when using High level of blending, didn't test properly before. Basic and Medium will still be broken so it is better to add the check back, comment can be updated tho that we require high level of blending.

plugins/GSdx/Renderers/Common/GSDevice.h
hibye8313: Oops, forgot not everyone is a Windows user =P, fixing...

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
hibye8313: Hmm, I guess so as all the changes for impossible blends like on Genji will only be enabled on High blending better to leave Medium/Basic alone. The FBMask = 0x80000000 for check is quite hacky though. I think it works because twice the alpha is incorrectly stored in the framebuffer (0xff instead of 0x7f I think) but then when its read back for the impossible blend only half its value is used again incorrectly so it cancels out.

gregory38:
A gsdevice virtual function will avoid the need of extra ifdef.

hibye8313:
@lightningterror Thanks for the tip, I will have to look at the OGL logging. I have been doing my own logging but the built in code is probably much more thorough.
Changed the PR title/other misc edits.
For the issue with God Of War could you please post a GS dump? I'll take a look and try to find out whats breaking.

@gregory38 I kept the blend map as a regular array and removed some of the constants since they were no longer needed. Hopefully I understood correctly the changes to make for the blend map. I decided to use a switch statement for conversion so that the order of the generic enum doesn't matter, since now the OGL/DX11 code are in separate files.

Ok the latest commit is mostly changes to the blend map structure. The conversion to OGL/DX11 constants is done on the fly in virtual functions now, but as discussed there shouldn't really be any perf impact at all. The benefit is that the code is cleaner and OGL/DX11 specific code is now in their own files.

Thanks for the reviews both.

plugins/GSdx/Renderers/Common/GSDevice.h
```diff
@@ -118,6 +124,37 @@ class GSDevice : public GSAlignedClass<32>
...
+
+		// HW blend operations
+		ADD, SUBTRACT, REV_SUBTRACT
+	};
+
+	struct HWBlend { uint16 flags, op, src, dst; };
+
+	// Convert the GS blend equations to HW specific blend factors/ops
+	// Index is computed as ((((A * 3 + B) * 3) + C) * 3) + D. A, B, C, D taken from ALPHA register.
+	HWBlend GSDevice::GetBlend(size_t index);
```
tadanokojin: Remove `GSDevice::` you don't need it

lightningterror:
Here is the GOW gs dump.
[God Of War.zip](https://github.com/PCSX2/pcsx2/files/3267350/God.Of.War.zip)

Check draw call 80.
On gl we hit free colclip on no overlap

On dx we hit colclip with sw blending

Here is a raw dump comparison.
image

gregory38:
May I suggest to post blending equation. I think it is "-Cs"

lightningterror:
(0 - Cs) * Af + 0

gregory38:
Ah yes. Do you have Af value ?

lightningterror:
Updated previous comment with full info

gregory38:
Honestly 2nd commit should be split in small part.

Anyway negative value is clamped in shader. Potentially uint isn't a good idea. I don't remember what I did on GL.

lightningterror:
> Changed the PR title/other misc edits.

@hibye8313 Don't forget commits too, you need to do that during rebase.

lightningterror:
> Anyway negative value is clamped in shader. Potentially uint isn't a good idea. I don't remember what I did on GL.

That seems to be it, dump looks good now, on gl we do ivec so it is fine.

plugins/GSdx/res/tfx.fx
```diff
@@ -871,6 +881,40 @@ void gs_main(line VS_OUTPUT input[2], inout TriangleStream<VS_OUTPUT> stream)
...
+		// Standard Clamp
+		if (PS_COLCLIP == 0 && PS_HDR == 0)
+			Cv = clamp(Cv, (float3)0.0f, (float3)255.0f);
+
+		// In 16 bits format, only 5 bits of color are used. It impacts shadows computation of Castlevania
+		// FIXME: Should Alpha be masked also here or not?
+		if (PS_DFMT == FMT_16)
+			Cv = (float3)((uint3)Cv & (uint3)0xF8);
+		else
```
lightningterror: Maybe `else if PS_COLCLIP == 1 && PS_HDR == 0` ? like on gl? I guess Greg will tell us which is better.

hibye8313: I saw those checks in the OGL shader but I don't know if is totally necessary. If doing clamping, it is already done before the masking so masking will not change the result. At worst it will just be doing an unnecessary mask. In the HDR case (not clamping) it will not affect the end result since we will be throwing away the upper bits in the final pass (in 16 bit format this is done anyway even in OGL).

I think it is clearer like this that the HDR algo is not affected by the masking, especially if something were to change with the HDR algo later on. Maybe there is something else I am not understanding here like maybe the perf impact of converting to int => masking => convert back to float. Like you said it's probably best to wait for gregory's input here.

gregory38:
What is the question for me ?

plugins/GSdx/Renderers/Common/GSDevice.h
hibye8313: Thanks, I removed it.

hibye8313:
@gregory38  The question was about the flags checking in the OGL pixel shader at:
```
#if PS_DFMT == FMT_16
    // In 16 bits format, only 5 bits of colors are used. It impacts shadows computation of Castlevania

    Color.rgb = vec3(ivec3(Color.rgb) & ivec3(0xF8));
#elif PS_COLCLIP == 1 && PS_HDR == 0
    Color.rgb = vec3(ivec3(Color.rgb) & ivec3(0xFF));
#endif
```
can this be just changed to:
```
#if PS_DFMT == FMT_16
    // In 16 bits format, only 5 bits of colors are used. It impacts shadows computation of Castlevania

    Color.rgb = vec3(ivec3(Color.rgb) & ivec3(0xF8));
#else
    Color.rgb = vec3(ivec3(Color.rgb) & ivec3(0xFF));
#endif
```
I think end result is exactly the same because clamping (if used) is already done. In some cases for 32 bit format there is just an extra masking done. To me it looks cleaner but maybe the reason was because of the perf impact of doing an unnecessary mask? I guess that is the reason for doing the HDR algo, to avoid the masking in the pixel shader and to do it in the final pass?

--------

Alright, latest restructure is broken into 3 commits for the blend map changes, the selector variables/pixel shader macros for DX11, and the main blending ports in the renderer and pixel shader. Hopefully this is a lot cleaner.

@gregory I incorporated most of the suggestions you made. std::array is now used for the blendMap, access modifiers are a bit better organized and I think are now correct, and I think the virtual/final declarations are correct (had to add an override to GSDeviceNull though).

@lightning going to look at that GOW dump. I honestly never considered a blend like (-Cs)*Af. That is a little bit crazy but I'll see how OGL handles it.

I think I made all of the other misc edits as well, if I missed any please let me know.

plugins/GSdx/res/tfx.fx
```diff
@@ -871,6 +881,40 @@ void gs_main(line VS_OUTPUT input[2], inout TriangleStream<VS_OUTPUT> stream)
...
+			Cv = (float3)((uint3)Cv & (uint3)0xFF);
```
lightningterror: int3 instead of uint3, gow issue.

plugins/GSdx/res/tfx.fx
```diff
@@ -871,6 +881,40 @@ void gs_main(line VS_OUTPUT input[2], inout TriangleStream<VS_OUTPUT> stream)
...
+		Cv = (PS_BLEND_A == PS_BLEND_B) ? D : trunc(((A - B) * C) + D);
+
+		// Standard Clamp
+		if (PS_COLCLIP == 0 && PS_HDR == 0)
+			Cv = clamp(Cv, (float3)0.0f, (float3)255.0f);
+
+		// In 16 bits format, only 5 bits of color are used. It impacts shadows computation of Castlevania
+		// FIXME: Should Alpha be masked also here or not?
+		if (PS_DFMT == FMT_16)
+			Cv = (float3)((uint3)Cv & (uint3)0xF8);
```
lightningterror: Ditto

plugins/GSdx/res/tfx.fx
hibye8313: Glad that was an easy fix, pushing changes now.

gregory38:
On pixel shader every opcode counts. A PS could run millions of time. Nothing is free.
Masking doesn't make sense when you need to clamp. It is either clamping or masking.

lightningterror:
Small nitpick on the commit names, I prefer if they are named starting with GSdx:

Example:
```
GSdx: Made a DX11/OGL-independent blend map in GSDevice.cpp. Made OGL/DX11 specific blend maps at init time.
GSdx-d3d11: Ported SW blending variables from OGL to DX11 PSSelector/BSSelector/pixel shader.
GSdx-d3d11: Partial port of EmulateBlending() from OGL renderer to DX11 renderer and ps_blend() from OGL shader to DX11 shader.
```

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -501,6 +501,114 @@ void GSRendererDX11::EmulateChannelShuffle(GSTexture** rt, const GSTextureCache:
...
+		else
+		{
+			//Breath of Fire Dragon Quarter triggers this in battles. Graphics are fine though.
+			//ASSERT(0);
+		}
+	}
+
+	// Accumulation blend
+	uint8 blend_index  = uint8(((ALPHA.A * 3 + ALPHA.B) * 3 + ALPHA.C) * 3 + ALPHA.D);
+	int blend_flag = m_dev->GetBlendFlags(blend_index);
```
lightningterror: Maybe use dev instead of m_dev here too, for consistency.

tadanokojin: So he can have a useless cast?
`GetBlendFlags` is a GSDevice function there's so reason to call it on `GSDevice11`

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -501,6 +501,114 @@ void GSRendererDX11::EmulateChannelShuffle(GSTexture** rt, const GSTextureCache:
...
+void GSRendererDX11::EmulateBlending()
+{
+	// Partial port of OGL SW blending. Currently only works for accumulation blend or when
+	// m_sw_blending >= ACC_BLEND_HIGH_D3D11 since it has not been fully tested. There is no check
+	// currently for primitive overlap like in OGL but it should give better results for post-processing effects
+	// that are done by non-overlapping sprites.
+	const GIFRegALPHA& ALPHA = m_context->ALPHA;
+	bool sw_blending         = false;
+
+	m_om_bsel.abe = PRIM->ABE || (PRIM->AA1 && m_vt.m_primclass == GS_LINE_CLASS);
```
lightningterror: I think we can do a return like on gl here for an optimization.

```
	if (PRIM->ABE || (PRIM->AA1 && m_vt.m_primclass == GS_LINE_CLASS))
		m_om_bsel.abe = 1;
	else
		return;
```

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -501,6 +501,114 @@ void GSRendererDX11::EmulateChannelShuffle(GSTexture** rt, const GSTextureCache:
...
+			// Disable HW blending
+			m_om_bsel.abe = false;
+
+			m_bind_rtsample |= !(blend_flag & BLEND_NO_BAR);
+		}
+
+		// Require the fix alpha vlaue
+		if (ALPHA.C == 2)
+			ps_cb.Af.x = (float)ALPHA.FIX / 128.0f;
+	}
```
lightningterror: I think we can also move clr blending here like on gl when there is no sw_blending, handle it the same on gl and remove the IsCLR1 bool in blend selector struct.

gregory38:
Sorry buggy app can't reply to comments.

Lt, I strongly advise to make correct code first and.then optimize later

lightningterror:
Sure, we can do the optimizations later.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
hibye8313: @lightningterror  True, it would be more consistent with the OGL EmulateBlending(). However it is not necessary to use dev = (GSDevice11*)m_dev here because the blending setting are handled using the OMBlendSelector struct (m_om_bssel) instead of through GSDevice11 methods.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
hibye8313: Actually, I think this might be a good idea to do now. Aside from the optimization, if blending is not even enabled then we shouldn't be making changes to the pixel shader settings. It might even give wrong results: if ABE = false but A, B, C, D are non zero we might end up doing SW blending in the shader when we shouldn't (IDK for sure because I have not tested and seems as if some games keep ABE = true by default).

gregory38:
My feeling about this PR. Short story

Pr should be split into 2 pr.
1st with accu blend (can keep sw blending infrastructure on shader and such)
2nd with remaining bits of sw blending.

You can keep mostly everything but the part that enable sw_blending in renderer.
Basically impossible blend, alpha check, fbmask check.... shall be removed

Accu blend is really accuracy. Sw blending in this pr is about fixing broken code with broken differently code.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
hibye8313: Actually, even this might be a good change to do now because really its part of the blend map changes and using the blend flags on the DX11 side. It also cleans up the code a bit and might be forgotten later if we leave it.

hibye8313:
@gregory38 Okey doke, this PR was originally supposed to be only for accu blend so I will keep it that way. I structured the PR again into 3 commits as before but on the last commit kept only accu blend in EmulateBlending().
> On pixel shader every opcode counts. A PS could run millions of time. Nothing is free.
> Masking doesn't make sense when you need to clamp. It is either clamping or masking.
I see, makes sense why you made it like that then. I removed the unnecessary mask from DX side as well.
COLCLAMP=0 is now handled only by HDR algo so COLCLIP is not needed in the shader and so is removed.

@lightningterror Fixed the PR titles, added the CLR1 check to EmulateBlending() and early exit when blending is disabled. Also removed the isCLR1() method from OMBSelector struct since it was no longer needed.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -501,6 +501,95 @@ void GSRendererDX11::EmulateChannelShuffle(GSTexture** rt, const GSTextureCache:
...
+	{
+	case ACC_BLEND_HIGH_D3D11:
+	case ACC_BLEND_MEDIUM_D3D11:
+	case ACC_BLEND_BASIC_D3D11:
+		sw_blending |= accumulation_blend;
+	default: break;
+	}
+
+	if (m_env.COLCLAMP.CLAMP == 0)
+		m_ps_sel.hdr = 1;
```
lightningterror: Shouldn't it be 
```
	if (m_env.COLCLAMP.CLAMP == 0)
	{
		if (accumulation_blend)
			sw_blending = true;
		m_ps_sel.hdr = 1;
	}
```

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
lightningterror: You need to remove this in gsdevice11.h since it will be useless with these new changes
```
		bool IsCLR1() const
		{
			return (key & 0x19f) == 0x93; // abe == 1 && a == 1 && b == 2 && d == 1
		}
```

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
hibye8313: Hmm, ok this will probably be better/more clear in case there are changes to SW blending later on. I had left it out because after I removed the other SW blending we actually have sw_blending == accumulation_blend (if basic or higher mode) since accu blend is the only sw blend.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -501,6 +501,99 @@ void GSRendererDX11::EmulateChannelShuffle(GSTexture** rt, const GSTextureCache:
...
+	if (sw_blending)
+	{
+		m_ps_sel.blend_a = ALPHA.A;
+		m_ps_sel.blend_b = ALPHA.B;
+		m_ps_sel.blend_c = ALPHA.C;
+		m_ps_sel.blend_d = ALPHA.D;
+
+		if (accumulation_blend)
+		{
+			m_om_bsel.accu_blend = 1;
```
lightningterror: You will need to add back
```
		if (bsel.accu_blend)
		{
			bd.RenderTarget[0].SrcBlend = D3D11_BLEND_ONE;
			bd.RenderTarget[0].DestBlend = D3D11_BLEND_ONE;
		}
```

In GSTextureFX11

plugins/GSdx/Renderers/DX11/GSDevice11.h
```diff
@@ -309,14 +318,6 @@ class GSDevice11 final : public GSDevice
 		}
```
lightningterror: You can purge IsCLR1() too.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
hibye8313: Ah, sorry too much rewriting of history is starting to get to me. I will fix this.

plugins/GSdx/Renderers/DX11/GSDevice11.h
hibye8313: I think I got it but I did is on the last commit since that was the one where I put in the check for the BLEND_C_CLR flag instead of using the IsCLR1().

hibye8313:
Latest update it mostly just fixing my own mistakes/minor structure changes.

I see now that alpha is already formatted correctly in the D3D11 shader in the alpha correction stage so I removed it from the blend stage (I had added it the last update).

@gregory38 I still don't know exactly what to do about the private/protected/public modifiers. I think it might be the simplest/cleanest to just make m_blendMap a protected (instead of private) member of GSDevice even though it will give access to GSDeviceOGL/GSDevice11.

lightningterror:
You'd still end up with two same modifiers depending on where the code is placed, either two protected or two public because code is read in order. Honestly I'd leave it as it is.

I'm fine with the current changes, some cleanups/improvements can be done here and there but let's get the base pr merged, they can be done afterwards.

hibye8313:
@gregory38 yes, I moved the struct out of the class so now we can have just three access modifiers, private, protected, public. I also changed the #defines for blend flags to an enum since that's how it seems to be handled for other constants in this file like ShaderConvert/ChannelFetch. I changed ONE, ZERO => CONST_ONE, CONST_ZERO, and ADD, SUBTRACT, REV_SUBTRACT => OP_ADD, OP_SUBTRACT, OP_REV_SUBTRACT.

@lightningterror Great, glad to see things are moving forward with this PR. Of course, much thanks/credit to you and gregory for helping to improve it a lot.

lightningterror:
Merged, I did some minor cleanups myself as well, anyway nice work :)
Now we can focus on other stuff, maybe the constant buffer issues can be next.

plugins/GSdx/res/tfx.fx
```diff
@@ -871,6 +880,37 @@ void gs_main(line VS_OUTPUT input[2], inout TriangleStream<VS_OUTPUT> stream)
...
+			Cv = clamp(Cv, (float3)0.0f, (float3)255.0f);
+
+		// In 16 bits format, only 5 bits of color are used. It impacts shadows computation of Castlevania
+		if (PS_DFMT == FMT_16)
+			Cv = (float3)((int3)Cv & (int3)0xF8);
+
+		Color.rgb = Cv / 255.0f;
+	}
+}
+
```
lightningterror: I'm not sure but I believe current implementation is wrong. I think the initial accu blend pr was the correct way because we want to avoid RtSampler on dx because then it will be very slow, especially for blending. Also in it's current state I think it will only work for 24 bit format because we aren't actually setting m_bind_rtsample on sw blending which we should (it will be painfully slow), on gl rt is always bind because we don't need to go around the dx limitations and copy the framebuffer so it is basically free.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -501,6 +501,99 @@ void GSRendererDX11::EmulateChannelShuffle(GSTexture** rt, const GSTextureCache:
...
+		m_ps_sel.hdr = 1;
+	}
+
+	if (sw_blending)
+	{
+		m_ps_sel.blend_a = ALPHA.A;
+		m_ps_sel.blend_b = ALPHA.B;
+		m_ps_sel.blend_c = ALPHA.C;
+		m_ps_sel.blend_d = ALPHA.D;
+
```
lightningterror: See previous comment , probably needs m_bind_rtsample true here but it will be painfully slow atm.

hibye8313:
@lightningterror Thanks for merging the PR! Unfortunately not sure how much time I'm going to have for future issues for at least some time because of job stuff coming up : (, but I'm glad we managed to finish this one.

avih:
> Unfortunately not sure how much time I'm going to have for future issues for at least some time because of job stuff coming up : (, but I'm glad we managed to finish this one.

@hibye8313 Code is rarely "final", and many times new code requires fixes and improvements after the initial merge, etc. As it's currently your code, you're expected to stay around and handle related issues, if such issues come up.

lightningterror:
@hibye8313 We can include BLEND_NO_BAR too, needs to be tested if it's actually useful for games that use it.

ssakash:
@avih 

> As it's currently your code, you're expected to stay around and handle related issues, if such issues come up.

Not really, @hibye8313 is just a contributor making changes to a module in PCSX2. He's under no obligation to stick around and fix issues in case the code ends up being wrong. The responsibility falls on the person who merges the code.

The way you phrase it, it makes it seem like there's some clause bounding him to stick to the project and look into the issues, you could've been a bit verbose to avoid a potential of it being interpreted as an imperative tone.

avih:
No one is under any obligation, issues do come up occasionally with new code, and the best person to handle them on such case is typically whoever wrote the new code (assuming it's not some minor patch).

Please let's not turn this into anything which it isn't.

ssakash:
> No one is under any obligation, issues do come up occasionally with new code, 


I'm personally aware that you didn't mean it that way, just saying there's the possibility of it being interpreted that way, which could've been avoided. 

>the best person to handle them on such case is typically whoever wrote the new code 

Yes, but that doesn't mean the team can pester contributors to keep maintaining the code by having unnecessary expectations on them.

avih:
The only one pestering here is you. Just drop it. If my clarification isn't enough for you then please just go on and on and on. Preferably elsewhere.

lightningterror:
@avih Sometimes the author can no longer maintain said feature for various reasons. In this case responsibility falls to team members to take care of (worst case scenario code can be disabled/reverted) until fixed properly.

What you should be doing here as a fellow team member which has some responsibility is encourage contribution, if someone can't continue maintaining it then that's perfectly fine, maintainers/team members can decide the best course of action so that everyone is happy.

avih:
you two really take it the wrong way. So be it, apologies all around. I'm out of here.

gregory38:
I think you did a god contribution to the code. You likely spent quite some times to understand GS drawing. And then read my awful GL code. Hopefully you can enjoy your favourite games with less bug and that all that matters :)

If you have time and are willing to contribute, you are more than welcome. Don't hesitate to ping me or Lt (whatever for an idea of contribution or specific issue to tackle)

lightningterror:
Regarding Genji fbmask maybe it is better to detect it based on alpha abcd value and disable it if it matches instead of mask value.

PR #2975, opened 2019-05-29
changes required to compile and run using VS 2019

bigianb:
in vs 2019, the system xhash used by unordered_map uses _Target_ in its template definition. pcsx2 #defines this in the r5900 and r3000 headers which leak cause a cryptic compilation error. A quick fix here is just to undef it in gamebase.h. A better fix would be to encapsulate the definition or use something more c++ to avoid the pre-processor.

tap.h did not have string defined. I guess this was included transitively somewhere with an earlier set of system headers.

Finally in biostools, the file is opened in text mode which causes the bios to be partially read. I'm not sure that's vs2019 specific but it's wrong in any case. Maybe the default mode parameter changed at some point. Maybe wxFile would be a better choice here rather than WxFFile because wxFile always open s in binary mode.

### DIFF
--- a/pcsx2/GameDatabase.h
+++ b/pcsx2/GameDatabase.h
@@ -18,6 +18,10 @@
 //#include "Common.h"
 #include "AppConfig.h"
 
+// _Target_ is defined by R300A.h and R5900.h and the definition leaks to here.
+// The problem, at least with Visual Studio 2019 on Windows,
+// is that unordered_map includes xhash which uses _Target_ as a template
+// parameter. Unless we undef it here, the build breaks with a cryptic error message.
 #undef _Target_
 #include <unordered_map>
 #include <wx/wfstream.h>
--- a/pcsx2/GameDatabase.h
+++ b/pcsx2/GameDatabase.h
@@ -18,6 +18,7 @@
 //#include "Common.h"
 #include "AppConfig.h"
 
+#undef _Target_
 #include <unordered_map>
 #include <wx/wfstream.h>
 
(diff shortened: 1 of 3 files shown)


gregory38:
Interesting, I think it is forbidden to use single underscore define as it is reserved for implementation.

I think it would be nice to put a comment for the undef to understand the purpose of it.

bigianb:
> Interesting, I think it is forbidden to use single underscore define as it is reserved for implementation.
> 
> I think it would be nice to put a comment for the undef to understand the purpose of it.

agreed. The correct fix is to rename the _Target_ define or to write it in a meta programming style rather than via the pre-processor. This at least provides a band aid though.
I added the comment.

arcum42:
Yeah, and if we renamed the _Target_ define, we'd want to rename a bunch of other ones for consistency (perhaps putting a prefix before them).. Rewriting would ultimately be better, but the worry there is always introducing new issues along the way, and even if that was done, we really should do either the undef or renaming the defines first, to get things working in the meantime

lightningterror:
There are a couple of instances where _Target_ is used so not that many (github search).

arcum42:
It's mainly that, for example, if you rename it here\https://github.com/PCSX2/pcsx2/blob/master/pcsx2/R5900.h#L240

You probably want to rename all the defines above and below it (say, adding a prefix to all of them) for consistency. Doable, but a bit of a pain.

lightningterror:
> Doable, but a bit of a pain.

I'll settle with the current undef workaround for now.

arcum42:
Yeah, I'd rather we just go with the unddef for now. We can rename them in a code cleanup later, and further still possibly rewrite. It is the sort of area that should probably be updated to something more modern...

lightningterror:
So everyone fine with the changes so I can merge it?

arcum42:
Haven't actually had a chance to try building it myself, but it builds in Travis and AppVeyor. I'm fine with it.

PR #2986, opened 2019-06-09
[GSdx] Added temperature setting

Dogway:
Port from libretro's white_point shader.

PR #2987, opened 2019-06-09
[GSdx.fx] Added temperature setting

Dogway:
Port from libretro's white_point shader.


lightningterror:
Can the work be done in 1 pr instead of opening multiple ones for the same feature?
Duplicate of #2986

PR #2989, opened 2019-06-09
[GSdx] Add temperature setting

Dogway:
Ported libretro's white_point shader to GSdx.
Based on the work described [here](http://www.zombieprototypes.com/?p=210). 

And sorry for the pull request mess.


lightningterror:
Was this tested if it works properly?

Dogway:
Looks good to me
https://pasteboard.co/IiEmkfp.png
https://pasteboard.co/IiEmq4J.png

https://pasteboard.co/IiP9YF0.png
https://pasteboard.co/IiPa4gG.png

https://pasteboard.co/IiPa9LL.png
https://pasteboard.co/IiPafkG.png

KrossX:
I don't have a CRT to compare, even then it would have to be properly calibrated to test, but low values seem warmer and higher values seem cooler as expected. Also, maybe the effect can be done with the other shaders? But converting temperature to presets would be quite the problem.

On my very old Radeon card, the lines 1951 and 1965 give me an error. `float3 wp = 255.0;` and `wp.rgb = clamp(wp.rgb, 0.0, 255.0) / 255.0;` worked. Using a value of 1000 gives the 1954/1958 `indefinite logarithm` warning. Those paths shouldn't be used so it ain't a problem. I guess it just shows the shader wasn't compiled in an optimized way and the constant is actually being checked during runtime and boths paths are being run.

Was there any log message on the console when you run this?

Dogway:
The default value of 6500K should be a no-op. I found that values between 8000-9000 are good for matching CRT temperature.
In my case using a recent card, a GTX 1070 the 1951 line triplet has to be explicitly defined or the output is black.
I didn't see any error on the log, but modify the code as you find necessary for old card support or code cosmetics.

gregory38:
Imho, it would be better to check the full value in if condition rather than the divided by 100 variable. It would make the job easier for the compiler.

Dogway:
Included a range check for the input variable and raised minimum to 2000K for good measure.
Hopefully all is fine now on the compiler.

lightningterror:
Is pr ready for merge?

Dogway:
All fine on my side

lightningterror:
KrossX reported that he's still getting some warnings when using low values, it will be nice do address those.

Dogway:
Added an if else condition, if it fails report where it does

lightningterror:
I gave it a quick test and lower values seem to break dx11 in my case.
```
D:\Program Files\PCSX2\shaders\GSdx.fx(2154,113-128): warning X4119: indefinite logarithm
D:\Program Files\PCSX2\shaders\GSdx.fx(2158,97-112): warning X4119: indefinite logarithm
```

Dogway:
Not sure why those are even evaluated when (temp <= 66.0), but in any case now it should be fine.

lightningterror:
I think it's all good now, both gl and dx work with high and low values.

lightningterror:
Also a quick note it seems that it doesn't work on llvmpipe mesa. I don't know if that's an expected behavior.

KrossX:
This code seems to work, no warnings. I think it should output the same image but it should be tested. No clamping because... "try values outside suggested range at your own risk" =P

```hlsl
#if COLOR_TEMPERATURE == 1
float4 TemperaturePass(float4 color, float2 texcoord)
{
   float temp = White_Point / 100.0f;

   // all calculations assume a scale of 255. We'll normalize this at the end
   float3 wp = float3(255.0,255.0,255.0);
   
   if(White_Point <= 6600.0) {
	   wp.r = 255.0;
	   wp.g = - 155.25485562709179 - 0.44596950469579133 * (temp - 2.0)  + 104.49216199393888 * log(temp - 2.0);
	   wp.b = White_Point <= 1900 ? 0.0 : - 254.76935184120902 + 0.8274096064007395 * (temp - 10.0) + 115.67994401066147 * log(temp - 10.0) ;
   } else {
	   wp.r = 351.97690566805693 + 0.114206453784165 * (temp - 55.0) - 40.25366309332127 * log(temp - 55.0);
	   wp.g = 325.4494125711974  + 0.07943456536662342 * (temp - 50.0) - 28.0852963507957   * log(temp - 50.0);
	   wp.b = 255.0;
   }
   
   // clamp and normalize
   wp.rgb = clamp(wp.rgb, 0.0, 255.0) / 255.0;

   float3 adjusted = color.rgb * wp;
   float3 base_luma = XYZtoYxy(RGBtoXYZ(color.rgb));
   float3 adjusted_luma = XYZtoYxy(RGBtoXYZ(adjusted));
   adjusted = adjusted_luma + (float3(base_luma.r,0.0,0.0) - float3(adjusted_luma.r,0.0,0.0));
   color = float4(XYZtoRGB(YxytoXYZ(adjusted)), 1.0);

   return color;
}
#endif
```

Dogway:
Code wise it reads better. I thought it could look different at 6600 since that should be 255.0 not the polynomial but they match, after all it's the inflection of the curve. I had to re-apply the clamp to the indicated range for sanity and because it still spits "indefinite logarithm" errors due to potential log(-n) (ie. 900K).

lightningterror:
The issue with mesa not working was an unsupported extension so it's not our issue, anyway tested again and everything seems to work.

PR #2993, opened 2019-06-14
Gsdx-hw/gui: Purge Sprite hack.

### LINKED ISSUE #2615: UserHacks_SpriteHack review request
This hack was made in order to reduce some odd black edge on sprite parts in games like ArTonelico, Mana Khemia and Tales of Destiny DC. This problem appears to have been fixed already so unless there are other games that benefit from this particular hack, it might be time for it to be removed.

Official 1.4.0 release still shows the issue but the oldest automatic build from 2016-11-12 does not. I haven't checked what actually fixed it yet, but it was something in 2016 for sure. (EDIT) Fixed by PR #1538.

[Old forum thread about the issue](https://forums.pcsx2.net/Thread-Some-kinda-texture-bug-Ar-Tonelico-2-Mana-Khemia-2)
[Screenshot comparison gallery](https://imgur.com/a/TM30h9f)

lightningterror:
The hack is no longer useful and is replaced by ATST
which works much better.

Allow to toggle the special Nvidia hack with "Disable Safe Features"
option instead.

Bonus Direct3D11: Update shader macros handling.

Close #2615


plugins/GSdx/Renderers/DX11/GSDevice11.cpp
```diff
@@ -243,9 +243,9 @@ bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
 			dxd->Release();
 		}
 
-		bool spritehack_enabled = theApp.GetConfigB("UserHacks") && theApp.GetConfigI("UserHacks_SpriteHack");
+		bool nvidiahack_enabled = theApp.GetConfigB("UserHacks") && theApp.GetConfigB("UserHacks_Disable_Safe_Features");
 
-		m_hack_topleft_offset = (!nvidia_gpu || m_upscale_multiplier == 1 || spritehack_enabled) ? 0.0f : -0.01f;
+		m_hack_topleft_offset = (!nvidia_gpu || m_upscale_multiplier == 1 || nvidiahack_enabled) ? 0.0f : -0.01f;
```
tadanokojin: It's probably better to move this code up (or at least NV detection part) so you don't have to query the api for the adapter description twice.

plugins/GSdx/Renderers/DX11/GSDevice11.cpp
```diff
@@ -243,9 +243,9 @@ bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
...
 	}
 
 	D3D11_FEATURE_DATA_D3D10_X_HARDWARE_OPTIONS options;
```
tadanokojin: You can probably remove this too while you're here. It's unused as far as I can tell.

plugins/GSdx/Renderers/DX11/GSTextureFX11.cpp
```diff
@@ -166,19 +166,22 @@ void GSDevice11::SetupGS(GSSelector sel, const GSConstantBuffer* cb)
...
 			std::string str[4];
 
-			str[0] = format("%d", sel.iip);
-			str[1] = format("%d", sel.prim);
-			str[2] = format("%d", sel.point);
-			str[3] = format("%d", sel.line);
+			str[id++] = format("%d", sel.iip);
+			str[id++] = format("%d", sel.prim);
+			str[id++] = format("%d", sel.point);
+			str[id++] = format("%d", sel.line);
```
tadanokojin: Honestly I think it's better the old way as this makes me do math in my head in order to read.

KrossX: But the previous way forced me to count and I'm bad at it, so it was error prone.

pgert:
Fatal Frame 1 (Project Zero 1) made use of the Sprite hack.
Setting it to Half fixed the upscaling problems with the photos.
So what is this ATST?

PR #2998, opened 2019-06-19
GSdx-hw: Sonic Unleashed OI hack

lightningterror:
- Purge existing crc hack for Sonic Unleashed
- Add support for rgba channel selection to StretchRect d3d11/gl
- Add an OI hack for Sonic Unleashed that will allow us to replace draws by a copy, it will allow to render the shadows properly now on Sonic. Requires crc level above minimum for it to work.

Collaborators/credits to Gregory and Kojin for helping with the code.


gregory38:
I think we are good now !

PR #3003, opened 2019-06-29
GSdx-d3d11: Enable Blend no barrier support on d3d11

lightningterror:
gsdx-d3d11: Extend sw blending to BLEND_NO_BAR which is also free and doesn't require rt sample nor any barriers.
It helps light rendering in SW Battlefront 2, VP2 also triggers it but needs to be further tested what it does.

gsdx-d3d11: Extend BLEND_NO_BAR to also hit on clamp 0, it's faster than standard HDR algo.

GOW shows a nice fps improvement. Spyro too maybe.

### DIFF
--- a/plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
+++ b/plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
@@ -547,10 +547,24 @@ void GSRendererDX11::EmulateBlending()
 
 	if (m_env.COLCLAMP.CLAMP == 0)
 	{
-		// fprintf(stderr, "%d: COLCLIP HDR mode%s\n", s_n, accumulation_blend ? " with accumulation blend" : "");
 		if (accumulation_blend)
+		{
+			// fprintf(stderr, "%d: COLCLIP HDR mode with accumulation blend\n", s_n);
 			sw_blending = true;
-		m_ps_sel.hdr = 1;
+			m_ps_sel.hdr = 1;
+		}
+		else if (sw_blending)
+		{
+			// So far only BLEND_NO_BAR should hit this path, it's faster than standard HDR algo.
+			// Note: Isolate the code to BLEND_NO_BAR if other blending conditions are added.
+			// fprintf(stderr, "%d: COLCLIP SW ENABLED (blending is %d/%d/%d/%d)\n", s_n, ALPHA.A, ALPHA.B, ALPHA.C, ALPHA.D);
+			m_ps_sel.colclip = 1;
+		}
+		else
+		{
+			// fprintf(stderr, "%d: COLCLIP HDR mode\n", s_n);
+			m_ps_sel.hdr = 1;
+		}
 	}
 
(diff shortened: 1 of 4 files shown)


plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -529,26 +529,47 @@ void GSRendererDX11::EmulateBlending()
...
+		{
+			// fprintf(stderr, "%d: COLCLIP HDR mode\n", s_n);
+			m_ps_sel.hdr = 1;
+		}
 	}
 
+#if 0
+	fprintf(stderr, "%d: BLEND_INFO: %d/%d/%d/%d. Clamp:%d. Prim:%d number %d (sw %d)\n",
+		s_n, ALPHA.A, ALPHA.B, ALPHA.C, ALPHA.D, m_env.COLCLAMP.CLAMP, m_vt.m_primclass, m_vertex.next, sw_blending);
+#endif
```
tadanokojin: Is this ifdef needed?

tadanokojin:
Honestly I'm concerned about this PR for 1.6.0 because it appears to just be an optimization and doesn't appear to fix anything.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
ssakash: Seems like this is for debugging, I personally don't mind if it'll be useful for some quick testing or anything. I've few of such stuff for PCRTC operations but under a different commented out macro in ``config.h``

tadanokojin: yeah but maybe we can just comment it?

ssakash: Sure, I don't mind either. The ``#ifdef macro`` case might be better in case he has a bunch of different debug statements to be enabled just with a single edit, whereas at such cases it might be a PITA to uncomment all other different statements. (but that doesn't seem to be the case here)

ssakash:
> GOW shows a nice fps improvement. Spyro too maybe.

Some statistics will be nice.

lightningterror:
> Some statistics will be nice.

I posted some results last week.

> hdr colclip 
> 169fps native, 61fps 3x
> 
> sw blend no barrier sw colclip
> 175fps native, 67fps 3x

And here are the visual changes on SW Battlefront 2
Master:
https://cdn.discordapp.com/attachments/514592552389967881/589811619618357259/pcsx2_2019-06-16_15-33-51-89.png

pr:
https://cdn.discordapp.com/attachments/514592552389967881/589811665676140554/pcsx2_2019-06-16_15-39-56-49.png

sw/gl:
https://cdn.discordapp.com/attachments/514592552389967881/589811746152251392/pcsx2_2019-06-16_15-37-30-70.png

hw gl excluding sw blend with no prim overlap check:
https://cdn.discordapp.com/attachments/514592552389967881/589952124998123562/pcsx2_2019-06-16_22-22-20-25.png

PR #3004, opened 2019-06-30
Readme: Drop official support of Vista

tadanokojin:
While it should be possible to run PCSX2 on Vista, it belongs in an OS graveyard as MS dropped support of it along with XP. So let's drop official mention of it and not continue to worry if future commits will break it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,15 +16,15 @@ Development builds are also available from [our website](https://pcsx2.net/downl
 # System Requirements
 
 ## Minimum
-* OS: Windows Vista SP2 or newer or GNU/Linux (32-bit or 64-bit)
-* CPU: Any that supports SSE2 (Pentium 4 and up, Athlon64 and up) @ 1600 STR or better
-* GPU: DirectX 10 GPU or better
-* RAM: 2GB or more
+* OS: Windows 7 or GNU/Linux
+* CPU: Any that supports SSE2 @ [1600 STR](#Notes)
+* GPU: DirectX 10 support
+* RAM: 2GB
 
 ## Recommended
-* OS: Windows 7/8/8.1/10 (64-bit) or GNU/Linux (64-bit)
-* CPU: Intel Haswell (or AMD equivalent) @ 2000 STR or better
-* GPU: DirectX 11 GPU or greater
+* OS: Windows 10 (64-bit) or GNU/Linux (64-bit)
+* CPU: Any that supports AVX2 (Core series Haswell or Ryzen and newer) @ [2000 STR](#Notes) or better
+* GPU: DirectX 11 support or better
 * RAM: 4GB or more
 
 ## Notes


refractionpcsx2:
Or just do as we did with XP and continue as normal until we do come across something that kills it, then drop support.

tadanokojin:
I understand how you mean we did a similar with with dx9
However, rather than surprise people by breaking it one day I'd like to remove it from our list of officially supported OSes now rather than then.

How about this as a compromise?
```
Windows 7 SP1 or newer (Vista SP2 is known to work but not officially supported)
```

lightningterror:
> How about this as a compromise?

It makes it a little bit more complicated, I prefer current changes, also wasn't dx9 removed from readme long before dx9 was removed?

lightningterror:
While you're at it you can change cpu recommended requirements, AVX2 support Haswell/ Ryzen or better.

ssakash:
I personally agree with @refractionpcsx2, there's no real need to discourage the rare percentile of Vista users from trying PCSX2 when there are no known issues with it so far.

gregory38:
Seriously kill it. Win7 is nearly dead too. People are free to run the program even on minimal conf. But the fact is nobody tested if vista is still working. And the cost to test it is too much for the 10 users.

PSI-Rockin:
We have git for a reason. People who are stuck on older systems are free to use older builds, and PCSX2 should not be afraid of dropping support for <1% of users. This insistence on supporting a small minority of users whose OS is not supported by the creator has hurt not only PCSX2, but also numerous other projects in the past. You got rid of DX9 for a reason, after all.

atomic83GitHub:
Honesty, Vista is deprecated,slow and dangerous to use with meltdown and spectre. PCSX2 share the same system requirement as Windows 10, that mean SSE2 as the bare minimum (Athlon 64 and Pentium 4), so it makes no sense not to upgrade for it.

ssakash:
Seems like the consensus is on removing it. You're cool with it @refractionpcsx2?

@tadanokojin - Doesn't it feel redundant to have Windows 7 mentioned again in recommended when it is also in minimum? (Like are there even any users who stay at SP1 when updates are forced?)

gregory38:
Recommended should be win10.

refractionpcsx2:
@ssakash I guess I've been outvoted. I still don't see the point, I mean saying Win7 is the minimum is a blatant lie as it will work on Vista.  We're supporting the program not the OS, so I don't see why it should be a problem while Vista does still actually work, we're just not taking any active steps to preserve that functionality.

But since everybody else wants to change it... go ahead.

gregory38:
You are pedantic. If people can't upgrade because of crappy laptop, they won't meet the other requirements anyway. If it is a money issue. Don't use windows in the first place. And there are legal solution for cheap if not free windows 10. We have already some horrors in the build system to handle win7. So yes it may or not run. But what the point ?

refractionpcsx2:
My point is what is the point in saying Win 7 is the minimum requirements when it blatantly isn't? This isn't anything to do with people not being rich enough to upgrade or whatever, it's just a plain fact that the emulator DOES work with Vista, so saying the minimum requirement is Windows 7 is an outright lie. What do we gain from saying Vista isn't the minimum? Absolutely nothing. I've never come across an issue (personally) where something didn't work because their OS was too old and we don't have to fix bugs which happen only in Vista, we can tell the user to upgrade if this does happen.

tadanokojin:
I updated with some suggestions though I feel a little strange about the language for recommended cpu being Haswell and up considering modern Pentiums don't support AVX2. If anyone has a better way to word it please let me know.

refractionpcsx2:
@tadanokojin I guess all you can say "haswell and up with AVX2 support"
or maybe "Any Intel Core processor from 4000 series and up", since that kind of excludes pentiums

lightningterror:
Core i3 or better works too.

atomic83GitHub:
Why not mention AMD apus (ryzen 3400g) in the read-me for people that want a cheap but powerfull upgrade?

refractionpcsx2:
Because it's not out yet atom? xD

atomic83GitHub:
In a few days :) the 7

README.md
```diff
@@ -16,14 +16,14 @@ Development builds are also available from [our website](https://pcsx2.net/downl
 # System Requirements
 
 ## Minimum
-* OS: Windows Vista SP2 or newer or GNU/Linux (32-bit or 64-bit)
+* OS: Windows 7 SP1 or newer or GNU/Linux (32-bit or 64-bit)
 * CPU: Any that supports SSE2 (Pentium 4 and up, Athlon64 and up) @ 1600 STR or better
```
lightningterror: Maybe we can remove the Pentium/Athlon part.

README.md
```diff
@@ -16,14 +16,14 @@ Development builds are also available from [our website](https://pcsx2.net/downl
...
 * GPU: DirectX 10 GPU or better
 * RAM: 2GB or more
 
 ## Recommended
-* OS: Windows 7/8/8.1/10 (64-bit) or GNU/Linux (64-bit)
-* CPU: Intel Haswell (or AMD equivalent) @ 2000 STR or better
+* OS: Windows 10 (64-bit) or GNU/Linux (64-bit)
+* CPU: Any that supports AVX2 (Haswell and up, Ryzen and up) @ 2000 STR or better
```
ssakash: Honestly not all end users might be familiar of CPU architectures, we need something more simple for users to catch I feel. Also might be nice to hyperlink to hyper link the STR value to the STR description sectio.

README.md
atomic83GitHub: Core 2 duo and Athlon 64 X2 would better fit for minimum (even if those single core works).

README.md
atomic83GitHub: Something like Core I7 4770K and Ryzen 5 2600?

lightningterror: Just the STR is fine, no need to mention models unless you absolutely have to (ex avx2 support).

gregory38:
Ref I understand you. But honestly it should be read as minimal *supported* requirements.

So either we add supported. Or we add "older OS may or may not work. No support will be provided"

Otherwise you have crazy situation where Kojin installed a dead OS to check/fix codes. So Vista is never broken because people fixes it based on the readme.

gregory38:
Otherwise we can break vista. And at least use dx11 feature. For example to implement accurate date on dx.

tadanokojin:
How about something like this. I think we can drop the 64/32 bit mention in minimal OS as I think it's already implied by not specifying. Additionally drop the "or better" language from from minimums as that is already implied. I can add links to the STR values as @ssakash suggested.

Thoughts?

## Minimum
* OS: Windows 7 SP1 or GNU/Linux
* CPU: Any that supports SSE2 (Pentium 4 or Athlon64) @ 1600 STR
* GPU: DirectX 10 GPU
* RAM: 2GB 

## Recommended
* OS: Windows 10 (64-bit) or GNU/Linux (64-bit)
* CPU: Any that supports AVX2 (Core series Haswell or Ryzen and newer) @ 2000 STR or better
* GPU: DirectX 11 GPU or better
* RAM: 4GB or more

lightningterror:
I still think `CPU: Any that supports SSE2 @ 1600 STR` would be better.

tadanokojin:
Updated with suggestions

atomic83GitHub:
Sorry but there are Haswell CPUs which don't support AVX2, the celeron ones.

lightningterror:
> Sorry but there are Haswell CPUs which don't support AVX2, the celeron ones.


`CPU: Any that supports AVX2 (Core series Haswell or Ryzen and newer) @ [2000 STR](#Notes) or better`

PR #3007, opened 2019-07-02
GSdx: Add CRC hack for Sniper Elite

tadanokojin:
Game has ghosting even on native due to an unimplemented CS effect.
Temporary fix until it's properly implemented.

### DIFF
--- a/plugins/GSdx/GSCrc.cpp
+++ b/plugins/GSdx/GSCrc.cpp
@@ -165,6 +165,7 @@ CRC::Game CRC::m_games[] =
 	{0x6BA2F6B9, ResidentEvil4, EU, 0},
 	{0x60FA8C69, ResidentEvil4, JP, 0},
 	{0x5F254B7C, ResidentEvil4, KO, 0},
+	{0x395779C5, SniperElite, US, 0},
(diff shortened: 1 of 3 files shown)


turtleli:
What's the status on this?

tadanokojin:
Would be nice to avoid it.
Comment about unimplemented CS is incorrect. Actually this issue is SW too but SW enables autoflush by default. So autoflush on HW will avoid this issue.

Haven't had a moment to debug it further or get Greg's thoughts yet so I just left this open for now.

lightningterror:
> Comment about unimplemented CS is incorrect. Actually this issue is SW too but SW enables autoflush by default. So autoflush on HW will avoid this issue.

If autoflush helps improve/solve the issue then we should stick with that, not in favor in adding more hacks like these otherwise. That's just my opinion.

tadanokojin:
This one kinda sucks because there's really no acceptable workaround while scaling.

tadanokojin:
On second thought, I'm going to close this out as it doesn't skip the whole effect anyway and I don't feel like debugging it again to get this hack right. For now native res + auto flush is just going to have to suffice.

PR #3010, opened 2019-07-07
Makes VU0f registers display as floats in the debugger

monster860:
image


tadanokojin:
Isn't this already open at #2779?

ramapcsx2:
Yep, looks like 2 implementations of the same thing.
I'm not actively using the debugger, and I don't have much external input on it either.
If you guys really want this "display as float" feature, I suppose we should do it (and merge one of these) :p

lightningterror:
Which one is better?

ramapcsx2:
https://github.com/PCSX2/pcsx2/pull/2779 has the nicer formatting.
This PR currently shows a higher precision but I don't think this more useful than the easier readability of the other.
So yeah, 2779 gets my vote :p

lightningterror:
Closing in favor of #2779

PR #3011, opened 2019-07-09
Implements IOP breakpoints

### LINKED ISSUE #1746: R3000 breakpoints not working
***PCSX2 version:***
PCSX2 1.4.0-20160105132032- compiled on Jan  5 2016

***PCSX2 options:***
Default.

I also tested changing these settings without any luck:
- IOP -> Interpreter
- EmotionEngine -> Interpreter
- Enable speedhacks -> disabled
- Automatic gamefixes -> disabled

***Plugins used:***
Default.

***BIOS used:***
**Japan   v01.60(27/04/2001)  Console** (SCPH-30000 version 5.0 04/27/01 J)

***Games tested:***
- Final Fantasy X (SLPS-25050)
- Rakugaki Okoku 2 Maojo no Tatakai (SLPM-65637)
- World Soccer Winning Eleven 5 (SLPM-62053)

Breakpoints on the R3000 core doesn't appear to be working. Both cores give me the message "unimplemented" if I explicitly right-click on an opcode and press "toggle breakpoint", but evidently breakpoints *are* implemented at least for R5900, as they do work fine there.

I am positive that the addresses I put breakpoints on are reached. I have even been able to put in a printf call via the "assemble opcode" context menu function to verify it via the PCSX2 log.

I have tried the latest development version as well (v1.5.0-dev-1764-g80be9d9) and it seems the problem is still present there.

Select and click an opcode to set a breakpoint. Alternatively, add it in the breakpoints tab, or click the "breakpoint" button at the top.

?

***PC specifications:***
CPU: Intel Core i7-6800K
GPU: ASUS GeForce GTX970
OS: Windows 7 SP1 64 bit

monster860:
Breakpoints on the IOP will now actually break, and you can now step through it and shizz. I've only implemented this on recompiler, and not the interpreter.

Fixes #1746

### DIFF
--- a/pcsx2/DebugTools/DebugInterface.h
+++ b/pcsx2/DebugTools/DebugInterface.h
@@ -19,6 +19,12 @@
 
 enum { EECAT_GPR, EECAT_CP0, EECAT_FPR, EECAT_FCR, EECAT_VU0F, EECAT_VU0I, EECAT_COUNT };
 enum { IOPCAT_GPR, IOPCAT_COUNT };
+enum BreakPointCpu
+{
+	BREAKPOINT_EE = 0x01,
+	BREAKPOINT_IOP = 0x02,
+	BREAKPOINT_IOP_AND_EE = 0x03
+};
 
 class DebugInterface
 {
@@ -51,6 +57,7 @@ public:
 	virtual std::string disasm(u32 address, bool simplify) = 0;
 	virtual bool isValidAddress(u32 address) = 0;
 	virtual u32 getCycles() = 0;
+	virtual BreakPointCpu getCpuType() = 0;
 	
 	bool initExpression(const char* exp, PostfixExpression& dest);
 	bool parseExpression(PostfixExpression& exp, u64& dest);
@@ -90,6 +97,7 @@ public:
 	virtual std::string disasm(u32 address, bool simplify);
 	virtual bool isValidAddress(u32 address);
 	virtual u32 getCycles();
+	virtual BreakPointCpu getCpuType();
 };
 
 
@@ -122,6 +130,7 @@ public:
 	virtual std::string disasm(u32 address, bool simplify);
 	virtual bool isValidAddress(u32 address);
 	virtual u32 getCycles();
+	virtual BreakPointCpu getCpuType();
 };
 
 extern R5900DebugInterface r5900Debug;
--- a/pcsx2/x86/iR3000A.cpp
+++ b/pcsx2/x86/iR3000A.cpp
@@ -1055,7 +1055,7 @@ void psxDynarecCheckBreakpoint()
 		return;
 
 	CBreakPoints::SetBreakpointTriggered(true);
-	GetCoreThread().PauseSelf();
+	GetCoreThread().PauseSelfDebug();
 	iopBreakpoint = true;
 }
 
@@ -1066,7 +1066,7 @@ void psxDynarecMemcheck()
 		return;
 
 	CBreakPoints::SetBreakpointTriggered(true);
-	GetCoreThread().PauseSelf();
+	GetCoreThread().PauseSelfDebug();
 	iopBreakpoint = true;
 }
 
--- a/pcsx2/x86/iR3000A.cpp
+++ b/pcsx2/x86/iR3000A.cpp
@@ -23,6 +23,8 @@
 #include "iR3000A.h"
 #include "BaseblockEx.h"
 #include "System/RecTypes.h"
+#include "System/SysThreads.h"
+#include "R5900OpcodeTables.h"
 
 #include <time.h>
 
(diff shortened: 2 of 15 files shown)


pcsx2/DebugTools/DebugInterface.h
```diff
@@ -19,6 +19,12 @@
 
 enum { EECAT_GPR, EECAT_CP0, EECAT_FPR, EECAT_FCR, EECAT_VU0F, EECAT_VU0I, EECAT_COUNT };
 enum { IOPCAT_GPR, IOPCAT_COUNT };
+enum BreakPointCpu
+{
+	BREAKPOINT_EE = 0x01,
+	BREAKPOINT_IOP = 0x02,
+	BREAKPOINT_IOP_AND_EE = 0x03
```
ramapcsx2: BREAKPOINT_IOP_AND_EE? 
I wonder how that would look like?

pcsx2/x86/iR3000A.cpp
```diff
@@ -1021,11 +1029,167 @@ void rpsxBREAK()
...
+
 void psxRecompileNextInstruction(int delayslot)
 {
 	// pblock isn't used elsewhere in this function.
 	//BASEBLOCK* pblock = PSX_GETBLOCK(psxpc);
 
+	// add breakpoint
+	if (!delayslot)
+	{
+		psxEncodeBreakpoint();
```
ramapcsx2:
```
psxEncodeBreakpoint();
psxEncodeMemcheck();
```
Regarding run-time performance:
Are these emitting any extra code while the debugger isn't in use?
Can they be disabled / bypassed for release builds?

monster860: They do not emit any additional code if there are no breakpoints.

ramapcsx2:
Great PR, this should be quite useful indeed!

I think the slight redesign of the breakpoints that you did is fine.
There has to be a system to differentiate between EE and IOP now, and this seems to do it just fine.

I'm not sure whether the few extra checks in the recompiler will be called in a release build, or whether they could affect non-debugger scenarios at all.
It would be best if any new debugger related code would simply not be used in a release build, so we can be sure there won't be any performance or stability affects.
In case this is problematic, can you please explain why that is, and what our options are?

Thanks :)

pcsx2/x86/iR3000A.cpp
```diff
@@ -434,6 +437,11 @@ void _psxFlushCall(int flushtype)
 	_freeX86reg( ecx );
 	_freeX86reg( edx );
 
+	if ((flushtype & FLUSH_PC)/*&& !g_cpuFlushedPC*/) {
```
ramapcsx2: Did you fix a register flush bug here, and/or was this required for the IOP breakpoints?

ramapcsx2: (This is the last review question from my side ;p)

monster860: This is required for IOP breakpoints otherwise the breakpoint will think it's on the wrong instruction, but this flush type isn't implemented on IOP side for some reason so I just copypasted the code for this from the EE side.

ramapcsx2:
Thanks @arcum42, I've checked as much as I could and it looks alright.
I'm lacking understanding on potential effects on the recompilers though.
It'd be great if a rec guy could do a sanity check on that.

pcsx2/x86/iR3000A.cpp
```diff
@@ -1021,11 +1029,167 @@ void rpsxBREAK()
...
+}
+
+void psxDynarecMemcheck()
+{
+	u32 pc = psxRegs.pc;
+	if (CBreakPoints::CheckSkipFirst(BREAKPOINT_IOP, pc) == pc)
+		return;
+
+	CBreakPoints::SetBreakpointTriggered(true);
+	GetCoreThread().PauseSelf();
```
AmFobes, line 1069: After hours of debugging, I found that you needed to set this is `GetCoreThread().PauseSelfDebug();` 
This fixes the  `Step over/into/out` buttons and matches the way the R5900 dynarec functions
Currently, when you use the stepping buttons on the R3000 the `break/run` button isn't properly updated and the temporary breakpoints aren't removed.

AmFobes (changes requested):
From my testing, everything besides the issue I pointed out looks good to me.

RedPanda4552:
Breakpoints work as advertised, works on recompiler but not interpreter, can now breakpoint in the R3000A. Tested on Final Fantasy 8 (PS1).

AmFobes:
> Breakpoints work as advertised, works on recompiler but not interpreter, can now breakpoint in the R3000A. Tested on Final Fantasy 8 (PS1).

That's expected, the interpreter for the EE doesn't support breakpoints (or they are broken)

PR #3013, opened 2019-07-10
pcsx2: Auto select GSdx during plugin selection

KrossX:
When there's no previous plugin set, auto select the GSdx plugin in order AVX2 -> SSE4 -> SSE2. Might help avoid having the wrong plugin selected on the first time wizard.

PS: No idea if it works, I only get the SSE2 plugin.

### DIFF
--- a/pcsx2/gui/Panels/PluginSelectorPanel.cpp
+++ b/pcsx2/gui/Panels/PluginSelectorPanel.cpp
@@ -677,7 +677,31 @@ void Panels::PluginSelectorPanel::OnEnumComplete( wxCommandEvent& evt )
 
 		else if( m_ComponentBoxes->Get(pid).GetSelection() == wxNOT_FOUND )
 		{
-			m_ComponentBoxes->Get(pid).SetSelection( 0 );
+			if( pid == PluginId_GS )
+			{
+				int count = (int)m_ComponentBoxes->Get(pid).GetCount();
+
+				int index_avx2 = -1;
+				int index_sse4 = -1;
+				int index_sse2 = -1;
+
+				for( int i = 0; i < count; i++ )
+				{
+					auto str = m_ComponentBoxes->Get(pid).GetString( i );
+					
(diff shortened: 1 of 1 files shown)


pcsx2/gui/Panels/PluginSelectorPanel.cpp
```diff
@@ -677,7 +677,31 @@ void Panels::PluginSelectorPanel::OnEnumComplete( wxCommandEvent& evt )
...
+					auto str = m_ComponentBoxes->Get(pid).GetString(i);
+					
+					if (x86caps.hasAVX2 && str.Contains("AVX2")) index_avx2 = i;
+					if (x86caps.hasStreamingSIMD4Extensions && str.Contains("SSE4")) index_sse4 = i;
+					if (str.Contains("SSE2")) index_sse2 = i;
+				}
+
+				if(index_avx2 >= 0) m_ComponentBoxes->Get(pid).SetSelection(index_avx2);
+				else if(index_sse4 >= 0) m_ComponentBoxes->Get(pid).SetSelection(index_sse4);
+				else if(index_sse2 >= 0) m_ComponentBoxes->Get(pid).SetSelection(index_sse2);
```
lightningterror: Very very tiny ocd nitpick, add spaces between if and the brackets for consistency :p
Above lines too.

mirh:
I believe for AVX2, you should also check if the OS itself supports it.

lightningterror:
> I believe for AVX2, you should also check if the OS itself supports it.

It is fine since we don't care for Vista or lower.

mirh:
That also entails W7 without SP1.

lightningterror:
Honestly people shouldn't be running an outdated system and the worst case here is they will need to manually set the plugin like before.

mirh:
There are tons of things people shouldn't do in this world... 
Anyway, I just realized that we do actually are [already](https://github.com/PCSX2/pcsx2/blob/v1.5.0-dev/common/src/x86emitter/cpudetect.cpp#L273) checking the thing the right comprehensive way.

PR #3016, opened 2019-07-12
Fix microVU debug logging.

gibbed:
A [previous commit](https://github.com/PCSX2/pcsx2/commit/ea38e2eba58ba87e8e245c8cc795363fb95cb3be#diff-136d6ae872176c8acc158d97dc2cea19L31) inadvertently removed a vsprintf call necessary for microVU debug logging to function properly.


lightningterror:
Oops, looks like I made a mistake :p

PR #3017, opened 2019-07-13
Travis: Bump Ubuntu version to Xenial

tadanokojin:
- Update Ubuntu version to Xenial as Trusty is EOL as of April
- Drop GCC 4.9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ sudo: required
 
 cache: ccache
 
+dist: xenial
+
 matrix:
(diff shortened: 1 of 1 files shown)


gregory38:
It would be nice to enable c++17 (at least the most of it) after the release. Something like gcc7+.
Anyway maybe drop the 4.9/5.0 compile. It will speedup the build process.

tadanokojin:
Fine by me. Dropped 4.9.

PR #3020, opened 2019-07-14
PCSX2: Set pad plugin and bios defaults automatically.

arcum42:
This builds on the auto GS plugin selection to automatically set a pad plugin as well, picking the non-legacy version of onepad on Linux and lilypad on windows. (I'm largely doing this because it's possible to build lilypad for Linux, and it *shouldn't* be the default, but l is before o alphabetically...)

I also added some code to pick the bios by default. This isn't perfect; if you never go to the screen where it shows the bios, it's still not set. If you do go to that window, it'll automatically pick the first bios in the list, though, which is probably the only one.

This needs testing on Windows.

### DIFF
--- a/pcsx2/gui/Panels/PluginSelectorPanel.cpp
+++ b/pcsx2/gui/Panels/PluginSelectorPanel.cpp
@@ -579,7 +579,7 @@ bool Panels::PluginSelectorPanel::ValidateEnumerationStatus()
 
 	m_FileList.swap(pluginlist);
 
-	// set the gague length a little shorter than the plugin count.  2 reasons:
+	// set the gauge length a little shorter than the plugin count.  2 reasons:
 	//  * some of the plugins might be duds.
 	//  * on high end machines and Win7, the statusbar lags a lot and never gets to 100% before being hidden.
 	
@@ -669,7 +669,9 @@ void Panels::PluginSelectorPanel::OnEnumComplete( wxCommandEvent& evt )
 	//  (for now we just force it to selection zero if nothing's selected)
 
 	int emptyBoxes = 0;
-	const PluginInfo* pi = tbl_PluginInfo; do
+	const PluginInfo* pi = tbl_PluginInfo;
+
+	do
 	{
 		const PluginsEnum_t pid = pi->id;
 		if( m_ComponentBoxes->Get(pid).GetCount() <= 0 )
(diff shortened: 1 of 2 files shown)


pcsx2/gui/Panels/PluginSelectorPanel.cpp
```diff
@@ -699,6 +701,48 @@ void Panels::PluginSelectorPanel::OnEnumComplete( wxCommandEvent& evt )
...
+					if (str.Contains("lilypad")) index_lilypad = i;
+					if (str.Contains("onepad"))
+					{
+						if (str.Contains("legacy"))
+							index_onepad_legacy = i;
+						else
+							index_onepad = i;
+					}
+				}
+
```
lightningterror: Maybe a lilypad string check would be nice too, in case there are multiple plugins added by user on windows.

Bonus: Maybe code formatting can be adjusted to better suit the rest of the code, your call.

lightningterror:
> This isn't perfect; if you never go to the screen where it shows the bios, it's still not set. 

It saves 1 click on first setup so I consider that an improvement, more automatic setup :p

pcsx2/gui/Panels/PluginSelectorPanel.cpp
arcum42: Well, it's got a check for a plugin with "lilypad" in the name, and on windows, it's supposed to pick that one over any others, whereas linux picks onepad first, onepad_legacy second, lilypad third, and anything else afterwards.

I mainly wasn't sure if "lilypad" will match what we want properly on Windows, or if there's anything else it should be searching for there.

I could adjust the formatting more, but it'd probably be better just to run it through clang_format at some point with a nice set of defaults rather than fiddling with it too much here.

For that matter, if any other plugins need defaults set, we might end up wanting to turn this into a case statement later...

pcsx2/gui/Panels/PluginSelectorPanel.cpp
```diff
@@ -711,7 +711,7 @@ void Panels::PluginSelectorPanel::OnEnumComplete( wxCommandEvent& evt )
 
 				for( int i = 0; i < count; i++ )
 				{
-					auto str = m_ComponentBoxes->Get(pid).GetString(i);
+					auto str = m_ComponentBoxes->Get(pid).GetString(i).Lower();
```
lightningterror: Maybe worth adding it to gsdx as well?

arcum42: I could, and lowercase the strings. It wasn't absolutely necessary, though, and I didn't feel like messing up the gsdx autodetection accidentally.

pcsx2/gui/Panels/PluginSelectorPanel.cpp
lightningterror: Clang would be ideal indeed.

Doing the defaults for all plugins is a nice idea too but most of the time people use the default plugins that come with pcsx2 itself. Up to you to decide what you wanna do with that.

arcum42:
Alright, cool, merging.

PR #3027, opened 2019-07-17
GSdx-HW: Dithering on Hardware

KrossX:
WIP: Initial basic implementation of dithering on the hardware renderers.

There's currently no check for dither matrix changes for buffers to be updated. **Edit: done.**
There's also no check for framebuffer format, in case it should only be enabled on some. **Edit: done.**
It currently does not scale with increased internal resolution. I kinda like the smaller dithering though.  **Edit: done, setting.**

SW renderer shows a darker more pronounced dithering although pattern seems to match.   **Edit: OpenGL can be pretty close to SW dithering.**

There should also be a setting to toggle dithering off, as many people seem to dislike the effect.  **Edit: done, setting.**

~~**TODO:** Move dither matrix to its own buffer.~~ Should be done when doing a constant buffer overhaul.

### DIFF
--- a/plugins/GSdx/Renderers/DX11/GSDevice11.h
+++ b/plugins/GSdx/Renderers/DX11/GSDevice11.h
@@ -103,6 +103,7 @@ public:
 
 		GSVector4 TC_OffsetHack;
 		GSVector4 Af;
+		GSVector4 DitherMatrix[4];
 
 		PSConstantBuffer()
 		{
@@ -115,6 +116,11 @@ public:
 			ChannelShuffle = GSVector4i::zero();
 			FbMask = GSVector4i::zero();
 			Af = GSVector4::zero();
+
+			DitherMatrix[0] = GSVector4::zero();
+			DitherMatrix[1] = GSVector4::zero();
+			DitherMatrix[2] = GSVector4::zero();
+			DitherMatrix[3] = GSVector4::zero();
 		}
 
 		__forceinline bool Update(const PSConstantBuffer* cb)
@@ -123,7 +129,8 @@ public:
 			GSVector4i* b = (GSVector4i*)cb;
 
 			if(!((a[0] == b[0]) /*& (a[1] == b1)*/ & (a[2] == b[2]) & (a[3] == b[3]) & (a[4] == b[4]) & (a[5] == b[5]) &
-				(a[6] == b[6]) & (a[7] == b[7]) & (a[9] == b[9])).alltrue()) // if WH matches HalfTexel does too
+				(a[6] == b[6]) & (a[7] == b[7]) & (a[9] == b[9]) & // if WH matches HalfTexel does too
+				(a[10] == b[10]) & (a[11] == b[11]) & (a[12] == b[12]) & (a[13] == b[13])).alltrue())
 			{
 				a[0] = b[0];
 				a[1] = b[1];
@@ -134,6 +141,11 @@ public:
 				a[6] = b[6];
 				a[7] = b[7];
 				a[9] = b[9];
+
+				a[10] = b[10];
+				a[11] = b[11];
+				a[12] = b[12];
+				a[13] = b[13];
 
 				return true;
 			}
@@ -211,16 +223,20 @@ public:
 				uint32 fbmask:1;
 
 				// Blend and Colclip
+				uint32 hdr:1;
 				uint32 blend_a:2;
-				uint32 blend_b:2;
-				uint32 blend_c:2;
+				uint32 blend_b:2; // bit30/31
+				uint32 blend_c:2; // bit0
 				uint32 blend_d:2;
 				uint32 clr1:1;
-				uint32 hdr:1;
+
 				uint32 colclip:1;
 
 				// Others ways to fetch the texture
 				uint32 channel:3;
+
+				// Dithering
+				uint32 dither:2;
 
 				// Hack
 				uint32 tcoffsethack:1;
@@ -229,7 +245,7 @@ public:
 				uint32 point_sampler:1;
 				uint32 invalid_tex0:1; // Lupin the 3rd
 
-				uint32 _free:18;
+				uint32 _free:16;
 			};
 
 			uint64 key;
--- a/plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
+++ b/plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
@@ -982,6 +982,16 @@ void GSRendererDX11::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sou
 	}
 
 	m_ps_sel.fba = m_context->FBA.FBA;
+	m_ps_sel.dither = m_dithering > 0 && m_ps_sel.dfmt == 2 && m_env.DTHE.DTHE;
+
+	if(m_ps_sel.dither)
+	{
+		m_ps_sel.dither = m_dithering;
+		ps_cb.DitherMatrix[0] = GSVector4(m_env.DIMX.DM00, m_env.DIMX.DM10, m_env.DIMX.DM20, m_env.DIMX.DM30);
+		ps_cb.DitherMatrix[1] = GSVector4(m_env.DIMX.DM01, m_env.DIMX.DM11, m_env.DIMX.DM21, m_env.DIMX.DM31);
+		ps_cb.DitherMatrix[2] = GSVector4(m_env.DIMX.DM02, m_env.DIMX.DM12, m_env.DIMX.DM22, m_env.DIMX.DM32);
+		ps_cb.DitherMatrix[3] = GSVector4(m_env.DIMX.DM03, m_env.DIMX.DM13, m_env.DIMX.DM23, m_env.DIMX.DM33);
+	}
 
 	if (PRIM->FGE)
 	{
--- a/plugins/GSdx/Renderers/HW/GSRendererHW.cpp
+++ b/plugins/GSdx/Renderers/HW/GSRendererHW.cpp
@@ -28,7 +28,6 @@ GSRendererHW::GSRendererHW(GSTextureCache* tc)
 	, m_custom_width(1024)
 	, m_custom_height(1024)
 	, m_reset(false)
-	, m_upscale_multiplier(1)
 	, m_userhacks_ts_half_bottom(-1)
 	, m_tc(tc)
 	, m_src(nullptr)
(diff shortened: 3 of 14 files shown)


plugins/GSdx/Renderers/DX11/GSDevice11.h
```diff
@@ -211,25 +217,29 @@ class GSDevice11 final : public GSDevice
...
+				uint32 hdr:1;
 				uint32 blend_a:2;
-				uint32 blend_b:2;
-				uint32 blend_c:2;
+				uint32 blend_b:2; // bit30/31
+				uint32 blend_c:2; // bit0
 				uint32 blend_d:2;
 				uint32 clr1:1;
-				uint32 hdr:1;
+
```
lightningterror: hdr was fine in original spot.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -980,6 +980,15 @@ void GSRendererDX11::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sou
 	}
 
 	m_ps_sel.fba = m_context->FBA.FBA;
+	m_ps_sel.dither = m_env.DTHE.DTHE;
+
+	if(m_ps_sel.dither)
```
lightningterror: Maybe you can do if (m_env.DTHE.DTHE) instead and add m_ps_sel.dither = 1 inside the function.

plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
```diff
@@ -1200,6 +1200,15 @@ void GSRendererOGL::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 	}
 
 	m_ps_sel.fba = m_context->FBA.FBA;
+	m_ps_sel.dither = m_env.DTHE.DTHE;
+
+	if(m_ps_sel.dither)
```
lightningterror: Ditto.

plugins/GSdx/res/glsl/tfx_fs.glsl
```diff
@@ -617,6 +617,12 @@ vec4 ps_color()
     vec4 C = tfx(T, PSin.fc);
 #endif
 
+#if PS_DITHER
+	ivec2 fpos = ivec2(gl_FragCoord.xy);
+	float dthf = DitherMatrix[fpos.y&3][fpos.x&3];
+	C.rgb += dthf;
```
lightningterror: Spaces instead of tabs for glsl shaders.

plugins/GSdx/res/tfx.fx
```diff
@@ -650,9 +652,15 @@ float4 ps_color(PS_INPUT input)
...
+	
+#if PS_DITHER == 1
+	int2 fpos = int2(input.p.xy);
+	float dthf = DitherMatrix[fpos.y&3][fpos.x&3] / 255.0;
+	c.rgb += dthf;
+#endif
 
 	atst(c);
-
+
```
lightningterror: Whitespace.

plugins/GSdx/Renderers/OpenGL/GSDeviceOGL.h
```diff
@@ -111,7 +111,7 @@ class GSDepthStencilOGL {
 class GSDeviceOGL final : public GSDevice
 {
 public:
-	struct alignas(32) VSConstantBuffer
+	struct alignas(16) VSConstantBuffer
```
lightningterror: Are the alignas changes needed?

plugins/GSdx/Renderers/DX11/GSDevice11.h
KrossX: I changed it for better alignment. https://pastebin.com/5dDQXXfn
I could pad it instead.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
KrossX: I prefer the lowercase letters. If I add ` m_ps_sel.dither = 1` inside the if, then I would need an else ` m_ps_sel.dither = 0` and kinda looks ugly.

Dreadmoth:
Seems to produce some slightly darkened squares around character shadows in Snowblind Engine games.
[Master](https://i.imgur.com/lo1O8yh.png)
[PR](https://i.imgur.com/xbfJLpi.png)
The squares are present on all Accurate Blending levels, but look much darker with it set to Ultra:
[PR](https://i.imgur.com/KmhtiJC.png)

Tested b83f7cd and 1eda351

MrCK1:
That's a side effect of dithering with blending. If you compare to SW, it should look similar

KrossX:
@Dreadmoth can you post a gsdump please? That looks interesting.

Dreadmoth:
@KrossX [GSdump](https://drive.google.com/open?id=1Q9elklNIkGXh0cSSAI0VRBNiNjL5_FjR)

> That's a side effect of dithering with blending. If you compare to SW, it should look similar

@MrCK1 Software doesn't appear to have the squares: [screenshot](https://i.imgur.com/QE7YOEz.png)

MrCK1:
I can't tell if you're upscaling or not, try turning that off?

Dreadmoth:
> I can't tell if you're upscaling or not, try turning that off?

The squares occur at native resolution and with upscaled resolutions. Affects both D3D11 and OpenGL.

KrossX:
@Dreadmoth test again please.

plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
```diff
@@ -1200,6 +1200,16 @@ void GSRendererOGL::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 	}
 
 	m_ps_sel.fba = m_context->FBA.FBA;
+	m_ps_sel.dither = m_dithering > 0 && m_ps_sel.dfmt == 2 && m_env.DTHE.DTHE;
+
+	if(m_ps_sel.dither)
+	{
+		m_ps_sel.dither = m_dithering;
```
lightningterror: Log will be nice here

> GL_INS("DITHERING mode ENABLED");

Dreadmoth:
> @Dreadmoth test again please.

Squares are fixed. :+1:

plugins/GSdx/res/glsl/tfx_fs.glsl
```diff
@@ -842,6 +855,11 @@ void ps_main()
     return;
 #endif
 
+#if !SW_BLEND && PS_DITHER
+    ps_dither(C);
+    C.rgb -= 3.0;
```
turtleli: What does the `-= 3.0` do?

turtleli:
Makes Suikoden 4 and Suikoden Tactics look more similar to SW.

plugins/GSdx/res/glsl/tfx_fs.glsl
KrossX: The dither matrix values range is [-4, 3], so that -3 makes the range [-7, 0]. This is to avoid incrementing values as it causes artifacts to show up (see screenshots in comments). There's no such problem when software blending is used.

turtleli: It'd be worth adding a comment to the code explaining this (it can wait till after the other PR is merged though).

KrossX: Done.

plugins/GSdx/res/glsl/tfx_fs.glsl
```diff
@@ -692,7 +704,8 @@ void ps_blend(inout vec4 Color, float As)
     Color.rgb = trunc((A - B) * C + D);
 #endif
 
-    // FIXME dithering
+    // Dithering
+    ps_dither(Color);
```
lightningterror: Missing on dx?

KrossX: Done!

plugins/GSdx/res/glsl/tfx_fs.glsl
```diff
@@ -842,6 +855,13 @@ void ps_main()
     return;
 #endif
 
+#if !SW_BLEND && PS_DITHER
+    ps_dither(C);
+    // Dither matrix range is [-4,3] but positive values can cause issues when
+    // software blending is not used or is unavailable.
+    C.rgb -= 3.0;
```
turtleli: So, what if C.rgb is (0,0,0)? Would there be issues in that case if you subtract 3?

KrossX: Gets clamped to zero, eventually. It's a hackfix for the squares showing up as you can see in the posted screenshots in the comments.

ssakash:
I'm still not sure why there's a scaling mode (``value 2``), if this mode is needed to properly emulate the behavior on upscaled resolutions wouldn't it make sense to have an implicit transition between 1/2 depending on if upscaling is used or not? Rather than a manual change through INI file.

Also I think ``PS_Scale`` and ``ScalingFactor`` directly get their values from ``upscale_multiplier``, IIRC custom resolution has a value of 0, so there might be a potential for division by zero somewhere.

turtleli:
> Also I think `PS_Scale` and `ScalingFactor` directly get their values from `upscale_multiplier`, IIRC custom resolution has a value of 0, so there might be a potential for division by zero somewhere.

`PS_SCALE_FACTOR`/`ScalingFactor` is 1 in the custom resolution case.

plugins/GSdx/Renderers/HW/GSRendererHW.cpp
```diff
@@ -37,11 +37,13 @@ GSRendererHW::GSRendererHW(GSTextureCache* tc)
 	, m_userhacks_tcoffset_y(0)
 	, m_channel_shuffle(false)
 	, m_lod(GSVector2i(0,0))
+	, m_dithering(1)
```
turtleli, line 40: Isn't this line unnecessary if it gets reassigned a few lines later?

KrossX: I dunno, would need to verify it. I think I use copied `m_upscale_multiplier`.

KrossX: Confirmed unnecessary for `m_dithering` and `m_upscale_multiplier`. Should I remove both while at it?

lightningterror: I'd say remove both if unnecessary.

turtleli: It's up to you.

PR #3029, opened 2019-07-17
Gamedb: Patch for Jak X NTSC Public Beta

PlasmaLight:
Patches HookOrDie()

Emulator can't boot this without the patch. All Jak X builds have this problem, but soon I will patch the few remaining ones too.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -1,4 +1,4 @@
----------------------------------------------
+﻿---------------------------------------------
 -- PCSX2 Game DataBase!
 ---------------------------------------------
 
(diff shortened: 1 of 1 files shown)


PlasmaLight:
Will this get merged before 1.6?

lightningterror:
I think there was also a patch posted on the wiki, is this one better than that?

PlasmaLight:
I can't see any patches on the wiki.

lightningterror:
https://forums.pcsx2.net/Thread-Jak-X-Patched-NTSC-and-PAL

PlasmaLight:
Those are for the retail version. This patch is for the NTSC Multiplayer Beta. My patches work the same way as Ge-Force's patches as far as I understand.

PlasmaLight:
Is it clear now? Do I need to explain more? I did the exact same thing for the PAL Multiplayer Beta: https://github.com/PCSX2/pcsx2/commit/74aefcb9187f70c3a79851c37adca81d7dedd734

PR #3030, opened 2019-07-18
Spu2-x: Fix noise generator

monster860:
Fixes the noise generator outputting a 4.8 kHz tone instead of white noise. The random number generator used in GetNoiseValues currently repeats every 10 samples, which is really really awful for a random number generator.

Here's what it sounded like before the fix: https://streamable.com/pa1fi
Here's what it sounds like after the fix: https://streamable.com/u5zhi
Here's what it sounds like on a real PS2: https://streamable.com/ipi5s

Not entirely accurate but a massive improvement over playing a 4.8 kHz tone.

### DIFF
--- a/plugins/spu2-x/src/Mixer.cpp
+++ b/plugins/spu2-x/src/Mixer.cpp
@@ -257,7 +257,7 @@ static __forceinline void GetNextDataDummy(V_Core &thiscore, uint voiceidx)
 
 static s32 __forceinline GetNoiseValues()
 {
-    static s32 Seed = 0x41595321;
+    /*static s32 Seed = 0x41595321;
     s32 retval = 0x8000;
 
     if (Seed & 0x100)
@@ -272,7 +272,8 @@ static s32 __forceinline GetNoiseValues()
     y ^= x;
     Seed = _rotr(y, 0x3);
 
-    return retval;
+    return retval;*/
+    return (s32)(rand() & 0xFF) << 8;
 }
 /////////////////////////////////////////////////////////////////////////////////////////
 /////////////////////////////////////////////////////////////////////////////////////////
--- a/plugins/spu2-x/src/Mixer.cpp
+++ b/plugins/spu2-x/src/Mixer.cpp
@@ -257,23 +257,12 @@ static __forceinline void GetNextDataDummy(V_Core &thiscore, uint voiceidx)
 
 static s32 __forceinline GetNoiseValues()
 {
-    /*static s32 Seed = 0x41595321;
-    s32 retval = 0x8000;
+    static u16 lfsr = 0xC0FEu;
 
-    if (Seed & 0x100)
-        retval = (Seed & 0xff) << 8;
-    else if (Seed & 0xffff)
-        retval = 0x7fff;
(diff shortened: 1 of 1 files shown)


MrCK1:
Is there some sort of "list" of games that happen to use the white noise generator? I don't know of other titles that can be tested to check behavior.

arcum42:
Just straight rand leaves the implementation subject to variation by platform. There are some random number generators over in std::random. Might want to look at what difference the engine makes. And should this actually be white noise, or pink noise?

http://www.cplusplus.com/reference/random/

ramapcsx2:
@gigaherz Hey, any idea what's going on here? ;p

atomic83GitHub:
This fixes noise problems in Rathet and Clank 1 in Kalebo 3

gigaherz:
🤷‍♂️ I never found anything using noise so I never tried to make it work properly.

monster860:
I think Ape Escape 3 also uses the white noise.

willkuer:
But why was the initial implementation how it was? Is this some well-known random value generator algo?

PSI-Rockin:
nocash's documentation on the PSX describes the noise generator as such: http://problemkaputt.de/psx-spx.htm#spunoisegenerator

The PSX SPU is similar to SPU2, so it's likely that the SPU2 noise generator follows this same function. I've no idea where PCSX2's current noise implementation comes from.

gigaherz:
The original implementation was written by me in 2003-2004 with practically 0 knowledge of the ps1 or ps2. I had no clue how to implement a noise channel for audio purposes but I knew that random values sounded like noise if you listen to them, so that's what I coded.

PSI-Rockin:
If that's the case, then I recommend replacing the rand() implementation this PR has with the one I linked from nocash.

gigaherz:
Yeah, the thing described in that document matches what I believe is the most common form of noise generation on sound chips: [an LFSR](https://en.wikipedia.org/wiki/Linear-feedback_shift_register). One configured in such a way as to give a noise of a certain pitch and with a certain cycle length. It is still a random number generator, just one that's tuned to output predictable numbers that "sound right".
I know this is the case in other sound chips, including the NES APU and the SID chip used in the C64 and friends.

KrossX:
I tried this from nocash docs. Sounds noisy enough, but it's s16 as opposed to s32. I tested the commented out code and it just sounds like a tone indeed. 

```cpp
s16 lfsr(void)
{
	static u16 lfsr = 0xC0FEu;
	
	u16 bit = lfsr ^ (lfsr<<3) ^ (lfsr<<4) ^ (lfsr<<5);
	lfsr = (lfsr<<1) | (bit>>15);
	
	return (s16)lfsr;
}
```

gigaherz:
just ((s32)noise)<<16 to convert to spu2-x format

lightningterror:
Any other pending changes/reviews or can it be merged?

monster860:
should be mergable I think

KrossX:
Does it sound good enough?

ramapcsx2:
original: https://instaud.io/3YZI
improved: https://instaud.io/3YZJ

I'd say it's good :)

PR #3033, opened 2019-07-21
GameDB: Add fixes for Tom Clancy's Splinter Cell Chaos Theory

sotpapathe:
Enable the fix for showing FMVs in software mode. Tested personally on PAL SLES-53007 but also reported to be an issue in the NTSC-U version [here](https://wiki.pcsx2.net/Tom_Clancy%27s_Splinter_Cell_Chaos_Theory).

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -13356,6 +13356,7 @@ Serial = SLES-53007
 Name   = Tom Clancy's Splinter Cell - Chaos Theory
 Region = PAL-M5
 Compat = 5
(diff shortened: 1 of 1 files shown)


sotpapathe:
The fix does not seem to work after all, only manually switching to software mode and back to hardware fixes the issue.

willkuer:
So either your crc or the fmv is not properly detected.

sotpapathe:
Even if I manually enable the fix in Emulation Settings/Game Fixes it still does not work. So I would guess the problem is in the FMV detection. Any tips on how I can further investigate this?

willkuer:
I think you nailed it already. I believe this detection is fiddly as hell.

PR #3038, opened 2019-07-25
[Lilypad] Update Xinput description

atomic83GitHub:
This commit add the Xbox One controller mention in the Xinput description. This include the Mediatek wireless adapter. PS, yeah this is for the unkown,special and unique guy.

### DIFF
--- a/plugins/LilyPad/Tooltips.cpp
+++ b/plugins/LilyPad/Tooltips.cpp
@@ -64,7 +64,7 @@ LPWSTR dialog_message(int ID, bool *updateText)
                    L"Use this option to check if all the inputs on a controller function properly.\n\n"
                    L"Right-click and select \"Refresh\" to update the list of devices in case a recently connected device has not shown up yet.";
         case IDC_G_XI:
-            return L"Xbox 360 controllers(and devices imitating it) only";
+            return L"Xbox 360/Xbox One controllers (or devices/software imitating such) only";
         case ID_RESTORE_DEFAULTS:
             return L"Restores the default contents of LilyPad.ini, undoing all settings changes and bindings that have been set up.";
         // Pad tabs


plugins/LilyPad/Tooltips.cpp
```diff
@@ -64,7 +64,7 @@ LPWSTR dialog_message(int ID, bool *updateText)
                    L"Use this option to check if all the inputs on a controller function properly.\n\n"
                    L"Right-click and select \"Refresh\" to update the list of devices in case a recently connected device has not shown up yet.";
         case IDC_G_XI:
-            return L"Xbox 360 controllers(and devices imitating it) only";
+            return L"Xbox 360/Xbox ONE controllers (and devices or software imitating it) only";
```
CookiePLMonster: Official naming is `Xbox One`, so no need to have it in all caps.

atomic83GitHub: Fixed.

plugins/LilyPad/Tooltips.cpp
```diff
@@ -64,7 +64,7 @@ LPWSTR dialog_message(int ID, bool *updateText)
                    L"Use this option to check if all the inputs on a controller function properly.\n\n"
                    L"Right-click and select \"Refresh\" to update the list of devices in case a recently connected device has not shown up yet.";
         case IDC_G_XI:
-            return L"Xbox 360 controllers(and devices imitating it) only";
+            return L"Xbox 360/Xbox One controllers (and devices or software imitating it) only";
```
pgert: "Xbox 360/One controllers (or devices/software imitating such) only"?

atomic83GitHub: I put your sugestion into the commit. It's true that looks better.

atomic83GitHub:
Yes, my mistake, I will remove that commit from the PR.

Issue #2727, opened 2018-11-30
Build Error Spu2-X


PGAnonamous:
This has been for the last few commits. I have recently had to reinstall windows so I'm not sure if it is because I'm missing something but I cannot seem to figure it out.

The error I get is from windows header xaudio.h

"This version of XAudio2 is available only in Windows 8 or later. Use the XAudio2 headers and libraries from the DirectX SDK with applications that target Windows 7 and earlier versions."

I am running Windows 10 Education 64-bit 1809 (Build 17763.134) and am using Visual Studio 15 2017 15.9.3 with Windows 10 SDK 10.0.17763.0 installed. The error the header is raising doesn't seem to make any sense.

tadanokojin:
There's a define guard in xaudio that prevents you from building it without the proper sdk.
It forces your windows version to 7 and the SDK you're using complains about it.

PGAnonamous:
> There's a define guard in xaudio that prevents you from building it without the proper sdk.
> It forces your windows version to 7 and the SDK you're using complains about it.

I see. What version should I be using?

tadanokojin:
I've had luck installing directxtk_desktop_2013 from VS2017 nuget package manager but building spu2-x isn't a requirement. You can disable the build and use a release version of the plugin instead.

PGAnonamous:
That is what I ended up doing. It is just strange that I haven't run into the issue previously.

tadanokojin:
Perhaps your old install had the older sdk on it.

lightningterror:
Perhaps the build documentation can be updated a bit for Windows 8 and up.

tadanokojin:
I believe @AlessandroVetere expressed interest in that. If not, I can do it.

PGAnonamous:
It would be extremely appreciated as I still have not figured out what I had installed to make it build without a problem.

AlessandroVetere:
I didn't manage to get the plugin built so...

tadanokojin:
I'll update then. Been meaning to do it just been busy with other things.

PGAnonamous:
So I changed '#if _WIN32_WINNT >= 0x602' on line 21 in SndOut_XAudio2_27.cpp to '#if #if _WIN32_WINNT == _WIN32_WINNT_VISTA' and it happily throws the error that it is time to remove the module.

[...]

PR #3042, opened 2019-07-27
Fix SPU2x XAudio 2.7 compilation

KrossX:
Includes four headers from the DirectX SDK with some `#include <dxfile>` into `#include "dxfile"` changes.

### DIFF
--- a/3rdparty/xaudio2/XAudio2.h
+++ b/3rdparty/xaudio2/XAudio2.h
@@ -1,0 +1,1282 @@
+/**************************************************************************
+ *
+ * Copyright (c) Microsoft Corporation.  All rights reserved.
+ *
+ * File:    xaudio2.h
+ * Content: Declarations for the XAudio2 game audio API.
+ *
+ **************************************************************************/
+
+#ifndef __XAUDIO2_INCLUDED__
+#define __XAUDIO2_INCLUDED__
+
+
+/**************************************************************************
+ *
+ * XAudio2 COM object class and interface IDs.
+ *
+ **************************************************************************/
+
+#include "comdecl.h"        // For DEFINE_CLSID and DEFINE_IID
+
+// XAudio 2.0 (March 2008 SDK)
+//DEFINE_CLSID(XAudio2, fac23f48, 31f5, 45a8, b4, 9b, 52, 25, d6, 14, 01, aa);
+//DEFINE_CLSID(XAudio2_Debug, fac23f48, 31f5, 45a8, b4, 9b, 52, 25, d6, 14, 01, db);
+
+// XAudio 2.1 (June 2008 SDK)
+//DEFINE_CLSID(XAudio2, e21a7345, eb21, 468e, be, 50, 80, 4d, b9, 7c, f7, 08);
+//DEFINE_CLSID(XAudio2_Debug, f7a76c21, 53d4, 46bb, ac, 53, 8b, 45, 9c, ae, 46, bd);
+
+// XAudio 2.2 (August 2008 SDK)
+//DEFINE_CLSID(XAudio2, b802058a, 464a, 42db, bc, 10, b6, 50, d6, f2, 58, 6a);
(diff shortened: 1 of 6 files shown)


lightningterror:
I believe this should also be merged for 1.6 release as it provides us with a fully working build that compiles without issues which will be later the 1.6 branch.

XAudio 2.7 will be dropped for 1.8 release so current pr/code will be only temporarily there.

turtleli:
> as it provides us with a fully working build

? What currently doesn't work?

lightningterror:
There is an issue with newer VS 2017/19 with compiling spu2x on Windows 10, a misdetection between 2.7 and 2.8. I don't know the full details since I'm on w7 still.

KrossX:
This just makes XAudio 2.7 build without requiring DX SDK. It should also avoid it being skipped when built on a Windows 8+ OS so that such a build would include XAudio 2.7 for Windows 7 users.

Current automatic builds work just fine.

turtleli:
> It should also avoid it being skipped when built on a Windows 8+ OS so that such a build would include XAudio 2.7 for Windows 7 users.

This doesn't make sense to me. It should build the same regardless of operating system.

KrossX:
It just means that if `_WIN32_WINNT >= 0x602` there's no reason to skip 2.7. It's currently set as 0x600, if undefined.

Still, automatic builds are building just fine. This just makes it easier for me to build it by switching branches. Also, if W7 support will be removed after 1.6 release, there's no reason to merge this at all.

lightningterror:
> Also, if W7 support will be removed after 1.6 release, there's no reason to merge this at all.

Don't take away my baby :cry:

mirh:
> Don't take away my baby

You were the one saying this?
It's not like you can keep W7 if you drop 2.7. 

And of course it doesn't make sense for me to do so.

lightningterror:
We had a discussion on discord to use Xaudio on windows 10 and Portaudio on lower.

tadanokojin:
@turtleli Li, it does build but you need some SDK for win10. Otherwise there is a define guard that produces some nonsense build error about your windows not being recent.

turtleli:
It builds on my end without requiring modifications.

What error message pops up?

tadanokojin:
> This version of XAudio2 is available only in Windows 8 or later. Use the XAudio2 headers and libraries from the DirectX SDK with applications that target Windows 7 and earlier versions.

#2727

turtleli:
That's not actually talking about what Windows version you're using.

It's saying that you're trying to support Windows 7 and earlier (detected from the _WIN32_WINNT define) and that the header is for XAudio2.8+.

That error should only pop up if the DXSDK is not detected.

mirh:
> We had a discussion on discord to use Xaudio 2.8 on windows 10 and Portaudio on lower.

Eww. I don't know, it always gave me a very cheap impression I can't put my finger on. 
But whatever.

lightningterror:
I also tested a while back with appveyor vs 2019 and got the same result, maybe I was missing something extra?
https://ci.appveyor.com/project/lightningterror/pcsx2/builds/25959443/job/6mkahcx923lg1ep2

`         C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\xaudio2.h(20,1): error C1189:  #error:  "This version of XAudio2 is available only in Windows 8 or later. Use the XAudio2 headers and libraries from the DirectX SDK with applications that target Windows 7 and earlier versions." [C:\projects\pcsx2\plugins\spu2-x\src\Windows\Spu2-X.vcxproj]`

turtleli:
https://github.com/appveyor/ci/issues/2953

KrossX:
Turteli, click this thing -> https://discord.gg/TCz3t9k

tadanokojin:
> It's saying that you're trying to support Windows 7 and earlier (detected from the _WIN32_WINNT define) and that the header is for XAudio2.8+.

Would be nice if it just error out with the link to the SDK. Error is kinda confusing and I'm not the first to think so.

Anyway, not to be off-topic. Is it even legal to include MS code like this?

PR #3047, opened 2019-07-29
GSDumpGUI: Usability improvements

KrossX:
- [x] Changed the main loop to process messages per frame which makes the GSdx window more responsive and allows for the window title information to show up. The plugin has its own message loop that runs on vsync, GUI message loop then runs before the vsync package to avoid missed messages.
- [x] Process termination is done using the new message loop.
- [x] Window icon is now set after GSopen, as there's a valid window handle to use.
- [x] Added GSmakeSnapshot to the loaded functions and is used when pressing the F8 key.
- [x] Added FileSystemWatchers to check for changes in dump and dll paths to update the respective lists.
- [x] Replaced directory selection tree dialog with a normal file selection dialog.
- [x] Fixed path labels updating lists on change.
- [x] Enabled and fixed STDERR redirection for the plugin process.
- [x] Avoid leaving opened file handles.
- [x] Fixed some exceptions.
- [x] Make it work in 64bit.

### DIFF
--- a/tools/GSDumpGUI/Forms/frmMain.cs
+++ b/tools/GSDumpGUI/Forms/frmMain.cs
@@ -99,6 +99,8 @@ namespace GSDumpGUI
 
         private ConcurrentQueue<int> watcher_events;
 
+        private string gsdx_path_old, dump_path_old;
+
         public GSDumpGUI()
         {
             PortableXmlSettingsProvider.ApplyProvider(Settings);
@@ -114,8 +116,8 @@ namespace GSDumpGUI
 
             _availableGsDumps.OnIndexUpdatedEvent += UpdatePreviewImage;
 
-            txtGSDXDirectory.DataBindings.Add(nameof(TextBox.Text), Settings, nameof(Settings.GSDXDir));
-            txtDumpsDirectory.DataBindings.Add(nameof(TextBox.Text), Settings, nameof(Settings.DumpDir));
+            txtGSDXDirectory.Text = Settings.GSDXDir;
+            txtDumpsDirectory.Text = Settings.DumpDir;
 
             BindListControl(lstDumps, _availableGsDumps, g => g.Files, f => f.DisplayText, g => g.SelectedFileIndex);
             BindListControl(lstGSDX, _availableGsDlls, g => g.Files, f => f.DisplayText, g => g.SelectedFileIndex);
@@ -277,30 +279,46 @@ namespace GSDumpGUI
 
         private void cmdBrowseGSDX_Click(object sender, EventArgs e)
         {
-            FolderBrowserDialog fbd = new FolderBrowserDialog();
-            fbd.Description = "Select the GSdx DLL Directory";
-            fbd.SelectedPath = AppDomain.CurrentDomain.BaseDirectory;
-            if (fbd.ShowDialog() == DialogResult.OK)
-                Settings.GSDXDir = fbd.SelectedPath;
-            Settings.Save();
-            ReloadGsdxDlls();
+            OpenFileDialog ofd = new OpenFileDialog();
+            ofd.ValidateNames = false;
+            ofd.CheckFileExists = false;
+            ofd.CheckPathExists = true;
+            ofd.InitialDirectory = Settings.GSDXDir;
+            ofd.FileName = "Select Folder";
 
-            // Auto select GSdx dll
-            _availableGsDlls.Selected = _availableGsDlls.Files.FirstOrDefault();
+            if(ofd.ShowDialog() == DialogResult.OK)
+            {
+                string newpath = Path.GetDirectoryName(ofd.FileName);
+                if (!Settings.GSDXDir.ToLower().Equals(newpath.ToLower()))
+                {
+                    Settings.GSDXDir = newpath;
+                    Settings.Save();
+                    ReloadGsdxDlls();
+                    _availableGsDlls.Selected = _availableGsDlls.Files.FirstOrDefault();
+                }
+            }
         }
 
         private void cmdBrowseDumps_Click(object sender, EventArgs e)
         {
-            FolderBrowserDialog fbd = new FolderBrowserDialog();
-            fbd.Description = "Select the GSdx Dumps Directory";
-            fbd.SelectedPath = AppDomain.CurrentDomain.BaseDirectory;
-            if (fbd.ShowDialog() == DialogResult.OK)
-                Settings.DumpDir = fbd.SelectedPath;
-            Settings.Save();
-            ReloadGsdxDumps();
+            OpenFileDialog ofd = new OpenFileDialog();
+            ofd.ValidateNames = false;
+            ofd.CheckFileExists = false;
+            ofd.CheckPathExists = true;
+            ofd.InitialDirectory = Settings.DumpDir;
+            ofd.FileName = "Select Folder";
 
-            // Auto select GS dump
-            _availableGsDumps.Selected = _availableGsDumps.Files.FirstOrDefault();
+            if (ofd.ShowDialog() == DialogResult.OK)
+            {
+                string newpath = Path.GetDirectoryName(ofd.FileName);
+                if (!Settings.DumpDir.ToLower().Equals(newpath.ToLower()))
+                {
+                    Settings.DumpDir = newpath;
+                    Settings.Save();
+                    ReloadGsdxDumps();
+                    _availableGsDumps.Selected = _availableGsDumps.Files.FirstOrDefault();
+                }
+            }
         }
 
         private void cmdRun_Click(object sender, EventArgs e)
@@ -479,22 +497,44 @@ namespace GSDumpGUI
                 SelectedRad = Convert.ToInt32(itm.Tag);
         }
 
+        private void txtGSDXDirectory_Enter(object sender, EventArgs e)
+        {
+            gsdx_path_old = txtGSDXDirectory.Text;
+        }
+
+        private void txtDumpsDirectory_Enter(object sender, EventArgs e)
+        {
+            dump_path_old = txtDumpsDirectory.Text;
+        }
+
         private void txtGSDXDirectory_Leave(object sender, EventArgs e)
         {
-            Settings.Save();
-            ReloadGsdxDlls();
+            string newpath = txtGSDXDirectory.Text;
 
-            // Auto select GSdx dll
-            _availableGsDlls.Selected = _availableGsDlls.Files.FirstOrDefault();
+            if (!String.IsNullOrEmpty(newpath) &&
+                !gsdx_path_old.ToLower().Equals(newpath.ToLower()) &&
+                Directory.Exists(newpath))
+            {
+                Settings.GSDXDir = newpath;
+                Settings.Save();
+                ReloadGsdxDlls();
+                _availableGsDlls.Selected = _availableGsDlls.Files.FirstOrDefault();
+            }
         }
 
         private void txtDumpsDirectory_Leave(object sender, EventArgs e)
         {
-            Settings.Save();
-            ReloadGsdxDumps();
+            string newpath = txtDumpsDirectory.Text;
 
-            // Auto select GS dump
-            _availableGsDumps.Selected = _availableGsDumps.Files.FirstOrDefault();
+            if (!String.IsNullOrEmpty(newpath) &&
+                !dump_path_old.ToLower().Equals(newpath.ToLower()) &&
+                Directory.Exists(newpath))
+            {
+                Settings.DumpDir = newpath;
+                Settings.Save();
+                ReloadGsdxDumps();
+                _availableGsDumps.Selected = _availableGsDumps.Files.FirstOrDefault();
+            }
         }
 
         private void lstProcesses_SelectedIndexChanged(object sender, EventArgs e)
--- a/tools/GSDumpGUI/Forms/frmMain.cs
+++ b/tools/GSDumpGUI/Forms/frmMain.cs
@@ -494,7 +494,7 @@ namespace GSDumpGUI
 
         private void GSDumpGUI_KeyDown(object sender, KeyEventArgs e)
         {
-            if (e.KeyCode == Keys.Return)
+            if (e.KeyCode == Keys.Return && !txtGSDXDirectory.Focused && !txtDumpsDirectory.Focused)
                 cmdRun_Click(sender, e);
 
             if (e.KeyCode == Keys.F1)
@@ -524,7 +524,22 @@ namespace GSDumpGUI
         private void txtGSDXDirectory_Leave(object sender, EventArgs e)
         {
             string newpath = txtGSDXDirectory.Text;
+            if (!_gsdxPathOld.ToLower().Equals(newpath.ToLower()))
+                txtGSDXDirectory.Text = _gsdxPathOld;
+        }
 
+        private void txtDumpsDirectory_Leave(object sender, EventArgs e)
+        {
+            string newpath = txtDumpsDirectory.Text;
+            if(!_dumpPathOld.ToLower().Equals(newpath.ToLower()))
+                txtDumpsDirectory.Text = _dumpPathOld;
+        }
+
+        private void txtGSDXDirectory_KeyDown(object sender, KeyEventArgs e)
+        {
+            if (e.KeyCode != Keys.Return) return;
+
+            string newpath = txtGSDXDirectory.Text;
             if (!String.IsNullOrEmpty(newpath) &&
                 !_gsdxPathOld.ToLower().Equals(newpath.ToLower()) &&
                 Directory.Exists(newpath))
@@ -536,10 +551,11 @@ namespace GSDumpGUI
             }
         }
 
-        private void txtDumpsDirectory_Leave(object sender, EventArgs e)
+        private void txtDumpsDirectory_KeyDown(object sender, KeyEventArgs e)
         {
-            string newpath = txtDumpsDirectory.Text;
+            if (e.KeyCode != Keys.Return) return;
 
+            string newpath = txtDumpsDirectory.Text;
             if (!String.IsNullOrEmpty(newpath) &&
                 !_dumpPathOld.ToLower().Equals(newpath.ToLower()) &&
                 Directory.Exists(newpath))
--- a/tools/GSDumpGUI/Forms/frmMain.cs
+++ b/tools/GSDumpGUI/Forms/frmMain.cs
@@ -116,7 +116,7 @@ namespace GSDumpGUI
 
             _availableGsDumps.OnIndexUpdatedEvent += UpdatePreviewImage;
 
-            this.Text += " " + (IntPtr.Size * 8).ToString() + "bits";
+            this.Text += Environment.Is64BitProcess ? " 64bits" : " 32bits";
 
             if (String.IsNullOrEmpty(Settings.GSDXDir) || !Directory.Exists(Settings.GSDXDir))
                 Settings.GSDXDir = AppDomain.CurrentDomain.BaseDirectory;
@@ -217,7 +217,7 @@ namespace GSDumpGUI
 
             _dllWatcher.Clear();
 
-            for (int i = 0; i < 4; i++)
+            for (int i = 0; i < paths.Length; i++)
             {
                 try
                 {
@@ -256,7 +256,7 @@ namespace GSDumpGUI
 
             _dumpWatcher.Clear();
 
-            for (int i = 0; i < 3; i++)
+            for (int i = 0; i < paths.Length; i++)
             {
                 try
                 {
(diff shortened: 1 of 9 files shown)


willkuer:
I’d like to review this but I am currently out of time and don’t know when I am free.

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -107,15 +116,61 @@ public GSDumpGUI()
 
             _availableGsDumps.OnIndexUpdatedEvent += UpdatePreviewImage;
 
-            txtGSDXDirectory.DataBindings.Add(nameof(TextBox.Text), Settings, nameof(Settings.GSDXDir));
-            txtDumpsDirectory.DataBindings.Add(nameof(TextBox.Text), Settings, nameof(Settings.DumpDir));
+            this.Text += " " + (IntPtr.Size * 8).ToString() + "bits";
```
willkuer: That’s such a fancy non-.Net way to write this. I would’ve used Environment.Is64bit or something but it’s cool how c-based people think.

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -107,15 +116,61 @@ public GSDumpGUI()
 
             _availableGsDumps.OnIndexUpdatedEvent += UpdatePreviewImage;
 
-            txtGSDXDirectory.DataBindings.Add(nameof(TextBox.Text), Settings, nameof(Settings.GSDXDir));
```
willkuer: It was necessary to remove model binding? Typically you want to separate view and model logic because on data level you don’t care how and where data is displayed and on view level you don’t care who is when changing data but just that you need to update views if data changes.

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -107,15 +116,61 @@ public GSDumpGUI()
...
+        {
+            while (Program.isProgRunning)
+            {
+                bool dllReload = false;
+                bool dumpReload = false;
+
+                int evt;
+                while (_watcherEvents.TryDequeue(out evt))
+                {
+                    if (evt == 1) dllReload = true;
```
willkuer: I’d use an enum for 1 and 2 to improve readability/maintainability

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -107,15 +116,61 @@ public GSDumpGUI()
...
 
             Processes = new List<Process>();
 
             NoImage = CreateDefaultImage();
+
+            _dllWatcher = new List<FileSystemWatcher>();
+            _dumpWatcher = new List<FileSystemWatcher>();
+            _watcherEvents = new ConcurrentQueue<int>();
+
+            Thread watcherThread = new Thread(ChangesWatchdogThread);
```
willkuer: Zombie threads apparently were an issue before. So I would suggest to cancel and dispose this in dispose method.

willkuer: There is also something like a BackgroundThread for such work but might not be available on this framework version

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -107,15 +116,61 @@ public GSDumpGUI()
...
+                while (_watcherEvents.TryDequeue(out evt))
+                {
+                    if (evt == 1) dllReload = true;
+                    else if (evt == 2) dumpReload = true;
+                }
+
+                if (dllReload) this.Invoke((MethodInvoker)delegate { ReloadGsdxDlls(); });
+                if (dumpReload) this.Invoke((MethodInvoker)delegate { ReloadGsdxDumps(); });
+
+                Thread.Sleep(1000);
```
willkuer: With other words a Timer with interval of 1s would be an alternative without that you manually need to handle threads.

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -151,6 +206,31 @@ private void ReloadGsdxDlls()
...
+
+            foreach (FileSystemWatcher w in _dllWatcher)
+            {
+                w.EnableRaisingEvents = false;
+                w.Dispose();
+            }
+
+            _dllWatcher.Clear();
+
+            for (int i = 0; i < 4; i++)
```
willkuer: Path.Length

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -151,6 +206,31 @@ private void ReloadGsdxDlls()
 
             Settings.GSDXDir = gsdxFolder.FullName;
             _internalLogger.Information("Completed GSdx Loading Procedures");
+
+            string[] paths = { "", "\\plugins", "\\dll", "\\dlls" };
```
willkuer: Does that mean that there are no custom paths anymore?

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -165,6 +245,31 @@ private void ReloadGsdxDumps()
...
+
+            foreach (FileSystemWatcher w in _dumpWatcher)
+            {
+                w.EnableRaisingEvents = false;
+                w.Dispose();
+            }
+
+            _dumpWatcher.Clear();
+
+            for (int i = 0; i < 3; i++)
```
willkuer: Paths.length

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -151,6 +206,31 @@ private void ReloadGsdxDlls()
...
+                {
+                    FileSystemWatcher w = new FileSystemWatcher(Settings.GSDXDir + paths[i], "*.dll");
+                    //w.Changed += OnDllDirChange;
+                    w.Created += OnDllDirChange;
+                    w.Deleted += OnDllDirChange;
+                    w.Renamed += OnDllDirChange;
+                    w.EnableRaisingEvents = true;
+                    _dllWatcher.Add(w);
+                }
+                catch { }
```
willkuer: I think there should be a logger available here. Why not logging errors?

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -179,30 +284,48 @@ private void GSDumpGUI_Load(object sender, EventArgs e)
...
+            ofd.InitialDirectory = Settings.GSDXDir;
+            ofd.FileName = "Select Folder";
+
+            if(ofd.ShowDialog() == DialogResult.OK)
+            {
+                string newpath = Path.GetDirectoryName(ofd.FileName);
+                if (!Settings.GSDXDir.ToLower().Equals(newpath.ToLower()))
+                {
+                    txtGSDXDirectory.Text = newpath;
+                    Settings.GSDXDir = newpath;
```
willkuer: This double assignment is the reason you typically prefer modelbinding.

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -179,30 +284,48 @@ private void GSDumpGUI_Load(object sender, EventArgs e)
...
+            ofd.ValidateNames = false;
+            ofd.CheckFileExists = false;
+            ofd.CheckPathExists = true;
+            ofd.InitialDirectory = Settings.GSDXDir;
+            ofd.FileName = "Select Folder";
+
+            if(ofd.ShowDialog() == DialogResult.OK)
+            {
+                string newpath = Path.GetDirectoryName(ofd.FileName);
+                if (!Settings.GSDXDir.ToLower().Equals(newpath.ToLower()))
```
willkuer: Equals(other, StringComparison.OrdinalIgnoreCase) instead of lowering

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -364,7 +494,7 @@ private static void PreviewImageClick(object sender, EventArgs e)
 
         private void GSDumpGUI_KeyDown(object sender, KeyEventArgs e)
         {
-            if (e.KeyCode == Keys.Return)
+            if (e.KeyCode == Keys.Return && !txtGSDXDirectory.Focused && !txtDumpsDirectory.Focused)
```
willkuer: Nice

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -381,22 +511,60 @@ private void rda_CheckedChanged(object sender, EventArgs e)
...
+            string newpath = txtDumpsDirectory.Text;
+            if(!_dumpPathOld.ToLower().Equals(newpath.ToLower()))
+                txtDumpsDirectory.Text = _dumpPathOld;
+        }
 
-            // Auto select GS dump
-            _availableGsDumps.Selected = _availableGsDumps.Files.FirstOrDefault();
+        private void txtGSDXDirectory_KeyDown(object sender, KeyEventArgs e)
+        {
+            if (e.KeyCode != Keys.Return) return;
```
willkuer: My personal opinion but I don’t like to hide control flow jumps in row 123432543 in a line because it’s harder to read. But yeah there are different opinions about that.

tools/GSDumpGUI/Forms/frmMain.cs
```diff
@@ -92,6 +94,13 @@ public Int32 SelectedRad
         private readonly GsDumps _availableGsDumps;
         private readonly GsDlls _availableGsDlls;
 
+        private List<FileSystemWatcher> _dllWatcher;
```
willkuer: I would dispose there on form dispose as you did on each reload.

tools/GSDumpGUI/Library/GSDXWrapper.cs
```diff
@@ -39,14 +39,15 @@ namespace GSDumpGUI
     public delegate void GSreset();
     public delegate void GSreadFIFO2(IntPtr data, int size);
     public delegate void GSsetGameCRC(int crc, int options);
-    public delegate int GSfreeze(int mode, IntPtr data);
-    public delegate void GSopen(IntPtr hwnd, String Title, int renderer);
+    public delegate  int GSfreeze(int mode, IntPtr data);
+    public delegate  int GSopen(IntPtr hwnd, String Title, int renderer);
```
willkuer: Shouldn’t you then align on the longest type? :D

tools/GSDumpGUI/Library/GSDXWrapper.cs
```diff
@@ -205,115 +205,135 @@ public unsafe void Run(GSDump dump, int rendererOverride)
...
+                NativeMethods.SetClassLong(hWnd,/*GCL_HICON*/ -14, Program.hMainIcon);
+
                 fixed (byte* freeze = dump.StateData)
                 {
-                    byte[] GSFreez = new byte[8];
-                    Array.Copy(BitConverter.GetBytes(dump.StateData.Length), 0, GSFreez, 0, 4);
-                    Array.Copy(BitConverter.GetBytes(new IntPtr(freeze).ToInt32()), 0, GSFreez, 4, 4);
+                    byte[] GSFreez;
+
+                    if (IntPtr.Size > 4)
```
willkuer: Here I would use Environment.is64bit to make it readable for people after you

tools/GSDumpGUI/Library/GSDXWrapper.cs
```diff
@@ -205,115 +205,135 @@ public unsafe void Run(GSDump dump, int rendererOverride)
...
                         if (ris == -1)
                         {
                             DumpTooOld = true;
-                            return;
+                            Running = false;
                         }
                         GSVSync(1);
 
+                        GSreset();
+                        Marshal.Copy(dump.Registers, 0, new IntPtr(pointer), 8192);
```
willkuer: Magic numbers are dangerous. E.g. I don’t understand where it comes from if you do not even comment it. I guess const int descriptive_name = 8xxx; would be better

tools/GSDumpGUI/Library/GSDXWrapper.cs
```diff
@@ -205,115 +205,135 @@ public unsafe void Run(GSDump dump, int rendererOverride)
...
-                                            Program.Client.Send(Msg);
-
-                                            ExternalEvent.Set();
-                                        }
-                                        else
-                                        {
-                                            Step(itm, pointer);
-                                        }
+                                        case MessageType.Step:
+                                            if (debug_idx >= dump.Data.Count) debug_idx = 0;
```
willkuer: Also in all of these I’d say newlines improve readability

tools/GSDumpGUI/Library/GSDXWrapper.cs
```diff
@@ -100,60 +101,58 @@ public void Load(String DLL)
...
-                        this.GSshutdown = (GSshutdown) Marshal.GetDelegateForFunctionPointer(funcaddrShutdown, typeof(GSshutdown));
-                        this.GSfreeze = (GSfreeze) Marshal.GetDelegateForFunctionPointer(funcaddrFreeze, typeof(GSfreeze));
-                        this.GSreset = (GSreset) Marshal.GetDelegateForFunctionPointer(funcaddrGSReset, typeof(GSreset));
-                        this.GSreadFIFO2 = (GSreadFIFO2) Marshal.GetDelegateForFunctionPointer(funcaddrGSreadFIFO2, typeof(GSreadFIFO2));
-                        this.GSinit = (GSinit) Marshal.GetDelegateForFunctionPointer(funcaddrinit, typeof(GSinit));
-
-                        Loaded = true;
-                    }
+                Directory.SetCurrentDirectory(dir);
+                System.Diagnostics.Trace.WriteLine("LoadLibrary: " + DLL);
```
willkuer: I actually have some locally code to transfer logging through the tcp client into the server. Is that using the stderr/stdout of the console to get it back into the form?

tools/GSDumpGUI/Library/NativeMethods.cs
```diff
@@ -33,55 +33,66 @@ namespace GSDumpGUI
...
-        public static extern bool DispatchMessage(ref NativeMessage message);
+        public static extern bool TranslateMessage(ref NativeMessage lpMsg);
+
+        [SuppressUnmanagedCodeSecurityAttribute]
+        [DllImport("user32.dll", CharSet = CharSet.Auto)]
+        public static extern UInt32 DispatchMessage(ref NativeMessage lpMsg);
+
+        public static UIntPtr SetClassLong(IntPtr hWnd, Int32 index, IntPtr dwNewLong)
+        {
+            if (IntPtr.Size > 4) return SetClassLong64(hWnd, index, dwNewLong);
```
willkuer: Here again I would use environment.is64bit to make the intention clear

willkuer:
Great work! I would’ve done stuff differently at some points but I think these are cool additions. I would just consider to release unmanaged resources in dispose even if not necessary because of app shutdown.

I was only able to do reviewing on my mobile so sorry if I missed something or my writing sucks.

KrossX:
Thank you for your time! Much appreciated.

tools/GSDumpGUI/Forms/frmMain.cs
KrossX: I removed them because something was bringing the old path back after it was changed. Never happened again after I removed that. I could try adding it again and debug elsewhere.

tools/GSDumpGUI/Forms/frmMain.cs
KrossX: `System.ComponentModel.BackgroundWorker` it's apparently available, but it seems to be to fire a single task in the background and be done with it.

I made the thread because the filewatchers would spam with reloads. The thread kinda works as a way to reduce the spam and just run one reload per many requests. So it's the message consumer thread for the multiple producers. Would BackgroundWorker work for this?

tools/GSDumpGUI/Forms/frmMain.cs
KrossX: This one ? [System.Threading.Timer](https://docs.microsoft.com/en-us/dotnet/api/system.threading.timer) looks interesting. I'll give it a try.

tools/GSDumpGUI/Forms/frmMain.cs
KrossX: It was like that before though: [frmMain.cs@L147](https://github.com/PCSX2/pcsx2/blob/0814690534e585bf4e74da131d439ba2e087080f/tools/GSDumpGUI/Forms/frmMain.cs#L147)
`var gsdxFolder = _folderWithFallBackFinder.GetViaPatternWithFallback(Settings.GSDXDir, "*.dll", "", "plugins", "dll", "dlls");`
It checks for the selected path, which was and still is customizable, then also searches in those other places. I suppose it was a way to have dlls and dumps listed with a default path.

Another intersting option would be to have a list of paths exposed in the UI. Extra paths could then be added or removed from it.

tools/GSDumpGUI/Forms/frmMain.cs
KrossX: I thought it might just spam a few "path does not exists" every time.

tools/GSDumpGUI/Library/GSDXWrapper.cs
KrossX: I have absolutely no clue where that's from, it's old code. I can try to find out and put a comment.

tools/GSDumpGUI/Library/GSDXWrapper.cs
KrossX: Debug leftover, it was just to see the number of Load/Free on the debugger output window.

tools/GSDumpGUI/Library/GSDXWrapper.cs
KrossX: Done!

tools/GSDumpGUI/Library/GSDXWrapper.cs
KrossX: Is the new alignment better?

tools/GSDumpGUI/Forms/frmMain.cs
KrossX: Done!

tools/GSDumpGUI/Forms/frmMain.cs
KrossX: Done!

tools/GSDumpGUI/Forms/frmMain.cs
KrossX: Done!

tools/GSDumpGUI/Library/NativeMethods.cs
KrossX: Done!

tools/GSDumpGUI/Library/GSDXWrapper.cs
KrossX: Done! I think.

tools/GSDumpGUI/Library/GSDXWrapper.cs
willkuer: No. I thought you added another magic number. I was once considering to add 64bit compatibility without knowing anything about low level programming and basically stopped when I saw all these multiples of 4 going through the code without a single notification.

tools/GSDumpGUI/Forms/frmMain.cs
willkuer: Yes model binding sucks heavily in winforms. At least for this framework version. Just keep it removed. I wanted to relaunch the app in an MVC way anyways at one magical day in future.

tools/GSDumpGUI/Forms/frmMain.cs
willkuer: It’s basically the same as thread but with lower cpu prio and some special long running thread pool settings iirc. But a timer would be better.

tools/GSDumpGUI/Forms/frmMain.cs
willkuer: No better use the winforms timer to stay in the ui thread. You need to check whether it is disposable though and possibly dispose it as well on form dispose.

willkuer: https://docs.microsoft.com/de-de/dotnet/api/system.windows.forms.timer?view=netframework-4.8

tools/GSDumpGUI/Forms/frmMain.cs
willkuer: Ah yeah I was writing that code. See how easy I forget stuff :D

tools/GSDumpGUI/Forms/frmMain.cs
willkuer: Ah I missed that. In that case you’d typically first check for FileExists. Exceptions are a bad program flow (in the net world) and you should avoid them as much as possible. Although I don’t see how it could really introduce issues here.

tools/GSDumpGUI/Forms/frmMain.cs
willkuer: So the you could remove the this.Invoke and just call the update functions.

KrossX: Done! It seeems to be working alright.

tools/GSDumpGUI/Forms/frmMain.cs
KrossX: Done!

tools/GSDumpGUI/Forms/frmMain.cs
KrossX: Done!

lightningterror:
@willkuer is it ready for merge or you still have pending review/changes?

lightningterror:
I will merge this now, any other potential improvements can be done at a later pr.
Current PR is already big with many fixes/improvements :) Good job.

PR #3050, opened 2019-08-01
Move PSXCLK and setpsxspeed to proper reset function

kenshen112:
https://www.youtube.com/watch?v=pQzCptidLmw - This PR in a nutshell… It moves some lines of code for ps1 mode to a more correct spot. Also rama should the PCSX2 GO DISC SPEED stuff in CDVD cdvdWrite14 go in CdRom.cpp setpsxmode()?

### DIFF
--- a/pcsx2/R3000A.cpp
+++ b/pcsx2/R3000A.cpp
@@ -63,7 +63,8 @@ void psxReset()
 	g_iopNextEventCycle = psxRegs.cycle + 4;
 
 	psxHwReset();
-	PSXCLK = 36864000;
+	PSXCLK =  33868800;
+	setPsxSpeed();
 	ioman::reset();
 	psxBiosReset();
 }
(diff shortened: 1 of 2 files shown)


pcsx2/R3000A.cpp
```diff
@@ -63,7 +63,8 @@ void psxReset()
 	g_iopNextEventCycle = psxRegs.cycle + 4;
 
 	psxHwReset();
-	PSXCLK = 36864000;
+	PSXCLK =  33868800;
```
turtleli: Doesn't this change mean that the IOP runs too slowly when playing PS2 games?

kenshen112: This is specifically for ps1mode in CDVD.cpp the ps1mode reset this is specifically called

turtleli: I'm pretty sure that this'll affect both PS2 and PSX mode.

kenshen112: I just tested it with rachet and clank. It doesn't

turtleli: Set a breakpoint here in the debugger and see it if triggers when you run a PS2 game. Because it does on my end (even boot to BIOS triggers this code).

ramapcsx2:
There were more considerations with switching to PSX mode than just where to set the clocks.
Please leave the reset logic alone for now.

pcsx2/R3000A.cpp
refractionpcsx2: psxReset gets called when the PS2 side resets, it was 36Mhz for a reason.

PR #3051, opened 2019-08-01
cdvd: Fix end-of-disc issues

### LINKED ISSUE #2113: Ratchet and Clank 3 : Read error.
***PCSX2 version:***
Last 1.5 dev

***PCSX2 options:***
All default, speedhacks disabled except MTVU

***Plugins used:***
All default

Ratchet and Clank 3 always have a read error in the 5th Qwark video comic level, it cause the music to stop and a retry impossible.

***Note:***
Set the EEcyclerate to -3 don't solve the issue.

Load PCSX2 and Ratchet and Clank 3, go to the 5th Qwark video comic level and you will see the error aleatory in it.

***Blockdump:***
https://drive.google.com/file/d/0B0Ph0W1NYxIVQk9pT0xpTGMyTm8/view?usp=sharing

Not a regression, 1.2.1 and 1.4 are also affected.

***PC specifications:***
AMD fx-8350 GTX 780ti Windows 10 64 Bit

turtleli:
Changes:
 * Fix a few off by one errors.
 * Fix loading for some games that attempt to read past the end of disc by doing nothing instead of erroring out. Seems to fix All Star Baseball 2005. cdvdGigaherz change isn't tested since I only have a blockdump.

Fixes #2113.

### DIFF
--- a/pcsx2/CDVD/CDVDaccess.cpp
+++ b/pcsx2/CDVD/CDVDaccess.cpp
@@ -56,7 +56,7 @@ static int diskTypeCached = -1;
 
 // used to bridge the gap between the old getBuffer api and the new getBuffer2 api.
 int lastReadSize;
-int lastLSN;		// needed for block dumping
+u32 lastLSN;		// needed for block dumping
 
 // Records last read block length for block dumping
 //static int plsn = 0;
@@ -490,6 +490,12 @@ s32 DoCDVDgetBuffer(u8* buffer)
 
 	if (ret == 0 && blockDumpFile.IsOpened())
 	{
+		cdvdTD td;
+		CDVD->getTD(0, &td);
+
+		if (lastLSN >= td.lsn)
+			return 0;
+
 		if (blockDumpFile.GetBlockSize() == CD_FRAMESIZE_RAW && lastReadSize != 2352)
 		{
 			u8 blockDumpBuffer[CD_FRAMESIZE_RAW];
--- a/pcsx2/CDVD/CDVDisoReader.cpp
+++ b/pcsx2/CDVD/CDVDisoReader.cpp
@@ -362,7 +362,6 @@ s32 CALLBACK ISOreadTrack(u32 lsn, int mode)
 	int _lsn = lsn;
 
 	if (_lsn < 0) lsn = iso.GetBlockCount() + _lsn;
-	if (lsn >= iso.GetBlockCount()) return -1;
 
 	iso.BeginRead2(lsn);
 
--- a/pcsx2/CDVD/CDVDisoReader.cpp
+++ b/pcsx2/CDVD/CDVDisoReader.cpp
@@ -315,7 +315,7 @@ s32 CALLBACK ISOreadSector(u8* tempbuffer, u32 lsn, int mode)
 	int _lsn = lsn;
 
 	if (_lsn < 0) lsn = iso.GetBlockCount() + _lsn;
-	if (lsn > iso.GetBlockCount()) return -1;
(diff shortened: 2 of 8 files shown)


turtleli:
@prafullpcsx2 Can you test Test Drive - Eve of Destruction? The blockdump you sent me ages ago doesn't seem to work for me so I don't know if it's the same issue as All Star Baseball.

ramapcsx2:
@turtleli
Where in All Star Baseball does it attempt those reads?   
I don't think we know how real hardware reacts to that.

turtleli:
It attempts those reads almost immediately after loading the game.

prafullpcsx2:
Thanks. Eve of destruction is now fixed both in Internal iso and cdvd gigaherz.

ramapcsx2:
Wow, that's a lot of released games with a pretty bad problem. 
You'd think quality assurance would catch something like this? Dunno, seems weird..

Anyway, good work! :)

ramapcsx2:
@prafullpcsx2 
Do you have real hardware at hand?
If so, could you pop one of these games in and see if they do anything weird with the drive?

I'm not sure how the issue would be handled.
It probably either silently ignores the bad request, or it may actually move the laser to the max position :p 
Or well, maybe it'll start to move the laser, then stop at the max indication from the TOC?

prafullpcsx2:
@ramapcsx2 I tried one of these games on my PS2 but doesn't seem to be doing anything weird.

ramapcsx2:
Okay, thanks for testing :)

atomic83GitHub:
So, when the console show this, the game does read past the end of disc. If it's the case that fix simpsons the game from spamming CDVD errors.

image

atomic83GitHub:
That PR make the 5th videofun playable in Ratchet up your arsenal: https://github.com/PCSX2/pcsx2/issues/2113

turtleli:
Updated:
 * Improved commit descriptions and comments - additional game names were added to the comments (in case of future regressions), and past end of disc was replaced with out of bounds/non-existent (judging from the log in the related issue, R&C 3 seems to reference a sector in the middle of nowhere rather than just slightly past the end of disc, so it's probably a more fitting description).
 * Avoid writing non-existent sectors to the blockdump.

> So, when the console show this, the game does read past the end of disc?

Yes.

atomic83GitHub:
That's trivial but you forgot "The Simpsons game".

turtleli:
I intentionally didn't include it. If it worked before this PR then this PR hasn't fixed it.

I probably won't be adding any more games to the code comments since the list doesn't need to be comprehensive, but do comment if this PR fixes any game which hasn't yet been mentioned.

gregory38:
Welcome  back. Good job :+1:

ramapcsx2:
Tested this with some isos and directly from DVD, all is fine.
It also compiles with VS2019 in release and debug.

I didn't have any of the problem games to test, but I could provoke some error conditions by swapping DVDs on the fly and such, no issues there either.

So I'm good with merging :)

MrCK1:
With this PR, I'm pretty sure there's no other game that relies on Linuz or the other CDVD plugins then? If so, that's great news :)

prafullpcsx2:
I have tested this (internal iso version) with more than 100 games and didn't notice aby regressions. This should be merged before 1.6 release.

ramapcsx2:
Agreed. I'll do it now so it doesn't get lost.

lightningterror:
I think a rebase merge would've been better instead of squash.

turtleli:
Yeah, rebase would have been nicer. Was actually planning on merging later today.

PR #3052, opened 2019-08-01
PCSX2: Wait some frames between F9 SW render toggles

### LINKED ISSUE #370: add a minimum delay between consecutive F9
Quickly switching between hardware/software renderer is not very stable. The idea is to put a minimum delay of a couple of seconds between consecutive hit.

Some references:

http://forums.pcsx2.net/Thread-Bug-Report-Project-Zero-2-Fatal-Frame-2-PAL?pid=380261#pid380261

http://forums.pcsx2.net/Thread-Bug-Report-Shin-Megami-Tensei-Persona-4-NTSC-SLUS-21782?pid=380265#pid380265

KrossX:
Attempt to fix issue #370.

Setting the toggle variable is now at the vsync point. FMV toggle should still work as usual, but the F9 toggle should have to wait the full period of around 32 frames. This seems short enough while also depends on how fast those frames can rendered, avoiding having to set a timer in ms.

Fixes #370

### DIFF
--- a/pcsx2/gui/AppMain.cpp
+++ b/pcsx2/gui/AppMain.cpp
@@ -517,8 +517,9 @@ double FramerateManager::GetFramerate() const
 // times a second if not (ok, not quite, but you get the idea... I hope.)
 extern uint eecount_on_last_vdec;
 extern bool FMVstarted;
-extern bool renderswitch;
 extern bool EnableFMV;
+extern bool renderswitch;
+extern uint do_renderswitch;
 
 void DoFmvSwitch(bool on)
 {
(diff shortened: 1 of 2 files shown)


pcsx2/gui/AppMain.cpp
```diff
@@ -534,11 +535,7 @@ void DoFmvSwitch(bool on)
 				viewport->DoResize();
 	}
 
-	if (EmuConfig.Gamefixes.FMVinSoftwareHack) {
```
willkuer: Isn’t this if missing now?

KrossX: Oh yeah, that should still be there. I'll fix it in a few....

lightningterror:
Can confirm it fixes the issue, pcsx2 no longer hangs if you spam F9.

CookiePLMonster:
It'd probably be wise to squash this to a single commit so c74d0be2709e77ada2bd11206a765bd83605bd61 doesn't technically introduce a regression.

KrossX:
Done.

ramapcsx2:
This should also go into 1.6.
Funny that these kind of bugfixes always pile up in feature freezes :p 

But yeah, this is a good for reliability and should have always been done like this anyway.

lightningterror:
Better late than never I suppose, gotta have all those juicy bugfixes before release :)
Is the PR all good for merge?

lightningterror:
Ok I will merge this, everyone seems to be ok with it and testing shows it works good.

PR #3053, opened 2019-08-02
PAD: Delay load dependent DLLs

CookiePLMonster:
This PR "fixes" a possible crash when enumerating plugins. As LilyPad depends on dinput8.dll, it was loaded and unloaded together with LilyPad, however it seems that DInput8 on Windows 10 regressed and now such rapid load/unload can cause it to crash.

With this PR, dinput8.dll and other dependant DLLs are delay loaded, so they are loaded before they are **actually** needed. Only dinput8.dll required this to work around the crash, but I also included the others (hid.dll and setupapi.dll), since it looks like it's totally possible to go through an entire session without using their functions. This should also speed up enumeration and startup, since less DLLs load at these points now.

Overview of DInput8 bug which could have resulted in a crash when enumerating LilyPad:
https://cookieplmonster.github.io/2019/08/02/win10-dinput8-crash/

### DIFF
--- a/plugins/LilyPad/LilyPad.cpp
+++ b/plugins/LilyPad/LilyPad.cpp
@@ -330,17 +330,19 @@ void UpdateEnabledDevices(int updateList = 0)
 #ifdef _MSC_VER
 BOOL WINAPI DllMain(HINSTANCE hInstance, DWORD fdwReason, void *lpvReserved)
 {
-    hInst = hInstance;
     if (fdwReason == DLL_PROCESS_ATTACH) {
+        hInst = hInstance;
         InitializeCriticalSection(&updateLock);
 
         DisableThreadLibraryCalls(hInstance);
     } else if (fdwReason == DLL_PROCESS_DETACH) {
-        while (openCount)
-            PADclose();
-        PADshutdown();
-        UninitLibUsb();
-        DeleteCriticalSection(&updateLock);
+        if (lpvReserved == nullptr) {
+            while (openCount)
+                PADclose();
+            PADshutdown();
+            UninitLibUsb();
+            DeleteCriticalSection(&updateLock);
+        }
     }
     return 1;
 }
@@ -1183,7 +1185,7 @@ u8 CALLBACK PADpoll(u8 value)
                     return 0xF3;
                 }
                 // Fall through
-                
+
             // READ_DATA_AND_VIBRATE
             case 0x42:
                 query.response[2] = 0x5A;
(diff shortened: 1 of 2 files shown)


willkuer:
I just read a blog entry that such delayed loading can lead to unexpected behavior on resource release/dll unloading/app shutdown as the order of dll unloading is not known anymore to the app and hence all dll’s will be unloaded in arbitrary order and if one dll holds a reference to another dll’s content it will break.

Might occur when playing a game (hence loading all dlls) and then switching to another plugin dll or maybe even closing pcsx2.

I will check whether I can find the blog entry. Maybe it’s unrelated...

willkuer:
https://devblogs.microsoft.com/oldnewthing/20190718-00/?p=102719

CookiePLMonster:
You're absolutely right, I forgot about the shutdown sequences - although the article you linked, [and the one it links to](https://devblogs.microsoft.com/oldnewthing/20120105-00/?p=8683) provides answers for your concerns and also a proper way for me to ensure nothing bad happens:

> **Note that this problem occurs only at process shutdown.** If CONTOSO.DLL unloads via a runtime call to Free­Library, it will still be able to call into WIDGET.DLL because it hasn’t yet called Free­Library on WIDGET.DLL. But during process shutdown, the module loader needs to free all the things, and the outstanding Load­Library won’t prevent that from happening.

So basically, delay loading will not cause problems when switching plugins (since LilyPad unloads via `FreeLibrary`) but it might cause problems on terminating PCSX2. But then:

> **The solution is to bypass widget cleanup if the DLL_PROCESS_DETACH handler realizes that the process is terminating.** Just leak the widget. The building is being demolished. You don’t need to sweep the floors.

And that part is trivial - I'll edit the PR in a bit.

CookiePLMonster:
So I inspected the code again and even when terminating PCSX2, plugins are unloaded via `FreeLibrary`, so it's safe for delay loading. **That said**, I guarded process detach routines against attempting to free again if the process is terminating, as per MSFT guidelines. This should pre-emptively solve any issues with this.

MrCK1:
@CookiePLMonster Anything left to be done on this besides additional testing?

CookiePLMonster:
Nope, it's done and ready for review and testing.

MrCK1:
@CookiePLMonster sorry for the delay. In the meantime, would you be able to rebase? Hopefully this can get merged soon.

CookiePLMonster:
Rebased - looks like it's went through without error.

CookiePLMonster:
I tried to remake this PR for a now-merged PAD but I don't think it applies anymore. The root issue which made me submit this PR (crash when enumerating plugins) isn't an issue anymore, so I think this can go.

PR #3056, opened 2019-08-03
3rdparty/pthreads4w: Fixup process exit logic

CookiePLMonster:
This PR modifies a **third party module** - I realize normally those should be fixed upstream, but I **really** doubt pthreads4w would a) consider it a valid bug and b) fix it. To make all my changes easily visible, I wrap them all in `#if PCSX2_FIX`.

This PR fixes a process exit routine in pthreads. This third party module exploits CRT initialization order to inject their initializer/deinitializer earlier than the others by putting their functions in `.CRT$XCU` and `.CRT$XPU` pseudo regions. The problem comes when a module gets build with dynamic CRT (`/MD` or `/MDd`), like most of PCSX2 plugins, it doesn't actually use `.CRT$XPx` regions as terminators, and instead lets dynamic CRT handle them.

This PR corrects this issue by registering the terminator via `atexit`, so it works with both static and dynamic CRT. This resolves an issue where SPU2-X plugin (and potentially more) leaks TLS handles when unloaded.

### DIFF
--- a/3rdparty/pthreads4w/autostatic.c
+++ b/3rdparty/pthreads4w/autostatic.c
@@ -36,21 +36,33 @@
 
 #if defined(PTW32_STATIC_LIB)
 
+#define PCSX2_FIX 1 // Comment out to restore code to pristine state
+
 #if defined(__MINGW64__) || defined(__MINGW32__) || defined(_MSC_VER)
 
 #include "pthread.h"
 #include "implement.h"
 
+#if !PCSX2_FIX
 static void on_process_init(void)
 {
     pthread_win32_process_attach_np ();
 }
+#endif
 
 static void on_process_exit(void)
 {
     pthread_win32_thread_detach_np  ();
     pthread_win32_process_detach_np ();
 }
+
+#if PCSX2_FIX
+static void on_process_init(void)
+{
+    pthread_win32_process_attach_np();
+    atexit(on_process_exit);
+}
+#endif
 
 #if defined(__MINGW64__) || defined(__MINGW32__)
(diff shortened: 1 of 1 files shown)


lightningterror:
I still think it is worth also pring this in upstream too.

CookiePLMonster:
> I still think it is worth also pring this in upstream too.

Do you know where exactly this project is maintained?

tadanokojin:
Believe here: https://sourceforge.net/projects/pthreads4w/

CookiePLMonster:
It doesn't seem to have `autostatic.c` though.

tadanokojin:
Hmmm it's in 2.9.1 release. I wonder if they fixed it in a later version. They seemed to have removed the files.

CookiePLMonster:
Do you think upgrading it is viable? Maybe this issue is not even a thing in newer versions. How old is the version used by PCSX2?

tadanokojin:
Pretty old. We considered it but I think the consensus was we should move to `std::thread`. We're in a freeze right now so update might not be a good idea. I think this fix is fine.

CookiePLMonster:
That sounds reasonable yeah, let's stick to it then.

ramapcsx2:
This fix (and the cdvd off by one / out of bounds one) should be in 1.6 though, right?

CookiePLMonster:
I'm not sure, it feels like it will be yes but can't say for certain.

tadanokojin:
This has my approval for 1.6.0

lightningterror:
Gonna go ahead and merge this then, time to move along with the 1.6 progress.

PR #3059, opened 2019-08-04
Various: Adjust gui options

lightningterror:
**GSdx-gui:**
Rename Automatic CRC Level label to CRC Lookup Level.
Rename Accurate Date label to DATE Accuracy.
Rename Accurate Blending label to Blending Accuracy.

Remove 10x and 12x upscaling options, we don't want to hit the maximum
texture size limit and cause issues so better play it safe and allow
only up until 8x upscaling.

Reword the tooltip to make it a bit more fitting with it's current
behavior.
Add all games in the list that have some sort of aggressive state.
Update anisotropic filtering, autoflush sw, sparse texture, gl blending tooltips.

Redesign hw hacks gui on Windows.
Separate and group hack types in to Rendering and Upscaling hacks. This
should help users identify better what some hacks do and make it a
little bit easier to use.

**PCSX2-gui:**
Update About window.
Add forum and license links.
Remove Developers and Contributors names, replaced with:
"Big thanks to everyone who contributed to the project throughout the years."
Fix DPI scaling for about window, first time wizard,import settings, assertion dialog.
Fix DPI scaling for about window logo, also dynamically allocate the wxStaticBitmap control since its parent window will delete it, which can cause issues.

**Lilypad-gui:**
Fix gui overlapping for force feedback axis 2 value.


ssakash:
> Automatic CRC Level -> CRC Level

This should be CRC hack level, CRC level isn't clear enough. Is this all about getting rid of the word "hack" to make the users get a different impression of that option?  xD


>Accurate Date -> DATE Accuracy
>Accurate Blending -> Blending Accuracy

I don't think it's worth changing either, the old ones were clear enough.

lightningterror:
> This should be CRC hack level, CRC level isn't clear enough. Is this all about getting rid of the word "hack" to make the users get a different impression of that option? xD

Thing is we have plenty of rendering fixes that rely on crc lookup too so calling it a hack isn't right either. I'm out of ideas what alternative names we can call it.

ssakash:
> Thing is we have plenty of rendering fixes that rely on crc lookup too so calling it a hack isn't right either. 

Anything which relies on detecting if it's a specific game is a hack.

lightningterror:
I updated the pr with only crc label change and removal of 10x and 12x as it was discussed on discord for the removal of high upscaling which can cause issues.

turtleli:
Personally I prefer DATE Accuracy and Blending Accuracy. IMO Accurate DATE and Accurate Blending sound more like on/off things, whereas DATE Accuracy and Blending Accuracy sound like they have various levels (and they do).

tadanokojin:
I'm in agreement with Li here.

lightningterror:
I added a commit for the about window too, it is wip so it's missing some names in there.

lightningterror:
Here's how the about page looks like now
image

pcsx2/gui/Dialogs/AboutBoxDialog.cpp
```diff
@@ -37,69 +37,53 @@ Dialogs::AboutBoxDialog::AboutBoxDialog(wxWindow* parent)
...
+		_("Support Forums"), L"https://forums.pcsx2.net"
+	) | pxProportion(1).Center().Border(wxALL, 3);
+
+	*this += new wxHyperlinkCtrl(this, wxID_ANY,
+		_("Git Repository at GitHub"), L"https://github.com/PCSX2/pcsx2"
 		) | pxProportion(1).Center().Border(wxALL, 3);
 
-	*this += boxesContainer | StdCenter();
+	*this += new wxHyperlinkCtrl(this, wxID_ANY,
+		_("Contributors"), L"https://github.com/PCSX2/pcsx2/graphs/contributors"
```
ssakash: This ignores the old Google code contributors though.

lightningterror: How about something like this?
image

Link1 links to github, link2 to https://pcsx2.net/meet-the-team.html

ssakash: Looks really weird to be honest, why not just have a singular link to OpenHub instead which saves all contributor details?

MrCK1: I'm not a fan of "Link1" and "Link2" either. Doesn't look very clean :P

lightningterror: We can have all this information on the website, contributors/developers haven't been updated since 1.2.1 release, maybe @bositman can cook something up that provides info for 1.2.1 and below releases (google code) and 1.4 and above github.
Feel free to leave suggestions.

lightningterror:
Need to update sparse texture tooltip.

pcsx2/gui/Dialogs/AboutBoxDialog.cpp
```diff
@@ -37,69 +37,55 @@ Dialogs::AboutBoxDialog::AboutBoxDialog(wxWindow* parent)
...
+	*this += new wxHyperlinkCtrl(this, wxID_ANY,
+		_("Support Forums"), L"https://forums.pcsx2.net"
+	) | pxProportion(1).Center().Border(wxALL, 3);
+
+	*this += new wxHyperlinkCtrl(this, wxID_ANY,
+		_("Git Repository at GitHub"), L"https://github.com/PCSX2/pcsx2"
+	) | pxProportion(1).Center().Border(wxALL, 3);
+
+	*this += new wxHyperlinkCtrl(this, wxID_ANY,
+		_("Contributors for v1.6 release"), L"https://github.com/PCSX2/pcsx2/graphs/contributors?from=2016-01-08&to=2019-11-07&type=c"
```
lightningterror: Link will need to be updated again during release date.

tadanokojin: Do we need to do by release? I think the point of this was to not have to update it all the time. Maybe just the non-date-ranged link and just call it "Contributors" or w/e.

lightningterror: https://github.com/PCSX2/pcsx2/pull/3059#discussion_r327926696

lightningterror: We do mention previous contributors tho.

tadanokojin: Sorry not following, what does using github date range have to do with google?

lightningterror: Nothing really, idea was just to list active contributors for 1.6 release, just trying various things to see which people prefer the most.

tadanokojin: I say just link to the contributor list. If you have to have an additional link or some 3rd party side for google people then so be it. I just think it defeats the purpose Li set out with this to not have to update it every release if we make it a range that has to be updated every release.

pcsx2/gui/Dialogs/AboutBoxDialog.cpp
tadanokojin: This it?
https://www.openhub.net/p/Pcsx2/contributors/summary

lightningterror: Nice, this lists everyone.

lightningterror:
About window now:
image

PR should be good to go now.

pcsx2/gui/Dialogs/AboutBoxDialog.cpp
```diff
@@ -37,69 +37,55 @@ Dialogs::AboutBoxDialog::AboutBoxDialog(wxWindow* parent)
...
+
+	SetMinWidth(460);
+
 	*this += m_bitmap_logo | StdCenter();
 
-	*this += Text(_("PlayStation 2 Emulator"));
+	*this += Text(_("PlayStation 2 Emulator:"));
+
+	*this += new wxHyperlinkCtrl(this, wxID_ANY,
+		_("Main Website"), L"https://pcsx2.net"
```
turtleli: Is `Main` necessary?

pcsx2/gui/Dialogs/AboutBoxDialog.cpp
```diff
@@ -37,69 +37,55 @@ Dialogs::AboutBoxDialog::AboutBoxDialog(wxWindow* parent)
...
+	*this += new wxHyperlinkCtrl(this, wxID_ANY,
+		_("GitHub Repository"), L"https://github.com/PCSX2/pcsx2"
+	) | pxProportion(1).Center().Border(wxALL, 3);
+
+	*this += new wxHyperlinkCtrl(this, wxID_ANY,
+		_("Contributors"), L"https://www.openhub.net/p/Pcsx2/contributors/summary"
+	) | pxProportion(1).Center().Border(wxALL, 3);
+
+	*this += new wxHyperlinkCtrl(this, wxID_ANY,
+		_("License"), L"https://github.com/PCSX2/pcsx2/blob/master/COPYING.GPLv2"
```
turtleli: Is the linked license correct?

pcsx2/gui/Dialogs/AboutBoxDialog.cpp
```diff
@@ -37,69 +37,55 @@ Dialogs::AboutBoxDialog::AboutBoxDialog(wxWindow* parent)
...
+	*this += new wxHyperlinkCtrl(this, wxID_ANY,
+		_("Support Forums"), L"https://forums.pcsx2.net"
+	) | pxProportion(1).Center().Border(wxALL, 3);
+
+	*this += new wxHyperlinkCtrl(this, wxID_ANY,
+		_("GitHub Repository"), L"https://github.com/PCSX2/pcsx2"
+	) | pxProportion(1).Center().Border(wxALL, 3);
+
+	*this += new wxHyperlinkCtrl(this, wxID_ANY,
+		_("Contributors"), L"https://www.openhub.net/p/Pcsx2/contributors/summary"
```
turtleli: I wouldn't include this because it only really values coding contributors when there's other forms of contributors. But that's just my opinion.

pcsx2/gui/Dialogs/AboutBoxDialog.cpp
tadanokojin: Yeah we can remove that.

pcsx2/gui/Dialogs/AboutBoxDialog.cpp
lightningterror: I used the linked one on the repo.

image

pcsx2/gui/Dialogs/AboutBoxDialog.cpp
lightningterror: We do have a thank you for every contributor here https://github.com/PCSX2/pcsx2/pull/3059/files#diff-7242f63bb6c1fff6ed237d24f219c4e8R86

I vote to keep some sort of link, don't know about others.

pcsx2/gui/Dialogs/AboutBoxDialog.cpp
turtleli: Did you look at the license on the actual files instead of using what Github thinks it is?

lightningterror: Ah, I thought linking the gnu license was fine, I linked the one from the docs now.

lightningterror:
I pushed a new commit that redesigns the gsdx hacks gui and separately groups rendering and upscaling hacks, Windows only.

image

lightningterror:
About window has been updated again, new style currently looks like this

Windows:
image

Linux Manjaro LXQT:
image

lightningterror:
Following a discussion new update contains social media links removed as well as contributors link.

New style on windows:
image

PR #3060, opened 2019-08-05
Remove pthread dependency from onepad_legacy.

arcum42:
I switched onepad legacy to use an mt_queue instead, which onepad already did, and moved mt_queue over to Utilities, which both avoids two copies of the same file and allows it to be used anywhere a multithreaded queue may come in handy.


plugins/onepad_legacy/onepad.cpp
```diff
@@ -223,10 +216,6 @@ EXPORT_C_(void) PADsetLogDir(const char *dir)
 
 EXPORT_C_(void) PADclose()
 {
-    while (!ev_fifo.empty())
```
CookiePLMonster: Doesn't the new type queue need to be flushed here too?

arcum42: You know, I was largely copying greg's code from onepad over to onepad legacy, and he didn't flush it there. Still, this is on closing the plugin, and I'm not sure it's absolutely necessary. 

Actually, I'm pretty sure if the plugin was closed, that it'd have to call PADopen if it started again, and that resets the queue, so we're probably fine there.

CookiePLMonster: I guess. Do note that plugin may unload after closing it, but then it should just run the destructor and free the queue anyway.

arcum42:
Think I'll go ahead and commit this, and if any issues crop up, we can always add in that flush and troubleshoot it at that point, if need be.

PR #3063, opened 2019-08-05
Avoid hang when spamming F4 (Frame Limiter Toggle)

KrossX:
Hitting F4 toggles the master framelimiter and the `AlsoApplyToGui` property causes `AppApplySettings()` to be called. This makes it unresponsive compared to turbo keys and locks the application.

Setting it to false, I assumed the use of F4 will not be reflected in the UI but if you open the settings panel it shows as expected. So no bad side effect, possibly.

### DIFF
--- a/pcsx2/gui/GlobalCommands.cpp
+++ b/pcsx2/gui/GlobalCommands.cpp
@@ -672,7 +672,7 @@ static const GlobalCommandDescriptor CommandDeclarations[] =
 		Implementations::Framelimiter_MasterToggle,
 		NULL,
 		NULL,
-		true,
+		false,
 	},
 
 	{	"GSwindow_CycleAspectRatio",


MrCK1:
Works fine for me compared to previous behavior from #3052. No longer hangs when using F9.

ssakash:
>  I assumed the use of F4 will not be reflected in the UI but if you open the settings panel it shows as expected

The global configuration object (``g_conf``) is already being updated at its caller function, which is the same thing as call to ``AppApplySettings()`` just that the latter does it for every variables instead of the one (Framelimiter in this case) which needs to be updated.

No wonder it crashes, I'd presume the overhead on calling that large ass routine is rather high. Similar treatment should also be done to ``GSwindow_CycleAspectRatio`` I think, no point in calling this function for modifying a single variable in global object.

ssakash:
Either way, this commit is totally fine, so merging it right now.

turtleli:
`true` means that when F4 is pressed the state change is immediately visible in the GUI.

Test case:
Boot game
Open Emulation Settings Dialog, GS Panel
Press F4 with game window in focus -> The Disable Framelimiting checkbox state will toggle.

With false the checkbox will not toggle.

ssakash:
@turtleli 

I did overlook that behavior change, however do we need to retain that behavior? I'd categorize it to be not useful considering the less percentile of users who would be relying on this behavior, I might be wrong too.

turtleli:
I don't know the answer to that. It might confuse users if the Emulation Settings dialog can desync with the actual settings, so the behaviour is useful.

lightningterror:
The gui gets updated if you close/reopen the emulator.
It would be nice to keep the previous behavior of the gui, but fixing the hang takes priority.

KrossX:
What happens if you just reopen the settings dialog? If it works, it would only be a problem in case of having the settings dialog opened while using F4 on the active window to change the limiter.

lightningterror:
Kinda slipped my mind, yea it does update.

ssakash:
I personally like having this behavior on aspect ratio cycling, but not sure if it's worth it for Framelimiter toggle. You can always like cycle it back and it'll get saved when you reopen the dialog and the live feedback of the option status doesn't seem as important.

ssakash:
>   it would only be a problem in case of having the settings dialog opened while using F4 on the active window to change the limiter.

this is the case @turtleli  is concerned about as there's a possibility for desync in the hotkey settings vs GUI setting.

KrossX:
So, a slow change + potential lock/hang vs possible confusion in case of user using hotkeys while config dialog is opened. I think the latter is preferable.

turtleli:
The main issue IMO is that we're not fixing the underlying threading issue. Instead we're just doing stuff to avoid triggering it.

If the threading issues didn't exist then having the GUI immediately update would be preferable.

PR #3064, opened 2019-08-05
GSdx SW: Fix SSE2 mipmap code generation.

KrossX:
~~It seems SEL.LTF was beign set to the mipmap filter while SEL.MMIN was being set to the texture filter. Setting LTF to the first bit and MMIN to the last one instead, fixes Hot Pursuit 2 mipmap on SW.~~ That was just fine.

Problem seems to be a typo in a duplicated code of the SSE2 code generator part.

Before:
sw_before

After:
sw_after2

### DIFF
--- a/plugins/GSdx/Renderers/SW/GSDrawScanlineCodeGenerator.x86.cpp
+++ b/plugins/GSdx/Renderers/SW/GSDrawScanlineCodeGenerator.x86.cpp
@@ -1765,7 +1765,7 @@ void GSDrawScanlineCodeGenerator::SampleTextureLOD_SSE()
 			// c[0] = c00 & mask;
 			// c[1] = (c00 >> 8) & mask;
 
-			split16_2x8(xmm5, xmm6, xmm5);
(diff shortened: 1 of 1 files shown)


KrossX:
This adds banding to at least Jak and Daxter and Ape Escape. *sighs*
Closing until new fix.

bositman:
Wow nice catch!

ramapcsx2:
Weird though, if you look at the other uses of split16_2x8, they follow a different pattern.

KrossX:
That's quite possibly why the typo was made. The mipmap 0 code is here: [L1530](https://github.com/PCSX2/pcsx2/blob/fbf7be2a85d57de4f1ca19da457f4a9398c6f532/plugins/GSdx/Renderers/SW/GSDrawScanlineCodeGenerator.x86.cpp#L1530), the typo is in mipmap 1 code.

gregory38:
It was me that I created the split function. You can check the git history to see previous code. Maybe I introduced the regression.

Issue #1189, opened 2016-02-18
[GSdx] Auto-switching to GSdx software rendering (manual game fix) with Rogue Galaxy sometimes make PCSX2 freeze


piorrro33:
PCSX2 1.5.0-20160215202751 - compiled on Feb 15 2016
Emulation options: safest except MTVU and the game fix enabled
Plugins: http://a.pomf.cat/cfrtwr.png
GSdx settings: http://a.pomf.cat/fslxnj.png
SPU2-X settings: http://a.pomf.cat/tiuswt.png
Last known version to work: I think it never worked.
PC specs: Intel Core i7-4790 @ 3.60 GHz (working with turbo mode at 4 GHz); ASUS GeForce GTX 750 Ti 2GB
Game: Rogue Galaxy PAL SCES-54552

How to reproduce:
0) Enable this fix http://a.pomf.cat/iaakxb.png
1) Go into the Dorgenark, and travel to any planet (works too with any cutscene that toggles the SW mode)
2) Press Start then triangle to skip the cutscene
3) If you're too fast, it won't freeze, too slow, you'll get a black screen since it switched to HW but when you press start it'll be back working since it switches back to SW
4) Enjoy the freeze if you're between too fast and too slow.

Should also crash with D3D9 and OGL, but only tried with D3D11. (if you want me to try, tell me)
What I think is causing this problem, is that it tries to switch to the hardware renderer when you're pausing but since it doesn't have enough time to switch back to software before the cutscene gets skipped, it gets skipped in hardware mode which causes a freeze. I'm not sure it is that, though.

Side note: enabling this fix is MANDATORY (or you're forced to only play with SW), since if you don't, you can get missing text after flashback cutscenes (which are fucked in HW) and other bad stuff.

I hope it was complete enough, but if you need anything else, feel free to ask me.

EDIT: It seems like it doesn't freeze with the OGL renderer, since it doesn't pause the emulation when pressing the pause button, unlike D3D9 or D3D11.

EDIT N°2: If you don't get it to crash the first time, try, try again. It's very very random so sometimes you freeze but thankfully most of the time you don't.

MrCK1:
According to your description, it seems clear that the game is intended to be played in SW mode. 

Manual game fixes are disabled by default for a reason because they often cause more compatibility problems.

Otherwise, it could be a plugin issue related to layer searching. Rogue Galaxy has arbitrary layer sizes; so the search must start from an arbitrary position

[...]

PR #3072, opened 2019-08-14
PCSX2: Avoid hang when switching renders with hotkey.

KrossX:
This fixes a hang (deadlock?) when using the F9 key on certain situations. For example, going from SW render to HW during the codec scenes in MGS3.

However, the FMV software toggle still runs the code from the same place. This might lead to the same problem when using that option.

### DIFF
--- a/pcsx2/gui/AppMain.cpp
+++ b/pcsx2/gui/AppMain.cpp
@@ -519,7 +519,7 @@ extern uint eecount_on_last_vdec;
 extern bool FMVstarted;
 extern bool EnableFMV;
 extern bool renderswitch;
-extern uint do_renderswitch;
+extern uint renderswitch_delay;
 
 void DoFmvSwitch(bool on)
 {
@@ -535,8 +535,11 @@ void DoFmvSwitch(bool on)
 				viewport->DoResize();
 	}
 
-	if (EmuConfig.Gamefixes.FMVinSoftwareHack)
-		do_renderswitch = -1;
+	if (EmuConfig.Gamefixes.FMVinSoftwareHack) {
+		ScopedCoreThreadPause paused_core(new SysExecEvent_SaveSinglePlugin(PluginId_GS));
+		renderswitch = !renderswitch;
+		paused_core.AllowResume();
+	}
 }
 
 void Pcsx2App::LogicalVsync()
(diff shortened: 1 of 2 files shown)


ramapcsx2:
So what's going on there? Race condition?

KrossX:
I have absolutely no idea. When I tried to debug it, all I could see was a bunch of threads at WaitingForMultipleObjects.

ramapcsx2:
Well, it's hard to judge whether the solution is any good or safe when there's so little to go by.
Could you give some testing instructions? Maybe something that normally crashes / deadlocks in other situations than the MGS3 scene?

I'd just like to do a little testing, particularly how it used to behave and how it behaves with the patch, making sure there's no regression at least.

KrossX:
For testing current master, just start MGS3 and press start/circle enough to reach the first codec sequence (good point for savestate). Then, simply press F9, it hangs when going from SW to HW.

I simple moved the `ScopedCoreThreadPause` code back to where it was, you can check PR #3052 to compare the original code. Which is why I think the FMV hack was and still is prone to the same hang.

ramapcsx2:
Okay, so if I understand that correctly, the logical VSync is now counting down the renderer switch delay and not doing the switch directly.
So the crash fix behaviour of #3052 is still there?

Unfortunately, I don't have MGS3 to see this in action.
Does anybody else have the game and can verify this PR fixes problems there?

KrossX:
Yes, the possible crash when spamming F9 should still be fixed.

lightningterror:
Can confirm issue is fixed, fmv switch issue remains.

To reproduce the issue here are the steps:
Create a savestate during codec screen > make sure you're running sw renderer > have the fmv switch be turned on > enter a fmv, renderer will switch to hw mode automatically > load savestate while the fmv is active.

KrossX:
I tried generating events from vsync() to call `Sys_RenderToggle` but nothing I tried worked. Although not surprising since I no nothing of wxWidgets. However, whatever thread runs the RenderToggle seems to be alright, perhaps there's a way to only run the FMV switch if it's from that thread.

ramapcsx2:
So this PR fixes one out of two remaining issues with renderer switching?
That's a good start :)

MrCK1:
Works fine for me compared to previous behavior from #3052. No longer hangs when using F9.

ramapcsx2:
Just gonna merge this in then.
If you want to try fixing the FMV toggle as well, that'd be great @KrossX :)

mirh:
Is fmv auto switch issue #1189?

KrossX:
So it would seem.

PR #3073, opened 2019-08-15
3rdparty/pthreads4w: Ensure on_process_init early execution

CookiePLMonster:
This fixes a semi-consistent regression introduced by #3056, where `on_process_exit` would execute before other destructors relying on pthreads.

Now `on_process_init` has been moved to an earlier section of static initializers, **ensuring** it will be initialized first. Previously, this initializer was placed in the same section as any other initializer, making their order of execution non-deterministic across compilations - for example, I was unable to reproduce this issue as soon as I forced `pthreads4w.c` to recompile last!


gandalfthewhite19890404:
This commit cause hang on Settings wizard, tested twice - Windows 7 x64

tadanokojin:
We've been made aware of the issue but thank you for reporting it to us.

PR #3075, opened 2019-08-16
GSdx-hw: improve and use sw sprite render in DBZ:BT2, replacing OO & CU with OI

AlessandroVetere:
Improve and use `GSRendererHW::SwSpriteRender` for rendering sprites in DBZ:BT2, replacing the current OO and CU methods with an OI method, similar to what has been done for Jak games.

This both improves the performances and correctly renders some effects such as kaioken and character outlines also at upscaled resolutions.

Function `GSC_DBZBT2` in `GSHwHack.cpp` has been removed, as it clashes with the new OI method (if it is not removed, blur builds up around characters models).

Some assertions in `GSRendererHW::SwSpriteRender` have been improved to ease debugging.

An handy method `GSRendererHW::CanUseSwSpriteRender` has been implemented, to factorize the checks that are needed to be performed prior calling the `GSRendererHW::SwSpriteRender` to replace current draw call.
Looking forward to backport the usage of the usage of `GSRendererHW::CanUseSwSpriteRender` to `GSRendererHW::OI_JakGames` to reduce custom checking code.

Changelist:

- In `GSRendererHW::SwSpriteRender`, improve alpha blending by implementing all the possible `ALPHA.C` cases,
- In `GSRendererHW::SwSpriteRender`, improve texture function by implementing the `TEX0.TFX = 1` case,
- In `GSRendererHW::SwSpriteRender`, use `GS_PRIM` enums for `PRIM.PRIM` register checks in assertions instead of hardcoded values,
- In `GSRendererHW::SwSpriteRender`, use `m_r` as drawing region rectangle directly,
- In `GSRendererHW::SwSpriteRender`, correct assertions for no input texture offset, no input texture min/mag, no input texture UV wrap,
- In `GSRendererHW::SwSpriteRender`, allow non flat shading,
- In `GSRendererHW::SwSpriteRender`, ensure color equality only in case the draw is not a sprite draw,
- In `GSRendererHW::SwSpriteRender`, use the color of last vertex for the draw,
- In `GSRendererHW::SwSpriteRender`, fix the implementation of frame buffer masking using `FRAME.FBMSK`,
- Add `GSRendererHW::CanUseSwSpriteRender` method to factorize the checks needed before calling `GSRendererHW::SwSpriteRender` to replace current draw,
- In `GSHwHack.cpp` remove function `GSC_DBZBT2`,
- In `GSRendererHW`, replace OO & CU methods for DBZ:BT2 with a new OI method using the sw sprite render,

Tests shown up to 3ms frametime reduction on Intel i5-2500k and Nvidia GTX 770.

Needed https://github.com/PCSX2/pcsx2/pull/3091

### DIFF
--- a/plugins/GSdx/Renderers/HW/GSRendererHW.cpp
+++ b/plugins/GSdx/Renderers/HW/GSRendererHW.cpp
@@ -811,6 +811,7 @@ void GSRendererHW::SwSpriteRender()
 	bool fb_mask_enabled = m_context->FRAME.FBMSK != 0x0;
 	GSVector4i fb_mask = GSVector4i(m_context->FRAME.FBMSK).u8to16(); // 0x00AA00BB00GG00RR00AA00BB00GG00RR
 
+	uint8 tex0_tfx = m_context->TEX0.TFX;
 	uint8 tex0_tcc = m_context->TEX0.TCC;
 	uint8 alpha_b = m_context->ALPHA.B;
 	uint8 alpha_c = m_context->ALPHA.C;
@@ -833,8 +834,10 @@ void GSRendererHW::SwSpriteRender()
 				sc = GSVector4i::loadl(&s[scol[x]]).u8to16();  // 0x00AA00BB00GG00RR00aa00bb00gg00rr
 
 				// Apply TFX
-				ASSERT(m_context->TEX0.TFX == 0);
-				sc = sc.mul16l(vc).srl16(7).clamp8();  // clamp((sc * vc) >> 7, 0, 255), srl16 is ok because 16 bit values are unsigned
+				ASSERT(tex0_tfx == 0 || tex0_tfx == 1);
+				if (tex0_tfx == 0)
+					sc = sc.mul16l(vc).srl16(7).clamp8();  // clamp((sc * vc) >> 7, 0, 255), srl16 is ok because 16 bit values are unsigned
+
 				if (tex0_tcc == 0)
 					sc = sc.blend(vc, a_mask);
 			}
--- a/plugins/GSdx/Renderers/HW/GSRendererHW.cpp
+++ b/plugins/GSdx/Renderers/HW/GSRendererHW.cpp
@@ -724,7 +724,6 @@ void GSRendererHW::SwSpriteRender()
 {
 	// Supported drawing attributes
 	ASSERT(PRIM->PRIM == GS_TRIANGLESTRIP || PRIM->PRIM == GS_SPRITE);
-	ASSERT(!PRIM->IIP);  // Flat shading method
 	ASSERT(!PRIM->FGE);  // No FOG
 	ASSERT(!PRIM->AA1);  // No antialiasing
 	ASSERT(!PRIM->FIX);  // Normal fragment value control
@@ -744,9 +743,10 @@ void GSRendererHW::SwSpriteRender()
 	ASSERT(m_context->FRAME.PSM == PSM_PSMCT32);
 
 	// No rasterization required
-	ASSERT(m_vt.m_eq.rgba == 0xffff);
-	ASSERT(m_vt.m_eq.z == 0x1);
-	ASSERT(!PRIM->TME || PRIM->FST || m_vt.m_eq.q == 0x1);  // Check Q equality only if texturing enabled and STQ coords used
+	ASSERT(PRIM->PRIM == GS_SPRITE
+		|| ((PRIM->IIP || m_vt.m_eq.rgba == 0xffff)
+			&& m_vt.m_eq.z == 0x1
+			&& (!PRIM->TME || PRIM->FST || m_vt.m_eq.q == 0x1)));  // Check Q equality only if texturing enabled and STQ coords used
 
 	bool texture_mapping_enabled = PRIM->TME;
 
@@ -804,8 +804,9 @@ void GSRendererHW::SwSpriteRender()
 
 	bool alpha_blending_enabled = PRIM->ABE;
 
-	GSVector4i vc = m_vt.m_min.c;  // 0x000000AA000000BB000000GG000000RR
-	vc = vc.ps32();                // 0x00AA00BB00GG00RR00AA00BB00GG00RR
+	const GSVertex& v = m_vertex.buff[m_index.buff[m_index.tail - 1]];  // Last vertex.
+	const GSVector4i vc = GSVector4i(v.RGBAQ.R, v.RGBAQ.G, v.RGBAQ.B, v.RGBAQ.A)  // 0x000000AA000000BB000000GG000000RR
+							.ps32();  // 0x00AA00BB00GG00RR00AA00BB00GG00RR
 
 	GSVector4i a_mask = GSVector4i::xff000000().u8to16();  // 0x00FF00000000000000FF000000000000
 
--- a/plugins/GSdx/Renderers/HW/GSRendererHW.cpp
+++ b/plugins/GSdx/Renderers/HW/GSRendererHW.cpp
@@ -813,6 +813,8 @@ void GSRendererHW::SwSpriteRender()
 
 	uint8 tex0_tcc = m_context->TEX0.TCC;
 	uint8 alpha_b = m_context->ALPHA.B;
+	uint8 alpha_c = m_context->ALPHA.C;
+	uint8 alpha_fix = m_context->ALPHA.FIX;
 
 	for (int y = 0; y < h; y++, ++sy, ++dy)
 	{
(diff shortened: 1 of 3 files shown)


plugins/GSdx/Renderers/HW/GSRendererHW.cpp
```diff
@@ -2150,27 +2156,26 @@ bool GSRendererHW::OI_JakGames(GSTexture* rt, GSTexture* ds, GSTextureCache::Sou
...
+		return true;
+	GSVector4i r = m_r;
+	if (!(r == GSVector4i(0, 0, 64, 64)).alltrue())  // Rendering region is 64x64
+		return true;
+	if (m_vt.m_eq.rgba != 0xffff || m_vt.m_eq.z != 0x1 || (PRIM->TME && !PRIM->FST && m_vt.m_eq.q != 0x1))  // No rasterization
+		return true;
+	if (m_context->DepthRead() || m_context->DepthWrite())  // No depth handling
+		return true;
+	SwSpriteRender();
+	return false; // Skip current draw
```
lightningterror: Maybe add a empty row before return.
Bonus/optional, maybe move entire OI below Bigmutha OI instead?

AlessandroVetere: What's the point in moving it below Bigmutha OI ?

lightningterror: Nothing major just alphabetical order for convenience.

plugins/GSdx/Renderers/HW/GSRendererHW.cpp
```diff
@@ -1714,6 +1720,26 @@ bool GSRendererHW::OI_BigMuthaTruckers(GSTexture* rt, GSTexture* ds, GSTextureCa
...
 
+bool GSRendererHW::OI_DBZBT2(GSTexture* rt, GSTexture* ds, GSTextureCache::Source* t)
+{
+	// Sprite rendering
+	if (m_context->FRAME.PSM != PSM_PSMCT32)
+		return true;
+	if (PRIM->TME && m_context->TEX0.PSM != PSM_PSMCT32)
+		return true;
+	GSVector4i r = m_r;
+	if (!(r == GSVector4i(0, 0, 64, 64)).alltrue())  // Rendering region is 64x64
```
turtleli: Any reason you can't just use `m_r`?

plugins/GSdx/Renderers/HW/GSRendererHW.cpp
```diff
@@ -812,8 +815,11 @@ void GSRendererHW::SwSpriteRender()
 				sc = GSVector4i::loadl(&s[scol[x]]).u8to16();  // 0x00AA00BB00GG00RR00aa00bb00gg00rr
 
 				// Apply TFX
-				ASSERT(m_context->TEX0.TFX == 0);
-				sc = sc.mul16l(vc).srl16(7).clamp8();  // clamp((sc * vc) >> 7, 0, 255), srl16 is ok because 16 bit values are unsigned
+				ASSERT(tex0_tfx == 0 || tex0_tfx == 1);
+				sc = tex0_tfx == 0 ?
+					sc.mul16l(vc).srl16(7).clamp8() : // clamp((sc * vc) >> 7, 0, 255), srl16 is ok because 16 bit values are unsigned
+					sc
```
turtleli: Would avoiding the ternary improve readability?
```
if (tex0_tfx == 0)
    sc = sc.mul16l(vc).srl16(7).clamp8() : // clamp((sc * vc) >> 7, 0, 255), srl16 is ok because 16 bit values are unsigned
```

plugins/GSdx/Renderers/HW/GSRendererHW.cpp
```diff
@@ -837,23 +843,24 @@ void GSRendererHW::SwSpriteRender()
...
 				ASSERT(m_context->ALPHA.D == 1);
-				ASSERT(m_context->ALPHA.FIX == 0);
 
-				GSVector4i sc_alpha_vec =
-					sc.yyww()   // 0x00AA00BB00AA00BB00aa00bb00aa00bb
+				// Flag C
+				GSVector4i sc_alpha_vec = alpha_c == 2 ?
+					GSVector4i(alpha_fix).xxxx().ps32() :
+					(alpha_c == 0 ? sc : dc0)
+					.yyww()     // 0x00AA00BB00AA00BB00aa00bb00aa00bb
```
turtleli: Ditto.

plugins/GSdx/Renderers/HW/GSRendererHW.cpp
AlessandroVetere: No, it's just a leftover from previous code where the `r` was recomputed from scratch instead of reusing `m_r`

plugins/GSdx/Renderers/HW/GSRendererHW.cpp
AlessandroVetere: I avoid the ternary operator for the ALPHA.C = 2 case, but maybe for 0 and 1 I will keep it

plugins/GSdx/Renderers/HW/GSRendererHW.cpp
AlessandroVetere: Sure

lightningterror:
This improves the KaioKen red effect but at the same time it breaks the character outlines when the effect is used.

Master:
gsdx_20190924223148

PR:
gsdx_20190924223159

SW:
gsdx_20190924223152

GS Dump:
[DBZ BT2 OI-KaioKen.gs.zip](https://github.com/PCSX2/pcsx2/files/3649333/DBZ.BT2.OI-KaioKen.gs.zip)

tadanokojin:
Shall we postpone this for 1.8?

lightningterror:
It is postponed, just didn't add the milestone for it :)

AlessandroVetere:
I rebased and force pushed.

A noticeable change in the sprite renderer is that I'm now using the color of the last vertex for the draw.
This correctly implements the sprite primitive draw, and does not change the other cases where color equality for all the vertices is ensured (so the color of the last vertex is the same of all the other vertices).
Thanks to @tadanokojin for having highlighted this behaviour in the game Tom and Jerry: War of the Whiskers.


As a TODO I added to check if the last vertex of the draw (used to read the color) is equal to `m_vertex.buff[m_vertex.tail - 1]`.


The Goku character outline issue when using the Kaioken still persists and seems related to buggy/unconsistent depth buffer readback in the texture cache, triggered by different TC usage in the PR w.r.t master.

AlessandroVetere:
And the outline bug appears to be related to something wrong in the alpha channel draws rather than the Z buffer readback. With respect to the software renderer, both the master and the PR hardware renderers seem to draw the alpha channel wrong when drawing Goku, but yet the master is capable of computing his outline correctly and the PR is not 😄
Not yet sure of what is really going on here.

plugins/GSdx/Renderers/HW/GSRendererHW.cpp
```diff
@@ -804,16 +804,21 @@ void GSRendererHW::SwSpriteRender()
...
+	GSVertex& v = m_vertex.buff[m_vertex.tail - 1];  // Last vertex.
+	GSVector4i vc = GSVector4i(v.RGBAQ.R, v.RGBAQ.G, v.RGBAQ.B, v.RGBAQ.A);  // 0x000000AA000000BB000000GG000000RR
+	vc = vc.ps32();  // 0x00AA00BB00GG00RR00AA00BB00GG00RR
 
 	GSVector4i a_mask = GSVector4i::xff000000().u8to16();  // 0x00FF00000000000000FF000000000000
 
 	bool fb_mask_enabled = m_context->FRAME.FBMSK != 0x0;
 	GSVector4i fb_mask = GSVector4i(m_context->FRAME.FBMSK).xxxx(); // 0x????????????????MMMMMMMMMMMMMMMM
 
+	uint8 tex0_tfx = m_context->TEX0.TFX;
```
lightningterror, line 817: Maybe use constant variables when applicable throughout the code, just makes the code cleaner and less VS warnings to deal with.

AlessandroVetere:
The issue was a wrong FBMSK implementation in the SwSpriteRender function, and now it is fixed.
Thanks again LT for finding the issue in the first place.
Enjoy the kaioken effect and outlines at 4x res!

image

Looking forward to solve at least those TODOs:

* Is draw vertices count is equal to m_vertex.tail?
* Is the last vertex of the draw is equal to m_vertex.buff[m_vertex.tail - 1]?

Could anybody help with those questions? Thanks!

AlessandroVetere:
Feels ready for me now. Better to merge squash all the commits when reviewed and tested.

PR #3080, opened 2019-08-17
3rdparty/pthreads4w: Revert recent pthreads changes/commits

lightningterror:
This reverts commits:
https://github.com/PCSX2/pcsx2/commit/88a02941f698577ecc45450308269f57e1d066ad
https://github.com/PCSX2/pcsx2/commit/af6f04020225eaf3239bc9b4542f4afd84560540

It actually causes more issues instead of fixing so just revert the new
code. Porting to std::thread for 1.8 would be a far better alternative.

Current master is broken and hangs on pcsx2 launch.

### DIFF
--- a/3rdparty/pthreads4w/autostatic.c
+++ b/3rdparty/pthreads4w/autostatic.c
@@ -73,14 +73,9 @@ static void on_process_init(void)
 attribute_section(".ctors") void *gcc_ctor = on_process_init;
 attribute_section(".dtors") void *gcc_dtor = on_process_exit;
 
-#if !PCSX2_FIX
 attribute_section(".CRT$XCU") void *msc_ctor = on_process_init;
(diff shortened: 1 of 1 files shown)


yxmline:
https://github.com/PCSX2/pcsx2/commit/88a02941f698577ecc45450308269f57e1d066ad is broken and hangs on pcsx2 launch
this is ok

ramapcsx2:
It would have been nice to have this kind of problem fixed before release, exactly for the reasons it seems to cause trouble now.

If it can't be fixed, then this PR reverts it okay.

lightningterror:
I'll go ahead and merge this then, original author also kinda gave up on fixing this so let's leave any pthreads changes for 1.8.

I gave it a quick test right now to make sure it doesn't hang anymore and it seems to be working fine.

PR #3035, opened 2019-07-23
Translations: Revert recent changes to pt_BR

rffontenelle:
The contributor applied changes only to `pcsx2_Main.po`. By not changing `pcsx2_Iconized.po` as well caused translation inconsistency between `pcsx2_Main.po` and `pcsx2_Iconized.po`, which has not being solved for 3 months now.

For the record, the reverted commits contain predominantly (good) style changes, not about gross error or typo fixes of previous translations.

PR #3082, opened 2019-08-18
Update Portuguese Brazilian Translation

altiereslima:


pgert:
@GuilhermeDiasJ & @rffontenelle:
Any comments?

rffontenelle:
@pgert I notice that the OP closed the PR (not sure why), but if it is reopened or resubmitted, I have no problem with it - although I didn't reviewed the changes -, and in this case it would probably be better to close #3035

rffontenelle:
@altiereslima Consider reopening this PR, as the previous one was closed already.

PR #3084, opened 2019-08-18
GSdx-d3d11: Adjust maximum texture size limit based on available feature level

lightningterror:
d3d10 = 8192
d3d11 = 16384

Seems it is easier to hit the limit. Champions of Norrath with 8x upscale for example.

### DIFF
--- a/plugins/GSdx/Renderers/DX11/GSDevice11.cpp
+++ b/plugins/GSdx/Renderers/DX11/GSDevice11.cpp
@@ -226,6 +226,12 @@ bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
 		return false;
 	}
 
+	// Set maximum texture size limit based on supported feature level.
+	if (level >= D3D_FEATURE_LEVEL_11_0)
+		m_d3d_texsize = D3D11_REQ_TEXTURE2D_U_OR_V_DIMENSION;
+	else
+		m_d3d_texsize = D3D10_REQ_TEXTURE2D_U_OR_V_DIMENSION;
+
 	{	// HACK: check nVIDIA
 		// Note: It can cause issues on several games such as SOTC, Fatal Frame, plus it adds border offset.
 		bool disable_safe_features = theApp.GetConfigB("UserHacks") && theApp.GetConfigB("UserHacks_Disable_Safe_Features");
@@ -638,8 +644,9 @@ GSTexture* GSDevice11::CreateSurface(int type, int w, int h, int format)
 
 	memset(&desc, 0, sizeof(desc));
 
-	desc.Width = std::max(1, std::min(w, 8192)); // Texture limit for D3D10 min 1, max 8192
-	desc.Height = std::max(1, std::min(h, 8192));
+	// Texture limit for D3D10/11 min 1, max 8192 D3D10, max 16384 D3D11.
+	desc.Width = std::max(1, std::min(w, m_d3d_texsize));
+	desc.Height = std::max(1, std::min(h, m_d3d_texsize));
 	desc.Format = (DXGI_FORMAT)format;
 	desc.MipLevels = 1;
 	desc.ArraySize = 1;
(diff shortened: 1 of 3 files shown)


plugins/GSdx/Renderers/DX11/GSDevice11.cpp
```diff
@@ -226,6 +226,12 @@ bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
 		return false;
 	}
 
+	// Set texture size limit based on supported feature level.
+	if (level < D3D_FEATURE_LEVEL_11_0)
+		m_d3d_texsize = 8192;
+	else
+		m_d3d_texsize = 16384;
```
turtleli: Would it be better to use `D3D10_REQ_TEXTURE2D_U_OR_V_DIMENSION` and `D3D11_REQ_TEXTURE2D_U_OR_V_DIMENSION`?

lightningterror: Cool, much better indeed.

plugins/GSdx/Renderers/DX11/GSDevice11.cpp
```diff
@@ -632,14 +638,17 @@ void GSDevice11::ClearStencil(GSTexture* t, uint8 c)
...
 	HRESULT hr;
 
 	D3D11_TEXTURE2D_DESC desc;
 
 	memset(&desc, 0, sizeof(desc));
 
-	desc.Width = std::max(1, std::min(w, 8192)); // Texture limit for D3D10 min 1, max 8192
-	desc.Height = std::max(1, std::min(h, 8192));
+	desc.Width = std::max(1, std::min(w, m_d3d_texsize)); // Texture limit for D3D10 min 1, max 8192 D3D10, max 16384 D3D11.
+	desc.Height = std::max(1, std::min(h, m_d3d_texsize));
```
ssakash: Maybe  use ``std::clamp``instead?

plugins/GSdx/Renderers/DX11/GSDevice11.cpp
```diff
@@ -226,6 +226,12 @@ bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
 		return false;
 	}
 
+	// Set maximum texture size limit based on supported feature level.
+	if (level < D3D_FEATURE_LEVEL_11_0)
```
ssakash: I find it more readable when the conditions are inverted, D3D11+ cases at first and the lower ones as fallback. Do you agree?

plugins/GSdx/Renderers/OpenGL/GSTextureOGL.cpp
```diff
@@ -150,6 +150,9 @@ namespace PboPool {
 GSTextureOGL::GSTextureOGL(int type, int w, int h, int format, GLuint fbo_read, bool mipmap)
 	: m_clean(false), m_generate_mipmap(true), m_local_buffer(nullptr), m_r_x(0), m_r_y(0), m_r_w(0), m_r_h(0), m_layer(0)
 {
+	ASSERT(w > 0);
+	ASSERT(h > 0);
```
ssakash: I'm starting to wonder if there's any point in these asserts if we're just manually  enforcing them into a range later, also these asserts trigger on release builds or only on dev and lower?

tadanokojin: Only on debug. Just there to warn developer about potential hidden issue or to warn them about improper new code.

plugins/GSdx/Renderers/DX11/GSDevice11.cpp
CookiePLMonster: It's a C++17 feature so would probably cut off compilers - if anything it should be made in code somewhere.

lightningterror: Yeah, can't use clamp over here (vs 2015). Buildbot and appveyor also use 2015.

mirh: 2015.3 should [actually support](https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=vs-2015) clamp.

lightningterror: Yeah there is a switch to use c++17 but I'll leave it out. Plans to switch to c++17 are set for after 1.6 release.

Issue #2478, opened 2018-06-09
GS Dump code seems to dislike non-ASCII characters in path


feikname:
***PCSX2 version:*** PCSX2 1.5.0-20180608225456 - compiled on Jun  8 2018

***PCSX2 options:*** Preset 2 (safe)

***Plugins used:*** Default ones from the build provided by https://buildbot.orphis.net/pcsx2/

GSDX with everything on auto, no hardware hacks and in native resolution.

***Description of the issue:***
It seems PCSX2 can't handle paths with non-ASCII characters when attempting to save GS dumps.

``` 
GSDump: Error failed to open C:\Users\Fl_vio\Downloads\pcsx2-v1.5.0-dev-2405-g1c58d5acf-windows-x86\snaps\gsdx_20180609060135.gs.xz
```

(my username is Flávio)

Relevant code: 
https://github.com/PCSX2/pcsx2/blob/af3e55af63dd23075c08eb6b181a6fe62793d8c0/plugins/GSdx/GSDump.cpp#L25-L32

***How to reproduce the issue:***
Try to do a GS dump (<kbd>Shift</kbd>+<kbd>F8</kbd>) with pcsx2 inside a folder that contains non-ASCII characters such as latin accented letters (e.g à é í õ ü).

***Last known version to work:*** Not tested with previous versions

***PC specifications:*** Intel Core i5-8250U; Intel UHD Graphics 620 & AMD Radeon 530; Windows 10 64-bit

lightningterror:
It's not really a big issue, and the recommended install location doesn't include those characters.

I'll close this for now unless someone actually wants to look in to this.

turtleli:
I changed the title to better reflect the issue. This doesn't just affect GS dumps - it also affects the plugin ini and log locations.

This is a combination of a few problems:
1. Wide character versions of functions need to be used on Windows for proper Unicode support (_wfopen instead of fopen)
2. PCSX2 doesn't necessarily pass the plugins the correct filepath strings (not Windows specific).

mirh:
Related
https://github.com/PCSX2/pcsx2/blob/ba1689f6d634ef9a668c88103f99d445b6b25eac/plugins/GSdx/GSSettingsDlg.cpp#L81-L93

ssakash:
Does #2509 fix this issue?

lightningterror:
Partially I believe, it's stated in the PR changes.

PR #3085, opened 2019-08-19
gsdx: Use the same ini code on both Windows and Linux

turtleli:
Partially replace the Linux ini code so that it:
 - works on both Windows and Linux
 - is backwards compatible with older GSdx inis
 - properly handles UTF-8 ini pathnames

This fixes an issue with GSdx on Windows not being able to handle non-ASCII ini pathnames.

This should mean that only dev9ghzdrk is unable to handle non-ASCII ini pathnames though GSdx does still have Unicode issues (external shader/capture filenames).

Related to #2478.

### DIFF
--- a/plugins/GSdx/GSdx.cpp
+++ b/plugins/GSdx/GSdx.cpp
@@ -23,6 +23,7 @@
 #include "GSdx.h"
 #include "GS.h"
 #include "PSX/GPU.h"
+#include <fstream>
 
 static void* s_hModule;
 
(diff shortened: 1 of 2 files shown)


CookiePLMonster:
Looks good to me, proper UTF-16 handling on Windows is long overdue.

That said, untested.

turtleli:
Updated
 * Fix trimming of key string trailiing whitespace
 * Avoid unnecessary processing if the key isn't recognised

turtleli:
Updated to avoid an infinite loop if a file read error occurs while reading the ini.

PR #3090, opened 2019-08-24
ci:appveyor: Add VS2019 and remove VS2015

turtleli:
Adds a VS2019 build job and removes the VS2015 build job.

When this is merged, support for compiling with VS2015 will be dropped.


scribam:
I think the DXSDK is now available on VS2019 image: https://www.appveyor.com/updates/2019/10/08/ so, maybe, it is not required to install it manually.

turtleli:
Updated to remove workaround.

PR #3094, opened 2019-08-25
GSdx-hw: Enable automatic mipmapping for Jak X.

AlessandroVetere:
Fixes screen rendering.


lightningterror:
It doesn't cause any issues as a side effect?
@atomic83GitHub I believe you tested this one too?

AlessandroVetere:
Breaks sea rendering. I'll close this for now.

PR #3095, opened 2019-08-26
Make recLUT not hardcoded to 32 MB.

water111:
This change makes the EE recompiler not hardcoded to working with 32 MB of RAM, and instead work with the amount of RAM set in Ps2MemSize::MainRam. The rest of PCSX2 seems to work fine with more than 32 MB of RAM - it is only the EE recompiler that has trouble. If the Ps2MemSize::MainRam value is not changed from the default 32 MB, there should be no change: 32 MB / 0x10000 = 0x200, the value that was there previously.

This may be helpful if anybody else in the future wants to emulate a PS2 dev kit with 128 MB or RAM, or maybe the PSX dvr thing which I think has 64 MB of RAM. I've confirmed that with the change, you could set Ps2MemSize::MainRam to 128 MB, and execute code with the recompiler that's above the first 32 MB of RAM, and do VIF and scratchpad DMA transfers from this upper memory as well.

### DIFF
--- a/pcsx2/x86/ix86-32/iR5900-32.cpp
+++ b/pcsx2/x86/ix86-32/iR5900-32.cpp
@@ -519,7 +519,7 @@ static void recAlloc()
 	for (int i = 0; i < 0x10000; i++)
 		recLUT_SetPage(recLUT, 0, 0, 0, i, 0);
 
-	for ( int i = 0x0000; i < 0x0200; i++ )
+	for ( int i = 0x0000; i < Ps2MemSize::MainRam / 0x10000; i++ )
 	{
 		recLUT_SetPage(recLUT, hwLUT, recRAM, 0x0000, i, i);
 		recLUT_SetPage(recLUT, hwLUT, recRAM, 0x2000, i, i);


ssakash:
> This change makes the EE recompiler not hardcoded to working with 32 MB of RAM, and instead work with the amount of RAM set in Ps2MemSize::MainRam. 

Doesn't make a difference since we don't alter ``MainRam`` after its value initialization, just seems like a handy macro to be used at this point.

> I've confirmed that with the change, you could set Ps2MemSize::MainRam to 128 MB, and execute code with the recompiler that's above the first 32 MB of RAM, and do VIF and scratchpad DMA transfers from this upper memory as well.

Interesting, this might make it easier to experiment with it. Modifying namespace variable value **vs** finding some random loop counter variable.

water111:
> Interesting, this might make it easier to experiment with it. Modifying namespace variable value **vs** finding some random loop counter variable.

Yeah, that would be the only benefit.  I recently made a modified build of PCSX2 for accessing the debug menu in a Jak game (https://www.youtube.com/watch?v=EjICpZXrCyk), which executes code and does DMA transfers on RAM past 32 MB.  Somebody had tried to increase the RAM to 128 MB to get to this menu before me, but he was unable to find out why it was crashing - it wasn't straightforward to debug why the JIT was getting TLB miss errors.  Hopefully this would save somebody time in the future.

ssakash:
@water111 
Fair enough, add a comment near ``Ps2MemSize::MainRam`` value init about the cases for experimenting with non-default values and I'll merge.

lightningterror:
@water111 Can you update the pr based on Akash's request?

turtleli:
IMO PR is fine as it is, it makes the code clearer.

PR #3099, opened 2019-09-02
GSdx-ocl: Enable OpenCL build option on Linux

JibbityJobbity:
Easy as adding `--opencl` when running `./build.sh`.
Removed #error hello since latest 32-bit version of OpenCL is 1.2 anyway and there isn't any significance noted for it.
This doesn't improve the renderer but I thought this might help.

### DIFF
--- a/plugins/GSdx/CMakeLists.txt
+++ b/plugins/GSdx/CMakeLists.txt
@@ -48,7 +48,7 @@ if(EGL_API AND EGL_FOUND)
 endif()
 
 
-if(OPENCL_API)
+if(OPENCL_FOUND)
     set(GSdxFinalFlags ${GSdxFinalFlags} -DENABLE_OPENCL)
 endif()
 
--- a/plugins/GSdx/GSdx.cpp
+++ b/plugins/GSdx/GSdx.cpp
@@ -347,9 +347,6 @@ void GSdxApp::Init()
 	// D3D Blending option
 	m_default_configuration["accurate_blending_unit_d3d11"]               = "1";
 
-	// OpenCL device. Windows only for now.
-	m_default_configuration["ocldev"]                                     = "";
-
 	// PSX option. Not supported on linux.
 	m_default_configuration["dithering"]                                  = "1";
 	m_default_configuration["ModeRefreshRate"]                            = "0";
@@ -391,6 +388,7 @@ void GSdxApp::Init()
 	m_default_configuration["ModeHeight"]                                 = "480";
 	m_default_configuration["ModeWidth"]                                  = "640";
 	m_default_configuration["NTSC_Saturation"]                            = "1";
+	m_default_configuration["ocldev"]                                     = "";
 #ifdef _WIN32
 	m_default_configuration["osd_fontname"]                               = "C:\\Windows\\Fonts\\tahoma.ttf";
 #else
--- a/build.sh
+++ b/build.sh
@@ -219,6 +219,7 @@ for ARG in "$@"; do
         --no-cross-multilib ) useCross=0; ;;
         --coverity          ) CoverityBuild=1; cleanBuild=1; ;;
         --vtune             ) flags="$flags -DUSE_VTUNE=TRUE" ;;
+        --opencl            ) flags="$flags -DOPENCL_API=TRUE" ;;
         -D*                 ) flags="$flags $ARG" ;;
 
         *)
@@ -239,6 +240,7 @@ for ARG in "$@"; do
             echo "--glsl          : Replace CG backend of ZZogl by GLSL"
             echo "--egl           : Replace GLX by EGL (ZZogl/GSdx plugins)"
             echo "--cross-multilib: Build a 32bit PCSX2 on a 64bit machine using multilib."
+            echo "--opencl        : Enable experimental OpenCL support"
             echo
             echo "** Distribution Compatibilities **"
             echo "--sdl12         : Build with SDL1.2 (requires if wx is linked against SDL1.2)"
--- a/cmake/BuildParameters.cmake
+++ b/cmake/BuildParameters.cmake
@@ -35,7 +35,7 @@ option(USE_VTUNE "Plug VTUNE to profile GSdx JIT.")
 #-------------------------------------------------------------------------------
 option(GLSL_API "Replace ZZogl CG backend by GLSL (experimental option)")
 option(EGL_API "Use EGL on ZZogl/GSdx (experimental/developer option)")
-option(OPENCL_API "Add OpenCL suppport on GSdx")
+option(OPENCL_API "Add OpenCL support on GSdx")
 option(REBUILD_SHADER "Rebuild GLSL/CG shader (developer option)")
 option(BUILD_REPLAY_LOADERS "Build GS replayer to ease testing (developer option)")
 option(GSDX_LEGACY "Build a GSdx legacy plugin compatible with GL3.3")
--- a/plugins/GSdx/res/tfx.cl
+++ b/plugins/GSdx/res/tfx.cl
@@ -1,9 +1,3 @@
-#if defined(CL_VERSION_2_0)
-
-#error hello
-
-#endif
-
 #if defined(CL_VERSION_1_1) || defined(CL_VERSION_1_2) // make safe to include in resource file to enforce dependency
 
 #ifdef cl_amd_printf


plugins/GSdx/GSdx.cpp
```diff
@@ -217,8 +217,10 @@ void GSdxApp::Init()
 	// FIXME openCL isn't attached to a device (could be impacted by the window management stuff however)
 #ifdef _WIN32
 	m_gs_renderers.push_back(GSSetting(static_cast<uint32>(GSRendererType::DX1011_OpenCL),	"Direct3D 11",	"OpenCL"));
-#endif
 	m_gs_renderers.push_back(GSSetting(static_cast<uint32>(GSRendererType::OGL_OpenCL),		"OpenGL",		"OpenCL"));
+#else
+	m_gs_renderers.push_back(GSSetting(static_cast<uint32>(GSRendererType::OGL_OpenCL),		"OpenGL",		"OpenCL"));
```
lightningterror: Old code should've been good here?

plugins/GSdx/GSdx.cpp
```diff
@@ -391,6 +390,7 @@ void GSdxApp::Init()
 	m_default_configuration["ModeHeight"]                                 = "480";
 	m_default_configuration["ModeWidth"]                                  = "640";
 	m_default_configuration["NTSC_Saturation"]                            = "1";
+	m_default_configuration["ocldev"]                                     = "";
```
lightningterror: GUI option would be nice too.

plugins/GSdx/GSdx.cpp
```diff
@@ -219,6 +219,7 @@ void GSdxApp::Init()
 	m_gs_renderers.push_back(GSSetting(static_cast<uint32>(GSRendererType::DX1011_OpenCL),	"Direct3D 11",	"OpenCL"));
 #endif
 	m_gs_renderers.push_back(GSSetting(static_cast<uint32>(GSRendererType::OGL_OpenCL),		"OpenGL",		"OpenCL"));
+#endif
```
lightningterror: Forgot this one too.

cmake/SearchForStuff.cmake
```diff
@@ -103,7 +103,7 @@ if(EGL_API)
     check_lib(X11_XCB X11-xcb X11/Xlib-xcb.h)
 endif()
 if(OPENCL_API)
-    check_lib(OPENCL OpenCL CL/cl.hpp)
+    find_package(OpenCL)
```
turtleli: Did you try `check_lib(OpenCL OpenCL CL/cl.hpp)`?

plugins/GSdx/CMakeLists.txt
```diff
@@ -276,7 +276,7 @@ if(USE_VTUNE)
 endif()
 
 if(OPENCL_FOUND)
```
turtleli: This needs to be `if(OpenCL_FOUND)` for the `check_lib` approach to work.

mirh:
> since latest 32-bit version of OpenCL is 1.2 

Ehrm.. this sounds new to me. Are you sure?

JibbityJobbity:
> > since latest 32-bit version of OpenCL is 1.2
> 
> Ehrm.. this sounds new to me. Are you sure?

I had the same reaction, but if you look up OpenCL 2.0 32 bit on Google, you'll find a few results, especially on the AMD community forum that OpenCL 2.0 is 32 bit only. I can confirm this when I built and ran 32-bit clinfo, platform version returns 1.2 on all of my devices where it returns 2.0 or higher in 64 bit. :disappointed:

Degerz:
Eh, I would remove OpenCL support entirely since it has been effectively depreciated. When you can't report bugs or they ignore them on 2 out of the 3 hardware vendors you're targeting, there's no future anymore for it ...

mirh:
> especially on the AMD community forum that OpenCL 2.0 is 32 bit only

:\

That's more like \*only\* there, because last time I checked it was simply them being lazy on QA. `GPU_FORCE_OCL20_32BIT` should unlock it iirc. 

> When you can't report bugs or they ignore them on 2 out of the 3 hardware vendors you're targeting

Uh, says who? Putting even aside that in linux for 2 of the vendors, the drivers are open

JibbityJobbity:
> :\
> 
> That's more like *only* there, because last time I checked it was simply them being lazy on QA. `GPU_FORCE_OCL20_32BIT` should unlock it iirc.
Agh! Forum replies I read suggested it was for the API and not just AMD's implementation of it. I can see results when I put `GPU_FORCE_OCL20_32BIT`, it's probably a thing but I'm not seeing any official documentation for it. I think you're onto something though since `CL_VERSION_2_0` shouldn't be defined otherwise. That's assuming they're different headers for 32-bit and 64-bit.

Degerz:
> Uh, says who? Putting even aside that in linux for 2 of the vendors, the drivers are open

Nvidia does not even have a developer subforum to report OpenCL issues so good luck sending reports of OpenCL driver issues to them and as for AMD they have straight neglected their OpenCL subforum on their official developer forum for a couple of years now. 

OpenCL on Linux is also a joke as well since much of the community aren't interested in developing the clover stack so progress over there has pretty much stalled and AMD exhibits the same behaviour for their ROCm OpenCL by ignoring issues over there as well. The only vendor that cares about OpenCL left is Intel ...

2 out of the 3 vendors ignoring OpenCL issues means that it's a dead API and should probably not use it unless the developers are okay with maintaining it to only reliably run on Intel graphics hardware!

mirh:
Amd has not neglected opencl any more than anything else (actually, to be honest, now that I check them they seem more responsive than for when we report about opengl troubles)
Also [here](https://github.com/RadeonOpenCompute/ROCm-OpenCL-Runtime/issues)
Nvidia doesn't have a forum section, but from what I could see anywhere in Accelerated Computing (yes, including the cuda sections) would be cool.

Degerz:
> (actually, to be honest, now that I check them they seem more responsive than for when we report about opengl troubles)

You need to see the reality for what it truly is. At least AMD's OpenGL drivers get real bug fixes from time to time while they've effectively stopped trying to advertise OpenCL support since they haven't submitted anything to the CTS in nearly the past 4 years! 

With Nvidia, there's no visibility on OpenCL and they're just doing the bare minimum to maintain compatibility. They are also the reason why OpenCL isn't getting developed any further!

OpenCL has no future whether you like it or not because maintenance for it has consistently kept getting lower over the past years. (OpenCL is the new glide at this point)

mirh:
OCL is stagnating because it has as much features as you want, but it is an ass to program (in comparison to cuda). Nvidia has no faults there. 

Check what SYCL is to see how khronos (also intel, and the foss community) are advancing it horizontally. 

Also, speaking of CTS, they just this year made the humongous step to move it totally open on github. No trifles. 

Speaking of amd instead............ We still have opengl bugs opened since 3-4 years.

Degerz:
OCL's features are useless in practice if 2 of the vendors you're targeting don't have any intention of supporting CL 2 x and that especially applies to Nvidia if they're going to gate features like SVM behind CUDA rather than support it in OCL ... (Also why would features be an argument in favour of OCL when the likes of CUDA is a lower level API ?)

FOSS community effort on SYCL depends on clover stack so it's useless ...

What exactly would those GL bugs be on AMD ?

JibbityJobbity:
Can't help but notice this is starting to get off topic from the actual PR. This is just about adding a flag on Linux to enable building with OpenCL support. Can you please take this talk over to somewhere else like the forum?

arcum42:
The main thing I notice here is that if opencl isn't installed, it tries to build gsdx with opencl anyways and fails, whereas it would be preferable for it to either build gsdx without opencl in that case, or skip gsdx.

JibbityJobbity:
> The main thing I notice here is that if opencl isn't installed, it tries to build gsdx with opencl anyways and fails, whereas it would be preferable for it to either build gsdx without opencl in that case, or skip gsdx.

This is why we have `-DENABLE_OPENCL` in GSdx's CMakeLists.txt. Only thing it does if OpenCL wasn't found is it still tries to link OpenCL libraries. I'll get onto that soon.

arcum42:
It tries to include header files for OpenCL if you enable opencl, regardless of whether it was found, which is where it errored out for me, since I apparently don't have it installed. 

It's probably just a matter of adding a check to SelectPcsx2Plugins.cmake checking if the option is enabled and it's not found.

Otherwise, I see no reason not to merge it. Don't think it's far enough to need a gui option, but having a option to build it makes sense for testing.

lightningterror:
> Otherwise, I see no reason not to merge it. Don't think it's far enough to need a gui option, but having a option to build it makes sense for testing.

I'm fine with the current pr, gui option was an extra nice thing to have :)

On a related note maybe we can purge the cl_d3d9.

JibbityJobbity:
> It tries to include header files for OpenCL if you enable opencl, regardless of whether it was found, which is where it errored out for me, since I apparently don't have it installed.
> 
> It's probably just a matter of adding a check to SelectPcsx2Plugins.cmake checking if the option is enabled and it's not found.
> 
> Otherwise, I see no reason not to merge it. Don't think it's far enough to need a gui option, but having a option to build it makes sense for testing.

I think it's better to build without OpenCL instead of skipping GSdx. I feel like the header thing is because we're using check_lib instead of FindPackage.

arcum42:
Now that I've had a better chance to look at this, I think we're fine. I'll go ahead and merge it, and if we need to make any adjustments, we can do that later.

PR #3105, opened 2019-09-06
FreeBSD: Fix compilation

turtleli:
This is a re-attempt at fixing FreeBSD compilation (I reverted the previous attempt in a977f1e6c1c67f1c4fe55cb1dfc760bd7c648bca).

The changes are largely the same.

Changes:
 * Remove ambiguous wxString-related overloads - wxString can be constructed with char*, wxChar* and wxString, so having function overloads that accept all of these types can actually result in ambiguity errors. Compared to the previous attempt, I also removed an assert (which was what caused issues in the previous attempt, the assert doesn't do anything meaningful due to the changes I made) and a FastFormatUnicode overload.
 * Split the thread callback function - Prevents compilation issues on Windows.
 * wx-config change - The wx3.0 pkg on FreeBSD is now configured slightly differently (built against GTK3) so the path has been adjusted to reflect that.


lightningterror:
Is ISO swapping working fine now? @refractionpcsx2 can you test this?

refractionpcsx2:
> Is ISO swapping working fine now? @refractionpcsx2 can you test this?

Wat? I don't have FreeBSD!

lightningterror:
See https://github.com/PCSX2/pcsx2/commit/93d5b52df34c03a2ed859bba6e06fead3b875e70#commitcomment-30390863

refractionpcsx2:
oh right, uhh I'll try it, I can't remember exactly how to replicate it, can only guess though

refractionpcsx2:
TBH I can't get it to go wrong on Master anymore.

turtleli:
> * Compared to the previous attempt, I also removed an assert (which was what caused issues in the previous attempt, the assert doesn't do anything meaningful due to the changes I made)

turtleli:
I should mention that the ISO swapping issue only affected dev/debug builds before I reverted the commits.

refractionpcsx2:
That may be why then

PR #3106, opened 2019-09-08
GameDB: Xenosaga fixes for broken save points.

atomic83GitHub:
This PR add a patch for Xenosaga NTSC-U version. This fixes broken save points in the game.

Tested by pandubz.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -1029,6 +1029,16 @@ Serial = SCAJ-30001
 Name   = Xenosaga Episode I [PlayStation 2 The Best]
 Region = NTSC-Unk
 Compat = 5
+[patches = A3D63039]
+	// Disassembled with love by PSI and pandubz.
+	// The game loads a value from what appears to be an uninitialized memory
+	// address into register v0. A later slti op checks if register v0 is less than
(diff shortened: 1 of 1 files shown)


tadanokojin:
As I've already voiced I'm pretty against this.

This patch fixes a bug present on real hw. It's not the job of the gamedb to fix bugs developers left in the game. That's a job for a cheat system or rom hacking.

Less you want to include this one too https://www.romhacking.net/hacks/3309/

Also I had nothing to do with testing this.

willkuer:
Not sure about that. Isn’t pcsx2 also about giving a better experience than the original? If this is indeed a dev bug that doesn’t yield any nostalgia but is just an annoyance and it is basically free to fix it here... why not?

You could always disable gamefixes to get the original buggy game.

tadanokojin:
You missed out, we already debated and it was decided to include the patch. I won't reiterate my arguments here, but similar concerns were addressed already.

refractionpcsx2:
Yes similar concerns were adressed, my side was that if the basic functionality of the game doesn't work and causes the game to crash, this is a problem and should be patched.

But I also agree if it's an enhancement (removing lag, increasing framerate, adding items missed out in the final game etc) this should be a cheat and kept separate.

PR #3110, opened 2019-09-13
spu2-x: Make portaudio an optional dependency.

orbea:
This makes portaudio an optional dependency for the `spu2-x` audio plugin. I have confirmed the build succeeds without a 32-bit portaudio installed, without portaudio being linked in `libspu2x-2.0.0.so` and that sound still works with SDL2.

Here is what the audio settings looks like with this commit.

Please note:

* ~~I did not touch any windows code~~ since I can't build or test it. If anyone wants to test the build and runtime I can try to finish that or it can wait for a followup PR by someone else?
* I did not touch anything regarding the zerospu2 audio plugin which I haven't used before.

Please review this to make sure I didn't do anything wrong, thanks!


orbea:
Seems appyveyor can test windows builds, I'll see if I can fix it there.

plugins/spu2-x/src/Windows/Config.cpp
```diff
@@ -145,7 +145,9 @@ void ReadSettings()
     Config_WaveOut.NumBuffers = CfgReadInt(L"WAVEOUT", L"Buffer_Count", 4);
 
     DSoundOut->ReadSettings();
+#ifdef SPU2X_PORTAUDIO
```
turtleli: The changes in this file seem unnecessary.

turtleli:
I don't really like this change. It means an extra configuration to test, and I don't see how this improves things.

orbea:
> I don't see how this improves things.

On my system the only thing that requires portaudio is PCSX2 and I don't use any of the features it provides for pcsx2 since SDL2 is generally better for audio in my experience. This would allow one less dependency to be acquired for anyone using PCSX2 on Slackware and potentially other places. Also because no one else uses a 32-bit multilib portaudio its much harder to get on Slackware than the equivalent package for SDL2. So this would really improve the situation here. :)

> It means an extra configuration to test

I'm willing and capable of periodically testing this build configuration and spending time on fixing issues which probably won't be too hard.

I'll do the suggested changes and update this PR in a little bit, thanks!

orbea:
I rebased this PR against master, implemented the suggested changes and dropped the second commit with windows fixes since it was suggested as not needed and it makes no difference to me.

arcum42:
I'm personally fine with making portaudio an optional dependency under Linux, since I'm the one who went and made sdl the default on Linux. SDL generally seemed to be working better, and is already a dependency for onepad, so it'll be there on Linux, though Windows doesn't use it as of yet.

Besides, Portaudio spams a bunch of Alsa warnings any time I try it on pcsx2. :P

orbea:
@arcum42 Any news on this PR? Not sure there is much more to be done here.

turtleli:
I think it'd be better to add a cmake option to disable PortAudio instead of relying on the detection of PortAudio. Otherwise it would be a bit inconvenient for a dev/tester to test SPU2-X both with and without PortAudio.

orbea:
Good idea, I will hopefully have time to add it today.

orbea:
I did a few things.


* I rebased this PR.
* I fixed a small, but critical typo breaking PCSX2 with portaudio in `plugins/spu2-x/src/CMakeLists.txt` (oops).
* I added a second commit with `PORTAUDIO_API` to optionally skip portaudio support. It defaults to `ON`.
* I tested that `-DPORTAUDIO_API=OFF` disables portaudio support even when its installed with both `SDL1` and `SDL2`.
* I added `--no-portaudio` to `build.sh`.

@turtleli Do you see anything else missing?

plugins/spu2-x/src/Linux/Config.cpp
```diff
@@ -124,12 +126,13 @@ void ReadSettings()
 
 #if SDL_MAJOR_VERSION >= 2
     CfgReadStr(L"OUTPUT", L"Output_Module", temp, SDLOut->GetIdent());
-#else
+#elif defined(SPU2X_PORTAUDIO)
```
turtleli: I think this change is incorrect. SDL1.2, no PortAudio SPU2-X builds will not work properly because the `OUTPUT/Output_Module` ini key won't be read.

cmake/SelectPcsx2Plugins.cmake
```diff
@@ -267,12 +267,13 @@ endif()
...
-#			-ALSA
-#           -Portaudio
+#           -ALSA
 #           -SDL
 #           -common_libs
+#
+# optional: -Portaudio
 #---------------------------------------
-if((PORTAUDIO_FOUND AND SOUNDTOUCH_FOUND AND SDLn_FOUND AND common_libs)
+if((SOUNDTOUCH_FOUND AND SDLn_FOUND AND common_libs)
```
turtleli: I think it's better to skip the build if PortAudio isn't found and PortAudio isn't explicitly disabled.

arcum42: Honestly, I think I'd prefer the build succeeds if it's able to succeed without portaudio. SDL's been set to the default anyways at this point on Linux, as it tends to behave better than portaudio.

plugins/spu2-x/src/Linux/Config.cpp
arcum42: Hmm, yeah, sdl 1.2 is still a possibility, as onepad-legacy still supports it (as well as 2.0), even if onepad doesn't.

orbea: How does someone test with SDL1?
```
CMake Error at cmake/SelectPcsx2Plugins.cmake:21 (message):
  Skip build of onepad: missing dependencies:check these libraries -> sdl2,
  X11, gtk2
```
It seems my previous test with SDL1 was mistaken and I had SDL2 installed.

cmake/SelectPcsx2Plugins.cmake
orbea: I would also prefer to let the cmake check dynamically enable/disable portaudio support and leave `-DPORTAUDIO_API=OFF` only for skipping it when portaudio is installed.

plugins/spu2-x/src/Linux/Config.cpp
orbea: I got past that with `rm -r plugins/onepad` and `-DSDL2_API=FALSE`

However it now fails with many errors like this.
```
/tmp/SBo/pcsx2/build/plugins/onepad_legacy/ImgHeader/dualshock2.h:9:29: error: expected primary-expression before ';' token
  static const uint Length = ;
                             ^
/tmp/SBo/pcsx2/build/plugins/onepad_legacy/ImgHeader/dualshock2.h:10:29: error: size of array 'Data' is not an integral constant-expression
  static const u8 Data[Length];
                             ^
/tmp/SBo/pcsx2/build/plugins/onepad_legacy/ImgHeader/dualshock2.h:14:37: error: size of array 'Data' is not an integral constant-expression
 const u8 res_dualshock2::Data[Length] =
                                     ^
```
This is using `sdl-1.2.15` and I won't be able to confirm these concerns or try to fix them until this is fixed. It might be just better to remove SDL1 support instead? I'm not sure anyone does or should use it with pcsx2?

arcum42: Only real reason I can think of to leave SDL 1.2 support in there is for any platforms that have 1.2 but not 2.0, and while they did drop platforms in between the two, I'm pretty sure people shouldn't be trying to compile pcsx2 on, say, BeOS or an Amiga.

onepad does use both the 1.2 and 2.0 sdl api for controllers, but the 1.2 api is still in there, and isn't actually depreciated, just the 2.0 one is better for controllers as opposed to joysticks.

Either fixing onepad_legacy with sdl 1.2 or removing support is fairly outside the scope of this pr, though. As long as we allow for the possibility of it being sdl 1.2 in spu2-x for now, I think we can handle onepad_legacy being broken with sdl 1.2 as a separate issue later. :P

I'm not sure yet if I want to fix that, or just drop sdl 1.2 support.

Those are autogenerated files from images for all the pictures in the config screen, btw...

orbea: Yea, I agree that its out of scope. I think the options are to leave the remaining SDL1 issues for later when they can be tested properly or to wait until they are fixed and then rebase this PR accordingly. I am not sure the latter option will ever be done unless someone who understands how these files are generated better can take the initiative.

Regardless I don't think using SDL1 with pcsx2 on Slackware is a good idea where SDL2 is readily available and this PR seems ready as far as SDL2 is concerned.

turtleli: The CMake error is poor handling on our end.

onepad_legacy compiles fine on my end (ArchLinux) with SDL1.2 without modifications. So the errors you had are a bit strange.

As for SDL1.2 support - some distros link WX to SDL1.2. Otherwise it probably would have been dropped already. The situation might have changed - the two distros that I know of (Fedora and OpenSUSE) that used to link WX to SDL1.2 now seem to link WX to SDL2 instead.

orbea: It seems the real issue is was higher up in the log.
```
[ 34%] Generating ImgHeader/dualshock2.h
binmode() on closed filehandle IN at /tmp/SBo/pcsx2/linux_various/hex2h.pl line 63.
read() on closed filehandle IN at /tmp/SBo/pcsx2/linux_various/hex2h.pl line 64.
Use of uninitialized value $byte in numeric ne (!=) at /tmp/SBo/pcsx2/linux_various/hex2h.pl line 64.
Use of uninitialized value $filesize in concatenation (.) or string at /tmp/SBo/pcsx2/linux_various/hex2h.pl line 87.
```

It seems WX uses neither SDL1 or SDL2 here.

orbea: @turtleli You were right, that change was not correct with SDL1 and without portaudio. I rebased against master and included a small adjustment in the first commit to fix this.

orbea:
One related and preexisting bug which someone might run into after this PR is that if the the `Output_Module` line in `~/.config/PCSX2/inis/spu2-x.ini` is set to any incorrect value PCSX2 will crash when opening the audio settings.

The correct behavior would be to default to a working `Output_Module` if the current one does not exist anymore or is just wrong (Typo). Perhaps it should try in the order of SDL, portaudio and then the null module. I am not sure how to implement this though.

orbea:
Any news on this? I'm not sure the reasons why to delay this?

RedDevilus:
1.6 stable is soon. This will probably for the dev builds with 1.7 version.

PR #3115, opened 2019-09-21
Linux: force analog stick updates in onepad

lukedupin:
Linux has a bug with some games that don't detect the analog sticks.  Sky Odyssey is the game I observed this behavior on.

This update adds default behavior to the Guide button (Tested on Xbox 360 controller) which forces the analog sticks to update.  Previously the guide button wasn't referenced in OnePad.

I've added a few extra state-full variables that track the Guide button.  None of the core logic for handling key presses was touched in OnePad.

I've tested for several hours with multiple games, the change appears to be stable.

Usage:

- Load a game.
- Once the game has started, the user can press the guide button to toggle between normal and force analog mode.
- In force analog mode, the analog sticks function as expected in game.
- A printf is sent to the log window
- The default state is untouched
- The patch always starts in the default state.

### DIFF
--- a/plugins/GSdx/GSdx.cpp
+++ b/plugins/GSdx/GSdx.cpp
@@ -355,7 +355,6 @@ void GSdxApp::Init()
 
 	// PSX option. Not supported on linux.
 	m_default_configuration["dithering"]                                  = "1";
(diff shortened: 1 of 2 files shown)


lukedupin:
Mistakenly closed.

PR #3116, opened 2019-09-21
OnePad: Fixes analog sticks not registering in some games.

lukedupin:
Linux has a bug with some games that don't detect the analog sticks. Sky Odyssey is the game I observed this behavior on.

This update adds default behavior to the Guide button (Tested on Xbox 360 controller) which forces the analog sticks to update. Previously the guide button wasn't referenced in OnePad.

I've added a few extra state-full variables that track the Guide button. None of the core logic for handling key presses was touched in OnePad.

I've tested for several hours with multiple games, the change appears to be stable.

Usage:

    Load a game.
    Once the game has started, the user can press the guide button to toggle between normal and force analog mode.
    In force analog mode, the analog sticks function as expected in game.
    A printf is sent to the log window
    The default state is untouched
    The patch always starts in the default state.

### DIFF
--- a/plugins/onepad/GamePad.h
+++ b/plugins/onepad/GamePad.h
@@ -57,6 +57,10 @@ public:
      * Rumble will differ according to type which is either 0(small motor) or 1(big motor)
      */
     virtual void Rumble(unsigned type, unsigned pad) {}
+
+    //Returns true if the user expects the joystict to be analog
+    virtual bool GetGuideButton() { return false; }
+
     /*
      * Safely dispatch to the Rumble method above
      */
--- a/plugins/onepad/KeyStatus.cpp
+++ b/plugins/onepad/KeyStatus.cpp
@@ -29,6 +29,9 @@ void KeyStatus::Init()
         m_internal_button_joy[pad] = 0xFFFF;
         m_state_acces[pad] = false;
 
+        m_guide_button[pad] = false;
+        m_guide_button_pressed[pad] = false;
+
         for (int index = 0; index < MAX_KEYS; index++) {
             m_button_pressure[pad][index] = 0xFF;
             m_internal_button_pressure[pad][index] = 0xFF;
@@ -87,6 +90,25 @@ void KeyStatus::release(u32 pad, u32 index)
     } else {
         analog_set(pad, index, m_analog_released_val);
     }
+}
+
+void KeyStatus::pressGuide( u32 pad )
+{
+    m_guide_button_pressed[pad] = true;
+}
+
+void KeyStatus::releaseGuide( u32 pad )
+{
+    if ( m_guide_button_pressed[pad] )
+    {
+        m_guide_button_pressed[pad] = false;
+        m_guide_button[pad] = !m_guide_button[pad];
+    }
+}
+
+bool KeyStatus::getGuide(u32 pad)
+{
+    return m_guide_button[pad];
 }
 
 u16 KeyStatus::get(u32 pad)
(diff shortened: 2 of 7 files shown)


KrossX:
Does this also happen with LilyPad on Windows? Maybe there's some missing logic like mode lock of the DualShock.

lukedupin:
I ran LilyPad through wine and everything worked great.  I don't have an understanding for what happens when the controller update packet is passed outside of the plugin.  Perhaps Linux handles that exchange differently than windows?

LilyPad doesn't work on Linux, but I have tested both onepad and onepad_legacy; both of those plugins have the same behavior.

KrossX:
Well, if LilyPad worked fine then there's some bug in OnePad. I checked again and modelock is just to avoid the analog button from changing the mode, so it's ignored in OnePad and shouldn't have caused the issue. CMD 0x44 seems to be similar to LilyPad, Pokopom as well but maybe that plugin is bugged too. It would be nice to debug OnePad's mode change and see what comes up, unless the game is bugged is actually switches to digital making the player press the button but LilyPad has some hackfix? Dunno.

[OnePad CMD 0x44](https://github.com/PCSX2/pcsx2/blob/master/plugins/onepad/state_management.cpp#L406) | [OnePad Legacy CMD 0x44](https://github.com/PCSX2/pcsx2/blob/master/plugins/onepad_legacy/state_management.cpp#L351) | [LilyPad CMD 0x44](https://github.com/PCSX2/pcsx2/blob/master/plugins/LilyPad/LilyPad.cpp#L1455) | [Pokopom CMD 0x44](https://github.com/KrossX/Pokopom/blob/master/Pokopom/playstation_dualshock.cpp#L266)

lukedupin:
I wrote a quick mode tracker that watches for changes of the mode in the CMD_READ_DATA_AND_VIBRATE case statement.  I tried out 3 random games, here's what I found.

Sky Odyssey only asked for mode 0x41.  Analog doesn't work.

Jak and Daxter first asked for mode 0x73, then 0x79.  Analog works fine with one pad.

MX vs ATV starts with mode 0x41, then 0x73, 0x79.  It bounces around between 0x73 and 0x79, but when playing the game, its 0x79.  Analog works fine with one pad.

Not sure what this tells us.  Sky Odyssey is a problem child?  Lilypad must have extra logic in one of the other case statements that handles this edge case?

KrossX:
Can you test Pokopom under Wine and see if it behaves like LilyPad or OnePad? If it behaves like Lilypad then you should log all the stream since the beginning and see where OnePad/LilyPad differs. Maybe it's just some ID code that's different and the game doesn't recognize the controller as proper DualShock 2.

lukedupin:
I'll give it a go.

Is there a way to log the stream in the program or is adding printf's the best method?  I don't want to recreate the wheel if its as easy as selecting a menu option.

KrossX:
I just added fprintf to log into a file. Just this two functions should be enough: [padPokopom DebugPrint](https://github.com/KrossX/Pokopom/blob/master/Pokopom/psemupro.cpp#L269).

lukedupin:
Hey KrossX,

I found the problem.  There are two places where pad->mode should be set to analog that exists in LilyPad, but not in OnePad.  I copied the commands from LilyPad, and the analog stick worked as expected in Sky Odyssey.

I checked out OnePad_Legacy, it has the same bug.  I'll create another pull request that adds these checks in both OnePad implementations.

I think this pull request should be closed.

KrossX:
You can close your own PRs. =P

arcum42:
I'll go ahead and close this one, then, in favor of the other.

lightningterror:
It would've been ideal to focus on editing the existing pr.

PR #3117, opened 2019-09-21
cmake: Fix build with cmake-3.5.2.

orbea:
For Slackware stable (`14.2`) it has the older `cmake-3.5.2` which does not seem to support `GREATER_EQUAL`. So I replaced it with `VERSION_EQUAL` and `VERSION_GREATER` which is more verbose, but works with both newer and older cmake versions.

Unfortunately I am stuck testing against older cmake versions until Slackware 15.0 is out (When its ready).
```
CMake Error at cmake/SearchForStuff.cmake:225 (if):
  if given arguments:

    "GCC_VERSION" "GREATER_EQUAL" "9.0" "AND" "GCC_VERSION" "LESS" "9.2"

  Unknown arguments specified
  CMakeLists.txt:27 (include)


-- Configuring incomplete, errors occurred!
See also "/tmp/SBo/pcsx2/build/CMakeFiles/CMakeOutput.log".
See also "/tmp/SBo/pcsx2/build/CMakeFiles/CMakeError.log".
```

### DIFF
--- a/cmake/SearchForStuff.cmake
+++ b/cmake/SearchForStuff.cmake
@@ -220,7 +220,7 @@ if(GCC_VERSION VERSION_EQUAL "7.0" OR GCC_VERSION VERSION_EQUAL "7.1")
     GCC7_BUG()
 endif()
 
-if(GCC_VERSION GREATER_EQUAL "9.0" AND GCC_VERSION LESS "9.2")
+if((GCC_VERSION VERSION_EQUAL "9.0" OR GCC_VERSION VERSION_GREATER "9.0") AND GCC_VERSION LESS "9.2")
     message(WARNING "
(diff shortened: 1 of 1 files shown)


orbea:
It probably could also be changed to check against `9.0` and `9.1` to have the same effect, I'm not sure which is better or that it really matters. The current PR is perhaps more explicit which this would be less verbose.
```
if(GCC_VERSION VERSION_EQUAL "9.0" OR GCC_VERSION VERSION_EQUAL "9.1")
```

arcum42:
Works fine on my side with a newer version of cmake, tested both with gcc 9.1.0 and clang. I could go either way on which version of the check to use, but I'm fine with the current one, so I'll merge that.

Issue #1904, opened 2017-04-17
Lilypad Regression: Controllers do not work on some games


lightningterror:
Last working version: pcsx2-v1.5.0-dev-1934-ga3efc77
Regression version: pcsx2-v1.5.0-dev-1939-gf3a89f5

Some games don't read controller inputs , pcsx2 reads the inputs fine fine but pressing the buttons in game does nothing.

Affected games: The Getaway Black Monday , Tekken 5
There could be more but these are the only ones that showed these results for me.
Some others showed some weird results as well but those sometimes work and sometimes not.

All default settings used without speedhacks , cheats ...etc
i3 4160, amd r7-360, 16gbram, windows 7 64 bit

turtleli:
pgert reported on #1831 that the Network Access Disc didn't respond either.

Two guesses (try separately at first, then combined if neither work):
 * Try changing https://github.com/PCSX2/pcsx2/blob/master/plugins/LilyPad/LilyPad.cpp#L1367 to:
```
if (pad->mode == MODE_DIGITAL || pad->mode == MODE_PS1_MOUSE || pad->mode == MODE_NEGCON) {
```
* Try changing https://github.com/PCSX2/pcsx2/blob/master/plugins/LilyPad/LilyPad.cpp#L1333 to:
```
 if (pad->mode != MODE_ANALOG) {
```

FlatOutPS2:
Disable "Use analog mode if possible".

lightningterror:
@turtleli no result , still doesn't work.
@FlatOutPS2 that doesn't work either.

FlatOutPS2:
> that doesn't work either.

Restart the game.

lightningterror:
> Restart the game.
Ops I forgot to do that xD
It works now.

FlatOutPS2:
The function is not enabled by default(and it is a regression, though the option only worked with PS1 emulators before). Native PS2 behaviour doesn't always enable analog controls if it's supported or even recommended.

pgert:
The Projects Zero (Fatal Frame) games are also more or less unresponsive to DS2 input since the last commit-batch. Why not just revert?

turtleli:
So do the issues only exist when "Use Analog mode is possible" is enabled, or are there other regressions?

lightningterror:
For the games I have disabling ""Use Analog mode is possible" fixes the issues.
Probably more than half of my games library are affected.

There are no other regressions.

pgert:
[...]

PR #3122, opened 2019-09-28
OnePad: Fix analog sticks not registering in some games.

lukedupin:
This is the replacement pull request to  #3116

OnePad, and OnePad legacy are missing a check during the CMD_SET_MODE_AND_LOCK case in the statemachine.  If they are digital, they should be set to analog.

This change is taken directly from LilyPad which doesn't have the bug of some games not recognizing the analog sticks.  Sky Odyssey specifically has this bug.

This change fixes both OnePad and OnePad legacy.  Both are tested with Sky Odyssey.


KrossX:
Alright, I tested Pokopom and it has the same issue as OnePad. So, I dunno if the change is just a hackfix on LilyPad and it ain't just a game bug. 

Also, why the change to `SDL_GameControllerGetAxis` ?

lukedupin:
Not sure I can comment to the first part.  I just know onepad's behavior is on par with LilyPad for that game.

The SDL change isn't needed with modern compilers.  When I was building the last patch I made that change.  While going between floats and ints, I prefer to be clear.  It's a hold over from my last patch, your call to keep it or drop it.

turtleli:
I looked at the LilyPad code. When PCSX2 uses LilyPad the Auto Analog feature is actually disabled (it's very, very non-obvious though). So the analog mode forcing code you've added to this PR doesn't actually execute on LilyPad and might actually break things (See #1904).

KrossX:
Atomic tested the game and the gamepad does not have analog on when starting the game. So it's a game bug then and I think Disgaea does the same maybe due to old buggy SDK perhaps. Instead of setting cmd44 byte3 to 1 it sets byte4 to 1 when byte4 is for locking.

Does OnePad have an analog key? Maybe just set the guide button as the analog toggle button. It should not do anything if the controller mode lock is enabled though.

lukedupin:
Hey KrossX.  One pad does not have an analog key.  That is what my first patch was accomplishing.  Perhaps that's a better solution?

KrossX:
Implementing the analog key would be well, implementing controller functionality as opposed to a hackfix so I would think so. =)

Not many controllers have a guide button available though so maybe have a fall back key or keys as well? I dunno if OnePad allows customizing input to map such key.

While the previous patch did indeed check for a Guide button it was forcing analog instead of trying to toggle current mode using OnePad's set_mode, if the lock is not enabled. Since this analog button is not transmitted, on Pokopom I just check for it only once per... command?. StartPoll might be a good place for example, as it happens once per command and before the reponse that reports mode.

lukedupin:
I agree both changes are on the hacky side.  If the first patch is desirable, I'd be willing to update that patch to use the set_mode() and resubmit?  

OnePad doesn't allow configuration, but OnePad_legacy does.  I could check out what it'd take to update the config UI to allow the Analog button to be bound in legacy.  If its pretty straight forward I could apply the change there as well.

Please let me know what you think best.

KrossX:
Just reuse this PR with the new changes, you can revert commits, rebase, squash, force push, etc.

lukedupin:
Hey KrossX,

Here is a rather big push.  

I revamped OnePad_legacy's button enumerations.  There were three different enums all interacting with each other.  Changing anything in one broke the other two, very rigid and difficult to modify that way.  Now they're all based on the main button enum.

The button handling all works the same now, instead of the side loading hack job from before.  The lower button 16bits remain untouched, analog is on bit 24, which means its ignored for normal update operations.  There is one new method which gives the current button status for analog.

Setting analog mode is done through set_mode() now.  Hitting the button repeatedly toggles between the game set_mode() and MODE_ANALOG.

I found a some bugs in the file saving system that are now fixed.  It was possible to hang the program if you had an order error, or unexpected character.  I confirmed the current ini settings format is handled gracefully with the new analog item.

OnePad's analog handling is almost identical to legacy's.

I did NOT test this on Windows.  I don't have a Windows dev environment handy and I need to get focused on my moon light project again.  I hope you like what you see.

KrossX:
Nice work! OnePad does not work on Windows so you're one of the few who can fix it. =P

From what I can see, this implements the foce analog hack rather than the controller analog button. The analog button just toggles the mode, if analog set digital and vice-versa but only of the mode is not locked (CMD 0x44, pad->modeLock). If the lock is enabled, the analog button does nothing.

lukedupin:
This does implement the force hack.

I just read up on ps2 protocol:
http://store.curiousinventor.com/guides/PS2
https://gist.github.com/scanlime/5042071

I kinda get it, but the response packets PCSX2 is sending are different lengths from these articles.

Do you have example code of sending the set_analog command, both on and off?  Perhaps better reading?  If not, I'll poke at it sending some commands and seeing what it does.

KrossX:
The set mode and lock is already implemented in OnePad, [CMD 0x44 Set Mode and Lock](https://github.com/PCSX2/pcsx2/blob/277f18b73057548e5421e6d9930919f231436a3b/plugins/onepad_legacy/state_management.cpp#L380) you just gotta use it. 

Keep in mind that the analog button is not polled on a serial command. The state does affect certain command responses and that's how a game checks for it but the button itself, nope. That's why I suggested to check for this during pollstart as that's a function that happens once per command and also cuz I'm biased as that's how I implemented it. [Pokopom CMD byte 0](https://github.com/KrossX/Pokopom/blob/2f3f6a3b3dd02d73762b1c7badd1d4cae0984103/Pokopom/playstation_dualshock.cpp#L161)

The command lengths are the same is just a matter of how you count them, the mode tells you how many extra words there are from the header. The digital reponse is 0x41 (1 extra word), the dualshock is 0x73 (3 extra word) and with pressure enabled becomes 0x79 for 9 extra words after the header (3 bytes) for buttons (1 word, 2 bytes), analog sticks (2 words, 4 bytes, one per axis) and pressure (6 words, 12 bytes one per button). Here's an image I made some times ago: [DualShock packet length](http://i.imgur.com/WjoZyTd.png)

lukedupin:
If it's a quick update and you already have the solution in mind, do you want to add the code into state_machine and I'll test it?

KrossX:
Alright, here's a diff with an attempt. I don't even know if it compiles, sorry in advance.

[analog_changes_diff.zip](https://github.com/PCSX2/pcsx2/files/3693728/analog_changes_diff.zip)

lukedupin:
I like the look of these changes.  I'll apply them and let you know how it goes.

lukedupin:
Hey Kross,

I tried out the patch.  I got it to compile with a couple small changes, but the code isn't successful at changing modes anymore.  If you want to me to take on your code style but leave the current pull request functionality I'll do that.

Please advise.

KrossX:
Oh well, worth a shot. Thanks for trying it.

Just try to adapt your PR I would say, make sure it respects the mode lock (as opposed to a forced change) and the button should toggle the mode instead of always setting it to analog mode.

lukedupin:
I made a few minor tweaks and things are now running with your patch.  The Analog only works with the binary buttons, not sure how I missed that the first time.

I had to add a #define in another part of the code.  I run Arch Linux and clang upgraded to version 9 this week.  For some reason version 9 complains about a couple clang only helper functions.  The #define prevents these functions from being used if clang is >= 9.  

The new RC is pushed.

KrossX:
It works now? But it failed to compile.

lukedupin:
It looks like the clang 9 change was added upstream.  I resolved the conflict.  We should be all set now.  Everything works on my end.

common/include/x86emitter/x86_intrin.h
```diff
@@ -85,6 +85,7 @@ static unsigned int _rotl(unsigned int x, int s)
 
 #pragma clang diagnostic pop
 #endif
+#endif
```
lightningterror: Should be removed.

lukedupin:
Sorry lightningterror.  I synced up, fixed the conflict and re-pushed.  Confirmed compiling and running on my machine.

KrossX:
Now there's a green check! Yay!

lukedupin:
=) Thanks KrossX.  Its been a fun process.

lightningterror:
Would be nice to clean up/squash commit history, also split it in to two commits (onepad and onepad legacy).

lukedupin:
I made two pull requests, one for each plugin update.

Issue #3124, opened 2019-09-29
Crash when opening the configuration menu.


orbea:
***PCSX2 version:***
https://github.com/PCSX2/pcsx2/commit/d943870413d7719d71ba4de61f1b35202f9a4529

***Plugins used:***
`SPU2`

***Description of the issue:***
If the `Output_Module` variable in `spu2-x.ini` is wrong, for example if it was changed manually with a typo pcsx2 will crash when opening the audio configuration menu.
```
Thread 1 "PCSX2" received signal SIGSEGV, Segmentation fault.
0xeea818f3 in WriteSettings () at ../plugins/spu2-x/src/Linux/Config.cpp:205
205         CfgWriteStr(L"OUTPUT", L"Output_Module", mods[OutputModule]->GetIdent());
(gdb) bt
#0  0xeea818f3 in WriteSettings () at ../plugins/spu2-x/src/Linux/Config.cpp:205
#1  0xeea82088 in ReadSettings () at ../plugins/spu2-x/src/Linux/Config.cpp:177
#2  0xeea83110 in configure () at ../plugins/spu2-x/src/Linux/Config.cpp:483
#3  0x080f02a2 in SysCorePlugins::Configure (this=0xa7047f0 <CorePlugins>, 
    pid=PluginId_SPU2) at ../pcsx2/PluginManager.cpp:1756
#4  0x08288f0a in MainEmuFrame::Menu_ConfigPlugin_Click (this=0xacad290, event=...)
    at ../pcsx2/gui/MainMenuClicks.cpp:671
#5  0x08210c9c in Pcsx2App::HandleEvent (this=0xab319c0, handler=0xacad290, 
    func=<optimized out>, event=...) at ../pcsx2/gui/AppMain.cpp:640
#6  0xf7cced4c in wxAppConsoleBase::CallEventHandler(wxEvtHandler*, wxEventFunctor&, wxEvent&) const () from /usr/lib/libwx_baseu-3.0.so.0
#7  0xf7e9a0b4 in wxEvtHandler::ProcessEventIfMatchesId(wxEventTableEntryBase const&, wxEvtHandler*, wxEvent&) () from /usr/lib/libwx_baseu-3.0.so.0
#8  0xf7e9a52a in wxEvtHandler::SearchDynamicEventTable(wxEvent&) ()
   from /usr/lib/libwx_baseu-3.0.so.0
[...]

PR #3125, opened 2019-09-30
spu2-x: Fix crash with an invalid output module.

orbea:
Fixes https://github.com/PCSX2/pcsx2/issues/3124

This will default to `SDLOut` to match the current default and also its unlikely that someone will use PCSX2 without SDL installed since its required for onepad.

### DIFF
--- a/plugins/spu2-x/src/Linux/Config.cpp
+++ b/plugins/spu2-x/src/Linux/Config.cpp
@@ -132,24 +132,20 @@ void ReadSettings()
 // find current API
 #ifdef __linux__
     CfgReadStr(L"PORTAUDIO", L"HostApi", temp, L"ALSA");
-    OutputAPI = -1;
-    if (temp == L"ALSA")
-        OutputAPI = 0;
     if (temp == L"OSS")
         OutputAPI = 1;
-    if (temp == L"JACK")
+    else if (temp == L"JACK")
         OutputAPI = 2;
+    else // L"ALSA"
+        OutputAPI = 0;
 #else
     CfgReadStr(L"PORTAUDIO", L"HostApi", temp, L"OSS");
-    OutputAPI = -1;
-
-    if (temp == L"OSS")
-        OutputAPI = 0;
+    OutputAPI = 0; // L"OSS"
 #endif
 
 #ifdef __unix__
     CfgReadStr(L"SDL", L"HostApi", temp, L"pulseaudio");
-    SdlOutputAPI = -1;
+    SdlOutputAPI = 0;
 #if SDL_MAJOR_VERSION >= 2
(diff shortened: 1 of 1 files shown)


plugins/spu2-x/src/Linux/Config.cpp
```diff
@@ -174,6 +174,12 @@ void ReadSettings()
 
     Clampify(SndOutLatencyMS, LATENCY_MIN, LATENCY_MAX);
 
+    if (mods[OutputModule] == NULL) {
+        fwprintf(stderr, L"* SPU2-X: Unknown output module '%s' specified in configuration file.\n", temp);
```
arcum42: Clang errored out on this line. Looking closer, temp is a wxString, so would need to be converted. I believe it'd be temp.c_str() .

orbea: Oops, while clang does not build here, it gets far enough to have caught this. However `c_str()` does not work, but looking at the 3rdparty wx code it seems `wc_str()` does. I updated the PR with this change, does this look better to you?

orbea:
@arcum42 Just curious, what clang version are you testing with?

turtleli:
Travis CI currently uses clang 7 (ignore the build job saying `VERSION=3.8`, that's a leftover from when we tried to enforce clang-format).

arcum42:
When I actually test it on my computer, I use clang 8.0.1, but I was going by the travis failure, which is 7, as turtleli said. 

You can click on the checkmark or x by the build, hit Details by travis, and it'll show you what passed and what didn't, and clicking on the build jobs will show you the log. I *always* check the log on failures, too, because sometimes it fails to set up the environment properly, and it wasn't a compilation problem. This time it was, though.

orbea:
I am familiar with travis having worked with it in other projects, but for some reason I just didn't notice or look at travis this time. I'm curious how your clang differs from mine allowing you to build with `8.0.1`, but that is not relevant for this issue.

arcum42:
Looks like it works fine to me. Though, while we're addressing this issue, it occurs to me, in the same function, that perhaps the bits that initialise OutputApi and SdlOutputApi to -1 should initialize them to 0, instead. Currently, if you manually change the fields those correspond to in the ini file, it leaves them blank, and if you are using pulseaudio or sdl, you don't really want that setting not to be set to anything...

orbea:
I made a second commit that implements this which seems to work here.

orbea:
@arcum42 Any news on this PR? It would be nice to at least get the first commit merged because it prevents crashes where the second issue can always be fixed manually in the GUI.

arcum42:
Everything's looking good. I'll go ahead and merge it.

PR #3131, opened 2019-10-09
GSdx-hw, TC: invalidate Source w overlapping draw.

AlessandroVetere:
Improve texture cache `Source` objects invalidation logic taking into account overlapping with current draw.
Fixes eyes rendering for the Major in Jak 1.

### DIFF
--- a/plugins/GSdx/Renderers/HW/GSTextureCache.cpp
+++ b/plugins/GSdx/Renderers/HW/GSTextureCache.cpp
@@ -161,6 +161,7 @@ GSTextureCache::Source* GSTextureCache::LookupDepthSource(const GIFRegTEX0& TEX0
 		src->m_from_target_TEX0 = dst->m_TEX0;
 		src->m_32_bits_fmt = dst->m_32_bits_fmt;
 		src->m_valid_rect = dst->m_valid;
+		src->m_end_block = dst->m_end_block;
 
 		// Insert the texture in the hash set to keep track of it. But don't bother with
 		// texture cache list. It means that a new Source is created everytime we need it.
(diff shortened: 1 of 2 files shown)


AlessandroVetere:
I force pushed some changes:

* Use draw and Source rectangles x and y for computing real start block,
* Use Source psm instead of context psm for computing Source start end addresses,
* Remove depth psm exclusion, as the code is non correct for all psm: highlight this in TODO comment, the real (but slow) check would require checking overlaps block per block.

AlessandroVetere:
Force pushed again, now the correct code is implemented, albeit it might be slower.
All the blocks interested by the draw are computed in a lazy fashion and compared with the ones of the texture for overlap check.

lightningterror:
Maybe some logs for those draws will be nice.

AlessandroVetere:
Force pushed with requested logs and typo corrected in a comment.

AlessandroVetere:
I think what we are looking for here is, aside from the review of the PR correctness, the evaluation of the performance penalty of such algorithm.
In case the penalty would be too big for some cases, the algorithm might be simplified to a block ranges overlap check, rather than a full block per block overlap check.
This optimization might give the same correct results in almost all the relevant cases, but will be most likely wrong in a small subset of the cases due to swizzled memory for some pixel formats not working well with block ranges calculations and comparison.

lightningterror:
How much of a speed penalty are we talking about?

AlessandroVetere:
Very difficult to estimate in the general case as the added code is seldom executed. Worst case here seems to be around 1-2k loops to compute and compare the blocks of the current draw with the blocks of the possibly overlapping textures from RT.
With the approximated range overlap check you never loop, but I have yet to verify if the results would be the same.
I think testing performance impact on some TC heavy games might reveal the true cost of this feature in the current form.
I will test it on Jak in the following days and report some figures.

AlessandroVetere:
My benchmarks shown that in some cases with very high fps (around 300-400 fps), the precise overlap check had a strong impact (10-20 fps).
So I decided to switch to the fast approximate overlap check, that goes also with a much easier and possibly cleaner implementation, and has almost no impact itself on performances.
I removed also the log while making the changes, but I can re-add it if needed.

lightningterror:
I will merge this, so far all testing that was done showed no issues.
Testing was done properly throughout the months.

PR #3135, opened 2019-10-18
Removed Special Case in Pnatch Parser for Pointer Commands When Data Field is 0

Some1fromthedark:
I removed the special case for when the data field is 0 for an extended code that is pointer type. I tested this case on hardware with Codebreaker V10 and it worked normally.

### DIFF
--- a/pcsx2/Patch_Memory.cpp
+++ b/pcsx2/Patch_Memory.cpp
@@ -84,47 +84,6 @@ void handle_extended_t(IniPatch *p)
 		break;
 
 	case 0x6000: // 000Xnnnn iiiiiiii
-		if (IterationIncrement == 0x0)
-		{
-			//LastType = ((u32)p->addr & 0x000F0000) >> 16;
-			u32 mem = memRead32(PrevCheatAddr);
-			if ((u32)p->addr < 0x100)
-			{
-				LastType = 0x0;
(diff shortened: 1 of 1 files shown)


pcsx2/Patch_Memory.cpp
```diff
@@ -84,47 +84,6 @@ void handle_extended_t(IniPatch *p)
...
-				case 0x2:
-					memWrite32(PrevCheatAddr, (u32)p->data);
-					break;
-				default:
-					break;
-				}
-			}
-		}
-		else
 		{
```
turtleli: The indentation of this block needs to be reduced by 1 tab to match the surrounding code style (I can do this for you if you want). Otherwise, the commit is fine, the special case makes no sense to me.

pcsx2/Patch_Memory.cpp
```diff
@@ -84,76 +84,35 @@ void handle_extended_t(IniPatch *p)
...
-				if (((mem & 0x0FFFFFFF) & 0x3FFFFFFC) == 0)
-					PrevCheatType = 0;
-				else
-					PrevCheatType = 0x6001;
-			}
+				PrevCheatType = 0x6001;
 		}
-		break;
+	}
+	break;
```
turtleli: This break needs a tab.
```suggestion
		break;
```

turtleli:
Thanks.

PR #3145, opened 2019-10-26
onepad-legacy: Allows the user to map the analog button.

lukedupin:
When the analog button is pressed, the analog stick is toggled.  Fix for
Sky Odyssey


lightningterror:
I'll close this since onepad legacy won't be merged in to the core.

PR #3149, opened 2019-11-05
travis: use system number of cores instead of `3`

ahimta:
Using `nproc` which is part of the `GNU coreutils` we can do something like `make -j $(nproc)` instead of `make -j 3`:smiley:.

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -89,7 +89,7 @@ linux_32_script() {
 		..
 
 	# Documentation says 1.5 cores, so 2 or 3 threads should work ok.
-	make -j3 install
+	make -j `nproc` install
 }
 
 
@@ -132,7 +132,7 @@ linux_64_script() {
 		..
 
 	# Documentation says 1.5 cores, so 2 or 3 threads should work ok.
-	make -j3 install
+	make -j `nproc` install
 }
 
 linux_after_success() {


tadanokojin:
Li, this seems sensible to me shall we have it for 1.6.0? It's build stuff anyway.

travis.sh
```diff
@@ -89,7 +89,7 @@ linux_32_script() {
 		..
 
 	# Documentation says 1.5 cores, so 2 or 3 threads should work ok.
-	make -j3 install
+	make -j `nproc` install
```
turtleli: I think it's better to use `$(nproc)`? In your commit message you also wrote `make -j $(nproc)` so I'm not sure why you've opted to use backticks instead.

turtleli:
Also this change is equivalent to `make -j 2`, though it doesn't have much effect on compilation times (Travis CI documentation now states that it uses 2 cores).

travis.sh
ahimta: I just used `$(nproc)` in the commit message because the backtick is used in Markdown. And even if I escaped it, the commit message (e.g: using `git log`) would still have the escape character:sweat_smile:.

I see that you've mentioned that Travis CI documentation states that it uses 2 cores and could be fine to just hardcode `2`.

I just made this PR while looking into `travis.sh` and working on other PCSX2 stuff:sweat_smile:.

`nproc` is nice in the sense that it's portable and uses the level of concurrency the system provides. And tools (e.g: Travis CI, GitHub) either change with time in case a project lives long enough or the project dies before the next-gen tools are released:sweat_smile:. And major changes can even happen in the same tool by a new version or config/feature:sweat_smile:.

And finally, it may actually be the case that the `j` option has no effect in `make install`:sweat_smile: but regardless I'd use `nproc` even if just so that the reader can know of its existence and use it later (e.g: when compiling with `make -j $(nproc)`):sweat_smile:.

turtleli: The point I'm making is that `$(command)` is usually preferable to `` `command` `` when working with shell scripts (though in this case it makes no difference since there's no command nesting). So it is a bit surprising that you've used backticks in the shell script when you have used `$()` in the commit message.

Please do not use markdown in the git commit message.

> And finally, it may actually be the case that the `j` option has no effect in `make install`

It does have a effect. It's just that in this case it's not really noticeable.

ahimta: I find myself always using the backtick and `$()` only when I have to (probably because it's more concise):sweat_smile:.

> Please do not use markdown in the git commit message.

Sure. I intentionally don't use much Markdown in the PR description just in case it creeps into the commit message:smiley:.

> It does have a effect. It's just that in this case it's not really noticeable.

Good to know:yum:.

arcum42:
So... are we committing this one line code change?

travis.sh
orbea: Please do not use backticks for anything. Its a legacy syntax that has issues.

See: 
https://github.com/koalaman/shellcheck/wiki/SC2006
https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03

ahimta: I don't think the backtick issues apply here. I use ShellCheck in most projects I work on and I never use backticks on them because I'd  have to add line-exceptions or silence errors altogether and both are things to avoid.

Classes of issues (diseases) are what matters and not individual issues (symptoms) themselves. In this case, backticks and many other dubious constructs can easily be prevented by integrating fitness-functiony/CI practices like ShellCheck.

This is an obvious one and others less so are the following:
1. Code formatting using `clang-format` and verifying compliance on merge/push in CI pipeline.
2. Static analysis using `clang-tidy` and `cppcheck` (to name a few) and verifying compliance on merge/push in CI pipeline.
3. Using strict-compile options for the whole build and even stricter for project sources (as opposed to dependencies).

This may seem a bit defensive of me because it's probably is a teeny, tiny bit:stuck_out_tongue_closed_eyes:. Sorry I'm just a human.

And finally, for the whole period this issue has been open. Seriously, just flip a coin. Indecision over many issues can be much more harmful than making the wrong choice especially for two-way door decisions. Whoever most familiar with an area can just scope it to 10-15minutes and make a reasonable decision that will probably be hard to improve upon in any reasonable time.

Sorry if I'm being too much of a jerk and I'd understand if you block me from participating in the future to preserve the non-jerkiness of the community.

orbea: You're taking this out of perspective, its a matter of quality. Even if this individual case is trivial and unlikely to cause problems, its best to avoid them whenever possible and discourage their continued use by future developers.

lightningterror:
I will close this since we moved to actions.

PR #3158, opened 2019-11-15
Update locale information in Linux

JibbityJobbity:
Worth noting that I also gave executable permissions to generate_pot.sh. Other than that, only thing that was done was running the script.


pgert:
It makes it difficult for translators to spot where changes have been made if the po-files are updated without translation-update. With the mo-files there's no problem.

Also, it shouldn't make any difference if you update the files using Linux or Windows.
"Update locale information in Linux" -> "Update locale information".

knight-ryu12:
Worth noting here that

1. With this, all translation would need to link from POT file.
2. @pgert it's really not that problem if someone maintain POT file. (that all other PO is dependent on)

locales/sv_SE/pcsx2_Iconized.po
```diff
@@ -348,19 +348,7 @@ msgstr ""
 msgid ""
 "Off: Disables temporary aspect ratio switch.\n"
 "\n"
-"4:3: Temporarily switch to a 4:3 aspect ratio while an FMV plays to "
-"correctly display an 4:3 FMV.\n"
-"\n"
-"16:9: Temporarily switch to a 16:9 aspect ratio while an FMV plays to "
-"correctly display a widescreen 16:9 FMV."
```
turtleli: Hmm. Seems it doesn't like strings being split up into multiple lines. Needs investigating.

PR #3168, opened 2019-12-01
Gamedb: Add vuClampMode 2 gamefix to Soul Calibur 2

lightningterror:
Fixes respawn issues as well as SPS.
Also add a missing DB entry for a JAP release of SC3 Trial. The demo
discs may not actually need the gamefixes so they aren't included.

Wasn't able to verify if the gamefixes are required for SC3.


tadanokojin:
Split the commit.

PR #3172, opened 2019-12-18
pcsx2: DI execution is delayed by one instruction

lightningterror:
Core: Fixes booting issues in the following games:
Jak X, Namco 50th anniversary, Spongebob the Movie, Spongebob Battle for Bikini Bottom,
The Incredibles, The Incredibles rize of the underminer, Soukou kihei armodyne, Garfield Saving Arlene, Tales of Fandom Vol. 2.
The games will no longer require a patch to boot.

Gamedb: remove gamefix patches for DI execution fixes. Jak X, Spongebob Squarepants( The movie  and Battle for Bikini Bottom), namCollection - Namco 50th Anniversary, Soukou kihei armodyne, garfield saving arlene.

Credits go to Refraction, fix was backported from Dobiestation, original contributors Refraction and PSI

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -2167,10 +2167,6 @@ Name   = Jak X - Combat Racing
 Region = PAL-M7
 // reads Ratchet Gladiator data
 MemCardFilter = SCES-53286/SCES-53285
-[patches = df659e77]
-	comment=patched by Ge-Force
-	patch=1,EE,00275CF7,word,00000000
-[/patches]
 ---------------------------------------------
 Serial = SCES-53300
 Name   = SOCOM 3 - U.S. Navy SEALs
@@ -5562,10 +5558,6 @@ Name   = Jak X - Combat Racing
 Region = NTSC-U
 Compat = 5
 MemCardFilter = SCUS-97429/SCUS-97465
-[patches = 3091e6fb]
-	comment=patched by Ge-Force
-	patch=1,EE,00275C67,word,00000000
-[/patches]
 ---------------------------------------------
 Serial = SCUS-97430
 Name   = Hot Shots Golf FORE! [Regular Demo]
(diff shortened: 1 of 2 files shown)


ramapcsx2:
Nice!
Any info where this is from / how it was discovered?

refractionpcsx2:
> Nice!
> Any info where this is from / how it was discovered?

by PSI and Dobiestation, through testing :P

atomic83GitHub:
This is a awsome fix! You forgot to mention the Incredibles games but it's not that important as it is the same engine as those 2 spongebob games.

lightningterror:
> This is a awsome fix! You forgot to mention the Incredibles games but it's not that important as it is the same engine as those 2 spongebob games.

They do have a boot patch too, if it fixes those games too I can remove it as well.

ramapcsx2:
Well, code that uses DI is timing critical. This fix could be pretty big ;p

MrCK1:
As I've already told Ref, none of my other games have shown regressions. Everything is working fine so far :). He did mention on Discord that it shouldn't affect anything that didn't already have a big patch :P

ramapcsx2:
I suppose a good follow up would be to check IOP as well :)

atomic83GitHub:
This PR solve a game hanging at start on the title "Soukou kihei armodyne"

image

atomic83GitHub:
And a another game fixed by this PR "garfield saving arlene"

image

PSI-Rockin:
I'll give an in-depth explanation on what's going on.

These games rely on a bug in the EE kernel. Just before it starts executing a game, the kernel tries to disable interrupts with the following code:

```
di
mfc0 k0, Status
ori k0, 0x13
mtc0 k0, Status
sync
```

However, DI is not atomic. That means it does not complete after the instruction has executed, which means that the MFC0 will be executed before DI completes. That is to say... MFC0 will read the old value of the master interrupt enable, which is set to 1. The DI completes before MTC0 is executed, but MTC0 will overwrite the master interrupt enable as a result. The tl;dr here is that the DI has no effect because the kernel doesn't properly wait for interrupts to be disabled.

All these games check if interrupts are enabled when they start booting by reading the master interrupt enable bit. If interrupts are disabled, they will crash or hang in various ways - Jak X calls the Exit syscall, Namco 50 hangs on a black screen, Heavy Iron games hang on loading screens. There is no reason for them to do this other than to mess with emulators. Clearly it worked, given how long PCSX2 has needed patches to deal with the more popular games, and how the less popular games were simply doomed to never work.

atomic83GitHub:
Seriously, please merge the PR now, it fixes hanging problem in "Tales of Fandom Vol. 2"

image

PR #3181, opened 2019-12-21
pcsx2: mBit Synchronization Improvement

lightningterror:
Slightly improves mbit sensitive games, but still not perfect.

Here's the status on the known games:
Totally Spies - Less shit but still spikey,
Mike Tyson Heavyweight - The characters now have animation but they're
all black and spikey,
My Street - You can now see the characters but they are a bit spikey,
but probably playable.

All credits go to Refraction.

Description: This PR will improve synchronisation which gets done between COP2 and VU0 when hitting m-Bit flags on VU0 micro programs. Currently the microVU recompiler massively overshoots these, this PR serves to resolve this.

### DIFF
--- a/pcsx2/VU0microInterp.cpp
+++ b/pcsx2/VU0microInterp.cpp
@@ -205,8 +205,9 @@ void InterpVU0::Step()
 
 void InterpVU0::Execute(u32 cycles)
 {
-	for (int i = (int)cycles; i > 0 ; i--) {
-		if (!(VU0.VI[REG_VPU_STAT].UL & 0x1)) {
+	VU0.flags &= ~VUFLAG_MFLAGSET;
+	for (int i = (int)cycles; i > 0; i--) {
+		if (!(VU0.VI[REG_VPU_STAT].UL & 0x1) || (VU0.flags & VUFLAG_MFLAGSET)) {
 			if (VU0.branch || VU0.ebit) {
 				vu0Exec(&VU0); // run branch delay slot?
 			}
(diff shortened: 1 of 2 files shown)


refractionpcsx2:
Just to note, although these games aren't perfect, it's probably the closest they're going to get on PCSX2 without completely overhauling COP2/VU0/EE synchronisation, which isn't going to happen.

lightningterror:
It breaks Tri-Ace games so we will close this.

kozarovv:
Sorry for reviving that PR, but maybe is something worth to think about. 

[https://i.imgur.com/LpYuG5f.jpg](url)

Maybe this can be added as gamefix? That way it won't break anything cause it can be enabled only while game that need it is loaded. I compiled version for test purposes, and things seems to work fine. Basically creating gamefix from this is as simple as adding interface/setting itself, and minor change to @refractionpcsx2 code in microVU_Compile.inl to check for this setting while processing mbit. As a example change made to recompiler code: 
```
			else if (mVUup.mBit && !mVUup.eBit && !mVUinfo.isEOB && CHECK_MBITHACK) {
				mVUsetupRange(mVU, xPC, false);
				incPC(2);
				mVUendProgram(mVU, &mFC, 0);
				incPC(-2);
				goto perf_and_return;
			}
		}
```

There are fixes for 1 game like FFX ipu one or, TLB preload for Goemon, or Scarface. So adding fix that get few games to work as playable, and few others to near to playable seems to be worth it. 

**Games that goes to playable are:**

24 The Game - require additional 2 lines pnach, and correct rounding, fix massive slowdown
Gift - Fix in game FPS (game show 60 fps, but is not even close without mbit handling)
Woody Woodpecker - same as above. Even more noticeable. 
My street - Almost perfect on mbit branch with additional pnach.
Probably other Eko Software games.

**Games that work better, but not 100% fixed:**

Mike Tyson Heavyweight - require additional 3 lines pnach
Totally Spies - Less shit but still spikey.

I'm aware this isn't perfect solution. Some testing is required.

PR #3184, opened 2019-12-23
Fix VIF unpack log message in Vif_Unpack.cpp

chaoticgd:
The relevant VIF_LOG statement previously printed the packet size from the structure relating to VIF0, regardless of whether the log message was for VIF0 or VIF1. I'm pretty sure this is just a typo.


refractionpcsx2:
Good catch, it was most certainly a typo :)

PR #3185, opened 2019-12-23
GameDB: add International Cue Club 2 and burnout3  demo and telly addics 

Mrlinkwii:
This adds  burnout3 demo , International Cue Club 2 and Telly Addicts  to  Game DB

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -6505,7 +6505,7 @@ Serial = SLED-52488
 Name   = The Suffering [Demo]
 Region = PAL-E
 ---------------------------------------------
-Serial = SLED-52929
+Serial = SLED-52597
 Name   = Burnout 3: Takedown [Demo]
 Region = PAL-Unk 
 ---------------------------------------------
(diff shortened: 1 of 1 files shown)


bin/GameIndex.dbf
```diff
@@ -6505,6 +6505,10 @@ Serial = SLED-52488
 Name   = The Suffering [Demo]
 Region = PAL-E
 ---------------------------------------------
+Serial = SLED-52597
```
lightningterror: Might be worth to check if the crc id is different, and if that needs to be added to crc database as well for hw renderer.

Can you tell me the crc id of the demo so I can check?

Mrlinkwii: image


this is what pcsx2  says

Mrlinkwii: image

lightningterror: Thanks, it's already on the list.

bin/GameIndex.dbf
```diff
@@ -6505,6 +6505,10 @@ Serial = SLED-52488
...
+Name   = Burnout 3 -Takedown [Demo]
```
lightningterror: Little nitpick, add a space after -

PR #3189, opened 2019-12-26
Update Debian/Ubuntu dependencies to avoid system package broken state

raelgc:
After install any package that requires `portaudio19-dev`, and request to install `pcsx2-unstable`, `apt` will be in a broken dependencies state.

Because `pcsx2` depends on `libjack-jackd2-0:i386`, which is incompatible with `portaudio19-dev`.

I need to manually install `libjack-jackd2-0:i386` (this will fix the broken packages state), in order to install `pcsx2-unstable` later.

### DIFF
--- a/debian-packager/control
+++ b/debian-packager/control
@@ -30,10 +30,12 @@ Package: pcsx2-unstable
 Architecture: i386
 # -i686 allow to use SSE into libc6
 Depends: ${shlibs:Depends}, ${misc:Depends}, libc6-i686
-# libasound => alsa plugin for pulseaudio
-# libusb    => ps3 controller (and probably others)
+# libasound        => alsa plugin for pulseaudio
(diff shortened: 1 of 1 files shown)


orbea:
Shouldn't this be maintained in debian repos rather than upstream?

Also fwiw you can apply PR https://github.com/PCSX2/pcsx2/pull/3110 to the `1.6.0` release and disable portaudio.

lightningterror:
Is this still an issue since portaudio is now an optional dependency?

raelgc:
If you guys don't mind, I'll test this during the next weekend :)

raelgc:
Ok. Installed Retropie on 20.04: compiled Dolphin, installed the psx2 PPA, no more package conflicts! Sorry for the super late feedback and thanks for fixing!

PR #3194, opened 2020-01-01
GSdx-d3d11: Update point sampler handle

lightningterror:
gsdx-d3d11: Update point sampler code.
Make sure point sampler is enabled only on amd gpus. Not needed for
nvidia and intel, behavior is the same on opengl as well. Helps #1026

### DIFF
--- a/plugins/GSdx/Renderers/DX11/GSDevice11.cpp
+++ b/plugins/GSdx/Renderers/DX11/GSDevice11.cpp
@@ -130,6 +130,7 @@ bool GSDevice11::SetFeatureLevel(D3D_FEATURE_LEVEL level, bool compat_mode)
 bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
 {
 	bool nvidia_vendor = false;
+	m_amd_vendor = false;
 
 	if(!__super::Create(wnd))
 	{
@@ -170,6 +171,8 @@ bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
 				{
 					if (desc.VendorId == 0x10DE)
 						nvidia_vendor = true;
+					else if (desc.VendorId == 0x1002)
+						m_amd_vendor = true;
 
 					adapter = enum_adapter;
 					driver_type = D3D_DRIVER_TYPE_UNKNOWN;
--- a/plugins/GSdx/Renderers/DX11/GSDevice11.h
+++ b/plugins/GSdx/Renderers/DX11/GSDevice11.h
@@ -343,6 +343,7 @@ private:
 	int m_upscale_multiplier;
 	int m_mipmap;
 	int m_d3d_texsize;
+	bool m_amd_vendor;
 
 	GSTexture* CreateSurface(int type, int w, int h, int format);
 	GSTexture* FetchSurface(int type, int w, int h, int format);
--- a/plugins/GSdx/Renderers/DX11/GSTextureFX11.cpp
+++ b/plugins/GSdx/Renderers/DX11/GSTextureFX11.cpp
@@ -186,6 +186,8 @@ void GSDevice11::SetupGS(GSSelector sel, const GSConstantBuffer* cb)
 
 void GSDevice11::SetupPS(PSSelector sel, const PSConstantBuffer* cb, PSSamplerSelector ssel)
 {
+	const bool amd_point_sampler = sel.point_sampler && m_amd_vendor;
+
 	auto i = std::as_const(m_ps).find(sel);
 
 	if(i == m_ps.end())
@@ -207,7 +209,7 @@ void GSDevice11::SetupPS(PSSelector sel, const PSConstantBuffer* cb, PSSamplerSe
 		sm.AddMacro("PS_FBMASK", sel.fbmask);
 		sm.AddMacro("PS_LTF", sel.ltf);
 		sm.AddMacro("PS_TCOFFSETHACK", sel.tcoffsethack);
-		sm.AddMacro("PS_POINT_SAMPLER", sel.point_sampler);
+		sm.AddMacro("PS_POINT_SAMPLER", amd_point_sampler);
 		sm.AddMacro("PS_SHUFFLE", sel.shuffle);
 		sm.AddMacro("PS_READ_BA", sel.read_ba);
 		sm.AddMacro("PS_CHANNEL_FETCH", sel.channel);


mirh:
TIL we have hacks even on dx. 

Then though... why are you checking against amd's [chipsets](https://pci-ids.ucw.cz/read/PC/1022), and intel's [network cards](https://pci-ids.ucw.cz/read/PC/8087) (and whatever 0x163C even is)? They all always ever used just one vendor id for gpus.

plugins/GSdx/Renderers/DX11/GSDevice11.cpp
```diff
@@ -168,8 +166,15 @@ bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
 				hr = enum_adapter->GetDesc1(&desc);
 				if (S_OK == hr && (GSAdapter(desc) == adapter_id || adapter_id == "default"))
 				{
+					// D3D GPU vendor check.
 					if (desc.VendorId == 0x10DE)
-						nvidia_vendor = true;
+						m_d3d_gpu_vendor = D3D11_VENDOR_NVIDIA;
```
tadanokojin: Don't use `D3D11_` namespace. It is reserved by the api.
I don't think we need the `d3d_` prefix either.

plugins/GSdx/Renderers/DX11/GSDevice11.cpp
```diff
@@ -168,8 +166,15 @@ bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
...
+					else if (desc.VendorId == 0x1002 || desc.VendorId == 0x1022)
+						m_d3d_gpu_vendor = D3D11_VENDOR_AMD;
+					else if (desc.VendorId == 0x163C || desc.VendorId == 0x8086 || desc.VendorId == 0x8087)
+						m_d3d_gpu_vendor = D3D11_VENDOR_INTEL;
+					else
+						m_d3d_gpu_vendor = D3D11_VENDOR_UNKNOWN;
```
tadanokojin: I would probably write something like this

```
switch(desc.VendorId)
{
case 0x163C:
case 0x8086:
case 0x8087:
	m_d3d_gpu_vendor = D3D11_VENDOR_INTEL;
    break;
....
}

File: plugins/GSdx/Renderers/DX11/GSDevice11.h [hunk 3]
```diff
@@ -481,6 +489,8 @@ class GSDevice11 final : public GSDevice
 	static bool s_old_d3d_compiler_dll;
 
 public:
+	int m_gpu_vendor;
```
turtleli: Why int?

lightningterror:
> and whatever 0x163C even is?

This is what I found https://web.archive.org/web/20170704040507/http://pcidatabase.com/vendor_details.php?id=1594

plugins/GSdx/Renderers/DX11/GSDevice11.h
```diff
@@ -481,6 +489,8 @@ class GSDevice11 final : public GSDevice
 	static bool s_old_d3d_compiler_dll;
 
 public:
+	uint8 m_gpu_vendor;
```
turtleli: Why uint8?

lightningterror: Why not?

turtleli: Why not enum?

plugins/GSdx/Renderers/DX11/GSDevice11.cpp
```diff
@@ -168,8 +166,22 @@ bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
 				hr = enum_adapter->GetDesc1(&desc);
 				if (S_OK == hr && (GSAdapter(desc) == adapter_id || adapter_id == "default"))
 				{
-					if (desc.VendorId == 0x10DE)
-						nvidia_vendor = true;
+					// D3D GPU vendor check.
+					switch (desc.VendorId)
+					{
+						case 0x10DE:
+							m_gpu_vendor = GPU_VENDOR::VENDOR_NVIDIA;
```
tadanokojin: if you use an enum class (which I think is good btw) then you have nicer scoping and you don't need something like `GPU_VENDOR::VENDOR_NVIDIA` you can just go `GPU_VENDOR::NVIDIA`.

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -754,7 +756,7 @@ void GSRendererDX11::EmulateTextureSampler(const GSTextureCache::Source* tex)
 
 	m_ps_sel.ltf = bilinear && shader_emulated_sampler;
 
-	m_ps_sel.point_sampler = !bilinear || shader_emulated_sampler;
+	m_ps_sel.point_sampler = dev->m_gpu_vendor == GSDevice11::GPU_VENDOR::VENDOR_AMD && (!bilinear || shader_emulated_sampler);
```
tadanokojin: I think it would be better to define `m_gpu_member` as a protected member of `GSRendererHW`. You can write some nice utility function like `bool isAMD()` to make this code a bit nicer. Also OGL has a way to get vendor information. See https://www.khronos.org/opengl/wiki/OpenGL_Context#Context_information_queries

Maybe it has no use for OGL now but I think it's simple to add and won't hurt anything. @turtleli thoughts?

mirh: We already have hacks and checks around GL_VENDOR. 
It's kinda useless alone though, since driver version is usually the most important element #2696

lightningterror: Yeah we have vendor checks in GLLoader so it has no use for gl, with that in mind putting it in gsdevice now someone might use the wrong vendor check for gl.

tadanokojin: The idea is that it should be generic. All renderers have a way to check vendor. There's no reason we need 2-3 different ways to check depending on renderer. Implementation details shall be the responsibility of the respective renderers.

lightningterror:
Testing reveals that the point sampler may only not be needed on nvidia. Selecting any other adapter and the point sampler is needed so I will close this pr. I can exclude nvidia only but don't wanna risk breaking something before release.

PR #3208, opened 2020-01-15
Updated Turkish translations for PCSX2(tr_TR)

JudgeJE:
Updated the version number in the files that I sadly overlooked


lightningterror:
Looks good, I went through the gui and everything seems translated (well I don't know Turkish but strings were showing translations properly).
We need to fix the Presets tooltip which can be translated afterwards as well.

PR #3209, opened 2020-01-15
GSdx-hw: Restore GSC_LordOfTheRingsTwoTowers crc hack to PAL regions only

lightningterror:
Currently we can't use the TS half bottom fix on PAL regions because we
get a height delta bigger than 64. NTSC versions seem fine.

Fixes regression with PAL regions experiencing a half screen bottom
issue.

gs dump for testing
[LordOfTheRingsTwoTowers.gs.zip](https://github.com/PCSX2/pcsx2/files/4068040/LordOfTheRingsTwoTowers.gs.zip)

Maybe instead of reintroducing the crc hack we just force enable the half screen fix on Lord of the Rings
Will wait for feedback.


tadanokojin:
Discussed on discord. We will add an additional option to GSdx which will allow for our currently flawed detection as well as forced enabled and disabled options.

Issue #3211, opened 2020-01-16
Halfscreen detection is flawed


tadanokojin:
Sometime ago I worked on https://github.com/PCSX2/pcsx2/pull/2934 which was a fix to handle games which split the frame buffer in order to do texture shuffle. Why they do this I am unsure. It may have been an optimization of some kind.

However, at the time I relied on some heuristics which have turned out to be flawed.
See:
https://github.com/PCSX2/pcsx2/pull/3162
https://github.com/PCSX2/pcsx2/pull/3209

Admittedly it is a shortcut and now the chickens have come home to roost.

Instead of this heuristic, what I want to do is handle the issue in the cache. @AlessandroVetere has made great progress in handling mid-texture hits in the cache for the Jak games with his tex_inside_rt work. I believe that this work can also benefit us here too.

TS relies on reading back a previous rt in a different format and mapping the pixels accordingly. Currently the issue with halfscreen is we need to be able to hit mid rt in order to get the other half. The hit will not always be dead center.

See related issues:
https://github.com/PCSX2/pcsx2/issues/2327
https://github.com/PCSX2/pcsx2/issues/1339

lightningterror:
With the addition of https://github.com/PCSX2/pcsx2/pull/5628 this has been greatly improved, real solution will be to fix the cache but this will do for now.

PR #3215, opened 2020-01-16
GSdx: Split half bottom detection in 3 levels controlled by a gui option

lightningterror:
Add an option in hw hacks to control the half bottom detection.

Automatic, Force Disabled, Force Enabled.

Automatic - automatically determinate if a game needs the half screen
fix.
Force Disabled - fully disables the half bottom fix. It will help
Xenosaga games.
Force Enabled - fully enables the half bottom fix. Will help Lord of the
Rings Two Towers, Superman Shadows of Apokolips, Midnight Club 3.

**Note this will be a temporary solution for 1.6 release so we can eliminate the possibility of regressions and allow us to fix a few more games as well with the option.**

Related https://github.com/PCSX2/pcsx2/issues/3211 https://github.com/PCSX2/pcsx2/issues/1339


plugins/GSdx/Renderers/HW/GSRendererHW.cpp
```diff
@@ -462,12 +463,30 @@ void GSRendererHW::ConvertSpriteTextureShuffle(bool& write_ba, bool& read_ba)
 	//
 	// 32bits emulation means we can do the effect once but double the size.
 	// Test cases: Crash Twinsantiy and DBZ BT3
-	int height_delta = m_src->m_valid_rect.height() - m_r.height();
-	// Xenosaga handles the half bottom as an vertex offset instead of a buffer offset which does the effect twice.
-	// Half bottom won't trigger a cache miss that skip the draw because it is still the normal buffer but with a vertices offset.
-	const bool cant_handle_half_bottom_fix = m_game.title == CRC::XenosagaE3;
-	// Test Case: NFS: HP2 splits the effect h:256 and h:192 so 64
-	const bool half_bottom = abs(height_delta) <= 64 && !(m_disable_ts_half_bottom || cant_handle_half_bottom_fix);
+	const int height_delta = m_src->m_valid_rect.height() - m_r.height();
```
tadanokojin: `height_delta` is only used in the `default` and `-1` case. Move there.

PR #3216, opened 2020-01-17
zh_TW Localization

hhhhhojeihsu:
Localization for Traditional Chinese (zh_TW). WIP


locales/zh_TW/pcsx2_Iconized.po
```diff
@@ -3,18 +3,17 @@ msgstr ""
 "Project-Id-Version: PCSX2 1.3.1 1201\n"
```
lightningterror: Version can be bumped to 1.6.0

hhhhhojeihsu: Got it. Will be changed in the next commit.

turtleli:
Sorry about this, but we've only just noticed that we were missing a translatable string.

If possible, can you rebase to latest master branch before resuming work on the translation? Thanks.

hhhhhojeihsu:
> Sorry about this, but we've only just noticed that we were missing a translatable string.
> 
> If possible, can you rebase to latest master branch before resuming work on the translation? Thanks.

No problem. It has been done.

lightningterror:
Some stuff that needs to be translated, Enable Recording tools button as well as popup window, Capture menu along with submenus, some strings in gamefixes, in GS Window fmv aspect ratio text, Vsync text, some strings in speedhacks.

hhhhhojeihsu:
All files are translated. And I've previewed the newly compiled version downloaded from appveyor and went through most of the interface.
And I'm glad the previous translator poured effort into the translation. Makes it easier to translate the rest.

hhhhhojeihsu:
> Everything looks translated, just slightly concerned with the other changes from previous commits that are showing up if they cause any issues during merge.

Rebased to the latest master branch. I think we're all set.

PR #3230, opened 2020-01-31
PCSX2 GUI: Update of German Translation

XXXBold:
Updated the German translation for 1.6.

Some parts of the translation (not limited to the new strings) could need more polishment, but it's usable for now I guess.


locales/de_DE/pcsx2_Iconized.po
```diff
@@ -644,6 +698,11 @@ msgid ""
 "CPU requirements substantially. Lower values will reduce the CPU load "
 "allowing lightweight games to run full speed on weaker CPUs."
 msgstr ""
+"Modifiziert den emulierten Emotion-Enigne CPU-Takt. Höhere Werte können die "
```
RedDevilus: Emotion-Engine

XXXBold: Oops, I'll look at it, thanks for pointing out.

XXXBold: Fixed

lightningterror:
First time wizard window is a bit odd, don't know if it's a string issue changing the width or it's intended to look like that on German.

image

XXXBold:
@lightningterror I've just tested this inside my build-VM for Windows 10. I cannot reproduce this, however, I'm using an older version of PCSX2 1.5.0 dev there. It also seems the window is adjusted, depending on the language, see here as comparison Czech and German:
fl_czech
fl_german
I will try to update PCSX2 to the most recent devbuild, but I guess that's not an issue related to the translation.

lightningterror:
The string that is causing the issue is https://github.com/PCSX2/pcsx2/blob/8d665cc98139278bb1c14f7a350d1949e8c51db0/locales/de_DE/pcsx2_Iconized.po#L78
You will need to split it in several lines, just like the original.

XXXBold:
Okay, I'll look through it, and will also improve some other strings aswell.

XXXBold:
Okay, fixed this, Wizard-UI now looks proper, I've also made some additional changes/improvements.

PR #3235, opened 2020-02-03
Environment variables for use in containers that change path at run time

ferion11:
* Environment variables for use in containers that change path at run time (like AppImage).
* Environment variables:
PCSX2_PLUGIN_DIR, PCSX2_GAMEINDEX_DIR and PCSX2_DOC_DIR
* This code only run for builds that have changed the default path at compilation time first (hence the need for different paths).


ferion11:
It should have worked better. I have to test it a little more. I will make the pull request later again.

PR #3236, opened 2020-02-04
Environment variables to use in containers that change path at run time

ferion11:
* Environment variables to use in containers that change path at run time (reworked). Working and tested at Appimage: https://github.com/ferion11/PCSX2_Appimage/releases/tag/continuous
* Environment variables:
PCSX2_PLUGIN_DIR, PCSX2_GAMEINDEX_DIR and PCSX2_DOC_DIR


ferion11:
Removed PCSX2_DOC_DIR for now (that was a bonus anyway)

pcsx2/gui/AppConfig.cpp
```diff
@@ -173,6 +173,8 @@ namespace PathDefs
 
 	wxDirName GetProgramDataDir()
 	{
+		char * evar_curr = getenv( "PCSX2_GAMEINDEX_DIR" );
+		if ( evar_curr != NULL ) return wxDirName( evar_curr );
```
tadanokojin, line 177: would prefer more this style:
```cpp
char* evar = getenv("PCSX2_GAMEINDEX_DIR");
if(evar)
	return wxDirName(evar);
```

PR #3241, opened 2020-02-07
spu2-x: Update XAudio to 2.9

### LINKED ISSUE #3232: Update to XAudio 2.9
While I was playing with *PreprocessorDefinitions Conditions* to get a graceful fallback in cases you didn't have the DXSDK.. I eventually found out microsoft has silently released 2.9 for W7 and W8.  
https://docs.microsoft.com/en-us/windows/win32/xaudio2/xaudio2-redistributable

I'd even try to make this work myself, but I could never understand [nuget](https://www.nuget.org/packages/Microsoft.XAudio2.Redist/).

MonJamp:
This allows simplification of XAudio code in spu2-x since XAudio 2.9 comes with a redistributable dll which is backwards compatible with Windows 7/8. I have tested the compiled plugin on a Windows 7 VM and can confirm it works. I removed all code which checks for the version of Windows for XAudio since it is no longer needed.

I'm pretty new to contributing to open source so I would like some feedback on some of the changes I made.

The NuGet package manager is needed to download the package which includes the headers and libs. I chose to exclude the `packages` directory and instead let NuGet automatically download it when building spu2-x. This does mean there's an extra step to building PCSX2. Another route I could've taken is to leave the `packages` directory with the XAudio 2.9 package. This would mean distributing Microsoft's dll files and headers which contain a license although I believe they do allow distributing those files without modification.

In order to load the library I had to use this bit of code `xAudio2DLL = LoadLibrary(L".\\plugins\\xaudio2_9redist.dll");`. This hard coded string to a specific directory is inevitable since the NuGet Package Manager downloads binaries to a project's output directory.  I was wondering if there was a better way around this. AFAIK NuGet is pretty rigid and this can't be easily changed.

I'm open to any suggestions and making any changes.

#Resolves #3232 

~~Edit: Looks like CI fails since it doesn't download the NuGet package. So I guess the best way would be to include the package in the repo.~~


mirh:
Maybe I'm wrong, but wouldn't you need to add 
```
<PackageReference Include="Microsoft.XAudio2.Redist" Version="1.2.0" />
```
somewhere (as hinted on the website) for automatic download to happen?
Or maybe it's just [one line](https://www.appveyor.com/blog/2014/03/18/about-nuget-package-restore/) to add in the appveyor config.

tadanokojin:
I don't want all this faffing for an EOL OS.

MonJamp:
> Or maybe it's just [one line](https://www.appveyor.com/blog/2014/03/18/about-nuget-package-restore/) to add in the appveyor config.

Nuget restore works on appveyor now.

> I don't want all this faffing for an EOL OS.

Using XAudio 2.9 allows backwards compatibility with both Windows 7 and 8 while simplifying code for spu2-x. It is annoying dealing with nuget but the process of getting the packages for compiling happens in the background. 

It would be nice to have the `xuadio2_9redis.dll` next to the executable file instead of in the plugins folder but I'm having difficult figuring out how to do that since it seems we don't get control over where a nuget package installs the binaries.

tadanokojin:
> It is annoying dealing with nuget but the process of getting the packages for compiling happens in the background.

Yes it's annoying. Which is why I don't want to deal with it now or in the future.
You can just update it to 2.8 and we can move on from win7. We'll end up killing it one way or another anyway.

MonJamp:
Fair enough, is it safe to assume Windows 7 won't be supported for 1.8 release? If that's the case I can still simplify the code without dealing with nuget. Or I can include the contents of the nuget package into the 3rdparty folder that way we still don't have to deal with nuget and still use the same code for win7/8/10.

mirh:
*Sigh*, I could get 2 seconds on the first start are a downer... But doesn't xz init take already even more than that?

> It would be nice to have the xuadio2_9redis.dll next to the executable file instead of in the plugins folder but I'm having difficult figuring out how to do that since it seems we don't get control over where a nuget package installs the binaries.

[[1]](https://stackoverflow.com/questions/4092759/is-it-possible-to-change-the-location-of-packages-for-nuget)? [[2]](https://stackoverflow.com/questions/7676640/how-do-i-specify-the-directory-where-nuget-packages-are-installed)?
Then putting aside you could hardcode the path, you can just have the "package part" of the build script move it.

tadanokojin:
@MonJamp we plan to drop it following 1.6.0 which is the upcoming release. So it'll get dropped in 1.7.0 and 1.8.0 (won't be out for some time) will not support win7.

Sorry you put this effort in just for me to tell you to kill it.

MonJamp:
> Sorry you put this effort in just for me to tell you to kill it.

lol it's fine, I had a feeling this pr would get killed due to nuget. The changes needed to drop XAudio 2.7 (dropping win7 support) are basically the same anyways.

tadanokojin:
You can still make those changes if you feel inclined to do so btw.
We'll merge it after 1.6.0.

Squall-Leonhart:
Sorry, this is a foolish direction and turtle needs to have his perms revoked.

PR #3244, opened 2020-02-09
spu2-x: Drop XAudio 2.7

MonJamp:
Simplifies code for spu2-x while removing support for Windows 7


seta-san:
don't forget in WinConfig.h you should change to this to specifically target xaudio2.9 
"#ifndef _WIN32_WINNT
#define _WIN32_WINNT 0x0A00
#endif

turtleli:
This PR is for dropping XAudio 2.7. Targeting XAudio 2.9 has no benefit if we're not using its features.

seta-san:
> is PR is for dropping XAudio 2.7. Targeting XAudio
either you target 2.8 or 2.9 after dropping 2.7. 

https://docs.microsoft.com/en-us/windows/win32/xaudio2/xaudio2-redistributable

Microsoft provides a NuGet package for 2.9 that can be ran on windows 7 sp1 and windows 8. There's really no good reason for stopping off at 2.8 when we can simply target 2.9 and still have compatibility with all supported Operating Systems. The 2.9 dll is still supported and still receives updates and bug fixes from Microsoft.

turtleli:
The plan is to drop Windows 7 support after 1.6 is released. Please see #3241 for previous discussion.

seta-san:
> The plan is to drop Windows 7 support after 1.6 is released. Please see #3241 for previous discussion.


cool to see you already had the discussion. i get not wanting to deal with a nuget but i still think targeting 2.9 is superior since Microsoft is still letting people with windows 8 upgrade to 10 for free although they aren't really advertising it. still, if we're targeting 2.8 the line in winconfig.h should be 0x602

turtleli:
There's a line in SndOut_XAudio2.cpp that sets _WIN32_WINNT to 0x602 but yes, what you suggested also works and might be a better way of doing it.

seta-san:
> N32_WINNT to 0x602 but yes, what you suggested also works and might be a better way of doing it.

thanks for listening. This has been a pet issue of mine for a little bit.
https://forums.pcsx2.net/Thread-how-to-get-SPU2-to-compile-on-visual-studio-2019-and-to-modern-xaudio
:)

MonJamp:
Using the ListDlls utility with this build I see that both xaudio 2.8 and xaudio 2.9 dlls are mapped on Windows 10. I assume that if W10 is detected there is a forward declaration to load xaudio 2.9 dll even if the exe is targeting 2.8. And just for fun I opened the xaudio2_8.dll found on W10 in a disassembler and confirm that all it's doing is forwarding function calls to xaudio2_9.dll

So there's no harm targeting xaudio 2.8 and it allows us to keep supporting W8. Although maybe I should add a comment to target xaudio 2.9 when W8 support is dropped.

turtleli:
SndOut_XAudio2_27.cpp and SndOut_XAudio2.inl also need to be removed from the SPU2-X CMakeLists.txt file.

turtleli:
Thanks.

Anths2016:
How can I use new plugin with Win7?

refractionpcsx2:
Use the one that comes with 1.6.0

kenshen112:
@Anths2016 You don't. We're not supporting Windows 7 anymore. You can either use 1.6's plugin or upgrade

Anths2016:
Are you affiliated with Microsoft? Otherwise I don't understand the reason ...

refractionpcsx2:
We are not no. However we have been using the DirectX SDK since 2010 and that was the only thing left using it, everything has been integrated in to the Windows SDK since Windows 8, and now Windows 7 is end of life we figured it time to remove the legacy tools.

We're not stopping you using the current release (or just the audio plugin with the newer builds) but by the time 1.8.0 is released, Windows 7 will be end of life by about 4-5 years.

KrossX:
You can still use the new plugin though, is just that old XAudio backend won't be available. You can use the other backends instead. However, I don't remember any fix going into SPU so using old plugin should be just fine and you wouldn't be missing anything.

tadanokojin:
Any further discussion about this the hows and whys of this can happen at https://forums.pcsx2.net/
Don't need to fill the poor guys PR.

PR #3248, opened 2020-02-11
Remove Arabic on Windows for release

MrCK1:


avih:
Is it not maintained?

MrCK1:
@avih The issue was that the formatting is broken on Windows due to Wx. It won't be present in 1.6 for Windows users only. Anyways, I will just exclude the language from being compiled into the installer, making this PR pointless.

Also you're welcome to join the rest of the devs on Discord if that's your thing :)
https://discord.gg/cCrAHpu

avih:
Thanks for the info. I think it belongs at the commit message though. It helps if the message explain also why and not only what.

PR #3255, opened 2020-02-15
Added VU patch for Powerdrome PAL

refractionpcsx2:
This patches some values in Powerdromes VU microprogram, which restores the 3d models of the characters.
The game uses the extended PS2 float range to do some culling calculations, so I have brung them down to be usable by x86, no other way to really fix this game.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -11994,8 +11994,8 @@ Name   = Powerdrome
 Region = PAL-M5
 Compat = 4
 [patches = 52085DF1]
-	//Game takes advantage of the VU's extended float range.
-	//This limits it so x86 can handle it.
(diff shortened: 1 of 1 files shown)


bin/GameIndex.dbf
```diff
@@ -11993,6 +11993,13 @@ Serial = SLES-52418
 Name   = Powerdrome
 Region = PAL-M5
 Compat = 4
+[patches = 52085DF1]
+	//Game takes advantage of the VU's extended float range
+	//This limits it so x86 can handle it
+	comment=VU Float patch by refraction
```
lightningterror: `author = refraction` would be better
Comments can be slightly better formatted, add spaces after // and periods at end (ocd satisfaction).

PR #3263, opened 2020-02-23
gsdx: mirror RGBAQ to 0x11

### LINKED ISSUE #2427: [GSDx] Ridge Racer V [SLUS-20002] Missing Shadows - All Renders
***PCSX2 version:***
All Versions

- **Notes**:
All Renders.

***PCSX2 options:***
No options have effect.

- **Note**: 

***Plugins used:***
GSdx32-SSE4.dll [GSdx 20180503003203 (MSVC 19.00 SSE4.1/AVX) 1.1.0]

- **Note**: 
DX11HW/SW

Car shadows alternate at a steady interval from white to invisible

https://i.gyazo.com/0d1941ba6e881329dbedfc43cff66dc6.png

https://i.gyazo.com/d153d51d23f66e461e695ddc0bcb35ab.png

GS Dumps avail here:
https://forums.pcsx2.net/Thread-Vector-Units-Bug-Report-Ridge-Racer-V-PAL-NTSC

Start the game. Enter a race. Observe lack of shadows.

I checked latest annual builds for 2011 thru 2017 (missing a 2015 build)
Car Shadows have 'Never worked' period. All renders. 

***PC specifications:***
CPU: AMD 4.0GHz X4 GPU: R9 270X 2GB OS: Win7 x64

tadanokojin:
fixes #2427

### DIFF
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -329,6 +329,7 @@ void GSState::ResetHandlers()
 
 	m_fpGIFRegHandlers[GIF_A_D_REG_PRIM] = &GSState::GIFRegHandlerPRIM;
 	m_fpGIFRegHandlers[GIF_A_D_REG_RGBAQ] = &GSState::GIFRegHandlerRGBAQ;
+	m_fpGIFRegHandlers[GIF_A_D_REG_RGBAQ + 0x10] = &GSState::GIFRegHandlerRGBAQ;
 	m_fpGIFRegHandlers[GIF_A_D_REG_ST] = &GSState::GIFRegHandlerST;
 	m_fpGIFRegHandlers[GIF_A_D_REG_UV] = m_userhacks_wildhack ? &GSState::GIFRegHandlerUV_Hack : &GSState::GIFRegHandlerUV;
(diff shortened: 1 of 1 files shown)


ramapcsx2:
Is it likely that the other regs are mirrored and used in the same way?
If so, we probably don't want countless duplicate entries right away, but a warning on use of a potentially mirrored reg would be good.

refractionpcsx2:
It needs to be tested on a ps2, something we plan on doing at some point

tadanokojin:
You've got 0x10, 0x12, and 0x13 available. Could account for PRIM, ST and UV.

ramapcsx2:
Just those?
I would just add them pro-forma in that case.
If you've got a call to the register handler with that index, what else should happen?
It can either mirror, or drop the call (on hardware). And I mostly doubt it should be dropped.

refractionpcsx2:
I don't think it's a good idea to just guess add them, in case that is wrong, or there's some weird mapping, it would be better to run tests first.

ramapcsx2:
Yeaahh.. Well, as long as it's not forgotten :p 
This PR itself should be good though, right?

gregory38:
Good job.

PR #3265, opened 2020-02-26
Change order of "Video (GS)" submenu

### LINKED ISSUE #3262: [Feature Request]  GUI (ConfigVideoGS) Change of the list.
VideoGsGui

I would like to change the ordering in the GUI. As a developer it makes sense.
As a normal user it is consistent with the others like Controllers. It looks more streamlined to me.

MonJamp:
Makes UI more consistent as per #3262

Close #3262


lightningterror:
Could be nice to get more feedback on this from others.

MonJamp:
I think it's a good change. Don't have much more to say than what's said on #3262. It makes the UI more consistent which is important when designing UI.

RedDevilus:
I requested it because not only does its consistent but i also change settings quite alot and i want to quickly go to plugin settings due to muscle memory. Also reduces people going into the wrong setting when we ask them to go into the GSDX. I would assume this would reduce frustation to endusers.

PR #3267, opened 2020-02-27
Locales: Update Lithuanian Translations lt_LT

xddxd:


RedDevilus:
Perhaps change the title to:

Locales: Update Lithuanian Translations lt_LT

Otherwise seems good.

PR #3091, opened 2019-08-24
GSdx-d3d11: Some HLSL shader work

KrossX:
Make the D3D shader closer to OpenGL's one, possibly fixing some bugs during previous shader port. Also, quite possibly adding new bugs. Yay!

Improves bloom/blur in DBZ BT2, SOTC matching OpenGL visuals.
Improves https://github.com/PCSX2/pcsx2/issues/2864 matching OpenGL visuals.

PR #3268, opened 2020-02-28
gsdx-d3d11: Truncate fog in tfx.fx shader.

lightningterror:
The change should've been pushed in #3091

Fixes regression with fog rendering.


lightningterror:
Eye candy for anyone interested, Castlevania testing:

master
image

pr
image

opengl
image

Issue #2226, opened 2017-12-20
Twisted Metal Head-on [Extra Twisted Edition] : broken fire effect + incorrect resolution rendering


atomic83GitHub:
***PCSX2 version:***
Last 1.5dev

***PCSX2 options:***
All default, speedhacks disabled. Core hacks and Clamp/round mode don't help here.

***Plugins used:***
GSDX software and Hardware OpenGL

***Description of the issue:***
The game Twisted Metal Head-on [Extra Twisted Edition] have some issue in PCSX2:

-Broken fire effect (GSDX Software/Hardware and ZZogl show the same issue (core issue?))
![image](https://user-images.githubusercontent.com/26351275/34200289-c796e694-e570-11e7-8912-74ca15e09b0f.png)

Incorrect FMV rendering: https://github.com/PCSX2/pcsx2/issues/961
![image](https://user-images.githubusercontent.com/26351275/34200351-133ec706-e571-11e7-915b-aeff31a45bde.png)

***How to reproduce the issue:***
Load PCSX2 with Twisted Metal Head-on [Extra Twisted Edition]

***Not a regression***

***PC specifications:***
AMD fx8350  GTX 780Ti  Windows 10 64 Bit

***GS/Blockdump***
https://drive.google.com/file/d/1YdhUuh6iLExrV8ePawTN_E60njJWCTKF/view?usp=sharing

MrCK1:
The fire looks fine to me? Did you try Full blending accuracy, I don't see the issue.

As for the rendering resolution, 720x480 is Progressive Scan aka 480P. I would only be concerned If the game doesn't support it, or if there were rendering glitches as a result.

atomic83GitHub:
Not all fire effect are broken, some are fine, some other are completly broken as you can see:
![image](https://user-images.githubusercontent.com/26351275/34222956-c20c8f60-e5bd-11e7-8ec4-27e911855f39.png)

And for the resolution issue I was talking about the right of the screen wich is black in FMV's even in software mode.

Ghaleon777:
Nobody seems to completely fix this awesome game, it also as got some weird black glitches (textures) on most cars when battling. For example look at this weird glitch..
![new bitmap image](https://user-images.githubusercontent.com/36754054/36578418-dfcab2ba-183b-11e8-9ff5-e758aa37de74.png)

Can it be fixed please?

MrCK1:
[...]

PR #3276, opened 2020-03-07
GameDB: Add FMV fix for Twisted Metal - Head-On [Extra Twisted Edition]

Aced14:
This game's FMVs are meant to be letterboxed (640x368 with 40 pixel top/bottom bars). Without this patch, FMVs are rendered at 720x480 with bars on 3 sides (40 pixels at the top, 80 on the right, 72 at the bottom).

Related to #961 and #2226.

**Screenshots (GSdx SW+HW)...**

**Before:**
gsdx_20200301211148

**After:**
gsdx_20200301212003


lightningterror:
It was discussed that a proper solution will be better instead of a patch.

PR #3292, opened 2020-03-19
pcsx2-gui: Fix DPI scaling for speedhacks window when moving the sliders

lightningterror:
We can have it for either 1.6 or 1.8. Leaving it to the rest to cast their vote.

Testing was done on 1080p, 150% scale.
Before:
image

After:
image


RedDevilus:
1.6 release. This is very annoying if u want to change it. (Although non-critical)

Issue #2965, opened 2019-05-18
TAS recording functionality not working on Linux


thegamingrestoration:
***PCSX2 version:***
PCSX2 1.5.0-20190517234655 - compiled on May 18 2019

***PCSX2 options:***
Preset: Safe (Default)

**Relevant**: https://github.com/PCSX2/pcsx2/pull/2412

***Description of the issue:***
I've been testing the new TAS recording options on Windows, and I was pleasantly surprised by how good they work; but when I tried them on Linux, the emulator couldn't start the game. It just froze before even initializing the BIOS:

```
PCSX2 1.5.0-20190517234655 - compiled on May 18 2019
Savestate version: 0x9a0e0000

Host Machine Init:
	Operating System =  Linux 5.0.11-24-tkg-pds x86_64
	Physical RAM     =  5864 MB
	CPU name         =  Intel(R) Core(TM) i7-2600 CPU @ 3.40GHz
	Vendor/Model     =  GenuineIntel (stepping 07)
	CPU speed        =  3.389 ghz (8 logical threads)
	x86PType         =  Standard OEM
	x86Flags         =  bfebfbff 1fbae3ff
	x86EFlags        =  28100000

x86 Features Detected:
	SSE2.. SSE3.. SSSE3.. SSE4.1.. SSE4.2.. AVX

Installing POSIX SIGSEGV handler...
Reserving memory for recompilers...

Loading plugins from /usr/lib32/pcsx2...
(GameDB) 9771 games on record (loaded in 143ms)
	Bound   GS: libGSdx.so [GSdx (GCC 8.3.0 AVX/AVX) 1.1.0]
	Bound  PAD: libonepad-legacy.so [OnePAD 20190517234655 1.3.0]
	Bound SPU2: libspu2x-2.0.0.so [SPU2-X 2.0.0]
	Bound CDVD: libCDVDnull.so [CDVDnull Driver 20190517234655 0.6.0]
	Bound  USB: libUSBnull-0.7.0.so [USBnull Driver 20190517234655 0.7.0]
	Bound   FW: libFWnull-0.7.0.so [FWnull Driver 20190517234655 0.7.0]
	Bound DEV9: libdev9null-0.5.0.so [DEV9null Driver 20190517234655 0.5.0]
Plugins loaded successfully.


** (PCSX2:15431): WARNING **: 19:45:20.098: Invalid borders specified for theme pixmap:
[...]

PR #3293, opened 2020-03-20
Resolve Recording Issues for Linux / OnePad users

xTVaser:
This is primarily to resolve https://github.com/PCSX2/pcsx2/issues/2965

## Issues Resolved
- OnePad now will return a populated `keyEvent` on `PADkeyEvent()`
  - This not only resolves the usage by the recording tools, but also the usage at the end of the main Vsync - https://github.com/PCSX2/pcsx2/blob/79db01d7ed01a4f3de06a8d8dc5c5e7099ef9d2e/pcsx2/gui/AppMain.cpp#L579
  - This means that emulation should no longer hang when a recording is started and subsequently: when emulation is paused, as key-bindings will be propagated and handled.
- Added some additional capitalized keybindings for linux.  Without these, the key-bindings would frequently be ignored.  The capitalized variant is what is typically caught, and has a distinct key-code to the lower-case variant.
- Recording files created on linux do not have the extension properly appended to them due to wxWidget's wxGTK implementation (if I understood this ticket correctly, this is expected) https://trac.wxwidgets.org/ticket/15285
- Some small refactors and renaming around these changes.

## Testing Done
- On Linux > onepad-legacy
  - Able to pause/resume emulation outside of an active recording
  - Able to make a new recording, resume the game, play for a bit, stop, and playback successfully.
  - Recording file properly saved
- On Windows > LilyPad
  - Everything listed above


InfamousKnight:
This was made 2 months ago, hope nobody forgot about this.

plugins/onepad/keyboard.cpp
```diff
@@ -117,7 +117,7 @@ static bool s_Shift = false;
 static unsigned int s_previous_mouse_x = 0;
 static unsigned int s_previous_mouse_y = 0;
 
-static void AnalyzeKeyEvent(keyEvent &evt)
+extern void AnalyzeKeyEvent(keyEvent &evt)
```
arcum42, line 120: Pretty sure the extern shouldn't be here.

arcum42:
You know, I fired this up with both onepad and onepad-legacy, and in both cases, it still seemed to freeze if I started recording...

xTVaser:
> You know, I fired this up with both onepad and onepad-legacy, and in both cases, it still seemed to freeze if I started recording...

It's expected for the game to pause after you start a new recording.  The issue was on onepad it wasn't possible to resume the emulation via the key-binding.  Is that still the case, if so I'll take another look again? @arcum42

arcum42:
Alright, now that I have a better idea of how to use it, it seems to work.

From what I can tell, it pauses it without telling you that's what it's doing or how to unpause it when you start recording, you have to hit shift-P,and then it actually says it's paused and hitting shift-P again unpauses things. I was then able to record what i was doing and then play it back.

We can probably merge this to get linux working, though it'd be nice to have things be more intuitive.

xTVaser:
@arcum42 
> We can probably merge this to get linux working, though it'd be nice to have things be more intuitive.

Agreed, probably some better console logs / onscreen logs would help (in the works, waiting on my other PRs to merge).  But also I _think_ `<PAUSED>` shows up in the title as you'd expect and if not, that may already be fixed in https://github.com/PCSX2/pcsx2/pull/3530

Another way to make things clearer would be to create menu options in the `Recording` menu for pause/frame-advance etc with the keyboard shortcuts, right now they aren't user-facing like the debug window key-bindings.

arcum42:
Having Pause show up was working after the initial pause, but not for the first one, IIRC. And yeah, a console log mentioning to press shift-P would help, as would menu options.

PR #3299, opened 2020-03-24
Fix recording related hang when PAD plugin is open/closed while game is paused

xTVaser:
I discovered this hang the other day, and I'm hoping it is the problem that has been mentioned which rarely occurs on Windows.  I figure since changing controller settings is probably one of the more frequent actions, it's likely someone has ran into this.

I added a fairly verbose comment explaining the problem, to keep it brief: if the `CoreThread` is paused prior to opening the PAD plugin settings then when they are closed the PAD will not re-open.  Since the PAD is used to resume the `CoreThread`, we hang.  

I believe this fix will prevent the recording pause behaviour from impacting any event, plugin or otherwise, by placing the emulator into it's expected state (running).

---

I'm not 100% sure if this is the best way to solve this either, at first I wanted to do the change here https://github.com/PCSX2/pcsx2/blob/master/pcsx2/gui/AppCorePlugins.cpp#L552 but the `CoreThread` is already expected to be paused by this point via opening the plugin menu itself.  So I looked for any location where `CoreThread.Pause()` is called, and this is the only spot other than where it's used in the recording code.

It also looks a little strange, considering I'm resuming only to immediately pause again.  But I'm thinking this is either because I don't fully understand how this works, _or_ related to how the recording uses `CoreThread.PauseSelf()` instead of `CoreThread.Pause()` and resuming puts things back into a workable state?


sonicfind:
I don't know exactly what causes this, but I'm getting some inconsistent behavior with opening the PAD plugin settings. Normally, when opening the PAD plugin window, you see this in the logs:
image
However, sometimes, with this change, you end up with this happening causing the plugin window not to open:
image
It will happen only when the emulation is paused and will happen randomly, but when it happens, it will usually happen multiple times in a row as long as its in the paused state - including frame advancement. 
I can only guess that setting emulation to resume before opening the window is causing some mistimings in the threading, so when a plugin window takes a bit longer to load, it ends up calling the applicable "exit window" function before opening.

A similar issues appeared when opening the USB plugin window (and audio), however in those cases, the window still opened even though the logs displayed out of order:
image
image

xTVaser:
Hmm yeah that does seem odd but your theory sounds believable to me.  With most of these being async events (iirc), and there being no easy way to await for them, I wouldn't be surprised if sometimes the plugin settings opening vs the emulation resuming mis-align.

That's kinda part of the reason why I felt like this was a hack at best, but wasn't sure of a better way to do it as this area of pcsx2 is pretty foreign to me.

After thinking on this again here is my next best idea: rather than trying to do this gracefully behind the scenes.  If the user tries to open the PAD plugin settings while emulation is paused (due to recording).  We could just flat our forbid the action with an alert message of some kind explaining to resume first?  Possibly just disable the menu options (but this feels less explicit).  I'm pretty sure this would be easy to do as I'm sure it would be easy to hook into the menu options.

Thoughts?

My hope is that eventually once the controller inputs aren't reliant on plugins and it's been integrated into the core, these problems will be less significant (we won't need to rely on lilypad/etc to catch the key-bindings to resume emulation).

sonicfind:
This could become a problem for people who need to switch controllers on the fly but don't have the luxury of being able to resume the emulation.
Now funnily enough, I did some digging in my own old WIP stuff. I actually found my own version of a "fix" for this issue involving pausing and resuming in the plugin_click function in Main_Menu_Clicks. However, after testing, I noticed that while it did solve the issue with PAD, it caused an issue with the other plugins - requiring two presses of Shift+P to truly resume. Frame advance was fine however. Comparatively, this may be a better alternative to try out.
```
void MainEmuFrame::Menu_ConfigPlugin_Click(wxCommandEvent &event)
{
	const int eventId = event.GetId() - MenuId_PluginBase_Settings;

	PluginsEnum_t pid = (PluginsEnum_t)(eventId / PluginMenuId_Interval);

	// Don't try to call the Patches config dialog until we write one.
	if (event.GetId() == MenuId_Config_Patches) return;

	if( !pxAssertDev( (eventId >= 0) || (pid < PluginId_Count), "Invalid plugin identifier passed to ConfigPlugin event handler." ) ) return;

	wxWindowDisabler disabler;
	ScopedCoreThreadPause paused_core( new SysExecEvent_SaveSinglePlugin(pid) );
#ifndef DISABLE_RECORDING
	if (g_RecordingControls.IsEmulationAndRecordingPaused())
	{
		g_RecordingControls.Unpause();
		GetCorePlugins().Configure(pid);
		g_RecordingControls.Pause();
	}
	else
		GetCorePlugins().Configure(pid);
#else
	GetCorePlugins().Configure(pid);
#endif // !DISABLE_RECORDING
}
```
It would require adding RecordingControls.h as an include however. Granted, that may be inevitable.

xTVaser:
> This could become a problem for people who need to switch controllers on the fly but don't have the luxury of being able to resume the emulation.

Hmm, that's an interesting use-case that I can't really wrap my head around.  For example - someone needs to swap controllers on a certain frame, and so they want to keep it paused?  I don't think these changes can 100% guarantee that would be the case since we are calling `Unpause` after all.  

I tried your solution and it works, but as I suspected, it seems as if frames are advanced after the plugin settings window is closed, so I'm not sure how useful this is for users that can't resume emulation.  I avoided the issues you mentioned by scoping this behaviour to only the PAD plugin.

sonicfind:
I guess that example wasn't exactly the greatest, ha. A more likely situation would probably be a mismapped button or something that needed to be fix. But it's whatevs.

And yea, that was the other problem I was worried about. Since Pause() waits for a vsync, it is definitely possible that it may skip some frames (I was hoping to test pauseImmediately() from the other branch to see if that gave better results, but I haven't gotten to that yet).
And can I just say: _I'm dumb as hell for not noticed PluginId_PAD was a enum that could be used._

sonicfind:
Update: _pauseImmediately() didn't work in this case for this situation._ Pause() is the best we get for now.
Ended up pausing sooo immediately that the closing plugin window resumed the emulation.

xTVaser:
## Testing Steps:
1. Verify that Recording Tools are enabled (PCSX2 > Game Settings)
2. Pause the game via Shift+P
3. Open the PAD plugin settings
4. Close the PAD plugin settings
5. You should be able to unpause the game with Shift+P.
6. Repeat steps 2-5 by using frame-advance (Space) instead

If you repeat steps 2-6 on master / a build not from this PR.  You will _not_ be able to unpause after closing the plugin settings.

sonicfind (approved):
While there exists an issue where pausing using the frame advance action before opening the PAD plugin requires you to press Shift+P twice afterwards to resume, the major hang this was intended to fix has been addressed.
If you _can_ manage to fix that additional, but much more minor, issue, that would be great; but it does feel very acceptable at this stage.

xTVaser:
Since it's around frame-advancing.  I'd rather keep track of that issue and see if it's resolved with https://github.com/PCSX2/pcsx2/pull/3530 If it still persists, we can follow up.  As you say, fixing the hang is the most important thing.

PR #3304, opened 2020-04-01
GSdx-gui: Remove custom resolution from the windows gui

lightningterror:
Leave it as an ini option only.
People should never use custom resolution as it causes many issues and
many features aren't supported with it.
On linux option was removed a long time ago.


vsub:
Why not just change the name from
Custom (Not Recommended)
to
Custom (Can Case Problems)
That will probably have a better effect at scaring them away from it and give them some clue why they are getting weird graphics

tadanokojin:
People don't read "Not Recommended" and you think they will read "Can Cause Problems"? It's not that they don't get what we mean, they ignore it. I've had countless conversations with users to this effect:


Average users should be scared away from it. It doesn't work properly.

RedDevilus:
> 
> 
> Why not just change the name from
> Custom (Not Recommended)
> to
> Custom (Can Case Problems)
> That will probably have a better effect at scaring them away from it and give them some clue why they are getting weird graphics

End users: youtube + click,click,click

Not recommended is obvious what it means to all.
Can cause problems u can apply to more than that and is too vague.

pixelherodev:
If the problem is that they're not reading it, why not just add a popup window that says in bolded text that "This option can and probably _will_ break your games. Are you sure you want your games broken?" That seems to be to be the best compromise; if a user really wants to do it, they can, but it reduces the incidence rate of "Yeah I didn't see the warning."

RedDevilus:
No offence to end users but most wont read it anyway no matter the methods. Cutting it out of GUI and stay as ini or HW hacks are the only solutions.

RedDevilus:
My personal opinion was to move it into a HW hack. That way it can be easily used for screenshots and the low amounts of games that truely need it. Best compromise to me and u solve the problem that 

1: People dont pick custom resolution unless they want it or at least reduce the complaints for everyone. 

2 Sure you can call it hacky but its also handy tontest something quickly and not gimmicky.

Otherwise besides screenshots that arent stretched and 4k and fix for an amount of games there is nothing i can think why people should use it.

For most people just use native or 3x native or the multipliers. I still see the potential dont take my previous comment to be one where im for removing as my favorite option.

PlasmaLight:
What about removing it only from the 1.6 release but keeping it in further dev builds? If the real problem is that people are using it without being aware of the problems it can cause... I guess most of these people are also unaware of the dev builds.

lightningterror:
> My personal opinion was to move it into a HW hack.

That will require work and it's just more burden to maintain a broken feature.

> I feel just because a portion of the users are dumb doesn't make it reasonable to remove extra functionality from the intermediate and above users who might find a need for it.

They can use either an older version (1.6) or edit the ini files to use it.

vsub:
I can imagine how guides start appearing how to change the resolution from the ini

"How to change the resolution,it will not allow me to set my desktop resolution"...like setting your desktop resolution in the custom resolution helps in any way

tadanokojin:
> I feel just because a portion of the users are dumb doesn't make it reasonable to remove extra functionality from the intermediate and above users who might find a need for it.

The cases you mention are few and far between. The case of a user joining and asking us why their games is broken because they saw a video is far more common. This isn't a new idea. We have plenty of secret ini option which have some usefulness but are dangerous for end users. Is it unreasonable for those too?

> I can imagine how guides start appearing how to change the resolution from the ini

There are youtube guides for other secret inis too. The thing is people are less included to follow that because it's a hassle for them to find and edit an ini by hand. This is why Akash is making his point, because he knows this is the case and he doesn't want to sacrifice those few and far between users for the good of the average user. Thing is those few and far between users who know what they are doing *are* inclined to edit an ini and no one is suggesting we remove it full stop.

PSI-Rockin:
Advanced options meant for advanced users should be hidden away from the average user. The vast majority of people only need the resolution multiplier, and there's no reason for an option few people need that causes problems to be right next to an option most people need.

RedDevilus:
I agree that the option should be removed but from what i understand wxwidgets makes it hard to move it to HW hacks and besides which will slow down development. I don't want it to become the next DX9 that isn't removed. If you compare the situation this is very comparable. On the short run the feelings of a few is hurt but i still believe its better in the long run. 

Perhaps even adjusting the code for the screenshot icon from emulation settings or even a way to make a screenshot the way some games + steam do it. If you have any suggestions you can always say them.

image

Besides this is aimed for 1.8 release and not 1.6 release.
So 1.6 stable is still with W7 support with custom resolution.

Dreadmoth:
> 
> > The highest multiplier I can use is 12 (6144x5376 in Ratchet and Clank, for example). Anything higher than that causes a Visual C++ runtime error. 
> 
> From: https://forums.pcsx2.net/Thread-Removal-of-Custom-resolution-from-GSdx?pid=537357#pid537357

This issue seems to have been fixed since that was posted - I tried up to 32x Native without crashes or error messages. Usable maximum resolution appears to be limited by VRAM or PCSX2's ability to allocate it (exceeding ~7GB usage on my 1070 Ti 8GB results in missing graphics or a black screen).
[Ratchet & Clank at 28x Native](https://user-images.githubusercontent.com/8492684/78632069-607cfc80-78e1-11ea-9337-ea7b22fe681e.png)
This screenshot shows the limitations of the current downscaling implementation - there is almost as much aliasing as a game running at native monitor resolution without AA. A comparison showing how it might look if a more effective algorithm could be implemented: https://imgsli.com/MTQyMzU

PR #3306, opened 2020-04-08
VU: Log the index of VU microprograms upon execution

chaoticgd:
microVU supports [dumping out the disassembly of VU microprograms](https://github.com/PCSX2/pcsx2/blob/49e316e488c224bf50df0ee1e3bfbc55fa738343/pcsx2/x86/microVU.h#L17) with file names in the form "microVU1 prog - %d.html" where %d is the index of the microprogram.

The only time I've been able to find this index being printed out is when the microprogram is recompiled, which seems hard to correlate with e.g. the execution of a VIF MPG instruction. To get around this, I've added log messages to print out the index of a microprogram when it is executed.

It seems much easier to figure out which dumped microprogram is which now, unless I'm just blind and it's already logging everything I need, in which case ignore this PR.

I've tested this with a VU1 microprogram and it seems to be printing out the right numbers.

### DIFF
--- a/pcsx2/x86/microVU.cpp
+++ b/pcsx2/x86/microVU.cpp
@@ -350,7 +350,9 @@ void recMicroVU1::Reset() {
 
 void recMicroVU0::Execute(u32 cycles) {
 	pxAssert(m_Reserved); // please allocate me first! :|
-
+	
+	VUM_LOG("VU0 Microprogram Index: %d", microVU0.prog.cur->idx);
+	
 	if(!(VU0.VI[REG_VPU_STAT].UL & 1)) return;
 
 	// Sometimes games spin on vu0, so be careful with this value
@@ -366,7 +368,9 @@ void recMicroVU0::Execute(u32 cycles) {
 }
 void recMicroVU1::Execute(u32 cycles) {
 	pxAssert(m_Reserved); // please allocate me first! :|
-
+	
+	VUM_LOG("VU1 Microprogram Index: %d", microVU1.prog.cur->idx);
+	
 	if (!THREAD_VU1) {
 		if(!(VU0.VI[REG_VPU_STAT].UL & 0x100)) return;
 	}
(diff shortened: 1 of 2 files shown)


refractionpcsx2:
Hmm, does that variable get updated before the program is recompiled? I didn't think it did, so not sure if you can use that.  Also the game will use multiple micro programs per frame, so unless you know what you're looking for, you kinda need to look at all the programs.

chaoticgd:
I was trying to figure out what a specific VIF chain, loaded from a known address was doing, so my logic was I could search emuLog.txt for that address, see where the VIF instructions were being executed, and see what microprogram was run after that. Editing said microprogram seemed to modify the object I expected it to. That said, I'm not very familiar with the recompiler, so if you're not sure about the change it's probably best just to close this PR.

refractionpcsx2:
Well I'll leave it for now, I suggest you double check where that variable is changed to make sure your log message appears in the right place, I have a feeling that updates somewhere else, but I can't be certain.  If it's fine then that's okay.

chaoticgd:
I've had another look at it and it seems that the variable is set [right after the microProgram struct is malloc'd](https://github.com/PCSX2/pcsx2/blob/49e316e488c224bf50df0ee1e3bfbc55fa738343/pcsx2/x86/microVU.cpp#L175-L177). I can't find any other code that updates its value. I added the log message right before it seemingly jumps into recompiled code, so I'm not sure how it can be too early. That said, I've modified the position of the log messages slightly.

refractionpcsx2:
Okay cool, as long as its already set by that point, should be fine :)

chaoticgd:
Actually, cur seems to be set by mVUsearchProg, which is called by an instruction generated in mVUdispatcherAB. Since I now have no idea as to whether this is right, I think I'll just close this.

chaoticgd:
This seems right now. mVU.prog.cur is set by the call right before the VUM_LOG.

PR #1134, opened 2016-01-20
Recompiler memory optimization

gregory38:
Open PR to open a discussion on direct memory access. It is only a proof of concept. Various issues likely remain. It is related to #1019 

Part that need more work:
- Remove completely lazy allocation on x86 buffer (otherwise it might generate nested SIGSEGV)
- Use a 64KB mirror below the start address instead of 32KB
- Redo (or drop) the x86emitter immediate optimization
- Windows support?
- Requires BMI2 (haswell and later). So far code rely on ifdef.
- Consume 256MB of Memory (large aware allow to us get 4GB).

PR #3308, opened 2020-04-14
Make the interpreter work on x86_64 unix

tellowkrinkle:
Unless I'm missing something, there doesn't seem to be a mode for vif emulation that doesn't use some sort of dynamic codegen
This makes the parts needed by it work on x86_64 unix (or at least work well enough to load the BIOS screen and Persona 3 FES's title screen, haven't tested past that yet)
## Changes:
- `__attribute__((fastcall))` does nothing on x86_64, swap it with `__attribute__((ms_abi))` instead
- In x86_64, only `mov`s take 64-bit immediates, all other instructions take 32-bit immediates even when targeting 64-bit registers.  I feel like there should be a better workaround than "just say whether it's a `mov` or not as an extra function argument" but at the moment I can't think of anything better.


arcum42:
I expect you and @pixelherodev may want to get together on this one...

pixelherodev:
Awesome!

Definitely going to take a look at this one, thanks for the mention!

pixelherodev:
Immediate comment.

This shouldn't be needed:
```
#if defined(__x86_64__) || defined(_M_AMD64)
#define __M_X86_64
#endif
```

Unless it's a change *I* made, BuildParameters.cmake should already do that.

pixelherodev:
So if I'm understanding this correctly, this just patches the VIF recompiler to function on both x64 and x86? Smart; I was going to write an interpreter for it :sweat:

pixelherodev:
Wait, the emitter uses the Microsoft ABI and not the SystemV one? That's interesting...

tellowkrinkle:
> Immediate comment.
> 
> This shouldn't be needed:
> 
> ```
> #if defined(__x86_64__) || defined(_M_AMD64)
> #define __M_X86_64
> #endif
> ```
> 
> Unless it's a change _I_ made, BuildParameters.cmake should already do that.

It seems you define one with one underscore, but the one with *two* underscores...

```
$ grep -r __M_X86_64 .
common/include/Pcsx2Defs.h:#define __M_X86_64
common/include/Pcsx2Defs.h:# ifdef __M_X86_64
common/include/x86emitter/internal.h:#ifdef __M_X86_64
common/include/x86emitter/internal.h:#ifdef __M_X86_64
common/include/x86emitter/x86types.h:#ifdef __M_X86_64
common/include/x86emitter/x86types.h:#ifdef __M_X86_64
common/include/x86emitter/x86types.h:#ifdef __M_X86_64
common/include/x86emitter/instructions.h:#ifdef __M_X86_64
common/include/x86emitter/implement/jmpcall.h:#ifdef __M_X86_64
common/include/x86emitter/implement/jmpcall.h:#ifdef __M_X86_64
common/include/x86emitter/implement/jmpcall.h:#ifdef __M_X86_64
common/include/x86emitter/implement/jmpcall.h:#ifdef __M_X86_64
common/include/x86emitter/implement/jmpcall.h:#ifdef __M_X86_64
common/include/x86emitter/implement/jmpcall.h:#ifdef __M_X86_64
common/include/x86emitter/implement/jmpcall.h:#ifdef __M_X86_64
common/src/x86emitter/simd.cpp:#ifdef __M_X86_64
common/src/x86emitter/cpudetect.cpp:#ifdef __M_X86_64
common/src/x86emitter/cpudetect.cpp:#ifdef __M_X86_64
common/src/x86emitter/cpudetect.cpp:#ifdef __M_X86_64
common/src/x86emitter/movs.cpp:#ifdef __M_X86_64
common/src/x86emitter/movs.cpp:#ifdef __M_X86_64
common/src/x86emitter/x86emitter.cpp:#ifdef __M_X86_64
common/src/x86emitter/x86emitter.cpp:#ifdef __M_X86_64
common/src/x86emitter/x86emitter.cpp:#ifdef __M_X86_64
common/src/x86emitter/x86emitter.cpp:#ifdef __M_X86_64
common/src/x86emitter/x86emitter.cpp:#ifdef __M_X86_64
common/src/x86emitter/x86emitter.cpp:#ifdef __M_X86_64
common/src/x86emitter/x86emitter.cpp:#ifdef __M_X86_64
common/src/x86emitter/x86emitter.cpp:#ifdef __M_X86_64
common/src/x86emitter/jmp.cpp:#ifdef __M_X86_64
pcsx2/System.cpp:#ifdef __M_X86_64
```

pixelherodev:
Maybe fix that in the CMake file then, instead?

arcum42:
That's probably legacy of when I was going through unifying what the code was using to check for 64 bits. I swear there was some oddball time when it wasn't being defined (and bear in mind I was trying to make sure both gcc and clang were fine with it). Though one underscore vs. two is exactly the sort of thing I could see tripping me up...

tellowkrinkle:
> Maybe fix that in the CMake file then, instead?

Or should we coalesce all the usages onto single-underscores?

tellowkrinkle:
Wait I just realized I went through this thinking that the code was working on windows x64 but not on Linux (hence the title specifying Unix)
But given your replies that may not be the case
If this hasn't previously worked for x86_64 on any platform, I would want to look at it a bit more closely, the current changes are just the result of me going "hey the program crashed, wait that assembly makes no sense!"

gregory38:
VIF is mostly SSE, except the function prologue/epilogue there is 0 difference between 32/64 bits. (you only gain some logical registers) (side note: 64 bits won't change/improve float support neither)

As far as I remember we use native ABI. So mswindow uses msbuild, and unix systemV. However, msbuild does bad thing with SSE register (i.e. use them as argument and they must be saved by callee).

gregory38:
On the move stuff. If I remember, only a single mov instruction support 64 bits immediate. Maybe it would make more sense to limit default code to 32 bits immediate and use a special 64 bits mov when we need it (which shall be very limited, i.e. loading recompiler data structure when entering recompiler).

gregory38:
Potentially if imm fit in 32 bits then use 32 bits immediate otherwise 64 bits.

tellowkrinkle:
At least for the vif functions I'm pretty sure they were marked __fastcall and always generated with arguments from rcx and rdx (though I'm on my phone right now so I can't really find it easily)
And for the codegen that makes sense, I'll look into it tomorrow

pixelherodev:
```
Potentially if imm fit in 32 bits then use 32 bits immediate otherwise 64 bits.
```

Heck, I'd argue that it's worth teaching the emitter to use smaller instructions when possible. If an IMM is passed as 32-bits but is known to be 16-bit, use a 16-bit load, and so on.

gregory38:
I don't deny the calling convention was wrong. So yeah it need to be fixed, fix is likely use native convention. However maybe the frame is still wrong

gregory38:
@pixelherodev micro-optimization can wait. It could be tricky. 64 bits immediate is really a special case of the ISA.

tellowkrinkle:
How does that look?

> Heck, I'd argue that it's worth teaching the emitter to use smaller instructions when possible. If an IMM is passed as 32-bits but is known to be 16-bit, use a 16-bit load, and so on.

16-bit and 8-bit load doesn't zero the rest of the register like 32-bit load does (yay for x86 backwards compatibility)

pixelherodev:
Yeah, I just found that out the hard way with my own project :(

gregory38:
Immediate support need to be squared. If argument is an int, it means it can only be 4 bytes. I think we should keep/enforce it this way. And create a MOV64(reg, immediate) pseudo instruction that allow to load a 64 bits immediate into a register would. Load a 64 bits immediate is really a special case of ISA. It would make core more readable to separate it from the standard move.

tellowkrinkle:
Something like that?
(BTW should I remove the commits that have had their changes undone by later commits?)

pixelherodev:
Good work on this one!

gregory38:
May I suggest to use MovImm64 (just to understand it is only an immediate move).

tellowkrinkle:
Like that?  Or should I rename the `xMOV64` pseudo-instruction as well?

tellowkrinkle:
Got controller input working and got past the title screen in P3FES, allowing me to run into another issue with the codegen used by the vif jit

When the jit attempts to generate the use of a fixed memory address, in x86-32 [it generates an instruction directly referencing the address](https://github.com/PCSX2/pcsx2/blob/58f5a5b915915293b90e7e8f34c33f9c0424d1e8/common/include/x86emitter/x86types.h#L991).  In x86-64, this has two issues:
- The memory address may not fit in 32 bits
- An instruction that used a fixed memory address in x86-32 is a rip-relative offset in x86-64

At least on my computer (macOS Catalina), all addresses (except the stack) are within 4GB of each other, but lie outside of the bottom 4GB of address space, meaning that both of these issues would be solved by modifying the codegen to subtract the next instruction's address on x86-64.

The only place I can think of that will know where the start of the next instruction will be is [`EmitSibMagic`](https://github.com/PCSX2/pcsx2/blob/58f5a5b915915293b90e7e8f34c33f9c0424d1e8/common/src/x86emitter/x86emitter.cpp#L262).  To put the logic here, `xIndirectVoid` would need to be updated to store `Displacement` as an `sptr` since it needs to stay 64 bits until it gets to `EmitSibMagic`

Can you think of any better places to put it?

pixelherodev:
"The memory address may not fit in 32 bits" This can be circumvented rather trivially: when mapping in memory for the recompilers, simply mark it with `MAP_32BITS` and it'll be placed in the lower 32-bits of the address space (at least on Linux, I'm not sure if other OSes support this).

pixelherodev:
```
returning virtual addresses <4G requires entitlement on Apple platforms
```
Drats. Never mind then.

pixelherodev:
A couple of notes from my own work (https://git.sr.ht/~pixelherodev/pcsx2), in case you (or anyone else) attempt to do more work on amd64 support:

* The first issue you'll run into is that register sizes are _wrong_. In amd64, the 32-bit registers and 64-bit registers are both declared as type xAddressReg, which extends xRegisterLong. On amd64, xRegisterLong is defined as xRegister64. You can't just change the 32-bit regs to xRegister32, for one simple reason: xAddressReg is the type used for indirect addressing. The 32-bit registers need to _remain_ as xAddressReg. It's the _64-bit_ ones that need to be changed to xRegister64 directly from xAddressReg (because they _can't_ be used for addressing), and xAddressReg should directly extend xRegister32 instead of xRegisterLong. This, annoyingly, requires tiny updates to a bunnnnch of other files as well, as fixing register sizing results in e.g. `xFastCall(func, ecx);` failing because xFastCall takes an xRegisterLong and on x64 ecx is, of course, *not*. Another option is to implement another xFastCall copy that takes in 32-bit registers and tail calls the other xFastCall with the 64-bit variant.
* The emitter currently is broken for pushes and pops in amd64, and likely others. It simply emits `50 + rd`, which is correct _for non-extended registers._ However, ScopedStackFrames push and pop r11-r15, and they currently generate incorrect instructions (pushes generate _pops_, which breaks the stack). This is trivial to fix; simply update xPUSH and xPOP to `EmitRex(0, reg)` first.
* Fixing those two bugs exposes a page fault on NULL, which likely suggests usage of a function before it is JITed, or that another instruction is generating incorrectly.

pixelherodev:
also,
```
            case 8:
                // Only mov's take 64-bit immediates
                xWrite32(imm);
                break;
```

Shouldn't that trigger an assertion failure instead? If an attempt to use a 64-bit immediate is made, we shouldn't be implicitly truncating it.

tellowkrinkle:
> The first issue you'll run into is that register sizes are _wrong_. In amd64, the 32-bit registers and 64-bit registers are both declared as type xAddressReg, which extends xRegisterLong. On amd64, xRegisterLong is defined as xRegister64. You can't just change the 32-bit regs to xRegister32, for one simple reason: xAddressReg is the type used for indirect addressing. The 32-bit registers need to _remain_ as xAddressReg. It's the _64-bit_ ones that need to be changed to xRegister64 directly from xAddressReg (because they _can't_ be used for addressing), and xAddressReg should directly extend xRegister32 instead of xRegisterLong. This, annoyingly, requires tiny updates to a bunnnnch of other files as well, as fixing register sizing results in e.g. `xFastCall(func, ecx);` failing because xFastCall takes an xRegisterLong and on x64 ecx is, of course, _not_. Another option is to implement another xFastCall copy that takes in 32-bit registers and tail calls the other xFastCall with the 64-bit variant.

Wait wouldn't the 64-bit ones be the ones used for addressing on x86-64?  (though I guess it doesn't matter too much, since there's not any instructions to address with 32-bit registers)
Personally, I think the r** registers should stay xRegisterLong as it's useful to have something that's the native register size on both platforms, and it's not like the r** registers are otherwise needed on x86-32 anyways.
IMO the ideal setup would be e** → xRegister32, r** → xAddressReg, though I do realize that would require a lot of messing with existing code...

> The emitter currently is broken for pushes and pops in amd64, and likely others. It simply emits `50 + rd`, which is correct _for non-extended registers._ However, ScopedStackFrames push and pop r11-r15, and they currently generate incorrect instructions (pushes generate _pops_, which breaks the stack). This is trivial to fix; simply update xPUSH and xPOP to `EmitRex(0, reg)` first.

We've got a bit more to fix than just that.  I ended up giving up on attempting to fix things as they come up, and am [attempting to write some unit tests for the codegen](https://github.com/tellowkrinkle/pcsx2/commit/639480860d80dab55f8ac487bda9ce5f2c6bae38).  I've only added mov, push, and pop tests so far but [it seems we have a lot of issues with handling of extended registers, not just in push/pop](https://gist.github.com/tellowkrinkle/a08596a025bc3f6e95c6cca2d5814285)

> Shouldn't that trigger an assertion failure instead? If an attempt to use a 64-bit immediate is made, we shouldn't be implicitly truncating it.

That does seem like a good idea

pixelherodev:
> Wait wouldn't the 64-bit ones be the ones used for addressing on x86-64? (though I guess it doesn't matter too much, since there's not any instructions to address with 32-bit registers)

Wrong, and wrong. Even in amd64, 64-bit registers _cannot_ be used for addressing; only the 32-bit ones can. The only way to extend beyond 32-bit addressing is via the REX prefix, but the default is still 32-bit registers, and the recompilers as they are today will still use the 32-bit ones (and likely will forever, since there's no gain to switching them). _There are no instructions that don't use 32-bit registers for addressing_.

Note that addressing here refers specifically to usage for indirect references via MODr/m + SIB.

> Personally, I think the r** registers should stay xRegisterLong as it's useful to have something that's the native register size on both platforms, and it's not like the r** registers are otherwise needed on x86-32 anyways.

They are. xRegisterLong is defined as xRegister64 in a macro on x64. However, it still makes sense to define the registers using the xRegister64 name as it's guaranteed to be correct regardless of platform.

> it seems we have a lot of issues with handling of extended registers, not just in push/pop

I'm not surprised. Thanks for the direction; I'll add fixes to xMOV to my branch on top of the PUSH/POP fixes.

tellowkrinkle:
> Wrong, and wrong. Even in amd64, 64-bit registers _cannot_ be used for addressing; only the 32-bit ones can. The only way to extend beyond 32-bit addressing is via the REX prefix, but the default is still 32-bit registers, and the recompilers as they are today will still use the 32-bit ones (and likely will forever, since there's no gain to switching them). _There are no instructions that don't use 32-bit registers for addressing_.
> 
> Note that addressing here refers specifically to usage for indirect references via MODr/m + SIB.

If you mean that they don't need any REX prefixes then yes, but the machine code for `mov eax, [ebx]` in x86-32 is the equivalent of `mov eax, [rbx]` in x86-64 (it uses the entirety of rbx, not just the bottom 32 bits), so I feel like it makes more sense to consider that a use of the 64-bit register (in the same way that `push rax` doesn't use a REX prefix but uses `rax`, not `eax`)

pixelherodev:
> If you mean that they don't need any REX prefixes then yes, but the machine code for mov eax, [ebx] in x86-32 is the equivalent of mov eax, [rbx] in x86-64 (it uses the entirety of rbx, not just the bottom 32 bits), so I feel like it makes more sense to consider that a use of the 64-bit register (in the same way that push rax doesn't use a REX prefix but uses rax, not eax)

It's not about that though.

Fundamentally, indirect addressing (via SIB) still uses 32-bit addresses. xAddressReg is specifically (as I understand it) the register used for SIB addressing. If you set xAddressReg to xRegister64 a lot breaks because it can't be used in such a manner. The only reason it currently works is that PCSX2 thinks 64-bit == 32-bit (it can't tell the difference between eax and rax).

pixelherodev:
The part of the manual I've been referencing, if it wasn't clear:
```
2.2.1.3
Displacement
Addressing in 64-bit mode uses existing 32-bit ModR/M and SIB encodings. The ModR/M and SIB displacement
sizes do not change. They remain 8 bits or 32 bits and are sign-extended to 64 bits.
```

tellowkrinkle:
Okay just tested on a linux computer and it looks like linux does not put static variables anywhere near pages from mmap
So I added a system to copy the required static variables to a codegen accessible area (`MAP_32BIT` allocated page on linux, generic mmap page that we hope is near other mmap pages on macOS)

pixelherodev:
You can use a linker script to change where they're placed on Linux.
That's what I've done in the past.

pixelherodev:
Here's an idea: if all allocations need to be near each other, why not just pre-allocate all the memory PCSX2 needs as a continuous block and then hand different portions to various components malloc()-style? Instead of mmap()ing in blocks separately for each component, just add the total amount needed, mmap() it all as a single contiguous block, write a fixed buffer memory allocator which simply increments an index into the block and hands it off, and have various components call `pcsx2_alloc(amount)`?

pixelherodev:
Also, found a small nit to pick:
```
static const int ModRm_EIZ = 4;       // same index value as ESP (used in Sib Index field)
```

This should be `SIB_EIZ`, not `ModRm_EIZ`, as it's not actually a ModRm value.

tellowkrinkle:
> This should be `SIB_EIZ`, not `ModRm_EIZ`, as it's not actually a ModRm value.

Should I do anything about the fact that `ModRm_UseDisp32` is used in both [ModRm](https://github.com/PCSX2/pcsx2/blob/de1717c1a1716fd9e4762dbaa5f6567670481353/common/src/x86emitter/x86emitter.cpp#L255) and [SIB](https://github.com/PCSX2/pcsx2/blob/de1717c1a1716fd9e4762dbaa5f6567670481353/common/src/x86emitter/x86emitter.cpp#L330)?

arcum42:
What's the current status on this?

tellowkrinkle:
On the `ModRM_EIZ`, I was kind of hoping for a response to whether I should change `ModRm_UseDisp32` as well

On changing the PCSX2 memory allocation system, I do think it's a good idea, but personally think it should be done as a separate change.  We would still need to be able to move stuff from the data segment to the allocation, so it wouldn't really replace anything here.  If anyone disagrees though I'm happy to implement it now

One other thing I'm not sure about is my placement of `MakeCodegenAccessible` in `HostSys`.  I couldn't think of any better place to put it, but if someone who knows the PCSX2 namespaces better has a better place for it that would be great.  Or maybe make a separate namespace for it?

arcum42:
Trouble is that this is one of these murky spots where there aren't that many people that still have much in the way of knowledge of the code, I think (Though, again, @gregory38 gregory would know more.).

HostSys is mostly platform-specific memory code. There was a note in there saying:
"// (this namespace name sucks, and is a throw-back to an older attempt to make things cross
// platform prior to wxWidgets .. it should prolly be removed -- air)"
though, obviously air is long gone. 

Memory.cpp and VirtualMemory.cpp are the other places I can think of involved. Most of HostSys is in two Linux and Windows only files.

Don't really see an issue with keeping it where it is for now, and figuring out if we need to move it later, though.

A separate pr for changing the memory management seems like a good idea. There's actually a really old broken pr for doing memory optimization changes. I tried to rebase it once and utterly failed. Might be worth looking to see if anything from there was useful.
https://github.com/PCSX2/pcsx2/pull/1134

Not really sure on ModRM myself, I'm afraid.

gregory38:
I have 0 (if not negative) free times.

gregory38:
IMHO, we should allocate a big data structure (something like a 2GB area) which will contains memory for recompiler and stuff accessible by the recompiler. This way, you know you can use RIP addressing easily

gregory38:
The bonus point of 2B area is that it might use huge translation table. So it is better for v2p cache

gregory38:
Issue of mapping 2 areas is that it might fail.

pixelherodev:
> On the ModRM_EIZ, I was kind of hoping for a response to whether I should change ModRm_UseDisp32 as well

Sorry, completely forgot to answer this one. I'd say yes, a second value SIB_UseDisp32 makes the most sense. The values are completely identical, but the _meaning_ is different, and using a ModR/M value for SIB encoding is a bit weird. Making a second value makes it more clear what it's being used for, and has literally zero cost.

beaumanvienna:
Hi, I noticed you guys are working on the x64 code, which is awesome! I started reviewing the emitter code myself. I noticed in xAlignPtr the 32-bit mask is chopping of the leading part of x86Ptr (now that we have 8-byte pointers) and suggest instead:
```
void xAlignPtr(uint bytes)
{
    // forward align
    #ifdef __M_X86_64 
      unsigned long mask64bits;
      mask64bits = bytes;
      mask64bits = ~(mask64bits - 1);
      x86Ptr = (u8 *)(((uptr)x86Ptr + bytes - 1) & mask64bits);
    #else
      x86Ptr = (u8 *)(((uptr)x86Ptr + bytes - 1) & ~(bytes - 1));
    #endif
} 
```
I would be interested in a general conversation about the x64 port. Would you guys like to open a dedicated thread for that? 
I also started to write some documentation here: https://github.com/beaumanvienna/pcsx2-testing/tree/master/doc
LMK if you want me to add xAlignPtr and/or the doc file to PR #3434. 
FYI, this is my motivation: https://github.com/beaumanvienna/marley

My approach for debugging is to compile an x64 and i386 version and to compare the outputs. I put in a little debug code so that they use different config folders. I work with gdbgui and like to put into the code a lot of printf's to pinpoint the location of a crash.
Regards,
JC

arcum42:
@beaumanvienna 
As far as the documentation goes, first, you can probably remove the references to SuperVU/sVU, as it was recently deleted.
GSdx also has its own recompiler, using xybak rather than x86emitter.
As far as central files go, the /pcsx2/pcsx2/x86 directory is where a lot of them are, and /pcsx2/pcsx2/x86/iCore.cpp and /pcsx2/pcsx2/x86/ix86/icore-32.cpp are probably worth a mention, as is /pcsx2/pcsx2/x86/BaseblockEx.cpp.

Given that #3434 is specifically for defines for opcodes, it's probably best to keep each pr focussed on what that particular pr is for. Adding documentation is a good idea, just something that I think would be added in a different pr.

The fix probably also more belongs in this pr or a separate one. 3434 requires rather less testing without it, so I can get it merged relatively soon, once I've had a chance to sanity-check things.

If anyone does want to open someplace else for general discussion of porting to x64, like an issue or a forum thread or such and link it here, they can. I just figure the conversation is already here, and moving conversations sometimes causes them to die out...

I will say I've never written a JIT and am *not* the most familiar with assembly, though both are topics I'm interested in and learning more about. I tend to know a little bit of everything (but not necessarily a lot), and am fairly familiar with pcsx2's source code. And, of course, I can commit things...

pixelherodev:
After having worked on this for a while, I'm not willing to continue porting x86emitter to x64. Others can keep going; I'll be working on a new x64 backend soon-ish (as time permits) and porting over the EmotionEngine so that it can be selected at compile time, instead. The x86emitter is... I'm honestly not a fan, even if it can be made to work properly.

tellowkrinkle:
Okay I changed everything to allocate from a single memory pool

beaumanvienna:
@arcum42 : I opened a new PR per your suggestion. We can discuss my x64 changes there. I added the doc-file. 
@pixelherodev : Noted, I will refrain from asking too many questions LOL. If you still find the time to review my changes, it would be appreciated. You seem to be very knowledgeable. 
@tellowkrinkle : I just created PR #3451 , which looks very similar to this PR. Your comments and feedback would also be highly appreciated. 

Does any of you guys know what we need to change to make the tables "recLUT" and "recLUTReserve_RAM" work properly under x64?

pixelherodev:
@beaumanvienna Ask as many questions as you want on the x86 / x64 ISA, I'm glad to answer those (and I don't need to look at the PCSX2 source for that). Seriously, I'm more than happy to help, ask away. With regards to the PCSX2 source... I'm deliberately not learning more about the current recompilers, so I'll only be able to answer what I've already gleaned in the past.

tellowkrinkle:
Changes here are now a part of #3512 and #3451, closing

PR #3316, opened 2020-04-14
Mark CMP0054 NEW; we don't rely on legacy behavior

### LINKED ISSUE #3310: Build outputing warning "Policy CMP0054 is not set" since CMake 3.17
***PCSX2 version:***
1.5.0-20200408193126 - compiled on Apr 14 2020

***PCSX2 options:***
N/A

***Plugins used:***
N/A

Since CMake 3.17, a deprecation warning message is displayed when "Policy CMP0054 is not set":

```
CMake Warning (dev) at /usr/share/cmake-3.17/Modules/FindPNG.cmake:97 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "Linux" will no longer be dereferenced when the
  policy is set to NEW.  Since the policy is not set the OLD behavior will be
  used.
  cmake/SearchForStuff.cmake:21 (find_package)
  CMakeLists.txt:27 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```
See full build log: 
[pcsx2-git-1.5.0.r3401.49e316e48-1-x86_64-build.log](https://github.com/PCSX2/pcsx2/files/4476715/pcsx2-git-1.5.0.r3401.49e316e48-1-x86_64-build.log)

Build PCSX2 with cmake>=3.17 (see [CMP0054 in CMake docs](https://cmake.org/cmake/help/v3.17/policy/CMP0054.html))

For instance, build AUR package [pcsx2-git](aur.archlinux.org/packages/pcsx2-git/), as Arch Linux has cmake>3.17

N/A

***PC specifications:***
Arch Linux (rolling, 64-bit), cmake 3.17.1

pixelherodev:
This closes #3310.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ project(Pcsx2)
 # Bumping up to 3.0 seems reasonable at this point, and will let us modernize
 # things a bit.
 cmake_minimum_required(VERSION 3.0.2)
+cmake_policy(SET CMP0054 NEW)
 
 # Variable to check that people use the good file
 set(TOP_CMAKE_WAS_SOURCED TRUE)


pixelherodev:
The good news is we don't rely on the old behavior anyways as far as I can tell; everything appears correct with the policy marked as NEW.

gregory38:
Might need to be tested on older cmake. To be sure they doesn't complain the policy doesn't exist.

pixelherodev:
Good call, I'll do that sometime in the next few days (a bit busy at the moment finishing work in my lengthy backlog).

arcum42:
Interestingly enough, it looks like you can set the minimum cmake version to a range, like this:
cmake_minimum_required(VERSION 3.0.2...3.17)

Doing this tells it that you also support the newer versions, so it sets the policy to new instead of old.

I'll still probably want to test that with older versions, though.

arcum42:
Closing, since I merged the build_changes branch.

PR #3324, opened 2020-04-24
Implemented memory mode check in cop0.cpp

prafullpcsx2:
This makes Next Generation Tennis 2003 (Roland Garros French Open 2003) work.


refractionpcsx2:
This also fixes the PAL release of Spongebob Squarepants Battle for Bikini Bottom

PR #3337, opened 2020-04-30
Locale:ja_JP:Update Hotkey notation

knight-ryu12:
to follow Japanese Windows Window to look good


lightningterror:
Noticed a bit too late, this breaks keyboard shortcuts tho.

Issue #3214, opened 2020-01-16
[Feature Request] Ability to completly unbind hotkeys


megapro17:
I know, that it is possible to change hotkeys in file PCSX2_keys.ini, but if I remove hotkey it uses default value. It is possible to add something like "NONE" or empty value to disable hotkey? I didn't find any messages about this, only questions about rebinding. Tried on latest git snapshot.

lightningterror:
Done on Qt, won't be done on wx.

PR #3345, opened 2020-05-01
Ability to completly unbind hotkeys

megapro17:
I spent three hours trying to do better it way. WXK_NONE is not working, it just uses a default hotkey. I had to use some unused special key, that actually doesn't exist on keyboards

Fixes https://github.com/PCSX2/pcsx2/issues/3214


arcum42:
Anything changing the sourcecode for wxwidgets is going to be windows only, since Linux uses the system version, and also makes it harder to update between versions of wxwidgets...

megapro17:
Maybe there is a better way?

avih:
I also think changing WX source is not great.

The special keys names are listed at `PCSX2_keys.ini.default`, and you can always "un-assigne" by assigning to some combo which you're highly unlikely to invoke, like ctrl+shift+alt+z etc.

Bottom line, unless you find a clean and preferably simple way to do that from PCSX2 itself (i.e. without changing the WX sources), I don't think it poses a real enough issue to address.

arcum42:
I'd imagine changing it properly would be done  over in gui/GlobalCommands.cpp. If we ever got rid of the controller plugins and merged them into pcsx2, that's an area I'd probably want to tie directly into them.

I've got limited time to work on pcsx2, and many other sections my attention is going to, though.

arcum42:
I'll close this for the moment. If anyone comes up with a way to change the keybindings that doesn't hack wxwidgets, they can always open a new pr.

Issue #3008, opened 2019-07-04
High Input Lag on 3D games


Dogway:
***PCSX2 version:***
pcsx2-v1.5.0-dev-3138-gfbafd4420-windows-x86

***PCSX2 options:***

Software or Hardware mode indistinctly. Using Safe Preset

***Plugins used:***
Default:
GSdx32-AVX2
Spu2-X
LilyPad 0.12.1
FWNull Driver 0.7.0

***Description of the issue:***
There's a high game input lag in 3D based games, rarely below 6 frames of lag. Here are some tests done with frame advance using the minimum recorded frames for fastest event paths.

MGS3 Subsistance (6 frames)
GTA Vice City (6 frames)
GTA San Andreas (10 frames, 166ms!)
Sengoku Basara X (6 frames)
Tokyo Xtreme Racer 3 (7 frames)
Taiko Drum Master (6 frames)
Silent Scope (6 frames)
Time Crisis 3 (5 frames)
Capcom Fighting Evolution (4 frames)

Generally it's 6 frames of game input lag which is one of the highest recorded on current emulation software compared to Dolphin (5), Citra (4), [Retroarch ](https://forums.libretro.com/t/input-lag-investigation-on-3d-systems/21684)(4-5) and possibly Cemu and RPCS3 (they feel snappy)

***How to reproduce the issue:***

-Enable "Recording Tools"
-Load a game
-Press the frame advance hotkey
-Press a fast event triggering button on your pad
-Press again the frame advance hotkey as many times (counting them) until the event starts the animation 

***PC specifications:***
Operating System: Windows 7 x64 SP1
CPU: Intel i5-4670K @ 4.1Ghz
GPU: Nvidia GeForce GTX 1070

jacob1218:
I’ve noticed that too, what controller are you using?

Dogway:
I use a 360 wireless controller but this is unrelated because I'm only measuring the emulation path input lag through the frame advance procedure. I guess judging by the picture you might already know but [here ](https://www.youtube.com/watch?v=Cv-OOn7iYio)is a good review of controller input lag.

KrossX:
Is there data on PS2 + CRT input lag? It would be nice to have some, specially for GTA-SA. Might be tricky though as you could need a light in the controller for when a button is pressed and when pressure is greater than deadzone.

tadanokojin:
Honestly I have a lot of problems with these numbers. Not the least of which is I can't independently verify any of them.

[...]

PR #3346, opened 2020-05-01
pcsx2: Use old saveslot UI.

lightningterror:
There's a callback to the ui to update the savestates gui whenever the
recompiler exits. the problem with this is that the recompiler exits a
ton, especially when a game boots or on a game with self modifying code
like Jak. This causes core to spam the message system to update the ui,
which in turn causes the savestates gui to attempt to perform a bunch of
OS i/o.

### DIFF
--- a/pcsx2/gui/Saveslots.h
+++ b/pcsx2/gui/Saveslots.h
@@ -21,7 +21,13 @@
 #include "Elfheader.h"
 
 // Uncomment to turn on the new saveslot UI.
-#define USE_NEW_SAVESLOTS_UI
+// Issue with new saveslot UI:
+// There's a callback to the ui to update the savestates gui whenever the recompiler exits.
+// The problem with this is that the recompiler exits a ton,
+// especially when a game boots or on a game with self modifying code like Jak.
+// This causes core to spam the messages system to update the ui,
+// which in turn causes the savestates gui to attempt to perform a bunch of OS i/o.
+//#define USE_NEW_SAVESLOTS_UI
 
 // Uncomment to turn on the extra UI updates *without* the UI.
 //#define USE_SAVESLOT_UI_UPDATES


arcum42:
Except that it quite clearly checks saveslot_cache[i].invalid_cache first, and only does UpdateCache, which does the disk i/o, if that is true. And that is set to true when the save slots are first initialized, when the crc changes, or the disk serial number changes. Not exactly things that happen often.

tadanokojin:
arcum see https://github.com/PCSX2/pcsx2/issues/3008#issuecomment-611635587

arcum42:
Comment out the call to Sstates_updateLoadBackupMenuItem in _SaveLoadStuff, then. It's called enough other places that it'll usually be up to date.

tadanokojin:
Long as it doesn't create any further issues with the ui I don't object to that.

arcum42:
Any issues it causes would have already been there in the old saveslot code. :P
It's technically possible it might not update in a timely fashion if the backup file is taking a while to be written, IIRC. It'd update next time you did anything to a savestate, in that case.

arcum42:
Created a saveslots branch, made the change there and refactored. The backup slot behavior is actually pretty weird, but in the same ways it was before this change and before my new saveslot code.

(By rights, we probably should have a backup saveslot *menu* instead of the current hack.)

tadanokojin:
Alright let's get testing on that then and if all goes well we won't have to resort to the old ui.

Thanks for the work arcum.

arcum42:
np. I'll admit I could have just gotten rid of the one line, but... that honestly should have already been a ranged for loop. 

Also, if the behavior of that backup menu item seems weird, keep in mind it's showing you the current backup or lack thereof for the *current save slot*...

lightningterror:
Closing as arcum has an alternative branch.

Issue #3313, opened 2020-04-14
CMake warning on unmatching HarfBuzz package/calling name


rffontenelle:
***PCSX2 version:***
1.5.0-20200408193126 - compiled on Apr 14 2020

***PCSX2 options:***
N/A

***Plugins used:***
N/A

***Description of the issue:***

The following warning message is printed:

```
CMake Warning (dev) at /usr/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:272 (message):
  The package name passed to `find_package_handle_standard_args` (PkgConfig)
  does not match the name of the calling package (HarfBuzz).  This can lead
  to problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  /usr/share/cmake-3.17/Modules/FindPkgConfig.cmake:45 (find_package_handle_standard_args)
  cmake/FindHarfBuzz.cmake:33 (INCLUDE)
  cmake/SearchForStuff.cmake:194 (find_package)
  CMakeLists.txt:27 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Checking for module 'harfbuzz>=0.9.0'
--   Found harfbuzz, version 2.6.4
-- Found HarfBuzz: /usr/include/harfbuzz
```

See full build log: 
[pcsx2-git-1.5.0.r3401.49e316e48-1-x86_64-build.log](https://github.com/PCSX2/pcsx2/files/4476715/pcsx2-git-1.5.0.r3401.49e316e48-1-x86_64-build.log)

***How to reproduce the issue:***
Build PCSX2

In my case, I built the AUR package [pcsx2-git](aur.archlinux.org/packages/pcsx2-git/)

***Last known version to work:***
N/A

***PC specifications:***
Arch Linux (rolling, 64-bit), cmake 3.17.1, harfbuzz 2.6.4

PR #3352, opened 2020-05-02
Fix various cmake issues

arcum42:
This does the following:
The Harzbuzz module is updated, resolving issue #3313 
The cmake minimum version is now a range, causing it to use the newer flags by default on newer versions up to 3.17, resolving Issue #3310
There's a new option to compile with flag -ftime-trace with clang, which can be useful for determining things that slow compile time.
If you are on gcc and cmake 3.16+, actually precompile PrecompiledHeader.h.

### DIFF
--- a/pcsx2/CMakeLists.txt
+++ b/pcsx2/CMakeLists.txt
@@ -739,7 +739,7 @@ endif()
 add_pcsx2_executable(${Output} "${pcsx2FinalSources}" "${pcsx2FinalLibs}" "${pcsx2FinalFlags}")
 
 if(NOT USE_CLANG)
-	if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16")
+	if(COMMAND target_precompile_headers)
 		message("Using precompiled headers.")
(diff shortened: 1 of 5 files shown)


pcsx2/CMakeLists.txt
```diff
@@ -738,6 +738,13 @@ endif()
 
 add_pcsx2_executable(${Output} "${pcsx2FinalSources}" "${pcsx2FinalLibs}" "${pcsx2FinalFlags}")
 
+if(NOT USE_CLANG)
+	if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16")
```
scribam, line 742:
```suggestion
	if(COMMAND target_precompile_headers)
```

> if(COMMAND command-name)
    True if the given name is a command, macro or function that can be invoked.

Source: https://cmake.org/cmake/help/v3.0/command/if.html

arcum42:
Effectively the same, since that's the version in which the command was introduced, but it makes the code clearer, so I'll go with it. I'll keep that one in mind...

Oh, and the check to make sure we don't use that command on clang is because when I tested, it broke compilation in clang, incidentally.

PR #3356, opened 2020-05-04
recording: Overhaul of Recording Tool's VirtualPad

xTVaser:
A note on the files modified:
- Edits to pcsx2's core source are mostly **reductions** in the recording footprint, namely in the Sio.cpp's interrupt.
- I was able to delete a few large chunks of the old code in the process, there are other unrelated (and smaller) refactors I'd like to do, but I kept them out of this PR as they weren't related to the VirtualPad.

Demo Video - https://www.youtube.com/watch?v=eAWn_QYsU08

The original Virtual Pad UI was not very nice visually, so the most obvious difference is this PR creates a much more visually appealing interface.  All images used to make it were made by me so I didn't have to concern myself with image licenses.

However, the original VirtualPad was also lacking a lot of features and had usability problems.  If the VirtualPad was opened _all_ inputs would be ignored and as a result it also did not show any feedback when replaying a movie.  A great example of a good TAS input is Dolphin's and I largely copied the behaviour of Dolphin's:
- Will update every frame to show active controller state, unless the real controller is explicitly ignored.
- If the real controller is not explicitly ignored you can use the VirtualPad and your controller in tandem, for example:
  - Lock the analog stick to a specific direction, and press face-buttons yourself.  As soon as you touch the analog stick on the real controller, the VirtualPad will cease to override it.
- When replaying a movie, the VirtualPad is largely disabled.

During development I found that if the GS plugin uses CPU rendering, then doing even a small amount of GUI events has a noticeable impact on the emulation performance.  This can be seen outside of this new window, if you scroll through the debugger window, the branch line drawing causes emulation to slow down.  For this reason I added a bunch of micro-optimizations to reduce the amount of widget updates and re-draw events and got it to the point where the performance impact while using CPU rendering is negligible, which means it works even better when GPU rendering is used.


xTVaser:
1080p100%Scaling:
image
4k150%Scaling:
image

RedDevilus:
1080P(UltraWide) 150%Scaling:
image

pcsx2/Recording/VirtualPad/VirtualPad.cpp
```diff
@@ -0,0 +1,438 @@
...
+	for (VirtualPadElement* virtualPadElement : virtualPadElements)
+		virtualPadElement->UpdateGuiElement(renderQueue, clearScreenRequired);
+
+	// Update Graphic Elements off render stack
+	// Before we start rendering (if we have to) clear and re-draw the background
+	if (!manualRedrawMode || clearScreenRequired || !renderQueue.empty())
+	{
+		bdc.SetBrush(*wxRED);
+		bdc.DrawRectangle(wxPoint(0, 0), bdc.GetSize());
+		bdc.SetBrush(wxNullBrush);
```
sonicfind, line 204: I do wonder if this is 100% necessary. Since the bitmap stretches over the entire window to begin with, I feel like this would just get covered up and may even cause a weird flicker of red in certain situations.

RedDevilus: It was used for testing purposes to see drawing artifacts from what i gathered.

xTVaser: Ah i forgot to change this to white, which would alleviate the flicker concern.  It's been a long time since I originally wrote this code, I believe properly clearing the screen solved some issues, in general I think it's just a best practice.

sonicfind: Fair

sonicfind:
768p100%dpi
image
Should be gucci once the small bar on the right side of this is fixed. x = 458 seems to do the trick (unless upscaling screws it). I see literally nothing else to worry about after that.
_Please disregard me placing this message in the wrong PR initially_

PR #3357, opened 2020-05-04
macOS fixes

tellowkrinkle:
Gets everything working on macOS except GSdx

Left GSdx out as it's going to be fairly large, requiring some heavy gl feature workaround necromancy.  Also I haven't finished reviving the gl feature workarounds needed for the HW renderer.


tellowkrinkle:
Actually should I split this apart more?  Still seems kind of big...

lightningterror:
Maybe not, it's not a big pr, would just keep separate commits as it is, for gsdx tho I think a separate pr would be better.

tellowkrinkle:
Oh I guess I should mention
Two issues I haven't been able to work out:
- When the app first launches the menus don't work.  Sending pcsx2 to the the background (clicking any other application) and then going back to pcsx2 fixes it
- Trying to open gdk configuration windows (currently just spu2-x but GSdx has one as well) more than once in a session crashes.  Maybe something to do with the fact that we aren't touching gdk in any other part of the app (since those two configuration windows are the only things that use gdk on macOS)?  (And calling `gdk_init` multiple times isn't it, I tried it with a static global to track whether we had initted before and it didn't help.  Then I checked the gdk source and `init` has a check in the beginning for if it was already initted)

bositman:
That's great work we get tons of emails from people asking for a Mac OS port :)

Mergenmen:
Yes, to relive all the games I bought and completed as a child is an absolute dream. Having a Mac is hard sometimes when applications are Windows only. I tried to fix my PlayStation 2, I got in 2002, but was unsuccessful and had a moment before putting it in the bin. I realised the memories are held within the games and not a sentimental console. 
Much love for the makers of this divine throw back.

arcum42:
This'll need a rebase at this point. Any luck with the issue with the menus?

With the controller database, I assume you're just adding the Mac ones in from the original source?

oscarbg:
that's great! even with early state would be good sharing a precompiled binary (Mac app) to play with it..
thanks..

tellowkrinkle:
> With the controller database, I assume you're just adding the Mac ones in from the original source?

Yes

> Any luck with the issue with the menus?

Well I found out that the issue doesn't exist on macOS 10.12 (3 year old OS, my main computer is 10.15).  On the other hand the GTK configs crash on 10.12 and I haven't looked into why yet.

> even with early state would be good sharing a precompiled binary (Mac app) to play with it..

I still need to untangle the binary from my macports/homebrew install or the binary won't run on anyone else's computer (unless they have all the same packages in the same places)
As a side note, if you ever need to compile 32-bit apps on macOS, MacPorts is the way to go, all its packages can be made into fat binaries by adding `+universal` (unlike Homebrew which is like "32-bit?  haha who needs that lol")

arcum42:
Alright, I'll just mainly want to make sure that either when this gets merged, the database is up to the current version or newer, or make sure we update it again afterwards. Though we'll have to make sure the mac controllers don't get deleted accidentally.

I also really need to get around to compiling this branch on linux and making sure nothing broke. 

And obviously, while this isn't going to be the last mac pr, since GSdx was separate, I don't want to commit this until you think it's ready. It *would* be nice to have pcsx2 compile for Mac OS X and work, though.

tellowkrinkle:
Okay looks like the crash was because I accidentally installed gtk-x11 instead of gtk-quartz and wasn't running an x server.  Switching to gtk-quartz fixed it

Don't merge quite yet, I'd like to add proper bundle packaging scripts

And for now, a 32-bit build I packaged using Dolphin's bundle packaging script.  OpenGL HW is kind of glitchy but SW should work [PCSX2.tar.gz](https://github.com/PCSX2/pcsx2/files/4627611/PCSX2.tar.gz)

oscarbg:
@tellowkrinkle many thanks for the build!.. sadly being 32bit only is not for Catalina, right? do you expect to be able to share a "equivalent" 64bit build later on.. say in a few weeks or so? thanks..

tellowkrinkle:
Okay packaging should be working
And it looks like the issue where menus weren't selectable may have been related to the fact that the app wasn't packaged in an app bundle?  Either that or it just randomly decided to fix itself.

BTW are there any higher resolution (>64x64) icon assets for PCSX2?  MacOS likes high-resolution icons

> do you expect to be able to share a "equivalent" 64bit build later on..

A 64-bit Mac build will be available when there's a 64-bit linux and windows build (not for a while)
...unless you want to run interpreter-only (10-20% speed)

oscarbg:
@tellowkrinkle thanks for answer.. happy to test whatever 64bit build, even if it's in an "interpreter-only" mode one..

Mergenmen:
Hiya, could you send me a step by step on how to get the 32bit please?

oscarbg:
@Mergenmen it’s in this thread!

Mergenmen:
I’m not understanding any of it”!”

tellowkrinkle:
To keep this thread for the PR, I've made a forum post for the build, please put your comments and questions there
https://forums.pcsx2.net/Thread-Native-Mac-Testing-Build-Mojave-and-lower

arcum42:
Went ahead and tested to make sure Linux was still working on this branch. Seems fine at the moment with either gcc or clang. I may want to fiddle with onepad, but I can always do that after the pr is merged.

Not really sure what the biggest copy of the logo is. Here's one that was around:
https://i.imgur.com/LRQxFBP.png

I'm inclined to go ahead and merge, since we can always do further prs later if need be, and I'm sure it'd be easier if these changes were already in the trunk. Any more changes you wanted to do first?

tellowkrinkle:
Hmm that image doesn't look so good, so why don't we stick with the current 64x64 (upscaled to 128x128 because the icns sizes are 16, 32, 128, 256, 512)
Aside from that I don't have anything I wanted to add, go ahead and merge

arcum42:
Done.

lightningterror:
Maybe we can rename the linux directories to unix, since they are shared for macos and linux now.

PR #3272, opened 2020-03-01
Add reading of 8bit EE timers and masking any offset. Fixes Robin Hood

refractionpcsx2:

PR #3361, opened 2020-05-06
Allow reading of 8bit timers (and other hardware pages)

refractionpcsx2:
Fixes Robin Hood


refractionpcsx2:
Same as Issue #3272 but without the rebase broken mess

PR #3367, opened 2020-05-08
GSdx: Pick opengl renderer as best default renderer for intel igpu.

lightningterror:
If gpu is dx12 capable.


lightningterror:
On second thought I will close it, GL_ARB_direct_state_access is not supported by intel 5th gen and I don't wanna set opengl as default if some extensions are not supported.

PR #3372, opened 2020-05-09
WinMisc: Flag versions older than 6.3 (Win8) as unsupported

MrCK1:


common/src/Utilities/Windows/WinMisc.cpp
```diff
@@ -108,7 +103,7 @@ wxString GetOSVersionString()
     else
         GetSystemInfo(&si);
 
-    if (VER_PLATFORM_WIN32_NT != osvi.dwPlatformId || osvi.dwMajorVersion <= 5)
+    if (VER_PLATFORM_WIN32_NT != osvi.dwPlatformId || osvi.dwMajorVersion <= 6 && osvi.dwMinorVersion <= 1)
```
turtleli, line 106: Maybe include Windows 8 as an unsupported operating system as well since that's EOL (I'm not referring to Windows 8.1, that's not EOL).

refractionpcsx2: Windows 8 should still work though, we're only flagging Win7 because we're removing something it needs. We don't want to cripple an OS just for lulz

lightningterror: I agree, w8 as minimum is fine. Might want to update github readme too.

turtleli: Changing this number doesn't actually prevent things from running. It'll just show up in the logs differently.

It'll still run until something that is necessary for it to work gets removed. Windows 7 probably still works at this point, it's just that XAudio2 can't be used on it anymore.

I think if the OS is EOL then we should opt to not provide support for it even if it still works and encourage the user to upgrade.

lightningterror: Alright then, CK go ahead with the change.

refractionpcsx2: I'm not sure if this if statement will work in its current form, what if the major/minor version is 5.2 (windows xp) ? it won't meet both conditions.  You might need to split it out so the minor version is checked afterwards

refractionpcsx2: if (VER_PLATFORM_WIN32_NT != osvi.dwPlatformId || osvi.dwMajorVersion <= 5 || (osvi.dwMajorVersion == 6 && osvi.dwMinorVersion <= 1))

That would work

turtleli:
You should also adjust this code since the Windows 8 branch can't be reached anymore:
https://github.com/PCSX2/pcsx2/blob/efe3bc1613aaee383ed896b786c43f764f9a9342/common/src/Utilities/Windows/WinMisc.cpp#L160-L162

CookiePLMonster:
Don't compare version infos like this - even if you got it right, it shouldn't be done and it's been a source of compatibility issues in the past (also the reason Windows currently lies about the OS version). Use `VerifyVersionInfo` instead.

insaneninja117:
Perhaps a message in the bootup log or in the readme to say that while PCSX2 might still continue to work on Win 7 the devs will not be offering any assistance, you can never say anything too clearly.

lightningterror:
@MrCK1 Status?

common/src/Utilities/Windows/WinMisc.cpp
```diff
@@ -108,7 +103,7 @@ wxString GetOSVersionString()
     else
         GetSystemInfo(&si);
 
-    if (VER_PLATFORM_WIN32_NT != osvi.dwPlatformId || osvi.dwMajorVersion <= 5)
+    if (VER_PLATFORM_WIN32_NT != osvi.dwPlatformId || osvi.dwMajorVersion <= 5 || (osvi.dwMajorVersion == 6 && osvi.dwMinorVersion <= 2))
```
CookiePLMonster, line 106: Just use `VerifyVersionInfo` and save yourself the trouble of comparing version numbers manually.

avih:
What problem does this solve? I think windows 7 is still supported with some enterprise environments.

kenshen112:
@avih we've decided to drop support for windows 7

avih:
> @avih we've decided to drop support for windows 7

That did not answer my question.

kenshen112:
Point is were purposely making every move to remove support officially for windows 7. But since you're probably still going to whine but y tho. 

We were having issues with xaudio and a few other packages causing complications with compiling. We got tired of that and here we are upgrading everything and removing this issues. This in particular is just the nail in the coffin

tellowkrinkle:
The problem this solves is "fixing things for people on ancient operating systems takes time from developers that could have been spent improving PCSX2 on newer operating systems"

avih:
Next time please include a meaningful commit message which includes the reason for the change. It's really not enough to say "change X to Y" without explaining why X is a problem.

NewMoonHD:
https://www.microsoft.com/en-us/windows/windows-7-end-of-life-support-information

tellowkrinkle:
Normally I'd agree, but I think the reason for a commit titled "flag version x of operating system as unsupported" is pretty obvious, it's because we no longer support version x of the operating system

avih:
> https://www.microsoft.com/en-us/windows/windows-7-end-of-life-support-information

On it's own, this is not really a good reason.

> "fixing things for people on ancient operating systems takes time from developers that could have been spent improving PCSX2 on newer operating systems"

Great, this is a good reason, preferably when it comes with an example or two. And it's best if in the future such things are included at the commit message, for posterity.

kenshen112:
I mean it seems like a good reason to me. We don't want to support old os. That's clearly stated in the title. Doesn't matter if it fixes things or doesn't.

RedDevilus:
Code should be easily readable, just like commit description/name. If this would happen to all commits, then the readability will go down till it becomes a mess. Everyone agreed for dropping it, i don't see what the problem is. Besides how many have W7 extended support? Many are on newer versions which helps speed up progress, refactor easier, clean-up , formatting, etc. The same happened to Windows XP.

avih:
Whatever the reason is, next time please include it at the commit message. That's my main point.

kenshen112:
Which was presently done with the original post :)

avih:
> If this would happen to all commits

This (stating a reason at the commit message) SHOULD happen at all commits.

The code should stay clean and explain only things which the CODE does.

Commit messages should explain the reason for the CHANGE.

tadanokojin:
Your point seems like a bit of an afterthought Avih considering you edited your original message to include it.
We've had conversations about why already. Are you suggesting we need to include our reasoning every time we do a breaking change to win7?

We don't support it anymore. So change reflect we don't support it anymore. Seems pretty straightforward to me.

avih:
> Your point seems like a bit of an afterthought Avih considering you edited your original message to include it.

It's not an afterthought, and I stand behind it. The first thing I did was to look at the commit message to understand why, and when I didn't see the reason I asked here.


> We don't support it anymore. So change reflect we don't support it anymore. Seems pretty straightforward to me.

This is another great reason which should have been at the commit message as well.

Don't make a fuss, I'm not arguing the action itself, just that commit messages should include the reasoning for changes, especially when it's a big change like this one.

tadanokojin:
That did not answer my question.

avih:
> Are you suggesting we need to include our reasoning every time we do a breaking change to win7?

No, and I'm not suggesting the opposite either.

I'm only requesting that next time commit messages should include the reasoning for the change.

CookiePLMonster:
Basing on the feedback from this issue, I vote towards dropping support for anything but Windows 10 2004.

PR #3376, opened 2020-05-10
Performance fix for Primal and Ghosthunter. 

RazielZnot:
Fix for terrible performance in US versions of Primal (all the time) and Ghosthunter (when encountering Teddy Bear and in the moment of Gator-Man's rush attack, maybe few other cases). Both games use shared code with pretty dumb wait loops in the microcode, those are absolutely unnecessary and significantly kill performance on the emulator.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -4685,7 +4685,10 @@ Compat = 5
 [patches = FCD89DC3]
 	author=Raziel I.C.H.I. Znot
 	comment=Performance fix
-	//microcode, remove wait loop
+	//PCSX2 attempts to run VU0 for thousands of cycles as soon as it is started,
+	//so the game will wait endlessly in a wait loop until COP2 gets around to
+	//setting vi5(to break the loop) and vf9(to finish calculation) which is many thousands of cycles later.
+	//VU0 side, remove wait loop
 	patch=1,EE,0040F110,word,8000033c
 	patch=1,EE,0040F128,word,8000033c
 	patch=1,EE,0040F168,word,8000033c
@@ -4695,7 +4698,7 @@ Compat = 5
 	patch=1,EE,0040F738,word,8000033c
 	patch=1,EE,0040F778,word,8000033c
 	patch=1,EE,0040F790,word,8000033c
-	//EE code, rearrange code so VU code receives all data at once
+	//EE side, rearrange and clean code so VU0 receives all data at once
 	patch=1,EE,00399B14,word,00000000
 	patch=1,EE,00399B58,word,48A44800
 	patch=1,EE,00399B5C,word,4A00D839
@@ -39734,7 +39737,10 @@ Compat = 5
 [patches = 29C7DA54]
 	author=Raziel I.C.H.I. Znot
 	comment=Performance fix
-	//microcode, remove wait loop
+	//PCSX2 attempts to run VU0 for thousands of cycles as soon as it is started,
+	//so the game will wait endlessly in a wait loop until COP2 gets around to
+	//setting vi5(to break the loop) and vf9(to finish calculation) which is many thousands of cycles later.
+	//VU0 side, remove wait loop
 	patch=1,EE,00461C68,word,8000033c
 	patch=1,EE,00461C80,word,8000033c
 	patch=1,EE,00461CC0,word,8000033c
@@ -39744,7 +39750,7 @@ Compat = 5
 	patch=1,EE,00463130,word,8000033c
 	patch=1,EE,00463170,word,8000033c
 	patch=1,EE,00463188,word,8000033c
-	//EE code, rearrange code so VU code receives all data at once
+	//EE side, rearrange and clean code so VU0 receives all data at once
 	patch=1,EE,003F8544,word,00000000
 	patch=1,EE,003F8588,word,48A44800
 	patch=1,EE,003F858C,word,4A00D839
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -4682,6 +4682,29 @@ Serial = SCUS-97142
 Name   = Primal - Civilization Is Only Skin Deep
 Region = NTSC-U
 Compat = 5
+[patches = FCD89DC3]
+	author=Raziel I.C.H.I. Znot
+	comment=Performance fix
+	//microcode
+	patch=1,EE,0040F110,word,8000033c
+	patch=1,EE,0040F128,word,8000033c
+	patch=1,EE,0040F168,word,8000033c
+	patch=1,EE,0040F180,word,8000033c
+	//duplicate, not sure that it's used, but patched anyway
+	patch=1,EE,0040F720,word,8000033c
+	patch=1,EE,0040F738,word,8000033c
+	patch=1,EE,0040F778,word,8000033c
+	patch=1,EE,0040F790,word,8000033c
+	//rearrange EE code accordingly
+	patch=1,EE,00399B14,word,00000000
+	patch=1,EE,00399B58,word,48A44800
+	patch=1,EE,00399B5C,word,4A00D839
+	patch=1,EE,00399B60,word,00000000
+	patch=1,EE,00399BA4,word,00000000
+	patch=1,EE,00399C10,word,48A44800
+	patch=1,EE,00399C14,word,4A00D839
+	patch=1,EE,00399C18,word,00000000
+[/patches]
 ---------------------------------------------
 Serial = SCUS-97144
 Name   = PlayStation Underground 5.1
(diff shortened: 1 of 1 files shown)


MrCK1:
Hello, and thanks for the patch. If you wouldn't mind, could you please give a more detailed explanation as to what is being changed/reasoning behind those changes. If you're not confident that some of the patched values are necessary, I think it would be better to remove them. Thanks.

For example:
```
// Due to an edge case with SIF0 DMA that isn't handled properly in PCSX2
	// the game accidentally resets DVD streams, which causes it to crash.
	// This patch forces the streaming module to use different buffers
	// which prevents the DVD streams from being cleared.

**Comment by pgert** [comment]
Are you aware of this?
https://forums.pcsx2.net/Thread-Fixing-unplayable-games?pid=608580#pid608580

The fixes by @prafullpcsx2 are still in trial.

**Comment by RazielZnot** [comment]
> 
> 
> Are you aware of this?
> https://forums.pcsx2.net/Thread-Fixing-unplayable-games?pid=608580#pid608580
> 
> The fixes by @prafullpcsx2 are still in trial.

I wasn't aware about that, I made this patch independently and it seems that my approach is more radical, I opted to totally remove wait loop instead of just only rearranging EE code.

> 
> 
> Hello, and thanks for the patch. If you wouldn't mind, could you please give a more detailed explanation as to what is being changed/reasoning behind those changes. If you're not confident that some of the patched values are necessary, I think it would be better to remove them. Thanks.
> 
> For example:
> 
> ```
> // Due to an edge case with SIF0 DMA that isn't handled properly in PCSX2
> 	// the game accidentally resets DVD streams, which causes it to crash.
> 	// This patch forces the streaming module to use different buffers
> 	// which prevents the DVD streams from being cleared.
> ```

This patch is unrelated to DMA, it patches very lame code which transforms vertices and normals for skinned models. It is totally safe to use it. Original code did some calculations and waited for new data from EE, but it is totally possible to fetch that data before calling VU code. And the loop looks like this:

ibne $vi5, $vi0, +0 (if $vi5 !=0 repeat this check again)

Seems like PCSX2 doesn't treat priorities very well for accessing COP2 registers, so this loop steals all available processing time for nothing.

By the way, during my own testing I noticed few other issues with this game, like SPS in some places and one weird issue with textures when you approach the place where Scree shows his ability to possess statues, the door of the temple doesn't render right in there.

**Comment by RazielZnot** [comment]
Also here is little thing I made for myself but maybe it will be useful for other people:
patch=1,EE,00299468,word,10000035
This code totally removes depth of field, so you can clearly see everything in the distance. I personally hate this effect and in this game it is just obnoxious.

And lastly some hint about black screen in HW mode. It happens only when main characters rendered. When I did research about bad performance, I found the way to disable rendering of all skinned models, that's how I realized what's the issue with this game, and when only level geometry is rendered, there is no problem showing image in HW mode. Seems like game does something weird when it renders those models and as a result plugin treats entire frame as the object to discard.

**Comment by refractionpcsx2** [comment]
> This patch is unrelated to DMA, it patches very lame code which transforms vertices and normals for skinned models. It is totally safe to use it. Original code did some calculations and waited for new data from EE, but it is totally possible to fetch that data before calling VU code. And the loop looks like this:

The comment was an example, we didn't mean copy that verbatim, we want you to comment the patch correctly to explain what the patch does and why it does it.

**Comment by PSI-Rockin** [comment]
It sounds like what's happening is that the VU0 microprogram is expecting COP2 on the EE side to modify a register. PCSX2 has major timing problems with the VUs: it gives micros a headstart of several thousand cycles, so if a VU0 micro expects synchronization with COP2, it could be running hundreds of times needlessly. A similar issue can be seen in 24: The Game.

The patch in its current form gives no useful information to someone who might be able to solve this problem, hence why others are asking for more detailed comments.

**Comment by RazielZnot** [comment]
> as for the ibne $vi5, $vi0, +0 loop, can you show us what is there? (don't forget the instruction after the branch also gets executed) include any relevant register loads etc from before the loop

Here is the entire microcode for vertices:
```
$81e4237c		# 0040f0d0	lqi.xyzw     vf4, ($vi4++)
$000002ff		# 0040f0d4	nop          
$81e5237c		# 0040f0d8	lqi.xyzw     vf5, ($vi4++)
$000002ff		# 0040f0dc	nop          
$81e6237c		# 0040f0e0	lqi.xyzw     vf6, ($vi4++)
$000002ff		# 0040f0e4	nop          
$81e7237c		# 0040f0e8	lqi.xyzw     vf7, ($vi4++)
$000002ff		# 0040f0ec	nop          
$803e433c		# 0040f0f0	move.w       vf30, vf8
$01c821bc		# 0040f0f4	mulax.xyz    ACC, vf4, vf8x
$8000033c		# 0040f0f8	nop          
$01c828bd		# 0040f0fc	madday.xyz   ACC, vf5, vf8y
$8000033c		# 0040f100	nop          
$01c830be		# 0040f104	maddaz.xyz   ACC, vf6, vf8z
$8000033c		# 0040f108	nop          
$01c03b0b		# 0040f10c	maddw.xyz    vf12, vf7, vf0w
$520507ff		# 0040f110	ibne         $vi5, $vi0, +0
$000002ff		# 0040f114	nop          
$8000033c		# 0040f118	nop          
$000002ff		# 0040f11c	nop          
$8000033c		# 0040f120	nop          
$41c0f1bf		# 0040f124	mulaw.xyz    ACC, vf30, vf0w[E]
$10050001		# 0040f128	iaddiu       $vi5, $vi0, $1
$01c96788		# 0040f12c	maddx.xyz    vf30, vf12, vf9x
```

> The patch in its current form gives no useful information to someone who might be able to solve this problem, hence why others are asking for more detailed comments.

Here is the explanation. Patch puts nops in microcode at 0040f110 and 0040f128 to get rid of this loop. Second noped instruction was there to set default state for the next usage. This microcode is called from loop on EE side which feeds list of vertices to it. On EE side I rearranged code so it puts needed value into vf9 before calling the microcode. Before these changes EE code did it after calling microcode and right after it supposed to clear vi5 register so microcode could proceed further. As you can see in the provided snippet, vf9 isn't used until the last instruction, so it is totally safe to put value in it before calling microcode.

On EE side original code looks like this:
```
00399b58:4a00d839    vcallmsr CMSAR0
00399b5c:48a44800    qmtc2 a0, vf9
00399b60:48c02800    ctc2  zero, $vi5
```
Patch swaps places for "vcallmsr CMSAR0" and "qmtc2 a0, vf9" and removes "ctc2  zero, $vi5", it also removes setting register $vi5 to 1 upper in the code, just to keep it clean. Same is done for normals. Technically it is enough to only swap backwards order of these three instructions on EE side without touching VU code, but I decided to clean it properly.

refractionpcsx2:
okay so yes, the game was waiting for the EE to set vi5 to 0 (with that ctc2) before it proceeded, probably to make sure the both are in perfect sync, by rearranging them as you've done it doesn't need to wait, which *should* be fine from the code I've seen.

tadanokojin:
Just make sure explanation is in comment of the patch please. Current comments explain nothing.

RazielZnot:
> 
> 
> Just make sure explanation is in comment of the patch please. Current comments explain nothing.

Do you really need such clutter? Most patches don't explain details either. It should be enough to have provided explanation here.

MrCK1:
> Do you really need such clutter? Most patches don't explain details either. It should be enough to have provided explanation here.

We're aware that many of the older patches lack comments, which is why we're trying to change that. Patches without comments will get lost at some point of time. Overall, it's just easier to include it in the patch to help future contributors if necessary.

refractionpcsx2:
Lacking comments doesn't help us fix the real issues later down the line, it actually leaves us with no idea what the problem is.

RazielZnot:
Ok, done.

refractionpcsx2:
Is the wait loop removal necessary after rearranging the COP2 operations? It doesn't seem like it should make a difference from what little code I've seen.
Also can you add a comment to the top explaining that the reason this patch works is because PCSX2 attempts to run VU0 for thousands of cycles as soon as it is started, so the game will wait endlessly in a wait loop until COP2 gets around to setting those two values which is many thousands of cycles later.

RazielZnot:
> Is the wait loop removal necessary after rearranging the COP2 operations? It doesn't seem like it should make a difference from what little code I've seen.

As I said before, I did it to keep VU code clean, why keep it, if it's totally useless anyway? Less instructions to recompile, though it's negligible, but that's just how I usually do things and I tested these games in such conditions.

> can you add a comment to the top

To the top of what? Pull request or the patch?

Also why this pull request flagged as needs testing? I tested both games, wasted too much time on that, actually, even scanned game's code few times to be sure and if I knew that it will be like that I could have spend that time on something else and released the patch much earlier. It's just a game-specific patch which can't break anything outside. So much bureaucracy here for a simple patch, which could have been made years ago if people just cared to investigate this issue.

refractionpcsx2:
> To the top of what? Pull request or the patch?

To the top of the patch, just so if we look at it, we know exactly what's going on

RazielZnot:
Done.

refractionpcsx2:
Thanks :)

rashvash:
Still no way to run Primal in HW mode?

dhillel:
Here is some info:
https://forums.pcsx2.net/Thread-Fixing-unplayable-games?pid=608968#pid608968

PR #3383, opened 2020-05-12
GSdx: Select Null renderer if feature level isn't supported, useful for VM testing.

lightningterror:


lightningterror:
Will close this as kojin has some ideas to redo the best renderer selection anyway.

PR #3388, opened 2020-05-13
Savestate: Remove unused exeception/variable

MrCK1:
Removes an exception that was never fully implemented to be used with a CRC check.

Removed an unused variable "m_didBios"


LaneSh4d0w:
What's the utility of removing it if it can be implemented in the future ? For now it is not the main priority to do a CRC check, their priority resides in RPCS2's game motor and such.

MrCK1:
> What's the utility of removing it if it can be implemented in the future ? For now it is not the main priority to do a CRC check, their priority resides in RPCS2's game motor and such.

The savestate is already checked against the game CRC (though only by filename). The only time I could see someone running into issues is if they intentionally renamed their savestate files to point to another game.

Also keep in mind that implementing the feature would also require changes to the CDVD plugin (which will hopefully be merged into the emulator at some point in the future)

LaneSh4d0w:
I see. Indeed, if this is already implemented elsewhere in the code it is legitimate.

pcsx2/SaveState.h
```diff
@@ -35,7 +35,7 @@ namespace Exception
 {
 	// ---------------------------------------------------------------------------------------
 	// Savestate Exceptions:
-	//   UnsupportedStateVersion / StateCrcMismatch
+	//   UnsupportedStateVersion
```
turtleli, line 38: `UnsupportedStateVersion` also seems to be unused. Maybe remove it as well?

MrCK1: Looks like you're right, changes pushed.

Issue #3183, opened 2019-12-22
Monster Hunter: Severe Performance issues in OpenGL HW


tadanokojin:
Setting blending to none will resolve the issue but ideally a solution which doesn't require this.
LT reports split of many draws (thousands)
D3D is fine.

Dump: https://dumps.pcsx2.net/index.php/s/67aRriKL4RrGS3N

lightningterror:
On D3D you need to set High blending to run fbmask for this game.
Issue is similar/same to https://github.com/PCSX2/pcsx2/issues/3083

lightningterror:
Game will perform much better on Vulkan, otherwise on Opengl blending accuracy can be dropped to Minimum.

Issue #3083, opened 2019-08-18
[Jojo no Kimyou na Bouken - Ougon no Kaze] 80-100% GPU usage and slowdown on OpenGL, 10-30% usage and 60 FPS on DirectX 11 (GTX 1070)


Reonu:
***PCSX2 version:***
1.5.0-dev-3260-g33571dd

- **Notes**:
  - It's preferred to only report bugs which are present on the [latest development builds](https://pcsx2.net/download/development.html).
  - The PSX backward compatibility mode is still in its early stage of development. Issue reports are too early at this point.

***PCSX2 options:***
Default speedhacks + MTVU and fast CDVD

***Plugins used:***
GSDX AVX
Running at 3X native, everything else default

GSDX dump in OpenGL mode (GPU usage was 99% and there was slowdown):
[OpenGL.zip](https://github.com/PCSX2/pcsx2/files/3513290/OpenGL.zip)
GSDX dump in DX11 mode (GPU usage was 10% and it was 60 FPS):
[DX11.zip](https://github.com/PCSX2/pcsx2/files/3513291/DX11.zip)
***Description of the issue:***
In DX11, the performance is perfect and the GPU usage is very low. In OpenGL, the emulator is suddenly GPU bottlenecked. 

***How to reproduce the issue:***
Play with OpenGL and with DX11, compare performance

***Last known version to work:***
Don't know.

***PC specifications:***
i7 6700k @ 4.2 GHz
16 GB DDR4 2400 MHz 15-15-15-35 RAM
GTX 1070
Windows 10 64 bit

lightningterror:
It might be blending related, I'm seeing `FBMASK SW emulated fb_mask:1 on 16 bits format` used a lot here. Disabling Blending provides quite a high performance boost on my end.

Can you compare both dx and gl with blending disabled, next compare dx blending set to high vs gl blending set to basic.

Leon422:
Note: On DX11, the emulator caps itself to 144 FPS (my refresh rate) no matter what, even though vsync is "disabled". If PCSX2 is not in focus, this limit dissapears. So for the DX11 tests I had Firefox on focus instead of PCSX2.
On the same scene:
GL with blending disabled: ~260 FPS
DX11 with blending disabled: ~350 FPS
DX11 with blending on high:  ~40 FPS
OpenGL with blending on basic: ~58 FPS

Do you need dumps for all these cases?

lightningterror:
[...]

PR #3406, opened 2020-05-19
GSdx-ogl: Optimize SW Fbmask.

lightningterror:
Optimization idea:
Monster hunter does SW FBMASK on almost every draw per frame on red
channel with a draw split of over 1k in many situations.
Idea is to use Unsafe path when draw split is above 500 on Basic
Blending.
SW Path above 500 will be active on High
Blending.

Issues #3183 #3083


lightningterror:
Closing as it was discussed a better solution would be rov support.

Issue #3408, opened 2020-05-21
something with the interface? i don't know


Danilopereira82:
***PCSX2 version:***

1.7.0.r45.f2560998c-1
the one from the AUR

- **Notes**:
  Interface is initializing.  Entering Pcsx2App::OnInit!
Applying operating system default language...
Loading language translation databases for 'Portuguese (Brazilian)' [pt_BR]
Command line parsing...
Command line parsed!
terminate called after throwing an instance of 'std::length_error'
  what():  basic_string::_M_create
Abortado (imagem do núcleo gravada)

***PCSX2 options:***
default, i didn't change anything

***Plugins used:***
everything default

***Description of the issue:***
looks like the language is causing some problem
because when i delete the config file the emulator works normal

***How to reproduce the issue:***

just start the emulator and set the language to system default
and when i close i can't open it again

***PC specifications:***
``(For example: CPU: Intel Core 2 Duo E4600 GPU: Gigabyte GTX 750Ti OS: Windows 10 64 Bit)``
intel i5-7200U and HD Graphics 620
arch linux

kenshen112:
When did you install the package?

Danilopereira82:
some minutes ago
i was testing castlevania and closed the emulator
when i tried to open the emulator again i got this log

kenshen112:
I'm the package maintainer. Can you uninstall it reinstall it making sure to select the clean build option then see if the issue occurs again?

Danilopereira82:
the only debugging i can do is read the log in the terminal lol

Danilopereira82:
while we are here i wanted to say that i tried to run the iso for god of war
and the emulator just crashes the entire screen
i don't know what, but i don't think its because its using too much ram
when i boot the game i just see the first text and the screen just stops
the mouse keeps working and the system works fine
i can change the screen with ctrl + alt + f2 etc
but everything in f7 except the mouse doesn't work

kenshen112:
Wait, you got it to start again?

Danilopereira82:
yes the emulator works if i delete the config files

Danilopereira82:
something is wrong with 
"terminate called after throwing an instance of 'std::length_error'
what(): basic_string::_M_create"

kenshen112:
What happens if you clone the repository, with 

`git clone https://github.com/PCSX2/pcsx2.git` go into the generated pcsx2 folder run ./build.sh and try to run the generated portable version?

[...]

PR #3411, opened 2020-05-21
This is a fix for a Voodoo issue that occurs on Arch Linux AUR.

kenshen112:
This is a PR to fix what could be considered a voodoo issue that occurs on Arch Linux with the AUR. I say voodoo as this issue is known not to happen to everyone and has happened with different circumstances. If I build the PKGBUILD file with the clang compiler the crash this amends will occur but not with GCC. As of issue #3408 it's discovered this can occur with the GCC compiler as well. Language selection has no known effect on the issue.

The issue this pr is stated to amend is thus. For previous users of pcsx2 there are settings files that pcsx2 will attempt to load. Doing so causes the WXstring in IniInterface.cpp Ln 173 to report an incorrect length thus crashing pcsx2. Clearing all settings files and folders is known to work around the issue being discussed. 
```
 12:49:07: Warning: Mismatch between the program and library build versions detected.
The library used 3.0 (wchar_t,compiler with C++ ABI 1011,wx containers,compatible with 2.8),
and your program used 3.0 (wchar_t,compiler with C++ ABI 1002,wx containers,compatible with 2.8).

(PCSX2:90102): Gtk-WARNING **: 12:49:07.439: Unable to locate theme engine in module_path: "adwaita",

(PCSX2:90102): Gtk-WARNING **: 12:49:07.442: Unable to locate theme engine in module_path: "adwaita",
Interface is initializing.  Entering Pcsx2App::OnInit!
Applying operating system default language...
Command line parsing...
Command line parsed!
terminate called after throwing an instance of 'std::length_error'
  what():  basic_string::_M_create
 Aborted (core dumped)
```

Is the prior error this pr deals with

### DIFF
--- a/common/src/Utilities/IniInterface.cpp
+++ b/common/src/Utilities/IniInterface.cpp
@@ -170,7 +170,7 @@ void IniLoader::Entry(const wxString &var, uint &value, const uint defvalue)
 void IniLoader::Entry(const wxString &var, bool &value, const bool defvalue)
 {
     // TODO : Stricter value checking on enabled/disabled?
-    wxString dest(defvalue ? L"enabled" : L"disabled");
+    wxString dest(defvalue ? "enabled" : "disabled", 8);
     if (m_Config)
         m_Config->Read(var, &dest, dest);
     value = (dest == L"enabled") || (dest == L"1");


Danilopereira82:
i was mistaken
i compiled in a terminal with clang in the environment variables

PR #3414, opened 2020-05-23
GSdx-hw: EmulateZbuffer changes

### LINKED ISSUE #2271: F1 2004 : starting lights are black instead of red
***PCSX2 version:***
Last 1.5dev

***PCSX2 options:***
All default, no speedhacks, no hacks can solve this.

***Plugins used:***
All default

The game don't display starting lights as it should, they are black instead of red, look:

-OpenGL hardware BA full:
capture

-Software render (with a more closer picture):
capture2

-Real PS2:

Load a race in F1 2004 Under PCSX2

***Last version to work: PCSX2 0.94***

***PC specifications:***
AMD fx8350 GTX 1060Ti Windows 7 64 Bit

***Note:***
-Like https://github.com/PCSX2/pcsx2/issues/2108 Blockdump don't even work with the cdvdLinuz plugin.

### LINKED ISSUE #2623: Tomb Raider: Angel of Darkness PAL - Missing (obscured) notebook illustrations
***PCSX2 version:***

PCSX2 v1.5.0-dev-2584-g940263d49

***PCSX2 options:***

Problem only occurs while using software rendering (any backend) or OpenGL hardware.

***Plugins used:***

Default

Early in the game Lara obtains a notebook containing various pieces of info. This notebook is accessible through the pause menu (select button in game), and some of the entries in the notebook are accompanied by pictures. At least one of these is needed to solve a puzzle. 

The problem is that the pictures are not usually visible in PCSX2 in software mode or using OpenGL hardware. They are there, but they seem to be being rendered behind the notebook graphic rather than in front of it (setting VU round mode to nearest shrinks the notebook vertically, revealing part of the image behind it) :

gsdx_20181007215802

This problem is present when using OpenGL hardware mode, or any software mode. Messing with EE or VU settings achieves nothing other than further graphic problems, however, for some reason the images seem to display just fine when using the DirectX 9 or 11 hardware renderers:  

gsdx_20181007213004

I haven't found any GSdx settings that have an impact other than changing the backend. Also, while the DX11 renderer is largely fine, it doesn't display the icons when you pick up items, only the dark circle that should be behind the icon. I don't know if that could be related (though on testing I see that the game's missing shadows in DX11 too, so probably not).

GSDump: 

[TombRaiderAOD.zip](https://github.com/PCSX2/pcsx2/files/2454323/TombRaiderAOD.zip)

Incidentally, it's probably unrelated but this game also has issues with garbage polygons sometimes appearing in the viewport. This usually happens when the camera view is looking up towards Lara and a wall, floor, or object would really be blocking the view (the polygons are garbage though, not a properly rendered wall).
 
***PC specifications:***

Intel i5 3570K, Nvidia GTX 1080, Windows 10 1809

### LINKED ISSUE #3061: GSdx-HW: Rozen Maiden DuellWalzer - Z buffer issue maybe
Current build v1.5.0-dev-3221

Rozen Maiden DuellWalzer has an issue where some text doesn't appear in HW mode. SW mode is fine.
It seems that we trigger `ASSERT(m_vt.m_min.p.z > max_z);` several times
https://github.com/PCSX2/pcsx2/blob/dbfe21c444bea89abfe48e1cf03f5aee346e3e7b/plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp#L190
In these cases `m_vt.m_min.p.z` is 0.

Comparison between hw and sw mode:
SW:
gsdx_20190805095108

HW:
gsdx_20190805095117

First difference between hw and sw is on draw 4 rz1:
SW:
_00004_f5000_rz1_01a40_Z_16S (2)

HW:
_00004_f5000_rz1_01a40_Z_16S

Context of draw 4:
https://pastebin.com/Mu6VH0Xb

First actual visual difference I see is on draw 56 rt1

SW:
_00056_f5000_rt1_008c0_C_24

HW:
_00056_f5000_rt1_008c0_C_24 (2)

Context of 56:
https://pastebin.com/a8sQGuT6

GS Dump:
[Rozen Maiden Duellwalzer.zip](https://github.com/PCSX2/pcsx2/files/3466469/Rozen.Maiden.Duellwalzer.zip)

### LINKED ISSUE #3396: Midnight Club 3 Character Bug [HW]
***PCSX2 version:***
``pcsx2-v1.7.0-dev-37-ga30a6583c-windows-x86``

***PCSX2 options:***
Default. Safe preset. Didn't change any of settings beside renderer.

***Plugins used:***
Only happens with D3D11/OGL in [HW]
Software renderer is fine.

https://imgur.com/a/LJdNiOU
[emuLog.txt](https://github.com/PCSX2/pcsx2/files/4640015/emuLog.txt)
https://files.catbox.moe/hpavo5.7z

Character missing when trying to crate a new profile.

Just try to crate a new profile after booting the game.

Unknown.

***PC specifications:***
``CPU: Ryzen 5 2400G GPU: Radeon™ RX Vega 11 Graphics 20.4.1 OS: Windows 10 64 Bit 1909)``

lightningterror:
gsdx-d3d11: Port/add depth mask support to EmulateZbuffer.
Add support for depthmasking to EmulateZbuffer, previous old code had
support but wasn't ported properly with the new code a few years back.
VS Constant buffer is now properly setup.

gsdx-hw: Clamp zbuffer to depthmask instead of wrapping.
The behavior was verified on Dobie to be correct.
The code needs to be ported to SW renderer too to
improve rendering on SW side.
Current PR will fix plenty of games on HW renderer
that had/have zbuffer issues before.
v2. Set DepthMask to the maximum the current depth format allows.
Will properly clamp for 16bit and 24bit formats.
v3. gl: Fix uniform buffer upload/cache for VSConstantBuffer.

gsdx-hw: Remove the code block for the bad Z sise check in EmulateZbeffer.
Also fixes Itadaki Street text and F1 2004 starting lights.

gsdx-hw: Cleanup a bit EmulateZbuffer.
Update the comment to reflect recent changes,
Rename DepthMask to MaxDepth,
Initialize max_z to 0 at start so it doesn't complain,
Do vs cb padding on d3d11.

gsdx-hw: Do z clamping only on sprites.
For some reason for triangle primitives even without clamping games look
fine, doing clamping on ps/fs will be more costly so let's leave it out
till we find a game that has issues.

gsdx-hw: Add zclamping to ps/fs.
Add zclamping to ps/fs, enable vs, ps/fs clamp when needed with a macro.

~~Fix also needs to be ported to SW renderer.~~

Fixes #2623
Fixes #2271
Fixes #3061
Fixes #3396

### DIFF
--- a/plugins/GSdx/Renderers/DX11/GSDevice11.h
+++ b/plugins/GSdx/Renderers/DX11/GSDevice11.h
@@ -105,7 +105,7 @@ public:
 		GSVector4i FbMask;
 
 		GSVector4 TC_OffsetHack;
-		GSVector4 Af;
+		GSVector4 Af_MaxDepth;
 		GSVector4 DitherMatrix[4];
 
 		PSConstantBuffer()
@@ -118,7 +118,7 @@ public:
 			MskFix = GSVector4i::zero();
 			ChannelShuffle = GSVector4i::zero();
 			FbMask = GSVector4i::zero();
-			Af = GSVector4::zero();
+			Af_MaxDepth = GSVector4::zero();
 
 			DitherMatrix[0] = GSVector4::zero();
 			DitherMatrix[1] = GSVector4::zero();
@@ -241,6 +241,9 @@ public:
 				// Dithering
 				uint32 dither:2;
 
+				// Depth clamp
+				uint32 zclamp:1;
+
 				// Hack
 				uint32 tcoffsethack:1;
 				uint32 urban_chaos_hle:1;
@@ -248,7 +251,7 @@ public:
 				uint32 point_sampler:1;
 				uint32 invalid_tex0:1; // Lupin the 3rd
 
-				uint32 _free:16;
+				uint32 _free:15;
 			};
 
 			uint64 key;
--- a/plugins/GSdx/Renderers/DX11/GSDevice11.h
+++ b/plugins/GSdx/Renderers/DX11/GSDevice11.h
@@ -42,11 +42,14 @@ public:
 		GSVector4 VertexOffset;
 		GSVector4 Texture_Scale_Offset;
 
+		GSVector2i DepthMask;
+
 		VSConstantBuffer()
 		{
-			VertexScale = GSVector4::zero();
-			VertexOffset = GSVector4::zero();
+			VertexScale          = GSVector4::zero();
+			VertexOffset         = GSVector4::zero();
 			Texture_Scale_Offset = GSVector4::zero();
+			DepthMask            = GSVector2i(0);
 		}
 
 		__forceinline bool Update(const VSConstantBuffer* cb)
@@ -74,11 +77,10 @@ public:
 		{
 			struct
 			{
-				uint32 bppz:2;
 				uint32 tme:1;
 				uint32 fst:1;
 
-				uint32 _free:28;
+				uint32 _free:30;
 			};
 
 			uint32 key;
--- a/plugins/GSdx/Renderers/DX11/GSDevice11.h
+++ b/plugins/GSdx/Renderers/DX11/GSDevice11.h
@@ -41,15 +41,16 @@ public:
 		GSVector4 VertexScale;
 		GSVector4 VertexOffset;
 		GSVector4 Texture_Scale_Offset;
-
-		GSVector2i DepthMask;
+		GSVector2i MaxDepth;
+		GSVector2i pad_vscb;
 
 		VSConstantBuffer()
 		{
 			VertexScale          = GSVector4::zero();
 			VertexOffset         = GSVector4::zero();
 			Texture_Scale_Offset = GSVector4::zero();
-			DepthMask            = GSVector2i(0);
+			MaxDepth             = GSVector2i(0);
+			pad_vscb             = GSVector2i(0);
 		}
 
 		__forceinline bool Update(const VSConstantBuffer* cb)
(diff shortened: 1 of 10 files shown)


plugins/GSdx/Renderers/DX11/GSDevice11.h
```diff
@@ -42,11 +42,14 @@ class GSDevice11 final : public GSDevice
 		GSVector4 VertexOffset;
 		GSVector4 Texture_Scale_Offset;
 
+		GSVector2i DepthMask;
```
AlessandroVetere, line 45: Why isn't the mask stored in a single unsigned integer?

plugins/GSdx/Renderers/DX11/GSRendererDX11.cpp
```diff
@@ -159,39 +159,17 @@ void GSRendererDX11::EmulateZbuffer()
...
-			{
-#ifdef _DEBUG
-				fprintf(stdout, "%d: Bad Z size on %s buffers\n", s_n, psm_str(m_context->ZBUF.PSM));
-#endif
-				m_om_dssel.ztst = ZTST_ALWAYS;
-			}
-		}
-	}
+	// On the real GS we appear to do clamping on the max z value the format allows.
+	// Clamping is done after rasterizaion.
```
lightningterror, line 171: Typo, need to correct.

plugins/GSdx/Renderers/DX11/GSDevice11.h
lightningterror: When I initially ported it I only followed what was done on gl but looking at it now you're right.
I don't see any use case where two vectors will be used in the shader.

lightningterror:
I ran a final test on a bunch of dumps, as well as ps2 test. Everything checks out so let's merge it.

PR #3418, opened 2020-05-23
GameDB: Add Xenosaga save point crash patch

RedPanda4552:
Adds the Xenosaga save point crash patch to GameDB.

Patch location on the forums: https://forums.pcsx2.net/Thread-Post-your-PCSX2-cheats-patches-here?pid=603236#pid603236

The two questions that need answered:

1. Is the patch proven to be stable?
2. Is it a good idea to, even though PCSX2 accuracy makes it worse than a real PS2, patch something that's technically correct behavior for a better user experience?

My answers:

1. Yes, it's been in the field for a while and there's been several hundred downloads with no reported issues.
2. I do think so, simply in that users would appreciate not having to manually search up and retrieve the patch.

There's probably plenty of conflicting opinions, so lets discuss. To merge, or not to merge?

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -4409,6 +4409,14 @@ Region = NTSC-J
 Serial = SCPS-55901
 Name   = Xenosaga Episode I - Der Wille zur Macht
 Region = NTSC-J
+[patches = A3D63039]
+	// Reduces JPEG quality to its minimum for save file thumbnails. This prevents
+	// the game from calling its own exit function, which was strangely its safety
+	// mechanism for when a captured thumbnail exceeded 4 KB in size. 
+	comment=Save Point Crash Prevention
+	author=pandubz, PSI, turtleli
+	patch=1,EE,20285f48,extended,24040000
+[/patches]
 ---------------------------------------------
 Serial = SCPS-55902
 Name   = Gran Turismo - Concept 2002 Tokyo-Geneva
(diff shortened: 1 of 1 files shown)


RedDevilus:
Well i would just merge it , just looking at the downloads alone (around a 1000) and no report that it's bad.

tadanokojin:
I'm fine with it. I wish we had a nice cheat system but we don't so I don't see a better way for now.

lightningterror:
If the patch is stable and tested properly I'm fine with it, otherwise no.

bin/GameIndex.dbf
```diff
@@ -34767,6 +34791,14 @@ vuClampMode = 3 // fix white shiny weapons
...
 Name   = Xenosaga Episode 1 - Der Wille zur Macht [Reloaded]
 Region = NTSC-J
+[patches = A3D63039]
+	// Reduces JPEG quality to its minimum for save file thumbnails. This prevents
+	// the game from calling its own exit function, which was strangely its safety
+	// mechanism for when a captured thumbnail exceeded 4 KB in size. 
+	comment=Save Point Crash Prevention
+	author=pandubz, PSI, turtleli
+	patch=1,EE,20285f48,extended,24040000
+[/patches]
```
turtleli, line 34801: I think this patch should be removed. Reloaded is different from the other NTSC-J versions.

bin/GameIndex.dbf
```diff
@@ -35182,6 +35214,14 @@ Region = NTSC-J
 Serial = SLPS-73901
 Name   = Xenosaga Episode 1 - Der Wille zur Macht [PlayStation 2 The Best]
 Region = NTSC-J
+[patches = A3D63039]
```
turtleli, line 35217: Side note (doesn't need to be addressed): I don't think this GameDB entry is actually accessible. Redump info shows that this disc is identical to SLPS-29002.

pgert:
@RedDevilus:
NTSC-J: "Xenosaga Episode 1 - Chikara e no Ishi" ?

RedDevilus:
> 
> 
> @RedDevilus:
> NTSC-J: "Xenosaga Episode 1 - Chikara e no Ishi" ?

What about it?

pgert:
Is it correct - should it be "Chikara e no Ishi" or "Der Wille zur Macht" in JP-versions?

RedDevilus:
Xenosaga Episode I: Chikara e no Ishi is the correct one, but it hasnt been added to the GameDB. Nitpicky i guess.
ゼノサーガ エピソードI 力への意志 If u want to be correct in the Kanji.

pgert:
There are 4 JP "Xenosaga Episode I" discs, all affected by this PR.
My point is: are the names of these discs as they should be?

RedDevilus:
Well that's just the localised name, it's the same but ask the OP if it needs to be changed.

PR #3419, opened 2020-05-23
Appveyor: Drop vs2017 builds.

lightningterror:
It will cut compile time half.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,10 +9,6 @@ environment:
     target: AppVeyor
     visualstudio_string: vs2019
     APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
-  - platform: Win32
-    target: AppVeyor
-    visualstudio_string: vs2017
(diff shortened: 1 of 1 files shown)


tadanokojin:
Hm rather than drop build, I think Li got pcsx2 in github actions. It should be faster than appveyor. Perhaps we should look into that instead?

lightningterror:
Sounds like a plan, will leave this one open anyway, maybe a temporary improvement for appveyor till then?

tadanokojin:
Yeah let's see what @turtleli thinks.

turtleli:
IMO merging this PR means dropping support for compiling with VS2017, because we won't know when a commit or PR will break VS2017 compilation if we don't have a buildbot compiling with it.

Personally I would drop it because VS2019 community is free for open source anyway, so it shouldn't be too much of an issue for people to use it.

As for Actions - It's worth considering. I think it allows 20 concurrent jobs (free plan), and testing on my fork suggests that it's slightly faster per job.

lightningterror:
Closing as it was discussed github actions will be a better alternative.

PR #3420, opened 2020-05-24
Switch over the null plugins dialog code to use wxWidgets instead of gtk on Linux.

arcum42:
I left the windows code as-is because I didn't want to accidentally break the windows build, though if anyone else wants to make windows use the wx code as well, it'd be a good idea.

I'd largely like to get rid of as much gtk code as we can, regardless of whether we end up getting rid of wx at some point.

Naturally, other plugins should be converted to wx dialogs as well, but one step at a time...

### DIFF
--- a/common/include/PS2Eext.h
+++ b/common/include/PS2Eext.h
@@ -32,8 +32,8 @@
 
 #elif defined(__unix__)
 
-#include <gtk/gtk.h>
 #include <cstring>
+#include <wx/msgdlg.h>
 
 #define EXPORT_C_(type) extern "C" __attribute__((stdcall, externally_visible, visibility("default"))) type
 
@@ -203,7 +203,6 @@ struct PluginConf
 };
 
 #if defined(__unix__)
-
 static void SysMessage(const char *fmt, ...)
 {
     va_list list;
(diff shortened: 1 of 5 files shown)


common/include/PS2Eext.h
```diff
@@ -216,44 +215,8 @@ static void SysMessage(const char *fmt, ...)
...
-    /* Add all our widgets, and show everything we've added to the dialog. */
-    gtk_container_add(GTK_CONTAINER(gtk_dialog_get_content_area(GTK_DIALOG(dialog))), label);
-    gtk_container_add(GTK_CONTAINER(gtk_dialog_get_content_area(GTK_DIALOG(dialog))), check_box);
-    gtk_dialog_add_button(GTK_DIALOG(dialog), "Ok", 0);
-
-    gtk_widget_show_all(dialog);
-
-    gtk_dialog_run(GTK_DIALOG(dialog));
-    gtk_widget_destroy(dialog);
+    wxMessageDialog *dial = new wxMessageDialog(NULL, msg, wxT("Info"), wxOK);
```
turtleli, line 218: This doesn't get deleted, so I think it'll cause a memleak.

I don't think it's necessary to use a pointer in this case since it isn't a child window. You should be able to just do
```
wxMessageDialog dialog(nullptr, msg, "Info", wxOK);
dialog.ShowModal();
```

arcum42: Alright, taken care of.

lightningterror:
> I left the windows code as-is because I didn't want to accidentally break the windows build, though if anyone else wants to make windows use the wx code as well, it'd be a good idea.

It doesn't use wx, uses windows gui.

arcum42:
Right, I changed the code from using gtk on Linux and the Windows gui on Windows to use wx on Linux, but left the Windows part alone. We potentially could use the wx code for both...

PR #3421, opened 2020-05-25
macOS gsdx

tellowkrinkle:
Makes GSdx compile on macOS and brings back the required feature workarounds for macOS's OpenGL 4.1 implementation

With any luck, the revived feature workarounds will also allow OpenGL to be used from within wine32on64 as well for all the people running that (since it's the only way that currently works on Catalina), but I don't have a computer set up to build windows versions to test

A few things to note / questions:
- The `Fix crash on shutdown in macOS` commit isn't for GSdx (I forgot about it in the previous PR).  I feel like the change I made in it should work on linux as well, but [the comment in `MmapResetPtr`](https://github.com/PCSX2/pcsx2/blob/5903ee95fe05df1684791ac32da3dcd1707414f3/common/src/Utilities/Linux/LnxHostSys.cpp#L181-L182) makes me doubt myself...  Does anyone know why mmap with `MAP_FIXED` might not do what we want?  For now it's ifdef'd for macOS only but if it does work on Linux we could use it all the time
- Does anyone know how often [this happens](https://github.com/PCSX2/pcsx2/blob/5903ee95fe05df1684791ac32da3dcd1707414f3/plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp#L183-L185)?  To avoid all the precision issues we were having with the old ARB_clip_control workaround (logz helped but still didn't work super well), I wanted to divide by `DepthMask` instead of 2^31 (so if a game is using 16-bit depth, its values actually go from -1 to 1 instead of -1 to -1+2^-15), but that requires turning off this "always use 32-bit depth" workaround for the listed issue, and I'd rather not break more games than I fix this way.


tellowkrinkle:
Actually it looks like #3414 gets rid of the thing I was asking about in the second bullet so if you merge that first I can rebase on top of it and everything will be happy

lightningterror:
I'm not really a fan of bringing a lot of legacy stuff back, idea was to bump up gl context to minimum 4.3 or 4.4 depending on time and use more modern features (more mandatory extensions).

arcum42:
The trouble there is that OpenGL on Mac OS never got updated past OpenGl 4.1, and is deprecated. The only other real options on Mac OS X would be a Metal renderer, or more likely, a Vulkan renderer with a Metal compatibility layer.

wine would use dxvk to translate DirectX into Vulkan, but I don't know if there's some way to use that without wine.

lightningterror:
> The only other real options on Mac OS X would be a Metal renderer, or more likely, a Vulkan renderer with a Metal compatibility layer.

It is the better/best solution since it lets us move forward with progress.
Otherwise I don't want to regress gsdx to a 1.4 release state so I'm against adding/merging these changes.

orbea:
I personally feel strongly against bending over backwards for a proprietary platform and their deprecated graphics driver that will only get worse in time.

tellowkrinkle:
Okay.  We'll wait on a Vulkan or Metal renderer then
Note: If anyone is looking for these changes in the future I'll try to keep them up to date on my [MacOS branch](https://github.com/tellowkrinkle/pcsx2/tree/MacOS)

uyjulian:
Why not add this to `GSdx_legacy` instead ?

orbea:
My impression that GSdx_legacy is mostly for regression testing at this point and may eventually be removed.

nonoche2:
> Okay. We'll wait on a Vulkan or Metal renderer then

there's one here : https://github.com/KhronosGroup/MoltenVK

tellowkrinkle:
> there's a Vulkan to Metal wrapper here : https://github.com/KhronosGroup/MoltenVK

Hence the wait for a Vulkan **or** Metal renderer

PR #3423, opened 2020-05-25
PCSX2-GUI: Update BackgroundLogo

Ryudo300:
~~Expanded the height of the logo to 200px from the original 180px. This change was done to fix the lines appearing when using 125% DPI mode in Windows. WX scaling apparently wasn't a fan of the original.~~

~~- Only 125% is fixed for now. 150% is still showing lines.~~

Update: Fixed up the original logo so I was able to keep the 180px intact. This now works on both 125% DPI and 150% DPI scaling.


Ryudo300:
Some notes. Turns out that this only fixes it for 125% DPI scaling at the moment. I didn't know it breaks -again- on 150% DPI scaling.

Ryudo300:
Victory over WX! 125% and 150% fixed with 180px height (original size logo).

PR #3424, opened 2020-05-26
Fix a few crashes in macOS

tellowkrinkle:
I feel like the change for the shutdown crash I made should work on linux as well, but [the comment in `MmapResetPtr`](https://github.com/PCSX2/pcsx2/blob/5903ee95fe05df1684791ac32da3dcd1707414f3/common/src/Utilities/Linux/LnxHostSys.cpp#L181-L182) makes me doubt myself...  Does anyone know why mmap with `MAP_FIXED` might not do what we want?  For now it's ifdef'd for macOS only but if it does work on Linux we could use it all the time

### DIFF
--- a/common/src/Utilities/Linux/LnxHostSys.cpp
+++ b/common/src/Utilities/Linux/LnxHostSys.cpp
@@ -157,6 +157,15 @@ void *HostSys::MmapReservePtr(void *base, size_t size)
     return mmap(base, size, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
 }
 
+#ifdef __APPLE__
+static void * MmapReservePtrFixed(void *base, size_t size)
+{
+    PageSizeAssertionTest(size);
+
+    return mmap(base, size, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0);
+}
+#endif
+
 bool HostSys::MmapCommitPtr(void *base, size_t size, const PageProtectionMode &mode)
 {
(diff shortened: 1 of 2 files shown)


tadanokojin:
thank you for your hard work

arcum42:
I tested, and using MmapReservePtrFixed in Linux does seem like it works, though it's always possible a crash might show up later down the line. Still, I launched a game, paused and resumed, swapped to a few other games, chose shutdown, and closed and opened the emulator without any issues...

tellowkrinkle:
The one other thing to check would be does memory usage still go down when you choose shutdown

There is a chance that the OS won't consider the memory free after `MmapReservePtrFixed`.  In my testing on macOS, doing an `mprotect` with `PROT_NONE` didn't free the memory but remapping using `mmap` with `MAP_FIXED` did

common/src/Utilities/Linux/LnxHostSys.cpp
```diff
@@ -157,6 +157,15 @@ void *HostSys::MmapReservePtr(void *base, size_t size)
     return mmap(base, size, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
 }
 
+#ifdef __APPLE__
```
lightningterror, line 160: If MmapReservePtr isn't used anywhere else on mac os then maybe you can put the function in else macro here too. Just a minor optimisation.

tellowkrinkle:
Okay I tested on Linux and overwriting with `MAP_FIXED` frees memory as expected

I also inlined the mmap fixed as recommended by @lightningterror

PR #3435, opened 2020-06-01
NSIS: Add support for command line flags and fix silent install

### LINKED ISSUE #3431: [Feature Request] Command line options for Windows installer
It would be nice if the Windows installer could have command line flags/options added/supported. 

There are two options that come by default with NSIS-
`/S`  - Silent install. It appears to not work for 1.6.0
`/D` - Change install directory. It works for portable installation but does not work for normal installation as far as I can tell.

Then there are a couple more that I think would be nice to have added. The exact name of the option can be changed if desired. 
`/NoDesktop` - Do not create a desktop icon.
`/NoStart` - Do not create a start menu shortcut. 
`/Portable` or `/Install` - to switch the install type.
`/?` and `/h` - help message to document these options.

TheCakeIsNaOH:
Fixes #3431 

- Fixes `/D` install directory override, works for both portable and install
- Fixes `/S` silent install, defaults to install
- Adds `/?` and `/H` which show a help message with the available flags
- Adds `/NoStart` and `/NoDesktop` which respectively disable the creation of start menu and desktop shortcuts.
- Adds `/Portable`, which switches the installer over to portable mode if it is silent(`/S`) as opposed to the default mode of install.

### DIFF
--- a/nsis/SharedDefs.nsh
+++ b/nsis/SharedDefs.nsh
@@ -3,6 +3,8 @@
 ; These definitions are shared between the 2 installers (pre-install/portable and full)
 ; This reduces duplicate code throughout both installers.
 
+!include "FileFunc.nsh"
+
 ManifestDPIAware true
 Unicode true
 ShowInstDetails nevershow
@@ -62,4 +64,58 @@ ${If} $UserPrivileges == "Admin"
     ${ElseIf} $UserPrivileges == "User"
     StrCpy $IsAdmin 0
 ${EndIf}
+FunctionEnd
+
+Function ShowHelpMessage
+  !define line1 "Command line options:$\r$\n$\r$\n"
+	!define line2 "/S - silent install (must be uppercase)$\r$\n"
+	!define line3 "/D=path\to\install\folder - Change install directory$\r$\n"
+	!define line4 "   (Must be uppercase, the last option given and no quotes)$\r$\n$\r$\n"
+	!define line5 "/NoStart - Do not create start menu shortcut$\r$\n"
+	!define line6 "/NoDesktop - Do not create desktop shortcut$\r$\n"
+  !define line7 "/Portable- Install in portable mode instead of full install, no effect unless /S is passed as well"
+	MessageBox MB_OK "${line1}${line2}${line3}${line4}${line5}${line6}${line7}"
+	Abort
+FunctionEnd
+
+Function .onInit
+    var /GLOBAL cmdLineParams
+    Push $R0
+    ${GetParameters} $cmdLineParams
+    ClearErrors
+
+    ${GetOptions} $cmdLineParams '/?' $R0
+    IfErrors +2 0
+    Call ShowHelpMessage
+
+  	${GetOptions} $cmdLineParams '/H' $R0
+    IfErrors +2 0
+    Call ShowHelpMessage
+
+    Pop $R0
+
+
+    Var /GLOBAL option_startMenu
+    Var /GLOBAL option_desktop
+    Var /GLOBAL option_portable
+    StrCpy $option_startMenu     1
+  	StrCpy $option_desktop       1
+    StrCpy $option_portable      0
+
+    Push $R0
+
+  	${GetOptions} $cmdLineParams '/NoStart' $R0
+    IfErrors +2 0
+  	StrCpy $option_startMenu 0
+
+    ${GetOptions} $cmdLineParams '/NoDesktop' $R0
+    IfErrors +2 0
+    StrCpy $option_desktop 0
+
+    ${GetOptions} $cmdLineParams '/Portable' $R0
+    IfErrors +2 0
+    StrCpy $option_portable 1
+
+    Pop $R0
+    
 FunctionEnd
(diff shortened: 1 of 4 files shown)


TheCakeIsNaOH:
Is there anything I can do to help this get merged?

MrCK1:
Sorry for the delay, I will get back to you when I can. I'm a bit busy with work at the moment.

nsis/SharedDefs.nsh
```diff
@@ -62,4 +64,58 @@ ${If} $UserPrivileges == "Admin"
     ${ElseIf} $UserPrivileges == "User"
     StrCpy $IsAdmin 0
 ${EndIf}
+FunctionEnd
+
+Function ShowHelpMessage
+  !define line1 "Command line options:$\r$\n$\r$\n"
+	!define line2 "/S - silent install (must be uppercase)$\r$\n"
```
lightningterror, line 71: Use spaces for indentation as the rest of the file, formatting could be a bit better too.

TheCakeIsNaOH: How does it look now?

MrCK1:
Sorry again for the delay, everything seems to work fine except for the directory override - it still is installing to the default directory on my end.

Also, silent uninstall will still throw this dialog if a previous version is installed - so you'll have to modify the code to get around that as well unless you'd rather have it displayed.

image

TheCakeIsNaOH:
I went ahead and selected a default choice, ok, for the message box.

The directory override is working on my end. Are you having issues with the install or the portable?

MrCK1:
Neither is able to work for me, can you paste an example of your syntax when running with the command line option?

TheCakeIsNaOH:
`pcsx2-1.6.0-setup.exe /S /Portable /D=C:\users\user\desktop\pcsx2`
or
`pcsx2-1.6.0-setup.exe /S /D=C:\Program Files (x86)\override`

Due to how NSIS is programmed, the D has to be uppercase and be the last argument given.

Also, if you are running from Powershell, use something like this.
`pcsx2-1.6.0-setup.exe --% /S /D=C:\Program Files (x86)\override` 
The `--%` tells PowerShell to stop interpreting the arguments, and just pass them directly to the program.

nsis/SharedDefs.nsh
```diff
@@ -62,4 +64,57 @@ ${If} $UserPrivileges == "Admin"
...
+    var /GLOBAL cmdLineParams
+    Push $R0
+    ${GetParameters} $cmdLineParams
+    ClearErrors
+
+    ${GetOptions} $cmdLineParams '/?' $R0
+    IfErrors +2 0
+    Call ShowHelpMessage
+
+  	${GetOptions} $cmdLineParams '/H' $R0
```
lightningterror, line 90: Forgot a tab here.

nsis/SharedDefs.nsh
```diff
@@ -62,4 +64,57 @@ ${If} $UserPrivileges == "Admin"
...
+    Call ShowHelpMessage
+
+    Pop $R0
+
+
+    Var /GLOBAL option_startMenu
+    Var /GLOBAL option_desktop
+    Var /GLOBAL option_portable
+    StrCpy $option_startMenu     1
+  	StrCpy $option_desktop       1
```
lightningterror, line 101: Ditto.

nsis/SharedDefs.nsh
```diff
@@ -62,4 +64,57 @@ ${If} $UserPrivileges == "Admin"
...
+    StrCpy $option_portable      0
+
+    Push $R0
+
+  	${GetOptions} $cmdLineParams '/NoStart' $R0
+    IfErrors +2 0
+  	StrCpy $option_startMenu 0
```
lightningterror, line 108: Here as well, and above first line.

TheCakeIsNaOH: Done

nsis/SharedDefs.nsh
TheCakeIsNaOH: Done.

nsis/SharedDefs.nsh
TheCakeIsNaOH: Done.

MrCK1:
Thanks, I just needed the additional argument for Powershell - everything is working fine. 

Also I just had another thought as I was testing this. If you are interested, you could add an argument that hooks into the functionality at the end of the installer - which would allow PCSX2 to launch after the silent install is complete (just as the GUI already does)

TheCakeIsNaOH:
Hmm, that sounds fine, I can add that. It did not register to me as something to add, because presumably most people using the command line options would be using installing as an administrative function, in which case they would not want to launch it. 

The context for why I noticed that the silent switch was not working is because I was trying to update the Chocolatey package for PCSX2. I then decided I might as well fix up the rest of the command line arguments.
https://chocolatey.org/packages/pcsx2

MrCK1:
@TheCakeIsNaOH do you still want to make the changes to your PR, or would you rather merge it currently as-is?

TheCakeIsNaOH:
@MrCK1 At this point, let's just merge it as is. I'll plan to make a follow up, I just got busy, so it might not be for a little while.

MrCK1:
Sounds good, thanks for replying quickly.

MrCK1 (approved):

Issue #3428, opened 2020-05-29
[Onimusha 3] [NTSC] Blur effect when picking up item is broken


PolygonSlayer:
PCSX2 version:   v1.7.0-dev-18-g5a5dc7f67

***PCSX2 options:***
Issue occurs when using Hardware of any kind. I tried messing with all the settings, but nothing seemed to fix the issue. The issue does not occur in Software however where it looks like it does on a real PS2.

***Description of the issue:***
When picking up any items in the game the screen should blur and a small window showing the item appears. In Software this effect is as it should, but when playing in Hardware mode the screen instead goes slightly white, but remains sharp.
The issue is minor and does not hinder gameplay at all, but it's not accurate so thought I'd report it :)
Screenshots from both hardware and software mode attached.
![Onimusha3_hardware](https://user-images.githubusercontent.com/44800819/83308503-cb222880-a1fe-11ea-8def-61a4ab8658c5.jpg)
![Onimusha3_software](https://user-images.githubusercontent.com/44800819/83308506-cbbabf00-a1fe-11ea-9edd-1a5463e9fc3b.jpg)

***How to reproduce the issue:***
You can see the issue as soon as you pick up your first item (only a few minutes into the game). Switching between hardware and software will highlight the issue.

lightningterror:
Same like the other issue, gsdump will be nice.

PolygonSlayer:
Attached is the gsdump when picking up an item:

[Onimusha3_NTSC_missingBlurEffect.zip](https://github.com/PCSX2/pcsx2/files/4704678/Onimusha3_NTSC_missingBlurEffect.zip)

lightningterror:
Looks like 1.4 is fine so it might be a regression. Need to track it down.

lightningterror:
Looks like regression commit is https://github.com/PCSX2/pcsx2/commit/52a2bb55241ffdf3b580f567400fe047c96b56fc

PolygonSlayer:
I noticed this got merged with the other green filter issue. Not sure if it's any help, but I'll attach the dumps I had during my recent run for it just in case :)

[Onimusha3_NTSC_greenFilter.zip](https://github.com/PCSX2/pcsx2/files/4740745/Onimusha3_NTSC_greenFilter.zip)
[Onimusha3_NTSC_endBoss_Green.zip](https://github.com/PCSX2/pcsx2/files/4740746/Onimusha3_NTSC_endBoss_Green.zip)
[...]

PR #3436, opened 2020-06-03
GSdx-hw: Onimusha 3 crc hack and texture shuffling

### LINKED ISSUE #3429: [Onimusha 3] [NTSC] Certain shadow effects don't appear in hardware mode
**PCSX2 version:** v1.7.0-dev-18-g5a5dc7f67

PCSX2 options:
Shadow issues occur in hardware mode regardless of any options I tried. In software mode the shadows always appear correctly.

The game seem to be using at least a couple of ways to create shadows. One method where it's more like "blobs" below the character seem to work fine and this is the shadows that most enemies will use as well as certain scene with the playable characters. However, in some places more advanced shadows (I presume shadow maps) are used and in hardware mode these do not seem to appear at all.
Screenshots attached show one example of the "blob shadow" (a bit hard to see, but it's there and easier to see in motion) and one example where the shadow maps are used. Screenshot examples show both hardware and software:

Blob shadows (these work fine):

Shadow maps (are broken in hardware):

The shadows can easily be observed early on in the game. I've highlighted that the game uses two types of shadows in case one see the blob type shadow an assume they all work fine.

lightningterror:
gsdx-hw: Remove GSC_Onimusha3 crc hack. 
It will allow to render shadows in certain locations in Ominusha 3.
Issue #3429

gsdx-hw: Add a temporary fix for Onimusha 3.
Allow texture shuffle to run. It will allow to render
shadows and other effects properly.
Would be nice to have a better solution with no crc in the future.

Close #3429 #3428

### DIFF
--- a/plugins/GSdx/Renderers/HW/GSHwHack.cpp
+++ b/plugins/GSdx/Renderers/HW/GSHwHack.cpp
@@ -310,16 +310,6 @@ bool GSC_IkkiTousen(const GSFrameInfo& fi, int& skip)
 	return true;
 }
 
-bool GSC_Onimusha3(const GSFrameInfo& fi, int& skip)
-{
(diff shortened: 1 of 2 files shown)


lightningterror:
Might as well look in to dmc3 as well since it looks like a similar issue.

PR #3440, opened 2020-06-05
GSdx-d3d11: Remove Alpha Stencil hack.

lightningterror:
Hack is obsolete/replaced with accurate date/date_one option/code.

There's pretty much no usecase for it remaining other than using it as
an alternative to skipdraw which removes/breaks transparency effects.


incognitoh:
this is silly. there are still plenty of cases when this is useful.  there is no reason to keep deleting features that are still needed.  at least give us an option in the .ini to enable it at the very least.

kenshen112:
@incognitoh Oh? And if there's plenty of cases. Surely you can list some of them?

PR #3442, opened 2020-06-06
gsdx-ogl: Adjust some gl logs/cleanup errors

lightningterror:


RedDevilus:
Don't disable sparse texture by* instead of ny?

PR #3445, opened 2020-06-06
Update Background Logo to Aquatic Theme

RedDevilus:
This will give it a watery filter on top of its current logo instead of the old one

PR #3450, opened 2020-06-07
Update Background Logo to Aquatic Theme

RedDevilus:
Improves background logo that matches the colors.

Also #3445 (Accidentaly pushed old files)


RedDevilus:
Original logo
Original
First version which applies the Water effect on whole logo
Commit1
Second Version which does the same but doesn't apply it on the text
Commit2

RedDevilus:
Any suggestions/opinions about this? If needed i can still reduce the water effect. Input is handy for everyone and i can't read minds.

lightningterror:
I prefer the original, water effect feels odd with a circuit next to it too.

asp0909:
I prefer the original

turtleli:
I prefer the original.

lightningterror:
Closing as original is preferred.

PR #3402, opened 2020-05-19
GameDB: Fix Multiple Games

RedDevilus:
#3391 Sorry for the mess. I bluescreened and not sure whats happened. Should be fixed now

PR #3452, opened 2020-06-08
GameDB: List of patches for 21 games (or 15 if u don't count all vers…

### LINKED ISSUE #1001: True Crime:Streets of LA - The source of bugs
I found this report here:
https://github.com/jpd002/Play-/issues/343

> So I thought I'd give this game a go (SLUS-20550) just to see how it fares on Play! vs. PCSX2 since PCSX2 still cannot reliably run this game past the title screen, and even with a patch the game still crashes constantly and is hardly playable, but I've found a few culprits from debugging:
> Tested on Play!, the game does boot but will crash after playing this video file (MOVIESVCNTSC.PSS) and it could be because it hits a VU1 opcode that maybe isn't emulated correctly or at all (you be the judge: it's '3ff8') that may or may not work correctly; or it's a stream command sent to the GS in which something goes terribly wrong (PCSX2 crashes here too unless you use PCSX2's "SuperVU0/1"), so it's safe to say that the issue is probably actually caused by either a GS transfer and/or VU1 itself.
> Second, this game writes to the TLB and I can imagine that would cause many issues since neither Play! nor PCSX2 seem to reliably handle TLB writes/reads in a safe way for all games to run right.
> In PCSX2, the patch likely helps cheat past the TLB and other stuff, but when playing in-game it eventually crashes since this hackery doesn't seem to hold its own for a long time.
> Thought I'd put this here since I dug a little and found out what this game actually tries to do.
> I know that it crashes right before the title screen is about to be drawn, right after the video is finished

### LINKED ISSUE #3049: Socom 3: U.S. Navy SEALs (NTSC-U) - Slowdown when looking at the sun.
***PCSX2 version:*** 1.5.0-dev-3212

***PCSX2 options:*** Tried several combinations of round and clamping, and also several gamefixes. No effect.

***Plugins used:*** Default. Happens in both HW and SW mode.

***Description of the issue:*** When looking around, the game works fine and no message appear on the log. But when looking at the sun, the game becomes sluggish and the following message keeps popping up in the log:

sceGsSyncPath: VIF1 does not terminate
	<D1_CHCR=10000045:D1_TADR=00121e50:D1_MADR=00121e50:D1_QWC=00000000>
	<D2_CHCR=70000045:D2_TADR=00b3daf0:D2_MADR=00b3db10:D2_QWC=00000000>
	<VIF1_STAT=00000043:GIF_STAT=00000004>

The VU and GS drops to 0%, but the EE does not change, maintaining the average of 37%. Also, the speed does not change too (100% - 60FPS). My CPU usage drops from 40% to 10%.

***How to reproduce the issue:*** Right at the beginning of the game. Skip all cutscenes and, when you gain control of the character, look at the sun.

***Last known version to work:*** --

***PC specifications:*** CPU: Intel Core I5 4690 - GPU: EVGA GTX 750Ti - OS: Windows 7 64-bit

RedDevilus:
…ions as a single entry);

- 24 : The game
- Amplitude
- Freaky Flyers
- Free Running
- Ghosthunter
- Next Generation Tennis 2003
- NHL 2K6
- NHL 2K9
- Primal
- Salt Lake 2002
- Savage Skies
- SOCOM 3
- Tomb Raider: Angel of Darkness
- True Crime: Streets of LA
- Yanya Caballista: City Skater

Fixes #1001 (fixed on 1.6 for NTSC-U already, this PR ports that to the PAL)
Fixes #3049 (fixes Socom 3)

### DIFF
--- a/plugins/GSdx/Renderers/HW/GSRendererHW.cpp
+++ b/plugins/GSdx/Renderers/HW/GSRendererHW.cpp
@@ -1294,20 +1294,27 @@ void GSRendererHW::Draw()
 		m_texture_shuffle = (GSLocalMemory::m_psm[context->FRAME.PSM].bpp == 16) && (tex_psm.bpp == 16)
 			&& draw_sprite_tex && m_src->m_32_bits_fmt;
 
-		// Shadow_of_memories_Shadow_Flickering (Okami mustn't call this code)
+		// Okami mustn't call this code
 		if (m_texture_shuffle && m_vertex.next < 3 && PRIM->FST && (m_context->FRAME.FBMSK == 0)) {
 			// Avious dubious call to m_texture_shuffle on 16 bits games
 			// The pattern is severals column of 8 pixels. A single sprite
 			// smell fishy but a big sprite is wrong.
 
-			// Tomb Raider Angel of Darkness relies on this behavior to produce a fog effect.
-			// In this case, the address of the framebuffer and texture are the same. 
-			// The game will take RG => BA and then the BA => RG of next pixels. 
-			// However, only RG => BA needs to be emulated because RG isn't used.
-			GL_INS("WARNING: Possible misdetection of a texture shuffle effect");
-
+			// Shadow of Memories/Destiny shouldn't call this code.
+			// Causes shadow flickering.
 			GSVertex* v = &m_vertex.buff[0];
-			m_texture_shuffle = ((v[1].U - v[0].U) < 256) || m_context->FRAME.Block() == m_context->TEX0.TBP0;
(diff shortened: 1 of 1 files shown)


RedDevilus:
I have added patches from Prafull and PSI. True Crime LA (PAL) is ported from the NTSC-U version etc. Everything is in the comments and i only added official patches and the ones that have been tested extensively. (Other patches will need to be reviewed and tested).

RedDevilus:
Fixes #909 (Seems to be already fixed with previous GameDB #3402)
Fixes #1001 (fixed on 1.6 for NTSC-U already, this PR ports that to the PAL)
Fixes #3049 (fixes Socom 3)

PR #3454, opened 2020-06-11
Rewrite the spu2-x dialog on Linux in wxwidgets.

arcum42:
It looks a fair amount different than the gtk version, but the gtk version didn't quite look like the windows version anyways. 

This actually adds in a missing option that just wasn't there in the gtk version, and seems generally better than the original to me. I put it in a separate folder from the Linux folder in case we decide to make windows use the wx version at some point...


arcum42:
No idea if this breaks Mac support.
@tellowkrinkle

tellowkrinkle:
Minor changes for macOS, mostly removing the workaround I had put in place for GTK crashes
[0001-MacOS-Fixes.patch.gz](https://github.com/PCSX2/pcsx2/files/4763896/0001-MacOS-Fixes.patch.gz)

Though I feel like the right side looks a little squished, the sliders are super short and there's not enough space for the full text of the Audio Expansion box's Stereo option text
image

Same for the debug settings, the last column gets cut off
image

Also the SDL API chooser box goes below the now invisible PortAudio API chooser box 
image

Note: Some of this may be due to bugs in wx's Mac implementation, in which case I can go complain over there (I haven't seen how the UI looks on Linux)

Also `Interpolatio` → `Interpolation`

arcum42:
With the sizers and options, a lot of times I was fiddling with them until things worked, as I honestly find sizers fairly frustrating to deal with in wx, and I could easily have set some things that just happen to work in Linux. I'll do another pass and see if I can make the flags a little more sensible when I get a chance. (Hopefully this weekend.)

Here's what it looks like in Linux.

screenshot

Weirdly, there actually already is an n on Interpolation. With Portaudio and SDL, I actually had a fair amount of trouble with that. I wanted to hide the sizers that the ones not relevant were in, but that didn't hide a thing, so I hid the text and pulldown instead, which looked perfect in Linux, but apparently not so much in Mac OS X.

Thanks for the patch. I'll go ahead and apply that as well.

arcum42:
@tellowkrinkle When you get a chance, can you see what the latest versions look like on the Mac?

I redid the hiding and showing code, and the windows sizing is now hopefully saner, as I managed to get it to actually pick sane defaults instead of having to manually pick out a size for it...

tellowkrinkle:
Well the sizing works but now there's no titlebar which makes it impossible to close the windows

image

image

It looks like you need to add `wxCAPTION | wxCLOSE_BOX` to the spots where you currently have `wxRESIZE_BORDER`

Also `wxRESIZE_BORDER` isn't very useful, as resizing the window reveals a whole lot of nothing, so maybe we should just use `wxCaption | wxCLOSE_BOX`
image

tellowkrinkle:
Also you apparently need to escape `&` as `&&` in the `Core & Voice Stats`

tellowkrinkle:
The `Interpolatio` issue [was a wxMac bug](https://github.com/wxWidgets/wxWidgets/pull/1905), ignore it for now

arcum42:
@tellowkrinkle Alright, made those changes. The resizable border was mostly from when I was worried that it'd still have size problems. Oddly, on Linux, it still has a close box without the flag.

Nice to know the missing n is a Mac bug. If there's a workaround, like adding spaces to the text, I can always do that...

tellowkrinkle:
Adding a space after it does fix it, if you want to do that
I just figured that it's a minor enough issue that we might as well just wait for it to get fixed and not try to work around it

orbea:
@arcum42 This broke the builds without portaudio.

PR #3457, opened 2020-06-13
PCSX2-GUI: Revision of tooltips/descriptions and other fixes for PCSX2 (Lilypad + others)

RedDevilus:
Changes to Large Framebuffer, 8-bit textures, Xinput, Dinput, huge chunk of memory cards section, Framelimiter, Zoom.
Recording and Recording to Input Recording and Screenrecorder.
Miscellaneous improvements to DPI scaling or small revisions (adding or rehauling strings).
And more...

GUI clean-up and change the tooltips and descriptions to be more easy to understand to endusers.
Todo:
- [x] Tooltip/description changes
- [x] Error-checking 100% done


RedDevilus:
The new default
image

RedDevilus:
So how do people feel about this? I know this will help certain controllers like the DS4 and others, but on the other side of this discussion some controllers only have Dinput capabilities.

lightningterror:
Not really in favor since there are plenty of Dinput controllers as well.

refractionpcsx2:
Nor am I. There's no harm in DInput being enabled.

RedDevilus:
That's fair, perhaps the detection system could be different like when dinput and xinput are available it autopicks, or revise tooltips.
Most times its DS4 which have problems but there are others, wouldn't be improvement to a pretty huge userbase of dinput users.

refractionpcsx2:
Some sort of auto detection would be nice if you can do that.  You can update this one if you like, just make sure you edit the title/description

lightningterror:
You can edit this one.

RedDevilus:
I'll try looking into it, hopefully i am able to do that. Perhaps extrapolate extra types of controllers who have the same problem and assign them the right type. I think it was Switch controller and some others if i remember correctly.

RedDevilus:
Seems autodetecting DS4 and changing the case of Dinput to off is out of my scope of expertise. I'll atleast change descriptions/tooltips on the GUI (like the plugins).

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -94,7 +94,8 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_PALTEX:
 			return "When checked 4/8 bits texture will be send to the GPU with a palette. GPU will be in charge of the conversion.\n\n"
 				"When unchecked the CPU will convert directly the texture to 32 bits.\n\n"
-				"It is basically a trade-off between GPU/CPU.";
+				"It is basically a trade-off between GPU/CPU.\n"
+				"For example you have a CPU that meets minimum of STR 1600 but not 2000 and a strong GPU GTX 1060 or stronger.";
```
lightningterror, line 98: Old one was fine.

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -198,8 +199,9 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_LARGE_FB:
 			return "Allocate a large framebuffer to be compliant with GS memory (Prevents FMV flickering).\n"
 				"It increases GPU/memory requirements.\n\n"
-				"Note: It should be disabled for Armored Core, Destroy All Humans, and Gran Turismo.\n"
-				"It will amplify RAM/VRAM spikes. Other games might be affected as well.";
+				"Note: It should be disabled for Armored Core, Destroy All Humans, Gran Turismo and many others .\n"
+				"It will amplify RAM/VRAM spikes with a chance of not improving anything for you.\n\n"
+				"RECOMMENDED to leave this disabled in most cases as it only helps select titles.";
```
lightningterror, line 204: Likewise.

plugins/LilyPad/Tooltips.cpp
```diff
@@ -63,8 +63,13 @@ LPWSTR dialog_message(int ID, bool *updateText)
...
                    L"Right-click and select \"Refresh\" to update the list of devices in case a recently connected device has not shown up yet.";
-        case IDC_G_XI:
-            return L"Xbox 360/Xbox One controllers (or devices/software imitating such) only";
+        case IDC_G_DI:
+			return L"(Legacy) Enable this if your gamepad doesn't support Xinput.\n\n"
+				   L"This will check your inputs (button presses) and maps them to corresponding action";
+		case IDC_G_XI:
+            return L"Imitates a Xbox (360) controller for a wide range of controllers.\n\n"
+				   L"The computer doesn't notice even if it's a totally different controller.\n\n"
+				   L"Steam also uses this method (but Steam also convert non-compatible controllers";
```
lightningterror, line 72: Maybe
```
        case IDC_G_XI:
            return L"Xbox 360/Xbox One controllers (or devices/software imitating such) only.\n\n";
                       L"Running through Steam might be required for the controllers to work properly.";
```

plugins/GSdx/Window/GSSetting.cpp
RedDevilus: I think this one needs revision. I would expect that to a normal end-user will read this option is in 'most cases fine especially on a good rig'.

plugins/LilyPad/Default.ini
```diff
@@ -5,7 +5,7 @@ Force Cursor Hide=0
 Mouse Unfocus=1
 Background=1
 Multiple Bindings=0
-DirectInput Game Devices=1
+DirectInput Game Devices=0
```
lightningterror, line 8: Should be reverted.

plugins/GSdx/Window/GSSetting.cpp
RedDevilus: Yeah perhaps shouldve been worded better, i think there is potential to improve it.

plugins/LilyPad/Tooltips.cpp
```diff
@@ -63,8 +63,13 @@ LPWSTR dialog_message(int ID, bool *updateText)
                    L"Double-click a device in the list or right-click it and select \"Test Device\" to display a continuously updated list of the state of all inputs on the selected device.\n"
                    L"Use this option to check if all the inputs on a controller function properly.\n\n"
                    L"Right-click and select \"Refresh\" to update the list of devices in case a recently connected device has not shown up yet.";
-        case IDC_G_XI:
-            return L"Xbox 360/Xbox One controllers (or devices/software imitating such) only";
+        case IDC_G_DI:
+			return L"(Legacy) Enable this if your gamepad doesn't support Xinput.\n\n"
+				   L"This will check your inputs (button presses) and maps them to corresponding action";
```
lightningterror, line 68: Maybe keep only the first sentence.

plugins/LilyPad/Tooltips.cpp
RedDevilus: While that is kinda true, the first sentence feels off. Perhaps removing the word 'only' because it feels redundant?
The brackets help readability however the meaning is a bit lost to me. (or devices/software imitating such as other controllers)

plugins/LilyPad/Tooltips.cpp
```diff
@@ -63,8 +63,13 @@ LPWSTR dialog_message(int ID, bool *updateText)
                    L"Double-click a device in the list or right-click it and select \"Test Device\" to display a continuously updated list of the state of all inputs on the selected device.\n"
                    L"Use this option to check if all the inputs on a controller function properly.\n\n"
                    L"Right-click and select \"Refresh\" to update the list of devices in case a recently connected device has not shown up yet.";
-        case IDC_G_XI:
-            return L"Xbox 360/Xbox One controllers (or devices/software imitating such) only";
+        case IDC_G_DI:
```
RedDevilus, line 66: I guess i can't use this method to create tooltips which don't exist yet.

RedDevilus:
I've rebased and reduced some commits, i'll take some time to do changes that helps people instead of making it worse or harder to understand.

RedDevilus:
Memory cards:
image

LilyPad:
image

Emulation settings:
image
image

I basically reworded sentences/ removed experimental mentions for folders (i have been using them for years now without trouble). I still have to do the others and perhaps revise them.

RedDevilus:
This will change translatable strings while being more accessible and user-friendly for reading.

pcsx2/gui/Panels/VideoPanel.cpp
```diff
@@ -30,7 +30,7 @@ Panels::FramelimiterPanel::FramelimiterPanel( wxWindow* parent )
 	: BaseApplicableConfigPanel_SpecificConfig( parent )
 {
 	m_check_LimiterDisable = new pxCheckBox( this, _("Disable Framelimiting"),
-		_("Useful for running benchmarks. Toggle this option in-game by pressing F4.") );
+		_("Useful for running benchmarks. Toggle this option in-game by pressing F4. It basically uncaps FPS to unlimited") );
```
lightningterror, line 33: Isn't the option name itself self explanatory?

plugins/LilyPad/LilyPad.rc
```diff
@@ -450,21 +450,21 @@ BEGIN
...
+    CONTROL         "XInput (For modern controllers)",IDC_G_XI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,94,125,10
     CONTROL         "DualShock 3 native mode (Requires libusb)",IDC_G_DS3,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,106,155,10
     CONTROL         "Monitor when in background",IDC_BACKGROUND,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,118,107,10
     GROUPBOX        "Mouse API",IDC_STATIC,216,16,192,85
     CONTROL         "Windows messaging (Recommended)",IDC_M_WM,"Button",BS_AUTORADIOBUTTON | WS_GROUP,223,27,137,10
     CONTROL         "Raw input",IDC_M_RAW,"Button",BS_AUTORADIOBUTTON,223,39,112,10
     CONTROL         "DirectInput",IDC_M_DI,"Button",BS_AUTORADIOBUTTON,223,51,112,10
     CONTROL         "Disable",IDC_M_DISABLE,"Button",BS_AUTORADIOBUTTON,223,63,39,10
-    CONTROL         "Start without mouse focus",IDC_MOUSE_UNFOCUS,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,223,75,98,10
+    CONTROL         "Start with unfocused mouse",IDC_MOUSE_UNFOCUS,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,223,75,105,10
```
lightningterror, line 462: Older was better.

pcsx2/gui/Panels/VideoPanel.cpp
RedDevilus: Perhaps, i just felt it was even more obvious what it does. It feels like a double negative, maybe a bit too nitpicky? I don't see the harm for expanding this part.

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -196,10 +196,10 @@ const char* dialog_message(int ID, bool* updateText) {
...
 				"\n\nThis hack replaces the partial invalidation with a complete deletion of the texture to reduce the CPU load.\n\nIt helps snowblind engine games.";
 		case IDC_LARGE_FB:
-			return "Allocate a large framebuffer to be compliant with GS memory (Prevents FMV flickering).\n"
-				"It increases GPU/memory requirements.\n\n"
-				"Note: It should be disabled for Armored Core, Destroy All Humans, and Gran Turismo.\n"
-				"It will amplify RAM/VRAM spikes. Other games might be affected as well.";
+			return "Reserves a bigger sized framebuffer for textures to prevent FMV flickers.\n"
+				"Increases GPU/memory requirements but also easier to hit the limit with lower FPS.\n"
+				"Enabling this can amplify stuttering due to low RAM/VRAM.\n\n"
+				"Note: It should be disabled (especially for Armored Core, Destroy All Humans, Gran Turismo and many others).";
```
lightningterror, line 202: Likewise here, this just makes it seem that user should never turn this option on which isn't the case.
Option is a lot more useful.

pcsx2/gui/Panels/VideoPanel.cpp
arcum42: I'd probably change that last sentence to just "Uncaps FPS.". Basically is one of these words that can almost always be deleted without consequence, and I think unlimited is already implied in "uncaps".

Seems to me like it should be at the beginning, not the end, too. In fact, maybe:
"Uncaps FPS. Toggle in-game by pressing F4. (Useful for running benchmarks.)" ?

plugins/GSdx/Window/GSSetting.cpp
RedDevilus: I'll just drop 'many' and make the language not as forceful for not enabling it. In the previous tooltip it was like you should always enable it, i even did that for years before contributing.

pcsx2/gui/Panels/VideoPanel.cpp
RedDevilus: Ah yes fair point, i was going to change that and forgot about it.

plugins/LilyPad/LilyPad.rc
RedDevilus: Fair, i was playing around the idea to make it more readable but i'd prefer the variable to be IDC_MOUSE_FOCUS and say Start with mouse focus and the option by default disabled but that's just nitpicking from my side.

RedDevilus:
This PR should be done (checked and tested). I balanced that the 'jargon' is more userfriendly and it's readability.

pcsx2/gui/Panels/VideoPanel.cpp
```diff
@@ -43,7 +43,7 @@ Panels::FramelimiterPanel::FramelimiterPanel( wxWindow* parent )
 	m_text_BasePal		= CreateNumericalTextCtrl( this, 7 );
 
 	m_spin_NominalPct	->SetRange( 10,  1000 );
-	m_spin_SlomoPct		->SetRange(  1,  1000 );
+	m_spin_SlomoPct		->SetRange(  5,  1000 );
```
lightningterror, line 46: Why was the range changed?

RedDevilus: I gave already an example but to reiterate, it doesn't allow lower than 5 as its input.
1.6 stable + 1.7 dev build before apply:
image
After apply:
image

RedDevilus: So if they changed the range when someone puts for example 1 it will automatically change to the preferred 5 already.
Or if they use the arrow keys it will only go to 5 as lowest number. Not sure if it's possible to force it do frame per frame instead of a slow motion (besides the point for this PR)

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -196,10 +196,11 @@ const char* dialog_message(int ID, bool* updateText) {
 			return "By default, the texture cache handles partial invalidations. Unfortunately it is very costly to compute CPU wise."
 				"\n\nThis hack replaces the partial invalidation with a complete deletion of the texture to reduce the CPU load.\n\nIt helps snowblind engine games.";
 		case IDC_LARGE_FB:
-			return "Allocate a large framebuffer to be compliant with GS memory (Prevents FMV flickering).\n"
-				"It increases GPU/memory requirements.\n\n"
-				"Note: It should be disabled for Armored Core, Destroy All Humans, and Gran Turismo.\n"
-				"It will amplify RAM/VRAM spikes. Other games might be affected as well.";
+			return "Reserves a bigger sized framebuffer for textures to prevent FMV flickers.\n"
```
refractionpcsx2, line 199: Not very good English, would be better to say "Reserves a larger framebuffer for textures to prevent FMV flickering.\n"

plugins/LilyPad/Tooltips.cpp
```diff
@@ -35,8 +35,8 @@ LPWSTR dialog_message(int ID, bool *updateText)
                    L"Note 2: This does not enable the mouse to function as an in-game mouse in PS2 games that support a USB mouse or lightgun."
                    L"This requires a USB plugin, while LilyPad is a PAD plugin(PAD means it emulates devices that are to be plugged into the PlayStation controller port that connects the DualShock 2 controller).";
         case IDC_MOUSE_UNFOCUS:
-            return L"Starts the emulation without the mouse being in focus by the application.\n\n"
-                   L"Without mouse focus, the mouse can still be used outside of the emulation.\nWith the mouse in focus it can only be used as an input for the emulator.";
+            return L"Enabled: Mouse is unfocused and can be used with emulation and outside it for general usage.\n\n"
```
refractionpcsx2, line 38: Possibly a better way to put this? like "Enabled: Mouse is unfocused and can be used in emulation and generally inside your OS.\n\n"

pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -81,7 +81,7 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 	m_check_HideMouse->SetToolTip( pxEt( L"Check this to force the mouse cursor invisible inside the GS window; useful if using the mouse as a primary control device for gaming.  By default the mouse auto-hides after 2 seconds of inactivity."
 	) );
 
-	m_check_Fullscreen->SetToolTip( pxEt( L"Enables automatic mode switch to fullscreen when starting or resuming emulation. You can still toggle fullscreen display at any time using alt-enter."
+	m_check_Fullscreen->SetToolTip( pxEt( L"Enables automatic mode switch to fullscreen when starting or resuming emulation. You can still toggle fullscreen display at any time using \nAlt-Enter or double clicking the emulion window (game)."
```
refractionpcsx2, line 84: emulation*  and I would call it the output window, or something.

pcsx2/gui/Panels/MemoryCardListView.cpp
```diff
@@ -98,9 +98,9 @@ const ListViewColumnInfo& MemoryCardListView_Simple::GetDefaultColumnInfo( uint
 	{
 		{ _("PS2 Port")		, 170 , wxLIST_FORMAT_LEFT	},
 		//{ _("Port status")	, 80  , wxLIST_FORMAT_LEFT	},
-		{ _("Card (file) name")	, 145 , wxLIST_FORMAT_LEFT	},
+		{ _("Active Memory Card")	, 145 , wxLIST_FORMAT_LEFT	},
```
refractionpcsx2, line 101: Is this not the card names in the list? they aren't *all* active, surely?

RedDevilus: Actually they are (slot1 and slot2). Under that are the unused cards. So it fits.

refractionpcsx2: yes but the column is saying what is underneath it, including the inactive/unmounted  cards, it's not what's active.

RedDevilus: Haven't thought about that way. Unless the line is replaced with a column to separate them so i can keep this. If its not feasible i'll revert the changes.

refractionpcsx2: I don't know if that is feasible, the boxes would have to be split in order to do that, but then a heading for the table would make more sense in that scenario than changing the column name

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -92,9 +92,9 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_OSD_MONITOR:
 			return "Continuously prints the FPS and the EE, GS and VU(if the MTVU speedhack is enabled) percentages onscreen.";
 		case IDC_PALTEX:
-			return "When checked 4/8 bits texture will be send to the GPU with a palette. GPU will be in charge of the conversion.\n\n"
-				"When unchecked the CPU will convert directly the texture to 32 bits.\n\n"
-				"It is basically a trade-off between GPU/CPU.";
+			return "Enabled: GPU will handle 4/8 bits textures while CPU does the rest.\n\n"
+				   "Disabled: CPU will have full control and directly convert the textures to 32 bits.\n\n"
+				   "It is basically a trade-off between GPU/CPU vs CPU alone.";
```
lightningterror, line 97: Formatting, extra spaces.

plugins/LilyPad/Tooltips.cpp
```diff
@@ -63,8 +63,11 @@ LPWSTR dialog_message(int ID, bool *updateText)
                    L"Double-click a device in the list or right-click it and select \"Test Device\" to display a continuously updated list of the state of all inputs on the selected device.\n"
                    L"Use this option to check if all the inputs on a controller function properly.\n\n"
                    L"Right-click and select \"Refresh\" to update the list of devices in case a recently connected device has not shown up yet.";
-        case IDC_G_XI:
-            return L"Xbox 360/Xbox One controllers (or devices/software imitating such) only";
+        case IDC_G_DI:
+			return L"(Legacy) Enable this if your gamepad doesn't support Xinput.";
+		case IDC_G_XI:
+            return L"For Xbox 360/ Xbox One controllers (or devices supporting Xinput).\n\n"
+				   L"Running through Steam as a non-Steam game might be required for the controllers to work properly.";
```
lightningterror, line 70: Formatting, should use spaces.

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -196,10 +196,11 @@ const char* dialog_message(int ID, bool* updateText) {
...
 		case IDC_LARGE_FB:
-			return "Allocate a large framebuffer to be compliant with GS memory (Prevents FMV flickering).\n"
-				"It increases GPU/memory requirements.\n\n"
-				"Note: It should be disabled for Armored Core, Destroy All Humans, and Gran Turismo.\n"
-				"It will amplify RAM/VRAM spikes. Other games might be affected as well.";
+			return "Reserves a bigger sized framebuffer for textures to prevent FMV flickers.\n"
+				"Increases GPU/memory requirements.\n"
+				"Enabling this can amplify stuttering due to low RAM/VRAM.\n\n"
+				"Note: It should be disabled (for Armored Core, Destroy All Humans, Gran Turismo and others).\n"
+				"This option does not improve the graphics or the FPS.";
```
lightningterror, line 203: Last line seems a bit useless.

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -196,10 +196,11 @@ const char* dialog_message(int ID, bool* updateText) {
...
 				"\n\nThis hack replaces the partial invalidation with a complete deletion of the texture to reduce the CPU load.\n\nIt helps snowblind engine games.";
 		case IDC_LARGE_FB:
-			return "Allocate a large framebuffer to be compliant with GS memory (Prevents FMV flickering).\n"
-				"It increases GPU/memory requirements.\n\n"
-				"Note: It should be disabled for Armored Core, Destroy All Humans, and Gran Turismo.\n"
-				"It will amplify RAM/VRAM spikes. Other games might be affected as well.";
+			return "Reserves a bigger sized framebuffer for textures to prevent FMV flickers.\n"
+				"Increases GPU/memory requirements.\n"
+				"Enabling this can amplify stuttering due to low RAM/VRAM.\n\n"
+				"Note: It should be disabled (for Armored Core, Destroy All Humans, Gran Turismo and others).\n"
```
lightningterror, line 202: Maybe drop the brackets, at the end you can adjust it to `and possibly others.`

plugins/GSdx/Window/GSSetting.cpp
tadanokojin: I can understand why he put it there though :P

RedDevilus: It's the balance i made where it implied it's a bad option, i made this extra line to make it obvious to people (even my former self) that it should only be used for FMV flickering.

pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -70,9 +70,9 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
 	m_check_Fullscreen	= new pxCheckBox( this, _("Default to fullscreen mode on open") );
 	m_check_DclickFullscreen = new pxCheckBox( this, _("Double-click toggles fullscreen mode") );
 
-	m_combo_FMVAspectRatioSwitch->SetToolTip( pxEt( L"Off: Disables temporary aspect ratio switch.\n\n4:3: Temporarily switch to a 4:3 aspect ratio while an FMV plays to correctly display an 4:3 FMV.\n\n16:9: Temporarily switch to a 16:9 aspect ratio while an FMV plays to correctly display a widescreen 16:9 FMV.") );
+	m_combo_FMVAspectRatioSwitch->SetToolTip( pxEt( L"Off: Disables temporary aspect ratio switch. (Will use the above setting from Aspect Ratio.)\n\n4:3: Temporarily switch to a 4:3 aspect ratio while an FMV plays to correctly display an 4:3 FMV.\n\n16:9: Temporarily switch to a 16:9 aspect ratio while an FMV plays to correctly display a widescreen 16:9 FMV.") );
```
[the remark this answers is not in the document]
RedDevilus: image

In the code if FMV override is on off then Aspect Ratio setting will take over.
Sounds better if i add 'Instead' to make it more obvious.

pcsx2/gui/Dialogs/CreateMemoryCardDialog.cpp
```diff
@@ -252,8 +252,8 @@ void Dialogs::CreateMemoryCardDialog::CreateControls()
 		.	SetToolTip(_t("Use at your own risk.  Erratic memory card behavior is possible (though unlikely)."))
 		.	SetInt(64),
 
-		RadioPanelItem(_("Folder [experimental]"), _("Store memory card contents in the host filesystem instead of a file."))
-		.	SetToolTip(_t("Automatically manages memory card contents so that the console only sees files related to the currently running software. Allows you to drag-and-drop files in and out of the memory card with your standard file explorer. This is still experimental, so use at your own risk!"))
+		RadioPanelItem(_("Folder"), _("Transforms the memory card into an usable folder with files inside"))
+		.	SetToolTip(_t("Gives every game 8MB for the saves (essentialy a whole memcard).You can see the structure and the saves with your File Explorer. Easier than using mymc(plus) but doesn't work with this folder type"))
```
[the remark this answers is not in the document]
RedDevilus: 'Incompatible with mymc(plus)' was my goal

pcsx2/gui/Panels/MemoryCardListView.cpp
RedDevilus: image

This is why i thought it was separated in the first place, still trying to figure out a good alternative.

RedDevilus: Changed Active Memory card to Memory card and Unused is changed to disabled. I tried to find my way around WX, PX and whatever is relevant to it.
Maybe unused as inactive if disabled isnt fitting?

image

RedDevilus:
Rebased and added this:

image
Not sure about this one
image

Reason:
image
They are 2 different things

Capture video will dump frames with audio and a separate audio one
Recording is actually Input Recording which also has a virtual pad. While it has TAS functionality it's not strictly one because you can't do macros or rewind stuff (if i got that right.

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -92,9 +92,9 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_OSD_MONITOR:
 			return "Continuously prints the FPS and the EE, GS and VU(if the MTVU speedhack is enabled) percentages onscreen.";
 		case IDC_PALTEX:
-			return "When checked 4/8 bits texture will be send to the GPU with a palette. GPU will be in charge of the conversion.\n\n"
-				"When unchecked the CPU will convert directly the texture to 32 bits.\n\n"
-				"It is basically a trade-off between GPU/CPU.";
+			return "Enabled: GPU will handle 4/8 bits textures while CPU does the rest.\n\n"
+				"Disabled: CPU will have full control and directly convert the textures to 32 bits.\n\n"
+			"It is basically a trade-off between GPU/CPU vs CPU alone.";
```
lightningterror, line 97: Tab.

plugins/LilyPad/Tooltips.cpp
```diff
@@ -63,8 +63,11 @@ LPWSTR dialog_message(int ID, bool *updateText)
                    L"Double-click a device in the list or right-click it and select \"Test Device\" to display a continuously updated list of the state of all inputs on the selected device.\n"
                    L"Use this option to check if all the inputs on a controller function properly.\n\n"
                    L"Right-click and select \"Refresh\" to update the list of devices in case a recently connected device has not shown up yet.";
-        case IDC_G_XI:
-            return L"Xbox 360/Xbox One controllers (or devices/software imitating such) only";
+        case IDC_G_DI:
+	        return L"(Legacy) Enable this if your gamepad doesn't support Xinput.";
+	    case IDC_G_XI:
+            return L"For Xbox 360/ Xbox One controllers (or devices supporting Xinput).\n\n"
```
lightningterror, line 69: Should use spaces, also spacing is wrong. Above line as well.

plugins/LilyPad/LilyPad.rc
```diff
@@ -450,8 +450,8 @@ BEGIN
     CONTROL         "Raw input",IDC_KB_RAW,"Button",BS_AUTORADIOBUTTON,22,40,112,10
     CONTROL         "DirectInput",IDC_KB_DI,"Button",BS_AUTORADIOBUTTON,22,52,112,10
     GROUPBOX        "Game Device APIs",IDC_STATIC,16,70,191,62
-    CONTROL         "DirectInput",IDC_G_DI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,82,65,10
-    CONTROL         "XInput",IDC_G_XI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,94,125,10
+    CONTROL         "DirectInput (For older controllers. Disable for DS4.)",IDC_G_DI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,82,183,10
+    CONTROL         "XInput (For modern controllers.)",IDC_G_XI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,94,125,10
```
lightningterror, line 454: The extra info should be in the new tooltips. You can do `DirectInput (Legacy)`.

RedDevilus: The tooltips for DirectInput and others aren't working, that's the reason i did it this way.

plugins/LilyPad/Tooltips.cpp
RedDevilus: I have used spaces though in the correct format:
image

lightningterror: ctrl + r + w to show spacing marks. Tabs will be marked with arrows, spaces with dots.

plugins/LilyPad/LilyPad.rc
lightningterror: You're missing a line for the tooltip.
https://github.com/PCSX2/pcsx2/blob/9aab6ae7fad4d34b5888404c6d4a5bc32496e307/plugins/LilyPad/Config.cpp#L2346

RedDevilus: Thanks, i'll make the changes.

RedDevilus: image
Works now, i'll revise the text.

pcsx2/gui/MainFrame.cpp
```diff
@@ -450,8 +450,8 @@ void MainEmuFrame::CreateWindowsMenu()
 void MainEmuFrame::CreateCaptureMenu()
 {
 	m_menuCapture.Append(MenuId_Capture_Video, _("Video"), &m_submenuVideoCapture);
-	m_submenuVideoCapture.Append(MenuId_Capture_Video_Record, _("Start Recording"));
-	m_submenuVideoCapture.Append(MenuId_Capture_Video_Stop, _("Stop Recording"))->Enable(false);
+	m_submenuVideoCapture.Append(MenuId_Capture_Video_Record, _("Start screenrecorder (Video with audio + audio-only)"));
```
sonicfind, line 453: There was a small miscommunication here. The video file does come with audio embedded in it. AVI - video only; WAV - audio only.
For now, just saying "Start screenrecorder" is probably sufficient.

plugins/LilyPad/LilyPad.rc
```diff
@@ -450,8 +450,8 @@ BEGIN
     CONTROL         "Raw input",IDC_KB_RAW,"Button",BS_AUTORADIOBUTTON,22,40,112,10
     CONTROL         "DirectInput",IDC_KB_DI,"Button",BS_AUTORADIOBUTTON,22,52,112,10
     GROUPBOX        "Game Device APIs",IDC_STATIC,16,70,191,62
-    CONTROL         "DirectInput",IDC_G_DI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,82,65,10
-    CONTROL         "XInput",IDC_G_XI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,94,125,10
+    CONTROL         "DirectInput (Legacy.)",IDC_G_DI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,82,183,10
+    CONTROL         "XInput (For modern controllers.)",IDC_G_XI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,94,125,10
```
lightningterror, line 454: Periods are not needed.

plugins/LilyPad/Tooltips.cpp
```diff
@@ -63,8 +63,13 @@ LPWSTR dialog_message(int ID, bool *updateText)
...
                    L"Use this option to check if all the inputs on a controller function properly.\n\n"
                    L"Right-click and select \"Refresh\" to update the list of devices in case a recently connected device has not shown up yet.";
+        case IDC_G_DI:
+            return L"(Legacy) Enable this if your gamepad doesn't support Xinput.\n"
+                   L"Disable for DualShock 4 (PS4 controllers) and probably others.";
         case IDC_G_XI:
-            return L"Xbox 360/Xbox One controllers (or devices/software imitating such) only";
+            return L"For Xbox 360/ Xbox One controllers (or devices supporting Xinput).\n\n"
+                   L"Running through Steam as a non-Steam game might be required for the controllers to work properly."
+                   L"Test your controller on https://gamepad-tester.com/ and if it says Xinput on top";
```
lightningterror, line 72: Missing period at end.

plugins/LilyPad/LilyPad.rc
```diff
@@ -450,8 +450,8 @@ BEGIN
     CONTROL         "Raw input",IDC_KB_RAW,"Button",BS_AUTORADIOBUTTON,22,40,112,10
     CONTROL         "DirectInput",IDC_KB_DI,"Button",BS_AUTORADIOBUTTON,22,52,112,10
     GROUPBOX        "Game Device APIs",IDC_STATIC,16,70,191,62
-    CONTROL         "DirectInput",IDC_G_DI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,82,65,10
-    CONTROL         "XInput",IDC_G_XI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,94,125,10
+    CONTROL         "DirectInput (Legacy)",IDC_G_DI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,82,100,10
+    CONTROL         "XInput (For modern controllers)",IDC_G_XI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,94,125,10
```
lightningterror, line 454: Maybe just `XInput (Modern)` ?

RedDevilus: Yeah, good idea, just had that there to be obvious it's for controllers.

lightningterror:
Commits can be better named with prefixes
example
`pcsx2-gui: Update Lilypad descriptions/tooltips.` to `Lilypad-gui: Update descriptions/tooltips.`

RedDevilus:
Changed input recording to input recording because of DPI on 100% maligns it
image

Capture:
image

and many other minor changes

pcsx2/gui/Panels/VideoPanel.cpp
```diff
@@ -30,7 +30,7 @@ Panels::FramelimiterPanel::FramelimiterPanel( wxWindow* parent )
 	: BaseApplicableConfigPanel_SpecificConfig( parent )
 {
 	m_check_LimiterDisable = new pxCheckBox( this, _("Disable Framelimiting"),
-		_("Useful for running benchmarks. Toggle this option in-game by pressing F4.") );
+		_("Uncaps FPS. Toggle in-game by pressing F4. (Useful for running benchmarks.)" ) );
```
lightningterror, line 33: Brackets seem odd here, maybe remove them.
`_("Uncaps FPS. Toggle in-game by pressing F4. Useful for running benchmarks." ) );`

pcsx2/gui/Panels/VideoPanel.cpp
```diff
@@ -288,12 +288,12 @@ Panels::VideoPanel::VideoPanel( wxWindow* parent ) :
 	wxPanelWithHelpers* right	= new wxPanelWithHelpers( this, wxVERTICAL );
 
 	m_check_SynchronousGS = new pxCheckBox( left, _("Use Synchronized MTGS"),
-		_t("For troubleshooting potential bugs in the MTGS only, as it is potentially very slow.")
+		_t("For troubleshooting potential bugs in the MTGS only, as it is potentially very slow. (Debug)")
```
lightningterror, line 291: Debug is redundant here.

pcsx2/gui/Panels/MemoryCardListView.cpp
```diff
@@ -162,9 +162,9 @@ wxString MemoryCardListView_Simple::OnGetItemText(long item, long column) const
 				//  it cannot access GetMcdProvider() which isn't (and shouldn't be) const.
 				//so.. a plain old cast does the trick. Hope it's not too bad. - avih.
 				if (((MemoryCardListView_Simple*)this)->GetMcdProvider().IsNonEmptyFilesystemCards())
-					return _("[-- Unused cards --]");
+					return _("[-- Disabled cards --]");
 				else
-					return _("[-- No unused cards --]");
+					return _("[-- No disabled cards --]");
```
lightningterror, line 167: Doesn't unused seem better?

RedDevilus: I think we should just make it into *No Cards*, No unused or even no disabled looks confusing. If you delete all memory cards then it will say the same which make it technically wrong..

RedDevilus: It will also solve the problem of DPI scaling i am having, but it is fine on 100%.

RedDevilus: Examples:
image

image

Also avoids a double negative.

pcsx2/gui/Dialogs/CreateMemoryCardDialog.cpp
```diff
@@ -252,8 +252,8 @@ void Dialogs::CreateMemoryCardDialog::CreateControls()
 		.	SetToolTip(_t("Use at your own risk.  Erratic memory card behavior is possible (though unlikely)."))
 		.	SetInt(64),
 
-		RadioPanelItem(_("Folder [experimental]"), _("Store memory card contents in the host filesystem instead of a file."))
-		.	SetToolTip(_t("Automatically manages memory card contents so that the console only sees files related to the currently running software. Allows you to drag-and-drop files in and out of the memory card with your standard file explorer. This is still experimental, so use at your own risk!"))
+		RadioPanelItem(_("Folder [best with PCSX2]"), _("Transforms the memory card into an usable folder with files inside."))
+		.	SetToolTip(_t("Gives every game 8MB for the saves (essentially a whole memcard). You can see the structure and the saves with your File Explorer. Can be used to back-up individual gamesaves instead of all saves on a memcard. Incompatible with mymc(plus) which is a save manager.You can always convert back between folder type and single memcard file."))
```
lightningterror, line 256: Maybe you can split the line code in to several so it's not one big line.

pcsx2/gui/Dialogs/ConvertMemoryCardDialog.cpp
```diff
@@ -84,15 +84,15 @@ Dialogs::ConvertMemoryCardDialog::ConvertMemoryCardDialog( wxWindow* parent, con
...
 		.SetInt( MemoryCardConversionType::MemoryCardConversion_File_16MB );
-	RadioPanelItem toFile32MB = RadioPanelItem( _( "32MB File" ), pxE( L"Convert this memory card to a 32 MB Memory Card .ps2 file." ) )
+	RadioPanelItem toFile32MB = RadioPanelItem( _( "32 MB File" ), pxE( L"Convert this memory card to a 32 MB Memory Card .ps2 file." ) )
 		.SetInt( MemoryCardConversionType::MemoryCardConversion_File_32MB );
-	RadioPanelItem toFile64MB = RadioPanelItem( _( "64MB File" ), pxE( L"Convert this memory card to a 64 MB Memory Card .ps2 file." ) )
+	RadioPanelItem toFile64MB = RadioPanelItem( _( "64 MB File" ), pxE( L"Convert this memory card to a 64 MB Memory Card .ps2 file." ) )
 		.SetInt( MemoryCardConversionType::MemoryCardConversion_File_64MB );
-	RadioPanelItem toFolder = RadioPanelItem( _( "Folder" ), _( "Convert this memory card to a folder of individual saves." ) )
+	RadioPanelItem toFolder = RadioPanelItem( _( "Folder" ), _( "Convert this memory card to a folder of individual saves. Choose this to go beyond the save-limits of 8-64MB cards and if you only use PCSX2. You can always convert back." ) )
```
lightningterror, line 95: This one too.

pcsx2/gui/Panels/GSWindowPanel.cpp
```diff
@@ -70,9 +70,9 @@ Panels::GSWindowSettingsPanel::GSWindowSettingsPanel( wxWindow* parent )
...
 
-	m_text_Zoom->SetToolTip( pxEt( L"Zoom = 100: Fit the entire image to the window without any cropping.\nAbove/Below 100: Zoom In/Out\n0: Automatic-Zoom-In untill the black-bars are gone (Aspect ratio is kept, some of the image goes out of screen).\n  NOTE: Some games draw their own black-bars, which will not be removed with '0'.\n\nKeyboard: CTRL + NUMPAD-PLUS: Zoom-In, CTRL + NUMPAD-MINUS: Zoom-Out, CTRL + NUMPAD-*: Toggle 100/0"
+	m_text_Zoom->SetToolTip( pxEt( L"Zoom = 100: Fit the entire image to the window without any cropping.\nAbove/Below 100: Zoom In/Out.\n\n0: Automatic-Zoom-In until the black-bars are gone (Aspect ratio is kept, some of the image goes out of screen).\n\nNOTE: Some games draw their own black-bars, which will not be removed with '0'.\n\nKeyboard: \nCTRL + NUMPAD-PLUS: Zoom-In, \nCTRL + NUMPAD-MINUS: Zoom-Out, \nCTRL + NUMPAD-*: Toggle 100/0"
```
lightningterror, line 75: These too in this file.

pcsx2/gui/Dialogs/CreateMemoryCardDialog.cpp
lightningterror: 8MB to 8 MB, so it matches the other changes.

RedDevilus: Noticed that PSX memcard says 1Mbit, should that be changed to 1MiB or 1MB?

lightningterror: It's fine as it is, maybe just add space after 1.

lightningterror: On the other hand change it, seems size is 1 MB.

pcsx2/gui/Dialogs/ConvertMemoryCardDialog.cpp
RedDevilus: unexpected end of file in macro expansion.
According to WX documentation: wxSTRINGIZE_T (  x  )  
I've tried different things but it gives those errors

pcsx2/gui/Dialogs/CreateMemoryCardDialog.cpp
RedDevilus: 128 KiB (PSX) is what the radiopanel says, isn't that also wrong?

lightningterror: Did a bit of search and original values are correct so leave them as they are for psx.

RedDevilus: ( !it.IsPresent ? L"" : (it.IsPSX? pxsFmt( L"%u MBit", it.SizeInMB ) : ( it.SizeInMB > 0 ? pxsFmt( L"%u MiB", it.SizeInMB )
Yeah probably this

pcsx2/gui/Dialogs/ConvertMemoryCardDialog.cpp
lightningterror: Try
```
	RadioPanelItem toFolder = RadioPanelItem( _( "Folder" ), _(
		"Convert this memory card to a folder of individual saves. "
		"Choose this to go beyond the save-limits of 8-64MB cards and if you only use PCSX2. "
		"You can always convert back." ) )
```

RedDevilus: Thanks, seems i was missing the space between ) )

pcsx2/gui/Dialogs/CreateMemoryCardDialog.cpp
RedDevilus: Is this a different style? I don't know why it's not working.

pcsx2/gui/Dialogs/ConvertMemoryCardDialog.cpp
```diff
@@ -84,15 +84,18 @@ Dialogs::ConvertMemoryCardDialog::ConvertMemoryCardDialog( wxWindow* parent, con
 void Dialogs::ConvertMemoryCardDialog::CreateControls( const MemoryCardType sourceType ) {
 	m_text_filenameInput = new wxTextCtrl( this, wxID_ANY, wxEmptyString, wxDefaultPosition, wxDefaultSize, wxTE_PROCESS_ENTER );
 
-	RadioPanelItem toFile8MB = RadioPanelItem( _( "8MB File" ), pxE( L"Convert this memory card to a standard 8 MB Memory Card .ps2 file." ) )
+	RadioPanelItem toFile8MB = RadioPanelItem( _( "8 MB File [most compatible]" ), pxE( L"Convert this memory card to a standard 8 MB Memory Card .ps2 file." ) )
```
lightningterror, line 87: Maybe change the brackets for [most compatible] to (most compatible).

pcsx2/gui/Panels/MemoryCardListView.cpp
RedDevilus: I changed it back to avoid unicode problem

lightningterror (approved):

xTVaser:
Input Recording name change is fine by me.  We'll do a pass to make sure internal naming is updated once things die down a bit. 👍

pcsx2/gui/Dialogs/ConvertMemoryCardDialog.cpp
```diff
@@ -84,15 +84,18 @@ Dialogs::ConvertMemoryCardDialog::ConvertMemoryCardDialog( wxWindow* parent, con
...
-	RadioPanelItem toFile32MB = RadioPanelItem( _( "32MB File" ), pxE( L"Convert this memory card to a 32 MB Memory Card .ps2 file." ) )
+	RadioPanelItem toFile32MB = RadioPanelItem( _( "32 MB File" ), pxE( L"Convert this memory card to a 32 MB Memory Card .ps2 file." ) )
 		.SetInt( MemoryCardConversionType::MemoryCardConversion_File_32MB );
-	RadioPanelItem toFile64MB = RadioPanelItem( _( "64MB File" ), pxE( L"Convert this memory card to a 64 MB Memory Card .ps2 file." ) )
+	RadioPanelItem toFile64MB = RadioPanelItem( _( "64 MB File" ), pxE( L"Convert this memory card to a 64 MB Memory Card .ps2 file." ) )
 		.SetInt( MemoryCardConversionType::MemoryCardConversion_File_64MB );
-	RadioPanelItem toFolder = RadioPanelItem( _( "Folder" ), _( "Convert this memory card to a folder of individual saves." ) )
+	RadioPanelItem toFolder = RadioPanelItem( _( "Folder" ), _(
+		"Convert this memory card to a folder of individual saves. "
+		"Choose this to go beyond the save-limits of 8-64 MB cards and if you only use PCSX2. "
```
RedPanda4552, line 97: > "Choose this to go beyond the save-limits of 8-64 MB cards and if you only use PCSX2

This does work but I think some of the phrasing is a bit ambiguous. The major points to highlight here should be

1. There is no maximum capacity. Don't worry about saying it "goes beyond what file memcards do", it would be better to just tell it how it is, "No capacity limit".
2. Folder memcards are exclusive to PCSX2. "If you only use PCSX2" is ok, but I am personally biased towards "Not compatible with other PS2 emulators.
3. One selling point of folder memcards is that they make it easy to directly access the save files the game uses. I would consider adding a bit about that, maybe something like "Allows direct access to individual save files".
4. I think "You can always convert back" is pointless; it's already implied since the user is already converting.

pcsx2/gui/Dialogs/CreateMemoryCardDialog.cpp
```diff
@@ -252,8 +252,12 @@ void Dialogs::CreateMemoryCardDialog::CreateControls()
 		.	SetToolTip(_t("Use at your own risk.  Erratic memory card behavior is possible (though unlikely)."))
 		.	SetInt(64),
 
-		RadioPanelItem(_("Folder [experimental]"), _("Store memory card contents in the host filesystem instead of a file."))
-		.	SetToolTip(_t("Automatically manages memory card contents so that the console only sees files related to the currently running software. Allows you to drag-and-drop files in and out of the memory card with your standard file explorer. This is still experimental, so use at your own risk!"))
+		RadioPanelItem(_("Folder [best with PCSX2]"), _("Transforms the memory card into an usable folder with files inside."))
```
RedPanda4552, line 255: Not sure I like "best with PCSX2", I think "Recommended" would fit the same theme as the rest of the application. Usually we just say "Recommended" or "Not Recommended". 

"Transforms the memory card into an usable folder with files inside." - I do not like this, it does not really say anything relevant to what the folder memory card is. I much prefer the original to this.

pcsx2/gui/Dialogs/CreateMemoryCardDialog.cpp
```diff
@@ -252,8 +252,12 @@ void Dialogs::CreateMemoryCardDialog::CreateControls()
...
+		.	SetToolTip(_t("Gives every game 8 MB for the saves (essentially a whole memcard).\n"
```
RedPanda4552, line 256: I suppose this is sort of accurate but it can be confusing; it may imply that it's allocating 8 MB on the host system for each game, which is not the case. I would maybe consider saying something like "Stores memory card contents in a folder. Only exposes the save files for the running game, rather than the entire memory card".

pcsx2/gui/Dialogs/CreateMemoryCardDialog.cpp
```diff
@@ -252,8 +252,12 @@ void Dialogs::CreateMemoryCardDialog::CreateControls()
 		.	SetToolTip(_t("Use at your own risk.  Erratic memory card behavior is possible (though unlikely)."))
 		.	SetInt(64),
 
-		RadioPanelItem(_("Folder [experimental]"), _("Store memory card contents in the host filesystem instead of a file."))
-		.	SetToolTip(_t("Automatically manages memory card contents so that the console only sees files related to the currently running software. Allows you to drag-and-drop files in and out of the memory card with your standard file explorer. This is still experimental, so use at your own risk!"))
+		RadioPanelItem(_("Folder [best with PCSX2]"), _("Transforms the memory card into an usable folder with files inside."))
+		.	SetToolTip(_t("Gives every game 8 MB for the saves (essentially a whole memcard).\n" 
+						  "You can see the structure and the saves with your File Explorer.\n"
+						  "Can be used to back-up individual saves instead of all saves on a memcard.\n" 
+						  "Incompatible with mymc(plus) which is a save manager.\n"
```
RedPanda4552, line 259: I do hesitate to list off external programs by name but if necessary (I feel it kind of is because MyMC is a bit of a lesser known tool), I would consider "Incompatible with PS2 memory card editing tools (e.g. MyMC, MyMCPlus)". This way it says the tools as a general concept are incompatible, not just MyMC.

pcsx2/gui/Dialogs/McdConfigDialog.cpp
```diff
@@ -36,13 +36,13 @@ Panels::McdConfigPanel_Toggles::McdConfigPanel_Toggles(wxWindow *parent)
...
 		_("Auto-eject memory cards when loading savestates"),
-		pxE( L"Avoids memory card corruption by forcing games to re-index card contents after loading from savestates.  May not be compatible with all games (Guitar Hero)."
+		pxE( L"Avoids broken memory card saves. May not work with some games such as Guitar Hero."
 		)
 	);
 
 	m_folderAutoIndex = new pxCheckBox( this,
 		_( "Automatically manage saves based on running game" ),
-		pxE( L"(Folder type only) Re-index memory card content every time the running software changes. This prevents the memory card from running out of space for saves."
+		pxE( L"(Folder type only / Card size: Auto) Checks the memory card every time you start a game. Avoids running out of space for saves."
```
RedPanda4552, line 45: > Card size: Auto

Remove this, it doesn't make sense to add this detail here.

> Checks the memory card every time you start a game

Doesn't really tell us much. I would consider something like "Loads only the booted game's save file and ignores all others".

pcsx2/gui/Panels/VideoPanel.cpp
```diff
@@ -293,7 +293,7 @@ Panels::VideoPanel::VideoPanel( wxWindow* parent ) :
 
 	m_restore_defaults = new wxButton(right, wxID_DEFAULT, _("Restore Defaults"));
 
-	m_check_SynchronousGS->SetToolTip( pxEt( L"Enable this if you think MTGS thread sync is causing crashes or graphical errors.")
+	m_check_SynchronousGS->SetToolTip( pxEt( L"Enable this if you think MTGS thread sync is causing crashes or graphical errors. For debugging to see if GS is running at the correct speed.")
```
RedPanda4552, line 296: > For debugging to see if GS is running at the correct speed

Not sure what the validity behind this statement is and considering that I wouldn't feel comfortable saying it's one of the reasons the setting exists.

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -92,9 +92,9 @@ const char* dialog_message(int ID, bool* updateText) {
 		case IDC_OSD_MONITOR:
 			return "Continuously prints the FPS and the EE, GS and VU(if the MTVU speedhack is enabled) percentages onscreen.";
 		case IDC_PALTEX:
-			return "When checked 4/8 bits texture will be send to the GPU with a palette. GPU will be in charge of the conversion.\n\n"
-				"When unchecked the CPU will convert directly the texture to 32 bits.\n\n"
-				"It is basically a trade-off between GPU/CPU.";
+			return "Enabled: GPU will handle 4/8 bits textures while CPU does the rest.\n"
```
RedPanda4552, line 95: I would want a GS expert to proof this one but to me this doesn't sound right. To me, what I think it should be:

Enabled: GPU will handle conversion of 4 and 8 bit textures to 32 bit.
Disabled: CPU will handle conversion of 4 and 8 bit textures to 32 bit.

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -196,10 +199,11 @@ const char* dialog_message(int ID, bool* updateText) {
 			return "By default, the texture cache handles partial invalidations. Unfortunately it is very costly to compute CPU wise."
 				"\n\nThis hack replaces the partial invalidation with a complete deletion of the texture to reduce the CPU load.\n\nIt helps snowblind engine games.";
 		case IDC_LARGE_FB:
-			return "Allocate a large framebuffer to be compliant with GS memory (Prevents FMV flickering).\n"
-				"It increases GPU/memory requirements.\n\n"
-				"Note: It should be disabled for Armored Core, Destroy All Humans, and Gran Turismo.\n"
-				"It will amplify RAM/VRAM spikes. Other games might be affected as well.";
+			return "Reserves a larger framebuffer for textures to prevent FMV flickers.\n"
```
RedPanda4552, line 202: I don't think this is for textures; isn't this for the actual framebuffer used to draw to screen? Maybe just remove "for textures" and that would be all it takes.

plugins/LilyPad/LilyPad.rc
```diff
@@ -450,8 +450,8 @@ BEGIN
     CONTROL         "Raw input",IDC_KB_RAW,"Button",BS_AUTORADIOBUTTON,22,40,112,10
     CONTROL         "DirectInput",IDC_KB_DI,"Button",BS_AUTORADIOBUTTON,22,52,112,10
     GROUPBOX        "Game Device APIs",IDC_STATIC,16,70,191,62
-    CONTROL         "DirectInput",IDC_G_DI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,82,65,10
-    CONTROL         "XInput",IDC_G_XI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,94,125,10
+    CONTROL         "DirectInput (Legacy)",IDC_G_DI,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,22,82,100,10
```
RedPanda4552, line 453: *palpatine_good_good.jpg*

plugins/LilyPad/Tooltips.cpp
```diff
@@ -35,8 +35,8 @@ LPWSTR dialog_message(int ID, bool *updateText)
                    L"Note 2: This does not enable the mouse to function as an in-game mouse in PS2 games that support a USB mouse or lightgun."
                    L"This requires a USB plugin, while LilyPad is a PAD plugin(PAD means it emulates devices that are to be plugged into the PlayStation controller port that connects the DualShock 2 controller).";
         case IDC_MOUSE_UNFOCUS:
-            return L"Starts the emulation without the mouse being in focus by the application.\n\n"
-                   L"Without mouse focus, the mouse can still be used outside of the emulation.\nWith the mouse in focus it can only be used as an input for the emulator.";
+            return L"Enabled: Mouse is unfocused and can be used with emulation and generally in your OS.\n\n"
```
RedPanda4552, line 38: > can be used with emulation and generally in your OS

Not sure what this is supposed to mean, this almost makes it sound like PCSX2 would normally steal mouse control away from your PC entirely which isn't right. Maybe "Enabled: Emulation starts with the mouse unfocused"?

plugins/LilyPad/Tooltips.cpp
```diff
@@ -35,8 +35,8 @@ LPWSTR dialog_message(int ID, bool *updateText)
...
+                   L"Disabled: Mouse is focused and can ONLY be used with emulation.";
```
RedPanda4552, line 39: Yeah, see above, definitely consider changing this, sounds almost... predatory?

plugins/LilyPad/Tooltips.cpp
```diff
@@ -63,8 +63,13 @@ LPWSTR dialog_message(int ID, bool *updateText)
                    L"Double-click a device in the list or right-click it and select \"Test Device\" to display a continuously updated list of the state of all inputs on the selected device.\n"
                    L"Use this option to check if all the inputs on a controller function properly.\n\n"
                    L"Right-click and select \"Refresh\" to update the list of devices in case a recently connected device has not shown up yet.";
+        case IDC_G_DI:
+            return L"(Legacy) Enable this if your gamepad doesn't support Xinput.\n"
+                   L"Disable for DualShock 4 (PS4 controllers) and probably others.";
         case IDC_G_XI:
-            return L"Xbox 360/Xbox One controllers (or devices/software imitating such) only";
+            return L"For Xbox 360/ Xbox One controllers (or devices supporting Xinput).\n\n"
+                   L"Running through Steam as a non-Steam game might be required for the controllers to work properly.\n\n"
```
RedPanda4552, line 71: > "Running through Steam as a non-Steam game might be required for the controllers to work properly"

I would specify that this is for controllers which do not support Xinput. For such controllers Steam ends up converting the controller's inputs into Xinput, allowing you to use Xinput bindings in Lilypad. Steam is completely unnecessary for controllers which already use Xinput, and this makes it sound like Steam is still sometimes needed for Xinput controllers.

RedPanda4552 (approved):
Had a healthy number of comments to make but still overall would approve merging as is. Would like to see some changes but either way is fine.

pcsx2/gui/Dialogs/McdConfigDialog.cpp
RedDevilus: > Card size: Auto

To me it makes more sense though unless you want to change card size to folder? Not everyone knows that card size = folder technically.

> Doesn't really tell us much. I would consider something like "Loads only the booted game's save file and ignores all others".

Suggestion: Limit to the relevant booted game saves, ignoring others.

pcsx2/gui/Panels/VideoPanel.cpp
RedDevilus: To be honest i've heard multiple things about this setting, to a legacy setting to putting on a separate thread. I've already deleted Debug tag on the description itself.

plugins/GSdx/Window/GSSetting.cpp
RedDevilus: Console changes stuff like 512x512 to 512x3012 if enabled. I'll remove the words though.

plugins/LilyPad/Tooltips.cpp
RedDevilus: Wasn't my original description.
This one was:
L"Enabled: Mouse is unfocused and can be used with emulation and outside it for general usage.\n\n"
Still current one is heavy

plugins/LilyPad/Tooltips.cpp
RedDevilus: I would call it to the point, perhaps a bit too blunt for some, i'll water it down a bit.

plugins/LilyPad/Tooltips.cpp
RedDevilus: Well i tried to be open with it:
Enable this if you have Xinput, if it doesnt work then do it via steam.

plugins/GSdx/Window/GSSetting.cpp
RedDevilus: Also the reason i want to change the tooltips because they even confuse me.

RedDevilus:
I haven't added more changes to the PR for delay hell and to accelerate other PRs.

PR #3463, opened 2020-06-16
GSdx-hw: use CanUseSwSpriteRender in OI_JakGames.

AlessandroVetere:
In GSRendererHW, reuse the method CanUseSwSpriteRender for OI_JakGames.

As in current master, 16x16 "flat" sprite draws are performed via CPU: those are palette rendering better emulated with the SwSpriteRender.

Jak games do also many 64x64 "flat" draws, but mostly for render to texture effects that are already correctly emulated in the TC, so those draws are not done via CPU (as on the contrary happens for DBZ BT games).

### DIFF
--- a/plugins/GSdx/Renderers/HW/GSRendererHW.cpp
+++ b/plugins/GSdx/Renderers/HW/GSRendererHW.cpp
@@ -2193,60 +2193,13 @@ bool GSRendererHW::OI_ItadakiStreet(GSTexture* rt, GSTexture* ds, GSTextureCache
 
 bool GSRendererHW::OI_JakGames(GSTexture* rt, GSTexture* ds, GSTextureCache::Source* t)
 {
-	GSVector4i r = GSVector4i(m_vt.m_min.p.xyxy(m_vt.m_max.p)).rintersect(GSVector4i(m_context->scissor.in));
-	GSVector4i r_p = GSVector4i(0, 0, 16, 16);
-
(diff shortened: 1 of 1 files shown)


lightningterror:
Maybe we can have a generic solution that is always on/auto detected, will probably help other games too,

AlessandroVetere:
The detection is rather good, but as you said, not all the relevant GS features are supported yet in the rendering code.

We could implement a subset of those missing features which implementation is feasible, and modify the detection code to skip those not supported (dithering or antialiasing for example).

Then we could keep the solution always on/auto detected, as you suggested, and it would be surely beneficial for the palette rendering (and small sprites too) due to integer math and no upscaling in the rendering code.

lightningterror:
Jak games seem fine, an odd issue also present on master on Jak X I noticed are the car wheels looking a bit off, other than that everything seems fine from the few dumps I tested and a short gameplay of Jak X.


[Jak X car wheels.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/4806765/Jak.X.car.wheels.gs.xz.zip)

AlessandroVetere:
Thank you for testing, I checked and the issue is present also in current master.
May it be related to the error message: `T:Error	ID:1619	S:Mid	=> Reading RT as a packed-indexed 8 bits format` ?
Anyway, if it is confirmed that the Jak X issue is not a regression, we might address it separately.

PR #3475, opened 2020-06-19
gsdx-ogl: Fix MaxDepthPS byte alignment.

lightningterror:
Bigger value comes first, then smaller to fill the gap so the alignment
can be proper.
vec3
float


plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
```diff
@@ -177,14 +177,14 @@ void GSRendererOGL::EmulateZbuffer()
 	const bool clamp_z = (uint32)(GSVector4i(m_vt.m_max.p).z) > max_z;
 
 	vs_cb.MaxDepth = GSVector2i(0xFFFFFFFF);
```
AlessandroVetere, line 179: Did you check if the same could apply also here ?

refractionpcsx2: We checked those, that all aligns fine on the vertex buffer

PR #3476, opened 2020-06-19
Updating Spanish translation

IlDucci:
Overhauls the previous batch of changes and adds the latest changes. Changes include:
 - Unification of terminology between all the strings.
 - Changing quotation marks with the ones used in Spanish.
 - Removal of some GT-like translations.
 - Unification of parenthesis symbols.
 - Removal of title uppercasing (Spanish language doesn't use that many uppercased words).
 - Checked fuzzy entries.


locales/es_ES/pcsx2_Iconized.po
```diff
@@ -8,16 +8,16 @@ msgstr ""
 "Project-Id-Version: PCSX2 1.6.0\n"
```
lightningterror, line 8: Version 1.7.0

locales/es_ES/pcsx2_Main.po
```diff
@@ -8,16 +8,16 @@ msgstr ""
 "Project-Id-Version: PCSX2 1.6.0\n"
```
lightningterror, line 8: Here too.

IlDucci:
Solved, @lightningterror .

locales/es_ES/pcsx2_Main.po
```diff
@@ -420,16 +419,16 @@ msgid ""
...
 "successfully but could not be moved to its final resting place."
 msgstr ""
-"El guardado rápido no se guardó correctamente.  El archivo temporal ha sido "
+"El guardado rápido no se guardó correctamente. El archivo temporal ha sido "
 "creado con éxito, pero no se pudo trasladar a su destino final."
 
 #: pcsx2/gui/AppConfig.cpp:976
 msgid "Safest (No hacks)"
-msgstr "Más seguro (Sin hacks)"
+msgstr "Lo más seguro (sin arreglos)"
```
lightningterror, line 427: Maybe `sin` was fine with caps, since it matched the original.

locales/es_ES/pcsx2_Main.po
```diff
@@ -420,16 +419,16 @@ msgid ""
...
 
 #: pcsx2/gui/AppConfig.cpp:977
 msgid "Safe (Default)"
-msgstr "Seguro (Predeterminado)"
+msgstr "Seguro (predeterminado)"
```
lightningterror, line 431: Ditto, didn't check all changes.

IlDucci:
All instances of text between parentheses had their first character lowercased following the RAE's rules.

locales/es_ES/pcsx2_Main.po
```diff
@@ -445,7 +444,7 @@ msgstr "Muy agresivo"
 
 #: pcsx2/gui/AppConfig.cpp:981
 msgid "Mostly Harmful"
-msgstr "Dañino"
+msgstr "Perjudicial"
```
RedDevilus, line 447: English is Mostly Harmful.
Doesnt it need Mayormente?

IlDucci: You are right, that does need something that separates "Harmful" from "Mostly Harmful". Correcting.

RedDevilus:
bin/Langs/es_ES/pcsx2_Iconized.mo
 bin/Langs/es_ES/pcsx2_Main.mo
Are missing for PCSX2.
They are the actual translations to be used for the program. The PO is only to make it easier for checking translation and have to be the same as the MO. Poedit has an option to compile to mo.

IlDucci:
Sorry about that, last time I did edits for PCSX2, the MO files were not a requirement. Compiling.

locales/es_ES/pcsx2_Main.po
```diff
@@ -1239,14 +1236,12 @@ msgid "Welcome to PCSX2!"
...
-#, fuzzy
 msgid "Configuration Guide"
-msgstr "Guías de configuración (internet)"
+msgstr "Guía de configuración"
 
 #: pcsx2/gui/Dialogs/FirstTimeWizard.cpp:110
-#, fuzzy
 msgid "Readme / FAQ"
-msgstr "Léeme / Preguntas frecuentes (desconectado/PDF)"
+msgstr "Léeme/Preguntas frecuentes"
```
RedDevilus, line 1244: Are you sure the empty space must be gone?

IlDucci: In this case, the slash is used to separate two possible options or choices, so RAE says spaces shouldn't be applied here. Would be point 1. b) in here: https://www.rae.es/dpd/barra

IlDucci: When the slash is used to offer options, according to RAE, no spaces should be in between the slash. It should be rule 1. b) here: https://www.rae.es/dpd/barra

And you can also see examples in Wikilengua: https://es.wikipedia.org/wiki/Barra_(tipograf%C3%ADa)

PR #3477, opened 2020-06-20
Misc: plugins gui changes.

lightningterror:
Fix text being cut off on some options when using dpi scaling.


RedDevilus:
Tested with 2560:1920 on 175%
PR:
image
While it fixes F5 , buttons are hiding under my taskbar

image
Seems fine

image
Seems fine too

Master:
image
image
image

PR #3478, opened 2020-06-20
GSdx-hw, TC: avoid 4/8b convert false positives.

iMineLink:
In GSdx hardware TC, avoid fishy RT conversion to 4/8 bit format false positives, by considering as valid conversion cases only those with full RT readback.

Fixes Jak X wheels rendering.


lightningterror:
It breaks Harry Potter, don't know which other games are broken.

[Harry Potter-8bit.zip](https://github.com/PCSX2/pcsx2/files/4808675/Harry.Potter-8bit.zip)

PR #3479, opened 2020-06-21
GSdx-hw, TC: improve EE write overlap detection.

iMineLink:
Improve EE write overlap detection reusing tex in rt code.

Fixes Jak X wheels rendering.
Fixes Jak 3 striped and shimmering textures.
Makes MCIII worse by side effect, evaluating a cure.

### DIFF
--- a/plugins/GSdx/Renderers/HW/GSTextureCache.cpp
+++ b/plugins/GSdx/Renderers/HW/GSTextureCache.cpp
@@ -321,7 +321,7 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
 					sok.surf_m_end_block = t->m_end_block;
 					sok.surf_tex0_tbp0 = t->m_TEX0.TBP0;
 					const SurfaceOffset so = ComputeSurfaceOffset(sok);
-					if (so.is_valid_offset)
+					if (so.is_valid)
 					{
 						dst = t;
 						x_offset = so.offset.x;
@@ -863,8 +863,8 @@ void GSTextureCache::InvalidateVideoMem(GSOffset* off, const GSVector4i& rect, b
 			// GH: Try to detect texture write that will overlap with a target buffer
 			if(GSUtil::HasSharedBits(psm, t->m_TEX0.PSM))
 			{
-				const SurfaceOffset so = ComputeSurfaceWriteOffset(off, r, t);
-				if (so.is_valid_offset)
+				const SurfaceOffset so = ComputeSurfaceOffset(off, r, t);
+				if (so.is_valid)
 				{
 					GL_CACHE("TC: Dirty After Target(%s) %d (0x%x->0x%x) pos(%d,%d => %d,%d) bw:%u",
 						to_string(type),
@@ -2013,7 +2013,7 @@ void GSTextureCache::AttachPaletteToSource(Source* s, uint16 pal, bool need_gs_t
 	s->m_palette = need_gs_texture ? s->m_palette_obj->GetPaletteGSTexture() : nullptr;
 }
 
-GSTextureCache::SurfaceOffset GSTextureCache::ComputeSurfaceWriteOffset(GSOffset* off, const GSVector4i& r, Target* t)
+GSTextureCache::SurfaceOffset GSTextureCache::ComputeSurfaceOffset(const GSOffset* off, const GSVector4i& r, Target* t)
 {
 	if (!off || !t)
 		return { false, GSVector4i(0, 0, 0, 0) };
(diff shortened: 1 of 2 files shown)


Dreadmoth:
Also fixes striped, shimmering textures in Jak 3:

[Master](https://user-images.githubusercontent.com/8492684/85218037-b9850880-b3d9-11ea-8e1b-4d6f790944e6.png)

[PR](https://user-images.githubusercontent.com/8492684/85218042-be49bc80-b3d9-11ea-86f8-31e6a8973c29.png)

iMineLink:
I'm going to make some important changes here. Thanks for testing the current version. Do not merge.

plugins/GSdx/Renderers/HW/GSTextureCache.cpp
```diff
@@ -2108,6 +2013,122 @@ void GSTextureCache::AttachPaletteToSource(Source* s, uint16 pal, bool need_gs_t
...
+					so.offset.w = surf_max_valid_w;
+				}
+				break;
+			}
+		}
+		if (so.is_valid_offset)
+			break;
+	}
+	// Clear cache if size too big.
+	while (m_surface_offset_cache.size() + 1 > S_SURFACE_OFFSET_CACHE_MAX_SIZE)
```
tadanokojin, line 2112: why a while instead of an if

tadanokojin:
Link,

Would you consider looking into Burnout engine games? I have debugged the black sky issue and tex in rt seems an appropriate fix. Current master however will break reflections.

Rendering pattern for Burnout Revenge:
Renders sky data to RT (0x2c00) with offset in x direction by 32 pixels.
Samples sky data with no offset (0x2c10)

Reflection issue appears to be that they are using an old buffer for the reflections (loading screen) and this is causing issues with master's tex in rt code.

I tried this PR with tex in rt set in the config but it appears to have regressed back to just black sky.

plugins/GSdx/Renderers/HW/GSTextureCache.cpp
iMineLink: Given that the map is cleared there is no valid reason, will switch to if

iMineLink:
This PR is aggressively making dirty old RTs and this breaks also MCIII (even with Tex in RT disabled): to fix this I'm making some changes in how dirty targets are handled (see Target::Update), basically allowing to copy data from one target to another without reading from GS memory if the dirty offset is matched in one valid target currently in the cache ("Tex in RT for the overlapped writes").

The major hassle will be format conversion, but that might be solved with a forced readback/reupload if we do not have the fast way (shaders) to copy convert (I'm still studying the current code so the solution might already be there).

MCIII is doing overlapped writes (with format conversion...) for the headlights effect.

That said, I'll surely give a look at Burnout, could you drop here a reference gsdump please?

tadanokojin:
I can drop you once but given the nature of the issue, I have to provide you one which encompasses the entire loading screen. If that is fine by you then I'll get it to you later today.

tadanokojin:
Here is the dump for Burnout Revenge. Be warned it's many frames (sky draw is done during loading)
https://dumps.pcsx2.net/index.php/s/P7YqMom56F8gSoB

iMineLink:
Thank you. That would be though to analyze in depth, I'll see how it reacts to MCIII cure.

tadanokojin:
Good news is burnout games are already in the CRC list so it's simple enough to add them. But you know where to find me if you need my help. Wish you luck.

plugins/GSdx/Renderers/HW/GSTextureCache.h
```diff
@@ -213,8 +228,8 @@ class GSTextureCache
 	static bool m_disable_partial_invalidation;
 	bool m_texture_inside_rt;
 	static bool m_wrap_gs_mem;
-	uint8 m_texture_inside_rt_cache_size = 255;
-	std::vector<TexInsideRtCacheEntry> m_texture_inside_rt_cache;
+	constexpr static size_t S_SURFACE_OFFSET_CACHE_MAX_SIZE = 65536;
```
tadanokojin, line 231: `UINT16_MAX` or s`td::numerical_limits<uint16>::max()` would be clearer.

ghenry22:
Curious if you managed to find anything regarding the black sky bug in burnout engine games?  It's probably the only major thing that is still broken in them when in HW rendering mode.

Would be happy to help test if that is useful.

iMineLink:
This will need a rework after #3895 is, if ever, merged.
Changing how VRAM data is invalidated and/or read passthough is not easy though in the current texture cache implementation.
I'm not positive that this "spot fix" approach is the correct one, given the small regressions that easily spawn while doing such work.

iMineLink:
Superseded by #3895.

PR #3480, opened 2020-06-21
Corrections to recording file's header data

xTVaser:
Two things updated about the recording's header data, none of which are currently breaking to the feature, but just some extra metadata:

1. The emulator version was hard-coded to `PCSX2-1.5.X` which is no longer accurate, it is no longer hard-coded and will use the correct version number the same way it is calculated to display in the window title.
2. When creating a recording, the game name is preferred over the ISO name.  This is determined via the GameDB.  When playing back the recording, a simple check occurs to see if the game running is the same one that was used to make the recording.  On the playback side, it always only checked with the ISO filename.


pcsx2/Recording/InputRecording.cpp
```diff
@@ -191,20 +179,38 @@ void InputRecording::Play(wxString FileName, bool fromSaveState)
...
 	recordingConLog(wxString::Format(L"[REC]: MaxFrame: %d\n", InputRecordingData.GetMaxFrame()));
 	recordingConLog(wxString::Format(L"[REC]: UndoCount: %d\n", InputRecordingData.GetUndoCount()));
 }
 
+wxString InputRecording::resolveGameName()
+{
+	// Code loosely taken from AppCoreThread::_ApplySettings to resolve the Game Name
+	wxString gameName;
+	const wxString gameKey(SysGetDiscID());
+	if (!gameKey.IsEmpty()) {
```
lightningterror, line 202: Would be nice if the formatting remained the same.

PR #3482, opened 2020-06-21
GameDB: Ratchet 2/3/4 shadow fixes.

atomic83GitHub:
This PR add several vu clamping mode fixes to various Ratchet and Clank titles.

- Resolve missing shadow geometry and some transparency issues.

***Before the fixes***

Capture

***After the fixes***

Capture2

Please note that Japanesse versions of Ratchet and Clank games are not perfect and might need a better COP2 handling in the futur.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -650,6 +650,7 @@ Region = NTSC-Unk
 Serial = SCAJ-20109
 Name   = Ratchet & Clank 3 - Up your Arsenal
 Region = NTSC-Unk
+vuClampMode = 2 // Fixes Shadow issues with transparency and geometry.
 ---------------------------------------------
 Serial = SCAJ-20110
 Name   = Dragon Quest VIII - Sora to Daichi to Norowareshi Himegimi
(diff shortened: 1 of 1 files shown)


Dreadmoth:
Maybe I'm missing something, but I don't see any changes to shadow rendering with these settings applied. 

Current master with default settings, except 4x Native resolution:
gsdx_20200622135127_

Checked a variety of scenarios in R&C 2/3, including overlapping transparent textures and geometry / water.

R&C 2/3 PAL Platinum releases
Windows 10 64-bit, version 2004
Intel Core i7-4770
Nvidia GTX 1070 Ti, driver version 446.14

Hirato:
I think the Auto Flush hack was meant to do the shadows properly (Hacks section of Advanced Settings)
It's slow, so long ago, gregory added a shader to draw the shadows instead (for HW only) in a performant way.

So for Software rendering,
in R&C2 the shadows produce a very intersecting effect
and in R&C3, they're much too thick.
Both mostly go away if I enable Auto-Flush for the software renderer, but the settings offered here do not help any.

MonJamp:
Shadows are fixed with OpenGL hardware renderer with Blending Accuracy set to Basic. In both Direct3D and OpenGL messing with the VU clamping or toggling `Delay V1F1 Stalls` makes no difference. I'm unable to reproduce how you got the before pic but I'm pretty sure these settings aren't necessary.

atomic83GitHub:
I actually found out that VU clamping alone is enough to solve minor shadow edges issues.

lightningterror:
So what are we going to do with this?

PR #3483, opened 2020-06-21
Add new debugger controls to go-to specific disassembly or memory addresses

xTVaser:
Lately I've been using pcsx2's debug view a lot to look at specific memory addresses or disassembly values.  Currently the only way I've found to jump to addresses is to create a breakpoint and then use that breakpoint to jump to the disassembly or memory value by double clicking it.

This is pretty clunky, but is especially annoying if whatever you are debugging crashes the emulator / requires the emulator to be restarted between iterations as you have to setup the breakpoint each time.

So this just adds some simple text fields next to the existing debug control buttons to jump to an explicit memory address or disassembly address.

Demo Video - https://www.youtube.com/watch?v=3MdNQaj9kM0

### DIFF
--- a/pcsx2/gui/Debugger/DisassemblyDialog.cpp
+++ b/pcsx2/gui/Debugger/DisassemblyDialog.cpp
@@ -266,9 +266,21 @@ DisassemblyDialog::DisassemblyDialog(wxWindow* parent):
 	Bind(wxEVT_BUTTON, &DisassemblyDialog::onBreakpointClicked, this, breakpointButton->GetId());
 	topRowSizer->Add(breakpointButton);
 
-	helpButton = new wxButton( panel, wxID_ANY, L"Help");
+	helpButton = new wxButton(panel, wxID_ANY, L"Help");
 	Bind(wxEVT_BUTTON, &DisassemblyDialog::onHelpClicked, this, helpButton->GetId());
-	topRowSizer->Add(helpButton);
+	topRowSizer->Add(helpButton, 0, wxRIGHT, 8);
+
+	goToDisassemblyAddressField = new wxTextCtrl(panel, wxID_ANY, wxEmptyString, wxDefaultPosition, wxDefaultSize, wxTE_PROCESS_ENTER);
+	goToDisassemblyAddressField->SetHint("Go to Address in Disassembly");
+	goToDisassemblyAddressField->SetMinSize(wxSize(goToDisassemblyAddressField->GetTextExtent("Go to Address in Disassembly   ").GetWidth(), wxDefaultSize.GetHeight()));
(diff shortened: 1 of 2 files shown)


kozarovv:
You can do that already by pressing CTRL + G, work for both memory, and disassembly view. Just need to click on any of them to select memory, or disasm. You probably missed it, because window have weird description "Enter expression".
Actually this is explained in docs here: https://github.com/PCSX2/pcsx2/blob/master/bin/docs/debugger.txt

image

xTVaser:
Ah I see, yeah that is quite hidden, I wasn't even aware of the `debugger.txt` file.  One benefit to this PR is the UI is more clear in it's intention instead of the generic "Enter Expression".  But also it does not require an additional dialog window.  For me that feels a bit more usable, but I'm biased so I'll leave it up to the review.

pcsx2/gui/Debugger/DisassemblyDialog.cpp
```diff
@@ -265,9 +265,21 @@ DisassemblyDialog::DisassemblyDialog(wxWindow* parent):
 	Bind(wxEVT_BUTTON, &DisassemblyDialog::onBreakpointClicked, this, breakpointButton->GetId());
 	topRowSizer->Add(breakpointButton);
 
-	helpButton = new wxButton( panel, wxID_ANY, L"Help");
+    helpButton = new wxButton( panel, wxID_ANY, L"Help");
```
lightningterror, line 268: Formatting.

pcsx2/gui/Debugger/DisassemblyDialog.h
```diff
@@ -102,18 +102,30 @@ class DisassemblyDialog : public wxFrame
...
 	wxNotebook* middleBook;
 
 	wxBoxSizer* topSizer;
-	wxButton *breakRunButton, *stepIntoButton, *stepOverButton, *stepOutButton, *breakpointButton, *helpButton;
+	wxButton* breakRunButton;
+	wxButton* stepIntoButton;
+	wxButton* stepOverButton;
+	wxButton* stepOutButton;
+	wxButton* breakpointButton;
+    wxButton* helpButton;
```
lightningterror, line 126: Here as well.

pcsx2/gui/Debugger/DisassemblyDialog.cpp
```diff
@@ -589,6 +624,36 @@ void DisassemblyDialog::gotoPc()
 	iopTab->getDisassembly()->gotoPc();
 }
 
+void DisassemblyDialog::gotoAddress(wxString fieldText, bool disasmAddress)
+{
+	if (currentCpu == NULL) {
```
lightningterror, line 629: Would be nice to follow clang style, same for the changes below.

GovanifY:
As the functionality already exists, even if not that well explained, I'll close this PR. @RedDevilus is trying in #4380 to get the debug UI on a better standing as to make this easier, so we should either move conversation there or open a new issue/PR

PR #3497, opened 2020-06-25
GSdx-gui: Limit spin buttons input properly.

### LINKED ISSUE #3492: [BUG] Can not make more threads, then 16
Good night!
I have 2 sockets: Intel Core Xeon 
So, I have 20 real and 40 virtual proccessor cores.
I tried to plase 18 cores.
image
But, when I open plugin windows again - cores are only 16
image

lightningterror:
v1 It won't allow users to try to input a value that isn't supported
osd (size, timeout, max), tc offsets (x, y), rendering threads.

v2 Use const if applicable for variables throughout
GSSettingsDlg.cpp. Will cleanup some warnings.

Close #3492

### DIFF
--- a/plugins/GSdx/Window/GSSettingsDlg.cpp
+++ b/plugins/GSdx/Window/GSSettingsDlg.cpp
@@ -81,7 +81,7 @@ std::vector<GSSettingsDlg::Adapter> GSSettingsDlg::EnumerateD3D11Adapters()
 
 		D3D_FEATURE_LEVEL level = GSUtil::CheckDirect3D11Level(adapter, D3D_DRIVER_TYPE_UNKNOWN);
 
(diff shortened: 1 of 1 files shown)


RedDevilus:
Same for;

GSWindow: Zoom and Custom Window Size
image
Apply gets this warning: 
image
And the wrong numbers are still there (not reset to previous or working numbers).

Framelimiter:
image
If you put lots of numbers it will default to 0% but not for slow motion
Exception:
image

Frameskipping:
image
Putting lots of numbers will default to 0 if you go to different one (not press apply)
Apply gets this warning: 
image
Exception again:
image
You will get the fixed int warning if you apply but the zeros still stick.

The only one who works as intended is Skipdraw if u put 4 numbers:
image

And if u put 5 numbers:
image

lightningterror:
PR is only for gsdx spin buttons.

RedDevilus:
Oh i see, thought it also was for GUI.
Like this weird behavior. Perhaps in future PR.
image

PR #3501, opened 2020-06-27
Mpeg.cpp/h indentation

Florin9doi:


ssakash:
While I do appreciate code cleanups, it feels like pushing commits for small cleans up on individual files here and there only ends up making a mess of the Git history.

If you want to do a clean up, it might be better to do it on one whole submodule rather than just picking a specific file and fixing some indentations or spaces, if it was some functional change included it makes sense, but just for fixing some indentation in a file on a project with over 300 files rather insignificant.

ssakash:
That was just my opinion regarding it, think we'll come to a stance regarding it later, your PR is fine for now, it'd be abrupt to simply diss off some changes since someone had a change of mind regarding clean up commits.

Florin9doi:
Are you rejecting a patch because it messing the history? I never heard this before. You should squash all the histroy in a single git commit or keep the sources in archives if you're so concerned about messing the histroy.

I read these files by chance and I fixed them by instinct. No, I wont code review a entire module just to persuade you to merge my PR. Take it or leave it.

I don't undestand why you (as a project, not you as a person) are so hostile about everything: closing valid tickets because clutters the bug tracker and rejecting PRs because clutters the git. Isn't my business, but your attitude isn't benefical for nobody, driving away possible contributors.

refractionpcsx2:
Looks good to me, thanks for your efforts

PR #3510, opened 2020-06-29
Fix no portaudio build

xddxd:
PCSX2 with no portaudio was failing to build because of an undeclared variable.


xddxd:
One additional thing I've noticed is that portaudio related settings are still present, even when it's not built. I do not know how to fix that though.

arcum42:
The section you'd want to look at for portaudio showing up when it shouldn't would probably be right here:
https://github.com/PCSX2/pcsx2/blob/master/plugins/spu2-x/src/wx/wxConfig.cpp#L155

in the switch statement for Dialog::Reconfigure, because I'd imagine the number selected would be different.

xddxd:
Pushed, don't know if it's the right way to do it, but I tested it and it works, --no-portaudio build doesn't show portaudio anymore, SDL audio works, on the regular build both SDL and portaudio work. Needs a clean build to see the changes though.

orbea:
Its sad how fast this regressed since https://github.com/PCSX2/pcsx2/pull/3110...

PR #3451, opened 2020-06-08
Work on X86_64 support

beaumanvienna:
* Dedicated pull request for x64 support
 * Contains commits for #3434 "defines for x86 opcodes added", as the same files are involved. Actual changes start on June 07
 * Related to #3308 "Make the interpreter work on x86_64 Unix". 
 * Status: The x64 recompilers work on Windows, macOS and Linux.  Gameplay works fine.

Give it a try, guys! We would like to know if it runs on your 64-bit OS as well. Please checkout this here branch or @tellowkrinkle's JIT64 [branch](https://github.com/tellowkrinkle/pcsx2/tree/JIT64). 
 * [Download artifacts for Windows](https://ci.appveyor.com/project/gregory38/pcsx2/builds/33942032/job/jt83csobgsi09f46)
 * Ubuntu PPA: [Launchpad](https://launchpad.net/~beauman/+archive/ubuntu/marley)
 * [Compile instructions for Fedora](https://github.com/PCSX2/pcsx2/pull/3451#issuecomment-654006105)
 * [Required packages in general](https://github.com/beaumanvienna/pcsx2/blob/x86_64-support/debian-native-packager/pcsx2_1.7.0.1-ce7889939%2Bdfsg1-1ubuntu1.18.04.4.dsc)

We are preparing Tellow's branch to be pulled in. The changes will be split into four PRs. Three out of four have been issued so far: #3512 #3523 #3524

PR #3512, opened 2020-06-30
Fix codegen on x86-64

tellowkrinkle:
Part 1 of the changes being worked on in #3451 

Makes x86emitter emit the x86-64 machine code you would expect it to

Also adds some unit tests to verify that things are working


RedDevilus:
I have done several tests on Windows 10 just doing what a normal end-user would do (First time wizard, boot bios,format memcards, play games and change settings on the fly). It seems to work fine from what i see.

PR #3513, opened 2020-07-01
Resolve issues around playing and creating input recordings under certain scenarios

xTVaser:
These are the major issues that this PR resolves:
- Forbid playing a recording that starts on a save-state if there is no game currently active
- Previously, if a recording was attempted to be played, but failed, the menu options would still be modified as if the recording was successfully played
- Similarly, if the dialog to play a recording was cancelled, the active recording would be cancelled which it shouldn't have been
- Correctly pause/unpause emulation around theses scenarios (especially resuming emulation in the event of aborting playing/creating a recording)

In order to accomplish some of these (especially the last one), modifications had to happen to `InputRecordingFile` and I took this opportunity to do a full refactoring pass / documentation.  Which makes it the last recording related file that needed to be refactored since the initial spike (other files are refactored in other open PRs)


xTVaser:
Superseeded by https://github.com/PCSX2/pcsx2/pull/3669 which is already merged.  Closing

PR #3523, opened 2020-07-05
Allocate memory in an x86-64-compatible way

tellowkrinkle:
Another part of #3451

Note: While this shouldn't change how anything works, it's been the #1 source of breakage of 32-bit builds in #3451 (it was the cause for the failure of win32 to allocate memory and the failure of linux-32 afterward) so we should definitely make sure it gets tested

There are a number of solutions to the problem so I'd like to list them here

The problem: On x86-64, most instructions that reference memory do it with a 32-bit offset from the current instruction pointer.  This means that code in the memory allocated without requesting a specific location probably won't be able access variables in global variables without a large and unwieldy `mov tmpRegister, address` before each usage (which would bloat access to the EE registers horribly)

Possible solutions
1. Move all code allocations to global arrays
	- Pros:
		- Modern debuggers will notice memory accesses are offsets into the static arrays and will annotate them accordingly (e.g. `HostMemoryMap::eeMem + 128` commented after an access to `eeMem->Main[128]`)
	- Cons:
		- On linux with overcommit turned off, will require the full 640mb of reserved memory available to launch (but will release it immediately after)
		- On windows, will be completely unable to decommit memory, meaning that all the memory must be available (even if just in the page file), and the emulator will be unable to tell Windows that it's done using pieces of memory that were allocated for recompilers or the EE.
2. Allocate things by requesting addresses near the executable
	- Cons:
		- Could fail if the system loads the executable in a 4GB sea of small allocations or memory that can't be mmap'd (we need 640mb contiguous space).  This hasn't happened on any of the systems tested (Windows 10 2004 32/64, Ubuntu 18.04 64, macOS 10.12 32/64, macOS 10.15 64) and I doubt it will in the future but you never know
3. Allocate things anywhere, use `mov rax, address; jmp rax` to call functions in the executable, and move all global variables into the allocation
	- Cons:
		- Will dirty extra memory for global constants (minor, there's not many)
		- Debuggers won't notice global variables, e.g. this:
			```asm
			0x15002cf25: mov    dword ptr [rip - 0x100174c7], esi
		    0x15002cf2b: nop    
		    0x15002cf2c: mov    r13d, dword ptr [rip - 0x4da765c3] ; cpuRegs + 496
		    0x15002cf33: nop    
		    0x15002cf34: mov    dword ptr [rip - 0x4da76513], r13d ; cpuRegs + 680
		    0x15002cf3b: mov    dword ptr [rip - 0x4da767b5], 0x80010000 ; cpuRegs + 12
		    ```
		    will turn into this:
		    ```asm
			0x15002cf25: mov    dword ptr [rip - 0x100174c7], esi
		    0x15002cf2b: nop    
		    0x15002cf2c: mov    r13d, dword ptr [rip - 0x4da765c3] ; 0x1025b6970
		    0x15002cf33: nop    
		    0x15002cf34: mov    dword ptr [rip - 0x4da76513], r13d ; 0x1025b6a28
		    0x15002cf3b: mov    dword ptr [rip - 0x4da767b5], 0x80010000 ; 0x1025b6790
		    ```

This PR implements option 2 (my current preference), though I'd like others' thoughts on which they think would be best

### DIFF
--- a/common/include/Utilities/PageFaultSource.h
+++ b/common/include/Utilities/PageFaultSource.h
@@ -25,6 +25,7 @@
 // [TODO] OS-X (Darwin) platforms should use the Mach exception model (not implemented)
 
 #include "EventSource.h"
+#include <atomic>
 
 struct PageFaultInfo
 {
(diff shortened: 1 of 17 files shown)


beaumanvienna:
I wonder if we can/should still use `-fPIC Generate position-independent code (PIC)` if the code position (relative to our allocated memory for tables) is in fact crucial. Why are we using this in the 1st place? If the recompilers depend on the position of the code, shouldn't all modules support position-depended code? Also, position-independent sounds like the compiler is using that extra step to load a 64-bit address. It might affect performance.

tellowkrinkle:
> I wonder if we can/should still use `-fPIC Generate position-independent code (PIC)` if the code position (relative to our allocated memory for tables) is in fact crucial. Why are we using this in the 1st place? If the recompilers depend on the position of the code, shouldn't all modules support position-depended code? Also, position-independent sounds like the compiler is using that extra step to load a 64-bit address. It might affect performance. We should also look into '-no-pie / '-fno-pie'.

We use the position of the code to choose where to put the recompiler allocations, but we don't need it to be in a specific place (which is what would require no-pie) since we generate code based on the runtime locations of things (not the compile-time locations of things).  Thanks to rip-relative addressing, PIC is very cheap on x86-64 (but not on i386) so pretty much every OS has it enabled by default now

PR #3525, opened 2020-07-05
More gtk removals and Onepad dialog tweaks

arcum42:
I reworked the secondary dialogs a bit in onepad, using things from when I redid spu2-x's dialogs.
I converted config.inl to use wx, finishing the null plugins conversion to wx. (And told it to use that for Macs, too. Needs to be checked.)
I also removed a number of places where gtk headers were being unnecessarily included, as well as a header that was no longer being used.


arcum42:
@tellowkrinkle When I redid the null plugin config code, I told it to go ahead and use the wxwidget dialog I put in for Apple as well as unix, whereas before Apple would just get empty function calls. Can you make sure that didn't break things?

I also redid the secondary dialogs in onepad, so it might also be a good idea to go into its configuration and hit the gamepad config and joystick config buttons, to make sure they aren't messed up. They're pretty simple, so hopefully not...

tellowkrinkle:
Do you know what makes the null plugins link with wx in linux?  On macOS I'm getting lots of linker errors due to wx not being included, and I can't see anything in the cmake files that would make it link with wx, e.g. [FWnull has no dependencies](https://github.com/PCSX2/pcsx2/blob/f0d1bb964be00d0981fbbbcf88bb0bcc96e97fd1/plugins/FWnull/CMakeLists.txt#L50-L51)

arcum42:
Judging by onepad, I'm guessing ${wxWidgets_LIBRARIES} should probably be in there, though I don't see it in spu2-x, which seemed to be fine linking to wx...

tellowkrinkle:
TIL `cmake --graphviz=` is cool

Looks like spu2-x gets its dependency transitively through `Utilities_NO_TLS`

tellowkrinkle:
[0001-macOS-compile.patch.gz](https://github.com/PCSX2/pcsx2/files/4942848/0001-macOS-compile.patch.gz)
Okay it looks like the reason it works on Linux is that `.so` files don't actually need their dependencies to link, I guess they just hope that their dependencies will exist in whatever they're getting loaded into

arcum42:
That certainly seems problematic. Well, I'll apply your patch, then probably go ahead and merge this in.

PR #3531, opened 2020-07-09
Add "Screenshot As..." feature to Capture options

xTVaser:
This was a new feature that was dropped from the initial recording tools PR as it was out of scope, adding it back now https://github.com/PCSX2/pcsx2/commit/64104ca9fdeec1b5b19b26e75b5c564ed230613b

Allows you to specify where you want the screenshot to be saved via a file browser. 
image

This also changes GSdx to save screenshots with the `png` extension, as that is actually the compression used - https://github.com/PCSX2/pcsx2/blob/master/plugins/GSdx/GSPng.cpp

### DIFF
--- a/pcsx2/gui/MainMenuClicks.cpp
+++ b/pcsx2/gui/MainMenuClicks.cpp
@@ -924,10 +924,19 @@ void MainEmuFrame::Menu_Capture_Screenshot_Screenshot_As_Click(wxCommandEvent &e
 	if (!CoreThread.IsOpen())
 		return;
 
-	wxFileDialog fileDialog(this, "Select a file", g_Conf->Folders.Snapshots.ToAscii(), wxEmptyString, "PNG files (*.png)|*.png", wxFD_SAVE | wxFD_OVERWRITE_PROMPT);
+	// Ensure emulation is paused so that the correct image is captured
+	bool wasPaused = CoreThread.IsPaused();
+	if (!wasPaused)
+		CoreThread.Pause();
+
+	wxFileDialog fileDialog(this, _("Select a file"), g_Conf->Folders.Snapshots.ToAscii(), wxEmptyString, "PNG files (*.png)|*.png", wxFD_SAVE | wxFD_OVERWRITE_PROMPT);
 
 	if (fileDialog.ShowModal() == wxID_OK)
 		GSmakeSnapshot(fileDialog.GetPath());
+
+	// Resume emulation
+	if (!wasPaused)
+		CoreThread.Resume();
 }
 
 #ifndef DISABLE_RECORDING
--- a/pcsx2/gui/MainMenuClicks.cpp
+++ b/pcsx2/gui/MainMenuClicks.cpp
@@ -919,6 +919,17 @@ void MainEmuFrame::Menu_Capture_Screenshot_Screenshot_Click(wxCommandEvent& even
 	GSmakeSnapshot(g_Conf->Folders.Snapshots.ToAscii());
 }
 
+void MainEmuFrame::Menu_Capture_Screenshot_Screenshot_As_Click(wxCommandEvent &event)
+{
+	if (!CoreThread.IsOpen())
+		return;
+
+	wxFileDialog fileDialog(this, "Select a file", g_Conf->Folders.Snapshots.ToAscii(), wxEmptyString, "PNG files (*.png)|*.png", wxFD_SAVE | wxFD_OVERWRITE_PROMPT);
+
+	if (fileDialog.ShowModal() == wxID_OK)
+		GSmakeSnapshot(fileDialog.GetPath());
+}
+
 #ifndef DISABLE_RECORDING
 void MainEmuFrame::Menu_Recording_New_Click(wxCommandEvent& event)
 {
--- a/plugins/GSdx/GS.cpp
+++ b/plugins/GSdx/GS.cpp
@@ -679,7 +679,9 @@ EXPORT_C_(uint32) GSmakeSnapshot(char* path)
 		if (!s.empty())
 		{
 			// Allows for providing a complete path
-			if (s.substr(s.size() - 4, 4) == ".png")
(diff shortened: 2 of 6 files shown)


pgert:
Wouldn't "Snapshot" be a better word than "Screenshot" since the photo might be of a framed (windowed) display?
The folder for photos is called "snaps" anyway.

Also I've noticed that the photo is in 4:3 even if the display is in widescreen.

xTVaser:
> Wouldn't "Snapshot" be a better word than "Screenshot" since the photo might be of a framed (windowed) display?
The folder for photos is called "snaps" anyway.

The internal naming is Snapshot as well, but the existing naming on the GUI is already `Screenshot`.  I'd be happy to change all the references if more people get behind that idea.

> Also I've noticed that the photo is in 4:3 even if the display is in widescreen.

I can't seem to replicate this problem.  I'm able to save screenshots that are definitely not in a 4:3 aspect ratio.

pcsx2/gui/MainMenuClicks.cpp
```diff
@@ -469,7 +469,7 @@ void MainEmuFrame::Menu_EnableBackupStates_Click( wxCommandEvent& )
 	//  (1st save after the toggle keeps the old pre-toggle value)..
 	//  wonder what that means for all the other menu checkboxes which only use AppSaveSettings... (avih)
 	AppApplySettings();
-    
+
```
lightningterror, line 472: Might just remove the whitespace if you're at it.

pcsx2/gui/MainMenuClicks.cpp
```diff
@@ -806,6 +806,21 @@ void MainEmuFrame::Menu_Capture_Screenshot_Screenshot_Click(wxCommandEvent & eve
 	GSmakeSnapshot(g_Conf->Folders.Snapshots.ToAscii());
 }
 
+void MainEmuFrame::Menu_Capture_Screenshot_Screenshot_As_Click(wxCommandEvent &event)
+{
+	if (!CoreThread.IsOpen())
+	{
+		return;
+	}
```
lightningterror, line 814: Personally I wouldn't add brackets for 1 line code, 2 less lines. There are others similar through the pr.

lightningterror:
Can you squash the formatting changes to the appropriate commits since gsdx and core changes are in their separate commits?

lightningterror:
Aren't snapshots saved as png already? A bit confused.

xTVaser:
> Aren't snapshots saved as png already? A bit confused.
They are stored as a png yes, but the file extension is `.bmp`

sonicfind:
If you could make it so the coreThread is paused when the _Screenshot As_ dialog window is open, that would be a good QoL change. It could take a user some time to choose a good filename or location, so it's possible that whatever they wanted to screenshot already disappeared.
Unlike just _Screenshot_ which is instantaneous.

pcsx2/gui/MainMenuClicks.cpp
```diff
@@ -876,10 +876,27 @@ void MainEmuFrame::Menu_Capture_Screenshot_Screenshot_As_Click(wxCommandEvent &e
...
 
-	wxFileDialog fileDialog(this, "Select a file", g_Conf->Folders.Snapshots.ToAscii(), wxEmptyString, "PNG files (*.png)|*.png", wxFD_SAVE | wxFD_OVERWRITE_PROMPT);
+	// Ensure emulation is paused so that the correct image is captured
+#ifndef DISABLE_RECORDING
+	bool wasPausedViaInputRecTools = g_RecordingControls.IsEmulationAndRecordingPaused();
+	if (!wasPausedViaInputRecTools)
+		CoreThread.Pause();
+#else
+	CoreThread.Pause();
+#endif
```
sonicfind, line 886: It would probably be best to to utilize only CoreThread.IsPaused() in this situation so we can limit input-recording related preprocessor conditions to only code directly related to recording.

xTVaser: Ah that's true, I had that originally but went through a few iterations and forgot.

plugins/GSdx/GS.cpp
```diff
@@ -723,9 +723,13 @@ EXPORT_C_(uint32) GSmakeSnapshot(char* path)
 	{
 		std::string s{path};
 
-		if(!s.empty() && s[s.length() - 1] != DIRECTORY_SEPARATOR)
+		if (!s.empty())
 		{
-			s = s + DIRECTORY_SEPARATOR;
+			// Allows for providing a complete path
+			if (s.substr(s.size() - 4, 4) == ".png")
```
sonicfind, line 729: Oh shoot. Just noticed this. The basic std::string operator== doesn't compare the contents of the each string. It returns whether the addresses of the two strings are the same.
It would be best to go with .compare(".png") == 0. Furthermore, I'd recommend something like I did here to ensure that it accounts for lowercase & uppercase: https://github.com/PCSX2/pcsx2/blob/2170b6c3a41f00f539123ca982f1b7f9325d1c7d/plugins/GSdx/GSCapture.cpp#L427

xTVaser: Hmm, I don't think that's true in C++, see the documentation http://www.cplusplus.com/reference/string/string/operators/ otherwise the normal `Screenshot` option wouldn't work. `==` internally uses `compare`.  But I'll change it to lower-case because why not.  https://repl.it/repls/AnnualTediousGenerics#main.cpp

plugins/GSdx/Renderers/Common/GSRenderer.cpp
```diff
@@ -80,7 +80,7 @@ bool GSRenderer::CreateDevice(GSDevice* dev)
 
 void GSRenderer::ResetDevice()
 {
-    if(m_dev) m_dev->Reset(1, 1);
+	if(m_dev) m_dev->Reset(1, 1);
```
refractionpcsx2, line 83: Spacing changed from tab to space?

refractionpcsx2 (changes requested):

PR #3534, opened 2020-07-10
GSdx-hw: Improve HW dithering

refractionpcsx2:
Gets rid of line artifacts down the screen due to 16bit colour limitations

### DIFF
--- a/plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
+++ b/plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
@@ -1216,10 +1216,10 @@ void GSRendererOGL::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 		GL_DBG("DITHERING mode ENABLED (%d)", m_dithering);
 
 		m_ps_sel.dither = m_dithering;
-		ps_cb.DitherMatrix[0] = GSVector4((int16_t)m_env.dimx[6].w, (int16_t)m_env.dimx[6].z, (int16_t)m_env.dimx[6].y, (int16_t)m_env.dimx[6].x);
-		ps_cb.DitherMatrix[1] = GSVector4((int16_t)m_env.dimx[4].w, (int16_t)m_env.dimx[4].z, (int16_t)m_env.dimx[4].y, (int16_t)m_env.dimx[4].x);
-		ps_cb.DitherMatrix[2] = GSVector4((int16_t)m_env.dimx[2].w, (int16_t)m_env.dimx[2].z, (int16_t)m_env.dimx[2].y, (int16_t)m_env.dimx[2].x);
-		ps_cb.DitherMatrix[3] = GSVector4((int16_t)m_env.dimx[0].w, (int16_t)m_env.dimx[0].z, (int16_t)m_env.dimx[0].y, (int16_t)m_env.dimx[0].x);
+		ps_cb.DitherMatrix[0] = GSVector4(m_env.DIMX.DM00, m_env.DIMX.DM01, m_env.DIMX.DM02, m_env.DIMX.DM03);
+		ps_cb.DitherMatrix[1] = GSVector4(m_env.DIMX.DM10, m_env.DIMX.DM11, m_env.DIMX.DM12, m_env.DIMX.DM13);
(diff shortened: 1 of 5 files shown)


plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
```diff
@@ -1216,10 +1216,10 @@ void GSRendererOGL::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 		GL_DBG("DITHERING mode ENABLED (%d)", m_dithering);
 
 		m_ps_sel.dither = m_dithering;
-		ps_cb.DitherMatrix[0] = GSVector4(m_env.DIMX.DM00, m_env.DIMX.DM01, m_env.DIMX.DM02, m_env.DIMX.DM03);
-		ps_cb.DitherMatrix[1] = GSVector4(m_env.DIMX.DM10, m_env.DIMX.DM11, m_env.DIMX.DM12, m_env.DIMX.DM13);
-		ps_cb.DitherMatrix[2] = GSVector4(m_env.DIMX.DM20, m_env.DIMX.DM21, m_env.DIMX.DM22, m_env.DIMX.DM23);
-		ps_cb.DitherMatrix[3] = GSVector4(m_env.DIMX.DM30, m_env.DIMX.DM31, m_env.DIMX.DM32, m_env.DIMX.DM33);
+		ps_cb.DitherMatrix[0] = GSVector4((int16_t)m_env.dimx[6].w, (int16_t)m_env.dimx[6].z, (int16_t)m_env.dimx[6].y, (int16_t)m_env.dimx[6].x);
```
iMineLink, line 1219: I may be missing something obvious, but why was in old code and is in the new code the loading of the DitherMatrix so different in OGL w.r.t. the DX code?

plugins/GSdx/res/tfx.fx
```diff
@@ -675,14 +675,17 @@ void ps_fbmask(inout float4 C, float2 pos_xy)
...
-	#else
-	int2 fpos = int2(pos_xy / (float)PS_SCALE_FACTOR);
-	#endif
-	C += DitherMatrix[fpos.y&3][fpos.x&3];
-#endif
+	if (PS_DITHER)
+	{
+		int2 fpos;
+
+		if(PS_DITHER == 2)
```
iMineLink, line 682: Minor stuff: missing space after if.

plugins/GSdx/res/tfx.fx
```diff
@@ -769,17 +772,22 @@ PS_OUTPUT ps_main(PS_INPUT input)
...
 	ps_blend(C, alpha_blend, input.p.xy);
 
 	ps_fbmask(C, input.p.xy);
 
+	// When dithering the bottom 3 bits become meaningless and cause lines in the picture
+	// so we need to limit the color depth on dithered items
+	// SW_BLEND already deals with this so no need to do in those cases
+	if (!SW_BLEND && PS_DITHER && PS_DFMT == FMT_16 && !PS_COLCLIP)
+	{
+		C.rgb = clamp(C.rgb, (float3)0.0f, (float3)255.0f);
```
iMineLink, line 787: I see this casting pattern like `(float3)0.0f` is applied thoroughly, but what about `0.0f.rgb`  to achieve the same?

plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
refractionpcsx2: Apparently it's handled differently on dx and ogl, I don't know, you need to ask krossx

plugins/GSdx/res/tfx.fx
refractionpcsx2: And what difference would that make? I just copied the code from elsewhere in the shader so it's consistent, I'm not here to overhaul the code

iMineLink: No difference at all.
Mine was just an arguable consideration about the style.

refractionpcsx2: Fair enough, maybe somebody can address that in the future if they tidy up the shaders, right now I just want the style/code to match the rest of the shader :)

PR #3535, opened 2020-07-11
Update .clang-format to better match upcoming style guide

tellowkrinkle:
People were turning it off because it was doing stupid things, hopefully it will now do less stupid things

I added `BraceWrapping: AfterCaseLabel: true`, which makes this happen:
```cpp
switch (something) {
	case option: // { doesn't go here
	{ // goes here instead

	}
```
Not sure what we prefer, and it wasn't in the style guide, but it seems like we mostly put them on separate lines for everything else

Also set `ReflowComments` to `false` to better match the `ColumnLimit: 0`, though I'm not sure if it does anything with `ColumnLimit: 0` anyways.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -11,7 +11,7 @@ AllowAllParametersOfDeclarationOnNextLine: true
 AllowShortBlocksOnASingleLine: false
 AllowShortCaseLabelsOnASingleLine: false
 AllowShortFunctionsOnASingleLine: All
-AllowShortIfStatementsOnASingleLine: false
+AllowShortIfStatementsOnASingleLine: Never
 AllowShortLoopsOnASingleLine: false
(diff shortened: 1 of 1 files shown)


tellowkrinkle:
Style guide draft here: https://gist.github.com/tadanokojin/7d42ce18a7e836c4a93685e1fedfb13a

.clang-format
```diff
@@ -85,5 +86,5 @@ SpacesInParentheses: false
 SpacesInSquareBrackets: false
 Standard:        Cpp11
 TabWidth:        4
-UseTab:          Never
+UseTab:          ForIndentation
```
lightningterror, line 89: ForContinuationAndIndentation seems better.

PR #3536, opened 2020-07-11
GSdx: Fix SW renderer on win64

tadanokojin:


plugins/GSdx/Renderers/SW/GSDrawScanlineCodeGenerator.x64.avx.cpp
```diff
@@ -78,8 +78,9 @@ void GSDrawScanlineCodeGenerator::Generate_AVX()
 	push(rbp);
 	push(r12);
 	push(r13);
+	push(t2);
 
-	sub(rsp, 8 + 10 * 16);
+	sub(rsp, 200 + 8);
```
iMineLink, line 83: What about a constant for the `200` magic number?

tadanokojin: think I'm gonna make an attempt here to properly allocate/align the stack. I just threw 200+8 as some arbitrary large number so I knew I had extra space.

PR #3546, opened 2020-07-15
Use precompiled headers with Clang

tellowkrinkle:
Full compile of PCSX2 (without plugins) on my laptop: 230s → 85s

@arcum42 When you added PCH you said it wasn't working with Clang, what was breaking for you?


arcum42:
It was throwing up a bunch of things as not being defined that were defined when I was testing it at the time, which was rather bizarre. Compiled with no problems when I tried it with your branch just now, though.

PR #3549, opened 2020-07-15
gsdx-d3d: switch to flip model

tadanokojin:
Enables the flip model for Win8 and Win10

**TODO:**
Need to make it so the GSFrame is destroyed when the renderer changes. Have some local code for this but want to spend some extra time with it. In the meantime avoid switching renderers.

**Enabling screen tearing**: Config -> Emulation Settings -> GS Window -> Aspect Ratio: Fit To Window/Screen
Enter fullscreen normally you should be able to observe tearing.


OCRBonk:
Ok, so testing this after figuring out how to reliably get it to switch to tearing mode. It actually works great.
I can use Scanline Sync or even force Vsync through the driver (The frame limiter needs to be disabled for both to work correctly.)
I was concerned about fit to window for 4:3 content as I don't always like using Widescreen patches. But using a custom 4:3 resolution is a reasonable compromise and it works perfectly then.

Using the same mode built in for OpenGL works as well though there can be performance differences. In Xenosaga I for example, using any kind of Blending Accuracy is needed for shadows to render properly but it absolutely wrecks performance in cutscenes even on an RTX 2080. (Difference on average just from None to Basic can be 30-40+% GPU usage. Less so during gameplay but when using AWGS in battle with Blending Accuracy Basic in OGL causes the same GPU usage jump without any visual differences at all when I took comparison screens.).

Back on topic, the shadows aren't rendered correctly I think no matter what in DX11 (Not that they look bad or anything) so that's not an issue there. So your build seems to work really well to enable tearing and allow the driver to take over Vsync control in Flip Model or use Scanline Sync and get the best frame pacing without major stuttering in Windows 10.  (And introduce the least latency in the process). So I hope this ends up getting merged. Is there a reason why it doesn't work when using a 4:3 aspect ratio in a 16:9 window?

tadanokojin:
Thank you for posting your experience

> So I hope this ends up getting merged. Is there a reason why it doesn't work when using a 4:3 aspect ratio in a 16:9 window?

Yes, it's because of how we handle aspect ratio via the gui instead of the rendering so driver fails to detect it as fullscreen. Waiting for #3588 to be merged before proceeding. Should be sorted after that.

masterotaku:
I hope this can get merged. I tried a build with it a few months ago and it's exactly what I needed to make PCSX2 trigger 3D Vision in DX11 without the need of using an old Special K build in a hacky way.

Also, playing in 2D with G-Sync is a lot better than without it, when you can get some fps drops.

tadanokojin:
Cool, good to hear that it's offering some improvements there.

As far as merging is concerned there are two major problems with this PR:
1. Because GL and D3D11 share a surface in PCSX2, this creates a problem where switching to OpenGL results in a black screen which is obviously not something I can merge into master. I did some research into this and the consensus seems to be that you have to destroy it which makes sense considering I've seen Dolphin code that suggest this is also a problem with VK drivers and as far as I have read sharing a surface between APIs like this is undefined behavior. So the solution to that has to be to destroy and recreate the window whenever the emulation is paused and resumed (ie when settings are opened and then closed).
2. A slightly less annoying problem is the way we are handling aspect ratio via Wx which breaks driver heuristics causing the effects of this PR to only be enabled when the rendering takes up the entire window (ie stretch). #3588 was created to address that problem but the author has since closed it (not sure I know why).

Right now GSdx is getting ready to be merged into core. So I'm going to wait for that as I imagine the logistics of that will be taking much of my time. As I understand it the TAS guys need 1 anyway so that should be likely to be addressed sometime hopefully in the semi-near future. Again, hopefully after the GSdx merge.

JordanTheToaster:
I resized the window after switching from OpenGL to D3D11 and caused this to happen.
wrehwerhntejhntr

tadanokojin:
Rebased and pulled in a commit from #4615 (for testing, I'll rebase it off master later).
Additionally added some experimental tearing support for OpenGL.

Details in updated pr description.

JordanTheToaster:
Works great only problem I could find was if you are full screen in Direct3D11 and switch to OpenGL then double click to go back to windowed it will be frozen on an old frame.

refractionpcsx2:
@JordanTheToaster can you test if that's a result of #4615 or this one please?

JordanTheToaster:
> @JordanTheToaster can you test if that's a result of #4615 or this one please? (since Kojin has temporarily merged the two)

Works correctly on #4615 so it's this PR that causes the issue.

tadanokojin:
> Works great only problem I could find was if you are full screen in Direct3D11 and switch to OpenGL then double click to go back to windowed it will be frozen on an old frame.

I mentioned it in the known issues section.

Immersion95:
I can confirm that AMD works (GL & DX11).

refractionpcsx2:
> I can confirm that AMD works (GL & DX11).

I don't think GL has any code for this yet?

HighBunkerHxH:
> 
> 
> > I can confirm that AMD works (GL & DX11).
> 
> I don't think GL has any code for this yet?

It sorta does, but it's frame pacing is terrible when using G-Sync on OGL

GiladN876:
I would like to test this as well if possible, but I can't download the builds. Can anyone send me a link to the AVX2 build?

JordanTheToaster:
> Is there any reason it was closed ? This is the only way to have lagless D3D11 (I have AMD so OpenGL in Hardware mode is not a great choice for me).

It's being superseded by something coming in the future.

HighBunkerHxH:
> > Is there any reason it was closed ? This is the only way to have lagless D3D11 (I have AMD so OpenGL in Hardware mode is not a great choice for me).
> 
> It's being superseded by something coming in the future.

Superseded by what?
I'm curious?

refractionpcsx2:
You'll find out when it comes, but you will get your VRR eventually.

PR #3550, opened 2020-07-16
Add Disc Swap Support

seta-san:
fldef did a lot of the work


seta-san:
> You reformatted an entire file which makes this pretty much impossible to review.

sorry. i didn't realize that happened. i picked it up after fldef gave up. how would i fix this?

tadanokojin:
Probably best just to delete the commit, reapply your changes, and commit it again but making sure that you don't reformat the file when you save it.

The build is failing because you (or rather fldef) made a function `static` which is also `extern`. This doesn't make any sense.

RedDevilus:
Could be easier to just get the master files and squash and then apply the changes.
Also the branch seems to be gone.
image

fldef:
The fixes were not complete and included various known-breaking changes. I apologize for the confusion.

PR #3553, opened 2020-07-17
GSDX-PCRTC: Delay output circuit dimension transitions

ssakash:
Delay applying changes to the output circuit dimension values regulated by the DISP register by 1 Vsync, this is under the assumption that there's a delay in the original HW whenever the register values are altered in a new frame (seems to be the consensus after inquiring few homebrew devs), this patch helps in emulating such a delay.

This is a more generic version of a older delay I had which was exclusive for the DH, which made no sense as to why it needs to be limited to that bitfield.


ssakash:
Relevant bug report: http://forums.pcsx2.net/Thread-GSDX-PCRTC-Bug-Report-Mizuiro-NTSC-J
Superseding PR #1967

PR #3556, opened 2020-07-18
gui: If available, enable menu option to load backup savestate when a game is first launched

xTVaser:
I noticed this very small issue when working on something bigger than involved savestates.  When PCSX2 is first opened, and a game is launched saveslot `0` is selected, but even if there is a backup for slot 0 the menu option is not enabled (until you perform other savestate actions)

I deleted the comment labeling the function as potentially problematic because after simplifying the function and thinking through the possibilities, it looks fine to me.  The only hairy path is during saving:
- If backups are disabled, we look for an existing `.backup` savestate which shouldn't be touched by the save routine (and we don't expect a new one to be created)
- If backups are enabled, the `.backup` file will have already been created via renaming the original save-state.  The asynchronous process is dumping the _current emulation state_, so by the time we've hit this function the rename has already taken place so we have a guarantee to go off.
  - This solves a potential bug in the existing code (which probably would have never been hit):
    1. Save to a slot that has no current backup
    2. Rename fails
    3. Update method looks for the file without `.backup`, which is still there.  
    4. Load backup option is enabled despite there not actually being a backup to load.
- And for loading, files are not modified or renamed so we have a guarantee around what we are checking.

### DIFF
--- a/pcsx2/gui/Saveslots.cpp
+++ b/pcsx2/gui/Saveslots.cpp
@@ -71,7 +71,6 @@ void States_FreezeCurrentSlot()
 		Console.WriteLn("Load or save action is already pending.");
 		return;
 	}
-	States_updateLoadBackupMenuItem(true);
 
 	GSchangeSaveState(StatesC, SaveStateBase::GetFilename(StatesC).ToUTF8());
 	StateCopy_SaveToSlot(StatesC);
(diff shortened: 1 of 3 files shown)


lightningterror:
Can this be rebased just in case, Red reports that it hard crashes the emu.

RedDevilus:
Brought the changes to master and it seems to work well. You trigger the backup savestate if you overwrite an existing savestate. A future improvement could be making more back-ups  than just a single back-up. Like 5 (as example) per savestate slot like when you overwrite the savestate for the 3rd, 4th or more times. 

Though this improves upon master for current PR.

williamtorres1 (approved):
I approve these changes

PR #3558, opened 2020-07-19
CDVD: Add disc swapping support

seta-san:
may god have mercy on my soul. almost all the credit for this goes to fldef

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -556,7 +556,7 @@ static __fi void cdvdUpdateTrayState()
 	}
 }
 
-static s32 cdvdCtrlTrayClose()
+s32 cdvdCtrlTrayClose()
 {
 	DevCon.WriteLn( Color_Green, L"Close virtual disk tray");
 	cdvd.Ready = CDVD_READY1;
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -41,6 +41,10 @@ static cdvdStruct cdvd;
 
 s64 PSXCLK = 36864000;
 
+#define TRAY_PAUSE_TRIGGER 1
+#define TRAY_SEEK_TRIGGER 4
+#define TRAY_CLOSE_TRIGGER 7
+#define TRAY_MAX_TIMEOUT 10
 
 static __fi void SetResultSize(u8 size)
 {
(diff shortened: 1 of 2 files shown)


lightningterror:
Would be nice if the formatting matches the rest of the code in the files
Example
```
If (something)
{
    code;
}
```

fldef:
This PR is based on my early code. I am still actively developing, documenting, and testing the changes. 

There are problems with this early code that should keep it from being used, for example `MediaChanged` should be reset in SCMD 0x05 (not on read), and the `NODISC` state should never enter `SEEK` status (you'll break Monster Rancher EVO).

@seta-san Please work with me on my fork and let me submit a PR. Thanks.

refractionpcsx2:
Can @fldef and @seta-san work together on this please then, and what are we doing about this PR? I guess I'll leave it here for now as a proof of concept and we can close it later in favour of the finished PR.

I feel some things are still incorrect/missing.  For one it doesn't look to trigger an "Eject" interrupt when the disc swap happens, and the Tray State check (which has been renamed to Media Check, I don't like that as it's not right) is only "Open" for 1 check, it should be open until the Timeout has expired and the disc is "inserted" again.

fldef:
@refractionpcsx2 Thanks for your feedback. I'll address briefly what I can. As mentioned earlier, more work (including documentation) is in process outside of this PR.

Re eject interrupt: No ejecting takes place. The same process PCSX2 uses today to change discs remains: when the tray is opened, close the tray after a certain number of cycles. The difference is in the statuses that are visited after closing. Rather than OPEN -> PAUSE, it's now OPEN -> SEEK -> PAUSE. This is documented in the latest code, which is not currently part of this PR.

Re Tray State vs Media Changed: This was an early name based on my difficulty finding documentation on the flag, and has been changed to `TrayChanged` in the latest code.  The flag represents whether the tray state has changed since a previous test, hence the "changed" in the name. The flag does not represent the current state of the tray -- the CDVD status provides that. Again, this is documented in newer code.

Thank you for your patience.

refractionpcsx2:
I see, fair enough, I misunderstood the tray state.

I know no eject is taking place, from our perspective, but tell me, how do you swap discs in a PS2 without pressing "eject"? That would generate an eject interrupt, so if we're going for accurate emulation of this we should really do that also.

fldef:
> [H]ow do you swap discs in a PS2 without pressing "eject"? That would generate an eject interrupt...

I think I'm misinterpreting "eject" in the context of the code changes, so I'll go over what I know (at the risk of sounding pedantic; sorry for everything obvious)...

Existing PCSX2 functionality:
1. The player starts a game that requires disc swapping. 
2. The game eventually prompts the player for a disc swap.
3. The player opens the CDVD window menu in PCSX2, and changes ISOs.
4. PCSX2 prompts the player if they want to swap discs or reset.
5. The player chooses to swap discs.
6. PCSX2 simulates the swap by changing some internal state and then moving the CDVD status from `OPEN` to `PAUSE`. (see [`cdvdCtrlTrayClose`](https://github.com/PCSX2/pcsx2/blob/6a067f5cd6d30fbb76bfba20bd27c8757940fe0c/pcsx2/CDVD/CDVD.cpp#L531-L542) for the last part.)

For step 6, PS2 hardware normally goes from `OPEN` to `SEEK` to `PAUSE`. SingStar games use disc swapping and explicitly look for the `SEEK` state when testing for a swap. If they don't find it, they won't recognize the swap.

Getting from `OPEN` to `PAUSE` is easy enough. Putting `SEEK` into the sequence (and giving the game time to see it) requires introducing new timing code.

seta-san:
How did I get it to work?

refractionpcsx2:
> > [H]ow do you swap discs in a PS2 without pressing "eject"? That would generate an eject interrupt...
> 
> I think I'm misinterpreting "eject" in the context of the code changes, so I'll go over what I know (at the risk of sounding pedantic; sorry for everything obvious)...
> 
> Existing PCSX2 functionality:
> 
> 1. The player starts a game that requires disc swapping.
> 2. The game eventually prompts the player for a disc swap.
> 3. The player opens the CDVD window menu in PCSX2, and changes ISOs.
> 4. PCSX2 prompts the player if they want to swap discs or reset.
> 5. The player chooses to swap discs.
> 6. PCSX2 simulates the swap by changing some internal state and then moving the CDVD status from `OPEN` to `PAUSE`. (see [`cdvdCtrlTrayClose`](https://github.com/PCSX2/pcsx2/blob/6a067f5cd6d30fbb76bfba20bd27c8757940fe0c/pcsx2/CDVD/CDVD.cpp#L531-L542) for the last part.)
> 
> For step 6, PS2 hardware normally goes from `OPEN` to `SEEK` to `PAUSE`. SingStar games use disc swapping and explicitly look for the `SEEK` state when testing for a swap. If they don't find it, they won't recognize the swap.
> 
> Getting from `OPEN` to `PAUSE` is easy enough. Putting `SEEK` into the sequence (and giving the game time to see it) requires introducing new timing code.

yes I get what WE are doing, but we're supposed to make it look like what a PS2 does, the PS2 doesn't have a CDVD menu to select ISO's from

So on a PS2 the user would press the "eject" button, this would fire an "Eject" interrupt on the CDVD at the same time the tray opens, then the tray state goes in to open. The user puts the disc in, closes the tray and it goes in to the SEEK/PAUSE routine which you mentioned above. 

My point was if we were to do it properly we would also simulate the "eject" procedure too.

refractionpcsx2:
Had a user try this with Dynasty Warriors 5 since you can insert another disc to load some data and they got stuck on this screen with it

seta-san:
Another thing we might need to test are old ps1 games that require disc swapping.

fldef:
The code in this PR is copy-and-pasted -- not forked -- from the work I did. As such I'd like to request that this PR be closed so that my contributions are correctly attributed in a future PR. Thanks.

seta-san:
> The code in this PR is copy-and-pasted -- not forked -- from the work I did. As such I'd like to request that this PR be closed so that my contributions are correctly attributed in a future PR. Thanks.

do you have a later build because i still have an issue where it doesn't always swap.

tadanokojin:
@seta-san , it seems like @fldef is tracking issues over at his branch. Maybe it's an idea to work with him over there rather than confusing the rest of us.

seta-san:
Fldef retired his account. Don't know what's going on now.

refractionpcsx2:
Evidently he didn't and wants to work on it again

seta-san:
@fldef how are you doing?

seta-san:
@fldef are you still alive?

PR #3560, opened 2020-07-21
vs-build: use the win10 sdk

tadanokojin:
Makes the project build with the latest win10 SDK on the system. Uses a workaround for vs2017 to be replaced whenever we drop that build.

Doing this to address build issues in #3549


common/vsprops/WinSDK.props
```diff
@@ -0,0 +1,9 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!-- TODO: vs2017 doesn't support using the lastest installed, 2019 does, this is the workaround -->
```
tadanokojin, line 3: note to self: spelling

PR #3562, opened 2020-07-22
GSdx: Remove GLX in favor of EGL

JibbityJobbity:
As title suggests, removes loading OpenGL using GLX in favor of EGL which is newer and supports Wayland.


orbea:
Is removing it really the best idea? How about just changing the default?

tadanokojin:
I had jib make this PR assuming we might remove it. But it's just for testing right now to see how well things go.

kenshen112:
[https://imgur.com/a/pisjAGQ](Specs I tested this pr with) There's no noticeable change in performance once we figured out the cmake issue jib was having and got gsdx to compile correctly

JibbityJobbity:
I have an Nvidia GTX 980 ti with driver version `450.57`, would be nice to have someone testing AMD on Ubuntu 20.04.

kimberly990:
Works fine for me. AMD RX 570 on Mesa 20.1.3-1, tested on Manjaro KDE 20.0.3

cmake/SearchForStuff.cmake
```diff
@@ -87,6 +87,8 @@ endif()
 ## Use CheckLib package to find module
 include(CheckLib)
 if(Linux)
+    check_lib(EGL EGL EGL/egl.h)
+    check_lib(X11_XCB X11-xcb X11/Xlib-xcb.h)
```
turtleli, line 91: EGL isn't limited to Linux.

PR #3565, opened 2020-07-24
GUI: Add Image Previews when Loading or Saving Savestates

xTVaser:
Given the changes in this PR https://github.com/PCSX2/pcsx2/pull/3531 (minus the new menu options) it's now very trivial to dump a screenshot with a specific file path.  This is an idea I've had for a while that in my opinion might be considered useful - given a whole bunch of savestates it can be easy to forget which one is which before loading or overwriting it.  So when a save-state is saved, we just save a screenshot alongside it, the screenshot can be displayed somewhere in the GUI when hovering over the menu option.

Now the part I'm not totally happy about - positioning of the image popup.  wxWidgets doesn't give a whole lot of flexibility in this area, originally I wanted it to align with the save/load sub-menu but there's no way to get accurate screen coordinates for that.  So I put it in the least intrusive spot that felt best and figured I'd get it up for review because if the feature isn't desired at all, no point wasting time on fighting with wxWidgets any longer.

Demo - https://www.youtube.com/watch?v=QUDr5GUrB0s

### Some areas that probably need more testing
- I've tested a few different screen resolutions but not different DPI settings to see if this looks good yet
  - May be an issue because I've already seen issues where the main-frame is cut off on higher scaling (200%+).  Though for this feature it might not be a big deal because if anything the text menus would be wider, not narrower.  Lower scaling values might be more problematic.


lightningterror:
Can the image fit the entire menu width? It will probably look better that way.

prafullpcsx2:
I was waiting for this feature since a long time.

avih:
Cute. Maybe the menu item could have some hover window which wx positions on its own? or just pop it some fixed offset from the mouse cursor position, e.g. few pixels to the right and bottom (but not lower than the bottom screen edge, and maybe to the left if it's otherwise overflowing the right screen edge), or some such.

xTVaser:
> Can the image fit the entire menu width? It will probably look better that way.
I'm not going to say that it's impossible, but it ties back to the lack of flexibility from wxWidgets.  The closest I found to a workaround was someone hacking around it by calculating font width.  This feels like a hopeful approximation at best.
- https://forums.wxwidgets.org/viewtopic.php?t=29124
- https://forums.wxwidgets.org/viewtopic.php?t=42709

> Cute. Maybe the menu item could have some hover window which wx positions on its own?
Not super familiar with wxWidgets in this area, are you suggesting there is some sort of way to let wx automatically position the windows or something?

> or just pop it some fixed offset from the mouse cursor position
Yup this was my first attempt and it's not difficult, the only issue I would have to resolve is getting the window to follow the mouse if you move around on the same element or else it felt awkward/un-responsive.  If that's the best option then it should just involve some more event wrangling.

arcum42:
Tested, and it is indeed showing pictures properly for my saveslots, and my deleting the pictures and then going to the saveslots didn't crash pcsx2 (though it would've been nice if it would have taken a screenshot on loading a saveslot that didn't have one).

What I actually think would be nice is adding a new savestate window that you could hide and show via a menu item. One side lists the savestates, if you select one, it shows the screenshot on the other side, and double-clicking loads the savestate. Admittedly, this is something I've thought about before, both for savestates and isos...

xTVaser:
> (though it would've been nice if it would have taken a screenshot on loading a saveslot that didn't have one
Good suggestion, though I think this might be a little tricky - there would need to be a way to take the screenshot once a proper full-frame has been drawn _after_ the save-state was loaded.

I like the idea of a savestate/ISO window as well, that would certainly give _alot_ more flexibility UI-wise as well, I've had similar ideas.  Just to be clear, we're talking about another separate frame, much like the console window yeah?  On the topic of listing ISOs, I remember Dolphin's displays icons...wonder how that's accomplished 🤔 

I'd be interested in tackling the separate window idea and abandoning the menu highlighting functionality.

arcum42:
That is a good point. Not absolutely necessary, anyways. You can always just resave the save slot if you want a picture.

And I am thinking very much like the console window, including showing and hiding it through a menu item. If we put being able to save saveslots in the window, too, we might be able to totally ditch the menus for them, even.

I've felt like a lot of the gui could use some work for a while, to be honest.

xTVaser:
Closing in favor of https://github.com/PCSX2/pcsx2/pull/3615

PR #3569, opened 2020-07-26
Refactor a repeatedly used plugin loop.

arcum42:
Nothing too critical. I just noticed the same while loop over and over again, and figured there had to be a way to factor it out.

So I did.

If you want to loop through the plugins, pass the ForPlugins function a lambda with [&] (const PluginInfo * pi), and do everything with the plugin info from pi. If you want to loop through and halt when something is true, use IfPlugins instead, and return a bool.

I think it's more readable than:
const PluginInfo* pi = tbl_PluginInfo; do {
		...
	}  while(++pi, pi->shortname != NULL);

Especially since there's the weirdness that tbl_PluginInfo actually has all the plugins in it, then NULL, to cause the loop to halt, and then the memory card "plugin".

I could use to make sure it doesn't break anything, though, since I did have to make changes to anything that used continue or return in the loop.

### DIFF
--- a/pcsx2/PluginManager.cpp
+++ b/pcsx2/PluginManager.cpp
@@ -1097,12 +1097,12 @@ void SysCorePlugins::Load( const wxString (&folders)[PluginId_Count] )
 	Console.WriteLn(Color_StrongBlue, L"\nLoading plugins from %s...", WX_STR(g_Conf->Folders[FolderId_Plugins].ToString()));
 
 	ConsoleIndentScope indent;
-	const PluginInfo* pi = tbl_PluginInfo; do
-	{
+
+	ForPlugins([&] (const PluginInfo * pi) {
 		Load( pi->id, folders[pi->id] );
 		pxYield( 2 );
+	});
 
-	} while( ++pi, pi->shortname != NULL );
 	indent.LeaveScope();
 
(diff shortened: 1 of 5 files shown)


refractionpcsx2:
This good to go? CK says he tested it and it was fine.

arcum42:
It's probably good to go. Most of the loops are the same, just refactored. There are just a few I had to change a bit, and I wasn't sure if I'd broken anything. Most of what it'd affect is looping through plugins, and we have less and less of those in any case, though...

Hard drive died on my main development computer on Friday, and I put a new one in and reinstalled Linux Sunday, so that killed most of my programming for the weekend. :(

refractionpcsx2:
That sucks :( I hope you didn't lose too much!

arcum42:
Don't worry, I have a small NAS server in my apartment, and most of the stuff I care about is there. There's actually a pretty good chance I could recover things from the hard drive, too, since it failed, fdisk got it working, then it happened again within an hour or two. Haven't bothered yet, though, since most of what I really wanted back on it, I copied straight back off the server and github.

It does mean I had to set up everything again on the computer, though. Ah well, it probably debloated my arch install a fair amount, reinstalling it from scratch.

PR #3576, opened 2020-07-31
GSdx-hw: Make d3d11 not crash on W7.

KrossX:
Changes `CreateDXGIFactory2` to `CreateDXGIFactory1` which is available in Windows7 and is also able to create `IDXGIFactory2` interfaces.

[CreateDXGIFactory2 function](https://docs.microsoft.com/en-us/windows/win32/api/dxgi1_3/nf-dxgi1_3-createdxgifactory2)

> **Remarks**
> This function accepts a flag indicating whether DXGIDebug.dll is loaded. The function otherwise behaves identically to CreateDXGIFactory1.

### DIFF
--- a/plugins/GSdx/Renderers/DX11/GSDevice11.cpp
+++ b/plugins/GSdx/Renderers/DX11/GSDevice11.cpp
@@ -105,7 +105,7 @@ bool GSDevice11::Create(const std::shared_ptr<GSWnd> &wnd)
 
 	// create factory
 	{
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Majority of the team is against the change as it brings no benefit considering Win7 support is dropped officially and this is only a problem for windows 7.  If somebody wants to make a "win7 compatible" fork, that's fine.

okt04175:
Win7 use DXVK can slove this problem.

PR #3579, opened 2020-08-02
GSdx: Merge SetupPrim asm

tellowkrinkle:
SetupPrim only version of #3577 for easier testing/review

Same concerns as #3577

Note: This was made by filtering the commits from #3577, so a lot of the commit messages will be referencing changes that don't exist


plugins/GSdx/Renderers/SW/GSSetupPrimCodeGenerator.all.cpp
```diff
@@ -0,0 +1,572 @@
...
+	} while (0)
+
+#if _M_SSE >= 0x501
+/// On AVX2, uses the given broadcast to load from memory, otherwise uses the load
+# define BROADCAST_OR_LOAD(broadcast, load, dst, src) \
+	broadcast(dst, src)
+# define _rip_local_d(x) _rip_local(d8.x)
+# define _rip_local_d_p(x) _rip_local_d(p.x)
+#else
+/// On AVX2, uses the given broadcast to load from memory, otherwise uses a the load
```
gregory38, line 64: This comment is a bit confusing as the "else" isn't avx2. Maybe 1 comment before the if would be better. or 1 comment with avx2 broadcast and 2nd comment otherwise.

gregory38 (commented):
To be honest it is still readable. And hopefully 32 bits is dropped soon enough.

Just and idea to test your change, generate all/most JIT code and compare them before/after your change. But I guess you could also test random games.

Anyway, complexity bump remains reasonable here. And it likely fixes all OS in 1 shot, so it is a good idea.

plugins/GSdx/Renderers/SW/GSSetupPrimCodeGenerator.all.cpp
tellowkrinkle: The comment was meant to show up in IDEs if you moused over `BROADCAST_OR_LOAD`, no matter which mode your IDE thought you were compiling with at the given time

(So the comment describes `BROADCAST_OR_LOAD` itself, not the specific implementation)

tellowkrinkle:
> Just and idea to test your change, generate all/most JIT code and compare them before/after your change. But I guess you could also test random games.

The problem is that in order to more easily merge i386 and x86-64, I changed a bunch of the temporary register assignments (mostly in x86-64, but some changed in i386 as well), so the codegen isn't identical

plugins/GSdx/Renderers/SW/GSSetupPrimCodeGenerator.all.cpp
gregory38: Oh ok. Not a clever IDE if comment need to be duplicated N times. It isn't important.

```
load128(reg, mem)

**Comment by gregory38** [comment]
```

plugins/GSdx/Renderers/SW/GSSetupPrimCodeGenerator.all.cpp
tellowkrinkle: Previously I had broadcasts of other sizes being used with BROADCAST_OR_LOAD but it seems they're all gone now so yeah I can switch it over

tellowkrinkle: Actually it I somehow missed DrawScanline needs the following pairings:
- vbroadcastf128/movaps
- vbroadcasti128/movdqa
- vbroadcastss/movaps
- vpbroadcastq/movdqa
- vpbroadcastd/movdqa
- vpbroadcastw/movdqa

I made the change to both this and the PR for DrawScanline, though we could do it for just one if you think that would be better

lightningterror (changes requested):
Same as the other one, copyright headers need to match the ones we currently use in GS and the entire project.

lightningterror (approved):

lightningterror:
Merged in #3577

PR #3588, opened 2020-08-10
WIP: Move aspect ratio changes to the GS plugin.

KrossX:
The GS output is handled by the wxWindow by modifying the viewport. This might be in conflict with how drivers try to detect a full screen mode. For example, this makes stretch mode the only way to actually see the tearing effect.

The changes make sure the viewport is always the size of the client window and reports to the plugin. Currently only the aspect ratio is handled, but other features like zoom and black bar removal (?) should be reimplemented as well.


plugins/GSdx/Renderers/Common/GSRenderer.cpp
```diff
@@ -75,6 +75,12 @@ bool GSRenderer::CreateDevice(GSDevice* dev)
 	m_dev = dev;
 	m_dev->SetVSync(m_vsync);
 
+	extern int g_aspect;
```
tadanokojin, line 78: Do we need this extern here? Can we just call SetAspectRatio and SetZoom where this is called.

KrossX: SetAspectZoom gets called before a renderer is ready, so it gets stored in the globals instead (GS.cpp). On render creation (and m_dev is set), it uses the stored global for initial state.

tadanokojin: Right, but create device is also called in gs.cpp during init so I'm confused why you can't just use this pattern.

```cpp
if(!s_gs->CreateDevice(dev))
{
	GSclose();
	return -1;
}

s_gs->SetZoom(g_zoom);
s_gs->SetAspectRatio(g_aspect);
```

KrossX: Yes, it can be moved there and the globals made static again. I'll give it a try.

Immersion95:
I built it as I wanted to test is asap. It works perfectly !

No need to change the aspect ratio with the TV anymore :)

pokeshark:
will this be merged? it helps with alleviating input lag

KrossX:
Nope. This code is garbage.

PR #3157, opened 2019-11-15
PNACH Improvements (Read the Description)

Topaz-Reality:
Using some functions unused by PCSX2 to implement some advanced features that will benefit modders all across the world wide web. Also cleaned up the code some (and will clean up some more)

These functions are not official codebreaker functions. Most of them are made by replacing Master commands (which PCSX2 does not need) to accommodate the needs of modders to make advanced RAM hacking possible without trainer help. These functions were added:

```
8AAAAAAA OOOOOOOO |-> Copies the value at the pointed address of A + O into
0TTTTTTT 00000000 |  address T

9AAAAAAA 0TTTTTTT -> Copies the value stored at address A to address T
 
ADDDDDDD 00000000 |-> Copies the value stored at address D to the pointed
0PPPPPPP 0IIIIIII |  address of P + I 

BXXSSSSS 00000000 -> Activates X amount of lines every S milliseconds.

CAAAAAAA 00000000 |-> Compares the value at address A with the value at
0TTTTTTT YZXXXXXX |  address T. If true, activates X amount of lines.

Value Type (Y): 0 = 32-Bit, 1 = 16-Bit, 2 = 8-Bit
Operations (Z): 00 = Equal, 10 = Not Equal, 20 = Lesser, 30 = Greater
```

Improved the formatting within Patch_Memory.cpp by switching from a mix mash of IF-ELSE and SWITCH-CASE statements to a pure SWITCH-CASE block. IFs are still used in some places, as needed.

PR #3591, opened 2020-08-10
Socket IPC implementation

GovanifY:
This pull request implements a small socket based IPC for pcsx2.

# Why

Among the years [several](https://github.com/GovanifY/Teamod/blob/master/Teamod_2/PCSX2_DMA.cs) [tools](https://www.youtube.com/watch?v=6Ust-45nOWc) [have](https://www.youtube.com/watch?v=snDEzmUciZw) been developed by various romhacking communities to, say, test mods in real time and/or expand on the game features. They all have been implemented by using the equivalent of `ptrace pcsx2` and scanning the memory to find the correct addresses.
Needless to say this was prone to breakage, could differ between versions (and emulators) and was simply a bad idea. If you need to visualize the context think about tools that rewrote the JIT region in realtime.
  
An IPC for a tool->game communication makes sense, but the opposite is also true: due to memory limitations and complexity of implementation some mods have extended the logic of the game outside of the emulator by this mean. One tool I'm thinking of hooked to the file loading routine, while another overlayed images on top of the OpenGL renderer of PCSX2 while communicating with the game.  

Dolphin implemented an IPC for a game<->emulator protocol ( https://dolphin-emu.org/blog/2019/12/06/dolphin-progress-report-november-2019/#50-11186-add-devdolphin-device-for-homebrew-to-communicate-with-dolphin-by-leseratte ) but in this case I believe an OS<->game IPC to be more generally useful to extend logic.

PNACH have also been investigated in the past but because of its one way communication, emu->game only and severe performance hit(each memory patching instruction is executed at each VSync) it has been quickly deemed a bad idea(tm). A previous PR that has been designed in order to extend pnach https://github.com/PCSX2/pcsx2/pull/3157 has, for example, implemented timers in pnach which effectively checks a timer at every vsync to verify if it should execute lines. This is simply hacky and not a good idea..
Lua has also been considered to replicate something like Bizhawk LUA API but ultimately I decided against it to avoid external dependencies, platform bloat should PCSX2 one day decide to have a non wxwidgets/plugin centric codebase, and to avoid bad performance while also allowing more extensibility in the toolset and possible code reuse of the IPC mechanism, which could be reused for an event syscall-based IPC for a game<->emu communication without any external tooling.

Concrete example: Kingdom Hearts 2 Final Mix does not have enough memory to implement streamed music at every point in the game. We would like to run said music without having to hardcode a music loading routine into PCSX2 or ptrace into PCSX2.

# Implementation details

This IPC is event-based, thanks to a socket implementation.  
It has thus two-way communication with virtually no performance hit.  
The BSD implementation uses a unix socket at the path `/tmp/pcsx2.sock` while the windows version uses a local TCP socket on port `28011`. 
A socket based implementation has been preferred over named pipes for its two way communication and possible code reuse in the future if we were to implement a gdbstub.  

The IPC currently implements those commands:
```cpp
        // possible command messages
        enum IPCCommand {
            MsgRead8 = 0,
            MsgRead16 = 1,
            MsgRead32 = 2,
            MsgRead64 = 3,
            MsgWrite8 = 4,
            MsgWrite16 = 5,
            MsgWrite32 = 6,
            MsgWrite64 = 7
        };
```
which are the simplest way to interact with the emulated game in a deterministic fashion. PS2 executables not being PIE allows us to know where a global variable is at any time.  
A menu option has been added, "Enable IPC" in the included picture. The IPC thread automatically starts itself when enabled upon game startup.  
image
Safety of the memRead/Write operations is ensured by verifying the state of the vm before executing those operations.  

The IPC could also theoretically be extended to change in real time emulator specific settings and/or emulation itself, but is not currently implemented. I will probably extend the IPC protocol once merged to allow stuff like this.
A 3 way communication can thus be established, game->OS; OS->game; OS->emu and game->emu.  
game->OS IPC is poll based, OS->game event, game->emu poll and OS->emu event.  
This is due to how the IPC is implemented: Dolphin IPC implements an event based IPC for emu<->game at the cost of having to modify the executable code of the game to implement this.  
The upside of this PR is thus that you do not need to modify the game executable, only to reverse engineer it to find state variables and read off from it. As such this is not a no-cost implementation if the logic requires this.

In a nutshell:
* PR: reads global variable every 0.5seconds to verify if state has changed
* Dolphin: game sends a trap interrupt to the emulator and parse its reply

Of course an event based IPC for game<->emu communication could eventually be implemented but is out of the scope of this PR.  
A client implementation is available [here](https://code.govanify.com/govanify/pcsx2_ipc/).

This PR has been tested to work under windows and linux. MacOS should normally work out of the box since they use BSD sockets too.

The format of IPC message is the following:
```cpp
    //         IPC Message event (1 byte)
    //         |  Memory address (4 byte)
    //         |  |           argument (VLE)
    //         |  |           |
    // format: XX YY YY YY YY ZZ ZZ ZZ ZZ
    //        reply code: 00 = OK, FF = NOT OK
    //        |  return value (VLE)
    //        |  |
    // reply: XX ZZ ZZ ZZ ZZ
```
In the future address could become unecessary but for the memRead/Write operations it is always present.

If you need to generate any doc on the format or any additional information ping me sooner rather than later!

### DIFF
--- a/pcsx2/IPC.cpp
+++ b/pcsx2/IPC.cpp
@@ -73,6 +73,16 @@ SocketIPC::SocketIPC(SysCoreThread* vm)
 	}
 
 #else
+	// XXX: go back whenever we want to have multiple IPC instances with
+	// multiple emulators running and make this a folder
+	char* runtime_dir = std::getenv("XDG_RUNTIME_DIR");
+	// fallback in case macOS or other OSes don't implement the XDG base
+	// spec
+	if (runtime_dir == NULL)
+		m_socket_name = (char*)"/tmp/pcsx2.sock";
+	else
+		m_socket_name = strcat(runtime_dir, "/pcsx2.sock");
+
 	struct sockaddr_un server;
 
 	m_sock = socket(AF_UNIX, SOCK_STREAM, 0);
@@ -82,11 +92,11 @@ SocketIPC::SocketIPC(SysCoreThread* vm)
 		return;
 	}
 	server.sun_family = AF_UNIX;
-	strcpy(server.sun_path, SOCKET_NAME);
+	strcpy(server.sun_path, m_socket_name);
 
 	// we unlink the socket so that when releasing this thread the socket gets
 	// freed even if we didn't close correctly the loop
-	unlink(SOCKET_NAME);
+	unlink(m_socket_name);
 	if (bind(m_sock, (struct sockaddr*)&server, sizeof(struct sockaddr_un)))
 	{
 		Console.WriteLn(Color_Red, "IPC: Error while binding to socket! Shutting down...");
@@ -213,7 +223,7 @@ SocketIPC::~SocketIPC()
 #ifdef _WIN32
 	WSACleanup();
 #else
-	unlink(SOCKET_NAME);
+	unlink(m_socket_name);
 #endif
 	close_portable(m_sock);
 	close_portable(m_msgsock);
--- a/pcsx2/IPC.cpp
+++ b/pcsx2/IPC.cpp
@@ -169,24 +169,24 @@ char* SocketIPC::MakeFailIPC(char* ret_buffer)
 SocketIPC::IPCBuffer SocketIPC::ParseCommand(char* buf, char* ret_buffer)
 {
 	// currently all our instructions require a running VM so we check once
-	// here, will help perf when/if we implement multi-ipc processing in one
-	// socket roundtrip.
+	// here, slightly helps performance
 	if (!m_vm->HasActiveMachine())
 		return IPCBuffer{1, MakeFailIPC(ret_buffer)};
 
 
 	u16 batch = 1;
 	u32 ret_cnt = 1;
+	u32 buf_cnt = 0;
 	if ((IPCCommand)buf[0] == MsgMultiCommand)
 	{
-		batch = FromArray<u16>(buf, 1);
-		buf += 3;
+		batch = FromArray<u16>(&buf[buf_cnt], 1);
+		buf_cnt += 3;
 	}
 
 	for (u16 i = 0; i < batch; i++)
 	{
 		// YY YY YY YY from schema below
-		u32 a = FromArray<u32>(buf, 1);
+		u32 a = FromArray<u32>(&buf[buf_cnt], 1);
 
 		//         IPC Message event (1 byte)
 		//         |  Memory address (4 byte)
@@ -197,82 +197,87 @@ SocketIPC::IPCBuffer SocketIPC::ParseCommand(char* buf, char* ret_buffer)
 		//        |  return value (VLE)
 		//        |  |
 		// reply: XX ZZ ZZ ZZ ZZ
-		//
-		// NB: memory safety checking would be very expansive in our case,
-		// implemented per command and simply a mess. As our threat model is
-		// nonexistant, knowing we can disable the IPC at any time and having
-		// checks client-side it simply makes more sense to not do check
-		// server-side, as bad as this sounds.
-		// Re security threat model: we control the entire emulated memory of
-		// the emulated game, we can DoS our emulator easily by abusing the IPC
-		// features already, and regardless of where you read this there's
-		// probably no sandbox implemented, so simply being able to write to the
-		// game memory code region would make us control the JIT and have probably
-		// full access over the host machine.
-		switch ((IPCCommand)buf[0])
+		switch ((IPCCommand)buf[buf_cnt])
 		{
 			case MsgRead8:
 			{
+				if (!SafetyChecks(buf_cnt, 5, ret_cnt, 1))
+					goto error;
 				u8 res;
 				res = memRead8(a);
 				ToArray(ret_buffer, res, ret_cnt);
 				ret_cnt += 1;
-				buf += 5;
+				buf_cnt += 5;
 				break;
 			}
 			case MsgRead16:
 			{
+				if (!SafetyChecks(buf_cnt, 5, ret_cnt, 2))
+					goto error;
 				u16 res;
 				res = memRead16(a);
 				ToArray(ret_buffer, res, ret_cnt);
 				ret_cnt += 2;
-				buf += 5;
+				buf_cnt += 5;
 				break;
 			}
 			case MsgRead32:
 			{
+				if (!SafetyChecks(buf_cnt, 5, ret_cnt, 4))
+					goto error;
 				u32 res;
 				res = memRead32(a);
 				ToArray(ret_buffer, res, ret_cnt);
 				ret_cnt += 4;
-				buf += 5;
+				buf_cnt += 5;
 				break;
 			}
 			case MsgRead64:
 			{
+				if (!SafetyChecks(buf_cnt, 5, ret_cnt, 8))
+					goto error;
 				u64 res;
 				memRead64(a, &res);
 				ToArray(ret_buffer, res, ret_cnt);
 				ret_cnt += 8;
-				buf += 5;
+				buf_cnt += 5;
 				break;
 			}
 			case MsgWrite8:
 			{
-				memWrite8(a, FromArray<u8>(buf, 5));
-				buf += 6;
+				if (!SafetyChecks(buf_cnt, 6, ret_cnt))
+					goto error;
+				memWrite8(a, FromArray<u8>(&buf[buf_cnt], 5));
+				buf_cnt += 6;
 				break;
 			}
 			case MsgWrite16:
 			{
-				memWrite16(a, FromArray<u16>(buf, 5));
-				buf += 7;
+				if (!SafetyChecks(buf_cnt, 7, ret_cnt))
+					goto error;
+				memWrite16(a, FromArray<u16>(&buf[buf_cnt], 5));
+				buf_cnt += 7;
 				break;
 			}
 			case MsgWrite32:
 			{
-				memWrite32(a, FromArray<u32>(buf, 5));
-				buf += 9;
+				if (!SafetyChecks(buf_cnt, 9, ret_cnt))
+					goto error;
+				memWrite32(a, FromArray<u32>(&buf[buf_cnt], 5));
+				buf_cnt += 9;
 				break;
 			}
 			case MsgWrite64:
 			{
-				memWrite64(a, FromArray<u64>(buf, 5));
-				buf += 13;
+				if (!SafetyChecks(buf_cnt, 13, ret_cnt))
+					goto error;
+				memWrite64(a, FromArray<u64>(&buf[buf_cnt], 5));
+				buf_cnt += 13;
 				break;
 			}
 			default:
 			{
+			error:
 				return IPCBuffer{1, MakeFailIPC(ret_buffer)};
 			}
 		}
--- a/pcsx2/IPC.cpp
+++ b/pcsx2/IPC.cpp
@@ -141,21 +141,21 @@ void SocketIPC::ExecuteTaskInThread()
 
 #ifdef _WIN32
 			// socket timeout
-			DWORD timeout = 10 * 1000; // 10 seconds
-			setsockopt(m_msgsock, SOL_SOCKET, SO_RCVTIMEO, (const char*)&timeout, sizeof timeout);
+			DWORD tv = 10 * 1000; // 10 seconds
 #else
 			// socket timeout
 			struct timeval tv;
 			tv.tv_sec = 10;
 			tv.tv_usec = 0;
-			setsockopt(m_msgsock, SOL_SOCKET, SO_RCVTIMEO, (const char*)&tv, sizeof tv);
 #endif
+			setsockopt(m_msgsock, SOL_SOCKET, SO_RCVTIMEO, (const char*)&tv, sizeof tv);
+			setsockopt(m_msgsock, SOL_SOCKET, SO_SNDTIMEO, (const char*)&tv, sizeof tv);
 
 
 			// either int or ssize_t depending on the platform, so we have to
 			// use a bunch of auto
 			auto receive_length = read_portable(m_msgsock, m_ipc_buffer, MAX_IPC_SIZE);
-			if (receive_length >= 0)
+			if (receive_length > 0)
 			{
 				// if we already received at least the length then we read it
 				auto end_length = 4;
@@ -227,15 +227,17 @@ SocketIPC::~SocketIPC()
 	DESTRUCTOR_CATCHALL
 }
 
-char* SocketIPC::MakeOkIPC(char* ret_buffer)
+char* SocketIPC::MakeOkIPC(char* ret_buffer, uint32_t size = 5)
 {
-	ret_buffer[0] = IPC_OK;
+	ToArray<uint32_t>(ret_buffer, size, 0);
+	ret_buffer[4] = IPC_OK;
 	return ret_buffer;
 }
 
-char* SocketIPC::MakeFailIPC(char* ret_buffer)
+char* SocketIPC::MakeFailIPC(char* ret_buffer, uint32_t size = 5)
 {
-	ret_buffer[0] = IPC_FAIL;
+	ToArray<uint32_t>(ret_buffer, size, 0);
+	ret_buffer[4] = IPC_FAIL;
 	return ret_buffer;
 }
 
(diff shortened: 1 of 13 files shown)


pcsx2/System/SysThreads.h
```diff
@@ -170,6 +171,16 @@ class SysCoreThread : public SysThreadBase
...
 
+	// Stores the state of the socket IPC thread.
+	std::unique_ptr<SocketIPC> m_socketIpc;
+
+    // Current state of the IPC thread
+    enum StateIPC {
+        OFF,
+        ON
+    };
+    StateIPC m_IpcState = OFF;
```
lightningterror, line 182: Use tabs instead of spaces, check the rest of the code too.

lightningterror:
You can run clang format on the 2 new files to fix the formatting.

GovanifY:
Just ran it, completely forgot there was one!

pcsx2/IPC.cpp
```diff
@@ -0,0 +1,304 @@
...
+			rval = std::make_pair(3, ToArray(MakeOkIPC(3), res, 1));
+			break;
+		}
+		case MsgRead32:
+		{
+			u32 res;
+			if (m_vm->HasActiveMachine() == true)
+			{
+				res = memRead32(a);
+			}
```
lightningterror, line 222: Don't really need brackets for 1 code lines.

GovanifY:
Just fixed styling issues on other files, all my changes should fit the clang-format spec now

pcsx2/IPC.cpp
GovanifY: Just fixed it for IPC{.cpp,.h}

lightningterror: I meant remove the brackets altogether, otherwise it was fine how clang formatted.
```
			if (m_vm->HasActiveMachine() == true)
				res = memRead32(a);
```

GovanifY: Ah! I don't usually use bracketless if because of bugs they can cause, let me fix that

pcsx2/IPC.cpp
```diff
@@ -0,0 +1,272 @@
...
+				res = memRead16(a);
+			else 
+				goto error;
+			rval = std::make_pair(3, ToArray(MakeOkIPC(3), res, 1));
+			break;
+		}
+		case MsgRead32:
+		{
+			u32 res;
+			if (m_vm->HasActiveMachine() == true)
```
lightningterror, line 211: You can drop the true checks and just do `if (m_vm->HasActiveMachine())`

GovanifY: done

lightningterror:
Can you run clang on IPC.h again?

GovanifY:
Done, also switched to bracketless for in the template expressions

GovanifY:
I just made a fairly complete Doxygen based documentation of the entire IPC API [in the client-side repository](https://code.govanify.com/govanify/pcsx2_ipc/).
It contains the binary format of all IPC commands and a more general documentation of the IPC and the reasonning behind some of the implementation choices.

GovanifY:
I implemented a batch processing mode, which will be the last main change I'll do to this PR before merging. When sending a lot of recurring IPC messages, due to how the socket stack works, it is preferred to send them in batch all at once, this leads to a speedup of at least 1000x in those situations, at the detriment of having a little less of handholding when it comes to the client-side API.   
I tested it and I can process 1000 Write<uint8_t> in less than 100µs.   

I'll abstain from touching this PR from now on, and I'll let you review it.

pcsx2/IPC.h
```diff
@@ -54,6 +54,11 @@ class SocketIPC : public pxThread
      */
 #define MAX_IPC_RETURN_SIZE 450000
 
+	/**
+     * Maximum number of socket errors in a row before the thread stops.
```
lightningterror, line 58: Why multiple line code comments for 1 line?

GovanifY: That was because they were originally doxygen-formatted comments, taken from the client implementation. 
Just fixed the styling to make it use // comments for 1 line comments.

GovanifY:
Hey, is there any news regarding the review of this PR? We haven't been discussing anything else than code style so far so I wonder how it's coming along. Can I do anything to make it faster?

GovanifY:
The issue linked above shows that the `ptrace pcsx2` technique has indeed been broken as of the master branch.
As such I believe this PR should get merged for the next release to at least provide a stable alternative.

Xeeynamo:
Since this pull request have been mentioned in #3638 as "the solution" but it has not even been merged or approved, I am going to address some of my concerns since I would eventually need to use this IPC.

1. Why always using `28011` as a port? This should definitively be configurable by the UI. If it happens to have multiple instances of PCSX2 running, the second instance will fail. You should have constantly the console window open to at least have the feedback if the IPC socket initialisation has been performed successfully.

2. Why shutting down the thread, without any feedback, after 600 errors? The choice does not make any sense and the arbitrary number you decided does not either.

3. Do you realise how ungracefully you are shutting down the running thread? You are setting `m_end` to `true` on the destructor and forcefully closing all the socket resources. At that point the socket might be ready to listen or to write back to the client, failing miserably. You know you can just use a semaphore to make it thread safe? 

4. If you disable IPC from the menu, it does not close the connection. I am not sure what is the logic there, but it looks like you are just starting new threads over and over.

5. This is my main issue with the current implementation. The documentation you provided is bland and incomplete and I can not figure out the basics based on the explanation on the top. Not even all the commands are mentioned (eg. `MsgMultiCommand`). Every decent API over the NET should at least provide the length of the expected request and response, or both client or server  might risk to end up to be stuck or even mis-aligning the next command. This implementation is not future proof whatsoever as it relies to fixed length on the content pipe you're arbitrarily de/serialising.
You are also mentioning that you performed some kind of benchmark: can we run this load test and see the result ourselves? Also I might suggest to read and write memory in chunk, defining the length of the byte array rather than just relying on the magical 8, 16, 32 and 64 bit values, also because you're just assuming that every thing will be little-endian.

6. Last, but not least, if you mention DLL to inject to existing applications to use THIS implementation (you referred to a certain `AppInit_DLL` on #3638 ), you might want to ship it with this pull request. or at least provide samples and bindings to your IPC interpretation.

By just briefly looking at your PR, it's clear that it is not ready to be merged into `master` yet.

GovanifY:
@Xeeynamo 
1. The port number was picked by looking at unused ports, since this protocol is going to be used by other emulators in the long run. Having multiple emulators opened is, currently, not supported for the IPC scripting but support can be easily added.

2. What happens when you don't have the permissions to write to the thread? Or the remote process sends too much too fast? We don't want to stall the emulator and let the user figure it out.

3&4. That's the idea, sherlock. It is hooked onto the VM when we start the game and, since some connection might have started, even if we disable it, we do not want to kill it until pcsx2 restart. The thread is ungracefully killed because it is on PCSX2 exist, and we actually handle this case in the API, go read the coverage report.

5. On linux, write 10.000 Write64 commands in a batch command and send them over to PCSX2. Any more than that and TLB misses happens. It is the speed of linux socket memory proceeding which you can check with `socat`. All commands are mentioned if you read the documentation correctly :), what are batch commands? Also it IS future proof, it relies on VLE to avoid sending unnecessary data and actually having something fast. The .NET API is C bindings, and they're fairly documented. If you have an actual concern about it please voice it without being angry toward this project.

6. This is going to be added, of course, but I hope you did not expect me to realistically respond to your childish endeavors in a few hours

By briefly looking at your comment, it is clear you are not suited for constructive criticisms yet.
Also any more rampage on this PR(thumbs down, etc) and it will be locked to contributors.

FerocityVine:
I agree with @Xeeynamo. This IPC implementation is bad and largely insecure.

1. The port is 28011. It cannot be altered. And that port just happens to be the port to my Minecraft Development Server, which is forwarded. Also I know people who use ports between 28000 and 28050, because of the convenience and emptiness of the ports. Which means the ports are not empty at all in some cases.

If the port is forwarded to the internet, someone with decent knowledge could take advantage of it over the IPC. The best case scenario is that some programs will collice, which is a minor annoyance. The worst case is that core files can be altered via the port through the IPC if File IO is ever implemented. And given how important File IO can be, Yikes.

Even if the port is not forwarded or used, some people use 2-10 instances of PCSX2 at once to test many things. Like, why cripple it like this?

2. IPC does not allow for specific programs to use it, like Cheat Engine and EmuHaste. For these programs, a static address is extremely necessary. I am only stating this because I came here from #3638 and I am livid to see this being passed off as a solution. I am using CE on a daily basis for reversing, and now I wake up to see all of my cheat tables have been rendered useless. Where is the fun in that?

3. What is the true use for IPC? I have seen #3157, and I think that it is more than extremely sufficient for everything one may want to do. If not, something like Internal Python can easily be implemented. I'd rather test my code real-time by just writing some text than to go through the hassle of downloading an IDE, writing and testing code, writing an IO Protocol to support IPC, compile, test, compile, test and repeat. Also the port may need to be altered for each executable if more than one PCSX2 instance is used. More hassle down the road.

When things like PNACH and Internal Python exist, an IPC is largely inconvenient because of the sheer amount of work it takes for one to make something to work with it. Why work with the hassle of IDE and compilations when I can just write and have it work? I cannot say anything for Internal Python, but PCSX2 Developers made the processing of PNACH to be psuedo-compilation, and in my tests, IPC is faster than PNACH. But it is such an insignificant difference that it is not worth it.

Say it with me as Technology Connections have said: "Most people will happily sacrifice minor performance benefits, if the experience after doing so is more convenient and is simpler."

Also considering a fixed x64 build may be around the corner, PNACH can be further improved so that it is faster than IPC anyway.

I am preplexed, and confused. Valve, plz fix.

kenshen112:
Homie. We're not going to change the addresses back in the pnach system

refractionpcsx2:
I'm sure TellowKrinkle will be along later to explain the reason for the change and if it was completely necessary or not, however by the fact he did it, tells me it was required for 64bit to work, which has been requested by a lot of users for many years.

If this is the case, then I'm sorry but either the tools will need to be updated or you will have to write new tools to deal with it, it isn't our responsibility to keep compatibility with 3rd party tools, that's up to the 3rd party tool developers to do, so if you need to complain to somebody, it should be them, don't shout down perfectly valid tools/PR's because it doesn't do what you currently do, this isn't the way to move forward.

sirocyl:
Is there a reason why this PR in particular is being brigaded against? It adds a feature at no cost to the emulator, code quality, or stability.

Contributors don't grow on trees, and if this is the experience and behavior you want novice PCSX2 developers to expect before joining the project, I guarantee stagnant progress, if any.

FerocityVine:
Oh, I do not know @sirocyl, maybe because of:

- Extremely bloaty code
- Crippled functionality
- Insecure code
- Unnecessary complication

and so much more.

tellowkrinkle:
> Is there a reason why this PR in particular is being brigaded against?

It's because I broke a bunch of people's stuff in the 64-bit update and mentioned this PR as a possible solution
Sorry @GovanifY for any trouble I may have caused by doing that

Reminder: If you want to complain about how this is not a solution to problems caused by the 64-bit update, do that on a thread about the 64-bit update, not this one.

There also appear to be some misunderstandings about what is and isn't broken by this patch and the 64-bit patch.

Things broken by 64-bit:
- Cheat Engine
- Things using `ptrace`/`ReadProcessMemory`
- Anything else that reads PCSX2's memory directly and assumes things will be at specific addresses

Things NOT broken by 64-bit:
- PNACH
- Anything else where PCSX2 is in control of the memory transfer (if it is broken, that would be considered a bug, so please report it using a new issue)

Things broken by this PR:
- Nothing

In addition, the PR only listens to localhost (see IPC.cpp:66), so random people on the internet would not be able to control the emulator even if you had the port open to the internet

> What is the true use for IPC?

PNACH will never support communication between outside programs and the emulator, which is what all the programs using `ptrace`/`ReadProcessMemory` needed.  e.g. you would never be able to dump KH2FM animations with PNACH

pcsx2/IPC.cpp
```diff
@@ -0,0 +1,315 @@
...
+	if ((IPCCommand)buf[0] == MsgMultiCommand)
+	{
+		batch = FromArray<u16>(&buf[buf_cnt], 1);
+		buf_cnt += 3;
+	}
+
+	for (u16 i = 0; i < batch; i++)
+	{
+		// YY YY YY YY from schema below
+		u32 a = FromArray<u32>(&buf[buf_cnt], 1);
```
tellowkrinkle, line 218: I think you need a bounds check here

pcsx2/IPC.cpp
```diff
@@ -0,0 +1,315 @@
...
+		if (msgsock == -1)
+		{
+			if ((m_error_count > MAX_SOCKET_ERRORS) || m_end)
+				break;
+			else
+				m_error_count++;
+		}
+		else
+		{
+			if (read_portable(msgsock, m_ipc_buffer, MAX_IPC_SIZE) < 0)
```
tellowkrinkle, line 130: Is there any guarantee in unix sockets / tcp that says that the message won't be split into multiple pieces?  It seems at least somewhat likely for things from localhost, but I'd rather not have things mysteriously fail because we didn't take this into account

pcsx2/IPC.cpp
```diff
@@ -142,6 +142,7 @@ void SocketIPC::ExecuteTaskInThread()
 			if (!(errno == ECONNABORTED || errno == EINTR || errno == EAGAIN || errno == EWOULDBLOCK))
 			{
 #endif
+				fprintf(stderr, "IPC: An unrecoverable error happened! Shutting down...");
```
tellowkrinkle, line 145:
```suggestion
				fprintf(stderr, "IPC: An unrecoverable error happened! Shutting down...\n");
```

pcsx2/IPC.cpp
```diff
@@ -0,0 +1,375 @@
...
+						end_length = FromArray<u32>(m_ipc_buffer, 0);
+
+					receive_length += tmp_length;
+				}
+
+				if (receive_length == 0)
+					break;
+
+				// we'd like to avoid a client trying to do OOB
+				if (receive_length > MAX_IPC_SIZE)
```
tellowkrinkle, line 205: If it's like this here, that means the OOB already happened in the `read`

pcsx2/IPC.cpp
```diff
@@ -0,0 +1,378 @@
...
+			tv.tv_usec = 0;
+#endif
+			setsockopt(m_msgsock, SOL_SOCKET, SO_RCVTIMEO, (const char*)&tv, sizeof tv);
+			setsockopt(m_msgsock, SOL_SOCKET, SO_SNDTIMEO, (const char*)&tv, sizeof tv);
+
+
+			// either int or ssize_t depending on the platform, so we have to
+			// use a bunch of auto
+			auto receive_length = read_portable(m_msgsock, m_ipc_buffer, MAX_IPC_SIZE);
+			if (receive_length > 0)
```
tellowkrinkle, line 172: I think you can avoid some of initialization by just starting with `receive_length = 0` and `end_length = 4`, then you can skip the check here and just let everything be read and checked by the while loop

pcsx2/IPC.cpp
```diff
@@ -0,0 +1,378 @@
...
+					}
+
+					receive_length += tmp_length;
+
+					// if we got at least the final size then update
+					if (end_length == 4 && receive_length >= 4)
+					{
+						end_length = FromArray<u32>(m_ipc_buffer, 0);
+						// we'd like to avoid a client trying to do OOB
+						if (end_length > MAX_IPC_SIZE)
```
tellowkrinkle, line 203: I think we should respond with an error if they make a request that's too large instead of just ignoring part of it

Otherwise it'll be confusing for the client to debug

RedDevilus (approved):

GovanifY:
Squashed and rebased to master for final review.

pcsx2/System/SysCoreThread.cpp
```diff
@@ -239,6 +240,13 @@ void SysCoreThread::GameStartingInThread()
 #ifdef USE_SAVESLOT_UI_UPDATES
 	UI_UpdateSysControls();
 #endif
+	if(EmuConfig.EnableIPC && m_IpcState == OFF)
```
lightningterror, line 243: Small nitpick, add space after if.

pcsx2/IPC.cpp
```diff
@@ -0,0 +1,371 @@
...
+		//        |  return value (VLE)
+		//        |  |
+		// reply: XX ZZ ZZ ZZ ZZ
+		switch ((IPCCommand)buf[buf_cnt])
+		{
+			case MsgRead8:
+			{
+				if (!SafetyChecks(buf_cnt, 5, ret_cnt, 1, buf_size))
+					goto error;
+				u8 res;
```
lightningterror, line 291: Maybe `const u8 res = memRead8(a);` instead?
Same for the others.

GovanifY: Done!

GovanifY:
Rebased on master

PR #3596, opened 2020-08-14
GameDB: Beatmania Da! Da1 Da!

seta-san:
Skip Konami keyboard check.


seta-san:
closing since we're getting an actual emulation of the keyboard so we don't need a hack.

PR #3518, opened 2020-07-02
Merge CDVD plugin into core

MonJamp:
~~Not ready to merge yet.~~ Closes #3515 

Things that have to be done:
- [x] Merge linux/unix code
~~Fix memory leak~~ Edit: Looks like the leak existed before my commits
- [x] Remove cdvdGigaherz plugin and CDVD plugin interface
- [x] Double check CMakeLists
- [x] Test `usecd` command line option
- [x] Resolve false error `* CDVD: Error opening source`
- [ ] Thorough testing

Currently I'm testing this by mounting ISOs. I don't have a disk drive on my PC. It'd be nice if someone could test this using a disk drive.

PR #3597, opened 2020-08-14
pcsx2-gui: Some menu changes

arcum42:
This pretty much rearranges the menus a bit in a way that makes sense to me.

Descriptions of the changes follow, though you'd probably get a better idea from compiling and running it.

The System menu is now the PCSX2 menu, and has been substantially rearranged. At the top, we have Boot ISO, Pause/Resume, Shutdown, and Run ELF. Fast Boot is now a checkbox, followed by "Create Blockdump", and then a Game Settings submenu with the other settings that used to be directly in the menu. then you have the normal save state menus, the backup save item, and exit, as normal.

CDVD, I didn't touch, as I didn't want to interfere with the CDVD plugin merge.

Config is mostly the same, except that Change Language is now in there.

Misc and Debug are gone.

Next we have the Window Menu, with "Show Debug", "Show Console", and "Console to Stdio". I'm likely to add more items here, like a savestate window, an iso library window, etc..., later.

Then Capture and Recording are the same.

And there's a new Help menu, which is mostly menu items that go to websites. The Getting Started page, the Compatibility page, the forums, the main website, the wiki, the github page, and, naturally, about.

Also, the status bar shows if fast boot is on, the current iso name, and if it's a 32 or 64 bit build.

I'm curious what people think, and if there are other changes I should make here...

### DIFF
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -53,6 +53,26 @@ wxMenu* MainEmuFrame::MakeStatesSubMenu( int baseid, int loadBackupId ) const
 	return mnuSubstates;
 }
 
+void MainEmuFrame::UpdateStatusBar()
+{
+	wxString temp(wxEmptyString);
+
+	if (g_Conf->EnableFastBoot)
+		temp += "Fast Boot - ";
+
+	if (g_Conf->CdvdSource == CDVD_SourceType::Iso)
+		temp += "Selected: '" + wxFileName(g_Conf->CurrentIso).GetFullName() +"' ";
+
+#ifdef __M_X86_64
+	temp+= "(64 bit) ";
+#else
+	temp += "(32 bit) ";
+#endif
+
+	m_statusbar.SetStatusText(temp, 0);
+	m_statusbar.SetStatusText(CDVD_SourceLabels[enum_cast(g_Conf->CdvdSource)], 1);
+}
+
 void MainEmuFrame::UpdateIsoSrcSelection()
 {
 	MenuIdentifiers cdsrc = MenuId_Src_Iso;
@@ -66,12 +86,8 @@ void MainEmuFrame::UpdateIsoSrcSelection()
 		jNO_DEFAULT
 	}
 	sMenuBar.Check( cdsrc, true );
-	m_statusbar.SetStatusText( CDVD_SourceLabels[enum_cast(g_Conf->CdvdSource)], 1 );
-
+	UpdateStatusBar();
 	EnableCdvdPluginSubmenu( cdsrc == MenuId_Src_Plugin );
-
-	//sMenuBar.SetLabel( MenuId_Src_Iso, wxsFormat( L"%s -> %s", _("Iso"),
-	//	exists ? Path::GetFilename(g_Conf->CurrentIso).c_str() : _("Empty") ) );
 }
 
 bool MainEmuFrame::Destroy()
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -418,17 +418,18 @@ void MainEmuFrame::CreateConfigMenu()
 
 	m_menuConfig.AppendSeparator();
 
-	m_menuConfig.Append( MenuId_ChangeLang,			L"Change &Language" ); // Always in English
+	m_menuConfig.Append(MenuId_ChangeLang,			L"Change &Language" ); // Always in English
 	m_menuConfig.Append(MenuId_Config_ResetAll,	_("C&lear all settings..."),
 		AddAppName(_("Clears all %s settings and re-runs the startup wizard.")));
 }
 
 void MainEmuFrame::CreateWindowsMenu()
 {
-	m_menuWindow.Append(MenuId_Debug_Open,		_("&Open Debug Window..."),	wxEmptyString);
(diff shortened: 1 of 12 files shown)


arcum42:
menus-screenshot

arcum42:
help
window

pcsx2/gui/MainFrame.cpp
```diff
@@ -322,6 +338,147 @@ static int GetPluginMenuId_Name( PluginsEnum_t pid )
...
+
+void MainEmuFrame::CreateHelpMenu()
+{
+	m_menuHelp.Append(MenuId_Help_GetStarted,				_("&Getting Started") );
+	m_menuHelp.Append(MenuId_Help_Compatibility,			_("&Compatibility") );
+	m_menuHelp.AppendSeparator();
+	m_menuHelp.Append(MenuId_Help_Forums,					_("&Pcsx2 Forums") );
+	m_menuHelp.Append(MenuId_Help_Website,					_("&Pcsx2 Website") );
+	m_menuHelp.Append(MenuId_Help_Wiki,						_("&Pcsx2 Wiki") );
+	m_menuHelp.Append(MenuId_Help_Github,					_("&Pcsx2 Github Project") );
```
lightningterror, line 477: Dunno about having them duplicated both here and the about page. Also you can drop the pcsx2 part on the menus.

arcum42: I'm not sure what I'd replace them with in the About window, though, and the License link isn't in the menu, currently. The About is already pretty bare. I figured I'd just leave it as is for now.

I went ahead and dropped "Pcsx2" from the names, and rearranged them slightly.

pcsx2/gui/MainFrame.cpp
```diff
@@ -53,6 +53,26 @@ wxMenu* MainEmuFrame::MakeStatesSubMenu( int baseid, int loadBackupId ) const
...
+		temp += "Fast Boot - ";
+
+	if (g_Conf->CdvdSource == CDVD_SourceType::Iso)
+		temp += "Selected: '" + wxFileName(g_Conf->CurrentIso).GetFullName() +"' ";
+
+#ifdef __M_X86_64
+	temp+= "(64 bit) ";
+#else
+	temp += "(32 bit) ";
+#endif
```
lightningterror, line 70: I think it would be better for 32/64bit info to be moved in it's own bar/column like CDVD selected boot type. Maybe move it at the end and also shrink the cdvd one.

Currently if the file name is too big build info gets cut.

lightningterror:
If you enable recording tools and then click Cancel on the popup it will crash the emulator.

sonicfind:
It crashes on cancel because in MainMenuClick's function **Menu_EnableRecordingTools_Click**, it's still searching for **MenuId_EnableRecordingTools** in the **m_menuSys** menu instead of **m_GameSettingsSubmenu** - where it was moved to.

arcum42:
> It crashes on cancel because in MainMenuClick's function **Menu_EnableRecordingTools_Click**, it's still searching for **MenuId_EnableRecordingTools** in the **m_menuSys** menu instead of **m_GameSettingsSubmenu** - where it was moved to.

Taken care of. Thanks!

pcsx2/gui/MainFrame.cpp
arcum42: You might see how it looks now. Getting it not to cut off "NoDisc" when resizing those is a bit tricky, too.

RedDevilus:
The only obvious thing we are missing is what other emulators use:
- A list of titles where u can double click on to play a game instead of browsing to directory and finding your specific game.
DobieStation is simple but effective:
image
Doesn't need to be this fancy or add a thumbnail to the left and be done with it. Maybe worth looking into.

- Pergame configuration (i leave this out because we have spectabis and other solutions.)

arcum42:
> The only obvious thing we are missing is what other emulators use:
> 
> * A list of titles where u can double click on to play a game instead of browsing to directory and finding your specific game.
>   DobieStation is simple but effective:
>   image
>   Doesn't need to be this fancy or add a thumbnail to the left and be done with it. Maybe worth looking into.
> * Pergame configuration (i leave this out because we have spectabis and other solutions.)

Those are things I'm interested in, and part of why I went with a Windows menu was to have an obvious place for new windows to go. I was thinking about adding an iso window, but decided it was going to be complicated enough I'd rather get this in first. (We have to store all of the information for the isos and have it be persistent, after all, and I don't just want the filename. I'd like the game name, CRC, cover image, maybe descriptions. Heck, category would be nice...)

Also, if you look at pr #3615 , while that is a savestate window, you can see in his description that he plans later to add an iso tab. So I'm assuming one of us will get there.

Per game configuration would be nice, but yeah, is a separate pr in itself, especially since we'd really want it to cover not just pcsx2 settings, but GSdx settings, and maybe spu2-x settings. It might also help if it was done after the iso window is there, since I'm sure we'll need to store metadata for the games, and we can add the settings to that.

Changing up the cdvd menu would be nice, too, but I want #3518 in first.

I will undoubtedly make more changes. I mostly want this in as a start at it.

RedDevilus:
Also doing the GUI #3457. But that should not affect on the order you want to merge (replaces strings mostly within its own categories). Besides the CRC it's best to also have the ability to show serial and even make entries show up as heavy games to run like GT4 or Ratchet,sly,.. I maybe getting ahead of myself. We will see then.

arcum42:
I'll go ahead and merge this in, anyways. I'm sure there will be further gui changes later.

PR #3602, opened 2020-08-15
Minor fix to the C++ GS Dump functionality (for debugging purposes)

hibye8313:
Minor update the C++ GSReplay functions (for debugging purposes) to be more accurate and match the C# GSDumpGUI app.

When I was debugging the GS a year or so ago I was using the GSReplay function in GS.cpp noticed that it was slightly different from the GSDumpGUI app because of the way VSync is handled. The C++ function used the FIELD value from the dump, but the C# version uses the FIELD values in the CSR register:
GSDXWrapper.cs line 386:
`GSVSync((*((int*)(registers + 4096)) & 0x2000) > 0 ? (byte)1 : (byte)0);`

Its not a terribly important change and I don't know if anyone else uses this function for debugging but it might save someone else some headache.

PS. How do people usually debug graphical problems? I have been using C++ GSReplay because that way I can set breakpoints in the actual GS code and dump textures, do logging, etc. To me the GSDumpGUI app is not very useful because it runs in a separate process from the GS (can't set breakpoints).

### DIFF
--- a/pcsx2/MTGS.cpp
+++ b/pcsx2/MTGS.cpp
@@ -444,8 +444,8 @@ void SysMtgsThread::ExecuteTaskInThread()
 							((u32&)RingBuffer.Regs[0x1010])				= remainder[1];
 							((GSRegSIGBLID&)RingBuffer.Regs[0x1080])	= (GSRegSIGBLID&)remainder[2];
 
-							// CSR & 0x2000; is the pageflip id.
-							GSvsync(((u32&)RingBuffer.Regs[0x1000]) & 0x2000);
+							// CSR & 0x2000 is the FIELD value (1 for odd, 0 for even).
+							GSvsync(!!(((u32&)RingBuffer.Regs[0x1000]) & 0x2000));
 							gsFrameSkip();
 
 							// if we're not using GSOpen2, then the GS window is on this thread (MTGS thread),
--- a/plugins/GSdx/GSDump.cpp
+++ b/plugins/GSdx/GSDump.cpp
@@ -65,7 +65,7 @@ void GSDumpBase::ReadFIFO(uint32 size)
 	AppendRawData(&size, 4);
 }
 
-bool GSDumpBase::VSync(int field, bool last, const GSPrivRegSet* regs)
+bool GSDumpBase::VSync(bool odd_field, bool last, const GSPrivRegSet* regs)
 {
 	// dump file is bad, return done to delete the object
 	if (!m_gs)
@@ -75,7 +75,7 @@ bool GSDumpBase::VSync(int field, bool last, const GSPrivRegSet* regs)
 	AppendRawData(regs, sizeof(*regs));
 
 	AppendRawData(1);
-	AppendRawData(static_cast<uint8>(field));
+	AppendRawData(odd_field ? 1 : 0);
 
 	if (last)
 		m_extra_frames--;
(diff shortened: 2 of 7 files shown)


tadanokojin:
I wouldn't imagine it makes a difference. Does it fix something?
FWIW I'm working on a portable version of GSDumpGUI and I use the value from the dump.

For debugging typically you'd want to dump the draw calls and look at the GL log.
<https://wiki.pcsx2.net/PCSX2_Documentation/GSdx_Debug>

hibye8313:
I think the reason I noticed this was because there was a difference in a certain game (Genji I believe) where it was not updating the correct field so you got a black "comb" through the image (every other line black). TBH I don't remember exactly because it was  so long ago. Maybe it makes no difference, let me look at the code and see if the dumped values is the same as the CSR register value.

Nice info thank you. I didn't know you could dump textures also in the GS dump. I used to just putting in temporary code to dump texture to png files.

hibye8313:
Ok so the problem was actually with the dumping function not the replaying function. MTGS was giving the GSVsync a FIELD value of 8192 or 0 (for FIELD = 1 or FIELD = 0). In the interlace function there was no problem because it was being converted to 1 or 0 with code:
`field ? 1 : 0`
But in the dumping function it was being used as is and then begin cast to uint8 so both 8192 and 0 were converted to just 0. I just added the same line of code to the dumping function. Now I can verify that the GS Dumps are recording the correct value of FIELD. This probably doesn't make much difference unless games are using some specific interlace mode (field or frame, I don't know which).

If some one can review this it would be greatly appreciated. Also, would it be acceptable to change field from and int to bool everywhere in the code? I think it would make it more clear that it should be 0 or 1 only and less prone to such errors.

tadanokojin:
Okay, likely what is happening here is you are getting a CSR bit but without the shift.
So FIELD bit of CSR is 13th bit. `1 << 13 = 8192`.

So you could do the most obvious:
```
(field >> 13) & 0x1
// or assuming no other bits set
field >> 13
```

Or assuming no other bits are set (which is what the current code assumes), the less obvious cast to bool:
```
!!field
```

I'm not sure I understand why an int was used here to begin with. Unless I missed something you might as well make it `bool odd_field` and do your cast in GS.cpp.

Anyway good find, I agree the code isn't clear which is probably what led to this bug.

hibye8313:
Thanks for the reply. Yeah I can go ahead and change field -> odd_field. It is used is a few different functions so it might be a bigger code change than it might appear (the field variable is passed down to the subclasses 2-3 levels deep). Mostly likely there are no bugs due to this in the core functionality of pcsx2 (since it would have been noticed long before) but at least it will prevent other mistakes from creeping in.

tadanokojin:
Sounds good.
Thank you for your time it will save me trouble when we switch away from GSDumpGUI.

hibye8313:
No problem, I think I changed everything relevant to the the field parameter and tried to clean up the bitwise code a little bit in GSDevice.cpp. Let me know if there are any problems so I can make any needed changes.

@tadanokojin
A new version of GSDumpGUI sounds like a good idea. If its ok for me to make a suggestion it would be very useful to have the tool built right into PCSX2 (in C++) that way it could be much more convenient and (more importantly) it would allow devs to set breakpoints or add printf's directly in the GS code directly and quickly recompile and execute.

PR #3604, opened 2020-08-16
PNACH: New patch type: apply repeating codes also on load

GovanifY:
There is no real reason to make the behavior of type-{0,1} pnach different on load. Keeping the behavior consistent not only makes more sense but also avoids some weird edge case we could run into.

### DIFF
--- a/pcsx2/Patch.h
+++ b/pcsx2/Patch.h
@@ -61,9 +61,20 @@ enum patch_data_type {
 // PCSX2 currently supports the following values:
 // 0 - apply the patch line once on game boot/startup
 // 1 - apply the patch line continuously (technically - on every vsync)
+// 2 - effect of 0 and 1 combined, see below 
+// Note:
+// - while it may seem that a value of 1 does the same as 0, but also later
+//   continues to apply the patch on every vsync - it's not.
+//   The current (and past) behavior is that these patches are applied at different
+//   places at the code, and it's possible, depending on circumstances, that 0 patches
+//   will get applied before the first vsync and therefore earlier than 1 patches.
+// - There's no "place" value which indicates to apply both once on startup
+//   and then also continuously, however such behavior can be achieved by
+//   duplicating the line where one has a 0 place and the other has a 1 place. 
 enum patch_place_type {
 	PPT_ONCE_ON_LOAD = 0,
 	PPT_CONTINUOUSLY = 1,
+	PPT_COMBINED_0_1 = 2,
 
 	_PPT_END_MARKER
 };
--- a/pcsx2/Patch.h
+++ b/pcsx2/Patch.h
@@ -61,16 +61,6 @@ enum patch_data_type {
 // PCSX2 currently supports the following values:
 // 0 - apply the patch line once on game boot/startup
 // 1 - apply the patch line continuously (technically - on every vsync)
-// Note:
-// - while it may seem that a value of 1 does the same as 0, but also later
-//   continues to apply the patch on every vsync - it's not.
-//   The current (and past) behavior is that these patches are applied at different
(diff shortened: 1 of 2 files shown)


kozarovv:
Little bit late for change like that to pnach system. This break WRC4 patch, and possibly few others. For example my patch for GOW2 make game widescreen/progressive at start as a default setting, but user can go back to 4:3/interlace. Without current patch=0 behavior that patch need to be translated into D or E conditional code. For WRC4 there is no other way than patch=0 because patch hit entrypoint of game, and need to be cleaned right after it.  

Maybe adding patch=2 will be better idea? That can be mixed 0 +1 behavior.

RedDevilus:
Some ideas i have;
PPT_ALWAYS_ON = 2,
PPT_ENDLESS_ON = 2,
PPT_COMBINED_0_1 = 2,

Not sure what would fit best in this case. Could be nice to expand the system.

GovanifY:
@RedDevilus @kozarovv Done!

prafullpcsx2:
This PR can break a lot of existing patches. Can you please explain in detail what benefits does this pr actually brings?

GovanifY:
@prafullpcsx2 This PR has been changed to allow a new patch=2 type which allows the combined effects of 0 and 1, aka executing every VSync AND at the start of the game. This does not break any implementation as no pnach of type patch=2 was supported up until now.

refractionpcsx2:
Okay but what benefit will it bring being one frame earlier than mode 1? if it's something that affects in-game then this will not help, if it only  affects initial boot, 0 mode is fine, as is the existing 1 mode, which seems to work effectively. In that first frame it is mostly BIOS boot initialisation that happens, so I'm struggling to see the benefit of this.  

Have you got an example of where this is an advantage?

kozarovv:
> In that first frame it is mostly BIOS boot initialisation that happens, so I'm struggling to see the benefit of this.

0 patches are applied on eeGameStarting, so right after bios for first recompiled blocks of the game. Those blocks sometime load float value that keep menu ratio, etc (GOW2 for example, patch=1 is too late).

refractionpcsx2:
> > In that first frame it is mostly BIOS boot initialisation that happens, so I'm struggling to see the benefit of this.
> 
> 0 patches are applied after eeGameStarting, so right after bios for first recompiled blocks of the game. Those blocks sometime load float value that keep menu ratio, etc (GOW2 for example, patch=1 is too late).

but surely that's a good thing if you want to change the menu ratio? you want to write it *after* the value has been written to memory and the menu itself won't be loaded in that first frame.  Writing the patch value before it has been written will achieve nothing as far as I understand it.

GovanifY:
> Have you got an example of where this is an advantage?
Basically any flag which fires up an initializer at startup(eg, mallocs and the like) and which then is changed at runtime later, eg by parsing assets. Making a patch=0 would disable the feature, making the patch=1 would crash the game, requiring either to patch the game code or this patch=2.

refractionpcsx2:
> > Have you got an example of where this is an advantage?
> 
> Basically any flag which fires up an initializer at startup(eg, mallocs and the like) and which then is changed at runtime later, eg by parsing assets. Making a patch=0 would disable the feature, making the patch=1 would crash the game, requiring either to patch the game code or this patch=2.

surely if it crashes on patch=1 because it's changed later that patch=2 would also crash no?

GovanifY:
> surely if it crashes on patch=1 because it's changed later that patch=2 would also crash no?

Nope, patch=1 crashes because it was not set on init, patch=2 sets it on initialization but continue every VSync, forcing the flag to be enabled even though its state changes later.

refractionpcsx2:
I see, okay, i guess it can be useful then

PR #3605, opened 2020-08-17
GSdx: Add Default OSD Font Resource

### LINKED ISSUE #3592: Missing TT fonts under Arch
Suggested solution

Add to PKGBUILD:

 * `sed -i 's/\/usr\/share\/fonts\/truetype\/freefont\/FreeSerif.ttf/\/usr\/share\/fonts\/gnu-free\/FreeSerif.otf/g' plugins/GSdx/GSdx.cpp`
 * ` gnu-free-fonts` to the dependencies

beaumanvienna:
fixes #3592


tellowkrinkle:
`_WIN32` is defined on all Windows platforms, no need to check both it and `_WIN64`

For the sans serif font, how about DejaVu Sans?

tadanokojin:
I'm okay with whatever sans serif is good with our license (not an expert on the legal stuff). I think I suggested Roboto but don't know if it will work with the GPL.

PR #3608, opened 2020-08-19
64-bit recompilers

tellowkrinkle:
The main piece of #3451

Makes recompilers compatible with x86-64


lightningterror:
Maybe some commits can be squashed and better named.

tellowkrinkle:
How's that for squashing?

lightningterror:
Yeah it's better, I'd name the commits better but it's fine.

PR #3611, opened 2020-08-19
Remove GSdx Legacy from master branch

lightningterror:
Linux only, same as gsdx from 1.4 branch.

Follow up from #2842


orbea:
Given PR https://github.com/PCSX2/pcsx2/pull/3562 does `-DEGL_API` actually do anything after this?

lightningterror:
Used for zzogl I guess?

build.sh
```diff
@@ -240,7 +240,7 @@ for ARG in "$@"; do
             echo
             echo "** Developer option **"
             echo "--glsl          : Replace CG backend of ZZogl by GLSL"
-            echo "--egl           : Replace GLX by EGL (ZZogl/GSdx_legacy plugins)"
+            echo "--egl           : Replace GLX by EGL (ZZogl plugin)"
```
RedDevilus, line 243: ZZogl/GSdx plugins?
Atleast it has no backspace.

orbea:
> Used for zzogl I guess?

Does anyone use zzogl?  The last commit was in 2018 (https://github.com/PCSX2/pcsx2/commit/38da174670d8bd058f293ac39ddd3bf56a9d14c8) and it doesn't seem maintained. Maybe this is something else that should be purged? I'm not very familiar with it.

lightningterror:
We'll get to those too, this was mainly done so going for it first.

arcum42:
> > Used for zzogl I guess?
> 
> Does anyone use zzogl? The last commit was in 2018 ([38da174](https://github.com/PCSX2/pcsx2/commit/38da174670d8bd058f293ac39ddd3bf56a9d14c8)) and it doesn't seem maintained. Maybe this is something else that should be purged? I'm not very familiar with it.

If anyone was maintaining it, it'd be me at this point. But once GSdx was ported to work on linux, it wasn't really needed any more, and I doubt most people use it.

It's based off of ZeroGS, which would be another candidate for deletion, as would ZeroSPU2.

PR #3614, opened 2020-08-20
Remove zzogl from master branch

lightningterror:


Florin9doi:
The wiki should be updated too: https://github.com/PCSX2/pcsx2/wiki/Installing-on-Linux

arcum42:
Areas that still have mentions of zzogl I noticed:
.gitignore - Line 25
cmake/SelectPcsx2Plugins.cmake - Lines 13-17
debian-packager/create_built_tarball.sh - Line 103

In linux_various, validate_glsl.sh and glsl2h.pl look zzogl-specific, too, though the former has an unimplemented section for testing GSdx, so I'm not absolutely sure whether they are likely to be used at some future point...

PR #3615, opened 2020-08-21
Add new "Game Manager" window with Saveslot listing and screenshot previews

xTVaser:
Successor to - https://github.com/PCSX2/pcsx2/pull/3565#issuecomment-663930630

This adds a new window to PCSX2 that for right now, only has a listing of all save-states with some basic information (what is already in the System menu).  And on the right hand column, the slot preview image.

The selected slot "row" (current + backup) is highlighted in a pale blue, while left clicking on a slot expands the preview image and highlights with a darker blue.

image

There are some basic menu options, currently just auto-docking like with the console logger, and toggling displaying all backup rows:
image
image

Lastly, this is visible through the "Misc" menu, just like the console logger.  Though I believe this is changing to the "Window" menu as per https://github.com/PCSX2/pcsx2/pull/3597 (I expect there's going to be atleast a few conflicts)

image

## Issues
- I ran into some weirdness where when the `ini` file is first created, booleans that are defaulted to true, aren't set to true.  So this might result in some unexpected behaviour.
- There is some slight flickering in the UI that I could likely avoid, my plan is to add an ISO tab in another PR which will more or less follow the same format so something I'll continue to try to improve.


xTVaser:
Hmm, yeah that's pretty confusing to me.  My thought was that `updateSlot` was being called prior to being properly initialized like you mentioned earlier.  However that doesn't really make sense.  If `updateSlot` is able to be called, I'd expect there to be a proper instance of `SavestateTab` and the vector should be initialized properly.  However, even a vector that is only declared, starts with size `0`.

Something weirder is probably happening, I can't seem to reproduce in my environment so maybe it's an `ini` issue, or there's a race-condition elsewhere.  Working on setting up my linux environment again, might take me a bit as the 20.04 steps in the wiki don't seem to work for compiling GSdx.

Just to be clear, you are crashing immediately on opening PCSX2, or immediately when booting a game?

arcum42:
It's immediately on opening pcsx2. I'll try pulling a fresh copy, just in case...

Re: the build error, do you have libx11-xcb-dev installed?

Sorry about all the conflicts, btw. I think basically anything that touches a menu needs a rebase now...

xTVaser:
No worries about the conflicts, I've dealt with far far worse :).  I'll give that package a shot, I don't remember explicitly checking/installing that.

xTVaser:
The last commit resolves the immediate crash issue.  For whatever reason `FindWindowById()` was not returning NULL before `m_id_GameManagerFrame` was properly initialized (still `0`).  Though, running these changes on linux reveals a lot to be desired:
- I need to adjust the color scheme if the user is using dark mode, hopefully theres a way to detect that easily.
screenshot-from-2020-08-24-20-
- The implicit gap between windows that exists on Windows is absent on linux, either I should remove what I added, or maybe add a gap to the console log window.
screenshot-from-2020-08-24-20-(1)
- But worst of all, seems like the GTK wxWidgets doesn't behave the same way it does in windows.  The window/sizers do not adjust in the same way which leads to basically an ugly/unusable UI.  The console log / game manager also seem to lock up when I'm running a game but wasn't able to confirm if this is an issue on master (i assume its not or else I wouldn't be the first) because I get a GSdx failed to load upon booting a game.  (I see `Error: GLLoader failed to get GL version` in the logs).

So probably going to take a bit of time to work through these problems.  How does one debug pcsx2 effectively on linux, if I just naively attach gdb to the process, I constantly get hung up on exceptions in iR5900-32.cpp

arcum42:
Glad that package did it! I'd seen an issue go through recently where apparently that used to be a dependency of one of the other dependencies in Ubuntu, but isn't any more.

Yeah, unfortunately wx's behavior can vary from platform to platform visually. I was definitely running into spots where Mac OS X didn't act the same way as gtk on things. With the crash gone, I'll take a look, though. (There are reasons why we repeatedly talk about redoing the gui in things like Qt instead...)

I hate to say that most of the time, I just add in Console.WriteLn commands everywhere and tell it to write to stdio. There's a section in the pcsx2 wiki for gdb, though:
https://wiki.pcsx2.net/PCSX2_Documentation/GNU_Debugger_Tips

The main important bit is typing:
handle SIGSEGV nostop

so it doesn't stop all the time.

xTVaser:
Just to provide an update: I think I've fought enough with my linux environment that I can start to cobble together some fixes.  I still have issues using GSdx on master, and might as well forget about having intellisense in my editor, but I can manage.  In general, it would be a _huge_ help if there was a page on the repo's wiki about setting up a linux-dev environment for pcsx2.

I think I've found the majority of the issues, have solutions in my head to resolve them.  Hopefully by the end of the week I'll have this cleaned up.

arcum42:
Main image that stood out in Linux was that none of the images were loading.

Going over to SavestateSlotPanel::initPreview, if I change wxBitmap(img, wxBITMAP_TYPE_PNG) to wxBitmap(img) in both spots, the images start loading, and some warnings that were spamming the console went away.

xTVaser:
@arcum42 Thanks for taking a look! I was hoping the big issues I was seeing were just on my end...sounds like that's mostly the case.  And yeah I had some issues around the images in linux, but removing the image type is a strange fix...wouldn't expect that to be the problem.  What were the warnings?  The latest changes should be fairly safe (if the image fails to load, it should just not update the preview, but I'm _sure_ wx loves to log an error anyway)

Are the images saved alongside your save-states not `.png`s?  Can you attach one here so i could try it out on my end.  I assume `wxBitmap(img)` figures out the image type, so maybe it's the better way to go anyway.

arcum42:
```(PCSX2:1016481): Gdk-WARNING **: 20:48:11.980: The gdk_draw_*_image require the drawable argument to
have a specified colormap. All windows have a colormap,
however, pixmaps only have colormap by default if they
were created with a non-NULL window argument. Otherwise
a colormap must be set on them with gdk_drawable_set_colormap

(PCSX2:1016481): Gdk-WARNING **: 20:48:11.991: The gdk_draw_*_image require the drawable argument to
have a specified colormap. All windows have a colormap,
however, pixmaps only have colormap by default if they
were created with a non-NULL window argument. Otherwise
a colormap must be set on them with gdk_drawable_set_colormap

(PCSX2:1016481): Gdk-WARNING **: 20:48:11.998: Attempt to draw a drawable with depth 15 to a drawable with depth 24

(PCSX2:1016481): Gdk-WARNING **: 20:48:11.999: Attempt to draw a drawable with depth 15 to a drawable with depth 24

(PCSX2:1016481): Gdk-WARNING **: 20:48:12.000: Attempt to draw a drawable with depth 15 to a drawable with depth 24

(PCSX2:1016481): Gdk-WARNING **: 20:48:12.001: Attempt to draw a drawable with depth 15 to a drawable with depth 24
```

example

If I make the change I gave you:
example2

Here's one of the sstate pictures from the sstate dir:
SLUS-21445 (4437F4B1) 06 p2s_alpha
SLUS-21445 (4437F4B1) 06 p2s

xTVaser:
@GovanifY 
I shelved this due to design constraints imposed by wxWidgets.  I'm not able to make something that I'm happy with / nor that looks consistent cross-platforms with it.  Feel free to close if it's getting in way, though it could remain as a PoC / I'll update it as soon as we get Qt established.

PR #3627, opened 2020-08-24
GUI/Recording - Add menu items for recording controls with keyboard shortcuts.

xTVaser:
This makes the recording tools a little more intuitive to use since there still is no proper documentation around it.

In the process, I found out that the current way of adding the keyboard shortcuts to the menu items would omit modifiers such as `shift`.  So I fixed that in both places.

By default wxWidget's `ToString` will return the keyboard character in it's correct case.  For example, if the bind is `Shift+p` it will be printed as such.  However, in most UI's I've seen, the key is capitalized anyway.  So I had to write a small routine to do so.

New menu options (note that the keybinds will only appear once the GS window has been opened, as this is when the majority of pcsx2's keybindings are assigned)
image

Existing binds are still set properly after my changes:
image

### DIFF
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -471,6 +471,10 @@ void MainEmuFrame::CreateRecordMenu()
 	m_menuRecording.Append(MenuId_Recording_Stop, _("Stop"))->Enable(false);
 	m_menuRecording.Append(MenuId_Recording_Play, _("Play"));
 	m_menuRecording.AppendSeparator();
+	m_menuRecording.Append(MenuId_Recording_TogglePause, _("Toggle Pause"));
+	m_menuRecording.Append(MenuId_Recording_FrameAdvance, _("Frame Advance"));
+	m_menuRecording.Append(MenuId_Recording_ToggleRecordingMode, _("Toggle Recording Mode"));
+	m_menuRecording.AppendSeparator();
 	m_menuRecording.Append(MenuId_Recording_VirtualPad_Port0, _("Virtual Pad (Port 1)"));
 	m_menuRecording.Append(MenuId_Recording_VirtualPad_Port1, _("Virtual Pad (Port 2)"));
 #endif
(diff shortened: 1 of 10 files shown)


RedDevilus:
What's the purpose of Toggle Recording Mode?

xTVaser:
> What's the purpose of Toggle Recording Mode?

There's currently two recording "modes":
- Recording:
  - new frame's inputs will be saved to the recording file
  - existing frame's inputs will be overwritten
- Replaying:
  - Existing inputs are not modified
  - No new frames will be recorded and will pause at the end of the recording

RedDevilus:
Recording as in it will save to a MP4?
Replaying as in blockdumps or macros?

xTVaser:
None of the above.  The recording tools in pcsx2 record the controller's inputs (an array of bytes) to a file every frame.  When the time comes that you want to export it as an actual video, you would replay it and turn on pcsx2's video dumping feature (F12...if I recall correctly).

pcsx2/gui/MainFrame.cpp
```diff
@@ -765,24 +776,32 @@ void MainEmuFrame::CommitPreset_noTrigger()
...
+void MainEmuFrame::initializeRecordingMenuItem(MenuIdentifiers menuId, wxString keyCodeStr) {
+	wxMenuItem& item = *m_menuRecording.FindChildItem(menuId);
+	wxString text = item.GetItemLabel();
+	const size_t tabPos = text.rfind(L'\t');
+	item.SetItemLabel(text.Mid(0, tabPos ) + L"\t" + keyCodeStr);
+	item.Enable(true);
+}
+void MainEmuFrame::enableRecordingMenuItem(MenuIdentifiers menuId, bool enable)
+{
+	wxMenuItem& item = *m_menuRecording.FindChildItem(menuId);
```
sonicfind, line 801: The debugger threw a read access violation exception at this point - showing that `m_menuRecording.FindChildItem(menuId)` is already nullptr. This error will appear when you exit the emulator so long as you have booted an ISO at least one time - regardless of if recording tools are enabled or not.

xTVaser: Ah ok, that might be due to the recent GUI refactor, I recall you finding an issue there as well?  I can solve that here at the same time...hopefully.  Looking into it.

xTVaser: Should be resolved with the latest commit.  The `s*` macros automatically handle the `null` case for windows.

sonicfind:
I noticed that the menu options aren't actually getting disabled on pause or shutdown.
They are however getting enabled on resume or boot. So I went to see why this was. Two big things:
1. The piece that disables the options should be in `Pcsx2App::CloseGsPanel()` instead of `Pcsx2App::OnGsFrameClosed( wxWindowID id )` since the latter is only called on exit... Or at least, I would say put it in `Pcsx2App::CloseGsPanel()`, but
2. `Pcsx2App::CloseGsPanel()` is never actually getting called from anywhere - all thanks to this value in App.h:
```
// Forces the Interface to destroy the GS viewport window when the GS plugin is
// destroyed.  This has the side effect of forcing all plugins to close and re-open
// along with the GS, since the GS viewport window handle will have changed.
static const bool CloseViewportWithPlugins = false;
```
The only time it would get called (AppCorePlugins.cpp) gets short-circuited to false
```
// Yay, this plugin is guaranteed to always be opened first and closed last.
void AppCorePlugins::ClosePlugin_GS()
{
	_parent::ClosePlugin_GS();
	if( CloseViewportWithPlugins && GetMTGS().IsSelf() && GSopen2 ) sApp.CloseGsPanel();
}
```

RedDevilus (approved):

sonicfind:
768p100%dpi
image
Should be gucci once the small bar on the right side of this is fixed. I see literally nothing else to worry about after that.

xTVaser:
Ran locally.  Shortcuts still showing up:
image
image

lgtm

pcsx2/gui/MainFrame.cpp
```diff
@@ -776,24 +783,32 @@ void MainEmuFrame::CommitPreset_noTrigger()
...
 void MainEmuFrame::AppendKeycodeNamesToMenuOptions() {
-	AppendShortcutToMenuOption( *m_menuSys.FindChildItem( MenuId_Sys_LoadStates ), "States_DefrostCurrentSlot" );
-	AppendShortcutToMenuOption( *m_menuSys.FindChildItem( MenuId_Sys_SaveStates ), "States_FreezeCurrentSlot" );
+	
+	AppendShortcutToMenuOption(*m_menuSys.FindChildItem( MenuId_Sys_LoadStates ), wxGetApp().GlobalAccels->findKeycodeWithCommandId("States_DefrostCurrentSlot").toTitleizedString());
+	AppendShortcutToMenuOption(*m_menuSys.FindChildItem( MenuId_Sys_SaveStates ), wxGetApp().GlobalAccels->findKeycodeWithCommandId("States_FreezeCurrentSlot").toTitleizedString());
 }
 
+#ifndef DISABLE_RECORDING
+void MainEmuFrame::initializeRecordingMenuItem(MenuIdentifiers menuId, wxString keyCodeStr, bool enable) {
```
refractionpcsx2, line 799: Bracket needs to go on to the next line

pcsx2/gui/MainFrame.cpp
```diff
@@ -776,24 +783,32 @@ void MainEmuFrame::CommitPreset_noTrigger()
...
+	wxMenuItem& item = *m_menuRecording.FindChildItem(menuId);
+	wxString text = item.GetItemLabel();
+	const size_t tabPos = text.rfind(L'\t');
+	item.SetItemLabel(text.Mid(0, tabPos ) + L"\t" + keyCodeStr);
+	item.Enable(enable);
+}
+void MainEmuFrame::enableRecordingMenuItem(MenuIdentifiers menuId, bool enable)
```
refractionpcsx2, line 806: maybe add a space between functions

refractionpcsx2 (changes requested):

refractionpcsx2 (approved):

PR #3633, opened 2020-08-27
Misc: Mac build/fixes

tellowkrinkle:
Changes to make GSdx build on macOS minus old GL workarounds (so the built GSdx won't run any games, but good for getting CI working)

As this gets all plugins building on macOS, I also added macOS to the CI

BTW do you think I should include the macOS window creation files (`plugins/GSdx/Window/GSWndCGL*`)?


lightningterror:
I'd say skip the window files for now as we don't have a proper renderer.

plugins/GSdx/GS.cpp
```diff
@@ -476,7 +485,7 @@ EXPORT_C_(int) GSopen2(void** dsp, uint32 flags)
 		}
 
 #endif
-#if defined(__unix__)
+#if defined(__unix__) || defined(__APPLE__)
```
lightningterror, line 488: Maybe setting Null renderer as default for macos would be better too.

tellowkrinkle: At the moment GSopen2 fails if it can't make a window (and the GL window considers itself failed if your GL implementation doesn't at least meet the requirements of the software renderer) so it doesn't really matter

lightningterror (approved):

plugins/onepad/keyboard.cpp
```diff
@@ -99,7 +99,7 @@ bool PollForNewKeyboardKeys(u32 &pkey) {
     // keys that aren't being recognized, bump this number up!
     for (int key = 0; key < 0x80; key++) {
         if (CGEventSourceKeyState(kCGEventSourceStateHIDSystemState, key)) {
-            pkey = key == kVK_Escape ? 0 : key;
+            pkey = key == kVK_Escape ? -1 : key;
```
arcum42, line 102: pKey is unsigned, and it's potentially being set to -1.

tellowkrinkle: Would you prefer `(u32)-1` or `0xFFFFFFFF`?

tadanokojin: Why not ```UINT32_MAX```

lightningterror:
Maybe this should wait for https://github.com/PCSX2/pcsx2/pull/3642 to be merged.

xTVaser:
MacOS github actions workflow based on this branch working - https://github.com/xTVaser/pcsx2-rr/pull/36/

lightningterror (approved):

PR #3642, opened 2020-08-30
Convert GS Settings dialog to use wxwidgets

arcum42:
This scraps the old gtk GSdx dialog box, and replaces it with a wx version.

Current issues:
While I added in tooltips, they don't seem to actually be showing for some reason.
It still has the code for adding the pictures into a gtk resource and bundling them with the program, and the wx dialog does not use those pictures. Instead, it adds a logo if it finds one in the plugin folder. It's entirely possible we may want to remove the logo anyways, so not a major issue.

Otherwise, the dialog box should be hooked up and working, and I felt it was a good idea to go ahead and open the pr for it and get more eyes on it, though I would like to have the tooltip issue resolved before actually merging.


arcum42:
Without logo:
gsdx-nologo

With logo:
gsdx-logo

lightningterror:
With logo looks better, also for adapter maybe comment it out until it's used?

tadanokojin:
I mentioned before in discord, but I prefer if we just got rid of the branding. It's kind of pointless for the direction we are going. Do you need help with the adapter stuff?

arcum42:
With the logo, the current code shows the logo as in the second screenshot if it sees a copy of logo-ogl.bmp in the plugin folder, otherwise you get the first screenshot. 

If we don't want the logo, then leaving it as is would be fine. If we do, the choices would either be dropping a copy of that picture in that folder, or rewriting the resource code to pull it in to be wx, which will rather be a pain, though I'm sure the main program does that, so I could potentially figure out how to do it from that.

As far as the adapter goes, depends how far we're going to go with it. If we want to leave it as nonfunctional, I could disable it or hide it, I suppose. If we want to be able to pick a video card from it for OpenGL, that I could probably use help with. I mainly implemented it for later use if we replace the windows dialog box with it as well.

Still baffled by the fact the tooltips aren't working...

tadanokojin:
Driver selects the adapter for GL so you might as well nop it out of Linux (VK can select an adapter later down the line). Really it's just DX. That's why I'm asking if you need my help on it since that's obviously going to be win only :P

arcum42:
Well, making this dialog work in Windows is probably outside the scope of this pr. I was figuring I may go back later, and convert the various wx plugin dialog boxes to be used for windows as well in a separate pr or prs, and when we get to that point, I could probably use help on getting the adapters working.

tadanokojin:
Fair enough, thought I'd ask since I saw the adapter code.

arcum42:
Not saying I'd object if code to make the whole thing work in windows as well got dropped into the branch, mind you. :p 

Technically, I can work in windows, it just involves going to a different computer with a development environment I'm less comfortable with and only one monitor hooked up. I figure if I'm going to do that, might as well get it all out of the way at the same time...

tellowkrinkle:
Working on macOS as well 
image

arcum42:
Cool. TBH, I think it looks better in the Mac OS screenshot than in Linux...

tadanokojin:
I took a quick look at windows and it looks like character encoding is going to make this a pain. I couldn't get wx to compile with multibyte (what gsdx currently uses) and it would require some work to get it to compile with unicode so maybe you are wise to hold off on it for a little bit.

plugins/GSdx/Window/GSwxDialog.cpp
```diff
@@ -0,0 +1,917 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with GNU Make; see the file COPYING.  If not, write to
+ *  the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA USA.
+ *  http://www.gnu.org/copyleft/gpl.html
+ *
+ */
+
+#include "GSwxDialog.h"
+
+void add_tooltip(wxWindow *widget, int tooltip)
```
tadanokojin, line 23: I think it would be good to make these helper functions static or use an anonymous namespace so it is unique to the translation unit

plugins/GSdx/Window/GSwxDialog.cpp
```diff
@@ -0,0 +1,917 @@
...
+    tab_box->Add(hacks_check, wxSizerFlags().Left());
+    tab_box->Add(rend_hacks_box, wxSizerFlags().Centre().Expand());
+    tab_box->Add(upscale_hacks_box, wxSizerFlags().Centre().Expand());
+
+    SetSizerAndFit(tab_box);
+    Bind(wxEVT_SPINCTRL, &HacksTab::CallUpdate, this);
+    Bind(wxEVT_CHECKBOX, &HacksTab::CallUpdate, this);
+}
+
+void HacksTab::CallUpdate(wxCommandEvent &event)
```
tadanokojin, line 288: You can do something like
```cpp
void HacksTab::CallUpdate(wxCommandEvent& /*event*/)
{
}
```

It'll save you a compiler warning. Same goes for anywhere else this occurs. It's a minor thing though so I'll leave it up to you.

plugins/GSdx/GS.cpp
```diff
@@ -45,6 +45,7 @@ static HRESULT s_hr = E_FAIL;
 #include "Window/GSWndEGL.h"
 
 extern bool RunLinuxDialog();
+extern bool RunwxDialog();
```
tadanokojin, line 48: You don't really need the extra extern here. You already extern it in the header, so you can just include the header and you'll have the function. I prefer it that way anyway.

plugins/GSdx/Window/GSwxDialog.cpp
arcum42: I moved them to an anonymous namespace. Since my header has lots of "Type *name, *name2, *name3, *name4; ", trying to put the * on the Type starts looking fairly inconsistent. I just ran the files through clang-format, though, so they should adhere to any standards set in the format file now.

plugins/GSdx/Window/GSwxDialog.cpp
arcum42: Done.

plugins/GSdx/GS.cpp
arcum42: Taken care of, and I removed the commented old code while I was at it.

arcum42:
The main blocker on this one is that the tooltips aren't working, btw. Not sure I actually have put tooltips in any wx dialog that's in a plugin before, though, so it could very well be that something with that's blocking it.

tellowkrinkle:
I rebased and cleaned up the code a bit

Tooltips seem to work fine on macOS, have you checked to see if PCSX2's main UI's tooltips work for you?

arcum42:
@tellowkrinkle 
Thanks!

Checked out the branch and did a clean build. Still seems like no tooltips. My main thought was that once GSdx is merged, we'll see how it does as far as tooltips with the merged version, because I've had this feeling it might have to do with being in a plugin and not part of the main program...

arcum42:
Rebasing this branch seemed tricky enough that I just copied the changes to a wx-gs branch instead...

arcum42:
Since the rebase has been taken care of, I'll delete the other branch I made.

Tooltips are sort of working half the time and not working the other half for me on Linux. At this point, I'd be fine with just moving forward with merging it if that's the only issue. I'm sure we'll rewrite without wx later anyways...

pcsx2/GS/Window/GSwxDialog.cpp
```diff
@@ -0,0 +1,659 @@
...
+
+void DebugTab::Update()
+{
+	m_ui.Update();
+	if (end_dump_spin->GetValue() < start_dump_spin->GetValue())
+		end_dump_spin->SetValue(start_dump_spin->GetValue());
+}
+
+Dialog::Dialog()
+	: wxDialog(nullptr, wxID_ANY, "GS Config", wxDefaultPosition, wxDefaultSize, wxCAPTION | wxCLOSE_BOX | wxRESIZE_BORDER)
```
Florin9doi, line 541:
```suggestion
	: wxDialog(nullptr, wxID_ANY, "Graphics Settings", wxDefaultPosition, wxDefaultSize, wxCAPTION | wxCLOSE_BOX | wxRESIZE_BORDER)
```

tadanokojin:
> At this point, I'd be fine with just moving forward with merging it if that's the only issue.

Gotta sneak a commit before that for making the adapter selection work on windows.

> Tooltips are sort of working half the time and not working the other half for me on Linux. 

I wanna pull them out of the resource header and inline them. Eventually I wanna get us setup for translations there but it can wait.

arcum42:
Yeah, we'll want to take care of anything still needed for getting it to work properly on Windows first, and any other touchups that are needed. I'll be glad to see things out of that resource file, too...

pcsx2/GS/Renderers/DX11/D3D.cpp
```diff
@@ -0,0 +1,133 @@
...
+		// we use CreateDXGIFactory2 because we assume at least windows 8.1 anyway
+		CComPtr<IDXGIFactory2> factory;
+		HRESULT hr = CreateDXGIFactory2(flags, IID_PPV_ARGS(&factory));
+
+		// if we failed to create a factory with debug support
+		// try one without
+		if (FAILED(hr) && debug)
+		{
+			fprintf(stderr, "D3D: failed to create debug dxgi factory, trying without debugging\n");
+			hr = CreateDXGIFactory2(0, IID_PPV_ARGS(&factory));;
```
tadanokojin, line 24: note to self

pcsx2/GS/Window/GSwxDialog.h
```diff
@@ -0,0 +1,204 @@
...
+	};
+
+	class Dialog : public wxDialog
+	{
+		GSUIElementHolder m_ui;
+
+		wxBoxSizer* m_top_box;
+		wxChoice* m_renderer_select;
+		wxChoice* m_adapter_select;
+		wxArrayString m_adapter_arr_string;
```
tadanokojin, line 182: note to self

tadanokojin:
I added the adapter select but there's still win32 problems.

For one the tab text rendering is broken:
image

Additionally, this would remove a bunch of UX from the Windows config. Namely things are no longer contextually disabled anymore. For example it's possible now to enable hw hacks while the sw renderer is selected or to enable trilinear with d3d hw selected.

pcsx2/GS/Window/GSwxDialog.cpp
```diff
@@ -0,0 +1,713 @@
...
+	skip_y_spin = m_ui.addSpin(skip_box, "UserHacks_SkipDraw",        0, 10000, 0, IDC_SKIPDRAWHACK,   hacks_check);
+
+	rend_hack_choice_grid->Add(skip_box);
+
+	// Upscale Hacks:
+	m_ui.addComboBoxAndLabel(upscale_hack_choice_grid, "Half-Pixel Offset:", "UserHacks_Half_Bottom_Override", &theApp.m_gs_offset_hack, IDC_OFFSETHACK,   hacks_check);
+	m_ui.addComboBoxAndLabel(upscale_hack_choice_grid, "Round Sprite:",      "UserHacks_round_sprite_offset",  &theApp.m_gs_hack,        IDC_ROUND_SPRITE, hacks_check);
+
+	// Texture Offsets
+	add_label(this, upscale_hack_choice_grid, "Texture Offsets:", IDC_TCOFFSETX);
```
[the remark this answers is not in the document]
tellowkrinkle: Added

lightningterror (commented):
Regarding the TC Offset hack, currently it is allowed to be changed on master and not greyed out on native res, rest are.

In Debug/OGL Geometry shader, Image load store, Sparse texture should be greyed out in Null and D3D.

Cosmetics:
Maybe the upper ones width can be increased to match the lower ones
image

I guess nothing can be done with the gap on d3d blending option?
image

For TC offsets, maybe a space between X: and the button, and Y: space between both buttons.

There are some other alignments or increasing with of spin buttons that can be done too but since QT is around the corner maybe they can be left out? For example in OSD increasing the width of timeout, size, max messages to match the sliders.

pcsx2/GS/Renderers/DX11/D3D.cpp
```diff
@@ -0,0 +1,120 @@
+#include "PrecompiledHeader.h"
```
lightningterror, line 1: Missing copyright header.

pcsx2/GS/Renderers/DX11/D3D.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
```
lightningterror, line 1: Ditto.

pcsx2/GS/Window/GSwxDialog.cpp
```diff
@@ -0,0 +1,755 @@
+/*
```
lightningterror, line 1: Inconsistent copyright header, should be
https://github.com/PCSX2/pcsx2/blob/master/pcsx2/GS/GS.cpp

pcsx2/GS/Window/GSwxDialog.h
```diff
@@ -0,0 +1,216 @@
+/*
```
lightningterror, line 1: Ditto.

pcsx2/GS/Window/GSwxDialog.cpp
```diff
@@ -0,0 +1,755 @@
...
+void GSUIElementHolder::DisableAll()
+{
+	for (const UIElem& elem : m_elems)
+	{
+		if (elem.prereq)
+			elem.control->Enable(false);
+	}
+}
+
+RendererTab::RendererTab(wxWindow* parent)
```
lightningterror, line 242: Would be nice to fix codacy warning too.

lightningterror (commented):
Would be nice to fix the codacy warning, also the gl options sparse texture, geometry shader, image load store should be available on release builds, so it's better to hide the debug block portion instead of the whole tab.

refractionpcsx2:
> Would be nice to fix the codacy warning, also the gl options sparse texture, geometry shader, image load store should be available on release builds, so it's better to hide the debug block portion instead of the whole tab.

I suggest rearranging the tab if it's gonna look weird if you cut it out. Debug stuff should probably be at the bottom anyway.

pcsx2/GS/Window/GSwxDialog.cpp
```diff
@@ -0,0 +1,802 @@
...
+		m_blend_mode.second->Show();
+	}
+#endif
+}
+
+HacksTab::HacksTab(wxWindow* parent)
+	: wxPanel(parent, wxID_ANY)
+	, m_ui(this)
+{
+	int space = wxSizerFlags().Border().GetBorderInPixels();
```
lightningterror, line 343: Can be marked as const, same for others applicable.

lightningterror (approved):
Aside from the minor tweaks that can be done, looks good.

JordanTheToaster (approved):

Issue #3638, opened 2020-08-28
EE Memory Base changed, breaks compatibility with tools and cheat tables


Xeeynamo:
***PCSX2 version:***
``v1.7.0-dev-164``

***Description of the issue:***
Previous to #3523 merge, we were able to create tools that were able to attach to PCSX2 process to read and write in real time values from the EE RAM at `0x20000000`. This was very convenient and hasn't been changed for at least 10 years from what I remember. Due to that change, basically all our tools and cheat tables will break. Most of [our tools](https://github.com/xeeynamo/openkh) (and more) depends from this specific address.

**Proposed solutions:**

1. Keep the content of that pull request, but make `(uptr)MainMemory()->GetBase()` to return `0x20000000` (this is from `SysMainMemory` contructor, in `System.cpp`).

2. If the intention is to not fix it and stick with a more dynamic address solution, provide an API to return a specific mapped memory location. This can be provided as an exported function or via TCP from a specified default port.

***Last known version to work:***
Before #3523

lightningterror:
See https://forums.pcsx2.net/Thread-Is-memory-handled-differently-in-the-1-7-0-builds?pid=612571#pid612571

Xeeynamo:
> See https://forums.pcsx2.net/Thread-Is-memory-handled-differently-in-the-1-7-0-builds?pid=612571#pid612571

I am afraid that the proposed solution to implement an IPC would costs a consistent rewrite of all our existing tools, it would break all the closed-source and currently un-maintained tools and breaks for ever any Cheat Table.

lightningterror:
Anyway the changes were done to get the recompilers working on 64bit builds, a bit unfortunate that it breaks compatibility with many modding tools.

tadanokojin:
tagging @tellowkrinkle

Xeeynamo:
> Anyway the changes were done to get the recompilers working on 64bit builds, a bit unfortunate that it breaks compatibility with many modding tools.

If this will not be fixed to map it back to `0x20000000`, can we at least consider the second option I proposed? The solution that @tellowkrinkle proposed to the forum, which suggests to attach a debugger and read from `HostMemoryMap::EEmem`, might require elevated permissions, a debugger and would not work easily (or at all) on modern managed languages .

tadanokojin:
I didn't really look at it but what about #3591?

GovanifY:
[...]

PR #3644, opened 2020-08-30
Prioritise the legacy VMM address during initialisation

Xeeynamo:
This pull request aims to fix the problem described in #3638  .

The new system introduced in #3523 proposes various base addresses and check if they can be actually used by the Virtual Memory Manager. This change just proposes to try to initialise the VMM by using `0x20000000` as base address first, for ~both x64 and~ x86 builds.
Of course this does not guarantee that the legacy memory address will be used, but it will help third party software to work to new builds and to slowly migrate to an API that eventually will be exposed by the emulator.

### DIFF
--- a/pcsx2/System.cpp
+++ b/pcsx2/System.cpp
@@ -359,13 +359,19 @@ namespace HostMemoryMap {
 /// Attempts to find a spot near static variables for the main memory
 static VirtualMemoryManagerPtr makeMainMemoryManager() {
 	const wxString memoryManagerName = "Main Memory Manager";
+	
+	auto tryMakeAt = [&memoryManagerName](uptr base) -> VirtualMemoryManagerPtr {
+		auto mgr = std::make_shared<VirtualMemoryManager>(memoryManagerName, base, HostMemoryMap::Size, /*upper_bounds=*/0, /*strict=*/true);
+		if (mgr->IsOk())
+			return mgr;
+		return nullptr;
+	};
 
 	// Historically, the base address has always been 0x20000000 for x86 builds. Many cheat tables and third party
 	// tools relies to this specific address, therefore try to initialize the memory manager using this base
 	// address first. This does not guarantee it, but it prioritise the legacy memory address.
 	const uptr LegacyMemoryBase = 0x20000000U;
-	auto mgr = std::make_shared<VirtualMemoryManager>(memoryManagerName, LegacyMemoryBase, HostMemoryMap::Size, /*upper_bounds=*/0, /*strict=*/true);
-	if (mgr->IsOk())
+	if (auto mgr = tryMakeAt(LegacyMemoryBase))
 		return mgr;
 
 	// Everything looks nicer when the start of all the sections is a nice round looking number.
@@ -383,10 +389,9 @@ static VirtualMemoryManagerPtr makeMainMemoryManager() {
 			// VTLB will throw a fit if we try to put EE main memory here
 			continue;
 		}
-		auto mgr = std::make_shared<VirtualMemoryManager>(memoryManagerName, base, HostMemoryMap::Size, /*upper_bounds=*/0, /*strict=*/true);
-		if (mgr->IsOk()) {
+
+		if (auto mgr = tryMakeAt(base))
 			return mgr;
-		}
 	}
 
 	// If the above failed and it's x86-64, recompiled code is going to break!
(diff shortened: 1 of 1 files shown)


tellowkrinkle:
On Ubuntu 18.04.5 x64:
Assertion Failed: SIB target is too far away, needs an indirect register
stack trace:
stack trace screenshot

pcsx2/System.cpp
```diff
@@ -358,6 +358,18 @@ namespace HostMemoryMap {
 
 /// Attempts to find a spot near static variables for the main memory
 static VirtualMemoryManagerPtr makeMainMemoryManager() {
+	const uptr UpperBounds = 0U;
+	const bool Strict = true;
```
tellowkrinkle, line 362: I'd prefer if the fact that the value was `true` was clearly visible at the call site, not way at the top of the function.  This option is decided separately for each time we try to make a VMM, and isn't even the same for all three.

tellowkrinkle: If you want to consolidate things, make a
```cpp
auto tryMakeAt = [&memoryManagerName](uptr base) -> VirtualMemoryManagerPtr {
	std::make_shared<VirtualMemoryManager>(memoryManagerName, base, HostMemoryMap::Size, /*upper_bounds=*/0, /*strict=*/true);
	if (mgr->IsOk())
 		return mgr;
	return nullptr;
}
```
Then you can
```cpp
if (auto mgr = tryMakeAt(LegacyMemoryBase))
	return mgr;
```

Xeeynamo:
Hi @tellowkrinkle and thanks for the code review, I modified the source code as you suggested.

Since it is just refactoring, I expect that the assertion will still continue to fail. Could you provide me more details on how to reproduce this crash? I built and ran it on both Windows 10 x64 and Arch Linux with Multilib and tested with SLUS-21174. I had no problems with `master` and the change I performed.

I rebased this branch to the latest changed on `master` just in case.

Please keep me updated 🙂

tellowkrinkle:
It's caused by the `0x20000000` allocation being too far away from `SSEXYZWMask[0]` referenced in `newVif_UnpackSSE.cpp:188`

Reproducible by attempting to boot the bios

Xeeynamo:
@tellowkrinkle I was able to reproduce the crash and it is x64 specific. Therefore I added a condition to not attempt to use that "magic" address on x64 builds.

Please let me know if there are other concerns or if it is ready to be approved.

pcsx2/System.cpp
```diff
@@ -358,6 +358,27 @@ namespace HostMemoryMap {
...
+		if (mgr->IsOk())
+			return mgr;
+		return nullptr;
+	};
+
+	// Historically, the base address has always been 0x20000000 for x86 builds. Many cheat tables and third party
+	// tools relies to this specific address, therefore try to initialize the memory manager using this base
+	// address first. This does not guarantee it, but it prioritise the legacy memory address.
+	// This is only valid for x86 builds, as x64 have a high chance of failing an assertion by using this
+	// specific address. x64 builds currently uses 0x60000000.
```
tellowkrinkle, line 374: Position in x64 depends on where your linker decides to put PCSX2, and will vary by platform, so I wouldn't put it in the comments here

e.g. on my computer (macOS 10.15) it's at 0x140000000

Xeeynamo: Comment removed!

pcsx2/System.cpp
```diff
@@ -358,6 +358,27 @@ namespace HostMemoryMap {
...
+			return mgr;
+		return nullptr;
+	};
+
+	// Historically, the base address has always been 0x20000000 for x86 builds. Many cheat tables and third party
+	// tools relies to this specific address, therefore try to initialize the memory manager using this base
+	// address first. This does not guarantee it, but it prioritise the legacy memory address.
+	// This is only valid for x86 builds, as x64 have a high chance of failing an assertion by using this
+	// specific address.
+	//if (sizeof(void*) == 4)
```
tellowkrinkle, line 375: You accidentally left the `//` in

Xeeynamo: That was definitively a distraction. Thanks for the flawless code review.

tellowkrinkle (approved):
BTW just so you know, 32-bit support isn't going to stay forever, though I'm happy to keep the old memory layout as long as we continue support it

Xeeynamo:
> BTW just so you know, 32-bit support isn't going to stay forever, though I'm happy to keep the old memory layout as long as we continue support it

Yes, I imagined that 😃. But as the description of the PR says, it would be good to give a window of migration for the existing 3rd party programs to a more future-proof solution (eg. the IPC PR).
Thanks for your time to review this!

F0bes:
Closing as the next release will be 64-bit only and this will have no effect.

PR #3649, opened 2020-08-30
Change all occurrences of NULL to nullptr.

arcum42:
Just a clang-tidy sweep to modernize things to use nullptr rather than NULL.


lightningterror:
Forgot to do dx11 renderer too.

arcum42:
That'd be a limitation of the way I did it, I suppose. I had the build.sh file build a json file when compiling pcsx2, then told clang-tidy to use that json file, and run modernize-null-ptr on all the files in it. I could have done this on any of clang tidy's checks, but this is one that's all over the place and unlikely to break anything.

Linux obviously doesn't build DX11, though, so I suspect its files weren't in the json file...

arcum42:
Think I'll just hold off on this, and maybe redo it as a few prs later on...

PR #3655, opened 2020-08-31
Address filename issues relating to capturing Video (.AVI) & Audio (.WAV)

sonicfind:
Before, if you were to insert a filename in the dialog box, it would be taken as is. As a result, pcsx2 would export the avi file with a different extension or no extension althogether. This simply fixes that.
Additionally, the filename set for the AVI will be transfered over to the WAV allowing them to match up. No more overrding an alreadt existing recording.wav file.

### DIFF
--- a/plugins/GSdx/GSCapture.cpp
+++ b/plugins/GSdx/GSCapture.cpp
@@ -417,11 +417,24 @@ bool GSCapture::BeginCapture(float fps, GSVector2i recommendedResolution, float 
 
 	GSCaptureDlg dlg;
 
-	if(IDOK != dlg.DoModal()) return false;
+	if (IDOK != dlg.DoModal())
+		return false;
 
 	m_size.x = (dlg.m_width + 7) & ~7;
 	m_size.y = (dlg.m_height + 7) & ~7;
 
+	{
+		int start = dlg.m_filename.length() - 4;
+		if (start > 0)
+		{
+			std::string test = dlg.m_filename.substr(start);
+			std::transform(test.begin(), test.end(), test.begin(), tolower);
+			if (test.compare(".avi") != 0)
+				dlg.m_filename += ".avi";
+		}
+		else
+			dlg.m_filename += ".avi";
+	}
 	std::wstring fn{dlg.m_filename.begin(), dlg.m_filename.end()};
 
 	//
--- a/plugins/GSdx/GSCapture.cpp
+++ b/plugins/GSdx/GSCapture.cpp
@@ -404,7 +404,7 @@ GSCapture::~GSCapture()
 	EndCapture();
 }
 
-bool GSCapture::BeginCapture(float fps, GSVector2i recommendedResolution, float aspect)
+std::wstring* GSCapture::BeginCapture(float fps, GSVector2i recommendedResolution, float aspect)
 {
 	printf("Recommended resolution: %d x %d, DAR for muxing: %.4f\n", recommendedResolution.x, recommendedResolution.y, aspect);
 	std::lock_guard<std::recursive_mutex> lock(m_lock);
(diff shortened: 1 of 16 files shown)


xTVaser:
Pulled this down and tested it out, looks good to me:
`audio_recording.wav` and `recording.avi` saved as expected.  Even when typing in the path manually (not browsing)
image

It would probably be nice if the audio filename could just match the video file's.  But that's probably out of scope (and not necessary right now), would require GSdx to send that info to spu2-x...i think?

I did find one new bug, doesn't seem to be caused by your changes as you just changed the file name.  Seems like the audio recording's folder does not get updated.  Repro steps:
- Start capturing to folder X
- Finish capturing, you should have an audio and a video file in folder X
- Delete the audio file
- Start capturing to folder Y
- Your new video file will be in folder Y, but the new audio file will still be dumped to folder X.  Video file in folder X is untouched as expected.

sonicfind:
As long as I am using the file browser and click "save," I don't seem to get this issue. I assume that these problems stem from spu2-x defaulting to the base directory, which only gets updated on every successful browse. I think the only way to remedy this would be to return a string from the Capture dialog, probably even the address to the dialog box's string to allow use of "true if success" statements (but that's another PR in itself).

sonicfind:
Anddddd I lied. This PR babyyyyy

plugins/GSdx/GSCapture.cpp
```diff
@@ -505,7 +518,7 @@ bool GSCapture::BeginCapture(float fps, GSVector2i recommendedResolution, float
 
 	m_capturing = true;
 
-	return true;
+	return new std::wstring(fn);
```
xTVaser, line 521: This seems to only be defined in the `_WIN32` ifdef so you're getting linux compiler errors - https://travis-ci.org/github/PCSX2/pcsx2/jobs/723997550#L1077

sonicfind: ah right

sonicfind: Only one problem, I have no idea what the linux pieces are doing.

plugins/spu2-x/src/PS2E-spu2.cpp
```diff
@@ -627,12 +627,12 @@ SPU2write(u32 rmem, u16 value)
...
 // for now, pData is not used
 EXPORT_C_(int)
-SPU2setupRecording(int start, void *pData)
+SPU2setupRecording(int start, std::wstring* filename)
 {
     if (start == 0)
         RecordStop();
     else if (start == 1)
-        RecordStart();
+		RecordStart(filename);
```
lightningterror, line 635: Use spaces for indentation.

plugins/spu2-x/src/Wavedump_wav.cpp
```diff
@@ -104,18 +104,25 @@ bool WavRecordEnabled = false;
...
         ScopedLock lock(WavRecordMutex);
         safe_delete(m_wavrecord);
-        m_wavrecord = new WavOutFile("recording.wav", 48000, 16, 2);
+#ifdef _WIN32
+		if (filename)
+			m_wavrecord = new WavOutFile((*filename) + "wav", 48000, 16, 2);
+		else
+			m_wavrecord = new WavOutFile("audio_recording.wav", 48000, 16, 2);
+#elif defined(__unix__)
+		m_wavrecord = new WavOutFile("audio_recording.wav", 48000, 16, 2);
```
lightningterror, line 120: Here and the above chunk as well.

sonicfind: I'm gonna have to probably do all that formatting on a separate PR... because oh my god, so much of the spu2-x code uses spaces instead of tabs. WHY???? I'll at least get these out of the way though.

sonicfind:
Spaces applied

xTVaser (changes requested):
Video and audio recording have matching file names under the following situations:
- [x] Browsing to a folder
- [x] Typing in a path/file-name manually
- [x] Browsing to a different folder after previous recording (audio file goes where it's supposed to).

I did find one new issue though:
- If you manually type in an invalid path spu2-x will throw a "couldn't record audio to that file/path..." error as you would expect.  But GSdx doesn't seem to care and capturing still initiates.  Probably need to check for path validity in both areas, and fail the capturing process if so.

xTVaser (approved):
Latest commit resolves issue I mentioned, looks good to me.

PR #3530, opened 2020-07-09
Recording: Resolve issues which lead to recordings prematurely stopping

xTVaser:
In the old/original code, the emulators internal `g_FrameCount` was mutated and used as a means to track if a recording had reached the end or not.  This relies on `g_FrameCount` starting at `0`, and was set to `0` accordingly.  However, `g_FrameCount` is also stored along with the savestate.  Sometimes a recording's starting save-state would not contain `0` and would begin with a typically very large frame counter, exceeding the total frames in the recording, and the recording would be placed into a "stopped" state.  Example:
problem

The solution to this is simple, and seems to be an overall improvement: _stop mutating the global internal frame counter_.  The recording code now keeps track of the recording's progress with it's own separate counter, only using the `g_FrameCount` (reading) to calculate adjustments to the frame counter if a save-state is loaded mid-recording/replay (this can be used to jump to sections in a long recording / redo a section over and over)

Along with this, some smaller issues were resolved in the process that got impacted by this change:
- GS Window title updating is no longer force-updated when frame advancing.  This was unnecessary and led to flickering (best way to describe it?) in the title string.
- Logic around how frame advancing and pausing works feels a lot more simplified and explicit
- InputRecording.cpp in general was cleaned up and refactored and headers have appropriate docstrings

I've attached a file detailing the test cases I went through, I would share an example of a broken recording that shows off these problems, but it depends on a modified pcsx2 that has 128MB of RAM instead of the default 32MB.

[TEST_CASES.txt](https://github.com/PCSX2/pcsx2/files/4894039/TEST_CASES.txt)

PR #3669, opened 2020-09-04
Resolve issues around playing and creating input recordings under certain scenarios (Extended)

sonicfind:
An extended version of @xTVaser's original PR of the same name.
Original:

> These are the major issues that this PR resolves:
>
>- Forbid playing a recording that starts on a save-state if there is no game currently active
>- Previously, if a recording was attempted to be played, but failed, the menu options would still be modified as if the recording was successfully played
>- Similarly, if the dialog to play a recording was cancelled, the active recording would be cancelled which it shouldn't have been
>- Correctly pause/unpause emulation around theses scenarios (especially resuming emulation in the event of aborting playing/creating a recording)
>
>In order to accomplish some of these (especially the last one), modifications had to happen to InputRecordingFile and I took this opportunity to do a full refactoring pass / documentation. Which makes it the last recording related file that needed to be refactored since the initial spike (other files are refactored in other open PRs)

Additional changes:
- If either the Create or Play dialog boxes are open, emulation will be paused if it isn't already. It will then also resume if the user chooses to cancel either action if it wasn't manually paused beforehand.
- A running recording file will not be unloaded if they cancel out of the file browser after choosing Play in the menu. It will however unload properly and resuming paused emulation if recording tools get disabled.
- Virtual Pads now *truly* only record and replay inputs to and from Slots 1A & 2A instead of every active controller in their respective ports.
- Initially loading a recording file from power-on will now, by default, run with un-paused emulation. This was mainly to ensure that no issues arise from certain settings being paused during a boot - something I personally ran into a lot.
- A few extra simplifications in the code with opening a recording file and Virtual Pad loading.

### DIFF
--- a/pcsx2/Recording/InputRecording.cpp
+++ b/pcsx2/Recording/InputRecording.cpp
@@ -37,7 +37,6 @@ void SaveStateBase::InputRecordingFreeze()
 #ifndef DISABLE_RECORDING
 	if (g_FrameCount > 0 && IsLoading())
 	{
-		// TODO - savestates can be used to jump ahead in the movie, and currently those are flagged as an undo
 		g_InputRecordingData.IncrementUndoCount();
 	}
 #endif
--- a/pcsx2/Recording/InputRecording.cpp
+++ b/pcsx2/Recording/InputRecording.cpp
@@ -50,11 +50,6 @@ InputRecording g_InputRecording;
 void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 buf[])
 {
 	// TODO - Multi-Tap Support
-	// Only examine controllers 1 / 2
-	if (port != 0 && port != 1)
-	{
-		return;
-	}
 
 	/*
 		This appears to try to ensure that we are only paying attention
--- a/pcsx2/Recording/InputRecording.cpp
+++ b/pcsx2/Recording/InputRecording.cpp
@@ -161,37 +161,59 @@ void InputRecording::Create(wxString FileName, bool fromSaveState, wxString auth
 }
 
 // GUI Handler - Play a recording
-void InputRecording::Play(wxString FileName, bool fromSaveState)
+void InputRecording::Play(wxString fileName, bool fromSaveState)
 {
 	g_RecordingControls.Pause();
 	Stop();
 
-	if (!InputRecordingData.Open(FileName, false, false))
+	// Open the file and verify if it can be played
+	if (!InputRecordingData.OpenExisting(fileName))
 	{
+		g_RecordingControls.Unpause();
 		return;
 	}
-	if (!InputRecordingData.ReadHeaderAndCheck())
+	// Either load the savestate, or restart the game
+	if (InputRecordingData.FromSaveState())
 	{
-		recordingConLog(L"[REC]: This file is not a correct InputRecording file.\n");
-		InputRecordingData.Close();
-		return;
+		if (!CoreThread.IsOpen())
+		{
+			recordingConLog(L"[REC]: Game is not open, aborting playing input recording which starts on a save-state.\n");
+			g_RecordingControls.Unpause();
+			InputRecordingData.Close();
+			return;
+		}
+		FILE* ssFileCheck = wxFopen(InputRecordingData.GetFilename() + "_SaveState.p2s", "r");
+		if (ssFileCheck == NULL)
+		{
+			recordingConLog(wxString::Format("[REC]: Could not locate savestate file at location - %s_SaveState.p2s\n", InputRecordingData.GetFilename()));
+			g_RecordingControls.Unpause();
+			InputRecordingData.Close();
+			return;
+		}
+		fclose(ssFileCheck);
+		StateCopy_LoadFromFile(InputRecordingData.GetFilename() + "_SaveState.p2s");
 	}
-	// Check author name
+	else
+	{
+		sApp.SysExecute();
+	}
+
+	// Check if the current game matches with the one used to make the original recording
 	if (!g_Conf->CurrentIso.IsEmpty())
 	{
 		if (resolveGameName() != InputRecordingData.GetHeader().gameName)
 		{
-			recordingConLog(L"[REC]: Recording was possibly recorded on a different game.\n");
+			recordingConLog(L"[REC]: Recording was possibly constructed for a different game.\n");
 		}
 	}
 	state = INPUT_RECORDING_MODE_REPLAY;
-	recordingConLog(wxString::Format(L"[REC]: Replaying movie - [%s]\n", FileName));
+	recordingConLog(wxString::Format(L"[REC]: Replaying input recording - [%s]\n", InputRecordingData.GetFilename()));
 	recordingConLog(wxString::Format(L"[REC]: PCSX2 Version Used: %s\n", InputRecordingData.GetHeader().emu));
 	recordingConLog(wxString::Format(L"[REC]: Recording File Version: %d\n", InputRecordingData.GetHeader().version));
 	recordingConLog(wxString::Format(L"[REC]: Associated Game Name or ISO Filename: %s\n", InputRecordingData.GetHeader().gameName));
 	recordingConLog(wxString::Format(L"[REC]: Author: %s\n", InputRecordingData.GetHeader().author));
-	recordingConLog(wxString::Format(L"[REC]: MaxFrame: %d\n", InputRecordingData.GetMaxFrame()));
-	recordingConLog(wxString::Format(L"[REC]: UndoCount: %d\n", InputRecordingData.GetUndoCount()));
+	recordingConLog(wxString::Format(L"[REC]: Total Frames: %d\n", InputRecordingData.GetTotalFrames()));
+	recordingConLog(wxString::Format(L"[REC]: Undo Count: %d\n", InputRecordingData.GetUndoCount()));
 }
 
 wxString InputRecording::resolveGameName()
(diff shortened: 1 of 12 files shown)


xTVaser:
The PR being referenced - https://github.com/PCSX2/pcsx2/pull/3513 since this contains the same commits, it makes sense that this one should be the one reviewed and merged.  I'll keep mine up in the meantime since I can't commit to your branch.  But hopefully this just gets merged soon and mine can be simply closed.

pcsx2/Recording/InputRecording.cpp
```diff
@@ -50,11 +50,6 @@ InputRecording g_InputRecording;
 void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 buf[])
 {
 	// TODO - Multi-Tap Support
-	// Only examine controllers 1 / 2
-	if (port != 0 && port != 1)
-	{
-		return;
-	}
```
xTVaser, line 57: I think this check should be maintained until multi-tap support is committed.

pcsx2/Recording/InputRecordingFile.cpp
```diff
@@ -24,119 +24,158 @@
...
+		if (CoreThread.IsOpen())
+		{
+			if (open(path, true))
+			{
+				savestate.fromSavestate = true;
+				if (wxFileExists(path + "_SaveState.p2s"))
+				{
+					wxString backupPath = path + "_SaveState.p2s.bak";
+					while (wxFileExists(backupPath))
+						backupPath += ".bak";
```
xTVaser, line 147: I think this should be an `if` instead of a `while`

pcsx2/Recording/PadData.cpp
```diff
@@ -76,10 +76,6 @@ void PadData::SetNormalButtons(int port, std::vector<int> buttons)
 
 void PadData::SetNormalButton(int port, PadData_NormalButton button, int fpushed)
 {
-	if (port < 0 || 1 < port)
-	{
-		return;
-	}
```
xTVaser, line 82: Once again, this looks like content that should be left alone until multi-tap support is added.  Though I'm not too concerns about this file since I basically completely remove it with the virtualpad overhaul.  So feel free to ignore this one!

pcsx2/Recording/RecordingInputManager.cpp
```diff
@@ -32,10 +32,6 @@ RecordingInputManager::RecordingInputManager()
 
 void RecordingInputManager::ControllerInterrupt(u8 & data, u8 & port, u16 & BufCount, u8 buf[])
 {
-	if (port >= 2)
-	{
-		return;
-	}
```
xTVaser, line 38: leave alone until multi-tap?

pcsx2/gui/MainMenuClicks.cpp
```diff
@@ -893,18 +908,12 @@ void MainEmuFrame::Menu_Recording_Stop_Click(wxCommandEvent &event)
...
-		vp = wxGetApp().GetVirtualPadPtr(1);
-	}
-	if (vp != NULL)
+	switch (event.GetId() - MenuId_Recording_VirtualPad_Port0)
 	{
-		vp->Show();
+	case 0:
+		wxGetApp().GetVirtualPadPtr(0)->Show(); break;
+	case 1:
+		wxGetApp().GetVirtualPadPtr(1)->Show(); break;
```
xTVaser, line 916: just pass the result from `event.GetId() - MenuId_Recording_VirtualPad_Port0` into `GetVirtualPadPtr`.  Handling out of range values can be done in that function if needed.

xTVaser (commented):
I'll start testing now, most of my comments are just around some changes being done early.  I think the multi-tap prep work can wait for the multi-tap PR.

pcsx2/Recording/RecordingControls.cpp
```diff
@@ -110,13 +110,11 @@ void RecordingControls::TogglePause()
...
-	fFrameAdvance = true;
 }
 
 void RecordingControls::Unpause()
 {
 	fStop = false;
 	fStart = true;
-	fFrameAdvance = true;
 }
 #endif
```
xTVaser, line 120: why these changes exactly?

xTVaser:
Tested the following cases:
- [x] Forbid playing a recording that starts on a save-state if there is no game currently active
- [x] Previously, if a recording was attempted to be played, but failed, the menu options would still be modified as if the recording was successfully played
- [x] Similarly, if the dialog to play a recording was cancelled, the active recording would be cancelled which it shouldn't have been
- [x] Correctly pause/unpause emulation around theses scenarios (especially resuming emulation in the event of aborting playing/creating a recording)
- [x] If either the Create or Play dialog boxes are open, emulation will be paused if it isn't already. It will then also resume if the user chooses to cancel either action if it wasn't manually paused beforehand.
- [x] A running recording file will not be unloaded if they cancel out of the file browser after choosing Play in the menu. It will however unload properly and resuming paused emulation if recording tools get disabled.
- [x] Initially loading a recording file from power-on will now, by default, run with un-paused emulation. This was mainly to ensure that no issues arise from certain settings being paused during a boot - something I personally ran into a lot.

I'm not too concerned about the VirtualPad changes given the overhaul PR that exists, but thanks for catching the issues around it.

Overall this looks good to me, no regressions on my fixes that I could find, and no major new issues.

pcsx2/Recording/RecordingInputManager.cpp
sonicfind: The problem here is that sio.port will only ever be 0 or 1. It's when sio.slot[sio.port] is used that the values can be greater, but nothing was actually pulling that. Everything that is called in sio.cpp only ever used sio.port as the parameter, so having these here made no difference.
Addtitionally, I added a slot check in sio.cpp that makes sure to only call those functions only if it's writing controllers 1A & 2A. (https://github.com/PCSX2/pcsx2/pull/3669/commits/7ed21d4c5ae1a696b6c40f38636e43a406964bdd)

pcsx2/Recording/InputRecordingFile.cpp
sonicfind: The problem with using an if here is that if there is already a backup in place, the existing savestate would then not get backed up due to how wxCopyFile works. By using a while, it ensures that no matter what, a backup of the existing savestate is saved by adding an extra ".bak" tag.
If the behavior was instead that the existing backup was overridden, then I would be using "if," but that's sadly not the case.

pcsx2/gui/MainMenuClicks.cpp
sonicfind: Ooooo, was thinking this one over and that's actually not a bad idea. Could go well with multitap too with some bitwise work.

pcsx2/Recording/RecordingControls.cpp
sonicfind: That was something from one of your commits. (https://github.com/PCSX2/pcsx2/pull/3669/commits/85560a4bf144ccd5bb1e75c9b3bcb2817a56f4af)

pcsx2/Recording/InputRecordingFile.cpp
sonicfind: Oooooo and there's a boolean parameter I can use for overwrite. Poggers. Turning that to an if.

pcsx2/Recording/RecordingControls.cpp
xTVaser: 🤦 thank god i described that commit.

pcsx2/Recording/RecordingInputManager.cpp
xTVaser: Right, I saw that near the end of the review.  All of these checks are redundant given that top-level check, good cleanup!  Though I think in the future we'll want to shift that check back into the input recording's code, just for no reason other than reducing the recording footprint in core's code.  But a later concern as it will soon change with your multitap PR

pcsx2/Recording/InputRecordingFile.cpp
xTVaser: ah i hadn't considered that, but nice to know there's an even better workaround.  good find

xTVaser:
Latest changes to address feedback look to me.  Just a cleanup around `wxCopyFile` and replacing a switch statement, pretty confident looking at the code that it won't impact any of the functionality already tested.
👍

xTVaser (approved):

xTVaser (approved):
Issues Red discovered that may/may-not be solved in #3530 but are unrelated to this PR:
- Oddities around the title bar
- Recording stopped when playing it back
- 0 frame recordings may lead to problems
- Shutting down when recording
- Pausing the GS window leads to recording issues?  Needs more investigating
- Rebooting leads to weird behaviour, should likely stop recording as well

New issues Red found that can be resolved separately:
- Force the user to browse when creating a recording.  In the future we may add a folder like the `snaps` folder.

Issues Red found that sonic resolved that pertain to the goal of this PR:
- [x] Don't start a new recording if emulation is not running

New Issue I found:
- If recording is paused prior to disabling recording tools.  The game will remain paused, and since we unbind the recording keybindings.  You can't unpause emulation (well you can, pause/unpause the PCSX2 menu option).
  - On this topic, the PCSX2 menu option for pausing/resuming should probably match whatever playing/pausing with the recording tools does.

Previous issues re-tested, still looks good to me - https://github.com/PCSX2/pcsx2/pull/3669#issuecomment-687690770

@RedDevilus fyi, in case I missed something in my lists above.  Will use this list going forward.

In my opinion this is ready for merging, it solves a lot of problems and doesn't introduce any regressions.  Just the code is finally be exhaustively tested, can't fix them all in one PR :)

RedDevilus (approved):
Due to fixes in other PRs and the many variables, it's best that we simplify the process by merging and then fix as we go. Luckily Input Recording should not mess with users not using this function. I'll do some more re-testing and see what else comes up.

PR #3670, opened 2020-09-05
Revamp the wx spu2-x dialog.

arcum42:
Now that I've gone and spent a bunch of time on the GSdx dialog, the wx one I did for spu2-x felt like I could do it better. Here it what it currently looks like:
old-dialog

So, lets make this have a tabbed interface, combine all the extra dialogs into the main dialog, and add Ok/Cancel buttons.
spu2-x1
spu2-x2
spu2-x3

### DIFF
--- a/plugins/spu2-x/src/wx/wxConfig.cpp
+++ b/plugins/spu2-x/src/wx/wxConfig.cpp
@@ -18,139 +18,397 @@
 #include "Global.h"
 #include "wxConfig.h"
 
-Dialog::Dialog()
-    : wxDialog(nullptr, wxID_ANY, "SPU2-X Config", wxDefaultPosition, wxDefaultSize, wxCAPTION | wxCLOSE_BOX)
+MixerTab::MixerTab(wxWindow* parent)
+	: wxPanel(parent, wxID_ANY)
 {
-    m_top_box = new wxBoxSizer(wxHORIZONTAL);
-    m_left_box = new wxBoxSizer(wxVERTICAL);
-    m_right_box = new wxBoxSizer(wxVERTICAL);
+	auto* top_box = new wxBoxSizer(wxVERTICAL);
+
+	// Mixing Settings
+	top_box->Add(new wxStaticText(this, wxID_ANY, "Interpolation"), wxSizerFlags().Centre());
+
+	m_interpolation.Add("Nearest (Fastest/bad quality)");
+	m_interpolation.Add("Linear (Simple/okay sound)");
+	m_interpolation.Add("Cubic (Artificial highs)");
+	m_interpolation.Add("Hermite (Better highs)");
+	m_interpolation.Add("Catmull-Rom (PS2-like/slow)");
+
+	m_inter_select = new wxChoice(this, wxID_ANY, wxDefaultPosition, wxDefaultSize, m_interpolation);
+
+	effect_check = new wxCheckBox(this, wxID_ANY, "Disable Effects Processing (Speedup)");
+	dealias_check = new wxCheckBox(this, wxID_ANY, "Use the de-alias filter (Overemphasizes the highs) ");
+
+	// Latency Slider
+	const int min_latency = SynchMode == 0 ? LATENCY_MIN_TIMESTRETCH : LATENCY_MIN;
+
+	m_latency_box = new wxStaticBoxSizer(wxVERTICAL, this, "Latency");
+	m_latency_slider = new wxSlider(this, wxID_ANY, SndOutLatencyMS, min_latency, LATENCY_MAX, wxDefaultPosition, wxDefaultSize, wxSL_LABELS);
+	m_latency_box->Add(m_latency_slider, wxSizerFlags().Expand());
+
+	// Volume Slider
+	m_volume_box = new wxStaticBoxSizer(wxVERTICAL, this, "Volume");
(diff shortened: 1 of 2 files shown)


tellowkrinkle:
Is it possible to make this a label and not a text box?
Also it's scrollable for some reason 
image

arcum42:
@tellowkrinkle The control itself is on line 120 of wxConfig.cpp. If you fiddle with the size on that line, the scroll should go away if it's large enough (That's why the size was set. Otherwise, it was smaller and giving a scrollbar on my side, too.). There is a wxTE_NO_VSCROLL attribute you could stick in the list, but it said it was only implemented on MSW and Gtk.

I originally had it as three, very fiddly wxStaticText labels that liked to cut off the end of the text with "..." and such. I just got fed up with messing with them, and decided to use a control that was actually designed for larger amounts of text and set it as read-only.

sonicfind:
Also, it's "more than." LOL

plugins/spu2-x/src/wx/wxConfig.cpp
```diff
@@ -18,471 +18,486 @@
...
-        auto adv_dialog = new SoundtouchCfg::AdvDialog;
-        adv_dialog->Display();
-        wxDELETE(adv_dialog);
-    }
+	Update();
 }
 
-void Dialog::ResetToValues()
+Dialog::Dialog()
+	: wxDialog(nullptr, wxID_ANY, "SPU2-X Config", wxDefaultPosition, wxDefaultSize, wxCAPTION | wxCLOSE_BOX)
```
lightningterror, line 345: If we are getting rid of the gsdx branding then we might as well update this one too.

arcum42:
@sonicfind I suppose I could change the text. I'd just copied it directly from what both the earlier wx dialog and the gtk dialog before it had. Tightening it up a bit so it takes up less space might be a good idea...

@lightningterror Sure, I could do that.

arcum42:
spu2-x4
@tellowkrinkle It's back to being a label. See how that looks.
@sonicfind Majorly revised the text.
@lightningterror Changed the name.

RedDevilus:
Is that your font making SPU2 config be cut off a bit on top? That second line best aligned to center, seems a bit too aligned to the left. The buttons themselves are also close to other text. Otherwise looks good to me.

arcum42:
@RedDevilus This is wxWidgets. If things look even remotely close to where you want them, it's pretty much a miracle, and takes endless fiddling with alignment tags and things. Positioning is fiddly enough on gtk, and then wx adds its own layer on top...

And, yeah, that's probably the current theme making the font slightly cut off on the title bar.

RedDevilus:
Sorry to bother you with that, didn't think it was that fiddly. 🙇

arcum42:
Seriously, sometimes I'm staring at half a dozen lines of code with "Centre()" on them, and wondering why the bloody widgets aren't centered. Earlier the labels were centered, but only going 3/4's of the way across the dialog. There's a reason why I'm totally along with the "lets get rid of wx" train, despite also working in it a lot on pcsx2.

It is possible to absolutely position widgets (like in onepad), but that's the sort of decision you make before making the dialog, and is liable to give you dpi issues across platforms.

Part of the trouble is the stable version of wxwidgets, 3.0, is ancient. The unstable version, 3.1, I think has all sorts of fixes, including things like dpi, but the linux distros don't package it because it's "unstable". Version 3.2 is likely to come out by the end of the year, so things may improve.

Or we could rewrite everything in a different gui, which would also improve things. This rewrite gives us something to base that one on, though.

plugins/spu2-x/src/wx/wxConfig.cpp
```diff
@@ -18,471 +18,484 @@
...
-    m_sdl_box->Add(m_sdl_text, wxSizerFlags().Centre());
+	auto* adv_box = new wxStaticBoxSizer(wxVERTICAL, this, "Advanced");
 
-    for (int i = 0; i < SDL_GetNumAudioDrivers(); ++i)
-        m_sdl.Add(SDL_GetAudioDriver(i));
+	auto* babble_label = new wxStaticText(this, wxID_ANY, \
+	"For fine-tuning time stretching.\n"\
+	"Larger is better for slowdown, && smaller for speedup (60+ fps).\n"\
+	"All options in microseconds.",\
+	wxDefaultPosition, wxSize(200, 65), wxALIGN_CENTRE_HORIZONTAL);
```
tellowkrinkle, line 121: It's cutting off the last line 
image

Could you try this?
```suggestion
	wxDefaultPosition, wxDefaultSize, wxALIGN_CENTRE_HORIZONTAL);
	babble_label->Wrap(300);
```

arcum42: Here's what that winds up looking like on Linux.
spu2-x5
IIRC, this is where I was before I made those changes, with everything centered off to the left for some reason...

tellowkrinkle: It stays centered if you remove the `.Expand()` when adding it to the sizer

arcum42: Alright, got it. Committed.

arcum42:
Seems like this is about ready to me. I'll go ahead and merge it.

PR #3671, opened 2020-09-06
HostFS: free fds when shutting down vm

GovanifY:
On express demand of @tadanokojin this fixes an issue where PCSX2 was not closing file handles used by the Host Filesystem subsystem until the process was closed, which was then freed by the OS.

As the ioman HostFS is implemented under a namespace I unfortunately can't put the reset into a destructor and had to put it in the core, which, while the changes are minimal, seems pretty dirty.
We should eventually move to a class for this as it assumes an array is initd at 0 when it very possibly cannot, but this requires basically rewriting the entire HostFS codebase, which I'm not very fond of doing at the moment.

I'll let you review and try to find a solution not quite as dirty

### DIFF
--- a/pcsx2/IopBios.cpp
+++ b/pcsx2/IopBios.cpp
@@ -326,7 +326,14 @@ namespace ioman {
 	void reset()
 	{
 		for (int i = 0; i < maxfds; i++)
-			fds[i].close();
+		{
+			// WARNING: the original IopBios code assumes that arrays are initialized at 0,
+			// which is very dumb but I don't feel like writing an initializer for the whole thing so 
+			// you'll just have to read my comment and fix it when something breaks. 
+			// Please don't do a git blame I'm not the cause of this madness
+			if (fds[i])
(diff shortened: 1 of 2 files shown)


tellowkrinkle:
If you add
```cpp
constexpr filedesc(): type(FILE_FREE), file(nullptr) {}
```
to the definition of `filedesc` it will no longer assume arrays are initialized to 0

Also `filedesc::close` already checks for the `FILE_FREE` case so you don't actually have to edit the reset function, though I can understand wanting the check there anyways

GovanifY:
I must have been fairly sleepy yesterday not to have noticed the struct, thanks!
I removed my comment about array being undefined but kept my check because we are uncertain in which scope reset can be called and I believe it to be more fase this way(eg what happens if we do a quick restart? does it actually close or just reset and go back to doing its own thing?)

tadanokojin (approved):
It does the thing

PR #3678, opened 2020-09-08
GSdx: Console printf CRC fix for starter character 0.

Mrlinkwii:


plugins/GSdx/GSCrc.cpp
```diff
@@ -575,19 +575,19 @@ bool IsCrcExcluded(std::string exclusionList, uint32 crc)
...
 CRC::Game CRC::Lookup(uint32 crc)
 {
-	printf("GSdx Lookup CRC:%X\n", crc);
+	printf("GSdx Lookup CRC:%08X\n", crc);
 	if(m_map.empty())
 	{
 		std::string exclusions = theApp.GetConfigS("CrcHacksExclusions");
 		if (exclusions.length() != 0)
-			printf( "GSdx: CrcHacksExclusions: %s\n", exclusions.c_str() );
+			printf( "GSdx: CrcHacksExclusions: %08X\n", exclusions.c_str() );
```
lightningterror, line 583: This one shouldn't be changed since it reads string.

lightningterror (approved):

RedDevilus (approved):

plugins/GSdx/GSCrc.cpp
sonicfind: Couldn't you still do `%08s` to achieve the intended correct result with the string?

lightningterror: Not worth it since user puts the info in ("all" or several crc at once), see https://github.com/PCSX2/pcsx2/blob/a20b028fcbc907145f57cde37ec673da829c9f14/plugins/GSdx/GSCrc.cpp#L568

sonicfind: ah i see

Issue #3610, opened 2020-08-19
Vampire Panic - Graphical distorations with hardware renderer


ghost:
***PCSX2 version:***
both stable 1.6.0 and the latest dev build 1.7.0

***PCSX2 options:***
default options with both hardware renderers and using the "safest" preset

- **Note**: Generally it's recommended to have settings as default as possible and use the ``Safest`` preset. If there's no issue with these settings but there is an issue with some other non-default option, then please state it explicitly.

***Plugins used:***
default plugins

- **Note**: In case the issue is a graphical problem with the GSdx plugin, please provide basic information like 
I have provided a GS dump of the issue 
[gsdx_20200819232917.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/5099122/gsdx_20200819232917.gs.xz.zip)

***Description of the issue:***
distorations of the graphics.
Screenshot : 
![gsdx_20200819232917](https://user-images.githubusercontent.com/33392696/90686566-33b7f900-e274-11ea-9cc6-dd262001d6fc.png)
Also this is using the recently released english fan translation patch.
***How to reproduce the issue:***
Just play the game by using any of the hardware renderers.
The game works fine with the software renderer.

***Last known version to work:***
no idea but it happens in both 1.6.0 and latest dev build.

***PC specifications:***
CPU: Intel i7 8550U
GPU: Radeon 530.
(yeah its a laptop but it runs pretty well with it).

tadanokojin:
Ouch, looks like the game is decoding the result of a shuffle with sprites to deswizzle it.

FuxorLuck:
Here is a hack that removes this Effect

//Remove Broken Effect
patch=1,EE,203FC3E4,word,00400000

Mrlinkwii:
still an issue as of v1.7.4508
![image](https://github.com/PCSX2/pcsx2/assets/17685135/5d020ba6-465f-4c85-b35f-e7c66c020f9d)

Mrlinkwii:
is a bit changed on v1.7.4867 

![image](https://github.com/PCSX2/pcsx2/assets/17685135/4c05cc91-48b7-4409-a0ac-6f3af455bf4d)

Mrlinkwii:
still broken on v1.7.4993


![image](https://github.com/PCSX2/pcsx2/assets/17685135/4412c7c7-85c2-48fb-ae3f-6625ef3d841f)

PR #3680, opened 2020-09-09
Improved texture shuffle detection

hibye8313:
This PR is in reference to issue at https://github.com/PCSX2/pcsx2/issues/3610. tadanokojin already pointed out the problem: the game is doing a texture shuffle on its depth buffer then swizzling the result to use it as a color texture. It seems the problem was that the TextureCache was not correctly detecting textures that are cast from a depth texture (PSMZ32, PSMZ24) to a color texture (PSMCT32, PSMCT24, PSMT8, PSMT4HH, PSMT4HL) with a texture shuffle in between. Example of Vampire Panic (link above) : depth buffer starts as PSMZ24, is cast to PSMCT16 for a texture shuffle, then cast back to PSMT8H in order to use alpha byte as a palette index. Since the texture started as depth but was used as color it must be swizzled correctly or the pixels will be scrambled.

Texture cache surfaces already have a field (m_32_bits_fmt) that allows detecting if a 32 bpp texture was cast to 16 bpp for a texture shuffle then back to 32 bpp. However, it did not record if the orginal texture was depth or color. I just modified it to keep track of 3 different values (None32Bpp, Color32Bpp, Depth32Bpp) so that when cast back to 32 bpp it could be swizzled if necessary. I also added the code for swizzling for OGL and DX11 (DX11 is not tested, I am using linux right now).

I hope someone more experienced in the texture cache can give feedback on whether this is a good solution. There is a lot of code in the texture cache for special cases that I don't understand so for all I know this functionality is hidden somewhere else in the code (If it does exist I am guessing it must use GSLocalMemory to do swizzling). I just added the fixes to the path that Vampire Panic takes.

It might have been possible to implement a more efficient solution for Vampire Panic alone but since we are trying to stay away from hacks this is my attempt at a more general fix. This might need some more testing (especially DX11) to be considered acceptable. If anyone has GS dumps would be much appreciated.

Edit: There is other code in the texture cache for tracing the original format of a texture (in Surface class:  m_target, m_from_target, m_from_target_TEX0) but this does not trace the original 32 bpp texture after a texture shuffle because it gets set to the 16 bpp texture that it was cast to. So I think the m_32_bits_fmt field s needed specifically for this purpose.

Edit2: Another detail is that this issue is actually caused by the game itself doing the (un)swizzling with sprites so the swizzling in GSdx is required to cancel this out. It would be much more efficient to have detected the game's swizzle and to just replace it with a regular sprite, but I don't know if this is a bit too 'hacky' (maybe won't work in the general case, ie. if some game actually requires the texture to be scrambled)?

Edit3: Sorry, windows is not compiling. I will try and fix this soon.


gregory38:
Unlike others peoples, I think a "hack" would be a saner solution. Channel shuffle have some interesting bits to read. The effect also do a (complex) un/swizzling with sprite. I replaced all the mess with a single sprite and few lines of shader.

Anyway, just some random info, texture cache is basically an upscaled and linear view of GS memory (aka compatible with GPU). Things get awful because GS "texture/framebuffer" are fully mutable, you can read/write with any format/size.

So far, texture shuffle was basically "if the framebuffer was rendered as 32 bits but read as a 16 bits texture, then do a TS". There is no detection of 16b->32b (or I remember it wrong). And color/depth detection was useless so far. Magically with 2 TS, stuff descramble without the need of sprite un/swizzling. There is a log somewhere in the Texture Cache. Anyway, I'm bit afraid that you might trigger the new un/swizzling code in case you don't need it which will eventually break another game. It must be thoroughly tested before the code could be merged.

hibye8313:
Thanks for the explanation Gregory, I also looked at your old dev blog on texture shuffle which was very helpful.

> Unlike others peoples, I think a "hack" would be a saner solution.

Possibly, especially if other games do not benefit from the change. Of course, the goal of the HW renderer is not 100% accuracy. Would need more tests to find if other game take the same path and if it is too inefficient.

> There is no detection of 16b->32b (or I remember it wrong)

Yes, I think you're right the field Surface::m_32_bits_fmt only keeps track of 32 bpp -> 16 bpp conversion to trigger a texture shuffle. I didn't see any code to detect the opposite  but I did see a comment (maybe yours?) here (GSTextureCache::LookupSource):
```
		// Pure depth texture format will be fetched by LookupDepthSource.
		// However guess what, some games (GoW) read the depth as a standard
		// color format (instead of a depth format). All pixels are scrambled
		// (because color and depth don't have same location). They don't care
		// pixel will be several draw calls later.
		//
		// Sigh... They don't help us.
```

So maybe GoW can benefit from the depth detection in that path? Of course there is a lot of code in the texture cache that I don't understand so I would need to run more dumps to understand what they do (I'll try and read the comments also =p).

> Magically with 2 TS, stuff descramble without the need of sprite un/swizzling"

That sounds complicated. Are you saying that there is a way for games to un-swizzle 32 bpp depth -> color in 16-bpp format with a 2-pass texture shuffle and this was being emulated in the shader with HLE? If this is true then I am guessing that the reason the case for Vampire Panic was not detected was because it did the unswizzling *after* converting back to a 32 bpp format instead of in 16 bpp.

> Anyway, I'm bit afraid that you might trigger the new un/swizzling code in case you don't need it which will eventually break another game. It must be thoroughly tested before the code could be merged.

I agree, this is only tested on a single GS dump right now. I'll try and look through the Issues section and test on other dumps with texture shuffles. Anyway a hack might be a better idea. I hope it's ok to keep this PR open at the moment for other feedback or discussion.

Thank you.

tadanokojin:
> That sounds complicated. Are you saying that there is a way for games to un-swizzle 32 bpp depth -> color in 16-bpp format with a 2-pass texture shuffle and this was being emulated in the shader with HLE? If this is true then I am guessing that the reason the case for Vampire Panic was not detected was because it did the unswizzling after converting back to a 32 bpp format instead of in 16 bpp.

No, you don't need to unswizzle for shuffle. Swizzle is the entire point of the shuffle. Most games will create a copy. IE sample Z32 and write C32 if they need depth then sample 16S on the 32bit color format and write 16S. This game does something different. Instead of the copy at the start they do the shuffle on Z32 and then deswizzle it with sprites.

hibye8313:
@tadanokojin Thank you for the explanation, that clears things up a lot. I didn't realize that PSMZ24/32 could be sampled as a texture and that this could be used simply to de-swizzle the depth buffer. It make sense now why other games do not hit this code path.

Thanks both for the info, it helps cut down a lot of trial and error trying to figure out this stuff.

PR #3684, opened 2020-09-11
WIP - Spike out Github Actions

xTVaser:
Windows was fairly easy to do, especially starting from Turtleli's workflow.  Linux was a bit more of a challenge:
- Travis is using xenial, a pretty old ubuntu image at this point.  I suspect there are a lot of packages i can omit, but the ubuntu build seems to be pretty consistent now atleast.
- I wasn't confident at all about the clang build.  clang 3.8 is >4 years old or something at this point and I couldn't really get it to work at this version.  The clang build also seems inconsistent, this might be due to ccache usage:
  - Passed - https://github.com/xTVaser/pcsx2-rr/runs/1100244910?check_suite_focus=true
  - Failed over a precompiled header - https://github.com/xTVaser/pcsx2-rr/runs/1100251941?check_suite_focus=true

Putting this up to maybe get some feedback / opinions / eyes on it.

Actions will need to be enabled on the repository:
image


.github/workflows/windows-workflow.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Restore DXSDX from Cache
+        id: dxsdk-cache
+        uses: actions/cache@v2
+        with:
+          key: dxsdk-cache
+          path: DXSDK
+
+      # Download DXSDK if needed, otherwise use cache
+      # Only a few include files are necessary
+      - name: Download DXSDK
```
tadanokojin, line 69: Is this needed anymore? I thought it was dead due to win7 drop.

tadanokojin:
I enabled actions for you.

.github/workflows/windows-workflow.yml
xTVaser: I think i recall reading a message around it not being required?  I didn't attempt it, let me try.

xTVaser:
Bringing this out of draft so maybe the actions start running?  Maybe they can't start running on another repo until there is _something_ already existing.

xTVaser:
K running on the theory that something will have to be merged first before they start running for a forked PR.  https://github.com/xTVaser/pcsx2-rr/pull/31 can be used instead.  I'll re-open once the clang issue is sorted.  No point spamming commits here if they won't even show up.

PR #3686, opened 2020-09-11
Support wxGTK3 3.2 and wxQt

guoyunhe:
It is not necessary to limit wxWidgets to 3.0 because 3.2 also works fine.

It was impossible to build PCSX2 with GTK3 if it is limited to 3.0. In some Linux distributions, only wxGTK3 3.2 is available.

Also, wxQt can also be supported for KDE/LXQt users.

### DIFF
--- a/cmake/SearchForStuff.cmake
+++ b/cmake/SearchForStuff.cmake
@@ -35,9 +35,14 @@ else()
     set(wxWidgets_CONFIG_OPTIONS --unicode=yes)
 endif()
 
-list(APPEND wxWidgets_CONFIG_OPTIONS --version=3.2)
+# Use system wxWidgets version when packaging for Linux
+if(NOT PACKAGE_MODE)
+    list(APPEND wxWidgets_CONFIG_OPTIONS --version=3.0)
+endif()
 
-if(GTK3_API AND NOT APPLE)
+if(QT_API AND NOT APPLE)
+    list(APPEND wxWidgets_CONFIG_OPTIONS --toolkit=qt)
+elseif(GTK3_API AND NOT APPLE)
     list(APPEND wxWidgets_CONFIG_OPTIONS --toolkit=gtk3)
 elseif(NOT APPLE)
     list(APPEND wxWidgets_CONFIG_OPTIONS --toolkit=gtk2)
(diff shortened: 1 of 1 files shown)


arcum42:
Some distributions are only shipping with wxwidgets 3.2?

This seems highly surprising for the very simple reason that wxwidgets 3.2 isn't currently out. The current versions are the stable branch, 3.0.5, and the development branch, 3.1.4. 3.2 *is* on the horizon, and last I heard was planned to come out by the end of the year, but hasn't been released as of yet. I've been checking periodically.

I'd have to imagine that if a distribution is packaging wxwidgets 3.2, that they are using that as a name for current git versions, which I'd expect would cause confusion when it actually releases. (Or they could be packaging 3.*1*.2, I suppose.)

That said, I don't necessarily have an issue with being able to compile against unstable or git versions. I just don't know what will break with them. The plan at the moment was to test things once 3.2 actually hits, and once everything was known to be working properly, switch things over (and we'd probably want to update the Windows version at that point as well).

As far as wxQt, the impression I always got was that it didn't have everything fully implemented, and I wasn't sure if we were using anything that wasn't yet working in it.

My biggest concern, though, would be making sure that it works if 3.0 is all you have installed. And I'd prefer not to suddenly find out that some distributions don't include "/usr/bin/wx-config", just versions with the version number. Distributions package things strangely sometimes...

guoyunhe:
I changed the description. This PR is mainly for GTK3. If you want to compile PCSX2 with GTK3, you have to install wxGTK3-devel package. But wxGTK3-devel only has 3.2 version on openSUSE Tumbleweed. And I have tested it works perfectly with wxGTK3 3.2.

arcum42:
I compiled wxwidgets from the current master git branch, and this code didn't detect it. What I think I may do is not have it do a version test in wx-config, and instead check the version after it finds the default wx, and make sure it's within an acceptable version number. And likely make gtk3 default.

arcum42:
Most of this besides the wxQt support was covered in:
https://github.com/PCSX2/pcsx2/commit/540c69897e640a72fd5f3e9101c26a48eb670cd8

in different form. Closing.

PR #3687, opened 2020-09-12
Use Github Actions to Replace Travis/AppVeyor

xTVaser:
Had to make a new PR as apparently you can't re-open a PR that you've force pushed to.

See them in _action_ here - https://github.com/xTVaser/pcsx2-rr/pull/31

Currently added actions to do the following:
- Builds 32 and 64 bit on windows
  - Windows artifact is added as an artifact to the PR
- Builds 32 and 64 bit on ubuntu 20.04 (GCC 10/GTK3)
- Added a nightly cronjob to remove 30 day old artifacts

For whatever reason, the actions don't seem to run here.  I _think_ it's because there needs to be some sort of workflow in the repo, but if I'm wrong I can do a follow-up.  This same setup works in another repo :shrug:

I'd suggest that we don't jump to get rid of appveyor/travis-ci.  Let the new actions bake for a bit, see if there's issues that need to be resolved.


xTVaser:
Slight issue with my shell-script causing a failing linux build to not fail the workflow - https://github.com/xTVaser/pcsx2-rr/pull/32/checks?check_run_id=1104555102 looking into it.

PR #3690, opened 2020-09-13
Recording: Resolve issues which lead to recordings prematurely stopping (Extended)

sonicfind:
Extended version of PR #3530, just to address some additional issues.
> In the old/original code, the emulators internal `g_FrameCount` was mutated and used as a means to track if a recording had reached the end or not. This relies on `g_FrameCount` starting at `0`, and was set to `0` accordingly. However, `g_FrameCount` is also stored along with the savestate. Sometimes a recording's starting save-state would not contain `0` and would begin with a typically very large frame counter, exceeding the total frames in the recording, and the recording would be placed into a "stopped" state. Example:
> problem
> 
> The solution to this is simple, and seems to be an overall improvement: _stop mutating the global internal frame counter_. The recording code now keeps track of the recording's progress with it's own separate counter, only using the `g_FrameCount` (reading) to calculate adjustments to the frame counter if a save-state is loaded mid-recording/replay (this can be used to jump to sections in a long recording / redo a section over and over)
> 
> Along with this, some smaller issues were resolved in the process that got impacted by this change:
> 
> * GS Window title updating is no longer force-updated when frame advancing.  This was unnecessary and led to flickering (best way to describe it?) in the title string.
> * Logic around how frame advancing and pausing works feels a lot more simplified and explicit
> * InputRecording.cpp in general was cleaned up and refactored and headers have appropriate docstrings
> 
> I've attached a file detailing the test cases I went through, I would share an example of a broken recording that shows off these problems, but it depends on a modified pcsx2 that has 128MB of RAM instead of the default 32MB.
> 
> [TEST_CASES.txt](https://github.com/PCSX2/pcsx2/files/4894039/TEST_CASES.txt)

Additionally:

* Undo count will increment more accurately. Only specifically counts _recording mode_ after a savestate, and only on a previously made frame.
* Recording mode will auto-swap depending on when during a recording a boot or savestate places you: Replay - before the first frame; Record - after the last frame.


sonicfind:
One new issue I'm noticing: while unpaused and recording inputs, setting the recording to replay mode without pausing first will lead to a frame that gets skipped. If you were to immediately switch back to record mode, that frame will remain empty unless overwritten later. Hmmm...

pcsx2/Recording/InputRecording.cpp
```diff
@@ -23,31 +23,72 @@
...
+
+		if (IsLoading())
+		{
+			g_InputRecording.SetFrameCounter(g_FrameCount);
+		}
 	}
 #endif
 }
 
 #ifndef DISABLE_RECORDING
```
xTVaser, line 64: This `#if` should end after the `RecordingReset` function definition

pcsx2/Recording/InputRecording.cpp
```diff
@@ -23,31 +23,72 @@
...
+	}
+	else if (g_InputRecording.IsActive())
+	{
+		// Explicitly set the frame change tracking variable as to not
+		// detect rebooting as a frame being drawn
+		g_InputRecordingControls.SetFrameCountTracker(0);
+		g_InputRecording.SetFrameCounter(0);
+	}
+}
+
```
xTVaser, line 88: Rather than defining a function in the App.h file, why not just define it as part of `InputRecording` and call it in AppMain.cpp.  The header already is included in that file where this is presently called.

xTVaser (commented):
Initial things i noticed when messing around with this branch, I'll do a full review once the latest commits are incorporated.

pcsx2/gui/AppMain.cpp
```diff
@@ -620,7 +620,7 @@ void Pcsx2App::HandleEvent(wxEvtHandler* handler, wxEventFunction func, wxEvent&
 #ifndef DISABLE_RECORDING
 		if (g_Conf->EmuOptions.EnableRecordingTools)
 		{
-			if (g_RecordingControls.IsEmulationAndRecordingPaused())
+			if (g_InputRecordingControls.IsRecordingPaused())
```
xTVaser, line 623: Naming this function is hard because it's not really tied to whether or not _recording_ is paused (you can frame-advance and pause outside of a recording of course).  But the old name was also inaccurate, how about:
`IsEmulationPausedByInputRecordingTools`

kinda verbose but, atleast it's accurate.

pcsx2/gui/AppMain.cpp
```diff
@@ -1152,6 +1152,9 @@ void Pcsx2App::SysExecute()
 void Pcsx2App::SysExecute( CDVD_SourceType cdvdsrc, const wxString& elf_override )
 {
 	SysExecutorThread.PostEvent( new SysExecEvent_Execute(cdvdsrc, elf_override) );
+#ifndef DISABLE_RECORDING
+	RecordingReset();
+#endif
```
xTVaser, line 1157: if my suggestion is taken, i imagine this would change to `g_InputRecording.RecordingReset()`

xTVaser (changes requested):
Overall this looks really good, and as I've already tested it a few commits ago, everything seems solid.  Most of my comments are stylistic changes, but a few are actual issues / suggestions.

I'll fully retest once the review has been addressed.

pcsx2/gui/AppMain.cpp
sonicfind: We could probably get away with just IsPaused(). Adding ByInputRecordingTools is a bit redundant since it must be called by linking it with the `g_InputRecordingControls` object anyways.

pcsx2/Recording/InputRecording.cpp
```diff
@@ -78,131 +113,256 @@ void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 b
...
-		StateCopy_LoadFromFile(InputRecordingData.GetFilename() + "_SaveState.p2s");
+		initialLoad = true;
+		StateCopy_LoadFromFile(inputRecordingData.GetFilename() + "_SaveState.p2s");
 	}
 	else
 	{
-		g_RecordingControls.Unpause();
-		sApp.SysExecute();
+		initialLoad = true;
+		sApp.SysExecute(g_Conf->CdvdSource);
```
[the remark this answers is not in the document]
sonicfind: The SysExecute with the parameter was just so it calls from the correct source like all other calls to the function - it otherwise defaults to ISO.

xTVaser (approved):
This PR resolves a huge amount of issues around recording/playing back recordings.  I tested the following cases successfully:
- [x] Can play old broken recording [on startup / after playing / recording another]
- [x] Can play old not-broken recording [on startup / after playing / recording another]
- [x] Can record new recording [on startup / after playing / recording another]
- [x] Can play new recording [on startup / after playing / recording another]
- [x] Load a power-on recording after [on start-up / playing / recording] a current-frame recording
- [x] Record a power-on recording after [on start-up / playing / recording another]
- [x] Frame advance works [outside of recording or playback / during playback / during recording]
- [x] Pausing works while [outside of recording or playback / during playback / during recording]
- [x] Undocount is accurate
- [x] Can load a savestate to jump [back / ahead] in the recording, no blank frames are recorded
- [x] Can load a savestate to jump [back / ahead] while playing back
- [x] No blank frames recorded at the end of the movie

This looks good to me, and should mark the end of a _lot_ of fundamental input recording problems / our branches being tied up together. 👍

xTVaser (approved):
Testing latest commit - Made a recording around ~500 frames in length, frame-advanced half-way through, stopped the recording and played it back.  Saw no issues.

RedDevilus (approved):
Looking forward to upcoming changes.

PR #3723, opened 2020-09-20
GSdx: Purge D3D11 Software

tadanokojin:
1.6.0:

This PR:

1.6.0:

This PR:

F9 has been modified on Windows to have the following functionality:
- If the renderer in the config is D3D11, switch between D3D11 and SW
- If the renderer in the config is OGL, switch between OGL and SW
- If the renderer in the config is SW, switch between SW and the renderer returned by GetBestRenderer()

### DIFF
--- a/plugins/GSdx/GS.cpp
+++ b/plugins/GSdx/GS.cpp
@@ -57,8 +57,7 @@ static void (*s_irq)() = NULL;
 static uint8* s_basemem = NULL;
 static int s_vsync = 0;
 static bool s_exclusive = true;
-static const char *s_renderer_name = "";
-static const char *s_renderer_type = "";
+static std::string s_renderer_name;
 bool gsopen_done = false; // crash guard for GSgetTitleInfo2 and GSKeyEvent (replace with lock?)
 
 EXPORT_C_(uint32) PS2EgetLibType()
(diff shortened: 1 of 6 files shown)


atomic83GitHub:
Small nitpick, I think the "Direct 3D" name should be changed to "DirectX". Also, I am not in favor of the "null" name but that's subjective.

refractionpcsx2:
> Small nitpick, I think the "Direct 3D" name should be changed to "DirectX". Also, I am not in favor of the "null" name but that's subjective.

DirectX is the suite of API tools, Direct3D is the graphics API component

atomic83GitHub:
Oh ok, well in that case Direct3D is fine.

tadanokojin:
I went with Null because it matches the logo which uses the ∅ symbol. It's also referred internally as Null but it doesn't really matter to me, I can change it back to "None" if it bothers anyone.

RedDevilus:
It matches the plugin names for FW,USB and DEV too. Seems fine to me.

beaumanvienna:
@tadanokojin For renderer ==  "Software", is that not based on any hardware acceleration at all? What's the difference between "OpenGL" and "Software"?

refractionpcsx2:
Software mode uses an OpenGL interface to display the picture.

tadanokojin:
@beaumanvienna 
**Software:** scene rasterized on the CPU
**OpenGL/DX11:** scene rasterized on the GPU
**both:** display merger and post-process (external shader, FXAA, deinterlace) performed on the GPU

PR #3726, opened 2020-09-20
Better Apple icon file

MaddTheSane:
The new icns file contains the following enhancements:
* 256 px taken from the Windows icon.
* Created new, better-looking, 128 px by scaling down 256 px (the old version looked like it was scaled up).
* Add retina versions of 16 and 128 px (taken from the 32 and 256 px versions, respectively).


tellowkrinkle:
Ahh nice, didn't realize the windows icon was from there

(I got the previous one from the files in `pcsx2/pcsx2/gui/Resources/`, where the largest image was 64x64)

PR #3732, opened 2020-09-21
Capture Optimizations and Input Recording Interactivity

### LINKED ISSUE #3729: [Linux] screenrecorder option does nothing

sonicfind:
Video capture will now end if an input recording reaches its final frame in Replay mode.
Additionally, should fix some oversights with filenames transferring to Spu2, especially on Linux.
Directly related to xTVaser#14
Possibly fixes #3729

### DIFF
--- a/common/include/PS2Edefs.h
+++ b/common/include/PS2Edefs.h
@@ -172,7 +172,7 @@ void CALLBACK GSsetFrameSkip(int frameskip);
 
 // Starts recording GS frame data
 // returns a non zero value if successful
-int CALLBACK GSsetupRecording(std::wstring& filename);
+int CALLBACK GSsetupRecording(std::string& filename);
 
 // Stops recording GS frame data
 void CALLBACK GSendRecording();
@@ -269,7 +269,7 @@ typedef void(CALLBACK *_GSsetGameCRC)(int, int);
 typedef void(CALLBACK *_GSsetFrameSkip)(int frameskip);
 typedef void(CALLBACK *_GSsetVsync)(int enabled);
 typedef void(CALLBACK *_GSsetExclusive)(int isExclusive);
-typedef int(CALLBACK* _GSsetupRecording)(std::wstring&);
+typedef int(CALLBACK* _GSsetupRecording)(std::string&);
 typedef void(CALLBACK* _GSendRecording)();
 typedef void(CALLBACK *_GSreset)();
 typedef void(CALLBACK *_GSwriteCSR)(u32 value);
--- a/common/include/PS2Edefs.h
+++ b/common/include/PS2Edefs.h
@@ -173,7 +173,7 @@ void CALLBACK GSsetFrameSkip(int frameskip);
 // if start is 1, starts recording spu2 data, else stops
 // returns a non zero value if successful
 // for now, pData is not used
-int CALLBACK GSsetupRecording(int start, void *pData);
+std::wstring* CALLBACK GSsetupRecording(int start);
 
 void CALLBACK GSreset();
 //deprecated: GSgetTitleInfo was used in PCSX2 but no plugin supported it prior to r4070:
--- a/common/include/PS2Edefs.h
+++ b/common/include/PS2Edefs.h
@@ -170,10 +170,12 @@ void CALLBACK GSsetGameCRC(int crc, int gameoptions);
 // controls frame skipping in the GS, if this routine isn't present, frame skipping won't be done
 void CALLBACK GSsetFrameSkip(int frameskip);
 
-// if start is 1, starts recording spu2 data, else stops
+// Starts recording GS frame data
 // returns a non zero value if successful
-// for now, pData is not used
-std::wstring* CALLBACK GSsetupRecording(int start);
+int CALLBACK GSsetupRecording(std::wstring& filename);
+
+// Stops recording GS frame data
+void CALLBACK GSendRecording();
 
 void CALLBACK GSreset();
 //deprecated: GSgetTitleInfo was used in PCSX2 but no plugin supported it prior to r4070:
@@ -267,7 +269,8 @@ typedef void(CALLBACK *_GSsetGameCRC)(int, int);
 typedef void(CALLBACK *_GSsetFrameSkip)(int frameskip);
 typedef void(CALLBACK *_GSsetVsync)(int enabled);
 typedef void(CALLBACK *_GSsetExclusive)(int isExclusive);
-typedef std::wstring*(CALLBACK *_GSsetupRecording)(int);
+typedef int(CALLBACK* _GSsetupRecording)(std::wstring&);
+typedef void(CALLBACK* _GSendRecording)();
 typedef void(CALLBACK *_GSreset)();
 typedef void(CALLBACK *_GSwriteCSR)(u32 value);
 typedef bool(CALLBACK *_GSmakeSnapshot)(const char *path);
@@ -316,6 +319,7 @@ extern _GSsetGameCRC GSsetGameCRC;
 extern _GSsetFrameSkip GSsetFrameSkip;
 extern _GSsetVsync GSsetVsync;
 extern _GSsetupRecording GSsetupRecording;
+extern _GSendRecording GSendRecording;
 extern _GSreset GSreset;
 extern _GSwriteCSR GSwriteCSR;
 #endif
--- a/pcsx2/Recording/InputRecordingControls.cpp
+++ b/pcsx2/Recording/InputRecordingControls.cpp
@@ -20,7 +20,7 @@
 #include "App.h"
 #include "Counters.h"
 #include "DebugTools/Debug.h"
-#include "GSFrame.h"
+#include "MainFrame.h"
 #include "MemoryTypes.h"
 
 #include "InputRecording.h"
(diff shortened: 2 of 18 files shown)


xTVaser:
Old related issue - https://github.com/xTVaser/pcsx2-rr/issues/14

pcsx2/gui/GlobalCommands.cpp
```diff
@@ -444,10 +444,14 @@ namespace Implementations
...
-		g_Pcsx2Recording ^= 1;
+		if (wxGetApp().HasGUI())
+		{
+			sMainFrame.VideoCaptureToggle();
+			return;
+		}
 
 		GetMTGS().WaitGS(); // make sure GS is in sync with the audio stream when we start.
-		if (g_Pcsx2Recording)
+		if (g_Pcsx2Recording ^= 1)
```
xTVaser, line 454: can we change this to `g_Pcsx2Recording = !g_Pcsx2Recording` at the least to get rid of the bitwise fanciness.  Also perhaps it might be a good idea to keep it outside of the `if`, not sure how the rest of the pcsx2 devs feel about it, but assignment in the if clause for me isn't the prettiest thing

pcsx2/gui/MainFrame.cpp
```diff
@@ -277,8 +277,8 @@ void MainEmuFrame::ConnectMenus()
 	Bind(wxEVT_MENU, &MainEmuFrame::Menu_Debug_Open_Click, this, MenuId_Debug_Open);
 
 	// Capture
-	Bind(wxEVT_MENU, &MainEmuFrame::Menu_Capture_Video_Record_Click, this, MenuId_Capture_Video_Record);
-	Bind(wxEVT_MENU, &MainEmuFrame::Menu_Capture_Video_Stop_Click, this, MenuId_Capture_Video_Stop);
+	Bind(wxEVT_MENU, &MainEmuFrame::Menu_Capture_Video_Capturing_Click, this, MenuId_Capture_Video_Record);
+	Bind(wxEVT_MENU, &MainEmuFrame::Menu_Capture_Video_Capturing_Click, this, MenuId_Capture_Video_Stop);
```
xTVaser, line 281: maybe change the MainEmuFrame function naming to `Menu_Capture_Video_Toggle_Capture_Click` since there is now 1 function to rule them all

pcsx2/Recording/InputRecordingControls.cpp
```diff
@@ -201,4 +206,21 @@ void InputRecordingControls::Lock(u32 frame)
...
+void InputRecordingControls::EndVideoCapture() const
+{
+	if (MainEmuFrame* mainFrame = GetMainFramePtr())
+	{
+		if (mainFrame->IsCapturing())
+		{
+			mainFrame->VideoCaptureToggle();
+			recordingConLog(L"Video Capture Completed\n");
+		}
+	}
```
xTVaser, line 224: could be combined into a single `if`

pcsx2/gui/MainFrame.cpp
sonicfind: True

pcsx2/gui/GlobalCommands.cpp
sonicfind: Sure, since it won't really change performance. Was just a stylistic choice here.

pcsx2/Recording/InputRecordingControls.h
```diff
@@ -80,6 +80,8 @@ class InputRecordingControls
 	bool switchToReplay = false;
 	// Used to stop recording frames from incrementing during a reset
 	bool frameLock = false;
+	bool IsReplayingAtEnd() const;
```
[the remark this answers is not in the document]
sonicfind: **Don't mind me. Just banging my head against the wall for** _not thinking of such an obvious name._ AAAAAA

pcsx2/Recording/InputRecordingControls.cpp
```diff
@@ -79,9 +79,12 @@ void InputRecordingControls::HandleFrameAdvanceAndPausing()
 		switchToReplay = false;
 	}
 
-	if ((g_InputRecording.IsReplaying() && g_InputRecording.GetFrameCounter() >= g_InputRecording.GetInputRecordingData().GetTotalFrames())
-		|| g_InputRecording.GetFrameCounter() == INT_MAX)
+	bool endCapture = false;
+	if (IsReplayingAtEnd() || g_InputRecording.GetFrameCounter() == INT_MAX)
+	{
 		pauseEmulation = true;
+		endCapture = true;
```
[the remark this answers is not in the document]
sonicfind: It's been a minute since I touched this, but I believe it was a matter of making sure to wait until emulation is paused before calling for the capturing to stop. All other called to the function are in situations where emulation is already pre-paused.

pcsx2/Recording/InputRecordingControls.cpp
sonicfind: This style of nesting the second condition after doing something similar to `MainEmuFrame* mainFrame = GetMainFramePtr()` matches other iterations in the solution. Seems a bit easier to follow readability-wise as well.

pcsx2/Recording/InputRecordingControls.cpp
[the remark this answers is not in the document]
xTVaser: I wondered if that was the case, and yeah I later noticed the other functions setting this bool so probably nothing that needs to change. 👍

pcsx2/Recording/InputRecordingControls.cpp
xTVaser: oh my mistake, i didn't notice it was doing assignment in the if statement.  This could be minimized down using the `sMainFrame` macro I believe.  Something like `if (sMainFrame.IsCapturing())`.  But I don't have a strong opinion this _should_ also safeguard against an uninitialized mainframe.

sonicfind: That was my first attempt. Didn't work so I had to do this.

xTVaser:
Tried to test this, when attempting to starting recording, I hit an exception, seems to be around the changes in wavedump_wav.cpp.  Here's what i think is the relevant part of the stacktrace:
```
 	pcsx2-dbg.exe!wxString::wxString(const std::wstring & str) Line 1317	C++
>	pcsx2-dbg.exe!RecordStart(std::wstring * filename) Line 122	C++
 	pcsx2-dbg.exe!SPU2setupRecording(int start, std::wstring * filename) Line 599	C++
 	pcsx2-dbg.exe!MainEmuFrame::VideoCaptureToggle() Line 867	C++
 	pcsx2-dbg.exe!MainEmuFrame::Menu_Capture_Video_ToggleCapture_Click(wxCommandEvent & event) Line 842	C++

```

sonicfind:
@xTVaser Are you able to reproduce this on my just updated commits? I can't seem to get this to happen and just want to be sure.

xTVaser:
the GS plugin is returning the right value:
image
But the way it is used in MainMenuClicks is problematic:
image

I think this problem is due to returning a pointer to a local variable from a function, which it appears to be.  But I don't know why this wasn't an issue in your previous PR which refactored to this style https://github.com/PCSX2/pcsx2/commit/1ff67c6c1bbc618f6b37018cc3738fde1855606a#diff-5fa8a882d01c9105bf7f5c58c525f660

sonicfind:
Ohhhh I see, what's happening. I remember something similar happening... when the GSdx and Spu2 plugins didn't match. I'll need to add a try:catch block here for that.

sonicfind:
As a double check, I'm also not getting this exception when using the artifact builds.

xTVaser:
I did multiple complete cleans, so my plugins are freshly built on the installation.

> Also, to be more accurate, it's not that it's returning an address to a local variable, it's creating a completely new variable with the same string inside it. It's just that in this case, for some reason, the GSdx plugin that you're using is still return an int instead of a string.

I don't think that is true...from the first screenshot above you can see that `GSSetupRecording` is returning a string, but once its set to the local var in `MainMenuClicks` / we continue beyond the function call, the pointer no longer points to a valid string.

sonicfind:
Mind testing what happens if you do
```
std::wstring* filename = GSSetupRecording(true);
if (filename)
```
directly?

xTVaser:
I already tested that, same result

sonicfind:
Screw it, we're going in.

xTVaser (approved):
Successfully worked for me, was able to capture/stop capture a recording normally.

And then I tried with an input recording, and it stopped the capture process automatically once it finished.

image

xTVaser (approved):
Repeated test, still works on latest rebased commits.
Would still be nice to see if this fixes screen recording on linux, but otherwise this looks good to me.

common/include/PS2Edefs.h
```diff
@@ -181,10 +181,12 @@ void CALLBACK GSsetGameCRC(int crc, int gameoptions);
 // controls frame skipping in the GS, if this routine isn't present, frame skipping won't be done
 void CALLBACK GSsetFrameSkip(int frameskip);
 
-// if start is 1, starts recording spu2 data, else stops
+// Starts recording GS frame data
 // returns a non zero value if successful
-// for now, pData is not used
-int CALLBACK GSsetupRecording(int start, void *pData);
+int CALLBACK GSsetupRecording(std::wstring& filename);
```
tadanokojin, line 186: I'm concerned about the usage of wstring as outlined here:
https://stackoverflow.com/a/50407375

This goes for your other usages of wstring as well. It's also inconsistent with the current API used by the OSD.

PR #3734, opened 2020-09-21
Updated GameDB.

YukiXXL:
Added fixes for Syphon Filter The Omega Strain Public Beta 1 and 2.

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -5569,7 +5569,7 @@ Region = NTSC-U
 EETimingHack = 1 // Fixes random hangs.
 [patches = C909A32E]
 	author=Refraction & YukiXXL
-	//Flashlight and Nightvision not working propetly fix.
+	// Flashlight and Nightvision not working propetly fix.
 	patch=0,EE,00367798,word,48438000
 	patch=0,EE,0036779C,word,4B06521B
 	//Offline button fix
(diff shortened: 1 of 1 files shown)


bin/GameIndex.dbf
```diff
@@ -5563,6 +5563,19 @@ Serial = SCUS-97396
...
+Serial = SCUS-97397
+Name   = Syphon Filter - The Omega Strain Public Beta 1
+Region = NTSC-U
+EETimingHack = 1 // Fixes random hangs.
+[patches = C909A32E]
+	author=Refraction & YukiXXL
+	//COP2 fix.
+	patch=0,EE,00367798,word,48438000
+	patch=0,EE,0036779C,word,4B06521B
+	//Offline button fix
```
lightningterror, line 5575: Maybe format the comments better so they match the rest of the file.

YukiXXL: Huh?

tlbjr: The one existing example I looked up had a space after `//` and before the actual comment.
```
// COP2 fix.
```

atomic83GitHub: Well, "COP2 fix" do not explain much what is going on here, I think you should write more about why does it need a patch, don't hesitate to contact the team as someone will probably know that.

YukiXXL:
I don't know what I did wrong or what needs to be done... I'am lost O.o

lightningterror:
+45,201 −45,180 changes, that doesn't look right.

lightningterror (changes requested):
It's showing changes from other commits, I suggest you rebase with current master and then add your changes.

lightningterror:
I'll close this since it's not being worked on, maybe someone else can add the patches instead.

PR #3739, opened 2020-09-22
gsdx-ogl: Swap DATE_GL42 with DATE_GL45 on sw blending draw.

### LINKED ISSUE #3738: Gran Turismo 4: Flickering car reflections with OGL HW
***PCSX2 version:***
v1.7.0-dev-328-g284977605

***PCSX2 options:***
Default except EE clamping to Full and VU clamping to Extra + Preserve Sign 

Reflective surfaces on car tend to flicker at random, even the player car hood if using the hood camera.
Reflective surfaces on buildings don't seem affected.
The issue also happens at native resolution.

Of all the settings I tried the following affects the issue:
-Blending Accuracy to High-Ultra coupled with Accurate DATE set to Fast or Off, Full Accurate DATE will cause the issue to reappear regardless of blending accuracy level and High blending accuracy causes severe performance issues at various point of the game (sub 20fps with GPU pegged at 100% usage)
-Turning GIF FIFO on , which instead breaks some GUI elements by flashing random garbage textures in their place
-Using the EE Timing Hack, which doesn't seem to break anything obvious.

Here's two multiframe GS dump taken while the issue was happening, one while paused the other while in a race.

[gsdx_20200922134636.gs.zip](https://github.com/PCSX2/pcsx2/files/5261416/gsdx_20200922134636.gs.zip)
[gsdx_20200922135629.gs.zip](https://github.com/PCSX2/pcsx2/files/5261425/gsdx_20200922135629.gs.zip)

Start any race with a rolling start while using the OGL HW renderer and blending set to basic or higher.
Different cars tend to have varying degree of said issue

None that I could find that didn't have other issues masking that one

***PC specifications:***
Windows 10 1909 64bit
Ryzen 7 3800X
Nvidia GTX 1080Ti w/ 451.85 drivers

lightningterror:
It will allow to run sw blending with DATE draw which was previously
DATE_GL42 by default.

Fix #3738

### DIFF
--- a/plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
+++ b/plugins/GSdx/Renderers/OpenGL/GSRendererOGL.cpp
@@ -446,7 +446,7 @@ void GSRendererOGL::EmulateChannelShuffle(GSTexture** rt, const GSTextureCache::
 	}
 }
 
-void GSRendererOGL::EmulateBlending(bool DATE_GL42)
+void GSRendererOGL::EmulateBlending(bool& DATE_GL42, bool& DATE_GL45)
 {
 	GSDeviceOGL* dev         = (GSDeviceOGL*)m_dev;
 	const GIFRegALPHA& ALPHA = m_context->ALPHA;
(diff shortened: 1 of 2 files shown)


sonicfind:
Crashes whenever I attempt to load a game in SW rendering or switch to SW rendering midgame. Checked with master, no crashing there.

PR #3741, opened 2020-09-23
ci: Create AppImage artifacts for Linux Builds

xTVaser:
This aims to provide quick, testable artifacts for linux in much the same way as what is currently available for Windows builds.
- Provides a way for linux users who are not familiar with compiling from source to test PRs and devbuilds (master commits)
- Facilitates a faster testing workflow, as not every PR needs to be tested end-to-end (installing packages > etc)

However, this last point is important to keep in mind as unless going forward PCSX2 on linux will be distributed solely with AppImages, it cannot be the _sole way to test pcsx2 on linux_.  AppImages are not concerned with dependencies that may be optimized for a particular distro, they also have an exclusion list, of which a handful of packages that pcsx2 depends on are included https://github.com/AppImage/pkg2appimage/blob/master/excludelist

This means that we should still likely build `.deb` and `.rpm` artifacts to support testing how the end-user is going to install and use pcsx2 to safeguard against the scenario where a change breaks the app on a particular distro.

It also means that even if you are using the AppImage, you still need to ensure that your environment has the necessary packages installed that won't come with them.  These are documented in the wiki.

## Main caveats:
- It's an AppImage within a tar within a zip, because github actions artifacts don't preserve file permissions (AppImage's need to be executable / allows you to just double-click it)...this is actually their recommended method...hopefully that improves lol
- Can't use portable mode because the AppImage "filesystem" is read-only and pcsx2 tries to create folders and such / write logs, etc.


tellowkrinkle:
You should be able to compile all GSdx variants by doing `-DDISABLE_ADVANCE_SIMD=ON`

xTVaser:
Opening this up for review as it's worked on atleast one other person's environment successfully.  So as far as I can tell, the AppImage is being constructed properly.

.github/workflows/scripts/build-linux.sh
```diff
@@ -99,16 +125,65 @@ compile() {
...
+	mkdir -p ./.appdir/usr/bin/
+	cp -R ./bin/docs ./.appdir/usr/bin/
+	cp -R ./bin/Langs ./.appdir/usr/bin/
+	cp -R ./bin/shaders ./.appdir/usr/bin/
+	# Assumes plugins have been built
+	cp -R ./bin/plugins ./.appdir/usr/bin/
+	cp ./bin/cheats_ws.zip ./.appdir/usr/bin/
+	cp ./bin/GameIndex.dbf ./.appdir/usr/bin/
+	# NOTE - we cannot distribute a portable AppImage because then it tries to
+	# write to it's read-only file-system
```
NexisHunter, line 151: :wave: According to the [documentation](https://github.com/AppImage/AppImageKit#special-directories) for special directories, it may be possible to provide these as an external directory allowing for portability.

Maybe having something like `PCSX2.config` could be used to store the ini file? Granted this requires some prior knowledge to ensure those files are created.

Also this potentially allows to have the plugins side loaded in a similar manner, thus allowing for custom plugins to be loaded (Unsure if this is currently an option).

xTVaser: 👋 That's good to know, sounds like we can make an AppImage with portable mode but might take some doing.  There's a lot in flux that might make this difficult (you mentioned plugins, they are currently being incorporated into the core as we speak).  But also the configuration setup in general may change.

However the big unknown to me is how pcsx2 determines the folder paths, I assume it just uses the binary's location but I'm not 100% sure.  If it does, there will likely need to be a code change to allow the `portable.ini` file specify a location.  Or maybe the AppImage documentation you reference is enough (needs to be tested).

qurious-pixel: hi @xTVaser, the AppImage uses the AppDir as root </> and then all folders are relative to this. 
Like in a Linux FileSystem /usr/bin would be AppDir/usr/bin. 
By default on first run, PCSX2 will write to the users actual ~/.config/PCSX2. Those files will be read/write.
AppImages respect creating a custom user home or user config by creating a folder with the same name and append .home or .config to the folder.
`PCSX2.AppImage` would be the app, and if found in the same directory `PCSX2.AppImage.config` all the writeable folders/files will go there. This is how I create a portable PCSX2. 

AppImages can also make use of an updater, so I have PCSX2 update when I pull new images (weekly)

Also, it's bad practice to zip or tar the AppImage. [AppImage Packaging Guide](https://docs.appimage.org/packaging-guide/distribution.html#do-not-put-appimages-into-other-archives)

xTVaser:
superseeded by https://github.com/PCSX2/pcsx2/pull/4103

PR #3742, opened 2020-09-23
nsis: Bring the installer up to date

lightningterror:
Will need updating again down the line but this matches the current work that's been done.


nsis/SectionVersionCheck.nsh
```diff
@@ -7,17 +7,17 @@ Function UninstallPrevious
...
 ${If} $R1 != ""
 ${AndIf} $R2 != ""
     Goto UserPrompt
 ${EndIf}
 
-; Search for 1.4.0
-ReadRegStr $R3 HKLM "${INSTDIR_REG_KEY}" "Uninst-pcsx2 1.4.0Directory"
-${StrContains} "$2" "1.4.0" "$R3"
+; Search for previous/older version
+ReadRegStr $R3 HKLM "${INSTDIR_REG_KEY}" "Uninst-pcsx2 1.6.0Directory"
```
MrCK1, line 19: This block of code is actually no longer necessary and won't work with any version past 1.4 - it can be removed completely unless we want to detect 1.4 installations.

 The uninstall path no longer contains a version number as it reduces the chance of possible bugs, ect. The "primary verison check" is compatible with all future PCSX2 versions going forward as it detects the presence of the new version string and an install path.

nsis/SectionVersionCheck.nsh
```diff
@@ -42,8 +42,8 @@ ${If} $R1 != ""
 ${EndIf}
 
 ${If} $R3 != ""
-    ${AndIf} $2 == "1.4.0"
-    Goto Exec1.4.0
+    ${AndIf} $2 == "1.6.0"
+    Goto ExecPreviousVersion
```
MrCK1, line 46: Refer to previous comment

nsis/SectionVersionCheck.nsh
```diff
@@ -53,11 +53,11 @@ ExecNormal:
...
 
-Exec1.4.0:
+ExecPreviousVersion:
     SetOutPath "$TEMP"
-    CopyFiles /SILENT /FILESONLY "$R3\Uninst-pcsx2 1.4.0.exe" "$TEMP"
-        ExecWait '"$TEMP\Uninst-pcsx2 1.4.0.exe" /S _?=$R3'
-        Delete "$TEMP\Uninst-pcsx2 1.4.0.exe"
+    CopyFiles /SILENT /FILESONLY "$R3\Uninst-pcsx2 1.6.0.exe" "$TEMP"
+        ExecWait '"$TEMP\Uninst-pcsx2 1.6.0.exe" /S _?=$R3'
+        Delete "$TEMP\Uninst-pcsx2 1.6.0.exe"
```
MrCK1, line 60: Refer to previous comment

nsis/SharedCore.nsh
```diff
@@ -19,24 +19,10 @@ Function RedistInstallation
 ; Check if the VC runtimes are installed
 ReadRegDword $R5 HKLM "SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x86" "Installed"
 
-${If} $R5 == "1"
-    Goto DxSetup
-${EndIf}
```
MrCK1, line 24: Replace with a `return` when `$R5` is present, it will skip the VC runtime from being downloaded/installed.

nsis/SharedDefs.nsh
```diff
@@ -17,7 +17,7 @@ Var UserPrivileges
 Var IsAdmin
 
 !ifndef APP_VERSION
-  !define APP_VERSION      "1.6.0"
+  !define APP_VERSION      "1.8.0"
```
MrCK1, line 20: For 1.8, the OS version check can be updated to reflect the minimum requirement of Windows 8.1 or greater - https://github.com/PCSX2/pcsx2/pull/3742/files#diff-25f3166319cb3cfbcbc8aed25acc2db8R45

MrCK1 (changes requested):

MrCK1 (approved):

PR #3744, opened 2020-09-23
gsdx-debug: Update debug logs a bit in renderer hw/gl/dx.

lightningterror:
Add prefixes to logs, clean up some stuff.


bositman:
I'm not sure if doing stuff like Channel Shuffle ->CS would benefit new people from actually knowing what that is..is there a reason for adding abbreviations there?

lightningterror:
A bit more consistent logging naming that can follow some format, regarding CS you're right so maybe I'll use full naming `Channel shuffle: ...`

lightningterror:
Trivial pr/changes, will do something in the future.

PR #3746, opened 2020-09-24
Bump up C++ version to c++17 for gcc & clang

AmFobes:
_This does not implement c++17 std::filesystem!_

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ include(Pcsx2Utils)
 
 check_no_parenthesis_in_path()
 detectOperatingSystem()
-check_compiler_version("4.8" "4.8")
+check_compiler_version("7.0" "7.0")
 
 #-------------------------------------------------------------------------------
 # Include specific module
--- a/cmake/BuildParameters.cmake
+++ b/cmake/BuildParameters.cmake
@@ -391,7 +391,7 @@ endif()
 # Note: -DGTK_DISABLE_DEPRECATED can be used to test a build without gtk deprecated feature. It could be useful to port to a newer API
 set(DEFAULT_GCC_FLAG "${ARCH_FLAG} ${COMMON_FLAG} ${DEFAULT_WARNINGS} ${AGGRESSIVE_WARNING} ${HARDENING_FLAG} ${DEBUG_FLAG} ${ASAN_FLAG} ${OPTIMIZATION_FLAG} ${LTO_FLAGS} ${PGO_FLAGS} ${PLUGIN_SUPPORT}")
 # c++ only flags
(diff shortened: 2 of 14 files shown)


tellowkrinkle:
Personally I feel like we should be switching over to `target_compile_features(target PRIVATE cxx_std_17)` for requesting a c++ standard

AmFobes:
> Personally I feel like we should be switching over to `target_compile_features(target PRIVATE cxx_std_17)` for requesting a c++ standard
You want to set that for each target or is there a way to globally set that project wide?

tellowkrinkle:
I would say per target, but there is `CMAKE_CXX_STANDARD` which is global

orbea:
Can this be postponed till after `1.7.0` is released?

AmFobes:
What's the benefit of waiting?

orbea:
I just want at least one 64-bit compatible release that doesn't require newer compilers.

AmFobes:
Alright well windows is already using the newer standard. It's not my call anyways.

gregory38:
Release won't happen tomorrow. 2-3 years old compiler should be fine. It is "recent" but not bleeding edge.

orbea:
2-3 years is still relatively modern when you consider older distros...

gregory38:
Debian stable has gcc 8 already. And next year it will become old stable. We won't lag dev for distribution that have 3+ years old toolchain. They can still keep 1.6 (which is already a damn good emulator).

orbea:
> They can still keep 1.6 (which is already a damn good emulator).

Which still requires multilib and I do not understand how using an older C++ standard for a little longer is holding back development, but I suppose it doesn't matter to anyone else I'm stuck supporting gcc-5.5.0...

refractionpcsx2:
We have some coding going on in the background which uses C++17 and will probably be PR'd in the next couple of weeks.  Holding off until then won't be of any benefit for you.

If you have an outdated compiler, then use an older version, we aren't going to hold back progress and not use features because you don't keep your system up to date.

gregory38:
Besides there is no issue to use multilib. It just takes some extra megabytes. Even ssd GB are pretty cheap nowadays. 64 bits won't bring you any extra performance anyway.

orbea:
> Besides there is no issue to use multilib.

You vastly underestimate the extra burden multilib has on package maintainers.

kenshen112:
Maintainer of Arches pcsx2-git here. There's a burden?

AmFobes:
Alright ladies and gentlemen I've decided to not be lazy. 
I have set the c++17 standard for PCSX2 and the relevant plugins individually.

tadanokojin (approved):

lightningterror (approved):

Issue #3735, opened 2020-09-22
Frame advance doesn't release held buttons


InfamousKnight:
***PCSX2 version:***

79f2468

If you hold frame advance and a key like left and then let go, it will keep pressing left even though you're not physically touching it.

To work around it, let go of frame advance, press some random key(preferable one that doesn't do anything for the time), and press frame advance once. This is pretty annoying to work with. Thankfully I figured out the real issue.

MrCK1:
Closing due to no response

PR #3752, opened 2020-09-25
recording: Don't reload cheats/settings on every frame-advance

xTVaser:
I believe the intention of the cheats/settings reloading is for the conventional pause/resume from the MainFrame's menu.

However, with input recording, nothing ever actually gets closed so I doubt the settings need to be reloaded.  These changes stop that from happening on every individual frame advance.  

However, in case im wrong and it is needed, they do get re-applied when emulation is fully resumed (Shift+P by default).  This just dramatically reduces the console spam.

Now:
image

Before:
image

fyi - @sonicfind

### DIFF
--- a/pcsx2/Recording/InputRecordingControls.h
+++ b/pcsx2/Recording/InputRecordingControls.h
@@ -39,6 +39,8 @@ public:
 	
 	// Resume emulation (incase the emulation is currently paused) and pause after a single frame has passed
 	void FrameAdvance();
+	// Returns true if emulation is currently set up to frame advance.
+	bool IsFrameAdvancing();
 	// Returns true if the input recording has been paused, which can occur:
 	// - After a single frame has passed after InputRecordingControls::FrameAdvance
 	// - Explicitly paused via an InputRecordingControls function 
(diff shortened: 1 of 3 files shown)


pcsx2/Recording/InputRecordingControls.h
```diff
@@ -50,7 +52,7 @@ class InputRecordingControls
 	void Resume();
 	void SetFrameCountTracker(u32 newFrame);
 	// Sets frameAdvancing variable to false
-	// Used to restrict a frameAdvaceTravker value from transferring between recordings
+	// Used to restrict a frameAdvaceTracker value from transferring between recordings
```
sonicfind, line 55: Thank you for fixing that before I did. lol

InfamousKnight:
Is this  #3735 related to this?

pcsx2/Recording/InputRecordingControls.h
RedDevilus: Typo: frameAdvanceTracker , maybe capitalise on first letter too?

xTVaser:
No I don't believe the issues are related.

pcsx2/Recording/InputRecordingControls.h
xTVaser: The capitalization matches with the variable so no changes to make there, but yeah I didnt notice to other missing letter, I'll fix that lol

PR #3762, opened 2020-09-27
GameDB: Fix trees in Legend of Kay

Mrlinkwii:
update Legend of Kay in gameDB


bin/GameIndex.dbf
```diff
@@ -41038,7 +41039,7 @@ Serial = SLUS-21248
 Name   = Legend of Kay
 Region = NTSC-U
 Compat = 5
-EETimingHack = 1 // For freezes in some scene transitions.
```
lightningterror, line 41041: Is the EETimingHack no longer needed?

PR #3763, opened 2020-09-28
Recording: Upgrade to Input Recording file V2 (Multitap support+) and other Input Recording Improvements

sonicfind:
With the integration of multitap in input recordings, a few additional accompanying updates had to be made:
* The number of Virtual Pads has also increased to 8 and their menu options have been split into two submenus of 4 by port.
Multitap 1 active:
image
Multitap 2 inactive:
image

* Automatic enabling and disabling of multitaps and of the fast boot configuration option now occurs when opening a recording according to necessity.
* With the increased possible pad count, a recording file size optimization was needed. Now, every frame in *.pirec files (previously *.p2m2) only hold the input data for the number of pads used - in comparison to version 1's consistent two-pad style. Singleplayer recordings will now be about *HALF* the size.
* **Note:** *Steps were taken to ensure backwards compatibility with version 1 recording files. However, there will be no converter from V1 to V2, at least not right now.*

Additionally, all instances of UndoCount have been changed to RedoCount to more accurately describe that behavior.

The actual use and construction of multitap-based recordings are only *fully* compatible with lilypad. Once onepad fully supports multitap, that will change.

### DIFF
--- a/common/include/PS2Edefs.h
+++ b/common/include/PS2Edefs.h
@@ -236,6 +236,8 @@ void CALLBACK PADupdate(int pad);
 // before the pad saw them. So the gui needs to send them back to the pad.
 void CALLBACK PADWriteEvent(keyEvent &evt);
 
+void CALLBACK PADSetupInputRecording(bool start, char padsRequested);
+
 // extended funcs
 
 void CALLBACK PADgsDriverInfo(GSdriverInfo *info);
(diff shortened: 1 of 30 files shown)


sonicfind:
Also, once the prerequisite PR #3356 is merged, I can rebase to remove some the extra weirdness in terms of re-added master commits.

sonicfind:
Splitting this into two smaller but easier to manage PRs (although the second will be made after the first one is good for merge).

Issue #3758, opened 2020-09-26
CD/DVD wont load when swaped  in BIOS 


Mrlinkwii:
***PCSX2 version:***

v1.7.0-dev-340-g5de82289d

***PCSX2 options:***
default - 2 (safe) 

***Plugins used:***
all defaults 

***Description of the issue:***
ISO wont read when swaping disc  in BIOS 

***How to reproduce the issue:***

set to no Disc and boot BIOS and when in the BIOS  swap to an ISO , the ISO wont load and get the error      

```
(CdvdSource) HotSwapping CDVD source types from NoDisc to ISO.
Patches: No CRC found, using 00000000 instead.
(IsoFS) Invalid partition descriptor encountered at block 0x10: ''
Patches: No CRC found, using 00000000 instead.

```

***PC specifications:***
i5-3210m
intel hd 4000 
ubuntu 20.04

GovanifY:
commit https://github.com/PCSX2/pcsx2/commit/551847411f99a98c94777176472c61b99b76549c caused this issue, but cannot be reverted for obvious reasons, so we'll have to untangle that mess

PR #3770, opened 2020-09-30
CDVD: fix for loading games in BIOS

GovanifY:
Probably fixes https://github.com/PCSX2/pcsx2/issues/3758


GovanifY:
Trsting confirms it to work

lightningterror (approved):
Tested and it does fix the issue.

PR #3778, opened 2020-10-02
IOP DMA: Always fire interrupts for SIF0/SIF1

refractionpcsx2:
Also cleaned up a bit of white space

For details please see this post: https://www.patreon.com/posts/sins-of-ps2-2001-42262496

Fixes Knockout Kings 2001 hang on boot

Based on information provided by Dobiestation author PSI.


Felipefpl:
Good to see old issues such as this one fixed. I wonder if there are others in the same situation.

PR #3780, opened 2020-10-02
Recording: Fix VS filters on new image files and safeguards around save-state code

xTVaser:
The SaveState handler in `InputRecording` is always accessible because it needs to be, even if input-recording tools are disabled.  However, the extra code within it, should not be executed if the recording tools are disabled, so I fixed that.

Also at the same time I noticed that since I didn't specify a filter for the new VirtualPad image files, they get shoved into the top level project folder so fixed that up at the same time.

### DIFF
--- a/pcsx2/Recording/InputRecording.cpp
+++ b/pcsx2/Recording/InputRecording.cpp
@@ -32,23 +32,28 @@
 
 void SaveStateBase::InputRecordingFreeze()
 {
+	// NOTE - BE CAREFUL
+	// CHANGING THIS WILL BREAK BACKWARDS COMPATIBILITY ON SAVESTATES
 	FreezeTag("InputRecording");
 	Freeze(g_FrameCount);
 
 #ifndef DISABLE_RECORDING
-	// Loading a save-state is an asynchronous task. If we are playing a recording
-	// that starts from a savestate (not power-on) and the starting (pcsx2 internal) frame
-	// marker has not been set (which comes from the save-state), we initialize it.
-	if (g_InputRecording.IsInitialLoad())
-		g_InputRecording.SetStartingFrame(g_FrameCount);
-	else if (g_InputRecording.IsActive())
+	if (g_Conf->EmuOptions.EnableRecordingTools)
 	{
-		// Explicitly set the frame change tracking variable as to not
-		// detect saving or loading a savestate as a frame being drawn
-		g_InputRecordingControls.SetFrameCountTracker(g_FrameCount);
+		// Loading a save-state is an asynchronous task. If we are playing a recording
+		// that starts from a savestate (not power-on) and the starting (pcsx2 internal) frame
+		// marker has not been set (which comes from the save-state), we initialize it.
+		if (g_InputRecording.IsInitialLoad())
+			g_InputRecording.SetStartingFrame(g_FrameCount);
+		else if (g_InputRecording.IsActive())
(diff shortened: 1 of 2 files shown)


pcsx2/Recording/InputRecording.cpp
```diff
@@ -32,23 +32,28 @@
...
+		{
+			// Explicitly set the frame change tracking variable as to not
+			// detect saving or loading a savestate as a frame being drawn
+			g_InputRecordingControls.SetFrameCountTracker(g_FrameCount);
 
-		if (IsLoading())
-			g_InputRecording.SetFrameCounter(g_FrameCount);
+			if (IsLoading())
+				g_InputRecording.SetFrameCounter(g_FrameCount);
+		}
```
sonicfind, line 56: There is actually very meticulous reason for how this was originally set up. Thanks to the the sanity checks and updates added in the last few months, both of these conditions _should_ only ever be true in very specific circumstances that already *require* `g_Conf->EmuOptions.EnableRecordingTools` to be active. Therefore, adding an additional check for that alone would be a little redundant (even if the redundancy is literally just one line of code executed very rarely).
Unless of course you found edge cases where that isn't true, then good call.

xTVaser: Yeah I imagine the current code is designed in such a way that it is resilient to whether or not recording is enabled or not.  However my concern is about easily weeding out "is this problem because of recording" -- someone can just flip the flag and try loading/saving the save-state and then we are in the clear. A lot of the current entrypoints into recording code is gated behind the global option, this is one of the exceptions to that. 

This came up the other day when it was discovered there was a breaking change to the save-states in an open PR.

sonicfind: I guess that's fair.
Funnily enough, just a bit ago, I did discover an edge case that we glossed over. We never had a sanity check for when the base savestate of a recording is incompatible with that version of PCSX2. It would continue to setup the recording as if the savestate successfully load *but* it wouldn't ever reach `InputRecordingFreeze`; thus, it would never call `SetStartingFrame` or unset `initialLoad`.

sonicfind: Would be good to add this underneath the `StateCopy_LoadFromFile` call in Play();
```
if (initialLoad) // If the value is still true, the savestate failed to load for some reason
{
	recordingConLog(wxString::Format("[REC]: %s_SaveState.p2s is not compatible with this version of PCSX2.\n", inputRecordingData.GetFilename()));
	recordingConLog(wxString::Format("[REC]: Original PCSX2 version used: %s\n", inputRecordingData.GetHeader().emu));
	inputRecordingData.Close();
	initialLoad = false;
	return false;
}
```

xTVaser: Suggestion added

xTVaser: Tested the suggestion and had to remove, it causes a regression because the call to `StateCopy_LoadFromFile` is not a blocking call, it is async.  I'll track this issue and we can follow up in a separate PR.  Likely need to handle it within the freeze handler, although long-term, I'd like a more explicit hook around loading save-states, adding side-effects onto the freeze feels like a temporary hack.

sonicfind: What exactly was the regression in this case?

PR #3781, opened 2020-10-03
Recording: Add on-screen-display logging around appropriate Input Recording actions

xTVaser:
This adds OSD logs around input recording actions, specifically:
- Pausing (note that since its pausing emulation, the log won't show up until the next frame)
- Resuming
- Recording Mode Toggling
- Starting a new Recording
- Playing an existing Recording
- Stopping a Recording

I did not include actions for frame-advance, because it would just fill the screen with logs and as mentioned before, the log will only show up the next frame.

I also added a small logging utility header which simplifies the logging story.

### Examples

image
image
image

### DIFF
--- a/pcsx2/CMakeLists.txt
+++ b/pcsx2/CMakeLists.txt
@@ -555,6 +555,7 @@ set(pcsx2RecordingHeaders
 	${rec_src}/InputRecordingFile.h
 	${rec_src}/NewRecordingFrame.h
 	${rec_src}/PadData.h
+    ${rec_src}/Utilities/InputRecordingLogger.h
 	${rec_vp_src}/VirtualPad.h
 	${rec_vp_src}/VirtualPadData.h
 	${rec_vp_src}/VirtualPadResources.h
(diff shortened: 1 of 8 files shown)


RedDevilus:
Couldn't you implement a ms delay on pause so it will show Pausing just before it goes into paused state? But i see why you did it that way, it's pretty intrusive for a minor nitpick.

xTVaser:
> Couldn't you implement a ms delay on pause so it will show Pausing just before it goes into paused state? But i see why you did it that way, it's pretty intrusive for a minor nitpick.
Unfortunately no I don't think so.  Pausing is directly tied to the frame count.  (You pause on a specific frame, not a specific timestamp). And since GS draws once per frame...I don't think there's a way.

I thought about just not doing the `Paused Emulation` log at all, but I figured if someone accidentally paused and then resumed/frame-advanced, maybe it would help clue them in on what they did 🤷

sonicfind:
Just a natural consequence on how OSD logs are rendered. Nothing we can do about that. Better to have it just in case.

Issue #3720, opened 2020-09-20
uLaunchELF / wLaunchELF won't boot directly using Run ELF


voltagex:
***PCSX2 version:***
``(For example, stable build v1.6.0 (or) development builds v1.7.0-dev-xxx)``
Stable v1.6.0

***Plugins used:***
``(Relevant/Non-default settings of GS/PAD/SPU2/CDVD/USB plugins)``
```
	Bound   GS: GSdx32-AVX2.dll [GSdx 20200506140834 (MSVC 19.25 AVX2/AVX2) 1.2.0]
	Bound  PAD: LilyPad.dll [LilyPad (20200506140834) 0.12.1]
	Bound SPU2: Spu2-X.dll [SPU2-X 20200506140834 2.0.0]
	Bound CDVD: CDVDnull.dll [CDVDnull Driver 20151231175709 0.6.0]
	Bound  USB: USBnull.dll [USBnull Driver 20200506140834 0.7.0]
	Bound   FW: FWnull.dll [FWnull Driver 20200506140834 0.7.0]
	Bound DEV9: CLR_DEV9.dll [CLR DEV9 0.8.9]

```
- **Note**: In case the issue is a graphical problem with the GSdx plugin, please provide basic information like ``Renderer/Internal resolution/CRC Hack level/Other relevant options`` which were used to reproduce the issue. Additionally, providing a [GS dump](https://forums.pcsx2.net/Thread-How-to-create-a-proper-GS-dump) of the issue would also be helpful.

***Description of the issue:***
``(What's the problem? - Screenshots showing the issue if applicable)``

![image](https://user-images.githubusercontent.com/159567/93694147-c1288a80-fb4b-11ea-968e-15787a5511c3.png)
There's `CDVDiso Error: No filename specified.` in the log, even though I'm not trying to load an ISO, even with CDVDnull loaded.

If I "insert" an ISO, Launch ELF will 'work', but black screen with the following log:

```
(IsoFS) Block 0x10: Primary partition info.
(IsoFS) Filesystem is ISO9660
(SYSTEM.CNF) Detected PS2 Disc = cdrom0:\ULE_4_38.ELF;1
(SYSTEM.CNF) Software version = 1.00
(SYSTEM.CNF) Disc region type = NTSC


IOP Realtime Kernel Ver.0.9.1

    Copyright 1999 (C) Sony Computer Entertainment Inc. 

Protected page @ 0x00084
[...]

PR #3784, opened 2020-10-04
CDVD: Allow executing ELFs with "ISO" selected 

AmFobes:
This fix allows us to tell CDVD to check if we are loading an ELF or not. This is required due to the fact that CDVD loads whatever it can on plugin load.

As shown in issue #3720, whenever there is no ISO selected in the ISO selector and CDVD is set to "ISO" mode, it'll prompt you with an error message.

Also: I attempted to at least allow CDVD to try and load when executing an ELF, this is for those who use homebrew but still need to emulate a DVD drive.

### DIFF
--- a/pcsx2/CDVD/CDVDaccess.cpp
+++ b/pcsx2/CDVD/CDVDaccess.cpp
@@ -368,12 +368,20 @@ bool DoCDVDopen()
 
 	auto CurrentSourceType = enum_cast(m_CurrentSourceType);
 	int ret = CDVD->open(!m_SourceFilename[CurrentSourceType].IsEmpty() ?
-							 static_cast<const char*>(m_SourceFilename[CurrentSourceType].ToUTF8()) :
-							 (char*)NULL);
+							static_cast<const char*>(m_SourceFilename[CurrentSourceType].ToUTF8()) :
+							(char*)NULL,
+							(CurrentSourceType == CDVD_TYPE_NODISC));
 
 	if (ret == -1)
-		return false; // error! (handled by caller)
-	//if( ret == 1 )	throw Exception::CancelEvent(L"User canceled the CDVD plugin's open dialog."); <--- TODO_CDVD is this still needed?
+	{
+		// CDVD was unable to open, error handled by caller
+		return false;
+	}
+	if (ret == -2)
+	{
+		// CDVD was unable to open but it doesn't matter because we are launching an ELF
+		return true;
+	}
 
 	int cdtype = DoCDVDdetectDiskType();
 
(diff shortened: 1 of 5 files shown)


pcsx2/CDVD/CDVDisoReader.cpp
```diff
@@ -40,12 +40,16 @@ void CALLBACK ISOclose()
...
 }
 
-s32 CALLBACK ISOopen(const char* pTitle)
+s32 CALLBACK ISOopen(const char* pTitle, bool launchELF)
 {
 	ISOclose(); // just in case
 
 	if ((pTitle == NULL) || (pTitle[0] == 0))
 	{
+		if (launchELF) {
```
lightningterror, line 49: Formatting.

pcsx2/CDVD/CDVDaccess.cpp
```diff
@@ -369,11 +369,16 @@ bool DoCDVDopen()
...
+							 (char*)NULL,(CurrentSourceType == CDVD_TYPE_NODISC));
 
-	if (ret == -1)
-		return false; // error! (handled by caller)
-	//if( ret == 1 )	throw Exception::CancelEvent(L"User canceled the CDVD plugin's open dialog."); <--- TODO_CDVD is this still needed?
+	if (ret == -1) {
+		// CDVD was unable to open, error handled by caller
+		return false; 
+	}
+	if (ret == -2) {
```
lightningterror, line 378: These too.

lightningterror:
Maybe just run clang on the files in a separate commit?

AmFobes:
Something is different with my local machine. It just doesn't match.

AmFobes:
oki doki ran clang format

pcsx2/CDVD/CDVDaccess.cpp
```diff
@@ -368,12 +368,20 @@ bool DoCDVDopen()
 
 	auto CurrentSourceType = enum_cast(m_CurrentSourceType);
 	int ret = CDVD->open(!m_SourceFilename[CurrentSourceType].IsEmpty() ?
-							 static_cast<const char*>(m_SourceFilename[CurrentSourceType].ToUTF8()) :
-							 (char*)NULL);
+                             static_cast<const char*>(m_SourceFilename[CurrentSourceType].ToUTF8()) :
+                             (char*)NULL,
```
lightningterror, line 372: Tabs instead of spaces, above too.

AmFobes: Apparently, clang-format prefers spaces. fixed...

lightningterror: Need to adjust that clang setting then.

lightningterror:
I can no longer reproduce the issue on latest master, can someone else verify so we can decide what to do with the pr?

AmFobes:
Yeah I can't reproduce the error now 
I can't reproduce it even on a [30 day old commit](https://github.com/PCSX2/pcsx2/commit/0d1a5831449c4f31993ca4838dd1c39707e96cce)
You can close this if you want...

PR #3785, opened 2020-10-04
Counters: Tighten tolerances on framelimiter

RedPanda4552:
Smooths out frame pace and keeps average framerate closer to target.

Comparing PCSX2's old framelimiter with RTSS showed there was room for improvement. Frametimes regularly would jump up and down in various situations. This PR smooths out frame pace substantially (easily visible in RTSS) and, drum roll please, may appease the input lag folks for a time. We'll find out.

Extensive user tests required; some vsync functions were shuffled around, and this was tested on a high performance system. We should test low end systems, and a wide variety of games to make sure there are no outlier games that needed the old behavior for some reason.

Shoutout to Refraction for greasing the wheels and helping with the various vsync aspects.

Before (above) and after (below) of Xenosaga Ep. 1 save point inside the Elsa:
image

### DIFF
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -452,6 +452,9 @@ static __fi void frameLimit()
 
 static __fi void VSyncStart(u32 sCycle)
 {
+	frameLimit(); // limit FPS
+	gsPostVsyncStart(); // MUST be after framelimit; doing so before causes funk with frame times!
+
 	if(EmuConfig.Trace.Enabled && EmuConfig.Trace.EE.m_EnableAll)
 		SysTrace.EE.Counters.Write( "    ================  EE COUNTER VSYNC START (frame: %d)  ================", g_FrameCount );
 
@@ -518,9 +521,6 @@ static __fi void VSyncEnd(u32 sCycle)
 	// Call it every 60 frames
 	if (!(g_FrameCount % 60))
 		sioNextFrame();
-
-	frameLimit(); // limit FPS
-	gsPostVsyncStart(); // MUST be after framelimit; doing so before causes funk with frame times!
 
 	// This doesn't seem to be needed here.  Games only seem to break with regard to the
 	// vsyncstart irq.
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -433,9 +433,9 @@ static __fi void frameLimit()
 	// If any integer value of milliseconds exists, sleep it off.
 	// Prior comments suggested that 1-2 ms sleeps were inaccurate on some OSes;
 	// further testing suggests instead that this was utter bullshit. 
-	if (msec >= 1)
+	if (msec > 1)
 	{
-		Threading::Sleep(msec);
+		Threading::Sleep(msec - 1);
 	}
 	
 	// Conversion to milliseconds loses some precision; after sleeping off whole milliseconds,
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -103,7 +103,7 @@ static __fi void _rcntSet( int cntidx )
 	if (c < nextCounter)
 	{
 		nextCounter = c;
-		cpuSetNextEvent( nextsCounter, nextCounter );	//Need to update on counter resets/target changes
+		cpuSetNextEvent( nextsCounter, nextCounter ); // Need to update on counter resets/target changes
 	}
 
 	// Ignore target diff if target is currently disabled.
@@ -121,7 +121,7 @@ static __fi void _rcntSet( int cntidx )
 		if (c < nextCounter)
 		{
 			nextCounter = c;
-			cpuSetNextEvent( nextsCounter, nextCounter );	//Need to update on counter resets/target changes
+			cpuSetNextEvent(nextsCounter, nextCounter); // Need to update on counter resets/target changes
 		}
 	}
 }
@@ -139,6 +139,7 @@ static __fi void cpuRcntSet()
 
 	// sanity check!
 	if( nextCounter < 0 ) nextCounter = 0;
+	cpuSetNextEvent(nextsCounter, nextCounter); // Need to update on counter resets/target changes
 }
 
 void rcntInit()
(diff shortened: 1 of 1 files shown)


pcsx2/Counters.cpp
```diff
@@ -515,6 +520,7 @@ static __fi void VSyncEnd(u32 sCycle)
 		sioNextFrame();
 
 	frameLimit(); // limit FPS
+	gsPostVsyncStart(); // MUST be after framelimit; doing so before causes funk with frame times!
```
refractionpcsx2, line 523: Move these 2 lines to the beginning of static __fi void VSyncStart(u32 sCycle)
Needs to be on the start function to fix some graphical errors

refractionpcsx2 (changes requested):

MrCK1:
Example comparison of master vs PR - Midway Arcade Treasures title screen
image

refractionpcsx2 (approved):

MrCK1:
*Bonus*
R&C3 (Obiani Draco)

PR:
image

Master:
image

MrCK1 (approved):

pcsx2/Counters.cpp
```diff
@@ -394,58 +395,65 @@ void frameLimitReset()
...
-	// use the expected frame completion time as our starting point.
-	// improves smoothness by making the framelimiter more adaptive to the
-	// imperfect TIMESLICE() wait, and allows it to speed up a wee bit after
-	// slow frames to "catch up."
+	// Conversion of delta from CPU ticks (microseconds) to milliseconds
+	s32 msec = (int) ((sDeltaTime * -1000) / (s64) GetTickFrequency());
+	
+	// If any integer value of milliseconds exists, sleep it off.
+	// Prior comments suggested that 1-2 ms sleeps were inaccurate on some OSes;
+	// further testing suggests instead that this was utter bullshit.
```
CookiePLMonster, line 435: This is true and might compromise accuracy a bit - it's worth running a test without sleep at all.

CookiePLMonster:
NASCAR Heat 2002 had good frametimes during the race, but entering a pit lane resulted in frametimes being all over the place:
image

With this PR, it doesn't seem to occur anymore:
image

pcsx2/Counters.cpp
refractionpcsx2: I just tested the theory, it does indeed make it smoother, how noticeable it'll be though is any bodies guess

sleepmsecs

refractionpcsx2: Other games such as SotC don't seem to care too much
sleepmsecs2

Dreadmoth:
A _different_ test (Drakan: The Ancients' Gates with FrameratePAL=119.98)

Master:
Master120
PR:
PR120

👍

pcsx2/Counters.cpp
CookiePLMonster: Might be that those games don't hit the condition for a sleep?

refractionpcsx2: I run at well above the frame cap in those games. I get about 70fps in SotC, so there should be about 3.334ms my CPU is waiting every frame.

CookiePLMonster: Interesting, maybe then you are hitting some non-contractual behaviour of `Sleep` where it special cases smaller sleeps? Even if this was the case I would not rely on it.

tadanokojin (approved):

pcsx2/Counters.cpp
refractionpcsx2: Well from my experience (and what I've read around the web) the inaccuracy of sleep is < 1ms, I would guess I'm seeing maybe 0.3ms inaccuracy which I don't think anybody is going to honestly notice, usually for stutters you need to see a good couple of ms to notice it.  I think pandubz is going to do this change anyway (maybe some others, dunno) so that will reduce it, I just don't know how much that'll affect people on the edge of the frame cap, spinlocking is great but it uses CPU power to do it.

CookiePLMonster: It might depend on the power plan as my experience with sleeping inside a frame limiter was drastically different - to the extent where e.g. a loop intended to lock to 120 FPS ended up at 112 FPS.

refractionpcsx2: yeah I mean that's 0.59ms which works out as 2fps either way, which isn't brilliant I'd admit, but as experienced, I changed it to do if(msec > 1) sleep(msec-1) so there was always at least 1ms we had to spinlock on, that got me the above results, which didn't improve things drastically.

You're not going to run PS2 games at 120fps that's for sure :P

RedDevilus:
All tested on default settings (commit-93823a3)
Game 1 Baroque;

Master:
image

PR:
image

Game 2 True Crime NY (Game needs a patch/fix);

Master:
image


PR:
image

Game 3 Prince Caspian Chapter 2;

Master:
image

PR:
image

Game 4 Rogue Galaxy;

Master:
image

PR:
image

pcsx2/Counters.cpp
RedPanda4552: Another look at the Xenosaga Ep. 1 Elsa save point. Master (top), `msec >=1 then sleep(msec)` (middle), `msec > 1 then sleep(msec - 1)` (bottom). I think we have our winner.
image

CookiePLMonster: That's a good approach, but I would go even further and make it something like `msec - 3` or even 5. The threshold has to be pretty big or else you'll always risk people losing time on those e.g. due to running on a laptop battery.

RedPanda4552:
Another change has just been pushed that should (sometimes) tighten this up even further.

pcsx2/Counters.cpp
refractionpcsx2: Well I don't recommend people run the emulator on a laptop and if they do they should be using high performance mode and not running off battery (that's a really dumb idea :P) I'm honestly not concerned about people crazy enough to do that, it will still be head over shoulders above master (which was 1ms buffer)

CookiePLMonster: Sure, that should be good enough as a starting point and could be revisited in the future if it's ever needed.

RedDevilus: I'm using a gaming laptop plugged in with high performance mode and it does improve significantly on specific games, u could stare blindly at squeezing an extra 5%-10%.

MrCK1:
Some more testing with new changes (I'll edit if I find anything substantial)

CoD Finest Hour (PR on top, master on bottom)
image

RedDevilus:
Latest PR (8aaae36)
Prince caspian Chapter 2
Seems an improvement did happen.

image

Dreadmoth:
Ratchet & Clank 2
RC2frametime

Champions - Return to Arms
CRTAframetime

Metal Arms - Glitch in the System 💯 
MAframetime

pcsx2/Counters.cpp
```diff
@@ -394,58 +395,65 @@ void frameLimitReset()
...
+		frameLimitUpdateCore();
 		return;
 	}
 
-	// use the expected frame completion time as our starting point.
-	// improves smoothness by making the framelimiter more adaptive to the
-	// imperfect TIMESLICE() wait, and allows it to speed up a wee bit after
-	// slow frames to "catch up."
+	// Conversion of delta from CPU ticks (microseconds) to milliseconds
+	s32 msec = (int) ((sDeltaTime * -1000) / (s64) GetTickFrequency());
```
lightningterror, line 431: Mark it as a const?

RedDevilus:
Added more examples along with my system info

Laracroft Legend:
LaraCroftLegend-3758

Alter Echo:
AlterEcho-3758

FinalFantasyX:
FinalFantasyX-3758

FinalFantasyX-2:
FinalFantasyX-2-3758

RedDevilus (approved):

RedPanda4552:
Has anyone checked Linux? RTSS is not available but at least to make sure stability is still good?

RedDevilus:
Examples when i change the msec

if (msec > 1):  (original PR)
image
if (msec > 2):
image
if (msec > 3):
image


@RedPanda4552 
It reduces spikes a bit more, but i'll let you decide what you want to do with it.

auscompgeek:
Just tested on Linux x86 with EGL and GTK2, with a few Kingdom Hearts games (KH1, KH2, Re:CoM). Seems stable.

RedPanda4552:
Discovered this article thanks to a forum user: https://randomascii.wordpress.com/2020/10/04/windows-timer-resolution-the-great-rule-change/

We may need to play with Win10 2004 release. Sounds like they are changing how thread sleeping will behave and sleep(1) is going to be a LOT more latent than ever before, if this is right.

pcsx2/Counters.cpp
CookiePLMonster: Good timing, as Bruce Dawson just published a somewhat-related post:
https://randomascii.wordpress.com/2020/10/04/windows-timer-resolution-the-great-rule-change/

RedPanda4552:
Update on the above: multiple testers have disclosed they did their tests on 2004 and had positive results. I am comfortable with where this is but more testing is always welcome. Up to everyone else.

OCRBonk:
Frame pacing did seem quite a bit better in a few games I tried, although Xenosaga is one that still kind of gives weird readings.
As far as input lag goes, I tested DMC which is already a bit of a slightly laggy game on real hardware. I couldn't feel the difference between builds when I tested them back and forth. Mind you this was on Windows 7. I'll have to try again on my Win 10 drive

PR #3787, opened 2020-10-04
SPU2: Change default volume from 100% to 50%

RedDevilus:
This will make PCSX2 have half volume instead of loud sounds if your system is also on 100%. Anti-earbleeds.

### DIFF
--- a/pcsx2/SPU2/Linux/Config.cpp
+++ b/pcsx2/SPU2/Linux/Config.cpp
@@ -88,7 +88,7 @@ void ReadSettings()
 	Interpolation = CfgReadInt(L"MIXING", L"Interpolation", 4);
 	EffectsDisabled = CfgReadBool(L"MIXING", L"Disable_Effects", false);
 	postprocess_filter_dealias = CfgReadBool(L"MIXING", L"DealiasFilter", false);
-	FinalVolume = ((float)CfgReadInt(L"MIXING", L"FinalVolume", 100)) / 100;
+	FinalVolume = ((float)CfgReadInt(L"MIXING", L"FinalVolume", 100)) / 200;
 	if (FinalVolume > 1.0f)
 		FinalVolume = 1.0f;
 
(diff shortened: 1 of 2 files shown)


seta-san:
i've actually wondered why the emulator has a volume control to begin with. can't you just adjust your computer's main volume? or adjust pcsx2's through the OS's per program adjustments?

RedDevilus:
You can do that but i like it separated. Then you have more granular controls and more ways for users to adjust.

kattjevfel:
Sounds like user error if they ram the volume to the max without expecting loud noises.

sotpapathe:
I don't really see the point in this. If I set the system volume to 100% I expect to hear loud sounds. Every other program I've used sets its internal volume to 100% by default, so setting it to 50% on PCSX2 I think would go against what most users expect.

tadanokojin:
> Sounds like user error if they ram the volume to the max without expecting loud noises.

They don't "ram the volume". It's the default.

kattjevfel:
> > Sounds like user error if they ram the volume to the max without expecting loud noises.
> 
> They don't "ram the volume". It's the default.

I am referring to inside their OS, where it is *not* the default. Either way I really don't see why you want to go against the norm here of setting full volume. Since this is *your* personal preference perhaps it should stay on *your* system?

tadanokojin:
The default is louder than any other app on my system including discord (with output volume at 200%) and my web browser by like a factor of 4 and I shouldn't have to change per-application volume or pcsx2 settings to prevent getting my ears murdered every time I grab a new build for the same reason I shouldn't be served a 150C cup of coffee and told "just blow on".

But hey, we can both be happy here. We can make the whole scale half as loud. That way you can keep your arbitrary 100% value ;)

image

seta-san:
For the record. I'm against the emulator having it's 
own volume control. I don't particularly care what it's highest volume is. I just think that you can adjust your speaker or OS volume control

RedDevilus:
Result:
image

sotpapathe:
I think we're missing the point here. There seem to be 2 reports (@RedDevilus and @tadanokojin) that PCSX2 is significantly louder than other applications on their systems. This must be a bug in PCSX2 not handling some particular system configurations correctly. Lowering the default volume for everyone does not fix the real issue. Opening a bug report with the sound plugins you use and system info is the way to go I think.

Changing the volume slider so that 100% is now half as loud as before does nothing more than confuse users which suddenly receive much lower volume output from PCSX2. You're essentially forcing your volume preferences an all other users. And it seems the majority of users have to issue with the default volume being what it currently is.

This seems like a bug and should be handled as such.

RedDevilus:
It's not a bug, it has always been louder than any application. And you can always change it to 200% and i changed it for 50% to become the new default. It's still pretty loud to me but i didn't want users to say its too silent now. Did you compare the difference from 100% and half the volume?

RedDevilus:
Besides it's the exact same thing that discord and other applications does:

image

To me it's more semantics and i always need to lower system volume and it just proves how powerful PCSX2 is, a bit too powerful in the sound department. I don't see anyone complaining that Discord or other applications do it like this.

RedDevilus:
#3962 has fixed this issue, i will however bring the improvements to slider in another PR (0-100% with 100% still as default)

PR #3789, opened 2020-10-04
recording: Resolve VirtualPads not drawing for some games

sonicfind:
The call for the virtualPads to redraw their visual assets originally required for the sio to provide a `bufCount` of **20** to `controllerInterrupt` - the last value index used for polling controller data (reinterpreted as **17**). However, for some of the games that do not utilize button pressure, that value will only ever reach a max of  **8** (Ex: Sonic Mega Collection Plus); consequently, the virtual pad will never be drawn.

By unhooking the padData logging and virtualPad redrawing from this conditional and instead calling it from a static location, both will now occur regardless of max `bufCount` value. `ControllerInterrupt` was also rearranged to allow virtualPads to be utilized outside of the limited frame window of an input recording (with limitations in record mode).

### DIFF
--- a/pcsx2/Recording/InputRecording.cpp
+++ b/pcsx2/Recording/InputRecording.cpp
@@ -111,7 +111,7 @@ void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 b
 			if (frameCounter >= 0 && frameCounter < INT_MAX)
 			{
 				if (!inputRecordingData.ReadKeyBuffer(bufVal, frameCounter, port, bufIndex))
-					inputRec::consoleLog(fmt::format("[REC]: Failed to read input data at frame {}", frameCounter));
+					inputRec::consoleLog(fmt::format("Failed to read input data at frame {}", frameCounter));
 
 				// Update controller data state for future VirtualPad / logging usage.
 				padData[port]->UpdateControllerData(bufIndex, bufVal);
@@ -142,7 +142,7 @@ void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 b
 					}
 
 					if (frameCounter < INT_MAX && !inputRecordingData.WriteKeyBuffer(frameCounter, port, bufIndex, bufVal))
-						inputRec::consoleLog(fmt::format("[REC]: Failed to write input data at frame {}", frameCounter));
+						inputRec::consoleLog(fmt::format("Failed to write input data at frame {}", frameCounter));
 				}
 			}
 			// If the VirtualPad updated the PadData, we have to update the buffer
--- a/pcsx2/Recording/InputRecording.cpp
+++ b/pcsx2/Recording/InputRecording.cpp
@@ -97,58 +97,58 @@ void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 b
 
 	if (bufCount == 1)
 		fInterruptFrame = data == READ_DATA_AND_VIBRATE_FIRST_BYTE;
-	else if (bufCount == 2 && buf[bufCount] != READ_DATA_AND_VIBRATE_SECOND_BYTE)
-		fInterruptFrame = false;
-
-	// We do not want to record or save the first two bytes in the data returned from the PAD plugin
-	else if (fInterruptFrame && bufCount >= 3 && frameCounter >= 0 && frameCounter < INT_MAX)
+	else if (bufCount == 2)
+	{
+		if (buf[bufCount] != READ_DATA_AND_VIBRATE_SECOND_BYTE)
+			fInterruptFrame = false;
+	}
+	else if (fInterruptFrame)
 	{
 		u8& bufVal = buf[bufCount];
 		const u16 bufIndex = bufCount - 3;
-
 		if (state == InputRecordingMode::Replaying)
 		{
-			u8 tmp = 0;
-			if (inputRecordingData.ReadKeyBuffer(tmp, frameCounter, port, bufIndex))
+			if (frameCounter >= 0 && frameCounter < INT_MAX)
 			{
-				// Overwrite value originally provided by the PAD plugin
-				bufVal = tmp;
+				if (!inputRecordingData.ReadKeyBuffer(bufVal, frameCounter, port, bufIndex))
+					inputRec::consoleLog(fmt::format("[REC]: Failed to read input data at frame {}", frameCounter));
+
+				// Update controller data state for future VirtualPad / logging usage.
+				padData[port]->UpdateControllerData(bufIndex, bufVal);
+				
+				if (virtualPads[port]->IsShown())
+					virtualPads[port]->UpdateControllerData(bufIndex, padData[port]);
 			}
 		}
-
-		// Update controller data state for future VirtualPad / logging usage.
-		padData[port]->UpdateControllerData(bufIndex, bufVal);
-
-		if (virtualPads[port] &&
-			virtualPads[port]->IsShown() &&
-			virtualPads[port]->UpdateControllerData(bufIndex, padData[port]) &&
-			state != InputRecordingMode::Replaying)
+		else
 		{
+			// Update controller data state for future VirtualPad / logging usage.
+			padData[port]->UpdateControllerData(bufIndex, bufVal);
+
+			// Commit the byte to the movie file if we are recording
+			if (state == InputRecordingMode::Recording)
+			{
+				if (frameCounter >= 0)
+				{
+					// If the VirtualPad updated the PadData, we have to update the buffer
+					// before committing it to the recording / sending it to the game
+					if (virtualPads[port]->IsShown() && virtualPads[port]->UpdateControllerData(bufIndex, padData[port]))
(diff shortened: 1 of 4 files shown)


pcsx2/Recording/InputRecording.cpp
```diff
@@ -91,58 +91,57 @@ void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 b
...
 		fInterruptFrame = data == READ_DATA_AND_VIBRATE_FIRST_BYTE;
-	else if (bufCount == 2 && buf[bufCount] != READ_DATA_AND_VIBRATE_SECOND_BYTE)
-		fInterruptFrame = false;
-
-	// We do not want to record or save the first two bytes in the data returned from the PAD plugin
-	else if (fInterruptFrame && bufCount >= 3 && frameCounter >= 0 && frameCounter < INT_MAX)
+	else if (bufCount == 2)
+	{
+		if (buf[bufCount] != READ_DATA_AND_VIBRATE_SECOND_BYTE)
+			fInterruptFrame = false;
```
xTVaser, line 97: I don't see the need for a nested if, can just be an AND like before

pcsx2/Recording/InputRecording.cpp
```diff
@@ -91,58 +91,57 @@ void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 b
...
+			if (state == InputRecordingMode::Recording)
 			{
-				inputRecordingData.IncrementUndoCount();
-				incrementUndo = false;
+				if (frameCounter >= 0)
+				{
+					u8& bufVal = buf[bufCount];
+					// If the VirtualPad updated the PadData, we have to update the buffer
+					// before committing it to the recording / sending it to the game
+					if (virtualPads[port]->IsShown() && virtualPads[port]->UpdateControllerData(bufIndex, padData[port]))
```
xTVaser, line 128: probably want to add the `virtualPads[port]` to this list of checks incase the VP isn't initialized

pcsx2/Recording/InputRecording.cpp
```diff
@@ -91,58 +91,57 @@ void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 b
...
-			if (virtualPads[port] && virtualPads[port]->IsShown() && !virtualPads[port]->IsIconized())
-				virtualPads[port]->Redraw();
-		}
-
-		// Finally, commit the byte to the movie file if we are recording
-		if (state == InputRecordingMode::Recording)
+		else
 		{
-			if (incrementUndo)
+			const u16 bufIndex = bufCount - 3;
```
xTVaser, line 119: this value is calculated twice identically, can probably be calculated earlier, once

pcsx2/Recording/InputRecording.cpp
```diff
@@ -173,6 +172,18 @@ void InputRecording::IncrementFrameCounter()
...
+{
+	for (u8 port = 0; port < 2; port++)
+	{
+		padData[port]->LogPadData(port);
+		// As well as re-render the virtual pad UI, if applicable
+		// - Don't render if it's minimized
+		if (virtualPads[port]->IsShown() && !virtualPads[port]->IsIconized())
+			virtualPads[port]->Redraw();
+	}
+}
```
xTVaser, line 185: 👍

pcsx2/Recording/InputRecordingControls.cpp
```diff
@@ -60,6 +60,7 @@ void InputRecordingControls::HandleFrameAdvanceAndPausing()
 	else if (frameCountTracker != g_FrameCount)
 	{
 		frameCountTracker = g_FrameCount;
+		g_InputRecording.LogAndRedraw();
```
xTVaser, line 63: As we add more and more to this class, its becoming less about just being an interface for the key-bindings and more about handling frame-to-frame interactions.  In the future, we may want to split that up / pull it into `InputRecording.cpp`.  But this isn't something for this PR by any means.

pcsx2/Recording/InputRecordingFile.cpp
```diff
@@ -158,29 +158,19 @@ bool InputRecordingFile::OpenExisting(const wxString path)
...
-	}
-
 	long seek = getRecordingBlockSeekPoint(frame) + controllerInputBytes * port + bufIndex;
-	if (fseek(recordingFile, seek, SEEK_SET) != 0 || fread(&result, 1, 1, recordingFile) != 1)
-	{
-		return false;
-	}
-
-	return true;
+	return fseek(recordingFile, seek, SEEK_SET) == 0 && fread(&result, 1, 1, recordingFile) == 1;
```
xTVaser, line 162: very nice refactor.  However, we are dropping the `nullptr` check, perhaps that should remain?

pcsx2/Recording/InputRecordingFile.cpp
```diff
@@ -158,29 +158,19 @@ bool InputRecordingFile::OpenExisting(const wxString path)
...
 	{
-		return;
+		totalFrames = frame;
+		fseek(recordingFile, seekpointTotalFrames, SEEK_SET);
+		fwrite(&totalFrames, 4, 1, recordingFile);
 	}
-	totalFrames = frame;
-	fseek(recordingFile, seekpointTotalFrames, SEEK_SET);
-	fwrite(&totalFrames, 4, 1, recordingFile);
+	fflush(recordingFile);
```
xTVaser, line 173: I'm not personally very familiar with C file management.  Why do we need the `fflush` here, and removed it from `WriteKeyBuffer`?  Otherwise this looks great as well, but same question around removing the `nullptr` check

pcsx2/Recording/InputRecordingFile.cpp
```diff
@@ -202,20 +192,8 @@ bool InputRecordingFile::WriteHeader()
...
 	long seek = getRecordingBlockSeekPoint(frame) + 18 * port + bufIndex;
-
-	if (fseek(recordingFile, seek, SEEK_SET) != 0 || fwrite(&buf, 1, 1, recordingFile) != 1)
-	{
-		return false;
-	}
-
-	fflush(recordingFile);
-	return true;
+	return fseek(recordingFile, seek, SEEK_SET) == 0 && fwrite(&buf, 1, 1, recordingFile) == 1;
```
xTVaser, line 196: 👍 but same question around `nullptr`

pcsx2/Recording/InputRecording.cpp
sonicfind: Not necessarily. If, for example, `state == InputRecordingMode::Replaying` *but* you loaded a savestate that is set before the first frame of the recording, bufIndex would not get used until you reach frame 0. Having it calculated beforehand would be pointless.

pcsx2/Recording/InputRecording.cpp
sonicfind: Because regardless of if `buf[bufCount] is equal to READ_DATA_AND_VIBRATE_SECOND_BYTE`, bufCount is *still* 2. Therefore, the very next else if (originally) would already be false before checking its conditions - `fInterruptFrame && bufCount >= 3 && frameCounter >= 0 && frameCounter < INT_MAX`.
By nesting the `buf[bufCount] != READ_DATA_AND_VIBRATE_SECOND_BYTE`, we negate the need for `bufCount >= 3`.

pcsx2/Recording/InputRecordingFile.cpp
sonicfind: ReadKeyBuffer is only called when a recording is in Replay mode. With the changes brought on by play-fixes and such, Replay mode should only possible with a file loaded. Therefore, a nullptr check wouldn't be necessary.

pcsx2/Recording/InputRecording.cpp
sonicfind: What kind of situation would cause it to be uninitialized? I always figured, with how it gets set up, that it would initialized no matter what.
.... is it a case of wx being dumb? I swear...

xTVaser: It's probably a very unlikely rare event that could happen, it shouldn't happen normally but technically the entrypoint for this function is completely independent from the GUI initialization so it could happen.

pcsx2/Recording/InputRecordingFile.cpp
xTVaser: Fair enough, it just feels a bit fragile depending on those external guarantees.  If those were to ever change, the file handling code can no longer stand on it's own.

xTVaser (changes requested):
This seems to have broken the virtualpad functionality.  It seems to only display updates when a recording is being replayed back.  I can't see any updates when:
- outside a recording
- recording mode in an active input recording

sonicfind:
oops, I immediately see why. My bad

xTVaser (approved):
Pulled down and tested on a game that has the conventional 18bytes per frame:
- VirtualPad worked inside and outside of a recording as I'd expect
- As did the controller log

I assume you've tested it on the games of the type you mentioned in your description as that's how you found the issue to begin with, so looks good from my perspective.

PR #3790, opened 2020-10-05
Add SDL2 into 3rdparty, and enable it as a module on Windows.

arcum42:
It still puzzles me that we only put SDL2 in as a module for Linux when it's completely crossplatform. So I thought I'd see how much was needed to add it to windows, and, well, really not terribly much, apparently.

As onepad requires SDL 2, having it in would make it a lot easier to port onepad's controller code to windows as well.

Bear in mind if any changes to the project in Windows seem odd, I'm a Linux person, and really have no idea what I'm doing in Visual Studio...


arcum42:
Changes I made to SDL 2:
I changed it to generate a lib instead of a dll.
I added #define HAVE_LIBC 1 to SDL_config.h.

I also added imm32.lib;version.lib;SDL2.lib; to the libraries pcsx2 pulls in.

lightningterror:
I'm not sure if it's worth the trouble if we aren't merging onepad, need to decide what we are going to do with pad for core merge.

arcum42:
Believe me, the trouble in this case was just getting sdl2 to compile and link. The code for sdl in spu2 was already totally crossplatform, it just wasn't added to the module list or compiled in Windows. I always intended eventually to get onepad working on Windows through sdl. The main blockers were having to write a new windows project for it, link in SDL, and totally rewrite the keyboard code on the Windows side. Writing, say, a cubeb module, would have been way harder.

As far as the pad merger goes: 
If we only merge in only lilypad, well, there's some preliminary Linux code in there, but I have no idea if it works. It'd probably be coding the Linux side largely from scratch, and given that that's generally code no one likes working with...

If we merge onepad in, at least on Linux, it'll be using SDL unless someone rewrites all the gamepad code. I probably could get something onepad based working in pcsx2 for both platforms, honestly, but if I did, I'd be using SDL on both sides...

And if we code a new replacement, either someone will be totally rewriting things from scratch, or we'll be borrowing from existing plugin code. Which still probably means borrowing gamepad code from one of the onepad's for Linux, which is in SDL. Also keep in mind that both the gamepad API's I'm familiar with are SDL. Any help I can give on recoding the pad plugin into pcsx2 becomes more limited if we aren't using it.

I do realise SDL has to be compiled, and that adds time, but it's not like you usually clean and rebuild the whole project from scratch every time. Technically speaking, we could actually avoid that if we had it link to a dll instead, and didn't actually build SDL, too, but that didn't really seem to be how we were handing any other 3rd party dependencies (though that's usually how it's included).

(And while, yes, we could do SDL everything and put GSdx in an SDL window, unless we go with imgui for all the settings, that's not really a route I'd want to go. Technically, you can get a Vulkan or OpenGL or DirectX window out of it, but you also can from wxwidgets or QT...)

carmiker:
It is rarely a good idea to import massive, full-featured, and widely available libraries into a codebase. Maintenance burden should be handled by the OS distributor, and in the case of Windows builds, distributing binary releases with a .dll is not difficult.

1. Keeping in sync with upstream releases becomes a chore. This becomes an even bigger problem when people don't know if they're building against the included library or the system-wide library.
2. There will inevitably be a temptation to internally modify the dependency. Then, keeping in sync is even more difficult, and despite any illusions any of us may have about ourselves and our programming abilities, it is unlikely that changes to a library like SDL will not have unintended consequences.
3. Increased build complexity for components that should be handled elsewhere.

This is not a personal attack, just a collection of observations based on many years of working with open source software. I have seen this scenario play out countless times, at the project level and Linux distribution level. Don't do it!

lightningterror:
Pandubz is working on a pad system built in to core https://github.com/RedPanda4552/pcsx2/tree/pad-core

Anyway I'm not in favor of merging this. Both Onepad and Lilypad suck and lack features that the other has.

PR #3791, opened 2020-10-05
Correct PGIF Register bit masks

Nobbs66:
The PGIF registers are 20 bits, but the bitmasks allowed 21 bits to be written.

### DIFF
--- a/pcsx2/ps2/pgif.cpp
+++ b/pcsx2/ps2/pgif.cpp
@@ -392,13 +392,13 @@ u32 immRespHndl(u32 cmd, u32 data)
         case 0x01:
             break;  //Returns Nothing (old value in GPUREAD remains unchanged)
         case 0x02:
-            data = pgif1reg & 0x001FFFFF;
+            data = pgif1reg & 0x000FFFFF;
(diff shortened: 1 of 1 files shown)


pcsx2/ps2/pgif.cpp
```diff
@@ -392,13 +392,13 @@ u32 immRespHndl(u32 cmd, u32 data)
         case 0x01:
             break;  //Returns Nothing (old value in GPUREAD remains unchanged)
         case 0x02:
-            data = pgif1reg & 0x001FFFFF;
+            data = pgif1reg & 0x000FFFFF;
             break;  //Read Texture Window setting  ;GP0(E2h) ;20bit/MSBs=Nothing
         case 0x03:
-            data = pgif2reg & 0x001FFFFF;
+            data = pgif2reg & 0x00FFFFF;
```
refractionpcsx2, line 398: Just an annoyance, but if you're gonna pad it with zero's can you make sure it's of even length (just needs one more zero)

refractionpcsx2 (changes requested):

PR #3793, opened 2020-10-05
Generate a Position Independant Executable

GovanifY:
PIE/PIC is pretty much mandatory when dealing with libraries in modern OSes. Our plugins are already PIC(64 bit only currently but moot point since I'm merging plugins) but not the main executable. As the reasons why PIC was disabled on 32bit are pretty much moot now I enabled PIE globally.

~~This currently only implements it in CMake, I'll check msbuild later today.~~ msbuild seems to enable it by default.

This probably doesn't require much testing, if it boots and the jit doesn't crash we're good basically

### DIFF
--- a/cmake/BuildParameters.cmake
+++ b/cmake/BuildParameters.cmake
@@ -185,7 +185,7 @@ if(${PCSX2_TARGET_ARCHITECTURES} MATCHES "i386")
     #     - code uses the ebx register so it's not compliant with PIC.
     #     - Impacts the performance too much.
     #     - Only plugins. No package will link to them.
-    set(CMAKE_POSITION_INDEPENDENT_CODE OFF)
+    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
     if(NOT DEFINED ARCH_FLAG)
         if (DISABLE_ADVANCE_SIMD)
--- a/common/include/Utilities/StringHelpers.h
+++ b/common/include/Utilities/StringHelpers.h
@@ -226,7 +226,7 @@ extern bool pxParseAssignmentString(const wxString &src, wxString &ldest, wxStri
 #define pxsFmt FastFormatUnicode().Write
 #define pxsFmtV FastFormatUnicode().WriteV
 
-#define pxsPtr(ptr) pxsFmt("0x%08X", (ptr)).c_str()
+#define pxsPtr(ptr) pxsFmt("0x%08llX", (u64)(uptr)(ptr)).c_str()
 
 extern wxString &operator+=(wxString &str1, const FastFormatUnicode &str2);
 extern wxString operator+(const wxString &str1, const FastFormatUnicode &str2);
--- a/common/src/Utilities/Linux/LnxHostSys.cpp
+++ b/common/src/Utilities/Linux/LnxHostSys.cpp
@@ -23,6 +23,8 @@
 #include <signal.h>
 #include <errno.h>
 #include <unistd.h>
+#include <sys/stat.h>
+#include <fcntl.h>
 
 // Apple uses the MAP_ANON define instead of MAP_ANONYMOUS, but they mean
 // the same thing.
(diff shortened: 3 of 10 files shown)


GovanifY:
Actually according to `readpe`, unless I'm mistaken, pcsx2 seems to already be compiled with DEP out of the box(windows acronym for PIE)
```
    DLL characteristics names                                                                                                           
                                         IMAGE_DLLCHARACTERISTICS_DYNAMIC_BASE                                                          
                                         IMAGE_DLLCHARACTERISTICS_NX_COMPAT                                                                                                                                                                                                      
                                         IMAGE_DLLCHARACTERISTICS_TERMINAL_SERVER_AWARE                                                                                                                                                                                          
```

GovanifY:
I removed DevCon logs that print the mapped addresses on release to avoid obvious leaks and made all mmap work on system-provided addresses unless the `JIT_DEBUG` flag is set, keeping our static address behaviour when debugging is necessary and otherwise using sane defaults.

@tellowkrinkle is that fine with you?

GovanifY:
I broke asserts at some point when making this PR, I'll fix that when I have the time

GovanifY:
@tellowkrinkle whenever you review this the base address generation part I introduced in this PR is incorrect, do you have any ideas on how to do it cleanly/what are the requirements to make the JIT work?

tellowkrinkle:
Yeah it's not going to work on 64-bit, we'd need to either rewrite the JIT to produce code that follows the large code model or stop touching C++ global variables from JITed code if we want to allocate anywhere

If you page-align `codeBase` instead of aligning it to `1<<28`, it should follow the ASLR of the main executable, is that fine?

Also, on the disabling of address printing, if the program is able to read the console (which is outside of the VM), wouldn't that mean they've already defeated ASLR?

GovanifY:
> If you page-align `codeBase` instead of aligning it to `1<<28`, it should follow the ASLR of the main executable, is that fine?

It should probably be fine, not the cleanest solution but until the JIT has a major rewrite it seems like the best solution.

> Also, on the disabling of address printing, if the program is able to read the console (which is outside of the VM), wouldn't that mean they've already defeated ASLR?

Well logs gets written to a file, and hostfs can  read the filesystem :)

GovanifY:
@tellowkrinkle does this seem good enough for you on 64bit?

tellowkrinkle:
`MmapReservePtr` maps straight to mmap, it will either allocate at the requested address or wherever it feels like, which is why we try multiple addresses in `makeMainMemoryManager` in the first place, so giving it a the same address every time isn't going to work

GovanifY:
Well we don't give it the same address each time in this case, we use the function pointer base which is different each time since the executable is a PIE. mmap(NULL) won't work either as it wouldn't necessarily fit the +- 2GB region in the 64bit mapping

tellowkrinkle:
[If another mapping already exists there, the kernel picks a new address that **may or may not depend on the hint.**](https://www.man7.org/linux/man-pages/man2/mmap.2.html)

We can test whether it does or doesn't depend on the hint in various OSes:
Run this program:
```c
#include <sys/mman.h>
#include <stdio.h>
#include <stdint.h>

void test(intptr_t addr) {
	void* buf = mmap((void*)addr, 4096, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
	printf("Mmap with hint %p was mapped at %p\n", (void*)addr, buf);
	munmap(buf, 4096);
}

	void* extra = (void*)0x400000000;
	printf("Main is at %p\n", main);
	printf("Stack is at %p\n", &extra);
	printf("Extra mapping at %p\n", extra);
	mmap(extra, 4096, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE, -1, 0);
	test(0);
	test((intptr_t)main & ~4095ull);
	test((intptr_t)main);
	test((intptr_t)&extra);
	test((intptr_t)extra);
	test((intptr_t)extra + 4096);
	return 0;
}
```

On macOS, it prints this, which would indicate that hinting a taken address will work fine:
```
Main is at 0x100003e30
Stack is at 0x7ffeefbff478
Extra mapping at 0x400000000
Mmap with hint 0x0 was mapped at 0x1003d9000
Mmap with hint 0x100003000 was mapped at 0x1003d9000
Mmap with hint 0x100003e30 was mapped at 0x1003d9000
Mmap with hint 0x7ffeefbff478 was mapped at 0x7ffeefc00000
Mmap with hint 0x400000000 was mapped at 0x400001000
Mmap with hint 0x400001000 was mapped at 0x400001000
```
But on my Linux computer (Ubuntu 18.04), I get this:
```
Main is at 0x4005d0
Stack is at 0x7ffd32c47a08
Extra mapping at 0x400000000
Mmap with hint (nil) was mapped at 0x7fcd35c5d000
Mmap with hint 0x400000 was mapped at 0x7fcd35c5d000
Mmap with hint 0x4005d0 was mapped at 0x7fcd35c5d000
Mmap with hint 0x7ffd32c47a08 was mapped at 0x7fcd35c5d000
Mmap with hint 0x400000000 was mapped at 0x7fcd35c5d000
Mmap with hint 0x400001000 was mapped at 0x400001000
```
If we don't probe around, we will not get a mapping anywhere near where we want

GovanifY:
Thank you for correcting my naive thinking that kernels would share the same behaviour. I'll implement something else for 64 bit when I have time.
All of this is probably a temporary hack until I decide myself to work  on the JIT though, as I'm planning to work on it. It'll definitely come after 64 bit, plugin merging and the new GUI are in though, so that'll take a while.

GovanifY:
I began working on rewriting the VTLB. For now I was able to only remove the sign hack from the physical TLB as the virtual one is casted by the JIT, and simply editing a field of the structure will cause the JIT to bork. I can fully get the sign hack out of virtual on interpreter mode currently.

I'll see what I can do for virtual now.

GovanifY:
Hey @tellowkrinkle , I've been hitting a brick wall when it comes to fixing the VTLB recompiler. It uses the struct as a raw uptr pointer, which isn't true when you modify the struct, thus even adding an element breaks everything. I've calculated the offset between the raw pointer and the uptr but even then I cannot get the recompiler to work, the Interpreter works just fine. Any clue on how I could go around to fix this?

tellowkrinkle:
You need to update for size of the struct as well, I probably should have used `sizeof(VTLBVirtual)` instead of `wordsize` here
You'll probably have to switch it with a separate SHL though, the maximum multiplication you can do in an x86 instruction is 8 bytes, and the new VTLBVirtual is 16 bytes on x86-64

GovanifY:
I'm closing this for now as this will require a bit more work on my side to get it up to par but I still intend to work on that later, as having a non relocatable JIT is pretty effin terrible imho.

I'll open a new PR when I have time to make something that actually works

PR #3801, opened 2020-10-06
SPU2: more missing PCH

GovanifY:
This should fix travis for good this time


arcum42:
It's also missing in SndOut_SDL.cpp...

GovanifY:
@arcum42 thanks for noticing! Forgot SDL was linux only for a bit, just woke up :/

arcum42:
np. When I decided to try to see if I could get the SDL SPU2 module running in Windows , that was one of the two or so lines in pcsx2 I had to add. It's really already written as cross platform, just no one added it on the Windows side.

PR #3809, opened 2020-10-08
Locales: Updating Portuguese Brazilian translation.

altiereslima:


lightningterror:
Suggest postponing this until after dev9 is merged in to core.

PR #3816, opened 2020-10-11
FPU: Removed forced rounding on DIV/SQRT/RSQRT (Do Not merge!)

refractionpcsx2:
Added gamefix for Mortal Kombat Shaolin Monks

Description:

For a very long time PCSX2 has had the rounding mode of DIV, SQRT and RSQRT on the FPU forced to "Nearest" instead of Chop/Zero (unless you use the Neg DIV hack), presumably because it fixed some game and seemed like a good idea.  This PR aims to remove it, but it needs to be tested thoroughly before we even think about merging it.  I suspect a few games will require a gamefix (like Shaolin Monks did) but many other games might see a positive benefit from doing this.

Please reply below if any game needs any other rounding setting and if it has any adverse effects on the game.

Improvements:
Stuntman - AI is improved (but still not brilliant)
Freakstyle - Less issue with invisible walls
ONI - Interactable doors now fixed

Regressions:
Mortal Kombat Shaolin Monks - Texture sizes, fixed by setting rounding to positive

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -14612,10 +14612,12 @@ vuRoundMode = 0 // Fixes textures and crashes.
 Serial = SLES-53524
 Name   = Mortal Kombat - Shaolin Monks
 Region = PAL-M5
+eeRoundMode = 2 // Fixes texture sizes
 ---------------------------------------------
 Serial = SLES-53525
 Name   = Mortal Kombat - Shaolin Monks
 Region = PAL-G
+eeRoundMode = 2 // Fixes texture sizes
 ---------------------------------------------
 Serial = SLES-53526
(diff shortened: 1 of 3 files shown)


boringhexi:
May we also list games that seem unaffected? This could save other testers some time.

Gitaroo Man: Did not notice any new graphical bugs, nor any improvement of the existing graphical bug (i.e. Stage 4 flashing). (Tested 30-60 seconds of each stage including the multiplayer one.)

RedDevilus:
Better to test several games and mention where you see improvements or more importantly problems like regressions. Unless you want to mention several of the 4K distinct games (or 10K+ entries of the GameDB). Unaffected games don't really lower or speed-up the merging process in this manner.

PR #3821, opened 2020-10-12
gsdx-d3d: Disable fbmask on Destroy All Humans on basic blending.

lightningterror:
gsdx-d3d11: Disable fbmask on Destroy All Humans on basic blending.
Hopefully not many games hit the mask value, otherwise blending will
need to be turned up to High.


BreakPoints:
Would this also apply to Destroy All Humans 2?

PR #3822, opened 2020-10-13
gsdx-hw: Invert Automatic mipmapping code.

lightningterror:
Always enable Basic mipmapping except for the following games that have
issues with it.

This way we rely on less crc and more games get to use it which need it
with adding less crc.

Following games need to be tested:
The Polar Express,
Nicktoons: Battle for Volcano Island,
Nicktoons: Attack of the Toybots,
El Tigre: The Adventures of Manny Rivera

And others possible too.


gregory38:
Honestly it is a bad idea.

Mipmap basic
1/ can screwed up cache badly
2/ reduce texture quality (which is the goal of mipmap in the first place). It can help for shimmering and aniso. But I doubt it will play well with upscaling.

refractionpcsx2:
I have to agree.  For a lot of games mipmapping will stay on the maximum texture quality and with upscaling that works great and people appreciate that due to the higher quality of the games.

On the flip side, this removes a bunch of hacks so....

Positives said, if people don't like it, they can disable mipmapping for the game  they want the mipmapping not to happen on.

RedDevilus:
I have tested this with 16x aniso and native + upscaling 6x to 8x. On some games it has a positive effect to others in my collection having no obvious changes. (Like tomb raider legend likes it on basic for example the vegetation model problems).

It doesn't seem to be performance taxing like blending is on higher levels of accuracy.

refractionpcsx2:
The biggest issue here is that Mipmapping on HW is a hack, so we don't know *when* it will break something.

gregory38:
This pr doesn't remove/add hack. It is about auto configuration of mipmap.
Unfortunately, there is no sane default. The real issue isn't limited to mipmap. And I'm afraid, you will never get a single configuration best for all games.
What we need is a database that give default/quality/speed/best/whatever preset setting. At least with plugin merge it could be shared with core

lightningterror:
So shall I close this then? Guess I/we'll need to add more games to crc list.

PR #3823, opened 2020-10-14
GuiConsole: Fix autodock and set it to a small gap

RedDevilus:


lightningterror:
Was it tested on Windows 8.1, MAC, Linux?

RedDevilus:
Linux: Problem with compiling (not tested)
W8.1: Seems to not move with the main window and sit in a specific spot)
W10: Works perfectly

RedDevilus:
Hotfix results;
Linux: Still needs testing
W8.1: Attaches to main window with no gap
image
W10: Attaches to main window with gap
image

lightningterror:
I'd just leave it as it is honestly.

RedDevilus:
Master autodock for linux (shadows get on top of the console) :
image


PR for linux:
image

RedDevilus:
Reason for this PR is that WX is calculating the topright wrong which gives it an extra 16px gap for no reason in windows but it doesn't do this in Linux. Whereas Linux does have shadows that interfere with the look of the whole program.

lightningterror:
I prefer with gap because of the shadow.

pcsx2/gui/MainFrame.cpp
```diff
@@ -192,7 +192,11 @@ void MainEmuFrame::OnMoveAround(wxMoveEvent& evt)
 		{
 			if (!proglog->IsMaximized())
 			{
-				g_Conf->ProgLogBox.DisplayPosition = GetRect().GetTopRight();
+				// GetSize() is wrong on windows.
+				// GetTopRight() is also wrong on windows since it probably uses GetSize() while it gives the wrong values for autodock.
+				// ConsoleLogger.cpp L713: if( wxWindow* main = GetParent() ) ;always triggers on false and it's corresponding lines never get executed.
+				int xCoord = GetRect().GetTopLeft().x + GetClientSize().GetWidth() + 10; // Adds a 10px gap between the main window and console.
```
lightningterror, line 198: You can use const here.

lightningterror:
Can someone post a ss how it looks on windows 8.1?

RedDevilus:
8.1 Windows: 
image

lightningterror:
Doesn't look good, main window overlaps with the console.

RedDevilus:
I could change 10 to 15

RedDevilus:
8.1:
image

10:
image

Changed back to original value of 20 but eliminates some issues people could have from wrong calculation for screen position.

lightningterror:
There's still some gap.

RedDevilus:
Making it a bigger gap will make it uglier for w10 just appealing to how w8.1 works.

lightningterror:
I think we should leave this out as it feels more like a hackfix.

RedDevilus:
I could look into make it seperate for 8.1 because of how the gaps work differently. Current code already is a hack on it's own.

tellowkrinkle:
macOS looks like Linux, though it uses less aggressive shadows
Screenshot of PCSX2 with no gap between windows
Screenshot of PCSX2 with gap between windows

GovanifY:
Looks like we like the gap after all

PR #3825, opened 2020-10-14
SPU2: Change default of latency slider

RedDevilus:
Linux: 300 to 100 (which should be fine for everyone), windows was already 100ms.
Windows: Change the steps to start from 3 and increment on higher numbers instead of 0.5 steps which is just stupid to be that granular.


lightningterror:
Is it a good idea? It was 150 in the past, maybe ask @ramapcsx2 for more details why it was changed.
https://github.com/PCSX2/pcsx2/commit/8320fa704a8a9a54e81cdade188680db22faf7a2

ramapcsx2 (approved):
As discussed in Discord: 100ms should be fine on both platforms these days.
If not, there are some serious problems anyway!

PR #3756, opened 2020-09-25
Merging SPU2-X into core

GovanifY:
Another day, another plugin to merge into the core.

This time around the PR is mostly done, ~~but still have some critical dumb things to fix:~~ and is ready for review!

~~* Windows support. I could not figure for the life of me how to fix includes hell from MSVC and I'll probably need a fresh brain to solve this~~
~~* Interrupt/DMA handlers might not be configured correctly, it currently freezes games at boot which is a little bit annoying.~~

The save state format had a small redesign to phase out from the Plugin system and seems to work well. The configuration GUI has been merged into the main Config button and now uses SPU2.ini instead of SPU2-X.ini

Probably forgot plenty of things I've done as this is a hefty PR but you'll probably see them during review anyways.

Issue #1775, opened 2017-01-16
No sound with portaudio after upgrading to alsa-lib-1.1.3


orbea:
***PCSX2 version:***
`alsa-lib-compat32-1.1.3-x86_64-1compat32`
`portaudio-compat32-v190600_20161030-x86_64-1_SBocompat32`
`pcsx2-2017.01.15_f9c202514_master-x86_64-1_git`

***Plugins used:***
`SPU2-X 2.0.0 [libspu2x-2.0.0]`

***Description of the issue:***
After Slackware current upgraded to `alsa-lib-1.1.3` from `alsa-lib-1.1.2` and when using the portaudio module PCSX2 will do one of two things.
1. Games will have no audio.
2. Games will start with audio, but after a minute at most audio will suddenly die.

PCSX2 will now hang for a little while when trying to shutdown games.

The SDL2 audio module has no issues and downgrading to `alsa-lib-1.1.2` will allow audio to work again.
I also do not have pulseaudio installed so this happens with just alsa.

I am not sure if this is a pcsx2, portaudio or alsa issue, but hopefully starting here will at least help narrow it down. Is there anything else that can be useful to debug this?

***How to reproduce the issue:***
1. Install `alsa-lib-1.1.3`.
2. Start a game with the portaudio module.

***PC specifications:***
OS: `Slackware64-current`

orbea:
As a sanity check I compiled alsa-lib-1.1.3 from source instead of using the slackware packages and was able to reproduce the issue. I do not have any other programs which I can really test portaudio with though.

For reference here is the alsa-lib v1.1.2 to v1.1.3 changelog.
http://www.alsa-project.org/main/index.php/Changes_v1.1.2_v1.1.3

gregory38:
Well the issue is likely alsa or portaudio. Need to search on Google if projects got done issues. (Unfortunately Alsa 1.1.3 is quite young)

gregory38:
https://github.com/JvanKatwijk/sdr-j-dab/issues/9

Could you check the env variable of the thread ?

np511:
On Monday, January 16, 2017 11:59:59 AM EST Gregory Hainaut wrote:

> https://github.com/JvanKatwijk/sdr-j-dab/issues/9

>

> Could you check the env variable of the thread ?

If it is alsa or portaudio, why not try with PortAudio using a different backend?

np511:
[...]

PR #3827, opened 2020-10-15
cmake: Fix the build without portaudio (Again).

orbea:
I think this was broken after PR https://github.com/PCSX2/pcsx2/pull/3756.

### DIFF
--- a/pcsx2/CMakeLists.txt
+++ b/pcsx2/CMakeLists.txt
@@ -39,7 +39,11 @@ else()
     set(pcsx2FinalFlags "")
 endif()
 
-set(pcsx2FinalFlags ${pcsx2FinalFlags} "-DSPU2X_PORTAUDIO" ${CommonFlags})
+set(pcsx2FinalFlags ${pcsx2FinalFlags} ${CommonFlags})
+
+if(PORTAUDIO_FOUND)
+    set(pcsx2FinalFlags ${pcsx2FinalFlags} -DSPU2X_PORTAUDIO)
+endif()
 
 if(XDG_STD)
     set(pcsx2FinalFlags ${pcsx2FinalFlags} -DXDG_STD)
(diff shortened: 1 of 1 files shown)


GovanifY:
PortAudio support is still hardcoded in for windows. If we want to reintroduce portaudio-less compilation we might want to add this in.
Is there an actual reason to compile without PortAudio though? It is currently, by far, our best audio output latency wise.

orbea:
I am not sure if windows was ever a consideration, regardless its not something I can test. This is mostly important for *unix systems that may not always have portaudio, but do have SDL2.

GovanifY:
LGTM then, but I would _highly_ recommend distros that do not ship PulseAudio to do so. Slackware thought process is "it adds complexity" but they fail to register that SDL adds even more in this corner case, and ends up recommending SDL over PA as a sound driver, which is pretty insane to me ngl.

GovanifY:
LGTM then but aren't most modern distros supporting portaudio? As far as I know you are a manager of Slackware, so how is https://slackbuilds.org/repository/14.2/development/portaudio/ working out with this?
I'm fine with merging this either way, it's just curiosity at this point.

orbea:
In my personal case pcsx2 is the only program which can use portaudio while many programs want SDL2, I don't think this uncommon. Also portaudio only builds with `-j1` which is annoying.

GovanifY (approved):

gregory38:
Extra note: Portaudio got issue with pulseaudio too. It was the reason of sdl2 module. I don't know if it is fixed nowadays (unlikely)

orbea:
Portaudio was also broken with alsa for a time, but that is now fixed. https://github.com/PCSX2/pcsx2/issues/1775

I think in general a lot more people are testing SDL2.

PR #3834, opened 2020-10-16
Actions: Add Clang Linux Build to Actions

xTVaser:
In the effort to not continue to find out about clang failures only in master, I've added a clang build.  I've also cleaned up the linux build a bit.

I've _also_ removed the cron-job artifact cleanup, as github has recently added first-class support for it and the cron-job was throwing API rate-limits (though this bug appears to be fixed around the same time) https://github.blog/changelog/2020-10-08-github-actions-ability-to-change-retention-days-for-artifacts-and-logs/


xTVaser:
ccache seems to be fine, not seeing the same problem i used to, repeated builds pass 
image

PR #3838, opened 2020-10-17
Support build with system fmt

nE0sIghT:
```
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld.bfd: pcsx2/CMakeFiles/PCSX2.dir/gui/FrameForGS.cpp.o: in function `GSPanel::InitRecordingAccelerators()':
FrameForGS.cpp:(.text+0x4ffd): undefined reference to `fmt::v7::detail::vformat[abi:cxx11](fmt::v7::basic_string_view<char>, fmt::v7::format_args)'
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld.bfd: pcsx2/CMakeFiles/PCSX2.dir/Recording/InputRecording.cpp.o: in function `inputRec::consoleLog(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)':
InputRecording.cpp:(.text+0x1685): undefined reference to `fmt::v7::detail::vformat[abi:cxx11](fmt::v7::basic_string_view<char>, fmt::v7::format_args)'
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld.bfd: pcsx2/CMakeFiles/PCSX2.dir/Recording/InputRecording.cpp.o: in function `InputRecording::SetStartingFrame(unsigned int)':
InputRecording.cpp:(.text+0x187d): undefined reference to `fmt::v7::detail::vformat[abi:cxx11](fmt::v7::basic_string_view<char>, fmt::v7::format_args)'
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld.bfd: pcsx2/CMakeFiles/PCSX2.dir/Recording/InputRecording.cpp.o: in function `inputRec::log(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)':
InputRecording.cpp:(.text+0x198d): undefined reference to `fmt::v7::detail::vformat[abi:cxx11](fmt::v7::basic_string_view<char>, fmt::v7::format_args)'
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld.bfd: pcsx2/CMakeFiles/PCSX2.dir/Recording/InputRecording.cpp.o: in function `InputRecording::Create(wxString, bool, wxString)':
InputRecording.cpp:(.text+0x253b): undefined reference to `fmt::v7::detail::vformat[abi:cxx11](fmt::v7::basic_string_view<char>, fmt::v7::format_args)'
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld.bfd: pcsx2/CMakeFiles/PCSX2.dir/Recording/InputRecording.cpp.o:InputRecording.cpp:(.text+0x2b7c): more undefined references to `fmt::v7::detail::vformat[abi:cxx11](fmt::v7::basic_string_view<char>, fmt::v7::format_args)' follow
collect2: error: ld returned 1 exit status
```

### DIFF
--- a/cmake/SearchForStuff.cmake
+++ b/cmake/SearchForStuff.cmake
@@ -242,3 +242,15 @@ if((GCC_VERSION VERSION_EQUAL "9.0" OR GCC_VERSION VERSION_GREATER "9.0") AND GC
     Even with that patch, compiling with LTO may still segfault. Use at your own risk!
     This text being in a compile log in an open issue may cause it to be closed.")
 endif()
+
+find_package(fmt "7.0.3" QUIET)
+if(NOT fmt_FOUND)
+    if(EXISTS "${CMAKE_SOURCE_DIR}/3rdparty/fmt/fmt/CMakeLists.txt")
+        message(STATUS "No system fmt was found. Using bundled")
+        add_subdirectory(3rdparty/fmt/fmt)
+    else()
+        message(FATAL_ERROR "No system or bundled fmt was found")
+    endif()
+else()
+    message(STATUS "Found fmt: ${fmt_VERSION}")
+endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,11 +41,6 @@ if(NOT NO_TRANSLATION)
     endif()
 endif()
 
-# 3rdpary Libraries
-if(EXISTS "${CMAKE_SOURCE_DIR}/3rdparty/fmt/fmt")
-    add_subdirectory(3rdparty/fmt/fmt)
-endif()
-
 # make common
 if(common_libs)
     add_subdirectory(common/src/Utilities)


matoro:
This should require at least v7.0.3 of libfmt I believe:  https://github.com/fmtlib/fmt/issues/795

refractionpcsx2:
tagging @xTVaser

samuelgrf (changes requested):
This seems like a nicer approach, because all code related to searching packages would be inside `SearchForStuff.cmake` instead of being in two different files.

Note: this is from https://github.com/archlinux/svntogit-community/blob/41d7c3e291905e096e3f541cdbcf06faee4ea32e/trunk/pcsx2-system-fmt.patch

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 1bc43a8e3..a8dd6e2ea 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,11 +41,6 @@ if(NOT NO_TRANSLATION)
     endif()
 endif()

-# 3rdpary Libraries
-if(EXISTS "${CMAKE_SOURCE_DIR}/3rdparty/fmt/fmt")
-    add_subdirectory(3rdparty/fmt/fmt)
-endif()
-
 # make common
 if(common_libs)
     add_subdirectory(common/src/Utilities)
diff --git a/cmake/SearchForStuff.cmake b/cmake/SearchForStuff.cmake
index 7a3871535..236682e6a 100644
--- a/cmake/SearchForStuff.cmake
+++ b/cmake/SearchForStuff.cmake
@@ -74,6 +74,11 @@ endif()
 find_package(wxWidgets COMPONENTS base core adv)
 find_package(ZLIB)

+find_package(fmt)
+if(NOT fmt_FOUND)
+    add_subdirectory(${CMAKE_SOURCE_DIR}/3rdparty/fmt/fmt)
+endif()
+
 ## Use pcsx2 package to find module
 include(FindLibc)
```

xTVaser (changes requested):
Seems fine to me for people who prefer to use their system's libraries.  However:
- Should probably add a check to ensure that atleast `v7` will be installed on the system? (assuming such a thing is possible)
- This suggestion does look better https://github.com/PCSX2/pcsx2/pull/3838#pullrequestreview-513135597 but it omits the directory check, which from looking at the cmake documentation https://cmake.org/cmake/help/v3.0/command/add_subdirectory.html does not have any implicit check for the directory existing, so that should probably be kept.

Also, before this is merged, it will need to be ran through the Travis-CI system on a fork, as that was causing issues before.

nE0sIghT:
I moved add_subdirectory call to SearchStuff, added target fmt version and added status message with fmt version found.

cmake/SearchForStuff.cmake
```diff
@@ -233,3 +233,12 @@ if((GCC_VERSION VERSION_EQUAL "9.0" OR GCC_VERSION VERSION_GREATER "9.0") AND GC
     Even with that patch, compiling with LTO may still segfault. Use at your own risk!
     This text being in a compile log in an open issue may cause it to be closed.")
 endif()
+
+find_package(fmt "7.0.3")
```
samuelgrf, line 237:
```suggestion
find_package(fmt "6.2.1")
```
I am running this on NixOS and that is the fmt version it comes with, I successfully built pcsx2 and booted into the bios.
I don't know much about the CMake syntax, so I don't know if this also allows using higher versions.

nE0sIghT: >  don't know much about the CMake syntax, so I don't know if this also allows using higher versions.

It allows higher versions so your suggestion is right. However @xTVaser requested "atleast `v7`" earlier.

samuelgrf: I guess it would be best to just stick to the recommended version to prevent issues down the line, even though this older one seems to work.

samuelgrf (approved):
LGTM 👍

xTVaser (approved):
Looks good to me as well, Travis-CI build is passing https://travis-ci.org/github/xTVaser/pcsx2-rr/builds/737850344

arcum42:
One thought on this is that, while it works, the default build is 32 bit (currently), and it's rather unlikely to have a 32 bit version of fmt installed, which means you have a warning that looks like a potential problem on the default build of pcsx2. 

`CMake Warning at cmake/SearchForStuff.cmake:237 (find_package):
  Could not find a configuration file for package "fmt" that is compatible
  with requested version "7.0.3".

  The following configuration files were considered but not accepted:

    /usr/lib/cmake/fmt/fmt-config.cmake, version: 7.1.0 (64bit)
    /lib/cmake/fmt/fmt-config.cmake, version: 7.1.0 (64bit)

  CMakeLists.txt:30 (include)
`

It'd be nice if it looked less like an issue, and told you that it was using a built-in version instead.

nE0sIghT:
@arcum42 

It should be fixed now. I silenced standard cmake messages and added custom status messages instead.

arcum42:
Thanks! Looks a lot less like something someone's going to point at when their build fails. ^_^

Issue #2699, opened 2018-11-20
cmake: Installation issues.


orbea:
***PCSX2 version:***
https://github.com/PCSX2/pcsx2/commit/651d287f1da5047cda12ae81de832755c70aaef5

***Description of the issue:***

The following code in `pcsx2/CMakeLists.txt` has the following three problems, the first one is critical and makes it entirely non-working.

https://github.com/PCSX2/pcsx2/blob/651d287f1da5047cda12ae81de832755c70aaef5/pcsx2/CMakeLists.txt#L713

1. It is modifying `${BIN_DIR}/PCSX2` (`/usr/bin/PCSX2`) and not `$DESTDIR/${BIN_DIR}/PCSX2`, if `${BIN_DIR}/PCSX2` is not already installed on the system then it will fail. This results in modifying the old file to be replaced and not the new file to be installed. This can never work because `cmake` does not know the value of `${DESTDIR}` and only `make/ninja/samu/etc.` do. The alternative would be to modify it before its installed, but there doesn't seem to be a good place to do this without modifying `add_pcsx2_executable` in `cmake/Pcsx2Utils.cmake` to not actually install the files which would be messy.
2. Not all users have `/bin/bash`, but it should be entirely safe to use `/bin/sh` here as even with true bourne shells such as `hsh` this should work.
3. Not all users have `sudo` and using `sudo` inside of a cmake installation is a really terrible thing from a package management perspective.

***Last known version to work:***
This was introduced in PR https://github.com/PCSX2/pcsx2/pull/2586.

***PC specifications:***
OS: `Slackware64-current`
`cmake-3.12.4`

lowlyocean:
Does slackware have something similar to debian/postinst? See https://github.com/PCSX2/pcsx2/pull/2638/files

orbea:
For reference here are distros that package pcsx2, it seems most of them are not recent enough versions for this.

lowlyocean:
That is outside my scope of influence. In fact, this Merge is not in a major release so it won't even show up in well-supported distro's (Ubuntu) official repos. It will only show up in daily build PPA

orbea:
I think it will show up for the gentoo 9999 and AUR builds, OpenSuse and slackbuilds.org are also following git checkouts so it will show up there when they are updated. The latter I maintain so this could be potentially soon.

And eventually if and when PCSX2 tags another release other distros will also upgrade.

lowlyocean:
[...]

PR #3840, opened 2020-10-17
Added cmake options to bypass setcap execution.

nE0sIghT:
This call became a problem with distributions (eg Gentoo GNU/Linux) that
builds PCSX2 in sandboxed environment which disallows setcap execution.


orbea:
Related issue. https://github.com/PCSX2/pcsx2/issues/2699

orbea:
Also another related issue, since you are involved with gentoo can you get someone to fix this finally? :)

https://bugs.gentoo.org/672566

PR #3844, opened 2020-10-17
CDVD: Fix iLinkId issues (.nvm)

### LINKED ISSUE #2475: Time Crisis 2/3 Timing issues with some BIOS revisions
***PCSX2 version:***
1.5dev 2400

***PCSX2 options:***
All default, speedhacks off

***Plugins used:***
All defaults

The games "Time Crisis 2" and "Time Crisis 3" fails to load on some BIOS revisions (Tested on 2 differents PS2 slim BIOS)

- PAL 2.0 (rev 14/06/2004) load them fine
- USA 2.2 (rev 10/02/2006) fails to load them with the following error: sce1394SbEui64(): failed -1

Note: set the EE cyclerate to -3 allow the games to work on the 2.2 BIOS.

Load Time Crisis 2 or 3 on a 2.2 BIOS (or other than 2.0).

***Not a regression***

***PC specifications:***
AMD fx8350 GTX 1060 Windows 10 64bits

kozarovv:
Previously iLinkId was writen/overwriten to .nvm only when user not supplied own file.
But when user supplied file with empty iLinkId, then value is never set. So games fail to get correct data.
That implementation return hardcoded ilinkId when read is requested, and correct iLinkId is not found.
fixes #2475

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -1677,6 +1677,18 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
 			case 0x12: // sceCdReadILinkId (0:9)
 				SetResultSize(9);
 				cdvdReadILinkID(&cdvd.Result[1]);
+				if ((!cdvd.Result[3]) && (!cdvd.Result[4])) // nvm file is missing correct iLinkId, return hardcoded one
+				{
+					cdvd.Result[0] = 0x00;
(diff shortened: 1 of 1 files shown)


kozarovv:
I pushed better solution, writing to nvm file when read is requested wasn't brightest idea. 
This solution not touch nvm file, instead return pcsx2 hardcoded iLinkId when nvm file don't have one.

PR #3845, opened 2020-10-18
pcsx2-gui: Internalize VirtualPad and Upgrade NewRecordingFrame

sonicfind:
Simplifies virtualPad access by solely holding the virtualPad window pointers in the g_inputRecording object. This will be useful later down the line when more virtualPads are added.

NewRecordingFrame has improved functionality:
* Users can now only choose to create a savestate-based recording if GS is open.
* Forced file browsing in certain situations.

Also, forces a game to have been loaded before being allowed to mess with input recordings.

### DIFF
--- a/pcsx2/Recording/InputRecording.cpp
+++ b/pcsx2/Recording/InputRecording.cpp
@@ -63,16 +63,26 @@ void SaveStateBase::InputRecordingFreeze()
 
 InputRecording g_InputRecording;
 
-InputRecording::InputRecording()
+InputRecording::InputRecordingPad::InputRecordingPad()
 {
-	// NOTE - No multi-tap support, only two controllers
-	padData[CONTROLLER_PORT_ONE] = new PadData();
-	padData[CONTROLLER_PORT_TWO] = new PadData();
+	padData = new PadData;
 }
 
-void InputRecording::setVirtualPadPtr(VirtualPad* ptr, int const port)
+InputRecording::InputRecordingPad::~InputRecordingPad()
 {
-	virtualPads[port] = ptr;
+	delete padData;
+}
+
+void InputRecording::InitVirtualPadWindows(wxWindow* parent)
+{
+	for (int port = 0; port < 2; ++port)
+		if (!pads[port].virtualPad)
+			pads[port].virtualPad = new VirtualPad(parent, port, g_Conf->inputRecording);
+}
+
+void InputRecording::ShowVirtualPad(const int port)
+{
+	pads[port].virtualPad->Show();
 }
 
 void InputRecording::RecordingReset()
@@ -114,16 +124,16 @@ void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 b
 					inputRec::consoleLog(fmt::format("Failed to read input data at frame {}", frameCounter));
 
 				// Update controller data state for future VirtualPad / logging usage.
-				padData[port]->UpdateControllerData(bufIndex, bufVal);
+				pads[port].padData->UpdateControllerData(bufIndex, bufVal);
 				
-				if (virtualPads[port]->IsShown())
-					virtualPads[port]->UpdateControllerData(bufIndex, padData[port]);
+				if (pads[port].virtualPad->IsShown())
+					pads[port].virtualPad->UpdateControllerData(bufIndex, pads[port].padData);
 			}
 		}
 		else
 		{
 			// Update controller data state for future VirtualPad / logging usage.
-			padData[port]->UpdateControllerData(bufIndex, bufVal);
+			pads[port].padData->UpdateControllerData(bufIndex, bufVal);
 
 			// Commit the byte to the movie file if we are recording
 			if (state == InputRecordingMode::Recording)
@@ -132,8 +142,8 @@ void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 b
 				{
 					// If the VirtualPad updated the PadData, we have to update the buffer
 					// before committing it to the recording / sending it to the game
-					if (virtualPads[port]->IsShown() && virtualPads[port]->UpdateControllerData(bufIndex, padData[port]))
-						bufVal = padData[port]->PollControllerData(bufIndex);
+					if (pads[port].virtualPad->IsShown() && pads[port].virtualPad->UpdateControllerData(bufIndex, pads[port].padData))
+						bufVal = pads[port].padData->PollControllerData(bufIndex);
 
 					if (incrementUndo)
 					{
@@ -147,8 +157,8 @@ void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 b
 			}
 			// If the VirtualPad updated the PadData, we have to update the buffer
 			// before sending it to the game
-			else if (virtualPads[port]->IsShown() && virtualPads[port]->UpdateControllerData(bufIndex, padData[port]))
-				bufVal = padData[port]->PollControllerData(bufIndex);
+			else if (pads[port].virtualPad->IsShown() && pads[port].virtualPad->UpdateControllerData(bufIndex, pads[port].padData))
+				bufVal = pads[port].padData->PollControllerData(bufIndex);
 		}
 	}
 }
@@ -190,11 +200,11 @@ void InputRecording::LogAndRedraw()
 {
 	for (u8 port = 0; port < 2; port++)
 	{
-		padData[port]->LogPadData(port);
+		pads[port].padData->LogPadData(port);
 		// As well as re-render the virtual pad UI, if applicable
 		// - Don't render if it's minimized
-		if (virtualPads[port]->IsShown() && !virtualPads[port]->IsIconized())
-			virtualPads[port]->Redraw();
+		if (pads[port].virtualPad->IsShown() && !pads[port].virtualPad->IsIconized())
+			pads[port].virtualPad->Redraw();
 	}
 }
 
@@ -239,16 +249,16 @@ wxString InputRecording::RecordingModeTitleSegment()
 void InputRecording::SetToRecordMode()
 {
 	state = InputRecordingMode::Recording;
-	virtualPads[CONTROLLER_PORT_ONE]->SetReadOnlyMode(false);
-	virtualPads[CONTROLLER_PORT_TWO]->SetReadOnlyMode(false);
+	pads[CONTROLLER_PORT_ONE].virtualPad->SetReadOnlyMode(false);
+	pads[CONTROLLER_PORT_TWO].virtualPad->SetReadOnlyMode(false);
 	inputRec::log("Record mode ON");
 }
 
 void InputRecording::SetToReplayMode()
 {
 	state = InputRecordingMode::Replaying;
-	virtualPads[CONTROLLER_PORT_ONE]->SetReadOnlyMode(true);
-	virtualPads[CONTROLLER_PORT_TWO]->SetReadOnlyMode(true);
+	pads[CONTROLLER_PORT_ONE].virtualPad->SetReadOnlyMode(true);
+	pads[CONTROLLER_PORT_TWO].virtualPad->SetReadOnlyMode(true);
 	inputRec::log("Replay mode ON");
 }
 
@@ -327,25 +337,25 @@ void InputRecording::FailedSavestate()
 void InputRecording::Stop()
 {
 	state = InputRecordingMode::NotActive;
-	virtualPads[CONTROLLER_PORT_ONE]->SetReadOnlyMode(false);
-	virtualPads[CONTROLLER_PORT_TWO]->SetReadOnlyMode(false);
+	pads[CONTROLLER_PORT_ONE].virtualPad->SetReadOnlyMode(false);
+	pads[CONTROLLER_PORT_TWO].virtualPad->SetReadOnlyMode(false);
 	incrementUndo = false;
 	if (inputRecordingData.Close())
 		inputRec::log("Input recording stopped");
 }
 
-bool InputRecording::Create(wxString FileName, bool fromSaveState, wxString authorName)
+bool InputRecording::Create(wxString fileName, const bool fromSaveState, wxString authorName)
 {
-	if (!inputRecordingData.OpenNew(FileName, fromSaveState))
+	if (!inputRecordingData.OpenNew(fileName, fromSaveState))
 		return false;
 
 	initialLoad = true;
 	state = InputRecordingMode::Recording;
 	if (fromSaveState)
 	{
-		if (wxFileExists(FileName + "_SaveState.p2s"))
-			wxCopyFile(FileName + "_SaveState.p2s", FileName + "_SaveState.p2s.bak", true);
-		StateCopy_SaveToFile(FileName + "_SaveState.p2s");
+		if (wxFileExists(fileName + "_SaveState.p2s"))
+			wxCopyFile(fileName + "_SaveState.p2s", fileName + "_SaveState.p2s.bak", true);
+		StateCopy_SaveToFile(fileName + "_SaveState.p2s");
 	}
 	else
 		sApp.SysExecute(g_Conf->CdvdSource);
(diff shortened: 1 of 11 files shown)


pcsx2/Recording/InputRecording.cpp
```diff
@@ -63,28 +63,51 @@ void SaveStateBase::InputRecordingFreeze()
...
 
-InputRecording::InputRecording()
+InputRecording::~InputRecording()
 {
-	// NOTE - No multi-tap support, only two controllers
-	padData[CONTROLLER_PORT_ONE] = new PadData();
-	padData[CONTROLLER_PORT_TWO] = new PadData();
+	// As to not attempt a double delete from WX
+	newRecordingFrame.release();
+	openFileDialog.release();
```
xTVaser, line 70: What's the scenario that leads to this double delete you speak of?  Once the object is destroyed these smart pointers should get automatically cleaned up should they not?

pcsx2/Recording/InputRecording.cpp
```diff
@@ -63,28 +63,51 @@ void SaveStateBase::InputRecordingFreeze()
...
 {
-	virtualPads[port] = ptr;
+	padData = std::make_unique<PadData>();
 }
 
-void InputRecording::RecordingReset()
+InputRecording::InputRecordingPad::~InputRecordingPad()
+{
+	// As to not attempt a double delete from WX
+	virtualPad.release();
```
xTVaser, line 81: same question here

pcsx2/Recording/InputRecording.cpp
```diff
@@ -63,28 +63,51 @@ void SaveStateBase::InputRecordingFreeze()
...
+}
+
+void InputRecording::InitInputRecordingWindows(wxWindow* parent)
+{
+	if (newRecordingFrame.get() != nullptr)
+		return;
```
xTVaser, line 87: this fundamentally works, but seems like a bit of an anti-pattern = using the initialization of 1/4 of the preceding variables as a check for all of them.  

I'd much rather see checking before initializing each, or have an all-encompassing `bool` to check if Recording GUI elements have already been initialized.

pcsx2/Recording/InputRecording.cpp
```diff
@@ -101,7 +124,7 @@ void InputRecording::ControllerInterrupt(u8& data, u8& port, u16& bufCount, u8 b
 		fInterruptFrame = false;
 
 	// We do not want to record or save the first two bytes in the data returned from the PAD plugin
-	else if (fInterruptFrame && bufCount >= 3 && frameCounter >= 0 && frameCounter < INT_MAX)
+	else if (fInterruptFrame && frameCounter >= 0 && frameCounter < INT_MAX)
```
xTVaser, line 127: this change appears unrelated in the scope of this PR

pcsx2/Recording/InputRecording.cpp
```diff
@@ -220,28 +239,32 @@ wxString InputRecording::RecordingModeTitleSegment()
...
+	pads[CONTROLLER_PORT_ONE].virtualPad->SetReadOnlyMode(true);
+	pads[CONTROLLER_PORT_TWO].virtualPad->SetReadOnlyMode(true);
 	inputRec::log("Replay mode ON");
 }
 
 void InputRecording::SetFrameCounter(u32 newGFrameCount)
 {
-	if (newGFrameCount > startingFrame + (u32)g_InputRecording.GetInputRecordingData().GetTotalFrames())
+	const u32 endFrame = startingFrame + inputRecordingData.GetTotalFrames();
+	if (newGFrameCount >= endFrame)
```
xTVaser, line 258: hmm, why the change to `>=` does `GetTotalFrames` return something different now? (havn't gotten to that point in the review yet) or does this resolve a bug of some sort?

pcsx2/Recording/InputRecording.cpp
```diff
@@ -298,8 +324,11 @@ bool InputRecording::Create(wxString FileName, bool fromSaveState, wxString auth
 	inputRecordingData.GetHeader().SetEmulatorVersion();
 
 	// Set author name
-	if (!authorName.IsEmpty())
-		inputRecordingData.GetHeader().SetAuthor(authorName);
+	{
+		const wxString authorName = newRecordingFrame->GetAuthor();
+		if (!authorName.IsEmpty())
+			inputRecordingData.GetHeader().SetAuthor(authorName);
+	}
```
xTVaser, line 331: these brace brackets appear to be wrapping nothing

pcsx2/Recording/InputRecording.cpp
```diff
@@ -308,17 +337,20 @@ bool InputRecording::Create(wxString FileName, bool fromSaveState, wxString auth
 	SetToRecordMode();
 	g_InputRecordingControls.DisableFrameAdvance();
 	inputRec::log("Started new input recording");
-	inputRec::consoleLog(fmt::format("Filename {}", std::string(FileName)));
+	inputRec::consoleLog(fmt::format("Filename {}", std::string(inputRecordingData.GetFilename())));
 	return true;
 }
 
-bool InputRecording::Play(wxString fileName)
+int InputRecording::Play()
```
xTVaser, line 344: If we need to return more than 2 states, I'd rather it return an `enum class` so that is reads better on the consuming side.  Returning hard-coded `int`s requires knowing the internals of this function.  Though maybe it's an indication of this function trying to do too much on it's own

pcsx2/Recording/InputRecording.cpp
```diff
@@ -361,6 +393,27 @@ bool InputRecording::Play(wxString fileName)
...
+}
+
+bool InputRecording::GoToFirstFrame()
+{
+	if (inputRecordingData.FromSaveState())
+	{
+		if (!wxFileExists(inputRecordingData.GetFilename() + "_SaveState.p2s"))
+		{
+			recordingConLog(wxString::Format("[REC]: Could not locate savestate file at location - %s_SaveState.p2s\n",
+												inputRecordingData.GetFilename()));
```
xTVaser, line 406: use the new `inputRec::consoleLog`

pcsx2/gui/AppInit.cpp
```diff
@@ -77,19 +72,6 @@ void Pcsx2App::OpenMainFrame()
 	DisassemblyDialog* disassembly = new DisassemblyDialog( mainFrame );
 	m_id_Disassembler = disassembly->GetId();
 
-#ifndef DISABLE_RECORDING
```
xTVaser, line 80: Seems like the new location for initializing the recording windows is now a side-effect of the menubar being created?  Not sure how I feel about that, that feels unexpected and should probably remain here where other frames (like the debugger/console) are made.

pcsx2/gui/FrameForGS.cpp
```diff
@@ -134,18 +136,23 @@ void GSPanel::InitRecordingAccelerators()
...
+		GetAssociatedKeyCode("InputRecordingModeToggle"),
 		g_InputRecording.IsActive());
 
 	inputRec::consoleLog("Initialized Input Recording Key Bindings");
 }
 
+wxString GSPanel::GetAssociatedKeyCode(const char* id)
+{
+	return m_Accels->findKeycodeWithCommandId(id).toTitleizedString();
+}
```
xTVaser, line 154: 👍 though I wonder if this could go into the `AcceleratorDictionary` class and be used generically outside of this file.  Also `GetAssociatedKeyCode` is a bit misleading as it's already converting the KeyCode into it's string representation.  This is mostly a nit, feel free to ignore.

pcsx2/gui/GlobalCommands.cpp
```diff
@@ -531,6 +538,15 @@ namespace Implementations
...
 	}
 
+	void GoToFirstFrame()
+	{
+		if (g_Conf->EmuOptions.EnableRecordingTools && g_InputRecording.IsActive())
+		{
+			if (!g_InputRecording.GoToFirstFrame() && wxGetApp().HasGUI())
+				sMainFrame.RecordingMenuReset();
+		}
+	}
```
xTVaser, line 548: needs to be wrapped in `ifndef`

pcsx2/gui/MainFrame.cpp
```diff
@@ -57,11 +61,18 @@ void MainEmuFrame::UpdateStatusBar()
 {
 	wxString temp(wxEmptyString);
 
-	if (g_Conf->EnableFastBoot)
-		temp += "Fast Boot - ";
+#ifndef DISABLE_RECORDING
+	if (g_InputRecording.IsActive() && g_InputRecording.GetInputRecordingData().FromSaveState())
+		temp += "Base Savestate - " + g_InputRecording.GetInputRecordingData().GetFilename() + "_SaveState.p2s";
+	else
```
xTVaser, line 67: I'm not sure how I feel about modifying the system menu.  I mean it makes sense for what you're trying to accomplish.  But I can't help but feel like a better way to go would be to:
- gray out the boot option / disable it
- add a base savestate option to the `Input Record` menu.
To me it feels better UX-wise, but I could very well be wrong.

pcsx2/gui/MainFrame.cpp
```diff
@@ -57,11 +61,18 @@ void MainEmuFrame::UpdateStatusBar()
...
+#endif
+	{
+		if (g_Conf->EnableFastBoot)
+			temp += "Fast Boot - ";
 
-	if (g_Conf->CdvdSource == CDVD_SourceType::Iso)
-		temp += "Load: '" + wxFileName(g_Conf->CurrentIso).GetFullName() + "' ";
+		if (g_Conf->CdvdSource == CDVD_SourceType::Iso)
+			temp += "Load: '" + wxFileName(g_Conf->CurrentIso).GetFullName() + "' ";
+	}
```
xTVaser, line 75: i think these brackets will cause an issue if the `#ifndef` isn't present.  We'll have dangling brackets without an associated else...maybe that's valid considering my previous comment but it reads strangely currently.

pcsx2/gui/MainFrame.cpp
```diff
@@ -92,6 +103,12 @@ void MainEmuFrame::UpdateCdvdSrcSelection()
 			jNO_DEFAULT
 	}
 	sMenuBar.Check(cdsrc, true);
+#ifndef DISABLE_RECORDING
+	if (g_InputRecording.IsActive())
+		ApplyFirstFrameStatus();
+	else
+#endif
+		ApplyCDVDStatus();
```
xTVaser, line 111: indentation here seems to be messed up

pcsx2/gui/MainFrame.cpp
```diff
@@ -483,16 +500,18 @@ void MainEmuFrame::CreateCaptureMenu()
...
+	m_menuRecording.Append(MenuId_Recording_Play, _("Play"), _("Playback an existing input recording."))->Enable(false);
 	m_menuRecording.AppendSeparator();
 	m_menuRecording.Append(MenuId_Recording_TogglePause, _("Toggle Pause"), _("Pause or resume emulation on the fly."))->Enable(false);
 	m_menuRecording.Append(MenuId_Recording_FrameAdvance, _("Frame Advance"), _("Advance emulation forward by a single frame at a time."))->Enable(false);
 	m_menuRecording.Append(MenuId_Recording_ToggleRecordingMode, _("Toggle Recording Mode"), _("Save/playback inputs to/from the recording file."))->Enable(false);
 	m_menuRecording.AppendSeparator();
 	m_menuRecording.Append(MenuId_Recording_VirtualPad_Port0, _("Virtual Pad (Port 1)"));
 	m_menuRecording.Append(MenuId_Recording_VirtualPad_Port1, _("Virtual Pad (Port 2)"));
+	if (g_Conf->EmuOptions.EnableRecordingTools)
+		g_InputRecording.InitInputRecordingWindows(this);
```
xTVaser, line 514: once again, I'm not sure if this is where we want to do this.  This function is supposed to be for initializing the wxMenuBar, I don't think we want to overload it with other tasks

pcsx2/gui/MainFrame.cpp
```diff
@@ -825,9 +851,9 @@ void MainEmuFrame::initializeRecordingMenuItem(MenuIdentifiers menuId, wxString
 	item.Enable(enable);
 }
 
-void MainEmuFrame::enableRecordingMenuItem(MenuIdentifiers menuId, bool enable)
+void MainEmuFrame::enableRecordingMenuItem(int menuId, bool enable)
 {
-	wxMenuItem& item = *m_menuRecording.FindChildItem(menuId);
+	wxMenuItem& item = *m_menuRecording.FindItem(menuId);
```
xTVaser, line 856: can i get some context on these changes.  From my perspective using the `enum` added some extra type-safety, and `FindChildItem` makes sense as it's specifically for finding menu sub-menu elements, is this to support finding nested menu elements in the future?

pcsx2/gui/MainFrame.h
```diff
@@ -255,6 +261,7 @@ class MainEmuFrame : public wxFrame,
 	void Menu_Recording_New_Click(wxCommandEvent& event);
 	void Menu_Recording_Play_Click(wxCommandEvent& event);
 	void Menu_Recording_Stop_Click(wxCommandEvent& event);
+	void ApplyFirstFrameStatus();
```
xTVaser, line 264: does this need to be wrapped in an `ifndef`?

pcsx2/gui/MainMenuClicks.cpp
```diff
@@ -945,65 +1019,93 @@ void MainEmuFrame::Menu_Recording_New_Click(wxCommandEvent& event)
...
-			g_InputRecordingControls.Resume();
-		return;
-	}
 
-	wxString path = openFileDialog.GetPath();
 	const bool recordingLoaded = g_InputRecording.IsActive();
-	if (!g_InputRecording.Play(path))
+	switch (g_InputRecording.Play())
 	{
+	case 1:
```
xTVaser, line 1043: rather switching on enum values as mentioned above

pcsx2/gui/MainMenuClicks.cpp
```diff
@@ -945,65 +1019,93 @@ void MainEmuFrame::Menu_Recording_New_Click(wxCommandEvent& event)
...
 	{
 		m_menuRecording.FindChildItem(MenuId_Recording_New)->Enable(false);
 		m_menuRecording.FindChildItem(MenuId_Recording_Stop)->Enable(true);
+		m_menuRecording.FindChildItem(MenuId_Recording_ToggleRecordingMode)->Enable(true);
 	}
-	sMainFrame.enableRecordingMenuItem(MenuId_Recording_ToggleRecordingMode, g_InputRecording.IsActive());
+	ApplyFirstFrameStatus();
+}
+
+void MainEmuFrame::ApplyFirstFrameStatus()
```
xTVaser, line 1061: wrapped in an ifndef?

xTVaser (commented):
Overall this looks good, lots of nice improvements. With that said, I don't mind reviewing/testing large PRs like this but it would make things a lot simpler and faster if we could reduce the surface area per-PR. ie, this PR changes several things that could be broken apart:
- reducing core footprint
- adding new GUI options
- looks like a handful of bug fixes / alterations inside the recording logic itself, etc.

pcsx2/Recording/InputRecording.cpp
sonicfind: When the new operator is called on a wxWindow object, the point of the new object gets sent into a list of that all wxWindow pointers. WX deletes every pointer in that list, so if the smart pointer deletes it first...

pcsx2/gui/MainFrame.cpp
sonicfind: Fair.

pcsx2/Recording/InputRecording.cpp
sonicfind: Well luckily, it's getting reworked back into a bool thanks to changes to MainFrame::SetRecordingMenu (or RecordingMenuReset in the previous iteration).

xTVaser (approved):
Tested the following:
- VirtualPad still opens and functions correctly with new adjustments
- Browsing for the recording's file-path was required, even manually deleting the field properly updated the UI
- Was only able to make a power-on or "boot" recording with GS closed
- New/Play options were disabled until i booted a game for the first time upon opening the emulator.

Looks good to me.

PR #3858, opened 2020-10-23
gsdx: properly init clut ptrs v2

tadanokojin:
If testers test it they will be rehired.

GS reads lower 3 bits of PSM to determine if format is 8/4 bit and thus whether to write the clut into the clut buffer. This PR correctly emulates this behavior.

I added a log for ignored writes but I might remove it if it becomes to noisy.

[test (zip)](https://github.com/PCSX2/pcsx2/files/5427477/clut_reload.zip)
PR mimics real hw and reloads the clut for the following formats:

**As 8bit:** `0x3`, `0xB`, `0x13`, `0x1B`, `0x23`, `0x2B`, `0x33`, `0x3B`
**As 4bit:** `0x4`, `0xC`, `0x14`, `0x1C`, `0x24`, `0x2C`, `0x34`, `0x3C`

TODO: CSM2 and undocumented formats (FFX uses `0x9` as an alias for CT24)

### DIFF
--- a/plugins/GSdx/GSClut.cpp
+++ b/plugins/GSdx/GSClut.cpp
@@ -40,31 +40,46 @@ GSClut::GSClut(GSLocalMemory* mem)
 	{
 		for(int j = 0; j < 64; j++)
 		{
-			m_wc[0][i][j] = &GSClut::WriteCLUT_NULL;
+			// The GS seems to check the lower 3 bits to tell if the format is 8/4bit
+			// for the reload.
+			const bool eight_bit = (j & 0x7) == 0x3;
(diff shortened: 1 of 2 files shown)


refractionpcsx2 (approved):

refractionpcsx2:
CSM2 might be a pain to test, I don't know of any games which actually use CSM2, i suspect the reload CLUT behaviour may be the same as CSM1 as that expects IDTEX4 or IDTEX8 also, just the storage layout is more linear.

tadanokojin:
I'm about to perform what's called a pro-neko move.

PR #3860, opened 2020-10-23
FPU: Remove FPU Compare Hack

kozarovv:
Little explanation why we can remove this hack. At the time hack was created full clamping mode in PCSX2 preserved sign by default, so acted like current Extra + Preserve Sign. This made it incompatible with both titles that hack were needed. Now PCSX2 have FULL clamping mode that is able to not preserve sign, so hack is no longer required, we can just use full clamping instead.

This PR remove hack, adjust GameDB according to that change, and rename fpuFloat4 function to fpuFloat3.
Last change is because fpuFloat3 was just wrapper for fpuFloat4 with added check for compare hack. So is safe to remove it at all. 
Additionally fpuFloat2 was only function that called fpuFloat4 directly, so that one call was changed to fpuFloat3 to respect previous changes.


GovanifY (approved):
LGTM, is there any testing/things I'm not aware of regarding this PR? Otherwise I'm fine with merging it

atomic83GitHub:
You can merge it, every games which needed this hack are fixed with full clamping.

GovanifY (approved):
rebased and switched a /bin/bash to /bin/sh, merge time

PR #3879, opened 2020-10-31
Revert commit 7dc3b25

kenshen112:
This is an attempt to fix Fatal Frame or Project Zero 2 / 3 voice issues. This doesn't fix the audio crackle across rooms. This reverts https://github.com/PCSX2/pcsx2/commit/7dc3b25ddf5ae789bc0c897866eaf6af0c35cdef


refractionpcsx2:
After looking in to this, I don't think this is a solution as this just puts it back to Chaos Legion being broken.  I *could* fix both but I don't think that behaviour is correct either, so I'm going to say we leave this until some testing is done to come up with a proper solution.

PR #3884, opened 2020-11-01
Don't call DEV9open if isSuspendedthread is false in OnResumeThread()

TheLastRar:
During emulation, if you open a plug-in's config menu (such as changing lilypad's settings) OnSuspendInThread is not called, meaning DEV9close is not called, meaning DEV9's RxThread is still running

When the plug-in's config menu is closed, OnResumeInThread() is called with isSuspended set false

Before, we would always call DEV9open(), which would create a new RxThread without stopping the old thread (as the code never expected DEV9open to be called more than once without a DEV9close)

Now we check isSuspendedthread before calling DEV9open, preventing the above situation

### DIFF
--- a/pcsx2/System/SysCoreThread.cpp
+++ b/pcsx2/System/SysCoreThread.cpp
@@ -316,7 +316,8 @@ void SysCoreThread::OnResumeInThread(bool isSuspended)
 		DoCDVDopen();
 	FWopen();
 	SPU2open((void*)pDsp);
-	DEV9open((void*)pDsp);
+	if (isSuspended)
+		DEV9open((void*)pDsp);
 }
 
 
--- a/pcsx2/System/SysCoreThread.cpp
+++ b/pcsx2/System/SysCoreThread.cpp
@@ -303,21 +303,22 @@ void SysCoreThread::ExecuteTaskInThread()
 void SysCoreThread::OnSuspendInThread()
 {
 	GetCorePlugins().Close();
-	DEV9close();
 	DoCDVDclose();
 	FWclose();
 	SPU2close();
+	DEV9close();
 }
 
 void SysCoreThread::OnResumeInThread(bool isSuspended)
 {
 	GetCorePlugins().Open();
 	if (isSuspended)
+	{
 		DoCDVDopen();
+		DEV9open((void*)pDsp);
+	}
 	FWopen();
 	SPU2open((void*)pDsp);
-	if (isSuspended)
-		DEV9open((void*)pDsp);
 }
 
 


TheLastRar:
Totally didn't accidentally hit send before I finished writing the title/description

On a side note, is something similar required for SPU, given that it previously was also a plugin?

GovanifY:
@TheLastRar As far as I can remember I believe I took that into account when merging SPU. You/I can always double check but it doesn't seem like it caused any trouble, and it wasn't exactly lightly tested.

I remember stumbling upon this issue but forgot about it since I was afk from the project for a few weeks

pcsx2/System/SysCoreThread.cpp
```diff
@@ -316,7 +316,8 @@ void SysCoreThread::OnResumeInThread(bool isSuspended)
 		DoCDVDopen();
 	FWopen();
 	SPU2open((void*)pDsp);
-	DEV9open((void*)pDsp);
+	if (isSuspended)
+		DEV9open((void*)pDsp);
```
GovanifY, line 320: We already have an `if (isSuspended)` for CDVD, wouldn't it make more sense to move everything there?

GovanifY (changes requested):

pcsx2/System/SysCoreThread.cpp
TheLastRar: I know that plug-ins where always suspended/ resumed in a specific order (for the most part)
I think this was only important for plug-ins that where subclassing the GSDX window (which I think lilypad & my usb plugin did) 

I pretty sure DEV9 would be unaffected, but I was keeping that convention

TheLastRar: If you deem it unimportant I can just move it to the above condition

TheLastRar:
> @TheLastRar As far as I can remember I believe I took that into account when merging SPU. You/I can always double check but it doesn't seem like it caused any trouble, and it wasn't exactly lightly tested.

Fair enough, seems to be fine as is for SPU if no issues cropped up during testing

pcsx2/System/SysCoreThread.cpp
GovanifY: That's indeed a picky situation, but in our case we want critical plugins for IO, eg CDVD, to open before plugins that read from said IO, eg SPU2/GS.
You basically always want CDVD and Pad to be initialized ASAP, then initialize non critical plugins then at the end SPU2 and GS, probably.

DEV9 doesn't matter either way afaik, you can probably test it to make sure though.

TheLastRar: Additionally, I can also move DEV9 in OnSuspendInThread to preserve that convention

GovanifY: It would be appreciated, thanks!

GovanifY (approved):
LGTM for merging+squash, might wanna test first though.

RedDevilus (approved):
It doesn't nag anymore about RX-thread and i can make network config, host a server.

TheLastRar:
The changes this PR adds is already included in https://github.com/PCSX2/pcsx2/commit/a1c4077d83663282e1efd15b6fdc0e102bb95444

The reason the rx isn't nagging you is because, when this pr was made, the rx thread just silently leaked, rather than raising an exception as it does now (due to the switch to rx

GovanifY:
Oh shit I'm sorry, I didn't mean to appropriate your work in my commits @TheLastRar . Thanks for the great work on HDD et al and sorry again ._.

PR #3885, opened 2020-11-01
misc: Replace Fall through comments with c++17 attribute.

lightningterror:
Silences warnings about not having breaks on switch cases.
Added a few missing ones too.

Previously comments worked on gcc only.


arcum42:
Seems fine to me. Only reason I didn't go this route originally was because we weren't on C++17 yet.

PR #3886, opened 2020-11-01
misc: Remove merged plugins leftover stuff.

lightningterror:


arcum42:
I believe generally everything should be fine here.

PR #3891, opened 2020-11-03
gsdx-hw: Don't always set MaxDepth on ps/fs.

lightningterror:
Value will be read only when zclamp is enabled. It will avoid an extra upload to buffer maybe.


Helium-4:
Seems to be about 0.5 to 2 fps faster on the opening mission of Armored Core 3 which seems to be heavily bottlenecked by depth emulation ( about 41fps with depth emulation enabled, about 70fps without)

Issue #3155, opened 2019-11-14
[GSdx HW] Xenosaga III ntsc Error graphics on surfaces in the last dungeon


BlackTelomeres:
***PCSX2 version:***
PCSX2 1.5.0-dev-3302 - 20191023012101 - compiled on Oct 23 2019

***PCSX2 options:***
Default emulation settings (2 - safe)

***Plugins used:***
GSDX OGL, AVX2 and the defaults

Hardware mode OGL
Native res and up
Mipmapping, automatic CRC, accurate date and accurate blending all set as the defaults, but changing them up to full does not fix it either. Cannot find a HW hack that fixes it either.

Game generally needs the hacks disable safe features, merge sprite, preload frame data, and half pixel offset: special (texture) to look its best through most of the game in OGL hardware mode, but those do not fix this, nor does turning them off.

Virtually the same error exists in DX11 HW mode.

***Description of the issue:***
Error garbage graphics present in some surfaces in the last dungeon, most or all of the surfaces being somewhat transparent and "crystalline."
Different but yet similar looking (enough for them to probably all be the same root cause) forms of it are present throughout the last dungeon in the game.
https://ibb.co/LJ2nLC4 - example of the error
https://ibb.co/bvc93WP - what the same place looks like in software mode (correct)

***How to reproduce the issue:***
A GSDump is attached to this post.

A memory card save is also attached if needed.
Load the 36th save in the memory card, go to the right, then south/
backwards (towards the camera) and through the door and keep walking towards the camera and you'll run into one of the walls of it. You can run from any enemy battles encountered with the escape command. There are also more examples further on in the dungeon which can be reached by going backwards through the dungeon from the 37th save.

***Last known version to work:***
Most likely has never worked correctly in dx11 or ogl hardware mode given what I am seeing on the compatibility wiki, works in software mode

***PC specifications:***
Intel i5 6600, 8GB RAM, GTX 1060 3GB, Windows 8.1pro

[xenosaga 3 last dungeon graphics bug gsdump.zip](https://github.com/PCSX2/pcsx2/files/3844061/xenosaga.3.last.dungeon.graphics.bug.gsdump.zip)
[xenosaga 3 mcd001 for last dungeon graphics bug.zip](https://github.com/PCSX2/pcsx2/files/3844063/xenosaga.3.mcd001.for.last.dungeon.graphics.bug.zip)

MetrosexualGarbodor:
Some updated dumps for this issue in case they're needed.

[...]

Issue #2026, opened 2017-07-26
Wild Arms 4 PAL (SLES-54239) - Not showing refresh rate choice screen


LeCharlatan:
Hello,

Here are basics informations :

***PCSX2 :***

**Build :** Dev PCSX 1.5.0~git201707251955 (PPA PCSX2 Team Daily Builds)

***PCSX2 options :***

**GSDX:** OGL Backend
**Speed hacks :** OFF

***GSDX Options :***

**Texture Filtering :** Bilinear (PS2)
**Large Framebuffer :** ON (Checked)
**Internal Resolution :** 2x Native
**Anisotropic Filtering :** OFF
**Blending Unit Accuracy :** High (Recommended high-end PC)
**Automatic CRC level :** Partial (OpenGL Recommended)
**Mipmapping (Insert) :** Basic (Fast)
**Mipmapping :** ON (Checked)

**Description of the issue:**
PCSX2 isn't showing choice screen about refresh rate monitor in OGL HW (maybe just not showing caracters only, dunno)
![screenshot_20170726_161216](https://user-images.githubusercontent.com/22820335/28629558-12245802-7228-11e7-8fe2-5c84c55fb076.png)

But it does on OGL SW
![screenshot_20170726_161408](https://user-images.githubusercontent.com/22820335/28629568-1ac38762-7228-11e7-9f3f-71fc7e60d1c6.png)

***How to reproduce the issue:***
Just launch the game with OGL HW.
About DirectX 9/11 I've no idea since I'm not able to use it on Linux

***Last known version to work:***
None

***PC specifications :***

**OS :** Kubuntu 16.04 (Xenial)
**CPU :** Intel i5 4690K (Haswell)
**GPU :** Nvidia GTX 1070 + Proprietary drivers 381.22
**RAM :** 32Go DDR3-1600

**Little note :** Since @gregory38 told me to report bugs as soon as I find one, I'm reporting ! :p

ssakash:
Could you provide a [GS Dump](http://forums.pcsx2.net/Thread-How-to-create-a-proper-GS-dump) of the issue?

LeCharlatan:
Off course, here it is 

Single frame is enough I think... or hope

[gsdx_20170726175900.zip](https://github.com/PCSX2/pcsx2/files/1177239/gsdx_20170726175900.zip)

DIKH3D:
[...]

Issue #3694, opened 2020-09-14
Driving Emotion Type-S - severe problems


FMecha:
***PCSX2 version:***
pcsx2-v1.7.0-dev-233-g6229b204f-windows-x86

***PCSX2 options:***
Mostly default with Safest speedhack and MTVU

***Plugins used:***
GSdx32-AVX2 1.1.0 (Direct3D 11, mostly default settings)
(all other plugins are defaults))

***Description of the issue:***
This game's problems are already well-known that it's marked only in-game, but I've reporting it here for documentation purposes:

- Using Hardware mode will lead to corrupted rainbows on the top of the race screen. [GS Dump](https://github.com/PCSX2/pcsx2/files/5215489/DET-S.GS.Dump.zip)
- Urban Highway and Provence have severe physics problems - this prevents you from racing in Urban Highway as the game disqualifies you at 3 countdown due to the player being out of bounds (just watch the attract demo to see how it happens) while Provence's physics fail after one or two turns. (Suzuka East works fine though - I haven't tried other courses since they're locked)
![gsdx_20200914071937](https://user-images.githubusercontent.com/68824023/93032649-ebb2a900-f65c-11ea-9dd0-90857fc661e3.png)
![gsdx_20200914073652](https://user-images.githubusercontent.com/68824023/93032763-6b407800-f65d-11ea-8a9f-956d6618156c.png)
![gsdx_20200914072451](https://user-images.githubusercontent.com/68824023/93032766-6da2d200-f65d-11ea-8c08-4c2a34ab1e1d.png)
- Loading West Coast will cause the game to return to BIOS/browser, with the following console log (there's also failures to load course data files? every track suffer from this but only West Coast causes a reset):
```
can't open file -> data/monaco/monaco.l01 [0]
file size 32904
file size 2060
file size 292
file size 24676
file size 56
file size 140
file size 518
file size 16228
file size 152
file size 128
infinity light info.
[0] -0.0e-18 -0.0e+20 0
lens flara info.
[0] 0.0e+1 0 0.0e-18
wall fname [monaco.wal]
file size 2008
file size 461608
### talloc_info addr:1f8f300 size:70c00
tex num 50
tex 128 64
### talloc_info free
file size 1119656
### talloc_info addr:1eee880 size:111680
tex num 218
tex 64 64
### talloc_info free
file size 3452552
### talloc_info addr:1cb4f80 size:34af80
tex num 224
138 100
169 100
mipmap start
[...]

Issue #2409, opened 2018-04-27
Chaos Legion - Glitched shadows under summoned legions


Krusher97:
***PCSX2 version:***
v1.5.0-dev-2354-g1d7cef08a

***PCSX2 options:***
default

***Plugins used:***
default

***Description of the issue:***
The shadows under legions are not rendered correctly in the hardware renderer
![gsdx_20180427201457](https://user-images.githubusercontent.com/25664221/39369529-1645bba6-4a2c-11e8-9d97-bc9a992a03ae.png)
[gsdx_20180427201457.zip](https://github.com/PCSX2/pcsx2/files/1955651/gsdx_20180427201457.zip)

***How to reproduce the issue:***
While ingame press L1 to summon your legion in Hardware renderer.

***Last known version to work:***
none

***PC specifications:***
CPU:i5 6500 3.2Ghz
GPU:Nvidia GeForce 750 Ti
OS:Windows 7 Ultimate 64-bit SP1

MrCK1:
Works fine in OGL/DX11 SW, DX11 HW is totally busted 😛

Krusher97:
It happens only in HW renderers. The screenshot is from the OGL HW renderer.

lightningterror:
None of the hw hacks have an effect ?

Krusher97:
Nope

knight-ryu12:
I have same problem with NTSC-J release of Chaos Legion(SLPM-65249)
CRC = 5E191B9C
https://drive.google.com/open?id=1lhGxr2g1IE2obIDl6fyhImgVzEdqFD9x

MrCK1:
Still exists on latest git

Peduls:
Tested with the latest dev build. Sad to see this issue still exists.

refractionpcsx2:
This is resolved with Tex in Rt enabled, we will need to add this to the GameDB to automatically turn it on once #5593 is merged

PR #3895, opened 2020-11-03
GS-hw, TC: improve tex in rt.

iMineLink:
I improved the tex in rt code which logic is now encapsulated in the `GSTextureCache::ComputeSurfaceOffset` method.
The method considers two surfaces A and B, and computes the offset from B to A in terms of B coordinates.
In the tex in rt case, A is the Source object and B is the Target object.
We want to know the offset from B to A in order to associate to A the correct portion of data from B.

The improved logic is able to fix/improve:

- ~~Burnout 3 black sky (clouds still missing, sky texture not invalidated to GS local memory so black sky reappears if TC is cleared, better solution in #4094 )~~
- Xenosaga 3 last dungeon graphic bug (related to #3155)
- Pause menu in The Getaway
- Indiana Jones and the Emperor's Tomb HW mode
- PAL/NTSC selection menu in Wild Arms 4 HW mode (related to #2026)
- Beyond Good and Evil main character shadow

The logic can be enabled via .ini file (`UserHacks_TextureInsideRt = 1`) for non Jak games, or from the GS HW hacks settings GUI.

image

Due to the nature of GS-hw TC, the logic cannot be always enabled without causing regressions.

The `GSTextureCache::ComputeSurfaceOffset` method is intended as future proof, potentially useful for detecting overlaps between EE writes and cached Targets, or between cached Targets alone.

Bonus 1:

The order in which Target and Source are searched in the cache is inverted, now:

* The lookup source is executed, the call sees the target drawn in the previous draw as MRU
* The lookup target is executed

Previously, the lookup source call saw the current target as MRU framebuffer, but this was not correct as that framebuffer was not written yet in the current draw.
This fixed in a decent way the misbehaviour of Test Drive Unlimited with tex in rt.

Bonus 2:

Targets are now readback from LRU to MRU (older data is overwritten with newer data).
Did not fix any particular issue, but it is more correct behaviour nevertheless.

Bonus 3:

~~The `GSTextureCache::ComputeSurfaceOffset` is reused to detect draws that invalidate partially a framebuffer with increased precision in Jak 2/3/X to fix striped and shimmering textures and wheels rendering in Jak X.~~
~~Unfortunately some games do not digest this change really well, so this improved logic is executed only for the games cited previously.~~
POSTPONED to future PR.

Bonus 4:

Partially dirty targets are now searched for valid data.


plugins/GSdx/Renderers/HW/GSTextureCache.cpp
```diff
@@ -305,93 +301,33 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
...
-					// Only PSMCT32 to limit false hits
-
-					// Check if it is possible to hit with valid <x,y> offset on the given Target
-					// Fixes Jak eyes rendering
-
-					bool valid_offset_may_exist = true;
-
-					// CACHE SEARCH: <x,y> offset
-					for (auto& el : m_texture_inside_rt_cache)
+				} else if (bw == t->m_TEX0.TBW && psm == PSM_PSMCT32 && t->m_TEX0.PSM == psm && t->m_TEX0.TBP0 < bp && t->m_end_block >= bp) {
```
iMineLink, line 304: Do anybody think the BW equality check makes sense?

NewMoonHD:
Test Drive Unlimited didn't favor it a lot. Only HW rendering is affected (SW Renderer is safe for now)
[gsdx_20201104120707.gs.zip](https://github.com/PCSX2/pcsx2/files/5490270/gsdx_20201104120707.gs.zip)

tadanokojin:
Hello imine, good work. #563 improved and the sky issue is now fixed for Burnout Revenge. Black is still experiencing sky issues. I fear that it will be hard to debug however given the nature of it using the Burnout engine. I need a rather large dump starting at loading the level and ending when the sky is first visible. Perhaps @refractionpcsx2 can provide me with that so I can get you more details.

That being said, the reflections on the cars are still flickering. On master without tex in rt I was unable to reproduce the flickering. However, on master with tex in rt enabled the flickering is observed and worse. So I suspect there is some tuning in tex in rt for this game that is needed but this is an improvement over master's tex in rt implementation for sure.
gsdx_20201105143203
[gsdx_20201105143203.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/5496606/gsdx_20201105143203.gs.xz.zip)

Additionally, #3694 has seen improvement with this PR. However, it still exhibits some problems. I can get you a dump if this game if you need. This is more straightforward to debug than the Burnout games.

legend800:
**Burnout 3:**

Sky is broken on first load:
2020-11-06 22_28_21-Slot_ 0 _ Speed_  82% (48 95) _ Limiter_ Normal _ OGL _ 3200x2240 _ Automatic _

If you toggle to Software render:
2020-11-06 22_25_48-Slot_ 0 _ Speed_ 100% (59 94) _ Limiter_ Normal _ SW _ 640x448 _ Automatic _ ${o

Then back to HW, it's finally correct (until you return to menus):
2020-11-06 22_26_34-Slot_ 0 _ Speed_  89% (53 29) _ Limiter_ Normal _ OGL _ 3200x2240 _ Automatic _

**Burnout Dominator:**
Same, except if you switch to Software, sky is black and toggling to HW doesn't fix. 

2020-11-06 22_36_08-Slot_ 0 _ Speed_  78% (46 95) _ Limiter_ Normal _ OGL _ 3200x2240 _ Automatic _

tadanokojin:
I will try to do some debugging today and provide some information about Burnout/Black.

iMineLink:
Thanks kojin, I agree the Burnout engine is not easy to debug unfortunately. I saw previously on discord that Burnout Dominator was working though, but now it seems not, maybe due to the recent PR update.

Maybe having also a gsdump for #3694 can be of help.

tadanokojin:
Yeah I suspect that will be easier and will give you something to look at in the meantime
[driving emotion s.zip](https://github.com/PCSX2/pcsx2/files/5505155/driving.emotion.s.zip)

You're looking for the draw where depth/color gets downsized. They're going to copy them into a temp buffer and color is going to be right under depth but using the same address, then sampled back later (after EE processing) from a different address.

iMineLink:
The replay of Driving Emotion S you gave me looks fine here.

image

Here the first draw call that uses tex in rt:

```
T:Oth	ID:258	S:Cache	=> TC: Lookup Target(Color) 1280x1024, hit: 2 (0x800, C_32)
T:Oth	ID:258	S:Cache	=> TC: Lookup Source <0,0 => 128,128> (0x1d80, C_32, BW: 2)
T:Oth	ID:258	S:Cache	=> TC: ComputeSurfaceOffset - Cached HIT element (size 1), [B] BW 2, PSM C_32, BP 0x1c80 (END 0x1e7f) + OFF <0,128 => 128,256> ---> [A] BP 0x1d80 (END: 0x1e7f).
T:Oth	ID:258	S:Cache	=> TC: dst Color hit (full): 118 (0x1d80, C_32)
T:Oth	ID:258	S:Cache	=> TC: Remove Src Texture: 119 (0x800)
```

tadanokojin:
Alright, it might have been fixed since you added new commits. Let me give it a shot.

iMineLink:
The new commits were:

1. fixing some bug in the offset computation itself (Jak was not working anymore lol)
2. doing the tex in rt search in a second round in LookupSource w.r.t. same BP Source-Target check; this fixed test drive unlimited, where the tex in rt was stealing the correct target lookup (same BP); the reason behind this is because the TC supports overlapping targets for the same PSM, which is a very tricky feature to have along with tex in rt.

tadanokojin:
Looks like there's an issue with it over time:
gsdx_20201107122459
[DES.zip](https://github.com/PCSX2/pcsx2/files/5505204/DES.zip)
I start seeing graphical issues as the race continues. Here you can see the artifacting at the top. Not sure if a dump can help you with this one though, because the issue is different in my dump vs what I saw when I dumped it. An issue of debugging cache issues like this I'm sure.

RedDevilus:
HW mode pics:
gsdx_20201107211953
SLES-54681 PAL - Burnout dominator on commit 82438e0
Black sky when you change from default settings (even including SW mode)
gsdx_20201107212126

SW mode pics:
gsdx_20201107212255
gsdx_20201107212356

When you restart race the sky looks fixed but the car dissapears:
gsdx_20201107212520

But then it recovers on both SW and HW:
gsdx_20201107212635

No idea why it's inconsistent on testing situations. (upscaling / native resolution + Switch between SW and HW mode + starting fresh and restarting races)
You do see a cool lightning halo effect at a certain distance (right side of rails and partially trees on picture:
gsdx_20201107212812

tadanokojin:
Okay, testers are obviously confused so let me explain.

The Burnout engine is a special case for this. The game renders the sky to a texture. It then samples that texture back with a different offset. This means that the part of the texture it wants for the sky will be at a different memory address than where it was rendered. This is the problem tex in rt solves.

However, unlike other examples (jak eyes, driving emotion s) where the texture is rendered to every frame, Burnout only renders the texture when the level is loading. Additionally, sw rendering doesn't experience this problem because it's a different cache implementation. So when you load the level in software, it renders the texture to GS memory and then is able to invalidate it correctly and so when it gets sampled back, it is found. If you switch to HW renderer after this, the HW renderer will read the texture out of GS memory as it would any other texture transferred in from the EE. This is why that workaround works.

This has additional complications due to the fact that the HW renderer will not invalidated rt back into GS memory. This is a problem that is outside the scope of this PR but has some side effects when switching from HW to SW when the level is loaded in HW or opening the config for example.

RedDevilus:
I see, it does seem to have improved the sky, just was surprised the black sky problem was also happening in SW mode (first in HW mode and then switching to SW) but was fine after 2 restarts of the race.

Sorry :o

CookiePLMonster:
Fixes pause menu in **The Getaway** not working properly in Hardware mode. Without this PR, pause shows the last frame of Software mode (or nothing), with this PR it properly shows a "screenshot" of the underlying game.

image

plugins/GSdx/Renderers/HW/GSTextureCache.cpp
```diff
@@ -2108,6 +2058,194 @@ void GSTextureCache::AttachPaletteToSource(Source* s, uint16 pal, bool need_gs_t
...
+				b2a_offset.w = b_el.rect.w;
+			}
+		}
+	}
+
+	assert(!so.is_valid || b2a_offset.z > b2a_offset.x);
+	assert(!so.is_valid || b2a_offset.z <= b_el.rect.z);
+	assert(!so.is_valid || b2a_offset.w > b_el.rect.y);
+	assert(!so.is_valid || b2a_offset.w <= b_el.rect.w);
+	assert(!so.is_valid || b2a_offset.z % dx == 0);
```
iMineLink, line 2188: Reminder to myself: this is not necessarily true, remove assertion here and below.

iMineLink:
I gave up on Driving Emotion S in this PR.
Open to discuss about the Jak 2/3/X only change in TC (crc based code inside TC already exists in the codebase); In any case, it's just a small commit.
I think this might be given a round of tests in Burnout, I don't think this can improved any further without severely interfering with framebuffer cache invalidation.

atomic83GitHub:
I don't agrea with the jak only fix, this should be a general fix as it improves burnout, gateway and probably many others.

iMineLink:
I better specified in the PR description what is the Jak only change: it is a small bonus part of the PR, that uses the same logic of the tex in rt, but applied to framebuffer invalidation.
The PR is mainly about tex in rt instead, which code is now always run instead.

atomic83GitHub:
Ok, thanks for the explaination.

MrCK1:
Tested the updated code in Burnout 3, works fine on my end with no performance issues that I could see.

Only issue I noticed is that closing the GS window and reopening will cause the sky to be black again, but I don't think there's much we can do about that unless I'm mistaken.

legend800:
Skybox is still not being loaded up correctly on HW on latest commit. It's blue but missing clouds and has bad banding. Sky turns black after toggling to Software and is broken until you start again from menus.

Burnout 3 HW at load:
2020-12-13 15_35_17-Slot_ 0 _ Speed_  70% (41 78) _ Limiter_ Normal _ OGL _ 3200x2240 _ None _ ${omo

Burnout 3 SW at load:
2020-12-13 15_37_03-Slot_ 0 _ Speed_  66% (39 54) _ Limiter_ Normal _ OGL _ 3200x2240 _ None _ ${omo

Burnout 3 HW at load:
2020-12-13 15_39_01-Slot_ 0 _ Speed_  76% (45 27) _ Limiter_ Normal _ OGL _ 3200x2240 _ None _ ${omo

Burnout Takedown SW at load:
2020-12-13 15_39_53-Slot_ 0 _ Speed_ 100% (59 94) _ Limiter_ Normal _ SW _ 640x448 _ None _ ${omodec

RedDevilus:
Doing any savestates/changing settings or anything basically will make the sky break. That's a different issue.

legend800:
What? Read my comment again. The sky does not load correctly on this PR. I'm not talking about save states or changing settings. The sky does not turn black unless the sky is not initially cached properly. If you start it on software, it's never black regardless of setting changes at that point. 

But regardless, the issue is this PR does not do anything so far except make the sky partially render on HW at load (better than black), but loading stage with software is still needed to get it looking proper.

tadanokojin:
>The sky does not turn black unless the sky is not initially cached properly. If you start it on software, it's never black regardless of setting changes at that point.

See: https://github.com/PCSX2/pcsx2/pull/3895#issuecomment-723494703

Specifically:
>This has additional complications due to the fact that the HW renderer will not invalidate rt back into GS memory. This is a problem that is outside the scope of this PR but has some side effects when switching from HW to SW when the level is loaded in HW or opening the config for example.

Yes I see there are still issues with the sky and the reflections. Thank for your test.

legend800:
Ah gotcha, you're just referring to the specific comment where I mentioned HW > software triggers black sky. Ok, got it, separate thing. So even if the sky did load correctly in HW (doesn't yet), you'll still get the black sky if you go to Software. Thankfully, it'll be moot once this PR gets the sky looking solid on initial load in HW. ;)

prafullpcsx2:
This PR also fixes most issues in Indiana Jones and the Emperor's Tomb HW mode.

plugins/GSdx/Renderers/HW/GSTextureCache.cpp
```diff
@@ -257,20 +258,18 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
...
 		// else.)
 
-		bool texture_inside_rt = ShallSearchTextureInsideRt();
-
-		for(auto t : m_dst[RenderTarget]) {
-			if(t->m_used && t->m_dirty.empty()) {
+		bool found_t = false;
+		for(auto t : m_dst[RenderTarget])
+		{
+			if(t->m_used && t->m_dirty.empty())
```
iMineLink, line 271: Memo:
`t->m_dirty.empty()` is too strict and also incorrect for tex in rt.
Indiana Jones requires to relax this to correctly display the map.
The validity rectangle of t used for the tex in rt lookup must then take into account the `t->m_dirty` list.
Initially only `C_32` pixel formats can be considered.

atomic83GitHub:
This PR also makes the shadows better in "beyond good and evil" but they have a another unrelated issue.

iMineLink:
Burnout reflection regression is instrinsic in the nature in the current TC: at the beginning of the draw a big clearing draw is done (thus creating a huge valid frambuffer).
Later on, several overlapping draws are done, but they are not reported to the original framebuffer, which keeps the original big validity rectangle.
When it comes to compute the spherical mapping for the reflections, it sometimes happens that a part the old huge frambuffer (its validity still being at least that of the initial clear) is picked up with a given offset instead of the correct texture to be applied (computed some draw calls before).
In short terms, having overlapping targets is not feasible anymore if tex in rt is going to be always active.

Burnout 3 (and hopefully also the others) black sky can be solved in a much better way with a combination of sw sprite rendering and point list palette rendering, for which I'll be opening PR soon.

What do you suggest to do here? We collected some games that react positively to the improved tex in rt. Given the upcoming merge of GSdx (we will have easy access to game database from GSdx), it can be a possibility to keep tex in rt an "optional feature", instead of rethinking framebuffer invalidation in TC.

tadanokojin:
Another related issue: #2409

iMineLink:
This is not a draft really, Burnout games will not be fixed by this alone anyway.
PR #4094 is much better for tackling the black sky issue (maybe also in the Black game).
Beyond Good and Evil here needs some improvements in how wrapping RT are handled in the readback case, which can be delivered later.

Also keeping tex in rt enabled via ini for non-Jak games reduces a lot the regressions probability.
There are still some GSdx wide changes in this PR, but much safer than tex in rt.

I'll switch this again to "Open".

CookiePLMonster:
It would seem that GS impacted this PR severely, since a "naive" rebase doesn't carry over the improvements (i.e. pause menu in The Getaway is still broken).

iMineLink:
I was initially supposing that you did not set the `UserHacks_TextureInsideRt = 1` variable in the .ini, but that's clearly not the case.
Thanks for the additional debugging and for tracking down the breaking commit.
The wrapping render targets handling is a bit sketchy in the HW renderer TC anyway; I suspect it might also be affected by #4348 because it touches the `GSOffset`.
What's happening here might be that a surface is detected to be (wrongly or not) a wrapping RT and it steals the cache hit from the correct surface, due to the unfortunate RT ordering in the MRU list and/or RT not being correctly invalidated.
Anyway it is demonstrated via Burnout sky issue that the current code cannot be generalized without adding complex heuristics (a big clearing draw create a huge RT, never invalidated, which steals further correct tex in RT hits).
So IMHO, to alleviate current TC issues specific to the tex in RT handling it might be good to add some "gamefixes" in the gamedb specific for such cases.

Drop in a GS dump please for The Getaway issue, I'll look into it in the next days.

CookiePLMonster:
> Drop in a GS dump please for The Getaway issue, I'll look into it in the next days.

There you go:
[gs_20210710115954.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/6795020/gs_20210710115954.gs.xz.zip)

iMineLink:
This PR is now rebased and the checkbox for the tex in rt logic has been added in the GS HW hacks settings GUI. I have yet to look into The Getaway issue.

iMineLink:
With the latest build, I'm able to match the SW renderer with the HW one on The Getaway replay with the "Texture Inside RT" hack enabled. On OGL it also needs blending level high.

prafullpcsx2:
@iMineLink Indiana Jones videos are again divided into two. You fixed this bug earlier also and probably already have the dump.

iMineLink:
Indiana Jones has been sorted out and the relevant commit edited, it should now be working fine.
It is worth mentioning that the "PAL/NTSC selection menu in Wild Arms 4 HW mode" is NOT fixed in debug build but it is fixed in devel/release.
Viceversa, the "Driving Emotion S" issue seems fixed (it is in my replays) in debug mode but NOT in devel/release.

pcsx2/GS/Renderers/HW/GSRendererHW.cpp
```diff
@@ -1459,6 +1439,29 @@ void GSRendererHW::Draw()
...
+		rt_tex = rt->m_texture;
+	}
+
+	TEX0.TBP0 = context->ZBUF.Block();
+	TEX0.TBW = context->FRAME.FBW;
+	TEX0.PSM = context->ZBUF.PSM;
+
+	GSTextureCache::Target* ds = NULL;
+	GSTexture* ds_tex = NULL;
+	if (!no_ds) {
```
lightningterror, line 1460: Formatting, above too.

pcsx2/GS/Renderers/HW/GSRendererHW.cpp
```diff
@@ -1459,6 +1439,29 @@ void GSRendererHW::Draw()
...
+	if (!no_rt) {
+		rt = m_tc->LookupTarget(TEX0, m_width, m_height, GSTextureCache::RenderTarget, true, fm);
+		rt_tex = rt->m_texture;
+	}
+
+	TEX0.TBP0 = context->ZBUF.Block();
+	TEX0.TBW = context->FRAME.FBW;
+	TEX0.PSM = context->ZBUF.PSM;
+
+	GSTextureCache::Target* ds = NULL;
```
lightningterror, line 1458: Maybe use nullptr instead?

lightningterror (commented):
Needs linux option too.

https://github.com/PCSX2/pcsx2/blob/master/pcsx2/GS/Window/GSLinuxDialog.cpp#L454

iMineLink:
The Linux option should be ok now, but I couldn't test it.

iMineLink:
done

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -271,25 +273,22 @@ GSTextureCache::Source* GSTextureCache::LookupSource(const GIFRegTEX0& TEX0, con
...
 
-		bool texture_inside_rt = ShallSearchTextureInsideRt();
+		const bool texture_inside_rt = ShallSearchTextureInsideRt();
 
-		for (auto t : m_dst[RenderTarget])
+		bool found_t = false;
+		for(auto t : m_dst[RenderTarget])
 		{
-			if (t->m_used && t->m_dirty.empty())
+			if(t->m_used)
```
lightningterror, line 291: Space after if, same for the loop above.

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -1048,9 +1003,11 @@ void GSTextureCache::InvalidateLocalMem(const GSOffset& off, const GSVector4i& r
 		GL_INS("ERROR: InvalidateLocalMem depth format isn't supported (%d,%d to %d,%d)", r.x, r.y, r.z, r.w);
 		if (m_can_convert_depth)
 		{
-			for (auto t : m_dst[DepthStencil])
+			auto& dss = m_dst[DepthStencil];
+			for (auto it = dss.rbegin(); it != dss.rend(); ++it)  // Iterate targets from LRU to MRU.
 			{
-				if (GSUtil::HasSharedBits(bp, psm, t->m_TEX0.TBP0, t->m_TEX0.PSM))
+				Target* t = *it;
+				if(GSUtil::HasSharedBits(bp, psm, t->m_TEX0.TBP0, t->m_TEX0.PSM))
```
lightningterror, line 1010: Same.

refractionpcsx2:
Ran through a bunch of my dumps (and a couple of games), only improvement I saw was the noted BGE shadow, but nothing seemed to break with the option enabled either.

pcsx2/GS/Window/GSSetting.cpp
```diff
@@ -196,6 +196,9 @@ const char* dialog_message(int ID, bool* updateText)
 		case IDC_PRELOAD_TEXTURES:
 			return cvtString("Uploads entire textures at once instead of small pieces, avoiding redundant uploads when possible.\n"
 				   "Improves performance in most games, but can make a small selection slower.");
+		case IDC_TEX_IN_RT:
+			return cvtString("Allows the texture cache to reuse as an input texture the inner portion of a previous framebuffer.\n"
+				"By default this is enabled only in Jak games.");
```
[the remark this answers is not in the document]
iMineLink: No, it's following the most common formatting there.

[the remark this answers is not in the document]
lightningterror: Hmm, it looks like the tooltip above and 2 more have those, which is odd, but anyway this one is fine then.

refractionpcsx2:
It would be super nice if we could add the CRC's like with Jak so this is automatically on for the games listed

RedDevilus:
Fixes SSX 3 ( when upscaled and tex in rt checkbox)
Emulog:
image

Default upscaled:
image

Upscaled with tex in rt PR checkbox:
image

rtj503:
So I tested this added option in Xenosaga 3 with the Zarathustra Dungeon. 
gs_20220128213456
It didn't actually have any effect. There are two areas where it's most prominent. One area is this one right before the crystallized section. The other is just beyond this part where you can step on switches on the floor to make transparent panels move. 
Zarathustra Panel Room
For comparison, here's how it's all supposed to look in Software Mode. 
gs_20220128213527

RedDevilus:
Ref made this for easily seeing if the function gets triggered if anyone is interested, enable HW hacks and tex in rt and check the program log/ console:

[pcsx2-avx2-texrtlog.zip](https://github.com/PCSX2/pcsx2/files/7966628/pcsx2-avx2-texrtlog.zip)

image

rtj503:
I downloaded that. That launcher just made the game really slow. I didn't notice any text in the Program Log that said Tex in RT had been activated. So maybe it's not activating.

iMineLink:
Regarding xenosaga 3, I have a gs dump from the "crystallized section" that works correctly now on master. Can you try forcing the tex in rt option active from the hw hacks section? Maybe the CRC of your game version is not registered for auto enabling the tex in rt logic.

rtj503:
Maybe that's the issue. I have a save from the Zarathustra Dungeon that I used on my AMD laptop and my Desktop that has an RTX 3060 graphics card. The graphical problem still occurs on both. I enabled the option in the HW hacks section, and I looked to see if it was enabled in the ini file. (I'm not a computer programmer, although I'd like to contribute with code, I don't have the knowledge, so don't ask anything too advanced, lol). 

Here's what the program log said, I assume this was the relevant section:

Patches: No CRC found, using 00000000 instead.
(SYSTEM.CNF) Detected PS2 Disc = cdrom0:\SLUS_214.17;1
(SYSTEM.CNF) Software version = 1.00
(SYSTEM.CNF) Disc region type = NTSC
ELF (cdrom0:\SLUS_214.17;1) Game CRC = 0x2088950A, EntryPoint = 0x00120008
(SYSTEM.CNF) Detected PS2 Disc = cdrom0:\SLUS_214.17;1
(SYSTEM.CNF) Software version = 1.00
(SYSTEM.CNF) Disc region type = NTSC

So first it mentions not having a CRC code to apply a patch, but then it says midway through that it has a CRC code.

refractionpcsx2:
hmm that CRC is in there... can you try it on #5433 ?

rtj503:
> hmm that CRC is in there... can you try it on #5433 ?
Is that a build, or would that require compilation? 
I ran it all on the 1.7.2282 AVX2 64bit build. I tried compiling PCSX2 from source in Visual Studio once...didn't work. Not certain what happened. Did accomplish it in Linux once though a while back.

refractionpcsx2:
if you go to the checks at the bottom of 5433, then click on the "details" next to windows, then go to "summary", there are builds in there.

rtj503:
I did that, I downloaded the AVX 64bit build. Same glitch still popping up. Perfectly willing to do more testing for you guys after some more work was done on it. 
For reference, I posted some testing I did for Xenosaga 3 on the forums with pretty thorough documentation. https://forums.pcsx2.net/Thread-Xenosaga-Episode-III-Also-Sprach-Zarathustra-SLUS-21389-U--28443
I'm user tgg22. I tested on both my AMD laptop with integrated graphics, and my HP Omen Desktop with an AMD CPU and Nvidia GPU. Testing is admittedly for older builds, but those specs are representative of the options I typically like to pick. In case that helps.

refractionpcsx2:
@rtj503 can you provide a GSDump of Xenosaga, please? might make testing easier.

rtj503:
What's a GSDump and how do I do that? Sorry for asking the beginner questions. I'm perfectly fine helping out, but I need guidance doing since I don't know anything about the software other than what's on PCSX2 website. I couldn't find that in the PCSX2 documentation. I do have a memory card file. I've found Memory Cards work well finally in the newer builds.

refractionpcsx2:
Here's an explanation how to make one https://forums.pcsx2.net/Thread-How-to-create-a-proper-GS-dump

rtj503:
> Here's an explanation how to make one https://forums.pcsx2.net/Thread-How-to-create-a-proper-GS-dump

Okay, I followed the directions, and posted the GS Dump in the forums. Link here: https://forums.pcsx2.net/Thread-GSdx?pid=624888#pid624888
Glad to help!

PR #3897, opened 2020-11-03
DEV9: switch net.cpp to use std::threads

TheLastRar:
Also merge Windows & Linux net.ccp into one file, based on the windows code

Priority setting is maintained using platform specific code

This also unifies the yielding/sleeping behaviour in the event that nothing can be received.
Previously, windows would sleep 10ms, while Linux would not sleep or yield (for reference, CLR_DEV9 would sleep 1ms).
I now just call yield() instead.


pcsx2/DEV9/net.cpp
```diff
@@ -49,19 +53,22 @@ void InitNet(NetAdapter* ad)
 	nif = ad;
 	RxRunning = true;
 
+	rx_thread = std::thread(NetRxThread);
+
+#ifdef _WIN32
+	SetThreadPriority(rx_thread.native_handle(), THREAD_PRIORITY_HIGHEST);
+#elif defined(__linux__)
```
GovanifY, line 60: Not necessarily linux? I assume macos handles the POSIX API just fine

GovanifY (commented):
Overall lgtm otherwise, thanks for the work!

pcsx2/DEV9/net.cpp
RedDevilus: Isn't #elif defined(__unix__) more appropiate?

TheLastRar: Would `#elif defined(__unix__) || defined(__APPLE__)` cover everything?

RedDevilus: defined unix is used for linux and apple in the code recently, only config.inl uses that specific line so i guess it would work too.

GovanifY: yeah, this should cover everything

TheLastRar: > defined unix is used for linux and apple in the code recently, only config.inl uses that specific line so i guess it would work too.

A Google srarch seemed to impy that __unix__ wasn't defined when compiling for mac. Hence the extra check. Are you saying it actually is?

TheLastRar:
Now also sets priority on mac via pthreads

TheLastRar:
```E: Could not configure 'libc6:i386'. 
E: Could not perform immediate configuration on 'libgcc-s1:i386'. Please see man 5 apt.conf under APT::Immediate-Configure for details. (2)
Error: Process completed with exit code 100.```
Pretty sure this isn't my fault.

lightningterror:
Don't worry about it.

pcsx2/DEV9/net.cpp
tellowkrinkle: `__UNIX__` is defined on macOS by wx, I don't think we use it anywhere though
`__unix__` is not defined on macOS
`__POSIX__` is defined on macOS by `common/include/Pcsx2Defs.h`

If the functions you're using are defined by posix (I think they are in this case), you should probably use `__POSIX__`
If not, but they happen to exist on both Linux and macOS, use `defined(__linux__) || defined(__APPLE__)`

TheLastRar: Ah, so ``__POSIX__`` is what I should use in this case then, given it's just posix threads?
``common/include/Pcsx2Defs.h`` is included in `PrecompiledHeader.h` so I shouldn't need to include anything extra to get it

refractionpcsx2:
you can probably rebase if the builds not working bothers you

TheLastRar:
Done

GovanifY (approved):

RedDevilus (approved):

PR #3899, opened 2020-11-04
Check register 14 (ps1 status register) for cd speed and CDDA 

kenshen112:
Fix to CD speed courtesy of Ref. Restored amount of old pcsx code for CDDA

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -1186,8 +1186,6 @@ static void cdvdWrite04(u8 rt)
 { // NCOMMAND
 	CDVD_LOG("cdvdWrite04: NCMD %s (%x) (ParamP = %x)", nCmdName[rt], rt, cdvd.ParamP);
 
-	std::cout << "COMMAND: " << nCmdName[rt] << std::endl;
-
 	cdvd.nCommand = rt;
 	// Why fiddle with Status and PwOff here at all? (rama)
 	cdvd.Status = cdvd.Spinning ? CDVD_STATUS_SPIN : CDVD_STATUS_STOP; // checkme
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -37,7 +37,7 @@
 // this string will be empty.
 wxString DiscSerial;
 
-static cdvdStruct cdvd;
+cdvdStruct cdvd;
 
 s64 PSXCLK = 36864000;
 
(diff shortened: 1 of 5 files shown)


pcsx2/CDVD/CDVD.cpp
```diff
@@ -1186,6 +1186,8 @@ static void cdvdWrite04(u8 rt)
 { // NCOMMAND
 	CDVD_LOG("cdvdWrite04: NCMD %s (%x) (ParamP = %x)", nCmdName[rt], rt, cdvd.ParamP);
 
+	std::cout << "COMMAND: " << nCmdName[rt] << std::endl;
```
GovanifY, line 1189: You should use DevCon.WriteLn

pcsx2/CDVD/CDVD.h
```diff
@@ -137,6 +137,7 @@ struct cdvdStruct
 	bool Spinning;    // indicates if the Cdvd is spinning or needs a spinup delay
 };
 
+extern cdvdStruct cdvd;
```
GovanifY, line 140: Why an extern? Can't you define the struct in the header directly?

pcsx2/CDVD/CdRom.cpp
```diff
@@ -844,6 +841,26 @@ void cdrWrite1(u8 rt)
 			cdr.Mode = cdr.Param[0];
 			cdr.Ctrl |= 0x80;
 			cdr.Stat = NoIntr;
+			if (cdr.Mode && MODE_CDDA)
+			{
+				StopCdda();
+				cdvd.Type = CDVD_TYPE_CDDA;
+				if (cdr.Reading == 1)
```
GovanifY, line 848: why an == 1? isn't a standard if sufficient in this case or do we want the 1? If we do, why not an enum?

pcsx2/CDVD/CdRom.cpp
```diff
@@ -844,6 +841,26 @@ void cdrWrite1(u8 rt)
...
+				{
+					StopReading();
+				}
+				if (cdr.SetlocPending)
+				{
+					memcpy(cdr.SetSectorSeek, cdr.SetSector, 4);
```
GovanifY, line 854: Why do you need a memcpy? Can't you simply make a static allocation or a bitmask if you only want 4 bytes?

GovanifY (changes requested):

pcsx2/CDVD/CdRom.cpp
```diff
@@ -226,7 +225,7 @@ void cdrInterrupt()
 			cdr.Result[0] = cdr.StatP;
 			cdr.Stat = Acknowledge;
 			break;
-
+
```
refractionpcsx2, line 228: Whitespace

pcsx2/CDVD/CdRom.cpp
```diff
@@ -748,11 +745,22 @@ void cdrWrite1(u8 rt)
 
 			cdr.Ctrl |= 0x80;
 			cdr.Stat = NoIntr;
+			cdr.SetlocPending = 1;
 			AddIrqQueue(cdr.Cmd, 0x800); // the seek delay occurs on the next read / seek command (CdlReadS, CdlSeekL, etc)
 		}
 		break;
-
+		do_CdlPlay:
```
refractionpcsx2, line 752: What is this? does pcsxr have the goto in the middle of a switch? how does that even work? O_o

kenshen112: https://github.com/iCatButler/pcsxr/blob/62467b86871aee3d70c7445f3cb79f0858ec566e/libpcsxcore/cdrom.c#L653

PR #3901, opened 2020-11-05
Debugging 32-bit Ubuntu Package installation Failure

xTVaser:
Tracking - https://bugs.launchpad.net/ubuntu/focal/+source/apt/+bug/1871268

I'm not familiar with Canoical's package release process, but i assume they would test something in 20.10 before rolling it to 20.04.  It seems like this issue has been resolved and is currently baking in 20.10.

### DIFF
--- a/.github/workflows/scripts/linux/install-packages.sh
+++ b/.github/workflows/scripts/linux/install-packages.sh
@@ -119,5 +119,9 @@ PCSX2_PACKAGES_STR=""
 for i in "${PCSX2_PACKAGES[@]}"; do
   PCSX2_PACKAGES_STR="${PCSX2_PACKAGES_STR} ${i}${ARCH}"
 done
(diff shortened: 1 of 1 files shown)


turtleli:
Workaround mentioned in https://bugs.launchpad.net/ubuntu/focal/+source/apt/+bug/1871268/comments/26 might work.
https://github.com/turtleli/pcsx2/commit/b0df78e3c8ba438b0d5d2f2ea2725349faab3e11
https://github.com/turtleli/pcsx2/pull/7/checks

tylerwilding-wk:
@turtleli Ah thank you, I missed that comment when scanning the ticket!.  I did try adding that package but I did it as part of the larger installation instead of doing it on it's own.  Good catch

xTVaser:
@turtleli Thanks! I had missed that completely when scanning the ticket.  I had tried installing that package, but not early/on it's own, seems like it works from your run.

PR #3902, opened 2020-11-05
Add AUR Job to Github Actions

kenshen112:
This PR allows Actions to compile and push to the AUR package pcsx2-git everytime a push to master is made


lightningterror:
@xTVaser this one for you.

refractionpcsx2:
We need to decide what to do with this as we only have a limited number of actions, we already have 4 Linux ones and we need to leave room for macos at some point

turtleli:
If I understand this correctly then merging this PR will make us somewhat responsible for the maintenance of the AUR pcsx2-git package. Any future changes that the AUR pcsx2-git PKGBUILD will need will have to go through us instead of just being fixed directly. Doesn't that just make maintenance of the package more complicated?

Also, I see this note in https://wiki.archlinux.org/index.php/Arch_User_Repository
>  AUR maintainers should not commit mere pkgver bumps.

Doesn't this PR do the exact opposite?

I don't see this PR as a good idea.

turtleli:
Closing.

As stated previously, it'd make us responsible for maintaining the package. It's not our responsibility to maintain the AUR pcsx2-git package, it's yours since you are the maintainer.

PR #3904, opened 2020-11-06
SPU2: default to XAudio when failing to read conf

GovanifY:
also switched fprintf to Console.Warning

### DIFF
--- a/pcsx2/SPU2/Windows/Config.cpp
+++ b/pcsx2/SPU2/Windows/Config.cpp
@@ -154,7 +154,9 @@ void ReadSettings()
 	if (mods[OutputModule] == nullptr)
 	{
 		// Unsupported or legacy module.
-		fwprintf(stderr, L"* SPU2: Unknown output module '%s' specified in configuration file.\n", omodid);
+		Console.Warning("* SPU2: Unknown output module '%s' specified in configuration file.", omodid);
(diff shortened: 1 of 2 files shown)


lightningterror:
Forgot linux.

GovanifY:
@lightningterror If you look at the diff this bit of code was originally changed because I purged DSound from the Windows implem. I doubt I removed DSound from Linux too, so behaviour should stay consistent with past revisions.

lightningterror:
I was talking about adding a fallback to linux too, but it already has one so scratch what I said.

pcsx2/SPU2/Windows/Config.cpp
```diff
@@ -154,7 +154,9 @@ void ReadSettings()
 	if (mods[OutputModule] == nullptr)
 	{
 		// Unsupported or legacy module.
-		fwprintf(stderr, L"* SPU2: Unknown output module '%s' specified in configuration file.\n", omodid);
+		Console.Warning("* SPU2: Unknown output module '%s' specified in configuration file.\n", omodid);
+		Console.Warning("* SPU2: Defaulting to XAudio (%S).\n", XAudio2Out->GetIdent());
```
lightningterror, line 158: Is new line at the end needed? Console already does that.

GovanifY: good catch!

lightningterror (approved):

PR #3905, opened 2020-11-06
gsdx: Remove TV Shader hotkey toggle (F7).

lightningterror:
Plenty of users misclick this setting and don't know how to switch back
since it's a rarely used option.
Let's get rid of the hotkey toggle for a quality of life change.


RedDevilus (approved):
I agree, they can still go into gsdx shader configuration for the TV shaders. Doesn't need to be changed often for games.

tadanokojin:
Can you run clang on the linux dialog file, then it's good by me.

tadanokojin (approved):

karasuhebi:
Was an option added to where we can rebind this to hotkey via PCSX2_keys.ini?

PR #3907, opened 2020-11-06
dev9/spu2: Cleanup guis a bit.

lightningterror:
gui:
Rename window title: HDD Configure -> Network and HDD Settings
Remove leftover about dialog code.
Move around button positions.
Some other changes.

pcap:
Fix not defined macro warning.

spu2: Rename SPU2 Config -> Audio Settings.
We can have a better user friendly name as the plugin is merged in to
core.


pcsx2/DEV9/Linux/dev9.ui
```diff
@@ -5,7 +5,7 @@
   <object class="GtkDialog" id="IDD_CONFDLG">
     <property name="can_focus">False</property>
     <property name="border_width">5</property>
-    <property name="title" translatable="yes">HDD Configure</property>
+    <property name="title" translatable="yes">DEV9 Settings</property>
```
Florin9doi, line 8: This may make sense from a developer point of view but it's user unfriendly. Nobody cares about the internal chip's name

lightningterror: I went with that for consistency but yeah it's a bit unfriendly for normal users.
We can adjust the others with better naming as well
SPU2 Settings -> Audio Settings
GSdx -> Video Settings

lightningterror: I went ahead with the suggestion and also renamed spu2 since it's been merged in to the core. Can be changed after merge.

PR #3908, opened 2020-11-07
Replace Star Wars - Force Unleashed CRC hack with GameDB patch. 

### LINKED ISSUE #3175: Star Wars: The Force Unleashed (SLES 54658) Missing or glitched HUD graphics
***PCSX2 version:***
Current Git build as of this date.

***PCSX2 options:***
No options appear to have any impact on this bug

***Plugins used:***
Bug manifests using GSdx in hardware and software modes. 

Any status bars in the HUD do not function correctly. More specifically, they are rendered, but only visible in certain circumstances. Based on visual inspection only, the bars seem to be made up of two elements: a grid which overlays the bar to give an indication of how much health etc you have compared to the maximum, and the coloured bars themselves which change length as you lose, for example, health. The grid part displays properly, but the coloured bar does not most of the time. I say "most of the time" because it looks like it is being rendered, but geometry (backgrounds, characters, etc) is being drawn behind the grid overlay but in front of the coloured bar, rendering it invisible most of the time. 

Attached is a screenshot and GSDump showing a partially occluded bar from the opening level. Any geometry except for the far skybox occludes the bars.

gsdx_20191219150946

[gsdx_20191219150946.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/3984227/gsdx_20191219150946.gs.xz.zip)

kozarovv:
EE patches work also in software mode. 

Note: CRC 0x137C792E seems to be some mistake. That game have only one version for US region.

Close #3175

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -17305,12 +17305,8 @@ Name   = Star Wars - The Force Unleashed
 Region = PAL-M5
 [patches = 87109051]
 	author=kozarovv
-	// Fixes various graphics glitches.
-	// It's a problem which apply to all games made by Khrome studio.
-	// The EE seems to send invalid data to the GS and expect
-	// some kind of unknown behaviour to operate.
-	// This can be fixed on the GS side but the software render is impossible to fix.
-	// So this patch correct the issue on the EE side.
+	// Fix for graphical issues, QTE events, subtitles, health bar.
+	// Patch replace values passed to FPU to workaround x86 rounding issues.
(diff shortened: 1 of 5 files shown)


atomic83GitHub:
The comments on the patch should mention what the issue is, it's a rounding issue, the game uses floats values higher than X86 the X86 standard which can't be solve by normal rounding operations.

kozarovv:
Values doesn't seems to go outside of x86 range, but for sure this is rounding issue. I added proper description to patches, that should be enough info.

PR #3913, opened 2020-11-10
CDVD: disable cdvd reset on framelimiter

GovanifY:
Tentative fix on the CDVD reset happening on framelimiter


GovanifY:
Probably a good idea to close this before anyone thinks this is a good idea

PR #3923, opened 2020-11-11
[USB] Change wheel type to device subtype

Florin9doi:
It is used only for Steering Wheels now, but may allows later to group more devices like Keyboard/Beatmania, Microphone/Singstar, or RealPlay.


refractionpcsx2:
Looks like Linux isn't happy here

pcsx2/USB/usb-pad/raw/usb-pad-raw.cpp
```diff
@@ -253,7 +253,7 @@ namespace usb_pad
 					uint16_t btn = mapping->btnMap[usage[i] - pButtonCaps->Range.UsageMin];
 					for (int j = 0; j < 2; j++)
 					{
-						PS2WheelTypes wt = (PS2WheelTypes)conf.WheelType[1 - j];
+						PS2WheelTypes wt = (PS2WheelTypes) 0; // conf.WheelType[1 - j];
```
jackun, line 256: Pass type from RawInputPad::ParseRawInput?

jackun, line 256: Pass type from RawInputPad::ParseRawInput?

jackun:
Thought of "pulling" the device subtypes "upward" if it got any like Rock band drum kit and Buzz. Any point?

Florin9doi:
You mean adding BUZZ wired and wireless emulation? Unless they make any difference for games, I think it would add unnecessary worries for users.

jackun:
I mean, only have the single device combo box. Instead of "Wheel device", you have directly "Driving Force, DF Pro, ...blah..., Singstar, ..blah.., Keyboardmania" etc.

Florin9doi:
Having a single combo box would result in a really long list. There are still device to be added: lightguns, realplay, gametrak, dj hero, psp.

RedDevilus:
What about just having 2 selections? 1 side just gives context menu for Device types and then on right side the list of devices within your first selection of the device type.

pcsx2/USB/Win32/Config_usb.cpp
```diff
@@ -64,6 +64,18 @@ void SelChangedAPI(HWND hW, int port)
 	changedAPIs[std::make_pair(port, devName)] = *it;
 }
 
+void SelChangedSubtype(HWND hW, int port)
+{
+	int sel = SendDlgItemMessage(hW, port ? IDC_COMBO_WHEEL_TYPE1_USB : IDC_COMBO_WHEEL_TYPE2_USB, CB_GETCURSEL, 0, 0);
```
lightningterror, line 69: You can use const here, didn't check others but would be nice to change them too if there are any that can be changed.

jackun:
Seems good but return std::vector for static strings seems weird and I can't think a better way than some gls::span or c++20 thingy.

pcsx2/USB/usb-pad/usb-pad.h
```diff
@@ -59,6 +59,15 @@ namespace usb_pad
 		static const TCHAR* LongAPIName(const std::string& name);
 		static int Configure(int port, const std::string& api, void* data);
 		static int Freeze(int mode, USBDevice* dev, void* data);
+		static std::vector<std::string> SubTypes()
+		{
+			std::vector<std::string> ret;
```
jackun, line 64: This probably could just be a initializer list: `{ "blahh", }` etc

Florin9doi: done

jackun (approved):

Florin9doi:
anything else needed ?

PR #3925, opened 2020-11-12
USB: Resolve an EyeToy crash when no webcam is connected

Florin9doi:


jackun (approved):

GovanifY (approved):

CookiePLMonster:
That's a terrifying exception from the rule regarding `HRESULT` values MSFT has laid out themselves.

PR #3928, opened 2020-11-14
GameDB: Replace GameDB with a YAML equivalent

xTVaser:
This PR serves two purposes:
- Improve the format of the current GameDB slightly and simplify the code around the handling the file itself.  Before it was an extended .ini file format that relied on WX libraries.
  - The new GameDB will do some brief validation upon loading the file.  I found several instances of invalid GameFixes being used, going forward, these problems should be easy to identify and solve.
- Serve as a proof of concept for the YAML library / format that will be coming soon with the larger configuration overhaul.

There are some things to point out:
- If the GameDB changes before this is merged, I have a throw-away script to convert it that I can commit if desired, but I'll want to clean it up first (I said throw-away for a reason...it's not pretty).
- I made a unrelated change to solve what seemed to be a long-standing collision on MSVC around a `_Target_` definition in R5900.cpp Please review.
- yaml-cpp is vendored at their current master.  It's been over a year since a release has been cut and when you look back through the last year of commits, some fixes seem valuable so I avoided pinning to this 1 year old release.  This means that the `yaml-cpp` from a linux package repository _could_ be used, but wouldn't have any guarantee.


pcsx2/GameDatabase.cpp
```diff
@@ -14,63 +14,193 @@
...
+#include "fmt/core.h"
+#include "yaml-cpp/yaml.h"
+
+std::string GameDatabaseSchema::GameEntry::memcardFiltersAsString()
 {
+	if (memcardFilters.empty())
+		return "";
+
+	std::string filters;
+	for (int i = 0; i < memcardFilters.size(); i++)
```
arcum42, line 29: Size is unsigned, so to prevent a warning, I'd do:
for (u32 i = 0; i < memcardFilters.size(); i++)

pcsx2/GameDatabase.cpp
```diff
@@ -14,63 +14,193 @@
...
+		gameEntry.eeClampMode = static_cast<GameDatabaseSchema::ClampMode>(safeGetInt(node, "eeClampMode", enum_cast(gameEntry.eeClampMode)));
+		gameEntry.vuClampMode = static_cast<GameDatabaseSchema::ClampMode>(safeGetInt(node, "vuClampMode", enum_cast(gameEntry.vuClampMode)));
+
+		// Validate game fixes, invalid ones will be dropped!
+		for (std::string fix : safeGetStringList(node, "gameFixes"))
+		{
+			bool fixValidated = false;
+			for (GamefixId id = GamefixId_FIRST; id < pxEnumEnd; ++id)
+			{
+				std::string validFix = wxString(EnumToString(id)).ToStdString() + L"Hack";
```
arcum42, line 96: Doesn't compile. Let's do this:
std::string validFix = wxString(EnumToString(id)).Append(L"Hack").ToStdString();

pcsx2/gui/AppCoreThread.cpp
```diff
@@ -243,78 +243,78 @@ void AppCoreThread::OnPauseDebug()
...
 		dest.Speedhacks.vuFlagHack = vuFlagHack;
 		gf++;
 	}
 
+	// TODO - config - this could be simplified with maps instead of bitfields and enums
 	for (GamefixId id = GamefixId_FIRST; id < pxEnumEnd; ++id)
 	{
-		wxString key(EnumToString(id));
-		key += L"Hack";
+		std::string key = wxString(EnumToString(id)) + L"Hack";
```
arcum42, line 306: Same compilation issue, so:
std::string key = wxString(EnumToString(id)).Append(L"Hack").ToStdString();

pcsx2/Patch.cpp
```diff
@@ -120,37 +121,31 @@ static void inifile_command(const wxString& cmd)
...
 		}
-	}
 
-	if (patchFound) TrimPatches(patch);
+		if (patch.patchLines.size() > 0)
+		{
+			for (auto line : patch.patchLines)
+            {
+                inifile_command(line);
+            }
```
lightningterror, line 145: Keep formatting consistent.

pcsx2/gui/AppCoreThread.cpp
```diff
@@ -303,8 +303,8 @@ static int loadGameSettings(Pcsx2Config& dest, const GameDatabaseSchema::GameEnt
 	// TODO - config - this could be simplified with maps instead of bitfields and enums
 	for (GamefixId id = GamefixId_FIRST; id < pxEnumEnd; ++id)
 	{
-		std::string key = wxString(EnumToString(id)) + L"Hack";
-
+        std::string key = wxString(EnumToString(id)).Append(L"Hack").ToStdString();
+
```
lightningterror, line 307: Whitespace, above formatting too.
Maybe run clang?

da2ce7:
Consider: https://github.com/crdoconnor/strictyaml  ?

The default YAML has a few sharp corners.

xTVaser:
strictyaml appears to be a python library?  Are there C++ bindings or something available?  I don't really see a need to switch to that at a glance, as I doubt we will hit any of those sharp corners for our very simplistic usage of YAML, and we already have committed to using yaml-cpp in an upcoming much larger PR.

If anything I was considering switching to https://github.com/biojppm/rapidyaml as it very performant, already setup for C++ with a CMakeLists, and doesn't have memory leak issues that we had to workaround for yaml-cpp, but due to it being relatively new I decided against it.  strictyaml clearly states as well that performance isn't a priority, which is a concern given how massive this file is.

bin/GameIndex.md
```diff
@@ -0,0 +1,181 @@
...
+- `DMABusyHack`
+  - Mana Khemia, Metal Saga. Denies writes to the DMAC when it's busy.
+- `VIFFIFOHack`
+  - Transformers Armada, Test Drive Unlimited. Fixes slow booting issue.
+- `VIF1StallHack`
+  - SOCOM II HUD and Spy Hunter loading hang.
+- `GIFFIFOHack`
+  - Enables the GIF FIFO. Needed for Wallace & Grommit, Hot Wheels, DJ Hero.
+- `FMVinSoftwareHack`
+  - Silent Hill 2-3. Fixes FMVs that are obscured when using hardware rendering by switching to software rendering whil
```
RedDevilus, line 149:
```suggestion
  - Silent Hill 2-3. Fixes FMVs that are obscured when using hardware rendering by switching to software rendering.
```

bin/GameIndex.md
```diff
@@ -0,0 +1,181 @@
...
+- `XGKickHack`
+  - Erementar Gerad, adds more delay to VU XGkick instructions. Corrects the color of some graphics, but breaks Tri-ace
+- `IPUWaitHack`
+  - FFX FMV, makes GIF flush before doing IPU work. Fixes bad graphics overlay.
+- `EETimingHack`
+  - General purpose timing hack.
+- `SkipMPEGHack`
+  - Finds sceMpegIsEnd pattern in games and then recompiles code to say the videos are finished.
+- `OPHFlagHack`
+  - Bleach Bankais and others
```
RedDevilus, line 139: punctation

pcsx2/gui/AppGameDatabase.cpp
```diff
@@ -14,128 +14,19 @@
  */
 
 #include "PrecompiledHeader.h"
+
```
RedDevilus, line 17: Unintended whitespace?

pcsx2/GameDatabase.cpp
```diff
@@ -14,63 +14,193 @@
...
-	for (auto it = kList.begin(); it != kList.end(); ++it) {
-		if (it->CompareKey(key)) {
-			if( value.IsEmpty() )
-				kList.erase(it);
-			else
-				it->value = value;
-			return;
+GameDatabaseSchema::GameEntry YamlGameDatabaseImpl::findGame(const std::string serial)
+{
+	if (gameDb.count(serial) == 1)
```
RedDevilus, line 164:
```suggestion
	if (GameDB.count(serial) == 1)
```

RedDevilus: I see gameDB, GameDB and gameDb scattered around the files. Not sure if that's intentional.

RedDevilus (changes requested):

pcsx2/gui/AppGameDatabase.cpp
xTVaser: I typically group the PCH separate, but I don't really have a strong opinion, looks like it's a 50/50 chance of that style being used just looking at a few .cpp files 🤷

pcsx2/GameDatabase.cpp
xTVaser: In this case `gameDb` is needed, as that's the variable's name.  Seems like the instances of `GameDB` are for logs / comments.

RedDevilus:
Looks good. Just need people to test with essential patches and are easily detectable if they work or not:
Someone with Xenosaga 1 trying to trigger the save to memcard bug or black would be a good starting point.

bositman:
Is it supposed to apply when auto game fixes are enabled? It didn't work here on my test with Xenosaga ep1 save point fix (worked if loaded from a pnach cheat)

xTVaser:
Yeah likely a small issue around the patch handling then, will investigate.

RedDevilus:
I have noticed one issue;
Master:
Enabled automatic gamefixes, enable manual gamefixes (force everything to 0)

Gamefixes do not get applied which is correct.

PR:
Enabled automatic gamefixes, enable manual gamefixes (ignores manual gamefixes)

Gamefixes do get applied which is incorrect.

Tested on demo and beyond good and evil with vu0kickstart (basically any game with SPS to see this easily)


Settings used for both:
image
image

refractionpcsx2:
You sure that's right red? As far as I knew you could enable manual game fixes if auto was on, but you couldn't override ones set in the auto fixes.

RedDevilus:
I tested it on several games, manual gamefixes always override game settings for me if i enabled the checkbox.

RedDevilus:
Nvm, seems to be an issue with my settings.

xTVaser:
Yeah I'd be surprised if this PR changed anything around that, as it really only touches the gamefixes / patches loaded from the GameDB.  The ones from the UI go through a different code-path I believe

pcsx2/GameDatabase.cpp
```diff
@@ -14,63 +14,187 @@
...
+}
+
+bool YamlGameDatabaseImpl::initDatabase(const std::string filePath)
+{
+	try
+	{
+		// yaml-cpp has memory leak issues if you persist and modify a YAML::Node
+		// convert to a map and throw it away instead!
+		YAML::Node data = YAML::LoadFile(filePath);
+		for (YAML::const_iterator entry = data.begin(); entry != data.end(); entry++)
```
CookiePLMonster, line 178: A foreach loop should make this clearer.

pcsx2/GameDatabase.cpp
```diff
@@ -14,63 +14,187 @@
...
+	return gameDb.size();
+}
+
+bool YamlGameDatabaseImpl::initDatabase(const std::string filePath)
+{
+	try
+	{
+		// yaml-cpp has memory leak issues if you persist and modify a YAML::Node
+		// convert to a map and throw it away instead!
+		YAML::Node data = YAML::LoadFile(filePath);
```
CookiePLMonster, line 177: Is a full absolute path passed there? If so, this will break if the path to PCSX2 contains Unicode characters (on Windows).

pcsx2/gui/AppGameDatabase.cpp
```diff
@@ -148,42 +39,38 @@ AppGameDatabase& AppGameDatabase::LoadFromFile(const wxString& _file, const wxSt
...
 	{
 		Console.Error(L"(GameDB) Database Not Found! [%s]", WX_STR(file));
 		return *this;
 	}
 
-	wxFFileInputStream reader( file );
+	u64 qpc_Start = GetCPUTicks();
 
-	if (!reader.IsOk())
+	if (!this->initDatabase(std::string(file)))
```
CookiePLMonster, line 54: This also may be Unicode unfriendly on Windows if a full absolute path is passed. If it's only a relative path within PCSX2, disregard.

pcsx2/GameDatabase.cpp
xTVaser: If you could provide a snippet that uses a foreach I'd gladly switch to it, cause I tried to at first again.  Using a foreach results in empty `YAML::Node`s.  And all documentation points to this style https://github.com/jbeder/yaml-cpp/wiki/Tutorial#basic-parsing-and-node-editing

xTVaser:
I tested `Dead or Alive 2` which requires a patch else the game straight up doesn't boot.  I verified this by commenting out the patch in the YAML (it crashed) and then by running it with the patch enabled.

Title bar in the console was updated accordingly, and the game worked:
image

So if there is still a problem with Xenosaga, I suspect a serial/CRC mismatch of some kind.

pcsx2/GameDatabase.cpp
CookiePLMonster: A foreach loop worked fine for me here: https://github.com/CookiePLMonster/pcsx2/commit/f728effbc8a9ad566252377833db375c30750105#diff-607dc9234d98755565862a0bbf907e55d71605d0668256b997f6555d5b82aa91R1544
The caveat here is that you receive a node which is a key-value pair of nodes, so need to use `node.first` or `node.second`. Took me a while to realize that.

pcsx2/GameDatabase.cpp
```diff
@@ -14,63 +14,193 @@
...
+	if (!n[key])
+		return def;
+	return n[key].as<std::string>();
+}
 
-	GameDataHash::const_iterator iter( gHash.find(id) );
-	if( iter == gHash.end() ) {
-		dest.clear();
-		return false;
+int YamlGameDatabaseImpl::safeGetInt(const YAML::Node& n, std::string key, int def)
```
CookiePLMonster, line 47: `.as` has an overload to do just that. Just do `n[key].as<int>(def);`

pcsx2/GameDatabase.cpp
```diff
@@ -14,63 +14,193 @@
...
+#include "yaml-cpp/yaml.h"
+#include <fstream>
+
+std::string GameDatabaseSchema::GameEntry::memcardFiltersAsString()
 {
+	if (memcardFilters.empty())
+		return "";
+
+	std::string filters;
+	for (u32 i = 0; i < memcardFilters.size(); i++)
```
CookiePLMonster, line 30: This is trivially changeable to a foreach loop.

pcsx2/GameDatabase.cpp
```diff
@@ -14,63 +14,193 @@
...
+		if (i != memcardFilters.size() - 1)
+			filters.append(",");
+	}
+	return filters;
 }
 
-// Sets the current game to the one matching the serial id given
-// Returns true if game found, false if not found...
-bool BaseGameDatabaseImpl::findGame(Game_Data& dest, const wxString& id) {
+std::string YamlGameDatabaseImpl::safeGetString(const YAML::Node& n, std::string key, std::string def)
```
CookiePLMonster, line 40: `.as` has an overload to do just that. Just do `n[key].as<std::string>(def);`

pcsx2/GameDatabase.cpp
CookiePLMonster, line 47: Ditto

pcsx2/GameDatabase.cpp
```diff
@@ -14,63 +14,193 @@
...
-		dest.clear();
-		return false;
+int YamlGameDatabaseImpl::safeGetInt(const YAML::Node& n, std::string key, int def)
+{
+	if (!n[key])
+		return def;
+	return n[key].as<int>();
+}
+
+std::vector<std::string> YamlGameDatabaseImpl::safeGetMultilineString(const YAML::Node& n, std::string key, std::vector<std::string> def)
```
CookiePLMonster, line 54: Ditto.

pcsx2/GameDatabase.cpp
```diff
@@ -14,63 +14,193 @@
...
+		lines.push_back(line);
 	}
-	dest = iter->second;
-	return true;
+
+	return lines;
 }
 
-Game_Data* BaseGameDatabaseImpl::createNewGame( const wxString& id )
+std::vector<std::string> YamlGameDatabaseImpl::safeGetStringList(const YAML::Node& n, std::string key, std::vector<std::string> def)
```
CookiePLMonster, line 71: Ditto.

pcsx2/GameDatabase.cpp
```diff
@@ -14,63 +14,193 @@
...
+		gameEntry.name = safeGetString(node, "name");
+		gameEntry.region = safeGetString(node, "region");
+		gameEntry.compat = static_cast<GameDatabaseSchema::Compatibility>(safeGetInt(node, "compat", enum_cast(gameEntry.compat)));
+		gameEntry.eeRoundMode = static_cast<GameDatabaseSchema::RoundMode>(safeGetInt(node, "eeRoundMode", enum_cast(gameEntry.eeRoundMode)));
+		gameEntry.vuRoundMode = static_cast<GameDatabaseSchema::RoundMode>(safeGetInt(node, "vuRoundMode", enum_cast(gameEntry.vuRoundMode)));
+		gameEntry.eeClampMode = static_cast<GameDatabaseSchema::ClampMode>(safeGetInt(node, "eeClampMode", enum_cast(gameEntry.eeClampMode)));
+		gameEntry.vuClampMode = static_cast<GameDatabaseSchema::ClampMode>(safeGetInt(node, "vuClampMode", enum_cast(gameEntry.vuClampMode)));
+
+		// Validate game fixes, invalid ones will be dropped!
+		for (std::string fix : safeGetStringList(node, "gameFixes"))
```
CookiePLMonster, line 92: Does this need to make a copy of the string?

CookiePLMonster (commented):
👍 on the changes related to Unicode paths, a few other remarks otherwise.

bositman:
Huh I just tested DOA2 as well and it crashed / did not detect the patch either...tried deleting the pcsx2 configuration ini for good measure but nothing...could you send me some kinda debug build so we can trace this?

pcsx2/GameDatabase.cpp
xTVaser: I could, but I need to track the index so I can skip the delimiter on the final filter anyway.  So it ends up being more or less the same I believe?

CookiePLMonster: True, I missed that.

xTVaser: If only C++ had a proper `join` / `interleave` 👎

pcsx2/GameDatabase.cpp
```diff
@@ -14,63 +14,174 @@
...
+		if (i != memcardFilters.size() - 1)
+			filters.append(",");
+	}
+	return filters;
 }
 
-// Sets the current game to the one matching the serial id given
-// Returns true if game found, false if not found...
-bool BaseGameDatabaseImpl::findGame(Game_Data& dest, const wxString& id) {
+std::vector<std::string> YamlGameDatabaseImpl::convertMultiLineStringToVector(const std::string multiLineString)
```
CookiePLMonster, line 40: Nitpick: AFAIK it can be a const reference.

CookiePLMonster (approved):

kenshen112:
I Just tested Rachet and Clank and saw none of the previous SPS that would occur without the Kickstart hack enabled so I can confirm the auto-fixes are applying for me.

pcsx2/R5900.h
```diff
@@ -231,21 +231,21 @@ struct tlbs
...
 
-#define _Funct_  ((cpuRegs.code      ) & 0x3F)  // The funct part of the instruction register
-#define _Rd_     ((cpuRegs.code >> 11) & 0x1F)  // The rd part of the instruction register
-#define _Rt_     ((cpuRegs.code >> 16) & 0x1F)  // The rt part of the instruction register
-#define _Rs_     ((cpuRegs.code >> 21) & 0x1F)  // The rs part of the instruction register
-#define _Sa_     ((cpuRegs.code >>  6) & 0x1F)  // The sa part of the instruction register
-#define _Im_     ((u16)cpuRegs.code) // The immediate part of the instruction register
-#define _Target_ (cpuRegs.code & 0x03ffffff)    // The target part of the instruction register
+#define _Funct_			((cpuRegs.code      ) & 0x3F)  // The funct part of the instruction register
+#define _Rd_			((cpuRegs.code >> 11) & 0x1F)  // The rd part of the instruction register
```
kenshen112, line 235: Unnecessary formatting changes

pcsx2/gui/AppGameDatabase.cpp
```diff
@@ -148,42 +50,39 @@ AppGameDatabase& AppGameDatabase::LoadFromFile(const wxString& _file, const wxSt
 
 		//wxDirName dir = InstallFolder;
 		wxDirName dir = (wxDirName)wxFileName(wxStandardPaths::Get().GetExecutablePath()).GetPath();
-		file = ( dir + file ).GetFullPath();
+		file = (dir + file).GetFullPath();
 	}
-	
-	
+
+
```
kenshen112, line 56: Unessacary new lines

pcsx2/Pcsx2Config.cpp
```diff
@@ -35,10 +35,35 @@ void TraceLogFilters::LoadSave( IniInterface& ini )
...
+			vu1Instant = enabled;
+			break;
+			jNO_DEFAULT;
+	}
+}
+
 Pcsx2Config::SpeedhackOptions::SpeedhackOptions()
 {
 	DisableAll();
-
```
kenshen112, line 41: extra newline

pcsx2/Patch.cpp
```diff
@@ -225,11 +216,13 @@ int LoadPatchesFromZip(wxString gameCRC, const wxString& patchesArchiveFilename)
 	{
 		wxString name = entry->GetName();
 		name.MakeUpper();
-		if (name.Find(gameCRC) == 0 && name.Find(L".PNACH")+6u == name.Length()) {
+		if (name.Find(gameCRC) == 0 && name.Find(L".PNACH") + 6u == name.Length())
+		{
 			PatchesCon->WriteLn(Color_Green, L"Loading patch '%s' from archive '%s'",
-				WX_STR(entry->GetName()), WX_STR(patchesArchiveFilename));
+								WX_STR(entry->GetName()), WX_STR(patchesArchiveFilename));
```
kenshen112, line 222: Looks like your formatter added extra spacing?

pcsx2/Patch.cpp
```diff
@@ -191,16 +179,18 @@ static int _LoadPatchFiles(const wxDirName& folderName, wxString& fileSpec, cons
...
 			PatchesCon->WriteLn(Color_Green, L"Found %s file: '%s'", WX_STR(friendlyName), WX_STR(buffer));
 			int before = Patch.size();
 			f.Open(Path::Combine(dir.GetName(), buffer));
 			inifile_process(f);
 			f.Close();
 			int loaded = Patch.size() - before;
 			PatchesCon->WriteLn((loaded ? Color_Green : Color_Gray), L"Loaded %d %s from '%s' at '%s'",
-				loaded, WX_STR(friendlyName), WX_STR(buffer), WX_STR(folderName.ToString()));
+								loaded, WX_STR(friendlyName), WX_STR(buffer), WX_STR(folderName.ToString()));
 			numberFoundPatchFiles++;
```
kenshen112, line 194: Spacing issues here?

pcsx2/Patch.cpp
```diff
@@ -40,66 +41,65 @@ wxString strgametitle;
...
+	PATCHTABLEFUNC* func;
 };
 
 static const PatchTextTable commands_patch[] =
-{
-	{ 1, L"author",		PatchFunc::author},
-	{ 2, L"comment",	PatchFunc::comment },
-	{ 3, L"patch",		PatchFunc::patch },
-	{ 0, wxEmptyString, NULL } // Array Terminator
+	{
```
kenshen112, line 50: Extra spacing

kenshen112 (changes requested):

pcsx2/R5900.h
xTVaser: I think this is just classic "tabs look awful".  In visual studio it appears to have done this to properly line things up 
image
If i run clang on it, it removes this style of formatting which I've seen push back against before so I think we'll want to leave this.

pcsx2/Pcsx2Config.cpp
xTVaser: Clang output - it got rid of the trailing whitespace (line used to be just a tab).
https://github.com/PCSX2/pcsx2/blob/678829a5b2b8ca7a3e42d8edc9ab201bf00b0fe9/pcsx2/Pcsx2Config.cpp#L41

pcsx2/Patch.cpp
xTVaser: Yeah, this is clang's output

pcsx2/Pcsx2Config.cpp
xTVaser: output from clang

xTVaser:
Besides the code in r5900.h which is hand-formatted to be column-aligned somewhat (seems like VS tried to maintain it) the rest of the spacing comments were just related to clang-format's.  I'm not sure what the official stance is, it seems like we are migrating towards clang-format'ing the code as we go (ideally without bloating the PR diff), but I have seen back-and-forth on this.

In general, I'm not going to hand-format my code, I run clang on the files / functions I modify.

tadanokojin:
My stance has been that it's fine as a separate commit last in the set (assuming you're going to rebase the merge and not squash it, then it's kind of a mess) since that won't make a mess of the history. I disable clang format on my IDE and run it manually on the command line.

#3985 is an example of how I've done it.

pcsx2/R5900.h
tellowkrinkle: If you mean pushback from me, I like the alignment but would prefer if it was made with spaces (like the previous ones were)

Tabs are for indentation, not alignment

xTVaser:
Well I typically format the individual functions I'm modifying or I'm in a file that's already been clang formatted (hence why I avoided running it on the R5900.h file altogether). If the formatting was creating huge diffs I would take your advice, but for the most part I use formatters as a tool to speed up my writing, so saving it for the end would hinder me.

pcsx2/R5900.h
xTVaser: Ah i see now, I wouldn't have expected visual studio to change the entire alignment to tabs when running a simple find/replace so I just assumed it added similar characters to correctly align.  I'll fix that.  Sorry I'm not used to looking at whitespace this closely in diffs, I'm used to working exclusively with spaces.

refractionpcsx2 (approved):

RedDevilus (approved):

tellowkrinkle (approved):

lightningterror (approved):

AmFobes (approved):

tadanokojin (approved):

xTVaser:
Just squashing / cleaning up commits a bit in preparation for upcoming merge.

refractionpcsx2:
Seems to work good. One question, can we not name GameIndex.md as GameIndexHelp.txt or something? a .md file isn't very user friendly on windows

tadanokojin:
All our other documentation is .md so I take it vaser is just following protocol. I can suggest some good markdown readers for windows if you like :P

refractionpcsx2:
I mean I'm fine with it, since I have decent tools, just thinking more from a user friendlyness point of view :P

tadanokojin:
Well the idea of doing .md is you can easily convert it to HTML or PDF or whatever and github has a nice viewer for it.

refractionpcsx2:
alright, I guess leave it for now, but then it might be worth later on moving it in to Docs with the rest of them instead of having it in the bin root.

PR #3782, opened 2020-10-03
Merge DEV9ghrz into core

GovanifY:
As we all know, ethernet was an utter failure(see picture below) and that's the reason why we did not want to include it by default in PCSX2.
But plugin merging time has come, and so it is time to merge yet another plugin into the core!

~~I _highly_ recommend merging https://github.com/PCSX2/pcsx2/pull/3756 and waiting for me to rebase this PR before you review stylistic differences(eg tabs, brackets, etc etc) since the SPU2 PR formats core files which DEV9 modifies, but DEV9 is currently based on the master branch which does not yet have those modifications.~~
Ready for review!

image

PR #3929, opened 2020-11-14
CI: Remove Travis-CI and AppVeyor

xTVaser:
Travis-CI has been failing for over a week now, and I don't think there's intentions nor an incentive to fix it with Actions taking it's place.

Bring back the green checkmark? 
image


tadanokojin (approved):
#BringGreenBack

tellowkrinkle:
If we're not fixing travis, we need to increase the minimum required cmake to 3.16, as one of the reasons it's failing is because the build doesn't work without PCH (which was added in cmake 3.16)

arcum42:
That's odd, since it should just not precompile it, which is what it used to do all the time when you built pcsx2.

What I do see, looking at the travis failures, is that it's not finding libsamplerate, which isn't surprising because it's a new dependency due to one of the plugin merges.

turtleli:
> That's odd, since it should just not precompile it, which is what it used to do all the time when you built pcsx2.

The non precompiled header build broke when #3782 was merged.

tellowkrinkle:
When you enable PCH, the precompiled header is implicitly included by every cpp file

Some of our cpp files rely on things from the precompiled header but don't `#include` it

tadanokojin:
Is 3.16 going to be too high? Looking at Debian stable it seems they're on 3.13 https://packages.debian.org/buster/cmake

tellowkrinkle:
Personally I think we should keep non-PCH compatibility, and keep travis around to make sure we keep non-PCH compatibility

xTVaser:
Would it not be better for someone to add an Actions configuration that tests non-PCH compatibility?  Travis only runs on master, so you'd only find out _after_ a PR merges if there is a regression.  If we are going to care about the results of Travis, it should be moved back to running on every PR.

tellowkrinkle:
I think it's easiest to enable Travis on PRs, I don't think its speed was ever much of an issue since we get up to 5 builds running concurrently

RedDevilus:
What about the new price changes especially for mac building? 
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

lightningterror:
Looks like Red is offering to cover our costs.
It would be nice to fix travis first.

RedDevilus:
I mean if macOS builds are safe in the future? But i guess we can just look at Actions instead for those.

lightningterror (approved):

tadanokojin:
tellow took care of no-pch build in #3942 so this just needs to drop the no pch commit and rebase and and it should be good.

PR #3930, opened 2020-11-14
[USB+] Clean up Warnings

arcum42:
Clean up various warnings, largely in the recently merged USB code. Unused variables have had [[maybe_unused]] added or been deleted, various variables were cast due to sign mismatches, member variables were rearranged in places to match the class initialization, and so on.

This brings the warnings in gcc and clang down to a much more reasonable level.

I ignored several warnings along the line of:
../pcsx2/USB/USB.cpp:520:10: warning: ‘char* strncpy(char*, const char*, size_t)’ output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
  520 |   strncpy(usbd.freezeID, USBfreezeID, strlen(USBfreezeID));

As well as various preexisting warnings (Such as Lilypad doing sizeof on a pointer :P ).

### DIFF
--- a/pcsx2/USB/usb-mic/audiodev-pulse.cpp
+++ b/pcsx2/USB/usb-mic/audiodev-pulse.cpp
@@ -334,7 +334,7 @@ namespace usb_mic
 			if (mPAready == 3 && dur >= 1000)
 			{
 				mLastGetBuffer = now;
-				int ret = pa_context_connect(mPContext,
+				[[maybe_unused]]int ret = pa_context_connect(mPContext,
 											 mServer,
 											 PA_CONTEXT_NOFLAGS,
 											 NULL);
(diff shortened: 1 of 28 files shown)


jackun (approved):

pcsx2/USB/usb-mic/audiodev-pulse.cpp
```diff
@@ -334,7 +334,7 @@ namespace usb_mic
 			if (mPAready == 3 && dur >= 1000)
 			{
 				mLastGetBuffer = now;
-				int ret = pa_context_connect(mPContext,
+				[[maybe_unused]]int ret = pa_context_connect(mPContext,
```
GovanifY, line 337: This is a bit of a mouthful to read, why not put maybe_unused on the line above? Do we have a guideline on same-line [[]] blocks?

GovanifY (commented):
Thanks a lot for the work arcum! just a minor nitpick

pcsx2/USB/usb-mic/audiodev-pulse.cpp
arcum42: Don't think we have any sort of guideline on attributes yet, since we only recently started adding them to pcsx2.

lightningterror (approved):

pcsx2/USB/usb-mic/audiodev-pulse.cpp
arcum42: I went ahead and moved it up a line on this one file.

GovanifY (approved):

PR #3934, opened 2020-11-17
USB: Fix some issues with save states

jackun:
Fix some things I noticed. There could be lurking more.


Florin9doi:
is there any reason why `#if defined(BUILD_RAW)` is used ?

jackun:
Historically to allow disabling raw input support from being built but currently probably pointless as is BUILD_DX.

lightningterror:
Maybe some commits can be squashed.

PR #3935, opened 2020-11-18
Don't reopen everything when we just want to reopen the CDVD

TheLastRar:
This was causing DEV9 code to crash

Possibly also fixes some CDVD issues that where apparently happening?

### DIFF
--- a/pcsx2/System/SysThreadBase.cpp
+++ b/pcsx2/System/SysThreadBase.cpp
@@ -235,7 +235,7 @@ void SysThreadBase::Resume()
 			if (m_ExecMode == ExecMode_Opened)
 				return;
 		}
-			// fall through...
+			[[fallthrough]];
 
 		case ExecMode_Closing:
 		case ExecMode_Pausing:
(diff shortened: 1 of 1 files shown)


lightningterror:
Can you run clang format on the file while you're at it?

TheLastRar:
Done

pcsx2/System/SysThreadBase.cpp
```diff
@@ -199,48 +209,55 @@ void SysThreadBase::PauseSelfDebug()
...
 			Start();
-			if( !m_running || (m_ExecMode == ExecMode_NoThreadYet) )
+			if (!m_running || (m_ExecMode == ExecMode_NoThreadYet))
 				throw Exception::ThreadCreationError(this);
-			if( m_ExecMode == ExecMode_Opened ) return;
+			if (m_ExecMode == ExecMode_Opened)
+				return;
 		}
-		// fall through...
+			// fall through...
```
refractionpcsx2, line 238: Can you get rid of the space, dunno why Clang did that. maybe better still replace it with [[fallthrough]];

TheLastRar: Done

lightningterror (approved):

refractionpcsx2 (approved):

tadanokojin (approved):

PR #3940, opened 2020-11-21
GSVertexTrace::FindMinMax improvements

tellowkrinkle:
*when you spend a bunch of time optimizing a function because you think it accounts for 30% of runtime but that was actually due to the compiler being stupid and not issues with the algorithm*

This PR contains the following changes:

- Prevents clang from optimizing out our denormal-removal shuffles (10x faster than before for people who compile with clang!)
- Run divides on four elements at a time instead of two elements and two useless numbers
- ~~Run per-vertex instead of per-primitive for all things that need to apply to all vertices of a primitive anyways (prevents double-checking vertices in triangle strips, and also reduces pointer chases in the main loop)~~ This broke stuff
- Remove inaccurate stq
  - With the above division improvements, on processors with partially-pipelined division (Ivy Bridge and later, Bulldozer and later), accurate stq is actually faster (according to both IACA of [inaccurate](https://gist.github.com/tellowkrinkle/36124c0871ce3bd1ebdb5d6a7dd5e1b4#file-findminmax00-analysis-hsw-txt) vs [accurate](https://gist.github.com/tellowkrinkle/36124c0871ce3bd1ebdb5d6a7dd5e1b4#file-findminmax01-analysis-hsw-txt) and [LLVM MCA](https://analysis.godbolt.org/z/xqvY64)).  On older CPUs expect performance to be about 2/3 of the old algorithm before taking into account improvements from not double-checking vertices.  
  - ~~There seems to be a check of `accurate_stq` when the OGL backend is deciding whether to use geometry shaders to process sprites, does anyone know if that's important or just something someone threw in there that's okay to remove?~~ Reason found, added as comment

In the end, ignoring clang issues, GSVertexTrace::FindMinMax goes from taking about 3% of MTGS thread runtime to 1.5% on my computer.  (Most of the time was spent doing OpenGL things so if you have a more efficient OpenGL driver it might make more of a difference for you)

### DIFF
--- a/pcsx2/GS/GSVector4.h
+++ b/pcsx2/GS/GSVector4.h
@@ -205,6 +205,17 @@ public:
 		return m;
 	}
 
+	/// Makes Clang think that the whole vector is needed, preventing it from changing shuffles around because it thinks we don't need the whole vector
+	/// Useful for e.g. preventing clang from optimizing shuffles that remove possibly-denormal garbage data from vectors before computing with them
+	__forceinline GSVector4 noopt()
+	{
+		// Note: Clang is currently the only compiler that attempts to optimize vector intrinsics, if that changes in the future the implementation should be updated
+#ifdef __clang__
+		__asm__("":"+x"(m)::);
+#endif
+		return *this;
+	}
+
 	__forceinline uint32 rgba32() const
 	{
 		return GSVector4i(*this).rgba32();
--- a/pcsx2/GS/Renderers/Common/GSVertexTrace.cpp
+++ b/pcsx2/GS/Renderers/Common/GSVertexTrace.cpp
@@ -27,8 +27,7 @@ GSVertexTrace::GSVertexTrace(const GSState* state)
 	memset(&m_alpha, 0, sizeof(m_alpha));
 
 	#define InitUpdate3(P, IIP, TME, FST, COLOR) \
-		m_fmm[0][COLOR][FST][TME][IIP][P] = &GSVertexTrace::FindMinMax<P, IIP, TME, FST, COLOR, 0>; \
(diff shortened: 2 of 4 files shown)


orbea:
This brings the slowdowns on Xenosaga Episode 1 during the third cut scene from ~40-45 fps to ~50-55 fps when using clang11.

tellowkrinkle:
> This brings the slowdowns on Xenosaga Episode 1 during the third cut scene from ~40-45 fps to ~50-55 fps when using clang11.

Curious, after this patch, are clang builds slower than gcc builds here?

frymezim:
@iMineLink or @gregory38 might be able to help review this pull request.

orbea:
> Curious, after this patch, are clang builds slower than gcc builds here?

I tested this again with commit https://github.com/PCSX2/pcsx2/commit/88983caaba01ad69d90569ca16f1267977219fca in the first slowdown which starts in the third cutscene right after KOS-MOS appears and lasts for two lines of dialogue.

* gcc-9.3.0 without this PR: 50-55 fps
* clang-11.0.0 without this PR: ~45 fps (Clearly slower than gcc)
* gcc-9.3.0 with this PR: ~50-55 FPS (About the same or maybe slightly slower than without this PR?)
*clang-11.0.0 with this PR: 50-55 fps (Roughly the same as gcc)

I think this PR doesn't make a significant difference when gcc while it clearly improves performance with clang so that it matches gcc better.

lightningterror:
I'd add a prefix to the last commit `gsdx-ogl:`.

lightningterror:
Would be nice to rename GSdx to GS in commit names as GS was merged.

Issue #3918, opened 2020-11-10
Meta: Compatibility issues with folder memory cards


MrCK1:
This is a list of the current games that have compatibility issues with folder memory cards:

[ ] Freedom Fighters - Saving works, corrupted data upon load - https://github.com/PCSX2/pcsx2/issues/2048
[ ] Dogz - Saving doesn't work (unconfirmed) - https://github.com/PCSX2/pcsx2/issues/2809
[ ] SoTC (SCPS-15097) - no files appear on card after conversion from .p2s -  https://github.com/PCSX2/pcsx2/issues/3616

gabriele2000:
> 
> 
> This is a list of the current games that have compatibility issues with folder memory cards:
> 
> [ ] Freedom Fighters - Saving works, corrupted data upon load - #2048
> [ ] Dogz - Saving doesn't work (unconfirmed) - #2809
> [ ] SoTC (SCPS-15097) - no files appear on card after conversion from .p2s - #3616

[ ] SSX: On tour - Saving doesn't work for profiles, only settings are saved

MageKing17:
[ ] DDRMAX: Dance Dance Revolution - Saving seems to work, but loading results in a "System data is corrupted" message. (Confirmed fixed by #3941.)

failzers:
> [ ] SSX: On tour - Saving doesn't work for profiles, only settings are saved
Just tested on 1.7.3255, saves both perfectly

refractionpcsx2:
Thanks for the update :)

stoniemahonie:
Outrun 2 SP (SLPM-66628) seems to have problems, can save but can't load afterwards, doesn't find it. 8 MB Card work fine.

Attached folder- and 8mb card.

[outrun2sp_slpm-66628.zip](https://github.com/PCSX2/pcsx2/files/10410721/outrun2sp_slpm-66628.zip)

Mrlinkwii:
#6708 , #1752, #7166 are a list of other folder memcard issues

Scall:
There's also: https://github.com/PCSX2/pcsx2/issues/8599

TheTechnician27:
Sort of redundant, but I can confirm On Tour works without issue.

TheTechnician27:
[...]

PR #3941, opened 2020-11-22
Folder memcards: Add an index file to track order of files

### LINKED ISSUE #3888: [Folder Memory Card] Can't load GTA Vice City Stories save
***PCSX2 version:***
v1.6.0 from [flatpak](https://flathub.org/apps/details/net.pcsx2.PCSX2).

***PCSX2 options:***
- MTVU: Yes
- All other emulation options at recommended values

***Plugins used:***
- GSdx 1.2.0 AVX2
    - Renderer: OpenGL Hardware
    - All other options at recommended values
- SPU2-X 2.0.0
- OnePAD 0 2.0.0

In GTA Vice City Stories (PAL), when saving to a folder memory card, the saves are no longer visible from within the game after PCSX2 is restarted. Loading works as long as PCSX2 is not closed between saving and loading. Using a normal 8 MB memory card works as expected. Saves for both folder and 8 MB memory cards attached. Similar to issue #1136 but it seems closed and not getting any attention.

[GTAVCS_folder.ps2.zip](https://github.com/PCSX2/pcsx2/files/5475869/GTAVCS_folder.ps2.zip)
[GTAVCS.ps2.zip](https://github.com/PCSX2/pcsx2/files/5475868/GTAVCS.ps2.zip)

1. Save a game
2. Close PCSX2
3. Restart PCSX2
4. The game will start from the beginning and no saves will be visible in the load game screen.

None

***PC specifications:***
OS: Debian 10	
CPU: Intel Core i5-4570 CPU @ 3.20 GHz
GPU: NVIDIA GeForce GTX 760 2 GB
RAM: 8 GB

CookiePLMonster:
**~This PR depends on #3950 because of its addition of yaml-cpp. Do not merge it before that PR.~**

This PR adds an index file for folder memory cards in order to track timestamps and order of files in the FAT partition properly. This normally shouldn't matter, but there are at least a few games (ie. GTA games) which rely on the order of files on the memory card. Previously, folder memcards would not care about the order and populate the partition in whatever order the host filesystem provided the files (so, in the case of NTFS partitions, alphabetically). In reality, FAT partitions don't specify the order of files, which means in practice they are sorted by the order of creation.

I added a YAML based index file to every subdirectory in folder memory cards to retain the original order in which game writes data to the card. Additionally, instead of relying on host timestamps, they are preserved in the index file as well. This resolves issues where moving/copying the folder memory card around would change timestamps.

I verified this against all 5 GTAs and they work properly. Vice City Stories still decided to reject the memory card if I converted the files from a regular memory card to a folder memory card, but converting back and forth now works fine instead of producing "corrupted" blocks (as far as the game was concerned). This is something worth revisiting in the future, but it's an extreme edge case.

Fixes #3888. Also fixes saving in **all** GTA games, not just Vice City Stories.
Test with games from #3918, as some/all of those issues may be potentially fixed.

### DIFF
--- a/pcsx2/gui/MemoryCardFolder.cpp
+++ b/pcsx2/gui/MemoryCardFolder.cpp
@@ -22,9 +22,26 @@
 #include "System.h"
 #include "AppConfig.h"
 
+#include "yaml-cpp/yaml.h"
+
 #include "svnrev.h"
 
 bool RemoveDirectory( const wxString& dirname );
+
+// A helper function to parse the YAML file
+static YAML::Node LoadYAMLFromFile( const wxString& fileName ) {
+	YAML::Node index;
+
+	wxFFile indexFile;
+	if ( indexFile.Open( fileName, L"r" ) ) {
+		wxString fileContents;
+		if ( indexFile.ReadAll( &fileContents ) ) {
+			index = YAML::Load( fileContents.mbc_str() );
+		}
+	}
+
+	return index;
+}
 
 FolderMemoryCard::FolderMemoryCard() {
 	m_slot = 0;
@@ -312,10 +329,7 @@ bool FilterMatches( const wxString& fileName, const wxString& filter ) {
 }
 
 bool FolderMemoryCard::AddFolder( MemoryCardFileEntry* const dirEntry, const wxString& dirPath, MemoryCardFileMetadataReference* parent, const bool enableFiltering, const wxString& filter ) {
-	wxDir dir( dirPath );
-	if ( dir.IsOpened() ) {
-		wxString fileName;
-		bool hasNext;
+	if ( wxDir::Exists( dirPath ) ) {
 
 		wxString localFilter;
 		if ( enableFiltering ) {
@@ -328,39 +342,31 @@ bool FolderMemoryCard::AddFolder( MemoryCardFileEntry* const dirEntry, const wxS
 		}
 
 		int entryNumber = 2; // include . and ..
-		hasNext = dir.GetFirst( &fileName );
-		while ( hasNext ) {
-			if ( fileName.StartsWith( L"_pcsx2_" ) ) {
-				hasNext = dir.GetNext( &fileName );
-				continue;
-			}
+		for ( const auto& file : GetOrderedFiles( dirPath ) ) {
 
-			wxFileName fileInfo( dirPath, fileName );
+			wxFileName fileInfo( dirPath, file.m_fileName );
 			bool isFile = wxFile::Exists( fileInfo.GetFullPath() );
 
 			if ( isFile ) {
 				// don't load files in the root dir if we're filtering; no official software stores files there
 				if ( enableFiltering && parent == nullptr ) {
-					hasNext = dir.GetNext( &fileName );
 					continue;
 				}
-				if ( AddFile( dirEntry, dirPath, fileName, parent ) ) {
+				if ( AddFile( dirEntry, dirPath, file, parent ) ) {
 					++entryNumber;
 				}
 			} else {
 				// if possible filter added directories by game serial
 				// this has the effective result of only files relevant to the current game being loaded into the memory card
 				// which means every game essentially sees the memory card as if no other files exist
-				if ( enableFiltering && !FilterMatches( fileName, localFilter ) ) {
-					hasNext = dir.GetNext( &fileName );
+				if ( enableFiltering && !FilterMatches( file.m_fileName, localFilter ) ) {
 					continue;
 				}
 
 				// make sure we have enough space on the memcard for the directory
-				const u32 newNeededClusters = CalculateRequiredClustersOfDirectory( dirPath + L"/" + fileName ) + ( ( dirEntry->entry.data.length % 2 ) == 0 ? 1 : 0 );
+				const u32 newNeededClusters = CalculateRequiredClustersOfDirectory( dirPath + L"/" + file.m_fileName ) + ( ( dirEntry->entry.data.length % 2 ) == 0 ? 1 : 0 );
 				if ( newNeededClusters > GetAmountFreeDataClusters() ) {
-					Console.Warning( GetCardFullMessage( fileName ) );
-					hasNext = dir.GetNext( &fileName );
+					Console.Warning( GetCardFullMessage( file.m_fileName ) );
 					continue;
 				}
 
@@ -377,19 +383,19 @@ bool FolderMemoryCard::AddFolder( MemoryCardFileEntry* const dirEntry, const wxS
 
 				// set metadata
 				wxFileName metaFileName( dirPath, L"_pcsx2_meta_directory" );
-				metaFileName.AppendDir( fileName );
+				metaFileName.AppendDir( file.m_fileName );
 				wxFFile metaFile;
 				if ( metaFileName.FileExists() && metaFile.Open( metaFileName.GetFullPath(), L"rb" ) ) {
 					size_t bytesRead = metaFile.Read( &newDirEntry->entry.raw, sizeof( newDirEntry->entry.raw ) );
 					metaFile.Close();
 					if ( bytesRead < 0x60 ) {
-						strcpy( reinterpret_cast<char*>(newDirEntry->entry.data.name), fileName.mbc_str() );
+						strcpy( reinterpret_cast<char*>(newDirEntry->entry.data.name), file.m_fileName.mbc_str() );
 					}
 				} else {
 					newDirEntry->entry.data.mode = MemoryCardFileEntry::DefaultDirMode;
 					newDirEntry->entry.data.timeCreated = MemoryCardFileEntryDateTime::FromWxDateTime( creationTime );
 					newDirEntry->entry.data.timeModified = MemoryCardFileEntryDateTime::FromWxDateTime( modificationTime );
-					strcpy( reinterpret_cast<char*>(newDirEntry->entry.data.name), fileName.mbc_str() );
+					strcpy( reinterpret_cast<char*>(newDirEntry->entry.data.name), file.m_fileName.mbc_str() );
 				}
 
 				// create new cluster for . and .. entries
@@ -416,8 +422,6 @@ bool FolderMemoryCard::AddFolder( MemoryCardFileEntry* const dirEntry, const wxS
 				// and add all files in subdir
 				AddFolder( newDirEntry, fileInfo.GetFullPath(), dirRef );
 			}
-
-			hasNext = dir.GetNext( &fileName );
 		}
 
 		return true;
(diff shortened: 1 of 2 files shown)


CookiePLMonster:
Rebased on top of #3950 to remove commits from the previous "parent" PR. Still marked as draft until #3950 is merged.

CookiePLMonster:
The dependant PR is now merged, so this PR is ready for a review.

RedDevilus:
I've imported my old folder memcard filled with all my saves, upon inserting it to port1 and also starting will hard crash PCSX2 for  me.
```
McdSlot 0 [File]: [is not memcard file]
McdSlot 1 [File]: D:\PCSX2_backup\Special builds\PRs\3941\memcards\Mcd002.ps2
McdSlot 0: [Folder] D:\PCSX2_backup\Special builds\PRs\3941\memcards\RedDevilus3000.ps2
(FolderMcd) Indexing slot 0 with filter "".
Closing plugins...
[wx] can't open file 'D:\PCSX2_backup\Special builds\PRs\3941\memcards\RedDevilus3000.ps2\_pcsx2_index' (error 2: the system cannot find the file specified.)
[wx] can't open file 'D:\PCSX2_backup\Special builds\PRs\3941\memcards\RedDevilus3000.ps2\BADATA-SYSTEM\_pcsx2_index' (error 2: the system cannot find the file specified.)
    Closing PAD
    Closing GS
Delete 0 Shaders, 34 Programs, 28 Pipelines
Plugins closed successfully.
Shutting down plugins...
Plugins shutdown successfully.
```
Inside of BADATA-System it doesn't create the index:
image

I also can't rename my folder memcard but it works fine on the precreated ones.
image

```
Apply memory cards:
No active slots.
[wx] Failed to copy the file 'D:\PCSX2_backup\Special builds\PRs\3941\memcards\RedDevilus3000.ps2' to 'D:\PCSX2_backup\Special builds\PRs\3941\memcards\TestMcd001.ps2' (error 5: access is denied.)
[wx] File 'D:\PCSX2_backup\Special builds\PRs\3941\memcards\RedDevilus3000.ps2' couldn't be renamed 'D:\PCSX2_backup\Special builds\PRs\3941\memcards\TestMcd001.ps2' (error 0: the operation completed successfully.)
Apply memory cards:
slot[0]='RedDevilus3000.ps2'
Apply memory cards:
slot[0]='RedDevilus3000.ps2'
Apply memory cards:
slot[0]='RedDevilus3000.ps2
```
I checked on a 1.7 dev build to see if its not permission issues or the folder memcard itself being broken but everything works there.

RedDevilus:
Additionally formatting a file memcard, making a save and then converting to folder produces another error for the index. It doesn't seem to want to create it at first sight (while it does if i look into the folder memcard later)
```
[wx] can't open file 'D:\PCSX2_backup\Special builds\PRs\3941\memcards\Mcd001-Or_converted.ps2\BESLES-55145\_pcsx2_index' (error 2: the system cannot find the file specified.)
(FolderMcd) Done! Took 41 ms.
Apply memory cards:
slot[0]='Mcd001-Or.ps2'
Apply memory cards:
slot[0]='Mcd001-Or_converted.ps2'
```
Starting a game produces this:
image
```
McdSlot 0: [Folder] D:\PCSX2_backup\Special builds\PRs\3941\memcards\Mcd001-Or_converted.ps2
(FolderMcd) Indexing slot 0 with filter "SLES-55145".
[wx] can't open file 'D:\PCSX2_backup\Special builds\PRs\3941\memcards\Mcd001-Or_converted.ps2\_pcsx2_index' (error 2: the system cannot find the file specified.)
(pxActionEvent) (Pcsx2AppMethodEvent) STL Runtime Error: invalid node; first invalid key: "BESLES-55145"(SysExecute) 
```
When i want to convert the folder back to 8 MB file memcard and it hardcrashes afterwards.

image


This is the inside of the memcard:
image
image

CookiePLMonster:
Converted to draft until I get the chance to review the above issues.

CookiePLMonster:
Fixed bugs mentioned by @RedDevilus above, should be OK now.

RedDevilus (approved):
Other than
 `[wx] can't open file 'D:\PCSX2_backup\Special builds\PRs\3941\v2\memcards\McD001Conv.ps2\_pcsx2_index' (error 2: the system cannot find the file specified.)`
 which is normal from what i understood, everything seems to be working as normal. 🥇

CookiePLMonster:
Updated with a fix for a crash when writing the index for zero-sized files. Fixes a crash in **Guitar Hero III: Legends of Rock**.

RedPanda4552:
image
Got the above when starting a fresh card. Not sure if this message was erroneous or not, the index appears to be there.
image
Also worth noting most files appeared as 0 bytes in the above. After closing the emu, they were showing sizes. Again unsure if an error or if this is just Windows being weird about new files.
image

CookiePLMonster:
> Got the above when starting a fresh card. Not sure if this message was erroneous or not, the index appears to be there.

It's intended because the best way to check if the file exists is just to try to open it - if it fails, it doesn't exist.

> Also worth noting most files appeared as 0 bytes in the above. After closing the emu, they were showing sizes. Again unsure if an error or if this is just Windows being weird about new files.

Also intended and not affected by this PR - PCSX2 doesn't flush writes explicitly and Explorer is rather slow at updating those sizes. I know PCSX2 closes written files properly ASAP which should flush them quickly, but I guess Explorer doesn't update them instantly.

Anuskuss:
@CookiePLMonster Sorry for the dumb question, but if the order of files is determined by the creation date, why not just use the file timestamps for that? All mayor OSes track the creation date and it's less likely to break.

CookiePLMonster:
There are at least two reasons I can think of which make this a bad idea:
1. The order of files flushed to disk may not correspond to the order they are on the FAT partition.
2. File creation dates will almost certainly break with backups, and a "I restored my backed up memcard and now it's broken" is a terrible scenario.

Anuskuss:
Yeah, you're right, didn't really consider those scenarios. I just didn't like the idea of introducing an extra file just to keep track of the order.
What do you think about ZIPs? The order would still be preserved and the filesystem would be accessible, although you would lose the ability to just drop the files in there. Would probably need to write an importer inside PCSX2, meaning more work for you. So honestly this seems like a decent solution.

RedDevilus:
Introducing zips sounds more like a bad replication of how file memcard is handled.

Anuskuss:
Tbh, I didn't even know that you could have folder memory cards. I always thought if you want to use the savedata from somewhere, you'd have to set that as memcard2 and import it into your memcard1 in the BIOS. Anyway, making them ZIPs wouldn't be much different from just having a file explorer which reads memory cards inside PCSX2.

CookiePLMonster:
I'm not completely opposed to that idea, but I don't think it's worth the effort currently.

RedPanda4552:
> 
> 
> Introducing zips sounds more like a bad replication of how file memcard is handled.

I don't really agree with this. File based memcards are binary blobs. There is no directory structure accessible by the host OS. The blob file is literally what would be the entire contents of the flash on a real memory card. Using a zip file would allow us to still have the host OS access the contents, but also (provided it is true that zip files maintain their created dates internally regardless of how many times the zip is copied/moved) just let us read those times to determine order when rebuilding the FAT.

We could come up with all sorts of other solutions. Even going as far as storing a meta file for the FAT (which would be neat because that would be one way to have saves show up in the BIOS). But for now I think this is satisfactory and we can explore other alternatives once this is done.

CookiePLMonster:
> Even going as far as storing a meta file for the FAT

FYI this is already happening if the memory card block name has characters which are illegal in NTFS.

Anuskuss:
Well, ZIPs are really just files stored after one another so after one file gets put in there, it's position shouldn't change (atleast when tested in 7-zip). Also ZIPs are historically made with FAT in mind, support compression, have UTF-8 file names, can store comments (could maybe eliminate those other `_pcsx2` files) and can be opened natively in the Windows 10 explorer which would remove a lot of headache introduced by folder memcards. I'm a bit worried about non-compliant ZIPs implementation though (e.g. reordering files on change).

I think the ultimate goal would be a file explorer inside PCSX2, with the ability to import saves (from GameFAQs for example). But even then we could still use ZIPs as the on-disk format.

RedDevilus:
I'm not against the zips itself, just that it should maybe be addressed in a separate PR to avoid regressions.

I've made my statement a bit lazy that looks that i imply the notion to dismiss it now that i look at it, unless the author is confident enough for this PR to make some changes, but feature creep can settle in and make it a delay-hell for merge and testing. Not a big deal though at this time.

A save-manager inside PCSX2 itself would be cool, MyMC and MyMCPlus aren't the easiest save-managers although powerful and intituive in the way the can do multiple save types.

CookiePLMonster:
> I've made my statement a bit lazy that looks that i imply the notion to dismiss it now that i look at it, unless the author is confident enough for this PR to make some changes, but feature creep can settle in and make it a delay-hell for merge and testing. Not a big deal though at this time.

I agree, I intended this to be a PR improving folder memcard incompatibilities and transforming them to use zips sounds like introducing far more risk than the current changes. The idea itself is solid, but maybe not as a part of this PR.

refractionpcsx2:
Yeah making them zips is way out of scope of this PR, personally I don't see the advantage of doing that either, we already have NTFS compression for memcard folders so we can already compress them, adding zip just adds another layer of complication.

CookiePLMonster:
Rebased on the latest master and made the change required by Codacy.

RedDevilus (approved):

refractionpcsx2 (approved):

F0bes:
> > Got the above when starting a fresh card. Not sure if this message was erroneous or not, the index appears to be there.
> 
> It's intended because the best way to check if the file exists is just to try to open it - if it fails, it doesn't exist.

Touching on this again, I want to approve this pr because it functionally works but I find that the message looks too much like an error, because it technically is one.
I realize that this message comes from WX and is probably shared throughout pcsx2, therefore it's not like we can just change the message color, but is there _any_ way you can check if the file exists before trying to open it?

CookiePLMonster:
Does wx provide some functionality to suppress outputting that error? Some kind of signal blocker?

CookiePLMonster:
Rebased on latest master and added error suppression to the above case, as requested by @F0bes

F0bes (approved):
Functionally works with Liberty City Stories
Looks good to me.

SnkXIII:
This has allowed me to convert the VCS save that I took from my console in psu format to a pcsx2 folder type save that sticks.
The process went smoothly for me.

Thanks, I had no idea why it wasn't working before.

RedDevilus:
Other than using a prefix like `Folder memcards:`  for the rest of the commits, i see no reason to delay this PR any further.

CookiePLMonster:
> Other than using a prefix like `Folder memcards:` for the rest of the commits, i see no reason to delay this PR any further.

Reworded now!

lightningterror:
Maybe run clang on the two files as well, current formatting is awful.

lightningterror (approved):
It's been extensively tested by several people.

PR #3950, opened 2020-11-27
Add YAML-cpp submodule

xTVaser:
More than one active PR depend on this, and as to prevent either from being held up we can add the submodule independently.


CookiePLMonster (approved):
#3941 rebased on top of this PR works fine.

tadanokojin (approved):
normally I'm not big on just adding deps and not using them right away but I'll make an exception since it seems a lot of people have interest in this.

don't see anything wrong with the VS stuff, didn't check cmake.

kenshen112:
The CMake was tested working.

arcum42:
I would like to see yaml in the tree, partially because merged plugins often had separate configuration code for Windows and Linux, and if we were rewriting it, I hate to rewrite it not to be yaml, than redo it again as yaml later. (for example, with the merged copy of spu2-x, the separate config code is one reason I didn't just go ahead and make Windows use the wx dialog as well. The config is tangled in the gui.)

OTOH, when I compiled this with clang, it compiled, but with about a page of warnings from yaml, all from "-Wshadow". It'd be nice if we could suppress the compiler warnings from it, since we aren't going to be patching it to fix them...

tadanokojin:
@arcum42 
> I would like to see yaml in the tree

To be clear you're saying you want this merged and not objecting to it being a submodule?

> It'd be nice if we could suppress the compiler warnings from it

I don't think vaser is going to take a swing at that but I'm sure he'll accept a page or you can do so after it's merged.

arcum42:
I do want it merged, and it's not a blocker if it's adding warnings in both gcc & clang, I'd just prefer to avoid them if we can.

tadanokojin:
Alright, sounds good.

PR #3958, opened 2020-11-29
Fix build without pulseaudio.

orbea:
Fixes the build when pulseaudio is not installed. Sound when playing games still works fine.

### DIFF
--- a/cmake/SearchForStuff.cmake
+++ b/cmake/SearchForStuff.cmake
@@ -203,7 +203,6 @@ if(PULSEAUDIO_FOUND)
     include_directories(${PULSEAUDIO_INCLUDE_DIR})
 endif()
 
-
 if(LIBXML2_FOUND)
 	include_directories(${LIBXML2_INCLUDE_DIRS})
 endif()
--- a/pcsx2/CMakeLists.txt
+++ b/pcsx2/CMakeLists.txt
@@ -48,6 +48,10 @@ if(PORTAUDIO_FOUND)
     set(pcsx2FinalFlags ${pcsx2FinalFlags} -DSPU2X_PORTAUDIO)
 endif()
 
+if(PULSEAUDIO_FOUND)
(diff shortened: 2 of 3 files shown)


pcsx2/CMakeLists.txt
```diff
@@ -953,6 +958,10 @@ if(PORTAUDIO_FOUND)
     set(pcsx2FinalLibs ${pcsx2FinalLibs} ${PORTAUDIO_LIBRARIES})
 endif()
 
+if(PULSEAUDIO_FOUND)
+    set(pcsx2FinalLis ${pcsx2FinalLibs} ${PULSEAUDIO_LIBRARIES})
```
GovanifY, line 962: Probably wanted to use pcsx2FinalLibs there?

GovanifY (changes requested):
See comment above

pcsx2/CMakeLists.txt
orbea: Thanks, good catch.

cmake/SearchForStuff.cmake
```diff
@@ -203,7 +203,6 @@ if(PULSEAUDIO_FOUND)
     include_directories(${PULSEAUDIO_INCLUDE_DIR})
 endif()
 
-
```
GovanifY, line 206: Why remove this line?

GovanifY (commented):
Sorry, didn't catch that before, otherwise LGTM

cmake/SearchForStuff.cmake
orbea: Extra blank lines are a pet peeve of mine and I removed it by habit when looking at the above lines. I can revert this change if you wish? Although it fits the style of the rest of the file more like this.

GovanifY: No it's fine, I just wanted to know the reasonning, this is why I put it as a comment and not a request for changes

GovanifY (approved):
LGTM

PR #3972, opened 2020-12-04
Resolve a crash when EyeToy is opened the second time

Florin9doi:


pcsx2/USB/usb-eyetoy/cam-windows.cpp
```diff
@@ -455,6 +455,7 @@ namespace usb_eyetoy
 				samplegrabberfilter->Release();
 				samplegrabber->Release();
 				nullrenderer->Release();
+				sourcefilter = NULL;
```
lightningterror, line 458: Use nullptr instead?

lightningterror (approved):

lightningterror:
What are the steps to reproduce the issue?

Florin9doi:
Have a webcam connected and configured in usb configuration page and then boot EyeToy Kinetic twice

Florin9doi:
2 weeks to null a pointer ?

refractionpcsx2:
@jackun do you want to deal with this?

PR #3973, opened 2020-12-05
GSdx-GUI: Conservative buffer allocation, reordering and more

RedDevilus:
Conservative buffer allocation will be used as the default for GSdx.
Dithering gets formatting improvement where it switches places.
Monitoring while correct sounds vague even with tooltip, like RivaTuner Statistics server and stats/FPS counter which overlay your hardware usages. 

Maybe get some ideas if it fits or doesn't fit.


plugins/GSdx/GSdx.cpp
```diff
@@ -331,15 +331,15 @@ void GSdxApp::Init()
 	m_default_configuration["debug_glsl_shader"]                          = "0";
 	m_default_configuration["debug_opengl"]                               = "0";
 	m_default_configuration["disable_hw_gl_draw"]                         = "0";
-	m_default_configuration["dithering_ps2"]                              = "2";
+	m_default_configuration["dithering_ps2"]                              = "1";
```
lightningterror, line 334: It will be Scaled default now, you need to change it in Renderer code too.

plugins/GSdx/GSdx.rc
```diff
@@ -277,7 +277,7 @@ BEGIN
     GROUPBOX        "Hardware Renderer Settings",IDC_STATIC,6,121,230,140,BS_CENTER
     CONTROL         "Accurate DATE",IDC_ACCURATE_DATE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,11,132,63,10
     CONTROL         "Allow 8-Bit Textures",IDC_PALTEX,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,77,132,76,10
-    CONTROL         "Large Framebuffer",IDC_LARGE_FB,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,158,132,75,10
+    CONTROL         "CBA",IDC_LARGE_FB,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,158,132,75,10
```
lightningterror, line 280: Not a fan of the short name.

plugins/GSdx/GSdx.cpp
RedDevilus: I've reverted this to only use context menu.

plugins/GSdx/GSdx.rc
RedDevilus: I have changed it to Cons.Buffer unless you prefer something else?

tadanokojin: I prefer "Conservative Buffer Allocation" which is why I told you to change it to that :P 
Move the settings around if you have to.

I'd also prefer not to invert it by setting read but by setting write so as to retain the original setting semantics in the config.
In other words, when you got to save it in the gui, read the opposite value of the checkbox. If it's set write 0 and if it's unset write 1.

RedDevilus: Oh yeah this is a sin, not sure what i was thinking. I'll fix that write.

plugins/GSdx/GSdx.rc
```diff
@@ -269,38 +269,38 @@ BEGIN
...
-    COMBOBOX        IDC_CRC_LEVEL,104,211,127,118,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
-    LTEXT           "Blending Accuracy:",IDC_ACCURATE_BLEND_UNIT_TEXT,22,229,79,8
-    COMBOBOX        IDC_ACCURATE_BLEND_UNIT,104,227,127,118,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
-    COMBOBOX        IDC_ACCURATE_BLEND_UNIT_D3D11,104,227,127,118,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
-    CONTROL         "Enable HW Hacks",IDC_HACKS_ENABLED,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,11,245,71,10
-    PUSHBUTTON      "Advanced Settings and Hacks",IDC_HACKSBUTTON,104,243,127,14
+    GROUPBOX        "Hardware Renderer Settings",IDC_STATIC,6,114,230,154,BS_CENTER
+    CONTROL         "Accurate DATE",IDC_ACCURATE_DATE,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,11,124,63,10
+    CONTROL         "GPU Palette Conversion",IDC_PALTEX,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,11,140,85,10
+	CONTROL         "Conservative Buffer Allocation",IDC_LARGE_FB,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,104,140,105,10
```
lightningterror, line 280: Tab

RedDevilus:
Should i just squash to 1 commit or re-order it into correct split of commits?

RedDevilus:
@tadanokojin I forgot to append a commit, everything should work.

RedDevilus:
End results in GUI;

Tooltips revised and more clean this way
image

Ordering for dithering makes more sense as well:
image

Monitoring wasn't obvious that it displays the FPS counter (well not your actual computer stats, but the 'internal components' equivalent, except FPS which is not the real internal FPS, just how many frames the emulator itself displays):
image

RedDevilus:
I retested it on 100% DPI and it seems to work fine: 

image

GovanifY (approved):
LGTM, just requires squashing

plugins/GSdx/Renderers/HW/GSRendererHW.cpp
```diff
@@ -37,43 +37,48 @@ GSRendererHW::GSRendererHW(GSTextureCache* tc)
 	, m_userhacks_tcoffset_x(0)
 	, m_userhacks_tcoffset_y(0)
 	, m_channel_shuffle(false)
-	, m_lod(GSVector2i(0,0))
+	, m_lod(GSVector2i(0, 0))
 {
 	m_mipmap = theApp.GetConfigI("mipmap_hw");
 	m_upscale_multiplier = theApp.GetConfigI("upscale_multiplier");
-	m_large_framebuffer  = theApp.GetConfigB("large_framebuffer");
+	m_conservative_framebuffer = theApp.GetConfigB("conservative_framebuffer");
```
lightningterror, line 44: I don't think I see any other place where you inverted the option(or I haven't spotted it) so invert it here.

RedDevilus: It's changed from inverting to changing all the renderer code to not need inverting it in the code.

plugins/GSdx/Window/GSLinuxDialog.cpp
```diff
@@ -285,8 +285,8 @@ void populate_hw_table(GtkWidget* hw_table)
 {
 	GtkWidget* hack_enable_check = CreateCheckBox("Enable User Hacks", "UserHacks");
 	GtkWidget* acc_date_check = CreateCheckBox("Accurate DATE", "accurate_date");
-	GtkWidget* paltex_check = CreateCheckBox("Allow 8 bits textures", "paltex");
-	GtkWidget* large_fb_check = CreateCheckBox("Large Framebuffer", "large_framebuffer");
+	GtkWidget* paltex_check = CreateCheckBox("GPU Palette Conversion", "paltex");
+	GtkWidget* conservative_fb_check = CreateCheckBox("Conservative Buffer Allocation", "conservative_framebuffer"); // 2nd parameter large_framebuffer replaced with Conservative_Buffer_Allocation which is inverted on the renderer side.
```
GovanifY, line 289: This definitely needs to be fixed

GovanifY (changes requested):
Missed the inversion thingy the first time around

plugins/GSdx/Renderers/HW/GSRendererHW.cpp
```diff
@@ -125,8 +125,13 @@ void GSRendererHW::SetScaling()
...
 	// Until performance issue is properly fixed, let's keep an option to reduce the framebuffer size.
-	int fb_height = m_large_framebuffer ? 1280 :
-		(fb_width < 1024) ? std::max(512, crtc_size.y) : 1024;
+	//
+	// m_large_framebuffer has been inverted to m_conservative_framebuffer, it isn't an option that benefits being enabled all the time for everyone.
+	int fb_height = 1280;
+		if(m_conservative_framebuffer)
+	{
+	fb_height = fb_width < 1024 ? std::max(512, crtc_size.y) : 1024;
+	}
```
tadanokojin, line 134: formatting

RedDevilus: Yeah everything got unclanged

GovanifY (approved):

lightningterror:
I'll need to test it to make sure everything is good.

tadanokojin (approved):
quick test seems to work as expected (Windows)

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -191,14 +191,14 @@ const char* dialog_message(int ID, bool* updateText) {
...
-				"Note: It should be disabled for Armored Core, Destroy All Humans, Gran Turismo and possibly others.\n"
-				"This option does not improve the graphics or the FPS.";
-		// Windows only options.
+				   "\n\nThis hack replaces the partial invalidation with a complete deletion of the texture to reduce the CPU load.\n\nIt helps snowblind engine games.";
+		case IDC_CONSERVATIVE_FB:
+			return "Disabled: Reserves a larger framebuffer to prevent FMV flickers.\n"
+				   "Increases GPU/memory requirements.\n"
+				   "Disabling this can amplify stuttering due to low RAM/VRAM.\n\n"
+				   "Note: It should be enabled for Armored Core, Destroy All Humans, Gran Turismo and possibly others.\n"
+				   "This option does not improve the graphics or the FPS.";
```
lightningterror, line 200: Maybe last line isn't needed?

lightningterror (approved):

PR #3975, opened 2020-12-06
Switch yaml-cpp to default to the built-in version, and add an option to use a system version instead.

arcum42:
The text when using the system version stated it was preferable to use the built-in version to the system version, due to system versions being out of date, but you weren't able to use the built-in version if the system version was detected.

Additionally, it's entirely possible for it to pick up you have the 64-bit version of yaml-cpp installed when you are building pcsx2 for 32 bits, then have it error out because it can't link to it and don't have the 32 bit version installed, so using the built-in version simplifies things.

With this change, it'll use the built-in yaml-cpp unless you turn on the option "USE_SYSTEM_YAML" (or pass --use-system-yaml" when compiling with build.sh.)


GovanifY (approved):
LGTM, although we might want to pressure upstream into pushing a new release to avoid having to use the built in library

arcum42:
Even if they did a release, it's probably better to go built-in until 64 bit is the default. I was able to reproduce the install finding the 64 bit version on a 32 bit build, and erroring out on the install due to the mismatch of versions. I figure if you choose to use the system library, you can work out which one to install...

Issue #3631, opened 2020-08-25
[Feature Request] Add UI for configuring Hotkeys


Nicholas-Steel:
It would be greatly appreciated if we could get a UI for viewing/configuring hotkeys.

IngwiePhoenix:
Would love that too as well as being able to remap them to a controller button (i.e. fast-forward directly on the controller)

TomTurbine:
Will gladly give a +1 vote for this.

Setting up an emulationstation setup using a PS5 controller and trying to bind the Playstation button to exit back to the frontend.

Looking and can't find an option to close it.

lightningterror:
Done on Qt, won't be done on wx.

PR #3991, opened 2020-12-10
GUI: Keyboard Shortcuts Strings

### LINKED ISSUE #3212: Shift options
Recently I was looking to use the Slow motion option in Pcsx2, and could not remember how to enable it. Well with some googling I found the Forum thread about the "Shift" button options. It would be nice to see theses options listed in the program. i.e. a tooltip stating what key or key combo is associated with it.

Thanks for the Hard Work.

RedDevilus:
This will add extra hotkeys to the GUI. You can look up https://wiki.pcsx2.net/Hotkeys for all of them (not everything is added from there)

Fixes #3212 (Doesn't allow mapping in GUI yet, only from ini at this moment)
#3631 is for another PR
#3214 setting to none may also be possible in future PR

### DIFF
--- a/pcsx2/gui/GlobalCommands.cpp
+++ b/pcsx2/gui/GlobalCommands.cpp
@@ -390,7 +390,7 @@ namespace Implementations
 	void Sys_TakeSnapshot()
 	{
 		if (GSmakeSnapshot(g_Conf->Folders.Snapshots.ToUTF8()))
-		   OSDlog(ConsoleColors::Color_Black, true, "Snapshot taken");
+			OSDlog(ConsoleColors::Color_Black, true, "Snapshot taken");
 	}
 
 	void Sys_RenderToggle()
@@ -865,7 +865,6 @@ void AcceleratorDictionary::Map(const KeyAcceleratorCode& _acode, const char* se
 	KeyAcceleratorCode acode = _acode;
 	wxString overrideStr;
 	wxAcceleratorEntry codeParser; //Provides string parsing capabilities
-	wxString ye = GetUiKeysFilename();
 	wxFileConfig cfg(L"", L"", L"", GetUiKeysFilename(), wxCONFIG_USE_GLOBAL_FILE);
 	if (cfg.Read(wxString::FromUTF8(searchfor), &overrideStr))
 	{
@@ -1013,9 +1012,9 @@ void Pcsx2App::InitDefaultGlobalAccelerators()
 	GlobalAccels->Map(AAC(WXK_F4), "Framelimiter_MasterToggle");
 	GlobalAccels->Map(AAC(WXK_F4).Shift(), "Frameskip_Toggle");
 
-// Doesn't read from the ini file at this point because `AppConfig::GetUiKeysFilename` is blank at this point!
-// Used for custom hotkeys in the GUI.
-// It will read from the PCSX2_keys.ini in the ini folder based on PCSX2_keys.ini.default which get overridden
+	// Doesn't read from the ini file at this point because `AppConfig::GetUiKeysFilename` is blank at this point!
+	// Used for custom hotkeys in the GUI.
+	// It will read from the PCSX2_keys.ini in the ini folder based on PCSX2_keys.ini.default which get overridden
 	GlobalAccels->Map(AAC(WXK_TAB), "Framelimiter_TurboToggle");
 	GlobalAccels->Map(AAC(WXK_TAB).Shift(), "Framelimiter_SlomoToggle");
 
@@ -1024,12 +1023,12 @@ void Pcsx2App::InitDefaultGlobalAccelerators()
 
 	GlobalAccels->Map(AAC(WXK_ESCAPE), "Sys_SuspendResume");
 
-	GlobalAccels->Map(AAC( WXK_F8 ),"Sys_TakeSnapshot");
-	GlobalAccels->Map(AAC( WXK_F8 ).Shift(),	"Sys_TakeSnapshot");
-	GlobalAccels->Map(AAC( WXK_F8 ).Shift().Cmd(),"Sys_TakeSnapshot");
-	GlobalAccels->Map(AAC( WXK_F9 ), "Sys_RenderswitchToggle");
+	GlobalAccels->Map(AAC(WXK_F8), "Sys_TakeSnapshot");
+	GlobalAccels->Map(AAC(WXK_F8).Shift(), "Sys_TakeSnapshot");
+	GlobalAccels->Map(AAC(WXK_F8).Shift().Cmd(), "Sys_TakeSnapshot");
+	GlobalAccels->Map(AAC(WXK_F9), "Sys_RenderswitchToggle");
 
-//	GlobalAccels->Map(AAC(WXK_F10),	"Sys_LoggingToggle");
-//	GlobalAccels->Map(AAC(WXK_F11),	"Sys_FreezeGS");
-	GlobalAccels->Map(AAC(WXK_F12),	"Sys_RecordingToggle");
+	//	GlobalAccels->Map(AAC(WXK_F10),	"Sys_LoggingToggle");
+	//	GlobalAccels->Map(AAC(WXK_F11),	"Sys_FreezeGS");
+	GlobalAccels->Map(AAC(WXK_F12), "Sys_RecordingToggle");
 }
--- a/pcsx2/gui/GlobalCommands.cpp
+++ b/pcsx2/gui/GlobalCommands.cpp
@@ -865,6 +865,7 @@ void AcceleratorDictionary::Map(const KeyAcceleratorCode& _acode, const char* se
 	KeyAcceleratorCode acode = _acode;
 	wxString overrideStr;
 	wxAcceleratorEntry codeParser; //Provides string parsing capabilities
+	wxString ye = GetUiKeysFilename();
 	wxFileConfig cfg(L"", L"", L"", GetUiKeysFilename(), wxCONFIG_USE_GLOBAL_FILE);
 	if (cfg.Read(wxString::FromUTF8(searchfor), &overrideStr))
 	{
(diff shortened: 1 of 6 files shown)


xTVaser:
I'm not sure if these should be hardcoded into the strings, putting aside stylistic reasons (tabs look ugly in the diff).  I suspect the GUI will be inaccurate if the user customizes their keybindings.  It should probably lookup what key is bound to the action and convert that to it's string representation instead.

RedDevilus:
Yeah i made the mistake to append it to translable strings. I found the correct file for appending.
This is for F3 and F1
GlobalCommands.cpp
L997-1005

Changing it:
image

RedDevilus:
Additions:
Tab for Turbo
Shift + Tab for Slow Motion
Shift + F4 for Frame Skipping
F6 for Aspect Ratio
image
image

lightningterror:
You can do the same for Disable Frameskipping, you can remove the `Toggle in-game by pressing F4` sentence after.

RedDevilus:
I just noticed, i should push that change

lightningterror (approved):

avih:
What if you change the shortcuts via `PCSX2_keys.ini`?

RedDevilus:
It's too hardcoded and intertwined for custom hotkeys to pop-up and not breaking something along the way.

RedDevilus:
Got some help for Vaser for this, just struggling appending hotkeys without touching the translatable label. Emulation settings now has multiple custom hotkeys support variables. Hardcoded placeholder on the main window, i will add it to GS PR for the other custom ones.

RedDevilus:
Finally no hard-coded strings , bug got fixed , some other optimizations and doesn't break translations.
Just need to squash and give commits a better name.

Examples:
image
image
image
image
image

And it doesn't break current translation or future translations (i did add 2 changes for translations itself though)
Dutch:
image

French:
image

pcsx2/gui/GlobalCommands.cpp
```diff
@@ -1007,15 +1007,23 @@ void Pcsx2App::InitDefaultGlobalAccelerators()
 	GlobalAccels->Map(AAC(WXK_F4), "Framelimiter_MasterToggle");
 	GlobalAccels->Map(AAC(WXK_F4).Shift(), "Frameskip_Toggle");
 
-	/*GlobalAccels->Map( AAC( WXK_ESCAPE ),		"Sys_Suspend");
-	GlobalAccels->Map( AAC( WXK_F8 ),			"Sys_TakeSnapshot");
-	GlobalAccels->Map( AAC( WXK_F8 ).Shift(),	"Sys_TakeSnapshot");
-	GlobalAccels->Map( AAC( WXK_F8 ).Shift().Cmd(),"Sys_TakeSnapshot");
-	GlobalAccels->Map( AAC( WXK_F9 ),			"Sys_RenderswitchToggle");
+	// Doesn't read from the ini file at this point because `AppConfig::GetUiKeysFilename` is blank at this point!
+	// Used for custom hotkeys in the GUI.
```
xTVaser, line 1011: This portion of this comment is now inaccurate now that we have a better understanding of what's going on.

The problem is that at this point in the application initializing, it does not know if it's in portable or installed mode, and so it assumed installed mode, which i believe is the default.  https://github.com/PCSX2/pcsx2/blob/678829a5b2b8ca7a3e42d8edc9ab201bf00b0fe9/pcsx2/gui/AppInit.cpp#L479

This looks like it's been a long-standing issue, but not noticed because the majority of the keybindings are _actually_ set and used in `FrameForGS.cpp` by which point its figured things out.  This is something that should be addressed in the upcoming config work.  As a result, there may be some inconsistencies in the UI until that is fixed for users that use custom keybindings in portable mode.

I'd suggest changing this to something like:
```
// At this early stage of startup, the application assumes installed mode, so portable mode custom keybindings may present issues.
// Relevant - https://github.com/PCSX2/pcsx2/blob/678829a5b2b8ca7a3e42d8edc9ab201bf00b0fe9/pcsx2/gui/AppInit.cpp#L479
```

pcsx2/gui/MainFrame.cpp
```diff
@@ -402,9 +402,12 @@ void MainEmuFrame::CreatePcsx2Menu()
 								 wxEmptyString, wxITEM_CHECK);
 
 	m_menuSys.AppendSeparator();
-
-	m_menuSys.Append(MenuId_Sys_LoadStates, _("&Load state"), &m_LoadStatesSubmenu);
-	m_menuSys.Append(MenuId_Sys_SaveStates, _("&Save state"), &m_SaveStatesSubmenu);
+	//  Implement custom hotkeys (F3) with translatable string intact + not blank in GUI.
+	wxMenuItem* MainLoadStateLabel = m_menuSys.Append(MenuId_Sys_LoadStates, _("&Load state"), &m_LoadStatesSubmenu);
```
xTVaser, line 406: nit (and similar below): I'm not a fan of this variable naming.  The value of `MainLoadStateLabel` isn't really a label, it's a menu item.  Perhaps `sysLoadStateMenuItem` would be better?  I'm also not a fan of the uppercase on the first character, but I might be contradicting the styleguide here.

xTVaser (commented):
Looks mostly good to me, comments are just for style/documentation.  Long-term, it would be great if we could centralize the keybinding initialization to one place after determining portable/installed mode.  I can't think of an immediate reason why it needs to be done twice (once in the AppInit, once when the GS Frame is opened).

pcsx2/gui/Panels/VideoPanel.cpp
```diff
@@ -337,35 +338,35 @@ void Panels::VideoPanel::Defaults_Click(wxCommandEvent& evt)
...
 	{
-		m_span->ApplyConfigToGui( configToApply, true );
-		m_fpan->ApplyConfigToGui( configToApply, true );
+		m_span->ApplyConfigToGui(configToApply, true);
+		m_fpan->ApplyConfigToGui(configToApply, true);
 	}
 
 	Layout();
 }
-
```
lightningterror, line 371: Didn't need to remove this line.

pcsx2/gui/MainFrame.cpp
RedDevilus: Fair enough, i've also changed the wxMenuItems for Capture and Screenshot.

pcsx2/gui/Panels/VideoPanel.cpp
```diff
@@ -337,35 +338,36 @@ void Panels::VideoPanel::Defaults_Click(wxCommandEvent& evt)
...
+	if (flags & AppConfig::APPLY_FLAG_MANUALLY_PROPAGATE)
 	{
-		m_span->ApplyConfigToGui( configToApply, true );
-		m_fpan->ApplyConfigToGui( configToApply, true );
+		m_span->ApplyConfigToGui(configToApply, true);
+		m_fpan->ApplyConfigToGui(configToApply, true);
 	}
 
 	Layout();
-}
```
lightningterror, line 370: Extra lines before bracket, revert it as it originally was.

RedDevilus:
These are basically unselectable menu items (disabled by default, otherwise they could act like file...) used as labels for the shortcuts to cycling active saveslots. Only works in-game

Example:
image



Loading savestate:
image


Saving savestate:
image

pcsx2/gui/MainFrame.cpp
```diff
@@ -48,8 +49,14 @@ wxMenu* MainEmuFrame::MakeStatesSubMenu(int baseid, int loadBackupId) const
...
 	}
-
+	//  Implement custom hotkeys (F2) + (Shift + F2) with translatable string intact + not blank in GUI.
+	// baseid in the negatives will order in a different section, so if you want to increase more slots you can still easily do this, as -1 it will have the same function as opening file for savestates, which is bad
+	// For safety i also made them inactive aka grayed out to signify that's it's only for informational purposes, in the future this can still be expanded to cycle savestates in the GUI.
 	mnuSubstates->Append(baseid - 1, _("File..."));
+	wxMenuItem* CycleNext = mnuSubstates->Append(baseid - 2, _("Cycle to next slot") + wxString("  ") + fmt::format("({})", wxGetApp().GlobalAccels->findKeycodeWithCommandId("States_CycleSlotForward").toTitleizedString()));
+	CycleNext->Enable(false);
+	wxMenuItem* CycleBack = mnuSubstates->Append(baseid - 3, _("Cycle to previous slot") + wxString("  ") + fmt::format("({})", wxGetApp().GlobalAccels->findKeycodeWithCommandId("States_CycleSlotBackward").toTitleizedString()));
+	CycleBack->Enable(false);
```
xTVaser, line 59: yeah it would be nice to have these enable and actual commands, but perhaps a separate PR.
Here's some of the lines you'd be interested in:
https://github.com/PCSX2/pcsx2/blob/7b7278bc85fa3becd39c47168fe6c912523b4d43/pcsx2/gui/UpdateUI.cpp#L50-L51
https://github.com/PCSX2/pcsx2/blob/3c8d4029ddb601063bb011987824a879e29f7124/pcsx2/gui/Saveslots.cpp#L155-L165

pcsx2/gui/MainFrame.cpp
```diff
@@ -48,8 +49,14 @@ wxMenu* MainEmuFrame::MakeStatesSubMenu(int baseid, int loadBackupId) const
 		wxMenuItem* m = mnuSubstates->Append(loadBackupId, _("Backup"));
 		m->Enable(false);
 	}
-
+	//  Implement custom hotkeys (F2) + (Shift + F2) with translatable string intact + not blank in GUI.
```
xTVaser, line 52: Double space after `//` in a few spots like this, also you might want to leave out the specific keys from the comment just incase that changes in the future? Saying you are adding the hotkeys should be enough imo.

xTVaser (commented):
New menu options look nice!

pcsx2/gui/MainFrame.cpp
RedDevilus: The comments are used for placeholder and to easily find it in the code, they won't see it anyway.

pcsx2/gui/MainFrame.cpp
RedDevilus: I'll look into in the future, but i don't want to break anything for now especially with the current scope and size of this PR.

pcsx2/gui/Panels/VideoPanel.cpp
lightningterror, line 371: As I said before, leave the line after the bracket.

pcsx2/gui/MainFrame.cpp
```diff
@@ -803,20 +817,17 @@ void MainEmuFrame::CommitPreset_noTrigger()
 	g_Conf->EmuOptions.EnablePatches = menubar.IsChecked(MenuId_EnablePatches);
 }
 
-static void AppendShortcutToMenuOption(wxMenuItem& item, wxString keyCodeStr)
+void MainEmuFrame::AppendShortcutToMenuOption(wxMenuItem& item, wxString keyCodeStr)
 {
+	if (&item == nullptr)
+	{
+		return;
+	}
```
sonicfind, line 825: This return statement will never get hit due to how pass-by-reference variables of non-pointer types work.

RedDevilus: This PR went through various revisions, i think it's fine if i delete those 4 lines if i'm not mistaken. Have other stuff to do till end of the year, if it's a small fix i can do it still.

RedDevilus:
Changed the string for frameskipping, it didn't really fit in before.
image

lightningterror:
@sonicfind @xTVaser is the pr fine with you?

sonicfind:
My one complaint is basically just on structure. Due to needing to grab the string format for a command in so many places, it would be good to give it a designated function. Without it, I found it to look a bit cluttered.

Something simple like this would do:
```
wxString Pcsx2App::GetAssociatedKeyCode(const char* id)
{
	return GlobalAccels->findKeycodeWithCommandId(id).toTitleizedString();
}
```

Similar function for GSPanel would be nice to finally define too.

RedDevilus:
> 
> 
> My one complaint is basically just on structure. Due to needing to grab the string format for a command in so many places, it would be good to give it a designated function. Without it, I found it to look a bit cluttered.
> 
> Something simple like this would do:
> 
> ```
> wxString Pcsx2App::GetAssociatedKeyCode(const char* id)
> {
> 	return GlobalAccels->findKeycodeWithCommandId(id).toTitleizedString();
> }
> ```
> 
> Similar function for GSPanel would be nice to finally define too.

I was going to improve it in another PR, making cycle functions work in GUI and make it less cluttered by defining more functions. Luckily it still makes sense to a point and doesn't break anything , i agree it's a bit rough in readability.

sonicfind:
That's fair

RedDevilus:
I finally got my linux environment kinda working, some examples;
_Seems that linux doesn't like appending control string compared to windows_

Emulation Settings:
image

image

Capture settings:
image

xTVaser (commented):
On windows i see all keybinding entries where i expect them to be.  For now due to the weird complexity surrounding the `Sys_TakeSnapshot` event (how it automatically binds 2 modifier variants of itself), I think it should be omitted for now.

Havn't tested linux yet, about to.  But looks like you ran into no issues in linux _except_ for the aforementioned event.

RedDevilus:
Yeah i don't really get the reason why it's used like this, ease of use? A big variable which seperates into multiple?
Anyway, i disabled the hotkey for Snapshot/screen because it gives 2 different results for OS
Windows: Gives some long keybinding which would do GS dumps (not really useful) _Ctrl + Shift + F8_
Linux: Would nag about multiple duplicate mappings, not really clean. _Shift + F8_

This is now the temporary workaround which is the same as on master:
image

RedDevilus:
To check later in other PR;
F8 = Snapshot (single screenshot)
F8 + Shift = Snapshot + Single Frame dump (Linux used this on old code for hotkey)
F8 + Ctrl + Shift = Snapshot + Multi Frame Dump (Windows used this on old code for hotkey)

RedDevilus:
Got some help, GSdx hardcodes it, needs an overhaul on how it's handled.
image

xTVaser (approved):
Keybindings that i expect to see are on the linux side as well.  I also tested out the issue around Ctrl being displayed in linux, and i don't see an issue:
image
I think that was just a side-effect of the weird TakeSnapshot situation.

Looks good to me.

PR #3992, opened 2020-12-11
GUI: Resolve issue causing a crash when using `Close All Windows` while a game is running

### LINKED ISSUE #3996: [Regression] PCSX2 locks up when closing from the Taskbar
Starting with dev-385 (#3627), trying to close PCSX2 by right-clicking on the Taskbar in Windows 10 will lock up and crash under certain conditions.

**To Reproduce**
1. Make sure the option to "Combine Taskbar Buttons" is enabled:
grafik
(Right-Click on Taskbar --> Taskbar Settings)
2. Run PCSX2 and boot a Game.
3. Right-Click on PCSX2 and select "Close all Windows".
grafik
4. The Emu stops responding.

If Taskbar Buttons are split, PCSX2 will close as intended.
grafik
grafik

xTVaser:
Close #3996

### DIFF
--- a/pcsx2/gui/AppCorePlugins.cpp
+++ b/pcsx2/gui/AppCorePlugins.cpp
@@ -363,7 +363,8 @@ bool AppCorePlugins::OpenPlugin_GS()
 void AppCorePlugins::ClosePlugin_GS()
 {
 	_parent::ClosePlugin_GS();
-	if( GetMTGS().IsSelf() && GSopen2 ) sApp.CloseGsPanel();
+	if (CloseViewportWithPlugins && GetMTGS().IsSelf() && GSopen2)
+		sApp.CloseGsPanel();
 }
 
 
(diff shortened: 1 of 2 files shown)


lightningterror (approved):

tadanokojin (approved):

sonicfind:
While it does fix the crash, choosing the `pause` menu option nor pressing ESC will disable those select recording menu options.
As I already mentioned in the discord, my original intention was to disable the menu options from a single spot where all calls made to closing the GSwindow would end up. Now, in normal circumstances, this implementation worked fine as all the actions we would normally perform that close the window would apply on the main thread.
However, using the **Close All Windows** taskbar command forces the GS window to manually call `Pcsx2App::CloseGsPanel()` on the MTGS thread instead. This becomes a problem when you consider that `AppRpc_TryInvoke(&Pcsx2App::CloseGsPanel)` will use `SynchronousActionState::WaitForResult()` when anything but the main thread calls it.
```
bool Pcsx2App::AppRpc_TryInvoke( FnPtr_Pcsx2App method )
{
	if( wxThread::IsMain() ) return false;

	SynchronousActionState sync;
	PostEvent( Pcsx2AppMethodEvent( method, sync ) );
	sync.WaitForResult();

	return true;
}
```
Thus, we are forced to wait for a result that will never come - something proven true if you place a breakpoint on the `return true` before calling **Close All Windows**.

PR #3997, opened 2020-12-12
Add ghc filesystem. Remove wx variable types from config code. Rewrite Portable mode. Merge after #3928

kenshen112:
This is the start of a total rewrite of pcsx2 config. This removes much of WX from core and config alike standardizing all variable types to their filesystem and std counterparts where needed. This is a total rewrite of AppUserMode that would simplify the logic and apply the use of yaml-cpp. This is the start of AppConfig / Pcsx2Config rewrites that will come in future PR's


tadanokojin:
You have a bunch of style changes mixed in and it's all just a single giant commit.
You also apparently failed to resolve a merge conflict or something because you are reverting [this](https://github.com/PCSX2/pcsx2/commit/85c3ee823fa9d111e2039d6664bc5d63dc21672f) commit to name one.

kenshen112:
Understood. I'll fix the bad commit on my end and force all styling to the clang formatter

common/include/Utilities/PathUtils.h
```diff
@@ -0,0 +1,43 @@
+#pragma once
+#include <fstream>
+#include <iostream>
+#include <iomanip>
+#include "FixedPointTypes.inl"
+#include "filesystem.hpp"
```
tadanokojin, line 6: Inconsistent with the rest of the project

tadanokojin:
You also have GameIndex.yml from vaser's PR in here.

TheLastRar:
You're not using a submodule?

Is it preferable to include in this manor, or is using a submodule preferred?

I had PR to added it as a submodule in the works before I learned you where also planning on including it

kenshen112:
@TheLastRar I mean... It's a single file header? There's not much of a reason in my eyes to make a single file into a submodule.

xTVaser:
I'd advise rebasing ontop of my GameDB branch instead of master, that should cleanup all of that out of this PR, and should be fine considering you are saying for this to be held until #3928 merges anyway.
More commits would be helpful, maybe:
- one that captures the cmake changes
- one that captures the msvc changes
- one that implements the wxFileConfig
- one for portable mode.
- one that refactors all of the codebase to use the new globalconfig pointer and such

Even if these commits do not compile / work on their own right, it will make things more organized to review.

PR #3998, opened 2020-12-12
VU: Synchronise VU1, added speedhack for old behaviour

### LINKED ISSUE #1610: Parappa the Rapper 2 depth issues
Virtually all versions suffer from this, even the current master

Default settings and plugins. OGL + HW Depth fixes some issues but not completely, as it also creates other issues such as black noodles appearing in the top left corner

-Objects tend to get in front of the tap bar where a player must press the buttons to rap. This doesn't happen in real hardware.
-The colorful noodle effect when a player goes down to BAD or AWFUL is not emulated correctly. In OGL + HW the noodles are flat instead of wavy. This effect is only correctly emulated in software mode, but it's not perfect either, since black noodles tend to appear in the top left corner, same as in HW OGL + HW Depth. This is especially BAD near the end of the third stage, practically impassable without using SW mode or OGL with HW Depth not disabled (see youtube link for example)

To reproduce: play any level using Hardware mode. Software mode isn't perfect either.
Play badly if you want to see the missing noodle effect/semi working in SW mode.

Video showcasing the issues:
https://www.youtube.com/watch?v=WrxrKT2COmQ
PC Specs: Windows 10 Pro x64, Intel Pentium G3258, AMD Radeon R7 200 Graphics, 8GB RAM

refractionpcsx2:
This synchronises VU1 with the EE and VU0 (as close as we can).  A new speedhack has been added to replicate the old behaviour, which is on by default unless you're using the Safest preset.  Games which require the new behaviour have an entry in the GameDB to disable it.

Fixed Games:
Fixes SPS in Shadowman 2encond Coming
Fixes "Noodles" in Parappa the Rapper 2 (box in the top left)
Fixes floor corrupt textures and console error spam in Dropship
Slightly improves World series baseball 2k3 but it's still broken, just doesn't run at sub 1fps anymore

Broken Games:
Wakeboarding unleashed will only boot with MTVU enabled for now, however there's a hack in the code (disabled) which can fix the booting problem and fixes the waves in the game (with MTVU off and VU1 Instant off), will add it as a gamefix later.

Fixes #1610

### DIFF
--- a/bin/GameIndex.dbf
+++ b/bin/GameIndex.dbf
@@ -76,6 +76,7 @@
 -- Speed Hacks (SpeedHackName = <value>)
 ---------------------------------------------
 -- mvuFlagSpeedHack = 1 or 0 // Katamari Damacy have weird speed bug when this speed hack is enabled (and it is by default)
+-- InstantVU1SpeedHack = 1 or 0 // Games such as Parappa the Rapper 2 need VU1 to sync, so you can force disable the speedhack here
 
 ---------------------------------------------
 -- Memory Card Filter Override (MemCardFilter = s)
@@ -1391,6 +1392,7 @@ Serial = SCES-50408
 Name   = PaRappa the Rapper 2
 Region = PAL-M5
 Compat = 5
+InstantVU1SpeedHack = 0 // Fixes noodles.
 ---------------------------------------------
 Serial = SCES-50409
 Name   = MotoGP 2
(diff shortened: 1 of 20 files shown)


MrCK1:
Games that still require instant VU1:
The Suffering/Suffering Ties That Bind
Shox
Neversoft engine (Guitar Hero 3+, Tonky Hawk games)

refractionpcsx2:
> Games that still require instant VU1:
> The Suffering/Suffering Ties That Bind
> Shox
> Neversoft engine (Guitar Hero 3+, Tonky Hawk games)
> Mission Impossible Operation Surma

All fixed (or gamefixes applied)

atomic83GitHub:
Rolland Garros 2003 seems unafected unfortunately by this PR. And also, with VU1 sync superman returns internal framerate absolutely plummet.

PR #3999, opened 2020-12-12
Add GHC Filesystem 

kenshen112:
This is a PR to add GHC filesystem to pcsx2 common utilittes

### DIFF
--- a/3rdparty/include/ghc/filesystem.h
+++ b/3rdparty/include/ghc/filesystem.h
@@ -1,0 +1,5689 @@
+//---------------------------------------------------------------------------------------
+//
+// ghc::filesystem - A C++17-like filesystem implementation for C++11/C++14/C++17
+//
+//---------------------------------------------------------------------------------------
+//
+// Copyright (c) 2018, Steffen Schümann <s.schuemann@pobox.com>
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


xTVaser:
Thoughts on putting this into `3rdparty` instead as it is a 3rdparty dependency?

arcum42:
Yeah, given we didn't write it, 3rdparty would be a better location. While it is just one file, it's a third party file, and Utilities is mostly things we wrote. I can think of an exception or two, but then, maybe any exceptions should be moved, too...

kenshen112:
Roger

tellowkrinkle:
Thoughts on using a directory structure more like
```
3rdparty
└── filesystem
    ├── CMakeLists.txt
    ├── filesystem.vcxproj
    ├── filesystem.vcxproj.filters
    └── include
        └── ghc
            ├── filesystem.hpp
            └── fs_std.hpp
```

common/include/Utilities/filesystem.h
```diff
@@ -0,0 +1,5689 @@
+//---------------------------------------------------------------------------------------
+//
+// ghc::filesystem - A C++17-like filesystem implementation for C++11/C++14/C++17
+//
+//---------------------------------------------------------------------------------------
```
tadanokojin, line 5: duplicate file

pcsx2/CMakeLists.txt
```diff
@@ -945,7 +945,7 @@ set(pcsx2FinalSources
 set(pcsx2FinalLibs
     Utilities
     x86emitter
-    fmt::fmt
+	fmt::fmt
```
tadanokojin, line 948: Unnecessary formatting change

pcsx2/CMakeLists.txt
```diff
@@ -978,7 +978,8 @@ endif()
 # additonal include directories
 include_directories(
     gui
-    x86
+	x86
+	../3rdparty/include
```
tadanokojin, line 982: Unnecessary formatting change

TheLastRar:
What is the status of this pr?

xTVaser (changes requested):
Can we get these commits squashed perhaps as follows:
- deps: Add ghc::filesystem header
- vs: Add new 3rdparty include folder to project properties
- cmake: Configure new 3rdparty include folder

Other than cleaning up the commits I don't see a reason for this to be held up.  
- It's just adding a new library that is unused (yet)
- CI is building successfully so project config changes were successful.
- And it's already being used successfully here https://github.com/PCSX2/pcsx2/pull/4005 so the new library should be usable by whoever needs it once merged.

PR #4000, opened 2020-12-12
Clang-Format: Align with spaces

tellowkrinkle:
Switches to the new clang-format option [`tabs: AlignWithSpaces`](https://bugs.llvm.org/show_bug.cgi?id=38381)

Makes code look much nicer on GitHub (and for anyone who prefers 2-space tabs)

Note: Requires the fairly new clang-format 11

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -86,5 +86,5 @@ SpacesInParentheses: false
 SpacesInSquareBrackets: false
 Standard:        Cpp11
 TabWidth:        4
-UseTab:          ForContinuationAndIndentation
+UseTab:          AlignWithSpaces
 ...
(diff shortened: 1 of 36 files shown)


seta-san:
Won't spaces only align on fixed width fonts?

tellowkrinkle:
Yes, though I don't think it's possible to align on non-fixed-width fonts (or at least not on type of lines `AlignWithSpaces` affects, you can look at the diff to see what it changes)

tellowkrinkle:
BTW if we do want things to still look nice with non-fixed-width fonts, we may want to change the style guide to use constructs that don't require alignment.

e.g. instead of this:
```cpp
thing.method(long_thing_a, long_thing_b,
             long_thing_c, long_thing_d);
```
use this:
```cpp
thing.method(
	long_thing_a, long_thing_b,
	long_thing_c, long_thing_d);
```

Personally, I'm pretty neutral on this decision, and would be fine with both keeping the current system (use either one) and enforcing use of the second only.  The first method can use a lot of horizontal space in a way that doesn't play nice with soft wrap, so I would not be fine with enforcing use of it only

tellowkrinkle:
Closing as after further use I don't think this feature is well implemented in clang-format 11

PR #4007, opened 2020-12-16
core: fix speedhacks panel

tadanokojin:
attempt to fix the speedhacks panel


lightningterror (approved):
Works a lot better now than before.

RedDevilus (approved):
Best Christmas gift for the GUI lovers. Works great :)

tadanokojin:
I would have like to have sorted that last bug I had but I think this is just going to have to do for now.
QT when

PR #4008, opened 2020-12-16
cmake: Fix the install target.

orbea:
The install target was broken....


xTVaser (approved):
Good catch, i also updated the equivalent file path on the windows installer side. https://github.com/PCSX2/pcsx2/pull/4009/commits/c941a5c3e9dd0a07d06b1cd619fd93663a0d89e6  Didn't realize these paths were hard-coded.

xTVaser:
> Perhaps the CI should be testing `make install`.

It does https://github.com/PCSX2/pcsx2/blob/master/.github/workflows/scripts/linux/compile.sh#L15 so I guess that isn't sufficient.  Feel free to open a PR with an improvement if it's more obvious to you.

orbea:
You're right, but its missing `-DPACKAGE_MODE=ON` so that this install path is never tested.

PR #4019, opened 2020-12-18
Reorganize and realign InputRecordingControls

sonicfind:
Splits the checking of pause status and the actual pausing of emulation to separate functions called at different points during VSync. `VSyncEnd` is called once per frame, so checking for `g_FrameCount` increments is no longer required.

### DIFF
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -477,6 +477,15 @@ static __fi void frameLimit()
 
 static __fi void VSyncStart(u32 sCycle)
 {
+#ifndef DISABLE_RECORDING
+	if (g_Conf->EmuOptions.EnableRecordingTools)
+	{
+		// It is imperative that any frame locking that must happen occurs before Vsync is started
+		// Not doing so would sacrifice a frame of a savestate-based recording when loading any savestate
+		g_InputRecordingControls.HandleFrameCountLocking();
+	}
+#endif
+
 	frameLimit(); // limit FPS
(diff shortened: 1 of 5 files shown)


xTVaser (commented):
After a quick sanity check, I don't see any issues that I'd expect around frame-advancing and such.  However, seems like the VirtualPad is no longer rendering?  Perhaps this just needs a rebase ontop of current master, the commits are quite old.

Also, does this still allow for in the future, frame advancing more than 1 frame at a time?  Seems so with the use of the `frameLockTracker`, but just wondering if these changes will make features that require that behaviour more difficult to implement.

refractionpcsx2:
What's happening with this?

sonicfind:
> Also, does this still allow for in the future, frame advancing more than 1 frame at a time? Seems so with the use of the `frameLockTracker`, but just wondering if these changes will make features that require that behaviour more difficult to implement.

An easy way to do this would be to just have a frameCount value in InputRecordingControls, starting with the chosen # of frames, that decrements on each frame. If it hits 0, it pauses.
Could possibly use FrameCountTracker but I'd have to check whether that would possibly conflict with the framelock mechanism.

sonicfind:
> What's happening with this?
Just need to confirm that the VPad issue was truly dealt with. Should be good if that is the case.

xTVaser (approved):
Works great from what I can tell, made a recording and did a bunch of things.  VirtualPad worked fine this time.

PR #4021, opened 2020-12-18
DEV9: Cancel read of TAP device on suspend/shutdown

TheLastRar:
As above, only deals with the issue on windows.
winPcap (and thus Linux) needs to be looked at separately (and also isn't fixed on CLR_DEV9)

The implementation of the fix differs from what CLR_DEV9 does as I mostly used C#'s filestream, instead of direct Win32 calls.

This should speed up closing rx thread significantly.

I've only done minimal testing, just fyi

### DIFF
--- a/pcsx2/DEV9/Win32/tap-win32.cpp
+++ b/pcsx2/DEV9/Win32/tap-win32.cpp
@@ -271,6 +271,9 @@ TAPAdapter::TAPAdapter()
 	write.Offset = 0;
 	write.OffsetHigh = 0;
 	write.hEvent = CreateEvent(NULL, FALSE, FALSE, NULL);
+
+	cancel = CreateEvent(NULL, TRUE, FALSE, NULL);
+
 	isActive = true;
 }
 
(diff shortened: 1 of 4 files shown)


GovanifY (approved):
diff looks fine, requires testing

pcsx2/DEV9/Win32/tap-win32.cpp
```diff
@@ -298,9 +301,20 @@ bool TAPAdapter::recv(NetPacket* pkt)
...
-										 &read_size, FALSE);
+			HANDLE readHandles[]{read.hEvent, cancel};
+			const DWORD waitResult = WaitForMultipleObjects(2, readHandles, FALSE, INFINITE);
+
+			if (waitResult == WAIT_OBJECT_0 + 1)
+			{
+				CancelIo(htap);
+				//Wait for the I/O subsystem to acknowledge our cancellation
+				result = GetOverlappedResult(htap, &read,
+											 &read_size, TRUE);
```
lightningterror, line 312: Fit them in 1 line, same for below one.

TheLastRar: Okay

atomic83GitHub:
After some testing, I think this can be merged.

pcsx2/DEV9/Win32/tap-win32.cpp
```diff
@@ -298,9 +301,18 @@ bool TAPAdapter::recv(NetPacket* pkt)
...
+				CancelIo(htap);
+				//Wait for the I/O subsystem to acknowledge our cancellation
+				result = GetOverlappedResult(htap, &read, &read_size, TRUE);
+			}
+			else
+				result = GetOverlappedResult(htap, &read, &read_size, FALSE);
+
 			if (!result)
 			{
 			}
```
refractionpcsx2, line 318: Is there much point in an empty if statement?

pcsx2/DEV9/Win32/tap-win32.cpp
```diff
@@ -346,8 +358,8 @@ bool TAPAdapter::send(NetPacket* pkt)
 		if (dwError == ERROR_IO_PENDING)
 		{
 			WaitForSingleObject(write.hEvent, INFINITE);
-			result = GetOverlappedResult(htap, &write,
-										 &writen, FALSE);
+			result = GetOverlappedResult(htap, &write, &writen, FALSE);
+
 			if (!result)
 			{
 			}
```
refractionpcsx2, line 365: Another empty if statement, any need?

pcsx2/DEV9/Win32/tap-win32.cpp
TheLastRar: I guess not, I assume the idea was to have some kind-of error handling, but that if statement has been left that way since the plugin was added to PCSX2's repo many years ago.

May aswell just remove it this point.

PR #4033, opened 2020-12-21
Add option to throw memory-related exceptions from recompiler

Kingcom:
In some cases it can be useful to know when a game triggers invalid behavior, like unaligned memory access or TLB misses, that would crash a real system. This pull request adds a new option `Cpu.Recompiler.ThrowAddressExceptions` to achieve that. It can only be edited through the .ini file and is disabled by default.

TLB miss  exceptions were already thrown in developer builds, this is extended to when the new option is set. Loads and stores didn't check for unaligned access, so these checks are added. They are only written into the recompiled code if the option is enabled. Furthermore, as you explicitly opt-in to receive these exceptions, you definitely want to notice them - so any exceptions thrown due to the option are treated as immediately fatal.

### DIFF
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -259,6 +259,10 @@ struct Pcsx2Config
 				PreBlockCheckIOP:1;
 			bool
 				EnableEECache   :1;
+
+			bool
+				ThrowAddressExceptions : 1;
+
 		BITFIELD_END
 
 		RecompilerOptions();
--- a/pcsx2/Pcsx2Config.cpp
+++ b/pcsx2/Pcsx2Config.cpp
@@ -132,6 +132,8 @@ Pcsx2Config::RecompilerOptions::RecompilerOptions()
 	fpuOverflow	= true;
 	//fpuExtraOverflow = false;
 	//fpuFullMode = false;
+
+	ThrowAddressExceptions = false;
 }
 
 void Pcsx2Config::RecompilerOptions::ApplySanityCheck()
@@ -192,6 +194,8 @@ void Pcsx2Config::RecompilerOptions::LoadSave( IniInterface& ini )
 	IniBitBool( StackFrameChecks );
 	IniBitBool( PreBlockCheckEE );
 	IniBitBool( PreBlockCheckIOP );
+
+	IniBitBool(ThrowAddressExceptions);
 }
 
 Pcsx2Config::CpuOptions::CpuOptions()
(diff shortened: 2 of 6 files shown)


pcsx2/Config.h
```diff
@@ -259,6 +259,10 @@ struct Pcsx2Config
 				PreBlockCheckIOP:1;
 			bool
 				EnableEECache   :1;
+
+			bool
+				ThrowAddressExceptions :1;
+
```
RedDevilus, line 265: Whitespaces

pcsx2/Pcsx2Config.cpp
```diff
@@ -192,6 +194,8 @@ void Pcsx2Config::RecompilerOptions::LoadSave( IniInterface& ini )
 	IniBitBool( StackFrameChecks );
 	IniBitBool( PreBlockCheckEE );
 	IniBitBool( PreBlockCheckIOP );
+
```
RedDevilus, line 197: Whitespace here as well

tellowkrinkle:
Would you mind running `git clang-format HEAD~~` instead (which will only format your new changes)?  Clang format has issues with some of our macros (e.g. `jNO_DEFAULT` is a case and shouldn't be indented) so I'd rather not blindly run it on whole files right now

Kingcom:
Sure. Formatting the files was suggested in Discord due to the inconsistencies already inside them, but I can definitely see how it would be a problem. I've rebased any clang-format changes on my changes into the commits. All the new lines should conform now, regardless of what's around them.

pcsx2/x86/ix86-32/iR5900LoadStore.cpp
```diff
@@ -221,6 +292,21 @@ void recStore(u32 bits)
                 if (bits == 128)
                         xAND(arg1regd, ~0x0F);
 
+				if (EmuConfig.Cpu.Recompiler.ThrowAddressExceptions && bits != 128)
+				{
+					xTEST(arg1regd, alignment - 1);
+					xForwardJZ32 skipException;
```
tellowkrinkle, line 298: This should be able to be a J8

pcsx2/x86/ix86-32/iR5900LoadStore.cpp
```diff
@@ -221,6 +292,21 @@ void recStore(u32 bits)
...
+
+					// flush PC for the error message, preserve calculated address
+					xPUSH(arg1regd);
+					iFlushCall(FLUSH_PC);
+					xPOP(arg1regd);
```
tellowkrinkle, line 303: What's the register save/restore here for?

tellowkrinkle (commented):
I don't think you need to save registers around iFlushCall, which just writes things to memory (which on x86 doesn't need anything in registers)

lightningterror:
Status on this?

lightningterror:

PR #4036, opened 2020-12-22
Capture: Optimizations with file names, SPU2 recording, and capture menu gui

### LINKED ISSUE #3729: [Linux] screenrecorder option does nothing

sonicfind:
Addresses some oversights involving transferring a file name between the GS & SPU2 recording setups.
Adds a new audio toggle setting that allows a user to control whether a wav file should be generated alongside a video recording. In doing this, behavior for when SPU2 recording fails to be setup becomes more defined and consistent.
Additionally, F12 will now route to sMainFrame if gui is active as to keep the capture menu gui accurate to the current state. This change resolves #3729.

### DIFF
--- a/pcsx2/SPU2/Wavedump_wav.cpp
+++ b/pcsx2/SPU2/Wavedump_wav.cpp
@@ -110,7 +110,7 @@ bool WavRecordEnabled = false;
 static WavOutFile* m_wavrecord = nullptr;
 static Mutex WavRecordMutex;
 
-int RecordStart(const std::string* filename)
+bool RecordStart(const std::string* filename)
 {
 	try
 	{
@@ -121,16 +121,16 @@ int RecordStart(const std::string* filename)
 		else
 			m_wavrecord = new WavOutFile("audio_recording.wav", 48000, 16, 2);
 		WavRecordEnabled = true;
-		return 1;
+		return true;
 	}
 	catch (std::runtime_error&)
 	{
 		m_wavrecord = nullptr; // not needed, but what the heck. :)
 		if (filename)
-			SysMessage("SPU2-X couldn't open file for recording: %s.\nWavfile capture disabled.", filename->c_str());
+			SysMessage("SPU2 couldn't open file for recording: %s.\nWavfile capture disabled.", filename->c_str());
 		else
-			SysMessage("SPU2-X couldn't open file for recording: audio_recording.wav.\nWavfile capture disabled.");
-		return 0;
+			SysMessage("SPU2 couldn't open file for recording: audio_recording.wav.\nWavfile capture disabled.");
+		return false;
 	}
 }
 
--- a/pcsx2/SPU2/Wavedump_wav.cpp
+++ b/pcsx2/SPU2/Wavedump_wav.cpp
@@ -110,28 +110,27 @@ bool WavRecordEnabled = false;
 static WavOutFile* m_wavrecord = nullptr;
 static Mutex WavRecordMutex;
 
-void RecordStart(std::wstring* filename)
+int RecordStart(const std::string* filename)
 {
-	WavRecordEnabled = false;
-
 	try
 	{
 		ScopedLock lock(WavRecordMutex);
 		safe_delete(m_wavrecord);
-#ifdef _WIN32
 		if (filename)
-			m_wavrecord = new WavOutFile((*filename) + "wav", 48000, 16, 2);
+			m_wavrecord = new WavOutFile(filename->c_str(), 48000, 16, 2);
 		else
 			m_wavrecord = new WavOutFile("audio_recording.wav", 48000, 16, 2);
-#elif defined(__unix__)
-		m_wavrecord = new WavOutFile("audio_recording.wav", 48000, 16, 2);
-#endif
 		WavRecordEnabled = true;
+		return 1;
 	}
 	catch (std::runtime_error&)
 	{
 		m_wavrecord = nullptr; // not needed, but what the heck. :)
-		SysMessage("SPU2 couldn't open file for recording: %s.\nRecording to wavfile disabled.", "audio_recording.wav");
+		if (filename)
+			SysMessage("SPU2-X couldn't open file for recording: %s.\nWavfile capture disabled.", filename->c_str());
+		else
+			SysMessage("SPU2-X couldn't open file for recording: audio_recording.wav.\nWavfile capture disabled.");
+		return 0;
 	}
 }
 
--- a/pcsx2/gui/AppConfig.cpp
+++ b/pcsx2/gui/AppConfig.cpp
@@ -680,9 +680,9 @@ void AppConfig::LoadSave( IniInterface& ini )
 	GSWindow		.LoadSave( ini );
 	Framerate		.LoadSave( ini );
 #ifndef DISABLE_RECORDING
-	inputRecording	.loadSave( ini );
+	inputRecording.loadSave(ini);
 #endif
-	AudioCapture	.LoadSave( ini );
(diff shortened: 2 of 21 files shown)


pcsx2/SPU2/Wavedump_wav.cpp
```diff
@@ -110,28 +110,27 @@ bool WavRecordEnabled = false;
...
+		return 1;
 	}
 	catch (std::runtime_error&)
 	{
 		m_wavrecord = nullptr; // not needed, but what the heck. :)
-		SysMessage("SPU2 couldn't open file for recording: %s.\nRecording to wavfile disabled.", "audio_recording.wav");
+		if (filename)
+			SysMessage("SPU2-X couldn't open file for recording: %s.\nWavfile capture disabled.", filename->c_str());
+		else
+			SysMessage("SPU2-X couldn't open file for recording: audio_recording.wav.\nWavfile capture disabled.");
```
lightningterror, line 132: SPU2 is fine.

sonicfind: Oh yea, missed that

pcsx2/gui/AppConfig.cpp
```diff
@@ -674,8 +674,9 @@ void AppConfig::LoadSave( IniInterface& ini )
 	GSWindow		.LoadSave( ini );
 	Framerate		.LoadSave( ini );
 #ifndef DISABLE_RECORDING
-	inputRecording.loadSave(ini);
+	inputRecording	.loadSave( ini );
```
lightningterror, line 677: Formatting on this line was correct, it would be better to run clang on the entire file tbh.

pcsx2/gui/AppConfig.h
```diff
@@ -285,6 +285,15 @@ class AppConfig
 #endif
 	};
 
+	struct CaptureOptions
+	{
+		bool					EnableAudio;
```
lightningterror, line 290: Too many tabs between type and name.

plugins/GSdx/GSCapture.cpp
```diff
@@ -438,11 +438,11 @@ std::wstring* GSCapture::BeginCapture(float fps, GSVector2i recommendedResolutio
 		else
 		{
 			dlg.InvalidFile();
-			return nullptr;
+			return 0;
 		}
 	}
 
-	std::wstring fn{dlg.m_filename.begin(), dlg.m_filename.end()};
+	;
```
lightningterror, line 445: Leftover.

plugins/GSdx/GSCapture.cpp
```diff
@@ -525,7 +528,8 @@ std::wstring* GSCapture::BeginCapture(float fps, GSVector2i recommendedResolutio
 	}
 
 	m_capturing = true;
-	return new std::wstring();
+	filename = m_out_dir + "/audio_recording.wav";
+	return 1;
```
lightningterror, line 532: I don't think I saw any other values besides 0 and 1 so maybe change the function to bool instead?

pcsx2/gui/AppConfig.cpp
sonicfind: I'll just revert that change for now. We can do clang later.

xTVaser:
Does this PR now contain the fixes that were done here - https://github.com/PCSX2/pcsx2/pull/3732 ?

sonicfind:
It contains those fixes, yes. I also took your advice of splitting PRs to make them more singularly-defined and took out the Input recording stuff. Once this PR is merged, I'll then bring that back.

xTVaser (changes requested):
The app consistently hits an exception when pressing F12, that isn't hit when navigating from the GUI menu.  Though I can continue and set up the capture.

Once recording, I seem to hit an exceptions again, sometimes recovering, sometimes not.  I also can't seem to find the audio output.

Worth mentioning that once I try stopping/capturing again after the initial buggy capture, it seems to go a lot smoother.  I suppose some sort of first-time setup problems?  Still no audio to be found though.

sonicfind:
Ok now that I've come back to this:
That exception you're hitting is a problem with master. I'll have a check to see where it originated from, but it's not something that should be handled in this PR as it wasn't initially a problem when it first started.
Additionally, the exception is triggered before even hitting any of the code I changed, so yea...

xTVaser:
Oh interesting, I'll double check that, but I believe you / wouldn't be surprised.  Perhaps due to the SPU2 merge?

xTVaser (changes requested):
Confirmed the F12 exception is present on master.

Can still record fine after recent rebase.  But found a new issue:
- Start recording with "Include Audio" checked,
- Before stopping, uncheck "Include Audio"
- Stop recording

Audio recording will not stop and PCSX2 will continue dumping audio to the file forever.  Some ideas:
- Either disable the menu option while recording
- when its unchecked audio dumping should stop immediately.
- or when stopping recording, it should check to see if audio is dumping even if "include audio" is `false`

Not sure which is easiest or best, but seems undesirable.

sonicfind:
Shoot, must've forgotten to add the menu disabler for that option on record. My b

xTVaser (approved):
Looks good to me, latest changes with disabling the menu option prevents me from getting things into a bad state.

plugins/GSdx/GSCapture.cpp
```diff
@@ -456,32 +454,34 @@ std::wstring* GSCapture::BeginCapture(float fps, GSVector2i recommendedResolutio
 	if(FAILED(hr = m_graph.CoCreateInstance(CLSID_FilterGraph))
 	|| FAILED(hr = cgb.CoCreateInstance(CLSID_CaptureGraphBuilder2))
 	|| FAILED(hr = cgb->SetFiltergraph(m_graph))
-	|| FAILED(hr = cgb->SetOutputFileName(&MEDIASUBTYPE_Avi, fn.c_str(), &mux, NULL)))
+	|| FAILED(hr = cgb->SetOutputFileName(&MEDIASUBTYPE_Avi,
```
lightningterror, line 457: Maybe fit it in 1 line.

sonicfind:
I noticed that back in February that there was an effort to implement unicode wstring stuff. With that in mind, should I try to switch capture stuff to wstring as well? It can be considered in this scope.

xTVaser:
> I noticed that back in January that there was an effort to implement unicode wstring stuff into GS. With that in mind, should I try to switch capture stuff to wstring as well? It can be considered in this scope.

Hmm yes, if you use a path with unicode characters, it struggles on windows:
image

But I don't know if you want to do this in this PR?  I've learned that we don't just want to use `wstring` across the board everywhere, I suspect the current `std::string` works fine on linux, but we need `wstring` for windows unicode support.  That might be a bigger deal?  I'm not familiar with what was attempted back in January, was it abandoned for a reason?

Everything thing else continues to work as I'd expect though

sonicfind:
I was just planning on changing the `m_filename` variable of the GSCaptureDlg class to wstring. It's only used in a windows configuration to begin with, so doing so should be fine.

sonicfind:
Actually, that does present an issue. Since GS & SPU are suppose to share a filename (just with different extensions), I would need to share wstring throughout all these capture functions on windows without disturbing linux. Hmm, which may mean making two versions of the GSCapture functions.

xTVaser:
Personally I'd probably just do a follow-up PR that just adds the function versions as it's not like this is introducing a new issue (i think).  But if you want to do it here, don't amend it with your previous commits, if it's a separate commit it should be easy to re-approve.

sonicfind:
After looking over what would be necessary to make it work, a separate PR will be the best way to go, yea.

PR #4039, opened 2020-12-23
Use armips as submodule

Kingcom:
Updated (and very overdue) version of #1902.

What this does is remove the crippled assembler I put into the debugger many years ago and instead replace it with an armips submodule, similar to what was done in PPSSPP. The assemble feature so far is greatly limited and doesn't even support basic macros like li. Using armips directly, almost its entire feature set can be used with little extra work.


Kingcom:
Hm, I'd have expected at least the Windows builds to pass...

```
 Build started 12/23/2020 7:26:57 AM.
     1>Project "D:\a\pcsx2\pcsx2\buildbot.xml" on node 1 (ReleaseAll target(s)).
     1>Project "D:\a\pcsx2\pcsx2\buildbot.xml" (1) is building "D:\a\pcsx2\pcsx2\PCSX2_suite.sln" (2) on node 1 (Clean target(s)).
     2>D:\a\pcsx2\pcsx2\PCSX2_suite.sln(2,1): error MSB4025: The project file could not be loaded. Data at the root level is invalid. Line 2, position 1.
     2>Done Building Project "D:\a\pcsx2\pcsx2\PCSX2_suite.sln" (Clean target(s)) -- FAILED.
     1>Done Building Project "D:\a\pcsx2\pcsx2\buildbot.xml" (ReleaseAll target(s)) -- FAILED.

Build FAILED.

       "D:\a\pcsx2\pcsx2\buildbot.xml" (ReleaseAll target) (1) ->
       "D:\a\pcsx2\pcsx2\PCSX2_suite.sln" (Clean target) (2) ->
         D:\a\pcsx2\pcsx2\PCSX2_suite.sln(2,1): error MSB4025: The project file could not be loaded. Data at the root level is invalid. Line 2, position 1.
```
Not very helpful error messages, unfortunately...

PCSX2_suite.sln
```diff
@@ -73,6 +73,7 @@ EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "libsamplerate", "3rdparty\libsamplerate\libsamplerate.vcxproj", "{47AFDBEF-F15F-4BC0-B436-5BE443C3F80F}"
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "yaml", "3rdparty\yaml-cpp\yaml.vcxproj", "{48329442-E41B-4A1F-8364-36EEE1B71343}"
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "libarmips", "3rdparty\armips\libarmips.vcxproj", "{129E5E2B-39C1-4D84-96FE-DFD22DBB4A25}"
```
TheLastRar, line 76: I think you are missing an `EndProject` after the yaml project entry

Kingcom: Ah, thanks! I wonder why my local installation didn't complain about that.

Kingcom:
Phew, that was interesting... but looks like it builds everywhere now. I had to duplicate the vcxproj for armips due to configuration differences.

GovanifY (approved):
went ahead and rebased, should be good to go!

pcsx2/CMakeLists.txt
```diff
@@ -1112,6 +1111,7 @@ include_directories(
     gui
     x86
     ${CMAKE_BINARY_DIR}/pcsx2/gui
+    ${CMAKE_SOURCE_DIR}/3rdparty/armips
```
tellowkrinkle, line 1114: Have you tested without this?  If armips did things correctly it shouldn't be needed.

Kingcom: Unfortunately, the latest release version (which is what I referenced here) does not do it correctly. I've rewritten the whole CMake project for the master branch though, so everything is good there now. I could update to use that instead, too.

tellowkrinkle: That would be my preference

Kingcom: Okay. I'll update it to the current master branch. That will require C++17 and bring in a recursive submodule (https://github.com/gulrak/filesystem, as std::filesystem isn't available on all platforms - std::filesystem _can_ be used if it is available though), is that a problem?

Kingcom: I've updated it now. The VS solution uses std::filesystem, as Microsoft supports it well. On the CMake side it's off by default, but can be enabled through an option. Most recent Linux distros should have no problem it, but you need quite a recent MacOS version to get native support.

F0bes:
Tested before the gs-merge conflicts are resolved; Everything works great.
`li` and `lui` are now recognized as valid instructions.
`li v0, 0x123456` properly gets assembled to 
`lui v0,0x0012`
`addiu v0,0x3456`

Once the conflicts are resolved it should be good to go!

F0bes (approved):

pcsx2/DebugTools/MipsAssembler.h
```diff
@@ -1,5 +1,5 @@
 /*  PCSX2 - PS2 Emulator for PCs
- *  Copyright (C) 2014-2014  PCSX2 Dev Team
+ *  Copyright (C) 2014-2017  PCSX2 Dev Team
```
F0bes, line 2: Copyright year

F0bes (changes requested):

jensdraht1999:
Just asking, why this is not merged? Everything seems to work as far as I can see.

F0bes:
For one there are merge conflicts and secondly it doesn't follow how we implement 3rd party projects.

Kingcom:
That wasn't mentioned in any of the reviews. Is there any documentation on how dependencies should be used?

This is mutually incompatible with #4926, however. I don't see a good way to support that behavior with armips, so a decision for one or the other would have to be made.

lightningterror:
Closing as it has been decided to keep https://github.com/PCSX2/pcsx2/pull/4926 and it's already been merged.

PR #4047, opened 2020-12-26
Properly cancel an input recording on a bad initial savestate

sonicfind:
One of the problems with the current implementation was that it never accounted for a failed savestate load. If you were to load a recording that was based on a savestate that no longer works (corrupted or outdated), the emulator would not know to cancel loading the recording. This now accounts for that by forcing most input recording elements to wait until confirmation.


xTVaser (approved):
Pulled this down and tested it:
- Attempting to load an old recording that used a now invalid save-state was rejected, saw no ill-effects in UI
  - Was able to start a new recording after this, everything worked as expected
  - Was also able to play a functional recording after this, also worked as expected.

I wish there was a more explicit way to know if the savestate loaded or not.  Maybe in the future loading a save-state can be refactored to a proper async/await system for parts of the codebase that need this kind of functionality.  At a _minimum_ the savestate could atleast be verified and immediately return false.

Good catch, looks good to me.

PR #4048, opened 2020-12-26
gamedb: Remove unused compareStrNoCase function, convert crc and serial strings to uppercase for consistency.

lightningterror:


pcsx2/GameDatabase.cpp
```diff
@@ -163,7 +155,7 @@ GameDatabaseSchema::GameEntry YamlGameDatabaseImpl::entryFromYaml(const std::str
 		{
 			for (const auto& entry : patches)
 			{
-				std::string crc = strToLower(entry.first.as<std::string>());
+				std::string crc = strToUpper(entry.first.as<std::string>());
```
xTVaser, line 158: This will be a breaking change, the CRC's are generated internally (which get used to check the GameDB) are lowercase, so it won't be able to find any crc specific patches.  This why is the old GameDB compared without case sensitivity and why I left it alone for now.  I think the better solution is:
- Find where the CRC's are generated, and force them to be uppercase.
- When validating the YAML file, reject serials or CRCs that are not uppercased as well.

xTVaser (changes requested):

pcsx2/GameDatabase.cpp
lightningterror: Ah, that makes sense, I'll just leave it for now then.

PR #4052, opened 2020-12-28
GameDB: Cleanup comments

RedDevilus:
Adding punctation, removing duplicate serial and making english language more happy for reading.

### DIFF
--- a/bin/GameIndex.yaml
+++ b/bin/GameIndex.yaml
@@ -766,7 +766,7 @@ SCAJ-20159:
   name: "Soul Calibur III"
   region: "NTSC-Ch-E-J"
   gameFixes:
-    - EETimingHack # Fixes bad colours on charcter select when in Progressive Scan
+    - EETimingHack # Fixes bad colours on character select when in Progressive Scan.
 SCAJ-20160:
   name: "Yoshitsuneki"
   region: "NTSC-Unk"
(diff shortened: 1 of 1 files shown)


bin/GameIndex.yaml
```diff
@@ -12676,9 +12677,6 @@ SLES-53446:
   name: "Arcade USA"
   region: "PAL-E"
   compat: 5
-SLES-53450:
```
lightningterror, line 12679: Why was this removed?

refractionpcsx2: Yeah that's going to break the parsing of the file.

RedDevilus: Sces-53450 is the correct serial. Sles-53450 is wrong. http://redump.org/disc/12328/

lightningterror: @Mrlinkwii Can you confirm with Red what the correct serial is?

Mrlinkwii: done

refractionpcsx2: okay, I see the entry is already in there, Do you want to correct the title for the entry that's already in there, Red?

RedDevilus: I'll do a bigger PR later and include that as well.

PR #4058, opened 2020-12-29
recording: Update the GS Frame's freq. properly on enabling/disabling input recording

xTVaser:
Prior to this, if you enabled input recording after already having the GS window open, the title bar would continue to use it's slow default until you restarted the entire emulator.

This fixes that, and also should put the frequency back down when disabled.

To test this, you don't need to even do any recording, the speed difference should be obvious.  These steps should cover most cases:
- Open Game > Enable Recording Tools > Title should update quickly.
  - Disable Recording Tools, title should go back to being slow
- Restart Emu > Disable Recording Tools > Open Game > Title should update slowly
- Restart Emu > Enable Recording Tools > Open Game > Title should update quickly

### DIFF
--- a/pcsx2/gui/FrameForGS.cpp
+++ b/pcsx2/gui/FrameForGS.cpp
@@ -589,8 +589,7 @@ bool GSFrame::ShowFullScreen(bool show, bool updateConfig)
 	return retval;
 }
 
-
-void GSFrame::CoreThread_OnResumed()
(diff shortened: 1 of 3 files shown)


xTVaser:
Touching some GS stuff, fyi @tadanokojin

tadanokojin (approved):

lightningterror (approved):
Seems to update properly now.

PR #4059, opened 2020-12-30
DEV9: Generate unique MAC for TAP

TheLastRar:
A unique MAC is needed for full compatibility with XLink Kai.

Uses a method similar to what was done for pcap.

### DIFF
--- a/pcsx2/DEV9/net.cpp
+++ b/pcsx2/DEV9/net.cpp
@@ -82,3 +82,21 @@ void TermNet()
 		nif = nullptr;
 	}
 }
+
+NetAdapter::NetAdapter()
+{
+	//Ensure eeprom matches our default 
+	SetMACAddress(ps2MAC);
+}
+
+void NetAdapter::SetMACAddress(u8* mac)
+{
+	if (ps2MAC != mac)
+		memcpy(ps2MAC, mac, 6);
+
+	for (int i = 0; i < 3; i++)
+		dev9.eeprom[i] = ((u16*)mac)[i];
+
+	//The checksum seems to be all the values of the mac added up in 16bit chunks
+	dev9.eeprom[3] = (dev9.eeprom[0] + dev9.eeprom[1] + dev9.eeprom[2]) & 0xffff;
+}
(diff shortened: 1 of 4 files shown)


pcsx2/DEV9/pcap_io.cpp
```diff
@@ -345,13 +329,31 @@ char* pcap_io_get_dev_desc(int num)
...
+	newMAC.bytes[4] = hostMAC.bytes[5];
+
+	SetMACAddress((u8*)&newMAC);
+
+	if (pcap_io_init(config.Eth, newMAC) == -1)
 	{
 		SysMessage("Can't open Device '%s'\n", config.Eth);
 	}
+#else
+	SysMessage("pcap not supported on windows\n");
```
GovanifY, line 355: Might want to use Console.Error

GovanifY (commented):
LGTM apart from a single comment

pcsx2/DEV9/net.cpp
```diff
@@ -82,3 +82,21 @@ void TermNet()
...
+
+NetAdapter::NetAdapter()
+{
+	//Ensure eeprom matches our default 
+	SetMACAddress(ps2MAC);
+}
+
+void NetAdapter::SetMACAddress(u8* mac)
+{
+	if (ps2MAC != mac)
```
Warepire, line 94: This will always be false. If you want to compare the actual contents of the mac's it's better to use std::array or memcmp. With preference to std::array.

TheLastRar: The intention is to check if the two are referring to the same region of memory (same ptr) (to avoid copying to itself)

The default constructor for NetAdapter (just above) calls SetMACAddress() with ps2MAC, this is done to ensure that the MAC address in eeprom matches what is stored in ps2MAC (as it actually isn't). Admiralty this is a little awkward

pcsx2/DEV9/pcap_io.cpp
TheLastRar: Changed

pcsx2/DEV9/net.cpp
TheLastRar: I've tweaked it to be a little less confusing as to what I was trying to do here

PR #4060, opened 2020-12-30
GUI Improvements around savestate-based input recordings

sonicfind:
* Originally, a user would have to navigate through a few menus to load up the file dialog and then manually select to load the savestate. To rectify this, a new `GoTofirstFrame()` function (along with the accompanying command code `Shift + L`) has been added. Calling this will automatically search for and attempt to load the base savestate. This command will be tied to the Boot option with a recording active and will also change its label to "Load savestate" when appropriate.
* If an attempt to explicitly load the base savestate fails due to it not being located, a file dialog will prompt the user for a new one. Whatever savestate gets chosen will be saved as the new base for the rest of the session. **Note: changing the savestate mid session will NOT change the starting frame count.**
* Choosing the shutdown option with **any** recording active, regardless of start type, will close said recording - but only after confirming the action in a safety prompt.
* Swapping the current source or source type with a savestate recording will add a line of warning text to the prompt that normally appears. It's just to make sure users are aware of possible savestate incompatibilities.

### DIFF
--- a/pcsx2/gui/GlobalCommands.cpp
+++ b/pcsx2/gui/GlobalCommands.cpp
@@ -545,10 +545,8 @@ namespace Implementations
 	{
 		if (g_Conf->EmuOptions.EnableRecordingTools && g_InputRecording.IsActive())
 		{
-			if (!g_InputRecording.GoToFirstFrame())
-			{
-				sMainFrame.StopInputRecording();
-			}
+			// Assumes that gui is active, as you can't access recording options without it
+			g_InputRecording.GoToFirstFrame(GetMainFramePtr());
 		}
 	}
 
--- a/pcsx2/gui/GlobalCommands.cpp
+++ b/pcsx2/gui/GlobalCommands.cpp
@@ -541,6 +541,17 @@ namespace Implementations
 		}
 	}
 
+	void GoToFirstFrame()
+	{
+		if (g_Conf->EmuOptions.EnableRecordingTools && g_InputRecording.IsActive())
+		{
+			if (!g_InputRecording.GoToFirstFrame())
+			{
+				sMainFrame.StopInputRecording();
+			}
+		}
+	}
+
 	void States_SaveSlot(int slot)
 	{
 		States_SetCurrentSlot(slot);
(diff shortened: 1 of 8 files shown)


pcsx2/gui/GlobalCommands.cpp
```diff
@@ -832,8 +841,10 @@ static const GlobalCommandDescriptor CommandDeclarations[] =
 
 #ifndef DISABLE_RECORDING
 		{"FrameAdvance", Implementations::FrameAdvance, NULL, NULL, false},
-		{"TogglePause", Implementations::TogglePause, NULL, NULL, false},
+		{"TogglePause" , Implementations::TogglePause,  NULL, NULL, false},
```
lightningterror, line 844: Might want to undo the formatting changes on this line.

pcsx2/gui/MainMenuClicks.cpp
```diff
@@ -173,14 +173,18 @@ wxWindowID SwapOrReset_Iso(wxWindow* owner, IScopedCoreThread& core_control, con
...
 
-		result = pxIssueConfirmation(dialog, MsgButtons().Reset().Cancel().Custom(_("Swap Disc"), "swap"));
-		if (result == wxID_CANCEL)
+#ifndef DISABLE_RECORDING
+		if (g_InputRecording.IsActive() && g_InputRecording.GetInputRecordingData().FromSaveState())
+			dialog += dialog.Text(_("\n(Warning: The savestate accompanying the active input recording\nmay not be compatible with the new source)"));
+#endif
+
+		result = pxIssueConfirmation( dialog, MsgButtons().Reset().Cancel().Custom(_("Swap Disc"), "swap"));
+		if( result == wxID_CANCEL )
```
lightningterror, line 187: Formatting.

sonicfind: Clarification on this one?
Referencing the empty line or something else?

sonicfind: ohhhhhhh

lightningterror (approved):
I think it's good now, need someone to test it.

xTVaser (approved):
Pulled this down and built it, didn't encounter any issues.
- New warning on the ISO swapping prompt only appeared when a recording was active
- New Shift+L Feature worked as expected inside a SS recording and power-on recording
- If i hid the original save-state, i was able to load a temporary one and it did NOT persist to the recording file (reloading the recording used the original after i unhid it)  which is what I'd expect.

Suggestions for changes discussed in discord, they should be small wording changes / an .ini update so nothing that can't be easily approved without having to formally test again.

xTVaser (approved):
Hard to tell because of the rebase, but I believe the only changes since my test last night were wording changes in the UI / adding to the default keybindings file.

PR #4061, opened 2020-12-30
Documentation: Add script to generate PDF documentation from markdown via pandoc

xTVaser:
Used this for a decent PDF template - https://github.com/Wandmalfarbe/pandoc-latex-template

Could have done more with it like made title pages for each PDF, but thought that might have been annoying for users (a blank page to always scroll past). I didn't do a whole lot of formatting or overrode the default templates settings so if there are issues with the generated PDFs call them out and i can try to fix them up.

I included source links to the github repo in each PDF, my rationale being if a user found a problem they could easily PR a fix.  Though this can also be removed if not deemed useful.

Pandoc also has language support, so perhaps that is something that could be looked into in the future for atleast the main configuration guide https://github.com/Wandmalfarbe/pandoc-latex-template#changing-the-document-language though I'm not sure how exactly it works, may not even be useful in the way i think it would be.


lightningterror:
Is it ready?

xTVaser:
from my perspective yes, but it would be nice to know if people like the PDF format.  I'm not sure if anyone has looked at them yet.

RedDevilus:
Some examples on my end:

image

image

image

Besides a bit much whitespace on start and end of the files, which i don't really mind.

They look good overall as result.

RedDevilus:
Even the bookmarks work as intended:

image

RedDevilus (approved):

tadanokojin:
I like it

tadanokojin (approved):

tadanokojin:
> I included source links to the github repo in each PDF, my rationale being if a user found a problem they could easily PR a fix. Though this can also be removed if not deemed useful.

I noticed that you use 1.6.0 branch, so gameindex.md will not exist.

Mrlinkwii:
looking great with the PDF format  
image

PR #4066, opened 2020-12-31
IPU: Return contents of the bottom of the FIFO in CMD except FDEC/VDEC

refractionpcsx2:
This is based on the work by @PSI-Rockin on Dobiestation but modified based on my theory that it always returns the first 32bits of the FIFO unless an FDEC/VDEC command is executed, however FDEC is basically doing this anyway, but VDEC returns decoded information.

Fixed videos for probably all broken EA games, but namely:
F1 2001
F1 2002
Neo Contra
Shox
Theme Park Rollercoaster/World

Unrelated PR but a note for myself when removing video skip patches for games which are no longer broken:
Simple 2000 Series Vol.59 - The Uchyujin to Hanashi Sou!

### DIFF
--- a/bin/GameIndex.yaml
+++ b/bin/GameIndex.yaml
@@ -6872,18 +6872,6 @@ SLES-50423:
   name: "F1 2001"
   region: "PAL-M4"
   compat: 5
-  patches:
-    063FF7DB:
-      content: |-
-        author=prafull
-        // Skips IPU videos.
-        // Each videos can be called separately
-        // on this game.
-        // This seems like the only way of patching
-        // them all. The issue here is the PCSX2 IPU
-        // do not support well the EA proprietary IPU
-        // Library causing hanging in menus.
-        patch=1,EE,001f4d2c,word,00000000
 SLES-50424:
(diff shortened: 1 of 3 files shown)


bin/GameIndex.yaml
```diff
@@ -34907,8 +34859,6 @@ SLUS-20961:
     08901101:
       content: |-
         comment=Patch by Prafull
-        // New patch fixes most issues.
-        patch=0,EE,003a94f8,word,00000000
         // Fix hang at Level 4.
```
atomic83GitHub, line 34862: Are the others lines of patch necessary?

refractionpcsx2: Yes, I confirmed with Prafull the other lines are still required

atomic83GitHub: Is the cause of the issue known?

refractionpcsx2: From my brief look at the debugger, it's generating a NaN float in the EE FPU and looping around comparing it.

PR #4073, opened 2020-12-31
3rdparty:PortAudio: Upgrade PortAudio to latest master

refractionpcsx2:
Our PortAudio was 4 years old and they've done a bunch of work since then so I figured I'd upgrade it.


lightningterror (approved):
Testers reported no issues so I guess it's good to go?

PR #4074, opened 2020-12-31
SPU2: Attempt 96khz sample rate if 48khz fails (WASAPI)

refractionpcsx2:
WASAPI is terrible and doesn't support sample rate conversion in shared (normal) mode and requires exclusive mode.

This PR basically attempts to double the PS2 sample rate of 48khz to 96khz as this tends to be the other windows default.  If this fails it will put a log message to say you need to use Exclusive mode.

Note: When using PS1 game emulation you will need to use exclusive mode, nothing I can do about this (44.1khz doesn't divide nicely from 48khz or 96khz)

### DIFF
--- a/pcsx2/SPU2/Mixer.cpp
+++ b/pcsx2/SPU2/Mixer.cpp
@@ -893,6 +893,9 @@ __forceinline
 
 	SndBuffer::Write(Out);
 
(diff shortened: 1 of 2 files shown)


RedDevilus (approved):
Fixes my (cheapish :P) USB headset not playing on default PortAudio Settings.

PR #4075, opened 2021-01-01
Wayland support

### LINKED ISSUE #1629: Wayland Support
**_PCSX2 version:**_ All when compiled with GTK3

**_Plugins used:**_
GSDX HW/SW

**_Description of the issue:**_
PCSX2 crashes when compiled with --gtk3 and run on Wayland Session. Adding --egl (which is necessary on Wayland) doesn't help.

```

Gdk-WARNING **: gdkwindow-x11.c:5554 drawable is not a native X11 window
Current Renderer: OpenGL (Hardware mode)
EGL: Failed to get a window surface
Wrong X11 display/window
GSopen Failed: return code: 0xffffffff

```

There are two possible ways of fixing the issue:
*Making app request X11 backend (which is workaround more than fix)
*Fixing the app :)

First option is just adding "GDK_BACKEND=x11 " before executing the PCSX2 itself.

**_How to reproduce the issue:**_
Compile latest PCSX2 with --gtk3
Run PCSX2 on Wayland Session (Fedora 25 default)

### LINKED ISSUE #3903: GS plugin crash -- "EGL: failed to get a window surface"
***PCSX2 version:***
``v1.7.0`` daily build from 05/11/2020.

PCSX crashes with the notification: _"GS plugin failed to open"_.
By googling around the most common fix is  to change the version of the GS plugin (AVX, SSE, etc.), but none of them worked in my case. The weirdest thing is that PCSX2 v1.7.0 used to work before updating (had an older build version from +- 6 months ago).
Looking at the log file, it says: _"EGL: failed to get a window surface"_, and then closes all plugins and crashes.

I guess this is pretty machine specific? I am on Ubuntu (with OpenBox, no stock configuration).

I know I had a 1.7.0-git version working, but cannot remember when exactly I got it from the PPA.
I got it about 6 months ago I think, but not quite sure (sorry this is vague, I am a user and not a developer).
Right now I downgraded to using the normal version of the PPA (as opposed to daily), and that one is version 1.4.0 and works.

I don't know what changed, but 1.4.0 does not seem as fluid as 1.7.0 was, so I would love to be able to go back to using the daily builds! :)

***PC specifications:***
CPU: Intel i7-7700HQ (8) @ 3.800GHz  
GPU: NVIDIA GeForce GTX 1050 Mobile  
OS: Ubuntu 18.04.5 LTS x86_64  
Host: XPS 15 9560  
Kernel: 4.15.0-122-generic  
WM: Openbox  

Thanks for your awesome work with this project!!

bmwalters:
- Fixes #1629
- Fixes #3903 

These patches add support for running PCSX2 natively under Wayland on Linux. The launcher script no longer forces `GDK_BACKEND=x11`.

The main changes are as follows (also see the individual commits):
- Implemented the previously-stubbed `GSWndEGL_Wayland`
   - The bulk of the changes are for supporting the deprecated `GSopen` (GSopen1) interface which creates a standalone window.
   - The implementation of `AttachNativeWindow` expects a `wl_egl_window` to be passed to GS from the GUI. This is created in `GSFrame` in order to keep a reference to the `wl_egl_window` around to support resizing it in the wx `OnResize`.
- Updated PAD/Linux to only call into X11 libraries when GS is using an X11 window.

I have successfully used these changes when running the Sway compositor, with rendering, resizing, keyboard input, and gamepad input working as expected.


cmake/SearchForStuff.cmake
```diff
@@ -154,6 +157,11 @@ if(UNIX)
         include_directories(${GTK2_INCLUDE_DIRS})
 	endif()
 
+	if(WAYLAND_FOUND)
```
GovanifY, line 160: && WAYLAND_PROTOCOLS_FOUND maybe?

common/include/wayland-gsdisplay.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
```
GovanifY, line 1: Why in common/? We won't reuse it outside of the displaying system, will we?

pcsx2/PAD/Linux/PAD.h
```diff
@@ -32,12 +32,28 @@
 #include <windowsx.h>
 
 #else
-/*
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>
 #include <X11/keysym.h>
-*/
+// x11 is dumb like that
+#undef DisableScreenSaver
```
GovanifY, line 39: Why move from keyboard to the general header?

pcsx2/PAD/Linux/PAD.h
```diff
@@ -32,12 +32,28 @@
...
+#ifndef __APPLE__
+#include "wayland-gsdisplay.h"
+
+struct GSDisplayHandle
+{
+	bool is_wayland;
+
+	union
+	{
+		struct { Display* display; Window window; } x11;
```
GovanifY, line 50: Maybe try not to have it as a one-liner?

pcsx2/gui/AppMain.cpp
```diff
@@ -305,7 +308,7 @@ void Pcsx2App::PadKeyDispatch( const keyEvent& ev )
 	// outside of the window so the app isn't aware of the current key state.
 	m_kevt.m_shiftDown = wxGetKeyState(WXK_SHIFT);
 	m_kevt.m_controlDown = wxGetKeyState(WXK_CONTROL);
-	m_kevt.m_altDown = wxGetKeyState(WXK_MENU) || wxGetKeyState(WXK_ALT);
```
GovanifY, line 308: Why remove menu?

pcsx2/gui/AppMain.cpp
```diff
@@ -1008,21 +1011,42 @@ void Pcsx2App::OpenGsPanel()
...
+	{
+		pDsp[0] = (uptr)gsFrame->GetPluginDisplayPropertiesWaylandEGL();
+		pDsp[1] = (uptr)NULL;
+	}
+	else
+#endif
+#ifdef GDK_WINDOWING_X11
+	if (GDK_IS_X11_WINDOW(draw_window))
+	{
+		#if GTK_MAJOR_VERSION < 3
```
GovanifY, line 1029: indentation

pcsx2/gui/AppMain.cpp
```diff
@@ -1008,21 +1011,42 @@ void Pcsx2App::OpenGsPanel()
...
+	else
+#endif
+	{
+		pxAssertDev(false, "Unknown GDK display type. Can't initialize GS Plugin.");
+	}
 #else
 	pDsp[0] = (uptr)gsFrame->GetViewport()->GetHandle();
-	pDsp[1] = NULL;
-#endif
+	pDsp[1] = (uptr)NULL;
```
GovanifY, line 1046: nullptr?

pcsx2/gui/AppMain.cpp
```diff
@@ -1008,21 +1011,42 @@ void Pcsx2App::OpenGsPanel()
...
+	// so that the viewport's wl_surface is created before we access it.
+	gsFrame->ShowFullScreen( g_Conf->GSWindow.IsFullscreen );
 
-	pDsp[0] = (uptr)XDisplay;
-	pDsp[1] = (uptr)Xwindow;
+#ifdef GDK_WINDOWING_WAYLAND
+	if (GDK_IS_WAYLAND_WINDOW(draw_window))
+	{
+		pDsp[0] = (uptr)gsFrame->GetPluginDisplayPropertiesWaylandEGL();
+		pDsp[1] = (uptr)NULL;
```
GovanifY, line 1022: nullptr here too

plugins/GSdx/GS.cpp
```diff
@@ -269,7 +269,7 @@ static int _GSopen(void** dsp, const char* title, GSRendererType renderer, int t
 			int w = theApp.GetConfigI("ModeWidth");
 			int h = theApp.GetConfigI("ModeHeight");
 #if defined(__unix__)
-			void *win_handle = (void*)((uptr*)(dsp)+1);
+			void *win_handle = dsp;
```
GovanifY, line 272: I'm fine with the change, but doesn't that add a warning?

pcsx2/gui/FrameForGS.cpp
```diff
@@ -839,3 +856,136 @@ void GSFrame::OnResize( wxSizeEvent& evt )
 	// if we skip, the panel is auto-sized to fit our window anyway, which we do not want!
 	//evt.Skip();
 }
+
+#if defined(__WXGTK__) && defined(GDK_WINDOWING_WAYLAND)
+
+struct NecessaryWaylandGlobals
```
GovanifY, line 862: is Necessary necessary? Pun intended

pcsx2/gui/FrameForGS.cpp
```diff
@@ -839,3 +856,136 @@ void GSFrame::OnResize( wxSizeEvent& evt )
...
+	props_wl->display = display;
+	props_wl->egl_window = egl_window;
+	props_wl->surface = child_surface;
+	props_wl->subsurface = subsurface;
+	return props_wl;
+}
+
+void GSFrame::DestroyPluginDisplayPropertiesWayland(PluginDisplayPropertiesWayland* props_wl)
+{
+	if (props_wl == nullptr) return;
```
GovanifY, line 975: no one-liners

plugins/GSdx/Window/GSWndEGL.cpp
```diff
@@ -333,10 +331,18 @@ void *GSWndEGL_X11::CreateNativeWindow(int w, int h)
...
+void GSWndEGL_X11::AttachNativeWindow(void *handle, void **out_native_display, void **out_native_window)
 {
-	m_NativeWindow = *(Window*)handle;
-	return handle;
+	uptr *window_ptr = (uptr*)handle + 1;
+	m_NativeWindow = (Window)*window_ptr;
+
+	// We only need a native display when we manage the window ourself.
+	// By default, EGL will create its own native display. This way the driver knows
+	// that display will be thread safe and so it can enable multithread optimization.
```
GovanifY, line 341: This comment was related to the scope of CreateNativeDisplay and doesn't make much more sense anymore

pcsx2/gui/FrameForGS.cpp
```diff
@@ -839,3 +856,136 @@ void GSFrame::OnResize( wxSizeEvent& evt )
...
+	if (strcmp(interface, wl_compositor_interface.name) == 0) {
+		globals->compositor = (wl_compositor *)wl_registry_bind(registry, name, &wl_compositor_interface, 4);
+	}
+	if (strcmp(interface, wl_subcompositor_interface.name) == 0) {
+		globals->subcompositor = (wl_subcompositor *)wl_registry_bind(registry, name, &wl_subcompositor_interface, 1);
+	}
+}
+
+static void gs_wl_registry_remove_global(void *data, wl_registry *registry, uint32_t name)
+{}
```
GovanifY, line 881: please add a linefeed between curly brackets

plugins/GSdx/Window/GSWndEGL.cpp
```diff
@@ -41,19 +41,24 @@ int GSWndEGL::SelectPlatform()
 		return 0;
 	}
 
+	// Inspect the provided display handle to check if it is Wayland or X11.
+	// The Wayland GS display handle happens to set pDsp[1] to NULL, so check it.
+	bool is_wayland_display = display_handle == nullptr
+		|| ((void **)display_handle)[1] == nullptr;
```
GovanifY, line 47: Why not use a struct there? Would be more appropriate

plugins/GSdx/Window/GSWndEGL.cpp
```diff
@@ -171,10 +176,8 @@ bool GSWndEGL::Attach(void* handle, bool managed)
 {
 	m_managed = managed;
 
-	m_native_window = AttachNativeWindow(handle);
-
+	AttachNativeWindow(handle, &m_native_display, &m_native_window);
 	OpenEGLDisplay();
-
```
GovanifY, line 177: Why the linefeed removal? display vs init separation

plugins/GSdx/Window/GSWndEGL.cpp
```diff
@@ -205,10 +208,10 @@ bool GSWndEGL::Create(const std::string& title, int w, int h)
 
 	m_managed = true;
 
+	m_native_display = CreateNativeDisplay();
 	OpenEGLDisplay();
 
 	m_native_window = CreateNativeWindow(w, h);
-
```
GovanifY, line 211: same here

GovanifY (changes requested):
Hi, sway user there, thanks a lot for the PR!

I did not find any big issues with your sway implementation, mostly superfluous stylistic and implementation changes.
Light testing also proved to work well.

Thanks for the work!!

cmake/SearchForStuff.cmake
bmwalters: `WAYLAND_INCLUDE_DIRS` and `WAYLAND_DEFINITIONS` are actually unrelated to `wayland-protocols`.

The only current usage of `wayland-protocols` is in the implementation of the old (deprecated?) `GSopen` (aka GSopen1) API in which the GS plugin creates its own native window instead of taking over a window given to it by core. It uses the xdg-shell protocol from `wayland-protocols` to handle setting the window title, and as currently unimplemented, resizing. (I focused my development efforts on the non-deprecated `GSopen2` API).

So I think that we should only condition on `wayland-protocols` in the cmake code that uses it directly (only the `WAYLAND_ADD_PROTOCOL_CLIENT` macro's usage(s)).

I'm a cmake noob so I don't know how to get the best error message in case a dev tries to build without having `wayland-protocols`. Would adding an `if(WAYLAND_PROTOCOLS_FOUND)` here be the best way to do it? Or would it generate hard-to-understand error messages since the `.c` and `.h` protocol files would not be generated?
https://github.com/PCSX2/pcsx2/pull/4075/files#diff-273e7a9b69994f653862c746a10ed357c053b6807756d7878de4c06879f2a1adR225

common/include/wayland-gsdisplay.h
bmwalters: It's referenced by `gui` as you mention and also once by `PAD` (since `PAD` has per-display-system input polling). It will also be used by GSdx once it's merged (the struct is currently duplicated into GSdx code).

pcsx2/PAD/Linux/PAD.h
bmwalters: The types are used in this file now in order to define the `GSDisplayHandle` struct.

pcsx2/gui/AppMain.cpp
bmwalters: See commit c5f73ff53b7dce68dd28525e7d417bd540e14376 for context. Since we're checking for alt here I don't think this affects users?

pcsx2/gui/AppMain.cpp
bmwalters: NULL and nullptr both require a cast to `uptr` since `uptr` is an integer size type (size of a pointer) and not a pointer type.

pcsx2/gui/AppMain.cpp
bmwalters: NULL and nullptr both require a cast to `uptr` since `uptr` is an integer size type (size of a pointer) and not a pointer type.

plugins/GSdx/GS.cpp
bmwalters: The previous code was casting `dsp` in order to perform pointer arithmetic on it (getting the second element of the two-element dynamically typed array pointer that is `pDsp`). Now we want the "first element" or the array pointer itself, so we just store `dsp` and no casting is required.

plugins/GSdx/Window/GSWndEGL.cpp
bmwalters: I used a struct for this in `PAD` but didn't want to duplicate the code into GSdx since it'll be merged. After the GSdx merge a ton of indirection can be removed since we won't need to marshal window handles through the `pDsp` pointer anymore.

plugins/GSdx/Window/GSWndEGL.cpp
bmwalters: reasoning is dependencies, you need to set `m_native_display` before you can call `OpenEGLDisplay`, then you need to set `m_native_window` before you can call `FullContextInit`.

In the Attach case the `AttachNativeWindow` assigns both of those variables so I squished all three calls next to each other to signify that they're closely related and not separate steps I guess.

Let me know if you still prefer the newlines.

plugins/GSdx/Window/GSWndEGL.cpp
bmwalters: I still think it's relevant to explain why we use `nullptr` here when we have a perfectly valid X `Display` passed to us through `pDsp` (aka `handle`).

bmwalters (commented):
Thanks for the review!

One other comment I wanted to make is that you'll notice resizing isn't frame-perfect right now (game content clips outside the GS window while dragging to resize). This will become a lot easier to fix once GSdx is merged.

Explanation under https://ppaalanen.blogspot.com/2013/11/sub-surfaces-now.html heading "Independent application sub-modules" explains the coordination between surface and subsurface owners to get frame-perfect resize.

common/include/wayland-gsdisplay.h
GovanifY: PAD is currently a part of the core as much as GUI is. common is used when libraries are used in plugins && core && utilities or any combination of it. Eventually this will be a part of the frontend part of the code but that's a bit far ahead, so you can move it into gui for now

pcsx2/PAD/Linux/PAD.h
GovanifY: Ah right, my bad

plugins/GSdx/Window/GSWndEGL.cpp
GovanifY: You can always duplicate the struct for now and I'll remove it when I'll do the merge, this way I don't end up forgetting about it and casting structures

plugins/GSdx/Window/GSWndEGL.cpp
GovanifY: Right, makes sense, maybe not the best place to have it but I won't bother you with that

pcsx2/gui/AppMain.cpp
GovanifY: Yeah but I meant use a nullptr instead of NULL when possible, type safety etc etc

pcsx2/gui/AppMain.cpp
GovanifY: Probably isn't, I'll accept that until someone bugs me about it

pcsx2/gui/AppMain.cpp
GovanifY: same as https://github.com/PCSX2/pcsx2/pull/4075#discussion_r552731633

plugins/GSdx/GS.cpp
GovanifY: Makes sense, thanks!

cmake/SearchForStuff.cmake
GovanifY: your "FindX" stuff should simply error out if a dependency isn't found, and you'll add a check required dep function in the general cmake file.

GSOpen1 will eventually be completely removed so supporting it isn't necessarily necessary, indeed. I don't think we even use it anymore.

plugins/GSdx/Window/GSWndEGL.cpp
GovanifY: Fine by me

cmake/SearchForStuff.cmake
tadanokojin: We do use GSOpen1 for GSDumpGUI

sjnewbury:
There seems to be a problem with scaling in High DPI (2x), at least in Sway.

PR #4084, opened 2021-01-02
Update mscompile.cmd to build with VC2019

iwilliamson:
The code no longer compiles with VC2015 anyway.  You only actually need "4aa702d", but I'm a bit of a git neophyte.


mscompile.cmd
```diff
@@ -48,6 +48,7 @@ ECHO Using:
 cl 2>&1 | findstr "Version"
 ECHO.
 
+SET Platform=
```
refractionpcsx2, line 51: Is this required?

lightningterror (approved):

Issue #883, opened 2015-10-07
Burnout 3/Revenge/Dominator - Sun Glow Can Be Seen Through Surfaces


escape209:
This is a known issue, but I'm opening this issue here so that if there is any progress on it being fixed, then it can be posted here and eventually (hopefully) marked as solved, in addition to the fact that it hasn't been mentioned on the issue tracker yet.

Anyway, this bug seems to be a core bug, as it appears in all of GSdx's software and hardware renderers, as well as in ZeroGS. It's a pretty easy issue to understand - you can see the sun through walls, when it should disappear when your view of it is completely (not partially) obscured by an object. The bug happens in Burnout 3: Takedown, Burnout Revenge and Burnout Dominator, of which the versions for all regions are affected (basically, PAL and NTSC.)

No combination of settings in the Emulation settings window, nor the GSdx settings window has been able to fix the bug (including Synchronized MTGS.)

FirstRoboticsGuy999:
This isn't just a Burnout issue. It happens in games like Gran Turismo 4 as well. It probably has something to do with alpha sorting or that fact that the PS2 doesn't have depth buffer support.

I found a great write-up by Matt Godbolt that explains some of the PS2's rendering limitations. http://xania.org/201003/swat-ps2-renderer

gregory38:
PS2 has depth support. It misses stencil / color swizzle / shader ... Well it just misses everything.

There is still a bug in SW when doing read/write to a texture so it could still be a GSdx bug. But it could also be a core bug.

DPersonalized:
It seems like it's not the sun that's showing through walls and it's actually just its glow effect. Setting Skipdraw to 28 will remove the sun's glow, meaning it won't show through walls, but it will appear as just a yellow circle in the sky:

Skipdraw 0
![gsdx_20160110131542](https://cloud.githubusercontent.com/assets/8225831/12221792/205cb830-b7a1-11e5-87e9-e708fa9a4dcc.PNG)

Skipdraw 28
![gsdx_20160110131710](https://cloud.githubusercontent.com/assets/8225831/12221799/4615e146-b7a1-11e5-9b55-031cdd36a17d.png)

willkuer:
28?

I never heard such a large value. For 27 you can still see the glow?
Thanks for the hint. This makes support much easier.

DPersonalized:
The glow is indeed **still** there at 27 :scream:

FlatOutPS2:
[...]

PR #4094, opened 2021-01-02
GS-hw: fix Burnout games black sky.

iMineLink:
Added an OI hack to perform some point list (CLUT) and sprite (~~CLUT,~~ sky tex) draws via CPU directly in GS local memory, tested only on Burnout 3.
Solves the black sky issue also when TC is cleared.
Update: added on OO hack to download the cloud texture from GPU after rendering during level loading, which solves the missing clouds issue. Tested only in Burnout 3 PAL (50Hz and 60Hz), might need adjustments for different games ~~/regions~~ .
~~Update 2: added a CU hack to avoid upscaling of the clouds drawn during level loading.~~
Update 3: Fixed a bug in TC which assigned a wrong scale factor to native resolution draws.
Update 4: Fixed a bug in the HW renderer which provided too small RT to fit the current draw (upscaled clouds).
Update 5: Reworked and refactored the GS device surface fetching code to always clear the fetched rt or ds.
Update 6: Do not linearly interpolate data when downloading from GPU. Greatly improves clouds rendering with even (x2, x4...) upscaling factors.
~~Sometimes the clouds work, sometimes not.~~
~~Possibly this introduces issues when loading screen is shown multiple times, for example when starting a race, exiting, and starting again.~~
~~Sometimes this intereferes with the skipdraw used to remove the display of the unsupported texture shuffle effect.~~
Testing needed and welcome!

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
@@ -863,12 +863,14 @@ void GSRendererHW::SwSpriteRender()
 
 	const GSVector4i r = m_r;
 
+#ifndef NDEBUG
 	const int tw = 1 << m_context->TEX0.TW;
 	const int th = 1 << m_context->TEX0.TH;
 	const float meas_tw = m_vt.m_max.t.x - m_vt.m_min.t.x;
 	const float meas_th = m_vt.m_max.t.y - m_vt.m_min.t.y;
 	ASSERT(!PRIM->TME || (abs(meas_tw - r.width()) <= SSR_UV_TOLERANCE && abs(meas_th - r.height()) <= SSR_UV_TOLERANCE)); // No input texture min/mag, if any.
 	ASSERT(!PRIM->TME || (abs(m_vt.m_min.t.x) <= SSR_UV_TOLERANCE && abs(m_vt.m_min.t.y) <= SSR_UV_TOLERANCE && abs(meas_tw - tw) <= SSR_UV_TOLERANCE && abs(meas_th - th) <= SSR_UV_TOLERANCE)); // No texture UV wrap, if any.
+#endif
 
 	GIFRegTRXPOS trxpos = {};
 
--- a/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
@@ -17,8 +17,6 @@
 #include "GSRendererHW.h"
 #include "GS/GSGL.h"
 
-const float GSRendererHW::SSR_UV_TOLERANCE = 1e-3f;
-
 GSRendererHW::GSRendererHW()
 	: m_width(default_rt_size.x)
 	, m_height(default_rt_size.y)
@@ -863,36 +861,30 @@ void GSRendererHW::SwSpriteRender()
 
 	const bool texture_mapping_enabled = PRIM->TME;
 
-	// Setup registers for SW rendering
-	GIFRegBITBLTBUF bitbltbuf = {};
+	const GSVector4i r = m_r;
 
-	if (texture_mapping_enabled)
-	{
-		bitbltbuf.SBP = m_context->TEX0.TBP0;
-		bitbltbuf.SBW = m_context->TEX0.TBW;
-		bitbltbuf.SPSM = m_context->TEX0.PSM;
-	}
-
-	bitbltbuf.DBP = m_context->FRAME.Block();
-	bitbltbuf.DBW = m_context->FRAME.FBW;
-	bitbltbuf.DPSM = m_context->FRAME.PSM;
-
-	ASSERT(m_r.x == 0 && m_r.y == 0); // No rendering region offset
-	ASSERT(!PRIM->TME || (abs(m_vt.m_min.t.x) <= SSR_UV_TOLERANCE && abs(m_vt.m_min.t.y) <= SSR_UV_TOLERANCE)); // No input texture offset, if any
-	ASSERT(!PRIM->TME || (abs(m_vt.m_max.t.x - m_r.z) <= SSR_UV_TOLERANCE && abs(m_vt.m_max.t.y - m_r.w) <= SSR_UV_TOLERANCE)); // No input texture min/mag, if any
-	ASSERT(!PRIM->TME || (m_vt.m_max.t.x <= (1 << m_context->TEX0.TW) && m_vt.m_max.t.y <= (1 << m_context->TEX0.TH))); // No texture UV wrap, if any
+	const int tw = 1 << m_context->TEX0.TW;
+	const int th = 1 << m_context->TEX0.TH;
+	const float meas_tw = m_vt.m_max.t.x - m_vt.m_min.t.x;
+	const float meas_th = m_vt.m_max.t.y - m_vt.m_min.t.y;
+	ASSERT(!PRIM->TME || (abs(meas_tw - r.width()) <= SSR_UV_TOLERANCE && abs(meas_th - r.height()) <= SSR_UV_TOLERANCE)); // No input texture min/mag, if any.
+	ASSERT(!PRIM->TME || (abs(m_vt.m_min.t.x) <= SSR_UV_TOLERANCE && abs(m_vt.m_min.t.y) <= SSR_UV_TOLERANCE && abs(meas_tw - tw) <= SSR_UV_TOLERANCE && abs(meas_th - th) <= SSR_UV_TOLERANCE)); // No texture UV wrap, if any.
 
 	GIFRegTRXPOS trxpos = {};
 
-	trxpos.DSAX = 0;
-	trxpos.DSAY = 0;
-	trxpos.SSAX = 0;
-	trxpos.SSAY = 0;
+	trxpos.DSAX = r.x;
+	trxpos.DSAY = r.y;
+	trxpos.SSAX = static_cast<int>(m_vt.m_min.t.x / 2) * 2; // Rounded down to closest even integer.
+	trxpos.SSAY = static_cast<int>(m_vt.m_min.t.y / 2) * 2;
+
+	ASSERT(r.x % 2 == 0 && r.y % 2 == 0);
 
 	GIFRegTRXREG trxreg = {};
 
-	trxreg.RRW = m_r.z;
-	trxreg.RRH = m_r.w;
+	trxreg.RRW = r.width();
+	trxreg.RRH = r.height();
+
+	ASSERT(r.width() % 2 == 0 && r.height() % 2 == 0);
 
 	// SW rendering code, mainly taken from GSState::Move(), TRXPOS.DIR{X,Y} management excluded
 
@@ -903,20 +895,16 @@ void GSRendererHW::SwSpriteRender()
 	const int w = trxreg.RRW;
 	const int h = trxreg.RRH;
 
-	GL_INS("SwSpriteRender: Dest 0x%x W:%d F:%s, size(%d %d)", bitbltbuf.DBP, bitbltbuf.DBW, psm_str(bitbltbuf.DPSM), w, h);
+	GL_INS("SwSpriteRender: Dest 0x%x W:%d F:%s, size(%d %d)", m_context->FRAME.Block(), m_context->FRAME.FBW, psm_str(m_context->FRAME.PSM), w, h);
 
-	if (texture_mapping_enabled)
-		InvalidateLocalMem(bitbltbuf, GSVector4i(sx, sy, sx + w, sy + h));
-	InvalidateVideoMem(bitbltbuf, GSVector4i(dx, dy, dx + w, dy + h));
-
-	GSOffset spo = texture_mapping_enabled ? m_mem.GetOffset(bitbltbuf.SBP, bitbltbuf.SBW, bitbltbuf.SPSM) : GSOffset();
-	GSOffset dpo = m_mem.GetOffset(bitbltbuf.DBP, bitbltbuf.DBW, bitbltbuf.DPSM);
+	const GSOffset& spo = m_context->offset.tex;
+	const GSOffset& dpo = m_context->offset.fb;
 
 	const bool alpha_blending_enabled = PRIM->ABE;
 
 	const GSVertex& v = m_vertex.buff[m_index.buff[m_index.tail - 1]]; // Last vertex.
 	const GSVector4i vc = GSVector4i(v.RGBAQ.R, v.RGBAQ.G, v.RGBAQ.B, v.RGBAQ.A) // 0x000000AA000000BB000000GG000000RR
-	                        .ps32(); // 0x00AA00BB00GG00RR00AA00BB00GG00RR
+							  .ps32(); // 0x00AA00BB00GG00RR00AA00BB00GG00RR
 
 	const GSVector4i a_mask = GSVector4i::xff000000().u8to16(); // 0x00FF00000000000000FF000000000000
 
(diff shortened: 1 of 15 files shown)


jeremybris:
If you still need testers, I'll test this. Idk how though.

iMineLink:
Testing is still needed here, and much appreciated, thanks!
To test this PR the simplest way would be to play with Burnout 3 Takedown / Burnout Revenge / Burnout Dominator using the build from this PR artifacts (downloadable from the "Checks" tab) and check the game behaviour, in terms of sky+clouds rendering and visual glitchs/regressions.
I now rebased and force-pushed this PR on top of master, so the artifacts that were previously expired, should be downloadable as soon as the builds are ready.

jeremybris:
Works good for me. Thank you for your work. I just noticed one regression: when I switch to "Software" mode, the emulator crashes. There is a small glitch in all of the games, where the clouds are not present, but it's still better than nothing.

| PR: | PS2: |
| ------------- | ------------- |
| Capture2 | Capture16 |
| Capture5 |  |
| InkedCapture14_LI |  |

MrCK1:
> Works good for me. Thank you for your work. I just noticed one regression: when I switch to "Software" mode, the emulator crashes.

Soudns like you downloaded the 64-bit artifact, SW mode does not work on it at this time. The 32 bit artifact should be fine.

legend800:
That's exactly what I mentioned in the other CR months ago. It's not fixed - the sky is broken, just blue now. No clouds, weird gradient, switching to Software doesn't work/crashes.

JordanTheToaster:
Played through 5 races on Burnout 3 and didn't notice anything all looks good to me.

JordanTheToaster:
I take that back I've found 2 issues one there seems to be a giant seam in the sky and the lighting seems to be going through the terrain and objects again though I'm not 100% on that second one.

JordanTheToaster:
I can reproduce the seam in the sky booting the game normally the sky renders correctly but if I switch to progressive 480p SDTV the seam appears in the sky.

iMineLink:
@legend800 could you confirm that this patch still has issues in going back and forth from software mode, specifically the sky going black?

@JordanTheToaster definitely the clouds are not rendered correctly, as also shown by @jeremybris; On that more research should be done as the effect is not straightforward and possibly also the SW implementation, especially that of the point list rendering (for the cloud palette and other palettes), is bugged.

Thanks all!

iMineLink:
Thanks @legend800 for the extensive testing.
Clouds rendering still needs improvements here, but we are closer to a working solution with respect to the previous attempt.
If you stick in HW and go back and forth from main menu to races (same or different track), do you see any strange additional behaviour occurring? (Sky and clouds are drawn again via patched code when loading track, I'm not entirely sure that the texture overwriting is ok).

jeremybris:
@JordanTheToaster the sun "going through terrain", is already present, #883 , I think.

JordanTheToaster:
> @JordanTheToaster the sun "going through terrain", is already present, #883 , I think.

I'm just being forgetful don't worry about that then.

legend800:
> If you stick in HW and go back and forth from main menu to races (same or different track), do you see any strange additional behaviour occurring? (Sky and clouds are drawn again via patched code when loading track, I'm not entirely sure that the texture overwriting is ok).

- Restarting stage is fine
- Leaving and reentering stage is broken though (no clouds)

2021-03-31 20_06_32-GSdx-hw_ fix Burnout games black sky  by iMineLink · Pull Request #4094 · PCSX2_

Loading a different stage is fine and fixes previous stage:
2021-03-31 20_07_45-Slot_ 0 _ Speed_  82% (49 35) _ Limiter_ Normal _ OGL _ 3200x2240 _ Automatic _

So so long as you keep playing different stages, it's fine (minus the "seams").

pcsx2/GS/Renderers/HW/GSRendererHW.cpp
```diff
@@ -949,31 +959,16 @@ void GSRendererHW::SwSpriteRender()
...
-				GSVector4i sc_alpha_vec;
-
-				if (alpha_c == 2)
-					sc_alpha_vec = GSVector4i(alpha_fix).xxxx().ps32();
-				else
-					sc_alpha_vec = (alpha_c == 0 ? sc : dc0)
+				const GSVector4i A = alpha_a == 0 ? sc : alpha_a == 1 ? dc0 : GSVector4i::zero();
+				const GSVector4i B = alpha_b == 0 ? sc : alpha_b == 1 ? dc0 : GSVector4i::zero();
+				const GSVector4i C = alpha_c == 2 ? GSVector4i(alpha_fix).xxxx().ps32() :
+					(alpha_c == 0 ? sc : dc0)
```
lightningterror, line 965: I'd move this above, it's not that big.

RedDevilus:
Might be best to rename the commits to `GS` instead of `GSdx`

RedDevilus:
Can't recall if the clouds are supposed to cut-out so abruptly, should check on my console to reaffirm.
gs_20211005021526
[gs_20211005021526.zip](https://github.com/PCSX2/pcsx2/files/7281964/gs_20211005021526.zip)

iMineLink:
The clouds are bugged but I have to find the offending draw call(s). I feel that the bug is related to something wrong in the point list palette draw hack.

iMineLink:
The point list palette draw hack had blending completely broken.
I attempted to fix this by recycling the same rendering code used for the sw sprite renderer.
Gs dumps are mostly useless here.
The issue with gs dumps is that the sky related draws (hacked here) happen during level loading, so you do need a huge multiframe dump from the beginning of the loading to debug them and their result on the sky appearance.
Anyway, please go ahead and give this another testing round!

JordanTheToaster:
The giant seam in the sky in Burnout 3 is gone but the clouds are now gone which is a problem.

unknown

iMineLink:
There's more render to clut than expected here.
I'll try plugging in the true sw renderer to perform those draws.

iMineLink:
Not so easy to plug in the sw renderer, the rasterizer unfortunately requires quite a big setup to perform a draw which ultimately seems feasible to mantain only inside a `GSRendererSW` instance.
Anyway, I triple checked, and for the blending that was performed by the clouds palette draw (A = B = 2, C = 2, D = 0), without alpha correction and alpha/depth tests, it is ok to copy the RGBA color as is from the vertex to the framebuffer; So blending was broken in general but not for this specific case.
Instead I realized that the hack worked on top of [9832a935d7416fb06efe78922b55417a39949ae1](https://github.com/PCSX2/pcsx2/commit/9832a935d7416fb06efe78922b55417a39949ae1) (~3 months ago), but the same does not work on top of current master (no clouds).
I didn't try to rebase on top of intermediate commits though, but clearly something really subtle is going on; I'll continue the research.

JordanTheToaster:
Still no clouds sadly we'll get there one day.

iMineLink:
Please test now.

Clouds should work in Burnout 3 PAL, for other games I will need some help in retrieving the 32 bit tex copy draw call during level loading (later the clouds tex is rendered from the copy on top of the sky using a palette at each frame).

JordanTheToaster:
The color of the sky is correct but clouds are still missing from Burnout 3 NTSC.

image

iMineLink:
I attempted at generalizing the cloud rendering detection during level loading, and the code is now working for B3 PAL 50Hz and 60Hz (is it the same at NTSC?).
I wrongly supposed it was a 32 bit texture copy, the draw is instead a proper rendering with a palette.

JordanTheToaster:
Clouds now somewhat showup correctly though they appear to be full of holes and the giant seam is also back.

image

pcsx2/GS/GSVector4i.h
```diff
@@ -2203,4 +2206,11 @@ class alignas(16) GSVector4i
 
 	__forceinline static GSVector4i xff(int n) { return m_xff[n]; }
 	__forceinline static GSVector4i x0f(int n) { return m_x0f[n]; }
+
+	std::string to_string_xyzw()
```
lightningterror, line 2210: Maybe not worth it for a log to include a libraries.

iMineLink:
The buggy clouds ("holes" + "seam") are an upscaling only issue now.
With the last force push I fixed the "holes" by avoiding the upscale of the cloud texture draw during level loading (the one which is downloaded forcibly); Later it is used as an indexed texture at the beginning of each frame to draw the clouds on top of the sky so it's crucial to have it precisely drawn the first time.
Now only the "seam" remain, still an upscaling only issue, I think due to an offset introduced in the UV coordinates of the initial draw when there is upscaling, which is being investigated (I can already detect the draw luckily).
I hope that's the case.

iMineLink:
I must be honest, this was a tough one. There was a bug in the TC (....really??) which did not reset the scale factor to 1 when an old texture was reused as curent RT for a draw at forced native resolution (the clouds). So only a portion of the clouds were downloaded because the wrong (non 1) scale factor was chopping the area to be copied. Hence the "seam" in the clouds.

It should now be fixed.

There's only one thing that sometimes occurs when I load the game but I do not know if it is related to this PR or not: the yellow stripes, which the other dedicated CRC hack `GSC_BurnoutGames` should skip drawing, instead appear on screen.

If testing demonstrates that there are no regressions, I think we should be good to go (code review aside)!

iMineLink:
A fix for a bug in the HW renderer made also the clouds drawn at upscaled resolution possible (the RT was too small before).

RedDevilus:
Burnout Dominator works, needs testing of other TC-buggy games.

refractionpcsx2:
A kind of win/loss here

This PR introduces bars at the bottom of the screen in test drive, however it kinda breaks and half fixes it when you upscale.  Here's a dump for you to look at.
[testdrive.zip](https://github.com/PCSX2/pcsx2/files/7614835/testdrive.zip)

JordanTheToaster:
Burnout 3 NTSC works perfectly now upscaled and haven't noticed any other issues with the games I currently own so it all looks good to me.

iMineLink:
I reverted the automatic rt/ds growth to match the current draw size because it introduces regressions in Test Drive (it makes an unsupported effect, previously luckily cut off, appear).
So I re-added the CU hack to avoid clouds upscale (1024 * 256 draw at native res), which were going to be downloaded (so downscaled to native) in any case.
I also removed the code which allowed dumping of the alpha channel for all the textures when debugging.
Both the rt/ds automatic sizing and the alpha channel dumping will be addressed by dedicated PRs.
A small side effect of not having auto rt/ds size is that the reflections on the cars are a bit worse, but this can be fixed later too.

refractionpcsx2:
I'd rather have the regression in Test drive if it's more accurate TC behaviour, I don't want to avoid progress because it uncovers bugs. if a bug shows it's head, tough, we'll have to deal with it.

iMineLink:
I also think that the auto rt/ds resizing is a step towards more accurate TC (we should really call it texture manager or surface manager or gs memory proxy cause it's really more than a mere "cache") behaviour.
So I restored auto resizing and removed the CU hack.
As for the alpha channel dump, it is re-added but only for dev-builds (should be fine all the time).

This way, by merging this we regress Test Drive in HW, but it can be fixed later.

refractionpcsx2:
What is the alpha dumping going to do in dev builds? I presume nothing unless dumping is enabled for that draw?

lightningterror:
> I'd rather have the regression in Test drive if it's more accurate TC behaviour, I don't want to avoid progress because it uncovers bugs. if a bug shows it's head, tough, we'll have to deal with it.

Agreed.

iMineLink:
@refractionpcsx2 yeah pretty much. It says that when a gs texture is saved to png, also a grayscale png with the alpha content must be saved. Even though it's pretty buggy this way ("alpha" is stored also when there isn't any in the format), it's better than nothing.

ribolovac510:
Black (SLUS-21376) gets stretched out if upscaled on this PR and the black sky is still present if you don't switch to Software rendering during loading.

| PR | v1.7.2092 |
| :---: | :---: |
| 4x Native | 4x Native + F9 during loading |
| gs_20211203100118 | gs_20211203095933 |
| Native |  |
| gs_20211203100137 |  |

Tested it with #3895 also, the image does not get stretched out on that PR but the sky is still black without switching.

iMineLink:
The stretching might be a regression, I'll look into it.
Don't expect Black sky to be not black anymore from this PR, it's only addressing Burnout games.
Later with small effort it can be extended to Black.

iMineLink:
I fixed the auto rt/ds resizing code, Black looks ok also upscaled now.

DonelBueno:
Now Burnout games (at least Burnou 3: Takedown) have issues when upscaled:
PCSX2_2021_12_03_19_40_53_887

RedDevilus:
What's the issue exactly with that screenshot? That the clouds don't look sharp?

refractionpcsx2:
I'm assuming the bottom of the screen is cut off and it looks kinda stretched vertically.

iMineLink:
Could not reproduce, anyway I improved a bit TC. Anyway I noticed that above 3x upscaling the clouds are broken ("holes", probably downscaling is too aggressive when downloading from a huge texture).

refractionpcsx2:
Kingdom hearts shadows are interesting when upscaled

image

Test Drive has changed (in any resolution)

image


Couldn't reproduce any problem with Burnout 3.

iMineLink:
Test Drive is fixed now. Can you retest KH or provide a dump?

refractionpcsx2:
Confirmed test drive is now okay, this is the situation with KH now

image

I'll attach the gsdump so you can check yourself
[kingdomhearts.zip](https://github.com/PCSX2/pcsx2/files/7654701/kingdomhearts.zip)

refractionpcsx2:
Got another one for ya :) Simpsons freaks out when upscaling
[simpsons_upscale.zip](https://github.com/PCSX2/pcsx2/files/7654733/simpsons_upscale.zip)

refractionpcsx2:
actually don't worry about the Kingdom Hearts thing, that happens on master when upscaling too.

But the simpsons is a problem

DonelBueno:
Now Burnout looks perfect, no stretching or cut off image.

refractionpcsx2:
Ah so it was old data? Nice work :)

iMineLink:
It seems so!

B3 clouds only work with odd (x3, x5...) upscaling factors though.
It's a huge 32 bit texture draw (only alpha channel relevant), which gets downloaded (so downscaled) to gs memory.
Later the clouds data is read as 8 bit texture for drawing.
I guess it's some missing half pixel offset somewhere that breaks the rendering with even (x2, x4...) upscaling.

Either we force native resolution draw for the clouds (it is downscaled later in any case), or also this should be fixed.

Again, if we force native res draw, this PR can be split (burnout specific improvements here, improvements to rendering code in another one).

refractionpcsx2:
hmm I don't know why even scaling doesn't work, that's a bit strange, I mean every scale is a multiple so it should scale perfectly :/ Sounds like a bug of some sort, probably worth fixing.  I want to avoid CRC hacks if we can help it.

iMineLink:
With even upscaling, the pixel center in the native texture becomes the boundary between two adjacent pixels in the upscaled texture; this does not occur with odd upscaling, because the center will still be in the middle of a pixel.
This resulted, with even upscaling, in a blurred downscaled texture, which made the masked rendering wrong later (the interpolation of the alpha values broke the masking).
Avoiding linear interpolation when downloading (so downscaling) data greatly improves the rendering quality in this case, and looks a better strategy overall.

refractionpcsx2:
sounds good to me :) man I hate the pixel positions on PS2 lol. I wonder if we should round/correct vertecies/UV's/ST's as they come in before they're passed to the renderer and upscaled etc.

lightningterror:
@iMineLink looks like a pr caused a bunch of conflicts, can you resolve them, I want this next on the merge list since it's been open for a while.

iMineLink:
done

JordanTheToaster:
Clouds look low res and the seam is back at any upscaling. Native works correctly.

image

iMineLink:
a bug has been fixed but the rendering is broken now. needs time to be checked.

iMineLink:
rebased on latest master dropping the last bugfix commit which was already merged. working on my side now.

lightningterror:
Some new warnings which should be taken care of
image

JordanTheToaster (approved):
Works at native 2x 3x ect all fine clouds look perfect and the seam is gone.

image

refractionpcsx2 (approved):

refractionpcsx2:
Tested a bunch of games and GS dumps, all looks good :)

Mrlinkwii (approved):

RedDevilus (approved):
No more issues :)

Bring back the rainbow palette.

refractionpcsx2:
@iMineLink  I think this is good to merge now, you happy to do so or is there anything else you want to do to it?

iMineLink:
For me it's ok to merge!

lightningterror (approved):
Time for the public to enjoy the changes.

freedom85uk:
Hi,

I am new to emulation. Excuse my lack of knowledge, but has this been updated within the source code of PCSX2? I am running the latest nightly build (v1.7.4023). Will this build contain the changes or do I need to something in addition?

Thanks in advance
Luke

refractionpcsx2:
Every new version contains everything from older versions.

If you need further support please use our discord, github is not the place for it

freedom85uk:
Thanks. 

Apologies, I didn't realise I was unable to ask a question on here regarding this issue.

PR #4095, opened 2021-01-02
Discord integration

RedPanda4552:
Round two. Fight.

- Adds Discord SDK to the project and reports current running game, or when in BIOS.
- Controlled with setting in PCSX2 > Game Settings
- Does not support Linux yet, everything is `#ifdef WIN32`'d. Some other mad lad will need to deal with cmake for Linux.
- ~~Needs testing on x64 profile, my computer seems to be unable to run x64 PCSX2, despite having the VC++ runtime installed.~~ Solved it, runs fine.
- ~~If the `WIN32` macro is truly for 32 bit only (I suspect it is not), that will need updated to a macro which will be defined on both 32 and 64 bit Windows.~~ `WIN32` macro is in fact just deception turned into a C++ macro, and also works on the x64 profile.


refractionpcsx2 (approved):

tellowkrinkle:
What's our plan for GPL2 compatibility?

Note: there's always [the old rpc library](https://github.com/discord/discord-rpc) which they claim they'll [continue to support](https://github.com/discord/discord-rpc/issues/290)

kenshen112:
Just leaving a note. If it needs some customer CMake done. I can take a look

RedPanda4552:
Just kidding, Discord SDK isn't GPL compliant. RIP second attempt.

PR #4103, opened 2021-01-05
CI: Appimage

qurious-pixel:
### Description of Changes

Create x86 and x64 AppImage artifacts by adding an AppImage workflow to github actions

### Rationale behind Changes

This allows for the testing of proposed PRs without requiring additional compilation similar to Windows artifacts

### Suggested Testing Steps

Download the appropriate artifact that matches the user architecture. Take note of any errors or missing/outdated libraries.

AppImages must be made executable once downloaded. This can be accomplished by using chmod.
 `chmod +x PCXS2-*.AppImage`


xTVaser:
Hello!

Are you aware of this PR - https://github.com/PCSX2/pcsx2/pull/3741 I largely got AppImages working but there were some issues in some peoples environments and it has largely been abandoned since then as interest has died down.  So happy to see someone picking it up and possibly seeing it through to merging.

Some things I notice here:
1. An entire duplicate workflow should not be required, free OSS repos are limited to 20 concurrent github actions builds at a time, this adds 4 additional builds _per commit_ which I believe are unnecessary.  Instead you should be adding new _steps_ to the existing linux workflow, see this as an example https://github.com/PCSX2/pcsx2/pull/3741/files#diff-d69355d70496176d5fe111e0a1a72ca1edabf246af138f7d7b013f9141eebefbR122-R138
2. I see artifacts from RPCS3 and such that were likely used as a reference of some sort.  Putting aside accreditation issues, thats likely not something we are going to want to depend on (pcsx2's AppImage build process being dependent upon RPCS3's patches AppImageKit).  I was able to get by with using the official AppImageKit.
3. Similarly I see links to curl artifacts from your own personal repo, that is something that we wouldn't want long-term either.  That script should be committed as part of the PR and then no curling should be necessary.

qurious-pixel:
Thanks for the observations, and have made changes accordingly.

1. Eliminated the duplicate workflow. It was doubled to allow granularity and in case, Linux Artifacts were turned on, but have been removed to use the same build.

2. RPCS3 and others have used a patched version of the original AppRun runtime file. This allowed for some control over the C-lib that was passed to the binary. If the system lib was newer, it would use that one, otherwise use the bundled version. I have removed the patched version for simplicty. The original runtime author does reference this patched version for use cases.

3. I had used a custom docker image for the libs, but switched to Github Actions for libraries, and removed all other traces of my repository or other custom solutions. 

4. One of the soultions for the AppImage to find GSdx was eliminated to not alter the way the Linux artifacts are currently produced, and may no longer be needed in the future once GSdx is merged into the core. As of now, it may ask the user each time to select "OK" to proceed.

.github/workflows/linux-workflow.yml
```diff
@@ -31,23 +31,23 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: ubuntu-20.04
+          - os: ubuntu-18.04
```
xTVaser, line 34: Hmm, why the downgrade in ubuntu?  20.04 is the current LTS, and it would be nice to continue using LTS as a general build target.  If 18.04 is truly needed for AppImage generation, this seems a little heavy-handed and could perhaps be contained to just the relevant step.

.github/workflows/linux-workflow.yml
```diff
@@ -103,7 +119,7 @@ jobs:
           key: ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
           restore-keys: |
             ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-
-
+
```
xTVaser, line 122: nit: whitespace

.github/workflows/linux-workflow.yml
```diff
@@ -120,3 +136,20 @@ jobs:
       - name: Build PCSX2
         working-directory: ./build
         run: ../.github/workflows/scripts/linux/compile.sh
+
+      - name: Build AppImage Artifacts
+        env:
+          PLATFORM: ${{ matrix.platform }}
+        run: chmod +x ./.github/workflows/scripts/linux/appimage.sh && ./.github/workflows/scripts/linux/appimage.sh
```
xTVaser, line 143: indentation is off, this would likely lead to some unexpected results.  I'd suggest committing the script with the executable flag as well, if you're on windows you can do that via:
```
git update-index --chmod=+x ./.github/workflows/scripts/linux/appimage.sh
```

.github/workflows/scripts/linux/install-packages.sh
```diff
@@ -6,7 +6,7 @@ set -e
 declare -a BUILD_PACKAGES=(
   "ccache"
   "cmake"
-  "g++-10-multilib"
+  "g++-8-multilib"
```
xTVaser, line 9: not sure about this change, this goes against the current wiki steps, but perhaps this is a side-effect of the 20.04 -> 18.04 change? https://github.com/PCSX2/pcsx2/wiki/Installing-on-Linux#how-to-compile-pcsx2-in-2020-ubuntu-2004

xTVaser (commented):
Big improvement since the last time i reviewed, could probably do with a squash now to get rid of the test commits.  

My only hangup is around downgrading to 18.04 for the entire build, if it makes just the AppImage portion easier and it was localized to just that step I'd be much happier.

.github/workflows/scripts/linux/install-packages.sh
qurious-pixel: Tested against Ubuntu 18.04 i386. It was needed for the C-Lib symbols, to be able to run. It is especially needed since the Apprun patched repo was removed. Since 18.04 is in LTS, it should be the targeted platform. _technically 16.04, until April_

refractionpcsx2:
When you do a squash, if you could rename the commits to have Build: in front of them, just so we have an idea what each commit relates to, that'd be great and try to give a short description of what each commit does.

Thanks :)

.github/workflows/scripts/linux/install-packages.sh
xTVaser: I would suggest building the AppImage in a 18.04 docker environment than if it's required / copying in relevant files to speed up the docker build if required and copying out the AppImage artifact to be published.

I understand 16.04 is still LTS until April 2021, but there is a benefit to continuing to successfully build/target the _current_ LTS.

kenshen112: Hell you could even use something like Arch if that was an option much better multilib packages and they tend to work well with pcsx2

linux_various/PCSX2.desktop
```diff
@@ -0,0 +1,11 @@
+[Desktop Entry]
```
orbea, line 1: PCSX2 already has a desktop file.

orbea:
Downstream packaging should probably be merged in a different repo and not in the main pcsx2 repo where most users probably don't care about it enough to maintain it.

.github/workflows/scripts/linux/install-packages.sh
qurious-pixel: The original submission used a docker image with all of the relevant files loaded. The AppImage could be built with 20.04, or any version, but the newer the system, the more users that are left out. The reason for building the image is to have the correct libraries installed and extend support further than the latest LTS. 

A user running 20.04 could download the binary, and update one or two files. A user running 19.10, might not even have the version of the libraries supported anymore.

This is the benefit I see in packaging with an older LTS, whether in docker or otherwise.

xTVaser:
Are you suggesting that AppImages should not be constructed in the pcsx2 repo?  I disagree with this for a few reasons:
- There does seem to be a decent amount of interest in maintaining it from a few people.
- The primary benefit of AppImages is to provide quick linux test builds for PRs like Windows already has, using a separate repo massively complicates that workflow instead of just having an Action trigger on a PR push.
  - Even if we got this to work elegantly, it would force people to go to a separate repo just to get the PR's artifacts or necessitate _another_ workaround to provide feedback on the parent PR

orbea:
> Are you suggesting that AppImages should not be constructed in the pcsx2 repo?

Yes, the upstream repo should not carry downstream packaging. It will be a maintenance burden that should be delegated to actually interested parties instead of forced upon everyone including those who may not know anything about it or care about maintaining it.

Appimage is not a universal solution on linux and may not even be usable or desirable for some people.

orbea:
@xTVaser How do you expect to maintain all these appimage builds when they are merged into every project which may never update them again and you have countless projects in various states of appimage bitrot? Can't you just use a main appimage repo like any other sane package distribution? You'll be saving the appimage community a lot of effort down the line.

tadanokojin:
It's for testing. We don't distribute our CI artifacts for public use. If some guy wants to make an appimage downstream who care

xTVaser:
> How do you expect to maintain all these appimage builds when they are merged into every project
You'd have to explain/illustrate to me how our CI process would be merged into another unrelated project that would then require a maintenance burden?  That is going to be a nightmare no matter what, AppImage or not.

> It will be a maintenance burden that should be delegated to actually interested parties instead of forced upon everyone including those who may not know anything about it or care about maintaining it.

In this case, we are the interested party.  These are testing artifacts on PRs / master commit builds, they aren't artifacts generated on the release.  You seem to be speaking in general terms without keeping in mind the use-case for this.  When I read through your points, they all seem to be hand-waved away by the fact that these are _temporary testing artifacts_.  If I'm wrong, I'm all ears.

orbea:
> You'd have to explain/illustrate to me how our CI process would be merged into another unrelated project that would then require a maintenance burden? That is going to be a nightmare no matter what, AppImage or not. Why would downstream parties naively run our repo specific CI scripts?

What are you even talking about? Look at almost any distro, they have repos full of package scripts and that model works well. Expecting project specific devs to maintain an appimage build indefinitely is kidding yourself, in a few years it will be an unmaintainable mess where half of the projects have unmaintained appimage builds and some may have no one capable of merging fixes.

> This seems to be on the line of thinking that we are publishing AppImages for general consumption? We aren't, atleast not yet with these proposed changes as was explained prior. 

Again, what are you even talking about? If you publish something to the github releases it is intended for general consumption.


> There is no mention of a ancillary appimage-repo, just the obvious

This is an appimage problem then, you should bring it up with them instead.

GovanifY:
> What are you even talking about? Look at almost any distro, they have repos full of package scripts and that model works well.

Yes they do and we do not want to deprecate them. I am myself a member of the NixOS organization and maintain the NixOS package of pcsx2.

> Expecting project specific devs to maintain an appimage build indefinitely is kidding yourself, in a few years it will be an unmaintainable mess where half of the projects have unmaintained appimage builds and some may have no one capable of merging fixes.

Except it won't because if we don't our testing infrastructure simply won't work anymore on linux. Our current testing infra uses ubuntu's PPA, which as you know is info itelf less portable than AppImage.
 

> Again, what are you even talking about? If you publish something to the github releases it is intended for general consumption.

I would appreciate that you actually understand the subject and read the replies before complaining. Github Actions or any CI setup appear in the PR they are in, which is our main use. Then dev builds are released for wider usage. All distros maintain their own and usually, except a few exceptions a la Gentoo, stick to stable releases.

> This is an appimage problem then, you should bring it up with them instead.

This is not. It is part of our testing infrastructure and adds no cyclomatic complexity, this shouldn't even bother you! You are not our project manager nor will you be. Our testing infra sits on our main repo because it is related to the project and this won't change.

xTVaser:
You clearly aren't reading my replies in full so I don't think there's any point trying to communicate further.  For the third/fourth time, these only need to be maintained BY US for our testing purposes AT PRESENT TIME.

- We are not targeting general consumption
- **We are not targeting these for Github Releases (I was explaining that if WE WERE targetting general consumption, that your advice flies in the face of the referenced official AppImage documentation)**

> This is an appimage problem then, you should bring it up with them instead.

I tend to side with the recommendations and documentation of the AppImage project for this matter.  You seem passionate about disagreeing with their recommendations, I would direct _you_ to go and get that changed, not by proxy through me.

orbea:
> I would appreciate that you actually understand the subject and read the replies before complaining.

Right back at you. I am not the one that is failing to understand and I speak from experience that appimage is a terrible testing platform (As is ubuntu). If you want long term quality you should target something other than the lowest common denominator.

That said I didn't come here to argue argue and if you really want to even more technical debt then w/e... Its not like pcsx2 doesn't have countless real issues to solve.

refractionpcsx2:
> Right back at you. I am not the one that is failing to understand 

I think you are buddy, out of 4 people, some of which are *very* well versed in the way of linux, you are the only one who is failing to grasp what they're trying to tell you.

This is for OUR CI only, NOT for general consumption, how are you failing to grasp this?

tadanokojin:
> Right back at you. I am not the one that is failing to understand and I speak from experience that appimage is a terrible testing platform (As is ubuntu). If you want long term quality you should target something other than the lowest common denominator.

> That said I didn't come here to argue argue and if you really want to even more technical debt then w/e... Its not like pcsx2 doesn't have countless real issues to solve.

I've had enough of you orbea, another one of your temper tentrums and I'm throwing you out. You clearly came into this argument not understanding what you were on about and are now desperately trying to save face. That much is evident by the fact that you didn't even know what this PR does. The problem is your bully tactics don't work here. Not with me here. I will not allow you to harass my team into your religion.

tadanokojin:
I've gone ahead and removed the conversation that was here previously. I think it's unfair to the parties involved to have their work tarnished by what was an unproductive, unnecessary and unsightly argument.

I'd like to extend an apology to @qurious-pixel for what happened here.
I hope that we can continue now with our work with minimal disruption.

orbea:
I was done here already, but I need to point out that censorship is not cool.

.github/workflows/linux-workflow.yml
```diff
@@ -108,15 +118,40 @@ jobs:
...
           PLATFORM: ${{ matrix.platform }}
           COMPILER: ${{ matrix.compiler }}
           ADDITIONAL_CMAKE_ARGS: ${{ matrix.cmakeflags }}
+        if: ${{ matrix.detail != '-appimage' }}
         run: ./.github/workflows/scripts/linux/generate-cmake.sh
 
       - name: Build PCSX2
         working-directory: ./build
+        if: ${{ matrix.detail != '-appimage' }}
         run: ../.github/workflows/scripts/linux/compile.sh
```
xTVaser, line 135: Being as almost all steps are skipped if we are building the app-image, it would be possible to create a separate workflow that just checkouts the repo / creates the AppImage / uploads it.  Unlike originally, this wouldn't create a bunch of extra builds as that duplicated the entire workflow.

However, this depends on my understanding of how the AppImage is built, which may be wrong!

.github/workflows/scripts/linux/docker.sh
```diff
@@ -0,0 +1,29 @@
...
+    -DCMAKE_CXX_COMPILER=/usr/lib/ccache/g++    \
+    -DCMAKE_BUILD_TYPE=Release                  \
+    -DPACKAGE_MODE=TRUE                         \
+    -DXDG_STD=TRUE                              \
+    -DDISABLE_ADVANCE_SIMD=TRUE                 \
+    -DCMAKE_INSTALL_LIBDIR="/tmp/"              \
+    -DCMAKE_INSTALL_DATADIR="/tmp/"             \
+    -G Ninja
+    
+    #-DGTK3_API=TRUE                            \
```
xTVaser, line 21: any particular reason we aren't using GTK3?  If there's a good reason, might be a good idea to add an explanation here so we switch when the blocker is gone

.github/workflows/linux-workflow.yml
```diff
@@ -108,15 +118,40 @@ jobs:
...
+        env:
+          PLATFORM: ${{ matrix.platform }}
+          COMPILER: ${{ matrix.compiler }}
+        if: ${{ matrix.detail == '-appimage' }}
+        run: |
+          if [ ${{ matrix.platform }} = x86 ]; then 
+              docker pull quriouspixel/pcsx2:18.04-386
+              docker run -u root --env-file .github/workflows/scripts/linux/docker.env -v $(pwd):/pcsx2 -v "$GITHUB_WORKSPACE/.ccache":/root/.ccache quriouspixel/pcsx2:18.04-386 /bin/bash /pcsx2/.github/workflows/scripts/linux/docker.sh
+          else
+              docker pull quriouspixel/pcsx2:18.04
```
xTVaser, line 147: i presume these two images are built using the Dockerfiles you included?  It may be worth our time to setup an official pcsx2 dockerhub account so we can update the auto-build and push the images if we have to add/remove packages.

.github/workflows/linux-workflow.yml
```diff
@@ -53,6 +53,16 @@ jobs:
...
+          - os: ubuntu-18.04
+            platform: x86
+            compiler: gcc
+            detail: -appimage
+            experimental: false
+          - os: ubuntu-18.04
+            platform: x64
+            compiler: gcc
+            experimental: false
+            detail: -appimage
```
xTVaser, line 65: I'm wondering if we want to set `experimental` to `true` on these jobs so that if they fail, it won't fail the entire workflow via `continue-on-error`

xTVaser (commented):
Looks good, hopefully some of our testers can give the AppImages a shot soon.  Kinda unfortunate that the AppImage builds take substantially longer (10mins and 20minutes) but maybe thats a caching thing?  If not, I suspect it's something that could be iterated on.

One question I have (not suggesting to change anything) is what are the blockers for building AppImages on ubuntu 20.04 that I/we could track?  It would be nice if in the future we could just tack this onto the end of our normal builds as that would be ideal (shorter build times, etc).

Commits could use a cleanup as well once you are finished iterating.

.github/workflows/linux-workflow.yml
```diff
@@ -5,7 +5,7 @@ name: Linux Build
 on:
   push:
     branches:
-      - master
+      - docker
```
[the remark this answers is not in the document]
qurious-pixel: Yes, it is/was supposed to be set at master for merger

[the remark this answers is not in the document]
qurious-pixel: The appimages don't have a cache built up yet, when they do they are pretty fast vs standard gcc.

.github/workflows/linux-workflow.yml
qurious-pixel: It can be a separate workload. That was my original PR, but it was requested to make it part of the workload to reduce build time, that is also why it is parallel, rather than sequential. (As a matrix element)

.github/workflows/scripts/linux/docker.sh
qurious-pixel: Gtk3 is default, using the cmake flag causes a warning. I was using GTK2, before it broke with a GSdx merge. But kept in the flag as a comment.

.github/workflows/linux-workflow.yml
qurious-pixel: Yes, added the files so PCSX2 could build the image,or combine the two as a mulitlib docker.

.github/workflows/linux-workflow.yml
qurious-pixel: The appimage is set to continue on fail, so if it does not build the other elements of the matrix can succeed.

.github/workflows/linux-workflow.yml
xTVaser: Well the original implement duplicated the entire workflow file and added AppImage steps.  I'm suggesting a separate workflow file that only has the 2 AppImage steps you added now and remove the steps that you skip if `-appimage` is set.

Mrlinkwii:
ive been testing this  , just  a few thing ive  noticed that could  be an  improvement :  first of which would be naming ,it would be better to change the appimage name something similar to what the windows artifacts ( PCSX2-32bit-pr[1234]-sha[xyz]-title[xyz].appimage   etc ) the second thing being if theirs many appimages installed at over a period  which may be  the case when theirs many PRs to test it would be best to somehow isolate each config folder because their may be big changes that may be effect said PR

qurious-pixel:
> it would be better to change the appimage name something similar to what the windows artifacts ( PCSX2-32bit-pr[1234]-sha[xyz]-title[xyz].appimage etc ) 

I can look at using the windows style naming. I did not want the name to become unmanageable. 

Regarding the configuration folder location, AppImages allow you to create a folder to store the configuration on a per AppImage basis. 

PCSX2-32bit-pr[1234].AppImage if run in a directory with a folder named PCSX2-32bit-pr[1234].AppImage.config, all configuration files will be added to that folder.
So multiple versions of the same program can be run side-by-side with different configurations.

qurious-pixel:
Added update file .zsync used to binary update the file. 
As of now, it is included in the zip file, but could be hosted on the server to allow delta updates.

xTVaser (approved):
Looks sane to me and latest commit has all builds still passing with the new AppImage steps completely isolated.  

Maybe @Mrlinkwii or @RedDevilus could give the artifacts another test on their ends?

RedDevilus (approved):

RedDevilus:
AppImage inside a VMWare Workstation 16 + Zorin Os Beta 16:

image
image

There is only English (US)

Overwrite/import works
image


Ubuntu-based VM fails at seeing GSDX plugin:
image

Issue probably resolved when GSDX is merged.

qurious-pixel:
Add translations to AppImage build

image

GovanifY:
> What is needed to get this merged? Tester for the AppImages? Code review?

Everything is more or less good, we are waiting on the GS merge which has just started, then we'll do light testing and final review and we should integrate it in our CI, sorry for the wait!

.github/workflows/appimage-workflow.yml
```diff
@@ -0,0 +1,139 @@
...
+          path: |
+            ./.git/modules/
+            ./3rdparty/fmt
+            ./3rdparty/xz
+            ./3rdparty/yaml-cpp
+            ./3rdparty/gtest
+
+      - name: Checkout Submodules
+        if: steps.cache-submodules.outputs.cache-hit != 'true'
+        run: git submodule update --init --recursive --jobs 2
```
Margen67, line 77:
```suggestion
        run: git submodule update --init --recursive -j 2
```

.github/workflows/appimage-workflow.yml
```diff
@@ -0,0 +1,139 @@
...
+          string(TIMESTAMP current_date "%Y-%m-%d-%H;%M;%S" UTC)
+          message("::set-output name=timestamp::${current_date}")
+
+      - name: ccache cache files
+        uses: actions/cache@v2.1.4
+        with:
+          path: .ccache
+          key: ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
+          restore-keys: |
+            ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-
```
Margen67, line 93:
```suggestion
          restore-keys: ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-
```

.github/workflows/appimage-workflow.yml
```diff
@@ -0,0 +1,139 @@
...
+          else
+              docker pull quriouspixel/pcsx2:18.04
+              docker run -u root --env-file .github/workflows/scripts/linux/docker.env -v $(pwd):/pcsx2 -v "$GITHUB_WORKSPACE/.ccache":/root/.ccache quriouspixel/pcsx2:18.04 /bin/bash /pcsx2/.github/workflows/scripts/linux/docker.sh
+          fi   
+        
+      - name: Upload artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ steps.artifact-metadata.outputs.name }}
+          path: artifacts
```
Margen67, line 139:
```suggestion
          path: artifacts
          if-no-files-found: error
```

.github/workflows/scripts/docker/Dockerfile_x64
```diff
@@ -0,0 +1,97 @@
...
+    ccache \
+    make \
+    cmake \
+    git \
+    ninja-build \
+    apt-get clean autoclean && apt-get autoremove --yes && rm -rf /var/lib/apt /var/lib/cache /var/lib/log /var/lib/apt/lists/*
+    
+ENV CMAKEVER=3.18.4
+RUN \
+	curl -sLO https://cmake.org/files/v${CMAKEVER%.*}/cmake-${CMAKEVER}-Linux-x86_64.sh && \
```
Margen67, line 67: `-S` allows errors, and `-f` prevents a file from being output on error.
```suggestion
	curl -sSfLO https://cmake.org/files/v${CMAKEVER%.*}/cmake-${CMAKEVER}-Linux-x86_64.sh && \
```

.github/workflows/scripts/docker/Dockerfile_x64
```diff
@@ -0,0 +1,97 @@
...
+	sh cmake-${CMAKEVER}-Linux-x86_64.sh --prefix=/usr --skip-license && \
+	rm ./cmake*.sh && \
+	cmake --version
+
+RUN\
+	curl -sLO https://github.com/NixOS/patchelf/releases/download/0.12/patchelf-0.12.tar.bz2 && \
```
Margen67, line 73: ditto
```suggestion
	curl -sSfLO https://github.com/NixOS/patchelf/releases/download/0.12/patchelf-0.12.tar.bz2 && \
```

.github/workflows/scripts/docker/Dockerfile_x64
```diff
@@ -0,0 +1,97 @@
...
+	tar xvf patchelf-0.12.tar.bz2 && \
+	cd patchelf-0.12*/ && \
+	./configure && \
+	make && make install
+
+#ENV WXVER=3.1.4
+#RUN \
+#	cd /tmp && \
+#	curl -sLO https://github.com/wxWidgets/wxWidgets/releases/download/v${WXVER}/wxWidgets-${WXVER}.tar.bz2 && \
```
Margen67, line 82: ditto
```suggestion
#	curl -sSfLO https://github.com/wxWidgets/wxWidgets/releases/download/v${WXVER}/wxWidgets-${WXVER}.tar.bz2 && \
```

.github/workflows/scripts/docker/Dockerfile_x64
```diff
@@ -0,0 +1,97 @@
...
+#	tar xjf wxWidgets-${WXVER}.tar.bz2 && \
+#	cd wxWidgets-${WXVER} && \
+#	mkdir buildgtk && cd buildgtk && \
```
Margen67, line 85:
```suggestion
#	mkdir -p wxWidgets-${WXVER}/buildgtk && \
#	cd wxWidgets-${WXVER}/buildgtk && \
```

.github/workflows/scripts/docker/Dockerfile_x64
```diff
@@ -0,0 +1,97 @@
...
+#	../configure --with-gtk && \
+#	make && make install && \
```
Margen67, line 87: Maybe `-j $(nproc` can be used here to speed it up?

.github/workflows/scripts/docker/Dockerfile_x86
```diff
@@ -0,0 +1,111 @@
...
+    make \
+    cmake \
+    git \
+    ninja-build \
+    apt-get clean autoclean && apt-get autoremove --yes && rm -rf /var/lib/apt /var/lib/cache /var/lib/log /var/lib/apt/lists/*
+    
+#ENV CMAKEVER=3.18.4
+#RUN \
+#	cd /tmp && \
+#	curl -sLO https://cmake.org/files/v${CMAKEVER%.*}/cmake-${CMAKEVER}-Linux-x86_64.sh && \
```
Margen67, line 82: ditto
```suggestion
#	curl -sSfLO https://cmake.org/files/v${CMAKEVER%.*}/cmake-${CMAKEVER}-Linux-x86_64.sh && \
```

.github/workflows/scripts/linux/AppRun
```diff
@@ -0,0 +1,19 @@
...
+ln -sf $APPDIR/usr/bin/plugins /tmp/PCSX2 
+mkdir -p $HOME/.local/share/icons/hicolor/scalable/apps && cp $APPDIR/PCSX2.png $HOME/.local/share/icons/hicolor/scalable/apps
+
+if [ "$(ldd /tmp/PCSX2/libGSdx.so | grep "not found" | awk ' { print $1 } ')" = "libOpenGL.so.0" ]; then
+        export LD_PRELOAD=$APPDIR/usr/bin/vm/libOpenGL.so.0
+fi
+
+$APPDIR/AppRun-patched
+
+unlink /tmp/PCSX2
```
Margen67, line 19: eof newline

.github/workflows/scripts/linux/docker.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -ex
```
Margen67, line 1: Parameters shouldn't be passed to shebang. See here: https://github.com/PCSX2/pcsx2/pull/4427#discussion_r628949877
```suggestion
#!/bin/bash

set -ex
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash -ex
```
Margen67, line 1: Parameters shouldn't be passed to shebang. See here: https://github.com/PCSX2/pcsx2/pull/4427#discussion_r628949877
```suggestion
#!/bin/bash

set -ex
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,61 @@
...
+else
+	APPARCH="x86_64"
+	ARCH="x86_64"
+fi
+
+
+BUILDBIN=/pcsx2/build/pcsx2
+
+cd /tmp
+	curl -sLO "https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-$ARCH.AppImage"
```
Margen67, line 16: ditto
```suggestion
	curl -sSfLO "https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-$ARCH.AppImage"
```

.github/workflows/appimage-workflow.yml
```diff
@@ -0,0 +1,139 @@
...
+        uses: actions/cache@v2.1.4
+        with:
+          path: .ccache
+          key: ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
+          restore-keys: |
+            ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-
+
+      - name: Prepare Artifact Metadata
+        id: artifact-metadata
+        shell: bash
```
Margen67, line 97: This isn't running on Windows or Docker so this is redundant.
```suggestion
```

.github/workflows/scripts/docker/Dockerfile_x86
```diff
@@ -0,0 +1,111 @@
...
+    git \
+    gettext \
+    ccache \
+    make \
+    cmake \
+    git \
+    ninja-build \
+    apt-get clean autoclean && apt-get autoremove --yes && rm -rf /var/lib/apt /var/lib/cache /var/lib/log /var/lib/apt/lists/*
+    
+#ENV CMAKEVER=3.18.4
```
Margen67, line 79: Might also want to create a version variable for patchelf.

.github/workflows/scripts/docker/Dockerfile_x86
```diff
@@ -0,0 +1,111 @@
...
+#RUN \
+#	cd /tmp && \
+#	curl -sLO https://cmake.org/files/v${CMAKEVER%.*}/cmake-${CMAKEVER}-Linux-x86_64.sh && \
+#	sh cmake-${CMAKEVER}-Linux-x86_64.sh --prefix=/usr --skip-license && \
+#	rm ./cmake*.sh && \
+#	cmake --version
+
+RUN\
+	curl -sLO https://github.com/NixOS/patchelf/releases/download/0.12/patchelf-0.12.tar.bz2 && \
```
Margen67, line 88: ditto
```suggestion
	curl -sSfLO https://github.com/NixOS/patchelf/releases/download/0.12/patchelf-0.12.tar.bz2 && \
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,61 @@
...
+cp ./.github/workflows/scripts/linux/AppRun /pcsx2/squashfs-root/AppRun
+curl -sL "https://github.com/AppImage/AppImageKit/releases/download/continuous/AppRun-$APPARCH" -o /pcsx2/squashfs-root/AppRun-patched
+chmod a+x ./squashfs-root/AppRun
+chmod a+x ./squashfs-root/runtime
+chmod a+x ./squashfs-root/AppRun-patched
+
+echo "$GITHUB_RUN_ID" > /pcsx2/squashfs-root/version.txt
+
+mkdir -p /pcsx2/squashfs-root/usr/bin/plugins
+mkdir -p /pcsx2/squashfs-root/usr/bin/vm
```
Margen67, line 45:
```suggestion
mkdir -p /pcsx2/squashfs-root/usr/bin/{plugins,vm}
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,61 @@
...
+cp ./bin/GameIndex.yaml /pcsx2/squashfs-root/usr/bin/plugins/GameIndex.yaml
+export UPD_INFO="gh-releases-zsync|PCSX2|pcsx2|latest|$name.AppImage.zsync"
+export OUTPUT=$name.AppImage
+/tmp/squashfs-root/AppRun --appdir=/pcsx2/squashfs-root/ -d /pcsx2/squashfs-root/PCSX2.desktop -i /pcsx2/squashfs-root/PCSX2.png --output appimage
+
+mkdir /pcsx2/artifacts/
+ls -al .
+mv "$name.AppImage" /pcsx2/artifacts # && mv "$name.AppImage.zsync" /pcsx2/artifacts
+chmod -R 777 ./artifacts
+cd ./artifacts
```
Margen67, line 60:
```suggestion
cd artifacts
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,61 @@
...
+cp ./linux_various/PCSX2.desktop.in ./squashfs-root/PCSX2.desktop 
+sed -i -e 's|Categories=@PCSX2_MENU_CATEGORIES@|Categories=Game;Emulator;|g' ./squashfs-root/PCSX2.desktop
+curl -sL https://github.com/AppImage/AppImageKit/releases/download/continuous/runtime-$APPARCH -o ./squashfs-root/runtime
+mkdir -p squashfs-root/usr/share/applications && cp ./squashfs-root/PCSX2.desktop ./squashfs-root/usr/share/applications
+mkdir -p squashfs-root/usr/share/icons && cp ./squashfs-root/PCSX2.png ./squashfs-root/usr/share/icons
+mkdir -p squashfs-root/usr/share/icons/hicolor/scalable/apps && cp ./squashfs-root/PCSX2.png ./squashfs-root/usr/share/icons/hicolor/scalable/apps
+mkdir -p squashfs-root/usr/share/pixmaps && cp ./squashfs-root/PCSX2.png ./squashfs-root/usr/share/pixmaps
+mkdir -p squashfs-root/usr/lib/
+cp ./.github/workflows/scripts/linux/AppRun /pcsx2/squashfs-root/AppRun
+curl -sL "https://github.com/AppImage/AppImageKit/releases/download/continuous/AppRun-$APPARCH" -o /pcsx2/squashfs-root/AppRun-patched
```
Margen67, line 37: ditto
```suggestion
curl -sSfL "https://github.com/AppImage/AppImageKit/releases/download/continuous/AppRun-$APPARCH" -o /pcsx2/squashfs-root/AppRun-patched
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,61 @@
...
+cd /pcsx2
+mkdir -p squashfs-root/usr/bin
+ls -al $BUILDBIN
+cp -P "$BUILDBIN"/PCSX2 /pcsx2/squashfs-root/usr/bin/
+patchelf --set-rpath /tmp/PCSX2 /pcsx2/squashfs-root/usr/bin/PCSX2
+
+cp ./pcsx2/gui/Resources/AppIcon64.png ./squashfs-root/PCSX2.png
+cp ./linux_various/PCSX2.desktop.in ./squashfs-root/PCSX2.desktop 
+sed -i -e 's|Categories=@PCSX2_MENU_CATEGORIES@|Categories=Game;Emulator;|g' ./squashfs-root/PCSX2.desktop
+curl -sL https://github.com/AppImage/AppImageKit/releases/download/continuous/runtime-$APPARCH -o ./squashfs-root/runtime
```
Margen67, line 30: ditto
```suggestion
curl -sSfL https://github.com/AppImage/AppImageKit/releases/download/continuous/runtime-$APPARCH -o ./squashfs-root/runtime
```

.github/workflows/scripts/docker/Dockerfile_x86
```diff
@@ -0,0 +1,111 @@
...
+	tar xvf patchelf-0.12.tar.bz2 && \
+	cd patchelf-0.12*/ && \
+	./configure && \
+	make && make install
+
+#ENV WXVER=3.1.4
+#RUN \
+#	cd /tmp && \
+#	curl -sLO https://github.com/wxWidgets/wxWidgets/releases/download/v${WXVER}/wxWidgets-${WXVER}.tar.bz2 && \
```
Margen67, line 97: ditto
```suggestion
#	curl -sSfLO https://github.com/wxWidgets/wxWidgets/releases/download/v${WXVER}/wxWidgets-${WXVER}.tar.bz2 && \
```

.github/workflows/scripts/docker/Dockerfile_x86
```diff
@@ -0,0 +1,111 @@
...
+#	tar xjf wxWidgets-${WXVER}.tar.bz2 && \
+#	cd wxWidgets-${WXVER} && \
+#	mkdir buildgtk && cd buildgtk && \
```
Margen67, line 100:
```suggestion
#	mkdir -p wxWidgets-${WXVER}/buildgtk && \
#	cd wxWidgets-${WXVER}/buildgtk && \
```

.github/workflows/appimage-workflow.yml
```diff
@@ -0,0 +1,139 @@
...
+      # Prevent one build from failing everything (although maybe those should be included as experimental builds instead)
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-18.04
+            platform: x86
+            compiler: gcc
+            detail: -appimage
+            experimental: true
+          - os: ubuntu-18.04
```
Margen67, line 39: Might want to use the latest LTS for consistency.
```suggestion
          - os: ubuntu-20.04
            platform: x86
            compiler: gcc
            detail: -appimage
            experimental: true
          - os: ubuntu-20.04
```

.github/workflows/scripts/docker/Dockerfile_x86
```diff
@@ -0,0 +1,111 @@
...
+    wx-common \
+    zenity \
+    wget \
+    curl \
+    git \
+    gettext \
+    ccache \
+    make \
+    cmake \
+    git \
```
Margen67, line 75: This is a duplicate.
```suggestion
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,61 @@
...
+cp /usr/lib/$ARCH-linux-gnu/libOpenGL.so.0 /pcsx2/squashfs-root/usr/bin/vm/
+find $BUILDBIN/../plugins -iname '*.so' -exec cp {} /pcsx2/squashfs-root/usr/bin/plugins \;
+arr=( $(ls -d /pcsx2/squashfs-root/usr/bin/plugins/* ) )
+for i in "${arr[@]}"; do patchelf --set-rpath /tmp/PCSX2 "$i"; done
+cp ./bin/GameIndex.yaml /pcsx2/squashfs-root/usr/bin/plugins/GameIndex.yaml
+export UPD_INFO="gh-releases-zsync|PCSX2|pcsx2|latest|$name.AppImage.zsync"
+export OUTPUT=$name.AppImage
+/tmp/squashfs-root/AppRun --appdir=/pcsx2/squashfs-root/ -d /pcsx2/squashfs-root/PCSX2.desktop -i /pcsx2/squashfs-root/PCSX2.png --output appimage
+
+mkdir /pcsx2/artifacts/
```
Margen67, line 56:
```suggestion
mkdir -p /pcsx2/artifacts/
```

.github/workflows/scripts/docker/Dockerfile_x64
```diff
@@ -0,0 +1,97 @@
...
+    x11-common \
+    zenity \
+    wget \
+    curl \
+    git \
+    gettext \
+    ccache \
+    make \
+    cmake \
+    git \
```
Margen67, line 61: Again, duplicate.
```suggestion
```

.github/workflows/scripts/linux/docker.sh
```diff
@@ -0,0 +1,26 @@
...
+
+cd /pcsx2
+
+mkdir build && cd build || exit 1
```
Margen67, line 5: This might be redundant since `-e` is set.
```suggestion
mkdir build && cd build
```

GovanifY (changes requested):
You have updated the fmt submodule, unless you have a good reason to do so you shouldn't do that in this PR

qurious-pixel:
> You have updated the fmt submodule, unless you have a good reason to do so you shouldn't do that in this PR

It pulled in the pre-7.1.3 `fmt` when I rebased. I will revert that, and update to the 19-Mar commit.

hsigamerppsspp:
Cr App To ps2 Emu Not bug Log For Android 2 Gb ram Up

GovanifY:
@qurious-pixel So, sorry that it might be a little late to comment about that, I was busy with other things.

I noticed this PR uses a dockerfile instead of our https://github.com/PCSX2/pcsx2/blob/master/.github/workflows/scripts/linux/install-packages.sh script, is there any reason behind this? I think we could add a switch to add appimage dependent features there, since both are ubuntu based. 

Thanks again for the work! I anticipate to move along to final testing/review soon enough once this is taken care of.

qurious-pixel:
> I noticed this PR uses a dockerfile instead of our https://github.com/PCSX2/pcsx2/blob/master/.github/workflows/scripts/linux/install-packages.sh script, is there any reason behind this? I think we could add a switch to add appimage dependent features there, since both are ubuntu based.
> 
In the first iteration of this PR I did use the install-package script and even combined with the linux-build workflow, but there are some challenges since the AppImage uses an older distro (based on Ubuntu 18.04) instead of Ubuntu latest (20.04), as well as choosing a different version of gcc/g++ and adding both the i386 and x86_64 libraries. 

I felt Docker uses minimial setup (set once and forget), but it can also be done installing all of the packages on-the-fly.

If it is preferred to remove reliance on the Docker Image, I can show what the workflow file would look like.

GovanifY:
I'm not against it per se but I'm wondering: why would it be easier in this case? I understand it is because we use an older ubuntu version for this appimage setup, but why? Also why do we want a different version of gcc? We already have a clang/gcc toggle along with 32/64 builds on our environment

GovanifY:
@qurious-pixel I remember we discussed about that earlier but what was the reasoning behind putting all the libraries in the workflow file instead of install-packages.sh? Also is there any reason why the appimage workflow is separated from our current workflows? Couldn't we just add an AppImage step once the compilation is done to have artifacts for all of our different builds?

qurious-pixel:
> what was the reasoning behind putting all the libraries in the workflow file instead of install-packages.sh? 

There were version conflicts for some libs between 18.04 and 20.04, and some additional libs were needed for the appimage.  If a carve-out can be created, then they can all go in the same file.

>Couldn't we just add an AppImage step once the compilation is done to have artifacts for all of our different builds?

The workflow file was separated because of the distro versioning, and so that if the appimage fails, (or linux build) it does not fail the entire job (although the experimental flag was set)

.github/workflows/linux-workflow.yml
```diff
@@ -1,4 +1,4 @@
-name: Linux Build
+name: 🐧 Linux Builds
```
GovanifY, line 1: You didn't rebase correctly? Those are master commits

GovanifY (changes requested):
Sorry to make you go back to this, but the rebase wasn't successful it looks like :x

Btw don't hesitate to ping me/request for reviews early rather than waiting 5 days for it, I wasn't aware you needed one! Request early, request often :)

.github/workflows/linux-workflow.yml
qurious-pixel: Those changes were added in a separate PR 
 https://github.com/PCSX2/pcsx2/pull/4522 after the initial rebase, so added manually after.
https://github.com/PCSX2/pcsx2/pull/4529 will also be touching the same code.

GovanifY: Sure, but this still means the rebase isn't correct, those are now part of the master branch

You should

git checkout master
git pull origin master
git checkout app-test
git rebase master

.github/workflows/scripts/linux/generate-cmake.sh
```diff
@@ -10,17 +10,24 @@ else
   export CXX=clang++
 fi
 
-if [ "${PLATFORM}" = "x86" ]; then
-  ADDITIONAL_CMAKE_ARGS="$ADDITIONAL_CMAKE_ARGS -DCMAKE_TOOLCHAIN_FILE=cmake/linux-compiler-i386-multilib.cmake"
+if [ "${PLATFORM}" = x86 ]; then
+  ADDITIONAL_CMAKE_ARGS="$ADDITIONAL_CMAKE_ARGS"
```
tellowkrinkle, line 14: Any reason for separating the toolchain file out of `ADDITIONAL_CMAKE_ARGS`?

qurious-pixel: `Codacy` complains if the variable is not double-quoted, and the entire CMAKE command will fail if the value is non-empty and quoted.

https://github.com/PCSX2/pcsx2/blob/c5505168d32e8946c39c58a44af851fa6a37756e/.github/workflows/scripts/linux/generate-cmake.sh#L30-L31

By separating it into two lines, double-quotes work whether the value is empty or not.

GovanifY (changes requested):
Mostly a few finishing touches here and there but the codebase seems solid, great job!

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,58 @@
...
+if [[ -e "$BUILDPATH/plugins" ]]; then
+  find "$BUILDPATH/plugins" -iname '*.so' -exec cp {} "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/plugins \;
+fi
+cp ./bin/GameIndex.yaml "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/plugins/GameIndex.yaml
+export UPD_INFO="gh-releases-zsync|PCSX2|pcsx2|latest|$name.AppImage.zsync"
+export OUTPUT="$name.AppImage"
+/tmp/squashfs-root/AppRun --appdir="$GITHUB_WORKSPACE"/squashfs-root/ -d "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.desktop -i "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.png --output appimage
+mkdir -p "$GITHUB_WORKSPACE"/artifacts/
+ls -al .
+mv "$name.AppImage" "$GITHUB_WORKSPACE"/artifacts # && mv "$name.AppImage.zsync" $GITHUB_WORKSPACE/artifacts
```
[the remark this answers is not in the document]
qurious-pixel: No need for comment. AppImage can use an update file, `.zsync`, to update the binary. It was for potential future use. It performs a hash check, and updates only the changed bits.

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,58 @@
...
+chmod a+x ./squashfs-root/AppRun-patched
+echo "$name" > "$GITHUB_WORKSPACE"/squashfs-root/version.txt
+mkdir -p "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/{plugins,vm}
+cp -r "$GITHUB_WORKSPACE"/bin/Langs "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/
+cp "$GITHUB_WORKSPACE"/bin/docs/{Configuration_Guide.pdf,PCSX2_FAQ.pdf} "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/plugins
+cp "$GITHUB_WORKSPACE"/bin/cheats_ws.zip "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/plugins
+if [[ -e "$BUILDPATH/plugins" ]]; then
+  find "$BUILDPATH/plugins" -iname '*.so' -exec cp {} "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/plugins \;
+fi
+cp ./bin/GameIndex.yaml "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/plugins/GameIndex.yaml
```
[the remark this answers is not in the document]
qurious-pixel: Folder was named for when there was plugins, and kept files together. Now it is used for documents, cheats_ws, and GameIndex.yml.

.github/workflows/scripts/linux/AppRun
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+export LD_LIBRARY_PATH="/tmp/PCSX2:$APPDIR/usr/lib:$LD_LIBRARY_PATH"
+BINARY_NAME=$(basename "$ARGV0")
+if [[ ! -e "$PWD/$BINARY_NAME.config" ]]; then
+        mkdir "$PWD/$BINARY_NAME.config"
+fi       
+export XDG_CONFIG_HOME="$PWD/$BINARY_NAME.config"
+
+ln -sf $APPDIR/usr/bin/plugins /tmp/PCSX2
```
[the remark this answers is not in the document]
RedDevilus: Unless you want to re-use it for 1.6 AppImage.

.github/workflows/linux-workflow.yml
```diff
@@ -95,22 +97,62 @@ jobs:
         with:
           path: .ccache
           key: ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
-          restore-keys: |
-            ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-
+          restore-keys: ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-
```
GovanifY, line 100: For some reason matrix.detail and matrix.cache have a space there, and appimage has no dash

GovanifY (approved):
Minor nitpick but it LGTM! @xTVaser might want to review it as he deals with gha often, other than that final testing and then ready to merge!

.github/workflows/linux-workflow.yml
qurious-pixel: https://github.com/PCSX2/pcsx2/pull/4529 is going to remove the dashes, so i had started to, but since it is a separate PR, i left it alone. I removed the dash from `detail: appimage` since it has to match the `if...` statements for building the appimage artifacts and uploading them.

https://github.com/PCSX2/pcsx2/blob/db08ddb10410a73a846da62ac010ef827b8a75ee/.github/workflows/linux-workflow.yml#L151
https://github.com/PCSX2/pcsx2/blob/db08ddb10410a73a846da62ac010ef827b8a75ee/.github/workflows/linux-workflow.yml#L158

GovanifY: Seems fine to me if another PR is taking care of it, all good then!

RedDevilus:
Next testing phase in VM:

image
image
image

And then it hardcrashes in Hardware OpenGL, could be due to the virtualisation itself.

Software mode:

Works fine
image

.github/workflows/linux-workflow.yml
```diff
@@ -31,33 +31,35 @@ jobs:
...
-            detail: -nopch
+            detail: nopch
             experimental: false
-          - os: ubuntu-20.04
+          - os: ubuntu-18.04
             platform: x86
             compiler: gcc
             cmakeflags: -DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON -DARCH_FLAG=-march=haswell
-            detail: -avx2-nopch
+            detail: avx2 nopch
```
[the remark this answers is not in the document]
Margen67: #4529

.github/workflows/linux-workflow.yml
```diff
@@ -95,22 +97,62 @@ jobs:
...
-          restore-keys: |
-            ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-
+          restore-keys: ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-
 
+      - name: Prepare Artifact Metadata
+        id: artifact-metadata
+        shell: bash
+        run: |
+          ARCH=$([ "${{ matrix.platform }}" == x86 ] && echo "32bit" || echo "64bit")
+          ARTIFACT_NAME=""
```
Margen67, line 107:
```suggestion
```

.github/workflows/linux-workflow.yml
```diff
@@ -95,22 +97,62 @@ jobs:
         with:
           path: .ccache
           key: ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
-          restore-keys: |
-            ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-
+          restore-keys: ${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}${{ matrix.detail }}-ccache-
 
+      - name: Prepare Artifact Metadata
+        id: artifact-metadata
+        shell: bash
```
Margen67, line 104: This is redundant.
```suggestion
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,56 @@
...
+cd /tmp
+curl -sSfLO "https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-$ARCH.AppImage"
+chmod a+x linuxdeploy*.AppImage
+./linuxdeploy-"$ARCH".AppImage --appimage-extract
+mv /tmp/squashfs-root/usr/bin/patchelf /tmp/squashfs-root/usr/bin/patchelf.orig
+sudo cp /usr/local/bin/patchelf /tmp/squashfs-root/usr/bin/patchelf
+cd "$GITHUB_WORKSPACE"
+mkdir -p squashfs-root/usr/bin
+ls -al "$BUILDBIN"
+cp -P "$BUILDBIN"/PCSX2 "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/
```
Margen67, line 24:
```suggestion
cp -P "$BUILDBIN/PCSX2" "$GITHUB_WORKSPACE/squashfs-root/usr/bin/"
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,56 @@
...
+patchelf --set-rpath /tmp/PCSX2 "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/PCSX2
```
Margen67, line 25:
```suggestion
patchelf --set-rpath /tmp/PCSX2 "$GITHUB_WORKSPACE/squashfs-root/usr/bin/PCSX2"
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,56 @@
...
+mkdir -p "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/app
+cp -r "$GITHUB_WORKSPACE"/bin/Langs "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/
+cp "$GITHUB_WORKSPACE"/bin/docs/{Configuration_Guide.pdf,PCSX2_FAQ.pdf} "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/app
+cp "$GITHUB_WORKSPACE"/bin/cheats_ws.zip "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/app
+
+cp ./bin/GameIndex.yaml "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/app/GameIndex.yaml
+export UPD_INFO="gh-releases-zsync|PCSX2|pcsx2|latest|$name.AppImage.zsync"
+export OUTPUT="$name.AppImage"
+/tmp/squashfs-root/AppRun --appdir="$GITHUB_WORKSPACE"/squashfs-root/ -d "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.desktop -i "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.png --output appimage
+mkdir -p "$GITHUB_WORKSPACE"/artifacts/
```
Margen67, line 51:
```suggestion
mkdir -p "$GITHUB_WORKSPACE/artifacts/"
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,56 @@
...
+ls -al .
+mv "$name.AppImage" "$GITHUB_WORKSPACE"/artifacts # && mv "$name.AppImage.zsync" "$GITHUB_WORKSPACE"/artifacts
```
Margen67, line 53:
```suggestion
mv "$name.AppImage" "$GITHUB_WORKSPACE/artifacts" # && mv "$name.AppImage.zsync" "$GITHUB_WORKSPACE"/artifacts
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,56 @@
...
+echo "$name" > "$GITHUB_WORKSPACE"/squashfs-root/version.txt
+mkdir -p "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/app
+cp -r "$GITHUB_WORKSPACE"/bin/Langs "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/
+cp "$GITHUB_WORKSPACE"/bin/docs/{Configuration_Guide.pdf,PCSX2_FAQ.pdf} "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/app
+cp "$GITHUB_WORKSPACE"/bin/cheats_ws.zip "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/app
+
+cp ./bin/GameIndex.yaml "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/app/GameIndex.yaml
+export UPD_INFO="gh-releases-zsync|PCSX2|pcsx2|latest|$name.AppImage.zsync"
+export OUTPUT="$name.AppImage"
+/tmp/squashfs-root/AppRun --appdir="$GITHUB_WORKSPACE"/squashfs-root/ -d "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.desktop -i "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.png --output appimage
```
Margen67, line 50:
```suggestion
/tmp/squashfs-root/AppRun --appdir="$GITHUB_WORKSPACE/squashfs-root/" -d "$GITHUB_WORKSPACE/squashfs-root/PCSX2.desktop" -i "$GITHUB_WORKSPACE/squashfs-root/PCSX2.png" --output appimage
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,56 @@
...
+
+echo "${PLATFORM}"
+if [ "${PLATFORM}" == "x86" ]; then
+  APPARCH="i686"
+  ARCH="i386"
+else
+  APPARCH="x86_64"
+  ARCH="x86_64"
+fi
+BUILDPATH="$GITHUB_WORKSPACE"/build
```
Margen67, line 13:
```suggestion
BUILDPATH="$GITHUB_WORKSPACE/build"
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,56 @@
...
+BUILDBIN="$BUILDPATH"/pcsx2
```
Margen67, line 14:
```suggestion
BUILDBIN="$BUILDPATH/pcsx2"
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,56 @@
...
+sed -i -e 's|Categories=@PCSX2_MENU_CATEGORIES@|Categories=Game;Emulator;|g' ./squashfs-root/PCSX2.desktop
+sed -i -e 's|__GL_THREADED_OPTIMIZATIONS=1|__GL_THREADED_OPTIMIZATIONS=0|g' ./squashfs-root/PCSX2.desktop
+curl -sSfL "https://github.com/AppImage/AppImageKit/releases/download/continuous/runtime-$APPARCH" -o ./squashfs-root/runtime
+mkdir -p squashfs-root/usr/share/applications && cp ./squashfs-root/PCSX2.desktop ./squashfs-root/usr/share/applications
+mkdir -p squashfs-root/usr/share/icons && cp ./squashfs-root/PCSX2.png ./squashfs-root/usr/share/icons
+mkdir -p squashfs-root/usr/share/icons/hicolor/scalable/apps && cp ./squashfs-root/PCSX2.png ./squashfs-root/usr/share/icons/hicolor/scalable/apps
+mkdir -p squashfs-root/usr/share/pixmaps && cp ./squashfs-root/PCSX2.png ./squashfs-root/usr/share/pixmaps
+mkdir -p squashfs-root/usr/lib/
+cp ./.github/workflows/scripts/linux/AppRun "$GITHUB_WORKSPACE"/squashfs-root/AppRun
+curl -sSfL "https://github.com/AppImage/AppImageKit/releases/download/continuous/AppRun-$APPARCH" -o "$GITHUB_WORKSPACE"/squashfs-root/AppRun-patched
```
Margen67, line 37:
```suggestion
cp ./.github/workflows/scripts/linux/AppRun "$GITHUB_WORKSPACE"/squashfs-root/AppRun
curl -sSfL "https://github.com/AppImage/AppImageKit/releases/download/continuous/AppRun-$APPARCH" -o "$GITHUB_WORKSPACE/squashfs-root/AppRun-patched"
```

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -0,0 +1,56 @@
...
+chmod a+x ./squashfs-root/AppRun
+chmod a+x ./squashfs-root/runtime
+chmod a+x ./squashfs-root/AppRun-patched
+echo "$name" > "$GITHUB_WORKSPACE"/squashfs-root/version.txt
+mkdir -p "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/app
+cp -r "$GITHUB_WORKSPACE"/bin/Langs "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/
+cp "$GITHUB_WORKSPACE"/bin/docs/{Configuration_Guide.pdf,PCSX2_FAQ.pdf} "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/app
+cp "$GITHUB_WORKSPACE"/bin/cheats_ws.zip "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/app
+
+cp ./bin/GameIndex.yaml "$GITHUB_WORKSPACE"/squashfs-root/usr/bin/app/GameIndex.yaml
```
Margen67, line 47:
```suggestion
echo "$name" > "$GITHUB_WORKSPACE/squashfs-root/version.txt"
mkdir -p "$GITHUB_WORKSPACE/squashfs-root/usr/bin/app"
cp -r "$GITHUB_WORKSPACE/bin/Langs" "$GITHUB_WORKSPACE/squashfs-root/usr/bin/"
cp "$GITHUB_WORKSPACE/bin/docs/{Configuration_Guide.pdf,PCSX2_FAQ.pdf}" "$GITHUB_WORKSPACE/squashfs-root/usr/bin/app"
cp "$GITHUB_WORKSPACE/bin/cheats_ws.zip" "$GITHUB_WORKSPACE/squashfs-root/usr/bin/app"

cp ./bin/GameIndex.yaml "$GITHUB_WORKSPACE/squashfs-root/usr/bin/app/GameIndex.yaml"
```

tellowkrinkle:
> image

Have stacktraces always been broken in Linux or is this new?

qurious-pixel:
> > image
> 
> Have stacktraces always been broken in Linux or is this new?

Early in the GS Merge. 
Has been fixed since.

https://github.com/PCSX2/pcsx2/pull/4436#issuecomment-861745774

tellowkrinkle:
> Early in the GS Merge.
> Has been fixed since.

I mean how it doesn't show function names, I was expecting it to look a bit more like this 
image

But if that's not new, we can deal with it some other time and not here

Mrlinkwii:
seems to be working fine for me 
image

GovanifY:
@qurious-pixel I'm missing `libthai.so.0`, maybe a few others on NixOS

qurious-pixel:
> @qurious-pixel I'm missing `libthai.so.0`, maybe a few others on NixOS
> 
> EDIT: here's a list of the libraries found and not found: https://hastebin.com/miwuwajupe.js
> some of those are normal, eg opengl, others aren't, like libthai.

Those files that are missing are on the "Exclusion" list.
https://github.com/AppImage/pkg2appimage/blob/master/excludelist
They are expected to be on the user's system or known to cause issues.

I will follow-up/test with some other NixOS users to find a solution.

GovanifY:
@qurious-pixel Even libthai? damn
If you are interested on NixOS be sure to use appimage-run to run appimages, otherwise it won't be able to find any libraries.

`nix-env -iA nixos.appimage-run` on the minimal cd
If you can't fix it it's fine, not a blocker for a release imo, I just have a weird setup

qurious-pixel:
> @qurious-pixel Even libthai? damn
> If you are interested on NixOS be sure to use appimage-run to run appimages, otherwise it won't be able to find any libraries.

Literally just adding `libthai.so.0` allowed it to boot. This can be added for additional compatibility.

GovanifY:
LGTM!

screenshot

tadanokojin:
I think we're ready on this if there's anymore work to be done we can follow up at this point.

PR #4122, opened 2021-01-09
Core: Attempt to fix CDVD regression

### LINKED ISSUE #4119: [Regression] Disc swapping is more broken with CDVD merge in 1.7
**Describe the bug**
The disc doesn't appear when in the browser (BIOS) when you disc-swap, other disc-swapping features are not working as intended.

**To Reproduce**
Go into BIOS with No ISO and while in browser switch to ISO.

Result:

1.7
image

**Expected behavior**
The disc should pop-up after a moment to say it sees a disk.

1.6
image

 - PCSX2 Revision:  dev-844
 - OS: W10 Pro 20H2
 - CPU: i7-9750H
 - GPU: 2060M 6GB

TheLastRar:
Turns out there is more to CDVD than just DoCDVDOpen()

Hopefully this fixes #4119, and doesn't break anything else.

### DIFF
--- a/pcsx2/System/SysCoreThread.cpp
+++ b/pcsx2/System/SysCoreThread.cpp
@@ -167,7 +167,7 @@ void SysCoreThread::ApplySettings(const Pcsx2Config& src)
 	if (src == EmuConfig)
 		return;
 
-	if (!pxAssertDev(IsPaused(), "CoreThread is not paused; settings cannot be applied."))
+	if (!pxAssertDev(IsPaused() | IsSelf(), "CoreThread is not paused; settings cannot be applied."))
 		return;
 
 	m_resetRecompilers = (src.Cpu != EmuConfig.Cpu) || (src.Gamefixes != EmuConfig.Gamefixes) || (src.Speedhacks != EmuConfig.Speedhacks);
--- a/pcsx2/System/SysCoreThread.cpp
+++ b/pcsx2/System/SysCoreThread.cpp
@@ -326,7 +326,6 @@ void SysCoreThread::OnResumeInThread(bool isSuspended)
 	GetCorePlugins().Open();
 	if (isSuspended)
 	{
-		DoCDVDopen();
 		DEV9open((void*)pDsp);
 		USBopen((void*)pDsp);
 	}
--- a/pcsx2/System/SysThreadBase.cpp
+++ b/pcsx2/System/SysThreadBase.cpp
@@ -332,7 +332,9 @@ bool SysThreadBase::StateCheckInThread()
 			if (m_ExecMode != ExecMode_Closing)
 			{
 				if (g_CDVDReset)
-					DoCDVDopen();
+					//AppCoreThread deals with Reseting CDVD
+					OnResumeInThread(false);
+					
 				g_CDVDReset = false;
 				break;
 			}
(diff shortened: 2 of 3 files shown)


refractionpcsx2:
If you boot a game then swap disc and either resume that game or continue if you didn't pause in the first place, as soon as the game tries to read from the cdvd (even though it won't like it anyway) the entire emulator locks up, master doesn't do this, the game just crashes as you'd expect.

TheLastRar:
The lockup you reported doesn't occur always (it took me a few attempts to trigger it).

After a certain amount of time (See `cdvdVsync()`), PCSX2 calls `cdvdCtrlTrayClose()` if it was in an open state (due to  `cdvdCtrlTrayOpen()` in `AppCoreThread::OnResumeInThread()`)

`cdvdCtrlTrayClose()` does its thing, then calls `GetCoreThread().ApplySettings(g_Conf->EmuOptions);`

if `ApplySettings()` thinks the settings have changed, and emulation is running, it will attempt to pause emulation with `ScopedCoreThreadPause paused_core`, at which point we get stuck

refractionpcsx2:
Strange, it happened every single time for me, you swap disc, wait about 5 seconds (might vary per game, but it needs to be streaming from the disc) and everything locks up and you need to kill it in task manager

TheLastRar:
Sometimes, the game itself locks up before the cdvd lockup occurs, which sometimes seems to prevent this issue.

Other times, `ApplySettings()` believes my settings are unchanged and returns before the ScopedCoreThreadPause, avoiding the lockup

TheLastRar:
What settings are we trying to apply here anyway? https://github.com/PCSX2/pcsx2/blob/ee890c508e03eb38669338c605f3dbf6ef260bc7/pcsx2/CDVD/CDVD.cpp#L590

refractionpcsx2:
I have no idea..

TheLastRar:
Seems to be trying to apply the gamefixes applicable to the newly swapped to disc

TheLastRar:
Given that it's being called from CoreThread, I believe we can ignore checking if the thread is running, as I'm assuming that is done to avoid race conditions (and a thread can't race against itself).

Gamefixes are applied correctly after the call to ApplySettings, so it doesn't seem anything additional extra was needed.

TheLastRar:
> 
> 
> Given that it's being called from CoreThread, I believe we can ignore checking if the thread is running, as I'm assuming that is done to avoid race conditions (and a thread can't race against itself).
> 
> Gamefixes are appear to be applied after the call to ApplySettings, so it seems that is working as intended now.

refractionpcsx2:
that's better, it doesn't die anymore (well, the game does obviously, but at least the emulator doesn't freeze)

RedDevilus:
Boot with NoDisc, go to browser and go to ISO (don't give an ISO aka blank):
ISO switched to physical Disc 
Physical disc switched to ISO (with valid ISO)

[emulog-4122.txt](https://github.com/PCSX2/pcsx2/files/5823304/emulog-4122.txt)

```
What happens if you leave it blank....

(UpdateVSyncRate) Mode Changed to DVD PAL.
	(CdvdSource) HotSwapping CDVD source types from NoDisc to ISO.
Patches: No CRC found, using 00000000 instead.
CDVDiso Error: No filename specified.
(IsoFS) Invalid partition descriptor encountered at block 0x10: ''
```
If i switch to my physical disc which has a game it shows up a disc and u can play
If i give a valid iso for the iso method it also shows a disc and boots correctly.

Tl;dr Seems to fix the issue.

refractionpcsx2:
yes (IsoFS) Invalid partition descriptor encountered at block 0x10: always happens (not just on this PR, since forever) when there is no valid disc/iso available.

PR #4123, opened 2021-01-09
Core: Implement proper %s handling for syscall 117 (0x75)

### LINKED ISSUE #2865: Syscall 0x75 (System Out logging) causes access violation with format specifier "%s"
***PCSX2 version:***
1.5.0 r2843

- **Notes**:
This happens due to PCSX2 attempting to read a string at the address given (e.g. 0x40000) instead of the address that address is mapped to (e.g. 20400000).

AmFobes:
This fixes #2865 

This makes the syscall iterate through the format string which can be considered "slow" but this syscall is only going to be used for debugging purposes anyways :) 
 
Works with my little crude example program
image

### DIFF
--- a/pcsx2/R5900OpcodeImpl.cpp
+++ b/pcsx2/R5900OpcodeImpl.cpp
@@ -1027,15 +1027,47 @@ void SYSCALL()
 			{
 				// TODO: Only supports 7 format arguments. Need to read from the stack for more.
 				// Is there a function which collects PS2 arguments?
-				sysConLog(
-					ShiftJIS_ConvertString((char*)PSM(cpuRegs.GPR.n.a0.UL[0])),
+				char* fmt = (char*)PSM(cpuRegs.GPR.n.a0.UL[0]);
+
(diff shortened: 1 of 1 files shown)


AmFobes:
For those curious or those who'd like to test, here is how you can use this syscall:
```c
void syscallout(char* passedInA0, ...){
    asm __volatile__(
        "li $v1,0x75 \n"
        "syscall    \n" );
}
```
The syscall supports up to 7 arguments, the stack _is not_ used

AmFobes:
Verified working on Manjaro Linux x32 and x64 :)

refractionpcsx2:
What about japanese text? Pretty sure that's the reason it had the ShiftJIS in there

AmFobes:
> What about japanese text? Pretty sure that's the reason it had the ShiftJIS in there

Uh oh, I recall that having problems due to encoding. I'll look into it after work.

AmFobes:
image
Seems to work properly without the ShiftJIS stuff.

PR #4126, opened 2021-01-11
Use a struct for passing native window handles between components

bmwalters:
## Motivation
Prerequisite for #4075.

## Overview
The main improvement in these patches is to use a "tagged union", `NativeWindowHandle`, instead of marshalling `HWND`s on Windows and `Display/Window` on X11 through the `void* pDsp`. This improves type safety, removes some pointer arithmetic and casts, and also makes it easier to implement additional native window types, such as Wayland in #4075 or possibly Cocoa.

## Details
The responsibilities of the classes involved have changed very little in these patches.

Previously, when using `GSopen2`, `AppCorePlugins` and `MTGS` would call both `Pcsx2App.OpenGsPanel` and `GSopen2`. `OpenGsPanel` would set up the global `pDsp` with window handle data obtained from peeking inside `GSFrame`'s `wxWindow`. `GSopen2` would use the window handle data to create GL or DX contexts. Simultaneously, `PAD`, `USB`, and other components are able to access `pDsp` for their own purposes.

When using GSopen1 (`GSopen`; deprecated), `OpenGsPanel` is bypassed. `GSopen` instead creates its own native window and stores its window handle data in `pDsp` on its own. That way `PAD`, `USB`, or other components can still access the window handle data for their purposes.

The only changes in these patches are:
- The new `g_gsWindowHandle` is used instead of `pDsp`.
- `GSFrame` is now responsible for digging into itself to pull out the native window handle. This seems a more appropriate place to put it since `GSFrame` is the owner of its native window data, and it works out to be a bit cleaner in the Wayland PR too.
- `const` references are used where appropriate now instead of raw pointers. The patches prefer copying window handle data instead of storing pointers to it.

## Testing
I tested this on my own Linux machine on this branch using X11 and also with my Wayland patches applied in another branch. I was able to boot games, hide the GS window, show, shutdown, etc. I have not ran this code on Windows.

## Reviewing
Each commit is independently reviewable and should build successfully.

The first meaningful patch is "gui: Represent platform-native window handles with a struct" which adds a new `g_gsWindowHandle` value which coexists with the old `pDsp`. Subsequent commits switch components over to use the new way, then the commit "gui: Remove deprecated pDsp window handle" removes pDsp. That way each commit works independently and review should be a lot easier.

cc @GovanifY


common/include/NativeWindowHandle.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
```
lightningterror, line 1: Misses a copyright header, also formatting should be within clang config.

tadanokojin (commented):
This PR will break our debugging tools. [As I stated](https://github.com/PCSX2/pcsx2/pull/4075/files#r552736479), [GSDumpGUI](https://github.com/PCSX2/GSDumpGUI) uses the legacy `GSopen` API because it [doesn't manage it's own window](https://github.com/PCSX2/GSDumpGUI/blob/3d15f9f44ed20f1b70a4a737433ec2fcc256e146/Library/GSDXWrapper.cs#L208). This is really the only modern use of `GSopen` and it will be removed for the merger once we have an alternative. Details of which are tracked here: https://github.com/PCSX2/pcsx2/discussions/4100. There's really no reason that I can think of why this PR should even touch the legacy `GSopen` API anyway as core should never use it.

While I think in theory it's a good idea, you have 3 copies of the same header, which is a preprocessor mess and [probably could just be a struct with some void*](https://github.com/dolphin-emu/dolphin/blob/79a234eff728b2aeb68c9848f2e4fe861d4f3b91/Source/Core/Common/WindowSystemInfo.h#L19).

While you could remove the header duplication without waiting and maybe keep the original `GSopen` api by separating it out of `_GSopen`, it seems to me this would be much easier after the merger than before. Waiting will also prevent breaking our API before the merger is done.

bmwalters:
You are right that the header is duplicated and is intended to be removed during the GSdx merge. This solution was recommended to me here: https://github.com/PCSX2/pcsx2/pull/4075#discussion_r552729727.

Calling `NativeWindowHandle.h` a "preprocessor mess" is disingenuous. Furthermore, the benefits are reduced if the handle itself is represented with `void*` as this requires casting and pointer arithmetic like the old code did. Using appropriate types means the compiler can catch misuse of the handle data at no cost.

Regarding GSDumpGUI, I made sure to update all usages of the old GSopen API in the source tree to expect a struct pointer instead of an HWND*. I did not realize that there were out-of-tree components that depended on the old calling convention.

In order to proceed I'm fine with either approach: waiting for GSDumpGUI to be merged into this tree then updating it, or reverting GSopen to use the old calling convention (it could still use _GSopen).

arcum42:
Closing due to being obsolete after #4905.

PR #4131, opened 2021-01-12
gsdx: Use constexpr to initialize avx vectors without avx instructions

tellowkrinkle:
Previously, to avoid static initializers from executing avx code before we knew avx was supported, we moved initialization of all global vectors to separate initialize functions, forcing those variables to be non-const

It seems no matter what optimization mode you put them in, C++11 compilers will always constant-evaluate any `constexpr` functions / constructors, avoiding the execution of any instructions at all: https://godbolt.org/z/rz6box

This should also help with future attempts to merge sse/avx plugins, as it will reduce the number of functions that need to be called only sometimes.


gregory38:
Very good idead

PR #4134, opened 2021-01-12
SPU2: Switch to a more accurate noise algorithm

Ziemas:
This algorithm comes from pcsxr and is an implementation of Dr. Hell's research. It's supposed to be very accurate.

Bumps savestate because of new SPU core struct members.

The previous implementation was a simple LFSR and did not use the provided noise clock from the ATTR register.


refractionpcsx2:
Looks good to me, any good way to test this? any example games which make use of this?

atomic83GitHub:
Ratchet and Clank can be a good test I think, I hope I am not wrong with the "noise" meaning.

refractionpcsx2:
I check Neopets, i guess the sliding down ropes sound (which I checked uses noise) sounds a bit stronger, but didn't really notice any other difference. Seems to work though.

refractionpcsx2 (approved):

PR #4137, opened 2021-01-13
pad-linux: Add unsupported gamepad

AmFobes:


RedDevilus (changes requested):
Should be alphabetically instead of L514 use L459 or something it has the same type of controller.

lightningterror:
Can you also pr it upstream?

AmFobes:
You mean here?
https://github.com/gabomdq/SDL_GameControllerDB/pull/418

lightningterror:
Ah nice you already did it.

PR #4140, opened 2021-01-14
Remove wxString ↔︎ std::string implicit conversions

tellowkrinkle:
They're non-unicode on windows and were breaking everything

This removes them from our included wx, preventing them from inadvertently being used.  Hopefully this will make PCSX2 an application that manages two character sets (UTF-8, UTF16), rather than three (UTF-8, UTF-16, Current Windows Locale)

Also fixes up some plugins that were calling the non-unicode versions of various library calls

Note: Should probably test to make sure I didn't break anything in doing that

### DIFF
--- a/pcsx2/DebugTools/Debug.h
+++ b/pcsx2/DebugTools/Debug.h
@@ -242,6 +242,11 @@ public:
 
 		return false;
 	}
+
+	bool Write(const std::string& msg) const
+	{
+		return Write(fromUTF8(msg));
+	}
 };
 
 // --------------------------------------------------------------------------------------
--- a/3rdparty/wxwidgets3.0/include/wx/string.h
+++ b/3rdparty/wxwidgets3.0/include/wx/string.h
@@ -1319,7 +1319,7 @@ public:
     wxString(const wxStdWideString& str)
         { assign(str.c_str(), str.length()); }
   #endif
-
+#ifndef _WIN32 // PCSX2: We don't want to accidentally interpret utf-8 std::strings as something else!
   #if !wxUSE_UNICODE // ANSI build
     // FIXME-UTF8: do this in UTF8 build #if wxUSE_UTF8_LOCALE_ONLY, too
     wxString(const std::string& str) : m_impl(str) {}
@@ -1327,6 +1327,7 @@ public:
     wxString(const std::string& str)
         { assign(str.c_str(), str.length()); }
   #endif
+#endif
 #endif // wxUSE_STD_STRING
 
   // Also always provide explicit conversions to std::[w]string in any case,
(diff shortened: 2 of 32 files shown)


common/include/Utilities/Dependencies.h
```diff
@@ -267,6 +267,7 @@ extern const wxChar *__fastcall pxExpandMsg(const wxChar *englishContent);
 extern const wxChar *__fastcall pxGetTranslation(const wxChar *message);
 extern bool pxIsEnglish(int id);
 
+extern wxString fromUTF8(const std::string& str);
 extern wxString fromUTF8(const char *src);
```
CookiePLMonster, line 271: With `std::string_view`, they can be coalesced to a single function as string views provide "free" constructors from both `std::string` and `const char*`. You lose the luxury of a guaranteed null terminator, but those functions don't necessarily need it anyway.

common/src/Utilities/Console.cpp
```diff
@@ -453,6 +453,28 @@ bool IConsoleWriter::Warning(const wxString fmt, ...) const
     return false;
 }
 
+bool IConsoleWriter::WriteLn(ConsoleColors color, const std::string& str) const
```
CookiePLMonster, line 456: Am I reading this right that with `fromUTF8` using a string_view, **all** those methods could use it too?

pcsx2/DebugTools/Debug.h
```diff
@@ -239,6 +239,11 @@ class ConsoleLogFromVM : public BaseTraceLogSource
 
 		return false;
 	}
+
+	bool Write(const std::string& msg) const
```
CookiePLMonster, line 243: Same here.

CookiePLMonster (approved):
It's a good idea! Just left a few remarks mostly in regards to string views (if they are even available in PCSX2 at the moment?).

GovanifY (approved):
LGTM, good to merge to me!

common/include/Utilities/Dependencies.h
GovanifY: I'd prefer keeping the functions separate to add a clear difference between C++ strings and C strings, as those can be wide char on windows and a real pain in the arse to debug with bad conversions

RedPanda4552 (changes requested):
Automatic gamefixes are not applying.

Test case: Ratchet Deadlocked (NTSC-U)

Master: Applies VU Kickstart and EE Timing Hack automatically
PR: Fails to apply gamefixes even with Automatic Gamefixes enabled

RedPanda4552:
Issue with automatic gamefixes is resolved

refractionpcsx2 (approved):

PR #4149, opened 2021-01-17
DEV9: Re-add pcap (both bridged and switched) on windows

TheLastRar:
Hamachi and Radmin VPN, the, two VPN software I tested, don't seem to play nice with our TAP setup

Hamachi flat out fails when bridged.
Radmin VPN allows bridging, however, a manual network config with an unique IP different on radmin's network wasn't able to find games hosted on the other end.

PCAP (specifically bridged mode) allowed me to host and play LAN games with another player over hamachi.

Game tested was Tribes: Areal Assault.

I've opted to use dynamic loading instead of static linking, to allow systems that don't have winPcap or Npcap to still run PCSX2
I'm using the Npcap SDK for the headers, although it should support both Npcap and winPcap.

Requested by @prafullpcsx2, who helped me test over hamachi, and also tested switched mode for online games.


pcsx2/DEV9/Linux/Linux.cpp
```diff
@@ -69,11 +71,15 @@ void OnInitDialog()
...
+
+	for (size_t i = 0; i < adapters.size(); i++)
 	{
-		dev = pcap_io_get_dev_name(i);
-		gtk_combo_box_text_append_text((GtkComboBoxText*)gtk_builder_get_object(builder, "IDC_ETHDEV"), dev);
-		if (strcmp(dev, config.Eth) == 0)
+		std::string dev = fmt::format("{}: {}", (char*)NetApiToString(adapters[i].type), adapters[i].name.c_str());
+
+		gtk_combo_box_text_append_text((GtkComboBoxText*)gtk_builder_get_object(builder, "IDC_ETHDEV"), dev.c_str());
+		if (config.EthApi == adapters[i].type && strcmp(adapters[i].guid.c_str(), config.Eth) == 0)
```
GovanifY, line 82: inline if

pcsx2/DEV9/Win32/Win32.cpp
```diff
@@ -47,31 +49,29 @@ void SysMessage(char* fmt, ...)
...
+		std::wstring dev = fmt::format(L"{}: {}", (wchar_t*)NetApiToWstring(adapters[i].type), adapters[i].name.c_str());
+		int itm = ComboBox_AddString(GetDlgItem(hW, IDC_ETHDEV), dev.c_str());
+		ComboBox_SetItemData(GetDlgItem(hW, IDC_ETHDEV), itm, i);
 		char guid_char[256];
-		wcstombs(guid_char, al[0][i].guid, wcslen(al[0][i].guid) + 1);
-		ComboBox_SetItemData(GetDlgItem(hW, IDC_ETHDEV), itm, _strdup(guid_char));
-		if (strcmp(guid_char, config.Eth) == 0)
+		wcstombs(guid_char, adapters[i].guid.c_str(), wcslen(adapters[i].guid.c_str()) + 1);
+		//TODO, also check type
+		if (config.EthApi == adapters[i].type && strcmp(guid_char, config.Eth) == 0)
```
GovanifY, line 74: inline if

pcsx2/DEV9/Win32/pcap_io_win32.cpp
```diff
@@ -0,0 +1,284 @@
...
+		return fp_##name(a1, a2, a3, a4);                              \
+	}
+#define FUNCTION_SHIM_5_ARG(retType, name, type1, type2, type3, type4, type5) \
+	typedef retType (*fp_##name##_t)(type1, type2, type3, type4, type5);      \
+	fp_##name##_t fp_##name;                                                  \
+	retType name(type1 a1, type2 a2, type3 a3, type4 a4, type5 a5)            \
+	{                                                                         \
+		return fp_##name(a1, a2, a3, a4, a5);                                 \
+	}
+#define FUNCTION_SHIM_6_ARG(retType, name, type1, type2, type3, type4, type5, type6) \
```
GovanifY, line 74: Shouldn't you replace this define hellfest with C++ generics? That seems really hacky

pcsx2/DEV9/net.h
```diff
@@ -37,10 +37,31 @@ extern mtfifo<NetPacket*> rx_fifo;
 extern mtfifo<NetPacket*> tx_fifo;
 */
 
+enum struct NetApi : int
+{
+	Unset = 0,
+	PCAP_Bridged = 1,
+	PCAP_Switched = 2,
+	TAP = 3,
```
GovanifY, line 45: Do we need to define the elements of the struct? I assume so if we save it as is on config but just asking in case

pcsx2/DEV9/pcap_io.cpp
```diff
@@ -35,65 +35,95 @@
...
+}
+#endif
+
+// Fetches the MAC address and prints it
+int GetMACAddress(char* adapter, mac_address* addr)
+{
+	int retval = 0;
+#ifdef _WIN32
+	IP_ADAPTER_ADDRESSES adapterInfo;
+	PIP_ADAPTER_ADDRESSES buffer;
```
GovanifY, line 118: can't you use an object that has an automatic lifetime instead of having to call delete[] each time?

pcsx2/DEV9/net.h
TheLastRar: The idea was that if we change anything in the enum (i.e. reorder, remove or add elements) we retain backwards compatibility with older config files (given we reference by int rather than name)
I could probably use the auto assigned enum values if we assume that won't be an issue.

I see we have ``IniInterface::EnumEntry()``, which may be the preferred way of storing enums if we ever port DEV9's config code to IniInterface (Out of scope for this PR however).

pcsx2/DEV9/pcap_io.cpp
TheLastRar: Like std::vector? can do

pcsx2/DEV9/Win32/pcap_io_win32.cpp
TheLastRar: I'm not sure how best to convert this to templates? how would I go about that?

pcsx2/DEV9/pcap_io.cpp
TheLastRar: Ended up using a std::unique_ptr instead

pcsx2/DEV9/Win32/pcap_io_win32.cpp
GovanifY: Not sure you need a template actually, you can probably do just fine with an auto and a variadic function, here's an example: 
https://en.cppreference.com/w/cpp/utility/variadic

tellowkrinkle: There isn't really a way to do this without macros without repeating names multiple times (which is a great way to mess up)

Personal recommendation, to avoid repeating for the definitions and the `LOAD_FUNCTION` calls: Move the list of functions to a separate file, then `#include` it once with `FUNCTION_SHIM_5_ARG`, etc defined to what they currently are, and once with `FUNCTION_SHIM_5_ARG` defined to `LOAD_FUNCTION` instead of manually rewriting everything for LOAD_FUNCTION

For a concrete example, LLVM does this a lot:
Definition: https://github.com/llvm/llvm-project/blob/5f022ad6ed8dc23094bb50f83608d5c786a821a9/mlir/lib/Parser/TokenKinds.def
Usage: https://github.com/llvm/llvm-project/blob/5f022ad6ed8dc23094bb50f83608d5c786a821a9/mlir/lib/Parser/Lexer.cpp#L216-L217

pcsx2/DEV9/net.h
GovanifY: It's fine, just wanted to ensure this wasn't an internal state we didn't store anywhere else, LGTM!

pcsx2/DEV9/Win32/pcap_io_win32.cpp
TheLastRar: > 
> 
> Not sure you need a template actually, you can probably do just fine with an auto and a variadic function, here's an example:
> https://en.cppreference.com/w/cpp/utility/variadic
variadic functions can't be used to define other functions at compile time (which is what I'm using the FUNCTION_SHIM_*_ARG macro for)

I could spit the FUNCTION_SHIM_*_ARG into smaller variadic macros, leaving FUNCTION_SHIM_*_ARG to reuse those macros, but I couldn't figure out how to make FUNCTION_SHIM_*_ARG fully variadic.

Let me know if you prefer to spit up the FUNCTION_SHIM_*_ARG macors or not?

> 
> 
> Personal recommendation, to avoid repeating for the definitions and the `LOAD_FUNCTION` calls: Move the list of functions to a separate file, then `#include` it once with `FUNCTION_SHIM_5_ARG`, etc defined to what they currently are, and once with `FUNCTION_SHIM_5_ARG` defined to `LOAD_FUNCTION` instead of manually rewriting everything for LOAD_FUNCTION

Thats a decent Idea.

GovanifY: @TheLastRar What is your current implementation? I'll take a stab at it and see if I have a better idea, otherwise tellow's idea works too

TheLastRar: FUNCTION_SHIM_*_ARG defines a type, variable (holding the fp) and function (matching the signature of the function defined in the pcap headers). The function just calls the fp held in the variable the macro also defined

Take ``FUNCTION_SHIM_3_ARG(int, pcap_next_ex, pcap_t*, struct pcap_pkthdr**, const u_char**)``
It expands to;
```
typedef int (*fp_pcap_next_ex_t)(pcap_t*, struct pcap_pkthdr**, const u_char**);
pcap_next_ex_t fp_pcap_next_ex;
int pcap_next_ex(pcap_t* a1, struct pcap_pkthdr** a2, const u_char** a3)
{
    fp_pcap_next_ex(a1, a2, a3);
}
```

LOAD_FUNCTION takes the job of calling ``GetProcAddress()`` and assigning the returned function pointer to the correct variable.

Take ``LOAD_FUNCTION(pcap_next_ex);``
It expands to;
```
fp_pcap_next_ex = (fp_pcap_next_ex_t)GetProcAddress(hpcap, "pcap_next_ex");
if (fp_pcap_next_ex  == nullptr)
{
    FreeLibrary(hpcap);
    Console.Error("%s not found", "pcap_next_ex");
    hpcap = nullptr;
    return false;
}
```

Hopefully that is clear, let me know when you have attempted this

TheLastRar: To further clarify, ``pcap_next_ex()`` is defined in the pcap headers and used in ``pcap_io.cpp`` the ``FUNCTION_SHIM_*_ARG`` macro provides an implemtation for that already defined function, which then just calls the function pointer I load via ``LOAD_FUNCTION``

PR #4168, opened 2021-01-20
pad-linux: Update logs

lightningterror:


pgert:
"GameControllerDB"?

RedDevilus:
https://github.com/gabomdq/SDL_GameControllerDB uses all lowercase, so it easily can be used as submodule.

PR #4172, opened 2021-01-23
Repairing dead links

jeltaqq:


refractionpcsx2:
Thanks, didn't know that was broken

PR #4173, opened 2021-01-23
Misc: Some more warning cleanups

lightningterror:


refractionpcsx2:
The psxCounters changes where the comparison of psxRegs.cycle is done, this will probably break when psxRegs.cycle loops around at INT_MAX, if you cast the sCycle + cycleT to 64 bits.

PR #4180, opened 2021-01-25
CI: Add an explicit flag to retain debugging artifacts (.pdb, .exp, .lib, etc) in Windows builds

xTVaser:
When debugging a recent PR we ran into issues where the CI build would consistently produce an exception that could not be replicated conventionally via the local Visual Studio build.  It seems valuable to me to easily flip on these artifacts going forward, to make it easier to attach a debugger to the process and get a stacktrace at a minimum.

However, enabling these does increase the artifact size by 4x (numbers in the actions summary pages are the uncompressed file-size which are much larger).  So I've left them still off by default.

An example commit/build that flips this flag to `true` - https://github.com/xTVaser/pcsx2-rr/actions/runs/508048107

### DIFF
--- a/.github/workflows/windows-workflow.yml
+++ b/.github/workflows/windows-workflow.yml
@@ -23,6 +23,12 @@ on:
       - debian-packager/
       - bin/PCSX2_keys.ini.default
       - "pcsx2/PAD/Linux/**"
+  workflow_dispatch:
+    inputs:
+      retainDebugArtifacts:
+        description: 'Retain debug artifacts (.pdb/.exp/.lib) for an easier debugging experience'
+        required: true
+        default: 'false'
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


lightningterror (approved):

tellowkrinkle:
Would it be possible to use [workflow_dispatch](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) to allow someone to trigger a build without needing to make a new commit?

xTVaser:
> Would it be possible to use [workflow_dispatch](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) to allow someone to trigger a build without needing to make a new commit?

Hmm, that's a good idea, the only issue (needs to be tested, blog post isn't super comprehensive) is that I would assume someone (like myself) without write-access would not be able to trigger the workflow dispatch..

xTVaser:
Gave workflow dispatch a try, it looks like this in use (on the `Windows` page of the actions tab):
image

Compare artifacts on the workflow dispatch one - https://github.com/xTVaser/pcsx2-rr/actions/runs/659427067
VS a normal automatic build - https://github.com/xTVaser/pcsx2-rr/actions/runs/659425925

.github/workflows/windows-workflow.yml
```diff
@@ -23,6 +23,12 @@ on:
       - debian-packager/
       - bin/PCSX2_keys.ini.default
       - "pcsx2/PAD/Linux/**"
+  workflow_dispatch:
+    inputs:
+      retainDebugArtifacts:
+        description: 'Retain debug artifacts (.pdb/.exp/.lib) for an easier debugging experience'
```
RedDevilus, line 29: `'Retain debug artifacts (.pdb/.exp/.lib) for an easier debugging experience? (true or false)'
`
Perhaps?

PR #4182, opened 2021-01-25
GSdx-GUI: Fix alignment of font

RedDevilus:
Should fix invisible characters and move hardware renderer settings 2px down.


lightningterror:
Still need to move sw renderer stuff, ok, cancel buttons and increase the height.

lightningterror (approved):

PR #4184, opened 2021-01-26
DEV9: Additional logging for debugging DHCP issues with PCSX2

TheLastRar:
Using a port of the packet readers CLR_DEV9 uses, as such, it may not be the most optimal code in a CPP world (there is room for optimasition in it's C# version aswell)

Written as an attempt to debug issues with PCSX2 and DHCP comminucations.
The reasoning behind this is that this works with some virtual DHCP servers (CLR_DEV9 and the ICS service), but not with my physical router (as has been reported widly).

I had initially attempted to modify the recived DHCP packets (to include options missed off the request list), before I added ARP logging and discovered the issue (hence the commented code)

In my case, the issue was my router returning my PC's IP address in any DHCP offer to PCSX2.

I have a second router that provides a unique IP (as it should) in the offer, which allows PCSX2 to follow with a DHCP request, however, the router itself ignores that (probably packet loss?, it isn't very responsive to DHCP requests and took many attempts, windows has issues connecting so I guess that's par for the course).


lightningterror:

PR #4187, opened 2021-01-27
GameDB: Add patches for KOF series

kozarovv:
Description by kojin: 
Fix for Depth precision.
Game fills upper 16bits of depth with 0xFFFF.
This results in a really high 32 bit value which is then converted to float because both hw and sw renderers lack double precision the lower 16 bits of the initial 32 bit value lose precision.

### DIFF
--- a/bin/GameIndex.yaml
+++ b/bin/GameIndex.yaml
@@ -13922,6 +13922,19 @@ SLES-53998:
 SLES-53999:
   name: "King of Fighters, The - Neo Wave"
   region: "PAL-E"
+  patches:
+    F5090D15:
+      content: |-
+        author=kozarovv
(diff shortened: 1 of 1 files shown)


Immersion95:
Hi, thanks a lot for this fix.

I can see that there is some missing games & regions. The full list would be all the KOF games with 3D backgrounds : 

All regions :
KOF 2001 
KOF 2002
KOF 2003
KOF XI
KOF '94 Re-Bout
KOF '98 Ultimate Match
KOF 2002 Unlimited Match
KOF 2002 Unlimited Match (Tougeki)
KOF Neowave
NeoGeo Battle Coliseum

Most Neo Geo Online Collection aren't affected as they use emulation/2D elements.

PR #4188, opened 2021-01-28
GSdx: switch to unicode

GovanifY:
This fixes one of the blockers required for https://github.com/PCSX2/pcsx2/discussions/4100 .

It has been lightly tested, should be tested a bit more before merging but no need to be extremely thorough imho.

### DIFF
--- a/plugins/GSdx/GS.cpp
+++ b/plugins/GSdx/GS.cpp
@@ -942,7 +942,8 @@ public:
 
 			AllocConsole();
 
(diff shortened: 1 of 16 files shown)


AmFobes:
- [x] Compiles
- [x] Doesn't catch fire
- [x] Games play fine with no noticeable perf loss
- [x] UI still intact 
Seems to work well for me :^)

tadanokojin (approved):

dio-gh (approved):

PR #4190, opened 2021-01-29
Add fixes for Jak II/3 prototypes

LuminarLight:
For Jak 3 prototypes, running the DISKINFO.BIN file (it is an ELF) directly is necessary if you want to apply pnach cheats.

For Jak II prototypes, running the DRIVERS/FIREWIRE.IRX file (it is an ELF) directly is necessary if you want to apply pnach cheats.

All these prototypes have an SCES/SCUS ELF that is loaded first, which was used for the WIBU protection. This is the ELF that PCSX2 gets the CRC from. There is nothing wrong with this, unless you want to apply pnach cheats to these builds. In which case you have to run their real ELF files directly. But that changes the CRC, so automatic fixes will be gone. This pull requests fixes this.

### DIFF
--- a/bin/GameIndex.yaml
+++ b/bin/GameIndex.yaml
@@ -26,6 +26,26 @@
 # ---------------------------------------------
 # -- Game List
 # ---------------------------------------------
+3828BB66:
+  name: "Jak 3 [July 2004 Internal]"
+  region: "NTSC"
+  gameFixes:
+    - VU0KickstartHack # Fixes Character SPS.
+4FF025AF:
+  name: "Jak II: Renegade [July 2003 Preview]"
+  region: "PAL"
+  gameFixes:
+    - VU0KickstartHack # Fixes Character SPS.
+A6630572:
+  name: "Jak 3 [September 2004 Preview]"
+  region: "PAL"
+  gameFixes:
+    - VU0KickstartHack # Fixes Character SPS.
+C71BAB1E:
+  name: "Jak II: Renegade [August 2003 Review]"
+  region: "PAL"
+  gameFixes:
+    - VU0KickstartHack # Fixes Character SPS.
 GUST-00009:
   name: "Mana Khemia: Alchemists of Al-Revis [Premium Box]"
   region: "NTSC-J"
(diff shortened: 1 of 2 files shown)


bin/GameIndex.yaml
```diff
@@ -26,6 +26,26 @@
 # ---------------------------------------------
 # -- Game List
 # ---------------------------------------------
+3828BB66:
```
lightningterror, line 29: You need to check for the game serial.

refractionpcsx2: It's a prototype, there probably isn't one, so it might not work with our DB structure very well..

lightningterror: If the crc isn't changed then adding it in gsdx is probably fine, not a fan adding it to gamedb without a serial check tho.

LuminarLight: The serial is same as their final versions' serials. But when you use "Run ELF..." for their real ELFs (DISKINFO.BIN for Jak 3 prototypes, FIREWIRE.IRX for Jak II prototypes), the emulator doesn't see any serial. That is why I had to add the CRCs themselves.

refractionpcsx2: I realise this, but the problem with adding CRC's like this is they are not unique and may clash with other games and could apply fixes that break others games, and I honestly am not willing to add that risk to fix some prototypes when you can probably patch the ISO's SYSTEM.CNF files to run the correct ELF instead of the protected one.

LuminarLight: Yes, they tried to disguise it. What looks like a normal SCES/SCUS game ELF is actually the protection.

My problem is not the prototype protection. The problem is with the pnach system. Protection ELFs for prototypes of the same game have identical ELF files. So pcsx2 can't make difference between them, unless I run the real game ELF directly.

Well I had a feeling that these changes wouldn't be accepted. And it is quite understandable. Thank you for your time. Maybe in the future the pnach system will be changed somehow...

No it is not really needed, we will just add both these commits into our branches.

refractionpcsx2: As I say, I'm happy for us to add the GSdx CRC's but I'm just not overly happy with the GameDB Stuff.

So feel free to modify your PR to just have the GSdx stuff :)

LuminarLight:
Closing this, it is not that important and I don't want to break other games.

PR #4193, opened 2021-01-30
GS: Single build with all ISAs

TellowKrinkle:
Fixes one of the blockers on #4100

~~Note: Requires #4131 and #4658~~

~~Needs msbuild implementation, not sure how to set that up.  You need to split GSdx into a main piece and the ISA-specific piece, which should be compiled once per isa in ["avx2", "avx", "sse4", "sse2"] and with the added define `MULTI_ISA_UNSHARED_COMPILATION`.  Finally, combine everything into one massive GSdx dll.  In CMake I accomplished this by making static library targets for the ISA specific pieces and linking them with the main GSdx.~~ We will be doing Windows multi-isa builds with CMake, since they're really only need to be done on the CI.

Details:
~~Does the dynamic split at the GSRenderer class, initializes a different GSRenderer depending on the ISA.  I had to add an `IGSRenderer` virtual-only interface as GSRenderer didn't seem like it would be easy to separate from its implementation details~~
Does the dynamic split by choosing different functions to put into our many function lookup tables.  Plus a few manual ones for XXHash.


GovanifY:
I went around and rebased this PR to the latest constexpr changes master had

plugins/GSdx/MultiISA.h
```diff
@@ -0,0 +1,67 @@
...
+
+// TKR: These were in GSHwHack.cpp (unshared), I assume they were done this way so they would show up in a debugger so I moved them here instead of making them static
+extern bool s_nativeres;
+extern CRC::Region g_crc_region;
+
+namespace GSStateISAShared
+{
+	// TKR: No clue what this is, it was in GSState.h but it's used by `GSDevice`s, outside of the unshared area
+	// So now it's here
+	extern int s_n;
```
iMineLink, line 66: I think this is the incremental draw call number, used for debug/dump purposes.

tadanokojin: that's what it is indeed.

plugins/GSdx/stdafx.h
```diff
@@ -438,3 +424,30 @@ extern const std::string root_hw;
...
+# endif
+#else
+// Define to isa_native in shared section in addition to multi-isa-off so if someone tries to use it they'll hopefully get a linker error and notice
+# define CURRENT_ISA isa_native
+#endif
+
+#ifndef MULTI_ISA_SHARED_COMPILATION
+# define MULTI_ISA_UNSHARED_START namespace CURRENT_ISA {
+# define MULTI_ISA_UNSHARED_END }
+# define MULTI_ISA_UNSHARED_IMPL using namespace CURRENT_ISA
```
iMineLink, line 448: neat

plugins/GSdx/GSUtil.cpp
```diff
@@ -195,7 +194,7 @@ bool GSUtil::HasCompatibleBits(uint32 spsm, uint32 dpsm)
 
 bool GSUtil::CheckSSE()
 {
-	if (!g_cpu.has(Xbyak::util::Cpu::tSSE2))
```
iMineLink: Maybe the minimum is SSE4 now?

fantesykikachu:
With the GS merge is this still useful, if so can we get a rebase.

refractionpcsx2:
Yes it's still useful and planned, no those won't make it obsolete, and yes it will be rebased when Tellow gets time, he has a lot of PR's to rebase and update since the merge, so he will get around to it when he can.

tellowkrinkle:
> Edit: also won't this be mostly obsoleted by #3579 and #3577

I'd like to think I'm not dumb enough to PR a fix that's already been obsoleted by another PR that I made

> can we get a rebase

This one's waiting for some cmake changes before it can be merged so I probably won't rebase it for a bit.  We ended up deciding to use cmake for published builds on Windows to get this working there, which means that we need cmake builds working on Windows first

refractionpcsx2:
Now all the blockers are merged on this, maybe we can consider testing this and getting it ready?

seta-san:
Can’t wait for this one. One build!

tellowkrinkle:
For anyone wondering about the status of this, it's waiting until we figure out why cmake builds are slower than msbuild builds (even though they're using the same compiler...)

Since it would be pretty pointless to switch to a build that combines SSE4 and AVX2 but is slower than just giving everyone an SSE4 build

pcsx2/CMakeLists.txt
```diff
@@ -932,6 +941,37 @@ else(WIN32)
 	)
 endif(WIN32)
 
+if(DISABLE_ADVANCE_SIMD)
+	target_compile_definitions(PCSX2 PRIVATE MULTI_ISA_SHARED_COMPILATION)
+	if(WIN32)
+		set(compile_options_avx2 /arch:AVX2)
+		set(compile_options_avx  /arch:AVX)
+		if (_M_X86_32)
```
seta-san, line 949: 32bit is now out.

iMineLink:
You could give a try doing a diff of the vcxproj files generated by cmake and those handcrafted, you should see some differences in the flags (probably due to how cmake interpret by default the Release mode vs how we do).

seta-san:
Which build is supposed to be tested if this combines sse4 and avx2?

tellowkrinkle:
> Which build is supposed to be tested if this combines sse4 and avx2?

CMake

seta-san:
if anything i've found this build to be marginally faster. doing all testing with software renderer of course.

PR #4195, opened 2021-01-30
SPU2: Gaussian interpolation

Ziemas:
This implements the actual SPU interpolation as documented by nocash.


HelicopterP:
I don't know what the root cause of this but with Gausian interpolation I can hear a lot of high frequency noise during some sounds in games. 
I tried Vexx and Sphinx and the Cursed Mummy. I do not remeber areal PS2 sounding this scratchy.

MrCK1:
Tested on Sphinx the Cursed Mummy and I can confirm there's a slight issue with the algorithm. Compared to Catmull-Rom (top), there are buzzing artifacts with lower quality samples - manifests as the vertical blue blocks I've circled below.


MP3 samples:
image
[sphynx_samples.zip](https://github.com/PCSX2/pcsx2/files/5899956/sphynx_samples.zip)

Ziemas:
Out of ideas on what's even going on here for now. Not resetting SP worked in some situations but not all, why this would matter in the first place though I'm not sure about.

What's really strange is that this same algorithm doesn't cause these issues in Dobiestation and I don't really know what could possibly be different this early on in the audio path.

ADormant:
@Ziemas maybe it's a core bug?

refractionpcsx2:
> @Ziemas maybe it's a core bug?

It's weird, it only seems to happen with Gaussian interpolation. AFAIK none of the of the other methods exhibit this behaviour.  It's certainly very weird.

Ziemas:
I think I've fixed this. The issue was the ` > 0` check for incrementing the sample index. We need to get the next sample every whole 0x1000 increment. Otherwise the interpolation index is going to be wrong.

Ziemas:
### Before SP fix


### After SP fix

HelicopterP:
I've tried several games and everything seems fine now.

Issue #3233, opened 2020-02-02
increase onepad plugin gamepad number


adit787:
since many game support more than 2 player please change the default gamepad number to 4 or 8
i have tried to compile onepad with 8 pad but pcsx from ppa wont recognize the plugin, maybe because iam using ubuntu 64bit to compile

ordovice:
Would like to add a +1 to this to support the equivalent of multitap functionality.

arcum42:
If it gives some perspective on why this hasn't been done, I'm the most likely one to make changes beyond updating the gamepad database to onepad, and I've never used multitap. All the ps2 games I play, I've basically played solo, so I'm not familiar with it from a user *or* developer standpoint.

Beyond that, I was just looking through what games support multitap. Only thing I see I have on the list at the moment is Ape Escape: Pumped and Primed, and I didn't even realize that used multitap until now. I'm pretty sure I fished it out of a bargain bin because I remembered there being some issue it was a test case for... (Something with VIF?)

All in all, I highly encourage other people to hack multitap support into onepad and submit a PR.

ordovice:
List of Multitap games: 
https://en.wikipedia.org/wiki/PlayStation_Multitap

FYI: Lilypad has it, so might be worth looking in that code base.  @arcum42 Appreciate all you do for the community.  Sadly I'm not much of a coder in that regards so I wouldn't even know where to begin.

KrossX:
If I remember correctly, the API makes it quite simple. `PADqueryMtap` to see if there's a mtap device in that port, then `PADsetSlot` to set which pad to poll next.

refractionpcsx2:
Qt now contains multitap (which uses OnePad), so marking this as fixed.

PR #4196, opened 2021-01-30
multitap on linux

nadenislamarre:
implement the small part missing to make multitap working
on linux. This fixes https://github.com/PCSX2/pcsx2/issues/3233

Signed-off-by: Nicolas Adenis-Lamarre <nicolas.adenis.lamarre@gmail.com>

### DIFF
--- a/pcsx2/PAD/Linux/state_management.cpp
+++ b/pcsx2/PAD/Linux/state_management.cpp
@@ -176,6 +176,26 @@ u8 pad_poll(u8 value)
 
 	Pad* pad = &pads[query.port][query.slot];
 
+	int padNum = -1;
+	if (query.port == 0 && query.slot == 0 && s_vgamePad.size() > 0)
+		padNum = 0;
+	if (query.port == 0 && query.slot == 1 && s_vgamePad.size() > 2)
+		padNum = 2;
+	if (query.port == 0 && query.slot == 2 && s_vgamePad.size() > 3)
+		padNum = 3;
+	if (query.port == 0 && query.slot == 3 && s_vgamePad.size() > 4)
+		padNum = 4;
+	if (query.port == 1 && query.slot == 0 && s_vgamePad.size() > 1)
+		padNum = 1;
+	if (query.port == 1 && query.slot == 1 && s_vgamePad.size() > 5)
+		padNum = 5;
+	if (query.port == 1 && query.slot == 2 && s_vgamePad.size() > 6)
+		padNum = 6;
+	if (query.port == 1 && query.slot == 3 && s_vgamePad.size() > 7)
+		padNum = 7;
+	if (padNum == -1)
+		return 0;
+
 	if (query.lastByte == 0)
 	{
 		query.lastByte++;
@@ -231,7 +251,7 @@ u8 pad_poll(u8 value)
 					b1=b1 & 0x1f;
 #endif
 
-				uint16_t buttons = g_key_status.get(query.port);
+				uint16_t buttons = g_key_status.get(padNum);
 
 				query.numBytes = 5;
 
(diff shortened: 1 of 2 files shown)


pcsx2/PAD/Linux/state_management.cpp
```diff
@@ -231,7 +242,7 @@ u8 pad_poll(u8 value)
 					b1=b1 & 0x1f;
 #endif
 
-				uint16_t buttons = g_key_status.get(query.port);
+				uint16_t buttons = g_key_status.get(padNum);
```
RedDevilus, line 245: Will this still work fine even if multitap is disabled? I see you change the old variable but reuse query.port then as extra check to then pipe it to padNum in multitap mode.

nadenislamarre: i've tested from 1 player to 8 players. without multipad 1 & 2, with only 1, only 2.
specifically for these 2 lines yes : 
query.port is the port : it was always 0 or 1 in the previous implementation and mapped then on joystick sdl id 0 and 1.
g_key_status.get() require the pad number. on the first step, i just compute the correct pad value according to port, slot, and in case a pad is not here, i return 0 (otherwise it appears).

pcsx2/PAD/Linux/state_management.cpp
```diff
@@ -176,6 +176,17 @@ u8 pad_poll(u8 value)
...
+	int padNum = -1;
+	if(query.port == 0 && query.slot == 0 && s_vgamePad.size() > 0) padNum = 0;
+	if(query.port == 0 && query.slot == 1 && s_vgamePad.size() > 2) padNum = 2;
+	if(query.port == 0 && query.slot == 2 && s_vgamePad.size() > 3) padNum = 3;
+	if(query.port == 0 && query.slot == 3 && s_vgamePad.size() > 4) padNum = 4;
+	if(query.port == 1 && query.slot == 0 && s_vgamePad.size() > 1) padNum = 1;
+	if(query.port == 1 && query.slot == 1 && s_vgamePad.size() > 5) padNum = 5;
+	if(query.port == 1 && query.slot == 2 && s_vgamePad.size() > 6) padNum = 6;
+	if(query.port == 1 && query.slot == 3 && s_vgamePad.size() > 7) padNum = 7;
+	if(padNum == -1) return 0;
```
lightningterror, line 188: Formatting should follow clang style.

arcum42:
Right now, this always has 8 gamepads, 0-7, where it should really only have 8 if both multitaps are on. Pretty much each port should either have one gamepad or four, depending on whether the multitap is on.

What we really should be doing is having an [2][4] array of gamepads, with the port and slot. (If both multitaps are on, you get gamepads 1A-1D and 2A-2D, basically.)

It's pretty much something I was already planning on, just a bit lower on priority since multitap's not that easy for me to test...

sonicfind:
Is that how linux works with PAD? PCSX2 just constantly has 8 pads available in the s_vgamePad vector? I don't have/run linux so I'm just going off the files directly, but I was thinking that it would only have as many pads in the vector as are connected to the system (which would bring its own separate problem to this PR).

arcum42:
The thing to remember with a pad plugin (not a plugin now, but...) is you have the actual controller on the computer, which, yes, would be in a vector with as many as are hooked up, and then the emulated ps2 controllers.

On linux right now, there are two emulated ps2 controllers. If multitap was implemented, if you turn on multitap on either, it would expand out to potentially four on that particular port. 

With this PR at the moment, it just always shows as if there were 8 emulated controllers on the ps2, even when multitap is off.

nadenislamarre:
@arcum42 this is not what happen. If no multitap IS set, query IS set only with slot 1 if each port. If multitap IS set, query IS set with thé four slot if thé port 1. Samedi for port 2 and multitap 2. It IS not ignoring thé multitap configuration. When multitap IS off, youre limited to 2 players.

nadenislamarre:
> @arcum42 this is not what happen. If no multitap IS set, query IS set only with slot 1 of each port. If multitap IS set, query IS set with thé four slots of thé port 1. Same for port 2 and multitap 2. It IS not ignoring thé multitap configuration. When multitap IS off, youre limited to 2 players.

GovanifY (approved):
LGTM, fine to merge by me!

pcsx2/PAD/Linux/state_management.cpp
```diff
@@ -176,6 +176,26 @@ u8 pad_poll(u8 value)
...
+	if (query.port == 1 && query.slot == 0 && s_vgamePad.size() > 1)
+		padNum = 1;
+	if (query.port == 1 && query.slot == 1 && s_vgamePad.size() > 5)
+		padNum = 5;
+	if (query.port == 1 && query.slot == 2 && s_vgamePad.size() > 6)
+		padNum = 6;
+	if (query.port == 1 && query.slot == 3 && s_vgamePad.size() > 7)
+		padNum = 7;
+	if (padNum == -1)
+		return 0;
```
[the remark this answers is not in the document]
sonicfind: Not sure about the test_bit part (don't see, at the moment, how we'd trim it down), but something else I'd really like to mention.
If this is the state it's gonna stay in for now, why are there no else ifs here?

refractionpcsx2:
Multitap is implemented in Qt (onepad) so this is no longer required. Thanks.

PR #4200, opened 2021-01-31
ps1 mode: Logging and old code cleanup

kenshen112:
This is a quick PR to remove the incorrect and since unused reset function for ps1 mode that was commented out in CDVD as well as to clean up log statements in CDVD and cdrom

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -1475,25 +1475,13 @@ static __fi void cdvdWrite0F(u8 rt)
 }
 
 static __fi void cdvdWrite14(u8 rt)
-{ // PS1 MODE?? // This should be done in the SBUS_F240 bit 19 write in HwWrite.cpp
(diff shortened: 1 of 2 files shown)


pcsx2/CDVD/CDVD.cpp
```diff
@@ -17,6 +17,7 @@
 #include "IopCommon.h"
 #include "AppConfig.h"
 
+#include <bitset>
```
refractionpcsx2, line 20: What was this added for?

kenshen112: Oops. When I was printing the binary values of the reg I used cout and bitset. I'll clean that up real quick

refractionpcsx2: plz learn to read hex ;)

kenshen112: It's been removed :p

PR #4205, opened 2021-02-04
Don't `Destroy()` modals

tellowkrinkle:
On macOS it prevents the parent window from regaining focus

Also, JoystickConfiguration had an incredibly complicated system for setting preferences as the user clicked checkboxes and undoing them if they clicked cancel.  Replaced with just setting preferences at the end if the user clicks `OK`


tellowkrinkle:
@arcum42 Could you test on Linux and make sure this doesn't break anything there

arcum42:
Doesn't seem like it broke anything on Linux. Having just an Ok button when the main dialog has Ok and Cancel, and the other sub dialogs also have Ok and Cancel feels inconsistent, though.

I'd suggest getting rid of the Ok button and also both Ok and Cancel on the left and right joystick config. You can then just click the close box to close them, and have everything governed by the Apply, Ok, Cancel on the main dialog.

tellowkrinkle:
Like this?

arcum42:
Yeah, that looks better and is more consistent.

PR #4211, opened 2021-02-04
GameDB: Add missing Speed Kings patches

RedDevilus:
https://forums.pcsx2.net/Thread-Fixing-unplayable-games?pid=609412#pid609412

I somehow forgot to add these from the forum. Whoops.


atomic83GitHub:
I think it's important to write in the patch comment that it is necessary because we don't have a vif fifo in pcsx2.

PR #4214, opened 2021-02-07
DEV9: Merge _DEV9*() functions into *Net(), and check ethEnable before calling InitNet()

TheLastRar:
Previously, the ethEnable config value was not checked before starting network RX net thread.

This PR merges the platform specific  ``_DEV9open()`` into``InitNet()``, ``_DEV9close()`` into ``TermNet()`` and adds a check before calling InitNet()

### DIFF
--- a/pcsx2/DEV9/DEV9.cpp
+++ b/pcsx2/DEV9/DEV9.cpp
@@ -219,7 +219,10 @@ s32 DEV9open(void* pDsp)
 			config.hddEnable = false;
 	}
 
-	return _DEV9open();
+	if (config.ethEnable)
+		InitNet();
+
+	return 0;
 }
 
 void DEV9close()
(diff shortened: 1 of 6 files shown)


AmFobes:
(With ethernet disabled)
Current Master:
image

PR:
image

👍 Everything seems to work fine.

pcsx2/DEV9/DEV9.cpp
```diff
@@ -1076,3 +1083,74 @@ void DEV9setLogDir(const char* dir)
...
+			}
+		}
+		else
+			InitNet();
+	}
+	else if (oldConfig.ethEnable)
+		TermNet();
+
+		//Hdd
+		//Hdd Validate Path
```
GovanifY, line 1111: Why two comments here?

GovanifY (commented):
LGTM, just curious about this

pcsx2/DEV9/DEV9.cpp
TheLastRar: THe `//Hdd` comment is supposed to indicate that this block of code deals with hdd, the following `Hdd Validate Path` is supposed to refer to the immediately following if statements (up to the `//Hdd Compare with old config` comment)

GovanifY (approved):

Issue #4151, opened 2021-01-17
[DEV9] ISPs that perform DNS hijacking can prevent games with resurrected servers from connecting


TheLastRar:
A number of ISPs will intercept all DNS communications to there own DNS server.

For server revival projects that relay on their own DNS (RE: Outbreak) or a community DNS (R&C, which uses Cristian DNS) this prevents the game from contacting the configured DNS, instead receiving a response form the ISP DNS server, which is unaware of the resurrected server's IP.

Prior to the DEV9 plugin merge, CLR_DEV9's winsock backend had an internal DNS server which could be be configured with the IP address of each resurrected servers IP manually, thus workaround the DNS redirection.

Currently there is no equivalent as far as I'm aware in PCSX2

PR #4220, opened 2021-02-09
DEV9: Add Internal DNS server

TheLastRar:
Using the CLR-DEV9 approach to resolve https://github.com/PCSX2/pcsx2/issues/4151

Depends on (and includes) #4301

This is a port of CLR-DEV9's DNS server, adapted to work with TAP and pcap (CLR-DEV9's was winsock only)

This PR adds an internal DNS server, that can be opted in by setting you DNS in the PS2 Network config to 192.0.2.1.
It features its own host list (DEV9Hosts.ini), allowing you to remap a url to a different IP address.
Note that while CLR-DEV9 used XML for its settings, this version uses PCSX2's iniInterface format for its host file.

If the internal DNS fails to find an url in its host list, is uses OS specific functions to look up the address.

The internal DNS will replace 127.0.0.1 with the PC's local IP (specifically the address assigned to the selected adapter), allowing PCSX2 to communicate with servers running on the same PC (i.e. developers of server revival projects may use this to allow developing on a single machine).

pcap may not fully work with 127.0.0.1 remapping, as it relies on your router to loop packets back to your PC.

Currently no UI is provided (as is the case with CLR-DEV9), and I'm not planning on providing one in this PR.

### DIFF
--- a/pcsx2/DEV9/DEV9Config.cpp
+++ b/pcsx2/DEV9/DEV9Config.cpp
@@ -1,0 +1,114 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2021  PCSX2 Dev Team
+ *
+ *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
+ *  of the GNU Lesser General Public License as published by the Free Software Found-
+ *  ation, either version 3 of the License, or (at your option) any later version.
+ *
+ *  PCSX2 is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+ *  PURPOSE.  See the GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along with PCSX2.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "PrecompiledHeader.h"
+
+#include "ghc/filesystem.h"
+#include <wx/fileconf.h>
+
+#include "DEV9.h"
+#include "gui/AppConfig.h"
+#include "common/IniInterface.h"
+
+#ifdef _WIN32
+#include "ws2tcpip.h"
+#elif defined(__POSIX__)
+#include <arpa/inet.h>
+#endif
+
+void SaveDnsHosts()
+{
+	std::unique_ptr<wxFileConfig> hini(OpenFileConfig(GetSettingsFolder().Combine(wxString("DEV9Hosts.ini")).GetFullPath()));
+	IniSaver ini((wxConfigBase*)hini.get());
+
+	for (size_t i = 0; i < config.EthHosts.size(); i++)
+	{
+		std::wstring groupName(L"Host" + std::to_wstring(i));
+		ScopedIniGroup iniEntry(ini, groupName);
+		ConfigHost entry = config.EthHosts[i];
+
+		wxString url(entry.Url);
+		ini.Entry(L"Url", url);
+		//TODO UTF8(?)
+		wxString desc(entry.Desc);
+		ini.Entry(L"Desc", desc);
+
+		char addrBuff[INET_ADDRSTRLEN];
+		inet_ntop(AF_INET, entry.Address, addrBuff, INET_ADDRSTRLEN);
+		wxString address(addrBuff);
+		ini.Entry(L"Address", address);
+
+		ini.Entry(L"Enabled", entry.Enabled);
+	}
+	ini.Flush();
+}
+
+void LoadDnsHosts()
+{
+	wxFileName iniPath = GetSettingsFolder().Combine(wxString("DEV9Hosts.ini"));
+	config.EthHosts.clear();
+	//If no file exists, create one to provice an example config
+	if (!iniPath.FileExists())
+	{
+		//Load Default settings
+		ConfigHost exampleHost;
+		exampleHost.Url = "www.example.com";
+		exampleHost.Desc = "Set DNS to 192.0.2.1 to use this host list";
+		memset(exampleHost.Address, 0, 4);
+		exampleHost.Enabled = false;
+		config.EthHosts.push_back(exampleHost);
+		SaveDnsHosts();
+		return;
+	}
+
+	std::unique_ptr<wxFileConfig> hini(OpenFileConfig(iniPath.GetFullPath()));
+	IniLoader ini((wxConfigBase*)hini.get());
+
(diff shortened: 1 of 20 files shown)


TheLastRar:
Marking this as WIP again as I've noted that there are a few more things I want to change

pcsx2/DEV9/DEV9Config.cpp
```diff
@@ -0,0 +1,114 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2020  PCSX2 Dev Team
```
lightningterror, line 2: 2021 for new files, there are others as well.

TheLastRar: I believe I've corrected all the copyright headers on all the new files

RedDevilus: image
Just this one left.

TheLastRar: Ah, just took the line `2021 for new files` and didn't look at the file the comment was made on.
I did also notice I missed another file, but that change I squashed.

TheLastRar: Actually, I got confused by the two similarly named files, I'll amend that tomorrow

TheLastRar: Okay, _now_ we should be good

pcsx2/DEV9/InternalServers/DNS_Logger.cpp
```diff
@@ -0,0 +1,203 @@
...
+		DNS_Packet dns(udpPayload->data, udpPayload->GetLength());
+		LogPacket(&dns);
+	}
+
+	void DNS_Logger::LogPacket(DNS_Packet* dns)
+	{
+		Console.WriteLn("DEV9: DNS: ID %i", dns->id);
+		Console.WriteLn("DEV9: DNS: Is Response? %s", dns->GetQR() ? "True" : "False");
+		const char* opName;
+		switch ((DNS_OPCode)dns->GetOpCode())
```
Warepire, line 50: To reduce function size of LogPacket this switch case (and the next) could be broken out to static functions returning `const char*`. (e.g. `DNS_OpCodeToString(opcode)`)

pcsx2/DEV9/InternalServers/DNS_Logger.cpp
```diff
@@ -0,0 +1,203 @@
...
+		}
+		for (size_t i = 0; i < dns->answers.size(); i++)
+		{
+			DNS_ResponseEntry entry = dns->answers[i];
+			Console.WriteLn("DEV9: DNS: Ans%i Name %s", i, entry.name.c_str());
+			Console.WriteLn("DEV9: DNS: Ans%i Type %i", i, entry.entryType);
+			Console.WriteLn("DEV9: DNS: Ans%i Class %i", i, entry.entryClass);
+			Console.WriteLn("DEV9: DNS: Ans%i TTL %i", i, entry.timeToLive);
+
+			char* strArr = new char[entry.data.size() * 4 + 1];
```
Warepire, line 144: Perhaps a helper function that does something like this to keep code duplication down and make sure memory is properly deallocated?
```
std::string str;
//str.reserve(entry.data.size() * 4 + 1);  -- If this is a hot path, otherwise let std::string deal with the allocation
for (const auto& d : entry.data) {
  str += std::to_string(d) + ":";
}
return str;
```

pcsx2/DEV9/InternalServers/DNS_Server.cpp
```diff
@@ -0,0 +1,435 @@
...
+			outstandingQueries--;
+			return retPay;
+		}
+		return nullptr;
+	}
+
+	bool DNS_Server::Send(UDP_Packet* payload)
+	{
+		PayloadPtr* udpPayload = static_cast<PayloadPtr*>(payload->GetPayload());
+		DNS_Packet* dns = new DNS_Packet(udpPayload->data, udpPayload->GetLength());
```
Warepire, line 164: Would this and other allocations benefit from std::shared_ptr or std::unique_ptr to ensure that the memory is properly managed across all paths?

pcsx2/DEV9/InternalServers/DNS_Logger.h
```diff
@@ -30,6 +30,9 @@ namespace InternalServers
 		void InspectSend(PacketReader::IP::IP_Payload* payload);
 
 	private:
+		std::string VectorToString(std::vector<u8> data);
```
Warepire, line 33: If you pass the vector by const reference, you won't make a copy of it., e.g. `const std::vector<u8>& data`

pcsx2/DEV9/InternalServers/DNS_Logger.cpp
```diff
@@ -42,84 +42,87 @@ namespace InternalServers
...
-	void DNS_Logger::LogPacket(DNS_Packet* dns)
+	std::string DNS_Logger::VectorToString(std::vector<u8> data)
 	{
-		Console.WriteLn("DEV9: DNS: ID %i", dns->id);
-		Console.WriteLn("DEV9: DNS: Is Response? %s", dns->GetQR() ? "True" : "False");
-		const char* opName;
-		switch ((DNS_OPCode)dns->GetOpCode())
+		if (data.size() != 0)
+		{
+			std::string str;
```
Warepire, line 49: You can utilize RVO / Copy elision ( https://en.cppreference.com/w/cpp/language/copy_elision ) if you declare the string outside the if-statement and always return it.
An empty string is equivalent to `""`

pcsx2/DEV9/InternalServers/DNS_Server.cpp
TheLastRar: This particular line didn't need to be an allocation and has been corrected

At a glance I'm not sure what would most benefit from using smart pointers.

pcsx2/DEV9/InternalServers/DNS_Logger.h
TheLastRar: I always forget about reference (and apparently also forgot to make it const)

pcsx2/DEV9/InternalServers/DNS_Server.cpp
Warepire: Re-checking I don't see anything obvious beyond the line I commented on. You may resolve this.

RedDevilus (approved):

Dnawrkshp:
Hey this feature is extremely useful and I'd love to see this merged. Is there a timeline for when it will be?

PR #4222, opened 2021-02-14
PAD (Windows): Remove DirectInput and Force Hide cursor options. Clean up includes/remove old code

MrCK1:
- This PR will remove the DirectInput API option for PC mice and keyboards. It also gets rid of the "Force Hide" cursor option which is buggy and superseeded by the same option already present in the GS Window options. I ended up taking the time to adjust the GUI a bit but that can be left out of the PR if we don't want that in here.

- Tried to tidy up some of the library includes by moving them to the Global.h header (similar structure to SPU2). Also removed some unused includes and code in certain files.

image


GovanifY (approved):
LGTM by me!

CookiePLMonster:
Can DInput for keyboards go too? It kinda only makes sense for gamepads/controllers.

MrCK1:
> 
> 
> Can DInput for keyboards go too? It kinda only makes sense for gamepads/controllers.

I forgot this was still open, I can run it by the rest of the team and I'll remove it if we're OK with that as well. Shouldn't be any issue removing it anyway

MrCK1:
PR has been rebased again with requested changes from @CookiePLMonster . Here is the current GUI state

image

RedDevilus (approved):
Fix the alignment 🐃 and it's fine as-is

PR #4223, opened 2021-02-18
PGIF- Fix Castlevania SoTN

Nobbs66:
PS1DRV handles bit 28 of GPU Status register, not PGIF. This fixes several games.


refractionpcsx2:
Can you list the games?

kenshen112:
Twisted Metal and Symphony of the night for starters

refractionpcsx2:
okay it would just be nice to have a list of games for the report.

Nobbs66:
Castlevania: Symphony of the Night

Twisted Metal

Pink Panther

I'd imagine that the number of games that this helps is rather small, but the games I listed are picky about things and require the fix. The change also seems to be more accurate based on one of wisi's code comments

PR #4226, opened 2021-02-19
SPU2: Resample reverb input/output.

Ziemas:
The reverb runs at half the the samplerate of the rest of the SPU so a change of samplerate is required. They way this is currently done is by decimating when downsampling and duplicating samples when upsampling without doing any filtering.

This adds proper filtering to this process,  the filter coefficients are from mednafen but I *think* they were originally found by Neill Corlett.

I have not yet compared to real PS2 behavior and don't know when I'll get around to so I'm PR'ing this now.


Ziemas:
Master on the left PR on the right

PR #4227, opened 2021-02-20
SPU2: Purge Replay Utility

MrCK1:
The utility was intended to only work with SPU2 as a plugin module and is no longer maintained.


refractionpcsx2:
You need to purge all the SPU2_dummy_callback stuff too since that's related to the replay stuff

MrCK1:
> 
> 
> You need to purge all the SPU2_dummy_callback stuff too since that's related to the replay stuff

Changes pushed, thanks.

PR #4234, opened 2021-02-21
IOP: Execute breakpoints for the interpreter

AmFobes:
Added as a compile-time option, it's pretty friggin slow.


kenshen112:
On inital testing seems to work

kenshen112:
memchecks seem functional as well

PR #4235, opened 2021-02-21
SPU2/SDL: Remove intermediate buffer and write samples directly to the

Oschowa:
buffer provided by the SDL callback.

I at least can't see a reason for it existing and it's working fine without it.


Ziemas:
:+1: 

I guess it was done to use `SDL_MixAudio` but that is indeed not necessary here.

PR #4245, opened 2021-02-23
GameDB: Fix KOF98 patches

kozarovv:
Added check for NeoGeo mode, fix graphic in 2D mode.

### DIFF
--- a/bin/GameIndex.yaml
+++ b/bin/GameIndex.yaml
@@ -16610,8 +16610,12 @@ SLES-55280:
         patch=1,EE,000ffc24,word,3442a833
         patch=1,EE,000ffc28,word,08077ed2
         patch=1,EE,000ffc2c,word,ae020098
-        patch=1,EE,001Dfb40,word,0803ff08
-        patch=1,EE,001dfb44,word,3c0200ff
+        patch=1,EE,e0020000,extended,0052f1e0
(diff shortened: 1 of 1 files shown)


Silanda:
I only have the PAL version to test, but it's still not working here. Characters appear to be being drawn behind the background (they're partially visible in parts of certain backgrounds).

kozarovv:
Ok, so i can't fix it using patches. Game have 3 modes, 3D/2D/NeoGeo, with patch that fix China (boat) stage NeoGeo/2D mode are broken. Check for NeoGeo mode don't work for 2D stages, and check for 2D stages will make NeoGeo mode broken again. There is no good way to detect China stage because in 2D mode the same address is used. So i will remove those patches at all, I guess that will be better. Other games are unaffected because only KOF98 have that additional patch. I'm closing this, and after work i will make pr to remove KOF98 patches. 

I don't see other way than another HW hack in GSDX to make this work.

refractionpcsx2:
@kozarovv an alternative solution you can try. see if you can find the part of the GS packets which set the ZBUF register to set the ZBuffer format to 32bit and change it to use Z16 format instead, that should make it ignore the upper 16 bits in all cases.  If that doesn't work you can try Z24 (same memory layout as Z32 but ignores the upper 8 bits), that might be just enough.

PR #4249, opened 2021-02-24
GameDB: Fix KOF98 patches

kozarovv:
Update KOF98 patches to fix remaining issues.

### DIFF
--- a/bin/GameIndex.yaml
+++ b/bin/GameIndex.yaml
@@ -16595,23 +16595,26 @@ SLES-55280:
   patches:
     E7516A6E:
       content: |-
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
@Silanda can you test the EU patch? the menu looks broken (not too bad if you choose 60hz) but the rest of the games should be okay

Silanda:
I've only had a quick test, but it looks good here! I'll try to do a complete playthrough when I get the chance.

refractionpcsx2:
ok well I'll merge for now, we can revisit it later if there's a problem :)

PR #4250, opened 2021-02-24
Core: Preserve read only bits in the COP0 config register

AmFobes:
The read-only bits in question are the Instruction cache size (IC) and the Data cache size (DC)

Before this pr, MTC0 would freely write to the entire config register.

### DIFF
--- a/pcsx2/COP0.cpp
+++ b/pcsx2/COP0.cpp
@@ -41,10 +41,14 @@ void __fastcall WriteCP0Status(u32 value) {
 	//DMA_LOG("COP0 Status write = 0x%08x", value);
 
 	cpuRegs.CP0.n.Status.val = value;
-    cpuUpdateOperationMode();
     cpuSetNextEventDelta(4);
 }
 
+void __fastcall WriteCP0Config(u32 value) {
+	// Protect the read-only ICacheSize (IC) and DataCacheSize (DC) bits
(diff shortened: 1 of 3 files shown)


pcsx2/COP0.cpp
```diff
@@ -45,6 +45,13 @@ void __fastcall WriteCP0Status(u32 value) {
     cpuSetNextEventDelta(4);
 }
 
+void __fastcall WriteCP0Config(u32 value) {
+	// Protect the read-only ICacheSize (IC) and DataCacheSize (DC) bits
+	cpuRegs.CP0.n.Config = value & ~0xFC0;
+	cpuRegs.CP0.n.Config |= 0x440;
+	cpuUpdateOperationMode();
+	cpuSetNextEventDelta(4);
```
refractionpcsx2, line 53: I don't think it needs these 2 last lines. the top one of the two doesn't do anything anyway.

AmFobes: Should I purge those two calls for the status register as well or leave them be? Just wanted to be consistent.

refractionpcsx2: the cpuSetNextEventDelta needs to be there for the status register because of interrupts, but you don't need it for the Config register.

refractionpcsx2: the cpuUpdateOperationsMode thing can probably just go.. lol

PR #4252, opened 2021-02-24
Debugger: Add "Go to address" for both disasm and memory view context menus

AmFobes:
I also implemented the previously unimplemented "Toggle breakpoint" context menu item.

Also, note #3483 

Examples below:

image

image


Squall-Leonhart:
any way to add the ability to point registers to a certain memory address?

AmFobes:
You are able to overwrite the value of registers if that's what you're asking.

PR #4260, opened 2021-02-28
DEV9: Move ATA write queue logic into its own class

TheLastRar:
This will be useful for porting other CLR_DEV9 features into pcsx2

Also adds the ability for multiple threads to add to the queue, as needed by the DNS pr

### DIFF
--- a/pcsx2/DEV9/ATA/Commands/ATA_CmdDMA.cpp
+++ b/pcsx2/DEV9/ATA/Commands/ATA_CmdDMA.cpp
@@ -62,7 +62,11 @@ void ATA::DRQCmdDMADataFromHost()
 }
 void ATA::PostCmdDMADataFromHost()
 {
-	QueueWrite(currentWriteSectors, currentWrite, currentWriteLength);
+	WriteQueueEntry entry{0};
+	entry.data = currentWrite;
+	entry.length = currentWriteLength;
+	entry.sector = currentWriteSectors;
+	writeQueue.Enqueue(entry);
 	currentWrite = nullptr;
 	currentWriteLength = 0;
 	currentWriteSectors = 0;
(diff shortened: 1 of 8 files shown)


pcsx2/DEV9/SimpleQueue.h
```diff
@@ -0,0 +1,95 @@
+#pragma once
```
lightningterror, line 1: File needs copyright header.

TheLastRar: Woops, added

pcsx2/DEV9/ATA/Commands/ATA_CmdDMA.cpp
```diff
@@ -62,7 +62,11 @@ void ATA::DRQCmdDMADataFromHost()
 }
 void ATA::PostCmdDMADataFromHost()
 {
-	QueueWrite(currentWriteSectors, currentWrite, currentWriteLength);
+	WriteQueueEntry entry;
+	entry.data = currentWrite;
+	entry.length = currentWriteLength;
+	entry.sector = currentWriteSectors;
+	writeQueue.Enqueue(entry);
```
refractionpcsx2, line 69: entry.next not initialised

TheLastRar: Reiterating what was said in the discord just for reference;

entry is of type WriteQueueEntry (defined within class ATA), this pr removed the next field, so this seems like a false positive.

I noticed that SimpleQueue also defines a structure called WriteQueueEntry (defined within the class) which does have a field called next.

In theory this shouldn't cause issues (as they are defined within there own respective classes), but I've renamed one of them

TheLastRar: Zero initializing the whole struct made codacy happy

pcsx2/DEV9/SimpleQueue.h
```diff
@@ -0,0 +1,110 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2020  PCSX2 Dev Team
```
lightningterror, line 2: 2021

TheLastRar: Corrected

PR #4261, opened 2021-02-28
GSdx: Switch tooltips to unicode

AmFobes:
I didn't test the tooltips on #4188 (oops)

### DIFF
--- a/plugins/GSdx/Window/GSSetting.cpp
+++ b/plugins/GSdx/Window/GSSetting.cpp
@@ -25,30 +25,38 @@
 #include "resource.h"
 #endif
 
-const char* dialog_message(int ID, bool* updateText) {
+#ifdef _WIN32
+#define cvtString(s) L##s
+const wchar_t* dialog_message(int ID, bool* updateText)
+{
+#else
+#define cvtString(s) s
+const char* dialog_message(int ID, bool* updateText)
+{
+#endif
 	if (updateText)
 		*updateText = true;
 	switch (ID)
 	{
 		case IDC_FILTER:
-			return "Control the texture filtering of the emulation.\n\n"
+			return cvtString("Control the texture filtering of the emulation.\n\n"
 				"Nearest:\nAlways disable interpolation, rendering will be blocky.\n\n"
 				"Bilinear Forced (excluding sprite):\nAlways enable interpolation except for sprites (FMV/Text/2D elements)."
 				" Rendering is smoother but it could generate a few glitches. If upscaling is enabled, this setting is recommended over 'Bilinear Forced'\n\n"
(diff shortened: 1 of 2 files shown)


plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -25,30 +25,36 @@
 #include "resource.h"
 #endif
 
-const char* dialog_message(int ID, bool* updateText) {
+#ifdef _WIN32
+#define _cvtString( s ) L##s
```
turtleli, line 29: Leading underscores shouldn't be used. They're reserved for use by the standard for implementation use.

GovanifY (approved):

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -25,30 +25,36 @@
 #include "resource.h"
 #endif
 
-const char* dialog_message(int ID, bool* updateText) {
+#ifdef _WIN32
+#define cvtString( s ) L##s
+const wchar_t* dialog_message(int ID, bool* updateText) {
+#else
+#define cvtString( s ) s
+const char* dialog_message(int ID, bool* updateText){
```
lightningterror, line 33: Formatting can be better on these lines, above ones too.

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -215,6 +221,7 @@ const char* dialog_message(int ID, bool* updateText) {
 		default:
 			if (updateText)
 				*updateText = false;
-			return "";
+			return cvtString("");
 	}
 }
+#undef cvtString
```
turtleli, line 227: Is this line necessary?

AmFobes: Not necessarily although intellisense didn't complain when I tried to use the define elsewhere which led me to believe that it'll clutter up stuff.

turtleli: Ok. I haven't coded for a while and don't recall what's best. I guess it might be better to leave it there.

RedDevilus (approved):

plugins/GSdx/Window/GSSetting.cpp
```diff
@@ -25,30 +25,36 @@
 #include "resource.h"
 #endif
 
-const char* dialog_message(int ID, bool* updateText) {
+#ifdef _WIN32
+#define cvtString(s) L##s
+const wchar_t* dialog_message(int ID, bool* updateText) {
+#else
+#define cvtString(s) s
+const char* dialog_message(int ID, bool* updateText){
```
lightningterror, line 33: You can also move the brackets to new line for consistency.

lightningterror (approved):

PR #4263, opened 2021-03-01
IPC: Add savestate support & more

GovanifY:
https://github.com/PCSX2/pcsx2/pull/4262 needs to be merged first, a recent version of pcsx2_ipc needs to be used to get savestates support.


GovanifY:
I scope creeped the PR a bit and added GetGame{ID,UUID,Title} commands

GovanifY:
This PR also fixes a bug that sneaked around in https://github.com/PCSX2/pcsx2/pull/4262

RedDevilus (approved):
I don't understand it fully, but it's an improvement which is what matters.

PR #4267, opened 2021-03-03
IPC: add a slot system

GovanifY:
This allows multiple emulator sessions to run at the same time and windows user to configure their used port to their preferred one.

While slotting looks like ports they are not, they are simply used for compatiblity on systems that use TCP sockets to communicate, on linux for example unix sockets are used and the slot is transferred into the name of the socket.


GovanifY:
> Missing a space

already solved, thanks for the pre-pr review!

PR #4269, opened 2021-03-05
vs/deps: Workaround patch for libsamplerate on VS 16.9

xTVaser:
Taken from https://github.com/microsoft/vcpkg/pull/14348

I believe this is also fixed in the latest version `0.2.1` so likely good followup work (it removes float_case entirely) but I wasn't able to get VS to build it, atleast this gets things back to a build-able state in the interim.


xTVaser:
closing in favor of https://github.com/PCSX2/pcsx2/pull/4277

PR #4270, opened 2021-03-05
gsdx-d3d11: Cleanup gsdevice11.cpp warnings reported on codacy.

lightningterror:


plugins/GSdx/Renderers/DX11/GSDevice11.cpp
```diff
@@ -1468,7 +1467,7 @@ void GSDevice11::CreateShader(std::vector<char> source, const char* fn, ID3DIncl
 	}
 }
 
-void GSDevice11::CreateShader(std::vector<char> source, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11PixelShader** ps)
+void GSDevice11::CreateShader(const std::vector<char> source, const char* fn, ID3DInclude *include, const char* entry, D3D_SHADER_MACRO* macro, ID3D11PixelShader** ps)
```
turtleli, line 1470: Would `const std::vector<char>&` be better?

sonicfind: Probably, since it'd use up space equivalent to a pointer rather than an entire new vector object.
Codacy would most likely state to make it so anyways.

Squall-Leonhart:
this needs thorough testing, especially with the save state system, last time i saw someone "fix warnings" in an emulator they broke sound following a load state.

refractionpcsx2:
> this needs thorough testing, especially with the save state system, last time i saw someone "fix warnings" in an emulator they broke sound following a load state.

None of this information is stored in savestates.

lightningterror:
Maybe I'll exclude the m_state changes.

PR #4271, opened 2021-03-06
GS/Core: Implement the new generation of GS Debugger

GovanifY:
This is a full in-core reimplementation of GSDumpGUI. 
This is the biggest blocker of https://github.com/PCSX2/pcsx2/discussions/4100, and of any plugin merges by far.

It is a _much_ simpler reimplementation of the past debugger designed with expandability in mind, while still keeping at least feature parity.

~~This PR is not fully finalized but it's close to completion, notable pain points are:~~

~~- [x] The GS is currently being tanked by our thread, leaving it unable to display anything when running a dump, this is the biggest breaking point of the implementation and needs to be figured out ASAP.~~
~~- [x] The code overall design is very rough, I concentrated on making a PoC as fast as possible, so there are a few areas in this implementation where the code quality is just not good.~~
~~- [x] The UI/UX, FlexSizers do not behave according to wxWidgets documentation, as such the GUI looks horrendous and segfaults on stop.~~
~~- [x] I noticed some of the reg parsing treeview values are parsed as integers instead of their string representation, I need to look at ParseTreeReg and see which ones I forgot.~~

~~Please review this PR and feel free to leave comments, this is a high priority PR and I'm simply clueless regarding some issues, such as FlexSliders.~~

This was much more painful than you'd expect.

READY FOR QA!!!


pcsx2/gui/App.h
```diff
@@ -122,6 +122,7 @@ enum MenuIdentifiers
 	MenuId_GameSettingsSubMenu,
 	MenuId_EnablePatches,
 	MenuId_EnableCheats,
+    MenuId_GSDump,
```
lightningterror, line 125: Formatting.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,806 @@
...
+				m_gif_list->SetItemText(rootId, s);
+				rootId = m_gif_list->AppendItem(mainrootId, "VSync");
+			}
+			else 
+				m_gif_list->AppendItem(rootId, s);
+		i++;
+	}
+	m_gif_list->Delete(rootId);
+}
+
```
lightningterror, line 291: Extra row.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,806 @@
...
+			wxString s;
+			([&] { return element.id == Transfer; })() ? (s.Printf("%d - %s - %s - %d byte", i, GSTypeNames[element.id], GSTransferPathNames[element.path], element.length)) : 
+									s.Printf("%d - %s - %d byte", i, GSTypeNames[element.id], element.length);
+			if (element.id == VSync)
+			{
+				m_gif_list->SetItemText(rootId, s);
+				rootId = m_gif_list->AppendItem(mainrootId, "VSync");
+			}
+			else 
+				m_gif_list->AppendItem(rootId, s);
```
lightningterror, line 286: Extra tab, same goes for above in this function.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,806 @@
...
+}
+
+void Dialogs::GSDumpDialog::ToStart(wxCommandEvent& event)
+{
+	m_button_events.push_back(GSEvent{RunCursor, 0});
+}
+
+void Dialogs::GSDumpDialog::GenPacketList(std::vector<GSData>& dump)
+{
+	int i = 0;
```
lightningterror, line 271: Maybe move it inside the loop since it's only used inside?

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,806 @@
...
+			}
+			if (i >= m_root_window->m_dump_packets.size())
+				i = 0;
+		}
+	}
+
+	GetCorePlugins().Shutdown();
+	OnStop();
+	return;
+}
```
lightningterror, line 806: I'd add 1 more line for end of line, I think most files do the same.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,806 @@
...
+			if (packed)
+				s.Printf("F = %u", (u32)((data.hi >> 36) & ((u64)(1 << 8) - 1)));
+			else
+				s.Printf("F = %u", (u32)((data.lo >> 56) & ((u64)(1 << 8) - 1)));
+			m_gif_packet->AppendItem(rootId, s);
+			break;
+		}
+		case AD:
+		{
+			wxString s;
```
lightningterror, line 572: Can be moved inside the if.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,806 @@
...
+
+	if (!dump_file.IsOk())
+	{
+		OnStop();
+		return;
+	}
+
+	u32 crc = 0, ss = 0;
+	// XXX: check the numbers are correct
+	int renderer_override = m_root_window->m_renderer_overrides->GetSelection();
```
lightningterror, line 660: const

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,806 @@
...
+	if (!GetCorePlugins().DoFreeze(PluginId_GS, 0, &fd, true))
+		m_running = false;
+	GSvsync(1);
+	GSreset();
+	GSsetBaseMem((void*)regs);
+	GetCorePlugins().DoFreeze(PluginId_GS, 0, &fd, true);
+
+	size_t i = 0;
+	size_t RunTo = 0;
+	size_t debug_idx = 0;
```
lightningterror, line 742: This and the 2 above can be moved further in.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
GovanifY: Yes it is only used inside because it's a counter, if I put it inside the loop it'll reset to 0 at each iteration

pcsx2/gui/Dialogs/GSDumpDialog.cpp
GovanifY: no they cannot, otherwise they'll be reinit to 0 at each iteration of the loop

lightningterror: Ah right, forgot about that.

tellowkrinkle:
`std::vector<wxString> v(4);` will default initialize 4 wxStrings so you don't have to input 4 empty strings like you're doing now

GovanifY:
> 
> 
> `std::vector<wxString> v(4);` will default initialize 4 wxStrings so you don't have to input 4 empty strings like you're doing now

Thanks! completely forgot that was a thing, any other opinions on the code? I feel like some parts are pretty dirty

GovanifY:
The layout now seems pretty functional to me and I fixed the shutdown segfault, discussing about layout ideas to further improve the tool's design before fixing the last GS API related bug

pcsx2/gui/Dialogs/ModalPopups.h
```diff
@@ -20,54 +20,344 @@
...
+X(BITBLTBUF,0x50) \
+X(TRXPOS,0x51) \
+X(TRXREG,0x52) \
+X(TRXDIR,0x53) \
+X(HWREG,0x54) \
+X(SIGNAL,0x60) \
+X(FINISH,0x61) \
+X(LABEL,0x62)
+
+#define X(name, value) name = value,
```
tellowkrinkle, line 83: I think it'll be a bit more obvious what's going on if you do
```c
#define GSDUMP_GIFREG(X) \
	X(PRIM, 0x00)
```
etc

Then, don't define the `X` until you're about to instantiate, at which point you do

```c
enum GSDUMP_GIFREG_NAME : GSDUMP_GIFREG_TYPE
{
#define ENTRY(name, value) name = value,
 	GSDUMP_GIFREG(ENTRY)
#undef ENTRY
};
```

Also I would like it if you indent the entries and align the numbers like
```c
#define GSDUMP_GIFREG(X) \
	X(PRIM,       0x00) \
	X(RGBAQ,      0x01) \
	X(ST,         0x02) \
	X(UV,         0x03) \
	X(XYZF2,      0x04) \
	X(XYZ2,       0x05) \
	X(TEX0_1,     0x06) \
	X(TEX0_2,     0x07) \
	X(CLAMP_1,    0x08) \
	X(CLAMP_2,    0x09) \
	X(FOG,        0x0a) \
	X(XYZF3,      0x0c) \
	X(XYZ3,       0x0d) \
	X(AD,         0x0e) \
	X(NOP,        0x0f) \
	X(TEX1_1,     0x14) \
	X(TEX1_2,     0x15) \
	X(TEX2_1,     0x16) \
	X(TEX2_2,     0x17) \
	X(XYOFFSET_1, 0x18) \
	X(XYOFFSET_2, 0x19) \
	X(PRMODECONT, 0x1a) \
	X(PRMODE,     0x1b) \
	X(TEXCLUT,    0x1c) \
	X(SCANMSK,    0x22) \
	X(MIPTBP1_1,  0x34) \
	X(MIPTBP1_2,  0x35) \
	X(MIPTBP2_1,  0x36) \
	X(MIPTBP2_2,  0x37) \
	X(TEXA,       0x3b) \
	X(FOGCOL,     0x3d) \
	X(TEXFLUSH,   0x3f) \
	X(SCISSOR_1,  0x40) \
	X(SCISSOR_2,  0x41) \
	X(ALPHA_1,    0x42) \
	X(ALPHA_2,    0x43) \
	X(DIMX,       0x44) \
	X(DTHE,       0x45) \
	X(COLCLAMP,   0x46) \
	X(TEST_1,     0x47) \
	X(TEST_2,     0x48) \
	X(PABE,       0x49) \
	X(FBA_1,      0x4a) \
	X(FBA_2,      0x4b) \
	X(FRAME_1,    0x4c) \
	X(FRAME_2,    0x4d) \
	X(ZBUF_1,     0x4e) \
	X(ZBUF_2,     0x4f) \
	X(BITBLTBUF,  0x50) \
	X(TRXPOS,     0x51) \
	X(TRXREG,     0x52) \
	X(TRXDIR,     0x53) \
	X(HWREG,      0x54) \
	X(SIGNAL,     0x60) \
	X(FINISH,     0x61) \
	X(LABEL,      0x62)
```

pcsx2/gui/Dialogs/ModalPopups.h
```diff
@@ -20,54 +20,344 @@
...
+			"","","","","","","",
+			"PSMT4HH",
+			"","","",
+			"PSMZ32",
+			"PSMZ24",
+			"PSMZ16",
+			"","","","","","","",
+			"PSMZ16S"
+		};
+        // clang-format on
```
tellowkrinkle, line 310: formatting

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,800 @@
...
+				case Path2:
+				{
+					GSgifTransfer2((u32*)event.data, event.length / 16);
+					break;
+				}
+				case Path3:
+				{
+					GSgifTransfer3((u32*)event.data, event.length / 16);
+					break;
+				}
```
tellowkrinkle, line 225: Braces not needed here

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,800 @@
...
+							for (u32 i = 0; i < nreg; i++)
+							{
+								u128 reg_data;
+								reg_data.lo =  *(u64*)(dump.data + p);
+								ParseTreeReg(regId, (GIFReg)((regs >> (i * 4)) & ((u64)(1 << 4) - 1)), reg_data, false);
+								p += 8;
+							}
+						}
+						break;
+					}
```
tellowkrinkle, line 375: Braces not needed for these cases

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,800 @@
...
+				wxString s;
+				s.Printf("ReadFIFO2: Size = %d byte", dump.length);
+				m_gif_packet->AppendItem(rootId, s);
+				break;
+			}
+			case Registers:
+			{
+				m_gif_packet->AppendItem(rootId, "Registers");
+				break;
+			}
```
tellowkrinkle, line 407: Braces not needed here

tellowkrinkle (commented):
I didn't mark all of them, but I don't think we need to include braces on the switch cases that don't declare variables

Don't know others' opinions on this though, it's not in our style guide

pcsx2/gui/Dialogs/ModalPopups.h
```diff
@@ -20,54 +20,344 @@
...
+		};
+		static constexpr const char* TEXCSMNames[256] = {
+			"CSM1",
+			"CSM2"
+		};
+		// a GNU extension exists to initialize array at given indices which would be 
+		// exactly what we need here but, obviously, MSVC is at it again to make our 
+		// life harder than sandpaper on your skin, so we make do
+		// clang-format off
+		static constexpr const char* TEXCPSMNames[256] = {
```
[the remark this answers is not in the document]
GovanifY: "" and nullptr are both null terminated strings with no chars, so they resolve to the same thing if printed using %s unless i'm mistaken.
Also while your solution is an elegant solution it's more verbose than what I did in this case, so I think I'll keep my current implementation, fine by you?

pcsx2/gui/Dialogs/GSDumpDialog.cpp
GovanifY: Actually I'm curious about your thoughts regarding all this, if a case requires a variable assignment, shouldn't we make all cases use brackets for readability sake? That's why I kept using brackets within the whole file

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,798 @@
...
+				s.Printf("Transfer Path %s", GSTransferPathNames[dump.path]);
+				trootId = m_gif_packet->AppendItem(rootId, s);
+				u64 tag = *(u64*)(dump.data);
+				u64 regs = *(u64*)(dump.data + 8);
+				u8 nloop = tag & ((u64)(1 << 15) - 1);
+				u8 eop = (tag >> 15) & 1;
+				u8 pre = (tag >> 46) & 1;
+				u32 prim = (tag >> 47) & ((u64)(1 << 11) - 1);
+				u8 flg = ((tag >> 58) & 3);
+				u32 nreg = (u32)((tag >> 60) & ((u64)(1 << 4) - 1));
```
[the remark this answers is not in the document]
GovanifY: oh, duh, thanks!

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,798 @@
...
+{
+	int i = 0;
+	m_gif_list->DeleteAllItems();
+	wxTreeItemId mainrootId = m_gif_list->AddRoot("root");
+	wxTreeItemId rootId = m_gif_list->AppendItem(mainrootId, "0 - VSync");
+	for (auto& element : dump)
+	{
+		wxString s;
+		([&] { return element.id == Transfer; })() ? (s.Printf("%d - %s - %s - %d byte", i, GSTypeNames[element.id], GSTransferPathNames[element.path], element.length)) : 
+									s.Printf("%d - %s - %d byte", i, GSTypeNames[element.id], element.length);
```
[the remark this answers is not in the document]
GovanifY: I tried to do that but ran into some issue I couldn't remember, that's the whole reason I used a lambda, probably had a brainfart, will look at it again, thanks for pointing it out to me!

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,784 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2010  PCSX2 Dev Team
```
AmFobes, line 2: 2002-2010 ?

pcsx2/gui/Dialogs/GSDumpDialog.cpp
GovanifY: went on and removed unneeded braces, if someone complains i'll put them back

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,784 @@
...
+					case GIF_FLG_IMAGE2:
+					{
+						wxString z;
+						s.Printf("IMAGE %d bytes", nloop * 16);
+						m_gif_packet->AppendItem(regId, z);
+						break;
+					}
+				}
+
+				//m_gif_packet->AppendItem(trootId, s);
```
AmFobes, line 370: Don't forget about this comment.

AmFobes (commented):
Pointed out two things that I quickly noticed.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
GovanifY: don't we all prefer those simpler times?

pcsx2/gui/Dialogs/ModalPopups.h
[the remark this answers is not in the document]
GovanifY: I'll mark this as resolved for now, feel free to reopen/comment!

pcsx2/gui/Dialogs/ModalPopups.h
GovanifY: good idea! For the def/undef X and using it as an argument I feel like it's a matter of personal preference but I don't have anything against that.

Thanks for doing the indentation, definitely more readable this way :)

tadanokojin (commented):
small non-code review:

**There's some weird behavior with loading dumps from the list:**
First I tried loading a dump from jak 2 I had laying around, that didn't work. It showed this nifty little error message:
```
[wx] can't open file '' (error 0: the operation completed successfully.)
```
So I tried another dump I had laying around and that didn't work either for the same reason. So I figured I would try thinking like a govi and dumping from a game directly since I figured this is what you did to test it and that does work.

So I thought "okay, maybe it can't find the associated screenshot and govi didn't account for this" but it seems like now the second dump I tried works without a screenshot but the jak dump I had still doesn't work. So no clue what is going on there.

**GSDumpGui Feature Parity**
One thing I noticed that's missing from GSDumpGui is that I have to restart the entire emulator in order to load a new dump in.

You might wanna check [these docs](https://docs.wxwidgets.org/3.0/classwx_file_system_watcher.html)

It might be a good idea not to add this additional complexity until we sort out the freezing, though.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,768 @@
...
+		{
+			while (i < (m_root_window->m_dump_packets.size()-1))
+			{
+				m_root_window->ProcessDumpEvent(m_root_window->m_dump_packets[i++], regs);
+
+				if (m_root_window->m_dump_packets[i].id == VSync)
+					break;
+			}
+			if (i >= m_root_window->m_dump_packets.size())
+				i = 0;
```
tadanokojin, line 761: think about this code with me:

while loop: you compare `i < size -1`
if statement: you compare `i >= size`

so what if `i ==  size - 1`?

Am I misunderstanding something or did you just write an infinite loop?

GovanifY: The original code did that too, am I misunderstanding something wrt the inner workings of this loop?
```csharp
                                while (gs_idx < dump.Data.Count)
                                {
                                    GSData itm = dump.Data[gs_idx++];
                                    CurrentGIFPacket = itm;
                                    Step(itm, pointer);

                                    if (gs_idx < dump.Data.Count && dump.Data[gs_idx].id == GSType.VSync)
                                        break;
                                }

                                if (gs_idx >= dump.Data.Count) gs_idx = 0;
```

tadanokojin: that code doesn't subtract by 1 in one comparison and not in the other.

GovanifY:
> So I thought "okay, maybe it can't find the associated screenshot and govi didn't account for this" but it seems like now the second dump I tried works without a screenshot but the jak dump I had still doesn't work. So no clue what is going on there.

I fixed that

GovanifY:
> **GSDumpGui Feature Parity**
> One thing I noticed that's missing from GSDumpGui is that I have to restart the entire emulator in order to load a new dump in.

Done

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,778 @@
...
+	, m_gif_list(new wxTreeCtrl(this, ID_SEL_PACKET, wxDefaultPosition, wxSize(400, 300), wxTR_HIDE_ROOT | wxTR_HAS_BUTTONS | wxTR_LINES_AT_ROOT))
+	, m_gif_packet(new wxTreeCtrl(this, wxID_ANY, wxDefaultPosition, wxSize(400, 300), wxTR_HIDE_ROOT | wxTR_HAS_BUTTONS | wxTR_LINES_AT_ROOT))
+	, m_start(new wxButton(this, ID_RUN_START, _("Go to Start"), wxDefaultPosition, wxSize(150,50)))
+	, m_step(new wxButton(this, ID_RUN_START, _("Step"), wxDefaultPosition, wxSize(150, 50)))
+	, m_selection(new wxButton(this, ID_RUN_START, _("Run to Selection"), wxDefaultPosition, wxSize(150, 50)))
+	, m_vsync(new wxButton(this, ID_RUN_START, _("Go to next VSync"), wxDefaultPosition, wxSize(150, 50)))
+	, m_thread(std::make_unique<GSThread>(this))
+{
+	//TODO: figure out how to fix sliders so the destructor doesn't segfault
+	wxBoxSizer& dump_info(*new wxBoxSizer(wxVERTICAL));
```
Warepire, line 65: This leaks memory every time the dialog is created? Also, I think your sizers are going out of scope when you declare them as references like this, creating a "use after free" scenario?

GovanifY: The dialog is only created once, moreso wxwidgets destructor goes through everything listed in the *this object and deletes all of them, I ran into a bug related to that a few commits above :p

GovanifY: And yeah, no, there is no UAF there; a unique reference to a pointer is created which is then freed on object deletion, nothing wrong with this code

GovanifY: ohhh i left the todo there, might have confused you, let me remove it

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,798 @@
...
+			break;
+		}
+		case VSync:
+		{
+			GSvsync((*((int*)(regs + 4096)) & 0x2000) > 0 ? (u8)1 : (u8)0);
+			break;
+		}
+		case ReadFIFO2:
+		{
+			u64* arr = (u64*)malloc(*((int*)event.data));
```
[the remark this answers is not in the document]
GovanifY: I went to the shared_ptr route instead, since the pointer would get out of scope when assigning it to my struct, even after an std::move since I was calling a function and it was starting to look like a mess.

std::vectors, as you said so well, 0 initialize and have a bigger overhead; knowing I will never resize my arrays I don't need all the semantics it provides.

On to the alloc of elements during the loop: can't get around it, I can't predict the size all elements will take and trying to just alloc the biggest block possible seems like a huge hack for me.

GovanifY:
@tellowkrinkle uhhhhh is it me or is macos compiler not implementing c++17 char[] shared_ptr and that's the reason the check failed?

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,776 @@
...
+		{
+			while (i < (m_root_window->m_dump_packets.size()-1))
+			{
+				m_root_window->ProcessDumpEvent(m_root_window->m_dump_packets[i++], regs);
+
+				if (m_root_window->m_dump_packets[i].id == VSync)
+					break;
+			}
+			if (i >= m_root_window->m_dump_packets.size()-1)
+				i = 0;
```
sonicfind, line 764: Unless this is intended, you're now essentially ignoring the last packet entirely by comparing to `size - 1` instead of just size.

I'm going to guess that the intention was to simplify this part from the original code:
> ```cs
>    if (gs_idx < dump.Data.Count && dump.Data[gs_idx].id == GSType.VSync)
>        break;
> ```

However, it looks to have been structured this way on purpose. As long as packets exist, we always want to process at least one. Only after the first packet do we check for VSync'd, but we can only do that after confirming that the next index is valid to begin with - thus requiring the `gs_idx < dump.Data.Count` check. If none of them have vsync, then every single one should get processed.

Now granted, it should've been
> ```cs
>    if (gs_idx >= dump.Data.Count || dump.Data[gs_idx].id == GSType.VSync)
>        break;
> ```

just so it doesn't end up checking the same false condition twice for an iteration that will fail.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,776 @@
...
+			}
+			case Registers:
+			{
+				std::unique_ptr<char[]> regs_tmp(new char[8192]);
+				m_dump_file->Read(regs_tmp.get(), 8192);
+				GSData data = {id, std::move(regs_tmp), 8192, Dummy};
+				m_root_window->m_dump_packets.emplace_back(data);
+				break;
+			}
+		}
```
sonicfind, line 678: Repaste from discord:

Give GSData a constructor that takes the id and the file pointer. You can then move the switch statement inside the constructor and manipulate the unique_ptr member variable directly. You would no longer need std::move and would simply this section to a single emplace_back call.

pcsx2/gui/Dialogs/ModalPopups.h
```diff
@@ -18,56 +18,352 @@
...
+		{
+			ButtonState btn;
+			int index;
+		};
+		std::vector<GSEvent> m_button_events;
+		std::vector<GSData> m_dump_packets;
+
+		float m_stored_q = 1.0;
+		void ProcessDumpEvent(GSData& event, char* regs);
+		void GenPacketList(std::vector<GSData>& dump);
```
sonicfind, line 345: Assuming that the only times that this is currently used in the code will remain that way, you don't need `std::vector<GSData>& dump` as a parameter. When it gets used on lines 580 & 678, you end up passing the m_dump_packets member variable regardless.

tellowkrinkle:
> uhhhhh is it me or is macos compiler not implementing c++17 char[] shared_ptr and that's the reason the check failed?

Looks like it, wasn't added until Clang 11 which released in October, and Apple's fork usually lags a version or so behind https://gcc.godbolt.org/z/88no6x 
(Linux clang builds probably didn't have the issue because they still use libstdc++, not Clang's libc++)

GovanifY:
> 
> 
> > uhhhhh is it me or is macos compiler not implementing c++17 char[] shared_ptr and that's the reason the check failed?
> 
> Looks like it, wasn't added until Clang 11 which released in October, and Apple's fork usually lags a version or so behind https://gcc.godbolt.org/z/88no6x
> (Linux clang builds probably didn't have the issue because they still use libstdc++, not Clang's libc++)

You just missed an amazing 2 hours of staff trying to figure out C++ pointer semantics and getting this to compile with unique_ptr because share_ptr isn't implemented on macos yet. That was fun haha.

pcsx2/gui/Dialogs/ModalPopups.h
```diff
@@ -341,7 +341,7 @@ namespace Dialogs
 		std::vector<GSData> m_dump_packets;
 
 		float m_stored_q = 1.0;
-		void ProcessDumpEvent(GSData event, char* regs);
+		void ProcessDumpEvent(GSData& event, char* regs);
```
tellowkrinkle, line 344: Don't think you edit this, shouldn't it be a const ref?

GovanifY:
I implemented some basic code for updating the treeview on top of the GS in the UI so that the user gets feedback on their button press. Unfortunately it seems wx is acting up yet again and refuses to behave correctly, not a blocking for merge but if anyone wants to look at it here is the tentative commit that should theoretically implement this feature https://github.com/PCSX2/pcsx2/commit/bb5df08f50bc2e287ed3c7017ca11f2790c10d31

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,830 @@
...
+}
+
+using namespace pxSizerFlags;
+
+// --------------------------------------------------------------------------------------
+//  GSDumpDialog  Implementation
+// --------------------------------------------------------------------------------------
+
+Dialogs::GSDumpDialog::GSDumpDialog(wxWindow* parent)
+	: wxDialogWithHelpers(parent, _("GSDumpGov"), pxDialogFlags())
```
lightningterror, line 58: Why not GS Debugger or something similar?

GovanifY: Because I forgot about it! WIll change it

lightningterror:
Missing minimize button btw.

GovanifY:
> 
> 
> Missing minimize button btw.

Done

tadanokojin:
shall we hide it in the release builds? kind of don't want end users messing about with it.
I think anyone who would know how to use it would know how to build the project.

refractionpcsx2:
> shall we hide it in the release builds? kind of don't want end users messing about with it.
> I think anyone who would know how to use it would know how to build the project.

I agree, it's more of a development build thing, general users don't need it and it isn't much use to them.

lightningterror:
> shall we hide it in the release builds? kind of don't want end users messing about with it.
> I think anyone who would know how to use it would know how to build the project.

I was thinking the same.

GovanifY:
> 
> 
> shall we hide it in the release builds? kind of don't want end users messing about with it.
> I think anyone who would know how to use it would know how to build the project.

Done

GovanifY:
And this should be it, unless new bug arises!

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,860 @@
...
+{
+	if (!m_run->IsEnabled())
+		return;
+	m_thread->m_dump_file = std::make_unique<pxInputStream>(m_selected_dump, new wxFFileInputStream(m_selected_dump));
+
+	if (!(m_thread->m_dump_file)->IsOk())
+	{
+		wxString s;
+		s.Printf(_("Failed to load the dump %s !"), m_selected_dump);
+		wxMessageBox(s, _("GSDumpGov"), wxICON_ERROR);
```
lightningterror, line 185: Forgot this one too.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,860 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2020  PCSX2 Dev Team
```
lightningterror, line 2: 2021

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,860 @@
...
+	}
+	m_run->Disable();
+	m_debug_mode->Enable();
+	m_thread->Start();
+	return;
+}
+
+void Dialogs::GSDumpDialog::ProcessDumpEvent(const GSData& event, char* regs)
+{
+    switch (event.id)
```
turtleli, line 196: Tab instead of spaces.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,860 @@
...
+
+
+	std::unique_ptr<char[]> state_data(new char[ss]);
+	m_dump_file->Read(state_data.get(), ss);
+	m_dump_file->Read(&regs, 8192);
+
+	freezeData fd = {(int)ss, (s8*)state_data.get()};
+	m_root_window->m_dump_packets.clear();
+
+	while (	m_dump_file->Tell() < m_dump_file->Length())
```
turtleli, line 697: Unnecessary tab after `(`.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,860 @@
...
+	m_root_window->m_step->Disable();
+	m_root_window->m_selection->Disable();
+	m_root_window->m_vsync->Disable();
+	m_root_window->m_gif_list->DeleteAllItems();
+	m_root_window->m_gif_packet->DeleteAllItems();
+	m_root_window->m_button_events.clear();
+	m_root_window->m_debug_mode->SetValue(false);
+	m_root_window->m_run->Enable();
+	m_dump_file->Close();
+}
```
turtleli, line 660: This function modifies the GUI state from the task thread. Is that safe?

GovanifY: As far as I know it is, gif list and packets are generated inside the thread and only inside the thread, all other buttons cannot have their state changed while the thread is running, m_run enables you to start a new thread and is enabled last.

pcsx2/gui/Dialogs/GSDumpDialog.cpp
lightningterror: Looks like you reverted this?

pcsx2/gui/Dialogs/GSDumpDialog.cpp
```diff
@@ -0,0 +1,878 @@
...
+		case Registers:
+			m_gif_packet->AppendItem(rootId, "Registers");
+			break;
+	}
+	m_gif_packet->ExpandAll();
+}
+
+void Dialogs::GSDumpDialog::ParsePacket(wxTreeEvent& event)
+{
+	int id = wxAtoi(m_gif_list->GetItemText(event.GetItem()).BeforeFirst('-'));
```
lightningterror, line 429: const will be nice

iMineLink:
Tested it and found no issues.
Just remember to build pcsx2 in Debug/Devel profile to make it accessible ;)

PR #4274, opened 2021-03-08
MagicGate

balika011:


pcsx2/CDVD/CDVD.cpp
```diff
@@ -687,6 +868,36 @@ void cdvdReset()
 	cdvd.RTC.day = (u8)curtime.GetDay(wxDateTime::GMT9);
 	cdvd.RTC.month = (u8)curtime.GetMonth(wxDateTime::GMT9) + 1; // WX returns Jan as "0"
 	cdvd.RTC.year = (u8)(curtime.GetYear(wxDateTime::GMT9) - 2000);
+
+	{
+		FILE *f = fopen("eks.bin", "rb");
```
GovanifY, line 873: Hey, great to see this PR! We are working heavily on the GS right now so it might take a while for us to review/merge that.

Related to that snippet, you probably shouldn't be trying to read files as hardcoded as this, might want to use our ini settings class to set the path/ put it in the bios lookup system?

I'll wait for others to comment on the rest of the file as I'm working on another PR right now, but I assume this adds support for, in theory, official memcards and arcade systems? The protocol seems to be fully implemented, are there any pitfalls or things we should know since you left the description of your PR empty?

pcsx2/des.h
```diff
@@ -0,0 +1,65 @@
+ /**
```
GovanifY, line 1: Might want to throw that in 3rdparty instead of just dropping it in the main folder

pcsx2/Sio.cpp
```diff
@@ -329,24 +331,270 @@ SIO_WRITE MemcardResponse()
...
+	{
+		iv[i] = rand();
+		seed[i] = keysource[i] ^ iv[i];
+		nonce[i] = rand();
+	}
+}
+
+void generateResponse()
+{
+	uint8_t ChallengeIV[8] = { /* SHA256: e7b02f4f8d99a58b96dbca4db81c5d666ea7c46fbf6e1d5c045eaba0ee25416a */ };
```
GovanifY, line 395: you will _probably_ need to implement something to look up the IV there, probably from our ini

pcsx2/CDVD/CDVD.cpp
```diff
@@ -40,6 +41,67 @@ cdvdStruct cdvd;
...
+	0x0090, 0x00A8, 0x00A8, 0x0094, 0x00AC, 0x00AC, 0x0098, 0x00B0, 0x00B0, 0x009C, 0x00B4, 0x00B4, 0x00A0, 0x00B8, 0x00B8, 0x00A4, 0x00BC, 0x00BC
+};
+
+uint16_t g_KelfKeysIndex[4] = { 0x110, 0x110, 0xC4, 0x15C };
+
+uint16_t g_cardKeyStore[48] = {
+	/* SHA256: 04bcc6b13827829fb5cc8dbd86420d30f69a2bfd3b7719398b341e15368bd365 */
+};
+
+uint8_t g_KeyStoreKey[16] = { /* SHA256: 04bcc6b13827829fb5cc8dbd86420d30f69a2bfd3b7719398b341e15368bd365 */ };
```
GovanifY, line 102: ditto

GovanifY (changes requested):
From a cursory look it seems fine, no keys included so legally I'd say it's fine myself, but I'll ask for others to comment.

A cursory look at the protocol seems to implement it correctly, will need more time to review; thanks for the work but it'll probably still take quite a bit of work to see it merged

balika011:
Feel free to do anything you want with this.
Clean up, make some polish, add UI and stuff like that.
Don't forget to bring your own keys!

pcsx2/CDVD/CDVD.cpp
```diff
@@ -687,6 +868,36 @@ void cdvdReset()
...
+	{
+		FILE *f = fopen("kek.bin", "rb");
+		if (f)
+		{
+			fread(g_KeyStoreKey, 1, sizeof(g_KeyStoreKey), f);
+			fclose(f);
+		}
+	}
+
+	readAndDecryptKeyStore(1); // 0: dev, 1: retail, 2: proto?, 3: arcade
```
balika011, line 899: this needs to be set based on the loaded bios

pcsx2/CDVD/CDVD.cpp
balika011: Arcade (Namco System 246/256) has some extra security that fails on PCSX2. It initializes the memory card, opens boot.bin, but doesn't try to decrypt. Fails before that. If somebody finds out why this happens it likely means that arcade will work.

kenshen112:
Could it be the arcades keys are different?

refractionpcsx2:
Hi there,

Thanks for the work you've done in reversing MechaCon.  I don't know if we are going to pursue MagicGate emulation, legally it feels a little dangerous to be going down this path and without getting proper legal advice I'm not really comfortable putting the code in, regardless of the keys being there or not.

I will leave this code here as it will probably be interesting for somebody and we may revisit in the future, but for now, this is probably as far as it's going to go.

PR #4276, opened 2021-03-09
gsdump: minor cleanup in packet emplacement & parsing logic

sonicfind:
While it compiled on windows, I am unsure whether it will see success on different systems. Checking that would be nice.


GovanifY:
Backported the stylistic changes in https://github.com/PCSX2/pcsx2/pull/4271/commits/dd15f00d2b0f22d1e07d3224a1cad0492ecda411 , no temporary variables are created by this code so the struct initializer was unneeded, see https://godbolt.org/z/TM77Kd

PR #4277, opened 2021-03-09
3rd Party: Update Libsamplerate to 0.2.1

refractionpcsx2:


3rdparty/libsamplerate/config.h
```diff
@@ -195,8 +195,9 @@
 /* Version number of package */
 #define VERSION "0.1.9"
```
RedDevilus, line 196: Doesn't this need 0.2.1?

refractionpcsx2: Fixed now.

3rdparty/libsamplerate/src_sinc.c
```diff
@@ -1,18 +1,22 @@
...
+#include <assert.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <math.h>
 
-#include "config.h"
-#include "float_cast.h"
 #include "common.h"
+#include "config.h"
```
xTVaser, line 19: Rather than adding this include manually, I'd suggest leaving it how it originally was:
https://github.com/libsndfile/libsamplerate/blob/a147b6dc4046deda22b2aa772b348717de772e2b/src/src_sinc.c#L9-L11
And adding the `HAVE_CONFIG_H` flag to the preprocessor defs in the vs project file.  As all other files have a similar check

refractionpcsx2: I can put them back in, but I wonder why we need them? They certainly weren't in there for the previous version and it won't work if the file isn't there so it seems silly to make it optional

xTVaser: Right, `HAVE_CONFIG_H` seems to be a new thing in the latest version, and I'll be honest -- I have no idea what the use-case is without the config.  But modifying the original code as little as possible will make future dependency bumps easier (won't have to remember the edits we did), up to you.

refractionpcsx2: Fair enough :) Well I'll leave it for now, hopefully we won't need to upgrade again :P

PR #4278, opened 2021-03-09
GUI-Speedhacks: Grayout Instant VU1 when MTVU is enabled

RedDevilus:
MTVU checked + Instant VU1 checked = MTVU is on
MTVU checked + Instant VU1 off = MTVU is on
Essentially it doesn’t matter for Instant VU1 be checked or unchecked if MTVU is checked.
MTVU off + Instant VU1 checked = Instant VU1 is on
MTVU off + Instant VU1 off = both are off

Instant VU1 enabled is the old behaviour (so think of it always being enabled in the past). Some games however don’t like this and for those few games they should be off, those kind of games should’ve normally been added as an automatic gamefix or will be in the future.

### DIFF
--- a/pcsx2/gui/Panels/SpeedhacksPanel.cpp
+++ b/pcsx2/gui/Panels/SpeedhacksPanel.cpp
@@ -242,6 +242,7 @@ Panels::SpeedHacksPanel::SpeedHacksPanel( wxWindow* parent )
 	Bind(wxEVT_SCROLL_CHANGED, &SpeedHacksPanel::VUCycleRate_Scroll, this, m_slider_eeSkip->GetId());
 	Bind(wxEVT_CHECKBOX, &SpeedHacksPanel::OnEnable_Toggled, this, m_check_Enable->GetId());
 	Bind(wxEVT_BUTTON, &SpeedHacksPanel::Defaults_Click, this, wxID_DEFAULT);
+	Bind(wxEVT_CHECKBOX, &SpeedHacksPanel::VUThread_Enable, this, m_check_vuThread->GetId());
 }
 
 // Doesn't modify values - only locks(gray out)/unlocks as necessary.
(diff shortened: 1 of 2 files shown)


RedDevilus:
Examples after you apply the settings;

MTVU is on, so it disables Instant VU1 checkbox:
image

MTVU is off, so it doesn't touch Instant VU1 checkbox:
image

seta-san:
just curious. does the emulator do the same thing for games that enable one of these settings from the gamesdb?

pcsx2/gui/Panels/SpeedhacksPanel.cpp
```diff
@@ -368,3 +372,11 @@ void Panels::SpeedHacksPanel::VUCycleRate_Scroll(wxScrollEvent &event)
...
 	event.Skip();
 }
+
+void Panels::SpeedHacksPanel::OnClick_VUThread(wxCommandEvent& evt)
+{
+	if (m_check_vuThread->IsEnabled())
+		m_check_vu1Instant->Disable();
+	else
+		m_check_vu1Instant->Enable();
+}
```
lightningterror, line 382: Maybe add 1 more empty line after the bracket.

RedPanda4552:
> 
> 
> just curious. does the emulator do the same thing for games that enable one of these settings from the gamesdb?

Speedhacks are not allowed to be set via the games database. However, for gamefixes and the CPU settings, their changes are not reflected in the UI (this prevents a game database fix from permanently activating a gamefix just because you happened to open and apply one of the UI panels).

refractionpcsx2:
Ticking MTVU successfully disables Instant VU in the GUI, however unticking MTVU doesn't reverse that process.

RedDevilus:
In a process of fixing that issue and breaking other stuff again, i finally seem to have found a working formula.
I've also squashed it 🧑‍🏭

refractionpcsx2:
Sorry :P If you skip between presets you can get in to situations where the text is enabled and the box isn't

image

RedDevilus:
How to produce the issue of refraction:
- Try to click to the left and/or right without touching the slider button itself, but still on the slider itself
- Look at the tooltip text
It's pretty hard to trigger and i had made sanity checks to avoid this kind of issue.

https://user-images.githubusercontent.com/24227051/111801937-bc138080-88cd-11eb-9584-7536dbfe084f.mp4

RedDevilus:
All issues should be resolved, with hopefully no more glue code.

pcsx2/gui/Panels/SpeedhacksPanel.cpp
```diff
@@ -368,3 +372,9 @@ void Panels::SpeedHacksPanel::VUCycleRate_Scroll(wxScrollEvent &event)
 
 	event.Skip();
 }
+
+void Panels::SpeedHacksPanel::VUThread_Enable(wxCommandEvent& evt)
+{
+m_check_vu1Instant->Enable(!m_check_vuThread->GetValue());
```
GovanifY, line 378: glue code, indentation

refractionpcsx2:
It works, but it doesn't disable Instant VU1 when you disable speedhacks.

JordanTheToaster:
Click the things and it dose the thing it's intended to do and it all seems to work perfectly.

JordanTheToaster:
I have found a bug if you untick speedhacks and then re tick it MTVU dose not grey out Instant VU1 until you untick and re tick MTVU.

RedDevilus:
Probably because i deleted the Layout(). Guess i'll put it back.

RedDevilus:
This is probably as good as it gets (without hacking the original function) , there is an edgecase that the repainting event can fail which is mentioned in the code comments where you go from preset 2 to preset 3 disable preset, disable speedhacks, enable speedhacks. It doesn't trigger if you have put Apply once somewhere in the process and it doesnt trigger again even if you do mess with enable speedhacks.

RedDevilus:
Everything seems to work fine now, i was just being an idiot. 🤦

RedDevilus:
https://user-images.githubusercontent.com/24227051/113364210-370b7b00-9353-11eb-9cd2-aa115474555f.mp4

PR #4281, opened 2021-03-11
General cleanup of various linux pad code.

arcum42:
Various restructuring and header and file cleanup on the linux pad code (formerly onepad).

I decided to redo much of the cleanup I'd previously done without the experiments and a bunch of half done things I had in there, and to handle gui, multitap, and various other things as separate prs later.


pcsx2/CMakeLists.txt
```diff
@@ -495,8 +495,8 @@ add_custom_command(
 
 # PAD/Linux sources 
 set(pcsx2PADSources
-    PAD/Linux/controller.cpp
-    PAD/Linux/GamePad.cpp
+    PAD/Linux/Device.cpp
+	PAD/Linux/InputManager.cpp
```
RedDevilus, line 499: Formatting, other locations also.

pcsx2/CMakeLists.txt
```diff
@@ -506,16 +506,18 @@ set(pcsx2PADSources
...
 
 # PAD/Linux headers
 set(pcsx2PADHeaders
     PAD/Linux/bitwise.h
-    PAD/Linux/controller.h
-    PAD/Linux/GamePad.h
+    PAD/Linux/Config.h
+    PAD/Linux/Device.h
+	PAD/Linux/Global.h
+	PAD/Linux/InputManager.h
```
RedDevilus, line 520: Ditto

pcsx2/CMakeLists.txt
arcum42: Should be taken care of, though the auto-formatting I ran may have tweaked the formatting in a few spots.Still might need to fix up nopch compiling. We'll see.

pcsx2/PAD/Linux/Global.h
```diff
@@ -0,0 +1,107 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2020  PCSX2 Dev Team
```
RedDevilus, line 2: Copyright 2021

arcum42: Taken care of.

lightningterror:
Would be nice to rebase and run clang again with the new changes, also some commits can be squashed and better named.

tellowkrinkle (commented):
I'd prefer if you used `git clang-format` instead of `clang-format` to avoid applying clang-format's preferences to things you haven't changed, as these files have been formatted recently and the majority of the remaining things are just things clang-format is dumb about

Though if you rerun it with the latest config it should be at least a bit more reasonable, we finally stopped it from blasting function call continuations off into space with piles of tabs

GovanifY (approved):
fine to merge by me!

pcsx2/PAD/Linux/InputManager.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+std::unique_ptr<InputDeviceManager> device_manager(new InputDeviceManager);
+
+InputDeviceManager::InputDeviceManager()
+{
+}
+
+InputDeviceManager::~InputDeviceManager()
+{
+	device_manager->devices.clear();
```
tellowkrinkle, line 33: Any reason this references `device_manager` instead of `this`?  It's crashing on shutdown with clang's libc++ because unique_ptr's destructor copies the pointer into a temporary and wipes the original before calling the destructor.

PR #4282, opened 2021-03-12
pad-windows: Fix a couple of warnings, make codacy happy.

lightningterror:


pcsx2/PAD/Windows/DualShock3.cpp
```diff
@@ -184,10 +182,8 @@ struct LightState
 struct DS3Command
 {
 	unsigned char id;
-	unsigned char unsure;
```
CookiePLMonster, line 187: Is it not breaking ABI with some HID report or external program? This and the below change can technically change the size and layout of this struct.

lightningterror: I'll drop the structure changes then, better play it safe.

CookiePLMonster: FWIW I **think** `[[maybe_unused]]` is valid to be used even on data fields like that. Maybe that's the best way to move forward?

lightningterror: Ah right, forgot about that attribute, let's see if it solves it.

lightningterror: Seems that it still complains.

Issue #4286, opened 2021-03-12
[BUG] Performance regression (Need for Speed: Most Wanted/Carbon)


bamq:
**Describe the bug**
It seems there is a severe performance hit after https://github.com/PCSX2/pcsx2/commit/862518e7da5dd50f1f3990324ad8960f2371a020 testing the games Need for Speed: Most Wanted and Need for Speed: Carbon. Before, the games ran perfectly at 60fps, now they run about 45fps or lower. Revert back to 1.7.0-dev-1050 (build before that commit) and the issue is not there.

**To Reproduce**
Load up the game to at least the main menu.

**Expected behavior**
Game runs smoothly as it used to.

**Emulation Settings**
Turning off speed hacks seems to make some sort of difference, but they used to work fine with no problems.

**GS Window Screenshots**
![image](https://user-images.githubusercontent.com/6969348/110995779-848e5c80-8340-11eb-9fd3-4c2fcf5dffd1.png)

**System Info (please complete the following information):**
 - PCSX2 Revision: 1.7.0-dev-1072
 - OS: Windows 10 64bit
 - CPU: AMD Ryzen 7 3700X
 - GPU: NVIDIA GeForce RTX 2070 Super

lightningterror:
A gsdump would be nice for testing.

bamq:
Does this work? Will gladly provide more info if needed

[gsdx_20210312145754.gs.zip](https://github.com/PCSX2/pcsx2/files/6132905/gsdx_20210312145754.gs.zip)

bamq:
I have noticed another strange thing. After running the game to the main menu, trying to shutdown the game causes the console log to get stuck on this line:
`Decommitting host memory for virtual systems...`

and PCSX2 will not close or boot any game, has to be killed through task manager.

tellowkrinkle:
@bamq Can you verify with the above PR?

bamq:
I've downloaded the artifact from your pull request. The issue is still there. The performance actually has improved somewhat, though it still struggles with maintaining full speed emulation and there are dramatic slowdowns in some areas of the game where there were none before. The issue I mentioned above with shutting down the emulator while in the game's menu also persists.

bamq:
I've recorded a video comparing the old build without the issues and your pr build if this is at all useful: https://youtu.be/8vp32Azd7Hw

refractionpcsx2:
[...]

PR #4288, opened 2021-03-13
GSdx: Don't use constexpr vector constructors at runtime

tellowkrinkle:
The MSVC implementation is slow


tellowkrinkle:
Should fix #4286

refractionpcsx2:
Not even close, sorry bud :(

PR #4290, opened 2021-03-13
PCSX2-HW: Remove unused GS/SMODE2 struct, move defines

MrCK1:
According to [git blame](https://github.com/PCSX2/pcsx2/commit/a9165dfbf6266362b1754ccc346faec3a711db35) this was used for the GS Window at some point. Doesn't seem to be the case anymore. Tested compile of PCSX2/GSdx and output GS Window and the titlebar, doesn't seem to be missing anything with this change in place.

I'm not sure if the 2nd commit worth merging, but the GS header might as well be a better home. (the rest of HW.h seems to be EE hardware and DMAC registers whereas these are only used by the GS)


MrCK1:
PR is no longer needed, changes are already in master

PR #4295, opened 2021-03-14
Fix #4294 convert wxString to std::string for fmt::format() 

guoyunhe:

PR #4296, opened 2021-03-14
Dependencies: Bump {fmt} to `7.1.3`

xTVaser:
Issues have been reported surrounding the `{fmt}` library which seem to be environment specific https://github.com/PCSX2/pcsx2/pull/4295 .  These build errors occurred on the latest version `7.1.3`, and while there are people that have experienced no issues on this version and CI passes, it would be wise for us to keep the submodule up to date to ease troubleshooting.

### DIFF
--- a/.github/workflows/linux-workflow.yml
+++ b/.github/workflows/linux-workflow.yml
@@ -72,18 +72,6 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v2
 
-      - name: Cache Submodules
-        id: cache-submodules
-        uses: actions/cache@v2.1.4
-        with:
-          key: submodules-${{ hashFiles('./.gitmodules') }}
-          path: |
-            ./.git/modules/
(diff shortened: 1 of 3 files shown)


qurious-pixel:
You could do something like add a versioning to the key.
`key: submodules-cache-v1-${{ hashFiles('./.gitmodules') }}` then update cache version as submodules are updated. (Changing key name will create a new cache)

kenshen112:
That's a fair way to do it. I'd really prefer that, and if we could force cmake and the like to build with only the submodule instead of the system packages

xTVaser:
I'm not totally convinced, that requires people to remember and be mindful of that when bumping submodules to save 10seconds tops per build.  I had a similar thought which was to include the version in the submodule file as a comment, which would change the result of `hashFiles`, but still relies on everyone remembering to do so.  In my experience, these things tend to drift and the worst-case scenario is builds are successful when they shouldn't be.

Issue #4247, opened 2021-02-23
[Feature Request] Active savestate slot GUI


RedDevilus:
**Description**
Displaying/changing the active savestate slot in the GUI on saving and loading tab.

**Reason**
Easier overview for endusers especially those who use alot of savestate and cycle between them. QoL feature would be able to change the active saveslot in the GUI itself instead of booting up a game.
Maybe even expanding the system to make it not default to the first saveslot, even separately between loading and saving (maybe a bit too much, it's fine for 1 active slot for both).

**Examples**
Default behaviour:
![image](https://user-images.githubusercontent.com/24227051/108921459-aac0b680-7636-11eb-8738-e1da5a52c3bd.png)

Saving to Slot 5:
![image](https://user-images.githubusercontent.com/24227051/108922226-068b3f80-7637-11eb-8c25-c16298c50d99.png)

Changing the active savestate slot in the GUI:
![image](https://user-images.githubusercontent.com/24227051/108922813-f4f66780-7637-11eb-95e0-9292a381d74a.png)

RedDevilus:
This idea could even be implemented in (or future PR) https://github.com/PCSX2/pcsx2/pull/3615

PR #4306, opened 2021-03-18
gui: Visually indicate the current savestate slot in the load/save menu

xTVaser:
Resolves https://github.com/PCSX2/pcsx2/issues/4247

Comments on some of the ideas in that issue:
- Changing the active slot via the GUI
  - To me, this can already be accomplished with the current GUI.  If you save or load a savestate, it will switch and make that the active slot.  Having a way to select the active slot first in the GUI, doesn't really solve anything, it just adds an extra step that i think most people would avoid?  (ie. why switch to a slot, then save it, when you could just save it in one step)
- However, I do think it would be cool to retain the select slot between sessions in the config, instead of always defaulting to `0`
  - And agreed, I think having a separate active slot for loading/saving would get too confusing and im not sure what the use-case would be.

image
image

fyi @RedDevilus

### DIFF
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -43,7 +43,8 @@ wxMenu* MainEmuFrame::MakeStatesSubMenu(int baseid, int loadBackupId) const
 	for (int i = 0; i < 10; i++)
 	{
 		// Will be changed once an iso is loaded.
-		mnuSubstates->Append(baseid + i + 1, wxsFormat(_("Slot %d"), i));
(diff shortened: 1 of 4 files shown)


RedDevilus:
As someone who cycles and saves alot of savestates for testing, it's nice to see visually what was the latest saved slot.

Although it resets back to slot 0 if you exit the emulator, just like master.

RedDevilus:
image

Switching games seems to be the same as master itself (remembering which saveslot you last accessed).
Though it's an improvement atleast :)

RedDevilus (approved):

xTVaser:
Yeah I like persisting the last slot idea, but i didn't want to mess with configuration in this PR.  Storing a slot per game is also going to be really annoying with the current config setup, so that's a far-future goal.  You basically need per-game config for that.

PR #4310, opened 2021-03-20
USB: Improve Force Feedback test so it doesn't block UI

### LINKED ISSUE #3926: USB: Issues with Wheel device support
Upon attempting to use USB steering wheel emulation with **PCSX2 v1.7.0-dev-593-g35122e660** and **DInput device API** I quickly encountered an array of issues. This issue serves as a tracker for all those.

I currently have not gathered any logs. I assume that for more information I need to provide logs generated by enabling the logging option from **USB Settings**.

## FFB issues:
- [x] **Force Feedback randomly doesn't initialize** - this seems to be inconsistent, but launching PCSX2 can result in Force Feedback not working (not in game nor in the Test screen). No particular pattern (e.g. every second launch) seems to exist. This might very well be an issue of wheel's drivers, but in this case I assume DirectInput should report the failure somewhere down the line. **Fixed with #4042**
- [x] **Force Feedback strength seems to malfunction** - in cases where the above point doesn't apply (i.e. FFB initialized correctly), Force Feedback appears to be so strong it could be a hazard. Turning down the **Gain** option does not actually seem to decrease the force of effects. **Fixed with #4042**
- [x] **Pausing emulation does not disable Force Feedback effects** - effects playing at the time of pausing emulations don't stop, which, combined with the previous point, can result in a rather unpleasant (if not dangerous) experience. **Fixed with #4042**

## UI issues:
- [ ] **UI remains unresponsive during the Force Feedback test** - a minor UI issue, but in cases where Test doesn't play any FFB effects (e.g. like when it doesn't initialize) it can be rather confusing.

## Wheel emulation issues (maybe? Need confirmation):
- ~**GT Force wheel malfunctions with NASCAR Heat 2002** - all proof of the GT Force working with this game Is an article [HERE](https://www.gamezone.com/news/nascar_heat_2002_to_support_logitech_force_feedback_wheel/), but attempting to do so in PCSX2 results in a rather broken experience - the game lacks any in-game remapping for the steering wheel (so I assume it should have been properly bound out of the box), but pedals don't work as expected. Instead, Circle/Cross function as throttle/brake.~ **Invalid - it's a game bug. Restoring controller mappings to Default "fixes" it.**

All those issues were encountered with a Thrustmaster T300RS GT and a DInput API enabled in the USB settings.

CookiePLMonster:
This PR modifies DirectInput USB Force Feedback test not to block the UI. Instead of performing the test synchronously on the UI thread, a timer gets created which advances the test every half a second.

Closes #3926. With this change in place, all the points from this meta-issue are resolved.

EDIT: `AutoCenter` was removed because it doesn't work when the application submits custom FFB data anyway.

### DIFF
--- a/pcsx2/USB/usb-pad/dx/dinput-config.cpp
+++ b/pcsx2/USB/usb-pad/dx/dinput-config.cpp
@@ -53,6 +53,10 @@ namespace usb_pad
 		int32_t FFMULTI[2][1];
 		int32_t INVERTFORCES[2]{};
 
+		// FFB test
+		bool ffbTestRunning = false;
+		unsigned int ffbTestStage = 0;
+
 		bool dialogOpen = false;
 
 		HWND hKey;
@@ -685,6 +689,14 @@ namespace usb_pad
 			EndPaint(hWnd, &Ps);
 		}
 
+		void EndFFBTest()
+		{
+			if (std::exchange(ffbTestRunning, false))
+			{
+				KillTimer(hWnd, 23);
+			}
+		}
+
 		INT_PTR CALLBACK StaticProc(HWND hDlg, UINT uMsg, WPARAM wParam, LPARAM lParam)
 		{
 			(*pFnPrevFunc)(hDlg, uMsg, wParam, lParam);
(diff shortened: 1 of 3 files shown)


pcsx2/USB/usb-pad/dx/dinput-config.cpp
```diff
@@ -943,16 +966,28 @@ namespace usb_pad
...
 						case IDC_BUTTON1:
 						{
-							//MessageBeep(MB_ICONEXCLAMATION);
-							ApplySettings(s->port);
-							TestForce(s->port);
+							if (!ffbTestRunning)
+							{
+								ApplySettings(s->port);
+								if (StartTestForce(s->port))
+								{
```
jackun, line 981: Trailing space

jackun (approved):

PR #4315, opened 2021-03-21
gs: remove some pesky asserts

tadanokojin:
Really tired of my debugger crashing due to known behavior.

### DIFF
--- a/plugins/GSdx/GSState.cpp
+++ b/plugins/GSdx/GSState.cpp
@@ -1265,11 +1265,7 @@ void GSState::GIFRegHandlerSCISSOR(const GIFReg* RESTRICT r)
 template <int i>
 void GSState::GIFRegHandlerALPHA(const GIFReg* RESTRICT r)
 {
-	ASSERT(r->ALPHA.A != 3);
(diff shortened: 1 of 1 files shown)


plugins/GSdx/GSState.cpp
```diff
@@ -1265,11 +1265,6 @@ void GSState::GIFRegHandlerSCISSOR(const GIFReg* RESTRICT r)
 template <int i>
 void GSState::GIFRegHandlerALPHA(const GIFReg* RESTRICT r)
 {
-	ASSERT(r->ALPHA.A != 3);
```
lightningterror, line 1268: Maybe replace them with a log instead?

tadanokojin: I don't really think the info is all that useful but I can stick a GL_REG in there if that makes you feel better.

lightningterror: Yeah it would be nice to have it.

tadanokojin: done.

PR #4321, opened 2021-03-25
Move ISO decompression to a separate thread

### LINKED ISSUE #4319: Asynchronous decompression of compressed formats
Now that PCSX2 is getting more and more accurate with disc timings and compressed formats like CSO and [CHD](https://github.com/PCSX2/pcsx2/pull/4314) getting increasingly popular, it would be unwise not to (further) improve the decompression performance to avoid any stutters/glitches.

We currently support using the OS's async IO for reading isos, but don't have anything in place for decompressing on a separate thread

tellowkrinkle:
Closes #4319

When loading from a mostly-LZMA CHD (SSD, i7-4980HQ w/ turbo boost off)
Time to play P3FES intro cutscene (disk-heavy, linear, timed with phone stopwatch)
- Uncompressed ISO: 57s
- Async + Readahead: 57s
- Asyncronous: 61s
- Synchronous: 65s

I was expecting hunk size to make more of a difference, but results were about the same for both my default settings (~20kb hunks) chd and max size one (~1mb hunks, chdman refuses to make bigger ones), however this FMV reads 100% linearly, so that might be why.

Also for anyone on slow HDDs, this also means that using compressed formats won't take away your async file reading (previously none of the compressed disk readers used async file APIs.  They still don't but now they don't block the EE thread calling them)

### DIFF
--- a/pcsx2/CDVD/ChdFileReader.cpp
+++ b/pcsx2/CDVD/ChdFileReader.cpp
@@ -20,7 +20,6 @@
 
 #include <wx/dir.h>
 
-
 bool ChdFileReader::CanHandle(const wxString& fileName)
 {
 	if (!wxFileName::FileExists(fileName) || !fileName.Lower().EndsWith(L".chd"))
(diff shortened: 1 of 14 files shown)


pcsx2/CDVD/ChdFileReader.cpp
```diff
@@ -30,177 +29,143 @@ bool ChdFileReader::CanHandle(const wxString& fileName)
...
-				delete header;
-				delete parent_header;
-				return false;
-			}
-			bool found_parent = false;
-			wxFileName wxfilename(chds[chd_depth]);
-			wxString dir_path = wxfilename.GetPath();
-			wxDir dir(dir_path);
-			if (dir.IsOpened())
+			Console.Error(L"chd_open hit recursion limit searching for parents");
```
lightningterror, line 53: Adding the `CDVD:`  prefix to the logs would be nice, same for the others.

pcsx2/CDVD/ThreadedFileReader.cpp
```diff
@@ -0,0 +1,360 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2021  PCSX2 Dev Team
```
lightningterror, line 2: Maybe `Copyright (C) 2002-2021  PCSX2 Dev Team` for consistency with the others.

GovanifY (approved):

Ziemas (approved):

RedDevilus (approved):

lightningterror:
Needs rebase.

PR #4324, opened 2021-03-25
msbuild: fix libchdr include directory

tadanokojin:
I don't know what PublicIncludeDirectories is, it's not documented anywhere by MS.
Even if it doesn't fix the bot, it's at least consistent now.


refractionpcsx2:
Downloaded it and built manually, was successful

tadanokojin:
good luck buildbot kun

Issue #3906, opened 2020-11-06
SPU2-X: Crash after shutdown after switching from some sound modules


MrCK1:
***PCSX2 version:***
1.7.0-dev-515

***PCSX2 options:***
Defaults

***Plugins used:***
Defaults, rest are integrated

***Description of the issue:***
After switching to certain output modules and shutting down the VM, PCSX2 will crash.

***How to reproduce the issue:***

Testing Result (with core running):
No Sound -> PortAudio (doesn't apply properly w/o restarting VM) - No Crash
No Sound -> WaveOut (doesn't apply properly w/o restarting VM) - No Crash
No Sound -> XAudio (doesn't apply properly w/o restarting VM) - No Crash

**PortAudio -> WaveOut - Crash on shutdown**
**PortAudio -> XAudio - Crash on shutdown**
PortAudio -> NoSound - No Crash

XAudio -> PortAudio - No Crash
XAudio -> WaveOut - No Crash
XAudio -> NoSound - No Crash

**WaveOut -> PortAudio - Crash on shutdown**
WaveOut -> XAudio - No Crash
WaveOut -> NoSound - No Crash

***Last known version to work:***
Before SPU2 merge into core

***PC specifications:***
Windows 10 64-bit Build 2004, GTX 1070, i5-6600K, Sound Blaster X-Fi Titanium Sound Card

MrCK1:
Fixed in #4333

PR #4333, opened 2021-03-28
SPU2: Simplify reset function for sample rate

MrCK1:
This PR combines the previously separate sample rate functions for PS2 and PSX modes into a single one.

### DIFF
--- a/pcsx2/Hw.cpp
+++ b/pcsx2/Hw.cpp
@@ -65,12 +65,9 @@ void hwReset()
 	psHu32(DMAC_ENABLEW) = 0x1201;
 	psHu32(DMAC_ENABLER) = 0x1201;
 
-	if ((psxHu32(HW_ICFG) & (1 << 3)))
-	{
-		SPU2ps1reset();
-	}
-
-	SPU2reset();
+	// Sets SPU2 sample rate to PS2 standard (48KHz) whenever emulator is reset.
+	// For PSX mode sample rate setting, see HwWrite.cpp
+	SPU2reset(PS2Modes::PS2);
 
 	sifReset();
 
--- a/pcsx2/SPU2/spu2.cpp
+++ b/pcsx2/SPU2/spu2.cpp
@@ -112,13 +112,27 @@ void SPU2writeDMA7Mem(u16* pMem, u32 size)
 	Cores[1].DoDMAwrite(pMem, size);
 }
 
-s32 SPU2reset()
+s32 SPU2reset(PS2Modes isRunningPSXMode)
 {
-	if (SndBuffer::Test() == 0 && SampleRate != 48000)
-	{
-		SampleRate = 48000;
-		SndBuffer::Cleanup();
+	u32 requiredSampleRate = (isRunningPSXMode == PS2Modes::PSX) ? 44100 : 48000;
 
+	if (isRunningPSXMode == PS2Modes::PS2)
+	{
+		memset(spu2regs, 0, 0x010000);
+		memset(_spu2mem, 0, 0x200000);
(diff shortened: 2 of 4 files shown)


pcsx2/SPU2/spu2.h
```diff
@@ -22,8 +22,7 @@
 extern Threading::MutexRecursive mtx_SPU2Status;
 
 s32 SPU2init();
-s32 SPU2reset();
-s32 SPU2ps1reset();
+s32 SPU2reset(bool isRunningPSXMode);
```
Warepire, line 25: Might help readability if an enum class is used over a boolean. e.g.
```
enum class PS2Modes {
  PS2,
  PSX,
};
```

pcsx2/Hw.cpp
```diff
@@ -65,12 +65,7 @@ void hwReset()
 	psHu32(DMAC_ENABLEW) = 0x1201;
 	psHu32(DMAC_ENABLER) = 0x1201;
 
-	if ((psxHu32(HW_ICFG) & (1 << 3)))
-	{
-		SPU2ps1reset();
-	}
-
-	SPU2reset();
+	SPU2reset(false);
```
Warepire, line 68: Looks like this one is still using the boolean and somehow gets away with it?

pcsx2/SPU2/spu2.cpp
```diff
@@ -162,66 +162,38 @@ void SPU2writeDMA7Mem(u16* pMem, u32 size)
...
-	else
-		SampleRate = 48000;
-
-	memset(spu2regs, 0, 0x010000);
-	memset(_spu2mem, 0, 0x200000);
-	memset(_spu2mem + 0x2800, 7, 0x10); // from BIOS reversal. Locks the voices so they don't run free.
+	u32 requiredSampleRate = (isRunningPSXMode == PSX) ? 44100 : 48000;
 
-	Spdif.Info = 0; // Reset IRQ Status if it got set in a previously run game
+	if (isRunningPSXMode)
```
Warepire, line 169: Please be explicit here about which value you're testing for.

pcsx2/Hw.cpp
GovanifY: you're right, i forgot my git add, was about to do a second pass but as i'm multitasking, i was waiting for builds to run first; this is because the enum in this case is casted as a bool

Warepire:
Looks fine to me now.

pcsx2/SPU2/spu2.cpp
```diff
@@ -162,66 +162,38 @@ void SPU2writeDMA7Mem(u16* pMem, u32 size)
...
-	else
-		SampleRate = 48000;
-
-	memset(spu2regs, 0, 0x010000);
-	memset(_spu2mem, 0, 0x200000);
-	memset(_spu2mem + 0x2800, 7, 0x10); // from BIOS reversal. Locks the voices so they don't run free.
+	u32 requiredSampleRate = (isRunningPSXMode == PSX) ? 44100 : 48000;
 
-	Spdif.Info = 0; // Reset IRQ Status if it got set in a previously run game
+	if (isRunningPSXMode == PSX)
```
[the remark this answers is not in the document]
refractionpcsx2: Beard noticed it, I'm more the messenger :P

tellowkrinkle:
Could you separate actual changes and formatting changes?  It's really hard to find the actual changes here

kenshen112:
What else needs to be done to merge this?

MrCK1:
It needs some additional changes, I'll have to take a look at some point

MrCK1:
Added commit that restores ability to switch output modules on the fly (without being tied to GS window status). For some reason this functionality got lost after the SPU2 merge. Should address issues in #3906 but needs testing on Linux as well.

Issue #2648, opened 2018-10-17
Rendering on Linux ~50% slower than on Windows


Pixelnarium:
***PCSX2 version:***
git build from today

***PCSX2 options:***
Standard Settings with OpenGL Software Mode (2 Threads are set as standard)

***Plugins used:***
GSdx AVX2

***Description of the issue:***
I am not sure what is causing this. I tried it with the game Flatout 2. As you can see on both screenshots I get around 30% on Linux and about 70% on Windows. "top" shows a CPU usage of around 180% for the pcsx2 process no matter how many threads I set. Maybe multi-threading does not work on Linux?

***PC specifications:***
Intel i5-4460 3.2Ghz, NVidia GTX 1060
Linux: ArchLinux, XFCE4, Compositor disabled
Windows: Windows 10 1803

Linux:
![pcsx2_linux](https://user-images.githubusercontent.com/30435935/47118859-10cbf080-d269-11e8-9547-e6e36d5f4707.png)

Windows:
![pcsx2_win](https://user-images.githubusercontent.com/30435935/47118862-13c6e100-d269-11e8-9de2-a8a148c954ab.png)

Pixelnarium:
Same settings, yes "2 - safe" selection. MTVU is disabled but does not change anything. It seems like that on Windows the CPUs are better used. On Windows every core seems to be used >60% while on Linux its more like <40%. But it could just be the display.

GSdx threads are set to 4 for these pictures. 

Linux:
![pcsx2_linux](https://user-images.githubusercontent.com/30435935/47154066-912e3800-d2e1-11e8-9a25-d73e2b9ae8d0.png)

Windows:
![pcsx2_win](https://user-images.githubusercontent.com/30435935/47154072-94c1bf00-d2e1-11e8-9fc4-7270e0e1826f.png)

mirh:
.. ok something sketchy is really going on. 
Could you try to experiment with threads number on both OS?
Also, for the records, what are numbers for hardware renderer?

uyjulian:
Try wine with the windows version on Linux, and see how it performs

Pixelnarium:
In WINE it runs worse. Around ~15% Performance so half of the native performance. Changing the thread numbers seems to do pretty much nothing (maybe because I only have 4 cores without HyperThreading)?. The hardware renderer is at 100% on both platforms.

mirh:
Mhh, could you try to run some benchmark between windows and linux?
[...]

PR #4346, opened 2021-04-01
Spin before sleeping threads to reduce thread sleeps/wakes

tellowkrinkle:
Sleeping and waking a thread with a condition variable is pretty expensive.  [You can benchmark it on your computer with this (may need `-lpthread`)](https://gist.github.com/tellowkrinkle/56591861a014bf1f478132a061749120), it takes about 8µs on the computers I tested

We had a number of places where we could end up with threads constantly sleeping and waking each other with condition variables, most notably in the software renderer when something needed to sync.

Improves speed for games that ran into those situations


tellowkrinkle:
Just going to be lazy and wait until after GSdx merge for this one, it'll fix all my linker issues and errors trying to `#include` Utilities headers from GSdx

tadanokojin:
#2648 maybe related

JordanTheToaster:
Tested this heavily and found some moderate improvements on a Ryzen 5 2600 and Ryzen 5 5600x but one especially large increase is Ratchet and Clank 3 going from 75-80% to full speed.

PR #4347, opened 2021-04-02
PS1-Mode: Restore CDXA support

kenshen112:
This PR is an attempt to restore pcsx's original CDXA support in an enhanced fashion using newer understanding from No$ and other sources.


kenshen112:
This PR's a draft because it's not quite ready sound-wise. It'll scream at ya real good but not much else

Issue #4342, opened 2021-03-30
[Feature Request] Updates to the brazilian portuguese translation


Felipefpl:
[Brazilian Portuguese Pcsx2 1.6.0.zip](https://github.com/PCSX2/pcsx2/files/6232256/Brazilian.Portuguese.Pcsx2.1.6.0.zip)
I read the main page but i dont know about how to make a pull request, posting here is easier.

RedDevilus:
Hello there, nice of you to make an issue for update on the translations. As of this moment i am waiting for the opportunity to have weblate (cloud translations) instead of the current system of Poedit which can be quite taxing for some translators with less confidence in their computer knowledge. Somewhere in the feature freeze in the coming months should normally be implemented and so a spree of translation can happen.

Felipefpl:
Hi, i dont think of Poedit as a bad thing, regarding a translation'S site i believe that transifex is the best system (crowdin sucks A LOT) as long as it doesnt allow to use of the "Reviewed" status that blocks your attempts to fix a translation because of a tyrant's behaviour from the coordinator. ;)

I dont have/use discord, sorry about that. But i'm on irc channel on freenode as aztec2.

Felipefpl:
I understand, are you one of the main devs that deal with translations? :)

RedDevilus:
I try to be :P, i'd like to think i have some better experience with handling translations.

Felipefpl:
Does my translation have any problem? I see it was submitted yet. :/

RedDevilus:
The problem as already stated is that your translations target parts which do not exist or are moved since 1.6 , so they are already unmergeable at this moment, this is why we wait till weblate is here so you can transfer your work to there.

Felipefpl:
i see, good to know, although it kinda worries me that i cant send files to weblate, at least not the one where i'm a translator.

RedDevilus:
Because it's not available yet for translators, hence translators wait.

PR #4351, opened 2021-04-06
Update pcsx2_Iconized.po

vyteking:
added some untranslated parts


RedDevilus:
Hello there, thanks for making translations. However translations are vastly out of sync with the current translatable strings due to waiting on another translation method via weblate.

https://github.com/PCSX2/pcsx2/issues/4342#issuecomment-810584011

lightningterror:
As Red said, in the meantime you can hop on our discord server, once we are in a feature freeze we will let everyone know how translations can be done via the web method.
Anyway I will be closing this now.

PR #4353, opened 2021-04-07
Input Recording: Cleanup inputRec utility namespace

xTVaser:
When these utility functions / namespace were originally made it wasn't done in the ideal way, this just quickly cleans that up.

### DIFF
--- a/pcsx2/Recording/Utilities/InputRecordingLogger.cpp
+++ b/pcsx2/Recording/Utilities/InputRecordingLogger.cpp
@@ -1,0 +1,50 @@
+﻿/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2021  PCSX2 Dev Team
+ *
+ *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
+ *  of the GNU Lesser General Public License as published by the Free Software Found-
+ *  ation, either version 3 of the License, or (at your option) any later version.
+ *
+ *  PCSX2 is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+ *  PURPOSE.  See the GNU General Public License for more details.
+ *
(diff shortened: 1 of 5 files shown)


pcsx2/Recording/Utilities/InputRecordingLogger.h
```diff
@@ -28,32 +28,7 @@
...
-
-	static void consoleMultiLog(std::vector<std::string> logs)
-	{
-		std::string log;
-		for (std::string l : logs)
-			log.append(fmt::format("[REC]: {}\n", l));
-
-		recordingConLog(log);
-	}
+	extern void log(const std::string log);
```
Warepire, line 31: There's absolutely no need to extern these.

xTVaser: Can you explain further?  My understanding is that leaving them static duplicates the functions for every translation unit (everywhere it's included).  And it was recommended to me to switch to `extern` instead.

Warepire: You already put the functions in a different translation unit without declaring them static, that's enough to avoid them being declared in every translation unit. So, functions are "extern" by default when they are not static.

Further declaring them `extern` is completely unnecessary.

pcsx2/Recording/Utilities/InputRecordingLogger.cpp
```diff
@@ -0,0 +1,50 @@
...
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "PrecompiledHeader.h"
+
+#include "InputRecordingLogger.h"
+
+namespace inputRec
+{
+	void log(const std::string log)
```
lightningterror, line 22: Would be nice to address the codacy complaints.

pcsx2/Recording/Utilities/InputRecordingLogger.h
xTVaser: Ah I see, thanks for the clarification.

PR #4361, opened 2021-04-11
Add Folder Utils path handler

kenshen112:
This is a PR that adds the Folder Utils path handler, class to better handle conversion to and from basic string types and file path handles like std::filesystem::path as well as conversion to and from nonstandard WX types.

### DIFF
--- a/common/src/Utilities/FileUtils.cpp
+++ b/common/src/Utilities/FileUtils.cpp
@@ -1,0 +1,138 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2021  PCSX2 Dev Team
+ *
+ *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
+ *  of the GNU Lesser General Public License as published by the Free Software Found-
+ *  ation, either version 3 of the License, or (at your option) any later version.
+ *
+ *  PCSX2 is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+ *  PURPOSE.  See the GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along with PCSX2.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "PrecompiledHeader.h"
+#include "FileUtils.h"
+#include <wx/file.h>
+#include <wx/stdpaths.h>
+#include <wx/filename.h>
+#include "ghc/filesystem.h"
+#include <wx/utils.h>
+
+// ---------------------------------------------------------------------------------
+//  Path namespace
+// ---------------------------------------------------------------------------------
+bool FileUtils::IsDirectoryWithinDirectory(const fs::path& base, const fs::path& dir)
+{
+    fs::path relativePath = fs::relative(fs::absolute(dir), fs::absolute(base));
+    if (relativePath.empty())
+    {
+        return false;
+    }
+    return relativePath.string().rfind("..", 0) != 0;
+}
+
+u64 FileUtils::GetFileSize(const fs::path &path)
+{
+    if (!fs::exists(path))
+        return -1;
+    return (u64)fs::file_size(path);
+}
+
+fs::path FileUtils::Normalize(const fs::path &src)
+{
+    if (src.empty())
+    {
+        return fs::path();
+    }
+    return src.lexically_normal();
+}
+
+fs::path FileUtils::GetFilename(const fs::path &src)
+{
+    if (fs::exists(src))
+        return fs::path(src).filename();
+    else
+        return fs::path();
+}
+
+fs::path FileUtils::GetFilenameWithoutExt(const fs::path &src)
+{
+    if (fs::exists(src))
+        return src.stem();
+    else
+    return fs::path();
+}
+
+fs::path FileUtils::GetDirectory(const fs::path &src)
+{
+    if (fs::exists(src))
+        return src.parent_path();
+    else
+        return fs::path();
+}
+
+fs::path FileUtils::GetExecutableDirectory()
+{
+    fs::path exePath(FileUtils::FromWxString(wxStandardPaths::Get().GetExecutablePath()));
+    return exePath.parent_path();
+}
+
+// returns the base/root directory of the given path.
+// Example /this/that/something.txt -> dest == "/"
+fs::path FileUtils::GetRootDirectory(const fs::path &src)
+{
+    if (fs::exists(src))
+        return src.root_directory();
+    else
+        return fs::path();
+}
+
+bool FileUtils::Mkdir(const fs::path& path)
+{
+    try
+    {
+        fs::create_directories(path); // An attempt to create the User mode Dir which already exists
+    }
+    catch(const std::exception& e)
+    {
+        std::cerr << e.what() << '\n';
+        return false;
+    }
+    return true;
+}
+
+bool FileUtils::Rmdir(const fs::path& path)
+{
+    try
+    {
+        fs::remove_all(path);
+    }
+    catch(const std::exception& e)
+    {
+        std::cerr << e.what() << '\n';
+        return false;
+    }
+    return true;
+}
+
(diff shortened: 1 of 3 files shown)


tellowkrinkle (commented):
IMO we should not be encouraging mixing of paths and strings.  Paths are paths, strings are strings.  Convert at the point a string goes from being "maybe or maybe not a path" to "definitely a path".  If you do this, you'll never need to do path operations on strings, which this PR seems to add a lot of

common/include/Utilities/FolderUtils.h
```diff
@@ -0,0 +1,46 @@
+#pragma once
```
F0bes, line 1: License header

common/include/Utilities/FolderUtils.h
```diff
@@ -0,0 +1,46 @@
...
+namespace Path
+{
+    extern bool IsRelative(const std::string& path);
+
+    /************************************************************************
+     * @brief Indicates whether dir2 is contained within dir1
+     * @param dir1 Directory that may or may not contain the second directory
+     * @param dir2 The second directory
+     * @return true/false
+     *************************************************************************/
```
F0bes, line 21: I'd try and keep comments styles consistent within the codebase.
I personally haven't seen this type before.

common/src/Utilities/FolderUtils.cpp
```diff
@@ -0,0 +1,153 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2010  PCSX2 Dev Team
```
F0bes, line 2: It's 2021 :P

common/src/Utilities/FolderUtils.cpp
```diff
@@ -0,0 +1,153 @@
...
+std::string Path::ReplaceExtension(const wxString &src, const wxString &ext)
+{
+    wxFileName jojo(src);
+    jojo.SetExt(ext);
+    return std::string(jojo.GetFullPath().ToUTF8());
+}
+
+std::string Path::ReplaceFilename(const wxString &src, const wxString &newfilename)
+{
+    wxFileName jojo(src);
```
F0bes, line 83: Non-self-explanatory variable naming.
JoJo BowBow?
That one anime?

common/include/Utilities/FolderUtils.h
xTVaser: It looks like it's just doxygen with the extra row of astrix.  imo I like them since VS now supports them in the latest release, but if we are going to use doxygen, they should be used across the entire namespace.

common/include/Utilities/FolderUtils.h
```diff
@@ -0,0 +1,46 @@
...
+    extern bool IsDirectoryWithinDirectory(fs::path base, fs::path dir);
+
+    // Returns -1 if the file does not exist.
+    extern s64 GetFileSize(const fs::path& file);
```
xTVaser, line 25: `extern`s are not required, as I've learned here https://github.com/PCSX2/pcsx2/pull/4353

common/include/Utilities/FolderUtils.h
```diff
@@ -0,0 +1,46 @@
...
+    extern std::string GetFilenameWithoutExt(const fs::path& src);
+    extern std::string GetRootDirectory(const wxString& src);
+    extern fs::path GetExecutableDirectory();
+    extern wxString ToWxString(const fs::path&);
+    extern bool CreateFolder(const fs::path& path);
+    extern fs::path FromWxString(const wxString&);
+    // Is the folder empty
+    extern bool Empty(fs::path);
+    // Does the path exist
+    extern bool DoesExist(const fs::path& path);
```
xTVaser, line 45: these two are definitely not required, if you already have an `fs::path` you can use the relevant empty/exists functions wherever this is being called instead.

common/src/Utilities/FolderUtils.cpp
```diff
@@ -0,0 +1,153 @@
...
+    return (s64)fs::file_size(path);
+}
+
+// Note std filesystem contains no clean way to normalize so we must use GUI provided functionality
+std::string Path::Normalize(const std::string &src)
+{
+    wxFileName normalize(Path::ToWxString(src));
+    normalize.Normalize();
+    return Path::FromWxString(normalize.GetFullPath());
+}
```
xTVaser, line 55: If we truly need normalization, we should avoid wx and implement the following algorithm described in the docs instead https://en.cppreference.com/w/cpp/filesystem/path

common/include/Utilities/FolderUtils.h
```diff
@@ -0,0 +1,46 @@
...
+#include <iostream>
+#include <iomanip>
+#include "FixedPointTypes.inl"
+#include "ghc/filesystem.h"
+#include "Path.h"
+
+namespace fs = ghc::filesystem;
+
+
+namespace Path
```
xTVaser, line 12: Being as this seems to be a completely new namespace for future use, I'd recommend cutting out the fat that isn't needed anymore.  A lot of the functions inside are there because the old wx code relied on wxStrings heavily and such.  Going forward, if something is operating on a path/file, it should be using an `fs::path` to begin with, which removes the need for a lot of these functions.

I'm also not crazy about the non-descript namespace name of `Path`, I'd recommend leaving the old namespace as still `Path`, and make this one something more descriptive like `FileUtils`.  That will also have the side effect of minimizing the amount of `Path` -> `wxPath` changes you had to make here.

common/src/Utilities/FileUtils.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+fs::path FileUtils::GetFilenameWithoutExt(const fs::path &src)
+{
+    return src.stem();
+}
+
+fs::path FileUtils::GetDirectory(const fs::path &src)
+{
+    return src.parent_path();
+}
```
tellowkrinkle, line 93: Why are we adding helper functions that just call a single method on the `fs::path`?  Why not just call that method directly?  Future people who know C++ `std::filesystem` but not our special `FileUtils` will probably appreciate if you used the `fs::path` methods

common/src/Utilities/FileUtils.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+fs::path FileUtils::MakeAbsolute(const fs::path &src)
+{
+    return ghc::filesystem::absolute(src);
+}
+
+fs::path FileUtils::Combine(const fs::path &srcPath, const fs::path &srcFile)
+{
+    return (srcPath / srcFile).make_preferred();
```
tellowkrinkle, line 65: I'm still fairly sure the `make_preferred` is completely unnecessary, and we should just use the `/` operator

common/src/Utilities/FileUtils.cpp
kenshen112: https://github.com/PCSX2/pcsx2/blob/6cc969a89ccfa24c6f82a94154855f7934ed7f64/common/src/Utilities/PathUtils.cpp#L125 Going off the original's functions

common/src/Utilities/FileUtils.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <wx/utils.h>
+
+// ---------------------------------------------------------------------------------
+//  Path namespace
+// ---------------------------------------------------------------------------------
+bool FileUtils::IsRelative(const fs::path &path)
+{
+    if (!fs::exists(path))
+        return false;
+    return fs::path(path).is_relative();
```
Warepire, line 31: Unnecessary copy-construction of path just to test if it's relative.

Warepire (commented):
Are the "invalid paths" obfuscations really necessary? At least include some header documentation about it. As it looks now the utilities don't seem like they originate from any actual need, and only serve as some sort of basic abstraction of ghc::filesystem.

kenshen112:
> Are the "invalid paths" obfuscations really necessary? At least include some header documentation about it. As it looks now the utilities don't seem like they originate from any actual need, and only serve as some sort of basic abstraction of ghc::filesystem.

It's kind of a need in that I'm making a copy cat lib to PathUtils.cpp / the pcsx2 custom WXDir class

lightningterror:

PR #4362, opened 2021-04-12
Rename main menu back to System

tellowkrinkle:
There's already a `PCSX2` menu on macOS, adding a second one is incredibly confusing

### DIFF
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -607,7 +607,7 @@ MainEmuFrame::MainEmuFrame(wxWindow* parent, const wxString& title)
 	// Initial menubar setup.  This needs to be done first so that the menu bar's visible size
 	// can be factored into the window size (which ends up being background+status+menus)
 
-	m_menubar.Append(&m_menuSys, _("&PCSX2"));
+	m_menubar.Append(&m_menuSys, _("&System"));
 	m_menubar.Append(&m_menuCDVD, _("CD&VD"));
 	m_menubar.Append(&m_menuConfig, _("&Config"));
 	m_menubar.Append(&m_menuWindow, _("&Window"));


tellowkrinkle:
If people prefer `PCSX2` on Windows/Linux I can ifdef it for macOS but if no one cares we might as well switch it back

Mrlinkwii (approved):

lightningterror:
I was also thinking about renaming Window to Debug or Debug Menu since only debug stuff is in there, and maybe move blockdump option there too.

arcum42:
I'm fine with renaming the menu back. I'd originally intended it to merge with the MacOS menu of the same name.

With Window, the intent was to be adding other windows that would also be in that menu, such as a game library window, and saveslot window.

tellowkrinkle:
Ahh I see
In macOS it's actually pretty rare for applications to put much of their functionality in the main application menu

Usually it's just system stuff plus `About`, `Preferences`, and `Quit`, most operations go under the `File` menu

wx seems to populate `About` and `Quit` for us, so the only thing missing is `Preferences`, though for some reason our about just says `About`, while most applications would include the application name (e.g. `About PCSX2`)

In general I think wx does a good job of making PCSX2 have a natural feeling menu bar (a lot better than Discord's or GIMP's at least), the only weirdness is the fact that the things in the main `PCSX2` menu are duplicated in other menus (`Quit` is also in the `System` menu, `About` is also in the `Help` menu)

arcum42:
We could always ifdef the about and quit menu items so that they weren't duplicated in MacOS.

I used to be more familiar with MacOS, but it's been years, and I was more familiar with classic MacOS then MacOS X, which is a totally different operating system.

With GIMP, none of the widgets are native, since it's using Gtk. (In fact, Gtk started out in Gimp, and got split off.) And, of course, Discord is Electron. With wx, everything is native, just with wx in between, as it were...

PR #4363, opened 2021-04-12
Format GSdx

tellowkrinkle:
I wasn't a huge fan of the blind clang-formats that were run during the previous plugin merges, so I figured I'd format before the merge so no format had to be done during the merge.

In addition, I'd much rather rebase my open PRs for formatting separately from the merge, rather than both at once

Formats GSdx using clang-format, with any dumb formatting decisions clang-format made undone


da2ce7:
Maybe separate this into two pull requests.

1. Merging a script and formatting rules.
2. Preform the actual formatting of the repository according to the previously merged script and rules.

This allows for a more easy review, as the second pull-request will be deterministically verifiable by others by running the same script and comparing the diff.

tellowkrinkle:
We already have a clang format file.  It's what I used for about 90% of this PR

But it doesn't know that humans can find deviating from the normal rules more readable in some cases

It also doesn't understand macros

And I don't think it ever will

So we're either going to have terrible looking code or we're going to have to live with the fact that not everything will perfectly match what clang-format thinks it should be.  And personally, I prefer the latter

iMineLink:
It takes great courage to do, but must be done. Some deviations from the standard style can indeed be handled manually. Thanks for this PR.

angstsmurf:
In the long run, it might be worth the trouble to add `// clang-format off` comments wherever Clang can't be configured to act sensibly.

tellowkrinkle:
> In the long run, it might be worth the trouble to add `// clang-format off` comments absolutely everywhere that Clang can't be configured to act sensibly.

I did it some places, but not everywhere.  I have two issues with forcing `// clang-format` comments around everything that clang format doesn't like:

1. Having to add the comments makes me more likely to just let clang-format do stupid things when formatting.  If you ask whether I prefer
```cpp
return (((c & 0x8000) ? TEXA.TA1 : (!TEXA.AEM | c) ? TEXA.TA0 : 0) << 24) ...
```
or
```cpp
return (((c & 0x8000) ? TEXA.TA1 : (!TEXA.AEM | c) ? TEXA.TA0 : 
                                                     0)
        << 24) ...
```
I definitely prefer the former, but if you told me I had to add `// clang-format on` and `// clang-format off` to it or choose the latter, I probably would just let clang-format go with the latter.
2. `// clang-format off` affects everything, not just the issue you care about.  If we had put the comments around [these functions, which want digit alignment of some numbers](https://github.com/PCSX2/pcsx2/blob/fafbb3cc63080cbc6a9f2d46a3416357532697fe/plugins/GSdx/GSVector4i.h#L2046-L2054), they would have never been reformatted to add spaces before and after the content of the `{}` (which I did do in this commit, because clang-format edited them, then I looked at the edit and fixed it up to keep the alignment but not the lack of spacing)

PR #4371, opened 2021-04-17
Input-Recording: V2 File format

xTVaser:
### Description

My main motivation for breaking the file format for hopefully the final time is because of several reasons:
- The current format was not designed with flexibility/extensibility in mind.  It has a lot of cruft that we'd like to cleanup, etc.
- I needed a way to make an input recording file that I can mark certain inputs as ignored.  In other words, different "types" of input recordings beyond just how they begin (savestate/poweron)

Additionally, I corrected our handling of unicode strings in the recording logging namespace, and switched the vast majority away from WX's file handling in favor of ghc::filesystem.  The new file format implementation also moves away from C libraries in favor of the C++ standard lib.

For utf8 handling I referenced - http://utf8everywhere.org/
For binary file format best practices - https://fadden.com/tech/file-formats.html

This PR still likely needs some cleanup, but functionality it's decently tested (i can record/playback and accurately convert from V1).  However, I wanted to put it up "early" to get feedback, mostly from @sonicfind as I know his upcoming MultiTap work may benefit from this (or be ruined if i made a mistake!)


common/include/Utilities/StringUtils.h
```diff
@@ -0,0 +1,38 @@
...
+#include "wx/string.h"
+
+namespace StringUtils
+{
+	namespace UTF8
+	{
+// There should be no reason for a unix environment to have to convert string->wstring
+// See - http://utf8everywhere.org/#how.cvt
+#ifdef _WIN32
+		std::string narrow(const wchar_t* str);
```
Warepire, line 27: If C++17 is allowed: Using std::wstring_view and std::string_view for the functions below lets you implement everything with one function. No need for overloads.

common/src/Utilities/FileUtils.cpp
```diff
@@ -0,0 +1,110 @@
...
+}
+
+fs::path FileUtils::appendToFilename(fs::path file_path, const std::string& utf8_str)
+{
+#ifdef _WIN32
+	std::wstring curr_file_extension = file_path.extension().wstring();
+	std::wstring new_name = file_path.filename().replace_extension().wstring() + StringUtils::UTF8::widen(utf8_str) + curr_file_extension;
+	return file_path.replace_filename(fs::path(new_name));
+#else
+	std::string curr_file_extension = file_path.extension().wstring();
```
Warepire, line 77: Copy-paste error with the wstring at the end?

pcsx2/Recording/InputRecording.cpp
```diff
@@ -284,29 +297,29 @@ void InputRecording::SetupInitialState(u32 newStartingFrame)
 	if (state != InputRecordingMode::Replaying)
 	{
 		inputRec::log("Started new input recording");
-		inputRec::consoleLog(fmt::format("Filename {}", std::string(inputRecordingData.GetFilename())));
+		inputRec::consoleLog(fmt::format("Filename {}", inputRecordingData.getFileName().string()));
```
Warepire, line 300: Teach format to print ghc::filesystem::path instead?

Warepire:
General suggestion for the file format: If you want maximum flexibility, binary is not the best choice for the header. How about a Yaml header + binary input data for each frame. Doesn't even need to be the same file if you don't want to mix the different types of data if you use something like tar and optionally zstd to compress it on the fly.

xTVaser:
That's a good point, I abandoned the idea of YAML for the entire file for performance reasons, but I like the idea for just the header.

On the subject of compression, it was something we talked about in the past and the verdict was that it doesn't really feel worth it.  Hours and hours of recording takes up less file-size than a single PCSX2 save-state to put it in perspective.

Warepire:
I didn't know if it would make sense or not which is why I specified compression as optional, and you proved that it doesn't make sense to compress. So that part can definitely be ignored.

xTVaser:
Thought about the YAML change a bit more, it's problematic for just the header as well.  While it would be more flexible, there are values within the header that get updated every frame, and since the YAML data is going to be variable size, we can't selectively update those values in-place.  

I don't we're sacrificing too much flexibility wise.  No matter the format, new fields have to be handled using the version schema appropriately.  While it is certainly a bit more cumbersome to work with the binary format compared to the alternative, it affords us some optimizations we need while still keeping things fairly organized/centralized.

Warepire:
Updating in place can be done if you keep the data in a struct or something which you serialize to yaml as necessary. But if binary format overall is easier to manage in this situation, then I have nothing against the binary format.

xTVaser:
We should disable the following codacy remark lint check:
image
It's apparently prone to false-positives (all the flagged ones here are) https://github.com/remarkjs/remark-lint/issues/78

xTVaser:
Reasonably confident with this, did the following tests:
- Confirmed there were no issues removing the following workaround in linux: https://github.com/PCSX2/pcsx2/blob/721253bf49aa763e55d47ac93891ca95950ba7d1/pcsx2/Recording/NewRecordingFrame.cpp#L101-L107
- Created an input recording from power-on and from a savestate on `master` (legacy recordings)
  - Was able to play both back successfully.  Each were converted while the originals were left in-tact.
- Created an input recording from power-on and from a savestate on this branch (use the new format from the beginning)
  - Both of these also worked fine.

common/src/Utilities/StringUtils.cpp
```diff
@@ -0,0 +1,71 @@
...
+	return output;
+}
+
+std::wstring StringUtils::UTF8::widen(const std::string_view& str)
+{
+	auto const size =
+		MultiByteToWideChar(CP_UTF8, 0, str.data(), str.size(), nullptr, 0);
+
+	std::wstring output;
+	output.resize(size);
```
Warepire, line 50: You can save a line (and possibly a reallocation) if you pass size when constructing output: `std::wstring output(size);`

xTVaser:
Closing this PR, very out of date, especially with the migration to Qt -- and I've since realized I need to radically alter the file format to do more useful things anyway.  This format still holds too true to the original V1 and would be useless for planned features.

PR #4373, opened 2021-04-18
CMake file cleanup

tellowkrinkle:
- Increase cmake minimum version to 3.10 (seems to be supported by most recent-ish distros)
- Consistently use tabs on cmake files (previously was all over the place), add `?w=1` to the diff URL to view a cleaner diff
- No longer allow compiling only parts of the emulator.  If you don't have enough dependencies, the configure fails instead of continuing while not actually building anything useful (common source of confusion when building the emulator)
- Stop completely overriding cmake build settings.  May add some flags on top, but doesn't wipe out the existing flags
- Settings for different build types work properly in generators that support multiple build types at once (previously specifying `-DCMAKE_BUILD_TYPE=Debug` would make the `Debug`, `Release`, etc targets all identical to the `Debug` target)
- Replace custom LTO flag with cmake-builtin `CMAKE_INTERPROCEDURAL_OPTIMIZATION`
- Replace `Prof` build type with the cmake-builtin `RelWithDebInfo`
- Switch to imported targets and stop globally adding things to include directories
  - This may break things, as it greatly reduces the number of include directories for things (especially the `Utilities` build), please test
- Stop globally adding warning flags
  - Reduces warnings from `libchdr` as they it longer has our warning flags
- Enforce the `Utilities/` prefix for `Utilities` headers when referenced from the main `pcsx2` build
- Remove X11 dependency from the macOS build (should speed up CI dependency installation by a few minutes)


cmake/SearchForStuff.cmake
```diff
@@ -290,4 +290,3 @@ if(NOT USE_SYSTEM_YAML)
 endif()
 
 add_subdirectory(3rdparty/libchdr/libchdr EXCLUDE_FROM_ALL)
-include_directories(3rdparty/libchdr/libchdr/include)
```
kenshen112, line 293: Wouldn't this break the search for it as a submodule if we wanted to use the local copy?

CMakeLists.txt
```diff
@@ -44,15 +42,11 @@ if(NOT NO_TRANSLATION)
 endif()
 
 # make common
-if(common_libs)
-	add_subdirectory(common/src/Utilities)
-	add_subdirectory(common/src/x86emitter)
-endif()
+add_subdirectory(common/src/Utilities)
```
kenshen112, line 45: We should think about breaking Common into it's own sub project with a CMakeLists

cmake/FindPulseAudio.cmake
```diff
@@ -35,6 +36,13 @@ find_package_handle_standard_args(PulseAudio
 	VERSION_VAR PULSEAUDIO_VERSION_STRING
 )
 
+if(PULSEAUDIO_LIBRARY AND NOT TARGET PulseAudio::PulseAudio)
```
kenshen112, line 39: Consider adding that option for Jack

build.sh
```diff
@@ -191,7 +191,7 @@ for ARG in "$@"; do
         --clang             ) useClang=1 ;;
         --intel             ) useIcc=1 ;;
         --cppcheck          ) cppcheck=1 ;;
-        --dev|--devel       ) flags="$flags -DCMAKE_BUILD_TYPE=Devel"   ; build="$root/build_dev";;
+        --dev|--devel       ) flags="$flags -DCMAKE_BUILD_TYPE=RelWithDebInfo"; build="$root/build_dev";;
```
kenshen112, line 194: I'd prefer we keep the devel option as RelwithDebInfo feels like it should be release with extra sauce

cmake/SearchForStuff.cmake
tellowkrinkle: No more than every other place I deleted an `include_directories`

cmake/FindPulseAudio.cmake
tellowkrinkle: ??? We don't directly depend on Jack at all

build.sh
tellowkrinkle: Done

CMakeLists.txt
tellowkrinkle: If we did it would literally be one file with the content
```cmake
add_subdirectory(src/Utilities)
add_subdirectory(src/x86emitter)
```

kenshen112: Not nessacarily. I meant for Common having a list of each source and header so it's 

`
set(sources 
src/Utilities/path.h
)
`

Ziemas:
Devel builds aren't getting debug symbols built in, would be nice if they did.

CMakeLists.txt
tellowkrinkle: It already does, in a `CMakeLists.txt` in the directory referenced by the `add_subdirectory`
How do you think the `add_subdirectory` works?  Magic?

arcum42:
I mostly approve, though I have a lot of difficulty reading the diff on github due to all the formatting changes, and I kind of wish those had been separate.

The fact that when I tried to compile, it errored out with:
../pcsx2/CDVD/ChdFileReader.h:18:10: fatal error: libchdr/chd.h: No such file or directory

could be an issue. That file is there, so maybe a path didn't come over properly or something. It compiled in the automatic checks though, so that seems a bit odd...

tellowkrinkle:
> I mostly approve, though I have a lot of difficulty reading the diff on github due to all the formatting changes, and I kind of wish those had been separate.
> 

Add `?w=1` to the url and it should hide whitespace changes

> The fact that when I tried to compile, it errored out with:
> 
> ../pcsx2/CDVD/ChdFileReader.h:18:10: fatal error: libchdr/chd.h: No such file or directory

You need to `git submodule update`, I updated to the libchdr that fixes that in the PR

sjnewbury:
> 
>     * Replace custom LTO flag with cmake-builtin `CMAKE_INTERPROCEDURAL_OPTIMIZATION`
LTO is Link Time Optimization, IPA is INTERPROCEDURAL_OPTIMIZATION..?

Also, why drop support for system libraries?  FWIW, being as I work on Gentoo stuff, and maintain a PCSX2 ebuild I will have to revert this patch this since it's Gentoo policy not to use bundled libraries, see here: https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies

This patch series fails to build on my system since it's not including librt on the PCSX2 executable link command.  Testing prior to commit b8299fa is fine.
```

/usr/bin/x86_64-pc-linux-gnu-ld.bfd: pcsx2/CMakeFiles/PCSX2.dir/GS/GS.cpp.o: undefined reference to symbol 'shm_open@@GLIBC_2.2.5'
/usr/bin/x86_64-pc-linux-gnu-ld.bfd: /lib64/librt.so.1: error adding symbols: DSO missing from command line
```

tellowkrinkle:
> LTO is Link Time Optimization, IPA is INTERPROCEDURAL_OPTIMIZATION..?

[CMake seems to use them interchangeably](https://cmake.org/cmake/help/latest/module/CheckIPOSupported.html)

> Also, why drop support for system libraries?  FWIW, being as I work on Gentoo stuff, and maintain a PCSX2 ebuild I will have to revert this patch this since it's Gentoo policy not to use bundled libraries, see here: https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies

This shouldn't have dropped support for any system libraries.  If it did, please specify which ones so that can be fixed.

> This patch series fails to build on my system since it's not including librt on the PCSX2 executable link command.  Testing prior to commit [b8299fa](https://github.com/PCSX2/pcsx2/commit/b8299fa5f81bc3015dd1252b9787257f41613124) is fine.
> 
> 
> ```
> 
> /usr/bin/x86_64-pc-linux-gnu-ld.bfd: pcsx2/CMakeFiles/PCSX2.dir/GS/GS.cpp.o: undefined reference to symbol 'shm_open@@GLIBC_2.2.5'
> /usr/bin/x86_64-pc-linux-gnu-ld.bfd: /lib64/librt.so.1: error adding symbols: DSO missing from command line
> ```

Can you open an issue for that?

tadanokojin:
@sjnewbury can you confirm #4579 works?

sjnewbury:
@tadanokojin It may take me a few days, but I will confirm as soon as possible.

tadanokojin:
thanks we appreciate it

jeandestouches:
Hello,
I was having the same issue as sjnewbury, unfortunately, it still fails to build.
Here is the bug report from August (latest commit shows the same error)
https://bugs.gentoo.org/809614
or direct link for the build log : https://paste.debian.net/plainh/f2cf7526

If it is not related to this issue, sorry for the noise.

PR #4382, opened 2021-04-21
Input Recording: Resolve crash when closing emulator involving GUI elements

xTVaser:
### Description of Changes

The gist is we paused the CoreThread without resuming it (input recording's `Resume()` method will resume on the next VSync), the shutdown routine triggers it when all required window elements are already cleaned up, hence the crash.

image

### Rationale behind Changes

Solves a bug that results in an unexpected GUI error window when closing the emulator after very specific/edge-casey steps.

### Suggested Testing Steps

Reproduction Steps:
- boot game so gs window is open
- close the GS window
- open the dialog for a new input recording, then close it
- close pcsx2, should error.

fyi @sonicfind incase this breaks something to do with input-recording's frame locking behaviour.


RedDevilus:
Can you update the copyright headers to 2021 instead of 2020?

sonicfind:
Frame lock seems to be unaffected. As long as you're not getting that error anymore, then it should be good.

PR #4384, opened 2021-04-22
Core: Purge OS specific CPU usage providers

F0bes:
In the past, the goal was to have os specific thread usage providers. This didn't end up happening and we now use the _DefaultCpuUsageProvider_.

### Description of Changes
I removed the broken Windows provider that tried to use COM and changed the _DefaultCpuUsageProvider_ class to just _CpuUsageProvider_.
I didn't remove the _BaseCpuUsageProvider_ class in case the default provider doesn't work on some special platform and a special provider is needed.

### Rationale behind Changes
The MSW provider that was ifdef'd out was broken and not used for years. There's also no point in having nonportable providers.

### Suggested Testing Steps
This pr either works on not. If you still have working EE, GS, MTVU, and UI usage percents, there's a 99.9% chance it's working properly. 

~~Note: I need someone to test this on Linux :)~~ Looks to be working on linux


lightningterror (approved):
Looks good, needs testing.

JordanTheToaster:
Works perfectly for me on Windows.

Issue #2753, opened 2018-12-15
[GSdx SW] Zettai Zetsumei Toshi 2 (Raw Danger) Trial A - Rendering Issue


Souzooka:
This issue is being posted after being tested on PCSX2 dev-2733 concerning the game serial [SLPM-60254]. This occurs on both the OpenGL and DX11 renderer.

Playing the game on software rendering mode reveals inaccurate emulation of a depth+tonemapping-based postprocessing effect.

Recording: https://gfycat.com/RipeVigilantHart

Images:
SW rendering:
![](https://cdn.discordapp.com/attachments/480478597799346186/523446640527409152/gsdx_20181215002958.png)

HW rendering:
![](https://cdn.discordapp.com/attachments/480478597799346186/523446549003501579/gsdx_20181215002922.png)

This problem does not occur on the retail release. Of course, this version of the game is not readily available to be obtained at this time (unless you get really lucky on JP Ebay or some such), but I would be happy to provide any additional information required upon request.

gregory38:
Hw is wrong too. You can notice a small top left corner rectangle (which contains the post processing).

gregory38:
Could you check sw renderer without extra threads ?

Souzooka:
The same effect still occurs when using 1 rendering thread.

atomic83GitHub:
This problem also happen in the final game  but the hardware render have a crc hack to avoid this glitch (and prevent insanelly vram usage Wen upscaling).

tadanokojin:
I believe @Souzooka confirmed for me this is working on real hw.
This version of the game appears to do this effect without a TS.

tadanokojin:
Part of the issue appears to be that the channel shuffle is using a SCANMSK to mask out the odd y coordinate:
![rd](https://cdn.discordapp.com/attachments/430072460671451137/796845211447590922/006392.png)

tadanokojin:
After further investigation I realized I'd made a goof. Full issue is SCANMSK.
![000533](https://user-images.githubusercontent.com/26046960/119899394-0e49d100-bf11-11eb-8420-bd8fb06ef128.png)

refractionpcsx2:
Okay Confirmed with @Souzooka that it's fixed.

PR #4385, opened 2021-04-23
GS: Simulate scan mask (fix transparency in MGS2 & MGS3)

Sergeanur:
### Description of Changes
This adds simulation of a PS2 scanline masked drawing on hardware renderer. Masked drawing allowed to skip rendering of primitives on even/odd lines (depending on the mask). As different games may have used this mask for different techniques several options were implemented:
изображение

**1. Off (current behaviour)**
Ignores the scanmask. For cases when the end result is just fine without it.
изображение

**2. Unscaled (set as a default setting as this is what PS2 does)**
Do the masking with PS2 resolution in mind. The prohibited lines are skipped by failing depth test in fragment shader/pixel shader.
изображение

**3. Scaled**
Same as above except actual internal resolution is used.
изображение

**4. Transparent**
Suitable if masking was used to create an illusion of transparency.
No lines are skipped when drawing. Geometry is rendered with half-transparency (blending forcibly enabled) and disabled z-write.
изображение

### Rationale behind Changes
As shown on screenshots above, this fixes transparency issues in Metal Gear Solid games. It should also fix issues in games that used scanmask as well (I don't know any other games that do).

### Suggested Testing Steps
MGS2: check Otacon's glasses, also water (when you're on top of it). Bubbles should be seen around player's model when you swim. When entering first person view you should see small particles underwater.
MGS3: cockpit in the first cutscene.

I have not tested further in the game nor I haven't checked with different varieties of HW hacks and blending modes enables, so it would be good if someone could test it too, though I assume everything should be ok.

### DIFF
--- a/pcsx2/GS/Renderers/Common/GSRenderer.cpp
+++ b/pcsx2/GS/Renderers/Common/GSRenderer.cpp
@@ -220,7 +220,9 @@ bool GSRenderer::Merge(int field)
 			off.x = tex[i]->GetScale().x * frame_diff.x;
 		}
 
-		if (display_diff.y >= 4) // Shouldn't this be >= 2?
+		if (m_scanmask_used && display_diff.y == 1) // Scanmask effect wouldn't look correct if we scale the offset
+			off.y = display_diff.y;
+		else if (display_diff.y >= 4) // Shouldn't this be >= 2?
 		{
 			off.y = tex[i]->GetScale().y * display_diff.y;
 
--- a/pcsx2/GS/Renderers/SW/GSRasterizer.cpp
+++ b/pcsx2/GS/Renderers/SW/GSRasterizer.cpp
@@ -40,6 +40,7 @@ GSRasterizer::GSRasterizer(IDrawScanline* ds, int id, int threads, GSPerfMon* pe
 	, m_ds(ds)
 	, m_id(id)
 	, m_threads(threads)
+	, m_scanmsk_value(0)
 {
 	memset(&m_pixels, 0, sizeof(m_pixels));
 	m_primcount = 0;
@@ -158,6 +159,7 @@ void GSRasterizer::Draw(GSRasterizerData* data)
 	m_scissor = data->scissor;
 	m_fscissor_x = GSVector4(data->scissor).xzxz();
 	m_fscissor_y = GSVector4(data->scissor).ywyw();
+	m_scanmsk_value = data->scanmsk_value;
 
 	switch (data->primclass)
 	{
@@ -834,7 +836,7 @@ void GSRasterizer::DrawSprite(const GSVertexSW* vertex, const u32* index)
 
 	GSVertexSW scan = v[0];
 
-	if (m_ds->IsSolidRect())
+	if ((m_scanmsk_value & 2) == 0 && m_ds->IsSolidRect())
 	{
 		if (m_threads == 1)
 		{
@@ -1158,6 +1160,7 @@ void GSRasterizer::Flush(const GSVertexSW* vertex, const u32* index, const GSVer
 
 void GSRasterizer::DrawScanline(int pixels, int left, int top, const GSVertexSW& scan)
 {
+	if ((m_scanmsk_value & 2) && (m_scanmsk_value & 1) == (top & 1)) return;
 	m_pixels.actual += pixels;
 	m_pixels.total += ((left + pixels + (PIXELS_PER_LOOP - 1)) & ~(PIXELS_PER_LOOP - 1)) - (left & ~(PIXELS_PER_LOOP - 1));
 	//m_pixels.total += ((left + pixels + (PIXELS_PER_LOOP - 1)) & ~(PIXELS_PER_LOOP - 1)) - left;
@@ -1169,6 +1172,7 @@ void GSRasterizer::DrawScanline(int pixels, int left, int top, const GSVertexSW&
 
 void GSRasterizer::DrawEdge(int pixels, int left, int top, const GSVertexSW& scan)
 {
+	if ((m_scanmsk_value & 2) && (m_scanmsk_value & 1) == (top & 1)) return;
 	m_pixels.actual += 1;
 	m_pixels.total += PIXELS_PER_LOOP - 1;
 
--- a/bin/resources/shaders/dx11/tfx.fx
+++ b/bin/resources/shaders/dx11/tfx.fx
@@ -50,6 +50,7 @@
 #define PS_PABE 0
 #define PS_DITHER 0
 #define PS_ZCLAMP 0
+#define PS_SCANMSK 0
 #endif
 
 #define SW_BLEND (PS_BLEND_A || PS_BLEND_B || PS_BLEND_D)
@@ -719,6 +720,13 @@ PS_OUTPUT ps_main(PS_INPUT input)
 	float4 C = ps_color(input);
 
 	PS_OUTPUT output;
+
+	if (PS_SCANMSK & 2)
+	{
+		// fail depth test on prohibited lines
+		if ((int(input.p.y) & 1) == (PS_SCANMSK & 1))
+			discard;
+	}
 
 	if (PS_SHUFFLE)
 	{
--- a/bin/resources/shaders/opengl/tfx_fs.glsl
+++ b/bin/resources/shaders/opengl/tfx_fs.glsl
@@ -718,6 +718,11 @@ void ps_blend(inout vec4 Color, float As)
 
 void ps_main()
 {
+#if PS_SCANMSK & 2
+    // fail depth test on prohibited lines
+ 	if ((int(gl_FragCoord.y) & 1) == (PS_SCANMSK & 1))
+ 	 	discard;
+#endif
 #if ((PS_DATE & 3) == 1 || (PS_DATE & 3) == 2)
 
 #if PS_WRITE_RG == 1
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -34,6 +34,7 @@ GSState::GSState()
 	, m_crc(0)
 	, m_options(0)
 	, m_frameskip(0)
+	, m_scanmask_used(false)
 {
 	// m_nativeres seems to be a hack. Unfortunately it impacts draw call number which make debug painful in the replayer.
 	// Let's keep it disabled to ease debug.
@@ -219,6 +220,8 @@ void GSState::Reset()
 	m_vertex.tail = 0;
 	m_vertex.next = 0;
 	m_index.tail = 0;
+
+	m_scanmask_used = false;
 }
 
 void GSState::ResetHandlers()
@@ -1098,6 +1101,8 @@ void GSState::GIFRegHandlerSCANMSK(const GIFReg* RESTRICT r)
 		Flush();
 
 	m_env.SCANMSK = (GSVector4i)r->SCANMSK;
+	if (m_env.SCANMSK.MSK & 2)
+		m_scanmask_used = true;
 }
 
 template <int i>
--- a/pcsx2/GS/GSState.h
+++ b/pcsx2/GS/GSState.h
@@ -159,6 +159,8 @@ protected:
 	GSVector4i m_ofxy;
 	bool tex_flushed;
 
+	bool m_scanmask_used;
+
 	struct
 	{
 		GSVertex* buff;
--- a/pcsx2/GS/Renderers/Common/GSDevice.h
+++ b/pcsx2/GS/Renderers/Common/GSDevice.h
@@ -250,7 +250,10 @@ struct alignas(16) GSHWDrawConfig
 				u32 point_sampler : 1;
 				u32 invalid_tex0 : 1; // Lupin the 3rd
 
-				u32 _free2 : 6;
+				// Scan mask
+				u32 scanmsk : 2;
+
+				u32 _free2 : 4;
 			};
 
 			u64 key;
--- a/pcsx2/GS/Renderers/DX11/GSTextureFX11.cpp
+++ b/pcsx2/GS/Renderers/DX11/GSTextureFX11.cpp
@@ -189,6 +189,7 @@ void GSDevice11::SetupPS(PSSelector sel, const GSHWDrawConfig::PSConstantBuffer*
 		sm.AddMacro("PS_PABE", sel.pabe);
 		sm.AddMacro("PS_DITHER", sel.dither);
 		sm.AddMacro("PS_ZCLAMP", sel.zclamp);
+		sm.AddMacro("PS_SCANMSK", sel.scanmsk);
 
 		wil::com_ptr_nothrow<ID3D11PixelShader> ps;
 		CreateShader(m_tfx_source, "tfx.fx", nullptr, "ps_main", sm.GetPtr(), ps.put());
--- a/pcsx2/GS/Renderers/HW/GSRendererNew.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererNew.cpp
@@ -1177,6 +1177,7 @@ void GSRendererNew::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 
 	ResetStates();
 	m_conf.cb_vs.texture_offset = GSVector2(0, 0);
+	m_conf.ps.scanmsk = m_env.SCANMSK.MSK;
 
 	ASSERT(m_dev != NULL);
 
@@ -1303,6 +1304,9 @@ void GSRendererNew::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 	{
 		m_conf.blend = {}; // No blending please
 	}
+
+	if (m_conf.ps.scanmsk & 2)
+		DATE_GL42 = false; // to have discard in the shader work correctly
 
 	if (m_conf.ps.dfmt == 1)
 	{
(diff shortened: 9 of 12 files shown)


plugins/GSdx/res/glsl/tfx_fs.glsl
```diff
@@ -860,12 +860,28 @@ void ps_main()
...
+    alpha_blend /= 2.0f;
+#endif
     SV_Target0 = C / 255.0f;
     SV_Target1 = vec4(alpha_blend);
 
 #if PS_ZCLAMP
 	gl_FragDepth = min(gl_FragCoord.z, MaxDepthPS);
-#endif 
+#else
+ 	gl_FragDepth = gl_FragCoord.z;
```
tadanokojin, line 872: This is a terrible idea. I understand you did this because you wanted it to make the thing fail the depth test to discard it (and if you write depth in a branch you have to always write it), but even a normal discard probably would have been better. This code makes everything slower to the benefit of a single game.

plugins/GSdx/res/glsl/tfx_fs.glsl
```diff
@@ -860,12 +860,28 @@ void ps_main()
         // C.rgb = (C.rgb - 256.0f);
     // }
 // #endif
+#if PS_SCANMSK_TRANSPARENT
+    alpha_blend /= 2.0f;
+#endif
```
tadanokojin, line 865: This is just a hack.

The reason MGS doesn't work if you implement scanmsk is nothing to do with the blending equation. It has to do with our PCRTC implementation. You've simply bypassed actually addressing the problem and ignored the functionality of scanmsk to instead render at half transparency. This isn't a fix.

plugins/GSdx/res/glsl/tfx_fs.glsl
Sergeanur: Ah, I see, didn't know about discard. Changed that now.

plugins/GSdx/res/glsl/tfx_fs.glsl
Sergeanur: Naturally Transparent mode was only an option to escape the interlaced look given by scaled and unscaled, but you're saying there's more to it with PCRTC? I was under impression that illusion of transparency was given by the TV's of the era but there's also something else that PS2 does that could be fixed too?

tadanokojin: Okay, at least I can see where you're coming from then.

No, PCRTC has two displays and MGS sticks the framebuffer in both of them and offsets one of them before compositing them together. PCSX2 just removes this offsetting as a part of another hack that removes blur.

Sergeanur: I've tackled GSRenderer::Merge a bit so that the offsetting would depend on a mode. Scaled looks pretty neat now with no noticeable blur. Please check if it's doable.

nobleflame:
Hi there, where is this option located in the latest dev build?

F0bes:
It's not merged, you need to build this pr.

nobleflame:
Thanks, I'm not sure how to do that, so I'll leave it until this is fully integrated. Strange because Sergeanur commented:
"Merged with current master and made 'Scaled' as default."

F0bes:
Oh yeah I guess that could've been confusing.

Click on the `Checks` tab and click on windows builds.
Once they finish you'll be able to download a zip with this prs pcsx2

Regarding merging with master, they pulled in master (merged) and updated their pr to eliminate conflicts. If there are conflicts we can't merge it. Another word used for this is rebase.

nobleflame:
Thanks very much. You've been really helpful!

nobleflame:
Hey, so I’ve got it working. Thanks for putting this together - it works like a charm and now all transparent surfaces are displaying correctly.

Do you have any idea of this will be included in the next official release of PCSX2?

Thanks :)

RedDevilus:
When it's reviewed to be not hacky, enough perfomant, doesn't break that or other games. So no ETA and i'm not capable or knowledgeable enough for the GS side.

It needs to replace any gsdx mention to GS too.

refractionpcsx2:
> Hey, so I’ve got it working. Thanks for putting this together - it works like a charm and now all transparent surfaces are displaying correctly.
> 
> Do you have any idea of this will be included in the next official release of PCSX2?
> 
> Thanks :)

Not in its current state, I don't like it having options in a GUI and a "transparent mode" which is really just a hack I don't like, also I don't like the fact it's forcefully disabling the Z writes, when it should be still doing it for the pixels written.

arcum42:
Seems like the fact that all the dialog box changes were to a version of the dialog box that no longer exists would be a bit of an issue too. :P

nobleflame:
All fair points. Still, hack or not, it does work with MGS2/3. Perhaps it could be implemented in a different place in the menus and once it’s been tested of course.

Thanks to all who work on this stuff. I have absolutely no coding knowledge whatsoever, so your hard work is enabling people like me to enjoy their PS2 libraries (which were, up until recently, gathering dust).

refractionpcsx2:
Sure, it works for MGS, but it might not work for anything else. I'd rather it was done correctly, if you don't mind.

Sergeanur:
> > Hey, so I’ve got it working. Thanks for putting this together - it works like a charm and now all transparent surfaces are displaying correctly.
> > Do you have any idea of this will be included in the next official release of PCSX2?
> > Thanks :)
> 
> Not in its current state, I don't like it having options in a GUI and a "transparent mode" which is really just a hack I don't like, also I don't like the fact it's forcefully disabling the Z writes, when it should be still doing it for the pixels written. However if the scanmask should be scaled or not, that's another question, I'm not sure which would be most correct. From the emulation point of view, an unscaled version would be more accurate.

I don't think there's an ultimate win without having a GUI option. I don't know if there are games that use scanmask other than MGS but I assume there are, and I assume they may use it for different purposes other than "fake transparency". Unscaled option is most faithful to PS2 but it blurs the image on any resolution, so it's most recommended only if scanmask is relying on native resolution. With scaled you get less blur as you crank the resolution up, so if the game doesn't care about being exact to native resolution, it's a doable option. The problem with both of these modes if that you get "interlaced in motion" geometry, most noticeable in MGS2 when you look on the water in first person mode and move the camera around. You can get rid of it in scaled mode if you increase internal resolution higher than your monitor's but that of course comes with a cost. Because of that I made transparency mode that renders full geometry half-transparent, it eliminates the "interlaced in motion" look without need change the resolution. Z-write is disabled for that mode because in case of MGS2 there were bubbles drawn after water render and because of that Z-test would fail and they end up being not drawn at all. It makes sense that if the developers tried to make transparent objects with scanmask they wouldn't care about rendering order, so I only recommend this if scanmask is used to fake transparent geometry. It's a bit hacky, yeah, but I've played around with it and haven't seen any issues.

JordanTheToaster:
Fixes transparency on GT4 ghost car in time trial.

image

nobleflame:
@JordanTheToaster 

that’s interesting. I played through the whole of MGS2 and 3 and the hack does indeed work.

I’ve since switched over to the latest dev build for stability and since new builds come out all of the time, but I’d love it for something like this hack to be added to a future release. 

Hoping that someone can fix the blending accuracy issues with MGS2 as well - the skin tones are shown as much darker on some models. setting blending accuracy to high fixes the issue, but causes massive slow down.

refractionpcsx2:
> but I’d love it for something like this hack to be added to a future release.

Well kinda what I was saying is this doesn't need to be a hack if done properly. it is entirely plausible for this to work properly without hacks.   In fact a hack will need to be removed (though preferably made an option) to make this work.

Sergeanur:
For clarity, only 'Transparent' mode is a hack, the rest are not.

refractionpcsx2:
yes the whole transparent mask thing is a hack and that's not getting in to the emulator :)

but if you can make it work properly and let the merge circuit do the blending, that'd be great.

Sergeanur:
Ok, redone now with plain unscaled, no menu, no tranparency, rebased onto latest master too.

pcsx2/GS/Renderers/Common/GSRenderer.cpp
```diff
@@ -220,7 +220,7 @@ bool GSRenderer::Merge(int field)
 			off.x = tex[i]->GetScale().x * frame_diff.x;
 		}
 
-		if (display_diff.y >= 4) // Shouldn't this be >= 2?
+		if (display_diff.y != 0) // Shouldn't this be >= 2?
```
refractionpcsx2, line 223: we want to leave this as >= 4 when scanmsk isn't on else a ton of people are gonna freak out about their games being blurry. so maybe have a parameter which allows the >= 4 when scanmsk hasn't been set to 2 or 3, but if it does get set, switch the boolean (so it forces the merge circuit to act on any offset) and it only reverts on GSReset

pcsx2/GS/Renderers/SW/GSRasterizer.cpp
```diff
@@ -1131,7 +1133,9 @@ void GSRasterizer::Flush(const GSVertexSW* vertex, const u32* index, const GSVer
 				int left = e->_pad.I32[1];
 				int top = e->_pad.I32[2];
 
-				DrawScanline(pixels, left, top, *e++);
+				if ((m_scanmsk_value & 2) == 0 || (m_scanmsk_value & 1) == (top & 1))
```
refractionpcsx2, line 1136: Isn't it the opposite? When scanmsk is 2 (so bit 0 will be zero) drawing of even lines is prohibited.  so you want !=

refractionpcsx2:
Further to my comments, we discussed it on the discord and you might need to disable GL42 for the discard to work, else the depth value could end up incorrect due to depth write happening in early fragment testing.

Another thing on the shader, it looks like you're scaling the scanmsk, but I don't think the merge circuit does, so I don't think you need the scale factor on there.  Of course it'll be worth testing to see what works for MGS and the other games.

refractionpcsx2:
Sorry for ping @Sergeanur just updated my comment with more information, which I think you had done some of previously, realised you might have already read what I said and gone off trying to change it. If I confused you I apologise.

Sergeanur:
I implemented the changes you asked I think, please check.

refractionpcsx2:
Hopefully that will do the job, yea :)

You might be able to get away with setting m_scanmask_used in GSState where SCANMSK is set instead of having to do it in the renderer, it possibly doesn't matter too much, but it's nice to have it centralised.

More importantly, have you managed to test it on your end to see if it works?

refractionpcsx2:
image

Software mode looks to be working great there, however neither DX or OGL Hardware work, even in native, it looks like it's not discarding every other line.

nobleflame:
Just to say that I thank you all for your hard work. 

You’re swiftly making PCSX2 the best way to enjoy MGS2 and 3

refractionpcsx2:
Okay update now I've stopped being an idiot. Hardware mode working great, even with upscaling :)

image

pcsx2/GS/Renderers/Common/GSRenderer.cpp
```diff
@@ -220,7 +221,9 @@ bool GSRenderer::Merge(int field)
 			off.x = tex[i]->GetScale().x * frame_diff.x;
 		}
 
-		if (display_diff.y >= 4) // Shouldn't this be >= 2?
+		if (m_scanmask_used && display_diff.y == 1) // Scanmask effect wouldn't look correctly if we scale the offset
```
refractionpcsx2, line 224: correct, not correctly (I'm guessing you rephrased it at some point)

pcsx2/GS/Renderers/Common/GSRenderer.cpp
```diff
@@ -384,6 +387,7 @@ void GSRenderer::VSync(int field)
 	{
 		ResetDevice();
 	}
+	m_scanmask_used = false;
```
refractionpcsx2, line 390: Would be better to reset this in GSState::Reset() so m_scanmask_used stays on for the entire game until a reset occurs.  then set it to true in GSState::GIFRegHandlerSCANMSK if 2 is set

TransparentBlue:
>Somebody else will need to test MGS2 and GT4 (ghost cars?), I'm not sure I can check that.

Does still fix transparency for ghost cars in Gran Turismo 3 and 4.
Does still fix Otacon's glasses and the water at the beginning of the second part of Metal Gear Solid 2
 I can make a GSdumps for those if that would help

Sergeanur:
I found an issue with OpenGL renderer. When looking at the water in first person view with an even resolution scale (2x, 4x, 6x, etc.) you get the water blinking in the eyes area, juggling between fully opaque water and non-existant water. This doesn't happen on DirectX. I'm guessing the way OpenGL resizes the texture is by throwing away every odd line.
изображение

Sergeanur:
Also pushed all the suggested changes

refractionpcsx2:
This is an issue that iMine ran in to as well, it's to do with linear interpolation (i think) when doing stretchrect, it was causing problems with the burnout sky issue as well, probably not a lot we can do about it at the moment.

prafullpcsx2:
image

GT4 Works fine

gilderoylockhart01:
MGS 3 is also fine (just needs edge AA to be perfect, but outside of the scope of this PR), I feel like I will have to send a new video to Bositman ^^:

https://user-images.githubusercontent.com/90222976/146792310-e135f85c-9d58-4acf-b80d-ff4310c0b536.mp4

refractionpcsx2:
How does the MGS2 water look? 

Does anybody have Raw Danger? Apparently that makes use of Scanmsk too as noted by kojin here https://github.com/PCSX2/pcsx2/issues/2753#issuecomment-849954749

Sergeanur:
Water, Fatman's glass of wine and Harrier's glass in MGS2 look fine.

refractionpcsx2:
ace :)

RedDevilus:
Refraction made an addition to detect scanmask in the Program Log and I found out that all the network config screens use scanmsk, however on this PR it will automatically choose Bob Interlacing instead of Blending Interlacing

image
vs
image

lightningterror (commented):
Can this be rebased as some changes were made which caused conflicts?

gilderoylockhart01:
Is it possible to have this build where it logs scanmask effects?

Sergeanur:
Rebased

Sergeanur:
> Refraction made an addition to detect scanmask in the Program Log and I found out that all the network config screens use scanmsk, however on this PR it will automatically choose Bob Interlacing instead of Blending Interlacing. Keep in mind this is only for software mode.
> 
> image vs image
> 
> Video showcase:
> BobInterlacingNetworkScreen.mp4

Looks like an issue of a merge or the SW renderer. In this case scanmask is used with offsetY=0 and should fill the gaps of one frame with another. On HW renderer it makes the image a bit darker though. Not sure if real PS2 would do the same, can't check too.

refractionpcsx2:
What I think is happening, is for some reason they chose to only render the lines which would be rendered out to the TV on the next VSync, however because in some situations we can cheat and use "bob" method (this is kind of a fake progressive) which skips all the interlacing and we just display the whole buffer, it causes this effect.  

So what needs to be done is if scanmsk is detected and we're using Bob, it needs to change to Blend (preferably) or Weave.

Sergeanur:
Nah, it was DrawSprite that would draw the whole rect instead of scanmasked one, patched it now.

RedDevilus:
> Is it possible to have this build where it logs scanmask effects?
[pcsx2-scanmsk_with_log.zip](https://github.com/PCSX2/pcsx2/files/7763802/pcsx2-scanmsk_with_log.zip)

I tried many games, and alas didn't really find any new ones. Like I though Casper would be one as scanmask seems to layer with transparency but like with PS2 nothing is standard and seems to be quite rare.

RedDevilus:
Master:

https://user-images.githubusercontent.com/24227051/147126347-489b5487-e920-46c2-b98b-6f2866479449.mp4


PR:

https://user-images.githubusercontent.com/24227051/147126367-f6c4e96a-02cd-4841-a232-0a271b8e511f.mp4



You can see bob effect on buttons, not sure if intended. Atleast it's not the whole screen flickering now.

refractionpcsx2:
That just looks like interlacing is disabled, which is probably fine for Bob

RedDevilus:
Screenshots if you can't see it clearly, when you move buttons:
image
image

When idle, looks the same as master:
image

Sergeanur:
This is expected due to nature of the effect. Could try enabling Blend if interlacing is set to automatic and scanmask was used while both frames have the same offsetY.

gilderoylockhart01:
PS1DRV uses scanmask, what a surprise.

Alex10005:
How close thing merge into master build of pcsx2.

gilderoylockhart01:
So what is the status of this PR, is the implementation of scanmask proper now? And what about the "antiblurr" hack?

nobleflame:
> So what is the status of this PR, is the implementation of scanmask proper now? And what about the "antiblurr" hack?

Patience. Devs are testing it and it will probably merge when they're confident it's been properly implemented.

refractionpcsx2:
I think the PR is good to go, I'm just wanting a code review from @lightningterror and @tellowkrinkle since they are the main GS guys, just to make sure they're happy with the placement of things, mainly the software renderer changes, I don't know if there's a better way to do those (don't know enough about it personally)

lightningterror (approved):

tellowkrinkle (approved):

refractionpcsx2:
We might modify this later but this is fine for now :)

PR #4393, opened 2021-04-24
Input Recording - Allow configuration of the frame advance amount

xTVaser:
### Description of Changes

Adds a GUI configurable setting to control how many frames the frame-advance feature should advance.

imageimage

### Rationale behind Changes

It's not unreasonable to imagine a use-case where you may want to advance frames in multiples other than `1s`.  So this gives that ability to the user.

In the future, it would probably be nice to have 2 key-bindings, one that always advances by `1s` and another that uses a custom/configurable value.  Wouldn't be too difficult to add but I think I'm tempted to wait until PCSX2 in general has better key-binding configuration for features like that.

### Suggested Testing Steps

Easier to confirm the right amount of frames are being advanced if you are recording/playing back a recording.  Change the value, restart the emulator (should persist), etc.

### DIFF
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -22,6 +22,7 @@
 #include "Dialogs/ModalPopups.h"
 #include "IsoDropTarget.h"
 
+#include "fmt/core.h"
 #include <wx/iconbndl.h>
 
 #include <unordered_map>
@@ -29,12 +30,12 @@
 
 #include "svnrev.h"
 #include "Saveslots.h"
+
 #ifndef DISABLE_RECORDING
 #include "Recording/InputRecording.h"
+#include "Recording/InputRecordingControls.h"
 #endif
 
-
-#include "fmt/core.h"
 // ------------------------------------------------------------------------
 wxMenu* MainEmuFrame::MakeStatesSubMenu(int baseid, int loadBackupId) const
 {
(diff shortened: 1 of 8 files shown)


pcsx2/gui/MainFrame.cpp
```diff
@@ -533,17 +535,25 @@ void MainEmuFrame::CreateCaptureMenu()
...
-void MainEmuFrame::CreateRecordMenu()
+void MainEmuFrame::CreateInputRecordingMenu()
 {
 #ifndef DISABLE_RECORDING
 	m_menuRecording.Append(MenuId_Recording_New, _("New"), _("Create a new input recording."))->Enable(false);
 	m_menuRecording.Append(MenuId_Recording_Stop, _("Stop"), _("Stop the active input recording."))->Enable(false);
 	m_menuRecording.Append(MenuId_Recording_Play, _("Play"), _("Playback an existing input recording."))->Enable(false);
 	m_menuRecording.AppendSeparator();
+
+	m_menuRecording.Append(MenuId_Recording_Settings, _("Settings"), &m_submenu_recording_settings);
```
lightningterror, line 546: I don't think a new sub menu is needed for just 1 option.

xTVaser: This is true, however I have more planned settings to go into this sub-menu, so I went in this direction to get things ready for that.

lightningterror: Makes sense, it's fine then.

PR #4364, opened 2021-04-12
DEV9: add ATA Idle Immediate (E1) stub

macmenot:
Fixes https://github.com/PCSX2/pcsx2/issues/4358

PR #4403, opened 2021-04-27
DEV9: consistently prefix all console output with `DEV9: `

macmenot:
Around half of the debug and console output from the DEV9 tree already
had this prefix. Adding it everywhere for consistency. Also fixed
misspelling of "Unknown" whilst there.

This is a follow-on from https://github.com/PCSX2/pcsx2/pull/4364 as requested by @lightningterror


GovanifY (approved):
Big patch! Thanks for the change, LGTM :)

Issue #4120, opened 2021-01-08
[Regression] Defocused window on boot - Hotkeys don't respond anymore


RedDevilus:
**Describe the bug**
While testing i like to use F4 , tab and other hotkeys. I have a laptop screen with is lid closed and an external monitor which is my actual main screen. But i need to always click on the screen or alt-enter to make my hotkeys work on the emulation. (It acts like 1.6 correctly on rare occasions)

**To Reproduce**
Get new devbuild
Unzip and go through first-time-wizard
Change no settings
Start a random PS2 game, full or fast boot
You boot in windowed mode
Try any hotkey (F4, Tab , etc)

**Expected behavior**
In 1.6 it will make the emulation correctly an active window, no matter if its fullscreen or windowed mode.
![image](https://user-images.githubusercontent.com/24227051/104073481-10cfb680-520e-11eb-8773-42606f7acbc8.png)

**GS Window Screenshots**
![image](https://user-images.githubusercontent.com/24227051/104073304-a1f25d80-520d-11eb-892e-70c677e5de85.png)
Using this option on the devbuild seems to fix this issue on first-booting on devbuild in windowed mode.

**System Info (please complete the following information):**
 - PCSX2 Revision:  dev-844
 - OS: W10 Pro 20H2
 - CPU: i7-9750H
 - GPU: 2060M 6GB

lightningterror:
Regression build needs to be identified.

lightningterror:
Status?

RedDevilus:
Triggers even on 1.6, sometimes it fixes by clicking outside the emulator and back on emulation window. Other times it works out of the box and rest i need to alt+enter to make it work.

RedDevilus:
Fixed on PR 4404.

PR #4404, opened 2021-04-27
GSFrame: Pass focus to GSPanel once Frame receives focus

CookiePLMonster:
Passing focus on the focus message instead of the activate message resolves an issue where external processes (e.g. NVIDIA overlay) forced a focus change to the GS Frame, requiring the window to be refocused in order for keyboard inputs to work again.

Currently, GSFrame rejects focus messages and instead passes it to the child GSPanel.

To better illustrate the changes, let's compare the GSFrame message flow before and after this change:

# Before

## WM_SETFOCUS
* Do nothing special. GSFrame receives focus and accepts it.
* Continue with default window procedure.

## WM_ACTIVATE
* Set focus to GSPanel. This prevents `WM_ACTIVATE` from setting focus to GSFrame (as it is contractually done by the default window handler).
* Continue with default window procedure.

# After

## WM_SETFOCUS
* Set focus to GSPanel. This prevents GSFrame from getting focused at.
* Set the message as processed, **ignoring the default window procedure**. This essentially "rejects" the focus message.

## WM_ACTIVATE
* Do nothing special. GSFrame receives focus, but because of a new `WM_SETFOCUS` handler, it is going to reject it and pass it to GSPanel instead.
* Continue with default window procedure.

In the case of NVIDIA overlay (and possibly more), a focus message was sent to the GSFrame **without** `WM_ACTIVATE`, so the existing logic failed to pass focus to the GSPanel. I have not managed to locate what exactly sends this message, but I confirmed it does not come from the PCSX2 process, so it must be posted from elsewhere.

### DIFF
--- a/pcsx2/gui/FrameForGS.cpp
+++ b/pcsx2/gui/FrameForGS.cpp
@@ -565,7 +565,7 @@ GSFrame::GSFrame( const wxString& title)
 	Bind(wxEVT_CLOSE_WINDOW, &GSFrame::OnCloseWindow, this);
 	Bind(wxEVT_MOVE, &GSFrame::OnMove, this);
 	Bind(wxEVT_SIZE, &GSFrame::OnResize, this);
-	Bind(wxEVT_ACTIVATE, &GSFrame::OnActivate, this);
+	Bind(wxEVT_SET_FOCUS, &GSFrame::OnFocus, this);
 
 	Bind(wxEVT_TIMER, &GSFrame::OnUpdateTitle, this, m_timer_UpdateTitle.GetId());
 }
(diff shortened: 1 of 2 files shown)


RedDevilus:
Probably fixes: https://github.com/PCSX2/pcsx2/issues/4120

RedDevilus (approved):

MrCK1:
This will need to be retested after GS merge

CookiePLMonster:
Rebased on latest master and confirmed that:
1. The bug is still here without the PR.
2. This PR still fixes it.

Basically GS merge didn't affect this issue at all.

RedDevilus (approved):

PR #4406, opened 2021-04-28
docs: add a compiling guide for linux

GovanifY:
This is a first draft of our linux compilation guide.

Will need to be proofread and we might also want some small guide for windows, for whoever wants to explain
how to click buttons.


kenshen112:
https://wiki.pcsx2.net/PCSX2_Documentation/Compiling_on_Linux

https://github.com/PCSX2/pcsx2/wiki/Installing-on-Linux

GovanifY:
@kenshen112 those guides are outdated, not maintained by us, and for linux distribution specific. If you want to have even more fun, try to run the fedora installation guide, you'll see it won't work.

kenshen112:
> @kenshen112 those guides are outdated, not maintained by us, and for linux distribution specific. If you want to have even more fun, try to run the fedora installation guide, you'll see it won't work.

Been there done that with fedora. And while it's true those guides are out of date. If we're going to draft a new one up we should either close the others as they're hosted on our end or publish the updated info to them.

GovanifY:
> Been there done that with fedora. And while it's true those guides are out of date. If we're going to draft a new one up we should > either close the others

Probably a good idea once all of this is said and done, yup

kenshen112:
Also speaking of windows since you'd mentioned it above: https://wiki.pcsx2.net/PCSX2_Documentation/Compiling_on_Windows again is out of date but could be fixed

GovanifY:
@kenshen112 Yeah, we're probably going to do this in this PR, I just want to get some light review first so everyone agrees

qurious-pixel:
It would be nice to have more detailed summary of the `CMAKE options`, and remove the ones that no longer work. 

At least a hyperlink to `CMAKE expert mode`.

GovanifY:
> It would be nice to have more detailed summary of the `CMAKE options`, and remove the ones that no longer work.
> 
> At least a hyperlink to `CMAKE expert mode`.

I'm fine with an hyperlink, I'll probably add one to git's manual too since both are out of scope.
I don't think we have much "deprecated" flags regarding our cmake though, do you have any example of this?

kenshen112:
https://github.com/PCSX2/pcsx2/pull/4373 I'd refer to Tellow's cleanup PR he removed some of them IIRC

GovanifY:
@kenshen112 That's not even merged... 
Forgetting that the flags it deprecates are not cmake flags but project specific flags, which we removed _because_ native cmake flags can do the same.

tellowkrinkle:
> Forgetting that the flags it deprecates are not cmake flags but project specific flags, which we removed _because_ native cmake flags can do the same.

I did remove a few project specific flags that aren't also native flags but they seem like the kind of things we would have left out of a compiling guide

image

Side note, `ccmake .` is a great way to see the flags available to you

tellowkrinkle:
I think you should leave the `$` out of the terminal prompts

People end up pasting them into the terminal and then wondering why the command doesn't work.  They're only really helpful if you want to show both input and output anyway, which we don't here

README.md
```diff
@@ -18,6 +18,8 @@ Installers and binaries for both Windows and Linux are available from [our websi
 
 Development builds are also available from [our website](https://pcsx2.net/download/development.html).
 
+For compilation instructions please follow our [compilation guide](COMPILING.md).
```
RedDevilus, line 21: Might be nice to mention that it's for Linux distros in general. What if a windows user sees this line?

GovanifY: this is why i said we will probably add windows too in this PR, so both guides are available at the same place.

qurious-pixel:
No longer working flags:
`-D64BIT_BUILD_DONT_WORK`
`-DGTK3_API`

changed flags:
`-DPLUGIN_DIR` --> `-DCMAKE_INSTALL_LIBDIR`
`-DGAMEINDEX_DIR` --> `-DCMAKE_INSTALL_DATADIR`

Also, the development builds appear to be Windows only.
https://pcsx2.net/download/development.html
And Linux build was last compiled 5 years ago. 
https://pcsx2.net/download.html

GovanifY:
> Also, the development builds appear to be Windows only.
> https://pcsx2.net/download/development.html
> And Linux build was last compiled 5 years ago.
> https://pcsx2.net/download.html

Yup, we are planning to move out to use AppImage for linux, using your PR! We will probably wait for the GS merge first though, which should begin by the end of the week


> No longer working flags:
> `-D64BIT_BUILD_DONT_WORK`

To be fair, 64 bit builds do work so yeah

> `-DGTK3_API`

Doesn't wxWidgets build with GTK3 by default now?

> changed flags:
> `-DPLUGIN_DIR` --> `-DCMAKE_INSTALL_LIBDIR`
> `-DGAMEINDEX_DIR` --> `-DCMAKE_INSTALL_DATADIR`

No plugins anymore, so that makes sense too, and we moved to use native cmake flags, unless I'm mistaken.
AFAICT our current documentation thus should only need to concern itself with native cmake flags.

xTVaser:
As you moved the CONTRIBUTING file, can you update the following link:
https://github.com/PCSX2/pcsx2/blob/5639273cec2111afa95772fb2e4e72311c00fbb9/.github/workflows/pr-triage.yml#L24

GovanifY:
@xTVaser thanks for the heads up, fixed!

For the rest this guide needs a heads up explaining nixGL so it can run and boot using the system opengl libraries, which I'll do whenever i can!

PR #4410, opened 2021-04-30
UI: Asynchronous BIOS enumeration

F0bes:
### Description of Changes
Instead of blocking the UI, show the text "Enumerating..." and run a thread to enumerate through the BIOSes

### Rationale behind Changes
It can look like PCSX2 has crashed when verifying BIOSes in a directory.
Expensive file I/O shouldn't be on the UI thread

### Suggested Testing Steps
Duplicate your BIOS bin files and put them in a single directory.
Go to the BIOS selection page and set the BIOS directory to said directory.
You should see "Enumerating BIOSes..." as the first item in the BIOS list, the UI shouldn't freeze
Once the thread is done working, the list box should contain your BIOSes

### DIFF
--- a/pcsx2/gui/Panels/BiosSelectorPanel.cpp
+++ b/pcsx2/gui/Panels/BiosSelectorPanel.cpp
@@ -44,8 +44,8 @@ wxDEFINE_EVENT(pxEvt_BiosEnumerationFinished, wxCommandEvent);
 // upgraded to wx2.9/3.0, it should incorporate such functionality into this base class.  (for now
 // we just provide the user with a "refresh" button).
 //
-Panels::BaseSelectorPanel::BaseSelectorPanel( wxWindow* parent )
-	: BaseApplicableConfigPanel( parent, wxVERTICAL )
+Panels::BaseSelectorPanel::BaseSelectorPanel(wxWindow* parent)
+	: BaseApplicableConfigPanel(parent, wxVERTICAL)
 {
 	Bind(wxEVT_DIRPICKER_CHANGED, &BaseSelectorPanel::OnFolderChanged, this);
 	Bind(wxEVT_SHOW, &BaseSelectorPanel::OnShow, this);
@@ -54,22 +54,22 @@ Panels::BaseSelectorPanel::BaseSelectorPanel( wxWindow* parent )
 void Panels::BaseSelectorPanel::OnShow(wxShowEvent& evt)
 {
 	evt.Skip();
-	if( evt.IsShown() )
+	if (evt.IsShown())
 		OnShown();
 }
 
 void Panels::BaseSelectorPanel::OnShown()
 {
-	if( !ValidateEnumerationStatus() )
+	if (!ValidateEnumerationStatus())
 		DoRefresh();
 }
 
-bool Panels::BaseSelectorPanel::Show( bool visible )
+bool Panels::BaseSelectorPanel::Show(bool visible)
 {
-	if( visible )
+	if (visible)
 		OnShown();
 
-	return BaseApplicableConfigPanel::Show( visible );
+	return BaseApplicableConfigPanel::Show(visible);
 }
 
 void Panels::BaseSelectorPanel::RefreshSelections()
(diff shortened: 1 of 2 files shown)


F0bes:
Ok, I moved things to a separate thread.
I think I fixed the race conditions? My implementation probably sucks and I should be ashamed.

RedDevilus (approved):
It's so fast that i can barely read it :P

Atleast it works.

pcsx2/gui/Panels/BiosSelectorPanel.cpp
```diff
@@ -162,28 +167,48 @@ bool Panels::BiosSelectorPanel::ValidateEnumerationStatus()
...
+	auto enumThread = static_cast<EnumThread*>(evt.GetClientData());
+	// Sanity check, in case m_BiosList was updated by ValidateEnumerationStatus() while the EnumThread was running
+	if (m_EnumeratorThread.get() != enumThread || m_BiosList->size() < enumThread->Result.size()) return;
+
+	const wxFileName& currentBios = g_Conf->FullpathToBios();
+	m_ComboBox->Clear(); // Clear the "Enumerating BIOSes..."
 
-		if (wxFileName((*m_BiosList)[i] ) == right)
-		{
+	std::for_each(enumThread->Result.begin(), enumThread->Result.end(), [this, currentBios](std::pair<wxString, u32> result) {
```
tellowkrinkle, line 206:
```suggestion
	for (const std::pair<wxString, u32>& result : enumThread->Result)
	{
```
I know everyone loves their higher order functions, but C++ does a really bad job of making them look nice

pcsx2/gui/Panels/BiosSelectorPanel.cpp
```diff
@@ -162,28 +167,49 @@ bool Panels::BiosSelectorPanel::ValidateEnumerationStatus()
...
-			biosSet = true;
-		}
 	}
+};
 
-	if ((!biosSet) && !(m_ComboBox->IsEmpty()))
-		m_ComboBox->SetSelection(0);
-}
+Panels::BiosSelectorPanel::EnumThread::EnumThread(BiosSelectorPanel &parent) : m_parent(parent)
+{};
```
GovanifY, line 215: Is this how our clang-format does it? that seems weird to me, might be because the function name is overly long though

pcsx2/gui/Panels/ConfigurationPanels.h
```diff
@@ -461,11 +461,34 @@ namespace Panels
...
+
 	protected:
 		virtual void Apply();
 		virtual void AppStatusEvent_OnSettingsApplied();
 		virtual void DoRefresh();
+		virtual void OnEnumComplete(wxCommandEvent& evt);
 		virtual bool ValidateEnumerationStatus();
+		
+		std::unique_ptr<EnumThread> m_EnumeratorThread;
+		Threading::Mutex m_EnumeratorMutex;
```
GovanifY, line 491: unused variable

pcsx2/gui/Panels/BiosSelectorPanel.cpp
```diff
@@ -115,6 +118,8 @@ Panels::BiosSelectorPanel::BiosSelectorPanel( wxWindow* parent )
 	*this	+= 8;
 	*this	+= m_FolderPicker	| StdExpand();
 
+	Bind(pxEvt_BiosEnumerationFinished, &BiosSelectorPanel::OnEnumComplete, this);
+
```
GovanifY, line 122: Why the linefeed?

GovanifY (changes requested):
A few nitpicks, otherwise LGTM aalthough I haven't tested not looked too in depth about the change, the new thread seems to be implemented like it should be though

pcsx2/gui/Panels/BiosSelectorPanel.cpp
F0bes: Usually, there is a linefeed in between button click binds and all other binds. There are only two here so I could remove it if you'd like.

RedDevilus:
-ConfigurstionPanels.h  typo

pcsx2/gui/Panels/BiosSelectorPanel.cpp
GovanifY: Eh, I don't have any strong thoughts on that one, do as you prefer

GovanifY (approved):
Needs testing but otherwise LGTM

F0bes:
I tested on windows after I fixed conflicts with master, still works.

PR #4411, opened 2021-04-30
license: clarify we are licensing under LGPLv3+

GovanifY:
Since 2012 PCSX2 shipped with 4 COPYING files, indicating that it might have been licensed as (L)GPLv2+.
That is incorrect as this same commit introduced the new file header format, which is what we use on essentially all our files, saying very clearly that this file is licensed under the terms of the LGPLv3, or any later revisions.

https://github.com/PCSX2/pcsx2/commit/ea98450eb0ed4c1e7eda76582ffbedefa8ae76af#diff-1a10173fe2b8edb74ff22e2da529ec5fa18a4f4d0b137977158a5baf13b7e321R3

As such I removed LGPLv2 and GPLv2 licenses files from our project root.

I should note that the LGPL COPYING file only contains the extensions to the original GPL, as such we still need to ship the GPL.


lightningterror:
Just so we don't forget, merging this one after gsdx merge.

GovanifY:
Added it on top of my gs-merge branch, closing.

PR #4416, opened 2021-05-03
CDVD-GUI: Swap buttons `Browse...` and `Ask when booting` and default to unchecked state.

RedDevilus:
### Description of Changes

Swap the GUI options Ask when booting and Browse... in CDVD. Ask when booting is also no longer on by default.

Old:
image

New:
image

### Rationale behind Changes

Makes more sense for muscle memory. It does seem cleaner as well, though perhaps biased.
### Suggested Testing Steps

Test some games and swap between them. Disabling checkbox at another test to see if the GUI updates properly.


RedDevilus:
One annoying thing that happens in master:

image

It doesn't grayout browse, ask on boot is checked and it selected my previous game, when you boot it will go into browsing mode instead of resuming the selected game. Perhaps it's better to have ask when browsing be off ( first time it will always go into browsing mode no matter if ask on boot is checked so you don't lose that function)

PR #4418, opened 2021-05-05
CI: Validate GameDB changes in GitHub Actions

xTVaser:
### Description of Changes

Currently, the emulator will fully validate the GameDB, which is how we've been catching these recent mistakes.  But this adds a CI check to validate the file in a very similar manner at build time.

### Rationale behind Changes

Only validating via the emulator application means that merging with confidence requires that extra step.  It would be better in the long-term to remove that extra step, and do it in CI.

The only issue with this, is this means we now have validation code in two places that has to be kept up to date.  Though with how infrequently the GameDB schema changes, likely not that big of a deal.  If it becomes one, can be re-evaluated.

### Suggested Testing Steps

Here's some test commits / builds:
- https://github.com/xTVaser/pcsx2-rr/runs/2505386598
- https://github.com/xTVaser/pcsx2-rr/runs/2505391776
- https://github.com/xTVaser/pcsx2-rr/runs/2505395100


RedDevilus:
Is it also possible to let it look for missing punctuation or other minor nitpicky stuff for low warning level?

xTVaser:
> Is it also possible to let it look for missing punctuation or other minor nitpicky stuff for low warning level?

It's a generic python script, so nothing technically preventing it, feel free to extend the script.  But I would suggest the initial merge be limited to checking for things that will cause fatal errors.

GovanifY (approved):
LGTM! I think we honestly could have lived with a dead simple YAML parser but I'm not against having a full on linter, thanks for going the extra mile :)

PR #4421, opened 2021-05-07
Convert DEV9 config dialog to wxWidgets

tellowkrinkle:
### Description of Changes
Switches from separate config dialogs for windows and linux/macOS to one shared wxWidgets-based dialog

### Rationale behind Changes
Less split code and helps remove GTK dependency on macOS

### Suggested Testing Steps
Open DEV9 config menu and see if it still works


TheLastRar:
As a suggestion, can you have provide a dropdown for the API, with a second dropdown that is populated with the relevant adapters.

@RedDevilus made a couple suggestion about regarding making the newly added DHCP panel a bit easier to use;
The ability so export/import the DHCP settings
Or a dropdown for some of the settings (DNS servers (8.8.8.8 or common PS2 DNS), or template IP address (192.168.0.0, 10.0.0.0, etc)

Regarding the later point, It may be possible to auto-detect that based on the currently selected adapter.

The profiles and such could probably be done at a later point in time, but I'm just mentioning here for now.

tellowkrinkle:
It now complains if you try to save invalid settings

PR #4422, opened 2021-05-07
Update ghc::filesystem to 1.5.4

tellowkrinkle:
Updates `ghc::filesystem` dependency to 1.5.4

Makes the wchar-based implementation default on Windows, and makes it possible to initialize a `fs::path` from a `wchar*`

### DIFF
--- a/3rdparty/include/ghc/filesystem.h
+++ b/3rdparty/include/ghc/filesystem.h
@@ -1,6 +1,6 @@
 //---------------------------------------------------------------------------------------
 //
-// ghc::filesystem - A C++17-like filesystem implementation for C++11/C++14/C++17
+// ghc::filesystem - A C++17-like filesystem implementation for C++11/C++14/C++17/C++20
 //
 //---------------------------------------------------------------------------------------
 //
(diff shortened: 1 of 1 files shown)


lightningterror (approved):

TheLastRar:
>Makes the wchar-based implementation default on Windows

https://github.com/PCSX2/pcsx2/blob/6a2ed3d0781265e503d02ae907173da5b3368c76/pcsx2/DEV9/DEV9.cpp#L211-L216
https://github.com/PCSX2/pcsx2/blob/6f7890b709d5e3f7f5b824781e493455efc92339/pcsx2/DEV9/Win32/Win32.cpp#L120-L122

Rip

tellowkrinkle:
> Rip

A few extra Unicode conversions never hurt anyone
Though maybe we should at least remove the comment

std::filesystem is also UTF-16 on Windows

In the future you can construct from `.wx_str()` which will expose whatever wx is using internally, guaranteeing the minimum number of conversions

TheLastRar:
Was more miffed as this caught me unaware

>In the future you can construct from .wx_str()

Didn't know about that, thanks for letting me know

PR #4424, opened 2021-05-07
Force-include pch on vs

tellowkrinkle:
CMake does this by default.  This way we can make non-pch builds not take forever by only including what's required instead of the entire pch

### DIFF
--- a/pcsx2/pcsx2.vcxproj
+++ b/pcsx2/pcsx2.vcxproj
@@ -43,6 +43,7 @@
       <ExceptionHandling>Async</ExceptionHandling>
       <PrecompiledHeader>Use</PrecompiledHeader>
       <PrecompiledHeaderFile>PrecompiledHeader.h</PrecompiledHeaderFile>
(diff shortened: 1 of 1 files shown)


tadanokojin:
Something I'm missing here?

https://docs.microsoft.com/en-us/cpp/build/reference/fi-name-forced-include-file?view=msvc-160

> This option has the same effect as specifying the file with double quotation marks in an #include directive on the first line of every source file specified on the command line, in the CL environment variable, or in a command file.

We already have to do this when using a PCH so this should have no side effects outside of just making it less obvious we're using a PCH. I don't really see how this only includes the parts of the PCH that are required in a way that wouldn't also be true had this flag not been specified.

tellowkrinkle:
We currently explicitly include, but in the future I would like to reduce include dependencies (for non-pch builds) by including only required headers, which requires this option so the compiler doesn't scream at us when pch is on

tadanokojin:
Ah, okay I didn't realize this was preparation for something.

I'm slightly worried this will cause some additional inconsistency in the codebase as developers might not actually include the PCH moving forward. I think if we're going to move in the direction of not requiring the include we should remove it across the codebase.

Are you planning to remove the includes as well?

PR #4426, opened 2021-05-08
Add vcxproj filters to cmake

tellowkrinkle:
Adds a (terrible regex-based) parser of the `.vcxproj.filters` files to cmake to allow cmake-generated IDE projects to use the same filters (cmake defaults to throwing all the files into a `Source Files` and `Header Files` folder)

It's not a full xml parser so please be nice to it and make your additions to the `.vcxproj.filters` look like ones VS would generate

Also adds filters for files that were previously missing them.  Please look over to make sure they seem reasonable

BTW, why do some of our filters have separate `Include` directories for headers (`System\Include`, `AppHost\Include`, `System\Ps2\Include`) but not others?


GovanifY:
Shouldn't we try, instead of using filters a la MSVC, to organize the source code organically and not drop every file in the same folder? Would make more sense to handle this at source level instead of the project level

tellowkrinkle:
> Shouldn't we try, instead of using filters a la MSVC, to organize the source code organically and not drop every file in the same folder? Would make more sense to handle this at source level instead of the project level

I can try reorganizing files to match the current MSVC filters, does that seem reasonable?

tellowkrinkle:
Okay it turns out things are a bit messier than I thought.  [I made a script to automatically rename files and fixup includes based on the `.vcxproj.filters`](https://gist.github.com/tellowkrinkle/e00ddfb0c64a6012f57609a10452813b) (run `ruby vcxproj_filters_renamer.rb windows/VCprojects/pcsx2.vcxproj.filters --absolute PrecompiledHeader.h` from the `pcsx2` directory) but there are a bunch of things that would need to happen first:
- Linux-only CDVD / USB files need to be added to the `.vcxproj.filters` so they get picked up by the script
- @tadanokojin expressed a dislike of the current filter layout, and I think changes should be made before we rename everything based on it, as renaming every file is much more invasive than changing some lines in a `.vcxproj.filters` file
- Some filters currently contain multiple files with the same name, which needs to be changed before we can mirror that to a filesystem

Since the plugin merges, I've been running into issues where I search for a file and multiple identical-looking files pop up because the cmake-generated project puts every file in each target into the same group (and we have a large number of files with the same name across the former-plugins).  This PR fixes that issue, and I'd rather not deal with the above issues right now, so I'd like to keep it separate.  The script will remain available if you'd like to use it after dealing with the listed prerequisites.  [I also have a branch with a few edits to the cmake project that the script doesn't work properly with](https://github.com/tellowkrinkle/pcsx2/tree/dir-filter-match)

GovanifY:
> > Shouldn't we try, instead of using filters a la MSVC, to organize the source code organically and not drop every file in the same folder? Would make more sense to handle this at source level instead of the project level
> 
> I can try reorganizing files to match the current MSVC filters, does that seem reasonable?

That would be very much appreciated, yes!

PR #4427, opened 2021-05-08
Workflow improvements

Margen67:
### Description of Changes
Cleanup/maintenance.

### Rationale behind Changes
See description/commits.

### Suggested Testing Steps
Running the CI.


.github/workflows/linux-workflow.yml
```diff
@@ -74,7 +74,7 @@ jobs:
 
       - name: Checkout Submodules
         if: steps.cache-submodules.outputs.cache-hit != 'true'
-        run: git submodule update --init --recursive --jobs 2
+        run: git submodule update --init --recursive --depth 1 -j $(nproc)
```
GovanifY, line 77: why? according to git documentation depth is used to create a shallow clone of only the latest commit, with in this case only the master branch. As git submodules can, in theory, target branches, then that's not something applicable everywhere. nproc doesn't do much here either as it's not a CPU intensive task, but a network bound one, mostly.

.github/workflows/windows-workflow.yml
```diff
@@ -40,32 +40,31 @@ jobs:
     # Set some sort of timeout in the event of run-away builds.  We are limited on concurrent jobs so, get rid of them.
     timeout-minutes: 30
 
+    env:
+      POWERSHELL_TELEMETRY_OPTOUT: 1
```
GovanifY, line 44: we don't really care but nice catch

.github/workflows/windows-workflow.yml
```diff
@@ -95,10 +94,10 @@ jobs:
 
       - name: Prepare Artifact Folder
         shell: bash
-        working-directory: ./bin
+        working-directory: bin
         run: |
           shopt -s extglob dotglob
-          mkdir "${{ steps.artifact-metadata.outputs.name }}"
+          mkdir -p "${{ steps.artifact-metadata.outputs.name }}"
```
GovanifY, line 100: is there a case where the folder is recursive?

GovanifY (changes requested):
I have a few concerns regarding what you propose, otherwise it looks like to be mostly formatting related, which, while being subjective, seems fine to me

.github/workflows/scripts/linux/generate-cmake.sh
```diff
@@ -1,6 +1,4 @@
-#!/bin/bash
-
-set -e
+#!/bin/bash -e
```
tellowkrinkle, line 1: Prefer that we don't do this, it breaks running scripts with `bash script.sh`

.github/workflows/windows-workflow.yml
Margen67: No, it's just for safety.

.github/workflows/linux-workflow.yml
GovanifY: In that case shouldn't we use https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt---no-recommend-shallow instead?

GovanifY: Yes I know they aren't since we currently do a full clone, my point was wouldn't that be the best option there, in case we do need a non shallow clone?

.github/workflows/windows-workflow.yml
GovanifY: No, it's fine!

.github/workflows/linux-workflow.yml
GovanifY: Sure, You can go ahead and do it in this PR? thanks!

.github/workflows/windows-workflow.yml
GovanifY: I don't have any strong feelings on that one, it's fine either way to me, so :shrug:

.github/workflows/linux-workflow.yml
Margen67: I pushed the changes to .gitmodules.

AKuHAK:
I like changes in this PR, the only thing: it is better to use `$(getconf _NPROCESSORS_ONLN)` for calculation core number. The reason - this command will work in all environments (Ubuntu, FreeBSD, macOS, MSYS, etc). And I advise renaming the header of the commit message to `ci: Build improvements`

GovanifY:
Hey, now that the GS merge is done would you mind rebasing your PR? afaict it seems ready to go to me otherwise

GovanifY (approved):

tadanokojin:
Gonna merge this one as-is so if there's any tweaks feel free to follow up.

PR #4429, opened 2021-05-09
DEV9: Update ATA code for wchar GHC

TheLastRar:
### Description of Changes
Use wchar instead of UTF8 for ghc::filesystem::path constructor on windows

### Rationale behind Changes
The GHC dependency got updated to a version that uses wchar instead of UTF8 on windows, see https://github.com/PCSX2/pcsx2/pull/4422

### Suggested Testing Steps
Test unicode paths with HDD files (including testing with the inis folder in a unicode path)

### DIFF
--- a/pcsx2/DEV9/DEV9.cpp
+++ b/pcsx2/DEV9/DEV9.cpp
@@ -1089,11 +1089,7 @@ void ApplyConfigIfRunning(Config oldConfig)
 
 	//Hdd
 	//Hdd Validate Path
-#ifdef _WIN32
-	ghc::filesystem::path hddPath(std::wstring(config.Hdd));
-#else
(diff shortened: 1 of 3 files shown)


TheLastRar:
looking at this again, I realise that I missed making the same changes in https://github.com/PCSX2/pcsx2/blob/2eb6d9a0fb4a63e286efeca787cf12cab59030a9/pcsx2/DEV9/DEV9.cpp#L1097-L1110

GovanifY:
> > > looking at this again, I realise that I missed making the same changes in `ApplyConfigIfRunning()`
> > 
> > 
> > Went ahead and fixed that, should be good to go!
> 
> Thanks, you may want to use `ToString().wx_str()` instead of `ToString().data()` to be match the other fixes

done, thanks!

PR #4430, opened 2021-05-10
Misc: Pattern match to find the BIOS thread address

F0bes:
### Description of Changes
Instead of using hardcoded BIOS addresses, pattern match instructions in the kernel to find the correct address.
This allows BIOSes that have weird checksums or BIOSes that haven't been hardcoded to use the thread tools!

I do this by hooking two thread-related syscalls, `StartThread` and `ChangeThreadPriority`. When either of these are called for the _first time_  it will pattern match the kernel. If you don't know what this means and are worried, don't worry, you will not notice perf loss. 

### Rationale behind Changes
Hard coding the addresses poses a problem when a BIOS is modified or has been dumped in a way that messes with the checksum.

### Suggested Testing Steps
Test the thread and stack views in the debugger with your bios, if they are blank, something is wrong. If it is showing garbage data, something is wrong. There will be a message in the console when the instruction pattern couldn't be found. Please report this if this is the case.

### DIFF
--- a/pcsx2/R5900OpcodeImpl.cpp
+++ b/pcsx2/R5900OpcodeImpl.cpp
@@ -982,8 +982,43 @@ void SYSCALL()
 		case Syscall::SetVTLBRefillHandler:
 			DevCon.Warning("A tlb refill handler is set. New handler %x", (u32*)PSM(cpuRegs.GPR.n.a1.UL[0]));
 			break;
+		case Syscall::StartThread:
+		case Syscall::ChangeThreadPriority:
+		{
+			if (CurrentBiosInformation.threadListAddr == 0)
+			{
+				u32 offset = 0x0;
+				// Suprisingly not that slow :)
+				while (offset < 0x5000) // I find that the instructions are in between 0x4000 -> 0x5000
+				{
+					u32 addr = 0x80000000 + offset;
+					const u32 inst1 = memRead32(addr);
+					const u32 inst2 = memRead32(addr += 4);
+					const u32 inst3 = memRead32(addr += 4);
 
-
+					if (ThreadListInstructions[0] == inst1 && // sw v0,0x0(v0)
+						ThreadListInstructions[1] == inst2 && // no-op
+						ThreadListInstructions[2] == inst3) // no-op
+					{
+						// We've found the instruction pattern!
+						// We (well, I) know that the thread address is always 0x8001 + the immediate of the 6th instruction from here
+						const u32 op = memRead32(0x80000000 + offset + (sizeof(u32) * 6));
+						CurrentBiosInformation.threadListAddr = 0x80010000 + static_cast<u16>(op) - 8; // Subtract 8 because the address here is offset by 8.
+						DevCon.WriteLn("BIOS: Successfully found the instruction pattern. Assuming the thread list is here: %0x", CurrentBiosInformation.threadListAddr);
+						break;
+					}
+					offset += 4;
+				}
+				if (!CurrentBiosInformation.threadListAddr)
+				{
+					// We couldn't find the address
+					CurrentBiosInformation.threadListAddr = -1;
+					// If you're here because a user has reported this message, this means that the instruction pattern is not present on their bios, or it is aligned weirdly.
+					Console.Warning("BIOS Warning: Unable to get a thread list offset. The debugger thread and stack frame views will not be functional.");
+				}
+			}
+		}
+		break;
 		case Syscall::sceSifSetDma:
 			// The only thing this code is used for is the one log message, so don't execute it if we aren't logging bios messages.
 			if (SysTraceActive(EE.Bios))
--- a/pcsx2/ps2/BiosTools.cpp
+++ b/pcsx2/ps2/BiosTools.cpp
@@ -54,57 +54,7 @@ u32 BiosChecksum;
 u32 BiosRegion;
 bool NoOSD;
 wxString BiosDescription;
-const BiosDebugInformation* CurrentBiosInformation;
-
-const BiosDebugInformation biosVersions[] = {
-	// Notes: The SCPH versions have not been verified
-	
-	// USA     v02.00(14/06/2004) SCPH70012
-	{ 0x00000200, 0xD778DB8D, 0x8001A640 },
-	// USA     v01.60(19/03/2002) SCPH39004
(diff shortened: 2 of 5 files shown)


pcsx2/ps2/BiosTools.cpp
```diff
@@ -105,7 +107,9 @@ const BiosDebugInformation biosVersions[] = {
 	// HK      v02.00(14/06/2004) SCPH70006
 	{ 0x00000200, 0x2E5D0C98, 0x8001A640 }
 };
-
+#else
+BiosDebugInformation CurrentBiosInformation;
```
GovanifY, line 111: indentation

pcsx2/ps2/BiosTools.cpp
```diff
@@ -365,7 +370,10 @@ void LoadBIOS()
 		}
 
 		if (CurrentBiosInformation == NULL)
-			Console.Warning("BIOS Warning: Unknown BIOS version. The debugger thread and stack frame views will not be functional.");
+			Console.Warning("BIOS Warning: Unknown BIOS version. The debugger thread and stack frame views will not be functional.\n(This is harmless for 90\% of users)");
```
GovanifY, line 373: Do we really need to add this? Lambda users won't read our emuLog probably

pcsx2/ps2/BiosTools.h
```diff
@@ -14,6 +14,9 @@
  */
 
 #pragma once
+// Uncomment this to use hardcoded addresses for the thread view in the debugger
+// It supports the bioses _only_ hardcoded, nothing more 
+//#define USE_HARDCODED_BIOS_INFORMATION
```
GovanifY, line 19: We don't really need to add this here, it should be done at the project level, eg adding the use flag to cmake/VS

pcsx2/R5900OpcodeImpl.cpp
```diff
@@ -982,8 +982,46 @@ void SYSCALL()
...
+						break;
+					}
+					offset += 4; // We don't know the alignment, incrementing by 32 doesn't work on some bioses. Maybe 2 just to be safe ?
+				}
+				if (!CurrentBiosInformation.threadListAddr)
+				{
+					// We couldn't find the address
+					CurrentBiosInformation.threadListAddr = -1;
+					// If you're here because a user has reported this message, this means that the instruction pattern is not present on their bios and this method will need an update
+					Console.Warning("BIOS Warning: Unable to get a thread list offset. The debugger thread and stack frame views will not be functional.\n(For 90\% of users you just can report this or ignore it)");
```
GovanifY, line 1019: same here

pcsx2/ps2/BiosTools.h
```diff
@@ -28,19 +31,35 @@ namespace Exception
...
 
+const extern BiosDebugInformation* CurrentBiosInformation;
+#else
+// Pattern matching for today :>
+const u32 ThreadListInstructions[3] =
+{
+	0xac420000, // sw v0,0x0(v0)
+	0x00000000, // no-op
+	0x00000000, // no-op
+};
```
GovanifY, line 50: Uhhhh, is this safe? This seems like an awfully generic suite of instructions to me

pcsx2/R5900OpcodeImpl.cpp
```diff
@@ -982,8 +982,46 @@ void SYSCALL()
...
-
+#ifndef USE_HARDCODED_BIOS_INFORMATION
+		case Syscall::StartThread:
+		case Syscall::ChangeThreadPriority:
+		{
+			if (CurrentBiosInformation.threadListAddr == 0)
+			{
+				u32 offset = 0x0;
+				// Suprisingly not that slow :)
+				while (offset < 0x5000) // I pulled this number out of my ass. How far _should_ we read into the kernel?
```
GovanifY, line 993: Is this number accurate or not? Please try it out with more kernels to see if that works and delete that comment, or fix it otherwise :p

GovanifY (changes requested):
Thanks for the PR! I have a few commenst though

pcsx2/ps2/BiosTools.h
F0bes: Kernel isn't too big, from my tests there is only one instance of this instruction in the bios. It gets difficult choosing instructions because they may change depending on the bios.

pcsx2/R5900OpcodeImpl.cpp
F0bes: I find that the instructions are in the 0x4xxxx region. I'll remove the comment

pcsx2/ps2/BiosTools.cpp
F0bes: I'm unsure what lambda you're referring too but if I can make it silently fail I guess.

pcsx2/ps2/BiosTools.h
GovanifY: works for me then!

pcsx2/ps2/BiosTools.h
GovanifY: you can go ahead and remove the old way then, we can always revert

pcsx2/R5900OpcodeImpl.cpp
```diff
@@ -982,8 +982,53 @@ void SYSCALL()
...
+		case Syscall::StartThread:
+		case Syscall::ChangeThreadPriority:
+		{
+			if (CurrentBiosInformation.threadListAddr == 0)
+			{
+				u32 offset = 0x0;
+				// Suprisingly not that slow :)
+				while (offset < 0x5000) // I find that the instructions are in between 0x4000 -> 0x5000
+				{
+					// memRead128 only works with aligned addresses on linux
```
tellowkrinkle, line 994: memRead128 is aligned on *all platforms* and may break at any time when used unaligned on Windows (especially likely if we ever start allowing people to build with clang, but could also happen in the future with a compiler upgrade)

memRead64 is also aligned, and you're only getting away with it because there's no such thing as an aligned 64-bit load in x86

tellowkrinkle: Personally would recommend it

Technically if any of your reads cross page boundaries that aren't mapped in order (doubt it would actually happen as I think most of our pages are mapped pretty much 1-1) you'd be reading the wrong values even if you don't get an alignment exception

PR #4436, opened 2021-05-12
Merge GS, Mcd into core, remove plugin selector/subsystem

GovanifY:
This one has been long coming and a huge pain to get to this point, but it's finally there!

As always for those PR, it's not yet finished and available here for early review before being ready for QA.

This time around the nice list is:

- [x] Update the GUI to fit with our pluginless system
- [x] Remove unused windows files (eg *.def)
- [x] Port to Windows
- [x] Look through the codebase and remove the last references to our plugin system
- [x] Fix up TODOs I added regarding GS state as we can introspect it directly now
- [x] Look back at my savestates callbacks and add a wrapper akin to other subcomponents (with Opening/Closing %s)
- [x] Add a menu entry for GS
- [x] Agree on a consistent formatting everywhere (I am going to assume this is good enough and do the rest in another PR)

And probably more things I forgot!

I also modified a bunch of __forceinline, which we seemed to really love using at stupid places in the past. While I tried my best to avoid touching any hot code we might want to do some benchmarks once this PR is ready for QA to make sure I haven't fucked up everything.

It's time to synthesize 1.8!
screenshot

(yes there is a savestate bump)


qurious-pixel:
`FAILED: pcsx2/CMakeFiles/PCSX2.dir/GS/GSClut.cpp.o`
and
`FAILED: pcsx2/CMakeFiles/PCSX2.dir/GS/GSDrawingContext.cpp.o`

<details>
<summary>Inline error</summary>

`error: inlining failed in call to always_inline '__m128i _mm_blendv_epi8(__m128i, __m128i, __m128i)': target specific option mismatch
 _mm_blendv_epi8 (__m128i __X, __m128i __Y, __m128i __M)`
from "smmintrin.h"
</details>

pcsx2/GS/GSdx.rc
```diff
@@ -393,10 +393,10 @@ BEGIN
             VALUE "CompanyName", "Gabest"
             VALUE "FileDescription", "GS plugin for ps2 emulators"
             VALUE "FileVersion", "1, 2, 0, 0"
-            VALUE "InternalName", "GSdx.dll"
+            VALUE "InternalName", "GS.dll"
```
lightningterror, line 396: Can get rid of the entire VERSIONINFO code block.

pcsx2/GS/Renderers/Common/GSDevice.cpp
```diff
@@ -0,0 +1,519 @@
...
+	HWBlend blend = m_blendMap[index];
+	blend.op = ConvertBlendEnum(blend.op);
+	blend.src = ConvertBlendEnum(blend.src);
+	blend.dst = ConvertBlendEnum(blend.dst);
+	return blend;
+}
+
+uint16 GSDevice::GetBlendFlags(size_t index) { return m_blendMap[index].flags; }
+
+std::array<HWBlend, 3 * 3 * 3 * 3 + 1> GSDevice::m_blendMap =
```
lightningterror, line 435: You can add clang format off and keep the old formatting of the blend table.

pcsx2/GS/Renderers/OpenGL/GSDeviceOGL.cpp
```diff
@@ -991,45 +981,7 @@ GLuint GSDeviceOGL::CompileGS(GSSelector sel)
 
 GLuint GSDeviceOGL::CompilePS(PSSelector sel)
 {
-	std::string macro = format("#define PS_FST %d\n", sel.fst)
```
lightningterror, line 994: Clang off this one too.

GovanifY:
> `FAILED: pcsx2/CMakeFiles/PCSX2.dir/GS/GSClut.cpp.o`
> and
> `FAILED: pcsx2/CMakeFiles/PCSX2.dir/GS/GSDrawingContext.cpp.o`
> Inline error
> 
> `error: inlining failed in call to always_inline '__m128i _mm_blendv_epi8(__m128i, __m128i, __m128i)': target specific option mismatch _mm_blendv_epi8 (__m128i __X, __m128i __Y, __m128i __M)`
> from "smmintrin.h"

-msse4.1

pcsx2/Recording/Utilities/InputRecordingLogger.h
```diff
@@ -28,7 +28,31 @@
 
 namespace inputRec
 {
-	void log(const std::string& log);
-	void consoleLog(const std::string& log);
-	void consoleMultiLog(const std::vector<std::string>& logs);
+	static void log(const std::string log)
```
Warepire, line 31: This is reverting a commit on master.

pcsx2/GS/GSdx.def
```diff
@@ -1,4 +1,4 @@
-; GSdx.def : Declares the module parameters for the DLL.
+; GS.def : Declares the module parameters for the DLL.
```
Warepire, line 1: Forgotten removal of .def file?

GovanifY: not forgotten, I'm waiting for the windows port to do that, noted it in my todo list above :)

pcsx2/Recording/Utilities/InputRecordingLogger.h
GovanifY: Can you please link the commit? Tellow has done this rebase and not me, so I'm not sure about what was changed during the rebase unfortunately

pcsx2/GS/GSdx.rc
GovanifY: can get rid of the whole file* waiting for the windows port to do so

pcsx2/GS/GSdx.def
Warepire: All right.

pcsx2/Recording/Utilities/InputRecordingLogger.h
Warepire: Here's the PR that it's reverting: https://github.com/PCSX2/pcsx2/pull/4353

tellowkrinkle: Oops
I saw that in the merge conflict and thought the change had been in the other direction

pcsx2/Docs/PCSX2_FAQ.md
```diff
@@ -161,14 +160,6 @@ As the name says, speedhacks are hacks to make things faster. They will speed up
...
 
-### Question 26: What are PCSX2 plugins
-PCSX2 uses a plugin framework for various sections of the emulator. A plugin is a small, incomplete piece of software that, when plugged in to another piece of software, provides some sort of additional function. PCSX2 uses plugins for:
-*   Video (GSdx)
-*   Controls (Lilypad for Windows, OnePad for Linux)
-*   USB (No plugin included)
-
-Each plugin will have its own Plugin Settings menu, and can be found under the Config menu.
-
 ### Question 27: Why is my sound garbled up
```
Florin9doi, line 163: Question 26

pcsx2/GS/GS.cpp
```diff
@@ -0,0 +1,2345 @@
...
+	catch (GSRecoverableError)
+	{
+	}
+	catch (const std::bad_alloc&)
+	{
+		fprintf(stderr, "GS: Memory allocation error\n");
+	}
+}
+
+uint32 GSmakeSnapshot(char* path)
```
kenshen112, line 638: make it an std::string or a const char*

kenshen112 (changes requested):

pcsx2/GS/GS.cpp
GovanifY: I am NOT about to rewrite GS in this PR, we have already way more than enough shit to go through

qurious-pixel:
Linux pad is non-responsive even though initialized as printed in the console.
Suspect commit [c1f9861](https://github.com/PCSX2/pcsx2/pull/4436/commits/c1f986175fb25bfbd053c484fa1c41802746976a)

GovanifY:
> Linux pad is non-responsive even though initialized as printed in the console.
> Suspect commit [c1f9861](https://github.com/PCSX2/pcsx2/pull/4436/commits/c1f986175fb25bfbd053c484fa1c41802746976a)
> Program Log

Can you test fdc56e7 and tell me if it worked then?

iMineLink:
Here we go!
Gave this a quick run on Windows (debug build), found an issue.
If I close the GS window while playing the sound does not stop.
If I then close pcsx2 I get this error:

```
D:\projects\pcsx2\pcsx2\System\SysThreadBase.cpp(81) : assertion failed:
    Function:  SysThreadBase::Suspend
    Thread:    MTGS
    Condition: !IsSelf()
    Message:   Suspend/Resume are not allowed from this thread.

Stacktrace:
[00] SysThreadBase::Suspend                       \pcsx2\System\SysThreadBase.cpp:81
[01] SysMtgsThread::CloseGS                       \pcsx2\MTGS.cpp:561
[02] SysMtgsThread::OnSuspendInThread             \pcsx2\MTGS.cpp:567
[03] SysThreadBase::StateCheckInThread            \pcsx2\System\SysThreadBase.cpp:345
[04] SysMtgsThread::ExecuteTaskInThread           \pcsx2\MTGS.cpp:284
[05] Threading::pxThread::_try_virtual_invoke     \common\src\Utilities\ThreadTools.cpp:554
[06] Threading::pxThread::_internal_execute       \common\src\Utilities\ThreadTools.cpp:651
[07] Threading::pxThread::internal_callback_helper \common\src\Utilities\ThreadTools.cpp:702
[08] Threading::pxThread::_internal_callback      \common\src\Utilities\ThreadTools.cpp:689
[09] ptw32_threadStart                            \3rdparty\pthreads4w\ptw32_threadStart.c:184
[10] register_onexit_function                    
[11] BaseThreadInitThunk                         
[12] RtlGetAppContainerNamedObjectPath           
[13] RtlGetAppContainerNamedObjectPath           
```

The same error occurs if I try to pause the emulation while running.

GovanifY:
> Here we go!
> Gave this a quick run on Windows (debug build), found an issue.
> If I close the GS window while playing the sound does not stop.
> If I then close pcsx2 I get this error:

The issue when you close pcsx2/close GS is now fixed; will look more into the audio related bug.

qurious-pixel:
> 
> Can you test [fdc56e7](https://github.com/PCSX2/pcsx2/commit/fdc56e7237b5a06a9939c4ad19268629bc570017) and tell me if it worked then?

Neither [fdc56e7] nor [8bcbda2] worked with pad or keyboard (using default keys).

JordanTheToaster:
Internal resolution and I presume other settings currently do not work the program will crash when I attempt to close it and the percentages along the top look like this.

erdhetdjtrej

pcsx2/Docs/PCSX2_FAQ.md
```diff
@@ -161,18 +160,10 @@ As the name says, speedhacks are hacks to make things faster. They will speed up
...
-
-Each plugin will have its own Plugin Settings menu, and can be found under the Config menu.
-
-### Question 27: Why is my sound garbled up
+### Question 26: Why is my sound garbled up
 PCSX2's default audio playback mode is Timestretch, which will attempt to stretch sounds out when the emulator is unable to play at full speed, to fill the gaps in the sound playback and prevent even harsher stutters. If you are unable to get the game running at full speed or an acceptable speed where timestretching does not ruin the sound, you can try changing from Timestretch to Async, but this may break games.
 
-### Question 28: Is my controller supported
+### Question 27: Is my controller supported
 For Windows users running LilyPad, PCSX2 will ship with XInput bindings already created. This will automatically work with an Xbox One controller or any other XInput controller. Bindings can easily be made for DirectInput controllers, such as Xbox 360 controllers. PS3, PS4, or Switch Pro controllers should be set up using a compatibility software, such as Steam's Big Picture mode.
```
seta-san, line 167: reference to lilypad

pcsx2/Docs/PCSX2_FAQ.md
```diff
@@ -161,18 +160,10 @@ As the name says, speedhacks are hacks to make things faster. They will speed up
...
+### Question 26: Why is my sound garbled up
 PCSX2's default audio playback mode is Timestretch, which will attempt to stretch sounds out when the emulator is unable to play at full speed, to fill the gaps in the sound playback and prevent even harsher stutters. If you are unable to get the game running at full speed or an acceptable speed where timestretching does not ruin the sound, you can try changing from Timestretch to Async, but this may break games.
 
-### Question 28: Is my controller supported
+### Question 27: Is my controller supported
 For Windows users running LilyPad, PCSX2 will ship with XInput bindings already created. This will automatically work with an Xbox One controller or any other XInput controller. Bindings can easily be made for DirectInput controllers, such as Xbox 360 controllers. PS3, PS4, or Switch Pro controllers should be set up using a compatibility software, such as Steam's Big Picture mode. 
 
 For Linux users running OnePad, PCSX2 will automatically detect and bind controls to any recognizable controller. If your controller is not automatically detected, you can try using Onepad Legacy to manually bind controls for it.
```
seta-san, line 169: reference to onepad

lightningterror (dismissed):
FAQ and Config guide docs should be left alone until before release.
The website uses those which are meant for 1.6

pcsx2/GS/GSState.cpp
```diff
@@ -1682,7 +1682,7 @@ void GSState::FlushPrim()
 			{
 				// could be an unsupported draw call
 			}
-			catch (const std::bad_alloc& e)
+			catch (const std::bad_alloc& /* e */)
```
tellowkrinkle, line 1685: Any reason to have it at all?  It's not like knowing we would have named the error `e` helps anyone, given the unhelpfulness of the name `e`

tadanokojin: I'm not really bothered, I just commented it out quick but I can get rid of it all the same.

tadanokojin:
> Neither [fdc56e7] nor [8bcbda2] worked with pad or keyboard (using default keys).
>
> EDIT: Went back to [ff3af8b] and also no gamepad support.
> OnePAD: failed to load ini /path/to/PCSX2/inis/PAD.ini
> (normal message on first run)

> the percentages along the top look like this.

Both issues likely caused by the display not being opened. So both pad and GS are getting null display handles.  The call to `sApp.OpenGsPanel()` was never replaced following it's removal in https://github.com/PCSX2/pcsx2/pull/4436/commits/23c328de019e208f907132e23b323b5a9514f157

> the program will crash when I attempt to close it

Caused by https://github.com/PCSX2/pcsx2/pull/4436/commits/8bcbda2c0fdbd4a6186d639327a1fc35fe2cb8bd

GovanifY:
I'm finally able to work back on this! Rebased it on master,

So, the main issues I can remember of/read back so far are:

- [ ] PAD/Linux doesn't appear to work 
- [ ] Frame Limiter/Panel options do not work
- [ ] Crash on GS lifecycle calls at shutdown

The first two can probably be linked back to the GS Panel handle lifecycle not being respected.
The last one is an issue with GS lifecycle proper, so I'll have to investigate that.

We might also want, either post merge or after we polish up everything in this PR, to stop having each subcomponent its own sstate handling functions and move this to the main sstate struct definition as to avoid having the same setup defined 6 times, along with making GS use the standard sstate struct instead of its own specific one

GovanifY:
I removed the cruft that was left out on the save states for now, making everything but the serializing take place in SysState.cpp.
The object de/serializing is still a part of each component, but all the writer/reader things are now at the same place. It still isn't as clean as I'd like because we have a ton of code repetition(a template would probably make sense there) but good enough for now

GovanifY:
Every obvious bug/thing left to do I've noticed is done so.....


IT IS FINALLY READY FOR QA/TESTING!!!!!

RedDevilus:
Workaround for actions:
https://we.tl/t-Bhv6za8ows

qurious-pixel:
booting an already loaded file (BIOS for example) leads to assertion failure on linux.
on Main branch, it reloads the file after warning.

<details>
<summary> Assert Failure </summary>

assert-failed-PCSX2

</details>

OnePad and automapping work as expected. 

commit https://github.com/PCSX2/pcsx2/pull/4436/commits/388db199d7673e7265198034b18f5c5c6038c785

qurious-pixel:
> Workaround for actions:
> https://we.tl/t-Bhv6za8ows

commenting out these lines, built on Windows Build
https://github.com/PCSX2/pcsx2/blob/9c2f846de0dd4152076e210835127acaf5ccc028/buildbot.xml#L4-L6

GovanifY:
To keep everyone updated: this PR is mostly done, save for one bug now related to threading

Understanding this bug might take a bit of background with our threading mechanism so bear with me for a bit

PCSX2 uses a multi-threaded renderer which we call MTGS, basically every GIF packet is processed in parallel before the rasterizer puts everything together single threaded and we finally get the final picture.

In order to process events, MTGS has a ringbuffer with a fez primite synchronization mechanisms: a read pos, that is only ever touched upon by the GS and a write pos, that is touched outside of the GS. The packets not processed are then the [read-write] interval and as long as we don't have enough packets as to overflow the ringbuffer, everything works great.

Our freeze(understand: save state) mechanism worked by locking the ringbuffer in the GS and adding a new instruction to process. Once this instruction would be gotten to it would wait for past packets to be processed and then serialize GS memory, objects and whatnot.

In order for this process to work, you need the GS to be active, which requires opening a panel handle. This is handled by the main thread of the process: it waits for events, and then start back the GUI handle and give it back.

Unfortunately, the main thread can also be used to process some commands, a la RenderSwitch, memory card windows, etc which would make the GS wait infinitely, obviously not a wise choice, so the most obvious solution in this case would be to avoid a whole resume/shutdown cycle regarding the panel, but that only solves one threading issue.

There's then the calls that are done in the save states, which are executed on the SysExecutor threads. This is a thousand times worse for a single reason: SysExecutor is pretty much our event dispatcher, so GS will absolutely deadlock waiting for SysExecutor to resume, while SysExecutor will wait for GS to process the sstate event. Making the call non thread safe could then, in theory, be a solution, since MTGS is stalling for events right? Unfortunately, only the packet processing thread is stalled for events and we have no control over the other threads active, and a Wait event doesn't seem to work. 
While I think I pretty much fixed the latter I still get some nullptr troubles in GSfreeze, so I assume I fucked up something.


And that's where we are now, this is the single bug left that I'm aware of for this PR and it's turning out to be a huge pain. In the past this whole mechanism worked by having a PluginManager thread different from SysExecutor and/or our main thread, which made all calls to GSfreeze on a redundant thread not locking up anything, but that's a fairly poor design choice.

I've tried to rewrite our synchronization primitives, the freezing mechanism, tried to take advantage of the core being locked up etc but for now no dice, this issue is turning out to be a huge pita that none of us anticipated and that we can thank the plugin subsystem for.

tadanokojin:
With regards to windows:

Compile is improving, the issues with texture corruption should be sorted now. Additionally I added a build configuration for AVX2 as that had gone missing. Debug builds are throwing linker errors if they are compiled after a devel build. It's just `pcsx2.exe` for the target right now for both avx and avx2 which is going to have to change.

Additionally there was a mistake made when moving the gs PCH stuff around that is currently breaking the AVX2 compile. Linux and Windows handle the macros for AVX2 differently. Linux reads another macro and defines _M_SSE based on that where as windows always just defines `_M_SSE`. On unix this masks an issue that arises from `GS.h` (the old GSdx one) needing to be included in all fines that also include `GSVector.h` and thus `GSVector8i.h`. The issue is that `GSVector8i.h` expects the macro `ASSERT` to be defined. This macro was previously housed in the PCH but has now been moved to `GS.h`. However, since linux gets it's `_M_SSE` set in `GS.h` it only sets it in files that include that file. This means that in other files where `GS.h` is not included before hand, the contents of `GSVector8i.h` are effectively nopped out whereas on Windows it wouldn't be.

I think we need a more consistent way of dealing with this across the two platforms.

GovanifY:
The threading bug has finally been found and fixed, which _should_ put us at feature parity with master! This leaves us the msvc compiling defines to sort out and we should be ready for another, and hopefully final, round of testing!

lightningterror (changes requested):
Emulator still locks up if you try to close both windows when a game is running from the taskbar with "close all windows".

GovanifY:
> Emulator still locks up if you try to close both windows when a game is running from the taskbar with "close all windows".

Can't repro after 2b8a0b6

GovanifY:
> Emulator still locks up if you try to close both windows when a game is running from the taskbar with "close all windows".

fixed in e4aee82

lightningterror (approved):
The crash is fixed, I haven't done much testing tho, a job for the testers.

GovanifY:
Testing was already heavily done on pretty much everything we can think of, at this point everything on my side is ready

lightningterror:
Can you just properly rename Tellows formatting commit with prefix?
https://github.com/PCSX2/pcsx2/pull/4436/commits/31d3c537f2ac31f3c249aa5d3ab0e445c3ed5038

GovanifY:
done

tadanokojin:
I'm not sentimental. Good riddance.

GovanifY:
you may call me the ~~demon~~ plugin slayer

tymondouglas:
Does it have native support for PS3 pressure sensitive buttons now? Because that was reliant on a modified version of Lilypad.

RedDevilus:
https://forums.pcsx2.net/Thread-DsHidMini-Windows-10-driver-for-the-DualShock-3

Use this instead of modified Lilypad or SCP.

WMan22:
What does this mean for the light gun plugin like Nuvee, or the steering wheel plugin like USBQemu? Would be a shame to lose that functionality. Don't get me wrong; I'm not gonna gripe in the slightest about the loss of plugins as long as the edge case functionalities they provided are natively implemented instead of just simply unusable now. 

Peripheral based gaming is part of people's nostalgia for the PS2, so it would earnestly be a downer if that functionality is just simply gone now in newer versions of the emulator.

refractionpcsx2:
USB emulation is integrated in to the core.  Lightgun support is in development but isn't in there yet, so keep 1.6 on hand if you want to play lightgun games for now.

WMan22:
> USB emulation is integrated in to the core. Lightgun support is in development but isn't in there yet, so keep 1.6 on hand if you want to play lightgun games for now.

Well, as long as it's in development, no harm, no foul.

ROBYER1:
> USB emulation is integrated in to the core. Lightgun support is in development but isn't in there yet, so keep 1.6 on hand if you want to play lightgun games for now.

See Light gun support re-added here
https://github.com/PCSX2/pcsx2/pull/7534

PR #4445, opened 2021-05-18
Extended argument support

AKuHAK:
Enable support for the combination of CLI arguments: --gameargs and --elf (or --irx) if the path to .elf (or .irx) does not contain spaces.

### Description of Changes
This PR will allow combining CLI `--elf/--irx` with `--gameargs`. There is a silent check if the path contains spaces. If the path contains spaces `gameargs` will be ignored like it was before this PR.

### Rationale behind Changes
This will improve homebrew testing. Before developers were forced to make an ISO file to test an app with custom arguments or use some other workarounds.

The check for spaces in the path is necessary because of argument support logic in the project. If the path string contains spaces, later this string will be cut into arguments by space delimiter. The first part will be used as a path, other parts - as arguments. Thus it is necessary to perform this check, or completely rewrite argument parsing logic.

### DIFF
--- a/pcsx2/gui/AppInit.cpp
+++ b/pcsx2/gui/AppInit.cpp
@@ -513,6 +513,8 @@ bool Pcsx2App::OnInit()
 			g_Conf->EmuOptions.UseBOOT2Injection = true;
 
 			sApp.SysExecute(Startup.CdvdSource, Startup.ElfFile);
+			if (strchr(Startup.ElfFile, ' ') != NULL)
+				g_Conf->CurrentGameArgs = Startup.GameLaunchArgs;
 		}
 		else if (Startup.SysAutoRunIrx)
 		{
@@ -522,6 +524,8 @@ bool Pcsx2App::OnInit()
 
 			// FIXME: ElfFile is an irx it will crash
 			sApp.SysExecute(Startup.CdvdSource, Startup.ElfFile);
+			if (strchr(Startup.ElfFile, ' ') != NULL)
+				g_Conf->CurrentGameArgs = Startup.GameLaunchArgs;
 		}
 	}
 	// ----------------------------------------------------------------------------


AKuHAK:
@fjtrujy @F0bes this PR can be useful for you.

F0bes:
Have you tested this? I can't get it to work. It seems to not properly set `a0` (argc) so our eeloadHook doesn't copy the arguments over.
https://github.com/PCSX2/pcsx2/blob/6f7890b709d5e3f7f5b824781e493455efc92339/pcsx2/R5900.cpp#L609-L610

AKuHAK:
> Have you tested this? I can't get it to work. It seems to not properly set `a0` (argc) so our eeloadHook doesn't copy the arguments over.
> https://github.com/PCSX2/pcsx2/blob/6f7890b709d5e3f7f5b824781e493455efc92339/pcsx2/R5900.cpp#L609-L610

Yes, I tested Windows build on Retroarch and ps2sdk samples, arguments are accepted just fine. Could you please provide an example where it fails?

F0bes:
This is what I'm doing, I'm pretty sure it's correct

My homebrew:
```
int main(int argc, char **argv)
{
	printf("%d\n",argc);
	for(int i = 0; i < argc; i++)
	{
		printf("[%d] %s\n",i,argv[i]);
	}
       SleepThread();
}
```

My pcsx2 arguments:
`--gameargs="-arg1 -arg2" --elf="P://c++/vu1draw/vu1.elf"`
(I didn't create a new project to test this)

Output:
```
1
[0] host:P://c++/vu1draw/vu1.elf
```

I can also say that argv[1] _is_ a null string

AKuHAK:
@F0bes Thank you, don't know why I used reversed logic in the final commit. Simply replace "!=" with "==" did the trick. Here is my tests with wlaunchelf:
```
Loaded, host:.¥wle.elf
start address 0x1d00008
gp address 00000000
eeloadHook2: Supplying launch argument(s) '1 2 3 4 5' to ELF 'host:.\wle.elf'.
eeloadHook2: arg block is 'host:.\wle.elf 1 2 3 4 5'.
ParseArgumentString: Saving these strings:
0000000000090A80 -> 'host:.\wle.elf'.
0000000000090A8F -> '1'.
0000000000090A91 -> '2'.
0000000000090A93 -> '3'.
0000000000090A95 -> '4'.
0000000000090A97 -> '5'.
eeloadHook2: Writing address 0000000000090A80 to location 0000000000090A68.
eeloadHook2: Writing address 0000000000090A8F to location 0000000000090A6C.
eeloadHook2: Writing address 0000000000090A91 to location 0000000000090A70.
eeloadHook2: Writing address 0000000000090A93 to location 0000000000090A74.
eeloadHook2: Writing address 0000000000090A95 to location 0000000000090A78.
eeloadHook2: Writing address 0000000000090A97 to location 0000000000090A7C.
eeloadHook2: Saving 6 and 0000000000090A68 in $a0 and $a1.
# Restart Without Memory Clear.

```

F0bes (approved):
My example program now works.
Looks good to me.
Let me check on linux.

AKuHAK:
Tested on MacOS with command
```sh
open -a ~/Applications/pcsx2/PCSX2.app --args --elf=/Users/user/wle.elf --gameargs="1 2 3 4 5"
```
works as expected. The only thing: it seems that `--elf` doesn't support relative paths (like `~/wle.elf`) - only absolute.

tellowkrinkle:
> it seems that `--elf` doesn't support relative paths (like `~/wle.elf`) - only absolute.

Those are often expanded by the shell, not the program, which isn't happening because the argument isn't by itself (has `--elf=` before it).  If PCSX2 supports gnu-style arguments (`--elf ~/wle.elf`), that would solve it.  Otherwise there's always `$HOME/` which works anywhere

AKuHAK:
So far I tested this PR in all environments (Windows, Ubuntu, macOS), @PCSX2 team, is this PR ok?

AKuHAK:
> > it seems that `--elf` doesn't support relative paths (like `~/wle.elf`) - only absolute.
> 
> Those are often expanded by the shell, not the program, which isn't happening because the argument isn't by itself (has `--elf=` before it). If PCSX2 supports gnu-style arguments (`--elf ~/wle.elf`), that would solve it. Otherwise there's always `$HOME/` which works anywhere
$HOME or any other variables also are not working, only absolute paths. Probably to get them working changes, like in this PR: https://github.com/PCSX2/pcsx2/pull/4451 should be added.

RedDevilus:
What's up with the second and third commit of this, a way to rebase?

AKuHAK:
> What's up with the second and third commit of this, a way to rebase?

As I know these "merged" commits will not appear in the main pcsx2 history after merging. So you can just ignore them.

tellowkrinkle:
> $HOME or any other variables also are not working, only absolute paths. Probably to get them working changes, like in this PR: #4451 should be added.

Did you test this?  I just ran [the 20210327 build](https://github.com/tellowkrinkle/pcsx2/releases/tag/macos-20210327) and it worked just fine with both `--elf=$HOME/wle.elf` and with `--elf ~/wle.elf` in both bash and fish.

> As I know these "merged" commits will not appear in the main pcsx2 history after merging. So you can just ignore them.

AFAIK they will appear if you don't remove them.  To remove them, you'll want to rebase with `git rebase -i upstream/master` and delete them

AKuHAK:
@tellowkrinkle thank you for this information. Maybe your version has more changes in comparision to mainstream, cause I never get to work everything except of absolute paths on Linux and Mac. Anyway, this is not purpose of this PR and isnt directly related to it.

Regarding rebasing, thank you for the info, i was sure that this commits will appear only in my history and will not appear. I will rebase the PR later.

AKuHAK:
@RedDevilus @pcsx2 this PR is ready for merging now.

RedDevilus (approved):
`// FIXME: ElfFile is an irx it will crash`
Does still match i wonder? Otherwise seems fine.

F0bes (approved):
Rebased on to master and tested again. Works perfectly. Good to merge.

F0bes:
Ok, I rebased it and tested it. It works.

Issue #4443, opened 2021-05-16
[Feature Request] HDD fixed 100% allocation


RedDevilus:
**Description**

Dynamically allocation of HDD file or compress unused space. Not allocate all specified HDD space at creation.
**Reason**

There is an active PR that will increase the minimum of HDD size from 8 GiB to 40 GiB. Some games expect atleast this amount but for normal endusers this is wasted space if you only use to about 10 GiB. So if the enduser only uses about 9 GiB then it should grow from 8 GiB to 12 GiB , if the user uses 10.5 Gig it grows to 16 GiB, etc. Another method could be to change existing HDD files manually if needed.

In essence:
- Game won't nag about not enough space.
- End-users doesn't need to make more space on their OS and/or create new HDD file.
- Almost no wasted space.

**Examples**

VMWare method (dynamically allocation or the growth by certain amount):
![image](https://user-images.githubusercontent.com/24227051/118415203-5d8a3900-b6a9-11eb-8a9b-4e6a0ed7709a.png)

tadanokojin:
I think it might be a better idea to use sparse files:
https://docs.microsoft.com/en-us/windows/win32/fileio/sparse-files

I think linux supports it too. Only issue is you interface with an OS api.

qurious-pixel:
Linux does support sparse files, but there are some filesystems that do not, such as FAT, i.e. ExFat.

tellowkrinkle:
AFAIK linux and macOS will make a sparse file if your filesystem supports it if you resize with `ftruncate`, no need to do anything special

Alternatively, you can seek to the end and write a byte (something we already do... before writing zeros over the middle section)

RedDevilus:
Maybe something similar to how provisioning works in virtualisation?

```
Thin: Allocate and zero on first write
Thick Lazy: Allocate in advance and zero on first write
Thick Eager: Allocate and zero in advance
```

TheLastRar:
I didn't actually know about the OS api's for sparse files.

Windows doesn't automatically deallocate when zeros are written (and will allocate when zero is written normally), you have to use the ``FSCTL_SET_ZERO_DATA IOCTL`` instead.
Does linux have a similar requirement?
And do we even need to care about that? as erased files probably won't be zeroed anyway.

Is is possible to determine if the filesystem supports sparse files on linux (windows has ``GetVolumeInformation()``)
[...]

PR #4446, opened 2021-05-18
DEV9 HDD Sparse file support

TheLastRar:
### Description of Changes
Uses sparse files for HDD images.
It will also check for zero writes to the HDD file and hole punch if large enough.
Note that it will not make existing files sparse (although writes may hole punch on Linux & Mac)
Currently only 1mb is allocated, with the remainder of the file left unallocated.

### Rationale behind Changes
A minimum of 40GB is required for most games, but they won't need most of that size. The storage cost can be mitigated using the OS/filesystem sparse file support.
Resolves https://github.com/PCSX2/pcsx2/issues/4443

### Suggested Testing Steps
Test creating a new HDD file, test HDD enabled games or homebrew
On windows, you can check the "Size on disk" to see the amount of storage actually taken.
Mac and Linux also need testing.

### DIFF
--- a/pcsx2/DEV9/ATA/ATA_Transfer.cpp
+++ b/pcsx2/DEV9/ATA/ATA_Transfer.cpp
@@ -212,6 +212,112 @@ void ATA::IO_SparseCacheLoad()
 	// Store file pointer.
 	const s64 orgPos = FileSystem::FTell64(hddImage);
 
+	// Flush so that we know what is allocated.
+	std::fflush(hddImage);
+
+#ifdef _WIN32
+	// FlushFileBuffers is required, hddSparseBlock differs from actual file without it.
+	FlushFileBuffers(hddNativeHandle);
+	// Range to be examined (One Sparse block size).
+	FILE_ALLOCATED_RANGE_BUFFER queryRange;
+	queryRange.FileOffset.QuadPart = HddSparseStart;
+	queryRange.Length.QuadPart = hddSparseBlockSize;
+
+	// Allocated areas info.
+	FILE_ALLOCATED_RANGE_BUFFER allocRange;
+	DWORD dwRetBytes;
+	const BOOL ret = DeviceIoControl(hddNativeHandle, FSCTL_QUERY_ALLOCATED_RANGES, &queryRange, sizeof(queryRange), &allocRange, sizeof(allocRange), &dwRetBytes, nullptr);
+
+	if (ret == TRUE && dwRetBytes == 0)
+	{
+		// We are sparse.
+		memset(hddSparseBlock.get(), 0, hddSparseBlockSize);
+		hddSparseBlockValid = true;
+#if defined(PCSX2_DEBUG) || defined(PCSX2_DEVBUILD)
+
+		// Store file pointer.
+		const s64 orgPos = FileSystem::FTell64(hddImage);
+		pxAssert(orgPos != -1);
+
+		//Load into check buffer.
+		FileSystem::FSeek64(hddImage, HddSparseStart, SEEK_SET);
+
+		std::unique_ptr<u8[]> temp = std::make_unique<u8[]>(hddSparseBlockSize);
+		memset(temp.get(), 0, hddSparseBlockSize);
+
+		if (FileSystem::FSeek64(hddImage, HddSparseStart, SEEK_SET) != 0 ||
+			std::fread((char*)hddSparseBlock.get(), readSize, 1, hddImage) != 1)
+			pxAssert(false);
+
+		// Restore file pointer.
+		if (FileSystem::FSeek64(hddImage, orgPos, SEEK_SET) != 0)
+			pxAssert(false);
+
+		// Check if file is actully zeros.
+		if (memcmp(hddSparseBlock.get(), temp.get(), hddSparseBlockSize) != 0)
+		{
+			Console.WriteLn("DEV9: ATA: Sparse area not sparse, BlockStart: %s, BlockEnd: %s",
+				std::to_string(HddSparseStart).c_str(), std::to_string(HddSparseStart + hddSparseBlockSize).c_str());
+		}
+		else
+			Console.WriteLn("DEV9: ATA: Sparse area is sparse (Yay), BlockStart: %s, BlockEnd: %s",
+				std::to_string(HddSparseStart).c_str(), std::to_string(HddSparseStart + hddSparseBlockSize).c_str());
+
+		pxAssert(memcmp(hddSparseBlock.get(), temp.get(), hddSparseBlockSize) == 0);
+#endif
+		return;
+	}
+#elif defined(__POSIX__)
+#ifdef SEEK_HOLE
+	// Are we in a hole?
+	off_t ret = lseek(hddNativeHandle, HddSparseStart, SEEK_HOLE);
+	if (ret == (off_t)HddSparseStart)
+	{
+		// Seek to data.
+		ret = lseek(hddNativeHandle, HddSparseStart, SEEK_DATA);
+		if (ret >= (off_t)(HddSparseStart + hddSparseBlockSize))
+		{
+			// We are sparse.
+			memset(hddSparseBlock.get(), 0, hddSparseBlockSize);
+			hddSparseBlockValid = true;
+#if defined(PCSX2_DEBUG) || defined(PCSX2_DEVBUILD)
+
+			// Store file pointer.
+			const s64 orgPos = FileSystem::FTell64(hddImage);
+			pxAssert(orgPos != -1);
+
+			// Load into check buffer.
+			FileSystem::FSeek64(hddImage, HddSparseStart, SEEK_SET);
+
+			std::unique_ptr<u8[]> temp = std::make_unique<u8[]>(hddSparseBlockSize);
+			memset(temp.get(), 0, hddSparseBlockSize);
+
+			if (FileSystem::FSeek64(hddImage, HddSparseStart, SEEK_SET) != 0 ||
+				std::fread((char*)hddSparseBlock.get(), readSize, 1, hddImage) != 1)
+				pxAssert(false);
+
+			// Restore file pointer.
+			if (FileSystem::FSeek64(hddImage, orgPos, SEEK_SET) != 0)
+				pxAssert(false);
+
+			// Check if file is actully zeros.
+			if (memcmp(hddSparseBlock.get(), temp.get(), hddSparseBlockSize) != 0)
+			{
+				Console.WriteLn("DEV9: ATA: Sparse area not sparse, BlockStart: %s, BlockEnd: %s",
+					std::to_string(HddSparseStart).c_str(), std::to_string(HddSparseStart + hddSparseBlockSize).c_str());
+			}
+			else
+				Console.WriteLn("DEV9: ATA: Sparse area is sparse (Yay), BlockStart: %s, BlockEnd: %s",
+					std::to_string(HddSparseStart).c_str(), std::to_string(HddSparseStart + hddSparseBlockSize).c_str());
+
+			pxAssert(memcmp(hddSparseBlock.get(), temp.get(), hddSparseBlockSize) == 0);
+#endif
+			return;
+		}
+	}
+#endif
+#endif
+
 	// Load into cache.
 	if (orgPos == -1 ||
 		FileSystem::FSeek64(hddImage, HddSparseStart, SEEK_SET) != 0 ||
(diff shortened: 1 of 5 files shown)


RedDevilus:
120GB:
image
Recognised as 20GB.

40 GB is recognised as 40GB.

Though it screwed other filesizes in the explorer (i put in different folder on external hard drive:
image
image

Naturally it doesn't affect the actual sizes:
image

It could spook some users.

pcsx2/DEV9/ATA/ATA_Transfer.cpp
```diff
@@ -98,18 +107,98 @@ bool ATA::IO_Write()
...
 	{
-		Console.Error("DEV9: ATA: File write error");
-		pxAssert(false);
-		abort();
+		//What size block is best (must fit evenly into 512 bytes)?
+		const int sparseSize = 512;
+		const static u8 zeroBlock[sparseSize]{0};
+		for (size_t i = 0; i < entry.length; i += sparseSize)
+		{
+			if (memcmp(&entry.data[i], zeroBlock, sparseSize) == 0)
```
tellowkrinkle, line 117: I feel like there has to be a better way to do this than zero a block of stack and memcmp, especially for non-fixed block size

stackoverflow says `!x[0] && memcmp(x, x+1, len-1) == 0`
or you can have some fun with sse
```cpp
bool all_zero_aligned64(const void* data, size_t len)
{
	using V = __m128i;
	constexpr int cacheline = 64 / sizeof(V);
	const V* src = static_cast<const V*>(data);
	const V* end = reinterpret_cast<const V*>(static_cast<const char*>(data) + len);
	for (; src < end; src += cacheline)
	{
		V acc = _mm_setzero_si128();
		for (int j = 0; j < cacheline; j++)
			acc = _mm_or_si128(acc, _mm_load_si128(src + j));
		if (65535 == _mm_movemask_epi8(_mm_cmpeq_epi8(acc, _mm_setzero_si128())))
			return false;
	}
	return true;
}
```

pcsx2/DEV9/ATA/HddCreate.cpp
```diff
@@ -81,32 +86,102 @@ void HddCreate::WriteImage(ghc::filesystem::path hddPath, int reqSizeMiB)
...
 	}
+	newImage.close();
 
-	//Size file
-	newImage.seekp(((u64)reqSizeMiB) * 1024 * 1024 - 1, std::ios::beg);
-	const char zero = 0;
-	newImage.write(&zero, 1);
+	//Size file (Slow on FS that don't support sparse)
+	std::error_code ec;
+	ghc::filesystem::resize_file(hddPath, ((u64)fileMiB) * 1024 * 1024, ec);
```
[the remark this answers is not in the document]
TheLastRar: Looking at the ghc::filesystem source, it calls ``SetEndOfFile()``.
``SetEndOfFile()`` states that the contents of the extended part of the file is undefined.

On NTFS, extending a file reserves disk space but does not zero out the data immediately, instead It will do that lazily on demand, see https://devblogs.microsoft.com/oldnewthing/20110922-00/?p=9573 and https://devblogs.microsoft.com/oldnewthing/20150710-00/?p=45171.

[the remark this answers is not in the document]
TheLastRar: Maybe instead we should set it sparse before resizing?

pcsx2/DEV9/ATA/ATA_Transfer.cpp
TheLastRar: > stackoverflow says `!x[0] && memcmp(x, x+1, len-1) == 0`

I did do a search before writing that code to see if there was a better way. 
In one of those answers they benchmarked that method against memcmp on a temporary stack block (like I'm using) and found it to be 4x slower.

> or you can have some fun with sse

I wasn't sure about including a large amount of code to perform this task, but it seems like your suggesting that is the way to go.

TheLastRar: I couldn't get the sse version to work correctly, It was giving me false positives for some reason
I grabbed another implementation from stackoverflow.

pcsx2/DEV9/ATA/ATA_Transfer.cpp
```diff
@@ -98,18 +107,98 @@ bool ATA::IO_Write()
...
 
 	hddImage.seekp(entry.sector * 512, std::ios::beg);
-	hddImage.write((char*)entry.data, entry.length);
-	if (hddImage.fail())
+	if (hddSparse)
 	{
-		Console.Error("DEV9: ATA: File write error");
-		pxAssert(false);
-		abort();
+		//What size block is best (must fit evenly into 512 bytes)?
```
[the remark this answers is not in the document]
TheLastRar: Can confirm this is also an issue on windows, additionally, NTFS also needs 64Kb sparse blocks in most cases.

Managed to implement a system to address this, (by checking a whole block on each write).

TheLastRar:
I still need to look into the issue Red has encountered

qurious-pixel:
On Exfat, creates full 40GB file over ~30-40 seconds.
Might be worth a check to fail if not a sparse compatible filesystem.

<details>
<summary>Sparse File</summary>

image
</details>
<details>
<summary>ExFat drive</summary>

exfat-sparse
</details>
<details>
<summary>Ext4 drive</summary>

ext4-sparse

</details>

TheLastRar:
> On Exfat, creates full 40GB file over ~30-40 seconds.
> Might be worth a check to fail if not a sparse compatible filesystem.
> Sparse File
> ExFat drive
> Ext4 drive

I'm not sure if there is an easy way to check if sparse files are supported on linux & mac, at least without trying to make one.

On master do you wait a significant amount of time at 0 / 40960 MiB? or does it proceed writing instantly?

ExFat apparently supports Valid Data Length, as I'm surprised there is such a delay in creating the file, even if sparse file support is missing, Wonder if windows (and mac if it supports it) handles that.

That being said, I probably should do a full zero on such filesystems anyway to avoid a performance hit during emulation.

qurious-pixel:
> On master do you wait a significant amount of time at 0 / 40960 MiB? or does it proceed writing instantly?
> 
On master there is also a significant delay when writing to ExFat.
Writing to Ext4 ~6 seconds for 40GB.

tellowkrinkle:
> mac if it supports it

macOS appears to support it for exFAT, though HFS+ doesn't have it

However the `ftruncate` call is always fast (even on HFS+), the OS doesn't make you wait until you try to close the file, so the following flow would work:
```cpp
ftruncate(fd, size);
struct stat s;
fstat(fd, &s);
if (s.st_blocks == size / 512)
{
	// file is not sparse, write zeros with progress bar
}
```

<details>
<summary>Testing procedure and results</summary>
Code:

```cpp
#include <stdio.h>
#include <unistd.h>
#include <fcntl.h>
#include <time.h>
#include <stdlib.h>
#include <sys/stat.h>

	unsigned long long mb = 1024 * 1024;
	int fd = open(argv[1], O_RDWR | O_CREAT, 0644);
	int size = atoi(argv[2]);
	time_t start = time(NULL);
	printf("Size: %dMB\n", size);
	ftruncate(fd, size * mb);
	struct stat s;
	fstat(fd, &s);
	printf("%lds: ftruncate done, %lld blocks\n", time(NULL) - start, s.st_blocks);
	char zero = 0;
	write(fd, &zero, 1);
	printf("%lds: Write at beginning done\n", time(NULL) - start);
	lseek(fd, size * mb / 2 - 1, SEEK_SET);
	write(fd, &zero, 1);
	printf("%lds: Write at middle done\n", time(NULL) - start);
	lseek(fd, size * mb - 1, SEEK_SET);
	write(fd, &zero, 1);
	printf("%lds: Write at end done\n", time(NULL) - start);
	close(fd);
	printf("%lds: File closed\n", time(NULL) - start);
	return 0;
}
```

On APFS:

```
Size: 1024MB
0s: ftruncate done, 0 blocks
0s: Write at beginning done
0s: Write at middle done
0s: Write at end done
0s: File closed
```

On HFS+ (Note: The current implementation, which zeros every block instead of writing 3 times, fills some FS write cache and ends up with a fairly nice looking progress bar):

```
Size: 1024MB
0s: ftruncate done, 2097152 blocks
0s: Write at beginning done
0s: Write at middle done
0s: Write at end done
12s: File closed
```

On exFAT:

```
Size: 1024MB
0s: ftruncate done, 2097152 blocks
0s: Write at beginning done
8s: Write at middle done
16s: Write at end done
17s: File closed
```

</details>

tellowkrinkle:
@qurious-pixel could you try testing the code in "Testing procedure and results" on Linux?

Squall-Leonhart:
exfat is not sparse file capable.

pcsx2/DEV9/ATA/ATA_Transfer.cpp
```diff
@@ -210,7 +210,8 @@ void ATA::IO_SparseCacheLoad()
 	std::streampos orgPos = hddImage.tellp();
 	//Load into cache
 	hddImage.seekg(HddSparseStart, std::ios::beg);
-	hddImage.read((char*)hddSparseBlock.get(), readSize);
+	//Suppress Check buffer boundaries warning
+	hddImage.read((char*)hddSparseBlock.get(), readSize); /* Flawfinder: ignore */
```
TheLastRar, line 214: I suspect this is a false positive, but I have added additional bounds checking code into the function.

Not sure what further is needed, so I've chosen to add an ignore comment, I can remove the comment if that is preferred.

TheLastRar:
> 
> 
> 120GB:
> image
> Recognised as 20GB.
> 
> 40 GB is recognised as 40GB.
> 
> Though it screwed other filesizes in the explorer (i put in different folder on external hard drive:
> image
> image

I've not been able to reproduce this, can you see if you can reproduce, but I only checked on the later commits
Can you test if the issue still occurs on the lastest commits

RedDevilus:
Same thing on external HDD:
Rebooting fixes the view though:

image
image

RedDevilus:
image

I'm wondering if it's just how specific external HDD handle it, because it's fine in it's default location.

Squall-Leonhart:
thats a windows 10 sparse files bug

TheLastRar:
I've added code that uses OS functions to query for sparse zones when checking if the region of the file is already zeroed. I had issues implementing this in the past, but my second attempt seems to not have issues in my brief testing.

I had tried to replace the std::stream code with OS file functions as well, but was unable to get that functioning correctly.
As such, I'm still mixing OS file functions with std::stream.

knight-ryu12:
Test result: with 40GB Sparse.
Dark Chronicle: Worked.
Ace Combat 4: Worked.

Tested under Windows 11 Insider Preview.

Mrlinkwii (approved):

refractionpcsx2:
Aside from needing a rebase, is anything else required here?

TheLastRar:
Rebase, along with adjusting the code to take better advantage of the filesystem functions where possible, I think

TheLastRar:
That was a challenging rebase.

As we had switched from `std::fstream` to `FILE` i took the opportunity to use get the `HANDLE`/`fd` from the `FILE` handle, rather than having to reopen the file as before.

I also did a quick round of testing on both Windows and Linux, allowing me to catch and fix a few bugs that cropped up while working on the above changes.

refractionpcsx2:
Nice work :) So is this ready for testing?

TheLastRar:
Yeah

Mrlinkwii (approved):

JordanTheToaster (approved):

refractionpcsx2:
Only formatting things I can see which could do with addressing are:

your // commends should be followed by a space, and periods at the end of the comments

and if you can, could you sort this codacy warning if it's addressable?

> 
> pcsx2/DEV9/ATA/ATA_Transfer.cpp#L273
> 
> Skipping configuration 'PCSX2_DEBUG;PCSX2_DEVBUILD;SEEK_HOLE;__POSIX__' since the value of 'SEEK_HOLE' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly

lightningterror (commented):
Some new compiler warnings, should be taken care of
https://github.com/PCSX2/pcsx2/actions/runs/3587370985/jobs/6037636196#step:11:1139

TheLastRar:
I believe I have address the suggests for the comments & the compiler warning.

As for the codacy warning, I'm not sure I can address that, I guess that codacy isn't able to pick the value of `SEEK_HOLE` from the linux headers (It's `4` on my system) `SEEK_HOLE` is not POSIX standard but supposedly widely supported.

Maybe someone with access to the codacy patterns page for pcsx2 can manully add the value for SEEK_HOLE.
An alternative would be to hardcode a value 4 on in the offending lseek call, with a static_assert to ensure `SEEK_HOLE == 4`

refractionpcsx2:
That's fine then :) Don't worry too hard about it, I figured it might be Codacy just being stupid.

RedDevilus:
Works fine now (tested on C drive and external 3.5 HDD):

image

PR #4449, opened 2021-05-21
DEV9: Fix potential out of bounds access when manual DNS1 & automatic DNS2 are used together.

TheLastRar:
### Description of Changes
Fix an off-by-one error for array access.

### Rationale behind Changes
If DNS1 was manually specified, but DNS2 set to be automatically assigned, an off by one error could result in an out of bounds access to an array of adapter DNS addresses.

### Suggested Testing Steps
Enable DHCP intercept
Test manual DNS1 & automatic DNS2 with an adapter that has DNS addresses assigned to it.

### DIFF
--- a/pcsx2/DEV9/InternalServers/DHCP_Server.cpp
+++ b/pcsx2/DEV9/InternalServers/DHCP_Server.cpp
@@ -356,10 +356,12 @@ namespace InternalServers
 						Console.Error("DHCP: RouterIP missmatch");
 					break;
 				case 6:
+					// clang-format off
 					if (((((DHCPopDNS*)dhcp.options[i])->dnsServers.size() == 0 && dns1.integer == 0) ||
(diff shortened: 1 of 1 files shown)


pcsx2/DEV9/InternalServers/DHCP_Server.cpp
```diff
@@ -291,7 +291,7 @@ namespace InternalServers
 			//otherwise use adapter's DNS1
 
 			if (dnsIPs.size() >= 1)
-				dns2 = dnsIPs[std::min((size_t)2, dnsIPs.size())];
+				dns2 = dnsIPs[std::min((size_t)2, dnsIPs.size()) - 1];
```
Warepire, line 294: Might be a little more readable like this:
(multi line suggestion doesn't work on unchanged lines, if-statement applies to line above)
```suggestion
			if (!dnsIPs.empty())
				dns2 = dnsIPs[std::min<size_t>(1, dnsIPs.size() - 1)];
```

PR #4454, opened 2021-05-23
Remove PSX references to distinguish between PS1 and DVR

Florin9doi:
### Description of Changes
Remove PSX references to distinguish between PS1 and DVR


karasuhebi:
These are the kind of PRs that I love seeing. 10/10!

PR #4460, opened 2021-05-25
Core: Make DMA's instant during the BIOS

### LINKED ISSUE #4459: [BUG] Certain PS2 BIOS revisions render text incorrectly
**Describe the bug**
The Browser section of the PS2 BIOS does not properly render text on some BIOS revisions. Text is generally supposed to have a strong outline, however that is not the case in PCSX2. It is worth noting only certain BIOS revisions have this issue. 

**To Reproduce**
Boot the BIOS and select Browser. At the memory card screen, observe the text on the top of the screen, it will not have the hard outline it should (screenshot from PCSX2, running my NTSC-U v1.60 BIOS).
image

**Expected behavior**
Text should have a strong outline on it (picture taken from a physical PS2, mind the quality of the screen, from the same NTSC-U v1.60 BIOS system).
image

**GS Settings**
Software renderer, default options. Hardware renderer and its options do not help.

**Emulation Settings**
Preset 2.

Also tried EE/IOP and VU modes; all modes were ineffective.

Also tried gamefixes; all modes were ineffective or crashed the BIOS.

 -   PCSX2 Revision: 1.7.0-dev-1265
 -   OS: Win10
 -   CPU: Ryzen 7 5800X
 -   GPU: GTX 1070 Ti

**Logs and Dumps**
GS dump: [gsdx_20210524134652.zip](https://github.com/PCSX2/pcsx2/files/6534062/gsdx_20210524134652.zip)

refractionpcsx2:
Fixes #4459

This hackfixes the BIOS to avoid a Data Cache bug caused by a DMA buffer being overwritten during a transfer without waiting, which messes up the fonts in the BIOS.  Fixing this correctly would require implementing the Data Cache, which is something which would make the emulator basically unusable, so this is a reasonable solution, it is only enabled during the BIOS screen.

Thanks to the work of @tadanokojin and @PSI-Rockin we can now hackfix this, a proper fix isn't feasible without making the entire emulator slow to a crawl.

### DIFF
--- a/pcsx2/R5900.cpp
+++ b/pcsx2/R5900.cpp
@@ -269,7 +269,7 @@ static __fi void TESTINT( u8 n, void (*callback)() )
 {
 	if( !(cpuRegs.interrupt & (1 << n)) ) return;
 
-	if( cpuTestCycle( cpuRegs.sCycle[n], cpuRegs.eCycle[n] ) )
+	if(!g_GameStarted || cpuTestCycle( cpuRegs.sCycle[n], cpuRegs.eCycle[n] ) )
 	{
 		cpuClearInt( n );
 		callback();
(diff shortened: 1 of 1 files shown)


JordanTheToaster:
Works on EU V2.00 BIOS

1.7 1226
rtjrytkjmtyfkm

PR
erhetjrytjkjj

maxdlpee:
Can confirm that the text renders properly on US 2.00 as well.
Before
https://user-images.githubusercontent.com/77379259/119409035-11e51a00-bcbd-11eb-9e39-6ae88d8c36ea.png
After
pic2

F0bes:
USA v1.60 bios taken from my (39001 or 30001)
- [x] Formatting memory cards
- [x] Copying data to memory cards
- [x] Reading metadata from memory cards
- [x] Deleting data from memory cards
- [x] Clock adjustment
- [x] Changing screen size
- [x] Changing language
- [x] Text is outlined

atomic83GitHub:
The BIOS now throw this log no matter what the cyclerate value is:

image

This doesn't seems to cause issue but it's still an internal error which might be worth examining.

refractionpcsx2:
That's always been there and it probably won't be fixed ever unless sync vu1 happens to sort it (not sure it does).

NewMoonHD:
PR works on USA v2.20.

Dev 1265:
gsdx_20210526064845
gsdx_20210526064242

PR:
gsdx_20210525144728
gsdx_20210526064224

PR #4455, opened 2021-05-23
Eyetoy: Add experimental support for JPEG and multiple resolutions

Florin9doi:
### Description of Changes
Add experimental support for JPEG and multiple resolutions

### Rationale behind Changes
* JPEG support required for EyeToy: Cameo games – JPEG, 640x480 : https://en.wikipedia.org/wiki/Digimask
  * Tested with MLB 07
* Support for Eyetoy Antigrav – MPEG, 160x112 : #4240
  * Won't work with normal webcams because a scaler isn't implemented, but can be tested with OBS

Demo:
Screenshot_2021-05-23_11-05-30

PR #4464, opened 2021-05-26
usb: Fix a few more warnings.

lightningterror:
### Description of Changes

Fix a few warnings.
### Suggested Testing Steps

Compile.

### DIFF
--- a/pcsx2/USB/usb-eyetoy/cam-windows.h
+++ b/pcsx2/USB/usb-eyetoy/cam-windows.h
@@ -72,11 +72,11 @@ namespace usb_eyetoy
 
 		typedef void (*DShowVideoCaptureCallback)(unsigned char* data, int len, int bitsperpixel);
 
-		typedef struct
+		struct buffer_t
 		{
 			void* start = NULL;
 			size_t length = 0;
-		} buffer_t;
+		};
 
 		static const char* APINAME = "DirectShow";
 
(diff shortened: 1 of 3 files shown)


pcsx2/USB/usb-eyetoy/cam-windows.h
```diff
@@ -72,7 +72,7 @@ namespace usb_eyetoy
 
 		typedef void (*DShowVideoCaptureCallback)(unsigned char* data, int len, int bitsperpixel);
 
-		typedef struct
+		typedef struct MPEG_Buffer
```
CookiePLMonster, line 75:
```suggestion
		struct
```
Afaik anonymous structs that are used as a variable type are OK, but the warning comes from `typedef`.

tellowkrinkle: That changes what this does

Define a `struct A` and typedef it to `b`
```cpp
typedef struct A {} b;
```

Define a variable b of type \<anonymous struct\>
```cpp
struct {} b;
```

The other way to fix it would be to use the fact that this is C++ and not C, and do `struct buffer_t {`

Florin9doi:
Excelent job, you resolved 2 useless warnings which were already resolved in #4455 and now you also introduced conflicts to #4455 :neutral_face::neutral_face:

lightningterror:
> Excelent job, you resolved 2 useless warnings which were already resolved in #4455 and now you also introduced conflicts to #4455 😐😐

Oh, didn't know you fixed them too, they were on my radar for a while.

PR #4466, opened 2021-05-28
Fix gamepad choice not affecting rumble test on Linux

Lahvuun:
The rumble test using either the checkbox or the slider in the gamepad
configuration GUI always causes the first gamepad to rumble, even if
other gamepads are selected in the drop-down menu.

### Description of Changes
This change causes the currently selected gamepad to rumble instead of the first one.

### Rationale behind Changes
Rumble test running on the currently selected gamepad is the expected behavior.

### Suggested Testing Steps
- connect multiple gamepads;
- select each one and perform the rumble test;

Without this change the first gamepad will always be the one with rumble. With this change it will be the currently selected gamepad.

### DIFF
--- a/pcsx2/PAD/Linux/wx_dialog/GamepadConfiguration.cpp
+++ b/pcsx2/PAD/Linux/wx_dialog/GamepadConfiguration.cpp
@@ -114,7 +114,7 @@ void GamepadConfiguration::OnSliderReleased(wxCommandEvent& event)
 
 		// convert in a float value between 0 and 1, and run rumble feedback.
 		// 0 to 1 scales to 0x0 to 0x7FFF
-		s_vgamePad[m_pad_id]->TestForce(m_sl_rumble_intensity->GetValue() / (float)0x7FFF);
+		s_vgamePad[m_joy_map->GetSelection()]->TestForce(m_sl_rumble_intensity->GetValue() / (float)0x7FFF);
 	}
 	else if (sl_id == joy_slider_id)
 	{
@@ -135,7 +135,7 @@ void GamepadConfiguration::OnCheckboxChange(wxCommandEvent& event)
 		g_conf.pad_options[m_pad_id].forcefeedback = (m_cb_rumble->GetValue()) ? (u32)1 : (u32)0;
 		if (m_cb_rumble->GetValue())
 		{
-			s_vgamePad[m_pad_id]->TestForce();
+			s_vgamePad[m_joy_map->GetSelection()]->TestForce();
 			m_sl_rumble_intensity->Enable();
 		}
 		else


kenshen112:
Quick word of advice. If you go to GamepadConfiguration.cpp:152 `OnChoiceChange` change the set_joy_uid to `		g_conf.set_joy_uid(m_pad_id, GamePad::index_to_uid(m_joy_map->GetSelection()));
` the same changes you made to the rumble and you'll fix the entire selection box so people can change and select which controller is active.

pcsx2/PAD/Linux/wx_dialog/GamepadConfiguration.cpp
```diff
@@ -114,7 +114,7 @@ void GamepadConfiguration::OnSliderReleased(wxCommandEvent& event)
 
 		// convert in a float value between 0 and 1, and run rumble feedback.
 		// 0 to 1 scales to 0x0 to 0x7FFF
-		s_vgamePad[m_pad_id]->TestForce(m_sl_rumble_intensity->GetValue() / (float)0x7FFF);
```
kenshen112, line 117: It would make more sense to change m_pad_id to be assigned `m_joy_map->GetSelection();` rather than removing it's use here. Or drop m_pad_id entirely.

pcsx2/PAD/Linux/wx_dialog/GamepadConfiguration.cpp
```diff
@@ -135,7 +135,7 @@ void GamepadConfiguration::OnCheckboxChange(wxCommandEvent& event)
 		g_conf.pad_options[m_pad_id].forcefeedback = (m_cb_rumble->GetValue()) ? (u32)1 : (u32)0;
 		if (m_cb_rumble->GetValue())
 		{
-			s_vgamePad[m_pad_id]->TestForce();
+			s_vgamePad[m_joy_map->GetSelection()]->TestForce();
 			m_sl_rumble_intensity->Enable();
 		}
 		else
```
kenshen112, line 141: Note that below this in OnChoice Change ID is looking at a selection box that doesn't exist. It would be best to update it to use m_joy_map like you've done above so the selection box works as well

kenshen112 (changes requested):

pcsx2/PAD/Linux/wx_dialog/GamepadConfiguration.cpp
Lahvuun: > It would make more sense to change m_pad_id to be assigned `m_joy_map->GetSelection();`

I think m_pad_id is not meant to be changed.
From what I understand, m_pad_id corresponds to a PS2 gamepad slot, so it can only be 0 or 1. The value is set when the window is created, based on which gamepad tab you have selected:
image
This id (in the square brackets) is then used to update the selected gamepad in inis/PAD.ini: 
```ini
uid[0] = 5335467880578043658
uid[1] = 7874879626751023234
```

Please correct me if I'm wrong.

Lahvuun: @kenshen112 So, really the bug here: `s_vgamePad[m_pad_id]->TestForce(m_sl_rumble_intensity->GetValue() / (float)0x7FFF);` is that out of the vector of detected gamepads (s_vgamePad), it will always be the first or the second gamepad to vibrate, depending on which tab is selected.

pcsx2/PAD/Linux/wx_dialog/GamepadConfiguration.cpp
Lahvuun: I'm not sure I understand what you mean. For me the selection box works correctly.
Could you list the steps to get it to break?

kenshen112: That was fro a specific adaper and is more of a hack nvm on that one

pcsx2/PAD/Linux/wx_dialog/GamepadConfiguration.cpp
kenshen112: Ye kinda. Though m_pad_id is supposed to be the selected pad in general being that how it's currently updated or set is wrong. It would be more correct to have m_pad_id updated with the selected pad not the tab

Lahvuun: > m_pad_id is supposed to be the selected pad in general

What makes you think that? Because I think it's not.
If it were, then the widget would not know which PS2 pad to save the settings under in inis/PAD.ini, 0 or 1. 
m_pad_id basically only exists to do this, and is set only when the widget is created. 

The *selected* pad id is not stored, so it has to be retrieved with m_joy_map->GetSelection() for things like the force feedback test.

I don't see why anything would need to be changed here (unless I'm misunderstanding something about how the Linux pad GUI works)

kenshen112: Mostly from the original usage. In TestForce, and that it's set to "pad" in the constructor. Mind you that's the tab but it was at least originally intended to be the pad index

turtleli:
Thanks. Sorry for the delay.

PR #4468, opened 2021-05-29
ipc: rename to pine

GovanifY:
This is a long coming change related to our ipc implementation: as the protocol is slowly getting adopted by other projects/other contributors stepped in the name had to stop being related to pcsx2 specifically. 

PINE (Protocol for the Instrumentation of Emulators) was chosen, this patch reflects that.

### DIFF
--- a/pcsx2/CMakeLists.txt
+++ b/pcsx2/CMakeLists.txt
@@ -130,7 +130,7 @@ set(pcsx2Sources
 	IopIrq.cpp
 	IopMem.cpp
 	IopSio2.cpp
-	IPC.cpp
+	PINE.cpp
 	Mdec.cpp
 	Memory.cpp
 	MemoryCardFile.cpp
@@ -205,7 +205,7 @@ set(pcsx2Headers
 	IopHw.h
 	IopMem.h
 	IopSio2.h
-	IPC.h
+	PINE.h
 	Mdec.h
 	MTVU.h
 	Memory.h
@@ -908,7 +908,7 @@ set(pcsx2GuiSources
 	gui/Dialogs/McdConfigDialog.cpp
 	gui/Dialogs/PickUserModeDialog.cpp
 	gui/Dialogs/SysConfigDialog.cpp
-	gui/Dialogs/IPCDialog.cpp
+	gui/Dialogs/PINEDialog.cpp
 	gui/Debugger/BreakpointWindow.cpp
 	gui/Debugger/CtrlDisassemblyView.cpp
 	gui/Debugger/CtrlRegisterList.cpp
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -545,7 +545,7 @@ struct Pcsx2Config
 		CdvdShareWrite : 1, // allows the iso to be modified while it's loaded
 		EnablePatches : 1, // enables patch detection and application
 		EnableCheats : 1, // enables cheat detection and application
-		EnableIPC : 1, // enables inter-process communication
+		EnablePINE : 1, // enables inter-process communication
 		EnableWideScreenPatches : 1,
 #ifndef DISABLE_RECORDING
 		EnableRecordingTools : 1,
--- a/pcsx2/PINE.cpp
+++ b/pcsx2/PINE.cpp
@@ -40,10 +40,10 @@
 #include "gui/AppCoreThread.h"
 #include "System/SysThreads.h"
 #include "svnrev.h"
-#include "IPC.h"
+#include "PINE.h"
 
-SocketIPC::SocketIPC(SysCoreThread* vm, unsigned int slot)
-	: pxThread("IPC_Socket")
+PINEServer::PINEServer(SysCoreThread* vm, unsigned int slot)
+	: pxThread("PINE_Server")
 {
 #ifdef _WIN32
 	WSADATA wsa;
@@ -52,14 +52,14 @@ SocketIPC::SocketIPC(SysCoreThread* vm, unsigned int slot)
 
 	if (WSAStartup(MAKEWORD(2, 2), &wsa) != 0)
 	{
-		Console.WriteLn(Color_Red, "IPC: Cannot initialize winsock! Shutting down...");
+		Console.WriteLn(Color_Red, "PINE: Cannot initialize winsock! Shutting down...");
 		return;
 	}
 
 	m_sock = socket(AF_INET, SOCK_STREAM, 0);
 	if ((m_sock == INVALID_SOCKET) || slot > 65536)
 	{
-		Console.WriteLn(Color_Red, "IPC: Cannot open socket! Shutting down...");
+		Console.WriteLn(Color_Red, "PINE: Cannot open socket! Shutting down...");
 		return;
 	}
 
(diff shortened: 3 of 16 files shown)


tadanokojin:
You're welcome to call it whatever you like internally but we don't do branding anymore.

That's why SPU2-X is now just SPU.
It's why Lilypad and Onepad are now just Pad.
It's why GSdx will just be GS when it's done.
It's why when you tried to brand the new GSdump interface we had you change that too.

I don't see any reason to make an exception here.

GovanifY:
> You're welcome to call it whatever you like internally but we don't do branding anymore.
> 
> That's why SPU2-X is now just SPU.
> It's why Lilypad and Onepad are now just Pad.
> It's why GSdx will just be GS when it's done.
> It's why when you tried to brand the new GSdump interface we had you change that too.
> 
> I don't see any reason to make an exception here.

The exception is that this is a cross protocol emulator that is going to be supported outside of PCSX2. It needs to share a common name with other implementations, I won't call it "pcsx2 ipc" in rpcs3, when it has a shared interface.
Also that excuse seems fallacious, otherwise why would we keep, eg, pnach? Removing plugins and renaming a component that is shared across different projects seems very different to me, it's like if I renamed a library.

tadanokojin:
> Also that excuse seems fallacious, otherwise why would we keep, eg, pnach? 

I really dislike arguments like this. The point is to change all that, we haven't gotten to pnach yet. Or microVU for that matter. But just because the work isn't complete doesn't make it fallacious to have a goal.

Example: GSdx Hacks
If I start removing hacks and say to new hacks "We don't do hacks anymore" and someone responds "but what about these other hacks then?". Yes the goal is to remove them, by making this argument you essentially tell me the only valid way to remove them is to do it all at once. This is an unrealistic expectation. It's not fallacious to say "we don't want to go in this direction anymore".

> The exception is that this is a cross protocol emulator that is going to be supported outside of PCSX2. 

I don't find that to be a good reason. What about the fact its being used outside PCSX2 means you have to rename it in the code? Can't you just refer to it as that if you like the name? That seems fair enough to me.

> I won't call it "pcsx2 ipc" in rpcs3, when it has a shared interface.

Well the don't, just call it the IPC which is how it's referred to in the code. Or call it PINE or whatever you wanna call it, I'm not bothered what you call it that was kind of the point I was making.

> it's like if I renamed a library.

Well you're welcome to make it a library and call it whatever you want. That's not a reality reflected by the code currently however and I would still prefer the files that interface with it in PCSX2 refer to it as the IPC since that is much clearer as to it's function.

GovanifY:
> > Also that excuse seems fallacious, otherwise why would we keep, eg, pnach?
> 
> I really dislike arguments like this. The point is to change all that, we haven't gotten to pnach yet. Or microVU for that matter. But just because the work isn't complete doesn't make it fallacious to have a goal.

This was never a part of the work, this was never discussed, either in staff channels, in github, or personally. The reason we removed branding was to unify settings and to make it easier for users, because they are a _part of pcsx2_. PINE is a protocol used outside of PCSX2, just like pulseaudio is a protocol used outside of pcsx2.


> Example: GSdx Hacks
> If I start removing hacks and say to new hacks "We don't do hacks anymore" and someone responds "but what about these other hacks then?". Yes the goal is to remove them, by making this argument you essentially tell me the only valid way to remove them is to do it all at once. This is an unrealistic expectation. It's not fallacious to say "we don't want to go in this direction anymore".

Again, this argument is fallacious, no intent was discussed to remove branding beyond plugins, for the sake of clarity.

> 
> > The exception is that this is a cross protocol emulator that is going to be supported outside of PCSX2.
> 
> I don't find that to be a good reason. What about the fact its being used outside PCSX2 means you have to rename it in the code? Can't you just refer to it as that if you like the name? That seems fair enough to me.

This is supposed to be a take at a compatible, cross platform _protocol_. Calling it "IPC" is vague at best and doesn't actually explain its purpose nor its inner workings. Do we call HTTP "the www ipc protocol"?


> 
> > I won't call it "pcsx2 ipc" in rpcs3, when it has a shared interface.
> 
> Well the don't, just call it the IPC which is how it's referred to in the code. 
It's not. I switched this PR to use IPC to only note message serialization and encoding. PINE is a type of IPC protocol.

>Or call it PINE or whatever you wanna call it, I'm not bothered what you call it that was kind of the point I was making.

I'm not sure I follow

> 
> > it's like if I renamed a library.
> 
> Well you're welcome to make it a library 

The client is already a library. The server, either in pcsx2 or rpcs3, isn't because it requires some level of integration.

>and call it whatever you want. That's not a reality reflected by the code currently however and I would still prefer the files that interface with it in PCSX2 refer to it as the IPC since that is much clearer as to it's function.

PINE is an IPC protocol. I think it is rather clear what its function is when you bother to read through the first lines of comments, linking you to the client implementation. If you prefer another take at it then please do so.

Also I'd like that you would be more open to discussion regarding topics like this, this change has been discussed for months in our staff channels and never once did you say anything about it, you only decided to come in once the work was done. The discussions we had were

tadanokojin:
> This was never a part of the work, this was never discussed, either in staff channels, in github, or personally. 

It has been, though. The thing is I don't expect you to be privy to any particular discussion on Discord so I'm not down your throat about it. I'm giving you the information when I had the opportunity to give it to you.

> Again, this argument is fallacious, no intent was discussed to remove branding beyond plugins, for the sake of clarity.

Sorry if you feel like this wasn't communicated with you but that doesn't make it fallacious. Getting information you didn't have previously and needing to adjust your view accordingly is not how fallacies work.

> This is supposed to be a take at a compatible, cross platform protocol. Calling it "IPC" is vague at best and doesn't actually explain its purpose nor its inner workings. Do we call HTTP "the www ipc protocol"?

I don't believe this to be an equivalence. When you say "IPC" with relation to an emulator, most people are going to know what those files and interfaces are for. It doesn't need any further elaboration via a new acronym. What does "Protocol for Instrumentalization of Emulators" add that isn't already obvious by the fact that it's IPC being used in an emulator? Are you expecting people will think it's a protocol for the instrumentation of teapots?

> I'm not sure I follow

I'm saying you're welcome to call it PINE or whatever outside of the code. I don't really care. I don't see how your decision to call it something means we have to change the names of the files and the interface.

> PINE is an IPC protocol. I think it is rather clear what its function is when you bother to read through the first lines of comments, linking you to the client implementation. If you prefer another take at it then please tell me.

Okay but it would be kind of clear what "IFTR" means if I put it somewhere that it means "the interface for the renderer". But why give it a less clear name? Why not just call it RendererInterface because that's what it is. You don't need to look it up. You don't need to look up IPC, most people already know what that means and the name PINE isn't giving anymore clarity as I pointed out. It's doing the opposite.

> Also I'd like that you would be more open to discussion regarding topics like this, this change has been discussed for months in our staff channels and never once did you say anything about it, you only decided to come in once the work was done. The discussions we had were open to every staff member and you could have voiced your concerns then.

I'm not sure how you expect me to be "more open" to discussion about it. I'm discussing it with you now. If I was less open to it I would close this and tell you we're not doing it but that it a little too closed minded for me. I have given you the opportunity via this PR to disagree with me and to make your argument and I'm sorry but I just don't find it convincing.

I'm sorry if you or I miss some conversation but that's the nature of discord. You can't expect me to be around for any conversation about any particular thing. I did not see any conversation about renaming this if I had I would have brought it up then.

GovanifY:
This matter was discussed privately. It was agree to refer internally to it as it used to be (eg IPCSocket) and refer to it in the ui as the PINE Protocol/PINE IPC

GovanifY:
I've added to this PR support for VLE encoding of replies with runtime relocation or, when possible, ahead-of-time detection with no additional cost compared to the previous implementation. This adds an O(n) loop, which basically doesn't affect performance even in the ns realm afaict.

A fix for the game serial being sent with [ ] was also added.

GovanifY:
finally rebased this PR; it's been tested and known to work, we might want to merge this as the client implementation relies on VLE replies which I commited to this PR a few months ago, but somehow never made it into master

PR #4475, opened 2021-06-02
IOP: Run the SPU at a consistent rate

Ziemas:
I guess this is more of an RFC than a PR, but having this code here doesn't make much sense to me so let's figure out if it's still useful.

```
Run the SPU once per 768th IOP cycle. And stop adjusting the SPU's
target cycle in IopDma, doing this makes little sense the SPU2 is so eager
about running itself on interactions from the core.

If we need finer timing for DMA transfers we could just run the SPU2
more often than every 765th cycle, it should handle that fine as far as
I can tell.
```

### DIFF
--- a/pcsx2/IopCounters.cpp
+++ b/pcsx2/IopCounters.cpp
@@ -542,6 +542,8 @@ void psxRcntUpdate()
 		psxCounters[CNT_SPU_DMA].sCycleT = psxRegs.cycle;
 		psxCounters[CNT_SPU_DMA].CycleT = psxCounters[CNT_SPU_DMA].rate;
 		SPU2RunDma(difference);
+		// The spu might have updateed the next cycle target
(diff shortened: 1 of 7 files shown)


refractionpcsx2:
There's a bit of a problem with running every 768 cycles, that french game Titeuf runs slow and stutters internally if you sync every 768 cycles, * 12 is fine, any less and it gets weird.

I had changed this back to 768 cycles earlier until I discovered it caused problems and I changed it back again.

Ziemas:
Ok, good to know. I'll check that out myself and see if I can figure anything out.

gilderoylockhart01:

refractionpcsx2:
I believe last it was touched Titeuf Mega-compet still had the stuttering issue with this.

lightningterror:
Closing as it's not worked on atm.

PR #4476, opened 2021-06-02
Add PString

kenshen112:
### Description of Changes

This PR adds a new common string lib PString that can currently handle UTF8 and UTF16 types

### Rationale behind Changes

To enable easier conversion and handling of all string types through the use of a common type handler

### Suggested Testing Steps

Attempt use of string and wxString types to test string conversions and containment.


kenshen112:
Note I've experimented with use between const char* and basic string types here

basic_string: https://github.com/kenshen112/pcsx2-string/commit/fe97d84ea02466cd1a87c6c974d86315ab389382
const char*: https://github.com/kenshen112/pcsx2-string/commit/c13034eeacf9966df19b03cc56d111cc6a04c890

kenshen112:
Currently tested functionality: https://hatebin.com/vswpjetubl

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,157 @@
...
+	count = std::move(rhs.count);
+	string = std::move(rhs.string);
+}
+
+#ifdef _WIN32
+// Non default
+PString::PString(const std::wstring& str)
+{
+	count = str.size();
+	char* temp = new char[count];
```
Warepire, line 48: temp is leaking. Use a memory managed object like an std::string with a previous resize, like the internal string.

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,157 @@
...
+	const wchar_t* wStr = str.c_str();
+	size_t len = (wcslen(wStr) + 1) * sizeof(wchar_t);
+	wcstombs_s(&count, temp, len, wStr, len);
+	if (temp != nullptr)
```
Warepire, line 52: nullcheck too late? May try to use nullptr as input to wcstombs_s. Also, error code of wcstombs_s is not checked.

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,157 @@
...
+	{
+		string = temp;
+	}
+}
+
+PString& PString::operator=(const std::wstring& str)
+{
+	count = str.size();
+	char* temp = new char[count];
```
Warepire, line 61: Leaking again, in fact, all your allocations are leaking.

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+std::istream& getline(std::istream& is, PString& str, char delim)
+{
+	char* streambuf = new char[5000];
+	is.get(streambuf, 5000, delim);
+	str.string = streambuf;
+	return is;
+}
+
+PString::~PString()
```
Warepire, line 153: `PString::~PString() = default;` All data members have defined destructors, no need to "clean up" like this.

common/include/Utilities/PString.h
```diff
@@ -0,0 +1,95 @@
...
+#include <iostream>
+#include <stdio.h>
+#include <stdlib.h>
+#include <locale.h>
+#include <string>
+#include "Path.h"
+
+namespace fs = ghc::filesystem;
+
+class PString
```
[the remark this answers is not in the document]
kenshen112: PcsxString

common/include/Utilities/PString.h
```diff
@@ -0,0 +1,121 @@
...
+{
+private:
+	std::basic_string<char> string;
+public:
+	PString() = default;
+
+	// Non default constructors
+	PString(const char* string)
```
CookiePLMonster, line 21: Prefer using initializer lists.

common/include/Utilities/PString.h
```diff
@@ -0,0 +1,121 @@
...
+
+	// Copy Constructor
+	PString(const PString& rhs)
+	{
+		string = rhs.string;
+	}
+
+	PString(PString&& move)
+	{
+		string = move.string;
```
CookiePLMonster, line 54: This doesn't actually move, but copy. For move and copy constructors you might be able to use `= default;`.

common/include/Utilities/PString.h
```diff
@@ -0,0 +1,121 @@
...
+		return string.capacity();
+	}
+
+	size_t size() const noexcept
+	{
+		return string.size();
+	}
+
+	char& at(size_t pos);
+	char& at(size_t pos) const;
```
CookiePLMonster, line 82: prefer returning `const char&` or `char` for const overloads.

common/include/Utilities/PString.h
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	PString& operator=(PString&&);
+
+	// Copy Operator
+	const PString& operator=(const PString&);
+	
+	operator std::string() const;
+	operator wxString() const;
+	operator fs::path() const;
```
CookiePLMonster, line 69: These look quite scary, it might be better not to allow for implicit conversions to force the programmer to think twice about character conversions.

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,95 @@
...
+{
+	string = std::move(rhs.string);
+	return *this;
+}
+
+#ifdef _WIN32
+// Non default
+PString::PString(const std::wstring& str)
+{
+	char* temp = new char[count];
```
CookiePLMonster, line 22: `std::string` has a non-const `data()` overload now, so this can be done in-place with a properly done `.resize`.

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,95 @@
...
+	const wchar_t* wStr = str.c_str();
+	size_t len = (wcslen(wStr) + 1) * sizeof(wchar_t);
+	int err = wcstombs_s(&count, temp, len, wStr, len);
```
CookiePLMonster, line 25: `wcstombs` might not necessarily output the multibyte character encoding you want. If you want to ensure UTF-8, prefer using `WideCharToMultiByte` since it's a Windows only function anyway.

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,95 @@
...
+		delete temp;
+	}
+
+	string = temp;
+	delete temp;
+}
+
+PString::operator std::wstring()
+{
+	wchar_t* wstr = new wchar_t[count];
```
CookiePLMonster, line 37: Prefer an in-place construction using a non-const `std::wstring.data()`.

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,95 @@
...
+PString::operator fs::path() const
+{
+	fs::path str(string);
+	return str;
+}
+
+void PString::resize(size_t siz, char c)
+{
+	string.resize(siz);
+	string = c;
```
CookiePLMonster, line 72: This will be extremely confusing, is there any particular reason you opted for a different behaviour to `std::string::resize(siz, c);`?

CookiePLMonster (changes requested):

CookiePLMonster:
Consider adding a `PString(std::string str)` constructor that *moves* `str` to the class. Also consider replacing a `const char*` constructor with `std::string_view`, as it can do everything a plain char pointer constructor could do and more (gives the user an option to construct a string with a known length).

common/include/Utilities/PString.h
```diff
@@ -0,0 +1,121 @@
...
+
+	// Non default constructors
+	PString(const char* string)
+	{
+		string = string;
+	}
+
+	explicit PString(const fs::path& path)
+	{
+		string = path.string();
```
CookiePLMonster, line 28: Prefer `path.u8string();` to explicitly request a UTF-8 string.

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+PString::operator std::string() const
+{
+	std::string str(string);
+	return str;
+}
+
+PString::operator fs::path() const
+{
+	fs::path str(string);
```
CookiePLMonster, line 65: Prefer `u8path(string)` to explicitly treat the string as a UTF-8 string.

CookiePLMonster (changes requested):
Looking at all these review remarks, I believe this PR could work better as a set of conversion functions instead of a string wrapper, like:
* `ToWxString`
* `FromWxString`
* `ToWString`
* `FromWString`

common/include/Utilities/PString.h
kenshen112: Well most of them are using the inbuilt conversions or constructors to copy the string data over so it should be fine

common/src/Utilities/PString.cpp
kenshen112: I was having issues with data() and wcstombs in this functions it didn't seem to want to fill it in proper

kenshen112:
> Looking at all these review remarks, I believe this PR could work better as a set of conversion functions instead of a string wrapper, like:
> 
>     * `ToWxString`
> 
>     * `FromWxString`
> 
>     * `ToWString`
> 
>     * `FromWString`

The idea behind this PR was to make a full assed string class rather than a plain set of functions to make copying data or even just using it as a string itself easier / possible. Originally I had a char* base and was doing the memory management myself. That turned into this.

common/src/Utilities/PString.cpp
CookiePLMonster: You shouldn't use `wcstombs` in the first place, maybe that was the issue you encountered in the first place.

CookiePLMonster:
> The idea behind this PR was to make a full assed string class rather than a plain set of functions to make copying data or even just using it as a string itself easier / possible.

The fact you ended up with essentially a thin wrapper around `std::string` makes me think the thing you really need is a set of well defined conversion functions operating on `string/wstring/wxString`.

common/include/Utilities/PString.h
```diff
@@ -0,0 +1,121 @@
...
+#include "ghc/filesystem.h"
+#include <iostream>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <locale.h>
+#include <string>
+#include "Path.h"
+
+namespace fs = ghc::filesystem;
```
CookiePLMonster, line 11: This will make every file including `PString.h` use this alias, not a good idea. Probably better to move this alias to inside the class.

tellowkrinkle:
> The fact you ended up with essentially a thin wrapper around `std::string` makes me think the thing you really need is a set of well defined conversion functions operating on `string/wstring/wxString`.

I think the main thing I would want is a string type where the conversions to other types that most people would expect to be able use don't do anything dumb.  We currently use "UTF-8 in a `std::string`" and it has the following issues:

1. `std::string` ↔︎ `fs::path` currently works fine, but if we ever switch to `std::filesystem::path`, it will assume all `std::string`s are not UTF-8.  Since `std::string` is not our type, we have no say in this
2. `std::string` ↔︎ `wxString` is an implicit conversion and once again assumes `std::string`s are not UTF-8.  We have some control over this one, since we currently bundle our own copy of WX and could edit it, but it's still kind of messy
3. Other stdlib functions (like `std::iostream`) assume `std::string` is not UTF-8.  Less of an issue because we will hopefully be using `fs::path`s with them in the future, but the temptation will continue to exist.

In fact there already are well defined conversion functions operating on `wxString`.  There's `wxString::FromUTF8()` and `wxString::utf8_string()` built into wx.  But clearly the implicit conversions are just too easy to accidentally use, because we still have plenty of places where we use them.  We need to stop using types whose encoding isn't agreed on by everyone.  We could use `std::u8string` in C++20, but personally, I like having implicit conversions to other string types when possible, and it has zero.  Won't even have explicit conversions, we'll need helper methods for literally everything.  Also we're not on C++20 yet.

As for conversions, my preferences:
- `wxString`, `QString`: Implicit.  No chances of data loss or corruption in this conversion
- `std::u8string`: Implicit.  Like above, no chances of data loss or corruption in this conversion
- `std::wstring`: Implicit, Windows only.  No chance of data loss, but no one uses wstring/wchar_t outside of Windows.
- `std::string`: Not sure.  The other conversions will prevent (1) and (2) above leaving just (3).  I expect a lot of our interaction with `std::string` will actually be in file IO (e.g. yaml reading, etc), which will actually be UTF-8.  For now I'd go explicit and see how annoying that gets.  Add a `.u8()` method too, for easier conversion at least in one direction.  We'll probably also want a `.mb()` method to convert to platform encoding, for libraries that were designed for unix and don't support wchars (like chd).
- `fs::path`: Explicit.  Paths and strings should be treated separately.

Not sure how I feel about the `const char*` constructor for the same reason.  Maybe if you add a custom string literal operator we can continue to use string literals without having to have an implicit conversion from `const char*`

CookiePLMonster:
> As for conversions, my preferences:

While I generally don't disagree with these preferences, both conversions to wx/Qt types and `wstring` should be as rare as possible and so I don't see the need to make them implicit. Implicit conversion also has a potential side effect of the programmer unknowingly converting the strings more than it is needed.



> Not sure how I feel about the `const char*` constructor for the same reason.

I think the way wx/Qt solves this issue (static creation methods) is quite sensible - `PcsxString::FromUTF8(char_array);` is very consise and obvious, `PcsxString(char_array);` requires the programmer to be aware of UTF-8 behaviour.



> 1. `std::string` ↔︎ `fs::path` currently works fine, but if we ever switch to `std::filesystem::path`, it will assume all `std::string`s are not UTF-8.

I thought `ghc::filesystem` is supposed to be a drop-in replacement for `std::filesystem`? That is kind of worrying then.

---------------------------------------------------------------

I think I understand the need for a string type better now, then. In this case I believe it might be best to make this type as much of a thin wrapper as possible, and maybe expose a `const std::string& str() const { return m_stdString; }` method instead of forwarding half of the API to it? I think it might be a good idea to make such a type immutable, i.e. narrow it down to a set of constructors, conversion functions and a read-only UTF-8 getter.

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+	const int size = WideCharToMultiByte(CP_UTF8, 0, utf16_string.c_str(), utf16_string.size(), nullptr, 0, nullptr, nullptr);
+	std::string converted_string(size, 0);
+	WideCharToMultiByte(CP_UTF8, 0, utf16_string.c_str(), utf16_string.size(), converted_string.data(), converted_string.size(), nullptr, nullptr);
+	string = converted_string;
+}
+
+PString::operator std::wstring()
+{
+	int size = MultiByteToWideChar(CP_UTF8, 0, string.c_str(), -1, nullptr, 0);
```
CookiePLMonster, line 30: You have access to `string.size()` here, no need to rely on the null terminator.

gulrak:
> > 1. `std::string` ↔︎ `fs::path` currently works fine, but if we ever switch to `std::filesystem::path`, it will assume all `std::string`s are not UTF-8.
> 
> I thought `ghc::filesystem` is supposed to be a drop-in replacement for `std::filesystem`? That is kind of worrying then.

Just to clarify this: `ghc::filesystem` is developed as a drop-in replacement of `std::filesystem` and starting with 1.5.x, where the Windows backend is using `std::wstring` it came as close at it might get to that target. Still, it is an utf8-everywhere philosophy approach and I say so in the docs. What that means in respect to `std::filesystem` from C++17 is, that in `std::filesystem` std::string interfaces are the expected to be in the current locale and if that is an UTF-8 type of locale than both behave the same. If instead the current locale encoding is non-UTF-8 based then the behavior is different and only the explicit interfaces (`fs::u8path()` and `fs::path::u8string()`) are exactly behaving the same. `ghc::filesystem` contains no locale handling that is able to handle anything else than UTF-8 for `std::string` (besides 7-bit clean ASCII of course, as that is a subset).

All this is made to make Windows/Linux/macOS portable development as easy as possible and works in multiple projects with mixed `std::filesystem`/`ghc::filesystem` depending on the OS and compiler available on the target (e.g. all Win-API calls should be the W-Version and simply use `p.c_str()` _not `p.wstring().cstr()`_ as the parameter to ensure that even non-utf-8 supporting Windows-Version versions work well with international characters) and as long as you stick to UTF-8-locales or use the explicit interfaces into and from `fs::path` everything works the same when switching to `std::filesystem` (besides the differences between the existing `std::filesystem` implementations that is ;-)).

I admit that this approach might not be perfect or fitting for every project. I just want to make clear where the differences are and I'm sorry, if that might be an issue and was not clear enough from the readme.

CookiePLMonster:
> I admit that this approach might not be perfect or fitting for every project. I just want to make clear where the differences are and I'm sorry, if that might be an issue and was not clear enough from the readme.

From what you say `ghc::filesystem` is behaving as expected regardless, and any assumptions about `fs::path` being UTF-8 at all times (without using `u8path`) are dangerous as they're relying on an implementation detail.

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,108 @@
...
+	std::string converted_string(size, 0);
+	WideCharToMultiByte(CP_UTF8, 0, utf16_string.c_str(), utf16_string.size(), converted_string.data(), converted_string.size(), nullptr, nullptr);
+	string = converted_string;
+}
+
+PString::operator std::wstring()
+{
+	const int size = MultiByteToWideChar(CP_UTF8, 0, string.c_str(), string.size(), nullptr, 0);
+	std::vector<wchar_t> converted_string(size);
+	MultiByteToWideChar(CP_UTF8, 0, string.c_str(), -1, converted_string.data(), converted_string.size());
```
CookiePLMonster, line 42: This still relies on a null terminator (`-1` size).

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,108 @@
...
+	string.resize(siz, c);
+}
+
+std::ostream& operator<<(std::ostream& os, const PString& str)
+{
+	os << str.string;
+	return os;
+}
+
+std::string PString::u8()
```
CookiePLMonster, line 76:
```suggestion
const std::string& PString::u8() const
```

If the caller wants a copy, they can do it themselves.

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,108 @@
...
+{
+	return string;
+}
+
+std::string PString::mb()
```
CookiePLMonster, line 81:
```suggestion
std::string PString::mb() const
```

common/src/Utilities/PString.cpp
kenshen112: So, every time I've tried to call u8Path it gives me multiple definitions / not sure which u8Path is being called error. I've tried calling it like the above suggestion and calling it like

`u8path(string.begin(), string.end());` No luck

kenshen112: Tellow helped me figure it out

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,80 @@
+#include "PrecompiledHeader.h"
```
Mrlinkwii, line 1: you're missing license  header :slightly_smiling_face:

Mrlinkwii:
you seem to be updating libchdr  , that that intended  :confused: ?

tadanokojin:
Not an upgrade, downgrade. He needed to run `git submodule update` and instead committed the change.

common/include/Utilities/PString.h
```diff
@@ -0,0 +1,131 @@
...
+	// Return OS specific format
+	std::string mb() const;
+
+	// Return UTF8 explicit string
+	std::string u8() const
+	{
+		return string;
+	}
+
+	const bool operator==(const PString& rhs)
```
CookiePLMonster, line 63:
```suggestion
	const bool operator==(const PString& rhs) const
```

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+
+#ifdef _WIN32
+// Non default
+PString::PString(const std::wstring& utf16_string)
+{
+	const int size = WideCharToMultiByte(CP_UTF8, 0, utf16_string.c_str(), utf16_string.size(), nullptr, 0, nullptr, nullptr);
+	std::string converted_string(size, 0);
+	WideCharToMultiByte(CP_UTF8, 0, utf16_string.c_str(), utf16_string.size(), converted_string.data(), converted_string.size(), nullptr, nullptr);
+	string = converted_string;
```
CookiePLMonster, line 22: Any reason it's not being converted in place? `converted_string` is redundant.

common/src/Utilities/PString.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+PString::operator std::wstring()
+{
+	const int size = MultiByteToWideChar(CP_UTF8, 0, string.c_str(), string.size(), nullptr, 0);
+	std::vector<wchar_t> converted_string(size);
+	MultiByteToWideChar(CP_UTF8, 0, string.c_str(), string.size(), converted_string.data(), converted_string.size());
+	return converted_string.data();
```
CookiePLMonster, line 30: Any reason this isn't converted to `std::wstring` directly like the constructor above uses `std::string`?

common/include/Utilities/PString.h
```diff
@@ -0,0 +1,137 @@
+#pragma once
```
lightningterror, line 1: Missing copyright header here as well.

tellowkrinkle:
After discussion with other team members, we finally decided that the best solution to the problems this PR solves would be to stay away from `std::filesystem` and continue to use `ghc::filesystem` for the foreseeable future.

Maybe in the future we can migrate the whole codebase to `std::u8string`, but I wouldn't consider that high priority (especially since C++20 has to come first).

PR #4489, opened 2021-06-09
Debugger: Modify hotkeys and add delete breakpoint context action

F0bes:
### Description of Changes

- Disassembly view:  hotkeys don't need to be used with CTRL (they are still usable with CTRL, however)
- Memory view: Goto address hotkey (G) doesn't need to be used with CTRL
- Disassembly view: To reassemble an opcode, you  can press `M` or go through the context menu, pressing any non-hotkey letter won't trigger this anymore.
- You can now delete a breakpoint through its context menu. (See image)
image

### Rationale behind Changes
If you use Ghidra for example, you'd know that depending on CTRL for goto is odd.

Why not have delete as a context menu option for breakpoints?

### Suggested Testing Steps
Use the various debugger hotkeys, see if they work.

Make sure the delete breakpoint button works properly.

## Note
~~The debugger pdf is outdated. Sorry, but why use pdfs when they are probably the worst format to edit?~~

### DIFF
--- a/pcsx2/Docs/Debugger.md
+++ b/pcsx2/Docs/Debugger.md
@@ -8,10 +8,11 @@ urlcolor: "cyan"
 # Debugger Key Bindings
 
 ## Disassembly View
--   `Ctrl+G`   -   goto
--   `Ctrl+E`   -   edit breakpoint
--   `Ctrl+D`   -   enable/disable breakpoint
--   `Ctrl+B`   -   add breakpoint
+-   `G`   -   goto
+-   `E`   -   edit breakpoint
+-   `D`   -   enable/disable breakpoint
+-   `B`   -   add breakpoint
+-   `M`   -   assemble opcode
 -   `Right Arrow`   -   follow branch/position memory view to accessed address
 -   `Left Arrow`   -   go back one branch level/goto pc
 -   `Up Arrow`   -   move cursor up one line
@@ -26,7 +27,7 @@ urlcolor: "cyan"
 
 ## Memory View
 
--   `Ctrl+G`   -   goto
+-   `G`   -   goto
 -   `Ctrl+B`   -   add breakpoint
 -   `Left Arrow`   -   move cursor back one byte/nibble
 -   `Right Arrow`   -   move cursor ahead one byte/nibble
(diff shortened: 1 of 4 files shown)


kozarovv:
> Disassembly view: To reassemble an opcode, you have to go through the context menu, pressing any non-hotkey letter won't trigger this anymore.

Maybe will be good idea to add new hotkey for that action. Maybe "m" like modify, or anything else that allow to work with code using keyboard only. Other changes looks good to me. All disassemblers i know use g as goto, and ctrl+g for something else, so that was confusing. Also good catch with missing delete option in breakpoint context menu. Took me a while to figure out how to remove breapoint first time (delete kb button).

> The debugger pdf is outdated. Sorry, but why use pdfs when they are probably the worst format to edit?

PDF files are what pcsx2 is meant to use for docs afaik, so even removing txt will be probably fine instead. Editing PDF is easy as they are generated from .md files. https://github.com/PCSX2/pcsx2/blob/master/pcsx2/Docs/Debugger.md in that case.

F0bes:
@kozarovv Updated, `M` is the new assemble opcode hotkey.

StorMyu:
I agree with: Why not have delete as a context menu option for breakpoints?

But I don't think it's fair to change the hotkeys for this just because of Ghidra when this specific setup is used throughout a lot of other emulators (I can think of: no$ emulators / BGB / ppsspp)

Ghidra shares some hotkeys with IDA, but I don't think adjusting hotkeys from those isn't especially a great idea on an emulator.

F0bes:
Regarding changing hotkeys, I've only made it so you don't need to press CTRL. 

This is a debugger, you're likely going to be switching between this and tools like Ghidra or IDA. Why make it weird by _requiring_ CTRL?

kozarovv:
> But I don't think it's fair to change the hotkeys for this just because of Ghidra when this specific setup is used throughout a lot of other emulators (I can think of: no$ emulators / BGB / ppsspp)
> 
> Ghidra shares some hotkeys with IDA, but I don't think adjusting hotkeys from those isn't especially a great idea on an emulator.

I don't any problem here. 

> "Disassembly view: hotkeys don't need to be used with CTRL **(they are still usable with CTRL, however)**"

So you can still use CTRL option, while people that get used to Ghidra/IDA/ps2dis/pcsx2dis can use G without CTRL. Only thing that is removed is triggering code edit by random input. Which seems to be good change.

RedDevilus (approved):

PR #4490, opened 2021-06-09
Update vs-version to 16.10

qurious-pixel:


refractionpcsx2:
If only it was that easy.  Github are fixing it on monday, so we just have to be patient until then.

PR #4499, opened 2021-06-22
System | Windows: Print active power profile on application launch

F0bes:
### Description of Changes
Print the current power profile and the current min/max processor state

### Rationale behind Changes
This can help aid users who are experiencing slowdowns and can aid helpers in diagnosing possible slowdown reasons.

### Suggested Testing Steps
You should get a message stating your current power profile and processor min/max states.

### Example
image

### DIFF
--- a/pcsx2/windows/WinPowerProfile.cpp
+++ b/pcsx2/windows/WinPowerProfile.cpp
@@ -1,0 +1,48 @@
+/*  PCSX2 - PS2 Emulator for PCs
+*  Copyright (C) 2002-2021  PCSX2 Dev Team
+*
+*  PCSX2 is free software: you can redistribute it and/or modify it under the terms
+*  of the GNU Lesser General Public License as published by the Free Software Found-
+*  ation, either version 3 of the License, or (at your option) any later version.
+*
+*  PCSX2 is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+*  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+*  PURPOSE.  See the GNU General Public License for more details.
+*
+*  You should have received a copy of the GNU General Public License along with PCSX2.
+*  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "PrecompiledHeader.h"
+#include <powrprof.h>
+
+// Checks the current active power plan
+// If the power plan isn't named 'high performance', put a little message in the console
+// This function fails silently
+void CheckIsUserOnHighPerfPowerPlan()
+{
+	GUID* pPwrGUID;
+	DWORD ret = PowerGetActiveScheme(NULL, &pPwrGUID);
+	if (ret != ERROR_SUCCESS)
+		return;
+
+	UCHAR aBuffer[2048];
+	DWORD aBufferSize = sizeof(aBuffer);
+	ret = PowerReadFriendlyName(NULL, pPwrGUID, &NO_SUBGROUP_GUID, NULL, aBuffer, &aBufferSize);
+	if (ret != ERROR_SUCCESS)
+		goto cleanup;
+
+	DWORD acMax = 0,acMin = 0,dcMax = 0,dcMin = 0;
+
+	if (PowerReadACValueIndex(NULL, pPwrGUID, &GUID_PROCESSOR_SETTINGS_SUBGROUP, &GUID_PROCESSOR_THROTTLE_MAXIMUM, &acMax) ||
+		PowerReadACValueIndex(NULL, pPwrGUID, &GUID_PROCESSOR_SETTINGS_SUBGROUP, &GUID_PROCESSOR_THROTTLE_MINIMUM, &acMin) ||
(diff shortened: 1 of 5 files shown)


pcsx2/windows/WinPowerProfile.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "PrecompiledHeader.h"
+#include <powrprof.h>
+
+// Checks the current active power plan
+// If the power plan isn't named 'high performance', put a little message in the console
+// This function fails silently
+void CheckIsUserOnHighPerfPowerPlan()
+{
+	GUID* pPwrGUID;
+	DWORD ret = PowerGetActiveScheme(NULL, &pPwrGUID);
```
CookiePLMonster, line 25: This leaks memory. As per https://docs.microsoft.com/en-us/windows/win32/api/powersetting/nf-powersetting-powergetactivescheme:
> A pointer that receives a pointer to a GUID structure. Use the LocalFree function to free this memory.

F0bes: Oh, right. I even commented that on my basic sample. Oops.

Margen67:
You should also check for the Ultimate Performance power plan.

tellowkrinkle:
Instead of judging by name, why not read the actual values the power plan sets, since AFAIK users can modify power plans however they want, and you also don't have to worry about Ultimate vs High Power, etc
Details here: https://stackoverflow.com/a/23601787

pcsx2/windows/WinPowerProfile.cpp
```diff
@@ -0,0 +1,47 @@
...
+	if (ret != ERROR_SUCCESS)
+		return;
+
+	DWORD acMax = 0,acMin = 0,dcMax = 0,dcMin = 0;
+
+	if (PowerReadACValueIndex(NULL, pPwrGUID, &GUID_PROCESSOR_SETTINGS_SUBGROUP, &GUID_PROCESSOR_THROTTLE_MAXIMUM, &acMax) ||
+		PowerReadACValueIndex(NULL, pPwrGUID, &GUID_PROCESSOR_SETTINGS_SUBGROUP, &GUID_PROCESSOR_THROTTLE_MINIMUM, &acMin) ||
+		PowerReadDCValueIndex(NULL, pPwrGUID, &GUID_PROCESSOR_SETTINGS_SUBGROUP, &GUID_PROCESSOR_THROTTLE_MAXIMUM, &dcMax) ||
+		PowerReadDCValueIndex(NULL, pPwrGUID, &GUID_PROCESSOR_SETTINGS_SUBGROUP, &GUID_PROCESSOR_THROTTLE_MINIMUM, &dcMin))
+		return;
```
tellowkrinkle, line 41: Not sure how much we care but technically you leak `pPwrGUID` here

You can solve this with RAII or C-style with a label in front of the cleanup and `goto`

F0bes: Resolved

CookiePLMonster: IIRC WIL is not included in PCSX2?
https://github.com/microsoft/wil
It has RAII wrappers e.g. for types that need to be freed with `LocalFree`.

F0bes: It is not in pcsx2. Maybe if we interfaced with win32 a little more, but currently I don't see the need. 
I think the solution is fine without RAII, even if it uses the cursed goto.

pcsx2/windows/WinPowerProfile.cpp
```diff
@@ -0,0 +1,48 @@
...
+	GUID* pPwrGUID;
+	DWORD ret = PowerGetActiveScheme(NULL, &pPwrGUID);
+	if (ret != ERROR_SUCCESS)
+		return;
+
+	UCHAR aBuffer[2048];
+	DWORD aBufferSize = sizeof(aBuffer);
+	ret = PowerReadFriendlyName(NULL, pPwrGUID, &NO_SUBGROUP_GUID, NULL, aBuffer, &aBufferSize);
+	if (ret != ERROR_SUCCESS)
+		return;
```
turtleli, line 33: I think there's a memleak here if this if is true.

RedDevilus:
image

Should say 58% instead of 5%

image

F0bes:
> image
> 
> Should say 58% instead of 5%
> 
> image
> 
> Unless it doesn't print out current battery life.

It prints your minimum and maximum processor state while on battery, and plugged in / desktops (AC)

F0bes:
`gs-merge` conflicts are now resolved.

RedDevilus (approved):
```
The current power profile is 'High performance'.
The current min / max processor states
AC: 100% / 100%
Battery: 5% / 100%
```

```
The current power profile is 'Balanced'.
The current min / max processor states
AC: 5% / 100%
Battery: 5% / 100%
```
```
The current power profile is 'Power saver'.
The current min / max processor states
AC: 5% / 100%
Battery: 5% / 100%
```
```
The current power profile is 'Fobes is a god'.
The current min / max processor states
AC: 5% / 100%
Battery: 5% / 100%
```

lightningterror (approved):
Works fine here.

PR #4503, opened 2021-06-27
microVU: Enable T-Bit to work with MTVU

refractionpcsx2:
### Description of Changes
Enables T-Bit stops to work with MTVU enabled as before some games which required it would just hang/crash

### Rationale behind Changes
Due to the threading nature of MTVU and the fact it has always kept the "busy" bit kept off, games which used T-Bit would get out of sync and mess up quite wildly, this fixes that.

### Suggested Testing Steps
Try games which use T-Bit (maybe test some other game too for performance regressions), here's a short list of games which should work better now, however they may have been unaffected before.

The Sims
Spiderman 3
Sharks Tale - Confirmed to now work in MTVU
Maybe Over the Hedge
Pitfall (both demo and full game) - Confirmed to now work in MTVU
Gumball 3000 (maybe, can't remember)
Evolution Snowboarding

### DIFF
--- a/pcsx2/Gif_Unit.cpp
+++ b/pcsx2/Gif_Unit.cpp
@@ -112,15 +112,15 @@ bool Gif_HandlerAD_MTVU(u8* pMem)
 	if (reg == 0x60)
 	{ // SIGNAL
 		GUNIT_WARN("GIF Handler - SIGNAL");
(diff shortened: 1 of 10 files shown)


atomic83GitHub:
Mafia is for most part unchanged. It will now break at any cyclerate changes but that's due to a another problem related to the VIF FIFO and VU1 timings.

PR #4510, opened 2021-07-04
Remove __rdtsc define. Gets rid of 30 million warnings on Linux while compiling.

arcum42:
### Description of Changes
It now checks if __rdtsc is defined before trying to define it.

### Rationale behind Changes
Since __rdtsc already *is* defined, GS.h is included a lot, and __rdtsc is brought in in the pch with the compiler's version, gcc isn't very happy about this situation.

### Suggested Testing Steps
If anyone has any idea when the gcc 4.8 crash mentioned in the comments happened, try to reproduce it?

### DIFF
--- a/pcsx2/GS/GS.h
+++ b/pcsx2/GS/GS.h
@@ -229,27 +229,6 @@
 #endif
 	}
 
-	#ifdef __GNUC__
-
-	// gcc 4.8 define __rdtsc but unfortunately the compiler crash...
-	// The redefine allow to skip the gcc __rdtsc version -- Gregory
-	#define __rdtsc _lnx_rdtsc
-	//static unsigned long long __rdtsc()
-	static unsigned long long _lnx_rdtsc()
-	{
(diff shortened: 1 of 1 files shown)


tellowkrinkle:
We don't compile on GCC 4.8 anymore so why not just get rid of the thing entirely?

arcum42:
I figured it was better to leave it being defined if it was undefined rather than risking breaking compilation someplace I don't know of, but getting rid of it entirely would certainly work as well. I mostly just don't want to deal with practically every file complaining about it being redefined, burying any legitimate warnings...

tellowkrinkle:
It always existed, just used to be a function, which doesn't give any warnings if you define over it

arcum42:
Got rid of the define instead.

tellowkrinkle (approved):

refractionpcsx2 (approved):

F0bes:
For the record, I just finished counting and I got 31 million warnings 🙄

arcum42:
I could easily have lost count. :P

PR #4512, opened 2021-07-04
GUI: Re-order config settings

RedDevilus:
### Description of Changes

Re-order config setting to older order.

Master:
image

PR:
image

### Rationale behind Changes

Visual clarity between option and also muscle memory.

### DIFF
--- a/pcsx2/gui/MainFrame.cpp
+++ b/pcsx2/gui/MainFrame.cpp
@@ -447,14 +447,15 @@ void MainEmuFrame::CreateCdvdMenu()
 
 void MainEmuFrame::CreateConfigMenu()
 {
-	m_menuConfig.Append(MenuId_Config_SysSettings, _("Emulation &Settings..."));
-	m_menuConfig.Append(MenuId_Config_McdSettings, _("&Memory Cards..."));
-	m_menuConfig.Append(MenuId_Config_BIOS, _("&BIOS Selector..."));
-	m_menuConfig.Append(MenuId_Config_SPU2, _("&Audio Settings..."));
-	m_menuConfig.Append(MenuId_Config_DEV9, _("&Network and HDD Settings..."));
-	m_menuConfig.Append(MenuId_Config_USB, _("&USB Settings..."));
-	m_menuConfig.Append(MenuId_Config_PAD, _("&GamePad Settings..."));
-	m_menuConfig.Append(MenuId_Config_GS, _("&Graphical Settings..."));
+	m_menuConfig.Append(MenuId_Config_SysSettings, _("General &Settings"));
+	m_menuConfig.Append(MenuId_Config_McdSettings, _("&Memory Cards"));
+	m_menuConfig.Append(MenuId_Config_BIOS, _("&BIOS / Folder Selector"));
+	m_menuConfig.AppendSeparator();
+	m_menuConfig.Append(MenuId_Config_GS, _("&Graphics Settings"));
+	m_menuConfig.Append(MenuId_Config_SPU2, _("&Audio Settings"));
+	m_menuConfig.Append(MenuId_Config_PAD, _("Game&pad Settings"));
+	m_menuConfig.Append(MenuId_Config_DEV9, _("&Network and HDD Settings"));
+	m_menuConfig.Append(MenuId_Config_USB, _("&USB Settings"));
 	m_menuConfig.AppendSeparator();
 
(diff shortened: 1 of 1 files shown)


arcum42:
Seriously, the whole thing is a mess anyways. Eight separate setting dialog boxes in eight different menu items? And how many times can we use the word "settings" in a menu?

We really need to combine a lot of these...

RedDevilus:
Yeah, a poll would help in this case.

arcum42:
If we were just going to do some general renaming, I'd suggest something like:
General...
Memory cards...
Bios...
--
Graphics...
Sound...
Gamepads...
Network/Hard drive...
Usb...
--
Multitap 1
Multitap 2
--
Language...
Reset to defaults

Though I'm sure all the translations would break. Language *could* just be a menu, too...

RedDevilus:
Languages are held back anyway, so by anything they are already broken. So don't hold back if you have suggestions.

Shoegzer:
@arcum42 Out of curiosity, are all the ellipses necessary?  Just seems cleaner without them.

arcum42:
> @arcum42 Out of curiosity, are all the ellipses necessary? Just seems cleaner without them.

It's a standard ui convention that if a menu item opens a dialog box, it has an ellipse after it, and all of those open dialog boxes.

tellowkrinkle:
While you're at it, maybe move the `&` on `GamePad Settings` to `Game&Pad Settings` to reduce conflict?

> Though I'm sure all the translations would break.

There aren't any to break
screenshot broken translations

RedDevilus:
image

Macro is fine now as well.

RedDevilus:
image
Removed the punctations marks/ dots.

RedDevilus:
image
Added settings back and removed the dots for about.

RedDevilus:
4446 Current Design
Hopefully this is fine unless someone has some more input. It should be ready for merging.

pcsx2/gui/AppRes.cpp
```diff
@@ -42,7 +42,7 @@
 RecentIsoList::RecentIsoList(int firstIdForMenuItems_or_wxID_ANY)
 {
 	Menu = std::unique_ptr<wxMenu>(new wxMenu());
-	Menu->Append(MenuId_IsoBrowse, _("Browse..."), _("Browse for an ISO that is not in your recent history."));
+	Menu->Append(MenuId_IsoBrowse, _("Browse"), _("Browse for an ISO that is not in your recent history."));
```
lightningterror, line 45: The dots should stay here.

pcsx2/gui/MainFrame.cpp
```diff
@@ -61,7 +61,7 @@ wxMenu* MainEmuFrame::MakeStatesSubMenu(int baseid, int loadBackupId) const
 	// baseid in the negatives will order in a different section, so if you want to increase more slots you can still easily do this, as -1 it will have the same function as opening file for savestates, which is bad
 	// For safety i also made them inactive aka grayed out to signify that's it's only for informational purposes
 	// Fixme: In the future this can still be expanded to actually cycle savestates in the GUI.
-	mnuSubstates->Append(baseid - 1, _("File..."));
+	mnuSubstates->Append(baseid - 1, _("File"));
```
lightningterror, line 64: Here too.

pcsx2/gui/MainFrame.cpp
```diff
@@ -374,7 +374,7 @@ void MainEmuFrame::CreatePcsx2Menu()
 		_("Wipes all internal VM states."));
 	m_menuSys.FindItem(MenuId_Sys_Shutdown)->Enable(false);
 
-	m_menuSys.Append(MenuId_Boot_ELF, _("&Run ELF..."),
```
lightningterror, line 377: Ditto.

pcsx2/gui/MainFrame.cpp
```diff
@@ -506,7 +507,7 @@ void MainEmuFrame::CreateCaptureMenu()
 	keyCode.Shift(false);
 	keyCode.Cmd(false);
 	AppendShortcutToMenuOption(*sysScreenShotItem, keyCode.toTitleizedString());
-	m_submenuScreenshot.Append(MenuId_Capture_Screenshot_Screenshot_As, _("Screenshot As..."));
+	m_submenuScreenshot.Append(MenuId_Capture_Screenshot_Screenshot_As, _("Take Screenshot And Browse"));
```
lightningterror, line 510: Save Screenshot As... ?

PR #4516, opened 2021-07-05
Misc: Remove windows build number in the version string

F0bes:
### Description of Changes
No longer show the Windows build number on emulator launch.

### Rationale behind Changes
GetVersionEx was deprecated and we never use this number anyways.

### Suggested Testing Steps
Launch the emulator and look at the Windows version string.

`Operating System =  Microsoft Windows 10 (64-bit)` is mine

### DIFF
--- a/common/src/Utilities/Windows/WinMisc.cpp
+++ b/common/src/Utilities/Windows/WinMisc.cpp
@@ -25,31 +25,31 @@ static __aligned16 LARGE_INTEGER lfreq;
 
 void InitCPUTicks()
 {
-    QueryPerformanceFrequency(&lfreq);
+	QueryPerformanceFrequency(&lfreq);
 }
 
 u64 GetTickFrequency()
 {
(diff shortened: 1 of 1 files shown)


common/src/Utilities/Windows/WinMisc.cpp
```diff
@@ -118,7 +110,6 @@ wxString GetOSVersionString()
         retval += IsWindowsServer() ? L"Windows Server 2012 R2 " : L"Windows 8.1 ";
 
     pGPI = (PGPI)GetProcAddress(GetModuleHandle(TEXT("kernel32.dll")), "GetProductInfo");
-    pGPI(osvi.dwMajorVersion, osvi.dwMinorVersion, 0, 0, &dwType);
```
tadanokojin, line 121: what is setting dwType now?

F0bes:
More complicated than I imagined.

PR #4519, opened 2021-07-06
Clean up GS.h includes and beef up the PCH

Ziemas:
Hopefully cleaning up GS.h help build times a bit.

Adding stuff to the PCH will make the No-PCH builds even worse though. Let's consider if it's worth it.

### DIFF
--- a/pcsx2/GS/Renderers/Common/GSDevice.h
+++ b/pcsx2/GS/Renderers/Common/GSDevice.h
@@ -22,10 +22,6 @@
 #include "GS/GSAlignedClass.h"
 #include "GSOsdManager.h"
 #include <array>
-#ifdef _WIN32
-#include <dxgi1_3.h>
-#endif
-
 
 enum ShaderConvert
 {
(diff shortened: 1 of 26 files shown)


lightningterror:
Would be nice to properly name the first commit.

Ziemas:
Done

pcsx2/GS/Renderers/Common/GSDevice.h
```diff
@@ -21,6 +21,11 @@
 #include "GSVertex.h"
 #include "GS/GSAlignedClass.h"
 #include "GSOsdManager.h"
+#include <array>
+#ifdef _WIN32
+#include <dxgi1_3.h>
+#endif
```
tadanokojin, line 27: there some reason you stuck this here instead of device11?

Ziemas: No probably not, I think I just slapped includes around until it built.

Can move it.

Ziemas:
Obsoleted by #4709

PR #4521, opened 2021-07-06
Utilities: Remove dependency on GetVersionEx

F0bes:
Attempt # 3
### Description of Changes
Throw out the old version fetching we used to do.
Report if the user is on a version < 8.1

### Rationale behind Changes
GetVersionEx is deprecated, and we only need to know if the user is < 8.1 anyways

### Suggested Testing Steps
Start the application, check the os version printed.
My example: `Operating System =  Microsoft Windows 10, 64-bit`

### DIFF
--- a/common/src/Utilities/Windows/WinMisc.cpp
+++ b/common/src/Utilities/Windows/WinMisc.cpp
@@ -25,54 +25,54 @@ static __aligned16 LARGE_INTEGER lfreq;
 
 void InitCPUTicks()
 {
-    QueryPerformanceFrequency(&lfreq);
+	QueryPerformanceFrequency(&lfreq);
 }
 
 u64 GetTickFrequency()
 {
-    return lfreq.QuadPart;
+	return lfreq.QuadPart;
 }
 
 u64 GetCPUTicks()
 {
-    LARGE_INTEGER count;
-    QueryPerformanceCounter(&count);
-    return count.QuadPart;
+	LARGE_INTEGER count;
+	QueryPerformanceCounter(&count);
+	return count.QuadPart;
 }
 
 u64 GetPhysicalMemory()
 {
-    MEMORYSTATUSEX status;
-    status.dwLength = sizeof(status);
-    GlobalMemoryStatusEx(&status);
-    return status.ullTotalPhys;
+	MEMORYSTATUSEX status;
+	status.dwLength = sizeof(status);
+	GlobalMemoryStatusEx(&status);
+	return status.ullTotalPhys;
 }
 
 // Calculates the Windows OS Version and processor architecture, and returns it as a
 // human-readable string. :)
 wxString GetOSVersionString()
 {
-    wxString retval;
+	wxString retval;
 
-    SYSTEM_INFO si;
+	SYSTEM_INFO si;
 	GetNativeSystemInfo(&si);
 
-    if (!IsWindows8Point1OrGreater())
-        return L"Unsupported Operating System!";
+	if (!IsWindows8Point1OrGreater())
+		return L"Unsupported Operating System!";
 
-    retval += L"Microsoft ";
+	retval += L"Microsoft ";
 
-    if (IsWindows10OrGreater())
-        retval += IsWindowsServer() ? L"Windows Server 2016" : L"Windows 10";
-    else // IsWindows8Point1OrGreater()
-        retval += IsWindowsServer() ? L"Windows Server 2012 R2" : L"Windows 8.1";
+	if (IsWindows10OrGreater())
+		retval += IsWindowsServer() ? L"Windows Server 2016" : L"Windows 10";
+	else // IsWindows8Point1OrGreater()
+		retval += IsWindowsServer() ? L"Windows Server 2012 R2" : L"Windows 8.1";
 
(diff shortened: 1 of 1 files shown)


common/src/Utilities/Windows/WinMisc.cpp
```diff
@@ -25,31 +25,31 @@ static __aligned16 LARGE_INTEGER lfreq;
...
+	return status.ullTotalPhys;
 }
 
-typedef void(WINAPI *PGNSI)(LPSYSTEM_INFO);
-typedef BOOL(WINAPI *PGPI)(DWORD, DWORD, DWORD, DWORD, PDWORD);
+typedef void(WINAPI* PGNSI)(LPSYSTEM_INFO);
+typedef BOOL(WINAPI* PGPI)(DWORD, DWORD, DWORD, DWORD, PDWORD);
 
 // Win 10 SDK
 #ifndef PRODUCT_CORE_N
```
turtleli, line 55: These defines should be removed if they're no longer used.

common/src/Utilities/Windows/WinMisc.cpp
```diff
@@ -79,147 +79,75 @@ typedef BOOL(WINAPI *PGPI)(DWORD, DWORD, DWORD, DWORD, PDWORD);
...
-
-    return retval;
+	wxString retval;
+
+	SYSTEM_INFO si;
+	PGNSI pGNSI;
+
+	memzero(si);
+
+	// Call GetNativeSystemInfo if supported or GetSystemInfo otherwise.
```
turtleli, line 89: Note: `GetNativeSystemInfo` is XP onwards so it'll always be used.

common/src/Utilities/Windows/WinMisc.cpp
```diff
@@ -25,31 +25,31 @@ static __aligned16 LARGE_INTEGER lfreq;
...
+	MEMORYSTATUSEX status;
+	status.dwLength = sizeof(status);
+	GlobalMemoryStatusEx(&status);
+	return status.ullTotalPhys;
 }
 
-typedef void(WINAPI *PGNSI)(LPSYSTEM_INFO);
-typedef BOOL(WINAPI *PGPI)(DWORD, DWORD, DWORD, DWORD, PDWORD);
+typedef void(WINAPI* PGNSI)(LPSYSTEM_INFO);
+typedef BOOL(WINAPI* PGPI)(DWORD, DWORD, DWORD, DWORD, PDWORD);
```
turtleli, line 52: This typedef is no longer used.

common/src/Utilities/Windows/WinMisc.cpp
F0bes: I removed the entire check if you think that's okay. GetProcAddress shouldn't fail in this case correct?

turtleli: Does `GetNativeSystemInfo(&si);` not work?

F0bes: Oh jeez I got it mixed up

seta-san:
What a decomplication of a big mess that is

common/src/Utilities/Windows/WinMisc.cpp
```diff
@@ -48,133 +48,24 @@ u64 GetPhysicalMemory()
...
     if (!IsWindows8Point1OrGreater())
         return L"Unsupported Operating System!";
 
     retval += L"Microsoft ";
 
-    // Test for the specific product.
-
     if (IsWindows10OrGreater())
-        retval += IsWindowsServer() ? L"Windows Server 2016 " : L"Windows 10 ";
+        retval += IsWindowsServer() ? L"Windows Server 2016 " : L"Windows 10";
```
lightningterror, line 66: I believe the spacing after Windows 10 string is intentional.

common/src/Utilities/Windows/WinMisc.cpp
```diff
@@ -48,133 +48,24 @@ u64 GetPhysicalMemory()
...
-        case PRODUCT_CORE_N:
-            retval += L"Home N";
-            break;
-        case PRODUCT_EDUCATION_N:
-            retval += L"Education N";
-            break;
-    }
-
-    retval += wxsFormat(L" (build %d)", osvi.dwBuildNumber);
+        retval += IsWindowsServer() ? L"Windows Server 2012 R2 " : L"Windows 8.1";
```
lightningterror, line 68: Here too.

common/src/Utilities/Windows/WinMisc.cpp
F0bes: It would be "Windows 10 , 64 bit" then

lightningterror: `It would be "Windows 10 , 64 bit" then`
Your change is fine then.
`I did just now notice that I didn't do it for the other strings....`
Yes need to do the other 2.

CookiePLMonster:
Might be a good excuse to un-wxString this function, since it seems to have nothing to do with wx.

tadanokojin:
It might be, but I'm going to hold off asking to do that unless in a follow up PR.

PR #4523, opened 2021-07-07
Replace ATL with WIL

CookiePLMonster:
### Description of Changes
This PR replaces an old COM management solution using ATL with [WIL](https://github.com/microsoft/wil), a header-only library offering a modern, lightweight COM wrapper as well as RAII wrappers for practically every Windows resource. For this reason I considered it better than WRL that only offers a COM wrapper.

While passing through the code and replacing manually managed resources with WIL, I also improved a few archaic code parts - such as macro-based "lambdas" in `GSCapture`.

This PR does not cover every use of WinAPI resources and/or COM - there are still a few files with raw COM pointers, like `dinput.cpp` and `cam-windows.cpp`, but for this PR I only wanted to deprecate ATL and fix whatever else I can spot in the files I modify.

One more change I kind of had to make even though it's unrelated is a global define of `WINVER` to `0x0602` (Win8). The inclusion of WRL in one of the files revealed that our `WINVER` is actually globally set to `0x0600` because of wx, and mismatches can occur! I fixed it by defining it in project settings instead, so no compilation unit can miss it.

### Rationale behind Changes
ATL is ancient and has design flaws that resulted in code being more janky - implicit conversions everywhere, inability to release an object before attaching a new one etc. WIL has none of these issues.

For WinAPI resources, manually managing their lifetime is error prone and during the migration I "implicitly" fixed around a dozen such resource leaks. Since WIL offers `unique_ptr` like wrappers for these resources, their lifetime manages itself.

### Suggested Testing Steps
PCSX2 has been verified to still boot fine, run games with a DX11 backend (the one affected by these refactors), display things properly and not crash on exit. I have not tested for regressions beyond that, but they are extremely unlikely.


CookiePLMonster:
Rebased on the latest master.

CookiePLMonster:
Rebased on latest master, and also added a change making `CheckIsUserOnHighPerfPowerPlan` use WIL so explicit control flow for cleanup is not needed.

pcsx2/CMakeLists.txt
```diff
@@ -85,9 +85,12 @@ if(WIN32)
 	)
 	target_compile_definitions(PCSX2 PRIVATE
 		_M_SSE=0x401 # TODO: Multiple ISA
+		WINVER=0x0602
+		_WIN32_WINNT=0x0602
```
tadanokojin, line 89:
```suggestion
		WINVER=0x0603
		_WIN32_WINNT=0x0603
```

https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-160
> #define _WIN32_WINNT_WIN8                   0x0602 // Windows 8
> #define _WIN32_WINNT_WINBLUE                0x0603 // Windows 8.1

pcsx2/DEV9/Win32/tap-win32.cpp
```diff
@@ -22,15 +22,16 @@
...
-#include <atlbase.h>
 #include <devguid.h>
 
 #include <tchar.h>
 #include "tap.h"
 #include "..\dev9.h"
 #include <string>
 
+#include "wil/com.h"
+#include "wil/resource.h"
```
tadanokojin, line 33:
```suggestion
#include <wil/com.h>
#include <wil/resource.h>
```

pcsx2/GS/GS.h
```diff
@@ -33,8 +33,9 @@
 #include <d3dcompiler.h>
 #include <d3d11_1.h>
 #include <dxgi1_3.h>
-#include <comutil.h>
-#include <atlcomcli.h>
+
+#include "wil/com.h"
+#include "wil/resource.h"
```
tadanokojin, line 38:
```suggestion
#include <wil/com.h>
#include <wil/resource.h>
```

Additionally with this one I'd prefer if we didn't put it in the `GS.h` header but rather individual headers/cpp files that depend on it.

see #4709

tests/ctest/GS/CMakeLists.txt
```diff
@@ -23,4 +23,15 @@ foreach(isa "sse4" "avx" "avx2")
...
+	if(WIN32)
+		target_compile_definitions(swizzle_test_${isa} PRIVATE
+			WINVER=0x0602
+			_WIN32_WINNT=0x0602
+			WIN32_LEAN_AND_MEAN
+			WIL_SUPPRESS_EXCEPTIONS
+		)
+		target_link_libraries(swizzle_test_${isa} PRIVATE
+			WIL::WIL
+		)
```
tadanokojin, line 35: note: this might not be required after you get the headers out of `GS.h`.

pcsx2/SPU2/Windows/SndOut_XAudio2.cpp
```diff
@@ -17,19 +17,20 @@
...
 #include <cguid.h>
-#include <atlcomcli.h>
 #include <memory>
 #include <sstream>
 #include <stdexcept>
 #include <string>
 
+#include "wil/com.h"
+#include "wil/resource.h"
+#include "wil/win32_helpers.h"
```
tadanokojin, line 29:
```suggestion
#include <wil/com.h>
#include <wil/resource.h>
#include <wil/win32_helpers.h>
```

pcsx2/windows/WinPowerProfile.cpp
```diff
@@ -15,34 +15,33 @@
 
 #include "PrecompiledHeader.h"
 #include <powrprof.h>
+#include "wil/com.h"
```
tadanokojin, line 18:
```suggestion
#include <wil/com.h>
```

common/vsprops/BaseProperties.props
```diff
@@ -11,7 +11,7 @@
     <ClCompile>
       <IntrinsicFunctions>true</IntrinsicFunctions>
       <AdditionalIncludeDirectories>$(SolutionDir);$(SolutionDir)\3rdparty\wxWidgets3.0\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <PreprocessorDefinitions>__WIN32__;WIN32;_WINDOWS;_CRT_SECURE_NO_WARNINGS;_CRT_SECURE_NO_DEPRECATE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>__WIN32__;WIN32;_WINDOWS;_CRT_SECURE_NO_WARNINGS;_CRT_SECURE_NO_DEPRECATE;WINVER=0x0602;_WIN32_WINNT=0x0602;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
tadanokojin, line 14:
```suggestion
      <PreprocessorDefinitions>__WIN32__;WIN32;_WINDOWS;_CRT_SECURE_NO_WARNINGS;_CRT_SECURE_NO_DEPRECATE;WINVER=0x0603;_WIN32_WINNT=0x0603;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```

tadanokojin (commented):
Everything looks pretty good. Don't know much about networking or xaudio but as far as I can tell it's correct. Not including wil in the `GS.h` is my biggest concern right now as I believe our test project depends on that header right now so it's leaking a probably unneeded wil dependency.

AsmodeanBreaggAI:
Hello there. This change may have broken the recording menu's ability to index codecs on the computer. I'm on Windows 10 specifically.
image
Attached above is the list on the new version. However, it's missing the x264vfw codec (and many old ones that granted I never used anyways) that I know exists on my computer.
image
Attached above is a version of PCSX2 prior to this change. It seems to be missing anything with the VfW prefix. Let me know if any other information is needed.

CookiePLMonster:
@AsmodeanBreaggAI This doesn't seem to be true, I am not getting those codecs on builds before this PR too.

PR #4529, opened 2021-07-08
GHActions: Remove dashes from build names

tellowkrinkle:
### Description of Changes
Updates Linux CI names to not have dashes in the build names

### Rationale behind Changes
Build names were previously updated to use spaces instead of dashes but the `detail` section was left out for some reason

### Suggested Testing Steps
If this breaks something other than the CI, I don't know what to say

### DIFF
--- a/.github/workflows/linux-workflow.yml
+++ b/.github/workflows/linux-workflow.yml
@@ -70,12 +70,12 @@ jobs:
           - os: ubuntu-18.04
             platform: x86
             compiler: gcc
-            detail: appimage
+            appimage: true
             experimental: false
           - os: ubuntu-18.04
             platform: x64
             compiler: gcc
-            detail: appimage
+            appimage: true
             experimental: false
           - os: ubuntu-18.04
             platform: x86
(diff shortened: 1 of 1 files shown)


qurious-pixel:
Should there be dashes in the ccache keys between ${{ matrix.compiler }} and ${{ matrix.detail }}?

https://github.com/PCSX2/pcsx2/blob/e1f482cbae3596b7a0998044024d0603068ec9bb/.github/workflows/linux-workflow.yml#L97-L99

`${{ matrix.os }}-${{ matrix.platform }}-${{ matrix.compiler }}-${{ matrix.detail }}-ccache-`

tellowkrinkle:
> Should there be dashes in the ccache keys between ${{ matrix.compiler }} and ${{ matrix.detail }}?

I switched the ccache keys to use spaces as well

.github/workflows/linux-workflow.yml
```diff
@@ -45,22 +45,22 @@ jobs:
...
           - os: ubuntu-20.04
             platform: x86
             compiler: gcc
             cmakeflags: -DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON -DARCH_FLAG=-march=haswell
-            detail: -avx2-nopch
+            detail: " avx2 nopch"
             experimental: false
 
-    name: "${{ matrix.platform }} | ${{ matrix.compiler }} ${{ matrix.detail }}"
+    name: "${{ matrix.platform }} | ${{ matrix.compiler }}${{ matrix.detail }}"
```
Margen67, line 63: This doesn't need quotation marks.
```suggestion
    name: ${{ matrix.platform }} | ${{ matrix.compiler }}${{ matrix.detail }}
```

PR #4539, opened 2021-07-12
VU: Improve sync with VU Kickstart, loosen without kickstart

refractionpcsx2:
### Description of Changes
This tightens the VU Sync with the EE when Kickstart is enabled to fix some SPS in PAL Ratchet games, also loosened it slightly when Kickstart is disabled to improve the speed of games which don't require it.

### Rationale behind Changes
Ratchet games were broken since the VU Sync changes, this resolves it.

### Suggested Testing Steps
Test any games you can (Especially those that require Kickstart or have M-Bit), check for SPS and freaky Physics problems.


atomic83GitHub:
It is worth testing the framerate of all Ratchet games, the frametimes have seen some improvements on my side, especially for the first episode.

PR #4547, opened 2021-07-14
Capture: Implement audio file customization

sonicfind:
### Description of Changes
Gives users the ability to change the channel arrangement (mono or stereo) for a captured wav file.

### Rationale behind Changes
We're most likely leaning towards a complete separation of audio and video capture - or really, allowing for audio to be captured separately from video if a user chooses to do so. It makes sense to thus add some switchable options to expand the use cases.

Also, wavfile.h/cpp from soundtouch had a lot of irrelevant functions for our context, so moving towards a completely custom version of those files that can be used on any pc config is probably best.

### Suggested Testing Steps
1. Check to ensure that pcsx2_ui.ini has these written correctly.
2. Check in some external audio program (audacity pre-3.0 for example) that the wav format matches what settings were chosen.


sonicfind:
Yea see, I knew there had to be something I missed with regards to the other system configs. 😅

refractionpcsx2:
I'd recommend grabbing the sample rate from the SPU2 if you can and multiplying the number based on that, because I'm not sure how happy it would be if somebody decided to record PS1 mode since you have it hard coded to 48khz (and multiples)

sonicfind:
Ah, I don’t own any PS1 games, so I wasn’t aware it worked in PS1 mode. Good to know. Guess I am doing that.

sonicfind:
That was.... much easier than I expected.

lightningterror:
Closing as it's not currently being worked on.

PR #4548, opened 2021-07-15
IOP Interpreter: Warn on branches to zero

F0bes:
### Description of Changes
Whenever there is a branch to zero, put a warning in the console. 

### Rationale behind Changes
The interpreter can start executing garbage, and this will tell us why.
(Also helps me debug my poor IOP code :P)

### Suggested Testing Steps
Make the IOP branch to zero.

### DIFF
--- a/pcsx2/R3000AInterpreter.cpp
+++ b/pcsx2/R3000AInterpreter.cpp
@@ -1,5 +1,5 @@
 /*  PCSX2 - PS2 Emulator for PCs
- *  Copyright (C) 2002-2010  PCSX2 Dev Team
+ *  Copyright (C) 2002-2021  PCSX2 Dev Team
  *
  *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
  *  of the GNU Lesser General Public License as published by the Free Software Found-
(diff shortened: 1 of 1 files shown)


pcsx2/x86/iR3000A.cpp
```diff
@@ -866,11 +866,22 @@ static __fi void recClearIOP(u32 Addr, u32 Size)
...
+
 void psxSetBranchReg(u32 reg)
 {
 	psxbranch = 1;
 
 	if( reg != 0xffffffff ) {
+
+		if (PSX_IS_CONST1(reg) ? g_psxConstRegs[reg] == 0 : psxRegs.GPR.r[reg] == 0)
+			// We are still gonna hard crash and burn but at least the emulog tells us why :)
+			xFastCall((void*)recNotifyBranchToZero);
```
refractionpcsx2, line 883: Is there any reg flushing function on the IOP?  I'm nervous of calling stuff outside the JIT without flushing, that can be a bit dangerous

F0bes: I'll double check, it's hard to test because the recompiler is going to die anyways :p

F0bes:
Ok, this PR is no longer for the recompiler and it'll be for the interpreter only.
I can't for the life of me get the target address for branches whose register isn't const. 
Idk, I'm still no recompiler geek :(

PR #4561, opened 2021-07-17
vs: properly disable gm and warnings

tadanokojin:
### Description of Changes
Clean up a number of vs project files that had been polluted with duplicate settings.

### Rationale behind Changes
I want this to be cleaner

### Suggested Testing Steps
If it doesn't build it doesn't work


tadanokojin:
Hey look it works. I'm the best tester.

PR #4562, opened 2021-07-17
Debian: Remove debian specific packaging hooks.

GovanifY:
This removes our debian specific packaging hooks which were used by our PPA.
As we switched to a distribution agnostic format (appimage) and upstream has their own fork, we don't really need to maintain those anymore


.github/workflows/macos-workflow.yml
```diff
@@ -36,7 +35,6 @@ on:
     branches:
       - master
     paths-ignore:
-      - '**/*.md'
```
lightningterror, line 39: Reason for removing md files?

PR #4565, opened 2021-07-17
vs: drop asio build configuration for portaudio

tadanokojin:
### Description of Changes
Drops the multiple configurations for portaudio concerning ASIO and just builds without ASIO (as is master). Additionally I cleaned up some of the configuration.

### Rationale behind Changes
The general idea is that we don't really need these different build configurations in portaudio. ASIO was never used and while it's not as trivial now to add it back, it can be done with little effort if we need. Whether we use ASIO or not there's not much reason to maintain two different build configurations for every arch and build type.

### Suggested Testing Steps
Portaudio should still work and not have an option for ASIO as it does on master. I've verified this locally so there's not much reason to test it.


refractionpcsx2:
Another reason not to bother with ASIO is it requires additional licensing, so screw that.

PR #4567, opened 2021-07-18
Replace opaque display pointer with window info struct and GS cleanups

stenzek:
### Description of Changes

The current setup for passing the GS window information to the subsystems was error-prone, especially on Linux where multiple window systems exist. Wrapping the information up into a window information structure makes this safer, as well as supporting multiple window systems within the same binary.

This PR also drops the "managed window" concept from GS, relying solely on the frontend to provide a window to handle into.

### Rationale behind Changes

Safer code, flexibility to support more platforms in the future, and removing unused functions.

### Suggested Testing Steps

I've tested Windows, and Linux with X11. Wayland is still not supported by default, I have another branch which removes GSWnd completely and does work with Wayland.

### DIFF
--- a/common/include/Utilities/WindowInfo.h
+++ b/common/include/Utilities/WindowInfo.h
@@ -1,0 +1,22 @@
+#pragma once
+#include "Pcsx2Defs.h"
+
+// Contains the information required to create a graphics context in a window.
+struct WindowInfo
+{
+  enum class Type
+  {
+    Surfaceless,
+    Win32,
+    X11,
+    Wayland,
+    MacOS
+  };
+
+  Type type = Type::Surfaceless;
+  void* display_connection = nullptr;
+  void* window_handle = nullptr;
+  u32 surface_width = 0;
+  u32 surface_height = 0;
+  float surface_scale = 1.0f;
+};
--- a/pcsx2/MTGS.cpp
+++ b/pcsx2/MTGS.cpp
@@ -250,7 +250,7 @@ void SysMtgsThread::OpenGS()
 	GSsetBaseMem(RingBuffer.Regs);
 	GSirqCallback(dummyIrqCallback);
 
-	pxAssertMsg((GSopen2((void**)pDsp, 1 | (renderswitch ? 4 : 0)) == 0), "GS failed to open!");
+	pxAssertMsg((GSopen2(g_gs_window_info, 1 | (renderswitch ? 4 : 0)) == 0), "GS failed to open!");
 
 	GSsetVsync(EmuConfig.GS.GetVsync());
 
--- a/pcsx2/gui/FrameForGS.cpp
+++ b/pcsx2/gui/FrameForGS.cpp
@@ -43,6 +43,20 @@
 #include <sstream>
 #include <iomanip>
 
+#ifdef __WXGTK__
+#include <gdk/gdkx.h>
+#include <gtk/gtk.h>
+
+// Junk X11 macros...
+#ifdef DisableScreenSaver
+#undef DisableScreenSaver
+#endif
+#endif
+
+#ifdef GDK_WINDOWING_WAYLAND
+#include <gdk/gdkwayland.h>
+#endif
+
 //#define GSWindowScaleDebug
 
 static const KeyAcceleratorCode FULLSCREEN_TOGGLE_ACCELERATOR_GSPANEL=KeyAcceleratorCode( WXK_RETURN ).Alt();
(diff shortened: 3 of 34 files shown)


pcsx2/MTGS.cpp
```diff
@@ -250,7 +250,7 @@ void SysMtgsThread::OpenGS()
 	GSsetBaseMem(RingBuffer.Regs);
 	GSirqCallback(dummyIrqCallback);
 
-	pxAssertMsg((GSopen2((void**)pDsp, 1 | (renderswitch ? 4 : 0)) == 0), "GS failed to open!");
+	pxAssertMsg((GSopen2(g_gs_window_info, 1 | (renderswitch ? 4 : 0)) == 0), "GS failed to open!");
```
stenzek, line 253: I'm not a fan of storing this in a global, but I couldn't see any obvious way to get it through to MTGS/SysCoreThread. Ideally, this would be computed on demand by the host, and passed through on initialization.

common/include/Utilities/WindowInfo.h
```diff
@@ -0,0 +1,22 @@
...
+    MacOS
+  };
+
+  Type type = Type::Surfaceless;
+  void* display_connection = nullptr;
+  void* window_handle = nullptr;
+  u32 surface_width = 0;
+  u32 surface_height = 0;
+  float surface_scale = 1.0f;
+};
```
tellowkrinkle, line 22: Needs a clang-format for tabs

pcsx2/gui/FrameForGS.cpp
```diff
@@ -243,6 +257,66 @@ void GSPanel::DoShowMouse()
...
 }
 
+std::optional<WindowInfo> GSPanel::GetWindowInfo()
+{
+	WindowInfo ret;
+
+	const wxSize gs_vp_size(GetClientSize());
+	ret.surface_scale = static_cast<float>(GetContentScaleFactor());
+	ret.surface_width = static_cast<u32>(gs_vp_size.GetWidth());
+	ret.surface_height = static_cast<u32>(gs_vp_size.GetHeight());
```
tellowkrinkle, line 267: If we're tracking this info here, do you plan to call this every frame?  Otherwise how do you handle window resizes?  If you are, I don't know about other platforms, but on macOS these methods fall under the "must be called from the main thread" category, which makes calling it from the GS thread kind of difficult.

stenzek: At the moment, it's tracked by the GS thread, same as it is in master. This is just the initial dimensions, which are (currently) unused. Trying to avoid too many changes at once :)

For Wayland, it's impossible to query the size, so the size events need to be passed to the GS thread anyway. My plan was to add another command type to MTGS which passes the dimensions through, and resizes the target on the GL thread (which is safe).

For Macs, yeah, it's a pain. I haven't messed with it in a while, but for DuckStation I just use GCD to kick the GL context update back to the main thread. It's kinda janky to have GL stuff in the frontend/window side of things IMO.

stenzek: Also, for Macs, I can't see how it would even work at the moment given there's no window/GL context wrapper. Rather than implementing GSWnd and then tossing it away when I drop GSWnd completely, I was just going to leave that for later.

tellowkrinkle (commented):
Main comment is below, but I'm a bit confused as to what window size info is doing in the WindowInfo struct

pcsx2/gui/FrameForGS.cpp
tellowkrinkle: > Also, for Macs, I can't see how it would even work at the moment given there's no window/GL context wrapper. 

I'm currently maintaining it on a fork as it requires some extra work to work around Apple's ancient OpenGL implementation, but you can see it here: https://github.com/tellowkrinkle/pcsx2/blob/MacOS/pcsx2/GS/Window/GSWndCGL.mm

stenzek: Seems pretty similar to what I did: https://github.com/stenzek/duckstation/blob/master/src/common/gl/context_agl.mm

Indeed, there's a potential race between the backbuffer actually updating (on the main thread) and the rendering thread using the new size (passed in through the resize event). But that's going to happen regardless, even if you did the resize on the main thread, unless you paused rendering while the update happens, which I feel like has the potential to deadlock. As far as I'm aware, the worst that would happen is you render to a smaller/invalid viewport for one frame.

I guess another option could be to block the rendering thread until the main thread processes the update, but I haven't tried that.

tellowkrinkle: > But that's going to happen regardless, even if you did the resize on the main thread, unless you paused rendering while the update happens, which I feel like has the potential to deadlock. As far as I'm aware, the worst that would happen is you render to a smaller/invalid viewport for one frame.

Apple doesn't talk about the thread safety of trying to access NSView parameters from anywhere other than the main thread, other than to say "just don't do it".  It's *probably* okay, but it makes their automated checker scream and throw warnings everywhere, which kind of bothers me so I'd rather avoid it if possible

And yeah I think having the chance of being a frame behind in your resize is kind of unavoidable unless you're doing your rendering from the main thread (which we're not)

tellowkrinkle:
Oh and feel free to boost our minimum macOS version to 10.14 for `std::optional`

pcsx2/gui/FrameForGS.cpp
stenzek: > Apple doesn't talk about the thread safety of trying to access NSView parameters from anywhere other than the main thread, other than to say "just don't do it". It's probably okay with these read-only things, but it makes their automated checker scream and throw warnings everywhere, which kind of bothers me so I'd rather avoid it if possible

With the way I'm envisioning things, there won't be any `NSView` calls happening on the GS thread, it'll all be on the UI thread, and then handed off to the GS thread. Which might result in redundant resizes if the user is dragging the window out, but it should be safe.

CookiePLMonster:
> Oh and feel free to boost our minimum macOS version to 10.14 for `std::optional`, it's in `cmake/BuildParameters.cmake`

Since `std::optional` is C++17, won't this allow PCSX2 to remove `ghc::filesystem` in favour of `std::filesystem`?

tellowkrinkle:
> Since std::optional is C++17, won't this allow PCSX2 to remove ghc::filesystem in favour of std::filesystem?

C++17 was added over time, and `std::filesystem` was last, at 10.15

PR #4569, opened 2021-07-18
vs: consolate build configuration to props files

tadanokojin:
### Description of Changes
Creates 2 new props files and updates every project to use them:
- `BaseBuildConfig.props` contains the base build configuration present in every project. This includes `Debug`, `Devel` and `Release` configs for the `Win32` and `x64` platforms.
- `BuildConfigAVX2.props` contains the build configuration for AVX2 for the `Win32` and `x64` platforms.

### Rationale behind Changes
Having these configurations in the same place has the following benefits:
- Less boilerplate required when creating new projects
- Editing these settings no longer requires touching every project file
- Helps ensure consistency across all the project build configurations.

### Suggested Testing Steps
As usual with these sorts of changes there's really nothing that needs testing.


3rdparty/soundtouch/SoundTouch.vcxproj
```diff
@@ -1,31 +1,7 @@
...
-      <Configuration>Release</Configuration>
-      <Platform>Win32</Platform>
-    </ProjectConfiguration>
-    <ProjectConfiguration Include="Release|x64">
-      <Configuration>Release</Configuration>
-      <Platform>x64</Platform>
-    </ProjectConfiguration>
-  </ItemGroup>
+  <Import Project="$(SolutionDir)common\vsprops\BaseProjectConfig.props" />
+  <Import Project="$(SolutionDir)common\vsprops\WinSDK.props" />
```
stenzek, line 4: Possible side-effect here as it didn't contain `WinSDK.props` before, but consistent with the others.

stenzek (approved):
LGTM otherwise!

lightningterror (approved):

PR #4570, opened 2021-07-19
Dependencies: Update GHC 1.5.4 to 1.5.8

RedDevilus:
### Description of Changes

Make ghc act more like std::fs +  + consistency with C++17 and other fixes.
### Rationale behind Changes

https://github.com/gulrak/filesystem/releases has some interesting fixes like issues with mounted folders and more wchar support.
### Suggested Testing Steps

Test with MacOS and see if everything works fine.


RedDevilus:
Think it's fine as-is, it's just touches the header for GHC.

tadanokojin:
@tellowkrinkle any qualms about this one?

tellowkrinkle (approved):

PR #4572, opened 2021-07-19
vcxproj: Include shaders in vs project.

lightningterror:
### Description of Changes

Includes shaders in vs project.


tadanokojin (approved):
LGTM

PR #4578, opened 2021-07-21
CHD: Fix parent search on windows

Ziemas:
It was crashing in some string conversion function.

Apparently you can't use the constructor to concat a literal and a wxString like that.


DukeSkinny:
Thank you for this.

PR #4579, opened 2021-07-22
cmake: Fix missing librt include

tellowkrinkle:
### Description of Changes
Re-add link against librt

### Rationale behind Changes
Was accidentally removed in #4373

### Suggested Testing Steps
Try to build if you had trouble with librt failing since #4373 
@sjnewbury please test


sjnewbury:
@tadanokojin Build works fine now, thanks!

PR #4613, opened 2021-07-23
IPU: Store thresholds for color conversions in u16, bump savestate version.

kozarovv:
### Description of Changes
Store color conversion thresholds in u16 instead of u8. Bump save state version.
Fix accidentally created UB from my previous PR.
### Rationale behind Changes
While SETTH command was accepting 8:0 and 24:16 bits after merging #4526 , I didn't noticed that variable is declared to store only 8 bit values. So SETTH in result still masked bit 8, and 24, additionally creating risk of overflow.
This PR fix it.
### Suggested Testing Steps
Testing FMVs for regressions/improvements. Possibly affect IPU texturing (rare thing afaik).


RedDevilus:
Maybe update the copyright dates for the files as well?

kozarovv:
> Maybe update the copyright dates for the files as well?

Done.

PR #4615, opened 2021-07-24
GS: Manage draw rectangle in GS instead of wx

stenzek:
### Description of Changes

Instead of having a second subwindow inside the GS window which is rendered into, resulting in a mix of software/Win32 and hardware surfaces, this PR makes the GS panel take up the full client area of the window, and the renderer performs the aspect ratio correction instead.

### Rationale behind Changes

This is (imo) cleaner, as well as enabling fullscreen optimizations/directflip. It also removes the redundant code from GS.

### Suggested Testing Steps

Make sure all aspect ratios work, hotkey switching functions, ensure that FMV auto switching isn't broken.

### DIFF
--- a/pcsx2/GS/GS.cpp
+++ b/pcsx2/GS/GS.cpp
@@ -419,9 +419,6 @@ int GSopen2(void** dsp, uint32 flags)
 
 	int retval = _GSopen(dsp, "", current_renderer);
 
-	if (s_gs != NULL)
-		s_gs->SetAspectRatio(0); // PCSX2 manages the aspect ratios
-
 	gsopen_done = true;
 
 	return retval;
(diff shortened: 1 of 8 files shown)


Mrlinkwii:
on linux ,  when first  booting  a game with default settings in  HW mode their is  a black screen  apart form the very  left corner  ,changing between  HW/SW mode can fix it but not always   , 
image
image

stenzek:
Hmm, probably need to explicitly resize the child window then. Strange that this differs between Linux and windows, but it is wx...

stenzek:
> on linux , when first booting a game with default settings in HW mode their is a black screen apart form the very left corner ,changing between HW/SW mode can fix it but not always ,

Should be fixed now.

Ziemas:
Yup, it works for me now.

Immersion95:
Works great in Windows+Open GL. We now only needs DX11 to use the flip model :).

tadanokojin:
Thanks Sten, look forward to working with you.
image

PR #4617, opened 2021-07-24
X86Emitter: Use __rdtsc define on non-msvc compilers

F0bes:
### Description of Changes
Use __rdtsc compiler extension on all compilers

### Rationale behind Changes
GCC 4.8 isn't supported anymore (we are c++ 17 now), the supposed crash related to it is not a problem anymore.

### Suggested Testing Steps
Compile using gcc / clang and see if there is a crash.

I tested this on clang 12.0.1 and gcc 11.1.0 (64-bit) with no apparent crashes.


F0bes:
Ok the 32 bit appimage works for me (I cannot compile 32 bit) which tells me that the __rdtsc is fine on a modern GCC.

PR #4619, opened 2021-07-24
GIF: Reimplement GIF FIFO to enable only when it is needed.

refractionpcsx2:
### Description of Changes
Reworks the GIF FIFO to always be enabled and only kick in when needed.  Savestate version bumped

### Rationale behind Changes
The old GIF FIFO was a bit janky/broken and it was quite slow when used, the new changes allow it to always be on without impacting performance.

### Suggested Testing Steps
Test mainly games which used to have the GIF FIFO hack entry in the GameDB, or games which use PATH3 masking, these will be impacted the most

### DIFF
--- a/pcsx2/FiFo.cpp
+++ b/pcsx2/FiFo.cpp
@@ -125,10 +125,10 @@ void __fastcall WriteFIFO_VIF1(const mem128_t *value)
 void __fastcall WriteFIFO_GIF(const mem128_t *value)
 {
 	GUNIT_LOG("WriteFIFO_GIF()");
-	if (CHECK_GIFFIFOHACK) {
-		gif_fifo.write((u32*)value, 1);
-
-		gif_fifo.read(true);		
+	if ((!gifUnit.CanDoPath3() || gif_fifo.fifoSize > 0)) {
+		//DevCon.Warning("GIF FIFO HW Write");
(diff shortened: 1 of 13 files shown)


lightningterror:
Once changes are ready it would be nice to run clang on the files for consistent formatting.

pcsx2/FiFo.cpp
```diff
@@ -125,10 +125,10 @@ void __fastcall WriteFIFO_VIF1(const mem128_t *value)
 void __fastcall WriteFIFO_GIF(const mem128_t *value)
 {
 	GUNIT_LOG("WriteFIFO_GIF()");
-	if (CHECK_GIFFIFOHACK) {
-		gif_fifo.write((u32*)value, 1);
-
-		gif_fifo.read(true);		
+	if ((!gifUnit.CanDoPath3() || gif_fifo.fifoSize > 0)) {
```
lightningterror, line 128: This file didn't get formatted, oh well another time.

refractionpcsx2: Happy now? :P

lightningterror: Bellissimo.

HeroponRikiBestest:
Breaks skipping the opening FMV in Air Ranger 2.

refractionpcsx2:
Fixed in #5324

PR #4580, opened 2021-07-22
PCSX2-GUI: Restore Defaults button - speedhacks panel

RedDevilus:
### Description of Changes

Move to higher position of Restore Defauls in speedhacks panel.
Master:
BrokenMaster1
BrokenMaster2RestoreBelowOtherButtons
BrokenMaster3ResizeStdExpandSmall
PR:
2021-07-22_08-18 HackFix SpeedHacks

### Rationale behind Changes

Recent change to icon alignment made speedhacks panel act in odds with Restore Defaults. Presumably a DPI issue with combination of GUI issue.
### Suggested Testing Steps

Mess around with the sliders and the presets. Note this won't be perfect either way.

PR #4620, opened 2021-07-25
PCSX2-GUI: Rehaul BIOS / Folder Selector

RedDevilus:
### Description of Changes

- Remove button BIOS / Folder Selector in GUI
- Move BIOS / Folder Selector contents in General Settings
### Rationale behind Changes

- Since emulation settings has been renamed general settings, we can put BIOS and Folder in the same location where people have an overview and can just use tab key on icons or mouse. Less confusion and also fixes the speedhacks section *for the zillionth time*.
### Suggested Testing Steps

- Test if runtime wizard works, game starts and esspecialy emulation settings (while checking size of the window is good enough)

### DIFF
--- a/3rdparty/wxwidgets3.0/src/common/sizer.cpp
+++ b/3rdparty/wxwidgets3.0/src/common/sizer.cpp
@@ -1040,10 +1040,10 @@ wxSize wxSizer::GetMinSize()
     return ret;
 }
 
-void wxSizer::DoSetMinSize(int width, int height)
+void wxSizer::DoSetMinSize( int width, int height )
 {
-	m_minSize.x = width;
-	m_minSize.y = height;
+    m_minSize.x = width;
+    m_minSize.y = height;
 }
 
 bool wxSizer::DoSetItemMinSize( wxWindow *window, int width, int height )
--- a/3rdparty/wxwidgets3.0/src/common/sizer.cpp
+++ b/3rdparty/wxwidgets3.0/src/common/sizer.cpp
@@ -1040,10 +1040,10 @@ wxSize wxSizer::GetMinSize()
     return ret;
 }
 
-void wxSizer::DoSetMinSize( int width, int height )
+void wxSizer::DoSetMinSize(int width, int height)
 {
-    m_minSize.x = width;
-    m_minSize.y = height;
+	m_minSize.x = width;
+	m_minSize.y = height * 1.6; // Hackfix: There is no real fixed height, workaround for general settings where BIOS / Folders / GS Window / Speedhacks and others reside in.
 }
 
 bool wxSizer::DoSetItemMinSize( wxWindow *window, int width, int height )
--- a/3rdparty/wxwidgets3.0/src/common/sizer.cpp
+++ b/3rdparty/wxwidgets3.0/src/common/sizer.cpp
@@ -1043,7 +1043,7 @@ wxSize wxSizer::GetMinSize()
 void wxSizer::DoSetMinSize(int width, int height)
 {
 	m_minSize.x = width;
-	m_minSize.y = height * 1.6; // Hackfix: There is no real fixed height, workaround for general settings where BIOS / Folders / GS Window / Speedhacks and others reside in.
+	m_minSize.y = height;
 }
 
 bool wxSizer::DoSetItemMinSize( wxWindow *window, int width, int height )
--- a/common/include/Utilities/wxGuiTools.h
+++ b/common/include/Utilities/wxGuiTools.h
@@ -257,7 +257,6 @@ void operator+=(wxWindow &target, WinType &src)
         return;
     *target.GetSizer() += src;
 }
-
 template <typename WinType>
 void operator+=(wxWindow &target, const pxWindowAndFlags<WinType> &src)
 {
--- a/common/include/Utilities/wxGuiTools.h
+++ b/common/include/Utilities/wxGuiTools.h
@@ -257,6 +257,7 @@ void operator+=(wxWindow &target, WinType &src)
         return;
     *target.GetSizer() += src;
 }
+
 template <typename WinType>
 void operator+=(wxWindow &target, const pxWindowAndFlags<WinType> &src)
 {
--- a/pcsx2/gui/Dialogs/FirstTimeWizard.cpp
+++ b/pcsx2/gui/Dialogs/FirstTimeWizard.cpp
@@ -1,5 +1,5 @@
 /*  PCSX2 - PS2 Emulator for PCs
- *  Copyright (C) 2002-2010  PCSX2 Dev Team
+ *  Copyright (C) 2002-2021  PCSX2 Dev Team
  *
  *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
  *  of the GNU Lesser General Public License as published by the Free Software Found-
@@ -227,6 +227,6 @@ void FirstTimeWizard::OnPageChanging( wxWizardEvent& evt )
 
 void FirstTimeWizard::OnPageChanged( wxWizardEvent& evt )
 {
-	if( ((sptr)evt.GetPage() == (sptr)&m_page_bios) )
-		m_panel_BiosSel.OnShown();
+	//	if( ((sptr)evt.GetPage() == (sptr)&m_page_bios) )			// FirstTimeWizard panel which prevents the user for clicking on finish to the main PCSX2 menu , it has the warning message:
+	//	m_panel_BiosSel.OnShown();									// .SetUserMsg(pxE(L"Please select a valid BIOS.  If you are unable to make a valid selection then press Cancel to close the Configuration panel."));
 }
--- a/pcsx2/gui/Dialogs/FirstTimeWizard.cpp
+++ b/pcsx2/gui/Dialogs/FirstTimeWizard.cpp
@@ -228,5 +228,5 @@ void FirstTimeWizard::OnPageChanging( wxWizardEvent& evt )
 void FirstTimeWizard::OnPageChanged( wxWizardEvent& evt )
 {
 	if( ((sptr)evt.GetPage() == (sptr)&m_page_bios) )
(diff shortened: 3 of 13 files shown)


RedDevilus:
On my PC:

image
FirstTimeSetup
image

kleinkkk (approved):

kozarovv:
I'm not sure that is good idea to put two least used options on top of column.

RedDevilus:
Yeah and it's a mess to work correctly. But im bot against putting at bottom which is what i originally did.

Shoegzer:
Nice idea @RedDevilus.  Just a thought, but perhaps it's worth considering not having a separate section just for BIOS.  In other words, move the "BIOS Search Path" option to the bottom of the folders section and simply require the user to select a BIOS ROM file vs. a path.  This would at least keep all "folder/file selection" options in one place while retaining consistency with most other emulators in the way BIOS files are chosen.  The "Select a BIOS ROM" window could be retained if it was absolutely necessary though I wouldn't miss it.  Thoughts?

RedDevilus:
I'd rather have something like what DobieStation has:
Give a set path for browse+ pick bios.

I constantly have to browse to my correct path whenever i get development builds. I've noticed there are bigger issues lying beneath the BIOS / folder selector which makes it very hard to me and would need some help with the multi inheritance situation. I did increase the height as temporary fix so it could technically fit but WX just scares me and it will be torn out when QT comes so i have some thinking to do.

Shoegzer:
Fair points, they make sense to me.  And moving the BIOS selector to the folders area could be just food for thought until the Qt move anyway, I don't blame you at all for being WX-gunshy. ;)

3rdparty/wxwidgets3.0/src/common/sizer.cpp
```diff
@@ -1040,10 +1040,10 @@ wxSize wxSizer::GetMinSize()
     return ret;
 }
 
-void wxSizer::DoSetMinSize( int width, int height )
+void wxSizer::DoSetMinSize(int width, int height)
 {
-    m_minSize.x = width;
-    m_minSize.y = height;
+	m_minSize.x = width;
+	m_minSize.y = height * 1.6; // Hackfix: There is no real fixed height, workaround for general settings where BIOS / Folders / GS Window / Speedhacks and others reside in.
```
GovanifY, line 1046: no no no NO. no 3rdparty patches please damnit

GovanifY (changes requested):
LGTM code wise, others will have to approve on the design though; I don't mind either way.

Except for that 3rdparty patch. What even is this? Why do you add a hack on wxwidgets??

3rdparty/wxwidgets3.0/src/common/sizer.cpp
RedDevilus: Look at this problem: https://github.com/PCSX2/pcsx2/pull/4580

RedDevilus: Buttons like ok and cancel are dpi sensitive and restore restore defaults just dissapears. It's also bad on other languag

GovanifY: see how we handle multi-DPI scaling here: https://github.com/PCSX2/pcsx2/blob/master/pcsx2/gui/Dialogs/AboutBoxDialog.cpp#L38

also either way the bug isn't the issue here, the solution you're proposing is the issue! there is approximately 0% of chance seeing your pr merged with a 3rdparty hack

RedDevilus: I'll fix it then.

vsub:
That's nice but it makes the window longer
How about place that side panel at the top(my poor photoshop attempt)and maybe add the memory cards too so less dialogs
Untitled-1

lightningterror:
I believe it used to be at the top in the past but had issues?

kleinkkk (approved):

lightningterror:
Status?

RedDevilus:
Still need to fix the height factor in a non-hacky way.

lightningterror:
> Still need to fix the height factor in a non-hacky way.

Hmm, if the issue is also present on master we can still merge the current changes anyway.

refractionpcsx2:
> > Still need to fix the height factor in a non-hacky way.
> 
> Hmm, if the issue is also present on master we can still merge the current changes anyway.

The hack requires modifying WX, and we don't really want to do that.

lightningterror:
> > > Still need to fix the height factor in a non-hacky way.
> > 
> > 
> > Hmm, if the issue is also present on master we can still merge the current changes anyway.
> 
> The hack requires modifying WX, and we don't really want to do that.

I meant excluding the hackfix, which can be fixed in a different pr later on if it's an issue on master as well.

RedDevilus:
I suppose, WX is smart enough to add sliders for the icons so it's a minor inconvenience for now. I'll remove the hackfix.

RedDevilus:
I have removed the hackfix.

Hackfix:
image

Current:
image

Keep in mind there is some differences depending on resolution and DPI.

3rdparty/wxwidgets3.0/src/common/sizer.cpp
```diff
@@ -1040,10 +1040,10 @@ wxSize wxSizer::GetMinSize()
     return ret;
 }
 
-void wxSizer::DoSetMinSize( int width, int height )
+void wxSizer::DoSetMinSize(int width, int height)
 {
-    m_minSize.x = width;
-    m_minSize.y = height;
+	m_minSize.x = width;
+	m_minSize.y = height;
```
lightningterror, line 1046: Need to undo formatting changes too.

pcsx2/gui/Dialogs/FirstTimeWizard.cpp
```diff
@@ -228,5 +228,5 @@ void FirstTimeWizard::OnPageChanging( wxWizardEvent& evt )
 void FirstTimeWizard::OnPageChanged( wxWizardEvent& evt )
 {
 	if( ((sptr)evt.GetPage() == (sptr)&m_page_bios) )
-		m_panel_BiosSel.OnShown();
+	    m_panel_BiosSel.OnShown();
```
lightningterror, line 231: Spaces.

RedDevilus:
Formatting should be fine now

common/include/Utilities/wxGuiTools.h
```diff
@@ -257,7 +257,6 @@ void operator+=(wxWindow &target, WinType &src)
         return;
     *target.GetSizer() += src;
 }
-
```
lightningterror, line 260: This too, useless.

RedDevilus: Fixed

pcsx2/gui/Dialogs/SysConfigDialog.cpp
```diff
@@ -192,7 +192,7 @@ void Dialogs::SysConfigDialog::Cancel()
 }
 
 Dialogs::SysConfigDialog::SysConfigDialog(wxWindow* parent)
-	: BaseConfigurationDialog( parent, AddAppName(_("Emulation Settings - %s")), 580 )
+	: BaseConfigurationDialog( parent, AddAppName(_("General Settings - %s")), 600 )
```
lightningterror, line 195: Hm, why the 600 change?

RedDevilus: Testing but forgot

pcsx2/gui/Panels/PathsPanel.cpp
```diff
@@ -79,3 +74,13 @@ Panels::StandardPathsPanel::StandardPathsPanel( wxWindow* parent )
...
 
+void Panels::StandardPathsPanel::Apply()
+{
+
+}
+
+void Panels::StandardPathsPanel::AppStatusEvent_OnSettingsApplied()
+{
+
+}
```
lightningterror, line 85: What purpose do these functions serve?

RedDevilus: This fixes the warning in first time wizard and the function to block users to not press next until they provide a valid bios.

RedDevilus: It relies on the fuction  onselbios, a cursor or atleast an update to the screen where you pick a bios.

So it handles this:
image

Whoever created this made it hardcoded and obfuscated.

GovanifY: Yep, I can confirm those are necessary, they are assumed to be defined by inheritance. Other windows have them dummied out too; blame wx/air

lightningterror (approved):
Looks good, just need one last round of testing.

pcsx2/gui/Dialogs/SysConfigDialog.cpp
```diff
@@ -202,6 +202,8 @@ Dialogs::SysConfigDialog::SysConfigDialog(wxWindow* parent)
 
 	//NOTE: all pages which are added to SysConfigDialog must be of class BaseApplicableConfigPanel or derived.
 	//		see comment inside UpdateGuiForPreset implementation for more info.
+	AddPage<BiosSelectorPanel>		( pxL("BIOS"),			cfgid.Cpu );
+	AddPage<StandardPathsPanel>		( pxL("Folders"),		cfgid.Paths );
```
tadanokojin, line 206: formatting

RedDevilus: image

How is it displaying it like that?

tadanokojin: tabs sucking at aligning things maybe, it's not really a huge issue.

tadanokojin (approved):
LGTM. @GovanifY can you verified your requested change has been made?

GovanifY (approved):
Yep, all good!

PR #4629, opened 2021-07-27
GameDB: Ratchet and Clank (EEtiming)

### LINKED ISSUE #4650: [BUG] Ratchet and Clank Crashing on When Loading Quartu
I load the game, the loading cutscene plays, then right as the starship approaches the Planet the game stops, the sounds of the landing play, and then the game crashes. Only happens in 1.7.0, downloaded the latest version.

RedDevilus:
Fixes #4650
### Description of Changes

Add EETiming to Ratchet and clank to avoid TLB misses.
### Rationale behind Changes

Improvements ofcourse.
### Suggested Testing Steps

Run the right serial with and without automatic gamefixes.

### DIFF
--- a/bin/GameIndex.yaml
+++ b/bin/GameIndex.yaml
@@ -175,6 +175,7 @@ SCAJ-20001:
   region: "NTSC-Unk"
   gameFixes:
     - VUKickstartHack # Fixes Character SPS.
+    - EETimingHack # Fixes SPR errors while going in-game.
 SCAJ-20002:
   name: "Gallop Racer 6 - Revolution"
   region: "NTSC-Unk"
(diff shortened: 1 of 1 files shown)


lightningterror:
@atomic83GitHub can you confirm it's needed since this game is your forte?

atomic83GitHub:
I tested the PAL version a lot, I couldn't manage to trigger TLB misses at least in my version (which is 2.0 so there might have been fixes since). But I will test more regardless and I can't test NTSC as I don't have it.

refractionpcsx2:
I tried and failed to reproduce it :(

lightningterror:
Status?

RedDevilus:
Should be fine for merge.

lightningterror:
Was it confirmed that it's needed?

RedDevilus:
Seems to fix tlb issues for people so i assume so. Probably speed changes affected those games more but not in the past.

refractionpcsx2:
I did some investigation work with atomic and the Ratchet games all write to the DMA registers while it's still busy without the EE Timing Hack, so it's good to do this to be safe.

PR #4632, opened 2021-07-29
GS-d3d11: Cleanup fxaa and external shader.

lightningterror:
### Description of Changes

Instead of relying on bools for checks rely on the pixel shader pointer, also merge the functions.
### Rationale behind Changes

Clean stuff up.
### Suggested Testing Steps

Make sure FXAA and External shader work properly  on Direct3D11


tadanokojin (approved):
LGTM

I don't have any idea about the original bug this was trying to fix. Sounds like it was specific to vs 2010.

PR #4637, opened 2021-08-01
GS-hw: Adjust StretchRect for rgba write.

lightningterror:
### Description of Changes

Rename the function to StretchRectWRGBA, better since there are multiple.
Wrap the code and check if we actually use any channels first so we can run it.
### Rationale behind Changes

Minor tweaks, sanity checks.
### Suggested Testing Steps

Test Sonic Unleashed will be only affected, also make sure it compiles.


tellowkrinkle:
I don't like the idea of making our own magic numbers that just happen to match DX11 and OpenGL's.  If you want to merge the values, do it like how we do blend modes, where the number uses our own struct definition which is then converted separately by each backend to its own format.

lightningterror:
You have a point, I adjusted the dx values so we check the bits instead, as for gl I kept the new changes.

PR #4638, opened 2021-08-01
Debugger: Add option to display VU0F registers as floats

F0bes:
### Description of Changes
Add an option to the VU0f context menu that allows you to toggle the float view
Add a title to the VU0f register list that identifies the registers w/z/y/x subsections

### Rationale behind Changes
Reading float values as hex sucks

### Suggested Testing Steps
Play around with the VU0f register section

### Screenshots
New VU0F hex view
image

New VU0F float view
image

Display as float context menu
image

Display as float context menu (when display as float is disabled)
image

Also supports 32 and 64 bit views!
32 bit
image
64 bit
image

## Notes:
Also, large floating point numbers are kind of weird.
If anyone has a solution (maybe using %g as the format specifier?) I couldn't get a good result however.

Closes #2779


F0bes:
Ok I made it not shit

PR #4642, opened 2021-08-03
PGIF: Code refactoring/cleanup.

kozarovv:
### Description of Changes

- Fixed GP1(00h), ps1drv do most job for us.
- Fixed GP1(01h)
- Fixed GP1(04h), since ps1drv do nothing here, we need to care about all possible options.
- Fixed BCR BA 0 transfers (really 0x10000, based on duckstation solution)
- Fixed BCR BA decrementing, previously BA was decremented by 1 every transfered word. This is wrong (1BA = 1BS, which can be up to 10 words for GPU).
- Fixed From GPU DMA transfers using wrong addressess. That one improve many games.
- Cleaned up getUpdPgifCtrlReg to just update fifo state before returning result. Previous implementation was really overdone. 
- Moved DREQ (DMA Request bit) handling from getUpdPgpuStatReg to handleGp1Command. Since this bit is dependent on GP1(04h) it seems to be best to take care of that as soon as GP1(04h) is read by GPU.
- Other getUpdPgpuStatReg cleanups, there is still something wrong that we care here about RSEND (Vram to IOP data ready), but for now it need to be as is (close to old behavior).
- Removed #define CMD_INTERCEPT_AT_WR #define CMD_INTERCEPT_AT_RD flags. CMD_INTERCEPT_AT_RD stays as correct behavior.
- Converted else if to case switches where possible for better readability.
- Changed cryptic names to something more readable.
- Implemented structs/unions to allow bitfields instead of anding/oring whole registers everytime.
- Enabled IRQ1 in GP0(1Fh), hardware do this regardless of data, so we should do the same.
- Fixed GP0(E3h - E4h) to mask more bits (19 vs 20) since we are emulating old GPU model (based on no$ info). 
- Removed STAT_DIR_BY_CMD_BITS, no need for it since proper handling is there.
- Removed #define PGPU_STAT_IRQ1 0x01000000, now handled on pgpu struct bits access, and always enabled. 
- Removed experimental functions not strictly related to pgif/pgpu. Many of then didn't even had chance to work on pcsx2 without implementing more stuff. Backup here: https://github.com/PCSX2/pcsx2/blob/dc4b382b2022cf2c910570444c0fac7a0846c17f/pcsx2/ps2/pgif.cpp#L961
- Adjusted logs, and errors. Added few "when error happen" default solutions. 
- Added warnings for backward memory address stepping (is unknown that is possible on gpu).
- Removed pgifReset, there is no need to reset pgif/pgpu in that way, and it was wrong anyway (pgpu after reset should be set to 0x14802000, not 0, etc.) Now ps1drv take care of that.
- Remove operating on CHCR Trigger bit. This is manual bit for syncmode 0, we shouldn't touch it on GPU DMA according to available docs. 
- Remove psxDma2GpuR CHCR busy=0 hack, we should now handle that correctly.
- Disable INTC speed hack for PS1 games, further improve few games like Castlevania SOTN. (by refraction)

### Rationale behind Changes
Improve code, improve PS1 games compatibility. 
### Suggested Testing Steps
Playing PS1 games, and comparing to latest devbuild. Most importantly, since there was no good way to do that PR in few commits. Test for any single regression, because that need to be tracked right now as code base change completely in many places.

### DIFF
--- a/pcsx2/ps2/pgif.cpp
+++ b/pcsx2/ps2/pgif.cpp
@@ -1,5 +1,5 @@
 /*  PCSX2 - PS2 Emulator for PCs
-*  Copyright (C) 2016-2016  PCSX2 Dev Team
+*  Copyright (C) 2016-2021  PCSX2 Dev Team
 *  Copyright (C) 2016 Wisi
 *
 *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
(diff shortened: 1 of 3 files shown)


atomic83GitHub:
This PR brings a massive improvement across all of my games. Graphics artifacts are gone for the most part on many of them. Of course, there are plenty of issues remaining as expected but so far, it's impressive.

pcsx2/ps2/pgif.cpp
```diff
@@ -86,983 +33,733 @@ All the PS1 GPU info comes from psx-spx: http://problemkaputt.de/psx-spx.htm
...
+			// Pgpu reset, mostly handled by ps1drv (00201D64 in ps1drv). Comment for case 1 apply here too.
+			ringBufferClear(&rb_gp0);
+			break;
+		case 1: //Reset GP0 fifo, seems to check that something is empty, so maybe we should do the same before clear?
+			ringBufferClear(&rb_gp0);
+			break;
+		case 2: //Acknowledge GPU IRQ
+			ackGpuIrq1();
+			break;
+		case 4: //DMA Direction / Data Request. The PS1DRV doesn't care of this... Should we do something on pgif ctrl?
```
refractionpcsx2, line 251: Could potentially be linked to the SIF2 FIFO direction, however the SIF will also have the direction bits set by the DMA channels also, so this may have no effect at all.  Could be worth noting though.

atomic83GitHub:
Here are some example to illustrate the improvements:

- Spiderman

Before:
bad1

After:
good1

- Final Fantasy 4 (Anthology)

Before:
bad2 FF4

After:
good2 ff4

- Space Invaders

Before:
bad3

After:
good3

- Resident Evil

Before:
bad4

After:
good4

- Street Fighter EX Plus Alpha

Before:
bad5

After:
good5

PR #4645, opened 2021-08-05
CDVD: Fix CdlReadS for PS1 games with CDDA.

kozarovv:
### Description of Changes
Fix mistake from https://github.com/PCSX2/pcsx2/pull/3899 , where path for PS1 games with CDDA was changed to real CD audio path. PR should affect only PS1 games.
### Rationale behind Changes
Fix Legend of Dragoon (PS1), combined with (#4642) fix Tenchu. Maybe more.

### Suggested Testing Steps
Testing PS1 games

### DIFF
--- a/pcsx2/CDVD/CdRom.cpp
+++ b/pcsx2/CDVD/CdRom.cpp
@@ -926,7 +926,7 @@ void cdrWrite1(u8 rt)
 			break;
 
 		case CdlReadS:
-			if (cdvd.Type == CDVD_TYPE_CDDA) // Taken from pcsxr
+			if (cdvd.Type == CDVD_TYPE_PSCDDA) // Taken from pcsxr
 				goto do_CdlPlay;
 			cdr.Irq = 0;
 			StopReading();


Ziemas:
Alongside this the original PR added: https://github.com/kozarovv/pcsx2/blob/9d9b0434868da4db1c983490fd5c386d9e277127/pcsx2/CDVD/CdRom.cpp#L855

Is that still valid?

kozarovv:
> Alongside this the original PR added: https://github.com/kozarovv/pcsx2/blob/9d9b0434868da4db1c983490fd5c386d9e277127/pcsx2/CDVD/CdRom.cpp#L855
> 
> Is that still valid?

Good catch, to be honest i missed that. But correcting it to PS CDDA break games, and not only ones that were fixed earlier. 
Looking at code that inspired this line, it seems that pcsxr just stops cdda playback if mode is not PSX CDDA without forcing any disc type here. So most likely this code need to be fixed later, and most likely it will break something as is.

refractionpcsx2:
> > Alongside this the original PR added: https://github.com/kozarovv/pcsx2/blob/9d9b0434868da4db1c983490fd5c386d9e277127/pcsx2/CDVD/CdRom.cpp#L855
> > Is that still valid?
> 
> Good catch, to be honest i missed that. But correcting it to PS CDDA break games, and not only ones that were fixed earlier.
> Looking at code that inspired this line, it seems that pcsxr just stops cdda playback if mode is not PSX CDDA without forcing any disc type here. So most likely this code need to be fixed later, and most likely it will break something as is.

is that right though? This is mode, not type, so it's probably different.

kenshen112:
But yeah, when I wrote that I didn't realize we had CDDA vs PSCDDA. Sorry! 😅

kozarovv:
Well now its forcing CD Audio disc/type, if mode is CDDA everytime that CdlSetmode is triggered with MODE_CDDA bit active.  So is wrong for sure. I'm just not sure what should be done here yet. Pcsxr solution seems to be logical, just stop cdda here if mode IS NOT ps cdda, and don't do anything else with type/mode.

https://github.com/mirror/pcsxr/blob/51b2ad7c2bdc83ed6a0f297679233bc4d19de027/libpcsxcore/cdrom.c#L1362

PR #4646, opened 2021-08-06
CDVD: Implement SCMD 0x27

kozarovv:
### Description of Changes
Implement SCMD 0x27 to allow PS1 games to boot on SCPH-50009, and DTL-H50009 bios.
According to krat0s this command should return PS1 Disc Serial.
Disc Serial is later used by ps1drv itself to find matching game config. 

Note: While DTL consoles are known to use region free PS1 bios, SCPH-50009 use Asia/Japan PS1 bios, and region lock apply here.
### Rationale behind Changes
Fix PS1 mode for SCPH-50009, and DTL-H50009.

### Suggested Testing Steps
Launch PS1 game using BIOS dumped from SCPH-50009, or DTL-H50009.


tadanokojin (approved):
LTGM

PR #4652, opened 2021-08-09
Random CMake improvements

tellowkrinkle:
### Description of Changes
- Removes our special `add_pcsx2_xxx` commands in favor of using the cmake builtin commands
- Removes leftover `Utilities_NO_TLS` from when we had plugins
- Adds support for using the `cpp` version of `bin2cpp`

### Rationale behind Changes
CMake cleanup and preparation for Windows support in CMake

### Suggested Testing Steps
Make sure building still works


arcum42:
Builds as 32-bit and 64-bit in Linux and still starts up. Looks like more modern CMake than previously. I'm good with this.

PR #4657, opened 2021-08-11
github: Switch to new issue forms

xTVaser:
### Description of Changes

Github has improvements to the issue template system that has been in beta for a while.  They recently started making marketing videos promoting it, so I expect the schema has largely stabilized.

They look pretty good - check them out here https://github.com/xTVaser/pcsx2-rr/issues/new/choose

### Rationale behind Changes

These are better for a few reasons imo:
- Basic validation - minimize half-filled out issues
- Way better formatting, can actually use markdown and space things out instead of having a giant block of non-interactive comments
- I wouldn't be too surprised if there is an API integration at some point on the read side -- each form field has a unique `id` -- would be potentially useful if issues became more machine readable in the future.

### Suggested Testing Steps

Go here - https://github.com/xTVaser/pcsx2-rr/issues/new/choose


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,153 @@
...
+  - type: textarea
+    id: repro
+    attributes:
+      label: Reproduction Steps
+      description: "Steps to reproduce the behavior"
+      value: |
+        1.
+        2.
+        3.
+        ...
```
lightningterror, line 44: Don't think the values are needed, looks much cleaner without.

tadanokojin (approved):
This is fine for me and I figure we can sort out any minor changes in follow up PRs. That fine by you @lightningterror ?

RedDevilus (approved):
image

There is already a function which does the numbered list, you just need to press it and atleast 1 character then press enter to get `2.` and so on. Just a minor thing anyway.

PR #4664, opened 2021-08-14
GS-GUI: Add 7xMultiplier

RedDevilus:
### Description of Changes

7x was bullied for far too long (2520P / 4.5K resolution) is added for end-users.
### Rationale behind Changes

Beyond 8x is debatable if there is a benefit or maximum framebuffer sizes that the API/PC/emulator can allow.
### Suggested Testing Steps

Open GS settings and change to 7x, then boot a game. Watching titlebar resolution or even the program log will help.

### DIFF
--- a/pcsx2/GS/GS.cpp
+++ b/pcsx2/GS/GS.cpp
@@ -1525,9 +1525,10 @@ void GSApp::Init()
 	m_gs_upscale_multiplier.push_back(GSSetting(2, "2x Native", "~720p"));
 	m_gs_upscale_multiplier.push_back(GSSetting(3, "3x Native", "~1080p"));
 	m_gs_upscale_multiplier.push_back(GSSetting(4, "4x Native", "~1440p 2K"));
-	m_gs_upscale_multiplier.push_back(GSSetting(5, "5x Native", "~1620p 3K"));
+	m_gs_upscale_multiplier.push_back(GSSetting(5, "5x Native", "~1620p"));
 	m_gs_upscale_multiplier.push_back(GSSetting(6, "6x Native", "~2160p 4K"));
-	m_gs_upscale_multiplier.push_back(GSSetting(8, "8x Native", "~2880p 5K"));
+	m_gs_upscale_multiplier.push_back(GSSetting(7, "7x Native", "~2520p"));
+	m_gs_upscale_multiplier.push_back(GSSetting(8, "8x Native", "~2880p"));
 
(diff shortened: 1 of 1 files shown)


RedDevilus:
image

tadanokojin:
Honestly I think we should just drop the "K" post fixes (or remove all but 4K). I mean what is 4.5K? 1440p isn't 2K to begin with, it's sort of a colloquialism to call it that but it's really QHD. I just think outside of 4K (which I don't think is technically correct either because it's based on horizontal pixels) they're probably adding more confusion that anything.

Thoughts?

refractionpcsx2:
I guess that's just a "here's an idiots version of roughly what the resolution is", so in that respect, it's useful, but yes it is lies :P

tadanokojin:
Well yeah I see that point with 4K at least, but 4.5K is imaginary and 3K and 5K are not even things that most people would be familiar with (if that was the goal of adding them)

RedDevilus:
2K on monitors is 1440P or can be 1080P  ( 2048 × 1080).
Well besides this 'definition' the game doesn't even follow it exactly i saw some go beyond their multiplier rated resolution.
Though there is a 4.5K Retina display from Apple :P

tadanokojin:
I think the point of adding "4K" was that more people are familiar with the term "4K" than they are with exactly what resolution it is so the additional context is helpful. In the case of all the others though, I don't think that additional context adds anything. Who's really ever heard of 5K or 3K? And I think it looks kind of dumb to stick 4.5K next to something like anyone has any idea what they would look like :P

RedDevilus:
Fair enough i have only seen it briefly in stores or even in school. Should i remove the 4.5K mention?

refractionpcsx2:
out of those only 2K and 4K are the established marketing ones used, that I know of

tadanokojin:
I think you can remove 4.5K and I think you can do 3K and 5K while you're there. 4K and 2K are not technically accurate but I think I can live with them more than I can the others :P

RedDevilus:
To be honest i always looked at the 3K and 5K, oh well i suppose it's one of those things to get used to.

RedDevilus:
image

slx7R4GDZM:
@tadanokojin 
>Who's really ever heard of 5K or 3K?

LG advertises 5k monitors. Not sure other companies really do though.

RedDevilus:
Besides Apple, hardly any.

PR #4666, opened 2021-08-15
Fix crashes in pad with libc++

tellowkrinkle:
### Description of Changes
Fixes crash on PAD shutdown when using libc++

### Rationale behind Changes
Crashes are not fun, even if they happen when the emulator is shutting down

### Suggested Testing Steps
Compile with libc++, and try to close the emulator


arcum42 (approved):
Seems fine to me.

DamonGamerking:
Vishal Vka

PR #4670, opened 2021-08-15
add Linux distro information to program log 

Mrlinkwii:
### Description of Changes

Adds Distro information to  the program log for linux 

Master:
image

PR:
image

### Rationale behind Changes

It helps finding out what  distro a linux user is using if they need help with a particular issue 
### Suggested Testing Steps

run the PR or use the appimage on a linux OS , make sure i havent broken windows

### DIFF
--- a/common/src/Utilities/Linux/LnxMisc.cpp
+++ b/common/src/Utilities/Linux/LnxMisc.cpp
@@ -24,13 +24,13 @@
 // Returns 0 on failure (not supported by the operating system).
 u64 GetPhysicalMemory()
 {
-    u64 pages = 0;
+	u64 pages = 0;
 
 #ifdef _SC_PHYS_PAGES
-    pages = sysconf(_SC_PHYS_PAGES);
+	pages = sysconf(_SC_PHYS_PAGES);
 #endif
 
-    return pages * getpagesize();
+	return pages * getpagesize();
 }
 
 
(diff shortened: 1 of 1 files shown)


pcsx2/System.cpp
```diff
@@ -210,8 +210,14 @@ void SysLogMachineCaps()
...
 
-	Console.WriteLn( Color_StrongBlack, "Host Machine Init:" );
-
+	Console.WriteLn( Color_StrongBlack, "Host Machine Init:"); 
+	
+	#ifdef __unix__
+	Console.Indent().WriteLn(
+        L"Distro info      =  %s",
+        WX_STR(wxGetLinuxDistributionInfo().Description)
+    );
```
refractionpcsx2, line 219: Should be tabs and the close bracket is misaligned.

tadanokojin:
How about instead of
```
distro = ubuntu 20.04
operating system = Linux 5.11.0-25-generic blah blah
```

you just do

```
operating system = Ubuntu 20.04 (Linux 5.11.0-25-generic x86_64)
```

Since a kernel is not an OS anyway.

Or you can just drop the kernel string and put
```
operating system = Ubuntu 20.04
```

Since I don't think we care what kernel you run anyway.

RedDevilus:
Ifdeffed but it affects Windows but now i see you basically replaced the function on which affects windows
image

Duplicate the old code with other ifdef for windows.

pcsx2/System.cpp
```diff
@@ -211,12 +211,21 @@ void SysLogMachineCaps()
...
+		L"Operating System =  %s",
+    	WX_STR(wxGetLinuxDistributionInfo().Description)
+    );
+	#endif
+	#ifdef __WIN32__
+	Console.Indent().WriteLn(
+		L"Operating System =  %s\n",
+		WX_STR(GetOSVersionString())
+	);
+	#endif
```
tadanokojin, line 226: going to take a guess and say this breaks macos/bsd.
Don't you think something like this might be clearer?
```cpp
wxString os_string;
#ifdef __linux__
os_string = wxGetLinuxDistributionInfo().Description;
#else
os_string = GetOSVersionString();
#endif

Console.Indent().WriteLn(
	L"Operating System =  %s\n"
	L"Physical RAM     =  %u MB",
	os_string,
	(u32)(GetPhysicalMemory() / _1mb)
);
```

tadanokojin: better yet maybe just modify this function?
https://github.com/PCSX2/pcsx2/blob/6f7890b709d5e3f7f5b824781e493455efc92339/common/src/Utilities/Linux/LnxMisc.cpp#L53-L56

tadanokojin (approved):

common/src/Utilities/Linux/LnxMisc.cpp
```diff
@@ -52,7 +52,7 @@ u64 GetCPUTicks()
 
 wxString GetOSVersionString()
 {
-    return wxGetOsDescription();
+  return wxGetLinuxDistributionInfo().Description;
```
lightningterror, line 55: Re add the 2 spaces, tho ideally would be nice to clang format the file.

PR #4671, opened 2021-08-15
Debugger: Purge 64 / 32 bit register view options

F0bes:
### Description of Changes
Remove the option to view > 32 bit registers as 32 bit or 64 bit and instead display the _entire_ register value

### Rationale behind Changes
There is no need to hide the upper bits of the registers, it also _greatly_ simplifies how the register list is drawn (or sized to be specific)

### Suggested Testing Steps
Launch a game, go to the debugger, and within the register view, play around with the categories, enabling and disabling resolving of string pointers and displaying VU0F registers as floats.

### DIFF
--- a/pcsx2/gui/Debugger/CtrlRegisterList.cpp
+++ b/pcsx2/gui/Debugger/CtrlRegisterList.cpp
@@ -76,7 +76,6 @@ CtrlRegisterList::CtrlRegisterList(wxWindow* parent, DebugInterface* _cpu)
 	}
 
 	SetDoubleBuffered(true);
-	SetInitialSize(ClientToWindowSize(GetMinClientSize()));
 
 	const wxSize actualSize = getOptimalSize();
 	SetVirtualSize(actualSize);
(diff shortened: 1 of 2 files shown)


F0bes:
Ok, I'm pleased with the results, and now I don't need to install another distro in order to use the debugger.

Ready for review / merge.

RedDevilus:
32 bit:
image
64 bit:
image
128 bit:
image
128 bit + resolve:
image

F0bes:
Instead of just making the debugger register list a static size, I removed the 32/64 bit view functionality and will now make it so all bytes in a register are visible
Please refer to the original post for the new PR description.

image
image
image
image

Mrlinkwii (approved):

pcsx2/gui/Debugger/CtrlRegisterList.cpp
```diff
@@ -304,23 +284,32 @@ void CtrlRegisterList::OnDraw(wxDC& dc)
 					{
 						char str[256];
 						const u128 val = cpu->getRegister(category, i);
-						// Use std::bit_cast in C++20. The below is technically UB
-						sprintf(str, "%7.2f %7.2f %7.2f %7.2f", *(float*)&val._u32[3], *(float*)&val._u32[2], *(float*)&val._u32[1], *(float*)&val._u32[0]);
-						dc.DrawText(wxString(str), x, y + 2);
+						// Print each register part individually so we know exactly where each part starts
+						// and use this information for printing the VU0F titles
+						for(int j = 3; j >= 0; j--)
```
lightningterror, line 289: Formatting.

F0bes: Resolved.

RedDevilus (approved):
Can't wait till it beats Ghidra.

PR #4672, opened 2021-08-17
Remove gui/ directory from target-wide includes

stenzek:
### Description of Changes
Remove `gui/` from the target include directory, replacing includes in files outside of it with paths including `gui/`.

### Rationale behind Changes
There's a decent amount of dependence on GUI classes/functions in the emulation core. This PR removes `gui/` from the target's include path, to make it clearer where we need to look at moving things to make it less dependent on the UI, since we can simply search for `gui/` includes.

### Suggested Testing Steps
Make sure it builds on all platforms. I've checked Windows and Linux 64-bit.

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -15,7 +15,7 @@
 
 #include "PrecompiledHeader.h"
 #include "IopCommon.h"
-#include "AppConfig.h"
+#include "gui/AppConfig.h"
 
 #include <memory>
 #include <ctype.h>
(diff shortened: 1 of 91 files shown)


tadanokojin (approved):
Sounds good to me

lightningterror (approved):

tellowkrinkle (commented):
IIRC we prefer e.g. `gui/` over `../` for places where both are valid

stenzek:
Rebased and changed `../` to `gui/` where appropriate. This is the base of https://github.com/stenzek/pcsx2/tree/settings which I'll PR soon, just need to move save states out of the GUI, which is the base for adding new host abstractions (eventually leading to a Qt UI).

PR #4683, opened 2021-08-21
Various GSBlock shuffle improvements

tellowkrinkle:
### Description of Changes
A bunch of minor improvements to GSBlock deswizzles
Nets fairly large improvements in GSBenchmark, much smaller improvements in actual games (since there's also everything else running).  Biggest improvement I noticed is in CPU palette conversion (a few percent in Baldur's Gate).  Intel Skylake+ may see larger improvements in those situations from the new use of `vpgatherdd`, but I don't have a Skylake+ CPU to test.

### Rationale behind Changes
speed

### Suggested Testing Steps
Test various games, make sure graphics aren't all glitchy, see if anything got any faster
Stare at the CI and make sure unit tests pass

### DIFF
--- a/pcsx2/GS/GSBlock.cpp
+++ b/pcsx2/GS/GSBlock.cpp
@@ -19,10 +19,13 @@
 CONSTINIT const GSVector4i GSBlock::m_r16mask(0, 1, 4, 5, 8, 9, 12, 13, 2, 3, 6, 7, 10, 11, 14, 15);
 CONSTINIT const GSVector4i GSBlock::m_r8mask(0, 4, 2, 6, 8, 12, 10, 14, 1, 5, 3, 7, 9, 13, 11, 15);
 CONSTINIT const GSVector4i GSBlock::m_r4mask(0, 8, 4, 12, 1, 9, 5, 13, 2, 10, 6, 14, 3, 11, 7, 15);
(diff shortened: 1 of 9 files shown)


lightningterror:
Would be nice to rename commit prefix from gsdx to gs.

gilderoylockhart01:
There actually is a game where those improvements result in a quite significant performance boost.

Oni during the level introduction screens runs at a stupidly high frame-rate because the game runs completely uncapped and it can go as high as 400+ fps on real hardware. That same screen causes the GSDX performance to plumet (as well as other emulators, it's a game "bug").

On your PR GSDX seems to handle the massive workload much better, I can sustain 100% speed at 2x with my 3070 and 9900k cpu. Possibly at 3x too as I have a bit of headroom still.

image

refractionpcsx2:
Mostly I think it's just margin of error differences for me, however Warhammer seems to lose roughly 2fps, and considering my 3900x and 2080Ti can barely hold full speed, every FPS counts there.

See if you get similar results vs master 2290
[warhammer.zip](https://github.com/PCSX2/pcsx2/files/7969655/warhammer.zip)

refractionpcsx2:
Nevermind, it looks just like a difference in the CI builds.

PR #4684, opened 2021-08-21
Remove MakeUnique.h, which is unused.

arcum42:
### Description of Changes
Removes an unused header file.

### Rationale behind Changes
It's one less header in Utilities, and if we need make_unique, it is in C++14.

### Suggested Testing Steps
Compiling, mostly.


arcum42:
Or, you know, maybe I misread this and the reason it still works is my compiler supports c++14. :P 

Does pcsx2 actually compile in straight C+11?

Ziemas:
Pretty sure PCSX2 is c++17 at this point.

Ziemas:
See https://github.com/PCSX2/pcsx2/commit/5858f72748d0ec0810fc3c570c9e1803be54f76f

arcum42:
I *thought* we were past c++11, but wasn't absolutely sure. Still stands as something we can get rid of, then...

PR #4685, opened 2021-08-21
Cmake: Fix ENABLE_VTUNE build.

Ziemas:
Utilities needs to include/link Vtune. 

Also update the search paths to the modern setup for intel tools.


lightningterror (approved):
Was tested by Fobes that it works fine.

Issue #3568, opened 2020-07-26
GSdx Hack Tracking Thread


tadanokojin:
We have many hacks. Many have been lost to time and their function is unknown. Others can be patched using out patching system rather than relying on CRC hacks. This is a thread dedicated to researching/replacing/working around GSdx CRC hacks.

**Testers:** Please provide before and after images for anything you notice different between CRC hack levels.
**Patch authors:** Please attempt to replace any applicable CRC hacks with patches.

**Big Mutha Truckers**
Skips an extra texture shuffle call
- [ ] patch available
- [ ] fixed
- [ ] workaround

**Bully**
Broken upscaling causing blur
- [ ] patch available
- [ ] fixed
- [ ] workaround

**DBZ: BT3**
Upscaling misalignment
- [ ] patch available
- [ ] fixed
- [ ] workaround

**Death By Degrees**
Broken shuffle effect, broken upscaling, possible broken channel effect, last hack unknown
see: https://forums.pcsx2.net/Thread-GSDX-Texture-Cache-Bug-Report-Death-By-Degrees-SLUS-20934-NTSC
- [ ] patch available
- [ ] fixed
- [ ] workaround

**Ghost in The Shell**
Channel effect
- [ ] patch available
- [ ] fixed
- [ ] workaround

**God of War 2**
Speed hack that removes shadows. Possible haze/upscaling issue. Upscaling issue on water.
- [ ] patch available
- [ ] fixed
- [ ] workaround

**Wild Arms Games**
Unknown. Appears to skip calls relating to to the color attachment PSM being Z32.
- [ ] patch available
- [ ] fixed
- [ ] workaround

**Manhunt 2**
Channel effect. Gamma correction similar to GT4 but page-based.
- [ ] patch available
- [ ] fixed
- [ ] workaround

**Crash: Wrath of the Cortex**
Unknown.
- [ ] patch available
- [ ] fixed
- [ ] workaround

**Sacred Blaze**
Unknown. Listed a "rendering glitches"
- [ ] patch available
- [ ] fixed
- [ ] workaround

**Spartan**
Skips depth sampling
- [ ] patch available
- [ ] fixed
- [ ] workaround

**Ikki Tousen/Battle Vixens**
Upscaling blur
- [ ] patch available
- [ ] fixed
- [ ] workaround

**Genji**
Listed as "blur" and "white fog". Additional DX only hack for shuffling, may not be needed.
- [ ] patch available
- [ ] fixed
[...]

PR #4687, opened 2021-08-24
gs: add nightshade undub crc

swosho:
Snice the list already contains some CRCs of the modified executables, there's no reason for the undubbed versions not to be included as well when necessary.

### DIFF
--- a/pcsx2/GS/GSCrc.cpp
+++ b/pcsx2/GS/GSCrc.cpp
@@ -307,6 +307,7 @@ CRC::Game CRC::m_games[] =
 	{0x1F8640E0, TalesOfLegendia, JP, 0},
 	{0xE4F5DA2B, TalesOfLegendia, KO, 0},
 	{0x519E816B, Kunoichi, US, 0}, // Nightshade
+	{0x160B66C8, Kunoichi, US, 0}, // Undub
 	{0x3FB419FD, Kunoichi, JP, 0},
 	{0x086D198E, Kunoichi, CH, 0},
 	{0x3B470BBD, Kunoichi, EU, 0},


tadanokojin:
I don't accept CRCs for undubs or protos.

With regards to already including CRCs for modified versions, I I've gone ahead and just removed the ones I could spot here https://github.com/PCSX2/pcsx2/commit/a68d834f8f6c9dcf64d3cc2755c54752e740d011 so if there are any others let me know and I'll remove them as well. I'd rather remove them all than start accepting new ones.

swosho:
> I don't accept CRCs for undubs or protos.

Can I inquire why? Aren't they completely harmless? I can't imagine the maintainers of any other emulator ever actively gimping the modded titles support without a clear reason.

I'm cool with helping to nuke the remaining unofficial CRCs for consistency's sake. I only want it to be more transparent that it's the unwillingness to accept new CRCs to the list being the reason why some fan translated games, undubs, etc., have some visual glitches present on the hardware renderer.

RedDevilus:
Better to make patches (pnachs) than relying on CRC hacks. While visually the same it will seperate the hacks from the core and perhaps improve upon them by not being complacent for a quick dirty fix.

swosho:
Exactly. Making the CRC hacks setup (maybe optionally) external would've been a step in the right direction.

RedDevilus:
That's what the GameDatabase is for and the cheats folder. You might also want to read this: https://github.com/PCSX2/pcsx2/issues/3568

swosho:
There's the probable eventual solutions, fair enough.
Still, speaking for the here and now I can't comprehend why unofficial CRCs shouldn't be listed.

RedDevilus:
Example: A game only worked fine on CRC hack level, it was removed, suddenly they made a patch in a day. It's more a longterm vision than shortterm. Besides we don't support stuff like romhacks and if someone comes along and need support how are we supposed to help? I understand that as user you want to put as many CRC hacks. But it's a way to grow to be less hacky.

TL;DR
image

swosho:
>I understand that as user you want to put as many CRC hacks. But it's a way to grow to be less hacky.

Except I'm not advocating for putting in more CRC hacks, I only want the existing hack to be whitelisted for use with a specific modded version of the game.
Besides, if you remove the hack – it's gone for every whitelisted variant of the same game, official or otherwise.

In case the actual reason for gatekeeping is this:

>Besides we don't support stuff like romhacks and if someone comes along and need support how are we supposed to help?

Then perhaps you're simply not obliged to help with the unofficial / tampered versions of the games??
Is it worth it to actively cripple the romhacks compatibility over that?

F0bes:
> Then perhaps you're simply not obliged to help with the unofficial / tampered versions of the games??

Yes, it makes giving support infinitely easier and prevents people blaming the emulator for bad romhacks, because it's not supported.

> Is it worth it to actively cripple the romhacks compatibility over that?

It's either we cater to all non-official mods or none at all, and as shown by https://github.com/PCSX2/pcsx2/commit/a68d834f8f6c9dcf64d3cc2755c54752e740d011, it'll be the latter.

swosho:
Sabotaging the compatibility over such petty arbitrary reasons is mad. Change a single byte in the executable and you're SOL just because the PCSX2 maintainers said so.

> It's either we cater to all non-official mods or none at all, and as shown by a68d834, it'll be the latter.

In striking contrast, the actual PS2 hardware "caters by default" to all non-official mods and runs them all the same as the non-modded counterparts of those games.

tadanokojin:
> Sabotaging the compatibility over such petty arbitrary reasons is mad. 

Thinking adding a line that enables a hack that patches out rendering is an improvement to "compatibility" is mad. You're not improving the compatibility of anything. You're just making it look less broken to most people, it's the illusion of compatibility. You've fixed nothing. There's been zero improvement.

> Change a single byte in the executable and you're SOL just because the PCSX2 maintainers said so.

So change a few extra bytes and improve the "compatibility".

> In striking contrast, the actual PS2 hardware "caters by default" to all non-official mods and runs them all the same as the non-modded counterparts of those games.

In striking contrast the PS2 doesn't require nopping out draws to improve "compatibility" but you're welcome to do that and explain whoever consumes this undub on real hw your logic on this one.

If all you're going to do is make accusations of pettiness and gatekeeping instead of addressing the arguments being presented then I don't see any reason to continue this discussion further. It's clearly not going anywhere.

PR #4691, opened 2021-08-28
gs: purge clut reload before draw hack

tadanokojin:
### Description of Changes
remove this stupid hack

### Rationale behind Changes
I said I would remove it if people keep trying to use it and someone suggested I add it to the gui so I'm removing it instead.

### Suggested Testing Steps
don't test anything today, especially not my patience.

### DIFF
--- a/pcsx2/GS/GS.cpp
+++ b/pcsx2/GS/GS.cpp
@@ -1620,7 +1620,6 @@ void GSApp::Init()
 	m_default_configuration["capture_threads"]                            = "4";
 	m_default_configuration["CaptureHeight"]                              = "480";
 	m_default_configuration["CaptureWidth"]                               = "640";
-	m_default_configuration["clut_load_before_draw"]                      = "0";
 	m_default_configuration["crc_hack_level"]                             = std::to_string(static_cast<int8>(CRCHackLevel::Automatic));
 	m_default_configuration["CrcHacksExclusions"]                         = "";
 	m_default_configuration["debug_glsl_shader"]                          = "0";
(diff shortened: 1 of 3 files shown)


refractionpcsx2 (approved):

DragonautMk:
I'm serious, I don't understand a programming bite, but when I read the things you write I die laughing, in a good way. Keep it up!

Silanda:
This hack didn't work properly if rendering threads was set to anything other than zero, but it was necessary for the Sega Model 2 ports (Virtua Fighter 2, Fighting Vipers, Sega Rally, etc) in order to avoid incorrectly coloured textures. Is there an alternative solution now?

refractionpcsx2:
> This hack didn't work properly if rendering threads was set to anything other than zero, but it was necessary for the Sega Model 2 ports (Virtua Fighter 2, Fighting Vipers, Sega Rally, etc) in order to avoid incorrectly coloured textures. Is there an alternative solution now?

we have some code which fixes the actual problem, we're just looking at it/reviewing it and trying to improve it before it gets submitted.

refractionpcsx2:
As for the Sega Ages games, I can't reproduce them needing the clut reload before draw, I've tried with and without it and it's the same, HW is black and white and software mode is fine.  The games run super slow too..

Silanda:
There are texture imperfections in the Sega Ages games (and Sonic the Fighters in the Sonic Gems pack) in software mode. At least there are on my system. It looks like some polygons are just the wrong colour. 

The hack fixed that but only with rendering threads set to zero (which makes Fighting Vipers rather marginal performance wise for me). Set to anything else the graphics looked identical to without the hack, although the hack was obviously still doing something given that it could break other games if I accidentally left it on.

refractionpcsx2:
No idea, I just tried 0 threads and it didn't make a difference to me :/

Silanda:
How strange... To be fair, I don't think I've tested the hack on a build since GSdx was merged into the core.

refractionpcsx2:
I mean that shouldn't make any difference.  But unless the Sega Ages games I have are different from the ones you have, i dunno. I mean these games DO have CLUT issues in hardware mode at least, but that's unrelated to the hack.

PR #4693, opened 2021-08-28
Disable gl threaded optimization

kenshen112:
### Description of Changes

Disable gl threaded optimizations in Linux

### Rationale behind Changes

Due to recent issues with Nvidia driver in Linux breaking the optimization. Pcsx2 has not run correctly with it enabled

### Suggested Testing Steps

Run Nvidia Linux and compare master to this performance-wise. Preferably in rolling release distros with the latest drivers

### DIFF
--- a/linux_various/PCSX2.desktop.in
+++ b/linux_various/PCSX2.desktop.in
@@ -9,7 +9,7 @@ GenericName[zh_CN]=PlayStation 2 模拟器
 Comment=Sony PlayStation 2 emulator
 Comment[ru]=Эмулятор Sony PlayStation 2
 Comment[zh_CN]=索尼 PlayStation 2 模拟器
-Exec=env GDK_BACKEND=x11 __GL_THREADED_OPTIMIZATIONS=1 mesa_glthread=true MESA_NO_ERROR=1 PCSX2
+Exec=env GDK_BACKEND=x11 MESA_NO_ERROR=1 PCSX2
 Icon=PCSX2
 Keywords=game;emulator;
 Categories=@PCSX2_MENU_CATEGORIES@


tadanokojin:
`__GL_THREADED_OPTIMIZATIONS=0 mesa_glthread=true`

just delete.

tadanokojin:
`mesa_glthread=true`

delete

kenshen112:
> `__GL_THREADED_OPTIMIZATIONS=0 mesa_glthread=true`
> 
> just delete.

Done

kenshen112:
Should I drop the MESA_NO_ERROR as well?
```

MESA_NO_ERROR

if set to 1, error checking is disabled as per KHR_no_error. 
This will result in undefined behavior for invalid use of the API, 
but can reduce CPU use for apps that are known to be error free.
```

tadanokojin:
no

Trevonn:
If the issue is with the NVIDIA driver why was `mesa_glthread=true` removed? That only affects the Mesa driver correct?

tadanokojin:
not needed so I had him remove while he was here

Trevonn:
Okay I understand and sorry for deleting my comment I misread the commit and thought I had spoken in error

qurious-pixel:
A little late, but should this also be applied for `PCSX2-linux.sh` ?

https://github.com/PCSX2/pcsx2/blob/b61619cf4026f91ef05d70161dac1f5b5a4ea382/linux_various/PCSX2-linux.sh#L61

tadanokojin:
taken care of here https://github.com/PCSX2/pcsx2/commit/64cc1b106ceabb2bb74c101beff5114cfbdc0cd2

PR #4695, opened 2021-08-29
vs: misc fixup

tadanokojin:
### Description of Changes
some random cleanup for vs

### Rationale behind Changes
just cleaning up

### Suggested Testing Steps
if it builds it builds


refractionpcsx2:
I'm not sure on which projects you changed from release to devel in devel profiles, I don't know which one it is but one of the 3rd party things causes 64bit to crash if you build anything but release, no idea why.

We don't need debug stuff for 3rd party tools anyways, I'm not interested in debugging other peoples shit :P

Anyway, just figured I'd mention it in case that was one of the things that changed.

refractionpcsx2:
Compiled all the 32bit projects from start to finish, no errors. LGTM

refractionpcsx2 (approved):

PR #4701, opened 2021-08-31
App: Add Windows exe version information

refractionpcsx2:
### Description of Changes
Adds PCSX2 version information to windows executables

### Rationale behind Changes
Windows gets kinda fussy about exe's with no signature and no version information or description, this makes it a little less fussy.

Right now it shares version number with the rest of the app, but it cannot get the GIT revision because linux hates me and I can't think of a way to do it, but honestly 1.7.0 is enough information for the EXE properties.

### Suggested Testing Steps
Make sure it compiles, make sure the exe loads, check the windows properties->details for the exe, make sure it says PCSX2 PS2 emulator and version 1.7, all that fun stuff

### DIFF
--- a/pcsx2/CMakeLists.txt
+++ b/pcsx2/CMakeLists.txt
@@ -73,6 +73,7 @@ endif()
 if(WIN32)
 	# Resources
 	target_sources(PCSX2 PRIVATE
+		PCSX2.rc
 		GS/GS.rc
 		PAD/Windows/PAD.rc
 		SPU2/Windows/SPU2.rc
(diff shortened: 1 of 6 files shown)


RedDevilus:
Master:
image

PR:
image

CookiePLMonster (approved):

tadanokojin (commented):
LGTM but needs a commit for cmake. I'll add it later if Tellow doesn't beat me to it.

RedDevilus:
Cmake:
image

PR #4703, opened 2021-09-02
MicroVU: Skip VU1 instructions on VU0

F0bes:
### Description of Changes
Unlike the interpreters, mVU shares an instruction table (is it a table?) and currently VU1 only instructions could execute on VU0.
With this change, on pass 1 in MicroVU IR it'll check if we are running in the VU0 rec and if so, NOP the instruction.

The instructions were
`xgkick`
All EFU instructions (including `waitp` and `mfp`)
`xtop`
### Rationale behind Changes
Accuracy

### Suggested Testing Steps
[vu0xgkick.elf.zip](https://github.com/PCSX2/pcsx2/files/7095299/vu0xgkick.elf.zip)
Run this program, it uploads data to VU1 data memory and then calls `xgkick` on VU0.
In the current, broken, implementation your screen should turn blue. In a proper implementation it'll be black.

I haven't developed tests for EFU and TOP instructions, as I believe that would be unnecessary.

### DIFF
--- a/pcsx2/x86/microVU_Lower.inl
+++ b/pcsx2/x86/microVU_Lower.inl
@@ -1,5 +1,5 @@
 /*  PCSX2 - PS2 Emulator for PCs
- *  Copyright (C) 2002-2010  PCSX2 Dev Team
+ *  Copyright (C) 2002-2021  PCSX2 Dev Team
  *
  *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
  *  of the GNU Lesser General Public License as published by the Free Software Found-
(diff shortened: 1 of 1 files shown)


pcsx2/x86/microVU_Lower.inl
```diff
@@ -1229,6 +1231,8 @@ mVUop(mVU_XTOP) {
 
 mVUop(mVU_XITOP) {
 	pass1 {
+		if (isVU0) { mVUlow.isNOP = true; return; }
```
refractionpcsx2, line 1234: XITOP is valid on VU0

refractionpcsx2:
You should probably add WAITP and MFP too

F0bes:
Included `MFP` and removed `XITOP`

PR #4709, opened 2021-09-03
GS Include Cleanup

tellowkrinkle:
### Description of Changes
- Move preprocessor macros to a header that's included by everything, to prevent accidents with `ifdef`
- Remove custom countof macros in favor of `std::size` (or for-in loops / `std::fill` where applicable)
- Replace GS's integer typedefs with the ones used by the rest of PCSX2 (`uint32`, `int64` → `u32`, `s64`)
  - Note: A number of structs had conflicting members (e.g. `GSVector4::u32`), I fixed that by renaming those to use capital letters
- Split `GS.h` into multiple smaller headers
- Stop pretending `GS.h` is a PCH

Note: Builds upon #4708

### Rationale behind Changes
- Cleaner, more consistent code
- Better compile times and less files recompiled per change in a header

### Suggested Testing Steps
Make sure everything builds


common/include/Pcsx2Defs.h
```diff
@@ -121,64 +113,68 @@ static const int __pagesize = PCSX2_PAGESIZE;
...
-#define unlikely(x) __builtin_expect(!!(x), 0)
+	#ifndef __packed
+		#define __packed __attribute__((packed))
+	#endif
+	#ifndef __aligned
+		#define __aligned(alig) __attribute__((aligned(alig)))
+	#endif
+	#define __aligned16 __attribute__((aligned(16)))
+	#define __aligned32 __attribute__((aligned(32)))
+	#define __pagealigned __attribute__((aligned(PCSX2_PAGESIZE)))
```
arcum42, line 151: While _aligned16 and _pagealigned are used, _aligned32 isn't actually used anywhere in the code that I see. _aligned(32) and _aligned(64) are used a few times...

tellowkrinkle: I replaced them all with `alignas` since we have C++11 now

arcum42: Sounds good to me.

common/Pcsx2Defs.h
```diff
@@ -104,64 +83,57 @@ static const int __pagesize = PCSX2_PAGESIZE;
...
 #ifdef _MSC_VER
 
-// Using these breaks compat with VC2005; so we're not using it yet.
-//#	define __pack_begin		__pragma(pack(1))
-//#	define __pack_end		__pragma(pack())
-
-// This is the 2005/earlier compatible packing define, which must be used in conjunction
-// with #ifdef _MSC_VER/#pragma pack() directives (ugly).
-#define __packed
+	// Using these breaks compat with VC2005; so we're not using it yet.
```
seta-san, line 86: is visual studio 2005 still an option?

tellowkrinkle: Gone

Mrlinkwii (approved):

PR #4711, opened 2021-09-03
Clean up Linux CI dependencies

tellowkrinkle:
### Description of Changes
Cleans up unneeded dependencies on the Linux CI

### Rationale behind Changes
Faster builds, less messy looking dependency list

### Suggested Testing Steps
Make sure appimage builds still run

### DIFF
--- a/.github/workflows/scripts/linux/generate-cmake.sh
+++ b/.github/workflows/scripts/linux/generate-cmake.sh
@@ -3,8 +3,8 @@
 set -e
 
 if [ "${COMPILER}" = "gcc" ]; then
-  export CC=gcc-8
-  export CXX=g++-8
+  export CC=gcc-10
+  export CXX=g++-10
 else
   export CC=clang
   export CXX=clang++
--- a/.github/workflows/scripts/linux/generate-cmake.sh
+++ b/.github/workflows/scripts/linux/generate-cmake.sh
@@ -3,8 +3,8 @@
 set -e
 
 if [ "${COMPILER}" = "gcc" ]; then
-  export CC=gcc
-  export CXX=g++
+  export CC=gcc-8
+  export CXX=g++-8
 else
   export CC=clang
   export CXX=clang++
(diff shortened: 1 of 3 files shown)


.github/workflows/scripts/linux/generate-cmake.sh
```diff
@@ -3,8 +3,8 @@
 set -e
 
 if [ "${COMPILER}" = "gcc" ]; then
-  export CC=gcc
-  export CXX=g++
+  export CC=gcc-8
+  export CXX=g++-8
```
Margen67, line 7: Why not use gcc 9 or 10? https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#language-and-runtime

tellowkrinkle: We're on 18.04
IIRC it made the appimage easier

Margen67: 18.04 also has gcc 9 and 10: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md#language-and-runtime

qurious-pixel:
GCC-10 will cause some breakage with older systems. 
https://github.com/darealshinji/AppImageKit-checkrt works around this issue, and should be implemented if moving to GCC-10.

tellowkrinkle:
In that case I'll stick to gcc-8

If someone else wants to implement that fix they can

Margen67:
> GCC-10 will cause some breakage with older systems.
> https://github.com/darealshinji/AppImageKit-checkrt works around this issue, and should be implemented if moving to GCC-10.

What about GCC 9?

qurious-pixel:
> > GCC-10 will cause some breakage with older systems.
> > https://github.com/darealshinji/AppImageKit-checkrt works around this issue, and should be implemented if moving to GCC-10.
> 
> What about GCC 9?

Ubuntu 20.04 and the equivalent will work, since gcc-9 is bundled. 
https://github.com/PCSX2/pcsx2/pull/4711/commits/635bdf36aee40b054fe9782173b007a019d0a413 works, with the patched runtime. Tested on Ubuntu 18.04 with gcc-7.

refractionpcsx2:
Okay so is this PR going to be put back to GCC 8? aside from that are we good to merge since a few things now require the newer clang to get around some issues.

qurious-pixel:
Pushed a fix for the custom config file location would not appear until the second boot. Separated out environment files into their own script. (located in AppRun hooks folder)

PR #4712, opened 2021-09-04
Fix up some CMake defines

tellowkrinkle:
### Description of Changes
Edits CMake defines to better match the vs projects
Fixes missing portaudio backends

### Rationale behind Changes
Make CMake builds better match VS builds

### Suggested Testing Steps
See if portaudio has devices


tadanokojin (approved):
LTGM

PR #4718, opened 2021-09-05
GS: Replace our own format iostream with using fmt lib

lightningterror:
### Description of Changes

Removes/replaces our own format iostream with fmt lib.
### Rationale behind Changes

Cleaner, less complex.
### Suggested Testing Steps

Test if HW/SW renderers are working properly.

### DIFF
--- a/pcsx2/GS/GSCapture.h
+++ b/pcsx2/GS/GSCapture.h
@@ -22,6 +22,8 @@
 #include "Window/GSCaptureDlg.h"
 #endif
 
+#include "fmt/core.h"
+
 class GSCapture
 {
 	std::recursive_mutex m_lock;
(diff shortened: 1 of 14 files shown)


pcsx2/GS/GSCrc.cpp
```diff
@@ -17,6 +17,8 @@
 #include "GS.h"
 #include "GSCrc.h"
 
+#include "fmt/core.h"
```
tadanokojin, line 20: use `#include <...>` for externals/system includes

pcsx2/GS/GSCapture.h
```diff
@@ -20,6 +20,8 @@
 
 #ifdef _WIN32
 #include "Window/GSCaptureDlg.h"
+#else
+#include "fmt/core.h"
```
tadanokojin, line 24: just include normally, it's not going to fail to compile or anything.

lightningterror: I didn't because it's not used on windows here.

tadanokojin: I understand. Just include it anyway.

pcsx2/GS/GSCrc.cpp
lightningterror: It's a 3rdparty submodule, won't compile on linux/mac otherwise.

F0bes (approved):
HW and SW modes seem to work fine on x64 linux.
Tested with both homebrew and Colin McRae Rally 3.

RedDevilus (approved):
SW/HW renderers work fine on Windows, so does the internal video recorder. Does need a final rebase :)

PR #4722, opened 2021-09-05
pad-windows: Always use Xinput 1.4.

lightningterror:
### Description of Changes

Removes Xinput 1.3 alternative for W7.
This also removes compatibility with SCP Toolkit.
### Rationale behind Changes

Windows 7 is no longer supported.
### Suggested Testing Steps

Test pad if it still works on windows.

### DIFF
--- a/pcsx2/PAD/Windows/XInputEnum.cpp
+++ b/pcsx2/PAD/Windows/XInputEnum.cpp
@@ -292,15 +292,8 @@ void EnumXInputDevices()
 		if (xinputNotInstalled)
 			return;
 
-		// Prefer XInput 1.3 since SCP only has an XInput 1.3 wrapper right now.
-		// Also use LoadLibrary and not LoadLibraryEx for XInput 1.3, since some
-		// Windows 7 systems have issues with it.
-		// FIXME: Missing FreeLibrary call.
-		HMODULE hMod = LoadLibrary(L"xinput1_3.dll");
-		if (hMod == nullptr && IsWindows8OrGreater())
-		{
-			hMod = LoadLibraryEx(L"XInput1_4.dll", nullptr, LOAD_LIBRARY_SEARCH_APPLICATION_DIR | LOAD_LIBRARY_SEARCH_SYSTEM32);
-		}
+
+		HMODULE hMod = LoadLibraryEx(L"XInput1_4.dll", nullptr, LOAD_LIBRARY_SEARCH_APPLICATION_DIR | LOAD_LIBRARY_SEARCH_SYSTEM32);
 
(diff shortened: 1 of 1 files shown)


Squall-Leonhart:
Pointless change,  you don't use the audio component of the xbox control and thats all 1.4 is included for.

tadanokojin:
There's nothing wrong with using 1.4. We only maintain 1.3 for win7 and SCP both of which are dead and unmaintained.

Squall-Leonhart:
There are still people using SCP on 10 because of a range of issues that can happen with latter controller emulator software, not to mention folks that are still using the dll drop for X360CE instead of running it in Vigem mode

RedDevilus:
SCP has been replaced by https://forums.pcsx2.net/Thread-DsHidMini-Windows-10-driver-for-the-DualShock-3

tadanokojin:
Pretty sure you can just change the name of the dll.
Actually X360CE (the old version, the new one doesn't use this method according to them) recommends doing exactly that <https://www.x360ce.com/>
```
Run this program from the same directory as the game executable. XInput library files exist with several different names and some games require a change in its name. Known names:

• xinput1_4.dll
• xinput1_3.dll
• xinput1_2.dll
• xinput1_1.dll
• xinput9_1_0.dll
```

seta-san:
> Pointless change, you don't use the audio component of the xbox control and thats all 1.4 is included for.

sounds like a nice feature for the Sega seamic controller

tadanokojin:
> Are we still supporting windows 8.1? What version does it support?

https://docs.microsoft.com/en-us/windows/win32/xinput/xinput-versions
> XInput 1.4 ships today as a system component in Windows 8 as XINPUT1_4.DLL. It is available “inbox” and does not require redistribution with an application. The Windows Software Development Kit (SDK) contains the header and import library for statically linking against XINPUT1_4.DLL. To download the Windows 8 SDK, see Downloads for developing desktop apps.

https://walbourn.github.io/xinput-and-windows-8/

> Windows 8.x and Windows 10 include version 1.4 of the XInput API for use with Xbox Common Controller compatible game devices, and this API is fully supported for both Windows Store apps (including x86, x64, and Windows on ARM) and desktop Win32 applications.

RedPanda4552:
I say ship it, but if we're going to kill SCP we may consider removing the references to the extended SCP library while we're at it, otherwise we're orphaning those items with this.

CookiePLMonster:
If only `xinput1_4` is to be used, might as well link to that DLL statically instead of loading it explicitly via `LoadLibrary`.

lightningterror:
Suppressed by #4781

Issue #2090, opened 2017-10-07
Ratchet Gladiator: EmotionEngine interpreter crash the emulator ingame.


ghost:
***PCSX2 version:***
Last 1.5 dev

***PCSX2 options:***
All default, EE set to interpreter (with and without the cache), speedhacks disabled

***Plugins used:***
All default

***Description of the issue:***
Set the Emotion Engine to interpreter cause an emulator crash ingame and in loading screen.

***How to reproduce the issue:***
Load "Ratchet Gladiator" with EE set to interpreter and go ingame

***Last known version to work:***
Don't exist, was broken before.

***PC specifications:***
Intel Core i7 4770k  GTX 560Ti  Windows 10 64 Bit

MrCK1:
The EE interpreter is only used for diagnosis and the game already works in the recompiler.. I don't think there's much point in this issue, so I'm going to close it for now.

atomic83GitHub:
And by the way, interpreter crash on the ONI game with a PCSX2 DMA error.
![image](https://user-images.githubusercontent.com/26351275/31314647-bb798f1e-ac05-11e7-95ac-0458abb2bdba.png)

atomic83GitHub:
And by the way, it seems to happen in all Ratchet and Clank games

TheTechnician27:
Can confirm that this is still an issue for me in _Deadlocked_  (tested in 1.7.4110). The game actually doesn't crash during cutscenes or in the menu, but once you're in-game, it immediately crashes. Not only that, but you can't even exit out of the game or X out of PCSX2. The program actually will ask you if you want to exit out of the game (so it's not like it's entirely unresponsive), but if you select 'Yes', it will just sit there doing nothing forever. I had to manually either go to my task manager and kill the program or go to the terminal and run kill -9.

While I wouldn't call "a game runs poorly or not at all with EE recompilation off" necessarily an issue, I would call it an issue when it makes PCSX2 so unresponsive that you need to circumvent PCSX2's UI and personally tell your OS to take it out behind the shed.

PR #4726, opened 2021-09-06
Vu int improve

### LINKED ISSUE #1851: Burnout 3/R/D: VU1 Interpreter Breaks Various Effects
***PCSX2 version:***
1.5.0-dev-1921

***PCSX2 options:***
* VU1 set to Interpreter for the first screenshot
* VU1 set to MicroVU Recompiler for the second screenshot 

***Plugins used:***
GSdx D3D11 SW.

When VU1 is set to Interpreter, many of the effects become broken. One of those effects is the cars' exhaust fumes. An observation I made is that the result of setting VU1 to Interpreter is identical to what happens if you use MicroVU Recompiler and set the VU rounding mode to Positive, so it's probably a rounding issue.

gsdx_20170310232535
MicroVU Recompiler

gsdx_20170310232550
Interpreter

1. Get into a race.
2. Go to Config > Emulation Settings > VUs and set VU1 to Interpreter.

This behaviour has always been present AFAIK.

***PC specifications:***
CPU: Intel i5-6400
GPU: AMD Radeon HD 7770
RAM: 4GB DDR4 @ 2133MHz
OS: Windows 10 x64

refractionpcsx2:
### Description of Changes
Overhauls the VU interpreters to be as accurate as microVU, and then some.  Still some clean up work to do (could make the new XGKick stuff one function, for example), but should be head and shoulders above master and be twice as fast too.

### Rationale behind Changes
Our interpreters were old, unfinished, broken and slow, this PR aims to resolve all of that.

### Suggested Testing Steps
Just test any games, compare them to microVU in terms of compatibility, make sure they are as good or better.

Games fixes by this (only in the interpreters currently)
WRC World Championship no longer needs patches
Tennis Court Smash no longer has SPS

Fixes #1851

### DIFF
--- a/pcsx2/VUops.cpp
+++ b/pcsx2/VUops.cpp
@@ -108,12 +108,30 @@ static __ri bool _vuFMACflush(VURegs * VU) {
 				{
 				VUM_LOG("flushing FMAC pipe[%d] (macflag=%x status=%x) Cur Cycle %d Expected cycle %d", currentpipe, VU->fmac[currentpipe].macflag, (VU->VI[REG_STATUS_FLAG].UL & 0xF30) | (VU->fmac[currentpipe].statusflag & 0x3CF), VU->cycle, VU->fmac[currentpipe].sCycle + VU->fmac[currentpipe].Cycle);
 				// FMAC only affectx Z/S/I/O
-					VU->VI[REG_STATUS_FLAG].UL = (VU->VI[REG_STATUS_FLAG].UL & 0xFF0) | (VU->fmac[currentpipe].statusflag & 0x3CF);
+					VU->VI[REG_STATUS_FLAG].UL = (VU->VI[REG_STATUS_FLAG].UL & 0xFF0) | ((VU->fmac[currentpipe].statusflag & 0xF) | ((VU->fmac[currentpipe].statusflag & 0xF) << 6));
 					VU->VI[REG_MAC_FLAG].UL = VU->fmac[currentpipe].macflag;
 					lastmac = (VU->cycle - VU->fmac[currentpipe].sCycle);
 				}
 
 			}
+			didflush = true;
+		}
+	}
+	return didflush;
+}
+
+static __ri bool _vuIALUflush(VURegs* VU) {
+	bool didflush = false;
+
+	VUM_LOG("Flushing ALU stalls");
+
+	for (int i = 0; i < 8; i++) {
+		int currentpipe = i;
+
+		if (VU->ialu[currentpipe].enable == 0) continue;
+
+		if ((VU->cycle - VU->ialu[currentpipe].sCycle) >= VU->ialu[currentpipe].Cycle) {
+			VU->ialu[currentpipe].enable = 0;
 			didflush = true;
 		}
 	}
@@ -198,7 +216,7 @@ void _vuFlushAll(VURegs* VU)
 				else
 				{
 					// FMAC only affectx Z/S/I/O
-					VU->VI[REG_STATUS_FLAG].UL = (VU->VI[REG_STATUS_FLAG].UL & 0xFF0) | (VU->fmac[currentpipe].statusflag & 0x3CF);
+					VU->VI[REG_STATUS_FLAG].UL = (VU->VI[REG_STATUS_FLAG].UL & 0xFF0) | ((VU->fmac[currentpipe].statusflag & 0xF) | ((VU->fmac[currentpipe].statusflag & 0xF) << 6));
 					VU->VI[REG_MAC_FLAG].UL = VU->fmac[currentpipe].macflag;
 				}
 			}
@@ -229,18 +247,29 @@ void _vuFlushAll(VURegs* VU)
 			}
 		}
 
-		VU->cycle++;
+		for (i = 0; i < 8; i++) {
+			int currentpipe = i;
 
+			if (VU->ialu[currentpipe].enable == 0) continue;
+			nRepeat = 1;
+			if ((VU->cycle - VU->ialu[currentpipe].sCycle) >= VU->ialu[currentpipe].Cycle) {
+				VU->ialu[currentpipe].enable = 0;
+			}
+		}
+
+		VU->cycle++;
 	} while(nRepeat);
 }
 
 __fi void _vuTestPipes(VURegs * VU) {
 	bool flushed;
+	u32 startcycle = VU->cycle;
 	do {
 		flushed = false;
 		flushed |= _vuFMACflush(VU);
 		flushed |= _vuFDIVflush(VU);
 		flushed |= _vuEFUflush(VU);
+		flushed |= _vuIALUflush(VU);
 	} while (flushed == true);
 }
 
@@ -353,6 +382,34 @@ static __fi void _vuAddFMACStalls(VURegs * VU, _VURegsNum *VUregsn) {
 	_vuFMACAdd(VU, VUregsn);
 }
 
+static __ri void __fastcall _vuIALUAdd(VURegs* VU, _VURegsNum* VUregsn) {
+	int i;
+
+	if (VUregsn->cycles == 0)
+		return;
+	//If it's an FMAC which doesn't modify FMAC flags, just exit, no need to queue
+	/* find a free fmac pipe */
+	for (i = 0; i < 8; i++) {
+		if (VU->ialu[i].enable == 1) continue;
+		break;
+	}
+
+	if (i < 8) {
+		VUM_LOG("adding IALU pipe[%d]; reg=%x xyzw=%x flagreg=%x target=%x current %x", i, VUregsn->VFwrite, VUregsn->VFwxyzw, VUregsn->VIwrite, VU->cycle + 4, VU->cycle);
+		VU->ialu[i].enable = 1;
+		VU->ialu[i].sCycle = VU->cycle;
+		VU->ialu[i].Cycle = VUregsn->cycles;
+		VU->ialu[i].reg = VUregsn->VIwrite;
+	}
+	else {
+		Console.Error("*PCSX2*: error , out of ALU's %d", VU->cycle);
+	}
+}
+
+static __fi void _vuAddIALUStalls(VURegs* VU, _VURegsNum* VUregsn) {
+	_vuIALUAdd(VU, VUregsn);
+}
+
 static __fi void _vuTestFDIVStalls(VURegs * VU, _VURegsNum *VUregsn) {
 //	_vuTestFMACStalls(VURegs * VU, _VURegsNum *VUregsn);
 	_vuTestFMACStalls(VU, VUregsn);
@@ -383,12 +440,31 @@ __fi void _vuTestUpperStalls(VURegs * VU, _VURegsNum *VUregsn) {
 		case VUPIPE_FMAC: _vuTestFMACStalls(VU, VUregsn); break;
 	}
 }
+static __fi void _vuTestALUStalls(VURegs* VU, _VURegsNum* VUregsn) {
+	int i;
+
+	for (i = 0; i < 8; i++) {
+		if (VU->ialu[i].enable == 0) continue;
+		if ((VU->cycle - VU->ialu[i].sCycle) >= VU->ialu[i].Cycle) continue;
+		if (VU->ialu[i].reg & VUregsn->VIread) // Read and written VI regs share the same register
+			break;
+	}
+
+	if (i == 8) return;
+
+	u32 newCycle = VU->ialu[i].Cycle + VU->ialu[i].sCycle;
+
+	VUM_LOG("ALU[%d] stall %d", i, newCycle - VU->cycle);
+	if (newCycle > VU->cycle)
+		VU->cycle = newCycle;
+}
 
 __fi void _vuTestLowerStalls(VURegs * VU, _VURegsNum *VUregsn) {
 	switch (VUregsn->pipe) {
 		case VUPIPE_FMAC: _vuTestFMACStalls(VU, VUregsn); break;
 		case VUPIPE_FDIV: _vuTestFDIVStalls(VU, VUregsn); break;
 		case VUPIPE_EFU:  _vuTestEFUStalls(VU, VUregsn); break;
+		case VUPIPE_BRANCH: _vuTestALUStalls(VU, VUregsn); break;
 	}
 }
 
@@ -403,6 +479,7 @@ __fi void _vuAddLowerStalls(VURegs * VU, _VURegsNum *VUregsn) {
 		case VUPIPE_FMAC: _vuAddFMACStalls(VU, VUregsn); break;
 		case VUPIPE_FDIV: _vuAddFDIVStalls(VU, VUregsn); break;
 		case VUPIPE_EFU:  _vuAddEFUStalls(VU, VUregsn); break;
+		case VUPIPE_IALU:  _vuAddIALUStalls(VU, VUregsn); break;
 	}
 }
 
@@ -1809,8 +1886,6 @@ static __ri void _vuILW(VURegs * VU) {
 	u16 addr = ((imm + VU->VI[_Is_].SS[0]) * 16);
 	u16* ptr = (u16*)GET_VU_MEM(VU, addr);
 
-	_vuBackupVI(VU, _It_);
-
 	if (_X) VU->VI[_It_].US[0] = ptr[0];
 	if (_Y) VU->VI[_It_].US[0] = ptr[2];
 	if (_Z) VU->VI[_It_].US[0] = ptr[4];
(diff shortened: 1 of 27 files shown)


seta-san:
impressive to see tennis court smash look good.

gilderoylockhart01:
I tested several VU edge cases games and the testing results are very promising, however there are several things to report more or less related to the changes:

BGE demo: Implicit VU0-EE sync + kickstart, handled just fine now.

Capture2

Totally Spies: strict Mbit timings: The VU part is fine, however it's the EE now which becomes the issue.

Capture5

Mafia: The game hangs now if the VU1 int is active. This is not an issue with this PR, the fact it hangs before displaying the memory card screen is expected as it needs the VIF FIFO and instant VU1 bruteforce it's way into the problem, so that's a win still.

Enter the Matrix: same as Mafia, needs the VIF FIFO now to avoid ingame issues (ee timing hack kinda work here but it's an improper fix).

**Regression**

No luck, 'Find my own way' tech demo isn't happy with the new int, it now turns into a SPS mess and will crash the emulator if one VU is set to interpreter but not the other (massive sync??):

Capture

Also, Blue Shift games (Marvel Nemesis and co) doesn't work anymore on any VU interpreter. I am curious if it is possible now to introduce the "always sync" behaviour for those games as a toogle maybe given it's slow nature?

pcsx2/VUops.cpp
```diff
@@ -349,20 +477,19 @@ static __fi float vuADD_TriAceHack(u32 a, u32 b) {
...
-		DevCon.WriteLn("aExp = %d, bExp = %d", aExp, bExp);
-		DevCon.WriteLn("0x%08x + 0x%08x = 0x%08x", a, b, (u32&)ret);
-		DevCon.WriteLn("%f + %f = %f", vuDouble(a), vuDouble(b), ret);
-		return ret;
-	}
-	return vuDouble(a) + vuDouble(b);
+	// microVU uses something like this to get TriAce games working,
+	// but VU interpreters don't seem to need it currently:
+
+	// Update Sept 2021, now the interpreters don't suck, they do - Refraction
```
gilderoylockhart01, line 483: Well, the comment itself is fine but I think it's best to remove the old ones as in fact, the old broken code isn't here anymore if you see what I mean :).

RedDevilus:
Worth looking into:
https://github.com/PCSX2/pcsx2/issues/2090

gilderoylockhart01:
Ratchet and Clank VU issues are fixed too. However the EE issue persist.

refractionpcsx2:
Totally Spies needs EE Cycle Rate +3 for the game to work, this has always been the case with that one
image

As for Find My Own Way, that game is a nightmare fuel on toast, and doesn't work properly anyway, so I'm not going to get too hung up on it.

Blue shift engine, yep, I know, the game gets stuck doing a VU flush on the VIF through the FIFO, so I'm going to need to implement VIF FIFO to deal with that.

Glad the BGE demo is okay.

refractionpcsx2:
> Ratchet and Clank VU issues are fixed too. However the EE issue persist.

yes, I could add the rounding mode stuff to the EE interpreter, that will likely fix the Ratchet problems with the EE interpreter.

gilderoylockhart01:
Ok, anyway, EE issues are outside of the scope of this PR. Btw, sorry, there is an other regression which I just noticed, Shawn Murray Wakeboarding Unleashed, it black screens when going ingame.

refractionpcsx2:
Okay that's another game that does stupid stuff, I'm not going to mess up VU timings to make games that do stupid crap work.  There is a hack in VIF to deal with the fact Wakeboarding starts a VU program on VIF which first clears the VU memory, while VIF is unpacking data to it behind its clean up job. PCSX2 by default has already started the VU program by this point and it clears the data VIF has just uploaded, it's very stupid.

refractionpcsx2:
image

Wakeboarding on interpreter.

gilderoylockhart01:
You made it to work on it or it just works out of the box? If that's the case I tested the demo which might be different.

refractionpcsx2:
> You made it to work on it or it just works out of the box? If that's the case I tested the demo which might be different.

With the Vif hack I made especially for it, it works on this pr with no further changes

gilderoylockhart01:
Ok all is fine then. I tested Rayman Revolution just in case, the game is fine.

And one final question regarding what can be possible or not, for Powerdrome you can't make the VU1 floats to bypass x86 range?

refractionpcsx2:
Possibly, that's a future pr plan, to use doubles to extend our float limits. But there's a bunch of things I want to do and that's gonna be a big job

gilderoylockhart01:
Ok so, there is one game "Mercenaries 2" which have a strange behaviour when running the VU1 in int mode.

On the recompiler that game needs sync VU1 to fix distant flickering (gamedb have the timing hack which.. works but I don't think it's a proper fix), but with the interpreter it still flicker regardless.

refractionpcsx2:
when you say sync, do you mean disable instant VU?

gilderoylockhart01:
Yes exactly.

refractionpcsx2:
huh okay, well the changes I'm doing might help, I guess we'll see. I think I have mercenaries 2

gilderoylockhart01:
I forgot to test Resident Evil code Veronica X and as expected, it's bad. The game still require the kickstart gamefix even with the interpreters to fix the eyes of the man standing in front of the character:

Capture

refractionpcsx2:
sure, I mean kickstart may always be a thing, there's nothing I can do about that where we can only run one thing at a time and soft of be in sync

gilderoylockhart01:
But why some games like Ratchet and Clank doesn't need it anymore?

refractionpcsx2:
Ratchet should need it, as far as I know it isn't any different.

gilderoylockhart01:
Well, that's actually the case for some games other than Ratchet such as BGE, look by yourself on the first Ratchet for exemple:

Rec without gamefixes:

Capture2

int without gamefixes:

Capture

refractionpcsx2:
yeah for the Rec is what I mean, the VU Interpreter doesn't really use Kickstart, not like the Recompiler anyway.

gilderoylockhart01:
Maybe I am missing the point, but in the case of Resident Evil the interpreter still needs kickstart even when running booth VUs in the interpreter mode so apparently, the gamefix still does something in that game even in int mode.

What can actually cause such a difference between games regarding kickstart behaviour in int?

refractionpcsx2:
yes it will slightly adjust how frequently the VU's can be called. I'm interested in if my coming changes will resolve that somewhat..

refractionpcsx2:
Okay see if that's any better (once it compiles)

gilderoylockhart01:
No luck, booth Mercenaries 2 and Resident Evil are still bad.

refractionpcsx2:
Okay, as I said, if it needed the Kickstart before, it still will, I'm not fixing that, I was just fixing games which now needed it but didn't before.

gilderoylockhart01:
Ok so effectively, the games that needed it now are fine, except Ratchet which is still fine without it.

refractionpcsx2:
ok cool, thanks

seta-san:
anyone try world series baseball 2k3?

refractionpcsx2:
It won't work.

gilderoylockhart01:
PS: This is not a joke, ONI crashes on the VU1 interpreter now with a bunch of DMA errors:

Capture

refractionpcsx2:
Interesting, you're not using Instant DMA off or anything? 
how far in do you have to be in order to get the errors?

gilderoylockhart01:
I don't have any hacks enabled while testing the interpreters. 

In order to trigger the bug start the "TCTF HQ" level and when going ingame, the screen will else turns black with the VU0 rec and VU1 int or crashes soon after the cutsceen have started with booth VUs in int mode.

refractionpcsx2:
okay, I want you to have the default hacks enabled (recommended) that includes instant vu, since that being off does cause problems :P

but if it happens as soon as you go ingame, I can check that, otherwise I'll need a savestate (recommend you make it with VU JIT's)

gilderoylockhart01:
Enabling instant VU1 does solve the problem indeed, do you still want a savestate?

refractionpcsx2:
sure, why not

gilderoylockhart01:
There it is, use it with the blockdump inside this folder it should work.

https://drive.google.com/file/d/1elT2x5ycu8SDVoLUYKiUADy9jLG_iR-5/view

[ONI savestate.zip](https://github.com/PCSX2/pcsx2/files/7129320/ONI.savestate.zip)

refractionpcsx2:
Okay I see what's happening, because VIF is waiting for the XGKick to end and it doesn't loop when it's flushing, I need to add some code to check that when an XGKick finishes. 

I need to make some changes to the XGKick code anyway, so I'll drop a message here when I've posted the fix

gilderoylockhart01:
In fact, it needs the vif fifo to work, delay vif stalls just like Mafia make it to work.

refractionpcsx2:
No, that's just avoiding the timing break happening which stops VIF. I know what the problem is, it isn't the FIFO

gilderoylockhart01:
Ok, is the recompiler also impacted by this bug?

refractionpcsx2:
No, because XGKick is instant on the JIT.  The problem is that I've synced XGKick on the interpreters, so the XGKIck can run past the end of a VU program.

gilderoylockhart01:
Ok very cool stuff! I assume this sync XGkick will help with the OSDSYS and Blus Shift games as well.

refractionpcsx2:
Okay the ONI problem is fixed. Just an explanation of what happened

The game started a PATH3 transfer while a PATH1 (VU) transfer was in progress, so what our GIF Unit does is it buffers this transfer, so as far as the GIF DMA was concerned, it had finished.

However PATH3 relies on the DMA to finish the "Transfer" for PATH3 to get the APATH timing right (for games like Tomb Raider, and PATH3 masking stuff), but because the DMA had ended by the time VU1 had finished (and GIF Unit then transferred the queued PATH3 transfer), there was no DMA to finish it off, so PATH 3 was left being busy forever.

I've now added some code to check PATH3's status if it gets itself in to this situation again.

gilderoylockhart01:
There is an interesting note with the bios, loading it with instant VU1 makes the white towers not to show during the startup (as expected") but now, even VU kickstart doesn't work, the towers just refuses to show unless you make it instant.

refractionpcsx2:
> There is an interesting note with the bios, loading it without instant VU1 makes the white towers not to show during the startup (as "expected") but now, even VU kickstart doesn't work, the towers just refuses to show unless you make it instant.

so wait, should they be there or shouldn't they?  Also note with the new VU XGKick stuff (and possibly non-instant VU) that you no longer get the errors on the BIOS for GS Packet Size

gilderoylockhart01:
Ok so I will explain better. Now with instant VU1 disabled you don't get the packet size error with the interpreter but the towers won't show.

However, when using the VU1 recompiler with instant VU1 disabled, you can "cheat" by using VU kickstart which will make the towers blue instead.

That "cheat" doesn't work with the VU1 interpreter.

refractionpcsx2:
makes sense, kickstart doesn't really work in the same way on the interpreter

gilderoylockhart01:
And it is worth mentioning that kickstart even on the interpreters makes the packet size error to show again.

refractionpcsx2:
yes, more than likely.

Anyway I updated some VU sync stuff again last night, i dunno if that will have changed anything for the better, it was more an attempt to get the speed back that's been lost recently but maintaining compat.

gilderoylockhart01:
Dammit, ONI is still bad. Now the error shown is different.

Capture

Here is a savestate leading to where the problem happens:

[ONI vu int problem.zip](https://github.com/PCSX2/pcsx2/files/7135313/ONI.vu.int.problem.zip)

gilderoylockhart01:
Ok little update on this, this seems to be a timing issue on the EE side, -3 cyclerate makes the problem go away.

refractionpcsx2:
it's a GIF problem again, but this time PATH2. I can fix that, but now I'm tracing another problem using synced VU + VU Int in Destruction Derby seems to make the game explode.

So the fix will be along, eventually..

refractionpcsx2:
Okay well, ONI should be okay now.

The unfortunate reality is, there's gonna be games which don't work with Instant VU off and timed XG Kick, especially PATH3 masking type games, and this isn't due to this PR, it's due to the design of the GIF Unit, it doesn't pause the DMA's if PATH1 is busy for example, it continues to transfer it, which breaks the PATH3 Masking timing completely, and right now there's nothing I can do about that, and it's WAY out of scope of this PR.  I've already done a bunch of unrelated triage, and I'm not really willing to do much more.

gilderoylockhart01:
Thanks, yeah the game is fine now. Yeah sure, this isn't a gif PR and in it's actual state, the PR seems fine on my end, Mercenaries 2 now work properly without instant VU1 (which is needed for distant flickering).

refractionpcsx2:
Good to hear

refractionpcsx2:
GIF rewrite cancelled, came up with a way to make the gif unit pause while a PATH1 transfer is going on, solving the issue with Instant VU being off and using VU Int in PATH3 games.

(just be sure not to load savestates too quickly, there's a bug which currently causes infinite loops)

refractionpcsx2:
@gilderoylockhart01 you might have more luck with Instant VU off now. That K-1 Premium Dynamite game now works with Instant VU off.

gilderoylockhart01:
The results so far are good, now find my own way is on par with the recompiler. I would have liked to test Batman Vengeance on it but I don't have the game anymore unfortunately.

refractionpcsx2:
Well, thanks for all the testing you have done.

gilderoylockhart01:
I haven't finished yet to be honest but yeah, awsome work on that one. Actually, the VU0 intepreter is pretty usable in itself speed wise which helps with that BGE demo and it's implicit timings. VU1 is of course, a lot more taxing but that makes sense here.

gilderoylockhart01:
Btw it is interesting to see that proper XGkick doesn't help hasbro familly game night at all with it's sps on the game table.

I guess that will need a separate issue thread.

image

refractionpcsx2:
It's a lot worse on Dobiestation, I'm not quite sure what's causing that, if it's behaviour we don't understand yet or bad floats.

gilderoylockhart01:
Enter the Matrix is very interesting, proper XGkick + instant VU1 disabled makes the game to work without the timing hack, there will still be a need for a VIF FIFO however.

gilderoylockhart01:
Sorry, we still have a problem in a demo. The demo of NBA shootout 2003 produces a weird glitch on the characters hair with instant VU1 disabled and VU int.

Capture

https://drive.google.com/file/d/1_b49mz4UU9y14Vnan6p9rP4wc63W_J4N/view?usp=sharing

refractionpcsx2:
> Sorry, we still have a problem in a demo. The demo of NBA shootout 2003 produces a weird glitch on the characters hair with instant VU1 disabled and VU int.
> 
> Capture
> 
> https://drive.google.com/file/d/1_b49mz4UU9y14Vnan6p9rP4wc63W_J4N/view?usp=sharing

More timing issues, fun....
and I presume the JIT's are okay?

gilderoylockhart01:
Yeah the JITs are okay.

Btw you can add to the list of fixed games ("Jaws Unleashed", fixed level 2 crashing (only in VU1 int - XGkick related))

RedDevilus:
Don't forget Inline Skating, Conflict Desert Storm and Star Wars Episode 3 to the list of fixed games.

refractionpcsx2:
> Don't forget Inline Skating (XGKick), Conflict Desert Storm (XGKick) and Star Wars Episode 3 (GIF) to the list of fixed games.

they were working on the JIT's though?

RedDevilus:
I mean on the interpreters

refractionpcsx2:
oh, sure, the laundry list of games fixed on the interpreter by the PR is probably exceptionally long.  I think what gilderoylockhart01 is saying is it fixes a bug which happens on master with JIT's (and obviously didn't work in the Int's before), Like Tennis Court Smash

gilderoylockhart01:
Yeah that's what I meant, Jaws unleashed will never work if XGkick isn't handled properly.

refractionpcsx2:
> Yeah the JITs are okay.
> 
> Btw you can add to the list of fixed games ("Jaws Unleashed", fixed level 2 crashing (only in VU1 int - XGkick related))

I just tried it on the JIT's and they are also broken with Instant VU off.

gilderoylockhart01:
Actually you are right, yeah, blame me :)

Other than that it's all green for me.

legend800:
@refractionpcsx2  One of these recent VU changes broke Matrix Path of Neo. Now when an enemy dies, the polygons explode instead of a matrix code fade out. After this happens a few times the game freezes. Reproduced numerous times.

Not sure of the exact change since not all VU changes had builds. It was fine on the available build on bottom and broken by the next available build up top. 

2021-09-16 13_49_35-Window

[emuLog.txt](https://github.com/PCSX2/pcsx2/files/7180904/emuLog.txt)

Config:
Windows 10 64-bit
GeForce GTX 3070
AMD 5900x@ 4.8ghz
32GB RAM

refractionpcsx2:
@legend800 can you try enabling the VU Kickstart gamefix please

legend800:
@refractionpcsx2 Yep, works perfectly again with that on. Whew. Can we get that added to the GameIndex?

refractionpcsx2:
Yeah I'll get it added

PR #4733, opened 2021-09-06
MicroVU: Purge Min/Max speed hack

F0bes:
### Description of Changes
Remove a disabled, and apparently broken speed hack

### Rationale behind Changes
Absolutely none, I'm losing my mind

image


RedDevilus:
RIP Athlon X2 1% boost but you will be sacrificed for the greater good.

PR #4735, opened 2021-09-07
Change s8 definition to be int8_t

### LINKED ISSUE #4716: Explicitly define s8 as signed char instead of char
### Describe the Bug

Not really a bug as such but for future porting this could be problematic as other systems you need to be explicit with char and signed char types

### Reproduction Steps

N/A

### Expected Behavior

Devs stop losing their shit over char types and we all have a picnic*

*picnic is optional

### PCSX2 Revision

n/a

### Operating System

Windows 10 (32bit)

### If Linux - Specify Distro

_No response_

### CPU

n/a

### GPU

n/a

### GS Settings

_No response_

### Emulation Settings

_No response_

### GS Window Screenshots

_No response_

### Logs & Dumps

_No response_

Ziemas:
And fix up compiler errors caused by this.
Fixes #4716

### DIFF
--- a/common/Pcsx2Types.h
+++ b/common/Pcsx2Types.h
@@ -31,10 +31,7 @@ class FastFormatUnicode;
 //  Basic Atomic Types
 // --------------------------------------------------------------------------------------
 
-// Note: char does not have a default sign, unlike other types. As we actually want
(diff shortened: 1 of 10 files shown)


tadanokojin:
YOLO

PR #4742, opened 2021-09-07
Use 64-bit operations on x64

tellowkrinkle:
### Description of Changes
Adds dedicated x64 implementations for things that can be done more easily on x64

### Rationale behind Changes
Minor speed and code size improvements

### Suggested Testing Steps
Test x64 builds and make sure nothing broke


rspretty:
This patch caused incorrect frame limiting on windows.
For example, when booting BIOS without this patch,
(UpdateVSyncRate) FPS Limit Changed : 59.94 fps
appears in log.
With this patch, above line does not appear and FPS limit goes too low.

Runs faster with frame limiting disabled. Great!!

refractionpcsx2:
Sorry, merged other stuff so there's some conflicts, but if you can sort them, we can then merge

refractionpcsx2 (approved):

PR #4745, opened 2021-09-08
Replace scalar variable names with percent variable names

RedPanda4552:
Storage format for these variables was changed, so old configs would cause framelimiter lockup with incredibly low values.

### Description of Changes
Changes *Scalar variables to *Percent

### Rationale behind Changes
Prevents framelimiter lockup when importing old settings.

### Suggested Testing Steps
Validate framelimiter runs at 100% normally, and the configured slowmo and turbo speeds.

### DIFF
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -293,7 +293,7 @@ struct Pcsx2Config
 		int FramesToDraw{ 2 }; // number of consecutive frames (fields) to render
 		int FramesToSkip{ 2 }; // number of consecutive frames (fields) to skip
 
-		double LimitScalar{ 1.0 };
+		double LimitPercent{ 1.0 };
 		double FramerateNTSC{ 59.94 };
 		double FrameratePAL{ 50.00 };
 
@@ -311,7 +311,7 @@ struct Pcsx2Config
 				OpEqu( FrameLimitEnable )		&&
 				OpEqu( VsyncEnable )			&&
 
-				OpEqu( LimitScalar )			&&
+				OpEqu( LimitPercent )			&&
 				OpEqu( FramerateNTSC )			&&
 				OpEqu( FrameratePAL )			&&
 
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -339,7 +339,7 @@ u32 UpdateVSyncRate()
 	const double vertical_frequency = GetVerticalFrequency();
 
 	const double frames_per_second = vertical_frequency / 2.0;
-	const double frame_limit = frames_per_second * EmuConfig.GS.LimitScalar;
+	const double frame_limit = frames_per_second * EmuConfig.GS.LimitPercent;
 
 	const double tick_rate = GetTickFrequency() / 2.0;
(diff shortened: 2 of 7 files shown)


F0bes (approved):

F0bes:
Maybe rename the commit to have a suffix?

pcsx2/Config.h
```diff
@@ -293,7 +293,7 @@ struct Pcsx2Config
 		int FramesToDraw{ 2 }; // number of consecutive frames (fields) to render
 		int FramesToSkip{ 2 }; // number of consecutive frames (fields) to skip
 
-		double LimitScalar{ 1.0 };
+		double LimitPercent{ 1.0 };
```
RedDevilus, line 296: Nitpick: Not exactly a correct name when the value is 1.0 as this is the framelimiter on right?

tadanokojin:
With regards to this,

I realized the change I made was breaking when I made it, admittedly I wasn't expecting so much hoopla about it considering our stated policy is to recommend users against reusing old ini's.

The solution can only be one of a few things and remain consistent:
- the policy changes and all inis moving forward are backwards compatible
- the policy doesn't change, this pr doesn't get merged, we continue as always
- something about the state of the project changes that makes the entire thing irrelevant

I really don't fancy a solution which requires arbitrarily changing the ini key along with a number of variables every time we want to make a change to how that key functions. I see that as wasting a bunch of changes and time on a problem that could be solved more reasonably. Backwards compatibility is nice but it should not get in the way of thoughtful development decisions or at the burden of people who want to make reasonable changes to how these things function. Coming up with arbitrary name changes for variables and config keys and adding this extra gotcha for working on the project is not in that category for me. By the same token that it may be unreasonable to expect users to never overwrite their inis, it must also be unreasonable to expect developers to be aware that such changes, which are not often made, could cause a problem and that they won't be made by accident or even caught in a reasonable amount of time. Especially when we have no infrastructure for that outside of hoping it gets caught in review. 

Additionally, I don't want us making this change and then getting back into the old song and dance about how we don't support using old inis. You can't have your cake and eat it too. We aren't Microsoft. We don't set expectations to our users that we don't support something only do to so anyway when that causes a problem or proves inconvenient for us. That means if anything breaks in the future because of an ini change, that's going to be entirely on us and we're going to have to offer solutions and support for that. No more washing our hands of it saying "we put it in the instructions". The instructions would need to change in addition to this. You'll be setting the expectation with our users that if this becomes a problem again, that this will be the solution we implement for the foreseeable future. Do you think that's a reasonable expectation we should be setting for developers and users?

From my perspective, it really should not have to be a choice between a burden on dev vs a burden on users and I think we have solutions that are far less short sighted than I consider this one to be. I would really prefer in the future that people consulted me about the things I've changed rather than thinking I hadn't considered the exact solution they want to implement. The truth is I had considered this in great detail before deciding on what I did. It's just that it seems some people had different expectations than I did despite a policy to the contrary. I want to improve UX as much as the next guy and some people might take me as a perfectionist who overthinks everything but I just see the pitfalls of not considering the bigger picture, especially on a project like PCSX2 where this has been an ongoing problem for a while now.

PR #4748, opened 2021-09-09
R5900: Fix LWL not sign extending in interpreter

stenzek:
### Description of Changes

This was zero extending because of the implicit promotion from signed->unsigned, causing incorrect high bits in the GPR.

Funnily enough, this was noted in the source, but implemented incorrectly.

### Rationale behind Changes

It was incorrect :)

### Suggested Testing Steps

Probably not much dependent on this, and it's interpreter only.


refractionpcsx2:
Thanks, good to see you again, Sten :)

PR #4756, opened 2021-09-11
Stdio console on mac

tellowkrinkle:
### Description of Changes
Enables the "Program Log to Stdio" menu option on macOS
Also makes it check the console for color support before trying to use color escapes

### Rationale behind Changes
It's useful on macOS too

### Suggested Testing Steps
Make sure option works on macOS and Linux


F0bes (approved):
Re-based onto master locally.
Works fine on Linux, logging gets redirected to console and colours are still there.

Mrlinkwii (approved):

PR #4767, opened 2021-09-14
microVU: Replace XGKick hack with synced XGKick option

refractionpcsx2:
### Description of Changes

Replaces XGKick hack option with an XGKick Sync option which provides accurate behaviour for the VU

Fixes Tennis Court Smash and Love Smash games which previously couldn't be fixed.
WRC no longer requires a patch, just the xgkickhack option.
Simple 2000 Series Vol.26 - The Pinball X 3 SPS and console spam.

Note: it's not a hack anymore, it just has to be called that in the gamedb :P

### Rationale behind Changes
Some games couldn't be fixes with the traditional hack.

### Suggested Testing Steps
Try games that require the XGKick hack, make sure they still work (you can search the gamedb for "XGKick" if you need an aid).

If you have games that have always experienced errors in the console (GS packet exceeds VU memory) or spikey polygons, give the new XGKick Sync option a try, it might fix it.

### DIFF
--- a/pcsx2/VUops.cpp
+++ b/pcsx2/VUops.cpp
@@ -2625,7 +2625,8 @@ void _vuXGKICKTransfer(u32 cycles, bool flush)
 
 	VU1.xgkickcyclecount += cycles;
 	VU1.xgkicklastcycle += cycles;
-	VUM_LOG("Adding %d cycles, total XGKick cycles to run now %d", cycles, VU1.xgkickcyclecount);
+
+	VUM_LOG("Adding %d cycles, total XGKick cycles to run now %d flush %d enabled %d", cycles, VU1.xgkickcyclecount, flush, VU1.xgkickenable);
 
 	while (VU1.xgkickenable && (flush || VU1.xgkickcyclecount >= 2))
 	{
@@ -2633,7 +2634,8 @@ void _vuXGKICKTransfer(u32 cycles, bool flush)
 
 		if (VU1.xgkicksizeremaining == 0)
 		{
-			u32 size = gifUnit.GetGSPacketSize(GIF_PATH_1, VU1.Mem, VU1.xgkickaddr);
+			VUM_LOG("XGKICK reading new tag from %x", VU1.xgkickaddr);
+			u32 size = gifUnit.GetGSPacketSize(GIF_PATH_1, vuRegs[1].Mem, VU1.xgkickaddr, ~0u, flush);
 			VU1.xgkicksizeremaining = size & 0xFFFF;
 			VU1.xgkickendpacket = size >> 31;
 			VU1.xgkickdiff = 0x4000 - VU1.xgkickaddr;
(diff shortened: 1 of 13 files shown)


gilderoylockhart01:
Ok, I admit I predicted this would come ;) and yeah, great stuff!

However before doing the testing I have 2 things to report which in my opinion matters:

Putting this sync option into the hack section is misleading and can be confusing regarding what it does, I think it would be a perfect match for the speedhack section with an option like "enable instant XGkick" because yeah, the performace impact can be indeed big.

Secondly, well, in the gameDB "hack' refers as something which isn't the correct behaviour, which is not the case there. Maybe "instantXGkick 0" like the instantVU1 that already exist would be a much better fit I guess.

refractionpcsx2:
> Putting this sync option into the hack section is misleading and can be confusing regarding what it does, I think it would be a perfect match for the speedhack section with an option like "enable instant XGkick" because yeah, the performace impact can be indeed big.

I agree, however it's the "gamefixes" section, not hacks section :P And I was careful to not name it as a hack.  The performance impact isn't huge, but it can be 10-15% at high framerates.

> Secondly, well, in the gameDB "hack' refers as something which isn't the correct behaviour, which is not the case there. Maybe "instantXGkick 0" like the instantVU1 that already exist would be a much better fit I guess.

Yep, I agree, however, I can't without changing how the gamedb works, it auto appends "hack" on to the names it's looking for in the gamedb, so it needs to be on there, but as far as the user facing stuff goes, it doesn't say hack anywhere.

seta-san:
doesn't help world series baseball 2k3.. in fact it causes a lock up

refractionpcsx2:
> doesn't help world series baseball 2k3.. in fact it causes a lock up

Okay, I'm going to say this one more time.  WSB 2k3 and Marvell Nemesis are likely never being fixed, at all, they will be forever broken unless you see a PR called "Miracle making WSB and Marvel work"

seta-san:
yeah. but that was kind of said about things like boogie and hype and yet you keep pulling miracles out of the hat one after another. it's not that i don't listen to you. it's that i believe in you.

gilderoylockhart01:
And even still, we first need a vif fifo + only the interperters *could* handle those games but don't expect miracles here, it will be a sub 1fps experience.

Also, related to this PR it is important to add Jaws Unleashed to the list of XGkick GameDB entries:

https://wiki.pcsx2.net/Jaws_Unleashed

refractionpcsx2:
> And even still, we first need a vif fifo + only the interperters _could_ handle those games but don't expect miracles here, it will be a sub 1fps experience.
> 
> Also, related to this PR it is important to add Jaws Unleashed to the list of XGkick GameDB entries:
> 
> https://wiki.pcsx2.net/Jaws_Unleashed

What does it fix in Jaws? So I can add a comment.  There's no mention of its requirement in the wiki

gilderoylockhart01:
It should fix a crash at level 2. Kozarovv if I remember have that game if it needs testing.

gilderoylockhart01:
Does the interpreter handle this edge case right too?

refractionpcsx2:
I modified that too, yes

gilderoylockhart01:
K1 world GP 2004 will also require XGkick and instant VU1 disabled like 2006.

As for 2005 if the game is also producing a weird black screen flickering, it will also require intant VU1 disabled.

refractionpcsx2:
> K1 world GP 2004 will also require XGkick and instant VU1 disabled like 2006.
> 
> As for 2005 if the game is also producing a weird black screen flickering, it will also require intant VU1 disabled.

yes, the game has already got XGKickHack enabled in the GameDB

pcsx2/VUops.cpp
```diff
@@ -215,7 +215,7 @@ __fi void _vuTestPipes(VURegs* VU)
 	{
 		if (VU1.xgkickenable)
 		{
-			_vuXGKICKTransfer((VU1.cycle - VU1.xgkicklastcycle), false);
+			_vuXGKICKTransfer((VU1.cycle - VU1.xgkicklastcycle)-1, false);
```
lightningterror, line 218: Formatting.

gilderoylockhart01:
I sugest adding those GameDB entries:

ONI:

SLUS-20064

SLES-50134
SLES-50176
SLES-50177
SLES-50178
SLES-50179

XGkickhack // Fixes various VU1 timing behaviour causing audio glitches in some scenarios.

refractionpcsx2:
I'm not saying that, because that's a lie :P

gilderoylockhart01:
Well it's up to you really if you want to add it or not to be honest, as for the comment, I don't know what else to say really

PR #4774, opened 2021-09-16
SPU2: Disallow KeyOn within 2T of last KeyOn

refractionpcsx2:
### Description of Changes
Disallow KeyOn less than 2T (768x2 cycles) of the previous KeyOn for a voice.
Fixes Legend of Spyro New Beginning hang

### Rationale behind Changes
KeyOn/Off events shouldn't happen within 2T's of the last one for each voice (possibly the whole register, but unsure on that, per voice seems okay)

### Suggested Testing Steps
uhh, any games you know had problems previously with delay cycles, that would be great.

### DIFF
--- a/pcsx2/SPU2/spu2sys.cpp
+++ b/pcsx2/SPU2/spu2sys.cpp
@@ -1979,6 +1979,12 @@ void StartVoices(int core, u32 value)
 		if (!((value >> vc) & 1))
 			continue;
 
+		if ((Cycles - Cores[core].Voices[vc].PlayCycle) < 2)
+		{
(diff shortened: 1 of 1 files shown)


gilderoylockhart01:
This PR does work fine in my set of games. It doesn't solve the "unknown TCTF" audio issue in Oni tho (first issue is pure EE timing issue).

refractionpcsx2:
thanks for testing

gilderoylockhart01:
Actually, it does seems to help a bit with the first Oni issue, the problem doesn't happen anymore in the place I showed in the screenshots, however what can be interesting is this PR seems to make the game upset with default cyclerate and syncVU1, the game will skip some sound completely now. I have to else enable instantVU1 or make the EE slightly faster to cop with the changes.

PR #4781, opened 2021-09-17
common: add DynamicLibrary

tadanokojin:
### Description of Changes
Adds a wrapper for optional library dependencies that deals with OS-specific details so users don't have to.

### Rationale behind Changes
It's better than using something like `wxDynamicLibrary` which depends on a large gui toolkit to function. As it stands this isn't much of an issue now due to the fact that we no longer use plugins and macos/linux has no optional dependencies that I can find. I implemented UNIX anyway for the future.

### Suggested Testing Steps
TBD

### DIFF
--- a/common/DynamicLibrary.cpp
+++ b/common/DynamicLibrary.cpp
@@ -1,0 +1,168 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2021 PCSX2 Dev Team
+ *
+ *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
+ *  of the GNU Lesser General Public License as published by the Free Software Found-
+ *  ation, either version 3 of the License, or (at your option) any later version.
+ *
+ *  PCSX2 is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+ *  PURPOSE.  See the GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along with PCSX2.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifdef _WIN32
+#include <Windows.h>
+#include <wil/com.h>
+#else
+#include <dlfcn.h>
+#endif
+
+#include <fmt/core.h>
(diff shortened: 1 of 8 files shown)


common/DynamicLibrary.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+		void Close()
+		{
+			if (m_handle)
+				dlclose(m_handle);
+		}
+
+		void* GetSymbolByName(std::string_view name)
+		{
+			return dlsym(m_handle, name.data());
```
tellowkrinkle, line 100: Don't use `string_view` if you need a null-terminated string, `string_view` is one of the few places in C++ you can legitimately end up with non-null-terminated strings (by trying to substring a longer string)

common/DynamicLibrary.cpp
```diff
@@ -0,0 +1,166 @@
...
+		}
+
+		void* GetSymbolByOrdinal(const int /*ordinal*/)
+		{
+			return nullptr;
```
tellowkrinkle, line 105: If it's only used in ifdef'd Windows code, thoughts on leaving it out of the UNIX API completely?  That way attempting to use it would be a compiler error instead of a runtime one?

common/DynamicLibrary.cpp
```diff
@@ -0,0 +1,166 @@
...
+			return "dll";
+		}
+
+		bool Open(std::string_view name)
+		{
+			constexpr DWORD flags =
+				LOAD_LIBRARY_SEARCH_APPLICATION_DIR |
+				LOAD_LIBRARY_SEARCH_SYSTEM32;
+
+			m_handle.reset(LoadLibraryExA(name.data(), nullptr, flags));
```
CookiePLMonster, line 42: Likewise on null terminated strings. `LoadLibrary` requires one and `std::string_view` won't guarantee that.

common/DynamicLibrary.cpp
```diff
@@ -0,0 +1,166 @@
...
+		static constexpr const char* ExtName()
+		{
+#if defined(__APPLE__)
+			return "dylib";
+#else
+			return "so";
+#endif
+		}
+
+		bool Open(std::string_view name)
```
CookiePLMonster, line 85: ditto

CookiePLMonster (changes requested):

lightningterror:

PR #4784, opened 2021-09-17
Can i?

LeoKids:
Not sure if it wouldn't be accepted but it would be more clear.

### Description of Changes

Changes regarding downloading in a rom site.

### Rationale behind Changes

Because the readme had no mention of downloading roms in a rom website.


F0bes:
No.

refractionpcsx2:
We're not going to recommend people download anything, it's piracy and we don't condone it, at all.

RedDevilus:
Beside this applies to ALL EMULATORS. One way to kill the projects in a speedrun effort.

PR #4005, opened 2020-12-15
Remove wx variable types from config code. Rewrite Portable mode.

kenshen112:
This PR is a total rewrite of AppUserMode and the start of a rewrite to Config.h pcsx2Config.cpp AppConfig. This clears a whole chunk of needless wx variable type usage in favor of standard std / filesystem ones. This casts all types back to their wx counterparts when needed.

PR #4786, opened 2021-09-18
Remove configuration coupling to GUI

stenzek:
### Description of Changes
This pull requests moves a bunch of configuration options from `AppConfig` to the base config class, reducing coupling from the emulator core to the GUI.

Long term (imo) everything should be moved to a single ini, but for now I've maintained the old behaviour of saving the different areas to different inis, so that existing settings don't break.

The `PCSX2_CORE` define relates to the remainder of the decoupling work - it enables PCSX2 to be compiled without anything in the `gui/` directory, which will be needed for us to develop a Qt frontend. The reverse of the macro is used to maintain the current wx frontend behavior.

Memory cards and save states have been moved from the GUI to the emulator core. Save state compression has also been refactored, ditching the persistent thread (which gets removed anyway) for `std::thread`.

FMV switching has also been partially rewritten and simplified. `HostKeyEvent` has been added so that the pad classes don't need to pull in `App.h`, although this is a complete mess anyway (particularly for Linux).

The next thing I want to move is GS, and get rid of `GSApp`. But there's a *lot* of options here and risk for regressions, so I'm leaving that for a separate PR.

### Rationale behind Changes
Making the emulator core dependent on the GUI is bad.

### Suggested Testing Steps
There's quite a few things touched here, so it'll need testing in a few areas:
 - Making sure existing configurations don't break.
 - Making sure FMV switching isn't broken.
 - Making sure no pad functionality has been broken.


kenshen112:
https://github.com/PCSX2/pcsx2/pull/4005

RedDevilus:
Might be a good time to update the copyright header dates as well.

JordanTheToaster:
I've found no issues with it including loading an old config on Windows just needs a rebase now.

MrCK1:
I ran through the PR yesterday and tried a variety of setting changes. Loading old config seems to work as expected, there were no issues applying settings ect.

JordanTheToaster:
Tested again and everything seems golden on Windows.

stenzek:
A note for if/when this gets merged; you'll probably want to squash it, there were some commits in the middle where custom paths were broken. They're fine in the final commit obviously, but it could be an issue for bisecting in the future if it's not squashed.

Mrlinkwii:
everything seems fine here :+1:

refractionpcsx2:
Custom paths being broken at some point isn't really a big deal, I'd rather have the history there to go through the commits and deal with the possibly broken feature that I doubt any of us are using for 5 minutes to find a bug, than to have to guess an entire PR's worth of stuff where the bug is :P

PR #4787, opened 2021-09-18
wx: Convert GS Capture Settings window to wxWidgets

xTVaser:
### Description of Changes

Part of the effort to convert non-wx windows to wx to support translations.

This also makes the dialog usable on linux (minus the codec support)

This is currently blocked in a few ways by the recent WIL changes:
- recordings always crash the emulator when being stopped (quartz.dll?)
- F12 to open the window also results in a crash

Windows:
gsCaptureWindowWin32

### Rationale behind Changes

Effort towards the 1.8 release.

### Suggested Testing Steps

- Open the dialog with F12, and via the Capture menu.  It should only open if a game is running
- Try to create a recording on linux and on windows, both with and without a codec (for windows)
- Try to modify the filepath by hand, and via browsing
- Try configuring the codec / changing the width-height / colorspace


lightningterror:
Closing as wx is currently a legacy option, seems rather no point to continue work on wx. Can be re opened if work continues.

Issue #4714, opened 2021-09-04
[BUG]: Can't Rebind Controller


TheDiscordian:
### Describe the Bug

Switch controller defaults to having X and O swapped, as well as Triangle and Square are also swapped. Following other issues on here, I read I should be editing either `PAD.ini` or `OnePAD2.ini`, it's *extremely* unclear what these 2 files do, as they're identical. I noticed PAD.ini has a `mouse_sensitivity` param which is definitely getting r/w from as the emulator runs. I decided to edit both with this line:

`SDL2 = 03000000000000000000000000006800,Nintendo Switch Pro Controller,platform:Linux,a:b1,b:b0,back:b8,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b12,leftshoulder:b4,leftstick:b10,lefttrigger:b6,leftx:a0,lefty:a1,rightshoulder:b5,rightstick:b11,righttrigger:b7,rightx:a2,righty:a3,start:b9,x:b3,y:b2,`

Should clearly swap the A/B and X/Y buttons, fixing my issue. However this edit in reality actually does nothing.

### Reproduction Steps

Try to use a Nintendo Switch Pro Controller with PCSX2.

### Expected Behavior

Bottom button is X, right button is O, top button is Triangle, left button is Square. I expected the line I added to my configs to achieve that.

### PCSX2 Revision

dev-1632

### Operating System

Linux (64bit) - Specify Distro Below

### If Linux - Specify Distro

Manjaro

### CPU

Ryzen 7 1700x

### GPU

RX 580

### Logs & Dumps

`PAD: controller (Nintendo Switch Pro Controller) detected, GUID:03000000000000000000000000006800`

Mrlinkwii:
:wave:  atm you cant rebind the gamepad on linux , see #4414  , in terms of defaults, you need to go upstream for the SDL   (gabomdq /SDL_GameControllerDB ) and put an issue their :slightly_smiling_face:

TheDiscordian:
Thank you for the quick response! I'm not sure if it'd be appropriate to rebind all the Switch controllers for an SDL database, as some people like A to be the right side button. I believe the issue is specific to PCSX2 as with this software people will be aiming for a Playstation layout (perhaps even a "Swap A/B X/Y" option would resolve this across the board). 

[...]

Issue #4414, opened 2021-05-02
[BUG] Cannot bind controller buttons


XXXBold:
**Describe the bug**
I'm not able to bind any button from my wired XBox 360 Controller to any Key. Controller itself works, if I choose the "Gamepad Configuration" and toggle "Enable Rumble", it is vibrating. 
Inputs from keyboard / mouse are accepted as they should.

This issue seems to apply to a few ppl with different controllers, as it's mentioned in the forums aswell: https://forums.pcsx2.net/Thread-changing-controller-layout-doesnt-work

**To Reproduce**
Open Controller configuration, try to assign any key by pressing buttons on the Controller. See that no input seems to be accepted.

**Expected behavior**
Buttons should be recognized and bound accordingly.

**Emulation Settings**
Preset 2, default.

**System Info (please complete the following information):**
 -   PCSX2 Revision: latest available from PPA (For 20.04): https://launchpad.net/%7Epcsx2-team/+archive/ubuntu/pcsx2-daily
 -   OS: Linux Mint 20.1 (Ubuntu 20.04 based)
Systeminfo:
```
inxi -Fxz
System:    Kernel: 5.8.0-50-generic x86_64 bits: 64 compiler: N/A Desktop: Cinnamon 4.8.6 Distro: Linux Mint 20.1 Ulyssa 
           base: Ubuntu 20.04 focal 
Machine:   Type: Desktop System: Gigabyte product: X570 AORUS ELITE v: -CF serial: <filter> 
           Mobo: Gigabyte model: X570 AORUS ELITE v: x.x serial: <filter> UEFI: American Megatrends v: F20 date: 07/07/2020 
CPU:       Topology: 12-Core model: AMD Ryzen 9 3900X bits: 64 type: MT MCP arch: Zen L2 cache: 6144 KiB 
           flags: avx avx2 lm nx pae sse sse2 sse3 sse4_1 sse4_2 sse4a ssse3 svm bogomips: 182050 
           Speed: 2212 MHz min/max: 2200/3800 MHz Core speeds (MHz): 1: 2212 2: 2202 3: 2639 4: 2216 5: 2255 6: 2230 7: 2219 
           8: 3345 9: 2052 10: 1947 11: 2196 12: 2220 13: 2195 14: 2207 15: 2184 16: 2177 17: 3313 18: 3215 19: 2051 20: 2104 
           21: 2026 22: 2225 23: 2197 24: 2283 
[...]

PR #4809, opened 2021-09-22
GUI(linux): Abort 'set all buttons' & fixed window size

JosephLees:
### Description of Changes

Adds the ability to abort ‘set all buttons’ by pressing ESC, currently the user has to cycle the the entire mapping to regain control.  

The screen was also cutting off abit of the lower buttons and controller label so just bumped up the default height.

Also it seems to be a bit unclear as to if the gamepad keys can be remapped in the GUI (as seen in these  issues: #4414 #4714  ) so just clarified in the dialog box to hopefully reduce frustration. 

Note: made the 'if' comment style consistent within the file. Translated the french comment, was translated before but removed in a merge?

### Rationale behind Changes
Adding quality of life improvements to the linux Gamepad Settings

### Suggested Testing Steps

Open up the ‘Gamepad Settings’ config, select ‘Set All Buttons’, press the ESC key to abort

### DIFF
--- a/pcsx2/PAD/Linux/wx_dialog/dialog.cpp
+++ b/pcsx2/PAD/Linux/wx_dialog/dialog.cpp
@@ -472,12 +472,12 @@ void PADDialog::InitDialog()
 
 void PADDialog::OnButtonClicked(wxCommandEvent& event)
 {
-	// Affichage d'un message à chaque clic sur le bouton
+	// Display a message each time the button is clicked
 	wxButton* bt_tmp = (wxButton*)event.GetEventObject(); // get the button object
 	int bt_id = bt_tmp->GetId() - wxID_HIGHEST - 1;       // get the real ID
 	int gamepad_id = m_tab_gamepad->GetSelection();       // get the tab ID (equivalent to the gamepad id)
 	if (bt_id >= 0 && bt_id <= PAD_R_LEFT)
-	{                      // if the button ID is a gamepad button
+	{ // if the button ID is a gamepad button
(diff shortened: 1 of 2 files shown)


pcsx2/PAD/Linux/wx_dialog/dialog.cpp
```diff
@@ -611,12 +616,16 @@ void PADDialog::config_key(int pad, int key)
 				clear_key(pad, key);
 				set_keyboard_key(pad, key_pressed, key);
 				m_simulatedKeys[pad][key] = key_pressed;
+			} else {
```
lightningterror, line 619: Formatting/

lightningterror:
Commits need to be squashed and properly named with prefixes.

JosephLees:
Apologies, I should have re-read the contribution docs more diligently….

pcsx2/PAD/Linux/wx_dialog/dialog.cpp
```diff
@@ -540,7 +541,7 @@ void PADDialog::OnButtonClicked(wxCommandEvent& event)
 			}
 			m_pan_tabs[gamepad_id]->Refresh();
 			m_pan_tabs[gamepad_id]->Update();
-			config_key(gamepad_id, i);
+			key_captured = config_key(gamepad_id, i);
```
lightningterror, line 544: You could move the declaration here, that way it will silence the codacy complaint.

arcum42:
Seems like everything's working, and I don't see anything that should be an issue. Merging.

PR #4818, opened 2021-09-24
Add partial LTO to CMake

tellowkrinkle:
### Description of Changes
Add option to cmake that runs LTO on the core emulator but nothing else, and enable it in the CI

### Rationale behind Changes
The intermediate files for LTO on everything on Windows+Ninja manage to take up so much disk space we fill up the entire CI runner, which isn't great.  Running LTO on core only uses up about 230MB of (zstd-compressed) buildcache, so I assume 1-2gb of disk, which isn't anywhere close to the ~12GB we're allotted, so I don't expect to run out of space any time soon.

### Suggested Testing Steps
Test CMake binary and compare to msbuild SSE4 binaries, make sure performance is similar

### DIFF
--- a/.github/workflows/linux-workflow.yml
+++ b/.github/workflows/linux-workflow.yml
@@ -70,11 +70,13 @@ jobs:
           - os: ubuntu-18.04
             platform: x86
             compiler: gcc
+            cmakeflags: -DLTO_PCSX2_CORE=ON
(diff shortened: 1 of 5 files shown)


JordanTheToaster:
Tested and CMake 32 bit is within 4 fps of SSE4 32 bit,

PR #4822, opened 2021-09-25
EE: Fix JIT exits on 64-bit Windows

stenzek:
### Description of Changes
Currently exiting the recompiler (e.g. toggling frame limiting) crashes the emulator in 64-bit Windows builds, as it attempts to longjmp across the recompiler stack frames, which don't have any unwind information.

> Unhandled exception at 0x00007FFB7D43042D (ntdll.dll) in pcsx2x64-dbg.exe: 0xC00000FF: A malformed function table was encountered during an unwind operation.

This PR adds a replacement implementation for longjmp which simply saves/restores registers, and doesn't care about unwinding the stack. You shouldn't be creating any complex C++ objects between the recompiler exit point and where the execution state is checked anyway.

I used inline asm for Linux, and a separate asm file for Windows, mainly just to avoid the headache of cmake and supporting .s files. Strictly speaking we don't need to replace longjmp on Linux, but consistency across platforms is never a bad thing. At some point that SEH nonsense for 32-bit Windows should be dropped too.

### Rationale behind Changes
It was broken and crashing.

### Suggested Testing Steps
Make sure all platforms still work. I haven't checked Linux 32-bit.

### DIFF
--- a/common/FastJmp.cpp
+++ b/common/FastJmp.cpp
@@ -1,0 +1,92 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2021  PCSX2 Dev Team
+ *
+ *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
+ *  of the GNU Lesser General Public License as published by the Free Software Found-
+ *  ation, either version 3 of the License, or (at your option) any later version.
+ *
+ *  PCSX2 is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+ *  PURPOSE.  See the GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along with PCSX2.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "FastJmp.h"
+
+#ifndef _WIN32
+
+#if defined(__APPLE__)
+#define PREFIX "_"
+#else
+#define PREFIX ""
+#endif
+
+#if defined(_M_X86_64)
+
+asm(
+	"\t.global " PREFIX "fastjmp_set\n"
+	"\t.global " PREFIX "fastjmp_jmp\n"
+	"\t.text\n"
+	"\t" PREFIX "fastjmp_set:" R"(
+	movq 0(%rsp), %rax
+	movq %rsp, %rdx			# fixup stack pointer, so it doesn't include the call to fastjmp_set
+	addq $8, %rdx
+	movq %rax, 0(%rdi)	# actually rip
+	movq %rbx, 8(%rdi)
+	movq %rdx, 16(%rdi)	# actually rsp
+	movq %rbp, 24(%rdi)
+	movq %r12, 32(%rdi)
+	movq %r13, 40(%rdi)
+	movq %r14, 48(%rdi)
+	movq %r15, 56(%rdi)
+	xorl %eax, %eax
+	ret
+)"
+	"\t" PREFIX "fastjmp_jmp:" R"(
+	movl %esi, %eax
+	movq 0(%rdi), %rdx	# actually rip
+	movq 8(%rdi), %rbx
+	movq 16(%rdi), %rsp	# actually rsp
+	movq 24(%rdi), %rbp
+	movq 32(%rdi), %r12
+	movq 40(%rdi), %r13
+	movq 48(%rdi), %r14
+	movq 56(%rdi), %r15
+	jmp *%rdx
+)");
+
+#elif defined(_M_X86_32)
+
+asm(
+	"\t.global " PREFIX "fastjmp_set\n"
+	"\t.global " PREFIX "fastjmp_jmp\n"
+	"\t.text\n"
+	"\t" PREFIX "fastjmp_set:" R"(
+	movl %eax, 0(%esp)
+	movl %esp, %edx			# fixup stack pointer, so it doesn't include the call to fastjmp_set
+	addl $4, %edx
+	movl %eax, 0(%ecx)	# actually eip
+	movl %ebx, 4(%ecx)
+	movl %edx, 8(%ecx)	# actually esp
+	movl %ebp, 12(%ecx)
+	movl %esi, 16(%ecx)
+	movl %edi, 20(%ecx)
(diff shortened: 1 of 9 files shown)


tellowkrinkle:
MacOS prefixes C names with `_`, you can either use the preprocessor (`#define ASM_FUNCTION(x) _##x` on macOS), use some more ifdefs, or since you went with C inline asm, throw it in an `__attribute__((naked))` function instead of defining the symbol in asm to make everything work

common/FastJmp.h
```diff
@@ -0,0 +1,57 @@
...
+#elif !defined(_WIN32) && !defined(__x86_64)
+__attribute__((fastcall)) int fastjmp_set(fastjmp_buf* buf);
+__attribute__((fastcall, noreturn)) void fastjmp_jmp(const fastjmp_buf* buf, int ret);
+#elif defined(_WIN32)
+int fastjmp_set(fastjmp_buf* buf);
+__declspec(noreturn) int fastjmp_jmp(const fastjmp_buf* buf, int ret);
+#else
+int fastjmp_set(fastjmp_buf* buf);
+__attribute__((noreturn)) int fastjmp_jmp(const fastjmp_buf* buf, int ret);
+#endif
```
tellowkrinkle, line 56: We already define `__fastcall` on non-windows platforms  (and I have a PR open that will avoid the warning spam by defining it to nothing on x86_64) https://github.com/PCSX2/pcsx2/blob/c1fea5bc162f16c840dea23bdb8e1fdd34a01d22/common/Pcsx2Defs.h#L152

If you want to add something there for noreturn (can call it `NORETURN` or `_noreturn` or whatever) then we can avoid 4 different ifdefs for the different platforms

common/common.vcxproj
```diff
@@ -83,10 +88,16 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="EventSource.inl" />
+    <MASM Include="FastJmp.asm">
+      <FileType>Document</FileType>
+      <PreprocessorDefinitions Condition="'$(Platform)'=='Win32'">_M_X86;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions Condition="'$(Platform)'=='x64'">_M_X86;_M_X86_64;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
tellowkrinkle, line 94: Is it possible to define this more globally or re-use our C++ defs?  In case we want to add more in the future it would be useful to not have to put it on each asm file separately

common/FastJmp.cpp
```diff
@@ -0,0 +1,86 @@
...
+	"movl %esi, %eax\n\t"
+	"movq 0(%rdi), %rdx\n\t" // actually rip
+	"movq 8(%rdi), %rbx\n\t"
+	"movq 16(%rdi), %rsp\n\t" // actually rsp
+	"movq 24(%rdi), %rbp\n\t"
+	"movq 32(%rdi), %r12\n\t"
+	"movq 40(%rdi), %r13\n\t"
+	"movq 48(%rdi), %r14\n\t"
+	"movq 56(%rdi), %r15\n\t"
+	"jmp *%rdx\n\t");
```
tellowkrinkle, line 52: Since we have fancy C++ things, thoughts on using multiline strings for this?  That way you don't need all the `\n\t"` https://gcc.godbolt.org/z/eY448c53c

Assembler doesn't seem to take `//` though, so you'l have to use `#` or `/* */`

common/FastJmp.h
stenzek: Done. I named it `__noreturn` for consistency, but as far as I'm aware double-underscored identifiers are reserved, so we should probably do a mass rename at some point.

common/FastJmp.h
```diff
@@ -0,0 +1,57 @@
...
+#else
+#if defined(_M_ARM64)
+	static constexpr std::size_t BUF_SIZE = 168;
+#elif defined(__x86_64__)
+	static constexpr std::size_t BUF_SIZE = 64;
+#else
+	static constexpr std::size_t BUF_SIZE = 24;
+#endif
+#endif
```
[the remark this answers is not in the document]
stenzek: Done. The other architecture wasn't meant to be in here anyway.

common/common.vcxproj
stenzek: Unfortunately no, MASM seems to use a separate set of defines from C++. I renamed them to _M_X86_32/64 for consistency at least.

common/FastJmp.cpp
stenzek: Didn't end up being *as* clean because I had to add the function prefix in there for macs, but it's a bit easier on the eyes now at least.

common/common.vcxproj
```diff
@@ -83,10 +88,17 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="EventSource.inl" />
+    <MASM Include="FastJmp.asm">
+      <FileType>Document</FileType>
+      <PreprocessorDefinitions Condition="'$(Platform)'=='Win32'">_M_X86_32;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions Condition="'$(Platform)'=='x64'">_M_X86_64;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions Condition="$(Configuration.Contains(AVX2))">_M_X86_64;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
tellowkrinkle, line 95: I think this was supposed to be `_M_AVX2`?

Though I don't think you should have to `vzeroupper`, if the upper halves are volatile no one should be relying on them being any particular value, whether that's zero or whatever garbage the function that called `fastjmp_jmp` left there (probably also zero if the compiler is trying to avoid transition penalties).

stenzek: Whoops, indeed, copy/paste fail.

I was thinking of the rec frames more than C frames. But I suppose they don't use 256bit ops, so this is probably fine to omit?

tellowkrinkle: > But I suppose they don't use 256bit ops, so this is probably fine to omit?

Yeah not right now, x86emitter can't even emit them yet
Going to have to think about how we deal with it in the future, but there's a lot of other C functions we call from the recompiler that aren't fastjmp_jmp so we'll still need to deal with those either way

tellowkrinkle (approved):

PR #4837, opened 2021-09-26
Core: Fix some Wsign-compare warnings, fix Wimplicit-fallthrough warning

lightningterror:
### Description of Changes

Fix some Wsign-compare warnings in iCore , fix Wimplicit-fallthrough warning in debugger.
### Rationale behind Changes

The less warnings the better.
### Suggested Testing Steps

See if gcc complains still

### DIFF
--- a/pcsx2/x86/iCore.cpp
+++ b/pcsx2/x86/iCore.cpp
@@ -58,7 +58,7 @@ static int s_xmmchecknext = 0;
 
 void _backupNeededXMM()
 {
-	for (int i = 0; i < iREGCNT_XMM; i++)
+	for (size_t i = 0; i < iREGCNT_XMM; i++)
 	{
 		if (xmmregs[i].inuse)
 		{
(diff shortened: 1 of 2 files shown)


CookiePLMonster:
The type you **actually** want to use in those loop counters is `size_t`, as that's a natural word size type, ideal for array indexing.

pcsx2/x86/iCore.cpp
```diff
@@ -58,7 +58,7 @@ static int s_xmmchecknext = 0;
 
 void _backupNeededXMM()
 {
-	for (int i = 0; i < iREGCNT_XMM; i++)
+	for (unsigned int i = 0; i < iREGCNT_XMM; i++)
```
refractionpcsx2, line 61: unsigned or size_t? make up your mind :P

lightningterror: Forgot about this one.

Issue #4117, opened 2021-01-08
[META] Games which are CDVD Timing Sensitive


refractionpcsx2:
Some games seem to exhibit a behaviour where they do not function correctly if the CDVD is running at roughly the correct speed.  There seems to be no rhyme or reason to it, you can run the CDVD slower, or faster, and the game will work.

Looking at the IOP logs, everything seems to function normally, aside from when it reads the last interrupt, it acknowledges it, then never touches it again.  From playing around doing things like adjusting the VU timings or adjusting SIF timings seems to also change the issue, so I'm not sure if this is a game timing problem or a PCSX2 specific issue.

Games affected:

- [ ]  Digital Devil Saga (Hangs on certain cutscenes)
- [ ]  Silent Hill 2 (Hangs on certain cutscenes)
- [ ]  Shadowman 2 (Bad textures, prefers slower CDVD)

seta-san:
doesn't ecco the dolphin and Victorious Boxers Ippo's Road to Victory both have issues with cdvd timing

-Ecco has an evil race condition: one of its custom IOP modules resets CDVDMAN while it's loading a module from disc. The race condition is resolved by making the EE slower. This gives enough time for CDVDMAN to reset itself before the EE sends a request to load a module.

-victorious boxers will hang unless you set "fast cdvd"

refractionpcsx2:
Thanks, updated

seta-san:
Bob the Builder locks on opening video without fast CDVD

seta-san:
can we get checkmarks on bob, victorious, and shadowman and any others that we know are fixed?

refractionpcsx2:
yeah Digital Devil Saga is fine now

gilderoylockhart01:
Ecco the dolphin? Why is the game in that meta? Yeah, the game does stupid stings with the dvd driver, but I don't think it's directly a cdvd emulation problem.

seta-san:
curious. if it's not a cdvd issue what kind of issue would it be ?

HeroponRikiBestest:
According to https://www.patreon.com/posts/sins-of-ps2-ecco-38348208, it's an ee timing issue that just happens to cause a cdvd malfunction if it's not handled properly.

JPTIZ:
[...]

PR #4839, opened 2021-09-28
CDVD: Implement correct SpindleCtrl handling

refractionpcsx2:
### Description of Changes
 Implement correct SpindleCtrl handling
Semi decouple end of CDVD DMA's
Fix alternative reading method for reading not picking the correct speeds for DVD's
Restores partially working Audio CD playback we had before.

### Rationale behind Changes
CDVD was always full speed all the time regardless of the spindle settings which could have caused problems in reading on some games.  Also the CDRead function was not exclusive for CD's but it was using CD timings.

### Suggested Testing Steps
Test games and stuff, especially ones listed in #4117  (sorry but Silent Hill 2 still sucks, something else going on)

Seems to resolve the Digital Devil Saga videos not playing
Also the DMA timing change resolved the remaining texture issues on Shadowman

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -631,37 +631,52 @@ static s32 cdvdReadDvdDualInfo(s32* dualType, u32* layer1Start)
 	return CDVD->getDualInfo(dualType, layer1Start);
 }
 
+static bool cdvdIsDVD()
+{
+	if (cdvd.Type == CDVD_TYPE_DETCTDVDS || cdvd.Type == CDVD_TYPE_DETCTDVDD
+		|| cdvd.Type == CDVD_TYPE_PS2DVD || cdvd.Type == CDVD_TYPE_DVDV)
+		return true;
+	else
+		return false;
+}
+
 static uint cdvdBlockReadTime(CDVD_MODE_TYPE mode)
 {
 	int numSectors = 0;
 	int offset = 0;
-	// Sector counts are taken from google for Single layer, Dual layer DVD's and for 700MB CD's
-	switch (cdvd.Type)
-	{
-		case CDVD_TYPE_DETCTDVDS:
-		case CDVD_TYPE_PS2DVD:
-			numSectors = 2298496;
-			break;
-		case CDVD_TYPE_DETCTDVDD:
-			numSectors = 4173824 / 2; // Total sectors for both layers, assume half per layer
-			u32 layer1Start;
-			s32 dualType;
 
-			// Layer 1 needs an offset as it goes back to the middle of the disc
-			cdvdReadDvdDualInfo(&dualType, &layer1Start);
-			if (cdvd.Sector >= layer1Start)
-				offset = layer1Start;
-			break;
-		default: // Pretty much every CD format
-			numSectors = 360000;
-			break;
+	// CAV Read speed is roughly 41% in the centre full speed on outer edge. I imagine it's more logarithmic than this
+	if (cdvd.SpindlCtrl & CDVD_SPINDLE_CAV)
+	{
+		// Sector counts are taken from google for Single layer, Dual layer DVD's and for 700MB CD's
+		switch (cdvd.Type)
+		{
+			case CDVD_TYPE_DETCTDVDS:
+			case CDVD_TYPE_PS2DVD:
+				numSectors = 2298496;
+				break;
+			case CDVD_TYPE_DETCTDVDD:
+				numSectors = 4173824 / 2; // Total sectors for both layers, assume half per layer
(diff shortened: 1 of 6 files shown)


pcsx2/CDVD/CDVD.cpp
```diff
@@ -1307,7 +1339,48 @@ static void cdvdWrite04(u8 rt)
 			cdvd.nSectors = *(u32*)(cdvd.Param + 4);
 			cdvd.RetryCnt = (cdvd.Param[8] == 0) ? 0x100 : cdvd.Param[8];
 			cdvd.SpindlCtrl = cdvd.Param[9];
-			cdvd.Speed = 24;
+
+			switch (cdvd.SpindlCtrl & CDVD_SPINDLE_SPEED)
+			{
+			case 0: // Will use current speed
```
lightningterror, line 1345: Formatting on the cases.

gilderoylockhart01:
So, this PR seems to show good results across the board in the games I tested but I don't have a whole bunch of CDVD problematic games anyway so my testing can only be done to a very limited extent:

So far, 2 games seems affected by this:

- Silent Hill 2 PAL, the game now works properly but you will need the EETimingHack + syncVU1 to be effective booth at the same time.

- Oni, yeah, that one again, now it does the same problem as it used to do before the VU timing change, and perhaps even more funny is that sync XGkick is again, the only viable fix to this.

refractionpcsx2:
> So, this PR seems to show good results across the board in the games I tested but I don't have a whole bunch of CDVD problematic games anyway so my testing can only be done to a very limited extent:
> 
> So far, 2 games seems affected by this:
> 
> * Silent Hill 2 PAL, the game now works properly but you will need the EETimingHack + syncVU1 to be effective booth at the same time.
> * Oni, yeah, that one again, now it does the same problem as it used to do before the VU timing change, and perhaps even more funny is that sync XGkick is again, the only viable fix to this.

yep and both those games are a random wheel of fortune of timing, poking random buttons and hacks makes them work because it adjusts timings, it's not directly related to this.

gilderoylockhart01:
Well, at least now booth of those games can be fixed with one or more GameDB entries, that's still a + in my opinion.

kenshen112:
Klonoa 2 is still fine

seta-san:
I’ll test tonight

seta-san:
bad news.
bob the builder,
ecco the dolphin
vicctorious boxers
all still suffer from the same problems

good news. shadowman textures look like they have cleaned up
gs_20210929014517
gs_20210929014319
.

seta-san:
i'm going to guess this pr is going to fix most things that needed slower cdvd. why things need fast cdvd is still a big question.

refractionpcsx2:
Well, this gives the speeds to the disc the developers expected, with Shadowman the extra DMA delay helped, idk if that's *actually* what was missing but it seems to help for sure.,

As for the other games, I've no idea what they want, they're kinda dumb, especially Bob the Builder.  I might revamp the Fast CDVD speedhack to be less shit

seta-san:
well. bob the builder is just on a normal cd, not a dvd. so it's naturally stupid.

gilderoylockhart01:
It might be worth for the games that still have problems to try things like syncVU1 and the timing hack.

refractionpcsx2:
Bob the Builder is fine if you set the EE cycle rate to -1

refractionpcsx2:
Victorious Boxers, seems to work fine for me on standard cycle rate with no Fast CDVD?? When does it hang?

Ecco the Dolphin will always need a patch, it's not a CDVD Timing problem (not quite anyway)

gilderoylockhart01:
I know that might not make a lot of sense, but I sugest waiting until PCSX2 will have a proper vif fifo before going further with those games as it turns out that fast cdvd/timings can often hide this kind of problem.

At some point rolland garros, shadowman 2 and the 2 more games more upper are the main indicator that now the cdvd isn't to blame anymore.

seta-san:
> Victorious Boxers, yeah dunno, music gives up. Game will hang around for a while on the "round 2" but it does wake up, the music is long gone though.

wakes up from what? is there some sort of timeout function? what's it waiting for? or is it just some sort of dma transfer that needs to get unstuck?

refractionpcsx2:
> > Victorious Boxers, yeah dunno, music gives up. Game will hang around for a while on the "round 2" but it does wake up, the music is long gone though.
> 
> wakes up from what? is there some sort of timeout function? what's it waiting for? or is it just some sort of dma transfer that needs to get unstuck?

I mean the fight continues if you just wait a bit, i dunno what it's waiting for, probably trying to start the music off again and failing, but it just continues eventually anyway.  Why it dies, I don't know.

PR #4844, opened 2021-09-29
Move Vu overflow flag checks to a Gamefix + reorganise panel

### LINKED ISSUE #4841: [BUG]: The Chronicles of Narnia - Prince Caspian (PAL): The second level cannot currently be completed using the EE recompiler.
### Describe the Bug

This is a regression that appears to have started with f9d96f5 

Basically there are a pile of rocks in the second level that need to be picked up and moved. However, the rocks keep resetting to their original positions when you put them down, meaning that they can never be cleared. There is also an example in the first level of what I think the same problem where a giant cog that you need to pick up is dropped from above, but if you don't pick it up immediately its falling animation repeats. That one's not game breaking as the cog can be picked up and used still, but the rock pile problem is.

The problem only manifests while using the EE recompiler. Using the interpreter fixes the issue but it's slow.

To test if that was the commit that broke it, I built a current dev build with that commit's changes reverted and it does seem to fix it.

### Reproduction Steps

Play until the second level. Try to move the pile of rocks blocking the cave.

### Expected Behavior

You can move the pile of rocks and their positions won't reset.

### PCSX2 Revision

v1.7.0-dev-1828

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

### CPU

Ryzen 3700X

### GPU

RTX 3070

### GS Settings

Not applicable to this problem.

### Emulation Settings

Except for using the EE interpreter, none of the hacks, rounding, or clamping settings make any difference as far as I can tell.

### GS Window Screenshots

_No response_

### Logs & Dumps

Not sure what to provide here. Let me know if anything can be useful.

refractionpcsx2:
### Description of Changes
Moves the VU Overflow checks for MAC and Status flags to a game fix so it can be optionally enabled. Also rearranged the game fixes panel to make a bit more sense by grouping fixes together.

### Rationale behind Changes
x86 can't really check for overflowed PS2 floats because we don't have the range and it's not really possible to guess. Turns out our implementation breaks Chronicles or Narnia - Prince Caspian, but it's required for Superman Returns.
Fixes #4841

### Suggested Testing Steps
Not really much needs testing beyond Superman Returns and Chronicles of Narnia - Prince Caspian. I guess make sure the UI doesn't explode and the hacks turn on the correct hack when you manually enable them.

### DIFF
--- a/pcsx2/x86/microVU_Upper.inl
+++ b/pcsx2/x86/microVU_Upper.inl
@@ -78,17 +78,18 @@ static void mVUupdateFlags(mV, const xmm& reg, const xmm& regT1in = xEmptyReg, c
 	xMOVMSKPS(gprT2, regT1); // Used for Zero Flag Calculation
 
 	xAND(mReg, AND_XYZW); // Grab "Is Signed" bits from the previous calculation
-	xSHL(mReg, 4 + ADD_XYZW);
+	xSHL(mReg, 4);
 
 	//-------------------------Check for Zero flags------------------------------
 
 	xAND(gprT2, AND_XYZW); // Grab "Is Zero" bits from the previous calculation
-	if (mFLAG.doFlag)
-		SHIFT_XYZW(gprT2);
 	xOR(mReg, gprT2);
 
 	//-------------------------Overflow Flags-----------------------------------
-	if (sFLAG.doFlag)
+	// We can't really do this because of the limited range of x86 and the value MIGHT genuinely be FLT_MAX (x86)
+	// so this will need to remain as a gamefix for the one game that needs it (Superman Returns)
+	// until some sort of soft float implementation.
+	if (sFLAG.doFlag && CHECK_VUOVERFLOWHACK)
 	{
 		//Calculate overflow
 		xMOVAPS(regT1, regT2);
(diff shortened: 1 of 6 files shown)


pcsx2/x86/microVU_Upper.inl
```diff
@@ -98,15 +99,20 @@ static void mVUupdateFlags(mV, const xmm& reg, const xmm& regT1in = xEmptyReg, c
 		xAND(gprT2, AND_XYZW); // Grab "Is FLT_MAX" bits from the previous calculation
 		xForwardJump32 oJMP(Jcc_Zero);
 			xOR(sReg, 0x820000);
+			if (mFLAG.doFlag)
```
lightningterror, line 102: Formatting here as well, also the above line.

refractionpcsx2: You've moaned about this exact line before, and I told you last time I did that on purpsoe because it was a conditional statement, just like an if statement.

refractionpcsx2: Ugh fine, if it'll stop you complaining

PR #4845, opened 2021-09-29
USB: dshow: Initialize multi-threaded COM libs

jackun:
### Description of Changes
Initialize multi-threaded COM using `wil` helpers.

### Rationale behind Changes
Fix crashing caused by single/multi-threaded CoInitialize conflict.

### Suggested Testing Steps
Set USB to EyeToy with directshow api and run a game.

### DIFF
--- a/pcsx2/USB/usb-eyetoy/cam-windows.h
+++ b/pcsx2/USB/usb-eyetoy/cam-windows.h
@@ -21,6 +21,8 @@
 #include <windows.h>
 #include <dshow.h>
 
+#include <wil/com.h>
+
 extern "C" {
 extern GUID IID_ISampleGrabberCB;
 extern GUID CLSID_SampleGrabber;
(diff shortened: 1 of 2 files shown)


pcsx2/USB/usb-eyetoy/cam-windows.h
```diff
@@ -118,6 +120,7 @@ namespace usb_eyetoy
 			IBaseFilter* samplegrabberfilter;
 			ISampleGrabber* samplegrabber;
 			IBaseFilter* nullrenderer;
+			wil::unique_couninitialize_call dshowCoInitialize;
```
CookiePLMonster, line 123: This should probably be moved to the top, above `ICaptureGraphBuilder2* pGraphBuilder;` - right now it doesn't matter, but if they get modified to use `wil::com_ptr` you'd want the construction/destruction to happen in the correct order, i.e. coinit constructed first and destructed last.

JordanTheToaster:
Does not crash following the testing procedure.

PR #4853, opened 2021-10-01
CDVD: Fix possible uncaught exception in CheckDiskTypeFS

stenzek:
### Description of Changes
I noticed this one when I had reads broken - if you manage to load an invalid ISO image, it'll throw this exception on the CPU thread.

Also replaced an explicit unique_ptr construction with make_unique.

### Rationale behind Changes
This shouldn't be an exception in the first place, but at least now it doesn't kill the CPU thread/SIGABRT.

### Suggested Testing Steps
None really needed, since it's an exceptional (hah) scenario. I guess you could try loading a disc image without a valid ISO structure.

### DIFF
--- a/pcsx2/CDVD/CDVDaccess.cpp
+++ b/pcsx2/CDVD/CDVDaccess.cpp
@@ -77,45 +77,51 @@ static void CheckNullCDVD()
 static int CheckDiskTypeFS(int baseType)
 {
 	IsoFSCDVD isofs;
-	IsoDirectory rootdir(isofs);
 	try
 	{
-		IsoFile file(rootdir, L"SYSTEM.CNF;1");
+		IsoDirectory rootdir(isofs);
 
-		int size = file.getLength();
-
-		std::unique_ptr<char[]> buffer(new char[file.getLength() + 1]);
-		file.read(buffer.get(), size);
-		buffer[size] = '\0';
-
-		char* pos = strstr(buffer.get(), "BOOT2");
-		if (pos == NULL)
+		try
 		{
-			pos = strstr(buffer.get(), "BOOT");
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
So on master it does this:
image

On this PR it just says this in the log then continues as if there's no disc

> Open virtual disk tray
> Path: [Unnamed or unknown]
> ISO mounting failed: PCSX2 is unable to identify the ISO image type.

stenzek:
Right, the goal was to allow it to boot to the RSOD. In my Qt branch, since I don't have any of the cross-thread exception nonsense, it just caused it to `abort()`.

refractionpcsx2:
If that was intended, it's fine. But from a user perspective, they have no idea that their ISO is bad now, they might just think it's region protection or something, so it would be nice to get some sort of user feedback, beyond the RSOD :P

stenzek:
Relying on the ISO directory being missing alone isn't a great indicator of that, imo.

I'd suggest displaying an error if either the directory or system.cnf can't be found, instead of just "isofilesystem file not found" (which isn't very descriptive).

refractionpcsx2:
Okay, well if it's something that can be revisited later to be a bit more user friendly, then that's fine :) Like if all of your try/catches fail it just says "Look mate, your iso sucks, try another one or remake it" in a polite fashion.

pcsx2/CDVD/CDVDaccess.cpp
```diff
@@ -77,49 +77,55 @@ static void CheckNullCDVD()
...
+			return CDVD_TYPE_DVDV;
+		}
+		catch (Exception::FileNotFound&)
+		{
+		}
 	}
 	catch (Exception::FileNotFound&)
 	{
-	}
+	}
```
refractionpcsx2, line 128: whitespace

PR #4855, opened 2021-10-01
Fix some compile warnings spewed by MSVC x64

stenzek:
### Description of Changes
These have been pestering me for a while.

### Rationale behind Changes
Making MSVC shush.

### Suggested Testing Steps
Shouldn't be any functional changes here.

### DIFF
--- a/tools/bin2cpp/bin2cpp.cpp
+++ b/tools/bin2cpp/bin2cpp.cpp
@@ -105,8 +105,8 @@ int main(int argc, char* argv[])
 
 	strcpy(srcfile,argv[ARG_SRCFILE]);
 
-	int srcfn_len = strlen( srcfile );
-	if( srcfile[srcfn_len-4] != '.' )
+	const size_t srcfn_len = strlen( srcfile );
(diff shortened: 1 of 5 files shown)


tools/bin2cpp/bin2cpp.cpp
```diff
@@ -105,8 +105,8 @@ int main(int argc, char* argv[])
 
 	strcpy(srcfile,argv[ARG_SRCFILE]);
 
-	int srcfn_len = strlen( srcfile );
-	if( srcfile[srcfn_len-4] != '.' )
+	const size_t srcfn_len = strlen( srcfile );
+	if( srcfn_len < 0 || srcfile[srcfn_len-4] != '.' )
```
refractionpcsx2, line 109: size_t is unsigned, checking if less than 0

stenzek: Whoops, that was meant to be a 4 (to avoid the buffer underflow).

refractionpcsx2 (approved):

PR #4859, opened 2021-10-01
GS: Cleanup ini ranges for some values.

lightningterror:
### Description of Changes

Use clamp instead of min max for OSD and Shade Boost.
Don't allow negative value for upscale multiplier, avoids a black screen.
### Rationale behind Changes

Cleaner code, avoids a black screen if upscale multiplier is negative.
### Suggested Testing Steps

Make sure Shade boost, osd, upscaling still works.

Note: these changes are mostly for editing ini manually.


JordanTheToaster:
Shade boost OSD and upscaling all work and negative values no longer black screen.

Issue #4851, opened 2021-10-01
Meta: Disc Swapping Capabilities


RedDevilus:
Games that can use disc swapping to extend gameplay, unlock items, etc
Keep in mind disc swapping is broken at the moment.

- [x] SingStar Games (#3355) (Fixed in #4860)
- [x] Dual Disc Games (Linux Distro, God of War, Grandia 3, Xenosaga + others ) (Fixed in #4860)
- [x] Dragon Ball Z Budokai Tenkaichi 1/2 (+ others maybe) (Fixed in #4860)
- [x] Monster Rancher Games (EVO, 3, 4) (#3533) (Fixed in #4860)
- [x] Dynasty Warrior Games (#91) (Fixed in #4860)
- [x] Dance Factory (Fixed in #4860)
- [x] Singstar Take That (Annoying little gremlin)
- [x] Some versions of PS2 Linux, possibly

RedDevilus:
Part 2 fixed by (SingStar Take That and PS2 Linux):
https://github.com/PCSX2/pcsx2/pull/5174

PR #4860, opened 2021-10-02
CDVD: Implement Disc Swapping

### LINKED ISSUE #91: Disc swapping issues (Dynasty Warriors XL-Samurai Warriors XL series)
-PCSX2 version: 1.3.0 GIT 25-6-2014

-PCSX2 options: Defaults, no settings modifications affect the issue

-Plugins used: GSdx, SPU2-X, Lilypad, Linuz ISO/Gigaherz CDVD,rest null plugins

-Plugin settings: Defaults, no settings modifications affect the issue

-Description: The XL games of the series have an 'import from original' feature, which prompts you to eject the XL disc and insert the original disc to load additional content. While the actual disc swap seems to work, the game never imports the data saying 'The wrong disc has been inserted'

-How to replicate: Use any Dynasty Warriors 3/4/5 XL/ Samurai Warriors 1/2 XL game and select 'Import' in the menu, and disc swap to the original disc.

-Last known version to work: Never worked

-PC specifications: 
CPU: Intel Core i5 2500K @ 4.3Ghz
GPU: Gigabyte Geforce GTX660 Windforce OC 2 GB
OS: Windows 7 SP1

-Other comments: Some forum member discovered a way to import the data by merging the 2 discs in one ISO which partially works here: 
http://forums.pcsx2.net/Thread-Dynasty-Warriors-XL-problem-SOLVED
In general it looks like the ELF of the switched disc isn't detected properly.

### LINKED ISSUE #3355: [Feature Request] SingStar Disc Swap
***PCSX2 version:***
1.5.0 3403

All versions of SingStar except the original version and the disney version are supposed to support a disc swap feature where the disc but doesn't work in the emulator.

### LINKED ISSUE #3533: sceCdSearchFile ignores disc swaps, keeps returning first result (Monster Rancher EVO)
***PCSX2 version:***
stable build v1.6.0

***PCSX2 options:***
Tested Safe & Safest (no known working options)

***Plugins used:***
Reproduced with DVD source (cdvdGigaherz plugin) and ISO source (built-in).

Monster Rancher EVO creates monsters by reading data from other discs. The necessary swapping works fine for all disc types except other PS2 DVD games.

The game determines if the inserted PS2 DVD disc is the EVO game disc by calling `sceCdSearchFile(..., "\\SLUS_213.30;1")`. If the result is `TRUE`, EVO is inserted. However, on PCSX2 the result is almost always `TRUE` regardless of the disc. (There is a small chance that it correctly returns `FALSE` for a swapped disc, but I can't reliably make that happen.)
This causes the game to think every PS2 DVD is its own disc, preventing the swap.

I can reproduce the problem with a custom ELF.  Compile [this gist](https://gist.github.com/fldef/8f4476945320cc6c7a94abc8f9497ca6) with PS2SDK, changing `SEARCH_PATH` to fit the disc you're testing.
On PS2 hardware, `sceCdSearchFile` returns the appropriate result for the loaded disc.
In PCSX2, the return value for that filename is almost always the same as the first call despite disc swaps.

Unknown

***PC specifications:***
i5 4th gen; integrated video; Windows 10 Pro 64-bit

***See Also***
Bug #91 lists various games with a similar or identical problem. (h/t to the commenters there for their research.) However, I'm only able to reverse & test Monster Rancher EVO, so I'm treating this as a separate issue.

refractionpcsx2:
### Description of Changes
Inspired by the work by fldef in the repo here https://github.com/fldef/pcsx2-fldef and extended to work well with physical discs too

### Rationale behind Changes
Disc Swapping was completely broken in games which required it. This should hopefully resolve that.

### Suggested Testing Steps
Test swapping discs in games which need it, or the BIOS.  Dance Factory kinda works, but it doesn't display the track names, or something.

Fixes #3533
Fixes #3355
Fixes #91
Fixes Dance Factory (kinda, track names kinda freak out but not sure why that is, could be an issue with the DVD reader code, or just lack of track names)
Hopefully fixes everything on #4851

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -590,6 +590,9 @@ s32 cdvdCtrlTrayOpen()
 	cdvd.Status = CDVD_STATUS_TRAY_OPEN;
 	cdvd.Ready = CDVD_NOTREADY;
 
+	if (!g_GameStarted && g_SkipBiosHack && CDVDsys_GetSourceType() != CDVD_SourceType::Iso)
+		return 0;
+
 	cdvd.Tray.cdvdActionSeconds = 3;
 	cdvd.Tray.trayState = CDVD_DISC_EJECT;
 	return 0; // needs to be 0 for success according to homebrew test "CDVD"
@@ -600,14 +603,16 @@ s32 cdvdCtrlTrayClose()
 	DevCon.WriteLn(Color_Green, L"Close virtual disk tray");
 
 	cdvd.Ready = CDVD_READY1;
-	if (!g_GameStarted)
+	if (!g_GameStarted && g_SkipBiosHack && CDVDsys_GetSourceType() != CDVD_SourceType::Iso)
 	{
+		DevCon.WriteLn(Color_Green, L"Drive Engaging instantly");
 		cdvd.Status = CDVD_STATUS_PAUSE;
 		cdvd.Tray.trayState = CDVD_DISC_ENGAGED;
 		cdvd.Tray.cdvdActionSeconds = 0;
 	}
 	else
 	{
+		DevCon.WriteLn(Color_Green, L"Drive entering seek mode");
 		cdvd.Status = CDVD_STATUS_SEEK;
 		cdvd.Tray.trayState = CDVD_DISC_SEEKING;
 		cdvd.Tray.cdvdActionSeconds = 3;
@@ -640,7 +645,7 @@ static __fi void cdvdGetDiskType()
 	u32 oldtype = cdvd.Type;
 	cdvd.Type = DoCDVDdetectDiskType();
 
-	if (oldtype != cdvd.Type && CDVDsys_GetSourceType() != CDVD_SourceType::Iso)
+	if ((g_GameStarted || !g_SkipBiosHack) && oldtype != cdvd.Type && CDVDsys_GetSourceType() != CDVD_SourceType::Iso)
 	{
 		if (cdvd.Type == 0)
 		{
@@ -1143,13 +1148,11 @@ void cdvdVsync()
 			switch (cdvd.Tray.trayState)
 			{
 			case CDVD_DISC_EJECT:
-				DevCon.Warning("Tray handling closing tray and seeking");
-				cdvd.Status = CDVD_STATUS_SEEK;
 				cdvdCtrlTrayClose();
 				break;
 			case CDVD_DISC_SEEKING:
 			case CDVD_DISC_ENGAGED:
-				DevCon.Warning("Tray handling engaging disc");
+				DevCon.WriteLn(Color_Green, L"Drive Engaging");
 				cdvd.Tray.trayState = CDVD_DISC_ENGAGED;
 				cdvd.Status = CDVD_STATUS_PAUSE;
 				cdvd.mediaChanged = true;
@@ -1256,7 +1259,10 @@ u8 cdvdRead(u8 key)
 		case 0x0F: // TYPE
 			CDVD_LOG("cdvdRead0F(Disc Type) %x", cdvd.Type);
 			cdvdGetDiskType();
-			return cdvd.Type;
+			if (cdvd.Tray.trayState == CDVD_DISC_ENGAGED)
+				return cdvd.Type;
+			else
+				return (cdvd.Tray.trayState == CDVD_DISC_SEEKING) ? 1 : 0; // Detecting Disc / No Disc
 
 		case 0x13: // UNKNOWN
 			CDVD_LOG("cdvdRead13(Unknown) %x", 4);
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -441,7 +441,7 @@ void cdvdReloadElfInfo(wxString elfoverride)
 	// called from context of executing VM code (recompilers), so we need to trap exceptions
 	// and route them through the VM's exception handler.  (needed for non-SEH platforms, such
 	// as Linux/GCC)
-
+	DevCon.WriteLn(Color_Green, L"Reload ELF");
 	try
 	{
 		if (!elfoverride.IsEmpty())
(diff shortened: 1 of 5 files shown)


seta-san:
`ELF (cdrom0:\SCUS_976.16;1) Game CRC = 0xD5016BA5, EntryPoint = 0x00100008
Drive Engaging
Pfs_CheckDisc() called. Checking if pack cdrom0:¥PAK_IOP1.PAK;1 is on disc
Pfs_CheckDisc() called. Checking if pack cdrom0:¥PAK_IOP1.PAK;1 is on disc
`

singstar doesn't fully acknowledge the swap and never continues

refractionpcsx2:
> `ELF (cdrom0:\SCUS_976.16;1) Game CRC = 0xD5016BA5, EntryPoint = 0x00100008 Drive Engaging Pfs_CheckDisc() called. Checking if pack cdrom0:¥PAK_IOP1.PAK;1 is on disc Pfs_CheckDisc() called. Checking if pack cdrom0:¥PAK_IOP1.PAK;1 is on disc `
> 
> singstar doesn't fully acknowledge the swap and never continues

What are you swapping between? and is that the build after my second commit?

seta-san:
yes. it's after the second commit. i've tried a couple different singstar games

refractionpcsx2:
okay what combination are you doing, reproduction steps would be cool

seta-san:
singstar 80s and 90s. i get to the song select screen. i press the select button to initiate the proper swap function on the game. i perform the swap and it just hangs.

seta-san:
```
case CDVD_DISC_ENGAGED:
				DevCon.WriteLn(Color_Green, L"Drive Engaging");
				cdvd.Tray.trayState = CDVD_DISC_ENGAGED;
				cdvd.Status = CDVD_STATUS_PAUSE;
				cdvd.mediaChanged = true;
				break;
```
this is setting the traystate to something it already should be to be in this case and then just sets the cdvd status to pause. i can't tell if other code somewhere else is affecting this so i hope i'm not just making a jerk out of myself.

refractionpcsx2:
> ```
> case CDVD_DISC_ENGAGED:
> 				DevCon.WriteLn(Color_Green, L"Drive Engaging");
> 				cdvd.Tray.trayState = CDVD_DISC_ENGAGED;
> 				cdvd.Status = CDVD_STATUS_PAUSE;
> 				cdvd.mediaChanged = true;
> 				break;
> ```
> 
> this is setting the traystate to something it already should be to be in this case and then just sets the cdvd status to pause. i can't tell if other code somewhere else is affecting this so i hope i'm not just making a jerk out of myself.

I think you misunderstand what the code is doing.

kozarovv:
Is worth to mention that Dance Factory swap will work only from real CD/DVD Drive due to other issues. Also Monster Rancher games will possibly fail to use swap feature when internal image reader is used. Because game is supposed to detect and use data from ANY cd/dvd, and for now pcsx2 don't allow to read non ps2 discs afaik. Not sure that work from real drive in pcsx2 or not.

refractionpcsx2:
> Is worth to mention that Dance Factory swap will work only from real CD/DVD Drive due to other issues. Also Monster Rancher games will possibly fail to use swap feature when internal image reader is used. Because game is supposed to detect and use data from ANY cd/dvd, and for now pcsx2 don't allow to read non ps2 discs afaik. Not sure that work from real drive in pcsx2 or not.

what?

refractionpcsx2:
Interesting thing with singstar, if you have 2 copies of the same version of the game then try to swap to one of them, it'll fail, then swap to the other copy of the same game and it'll work..

seta-san:
i remember having this issue a year ago i don't remember how i resolved it.

kozarovv:
> > Is worth to mention that Dance Factory swap will work only from real CD/DVD Drive due to other issues. Also Monster Rancher games will possibly fail to use swap feature when internal image reader is used. Because game is supposed to detect and use data from ANY cd/dvd, and for now pcsx2 don't allow to read non ps2 discs afaik. Not sure that work from real drive in pcsx2 or not.
> 
> what?

I don't know what else to say. Just that pr shouldn't fix Monster Rancher, and Dance Factory if iso/bin/cue files are used, due to other issues in pcsx2. Only when real disc drive is used in pc that will work for mentioned games.

refractionpcsx2:
it won't fix audio games from cue/bin's no, but if you put in a real audio cd and swap to disc, it works.

seta-san:
after a disc swap shouldn't the cdvd status be set to seek before pause? like maybe this?

case CDVD_DISC_SEEKING:
cdvd.Status = CDVD_STATUS_ SEEK
			case CDVD_DISC_ENGAGED:
				DevCon.WriteLn(Color_Green, L"Drive Engaging");
				cdvd.Tray.trayState = CDVD_DISC_ENGAGED;
				cdvd.Status = CDVD_STATUS_PAUSE;
				cdvd.mediaChanged = true;
				break;

refractionpcsx2:
It's nothing to do with that, I've fixed it, just sorting some physical swapping issues

refractionpcsx2:
Okay, should be fixed now.  Opened a can of worms but I think I've wrestled it in to submission

gilderoylockhart01:
So far, I tested Star Ocean 3 and it work just fine with the new changes when changing from disc 2 to disc 1 (and vice versa).

Singstar is a bit more tricky but I don't know if that's a game or emulator issue yet, switching from "Take That" to "ABBA", "Anthems" and "Boy Bands vs Girl Bands" will work just fine but, switching from Anthems to "Take That" will not work (the game will not recognize the disc) and that only happens with Take That, strange.

refractionpcsx2:
I'm not sure if that's an emulator issue, I would expect behaviour to be consistent with any disc swapping.

refractionpcsx2:
Had an issue where it would keep resetting the disc when trying to full boot (because of the eeHook in our code) so I've fixed that now.

gilderoylockhart01:
The Singstar problem is an emulator problem, on real hardware switching from "ABBA" to "Take That" does work.

refractionpcsx2:
I pushed some small changes, can you tell if Singstar is any different? though I'm not expecting it, I mostly dealt with boot issues..

RedDevilus:
Singstar:

image
image
image
image
[emuLog.txt](https://github.com/PCSX2/pcsx2/files/7271866/emuLog.txt)

RedDevilus:
Fun fact:
SingStar doesn't need the change disc screen as it has 2 checks preventing wrong disc swaps ( although it can be done wrongly but it can recover again if you do it on the correct spot again)

image
Screen above is fine to disc swap or press select.

RedDevilus:
emulog for physical disc reader method:
https://cdn.discordapp.com/attachments/612095998642225162/893919865331675226/P_20211002_195814.jpg
[emuLog.txt](https://github.com/PCSX2/pcsx2/files/7271971/emuLog.txt)

refractionpcsx2:
I dunno what Take That's deal is, it seems to be only swapping to that game which is a problem, everything else seems to work fine.

I would suggest booting that game if you want to play Take That songs :P I don't know why the swap doesn't work with that one game, it's like it reads the contents then decides it isn't a Singstar disc, but it should be the same as every other Singstar game, so i dunno. Gonna ignore that, can be a new issue :P

refractionpcsx2:
I tried Take That with the original disc swap code from fldef, and well, it's an interesting outcome, it still doesn't really work 😂 

image

seta-san:
How does it behave on hardware?

gilderoylockhart01:
On real hardware it does swap as you would expect, if it doesn't work until the pr is merged I will open an issue about that.

refractionpcsx2:
> On real hardware it does swap as you would expect, if it doesn't work until the pr is merged I will open an issue about that.

well, I'm unsure why it doesn't work, the swap seems to work successfully, then it does a couple of reads then decides it doesn't like it.

refractionpcsx2:
okay I've finished messing with it now, I think it's good to go.  The only issue is the SingStar problem with the Take That disc, and I don't know what makes that disc so special, might need further investigation.

But if everything is okay with this, then I'll merge.

gilderoylockhart01:
Mmm, by bet would be on a timing issue or wrong cdvd flag for Singstar Take that, enabling the dcache or making the cdvd go faster will freak out that specific disc.

image

refractionpcsx2:
I did mess with the CDVD timing, I made CDVD slower and faster, it didn't make any difference. I think I tried the EE cycle rate slider too and that didn't help, so idk

PR #4871, opened 2021-10-03
CDVD: Time reads by sectors per second instead of bytes + Some rotational latency

refractionpcsx2:
### Description of Changes
Changes the block/sector timing to be based on sector per second instead of bytes per second
Also added rotational latency for reads which are not sequential to simulate waiting for the disc to go around, it's probably slower than it needs to be, but better safe than sorry (Shadowman)

### Rationale behind Changes
Doing timing by sectors per second is pretty much staple for PS1 emulators, but also calculating by the sector size based on a base CD speed (specifically) is incorrect, as a CD can read 75 sectors per second if in Data mode or CD Audio mode, despite CD Audio sectors being bigger.  This eradicates that problem.  The rotational latency adjusts the read timings for games such as Shadowman.  Also removed the DMA timing after the read, that probably isn't really needed.

### Suggested Testing Steps
Uhh test timing sensitive games I guess (SH2 still needs FastCDVD), CD Audio playback should be slightly smoother, but it still kinda sucks.

Games I don't care about:

Silent Hill 2
Victorious Boxers
Bob the fucking builder
Ecco the Dolphin (This is patched and that's all it will ever have)

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -689,11 +689,13 @@ static uint cdvdBlockReadTime(CDVD_MODE_TYPE mode)
 	
 		const float sectorSpeed = (((float)(cdvd.SeekToSector - offset) / numSectors) * 0.60f) + 0.40f;
 
-		return ((PSXCLK * cdvd.BlockSize) / ((float)(((mode == MODE_CDROM) ? PSX_CD_READSPEED : PSX_DVD_READSPEED) * cdvd.Speed) * sectorSpeed));
+		return (PSXCLK / ((((mode == MODE_CDROM) ? CD_SECTORS_PERSECOND : DVD_SECTORS_PERSECOND) * cdvd.Speed) * sectorSpeed));
+		//return ((PSXCLK * cdvd.BlockSize) / ((float)(((mode == MODE_CDROM) ? PSX_CD_READSPEED : PSX_DVD_READSPEED) * cdvd.Speed) * sectorSpeed));
 	}
 	
 	// CLV Read Speed is constant
-	return ((PSXCLK * cdvd.BlockSize) / (float)(((mode == MODE_CDROM) ? PSX_CD_READSPEED : PSX_DVD_READSPEED) * cdvd.Speed));
+	//return ((PSXCLK * cdvd.BlockSize) / (float)(((mode == MODE_CDROM) ? PSX_CD_READSPEED : PSX_DVD_READSPEED) * cdvd.Speed));
+	return (PSXCLK / (((mode == MODE_CDROM) ? CD_SECTORS_PERSECOND : DVD_SECTORS_PERSECOND) * cdvd.Speed));
 }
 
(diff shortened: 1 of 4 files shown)


kenshen112:
This is just ps2 cd games?

refractionpcsx2:
i mean, it affects everything, but apart from SH2 I don't know any DVD games with timing issues. Unless Klonoa 2 is DVD, then you can test that too i guess. I'm not expecting much change for normal CD's and DVD's, it's mostly CD Audio tbh..

kenshen112:
Yeah Klonoa's audio is based on DVD read speed so I'll test that. 

I was mostly asking since I know we also have https://github.com/PCSX2/pcsx2/blob/a6e188692e9199e47d84ee6a0e0868c5aad18a0c/pcsx2/CDVD/CdRom.cpp#L667 for ps1 mode to set speed depending on current read speed 1x vs 2x of the disc

refractionpcsx2:
yes, well that's what I was explaining in my rationale.

kenshen112:
I see

seta-san:
ecco same problem
bob the builder still locks up on fmv
victorious boxers same issue

and worst of all it refucked up that texture in shadow man

refractionpcsx2:
oh, that's a shame...

kenshen112:
Klonoa seems fine. I can still hear the title screen music

refractionpcsx2:
@seta-san can you provide a save at that point (not a savestate), I'm looking in to the benefit of rotational latency when not seeking to the next sequential sector.

seta-san:
[BASLUS-20413.zip](https://github.com/PCSX2/pcsx2/files/7279860/BASLUS-20413.zip)
it's a folder memory card.

leave the building. turn right.. take the first immediate right. follow it around and go straight at the intersection. it'll be the wall in front of you.

seta-san:
for the record. fast cdvd speed hack now breaks all the textures both on master and this pr.

refractionpcsx2:
> for the record. fast cdvd speed hack now breaks all the textures both on master and this pr.

fast cdvd is the last thing you ever want to use on this game, so I don't care.

thanks for the save, I'll see if this made a difference

refractionpcsx2:
image

I guess it helps then... I'll add it to the PR

seta-san:
on the inner/outer speed pr someone brought up the archemedian spiral implementation on dolphin. would that be more accurate even if slower?

refractionpcsx2:
> on the inner/outer speed pr someone brought up the archemedian spiral implementation on dolphin. would that be more accurate even if slower?

It would probably reduce the delay, I'm basically working out the rotational speed and assuming one whole turn is required.

seta-san:
I have a full day. Won’t be able to test for the next 10 to 12 hours

seta-san:
yeah. basically just refixed shadowman

refractionpcsx2:
cool

Mrlinkwii:
ran a few games over the PR found no issues :+1:

PR #4876, opened 2021-10-05
Fix memory card saving and trace log setting loading/saving

stenzek:
### Description of Changes
More regression fixes from the decoupling PR - this time, trace logs weren't being saved to the correct section, and memory cards setting changes weren't applying.

I also clang-formatted Config.h and Pcsx2Config.cpp, as they were an inconsistent mess. First two commits are the ones to look at.

### Rationale behind Changes
Regressions :(

### Suggested Testing Steps
Make sure memory card hotswitching works as expected now.

### DIFF
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -106,12 +106,12 @@ typename std::underlying_type<Enumeration>::type enum_cast(Enumeration E)
 	return static_cast<typename std::underlying_type<Enumeration>::type>(E);
 }
 
-ImplementEnumOperators( GamefixId );
-ImplementEnumOperators( SpeedhackId );
+ImplementEnumOperators(GamefixId);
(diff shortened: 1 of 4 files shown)


JordanTheToaster:
Memory card hotswapping still seems to work correctly on Windows as well as Aspect Ratio and FMV Aspect Ratio save correctly now.

refractionpcsx2:
A couple more bugs someone on the discord found

folder configs seem to be buggered if you do a fresh install so GS, SPU etc inis get created in the PCSX2 root folder instead of inis, I think the same is happening for the SPU2 debugging stuff (debug audio recordings) 

ISO list doesn't remember stuff after shutdown

Mrlinkwii:
Tested the PR , and it works as intended in terms of hotswitching memcards :+1:

PR #4882, opened 2021-10-06
GUI: remove presets 4,5,6

Mrlinkwii:
### Description of Changes

removes presets 4-6 and leaves presets  1-3 
### Rationale behind Changes

Preset 4-6 cause more issues than they help , so getting rid of them stops user error and user headache 

### Suggested Testing Steps

make sure the slider for the presets dont crash

### DIFF
--- a/pcsx2/gui/AppConfig.cpp
+++ b/pcsx2/gui/AppConfig.cpp
@@ -889,7 +889,7 @@ void AppConfig::UiTemplateOptions::LoadSave(IniInterface& ini)
 
 int AppConfig::GetMaxPresetIndex()
 {
-	return 5;
+	return 2;
 }
 
 bool AppConfig::isOkGetPresetTextAndColor(int n, wxString& label, wxColor& c)
@@ -898,12 +898,9 @@ bool AppConfig::isOkGetPresetTextAndColor(int n, wxString& label, wxColor& c)
 		{
 			{_t("Safest (No hacks)"), L"Blue"},
 			{_t("Safe (Default)"), L"Dark Green"},
-			{_t("Balanced"), L"Forest Green"},
-			{_t("Aggressive"), L"Orange"},
-			{_t("Very Aggressive"), L"Red"},
-			{_t("Mostly Harmful"), L"Purple"}};
-	if (n < 0 || n > GetMaxPresetIndex())
-		return false;
+			{_t("Balanced"), L"Forest Green"}
+		} ;
+		n  = std::clamp(n, 0, GetMaxPresetIndex());
 
 	label = wxsFormat(L"%d - ", n + 1) + presetNamesAndColors[n][0];
 	c = wxColor(presetNamesAndColors[n][1]);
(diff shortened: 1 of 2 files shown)


avih:
> Enough people have issues on 4-6

Absolutely. It's definitely outside the "recommended" or even "good" for many cases, and has always been the case.

> because they think it's more speedy or offsets the need for more powerful hardware

Yup. Users never read docs and ignore warnings. In any program - users are the hardest problem to fix ;)

That being said, it's orthogonal to the argument that it could still be useful to some other users.

> it works less than before because of accuracy.

That's a fair argument. If indeed there are barely games which can benefit from it for users with weak hardware, and/or if it now has more issues than in the past (like crashes or general bad behavior), then keeping it is indeed not a great idea.

These days I wouldn't know, because I do have fine HW, but in the past it was very useful to me (that's the reason I added the presets system - to make it easier for "normal" users to coordinate/tune a complex set of settings split between different configuration places using a simple slider).

So the question really is: are there (more than few) users who use these presets for some games and find them useful?

JordanTheToaster:
> > Enough people have issues on 4-6
> 
> Absolutely. It's definitely outside the "recommended" or even "good" for many cases, and has always been the case.
> 
> > because they think it's more speedy or offsets the need for more powerful hardware
> 
> Yup. Users never read docs and ignore warnings. In any program - users are the hardest problem to fix ;)
> 
> That being said, it's orthogonal to the argument that it could still be useful to some other users.
> 
> > it works less than before because of accuracy.
> 
> That's a fair argument. If indeed there are barely games which can benefit from it for users with weak hardware, and/or if it now has more issues than in the past (like crashes or general bad behavior), then keeping it is indeed not a great idea.
> 
> These days I wouldn't know, because I do have fine HW, but in the past it was very useful to me (that's the reason I added the presets system - to make it easier for "normal" users to coordinate/tune a complex set of settings split between different configuration places using a simple slider).
> 
> So the question really is: are there (more than few) users who use these presets for some games and find them useful?

So they can untick the preset box and set it manually the ability to change these things manually isn't being taken away just the misleading presets.

avih:
> > So they can untick the preset box and set it manually
> 
> That's also a good argument. Does it still hold true?

As it turns out, it doesn't hold true - https://github.com/PCSX2/pcsx2/pull/4883#issuecomment-938431086

Because right after removing the presets, the manual options to help those with slow systems are also removed (this is one option, but the point is clear).

And that's after a user explicitly gives an example of a game that's only playable on their system with this speed hack.

Of course, it could still be the case that this option is very bad in general (crashes a lot, introduces other bad bugs) - I don't know that, but THAT would be a good reason to remove it. But judging by this https://github.com/PCSX2/pcsx2/pull/4883#issuecomment-938311630 at least for this game it's apparently useful. My guess that it's not the only game, but it's just a guess.

Anyway, I'm not going to continue the discussion because my point was already expressed, and so has yours.

But do consider not throwing users with old systems under the bus without "proper" reason (like bugs, code complexity, etc).

Over and out.

refractionpcsx2:
Skipping 3 is rarely useful, if ever, it causes dramatic internal framerate problems, saying it makes a game playable is somewhat of a stretch, but it is rarely (if ever) useful over skipping 2, it's usually diminishing returns of usefulness after that point.

Also if a user has an old core 2 and they want good performance, 1.7 is not going to do them any favours and they are best served using an older version until they can upgrade.

The whole notion we have to hold on to the past and terrible hacks because somebody with an archaic machine might want it is ridiculous. The project needs to move forward to improve and holding on to old jank from the past to satisfy a few users is not worth our time or the technical debt it incurs.  Maybe not so much the skipping but it is considered an extremely harmful setting and we're moving away from having settings which are too harmful in the name of speed.  We have the reputation of being the emulator full of hacks for reasons like this, and this is a reputation we'd like to shake off.

dio-gh:
> flexibility to solve problems which the developers don't face, but *some* users do

But if this is a frequent problem on help channels, then clearly the number of users who use these levels despite not needing it surpasses the number of users who do need it, am I wrong? Does this not drive it home that it's a "footgun", that instead of benefitting users, it leads them astray instead?

And why do you discuss another PR here? I get that it's related, but still. This removal here is purely cosmetic, whether further levels are cut from different sliders in the suspectable future, that doesn't belong here.

avih:
> makes our life more difficult?

As I said earlier, this is a very valid reason (support). I can't judge this for you and I can't balance the factors for you.

> You can add a lot of "flexibility" but it also adds complexity

As I said earlier, that's very true in general, and definitely a valid reason to not add flexibility, and vice-verse - removing flexibility to make things simpler can be a great reason.

But as far as I can tell, not in this case - of simply removing support for some option values without actually making anything simpler at the code (that I could notice).

The existing solution to the UX problem of discouraging users from using not-recommended values by adding warning/colors is apparently not enough. I thought It is enough, but you claim it's not, and I believe you.

Finding a better solution is indeed effort, which I can't ask anyone (but myself) to make.

All of it comes down to the developers deciding what they care about, and how they balance the factors. I can't make decisions for other people.

Issue #2732, opened 2018-12-02
Silent Hill Origins - dark/white textures rendering on HW


Lyudmila17:
***PCSX2 version:***
pcsx2-v1.5.0-dev-2711-g8307cfc54-windows-x86

***PCSX2 options:***
![image](https://user-images.githubusercontent.com/14199521/49339902-c21bbf80-f638-11e8-8f34-45fbf941567a.png)

***Plugins used:***
![image](https://user-images.githubusercontent.com/14199521/49339906-deb7f780-f638-11e8-9354-bdc2673d91e3.png)

***Description of the issue:***
![gsdx_20181202134919](https://user-images.githubusercontent.com/14199521/49339949-a95fd980-f639-11e8-94c6-02549dd799ba.png)
![gsdx_20181202134938](https://user-images.githubusercontent.com/14199521/49339951-abc23380-f639-11e8-94a2-4ab7701a3a47.png)

***How to reproduce the issue:***
Going through these rooms (not sure if it happens to more)

***Last known version to work:***
Probably none.

***PC specifications:***
Windows 10 Pro x64 - Intel i5 2500 - 8GB RAM - GTX 660ti

***Snaps:***
[snaps.zip](https://github.com/PCSX2/pcsx2/files/2636923/snaps.zip)

RinMaru:
this happens in the 2nd floor of the hospital after exiting the elevator

Lyudmila17:
> this happens in the 2nd floor of the hospital after exiting the elevator

right, but happens same as the 2nd screenshot

RinMaru:
yea. only software mode fixes the issue.

Lyudmila17:
> have you tried turning on "fast texture invalidation" under hw hacks in graphic plugin ?

there is no "fast texture invalidation" hack

RinMaru:
>           have you tried turning on "fast texture invalidation" under hw hacks in graphic plugin ?

That worked but it introduces a new issue
![capture](https://user-images.githubusercontent.com/15071652/49676309-38f4f600-fa47-11e8-9c45-2d125877355a.JPG)

MrCK1:
status unchanged, the 2nd dump doesn't work in HW at all.

LukasThyWalls:
Tested with v1.7.0-dev-1211-g290183040 (20210424200746) , status unchanged.

[...]

Issue #1326, opened 2016-04-30
Silent Hill: Shattered Memories HW Mode Breaks Streaming


AmbientMalice:
-PCSX2 version: v1.5.0-20160429001616

-PCSX2 options: Full Boot.

-Plugins used: Defaults

-Plugin settings: GL HW Rendering & GL SW Rendering

-Description: Running in HW Mode causes various assets to not stream correctly. Pressing X to make Harry call for Sherry will result in the sound effect being clipped randomly. Objects will be missing textures in various ways. Sometimes Harry's head disappears. Sometimes Harry's jacket disappears. Wall textures disappear. HUD text often disappears, though this might possibly happen in SW mode, too. Switching back to SW mode isn't enough to fix missing assets. You'll have to get them to stream again by changing areas or rebooting game.

-How to replicate: Simply play the game. Interacting with doors will refresh the textures, causing random improvements and degradation, since the slight delay between touching doors and pulling them open masks streaming.

-Last known version to work: Never worked.

-PC specifications: CPU: Intel Q9400z GPU: Gigabyte GTX 750Ti OS: Windows 10 64 Bit

![gsdx_20160430113002](https://cloud.githubusercontent.com/assets/10704701/14933273/a0f1ce90-0ecc-11e6-80e8-c21bdfb1d32e.png)

![gsdx_20160430121452](https://cloud.githubusercontent.com/assets/10704701/14933299/3ee605d0-0ecd-11e6-894d-b477a10d4f6e.png)

![gsdx_20160430121500](https://cloud.githubusercontent.com/assets/10704701/14933300/46f5209e-0ecd-11e6-9a18-d04332c5a4a8.png)

Sunderland93:
I recorded all of bugs, that has in Silent Hill Shattered Memories on all render backends https://www.youtube.com/watch?v=QwOFionyNv8

Tested on Windows 8.1 with Nvidia driver 372.90 and Debian 8.4 with Nvidia driver 367.44

MrCK1:
@AmbientMalice Any status update on this?

AmbientMalice:
It affects all HW renders and when I tested a few weeks ago it was still an issue. SW mode has some of these problems, too, such as cutscenes running way too fast.

themanwiththeplan911:
[...]

PR #4887, opened 2021-10-08
gs-tc: propagate texture shuffle format on readback

tadanokojin:
### Description of Changes
Read back PSMCT32 when the texture being blit is the result of a channel effect. This is not exactly right but there's no harm in having the extra data in cpu memory since it's supposed to be there anyway. The only issue is it gets unscaled.

fixes https://github.com/PCSX2/pcsx2/issues/1326
fixes https://github.com/PCSX2/pcsx2/issues/2732

### Rationale behind Changes
Fixes silent hill flashlight.
image

### Suggested Testing Steps
If you plan on testing silent hill, the game still needs fast texture invalidation.


refractionpcsx2:
Wasn't sure what games this might fix (can't think of any with broken lighting/shadows off the top of my head), but I tested a bunch of games I know do GS downloads and they were all still peachy (except Star Wars Racer which is still broken, but probably outside the scope of this PR since it's depth download).  

so yeah, couldn't find anything else fixed I'm afraid, but also nothing broken, so LGTM

PR #4888, opened 2021-10-08
Add the current profile to the status bar.

arcum42:
### Description of Changes
Adds a spot on the status bar with the current profile that pcsx2 is set to, or "---" for none.

### Rationale behind Changes
It makes it easier to tell at a glance what profile pcsx2 is currently set to.

### Suggested Testing Steps
Enable and disable profiles, change the profile, look at the status bar.


JordanTheToaster:
Looks and functions correctly from my quick run through on Windows.

PR #4890, opened 2021-10-08
IOP Recompiler: Fix BIOS trace logging on 64bit

Ziemas:
### Description of Changes
Fixes IOP BIOS trace logging on 64bit calling conventions.

### Rationale behind Changes
It was crashing before.

### Suggested Testing Steps
Check that trace logging works/doesn't crash


Ziemas:
Updated to use `arg3reg` instead of `#ifdef`'s.

Issue #4824, opened 2021-09-25
[BUG]: Jak 3 graphical issues on 64-bit builds


EvilPlayer1:
### Describe the Bug

Graphical issues on textures of grass.

### Reproduction Steps

Bug appers after you enter the temple in wasteland.

### Expected Behavior

Textures rendered correctly.

### PCSX2 Revision

PCSX2-64bit-SSE4-pr[4786]-sha[7ecbf2b]

### Operating System

Windows 10 (64bit)

### CPU

i3-8100

### GPU

GTX 1660 SUPER

### GS Settings

![Screenshot 2021-09-25 11-23-37](https://user-images.githubusercontent.com/90133207/134764893-44110401-3d2d-4121-a37f-e089feca26b3.png)
![Screenshot 2021-09-25 11-20-55](https://user-images.githubusercontent.com/90133207/134764924-3b6664da-2898-4c9b-9e4f-8f516a50dab0.png)

### Emulation Settings

![image](https://user-images.githubusercontent.com/90133207/134764932-c28c6c81-f498-4f76-a5a2-fa538a5bd8c0.png)

### GS Window Screenshots

32-bit:
![Screenshot 2021-09-25 11-09-34](https://user-images.githubusercontent.com/90133207/134764940-a665d0d1-49db-491f-a07a-e6681447afa7.png)
64-bit:
![Screenshot 2021-09-25 11-10-52](https://user-images.githubusercontent.com/90133207/134764953-15fb9563-db62-480c-a96c-a65ba87a7666.png)

### Logs & Dumps

[log.txt](https://github.com/PCSX2/pcsx2/files/7229444/log.txt)
[gs_20210925113927.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/7229446/gs_20210925113927.gs.xz.zip)

stenzek:
Fantastic, thanks for checking!

PR #4910, opened 2021-10-17
microVU: Use uncached reg when clamping for FMAC instructions

stenzek:
### Description of Changes

This was causing broken lighting in 64-bit mode, due to the larger number of registers available. Ft gets loaded and cached, but then the FMAC instructions clamp it in all modes except none, which disturbs the cached value (mismatched with the VU state). Jak happens to rely on this value *not* being clamped, so it was "okay" in 8-register mode because it had to be reloaded.

### Rationale behind Changes

Broken rendering :)

Before:
image

After: 
image

### Suggested Testing Steps

Test tricky VU games in 64-bit mode. But this should be a relatively safe change I think, unless something was relying on the incorrect behaviour.


stenzek:
Also fixes #4824 (confirmed by reporter).

Issue #4770, opened 2021-09-15
[Feature Request] Add Hotkey for Resetting emulated system


ichee:
### Description

See Title.

### Reason

Easily restart system without breaking immersion.

### Examples

N/A

RedDevilus:
Soft reset like a Nintendo DS would be intresting.

gilderoylockhart01:
The PS2 does support that natively with the power button, I wonder what PCSX2 does with that.

refractionpcsx2:
PCSX2 doesn't emulate the power button, not sure how easy it would be, I know CDVD takes an interrupt from the power button, but I don't know if that will "turn off" or just reset

RedDevilus:
Linking full boot/fast boot from thr GUI to a hotkey will have a similar effect that might be feasible for completing this feature request.

gilderoylockhart01:
well, when the PS2 reset it also reset the CDVD with it so I assume this is what this interrupt is for.

refractionpcsx2:
Well the reset signal goes from the power switch in to the Syscon and it only goes out to the Mechacon, which then passes it to the CDVD subsystem.

So I think it's pretty responsible for it.

tadanokojin:
Is there really any benefit to LLEing the soft reset on the ps2 when you can just shutdown the emulation and start it again with the same settings?

refractionpcsx2:
not particularly, I'd be more inclined to link it up to the PCSX2 reset :P

RedDevilus:
Perhaps a candidate for when Qt is available where you just press a single button (or combo) and it automatically restarts the state (instead of going into the GUI to get a prompt to reset / swap disc).

RedDevilus:
Fixed by: https://github.com/PCSX2/pcsx2/pull/5757

PR #4913, opened 2021-10-18
CDVD: Reset VM on sceCdPowerOff writes

F0bes:
### Description of Changes
Sort of related to #4770.
Whenever CDVD SCOMMAND `sceCdPowerOff` (0xBF402016 = 0xF) is called, it'll reset (or shutdown, really) the PS2 VM.

### Rationale behind Changes
It acts more like a real authentic PS2 I guess.

### Suggested Testing Steps
If you have software whether it be homebrew or a game that shuts down the PS2, give it a spin.

This was tested on real hardware as well.

### Deeper Explaination
I dug into the PS2SDK source and found out how their poweroff was implemented.
https://github.com/ps2dev/ps2sdk/blob/11500fe1dcec03592e71406cbd9d34b39aa52589/iop/dev9/poweroff/src/poweroff.c#L102-L103

It writes 0 to 0xBF402017 (SDATAIN) -- Not too sure why this is required, my hardware shuts down without it
It writes F to 0xBF402016 (SCOMMAND sceCdPowerOff)

### Extended testing
Does this work on hardware _without_ a DEV9 adaptor? I have no means of testing without one currently (I use FreeHDBoot). Does it really matter?

Attached below is a zipped ELF that only calles the sceCdPowerOff command and does not write to SDATAIN. It turns off my hardware and PCSX2 with this commit.
[cdvdPowerOff.zip](https://github.com/PCSX2/pcsx2/files/7361599/cdvdPowerOff.zip)
.

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -1,5 +1,5 @@
 /*  PCSX2 - PS2 Emulator for PCs
- *  Copyright (C) 2002-2010  PCSX2 Dev Team
+ *  Copyright (C) 2002-2021  PCSX2 Dev Team
  *
  *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
  *  of the GNU Lesser General Public License as published by the Free Software Found-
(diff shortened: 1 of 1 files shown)


RedPanda4552:
Confirmed working, instantly powers off my SCPH-39001 model. Perhaps the other write is just trying to be cute and zero out some data lane?

PR #4917, opened 2021-10-18
Savestates: Add missing things from Savestates

refractionpcsx2:
### Description of Changes
Add missing variables from savestates and sync up MTVU properly on savestate load, hopefully makes things more deterministic.

### Rationale behind Changes
Some variables which could be important were missing from savestates, also not all of the VU's were backed up which could be a problem, especially for TAS where it relies on savestates and determinism

### Suggested Testing Steps
Load games at various points, make savestates, make sure you can load them (providing they saved correctly, of course)

### DIFF
--- a/pcsx2/SaveState.cpp
+++ b/pcsx2/SaveState.cpp
@@ -32,6 +32,7 @@
 
 #include "common/pxStreams.h"
 #include "common/SafeArray.inl"
+#include "common/StringUtil.h"
 #include "SPU2/spu2.h"
 #include "USB/USB.h"
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


MrCK1:
Tested a bunch of PS2 games and PSX (loading and saving with various settings). All seems to be working fine on my end.

stenzek:
It's more of a thing for start-and-load, which is a feature in Qt (I use it for CPU tracing). Since the normal boot routine never happens, it doesn't get a chance to populate these variables.

stenzek:
Great, thanks!

PR #4919, opened 2021-10-18
gs: gsstate (the old gsdx one) cleanup

tadanokojin:
### Description of Changes
just some spring cleaning since I'm here anyway.

### Rationale behind Changes
if I'm going to spend time in this file I'd like it to be somewhat clean

### Suggested Testing Steps
just test stuff if you want and check for crashes. don't suspect anything will but im only human.


refractionpcsx2:
Code changes look good, nice to clean up a few redundant bits and modernise others.

Ran it through a few different games just to make sure it didn't take a dump playing any games, seems fine :)

So, LGTM

PR #4922, opened 2021-10-21
Various GSDumpGUI Improvements

tellowkrinkle:
### Description of Changes
- Window is now resizable (bottom debug section grows)
- No longer prints integer numbers with 6 decimal places
- Now shows entire transfer instead of just the first packet
- Most entries have a summary line and default unexpanded, greatly increasing the information that fits on screen
- Adds support for xz-compressed gsdumps

### Rationale behind Changes
Less annoying

### Suggested Testing Steps
xz your gsdumps, try running them, try checking the debug box and looking at the debug views

### DIFF
--- a/pcsx2/GS/GSLzma.cpp
+++ b/pcsx2/GS/GSLzma.cpp
@@ -34,6 +34,11 @@ GSDumpFile::GSDumpFile(char* filename, const char* repack_filename)
 	}
 }
 
+GSDumpFile::GSDumpFile(FILE* file, FILE* repack_file)
+	: m_repack_fp(repack_file), m_fp(file)
+{
+}
+
 void GSDumpFile::Repack(void* ptr, size_t size)
 {
 	if (m_repack_fp == nullptr)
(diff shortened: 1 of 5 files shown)


lightningterror:
I have three main requests:
1. ESC key closes the running gs dump window (if on focus, which will allow me to swap between dumps fast).
2. Being able to change the snaps directory without needing to restart the emu every time for it (for someone that has over 600 dumps I like to keep separate directories for issues and requiring that I restart the emu every single time I want to swap to different directories is frustrating).
3. Adding another text field for GS logs (currently it's counter productive, every time I start the debugger I need to adjust the logs position so I can see what's going on).

Without these it's extremely annoying using the GS Debugger so I'm still relying on the old GSDumpGUI.

Bonus
4. Closing the emu automatically closes any running gs dump.

refractionpcsx2:
The sorting of the dump list seems to be case sensitive

image

refractionpcsx2:
LGTM what's there, I dunno if you wanted to address LT's requests before merging or not.

refractionpcsx2 (approved):

PR #4348, opened 2021-04-02
Replace cached GSOffset with live calculations

tellowkrinkle:
Should fix #2686, please test

The non-cached calculations are pretty fast, but generally not quite as fast as the cached versions.  I haven't noticed any slowdown in anything I've tested, but it might be good for others to test.

Also this makes heavy use of lambdas hoping they'll be inlined, so we should performance test MSVC and make sure it actually does

Finally, we haven't used them much before, so what are people's thoughts on formatting lambdas passed to functions? I went with this (indented like you would the content of a loop):
```cpp
s->m_pages.loopPages([this, s](uint32 page)
{
	s->m_erase_it[page] = m_map[page].InsertFront(s);
});
```
but we could also go with this
```cpp
s->m_pages.loopPages([this, s](uint32 page)
	{
		s->m_erase_it[page] = m_map[page].InsertFront(s);
	});
```
or this
```cpp
s->m_pages.loopPages(
	[this, s](uint32 page)
	{
		s->m_erase_it[page] = m_map[page].InsertFront(s);
	});
```
I personally would like to not go with clang-format's preference of this, it destroys horizontal space
```cpp
s->m_pages.loopPages([this, s](uint32 page)
                     {
                         s->m_erase_it[page] = m_map[page].InsertFront(s);
                     });
```
Same question for locals
```cpp
auto helper = [&](...)
{
};
```
or
```cpp
auto helper = [&](...)
	{
	};
```

PR #4941, opened 2021-10-26
GS: Use custom allocator for SW renderer

tellowkrinkle:
### Description of Changes
Some allocators really didn't like the SW renderer's allocation pattern
Replaces the system allocator with an overly specialized one that works very well at its one purpose

Hitman Blood Money title screen on i7-4980HQ with 2 SW renderer threads:
Windows:
29% → 31% (MTGS thread time in _aligned_malloc: 10% → 3.5%)
MacOS:
22% → 30% (MTGS thread time in _aligned_malloc: 30% → 3.5%)
Linux (default glibc allocator with Fedora 34):
22% → 30% (My perf flame graph was very broken, so no stats here)

There's one more malloc call from `GSOffset::GetPages`, which will get cleared up with #4348

(Note: I'm expecting the most performance increase on things that also benefitted from the spin wait PR)

### Rationale behind Changes
Faster software renderer, yay

### Suggested Testing Steps
Make sure everything works
Watch the console for `GSRingHeap: Refusing to grow to 128mb`, I threw in a maximum size so that if I messed anything up it wouldn't eat all the system's ram and crash, but I might have set it too low

### DIFF
--- a/pcsx2/CMakeLists.txt
+++ b/pcsx2/CMakeLists.txt
@@ -618,6 +618,7 @@ set(pcsx2GSSources
 	GS/GSLzma.cpp
 	GS/GSPerfMon.cpp
 	GS/GSPng.cpp
+	GS/GSRingHeap.cpp
 	GS/GSState.cpp
 	GS/GSTables.cpp
 	GS/GSUtil.cpp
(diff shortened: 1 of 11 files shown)


Shoegzer:
@TransparentBlue great results, what OS are you testing with?  This one is linux (with intel) as noted above.  While I haven't tested as comprehensively, I haven't encountered any negative perf regressions at least - all results are either the same or better than master, so pretty much in line with your conclusions.

refractionpcsx2 (approved):

Shoegzer:
Ran a few more tests yesterday and all seemed to be at least as performant as master, with some games such as Killzone seeming to show a modest speed improvement.

Is this dependent on another PR to merge?  Wondering what blockers exist at this point.

Mrlinkwii (approved):

PR #4930, opened 2021-10-23
DEV9: Allow Internal DHCP Server to find Gateway on Mac and FreeBSD

TheLastRar:
### Description of Changes
Allows Auto Gateway to function on Mac/FreeBSD

### Rationale behind Changes
Previously this would not function on those systems

### Suggested Testing Steps
Retracing the steps in issue #4828, except using FreeBSD or Mac as the operating system

Using the network access disc to test network connectivity while DHCP intercept is set & Gateway is set to auto

Testing other online games while DHCP intercept is set & Gateway is set to auto

Using the network setup utility found in various other games while DHCP intercept is set & auto Gateway is set

My testing ability on these OS is limited, so further testing is appreciated

PR #4944, opened 2021-10-26
DEV9: Add a socket based backend for DEV9 networking

TheLastRar:
### Description of Changes
Adds a socket based backend, similar to CLR_DEV9's Winsock backend

### Rationale behind Changes
Brings the integrated DEV9 closer to feature parity with CLR_DEV9.

This backend can be used with minimal configuration to join online/resurrected games

### Suggested Testing Steps
Test online games with the API set to Sockets

#### Regarding LAN games,
Hosting a game with Sockets is not expected to work,
You should allow you to join a game hosted by actual hardware or hosted by an instance of PCSX2 using TAP/Pcap.

#### Regarding Mac/FreeBSD
Auto adapter will not be able select an adapter, you currently must specify an adapter.
PR #4930 would be required to support this.

Handling of ICMP Echo packets (performed by the Network Access Disc) required platform specific code. I'm not able to test on these platforms so I have no guarantee that works correctly. If no testers are available for these platforms then I can disable support for this, as most games don't require this.


TheLastRar:
Qt Ui has been added

This PR should be ready

PR #4952, opened 2021-10-28
GS: Further state cleanup + fixes for older dump compatibility

refractionpcsx2:
### Description of Changes
Fixes compatibility with old savestates + remove PRMODE which is no longer used.  Also bumps savestate version.

### Rationale behind Changes
Recent changes temporarily broke compat with older states since it was merged before the PR was completely ready.

### Suggested Testing Steps
Test GS Dumps and savestates from PCSX2 to make sure they all work ok.

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -23,7 +23,7 @@
 int GSState::s_n = 0;
 
 GSState::GSState()
-	: m_version(6)
+	: m_version(7)
 	, m_gsc(NULL)
 	, m_skip(0)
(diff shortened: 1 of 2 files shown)


refractionpcsx2:
Of course, you can't go backwards..

prafullpcsx2:
Thanks for explanation. That means we lose ability to check for regressions unless we are going to ask users to make gs dumps using old pcsx2 versions. I hope the gains are far more than this loss.

refractionpcsx2:
Why can't you use the gsdump too in the new version?

PR #4965, opened 2021-10-30
GS: Let draw happen even if invalid, log invalid draws

### LINKED ISSUE #4948: Star Ocean battle transition effect broken with all renderers
### Describe the Bug

The battle transition effect when encountering an enemy on the filed is not rendered correctly in the latest builds, even in software mode. Not sure when this bug started.

### Reproduction Steps

Boot the game and run into an enemy on the field. 

### Expected Behavior

The effect to be rendered correctly especially in software mode. 

### PCSX2 Revision

v1.7.0-dev-1977-g23283b7f0a

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

### CPU

i7 8086k

### GPU

Nvidia GTX 1080ti

### GS Settings

I've tried software mode as well as default hardware mode settings. 

### Emulation Settings

I've tried various speedhack combinations and the defaults as well as all of them off. 

### GS Window Screenshots

The first image is how it currently renders and the second is how it used to render except my standard screenshot did not capture the ripple effect for some reason

Screenshot (13)
Screenshot (17)
. 

### Logs & Dumps

https://drive.google.com/file/d/1Qfb2bCurPM2ytnl7n-Na9kgALWTzPlWc/view?usp=sharing

refractionpcsx2:
### Description of Changes
Allow draws to happen regardless of frame/z format, log invalid draws that matter

### Rationale behind Changes
Star Ocean 3 was setting the Z Buffer to an invalid storage format, but the Z was masked and there was no depth test, so its format was irrelevant. Previously the draw got skipped, but if the draw is wrong, it's going to be broken anyway.

### Suggested Testing Steps
Test Star Ocean 3 battle transitions, uhh, not sure what else it'll affect, so just try random stuff to make sure it's okay I guess.

Fixes #4948

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -547,7 +547,7 @@ float GSState::GetTvRefreshRate()
 		case GSVideoMode::HDTV_1080I:
 			return 60;
 		default:
-			Console.Error("Unknown video mode. Please report: https://github.com/PCSX2/pcsx2/issues");
+			Console.Error("GS: Unknown video mode. Please report: https://github.com/PCSX2/pcsx2/issues");
 			return 0;
 	}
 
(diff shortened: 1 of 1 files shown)


pcsx2/GS/GSState.cpp
```diff
@@ -1433,37 +1433,38 @@ void GSState::FlushPrim()
 			ASSERT((int)unused < GSUtil::GetVertexCount(PRIM->PRIM));
 		}
 
-		if (GSLocalMemory::m_psm[m_context->FRAME.PSM].fmt < 3 && GSLocalMemory::m_psm[m_context->ZBUF.PSM].fmt < 3)
+		// If the PSM format of Z is invalid, but it is masked (no write) and ZTST is set to ALWAYS pass (no test, just allow)
+		// we can ignore the Z format, since it won't be used in the draw (Star Ocean 3 transitions)
+		bool ignoreZ = m_context->ZBUF.ZMSK && m_context->TEST.ZTST == 1;
```
lightningterror, line 1438: const would be nice.

pcsx2/GS/GSState.cpp
```diff
@@ -1433,37 +1433,38 @@ void GSState::FlushPrim()
...
+
+		if (GSLocalMemory::m_psm[m_context->FRAME.PSM].fmt >= 3 || (GSLocalMemory::m_psm[m_context->ZBUF.PSM].fmt >= 3 && !ignoreZ))
 		{
-			m_vt.Update(m_vertex.buff, m_index.buff, m_vertex.tail, m_index.tail, GSUtil::GetPrimClass(PRIM->PRIM));
+			Console.Warning("Possible invalid draw, Frame PSM %x ZPSM %x", m_context->FRAME.PSM, m_context->ZBUF.PSM);
```
lightningterror, line 1442: Maybe a prefix to the log `GS: Possible...`

pcsx2/GS/GSState.cpp
```diff
@@ -1433,37 +1433,38 @@ void GSState::FlushPrim()
...
 		{
-			Console.Error("Draw skipped due to invalid guest framebuffer format."
-			              "Please report: https://github.com/PCSX2/pcsx2/issues");
+			// could be an unsupported draw call
 		}
+		catch (const std::bad_alloc&)
+		{
+			// Texture Out Of Memory
+			PurgePool();
+			Console.Error("Memory allocation failure.");
```
lightningterror, line 1461: While we are at it why not this one too with GS: prefix too.

lightningterror (approved):
Looks good.

PR #4971, opened 2021-11-03
GS: Redo the Texture min/max opt

refractionpcsx2:
### Description of Changes
Reimplement the Texture Min Max optimisation to reduce texture sizes read

### Rationale behind Changes
Original code was slightly janky and didn't cover all scenarios, but it also didn't deal with erroneous situations, this should address that.

### Suggested Testing Steps
Test games, make sure no textures are corrupted, see if the performance is any better. Chronicles of Narnia, Test Drive Unlimited, Kaan are games known to be affected.

Note: The Test Drive Unlimited dump has a black bit on the road in the distance, I believe this is correct, the PS2 does this with distant road, but it is the correct colour.

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -2660,56 +2660,112 @@ void GSState::GetTextureMinMax(GSVector4i& r, const GIFRegTEX0& TEX0, const GIFR
 			__assume(0);
 	}
 
-	// Small optimisation left over from a much bigger (and broken) optimisation
-	// This makes Baldurs Gate 2 go brr. Without it, it's pretty slow due to cutting
-	// things up in to tiny pieces and the GS allocating a huge texture
-	if (wms == CLAMP_REPEAT && wmt == CLAMP_REPEAT)
+	bool skipClamp = false;
+
+	// If any of the min/max values are +-FLT_MAX we can't rely on them
+	// so just assume full texture.
+	if (m_vt.m_max.t.x >= FLT_MAX || m_vt.m_min.t.x <= -FLT_MAX ||
+		m_vt.m_max.t.y >= FLT_MAX || m_vt.m_min.t.y <= -FLT_MAX)
+		skipClamp = true;
+	
+	// Optimisation aims to reduce the amount of texture loaded to only the bit which will be read
+	if (!skipClamp)
 	{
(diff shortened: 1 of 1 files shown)


pcsx2/GS/GSState.cpp
```diff
@@ -2660,56 +2660,112 @@ void GSState::GetTextureMinMax(GSVector4i& r, const GIFRegTEX0& TEX0, const GIFR
...
+		switch (wms)
 		{
-			if (vr.x < u.x)
-				vr.x = u.x;
-			if (vr.z > u.z + 1)
-				vr.z = u.z + 1;
-		}
+			case CLAMP_REPEAT:
+			case CLAMP_REGION_REPEAT:
+					// This commented code cannot be used directly because it needs uv before the intersection
```
lightningterror, line 2721: Formatting.

pcsx2/GS/GSState.cpp
```diff
@@ -2660,56 +2660,112 @@ void GSState::GetTextureMinMax(GSVector4i& r, const GIFRegTEX0& TEX0, const GIFR
...
-		if (mask & 0xf000)
+		switch (wmt)
 		{
-			if (vr.y < v.y)
-				vr.y = v.y;
-			if (vr.w > v.w + 1)
-				vr.w = v.w + 1;
+			case CLAMP_REPEAT:
+			case CLAMP_REGION_REPEAT:
+					//if (uv_.y >> th == uv_.w >> th)
```
lightningterror, line 2749: Here too.

pcsx2/GS/GSState.cpp
```diff
@@ -2660,56 +2660,112 @@ void GSState::GetTextureMinMax(GSVector4i& r, const GIFRegTEX0& TEX0, const GIFR
...
-		// See commented code below for the meaning of mask
-		const GSVector4i u = uv & GSVector4i::xffffffff().srl32(32 - tw);
-		const GSVector4i v = uv & GSVector4i::xffffffff().srl32(32 - th);
+		// Checks for UV's going above the size of the texture (for wrapping)
+		if (wms == CLAMP_REGION_REPEAT)
+		{
+			u = uv & (maxu | minu);
+			uu = uv & ~(maxu | minu);
+		}
+		else if(wms == CLAMP_REPEAT)
```
lightningterror, line 2688: Ditto.

lightningterror (approved):

Issue #4963, opened 2021-10-30
[Feature Request]: Change "Clear ISO list" with "Delete ISO list" and add a "Clean ISO list" option at the ISO Selector


IlDucci:
### Description

Add an aditional option at CDVD -> ISO Selection -> ISO list, do the following additions:
 - Replace "Clear ISO list" with "Delete ISO list" or some other naming that makes it clear that the list itself will be emptied in its entirety.
 - Add a new option called "Clean ISO list", to be used to remove only any ISO files that are no longer available at their original location.

### Reason

This feature somewhat came to me because I mistranslated "Clear ISO list" as "Clean ISO list", and when I actually went to use it, I thought I would clean up all the ISOs I had moved from one place to another. Right now, having only the option to hit the nuclear button and redo the ISO gathering can be a bit tedious for people with multiple folders. A happy accident, I suppose.

### Examples

Cannot think on any examples here.

lightningterror:
Addressed in the linked pr.

PR #4975, opened 2021-11-04
Add "Clear missing files" option to recent ISO list

italodirenzo93:
### Description of Changes
Adds an option to the recent ISO list to remove all entries that can no longer be found in their original location.

### Rationale behind Changes
Solution for #4963. I decided on using the term "prune" instead of "clean" because it sounds too similar to the existing "clear". Would welcome input on what the terminology for this function should be.

### Suggested Testing Steps
1. Load an ISO file into PCSX2
2. In a separate file browser window, move that file to a different location on disk
3. Click the "Prune ISO list" option and the file will be removed from the list

### DIFF
--- a/pcsx2/gui/MainMenuClicks.cpp
+++ b/pcsx2/gui/MainMenuClicks.cpp
@@ -579,6 +579,23 @@ void MainEmuFrame::Menu_IsoClear_Click(wxCommandEvent& event)
 	}
 }
 
+void MainEmuFrame::Menu_IsoClearMissing_Click(wxCommandEvent& event)
+{
+	wxDialogWithHelpers dialog(this, _("Confirm clearing ISO list"));
+	dialog += dialog.Heading(_("This will remove all missing ISO files from the list. If an ISO is running it will remain in the list. Continue?"));
+
+	bool confirmed = pxIssueConfirmation(dialog, MsgButtons().YesNo()) == wxID_YES;
+
+	if (confirmed)
+	{
+		// If the CDVD mode is not ISO, or the system isn't running, wipe the CurrentIso field in INI file
+		if (g_Conf->CdvdSource != CDVD_SourceType::Iso || !SysHasValidState())
+			SysUpdateIsoSrcFile("");
+		wxGetApp().GetRecentIsoManager().ClearMissing();
+		AppSaveSettings();
+	}
+}
+
 void MainEmuFrame::Menu_Ask_On_Boot_Click(wxCommandEvent& event)
 {
 	g_Conf->AskOnBoot = event.IsChecked();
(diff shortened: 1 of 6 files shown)


lightningterror:
Maybe `Clear invalid ISO list` would make more sense.

IlDucci:
> 
> 
> Maybe `Clear invalid ISO list` would make more sense.

Or simply "Clear invalid ISO files". Maybe things could be made more obvious by referring to the entire ISO list (for the full removal) and to the individual ISO files (for the pruning of only any missing ISO files).

refractionpcsx2:
Or an even more user friendly way of writing it
`Clear missing files`

nf6429:
Agree with the above. "Clear invalid ISO" to an end user could indicate something is wrong with your file itself, not that it's just been moved, whilst "clear missing files" directly states what the issue is.

IlDucci:
> 
> 
> Agree with the above. "Clear invalid ISO" to an end user could indicate something is wrong with your file itself, not that it's just been moved, whilst "clear missing files" directly states what the issue is.

Agreed as well.

arcum42:
My main suggestion would be to show a list of what's going to be removed on the confirmation dialog box, before you actually remove it. Otherwise, I could see times where you could accidentally remove all or most of the entries because of something not being mounted at the moment...

RedDevilus:
```
* Game1
* Game2
* Game3
These will be removed from the listing. Do you confirm?
OK              Cancel
```

-----
Should be feasible to do something like this.

darkxex:
I find it very useful.

italodirenzo93:
I agree. Will update the PR with a new commit soon.

RedDevilus:
On your location you can't fully avoid WX but if you can, sure.

pcsx2/gui/MainMenuClicks.cpp
```diff
@@ -583,6 +583,46 @@ void MainEmuFrame::Menu_IsoClear_Click(wxCommandEvent& event)
...
+		for (const auto& file : missing_files)
+		{
+			missing_files_list += "* ";
+			missing_files_list += file.Filename;
+			missing_files_list += "\n";
+		}
+
+		dialog += dialog.Text(missing_files_list);
+
+		bool confirmed = pxIssueConfirmation(dialog, MsgButtons().YesNo()) == wxID_YES;
```
lightningterror, line 613: Const.

lightningterror (commented):
Commits should be squashed and properly named with prefixes.

lightningterror (approved):
Needs to be tested.

arcum42:
Works properly, from what I can see.

arcum42 (approved):

PR #4976, opened 2021-11-04
Restore Decode_XA.h

kenshen112:
### Description of Changes

Restores a file from pcsx. The structures were still in our code but had been strangely moved to cdrom.h where they don't belong
### Rationale behind Changes

Better code placement.
### Suggested Testing Steps

N/A

### DIFF
--- a/pcsx2/CDVD/Decode_XA.h
+++ b/pcsx2/CDVD/Decode_XA.h
@@ -13,7 +13,7 @@
  *  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#ifndef DECODEXA_H
+/*#ifndef DECODEXA_H
 #define DECODEXA_H
 
 typedef struct
@@ -31,8 +31,8 @@ typedef struct
 	s16 pcm[16384];
 } xa_decode_t;
 
-/*long xa_decode_sector(xa_decode_t* xdp,
+long xa_decode_sector(xa_decode_t* xdp,
 	unsigned char* sectorp,
-	int is_first_sector);*/
+	int is_first_sector);
 
-#endif
+#endif*/
--- a/pcsx2/CDVD/Decode_XA.h
+++ b/pcsx2/CDVD/Decode_XA.h
@@ -1,0 +1,38 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2021  PCSX2 Dev Team
+ *
+ *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
+ *  of the GNU Lesser General Public License as published by the Free Software Found-
+ *  ation, either version 3 of the License, or (at your option) any later version.
+ *
+ *  PCSX2 is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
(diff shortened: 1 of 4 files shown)


arcum42:
If those structures aren't used at all, wouldn't deleting them be the better code placement?

Also, am I missing something, or is the xa_decode_sector function in that header not actually defined anywhere in Pcsx2?

kenshen112:
No the function isn't defined. This would be the eventual lead into restoring CdXa support since we have ps1 mode. Assuming I can ever figure it out... Or someone else comes by

arcum42:
That makes a little more sense, though I'd probably comment out the function in the header until it actually has a matching function, presumably some variation of the code in here:
https://github.com/grumpycoders/pcsx-redux/blob/main/src/core/decode_xa.cc

My other concern, really, would be that this removes Xa from cdrStruct, and cdr, which is a cdrStruct, is frozen, so I could potentially see this being a savestate version bump...

refractionpcsx2:
Wouldn't it be better to restore this in a pr which restored xa functionality? Seems strange having a pr to restore a file to do absolutely nothing and it's not ideal for code cleanliness to have redundant stuff laying around. What if you never get around to restoring the xa functionality?

Anyway, that's my 2c

pcsx2/CDVD/Decode_XA.h
```diff
@@ -0,0 +1,28 @@
+//============================================
```
lightningterror, line 1: Missing copyright header.

pcsx2/CDVD/Decode_XA.h
```diff
@@ -0,0 +1,38 @@
...
+	s32 nsamples;
+	ADPCM_Decode_t left, right;
+	s16 pcm[16384];
+} xa_decode_t;
+
+/*long xa_decode_sector(xa_decode_t* xdp,
+	unsigned char* sectorp,
+	int is_first_sector);*/
+
+#endif
```
F0bes, line 38: Missing the newline at the end of file.

pcsx2/pcsx2.vcxproj
```diff
@@ -1170,4 +1171,4 @@
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets" />
-</Project>
+</Project>
```
F0bes, line 1174: Missing the newline at the end of file.

pcsx2/pcsx2.vcxproj.filters
```diff
@@ -2911,4 +2914,4 @@
       <Filter>AppHost\Resources</Filter>
     </Manifest>
   </ItemGroup>
-</Project>
+</Project>
```
F0bes, line 2917: Missing the newline at the end of file.

lightningterror (changes requested):
No need to have it included now if it does nothing, better comment it out in the file it currently is.

refractionpcsx2:
Remove the new file please.

kenshen112:
I'll make a seperate PR for that. This was to restore the header file. Which obviously wasn't accepted

PR #4980, opened 2021-11-05
Miscellaneous fixes for macOS

tellowkrinkle:
### Description of Changes
Fixes a number of random minor issues on macOS
Also increases the macOS deployment target so the CI stops complaining at people who use `std::optional`'s `value()` method

### Rationale behind Changes
Fixes things

### Suggested Testing Steps
Nothing in particular


stenzek (approved):
LGTM

PR #4981, opened 2021-11-05
GameDB: Add and fix SingStar Entries

RedDevilus:
### Description of Changes

We like higher GameDB numbers
### Rationale behind Changes

I noticed we had a ton of missing SingStar serial + minor other fixes.
### Suggested Testing Steps

Run SingStar games and look on the program log if it says a name and if it is , is it correct.

### DIFF
--- a/bin/GameIndex.yaml
+++ b/bin/GameIndex.yaml
@@ -1543,6 +1543,9 @@ SCED-52442:
 SCED-52452:
   name: "Official PlayStation 2 Magazine Demo 45"
   region: "PAL-M5"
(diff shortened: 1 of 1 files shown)


Mrlinkwii (approved):

refractionpcsx2:
I don't think "with microphone" is relevant, it's a list of games and if they are pre-release/promo etc, not if you got a microphone in the pack or not.  Please remove those.

RedDevilus:
Removed microphones/USB converter/game only mentions for SingStar.

PR #4982, opened 2021-11-05
GHActions: Don't shallow clone submodules

tellowkrinkle:
### Description of Changes
Shallow clones tend to be 5-10s slower

GHActions Windows submodule checkout run time (seconds):
image

Linux:
image

macOS (never used GH cache on submodules):
image

### Rationale behind Changes
Why waste 10 seconds when you don't have to?

### Suggested Testing Steps
Watch the CI run

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,23 +1,18 @@
 [submodule "3rdparty/xz/xz"]
 	path = 3rdparty/xz/xz
 	url = https://github.com/PCSX2/xz.git
-	shallow = true
 [submodule "3rdparty/gtest"]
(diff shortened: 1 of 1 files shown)


tellowkrinkle:
We can also do this by removing the shallow mark from `.gitmodules`, depending on whether we want to change the default for non-CI users checking out the repo.  Preferences?

lightningterror (approved):

xTVaser (approved):
this is really counter-intuitive.  Everything i read suggests that shallow clones are intuitively the right choice for temp CI builds, but also faster.  But I can't argue with the graphs :/

PR #4983, opened 2021-11-05
Debugger: Make the register list DPI aware (Windows)

F0bes:
### Description of Changes
On DPIs that weren't 100%, the register list would be cut off. 

This was 125% dpi for example

image

Note how you're missing the X column

This is the fixed version:

image

### Rationale behind Changes
The debugger was unusable for people running things other than 100% DPI

### Suggested Testing Steps
Test different DPIs and make sure the register list isn't cut off.


RedDevilus:
image

Width is fine for me but the `ACC` is not (above Memory Tab)

RedDevilus:
image

If I manually resize, I just have enough to see it.

image

RedDevilus:
image

image

RedDevilus (approved):

PR #4984, opened 2021-11-05
GameDB: Add and fix Power Pros series

Tokman5:
### Description of Changes

I added missing entries and fixed some entries with consistent title name.

### Rationale behind Changes

It makes GameDB better.
### Suggested Testing Steps

Check if nothing is broken.


Tokman5:
I tested running ``Jikkyou Powerful Pro Yakyuu 8`` and ``Jikkyou Powerful Major League 2009``, and it worked.

PR #4985, opened 2021-11-05
Create a shared pad header, get rid of the originals, and do some cleanup.

arcum42:
### Description of Changes
Initially, I created a new Gamepad.h header, made the rest of pcsx2 use that header, and put everything that was in both the Linux and Windows PAD.h files in it.

I then went to clean up the PAD.h files, and found I could actually get rid of what remained fairly easily, so I did so.

### Rationale behind Changes
It's one step towards sharing code between the two pad areas of code, and it also removes a lot of platform specific #ifdefs.

### Suggested Testing Steps
I've already compiled on both Windows and Linux, but testing to make sure nothing seems to be out of the ordinary gamepad-wise would be in order.


arcum42:
Still need to clean this up a bit. Looks like there are a few conflicts, and a few files were touched that shouldn't have been.

lightningterror (commented):
Are the translation changes intended?

arcum42:
> Are the translation changes intended?

No, it just modified them when I ran the build, and I didn't notice until after committing. Figuring out the best way to remove them from the commit was the main thing I have left on this besides trying to see if I missed anything in the rebase.

arcum42:
@lightningterror 
Managed to get rid of the language changes by checking out the previous copies of those two files, committing that, then squashing the commits together.

lightningterror (commented):
Renaming the commit with pad prefix would be nice.

arcum42:
> Renaming the commit with pad prefix would be nice.

Done.

PR #4986, opened 2021-11-05
PCSX2-GUI: Fix GUI inconsistencies

RedDevilus:
### Description of Changes

PCSX2-GUI: Fix Groupboxes to align on Sound and Gamepad and fix the look of DEV9
### Rationale behind Changes

It looks ugly
### Suggested Testing Steps

See if nothing breaks and looks nice.


pcsx2/DEV9/ConfigUI.cpp
```diff
@@ -184,16 +184,16 @@ class DEV9Dialog : public wxDialog
 		hdd_grid->Add(m_hdd_file,     wxSizerFlags().Expand());
 		auto* hdd_size_label = new wxStaticText(this, wxID_ANY, _("HDD Size (GiB):"));
 		auto* hdd_size_box = new wxBoxSizer(wxHORIZONTAL);
-		m_hdd_size_slider = new wxSlider(this, wxID_ANY, HDD_MIN_GB, HDD_MIN_GB, HDD_MAX_GB, wxDefaultPosition, wxDefaultSize, wxSL_HORIZONTAL | wxSL_MIN_MAX_LABELS | wxSL_AUTOTICKS);
+		m_hdd_size_slider = new wxSlider(this, wxID_ANY, 10, HDD_MIN_GB, HDD_MAX_GB, wxDefaultPosition, wxDefaultSize, wxSL_HORIZONTAL | wxSL_MIN_MAX_LABELS | wxSL_AUTOTICKS);
 		m_hdd_size_slider->SetPageSize(10);
-		for (int i = 15; i < HDD_MAX_GB; i += 5)
-			m_hdd_size_slider->SetTick(i);
+		for (int i = HDD_MIN_GB; i < HDD_MAX_GB; i += 5)
+			m_hdd_size_slider->SetTickFreq(10);
```
xTVaser, line 190: you shouldn't have to do this in a for-loop.  It looks like the original method was manually setting the ticks -- but the whole point of TickFrequency is specifying "hey i want you to put a tick every nth increment"

RedDevilus:
GamePad Settings (PAD):

Old:
image

New:
image

Audio Settings (SPU2):

Old:
image

New:
image

Network and HDD Settings (DEV9):
Old:
image

New:
image

RedDevilus:
I removed the min max labels, they only obscure the sight and could confuse people.
image

Mrlinkwii:
image
how it looks  on linux

pcsx2/DEV9/ConfigUI.cpp
```diff
@@ -184,16 +184,15 @@ class DEV9Dialog : public wxDialog
 		hdd_grid->Add(m_hdd_file,     wxSizerFlags().Expand());
 		auto* hdd_size_label = new wxStaticText(this, wxID_ANY, _("HDD Size (GiB):"));
 		auto* hdd_size_box = new wxBoxSizer(wxHORIZONTAL);
-		m_hdd_size_slider = new wxSlider(this, wxID_ANY, HDD_MIN_GB, HDD_MIN_GB, HDD_MAX_GB, wxDefaultPosition, wxDefaultSize, wxSL_HORIZONTAL | wxSL_MIN_MAX_LABELS | wxSL_AUTOTICKS);
+		m_hdd_size_slider = new wxSlider(this, wxID_ANY, 10, HDD_MIN_GB, HDD_MAX_GB, wxDefaultPosition, wxDefaultSize, wxSL_HORIZONTAL | wxSL_AUTOTICKS);
```
tellowkrinkle, line 187: The value you replaced with `10` is the initial value of the slider.  Putting a value less than the minimum just ends up with the minimum in a more confusing way.

pcsx2/DEV9/ConfigUI.cpp
```diff
@@ -184,16 +184,15 @@ class DEV9Dialog : public wxDialog
...
 		m_hdd_size_spin = new wxSpinCtrl(this, wxID_ANY, wxEmptyString, wxDefaultPosition, wxDefaultSize, wxSP_ARROW_KEYS, HDD_MIN_GB, HDD_MAX_GB, HDD_MIN_GB);
 		m_hdd_size_spin->SetSizeHints(m_hdd_size_spin->GetSizeFromTextSize(30));
-		hdd_size_box->Add(m_hdd_size_slider, wxSizerFlags(1).Centre());
-		hdd_size_box->Add(m_hdd_size_spin,   wxSizerFlags(0).Centre());
-		hdd_grid->Add(hdd_size_label, wxSizerFlags().Centre().Right());
-		hdd_grid->Add(hdd_size_box,   wxSizerFlags().Expand());
+		hdd_size_box->Add(m_hdd_size_slider, wxSizerFlags(4).Centre().Border(wxLEFT,4)); 
+		hdd_size_box->Add(m_hdd_size_spin,   wxSizerFlags(1).Centre().Border(wxLEFT,4)); 
+		hdd_grid->Add(hdd_size_label, wxSizerFlags().Centre().Right().Border(wxLEFT,4)); 
+		hdd_grid->Add(hdd_size_box,   wxSizerFlags().Expand().Border(wxLEFT,4));
```
tellowkrinkle, line 195: Please don't apply left borders to everything.  If you wanted to add space between the slider and spin control, you can `hdd_size_box->AddSpacer(5);` between the two controls.

I also don't see much point in messing with the proportions like that, there's no reason to grow the size box because it never needs to fit more digits than `HDD_MAX_GB`.  If you're finding that it doesn't fit that, adjust the size of the box by adjusting the number in the `GetSizeFromTextSize` call.

pcsx2/DEV9/ConfigUI.cpp
RedDevilus: Forgot I changed the value. Thanks.

pcsx2/DEV9/ConfigUI.cpp
RedDevilus: Probably why Linux/Mac looks so big as I am screwing with the window properties in janky method.

Mrlinkwii:
image
looks much better :+1:

RedDevilus:
Hopefully, latest revision as my VS didn't seem to like this file:
image

Mrlinkwii (approved):

RedDevilus:
A bit more white space above `Hard Disk Drive`. Should be fine as the final version.
image

IlDucci:
Can I dare to add some two cents?

> New:
> image

You might want to add a bit of space between the right side of the Interpolation dropdown box, so it matches the left side. You could also make all the three dropdown boxes of the Mixing Settings section equal in width.

RedDevilus:
@IlDucci 
image

Looks better? (Ignore the position of OK and Cancel, haven't changed that yet)

IlDucci:
Yes, that's much more uniform. Thank you!

I'd think that the only thing that doesn't look uniform now would be the right side of the Volume slider (compared with the Latency one).

RedDevilus:
I've already fixed it:

image

RedDevilus:
Old is left                               
New is right


Network and HDD:
image

Audio Settings:
image

Gamepad Settings:
image

RedDevilus:
Changed the max audio latency slider from 750 to 200. Also fixed the steps (with page down/page up/arrow keys when you select the slider instead of dragging it) when you go from TimeStretch to another mode , something I forgot due to how the code is duplicated on Windows. The Linux side is much saner in SPU2 so Windows will need a rewrite in the future. + 0.5f clamps for volume aren't probably needed either.

Default on TimeStretch:
image

Default on Async Mix:
image

Mrlinkwii:
differences in latency  MAX value linux 


master :
image
PR :
image

PR #4988, opened 2021-11-06
More preparation work for Qt

stenzek:
### Description of Changes
Adding new abstractions, moving stuff to core, getting rid of multiple sources of truth.

 - Adds new `FileSystem` wrappers which handle unicode paths on Windows, but store everything as UTF-8.
 - GameDatabase is now loaded on demand, and cached in a binary file (much faster and doesn't leak memory like yaml).
 - FPS/CPU usage is moved to a new PerformanceMetrics class instead of having it in 5 places.
 - Shaders, fonts and other resources are moved in to a `resources` directory in the exe folder, instead of being embedded in the exe. This means we can drop glib as a dependency later on for Linux, and less rc rubbish on Windows.
 - Gets rid of WX usage in CDVD and BIOS loading.

NOTE: When you're missing a BIOS, it'll no longer open the settings panel. Instead, you'll just get an error message. On the flipside, now if there's any valid BIOS file in the `bios` directory, it'll automatically use that. More idiot-proofing this way I hope..

### Rationale behind Changes
wx bad.

### Suggested Testing Steps

- Make sure game database related stuff hasn't broken. For example, game fix patches/widescreen patches.
- Make sure BIOS loading isn't broken with your existing config files.
- Usual stress tests for unicode correctness on windows.


lightningterror:
> Shaders, fonts and other resources are moved in to a resources directory in the exe folder, instead of being embedded in the exe. This means we can drop glib as a dependency later on for Linux, and less rc rubbish on Windows.

I actually prefer if they are embedded with the exe, less files to worry about, plus I wouldn't be surprised people not properly updating their dev builds and just partially updating with some old or missing files from previous versions.
It already happens with the gamedb.

seta-san:
> > Shaders, fonts and other resources are moved in to a resources directory in the exe folder, instead of being embedded in the exe. This means we can drop glib as a dependency later on for Linux, and less rc rubbish on Windows.
> 
> I actually prefer if they are embedded with the exe, less files to worry about, plus I wouldn't be surprised people not properly updating their dev builds and just partially updating with some old or missing files from previous versions. It already happens with the gamedb.

Agreed.

stenzek:
No, not going to happen.

1. Embedding stuff is a nightmare for development. You have to recompile all the time,.that's if it even detects the change in the place (it doesn't for me when I was editing shaders), and it takes forever with LTO.
2. Bloating the exe with megabytes of stuff is silly. Especially since we'll have avx/non avx and qt/wx variants. Sharing the data makes sense.
3. It breaks my game list cache since there's nothing to reference against to check if it needs to be rebuilt.
4. Game database or ws cheats can't be modified without recompiling, making it impossible for users.
4. If users are too stupid and expect the emu to be contained in one exe, then that's their problem. There's going to be tons of other issues. Even more so with Qt, since we'll have a bunch of dependent DLLs. Are you saying we should static link those and have exes in the order of 100mb in size? (Multiply by 2 for avx/nonavx)

refractionpcsx2:
If it's going to be a massive ballache and cause problems, then I say it's fine.  It's not like the build system is going to be a problem, all we have to do is remember to add the files to the build, then we're done, it's not a big deal.

As a bonus it makes the linux life easier.

So I'm fine with it.

stenzek:
It's globbed in cmake (for Mac) and the appimage script (Linux), so there's literally nothing you have to do to add a new file except drop it in.

In contrast to the current mess.....

arcum42:
I've never been the biggest fan of having files embedded in the exe. It's rather easier to work with on the development if it isn't. Any time I think about merging one of our nine billion config dialogs into the main one, it's always in the back of my mind that I'd have to grab an icon for it and figure out how to embed it...

lightningterror:
> No, not going to happen.
> 
> 1. Embedding stuff is a nightmare for development. You have to recompile all the time,.that's if it even detects the change in the place (it doesn't for me when I was editing shaders), and it takes forever with LTO.
> 2. Bloating the exe with megabytes of stuff is silly. Especially since we'll have avx/non avx and qt/wx variants. Sharing the data makes sense.
> 3. It breaks my game list cache since there's nothing to reference against to check if it needs to be rebuilt.
> 4. Game database or ws cheats can't be modified without recompiling, making it impossible for users.
> 5. If users are too stupid and expect the emu to be contained in one exe, then that's their problem. There's going to be tons of other issues. Even more so with Qt, since we'll have a bunch of dependent DLLs. Are you saying we should static link those and have exes in the order of 100mb in size? (Multiply by 2 for avx/nonavx)
> 6. The updating point is a non issue with automatic updates, which we will have in qt.

That's a valid explanation, in that case I'm in.
Maybe it would be nice to have some failsafes tho if the specified files are not found or an error reading them?

common/FileSystem.cpp
```diff
@@ -0,0 +1,1638 @@
...
+{
+#ifdef _WIN32
+	int filename_len = static_cast<int>(std::strlen(filename));
+	int mode_len = static_cast<int>(std::strlen(mode));
+	int wlen = MultiByteToWideChar(CP_UTF8, 0, filename, filename_len, nullptr, 0);
+	int wmodelen = MultiByteToWideChar(CP_UTF8, 0, mode, mode_len, nullptr, 0);
+	if (wlen > 0 && wmodelen > 0)
+	{
+		wchar_t* wfilename = static_cast<wchar_t*>(alloca(sizeof(wchar_t) * (wlen + 1)));
+		wchar_t* wmode = static_cast<wchar_t*>(alloca(sizeof(wchar_t) * (wmodelen + 1)));
```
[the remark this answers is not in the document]
stenzek: Avoids the heap allocation. Although not that opening file is particularly efficient anyway.. replaced all of them.

stenzek:
> Maybe it would be nice to have some failsafes tho if the specified files are not found or an error reading them?

It throws an error on startup if the resources directory doesn't exist. I guess we could check for specific files, but it seems silly to put checks for e.g. GS files in the GUI. It takes a special kind of skill to remove only some data files...

lightningterror (commented):
Would be nice to fix the new warnings
https://github.com/PCSX2/pcsx2/runs/4129597025?check_suite_focus=true#step:9:440
https://github.com/PCSX2/pcsx2/runs/4129597025?check_suite_focus=true#step:9:686
https://github.com/PCSX2/pcsx2/runs/4129597025?check_suite_focus=true#step:9:463
https://github.com/PCSX2/pcsx2/runs/4129597025?check_suite_focus=true#step:9:460
https://github.com/PCSX2/pcsx2/runs/4129597025?check_suite_focus=true#step:9:450

pcsx2/pcsx2.vcxproj.filters
```diff
@@ -2887,4 +2885,4 @@
       <Filter>AppHost\Resources</Filter>
     </Manifest>
   </ItemGroup>
-</Project>
+</Project>
```
refractionpcsx2, line 2888: Probably should put a new line so it doesn't moan :D

pcsx2/PerformanceMetrics.h
```diff
@@ -0,0 +1,42 @@
...
+	float GetAverageFrameTime();
+	float GetWorstFrameTime();
+
+	double GetCPUThreadUsage();
+	double GetCPUThreadAverageTime();
+	float GetGSThreadUsage();
+	float GetGSThreadAverageTime();
+	float GetVUThreadUsage();
+	float GetVUThreadAverageTime();
+} // namespace PerformanceMetrics
```
refractionpcsx2, line 42: New line again

refractionpcsx2:
BIOS files are showing multiple times and the string gets kinda busted sometimes, seems to repeat for all the files relating to that BIOS

image

refractionpcsx2:
Small request, if it's possible (shouldn't be too bad) can all the GS related stuff be put in a folder inside resources? Just to organise it away from the patch/db related stuff (which the user may modify), just to reduce temptation to screw with files they shouldn't.

stenzek:
> Bug1: BIOS files are showing multiple times and the string gets kinda busted sometimes, seems to repeat for all the files relating to that BIOS

Fixed.

> Bug 2: Create Blockdump doesn't seem to function

Fixed.

> Small request, if it's possible, can all the GS related stuff be put in a folder inside resources? Just to organise it away from the patch/db related stuff (which the user may modify), just to reduce temptation to screw with files they shouldn't.

Done, put the fonts into `fonts`, and GS shaders into `gs_dx11` and `gs_opengl` respectively.

arcum42:
Compiled in Linux, didn't immediately see any issues. Loaded Mana Khemia, and it said it found 'slus-21735' in the GameDB. Booted to the Bios without issue.

tellowkrinkle (commented):
Just some formatting things

pcsx2/GameDatabase.h
```diff
@@ -56,15 +54,10 @@ class GameDatabaseSchema
 		Full
 	};
 
-	struct Patch
-	{
-		std::string author;
-		std::vector<std::string> patchLines;
-	};
+	using Patch = std::vector<std::string>;
```
xTVaser, line 57: looks like we aren't bubbling up the author metadata for patches into the emulator anymore?  It doesn't really matter to me -- not sure if there are differing opinions on this though.

xTVaser (commented):
Did a pass over the code that I'm familiar with, it looks good to me.  I find the new GameDB code to be really complicated for what it's ultimately accomplishing (loading some data from a file) -- but I understand this is required for the caching to function properly, which is a nice performance boost.  None-the-less, it may read better if the bulk of the code that supported it was generalized and pulled out into `common` (lots of generic read/write functions added that aren't really GameDB specific).

On a clean local `Debug` build on 64bit (windows).  I hit the assertion at GSRendererHW.cppL918.

<details>
<summary>Stacktrace</summary>

```
 	ucrtbased.dll!00007ffe238ceaa5()	Unknown
 	ucrtbased.dll!00007ffe238ce8c3()	Unknown
 	ucrtbased.dll!00007ffe238d158f()	Unknown
>	pcsx2x64-avx2-dbg.exe!GSRendererHW::SwSpriteRender() Line 918	C++
 	pcsx2x64-avx2-dbg.exe!GSRendererHW::OI_JakGames(GSTexture * rt, GSTexture * ds, GSTextureCache::Source * t) Line 2323	C++
 	pcsx2x64-avx2-dbg.exe!GSRendererHW::Draw() Line 1523	C++
 	pcsx2x64-avx2-dbg.exe!GSState::FlushPrim() Line 1450	C++
 	pcsx2x64-avx2-dbg.exe!GSState::Flush() Line 1361	C++
 	pcsx2x64-avx2-dbg.exe!GSState::GIFRegHandlerTRXDIR(const GIFReg * r) Line 1325	C++
 	pcsx2x64-avx2-dbg.exe!GSState::Transfer<3>(const unsigned char * mem, unsigned int size) Line 1874	C++
 	pcsx2x64-avx2-dbg.exe!GSgifTransfer(const unsigned char * mem, unsigned int size) Line 435	C++
 	pcsx2x64-avx2-dbg.exe!SysMtgsThread::ExecuteTaskInThread() Line 428	C++
 	pcsx2x64-avx2-dbg.exe!Threading::pxThread::_try_virtual_invoke(void(Threading::pxThread::*)() method) Line 571	C++
 	pcsx2x64-avx2-dbg.exe!Threading::pxThread::_internal_execute() Line 671	C++
 	pcsx2x64-avx2-dbg.exe!Threading::pxThread::internal_callback_helper(void * itsme) Line 721	C++
 	pcsx2x64-avx2-dbg.exe!Threading::pxThread::_internal_callback(void * itsme) Line 710	C++
 	pcsx2x64-avx2-dbg.exe!ptw32_threadStart(void * vthreadParms) Line 184	C
```
</details>

arcum42:
Formatting seemed minor, so I'm thinking one of the main things we needed on this pr was a rebase?

refractionpcsx2:
There's a couple of other PR's which interfere with this one, though now I can't remember what they are, I think CK's Savestate PR (which he has a minor thing to fix I think)  and one of Tellows PR's, then it'll be good to rebase this, test it and merge it.

arcum42:
Alright. We'll probably want to see what we can do to get that rolling. because it'd be nice to get this code merged in.

pcsx2/ps2/BiosTools.cpp
```diff
@@ -31,147 +31,151 @@
...
+					region = 6;
+					break;
+				case 'C':
+					zone = "China";
+					region = 7;
+					break;
+				default:
+					zone.clear();
+					zone += romver[4];
+					break;
```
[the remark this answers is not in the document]
stenzek: Clang-format fun...

pcsx2/GameDatabase.h
stenzek: It'll still be in the file? But I don't see much use in loading it at runtime.

pcsx2/ps2/BiosTools.cpp
```diff
@@ -14,164 +14,171 @@
...
-				romver[12], romver[13],	// day
-				romver[10], romver[11],	// month
-				romver[6], romver[7], romver[8], romver[9],	// year!
-				(romver[5]=='C') ? "Console" : (romver[5]=='D') ? "Devel" : ""
-			);
+				romver[12], romver[13], // day
+				romver[10], romver[11], // month
+				romver[6], romver[7], romver[8], romver[9], // year!
+				(romver[5] == 'C') ? "Console" : (romver[5] == 'D') ? "Devel" :
+                                                                      "");
```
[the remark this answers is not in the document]
stenzek: clang-format fights me on this one.

xTVaser:
### Things I Tested
- Input Recording and Capturing
	- [x] InputRecording still functions properly
	- [x] as does exporting frames/audio
- File paths and unicode handling
  - [x] Ran the application in a directory containing unicode
  - [x] Booted an ISO with and without unicode paths
  - [x] Booted an ELF with and without unicode paths
  - [x] Memory cards with and without unicode paths
  - [x] Customized the folders in config with unicode and ensured they saved on exit and reload

> All of the above was tested with existing configuration, and then re-tested with fresh config to ensure backward's compatibility.  I also tested in Installed mode

- GameDB
  - [x] Ensured changes are picked up on an app restart
  - [x] GameFixes, Speedhacks, memcard filters and patches all seem to apply as expected.
  - [x] Invalid gamefixes/speedhacks are still complained about in the console log

The organization mentioned here doesn't seem to be done https://github.com/PCSX2/pcsx2/pull/4988#issuecomment-967027341 everything is in a single directory in `bin/resources` maybe im misunderstanding?

### Issues I Found
Unicode in the Game Name is displayed incorrectly in the console log header, likely needs to be converted to a wide-string on windows somewhere?
image

That's it, I can't find any other issues.  Nice work.

> This testing was all done on Windows 64bit Release

xTVaser:
The fix for the console title seems to be a small one:
AppCoreThread.cppL455:
`GameInfo::gameName = wxString::Format("%s (%s)", fromUTF8(game->name), game->region);`

or `UTF8StringToWxString`, whatever is the correct thing to use now.

xTVaser (approved):
We can always follow up on the two issues i mentioned, looks good to me other than conflicts from recently merged PRs.

stenzek:
> Unicode in the Game Name is displayed incorrectly in the console log title, likely needs to be converted to a wide-string on windows somewhere?

Fixed

> The organization mentioned here doesn't seem to be done More preparation work for Qt #4988 (comment) everything is in a single directory in bin/resources maybe im misunderstanding?

This seems to have been lost when I switched branches at some point. It's back in there now

xTVaser (approved):
Tested those two issues, they were fixed as stated.

Assuming that was the only thing that occurred in the rebase (organizing the resources dir and fixing the console log title unicode issue) this continues to look good to me.

JordanTheToaster:
Having unicode in the path for PCSX2 will cause widescreen patches to fail to load.

sl1pkn07:
seems need change the cheats_ws.zip path in the main CMakeLists.txt


https://github.com/PCSX2/pcsx2/blob/01855fc66aab7cf939b119311c9dd4ef07b138a4/CMakeLists.txt#L60-L63

if net, fail the installation when enable `PACKAGE_MODE`

~~~
-- Installing: /tmp/makepkg/sl1-pcsx2-git/pkg/sl1-pcsx2-git/usr/bin/PCSX2
CMake Error at cmake_install.cmake:66 (file):
  file INSTALL cannot find
  "/tmp/makepkg/sl1-pcsx2-git/src/pcsx2/bin/cheats_ws.zip": No such file or
  directory.


make: *** [Makefile:100: install] Error 1
~~~

in here https://github.com/PCSX2/pcsx2/pull/4988/commits/c8de8d861ec018f050f1530745b004b84fd09929

stenzek:
Package mode isn't something I care about, a linux packager can actually do something useful and fix it.

> Having unicode in the path for PCSX2 will cause widescreen patches to fail to load.

Fixed.

MarcoEstevez:
What happened?

refractionpcsx2:
The author decided he no longer wished to contribute this code due to various reasons. That's fine, I'm disappointed as this was a good step in the right direction, but I cannot force him to contribute, it's up to him.

guidodi:
Don't really know what happened but I hope you can work things out, Stenzek has been pretty active and making lots of commits lately.

PR #4989, opened 2021-11-06
wx: Fix `pxExplore` on macOS

tellowkrinkle:
### Description of Changes
Makes `pxExplore` work correctly on macOS

### Rationale behind Changes
It didn't work

### Suggested Testing Steps
Make sure you can still use the "Open in Explorer" buttons in the settings window

I would think `wxLaunchDefaultApplication` will work at least as well as the old solution on all platforms but who knows

### DIFF
--- a/pcsx2/gui/Panels/DirPickerPanel.cpp
+++ b/pcsx2/gui/Panels/DirPickerPanel.cpp
@@ -78,7 +78,7 @@ void Panels::DirPickerPanel::Explore_Click( wxCommandEvent &evt )
 		path.Mkdir();
 	}
 
-	pxExplore( path.ToString() );
+	wxLaunchDefaultApplication(path.ToString());
 }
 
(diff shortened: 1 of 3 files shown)


arcum42:
pxExplore is only ever used once. Couldn't we just get rid of it, and change that line of code?

arcum42:
Clicked browse in several places under Linux and it came up with a file browser in the right folders.

shivankmd763-dev:
> ### Description of Changes
> Makes `pxExplore` work correctly on macOS
> 
> It didn't work
> 
> ### Suggested Testing Steps
> Make sure you can still use the "Open in Explorer" buttons in the settings window
> 
> I would think `wxLaunchDefaultApplication` will work at least as well as the old solution on all platforms but who knows

https://github.com/PCSX2/pcsx2/pull/4989#issue-1046385708

PR #4999, opened 2021-11-07
CMake: Always default `CMAKE_BUILD_PO` off

tellowkrinkle:
### Description of Changes
Defaults the `CMAKE_BUILD_PO` option to off for all build modes

### Rationale behind Changes
Git files getting modified by build is just as much of a pain in release builds as it is in devel builds

### Suggested Testing Steps
Build in release and see if it modifies files tracked by git


F0bes (approved):
Works for me on Linux.

PR #5002, opened 2021-11-08
GS: add warning if skipdraw is enabled

Mrlinkwii:
### Description of Changes

Adds warning  if  skipdraw is enables since it can casuse bugs 
### Rationale behind Changes

 let the user know that s its enabled and anyone looking at the log 
### Suggested Testing Steps

boot  with skipdraw enabled and see the warning ,  enable skipdraw   mid game and see the  warning


lightningterror:
Honestly I don't see the need, users will still enable it if they want, and just creates more clutter in the logs.

PR #5013, opened 2021-11-12
GS: Fix Half Pixel Offset and Half Screen Fix config swap

tellowkrinkle:
### Description of Changes
Undoes an accidental swap of config options when switching the GS config to WX
\+ some more ui changes that didn't make it into the first PR

### Rationale behind Changes
It's nice when config options change the thing they say they change

### Suggested Testing Steps
Try using the config to change half pixel offset / half screen fix and make sure they change accordingly


arcum42:
Oops. Probably my mistake. Matching up the options to the checkboxes was a pain...

PR #5016, opened 2021-11-13
Debugger: Allow access to all of 0xBXXXXXXX

F0bes:
### Description of Changes
Show more of the kernel in the debugger memory view and disassembly views

### Rationale behind Changes
Give @Nobbs66 their sanity back
What is the point of blocking this memory anyways?

### Suggested Testing Steps
Try to access previously 'invalid' addresses such as ` 0xBFC000000`

### DIFF
--- a/pcsx2/DebugTools/DebugInterface.cpp
+++ b/pcsx2/DebugTools/DebugInterface.cpp
@@ -635,10 +635,7 @@ bool R5900DebugInterface::isValidAddress(u32 addr)
 		case 0xA:
 		case 0xB:
 			// [ 8000_0000 - BFFF_FFFF ] kernel
-			// We only need to access the EE kernel (which is 1 MB large)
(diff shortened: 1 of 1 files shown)


pcsx2/DebugTools/DebugInterface.cpp
```diff
@@ -635,8 +635,6 @@ bool R5900DebugInterface::isValidAddress(u32 addr)
 		case 0xA:
 		case 0xB:
 			// [ 8000_0000 - BFFF_FFFF ] kernel
-			// We only need to access the EE kernel (which is 1 MB large)
-			if (lopart < 0x100000)
 				return true;
 			break;
```
lightningterror, line 639: 1 tab less on return, and you won't need the break below anymore.

lightningterror (approved):

Issue #4879, opened 2021-10-06
[BUG]: Blokus Club with Bumpy Trot (Japan) - Various graphical issues on both renderers


HeroponRikiBestest:
### Describe the Bug

Various graphical issues are present in both the software and hardware renderers for Blokus Club with Bumpy Trot (Japan), mainly in software mode. I will combine the screenshots with this description, as there are multiple issues that I need to describe.
Software issues:
- Title screen is green when it should be white
SW renderer:
![image](https://user-images.githubusercontent.com/50224630/136131607-f46905da-ee83-4eec-bd0c-ad1983ccaab4.png)
HW renderer:
![image](https://user-images.githubusercontent.com/50224630/136131620-7888e1d8-f503-429f-91f3-bee20e57252f.png)
-Main menu is missing square icons next to menu choices
SW renderer:
![image](https://user-images.githubusercontent.com/50224630/136131642-178e49a9-084c-4f9e-821c-4282942298e3.png)
HW renderer:
![image](https://user-images.githubusercontent.com/50224630/136131647-141eb342-ef43-4e82-ac06-a5c97bc6caae.png)
- Character outlines are the wrong color (Green outside of a game, blue in a game, should be a lightish brown)
SW renderer:
![image](https://user-images.githubusercontent.com/50224630/136132244-54456ca5-3e3c-49c7-8619-9b323dc0a4ff.png)
HW renderer:
![image](https://user-images.githubusercontent.com/50224630/136132254-3ff98311-2577-43e5-96c7-f87ad802b9cd.png)
- Rendering issues on the water and sky in the coast stage
SW renderer:
![image](https://user-images.githubusercontent.com/50224630/136132554-9a0a2d22-0c88-4c6f-8e6e-890308e277c9.png)
HW renderer:
![image](https://user-images.githubusercontent.com/50224630/136132560-866c73c2-a383-4048-a18d-f6d67ca81024.png)

Hardware issues:
- During a match, some sort of graphical error appears in the top left. This is most noticeable on the coast stage, where it appears as 8 orange blocks with some clipping issues behind them
SW renderer:
![image](https://user-images.githubusercontent.com/50224630/136132437-85d2955a-983c-4e45-9796-0fa581d51134.png)
HW renderer:
![image](https://user-images.githubusercontent.com/50224630/136132453-453c2489-9899-4327-b6a7-41c8c845ab31.png)

Unsure which renderer has the issue:
[...]

PR #5024, opened 2021-11-16
GS SW: Handle flat prims without float conversion

### LINKED ISSUE #2980: SW GSdx: don't interpolate Z for flat triangle / Raw danger Menu
SW issue #1562 

Currently the code bypass rasterization of depth for sprite. We want to do the same for flat triangle. The main reason is to improve the accuracy of depth. Rasterization will convert an integer to float back and forth. It reduces depth accuracy which lead to depth fighting and bad rendering.

The ideal fix would be to use double precision. But it is quite expensive so flat triangle is a better trade-off.

Information to implement the code (use depth handling on sprite as an example)
* Flatness can be checked with
  * m_vt.m_eq.z
  * might need to call GSVertexTrace::CorrectDepthTrace to ensure correctness of above variable
* GSRendererSW::ConvertVertexBuffer
 * need a new template parameter
 * Insert the depth value in t.w
* Update JIT SetupCodeGenerator (all ISA)
  * add a new selector bit to select flat depth code path
  * load depth in m_local.p.z
* Update JIT DrawScanlineCodeGenerator (all ISA)
  * add a new selector bit to select flat depth code path
  * read depth from m_local.p.z in both Init and step
  * don't cast float to integer in TestZ
* Update non-JIT code
  * Same as JIT but in C++ (`GSDrawScanline.cpp`)
  * Uncomment `ENABLE_JIT_RASTERIZER` to disable JIT.

Note for all: be aware of Z overflow special case handling. You might need to handle flat triangle first

### LINKED ISSUE #3662: 007 Agent Under Fire (PAL): Title screen graphic errors in software mode
***PCSX2 version:***
v1.7.0-dev-208-g4629c8363

***PCSX2 options:***
None seem to make any difference other than switching to hardware rendering.

This is a known issue as it's documented in the wiki. However, there doesn't appear to be an open issue here, and as it's a problem with the software renderer I hope no-one minds me opening one.

Whatever the underlying cause is, the issue manifests like a priority or possibly blending problem. In software mode, the game's logo and some other elements are displayed behind, and obscured by, the moving "James Bond 007" sprites (?). In hardware mode everything displays correctly. GSDump attached. Ignore the missing text in the hardware shot; it was just taken slightly later.

Software mode:

gsdx_20200901225940

Hardware mode:

gsdx_20200901225952
 

Start the game.

None in the last several years at least.

***PC specifications:***
Windows 10 x64, Ryzen 7 3700x, Nvidia GTX 1080.

[AgentUnderFire.zip](https://github.com/PCSX2/pcsx2/files/5159075/AgentUnderFire.zip)

### LINKED ISSUE #4164: [BUG] Shinseiki Evangelion 2 (SLPS 25299): Missing graphics with the software renderer
**Describe the bug**

Some graphics are missing when using the software renderer. Most critically, all of the overlays (health bar, etc) are missing during battle scenes, and conveniently for demonstration purposes graphics are missing on the load game screen too. The game works fine with the OpenGL renderer, and DX11 is probably fine too as far as I can see. GS Dump attached.

EDIT: I just noticed this so I'll add it as it's possibly relevant. As you can see in the software renderer screenshot, the arrow in the bottom right corner is being obscured by the black area where a save slot should be. That arrow bounces up and down and should be on top of the save slot graphics, not behind them. Some kind of layering priority problem? Could it be related to #3662?

**To Reproduce**

Go to the load game screen or play through the game until you reach a battle.

**Expected behavior**

All graphics should be present, as when using the hardware renderers

**GS Settings**

No software renderer settings have any effect. 

**Emulation Settings**

No emulation settings have any effect as far as I can tell.

**GS Window Screenshots**

Software renderer:

gsdx_20210119134440

OpenGL renderer:

gsdx_20210119134443

 - PCSX2 Revision:  1.7.0-dev-903-gcf33fb3ca
 - OS: Windows 10 20H2
 - CPU: Ryzen 7 3700X
 - GPU: Nvidia RTX 3070

**Logs and Dumps**

[Evangelion2Missing.zip](https://github.com/PCSX2/pcsx2/files/5836234/Evangelion2Missing.zip)

refractionpcsx2:
### Description of Changes
Improve handling of large values for Z on the SW renderer by using integer Z for flat triangles.

### Rationale behind Changes
Large floats are not handled very well in the software renderer due to the range being limited by signed integers but also some precision is lost by being single floats which only have a precision of 24bits.  This PR makes it so flat triangles are treated like sprites and the Z values are passed as integer so no precision is lost, which fixes games which use flat triangles to draw UI/2D screens.

### Suggested Testing Steps
Test games in software mode, including 007: Agent Under Fire, make sure no graphics are missing or more broken than previous.

Fixes #3662
Fixes #4164
Also probably fixes issue with Raw Danger requiring integers on flat prims as mentioned in #1562 Fixes #2980
Also hugely improves the 2D UI in Hana to Taiyou to Ame to

### DIFF
--- a/pcsx2/GS/Renderers/SW/GSDrawScanline.cpp
+++ b/pcsx2/GS/Renderers/SW/GSDrawScanline.cpp
@@ -98,6 +98,7 @@ void GSDrawScanline::BeginDraw(const GSRasterizerData* data)
 	sel.fb = m_global.sel.fb;
 	sel.zb = m_global.sel.zb;
 	sel.zoverflow = m_global.sel.zoverflow;
+	sel.zequal = m_global.sel.zequal;
 	sel.notest = m_global.sel.notest;
 
 	m_sp = m_sp_map[sel];
@@ -143,13 +144,20 @@ void GSDrawScanline::SetupPrim(const GSVertexSW* vertex, const u32* index, const
 
 			if (has_z)
 			{
-				m_local.d8.p.z = dp8.extract32<2>();
-
-				GSVector8 dz = GSVector8::broadcast32(&dscan.p.z);
-
-				for (int i = 0; i < 8; i++)
+				if (sel.zequal)
 				{
-					m_local.d[i].z = dz * shift[1 + i];
+					m_local.p.z = vertex[index[1]].t.U32[3];
+				}
+
+				{
+					m_local.d8.p.z = dp8.extract32<2>();
+
+					const GSVector8 dz = GSVector8::broadcast32(&dscan.p.z);
+
+					for (int i = 0; i < 8; i++)
+					{
+						m_local.d[i].z = dz * shift[1 + i];
+					}
 				}
 			}
 		}
@@ -545,9 +553,14 @@ void GSDrawScanline::DrawScanline(int pixels, int left, int top, const GSVertexS
 
 				if (sel.prim != GS_SPRITE_CLASS)
 				{
+					// Need to handle when the float converts incorrectly
 					GSVector8 z = GSVector8::broadcast32(&scan.p.z) + zo;
 
-					if (sel.zoverflow)
+					if (sel.zequal)
+					{
+						zs = GSVector8i::broadcast32(&m_local.p.z);
+					}
+					else if (sel.zoverflow)
 					{
 						zs = (GSVector8i(z * 0.5f) << 1) | (GSVector8i(z) & GSVector8i::x00000001());
 					}
@@ -2787,7 +2800,7 @@ void GSDrawScanline::WritePixel(const T& src, int addr, int i, u32 psm)
 			*(u32*)dst = (src.U32[i] & 0xffffff) | (*(u32*)dst & 0xff000000);
 			break;
 		case 2:
-			*(u16*)dst = src.u16[i * 2];
+			*(u16*)dst = src.U16[i * 2];
 			break;
 	}
 }
--- a/pcsx2/GS/Renderers/SW/GSDrawScanlineCodeGenerator.all.cpp
+++ b/pcsx2/GS/Renderers/SW/GSDrawScanlineCodeGenerator.all.cpp
@@ -784,19 +784,24 @@ void GSDrawScanlineCodeGenerator2::Init()
 
 			if (m_sel.zb)
 			{
-				// z = vp.zzzz() + m_local.d[skip].z;
-				shufps(z, z, _MM_SHUFFLE(2, 2, 2, 2));
-				if (is64)
+				if (!m_sel.zequal)
 				{
-					addps(z, ptr[a1 + offsetof(GSScanlineLocalData::skip, z)]);
+					// z = vp.zzzz() + m_local.d[skip].z;
+					shufps(z, z, _MM_SHUFFLE(2, 2, 2, 2));
+					if (is64)
+					{
+						addps(z, ptr[a1 + offsetof(GSScanlineLocalData::skip, z)]);
+					}
+					else
+					{
+						movaps(ptr[&m_local.temp.z], z);
+						movaps(xym2, ptr[a1 + offsetof(GSScanlineLocalData::skip, z)]);
+						movaps(ptr[&m_local.temp.zo], xym2);
+						addps(z, xym2);
+					}
 				}
 				else
-				{
-					movaps(ptr[&m_local.temp.z], z);
-					movaps(xym2, ptr[a1 + offsetof(GSScanlineLocalData::skip, z)]);
-					movaps(ptr[&m_local.temp.zo], xym2);
-					addps(z, xym2);
-				}
+					pbroadcastdLocal(z, _rip_local(p.z));
 			}
 		}
 	}
@@ -995,7 +1000,11 @@ void GSDrawScanlineCodeGenerator2::Step()
 
 		if (m_sel.zb)
 		{
-			if (is32)
+			if (m_sel.zequal)
+			{
+				pbroadcastdLocal(z, _rip_local(p.z));
+			}
+			else if (is32)
 			{
 				broadcastssLocal(z, _rip_local_d_p(z));
 				addps(z, _rip_local(temp.zo));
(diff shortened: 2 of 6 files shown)


seta-san:
“ Also Fixes sorta #4164 but only in 64bit (which is okay since we're dumping 32bit soon anyway ;) ) ” shots fired

pcsx2/GS/Renderers/SW/GSDrawScanlineCodeGenerator.all.cpp
```diff
@@ -1188,22 +1188,35 @@ void GSDrawScanlineCodeGenerator2::TestZ(const XYm& temp1, const XYm& temp2)
...
 			// zs = (GSVector4i(z * 0.5f) << 1) | (GSVector4i(z) & GSVector4i::x00000001());
+			/*GSVector8 z = GSVector8::broadcast32(&scan.p.z) + zo;
+			z /= 2;
+			zs = GSVector8i(z, true);
+			zs = zs.min_u32(GSVector8i::x7fffffff());
+			zs = zs.sll32(1) | 1;*/
 
 			auto m_half = loadAddress(rax, &GSVector4::m_half);
+			auto m_one = loadAddress(rax, &GSVector4i::m_one);
+			auto  m_half_int_max = loadAddress(rax, &GSVector4i::m_half_int_max);
```
lightningterror, line 1199: Space after auto.

Immersion95:
Hi, just wanted to know if can solve KOF missing textures ? Is it a related/similiar kind of issue ? Thanks :)

refractionpcsx2:
No it probably won't help KoF, that is a precision issue, this just deals with a problem where converting INT_MAX to float then back to int makes it a value biggerr than INT_MAX, which is kinda dumb xD

Before it got around this by cutting off the bottom of the value, which is probably worse lol

pcsx2/GS/Renderers/SW/GSDrawScanlineCodeGenerator.all.cpp
```diff
@@ -1188,22 +1188,35 @@ void GSDrawScanlineCodeGenerator2::TestZ(const XYm& temp1, const XYm& temp2)
...
 		{
 			// zs = (GSVector4i(z * 0.5f) << 1) | (GSVector4i(z) & GSVector4i::x00000001());
+			/*GSVector8 z = GSVector8::broadcast32(&scan.p.z) + zo;
+			z /= 2;
+			zs = GSVector8i(z, true);
+			zs = zs.min_u32(GSVector8i::x7fffffff());
+			zs = zs.sll32(1) | 1;*/
 
 			auto m_half = loadAddress(rax, &GSVector4::m_half);
+			auto m_one = loadAddress(rax, &GSVector4i::m_one);
```
tellowkrinkle, line 1198: `loadAddress`  uses the given register, so doing it with the same register multiple times will result in all three being the address of `m_half_int_max`

This does explain the thing that only gets fixed on x64, on 32-bit `rax` isn't used as all addresses are in range, so 32-bit doesn't have any issues here.

You can get yourself constants like `7fffffff`, `80000000`, and `1` by using `cmpeqd` to get all ones, then shifting left/right

pcsx2/GS/Renderers/SW/GSDrawScanline.cpp
```diff
@@ -545,11 +545,14 @@ void GSDrawScanline::DrawScanline(int pixels, int left, int top, const GSVertexS
...
+					// Need to handle when the float converts incorrectly
 					GSVector8 z = GSVector8::broadcast32(&scan.p.z) + zo;
-
 					if (sel.zoverflow)
 					{
-						zs = (GSVector8i(z * 0.5f) << 1) | (GSVector8i(z) & GSVector8i::x00000001());
+						z *= 0.5f;
+						zs = GSVector8i(z, true);
+						zs = zs.min_u32(GSVector8i::x7fffffff());
+						zs = zs.sll32(1) | 1;
```
tellowkrinkle, line 555: I don't think this is correct
Let's say z is `10.f`
```
z *= 0.5f; // z = 5.f
zs = GSVector8i(z, true); // zs = 5
zs = zs.min_u32(GSVector8i::x7fffffff()); // zs = 5
zs = zs.sll32(1) | 1; // zs = 11(!)
```

tellowkrinkle (commented):
I'm confused as to how things were overflowing

The line in GSRendererSW.cpp should clamp z values to `0xffffff00`, which can successfully make the round trip without causing issues

pcsx2/GS/Renderers/SW/GSDrawScanline.cpp
refractionpcsx2: It's not correct, you're right, but before it cut off the lower 255, so if it was 255 it became 0 :P so although this is still wrong, it's less wrong, the ideal solution is to convert it all to integer, since it's impossible (without just using intrinsics) to convert to unsigned 32bit from float, which makes things more difficult.

In the case of 007, it draws triangles with UINT_MAX as the 32bit depth, then draws a sprite over the top, but it causes bad values to be left in the depth and it misses drawing half of the sprite, so the value from the triangle draw needs to be UINT_MAX and the only way to do that when we're dividing by 2 is to or in the extra 1.

But yeah, it's wrong, but it's less wrong than it was.

pcsx2/GS/Renderers/SW/GSDrawScanlineCodeGenerator.all.cpp
refractionpcsx2: Hmm that's a good point, I'll change that.

refractionpcsx2:
> I'm confused as to how things were overflowing
> 
> The line in GSRendererSW.cpp should clamp z values to `0xffffff00`, which can successfully make the round trip without causing issues

yes I removed the 0xffffff00 clamp because that was making the value too small, since the incoming value was UINT_MAX and it caused problems with large Z values, the issue is if you convert UINT_MAX to float, then use the cvttps2dq function to convert it back, it actually causes the value to be UINT_MAX+3

refractionpcsx2:
just noting this causes depth issues with a game called Hana to Taiyou to Ame to

refractionpcsx2:
ok this is finally ready for testing/review

pcsx2/GS/Renderers/SW/GSDrawScanline.cpp
```diff
@@ -143,13 +144,20 @@ void GSDrawScanline::SetupPrim(const GSVertexSW* vertex, const u32* index, const
...
+				{
+					m_local.p.z = vertex[index[1]].t.U32[3];
+				}
 
-				for (int i = 0; i < 8; i++)
 				{
-					m_local.d[i].z = dz * shift[1 + i];
+					m_local.d8.p.z = dp8.extract32<2>();
+
+					GSVector8 dz = GSVector8::broadcast32(&dscan.p.z);
```
lightningterror, line 155: Const

HeroponRikiBestest:
This fixes an issue with https://github.com/PCSX2/pcsx2/issues/4879 where some graphics didn't show up in the main menu. However, this exacerbates the exciting graphical issues with the KOF games. This screenshot/gsdump is from KOF 94 Rebout.
gs_20211203064859
[gs_20211203064859.gs.zip](https://github.com/PCSX2/pcsx2/files/7649364/gs_20211203064859.gs.zip)

refractionpcsx2:
KoF is back to how it was now, sorry about that, made a mistake restoring some old code.

This PR won't fix the KOF games since it's using triangles which are not at a flat depth, so the code in this PR won't touch it.

lightningterror (approved):
Needs testing.

HeroponRikiBestest:
This pull also improves depth handling in Hana to Taiyou to Ame to (Flower, Sun & Rain)

Mrlinkwii (approved):

PR #5026, opened 2021-11-17
GS-gui: Change blending option from None to Minimum.

lightningterror:
### Description of Changes

Change blending option from None to Minimum, update tooltip.
### Rationale behind Changes

Accurate representation. None doesn't actually fully disable sw blending, so the option is a bit inaccurate, rename it to Minimum instead.


RedDevilus:
Tooltip needs to be changed:
`"None:\nFast but introduces various rendering issues.\n"`

PR #5027, opened 2021-11-17
SPU: Fix videos in Stolen

Ziemas:
Stolen sets voice pitch to 0 to pause the voice, toggles key on, and only then DMA's the sample data to it. This means that grabbing the next sample must be delayed until the voice is unpaused by setting pitch to a non-zero value.

When i changed the comparisons for SP in order to fix the interpolation the behavior changed and it started grabbing the new block instantly on key on.

This restores the old behavior by initializing SP to -1 on key on.


LoStraniero91:
Tested and it works. Doesn't seem to break other games (for the ones I've tested).

PR #5031, opened 2021-11-18
Debugger: Make memory dialog transparent

F0bes:
Was supposed to do this in my initial memory search pr but ended up removing while fighting with git accidentally.

### Description of Changes
Makes the background of the memory area transparent

### Rationale behind Changes
image
Checkboxes / Labels are not the system colour in the memory search.

### Suggested Testing Steps
See if the check boxes are the system background colour for you.


F0bes:
image
Here's proof that it works :^)

PR #5032, opened 2021-11-18
USB: Gametrak/RealPlay

Florin9doi:
### Description of Changes
Gametrak and RealPlay controllers emulation. 

- Documentation : https://gist.github.com/Florin9doi/b952c3e210f93eb90b36fb4afbae14e1

- A PC gametrak (grey cable) is required. The PS2 versions (black cable) aren't yet compatible.
Untitled

Untitled


seta-san:
Kotaku was a week early. 😜

pcsx2/USB/usb-pad/dx/usb-pad-dx.cpp
```diff
@@ -57,8 +58,38 @@ namespace usb_pad
 				pad_copy_data(mType, buf, mWheelData);
 				return 5;
 			}
-
-			if (mType == WT_KEYBOARDMANIA_CONTROLLER)
```
RedDevilus, line 61: Was this a potential addition that was never completed?

seta-san: looks like a placeholder for the keyboardmania piano except there is already a place for keyboardmania below. i wonder if this would be more clear with a switch/case statement

Florin9doi: Still present at line 92

Florin9doi:
ready for review

seta-san:
what is needed for the ps2 versions?

refractionpcsx2:
Tried Real Racing, the controls aren't working at all for me, though going in to the USB options then coming out seems to press the "green" button, as far as the game cares, after briefly thinking it's been disconnected.

Florin9doi:
> what is needed for the ps2 versions?

USB passthrough support for PCSX2 or a device driver for each OS to enable the device.

> Tried Real Racing, the controls aren't working at all for me so I can't get past the menus, though going in to the USB options then coming out seems to press the "green" button, as far as the game cares, after briefly thinking it's been disconnected.
> I am mapping everything including Accelerate and Brake to the keyboard, so maybe the axis positions aren't working okay with the keyboard because they normally settle in the middle?

Are you using linux?

refractionpcsx2:
> Are you using linux?

Nope, Windows 10, just mapped all the buttons to the keyboard and they were pretty much unresponsive.

Florin9doi:
> Nope, Windows 10, just mapped all the buttons to the keyboard and they were pretty much unresponsive.

switch from Raw to Dinput

refractionpcsx2:
That worked better, but the accel XYZ don't seem to do anything and there's no wheel binding, so I can't steer :P

Dunno why Raw input doesn't work though, I'm not using raw input on pad

Shideravan:
I have the game but not the accessory. Is it possible to configure it to run using an XInput control or the keyboard? Or it can be playable only with the accessory? I can set a button for the foot mat, but not for the rest (or at least I don't know how to proceed)

seta-san:
right now as far as we know you would need the PC version of the accessory to play effectively

Shideravan:
Oh, that's sad. I'll open a request about the virtualization compatibility

PR #5035, opened 2021-11-20
GS: Improve FixedTEX0 accuracy

refractionpcsx2:
### Description of Changes
Improves the accuracy of TEX0 when calculating texture sizes

### Rationale behind Changes
Previously it expected the values to be too small and would overcompensate, but since we are now using Accurate STQ it gives good numbers and it broke this behaviour in Final Fantasy X (Possibly others)

### Suggested Testing Steps
Test games using upscaling, preferably 3x or 4x, make sure everything looks okay.

### DIFF
--- a/pcsx2/GS/GSDrawingContext.cpp
+++ b/pcsx2/GS/GSDrawingContext.cpp
@@ -56,7 +56,7 @@ static int findmax(int tl, int br, int limit, int wm, int minuv, int maxuv)
 
 static int reduce(int uv, int size)
 {
-	while (size > 3 && (1 << (size - 1)) >= uv + 1)
+	while (size > 3 && (1 << (size - 1)) >= uv)
 	{
 		size--;
 	}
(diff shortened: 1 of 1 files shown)


lightningterror (approved):
Needs testing, otherwise looks good.

JordanTheToaster (approved):

gilderoylockhart01:
I can help with the testing but I need a bit of an explaination into what I must check. Do I need to check for accurate STQ games or everything in general?

refractionpcsx2:
Uhh it's hard to say, basically the only thing I know what affected was FFX in cutscenes going weird, see here: https://forums.pcsx2.net/Thread-Final-Fantasy-X-Internal-Resolution-Scaling-Incorrect-During-Gameplay-Cutscenes
and it got worse with upscaling.

It basically just needs checking that textures look correct and not to big/small

Issue #3484, opened 2020-06-21
 /usr/local/lib/libatk-1.0.so: error adding symbols: file in wrong format


denn0n:
***PCSX2 version:***
 build v1.6.0
 
Fedora 30 x86_64 

I have fill all dependencies but still getting this complain if some one can help me i will appreciate 

Linking CXX shared module plugins/CDVDnull/libCDVDnull.so
FAILED: plugins/CDVDnull/libCDVDnull.so 
: && /usr/lib64/ccache/c++ -m32  -mfxsr -mxsave -march=native -pipe -fvisibility=hidden -pthread -fno-builtin-strcmp -fno-builtin-memcmp -mfpmath=sse -Wall -Wextra -Wno-attributes -Wno-unused-function -Wno-unused-parameter -Wno-missing-field-initializers -Wno-overloaded-virtual -Wno-unused-value  -Wstrict-aliasing -Wstrict-overflow=1  -D_FORTIFY_SOURCE=2  -Wformat -Wformat-security -ggdb3 -fno-omit-frame-pointer -DNDEBUG -DPCSX2_DEVBUILD -D_DEVEL  -O2    -std=c++11 -Wno-invalid-offsetof -pthread   -shared  -o plugins/CDVDnull/libCDVDnull.so plugins/CDVDnull/CMakeFiles/CDVDnull.dir/CDVD.cpp.o  /usr/lib/libglib-2.0.so  /usr/lib/libgobject-2.0.so  /usr/local/lib/libatk-1.0.so  /usr/lib/libgio-2.0.so  /usr/lib/libgthread-2.0.so  /usr/lib/libgmodule-2.0.so  /usr/lib/libgdk_pixbuf-2.0.so  /usr/local/lib/libpango-1.0.so  /usr/local/lib/libpangocairo-1.0.so  /usr/local/lib/libpangoft2-1.0.so  /usr/local/lib/libpangoxft-1.0.so  /usr/lib/libgdk-x11-2.0.so  /usr/lib/libgtk-x11-2.0.so && :
/usr/bin/ld: /usr/local/lib/libatk-1.0.so: error adding symbols: file in wrong format
collect2: error: ld returned 1 exit status

PC specifications:
AMD A8-7410 APU with AMD Radeon R5 Graphics 8 GB Ram

i add the full output  

```
./build.sh --cross-multilib
[...]

PR #5044, opened 2021-11-21
Pcsx2 fix cmake order

Ds886:
### Description of Changes
Change the linking order on Linux to prefer the system libraries in /usr/lib/

### Rationale behind Changes
I've encountered an issue in my environment where I had a stray SDL library installed on /usr/local/lib which wasn't included in the system ld.so.conf
In general the linking order seems comprehensive but only *BSD seems to utilize /usr/local/ as the system install directory while in Linux it is usually kept for user compiled modules(generally for development) or due to some hacked library in some package as far as I am aware

should resolve issue like 
https://github.com/PCSX2/pcsx2/issues/3484

### Suggested Testing Steps
I've tested  that it build in my machine after the issue was reproduced and on FreeBSD
I don't have a mac to test it on.


fantesykikachu:
Not saying this shouldn't be done but it does poses a separate issue; say the distribution's version of a library is too old to link, so the user compiles a more recent version, that version will be placed in /usr/local/lib/ , but with this ld will still try to link against the old version, and they will be stuck in another failure loop with very little information on how to fix it.

Ds886:
@fantesykikachu sorry for the late response was away from the comupter
This can be covered by allowing to specify the libdir  manually as a parameter and have instruction per distro to accommodate it
if someone have a special setup the person is in charge supporting it generally
In addition other project check during the configure process for the version(cmake process)

fantesykikachu:
> 3. In case of a special setup the person is in charge supporting it generally but this pull request should make it more persistent for different setup.

Actually your case would be considered the special setup as your stray SDL library is not following the unix standards, whereas my example is following the unix filesystem hierarchy.

Ds886:
I had a configuration issue but it should lead to a more stable build generally as the system libraries are more tested for interoperability and support.

As mentioned the total solution  for an inherent  stable build environment  will not come from this pull request but rather from:
1.  a better documentation of the building process(currently there is an outdated wiki and I was relying on the official arch build recipe and the AUR maintainer but also had to reverse a lot from CMakeLists) 
2. A more systematic view on the build targets and accommodate them in the cmake files.(for example currently the 3rd party repos are not friendly for compiling in general, havent fully looked but it is not really documented what to do to specialize the pathing)
3. make sure the building process is sufficiently customizable and documented for it's customization.

I will be happy to help documenting the building process on the repo for various platforms(might need assistance with mac as I don't have one) further on so such issues will be less likely to happen in the future or more systematize the build system and if needed add customization if missing though I take it shouldn't be all in the same basket.

lightningterror:
@Ds886 Is this still needed?

lightningterror:
Closing as it isn't being worked on and the issue it resolves seems closed as well.

PR #5053, opened 2021-11-22
Add crc hack for project metafalica (Again)

orbea:
### Description of Changes
Adds a crc for Project Metafalica which is a Ar Tonelico II relocalization. I am unsure why it was removed, but its still needed.

### Rationale behind Changes
Fixes https://github.com/PCSX2/pcsx2/issues/1762
See https://github.com/PCSX2/pcsx2/pull/1872

### Suggested Testing Steps
* Test that the world map is not broken in Project Metafalica (I did locally).

### DIFF
--- a/pcsx2/GS/GSCrc.cpp
+++ b/pcsx2/GS/GSCrc.cpp
@@ -25,6 +25,7 @@ const CRC::Game CRC::m_games[] =
 	{0xF95F37EE, ArTonelico2, US, 0},
 	{0x68CE6801, ArTonelico2, JP, 0},
 	{0xCE2C1DBF, ArTonelico2, EU, 0},
+	{0xC38067F4, ArTonelico2, NoRegion, 0}, // project metafalica 1.0
 	{0x2113EA2E, MetalSlug6, JP, 0},
 	{0x42E05BAF, TomoyoAfter, JP, PointListPalette},
 	{0x7800DC84, Clannad, JP, PointListPalette},


lightningterror:
https://github.com/PCSX2/pcsx2/pull/4082

orbea:
I would strongly prefer if this was not removed, its required to play the game and this is now the canonical English version of the game.

lightningterror:
No, only official discs released, no modifications.
But you can modify the file and compile locally.

orbea:
I really do not want to modify the file every time I update pcsx2 nor do I want to instruct everyone that wants to play the game to do that,  please reconsider.

gilderoylockhart01:
Well, to be honest I am a bit... mhee about crc hacks. At a time where we are massively removing them I feel we should continue that way. Why the problem can't be fixed in the GSDX code?

dio-gh:
This is quite an interesting predicament. Personally, I'm not sure about the following:

From the emulator's point of view:
- Why the CRCs of the game executables are relied on for hacks? In theory I can see this being important, but the current code seems to keep around a reductive mapping of CRCs to title enums, which to me suggests that the hacks that are in do not actually rely on this level of fine grained control. To me this suggests simply relying on product codes (`AAAA-00000`) would suffice, and this would resolve the issue of "romhacks changing the CRC thus making the hacks not trigger" in one fell swoop, too.
- Why are the mappings stored in the codebase as opposed to a config file? Seems a bit counter intuitive.

From the mod's point of view:
- You modify assets, cool, but then you also modify the executable. Why can't you turn the changes made to the executable into a game patch for PCSX2 to consume? I've only ever written patches for RPCS3, but I can't imagine the system here would be that much rougher.

arcum42:
> 
I'm just going to mention that this patch is actually patches to turn any of three disks (US, EU, and undubbed-US) into the same patched version, which retranslates all the text in the entire game, removes the dubs, and reencodes all the videos in the game. Apparently, the changes were large enough that it was impossible for them not to remove the dubs.

The project's here, for anyone that's interested: (And if you go to the about, you'll totally understand why people would want to play this version over the original. It's the localization that *should* have been done for the game...)
http://at2.metalbat.com/download/

orbea:
I would also like to add it fixes some game breaking bugs including a crash with a late game boss that happens on real hardware.

> Why the problem can't be fixed in the GSDX code?

That would be ideal, but it may not be easy going on the original discussion in the linked issue. If someone can do that then the hack could be removed for all of the versions of the game.

F0bes:
Closing, seems like the main issue here is #5062 and no one wants to add unofficial CRCs to the GS.

orbea:
@F0bes Please reconsider at least until a better solution is available.

PR #5054, opened 2021-11-22
Gs-hw: Adjust/minor optimization on sw blend shader a bit

lightningterror:
### Description of Changes

ogl:
Get rid of Color_pabe vector, makes more sense to use As and Cs in pabe code.

d3d11:
Use As and Cs in pabe code, makes more sense.
Get rid of Cv vector, use Color parameter for operations instead like on gl.
### Rationale behind Changes

Minor optimization.
### Suggested Testing Steps

Make sure hw renderer blending still works on both renderers.


stenzek:
I think this may result in different values being used in the comparison on original line 684 due to C being normally modified in lines 812-818. Was this intentional?

PR #5056, opened 2021-11-22
CDVD: Adjust DMA timing based on PS1 timings.

refractionpcsx2:
### Description of Changes
Adjusts the CDVD DMA (3) timing to that based on PS1 tested numbers.

### Rationale behind Changes
Bus width is 16bit on the PS2 between the DVD drive and the DSP, it is 8bit on PS1, so serves to reason the DMA would be twice the speed. (PS1 is 24 cycle per word)

### Suggested Testing Steps
test games, make sure they still work and don't die horribly.

Fixes Spongebob Lights, Camera, Pants!

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -1243,7 +1243,7 @@ __fi void cdvdReadInterrupt()
 				cdvd.Status = CDVD_STATUS_PAUSE;
 
 			cdvd.nCommand = 0;
-			//DevCon.Warning("Scheduling interrupt in %d cycles", cdvd.ReadTime - (cdvd.BlockSize / 4));
+			//DevCon.Warning("Scheduling interrupt in %d cycles", cdvd.ReadTime - ((cdvd.BlockSize / 4) * 12));
 			// Timing issues on command end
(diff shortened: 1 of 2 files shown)


seta-san:
totally fixes victorious boxers -  no longer requires fast cdvd - no delay, music continues into second round
fixes bob the builder - no longer need fast cdvd
fixes spongebob light camera pants - gets past loading screen and no gets past the swap in the second minigame 

didn't fix ecco the dolphin.

the other games in https://github.com/PCSX2/pcsx2/issues/4117 should be tested but i don't own.

refractionpcsx2:
Will you stop going on about Ecco, that will never be fixed by anything 📦 
I thought bob the builder worked on master, same for victorious boxers?

seta-san:
nevermind. didn't know victorious boxer and bob the builder were already fixed. i thought they both still needed fast cdvd

refractionpcsx2:
Not that I know of.  The main thing with this PR was fixing shit without breaking anything, which was being somewhat problematic, so I just went for something legitimate and it seemed to work well out of sheer luck I think lol

refractionpcsx2:
Silent Hill 2 video in the graveyard seems to be working for me as long as instant vu + mtvu is off, I guess that's a mild improvement, since master doesn't work regardless.

Issue #4674, opened 2021-08-17
[BUG]: Doko Demo Issho - Toro to Ippai (Japan) and Toro to Kyuujitsu (Japan) - Wireframe issues on OGL


HeroponRikiBestest:
### Describe the Bug

In Doko Demo Issho - Toro to Ippai (Japan); all Toros are rendered with their wireframe missing on OGL. Additionally, even in software mode, the black Toros seem to have texture issues with their eyes.
In Toro to Kyuujitsu (Japan); Toro is rendered with only his wireframe on OGL, as well as the 3d model of a hand that acts as your mouse. As far as I know, there are no black Toros in this game to check.

### Reproduction Steps

Doko Demo Issho - Toro to Ippai (Japan): Wait until the title screen appears
Toro to Kyuujitsu (Japan): Start a new game, get past the character customization menu.

### Expected Behavior

Models of Toro (and in the case of Toro to Kyuujitsu, the mouse/hand icon as well) should appear whole, with both the wireframe and the rest of the model intact. Additionally, the black toros should not have eye texture issues.

### PCSX2 Revision

PCSX2 1.7.0-20210817085737 - compiled on Aug 17 2021

### Operating System

Linux (64bit) - Specify Distro Below

### If Linux - Specify Distro

Linux Mint 20

### CPU

AMD Ryzen 7 PRO 4750U with Radeon Graphics

### GPU

Uses an iGPU, see cpu

### GS Window Screenshots

Doko Demo Issho - Toro to Ippai (Japan): Software mode (Eye texture issues visible)
![image](https://user-images.githubusercontent.com/50224630/129793592-b7f8aac3-6c61-4ce0-8621-f0f00d04a765.png)
Doko Demo Issho - Toro to Ippai (Japan): Hardware mode
![image](https://user-images.githubusercontent.com/50224630/129793635-d8668dfb-80f8-468b-a5c9-bc1258d0daa1.png)

Toro to Kyuujitsu (Japan): Software mode 
![image](https://user-images.githubusercontent.com/50224630/129793417-830e9eac-bbcf-49fe-a43c-5c507799d972.png)
Toro to Kyuujitsu (Japan): Hardware/OGL mode 
![image](https://user-images.githubusercontent.com/50224630/129793448-3b67c001-c2be-4d11-b8e9-2e66af6de1f4.png)

### Logs & Dumps

[...]

PR #5061, opened 2021-11-23
GS-hw: Improve how we handle AA1 draws

lightningterror:
### Description of Changes

There is no hw implementation of AA1.
Alpha blending shouldn't be enabled by AA1 flag, only ABE and PABE flags.
This caused alpha blending to be enabled when it shouldn't have.
Also disable Blending when AA1 and ABE or PABE are on on lines.
Lines aren't drawn so no need to do blending on them.

Mask out AA1 on triangles in GSState
AA1 is not supported on hw renderers so ignore flushing the prims on triangles.
Should provide a nice speed boost on games that use AA1 on triangle prims.
When AA1 flag is enabled Z is not written on lines since coverage is always less than 0x80.
### Rationale behind Changes

Accuracy, improves how we handle aa1 draws, while there's no actual aa1 hw implementation yet we can make sure the draws are handled properly.
Improves #4674
Master
image
PR
image

### Suggested Testing Steps

Test games which use AA1.

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -722,7 +722,13 @@ __forceinline void GSState::ApplyPRIM(u32 prim)
 {
 	if (GSUtil::GetPrimClass(m_env.PRIM.PRIM) == GSUtil::GetPrimClass(prim & 7)) // NOTE: assume strips/fans are converted to lists
 	{
-		if (m_env.PRMODECONT.AC == 1 && (m_env.PRIM.U32[0] ^ prim) & 0x7f8) // all fields except PRIM
+		u32 prim_mask = 0x7f8;
+		const bool is_hardware_renderer =
+			((theApp.GetCurrentRendererType() == GSRendererType::OGL_HW) || (theApp.GetCurrentRendererType() == GSRendererType::DX1011_HW));
+		if (is_hardware_renderer && GSUtil::GetPrimClass(prim & 7) == GS_TRIANGLE_CLASS)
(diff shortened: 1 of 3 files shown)


pcsx2/GS/GSState.cpp
```diff
@@ -722,7 +722,13 @@ __forceinline void GSState::ApplyPRIM(u32 prim)
 {
 	if (GSUtil::GetPrimClass(m_env.PRIM.PRIM) == GSUtil::GetPrimClass(prim & 7)) // NOTE: assume strips/fans are converted to lists
 	{
-		if (m_env.PRMODECONT.AC == 1 && (m_env.PRIM.U32[0] ^ prim) & 0x7f8) // all fields except PRIM
+		u32 prim_mask = 0x7f8;
+		const bool is_hardware_renderer =
+			((theApp.GetCurrentRendererType() == GSRendererType::OGL_HW) || (theApp.GetCurrentRendererType() == GSRendererType::DX1011_HW));
+		if (is_hardware_renderer && GSUtil::GetPrimClass(prim & 7))
```
refractionpcsx2, line 728: GSUtil::GetPrimClass(prim & 7) == GS_TRIANGLE_CLASS

RedDevilus:
This is Fifa 2002 (demo but i expect full game to be affected);
Old:
gs_20211124010038 1 7-2033
New:
gs_20211124005639 1 7-2076

Probably other games affected (not FFX as it is triangles not lines unfortunately)

PR #5067, opened 2021-11-24
GS/OpenGL: Use CreateRenderTarget() for temp HDR target

stenzek:
### Description of Changes
Probably won't change much, except maybe with sparse textures, where it'll use less video memory.

### Rationale behind Changes
It was wrong before.

### Suggested Testing Steps
Make sure colclip games are still fine (e.g. sly 2, ffx-2 opening)


lightningterror (approved):
Tested a few dumps that do hdr, they seem fine.

PR #5074, opened 2021-11-26
DEV9: Enable pcap non-blocking

TheLastRar:
### Description of Changes
Uses non-blocking mode for pcap

### Rationale behind Changes
When trying to close PCSX2, The Rx networking thread would hang trying to receive a packet
Use non-blocking mode to avoid the hang

### Suggested Testing Steps
Test network enabled games using pcap
Test intercept DHCP or the internal DNS, as non-blocking mode uses a different path (seemed fine in my testing)
Test performance between master and this PR when networking is enabled

### DIFF
--- a/pcsx2/DEV9/pcap_io.cpp
+++ b/pcsx2/DEV9/pcap_io.cpp
@@ -45,6 +45,7 @@ char errbuf[PCAP_ERRBUF_SIZE];
 
 int pcap_io_running = 0;
 bool pcap_io_switched;
+bool pcap_io_blocking;
 
 extern u8 eeprom[];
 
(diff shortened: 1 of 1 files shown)


F0bes:
Seems to fix the Waiting on RX thread problem. I don't have any networking home brew set up yet. Could someone test if networking still works on Linux?

F0bes (approved):
image
Works for me, and the DHCP intercept too.

pcsx2/DEV9/pcap_io.cpp
```diff
@@ -370,6 +379,12 @@ PCAPAdapter::PCAPAdapter()
 	host_mac = hostMAC;
 	ps2_mac = newMAC; //Needed outside of this class
 
+	if (pcap_io_init(config.Eth, config.EthApi == NetApi::PCAP_Switched, newMAC) == -1)
+	{
+		Console.Error("Can't open Device '%s'\n", config.Eth);
```
lightningterror, line 384: DEV9: prefix would be nice for the log, also remove the \n in the end.

TheLastRar: Done

Mrlinkwii (approved):

PR #5075, opened 2021-11-27
GS: Fix compiler warning

tellowkrinkle:
### Description of Changes
Fixes a memory leak we could have gotten if we used GSRingHeap::SharedPtr with exception-throwing constructors
(I don't think any of our constructors actually throw exceptions, but hey better safe than sorry)
Also clears an MSVC compiler warning

### Rationale behind Changes
Less compiler warnings

### Suggested Testing Steps
Look at CI, see if compiler warnings are gone
Run the SW renderer and make sure nothing breaks horribly

### DIFF
--- a/pcsx2/GS/GSAlignedClass.h
+++ b/pcsx2/GS/GSAlignedClass.h
@@ -38,6 +38,12 @@ public:
 		return ptr;
 	}
 
+	void operator delete(void* ptr, void* placement_ptr)
+	{
+		// Just here to satisfy compilers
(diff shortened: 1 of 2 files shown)


F0bes:
Linux x64

SW render works fine for me.

Code looks like magic to me but is functional :+1:

F0bes (approved):

Mrlinkwii (approved):

lightningterror (approved):
Tested a couple of dumps, seems fine.

PR #5076, opened 2021-11-27
GS: Add dedicated functions for choosing renderers and adapters

tellowkrinkle:
### Description of Changes
- Moves adapter list and default renderer selection to GSUtil.cpp, and out of GSwxDialog.cpp+GS.cpp
- Adds a separate option to be the default-selected adapter, rather than defaulting to adapter 0
- Enables the adapter selector list on all platforms.  I can disable this on platforms with no adapter-supporting graphics APIs if we'd prefer that.

### Rationale behind Changes
- This will be important for non-DX11, non-OGL APIs.
- On dual-gpu Macs, adapter 0 is the integrated GPU while adapter 1 is the discrete GPU, so it's probably better to let the API tell us what should be the default instead of picking 0 for ourselves

### Suggested Testing Steps
- Look at the graphics settings dialog and make sure everything still works
- Make sure F9 renderer switching still works

### DIFF
--- a/pcsx2/GS/GS.cpp
+++ b/pcsx2/GS/GS.cpp
@@ -1189,7 +1189,6 @@ void GSApp::Init()
 	// Avoid to clutter the ini file with useless options
 #ifdef _WIN32
 	// Per OS option.
-	m_default_configuration["adapter_index"]                              = "0";
 	m_default_configuration["CaptureFileName"]                            = "";
 	m_default_configuration["CaptureVideoCodecDisplayName"]               = "";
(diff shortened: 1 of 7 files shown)


JordanTheToaster:
Adapter selection and F9 is working correctly no issues with any graphics settings.

JordanTheToaster (approved):

Mrlinkwii (approved):

F0bes (approved):
Linux x64 KDE
image

Switching via GUI works.
Switching via F9 works.
Spamming  F9 is stable.

PR #5077, opened 2021-11-28
Fixes to CMake and version upgrade with CMake

kenshen112:
FindWayland.cmake: Changed incorrect use of `FIND_PACKAGE_HANDLE_STANDARD_ARGS` which is meant to link and set the main package to found rather then link individual libs

Added CMake 3.22.0 as max version. Added package version in CMakeLists as per standard CMP0048

### Description of Changes

Changes incorrect `FIND_PACKAGE_HANDLE_STANDARD_ARGS` to link wayland libs as imported targes. Increases the minimum Cmake version to include 3.11 and more modern 3.22 and sets "package version" as per https://cmake.org/cmake/help/latest/policy/CMP0048.html#cmp0048

### Rationale behind Changes

To use more modern CMAKE features for those with 3.22 and clear warnings related to that. To fix several incorrect linkages with wayland clearing warnings like
```CMake Warning (dev) at /usr/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (WAYLAND)
  does not match the name of the calling package (Wayland).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
  cmake/FindWayland.cmake:56 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  cmake/SearchForStuff.cmake:184 (find_package)
  CMakeLists.txt:31 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

### Suggested Testing Steps

See if it builds without those warnings


CMakeLists.txt
```diff
@@ -1,13 +1,19 @@
...
+endif (POLICY CMP0048)
+
 # Project Name
-project(Pcsx2)
+project(Pcsx2 VERSION 1.7)
 
 # Debian-based distributions require at least 2.8.5 due to multiarch.
 # Bumping up to 3.0 seems reasonable at this point, and will let us modernize
 # things a bit.
 #
```
F0bes, line 13: Remove this.

F0bes (changes requested):

F0bes (approved):
Gave it a spin with the Wayland API On.

- [x] Builds with no cmake warnings
- [x] Runs a game

Didn't notice anything immediately incorrect with the FindWayland script but I'm no cmake wizard.

cmake/FindWayland.cmake
```diff
@@ -31,29 +31,41 @@ IF (NOT WIN32)
...
 
 	include(FindPackageHandleStandardArgs)
 
-	FIND_PACKAGE_HANDLE_STANDARD_ARGS(WAYLAND_CLIENT  DEFAULT_MSG  WAYLAND_CLIENT_LIBRARIES  WAYLAND_CLIENT_INCLUDE_DIR)
-	FIND_PACKAGE_HANDLE_STANDARD_ARGS(WAYLAND_SERVER  DEFAULT_MSG  WAYLAND_SERVER_LIBRARIES  WAYLAND_SERVER_INCLUDE_DIR)
-	FIND_PACKAGE_HANDLE_STANDARD_ARGS(WAYLAND_EGL     DEFAULT_MSG  WAYLAND_EGL_LIBRARIES     WAYLAND_EGL_INCLUDE_DIR)
-	FIND_PACKAGE_HANDLE_STANDARD_ARGS(WAYLAND_CURSOR  DEFAULT_MSG  WAYLAND_CURSOR_LIBRARIES  WAYLAND_CURSOR_INCLUDE_DIR)
-	FIND_PACKAGE_HANDLE_STANDARD_ARGS(WAYLAND         DEFAULT_MSG  WAYLAND_LIBRARIES         WAYLAND_INCLUDE_DIR)
+	# FIND_PACKAGE_HANDLE_STANDARD_ARGS is just meant to find the main package and set package found. Not set variables or find individual libs
+	FIND_PACKAGE_HANDLE_STANDARD_ARGS(Wayland REQUIRED_VARS WAYLAND_LIBRARIES)
```
tellowkrinkle, line 68: This won't properly fail if it fails to find wayland.  It'll look, see that `WAYLAND_LIBRARIES` is indeed set (since it's set to `wayland::client;wayland::server;wayland::egl;wayland::cursor`), and be like "yep, all good"

You want to be checking things like `WAYLAND_SERVER_INCLUDE_DIR`, `WAYLAND_SERVER_LIBRARIES`, etc

cmake/FindWayland.cmake
```diff
@@ -31,29 +31,41 @@ IF (NOT WIN32)
...
+		add_library(wayland::egl STATIC IMPORTED)
+		set_target_properties(wayland::client PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${WAYLAND_EGL_INCLUDE_DIR}" INTERFACE_LINK_LIBRARIES "${WAYLAND_EGL_LIBRARIES}")
+	endif()
+	if (WAYLAND_CURSOR_INCLUDE_DIR AND WAYLAND_CURSOR_LIBRARIES AND NOT TARGET wayland::cursor)
+		add_library(wayland::cursor STATIC IMPORTED)
+		set_target_properties(wayland::cursor PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${WAYLAND_CURSOR_INCLUDE_DIR}" INTERFACE_LINK_LIBRARIES "${WAYLAND_CURSOR_LIBRARIES}")
+	endif()
 
-	set(WAYLAND_LIBRARIES ${WAYLAND_CLIENT_LIBRARIES} ${WAYLAND_SERVER_LIBRARIES} ${WAYLAND_EGL_LIBRARIES} ${WAYLAND_CURSOR_LIBRARIES})
+	set(WAYLAND_LIBRARIES wayland::client wayland::server wayland::egl wayland::cursor)
```
tellowkrinkle, line 61: While this technically works, that's not what people expect to happen with CMake

Variables like `WAYLAND_LIBRARIES` are part of usage of old-style CMake, where people expect that to be a list of `.so` files, which they can link to while including everything in `WAYLAND_INCLUDE_DIRS`.

tbh we don't have to support old-style CMake usage, we should be using imported targets for our own code anyways

cmake/FindWayland.cmake
```diff
@@ -31,29 +31,41 @@ IF (NOT WIN32)
...
-	FIND_LIBRARY(WAYLAND_EGL_LIBRARIES    NAMES wayland-egl      HINTS ${PKG_WAYLAND_LIBRARY_DIRS})
-	FIND_LIBRARY(WAYLAND_CURSOR_LIBRARIES NAMES wayland-cursor   HINTS ${PKG_WAYLAND_LIBRARY_DIRS})
+	FIND_LIBRARY(WAYLAND_CLIENT_LIBRARIES NAMES wayland-client)
+	FIND_LIBRARY(WAYLAND_SERVER_LIBRARIES NAMES wayland-server)
+	FIND_LIBRARY(WAYLAND_EGL_LIBRARIES    NAMES wayland-egl)
+	FIND_LIBRARY(WAYLAND_CURSOR_LIBRARIES NAMES wayland-cursor)
 
-	set(WAYLAND_INCLUDE_DIR ${WAYLAND_CLIENT_INCLUDE_DIR} ${WAYLAND_SERVER_INCLUDE_DIR} ${WAYLAND_EGL_INCLUDE_DIR} ${WAYLAND_CURSOR_INCLUDE_DIR})
+	if (WAYLAND_CLIENT_INCLUDE_DIR AND WAYLAND_CLIENT_LIBRARIES AND NOT TARGET wayland::client)
+		add_library(wayland::client STATIC IMPORTED)
```
tellowkrinkle, line 45: People generally expect the namespace name to match the package name, which in this case would have a capital `W`

The part after the `::` can technically be whatever, but I'd prefer if those were also capitalized

cmake/FindWayland.cmake
```diff
@@ -2,10 +2,10 @@
...
 # This will define:
 #
-#   WAYLAND_FOUND       - True if Wayland is found
-#   WAYLAND_LIBRARIES   - Link these to use Wayland
-#   WAYLAND_INCLUDE_DIR - Include directory for Wayland
-#   WAYLAND_DEFINITIONS - Compiler flags for using Wayland
+#   Wayland_FOUND       - True if Wayland is found
+#   Wayland_LIBRARIES   - Link these to use Wayland
+#   Wayland_INCLUDE_DIR - Include directory for Wayland
+#   Wayland_DEFINITIONS - Compiler flags for using Wayland
```
tellowkrinkle, line 8: If you delete `Wayland_LIBRARIES` and `Wayland_INCLUDE_DIR`, edit this accordingly

And document the imported targets we now make, e.g.
```
# Wayland::Client     - An imported target for Wayland Client
etc
```

tellowkrinkle (commented):
Since you're adding imported targets, could you also edit `pcsx2/CMakeLists.txt` to stop looking at things like `WAYLAND_CLIENT_FOUND` and `${WAYLAND_CLIENT_LIBRARIES}` and just link against `Wayland::Client`?

cmake/FindWayland.cmake
```diff
@@ -33,27 +31,38 @@ IF (NOT WIN32)
...
-	set(WAYLAND_INCLUDE_DIR ${WAYLAND_CLIENT_INCLUDE_DIR} ${WAYLAND_SERVER_INCLUDE_DIR} ${WAYLAND_EGL_INCLUDE_DIR} ${WAYLAND_CURSOR_INCLUDE_DIR})
+	FIND_LIBRARY(WAYLAND_CLIENT_LIBRARIES NAMES wayland-client HINTS ${PKG_WAYLAND_INCLUDE_DIRS})
+	FIND_LIBRARY(WAYLAND_SERVER_LIBRARIES NAMES wayland-server HINTS ${PKG_WAYLAND_INCLUDE_DIRS})
+	FIND_LIBRARY(WAYLAND_EGL_LIBRARIES    NAMES wayland-egl HINTS ${PKG_WAYLAND_INCLUDE_DIRS})
+	FIND_LIBRARY(WAYLAND_CURSOR_LIBRARIES NAMES wayland-cursor HINTS ${PKG_WAYLAND_INCLUDE_DIRS})
 
-	set(WAYLAND_LIBRARIES ${WAYLAND_CLIENT_LIBRARIES} ${WAYLAND_SERVER_LIBRARIES} ${WAYLAND_EGL_LIBRARIES} ${WAYLAND_CURSOR_LIBRARIES})
-
-	list(REMOVE_DUPLICATES WAYLAND_INCLUDE_DIR)
+	#list(REMOVE_DUPLICATES WAYLAND_INCLUDE_DIR)
```
tellowkrinkle, line 42: If `WAYLAND_INCLUDE_DIR` isn't a thing, I don't see why we would want this here

tellowkrinkle (approved):
This enables many years worth of CMake policies so we should get some testing on Linux to make sure nothing broke

kenshen112:
Tested between Xorg and Wayland. Package built correctly and on basic use with loading a game seemed functional as it should be. Wayland was active as noted by the pause and resume tested resulting in the current black window glitch speed seemed fast enough for the earlier fixes to be active nearing full-speed at times. But sluggish on Nvidia compared to Xorg still.

F0bes:
https://user-images.githubusercontent.com/29295048/143981017-3511f291-695c-426c-b05c-e04afcd945e6.mp4

Doesn't happen on master, both renderers, SW mode settings do not matter.
Happens with or without `-DWAYLAND_API=on` passed to cmake.

tellowkrinkle:
> Doesn't happen on master, both renderers, SW mode settings do not matter. Happens with or without `-DWAYLAND_API=on` passed to cmake.

Can you check which commit does it?  I would expect it's the CMake minimum version one but just to make sure

F0bes:
> > Doesn't happen on master, both renderers, SW mode settings do not matter. Happens with or without `-DWAYLAND_API=on` passed to cmake.
> 
> Can you check which commit does it? I would expect it's the CMake minimum version one but just to make sure

Nevermind, I just got unlucky when testing this PR. After many attempts you can reproduce it on master.

kenshen112:
As a side note. This works in the AUR environment as well

kenshen112:
Anything else needed to merge this?

Mrlinkwii (approved):
runing X11 and cmake 3.16.3 ,  the PR complied fine and  worked like master

PR #5082, opened 2021-11-29
Revert "GS: Remove Unofficial CRCs (#4082)"

orbea:
### Description of Changes
This reverts commit 6bcdb55f22d468b2eb8b95fd35bd071f9272a89c.

### Rationale behind Changes
These games hacks are still used and can be removed again when a better solution is available. Removing them before that is the case only regresses users.

### Suggested Testing Steps
N/A

### DIFF
--- a/pcsx2/GS/GSCrc.cpp
+++ b/pcsx2/GS/GSCrc.cpp
@@ -22,6 +22,8 @@ const CRC::Game CRC::m_games[] =
 {
 	// Note: IDs 0x7ACF7E03, 0x7D4EA48F, 0x37C53760 - shouldn't be added as it's from the multiloaders when packing games.
 	{0x00000000, NoTitle, NoRegion, 0},
+	{0xF46142D3, ArTonelico2, NoRegion, 0},
+	{0xC38067F4, ArTonelico2, NoRegion, 0}, // project metafalica 1.0
 	{0xF95F37EE, ArTonelico2, US, 0},
 	{0x68CE6801, ArTonelico2, JP, 0},
 	{0xCE2C1DBF, ArTonelico2, EU, 0},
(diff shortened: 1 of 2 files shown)


refractionpcsx2:
You can wait until there's a better solution available, we're not readding them.

F0bes:
Just because I didn't respond to you doesn't mean create another pr to revert a commit the team is absolutely fine with.

orbea:
@F0bes @refractionpcsx2 Why can't this just be removed when another solution is available? What are the negatives? As far as I can tell you are just hurting users for no good reason.

F0bes:
> If you are not willing to have a conversation you do not leave much choice.

At what point does a romhack get popular enough to hardcode it into PCSX2?
Does Johnny's romhack that incorporates farting noises every time there is an M1 Garand Ping in Call of Duty get added to PCSX2? It's an all or nothing situation and the consensus within the team is to **not** treat _all non official roms_ as special.

orbea:
Your example doesn't apply in this case. I am pretty sure none of the hacks in this commit are so pointless, but they are popular versions of games that people actually use. At least one of them is literally the only non-broken version of the game on real hardware or not...

RedPanda4552:
The original purpose of removing unofficial CRCs is that:

1) Doing so sets the standard that all romhacks must be supported, which means making code modifications for all romhacks on this planet which might use CRC hacks.
2) Removing CRC hacks for unofficial CRCs will in turn create pressure for problems to actually be solved instead of CRC hacked.

If you disagree with these, you can build from source or use a fork which has not merged this commit. The project stance is what it is, and this is now multiple people who have communicated this.

refractionpcsx2:
If you really need these CRC's in PCSX2 to play your rom hack, feel free to use an old version of the emulator.  These aren't going back in, end of story.

PR #5086, opened 2021-11-30
Switch to SDL_GameControllerRumble for rumble

tellowkrinkle:
### Description of Changes
- Switches to SDL_GameControllerRumble api for rumble
- Adds some extra SDL hints
- Adds support for users adding their own hints to PAD.ini

### Rationale behind Changes
According to https://github.com/libsdl-org/SDL/issues/4435#issuecomment-896432954, use of haptic interfaces for game controllers is deprecated.
The SDL_GameControllerRumble api is supported by a number of controllers that haptic is not.

### Suggested Testing Steps
(Linux only) Test all your controllers and make sure there are none that lose rumble support


tellowkrinkle:
...and it looks like we'll need to upgrade the CI

arcum42:
The TestForce function, which rumbles when you turn on "Enable Rumble" under "Gamepad Settings", "Gamepad Configuration", rumbles for way too long with the new api, and the number passed to SDL_GameControllerRumble likely needs to be rather less than 400. If you try it in the old version and the new one, you'll see the difference.

There are also two spots with "#if SDL_MINOR_VERSION >= 4" where it'll now always be more than 2.0.4, so doesn't need the #if/#endif around it.

tellowkrinkle:
> If you try it in the old version and the new one, you'll see the difference.

I don't have any controllers supported by the old API, so I'll have to leave getting the right number to someone who does

arcum42:
Turns out the only way I can get it to stop rumbling at a reasonable time is by doing this:
`bool JoystickInfo::TestForce(float strength = 0.60)
{
	u16 u16strength = static_cast<u16>(0x7fff * strength);

	int rumble = SDL_GameControllerRumble(m_controller, u16strength, u16strength, 1);
	sleep(1);
	SDL_GameControllerRumble(m_controller, 0, 0, 1);

	return rumble >= 0;
}`

Just changing it to 1 was better, but you're still left with a controller vibrating a fair bit after you get out of the dialog.

Of course, it does add a slight delay on the checkbox checking...

tellowkrinkle:
@arcum42 See if the latest commit works better

arcum42:
That one works exactly as I'd expect it to work. Looking good.

arcum42 (approved):

.github/workflows/scripts/linux/install-packages.sh
```diff
@@ -24,9 +24,9 @@ declare -a PCSX2_PACKAGES=(
   "libpng-dev"
   "libpulse-dev"
   "librsvg2-dev"
-  "libsdl2-dev"
```
arcum42, line 27: Wouldn't we still need sdl 2 in the list of libraries?

tellowkrinkle: We're compiling it ourselves now because the package manager's one is too old

arcum42: Ah, alright. Arch has a new enough version, so I hadn't noticed. Think we're probably good to merge...

PR #5087, opened 2021-11-30
GS: Unify HW texture caches

tellowkrinkle:
### Description of Changes
- Converts a number enums to enum classes, with arguments switching to typed values instead of `int`
- Stores a known enum in the `format` field of textures, removing issues where e.g. no one outside of `GSDeviceOGL` could actually understand what the format of a `GSTextureOGL` meant even though they could read it.  Now anyone can ask a `GSTexture*` for its format and understand what it is.
- Adds a `DownloadTexture` and an optional `DownloadTextureConvert` method to `GSDevice`, which downloads a texture and maps it to CPU memory
- Uses the above changes to add generic, shared versions of all the `virtual` `GSTextureCache` methods to `GSTextureCache` and get rid of the separate subclasses per subclass of `GSDevice`
- Renames convert shaders to not be `ps_convert##` because it's impossible to figure out what's what when only looking at the list of shaders

### Rationale behind Changes
Reduce per-renderer code so that adding a new renderer doesn't add as much maintenance cost

### Suggested Testing Steps
Make sure renderers still work


pcsx2/GS/Renderers/OpenGL/GSRendererOGL.cpp
```diff
@@ -1571,7 +1577,7 @@ void GSRendererOGL::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 	if (hdr_rt)
 	{
 		const GSVector4 sRect = GSVector4(commitRect) / GSVector4(rtsize.x, rtsize.y).xyxy();
-		dev->StretchRect(hdr_rt, sRect, rt, GSVector4(commitRect), ShaderConvert_MOD_256, false);
+        dev->StretchRect(hdr_rt, sRect, rt, GSVector4(commitRect), ShaderConvert::MOD_256, false);
```
lightningterror, line 1580: Tabs.

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -1540,7 +1540,7 @@ GSTextureCache::Source* GSTextureCache::CreateSource(const GIFRegTEX0& TEX0, con
 		// copy. Likely a speed boost and memory usage reduction.
 		bool linear = (TEX0.PSM == PSM_PSMCT32 || TEX0.PSM == PSM_PSMCT24);
 
-		if ((sRect == dRect).alltrue() && !shader)
+		if ((sRect == dRect).alltrue() && shader != ShaderConvert::COPY)
```
refractionpcsx2, line 1543: I did mention on discord, but incase it gets lost, check should be == not !=

JordanTheToaster:
Tested with my selection of games on both DX11 and OpenGL and both functioned correctly with no issues I could observe.

refractionpcsx2:
Ran it through a bunch of blockdumps on DX and OGL, looks good to me :)

refractionpcsx2 (approved):

lightningterror (approved):
Ran through some dumps, everything seemed fine.

pcsx2/GS/res/glsl/common_header.glsl
```diff
@@ -91,10 +92,10 @@ layout(std140, binding = 21) uniform cb21
 
     vec4 MinMax;
 
-    vec2 TextureScale;
+    vec2 pad1_cb21;
     vec2 TC_OffsetHack;
 
-    vec3 pad1_cb21;
+    vec3 pad2_cb21;
```
lightningterror, line 98: Hmm, we have 6 total free vectors, this can be optimized.

PR #5099, opened 2021-12-04
GameDB / Memcards - Replace `yaml-cpp` with `rapidyaml`

xTVaser:
### Description of Changes

- Replaces our YAML library with rapidyaml, a significantly faster and more actively maintained library without by-design memory leaks (to my knowledge)
  - it is MIT licensed, and it's dependencies are: MIT / Boost / Apache / BSD-2 Clause
  - it does have a slightly more tricky to use API and you must be a lot more strict (yaml-cpp let you get away with a lot of unsafe behaviour after going through this migration...) but it has significant documentation and examples.
- folder memory cards was the only other user of YAML, so I also had to migrate that

### Rationale behind Changes

- rapidyaml is a better library, I wanted to do this from the start a year ago, but getting it working with our visual studio configuration at the time was the main blocker.
  - from 300-400ms to 30-40ms loading the GameDB

### Suggested Testing Steps

- Thoroughly test folder memory cards and that i didn't break anything
- GameDB testing:
  - are speedhacks/gamefixes/memcard filters/patches loaded as expected?
  - if you have invalid yaml does the app catastrophically fail or gracefully continue (in Release mode! debug mode will hit assertions)


common/StringUtil.cpp
```diff
@@ -206,6 +207,22 @@ namespace StringUtil
 		return ss.str();
 	}
 
+	std::string toLower(const std::string& input)
```
CookiePLMonster, line 210: Use `std::string_view` for the input parameter.

common/StringUtil.cpp
```diff
@@ -206,6 +207,22 @@ namespace StringUtil
...
+	{
+		std::string newStr;
+		std::transform(input.begin(), input.end(), std::back_inserter(newStr),
+			[](unsigned char c) { return std::tolower(c); });
+		return newStr;
+	}
+
+	bool compareNoCase(const std::string& str1, const std::string& str2)
```
CookiePLMonster, line 218: Use `std::string_view` for input parameters.

common/StringUtil.cpp
```diff
@@ -206,6 +207,22 @@ namespace StringUtil
...
+	{
+		return std::equal(str1.begin(), str1.end(), str2.begin(),
+			[](char a, char b) {
+				return tolower(a) == tolower(b);
+			});
```
CookiePLMonster, line 223: This is almost certainly unsafe as access to `str2` is unbounded. You should return false before that line if string lengths are not equal.

pcsx2/GS/Renderers/HW/GSRendererHW.cpp
```diff
@@ -916,14 +916,14 @@ void GSRendererHW::SwSpriteRender()
 		for (int x = 0; x < w; x += 2)
 		{
 			u32* di = dpa.value(x);
-			ASSERT(*di + 1 == *dpa.value(x + 1)); // Destination pixel pair is adjacent in memory
+			ASSERT(di + 1 == dpa.value(x + 1)); // Destination pixel pair is adjacent in memory
```
CookiePLMonster, line 919: Unrelated changes?

pcsx2/GameDatabase.cpp
```diff
@@ -17,116 +17,158 @@
...
+					std::string validFix = fmt::format("{}Hack", wxString(EnumToString(id)).ToUTF8());
+					if (validFix == fix)
+					{
+						fixValidated = true;
+						break;
+					}
+				}
+				if (fixValidated)
+				{
+					gameEntry.gameFixes.push_back(fix);
```
CookiePLMonster, line 156:
```suggestion
					gameEntry.gameFixes.emplace_back(std::move(fix));
```

Nit

pcsx2/GameDatabase.cpp
```diff
@@ -147,33 +189,44 @@ GameDatabaseSchema::GameEntry YamlGameDatabaseImpl::entryFromYaml(const std::str
...
 
-		gameEntry.memcardFilters = node["memcardFilters"].as<std::vector<std::string>>(std::vector<std::string>());
+		// Memory Card Filters - Store as a vector to allow flexibility in the future
+		// - currently they are used as a '\n' delimited string in the app
+		if (node.has_child("memcardFilters") && node["memcardFilters"].has_children())
+		{
+			for (const ryml::NodeRef& n : node["memcardFilters"].children())
+			{
+				auto memcardFilter = std::string(n.val().str, n.val().len);
+				gameEntry.memcardFilters.push_back(memcardFilter);
```
CookiePLMonster, line 199:
```suggestion
				gameEntry.memcardFilters.emplace_back(std::move(memcardFilter));
```

Nit

pcsx2/GameDatabase.cpp
```diff
@@ -182,68 +235,129 @@ GameDatabaseSchema::GameEntry YamlGameDatabaseImpl::entryFromYaml(const std::str
...
+	return std::ifstream(path.c_str());
+#endif
 }
 
-bool YamlGameDatabaseImpl::initDatabase(std::ifstream& stream)
+void YamlGameDatabaseImpl::initDatabase()
 {
+	const ryml::Callbacks preserve_callbacks = ryml::get_callbacks();
+	const c4::error_callback_type preserve_c4_callback = c4::get_error_callback();
+	auto callbacks = ryml::get_callbacks();
```
CookiePLMonster, line 281: This is not actually set anywhere. The only `ryml::set_callbacks` is at the bottom, is it correct?

pcsx2/GameDatabase.cpp
```diff
@@ -182,68 +235,129 @@ GameDatabaseSchema::GameEntry YamlGameDatabaseImpl::entryFromYaml(const std::str
...
+		Console.WriteLn(fmt::format("[GameDB] Initialized in {}ms! {} games on record.", duration, gameDb.size()));
+	});
+
+	std::string serialLower = StringUtil::toLower(serial);
+	Console.WriteLn(fmt::format("[GameDB] Searching for '{}' in GameDB", serialLower));
+	if (gameDb.count(serialLower) == 1)
+	{
+		Console.WriteLn(fmt::format("[GameDB] Found '{}' in GameDB", serialLower));
+		return &gameDb[serialLower];
+	}
```
CookiePLMonster, line 359: If you refactor it to use `gameDb.find` and check the result instead, you'll get just one map lookup instead of two.

pcsx2/pcsx2.vcxproj
```diff
@@ -63,6 +63,7 @@
       <!-- SH 2/3 flashlight, explicitly set here don't change -->
       <!-- https://github.com/PCSX2/pcsx2/commit/16431653e4d92fda4069031897e24fbe4688d36a -->
       <FloatingPointModel>Precise</FloatingPointModel>
+      <RuntimeLibrary Condition="'$(Configuration)|$(Platform)'=='Release AVX2|x64'">MultiThreadedDLL</RuntimeLibrary>
```
CookiePLMonster, line 66: Unrelated change?

pcsx2/pcsx2.vcxproj.filters
```diff
@@ -1642,6 +1639,7 @@
     <ClCompile Include="gui\pxWindowTextWriter.cpp">
       <Filter>AppHost</Filter>
     </ClCompile>
+    <ClCompile Include="GS\Renderers\SW\GSSetupPrimCodeGenerator.all.cpp" />
```
CookiePLMonster, line 1642: Unrelated change?

pcsx2/pcsx2.vcxproj.filters
```diff
@@ -2755,6 +2747,7 @@
     <ClInclude Include="PAD\Gamepad.h">
       <Filter>System\Ps2\PAD</Filter>
     </ClInclude>
+    <ClInclude Include="GS.h" />
```
CookiePLMonster, line 2750: Unrelated change?

pcsx2/GS/Renderers/HW/GSRendererHW.cpp
xTVaser: yeah unrelated to the yaml stuff, but the game I test with crashes due to this bug so I had to fix it.

pcsx2/GameDatabase.cpp
xTVaser: The library has default callbacks set from my understanding, which is why i preserve and restore them after im finished here -- `auto callbacks = ryml::get_callbacks();` is a reference which i modify the error handler on the next line.

pcsx2/pcsx2.vcxproj
xTVaser: hmmm...for these VS project changes they do seem unrelated but i didn't do any of them myself.  Good catch

common/StringUtil.h
```diff
@@ -184,6 +184,9 @@ namespace StringUtil
 		return 0;
 	}
 
+	extern std::string toLower(const std::string& str);
+	extern bool compareNoCase(const std::string& str1, const std::string& str2);
```
Warepire, line 188: No need to mark these extern, the linker will figure it out anyway.

common/StringUtil.cpp
stenzek: Also redundant since `Strcasecmp` already exists, although a string_view overload could be nice.

Mrlinkwii (approved):

PR #5100, opened 2021-12-04
Linux and Windows CMake Builds: Change output executable to lowercase

F0bes:
### Description of Changes
bin/PCSX2 is now bin/pcsx2 -- update your scripts!

### Rationale behind Changes
Everything in Linux land is lowercase

### Suggested Testing Steps
Build PCSX2 and see if it is lowercase

-- I believe I got everything covered for this change? PCSX2-Linux.sh still works

### DIFF
--- a/pcsx2/CMakeLists.txt
+++ b/pcsx2/CMakeLists.txt
@@ -1483,6 +1483,12 @@ foreach(path IN LISTS RESOURCE_FILES)
 	pcsx2_resource(${path} ${CMAKE_SOURCE_DIR}/bin/resources/)
 endforeach()
 
+if (NOT APPLE)
+	set_target_properties(PCSX2 PROPERTIES
+		OUTPUT_NAME pcsx2
+	)
+endif()
+
 if (APPLE)
 	# MacOS defaults to having a maximum protection of the __DATA segment of rw (non-executable)
 	# We have a bunch of page-sized arrays in bss that we use for jit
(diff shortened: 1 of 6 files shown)


Mrlinkwii (approved):
Compiles here , works as intended :+1:

qurious-pixel:
I would expect it to fail to run from the desktop file.

`Exec=env GDK_BACKEND=x11 MESA_NO_ERROR=1 PCSX2`

Should the desktop file name, be lower-case?

lightningterror:
I prefer uppercase for PCSX2.

F0bes:
> I would expect it to fail to run from the desktop file.
> 
> `Exec=env GDK_BACKEND=x11 MESA_NO_ERROR=1 PCSX2`
> 
> Should the desktop file name, be lower-case?

You're right



> I prefer uppercase for PCSX2.

Why?

lightningterror:
> Why?

Seems more right since we use the upper case naming mostly everywhere.

refractionpcsx2:
Sure, in the app and website it's fine since it's an acronym, but the filename should be lower case, especially on linux.  Windows has always been lower case.

fantesykikachu:
> Windows has always been lower case

I just noticed that with the new cmake windows builds the executable filename is uppercase, maybe the PR should lowercase executables everywhere.

refractionpcsx2:
Maybe it does, but only with CMake, the VS projects still output a lower case file.

pcsx2/CMakeLists.txt
```diff
@@ -1525,6 +1525,12 @@ if(COMMAND target_precompile_headers)
 	target_precompile_headers(PCSX2_FLAGS INTERFACE PrecompiledHeader.h)
 endif()
 
+if (UNIX OR WIN32)
```
tellowkrinkle, line 1528:
```suggestion
if (NOT APPLE)
```
`UNIX` is true on macOS
I think the reason it's not actually renaming on macOS is because the bundle stuff doesn't look at OUTPUT_NAME, but I'd prefer if it was explicit

F0bes: Good catch.

F0bes:
Rebased.

PR #5112, opened 2021-12-07
Core: Stop using embedded resources

tellowkrinkle:
### Description of Changes
Replaces use of embedded resources with normal files in a resources directory somewhere (depends on OS)

### Rationale behind Changes
- Makes files easier to edit without recompilation
- Reduces dependencies on glib for non-Windows OSes

### Suggested Testing Steps
Check things that were moved into the `resources` directory, including...
- cheats_ws.zip (widescreen patches)
- gamedb
- sdl game controller database
- OpenGL and DX11 renderers

And make sure to check that the `PACKAGE_MODE` and `DISABLE_CHEATS_ZIP` still work


Mrlinkwii:
-  GameDB-  loaded  and applied automatic fixes  

- sdl game controller database - controller worked  

- OpenGL and DX11 renderers - since im on linux i could only check Opengl , which worked   :+1:

tellowkrinkle:
Stenzek reopened his qt prep PR so this should no longer be needed

PR #5115, opened 2021-12-08
Only pass C++ flags to C++ files.

arcum42:
### Description of Changes
Changed -fno-operator-names to only be passed to C++ files.

### Rationale behind Changes
3rd party libraries we compile no longer spew out a bunch of warnings that C files don't use that flag.

### Suggested Testing Steps
Compile pcsx2.


arcum42:
These automatic builds are so frigging broken...

PR #5130, opened 2021-12-10
CI - Remove 32bit Builds

xTVaser:
### Description of Changes

Removes 32bit builds from CI via the smallest possible change to do so.  There is still 32bit scaffolding and code throughout the repo, so until this is all removed, let's start small.

Also added a link to the new build page to the announcement / updated the scripts to no longer care about 32bit artifacts

### Rationale behind Changes

The time has come.

The next release is slated to be 64bit only, this is the first stepping stone in removing 32bit support.

### Suggested Testing Steps

Workflows work on my fork:
- https://github.com/xTVaser/pcsx2-rr/actions/runs/1561333647
- https://github.com/xTVaser/pcsx2-rr/actions/runs/1561333646

### DIFF
--- a/.github/workflows/linux-workflow.yml
+++ b/.github/workflows/linux-workflow.yml
@@ -45,19 +45,13 @@ jobs:
       matrix:
         include:
           - os: ubuntu-18.04
-            platform: x86
-            compiler: gcc
-            cmakeflags: -DLTO_PCSX2_CORE=ON
-            appimage: true
-            experimental: false
-          - os: ubuntu-18.04
             platform: x64
             compiler: gcc
             cmakeflags: -DLTO_PCSX2_CORE=ON
             appimage: true
             experimental: false
           - os: ubuntu-18.04
-            platform: x86
+            platform: x64
             compiler: clang
             # Need to disable PCH until cmake 3.17
             # (PCH conflicts with ccache, fixed by https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4400)
@@ -66,14 +60,14 @@ jobs:
             appimage: false
             experimental: false
           - os: ubuntu-18.04
-            platform: x86
+            platform: x64
             compiler: gcc
             cmakeflags: -DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON
             detail: " nopch"
             appimage: false
             experimental: false
           - os: ubuntu-18.04
-            platform: x86
+            platform: x64
             compiler: gcc
             cmakeflags: -DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON -DARCH_FLAG=-march=haswell
             detail: " avx2 nopch"
--- a/.github/workflows/linux-workflow.yml
+++ b/.github/workflows/linux-workflow.yml
@@ -74,7 +74,7 @@ jobs:
             appimage: false
             experimental: false
 
-    name: ${{ matrix.platform }} | ${{ matrix.compiler }}${{ matrix.detail }}
+    name: ${{ matrix.compiler }}${{ matrix.detail }}
     runs-on: ${{ matrix.os }}
     continue-on-error: ${{ matrix.experimental }}
     # Set some sort of timeout in the event of run-away builds.  We are limited on concurrent jobs so, get rid of them.
--- a/.github/workflows/macos-workflow.yml
+++ b/.github/workflows/macos-workflow.yml
@@ -70,7 +70,6 @@ jobs:
         platform: [x64]
         experimental: [false]
 
-    name: ${{ matrix.platform }}
     runs-on: ${{ matrix.os }}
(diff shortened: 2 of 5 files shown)


seta-san:
I’ll test the usb devices to make sure none of them are broken on 64 bit

lightningterror (approved):

Mrlinkwii (approved):

xTVaser:
@seta-san did you find anything?

.github/workflows/linux-workflow.yml
```diff
@@ -66,14 +60,14 @@ jobs:
...
-            platform: x86
+            platform: x64
             compiler: gcc
             cmakeflags: -DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON
             detail: " nopch"
             appimage: false
             experimental: false
           - os: ubuntu-18.04
-            platform: x86
+            platform: x64
```
Margen67, line 70: `platform` should be moved out of include so it doesn't need to be set for each job

xTVaser: unfortunately this is not how this works it seems, i tried it and it ends up not creating all the necessary builds.  Each include object seems to have to completely describe the run configuration -- and docs don't suggest otherwise https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

xTVaser:
blocked by the following - https://github.com/PCSX2/pcsx2/issues?q=is%3Aissue+is%3Aopen+label%3Ax64

Squall-Leonhart:
Theres an issue with the iso browser on 64bit builds when certain shell extensions are present (and only known to affect Windows 10, must be those pesky ASLR hardenings again)

the extensions in question tend to load a proxy executable, for example Alcohol 52/120% registers AxshlEx64.dll to the shell which calls AxshlExhlper.exe to determine if a ISO is mountable, etc.

In the presence of this shell extension, the browser becomes immediately non-responsive, this is bound to hit issues in other poorly made shell extensions though not be immediately obvious, so it may be prudent to build a loadlibrary hook and blocklist to mitigate the issue.

refractionpcsx2:
people still use Alcohol?

Squall-Leonhart:
Yes, but thats just one potential tool to do it, anything that runs in the shell to faciliate the loading of overlays on icons can do it too, for example google drive.

Heres a list of all known to be issues.
https://resource.dopus.com/t/shell-extensions-which-are-blocked-by-default/15350

refractionpcsx2:
Have you actually experienced this browser issue? of did you just see that list and go "ermagherd a prerblm"?

Squall-Leonhart:
Yes, thats why i knew to start poking at shell extensions because it matched that of another emulator displaying the same issue, the dialog hangs while trying to determine the correct overlay for files in the folder, in this case it doesn't matter if its trying to load the mds, mdf or iso overlay.

I would switch to daemon tools if it handled raw data/sub as well as alcohol did, it really is only second to CloneCD for PSX rips.

refractionpcsx2:
I'm curious why it doesn't affect 32bit then, we don't do anything different.

Squall-Leonhart:
My guess, is its the changes that microsoft introduced in 1607 that caused EZFRD64 to stop working, though in that rumble dlls case it really is just a improperly made library (64bit compiled dll that can only handle 32bit addresses? lol)

TheLastRar:
> Yes, but thats just one potential tool to do it, anything that runs in the shell to faciliate the loading of overlays on icons can do it too, for example google drive, if they happen to utilise an excutable handler.
> 
> Heres a list of all known to be issues. https://resource.dopus.com/t/shell-extensions-which-are-blocked-by-default/15350
> 
> I've certainly seen Acronis extensions cause the same, on Win 10, Winzip 17 and older too (same dialog hang) Others listed https://resource.dopus.com/t/crash-exit-or-high-cpu-when-viewing-certain-directories/1325

The links you provided also make reference to using tools such as ShellExView to block an shell extension system wide
If the official stance on bugged shell extensions is to not fix/workaround them (and why should we?) than that could allow a user to still have buggy software (i.e. Alcohol) installed, while not having crashes in PCSX2x64 (or other susceptible programs).

refractionpcsx2:
I agree, I don't think the onus should be on us to fix/workaround shitty extensions.

Squall-Leonhart:
The issue actually might be fixed already by dropping Wx for Qt, i'll poke a build and find out.

xTVaser:
@Squall-Leonhart Thanks for letting us know, but yeah i agree that it isn't necessarily our responsibility to support third-party customization tools.  

Regardless, I think this back and forth would be much better suited for an issue that mentions this PR -- rather than on the PR itself.  As after this PR merges, the context will be lost (if it ends up being relevant).

PR #5134, opened 2021-12-10
Rebase more de-wx-ifying from last month

stenzek:
Rebase of #4988. Only difference is the game database uses enums rather than strings now to store fixes/speedhacks, and the caching part was dropped, as apparently it's not wanted (keeping that one in my local branch).


JordanTheToaster (approved):
All looks good still from the last PR game fixes still work unicode spam works and it all seems functional to me.

Mrlinkwii (approved):
everything works from  my brief testing , everything seems to work , pad , renderers  , gamefixes etc :+1:

lightningterror (commented):
Moving fxaa.fx from gs_dx11 to gs_common would be more appropriate since it is shared between renderers.

lightningterror (changes requested):
Submodules need to be properly fetched/updated, libchdr was updated around a month ago and the pr reverts the changes.

lightningterror (changes requested):
I also see some new warnings which should be taken care of
https://github.com/PCSX2/pcsx2/runs/4483454025?check_suite_focus=true#step:9:406

pcsx2/GameDatabase.cpp
```diff
@@ -16,54 +16,93 @@
...
 #include "fmt/ranges.h"
 #include "yaml-cpp/yaml.h"
 #include <fstream>
 #include <algorithm>
+#include <mutex>
+#include <thread>
+
+static constexpr char GAMEDB_YAML_FILE_NAME[] = "GameIndex.yaml";
+static constexpr char GAMEDB_CACHE_FILE_NAME[] = "gamedb.cache";
+static constexpr u64 CACHE_FILE_MAGIC = UINT64_C(0x47414D4544423031); // GAMEDB01
```
TheLastRar, line 37: Leftover from the dropped Gamedb cache?

The commit `GameDatabase: Move out of gui and add cache` should also not mention a cache if none is added

TheLastRar: Additionally, I believe `Config: Add cache folder` can be dropped

stenzek: No, the cache is used for shaders and the game list in later commits.

TheLastRar: Ah, I only did a quick look though and didn't see anything using the cache folders yet, my mistake

stenzek:
> I also see some new warnings which should be taken care of https://github.com/PCSX2/pcsx2/runs/4483454025?check_suite_focus=true#step:9:406

These had no error checking before, but I'll add it now.

pcsx2/CDVD/CDVDaccess.cpp
```diff
@@ -376,33 +377,33 @@ bool DoCDVDopen()
...
+	std::string somepick(FileSystem::StripExtension(FileSystem::GetDisplayNameFromPath(m_SourceFilename[CurrentSourceType])));
 	//FWIW Disc serial availability doesn't seem reliable enough, sometimes it's there and sometime it's just null
 	//Shouldn't the serial be available all time? Potentially need to look into Elfreloadinfo() reliability
 	//TODO: Add extra fallback case for CRC.
-	if (somepick.IsEmpty() && !DiscSerial.IsEmpty())
-		somepick = L"Untitled-" + DiscSerial;
-	else if (somepick.IsEmpty())
-		somepick = L"Untitled";
+	if (somepick.empty() && !DiscSerial.IsEmpty())
+		somepick = StringUtil::StdStringFromFormat("Untitled-%s", static_cast<const char*>(DiscSerial.c_str()));
```
CookiePLMonster, line 385: What is it `static_cast`'ing from?

pcsx2/CDVD/ChdFileReader.cpp
```diff
@@ -16,71 +16,98 @@
...
+	while (CHDERR_REQUIRES_PARENT == (error = chd_open_wrapper(chds[chd_depth].c_str(), &fp, CHD_OPEN_READ, NULL, &child)))
 	{
 		if (chd_depth >= static_cast<int>(std::size(chds) - 1))
 		{
-			Console.Error(L"CDVD: chd_open hit recursion limit searching for parents");
+			Console.Error("CDVD: chd_open hit recursion limit searching for parents");
 			return false;
 		}
+
+		// TODO: This is still broken on Windows. Needs to be fixed in libchdr.
```
CookiePLMonster, line 81: I would suggest leaving a link to libchdr issue in the comment so it's easier to track. https://github.com/rtissera/libchdr/issues/55

pcsx2/CDVD/CsoFileReader.cpp
```diff
@@ -39,12 +40,12 @@ struct CsoHeader
 
 static const u32 CSO_READ_BUFFER_SIZE = 256 * 1024;
 
-bool CsoFileReader::CanHandle(const wxString& fileName)
+bool CsoFileReader::CanHandle(const std::string& fileName, const std::string& displayName)
 {
 	bool supported = false;
-	if (wxFileName::FileExists(fileName) && fileName.Lower().EndsWith(L".cso"))
+	if (StringUtil::EndsWith(displayName, ".cso"))
```
CookiePLMonster, line 46: Is this intended? Shouldn't this check `fileName`? Or maybe both are fine?

pcsx2/GameDatabase.cpp
```diff
@@ -16,54 +16,91 @@
...
+static bool compareStrNoCase(const std::string& str1, const std::string& str2)
 {
 	return std::equal(str1.begin(), str1.end(), str2.begin(),
-					  [](char a, char b) {
-						  return tolower(a) == tolower(b);
-					  });
+		[](char a, char b) {
+			return tolower(a) == tolower(b);
+		});
 }
```
CookiePLMonster, line 53: 1. Nitpick: `std::string_view` for arguments may work better.
2. Not nitpick: How does this work when strings are not equal length? `str2.end()` is not used so I'm not sure. Might be worth to safeguard this by comparing lengths first?

pcsx2/gui/Panels/BiosSelectorPanel.cpp
```diff
@@ -170,12 +170,13 @@ bool Panels::BiosSelectorPanel::ValidateEnumerationStatus()
 
 void Panels::BiosSelectorPanel::EnumThread::ExecuteTaskInThread()
 {
+	u32 region, version;
+	std::string description, zone;
```
CookiePLMonster, line 174: Were those part of a class and you made them locals? Hard to tell from the diff.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -153,90 +149,74 @@ NVMLayout* getNvmLayout()
...
-static void cdvdCreateNewNVM(const wxString& filename)
+static void cdvdCreateNewNVM(std::FILE* fp)
 {
-	wxFFile fp(filename, L"wb");
-	if (!fp.IsOpened())
-		throw Exception::CannotCreateStream(filename);
-
-	u8 zero[1024] = {0};
-	fp.Write(zero, sizeof(zero));
+	u8 zero[1024] = { 0 };
```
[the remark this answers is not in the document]
TheLastRar: My personal preference is `{0}` (no spaces, but explicit zero)
I see both `{}` and `{0}` in PCSX2 code, in roughly equal amounts, so I don't know what the preferred option is overall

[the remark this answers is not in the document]
CookiePLMonster: Technically `{}` is what you intend to convey:
* `{0}` is "first entry to 0, the rest is default initialized"
* `{}` is "all entries default initialized"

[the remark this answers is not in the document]
TheLastRar: Ah, okay, I'll keep that in mind

pcsx2/gui/Panels/BiosSelectorPanel.cpp
stenzek: They were globals before, but this is for scanning, so don't want to screw with global state.

pcsx2/GameDatabase.cpp
stenzek: Unused function from old code.

pcsx2/CDVD/CsoFileReader.cpp
stenzek: The aim was to support protocols/file systems where the display name is different from the file name, so the extension is checked based on the "visible" name, but the "internal" name is used to open it.

common/PathUtils.cpp
```diff
@@ -149,6 +149,21 @@ wxString Path::Combine(const wxString& srcPath, const wxDirName& srcFile)
...
+	const wxCharBuffer wxBuf(wxResult.ToUTF8());
+	return std::string(wxBuf.data(), wxBuf.length());
+}
+
+std::string Path::CombineStdString(const std::string_view& srcPath, const std::string_view& srcFile)
+{
+	const wxDirName srcPathDir(wxString::FromUTF8(srcPath.data(), srcPath.length()));
+	const wxString wxResult((srcPathDir + wxString::FromUTF8(srcFile.data(), srcFile.length())).GetFullPath());
+	const wxCharBuffer wxBuf(wxResult.ToUTF8());
+	return std::string(wxBuf.data(), wxBuf.length());
```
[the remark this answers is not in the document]
stenzek: Because the actual path joining is still using the wx class, for now (will be changed eventually).

PR #5135, opened 2021-12-10
GS-hw: Optimize pabe (per pixel alpha blending)

lightningterror:
### Description of Changes

Add an early condition to check if alpha is lower than 128, if so then no blending so early exit.
### Rationale behind Changes

Optimization, accuracy on hw blending.
### Suggested Testing Steps

Test the following games: SOTC, Breath of Fire Dragon Quarter, Strawberry Shortcake, Super Robot Wars, Cartoon Network Racing.


lightningterror:
Do not merge as I plan to include Blend Mix optimizations here.

PR #5140, opened 2021-12-12
Pad: Fix connection of macOS GameController framework controllers

tellowkrinkle:
### Description of Changes
Fixes an issue where SDL's backend for the macOS native game controller framework wouldn't initialize its controller list in time for PCSX2 to pick it up

### Rationale behind Changes
Prevents issues where PCSX2 doesn't recognize controllers that should otherwise be recognized
Happens most frequently around OS updates as they tend to break SDL's lower level APIs that could otherwise handle the controllers

### Suggested Testing Steps
Make sure pad still works on Linux and I didn't somehow break anything there

### DIFF
--- a/pcsx2/gui/AppInit.cpp
+++ b/pcsx2/gui/AppInit.cpp
@@ -43,6 +43,10 @@
 #endif
 #endif	// __WXGTK__
 
+#ifdef SDL_BUILD
+#include <SDL.h>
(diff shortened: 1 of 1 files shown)


Mrlinkwii (approved):
controllers work fine here on linux :slightly_smiling_face:

PR #5142, opened 2021-12-12
CDVD: Fix some read timing logic

refractionpcsx2:
### Description of Changes
Adjusts some of the reading logic to correct the timing of some events
Also don't set "READ" state when buffering, doing this breaks Pro Yakyuu Spirits 5 which does CD Standby and expects it to not be reading.

### Rationale behind Changes
In some scenarios it would spend too long reading/seeking when it shouldn't do since we'd already accounted for it.

### Suggested Testing Steps
Check games, especially CDVD sensitive ones, make sure they're still okay, not actually expecting much change.

SH2 is still janky, seems okayish for me with MTVU off... but I don't know what the proper fix is for that, just to jump in there first.

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -1038,7 +1038,7 @@ __fi void cdvdActionInterrupt()
 			cdvd.Spinning = true;
 			cdvd.Ready |= CDVD_DRIVE_READY; //check (rama)
 			cdvd.Sector = cdvd.SeekToSector;
-			cdvd.Status = CDVD_STATUS_READ;
+			cdvd.Status = CDVD_STATUS_PAUSE;
 			cdvd.nextSectorsBuffered = 0;
 			cdvd.triggerDataReady = true;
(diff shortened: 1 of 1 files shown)


gilderoylockhart01:
Strange, for me SH2 PAL is still not okay despite having MTVU and instantVU1 disabled. What an abomination this game is.

 will test other games as well.

refractionpcsx2:
> Strange, for me SH2 PAL is still not okay despite having MTVU and instantVU1 disabled. What an abomination this game is.
> 
> will test other games as well.

yeah SH2 is just cursed.

kenshen112:
> > Strange, for me SH2 PAL is still not okay despite having MTVU and instantVU1 disabled. What an abomination this game is.
> > will test other games as well.
> 
> yeah SH2 is just cursed.

Well, they don't call it a horror game for nothin xD

gilderoylockhart01:
Okay. 2 good news there. The patches for Indigo Prophecy can be removed.

And Sonic Mega Collection Plus doesn't seem to do the weird sound stall it can sometimes do when playing a mega drive game.

PR #5146, opened 2021-12-12
GSRenderer: Remove anti-blur hacks for altering display height

MrCK1:
### Description of Changes
Removes hacks that affect display height.

### Rationale behind Changes
Remove upscaling hacks from core code. Fixes "bouncing image" bug in The Suffering series when viewing security cameras.

### Suggested Testing Steps
Test games for bugs with native res and upscaling, specifically Time Crisis 2/3 (splitscreen), FFXII, Tenchu: Fatal Shadows, Worms 3D

Master:
pcsx2-dbg_2021_12_12_13_27_21_209

PR:
pcsx2-dev_2021_12_12_13_32_16_103

### DIFF
--- a/pcsx2/GS/Renderers/Common/GSRenderer.cpp
+++ b/pcsx2/GS/Renderers/Common/GSRenderer.cpp
@@ -205,34 +205,6 @@ bool GSRenderer::Merge(int field)
 		src_hw[i] = (GSVector4(r) + GSVector4(0, y_offset[i], 0, y_offset[i])) * scale / GSVector4(tex[i]->GetSize()).xyxy();
 
 		GSVector2 off(0);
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
That's not all anti-blur, some of that is split screen also and ofsetting the screen for games which do it (such as Devil May Cry), so as you can see from the top this is what your changes did

image

and you will have broken the time crisis split screen.

Additionally, I was going to suggest the anti-blur hacks (when we have just the right ones) get added as an option on GSDX, because a ton of people are gonna moan about this, so it at least gives them the option to have it on, if they want, they can turn it off for games like The Suffering

refractionpcsx2:
What you actually want to do, to get rid of the hacks are to change the "diff > 4" etc options to not be if statements. But if it's added as an option to the GUI you can do 

if(diff > 4 || !gs_no_blur)

PR #5151, opened 2021-12-13
Resource file cleanup

tellowkrinkle:
### Description of Changes
- Move locale files to resource directory
- Use `MACOSX_PACKAGE_LOCATION` to install locale files on macOS
- Use relative paths to reference data and docs directories in package mode
- Fixes missing translations in package mode

### Rationale behind Changes
- MacOS builds' locale files are tracked by make and properly regenerated if deleted
- PCSX2 no longer breaks if moved out of the original install directory as long as the relative location of data and docs is kept.  This removes the requirement for appimages to symlink the directories to a fixed location.

### Suggested Testing Steps
- Windows: Make sure the links in the firstrun wizard still work
- Linux: Make sure appimage works
- Linux: Make sure package mode works
- Linux: Make sure package mode install now has translations
- Linux: Make sure non-package-mode works


refractionpcsx2:
None of the languages appear in windows, it's fine on master, it's still looking in the old ./Langs location, as copying that from master brings the languages back

image

.github/workflows/scripts/linux/generate-cmake.sh
```diff
@@ -21,9 +21,8 @@ cmake                                       \
 -DCMAKE_BUILD_TYPE=Release                  \
 -DPACKAGE_MODE=TRUE                         \
 -DDISABLE_ADVANCE_SIMD=TRUE                 \
--DCMAKE_INSTALL_LIBDIR="/tmp/"              \
--DCMAKE_INSTALL_DATADIR="/tmp/"             \
--DCMAKE_INSTALL_DOCDIR="/tmp/PCSX2"         \
+-DDISABLE_PCSX2_WRAPPER=TRUE                \
+-DCMAKE_INSTALL_PREFIX="squashfs-root/usr/" \
```
qurious-pixel, line 25: AppImage cannot see "squashfs-root" it translates to `/`
The lookup will resolve to an unknown path.

tellowkrinkle:
> None of the languages appear in windows, it's fine on master, it's still looking in the old ./Langs location, as copying that from master brings the languages back

Try now

.github/workflows/scripts/linux/generate-cmake.sh
tellowkrinkle: This string is no longer embedded into the application, the application will get things like `../usr/share`

qurious-pixel: Relative paths is much cleaner.
Then `Configuration_Guide.pdf` & `PCSX2_FAQ.pdf` should go in ```../usr/share/PCSX2/``` ?

Mrlinkwii:
pcsx2 cant seem to find the languages files

image

its the  same on master 
image

Mrlinkwii (approved):

JordanTheToaster (approved):

tellowkrinkle:
Turns out the language file issue was wx finding the language file but failing to set the locale for other reasons

Not addressing that for now

PR #5164, opened 2021-12-15
Common: Fix GetWorkingDirectory on unix

### LINKED ISSUE #5163: [BUG]: Blockdumps do not work in Linux when built from source or using Appimage
### Describe the Bug

When built from source or using an Appimage, blockdumps do not work. Attempting to boot an iso with blockdumped ticked on either of these options, some form of the following window and console error message will appear.
image
`Path: /home/bestest/pcsx2/bin`
`(p) Unable to open the file for writing: General file/stream error [errno: 21](thread:EE Core)`
However, blockdumps do work on the PPA, being output in the user's home directory.

### Reproduction Steps

Tick the option to create blockdump, then boot an ISO.

### Expected Behavior

I expected PCSX2 to be able to create a blockdump.

### PCSX2 Revision

PCSX2 Nightly - v1.7.2133 Compiled on Dec 14 2021

### Operating System

### If Linux - Specify Distro

Linux Mint 20.2

tellowkrinkle:
### Description of Changes
Fixes an issue where `GetWorkingDirectory` would return a `std::string` containing null characters at the end
Fixes #5163

### Rationale behind Changes
When you concatenated those strings with other strings, the null characters would make most C utilities not notice the concatenation

### Suggested Testing Steps
Attempt to make a block dump in Linux

### DIFF
--- a/common/FileSystem.cpp
+++ b/common/FileSystem.cpp
@@ -1557,6 +1557,7 @@ std::string FileSystem::GetWorkingDirectory()
 		buffer.resize(buffer.size() * 2);
 	}
 
+	buffer.resize(std::strlen(buffer.c_str())); // Remove excess nulls
 	return buffer;
 }
 


HeroponRikiBestest:
I can confirm this fixes the issue.

common/FileSystem.cpp
```diff
@@ -1557,6 +1557,7 @@ std::string FileSystem::GetWorkingDirectory()
 		buffer.resize(buffer.size() * 2);
 	}
 
+	buffer.resize(strlen(buffer.c_str())); // Remove excess nulls
```
stenzek, line 1560: Nit: `std::strlen()` to follow the convention of the rest of the file.

Mrlinkwii (approved):

PR #5165, opened 2021-12-15
GS: Don't crash on failed init

### LINKED ISSUE #5156: [Feature Request]: Better GS exception handling
### Description

In cases where GS fails to init, PCSX2 unceremoniously crashes rather than throwing a proper exception to the user.  This can be seen in recent issue reports such as #5149.

### Reason

Proper exception handling in cases such as these should not only reduce confusion but also lead to better issue reporting.

### Examples

A popup exception with more useful and formatted information should suffice.

tellowkrinkle:
### Description of Changes
Exits emulation without crashing if GS fails to init

Fixes #5156

### Rationale behind Changes
Crashing on a random segfault doesn't make the cause of the problem very obvious

### Suggested Testing Steps
Do something that would make GS fail to init (e.g. delete resources) and make sure PCSX2 doesn't crash


stenzek:
FWIW, I have this implemented as a proper startup failure by returning false from `MTGS::WaitForOpen()` (no cross exception nonsense).

Shoegzer:
The message returned is certainly better:

pic

@tellowkrinkle for your example test step above you might consider informing testers to delete `resources/shaders` rather than just `resources`, as doing the latter will generate a non-GS exception that exists in master.

stenzek:
That's how I have it implemented. `Host::ReportErrorAsync` can be called from any thread and displays whatever error is desired on the UI thread with a message box (but doesn't block events). I'm using it for GPU feature checking at the moment.

tellowkrinkle:
> FWIW, I have this implemented as a proper startup failure by returning false from MTGS::WaitForOpen() (no cross thread exception nonsense).

Something like this?  Or do you think we should do something else when it fails?

stenzek:
Not really. That's still an exception mess.

I've already implemented this and was planning on PRing it eventually once I split the rest of the large qt commit out, but I don't really feel like dealing with more conflicts, so whatever, I'll just keep it in my local branch, forget it.

tellowkrinkle:
> Not really. That's still an exception mess, and will result in a double error popup on startup if the renderer posted its own message.

Sorry, I was going to deal with that part later
My main question is on what to do when `WaitForOpen()` fails.  Calling `Suspend()` isn't my favorite but it does seem to successfully stop the emulator before it hits anything that would make it crash.  The only other things I could think of without rewriting a large part of our threading code was throwing an exception (current threading code seems to like that a lot)
If you're planning on PRing a rewrite of the threading code that makes dealing with failure without exceptions easier, that would be good to know and I'm happy to delay this for later

Or if you meant "I have this all fixed in a change that I'll PR soon™ so just wait for that" that's fine too, I don't consider this to be an especially high priority issue and am happy to wait

stenzek:
> If you're planning on PRing a rewrite of the threading code that makes dealing with failure without exceptions easier, that would be good to know and I'm happy to delay this for later

That's exactly what I did in the qt branch. The failure happens on the GS thread, and gets propagated back here: 
https://github.com/stenzek/pcsx2/blob/qt-rebase/pcsx2/MTGS.cpp#L879-L914 and here https://github.com/stenzek/pcsx2/blob/qt-rebase/pcsx2/VMManager.cpp#L511

I also rewrote GS reopening, it doesn't really throw any exceptions anymore (except for D3D, still need to fix that up).

I didn't see much point in rewriting it for wx.

PR #5181, opened 2021-12-18
GS/OpenGL: Fix incorrect VAO cleanup

stenzek:
This wasn't an issue in wx, since the whole context/device gets torn down and recreated when you pause/unpause, or change settings.

But on Qt it is, since the context/device is preserved for a faster/lighter settings apply.


JordanTheToaster (approved):
All looks good on Windows just gotta wait for Linux testing now :)

Mrlinkwii (approved):

PR #5182, opened 2021-12-18
CDVD: Check file actually opened before proceeding

stenzek:
If you loaded an invalid file, particularly with compression, the file open may fail, leading to a division by zero (on block size), or trying to read from nothing.


gilderoylockhart01:
I guess I will write that here, but that PR doesn't seem to fix a very very weird problem I have with Wipeout 2097 where PCSX2 think it's a corrupted iso (it can't find the game executable file) despite comming straight from the game disc, and that same disc works fine in duckstation.

stenzek:
It's not going to. This just stops it crashing on corrupted files (at least in one place)

JordanTheToaster (approved):

PR #5197, opened 2021-12-20
GS-hw: Fix recent alpha max check.

lightningterror:
### Description of Changes

GS: Redo GetAlphaMinMax.
Cleaner this way, also fix recent regression on Full Blend with alpha max.

GS-hw: Move the blend mix alpha check after sw blending selection.
Optimization.
### Rationale behind Changes

Fix recent regression and optimization.
### Suggested Testing Steps

Test blending and make sure still works on both renderers.

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -2736,7 +2736,7 @@ void GSState::GetTextureMinMax(GSVector4i& r, const GIFRegTEX0& TEX0, const GIFR
 	r = vr;
 }
 
-void GSState::GetAlphaMinMax()
+void GSState::CalcAlphaMinMax()
 {
 	if (m_vt.m_alpha.valid)
 		return;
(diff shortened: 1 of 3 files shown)


tellowkrinkle:
Instead of this mess where we try to do all sorts of checks to only run `GetAlphaMinMax()` when needed, why not just have `GetAlphaMinMax()` return a reference to `m_vt.alpha` and then just do `GetAlphaMinMax().max`

Then we won't accidentally mess up conditions and we won't have to worry about that anymore at all

tellowkrinkle:
Side note, GetAlphaMinMax() already checks `m_vt.m_alpha.valid` for an early return.  But if you're worried about function call overhead, since compilers are terrible at splitting functions into a small inlinable part and a big rest of the function, you can do that manually and split `GetAlphaMinMax()` into
```cpp
void CalcAlphaMinMax();
GSVertexTrace::VertexAlpha& GetAlphaMinMax()
{
	if (!m_vt.m_alpha.valid)
		CalcAlphaMinMax();
	return m_vt.m_alpha;
}

PR #5138, opened 2021-12-12
GS: Replace separate HW renderers with single shared renderer

tellowkrinkle:
### Description of Changes
Replaces GSRendererOGL and GSRendererDX11 with one shared GSRendererNew and a new RenderHW virtual method on GSDevice

### Rationale behind Changes
- Less split code
- New renderers don't have to duplicate all 1500-ish lines of stuff that was in GSRendererOGL/GSRendererDX11.  The replacement method in GSDevice is usually about 250 lines and much more straightforward.

### Suggested Testing Steps
Please test DX11 and OGL renderers for regressions, I'm sure I accidentally broke stuff

### Future Changes
- Merge GSRendererNew and GSRendererHW (and make GSRendererHW final)

I want to delay this to give time for any existing forks of PCSX2 that have other renderers to migrate to the new interface and PR fixes for any issues that pop up in the process.  Any existing GSRendererHW subclasses should continue to work until this is carried out.

PR #5198, opened 2021-12-21
GS: Unify constant buffers between renderers

stenzek:
### Description of Changes

Trying to claw back some of the performance loss (~5% in extreme cases) from #5138. Copying all those structs around up to 20,000 times a second isn't free. Brief testing suggests a 10% performance uplift in draw call bound scenarios when running GSDumps.

### Rationale behind Changes

Better performance, less code.

### Suggested Testing Steps

Test a range of games to make sure no regressions were introduced. GS dumps would be fine too.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererNew.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererNew.cpp
@@ -1494,16 +1494,16 @@ void GSRendererNew::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 	const GSVector4& hacked_scissor = m_channel_shuffle ? GSVector4(0, 0, 1024, 1024) : m_context->scissor.in;
 	const GSVector4i scissor = GSVector4i(GSVector4(rtscale).xyxy() * hacked_scissor).rintersect(GSVector4i(rtsize).zwxy());
 
-	const GSVector4i commitRect = ComputeBoundingBox(rtscale, rtsize);
-	m_conf.scissor = (DATE && !DATE_GL45) ? scissor.rintersect(commitRect) : scissor;
+	m_conf.drawarea = scissor.rintersect(ComputeBoundingBox(rtscale, rtsize));
+	m_conf.scissor = (DATE && !DATE_GL45) ? m_conf.drawarea : scissor;
 
 	SetupIA(sx, sy);
 
 	if (rt)
-		rt->CommitRegion(GSVector2i(commitRect.z, commitRect.w));
+		rt->CommitRegion(GSVector2i(m_conf.drawarea.z, m_conf.drawarea.w));
 
 	if (ds)
-		ds->CommitRegion(GSVector2i(commitRect.z, commitRect.w));
+		ds->CommitRegion(GSVector2i(m_conf.drawarea.z, m_conf.drawarea.w));
 
 	m_conf.alpha_second_pass.enable = ate_second_pass;
 
--- a/pcsx2/GS/Renderers/HW/GSRendererNew.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererNew.cpp
@@ -1113,27 +1113,27 @@ void GSRendererNew::EmulateATST(float& AREF, GSHWDrawConfig::PSSelector& ps, boo
 	switch (atst)
 	{
 		case ATST_LESS:
-			AREF = (float)m_context->TEST.AREF - 0.1f;
+			AREF = static_cast<float>(m_context->TEST.AREF) - 0.1f;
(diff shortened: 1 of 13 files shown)


pcsx2/GS/Renderers/HW/GSRendererNew.cpp
```diff
@@ -1574,7 +1577,6 @@ void GSRendererNew::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 			return;
 
 		// RenderHW always renders first pass, replace first pass with second
-		memcpy(&m_conf.cb_ps,     &m_conf.alpha_second_pass.cb_ps,     sizeof(m_conf.cb_ps));
```
tellowkrinkle, line 1577: `m_conf.cb_ps.FogColor_AREF.a = m_conf.alpha_second_pass.ps_aref`

Though tbh I'm fairly sure we have our constants set up so opposite atst pairs will always use the same aref constant
So if you wanted you could probably get rid of `m_conf.alpha_second_pass.ps_aref` completely

pcsx2/GS/Renderers/HW/GSRendererNew.cpp
```diff
@@ -741,7 +738,7 @@ void GSRendererNew::EmulateBlending(bool& DATE_GL42, bool& DATE_GL45)
 
 		// Require the fix alpha vlaue
 		if (ALPHA.C == 2)
-			m_conf.cb_ps.alpha_fix = ALPHA.FIX;
+			m_conf.cb_ps.TA_MaxDepth_Af.a = ALPHA.FIX;
```
tellowkrinkle, line 741:
```suggestion
			m_conf.cb_ps.TA_MaxDepth_Af.a = ALPHA.FIX / 128.f;
```
Old was int, new is float

pcsx2/GS/Renderers/HW/GSRendererNew.cpp
```diff
@@ -165,7 +165,7 @@ void GSRendererNew::EmulateZbuffer()
 		}
 		else if (!m_context->ZBUF.ZMSK)
 		{
-			m_conf.cb_ps.max_depth = max_z;
+			m_conf.cb_ps.TA_MaxDepth_Af.z = max_z;
```
tellowkrinkle, line 168:
```suggestion
			m_conf.cb_ps.TA_MaxDepth_Af.z = max_z * 0x1p-32f;
```
Old was int, new is float

stenzek:
Reduced the copy area for DATE/HDR, this gives another 30% boost when GPU bottlenecked in games like Sly 2.

HighBunkerHxH:
> Reduced the copy area for DATE/HDR, this gives another 30% boost when GPU bottlenecked in games like Sly 2.

So games like GT3 & GT4 will benefit from this?

stenzek:
The goal was to restore performance to pre-5138 levels. It's now marginally faster than the old split renderers, but only in extreme cases, and probably not something really noticable.

lightningterror (commented):
One small warning to take care of
https://github.com/PCSX2/pcsx2/runs/4595662752?check_suite_focus=true#step:9:590

pcsx2/GS/Renderers/HW/GSRendererNew.cpp
```diff
@@ -1440,7 +1439,12 @@ void GSRendererNew::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 	}
 	else
 	{
-		EmulateATST(m_conf.cb_ps, m_conf.ps, false);
+		float aref;
```
lightningterror, line 1442: Maybe initializing it to 0 to silence the warning?

JordanTheToaster (approved):
Performance is back to what I'd have expected especially in games like Burnout 3. Haven't noticed any new issues and everything seems good on my end :)

refractionpcsx2:
Tried a bunch of GS Dumps all seemed pretty good, Sly 2 is about the same speed as master for me (and yes my miner was off :P )

stenzek:
It depends which path it was using. If it was using sw blending, it won't make a huge difference. But it you use dx11 or disable blending, it will at higher res

refractionpcsx2:
oh okay yeah I was mainly comparing OGL with basic blending enabled

pcsx2/GS/Renderers/HW/GSRendererNew.cpp
stenzek: Better yet; initialize it with the old value, possible reducing a cbuffer update :)

lightningterror (approved):
Dumps I tested were fine, looks good.

PR #5199, opened 2021-12-21
Merge the Qt branch: Part 1

stenzek:
### Description of Changes

This is the first part of a three-PR series which merges my work-in-progress Qt frontend. The new frontend is semi-usable already, the main missing feature is controller binding. If we can get it in-tree, others can work on it, and it avoids my rebase hell.

The WX interface should not be impacted from any of these changes. Except the second part, where it will use the new presentation system.

Currently, this is based on #5194 (and thus #5198).

The three parts:
1. Platform-independent PAD/SPU2 Config, Layered Config (this part)
2. New presentation system and on-screen display for GS.
3. VM manager (replaces all of the core thread junk, only used in Qt), and the Qt frontend itself.

What this part has:
 - Adds the layered settings interface for custom setting profiles/per-game settings. I don't see much point in implementing this in wx, but the code is there if someone wants to do it. Basically, `gamesettings/CRC.ini` is an overlay which gets applied on top of the config at runtime.
 - Adds the asynchronous error reporting mechanism discussed in another PR.
 - A platform independent PAD implementation based on the old Linux pad code. This is obviously lacking features compared to the current Windows plugin, these can be added later on. The PAD implementation should not handle bindings, those will be dispatched by common frontend code, that way you can use controller button combinations for hotkeys too.
 - SPU2 config backend which uses the new layered settings interface. The old split code will disappear eventually.

### Rationale behind Changes

The WX UI and the disgusting threading system is years past its overdue date. Everyone hates it.

### Suggested Testing Steps

Make sure wx doesn't break.


pcsx2/PAD/Host/Config.h
```diff
@@ -0,0 +1,131 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2020  PCSX2 Dev Team
```
RedDevilus, line 2: Minor nitpick: Some of these are 2002-2021 and others are 2002-2020.

Mewtwo5553:
stenzek so when done with three part will be per game settings because pcsx2 really needs it in order run properly different games needs different settings

pcsx2/Frontend/INISettingsInterface.cpp
```diff
@@ -0,0 +1,242 @@
...
+}
+
+void INISettingsInterface::Clear()
+{
+	m_ini.Reset();
+}
+
+bool INISettingsInterface::GetIntValue(const char* section, const char* key, int* value) const
+{
+	const char* str_value = m_ini.GetValue(section, key);
```
Warepire, line 74: Why is for example GetLongValue inappropriate here?

https://github.com/brofield/simpleini/blob/master/SimpleIni.h#L885

Guessing from the documentation, the default value is a problem because we cannot set a "value not found" value that is guaranteed to always work?

pcsx2/Frontend/INISettingsInterface.cpp
```diff
@@ -0,0 +1,242 @@
...
+std::vector<std::string> INISettingsInterface::GetStringList(const char* section, const char* key)
+{
+	std::list<CSimpleIniA::Entry> entries;
+	if (!m_ini.GetAllValues(section, key, entries))
+		return {};
+
+	std::vector<std::string> ret;
+	ret.reserve(entries.size());
+	std::transform(entries.begin(), entries.end(), std::back_inserter(ret),
+		[](const CSimpleIniA::Entry& it) { return std::string(it.pItem); });
```
Warepire, line 210: Any specific reason why something like below won't work? (just curious, no need to change it as in the end it probably doesn't matter)
```
for (const CSimpleIniA::Entry& it : entries)
     ret.push_back(str::string(it.pItem));
```

JordanTheToaster (approved):
Looks all good to me with my rounds of testing.

Mrlinkwii (approved):
works well

RedDevilus (approved):
No issues were detected.

MrCK1:
Everything looks all set for me, tested x64 exe too

PR #5215, opened 2021-12-27
FreeBSD: fix compilation

shiorid:
Following changes should fix FreeBSD build after [a083343](https://github.com/PCSX2/pcsx2/commit/a083343c6e918b890395ab463de1e0c1b030ddf2) and [9533fa2](https://github.com/PCSX2/pcsx2/commit/9533fa25c3f0c93a71d20c2fe5289ce94997bae2):

#### ryml:
Imported version of Rapid YAML doesn't build on Unices without `CLOCK_MONOTONIC_RAW`, which [should be Linux specific](https://linux.die.net/man/2/clock_gettime).
[450a27f](https://github.com/biojppm/c4core/commit/450a27f8b18ee23b18a3b102be90d542bc887b68) addresses this issue, so it is the current target.

#### FileSystem:
Fixes FreeBSD `ifdef` so it doesn't use `countof`. Instead, follow c++17 and use `std::count`, which is available in base clang compiler for all supported releases, and fits perfectly according to [`sysctl(3)`](https://www.freebsd.org/cgi/man.cgi?query=sysctl&apropos=0&sektion=3&manpath=FreeBSD+14.0-current&arch=default&format=html) manual page.

### Test environment:
FreeBSD 14.0-CURRENT main-c78058efb (fairly close to ^HEAD) with base clang 13 compiler. [Build log](https://reviews.freebsd.org/P532)

### DIFF
--- a/common/FileSystem.cpp
+++ b/common/FileSystem.cpp
@@ -1534,7 +1534,7 @@ std::string FileSystem::GetProgramPath()
 	int mib[4] = {CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, -1};
 	char buffer[PATH_MAX];
 	size_t cb = sizeof(buffer) - 1;
-	int res = sysctl(mib, countof(mib), buffer, &cb, nullptr, 0);
+	int res = sysctl(mib, std::size(mib), buffer, &cb, nullptr, 0);
 	if (res != 0)
 		return {};
 


refractionpcsx2:
Thanks for that :)

There might be a bit of a delay, actions is kinda screwed up right now..

shiorid:
Can _anything_ be done about submodules checkout fail?

refractionpcsx2:
> Can _anything_ be done about submodules checkout fail?

yes we can twiddle our thumbs and wait for Github to fix it.

tellowkrinkle:
Could you rebase instead of merging?

Also if there's no conflicts, you don't need to do anything

shiorid:
> Could you rebase instead of merging? 
> Also if there's no conflicts, you don't need to do anything

 Sorry, and thank you for the tip.
 I'm not used to _git forges_ workflow and wasn't meant to alter this PR.

 Should I rebase onto ^HEAD or leave as it is?

tellowkrinkle:
Yeah a rebase would be good

In the future, PRing from a branch that's not your `master` branch is a good way to avoid accidentally adding things to it after the fact

shiorid:
tellowkrinkle ***@***.***> writes:

> Yeah a rebase would be good
Done.

Once again, sorry for the trouble and thank you for the tips.

PR #5221, opened 2021-12-28
Merge the Qt branch: Part 2

stenzek:
### Description of Changes

This is the second part of a three-PR series which merges my work-in-progress Qt frontend. The new frontend is semi-usable already, the main missing feature is controller binding. If we can get it in-tree, others can work on it, and it avoids my rebase hell.

The three parts:

1. Platform-independent PAD/SPU2 Config, Layered Config.
2. New presentation system and on-screen display for GS  (this part).
3. VM manager (replaces all of the core thread junk, only used in Qt), and the Qt frontend itself.

What this part has:
 - Replaces all the GS window classes with a host display abstraction. The idea here is so that we can run the display without the emu core/GS going, which could be useful for a future UWP-based frontend.
 - Uses flip mode instead of blit mode for D3D. More efficient.
 - Fast software<->hardware transitions without tearing down and recreating the D3D/GL device.
 - New ImGui based OSD with improved statistics for the hardware renderers (e.g. readbacks/texture copies/texture uploads).
 - Pausing and resuming in wx doesn't tear down/recreate the device anymore (makes it much faster).

One quick note about imgui:
 - The imgui renderers aren't straight from upstream, because there's a number of issues in the GL renderer - mainly that it does a bunch of `glGet()` calls, which are serializing and slow on NVIDIA.

### Rationale behind Changes

The WX UI and the disgusting threading system is years past its overdue date. Everyone hates it.

### Suggested Testing Steps

Make sure there's no issues with graphics rendering on a range of drivers/platforms, no regressions in performance, and that pausing/resuming/changing options works as expected (or, as well as it does in wx).


stenzek:
Dropped the freetype removal commit for now since it makes reviewing too painful. Will add this separately afterwards.

Toxin60009:
Quick question will the Qt Ui have support for per game settings in the gui it self yes or no.

Toxin60009:
Quick question will the Qt Ui for pcsx2 have support for per game settings in the gui it self yes or no.

arcum42:
On line 1442 of CMakelists.txt, "HarfBuzz::HarfBuzz" needs to be removed.

arcum42:
In GS.cpp, looks like lines 1135 & 1136 are using enums that no longer exist.

arcum42:
Now compiles in Linux on my system.

JordanTheToaster:
Found an issue when closing graphics settings. "Option dither_ps2 doesn't have a default value"

stenzek:
> Found an issue when closing graphics settings. "Option dither_ps2 doesn't have a default value"

Should be fixed.

refractionpcsx2:
Couple of issues I've noticed

1. If you escape the emulation to close the GS window, then go in to the graphics settings, I think I changed the OSD settings, then Okayed it to come out, it opens a dead GS window, even though one shouldn't be open at this point. Although going to System-Resume does bring it back to life. This is apparently true in the QT Part 2 PR also.

2. Trying to go in to the GS settings while the emulation is running, the following happens:

Vulkan: prints "Writing 506645 bytes to 'G:\ps2\gitrepo\vulkan\cache\vulkan_pipelines.bin'" then crashes to desktop (on the other PR)
OpenGL: prints
"SetPixelFormat() failed: 0xC0070006
Failed to switch to surfaceless"
but does successfully go in to the options.

Mrlinkwii:
Tab(turbo) dosent seem to be working  , while f4 dose

stenzek:
> Tab(turbo) dosent seem to be working , while f4 dose

Should be fixed now.

Mrlinkwii:
> > Tab(turbo) dosent seem to be working , while f4 dose
> 
> Should be fixed now.

Works great :)

Mrlinkwii (approved):

JordanTheToaster (approved):
All looks good here.

gtgamer468:
Played around with a bunch of settings and found no issues tweaking them. Should be all good by now.

dio-gh:
Mildly related, but would it be viable to make the settings windows in WX non-pausing while on the way to Qt? I cannot know how much work it'd require, so it may not worth it, but it'd definitely be a welcome change to stick into 1.8.

> the settings windows in WX non-pausing

By that I mean, for the following settings panels not to pause the VM:
- memory cards,
- audio settings,
- gamepad settings,
- network and hdd settings,
- usb settings.

The rationale being, apart from these pauses being kind of lame and annoying, is that during online play this can prevent people from rebinding their controls, since pausing the VM while playing online is lowkey dicey.

stenzek:
> Mildly related, but would it be viable to make the settings windows in WX non-pausing while on the way to Qt? I cannot know how much work it'd require, so it may not worth it, but it'd definitely be a welcome change to stick into 1.8.

This is the case for graphics settings. I'm not touching anything else in wx, it's a complete clusterf**k where different things happen on different threads, with the risk of deadlocking everywhere.

MrCK1:
Everything looks good to me, no issues on my end 👍

refractionpcsx2 (approved):

Zeromegax:
So over time you will be able to use the mouse in this emulator.

refractionpcsx2:
Please don't pick random pull requests to ask questions. if you want support please use our Discord or Forums.

PR #5228, opened 2021-12-29
GS:OGL: Populate feature struct after checking support

tellowkrinkle:
### Description of Changes
Fixes an issue introduced in renderer merge where OGL would populate features with uninitialized GLLoader values

### Rationale behind Changes
Unbreaks things

### Suggested Testing Steps
Make sure games that use accurate destination alpha aren't slow the first time you run them (with subsequent runs being fine)


stenzek:
This was covered in the Vulkan merge (I just put it after the shader cache, not before, hence no conflict)

PR #5231, opened 2021-12-30
GS: Use flat shading for IIP=0 instead of geometry shaders

stenzek:
### Description of Changes

Because D3D defines flat shaded varyings to be from the first vertex, whereas OpenGL/Vulkan and GS use the last vertex, we previously used a geometry shader to apply the colour across all vertices.

To work around the provoking vertex difference, can simply swap the indices, as back-face culling is disabled anyway.

This PR also adds another change from the vulkan branch - using native hardware point/line expansion instead of geometry shaders where supported. Related change so I figured I'd toss it in to get tested at the same time.

### Rationale behind Changes

Geometry shaders are a pipeline bubble on modern GPUs, and it's silly to use them when there's native flat shading support. 

### Suggested Testing Steps

Make sure games aren't broken graphically. Test upscaling too, since the point/line expansion is used instead of GS now.


stenzek:
The only case I've found so far where this matters is Tony Hawk's American Wasteland (for the shadows). But I'm sure there's others.

pcsx2/GS/Renderers/OpenGL/GLState.cpp
```diff
@@ -22,6 +22,9 @@ namespace GLState
 	GSVector2i viewport;
 	GSVector4i scissor;
 
+	bool point_size = false;
```
refractionpcsx2, line 25: any chance we can give this a bit better name? "point_size = yes" sounds weird :P{

stenzek:
Merged as part of #5221.

PR #5232, opened 2021-12-30
Fix some gcc warnings.

arcum42:
### Description of Changes
Got rid of some warnings in gcc over in the usb code. I decided clang warnings could wait for another time.

### Rationale behind Changes
Less warnings while compiling.

### Suggested Testing Steps
Compile pcsx2.

### DIFF
--- a/pcsx2/USB/usb-pad/usb-pad.h
+++ b/pcsx2/USB/usb-pad/usb-pad.h
@@ -1516,7 +1516,7 @@ namespace usb_pad
 	// 0x0982 - PlayStation 2
 	// 0x0984 - ???
 
-	static uint8_t gametrak_dev_descriptor[] = {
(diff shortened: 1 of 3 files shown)


arcum42:
No idea why rapidyaml wants to be part of this commit, as I didn't touch it.

tellowkrinkle:
> No idea why rapidyaml wants to be part of this commit, as I didn't touch it.

Just Git submodules being great as always
git doesn't update submodules when you check out branches which makes it super easy to accidentally commit that

lightningterror (commented):
Commits can be better named.

arcum42:
Always planned on squashing the second commit with the first, which is why I wasn't too concerned with its name.

Mrlinkwii (approved):

PR #5234, opened 2021-12-30
GameDB: remove patches for  'Neo Contra' and add EE rounding and clamping

Mrlinkwii:
### Description of Changes

 remove patches for  'Neo Contra' and adds EE rounding and claming
### Rationale behind Changes

The patches  arent needed any more and the issue can be fixed with EE clamping and rounding 
### Suggested Testing Steps

 look at CI / make sure the game works as intended


gilderoylockhart01:
I guess the issue linked to it can also be closed.

Also, it might be worth checking Kings Field 4, I guess the patch can be replaced with ee rounding to negative?

Issue #354, opened 2014-11-19
Ratchet & Clank 2 - Megaturret Doesn't Work Correctly


erwark:
Core issue?

Issue title: Ratchet & Clank 2 (PAL/Australia) Megaturret Not Working

-PCSX2 version: 1.2.1 Stable

-PCSX2 options: Defaults, no settings modifications affect the issue

-Plugins used: GSdx (SW or HW), SPU2-X, Lilypad, Linuz or Gigaherz CDVD,rest null plugins. Same issue under ZeroGS - ruling out Graphics.

-Plugin settings: Defaults, no settings modifications affect the issue

-Description: The turrets from the Megaturret Glove (the upgraded Miniturrent Glove) just bounce up and down with their starting sound and animation looping. They don't work (shoot anything).

-How to replicate: Upgrade the Miniturret Glove to the Megaturret glove and use it.

-Last known version to work: Never worked (checked back to 0.9.7). Also checked most recent testing build.

-PC specifications: 
CPU: Intel Core i5 2500K @ 4.3Ghz 
GPU: Gigabyte Radeon 6950 1GB
OS: Windows 7 x64 SP1

-Other comments: Aside from this issue, the game runs fine under Software Mode. No settings/gamefixes seem to make a difference. 2 other people have posted the issue in the past with no resolution (but they stopped following up the issue pretty quickly). These people seemed to have 64-bit Windows and PAL-Region games, like me. I have not tried the NTSC release.

I've filled out the required info as I understand it and welcome any ideas or requests for more information.

Thanks.

Hirato:
This did happen on the consoles as well, but only occasionally.
The R&C wiki also suggests the issue plagues the NTSC version.

Something in PCSX2's operation seems to trigger the issue for every single turret, so they all end up stuck in an initialising state.
I've tried all the semi-relevant manual game hacks, as well as numerous combinations of the EE/VU precision options to no avail.
Is there anything I can do to help debug this?

Nobbs66:
I'll look into this sometime tonight.

FlatOutPS2:
There's also bug report on the forum about this:
http://forums.pcsx2.net/Thread-Bug-Report-Ratchet-and-Clank-2-Megaturrets-PAL-Australia

atomic83GitHub:
Still here with last 1.5 dev build, blockdump is comming.

atomic83GitHub:
Here is a blockdump of the issue:
[...]

PR #5235, opened 2021-12-30
Hack for Megaturrets on Ratchet & Clank 2 (#354)

Vogtinator:
Values found by @kozarovv

### Description of Changes
This adds a new gamefix/hack to fix deployment of Megaturrets in RC2.

I have absolutely no idea what the patched function does, but comparing against 1.5 instead of 6.5 fixes it...
Maybe it's somehow involved in collision detection?

### Rationale behind Changes
See #354

### Suggested Testing Steps
Throwing Megaturrets onto various surfaces on Ratchet & Clank 2 PAL edition.

### DIFF
--- a/.github/workflows/scripts/validation/lint-gamedb/lint-gamedb.py
+++ b/.github/workflows/scripts/validation/lint-gamedb/lint-gamedb.py
@@ -35,6 +35,7 @@ allowed_game_fixes = [
     "VUKickstartHack",
     "IbitHack",
     "VUOverflowHack",
+    "RC2MegaturretHack",
 ]
 allowed_speed_hacks = ["mvuFlagSpeedHack", "InstantVU1SpeedHack"]
 # Patches are allowed to have a 'default' key or a crc-32 key, followed by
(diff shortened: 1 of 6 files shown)


refractionpcsx2:
As I said in the other thread, I don't really want to hard code this, I'd rather have a proper solution that we can link to the GameDB

lightningterror:
Not a fan of it as well, especially now since we are moving away from hacks.

Vogtinator:
I'm not a fan of it either, but I know basically nothing about internal workings of RC2 and the PS2 to debug the root cause nor develop a more appropriate patching mechanism.

This PR is mostly just a formality to have at least something...

GreyXor:
Thanks for this PR, I confirm that the 'Megaturret Glove' works as it should.

Tested on [Ratchet & Clank 2 [SCES-51607]](http://redump.org/disc/234/)

gilderoylockhart01:
Erfff, I am skeptical about this, I guess I would prefer having it broken and waiting for a true hardware test to actually confirm what is going on here.

lightningterror:
I will close this as we aren't adding it.

PR #5243, opened 2022-01-01
FileSystem: Use _stat64 routines

xantares:
allows compilation with gcc/mingw

### DIFF
--- a/common/GL/ContextWGL.cpp
+++ b/common/GL/ContextWGL.cpp
@@ -20,12 +20,12 @@
 
 static void* GetProcAddressCallback(const char* name)
 {
-	void* addr = wglGetProcAddress(name);
+	void* addr = reinterpret_cast<void*>(wglGetProcAddress(name));
 	if (addr)
 		return addr;
 
 	// try opengl32.dll
-	return ::GetProcAddress(GetModuleHandleA("opengl32.dll"), name);
+	return reinterpret_cast<void*>(::GetProcAddress(GetModuleHandleA("opengl32.dll"), name));
 }
 
 namespace GL
@@ -364,7 +364,7 @@ namespace GL
 			}
 
 			// re-init glad-wgl
-			if (!gladLoadWGLLoader([](const char* name) -> void* { return wglGetProcAddress(name); }, m_dc))
+			if (!gladLoadWGLLoader([](const char* name) -> void* { return reinterpret_cast<void*>(wglGetProcAddress(name)); }, m_dc))
 			{
 				Console.Error("Loading GLAD WGL functions failed");
 				return false;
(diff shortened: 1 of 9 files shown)


common/GL/ContextWGL.cpp
```diff
@@ -241,7 +241,7 @@ namespace GL
 			}
 
 			// re-init glad-wgl
-			if (!gladLoadWGLLoader([](const char* name) -> void* { return wglGetProcAddress(name); }, m_dc))
+			if (!gladLoadWGLLoader([](const char* name) -> void* { return (void*)wglGetProcAddress(name); }, m_dc))
```
stenzek, line 244: `if (!gladLoadWGLLoader([](const char* name) -> { return reinterpret_cast<void*>(wglGetProcAddress(name)); }, m_dc))`

common/GL/ContextWGL.cpp
```diff
@@ -319,7 +319,7 @@ namespace GL
 			}
 
 			// re-init glad-wgl
-			if (make_current && !gladLoadWGLLoader([](const char* name) -> void* { return wglGetProcAddress(name); }, m_dc))
+			if (make_current && !gladLoadWGLLoader([](const char* name) -> void* { return (void*)wglGetProcAddress(name); }, m_dc))
```
stenzek, line 322: `if (make_current && !gladLoadWGLLoader([](const char* name) -> { return reinterpret_cast<void*>(wglGetProcAddress(name)); }, m_dc))`

common/Windows/WinHostSys.cpp
```diff
@@ -47,6 +48,16 @@ long __stdcall SysPageFaultExceptionFilter(EXCEPTION_POINTERS* eps)
 	{
 		return EXCEPTION_CONTINUE_SEARCH;
 	}
+#else
+  try
+  {
+    return DoSysPageFaultExceptionFilter(eps);
+  }
+  catch (...)
```
stenzek, line 56: This isn't going to work, unless your compiler transforms SEH exceptions into C++ exceptions. As far as I'm aware, MSVC is the only one which does this, but I could be wrong (although it's not really something you want, and it's disabled by default).

Fortunately, there's other options, like using vectored exception handlers, which we use on 64-bit anyway.

xantares: afaik gcc handles c++ exception as seh or sjlj

lightningterror:
@xantares Status?

xantares:
@lightningterror this is still needed

lightningterror (commented):
It was discussed that the try/catch bits are undesired and we also use VEH anyway.
Should also be properly formatted.

lightningterror:
Closing for the reasons stated before.

xantares:
@lightningterror what if I drop that commit, the rest should be good

stenzek:
I don't know why we would want to build with mingw anyway. There's plenty of issues with headers being out of date, subtle ABI differences, and it provides no advantage.

I have a branch with clang-cl support, which if you want to use non-MSVC codegen, works much better.

PR #5248, opened 2022-01-01
VTLB: Fix ppmap allocation in 64-bit

tellowkrinkle:
### Description of Changes
Makes ppmap allocation accessible from 64-bit JIT

### Rationale behind Changes
Fixes goemon on 64-bit

### Suggested Testing Steps
Test goemon


HeroponRikiBestest:
Successfully fixed the issue.
https://cdn.discordapp.com/attachments/923492653788692480/927010757638508644/unknown.png

PR #5249, opened 2022-01-01
GS:SW: Fix crash when scanline count wasn't divisible by thread count

tellowkrinkle:
### Description of Changes
Fixes crash when scanline count wasn't divisible by thread count (common with odd thread counts like 5)

### Rationale behind Changes
Less crashy more runny

### Suggested Testing Steps
Run SW renderer with weird thread counts

### DIFF
--- a/pcsx2/GS/GSPerfMon.h
+++ b/pcsx2/GS/GSPerfMon.h
@@ -22,9 +22,8 @@ public:
 	{
 		Main,
 		Sync,
-		WorkerDraw0, WorkerDraw1, WorkerDraw2, WorkerDraw3, WorkerDraw4, WorkerDraw5, WorkerDraw6, WorkerDraw7,
-		WorkerDraw8, WorkerDraw9, WorkerDraw10, WorkerDraw11, WorkerDraw12, WorkerDraw13, WorkerDraw14, WorkerDraw15,
-		TimerLast,
+		WorkerDraw0,
+		TimerLast = WorkerDraw0 + 32, // Enough space for 32 GS worker threads
 	};
 
 	enum counter_t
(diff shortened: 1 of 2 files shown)


JordanTheToaster:
Seems to run fine now upto around 25 threads afterwards it will start to crash again.

HeroponRikiBestest:
This works properly for me. As Jordan said, it now only begins crashing with 23 extra rendering threads or more. A better error message that doesn't require you to close the emulator or a hard cap on the # of threads at 22 might be helpful.

tellowkrinkle:
[Okay we currently only allocate enough space in GSPerfMon for 16 threads, anything past that starts corrupting memory](https://github.com/PCSX2/pcsx2/blob/0ddef1169d755c5c522a501a887056d90a959e1f/pcsx2/GS/GSPerfMon.h#L25-L26)

Should we limit threads to 16 or add space for 16 more threads in PerfMon?

F0bes:
I wouldn't mind having a 32 thread cap.

Mrlinkwii:
yeah 32 thread cap  sounds good  the way most CPUs are going its a good future proof number

JordanTheToaster (approved):
25+ threads no longer crashes and games are fine.

HeroponRikiBestest:
This PR causes a regression in Neo Atlas III, causing the game to crash during the dialog boxes again. Weirdly, this occurs on OpenGL as well, and on SW with 16 or less threads too. I double checked that this does not happen on master.

tellowkrinkle:
Wait never mind it's because this branch is based on PCSX2 from before that was fixed
Try now

JordanTheToaster:
Tested again and all still seems fine to me.

HeroponRikiBestest:
Tested again, the issues on Shangai and Neo Atlas are resolved.

PR #5252, opened 2022-01-02
R5900: Use unsigned math for add/sub instructions

stenzek:
### Description of Changes

For the interpreter and dynarec constant prop, this was causing different code to be generates at -O0 and -O3 in clang. As best I can tell, it's due to there being a possibility of signed overflow, thus the high bits are irrelevant. Except for us, when we're simulating a CPU.

"Bad" code was only generated in clang, msvc and gcc were fine.

Case which prompted this was Juiced 2, but there may be others.

### Rationale behind Changes

Getting rid of undefined behaviour and fixing game crashes, yay.

### Suggested Testing Steps

Quick testing in interpreter and dynarec to make sure nothing was relying on this broken behaviour. Preferably with clang.

### DIFF
--- a/pcsx2/R5900OpcodeImpl.cpp
+++ b/pcsx2/R5900OpcodeImpl.cpp
@@ -300,7 +300,7 @@ void ADDI()
 void ADDIU()
 {
 	if (!_Rt_) return;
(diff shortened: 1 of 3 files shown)


F0bes (approved):
Ran a few games with this compiled with Clang. No issues noticed.

PR #5269, opened 2022-01-06
SysThreadBase: Fix double init of SPU2 etc

### LINKED ISSUE #5267: [BUG]: Hotkey doesn't work after switching HW renderer
### Describe the Bug

When you switch the game while running a game, and change the renderer to another HW one in Graphics Settings, any hotkey does not work anymore.
If you shut down the game and load another one, this bug doesn't seem to be occurred.

This bug does not happen in v1.7.2186.
It started to happen since v1.7.2187 ( #5221 ).

### Reproduction Steps

- Set D3D11 or OGL renderer in Graphics Settings.
- Run a game.
- while running the game, load another game and reset the system.
- Switch the renderer to another HW one in Graphics Settings.

### Expected Behavior

Hotkey should work well.

### PCSX2 Revision

v1.7.2197 - it's still existed in this version

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

stenzek:
### Description of Changes

Typical pcsx2 threading nonsense.

Changing discs sets g_CDVDReset, which ends up double-opening PAD. So,
 1) it never gets closed properly, and
 2) when you change renderers, it should get reinitalized due to the window change (because it's nasty and hijacks the window thread), but because it has a gross "openCount" which never reaches zero, that never happens
 3) hotkeys don't work

### Rationale behind Changes
:rage:

Fixes #5267

### Suggested Testing Steps
Cry in a corner

### DIFF
--- a/pcsx2/System/SysThreadBase.cpp
+++ b/pcsx2/System/SysThreadBase.cpp
@@ -333,15 +333,11 @@ bool SysThreadBase::StateCheckInThread()
 			if (m_ExecMode != ExecMode_Closing)
 			{
 #ifndef PCSX2_CORE
-				if (g_CDVDReset)
-					// AppCoreThread deals with Reseting CDVD
-					// Reinit all but GS, USB, DEV9, CDVD (just like with isSuspend = false previously)
-					OnResumeInThread(static_cast<SystemsMask>(~(System_GS|System_USB|System_DEV9|System_CDVD)));
-				else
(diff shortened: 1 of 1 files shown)


Tokman5:
It seems to be fixed.

JordanTheToaster (approved):

Mrlinkwii (approved):

PR #5277, opened 2022-01-08
GS-hw: Don't enable blending when only alpha is written.

lightningterror:
### Description of Changes

Don't enable blending when only alpha is written.
### Rationale behind Changes

Optimization.
### Suggested Testing Steps

Check if games didn't break.


lightningterror:
I'll throw this in here for the q1 report, dbz bt2 gains around 4 fps on opengl.

PR #5283, opened 2022-01-08
CDVD: Adjust error timing and modify error used for high sector.

### LINKED ISSUE #5280: [BUG - Regression]: Silent Hill 2 black ribbon demo: Invalid sector error causing softlock.
### Describe the Bug

In the video demo "Silent Hill 2 black ribbon", there are 2 trailers playing continuously.

Before the problematic commit, the demo did a bunch of out of bound sector errors when reading the second trailer but did continue regardless.

On master, the software simply dies when doing so with the emulator constantly spamming invalid sector read.

### Reproduction Steps

Load the blockdump + savestate.

### Expected Behavior

The video demo should not crash when reading the second trailer.

### PCSX2 Revision

pcsx2-v1.7.2210-windows-x86

### Operating System

Windows 11

### If Linux - Specify Distro

_No response_

### CPU

I9 9900K

### GPU

RTX 3070

### GS Settings

_No response_

### Emulation Settings

_No response_

### GS Window Screenshots

The console log:
Capture d’écran 2022-01-08 110826

### Logs & Dumps

Blockdump + savestate: https://drive.google.com/file/d/1LsXSmonY7hYjQCrBDytCD6KYGPWAs48K/view?usp=sharing

refractionpcsx2:
### Description of Changes
Add timing to error calls and modify error for reading past end of the disc

### Rationale behind Changes
Games do terrible, terrible things. I don't know if this is right, but it seems to mimic PS2 behaviour on the Silent Hill 2 Black Ribbon demo disc.

### Suggested Testing Steps
test stuff that tries to read past the end of the disc

Fixes #5280

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -1094,6 +1094,11 @@ __fi void cdvdActionInterrupt()
 			cdvd.Sector = 0;
 			cdvdUpdateStatus(CDVD_STATUS_STOP);
 			break;
+
+		case cdvdAction_Error:
+			cdvd.Ready = CDVD_DRIVE_READY | CDVD_DRIVE_DEV9CON | CDVD_DRIVE_ERROR;
+			cdvdUpdateStatus(CDVD_STATUS_PAUSE);
+			break;
 	}
 	cdvd.Action = cdvdAction_None;
 
@@ -1151,7 +1156,7 @@ __fi void cdvdReadInterrupt()
 		}
 	}
 
-	if (cdvd.Sector > cdvd.MaxSector)
+	if (cdvd.Sector >= cdvd.MaxSector)
 	{
 		DevCon.Warning("Read past end of disc Sector %d Max Sector %d", cdvd.Sector, cdvd.MaxSector);
 		cdvd.Error = 0x32; // Outermost track reached during playback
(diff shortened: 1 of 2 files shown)


combo-pizza:
Nice, it gets past the warthog logo in mace griffin (U) now

gilderoylockhart01:
This PR fixes Ratchet and Clank UYA PAL 1.00 crash at the 5th vid comic when dying (also an issue with OPL btw).

As a general rule of thumb it is a good idea to check for games that are broken on OPL.

As for the log, it is not consistent depending which game:

This comes from ratchet and silent hill:
Capture d’écran 2022-01-08 174332

And this comes from the Simpson's game:

Capture d’écran 2022-01-08 174313

refractionpcsx2:
That's hitting a different thing

did it break Ratchet, SH or Simpsons?

gilderoylockhart01:
Ratchet and SH were broken before this PR, now they are booth fine. Simpson's remains the same (it is fine ^^).

refractionpcsx2:
cool, that's good to know :)

gilderoylockhart01:
Btw yeah, cdvd errors are like pokemons, you gota catch them all ^^.

refractionpcsx2:
yup! we've been slack until now.  I'm not sure this PR is correct, but doing what I think is correct doesn't work for SH2 Black Ribbon.

PR #5285, opened 2022-01-08
GS: fully support target rescaling in TC.

iMineLink:
### Description of Changes
Added full support for target rescaling in GS TC.
Previously only color to depth or depth to color rescaling was supported.
Also avoid redundant clears when the target is going to be completely drawn when rescaling.

EDIT: Bonus fix palette rendering in Majokko A-La-Mode 2, replacing CU+OO hack for nativeres palette rendering and downloading with direct palette rendering in gs local memory (point list draw).

### Rationale behind Changes
Improve TC behaviour when handling targets with different scaling, for example native res and upscaled, or potentially when changing internal resolution without flushing.
Avoid redundant target clears.

### Suggested Testing Steps
Test some games and check that no visual regression occurs, especially games with "CU" hacks that have native res draws even when upscaled (currently ~~"Majokko A La Mode 2" and~~ only "Tales Of Abyss").
Potentially appreciate performance improvement in depth to color and color to depth conversion due to clearing skip.

EDIT: Check that Burnout games do not have regressions due to improvements in point list palette "hack".

### DIFF
--- a/pcsx2/GS/GSCrc.cpp
+++ b/pcsx2/GS/GSCrc.cpp
@@ -203,7 +203,7 @@ const CRC::Game CRC::m_games[] =
 	{0xA3643EB1, GiTS, KO, 0},
 	{0x28557423, GiTS, RU, 0},
 	{0xBF6F101F, GiTS, EU, 0}, // same CRC as another US disc
-	{0xF442260C, MajokkoALaMode2, JP, 0},
+	{0xF442260C, MajokkoALaMode2, JP, PointListPalette},
 	{0xA616A6C2, TalesOfAbyss, US, 0},
 	{0x14FE77F7, TalesOfAbyss, US, 0},
 	{0xAA5EC3A3, TalesOfAbyss, JP, 0},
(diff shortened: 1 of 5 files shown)


refractionpcsx2:
I tried Tales of the Abyss and I don't know what I'm looking for so somebody else will need to try that.

I also tried Majokko A-La-Mode 2 and to be honest, it's pretty broken in HW, even with the CRC, but this PR doesn't seem to help

Here's a GS Dump for you to tr
[Majokko.zip](https://github.com/PCSX2/pcsx2/files/7876957/Majokko.zip)
y

iMineLink:
The first commit shouldn't give any benefit to Majokko A-La-Mode 2 really, I'm more curious about Tales Of Abyss.
I was a bit surprised to see that the rendering in Majokko worsened (the actual rescaling code was never hit in the dump), but anyway it should be properly fixed with some point list palette rendering directly in gs memory.

JordanTheToaster (approved):

coornio:
Leaving a comment here for parity's sake -- This commit appears to break certain games, such as Dynasty Warriors. The second title has FMV seizures when upscaling, the fifth title will run out of VRAM

To quote ref:
| Dynasty Warriors 5 has the same problem as Minority Report
| "Failed to allocate device memory for 65536x65536 texture"
| It's squaring the size of the texture
| That was supposed to be a 1024x1024 texture, but I'm at x8, so it resized to 8192x8192, then did it again

5d-0:
This particular PR breaks Manhunt 2, game now crashes in HW mode (no matter what api you are using) on the title screen if internal resolution is higher than default. You can actually get ingame by choosing SW mode or internal resolution being set to default and after that when you get to the menus you can notice another issue which is been introduced by this PR, it's broken scaling of level load pictures when internal resolution is higher than default. The higher internal resolution the smaller load pic becomes. That's what the log says: A lot of "GS: Memory allocation failure".

Examples: 

2x 
2x

4x 
4x

PR #5289, opened 2022-01-09
GS: Better support of dpi scaling

tellowkrinkle:
### Description of Changes
- Adds DPI change event support when available (wx 3.1.3+)
- Fixes double-applied DPI correction on GS OSD
- Fixes GS OSD not properly reacting to changes in DPI

### Rationale behind Changes
Note: None of this is applicable to Windows because wx is great there and doesn't realize DPI scaling is a thing.  It might be applicable on Linux (esp. Wayland).  It's definitely applicable on macOS.
- OSD isn't absolutely massive when running with 200% scaling
- Windows now notice when they're dragged between displays with different DPI scaling modes
- When windows notice that they were dragged between displays with different DPI scaling modes, the OSD properly rescales itself

### Suggested Testing Steps
Drag windows between displays with different DPI scaling settings

### DIFF
--- a/pcsx2/Frontend/ImGuiManager.cpp
+++ b/pcsx2/Frontend/ImGuiManager.cpp
@@ -66,7 +66,7 @@ bool ImGuiManager::Initialize()
 
 	s_global_scale = std::max(1.0f, display->GetWindowScale() * static_cast<float>(EmuConfig.GS.OsdScale / 100.0));
 
-	ImGui::GetIO().DisplayFramebufferScale = ImVec2(display->GetWindowScale(), display->GetWindowScale());
+	ImGui::GetIO().DisplayFramebufferScale = ImVec2(1, 1); // We already scale things ourselves, this would double-apply scaling
 	ImGui::GetIO().DisplaySize.x = static_cast<float>(display->GetWindowWidth());
 	ImGui::GetIO().DisplaySize.y = static_cast<float>(display->GetWindowHeight());
 	ImGui::GetStyle() = ImGuiStyle();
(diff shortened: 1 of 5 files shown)


pcsx2/Frontend/ImGuiManager.cpp
```diff
@@ -114,29 +114,26 @@ void ImGuiManager::WindowResized()
 	const float new_scale = (display ? display->GetWindowScale() : 1.0f);
 
 	ImGui::GetIO().DisplaySize = ImVec2(static_cast<float>(new_width), static_cast<float>(new_height));
-	ImGui::GetIO().DisplayFramebufferScale = ImVec2(new_scale, new_scale);
 
-	UpdateScale();
+	UpdateScale(std::max(new_scale * static_cast<float>(EmuConfig.GS.OsdScale / 100.0), 1.f));
```
stenzek, line 118: Any reason specifically behind passing this as a parameter, instead of computing it in `UpdateScale()`? (can read the display scale there).

Otherwise, LGTM

tellowkrinkle: Good question.  I have no clue why I did that
Now computed in `UpdateScale()`

stenzek (approved):

pcsx2/gui/AppHost.cpp
```diff
@@ -193,17 +195,19 @@ void Host::CheckForGSWindowResize()
 		return;
 
 	int width, height;
+	float scale;
```
lightningterror, line 198: Good practice to initialize all 3.

tellowkrinkle: They all get initialized under the lock

IMO if you think you initialize a variable on all paths that you use it, it's more obvious to future readers and static analyzers if you don't initialize to a dummy value

Mrlinkwii (approved):

PR #5292, opened 2022-01-09
GS/DX11: Implement merge feedback write

stenzek:
### Description of Changes

This was missing before, and is needed now since we default to DX11 SW on Windows.

### Suggested Testing Steps

Test games which uses it. I've checked the Xenosaga dream sequence gs dump.

### DIFF
--- a/bin/resources/shaders/dx11/convert.fx
+++ b/bin/resources/shaders/dx11/convert.fx
@@ -18,6 +18,20 @@ struct VS_OUTPUT
 	float4 c : COLOR;
 };
 
+cbuffer cb0
+{
+	float4 BGColor;
+	int EMODA;
+	int EMODC;
+};
+
+static const float3x3 rgb2yuv =
+{
+	{0.587, 0.114, 0.299},
+	{-0.311, 0.500, -0.169},
+	{-0.419, -0.081, 0.500}
+};
+
 Texture2D Texture;
 SamplerState TextureSampler;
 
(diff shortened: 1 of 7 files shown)


bin/resources/shaders/dx11/convert.fx
```diff
@@ -357,12 +357,63 @@ PS_OUTPUT ps_convert_rgba_8i(PS_INPUT input)
...
+	float4 i = sample_c(input.t);
+	float4 o;
+
+	float3 yuv = mul(rgb2yuv, i.gbr);
+
+	float Y = float(0xDB) / 255.0f * yuv.x + float(0x10) / 255.0f;
+	float Cr = float(0xE0) / 255.0f * yuv.y + float(0x80) / 255.0f;
+	float Cb = float(0xE0) / 255.0f * yuv.z + float(0x80) / 255.0f;
+
+	switch (EMODA) {
```
lightningterror, line 386: Formatting, bracket on next line, all elements inside switch case +1 more tab.

bin/resources/shaders/dx11/convert.fx
```diff
@@ -357,12 +357,65 @@ PS_OUTPUT ps_convert_rgba_8i(PS_INPUT input)
 	return output;
 }
 
-// DUMMY
+cbuffer cb0
```
lightningterror, line 360: Feels odd to put these here (cbuffer and rgb2yuv) , I'd have put them above, around line 40 would've made more sense.

lightningterror (approved):

PR #5293, opened 2022-01-09
GS/DX11: Add a shader cache

stenzek:
### Description of Changes

Adds a shader cache for DX11, similar to the other renderers.

### Rationale behind Changes

Switching hardware/software renderers is slow in DX11 because it recompiles a bunch of shaders every time.

### Suggested Testing Steps

Test software/hardware renderer switching, make sure it's faster, and that there's no random rendering errors in games.

Test shaders (fxaa/externalfx/shadeboost).


JordanTheToaster (approved):
No problems in any of the games I tested F9 switching is fast and shaders and shade boost all works LGTM.

PR #5304, opened 2022-01-10
GS: Support wrapping off-the-end offsets in new GSOffset

tellowkrinkle:
### Description of Changes
Extends GS row offset tables to 4096 entries for wrapping

### Rationale behind Changes
Fixes issue introduced by new GSOffset

### Suggested Testing Steps
Test WWE SmackDown vs. Raw 2008


refractionpcsx2:
WWE Smackdown Vs Raw 2008 works fine :) thanks :)

Is the extension of size required if it's wrapping with the % 2048?

PR #5306, opened 2022-01-10
Common: Add gsl::span and use it for GL::Context

tellowkrinkle:
### Description of Changes
- Adds LLVM's ArrayRef to Common
- Switches GL::Context's pointer + length usage to ArrayRef

### Rationale behind Changes
- Can use C++ iteration now
- The length of the return of `GL::Context::GetAllVersionsList` is no longer part of its type

### Suggested Testing Steps
Launch the emulator with a GL backend

### DIFF
--- a/3rdparty/include/gsl/algorithm
+++ b/3rdparty/include/gsl/algorithm
@@ -1,0 +1,63 @@
+///////////////////////////////////////////////////////////////////////////////
+//
+// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+//
+// This code is licensed under the MIT License (MIT).
+//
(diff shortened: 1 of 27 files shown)


CookiePLMonster:
There is also `std::span`/`gsl::span`, from what I can see by briefly looking into this code they are equivalent?
https://github.com/microsoft/GSL/blob/main/include/gsl/span

`std::span` is C++20 so that might be a dealbreaker, but I think `gsl::span` has an identical API so that'd make potential future migration simplistic.

tellowkrinkle:
Switched to `gsl`

Mrlinkwii (approved):

JordanTheToaster (approved):

PR #5310, opened 2022-01-10
GUI: Make aspect ratio/frame limiter OSD messages keyed

CookiePLMonster:
### Description of Changes
Keyed messages don't pile up so spamming turbo/frame limiter keys will not generate multiple messages now.

pcsx2-avx2-dev_e3Ei3yA3Zw

### Rationale behind Changes
OSD spam makes messages less clear to the end user.

### Suggested Testing Steps
Switch GS options via hotkeys, like aspect ratio, frame limiter, frame skipping, zoom etc.

### DIFF
--- a/pcsx2/gui/GlobalCommands.cpp
+++ b/pcsx2/gui/GlobalCommands.cpp
@@ -16,6 +16,7 @@
 #include "PrecompiledHeader.h"
 #include "MainFrame.h"
 #include "GSFrame.h"
+#include "Host.h"
 #include "ApplyState.h"
 #include "ConsoleLogger.h"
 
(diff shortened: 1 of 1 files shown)


RedDevilus:
What's the retention time for one message? 5 seconds?

CookiePLMonster:
> What's the retention time for one message? 5 seconds?

2 seconds

Mrlinkwii (approved):

stenzek (approved):
LGTM, you could probably remove `OSDlog` entirely now as well.

CookiePLMonster:
> `OSDlog`

it's still used in savestate selection so eh

PR #5324, opened 2022-01-12
GIF: Separate GIF Reset and GIF DMA Reset

refractionpcsx2:
### Description of Changes
Removes DMA reset stuff from a GIF reset.

### Rationale behind Changes
The two shouldn't be reset together anyway, the GIF has no control over the DMA, it was incorrect and causing currently running DMA's to get stuck in an infinite loop.

### Suggested Testing Steps
Run games, make sure they still work, no specific games here (Except Air Ranger 2) was known to be affected directly, but the adverse is unknown.

Fixes Air Ranger 2 - Rescue Helicopter infinite loop on boot.

### DIFF
--- a/pcsx2/Gif.cpp
+++ b/pcsx2/Gif.cpp
@@ -86,9 +86,9 @@ void GIF_Fifo::init()
 	memzero(data);
 	fifoSize = 0;
 	gifRegs.stat.FQC = 0;
-	CSRreg.FIFO = CSR_FIFO_EMPTY;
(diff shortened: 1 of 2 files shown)


HeroponRikiBestest:
Fixes air ranger 2.

JordanTheToaster (approved):

lightningterror (approved):

Mrlinkwii (approved):

PR #5346, opened 2022-01-16
GS: optimize xxHash usage in palette hashing.

iMineLink:
### Description of Changes
Optimize xxHash usage in palette hashing as suggested by @stenzek.

### Rationale behind Changes
Optimization.

### Suggested Testing Steps
I guess code review is enough here.


stenzek (commented):
LGTM

Issue #5317, opened 2022-01-11
[BUG]: Destruction Derby Arenas Lights go through walls and Fog Effect Doesn't Go Away from Beginning of Race 


ps1freak26:
### Describe the Bug

![DDA pcsx2 hw](https://user-images.githubusercontent.com/91406481/148874927-8cc443bf-72a3-470f-bcb3-4bd1fa572392.jpg)
Red fog at the introduction of race and through all of the race. Different races have different colored fog. Lights from powerups and lamps are see through on walls. Disabling Depth Emulation does get rid of fog but turns the sky black on steelworks. Disabling Depth Emulation does not effect the lights they are still see through.

### Reproduction Steps

Start any race but most noticeable in Steelworks. 

### Expected Behavior

![DDA ps2 snap 1](https://user-images.githubusercontent.com/91406481/148874914-71888ce4-5fbf-4142-a146-0d23504a94b5.jpg)
![DDA pcsx2 sw](https://user-images.githubusercontent.com/91406481/148874916-ac35b709-f3a9-47ee-84af-b7700763d4a3.png)

### PCSX2 Revision

PCSX2 Nightly - v1.7.2212

### Operating System

Windows 10 (64bit)

### CPU

i5 4950

### GPU

1660 super

### GS Settings

The settings are at their defaults expect 4x res but 1x res still has the issue expect software mode.

### Emulation Settings

All settings are at default and preset is at safe mode.

### Logs & Dumps

https://drive.google.com/file/d/1Ho9V0GG115m_xUv4RtmhX3Hp0iVUj7ME/view?usp=sharing

ps1freak26:
I've tested other versions of pcsx2 and 1.6.0 with recent nvidia graphics drivers seem to break the ui and stuff from half the screen and pcsx2 v1.7.2010 seem to have none of the issues. Here are snaps from these different build
![DDA pcsx2 1 6 0](https://user-images.githubusercontent.com/91406481/148879210-f7a74039-92a5-4e8c-8472-ee6054f0f5b2.jpg)
![DDA pcsx2 v1 7 2010](https://user-images.githubusercontent.com/91406481/148879217-9c7155aa-6d58-4179-8ef0-283bb290e319.jpg)
s

ps1freak26:
[...]

PR #5349, opened 2022-01-16
GS: preserve target on different format write. 

iMineLink:
### Description of Changes
~~Based on #3895.~~
Avoid discarding the entire target upon incompatible format write (same block pointer), just mark the related rectangle as dirty.

### Rationale behind Changes
See #5317.

### Suggested Testing Steps
Check that #5317 is fixed.
Make sure no regressions occur.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
+++ b/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
@@ -876,11 +876,40 @@ void GSTextureCache::InvalidateVideoMem(const GSOffset& off, const GSVector4i& r
 				}
 				else
 				{
-					list.erase(j);
-					GL_CACHE("TC: Remove Target(%s) %d (0x%x)", to_string(type),
-						t->m_texture ? t->m_texture->GetID() : 0,
-						t->m_TEX0.TBP0);
-					delete t;
+					// YOLO skipping t->m_TEX0.TBW = bw; It would change the surface offset results...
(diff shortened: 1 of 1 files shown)


gilderoylockhart01:
This breaks Beyond good and evil shadows which were fin with the previous hack:

image

Also, wouldn't be a better thing to move this hack to "disable safe features" instead?

refractionpcsx2:
Didn't notice any regressions, don't have an appropriate place in Destruction Derby to check either.

Mrlinkwii (approved):

iMineLink:
Ok I made the logging super paranoid, to make it easier to check what's going on.
This seems to help also https://forums.pcsx2.net/Thread-Bug-Report-Conflict-Desrt-Storm.
Also, please re-test the shadows in beyond good and evil, if they are still broken (I didn't change this PR logic since creation so it could be) I'll investigate further.

refractionpcsx2:
BGE Shadows look as good as master
Destruction Derby is fixed
Not noticed any other fixed (Don't have Conflict Desert Storm)

lightningterror (commented):
> Also, please re-test the shadows in beyond good and evil, if they are still broken (I didn't change this PR logic since creation so it could be) I'll investigate further.

They look fine, same behavior as master, needs Texture inside RT to look better.
Note: Tested only on a gsdump.

refractionpcsx2:
As noted above, improves linked issue is hugely improved in hw mode, but the top left side of the screen (Not quite like Desert Storm but kind of a similar symptom) is still black.

PR #5353, opened 2022-01-18
GS-hw: Implement BLEND CD.

lightningterror:
### Description of Changes

Allow hw blend only to do the math, output will be Cd, no need to run costly barriers.
Will also help not enabling colclip on all renderers since there is no wrapping.

Previously on these cases sw blend would run as would require costly barriers. When output is Cd(Color destination) let hw blend handle it.

Also we could potentially disable hw blend too, but let's leave it for now.
### Rationale behind Changes

Less texture barriers, optimizations.
### Suggested Testing Steps

Test games with various blending levels and make sure nothing broke.
So far the following games I spotted are: Sly 2, Remote control dandy, MGS3, Xenosaga 1, NFS Carbon and Undercover, SW Battlefront, Genji, Burnout Revenge, Final Fight streetwise and a coupe of other more.


BinhoKof99:
Why don't you solve the wrong textures of the Street Fighter and Darkstalkers collections in hardware mode???

lightningterror:
Forgot to mention but Driv3r gains around 4 fps on dx11 with the pr.

PR #5356, opened 2022-01-19
memcard: Fix folder memory card regressions

xTVaser:
### Description of Changes

Someone reported issues in GTA, i looked into it and found some bugs that were not caught during the migration to rapidyaml.  I believe they didn't come up because not all games care about the index file.
- the index file would not be initially created properly
- the index file wouldn't get updated properly when saving to slots

In the old file, `%ROOT` was used to keep track of metadata about the entire folder memory card, unfortunately rapidyaml has a small bug I'll need to upstream and it does not quote this key. `%` is a directive in YAML which are non-data args for the yaml processor, so the line gets discarded, screwing up the assumed syntax.
  - to fix this in the meantime (but i see no reason it can't remain) -- I switched to `$ROOT`

Also if the yaml file failed to parse, it would be a hard-crash, wrapped it with error handling like with the gameDB

### Rationale behind Changes

Breaks folder memory cards, atleast in games that require the index file, like GTA.

### Suggested Testing Steps

Please test with the same games / scenarios that were involved in this PR - https://github.com/PCSX2/pcsx2/pull/3941
- Use a Legacy Folder Memory Card (build before November 2020)
- Use a Folder Memory Card saved to prior to v1.7.2155 (they may or may not be usable due to these bugs - especially for games like GTA that require the index file)
- Manually screw up your pcsx2_index file -- things will still break, but the emulator will no longer crash and will log a warning

### DIFF
--- a/pcsx2/MemoryCardFolder.cpp
+++ b/pcsx2/MemoryCardFolder.cpp
@@ -1,5 +1,5 @@
 /*  PCSX2 - PS2 Emulator for PCs
- *  Copyright (C) 2002-2015  PCSX2 Dev Team
+ *  Copyright (C) 2002-2022  PCSX2 Dev Team
  *
  *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
  *  of the GNU Lesser General Public License as published by the Free Software Found-
@@ -22,6 +22,7 @@
 #include "System.h"
 #include "Config.h"
 
+#include <common/FileSystem.h>
 #include "common/StringUtil.h"
 
 #include "fmt/core.h"
@@ -56,18 +57,13 @@ std::optional<ryml::Tree> loadYamlFile(const wxString& filePath)
 	});
 	try
 	{
-		auto path = Path::FromWxString(filePath);
-		if (!fs::exists(path))
-		{
+		std::optional<std::string> buffer = FileSystem::ReadFileToString(StringUtil::wxStringToUTF8String(filePath).c_str());
+		if (!buffer.has_value())
 			return std::nullopt;
-		}
-
-		auto stream = fs::ifstream(path);
-		std::stringstream buffer;
-		buffer << stream.rdbuf();
+		const ryml::Tree tree = treeFromString(buffer.value());
 
 		ryml::reset_callbacks();
-		return std::make_optional(treeFromString(buffer.str()));
+		return std::make_optional(tree);
 	}
 	catch (const std::exception& e)
 	{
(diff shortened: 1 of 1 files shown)


pcsx2/MemoryCardFolder.cpp
```diff
@@ -41,17 +44,39 @@ const ryml::Tree treeFromString(const std::string& s)
...
+	ryml::set_callbacks(rymlCallbacks);
+	c4::set_error_callback([](const char* msg, size_t msg_size) {
+		throw std::runtime_error(fmt::format("[YAML] Internal Parsing error: {}",
+			msg));
+	});
+	try
 	{
-		return std::nullopt;
-	}
+		auto path = Path::FromWxString(filePath);
```
stenzek, line 59: All of this can be replaced by `FileSystem::ReadFileToString()` (and remove the wxString dependency as well).

xTVaser: imo I would save that for when this file is properly cleaned of wx types, there are over a hundred wx dependencies remaining in this file.

stenzek: But why introduce more? It's just adding more work for later.

refractionpcsx2: 1 out of over 125 isn't really a big deal, with 125 or 126 the person who converts this file is going to hate themselves :P

stenzek: Right, but with all respect, this is written pretty poorly. Redundant reads/copies, when there's a function which does all the work in our wrapper namespace (~10 lines down to 1). It's pretty similar to the bit I replaced in gamedb loading.

Literally just
```
std::optional<std::string> buffer = FileSystem::ReadFileToString(StringUtil::wxStringToUTF8String(path).c_str());
if (!buffer.has_value())
  return std::nullopt;
const ryml::Tree tree = treeFromString(buffer->c_str());
```

I know the old code was written this way, but given it's being modified anyway, it makes sense to switch it over and save time later on.

refractionpcsx2: Then I would recommend we make it work by massaging it for now, then somebody will need to rewrite it to make it not suck. Maybe Silent would like to look at it since he added all the index stuff.

xTVaser: Suggestion committed, thanks.

JordanTheToaster (approved):

Mrlinkwii (approved):

PR #5361, opened 2022-01-19
Qt Part 3

stenzek:
### Description of Changes

New core/VM/input systems. Everything except the dependencies and Qt directory itself.

### Suggested Testing Steps

Make sure nothing is broken in wx. It shouldn't be.

### DIFF
--- a/pcsx2/Frontend/SDLInputSource.cpp
+++ b/pcsx2/Frontend/SDLInputSource.cpp
@@ -1,0 +1,513 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2022  PCSX2 Dev Team
+ *
+ *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
+ *  of the GNU Lesser General Public License as published by the Free Software Found-
+ *  ation, either version 3 of the License, or (at your option) any later version.
+ *
+ *  PCSX2 is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+ *  PURPOSE.  See the GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along with PCSX2.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "PrecompiledHeader.h"
+#include "Frontend/SDLInputSource.h"
+#include "Frontend/InputManager.h"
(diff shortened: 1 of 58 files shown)


MrCK1:
Tested all general functions without issues

JordanTheToaster (approved):

pcsx2/VMManager.h
```diff
@@ -0,0 +1,190 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2021 PCSX2 Dev Team
```
lightningterror, line 2: 2022 copyright headers.

Mrlinkwii (approved):

RedDevilus (approved):

pcsx2/Frontend/SDLInputSource.cpp
```diff
@@ -0,0 +1,513 @@
...
+	{
+		Console.WriteLn("Enabling PS4/PS5 enhanced mode.");
+#if SDL_VERSION_ATLEAST(2, 0, 9)
+		SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS4, "true");
+		SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS4_RUMBLE, "true");
+#endif
+#if SDL_VERSION_ATLEAST(2, 0, 16)
+		SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS5, "true");
+		SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS5_RUMBLE, "true");
+#endif
```
tellowkrinkle, line 88: Can we include a method to set hints with a configuration file, [like we do on current PCSX2](https://github.com/PCSX2/pcsx2/blob/e5c4fe003b9cfd1821fe81ae48f7d5069d75e865/pcsx2/PAD/Linux/SDL/joystick.cpp#L46-L47)?

Also, can you set these using strings so that they'll automatically be enabled if the user swaps out the sdl dylib for a newer one without recompiling?

Also, please set the `SDL_GAMECONTROLLER_USE_BUTTON_LABELS=0` and `SDL_JOYSTICK_HIDAPI_SWITCH_HOME_LED=0` hints as well.

Side note: We currently require SDL 2.0.9.  You're welcome to lower that, but don't feel like you have to

stenzek: I'm planning on static linking SDL2 (with only the controller support enabled), so this won't be a problem.

Options can be added later when I finish the controller settings UI.

PR #5365, opened 2022-01-21
Debugger & Core: Lighten IOP breakpoints and fix the memory search

F0bes:
### Description of Changes
- **Enables interpreter breakpoints and memchecks on dev & debug builds**
- Removes the unnecessary calls to standardizeBreakpointAddress

	standardizeBreakpointAddress is pointless on the IOP, and when a memcheck is active _any_ load / store instruction executed will call this pointless function.
- Fixes up the memory search window, it used to by default cover the memory view.

	I tried to make it DPI aware, but wx isn't very helpful and Qt is coming soon anyways. So expect non 100% DPIs to be less than optimal (but still functional).

- Fixes the memory search window getting screwed up on resize.

### Rationale behind Changes
Trying to get more familiar with the core. Oh and speed when debugging :)

### Suggested Testing Steps
Test memory checks on the IOP. Test the debugger resizing and stuff.

### DIFF
--- a/pcsx2/DebugTools/Breakpoints.cpp
+++ b/pcsx2/DebugTools/Breakpoints.cpp
@@ -31,15 +31,7 @@ std::vector<MemCheck *> CBreakPoints::cleanupMemChecks_;
 bool CBreakPoints::breakpointTriggered_ = false;
 
 // called from the dynarec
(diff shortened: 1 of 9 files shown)


F0bes:
Quick tests, just in a game menu (not IOP heavy)
400% with no memory checks
250% with memory check on pr
230% with memory check on master
:)

JordanTheToaster (approved):

Issue #5081, opened 2021-11-28
Meta: Auto Flush destroying performance


RedDevilus:
### Description

Games that tank performance when using autoflush (SW and HW). On SW it's enabled by default. On HW it's disabled by default.

- [ ] Alter Echo
- [ ] Need For Speed Underground 2
- [ ] Need For Speed Undercover
- [ ] Hitman Contracts
- [ ] Spider-Man: The Movie
- [ ] Powerdrome

### Reason

Better to have a tracker for games that don't like autoflush.

### Examples

Alter Echo;
SW autoflush: 42% (20.79)
SW no autofl: 250% (124.90)
Lost VPS = -208%  (-104.11)

![SWAutoFlushAlterEcho1](https://user-images.githubusercontent.com/24227051/143778389-1a37e824-0e1a-4315-8b43-9415f1a1642d.png)
![SWNoAutoFlushAlterEcho2](https://user-images.githubusercontent.com/24227051/143778398-f4fe401b-2e14-4043-bfd4-31684434e1ce.png)

8x HW autoflush: 86% (43.02)
8x HW no autofl: 154% (76.93)
Lost VPS = -68%  (-33.91)

![8xHWAutoFlushAlterEcho3](https://user-images.githubusercontent.com/24227051/143778407-b2b341a2-9699-4a3d-a97c-2b150583dfb3.png)
![8xHWNoAutoFlushAlterEcho4](https://user-images.githubusercontent.com/24227051/143778412-63a53c54-6850-4709-a3ff-94e680efbfdf.png)

Satan86:
Need For Speed Underground 2 and Undercover also lose tonnes of FPS with Auto Flush
On
![image](https://user-images.githubusercontent.com/61447828/143788575-bce68b16-bddd-4a1d-b221-54ade00de29f.png)
![image](https://user-images.githubusercontent.com/61447828/143788788-7f04bb18-0028-4a3c-90c0-4ef7ef7025cc.png)

Off
![image](https://user-images.githubusercontent.com/61447828/143788582-d1f55e18-05ab-47a9-8875-6def8ab15e1b.png)
![image](https://user-images.githubusercontent.com/61447828/143788874-32962c0a-d8cf-4aef-8b95-435fa086d978.png)
This is on Software mode

gilderoylockhart01:
There are several games where autoflush is super slow on hardware but fast on software, which is... weird.

RedDevilus:
Tested on 1.7.2415;

[...]

PR #5377, opened 2022-01-22
GS: Only autoflush on move/self blend (Perf test)

refractionpcsx2:
Also Move overlap check to State

### Description of Changes
Attempts to improve performance when Auto Flush is enabled

### Rationale behind Changes
Auto Flush was very aggressive, this aims to cut down on aggressive flushing to hopefully improve performance a bit when Auto Flush is enabled (Especially in software)

### Suggested Testing Steps
Test games that require auto flush, make sure they aren't broken, compare performance with master.

Note: This might never get merged if it breaks a ton of stuff! If not then it's a good bonus!


RedDevilus:
https://github.com/PCSX2/pcsx2/issues/5081

Might help?

refractionpcsx2:
> #5081
> 
> Might help?

Forgot there was a meta, thanks!

DonelBueno:
Is there a list of games that need autoflush? The only ones I know are the Syphon Filter ports from the PSP.

JordanTheToaster:
Substatial performance increase with autoflush in software on NFS Underground 2.

Master
AF-Mster 1


PR
AF-PR 1

refractionpcsx2:
> Is there a list of games that need autoflush? The only ones I know are the Syphon Filter ports from the PSP.

Not sure there is, I know Driv3r and GTA SA need it, but beyond that, I'm basically relying on the Wiki

RedDevilus:
Default HW no autoflush:
image
Default SW with autoflush:
image
60-140 VPS (50 VPS is the normal because PAL)
HW with autoflush:
image
250-330 VPS but with artifacts

RedDevilus:
[AlterEchoHWautoflush.zip](https://github.com/PCSX2/pcsx2/files/7918921/AlterEchoHWautoflush.zip)

gilderoylockhart01:
I can handle the testing for that, I have several games which I know requires it at specific spots or with some specific effects.

I will update the thread with what I find.

HelicopterP:
From the top of my head - Burnout 3 (sun is shinig through objects without it) and Jak 2 (shadows are messed up without it) need autoflush. I am unable to test this games at the moment, unfortunatelly.

refractionpcsx2:
I just checked Jak 2, the shadows look the same to me with or without Auto Flush (even on master).  Not sure the best place to check Burnout 3, a GS Dump would be cool.

gilderoylockhart01:
In Jak 2 the performance is almost not improved at all, actually, the game needs auto flush for an other rendering problem which is hard to track:

When approaching a volumetric effect, there is a sort of trail following the character, this effect needs auto flush in order to work as you can see here:

No AF:
Capture d’écran 2022-01-22 192727

AF:
Capture d’écran 2022-01-22 192709

The problem here is the absolutely terrible framerate, it's extremely slow with the hardware render in comparison to the software render.

refractionpcsx2:
My Apologies, the HW auto flush option inteferes with the SW one, so they both need to be off if you want to toggle it for one renderer, mildly annoying! maybe we should combine the boxes.

Anyway I just made another change which makes it auto flush less, might be a bunch faster in some games.

gilderoylockhart01:
Yeah, better combining this for booth render, it's far from a hack anyway.

Also, as a side note, some games will need to have auto flush AND memory wrapping enabled for some effects to work, memory wrapping is still a hack, which makes... little sense I guess considering it's more accurate to what a real console does.

gilderoylockhart01:
In Jak 2, I can get full speed with auto flush and native resolution, anything higher (even 2x) will be super slow.

refractionpcsx2:
weird, for me Jak 2 (at the start of the game) I go down from 115fps to 83fps, roughly, that's at x4

gilderoylockhart01:
Excuse me, I didn't explain enough, Jak 2 will only be super slow when in front of that volumetric effect, if not the game will be fast yeah. I guess I will create a gsdump

As for Tekken Tag tournament, it's a *massive* improvement, goes from 50 ish % to almost 400% in the menus at 8x.

gilderoylockhart01:
Here is the GSdump for Jak 2 and it's slow auto flush problem:
[Jak 2 slow auto flush.zip](https://github.com/PCSX2/pcsx2/files/7919094/Jak.2.slow.auto.flush.zip)

gilderoylockhart01:
MGS2 transparency effects in the intro are not super slow as they used to be with auto flush hardware (there are still some slowdowns and minor glitches but those are different problems):

https://user-images.githubusercontent.com/90222976/150652647-e08ad719-ed53-4950-97f1-c0995cc5cb98.mp4

refractionpcsx2:
Okay I see the problem with Jak 2, the number of texture copies blows up, so yeah that will screw the speed, that will have to be a thing for now.

gilderoylockhart01:
007 is now fully accurate to what a real PS2 does (auto flush helps with the car's reflections), and this without tanking the frame-rate, isn't that beautiful ^^ (the only little gremlin is minor Z-fighting but that is common on the hardware renders):

https://user-images.githubusercontent.com/90222976/150653905-a0a595de-e89c-4c05-926b-1ebc000c6740.mp4

(PS, yeah, terrible video quality but hey, 10mb limit it is).

refractionpcsx2:
Nice :)

gilderoylockhart01:
I just have a random question about Ratchet shadows, why does the Vulkan and OpenGL hardware render renders them fine without auto flush?

refractionpcsx2:
> I just have a random question about Ratchet shadows, why does the Vulkan and OpenGL hardware render renders them fine without auto flush?

on DX Auto Flush is kind of a hack for Ratchet because DX doesn't have barriers.

gilderoylockhart01:
Ok, I said that because it's painfully slow with DX yeah.

gilderoylockhart01:
Mmm, I have one weird case, catwoman is still fine with the software render (the game needs auto flush to prevent an image ghosting problem), but the hardware render is still bad, even with auto flush on, that's not an issue with this PR, but it is still something worth to look into:

Hardware render:

gs_20220123141311

Here is the dump:
[Catwoman broken auto flush hardware.zip](https://github.com/PCSX2/pcsx2/files/7920558/Catwoman.broken.auto.flush.hardware.zip)

HelicopterP:
Ok, this is just for completion sake as this 2 games were mentioned before. Here are GS dumps with sun shining through a bridge in B3 and messed up shadows in J2 without autoflush. This two problems are unrelated to this PR. As far as I can see the behaviour is the same as master. Autoflush should be enabled in software mode for this two games to fix glitches. I repeat: unrelated to this PR. End of offtopic :p
[Burnout 3 sun.zip](https://github.com/PCSX2/pcsx2/files/7920622/Burnout.3.sun.zip)
[Jak 2 shadows.zip](https://github.com/PCSX2/pcsx2/files/7920623/Jak.2.shadows.zip)

PR #5386, opened 2022-01-23
GS-hw: Check if writable alpha is masked on Ad case, do sw blend.

lightningterror:
### Description of Changes

GS-hw: Check if writable alpha is masked on Ad case, do sw blend. 
opengl/vulkan: Blending can be done with one barrier which is much faster.
d3d11: Read the fb and do the blend, slower but it's doable on dx too.
### Rationale behind Changes

Less full texture barriers vk/gl, also allows d3d11 to sw blend with reading the fb but slower than gl/vk.
The specific blending can be done on Basic level.
### Suggested Testing Steps

Test basic blending, all hw renderers, list of affected games:
image


lightningterror:
Replaced by #5452

PR #5399, opened 2022-01-24
GS: revert target scaling.

iMineLink:
### Description of Changes
Fix a missing texture scale setting and then revert for now the target rescaling code introduced by #4094 and completed by #5285.

### Rationale behind Changes
The GS currently does allocate an upscaled target even for native resolution draws.
This was not taken into account in the recent changes and caused double upscaling of native resolution drawn targets (before the missing texture scale setting the target would be upscaled indefinitely).
Until the GS code is changed to allow native resolution targets for native resolution draws even with upscaling, this rescaling code must not be executed.

### Suggested Testing Steps
Check that the regressions spotted in #5285 are fixed.
No game was fixed by the cited PR in particular.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
+++ b/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
@@ -556,6 +556,7 @@ GSTextureCache::Target* GSTextureCache::LookupTarget(const GIFRegTEX0& TEX0, int
 			g_gs_device->StretchRect(dst->m_texture, sRect, tex, dRect, ShaderConvert::COPY, false);
 			g_gs_device->Recycle(dst->m_texture);
 			dst->m_texture = tex;
+			dst->m_texture->SetScale(new_s);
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Mostly okay, unfortunately this breaks the upscaling in some games.  Here is a pack of 3 game gsdumps (god of war 2, test drive, vakyrie profile 2), only 2x upscale is required to see the issues, hopefully you can triage it for now until the fix comes :)
[gsdumps.zip](https://github.com/PCSX2/pcsx2/files/7924361/gsdumps.zip)

5d-0:
Both issues with Manhunt 2 (crash and scaling) been fixed with this PR.

refractionpcsx2:
I have done a revert with #5404

Unfortunately this caused more issues, probably likely because of the original problems I have with the burnout PR, but reverting just the newer PR resolves most of the problems. So gonna close this one for now.

PR #5400, opened 2022-01-24
GS: Prefer one-frame-old textures

tellowkrinkle:
### Description of Changes
Prefer one-frame-old textures instead of new textures

### Rationale behind Changes
The main reason for preferring new textures is to prevent needing to split a render pass to upload to the texture (in Vulkan or Metal).  This split only happens if the texture was last written in the same frame it's being uploaded to.

By preferring one-frame-old textures we can improve texture reuse without affecting draw performance

### Suggested Testing Steps
Test Vulkan and watch memory usage (should hopefully be lower) and performance (should hopefully not be lower)

### DIFF
--- a/pcsx2/GS/Renderers/Common/GSDevice.cpp
+++ b/pcsx2/GS/Renderers/Common/GSDevice.cpp
@@ -125,7 +125,7 @@ GSTexture* GSDevice::FetchSurface(GSTexture::Type type, int width, int height, i
 
 		if (t->GetType() == type && t->GetFormat() == format && t->GetSize() == size && t->GetMipmapLevels() == levels)
 		{
-			if (!prefer_new_texture)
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Tested a few games, at x4 the vram usage dropped by about 100mb (about 10%) or none at all, depending on the game.

tellowkrinkle:
This is going to affect games with a lot of large textures (not render targets) the most (e.g. Baldur's Gate)

refractionpcsx2:
I tried Baldurs gate, I think that was one of the ones who saved about 100mb :P

hunter1998w:
i tested Legacy of Kain - Defiance, it save about 150 mb, at 1080P

stenzek (approved):
LGTM

PR #5411, opened 2022-01-26
GS-hw: Implement alternative hw blending for Cs*Ad, Cs*Ad + Cd, Cd - Cs*Ad

lightningterror:
### Description of Changes

GS-hw: Implement alternative hw blending for `Cs*Ad, Cs*Ad + Cd, Cd - Cs*Ad.`

Alpha destination value is wrong so let us try to compensate.

Multiply Cs by (255/128) in shader.
Will work best if Cs is 0.5 or less,  if it's higher than 0.5 then the closer to 1 the less accurate it gets.

Ofc it is best to use sw blending but it will help if sw blending is not present/selected for specific draw, will help d3d11 quite more.

GS-hw: Adjust how clr_blend is handled based on blend level on d3d11.
Blend level High and lower: Prefer clr_blend.
Blend level Full, prefer clr_blend when ALPHA.C != 1, otherwise sw when primitives don't overlap.
Blend level Ultra, remains the same but we can probably prefer clr_blend when alpha again is As or F, let's play it safe and leave it as it is for now.
### Rationale behind Changes

Accuracy, will help all renderers when sw blending is not present, especially dx11 since it lacks more
### Suggested Testing Steps

Some games I spotted, also test various blending levels on renderers, especially dx11.
image


Mrlinkwii (approved):

lightningterror:
Some eye candy for DX11 users, ultra blending is used on all renderers for best comparison.
OpenGL is used as a reference comparison since sw blending is supported.

Destroy all humans
DX11 PR:
image

GL:
image

DX11 Master:
image

NFS Carbon:
DX11 PR:
image

GL:
image

DX11 Master:
image

Scarface
DX11 PR:
image

GL:
image

DX11 Master:
image

GT4:
DX11 PR:
image

GL:
image


DX11 Master:
image

DonelBueno:
This makes GT3 and GT4 identical to software mode with blending accuracy on medium or lower when using Vulkan, which gives a nice speed boost. Before this PR, they required blending on high. =D

PR #5412, opened 2022-01-26
GS-wx: Add extra check for shader options when VK is selected

F0bes:
### Description of Changes
Thank you to AndroBanani#2026 on discord for reporting this issue.

image

Switching to D3D11 / OpenGL and checking `Enable Shader Boost` or `Enable External Shader` and switching to Vulkan will leave those inputs enabled.

### Rationale behind Changes
Those inputs should be disabled as they aren't used when Vulkan is selected.

### Suggested Testing Steps
Test the Shader panel in the GS settings window, switch the adapters, etc.


JordanTheToaster (approved):
Working as intended on Windows.

tellowkrinkle (approved):

PR #5419, opened 2022-01-28
GS: target rescaling take 2.

iMineLink:
### Description of Changes
#5285 take 2.
Added full support for target rescaling in GS TC, respecting the texture aspect ratio.
Avoid redundant clears when the target is going to be completely drawn when rescaling/converting.
Merge the two ways to lookup render targets (for draw and frame) into a single slightly more complex method, to recycle the rescaling logic and other code.
Avoid using upscaled size for targets requested at native resolution when upscaling.
Correctly set texture scale for targets created in a frame lookup.
Update targets from GS memory before converting color to depth.
Update targets validity region when uploading dirty data from GS memory (especially useful in combination with preload frame data, which makes the initial validity region of the target non empty).

### Rationale behind Changes
Improve TC behaviour when handling targets with different scaling, for example native res and upscaled, or potentially when changing internal resolution without flushing the TC.

### Suggested Testing Steps
Test some games and check that no visual regression occurs, especially games with "CU" hacks that have explicit native res draws even when upscaled (currently only "Tales Of Abyss").
Potentially appreciate performance improvement due to less target clearing and smaller render targets when big ones are not required.
Maybe some visual bug is fixed by updating the color targets before converting to depth, or by correctly setting texture scale for targets created in a frame lookup, or by updating the target validity region on dirty data upload from GS memory.
Test Manhunt 2 and Dynasty Warriors 5 which had severe regressions with #5285.

### DIFF
--- a/pcsx2/GS/GSVector.h
+++ b/pcsx2/GS/GSVector.h
@@ -72,6 +72,16 @@ public:
 	{
 		return x != v.x || y != v.y;
 	}
+
+	constexpr GSVector2T operator*(const GSVector2T& v) const
+	{
+		return { x * v.x, y * v.y };
+	}
+
+	constexpr GSVector2T operator/(const GSVector2T& v) const
+	{
+		return { x / v.x, y / v.y };
+	}
 };
 
 typedef GSVector2T<float> GSVector2;
--- a/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
@@ -333,7 +333,7 @@ GSTexture* GSRendererHW::GetOutput(int i, int& y_offset)
 
 	GSTexture* t = NULL;
 
-	if (GSTextureCache::Target* rt = m_tc->LookupTarget(TEX0, m_width, m_height, GetFramebufferHeight()))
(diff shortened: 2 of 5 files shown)


Tzerinas:
Tested, and I couldn't reproduce the issue with Manhunt 2, Dynasty Warriors 2 and Dynasty Warriors 5, except the Memory allocation issue above native for Manhunt2. (Works at 2x, but breaks above that for me)
That happens on Master as well though (with the commit reverted), so I'm not sure about this issue.

refractionpcsx2:
Tested a whole bunch of dumps and games, didn't notice any problems or fixes, except Gundam looks broken, however this was only a dump which only works with preload frame in hardware mode, this doesn't work with the actual game.

Tzerinas:
*Test some games and check that no visual regression occurs, especially games with "CU" hacks that have explicit native res draws even when upscaled (currently only "Tales Of Abyss").*

This CU hack does not seem to apply anymore. Which is a shame, because it + Half-Pixel Offset (Special-Texture) looked near perfect. I didn't see a commit that took this out, so is this a bug?

Tzerinas:
Or maybe I'm just going insane. Older versions seem clearer on the title screen, but that's it.

PR #5437, opened 2022-01-31
GUI: rename  preset 1 and  description

Mrlinkwii:
### Description of Changes

renames preset 1 and changes its  description to something more appropriate 
### Rationale behind Changes

To make it clear that preset 1 should be only used when requested 
### Suggested Testing Steps

make sure pcsx2 dosent  explode and the text is correct


RedDevilus:
Might as well remove it, Safest naming was basically a lie (slowest + more compatibility issues than the other 2 presets)

PR #5440, opened 2022-01-31
CI: Build Qt binaries on Windows

stenzek:
So I stop being a derp and pushing changes which don't compile.


stenzek:
NOTE: Qt doesn't require AVX2, but it's silly to spam more jobs on the runners when they're not going to be used.

xTVaser (approved):
Looks good to me, build is building, and the excluded one is excluded.  Changes to the build are what i was imagining.

PR #5443, opened 2022-01-31
GameDB: Urban Reign

RedDevilus:
### Description of Changes

EE cyclerate +1 will fix the issue but a compromise on default is to enable EE-Timing-hack. Ee cyclerate +1 can be combined with or without EETiming.
### Rationale behind Changes

It's a compromise to improve users by default even if it's the best fix)
### Suggested Testing Steps

Check the menus and in-game without changing settings for vertical shaking (like similar on other games that shake the screen on different interlacing settings.) Urban Reign shake is very severe on None Interlacing so maybe check there too (yeah It's not the correct naming for it)

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -763,7 +763,9 @@ SCAJ-20151:
   region: "NTSC-Unk"
 SCAJ-20152:
   name: "Urban Reign"
-  region: "NTSC-Unk"
+  region: "NTSC-J"
+  gameFixes:
+    - EETimingHack # Mitigates bounciness of vertical shaking but better fix with EE cyclerate +1.
 SCAJ-20153:
   name: "Code Age Commanders"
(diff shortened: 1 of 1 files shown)


Mrlinkwii (approved):

RedDevilus:
Yeah the full fix is changing EECyclerate to + 1
image

AhayriSG:
> Yeah the full fix is changing EECyclerate to + 1 but I can't force that via GameDB. image

I remember this story from somewhere :face_in_clouds: 
https://github.com/PCSX2/pcsx2/issues/569#issuecomment-1004889920

AhayriSG:
> Yeah the full fix is changing EECyclerate to + 1 but I can't force that via GameDB. image

https://github.com/PCSX2/pcsx2/issues/569#issuecomment-1004889920

gilderoylockhart01:
This does not happen on the pal version for some reasons.

PR #5422, opened 2022-01-30
GS: Get rid of extra binding for channel shuffle

stenzek:
### Description of Changes

Having this binding was redundant, as there's no "normal" texture sampled when we're doing a channel shuffle, and it caused issues in Vulkan when the render target or depth buffer the source.

Also fixes the Urban Chaos HLE shader.

Fixes validation errors in GT4, NFS: Carbon, Urban Chaos, probably others too.

For DX11, does a smaller copy for the RT when it's copied for sampling, which is a significant improvement (9fps->25fps with ultra blending in xenosaga, 3fps->25fps at 4x).

### Suggested Testing Steps

Test blending and channel shuffle games.

PR #5447, opened 2022-02-01
GS/HW: Don't disable depth testing for channel shuffle

stenzek:
### Description of Changes

Regression from #5422.

Mercenaries needs it. But we can skip it when z testing is off.

Handles this case in Vulkan, which would otherwise be a validation error (also affects Ico).

Technically we can handle this as another feedback loop, but in my opinion it's not worth it, since there's only two games I've seen which do this - Ico and Mercenaries.

### Suggested Testing Steps

Test Mercenaries and Ico. Nothing else seems affected that we know of.


lightningterror (approved):
Didn't spot an issue with the dumps I tested that use CS.

PR #5452, opened 2022-02-02
GS-hw: Implement hw, hw/sw, sw blending on Ad when alpha write is masked, and more optimizations

lightningterror:
### Description of Changes

GS-hw: Implement hw, hw/sw, sw blending on Ad when alpha write is masked.
Idea is to replace Ad with As when alpha write is masked,
then expand/let blend mix, accumulation blend non recursive blend or hw clr blend to
do the blending with Ad swapped as As.

We are doing this to try to bring some originally higher blending modes to lower levels
where we can do the draws with less texture barriers instead (gl/vk),
as for d3d11 this allows to run blending on the draws since previously the cases weren't handled properly,
it will be slower on d3d11 since we will be reading the frame buffer but it's better than nothing.

D3D11: It is enabled on Medium blending or higher, if draw is fbmask then it will enable
it on basic blending too.

OpenGL/Vulkan:
It is enabled based on the previous blending modes:
accumulation blend -> either minimum or basic level, depending on colclamp.
non recursive blend -> either minimum or basic level, depending on colclamp.
blend mix -> basic and higher level.
hw clr blend -> minimum and higher level.

All:
Prefer full sw blend when primitives don't overlap, sw fbmask or full barrier is used, it is more accurate.

GS-hw: Combine BLEND_C_CLR2_AF and BLEND_C_CLR3_AS in one shader bit. 

Free a shader bit.

S-hw: Move the sw blend for BLEND_C_CLR3 to High level and upper. 
Reading the frame buffer is slow, let's allow hw blend to run, can partially handle it.

GS-hw: Allow to run clr_blend1 and clr_blend2 over sw blending.
As or F case only, no Ad.
Allow condition to run when there are no texture barriers (gl/vk),
or copying the fb (d3d11).
Run when colclamp is 1 as hw blend will clamp value to 0-1.
Check for texture barriers, don't run if any is enabled.

### Rationale behind Changes

Accuracy, optimizations.
### Suggested Testing Steps

Test various blending levels on all hw renderers.

### DIFF
--- a/bin/resources/shaders/dx11/tfx.fx
+++ b/bin/resources/shaders/dx11/tfx.fx
@@ -769,17 +769,16 @@ void ps_blend(inout float4 Color, float As, float2 pos_xy)
 
 			Color.rgb = (float3)255.0f;
 		}
-		else if (PS_CLR_HW == 2 || PS_CLR_HW == 3)
+		else if (PS_CLR_HW == 2)
 		{
-			// PS_CLR_HW 2 Af, PS_CLR_HW 3 As
 			// Cd*As or Cd*F
 
-			float Alpha = PS_CLR_HW == 2 ? Af : As;
+			float Alpha = PS_BLEND_C == 2 ? Af : As;
 
 			Color.rgb = max((float3)0.0f, (Alpha - (float3)1.0f));
 			Color.rgb *= (float3)255.0f;
 		}
-		else if (PS_CLR_HW == 4)
+		else if (PS_CLR_HW == 3)
 		{
 			// Needed for Cs*Ad, Cs*Ad + Cd, Cd - Cs*Ad
 			// Multiply Color.rgb by (255/128) to compensate for wrong Ad/255 value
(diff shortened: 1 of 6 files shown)


lightningterror:
Some eyecandy, God of war now needs basic blending on gl/vk, medium on dx11.
And let's compare God of war vs master dx11:
PR:
image

Master:
image

Harry Potter COS, ultra blending dx11
PR:
image

Master:
image

OpenGL/Vulkan for reference:
image

JordanTheToaster:
NFS Carbon draw calls and GPU usage are massively down with this PR.

Master
image

PR
image

refractionpcsx2:
NFS MW previously needed high and had a perf hit.
Now on the PR Basic is enough, and even if you use high, there isn't a perf loss anymore.

Master VK Basic
image

Master VK High
image

PR VK Basic
image

Mrlinkwii (approved):

JordanTheToaster (approved):

tellowkrinkle (approved):

DonelBueno:
I notice that the performance increases in blending heavy scenes when changing to the software renderer and then going back do the HW renderer, at least on Vulkan. Do you know why? Is this a bug or the intended behavior?

PR #5457, opened 2022-02-02
cmake: copy resources in build folder.

iMineLink:
### Description of Changes
copy resources from bin to pcsx2 target build folder when building with cmake.

### Rationale behind Changes
make it easier to run the pcsx2 target via vscode cmake integration on windows.

### Suggested Testing Steps
check that building and running an all platforms is not broken.
on windows, check that you can run the pcsx2 target via vscode cmake integration and it finds the resources folder.


tellowkrinkle:
`add_custom_command` lets you specify multiple commands to run in series
If you do the two commands separately, on multicore builds, they could run out of order AFAIK

iMineLink:
thanks for the tip, now a single `add_custom_command` call specifies both directories creation and file copy commands.

PR #5465, opened 2022-02-04
Add option in build.sh to build with qt.

arcum42:
### Description of Changes
Adds a --qt option to build.sh, to build with qt.

### Rationale behind Changes
You can now do a qt build with build.sh.

### Suggested Testing Steps
Build pcsx2 with qt.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -196,6 +196,7 @@ for ARG in "$@"; do
         --use-system-yaml   ) flags="$flags -DUSE_SYSTEM_YAML=TRUE" ;;
         --asan              ) flags="$flags -DUSE_ASAN=TRUE" ;;
         --gtk2              ) flags="$flags -DGTK2_API=TRUE" ;;
+        --qt                ) flags="$flags -DQT_BUILD=TRUE" ;;
         --lto               ) flags="$flags -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=TRUE" ;;
         --pgo-optimize      ) flags="$flags -DUSE_PGO_OPTIMIZE=TRUE" ;;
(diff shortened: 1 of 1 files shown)


stenzek:
Not sure if this is a great idea yet. Qt isn't ready for user use, and it doesn't even build on linux (since nobody ships qt6).

arcum42:
> Not sure if this is a great idea yet. Qt isn't ready for user use, and it doesn't even build on linux (since nobody ships qt6).

I put "experimental" next to the option to indicate that it isn't ready. I could add "Do not use" or some such. (And come to think of it, maybe it should be --qt6 instead of just qt...)

I do actually have qt 6 installed on Linux on my system (Arch has it in the repositories) and it did build and run. I was really largely adding it to make it easier for me to work on the qt side of things.

stenzek:
What's the point if it only works on one distribution? And not the the OS that the builders use (Ubuntu) and probably a majority of users use.

We also need to patch Qt to fix the incorrect scaling of icons for non-integer scales. I did do a Qt build on Ubuntu 18.04 after a fair bit of messing around, but it ends up linking against an older version of icu and some float conversion library, so the resulting binaries don't work on 20.04 and newer. They would in the appimage, since they'd get pulled in, but not for someone trying to develop/work on PCSX2.

At that point I just threw my hands in the air and said "well, it's someone else's problem, I'm only worried about windows anyway". So if you have any ideas, I'm all ears. But just saying "screw anyone who doesn't use Arch" is unacceptable IMO.

arcum42:
Alright, was trying to get it where I could easily work on the qt side with Linux, but I'll leave qt to the Windows people then.

stenzek:
Not saying that Qt should be Windows-only, just that we need to find a solution which works with both the AppImage as well as the majority of distributions.

PR #5466, opened 2022-02-04
Debugger: add register name patterns

Tokman5:
### Description of Changes

The debugger uses 2 digits for floating point number registers' number (f00, f01... f31), but you can't use this format in ``Assemble Opcode(s)`` when the register's number is less than 10.
This PR adds ``f00`` - ``f09`` pattern so that debugger will be able to handle them correctly.
### Rationale behind Changes

Less confusing

For example, instruction code 0x46140800 is shown as ``add.s f00,f01,f20`` in debugger.
However, when you open ``Assemble Opcode(s)`` and type the same ``add.s f00,f01,f20`` opcode, you will get parameter failure error because PCSX2 can only find ``f0`` - ``f9`` for registers' name.
You should input ``add.s f0,f1,f20`` instead on master.
### Suggested Testing Steps

Try new opcodes which include ``f00`` - ``f09`` register.


refractionpcsx2:
Works great :) Thanks!

PR #5473, opened 2022-02-05
Qt: Resize Windows + Change Framerate Presets + Lower OSD time

RedDevilus:
### Description of Changes

The main window will now correctly display the full Toolbar. Interface Settings will not go out of bounds. The OSD message for cheats and patches has been changed and the overall timing of the static (not dynamic) OSD messages are now 3 seconds instead of 10 seconds which is pretty long. Tested on 1920 x 1080 + 2560 x 1080 on 125% and 150% DPI scaling.

Main Window;
Before (Master):
Master
After (PR):
PR

Don't worry about the bad scaling, it's a Qt bug with 125% DPI.
150%1080P

Speed percentage steps (presets);
Before (Master):
Master2
After (PR):
PR2

OSD naming change:
Before (Master):
Master3
After (PR):
PR3

### Rationale behind Changes

Keeping the end-user in mind with uncommon preferences without breaking the common preferences.
### Suggested Testing Steps

Get Qt to work and check all the window and if everything is visible and legible.

### DIFF
--- a/pcsx2-qt/MainWindow.ui
+++ b/pcsx2-qt/MainWindow.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>800</width>
-    <height>700</height>
+    <width>950</width>
+    <height>640</height>
    </rect>
   </property>
   <property name="acceptDrops">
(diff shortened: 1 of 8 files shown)


stenzek:
3 seconds is way too low for people to read the message, especially those not who are not native English speakers. 10 is a good balance, and with the keyed messages, it's not like it'll flood the screen.

RedDevilus:
The old OSD was 2 seconds, but I'll make it 5 seconds at minimum for now and see what users think of it when the time comes.

stenzek (approved):
LGTM

PR #5476, opened 2022-02-05
Fix _BitScanForward and _BitScanReverse on LP64 platforms.

MaddTheSane:
### Description of Changes
This changes `_BitScanForward` and `_BitScanReverse` to take in `unsigned int`s instead of `unsigned long`s.

### Rationale behind Changes
This quiets a warning for compilers where `-Wshorten-64-to-32` is enabled by default.

### Suggested Testing Steps
None come to mind.


MaddTheSane:
Bleh, stupid Windows…

tellowkrinkle:
Maybe replace them with a new function (with a new name) that takes a `u32` and either calls the Windows one or does the current thing on unix

PR #5477, opened 2022-02-05
GS: Remove upscaling dependency on PCTRC enabling.

iMineLink:
### Description of Changes
Remove the dependency of the GS upscaling from the PCRTC enabling (`PRMODE.EN1 | PRMODE.EN2`).
Skip draws where the frame buffer width is set to 0 or where it is detected that no color nor depth write is going to happen.

### Rationale behind Changes
The GS draw processing is not related to the PCRTC display.
A comment in the code that added the dependency of the upscaling to the PCRTC cited that when in the "ps2 logo to game transition" the display size may not be set correctly, so the draws where the frame buffer width is not defined or there is no output are now skipped.

### Suggested Testing Steps
Test various games at upscaled resolution, check that no visual regression occur.
Potentially some upscaled draws may have been improved (if the game renderered them without the PCRTC output buffer enabled, ref cited an homebrew which may have applied some texture decompression algorithm in this way).
Test the "ps2 logo to game transition" that was reported to have issues if allowed to render with upscaling.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
@@ -1241,6 +1241,12 @@ void GSRendererHW::Draw()
 	GSDrawingContext* context = m_context;
 	const GSLocalMemory::psm_t& tex_psm = GSLocalMemory::m_psm[m_context->TEX0.PSM];
 
+	if (!context->FRAME.FBW)
+	{
+		GL_CACHE("Skipping draw with FRAME.FBW = 0.");
+		return;
+	}
+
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
The BIOS/Game transition seems fine to me, I hadn't noticed any problems.

Haven't noticed any other differences besides that.

gilderoylockhart01:
Poosibly one more hack which is not needed anymore. And I guess plenty of them will also need to be removed, I spoted in the code some *horrible* ones like hardcodded crc hacks which just needs to be... eliminated ^^.

image

Mrlinkwii (approved):

Issue #5388, opened 2022-01-23
[Feature Request]: Moving the COP2 bad div/q detection system to retail builds.


gilderoylockhart01:
### Description

In PCSX2, there is a detection system that allows devs to spot a couple of specific COP2 errors due to bad div or q timings (excuse me if that's a terrible term ^^) and will print those to the console log.

This is very useful in case a game will have issues with the COP2 even if the issue is not obvious or even apparent.

Currently, you need a devel build or upper for that to work, this is a bit inconvenient as the dev builds are compiled with the release flag, and it's not exactly slower or anything.

### Reason

Makes testing *a lot* easier and allows the end users to knows if the game is running badly.

+ this will ease the work of bug squad and testers members, the user would just have to post the emulog, and they will already know where the problem might come from if it's related to that.

And yeah, this can have false positive but the persons who will write the patch will likely know how to separate those from the real ones.

### Examples

Devil May Cry has a COP2 problem which is *impossible* to notice with a release build, the patch was created for that thanks to that detection system, so why not the end users could take advantage of that?

F0bes:
I don't think I have enough time between now and work to make up a pr but I'd like benchmarks to be made.
Considering during recompilation [this](https://github.com/PCSX2/pcsx2/blob/4ccb922d1ab10b95f7ee391f2bcd2bedacaecdec/pcsx2/x86/ix86-32/iR5900-32.cpp#L1596-L1659) is going to be run now. And it's a lot of loops and branches. Maybe a game that likes to do self modifying code would be a good test.

PR #5480, opened 2022-02-05
COP2: Move COP2 timing messages to release builds.

F0bes:
### Description of Changes
The COP2 pipeline isn't fully emulated. A way to fix games that depend on the pipelined behaviour is to patch them. In order to find out what games need patching we can look at the EE code we are compiling and do some heuristics.

#5388 requested to allow this code to run in release builds as well.

### Rationale behind Changes
Allowing this extra logging can help us find what games need patching, not everyone is going to test their games with devel builds and this information can prove to be useful.

### Suggested Testing Steps
This needs to be bench marked against a release build with this disabled. This code is run during recompilation, so a game that likes to do a lot of self modifying code would be a great candidate. 
You can compare against this build: https://github.com/PCSX2/pcsx2/actions/runs/1799763630

---
fixes: #5388

### DIFF
--- a/pcsx2/x86/ix86-32/iR5900-32.cpp
+++ b/pcsx2/x86/ix86-32/iR5900-32.cpp
@@ -1174,7 +1174,6 @@ static void iBranchTest(u32 newpc)
 	}
 }
 
-#ifdef PCSX2_DEVBUILD
 // opcode 'code' modifies:
 // 1: status
 // 2: MAC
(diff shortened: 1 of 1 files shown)


gilderoylockhart01:
I ran Deadly Skies III which is a self-modifying heavy game, I see no performance regression of any sort. I also tested Test Drive Unlimited which is a self-modifying game and also run into the COP2 log, same trend, no perfmance regression at all.

PR #5485, opened 2022-02-06
Fix some vsync related issues, implement Vulkan adaptive vsync

stenzek:
### Description of Changes

Vsync was flaky since the frontend/host display refactoring (my fault). This fixes it and makes it reliable again.

Also adds adaptive vsync (FIFO_RELAXED) for Vulkan.

### Rationale behind Changes

Fixing my bugs.

### Suggested Testing Steps

Make sure vsync works as expected now.

### DIFF
--- a/common/Vulkan/Context.cpp
+++ b/common/Vulkan/Context.cpp
@@ -291,7 +291,7 @@ namespace Vulkan
 	}
 
 	bool Context::Create(std::string_view gpu_name, const WindowInfo* wi, std::unique_ptr<SwapChain>* out_swap_chain,
-		bool threaded_presentation, bool enable_debug_utils, bool enable_validation_layer)
+		bool vsync, bool threaded_presentation, bool enable_debug_utils, bool enable_validation_layer)
 	{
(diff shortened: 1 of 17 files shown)


gilderoylockhart01:
I have a question regarding Vsync, does it takes into account the Gsync technology for 50Hz games? (by that I mean, lowering down the refresh rate of the monitor)?

JordanTheToaster (approved):

stenzek:
How VRR gets activated depends on the API.

gilderoylockhart01:
With an Nvidia GPU, this will work with DirectX 9/10/11/12 if the window is in fullscreen and exclusive mode (or in flip mode).

stenzek:
Not according to MS's documentation: https://docs.microsoft.com/en-us/windows/win32/direct3ddxgi/variable-refresh-rate-displays

We set this flag when vsync is **off**.

gilderoylockhart01:
Okay, for some weird reasons this did work with those settings on my system in the past. Nevermind, now with the PR it works as expected.

PR #5487, opened 2022-02-06
GS-hw-tc: Remove JackieChanAdv crc hack.

lightningterror:
### Description of Changes

GS-hw-tc: Remove JackieChanAdv crc hack.
Game seems to work fine currently.
### Rationale behind Changes

Less hacks, not needed.
### Suggested Testing Steps

[JA_Halfscreen_Flicker_hitolddepthbuffer.zip](https://github.com/PCSX2/pcsx2/files/8010057/JA_Halfscreen_Flicker_hitolddepthbuffer.zip)

#1838

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
+++ b/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
@@ -102,16 +102,7 @@ GSTextureCache::Source* GSTextureCache::LookupDepthSource(const GIFRegTEX0& TEX0
 	if (!m_can_convert_depth)
 	{
 		GL_CACHE("LookupDepthSource not supported (0x%x, F:0x%x)", TEX0.TBP0, TEX0.PSM);
-		if (m_renderer->m_game.title == CRC::JackieChanAdv || m_renderer->m_game.title == CRC::SVCChaos)
-		{
-			// JackieChan and SVCChaos cause regressions when skipping the draw calls when depth is disabled/not supported.
-			// This way we make sure there are no regressions on D3D as well.
-			return LookupSource(TEX0, TEXA, r, false);
-		}
-		else
-		{
-			throw GSRecoverableError();
-		}
+		throw GSRecoverableError();
 	}
 
 	const GSLocalMemory::psm_t& psm_s = GSLocalMemory::m_psm[TEX0.PSM];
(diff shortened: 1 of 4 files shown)


JordanTheToaster (approved):

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -99,19 +99,15 @@ void GSTextureCache::RemoveAll()
 
 GSTextureCache::Source* GSTextureCache::LookupDepthSource(const GIFRegTEX0& TEX0, const GIFRegTEXA& TEXA, const GSVector4i& r, bool palette)
 {
-	if (!m_can_convert_depth)
+	if (m_renderer->m_game.title == CRC::SVCChaos)
+	{
+		// SVCChaos black screen on main menu, regardless of depth enabled or disabled.
+		return LookupSource(TEX0, TEXA, r, false);
```
gilderoylockhart01, line 105: Is this really needed? Having hardcodded crc hacks is really bad. I guess a patch would be a better solution to this.

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -196,21 +192,11 @@ GSTextureCache::Source* GSTextureCache::LookupDepthSource(const GIFRegTEX0& TEX0
...
-			// (don't exists yet for us due to the cache)
-			// Rendering is nicer (less garbage) if we skip the draw call.
-			throw GSRecoverableError();
-		}
-
-		//ASSERT(0);
+		// Full Spectrum Warrior: first draw call of cut-scene rendering
+		// The game tries to emulate a texture shuffle with an old depth buffer
+		// (don't exists yet for us due to the cache)
+		// Rendering is nicer (less garbage) if we skip the draw call.
```
gilderoylockhart01, line 198: The comment is a bit misleading, this suggest that this code path is only for Full Spectrum Warrior, maybe make it a bit more generic.

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
lightningterror: You're welcome to submit a proper fix.

gilderoylockhart01: I don't have the game, nor the knowledge of how to fix this, I just asked if it is still needed, which is apparently the case if you said so. Why this can't be moved to the crc hack database instead?

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
refractionpcsx2: And completely outside the scope of this PR, this PR isn't a "lets critique the texture cache" PR, it's removing a CRC hack.  If you have problems with other parts of the file, make a new issue.

gilderoylockhart01: Okay okay, this isn't meant to be free "critique", sorry if you felt it was the case, that's fine then.

Issue #2059, opened 2017-08-30
'High fidelity' v-sync mode


mirh:
*As tangentially suggested by [gregory](https://github.com/PCSX2/pcsx2/issues/2053#issuecomment-325966462).*
The idea went through a lot of mind work so I'll also describe this. 

At first I was going to just propose a "High Fidelity" mode (aka pcsx2 sets monitor refresh rate to match games one) unconditional to everything else, toggled by an additional checkbox. 

Assuming nobody would want to go down the scary rabbit hole of "[vendors](http://gpuopen.com/gaming-product/amd-gpu-services-ags-library/) [private](https://www.intel.com/content/www/us/en/support/graphics-drivers/000007175.html) [tools](https://developer.nvidia.com/nvapi)*" I realized only a "selection between already available modes" would have been possible with normal Windows functions. 
Which means you'd have to warn users for when "their expectation cannot be met" *and* instruct them for [how to](https://www.monitortests.com/forum/Thread-Custom-Resolution-Utility-CRU) [work around](https://forum.doom9.org/showthread.php?t=173571). 

So this got me thinking about who really is the kind of individual that (putting aside bugs in drivers, please) would find all this tinkering (and the possible nuisance of monitor 'loosing connection' to adjust to newer settings) worth it. 
Because I mean, perhaps it's my lazy eye, but even with PAL games under 60Hz DWM.. I don't really feel stutter. 

On the other hand, this would probably be a big deal breaker if I were to use v-sync (let's put aside the compositing variable for simplicity). It'd be bonkers to sync frame rendering to the wrong clock. 

Which ultimately lead us to my tl;dr:
* v-sync menu gets labeled refresh rate control
* disabled/adaptive stays basically like now
* standard gets named #something# and not only enforce v-sync but also try to instate the right refresh rate

Notes about the later mode:
* query available resolutions before set attempt, and warn+fallback if there's is none multiple of the desired one
* at least on linux possibly we could try ourselves to fix the problem for the user (ask him?)? [th](https://dri.freedesktop.org/docs/drm/gpu/i915.html#display-refresh-rate-switching-drrs)[is](https://lists.freedesktop.org/archives/intel-gfx/2013-November/036648.html)?
* as I were saying somewhere else, test whether we can **always** safely assume late vsync > normal 
(very important for the 'perfection' of the still 3 choices menu)
* maybe delay of some seconds this function on games boot, to allow them to set the right whatevers once?
* on game closure original resolution should restored 

[...]

PR #5488, opened 2022-02-06
GS: Add sync to host refresh rate option

stenzek:
### Description of Changes

This PR implements an option I had in DuckStation - to run the emulation approximately 0.1% faster, so that the guest vertical frequency matches the host refresh rate. It gets rid of the duplicate frames you would otherwise get around every 18 seconds or so.

Since we're running the emulation ever-so-slightly faster, the SPU output rate also changes to ~48048hz. Because we're using cubeb now, we just forward that through, and it takes care of resampling it for us.

In addition, when the host vsync and guest vsync queue size is set to 0, we can get rid of sleep-based frame pacing completely, relying exclusively on the host vsync. Polling controllers after host vsync should result in the lowest possible input latency, and the most stable frame times since we're not at the mercy of the OS scheduler so much.

For best results, use with exclusive fullscreen on D3D. This is only available in Qt, I mostly implemented it in Wx, only to find that it's still using that parent/child window mess, which prevents it, and I don't feel like rewriting that disgustingness.

To enable vsync-based pacing:
 - Set your host to 60hz or 59.94hz (for NTSC games), or 50hz (for PAL games). I'm considering making this switch automatic in the future.
 - Enable sync to refresh rate (GS window in wx, emulation settings in qt).
 - Set the vsync queue size to 0 (GS in wx, emulation settings -> optimal frame pacing in qt).
 - Enable vsync on the host (GS window in wx, graphics settings -> display in qt).

If it's working, it should display a message in the OSD saying it's using vsync for frame pacing.

I'm curious if anyone sensitive to input latency can notice an improvement here. Because I sure as heck can't :D

### Rationale behind Changes

Getting rid of the meme that pcsx2 sucks for input latency, when it doesn't.

### DIFF
--- a/pcsx2-qt/EmuThread.cpp
+++ b/pcsx2-qt/EmuThread.cpp
@@ -27,6 +27,7 @@
 #include "common/StringUtil.h"
 
 #include "pcsx2/CDVD/CDVD.h"
+#include "pcsx2/Counters.h"
 #include "pcsx2/Frontend/InputManager.h"
 #include "pcsx2/Frontend/ImGuiManager.h"
 #include "pcsx2/GS.h"
@@ -361,6 +362,9 @@ void EmuThread::setFullscreen(bool fullscreen)
 	m_is_fullscreen = fullscreen;
 	GetMTGS().UpdateDisplayWindow();
 	GetMTGS().WaitGS();
+
+	// If we're using exclusive fullscreen, the refresh rate may have changed.
+	UpdateVSyncRate();
 }
 
 void EmuThread::setSurfaceless(bool surfaceless)
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -435,6 +435,7 @@ struct Pcsx2Config
 					PCRTCOffsets : 1,
 					IntegerScaling : 1,
 					LinearPresent : 1,
+					SyncToHostRefreshRate : 1,
 					UseDebugDevice : 1,
 					UseBlitSwapChain : 1,
 					DisableShaderCache : 1,
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -31,6 +31,8 @@
 
 #include "ps2/HwInternal.h"
 #include "Sio.h"
+#include "HostDisplay.h"
+#include "SPU2/spu2.h"
 
 #ifndef PCSX2_CORE
 #include "gui/App.h"
@@ -46,6 +48,7 @@ using namespace Threading;
 extern u8 psxhblankgate;
 static const uint EECNT_FUTURE_TARGET = 0x10000000;
 static int gates = 0;
+static bool s_use_vsync_for_timing = false;
 
 uint g_FrameCount = 0;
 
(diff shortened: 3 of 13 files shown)


Immersion95:
Hi stenzek, what would be the behaviour with 120hz host ? Would it consider it being a multiple of 60hz ?

tabnk:
Gran Turismo 4 frame pacing issue almost fully fixed with this build except **George V Paris** track got tiny pacing issue when making a quick turn.

nixxquality:
> Getting rid of the meme that pcsx2 sucks for input latency, when it doesn't.
Admitting you have a problem is the first step.

I checked out the Qt version from the CI build and applied all of the settings you recommended.
My go to test here is Space Channel 5, as it's a game I had to give up on and use a Dreamcast emulator on before.

Unsurprisingly, it is still unplayably bad. Try it yourself if you don't believe me. Then maybe you can take the first step.

refractionpcsx2:
> > Getting rid of the meme that pcsx2 sucks for input latency, when it doesn't.
> > Admitting you have a problem is the first step.
> 
> I checked out the Qt version from the CI build and applied all of the settings you recommended. My go to test here is Space Channel 5, as it's a game I had to give up on and use a Dreamcast emulator on before.
> 
> Unsurprisingly, it is still unplayably bad. Try it yourself if you don't believe me. Then maybe you can take the first step.

Did you adjust your audio latency to also be as low as possible? By default it's quite high but Cubeb (xaudio sucks if you're a stickler for latency) can go pretty low, about 1 frame's worth of delay.

refractionpcsx2:
yes the Qt audio configuration is currently missing, hence us saying that Qt is unfinished and not ready for public consumption :P

Tzerinas:
Would it be possible to make a low audio latency checkbox (like optimal frame pacing) that automatically sets it to low latency with the tooltip mentioning its use?

RedDevilus:
Cubeb already automatically sets it much lower than the default 100ms. If it can't like a weak(er) PC keep up it will increase it's latency but is still lower than 100ms. You can test this doing it in WxWidgets , setting it to cubeb and watching the program log/ console / emulog.

Unix0815:
Could you please do a new build, based on the newest nightly? Your build is the only one I'm getting silky smooth framerates with perfekt frame pacing and vsync. With all official versions, I'm getting stuttering even with vsync.

RibShark:
If this is added, please make sure users are aware that their game is running slightly faster than real hardware. There are many speedrun communities that allow PCSX2, and if this is accidentally enabled during a speedrun it could cause issues.

refractionpcsx2:
This would close #2059

mirh:
Yes, but I would hope there could be a check for VRR first of all. 
No need of tinkering with refresh rates then. 

Also, the thing should accept *any* bigger multiple of the guest frequency, not just the same exact one. 

And last but not least having emulation run faster is awful. If a monitor can do 60hz, then even 59.94 and 50 are going to be possible. 
If d3d doesn't let you switch to them because they aren't available options in the supported resolutions query, you should just instruct the user to use the gpu control panel or CRU to manually add them.

refractionpcsx2:
> And last but not least having emulation run faster is awful. 

Are you honestly going to notice your game running 0.06fps faster? Unless you're some freaking super human, I don't think so.

mirh:
Likely not. On the other hand games themselves have had problems with far less AFAICT. 
I just cannot see what's the upside considering it's a workaround for a solved problem.

refractionpcsx2:
> Likely not. On the other hand games themselves have had problems with far less AFAICT. I

Are you talking about adjusting the internal framerate? because that's not what this does, it just makes the framelimiter in PCSX2 run a bit higher *if* this happens at all, the game is still running at the same speed as the PS2 expects, so I don't see how it would cause issues.

mirh:
Right.. _of course_. 
Well, I guess my annoyance just comes from the "order of preference". 
After some deep mental gymnastic, more than the already very solvable 59.94~60.0hz case, I could see this making sense in certain edge cases (my QHD monitor that at most could be overclocked to 147hz over HDMI 2.0, or whatever android port that may arise in the future)

But even if minimally, changing the speed should only be the last resort (if even, given you could be as well within the variable range of a freesync monitor). 

So I would imagine the settings something like a checkbox for "care about frequency and try to switch", and then only when you enable that "speedshift" becomes toggleable.

refractionpcsx2:
well VRR already does 59.94 anyway. it is for monitors that can't, it will only go to available refresh rates, with VRR obviously being anything over like, 35hz or whatever.

Issue #5426, opened 2022-01-30
[BUG]: Broken splashscreen in This is Football 2002


RedDevilus:
### Describe the Bug

So I was testing some random games on a recently dev build and noticed a bug:
Splashscreen is wrong:
![image](https://user-images.githubusercontent.com/24227051/151708739-5d257ec9-822e-4d67-afbf-7fbab20edcb0.png)

Seems to be broken in v1.7.2282 (tex in rt) which isn't what I expected (Thought it was 2283 for sure)

### Reproduction Steps

I tested it with a psmagazine demo disc 2001:
![image](https://user-images.githubusercontent.com/24227051/151708796-eb570543-4c6e-49ad-9cf7-450d5263a2e5.png)
Then booted the game , press start to skip FMV and then see broken splashscreen.

### Expected Behavior

![image](https://user-images.githubusercontent.com/24227051/151708833-b2face23-e0c8-4c27-8aef-443ebbf4b26e.png)

### PCSX2 Revision

v1.7.2183

### Operating System

Windows 10 (64bit)

### CPU

i7-9750H

### GPU

RTX 2060M

### Logs & Dumps

![image](https://user-images.githubusercontent.com/24227051/151708911-b0604566-d502-444f-951f-d44f85d16200.png)

lightningterror:
A blockdump would be nice.

RedDevilus:
https://we.tl/t-JCbriPJEVs

iMineLink:
This is the breaking commit: https://github.com/PCSX2/pcsx2/commit/c1e5e45431863bc206686c49b8fb4fa2775be9a2

Especially the changes in the lines regarding `GSTextureCache::Target::m_TEX0.TBW` updates.
Reverting those changes fixes the issue. But IIRC this will break Indiana Jones and the Emperor's Tomb.

In practice, the buffer width is used to regulate the uploads of the dirty data from local GS memory and also to compute the tex in rt and frame offsets.

It is not easy to attach a real buffer width to something that is not a persisted entity on the GS memory, but just a region of memory considered with a given buffer width for a specific GS command.
But by improving the way the uploads are handled (not merging all the dirty regions in a big one, but maybe splitting them by buffer width) the buffer width attached to a target can be just updated on draw/frame lookup and this may avoid breaking the tex in rt / frame offset calculations while preversing the capability of uploading EE writes in the correct texture region.

PR #5495, opened 2022-02-06
GS: update target pitch on write.

iMineLink:
### Description of Changes
Overwrite the target pitch when an overlapping write is detected with the write pitch itself.
Partially reverts c1e5e45431863bc206686c49b8fb4fa2775be9a2 (part of #3895), the commit which fixed Indiana Jones and the Emperor's Tomb.

### Rationale behind Changes
fixes the broken splashscreen in This is Football 2002 (#5426).

### Suggested Testing Steps
Check the This is Football 2002 splashcreen (#5426).
Check that Indiana Jones and the Emperor's Tomb is not broken (initial castle FMV and in game textures, especially the map texture).


RedDevilus:
Seems to work again:
image

PR #5501, opened 2022-02-08
DEV9: Merge settings file into Pcsx2Config

TheLastRar:
### Description of Changes
Integrate DEV9's main settings into the main Pcsx2Config
DNS hosts are still stored in a separate file for the time being

Various other changes are also included;
Fix an off-by-one error checking that LBA is in range for HDD access,
Use fs namespace instead of ghc,
Swap char strings with std::string in various locations in DEV9,
Cleanup unneeded functions
Remove LibXml dependency from CMake files,

### Rationale behind Changes
Currently DEV9 uses a different config format for Windows (ini) and Linux (xml), which doubles the effort for adding new settings.
Unifying the config code reduces this duplicated effort, and removes a dependency on LibXml.

### Suggested Testing Steps
Test configuring DEV9 settings, including during emulation
Test HDD files in paths containing Unicode characters
Test that networking still works


TheLastRar:
HDD size is now saved as the number of sectors, rather than a MiB size
I hadn't originally accounted for that for creating a HDD image, I've now fixed those bugs

Blackbird88:
Everything still works fine for me 👍
Even path such as this one still works
`HddFile=F:\\Consoles\\PS2\\🍪⟑η∏☉ⴤℹ︎∩₲ ₱⟑♰⫳🐱\\DEV9hdd.raw`

RedDevilus (approved):

PR #5513, opened 2022-02-11
CMake: fix Xrandr detection

SupervisedThinking:
### Description of Changes
At the moment CMake looks for `libxrandr.so` but `libXrandr.so` would be the correct shared object. Since the name it is looking for probably never exists the check will always fail.

- fix path to includes
- fix lib name

### Rationale behind Changes
- enable Xrandr


arcum42:
Seems like it detects XRandr either way on my system. Could be distribution dependent, though.

tellowkrinkle (approved):
It'll detect it either way if you have pkg-config installed

PR #5521, opened 2022-02-13
actions: Add windows category for SDL

RedDevilus:
### Description of Changes

Upcoming Qt will feature native support for multiple controllers + remapping for all relevant OSes. This makes the step a bit easier.

I've also tested it locally via Python so you can check out the text file which is appended to the PR if it looks fine which it does in my eyes.
### Rationale behind Changes

Windows was left out for automatic mapping :(

Basically:
Windows has cleaner code and no more DS4/DualSense/Switch Pro/... wrapper such as steam or DS4Windows (ryochan7) though will still work if you want it.
Linux will have OnePad legacy remapping
Mac will have feature parity to some extent.

### Suggested Testing Steps

In future test out-of-the-box mapping


.github/workflows/update-controller-db.yml
```diff
@@ -21,9 +21,9 @@ jobs:
...
         uses: peter-evans/create-pull-request@v3
         with:
-          title: "pad-linux: Update to latest controller database"
-          commit-message: "pad-linux: Update to latest controller database."
+          title: "pad-windows/unix: Update to latest controller database"
+          commit-message: "pad-windows/unix: Update to latest controller database."
           committer: "PCSX2 Bot <PCSX2Bot@users.noreply.github.com>"
           author: "PCSX2 Bot <PCSX2Bot@users.noreply.github.com>"
-          body: "Weekly automatic update of SDL Controller DB for Linux / Mac OS"
+          body: "Weekly automatic update of SDL Controller DB for Windows / Mac OS / Linux"
```
Florin9doi, line 28:
```suggestion
          body: "Weekly automatic update of SDL Controller DB for Windows / macOS / Linux"
```

PR #5523, opened 2022-02-14
GS-hw: Prefer sw blend over hw blend on high modes.

lightningterror:
### Description of Changes

Vulkan/Opengl: Disable accumulation blend on Ultra, prefer full sw blend instead, for clr_blend1_2 prefer sw blend if available over hw.

Direct3D11: Disable accumulation blend on ultra when prims don't overlap, prefer sw blend instead. also prefer sw blend if available over hw for clr_blend1_2.

### Rationale behind Changes

Accuracy, it is better to prefer sw blend on high/full/ultra if available.
### Suggested Testing Steps

Test blending modes on all hw renderers.


lightningterror:
> Probably a silly question, but any correlation in preferring sw blends in some situations equating to lower overall performance as well aside from an accuracy increase? Considering this change affects all higher levels of blending, I'm just trying to wrap my head around on whether this would be anything noticeable on the max framerate. Idle thoughts really, not criticizing or anything :)

Idea is it's more accurate, I'll redo the change that it only affects Ultra blending, however I want to know if it's worth doing where a game actually needs it, might also leave out dx11.

tellowkrinkle:
You don't have to disable accumulation blend, as you can't run into rounding issues with just addition
(Rounding problems happen with multiplication)

PR #5532, opened 2022-02-15
Qt: Add per game settings UI

stenzek:
All the emucore/implementation side of this was already done, this adds the UI to configure settings per game in Qt.

Right click in the game list and select properties, or for games which are manually booted, View -> Game Properties.

This also removes game fixes from the global settings, since there's no reason you should be setting any of those hacks glboally.

Note: I still haven't figured out what to do with the spinner boxes for per-game settings. Maybe make the global setting bold to differentiate when it has been changed.. not sure yet. So they just get disabled in game properties. But everything else should be fine.

### DIFF
--- a/pcsx2-qt/CMakeLists.txt
+++ b/pcsx2-qt/CMakeLists.txt
@@ -39,6 +39,9 @@ target_sources(pcsx2-qt PRIVATE
 	Settings/AdvancedSystemSettingsWidget.cpp
 	Settings/AdvancedSystemSettingsWidget.h
 	Settings/AdvancedSystemSettingsWidget.ui
+	Settings/AudioSettingsWidget.cpp
+	Settings/AudioSettingsWidget.h
+	Settings/AudioSettingsWidget.ui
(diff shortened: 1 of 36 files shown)


andercard0:
> It's not failing to hide the names perse, it's forcefully setting the option to enabled again. Somehow losing persistence in the ini files and forgetting the user setting it to disabled.

Not only this but I've noticed that the position of the zoom is not being memorized as well. if, that matters :)

RedDevilus:
Also why Qt is not given to the public, it's WIP and still has lots of work to do.

codered21:
> Reporting the bug here so developers won't forget to take a look when have time!
> 
> As reported to Stenzek yesterday there is a bug when set it to grid mode view and untick "show titles in (grid view)"
> 
> image
> 
> if you close and reopen emulator the option is reseted as follows:
> 
> image
> 
> Showing the name of the games back beneath it's covers

where can i find this version ?

F0bes:
It's not ready for public use yet.

Issue #3120, opened 2019-09-23
SPU2-X: Jak X (NTSC) 'Hang Time!' counter audio issue


TheTechnician27:
***PCSX2 version:***
v1.5.0-dev-3283-ge2d899231

***PCSX2 options:***
Interpolation: 4- Catmull-Rom
Output Module: 1- XAudio 2
Synchronizing Mode: TimeStretch (default configuration values)
Audio Expansion Mode: Stereo
Preset: 3 - Balanced
MTVU: Enabled

***In-game settings:***
Sound mode: Mono, Stereo, and Surround (the issue persists for all three options)

***Plugins used:***
GSdx 1.1.0
SPU2-X 2.0.0
LilyPad 0.12.1

***Description of the issue:***
In Jak X, there's a counter labeled 'Hang Time!' that appears on-screen when you get a significant amount of air time. This counter makes a ticking sound while incrementing and stops when you hit the ground. In PCSX2, the audio plays at the right time, but for whatever reason, instead of [the normal ticking sound effect](https://www.youtube.com/watch?v=lwYtRAkUjRg&t=12m24s) (it's somewhat faint here), it makes a loud, cacophonous honking noise (I don't have a clip of it, but it's easily achievable and recognizable in-game).

***How to reproduce the issue:***
Enter an event in Jak X with a large jump (for a new save file, the easiest place to do this is in the third stage of the mandatory tutorial). Get enough air time that the 'Hang Time!' counter on the right appears and begins to increment.

Note for tester(s): the game will hang upon being saved (auto-saves do not do this), but the data is actually saved, so if this happens, you just need to reload the game. Also, the audio will be muted until the main menu if you select 'Boot ISO (fast)', so don't let that bother you.

***PC specifications:***
CPU: Ryzen 7 1700 @3.4 GHz
RAM: 16 GB DDR4 @2933 MHz
GPU: GTX 1070
OS: Windows 10 64-bit

atomic83GitHub:
Can you try to put delay cycle to 3 and test the sound?

atomic83GitHub:
Can you send me a blockdump of your game with a savestate where the problem happen? Use your favorite file hosting service to upload it.

TheTechnician27:
@atomic83GitHub Sure thing. The archive file containing the dump and the save state [is here](https://send.tresorit.com/a#F1AjsYDn1j0nrkzzVamXmA). The password is just your username.

[...]

PR #5533, opened 2022-02-15
SPU2: Don't cap pitch setting on register write.

Ziemas:
While 0x3FFF is indeed the highest playable pitch I think that the
register is 16 bits wide and that the cap is enforced by the pitch
counter algorithm.

This fixes the air time ticking sound in Jak X.

### Suggested Testing Steps
Test to make sure there where no adverse affects in other titles, this code is used by everything.


TheTechnician27:
Tested Sly 2, Ratchet 2, Battle for Bikini Bottom, Scooby-Doo! Night of 100 Frights, Thrillville: Off the Rails, Jak X, MGS3: Subsistence, Duelists of the Roses, Conflict: Vietnam, and ATV Offroad Fury using this PR. On default audio settings, no audio issues detected whatsoever, and the PR seems to perfectly fix #3120 as intended.

JordanTheToaster (approved):

PR #5540, opened 2022-02-18
GS-hw: Don't wrap on 16bit destination format for blend mix.

lightningterror:
### Description of Changes

GS-hw: Rename alpha_clamp to blend_mix.

GS-hw: Don't wrap on 16bit destination format for blend mix.
Helps Dog's Life blending, we shouldn't wrap as we do more blend math afterward which will be wrong if we wrapped before.
### Rationale behind Changes

Solves blending issues on Dog's life.
### Suggested Testing Steps

Test blending levels


refractionpcsx2:
We should note this is also wrong as that means the result of the blend is stored as 32bit and not 16bit. We might be lucky and it doesn't really matter, but we should be aware that later we might have to say no hardware blend on 16bit target in order to get the correctly truncated colour values.

refractionpcsx2 (approved):

PR #5554, opened 2022-02-20
GS-hw: Optimize blending equations based on alpha value.

lightningterror:
### Description of Changes

Gs-hw: Optimize blending equations based on alpha value.
It will allow us to use free sw blending without texture barriers.

Will be especially helpful for opengl/vulkan.
### Rationale behind Changes

Optimization, brr.
### Suggested Testing Steps

Test blending, opengl/vulkan should see a speed bump.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererNew.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererNew.cpp
@@ -540,23 +540,67 @@ void GSRendererNew::EmulateBlending(bool& DATE_PRIMID, bool& DATE_BARRIER)
 
 	// Compute the blending equation to detect special case
 	const GIFRegALPHA& ALPHA = m_context->ALPHA;
+
+	// Set blending to shader bits
+	m_conf.ps.blend_a = ALPHA.A;
+	m_conf.ps.blend_b = ALPHA.B;
+	m_conf.ps.blend_c = ALPHA.C;
+	m_conf.ps.blend_d = ALPHA.D;
+
+	// Get alpha value
+	const bool alpha_c0_zero = (m_conf.ps.blend_c == 0 && GetAlphaMinMax().max == 0);
+	const bool alpha_c0_one = (m_conf.ps.blend_c == 0 && (GetAlphaMinMax().min == 128) && (GetAlphaMinMax().max == 128));
+	const bool alpha_c0_high_max_one = (m_conf.ps.blend_c == 0 && GetAlphaMinMax().max > 128);
+	const bool alpha_c2_zero = (m_conf.ps.blend_c == 2 && ALPHA.FIX == 0u);
+	const bool alpha_c2_one = (m_conf.ps.blend_c == 2 && ALPHA.FIX == 128u);
+	const bool alpha_c2_high_one = (m_conf.ps.blend_c == 2 && ALPHA.FIX > 128u);
+
+	// Optimize blending equations, must be done before index calculation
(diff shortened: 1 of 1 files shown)


lightningterror:
Sly 2 pr 128 fps , master 115 fps - OpenGL
Vulkan and dx remain around the same. I assume for amd it will help on Vulkan too.

tellowkrinkle:
Instead of flagging specific entries with another flag, apply this rule: "If C is 128 && D == B then swap D and A".  This will rely on our existing handling of the A == B case to also handle D == B cases when B is known to be multiplied by 1.

Also for future changes, C being 128 removes the chance of HW blend rounding issues, so you can use HW blend in Ultra

We should also work on normalization to reduce the number of different pipelines
If A == B then set A, B, and C all to zero

lightningterror:
> We should also work on normalization to reduce the number of different pipelines
> If A == B then set A, B, and C all to zero
Yeah, it was on my mind already.

> Instead of flagging specific entries with another flag, apply this rule: "If C is 128 && D == B then swap D and A". This will rely on our existing handling of the A == B case to also handle D == B cases when B is known to be multiplied by 1.
Yes but I have some other ideas so I might still go with a flag, well one flag instead of 2 like in this pr but changing blend mix 1 to 4 for it.
https://github.com/PCSX2/pcsx2/compare/blend-work12

pcsx2/GS/Renderers/HW/GSRendererNew.cpp
```diff
@@ -567,14 +567,25 @@ void GSRendererNew::EmulateBlending(bool& DATE_PRIMID, bool& DATE_BARRIER)
...
+	const bool blend_mix2_hw_sw = !!(blend_flag & BLEND_MIX2);
+	const bool blend_mix3_hw_sw = !!(blend_flag & BLEND_MIX3);
+	bool blend_mix_hw_sw = (blend_mix1_hw_sw || blend_mix2_hw_sw || blend_mix3_hw_sw);
+
+	// BLEND MIX SW selection, blend can be done in sw
+	const bool blend_mix1_sw = blend_mix1_hw_sw &&
+		((ALPHA.D == 2 && (alpha_c0_zero || alpha_c2_zero)) || (ALPHA.B == ALPHA.D && (alpha_c0_one || alpha_c2_one)));
+	const bool blend_mix2_sw = blend_mix2_hw_sw && (alpha_c0_zero || alpha_c2_zero);
+	const bool blend_mix3_sw = blend_mix3_hw_sw && (alpha_c0_zero || alpha_c2_zero);
+	const bool blend_mix_sw = (blend_mix1_sw || blend_mix2_sw || blend_mix3_sw);
```
tellowkrinkle, line 588: You could just do "if C is known to be zero, set A and B to 0" and it would solve pretty much all of this without needing 15 new flags

lightningterror: Ah right, good idea.

lightningterror:
Benchmark comparison eyecandy:

```
Getaway/Getaway Black Monday
opengl, ultra blending:
pr: 26 fps, 5720 draw calls
master: 13 fps, 13507 draw calls

vulkan, ultra blending:
pr: 49 fps, 5720 draw calls
master: 41 fps, 13507 draw calls
```

SOTC
```
opengl, ultra blending:
pr: 41 fps, 7143 draw calls
master: 20 fps, 17693 draw calls

vulkan, ultra blending:
pr: 91 fps, 7143 draw calls
master: 63 fps, 17693 draw calls
```

Other games are also affected just to a lesser extent but still show an improvement, some big, some small, some no change.

PR #5557, opened 2022-02-21
GS-hw: Don't trigger no overlap check for DATE_BARRIER on d3d11.

lightningterror:
### Description of Changes

GS-hw: Don't trigger no overlap check for DATE_BARRIER on d3d11.
d3d11 doesn't support DATE_BARRIER, it's better to let the Date one cases handle it as it will be more accurate.

### Rationale behind Changes

Accuracy, might have caused regressions that we didn't know about.
Needed to be changed when we added overlap check on dx, on dx we only do the overlap check for blending.

On gl/vk we use the no overlap check and do DATE_BARRIER as it is essentially quite fast, dx doesn't support it.


BinhoKof99:
you never fix the street fighter zero collections or the darkstalkrs collection because??? 🤦‍♂️🤦‍♂️🤦‍♂️

PR #5565, opened 2022-02-23
PCSX2-GUI: Rename Interlacing to Deinterlacing

RedDevilus:
### Description of Changes

Rename the string Interlacing to Deinterlacing for WX and Qt
### Rationale behind Changes

So to sketch what is an annoying issue is that this for years has been wrongly presented to users.
None Interlacing = Interlacing but if used with a no-interlacing patch it will look like progressive but PCSX2 or really the internal PS2 side only detects interlacing mode. Which blows my mind that the user wouldn't be wrong to assume that None Interlacing = Interlacing.
Weave = Interlacing method (saw-tooth)
Bob = Likely the sharpest but can do vertical bouncing.
Blending = A bit blurry but the most consistent and has no bouncing.

I didn't change the code just the visual presentation to the users.

Example in gif format: 
Interlaced_Animation

### DIFF
--- a/pcsx2-qt/Settings/GraphicsSettingsWidget.ui
+++ b/pcsx2-qt/Settings/GraphicsSettingsWidget.ui
@@ -133,7 +133,7 @@
        <item row="3" column="0">
         <widget class="QLabel" name="label_23">
          <property name="text">
-          <string>Interlacing:</string>
+          <string>Deinterlacing:</string>
          </property>
         </widget>
        </item>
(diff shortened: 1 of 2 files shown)


RedDevilus:
Example in WX:
image
Example in Qt:
image

fallaha56:
Sorry if this is an obvious question, but is there a reason that the GS output cannot simply be fed to the deinterlacing units of the graphics cards?

RedDevilus:
@fallaha56 I'm not a GS developer, I only know that PS2 emulation is a pain and that everything there is for a reason. PS2 does not have a GPU but it does have a rasterizer and that game developers did their own thing.

PR #5571, opened 2022-02-26
EE/MicroVU: Elide COP2 flag updates where safe-ish

stenzek:
### Description of Changes

Denormalizing the VU0 status flag before and renormalizing after the instruction adds a decent chunk of code bloat to the dynarec blocks, as well as having a measurable performance impact.

This PR adds a new "optimization" option which basically extends the microVU Flag Hack to COP2 instructions as well, that is, skipping computing flags as well as normalizing/denormalizing when possible.

It's not 100% "safe", as the sticky bits are only updated by the final/last instruction. However, as the mVU flag hack has shown, it's not something games tend to be sensitive to. Except Tekken Tag, apparently, which still works with this option (plus about a 5% performance boost).

We'll probably combine it with the MicroVU Flag Hack, as that makes sense, but for testing purposes it's still a separate option (for now).

The second commit eliminates the micro flag updates from COP2 instructions entirely, moving them to program start time. Since COP2 instructions will force the microprogram to finish (and thus normalize the flags), and starting another micro will wait for the COP2 instruction to finish, as I understand it this should be safe.

Performance gains can be anywhere from 3-10%, depending on the game.

### Rationale behind Changes

Making VU0/COP2 heavy games suck slightly less.

### Suggested Testing Steps

Thoroughly test VU0/COP2 heavy games. I've checked FFX, Tekken Tag and Ratchet briefly.

### DIFF
--- a/pcsx2/x86/microVU_Macro.inl
+++ b/pcsx2/x86/microVU_Macro.inl
@@ -102,16 +102,6 @@ void endMacroOp(int mode)
 
 	if (mode & 0x10)
 	{
-		if (!CHECK_VU_FLAGHACK || (g_pCurInstInfo->info & EEINST_COP2_STATUS_FLAG)) // Status/Mac Flags were Updated
-		{
-			// update micro_statusflags
-			const int t0reg = _allocTempXMMreg(XMMT_INT, -1);
-			xMOVDZX(xRegisterSSE(t0reg), gprF0);
-			xSHUF.PS(xRegisterSSE(t0reg), xRegisterSSE(t0reg), 0);
-			xMOVAPS(ptr128[&microVU0.regs().micro_statusflags], xRegisterSSE(t0reg));
-			_freeXMMreg(t0reg);
-		}
-
 		if (!CHECK_VU_FLAGHACK || g_pCurInstInfo->info & EEINST_COP2_NORMALIZE_STATUS_FLAG)
 		{
 			// Normalize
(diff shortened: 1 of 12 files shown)


refractionpcsx2:
Most games seem to work great, but the BIOS seems to be an edgecase which is not happy :(

pcsx2/x86/microVU_Macro.inl
```diff
@@ -82,29 +96,26 @@ void endMacroOp(int mode)
...
+		{
+			// Normalize
+			mVUallocSFLAGc(eax, gprF0, 0);
+			xMOV(ptr32[&vu0Regs.VI[REG_STATUS_FLAG].UL], eax);
+		}
+		else
+		{
+			// backup denormalized flags for the next instruction
+			// this is fine, because we'll normalize them again before this reg is accessed
+			xMOV(gprF0, ptr32[&vuRegs->VI[REG_STATUS_FLAG].UL]);
```
refractionpcsx2, line 115: Operands are backwards (this is what breaks the BIOS) but correcting it re-breaks Tekken Tag

refractionpcsx2, line 115: Operands are backwards (this is what breaks the BIOS) but correcting it re-breaks Tekken Tag

pcsx2/x86/microVU_Macro.inl
```diff
@@ -55,24 +55,38 @@ void setupMacroOp(int mode, const char* opName)
...
-		//Denormalize
-		mVUallocSFLAGd(&vu0Regs.VI[REG_STATUS_FLAG].UL);
-		
-		xMOV(gprF0, eax);
+	}
+	if (mode & 0x10)
+	{
+		if (!CHECK_COP2_FLAGHACK || (g_pCurInstInfo->info & EEINST_COP2_DENORMALIZE_STATUS_FLAG))
+		{
+			_freeX86reg(ebx);
```
refractionpcsx2, line 79: should this not be outside the if? since ebx == gprF0 and the other case is also modifying it

pcsx2/x86/microVU_Macro.inl
refractionpcsx2: seems to want status flags to be written on line 63
if (mode & 0x10 && (!CHECK_COP2_FLAGHACK || g_pCurInstInfo->info & EEINST_COP2_STATUS_FLAG)) // Update Status Flag

removing the EEINST_COP2_STATUS_FLAG check fixes it.

refractionpcsx2: Just an additional, I set it to always update those flags on the instructions that want it, and the performance was the same on x86 (maybe margin of error better), is it a huge problem on ARM?

refractionpcsx2: I had a ponder about this when I went for a nap. I guess you *could* do what you are doing now, but you'd have to scan forward on the block, see if it reads the status register, followed by an AND with a value including any bits of 0xFC0, if that's true, enable status/mac updates for that entire block, that will probably be more than enough for it.

MrCK1:
MDK2:
COP2 Hack Disabled: (~210FPS):
image



COP2 Hack Enabled (~238FPS):
image

JordanTheToaster (approved):
Ran through a fair few of my games and didn't notice any issues FPS seems to up across the board from very small to 10 fps in R&C 3.

PR #5572, opened 2022-02-27
VU: Run sync ahead on small blocks & Rework VU Kickstart

refractionpcsx2:
### Description of Changes
Attempt to run the VU (VU0 mainly) ahead if it's only running a small block, to try and reduce overhead.
VU Kickstart is now always on and is replaced with a VU Sync option as less games need the sync than the run ahead.

### Rationale behind Changes
Running blocks of like 3 cycles is very inefficient when you factor in entering and exiting a recompiler, so this makes sure at least a more reasonable amount of cycles is run.

### Suggested Testing Steps
test games, compare to master.

Some preliminary numbers:
Tekken Tag: 80-82 to 82-84
Ratchet & Clank: 102-105 to 110-113
Mike Tyson Heavyweight Boxing: 100 to 102
Ratchet & Clank 3: 75 to 82
Jak 3: 118 to 120

Mixed with #5571 (which has now been rebased in) Ratchet & Clank has gone from 105fps to 122fps at the beginning of the first level, a nice little boost, so these two PR's bring a combined boost of somewhere between 10-20% in a bunch of games.

NOTE: This will NOT affect all games, you have to be EE thread limited and the game has to be VU0 heavy (which isn't a great deal) for it to have much effect, so in most scenarios you probably won't even notice a difference.

### DIFF
--- a/pcsx2/VUmicro.cpp
+++ b/pcsx2/VUmicro.cpp
@@ -20,6 +20,19 @@
 #include "GS.h"
 #include "Gif_Unit.h"
 
+__inline u32 CalculateMinRunCycles(u32 cycles, bool requiresAccurateCycles)
+{
+	// If we're running an interlocked COP2 operation
+	// run for an exact amount of cycles
+	if(requiresAccurateCycles)
+		return cycles;
+
+	// Allow a minimum of 16 cycles to avoid running small blocks
+	// Running a block of like 3 cycles is highly inefficient
+	// so while sync isn't tight, it's okay to run ahead a little bit.
+	return std::max(16U, cycles);
+}
+
 // Executes a Block based on EE delta time
 void BaseVUmicroCPU::ExecuteBlock(bool startUp)
 {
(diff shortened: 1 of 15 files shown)


pcsx2/VUmicro.cpp
```diff
@@ -20,6 +20,19 @@
 #include "GS.h"
 #include "Gif_Unit.h"
 
+__inline u32 CalculateMinRunCycles(u32 cycles)
+{
+	// If we're running an interlocked COP2 operation
+	// run for an exact amount of cycles
+	if((cpuRegs.code & 0xFC000001) == 0x48000001)
```
stenzek, line 27: Is `code` going to be correct in all cases here? Iirc VCALLMS is a fallback but are there pathways here from other instructions which won't be? Might need a global flag for "in interlocked op"..

refractionpcsx2: We always use cpuRegs.code to check if it's interlocking if you check in microVU_Macro.inl and I crafted it specifically to ignore the "V" instructions because for those to run it has to finish the VU program that's running, plus it avoids false positives, so this cos is kind of irrelevant and doesn't matter if it runs a bit longer :)

stenzek: Right, but this routine gets called when the block is executing, not compiling right?

refractionpcsx2: tits, uhh you make a good point...

refractionpcsx2: Alright, I've changed it up so if the EE block uses interlocking, the rest of the sync points on the block will also use accurate cycles, so at runtime it should be correct. I always mix up compile time and run time lol

refractionpcsx2:
Blown the scope of this PR out of the water and reworked VU Kickstart to always be on and added a new VUSync option for the games which don't like it, since there's only a handful of games, instead of 222 entries for kickstart.

JordanTheToaster (approved):

PR #5574, opened 2022-02-28
GS: Texture replacement fixes

stenzek:
### Description of Changes

**[GS/TextureReplacements: Disable generated mipmaps on compressed textures]**

This was causing errors in D3D when replacements were loaded, since it was trying to enable/generate mipmaps on compressed formats, which aren't render targets.

**[GS/TextureCache: Don't count replacements in hash cache budget]**

Large replacement textures were causing the hash cache to disable itself. There's plenty of other massive VRAM sinks in PCSX2 that we need to consider before we can say we're managing VRAM well anyway.

### Rationale behind Changes

Fixing my bugs.

### Suggested Testing Steps

Don't really need any, since it's an experimental feature anyway, but I've briefly tested and replacements are still fine.

### DIFF
--- a/pcsx2/GS/Renderers/Common/GSTexture.cpp
+++ b/pcsx2/GS/Renderers/Common/GSTexture.cpp
@@ -131,7 +131,7 @@ u32 GSTexture::CalcUploadSize(u32 height, u32 pitch) const
 
 void GSTexture::GenerateMipmapsIfNeeded()
 {
-	if (!m_needs_mipmaps_generated || m_mipmap_levels <= 1)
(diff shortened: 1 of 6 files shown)


CelestePolo:
I tried v1.7.2402 and this PR version on SotC. The screen goes briefly black regularly. When this happen Vulkan reports this:
" Vulkan: Executing command buffer due to 'While waiting for 1048576 bytes in texture upload buffer' "
 
There is also this error on the left hand corner:
image

PR5547 was exempt of this problem.

PR #5577, opened 2022-03-01
Add Metal renderer

tellowkrinkle:
### Description of Changes
Adds Metal renderer

### Rationale behind Changes
Better Mac support

### Suggested Testing Steps
Make sure I didn't break anything outside of Metal

### DIFF
--- a/pcsx2/Pcsx2Config.cpp
+++ b/pcsx2/Pcsx2Config.cpp
@@ -90,7 +90,7 @@ void Pcsx2Config::SpeedhackOptions::Set(SpeedhackId id, bool enabled)
 		case Speedhack_MTVU:
 			vuThread = enabled;
 			break;
-			jNO_DEFAULT;
+        jNO_DEFAULT;
 	}
 }
 
@@ -269,20 +269,13 @@ const char* Pcsx2Config::GSOptions::GetRendererName(GSRendererType type)
 {
 	switch (type)
 	{
-		case GSRendererType::Auto:
-			return "Auto";
-		case GSRendererType::DX11:
-			return "Direct3D 11";
-		case GSRendererType::OGL:
-			return "OpenGL";
-		case GSRendererType::VK:
-			return "Vulkan";
-		case GSRendererType::SW:
-			return "Software";
-		case GSRendererType::Null:
-			return "Null";
-		default:
-			return "";
+		case GSRendererType::Auto:  return "Auto";
+		case GSRendererType::DX11:  return "Direct3D 11";
+		case GSRendererType::OGL:   return "OpenGL";
+		case GSRendererType::VK:    return "Vulkan";
+		case GSRendererType::SW:    return "Software";
+		case GSRendererType::Null:  return "Null";
+		default:                    return "";
 	}
 }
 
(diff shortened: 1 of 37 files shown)


pcsx2/Pcsx2Config.cpp
```diff
@@ -264,20 +264,14 @@ const char* Pcsx2Config::GSOptions::GetRendererName(GSRendererType type)
...
-		default:
-			return "";
+	case GSRendererType::Auto: return "Auto";
+	case GSRendererType::DX11: return "Direct3D 11";
+	case GSRendererType::Metal: return "Metal";
+	case GSRendererType::OGL: return "OpenGL";
+	case GSRendererType::VK: return "Vulkan";
+	case GSRendererType::SW: return "Software";
+	case GSRendererType::Null: return "Null";
+	default: return "";
```
TheLastRar, line 274: Might be worthwhile adding `// clang-format off` and `// clang-format on` here if you don't like what clang format does

tellowkrinkle: With the frequency of these single-item switch statements, I feel like having to add the clang format comments is kind of ridiculous

If you use `git clang-format` (which you should), this only becomes an issue if you actually edit the switch statement, since the braces have the right indentation which should protect it from any outside changes bleeding in

I should probably format that file nicely though, instead of just reverting the clang-format change and leaving it somewhere in the middle

Warepire:
Question because I don't know enough about Macs or the GS, so this PR made me curious:

I see other projects (like dolphin) that support Mac only via MoltenVK for example. Is there a benefit here to use Metal directly over something like that in PCSX2?

tellowkrinkle:
> I see other projects (like dolphin) that support Mac only via MoltenVK for example. Is there a benefit here to use Metal directly over something like that in PCSX2?

A few differences
- Nicer traces in Xcode.  It's nice when the commands you issue in code match the commands you see in captured GPU traces.
- Full access to Metal features.  Framebuffer fetch is still missing from Vulkan, though you could run a modified MVK like Dolphin does.
- Reduced CPU usage.  MoltenVK adds a second thread to encode GPU commands, which could affect dual-core Macs (which find themselves very heavily loaded by PCSX2).
- Better performance when CPU limited (AMD).  Surprisingly, encoding Vulkan commands into MVK is slower than encoding Metal commands into AMD's Metal driver.  On my laptop, Jak title screen GSdump at 1x (for maximum CPU bottleneck) is 150fps with 110% CPU usage in Metal, 140fps with 150% CPU usage in MVK.  On the other hand, the Intel GPU is 115fps on both, 110% Metal vs 160% Vulkan
- Better performance when GPU limited.  I don't know why, the commands are all going through Metal either way, but I'd estimate on average 10% better performance in Metal when GPU limited

Plus some weirdnesses that I have yet to look into properly:
- MVK performance on a 2014 MBP with an Nvidia 750M is terrible (like 25fps Vk vs 76fps Metal on a Sly GSdump).  Iris Pro on that laptop is closer, but still not great (60fps vs 90fps).  Seems to be a CPU thing, they get closer as you increase the resolution.

Warepire:
I see, thank you for explaining.

tellowkrinkle:
Ahh one more
It's much easier to realize that a performance issue exists when you have another renderer that should be the same speed but is instead performing much faster (and also easier to track down the cause).  I've gotten a number of speedups in the Metal renderer by comparing it to MVK, and 793ba944d634406b9d09c0f4f8e7d769b1b449a0 was found because MVK was performing 10x slower than Metal in a game.

tellowkrinkle:
Now requires #5630

refractionpcsx2:
#5630 has been merged

PR #5582, opened 2022-03-01
GS/TextureCache: Don't let hash cache memusage go negative

### LINKED ISSUE #5578: [Texture Replacement]: Hash cache has used (xxx) MB of VRAM, disabling.
### Describe the Bug

a huge amount of Texture replacement may cause a crazy amount of preload into Vram, the emulator would just spam this messages if i had the OSD "Show Messages" enabled:
hashcache
also trying to emulate the game in Higher Resolution than 5x native (like 6x native 4K) would just make it crash.

the game in Question are Mobile Suit Gundam: Gundam vs. Zeta Gundam (SLUS-20821) and here some screenshot of the emulator running with replaced textures (it work fine, with exception of Loading Screen which for some reason kinda glitched if its replaced and hashcache messages spam)
remaster
gs_20220301094754

### Reproduction Steps

have 1638 texture replacement (DDS BC5 format ,mostly 1024x1024 for half of it and the rest are 512x512 or 256x256)
Texture Preloading: None/Partial/Full
Renderer : OpenGL/Vulkan
Advanced: Asynch Texture Loading + Load Textures + Precache Texture (disabling precache wouldn't make the emulator replace the texture at all)
Internal Resolution Tested: Native 3X,4X,5X (6X would just make it crash)

### Expected Behavior

dynamically load only needed texture or texture replacement working without precache texture

### PCSX2 Revision

v.1.7.2401-64-bits AVX2 build

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

stenzek:
Regression fix for https://github.com/PCSX2/pcsx2/commit/ba27a46ac617bb170eb5524f2f58cc018e720a86.

Should fix #5578, but I have now way of testing it with larger packs and don't feel like creating one.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
+++ b/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
@@ -1138,7 +1138,9 @@ void GSTextureCache::IncAge()
 		HashCacheEntry& e = it->second;
 		if (e.refcount == 0 && ++e.age > max_hash_cache_age)
 		{
-			m_hash_cache_memory_usage -= e.texture->GetMemUsage();
+			if (!e.is_replacement)
+				m_hash_cache_memory_usage -= e.texture->GetMemUsage();
+
 			g_gs_device->Recycle(e.texture);
 			m_hash_cache.erase(it++);
 		}


DonelBueno:
Confirmed fixed.

danny199012:
I Can also test this out as this was a major issue with png causing graphical issues and major slowdowns for me so this should hopefulyy fix it all

PR #5587, opened 2022-03-01
WX: Make texture dump location configurable

### LINKED ISSUE #5586: [Feature Request]: Allow change in dumping and load directory for GS Texture Replacement
### Description

Allow users to define the directory to dump textures to within the PCSX2 GUI or as an INI setting.

### Reason

Currently dumping and loading textures must be from the textures directory; For emulators kept on an SSD this causes an excessive amounts of writes to the drive lowering the life span.

### Examples

F:\texture\dump
/mnt/some_drive/textures/dump

tellowkrinkle:
### Description of Changes
Makes texture dump location configurable

Closes #5586

### Rationale behind Changes
Texture dumps are big and you might not want them on your main volume

### Suggested Testing Steps
Make sure you can change the texture dump location in settings and have it actually affect where textures are dumped

### DIFF
--- a/pcsx2/gui/AppConfig.cpp
+++ b/pcsx2/gui/AppConfig.cpp
@@ -536,8 +536,8 @@ void AppConfig::FolderOptions::Set(FoldersEnum_t folderidx, const wxString& src,
 			break;
 
 		case FolderId_Textures:
-			Cache = src;
-			UseDefaultCache = useDefault;
+			Textures = src;
+			UseDefaultTextures = useDefault;
 			EmuFolders::Textures = GetResolvedFolder(FolderId_Textures);
 			EmuFolders::Textures.Mkdir();
 			break;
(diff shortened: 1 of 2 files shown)


SirFancyBacon:
Works like a charm!

stenzek:
My copy/paste fail in `AppConfig::FolderOptions::Set()` is probably why people were reporting the textures directory not getting automatically created. I only ever tested it in Qt..

refractionpcsx2:
Might be a good idea to fix that in this PR also then, so we can kill 2 birds with 1 stone (since they're kind of related)

stenzek:
> Might be a good idea to fix that in this PR also then, so we can kill 2 birds with 1 stone (since they're kind of related), or is that already fixed? I see there is some changes in there.

Already in there.

PR #5598, opened 2022-03-04
Qt: Assorted cleanups/add debug log options

stenzek:
### Description of Changes

Fixes warnings, gets rid of unnecessary ifdefs since we require Qt6 on all platforms. Also makes the log accessible in release builds.

### Suggested Testing Steps

None needed, I've already tested, and it's mostly cosmetic.

### DIFF
--- a/pcsx2-qt/DisplayWidget.cpp
+++ b/pcsx2-qt/DisplayWidget.cpp
@@ -49,12 +49,7 @@ DisplayWidget::~DisplayWidget() = default;
 
 qreal DisplayWidget::devicePixelRatioFromScreen() const
 {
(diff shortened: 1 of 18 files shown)


Zorklis:
Final steps towards full QT release?

WilliamVaz:
Naruto Uzumaki Chronicles 2 is freezing when choosing the "Gathering Intelligence" quest.
This did not happen in previous versions of the emulator.

refractionpcsx2:
> 

That has ABSOLUTELY nothing to do with this PR. If you have a problem, make a new issue and fill out the issue form.

PR #5601, opened 2022-03-05
GS: Support local to local transfers that overwrite themselves

### LINKED ISSUE #3118: Gitaroo Man (NTSC-U) [SLUS-20294] - Blank/disappearing face in Stage 8
***PCSX2 version:***
v1.5.0-dev-3290-gd92e4343e 	(i.e. 2019-09-22 03:00:52 from the buildbot)

Gitaroo Man (NTSC-U) SLUS-20294 blank face bug: the opponent's face disappears at a certain moment in Stage 8 (see How to reproduce).

screenshot blank face

***PCSX2 options:***
Default settings with the Safest preset.

In other tests (which all resulted in the same visual bug or just made him explode into polygons), I disabled Automatic game fixes and tried various Clamping/Rounding settings:
- all combinations of the EE/IOP Round/Clamp choices, with VU Round/Clamp set to their Safest defaults.
- all combinations of the VU Round/Clamp choices, with EE/IOP Round/Clamp set to their Safest defaults.

***Plugins used:***
- GSdx SSE2 with Direct3D 11 (Software) renderer
Everything else is default plugins/default settings.

GS dump: [gsdump blank face.zip](https://github.com/PCSX2/pcsx2/files/3640041/gsdump.blank.face.zip)

Load a complete save game from [Gitaroo Man (NTSC-U) memcard.zip](https://github.com/PCSX2/pcsx2/files/3640043/Gitaroo.Man.NTSC-U.memcard.zip) through the in-game Options menu. Go to Single Play -> Stage 8. Play the stage until the opponent says "Beauty is a sin! Oh God!" and as he throws his arms back dramatically, his face will disappear.

(If you'd rather not actually play the stage, you put `patch=1,EE,0017820c,word,10000026` in 7130C553.pnach to keep from losing, then just wait until it gets to that part.)

(Despite the SVN numbers, these are three git commits in a row, with no commits in between them.)
- SVN 4282, i.e. https://github.com/PCSX2/pcsx2/commit/e2d36a53a4e3396ade8d008186c1ee2069903d49: last known version to display his face properly.
- SVN 4287, i.e. https://github.com/PCSX2/pcsx2/commit/ca7abd983aed6762efa41d0c51b4bbbfb84134b5: probably introduced the problem, but I couldn't build it until the next commit's VS2008 fix.
- SVN 4288, i.e. https://github.com/PCSX2/pcsx2/commit/77f392edac1806fcba929dd72b7775dd850f8d6c: the first version I was able to build that has the blank face bug.

***PC specifications:***
- CPU: Intel Core i5-2450M
- GPU: Intel HD Graphics 3000
- OS: Windows 10 Pro (64 Bit, Version 1903, build 18362.356)

For what it's worth, I've also seen this problem on recent Linux PPA builds (in Ubuntu 18.04, running on the same hardware with GSdx's OpenGL software renderer). I don't know if the git commits above apply to the Linux version.

***Additional info:***
In Gitaroo Man, it seems face animations work by copying one region of a texture and pasting over to another region. (There may be a little more to it than that.) One particular face is causing this bug, due to the relationship between the copied and pasted coordinates.

In my attempt to figure out why the problem happens, I hacked the coordinates of the copied texture region to make the animation below. The left side shows the visual bug in PCSX2, and the right side is how it should look if the bug were fixed. Below that it shows the texture: the red box is the copied region, and the green box indicates where it's pasted to.

face demo

As the gif shows, the problem happens when the copied region extends off the left edge of the pasted region, and the problem disappears the moment the left edge of the copied region meets the left edge of the pasted region.

(Unrelated to the gif above) If it would be helpful, I can provide a savestate in which I've tracked down the exact value in memory responsible for the problem just before it becomes visible in normal gameplay. (Unfortunately, when I set a read breakpoint on this value, the debugger acts like the game never reads it back before overwriting it, even though there are obvious changes in-game when I edit that value.)

refractionpcsx2:
### Description of Changes
Adds support for local to local GS transfers that use the same source/destination address

### Rationale behind Changes
Some games (In this case Gitarooman will do local to local transfers to the same address but with the source only a few pixels behind the destination, causing the source to get overwritten by the destination (it will repeat the first few pixels), this is an undesired effect, so we need to do the copy backwards. On real hardware the page would likely be buffered when the read starts, so it will have a clean copy before it gets overridden, we don't have that luxury in the current code (also it'd be slower)

### Suggested Testing Steps
Check Gitarooman stage 8, not sure if any other games do this.

Fixes #3118

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -1702,14 +1702,59 @@ void GSState::Move()
 		int _sy = sy, _dy = dy; // Faster with local copied variables, compiler optimizations are dumb
 		if (xinc > 0)
 		{
-			for (int y = 0; y < h; y++, _sy += yinc, _dy += yinc)
+			const int page_width = GSLocalMemory::m_psm[m_env.BITBLTBUF.DPSM].pgs.x;
+			const int page_height = GSLocalMemory::m_psm[m_env.BITBLTBUF.DPSM].pgs.y;
+			const int xpage = sx & ~(page_width - 1);
+			const int ypage = _sy & ~(page_height - 1);
+			// Copying from itself to itself (rotating textures) used in Gitaroo Man stage 8
+			// What probably happens is because the copy is buffered, the source stays just ahead of the destination.
+			if (sbp == dbp && (((_sy < _dy) && ((ypage + page_height) > _dy)) || ((sx < dx) && ((xpage + page_width) > dx))))
 			{
-				auto s = getPAHelper(spo, sx, _sy);
-				auto d = getPAHelper(dpo, dx, _dy);
+				int starty = 0;
+				int endy = h;
+				int y_inc = yinc;
 
-				for (int x = 0; x < w; x++)
+				if (((_sy < _dy) && ((ypage + page_height) > _dy)))
 				{
-					pxCopyFn(d, s, x);
+					_sy += h;
+					_dy += h;
+					starty = h-1;
+					endy = -1;
+					y_inc = -y_inc;
+				}
+			
+				for (int y = starty; y != endy; y+= y_inc, _sy += y_inc, _dy += y_inc)
+				{
+					auto s = getPAHelper(spo, sx, _sy);
+					auto d = getPAHelper(dpo, dx, _dy);
+
+					if (((sx < dx) && ((xpage + page_width) > dx)))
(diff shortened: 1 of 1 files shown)


kenshen112:
Would this have an affect on Fatal Frame?

PR #5605, opened 2022-03-07
GUI: Log graphics driver info on startup

stenzek:
### Description of Changes

Logs graphics driver info to the console when starting up for the first time. Also stops the Vulkan renderer preventing initialization when the debug layers are unavailable but the option is enabled.

### Rationale behind Changes

Useful when dealing with people with driver bugs.

### Suggested Testing Steps

None needed, it's just logging.


lightningterror:
> Also stops the Vulkan renderer preventing initialization when the debug layers are unavailable but the option is enabled.

I needed this.

PR #5607, opened 2022-03-07
Qt: Fix live changing game settings not applying

stenzek:
### Description of Changes

What the title says. Game settings weren't updating after being changed in the UI because it wasn't reloading the VM side ini interface.

Also fixes incorrect values getting written to the ini which don't match the UI for some options (e.g. resolution scale).

### Rationale behind Changes

This was pretty annoying, and a bug from not properly testing on my part.

### Suggested Testing Steps

It's Qt, and I've already tested it.


andyten:
bug?
ps1

ps2

PR #5614, opened 2022-03-07
ps1 mode cd logic adjustments

kenshen112:
### Description of Changes

Fixes set status on seek and read. Added missing implementation for CdlStat and CdlMotor.
Fixes to CdlPlay and ReadS

### Rationale behind Changes

More accurate to ps1 behavior.

### Suggested Testing Steps

Make sure ps1 games aren't borked

### DIFF
--- a/pcsx2/CDVD/CdRom.cpp
+++ b/pcsx2/CDVD/CdRom.cpp
@@ -41,7 +41,7 @@ enum cdrom_registers
 	CdlDemute = 12,
 	CdlSetfilter = 13,
 	CdlSetmode = 14,
-	CdlGetmode = 15,
+	CdlGetparam = 15,
 	CdlGetlocL = 16,
 	CdlGetlocP = 17,
 	Cdl18 = 18,
@@ -54,7 +54,6 @@ enum cdrom_registers
 	CdlReadS = 27,
 	CdlReset = 28,
 	CdlReadToc = 30,
-
 	AUTOPAUSE = 249,
 	READ_ACK = 250,
 	READ = 251,
@@ -68,7 +67,7 @@ const char* CmdName[0x100] = {
 	"CdlSync", "CdlNop", "CdlSetloc", "CdlPlay",
 	"CdlForward", "CdlBackward", "CdlReadN", "CdlStandby",
 	"CdlStop", "CdlPause", "CdlInit", "CdlMute",
-	"CdlDemute", "CdlSetfilter", "CdlSetmode", "CdlGetmode",
+	"CdlDemute", "CdlSetfilter", "CdlSetmode", "CdlGetparam",
 	"CdlGetlocL", "CdlGetlocP", "Cdl18", "CdlGetTN",
 	"CdlGetTD", "CdlSeekL", "CdlSeekP", NULL,
 	NULL, "CdlTest", "CdlID", "CdlReadS",
@@ -93,6 +92,7 @@ u8 Test23[] = {0x43, 0x58, 0x44, 0x32, 0x39, 0x34, 0x30, 0x51};
 #define DiskError 5
 
 /* Modes flags */
+#define MODE_INIT (0 << 0)      // Init sets mode 00h or not all bits cleared
 #define MODE_SPEED (1 << 7)     // 0x80
 #define MODE_STRSND (1 << 6)    // 0x40 ADPCM on/off
 #define MODE_SIZE_2340 (1 << 5) // 0x20
@@ -108,8 +108,8 @@ u8 Test23[] = {0x43, 0x58, 0x44, 0x32, 0x39, 0x34, 0x30, 0x51};
 #define STATUS_SEEK (1 << 6)      // 0x40
 #define STATUS_READ (1 << 5)      // 0x20
 #define STATUS_SHELLOPEN (1 << 4) // 0x10
-#define STATUS_UNKNOWN3 (1 << 3)  // 0x08
-#define STATUS_UNKNOWN2 (1 << 2)  // 0x04
+#define STATUS_IDERROR (1 << 3)  // 0x08
+#define STATUS_SEEKERROR (1 << 2)  // 0x04
 #define STATUS_ROTATING (1 << 1)  // 0x02
 #define STATUS_ERROR (1 << 0)     // 0x01
 
@@ -134,6 +134,8 @@ static void AddIrqQueue(u8 irq, u32 ecycle);
 static __fi void StartReading(u32 type)
 {
 	cdr.Reading = type;
+	// Read's retry. If there's a status error clear and try, try again
+	cdr.StatP &~ STATUS_ERROR;
 	cdr.FirstSector = 1;
 	cdr.Readed = 0xff;
 	//DevCon.Warning("ReadN/ReadS delay: %d", sectorSeekReadDelay);
@@ -217,6 +219,7 @@ void cdrInterrupt()
 
 		case CdlNop:
 			SetResultSize(1);
+			cdr.StatP &= ~STATUS_SHELLOPEN;
 			cdr.Result[0] = cdr.StatP;
 			cdr.Stat = Acknowledge;
 			break;
@@ -256,7 +259,14 @@ void cdrInterrupt()
 		case CdlStandby:
 			cdr.CmdProcess = 0;
 			SetResultSize(1);
-			cdr.StatP |= STATUS_ROTATING;
+			if (cdr.StatP & STATUS_ROTATING)
+			{
+				cdr.StatP |= ERROR_INVALIDARG;
+			}
+			else
+			{
+				cdr.StatP |= STATUS_ROTATING;
+			}
 			cdr.Result[0] = cdr.StatP;
 			cdr.Stat = Complete;
 			break;
@@ -288,6 +298,7 @@ void cdrInterrupt()
 		case CdlInit:
 			SetResultSize(1);
 			cdr.StatP = STATUS_ROTATING;
+			cdr.Mode |= MODE_INIT;
 			cdr.Result[0] = cdr.StatP;
 			cdr.Stat = Acknowledge;
 			AddIrqQueue(CdlInit + 0x20, 0x800);
@@ -328,7 +339,7 @@ void cdrInterrupt()
 			cdr.Stat = Acknowledge;
 			break;
 
-		case CdlGetmode:
+		case CdlGetparam:
 			SetResultSize(6);
 			cdr.StatP |= STATUS_ROTATING;
 			cdr.Result[0] = cdr.StatP;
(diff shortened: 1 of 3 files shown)


F0bes:
Can you explain how you found this stuff out?
Cite sources, hardware tests, etc?

kenshen112:
No$ docs mostly

RedDevilus:
Can you link the relevant lines of no$ and what are the other parts that aren't included in the 'mostly' section?

kenshen112:
Sure. Give me a bit

kenshen112:
https://problemkaputt.de/psx-spx.htm#cdromstatuscommands the changes I made to the status defines like the newly added ones STATERROR you can see listed here.

Read the bottom of the GetStat section for the reason I changed cdlNop. 

https://problemkaputt.de/psx-spx.htm#cdromcdaudiocommands the comment I added to cdlPlay and the reason I added the readSouth read command in there and removed the irq call can be found in the play section near the bottom play is nearly identical

https://problemkaputt.de/psx-spx.htm#cdromcontrolcommands look at MotorOn and see the "Myth" to understand why I removed cdlStandby

kenshen112:
As for the submodule. My environment.keot trying to push a change. I tried to fight it back and it slipped through. I'll fix that in a sec

F0bes:
> https://problemkaputt.de/psx-spx.htm#cdromstatuscommands the changes I made to the status defines like the newly added ones STATERROR you can see listed here.
> 
> Read the bottom of the GetStat section for the reason I changed cdlNop. 
> 
> https://problemkaputt.de/psx-spx.htm#cdromcdaudiocommands the comment I added to cdlPlay and the reason I added the readSouth read command in there and removed the irq call can be found in the play section near the bottom play is nearly identical
> 
> https://problemkaputt.de/psx-spx.htm#cdromcontrolcommands look at MotorOn and see the "Myth" to understand why I removed cdlStandby
> 

Appreciated.

kenshen112:
> > https://problemkaputt.de/psx-spx.htm#cdromstatuscommands the changes I made to the status defines like the newly added ones STATERROR you can see listed here.
> > 
> > Read the bottom of the GetStat section for the reason I changed cdlNop. 
> > 
> > https://problemkaputt.de/psx-spx.htm#cdromcdaudiocommands the comment I added to cdlPlay and the reason I added the readSouth read command in there and removed the irq call can be found in the play section near the bottom play is nearly identical
> > 
> > https://problemkaputt.de/psx-spx.htm#cdromcontrolcommands look at MotorOn and see the "Myth" to understand why I removed cdlStandby
> > 
> 
> Appreciated.

There's probably other mistakes like that in our code currently that I'm just missing. Maybe even outside of cdrom

kenshen112:
So far Symphony of the night's a casualty

kenshen112:
The DMA3 not ready stuff seem related to ReadInterrupt being off

pcsx2/CDVD/CdRom.cpp
```diff
@@ -751,20 +765,18 @@ void cdrWrite1(u8 rt)
...
-			if (cdr.Reading)
-			{
-				StopReading();
-			}
 			if (cdr.SetlocPending)
 			{
 				memcpy(cdr.SetSectorSeek, cdr.SetSector, 4);
 				cdr.SetlocPending = 0;
 			}
+			// Need to figure out if data track or audio. Data tracks are treated as "00"
```
stenzek, line 773: Incorrect. Play with `track == 0` starts playing CDDA from the current hold position. The mechacon doesn't care whether it's data or audio for the purposes of the command.

pcsx2/CDVD/CdRom.cpp
```diff
@@ -926,10 +939,10 @@ void cdrWrite1(u8 rt)
 			break;
 
 		case CdlReadS:
-			if (cdvd.Type == CDVD_TYPE_PSCDDA) // Taken from pcsxr
-				goto do_CdlPlay;
 			cdr.Irq = 0;
 			StopReading();
+			if (cdvd.Type == CDVD_TYPE_PSCDDA) // Taken from pcsxr
```
stenzek, line 944: This is likely wrong. Again, the command doesn't care about the media/track type, how it gets delivered is (iirc) based on the subq of the sector itself.

pcsx2/CDVD/CdRom.cpp
```diff
@@ -848,22 +858,25 @@ void cdrWrite1(u8 rt)
...
 			if (cdr.Mode & MODE_CDDA)
 			{
 				StopCdda();
-				cdvd.Type = CDVD_TYPE_CDDA;
+				cdvd.Type = CDVD_TYPE_PSCDDA;
 			}
 
-			cdvd.Speed = 1 + ((cdr.Mode >> 7) & 0x1);
-			setPs1CDVDSpeed(cdvd.Speed);
+			if (cdr.Mode & MODE_SPEED && cdr.OldSpeed != cdr.Speed)
```
stenzek, line 872: This seems pointless. From a quick glance, `setPs1CDVDSpeed()` will have no side effects if the speed does not change. Also, this expression as is will break when switching from double back to single speed (very common, e.g. for CDDA).

From my hardware tests on a PS1 console, setting the mode starts changing the speed immediately, and it takes approximately 0.8 seconds to stabilize at the new speed from 1x->2x, or ~1 second from 2x->1x. Ideally, this should be taken into consideration when seeking/reading after a speed change.

stenzek (changes requested):

stenzek:
Also shouldn't rely on no$ docs exclusively, we changed/fixed several parts of it at https://psx-spx.consoledev.net/cdromdrive.

kenshen112:
@stenzek Right. I'll fix all that then. Do you have any clue where the DMA3 Not ready errors might be coming from? Far as I can tell, something's off timing-wise with ReadInterrupt which causes DMA3 to not have data. The old hack would force the disk to read again till it will be full afaik.

pcsx2/CDVD/CdRom.cpp
kenshen112: Yeeahhh. It was an old hack from pcsxr, I'll remove it

pcsx2/CDVD/CdRom.cpp
```diff
@@ -501,7 +511,7 @@ void cdrInterrupt()
 				return;
 
 			SetResultSize(1);
-			cdr.StatP |= STATUS_ROTATING;
+			cdr.StatP |= STATUS_READ;
```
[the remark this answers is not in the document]
kenshen112: No$ at least listed Gran Turismo as one of those

kenshen112:
Another goal for this PR to fix. Games like Metal Gear Solid used to work in pcsx2. Now in master will cause segfaults with the exception "File not found"

`{m_impl=L"cdrom:\\slus_005.94;1" m_convertedToChar={m_str=0x00000000 <NULL> m_len=3435973836 } }` is the path that's being passed from the image. It passes to IsoFs.cpp:128 and runs through the sector reads and passes to the files vector it runs a few rounds then crashes around there

coornio:
There's one title I found so far that the emulator craps itself with.

SLUS 00583 - Duke Nukem - Time to Kill

The only noteworthy message that appears in the log is: 
```
(SYSTEM.CNF) Detected PSX/PSone Disc = cdrom:\slus_005.83;1
File not found.
Path: slus_005.83;1
(pxActionEvent) (thread:EE Core)
```

Attempting to boot any other title, regardless of what it is, after this error took place freezes the emulator when enumerating the BIOS and promptly causes a hard crash.

kenshen112:
This PR is merge able unless there's anything else you guys want me to look at

lightningterror:
> This PR is merge able unless there's anything else you guys want me to look at

You didn't address Sten's review.

kenshen112:
> > This PR is merge able unless there's anything else you guys want me to look at
> 
> You didn't address Sten's review.

I cleared most of it. Unless you'd like me to change the names back like he suggested?

lightningterror:
Yes, change the names back.

kenshen112:
@lightningterror Done

RedPanda4552 (approved):
Tested against Digimon World 2 and Digimon World 3. Disc access is slightly faster across the board. FMVs are still horribly broken due to the pile of issues with PS1DRV, but while still busted, the flickering problem they had seems to be resolved.

May do more tests later but for now I approb.

pcsx2/CDVD/CdRom.cpp
```diff
@@ -552,7 +561,9 @@ void cdrReadInterrupt()
 
 	cdr.OCUP = 1;
 	SetResultSize(1);
-	cdr.StatP |= STATUS_READ | STATUS_ROTATING;
+	if (cdr.StatP & STATUS_SEEK)
+		cdr.StatP &= ~STATUS_SEEK;
```
refractionpcsx2, line 565: You don't really need to check if it's set to unset it, just unset it, it's probably more expensive to check the condition then unset it than just unset it every time.

refractionpcsx2:
You added rapidyaml to your PR.. again.

kenshen112:
> You added rapidyaml to your PR.. again.

Fixed it again

refractionpcsx2:
Initializing Elf: 73728 bytes
(ELF) Size of section headers is not standard

Is that a problem? I feel like this could cause issues reading the ELF headers. You should probably have it correctly detect the size of the ELF header (depending on if it's a PS1 header or PS2 header) and handle it appropriately, instead of YOLOing it, which is what you're doing right now.

kenshen112:
Which game is that?

refractionpcsx2:
That was Diablo 1 NTSC

kenshen112:
So, I made the change to yolo it here, https://github.com/PCSX2/pcsx2/pull/5614/commits/739929cd3e50a57887d385e14d32472674d3bca3 because it was segfaulting on games like playstation underground jampack winter 98 CK's ps1 mode elf parser calls that code here https://github.com/PCSX2/pcsx2/blob/170a03cac606e21581206be04e24e7bb4c99ed5d/pcsx2/CDVD/CDVD.cpp#L456 but my question is. Is that needed for ps1 games at all? psx elf Stenz mentioned shouldn't be parsed through that so I'll have to add a condition to ignore those games and continue. That's one case that could break. But I'm wondering if most games would or wouldn't need that logic?

kenshen112:
Also. Damn that's some solid formatting from github

refractionpcsx2:
the PSX Elf parser may not be correct, it was just a copied and stripped down version of the PS2 one so it didn't crash, it should probably be modified to handle PS1 ELF's correctly if they are more different.

kenshen112:
Rebased onto the remove IopCommon PR. Removed the make upper commit as that proved exceedingly dangerous for ps2 games causing the emulator to become nonfunctional. It seems only a total rewrite of the ps1 elf parser will offer a proper fix for the non-booting games.

refractionpcsx2:
Looks fine to me, but needs conflicts resolving.

I thought the reason we added the PSX Reload Elf thing was to get the console window to say the PSX game ELF information in the title instead of "Running BIOS" but apparently it doesn't even work in master, so IDK what's going on there... Maybe a future PR.

kenshen112:
Yeah, that's what it looked like. But that code was trying to parse the ps1 games like they were PS2's with elf headers and all. When the real parser is in https://github.com/PCSX2/pcsx2/blob/0a2584172ab5341a005ec4573d0510348f8ad597/pcsx2/CDVD/CDVDaccess.cpp#L148

I wonder if we should have a seperate database for ps1 games given how different they seem to be.

PR #5616, opened 2022-03-07
Gamedb: add VU clamping to 'DT Racer' and update MGS2 hack

Mrlinkwii:
### Description of Changes

Gamedb: add VU clamping to 'DT Racer' and update MGS2 hack
### Rationale behind Changes

fix said game(s)
### Suggested Testing Steps

check said games / look at CI


LoStraniero91:
DT Racer VuClamp should be 0 not 1.

PR #5622, opened 2022-03-09
Filesystem: Properly convert stat return to bool.

Ziemas:
Two of the overloads where wrong.

Fixes HLE of stat on the IOP.


stenzek (approved):
Whoops.

PR #5623, opened 2022-03-09
GS/WX: Fix default adapter causing unnecessary GS restarts

stenzek:
### Description of Changes

Easy one, leaving the adapter on default would cause the setting to internally toggle between "Default Adapter" and "", leading to unnecessary GS reopens on settings change.

### Rationale behind Changes

Fixes my bug when I changed adapters from ints to strings.

### Suggested Testing Steps

Already tested with D3D/VK.


refractionpcsx2:
Also fixes the "Default adapter not found" error that kept popping up, so that's good :)

PR #5625, opened 2022-03-10
GS: Properly detect 16bit format on Texture Shuffle + Convert

### LINKED ISSUE #4661: [BUG] SLUS-20724 Firefighter FD18
**Describe the bug**
Graphical glitching in game while playing.  Assets seem to vanish such as the fire, or other in game assets for example in the parking garage the cars will vanish too at times.  This was tested on multiple machines exhibiting the same bug in the same place.  Hitting f9 resolves the issue.

**To Reproduce**
This bug is repeatable by using the following steps on two different stages:

1.  Stage 3-2 play to mission 3 Parking Garage (There is no way to legit start at stage 3-2 mission 3 without using a save state). 
  a. Player will enter the parking garage at start of mission.
  b.  Spray the hose around and move camera and graphical glitches are visible.

2. Stage 4-2 Mission 2 Boss Battle
  a.  During the boss fight the boss vanishes
  b.  The fire vanishes as well.

**Expected behavior**
Have assets visible and not invisible or be missing.

**GS Settings**
Default settings

**Emulation Settings**
Default Settings

**GS Window Screenshots**
gsdx_20210812111938

 -   PCSX2 Revision: 1.6.0
 -   OS: Windows 10 
 -   CPU:   i7-9700 CPU   
 -   GPU:   Nvidia GeForce RTX 2070

**Logs and Dumps**
GSdump:    http://www.filedropper.com/gsdumpfirefighterfd18
MemCard:  [Mcd001_FirefighterFD18_SLUS-20724.zip](https://github.com/PCSX2/pcsx2/files/6977025/Mcd001_FirefighterFD18_SLUS-20724.zip)
sstate:        [sstate_FirefighterFD18_SLUS-20724.zip](https://github.com/PCSX2/pcsx2/files/6977027/sstate_FirefighterFD18_SLUS-20724.zip)

refractionpcsx2:
### Description of Changes
Fixes the detection of 16bit format on shuffle and format convert

### Rationale behind Changes
It was being misdetected as 16bit instead of 32bit when a shuffle happened due to texture shuffles explicitly forcing the texture to 32bit, causing issues.

### Suggested Testing Steps
Test games, make sure post processing isn't broken.

Fixes #4661 but does need Half screen hw renderer fix also (now in Master)

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
+++ b/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
@@ -546,7 +546,9 @@ GSTextureCache::Target* GSTextureCache::LookupTarget(const GIFRegTEX0& TEX0, con
 			dst = CreateTarget(TEX0, new_size.x, new_size.y, type, clear);
 			dst->m_32_bits_fmt = dst_match->m_32_bits_fmt;
 			ShaderConvert shader;
-			bool fmt_16_bits = (psm_s.bpp == 16 && GSLocalMemory::m_psm[dst_match->m_TEX0.PSM].bpp == 16);
+			// m_32_bits_fmt gets set on a shuffle or if the format isn't 16bit.
+			// In this case it needs to make sure it isn't part of a shuffle, where it needs to be interpreted as 32bits.
+			const bool fmt_16_bits = (psm_s.bpp == 16 && GSLocalMemory::m_psm[dst_match->m_TEX0.PSM].bpp == 16 && !dst->m_32_bits_fmt);
 			if (type == DepthStencil)
 			{
 				GL_CACHE("TC: Lookup Target(Depth) %dx%d, hit Color (0x%x, %s was %s)", new_size.x, new_size.y, bp, psm_str(TEX0.PSM), psm_str(dst_match->m_TEX0.PSM));


pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -576,7 +576,7 @@ GSTextureCache::Target* GSTextureCache::LookupTarget(const GIFRegTEX0& TEX0, con
 			dst = CreateTarget(TEX0, new_size.x, new_size.y, type, clear);
 			dst->m_32_bits_fmt = dst_match->m_32_bits_fmt;
 			ShaderConvert shader;
-			bool fmt_16_bits = (psm_s.bpp == 16 && GSLocalMemory::m_psm[dst_match->m_TEX0.PSM].bpp == 16);
+			bool fmt_16_bits = (psm_s.bpp == 16 && GSLocalMemory::m_psm[dst_match->m_TEX0.PSM].bpp == 16 && !dst->m_32_bits_fmt);
```
lightningterror, line 579: Bonus, make it a const.

stenzek:
Dunno if it's worth mentioning where that `m_32_bits_fmt` flag comes from (shuffles get forced to 32bit regardless of PSM). But noting it here for future reference anyway.

PR #5630, opened 2022-03-11
GS: Utilize framebuffer fetch for blending/fbmask where available

stenzek:
### Description of Changes

This PR changes the hardware renderers to use GL_EXT_shader_framebuffer_fetch and VK_ARM_rasterization_order_attachment_access if available. It can result in a massive performance improvement on these drivers.

For now, this only means Intel with OpenGL on Linux. But when/if Mesa supports the Vulkan extension, we'll be ready. It's also potentially applicable for Metal.

Also adds some unrelated changes, like supporting drivers without dual-source blending (probably not many or any left), and Vulkan drivers without D32S8. The latter can be taken advantage of when we use fbfetch, since the stencil buffer isn't needed because DATE goes through the faster fbfetch path instead.

Some quick performance testing at 2x resolution with High blending on a 7700HQ / HD 630 (KabyLake GT2):

 - NFS Carbon Main Menu: fbfetch off **20fps** | fbfetch on **51fps**
 - Monster Hunter: fbfetch off **59fps** | fbfetch on: **265fps**
 - Sly 2: fbfetch off **8fps** | fbfetch on: **78fps**

### Rationale behind Changes

Brr. Too bad more drivers don't support it.

### Suggested Testing Steps

As it touches a fair bit of code, probably need a decent amount of testing, both with and without fbfetch, to make sure there's no performance regressions from the additional paths.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererNew.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererNew.cpp
@@ -599,7 +599,7 @@ void GSRendererNew::EmulateBlending(bool& DATE_PRIMID, bool& DATE_BARRIER, bool&
 		blend_ad_alpha_masked = false;
 
 	u8 blend_index = u8(((m_conf.ps.blend_a * 3 + m_conf.ps.blend_b) * 3 + m_conf.ps.blend_c) * 3 + m_conf.ps.blend_d);
-	const int blend_flag = g_gs_device->GetBlendFlags(blend_index);
+	const int blend_flag = GSDevice::GetBlendFlags(blend_index);
 
 	// Re set alpha, it was modified, must be done after index calculation
 	if (blend_ad_alpha_masked)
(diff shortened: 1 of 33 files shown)


Hirato:
I see no mention of GL_EXT_shader_framebuffer_fetch in mesa's features.txt and almost none in the issue tracker itself.
It might be worthwhile filing an Enhancement Request ticket there, like there is for [VK_EXT_fragment_shader_interlock](https://gitlab.freedesktop.org/mesa/mesa/-/issues/3511) (even though that hasn't really gone anywhere), so mesa's developers at least know there's some demand for it.

tellowkrinkle:
For Vulkan, you'd be looking for `VK_ARM_rasterization_order_attachment_access`, not `GL_EXT_shader_framebuffer_fetch`

I think `GL_EXT_shader_framebuffer_fetch` is supported on all GPUs that can support it (Intel's and some mobile GPUs). [Intel's Windows driver has support for it too](https://opengl.gpuinfo.org/listreports.php?extension=GL_EXT_shader_framebuffer_fetch) but their driver's OGL performance isn't great.

JordanTheToaster:
Tested on an i3 1115G4 with Xe 48EU on Windows 10. Carbon main menu ultra blending 27 fps without 43 fps with but more or less CPU limited Sly 2 28 fps without 120 fps with Underground 2 Full blending 43 fps without 53 fps with.

JordanTheToaster (approved):

pcsx2/GS/Renderers/HW/GSRendererNew.cpp
```diff
@@ -1470,6 +1486,10 @@ void GSRendererNew::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sour
 		m_conf.require_full_barrier = false;
 	}
 
+	// Don't emit the second color output from the pixel shader when it's
+	// not going to be used (no or sw blending).
+	m_conf.ps.no_color1 |= (m_conf.blend.index == 0);
```
tellowkrinkle, line 1491: Mind making this fully check the blend mode and set this flag for all cases where dual source blend isn't used?  Intel GPUs on Metal (and MVK) have a driver bug where they don't write depth if dual source blend outputs go unused.

stenzek:
Extended the COLOR1 eliding to when the dual source output is unused, and also disabled blending when all colours are masked.

Ended up refactoring how blends were handled in the backend; the index only gets used in the frontend now.

Will need re-testing, since the latter is a nontrivial change.

PR #5640, opened 2022-03-13
GameDB: Metal Arms - Glitch in the System (PAL)

Dreadmoth:
### Description of Changes
Force disable mVU Flag Hack to prevent SPS.

### Rationale behind Changes
Prevents graphical glitches. Same fix is already applied to the NTSC version.

mVU Flag Hack enabled:
gs_20220313194752_Metal Arms - Glitch in the System _PAL-M5__SLES-51758

mVU Flag Hack disabled:
gs_20220313194805_Metal Arms - Glitch in the System _PAL-M5__SLES-51758

### Suggested Testing Steps
Test the game with mVU Flag Hack checked / unchecked, SPS should not occur.


bin/resources/GameIndex.yaml
```diff
@@ -11251,6 +11251,8 @@ SLES-51758:
   name: "Metal Arms - Glitch in the System"
   region: "PAL-M5"
   compat: 5
+  speedHacks:
+    mvuFlagSpeedHack: 0  # Fixes SPS.
```
refractionpcsx2, line 11255: can you get rid of the extra space after the 0 please? should only be one space (I know it's a bit inconsistent in the file but I'm trying to rectify that)

RedDevilus:
I bet it affects all the regions.

RedDevilus:
Nvm mvuflaghack was already added for the other serials.

RedDevilus:
Can you check if it needs what the other serials have:

```
  gameFixes:
    - EETimingHack # Texture flicker.
```

Dreadmoth:
> Can you check if it needs what the other serials have which is the EEtiming:

I haven't seen any flickering textures in the first five campaign levels or on any menus - any idea where they might occur?
I searched but couldn't find any specific details about it.

That fix appears to have been added for the NTSC version in 2011 (1cbc1ec), maybe it's no longer needed?

bin/resources/GameIndex.yaml
Dreadmoth: Fixed.

refractionpcsx2:
it seems to work (at the start) without it, so I'm not going to touch it, just in case, it's not doing any harm

PR #5648, opened 2022-03-14
GameDB: GS HW renderer fixes for Tales of the Abyss

Tzerinas:
### Description of Changes
Added Half-Pixel Offset: Special (Texture) to fix ghosting when you're upscaling. The wiki mentions black boxes as an issue too and suggests Workarounds (Trilinear + Mipmapping Full). However I can't reproduce the issue when Half-Pixel Offset is set to Special (Texture).

### Rationale behind Changes
Upscaling goodness & helping out the GameDB

### Suggested Testing Steps
Test Abyss. Bad ghosting shouldn't occur.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -881,6 +881,8 @@ SCAJ-20162:
 SCAJ-20163:
   name: "Tales of the Abyss"
   region: "NTSC-Unk"
+  gsHWFixes:
+    halfPixelOffset: 2
 SCAJ-20164:
   name: "Kingdom Hearts II"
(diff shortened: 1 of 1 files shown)


Ziemas:
If you rebase your branch on top of master it should get rid of the merge. `git rebase origin/master` while in your branch.

Ziemas:
> If you have merge commits it will block the squash, `git reset --hard origin/master` perhaps works?

That would just restore his branch to the same state as master.

RedDevilus:
I suppose, but he could easily apply a commit above but yeah my bad.

PR #5649, opened 2022-03-14
DEV9: Qt Settings Panel

TheLastRar:
### Description of Changes
Adds a Qt UI for configuring DEV9 settings
Mergers the host list into the main settings file in Qt/Core
Adds the option to configure DNS1 or DNS2 to use the internal server (Only Qt has UI for that)

Note the following;
- Validation of Ethernet and HDD settings are no longer performed by the UI, due to the live update nature of Qt I wasn't sure when to validate input.
- The Ethernet Api/Type dropdown doesn't actually set the corresponding settings field, that is performed by the device selection dropdown. This was done to keep these two values in sync (such as when switching API & device mid emulation)
- ~~HDD creation is done on emulation start, with progress logged to console~~
- HDD will not auto create an image, instead you must click `Create Image` to do so
- Modifying settings during emulation may behave unexpectedly (such as when enabling Ethernet before specifying an adapter, or adjusting HDD settings), but I'm not sure what the best course of action for resolving such issues are.

### Rationale behind Changes
Provide Qt UI for DEV9 settings

### Suggested Testing Steps
Test Qt UI
Test that HDD creation still works for Wx

### DIFF
--- a/pcsx2-qt/CMakeLists.txt
+++ b/pcsx2-qt/CMakeLists.txt
@@ -86,6 +86,11 @@ target_sources(pcsx2-qt PRIVATE
 	Settings/MemoryCardSettingsWidget.cpp
 	Settings/MemoryCardSettingsWidget.h
 	Settings/MemoryCardSettingsWidget.ui
+	Settings/DEV9SettingsWidget.cpp
+	Settings/DEV9SettingsWidget.h
+	Settings/DEV9SettingsWidget.ui
+	Settings/HddCreateQt.cpp
+	Settings/HddCreateQt.h
 	Settings/SettingsDialog.cpp
 	Settings/SettingsDialog.h
 	Settings/SettingsDialog.ui
--- a/pcsx2-qt/MainWindow.cpp
+++ b/pcsx2-qt/MainWindow.cpp
@@ -146,6 +146,7 @@ void MainWindow::connectSignals()
 	connect(m_ui.actionGraphicsSettings, &QAction::triggered, [this]() { doSettings("Graphics"); });
 	connect(m_ui.actionAudioSettings, &QAction::triggered, [this]() { doSettings("Audio"); });
 	connect(m_ui.actionMemoryCardSettings, &QAction::triggered, [this]() { doSettings("Memory Cards"); });
+	connect(m_ui.actionDEV9Settings, &QAction::triggered, [this]() { doSettings("Network & HDD"); });
 	connect(
 		m_ui.actionControllerSettings, &QAction::triggered, [this]() { doControllerSettings(ControllerSettingsDialog::Category::GlobalSettings); });
 	connect(m_ui.actionHotkeySettings, &QAction::triggered, [this]() { doControllerSettings(ControllerSettingsDialog::Category::HotkeySettings); });
--- a/pcsx2-qt/MainWindow.ui
+++ b/pcsx2-qt/MainWindow.ui
@@ -111,6 +111,7 @@
     <addaction name="actionGraphicsSettings"/>
     <addaction name="actionAudioSettings"/>
     <addaction name="actionMemoryCardSettings"/>
+    <addaction name="actionDEV9Settings"/>
     <addaction name="actionControllerSettings"/>
     <addaction name="actionHotkeySettings"/>
     <addaction name="separator"/>
@@ -553,6 +554,14 @@
    </property>
    <property name="text">
     <string>&amp;Memory Cards</string>
+   </property>
+  </action>
+  <action name="actionDEV9Settings">
+   <property name="text">
+    <string>&amp;Network &amp;&amp; HDD</string>
+   </property>
+   <property name="icon">
+    <iconset theme="dashboard-line"/>
    </property>
   </action>
   <action name="actionViewToolbar">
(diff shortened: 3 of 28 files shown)


JordanTheToaster:
I have no HDD or Network Games to test that functionality but the UI itself and settings seems to function and save properly so that all looks good to me.

refractionpcsx2:
Not sure what codacy is moaning about, maybe it wants it to be a constexpr if you're gonna do it like this?

TheLastRar:
> Not sure what codacy is moaning about, maybe it wants it to be a constexpr if you're gonna do it like this? (I mean, you could do that anyway)

constexpr doesn't work with QString or QRegularExpression, but I can make the QRegularExpression fields static const. hopefully that is close to what you are after

refractionpcsx2:
Seemed to be enough to shut Codacy up :) Thanks

TheLastRar:
I admittedly didn't consider per game settings

I've addressed the bugs you have shown, although I may need to give that another look though

I've also disabled the DNS hosts tab in the per game settings view for now, as I'm not sure how to add per game setting support to that.

Blackbird88:
Another minor bug in per-game settings.
If you change the DNS setting from **Use Global Setting** to **Manual** and back then it won't be greyed out (Auto doesn't have this problem)
image

TheLastRar:
> Another minor bug in per-game settings. If you change the DNS setting from **Use Global Setting** to **Manual** and back then it won't be greyed out (Auto doesn't have this problem) image

Is that still the case with the latest commit?

Blackbird88:
Yes, I forgot to say that the bugs I mentioned before are fixed and this one still happens on latest commit.

TheLastRar:
Last commit should have addressed the final issue you reported. thanks for testing

Blackbird88:
I wiped my game settings and everything still works fine! However I forgot about this visual issue.
When global settings are like this
image
per-game looks like this
image
It still works fine, but maybe it should also show "thing (Auto)" like the rest to not confuse the user?

Blackbird88:
Perfect! I don't know if this is a bug, but with the new change Ethernet Device shows one less thing in per-game
Global
image
image

RedDevilus:
Looks to be a packet sniffer like wireshark that is used for debugging your own network. Which is useless for dev9.

TheLastRar:
> Perfect! I don't know if this is a bug, but with the new change **Ethernet Device** shows one less thing in per-game Global image Per-Game image In all honesty I have no clue what that loopback thingy is. Some NMap debug stuff?

Intentional, you would need to pick the non-global `PCAP Bridged` to see the full list of adapters. I wanted to ensure that either both or none of those dropdowns where per-game set
I guess that wasn't clear in the UI, but I dunno what would be best way to show this is (maybe I could disable the adapter list when the globally set ethernet type is set?)

Blackbird88:
Oh I see now, my mistake. I think it's fine as it is. I am just blind :D

Blackbird88:
Noticed this one just now. This is my per-game ini
`[DEV9/Eth]`
`InterceptDHCP = true`
`PS2IP = 192.168.0.250`
`ModeDNS1 = Manual`
`DNS1 = 45.7.228.197`

But in GUI it shows DNS1 is Auto not Manual as it should be
image

TheLastRar:
> Noticed this one just now. This is my per-game ini 
> `[DEV9/Eth]`
> `InterceptDHCP = true`
> `PS2IP = 192.168.0.250`
> `ModeDNS1 = Manual`
> `DNS1 = 45.7.228.197`
> 
> But in GUI it shows DNS1 is Auto not Manual as it should be

Should be addressed now, also took the opportunity to squash my commits

Blackbird88:
Something odd with HDD size if you go below 40 in per-game
Global
image
Per-game
image
Is this actual feature or just bug?

TheLastRar:
Going below 40 in per game settings should allow you to use global settings, and it intentional
The UI for that however isn't final, and may get revised at some point.

The other issue is a bug, my bad, should be fixed now.

Blackbird88:
> Going below 40 in per game settings should allow you to use global settings, and it intentional

Well the problem is as you can see from the screenshots, it doesn't match, Global is set to 120 while per-game thinks Global is 40

TheLastRar:
> > Going below 40 in per game settings should allow you to use global settings, and it intentional
> 
> Well the problem is as you can see from the screenshots, it doesn't match, Global is set to 120 while per-game thinks Global is 40

I've seen that happen with some of the other DEV9 settings, will need to look into that further

TheLastRar:
Gave the HDD creation a second pass, now that has a Qt powered progress bar

You now need to click create Image to generate an image file, PCSX2 will no longer do it automatically for you in Qt
Doing it this way I think is best, given the live updating nature of settings in Qt

TheLastRar:
Also, seems Codacy's decided to be a pain

image

I think it's not rechecking the file after a force push, but I've got no clue

refractionpcsx2:
I refreshed Codacy and it's still saying the same thing, is it because you're initialising it to nullptr instead of a value it doesn't like it?

TheLastRar:
> I refreshed Codacy and it's still saying the same thing, is it because you're initialising it to nullptr instead of a value it doesn't like it?

I Could try initialising to something

TheLastRar:
Seems it's still not happy

refractionpcsx2:
Is this ready now?

TheLastRar:
Yeah

PR #5662, opened 2022-03-15
GameDB: Add gsHWFixes for GTA LCS/VCS

Blackbird88:
### Description of Changes
Adds upscaling fixes for Grand Theft Auto: Liberty City Stories and Grand Theft Auto: Vice City Stories

### Rationale behind Changes
LCS uses some fullscreen effect that causes it to be blurry when upscaled.
Before
image
After
image

VCS uses some kind of bloom when in-game Trails option is turned that becomes misaligned when upscaled
Before
image
After
image

### Suggested Testing Steps
Test the games if all works fine.


Blackbird88:
Dang I forgot about the commit naming. If someone could fix it I would be grateful 😃

PR #5676, opened 2022-03-16
Common: Removed array-index manipulation logic for more standardised abstractions

tcoyvwac:
### Description of Changes

* Manipulation of array indexes to assign data, changed to be more standardised and to strengthen `const`ness.
* Method to locate particular `EGLConfig`-data, changed to be more C++ standardised.

### Rationale behind Changes

* Standardisation, to use more C++ standard algorithms and IIFE (Immediately invoked function expressions).
  * This makes the source-code easier-to-read, as it is using common-abstracted C++ concepts.
* Much reduced code-size to maintain.
* Stronger `const`-correctness.

### Suggested Testing Steps
* CI should pass and build with no problems.
  * _Structurally, nothing has changed, only replaced previous code with refactored C++ common algorithm code._

---
_P.S. This is a great project! :smile_cat: Thank you for all your work so far! :tada:_

### DIFF
--- a/common/GL/ContextEGL.cpp
+++ b/common/GL/ContextEGL.cpp
@@ -334,17 +334,15 @@ namespace GL
 			config = configs.front();
 		}
 
-		int attribs[8];
-		int nattribs = 0;
-		if (version.profile != Profile::NoProfile)
-		{
-			attribs[nattribs++] = EGL_CONTEXT_MAJOR_VERSION;
-			attribs[nattribs++] = version.major_version;
-			attribs[nattribs++] = EGL_CONTEXT_MINOR_VERSION;
-			attribs[nattribs++] = version.minor_version;
-		}
-		attribs[nattribs++] = EGL_NONE;
-		attribs[nattribs++] = 0;
+		const auto attribs = [version]() -> std::array<int, 8> {
+			if (version.profile != Profile::NoProfile)
+				return {
+					EGL_CONTEXT_MAJOR_VERSION, version.major_version,
+					EGL_CONTEXT_MINOR_VERSION, version.minor_version,
+					EGL_NONE
+				};
+			return {EGL_NONE};
+		}();
 
 		if (!eglBindAPI((version.profile == Profile::ES) ? EGL_OPENGL_ES_API : EGL_OPENGL_API))
 		{
@@ -352,7 +350,7 @@ namespace GL
 			return false;
 		}
 
-		m_context = eglCreateContext(m_display, config.value(), share_context, attribs);
+		m_context = eglCreateContext(m_display, config.value(), share_context, attribs.data());
 		if (!m_context)
 		{
 			Console.Error("eglCreateContext() failed: %d", eglGetError());
--- a/common/GL/ContextEGL.cpp
+++ b/common/GL/ContextEGL.cpp
@@ -285,23 +285,18 @@ namespace GL
 		Console.WriteLn(
 			"Trying version %u.%u (%s)", version.major_version, version.minor_version,
 			version.profile == Context::Profile::ES ? "ES" : (version.profile == Context::Profile::Core ? "Core" : "None"));
-		int surface_attribs[16];
-		int nsurface_attribs = 0;
-		surface_attribs[nsurface_attribs++] = EGL_RENDERABLE_TYPE;
-		surface_attribs[nsurface_attribs++] = (version.profile == Profile::ES) ?
-                                                  ((version.major_version >= 3) ? EGL_OPENGL_ES3_BIT :
-                                                                                  ((version.major_version == 2) ? EGL_OPENGL_ES2_BIT : EGL_OPENGL_ES_BIT)) :
-                                                  EGL_OPENGL_BIT;
-		surface_attribs[nsurface_attribs++] = EGL_SURFACE_TYPE;
-		surface_attribs[nsurface_attribs++] = (m_wi.type != WindowInfo::Type::Surfaceless) ? EGL_WINDOW_BIT : 0;
-		surface_attribs[nsurface_attribs++] = EGL_RED_SIZE;
-		surface_attribs[nsurface_attribs++] = 8;
-		surface_attribs[nsurface_attribs++] = EGL_GREEN_SIZE;
-		surface_attribs[nsurface_attribs++] = 8;
-		surface_attribs[nsurface_attribs++] = EGL_BLUE_SIZE;
-		surface_attribs[nsurface_attribs++] = 8;
-		surface_attribs[nsurface_attribs++] = EGL_NONE;
-		surface_attribs[nsurface_attribs++] = 0;
+
+		const int renderable_type = version.profile == Profile::ES
+			? ((version.major_version >= 3) ? EGL_OPENGL_ES3_BIT : ((version.major_version == 2) ? EGL_OPENGL_ES2_BIT : EGL_OPENGL_ES_BIT))
+			: EGL_OPENGL_BIT;
+		const int surface_attribs[] = {
+			EGL_RENDERABLE_TYPE, renderable_type,
+			EGL_SURFACE_TYPE, (m_wi.type != WindowInfo::Type::Surfaceless) ? EGL_WINDOW_BIT : 0,
+			EGL_RED_SIZE, 8,
+			EGL_GREEN_SIZE, 8,
+			EGL_BLUE_SIZE, 8,
+			EGL_NONE
+		};
 
 		EGLint num_configs;
 		if (!eglChooseConfig(m_display, surface_attribs, nullptr, 0, &num_configs) || num_configs == 0)
--- a/common/GL/ContextEGL.cpp
+++ b/common/GL/ContextEGL.cpp
@@ -313,21 +313,18 @@ namespace GL
 		}
 		configs.resize(static_cast<u32>(num_configs));
 
-		std::optional<EGLConfig> config;
-		for (EGLConfig check_config : configs)
-		{
-			if (CheckConfigSurfaceFormat(check_config))
+		const auto m_config = [this, &configs]() {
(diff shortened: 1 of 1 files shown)


lightningterror (commented):
Formatting can be better.

common/GL/ContextEGL.cpp
```diff
@@ -285,74 +287,56 @@ namespace GL
...
-		surface_attribs[nsurface_attribs++] = EGL_BLUE_SIZE;
-		surface_attribs[nsurface_attribs++] = 8;
-		surface_attribs[nsurface_attribs++] = EGL_NONE;
-		surface_attribs[nsurface_attribs++] = 0;
+
+		const auto surface_attribs = [this, version]() -> std::array<int, 16> {
+			const auto renderable_type_value = (version.profile == Profile::ES) ?
+												   ((version.major_version >= 3) ? EGL_OPENGL_ES3_BIT :
+                                                                                   ((version.major_version == 2) ? EGL_OPENGL_ES2_BIT : EGL_OPENGL_ES_BIT)) :
+                                                   EGL_OPENGL_BIT;
```
refractionpcsx2, line 295: There seems to be a mix of tabs and spaces, better just stick to tabs

tcoyvwac:
Humm, is there a _new_ particular format / formatting which is preferred? The current format was done by the project's `.clang-format` so I am slightly confused! :sweat_smile: 

I am happy to adjust and add special `clang-format off` blocks if you like.

common/GL/ContextEGL.cpp
tcoyvwac: The formatting has been done by the project's `.clang-format`! :sweat_smile: (You can see that the previous code was somewhat similar styled)!

I am happy to make a special `clang-format off` for this section, and make the code _stick to tabs_ if needed / if you like.

tellowkrinkle (commented):
If you'd like to use clang-format, please stick to `git clang-format` instead of blindly formatting entire files.  This makes it so you don't have to throw clang-format off tags around every spot where clang-format is dumb and doesn't know how to indent with tabs.

And sorry, I'm a bit more picky with the formatting of these key-value lists, clang-format doesn't really understand them.

tcoyvwac:
> If you'd like to use clang-format, please stick to `git clang-format` ...
> [...] 
> And sorry, I'm a bit more picky with the formatting of these key-value lists, clang-format doesn't really understand them.

Interestingly and to confirm things, the formatting was done _exactly this way_, by using `git clang-format` as part of development before making the PR. Humm, I think if there are any problems with the formatting, the project's `.clang-format` may need adjusting. One can verify this by cloning this branch and run `git clang-format` to verify if you like. 

> ...**clang-format is dumb**...

Agree with you @tellowkrinkle! I have no personal preferences on formatting when it comes to open-source projects, I just let the project's `.clang-format` do its thing on the file, that's why I assume it is there! :wink: Happy to make any adjustments to accommodate `clang-format` dumbness.

common/GL/ContextEGL.cpp
```diff
@@ -285,74 +287,54 @@ namespace GL
...
-		surface_attribs[nsurface_attribs++] = EGL_NONE;
-		surface_attribs[nsurface_attribs++] = 0;
+
+		const auto surface_attribs = [this, version]() -> std::array<int, 16> {
+			const auto opengl_es_version_number = [](auto major_version) {return ((major_version >= 3) ? EGL_OPENGL_ES3_BIT : ((major_version == 2) ? EGL_OPENGL_ES2_BIT : EGL_OPENGL_ES_BIT)); };
+			const auto renderable_type_value = (version.profile == Profile::ES) ? opengl_es_version_number(version.major_version) : EGL_OPENGL_BIT;
+			const auto surface_type_value = (m_wi.type != WindowInfo::Type::Surfaceless) ? EGL_WINDOW_BIT : 0;
+
+			return {EGL_RENDERABLE_TYPE, renderable_type_value, EGL_SURFACE_TYPE, surface_type_value, EGL_RED_SIZE, 8, EGL_GREEN_SIZE, 8, EGL_BLUE_SIZE, 8, EGL_NONE, 0};
+		}();
```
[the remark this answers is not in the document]
tcoyvwac: _When dealing with unknown code in an unknown codebase, the lambda-form can be a lifesaver in determining and clearly-stating side-effects from unknown external variables!_...But I agree with you @tellowkrinkle , in this case, it is ridiculous now with your clarification! :sweat_smile: Thanks!

---

So... in the original code, the variable [`surface_attribs` was strongly stated as `surface_attribs[16]`](https://github.com/PCSX2/pcsx2/pull/5676/files#diff-5be66afa8049cdaefb8fd5b469f92661ece7f3459cbbfc74f2a1ae3fa6c7e206L288). Because of the `C`-form declaration, the remaining unused array elements were not initialised. This may cause erroneous behaviour and was the main reason why `surface_attribs` was changed to `const std::array<int 16>`.

Is it ok @tellowkrinkle to adjust the datatype you're suggesting (a variable-length `C` array) to `const std::array<int, 16>`? Currently, your form could cause problems if something is expecting an array-datatype with `16` elements...

common/GL/ContextEGL.cpp
```diff
@@ -227,8 +227,10 @@ namespace GL
 		const u32 height = std::max<u32>(m_wi.surface_height, 1);
 
 		EGLint attrib_list[] = {
-			EGL_WIDTH, static_cast<EGLint>(width),
-			EGL_HEIGHT, static_cast<EGLint>(height),
+			EGL_WIDTH,
+			static_cast<EGLint>(width),
+			EGL_HEIGHT,
+			static_cast<EGLint>(height),
```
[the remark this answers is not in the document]
tcoyvwac: `git clang-format` being dumb... Happy to ignore and revert!

common/GL/ContextEGL.cpp
```diff
@@ -285,74 +287,54 @@ namespace GL
...
-			attribs[nattribs++] = EGL_CONTEXT_MAJOR_VERSION;
-			attribs[nattribs++] = version.major_version;
-			attribs[nattribs++] = EGL_CONTEXT_MINOR_VERSION;
-			attribs[nattribs++] = version.minor_version;
-		}
-		attribs[nattribs++] = EGL_NONE;
-		attribs[nattribs++] = 0;
+		const auto attribs = [version]() -> std::array<int, 8> {
+			if (version.profile != Profile::NoProfile)
+				return {EGL_CONTEXT_MAJOR_VERSION, version.major_version, EGL_CONTEXT_MINOR_VERSION, version.minor_version, EGL_NONE, 0};
```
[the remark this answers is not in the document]
tcoyvwac: No problems, `git clang-format` being `git clang-format` here! :sweat_smile:

tellowkrinkle:
I assume that final formatting commit wasn't from `git clang-format`?

And yes, for the other ones we're going to have to fight with clang-format since the issues are with newly added code.  (I usually try not to complain about newly added code as long as it matches clang-format since I realize not everyone wants to manually format their code.  This is an exception, which is why I mentioned being picky about those key-value lists)

tcoyvwac:
> I assume that final formatting commit wasn't from git clang-format? That's what I was talking about when I mentioned using git clang-format, since all it does is edit a bunch of lines that weren't touched by the rest of the changes.

To double-check, I went to my bash-history to confirm the issue... Here are the results:
```
clang-format -- pcsx2/common/GL/ContextEGL.cpp  
git clang-format --  pcsx2/common/GL/ContextEGL.cpp  
```
_(`clang-format` is aliased to `clang-format -i`)_

---

Whoops haha! Mystery solved! :smile: 
Will delete the commit: https://github.com/PCSX2/pcsx2/pull/5676/commits/d6dfcfb5446e90b5151afef35cd2393998b82a53. :+1:

common/GL/ContextEGL.cpp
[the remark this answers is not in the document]
tellowkrinkle: `eglChooseConfig` uses `EGL_NONE` as a terminator, so it should be fine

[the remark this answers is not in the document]
tellowkrinkle: I assume the original choice of 16 was a "I don't want to count these, 16 should be enough" but with an initializer list, the compiler counts for us so we don't have to worry

common/GL/ContextEGL.cpp
```diff
@@ -318,41 +313,34 @@ namespace GL
...
-		for (EGLConfig check_config : configs)
-		{
-			if (CheckConfigSurfaceFormat(check_config))
-			{
-				config = check_config;
-				break;
-			}
-		}
+		auto config = [this, &configs]() {
+			const auto found_config = std::find_if(std::begin(configs), std::end(configs), [&](auto&& check_config) { return CheckConfigSurfaceFormat(check_config); });
```
tellowkrinkle, line 317: Is there anything preventing you from using `CheckConfigSurfaceFormat` directly instead of wrapping it in a lambda that calls it?

Also you can remove the usage of the `std::optional` completely if you pull the lower `!config.has_value()` check into here

common/GL/ContextEGL.cpp
RedDevilus: Clang formatting in PCSX2 has a lot to be desired unfortunately. Don't think there is a correct and easy way to go about it.

tcoyvwac: The extra formatting on certain untouched parts of the codebase were on my part (running `clang-format` instead of `git clang-format` sorry!), but all linters are not "optimal" in the end, so happy to just adjust to whatever helpful comments to correct the problems any time.

common/GL/ContextEGL.cpp
tellowkrinkle: I mean like this
```cpp
std::find_if(std::begin(configs), std::end(configs), CheckConfigSurfaceFormat)
```

tellowkrinkle (approved):

PR #5682, opened 2022-03-17
GS: Per-selector switch for whether to use JIT for DrawScanline

tellowkrinkle:
### Description of Changes
Adds a per-selector switch for using JIT for DrawScanline

### Rationale behind Changes
Easier to track down JIT bugs

### Suggested Testing Steps
Run with the environment variable `USE_C_DRAW_SCANLINE` set to a file name.  Open that file and edit some of the `N`s to `Y`s.  See if it still works (but slower).

Tested working on macOS x64.  x86 and Windows x64 are not yet tested


refractionpcsx2:
Do you wanna remove the 32bit stuff? :)

refractionpcsx2:
LGTM, ran a few dumps and games, all worked okay, also tried with the JIT disabled to make sure that still works, all seems fine

PR #5683, opened 2022-03-17
GameDB: Add DMABusy fix to Shining Wind

KrossX:
### Description of Changes
Booting Shining Wind results in a crash with the assert ["Gif Path Buffer Overflow!"](https://github.com/PCSX2/pcsx2/blob/515d9040a103ba928301580c5c81e4c87b95d885/pcsx2/Gif_Unit.h#L323). Either EETimingHack or DMABusyHack seems to fix booting, picked the DMA one to avoid Emotion.

### Suggested Testing Steps
Run the game with a master build, see if it crashes.
Run it again with manual game fixes, enable Handle DMAC, see if it boots.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -20636,8 +20636,10 @@ SLKA-25214:
   gsHWFixes:
     roundSprite: 2 # Fixes font artifacts.
 SLKA-25215:
-  name: "Shining Wild"
+  name: "Shining Wind"
   region: "NTSC-K"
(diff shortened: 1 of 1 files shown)


Mrlinkwii:
```
 SLKA-25215
 SLPM-74261
```

the game also has these serials , can you add it to them is well?

KrossX:
Done, although I cannot test if it was necessary for those titles. I suppose the "The Best" release is the very same, but is the korean version the same as well?

PR #5692, opened 2022-03-18
Qt: Fix bugs relating to per-game settings

TheLastRar:
### Description of Changes
Fix the following bugs
One of the overloads for BindWidgetToEnumSetting() would select the wrong index if a per-game setting differed from global
setStringSettingValue would try to save the string as a bool when saving as a per-game setting

### Rationale behind Changes
Bug fixes

### Suggested Testing Steps
The DEV9 Qt PR UI has these fixes, and code that makes use of these functions, so you can check if it works there.
Other than that I guess a unit test can be written to test these functions.

### DIFF
--- a/pcsx2-qt/SettingWidgetBinder.h
+++ b/pcsx2-qt/SettingWidgetBinder.h
@@ -435,7 +435,7 @@ namespace SettingWidgetBinder
 
 			int sif_value;
 			if (sif->GetIntValue(section.c_str(), key.c_str(), &sif_value))
-				Accessor::setNullableIntValue(widget, sif_value);
+				Accessor::setNullableIntValue(widget, sif_value - option_offset);
 			else
 				Accessor::setNullableIntValue(widget, std::nullopt);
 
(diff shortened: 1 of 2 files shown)


andyten:
not repaired 
org
org1
bug2
bug2

stenzek (approved):
LGTM

andyten:
bug3
bug4

TheLastRar:
Can you provide me reproduction steps for the Anisotropic Filtering. I think I'm aware of that issue, but I would need more information to confirm

andyten:
> Can you provide me reproduction steps for the Anisotropic Filtering. I think I'm aware of that issue, but I would need more information to confirm

1 open  per-game setting window
2 Anisotropic Filtering ,set 8x
3 close per-game setting window
4 close pcsx2
5 run pcsx2
6 open  per-game setting window

org

TheLastRar:
I did fix alot of the issues reported

The Renderer dropdown and Enable Game Fixes checkbox have not been fixed, I'm considering that outside the scope of this pr

PR #5707, opened 2022-03-19
Build: Remove 32bit code from project

refractionpcsx2:
### Description of Changes
Removes all 32bit only code and gets rid of "if 64bit" blocks.

### Rationale behind Changes
We have dropped 32bit, plus code cleanliness

### Suggested Testing Steps
Test game, SW renderer and all JIT's to make sure nothing is busted.


tealsuki:
The buildbot build for this (2486) threw a (hopefully) false positive in Windows Defender

image

refractionpcsx2:
We don't control buildbot builds, but yes it's a false positive. I don't recall coding in any trojans anyway.

PR #5715, opened 2022-03-20
GS: Add Direct3D 12 renderer

stenzek:
### Description of Changes

What the title says.

This seemed kinda redundant with the Vulkan renderer existing, until we found out that barriers were broken, and it was super unstable in the Intel drivers.

Even on AMD, DX12 ends up faster than Vulkan in some games (assuming barriers are off).

Long term, we'll probably look at doing sw blending via ROVs in DX12. But for now, it's only DX11 levels of accuracy (except it has my blending-based primitive ID DATE).

Larger line count is due to pulling in D3D12MemoryAllocator. The actual LOC increase in GS is about 3.5k.

### Rationale behind Changes

Brr on Intel.

### Suggested Testing Steps

Early days so probably plenty of bugs. Apparently some games have issues on Intel GPUs due to compiler issues which I still need to look into (e.g. NFSU2).


seta-san:
good future proofing update. windows 8.1 support ends on jan 10 2023.. after that dx11 is obsolete

refractionpcsx2:
What? No it isn't.

kaosjr:
> good future proofing update. windows 8.1 support ends on jan 10 2023.. after that dx11 is obsolete

What are you even on about? This is just flat out wrong. You can still run even DX9 and older versions on Win10 natively.

seta-san:
> What? No it isn't.

I’m saying we can safely drop support for dx11 after windows 8.1 loses support. Windows 10 and newer all support 12

refractionpcsx2:
yes but not all hardware that runs on win10 supports dx12.

lightningterror:
> > What? No it isn't.
> 
> I’m saying we can safely drop support for dx11 after windows 8.1 loses support. Windows 10 and newer all support 12

We won't.

K0bin:
> Long term, we'll probably look at doing sw blending via ROVs in DX12. But for now, it's only DX11 levels of accuracy (except it has my blending-based primitive ID DATE).

Do you also plan to implement that in the Vulkan renderer as well? I know AMD doesn't support VK_EXT_fragment_shader_interlock but Nvidia does and Mesa might add it in the future.

Medal2009:
> ### Description of Changes
> What the title says.
> 
> This seemed kinda redundant with the Vulkan renderer existing, until we found out that barriers were broken, and it was super unstable in the Intel drivers.
> 
> Even on AMD, DX12 ends up faster than Vulkan in some games (assuming barriers are off).
> 
> Long term, we'll probably look at doing sw blending via ROVs in DX12. But for now, it's only DX11 levels of accuracy (except it has my blending-based primitive ID DATE).
> 
> Larger line count is due to pulling in D3D12MemoryAllocator. The actual LOC increase in GS is about 3.5k.
> 
> Brr on Intel.
> 
> ### Suggested Testing Steps
> Early days so probably plenty of bugs. Apparently some games have issues on Intel GPUs due to compiler issues which I still need to look into (e.g. NFSU2).

Hi, i download new versions pcsx2 but find issues graphic, images games glitchy not normal, how to fix it.?

refractionpcsx2:
You use our discord or forums to try and work out what's going on.  picking pull requests on our github is not the place for you to get help.

bearoso:
> 



> I know AMD doesn't support VK_EXT_fragment_shader_interlock but Nvidia does and Mesa might add it in the future.

From what I know, it's not going to be very fast on AMD/Mesa, so don't get your hopes up.

refractionpcsx2:
> > 
> 
> > I know AMD doesn't support VK_EXT_fragment_shader_interlock but Nvidia does and Mesa might add it in the future.
> 
> From what I know, it's not going to be very fast on AMD/Mesa, so don't get your hopes up.

We've already tried ROV in the past, and while it wasn't as bad on Jak 2 as OpenGL, it was still slow compared to Nvidia.

K0bin:
I'm personally not particularly worried about AMD support, I'd just like to have this working in the Linux version.

stenzek:
Firmware limitations prevent blending from being fast on AMD. See the thread here: https://np.reddit.com/r/emulation/comments/rz7rqj/pcsx2_vulkan_released_in_latest_dev_builds/hrwoldv/?context=3

Personally I don't see much benefit in implementing FSI-based blending just for NVIDIA on Linux, when it could be done with ROV in Direct3D and support all three major vendors. NVIDIA still has barrier based blending on Linux anyway, and it's very fast.

lightningterror (commented):
Some tweaks:
On window title it says D3D12 SW instead of HW,
Adding D3D12 option for GS debugger,
The following dump crashes on dx12 [50 Cent - Bulletproof - Texture shuffle.zip](https://github.com/PCSX2/pcsx2/files/8362488/50.Cent.-.Bulletproof.-.Texture.shuffle.zip)

JordanTheToaster (approved):

PR #5717, opened 2022-03-20
Build: Remove 32bit configs from vsprops

stenzek:
What the title says. If you don't remove these, when you add a new project, it recreates 32bit again.


stenzek:
Qt build failure is unrelated (random CI failure it seems, it should've aborted when it downloaded the binaries, but didn't).

refractionpcsx2:
whoops, forgot about props! Thanks :)

PR #5718, opened 2022-03-20
GS: Use pixel format mask for FBMSK checks

refractionpcsx2:
### Description of Changes
Uses the preconfigured pixel format masks for doing FBMSK checks to see if the channel is entirely masked or not.

### Rationale behind Changes
Before it just did a blanket check of the whole 32bits (in some cases) which may not always be the case, this uses the pixel format mask to only check the bits we care about.
More brr in some scenarios?
Bit cleaner than having "this number, or this number, or this number" in places.

### Suggested Testing Steps
Test any games, especially ones with fancy post effects, make sure nothing is broken.

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -2730,7 +2730,7 @@ __forceinline void GSState::VertexKick(u32 skip)
 			break;
 	}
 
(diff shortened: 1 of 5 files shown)


lightningterror:
Tested some dumps, they seemed fine.

PR #5719, opened 2022-03-20
Gamedb: Remove VU rounding form 'Hitman - Contracts'

Mrlinkwii:
### Description of Changes

Remove VU rounding form 'Hitman - Contracts'
### Rationale behind Changes

Having the rounding setting breaks the lighting, not having it makes a curtain invisible and removes some light cones. Removing it is the lesser evil
### Suggested Testing Steps

look at CI/test said game


LoStraniero91:
With the new separate options for VUs, the VU1 Rounding to Nearest fixes the missing graphic without breaking the lighting.

PR #5725, opened 2022-03-21
32bit cleanup

lightningterror:
### Description of Changes

Removes leftover 32bit code.
### Rationale behind Changes

32bit code cleanup.
### Suggested Testing Steps

Builds, stuff checks out.


common/emitter/instructions.h
```diff
@@ -57,9 +57,9 @@ namespace x86Emitter
 	// flags.
 
 	extern const xImpl_Mov xMOV;
-#ifdef __M_X86_64
+
 	extern const xImpl_MovImm64 xMOV64;
-#endif
+
```
refractionpcsx2, line 62: any need to have the blank lines here?

common/emitter/movs.cpp
```diff
@@ -55,37 +55,15 @@ namespace x86Emitter
...
-		{
-			xOpAccWrite(to.GetPrefix16(), to.Is8BitOp() ? 0xa0 : 0xa1, to, src);
-			xWrite32(src.Displacement);
-		}
-		else
-#endif
-		{
-			xOpWrite(to.GetPrefix16(), to.Is8BitOp() ? 0x8a : 0x8b, to, src);
-		}
+		xOpWrite(to.GetPrefix16(), to.Is8BitOp() ? 0x8a : 0x8b, to, src);
```
refractionpcsx2, line 66: why did you remove all this code?

common/emitter/x86emitter.cpp
```diff
@@ -151,17 +151,12 @@ const xRegister8
...
     dh(6), bh(7);
 
-#if defined(_WIN32) || !defined(__M_X86_64)
+#if defined(_WIN32)
 const xAddressReg
     arg1reg = rcx,
     arg2reg = rdx,
-#ifdef __M_X86_64
     arg3reg = r8,
     arg4reg = r9,
```
refractionpcsx2, line 159: Not quite convinced the logic here is right, it should probably be
if win32
rcx, rdx
else
r8, r9

PR #5752, opened 2022-03-24
Qt: GS Settings Fixes

TheLastRar:
### Description of Changes
Sets value of Anisotropic off to `0` instead of `1`
Disable the Anisotropic dropdown when GPU palette conversion is on
Fix issues with the renderer dropdown regarding per-game settings
Fix typo with 8x scale
Remove `Custom` Resolution entry

### Rationale behind Changes
Core sets the default Anisotropic to 0 for Off, while the Qt UI sets the default to 1 (also for Off), make this consistent
Anisotropic off appears to be assumed 0 in a few locations (assuming I'm reading the code correctly) such as https://github.com/PCSX2/pcsx2/blob/45682c382f349c39089d27814b23251a9f721c0b/pcsx2/GS/Renderers/DX11/GSTextureFX11.cpp#L232

Disabling the Anisotropic dropdown when GPU palette conversion is on was requested by LT

Fix bugs & typos

We don't have Custom resolutions

### Suggested Testing Steps
Test the Anisotropic setting behaviour
Test if toggling GPU palette conversion disables/enables the Anisotropic dropdown
Also test the above with per-game settings
Test the render dropdown with per-game settings
Test that the resolution dropdown behaves correctly & selects the correct scaling


stenzek (approved):
Strictly speaking aniso off should be 1, not zero, but that sounds like a separate issue.

TheLastRar:
> Strictly speaking aniso off should be 1, not zero, but that sounds like a separate issue.

I based my decision on the default setting in config.h, along with if statements like the one I linked

I can, however, revert that change if you prefer.

stenzek:
Nah, it's fine here; we should change on the GS side.

PR #5758, opened 2022-03-25
128mb of ram toggle

ThirteenAG:
### Description of Changes
Bringing an option to have 128mb of EE ram, from pcsx2-rr project (https://github.com/xTVaser/pcsx2-rr/issues/102).
Since memory layout is set up at compile-time, this basically just allocates more memory regardless of what the toggle status is, but I added a check for it in couple of internal functions that will behave accordingly.
pcsx2x64-dbg_2022-03-25_14-29-52
pcsx2-qtx64-dbg_2022-03-25_17-03-31

### Rationale behind Changes
More memory can be used to allocate custom code and data in it, which greatly improves possibilities to create mods(including pnach).
With both, toggle enabled and disabled, GetMemorySize syscall returns `0x2000000`, so this does not make the games aware of more memory, and it's not the point of this PR. Also this shouldn't break anything compatibility wise etc.
Tested by manually creating a call to address beyond 32mb limit, the code executes with no issues:
pcsx2x64-dbg_2022-03-25_15-03-50

pcsx2x64-dbg_2022-03-25_15-04-34

### Suggested Testing Steps
You can see more memory being available at address 0x02000000 in the debugger.

Toggle enabled:
pcsx2x64-dbg_2022-03-25_14-29-14

Toggle disabled:
pcsx2x64-dbg_2022-03-25_14-28-52

### DIFF
--- a/pcsx2-qt/Settings/AdvancedSystemSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/AdvancedSystemSettingsWidget.cpp
@@ -33,6 +33,7 @@ AdvancedSystemSettingsWidget::AdvancedSystemSettingsWidget(SettingsDialog* dialo
 
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.eeRecompiler, "EmuCore/CPU/Recompiler", "EnableEE", true);
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.eeCache, "EmuCore/CPU/Recompiler", "EnableEECache", false);
+	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.ee128mbram, "EmuCore/CPU/Recompiler", "EnableEE128mbRam", false);
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.eeINTCSpinDetection, "EmuCore/Speedhacks", "IntcStat", true);
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.eeWaitLoopDetection, "EmuCore/Speedhacks", "WaitLoop", true);
 
--- a/pcsx2/System.h
+++ b/pcsx2/System.h
@@ -42,41 +42,43 @@ class RecompiledCodeReserve;
 
 namespace HostMemoryMap
 {
-	static const u32 Size = 0x28000000;
+	static constexpr u32 _128mbOffset = 0x6000000;
+
+	static constexpr u32 Size = 0x28000000 + _128mbOffset; // 160mb(base) + 96MB(extra needed to bring up to 128MB)
 
 	// The actual addresses may not be equivalent to Base + Offset in the event that allocation at Base failed
 	// Each of these offsets has a debugger-accessible equivalent variable without the Offset suffix that will hold the actual address (not here because we don't want code using it)
 
 	// PS2 main memory, SPR, and ROMs
-	static const u32 EEmemOffset   = 0x00000000;
+	static constexpr u32 EEmemOffset   = 0x00000000; // 128mb
 
 	// IOP main memory and ROMs
-	static const u32 IOPmemOffset  = 0x04000000;
+	static constexpr u32 IOPmemOffset  = 0x04000000 + _128mbOffset;
 
 	// VU0 and VU1 memory.
-	static const u32 VUmemOffset   = 0x08000000;
+	static constexpr u32 VUmemOffset   = 0x08000000 + _128mbOffset;
 
 	// EE recompiler code cache area (64mb)
-	static const u32 EErecOffset   = 0x10000000;
+	static constexpr u32 EErecOffset   = 0x10000000 + _128mbOffset;
 
 	// IOP recompiler code cache area (16 or 32mb)
-	static const u32 IOPrecOffset  = 0x14000000;
+	static constexpr u32 IOPrecOffset  = 0x14000000 + _128mbOffset;
 
 	// newVif0 recompiler code cache area (16mb)
-	static const u32 VIF0recOffset = 0x16000000;
+	static constexpr u32 VIF0recOffset = 0x16000000 + _128mbOffset;
 
 	// newVif1 recompiler code cache area (32mb)
-	static const u32 VIF1recOffset = 0x18000000;
+	static constexpr u32 VIF1recOffset = 0x18000000 + _128mbOffset;
 
 	// microVU1 recompiler code cache area (32 or 64mb)
-	static const u32 mVU0recOffset = 0x1C000000;
+	static constexpr u32 mVU0recOffset = 0x1C000000 + _128mbOffset;
 
 	// microVU0 recompiler code cache area (64mb)
-	static const u32 mVU1recOffset = 0x20000000;
+	static constexpr u32 mVU1recOffset = 0x20000000 + _128mbOffset;
 
 	// Bump allocator for any other small allocations
 	// size: Difference between it and HostMemoryMap::Size, so nothing should allocate higher than it!
-	static const u32 bumpAllocatorOffset = 0x24000000;
+	static constexpr u32 bumpAllocatorOffset = 0x24000000 + _128mbOffset;
 }
 
 // --------------------------------------------------------------------------------------
--- a/common/Dependencies.h
+++ b/common/Dependencies.h
@@ -249,6 +249,7 @@ static const s64 _8mb = _1mb * 8;
 static const s64 _16mb = _1mb * 16;
 static const s64 _32mb = _1mb * 32;
 static const s64 _64mb = _1mb * 64;
+static const s64 _128mb = _1mb * 128;
 static const s64 _256mb = _1mb * 256;
 static const s64 _1gb = _1mb * 1024;
 static const s64 _4gb = _1gb * 4;
--- a/pcsx2-qt/Settings/AdvancedSystemSettingsWidget.ui
+++ b/pcsx2-qt/Settings/AdvancedSystemSettingsWidget.ui
@@ -53,6 +53,13 @@
         </property>
        </widget>
       </item>
+      <item row="0" column="2">
+       <widget class="QCheckBox" name="ee128mbram">
+        <property name="text">
+         <string>Enable 128 MB of RAM</string>
+        </property>
+       </widget>
+      </item>
       <item row="2" column="1">
        <widget class="QCheckBox" name="eeINTCSpinDetection">
         <property name="text">
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -361,7 +361,8 @@ struct Pcsx2Config
 			PreBlockCheckEE : 1,
 			PreBlockCheckIOP : 1;
 		bool
-			EnableEECache : 1;
+			EnableEECache : 1,
+			EnableEE128mbRam : 1;
 		BITFIELD_END
 
 		RecompilerOptions();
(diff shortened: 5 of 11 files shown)


refractionpcsx2:
Did you get permission from Vaser to take his code?

ThirteenAG:
> Did you get permission from Vaser to take his code?

It's in this post: https://github.com/xTVaser/pcsx2-rr/issues/102#issuecomment-1078426319

refractionpcsx2:
Yep, I saw that after, thanks. Just wanted to be sure.

xTVaser (approved):
This looks good from my perspective -- but I defer to someone like @tellowkrinkle over the general changes to the memory management -- I have no idea if it has other implications.

I was able to boot games with configurations that _require_ 128mb to load properly.  Without the option enabled they crashed and with it enabled they worked!  I only tested the current wx GUI and not Qt.

Thanks for taking the time to get this merged, looks a lot easier than I was expecting.  Just curious -- do you think it would be possible to make this an arbitrary amount in the future?  Doesn't seem too far off, and I imagine for the use-case you're going for a configurable amount would be more useful than just 4x as much (although maybe that's plenty?)

tellowkrinkle (approved):

pcsx2-qt/Settings/AdvancedSystemSettingsWidget.cpp
```diff
@@ -33,6 +33,7 @@ AdvancedSystemSettingsWidget::AdvancedSystemSettingsWidget(SettingsDialog* dialo
 
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.eeRecompiler, "EmuCore/CPU/Recompiler", "EnableEE", true);
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.eeCache, "EmuCore/CPU/Recompiler", "EnableEECache", false);
+	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.ee128mbram, "EmuCore/CPU/Recompiler", "EnableEE128mbRam", false);
```
stenzek, line 36: This will cause problems if it's used for per-game configurations, since the config variable will get updated, but the memory map won't reflect that. Same deal with live configuration changes. The way to work around that would be to not read the config in `CHECK_128MBRAM`, but instead some other flag latched at memory map init time.

Personally, I don't think the option should be exposed as a global setting at all, and instead only set per-game. Less risk of users breaking their setup completely then. But you'd have to solve the aforementioned memory map issue, perhaps by saving memory state, resetting, and restoring.

stenzek (changes requested):

pcsx2/System.h
```diff
@@ -42,41 +42,43 @@ class RecompiledCodeReserve;
...
 	// The actual addresses may not be equivalent to Base + Offset in the event that allocation at Base failed
 	// Each of these offsets has a debugger-accessible equivalent variable without the Offset suffix that will hold the actual address (not here because we don't want code using it)
 
 	// PS2 main memory, SPR, and ROMs
-	static const u32 EEmemOffset   = 0x00000000;
+	static constexpr u32 EEmemOffset   = 0x00000000; // 128mb
 
 	// IOP main memory and ROMs
-	static const u32 IOPmemOffset  = 0x04000000;
+	static constexpr u32 IOPmemOffset  = 0x04000000 + _128mbOffset;
```
stenzek, line 56: This will also break anyone's cheat tables if modifying anything in IOP memory, regardless of whether the option is enabled or not.

ThirteenAG: I don't see any difference in addresses between pcsx2 builds with and without 128mb:
image
Seems like an unnecessary complication?

pcsx2-qt/Settings/AdvancedSystemSettingsWidget.cpp
ThirteenAG: > This will cause problems if it's used for per-game configurations, since the config variable will get updated, but the memory map won't reflect that. Same deal with live configuration changes. 

It won't cause problems, since the toggle is only there as a "storefront" so to say, to let users know that option exists. I specifically used it only with debugger functions, to have some "visible" changes. Another thing is external access, for example someone wants to redirect game entry point to somewhere safe, which is only after >32mb memory. He can check this option, make sure user allowed pcsx2 to use more memory and make necessary changes. With this disabled, user can have vanilla experience.

My proposal would be to make toggle greyed out with 
`m_check_EE128mbEnable->Enable(!configToApply.EnablePresets && !g_GameStarted && !g_GameLoading);`

pcsx2/System.h
stenzek: I'm not talking about the guest addresses. I'm talking about the offset to IOP memory in the main PCSX2 memory map, which is often used by people with programs like cheat engine.

pcsx2-qt/Settings/AdvancedSystemSettingsWidget.cpp
stenzek: `!g_GameStarted && !g_GameLoading`: These are set on the emu thread, and not safe to access from the UI thread.

> He can check this option, make sure user allowed pcsx2 to use more memory and make necessary changes. With this disabled, user can have vanilla experience.

pcsx2/System.h
stenzek: No. People like to use the base pointer to the memory map (or EEmem), and then offset that. They should use the IOPmem pointer instead, but plenty of modders seem to rely on fixed addresses from the start of the memory map.

pcsx2-qt/Settings/AdvancedSystemSettingsWidget.cpp
ThirteenAG: I see, reverted. Perhaps no toggle at all then?

pcsx2/System.h
ThirteenAG: > That's why I suggested relocating the additional memory to the end of the memory map, keeping the start of IOP memory the same.

I don't quite see how to reflect that in the code, mind pushing this change?

stenzek: Add an additional space between the end of IOP memory and the recompiler code area, and special case the physical EE pages above 32MB in VTLB when setting up the page pointers.

ThirteenAG: Wait, but wouldn't that mean the offset of that memory will be different then?
I need `EEMainMemoryStart + offset` external writings too.

ThirteenAG: Plus it's not the first time memory layout changes, I don't see why some abstract cheat engine users that use EE Mem to access IOP should be thought about here. Isn't it much better to provide an API to access the addresses of these memory regions?

stenzek: > Wait, but wouldn't that mean the offset of that memory will be different then?
> I need EEMainMemoryStart + offset external writings too.

Yes, that's what I was referring to (but it's only relevant for IOP memory, because the EE memory is still 0-32MB).

> Isn't it much better to provide an API to access the addresses of these memory regions?

There is one. You can use IPC, or the `EEmem` and `IOPmem` pointers. But they don't.

ThirteenAG: > But they don't.

Then they can PR this change if it's necessary, I'd rather keep the memory whole, better for external access and pattern matching and whatnot.

stenzek: So you're not willing to address any of the concerns I've raised? (race conditions, live configuration changes, resetting, per-game configuration, memory map changes), despite me suggesting multiple methods of how it could be solved?

If not, then this discussion is pointless, and I'm removing myself/won't reply any further.

ThirteenAG: I'm sorry, I just don't enough about the subject and don't understand some concerns 
- like, per-game configuration, memory is already allocated on pcsx2 startup, what's there related specifically to the game?
- memory map changes, "fixing" iop pointer that is used by 3rd party, but making EE memory two regions, requiring to find two pointers now? Remember that almost no games will use that memory and the whole point of this PR is to make it available for external tools, which then may result to more pull requests to pcsx2 which utilize it.
My solution is to remove the toggle and have more memory always.
Would be great if you could help code wise in a different pull request or this one with your vision of how the code should look like. Because I only need more memory, not this concrete implementation.

stenzek: Race condition: accessing GameLoaded/GameStarted on the UI thread is unsafe. Because it can change between opening the settings dialog and when the user interacts with it.

Per-game configuration: The Qt interface lets you set near any setting in a per-game profile, and those settings are applied when the ELF is loaded. Which in your setting's case, means that the 128MB flag won't reflect the TLB virtual state.

Live configuration changes: Same deal, except changing the setting while the game is running, which you didn't address in Qt. The "safe" way to do this, like I said, would be to reinitialize both the physical and virtual VTLB mappings when the 128MB flag changes. But then you risk messing with any mappings the game happened to put in this area.

Save states: 
1. You're breaking save state compatibility (which isn't that huge of a deal, but when it happens, users tend to get annoyed.
2. What happens if someone saves a state with 128MB enabled, and loads with it disabled? Or vice versa?

Memory map changes: As I've stated repeatedly now, you're moving the start of IOP further up the memory map, and that affects anyone with hardcoded offsets anywhere above 32MB.

Furthermore, perhaps more concerningly, this affects the physical side of the VTLB mapping; you're unconditionally physically mapping 128MB instead of 32MB, which means that the upper 96MB will be accessible to games regardless instead of raising an exception. I'm not aware of anything which necessarily depends on this, but it does mean debugging issues may be more difficult, as code "running away" due to a bug will take much longer to TLB miss, if at all. See https://github.com/PCSX2/pcsx2/blob/master/pcsx2/Memory.cpp#L152

ThirteenAG: > Race condition: accessing GameLoaded/GameStarted on the UI thread is unsafe. Because it can change between opening the settings dialog and when the user interacts with it.

I reverted that change. Let's remove the toggle next.

> Per-game configuration: The Qt interface lets you set near any setting in a per-game profile, and those settings are applied when the ELF is loaded. Which in your setting's case, means that the 128MB flag won't reflect the TLB virtual state.

> Live configuration changes: Same deal, except changing the setting while the game is running, which you didn't address in Qt. The "safe" way to do this, like I said, would be to reinitialize both the physical and virtual VTLB mappings when the 128MB flag changes. But then you risk messing with any mappings the game happened to put in this area.

Let's remove the toggle.

> Save states:

> You're breaking save state compatibility (which isn't that huge of a deal, but when it happens, users tend to get annoyed.
> What happens if someone saves a state with 128MB enabled, and loads with it disabled? Or vice versa?

Let's remove the toggle. Plus, in my opinion, the possibilities that more memory brings well worth some broken save states.

> Memory map changes: As I've stated repeatedly now, you're moving the start of IOP further up the memory map, and that affects anyone with hardcoded offsets anywhere above 32MB.
I have hardcoded offsets right now too that will break if you do this change preserving IOP offset. I have zero clue why would anyone need IOP. But when I used EE mem offset that was hardcoded to 0x20000000 in x86 pcsx2, nobody notified me when it changed. Then it changed again in x64 version, nobody told me either. From my perspective, things like this happen all the time. I understand your explanation, but don't understand how to implement it, so I won't.

> Furthermore, perhaps more concerningly, this affects the physical side of the VTLB mapping; you're unconditionally physically mapping 128MB instead of 32MB, which means that the upper 96MB will be accessible to games regardless instead of raising an exception. I'm not aware of anything which necessarily depends on this, but it does mean debugging issues may be more difficult, as code "running away" due to a bug will take much longer to TLB miss, if at all. See https://github.com/PCSX2/pcsx2/blob/master/pcsx2/Memory.cpp#L152

¯\_(ツ)_/¯

refractionpcsx2: > I understand your explanation, but don't understand how to implement it, so I won't.
Then learn

ThirteenAG: > > I understand your explanation, but don't understand how to implement it, so I won't.
> > Then learn

I prefer the way it is now. Complicating code to preserve a pointer that nobody should even access the way they allegedly do now, is not worth the effort for me, sorry.

ThirteenAG:
Should I open an issue then?

refractionpcsx2:
If you want somebody to implement it properly, an issue is the way to go, yes.

refractionpcsx2:
I meant in a way that's going to cause less problems by somebody actually willing to make changes, instead of the "I dunno how to do it, so I'm not going to" attitude.

ThirteenAG:
There's no proof of "problems".

refractionpcsx2:
People who know what they're doing telling you there's problems should be enough. But if you don't want to listen to seasoned emulator developers, well then that's your issue.

PR #5760, opened 2022-03-25
GS: Add automatic trilinear filtering level

stenzek:
### Description of Changes

The auto level will use the value in the game database if selected, otherwise the user's choice takes priority.

Based on discussion with @RedDevilus.

### Rationale behind Changes

Allows trilinear to be forced on via gamedb, and if we ever get to the point where we can default it on instead of off, it'll happen transparently.

### Suggested Testing Steps

Check behaviour matches expectations for gamedb overrides, e.g. Rogue Galaxy.

### DIFF
--- a/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
@@ -146,7 +146,9 @@ GraphicsSettingsWidget::GraphicsSettingsWidget(SettingsDialog* dialog, QWidget* 
 	SettingWidgetBinder::BindWidgetToIntSetting(sif, m_ui.texturePreloading, "EmuCore/GS", "texture_preloading",
 		static_cast<int>(TexturePreloadingLevel::Off));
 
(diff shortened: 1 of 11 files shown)


RedDevilus:
The emulator seems to correctly override the GameDB with the GUI changes.
However it seems trilinear itself seems not work anymore:

Trilinear forced should make these letters a purplish hue:
image

Bilinear (forced) gave that same effect:
image

refractionpcsx2:
Trilinear is working as expected for me, tested in x3 in Vulkan

RedDevilus:
Seems it's OpenGL being broken.

refractionpcsx2:
Okay so, I suspect the Vulkan is incorrect with Trilinear, but this is separate from the PR.

PR thing, works fine.

Vulkan Trilinear/Bilinear issue which I think has been mentioned before.

Example of Trilinear in Vulkan:

image

Example in OpenGL:
image

As you can notice, the filtering is less "pixely" in OGL, same for DX.

stenzek:
Everything should be consistent between renderers now.

Trilinear forced will also force bilinear (and disable the dropdown on Qt).

Per discussion on Discord, we probably should combine these into one `Texture Filtering` option at some point in the future.

PR #5768, opened 2022-03-27
Qt: Reduce max SPU - Change Speed Preset

RedDevilus:
### Description of Changes

SPU2 max latency changed from 3000ms (3 seconds) to 200 ms (0.2 seconds) value.
### Rationale behind Changes

Qt has a leftover from the old SPU2-X values which is 3000 ms (3 seconds) latency meaning that if an user sets it to the max it will do something like this: Videoframe ~150 / 180 happens and it makes the first sound after 3 seconds for the first videoframe which is far too long. Also changed 1% speed as it messed with the GSdump player and too stuttery to be usable in other usecases. The lower ms the better but Cubeb should handle it fine and ignore the default 100 ms to a very low number even if your computer is weak:

image

So in best case the audio latency is 0.01 seconds and in worst case 0.025 seconds which is neglible.

### Suggested Testing Steps

Check if the GUI responds normally.

### DIFF
--- a/pcsx2-qt/Settings/AudioSettingsWidget.ui
+++ b/pcsx2-qt/Settings/AudioSettingsWidget.ui
@@ -180,7 +180,7 @@
            <number>15</number>
           </property>
           <property name="maximum">
-           <number>3000</number>
+           <number>200</number>
           </property>
           <property name="value">
            <number>100</number>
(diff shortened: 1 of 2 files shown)


JordanTheToaster (approved):

stenzek (approved):
LGTM

Issue #5709, opened 2022-03-20
[BUG]: Knockout Kings 2001 missing graphics (Regression)


refractionpcsx2:
### Describe the Bug

Most of the graphics are missing ingame if you load the game using the HW renderer, starting in SW and changing once ingame is fine, was broken by #4094

Tagging @iMineLink 

### Reproduction Steps

Load the blockdump, keep hammering X and Start to skip videos and go ingame with HW mode active, witness missing everything :P

### Expected Behavior

It doesn't break :P

### PCSX2 Revision

Anything post 1.7.2125

### Operating System

Windows 10 (64bit)

### CPU

Ryzen 3900X

### GPU

RTX 2080 TI

### GS Settings

Any hardware mode, everything else irrelevant (afaik)

### GS Window Screenshots

![unknown](https://user-images.githubusercontent.com/6278726/159143016-af082ad7-c435-496f-b4eb-896792c67b77.png)

### Logs & Dumps

https://drive.google.com/file/d/11f_QhaqJ-aWuFFIY1d7hRIN2QLqntfdK/view?usp=sharing

RedDevilus:
So this game worked by accident and needs Disable Partial Invalidation (Older name was Fast Texture Invalidation)

PR #4747, opened 2021-09-08
Savestates: Small refactoring

MrCK1:
### Description of Changes
- Organized user-created savestates by saving into subfolders with game name and serial.
- Detect PSX CRC properly, format console titlebar.

- Fix bug with g_GameStarted and g_GameLoading not being reset on recompiler shutdown.
- Fix bug with g_GameStarted and g_GameLoading not being set to proper status when loading savestate.

### Rationale behind Changes
Cleaner directory for savestates, improve ease of saveestate management. Improve management of game state variables in regards to functionality.

### Suggested testing steps
Save/load savestates, look for errors in directory/savestate creation or directory naming. Check for errors/crashes switching games.

Closes #4625, #4744

PR #5771, opened 2022-03-27
GameDB: Serial with ':' to '-' + GS HW and other Fixes

RedDevilus:
### Description of Changes

GameDB serials with ':' replaced with ' - '
+
Spacing consistent for comments
+
GS HW Fixes and other fixes for:
- Adventures of Cookie & Cream, The (Fixes https://github.com/PCSX2/pcsx2/issues/5779)
- Brothers in Arms - Earned in Blood / Road to Hill 30
- Black
- Chaos Legion
- God Hand
- Knockout Kings 2001 (Fixes https://github.com/PCSX2/pcsx2/issues/5709)
- Kuon
- Project Eden
- Psi-Ops - The Mindgate 
- Punisher, The 
- Ratchet Deadlocked (USA) / Gladiator (Europe) / 3 Up Your Arsenal
- Silent Hills Origins / Shattered Memories / 3 / 4
- SoulCalibur II / III

### Rationale behind Changes

Windows doesn't like you to use ':' in folders this caused issues for when CK1 did savestates in folders and now it's also messing with unable to add covers in Qt so better to replace them and also to avoid other issues now and the future.

WX:
https://github.com/PCSX2/pcsx2/pull/4747
Relevant comment:
image

Qt:
image
image

### Suggested Testing Steps

Test relevant games.


RedPanda4552:
Is the true solution to this problem to go into GameIndex.yaml and modify every entry which may legitimately have a ":" in it? Or is it to look at code which is writing to the file system and make it stop writing illegal characters?

RedDevilus:
I just avoid the issue that it it's going through illegal characters, most of the GameDB entries already use '-' vs about 200 entries containing ':' so it also streamlines the GameDB.

RedPanda4552:
":" is not an illegal character for a game name though. Some may legitimately use it. So my angle is, how feasible is it to instead of overwriting GameIndex.yaml to correct for code that doesn't do error checking, write code that does error checking? If the names in GameIndex.yaml have to match names on the file system then I also have to call into question why we are using names as unique identifiers for files rather than serials or CRCs. Maybe I've been out of the loop but I don't see how display names having illegal file system characters can be an issue for something which, to my knowledge, shouldn't be written to the file system.

RedDevilus:
> ":" is not an illegal character for a game name though. Some may legitimately use it. So my angle is, how feasible is it to instead of overwriting GameIndex.yaml to correct for code that doesn't do error checking, write code that does error checking? If the names in GameIndex.yaml have to match names on the file system then I also have to call into question why we are using names as unique identifiers for files rather than serials or CRCs. Maybe I've been out of the loop but I don't see how display names having illegal file system characters can be an issue for something which, to my knowledge, shouldn't be written to the file system.

It's fine if you launch a game sure but it's when you want to make a cover for Qt, or an older version of the gamestate revision that copied the game name of the GameDB as the folder. So instead the solution was to completely ignore the GameDB which to me isn't a solution either and sure this may not be the best approach but I don't see a problem revising a minority of entries.

You can check it here
[GameSerialsPCSX2-27March2022.txt](https://github.com/PCSX2/pcsx2/files/8358553/GameSerialsPCSX2-27March2022.txt)

Like if we just look at Eyetoy most of them are using '-' and only a few use ':' in the name. I'm not against error checking perhaps we can combine that with the PR.

Blackbird88:
While you're at it, could you please add **[Demo]** to **Gran Turismo - Nissan Micra Edition**. I forgot to do it while adding it and it'd be silly to make whole PR for such small thing 😃

RedDevilus:
Silent Hills Origins;
Old:
image
New:
image

Lost track of how many GS HW Fixes I have added.

RedDevilus:
Other examples that should fix these games;

Ratchet 3 Up Your Arsenal:
Ratchet3UpYourArsenal
Ratchet Deadlocked (USA) Gladiator (Europe):
RatchetDeadlockedUSAGladiatorEurope

RedDevilus:
Black;
Old:
image
New:
image

RedDevilus:
Note: In future we should do error-checking on those special characters in case someone modifies it or there is an edge-case.

RedDevilus:
Project eden;

Old:
image

New:
image

RedDevilus:
SoulCalibur 3 clip provided from Discord user:


https://user-images.githubusercontent.com/24227051/160715774-c15cdbf1-e16d-4723-9a81-0d49ad7abea0.mov

RedDevilus:
Kuon;
MTVU on:
image

MTVU off:
image

RedDevilus:
Seem Soul Calibur II has the same issue as the third entry but to a lesser degree.

RedDevilus:
Thanks Git for making no sense.

Issue #5558, opened 2022-02-22
[BUG]: Command line options are not working with the Appimage


Ka-hu:
### Describe the Bug

I'm using **pcsx2-v1.7.2357-linux-AppImage-64bit.AppImage** but it seems the command line options are not working with the Appimage format. I tried to launch an ISO directly from command line, but it just opens the GUI. Also, I tried other options (--nogui for example) but they don't work either. Every time only the GUI opens and I have to launch the game from there. I'm using Arch Linux. I tried earlier Appimage versions but had the same problem.

### Reproduction Steps

1. Download an Appimage
2. Launch a game from command line with the full path to the ISO file
3. It opens the GUI instead of running the game

### Expected Behavior

It should launch the game (and parse other command line options too).

### PCSX2 Revision

v1.7.2357

### Operating System

Linux (64bit) - Specify Distro Below

### If Linux - Specify Distro

Arch Linux

qurious-pixel:
replacing https://github.com/PCSX2/pcsx2/blob/4e7ade8612b921c7fac22a8771bc3da4f8a6e159/.github/workflows/scripts/linux/AppRun#L3

with `$APPDIR/AppRun-patched "$@"` will pass the arguments.

7oxicshadow:
I have also confirmed that modifying AppRun-patched to add "$@" so that the parameters are passed though works well for me. Would be good if we could get a fix for this in the future

brooksytech:
Hello, I am also having this issue.

Mrlinkwii:
is this still an issue on latest master?

qurious-pixel:
Qt AppRun passes the commandline arguments

PR #5777, opened 2022-03-28
Allow commandline params to be passed through to the patched apprun file

7oxicshadow:
### Description of Changes
Modified AppRun to support passing of command line parameters 
I have (with some difficulty) run the .github/workflows/scripts/linux/appimage.sh script to confirm that the correct file gets modified within the appimage file.

### Rationale behind Changes
Appimage uses a wrapper that doesn't pass the parameters to the AppRun-patched file
See Bug #5558

### Suggested Testing Steps
Download appimage
Execute passing iso to launch.
PCSX2 will open without loading iso

### DIFF
--- a/.github/workflows/scripts/linux/AppRun
+++ b/.github/workflows/scripts/linux/AppRun
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-$APPDIR/AppRun-patched
+$APPDIR/AppRun-patched "$@"


F0bes:
@qurious-pixel could you review?

TheGag96:
Just encountered this issue. Would be really helpful for creating individual game launchers for the Steam Deck (outside of the FlatHub version, not sure if that works either).

brooksytech:
Can we get this reviewed and merged? As I am also having the issue of the arguments not being passed. TY!

arcum42 (approved):

PR #5778, opened 2022-03-29
BiosTools: (Experimental) BIOS region patching

F0bes:
# * This has nothing to do with region locking and whatever DRM SCE used*

### Description of Changes

Allows the user to patch their BIOS region by setting the region byte during ROM upload.

These are the results with my BIOS when going to OSDSYS

<table>
<tr>
	<td> T10K
	<td>TLB miss
<tr>
	<td>Test
	<td>TLB miss
<tr>
	<td>Japan
	<td>Works (On some BIOSes OSDSYS will be Japanese, buttons are always switched)
<tr>
	<td>USA
	<td>Works
<tr>
	<td>Europe
	<td>Works
<tr>
	<td>HK
	<td>Works (English text, is this expected?)
<tr>
	<td>Free
	<td>TLB miss
<tr>
	<td>China
	<td>TLB miss. Infinite rebooting on some BIOSes.
</table>

Of course each bios revision is unique and you may get different results.

### Rationale behind Changes
The recommended BIOS dumper ends up patching slim BIOSes as Japan when it's USA for example. 

### Suggested Testing Steps
Anything pertaining to emulation issues while patching aren't a feasible issue to solve.

### DIFF
--- a/pcsx2/VMManager.cpp
+++ b/pcsx2/VMManager.cpp
@@ -573,6 +573,7 @@ bool VMManager::ApplyBootParameters(const VMBootParameters& params)
 {
 	const bool default_fast_boot = Host::GetBoolSettingValue("EmuCore", "EnableFastBoot", true);
 	EmuConfig.UseBOOT2Injection = params.fast_boot.value_or(default_fast_boot);
+
 	s_elf_override = params.elf_override;
 	s_disc_path.clear();
 
--- a/pcsx2/ps2/BiosTools.h
+++ b/pcsx2/ps2/BiosTools.h
@@ -28,6 +28,23 @@ struct BiosDebugInformation
 	u32 threadListAddr;
 };
 
+static const char* BiosZoneStrings[] {
+	"T10K",
+	"Test",
+	"Japan",
+	"USA",
+	"Europe",
+	"HK",
+	"Free",
+	"China",
+	nullptr
+};
+
+static const char* BiosZoneBytes[]
+{
+	"T", "X", "J", "A", "E", "H", "P", "C", nullptr
+};
+
 extern BiosDebugInformation CurrentBiosInformation;
 extern u32 BiosVersion;		// Used by CDVD
 extern u32 BiosRegion;		// Used by CDVD
@@ -41,4 +58,3 @@ extern std::string BiosPath;
 extern bool LoadBIOS();
 extern bool IsBIOS(const char* filename, u32& version, std::string& description, u32& region, std::string& zone);
 extern bool IsBIOSAvailable(const std::string& full_path);
-
--- a/pcsx2-qt/Settings/BIOSSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/BIOSSettingsWidget.cpp
@@ -34,7 +34,11 @@ BIOSSettingsWidget::BIOSSettingsWidget(SettingsDialog* dialog, QWidget* parent)
 	m_ui.setupUi(this);
 
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.fastBoot, "EmuCore", "EnableFastBoot", true);
+	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.patchRegion, "EmuCore", "PatchBios", false);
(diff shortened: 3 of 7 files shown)


RedDevilus:
Hongkong used to be a British Colony till 1997 so it would make sense it's in English. As far as I know it's working as intended.

F0bes:
I also made a script you can use to patch the bios. It makes a backup of your bios but things can go wrong so be careful and backup!

https://gist.github.com/F0bes/1224ca17a8e71a0a5f93be9ced23b14b

F0bes:
Ditched the WX side of things. I can't make to configuration systems work at the same time, plus there's no reason to.
Should be ready.

Swat842:
The China bios doesn't do Infinite rebooting at all its just a black screen my friend tested it an told me that it won't boot at all just wanted to let you know that so it can be corrected.

refractionpcsx2:
> The China bios doesn't do Infinite rebooting at all its just a black screen my friend tested it an told me that it won't boot at all just wanted to let you know that so it can be corrected.

is that an issue with this PR? or has your friend just got trouble trying to use a Chinese BIOS?

Swat842:
> > The China bios doesn't do Infinite rebooting at all its just a black screen my friend tested it an told me that it won't boot at all just wanted to let you know that so it can be corrected.
> 
> is that an issue with this PR? or has your friend just got trouble trying to use a Chinese BIOS?

No he didn't get in trouble at all he has the ROM2 an it wont boot at all something is wrong with the BIOS... Either it wont work or it was a bad dump

F0bes:
> > > The China bios doesn't do Infinite rebooting at all its just a black screen my friend tested it an told me that it won't boot at all just wanted to let you know that so it can be corrected.
> > 
> > 
> > is that an issue with this PR? or has your friend just got trouble trying to use a Chinese BIOS?
> 
> No he didn't get in trouble at all he has the ROM2 an it wont boot at all something is wrong with the BIOS... Either it wont work or it was a bad dump

Okay well it's unrelated to this PR. The PR makes the BIOS think it's a different region. It doesn't affect the compatibility with BIOSes.

pcsx2/ps2/BiosTools.h
```diff
@@ -28,6 +28,23 @@ struct BiosDebugInformation
...
+	"Europe",
+	"HK",
+	"Free",
+	"China",
+	nullptr
+};
+
+static const char* BiosZoneBytes[]
+{
+	"T", "X", "J", "A", "E", "H", "P", "C", 0x0
```
stenzek, line 45: Strictly speaking this should be `nullptr` on the end (since it's a pointer array).

pcsx2/VMManager.cpp
```diff
@@ -573,6 +573,9 @@ bool VMManager::ApplyBootParameters(const VMBootParameters& params)
 {
 	const bool default_fast_boot = Host::GetBoolSettingValue("EmuCore", "EnableFastBoot", true);
 	EmuConfig.UseBOOT2Injection = params.fast_boot.value_or(default_fast_boot);
+	EmuConfig.PatchBios = Host::GetBoolSettingValue("EmuCore", "PatchBios", false);
```
stenzek, line 576: No need to load the setting here, since it's part of the config (and will get loaded in `Pcsx2Config`).

pcsx2/VMManager.cpp
```diff
@@ -573,6 +573,9 @@ bool VMManager::ApplyBootParameters(const VMBootParameters& params)
...
+	EmuConfig.PatchRegion = Host::GetStringSettingValue("EmuCore", "PatchRegion", BiosZoneBytes[0]);
```
stenzek, line 577: Probably should move this one's loading to the config (`SettingsWrapEntry`).

RedDevilus:
Shouldn't there be a mechanism to back-up the unmodified bios? Just like how memcards create a clone and convert that instead for protection and to avoid unforeseen freezes, blue screens or other situations.

F0bes:
When using the system provided in the emulator, the BIOS on disc is not touched. It only modifies what is uploaded to the guest memory.

RedDevilus:
That's fine then.

PR #5788, opened 2022-03-30
Various Gamefix code changes

TheLastRar:
### Description of Changes
Fixes saving the `EnableGameFixes` option in Qt
Moves application of Game Fixes into GameEntry (where GS Hardware fixes are currently applied)
Adds logging similar to GS Hardware Fixes for Game Fixes

### Rationale behind Changes
Fix Bugs
Removes code duplication
Centralises code for applying Game/GS fixes into one file

### Suggested Testing Steps
Test if game fixes apply (or don't if disabled) in both Qt and Wx
Check toggling Game Fixes saves in Qt

### DIFF
--- a/pcsx2/GameDatabase.cpp
+++ b/pcsx2/GameDatabase.cpp
@@ -18,6 +18,7 @@
 #include "GameDatabase.h"
 #include "Host.h"
 #include "Patch.h"
+#include "vtlb.h"
 
 #include "common/FileSystem.h"
 #include "common/Path.h"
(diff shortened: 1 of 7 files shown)


RedDevilus (changes requested):
Automatic gamefixes does nothing on WX and Qt even though it says fixes in the Program Log titlebar but not the log itself. Doing the manual gamefixes like the FMV switch to software also doesn't work.

pcsx2/GameDatabase.cpp
```diff
@@ -325,6 +326,118 @@ bool GameDatabaseSchema::isUserHackHWFix(GSHWFixId id)
 	}
 }
 
+u32 GameDatabaseSchema::GameEntry::applyGameFixes(Pcsx2Config& config) const
+{
+	// Only apply core game fixes if the user has enabled them.
+#ifdef PCSX2_CORE
+	const bool apply_auto_fixes = config.EnableGameFixes;
+#else
+	const bool apply_auto_fixes = config.EnablePatches;
```
Mrlinkwii, line 335: i think you meant to do this ?
```suggestion
	const bool apply_auto_fixes = config.EnableGameFixes;
```

TheLastRar: The Wx version doesn't have `EnableGameFixes`, `EnablePatches` controls both patches and Gamefixes

The Qt version (which uses PCSX2_CORE) introduced `EnableGameFixes`, hence the ifdef statement

TheLastRar:
> Automatic gamefixes does nothing on WX and Qt even though it says fixes in the Program Log titlebar but not the log itself. Doing the manual gamefixes like the FMV switch to software also doesn't work.

Can you test with the latest push

RedDevilus:
> > Automatic gamefixes does nothing on WX and Qt even though it says fixes in the Program Log titlebar but not the log itself. Doing the manual gamefixes like the FMV switch to software also doesn't work.
> 
> Can you test with the latest push?

I'll give it a quick look when it's built.

pcsx2/GameDatabase.cpp
stenzek: Could either just do it at the call site instead (VMManager), or make the whole routine `#ifdef PCSX2_CORE`, since it's not used in wx.

RedDevilus:
Still an issue:
Manual gamefixes + automatic gamefixes don't work on WX and Qt.

RedDevilus:
Hmm never mind it seems the FMV software switch seems to not work for me. So automatic settings work partially on my test.

Though I'm having a WX issue where I click apply and it's not doing anything (notice the grayed out button at bottom right)

image

TheLastRar:
I'm not able to reproduce the issue of manually setting the FMV software gamefix not working, it switches to SW mode for me in both Wx and Qt

I don't _think_ any of my changes would have broken a single fix anyway.

I'm also not sure what's triggering your issue with the greyed out button, It becomes clickable for me after changing a setting, but it seems like that isn't happening for you?

TheLastRar:
Actually, it seems that the FMV software gamefix doesn't work with the Automatic render

RedDevilus:
I'll make it an issue atleast but yeah same as on master.

RedDevilus:
> Alright
> 
> What about your Wx issue, are there any steps to reproduce this? I've not figured out how to trigger it While I don't think it's related to this pr, I would like to confirm that one way or the other

Seems I can't reproduce it anymore.

refractionpcsx2:

TheLastRar:
> What is the status of this PR?

I've acted on Stenzek's suggestion in discord
The issues reported I _believe_ are unrelated to this PR

Should be good to go assuming a code review has been performed.

refractionpcsx2:
Code looks good to me, also tested it and it seems to work fine :)

PR #5791, opened 2022-03-31
releases: simplify discord announcements, and link to the right page

xTVaser:
### Description of Changes

- Link to the new full website instead of the temporary page.
- Simplify the message, all links are organized on the website

### Rationale behind Changes

- Full site is live, the temporary github.io one needs to stop being linked to so it can be shut down.


refractionpcsx2:
Does this remove the changelog line? It would be good to keep that.

xTVaser:
Nope that should be kept still

PR #5802, opened 2022-04-01
GS-hw: Purge custom resolution entirely.

lightningterror:
### Description of Changes

GS-hw: Purge custom resolution entirely.

Code debt, not used anymore, get rid of it.
### Rationale behind Changes

Custom resolution is not used anymore, get rid of it.
### Suggested Testing Steps

Make sure upscale multiplier still works.

### DIFF
--- a/pcsx2/GS/GS.cpp
+++ b/pcsx2/GS/GS.cpp
@@ -1369,8 +1369,6 @@ void GSApp::Init()
 	m_default_configuration["PrecacheTextureReplacements"]                = "0";
 	m_default_configuration["preload_frame_with_gs_data"]                 = "0";
 	m_default_configuration["Renderer"]                                   = std::to_string(static_cast<int>(GSRendererType::Auto));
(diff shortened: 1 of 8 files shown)


JordanTheToaster (approved):

refractionpcsx2 (approved):

Tokman5:
why UpscaleMultiplier will be limited 8x?
imo 10x is needed for 8k resolution, even though you don't need to add such high res settings at the moment.

refractionpcsx2:
Upscaling would be broken entering custom upscales like that since we don't handle them internally (it needs to be done for each upscale), and unless you have an 8k monitor there isn't much point.  8x is plenty enough.

iMineLink:
Do we still need to support custom x/y float scaling in textures after this?

lightningterror:
> Do we still need to support custom x/y float scaling in textures after this?

We can merge the two.

PR #5806, opened 2022-04-03
EERec: Reset reliability improvements

stenzek:
### Description of Changes

**iR5900: Exit on NeedsReset not IsReset**

This is an edge case which can be hit in Goemon, but also when changing settings in Qt, as they will be applied at guest vsync time, which is called within the JIT.

Also, do the reset before entering JIT code, rather than when the first block is compiled.

**EERec: Purge PCSX2_SEH define**

No longer needed since we don't have 32-bit.

**VMManager: Make EE rec toggles reliable**

Before, it'd swap out the Cpu while it was executing... This one only affects Qt, but now you can toggle the recompiler on and off while the game is running, and enjoy your framerate tanking.

### Rationale behind Changes

Reliability.

### Suggested Testing Steps

Make sure Goemon still works as intended, test toggling CPU options while the VM is running in Qt.

### DIFF
--- a/pcsx2/Config.h
+++ b/pcsx2/Config.h
@@ -393,6 +393,8 @@ struct Pcsx2Config
 		void LoadSave(SettingsWrapper& wrap);
 		void ApplySanityCheck();
 
+		bool CpusChanged(const CpuOptions& right) const;
+
 		bool operator==(const CpuOptions& right) const
 		{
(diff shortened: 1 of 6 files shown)


refractionpcsx2:
Goemon is still good :) Fucked with a bunch of settings while running emulation in WX and QT, all worked as expected.

PR #5807, opened 2022-04-03
GS: Support compiling without OpenGL renderer

stenzek:
### Description of Changes

Mainly relevant for Macs, but also if someone ever does a UWP port.

### Rationale behind Changes

Reducing binary size when it's unnecessary.

### Suggested Testing Steps

Try building with `-DUSE_OPENGL=OFF`, and make sure everything works.

### DIFF
--- a/pcsx2/GS/GSUtil.cpp
+++ b/pcsx2/GS/GSUtil.cpp
@@ -190,14 +190,30 @@ CRCHackLevel GSUtil::GetRecommendedCRCHackLevel(GSRendererType type)
 
 GSRendererType GSUtil::GetPreferredRenderer()
 {
-#ifdef __APPLE__
+#if defined(__APPLE__)
+	// Mac: Prefer Metal hardware.
 	return GSRendererType::Metal;
-#endif
-#ifdef _WIN32
+#elif defined(_WIN32)
+#if defined(ENABLE_OPENGL)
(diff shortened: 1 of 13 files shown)


refractionpcsx2:
Windows seems good :)

pcsx2/GS/GSUtil.cpp
```diff
@@ -190,14 +190,30 @@ CRCHackLevel GSUtil::GetRecommendedCRCHackLevel(GSRendererType type)
...
 {
-#ifdef __APPLE__
+#if defined(__APPLE__)
+	// Mac: Prefer Metal hardware.
 	return GSRendererType::Metal;
-#endif
-#ifdef _WIN32
+#elif defined(_WIN32)
+#if defined(ENABLE_OPENGL)
+	// Windows: Prefer GL if available.
```
auscompgeek, line 198:
```suggestion
	// Windows: Prefer DX11 if available.
```

refractionpcsx2: Why? GL is better..

stenzek: No. DX11 has the worst accuracy on windows. If it wasn't for old GPUs, I'd make Vulkan the default. But regardless, the goal of this PR isn't to change the defaults.

PR #5812, opened 2022-04-03
GS-hw: Further optimize blending equations.

lightningterror:
### Description of Changes

GS-hw: Further optimize blending equations.
Allow to run free sw blend with no texture barriers, more speed for all renderers.

CLAMP 1, negative result will be clamped to 0.
Condition 1:
(0  - Cs)*Alpha +  0, (0  - Cd)*Alpha +  0
Condition 2:
Alpha is either As or F higher than 1.0f
(0  - Cd)*Alpha  + Cd, (0  - Cs)*F  + Cs

### Rationale behind Changes

More brr when using blending.
### Suggested Testing Steps

Test blending.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
@@ -2604,6 +2604,21 @@ void GSRendererHW::EmulateBlending(bool& DATE_PRIMID, bool& DATE_BARRIER, bool& 
 		m_conf.ps.blend_a = 0;
 		m_conf.ps.blend_b = 0;
 	}
+	else if (m_env.COLCLAMP.CLAMP && m_conf.ps.blend_a == 2
+		&& (m_conf.ps.blend_d == 2 || (m_conf.ps.blend_b == m_conf.ps.blend_d && (alpha_c0_high_min_one || alpha_c2_high_one))))
+	{
+		// CLAMP 1, negative result will be clamped to 0.
+		// Condition 1:
+		// (0  - Cs)*Alpha +  0, (0  - Cd)*Alpha +  0
+		// Condition 2:
+		// Alpha is either As or F higher than 1.0f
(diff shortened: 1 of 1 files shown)


pcsx2/GS/Renderers/HW/GSRendererNew.cpp
```diff
@@ -553,13 +553,25 @@ void GSRendererNew::EmulateBlending(bool& DATE_PRIMID, bool& DATE_BARRIER, bool&
 	// Get alpha value
 	const bool alpha_c0_zero = (m_conf.ps.blend_c == 0 && GetAlphaMinMax().max == 0);
 	const bool alpha_c0_one = (m_conf.ps.blend_c == 0 && (GetAlphaMinMax().min == 128) && (GetAlphaMinMax().max == 128));
+	const bool alpha_c0_high_min_one = (m_conf.ps.blend_c == 0 && GetAlphaMinMax().min > 128);
```
refractionpcsx2, line 556: unused

lightningterror: Oh, actually I should use that one instead.

refractionpcsx2: also are you sure you don't want >= ?

lightningterror: I do, good catch.

lightningterror:
With the changes cleaned up and done properly I wasn't able to find any test case that hits the code, maybe not worth committing.

labib2002:
I'm sorry but I've seen "more brr" in more PRs than I can count.
what does this mean?

refractionpcsx2:
Goes faster.

refractionpcsx2:
Do you wanna run this through the GS Dump runner (after a rebase)? see if anything changes

lightningterror:
I'll just merge it instead, should be fine, since I didn't manage to find a game that uses it.

PR #5813, opened 2022-04-03
PCSX2: Interlacing to Deinterlacing

RedDevilus:
### Description of Changes

Interlacing string to Deinterlacing strings
### Rationale behind Changes

Consistency and correctness
### Suggested Testing Steps

Test all the renderers


RedDevilus:
Not crazily confident that I did it correct, if it's needed I can just keep it to the minimum for the values for GameDB, GameDB itself and Documentation.

RedDevilus:
Note to fix the labeler but PR should work for WX and Qt now.

pcsx2/GS/GS.cpp
```diff
@@ -716,20 +715,14 @@ void GSgetStats(std::string& info)
...
 {
-	const char* api_name = HostDisplay::RenderAPIToString(s_render_api);
-	const char* hw_sw_name = (GSConfig.Renderer == GSRendererType::Null) ? " Null" : (GSConfig.UseHardwareRenderer() ? " HW" : " SW");
-	const char* deinterlace_mode = theApp.m_gs_interlace[static_cast<int>(GSConfig.InterlaceMode)].name.c_str();
-
-#ifndef PCSX2_CORE
 	int iwidth, iheight;
 	GSgetInternalResolution(&iwidth, &iheight);
 
-	info = StringUtil::StdStringFromFormat("%s%s | %s | %dx%d", api_name, hw_sw_name, deinterlace_mode,  iwidth, iheight);
```
stenzek, line 727: Looks like a merge conflict gone wrong here (wx and qt should emit different info).

pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
```diff
@@ -82,7 +82,7 @@ GraphicsSettingsWidget::GraphicsSettingsWidget(SettingsDialog* dialog, QWidget*
 		sif, m_ui.aspectRatio, "EmuCore/GS", "AspectRatio", Pcsx2Config::GSOptions::AspectRatioNames, AspectRatioType::R4_3);
 	SettingWidgetBinder::BindWidgetToEnumSetting(sif, m_ui.fmvAspectRatio, "EmuCore/GS", "FMVAspectRatioSwitch",
 		Pcsx2Config::GSOptions::FMVAspectRatioSwitchNames, FMVAspectRatioSwitchType::Off);
-	SettingWidgetBinder::BindWidgetToIntSetting(sif, m_ui.interlacing, "EmuCore/GS", "interlace", DEFAULT_INTERLACE_MODE);
+	SettingWidgetBinder::BindWidgetToIntSetting(sif, m_ui.interlacing, "EmuCore/GS", "deinterlace", DEFAULT_INTERLACE_MODE);
```
stenzek, line 85: Would it be worth leaving the config variable as-is, so not to break people's existing configurations? Although this one's probably low risk, I don't think many people change deinterlacing from auto.

pcsx2/GS/GS.cpp
RedDevilus: Yeah didn't know what I did, probably didn't pay attention I did see the issue but not the cause.

PR #5816, opened 2022-04-04
Input Recording: Begin the process of achieving parity in Qt

xTVaser:
### Description of Changes

Initial work of porting input recording over to Qt.  This is still rough and not feature complete, several things need re-writing and cleaning up.  But I'm hopeful that I can work on this iteratively while Qt is still in a state of flux.

https://user-images.githubusercontent.com/13153231/161457126-09c1fa05-4910-459d-8df5-5859ac6187fb.mov

#### Known Issues
- No frame advance yet
- OSD logs are not all perfect
- Virtual Pads not yet implemented
- Settings action/menu not yet implemeneted
- General UX issues

### Rationale behind Changes

I made the decision to duplicate the input recording code and hide the wxWidgets version behind the `PCSX2_CORE` define.  This allows me to essentially freeze the wxWidgets version, which means while Qt is in development there is no need to regression test it (I can focus my efforts only on Qt).  

Input recording's code has been very stable since last years work, I don't anticipate needing to do any backporting.

### Suggested Testing Steps

I've done some basic testing as shown above, I would suggest testing to make sure I havn't broken something else and unrelated in the process.


pcsx2-qt/EmuThread.cpp
```diff
@@ -889,4 +890,20 @@ DEFINE_HOTKEY("ToggleFullscreen", "General", "Toggle Fullscreen", [](bool presse
...
+		g_InputRecordingControls.RecordModeToggle();
+	}
+})
+// TODO - Vaser - the way the pause/resuming used to work is broken on Qt, needs a rewrite. 
+//	- Currently if you frame advance you can't get out of frame advancing!
+DEFINE_HOTKEY("InputRecFrameAdvance", "Input Recording", "Frame Advance", [](bool pressed) {
+	if (!pressed) // ?? - not pressed so it is on key up?
+	{
+		g_InputRecordingControls.FrameAdvance();
+		g_InputRecordingControls.ResumeCoreThreadIfStarted();
```
stenzek, line 906: There shouldn't be any mention of "core thread" in Qt code, it doesn't exist.

pcsx2-qt/MainWindow.cpp
```diff
@@ -1014,6 +1025,85 @@ void MainWindow::onLoggingOptionChanged()
 	QtHost::UpdateLogging();
 }
 
+void MainWindow::onInputRecNewActionTriggered()
+{
+	const bool wasPaused = VMManager::GetState() == VMState::Paused;
```
stenzek, line 1030: Not thread-safe; use `m_vm_valid` and  `m_vm_paused`.

pcsx2-qt/MainWindow.cpp
```diff
@@ -1014,6 +1025,85 @@ void MainWindow::onLoggingOptionChanged()
...
+	{
+		VMManager::SetPaused(false);
+	}
+}
+
+#include "pcsx2/Recording/InputRecordingControls.h"
+
+void MainWindow::onInputRecPlayActionTriggered()
+{
+	const bool wasPaused = VMManager::GetState() == VMState::Paused;
```
stenzek, line 1061: Ditto

pcsx2-qt/MainWindow.cpp
```diff
@@ -1014,6 +1025,85 @@ void MainWindow::onLoggingOptionChanged()
...
+
+	if (!wasPaused)
+		g_InputRecordingControls.PauseImmediately();
```
stenzek, line 1064: Not sure what this function does exactly, but I saw it being manipulated on the emu thread as well. If you need to call it from the UI thread, it should go through `EmuThread`, with a queued connection slot (see the other methods).

pcsx2-qt/input-rec/NewInputRecordingDlg.h
```diff
@@ -15,31 +15,38 @@
 
 #pragma once
 
-#include <QtWidgets/QDialog>
+#include "ui_NewInputRecordingDlg.h"
```
stenzek, line 18: Nitpick, but `input-rec` as a folder name doesn't match the rest of the directory structure.

stenzek (changes requested):

pcsx2-qt/EmuThread.cpp
```diff
@@ -889,4 +890,20 @@ DEFINE_HOTKEY("ToggleFullscreen", "General", "Toggle Fullscreen", [](bool presse
...
+	{
+		g_InputRecordingControls.RecordModeToggle();
+	}
+})
+// TODO - Vaser - the way the pause/resuming used to work is broken on Qt, needs a rewrite. 
+//	- Currently if you frame advance you can't get out of frame advancing!
+DEFINE_HOTKEY("InputRecFrameAdvance", "Input Recording", "Frame Advance", [](bool pressed) {
+	if (!pressed) // ?? - not pressed so it is on key up?
+	{
+		g_InputRecordingControls.FrameAdvance();
```
stenzek, line 905: IMO, frame advance shouldn't be part of input recording, it has use cases outside of it as well.

pcsx2-qt/MainWindow.cpp
```diff
@@ -44,6 +44,8 @@
 #include "Settings/InterfaceSettingsWidget.h"
 #include "SettingWidgetBinder.h"
 #include "svnrev.h"
+#include <input-rec/NewInputRecordingDlg.h>
```
stenzek, line 47:
```suggestion
#include "input-rec/NewInputRecordingDlg.h"
```

(and put it in alphabetical order)

pcsx2-qt/EmuThread.cpp
xTVaser: will be cleaned up when the class is cleaned up, the naming has not been touched yet.

pcsx2-qt/MainWindow.cpp
xTVaser: What does `m_vm_valid` correspond to?  Looks like `true` corresponds to the vm being started/starting -- but is that the intended useage in the long-term?

pcsx2-qt/MainWindow.cpp
xTVaser: I'll keep that in mind when I refactor the class, once again just replicating existing behaviour for this initial PR.

pcsx2-qt/input-rec/NewInputRecordingDlg.h
xTVaser: In what way? because it has a symbol or because its lower-case? I assume because of the symbol because there are folders with all lower-case and mixed-case currently.

pcsx2-qt/EmuThread.cpp
xTVaser: removed as you are apparently PR'ing the feature.  I agree, and im glad a proper pause feature has also been added to the main emulator's functions.

pcsx2-qt/MainWindow.cpp
stenzek: There's `Host::RunOnCPUThread()` which may be helpful here, but I never got around to implementing it in Qt. Will do that at some point, as well as the inverse (`RunOnUIThread()`).

pcsx2-qt/input-rec/NewInputRecordingDlg.h
stenzek: In Qt, everything is named TitleCase. (e.g. `Settings`).

pcsx2-qt/MainWindow.cpp
stenzek: They're set when the state changes, based on a signal emitted by the VM thread. That way, you don't need to read emu thread state, which can risk changing at any time, and opens you to race conditions.

`m_vm_valid` means that a VM exists (is initialized), but any actual VMManager calls should check that it's still valid when the code actually runs on the VM thread (see the wrapper methods in EmuThread which forward to VMManager).
`m_vm_paused` is what the title says; it's paused at the time of call. Remember that these are set via signals, so at the time you check it, the VM thread may have already paused (and the UI thread hasn't processed the signal to update the bool yet).

pcsx2-qt/input-rec/NewInputRecordingDlg.h
xTVaser: k, i switched mine to be title-case as well then.  Does that mean this directory should be renamed? https://github.com/PCSX2/pcsx2/tree/master/pcsx2-qt/resources

stenzek: Probably should, yeah.

PR #5820, opened 2022-04-04
R5900: Make CPU exits consistent and safe (+ Qt frame advance)

stenzek:
### Description of Changes

**R5900: Make CPU exits consistent and safe**

Previously, we would either throw an exception (ints), or longjmp out of the recompiler when the execution state was checked. Unfortunately for our stability, this happened at the end of the frame, just before it was pushed to the GS, and in the middle of processing EE events (!).

Doing so not only meant that we executed a bunch of event  testing/exception code twice (once after we paused, again when we resumed), but it also could potentially leave things in an inconsistent state.

So instead, let's do it safely with a flag, replacing the old `iopBreakpoint` flag, so there's no additional overhead on the hot path.

**VMManager: Add frame advance hotkey**

Now that exits are consistent, this is easy. Can advance a user-specified number of frames, but the hotkey defaults to 1.

### Rationale behind Changes

Stability, more features for Qt.

### Suggested Testing Steps

Make sure the debugger still works as intended, in particular IOP breakpoints.

Test frame advance and pausing in Qt.


Ziemas:
IOP breakpoints seem to deadlock the emulator.

stenzek:
> IOP breakpoints seem to deadlock the emulator.

Should be fine now, I ended up redoing how they were handled on the EE side completely. As an added bonus, the EE rec + IOP int configuration no longer blows up when you use bps.

Ziemas:
Confirmed working now :+1:

PR #5837, opened 2022-04-07
LnxMisc: Use clock_gettime() instead of gettimeofday()

stenzek:
### Description of Changes

`gettimeofday()` is not monotonically increasing, and can be subject to clock changes. `clock_gettime()` also has nanosecond resolution.

### Rationale behind Changes

Correctness.

### Suggested Testing Steps

Check Linux build still works as intended. Check Mac build, because apparently it's buggy/nonexistant there?

### DIFF
--- a/common/Linux/LnxMisc.cpp
+++ b/common/Linux/LnxMisc.cpp
@@ -41,14 +41,14 @@ void InitCPUTicks()
 
 u64 GetTickFrequency()
 {
-	return 1000000; // unix measures in microseconds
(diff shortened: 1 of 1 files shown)


tellowkrinkle (approved):
macOS uses a `mach_absolute_time`-based implementation in DarwinMisc so it shouldn't be affected

Ziemas:
Seems fine on linux

Mrlinkwii (approved):

PR #5838, opened 2022-04-08
Qt: Add Gzipped + Fix interlacing

RedDevilus:
### Description of Changes

Allows booting .gz roms inside the Qt GUI toolkit. (Also added a tiny regression fix)
### Rationale behind Changes

WX supported this, I don't see why not add to Qt. Though no idea why anyone wants to use it above CSO or CHD if you look at filesize.
### Suggested Testing Steps

Test any .gz

Don't do .iso.gz:
image

### DIFF
--- a/pcsx2-qt/MainWindow.cpp
+++ b/pcsx2-qt/MainWindow.cpp
@@ -46,11 +46,12 @@
 #include "svnrev.h"
 
 static constexpr char DISC_IMAGE_FILTER[] =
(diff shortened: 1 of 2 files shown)


RedDevilus:
If you want to do a quick test on windows:
You can do it via right clicking your iso when having 7z installed go to the tab and select add to archive:
image

Changing to gzip archive format:
image

And removing the .iso part:
image

Press ok and wait

stenzek (approved):

PR #5844, opened 2022-04-11
wxWidgets 3.1.6 Compatibility

tellowkrinkle:
### Description of Changes
Makes PCSX2 compile properly on wxWidgets 3.1.6

### Rationale behind Changes
It's nice when things compile

### Suggested Testing Steps
Try compiling with wxWidgets 3.1.6

PR #5848, opened 2022-04-11
Publish macOS builds

tellowkrinkle:
### Description of Changes
Updates Mac build script to make publishable builds and enables build publishing

Note: Don't merge until after #5792, #5807, and #5844 are merged

### Rationale behind Changes
Official Mac builds

### Suggested Testing Steps
Not sure how to test the publishing system without merging to master, if anyone knows how please test


.github/workflows/macos-workflow.yml
```diff
@@ -115,3 +124,55 @@ jobs:
...
+          ARTIFACT_NAME=""
+          if [ ${{ github.event_name }} == "pull_request" ]; then
+            PR_SHA=$(git rev-parse --short "${{ github.event.pull_request.head.sha }}")
+            ARTIFACT_NAME="PCSX2"
+            if [ ! -z "${{ github.event.pull_request.number }}" ]; then
+              PR_NUM=${{ github.event.pull_request.number }}
+              ARTIFACT_NAME="${ARTIFACT_NAME}-pr[${PR_NUM}]"
+            fi
+            ARTIFACT_NAME="${ARTIFACT_NAME}-sha[${PR_SHA}]"
+            if [ ! -z "${{ github.event.pull_request.title }}" ]; then
```
Margen67, line 141:
```suggestion
            if [ -n "${{ github.event.pull_request.title }}" ]; then
```

.github/workflows/macos-workflow.yml
```diff
@@ -115,3 +124,55 @@ jobs:
...
+
+      - name: Prepare Artifact Metadata
+        id: artifact-metadata
+        shell: bash
+        run: |
+          ARTIFACT_NAME=""
+          if [ ${{ github.event_name }} == "pull_request" ]; then
+            PR_SHA=$(git rev-parse --short "${{ github.event.pull_request.head.sha }}")
+            ARTIFACT_NAME="PCSX2"
+            if [ ! -z "${{ github.event.pull_request.number }}" ]; then
```
Margen67, line 136:
```suggestion
            if [ -n "${{ github.event.pull_request.number }}" ]; then
```

.github/workflows/macos-workflow.yml
```diff
@@ -115,3 +124,55 @@ jobs:
       - name: Run Tests
         working-directory: build
         run: make -j$(getconf _NPROCESSORS_ONLN) unittests
+
+      - name: Prepare Artifact Metadata
+        id: artifact-metadata
+        shell: bash
+        run: |
+          ARTIFACT_NAME=""
```
Margen67, line 132:
```suggestion
```

.github/workflows/macos-workflow.yml
```diff
@@ -115,3 +124,55 @@ jobs:
       - name: Run Tests
         working-directory: build
         run: make -j$(getconf _NPROCESSORS_ONLN) unittests
+
+      - name: Prepare Artifact Metadata
+        id: artifact-metadata
+        shell: bash
```
Margen67, line 130: This is the default shell.
```suggestion
        shell: bash
```

refractionpcsx2:
All the above are merged, so you can merge this when you're ready to publish them.

.github/workflows/macos-workflow.yml
```diff
@@ -115,3 +124,53 @@ jobs:
...
+          cp "${{ steps.artifact-metadata.outputs.name }}.tar.gz" ci-artifacts/macOS.tar.gz
+          ls
+
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ steps.artifact-metadata.outputs.name }}
+          path: ./*.tar.gz
+
+      - name: Upload Assets and Potential Publish Release
```
xTVaser, line 166: i believe this is going to make the release asset poorly named -- in the linux/windows steps there is an additional step that names them cleaner than the current PR/Commit artifact https://github.com/PCSX2/pcsx2/blob/master/.github/workflows/windows-workflow.yml#L188-L194

Additionally, should the name of the artifact include the version of macOS that is supported? I see the CI is using big sur.

Also this can be tested on a fork before merged, i can do so.

.github/workflows/macos-workflow.yml
```diff
@@ -115,3 +124,53 @@ jobs:
...
+          tar cvzf "${{ steps.artifact-metadata.outputs.name }}.tar.gz" "${{ steps.artifact-metadata.outputs.name }}.app"
+          mkdir ci-artifacts
+          cp "${{ steps.artifact-metadata.outputs.name }}.tar.gz" ci-artifacts/macOS.tar.gz
+          ls
+
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ steps.artifact-metadata.outputs.name }}
+          path: ./*.tar.gz
```
Margen67, line 164:
```suggestion
          path: *.tar.gz
```

.github/workflows/macos-workflow.yml
```diff
@@ -115,3 +124,53 @@ jobs:
...
+
+      - name: Upload Assets and Potential Publish Release
+        if: github.repository == 'PCSX2/pcsx2' && startsWith(github.ref, 'refs/tags/')
+        env:
+          GITHUB_TOKEN: ${{ secrets.BOT_PAT }}
+          ASSET_DIR: ${{ github.WORKSPACE }}/ci-artifacts
+          ASSET_EXTENSION: tar.gz
+          TAG_TO_SEARCH_FOR: ${{ github.REF }}
+        run: |
+          cd ./.github/workflows/scripts/releases/upload-release-artifacts
```
Margen67, line 174:
```suggestion
          cd .github/workflows/scripts/releases/upload-release-artifacts
```

.github/workflows/macos-workflow.yml
tellowkrinkle: The whole point of doing this is to make the build support all the way back to our minimum version (10.13 at the moment).  If we wanted to only support Big Sur and up, we could continue using Homebrew binaries.

I name it `macOS.tar.gz` since we only support one architecture and one SIMD ISA.  If that changes in the future, we could rename it, though ideally we'd make a fat binary and shove it all in one download.  Does that seem reasonable?

xTVaser: The name of the asset will be used to make it clear to the user on the frontend what version it supports.  I'm not saying it should be labelled as big sur (although if something is ever merged that breaks on 10.13, aka high sierra?, it's probably going to go unnoticed).

If you are targetting high sierra, then we should label it as such.

tellowkrinkle: The compiler is pretty good about making sure we keep compatibility, ignoring 10.13-specific bugs, so I don't expect compatibility breaks to happen very often.

If we're going to label macOS 10.13, should we making the Windows one "Windows 8.1"?

xTVaser: Yes I've thought about that argument.  From my experience with macOS apps, they usually label what version they support clearly -- I suppose this has something to do with how frequently macOS versions break previous app compatibility, but it seems to be the norm.

I guess its ultimately up to you, I have no experience building for macOS.  But the current naming this PR will commit will break the backing API's expectations as seen above -- I imagined that we would we labelling it with the version given my personal experience.

tellowkrinkle: I looked around at some other applications.  Chrome shows versions for both.  Opera does for neither.  Dolphin does for neither.  Sublime Text does for neither.  VSCode does for neither.  IntelliJ does neither.

What application were you thinking of that only shows a version for the macOS release?

xTVaser: Basically almost everyone I can remember downloading for my own mac 🤷 clearly it's not as standardized as I thought.

xTVaser (changes requested):
The `paths-ignore` here needs to be removed, and the workflow needs to run on tag pushes for the process to work https://github.com/PCSX2/pcsx2/pull/5848/files#diff-e24d0c032fd3ded7d8b2f5704cf5e79e808d7577926e58a1f9334c51c9c48decR9

That was the only change i had to make here - https://github.com/xTVaser/pcsx2-rr/commit/558f2c786836d2fa29ee12361a684df137f33960

Which was able to create a successful release - https://github.com/xTVaser/pcsx2-rr/releases/tag/v1.4.2

xTVaser (approved):
This looks good and should work after merging.  I'll follow up on the rest of the work to get the mac builds pushed through the rest of the process (on the downloads page).

tellowkrinkle:
Okay.  I just realized I need to make sure we put up a decent error if you try to run on a non-Metal-supporting mac, so I won't merge quite yet

Issue #5851, opened 2022-04-11
Regression: Capcom games suffering from lines on right part of screen


prafullpcsx2:
### Describe the Bug

Although this bug only happens while upscaling but since this is a regression I am posting this bug report.

Capcom games like Onimusha 3, Onimusha 4, Devil may Cry 3 and may be others show graphical artifacts in shape of lines on right side of the screen. These vary from game to game but location is almost same. My testing suggests that this bug started happening after https://github.com/PCSX2/pcsx2/pull/5387 was merged. 

Using round sprite Full reduces the problem but doesn't fix it completely.

### Reproduction Steps

Run these games and they suffer from the mentioned bug during some cutscenes/ingame. I have attached multiframe gsdumps so testing becomes easier.

### Expected Behavior

Game should run without any graphical issues.

### PCSX2 Revision

All versions after v1.7.2300 suffer from this problem

### Operating System

Windows 10 (64bit)

### CPU

i3-9100f

### GPU

750Ti

### GS Settings

Default settings with or without upscaling hacks.

### Logs & Dumps

http://filedropper.com/UvZnto1X

Dreadmoth:
Shadow of Rome is also affected by this issue:

[Shadow_of_Rome_right_edge_line.zip](https://github.com/PCSX2/pcsx2/files/8468682/Shadow_of_Rome_right_edge_line.zip)

tellowkrinkle:
Okay I looked at Shadow of Rome
Here's a draw **on a build not affected by this problem**: 
<img width="1536" alt="image" src="https://user-images.githubusercontent.com/3315070/163697230-da93c0ce-e8a9-4251-9eb5-6ca1e14f1e34.png">

Game renders a sprite to a 128x56 target from (28664, 30972) to (30712, 31868) with XYOFFSET (28672, 30976).  Convert that to pixel coordinates and you get (-0.5, -0.25) to (127.5, 55.75).  This means that it never touches the rightmost 0.5 of a pixel and the bottommost 0.25 of a pixel, which become whole pixels when upscaling.

So the game has always been broken, but our broken samplers were covering this up by trimming the right and bottom edges off textures.

IMO the proper solution to this is upscaling hacks that affect where the game draws, rather than how we sample textures, but curious of what others think @refractionpcsx2

refractionpcsx2:
[...]

PR #5387, opened 2022-01-23
GS: Enable small framebuffers at 1x

tellowkrinkle:
### Description of Changes
- Stops the texture cache from downscaling targets when used as a source
- Enables smaller framebuffers even when rendering at 1x (previously only enabled at 2x+)
- Adds some optimizations to move sampling to hardware when possible (minor perf improvement on GT4)

Note: I did this by editing the vertex and fragment shaders to support rescaling normalized coordinates.  An alternative way to do it would be to just allocate full scale textures in this case.  Preferences?  (sampling the target directly mentioned below would need this approach)
Note2: This changes the renderer's interpretation of a texture whose `size / scale` does not match `tw` and `th` from "stretch to fit" to "use a 1-1 pixel scale".  Not sure if this breaks anything?  If it does, maybe those things should have set their scale properly...

### Rationale behind Changes
- Improves speed of Burnout at 1x on integrated GPUs with low memory bandwidth due to its access patterns making TC convert framebuffers to and from depth formats constantly (by reducing the size of those framebuffers).
- Improves the image quality of Burnout when upscaled to be the multiplier you selected and not that times 80% or so then nearest-neighbor upscaled back to 100%
- Should make removing the extra copies we do when using a target as a source a bit easier (future improvements, not included in this PR)

Example at 2x upscale:
ExampleA
ExampleB

### Suggested Testing Steps
Test 1x and 2x upscale for lots of games

PR #5853, opened 2022-04-11
GS: Support upscaled region repeat

tellowkrinkle:
### Description of Changes
- Add support for upscaling region repeat
- Adjust region clamp coordinates for upscaling

### Rationale behind Changes
More upscaling is nice

### Suggested Testing Steps
- Make sure my changes to region clamp didn't make things worse.  I don't think they would, but who knows
- Test DX and GL (I only tested Metal and Vk)
- Test upscaled region repeat.  If you don't know of a game to test, comment out GSRendererHW:1373-1378 and GSRendererHW:1387-1391 and [run this trace](https://github.com/PCSX2/pcsx2/files/8468210/transformers_pixelation_pr5387.gs.xz.zip)

### DIFF
--- a/bin/resources/shaders/dx11/tfx.fx
+++ b/bin/resources/shaders/dx11/tfx.fx
@@ -201,46 +201,31 @@ float4 clamp_wrap_uv(float4 uv)
 	else
 		tex_size = WH.xyxy;
 
-	if(PS_WMS == PS_WMT)
+	if(PS_WMS == 2)
 	{
-		if(PS_WMS == 2)
-		{
-			uv = clamp(uv, MinMax.xyxy, MinMax.zwzw);
-		}
-		else if(PS_WMS == 3)
-		{
-			#if PS_FST == 0
-			// wrap negative uv coords to avoid an off by one error that shifted
-			// textures. Fixes Xenosaga's hair issue.
-			uv = frac(uv);
-			#endif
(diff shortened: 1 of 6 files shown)


refractionpcsx2:
Apparently this makes #5851 worse? I assume it was meant to address that.

tellowkrinkle:
If it's the last commit causing it, I mostly wanted to see what that *did* affect, and had no specific games in mind when I changed it

It makes it more in line with what our hw sampler would do to a similar upscale, which is nice as I don't have to worry about side effects of swapping between the two.

The change is to how clamping is handled.  The PS2 clamps with min and max values, rather than min and max+1.  This means that if you upscale a 4x4 texture to 16x16, the equivalent max value goes from 3 to 12, rather than 3 to 15.  Obviously if you use a hw sampler, clamping goes to the edge of the texture, which will of course be 15.  #5851 being affected by both this and #5387 makes sense though, since #5387 is what enabled clamping with the hw sampler when available, which this adjustment is attempting to match in the sw sampler.

seta-san:
can we please rebase this? there are a couple games I'd like to test it with.

TellowKrinkle:
It reveals a whole bunch more upscaling issues in games, so it's waiting on some improved upscale hacks which I keep not getting around to working on

PR #5854, opened 2022-04-12
WxCleanup: Purge wx from save states and pnach handling, use zstd compression

stenzek:
### Description of Changes

 - Replaces wx zip stream usage with libzip.
 - Gets rid of wx in pnach handling, and simplify parsing.
 - Get rid of wx in SYSTEM.CNF parsing, sadly it's still everywhere else in CDVD.

Save states now use zstd compression instead of deflate. This is **much** faster:
```
deflate:
Saving state took 4252.671 ms
Loading state took 207.523 ms

zstd:
Saving state took 244.878 ms
Loading state took 125.824 ms
```

### Rationale behind Changes

Getting rid of wx, chunk by chunk.

### Suggested Testing Steps

As this touches a fair bit of code, it'll need thorough testing. Things which need to be tested:
 - Save state loading/saving. Old states should be loadable, but new states won't be loadable in older builds.
 - Test booting a range of games to make sure the SYSTEM.CNF parser still works as intended.
 - Test games which have patches in the game database.
 - Test using pnach files in cheats and cheats_ws directories.


pcsx2/Patch.cpp
```diff
@@ -114,188 +107,116 @@ static int PatchTableExecute(const ParsedAssignmentString& set, const PatchTextT
...
-{
-	wxULongLong_t l;
-	str.ToULongLong(&l, base);
-	return l;
+		const std::optional<std::string> pnach_data(FileSystem::ReadFileToString(fd.FileName.c_str()));
+		if (!pnach_data.has_value())
+			continue;
+
+		const int loaded = LoadPatchesFromString(pnach_data.value());
+		PatchesCon->WriteLn((loaded ? Color_Green : Color_Gray), "Loaded %d %s from '%s'.",
```
CookiePLMonster, line 198:
```suggestion
		PatchesCon->WriteLn((loaded ? Color_Green : Color_Gray), "Loaded %d %s from '%.*s'.",
```

Wrong format string explodes debug builds rn

JordanTheToaster (approved):

PR #5870, opened 2022-04-15
GL/ContextEGL: Fix broken surface creation

stenzek:
Regression from #5676.

### DIFF
--- a/common/GL/ContextEGL.cpp
+++ b/common/GL/ContextEGL.cpp
@@ -313,8 +313,10 @@ namespace GL
 		}
 		configs.resize(static_cast<u32>(num_configs));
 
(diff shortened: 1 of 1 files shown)


tcoyvwac:
Thanks @stenzek , Sorry, `EGLConfig check_config` does indeed copy! :sweat:

stenzek:
The actual issue was assigning the resolved config to a local, not the class member (which is used later in surface creation). I just got rid of the rvalue reference there because it didn't make sense for a pointer type.

tcoyvwac:
Ahh, so [the variable was being shadowed](https://github.com/PCSX2/pcsx2/blob/master/common/GL/ContextEGL.h#L59). :ninja:

PR #5881, opened 2022-04-15
GS: Get display width for frame when offsets off

### LINKED ISSUE #5871: [BUG]: Mortal Kombat Armageddon - stretched and cut beyond visible area except movies
### Describe the Bug

Any non FMV frame is stretched and cut beyond visible area. Changing aspect ratio in settings doesn't fix anything. Zooming out shows that the frame is cut.

### Reproduction Steps

Look at any frame that's not FMV.

### Expected Behavior

_No response_

### PCSX2 Revision

v1.7.2605

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

### CPU

Ryzen 5 1600

### GPU

GTX 970

### GS Settings

_No response_

### Emulation Settings

_No response_

### Logs & Dumps

[snaps.zip](https://github.com/PCSX2/pcsx2/files/8495992/snaps.zip)

refractionpcsx2:
### Description of Changes
Calculate how much of the screen the picture should take up in real terms when screen offsets are disabled.

### Rationale behind Changes
some games take a large framebuffer such as Mortal Kombat Armageddon, where it renders 841 pixels and fits it in to like, 632 pixels on the screen due to the magnification, but when Screen Offsets are disabled, it doesn't account for this, but it needs to be restricted as some games try to render too much.

### Suggested Testing Steps
Test games with screen offsets disabled, especially known PCRTC problematic ones, make sure everything fits on the screen properly.

Fixes #5871 and situations like it.

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -416,6 +416,20 @@ GSVector4i GSState::GetFrameMagnifiedRect(int i)
 	return rectangle;
 }
 
+int GSState::GetDisplayHMagnification()
+{
+	// Pick one of the DISPLAY's and hope that they are both the same. Favour DISPLAY[1]
+	for (int i = 1; i >= 0; i--)
+	{
+		if (IsEnabled(i))
+			return m_regs->DISP[i].DISPLAY.MAGH + 1;
+	}
+
+	// If neither DISPLAY is enabled, fallback to resolution offset (should never happen)
+	const GSVideoMode videomode = GetVideoMode();
+	return VideoModeDividers[(int)videomode - 1].x + 1;
+}
+
 GSVector4i GSState::GetDisplayRect(int i)
 {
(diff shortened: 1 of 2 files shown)


JordanTheToaster (approved):

Anuskuss:
Why is PTRTC offsets even a config option? Why shouldn't it just be enabled permanently?

refractionpcsx2:
> Why is PTRTC offsets even a config option? Why shouldn't it just be enabled permanently?

I'd love to, but there's 2 problems

1. some games have bad offsets (Sony's advice was "Just play with it until you're happy") so the game ends up going off the screen in annoying ways.

2. People WILL complain about there being single black borders down one side (due to how some offsets are) or borders around their entire game.

So yeah, I'm leaving it with the old behaviour with the option to use the correct (but problematic) mode.

Here's an example of one of the annoying offset contributing to my not wanting it always on.

Normal:
image

With Offsets:
image

PR #5897, opened 2022-04-17
build: Remove i386_flag, as it is no longer used.

arcum42:
### Description of Changes
Remove i386_flag, as it is no longer used.

### Rationale behind Changes
I removed where i386_flag was used previously, and didn't notice this was no longer used.

### Suggested Testing Steps
Compile with build.sh.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -21,11 +21,6 @@ set -u
 set_ncpu_toolfile()
 {
     ncpu=$(getconf NPROCESSORS_ONLN 2>/dev/null || getconf _NPROCESSORS_ONLN)
-    if [ "$(uname -s)" = 'Darwin' ]; then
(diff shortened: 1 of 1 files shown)


tellowkrinkle:
You still need the `ncpu` for make

arcum42:
Ah, missed that line. I'll change that.

arcum42:
Taken care of, though I wonder if I should just move that line to set_make, thinking about it.

refractionpcsx2:
Seems a bit pointless having a function of its own now.

arcum42:
Unless anyone's noticed any issues, this should be ready to merge. I moved the ncpu code to set_make a few days back and got rid of the separate function.

PR #5905, opened 2022-04-17
GS: Clean up some code and correct heights when offset by 1

refractionpcsx2:
### Description of Changes
Cleans up some variable usage and corrects the height of the frame if the game is using small offsets to "blur" the picture.

### Rationale behind Changes
Tidy some things up, but also having slightly offset DISPLAY registers caused the height to be odd (like 449 instead of 448) due to blurring gamedevs put in to smooth the picture (which we remove).

### Suggested Testing Steps
Run games, make sure everything looks okay and that it doesn't crash.

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -380,7 +380,7 @@ GSVector4i GSState::GetFrameMagnifiedRect(int i)
 	if (!IsEnabled(0) && !IsEnabled(1))
 		return rectangle;
 
-	const GSVideoMode videomode = GetVideoMode();
+	const int videomode = static_cast<int>(GetVideoMode()) - 1;
 	const auto& DISP = m_regs->DISP[i].DISPLAY;
 	const bool ignore_offset = !GSConfig.PCRTCOffsets;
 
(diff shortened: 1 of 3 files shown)


refractionpcsx2:
@SamRH could you possibly test this and make sure nothing crashes, since this will reduce the height by 1. Thanks :)

SamRH:
Release build with LTO enabled on Linux using the Vulkan renderer. Fired up a couple of the games from the Jak and Daxter series (1 & 2) + Kelly Slaters Pro Surfer. All ran and looked fine.

refractionpcsx2:
Ace :) Thanks for checking! I just wanted to make sure we didn't run in to the same issue again!

refractionpcsx2:
@SamRH sorry, can you check one last time.  It is the last time, I super promise :D

SamRH:
Haha, no worries. I'll give it a test this afternoon (sometime in the next few hours for my timezone).

SamRH:
Re-tested with the same games and both OpenGL HW and Vulkan renderer at 4x internal resolution. All looked good :)

refractionpcsx2:
Awesome! Thanks for testing again :)

PR #5908, opened 2022-04-18
build: on linux use system wayland libs falling back to appimage ones

### LINKED ISSUE #5904: [BUG]: application crash upon booting any iso
### Describe the Bug

when attempting to launch any iso file, the emulator crashes and closes. This began after the latest patch on the discovery app store on the steam deck. Before the patch push everything worked great.

### Reproduction Steps

Launch any iso, i used ratchet & clank. The entire emulator will close.

### Expected Behavior

I expected the  game to launch properly.

### PCSX2 Revision

v1.7.2622

### Operating System

### If Linux - Specify Distro

Arch

SamRH:
### Description of Changes
Prefer system libwayland-* to the ones provided by the AppImage

### Rationale behind Changes
Currently the AppImage crashes when starting emulation with graphical output on certain stock configured Linux distributions like Fedora or SteamOS. Not sure on the exact internal details but relying on system wayland rather than a bundled one seems to solve this.

Fixes #5904 

### Suggested Testing Steps
- [x] Test a game starts on a Linux distribution that uses wayland 
- [x] Test a game starts on a Linux distribution without wayland

Has been confirmed working on a range of Linux distributions (OpenSuse, Debian, Fedora, Steam Deck/SteamOS) by various people in Discord.

### DIFF
--- a/.github/workflows/scripts/linux/app-variables.sh
+++ b/.github/workflows/scripts/linux/app-variables.sh
@@ -1,6 +1,12 @@
 #!/bin/bash
 
 export LD_LIBRARY_PATH="$APPDIR/usr/lib:$LD_LIBRARY_PATH"
+
+# use system wayland if available otherwise use the appimage provided wayland
+if [[ $(ldconfig -p | grep libwayland-client | wc -l) -lt 1 ]]; then
+        export LD_LIBRARY_PATH="$APPDIR/usr/lib/wayland:$LD_LIBRARY_PATH"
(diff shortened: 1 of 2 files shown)


SamRH:
Not the cleanest fix but not the worst.

SamRH:
For some reason I thought libwayland was being dlopened as needed. Have changed this so the system one will be used if available, otherwise one provided by the appimage will be used.

PR #5911, opened 2022-04-18
GameDB:add 'EETimingHack' to'Harry Potter and the Chamber of Secrets'

Mrlinkwii:
### Description of Changes

add 'EETimingHack' to 'Harry Potter and the Chamber of Secrets'
### Rationale behind Changes

fixes said game 
### Suggested Testing Steps

make sure ci passes or check the game


coornio:
For reference, this change is meant to alleviate most of the flickering between mipmaps that occurs with tree leaves in the game. They're not gone entirely though, they still flicker very occasionally.

PR #5915, opened 2022-04-18
Qt: Add RetroAchievements implementation

stenzek:
# THIS IS NOT FOR USER USE YET
## It is only for achievement developers for now! There are no achievement sets to complete anyway.

This PR is a work in progress to add RetroAchievements for PCSX2. It's based on my cross-platform implementation from DuckStation, so it won't be limited to Windows only.

**At the moment, there is no ingame UI.** This requires bringing over my "imgui fullscreen" library, which has not been done yet. The implementation also has to be made thread safe, so it can be accessed from the GS thread, where the UI will be rendered.

The reason I'm PRing this early is so that achievement set developers can test it, identify any issues, and start working on creating sets. Once we have some games with partial sets created, we can bring the full UI up in PCSX2.

**At this point, RetroAchievements are only planned for the Qt interface.** Changing anything in wx just tends to break other stuff, and due to the way this is integrated with the VM manager, plus the fact that it needs a UI for logging in, etc, I don't really have much desire to bring it across to wx.

### For RetroAchievements developers:

The procedure is almost the same as DuckStation. So I'm just borrowing my instructions from there :)

1. Download the Qt build from this PR, and extract it some place. PCSX2-Qt only operates in portable mode for now, so it won't conflict with any other installation. `Show All Checks` -> `Windows Builds / Qt (pull_request) (Details)` -> `Summary` -> `PCSX2-64bit-QT-pr[5915]...`.
2. Download RA_Integration-x64.dll from retroachievements.org. We have specifically disabled the auto-downloading in RAInterface, because it doesn't use TLS, and is therefore not safe to download and execute code since it's vulnerable to MITM attacks. Link: https://retroachievements.org/bin/RA_Integration-x64.dll
3. Rename RA_Integration-x64.dll to RA_Integration.dll, and move it into the directory you extracted PCSX2-Qt to.
4. Run PCSX2 for the first time to create the configuration file. You can also drop your BIOS file(s) in the `bios` directory at this point, as well as add games to scan via `Settings` -> `Add Game Directory`. Exit PCSX2.
5. Open PCSX2.ini in the `ini` directory, find the `[Achievements]` section, and add a new line `UseRAIntegration = true`. This will disable the built-in achievement handling, and forward everything to RAIntegration instead.
6. Start PCSX2 as normal. You will be prompted to log in to RetroAchivements the first time you start.
7. In the debug menu, there will be a RAIntegration submenu at the top which you can use to interact with RAIntegration.


pcsx2/VMManager.cpp
```diff
@@ -713,6 +728,11 @@ bool VMManager::Initialize(const VMBootParameters& boot_params)
 	frameLimitReset();
 	cpuReset();
 
+#ifdef ENABLE_ACHIEVEMENTS
+	if (EmuConfig.Achievements.Enabled)
```
Jamiras, line 732: needs to also check `!Achievements::IsUsingRAIntegration()` to prevent assertion in `Achievements::Initialize`

pcsx2/Frontend/Achievements.cpp
```diff
@@ -0,0 +1,1855 @@
...
+{
+	RA_InitClient((HWND)main_window_handle, "PCSX2", GIT_TAG);
+	RA_SetUserAgentDetail(Achievements::GetUserAgent().c_str());
+
+	RA_InstallSharedFunctions(RACallbackIsActive, RACallbackCauseUnpause, RACallbackCausePause, RACallbackRebuildMenu,
+		RACallbackEstimateTitle, RACallbackResetEmulator, RACallbackLoadROM);
+	RA_SetConsoleID(PlayStation2);
+
+	// For now, only the EE memory is exposed.. and only the bottom 2GB of it :S
+	RA_InstallMemoryBank(0, RACallbackReadEEMemory, RACallbackWriteEEMemory, std::numeric_limits<int>::max());
```
Jamiras, line 1695: You should only expose the 32MB of system memory. We don't need the entire addressable space.

stenzek: You do, because the addressing is virtual; games can remap the addresses how they like. Plus you have the IOP memory and scratchpad ;) This actually creates an issue since the size is signed, only the lower 2GB is accessible.

Jamiras: Every other system that RetroAchievements supports that has more than only 64KB of addressable memory only accesses the RAM chips (and possibly the SRAM chips) - not the full address space.

I understand that the code can access the RAM from various addresses to provide slightly different behavior, but the system still only has 32MB of RAM, so when we're looking for how many lives the player has left, we don't need to look at the ROM, echoes caused by disconnected pins, the video RAM, the memory cards, or whatever else may be virtualized through the bus. We only need the volatile memory that will be mutated by the program.

Because of the way we expect memory to be exposed, we don't expect our addresses to line up 1:1 with the addresses as they would appear on the system. For static memory, we just reference the value at the location where we expect to find it. For traversing a pointer, we manually calculate the adjustment needed to turn the real pointer into a bank address (usually by just masking off the uppermost bits).

Also, the search code will not handle this well. To find changes in memory, we take a snapshot of the exposed RAM and compare the current values to that snapshot. You've just provided us with a 2GB snapshot.

stenzek: I still suspect you're going to run into issues dereferencing pointers that are in the high address space, and you're assuming that the game will set up a 1:1 mapping between physical and virtual addresses (remember, the PS2 has a TLB and address translation).

The reason I went with virtual addresses is because that's what other external tools (through PINE) use, and it's the only way you can access IOP memory and the scratchpad.

But I'll change it for now, and if it creates issues we can deal with it later on I guess.

pcsx2/VMManager.cpp
stenzek: Moved the check inside `Achievements`.

3rdparty/rcheevos/include/rc_consoles.h
```diff
@@ -0,0 +1,127 @@
+#ifndef RC_CONSOLES_H
```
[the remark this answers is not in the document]
stenzek: No real reason, other than being slightly annoyed at the number of submodules we have :D But one more isn't a big deal, so switched it over.

`rc_api` wasn't a thing when I wrote the DuckStation implementation, but I could look at swapping over in the future.

pcsx2/Frontend/Achievements.cpp
refractionpcsx2: I will just add that trying to find stuff with physical addresses on the PS2 could be a real ball ache depending on what ytou wanna do, the chance of there being 1:1 mapping is highly unlikely, unless using your tools to make the achievements bypass any needing to debug it yourself to work out what's going on, it's going to make the life of people making the achievements harder.

CookiePLMonster: FWIW the physical address route will probably complicate relying on addresses obtained not from a memory scanner, but from a disassembler. In an ideal world, retroachievements could expose an inbuilt function to convert a virtual address to a physical address and vice versa - on PS1 you could e.g. get away by just doing `+/- 0x80000000` to the address, but by the sounds of it PS2 isn't that simple.

stenzek: Having a callback to translate a virtual to physical address is definitely something we could provide on our side.

CookiePLMonster: And vice versa, yeah. Other platforms could benefit from it too, e.g. allowing to drop that "mask highest bits" hack on the PS1.

stenzek: Going physical to virtual is a bit trickier, as we don't keep track of reverse mappings (after all, the things which work with physical addresses, such a DMA, don't need virtual addresses in the first place). Plus, the fact that you can have many virtual mappings for one physical address (for fastmem, I maintained a list).

CookiePLMonster: True, and since racheevos don't ever **write** you are unlikely to ever need to go this route. But for **reading**, virtual-to-physical is very much needed.

Jamiras:
The basic functionality all seems to be working:
* 32MB System RAM +16KB of scratch RAM exposed.
* Achievements are processed each frame and trigger appropriately
* Achievement state is included in save states and restored when loading save state

Things that aren't working:
* Framestepping and slowdown are available in hardcore mode. The application should not be allowed to run below 100% speed when hardcore mode is enabled.
* Pausing does not open the overlay. See `RA_SetPaused` and `RA_NavigateOverlay`. (Pausing is allowed in hardcore mode).
* Cheats should not be allowed in hardcore mode. Compatibility cheats are probably acceptable. I'm not certain if we have a stance on widescreen support.
* Reset re-identifies/re-activates the game, which discards all achievement state, including modifications to achievements.
* Shutdown does not close the game (`RA_ActivateGame(0)`). 
* Shutdown should also check for modifications. See `RA_ConfirmLoadNewRom`.
* Switching discs does not identify the new disc. Our general policy is that if the disc generates a hash but the server doesn't recognize the hash (`RA_IdentifyHash` returns 0), then the user must disable hardcore to continue (see `RA_WarnDisableHardcore`), or the emulator automatically switches back to the previous disc.

Minor:
* Menu check states are not reflected in the submenu. I understand this is a limitation of the implementation, and believe it can be addressed via https://github.com/RetroAchievements/RAIntegration/pull/881 when the next version of the DLL is released.
* The application title isn't showing the logged in user. see `RA_UpdateAppTitle`

refractionpcsx2:
@Jamiras what is the safety zone for hardcore zone? because the framerate on PS2 games can fluctuate a little, I guess some games will be impossible to run in hardcore mode since the framerate can halve in some cases.

Jamiras:
> @Jamiras what is the safety zone for hardcore mode? because the framerate on PS2 games can fluctuate a little, I guess some games will be impossible to run in hardcore mode since the framerate can halve in some cases.

I was just referring to the settings
image

There may be other settings that could indirectly affect framerate too. I imagine it's hard to maintain 60fps at 8x native resolution.

We don't usually enforce an actual maintained framerate. If you wanted to do something like that, I'd probably suggest a speed drop of more than 25% for a few seconds would be subject to enforcement. 

Are there a lot of games that are known to run slow for long periods of time?

RedDevilus:
Ratchet and Clank, Shadow of the Colossus, Jak, ...

Quite a few games that can turn into a lagfest.

refractionpcsx2:
It can depend on the hardware, but we get a lot of potatoes trying to run the games Red mentioned and they struggle to maintain a good speed on a budget machine.

if there's a 25% speed drop for more than a few seconds, I guess that's fair, 25% is a reasonable margin :)  Just wanted to check how strict it had to be.

stenzek:
> Framestepping and slowdown are available in hardcore mode. The application should not be allowed to run below 100% speed when hardcore mode is enabled.

Correct, none of the hardcode more checks/hooks are added yet. When I add them, it'll be the same between builtin/using integration.

> Pausing does not open the overlay. See RA_SetPaused and RA_NavigateOverlay. (Pausing is allowed in hardcore mode).

Should be fine to add an OnPaused() hook here to the core.

> Cheats should not be allowed in hardcore mode. Compatibility cheats are probably acceptable. I'm not certain if we have a stance on widescreen support.

See above.

> Reset re-identifies/re-activates the game, which discards all achievement state, including modifications to achievements.

This one is not going to be fixable. We track games based on the **ELF** which is loaded, not the disc. Resetting boots the IPL, which either loads OSDSYS or the game depending on whether fast boot is enabled or not. So, you'll get a "game change" event to the BIOS, and then again to the game once it's reloaded. One way could be to skip notifying the backend of the game change when it's switched to the BIOS, but I'm not really a fan of that idea, since from the achievement system's perspective, it's still monitoring a game which isn't running.

Using discs here also introduces another problem; a user can boot a custom/modded ELF, but use the original disc. Using the ELF on the disc would result in the same hash as the original, instead of the modded ELF.

> Shutdown does not close the game (RA_ActivateGame(0)).

> Shutdown should also check for modifications. See RA_ConfirmLoadNewRom.

I can call out to that on shutdown, but keep in mind there's no concept of a "ROM" here, as above, the closest thing is the ELF.

> Switching discs does not identify the new disc. Our general policy is that if the disc generates a hash but the server doesn't recognize the hash (RA_IdentifyHash returns 0), then the user must disable hardcore to continue (see A_WarnDisableHardcore), or the emulator automatically switches back to the previous disc.

See above note about using ELFs which are being loaded, not the disc. By the time the achievement implementation is notified of the ELF change, it's long after the disc change has happened, and impossible to stop. And that's assuming there's an ELF change in the first place, multidisc games tend to use the same ELF on different discs. And again, the ELF is the key identifier anyway, since that's what's going to determine the memory addresses you need to monitor. A disc change without an ELF change won't call `GameChanged()`, and really, from the achievement system's perspective, nothing has changed.

Minor:

> Menu check states are not reflected in the submenu. I understand this is a limitation of the implementation, and believe it can be addressed via https://github.com/RetroAchievements/RAIntegration/pull/881 when the next version of the DLL is released.

Should be doable.

> The application title isn't showing the logged in user. see RA_UpdateAppTitle

I'm not particularly a fan of the DLL messing with the title of the window. It's not the DLL's window, it's owned by the hosting application, so the only thing which should change it should be the application itself.

Jamiras:
>> Reset re-identifies/re-activates the game, which discards all achievement state, including modifications to achievements.

> This one is not going to be fixable. We track games based on the ELF which is loaded, not the disc. Resetting boots the IPL, which either loads OSDSYS or the game depending on whether fast boot is enabled or not. So, you'll get a "game change" event to the BIOS, and then again to the game once it's reloaded. One way could be to skip notifying the backend of the game change when it's switched to the BIOS, but I'm not really a fan of that idea, since from the achievement system's perspective, it's still monitoring a game which isn't running.

This is going to be a big annoyance.

Since hardcore prevents loading save states, players frequently reset the game when they fail a challenge to reload their in-game save. Reactivating the game would discard any progress that is not tracked by the game itself. It may not be an issue as I'm having trouble coming up with an examples for things you'd want to track _and_ persist through a reset.

For devs, a reset is a quick way to get back to the title screen. This can be circumvented by using save states, but it's still different than what devs are used to. If the reload can't be avoided, please call `RA_ConfirmLoadNewRom` to ensure the developer has a chance to save their work before the reload discards it.

> Using discs here also introduces another problem; a user can boot a custom/modded ELF, but use the original disc. Using the ELF on the disc would result in the same hash as the original, instead of the modded ELF.

I'm not sure I understand what you're saying here. This sounds kind of like loading a ROM through a GameGenie, in which case it should not be allowed in hardcore. 

Similarly, if you support softpatching, the hash should be calculated after applying the softpatch.

>  keep in mind there's no concept of a "ROM" here

Understood. `RA_ConfirmLoadNewRom` is just the name of the function. All it does is check for unsaved modifications and if any are found, it gives the user the chance to abort the operation so they don't lose them.

> multidisc games tend to use the same ELF on different discs.

Ideally, that's true. The logic to validate the secondary hashes on disc changes is to prevent the user from using a valid Disc 1 and a hacked Disc 2.

> A disc change without an ELF change won't call `GameChanged()`, and really, from the achievement system's perspective, nothing has changed.

That's fine. All of the hashing logic is just to ensure the player is using approved images. Limiting the hash to the executable code allows us to generate a hash quickly, as we assume that most hacks modify both the data and the executable code. It's not a perfect system, but it's the compromise we've arrived at.

> I'm not particularly a fan of the DLL messing with the title of the window.

That's fine. The function just provides a uniform presentation across emulators. Mostly, it just makes it easy for us to see the application version and logged in user when players post screenshots asking for help. Would it be possible for you to add the logged in user to the title bar?

RedDevilus:
> > Using discs here also introduces another problem; a user can boot a custom/modded ELF, but use the original disc. Using the ELF on the disc would result in the same hash as the original, instead of the modded ELF.
> 
> I'm not sure I understand what you're saying here. This sounds kind of like loading a ROM through a GameGenie, in which case it should not be allowed in hardcore.

Sounds like a situation where they can insert any version of the game (different region, modded image with cheats, etc) to circumvent triggering the cheating detection by the user.

stenzek:
> Pausing does not open the overlay. See RA_SetPaused and RA_NavigateOverlay. (Pausing is allowed in hardcore mode).

Done. As for the hardcore mode stuff, I'll look into that soonish, but I didn't prioritize it since there's nobody actually using it yet :)

> If the reload can't be avoided, please call RA_ConfirmLoadNewRom to ensure the developer has a chance to save their work before the reload discards it.

Done. I'm not a huge fan of the emu thread being blocked, since it's going to cause issues if it's in exclusive fullscreen, so you'll just have to careful of that (my built-in implementation renders using imgui on the graphics thread, so it's a non-issue there).

> I'm not sure I understand what you're saying here. This sounds kind of like loading a ROM through a GameGenie, in which case it should not be allowed in hardcore.
> Similarly, if you support softpatching, the hash should be calculated after applying the softpatch.

In PCSX2, you can override the ELF which is normally loaded by the IPL off the disc, with your own executable. This is a common way for users to mod games, since they can just "boot" using a different exe, instead of having to modify the disc itself. Hashing the copy on disk would therefore be incorrect here, but by doing things how I've currently got it set up, it'll hash the replacement ELF instead. Which will curb any trivial cheating attempts.

> Understood. RA_ConfirmLoadNewRom is just the name of the function. All it does is check for unsaved modifications and if any are found, it gives the user the chance to abort the operation so they don't lose them.

I've also called this on shutdown/quit.

> Would it be possible for you to add the logged in user to the title bar?

Having the emulator core completely isolated from the UI was a deliberate decision, and allows it to be reused for multiple platforms without an intrusive ifdef mess, or direct coupling to the UI (which has been a major source of pain in the old WX interface). The reverse is true too, there's no RAIntegration-specific code in the frontend (aside from the command line argument I added as an alternative way of enabling it).

So the short answer is it's not possible, at least not cleanly, because the frontend/UI maintains its own state (and its own title format), completely away from the emulator core (see `MainWindow::onGameChanged()`). The window title updates happen asynchronously (there's a game change event which gets delivered, and the UI updates whatever it needs to). Some of our potential target platforms (e.g. UWP, mobile, or headless CI runners) may not even have the concept of a window title, so an abstraction doesn't really make sense either.

stenzek:
Is there still any interest in this?

I'm in the process of bringing across my "big picture mode" UI to PCSX2, which the integrated achievements system uses to display unlocks, achievement lists/leaderboards/etc. If achievements for PS2 is not happening, I'll remove the code. But it would be silly to include it in what's going to be a relatively large PR, only to remove part of it later.

Tsearo:
Yep from the RetroAchievements side we are all very interested and hyped to be able to support PS2 in the future.

One of the things we've noticed in this version is that we can't load save state files from the file selector and it only works by using the F4 load hotkey. As a workaround we can rename the saved files as one of the numbered ones and then cycle to that number but it is pretty common for us to create many named Savestates when working on a set for testing so is this a known issue or something that could be looked into?

RedDevilus:
F3 will load the savestate or via WX: System > Load State > ... 
but yeah true you can't browse or drag an drop savestates.

stenzek:
@RedDevilus wx is irrelevant here, achievements are qt only.

Right click in the game list -> Load State File. Or, start the game, then system -> load state -> from file. This *should* work, although I can't remember if I hooked it up before or after this stuff was branched off (it needs a rebase anyway). Save states don't contain the disc information, so you can't load them without also providing the disc you want to mount (otherwise it'd be loading the state, but with "no disc" in the drive, which would cause a game crash).

Tsearo:
Sorry maybe I can clarify a little better now that I'm at my PC and can show an example. How do you specify the disc to load with them? Playing Rogue Galaxy I can use hotkeys to cycle to savestates 1-10 and save and load these with the Load hotkey just fine.
image

However if I go to Load from File... with the same exact savestate that I just created and loaded previously while the game is running
image
image

I get the following error
image

We can also rename other custom named states to match the above pattern and then use the Load hotkey to load these fine so it seems exclusive to the Load Savestate method from file from the System menu. If I right click on Rogue Galaxy and use Load State File... from the game list this way does actually work with any type of named file but requires exiting the game which is a bit of an inconvenience
image

stenzek:
Rebased on current master.

> However if I go to Load from File... with the same exact savestate that I just created and loaded previously while the game is running

Should be fixed with latest push. Also added drag and drop for states+disc images.

stenzek:
Rebased again. The native implementation is mostly finished now, except for hardcode mode, and leaderboard display in the ingame UI.

PR is much larger because it's now based on the big picture UI, this will go in first, separately, though.

Screenshot from 2022-06-05 22-21-52

PR #5920, opened 2022-04-19
Qt: Add crash handler/dumper

stenzek:
### Description of Changes

This PR adds a crash handler to PCSX2-Qt on Windows, which creates both a text file containing a backtrace, as well as a minidump file which can be opened in Visual Studio.

The text file might seem a bit redundant, but sometimes things can get into such a screwed up state that creating the minidump fails, so at least you'll get some kind of backtrace when that happens.

If you're running under a debugger, or attached one post-launch, no crash dump will be created, and the exception will be dispatched to the debugger as one would expect.

NOTE: The stack walker class is a bit old-style and clunky design-wise, but it works well, and I don't feel like reinventing the wheel here (or using wx, which we're going to remove eventually). I've fixed up the warnings that Codacity spews about it, and removed the old msvc version branches from it at least.

### Rationale behind Changes

Crash dumps are very useful when investigating issues reported by users, or perhaps more commonly, telling them to uninstall/update their broken controller drivers.

### Suggested Testing Steps

It's Qt, yolo. But I did test deliberately crashing, and it works as expected.

### DIFF
--- a/common/CrashHandler.cpp
+++ b/common/CrashHandler.cpp
@@ -1,0 +1,228 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2022  PCSX2 Dev Team
+ *
+ *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
+ *  of the GNU Lesser General Public License as published by the Free Software Found-
+ *  ation, either version 3 of the License, or (at your option) any later version.
+ *
+ *  PCSX2 is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+ *  PURPOSE.  See the GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along with PCSX2.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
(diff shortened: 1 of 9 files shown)


common/CrashHandler.cpp
```diff
@@ -0,0 +1,227 @@
...
+		return EXCEPTION_CONTINUE_SEARCH;
+
+	s_in_crash_handler = true;
+
+	wchar_t filename[1024] = {};
+	GenerateCrashFilename(filename, std::size(filename), s_write_directory.empty() ? nullptr : s_write_directory.c_str(), L"txt");
+
+	// might fail
+	HANDLE hFile = CreateFileW(filename, GENERIC_WRITE, 0, nullptr, CREATE_ALWAYS, 0, nullptr);
+	if (hFile)
```
cremno, line 142: From the [CreateFileW](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew#return-value) docs:

>If the function fails, the return value is **INVALID_HANDLE_VALUE**.

Which is `(HANDLE)-1` and not a null pointer.

cremno (changes requested):

common/CrashHandler.cpp
stenzek: Indeed, nice catch.

PR #5922, opened 2022-04-19
Counters: Present at VSync End

refractionpcsx2:
### Description of Changes
Moves the VSync present and framelimiter to VSync End interrupt

### Rationale behind Changes
Previously we were basically presenting the *previous* frame at the beginning of the next vsync start as a result of how things were working previously, we had to, however this will move it to VSync end, so the framebuffer pointers will have changed to the "just rendered" frame, which in theory should potentially reduce input lag by 1 frame. 

### Suggested Testing Steps
I can't remember what 2D fighting game used to have a problem with it this way around, but it's worth testing any you have.
Aside from that, test games for input lag (probably better on Qt) and make sure you don't get any black frame flashing that wasn't there previously.

### DIFF
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -485,16 +485,23 @@ static __fi void DoFMVSwitch()
 }
 
 // Convenience function to update UI thread and set patches. 
-static __fi void frameLimitUpdateCore()
+static __fi void VSyncUpdateCore()
 {
 	DoFMVSwitch();
 
 #ifndef PCSX2_CORE
 	GetCoreThread().VsyncInThread();
+#else
+	VMManager::Internal::VSyncOnCPUThread();
+#endif
+}
+
+static __fi void VSyncCheckExit()
+{
+#ifndef PCSX2_CORE
 	if (GetCoreThread().HasPendingStateChangeRequest())
 		Cpu->ExitExecution();
 #else
-	VMManager::Internal::VSyncOnCPUThread();
 	if (VMManager::Internal::IsExecutionInterrupted())
 		Cpu->ExitExecution();
 #endif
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Due to some Texture Cache issues (in Black PAL) and potential Pad issues on Qt, this is a no go for now.

Immersion95:
> ### Description of Changes
> Moves the VSync present and framelimiter to VSync End interrupt
> 
> Previously we were basically presenting the _previous_ frame at the beginning of the next vsync start as a result of how things were working previously, we had to, however this will move it to VSync end, so the framebuffer pointers will have changed to the "just rendered" frame, which in theory should potentially reduce input lag by 1 frame.
> 
> ### Suggested Testing Steps
> I can't remember what 2D fighting game used to have a problem with it this way around, but it's worth testing any you have. Aside from that, test games for input lag (probably better on Qt) and make sure you don't get any black frame flashing that wasn't there previously.

Hi, just so you know I had time to test it using internal frame advance.

That led me to those results : 

- Build from months ago : 

1. Capcom vs. SNK 2 - Jab - 5F
2. Capcom Fighting Jam - Jab - 4F
3. KOF 2003 - Move - 4F

- Latest build (almost certain that this is faithful to real PS2)
1. Capcom vs. SNK 2 - Jab - 4F
2. Capcom Fighting Jam - Jab - 3F
3. KOF 2003 - Move - 3F

- This PR (really great but I don't think it's accurate)
1. Capcom vs. SNK 2 - Jab - 2F
2. Capcom Fighting Jam - Jab - 2F
3. KOF 2003 - Move - 2F

This PR was feeling fantastic, I finally was able to consistently roll when on the ground in CVS2 (I'm a great Street Fighter player so I know the timing).

Unfortunately this messed up game with progressive patch (Tekken 4/5 with pnatch).

refractionpcsx2:
Messed up how? It would be nice to know for future reference and was Software mode affected too?

Immersion95:
> Messed up how? It would be nice to know for future reference and was Software mode affected too?

Here are the patches for reference.
[cheats.zip](https://github.com/PCSX2/pcsx2/files/8523742/cheats.zip)

Graphics were messed up in TTT, part of models were on black screen. In Tekken 5 the UI was flickering while in game. The thing is that those patches modify the games so i don't know if it's relevant.

refractionpcsx2:
I am a little suspicious that the game is fine without patches, makes me wonder if the patch has overlooked something.

Immersion95:
> I am a little suspicious that the game is fine without patches, makes me wonder if the patch has overlooked something.

Yes I wrote it in the previous post. The games are perfect without the patches that's why I'm unsure if this graphic bug is relevant.

refractionpcsx2:
Alright, well thank you for reporting it either way

PR #5925, opened 2022-04-20
Added VSCode Environment

kenshen112:
### Description of Changes

Adds launch.json, tasks.json, and settings.json for compilation and debugging across Linux and Windows platforms with VSCode.
Note. Does require CMake tools extension installed. And does require currently launching vs code from 64-bit dev console in windows.

### Rationale behind Changes

To add an easy unified platform for VSCode users to debug and compile with pcsx2.
To provide a graphical debug environment using GDB or MSVC and VSCode in Linux and Windows

### Suggested Testing Steps

Try opening VSCode

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,67 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Build PCSX2 - Debug GDB",
+            "type": "shell",
+            "command": "cd build && cmake -DEGL_API=TRUE -DCMAKE_BUILD_TYPE=Debug -GNinja -DBUILD_REPLAY_LOADERS=TRUE -DCMAKE_BUILD_PO=FALSE .. && ninja -j 10 && ninja install && cd ..",
+            "problemMatcher": [],
+            "group": {
+                "kind": "build",
+                "isDefault": true
+            }
+        },
+        {
+          "label": "Build PCSX2 - Devel GDB",
+          "type": "shell",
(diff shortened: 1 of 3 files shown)


.vscode/tasks.json
```diff
@@ -0,0 +1,67 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Build PCSX2 - Debug GDB",
+            "type": "shell",
+            "command": "cd build && cmake -DEGL_API=TRUE -DCMAKE_BUILD_TYPE=Debug -GNinja -DBUILD_REPLAY_LOADERS=TRUE -DCMAKE_BUILD_PO=FALSE .. && ninja -j 10 && ninja install && cd ..",
```
stenzek, line 9: Hardcoding the job count doesn't seem like a great idea to me. I have more than 10 CPUs ;)

Is there any way to only run the cmake command once? After it's been generated for the first time, there's no need to run it every time; it'll automatically regenerate the build file when needed by just invoking ninja. Running it unconditionally just adds a few seconds to the launch time.

.vscode/tasks.json
```diff
@@ -0,0 +1,67 @@
...
+          "problemMatcher": ["$msCompile"],
+          "group": {
+            "kind": "build",
+            "isDefault": true
+          }
+        },
+        {
+          "type": "shell",
+          "label": "Build PCSX2 - Devel MSVC",
+          "command": "cd build && cmake -DCMAKE_BUILD_TYPE=Devel -G Ninja -DCMAKE_CXX_COMPILER=cl.exe -DCMAKE_CXX_FLAGS=\"/Zi /EHsc\" ..",
```
stenzek, line 49: Why is `/Zi /EHsc` needed here? (or the compiler exe)

.vscode/tasks.json
Calinou: Ninja uses all CPU threads by default, so you do not need to pass `-j` to the build command.

Also, I recommend using the build system-agnostic `cmake --build .` and `cmake --build . target install` instead of `ninja` and `ninja install`

PR #5930, opened 2022-04-21
GS: Only copy texture coverage area for current FRAME/Z sources

stenzek:
### Description of Changes

This tends to be something which is pretty common, but only really has a performance hit when you have hundreds of copies in a frame. Basically, for things like auto flush and texture shuffles, we create a copy of the framebuffer to use as a texture, and then it gets immediately invalidated, because the FRAME pointer is also the TEX0 pointer.

Quite often, these draws tend to be quite small relative to the size of the framebuffer. For example, in shuffles, it's a vertical strip. Before, we were copying the *whole* texture, regardless of how much was actually being used. As you can imagine, that's pretty slow, and wastes a ton of memory bandwidth.

This PR changes the behaviour when FRAME/Z is used as the texture, only copying the area covered by the draw, based on the vertex trace UV result. It also adjusts said UV range when the draw gets scissor clipped, this happens in RE4.

### Rationale behind Changes

More brr. Some quick benchmarks on a RX580:

Jak 2 autoflush torture test: 27fps -> 49fps at 3x (Vulkan), 10fps -> 44fps (DX11)
Jak 2 after exiting prison: 25fps -> 115fps at 3x (Vulkan)

### Suggested Testing Steps

Test a range of games that use autoflush and texture/channel shuffle, make sure nothing is broken. There's a good chance there's an edge case I've missed.


refractionpcsx2:
Hitman Blood Money with Autoflush enabled is about a 100% uplift (nearly!), it goes from 197fps to 376fps on this PR :) Without Autoflush it gains about 20-30fps (but already doing like 500fps)

God of War with Autoflush enabled (needed for some effects) gets a good few % fps boost, though I don't have one of the heavier areas to check, where it actually matters, but I suspect it'll be decent.

All testing done in Vulkan

prafullpcsx2:
It introduces a bug during entrances in PDC darts 2008 game. The game has longstanding problems with lights anyway

[gs_20220421114642_PDC World Championship Darts 2008 _NTSC-U__SLUS-21773.gs.zip](https://github.com/PCSX2/pcsx2/files/8534331/gs_20220421114642_PDC.World.Championship.Darts.2008._NTSC-U__SLUS-21773.gs.zip)

gs_20220421114642_PDC World Championship Darts 2008 _NTSC-U__SLUS-21773

AhayriSG:
gs_20220422181609
GSDump: https://www.mediafire.com/file/hiwyrp9offo0qcu/SLUS-21426.7z/file

prafullpcsx2:
This pr seems to have some brightness and bloom issues with Prince of Persia Two thrones at native resolution. Gsdump attached
[gs_20220423232511_Prince of Persia - The Two Thrones _NTSC-U__SLUS-21287.gs.zip](https://github.com/PCSX2/pcsx2/files/8548378/gs_20220423232511_Prince.of.Persia.-.The.Two.Thrones._NTSC-U__SLUS-21287.gs.zip)
.

stenzek:
All of those should (hopefully) be fixed now.

MetrosexualGarbodor:
The performance hit when using autoflush in Xenosaga Episode I (required for detailed shadows in cutscenes) is completely gone now. Before updating I would get around 20fps in this shot. 

imagen

Nice work 👍

PR #5933, opened 2022-04-22
ElfObject: Ignore section headers when offset is invalid

stenzek:
### Description of Changes

This happens in V-Rally 3:

```
(ELF) Section header offset 14853196 is larger than file size 5898828
```

### Rationale behind Changes

Crashing/aborting PCSX2 on loading a game is never good.

### Suggested Testing Steps

Test both scanning and running V-Rally 3 with the patch applied (done, it seems to run okay from a quick test).

### DIFF
--- a/pcsx2/Elfheader.cpp
+++ b/pcsx2/Elfheader.cpp
@@ -62,17 +62,27 @@ void ElfObject::initElfHeaders(bool isPSXElf)
 
 	DevCon.WriteLn( L"Initializing Elf: %d bytes", data.GetSizeInBytes());
 
-	if ( header.e_phnum > 0 )
-		proghead = (ELF_PHR*)&data[header.e_phoff];
+	if (header.e_phnum > 0)
+	{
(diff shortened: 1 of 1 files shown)


pcsx2/Elfheader.cpp
```diff
@@ -62,17 +62,27 @@ void ElfObject::initElfHeaders(bool isPSXElf)
 
 	DevCon.WriteLn( L"Initializing Elf: %d bytes", data.GetSizeInBytes());
 
-	if ( header.e_phnum > 0 )
-		proghead = (ELF_PHR*)&data[header.e_phoff];
+	if (header.e_phnum > 0)
+	{
+		if ((header.e_phoff + sizeof(ELF_PHR)) <= data.GetSizeInBytes())
```
lightningterror, line 67: Hmm, a few new warnings on these, would be nice to take care of them.

https://github.com/PCSX2/pcsx2/runs/6128211125?check_suite_focus=true#step:11:503

PR #5939, opened 2022-04-23
build: Include CrashHandler.cpp in cmake. Fixes #5938.

### LINKED ISSUE #5938: [BUG]: PCSX2-Qt not building on Linux since introduction of CrashHandler
### Describe the Bug

`CrashHandler.cpp` (and the files it depends on) is not part of the build unless building on Windows. Still, `CrashHandler.h` is included by `Main.cpp`, creating an error at the moment of linking that makes the build fail.

By reverting to v1.7.2643, it builds correctly.

(I do realize the QT version is heavily WIP, especially on Linux, but this doesn't just break features or require workarounds on the user side, it actually makes the build completely fail, so I thought it might be worth it to report it anyway)

### Reproduction Steps

1. Try building with `-DQT_BUILD=TRUE`
2. Fail

### Expected Behavior

1. Try building with `-DQT_BUILD=TRUE`
2. Get a working executable

### PCSX2 Revision

v1.7.2647

### Operating System

### If Linux - Specify Distro

Gentoo

arcum42:
### Description of Changes
Add CrashHandler.cpp and CrashHandler.h to CMakeLists.txt.

### Rationale behind Changes
Pcsx2-qt will not compile without this change. Fixes #5938 .

### Suggested Testing Steps
Compile pcsx2-qt in Linux.

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -14,6 +14,7 @@ target_sources(common PRIVATE
 	EventSource.inl
 	SafeArray.inl
 	Console.cpp
+	CrashHandler.cpp
 	EventSource.cpp
 	Exceptions.cpp
 	FastFormatString.cpp
@@ -65,6 +66,7 @@ target_sources(common PRIVATE
 	Assertions.h
 	boost_spsc_queue.hpp
 	Console.h
+	CrashHandler.h
 	Dependencies.h
 	EnumOps.h
 	EventSource.h


Zekromaster:
I'm... actually surprised this works.

arcum42:
The file actually had stubs of the functions for non Windows operating systems. It just hadn't actually been added to cmake, so the linker wasn't finding them.

Zekromaster:
> The file actually had stubs of the functions for non Windows operating systems. It just hadn't actually been added to cmake, so the linker wasn't finding them.

Still, how... how doesn't it break through the inclusion of the likes of DbgHelp.h?

arcum42:
A few lines above that, it has:
#if defined(_WIN32) && !defined(_UWP)

so none of that's actually done on Linux.

Zekromaster:
> A few lines above that, it has: #if defined(_WIN32) && !defined(_UWP)
> 
> so none of that's actually done on Linux.

... oh, I missed the fact there's no `#endif` there

arcum42:
Yeah, there's an #else near the bottom.

Of course, fixing the linux version of pcsx2 is how I ended up on the team in the first place. I kept opening prs every time linux broke, which at the time was every few commits...

stenzek:
If someone wanted to set up the CI for Linux we'd know when things broke ;)

arcum42:
Now that Ubuntu 22.04 has released, that should be rather easier, once it's added in, anyways.

stenzek:
Not really. We'll still need our own Qt build for two reasons:
 - If you build on 22.04, the appimage won't run on older versions.
 - The package version is 6.2.4, we're on 6.3.0. It'll build on 6.2.x, for now anyway, but could create problems in the future.

Issue #5599, opened 2022-03-04
[BUG]: Qt frontend should respect the XDG standard


j8r:
### Describe the Bug

When using the Qt frontend inside Flatpak, at first the `resources` directory, which is located at `/app/share/PCSX2/resources` inside the sandbox, is not found: `resources directory missing, your installation is incomplete`. I don't know how the wxWidgets frontend handle that. A quick fix is to move the folder where the binaries are.

After fixing that, `pcsx2-qt` tries to write directly to `$HOME/PCSX2`, which is forbidden by the Flatpak sandbox.

Finally, a `textures` folder is trying to be created in the same directory where the binaries are.

---

Command:
```sh
flatpak run --branch=master --command=pcsx2-qt net.pcsx2.PCSX2
```

Output (after fixing the `resources` directory error):
```sh
Qt: Session management error: Could not open network socket
Failed to save settings to '/home/j8r/PCSX2/inis/PCSX2.ini'.
BIOS Directory: /home/j8r/PCSX2/bios
Snapshots Directory: /home/j8r/PCSX2/snaps
Savestates Directory: /home/j8r/PCSX2/sstates
MemoryCards Directory: /home/j8r/PCSX2/memcards
Logs Directory: /home/j8r/PCSX2/logs
Cheats Directory: /home/j8r/PCSX2/cheats
CheatsWS Directory: /home/j8r/PCSX2/cheats_ws
Covers Directory: /home/j8r/PCSX2/covers
Game Settings Directory: /home/j8r/PCSX2/gamesettings
Cache Directory: /home/j8r/PCSX2/cache
Textures Directory: /app/bin/textures
21:53:16: Error: Directory '/home/j8r/PCSX2' couldn't be created (error 30: Read-only file system)
21:53:16: Error: Directory '/home/j8r/PCSX2' couldn't be created (error 30: Read-only file system)
21:53:16: Error: Directory '/home/j8r/PCSX2' couldn't be created (error 30: Read-only file system)
21:53:16: Error: Directory '/home/j8r/PCSX2' couldn't be created (error 30: Read-only file system)
21:53:16: Error: Directory '/home/j8r/PCSX2' couldn't be created (error 30: Read-only file system)
21:53:16: Error: Directory '/home/j8r/PCSX2' couldn't be created (error 30: Read-only file system)
21:53:16: Error: Directory '/home/j8r/PCSX2' couldn't be created (error 30: Read-only file system)
[...]

Issue #5496, opened 2022-02-06
[BUG]: Qt build doesn't respect XDG_STD compile flag


bsdf:
### Describe the Bug

The Qt build appears to have a [hardcoded reference](https://github.com/PCSX2/pcsx2/blob/master/pcsx2-qt/QtHost.cpp#L88) to `$HOME/PCSX2` for its data even if the `XDG_STD` build flag is present. With this build flag, the proper behavior would be to check `~/.config/PCSX2` instead.

Personally, I think getting rid of `XDG_STD` and the hardcoded home dir and using [`QStandardPaths`](https://doc.qt.io/qt-6/qstandardpaths.html) for config and data locations would be the best solution

### Reproduction Steps

- `cmake -DXDG_STD=1 -DQT_BUILD=1 -DPACKAGE_MODE=1 -DCMAKE_INSTALL_PREFIX=/usr/local/stow/pcsx2 ..`
- `make && make install`
- `/usr/local/stow/pcsx2/bin/pcsx2-qt`
- observe error "Resources directory is missing"

### Expected Behavior

All needed files are present and installed paths in accordance with XDG standard dirs, no error is received

### PCSX2 Revision

061fff6f17efa28654bdf5c61121c78a3d65cf42

### Operating System

Linux (64bit) - Specify Distro Below

### If Linux - Specify Distro

debian sid

RedDevilus:
Linux doesn't even have a working Qt version at this time.

F0bes:
Qt isn't ready yet and as said above, the Linux side of things hasn't even been worked out.

bsdf:
Hello, I wanted to start dipping my toes into contributing to the project and thought opening a ticket to discuss the behavior, intended behavior, and potentially the proposed change would be better than a PR out of nowhere. I'm not sure what you two mean by not having a working linux version since I'm playing games over here in the Qt version just fine-- certainly working well enough to hack on.

RedDevilus:
That's fine but it's just that we don't accept any Qt issues at the moment as it's in big flux for missing features/screens and other issues and then the issues count would bloat as it's not intended for main usage yet. Besides there is too much other works happening to even track the WIP Qt issues. Qt PRs are a different matter in this context however.

Hopefully you can understand this reasoning.

bsdf:
Makes sense, thanks for the explanation

RedDevilus:
[...]

PR #5943, opened 2022-04-23
Qt: Implement proper data directory selection

stenzek:
### Description of Changes

Makes PCSX2-Qt follow the same data directory selection as the wx interface, i.e. `Documents/PCSX2` for Windows, `~/.config/PCSX2` for Linux, and `~/Library/Application Support/PCSX2` for Macs (not sure if correct).

Of course, `portable.ini` will continue to override and use the program directory.

I don't want to bikeshed about whether things should go in `XDG_CONFIG_HOME` or `XDG_DATA_HOME`, I'm just making Qt the same as WX. If you want to argue to change it, you can do so with a follow-up PR ;)

### Rationale behind Changes

Consistency, making it easier to migrate to Qt.

### Suggested Testing Steps

Test Windows and Linux Qt in both portable and non portable mode (I've tested Windows).

### DIFF
--- a/pcsx2-qt/QtHost.cpp
+++ b/pcsx2-qt/QtHost.cpp
@@ -22,6 +22,8 @@
 
 #ifdef _WIN32
 #include "common/RedtapeWindows.h"
+#include <KnownFolders.h>
+#include <ShlObj.h>
 #endif
 
 #include "common/Assertions.h"
@@ -52,6 +54,9 @@ static constexpr u32 SETTINGS_SAVE_DELAY = 1000;
 namespace QtHost {
 static void InitializeWxRubbish();
 static bool InitializeConfig();
+static bool ShouldUsePortableMode();
+static void SetResourcesDirectory();
+static void SetDataDirectory();
 static void HookSignals();
 static bool SetCriticalFolders();
 static void SetDefaultConfig();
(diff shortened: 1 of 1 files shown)


stenzek:
Tested on Linux, seems to work as expected. Although on my system I didn't have `$XDG_CONFIG_HOME` available.

j8r:
No that looks good as is, thanks @stenzek! The most important was to put at least in one of this standard locations.
Solves #5496 and #5599.

j8r:
Most is working well, but I noticed one thing about the `resources` directory in [Flatpak](https://github.com/flathub/net.pcsx2.PCSX2/pull/46), and maybe just Linux in general: `pcsx2` read them at `/app/share/PCSX2/resources` - which were installed there by CMake. However `pcsx2-qt` read them at `/app/bin/resources`, so I have to create a symlink.

I looked at `QtHost.cpp` and `AppConfig.cpp`, but I wasn't able to figure out the logic where Wx Widget sets the directory to `/app/share/PCSX2/resources`.

In both cases, with or without `-DQT_BUILD=TRUE`, CMake install the resources to `/app/share/PCSX2/resources`: `-- Installing: /app/share/PCSX2/resources/`.

The fixes are either:
- install the resources where the binary is when `-DQT_BUILD`
- or better, make the Qt binary read `/app/share/PCSX2/resources`, like the Wx Widget one

j8r:
Where did you find that this package is officially supported by you? From my point of view there is no such statement in the package. Usually people open issues in the dedicated repository, then us packagers communicate there - at least that's what we aim for.

I just wanted to know if it was supposed to work this way or if it was a Linux regression, since the behavior has changed from wxWidget to Qt.

Thanks to have clarified this, it is indeed supposes to work like done in my fix: place the resources directory alongside the binary.
So, logically, it could be done by CMake too.

stenzek:
> Where did you find that this package is officially supported by you?

Because a little while back we had an influx of steam deck people whinging about things being broken, and it had nothing to do with us.

> since the behavior has changed from wxWidget to Qt.

Qt is not wx. Plenty of things are done differently, wx was absolute garbage in many ways and the idea never was to replicate it.

> So, logically, it could be done by CMake too.

Don't really see the need to do anything here tbh. "make install" will copy the binary to the in-tree bin directory, where the resources already are. Alternatively, for development, you can just symlink it into the build directory (which is what I usually do).

j8r:
Thanks for the clarifications.
That's the risk, users not opening issues in the right place, and that's not always obvious to know if the bug comes from the application itself or caused by the packaging - even more in a sandboxed one.

I just noticed the issues link points currently to https://github.com/PCSX2/pcsx2/issues, I could change it to https://github.com/flathub/net.pcsx2.PCSX2/issues.
We could always redirect them here if the issue is an actual emulation problem, not a system one.

PR #5945, opened 2022-04-24
GUI: Append warning for Manual HW Hacks

RedDevilus:
### Description of Changes

Addendum to the Manual HW Hacks checkbox that warns people that enabling this option undoes automatic (renderer and upscaling) fixes.
### Rationale behind Changes

WX and Qt both will disable renderer and upscaling fixes if you want to manually handle the renderer and upscaling fixes. For years enabling this option didn't do much harm for users even if you didn't change any other setting, but since 1.7 development series automatised these to a degree it would be nice for users to know why their game looks potentially awful. Quite a lot of users don't know how to correctly handle this tab anyway and shouldn't touch something they don't get and get worse. Though in some cases like texture offsets you do need to override the automatic fixes.
### Suggested Testing Steps

Look if the text is correct.

### DIFF
--- a/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
@@ -252,6 +252,11 @@ GraphicsSettingsWidget::GraphicsSettingsWidget(SettingsDialog* dialog, QWidget* 
 	connect(m_ui.enableHWFixes, &QCheckBox::stateChanged, this, &GraphicsSettingsWidget::onEnableHardwareFixesChanged);
 	updateRendererDependentOptions();
 
+	dialog->registerWidgetHelp(m_ui.enableHWFixes, tr("Manual Hardware Renderer Fixes"), tr("Unchecked"),
+		tr("Enabling this option gives you the ability to change the renderer and upscaling fixes "
+		   "to your games. However IF you have ENABLED this, you WILL DISABLE AUTOMATIC "
+		   "SETTINGS and you can re-enable automatic settings by unchecking this option."));
+
 	dialog->registerWidgetHelp(m_ui.useBlitSwapChain, tr("Use Blit Swap Chain"), tr("Unchecked"),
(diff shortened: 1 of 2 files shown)


RedDevilus:
Preview:
image

refractionpcsx2:
What about Qt?

RedDevilus:
Not sure what to do with the Qt name.

RedDevilus:
image

Not enough space for the same text either way, I thought about the tooltip atleast in the future.

refractionpcsx2:
well if there's not enough space, Qt should move it to somewhere it does, the checkboxes kind of cascade

RedDevilus:
Guess I'll move the UI option down

refractionpcsx2:
other option, is make it part of the tooltip, instead of the name.  People are gonna ignore it anyways.

RedDevilus:
Don't want to fix Qt atm. I'll probably fix it later.

RedDevilus:
Maybe in a new PR even.

stenzek:
You could do a popup warning/confirm box (maybe one-time) instead.

refractionpcsx2:
> You could do a popup warning/confirm box (maybe with a one-time/don't show me again option) instead.

Oh I tried that when I did BIOS configurations for fast boot, NOBODY read it :D

RedDevilus:
I read it but yeah nobody cares about popups, it's like insects where you smash till it's death for users.

RedDevilus:
Biggest example is the OnePad popup that you can't configure it via the GUI (minus keyboard)

RedDevilus:
WX:
image

Qt:
image

RedDevilus:
I changed the Qt one to a tooltip:

image

As the text would become a translatable and would be better as a tooltip.

NoShotz:
I think this would be more correct grammar

> and you can re-enable the automatic settings by unchecking this option.

RedDevilus:
> I think this would be better grammar wise
> 
> > and you can re-enable the automatic settings by unchecking this option.

Already changed the text to better grammar.

RedDevilus:
image

PR #5949, opened 2022-04-24
Remove iMisc.cpp, and move the code to System.cpp/h.

arcum42:
### Description of Changes
iMisc.cpp was entirely deleted, and SetCPUState, g_sseVUMXCSR, & g_sseMXCSR were moved over to System.cpp.

### Rationale behind Changes
SetCPUState is only actually used in VMManager.cpp and SysCoreThread.cpp. Maintaining a separate Misc file for one function and two variables seemed pointless, and everything that uses them includes System.h.

### Suggested Testing Steps
Make sure everything still compiles and runs.


stenzek:
IMO, `x86/iCore` would be a better location, as this is x86 specific code.

refractionpcsx2:
Not *entirely* true, the VU interpreters use it, and the EE interpreter could in theory use it (and it probably should for compatibility purposes tbh)

arcum42:
Pretty much a breakdown of where these are used would be:
SetCPUState: VMManager.cpp, SysCoreThread.cpp
g_sseVUMXCSR: VU0microInterp.cpp, VU1microInterp.cpp, and microVU_Execute.inl
g_sseMXCSR: iFPU.cpp, iFPUd.cpp, and microVU_Execute.inl

Moving this to system and taking it out of common just meant adding one include. Moving it to iCore would mean adding several new includes (largely because the interpreters aren't including iCore) or sticking it back in common.

I'm open to moving it other places. I largely moved it to System based on where SetCPUState was used.

refractionpcsx2 (approved):

PR #5952, opened 2022-04-25
MTGS: Combine init/read FIFO and remove several MTGS waits

stenzek:
### Description of Changes

Gets rid of a bunch of GS thread waits, and combines init/copy for GS downloads.

### Rationale behind Changes

Slightly more brr for something which is already slow.

### Suggested Testing Steps

Test games which use readbacks and make sure they still function as expected. Need those which are actually sensitive to the contents of the download (e.g. I tried it with Oni which uses it for light visibility).


pcsx2/GSDumpReplayer.cpp
```diff
@@ -276,10 +276,8 @@ void GSDumpReplayerCpuStep()
 			u32 size;
 			std::memcpy(&size, packet.data, sizeof(size));
 
-			std::unique_ptr<u8[]> arr(new u8[size]);
-			GetMTGS().WaitGS();
-			GetMTGS().SendPointerPacket(GS_RINGTYPE_INIT_READ_FIFO2, size, arr.get());
-			GetMTGS().WaitGS(false); // wait without reg sync
+			std::unique_ptr<u8[]> arr(new u8[size * 16]);
```
tellowkrinkle, line 279: I assume this was broken before?

tellowkrinkle (approved):

pcsx2/GSDumpReplayer.cpp
stenzek: Yeah, but because it never actually did the read (only set it up/init'ed it), it never caused any issues.

tellowkrinkle: That's why readbacks didn't trigger in GSdumps isn't it

stenzek: They always seemed to trigger for me? It just never copied it to the DMA buffer (`arr` in this case).

tellowkrinkle: Really?  There are a number of games that do that for me
e.g. this VP2 trace had 1 readback during the game when it was taken, but 0 in trace playback: [VP2_MissingReadback.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/8551468/VP2_MissingReadback.gs.xz.zip)

Doesn't look like this changes it though.  Must be for a different reason.

stenzek: Pretty strange, the init fifo should get recorded into the dump when it happens. I'll double-check that.

refractionpcsx2:
Works great :) LTGM!

PR #5955, opened 2022-04-25
GS/HW: Implement moves in texture cache

### LINKED ISSUE #1725: Devil May Cry: Upper-third of screen is blurred (regression)
***PCSX2 version:***
v1.5.0-dev-1744-g8fecd35

#1613 (gsdx tc: miss a set scaling call) introduced the regression

***PCSX2 options:***
Defaults

***Plugins used:***
Defaults, GSdx OGL, 4x Native

During 3D portions of the game, the upper-third of the screen is blurry. You can see the separation in screenshot. The issue is only visible in resolutions above Native.

Boot the game, start a level and look at the top of the screen.

v1.5.0-dev-1426-g3b73393

***PC specifications:***
i5-6600k@4.7Ghz, GTX 1070 SC, Windows 10 64-bit Insider Build 14986

dmcregression

### LINKED ISSUE #3169: Armored Core: Last Raven (NTSC-U - SLUS_213.38) - Blurry graphics during missions (in HW mode).
***PCSX2 version:***
1.7.0 dev 1040 (also tried 1.6.0 and 1.4.0. No difference)

***PCSX2 options:***
Defaults, 'Safe' preset. I've also tried different settings to no effect. Speedhacks were disabled during testing.

***Plugins used:***
GSdx, SPU2-X, LilyPad, cdvdGigaherz, the rest are null plugins.

***Plugin settings:*** 
Tested GSdx in both OpenGL and Direct3D 11 modes, HW and SW. Tried all sort of settings combinations to no effect. "Preload Frame Data" hack enabled to avoid HW mode blackscreen.

GS dump: [gsdx_20191205031754.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/3925432/gsdx_20191205031754.gs.xz.zip)

The game simply looks wrong during missions in HW mode. The entire screen is a blurry mess. Graphics are perfectly fine while in menus (no blur + proper upscaling). This issue doesn't show up in the previous game in the series (Armored Core: Nexus). Image comparison included, screenshots taken in OpenGL HW mode.

gsdx_20191204205752 gsdx_20191204204827
HW mode upscaling working properly in menus, no blur present:
gsdx_20210317131417

Enter any mission, battle or AC Test mode.

Never worked right, as far as I'm aware.

***PC specifications:***
CPU: AMD Ryzen 5 2600
GPU: Geforce GTX 1060 6GB
OS: Windows 7 Ultimate SP1 x64

stenzek:
### Description of Changes

Currently, for local memory moves, we invalidate all hardware targets, and then do the copy on the CPU. Which works great, but it's slow (readbacks), and you lose upscaling.

This PR implements a subset of the possible move cases in the texture cache (when SBP+DBP exist in the target list, and the sizes/psm match). Which is enough for games such as Armored Core, Devil May Cry, Bust-A-Move Dance Summit 2001.

### Rationale behind Changes

More brr, more upscaling.

### Suggested Testing Steps

Test games which use GS moves.

Fixes #1725
Fixes #3169

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
+++ b/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
@@ -1045,6 +1045,68 @@ void GSTextureCache::InvalidateLocalMem(const GSOffset& off, const GSVector4i& r
 	// TODO: ds
 }
 
+bool GSTextureCache::Move(u32 SBP, u32 SBW, u32 SPSM, int sx, int sy, u32 DBP, u32 DBW, u32 DPSM, int dx, int dy, int w, int h)
+{
+	// TODO: In theory we could do channel swapping on the GPU, but we haven't found anything which needs it so far.
+	// Same with SBP == DBP, but this behavior could change based on direction?
+	if (SPSM != DPSM || SBP == DBP)
+	{
+		GL_CACHE("Skipping HW move from 0x%X to 0x%X with SPSM=%u DPSM=%u", SBP, DBP, SPSM, DPSM);
+		return false;
+	}
+
+	// DX11/12 is a bit lame and can't partial copy depth targets. We could do this with a blit instead,
+	// but so far haven't seen anything which needs it.
+	if (GSConfig.Renderer == GSRendererType::DX11 || GSConfig.Renderer == GSRendererType::DX12)
+	{
+		if (GSLocalMemory::m_psm[SPSM].depth || GSLocalMemory::m_psm[DPSM].depth)
+			return false;
+	}
+
+	// Look for an exact match on the targets.
+	GSTextureCache::Target* src = GetExactTarget(SBP, SBW, SPSM);
+	GSTextureCache::Target* dst = GetExactTarget(DBP, DBW, DPSM);
+	if (!src || !dst || src->m_texture->GetScale() != dst->m_texture->GetScale())
+		return false;
+
+	// Scale coordinates.
+	const GSVector2 scale(src->m_texture->GetScale());
+	const int scaled_sx = static_cast<int>(sx * scale.x);
+	const int scaled_sy = static_cast<int>(sy * scale.y);
+	const int scaled_dx = static_cast<int>(dx * scale.x);
+	const int scaled_dy = static_cast<int>(dy * scale.y);
+	const int scaled_w = static_cast<int>(w * scale.x);
+	const int scaled_h = static_cast<int>(h * scale.y);
+
+	// Make sure the copy doesn't go out of bounds (it shouldn't).
+	if ((scaled_sx + scaled_w) > src->m_texture->GetWidth() || (scaled_sy + scaled_h) > src->m_texture->GetHeight() ||
+		(scaled_dx + scaled_w) > dst->m_texture->GetWidth() || (scaled_dy + scaled_h) > dst->m_texture->GetHeight())
+	{
+		return false;
+	}
+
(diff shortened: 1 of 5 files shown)


pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -1047,6 +1047,68 @@ void GSTextureCache::InvalidateLocalMem(const GSOffset& off, const GSVector4i& r
...
+
+	// Invalidate any sources that overlap with the target (since they're now stale).
+	InvalidateVideoMem(m_renderer->m_mem.GetOffset(DBP, DBW, DPSM), GSVector4i(dx, dy, dx + w, dy + h), false);
+	return true;
+}
+
+GSTextureCache::Target* GSTextureCache::GetExactTarget(u32 BP, u32 BW, u32 PSM) const
+{
+	auto& rts = m_dst[GSLocalMemory::m_psm[PSM].depth ? DepthStencil : RenderTarget];
+	for (auto it = rts.rbegin(); it != rts.rend(); ++it) // Iterate targets from LRU to MRU.
```
iMineLink, line 1102: Here you may want to search starting from the MRU, so from the beginning of the list; Hopefully it won't make a difference though.

Dreadmoth:
The motion blur effect on zoom level transitions in Champions of Norrath and Champions: Return to Arms now upscales nicely:

Master (5x Native):
crta_master

PR (5x Native):
crta_pr

The fade to black on entering/exiting the inventory is similarly improved.

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
stenzek: Good point, I just grabbed it from the lookup functions.

But yeah, if we have more than one target with a matching BP, BW and PSM, we probably have bigger problems..

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -1047,6 +1047,68 @@ void GSTextureCache::InvalidateLocalMem(const GSOffset& off, const GSVector4i& r
...
+	// but so far haven't seen anything which needs it.
+	if (GSConfig.Renderer == GSRendererType::DX11 || GSConfig.Renderer == GSRendererType::DX12)
+	{
+		if (GSLocalMemory::m_psm[SPSM].depth || GSLocalMemory::m_psm[DPSM].depth)
+			return false;
+	}
+
+	// Look for an exact match on the targets.
+	GSTextureCache::Target* src = GetExactTarget(SBP, SBW, SPSM);
+	GSTextureCache::Target* dst = GetExactTarget(DBP, DBW, DPSM);
```
iMineLink, line 1070: Now that I'm thinking of it, it dst does not exist in cache you can just copy src, change the BP/BW fields and add it to the relevant list. Also it might be needed to check that no overlap exist between the src and dst GS memory regions...but... that would be an evil move.

stenzek: Good suggestion, but for the record, I wanted to keep this change as small as possible in case of regressions. Could be worth doing in the future, but I don't think there's any cases where it does happen currently (at least in the ones I tested, it was moves *inside* an existing target).

Also keep in mind that the missing-from-cache cache would only happen once, with the previous frame's target being around for the next frame. So we'd probably want to drop the old target in such a case.

washingleavesandthensome23-434:
very freakin nice my guys ....no idea how long ive been waiting for this for dmc1, big thanks !

PR #5969, opened 2022-04-26
GS: Merge GSRendererNew and GSRendererHW

stenzek:
### Description of Changes

We haven't had any non-"new" renderers for a while now, so it seems silly to keep half the hardware rendering stuff in a base class, and another in a derived class.

### Rationale behind Changes

Cleaning up legacy stuff, yay.

### Suggested Testing Steps

Make sure hardware renderers still work. It's a cut/paste of all the function definitions, so no behaviour should change.


fantesykikachu:
Just for future Github reference this finishes the work from #5138 .

PR #5978, opened 2022-04-27
GS: Fix ReadTexture on unaligned reads

TellowKrinkle:
### Description of Changes
When it sees an unaligned request, ReadTexture finds the largest aligned interior rectangle it can, and attempts to read the following pieces:
intended

Somehow, it ends up reading this instead:
actual

Since this is larger than the image it was asked to read, this ends up writing out of bounds.  Due to the nature of GPU upload buffers, these OOB reads often do nothing, but sometimes they actually write outside of a buffer and crash.  Fun!

This fixes ReadTexture to read into the intended location.

### Rationale behind Changes
Not crashing is good

### Suggested Testing Steps
[This GSdump was crashing in the Metal renderer previously](https://github.com/PCSX2/pcsx2/files/8569256/DragonBallZ-MetalCrash.gs.xz.zip)

### DIFF
--- a/pcsx2/GS/GSLocalMemory.cpp
+++ b/pcsx2/GS/GSLocalMemory.cpp
@@ -1670,15 +1670,9 @@ void GSLocalMemory::ReadTexture(const GSOffset& off, const GSVector4i& r, u8* ds
 		}
 		else
 		{
-			for (int y = r.top; y < cr.top; y++, dst += dstpitch)
-			{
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Strange how it didn't seem to affect windows (That I know of???) but interestingly, the GSDump is kinda busted anyways, it seems to be incomplete.

> (GSDump) Dropping last packet of 344080 bytes (we only have 341725 bytes)

PR #5903, opened 2022-04-17
Merge SSE4 and AVX2 C Rasterizer

tellowkrinkle:
### Description of Changes
Squish two functions into one

### Rationale behind Changes
Less code that's not used half the time and easy to accidentally forget to update

### Suggested Testing Steps
Enable the C rasterizer and make sure nothing broke

PR #5980, opened 2022-04-28
GS:SW: zclamp fixup

tellowkrinkle:
### Description of Changes
- ~~Reduces the number of times zclamp is run~~ Fixes an issue where zclamp would be run a second time to clamp out the destination z bits that had just been loaded in order to preserve them
- Fixes an issue in the C rasterizer where zclamp was run after setting the top bit in preparation for using an SSE signed compare as an unsigned compare, which messed up the comparison.
- Fixes an issue in the C rasterizer where the unused z bits (e.g. top 8 bits of Z24) would be wiped out by the C rasterizer
- Disables the `zoverflow` check for enabling signed comparison workarounds, since if `zoverflow && zpsm != 0` then zclamp would be enabled and would clamp the z value down to a size where signed comparison doesn't cause any issues.

Note: Requires #5903

### Rationale behind Changes
- It's nice when your reference renderer matches the ASM generator based off it
- ~~Why zclamp twice when you can zclamp once~~ Let's not clamp away the bits we were trying to preserve

### Suggested Testing Steps
- [Test this GSdump with the C rasterizer (tests zclamp top bit issue)](https://github.com/PCSX2/pcsx2/files/8578637/GodOfWar-BrightSun.gs.xz.zip)
- [Test this one too (tests unused z trashing issue)](https://github.com/PCSX2/pcsx2/files/8588500/ArmoredCore2-RadarGarbage.gs.xz.zip)
- Test zclamp dumps to make sure I didn't break any

### DIFF
--- a/pcsx2/GS/Renderers/SW/GSDrawScanline.cpp
+++ b/pcsx2/GS/Renderers/SW/GSDrawScanline.cpp
@@ -479,6 +479,9 @@ void GSDrawScanline::CDrawScanline(int pixels, int left, int top, const GSVertex
 					{
 						zs = VectorI(z);
 					}
+
+					if (sel.zclamp)
+						zs = zs.min_u32(VectorI::xffffffff().srl32(sel.zpsm * 8));
 				}
 				else
 				{
@@ -505,14 +508,11 @@ void GSDrawScanline::CDrawScanline(int pixels, int left, int top, const GSVertex
 					VectorI zso = zs;
 					VectorI zdo = zd;
 
-					if (sel.zoverflow || sel.zpsm == 0)
+					if (sel.zpsm == 0)
 					{
 						zso -= VectorI::x80000000();
 						zdo -= VectorI::x80000000();
 					}
-
-					if (sel.zclamp)
-						zso = zso.min_u32(VectorI::xffffffff().srl32(sel.zpsm * 8));
 
 					switch (sel.ztst)
 					{
@@ -1200,9 +1200,6 @@ void GSDrawScanline::CDrawScanline(int pixels, int left, int top, const GSVertex
 				{
 					zs = zs.blend8(zd, zm);
 				}
-
-				if (sel.zclamp)
-					zs = zs.min_u32(VectorI::xffffffff().srl32(sel.zpsm * 8));
 
 				bool fast = sel.ztest ? sel.zpsm < 2 : sel.zpsm == 0 && sel.notest;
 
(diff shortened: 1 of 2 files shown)


refractionpcsx2 (approved):

AhayriSG:
SLUS21639 & NASCAR 08 NTSC
3db1f88: no problem
4689f7d: crash
3db1f88;
zclampfixup_3db1f88
4689f7d;
zclampfixup_4689f7d
3db1f88 emuLog
[emuLog_3db1f88.txt](https://github.com/PCSX2/pcsx2/files/8791017/emuLog_3db1f88.txt)
4689f
[emuLog_4689f7d.txt](https://github.com/PCSX2/pcsx2/files/8791020/emuLog_4689f7d.txt)
7d emuLog

tellowkrinkle:
Seems like that isn't related to this PR
The closest commit on master under 3db1f88e3cf460c86f40067c2eb3af92d59d301f is dc08858e0fea4f6655abeb1e4d26723fc7741b6a
For 4689f7d7cb80970fd8999568319b4388597ae88b it's c2638131b4bec6a6fc980bca820ef498a22542ca
Go compare those and then you can bisect properly

AhayriSG:
> Seems like that isn't related to this PR The closest commit on master under [3db1f88](https://github.com/PCSX2/pcsx2/commit/3db1f88e3cf460c86f40067c2eb3af92d59d301f) is [dc08858](https://github.com/PCSX2/pcsx2/commit/dc08858e0fea4f6655abeb1e4d26723fc7741b6a) For [4689f7d](https://github.com/PCSX2/pcsx2/commit/4689f7d7cb80970fd8999568319b4388597ae88b) it's [c263813](https://github.com/PCSX2/pcsx2/commit/c2638131b4bec6a6fc980bca820ef498a22542ca) Go compare those and then you can bisect properly

v1.7.2683: no problem
v1.7.2684: crash (again, SW only)

https://github.com/PCSX2/pcsx2/compare/v1.7.2683...v1.7.2684

refractionpcsx2:
yep, I actually looked in to this and told Tellow about it on discord, but apparently he missed it so I will paste it here..

@tellowkrinkle could use your advice, not sure if it's something you need to check or not. So it's when reading the output merge circuit, it does a massive DBY (as seen above) which is the offset vertically from the FBP to read the frame
and it's crashing the memory lookup
but I'm not sure if that's a result of this not handling it right, or the memory read not handling it right :/
I'm not sure if it should be wrapping on 2048 or something...

image

PR #5990, opened 2022-04-30
GameDB: Crash Bandicoot - Wrath of Cortex

RedDevilus:
### Description of Changes

Adds missing gamefixes that influenced random corruptions and the famous fog wall.
### Rationale behind Changes

Less tinkering
### Suggested Testing Steps

Test Crash Bandicoot wrath of cortex.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -9188,6 +9188,9 @@ SLES-50386:
   name: "Crash Bandicoot - Wrath of Cortex"
   region: "PAL-M6"
   compat: 5
+  gsHWFixes:
+    preloadFrameData: 1 # Fixes random texture corruptions such as horizontal lines.
+    cpuFramebufferConversion: 1 # Fixes fog wall.
 SLES-50390:
   name: "Driven"
   region: "PAL-M5"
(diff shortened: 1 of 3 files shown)


dio-gh:
Works as expected.

**Note:** the DirectX backends are still affected by the fog wall issue.

ps1freak26:
I tried it and it didn't work also Droid Void the mech section is screwed up you can't even see. Setting various crc hack levels and hw renderers didn't provide the result I expected.
[gsdump.zip](https://github.com/PCSX2/pcsx2/files/8598172/gsdump.zip)

ps1freak26:
Okay I made a mistake in Arctic Antics I set the resolution to 4x but when I set to 1x with crc hacks off the fog shows up but with 4x I get pink lines. It even fixes Droid Void in the gsdump.

dio-gh:
Makes sense, this PR is primarily aimed to fix the rendering at native res (1x) when using HW mode. It makes the CRC hacks unnecessary there - so much so that it's worse with them on, as you could see.

Upscaling introduces further rendering issues unfortunately. Using CRC hacks can help, but it seems like they cause severe issues on the level you sent a gsdump about (and this PR doesn't change that).

My advice for the time being, is that if you want to use upscaling, utilize skipdraw instead. It's pretty heavy handed, but it's the only way I could make that specific scene in your gsdump look bareable. I use 3-10 as the range, cuts a lot of stuff (basically all of the post processing and transparency effects), but it for sure gets rid of these corruptions.

Issue #5796, opened 2022-03-31
[BUG]: Software FMV - Automatic Renderer


RedDevilus:
### Describe the Bug

The software FMV switch does not work in the automatic renderer (GameDB or manually) but does in Vulkan, OpenGL or Direct3D11.

### Reproduction Steps

- Use default settings
- Launch a game that has SoftwareFMV (ex. Monster Rancher 4)

- Or enable manually the gamefix on a game with FMVs

### Expected Behavior

Correctly change from HW to SW for FMV.

### PCSX2 Revision

1.7.2548

### Operating System

Windows 10 (64bit)

PR #5993, opened 2022-05-01
Counters: Fix software renderer FMV fix with auto renderer

stenzek:
Fixes the issue where the use-software-renderer-for-FMVs gamefix didn't activate when using the automatic renderer.


RedDevilus:
Fixes: https://github.com/PCSX2/pcsx2/issues/5796

PR #5995, opened 2022-05-01
GS: Account for frame offset in output circuit

refractionpcsx2:
### Description of Changes
Account for offset in framebuffer when calculating height

### Rationale behind Changes
If there was a framebuffer offset and the framebuffer size exceeded the height of the screen, it would cut it off too early. Regression from #5981

### Suggested Testing Steps
test stuff to make suire the screen isn't cut off, especially Silent Hill 2

Fixes Silent Hill 2 bug reported here: https://forums.pcsx2.net/Thread-Bug-Report-Silent-Hill-2-NTSC-U


lightningterror (approved):
Silent Hill 2 works good now.

PR #5996, opened 2022-05-01
ThreadedFileReader: Fix possible race with synchronous reads

stenzek:
### Description of Changes

There's a potential for a race here when doing synchronous reads. Basically, another request can come in, after we release the lock, but before we store null to indicate we're finished. So, we do a compare-exchange instead, to detect when this happens, and if so, reload all the inputs and try again.

This happened **very rarely** when booting a game, and either resulted in the CRC being misdetected, or SYSTEM.CNF not being parsed correctly.

### Rationale behind Changes

Racey bad.

### Suggested Testing Steps

Test booting games repeatedly, make sure it parses SYSTEM.CNF correctly.

### DIFF
--- a/pcsx2/CDVD/ThreadedFileReader.cpp
+++ b/pcsx2/CDVD/ThreadedFileReader.cpp
@@ -70,22 +70,35 @@ void ThreadedFileReader::Loop()
 		if (m_quit)
 			return;
 
-		u64 requestOffset = m_requestOffset;
-		u32 requestSize = m_requestSize;
-		void* ptr = m_requestPtr.load(std::memory_order_relaxed);
-
-		m_running = true;
-		lock.unlock();
+		void* ptr;
+		u64 requestOffset;
+		u32 requestSize;
 
 		bool ok = true;
+		m_running = true;
 
-		if (ptr)
+		for (;;)
 		{
-			ok = Decompress(ptr, requestOffset, requestSize);
-		}
+			ptr = m_requestPtr.load(std::memory_order_acquire);
+			requestOffset = m_requestOffset;
+			requestSize = m_requestSize;
+			lock.unlock();
 
-		m_requestPtr.store(nullptr, std::memory_order_release);
-		m_condition.notify_one();
+			if (ptr)
+				ok = Decompress(ptr, requestOffset, requestSize);
+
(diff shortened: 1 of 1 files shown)


tellowkrinkle:
Wait are we issuing synchronous read requests while async read requests are still in flight?

Or is there a way for this race to happen even when only one read request is in flight at a time?

stenzek:
> Wait are we issuing synchronous read requests while async read requests are still in flight?

It's theoretically possible, but not happening at the moment. We only do the ISOFile parsing stuff at EE load time, and the guest CDVD isn't active then.

> Or is there a way for this race to happen even when only one read request is in flight at a time?

Yes. In particular:
 - ReadSync for 3 sectors.
 - First one decompresses on calling thread. Kicks ISO thread for readahead.
 - Second one is satisfied from the cache (larger chunk size). Kicks ISO thread for readahead.
 - Third one is out of range, and needs to be decompressed. ISO thread gets woken up somewhere in between setting `m_running` to true (after ptr load), and before null is stored to the request pointer. Calling thread writes pointer and size, ISO thread unconditionally writes null, and it gets lost.

tellowkrinkle:
Ahh it's the readahead

This part probably should have been inside of the `if (ptr)`, as it was supposed to be relying on the fact that a second read wouldn't be issued until the first completed
```cpp
m_requestPtr.store(nullptr, std::memory_order_release);
m_condition.notify_one();
```

If we do want to support sync and async at once with this, there's one more race to consider:
- Sync or async read finishes main read and starts readahead (`m_requestPtr` is null, `m_running` is true)
- Async read starts (`m_requestPtr` is set to the async read target)
- Sync read starts (`m_requestPtr` is set to the sync read target)
- Decompression thread notices `m_requestPtr` is non-null, finishes readahead, and reads the sync read's section
- `m_requestPtr` is set to null again, signaling read completion, but the async read never happened

Otherwise, we can throw an `assert(!m_requestPtr, "Multiple simultaneous reads is currently unsupported");` after acquiring the lock in `BeginRead` and `ReadSync` (which I probably should have put in from the start, but better late than never)

(Also I probably should have documented this a bit better, debugging lock free documentation free code isn't great... I'll do that after this merges)

PR #5997, opened 2022-05-01
GameDB: Sniper Elite + Naruto Shippuuden Ninja 3-5 + GT4

RedDevilus:
### Description of Changes

Adds FMV switch for both and adds special texture half pixeloffset for Sniper Elite.

Sniper Elite;
Before:

After:

### Rationale behind Changes

Less tinkering
### Suggested Testing Steps

Test relevant games.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -135,6 +135,8 @@ PAPX-90512:
   region: "NTSC-J"
   clampModes:
     vuClampMode: 2 # Text in GT mode works.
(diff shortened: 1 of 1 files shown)


Blackbird88:
I'm sorry for being late, but I had to finish a race. HPO Normal isn't enough for GT4 in the NTSC versions. It works in the interlaced modes, but causes shaking in 480p mode in the NTSC versions that can't be handled with de-interlacing. HPO Special works fine with 480p.

PR #5999, opened 2022-05-01
GS: Fix up SW FMV switch behaviour

### LINKED ISSUE #5998: GAMEDB: Problem with the SoftwareRendererFMVHack setting
### Describe the Bug

I was trying today the game Rygar. This game has the SoftwareRendererFMVHack setting enabled. When the game switches to SW renderer it never goes back to HW renderer after the FMV. I tried another game that uses the setting (Silent Hill 3 NTSC) and the same happens. I searched the github but didn't find any report of this, I don't know if it's me doing something wrong. If you press F9 the game goes into HW mode again but it never goes back to SW when an FMV plays. I used the OpenGL renderer in case that is relevant and I have the latest nightly release

### Reproduction Steps

Just play any game that has this setting enabled in the GameDB and confirm this behavior 

### Expected Behavior

Switch to software when an FMV plays and then go back to HW when the actual game starts or continues

### PCSX2 Revision

1.7.2684

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

### CPU

i5-4690

### GPU

NVIDIA GeForce GTX 1070

### GS Settings

3x internal resolution

### Emulation Settings

All defaults

### GS Window Screenshots

_No response_

### Logs & Dumps

_No response_

refractionpcsx2:
### Description of Changes
Fix Software FMV switch, remembering the old renderer and if it was switched.

### Rationale behind Changes
#5993 changed it so the GSConfig got updated on the switch, then it checked the GSConfig at the end of the FMV to see if it was configured to use a HW renderer, so it never came out of Software.

Also added some extra protection in case the renderer is changed during the FMV.

### Suggested Testing Steps
Test different renderers, Auto, Software, Specific API's with the SW FMV gamefix and make sure switching happens properly

Fixes #5998

### DIFF
--- a/pcsx2/Counters.cpp
+++ b/pcsx2/Counters.cpp
@@ -433,6 +433,8 @@ void frameLimitReset()
 extern uint eecount_on_last_vdec;
 extern bool FMVstarted;
 extern bool EnableFMV;
+
+static bool RendererSwitched = false;
 static bool s_last_fmv_state = false;
 
 static __fi void DoFMVSwitch()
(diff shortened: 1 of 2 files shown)


refractionpcsx2:
@stenzek if you have a better solution, I'm happy to kill this.

RedDevilus:
Seems to work for all the (hardware) renderers.

PR #6001, opened 2022-05-02
GS: Limit scissor optimisation, fix repeat condition

refractionpcsx2:
### Description of Changes
Only allow scissors to happen on single draws, and make sure we only modify the range on non-repeating textures.

### Rationale behind Changes
multiple draws (likely using it as a texture map) will possibly falsely limit the range, causing texture corruption (As seen in .hack infection)

### Suggested Testing Steps
Test stuff, make sure the textures aren't broken in SW and HW (more than they used to be)


stenzek (approved):
RE4's stupid shuffle is still getting clipped, so, lgtm

PR #6002, opened 2022-05-02
Threading: Make CPU usage measurements less garbage on Windows

stenzek:
### Description of Changes

Currently, we use `GetThreadTimes()` for measuring CPU usage of our various threads. It's.. not the best function, both in terms of precision, but also the frequency at which it is updated. It's pretty easy for our polling to fall out of phase with when the system updates it, which leads to spurious 0% CPU usage measurements, and/or it bouncing around +/- up to 20%.

I've replaced it with `QueryThreadCycleTime()`, which works **much** better. **Technically**, we shouldn't be relying on the frequency for this function (MSDN says you shouldn't), *but*, I'm guessing that's just because it's using `rdtsc` internally, and that's not guaranteed to be stable on older CPUs (pre-2008ish). Since we don't support anything older than that anyway, it should be fine for us. But probably worth testing on Intel just to be safe.

Further note: While this introduces a new abstraction (`ThreadHandle`), I got rid of the old timer-based class, and eventually merge `ThreadHandle` into `pxThread`. Once we drop the wx crap there, anyhow.

### Rationale behind Changes

Making it so we can actually measure CPU performance improvements on Windows.

### Suggested Testing Steps

Test AMD and Intel processors on Windows to make sure the CPU usage measurements are correct. I've done Windows and Linux on AMD Zen.

### DIFF
--- a/common/Darwin/DarwinThreads.cpp
+++ b/common/Darwin/DarwinThreads.cpp
@@ -94,6 +94,52 @@ u64 Threading::GetThreadCpuTime()
 	return us;
 }
 
+Threading::ThreadHandle::ThreadHandle() = default;
+
+Threading::ThreadHandle::ThreadHandle(const ThreadHandle& handle)
+	: m_native_handle(handle.m_native_handle)
+{
+}
+
+Threading::ThreadHandle::ThreadHandle(ThreadHandle&& handle)
+	: m_native_handle(handle.m_native_handle)
+{
+	handle.m_native_handle = nullptr;
+}
+
+Threading::ThreadHandle::~ThreadHandle() = default;
+
+Threading::ThreadHandle Threading::ThreadHandle::GetForCallingThread()
+{
+	ThreadHandle ret;
+	ret.m_native_handle = pthread_self();
+	return ret;
+}
+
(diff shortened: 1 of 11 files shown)


common/Darwin/DarwinThreads.cpp
```diff
@@ -94,6 +94,57 @@ u64 Threading::GetThreadCpuTime()
...
+
+u64 Threading::ThreadHandle::GetCPUTime() const
+{
+	return getthreadtime(pthread_mach_thread_np((pthread_t)m_native_handle));
+}
+
+void Threading::ThreadHandle::SetThreadName(const char* name) const
+{
+	// Doesn't appear to be possible to set another thread's name in OSX...
+}
```
tellowkrinkle, line 140: Is there any reason to use this over `SetNameOfCurrentThread` which works on more platforms?

stenzek: Originally my intent was to switch pxThread over to these wrappers, and I wrote the other implementations.. before I realized macos couldn't do it. Probably just easier to drop this one.

JordanTheToaster (approved):

Issue #3496, opened 2020-06-25
King of Fighters 98 Ultimate Match and 2002 Unlimited Match Tougeki Version 3D background issue


IAMNOTRANA:
***PCSX2 version:***
v1.6.0 

***PCSX2 options:***
Running on safest preset and still get weird glitch fest 3D background

***Plugins used:***
D3D 11 Default setting with no hardware hack enabled

***Description of the issue:***
KOF 98 UM and 2002 UM both game face issue with 3D background no matter what setting you use.

Here's some comparison of 2002 UM Tougeki Version (J) to actual PS2 machine

[https://cdn.discordapp.com/attachments/440195976880455680/725685861852905592/1.png](url)

KOF 98 UM (USA) also suffer from same problems but only with 3D background (you've to turn on 3D background from option and by default game runs on 3D instead of 2D background)

***How to reproduce the issue:***
KOF 2002 UM Tougeki Version (J
Boot up game > go practice > pick greece map and start. 
You can compare this with screenshot.

KOF 98 UM (USA)
Just boot up the game with any setting. (literally anything) then go option > graphic setting > set background to 3D (by default it should be 3D)
go practice mode and pick china 1-3 to see obvious shadow glitching and hills missing parts

***Last known version to work:***
uh I tried googling and it seems like this problem exist since bloody 2015 lol but never fixed.

***PC specifications:***
Laptop
i5 7200U
GTX 940mx 4gb 
8GB ram

lightningterror:
Is the issue present on latest dev build too?

RedDevilus:
Screenshot only works if you copy it in discord. I reuploaded it.
![image](https://user-images.githubusercontent.com/24227051/85720607-e0e91780-b6f0-11ea-876c-16fc29c354d8.png)

Silanda:
You can add King of Fighters '94 Rebout to this list too, and IIRC the 3D backgrounds in KoF XI, though admittedly I haven't tested it lately. Some of the backgrounds in Rebout are a complete mess, and the health bar appears to have a priority problem.

IAMNOTRANA:
> Is the issue present on latest dev build too?

Yeah I checked. Glitch is also present in dev build



> Screenshot only works if you copy it in discord. I reuploaded it.

Thank you.




[...]

PR #6003, opened 2022-05-02
GameDB: Add patches for remaining KOF series

Immersion95:
There's an [ongoing issue](https://github.com/PCSX2/pcsx2/issues/3496) with the KOF games, it is 95% fixed with [these patches](https://github.com/PCSX2/pcsx2/pull/4187).

_Fix for Depth precision.
Game fills upper 16bits of depth with 0xFFFF.
This results in a really high 32 bit value which is then converted to float because both hw and sw renderers lack double precision the lower 16 bits of the initial 32 bit value lose precision._

Unfortunately some main games were left out : 

PAL : KOF2002+KOF2003
NTSC-J : NGBC+KOF2002+KOF2003

Check : 
- Mexico flag in stage F in KOF2002
- Town 3 in KOF 2003
- Stage C in NGBC

All affected KOF games across all regions should now be covered (apart for special reissues or unknown versions which are pointless).

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -15281,9 +15281,35 @@ SLES-53380:
 SLES-53381:
   name: "King of Fighters 2002"
   region: "PAL-E"
+  patches:
+    CF1D71EE:
+      content: |-
+        author=kozarovv (adapted by Immersion95)
+        // Fix for Depth precision.
+        // Game fills upper 16bits of depth with 0xFFFF.
+        // This results in a really high 32 bit value which is then converted to float
+        // because both hw and sw renderers lack double precision the lower 16 bits of the initial 32 bit value lose precision.
+        patch=1,EE,000ffc00,word,f88a0000
+        patch=1,EE,000ffc04,word,a080000b
+        patch=1,EE,000ffc08,word,08056677
+        patch=1,EE,001599D4,word,0803ff00
+        patch=1,EE,001599D8,word,4bea497d
 SLES-53382:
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
I find it suspicious (and quite unlikely) that all the modifications needed to be done at the same addresses.  Did you test these?

Immersion95:
> I find it suspicious (and quite unlikely) that all the modifications needed to be done at the same addresses across different games, let alone regions. Did you test these?

Hi, yes I tested this and it works. It's not the same adresses though. You can check on the stages that I mentionned.

refractionpcsx2:
> > I find it suspicious (and quite unlikely) that all the modifications needed to be done at the same addresses across different games, let alone regions. Did you test these?
> 
> Hi, yes I tested this and it works. It's not the same adresses though. You can check on the stages that I mentionned.


I meant that 60% of your patches start at address 000ffc00 and end at 000ffc08

Immersion95:
> > > I find it suspicious (and quite unlikely) that all the modifications needed to be done at the same addresses across different games, let alone regions. Did you test these?
> > 
> > 
> > Hi, yes I tested this and it works. It's not the same adresses though. You can check on the stages that I mentionned.
> 
> I meant that 60% of your patches start at address 000ffc00 and end at 000ffc08

I didn't create anything, I just adapted the patches from kozarovv https://github.com/PCSX2/pcsx2/pull/4187/files for the missing regions. They all start and end at these adresses in his ones.

refractionpcsx2:
oh okay

PR #6005, opened 2022-05-02
GS: Adjust the Crash Bandicoot: WoC CRC hack

dio-gh:
### Description of Changes
Follow-up to #5990.

Makes the Crash Bandicoot: Wrath of Cortex CRC hack only take effect when using the DX11 backend, or when upscaling. On DX11 this is still needed due to recursive draws not working out there, and due to upscaling introducing a different kind of artifact to the affected graphical effects (striping).

~~I've also adjusted the CRC Full mode so that it applies on DX12 as well, not just DX11.~~ (Dropped the commit, as I've confirmed the game to be fine on DX12.)

### Rationale behind Changes
Improves defaults.

Right now the CRC hack is active on HW mode no matter what by default, cutting out significant graphical effects (fog rendering, water caustics), and causing severe, game breaking graphical glitches on later levels of the game.

With this change, both the aforementioned graphical effects are retained and the game breaking graphical glitches are avoided in HW mode, unless the backend selection or the upscaling choice make this unfeasible.

### Suggested Testing Steps
Test the game in question if you wish, but should be fine.


ps1freak26:
Everything works as expected expect the corrupt squares on the bottom right in jungle rumble at native res. CRC hacks and GS hacks set to auto it looks correct as 1x native and behaves correctly with 2x or more.

dio-gh:
> except the corrupt squares on the bottom right in jungle rumble at native res

I cannot reproduce it, could you show a screenshot and further instructions?

ps1freak26:
https://user-images.githubusercontent.com/91406481/166390413-b3a5acfb-2601-44b4-8af3-ae5badc2bff7.mp4

I have a old nvidia graphics driver 497.29 and this happens on opengl and vulkan. This level is playable though the level with the glitches.

dio-gh:
I see, those are not present for me.

How does the level look on v1.7.2689 with CRC hacks off, native, on those renderers? What happens if you enable Frame Buffer Conversion, and then Preload Frame Data one after the other?

The thing I'm trying to find out here is that whether my PR introduces a regression, or if these settings and the crc hack off is worse for native res on those backends. I have a reasonable strong suspicion that they aren't, and that your graphics driver may be the culprit, so that's what I'm aiming to verify here.

ps1freak26:
I tried an old build v1.7.2613 and it does the same this build does. I will put blame on my graphics drivers but I will test a pc with an amd igpu to see if this problem happens there. I found this issue also on the coco levels.

dio-gh:
In the meantime I had Jordan test a gsdump I made, the flicker seems to be an NVIDIA-specific host rendering issue. I get some artifacts on Vulkan myself with AMD, but it looks starkly different, so it's probably a different issue.

Either way, this not a regression from my PR, and the visual quality is arguably more accurate and still acceptable this way. I asked Jordan to open a ticket (#6006) for the nvidia backends problems, it's not an actual emulation but an implementation problem.

So yeah, as far as I'm concerned, this is out of scope for me.

dio-gh:
Dropped the commit which made CRC_Full include DX12, as during testing it became clear that DX12 is capable of rendering the effects even without ROV. However, it does seem to be a recursive draw, which coincidentally also means that the NVIDIA square artifacts issue is progressively going to get worse over time during playthrough.

Until #6006 is addressed I can't recommend merging this, would be too damaging on NVIDIA. Changing this into a draft until then.

dio-gh:
@ps1freak26 by complete chance I have the same driver on my nvidia machine (my laptop), and I cannot reproduce the issue there either. What's your GPU model? Mine's a GTX 1060 (so Pascal family).

ps1freak26:
I have a 1660 super (Turing) and I set blending accuracy to basic.

Codexmsters:
#6180

Codexmsters:
#6180

lightningterror:
Closing as it's not currently being worked on and there may be other issues.

ps1freak26:
The fractional upscale of 1.5x got rid of this problem. It's very bizarre but it works.

PR #6008, opened 2022-05-03
GS: Adjust default renderer based on vendor and feature level.

lightningterror:
### Description of Changes

GS: Adjust default renderer based on vendor and feature level.
Vulkan default:
Nvidia Maxwell gen2 and higher.
OpenGL default:
From Fermi to Mawell gen1.
Everything else Direct3D11.

GS-d3d11: Get rid of vendor check in device.
Rely on the vendor check we already use for best renderer selection.
### Rationale behind Changes

Automatically pick the better best renderer for the user.
### Suggested Testing Steps

Make sure Automatic works as described above.

### DIFF
--- a/pcsx2/GS/GSUtil.cpp
+++ b/pcsx2/GS/GSUtil.cpp
@@ -194,16 +194,14 @@ GSRendererType GSUtil::GetPreferredRenderer()
 	// Mac: Prefer Metal hardware.
 	return GSRendererType::Metal;
 #elif defined(_WIN32)
+	if (D3D::ShouldPreferRenderer() == D3D::Renderer::Vulkan)
+		return GSRendererType::VK;
 #if defined(ENABLE_OPENGL)
-	// Windows: Prefer GL if available.
-	if (D3D::ShouldPreferD3D())
-		return GSRendererType::DX11;
-	else
+	else if (D3D::ShouldPreferRenderer() == D3D::Renderer::OpenGL)
 		return GSRendererType::OGL;
-#else
-	// DX11 is always available, otherwise.
-	return GSRendererType::DX11;
 #endif
+	else
+		return GSRendererType::DX11;
 #else
 	// Linux: Prefer GL/Vulkan, whatever is available.
 #if defined(ENABLE_OPENGL)
--- a/pcsx2/GS/Renderers/DX11/D3D.cpp
+++ b/pcsx2/GS/Renderers/DX11/D3D.cpp
@@ -69,44 +69,68 @@ namespace D3D
 		return adapter;
 	}
 
-	bool IsNvidia(IDXGIAdapter1* adapter)
+	u8 Vendor()
 	{
+		auto factory = CreateFactory(false);
+		auto adapter = GetAdapterFromIndex(factory.get(), 0);
+
 		ASSERT(adapter);
 
 		DXGI_ADAPTER_DESC1 desc = {};
 		if (FAILED(adapter->GetDesc1(&desc)))
 		{
 			fprintf(stderr, "D3D: failed to get the adapter description\n");
-			return false;
+			return VendorID::Unknown;
 		}
 
-		// NV magic number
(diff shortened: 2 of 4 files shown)


refractionpcsx2:
Is there a vendor check we can do on intel too to make that selection more cool? I presume Vulkan is preferred on Linux too unless the hardware sucks?

refractionpcsx2:
Also I feel we should move it out of DX, maybe put it in GSUtil? having a function in DX11 to tell it to use Vulkan seems kinda silly :P Plus if we make a function for Linux it would be good to have them both in the same place and switch between them.

lightningterror:
Thing is code is d3d specific, putting it in GSUtil will make things more ugly, iirc it was like that before kojin cleaned it up.

refractionpcsx2:
Sure, but then DX is selecting OpenGL and VK which also makes no sense.  I guess you could put the vendor lookup in DX, but have the "should prefer" function in GSUtil.  That's IMO anyway.

fantesykikachu:
I think it should stay in D3D as it calls the actual D3D API; but I do agree that it should be as explicit as possible, so a switch and vendorID should be added for intel, and for the Render enum Default should be replaced with Direct3D_11 and Direct3D_12 should be added.

Also IIRC, a few years back Dolphin was recommending DX12 on Intel, but I don't know how that stands now with the PCSX2 DX12 backend.

JordanTheToaster (approved):

tellowkrinkle:
Vulkan (barriers on) can get some pretty terrible performance on AMD cards
That hasn't really changed

e.g. VP2 is 30fps on Vulkan on a Radeon Pro 5600m (vs 180 on D3D)

refractionpcsx2:
yes AMD should probably default to either DX12 or DX11 (depending on feature level).

lightningterror:
> Vulkan (barriers on) can get some pretty terrible performance on AMD cards That hasn't really changed
> 
> e.g. VP2 is 30fps on Vulkan on a Radeon Pro 5600m (vs 180 on D3D)

Ouch, that's a lot, and here i thought it's acceptable to make Vulkan default for Amd.

Nerboruto:
maxwell gen 1 support latest vulkan and latest driver... why not?

JordanTheToaster:
> maxwell gen 1 support latest vulkan and latest driver... why not?

Maxwell 1 has a weaker feature set than 2 so OpenGL makes more sense for them.

refractionpcsx2:
Please note this is just the "Automatic" setting, nothing is stopping you ignoring it and selecting Vulkan yourself, if you disagree.

Nerboruto:
> > maxwell gen 1 support latest vulkan and latest driver... why not?
> 
> Maxwell 1 has a weaker feature set than 2 so OpenGL makes more sense for them.

https://en.wikipedia.org/wiki/Feature_levels_in_Direct3D

gen 1 lack Conservative rasterization, Rasterizer-ordered views and View instancing tier2...

lightningterror:
> > maxwell gen 1 support latest vulkan and latest driver... why not?
> 
> Maxwell 1 has a weaker feature set than 2 so OpenGL makes more sense for them.

Because detection is based on feature level, if I enable Vulkan on 11.0, it will also enable on Fermi which doesn't support Vulkan, and Kepler which I don't know if it properly works.

RedDevilus:
> maxwell gen 1 support latest vulkan and latest driver... why not?

The goal of this PR isn't to produce perfect detection but to have better detection (on the cautious side) you can always pick the renderer manually if you have a more uncommon setup and it can in the future be reviewed for improvements. So in this case more people will get more Vulkan detection than OpenGL for performance reasons and it has pretty much feature-parity with blending for the rest it will be Direct3D with some exceptions for OpenGL as stated in the PR opening.

RedDevilus (approved):

Nerboruto:
thank you all for the explanations .. it was not my intention to make controversy ...

PR #6016, opened 2022-05-05
Misc: Move pxThread and friends to gui

stenzek:
### Description of Changes

Second part of #6015, which gets rid of a bunch of unused threading crap, and moves the worst of it into `gui`, so it doesn't get pulled in for Qt.

### Rationale behind Changes

Getting rid of `wxThread` and other nonsense in the Qt build.

Out of sight, out of mind.

### Suggested Testing Steps

Just use the application as normal, and report anything which breaks. e.g. suspending/resuming.


arcum42:
Seems like this crashes if I compile as wx on linux, and try to switch between Vulkan and OpenGL if emulation has been started (I could switch before running an iso, but after running it, whether I left it running, paused it, or even shut down emulation, it hangs on switching renderers.)

When it hangs, it becomes unresponsive, if emulation was running, it stops, and the windows stop updating.

Works fine in qt, and on master.

refractionpcsx2:
Can confirm the hang switching renderer ( I went from Vulkan to OGL on the FFX intro), can confirm it doesn't happen on #6015

stenzek:
Hang and render switch issues should be resolved

refractionpcsx2:
The hang is fixed, however now if you swap renderer, pressing Escape no longer responds, sorry :(

Can confirm you can close the window manually and once you resume escape then responds again.

PR #6023, opened 2022-05-06
Qt: Rework display widget swapping

stenzek:
### Description of Changes

I really didn't want to have to make things work this way (messing with the central widget), but it seems to be the only way which works with newer version of wayland/gnome-shell/kwin....

Also fixed the long-running issue where fullscreen wouldn't be exited when displaying the "confirm exit" popup, corrected the window title when not using render-to-main, stopped things blowing up when you closed the non-render-to-main window, and added a confirm dialog for swapping discs from the game list (when it wasn't initiated from file->change disc).

### Rationale behind Changes

Reliability.

### Suggested Testing Steps

Check disc changing, fullscreen/render-to-main switching, view->game list, view->system display, etc. I've done a bit and I think I've nailed all the edge cases, but another set of eyes is always good.

### DIFF
--- a/pcsx2-qt/CMakeLists.txt
+++ b/pcsx2-qt/CMakeLists.txt
@@ -30,6 +30,7 @@ target_sources(pcsx2-qt PRIVATE
 	QtUtils.cpp
 	QtUtils.h
 	SettingWidgetBinder.h
+	GameList/EmptyGameListWidget.ui
 	GameList/GameListModel.cpp
 	GameList/GameListModel.h
 	GameList/GameListRefreshThread.cpp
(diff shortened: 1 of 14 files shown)


RedDevilus:
image
image

Probably my scuffed setup with laptop lid closed and attached to another display, all with different DPI and resolutions.

stenzek:
@RedDevilus Should be fixed now, hopefully.

RedDevilus:
:) looks to work nicely now

image

JordanTheToaster (approved):

RedDevilus (approved):

PR #6040, opened 2022-05-09
3rdparty/SDL2: Bump to 2.0.22 release

stenzek:
### Description of Changes

Fixes a crash on shutdown in Windows::Gaming::Input when controllers are connected. Crash manifests as a hang, because it tries to fire our already-torn-down pagefault handler.

### Rationale behind Changes

Making Qt not hang on shutdown.

### Suggested Testing Steps

Already tested, but @JordanTheToaster was the one who noticed this in the first place, so it would be good to get confirmation that the issue is fixed on his setup too.


qurious-pixel:
Bump Linux build to 2.0.22?

Currently on `SDL2-2.0.16`

stenzek:
Not sure what you mean. Qt builds the in-tree version, wx is unaffected by this change.

PR #5534, opened 2022-02-15
[WIP] GS: improve write overlap detection in TC.

iMineLink:
### Description of Changes

Please, consider this WIP.

Improve the write overlap detection in GS HW TC.
Avoid overwriting target FBW on EE write.
Allow dirty rectangles to overlap tex in rt search (update target if needed).
Rewrite target update logic to preserve EE writes mapping with underlying GS memory region (EE write and target can have different BW).
Allow tex in rt logic to be applied on same BP case (this enables more accurate dirty target and wrapping targets handling, even if the resulting offset, if valid, will start from 0,0).

Fixes #5530.
Fixes #4858.
Fixes Indiana Jones and the Emperor's Tomb map.

### Rationale behind Changes
In Jak X, if the initial FMV is skipped before the Naughty Dog logo is shown, the writes for the X of the Jak X logo and the split screen config occur in the middle of an existing target.
Before this PR those writes were not correctly registered, causing a missing update of the texture which resulted in black X and black split screens.

### Suggested Testing Steps
Test Jak X, check the X of the Jak X logo and the split screen when skipping the initial FMV.
Do generic testing for other games.

PR #6046, opened 2022-05-09
[RFC] GS/TextureCache: Update each subrectangle of dirty targets separately

stenzek:
This is more of an RFC than an actual change which should be merged at the moment.

While looking into the Dark Cloud regression, I ended up barking down the wrong tree with dirty areas in targets, and noticed that we update the union of all dirty areas on update, instead of just the actual areas which were written to. Which would mean we'd be potentially trashing parts of it, as the local memory wasn't in sync.

There's a few ways which we could go about this, but the simplest (and upscaling compatible) is what I noted in the source, take the union, decode the whole thing, and then copy/stretch the bits out which we want.

Question is, is it worthwhile? Hence RFC :)

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
+++ b/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
@@ -2145,16 +2145,8 @@ void GSTextureCache::Target::Update()
 {
 	Surface::UpdateAge();
 
-	// FIXME: the union of the rects may also update wrong parts of the render target (but a lot faster :)
-	// GH: it must be doable
-	// 1/ rescale the new t to the good size
-	// 2/ copy each rectangle (rescale the rectangle) (use CopyRect or multiple vertex)
-	// Alternate
-	// 1/ uses multiple vertex rectangle
-
-	GSVector4i unscaled_size = GSVector4i(GSVector4(m_texture->GetSize()) / GSVector4(m_texture->GetScale()));
-	GSVector4i r = m_dirty.GetDirtyRectAndClear(m_TEX0, GSVector2i(unscaled_size.x, unscaled_size.y));
-
+	const GSVector4i unscaled_size = GSVector4i(GSVector4(m_texture->GetSize()) / GSVector4(m_texture->GetScale()));
+	const GSVector4i r = m_dirty.GetDirtyRect(m_TEX0, GSVector2i(unscaled_size.x, unscaled_size.y));
 	if (r.rempty())
 		return;
 
@@ -2163,7 +2155,7 @@ void GSTextureCache::Target::Update()
 	{
 		// do the most likely thing a direct write would do, clear it
 		GL_INS("ERROR: Update DepthStencil dummy");
-
+		m_dirty.clear();
 		return;
 	}
 	else if (m_type == DepthStencil && g_gs_renderer->m_game.title == CRC::FFX2)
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
@iMineLink I feel you're probably one of the most qualified to answer this :)

@tellowkrinkle too?

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -2213,21 +2206,55 @@ void GSTextureCache::Target::Update()
...
 
-	// Copy the new GS memory content into the destination texture.
-	if (m_type == RenderTarget)
+	// The idea: If we have more than one dirty rectangle, we upload the whole thing, and then
+	// copy/upscale the specific areas we want into it.
+	if (m_dirty.size() > 1)
 	{
-		GL_INS("ERROR: Update RenderTarget 0x%x bw:%d (%d,%d => %d,%d)", m_TEX0.TBP0, m_TEX0.TBW, r.x, r.y, r.z, r.w);
+		const GSVector2i bs = GSLocalMemory::m_psm[m_TEX0.PSM].bs;
+		for (const auto& dr : m_dirty)
```
iMineLink, line 2214: I like this.
This is also kind of part of #5534 (https://github.com/PCSX2/pcsx2/pull/5534/files#diff-a08fccf34ca3d5af6f2ba46d887e2f60833415531482f5ac08a9f4848b9f8f45R1758).

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -2213,21 +2206,55 @@ void GSTextureCache::Target::Update()
...
+		{
+			// dirty area in unscaled space
+			GSVector4i drc(dr.GetDirtyRect(m_TEX0));
```
iMineLink, line 2217: Just to clarify a point, the `GSDirtyRect::GetDirtyRect` method used here (and it was the same before these changes) is a shortcut that properly works only in some cases.

Given the lack of pitch conversion you need to have buffer width compatibility here (between the draw and the target), also this code supposes that the writes occur with compatible memory mapping format (#5534 tries [and fails] to surpass this).
There's an issue with the pitch (`m_TEX0.TBW`) attached to our framebuffers: the original one is overwritten with this of the last write that is reported in the m_dirty of the same target (possibly trashing the old value) without any pitch conversion occurring in the process.
I fear that by being more precise here with the uploads, these limitations will result in some issues.

Also there's a chance that another draw "overwrites" the local data but it is not reported to the actual target nor its dirty rectangles list, hence even with this upgraded logic, some issues could persist due to the `m_dirty` referencing invalid data (`GSTextureCache::InvalidateVideoMem` not really invalidating). The only sane way to solve this properly seems to keep a data structure tracking each page/(block?) state (like I tried to do in my branch, if you wanna take a look https://github.com/iMineLink/pcsx2/blob/gs/hw-tc-page-ownership/pcsx2/GS/Renderers/HW/GSTextureCache.h#L159).

In conclusion, this is a step forward towards compatibility in my opinion as the upload is restricted to the dirty areas and not to a generic union, but it could make more obvious other limitations of the current caching system.

Consider making the optimizations done in the `GSTextureCache::Source::Write` to merge togheter the dirty rectangles and reduce the number of uploads.

iMineLink (commented):
Nice idea! Tried commenting the changes and giving a more general overview on some issues you might step into by touching these delicate code bits.

PR #6033, opened 2022-05-08
GS:SW: Use doubles for Z

tellowkrinkle:
### Description of Changes
Makes the SW renderer use doubles for Z

<del>Has a noticeable (~3%) performance hit, so leaving this a draft for now, will only use it for large z values in the future</del> I think we've gotten performance to an acceptable level

### Rationale behind Changes
Reduce Z fighting

### Suggested Testing Steps
Test KOF, other zequal games / games with z fighting

PR #6047, opened 2022-05-09
GS:SW: Clamp Z with unsigned clamp

tellowkrinkle:
### Description of Changes
Z is unsigned, so clamp it unsigned

### Rationale behind Changes
Correctness

### Suggested Testing Steps
Not sure any games hit this with the zequal test in place (which clamps outside of the JITed code), but removing zequal breaks trapt [trapt_zoverflow_zclamp.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/8653065/trapt_zoverflow_zclamp.gs.xz.zip) (see #6033)


refractionpcsx2:
Wasn't it set to signed because it moved the value back when it overflowed to try and get it in to range? I presume that's not needed anymore?

tellowkrinkle:
If the top bit was set, that would make it negative in the eyes of a signed min, which would mean that it would never be edited (since a negative number is definitely less than a positive number)

So if that's what it was supposed to do, that wasn't what it was actually doing
Also only the JIT had this, not the C rasterizer

PR #6052, opened 2022-05-10
Update DEV9.h

Hinkiii:
### Description of Changes
The minimum HDD size changed to 8 from 40 GB.

### Rationale behind Changes
I see no reason to change the limit from the original plugin as not everybody wants/needs 40 GB taken up when many things don't take up nearly as much space.

### Suggested Testing Steps
Creating a HDD file in the config prompt.

pcsx2x64-avx2-dbg_cbhDdHUzlk
image

### DIFF
--- a/pcsx2/DEV9/DEV9.h
+++ b/pcsx2/DEV9/DEV9.h
@@ -52,7 +52,7 @@ bool rx_fifo_can_rx();
 #define HDD_DEF "DEV9hdd.raw"
 #endif
 
-#define HDD_MIN_GB 40
+#define HDD_MIN_GB 8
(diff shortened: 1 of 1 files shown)


lightningterror:
See #4442

Hinkiii:
> See #4442

Could keep the default to 40 and/or make a way to where the sliders minimum is 40 and you can manually type it. It's kind of annoying having to use an old pcsx2 environment for creating smaller images.

refractionpcsx2:
why do you want a smaller image? If you want to save disc space, NTFS compress it, it will take the size down to 0 (if it's blank)

PR #6066, opened 2022-05-12
Qt: Add auto updater

stenzek:
### Description of Changes

See title. Large number of lines because of pulling in the LZMA/7Z SDK.

Windows only for now, but could be expanded out to Mac without too much work.

I've tested as much as I can (update extracting, downloading, etc) without this in releases, but that would be the next step. It'll only download releases with a "Qt" tag, and AVX2 if the running executable was AVX2.

### Rationale behind Changes

It's useful.

### Suggested Testing Steps

Difficult to test this when it's not in the release download list, but I'm not well versed in how all that operates, so I'm not adding it.


refractionpcsx2:
well, since it just tells me that this isn't an official build (because it's a PR), that makes it even harder to test xD

YOLO push it is

PR #6070, opened 2022-05-12
Qt: Hook up screenshot button.

arcum42:
### Description of Changes
Hooks up the screenshot button/menu item in QT.

### Rationale behind Changes
The screenshot button should work.

### Suggested Testing Steps
Take a screenshot.

### DIFF
--- a/pcsx2-qt/EmuThread.cpp
+++ b/pcsx2-qt/EmuThread.cpp
@@ -896,7 +896,8 @@ BEGIN_HOTKEY_LIST(g_host_hotkeys)
 DEFINE_HOTKEY("Screenshot", "General", "Save Screenshot", [](bool pressed) {
 	if (!pressed)
 	{
-		// TODO
+		Host::AddOSDMessage("Saved Screenshot.", 10.0f);
+		GSmakeSnapshot(EmuFolders::Snapshots.ToString().char_str());
 	}
 })
 DEFINE_HOTKEY("ShutdownVM", "System", "Shut Down Virtual Machine", [](bool pressed) {
--- a/pcsx2-qt/MainWindow.cpp
+++ b/pcsx2-qt/MainWindow.cpp
@@ -45,6 +45,8 @@
 #include "Settings/InterfaceSettingsWidget.h"
 #include "SettingWidgetBinder.h"
 
+extern u32 GSmakeSnapshot(char* path);
+
 static constexpr char DISC_IMAGE_FILTER[] =
 	QT_TRANSLATE_NOOP("MainWindow", "All File Types (*.bin *.iso *.cue *.chd *.cso *.gz *.elf *.irx *.m3u *.gs *.gs.xz);;"
 									"Single-Track Raw Images (*.bin *.iso);;"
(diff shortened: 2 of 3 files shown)


refractionpcsx2:
Still not working here.

JordanTheToaster:
Does not work for me.

arcum42:
Alright, suppose I'll need to build qt on Windows and test it then. Worked fine when I tried it in Linux.

arcum42:
Assuming I can actually get the Qt version to compile in Windows, that is. It wasn't picking up on the copy of Qt 6.2 I had installed, so we'll see if updating helps...

refractionpcsx2:
in windows I think you need to use the 6.3 pack we provide here: https://github.com/PCSX2/pcsx2-windows-dependencies/releases/tag/2022-04-13

arcum42:
Ah, fun, no picking it up from environment variables. Alright, grabbing it from there, then.

refractionpcsx2:
yeah if you extract the version folder in to 3rdparty/qt it should auto pick it up

arcum42:
Yep, got it. Compiled, copied over my bios and a copy of Dark Cloud, showed the toolbar, started Dark Cloud, hit the screenshot button, and it saved a screenshot into snaps, so it seems like it's working for me in windows as well...

refractionpcsx2:
can you download the windows artifact and extract it to a new folder so all the ini's etc are reset?

I certainly can't get it to work... I don't have media keys or anything, but pressing F8 does nothing, even though it's set up in the hotkeys.

arcum42:
This is the screenshot button in the gui next to cheats, not F8.

refractionpcsx2:
oh right, do you wanna hook up the hotkey as well then? :P I didn't even know there was a button lol

refractionpcsx2:
Can confirm the button works, now I know exactly what you've hooked up :P (Guess I should have read the first post)

arcum42:
Yeah, I was mostly concerned about getting buttons that didn't do anything in the gui working. Didn't even realise the hotkey wasn't working. :P

I'll take a look.

arcum42:
There we are, now the hotkey works, by my copying the same two lines of code I used for the button in place of a "TO DO" comment...

refractionpcsx2:
excellent lol, well if it works it works :D

arcum42:
It's mostly a matter of popping up something saying we saved it, then calling the same function wx uses. I just had to figure out the qt version of getting the path instead of the wx version. Honestly, figured it would be easy enough, just no one'd gotten to it.

refractionpcsx2:
Yeah, tbh I need to start chipping away at Qt stuff. I just don't know Qt at all, hate GUI stuff and I'm lazy, it's the tri-force of disappointment :D

arcum42:
I learned a little bit of Qt for working with dobiestation, and I already know how to do gui stuff in gtk and wx, plus a little bit of imgui, and I used to know Carbon/Cocoa on the Mac. The more gui systems you know, the easier it tends to be to pick up another one...

I'll admit I'm not even sure what that cheats button that doesn't work is supposed to do, though. :P

refractionpcsx2:
Not a clue, maybe it opens a cheats window? Idk :P @stenzek is the man to ask

arcum42:
Beyond the scope of this pr anyways, I suppose. As is fixing the silly string conversions we do here. Lets turn this path into a wxString, which we turn into a char*, which we then turn into a std::string in the actual function. Great...

refractionpcsx2:
Yeah, I presume all that will come with the pass of ripping out wxString, since the goal is to be able to cut WX out of the Qt build completely.

JordanTheToaster (approved):

PR #6072, opened 2022-05-13
CI: Upload Qt in releases

stenzek:
See title. Not sure if this is the optimal way of doing this, and/or if there's additional changes required on the site/API side.


xTVaser:
I believe this will "work" in it's current state, but it feels kinda forced in (using what normally would be the 32bit/64bit part as Qt to differentiate it).

Looking at what you had to change to support an SSE4 build, it should be very easy to add to https://github.com/PCSX2/pcsx2/pull/6076 the result should be cleaner as well as the other improvements I made along the way.

stenzek:
Yeah, abusing the platform is definitely not ideal, just seemed like the easiest way of sneaking it in there, without changing all the other scripts too.

xTVaser:
Qt SSE4 builds added to that PR, if they look good to you, I think this can be closed.

stenzek:
Superseded by #6076.

PR #6073, opened 2022-05-13
README.md: update website links

SgtSteel:
Updated the downloads link to point to the new page (old one gave 404)
Apparently there is no screenshot page on the new website, so the  "Check out the PCSX2 website" link has been changed to point to the pcsx2.net homepage.


refractionpcsx2:
Thanks, it could be worth changing the "Development builds" line underneath too, maybe combine it with the download link line since both are available there now :)

refractionpcsx2:
Thanks :)

PR #6074, opened 2022-05-13
GameDB: Adds SkipDraw ranges to 'Tales of the Abyss'

Trixarian:
### Description of Changes
GameDB: Adds SkipDraw ranges to 'Tales of the Abyss'

### Rationale behind Changes
This aims to correct the graphical issues introduced by the use of Upscaling and Half-Pixel Special (Texture) in Tataroo Valley, Engave, Keterburg, Keterburg bay, and outside the castle in Baticul (among others). This does introduce a minor graphical glitch in the Desert Oasis, but should be an acceptable compromise for everything else it corrects

### Suggested Testing Steps
Test suggested areas for any graphical issues

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -1081,6 +1081,8 @@ SCAJ-20163:
   region: "NTSC-Unk"
   gsHWFixes:
     halfPixelOffset: 2 # Fixes ghosting.
+    skipDrawStart: 100 # Fixes black line grids in skyboxes and ghosting in some instances
+    skipDrawEnd: 361 # Fixes black line grids in skyboxes and ghosting in some instances
 SCAJ-20164:
   name: "Kingdom Hearts II"
   region: "NTSC-Unk"
@@ -30692,6 +30694,8 @@ SLPM-66897:
   region: "NTSC-J"
   gsHWFixes:
     halfPixelOffset: 2 # Fixes ghosting.
+    skipDrawStart: 100 # Fixes black line grids in skyboxes and ghosting in some instances
+    skipDrawEnd: 361 # Fixes black line grids in skyboxes and ghosting in some instances
 SLPM-66898:
   name: "Spectral Gene [Limited Edition]"
   region: "NTSC-J"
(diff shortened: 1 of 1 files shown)


Ryudo300:
Not sure if this is still an issue with current Skipdraw settings, but in the past it caused everything in the game to become dark, forcing you to increase brightness to the max. (And as a result this also affected the GUI + Cutscenes, becoming overly bright).

Trixarian:
> Not sure if this is still an issue with current Skipdraw settings, but in the past it caused everything in the game to become dark, forcing you to increase brightness to the max. (And as a result this also affected the GUI + Cutscenes, becoming overly bright).

From what I tested, it does randomly cause some areas to become darker when loaded for the first time using manual hacks. Once loaded it doesn't seem to cause any further instances for me, but may warrant more in-depth testing to be sure

RedDevilus:
```
    skipDrawStart: 100 # Fixes black line grids in skyboxes and ghosting in some instances
    skipDrawEnd: 361 # Fixes black line grids in skyboxes and ghosting in some instances
```

That is very aggressive 
Skipdraw 1-2 is fine but beyond that causes multiple visual issues.

lightningterror:
Not in favor adding these since they are upscaling issues.

Trixarian:
> ```
>     skipDrawStart: 100 # Fixes black line grids in skyboxes and ghosting in some instances
>     skipDrawEnd: 361 # Fixes black line grids in skyboxes and ghosting in some instances
> ```
> 
> That is very aggressive Skipdraw 1-2 is fine in normal circumstances (not saying it will help that game) but beyond that causes multiple visual issues.

I tried lower values than the ones suggested on the game's wiki page, but it doesn't seem to correct the issues that that range seems to fix. It makes Engave playable without needing to drop to Native resolution and fixes some of the blur in Tataroo Valley  which Half-Pixel Special (Texture) seems to introduce there. I have noted that using an end of 354 corrects the visual issue it introduces in Desert Oasis, but breaks it everywhere else

lightningterror:
I will close this as it's not a good idea to add skipdraw for upscaling issues, this will break native res from looking proper.

PR #6078, opened 2022-05-14
GS: Redo a bunch of the ignore offset code

refractionpcsx2:
Plus clean up a bunch of stuff which isn't needed

### Description of Changes
Removes a bunch of dead code and made the ignore offsets code more reliable.

### Rationale behind Changes
There was a regression initially with Devil May Cry, however I also noticed a regression with Wipeout, so redesigning this code and cleaning up old dead stuff sorted that out.

### Suggested Testing Steps
Test games with strange display heights, make sure everything is OK.

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -510,16 +510,6 @@ GSVector2i GSState::GetResolution()
 		total_rect.w = std::min(total_rect.w, resolution.y);
 		resolution.x = total_rect.z;
 		resolution.y = total_rect.w;
-
-		// When we're ignoring offsets we need to account for pictures which are usually offset up off the screen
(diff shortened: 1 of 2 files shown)


pcsx2/GS/GSState.cpp
```diff
@@ -459,9 +459,9 @@ GSVector4i GSState::GetDisplayRect(int i)
 	// Set up the display rectangle based on the values obtained from DISPLAY registers
 	rectangle.left = offsets.x;
 	rectangle.top = offsets.y;
-
-	rectangle.right = rectangle.left + width;
-	rectangle.bottom = rectangle.top + height;
+
```
lightningterror, line 462: Tab.

lightningterror (approved):
Just needs testing.

PR #6079, opened 2022-05-15
Core: Misc cleanup and refactoring of HW functions

MrCK1:
### Description of Changes
- This PR addresses an issue where a shutdown function for the internal HW existed but did not get called from anywhere. Also refactored the code to group like components together a bit more.

- Removes some additional code and includes that are no longer necessary

### Rationale behind Changes
Code cleanliness

### Suggested Testing Steps
Tested PS2 and PSX modes, shutdown and reset. Make sure no glitches occur switching between games ect.


pcsx2/Sifcmd.h
```diff
@@ -16,178 +16,12 @@
 #ifndef __SIFCMD_H__
 #define __SIFCMD_H__
 
-/* from sifcmd.h */
```
stenzek, line 19: Would it worth keeping these around for documentation purposes? Or perhaps moving them into the documentation directory.

refractionpcsx2: Looks like all that information can be found here https://github.com/ps2dev/ps2sdk/blob/master/ee/kernel/include/sifrpc.h

So probably not a lot of reason to keep it hanging around in the code.

Issue #6062, opened 2022-05-11
[BUG]: Iffy texture replacement in Secret Agent Clank


ZwipZwapZapony:
### Describe the Bug

When using PCSX2's texture dump/replacement feature, Secret Agent Clank (SCES-55496) still manages to load some original/unmodified textures.

Not all textures can be replaced (some can't even be dumped!), and for those that can, they may revert to the original texture after some time (and may return to the replaced texture after some more time).

\
Related to (and may have the same root cause as?) #5862.

### Reproduction Steps

- Play through a level three times while dumping textures (while rebooting PCSX2 between each attempt).
- Rename the game's textures' "dumps" folder to "replacements".
- Make a 64x64 PNG image, with RGB colour 64,192,64 and 127 alpha, and save it as "green.png" in the replacements folder.
- Use `FOR %A IN (*) DO COPY /B /Y "green.png" "%~A"` in a command prompt within the replacements folder, to replace every dumped texture with the flat, green texture.
- Delete "green.png", to ensure that it doesn't affect PCSX2.
- Launch PCSX2, enable loading textures, and play through the level again.

\
Many textures will now be flat green, but some will also be unmodified (either immediately or after some time).

### Expected Behavior

All textures should have been flat green permanently.

### PCSX2 Revision

v1.7.2735

### Operating System

Windows 10 (64bit)

### CPU

Intel Core i5-8300H @ 2.3GHz

### GPU

NVIDIA GeForce GTX 1050

### GS Settings

![image](https://user-images.githubusercontent.com/14360666/167878891-3c261492-f9d7-4a9e-9b4e-f8c9b1c72f17.png)

### GS Window Screenshots

|Description|Screenshot|
|---|---|
|Staring into a corner, loading a cutscene, and skipping it (to reload textures):|![image](https://user-images.githubusercontent.com/14360666/167879208-4798f5c1-2b28-442e-8d58-16e51595f5cb.png)|
|Turning 180 degrees, which loads more content (and reverts some of the changed textures in the process):|![image](https://user-images.githubusercontent.com/14360666/167879307-ea87db98-d85e-4ef0-8558-44e6163f9725.png)|
|Turning 180 degrees again, back to the corner. Compare it to the first screenshot.|![image](https://user-images.githubusercontent.com/14360666/167879677-e6199853-cfd8-4062-abc9-4e281aa58faa.png)|

refractionpcsx2:
[...]

PR #6080, opened 2022-05-15
GS/TextureCache: Handle edge cases of replacements with paltex enabled

stenzek:
### Description of Changes

From the source comment:

```
// Using paltex without full preloading is a disaster case here. basically, unless *all* textures are
// replaced, any texture can get populated without the hash cache, which means it'll get partial invalidated,
// and unless it's 100% removed, this partial texture will always take precedence over future hash cache
// lookups, making replacements impossible.
```

also handles

```
// There's an edge case here; when there's multiple textures with the same vram data, but different
// palettes, if we don't replace all of them, the first one to get loaded in will prevent any of the
// others from getting tested for replacement. So, disable paltex for the textures when any of the
// palette variants have replacements.
```

Basically, it's a crappy solution (disabling paltex when full preloading isn't used), but it's better than forcing preloading, as that's problematic in more games.

### Rationale behind Changes

Fixing #6062 again.

### Suggested Testing Steps

Test replacements with paltex on.

### DIFF
--- a/pcsx2/GS/GSExtra.h
+++ b/pcsx2/GS/GSExtra.h
@@ -114,10 +114,11 @@ extern Pcsx2Config::GSOptions GSConfig;
 
 // Maximum texture size to skip preload/hash path.
 // This is the width/height from the registers, i.e. not the power of 2.
+static constexpr u32 MAXIMUM_TEXTURE_HASH_CACHE_SIZE = 10; // 1024
 __fi static bool CanCacheTextureSize(u32 tw, u32 th)
 {
-	static constexpr u32 MAXIMUM_CACHE_SIZE = 10; // 1024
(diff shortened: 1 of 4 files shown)


ZwipZwapZapony:
Texture loading seems fixed. :thumbsup:

But, texture *dumping* is still iffy if you're using GPU palette conversion, you're not using full texture pre-loading, and you're not loading any custom textures. Secret Agent Clank's PAL language select screen still only dumps the first flag in that case. (That's fixed by simply loading any replacement texture now, though, so it's not a big issue.)

stenzek:
Should be fine now, just had to extend the check to dumping too. It'll basically force preloading and paltex off when dumping, and only preloading when the texture has a replacement with another palette hash.

Basically, replacement and dumping is incompatible with paltex.

ZwipZwapZapony:
Both replacing and dumping textures seems fixed now. 👍

PR #6086, opened 2022-05-16
GameDB: Modify .hack // games to use 'dot hack' and '-'

LargoUsagi:
… characters for consistency and compatability with the QT UI

### Description of Changes
Changed all .hack games to use dot hack and no longer use / characters

### Rationale behind Changes
It broke functionality in the QT UI and there was inconsistencies across the games

### Suggested Testing Steps
Tested via replacing the file locally on the latest QT build from GH Actions and attempting to set a cover image for .hack part 1


refractionpcsx2:
Thanks, in future please keep commit messages concise and prefix with GameDB: if it's for the GameDB please :)

Thanks for the contribution

PR #6089, opened 2022-05-16
3rdparty/imgui: Update to 60bea05

stenzek:
### Description of Changes

Also redoes the renderers, so that multiple textures can be used, and gets rid of a bunch of unused code.

### Rationale behind Changes

Mainly for the multi textures, which I need for the big picture UI.

### Suggested Testing Steps

Test OSD on all renderers.


labib2002:
"big picture UI" is the upcoming UI design, right?

PR #6091, opened 2022-05-17
CMake: Set fp-contract=fast

tellowkrinkle:
### Description of Changes
Allows multiplies and adds to be merged into fma instructions

### Rationale behind Changes
Clang defaults to off, gcc defaults to fast.  MSVC supposedly defaulted on until VS2022, and I've seen it do the optimization but not very consistently.
- Gives a microscopic increase in speed to the software renderer
- Reduces the impact of Double Z (#6033) in AVX2 Clang to 1-2%

### Suggested Testing Steps
Watch things build

### DIFF
--- a/pcsx2/CMakeLists.txt
+++ b/pcsx2/CMakeLists.txt
@@ -39,8 +39,13 @@ if(NOT TOP_CMAKE_WAS_SOURCED)
 	It is advice to delete all wrongly generated cmake stuff => CMakeFiles & CMakeCache.txt")
 endif()
 
-if(NOT MSVC)
+if(MSVC)
+	if (MSVC_VERSION GREATER_EQUAL 1930)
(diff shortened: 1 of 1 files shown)


F0bes:
Just wondering, couldn't find an answer online. What is the tradeoff when setting this to fast?

refractionpcsx2:
Did the R&C TEst

CMake: 157-160fps
Master-AVX2: 160-163fps

tellowkrinkle:
> Just wondering, couldn't find an answer online. What is the tradeoff when setting this to fast?

It's an accuracy thing.  fma instructions don't round between multiplying and adding, so `a * b + c` with separate instructions is `round(round(a * b) + c)`, while with fma it's `round(a * b + c)` where `round` applies whatever floating point rounding mode you've configured in the MXCSR.  So technically you're doing a different calculation, though it's very similar.  And no, obviously this won't affect the JIT (it's mostly for GS).

PR #6094, opened 2022-05-18
Optionally use system libs

tellowkrinkle:
### Description of Changes
Adds the ability to use system libs for a number of dependencies

You can now specify `-DUSE_SYSTEM_LIBS=ON` to force PCSX2 to require system libs and fail if they can't be found.  You can also specify `-DUSE_SYSTEM_LIBS=OFF` to not attempt to use system libs at all.  The default (`AUTO`) uses system libs if they can be found, and bundled libs otherwise.  In addition, there are separate overrides for each of the optionally-bundled libraries.

### Rationale behind Changes
Users may prefer system libs for a number of reasons, so it's nice to have the option.

### Suggested Testing Steps
- Make sure save states are loadable with system libzip (tested by me, but might as well check)
- Make sure system SDL2 works in Qt (not tested by me)
- Make sure cmake rejects outdated packages (SDL2 is probably easiest to test, plenty of linux distros have a < 2.0.22 copy)
  - In this case, make sure the build continues with the bundled SDL in the default configuration, and fails if you specify `-DUSE_SYSTEM_SDL2=ON`


arcum42:
Yeah, it's generally best for all third party libraries to at least optionally use system libraries. Some distributions really want everything to be using system libraries anyways, and bundled libraries can fall out of date and cause potential security issues. (Though I'm not sure who'd be trying to hack your computer through pcsx2.)

Besides, the less files that have to compile when I build pcsx2, the better. :P

cmake/SearchForStuff.cmake
```diff
@@ -145,7 +145,7 @@ else()
 	check_lib(SAMPLERATE samplerate samplerate.h)
 
 	if(NOT QT_BUILD)
-		check_lib(SDL2 SDL2 SDL.h PATH_SUFFIXES SDL2)
+		find_optional_system_library(SDL2 3rdparty/sdl2 2.0.12)
```
arcum42, line 148: Wrong version here.. Should be 2.0.22, not 12.

tellowkrinkle: 2.0.12 is the minimum version needed for wx builds
Though we could require 2.0.22 anyways for better controller support

arcum42: Ah, alright, didn't realise we weren't going to 22 across the board (though with wx going away, we will soon enough anyways).

arcum42: Can verify that SDL2 does not build with -DUSE_SYSTEM_SDL2=ON on a qt build with sdl lower than 2.0.22 then, in which case I've successfully checked all the suggested testing steps. I just initially tried it with a wx build and sdl 2.0.20 and it compiled...

arcum42 (approved):

PR #6097, opened 2022-05-19
Everything: Remove a **lot** of wx, and px nonsense 

stenzek:
### Description of Changes

 - common has no wx left except for Path.
 - pcsx2core only has it in a few places (memory cards and path related stuff).

### Rationale behind Changes

Get lost wx, an emu core should never depend on a semi-dead UI framework library.

### Suggested Testing Steps

Test your usual workflow.


arcum42:
Actually, seems like moving those lines takes care of it for *gcc*, but not for clang, which still has the SafeArray linker problems, as well as a bunch of things about fmt, for some reason.

pcsx2/DEV9/Win32/tap-win32.cpp
```diff
@@ -355,7 +355,7 @@ bool TAPGetWin32Adapter(const std::string& name, PIP_ADAPTER_ADDRESSES adapter,
 
 	//We must be bridged
 	Console.WriteLn("DEV9: Current adapter is probably bridged");
-	Console.WriteLn(L"DEV9: Adapter Display name: %s", pAdapterInfo->FriendlyName);
+	Console.WriteLn("DEV9: Adapter Display name: %s", pAdapterInfo->FriendlyName);
```
TheLastRar, line 358: `pAdapterInfo->FriendlyName` is a wide string

I think wx let me get away with using `%s` instead of `%ls`

pcsx2/DEV9/Win32/tap-win32.cpp
```diff
@@ -407,7 +407,7 @@ bool TAPGetWin32Adapter(const std::string& name, PIP_ADAPTER_ADDRESSES adapter,
 				PIP_ADAPTER_ADDRESSES potentialAdapter = FindAdapterViaIndex(AdapterInfoReduced.get(), row.HigherLayerInterfaceIndex);
 				if (potentialAdapter != nullptr)
 				{
-					Console.WriteLn(L"DEV9: %s is possible bridge (Check 1 passed)", potentialAdapter->Description);
+					Console.WriteLn("DEV9: %s is possible bridge (Check 1 passed)", potentialAdapter->Description);
```
TheLastRar, line 410: `potentialAdapter->Description` is a wide string

pcsx2/DEV9/Win32/tap-win32.cpp
```diff
@@ -491,7 +491,7 @@ bool TAPGetWin32Adapter(const std::string& name, PIP_ADAPTER_ADDRESSES adapter,
 									wil::unique_cotaskmem_string dispName;
 									hr = component->GetDisplayName(dispName.put());
 									if (SUCCEEDED(hr))
-										Console.WriteLn(L"DEV9: %s is possible bridge (Check 2 passed)", dispName);
+										Console.WriteLn("DEV9: %s is possible bridge (Check 2 passed)", dispName);
```
TheLastRar, line 494: `dispName` is a wide string

pcsx2/DEV9/Win32/tap-win32.cpp
```diff
@@ -504,7 +504,7 @@ bool TAPGetWin32Adapter(const std::string& name, PIP_ADAPTER_ADDRESSES adapter,
 
 									hr = component->GetDisplayName(dispName.put());
 									if (SUCCEEDED(hr))
-										Console.WriteLn(L"DEV9: %s is bridge (Check 3 passed)", dispName);
+										Console.WriteLn("DEV9: %s is bridge (Check 3 passed)", dispName);
```
TheLastRar, line 507: ditto

pcsx2/ShiftJisToUnicode.cpp
```diff
@@ -844,9 +844,9 @@ wchar_t ShiftJIS_ConvertChar(const u8* input, int& used)
 	}
 }
 
-wxString ShiftJIS_ConvertString( const char* src )
+std::string ShiftJIS_ConvertString( const char* src )
 {
-	wxString result;
+	std::string result;
```
tellowkrinkle, line 849: We append wchars to this, you'll need to convert to UTF-8 either on append or after
e.g.
```cpp
static void AppendUCS2(std::string& str, u16 c)
{
	if (c & 0xf800)
	{
		str.push_back(0xe0 | (c >> 12));
		str.push_back(0x80 | ((c >> 6) & 0x3f));
		str.push_back(0x80 | (c & 0x3f));
	}
	else if (c & 0xff80)
	{
		str.push_back(0xc0 | (c >> 6));
		str.push_back(0x80 | (c & 0x3f));
	}
	else
	{
		str.push_back(c);
	}
}
```

pcsx2/DEV9/Win32/tap-win32.cpp
stenzek: %ls is potentially problematic, as _scprintf with a %ls string that has any codepoints above 255 will return -1, and blow up the string conversion. I'll just convert them to UTF-8 first.

stenzek:
Closing as #6099 which includes this is being merged.

Issue #6075, opened 2022-05-14
[BUG]: Regression: Tales of the Abyss


Ryudo300:
### Describe the Bug

Upscaling causes an unplayable blurry effect in certain towns where bloom is very active (Engeve being a prime example).

Each time you increase the upscale multiplier (2x to 3x, etc), it causes the blurred effect to fill the screen further to the point where it fully covers the screen.

After testing out various builds, it was discovered to be a regression caused by https://github.com/PCSX2/pcsx2/pull/5419.

Example Screenshots:

Native: (Bug not present)
![pcsx2x64-avx2_2022-05-14_03-55-29](https://user-images.githubusercontent.com/10039778/168406599-fb771eb3-9edf-4a96-b766-ad957da3d821.png)

2x Native:
![pcsx2x64-avx2_2022-05-14_03-56-30](https://user-images.githubusercontent.com/10039778/168406622-2ce6db6a-f8f7-4bd7-a600-86fba6049740.png)

3x Native:
![pcsx2x64-avx2_2022-05-14_03-56-52](https://user-images.githubusercontent.com/10039778/168406631-acad3d09-ed56-4021-b446-b9ec9bf82fe9.png)

### Reproduction Steps

One of the earlier locations you can spot this in, is in the town called Engeve. Simply enter the town and you'll notice these effects. Inside the buildings it's fine again (presumably because there's no bloom effect in there). 

### Expected Behavior

Expected behavior should be as seen below here in a screenshot taken on PCSX2 1.6.0

![pcsx2_2022-05-14_04-04-00](https://user-images.githubusercontent.com/10039778/168406834-1ad8b089-35cb-4169-a718-d4b6dd13f112.png)

### PCSX2 Revision

v1.7.2.2744

### Operating System

Windows 10 (64bit)

### CPU

AMD Ryzen 5 3600

### GPU

MSI GeForce GTX 1660 SUPER

### GS Settings

**Internal Resolution:** 3x Native
**Anisotropic Filtering:** Off
**Dithering:** Off
**Mipmapping:** Automatic (Default)
**CRC Hack Level:** Automatic (Default)
**Blending Accuracy:** Basic (Recommended)
**Texture Preloading:** None (Default)

### Logs & Dumps

GSdump (Using 3x native)

[...]

Issue #6054, opened 2022-05-10
Star Ocean, conservative buffer allocation off causes black screen on initial load file screen. 


BParks21:
### Describe the Bug

When conservative buffer allocation is unchecked when pressing continue to load your save file the screen is black, It's functional and working you just can't see anything. Unchecking it instantly fixes the issue. This also happens at native resolution when the setting is grayed out. I use to run this game always with it unchecked and it was fine because now having it enabled causes this weird graphical bugs randomly. See screenshot. 

### Reproduction Steps

Run the game at native res or at a higher res with the option unchecked

### Expected Behavior

No black screen when loading the save file.

### PCSX2 Revision

Nightly v1.7.2731

### Operating System

Windows 10 (64bit)

### CPU

i7 8086k

### GPU

Nvidia GTX 1080ti

### GS Settings

Opengl default. 

### Emulation Settings

Default

### GS Window Screenshots

![Screenshot (17)_LI](https://user-images.githubusercontent.com/12180630/167696201-7e0d8136-3d62-4ada-bf0e-5edff53b4807.jpg)

lightningterror:
Can you post a gsdump?

BParks21:
I always forget is there a hotkey for that?

BParks21:
Also real quick note here, something I just discovered. It only happens at native resolution if conservative buffer allocation is unchecked before enabling native resolution. Which is strange to me. If it's left checked and then native res is enabled and the game is started the black screen does not occur. Also this game has a progressive screen mode so triangle and X are required to be held down on boot to enable it otherwise the game looks blurry. If you guys are unaware.

lightningterror:
> I always forget is there a hotkey for that?

https://forums.pcsx2.net/Thread-How-to-create-a-proper-GS-dump

BParks21:
> > I always forget is there a hotkey for that?
> 
> https://forums.pcsx2.net/Thread-How-to-create-a-proper-GS-dump
[Desktop.zip](https://github.com/PCSX2/pcsx2/files/8664109/Desktop.zip)

lightningterror:
Can you take another dump from the actual scene that has the issue?

BParks21:
Oh well I took the dump where the screen is black. I can take a dump where the graphical bug occurs.

BParks21:
Here is the new dump
[snaps.zip](https://github.com/PCSX2/pcsx2/files/8664334/snaps.zip)

lightningterror:
[...]

PR #6102, opened 2022-05-19
gs-ogl: fix depth copy shader in stretch rect.

iMineLink:
### Description of Changes
Title says it.

### Rationale behind Changes
Less bugs is better.

### Suggested Testing Steps
Check that the issue #6075 on OGL is on par with the other hardware renderers.
Verify if it impacts #6054 somehow.


lightningterror (approved):
Looks like it was only missing on gl, maybe the recent depth changes that were done.

PR #6110, opened 2022-05-22
gs-hw: fix tales of abyss upscaled rendering.

iMineLink:
### Description of Changes
Fixed the upscaled rendering of tales of abyss, closing #6075 and improving it further.
Commits are mostly related to gs hardware texture cache stuff: after adding target rescaling/resizing, it could have happened that the texture reference stored in the source object became invalid before the rendering occurred; This is now fixed, and this prevents the fullscreen blurring in the tales of abyss game when upscaling.

Also, I found out that for the dummy texture case, the scale of the texture stored in the source object was wrong, and this is fixed (but likely does not make a difference).

I added some assertions in the source lookup code to ensure the scale is properly set on the texture and some other source properties are properly set too.

Finally, I removed the can upscale hack of the tales of abyss game which was just introducing ghosting in the upscaled rendering.

### Rationale behind Changes
Less bugs and hacks is better.

### Suggested Testing Steps
Test tales of abyss and verify that #6075 is fixed and that the rendering quality is improved when upscaling.
Test games that were using the dummy textures system (`GSRendererHW::IsDummyTexture` says "Jak series/tri-ace game") and verify that they work properly.
Test some games that use channel shuffling and verify that they work properly.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
@@ -3696,8 +3696,6 @@ GSRendererHW::Hacks::Hacks()
 	m_oi_list.push_back(HackEntry<OI_Ptr>(CRC::BurnoutGames, CRC::RegionCount, &GSRendererHW::OI_BurnoutGames));
 
 	m_oo_list.push_back(HackEntry<OO_Ptr>(CRC::BurnoutGames, CRC::RegionCount, &GSRendererHW::OO_BurnoutGames));
-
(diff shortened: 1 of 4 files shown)


Tzerinas:
Could this PR affect HPO? Because from my short testing the HPO moves perform a bit differently.

refractionpcsx2:
Went through my dumps, didn't notice anything worse or improved (don't have tales of the abyss)

iMineLink:
I'd say HPO can work differently now because the rendering is always done at upscaled resolution.

lightningterror (approved):
Tested dumps that use channel shuffle, didn't spot issues.

PR #6114, opened 2022-05-22
Qt: Cancel game list scanning when VM starts 

stenzek:
### Description of Changes

Because of CDVD global state, and the fact that we hijack CDVD for getting the game details, we have to cancel the scan. Otherwise, the scanner will clash with the game's CDVD access.

### Rationale behind Changes

Don't want users crashing the software.

### Suggested Testing Steps

Already tested the evil scenario of starting a game while scanning :-)


RedDevilus (approved):
Fixes 3 issues I had:
- While scanning for games and browsing for example BIOS it would hard-crash
- BIOS and Games unresponsive for booting (especially after scanning the first time)
- Minor other related issues that causes hard-crashing or even latency when starting game

PR #6121, opened 2022-05-22
ci: fix windows release asset structure

xTVaser:
### Description of Changes

Ensures all files in windows release assets are in the top level of the `7z` file (not within a sub folder)

### Rationale behind Changes

The auto-updater requires the release to be structured a certain way, this was a mistake that wasn't caught before the CI cleanup merge.

### Suggested Testing Steps

- Tested here, check that the assets look as expected https://github.com/xTVaser/pcsx2-rr/releases/tag/v1.4.5


RedDevilus:
image
Should fix the updater (not the 0.00MB)

PR #6124, opened 2022-05-22
Build: Update build.sh to remove a no longer used flag and add a few more.

arcum42:
### Description of Changes
Removed use system yaml, as it is no longer used.
Added flags for forcing using system libraries on, forcing using the bundled libraries, and doing a qt build.

### Rationale behind Changes
No longer needed flags shouldn't be in the build script. Qt is going more towards being used, so having an option here for it seems reasonable. Having switches for forcing using the system libraries or the bundled ones seems reasonable as well.

### Suggested Testing Steps
Compile with build.sh.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -158,9 +158,11 @@ for ARG in "$@"; do
         --rel|--release     ) flags="$flags -DCMAKE_BUILD_TYPE=Release" ; build="$root/build_rel";;
         --prof              ) flags="$flags -DCMAKE_BUILD_TYPE=RelWithDebInfo"; build="$root/build_prof";;
         --strip             ) flags="$flags -DCMAKE_BUILD_STRIP=TRUE" ;;
-        --use-system-yaml   ) flags="$flags -DUSE_SYSTEM_YAML=TRUE" ;;
         --asan              ) flags="$flags -DUSE_ASAN=TRUE" ;;
         --gtk2              ) flags="$flags -DGTK2_API=TRUE" ;;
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -158,9 +158,11 @@ for ARG in "$@"; do
         --rel|--release     ) flags="$flags -DCMAKE_BUILD_TYPE=Release" ; build="$root/build_rel";;
         --prof              ) flags="$flags -DCMAKE_BUILD_TYPE=RelWithDebInfo"; build="$root/build_prof";;
         --strip             ) flags="$flags -DCMAKE_BUILD_STRIP=TRUE" ;;
-        --use-system-yaml   ) flags="$flags -DUSE_SYSTEM_YAML=TRUE" ;;
         --asan              ) flags="$flags -DUSE_ASAN=TRUE" ;;
         --gtk2              ) flags="$flags -DGTK2_API=TRUE" ;;
+        --qt6               ) flags="$flags -DQT_BUILD=TRUE" ;;
+        --use-system ) flags="$flags -DUSE_SYSTEM_LIBS=ON" ;;
+        --use-bundled ) flags="$flags -DUSE_SYSTEM_LIBS=OFF" ;;
```
tellowkrinkle, line 165: Align these two like the rest of the options?
```suggestion
        --use-system        ) flags="$flags -DUSE_SYSTEM_LIBS=ON" ;;
        --use-bundled       ) flags="$flags -DUSE_SYSTEM_LIBS=OFF" ;;
```

arcum42: Take care of.

PR #6130, opened 2022-05-22
Qt: Disable/remove unused buttons.

lightningterror:
### Description of Changes

Qt: Disable/remove unused buttons
Not implemented yet.
Disable:
Pause On Start,
Pause On Focus Loss,
Hide Mouse Cursor,
Disable Window Resizing.

Remove:
Cheats menu and toolbar.
### Rationale behind Changes

Unused/ non functional.
### Suggested Testing Steps

GUI still works and looks good, compiles.

### DIFF
--- a/pcsx2-qt/MainWindow.cpp
+++ b/pcsx2-qt/MainWindow.cpp
@@ -594,11 +594,9 @@ void MainWindow::updateEmulationActions(bool starting, bool running)
 	m_ui.actionReset->setEnabled(running);
 	m_ui.actionPause->setEnabled(running);
 	m_ui.actionChangeDisc->setEnabled(running);
-	m_ui.actionCheats->setEnabled(running);
 	m_ui.actionScreenshot->setEnabled(running);
 	m_ui.actionViewSystemDisplay->setEnabled(starting_or_running);
 	m_ui.menuChangeDisc->setEnabled(running);
-	m_ui.menuCheats->setEnabled(running);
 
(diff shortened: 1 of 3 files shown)


BParks21:
So there will be no way to hide the mouse cursor automatically?

RedDevilus:
It's a temporary deletion but that doesn't mean it won't be inplemented in the future.

BParks21:
> It's a temporary deletion but that doesn't mean it won't be inplemented in the future.

I see, thanks for the reply.

stenzek:
Imo better to leave them there until they're implemented.

BParks21:
> Imo better to leave them there until they're implemented.
> 
> Edit: Another option could just be to leave them perma-disabled until they're hooked up.

Was just curious about the option because with it checked or unchecked or functional, the mouse still disappears when not active in fullscreen with wxWidgets builds.

lightningterror:
> Edit: Another option could just be to leave them perma-disabled until they're hooked up.

I thought about that but then people would be asking how to enable them, so I think it's better to remove until they are implemented.

arcum42:
I think permadisabled's probably better because that way we remember that they are supposed to be there and need to be implemented. It'd be easier to forget if the buttons not there at all.

stenzek:
The cheats item could probably go, but the others are reasonably straightforward to implement so disabling for them is better (might be able to do them this weekend).

PR #6131, opened 2022-05-22
QtHost.cpp: Fixed Resources and Config dir

kenshen112:
### Description of Changes

Fixed location of Resources Dir to proper usr/share location and XDG location of PCSX2 to .config

### Rationale behind Changes

To fix linux folder locations to match XDG standard.

### Suggested Testing Steps

See if it builds


arcum42:
If you look under "Files Changed", you can see some Codacy notes that should be addressed.

stenzek:
I specifically made this match the wx behaviour. I don't really see why it needs to be changed (and break existing installs)?

kenshen112:
Because it doesn't work. It's making the PCSX2 config dir in my home dir instead of .config and all package installs are broken because it's not looking for resources in /usr/share like it should be

stenzek:
Package mode doesn't exist in Qt.

And dropping the folder in the folder in the home directory unless XDG_CONFIG_DIR was defined was what wx did for me too.

kenshen112:
-DPACKAGE_MODE still seems to work in QT and even seems to offer an install target. And When that was enabled. When XDG_CONFIG_DIR wasn't defined it would still result in .config at least on my system, though I think that may have been distro dependent.

kenshen112:
I just did a force push to fix that @arcum42

arcum42:
Yeah, the "xdg_config_home &&" in the else statement was what it was having an issue with initially...

kenshen112:
Ok there we go

arcum42:
In the else if section, we already know it isn't empty from the first if, so the "!xdg_config_home.empty() &&" at the beginning is redundant.

kenshen112:
Fixed. Anything else?

stenzek:
Like I said, package mode isn't intended for Qt. Personally I am 100% against putting stuff in /usr/share; it's literally "shared data", but PCSX2's resources aren't used by anything else.

You're **still** changing the fallback when the environment variable isn't present, which will break my installs at least. Plus assigning it to a std::string, when getenv returns null, will crash.

arcum42:
Looking at "EmuFolders::Resources = Path::Canonicalize(Path::Combine(EmuFolders::AppRoot, PCSX2_APP_DATADIR "/resources"));", what we're saying here is "take the path to where the binary is, add the path to the data directory, then add "/resources" on the end, which doesn't seem like it'll come up with any sort of a valid path...

stenzek:
Fwiw, the long term plan was to move all of the resources into the exe with qrc for released builds (to make it easier to have multiple versions alongside), so there wouldn't be any need for this /usr/share nonsense in the first place in that scenario.

kenshen112:
The result of that on my machine is `/usr/share/PCSX2/resources`

kenshen112:
And the long term plan is fine. But right now, I'm just aware that it's at least preventing the packages like `pcsx2-git` from functioning proper and setting it back the way it was in WX repairs that https://github.com/PCSX2/pcsx2/blob/ef32b49e50b94b027cc9a1da7b7efe5ef14c5e9d/pcsx2/gui/AppConfig.cpp#L199

stenzek:
Then that's your problem, not ours, we don't support packages, only the appimage (which I still have to create a qt script for).

What arcum42 said is correct, the logic is clearly wrong here, and like I said, irrelevant long term when we move to qrc.

kenshen112:
Then I can fix that logic real quick. That said, where would you prefer that folder to go if not /usr/share?

And I don't think that supporting AppImage as the main method of distribution is the best call for anything like pcsx2. Mainly because of a few reasons that when compiled you run in a virtual environment that A. misses the whole config scheme that linux puts in place for packages. You miss any optimizations from distro's like PopOS and their gaming focus with the packages you add into the AppImage, and there's no permanence with AppImages so settings, if it's deleted or lost, are lost as well which while I'm sure can be worked around doesn't need to be with a package.

If it's an issue with package maintainers, might I suggest there's other CI's like the OpenSuseBuild system that can output packages for multiple distros?

RedDevilus:
> Then I can fix that logic real quick. That said, where would you prefer that folder to go if not /usr/share?
> 
> And I don't think that supporting AppImage as the main method of distribution is the best call for anything like pcsx2. Mainly because of a few reasons that when compiled you run in a virtual environment that A. misses the whole config scheme that linux puts in place for packages. You miss any optimizations from distro's like PopOS and their gaming focus with the packages you add into the AppImage, and there's no permanence with AppImages so settings, if it's deleted or lost, are lost as well which while I'm sure can be worked around doesn't need to be with a package.
> 
> If it's an issue with package maintainers messing things up or not listening or anything like that, might I suggest there's other CI's like the OpenSuseBuild system that can output packages for multiple distros? They auto-update and don't need much maintenance past that point. Plus, I'm all ears here my dudes I'll do whatever the hell you want

I get you don't find AppImages as graceful or against your nature as a package creator, but you are still expecting the Window users majority to need to know all the ins and outs of all the Linux distributions then with no idea if they are correct and wrong, artificially inflating and making it harder to support. 

Nothing against giving people the tools to update via PPA/Flatpak/... but AppImages will streamline the process and if you don't like the current AppImage you can always try to improve it to use less/more packages. You could even do a poll amongst Linux users if they like the current situation AppImage + compiling options for others. However, I never see anyone complaining what for example RPCS3 uses, which are effectively AppImages too.

I do still believe we can find some middle ground here that would satisfy the majority of users.

stenzek:
Sigh. This isn't really a debate I want to have...

> And I don't think that supporting AppImage as the main method of distribution is the best call for anything like pcsx2. 

AppImages are the best way of distribution for application developers, because they're (mostly) consistent across platforms. You don't have packagers messing with things, and breaking stuff like you've done here (e.g. you completely ignored my statement that assigning null to std::string will crash).

> A. misses the whole config scheme that linux puts in place for packages.

That makes no sense.

> You miss any optimizations from distro's like PopOS and their gaming focus with the packages you add into the AppImage

What optimizations? Breaking stuff like you've done here? We're not deliberately shipping unoptimized builds; an appimage is going to run just as well as a self-compiled build. Even more so for PCSX2, when the majority of CPU time is spent in jitted code.

> there's no permanence with AppImages so settings, if it's deleted or lost, are lost as well which while I'm sure can be worked around doesn't need to be with a package.

What? They're saved in $HOME, just like a self-compiled build.

To be brutally honest, I'd rather you didn't package/distribute qt at the moment anyway; **especially** if you're making changes like this which are broken/crash, it's just going to create headaches for us.

arcum42:
I'm assuming what you were intending to do was more along the line of "EmuFolders::Resources = Path::Canonicalize(Path::Combine(PCSX2_APP_DATADIR, "/resources"));"", though I haven't tested that.

(I personally would've probably made qt a separate package, or waited a bit to switch over to it until the kinks were worked out. Making initial test builds available is really the same as being ready to distribute as the main daily package...)

kenshen112:
Would it be better as a cmake option? And that's fine and dandy if you would rather wait on the main distribution. I was planning on doing some testing packages outside of the main repos

kenshen112:
And it is a separate package on arch. The main repo distributes WX builds.

kenshen112:
Is that more to your guys liking? That still works on my end.

stenzek:
I'm still -1 on hardcoding paths into the binary. It makes it unrelocatable (and thus unable to have multiple builds on the same system concurrently), that's the whole reason it uses relative paths here in the first place.

IMO, just wait until we merge all resources into the exe. It's not difficult to do, thanks to the abstractions, I just keep forgetting to do it.

tellowkrinkle:
> Looking at "EmuFolders::Resources = Path::Canonicalize(Path::Combine(EmuFolders::AppRoot, PCSX2_APP_DATADIR "/resources"));", what we're saying here is "take the path to where the binary is, add the path to the data directory, then add "/resources" on the end, which doesn't seem like it'll come up with any sort of a valid path...

This is correct.  `PCSX2_APP_DATADIR` is a relative path from the exe, to support moving the whole root directory around after building (which happens in appimages)

kenshen112:
If hard coding it's an issue. I could use XDG_DATA_HOME for resources

tellowkrinkle:
> Is that more to your guys liking? That still works on my end.

That looks like it's relying on the cwd to find resources, which doesn't look like a good idea.  `Path::Canonicalize(Path::Combine(EmuFolders::AppRoot, PCSX2_APP_DATADIR "/resources"));` is correct, since `PCSX2_APP_DATADIR` is relative to the directory containing the exe, not anything xdg-related.

kenshen112:
And I agree with you

stenzek:
IMO, something like "../share" as the define isn't really much better than the current setup. Ideally, just wait until we merge everything into the binary.

tellowkrinkle:
What *is* the "current setup"?
For wx builds, it's exactly that.  That's why that define is already there.
Qt builds are completely broken right now, which I wouldn't consider "better" than much of anything.

stenzek:
They work fine for me? (both running it inside the build directory, as well as `make install`'ed to `bin`)

Or, at least they did a couple of days ago..

tellowkrinkle:
This only affects `-DPACKAGE_MODE=ON`.  Without that, the define will not be there, and they'll take the existing path.

kenshen112:
^ or even just if it's portable mode it seems okay. Without it's dead.

stenzek:
> This only affects -DPACKAGE_MODE=ON

I stated multiple times above that I didn't want to support "package mode" in Qt. It's silly to support something, only to rip it out when we're planning on embedding all resources soon.

kenshen112:
So, I respect that. In the meantime could we come to this as a temporary fix or compromise until that's happened? Cause I'm totally down for merging resources into the main binary 

Then it allows packages like the AUR to at least function with `pcsx2-qt`

stenzek:
> Then it allows packages like the AUR to at least function with pcsx2-qt

Quoting what I just wrote in the previous post: Furthermore, packagers shouldn't be distributing Qt in its current state anyway. Windows is a different story, since we have the entire thing under our control, and can quickly push bugfixes (with the autoupdater).

We're not even distributing appimages yet, or mac builds, and it's been less than 12 hours since the effectively pre-alpha windows builds are in the release list. Just be patient...

kenshen112:
I understand

qurious-pixel:
$HOME/Depends has to exist, so I wasn't sure when the folder was created. 
And the cache is caching `~/deps`
Usually set an if statement, or install deps if cache hit miss, so it doesn't re-download and rebuild each run, if already configured/downloaded. Saves about 4 mins build time.

qurious-pixel:
No need to move SDL. 
`ld` cannot find libsdl because it is not in search path.

```
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$HOME/Depends 
```

Should be available globally, or in the `appimage.sh` script.

qurious-pixel:
Copy libSDL to `squashfs-root/usr/lib` which is the path to the AppImage bundle. (If not setting LD_LIBRARY_PATH or copying to a known path, like /usr/local/lib)

`/tmp/squashfs-root/usr/{lib,bin}` is the tool that builds the AppImage. Patchelf, goes to the build tool, SDL needs to be found by `ld` to bundle with the AppImage as a dependency of pcsx2.

qurious-pixel:
Have to set the `Qt6_DIR` variable.
Probably `qt/6.2.4/gcc_64/lib/cmake/qt6`
At both the build_depends step, and compile_cmake step.
And later on add qmake to PATH, or symlink.
`qt/6.2.4/gcc_64/bin/qmake/` to package the appimage.

qurious-pixel:
Set 
```
-DCMAKE_PREFIX_PATH=$HOME/Depends/qt/$QT/gcc_64
```
Or
```
export Qt6_DIR="${GITHUB_WORKSPACE}/qt/${QT}/gcc_64/lib/cmake/Qt6"
```

kenshen112:
This sucker's ready for review. The AppImage is a joint effort between, Tellow, Qurious and myself

pcsx2-qt/QtHost.cpp
```diff
@@ -161,12 +161,14 @@ void QtHost::SetAppRoot()
 
 void QtHost::SetResourcesDirectory()
 {
-#ifndef __APPLE__
+#ifdef __APPLE__
+	// On macOS, this is in the bundle resources directory.
+	EmuFolders::Resources = Path::Canonicalize(Path::Combine(EmuFolders::AppRoot, "../Resources"));
+#elif !defined(PCSX2_APP_DATADIR)
```
stenzek, line 167: Why is this in here? It's unrelated to the appimage.

.github/workflows/scripts/linux/generate-cmake.sh
```diff
@@ -2,32 +2,18 @@
...
--DXDG_STD=TRUE                              \
--DUSE_SYSTEM_ZSTD=FALSE                     \
-$ADDITIONAL_CMAKE_ARGS                      \
--GNinja                                     \
+-DCMAKE_PREFIX_PATH="$HOME/Depends"          \
+-DWAYLAND_API=ON                              \
+-DQT_BUILD=ON                                  \
+-DXDG_STD=TRUE                                  \
+-DUSE_VULKAN=ON                                  \
+-DPACKAGE_MODE=ON                                 \
```
stenzek, line 16: AppImage does not need `PACKAGE_MODE`.

.github/workflows/scripts/linux/generate-cmake.sh
```diff
@@ -2,32 +2,18 @@
...
 # Generate CMake into ./build
-cmake                                       \
--DCMAKE_CXX_COMPILER_LAUNCHER=ccache        \
--DCMAKE_BUILD_TYPE=Release                  \
--DPACKAGE_MODE=TRUE                         \
--DWAYLAND_API=TRUE                          \
--DDISABLE_ADVANCE_SIMD=TRUE                 \
--DDISABLE_PCSX2_WRAPPER=TRUE                \
+cmake                                     \
+-DCMAKE_BUILD_TYPE=RelWithDebInfo          \
```
stenzek, line 9: Why are you using `RelWithDebInfo`?

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -49,12 +53,12 @@ cp --dereference /lib/"$LIBARCH"/libgcc_s.so.1 "$GITHUB_WORKSPACE"/squashfs-root
...
 cp .github/workflows/scripts/linux/app-variables.sh "$GITHUB_WORKSPACE"/squashfs-root/apprun-hooks
 export UPD_INFO="gh-releases-zsync|PCSX2|pcsx2|latest|$name.AppImage.zsync"
-/tmp/squashfs-root/AppRun --appdir="$GITHUB_WORKSPACE"/squashfs-root/ --plugin gtk -d "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.desktop -i "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.png
+/tmp/squashfs-root/AppRun --appdir="$GITHUB_WORKSPACE"/squashfs-root/ --plugin qt -d "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.desktop -i "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.png
 # see LD_LIBRARY_PATH in app-variables.sh - the intent is to use system wayland if available but fall back to app-image provided
 # a little bit hacky but should ensure maximum compatibility
-mkdir -p squashfs-root/usr/lib/wayland
-mv squashfs-root/usr/lib/libwayland-* squashfs-root/usr/lib/wayland
-rm squashfs-root/usr/lib/libgmodule-2.0.so.0
+#mkdir -p squashfs-root/usr/lib/wayland
```
stenzek, line 59: Comments?

.github/workflows/scripts/linux/appimage.sh
```diff
@@ -49,12 +53,12 @@ cp --dereference /lib/"$LIBARCH"/libgcc_s.so.1 "$GITHUB_WORKSPACE"/squashfs-root
 chmod +x .github/workflows/scripts/linux/app-variables.sh
 cp .github/workflows/scripts/linux/app-variables.sh "$GITHUB_WORKSPACE"/squashfs-root/apprun-hooks
 export UPD_INFO="gh-releases-zsync|PCSX2|pcsx2|latest|$name.AppImage.zsync"
-/tmp/squashfs-root/AppRun --appdir="$GITHUB_WORKSPACE"/squashfs-root/ --plugin gtk -d "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.desktop -i "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.png
+/tmp/squashfs-root/AppRun --appdir="$GITHUB_WORKSPACE"/squashfs-root/ --plugin qt -d "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.desktop -i "$GITHUB_WORKSPACE"/squashfs-root/PCSX2.png
```
stenzek, line 56: Does the linuxdeploy plugin actually officially support qt6? I can't see any evidence to suggest as much.

That's why I just assembled everything myself (grabbing the qt plugins/patching RPATH).

.github/workflows/linux_build_matrix.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     with:
       jobName: "with LTO"
       compiler: gcc
-      cmakeflags: "-DLTO_PCSX2_CORE=ON"
+      cmakeflags: "" # "-DLTO_PCSX2_CORE=ON"
```
stenzek, line 18: Why are you disabling LTO?

.github/workflows/linux_build.yml
```diff
@@ -86,42 +85,26 @@ jobs:
...
+      
       - name: Generate CMake
         env:
           PLATFORM: ${{ inputs.platform }}
           COMPILER: ${{ inputs.compiler }}
           ADDITIONAL_CMAKE_ARGS: ${{ inputs.cmakeflags }}
         run: .github/workflows/scripts/linux/generate-cmake.sh
 
+      - name: QT 6 paths
+        run: sudo ln -s $(find $HOME/Depends/ -name qmake) /usr/bin
```
stenzek, line 106: This seems like it should be more specific. What if `find` returns more than one path?

.github/workflows/scripts/linux/generate-cmake.sh
```diff
@@ -2,32 +2,18 @@
...
--DOpenGL_GL_PREFERENCE="LEGACY"             \
--DOPENGL_opengl_LIBRARY=""                  \
--DXDG_STD=TRUE                              \
--DUSE_SYSTEM_ZSTD=FALSE                     \
-$ADDITIONAL_CMAKE_ARGS                      \
--GNinja                                     \
+-DCMAKE_PREFIX_PATH="$HOME/Depends"          \
+-DWAYLAND_API=ON                              \
+-DQT_BUILD=ON                                  \
+-DXDG_STD=TRUE                                  \
```
stenzek, line 14: Also unnecessary.

.github/workflows/scripts/linux/build-dependencies.sh
```diff
@@ -0,0 +1,125 @@
...
+
+    for f in qt*.tar*; do 
+        mkdir "${f/-*/}"
+        tar -xf "$f" -C "${f/-*/}" --strip-components=1; 
+    done
+
+
+	echo "Installing Qt Base..."
+	cd "qtbase"
+	cmake -B build -DCMAKE_PREFIX_PATH="$INSTALLDIR" -GNinja -DCMAKE_INSTALL_PREFIX="$INSTALLDIR" -DCMAKE_BUILD_TYPE=Release -DFEATURE_optimize_size=ON -DFEATURE_framework=OFF -DFEATURE_icu=OFF -DFEATURE_printsupport=OFF -DFEATURE_sql=OFF
```
stenzek, line 78: Why `-DFEATURE_framework=OFF`?
Why -`DFEATURE_icu=OFF`? (will this break unicode conversion?)
Why `-DFEATURE_optimize_size=ON`?

.github/workflows/scripts/linux/install-packages.sh
```diff
@@ -35,9 +37,9 @@ declare -a PCSX2_PACKAGES=(
 )
 
 if [ "${COMPILER}" = "gcc" ]; then
-  BUILD_PACKAGES+=("g++-10-multilib")
+  BUILD_PACKAGES+=("g++-10")
```
stenzek, line 40: Probably better off just using clang and dropping gcc.

stenzek (changes requested):
Have you actually tested that this works? Several things I see might prevent it from running.

Other issues outside of the files listed:

 - It's still doing the LD_LIBRARY_PATH dance for libwayland, this is unnecessary, Qt doesn't static link against libwayland-client, and including it in the appimage will break things (as the library has a circular dependency on wayland-egl, which comes from the driver).
 - The qt plugin for linuxdeploy is questionable. What libraries does it pull in? Does it grab all the necessary Qt plugins? What about the wayland issue from above? If it's naively ldd'ing the, it'll end up pulling in system versions.
 - AppImage does *not* need package mode.
 - You've effectively removed the wx appimage. We're not ready to drop it just yet.

I'm sure there's others but that's what I saw at a quick glance.

.github/workflows/scripts/linux/build-dependencies.sh
tellowkrinkle: These are pulled from the Mac build script, so I can tell you my reasoning for that
- `-DFEATURE_framework` is Mac specific, but I prefer dylibs there for bundling
- ICU is larger than the entire rest of the application combined, so I'd like to avoid it if possible.  I don't think we currently use it, but I'd expect it would be useful for things like splitting strings on unicode graphemes, locale-correct case conversions, etc.  I assume Windows currently builds with this off as well, since it doesn't ship with a 40mb ICU dll.  I doubt it'll break UTF-16 ↔︎ UTF-8 conversion, but it could break more obscure codecs if we used any (I hope we don't).
- I enabled optimize size because our dependencies tend to stay out of our hot paths, so I didn't think we needed `-O3`.  Saves a lot on SDL, though I don't know how much it saves on Qt.

.github/workflows/linux_build_matrix.yml
kenshen112: We were having some real bad linker issues with it earlier.

stenzek: It works fine for me (with clang at least).

.github/workflows/scripts/linux/appimage.sh
qurious-pixel: left it in as out-of-scope for qt, but still part of wX

.github/workflows/scripts/linux/appimage.sh
qurious-pixel: Qt6 is supported, yes. and confirmed to work.

.github/workflows/linux_build_matrix.yml
qurious-pixel: this is building with gcc, not clang. linker causing breakage with the qt translation.

.github/workflows/linux_build.yml
qurious-pixel: only one instance of qt is installed. only one instance of qmake at $HOME/Depends. Can use absolute path.

.github/workflows/scripts/linux/appimage.sh
refractionpcsx2: This app image is for Qt, not WX

stenzek:
Superseded by #6555.

PR #6135, opened 2022-05-23
input-rec: fix crash in wxWidgets if input recording is disabled

xTVaser:
### Description of Changes

Fixes the aforementioned crash, virtual pads are only initialized if input recording is enabled.  In the PR that merged earlier today, it exposed a potential path where a NULL virtual pad could be accessed.

I've added back the safeguard to only intercept the inputs if recording is enabled (only a wx concern) as well as eliminated the null pointer issue at the source.

### Rationale behind Changes

Crashes are bad

### Suggested Testing Steps

Restart the emulator on wx with input recording enabled and disabled.  In Qt, input recording is always enabled, and there is no ability to disable it -- so this should not be an issue.


dio-gh:
Working as intended, fix confirmed.

PR #6137, opened 2022-05-23
ci: differentiate cmake windows build from the other build configs

xTVaser:
### Description of Changes

Append `-cmake` to artifacts that are built with cmake (only currently applicable to windows)

### Rationale behind Changes

Learned something new about actions artifacts.  If they have the same name the file contents are effectively merged.  Looking at the recent artifacts, wxWidgets AVX2 Windows artifacts have 2 .exe's.  The `pcsx2.exe` is coming from the cmake build.

Differentiating the name solves this inconsistency.

All other windows builds were fine because we only built cmake with avx2 (and only on wx).  Linux and MacOS are _always_ done with cmake so its redundant to append there.

### Suggested Testing Steps

Look at the artifact in question.

### DIFF
--- a/.github/workflows/scripts/common/name-artifacts.sh
+++ b/.github/workflows/scripts/common/name-artifacts.sh
@@ -8,6 +8,7 @@
 
 # Inputs as env-vars
 # OS
+# BUILD_SYSTEM
 # GUI_FRAMEWORK
 # ARCH
 # SIMD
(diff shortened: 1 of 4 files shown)


dio-gh:
On one hand, no more sussy .exe files in the Wx AVX2 archive. On the other, there is now a CMake archive.

I guess this is a success? :D

xTVaser:
heh, well there used to be a cmake artifact -- just something else that went unnoticed, especially since we never used the cmake results for the releases (only for PR testing).  See - https://github.com/PCSX2/pcsx2/actions/runs/2365140226

dio-gh:
Ah alright then, LGTM. :)

Issue #6138, opened 2022-05-23
[BUG]: (Qt) Memory card folder format not detected when a game boots. 


BParks21:
### Describe the Bug

When booting a game using the memory card folder format it seems as though the memory card is being loaded too late triggering a no memory card detected message in games. However the card is actually loaded, and save files can still be accessed. 

### Reproduction Steps

Use the folder memory card format and boot a game like Kingdom Hearts 2 or Star Ocean. The message should be triggered as soon as the actual game boots up. 

### Expected Behavior

Memory cards should be detected the moment the game is started and no message should be triggered. wxWidgets builds behave as expected. 

### PCSX2 Revision

v1.7.2775

### Operating System

Windows 10 (64bit)

### CPU

i7 8086k

### GPU

Nvidia GTX 1080ti

### GS Settings

Default

### Emulation Settings

Default
![gs_20220523002706_Kingdom Hearts II - Final Mix __SLPM-66675](https://user-images.githubusercontent.com/12180630/169743900-1cc60095-4aed-4b8f-8f7d-ea216d455963.png)
![gs_20220522220434_Star Ocean 3 - Till the End of Time _Disc2of2__SLUS-20891](https://user-images.githubusercontent.com/12180630/169743901-c9a68b30-2d7e-4e69-b927-d7d2253c783a.png)

PR #6143, opened 2022-05-23
VMManager: Clear memory card eject timeout when game starts

stenzek:
If we don't reset the timer here, when using folder memcards the reindex will cause an eject, which a bunch of games don't
like since they access the memory card on boot.

This PR makes it behave the same as wx.

Fixes https://github.com/PCSX2/pcsx2/issues/6138


JordanTheToaster (approved):

CookiePLMonster:
Test Drive Unlimited confirmed fixed.

image

PR #6154, opened 2022-05-23
Qt: Add no-interlacing patches

stenzek:
### Description of Changes

Had this sitting on the branch for a while and forgot to PR.

Extend the built-in patch set system from just widescreen patches to no-interlacing patches too.

### Rationale behind Changes

Making games look less rubbish with little work from users.

### Suggested Testing Steps

Test the new feature in the games it affects. I've included a zip based on the one from the forums.


RedDevilus:
Think I'm still missing No-Interlacing patches in the latest zip:
https://u.pcloud.link/publink/show?code=kZDH81kZHH9txOzWRSfidhxYhMviSYe30lW7&lang=en#folder=10433533981&tpl=publicfoldergrid

Either way, probably don't need to actively maintain no-interlacing pnachs separately then.

Anuskuss:
MFing stenzek back at it again. Literally everything has been taken care of now and my custom launcher is obsolete. I guess having to manually set widescreen in the game settings is the last thing left for the user to do but I assume detecting that is just not possible.
Anyway, would it be feasible to turn those ZIPs into single YAMLs (`patches_ni.yaml` and `patches_ws.yaml`)? It's pain to update them and git doesn't really like binary files. Also the compression gains are negligible (actually leaving them uncompressed and serving 7zips should be better).

stenzek:
Yeah, I agree, moving to a different format would be better for version control. Just kept the consistency for now, but we can move it all over at once in the future.

JordanTheToaster (approved):

PR #6157, opened 2022-05-23
GameDB: Revise serial names + GSHWFixes

RedDevilus:
### Description of Changes

With Qt being released there were still illegal characters being used in the name (which stops you from using covers and more), also found some random typo mistakes.

Fixes https://github.com/PCSX2/pcsx2/issues/6170

GsHWFixes:
- Ratatouille
- Wild Arms series
- Tony Hawk Downhill Jam

Ratatouille;
Before:
image

After:
image

So it's not perfect, but you can use roundsprite to full but will give vertical line effect or you could use texture offsets.

Wild Arms Series;

Upscaled (no changes):
image
image

Half Roundsprite (used currently which wasn't perfect):
image
image

Wild Arms (new one for whole series):
image
image

Tony Hawk Downhill Jam;

Before:
image

After:
image

### Rationale behind Changes

Less tinkering.
### Suggested Testing Steps

Test ratatouille.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -49,7 +49,7 @@ ALCH-00010:
   name: "Kono Aozora ni Yakusoku wo - Melody of the Sun and Sea [Limited Edition]"
   region: "NTSC-J"
 ALCH-00011:
-  name: "Pure x Cure RE:Covery [Koi no Kyuukyuu Set]"
+  name: "Pure x Cure RE-Covery [Koi no Kyuukyuu Set]"
   region: "NTSC-J"
 ALCH-00014:
   name: "Aria - The Origination ~Aoi Hoshi no El Cielo~ [Limited Edition]"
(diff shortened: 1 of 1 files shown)


coornio:
The perfect fix is to use Skipdraw 3 and HPO: special texture aggressive. The former removes one layer of the weird ghosting (seemingly without any actual graphical loss compared to software) while retaining vibrancy, and the latter helps aligning the remaining ghosting exactly to where it needs to be. 

gs_20220524004900

Honestly, this skipdraw might be worth it, it's practically harmless.

RedDevilus:
Skipdraw might be for the second time be a good fix

coornio:
Small update, any skipdraw value between 1 and 3 does away with the pesky ghost, whether it's 1,1 or 1,3 or anything inbetween. Not sure why. Just mentioning this in case it's important. It *might* have some effect elsewhere in the game, like some menu, so I'd recommend a bit more testing on it to be sure.

RedDevilus:
No skipdraw:
image

Skipdraw 2:
image

The ghosting in hardware is because it's used as a blooming filter and looks off upscaled:
image

RedDevilus:
```
    skipDrawStart: 2 # Removes blooming filter that makes it look like ghosting.
    skipDrawEnd: 2 # Removes blooming filter that makes it look like ghosting.
```
I'll put it in here to remember there is an issue upscaled (also native with chandeliers not working properly, actually general lighting)

RedDevilus:
image
Should be fine now

PR #6163, opened 2022-05-24
GS: Add Skip Presenting Duplicate Frames option

### LINKED ISSUE #6103: [BUG]: Game is more performant when the frame limiter is off
### Describe the Bug

I have bouts of dips to 80% speed in NFS Underground 2 sometimes, but I noticed that if I uncap the speed limiter, it will shoot up closer to 100%, and sometimes even significantly beyond.

image

In this scene specifically, I get roughly 100% speed on Vulkan in windowed mode, but it drops to about 80% in full screen. When uncapped however, the speed shoots up all the way to 140%.

I can replicate this on all backends, even software, though in this specific scene I couldn't get DX11 to display this, as DX11 is too performant. I'm on the high performance energy profile, and have cross referenced HWinfo to verify that my CPU isn't downclocking or thermal/energy throttling.

**Edit:** Oh yeah, this is on an Intel i5 4440. OpenGL and Vulkan seem to have a tendency to trigger some AMD driver hack that enables exclusive full screen, but even on DX11 SW mode I can repro this issue, so it doesn't seem to be that.

### Reproduction Steps

Pull up NFSU 2 and give it a whirl as described.

### Expected Behavior

Consistent emulation speed that sticks to 100% if the hardware is capable of it (which it clearly is, as can be seen when uncapped).

### PCSX2 Revision

v1.7.2746

### Operating System

Windows 8.1 (64bit)

### If Linux - Specify Distro

_No response_

stenzek:
### Description of Changes

I found this didn't really help much on my desktop system, but dio found that it helped on his system when using vsync+exclusive fullscreen.

Detects when idle frames are being presented in 25/30fps games, and skips presenting those frames. The frame is still rendered, it just means the GPU has more time to complete it (this is NOT frame skipping). Can smooth our frame time fluctuations when the CPU/GPU are near maximum utilization, but makes frame pacing more inconsistent and can increase input lag.

### Rationale behind Changes

Fixes #6103 .

### Suggested Testing Steps

Test new option.

### DIFF
--- a/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
@@ -206,11 +206,13 @@ GraphicsSettingsWidget::GraphicsSettingsWidget(SettingsDialog* dialog, QWidget* 
 	//////////////////////////////////////////////////////////////////////////
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.useBlitSwapChain, "EmuCore/GS", "UseBlitSwapChain", false);
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.useDebugDevice, "EmuCore/GS", "UseDebugDevice", false);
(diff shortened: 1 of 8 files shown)


dio-gh:
Code matches the git patch I got from before, so while I can't test it directly, should be okay.

JordanTheToaster (approved):

PR #6177, opened 2022-05-24
Qt: Revision default hotkeys v2

RedDevilus:
### Description of Changes

There has been a re-order of keybindings, new keybindings and missing Single Frame GS Dump, Multi Frame GS Dump and Input Recording mode.

To be decided for the rest if it needs default hotkeys and which key or key combinations would be best.

image
image
image
image
image

### Rationale behind Changes

Consistency with older hotkeys, better code readability which mimics GUI structure.
### Suggested Testing Steps

Test hotkeys.

### DIFF
--- a/pcsx2/PAD/Host/PAD.cpp
+++ b/pcsx2/PAD/Host/PAD.cpp
@@ -227,19 +227,27 @@ void PAD::SetDefaultConfig(SettingsInterface& si)
 	// PCSX2 Controller Settings - Hotkeys
 
 	// PCSX2 Controller Settings - Hotkeys - General
-	si.SetStringValue("Hotkeys", "Screenshot", "Keyboard/F8");
 	si.SetStringValue("Hotkeys", "ToggleFullscreen", "Keyboard/Alt & Keyboard/Return");
 
 	// PCSX2 Controller Settings - Hotkeys - Graphics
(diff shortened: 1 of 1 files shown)


RedDevilus:
F8:
image

Shift + F8: 
image

Control + Shift + F8:
image

Spamming different combination seems to work fine:
image

coornio:
Perhaps it might be a good idea to consider a combination of keys friendlier to one-handed use for the GS dumps? From the left ctrl/shift the F8 key is too far, and using the right ctrl/shift with your left hand is awkward. You can surely use both hands at once but in some situations you might be required to do an action ingame while the recording is happening.

With this in mind, perhaps we should think about how to make one-handed use of multi-frame dumps easier, or even the possibility of changing it to a start/stop record function altogether so you won't have to hold the buttons down for the duration (with failsafe duration in case someone forgets and automatic cancel if they save/load a savestate).

stenzek:
There's always the gs dump item in the debug menu.

RedDevilus:
> There's always the gs dump item in the debug menu.

Yeah, that's what people use now, but it's nice to have the same hotkeys on Qt.

Rex000:
A toggle for muting audio would be nice in the "General" tab.

PR #6206, opened 2022-05-26
GameDB: Tales of Legendia + Batman Begin + ...

RedDevilus:
### Description of Changes

- Added FMV switch for Tales of Legendia which can show garbage textures
- Added Batman Begins (Autoflush and Special Texture half pixel offset for the lighting misalignment)
- Some extra clean-up on Spartan - Total Warrior

Tales of Legendia;

Before:
TalesOfLegendiaBefore

After:
TalesOfLegendiaAfter

Batman Begins;

Before:
BatmanBefore

After:
BatmanAfter

### Rationale behind Changes

Less tinkering.
### Suggested Testing Steps

Test relevant games.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -983,6 +983,8 @@ SCAJ-20144:
 SCAJ-20145:
   name: "Tales of Legendia"
   region: "NTSC-J"
+  gameFixes:
+    - SoftwareRendererFMVHack # Fixes garbage textures presumably from texture cache issue.
 SCAJ-20146:
   name: "Shadow of the Colossus"
   region: "NTSC-Ch-E-J"
@@ -4476,6 +4478,8 @@ SCKA-20050:
   name: "Tales of Legendia"
   region: "NTSC-K"
   compat: 5
+  gameFixes:
+    - SoftwareRendererFMVHack # Fixes garbage textures presumably from texture cache issue.
 SCKA-20051:
   name: "Minna Daisuki Katamari Damacy"
   region: "NTSC-K"
@@ -8332,7 +8336,7 @@ SLED-53732:
   name: "Spartan - Total Warrior [Demo]"
   region: "PAL"
   gameFixes:
-    - EETimingHack # Fixes garbage textures flashing on the character model
+    - EETimingHack # Fixes garbage textures flashing on the character model.
 SLED-53745:
   name: "Total Overdose [Demo]"
   region: "PAL-M5"
@@ -15369,6 +15373,9 @@ SLES-53386:
 SLES-53387:
   name: "Batman Begins"
   region: "PAL-M7"
+  gsHWFixes:
+    autoFlush: 1 # Fixes lighting misalignment.
+    halfPixelOffset: 2 # Fixes lighting misalignment.
 SLES-53388:
   name: "Rhythmic Star!"
   region: "PAL-M5"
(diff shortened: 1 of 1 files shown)


Rex000:
I can't quite tell the difference in Batman Begins. I have enabled Autoflush and Special Texture half pixel offset already.

refractionpcsx2:
You shouldn't need to manually enable it.

but if you compare the screenshots it fixes the window light alignment.

Rex000:
Oh I see.

RedDevilus:
> I can't quite tell the difference in Batman Begins. I have enabled Autoflush and Special Texture half pixel offset already.

Yeah, it's to avoid doing what you did there exactly. Just like we do automatic rounding/clamping and other options.

Rex000:
> > I can't quite tell the difference in Batman Begins. I have enabled Autoflush and Special Texture half pixel offset already.
> 
> Yeah, it's to avoid doing what you did there exactly. Just like we do automatic rounding/clamping and other options.

Hmm I understand now. I didn't know these were automatic fixes and didn't require enabling manual hardware renderer fixes.

refractionpcsx2:
yes the idea of all these changes is so users don't need to configure anything, as one of the biggest complaints we get is the fact you have to have different configurations for different games when upscaling.

RedDevilus:
WX:
image

Qt:
image

Rex000:
> yes the idea of all these changes is so users don't need to configure anything, as one of the biggest complaints we get is the fact you have to have different configurations for different games when upscaling.

Oh indeed it's the biggest complaint. It's great to have them automatically.

Something just came to me. If for some reason the user needs to try more fixes, then they would surely need to enable manual hardware renderer fixes. But then how would they know what automatic settings got disabled in the process?

So in order to avoid this, I think it may be a good idea to enable those settings in  Hardware-settings/Upscaling-fixes by default? This way when the user enables manual fixes, they would immediately know what automatic fixes were enabled before. Then they can turn them off or change them along with any other manual changes. 

I hope I have explained this well.

RedDevilus:
> > yes the idea of all these changes is so users don't need to configure anything, as one of the biggest complaints we get is the fact you have to have different configurations for different games when upscaling.
> 
> Oh indeed it's the biggest complaint. It's great to have them automatically.
> 
> Something just came to me. If for some reason the user needs to try more fixes, then they would surely need to enable manual hardware renderer fixes. But then how would they know what automatic settings got disabled in the process?
> 
> So in order to avoid this, I think it may be a good idea to enable those settings in Hardware-settings/Upscaling-fixes by default? This way when the user enables manual fixes, they would immediately know what automatic fixes were enabled before. Then they can turn them off or change them along with any other manual changes.
> 
> I hope I have explained this well.

You underestimate the users' talent to not read something, there is a reason why it's hidden behind a checkbox even with text not to. But perhaps in the future we could do something like reflect what fixes are enabled by default (as reference without disabling anything) with a separate option, please do that and then add texture offsets separately without losing automatic settings.

But that's another thing on the endless list of things to do.

Rex000:
> You underestimate the users' talent to not read something, there is a reason why it's hidden behind a checkbox even with text not to. But perhaps in the future we could do something like reflect what fixes are enabled by default (as reference without disabling anything) with a separate option, please do that and then add texture offsets separately without losing automatic settings.
> 
> But that's another thing on the endless list of things to do.
I guess I failed to explain it properly. I'm not saying the users can't or won't read. What I'm saying is that there should be some sort of indication of what automatic fixes a game already has enabled. Even if it's just text in the game's properties.

For instance in this case, if somebody hasn't read this page, they won't ever know that Batman Begins already has Autoflush and Special Texture half pixel offset automatically enabled. Therefore if they decide to play around with manual hardware fixes, they'll obviously disable those automatic changes for themselves without knowing what those automatic settings actually were.

RedDevilus:
I actually listened to that as it's not as esy to see grayed out section for now (the reference part)

Rex000:
> I actually listened to that as it's not as easy to see grayed out section for now (the reference part)

Yeah, but maybe we can have another tab in game properties by the name of "Important Notes" that would list the automatic changes among other things? Something of that sort.

RedDevilus:
It already says that in the emulog, people shouldn't adjust normally. But I get what you mean for tinkerers.

PR #6213, opened 2022-05-27
Qt: Add hardware check for SSE4 and AVX2

stenzek:
### Description of Changes

Make sure the user has the required CPU instructions before PCSX2 properly initializes.

### Rationale behind Changes

Stopping people running the wrong build then complaining it crashes.

### Suggested Testing Steps

Difficult to test unless you have an ancient CPU. But the code can be reviewed :)

### DIFF
--- a/common/emitter/cpudetect.cpp
+++ b/common/emitter/cpudetect.cpp
@@ -46,6 +46,11 @@ using namespace x86Emitter;
 
 alignas(16) x86capabilities x86caps;
 
+#ifdef _MSC_VER
+// We disable optimizations for this function, because we need x86capabilities for AVX
+// detection, but if we keep opts on, it'll use AVX instructions for inlining memzero.
+#pragma optimize("", off)
+#endif
 x86capabilities::x86capabilities()
 	: isIdentified(false)
 	, VendorID(x86Vendor_Unknown)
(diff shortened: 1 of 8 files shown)


Ryudo300:
Gave this a test drive on an Intel Core i3-350M which has no AVX2 support. (only has up to SSE4)

The program never opens up, and no message shows up either.

stenzek:
> Gave this a test drive on an Intel Core i3-350M which has no AVX2 support. (only has up to SSE4)

Should be fine now. Had to do a massive hack to get it to run early enough, and no fancy message box with clickable URLs :(

Ryudo300:
Unfortunately still no message showing up on commit 1d5ae55. 😟 Nado reported the same from his side

stenzek:
Hopefully *now* we should be good. Disgusting hacks :(

shaggi72:
works for me

Core i7 3770
Sin título

Ryudo300:
> works for me
> 
> Core i7 3770 Sin título

Your CPU supports AVX, and that commit you're using supposedly already worked with it. (This is proof of that fact). Right now we still need to test CPUs that only support up to SSE4.

Xanado95:
pr- 6213(6b5c4be) Avx2 QT now produces a pop up as intended when trying  run it on an unsupported cpu even on cpu's without avx instructios.

NadoPS:
pr- 6213(6b5c4be) Avx2 QT now produces a pop up as intended when trying  run it on an unsupported cpu even on cpu's without avx instructions.

PR #6214, opened 2022-05-27
GUI/Qt: Prioritize serial in cover lookup

### LINKED ISSUE #6212: [BUG]: Covers separation per Region or Serial Number
### Describe the Bug

I think covers must be separated for every variation of same game using serial number or region.

 

### Reproduction Steps

I have all variation of Bloody Roar 3. If I change cover for one of them, they all will change too.

### Expected Behavior

I change cover only for one variation of game, not for all of them

### PCSX2 Revision

v.1.7.2813

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

refractionpcsx2:
### Description of Changes
Change the cover lookup priority to check serial first.

### Rationale behind Changes
Generic cover names were used first, which is probably a last reset "generic" filename, serial is much more specific and should be loaded if it exists.

### Suggested Testing Steps
If you have 2 versions of the same game (different regions), try one with the generic game name, and another with the serial.

Fixes #6212

### DIFF
--- a/pcsx2/Frontend/GameList.cpp
+++ b/pcsx2/Frontend/GameList.cpp
@@ -683,7 +683,17 @@ std::string GameList::GetCoverImagePath(const std::string& path, const std::stri
 	std::string cover_path;
 	for (const char* extension : extensions)
 	{
-		// use the file title if it differs (e.g. modded games)
+
+		// Prioritize lookup by serial (Most specific)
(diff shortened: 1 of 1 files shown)


metalslayer:
Renaming files to serial numbers fixed my issue. But still...

PR #6232, opened 2022-05-28
Build: Don't rewrite svnrev.h unconditionally

stenzek:
### Description of Changes

Stops VS from recompiling the files which include it every time you hit run.

### Rationale behind Changes

It's really annoying.

### Suggested Testing Steps

Make sure it still builds and gets the correct version number.


refractionpcsx2:
Can't tell from a PR so I guess we'll just YOLO it and see what happens :D

PR #6234, opened 2022-05-28
GameDB: fix up game names

Mrlinkwii:
fixes game names in the GameDB

### Suggested Testing Steps

check said game


coornio:
Can you include an edit for yugioh capsule monster coliseum? One of the entries has a dash but it shouldn't.

coornio:
Many thanks <3

PR #6236, opened 2022-05-28
Updater: Making it fancier

CookiePLMonster:
**CAUTION: Please review and test this thoroughly, it would suck to break the updater by accident.**

### Description of Changes
This PR gives the Windows updater some love:
* UI has been moved to a thread, so it is properly responsive and long running operations don't make Windows think the window is stuck.
* Taskbar progress bar has been added, indicating update progress on the taskbar icon, much like many modern installers/updaters do.
* Initial indeterminate stages of updating (waiting for `pcsx2.exe` to exit, parsing the ZIP file etc.) now set the progress bar and the taskbar progress bar into a marquee/indeterminate state. This means that a typical _"something is going on, but I can't give you an ETA"_ sliding progress bar is now used in place of a stuck empty bar.
* Maximize and Close buttons have been disabled because they were breaking things and the updater was never adapted for layout changes/aborting.

Indeterminate progress now looks like this:
image

While PCSX2 could benefit from a taskbar progress bar, it's been removed from Qt6 so it'll likely need a pure WinAPI solution later, much like the one here.

### Rationale behind Changes
1. Because I can.
2. Because with those improvements the updater feels a little more native.

### Suggested Testing Steps
* Nudge the updater a little, verify that it looks and works properly.
* Ensure there are no functional regressions and updating works as expected.

To test the Updater manually:
1. Make a copy of any pcsx2 ZIP from the build bot. The updater will wipe it after a successful update, so make sure you back it up to save yourself some hassle when retesting.
2. Create a temporary directory somewhere, e.g. `pcsx2_tmp` or anything.
3. Launch `updater.exe` manually (e.g. from a batch file or Command Prompt) like: `updater.exe 1 "path\to\pcsx2_tmp" "path\to\pcsx2_update.zip" "path\to\pcsx2_tmp\pcsx2-qtx64-avx2.exe"`.
4. If done correctly, updater will unpack that "fake" update to a temporary directory created earlier and then run this unpacked version.

### DIFF
--- a/common/ProgressCallback.cpp
+++ b/common/ProgressCallback.cpp
@@ -128,6 +128,7 @@ public:
 	void SetProgressRange(u32 range) override {}
 	void SetProgressValue(u32 value) override {}
 	void IncrementProgressValue() override {}
+	void SetProgressState(ProgressState state) override {}
 
 	void DisplayError(const char* message) override { Console.Error("%s", message); }
 	void DisplayWarning(const char* message) override { Console.Warning("%s", message); }
@@ -147,12 +148,6 @@ static NullProgressCallbacks s_nullProgressCallbacks;
 ProgressCallback* ProgressCallback::NullProgressCallback = &s_nullProgressCallbacks;
 
 BaseProgressCallback::BaseProgressCallback()
-	: m_cancellable(false)
-	, m_cancelled(false)
-	, m_progress_range(1)
-	, m_progress_value(0)
-	, m_base_progress_value(0)
-	, m_saved_state(NULL)
 {
 }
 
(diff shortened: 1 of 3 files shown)


coornio:
Wait, does this mean that we'll lose that tiny little window that appeared at the end that says "Update completed successfully" with an OK button we must press, but it occasionally fails to render on top of other windows so I have to dig it out by minimizing everything else?

Awesome!

CookiePLMonster:
No, that's not changed.

coornio:
> No, that's not changed.

shit

Daisouji:
It's taking forever on waiting to parent process # process

PR #6247, opened 2022-05-28
DEV9: Exclude pcap NPF_ Prefix from GUID on Windows

TheLastRar:
### Description of Changes
On windows, PCAP adds a prefix onto the adapter GUID, which we save to settings file
This pr instead strips and readds that prefix to the GUID instead of saving it.
This does mean you will need to reselect the adapter in the network options

### Rationale behind Changes
Makes it slightly easier to consolidate code
Also Fixes the QT ui not being able to match the same adapter between pcap and sockets, when switching between the those API types

### Suggested Testing Steps
Test if pcap initialises correctly with a selected adapter on windows

### DIFF
--- a/pcsx2/DEV9/pcap_io.cpp
+++ b/pcsx2/DEV9/pcap_io.cpp
@@ -42,6 +42,10 @@
 #define PCAP_NETMASK_UNKNOWN 0xffffffff
 #endif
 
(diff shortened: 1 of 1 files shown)


lightningterror:
Suppressed by #6248

PR #6248, opened 2022-05-28
DEV9: Deduplicate network adapter code

TheLastRar:
### Description of Changes
Unifies a bunch of code across the DEV9 networking code
includes https://github.com/PCSX2/pcsx2/pull/6247

### Rationale behind Changes
Less code duplication

### Suggested Testing Steps
Test if internal DHCP still works
Test if pcap still works
Test if sockets still works

### DIFF
--- a/pcsx2/CMakeLists.txt
+++ b/pcsx2/CMakeLists.txt
@@ -343,6 +343,7 @@ set(pcsx2SPU2Headers
 
 # DEV9 sources
 set(pcsx2DEV9Sources
+	DEV9/AdapterUtils.cpp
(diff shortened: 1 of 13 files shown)


JaduZ:
Hello, could you see the problem why Pro Evolution 2014, whether in the normal version, has errors when playing, slow down, and in the Qt version, it gives a distortion when playing, is there a solution so that it does not happen?

TheLastRar:
> Hello, could you see the problem why Pro Evolution 2014, whether in the normal version, has errors when playing, slow down, and in the Qt version, it gives a distortion when playing, is there a solution so that it does not happen?

Is that a recent problem, has this been an issue for a while?

PR #6256, opened 2022-05-29
GS: Fix CSR.FIELD for progressive mode

### LINKED ISSUE #1413: Tekken 5 NTSC-U : Progressive Scan Screen Shaking
**Game** : Tekken 5 (NTSC-U Version)
**Activate Progressive Scan :** Go to Options menu and select it
**Mode :** Software and Hardware have the issue
**Interlacing :** Auto

The screen will shake during the game after the option activated, can be mostly seen in the pause menu.

This is not an issue with my TV, I can play Burnout 2 or Dragon Ball Z Budokai 3 in progressive scan without screen shaking. Also, some people on the forums mentionned the same bug too.

refractionpcsx2:
### Description of Changes
Fixes the CSR.FIELD bit to set when in 480 progressive mode (and others)

### Rationale behind Changes
Not fixing this causes games to shake in progressive mode. @F0bes tested this on the real hardware and it is fixed set when in 480p mode.

### Suggested Testing Steps
Test games in progressive, make sure they don't shake.

Fixes #1413

### DIFF
--- a/pcsx2/GS.h
+++ b/pcsx2/GS.h
@@ -28,10 +28,10 @@ alignas(16) extern u8 g_RealGSMem[Ps2MemSize::GSregs];
 
 enum CSR_FifoState
 {
-    CSR_FIFO_NORMAL = 0,	// Somwhere in between (Neither empty or almost full).
-    CSR_FIFO_EMPTY,			// Empty
-    CSR_FIFO_FULL,			// Almost Full
-    CSR_FIFO_RESERVED		// Reserved / Unused.
+	CSR_FIFO_NORMAL = 0, // Somwhere in between (Neither empty or almost full).
+	CSR_FIFO_EMPTY, // Empty
+	CSR_FIFO_FULL, // Almost Full
+	CSR_FIFO_RESERVED // Reserved / Unused.
 };
 
 // --------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


F0bes:
Related test:
https://github.com/PCSX2/HardwareTests/tree/main/progressive_field

Immersion95:
This totally solves https://github.com/PCSX2/pcsx2/issues/1413

pcsx2/GS.h
```diff
@@ -145,6 +145,11 @@ union tGS_CSR
 		_u32 ^= 0x2000;
 	}
 
+	void SetField()
```
lightningterror, line 148: Formatting.

refractionpcsx2:
> This totally solves #1413

yep, that's what I was targeting :D Thanks for checking (and bringing the Issue to my attention, I forgot about it)

felixthecat1970:
i was testing some custom codes for enable sdtv modes in games (for no interleacing and less blur image) past month and see this PR the games with video playing start to crash, tracing where games crash always  after sceGsSynV function, withow know much about GS internal "i guess" this because internal PCRT circuit still thinks is in "interleaced signal" so maybe you have a advice what kind of functions start looking for or advice how to workaround this video crash, thanks.

refractionpcsx2:
We now stop the FIELD from swapping if CMOD in SMODE1 is set to 0 or the lowest bit of SYNCV is set (It's weird, I know, but this is how the PS2 reacts), and this makes actual progressive games work.

if games are crashing waiting for the field to swap, then those patches need to be changed/updated to not check it.

Issue #2132, opened 2017-11-16
Meta: Games with IPU issues


MrCK1:
I figured a meta issue to track games with various IPU issues would be helpful since they're not always well documented/found easily. FMV issues related to GSdx are not included here.

Feel to make corrections as necesssary. The games here are based off my own tests, and data from the wiki, so re-tests would always be nice.

- [ ] Test Drive (in-game freezing) - https://github.com/PCSX2/pcsx2/issues/1863

- [ ] Call of Duty World at War: Final Fronts (corruption)

- [ ] Harry Potter Quidditch World Cup (corruption)

- [ ] James Bond 007: Everything or Nothing

- [ ] Black & Bruised (hangs without SkipMpeg)

- [ ] The Guy Game

- [ ] The Incredible Hulk (hangs without SkipMpeg)

- [ ] Test Drive Off-Road: Wide Open (blank videos, might be GSdx issue instead?)

- [ ] Tiger Woods PGA Tour 06 (hangs without SkipMpeg)

- [ ] Q-Ball Billards Master (unsure if this one has issues still, may be fixed)

- [ ] Classic British Motor Racing (hangs without SkipMpeg)

- [ ] Knight Rider: The Game (hangs without SkipMpeg)

Unconfirmed:

- [ ] Spyro: A Hero's Tail - https://github.com/PCSX2/pcsx2/issues/858

- [ ] Guitar Hero - https://github.com/PCSX2/pcsx2/issues/2108 (freezes at end of videos)

- [ ] Valkyrie Profile 2: Silmera - https://github.com/PCSX2/pcsx2/issues/1634 (freezes at end of videos)

atomic83GitHub:
You can add Onimusha - Dawn of Dreams, it have some corruption in opening/title screen and have a patch to avoid some IPU issues.

atomic83GitHub:
I have update my GameDB PR with games that need skipmpeg to work. Also you can add "lethal skies 2" and "Paris dakar rally".

atomic83GitHub:
Only Paris-Dakar and lethal skies 2 booth pal.

MrCK1:
Oh you're right about the Hulk. I'll remove that then.

atomic83GitHub:
Hum, you are right with lethal skies 2, I have test it on the 1.4 version but it work just fine in 1.5dev, unfortunatly, paris dakar is still broken.

atomic83GitHub:
[...]

PR #6270, opened 2022-05-29
IPU: Implement start code validation for BDEC/IDEC

### LINKED ISSUE #1141: Graphical Corruption on opening screens and title screen in Onimusha - Dawn of Dreams
When started, the game will first ask for a system save on a mem card, after that the opening few screens and the title screen will have major graphical corruption. Opening video is not affected. Log is spammed with "libdma: sync timeout".

Version - pcsx2-v1.5.0-dev-165-g6dd83d3-windows-x86
Video - https://youtu.be/BBIs9BKQbew
Log - https://gist.github.com/deadair3210/b954d2ce710ffd9ff079

Goatman13:
### Description of Changes
According to MPEG 1/2 specs when 8 "0" bits are found, we should check for 000001xxh. Recommended way is to discard bits until stream is aligned. Then check 24 bits for 000001h, and if they won't match discard 8 bits and try again until valid code is found, or stream ends.

### Rationale behind Changes
Fixes #1141
Also helps Tiger Woods 06/08 from https://github.com/PCSX2/pcsx2/issues/2132
WRC3 patch removed as a bonus, game work also on master.

### Suggested Testing Steps
Test that other games aren't broken while playing movies. Test that movies aren't corrupted, don't skip, etc.
I tested many random games i own this time, and all of then worked. I wasn't able to test "billboards" in Test Drive game, and i suspect that PR can help it. Games which still have IPU entries in GameDB are still broken, no need to test them really. 

Thx to @raul3d for pointing out Play! commit which fixed Onimusha, that helped a lot.
This implementation is little bit different from what Play! do, I extended code search to BDEC (more precisely to mpeg2_slice). We also don't accept code if there is no data after first 8 bits after alignment. This break many games in non intra mode. But if we return false instead, and wait for data, everything seems to work ok. This was needed for Tiger Woods PGA games.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -15847,8 +15847,6 @@ SLES-53541:
   region: "PAL-M4"
   clampModes:
     vuClampMode: 3 # Fix black textures on characters.
-  gameFixes:
-    - SkipMPEGHack # Fixes hang after EA intro (IPU).
 SLES-53542:
   name: "Shadow the Hedgehog"
   region: "PAL-M5"
@@ -18979,8 +18977,6 @@ SLES-54867:
   region: "PAL-E-F"
   clampModes:
     vuClampMode: 2 # Fix black textures on characters.
-  gameFixes:
-    - SkipMPEGHack # Fixes hang after EA intro.
 SLES-54870:
   name: "FIFA '08"
   region: "PAL-E"
@@ -20825,25 +20821,6 @@ SLES-82038:
   compat: 5
   gameFixes:
     - SoftwareRendererFMVHack # Wrong white textures in FMV.
-  patches:
-    812C5A96:
-      content: |-
-        comment=patched by Kozarrov and ported by Prafull
-        //Improve menus, and other 2D displays handled on IPU
-        // Set IPU_DATA IVF to Intra
-        patch=1,EE,003c3678,word,3C040020
-        // Set IPU_CMD to IDEC, and set QSC to 2
-        // Why 2? Because it seems to work best.
-        patch=1,EE,003c368c,word,3C021002
-        // Old patch by shadow lady (still needed)
-        patch=1,EE,00104170,word,00000000
-        // Disable print - we need that loop!
-        // Without loop ipu break itself.
-        patch=1,EE,0010FDC8,word,34190180
-        patch=1,EE,0010FDCC,word,1720FFFF
-        patch=1,EE,0010FDD0,word,2739FFFF
-        patch=1,EE,0010FDD4,word,03E00008
-        patch=1,EE,0010FDD8,word,00000000
 SLES-82039:
   name: "Onimusha - Dawn of Dreams [Disc2of2]"
   region: "PAL-M5"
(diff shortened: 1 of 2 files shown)


refractionpcsx2:
Okay I went through games which have skipmpeg, plus one that I know has a patch, and these were my results

Penny Racers videos more reliable, still needs patches to avoid jump to null.
David Beckham Soccer still hangs (But does actually work with mild EE underclock, so likely timing related and not a result of this PR.)
ESPN NBA 2K5 still broken (Still requires SkipMPEG else the end of the ESPN video hangs)
Tiger Woods 06 Fixed
Tiger Woods 08 Fixed
Onimusha Dawn of Dreams Fixed (as promised)
Pro Mahjong Kiwame Next (Version 1.00) still broken, seems to loop reading on a CD sector (could be a bad old CDVD driver at fault) so still needs patch. 3.01 works fine (on master too)


But that's 3 (and I guess a fourth too) fixed, I didn't find anything broken. nice work! :)

seta-san:
> N NBA 2K5 still broken (Still requires SkipMPEG else the end of the ESPN video hangs)

the 2k games are more likely dev9 issues

refractionpcsx2:
This is unrelated, the DEV9 issue is patched.

refractionpcsx2:
> This broke videos on Tales of Rebirth. I cant send log or gsdump right now, wont be in my house for some time but I thought I would mention it anyway.

a GSDump is no good, it's a core problem. I have actually already noted it over at #2132

refractionpcsx2:
Okay so this PR does cause the videos to hang at the end of the video, the very last frame stops and it loops the last 5-10 frames, but the corruption can be solved with the EE Timing hack.

I'm not sure if this is a DMA problem that's readed its ugly head or if it's just forever checking for a start code and failing or something, idk

refractionpcsx2:
Looks like this PR basically showed up DMA timing errors, so I have redone them somewhat in #6313 and it seems to resolve both the glitching and the hangs.

refractionpcsx2:
hey, sorry, got another one for you @Goatman13 which broke with the addition of your PR.

Onimusha 3 Demon Siege dies right at the beginning during the lang/refresh selection, wonder if you could look in to it :)
[Onimusha 3 [SLES 51914] (E) (2022-05-03 21-26-16).zip](https://github.com/PCSX2/pcsx2/files/8835534/Onimusha.3.SLES.51914.E.2022-05-03.21-26-16.zip)

refractionpcsx2:
isn't 1 the only valid thing for start code detection though?

refractionpcsx2:
yeah the manual I have (which I should really read lol) says this

image

but it doesn't mention checking for 8bits of 0 beforehand like we do, idk where that comes from, or why we do it unaligned, it's all very confusing to me

refractionpcsx2:
I see.... I'm a little confused to what you mean, cos we don't remove zero bits until we're aligned, we read 8 bits, presumably to check for stuffing, and if it's 0 stuffed, then we align the stream, THEN check for start code.

refractionpcsx2:
well, yeah, it adds 7 then chops it to the next 8bit aligned position down from there, it doesn't quite remove 0's until it's aligned, more skips to an aligned state xD

just looking at the code for getbits8, it doesn't look like it returns 0 for the unaligned bits :/

		uint mask = (0xff >> shift);
		*(u8*)address = (((~mask) & readpos[1]) >> (8 - shift)) | (((mask) & *readpos) << shift);

refractionpcsx2:
Oh I never said it was wrong :P

refractionpcsx2:
Well the one I have above seems to just imply that keep going until it is 1, or is it interpreted as an error code? cos surely the only 2 things it should be expecting is 

1. zero stuffing
2. a start code

Am I wrong there? :/

refractionpcsx2:
I did just try changing it to say if it's greater than 1, set ECD and break, and that worked for Onimusha yeah

refractionpcsx2:
Good luck :) It'd be nice to have some sort of solution (hopefully helping other games too)

PR #6282, opened 2022-05-30
Fix blockdump loading on Mac and Linux

tellowkrinkle:
### Description of Changes
New blockdump code relies on the return value of FlatFileReader being the number of bytes read and not just -1 or 1

### Rationale behind Changes
It's nice to be able to load block dumps

### Suggested Testing Steps
Linux testing.  The old code was definitely wrong, but aio docs weren't clear enough for me to be sure that the new code is right.  Untested.


stenzek (approved):
Linux side sounds okay to me based on what I can see online about aio, but I haven't tested it.

refractionpcsx2:
Link tested it afaik, all was peachy. Might as well merge :)

Issue #6185, opened 2022-05-25
[BUG]: (Qt)Emulated controller detected as PS1 digital pad in Ridge Racer V


stuken:
### Describe the Bug

Using the Qt interface Ridge Racer V detects the emulated controller as a PS1 digital pad and no analogue controls are useable.
wxWidgets does not have this issue and the game will detect the controller as a Dualshock2.

Happens under both xinput or sdl input, with default out of box keyboard input and after binding controls.

### Reproduction Steps

Automatically bind controller inputs using xinput or sdl as the input source, boot game and navigate to options.
Controller options will display a ps1 contoller and analogue input is not useable.

### Expected Behavior

Parity with wxWidgets where game detects emulated controller as a Dualshock2 and analogue input is available.

### PCSX2 Revision

v1.7.2800

### Operating System

Windows 10 (64bit)

stuken:
Found the inconsistency between qt and wx. wx maps the stick to dpad inputs here:
https://github.com/PCSX2/pcsx2/blob/a26b91291ca4bcdb1a4357863683b4af3a6205c8/pcsx2/PAD/Windows/PAD.cpp#L668-L681

Ridge Racer V is one of those weird edge cases that will forcibly disable analogue so Qt still needs the analogue button exposed through the controller mapping like wx does.

stuken:
I messed around with the default pad state on reset to be analogue and ds2 and the game will send a change mode command with a to digital payload if it finds the pad in either of those states. I dont have access to my old ps2 to 100% verify the behaviour but this lines up with some forum and old gamefaqs posts asking how to enable analogue controls implying that they didnt work out of the box. So it seems to follow from there that the emulation behaviour is correct.

As for the wx side it to isnt setting it to ds2 on boot, I simply didnt notice the difference in control sensitivity from the true analogue mode and the dpad mapping and it does need to be explicitly enabled with the mapped analogue mode button.

This probably isnt a bug, but a feature request for exposing the analogue mode button.

RedPanda4552:
Makes sense. This behavior is in line with what I observed from the PS2 BIOS as well, if you program a controller to report that it is already working as an analog controller, the BIOS will send a request for it to revert to digital, something it will not do if the controller is already reporting digital.

I'm not entirely convinced that this is the full story, but for now I think it's enough to justify a feature request exposing the button. I don't know if I will be the one to do it, but I am watching this issue so if no one beats me to it I may eventually try my hand at this.

PR #6284, opened 2022-05-30
pad: Expose Analog Mode button.

stuken:
### Description of Changes
Exposes a bindable analog mode toggle through the ui and wires button events into pad state.

### Rationale behind Changes
Issue raised in #6185. Brings Qt to parity with wxWidgets which has the button bindable.

### Suggested Testing Steps
Test that button binding is saved and acted upon, also that well behaved games still accept input and will not switch back to digital when pad state is locked.
Currently tested against Ridge Racer V and Grand Turismo 2000 which both fail to enable analog controls by default. PR allows both to toggle between modes.
Also tested against Gran Turismo 3 which does not allow falling back to digital once ds2 mode is enabled.

### DIFF
--- a/pcsx2/PAD/Host/KeyStatus.cpp
+++ b/pcsx2/PAD/Host/KeyStatus.cpp
@@ -31,7 +31,7 @@ void KeyStatus::Init()
 {
 	for (u32 pad = 0; pad < GAMEPAD_NUMBER; pad++)
 	{
-		m_button[pad] = 0xFFFF;
+		m_button[pad] = 0xFFFFFFFF;
 
 		for (u32 index = 0; index < MAX_KEYS; index++)
 			m_button_pressure[pad][index] = 0;
@@ -107,6 +107,7 @@ void KeyStatus::Set(u32 pad, u32 index, float value)
 			1, // PAD_R2
 			9, // PAD_L3
 			10, // PAD_R3
+			16, // Analog
 			// remainder are analogs and not used here
 		} };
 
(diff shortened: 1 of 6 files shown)


pcsx2/PAD/Host/StateManagement.cpp
```diff
@@ -241,7 +241,21 @@ u8 pad_poll(u8 value)
 					b1=b1 & 0x1f;
 #endif
 
-				uint16_t buttons = g_key_status.GetButtons(query.port);
+				uint32_t buttons = g_key_status.GetButtons(query.port);
+				if (!test_bit(buttons, 16) && !pad->modeLock)
```
stenzek, line 245:
```suggestion
				if (!test_bit(buttons, PAD_ANALOG) && !pad->modeLock)
```

Avoids the magic number if the indices ever change.

pcsx2/PAD/Host/KeyStatus.cpp
```diff
@@ -107,6 +107,7 @@ void KeyStatus::Set(u32 pad, u32 index, float value)
 			1, // PAD_R2
 			9, // PAD_L3
 			10, // PAD_R3
+			16, // Analog
```
stenzek, line 110: Magic number here too, but we could probably just leave this one since the rest also aren't using the enum either... (I think it was originally a different order here, until I synced them up, the onepad code was quite... gross).

stenzek (approved):
Nice work! LGTM code-wise, difficult to tell about the UI from the xml alone, but I'm assuming it's fine.

pcsx2/PAD/Host/KeyStatus.cpp
RedDevilus: *LilyPad code sweating in bad code*

lightningterror (commented):
Can the commits the properly named with prefixes and such?

PR #6295, opened 2022-05-31
Qt: wire in tv shaders

stuken:
### Description of Changes
Adds TV shader to postprocessing section of graphics config in Qt interface.

### Rationale behind Changes
Feature is missing in Qt interface. Time to add it back.

### Suggested Testing Steps
Test that config is correctly persisted to ini, is updateable at runtime, correct post processing is applied and things dont fall over.


stenzek (approved):
Code changes LGTM, but untested

PR #6300, opened 2022-05-31
GS: Fix up wave filter shaders.

lightningterror:
### Description of Changes

GS: Fix up wave filter shaders.
Follow up from #6298
### Rationale behind Changes

Fixes broken shaders.
### Suggested Testing Steps

Test that the tv shader still works.

### DIFF
--- a/bin/resources/shaders/dx11/present.fx
+++ b/bin/resources/shaders/dx11/present.fx
@@ -135,12 +135,10 @@ PS_OUTPUT ps_filter_complex(PS_INPUT input) // triangular
 {
 	PS_OUTPUT output;
 
-	float2 texdim, halfpixel; 
-	Texture.GetDimensions(texdim.x, texdim.y); 
-	if (ddy(input.t.y) * input.t.y > 0.5) 
-		output.c = sample_c(input.t); 
-	else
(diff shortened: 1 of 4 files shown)


tellowkrinkle:
What is this `if` supposed to be checking?  With the new check, it'll be more likely to be true the further down you get in the image.  Is that what it's supposed to be?  Seems like a weird effect...

refractionpcsx2:
Wave is a weird effect, it's kind of like scanlines, but much more chonky.

tellowkrinkle:
Actually, `data.t.y` is a number from 0 to 1.  `dfdy(data.t.y)` is a very small number.  Multiplying those together will always get you a number less than 0.5, so that if statement is completely useless.  Either delete the if statement completely, or use `texdim.y`, which is clearly the originally intended value.

refractionpcsx2:
texdim.y does nothing, the effect is completely broken, unless you mean completely change the statement.

tellowkrinkle:
The point of the if statement in the current shader seems to be to disable the shader if the output resolution is lower than 2x the resolution of the image.  If you don't like it, delete the if statement and use the second case only.  But `dfdy(data.t.y) * data.t.y > 0.5` is wrong no matter how you look at it.

refractionpcsx2:
okay, that's fine, and you're right, it probably is wrong. but this was just bringing it in to par with the other renderers, because how it is right now, it actually does nothing, but with this change it will look like this on all renderers 

image

but yes it might need to be modified like the interlacing functions were, but I don't know the intended effect.

lightningterror:
Removed the first condition as suggested.

PR #6307, opened 2022-06-01
GS/TextureReplacement: Ignore replaced textures in DumpTexture

nev3rfail:
### Description of Changes
Added code that returns from DumpTexture function if texture is present in replacement texture cache or if texture exists in replacements folder.
It searches _only_ in cache if texture precaching is enabled and falls back to fs search otherwise.

### Rationale behind Changes
With this change we can dump only textures that need replacements instead of dumping every texture and then removing unneeded textures with scripts or by hand. 

### Suggested Testing Steps
Start dumping any game, edit texture with game logo with image editor, copy texture to replacements folder, delete texture from dumps folder, then start dumping the game again. You should not see game logo dumped for the second time.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSTextureReplacements.cpp
+++ b/pcsx2/GS/Renderers/HW/GSTextureReplacements.cpp
@@ -583,9 +583,9 @@ void GSTextureReplacements::ProcessAsyncLoadedTextures()
 
 void GSTextureReplacements::DumpTexture(const GSTextureCache::HashCacheKey& hash, const GIFRegTEX0& TEX0, const GIFRegTEXA& TEXA, GSLocalMemory& mem, u32 level)
 {
-	// check if it's been dumped already
+	// check if it's been dumped or replaced already
 	const TextureName name(CreateTextureName(hash, level));
-	if (s_dumped_textures.find(name) != s_dumped_textures.end())
+	if (s_dumped_textures.find(name) != s_dumped_textures.end() || s_replacement_texture_filenames.find(name) != s_replacement_texture_filenames.end())
 		return;
 
 	s_dumped_textures.insert(name);


pcsx2/GS/Renderers/HW/GSTextureReplacements.cpp
```diff
@@ -212,7 +212,7 @@ std::string GSTextureReplacements::GetGameTextureDirectory()
 	return Path::Combine(EmuFolders::Textures, s_current_serial);
 }
 
-std::string GSTextureReplacements::GetDumpFilename(const TextureName& name, u32 level)
+std::string GSTextureReplacements::GetFilename(const TextureName& name, const std::string& what, u32 level)
```
refractionpcsx2, line 215: Can you use a better variable name than "what" like "file_name" or something?

nev3rfail: oops. fixed.

refractionpcsx2: Thanks :)

pcsx2/GS/Renderers/HW/GSTextureReplacements.cpp
```diff
@@ -112,7 +112,7 @@ namespace GSTextureReplacements
 	static GSTextureCache::HashCacheKey HashCacheKeyFromTextureName(const TextureName& tn);
 	static std::optional<TextureName> ParseReplacementName(const std::string& filename);
 	static std::string GetGameTextureDirectory();
-	static std::string GetDumpFilename(const TextureName& name, u32 level);
+	std::string GetFilename(const TextureName& name, const std::string& subdirectory, u32 level);
```
stenzek, line 115: What was the point of changing this function?

stenzek (changes requested):

lightningterror:
I take it the pr was closed by accident?

pcsx2/GS/Renderers/HW/GSTextureReplacements.cpp
nev3rfail: To reuse it for replacement texture existence. It is not needed now, I removed it.

stenzek (approved):
LGTM

PR #6314, opened 2022-06-03
Port usb settings from wx to qt

kookycake:
Why do I want this change to be made?

This is because that way we can play the eyetoy games, singstar games, and other games that require usb settings. Look, I know that wx does this already, but I think it'd be much more convenient to have them all in one place, rather than going from wx specifically for the eyetoy games, and qt for the normal games.


F0bes:
What in the world is this.
Please make a feature request in the Issues category.

PR #6316, opened 2022-06-03
GameDB: Replace Choro Q HG, and Penny/Gadget Racers patches.

Goatman13:
### Description of Changes
Replace patches to less invasive solution. Patch for stack will likely always gonna be needed.
Second patch skip Vsync in tight loop on initial memory card check. So it fix game loading.
For more info why patch for Vsync is needed: https://github.com/PCSX2/pcsx2/blob/2333ff7b2df5a90698e25a4b86df9fccf5f27ec8/pcsx2/Counters.cpp#L602

### Rationale behind Changes
Fix Videos, no longer TLB miss. Vsync fully work apart from initial memory card check screen. 

### Suggested Testing Steps
Test Choro Q HG, and Penny/Gadget Racers games. 
Note: Make sure to test correct games. Gadget Racers SLUS_202.25 is correct game to test, Gadget Racers SLES_XXX.XX is really sequel to that game..

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -9061,13 +9061,12 @@ SLES-50252:
   patches:
     FBE2613D:
       content: |-
-        comment=Patch by Prafull
-        // Fixes VSync issues.
-        patch=0,EE,001a66ac,word,03e00008
-        // Avoids hanging at first loading screen.
-        patch=0,EE,001a6688,word,00000000
-        // Skips movies.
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Thanks for the new patches! Do we have any idea why the stack corruption happens?

refractionpcsx2:
Nice :) Thanks for the explanation! 

I can try a couple of the games, will report back

refractionpcsx2:
All 3 games work great, thanks!

PR #6317, opened 2022-06-03
MipsAssembler: Support li pseudo instruction

F0bes:
### Description of Changes
Support the li pseudo instruction

### Rationale behind Changes
It's valid mips

### Suggested Testing Steps
Try assembling with `li` in the debugger (wx only on master)

https://user-images.githubusercontent.com/29295048/171924953-655c7c93-bd60-4a03-88ac-8ffcfa031280.mp4


F0bes:
The resources I've used were incorrect and therefore, this is wrong :(

PR #6323, opened 2022-06-04
GS: Split convert and present shaders

stenzek:
### Description of Changes

Splits the presentation shaders (filters) out of the convert shader pack, and adds resolution-related uniforms (needed for CRT shaders).

TODO: Define the uniform buffer in Metal, test Metal.

### Rationale behind Changes

This is just a bit of stop-gap measure to allow other shaders to be added, until I have the time/motivation to integrate reshade into PCSX2.

### Suggested Testing Steps

Test post shaders on all APIs.

### DIFF
--- a/bin/resources/shaders/dx11/convert.fx
+++ b/bin/resources/shaders/dx11/convert.fx
@@ -91,30 +91,6 @@ PS_OUTPUT ps_filter_transparency(PS_INPUT input)
 	return output;
 }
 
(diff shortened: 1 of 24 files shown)


JordanTheToaster (approved):

labib2002:
wow dude, you really have been pulling miracles on pcsx2 these  past months, I think adding reshade would be another milestone

PR #6339, opened 2022-06-04
GHActions:macOS: Add Qt SVG plugin

tellowkrinkle:
### Description of Changes
Adds Qt SVG plugin to the GitHub Actions Mac Qt build

### Rationale behind Changes
We now use SVGs in our assets

### Suggested Testing Steps
Run the Qt build on macOS and see if icons render


nolrinale:
The icons are rendered correctly.

PR #6352, opened 2022-06-05
GameDB: Change upscaling fix Valkyrie Profile 2

RedDevilus:
### Description of Changes

HPO 1 to 2
### Rationale behind Changes

Normal Vertex has more issues than Special Vertex
### Suggested Testing Steps

Test VP2

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -1210,7 +1210,7 @@ SCAJ-20177:
   gameFixes:
     - VuAddSubHack
   gsHWFixes:
-    halfPixelOffset: 1 # Reduces bloom misalignment.
+    halfPixelOffset: 2 # Reduces bloom misalignment.
     roundSprite: 1 # Fixes area transition vertical lines and lessens red forest vertical lines.
 SCAJ-20178:
   name: "Ape Escape - Million Monkeys"
@@ -1301,7 +1301,7 @@ SCAJ-20197:
   gameFixes:
     - VuAddSubHack
   gsHWFixes:
-    halfPixelOffset: 1 # Reduces bloom misalignment.
+    halfPixelOffset: 2 # Reduces bloom misalignment.
     roundSprite: 1 # Fixes area transition vertical lines and lessens red forest vertical lines.
 SCAJ-20198:
   name: "Everybody's Tennis [PlayStation 2 The Best]"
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Special (Texture) causes graphical errors in the forest


Normal Vertex:
image

Special Texture:
image

JordanTheToaster:
> Special (Texture) causes graphical errors in the forest
> 
> Normal Vertex: image
> 
> Special Texture: image

Normal Vertex causes lines in trees 

image

Special 

image

refractionpcsx2:
I guess it's a pick your poison thing then..

Does round sprite full also cause any issues elsewhere? in the forest it lessens the green boxes further than half

coornio:
> I guess it's a pick your poison thing then..
> 
> Does round sprite full also cause any issues elsewhere? in the forest it lessens the green boxes further than half

It has an effect on the ghosting that HPO removes, so if you go for Full then the ghosting comes back. 
Speaking of, where may I get a dump of the red forest? :)

refractionpcsx2:
[vp2forest.zip](https://github.com/PCSX2/pcsx2/files/8840833/vp2forest.zip)

here you go

coornio:
Thanks. Frame Buffer Conversion + Preload Frame Data removes the red lines/noise introduced with HPO Special Texture. No effect on anything else from the looks of it.

refractionpcsx2:
Can confirm those options do seem to resolve it...

@JordanTheToaster @RedDevilus do those options have a negative impact elsewhere?

refractionpcsx2:
Preload Frame makes the FMVs flicker, so the SW mode switch for FMV's would also be required.

coornio:
Also noticed that "Disable Depth Emulation" resolves the thick red line on the bottom of the screen in the red forest.

refractionpcsx2:
I'd rather not disable depth, if it can be helped, that can fuck up post effects (and this game does a lot of them in cutscenes)

coornio:
> I'd rather not disable depth, if it can be helped, that can fuck up post effects (and this game does a lot of them in cutscenes)

Sure thing, just pointed it out in case it can turn out to be useful. I don't have the game so I can only check these through the dumps :)

lightningterror (changes requested):
I'd just remove HPO and let the user tinker with it since there's no good option.

lightningterror:
So what will we do about this?

refractionpcsx2:
Since no progress has been made on this and it's really down to user preference tbh, gonna close this.  Users can set it per game if they want something different.

PR #6361, opened 2022-06-06
GS: Add Lottes CRT to TV shaders.

stuken:
### Description of Changes
Adds a quick port of the GS.fx Lottes CRT implementation to the tv shaders option using the newly exposed frame and render window size uniforms. These allow for a nicer scaling of the effect at higher output resolutions compared to the old external fx implementation.

### Rationale behind Changes
GS.fx is not exposed via Qt and does not work with the vulkan renderer. This brings in a small subset of the missing functionality to the built in filtering options.

### Suggested Testing Steps
Locally teseted with idential output from each rendering backend. The source shader is simple enough but has not been used on vulkan that much. Should probably be verifed accross multiple gpu vendors.


lightningterror (changes requested):
Formatting can be better, misses metal implementation.

stuken:
Cleaned up the styling a bit, shoud be more sane than what was directly lifted from GS.fx. As for a metal implementation I dont have access to a mac so I'm not in a positon to verify anything I could try put together.

lightningterror:
Hmm, formatting still seems a bit wrong, should use tabs for indentation, probably run clang.

stuken:
Added metal implementation totally blind. After much ci abuse it's compiling but is completetly untested.

stuken:
Added tellowkrinkle's suggestions for cleaning up uniform access in the metal shader and some small cleanups of unused entry point arguments in the other backends. Output has been verified as working identically on every renderer now.

lightningterror (approved):
Needs to be tested.

refractionpcsx2:
> On the Windows Qt build the window size/aspect ratio isn't being applied correctly. The output seems to be stretched to 16:9 no matter the aspect ratio option and that frame is then cropped to the chosen aspect ratio in the settings.
> 
> DaH set to 4:3: image
> 
> In windowed mode the frame is stretched horizontally if the window is wider than what would provide a 16:9 area: image

to be fair, both of them are wrong :P

stuken:
This looks like the input uniforms are not being updated when you change the rendering aspect ratio. If you select the window scale multiplier menu option after changing the aspect ratio it does correctly change and you wont have a chopped image.

stenzek:
Uniforms are set immediately before the draw: https://github.com/PCSX2/pcsx2/blob/7b493ff4c807ef9d2fbd11767c82495155e1748e/pcsx2/GS/Renderers/DX11/GSDevice11.cpp#L709-L713

stuken:
The cropping should be fixed now. Tested on dx11/12, open gl and vulkan and full images are maintained at fixed aspect ratios and stretch to window. Still needs mac testing.

BreakPoints:
I can confirm, latest commit fixes the issue for me

stuken:
Rebased and commits squashed.

Issue #6268, opened 2022-05-29
[Feature Request]: Save State Hotkey


snowpin:
### Description

I know there's the default hotkeys we can customizer currently for flipping through save states and loading. 

I'm asking for the ability to find hotkeys for saving/loading each save state number individually without the need for cycling.

### Reason

This may not be something most people will use, but at least on the RetroAchievement Developer side I use this function religiously and having to cycle through or using a  drop down menu when scanning for memory addresses and trying to see differences adds a lot of time to my searches.

### Examples

See attached photo

Save State 1 - *hotkey here*
Load State 1 - *hotkey here*
Save State 2 - *hotkey here*
Load State 2 - *hotkey here*
Etc...

![RALibretro_8hYej4i7YD](https://user-images.githubusercontent.com/6635720/170876597-77a88601-17f5-42f1-9436-1cef4194fc66.png)

timenoe:
+1 to this request. It would be much appreciated!

redwizard42:
Agreed. Definite time save and help prevent saving to wrong slot.

coornio:
Give it a while. In the latest Wx, the numpad was used to save/load individual savestates (annoyingly for me who use it for my right half of the ps2 controller inputs) and I imagine that eventually the same will be exposed on the Qt end. There's so many rough edges to iron out :D

grimpunch:
My PR addressing this has passed all checks and is squashed now, and I believe @stenzek  has approved it, just waiting for merge 👍

grimpunch:
@snowpin does https://github.com/PCSX2/pcsx2/commit/8a6f55a63e16a54c664aa6e3b0dfc8a315f4ba76 address your needs well enough for this issue to be closed?

snowpin:
Sorry, for some reason it never notified me. Yes, the update was great :D Thanks.

PR #6373, opened 2022-06-07
Save/Load States via hotkey in QT

grimpunch:
QT: Adds the commands for loading/saving all numbered states via hotkeys, leaving them all unbound by default.

Addressing https://github.com/PCSX2/pcsx2/issues/6268.


stenzek (approved):
Not a fan of the 0-prefix numbering, but I'm considering removing the (later) sorting of hotkeys, which will solve this issue anyway.

PR #6381, opened 2022-06-09
Gamedb: add upscaling fixes  to 'Fast and the Furious'

Mrlinkwii:
helps said game upscale 
### Suggested Testing Steps

make sure CI  is happy


Maarcis12:
It seems that it just broke the FMV's in the game, suddenly game freezes whenever FMV starts to play.

RedDevilus:
> It seems that it just broke the FMV's in the game, suddenly game freezes whenever FMV starts to play.

These upscaling fixes should not have any effect that would cause freezing.

Issue #6237, opened 2022-05-28
[Feature Request]: More forgiving scroll in gamelist


coornio:
### Description

Each tick of the scroll in the gamelist when it is in Grid View moves the list by nearly a full page of items (if you can fit comfortably 3 rows, then it does 3 rows, if you can only fit 2 rows then it's 2 rows and so on).

I personally find that to be quite jarring. Since there's no smooth scroll start/stop animation to give a visual feedback on which way you scrolled (I'd really like that option, please consider :) I'll love you lots), the game list ends up simply popping up with a different chunk of games in a very abrupt way.

I would like if possible to give the user the option to define how many rows each tick of a scroll will move the list, be it 1 row, 1.2 rows, or 5. The experience will dramatically improve as a result, especially if there's animation like mentioned before.

This is a great website to review and test different easing algorithms for such a job. I think easeOutExpo is a great match for scrolling.
https://easings.net/#

Thank you for the consideration!

grimpunch:
By default you have 2 scroll modes in item views in Qt,
enum ScrollMode {
        ScrollPerItem,
        ScrollPerPixel
    };

I'm not sure if specific easing can be applied but swapping out the per-item scrolling for per-pixel would alleviate the choppiness.
Adding this line:
`m_list_view->setVerticalScrollMode(QAbstractItemView::ScrollMode::ScrollPerPixel);`
here
https://github.com/PCSX2/pcsx2/blob/aeef16334948570b59cc175690db2ac6663d364d/pcsx2-qt/GameList/GameListWidget.cpp#L121
would achieve that.

coornio:
a relative scroll unit can be adjusted if it's possible to calculate how many pixels a row of entries (zoom) would take up and extrapolate from that.

Surely an easing would make the animation much smoother and by extension a bit more professional looking, though not the aim of this feature request; more so the ability to scroll smaller amounts so that you know which way you're actually going :)

F0bes:
Closing as the PR issuing this has been merged. (Please put 'Fixes #XXXX' in your PR descriptions)

PR #6382, opened 2022-06-09
Qt: Update GameList scrolling to be per pixel 

grimpunch:
Qt: Update GameList scrolling to be per pixel to better indicate direction of scrolling to user

Addressing issue https://github.com/PCSX2/pcsx2/issues/6237


lightningterror:
Can you show comparisons, I'm unable to see a difference.

grimpunch:
Default scrolls multiple items per scroll unit, replacement Scrolls an amount of vertical pixels per scroll unit, in my case it was scrolling different amounts of items if the window was bigger/smaller before my change, then after it was consistent

stenzek (approved):

PR #6386, opened 2022-06-09
GameDB: SOCOM II: U.S NAVY SEALs NTSC Multi-Patch

xCENTx:
GameDB: SOCOM II: U.S NAVY SEALs NTSC Multi-Patch

This is a GameDB patch that will apply based on the current running game version. SOCOM 2 has 2 internal patches that will shift the .text segment of data . This change also applies to the single player version of the game as there is no difference in game memory from Single Player to Multi - Player/LAN

**DESCRIPTION OF CHANGES**
SOCOM II has an active community that has been successful in reviving the online multiplayer portion of the game.
With that accomplishment brought upon some patches that needed to be applied to improve the games playability.

One of these patches is the "Render Fix" (Originally created by Harry62)

The  Render Fix adjusts the GameWorld brightness by patching an instruction in the PlayerEyeAdjustment Function. Without this patch the game is susceptible to severe slowdowns (as low as 5-10fps) on certain maps, both for the single player campaign as well as the online multiplayer component.

The online team has been successful in supplying a workaround for this by pushing this patch directly to the user by means of memory poke when the client establishes a connection.

Being that there is now 2 versions of SOCOM 2 out in the wild (r0001 & r0004) and the Render Fix being at a different address in memory per version, its has become increasingly difficult to provide an all around fix. I originally intended to include the render fix with my last pull request but remembered that it would have potential issues and the last thing I wanted to do was disrupt the online service in any way.

Here is a video highlighting the problems encountered without the render fix being applied:
<p align="center">
 <video src="https://user-images.githubusercontent.com/80198020/172934613-2aae2171-c098-4ece-b33d-b5d7ef542bd1.mp4" alt="VIDEO">
</p>

**RATIONALE BEHIND CHANGES**
While the online portion of the game has received a lot of love, the same can't be said for the single player version of the game. I find this to be of a disservice to the greater PCSX2 community abroad as all of our findings should be shared and applied at a source level, but I'm going to assume they just didn't have a clue about how to provide the patch to everybody in a more efficient and official way. I've spent a long time trying to come up with a solution to this problem so that everybody can play the game , no matter the version and whether they are online or not, whilst engaging in an enjoyable experience.

**SUGGESTED TESTING STEPS**
[...]

PR #6390, opened 2022-06-10
GameDB: SOCOM II: U.S NAVY SEALs NTSC Multi-Patch

xCENTx:
### Description of Changes
This is a GameDB patch that will apply based on the current running game version. SOCOM 2 has 2 internal patches that will shift the .text segment of data . This change also applies to the single player version of the game as there is no difference in game memory from Single Player to Multi - Player/LAN

### Rationale behind Changes
This fix will bring the game to full speed as well as retain all normal game functions

### Suggested Testing Steps
Launch SOCOM 2 without a memory card and launch the single player campaign. Upon loading into the first mission a cutscene will ensue explaining the objectives for the mission ahead. Its during this brief intermission that we get our first glimpse of frame drops. After the cutscene ends and you gain control of your player, you will feel that the game is playing rather sluggishly even though the FPS counter will say the game is running at full speed.

Upon applying the r0001 render fix the game will immediately start playing at full speed and you will be able to play the mission at a clean 60fps.

The very same can be said for the r0004 version of the game , however to get your hands on the patch you will first need install a memory card. Then connect to the online portion of the game to get the download prompt. Proceed with the original testing steps after the patch has been downloaded

This patch does not lower the overall brightness from the game
[Video Link](https://youtu.be/oY1GTeTYKqM) for a full detailed breakdown on how this patch works compared to [my last pull request](https://github.com/PCSX2/pcsx2/pull/6386)
image

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -6984,6 +6984,17 @@ SCUS-97275:
   gameFixes:
     - EETimingHack # Get rids of DMA8/9 non-fatal errors.
     - VIF1StallHack # Fixes broken HUD.
+  patches:
+    default:
+      author: "NightFyre"
+      content: |-
+        author=NightFyre
+        // This will only be applied if the game is version r0001
+        patch=1,EE,D033CD90,extended,0000EF64
+        patch=1,EE,2033CD90,extended,00000000
+        // This will only be applied if the game is version r0004
+        patch=1,EE,D035A320,extended,000064C8
+        patch=1,EE,2035A320,extended,00000000
 SCUS-97276:
   name: "NFL GameDay 2004"
   region: "NTSC-U"


prafullpcsx2:
Try looking at the sky with your patch enabled.

xCENTx:
> Try looking at the sky with your patch enabled.

Is there any specific I am supposed to be looking at in the sky? I see no difference
##DISABLED
image

##ENABLED
image

refractionpcsx2:
As I said before, performance patches do not belong in the GameDB, not unless it's completely unplayable without it (you could probably disable hw readbacks to increase the speed instead of patching it)

if you want to put the patches somewhere, they go on the forum.

refractionpcsx2:
i decided to reverse my decision to close this, since we do already have patches which do this in our GameDB, and to not seem like a hypocrite, as long as it doesn't break anything, I'm fine with it.

prafullpcsx2:
<iframe width="560" height="315" src="https://www.youtube.com/embed/lhlAjvN8VWY" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>

refractionpcsx2:
> [youtube.com/watch?v=lhlAjvN8VWY](https://www.youtube.com/watch?v=lhlAjvN8VWY)

What does your patch do for Combined Assault, which we already have in the GameDB?

xCENTx:
> https://www.youtube.com/watch?v=lhlAjvN8VWY

The only way in which I am able to reproduce this brightness level is by having my in game brightness set to max., in which case, even without my patch ... you would get the very same results. Please verify that your brightness level has not be adjusted to compensate for the other render fix that is applied on the Socom Online Server.

I have recorded a short video to show that toggling this patch has no effect on the game world brightness whatsoever
LINK: https://youtu.be/pgCn0JmdSVQ
Notes: I am using an internal hook to apply the patch whilst being able to give a visual of everything that is taking place (Cheat engine shows the values at the addresses). I understand that this method of patching may not provide as conclusive evidence due to the nature of its application method ... but i could think of no other way to give an example of both patches being applied and reverted to give an example as to how it truly affects the game without multiple restarts.

prafullpcsx2:
> > [youtube.com/watch?v=lhlAjvN8VWY](https://www.youtube.com/watch?v=lhlAjvN8VWY)
> 
> What does your patch do for Combined Assault, which we already have in the GameDB?

When looking at the sun the game fps falls down by 90% or so due to the need of vif fifo. The patch skips that. It was initially made by CK (discussed with you) and ported by me.

refractionpcsx2:
yeah I thought it was a VIF patch, but I can't remember what it does and if it skips the sun (like this patch) or not.

prafullpcsx2:
> yeah I thought it was a VIF patch, but I can't remember what it does and if it skips the sun (like this patch) or not.

It just skips the need for vif fifo which fixes the game crawling. The sun and all surroundings still look the same.

refractionpcsx2:
Alright, then I think we've entertained this enough.

If you can come up with a patch which doesn't destroy the sun (and don't make up bullshit about only applying when you open the debugger), then maybe we can consider it.

Until then, patches go on the forum please.

refractionpcsx2:
Just FYI this is normal, speed is barely full

image

Then this is no patch, but ticking "Disable HW Readbacks", full speed 3 times over
image


So just ticking that option and the sky is less broken (but still broken) than yours and didn't require any special patches.

Issue #5914, opened 2022-04-18
[BUG]: Kung Fu Panda Broken Graphics


Levan7:
### Describe the Bug

In hardware mode the graphics are completely broken.

![gs_20220418155636_Kung Fu Panda _NTSC-U__SLUS-21757](https://user-images.githubusercontent.com/5770076/163805206-0bac3ffe-5a7f-457c-b2b3-e295de616d62.png)

Here is the gs dump
[snaps.zip](https://github.com/PCSX2/pcsx2/files/8505486/snaps.zip)

### Reproduction Steps

Start game in hardware mode

### PCSX2 Revision

v1.7.2633

### Operating System

Windows 10 (64bit)

### CPU

3570

### GPU

1050

### GS Settings

DX 12, 2x resolution scaling.

Issue #5366, opened 2022-01-21
[BUG]: Transformers: Revenge of the Fallen has several graphical bugs in HW mode


icup321:
### Describe the Bug

Transformers: Revenge of the Fallen has several graphical bugs in-game when running in HW mode.

### Reproduction Steps

>Start new campaign
>Skip first FMV cutscene
>Observe in-game cutscenes and gameplay graphics

### Expected Behavior

To not have major graphical glitches like other Krome Studios games.

Side note though, Star Wars: The Clone Wars - Republic Heroes is another Krome Studios game that is known to have graphical bugs but has been partially fixed with a patch, perhaps this game has a similar issue?

### PCSX2 Revision

v1.7.2252

### Operating System

Windows 10 (64bit)

### CPU

i5-8600K

### GPU

GTX 1050 Ti

### GS Settings

Issue is present with default settings at native resolution regardless of renderer.
None of the settings or hacks seem to affect it.

### Emulation Settings

Tested with default settings.

### GS Window Screenshots

Example 1
HW:
![gs_20220121043227](https://user-images.githubusercontent.com/28464642/150506977-50b329bb-6a9b-45a7-82ca-3f74182a0988.png)
SW:
![gs_20220121043233](https://user-images.githubusercontent.com/28464642/150507021-d067f965-91f6-4b10-a425-4cc58b8c8ca3.png)

Example 2
HW:
![gs_20220121043249](https://user-images.githubusercontent.com/28464642/150507067-36d6c56c-eb09-4a14-96cd-82f8ca126cb2.png)
SW:
![gs_20220121043300](https://user-images.githubusercontent.com/28464642/150507099-552ad7e0-b1f7-4e2b-ac75-4aa1e5cb7ec8.png)

### Logs & Dumps

[hw.txt](https://github.com/PCSX2/pcsx2/files/7912274/hw.txt)
[sw.txt](https://github.com/PCSX2/pcsx2/files/7912276/sw.txt)

Issue #3138, opened 2019-10-23
Call of duty 2 graphics issues 


Levan7:
Using v1.5.0-dev-3309-g5b6b38670 of PCSX2 speed settings are default (2)

Getting graphical issues when starting the game, if i run it in hardware mode (the dx11 just runs out of memory, the open gl doesn't).

![aa](https://user-images.githubusercontent.com/5770076/67427087-75442d00-f5ec-11e9-8e87-3f766974c4b2.gif)
[emuLog.txt](https://github.com/PCSX2/pcsx2/files/3764085/emuLog.txt)
[snaps.zip](https://github.com/PCSX2/pcsx2/files/3764086/snaps.zip)

kozarovv:
This bug affect other Treyarch PS2 games. Spiderman 2, COD2/3, Ultimate SpiderMan. Spiderman 3 texturing issue seems to be related, but partially fixable by enabling frame buffer conversion, maybe is good hint to start searching cause for other games there.

It looks little bit different on every title, but seems to be the same issue. Iirc there was idea that somehow mipmaping is cause of that, at least for Spiderman2.

atomic83GitHub:
No spiderman 3 is unrated as it's not a treyarch games and use a inhouse engine by Vicarious.

kozarovv:
Right, looks like Spiderman3 use some Vicarious proprietary engine, so is unrelated.

JeBobs:
Can confirm the same thing on g9c65b73ae (on both DX11/OGL) with CoD 3, testing on a GTX 1660 ti, Vega 10, and R9 380. Setting it to software fixes the texture issues, but the game becomes unplayable.

Shoegzer:
Confirmed here too, as tested against commit d05b31e91ce815cb92a3d23a6a8d9d566df27857.  This is by far the worst performing ps2 game of those I've tested - average 60% speed in OGL HW mode and 45% speed in OGL SW mode on a reasonably modern system.  The serious graphics issues with menus and ingame as others have reported related to HW modes still occur as well.  It's a shame for such a great and high-profile title in the ps2 library but understandable if the underlying issues are complex (timing related?).

Test system is liunx mint 19.3 and kernel 5.3 using pcsx2-flatpak self-compiled with gcc 9.3.0.  Nvidia gtx 1070M with 440.82 driver, and an Intel i7-7820HK.

[...]

Issue #3055, opened 2019-08-03
Shrek Super Slam - Texture issue


jy95:
***PCSX2 version:***
v1.5.0-dev-3219-g11187cff9

***Description of the issue:***
``(What's the problem? - Screenshots showing the issue if applicable)``

There is texture issue in the menus and the game : 
![gsdx_20190803130900](https://user-images.githubusercontent.com/9306961/62411279-da414580-b5f0-11e9-9195-3c411cdd98f8.png)
![gsdx_20190803130916](https://user-images.githubusercontent.com/9306961/62411283-df05f980-b5f0-11e9-9474-b7d8f10dfc31.png)
![gsdx_20190803131727](https://user-images.githubusercontent.com/9306961/62411301-1aa0c380-b5f1-11e9-992d-ee5a26c19c87.png)
![gsdx_20190803131740](https://user-images.githubusercontent.com/9306961/62411303-1e344a80-b5f1-11e9-9e97-9742a810c610.png)
![gsdx_20190803131747](https://user-images.githubusercontent.com/9306961/62411306-21c7d180-b5f1-11e9-9fea-0bd279ffc2be.png)

***PCSX2 log file:***

```
PCSX2 1.5.0-20190802200650 - compiled on Aug  3 2019
Savestate version: 0x9a0e0000

Host Machine Init:
	Operating System =  Microsoft Windows 10 Home (build 18362), 64-bit
	Physical RAM     =  16226 MB
	CPU name         =  Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
	Vendor/Model     =  GenuineIntel (stepping 0A)
	CPU speed        =  2.207 ghz (12 logical threads)
	x86PType         =  Standard OEM
	x86Flags         =  bfebfbff 7ffafbbf
	x86EFlags        =  2c100000

x86 Features Detected:
	SSE2.. SSE3.. SSSE3.. SSE4.1.. SSE4.2.. AVX.. AVX2.. FMA

Reserving memory for recompilers...

Loading plugins from D:\pcsx2-v1.5.0-dev-3219-g11187cff9-windows-x86\plugins...
	Bound   GS: GSdx32-AVX2.dll [GSdx 20190802200650 (MSVC 19.00 AVX2/AVX2) 1.1.0]
	Bound  PAD: LilyPad.dll [LilyPad (20190802200650) 0.12.1]
	Bound SPU2: Spu2-X.dll [SPU2-X 20190802200650 2.0.0]
[...]

PR #6401, opened 2022-06-12
GS: Support using SW renderer for texture decompression sprites

stenzek:
### Description of Changes

A number of games including the Spiderman series, True Crime: New York City, Kung Fu Panda, L.A. Rush, Call of Duty series, and others, take the idea of texture streaming a step further. Instead of just transferring the textures on demand, they also stored them in RAM in a "compressed" form (similar to VQ texture compression), and "decompressed" them on the GS just before they were drawn! A very clever technique, since the GS had no form of texture compression, outside of "normal" indexed/palette textures. They did this by performing several small sprite draws to apply an additional palette prior to the normal draw.

The result was a disaster for our texture cache. Not only did you have to worry about keeping all these tiny "textures" inside one larger framebuffer, but you also had to merge them back together into a single texture prior to draw, and dispatch all of the sprite draws, which can be upwards of 1,000 per frame. As expected, mobile GPUs absolutely hated it, and desktop GPUs weren't much better. Not only that, but in most cases, the TC couldn't track all this mayhem going on, and you got completely broken textures in the hardware renderers.

Previously, the only solution was to use the software renderer. But, can we identify these "texture decompression" draws, and use the software renderer only for these? The answer was yes, after some experimentation, we found a heuristic which worked for these draws, but did not break others. Another advantage, because the texture data is now present in local memory (due to software being used), they result in hits for the hash cache, further improving performance, and enabling texture replacement.

I don't like the hack for invalidation, but outside of a riskier change rewriting the whole thing (and improving my understanding of local memory at the same time), at least it's safe..

### Rationale behind Changes

Fixes https://github.com/PCSX2/pcsx2/issues/5914
Fixes https://github.com/PCSX2/pcsx2/issues/1320
Fixes https://github.com/PCSX2/pcsx2/issues/5366
Fixes https://github.com/PCSX2/pcsx2/issues/3138
Fixes https://github.com/PCSX2/pcsx2/issues/775
Fixes https://github.com/PCSX2/pcsx2/issues/3055

Any others?

### Suggested Testing Steps

Test affected games.

### DIFF
--- a/.github/workflows/scripts/lint/gamedb/lint.py
+++ b/.github/workflows/scripts/lint/gamedb/lint.py
@@ -61,6 +61,7 @@ allowed_gs_hw_fixes = [
     "roundSprite",
     "texturePreloading",
     "deinterlace",
+    "cpuSpriteRenderBW",
 ]
 gs_hw_fix_ranges = {
     "mipmap": (0, 2),
@@ -70,6 +71,7 @@ gs_hw_fix_ranges = {
     "halfPixelOffset": (0, 3),
     "roundSprite": (0, 2),
     "deinterlace": (0, 7),
+    "cpuSpriteRenderBW": (1, 10),
 }
 allowed_speed_hacks = ["mvuFlagSpeedHack", "InstantVU1SpeedHack", "MTVUSpeedHack"]
 # Patches are allowed to have a 'default' key or a crc-32 key, followed by
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -9433,6 +9433,8 @@ SLES-50288:
   name: "Stuntman"
   region: "PAL-M5"
   compat: 5
+  gsHWFixes:
+    cpuSpriteRenderBW: 4 # Fixes textures.
 SLES-50296:
   name: "Gift"
   region: "PAL-F"
@@ -13507,6 +13509,8 @@ SLES-52371:
 SLES-52372:
   name: "Spider-Man 2"
   region: "PAL-M5"
+  gsHWFixes:
+    cpuSpriteRenderBW: 1 # Fixes textures.
 SLES-52373:
   name: "Champions of Norrath"
   region: "PAL-E-S"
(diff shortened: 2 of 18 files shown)


stenzek:
> That is a very big commit regarding compatibility indeed. I wonder if the hash cache will have a negative impact on this.

As the textures are block-decompressed in local memory, the hash cache works as expected; some of these games end up with near zero texture uploads per frame. That also means that replacing these textures should be possible.

> I have one question regarding this, can the logic be extended to the FBW screen-clear games as well?

I already believe some of these are already detected in the existing code.

AhayriSG:
https://youtu.be/aIuVne2HKAg?t=310

stenzek:
> As for FBW screan-clear, this seems to only be partially detected but the texture-cache itself stands no chance of actually being able to emulate such a behavior. And as far as I am aware, the only solution to this is the software render.

I'd have to look into what you're describing more closely, but if the expectation is that the clear is propagated to the hardware targets, this PR may not help (it only works in local memory, not the hardware targets). However, in theory, it should set the dirty area of any overlapping targets (and then that would get preloaded)... but the TC tends to struggle when there's overlap involved and it's not pointing directly at the target.

RedDevilus:
Probably the shooter games like call of duty are affected with texture issues.

JordanTheToaster (approved):

refractionpcsx2:
Can you provide your graphics settings @githubribbon45 ? Need to know what your texture preloading etc are set to.

refractionpcsx2:
good god, what the fuck is that? Okay.

stenzek:
Your settings are completely stupid. Reset to defaults, only enable hash cache and upscaling, and it'll run fine.

refractionpcsx2:
And in the process completely ignored the "Recommended value" part of the description.

Mrlinkwii (approved):

Anuskuss:
So is it really supposed to be called `cpuSpriteRenderBW`?

Sportik-228:
Hello! can u help me with True crime NYC? PCSX2 v 2.1.56 has flickering textures in the form of dark squares, i used different game versions, graphic settings vulkan, ogl, d3d (SW is good) and have same problem. I dont understand how to install your fix, excuse me
True Crime - New York City_SLES-53616_20240810103001
True Crime - New York City_SLES-53616_20240810103002

PR #6406, opened 2022-06-13
Core: Tweaks to FIELD switch timing and INTC Hack

refractionpcsx2:
### Description of Changes
Adjusts the FIELD switch timing for PAL which is 3 scanlines instead of 3.5 (NTSC), probably won't affect much, if anything.
Adjusted the INTC hack to wait 8 cycles after the last event for INTC reads

### Rationale behind Changes
INTC:  if a game is in a tight loop reading it, else it could errantly skip to the next event which could be thousands of cycles away, this could incorrectly cause the EE to slow down.

Counter/GS: Field flip time was too late on PAL as it was timed for NTSC, now corrected.

### Suggested Testing Steps
Eh probably not worth testing, I guess just make sure nothing explodes.

No known fixes, just an accuracy update.

### DIFF
--- a/pcsx2/R5900.h
+++ b/pcsx2/R5900.h
@@ -260,6 +260,7 @@ alignas(16) extern fpuRegisters fpuRegs;
 alignas(16) extern tlbs tlb[48];
 
 extern u32 g_nextEventCycle;
+extern u32 g_lastEventCycle;
 extern bool eeEventTestIsActive;
 extern u32 s_iLastCOP0Cycle;
 extern u32 s_iLastPERFCycle[2];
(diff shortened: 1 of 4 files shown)


pcsx2/R5900.h
```diff
@@ -260,6 +260,7 @@ alignas(16) extern fpuRegisters fpuRegs;
 alignas(16) extern tlbs tlb[48];
 
 extern u32 g_nextEventCycle;
+extern u32 g_lastEventCycle;
```
stenzek, line 263: Should this be serialized/saved to state? (given it's accessed in HwRead, which doesn't have cpuEventTest in its path).

refractionpcsx2: It could be, but it doesn't really matter since it will basically get updated pretty darn quickly and it's only used for the INTC hack. So it could be, I just couldn't justify bumping savestates just for that.

PR #6409, opened 2022-06-13
PAD Linux: correct conversion of analog stick values.

### LINKED ISSUE #6408: [BUG]: Right analog stick input fed to the game is asymmetrical on Linux version.
### Describe the Bug

Version: v1.7.2889 and v1.7.2924 installed from Flathub
OS: SteamOS (Arch-based) on Steam Deck.
Tested game: Call of Duty: Finest Hour

Input data of right analog stick that is fed into the game is not symmetrical. It looks like pushing the right stick to the left causes proper response (camera turning left with full speed), but when pushing it to the right the response is much weaker (camera is turning slower). This issue is present when using Steam Deck controls, external USB controller, as well as keyboard keys mapped to right analog stick in the Gamepad Settings.

The issue is not present in Windows version of PCSX2 (in the same game).

### Reproduction Steps

Using keyboard:

1. Use gamepad settings to map LEFT ARROW key to right joystick left direction.
2. Use gamepad settings to map RIGHT ARROW key to right joystick right direction.
3. Start a game that uses right analog stick input.
4. Press LEFT ARROW key. Notice the response to analog input.
5. Press RIGHT ARROW key. Notice the response to analog input.
6. Notice the difference in response strength.

Using gamepad:

1. Start a game that uses right analog stick input.
2. Push gamepad's right stick to the left edge. Notice the response to analog input.
3. Push gamepad's right stick to the right edge. Notice the response to analog input.
4. Notice the difference in response strength.

### Expected Behavior

Pushing right stick to the left edge (or pressing a key mapped to the left direction of right stick) should generate a response of the same strength (camera speed) as pushing the stick to the right edge (or pressing a key mapped to the right direction of the stick). Correct behavior may be observed on Windows version of the emulator.

### PCSX2 Revision

v1.7.2924

### Operating System

### If Linux - Specify Distro

SteamOS (Arch-based)

kmicki:
### Description of Changes
Fixes #6408.
Makes feeding the emulator with analog stick values spanning 0x00-0xFF instead of 0x00-0xFE. That fixes assymetry of response to analog stick input in games.

Replacing division with bit shift got rid of rounding negative values up and allowed force to span 0x80 - 0x7F. Before, it spanned 0x81 - 0x7F.  For non-reversed input, 1 is added to the final value. This makes the result span values 0x00-0xFF instead 0x00-0xFE.

### Rationale behind Changes
Before fix, conversion generated values between 0x00 and 0xFE, leaving 0xFF out.
That caused assymetrical response from some games between between up and down or left and right camera/player movement.

### Suggested Testing Steps
Tested in Call of Duty: Finest Hour. In that game a difference between 0xFE and 0xFF in analog stick value is very perceptible for some reason. Might be tested also in other games.

### DIFF
--- a/pcsx2/PAD/Linux/KeyStatus.cpp
+++ b/pcsx2/PAD/Linux/KeyStatus.cpp
@@ -69,11 +69,11 @@ void KeyStatus::press(u32 pad, u32 index, s32 value)
 		// Force range :			  80 -> 0  -> 7F
 		// Normal mode : expect value 0  -> 80 -> FF
 		// Reverse mode: expect value FF -> 7F -> 0
-		u8 force = (value / 256);
+		u8 force = (value >> 8);
 		if (analog_is_reversed(pad, index))
 			analog_set(pad, index, m_analog_released_val - force);
 		else
-			analog_set(pad, index, m_analog_released_val + force);
+			analog_set(pad, index, m_analog_released_val + force + 1);
 	}
 }
 


lightningterror (commented):
Can you properly name the commit along with prefixes?

PR #6412, opened 2022-06-14
Misc: Fix a couple of things in GS config and d3d12 renderer

lightningterror:
### Description of Changes

GS-d3d12: Remove scale factor clamp.
Already handled in config so it's duplicate.

GS: Make sure upscaling hacks are disabled on native res.
There was a regression that caused upscaling hacks to still be enabled on native res.

GS-wx: Disable TX x y offsets on native res:
We already gray out the option on Qt, and it should be an upscaling hack only.
### Rationale behind Changes

Regression fixes, consistency.
### Suggested Testing Steps

Make sure the options work as described.

### DIFF
--- a/pcsx2/Pcsx2Config.cpp
+++ b/pcsx2/Pcsx2Config.cpp
@@ -654,13 +654,16 @@ void Pcsx2Config::GSOptions::MaskUserHacks()
 
 void Pcsx2Config::GSOptions::MaskUpscalingHacks()
 {
-	if (UpscaleMultiplier == 1 || ManualUserHacks)
+	if (UpscaleMultiplier != 1 && ManualUserHacks)
 		return;
 
 	UserHacks_AlignSpriteX = false;
 	UserHacks_MergePPSprite = false;
+	UserHacks_WildHack = false;
 	UserHacks_HalfPixelOffset = 0;
 	UserHacks_RoundSprite = 0;
+	UserHacks_TCOffsetX = 0;
+	UserHacks_TCOffsetY = 0;
 }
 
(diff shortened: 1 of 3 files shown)


JordanTheToaster (approved):

pcsx2/Pcsx2Config.cpp
```diff
@@ -609,6 +609,9 @@ void Pcsx2Config::GSOptions::ReloadIniSettings()
 	GSSettingStringEx(ShaderFX_Conf, "shaderfx_conf");
 	GSSettingStringEx(ShaderFX_GLSL, "shaderfx_glsl");
 
+	if (UpscaleMultiplier == 1u)
```
stenzek, line 612: Just commenting here for completeness, this already happens in `VMManager`.

pcsx2/Pcsx2Config.cpp
```diff
@@ -650,15 +653,23 @@ void Pcsx2Config::GSOptions::MaskUserHacks()
 #endif
 }
 
-void Pcsx2Config::GSOptions::MaskUpscalingHacks()
+void Pcsx2Config::GSOptions::UpscalingHacks()
 {
-	if (UpscaleMultiplier == 1 || ManualUserHacks)
```
stenzek, line 655: We discussed this on discord, but the only issue here is the condition should be inverted (`if (UpscaleMultiplier != 1 && !ManualUserHacks)`) in `MaskUpscalingHacks()`, no need to create an additional function.

stenzek (changes requested):

stenzek (approved):
lgtm

PR #6413, opened 2022-06-14
GS: Migrate function pointer to lambda

ricky65:
Compilers can inline lambda for superior performance.


stenzek (approved):
I think you'd find it'd produce exactly the same code either way here (tiny static function is basically guaranteed to get inlined). But it's only used once, so this is fine.

PR #6415, opened 2022-06-14
Turbo (Hold) Hotkey binding

### LINKED ISSUE #5538: [Feature Request]: Fast Forward (Hold)
### Description

When you Press the Fast Forward Button it stays in Fast Forward and Turns Off when you let go instead of of Pressing to Start and Then Press again to Stop.

### Reason

Find it's easier to use the Hold instead of Pressing. Just makes it Easier to use Fast Forward and don't fast forward anything Accidentally

### Examples

FF RA

grimpunch:
Fixes #5538
Acts as a temporary flip-flop switch while held that either enables turbo while held or disables turbo while held if you've already toggled turbo.

Default bind is Keyboard/Period but happy to make this unbound if preferred.

### DIFF
--- a/pcsx2/VMManager.cpp
+++ b/pcsx2/VMManager.cpp
@@ -134,6 +134,7 @@ static u32 s_active_no_interlacing_patches = 0;
 static s32 s_current_save_slot = 1;
 static u32 s_frame_advance_count = 0;
 static u32 s_mxcsr_saved;
+static std::optional<LimiterModeType> s_limiter_mode_prior_to_hold_interaction;
 
 bool VMManager::PerformEarlyHardwareChecks(const char** error)
 {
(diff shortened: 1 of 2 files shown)


lightningterror (commented):
Commit should be better named with prefix.

grimpunch:
> Commit should be better named with prefix.

I've renamed the commit with the prefix

pcsx2/VMManager.cpp
```diff
@@ -133,6 +133,8 @@ static u32 s_active_no_interlacing_patches = 0;
 static s32 s_current_save_slot = 1;
 static u32 s_frame_advance_count = 0;
 static u32 s_mxcsr_saved;
+static LimiterModeType s_limiter_mode_prior_to_hold_interaction;
```
stenzek, line 136: Could probably collapse this to a `std::optional`. And it should be reset on shutdown/initialize.

grimpunch:
Updated with std::optional usage and reseting on initalize/shutdown, 
should be working as it did prior with cleaner code.

grimpunch:
Sorry, bad squash, i've picked up the changes from updating. 
I'll fix my changes

grimpunch:
ok should be good to go now.

stenzek (approved):

PR #6416, opened 2022-06-14
GS: Only enable scanmsk offset on frames that need it

### LINKED ISSUE #6411: [BUG]: Metal Gear Solid 2 possible regression; Interlaced blur after throwing punch/starting a Codec call
### Describe the Bug

Possible regression from this:  https://github.com/PCSX2/pcsx2/pull/4385 .

Blur, looks like an interlaced frame, Happens in Metal Gear Solid 2 (PAL and NTSC) after throwing a punch or making a Codec call (it will appear when the faces show up, after the ringing noise). This blur remains until you load a quick save or restart the game.

I found the two builds where the difference started (using https://buildbot.orphis.net/pcsx2/index.php?m=fulllist ) :

Working:
BEFORE (and incl.) v1.7.2159    refractionpcsx2    2021-12-26 11:37:43    Download    VS: Fix filters

Broken:
AFTER (And incl.) v1.7.2160    Sergeanur    2021-12-22 11:56:38    Download    GS: Simulate scan mask

I tested quite a few builds before and after these, showing the same difference (meaning it probably started here). To test I used the normal pcsx2.exe, changed from opengl to software and turned off Texture Filtering of Display. It also happens in HW mode. No deinterlacing modes change anything.

### Reproduction Steps

Anywhere in the game, create a quick save, then load it. This will reset the image to "sharp". Throw a punch or start a Codec call (when the faces appear), and the blur will appear, remaining until you quick load again. No Texture Filtering of Display will help to see it more clearly.

### Expected Behavior

The punch should not introduce any blur of either 2D or 3D/fullscreen elements. The older builds don't have this issue.

### PCSX2 Revision

v1.7.2908

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

### CPU

i7-9700

### GPU

RTX 2080

### GS Settings

Default, but disable Texture Filtering of Display to more clearly see the issue.

### Emulation Settings

_No response_

### GS Window Screenshots

BEFORE
AFTER

### Logs & Dumps

The snaps/dumps are too big to upload, hopefully I am allowed to link it here: https://drive.google.com/file/d/1KEBpaVVNt62UmhDr-bVn1x73Rph7SGRe/view?usp=sharing 

They show before and after the punch, and also a multi where the punch button is pressed.

refractionpcsx2:
### Description of Changes
Only enable DISPLAY merge offset for scan mask for frames which actually use scan mask

### Rationale behind Changes
Some users found it frustrating that once the game used scanmask it made the picture "blurry" and then it never went back again.  Problem is, in the case of Metal Gear Solid 2, the game is *supposed* to have that blur all the time, but we disable it to give people a nicer picture, however scan mask (MGS2 uses for some transparency effects) doesn't work if we do that, so it needs to be offset by 1 pixel, causing the picture to shift down and the picture to become blurrier.  This can be mostly mitigated/made less noticeable by increasing your internal resolution.

### Suggested Testing Steps
Test GT4, MGS2, MGS3, Alpine Racer 3, make sure they all still work okay.

Fixes #6411 as best as possible. (it was either this or always blurry)

### DIFF
--- a/pcsx2/GS/Renderers/Common/GSRenderer.cpp
+++ b/pcsx2/GS/Renderers/Common/GSRenderer.cpp
@@ -374,6 +374,8 @@ bool GSRenderer::Merge(int field)
 		}
 	}
 
+	m_scanmask_used = false;
+
 	return true;
 }
 


lightningterror:
Tested GT4 and is fine, didn't test others.

refractionpcsx2:
I tested dumps of MGS2,  3 and Alpine Racer 3

PR #6106, opened 2022-05-20
GS: Reduce blurring when upscaling on some games

refractionpcsx2:
### Description of Changes
Reduces the blurring in some games which expect the interlacing to give so much of an offset and it not gelling well with upscaling which until now reduced this offset.

Option added in the Graphics Settings to disable it as games such as ICO respond poorly to this new option.

Thanks to Stenzek for helping diagnose what was going on with my changes (which didn't initially work properly)

### Rationale behind Changes
Offset for interlace was per line, which is kind of what we want but it breaks the picture in some games, this corrects it. This applies only to games which use FFMD (Full Frame Rendering, aka they only drew half a frame and it's being stretched over the screen)

### Suggested Testing Steps
Compare games when upscaled to see if the picture is any clearer, won't affect everything.

--------------------------------------------------------------------------------------------------------------

Examples at x6 upscale, please view at full size in order to see the changes clearly.

Raging Blades: (Everything looks blurry, still not perfect, requires Blend BFF deinterlacing):

Before:
image

After:
image

Baldurs Gate 2: (Check the text for the language select)

Before:
image

After:
image

Armored Core Last Raven:  (Look at the lines on the platform and the compass at the top)

Before:
image

After:
image

WRC Extreme II: (Look at the barrier adverts or hills in the distance)

Before:
image

After:
image

EverBlue: (Not really a fix but blurring severely reduced when upscaled)

Before:
image
After:
image

PR #6417, opened 2022-06-14
GS-PCRTC: Add option to Enable/Disable Anti-Blur mechanisms

refractionpcsx2:
### Description of Changes
Adds option to enable/disable the internal Anti-Blur hacks to display exactly as developers intended, to blur out your CRT screen.
Adjusts single pixel DISPLAY offsets to work better with upscaling.
Cleaned up some FIELD nonsense in the interlacing code.
Fixed bob deinterlacing, broken by changing how the interlace offsets work in #6106

### Rationale behind Changes
For... accuracy?  I guess it stops MGS3 "bobbing" when it goes in to scan mask mode.
Other stuff was just tweaking or regression fixing.

### Suggested Testing Steps
Run games, toggle the "Anti-Blur" PCRTC option, make sure it doesn't explode and look extra shitty like you've smeared vaseline on the screen*

*doesn't happen with quite a lot of games.


jonasos3:
Thanks for helping me with figuring out how to download these builds! 

This option appears to do what it says, and increased accuracy is always a plus IMO. Tested with MGS2, looks perfect(ly blurry, as was intended)! :)

jonasos3:
Although.. This works as intended when using software rendering, but on any of the hw renderers WHEN UPSCALING, the issues from before creep back again. When punching or starting a codec call, the screen changes for a split second on each punch/kick, or for the duration of the call. This happens when the new option is either on or off.

refractionpcsx2:
> Although.. This works as intended when using software rendering, but on any of the hw renderers WHEN UPSCALING, the issues from before creep back again. When punching or starting a codec call, the screen changes for a split second on each punch/kick, or for the duration of the call. This happens when the new option is either on or off.
> 
> EDIT - To simplify, the new option works as intended when running at native res. When upscaling however, there seems to be an issue whether the option is on or off; blur/"bobbing" when throwing a punch or when in a call. This blurring is sometimes also accompanied by a stutter or framedrop, making it very noticable even on higher resolutions where the visual effect is lessened.

of course it will, the problem is kind of in reverse. Here's the problem.

When we interlace/scanmask, that means we have to alternate every line, as you upscale those lines get thinner, much thinner than they were supposed to be.

So in theory, if you're upscaling x3, that means we should draw 3 lines and skip 3 lines (interlace) right? Nope! This would look terrible when upscaling and completely negate the effect, so when scan mask is in use, the offset needs to be 1 line, regardless of upscale.

So what you're experiencing now, is the more you upscale, the more it shifts UP (instead of down)

refractionpcsx2:
I'll see if I can cheat with this.

refractionpcsx2:
okay @jonasos3 after 4 hours of screwing around, I've gotten it to work nicely with upscaling.  

Of course it still jots down without the "Disable Anti-Blur" and it's now consistently doing that through upscaling. So anybody wanting to have it look nice and not jot down a little will need to disable them, it looks pretty good at high resolutions anyway.

I did try to get the best of both worlds, but it really wasn't happening, that was an absolute no go.

jonasos3:
Incredible! I'll check it out when I get home from work, thanks for all your hard work

jonasos3:
This seems to work as intended! Tested with MGS2 on native, upscaled, etc, works as you explained :) Happy to have a way of displaying the game (and other games) as they were made at the time, warts (blur) and all.

For people who want to upscale, but don't want the occasional stutters when punching/Codec calling, there could maybe be a way of adding a toggle for the scanmask filter itself? This would again break transparancy of water effects, as before the original fix, but also would allow for perfect (non-blurry) pixels without the "bobbing" that happens when the Disable Anti-Blur option is disabled. This would also help people who want the game in native res without the blur, but don't mind the water being broken if it means a smoother experience (with no occasional stutters).

refractionpcsx2:
if we disable scanmask, the windows, water, glasses, radar etc, will all break, so that's not happening.

Anuskuss:
Could this option be inverted? Defaulting to on and called simply "Anti-Blur". Double negatives are sometimes hard to understand and calling it like that better conveys that it's an (optional) enhancement.

refractionpcsx2:
No, because the collective internet would shit a brick, and I don't want to deal with the support.

Anuskuss:
The behaviour wouldn't change though - just the name.

refractionpcsx2:
Oh the behaviour has to be reversed for what you suggested.

Anyway, I've done it, so it's called "Anti-Blur" and it's ticked by default.

Shoegzer:
This is fantastic, thank you!  Even if defaults should IMO remain true to PS2 accuracy as a guiding principle, at least there's an option to turn such features off.  I understand certain people want stuff like this, just never understood why for various reasons, including the fact that most ps2 games can now be experienced in many other ways.  Over time I hope pcsx2 can be the reference implementation that preserves the original experience as its game devs intended, filtering out the noise of modern "improvements" to image quality etc.

Anuskuss:
> Anyway, I've done it, so it's called "Anti-Blur" and it's ticked by default.

Perfect, thanks!

I just tested it and I can't tell which one I like more - but it's pretty subtle either way (from couch distance anyway). Kinda reminds me of when N64 started emulating the VI. I never realized how blurry it's actually supposed to look like.

refractionpcsx2:
It becomes more subtle the higher the resolution, however some games it's very obvious

jonasos3:
> if we disable scanmask, the windows, water, glasses, radar etc, will all break, so that's not happening.

Per now we have a few different ways of enjoying the game, simply put:
1. Anti-Blur on, the new default toggle; here the pixels are sharp and sexy, but the bobbing when doing a punch-punch-kick combo for example is quite annoying visually, especially in native res. When displaying water, windows, reflections etc it's less of an issue, but still a noticeable shift when moving around a stage and water enters the frame. The bobbing is sometimes also accompanied by a stutter/framedrop, making it noticeable even on high resolutions.
2. Anti-Blur off. The most accurate way as compared to real hardware. Looks great, although a bit too blurry for some. Mitigated by upscaling, but some people will want to run at native (me included).

A 3rd way would be to add an option to revert to before this PR https://github.com/PCSX2/pcsx2/pull/4385 . This again breaks transparancies, water etc, but allows for playing the game with perfectly sharp pixels WITHOUT the annoying bobbing. Given the chance, this would be my preferred way of playing. Just adding my thoughts on the issue, always looking for the (closest to) perfect way of playing my fav games :) Thanks for all you have done, sorry to bug you again!

refractionpcsx2:
3 is never happening, I told you that before, 2 is the best you're going to get.

jonasos3:
Fair enough!

jonasos3:
That's too bad, but that's how it is. The old builds from before https://github.com/PCSX2/pcsx2/pull/4385 fulfill the criteria of "3" for anyone interested in sharp pixels (and are distracted by the "bobbing") but don't mind the broken water, glasses etc :)

refractionpcsx2:
Well the old builds had incomplete emulation, and we're not going to remove emulation just to make a hack work better, that's the worst idea in the history of emulation.

jonasos3:
The end result in this particular case is that a certain way of enjoying the game has been altered/removed, and that an older build is better suited for those who want this exact (nr. 3) experience. Of course I'm not going to argue about it with someone who knows his shit the way you do, I'm just putting forward a highly specific end-user viewpoint :) Other available settings like Mipmapping, Auto-Flush etc will make certain games "break"/render incorrectly when on/off, and this is specified in the GUI/tooltips, so I guess I just don't see what would be different about this particular thing

refractionpcsx2:
Well if people are unhappy with the emulator improving properly, they are free to use an old version. There is no way I can make this game work properly and not be blurry in lower resolutions.

As stated as long as you put the resolution up, you can barely (if at all) notice it.

jonasos3:
Sure, I understand many things have to be a compromise. Only, after https://github.com/PCSX2/pcsx2/pull/4385 one of the available compromises for selection has been removed (and another one added, which is great!). This is a highly specific use-case I'm sure, I might be the only person in the world who would even care

refractionpcsx2:
Going on about it isn't going to change my mind.

PR #6425, opened 2022-06-15
PAD: changed freeze data size in Linux for cross-platform compatibility

### LINKED ISSUE #6424: [BUG]: Error in Windows version when trying to load state saved in Linux version
### Describe the Bug

In Windows version of PCSX2, when trying to load save state that was saved in the Linux version (the same version number), following error occurs:

image

It works the other way around. Linux version does not have any problems loading state saved in Windows.

### Reproduction Steps

1. Start a game in Linux version of PCSX2.
2. Save state.
3. Copy the contents of savestate folder to Windows version of PCSX2.
4. Start a game in Windows version of PCSX2.
5. Load state saved in step 2.
6. The error occurs.

### Expected Behavior

_No response_

### PCSX2 Revision

v1.7.2948

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

kmicki:
### Description of Changes
Padding of 8 bytes was added to the Linux's PAD's freeze data structure so the size matches the Windows' PAD's freeze data structure's size.

### Rationale behind Changes
Fixes #6424.

When PAD's freeze (state) data has lower size than expected, the state loading will abort with an error, even though pad data is not essential and the load method checks if data is incompatible and skips loading if it is.

Because of that, states created in Linux were not loadable in Windows. But it was possible the other way around.

Matching the sizes between Linux and Windows prevents aborting and states are properly loaded without PAD data when switching between platforms.

Unfortunately, the states saved in older version of PCSX2 on Linux **will not be loadable after this fix**.

See #6424.

### Suggested Testing Steps
Save state in Linux and load in Windows and vice versa.

### DIFF
--- a/pcsx2/PAD/Linux/state_management.h
+++ b/pcsx2/PAD/Linux/state_management.h
@@ -120,6 +120,9 @@ struct PadFullFreezeData
 	u8 slot[2];
 	PadFreezeData padData[2][4];
 	QueryInfo query;
+	// unused padding data, present only so that data size
+	// matches Windows version for cross-platform save/load
+	char padding[8];
 };
 
 extern QueryInfo query;


RedDevilus:
Probably need to bump the savestate version.

stenzek:
IMO not worth bumping the save state version just for wx/linux. Just let it fail to load.

kmicki:
> IMO not worth bumping the save state version just for wx/linux. Just let it fail to load.

That's what I thought. If not for the initial size check (which is global to all the state components so probably can't be disabled just for PAD) the old state would still load properly so I wouldn't change the version.

PR #6426, opened 2022-06-16
Input Recording: Qt cleanup and refactor

xTVaser:
### Description of Changes

This cleans up the initial Qt implementation of input recording, simplifying things where possible, adding back features, and fixing bugs that were not fixed in the initial code migration.

I did a fair amount of testing (documented below) but there may definitely still be some slight issues / hard to catch things in terms of recording accuracy.  Similar to the last PR, this is the not the last and there is still a little bit more to port over so I'm sure those problems will be found and squashed as the process continues.

### Rationale behind Changes

Continuing to work on achieving parity with wx.

### Suggested Testing Steps

I went back through some of our old PRs and tested with the same conditions / modified them if they no longer made sense:

PRs I checked:
- https://github.com/PCSX2/pcsx2/pull/3845#pullrequestreview-559705772
- https://github.com/PCSX2/pcsx2/pull/4047#pullrequestreview-559678109
- https://github.com/PCSX2/pcsx2/pull/3690#pullrequestreview-491043980

<details>
<summary>Results:</summary>

- [x] Browsing for the recording's file-path was required, even manually deleting the field properly updated the UI
  - Just forbid manual editing of the field
- [x] Was only able to make a power-on or "boot" recording with GS closed
  - Is this possible in the Qt PR - can you "load" an ISO and perform actions upon it basically?
    - It's not, so this is no longer relevant, the options should be disabled unless a game is open
- [x] New/Play options were disabled until i booted a game for the first time upon opening the emulator.
  - same as above, no longer relevant
- [x] Undo Count doesn't inc on initial load
- [x] Can record a savestate recording [after playing / recording another]
- [x] Can play a savestate recording [after playing / recording another]
- [x] Load a power-on recording after [playing / recording] a save-state recording
- [x] Record a power-on recording after [playing / recording another]
- [x] Frame advance works [outside of recording or playback / during playback / during recording]
- [x] Undo count is accurate
- [x] Pausing works while [outside of recording or playback / during playback / during recording]
- [x] Can load a savestate to jump [back / ahead] in the recording, no blank frames are recorded
- [x] Can load a savestate to jump [back / ahead] while playing back
- [x] No blank frames recorded at the end of the movie
- [x] Play till the end of the recording, pauses on the right frame, resuming will switch to record mode and record properly!
- [x] OSD is accurate and is only displayed when a recording is active
  - [x] Frame Counter
  - [x] Record Mode Status
  - [x] Undo counter
- [x] Loading a save-state recording with a broken save-state version is handled
- [x] Loading a save-state recording without a corresponding savestate is handled appropriately
- [x] Loading a save-state before the recording's first frame or after the final frame is handled appropriately
- [x] Playing a game normally works fine

</details>


pcsx2-qt/MainWindow.cpp
```diff
@@ -1273,14 +1278,13 @@ void MainWindow::onInputRecNewActionTriggered()
...
-#include "pcsx2/Recording/InputRecordingControls.h"
-
 void MainWindow::onInputRecPlayActionTriggered()
 {
 	const bool wasPaused = m_vm_paused;
-
 	if (!wasPaused)
-		g_InputRecordingControls.PauseImmediately();
+	{
+		VMManager::SetPaused(true);
```
stenzek, line 1286: You shouldn't be calling VMManager functions on the UI thread. Always pass through g_emu_thread.

pcsx2-qt/MainWindow.cpp
```diff
@@ -1291,30 +1295,37 @@ void MainWindow::onInputRecPlayActionTriggered()
 	{
 		fileNames = dialog.selectedFiles();
 	}
+	else
+	{
+		if (!wasPaused)
+		{
+			VMManager::SetPaused(false);
```
stenzek, line 1302: Ditto

pcsx2-qt/MainWindow.cpp
```diff
@@ -1291,30 +1295,37 @@ void MainWindow::onInputRecPlayActionTriggered()
...
-	}
 }
 
 void MainWindow::onInputRecStopActionTriggered()
 {
-	if (g_InputRecording.IsActive())
+	if (g_InputRecording.isActive())
 	{
-		g_InputRecording.Stop();
+		g_InputRecording.stop();
```
stenzek, line 1326: Should this execute on the EE/CPU thread too? (`Host::RunOnCPUThread()`)

pcsx2/CDVD/CDVD.cpp
```diff
@@ -839,7 +839,7 @@ void cdvdReset()
 
 	// If we are recording, always use the same RTC setting
 	// for games that use the RTC to seed their RNG -- this is very important to be the same everytime!
-#ifndef DISABLE_RECORDING
+#ifndef PCSX2_CORE
```
stenzek, line 842:
```suggestion
#ifndef PCSX2_CORE
const bool input_recording_active = g_InputRecording.IsActive();
#else
const bool input_recording_active = g_InputRecording.isActive();
#endif
if (input_recording_active)
```
Although IMO just making the naming consistent would be better (all the core stuff follows TitleCase, not camelCase).

pcsx2/Frontend/ImGuiManager.cpp
```diff
@@ -521,25 +522,48 @@ static void DrawPerformanceOverlay()
 	constexpr bool paused = false;
 #endif
 
+#ifdef PCSX2_CORE
+	// TODO - is there a way to forcibly update the imgui frame without requiring to wait until the next vsync?
```
stenzek, line 526: `GetMTGS().PresentCurrentFrame()`, but if you want to call that from anywhere other than the CPU thread, you'll need to wrap it in `EmuThread` Qt-side.

pcsx2/Recording/InputRecording.cpp
```diff
@@ -465,364 +465,342 @@ wxString InputRecording::resolveGameName()
...
+		{
+			InputRec::consoleLog(fmt::format("Could not locate savestate file at location - {}", savestatePath));
+			InputRec::log("Savestate load failed");
+			m_file.Close();
+			return false;
+		}
+		m_type = Type::FROM_SAVESTATE;
+		m_initialSavestateLoadComplete = false;
+		m_isActive = true;
+		const auto loaded = VMManager::LoadState(savestatePath.c_str());
```
stenzek, line 563: Some concerns about threading here, if this function executes on the CPU thread, it's fine, but otherwise you'll need to use `Host::RunOnCPUThread().` (mainly because of the control calls below)

pcsx2/Sio.cpp
```diff
@@ -210,7 +210,11 @@ SIO_WRITE sioWriteController(u8 data)
 			// Only examine controllers 1 / 2
 			if (sio.slot[sio.port] == 0 || sio.slot[sio.port] == 1)
 			{
+#ifdef PCSX2_CORE
```
stenzek, line 213: Ditto here re `TitleCase`

stenzek (changes requested):
Mostly seems okay from a quick glance, just some concerns about races/things being called from the wrong thread.

pcsx2/CDVD/CDVD.cpp
xTVaser: Yeah we probably need to actually codify a style guide, because that's contrary to what was proposed a while ago https://github.com/tadanokojin/pcsx2/blob/coding-guide/pcsx2/Docs/Coding_Guidelines.md#camel-case-for-function-names which is what i try to follow

stenzek: All the code outside of the Qt classes (because they, themselves, use camelCase) is TitleCase, perhaps the style guide should be adjusted to suit..

xTVaser: I believe the code was that way (TitleCase) at the time the style guide was originally drafted as well.  But yes either the style guide is adjusted to fit what the majority of the code base already is, or new code is incrementally written in the desired style.  

Something has to be committed to or there is no hope for consistency (if we care about it)

pcsx2-qt/MainWindow.cpp
xTVaser: What is the accepted pattern for getting a value out of the CPU thread to inform a UI action?  For example below this line:
```cpp
if (g_InputRecording.play(fileNames.first().toStdString()))
{
	m_ui.actionInputRecStop->setEnabled(true);
	return;
}
```
I assume I should also be running `play` on the CPU thread, but how am I supposed to get the bool result out to perform further UI actions?  I assume `RunOnCPUThread` is not blocking, as well as it currently returns `void` so doing something like the following wouldn't help anyway.

```cpp
Host::RunOnCPUThread([&]() {
	if (g_InputRecording.play(fileNames.first().toStdString()))
	{
		m_ui.actionInputRecStop->setEnabled(true);
		return;
	}
});
```

pcsx2/Frontend/ImGuiOverlays.cpp
```diff
@@ -452,10 +454,63 @@ void ImGuiManager::DrawInputsOverlay()
...
+	const float spacing = std::ceil(5.0f * scale);
+	float position_y = margin;
+
+	ImFont* const fixed_font = ImGuiManager::GetFixedFont();
+	ImFont* const standard_font = ImGuiManager::GetStandardFont();
+
+	ImDrawList* dl = ImGui::GetBackgroundDrawList();
+	std::string text;
+	ImVec2 text_size;
+	bool first = true;
```
lightningterror, line 472: Not used?

Mrlinkwii (approved):

PR #6435, opened 2022-06-16
DEV9: Add Per Game Host List Support

TheLastRar:
### Description of Changes
Adds the ability to use per game settings for the internal DNS host list

### Rationale behind Changes
More per-game settings

### Suggested Testing Steps
Test if the DNS host list works on a per game basis (when overridden)

### DIFF
--- a/pcsx2-qt/Settings/DEV9SettingsWidget.cpp
+++ b/pcsx2-qt/Settings/DEV9SettingsWidget.cpp
@@ -251,8 +251,7 @@ DEV9SettingsWidget::DEV9SettingsWidget(SettingsDialog* dialog, QWidget* parent)
 	connect(m_ui.ethHostExport, &QPushButton::clicked, this, &DEV9SettingsWidget::onEthHostExport);
 	connect(m_ui.ethHostImport, &QPushButton::clicked, this, &DEV9SettingsWidget::onEthHostImport);
 
-	if (m_dialog->isPerGameSettings())
-		m_ui.ethTabWidget->setTabEnabled(1, false);
+	connect(m_ui.ethHostPerGame, &QPushButton::clicked, this, &DEV9SettingsWidget::onEthHostPerGame);
 
 	//////////////////////////////////////////////////////////////////////////
 	// HDD Settings
@@ -495,7 +494,7 @@ void DEV9SettingsWidget::onEthHostDel()
 
 void DEV9SettingsWidget::onEthHostExport()
 {
-	std::vector<HostEntryUi> hosts = ListHostsConfig();
+	std::vector<HostEntryUi> hosts = ListHostsConfig().value();
 
 	DEV9DnsHostDialog exportDialog(hosts, this);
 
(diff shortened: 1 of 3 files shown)


pcsx2-qt/Settings/DEV9SettingsWidget.cpp
```diff
@@ -493,7 +492,7 @@ void DEV9SettingsWidget::onEthHostDel()
 
 void DEV9SettingsWidget::onEthHostExport()
 {
-	std::vector<HostEntryUi> hosts = ListHostsConfig();
+	std::vector<HostEntryUi> hosts = ListHostsConfig().value();
```
stenzek, line 495: Is `ListsHostsConfig()` ever going to return nullopt here? If so, you should check that it has a value.

TheLastRar: The Export button is only accessible in the global settings panel, or the per-game settings panel when a per game host list is present, I `ListHostsConfig()` should never return nullopt in those cases.

I can however add a check regardless, if preferred

stenzek: Nah, it's fine, if it's never going to be nullopt there's no problem, was just concerned about it blowing up at runtime on some crazy config :)

TheLastRar: I did mean to clarify the logic, but I haven't done so until now

`ListHostsConfig()` only returns nullopt when it's call to `m_dialog->getIntValue("DEV9/Eth/Hosts", "Count", std::nullopt);` returns nullopt.
I can see that when getIntValue is not used with per game settings, std::nullopt gets treated as zero instead, so returning nullopt would only happen when `[DEV9/Eth/Hosts] Count` isn't found in the per game settings file*
In such a case, `RefreshHostList()` (called in the constructor) would disable the host Add/Delete/Export & Import buttons, which, as `onEthHostExport` is only called via a button click, effectively ensuring that code only running with valid settings

I’ll also add that `RefreshHostList()` assumes this behaviour and only checks for nullopt when working with per-game settings

In response to your comment, I’ve also briefly tested;
A clean Install, of which PCSX2 auto created the `Count` entry, which worked fine
Manually deleting the `Count` entry, which also worked fine, treating the missing entry as zero.

kokroo:
> ### Description of Changes
> Adds the ability to use per game settings for the internal DNS host list
> 
> More per-game settings
> 
> ### Suggested Testing Steps
> Test if the DNS host list works on a per game basis (when overridden)

Do you want me to test this on macOS?

refractionpcsx2:
What's the status of this? do we know?

pcsx2-qt/Settings/DEV9SettingsWidget.cpp
TheLastRar: I've since adjusted the logic so that `ListHostsConfig()` is no longer dependent on `m_dialog->getIntValue()` returning 0 for a default value of `std::nullopt` when not used with per game settings

TheLastRar:
I've rebased this and did some brief testing myself to make sure things worked.

I did a quick round of testing to make sure things worked

Mrlinkwii (approved):

JordanTheToaster (approved):

kokroo:
> I've rebased this and did some brief testing myself to make sure things worked.
> 
> I did a quick round of testing to make sure things worked

Did you test it with any games?

TheLastRar:
RE; Outbreak, although I tested it with the fixes included in #7270, without it you may have to manually specify the internal DNS (as 192.0.2.1) to actually use the Internal DNS

Issue #6437, opened 2022-06-17
[BUG]: Weird error loading save state


dezraj:
### Describe the Bug

Hello! I got the following error while I was playing KOF Nests Hen Collection and tried to load a state:

![image](https://user-images.githubusercontent.com/27839354/174227431-194b4f98-164b-4897-ae35-1e8077b1cfbd.png)

### Reproduction Steps

Just hit F3 key to load the saved state and couldn't.

I don't know if it matters but I converted my ISOs to CHD to save space. Aren't save states supported with CHD files?

### Expected Behavior

It was supposed to load normally I guess.

Thank you for your time and work.

### PCSX2 Revision

v.1.7.2950

### Operating System

Windows 10 (64bit)

stenzek:
Invalid, save states got bumped a few versions back. The error explains exactly what's going on.

dezraj:
Hello! I noticed that for some reason the state wasn't even created, that's why it wasn't loading and showing this message. Maybe I didn't hit the key with enough pressure?

Anyway, I'm sorry for the misunderstanding. Next time I'll be sure to check whether a save state file was created.

Thank you for the fast reply.

stenzek:
If there wasn't a state in that slot, we probably should display a better error message.

dezraj:
> 

Yes, that was indeed the case. I just deleted my save state file and tried to load it to confirm that the same error message shows.

So a better error message would have probably helped me better to understand what was happening in my case (no save file found). Opening this issue wasn't so useless in the end.

Thank you again for your hard work. 🙏

PR #6438, opened 2022-06-17
VMManager: Don't try to load state when none exists

stenzek:
See title. Makes the scenario mentioned in #6437 less confusing.

### DIFF
--- a/pcsx2/VMManager.cpp
+++ b/pcsx2/VMManager.cpp
@@ -1654,6 +1654,23 @@ static void HotkeyCycleSaveSlot(s32 delta)
 	}
 }
 
+static void HotkeyLoadStateSlot(s32 slot)
+{
+	if (s_game_crc == 0)
+	{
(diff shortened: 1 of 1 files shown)


pcsx2/VMManager.cpp
```diff
@@ -1654,6 +1654,23 @@ static void HotkeyCycleSaveSlot(s32 delta)
 	}
 }
 
+static void HotkeyLoadStateSlot(s32 slot)
+{
+	if (s_game_crc == 0)
```
lightningterror, line 1659: Won't this prevent save/load in the bios?

refractionpcsx2: why would you need to load a state in the bios?

However a more applicable problem related to this, PS1 games are functionally registered as running the BIOS, so you won't be able to save states in PS1 games.

stenzek: It won't save anyway currently (the file generator returns an empty string).

You can use the manual save to file (in the menu, pick a filename) if you really want to.

JordanTheToaster (approved):

PR #6449, opened 2022-06-18
Updater: Do not extract portable.ini

CharlesThobe:
### Description of Changes
Makes the updater skip "portable.ini" while extracting the update.

### Rationale behind Changes
Users who delete portable.txt definitely don't want their pcsx2 to be "portable".

### Suggested Testing Steps
make sure the updater works as it's supposed to.

### DIFF
--- a/updater/Updater.cpp
+++ b/updater/Updater.cpp
@@ -238,7 +238,9 @@ bool Updater::ParseZip()
 		if (!entry.destination_filename.empty() && entry.destination_filename.back() != FS_OSPATH_SEPARATOR_CHARACTER)
 		{
 			// skip updater itself, since it was already pre-extracted.
-			if (StringUtil::Strcasecmp(entry.destination_filename.c_str(), "updater.exe") != 0)
+			// also skips portable.ini to not mess with future non-portable installs.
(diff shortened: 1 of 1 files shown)


CharlesThobe:
btw it doesn't work, yet

CharlesThobe:
Back in business.

refractionpcsx2:
Why are you skipping updating the updater? what if we update it?

CharlesThobe:
I'm not, that's the original behavior from stenzek.
in the update dialogue from pcsx2-qt it extracts updater.exe before launching it:
function call: https://github.com/PCSX2/pcsx2/blob/d1cdfafe225f636c92f8473b65ed747f4cbdc4ca/pcsx2-qt/AutoUpdaterDialog.cpp#L493

definition: https://github.com/PCSX2/pcsx2/blob/fbf9585b5a8746d0f8abb0fa43a27a331cb84cb1/updater/UpdaterExtractor.h#L41

CharlesThobe:
In short, I only added a check for "not" portable.ini, the check for "not" updater.exe is there already in master.

stenzek (approved):
lgtm

Issue #6447, opened 2022-06-18
[BUG]: Crashes When Changing Internal Resolutions From 8x to 6x while Sync To Host Is Enabled


mrchrismania:
### Describe the Bug

The Emulator itself even when theres no game booted up it Crashes When Changing Internal Resolutions From 8x to 6x while Sync To Host Is Enabled its a bug with the Application itself

### Reproduction Steps

General Settings >> GS Window If I Have Sync To Host Ticked

Then i go to Graphics Settings >> If I then Switch to 8x Native Resolution Then Switch it to 6x Internal Resolution It causes the Emulator to crash. i have spent a while testing this and it only happens when Sync to host refresh rate is ticked inside the GS Window

I was able to reproduce this error on both my Desktop PC and Laptop

### Expected Behavior

was trying to play with sync to host enabled and i needed to change the resolution from 8x down to 6x when Playing GTA Vice City Stories which caused it to crash since i had the sync to host setting enabled.

### PCSX2 Revision

v1.7.2954

### Operating System

Windows 10 (64bit)

stenzek:
No crash for me, even in VCS when changing resolutions.

Can't really do much unless you provide the crash dump (use Qt).

mrchrismania:
> No crash for me, even in VCS when changing resolutions.
> 
> Can't really do much unless you provide the crash dump (use Qt).

did you also have wait for vsync on refresh on standard and Sync To Host Refresh Rate Ticked?

Heres the crash dump i had a few of them in my crash dump folder.

[Crash Dump.zip](https://github.com/PCSX2/pcsx2/files/8933805/Crash.Dump.zip)

stenzek:
That's not a Qt minidump.

stenzek:
Closing since https://github.com/PCSX2/pcsx2/commit/c86e82039a0a4f4063a87d60455b37f328474376 was merged, this feature no longer exists in wx. If you're still getting crashes, disable it in the ini file, or switch to Qt.

PR #6454, opened 2022-06-19
Wx: Remove sync to host refresh rate option

stenzek:
### Description of Changes

Still available in Qt obviously. Apparently there's COM reinitialization issues on the CPU thread in wx, and I don't want to deal with that nonsense.

### Rationale behind Changes

Possible fix for https://github.com/PCSX2/pcsx2/issues/6447.

### Suggested Testing Steps

Make sure it builds.


F0bes (approved):
With a fresh ini, I could no longer reproduce the crash #6447 was probably running into.

PR #6461, opened 2022-06-19
Gamedb: DB Updates

Florin9doi:

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -241,6 +241,24 @@ PBPX-95224:
 PBPX-95228:
   name: "DVD Player Version 3.00"
   region: "NTSC-J"
+PBPX-95239:
+  name: "Online Start-Up Disc v3.0"
+  region: "NTSC-U"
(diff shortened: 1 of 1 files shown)


Mrlinkwii (approved):

bin/resources/GameIndex.yaml
```diff
@@ -3036,7 +3054,7 @@ SCES-51533:
     roundSprite: 1 # Fixes vertical lines and minor ghosting.
     wrapGSMem: 1 # Fixes FMV when in progressive mode.
 SCES-51578:
-  name: "Network Access Disc [Original, v4.02 & v4.03]"
+  name: "Network Access Disc [v1.03 - v6.00]"
```
refractionpcsx2, line 3057: Is this just all versions then? if so it's probably not worth mentioning the version at all, since it's kinda moot.

Florin9doi: There are two other serials, one for v0.02 and one for v1.02 : https://wiki.pcsx2.net/Network_Access_Disc

PR #6486, opened 2022-06-21
GUI: Only add extension to new memcards on save

refractionpcsx2:
### Description of Changes
Changes the memcard screen to only add the extension to new memcards on save.

### Rationale behind Changes
Because overengineering it to automatically take off the old extension and put it back on constantly not only causes problems (because it didn't work right) there's also no need, better just to add the extension when you save, and deny the use of decimal points.

### Suggested Testing Steps
Make some memcards, try different names, I'm sure something will break it and I won't care :)

### DIFF
--- a/pcsx2-qt/Settings/CreateMemoryCardDialog.cpp
+++ b/pcsx2-qt/Settings/CreateMemoryCardDialog.cpp
@@ -60,20 +60,16 @@ CreateMemoryCardDialog::~CreateMemoryCardDialog() = default;
 
 void CreateMemoryCardDialog::nameTextChanged()
 {
-	const QString controlName(m_ui.name->text());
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
I'm not sure how I'd do that, however I've changed it so if you try to put an extension (or specifically a decimal point) it won't let you, it'll remove any decimals.

IMO having to deal with extensions is more confusing for the user, unless of course they spent a lot of time making memory cards in PCSX2, like yourself :P

RedDevilus:
On Qt, PS1 memcards make .ps2 instead of .mcr

refractionpcsx2:
That's what this fixes?

image

RedDevilus:
Yeah, that's what I meant that it fixed that behavior.

PR #6498, opened 2022-06-23
GS: Add internal FFMD offset to interlace.

refractionpcsx2:
### Description of Changes
Account for the internal offset on the PCRTC circuit when interlacing, not just the interlace offset.

### Rationale behind Changes
We weren't accounting for this before, and it makes stuff look clearer (see below)

### Suggested Testing Steps
Test games, compare to master, make sure they don't look blurrier. Also test games which looked a little blurry still when upscaled.

(Note the cliffs and the clarity of the UI)

Raging Master

Raging PR

(Look at the languages, especially the G in English, also the stepping on the letters in "Dark Alliance")

BG2 Master

BG2 PR

### DIFF
--- a/pcsx2/GS/Renderers/Common/GSDevice.cpp
+++ b/pcsx2/GS/Renderers/Common/GSDevice.cpp
@@ -378,9 +378,9 @@ void GSDevice::Interlace(const GSVector2i& ds, int field, int mode, float yoffse
 	if (mode == 0 || mode == 2) // weave or blend
 	{
 		// weave first
-		const int offset = static_cast<int>(yoffset) * field;
+		const float offset = -yoffset * static_cast<float>(field);
 
-		DoInterlace(m_merge, m_weavebob, field, false, GSConfig.DisableInterlaceOffset ? 0 : offset);
+		DoInterlace(m_merge, m_weavebob, field, false, GSConfig.DisableInterlaceOffset ? 0.0f : offset);
 
 		if (mode == 2)
 		{
--- a/pcsx2/GS/Renderers/Common/GSDevice.cpp
+++ b/pcsx2/GS/Renderers/Common/GSDevice.cpp
@@ -378,7 +378,7 @@ void GSDevice::Interlace(const GSVector2i& ds, int field, int mode, float yoffse
 	if (mode == 0 || mode == 2) // weave or blend
 	{
 		// weave first
-		const float offset = -yoffset * static_cast<float>(field);
+		const float offset = yoffset * static_cast<float>(field);
 
 		DoInterlace(m_merge, m_weavebob, field, false, GSConfig.DisableInterlaceOffset ? 0.0f : offset);
 
--- a/pcsx2/GS/Renderers/Common/GSRenderer.cpp
+++ b/pcsx2/GS/Renderers/Common/GSRenderer.cpp
@@ -349,9 +349,15 @@ bool GSRenderer::Merge(int field)
 
 		g_gs_device->Merge(tex, src_gs_read, dst, fs, m_regs->PMODE, m_regs->EXTBUF, c);
 
-		// Use offset for bob deinterlacing, for normal deinterlacing we add an extra 1 since it's clearer, with bob that causes extra shake.
+		// Use offset for bob deinterlacing always, extra offset added later for FFMD mode.
 		const bool is_bob = GSConfig.InterlaceMode == GSInterlaceMode::BobTFF || GSConfig.InterlaceMode == GSInterlaceMode::BobBFF;
-		const float offset = (!m_regs->SMODE2.FFMD && !is_bob) ? 0 : ((tex[1] ? tex[1]->GetScale().y : tex[0]->GetScale().y));
+		float offset = (!m_regs->SMODE2.FFMD && !is_bob) ? 0 : ((tex[1] ? tex[1]->GetScale().y : tex[0]->GetScale().y));
+
+		// Offset accounts for the upscaling interlace offset difference, but we need to also account for the internal framebuffer movement in Frame mode.
+		if (m_regs->SMODE2.FFMD && !is_bob)
+		{
+			offset += 1.0f;
+		}
 
 		if (isReallyInterlaced() && GSConfig.InterlaceMode != GSInterlaceMode::Off)
 		{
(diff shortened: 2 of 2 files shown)


refractionpcsx2:
I did try other stuff that was improved like Jak and Armored Core Last Raven, and to be honest, didn't really change a lot, just seemed to move down a tiny bit, that was about it lol

felixthecat1970:
testing some games have "no interleacing" patches seems not work anymore, that because many this patches relies in hack / patch / remove the sceGsSetHalfoffset function, now taking account the PCRTC circuit for interleaced signal this patches will not work anymore (i guess), but doing some tests past month i find enabling 480p mode, unlocking it or reworked the SDTV / progressive video function in games, the interleacing is disabled, also look less blurry (same effect using antiblur option tested some games) and this kind codes works this PR too, still need to more code "workaround or how to access internal PCRT circuit" because videos for progressive modes in reworked / call sdtv function, games crash in sceVsync function after [PR#6256](https://github.com/PCSX2/pcsx2/pull/6256) ) so before people start flowing the issues section this is the cause. " maybe " the solution or the way to go for no interleacing is (still testing) start looking, making, rework functions 480p mode codes, or using native SDTV progressive modes in games .

refractionpcsx2:
Ugh I really dislike no interlace patches, they're such a bodge.....  I mean it shouldn't matter, because as long as your interlacing mode is set to automatic, my changes have no effect (PCSX2 will automatically set the interlacing mode to "none", rendering my changes moot).

So if that's the case and it's still broken, I need examples of games which broke with it.

refractionpcsx2:
okay well, I've disabled this new code if deinterlacing is disabled (usual when using no interlacing patches), as that was intefering.

Should be fine in 1.7.3005

felixthecat1970:
i think too "no interleacing" hacks based in sceGshalfoffsett function are very hacky, and looks like this patches codes was working because error you fixed (no interleacing codes based on sceGshalfoffsett  was taking avantage a error) correct me is my deduction is wrong, also this fix led me guest why this patches not work in real hardware too, also i was testing the antiblur option + deinterleacing automatic seems to work good same image as many "no interleacing codes" so i think this is the way to go and remove the "no interleacing codes" from main PCSX2; 

based in some test:
--street figther alpha anthology (USA) : enabling native progressive mode
--using default deinterleacing automatic + antiblur options default NTSC mode "same image quality as progressive mode"

--soul calibur III (PAL): enabling native progressive mode with custom code
--using antiblur enabled with this code improves the image clearity (looks better as only progressive)

games with no intrleacing codes
-- marvel vs capcom 2 no interleacing codes + deinterleacing to : none
-- marvel vs capcom 2 automatic deinterlacing + antiblur looks same as above

and other bunch games has the same behaviur looks same with no interleacing codes or better when antiblur options is used and automatic deinterleacing

maybe this is not notice this because deinterleacing option is default "automatic" and using "no interleacing codes rely in sceGshalfoffset hack" do nothing (placebo effect) other codes rely in sharp front buffer or sdtv modes are unnafected this PR

refractionpcsx2:
Yes there is no way in hell no interlacing patches would work on the real hardware in like 95% of cases, it's all bodged to work on PCSX2 (due to how we detected stuff), and now I've had to add a bunch of conditions to disable the handling like the PS2 would do to accommodate these patches.  From somebody who likes accurate emulation, I find this very frustrating.

Anyway, as said above, the regression from this PR will be fixed in 3005.

felixthecat1970:
i think is this PR not a regresion, because you fixes correct behaviur, (also no interleaced codes not rely in sceGsHalfoffsett hack works ok ) better i think should be consider to remove the "no interleacing patches"  too or rename other thing example "sdtv mode patches o image patches etc" because many people uses this patches with interleacing automatic giving a "placebo effect" what do you think if i make a pull request for remove o change "no interleacing" patches options to better correct codes instead add more hacky things as you mentioned?

Dreadmoth:
I'm getting mixed results with the Snowblind engine games (PAL, default settings except 4x native, no patches):

Dark Alliance, Champions of Norrath, Champions - Return to Arms: 
3D scenes were clearer before these changes - there is now a vertical double image effect.
UI is clearer or less aliased on some screens, others are blurrier or more aliased.

Dark Alliance 2: 
3D scenes were clearer before these changes - there is now a vertical double image effect.
UI is significantly clearer after changes (matches your example screenshot).

refractionpcsx2:
Can you show me examples of the 3D scenes?

refractionpcsx2:
The 3D looks okay to me at x4

image

refractionpcsx2:
Okay here's from an older version 2991

image

I will agree that the old version looks marginally clearer on the 3D, but the affect on the UI is huge.

I could probably get the 3D clear but then that would make other games blurry.  

Can you also check different upscales, I suspect x3, x4 and x5 will have varying levels of clarity in the old version

Dreadmoth:
> Can you show me examples of the 3D scenes?

Best viewed full-size, most noticeable in the centre of the screen (stripes on the player character).

2996
Champions RTA, pcsx2-v1.7.2996

3000
Champions RTA, pcsx2-v1.7.3000

At higher upscaling multipliers the amount of image separation is the same on-screen, but the double image effect is more noticeable (two overlapped crisp images instead of two softer ones that blur together somewhat).

>Edit: so I did some testing, and due to how the game is designed, you can't have both. Either the 3D is slightly blurry (which is quite difficult to see unless you're comparing side by side) or the UI/Text looks messy.

For Dark Alliance 2 I agree, the UI looks significantly better and the trade-off is probably worth it to most users.
The other games I mentioned don't seem to benefit from the changes, though (most of their UI clarity issues were fixed by #6106).

felixthecat1970:
my bad i was hopping do a pull request for the remove "no interleacing" codes resources and options because new antiblur work fine many games i tested but looking now games rely in no interleacing codes which adresses the sceHalfoffset function, they start looking blurry now in movement, i agree emulation accuracy need to prevail  over hacks and this games can work again less blurry finding - using  SDTV progressive modes, i have been testing the SDTV modes and many games i test last weeks disables natively the inteleacing blurry image leading better output in emulator or consoles, still i will do the pull request and reasoning for no depend in this hacks and lead to find more improved codes image (thinking can go in CHEATS section) withow developers change ps2 emulation behavior.

Tested games are blurry now with withow no intrleacing codes:
Marvel vs capcom 2 (movement blurry)
Capcom vs snk (movement blurry)
project zero (dont matter this game is blurry by default with-withow codes other codes hack blur function) 

Games have native or hacked codes to SDTV-480p or sharp backbuffer - front buffer
all tested and work fine this releace
tekken 5 (usa,japan, europe)
street figther anthology (USA)
assault suits valken (europe)

PR #6512, opened 2022-06-25
VU: YOLO T-Bit on MTVU

refractionpcsx2:
Maybe we can re-enable it in the future if some threading genius figures it out.

### Description of Changes
Disables VU1 busy flag on T-Bit handling when MTVU is enabled.

### Rationale behind Changes
More games than I expected just turn T-Bit on but never actually use it, which causes poor performance in MTVU due to threading.  If there is a game that uses it, we can disable MTVU in the GameDB, since it will break the performance of the game, even if we try to handle it correctly.

### Suggested Testing Steps
Test Tak - The Great Juju Challenge


refractionpcsx2:
I know I haven't removed the patches for Midnight Club 3, they are innocuous and if we do turn this back on, I don't wanna have to find out what they were again...

PR #6513, opened 2022-06-25
Misc: Replace old include guard with pragma once

### LINKED ISSUE #494: reserved identifier violation
I would like to point out that identifiers like "[`__CACHE_H__`](https://github.com/PCSX2/pcsx2/blob/dde94adec321c54d2b6bb875e06223b20c1d39e1/pcsx2/Cache.h#L16)" and "[`__SPR_H__`](https://github.com/PCSX2/pcsx2/blob/3271048ab6c73708dda7b6919f595ea6786fc200/pcsx2/SPR.h#L17)" [do not fit](https://www.securecoding.cert.org/confluence/display/cplusplus/DCL51-CPP.+Do+not+declare+or+define+a+reserved+identifier#DCL51-CPP.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28HeaderGuard%29) to the expected naming convention of the C++ language standard.
Would you like to adjust your selection for unique names?

F0bes:
### Description of Changes
Fixes #494
Only done for whatever is pulled in the QT solution, all other files that are only for WX are gonna be nuked at some point anyways.

### Rationale behind Changes
See #494

### Suggested Testing Steps
If it builds it should be fine.


elfring:
:thought_balloon: [The usage of the preprocessor directive “`pragma once`”](https://en.wikipedia.org/wiki/Pragma_once "Description") will trigger its own development concerns, won't it?

F0bes:
It's non-standard but supported on every platform and compiler we support. It also follows every other file in the code base.

PR #6519, opened 2022-06-26
VU: Don't clamp VF00 or I Reg

refractionpcsx2:
### Description of Changes
Elides clamping VF00 and the I register on every op that uses these registers. 

### Rationale behind Changes
VF00 never changes to a state which needs clamping and I Reg gets clamped when it's set (unless the I-Bit hack is on). Lead to extra Brrr.

### Suggested Testing Steps
Test especially on low end machines with MTVU disabled for maximum effect (of course test with too).

I found maximum impact was with MTVU off and VU Clamping set to Extra + Sign, shows about a 2-3% performance increase, but with normal settings on a low end machine would also be good.

Observed changes:

R&C with normal clamping (with MTVU):
before: 150-154fps
after: 152-156fps

MGS3 (without MTVU)

Before
None: 144-146
Normal: 141-143
Extra+Sign: 129-131

After
None: 145-147
Normal: 142-144
Extra+Sign: 134-136

Tekken Tag (without MTVU):

Before: 
Normal: 125
Extra + Sign: 115

After:
Normal: 127
Extra + Sign: 10fps

### DIFF
--- a/pcsx2/x86/microVU_IR.h
+++ b/pcsx2/x86/microVU_IR.h
@@ -220,6 +220,7 @@ struct microMapXMM
 	int  xyzw;     // xyzw to write back (0 = Don't write back anything AND cached vfReg has all vectors valid)
 	int  count;    // Count of when last used
 	bool isNeeded; // Is needed for current instruction
+	bool isZero;   // Register was loaded from VF00 and doesn't need clamping
 };
 
 class microRegAlloc
@@ -330,6 +331,14 @@ public:
 		}
 	}
 
+	bool checkVFClamp(int regId)
+	{
+		if ((xmmMap[regId].VFreg == 33 && !EmuConfig.Gamefixes.IbitHack) || xmmMap[regId].isZero)
+			return false;
+		else
+			return true;
+	}
+
 	bool checkCachedReg(int regId)
 	{
 		if (regId < xmmTotal)
@@ -346,6 +355,7 @@ public:
 		clear.count    =  0;
 		clear.xyzw     =  0;
 		clear.isNeeded =  0;
+		clear.isZero   =  0;
 	}
 
 	void clearRegVF(int VFreg)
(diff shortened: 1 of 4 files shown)


pcsx2/x86/microVU_IR.h
```diff
@@ -330,6 +330,16 @@ class microRegAlloc
 		}
 	}
 
+	bool checkVFClamp(int regId)
+	{
+		if(CHECK_SKIPMPEGHACK)
```
F0bes, line 335: Just curious, why skipmpeg?

F0bes (commented):
So what ends up happening to the I reg when the I-Bit hack is on now?

pcsx2/x86/microVU_IR.h
refractionpcsx2: I just wanted to pick a gamefix to hook it up to which was unlikely to affect 3d performance. No other reason :)

refractionpcsx2:
> So what ends up happening to the I reg when the I-Bit hack is on now?

If the I-Bit hack is disabled, it will only optimise on the VF00 clamps, it will clamp the I_REG again.

pcsx2/x86/microVU_IR.h
lightningterror, line 335: Formatting.

refractionpcsx2: Are you freaking serious?

lightningterror: Space after the if.

refractionpcsx2: It's temporary code... do you really think I'm leaving a skipmpeg link in the VU code? Use your brain LT

lightningterror: Okay, didn't know it was temporary, ignore the comment then.

PR #6520, opened 2022-06-26
iR5900: Elide VU0 micro finish calls when safe

stenzek:
### Description of Changes

This makes a difference in COP2-heavy games, where a chain of instructions will repeatedly test the VU0 idle bit unnecessarily, as it is impossible for a micro to be started inbetween the instruction chain.

Saves a bit of code size (for register backup/restore), as well as getting rid of branches. Seems to make a 1-2% difference in performance in Ratchet on a 3900X, but if we're lucky, more on slower chips.

### Rationale behind Changes

Brr maybe?

### Suggested Testing Steps

Test VU0 games, make sure nothing broke.

### DIFF
--- a/pcsx2/x86/iCore.h
+++ b/pcsx2/x86/iCore.h
@@ -220,6 +220,7 @@ int _signExtendXMMtoM(uptr to, x86SSERegType from, int candestroy); // returns t
 #define EEINST_COP2_STATUS_FLAG 0x400
 #define EEINST_COP2_MAC_FLAG 0x800
 #define EEINST_COP2_CLIP_FLAG 0x1000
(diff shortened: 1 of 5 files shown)


gysiguy:
Hello stenzek, this update breaks Ratchet & Clank: Up Your Arsenal for me.. Ratchet is in a perpetual state of falling/clipping through the ground and blinking, making the game unplayable. Every update afterwards is also broken, but v1.7.3013 works. I can provide more details and/or video on request. Thanks for your hard work, PCSX2 is awesome!!

refractionpcsx2:
This will be fixed with #6531

gysiguy:
Thanks! :)

PR #6527, opened 2022-06-27
Qt: Add pressure modifier button for pads

### LINKED ISSUE #6198: [Feature Request]: Inputs that allow "button pressure" to be specified in the new QT UI
### Description

The new input plugin for the QT version of PCSX2 should allow button inputs (specifically ones mapped to the face buttons) to have their button pressure specified.

### Reason

In the old version of PCSX2, you could use the LilyPad input plugin to specify how much pressure a button input would have. This is very useful because it allows you to play games that use the pressure-sensitive buttons of the DualShock 2 controller even if you don't have that controller. Some very iconic games that use this feature include the various GTA games, Gran Turismo, Okami, MGS2 and MGS3. 

In the latest QT update, the input plugin has been replaced with seemingly no alternative, so it seems to be impossible to properly play some games with a standard modern controller.

### Examples

This image shows an example of how to set button pressure in the old LilyPad input plugin:
2022-05-25 23_29_13-New Issue · PCSX2_pcsx2 — Mozilla Firefox

### LINKED ISSUE #6233: [Feature Request]: Support for pressure sensitive buttons of the DualShock 2/3
### Description

I use a Dualshock 3 in combination with DsHidMini/BthPS3, and it is great for using the controller on PC with all its features. For some reason, though, PCSX2 no longer supports the anolog pressure sensitive input of these controllers.

### Reason

Currently, can't get the most from the games that make use of it.

### Examples

N/A

stenzek:
### Description of Changes

Adds a "pressure modifier" button, which can have a configurable amount that scales any button presses while it is held.

### Rationale behind Changes

Feature parity with wx.

Closes #6198.
Closes #6233.

### Suggested Testing Steps

Test pressure sensitive games, or use something like padtest (checked myself).

Choose desired pressure in controller settings, bind button/key to pressure, hold pressure button and push desired button.

### DIFF
--- a/pcsx2-qt/Settings/ControllerBindingWidget_DualShock2.ui
+++ b/pcsx2-qt/Settings/ControllerBindingWidget_DualShock2.ui
@@ -25,7 +25,7 @@
   <property name="windowTitle">
    <string>Form</string>
   </property>
-  <layout class="QHBoxLayout" name="horizontalLayout">
+  <layout class="QGridLayout" name="gridLayout_35">
    <property name="leftMargin">
     <number>0</number>
    </property>
@@ -38,7 +38,7 @@
    <property name="bottomMargin">
     <number>0</number>
    </property>
-   <item>
+   <item row="0" column="0" rowspan="3">
     <layout class="QVBoxLayout" name="verticalLayout_2">
      <item>
       <widget class="QGroupBox" name="groupBox">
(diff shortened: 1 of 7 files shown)


jonasos3:
Great work! One gripe with how it is implemented atm; it now functions as a toggle of sorts, and it requires being held in before the affected button is pushed.

Ex.: To switch from a run to a walk in MGS2 one has to stop, release the d-pad button, press and hold "pressure mod" and then hold the d-pad button again. It should perhaps change the pressure value only when actually being held, as opposed to how it is now?

nobleflame:
I’m not sure this is what was requested. It’s a nice idea and might work for some games, but in the WX version every button is individually configurable.

stenzek:
We decided this is a better solution. Just because it was done one way in wx doesn't mean it will be done the same going forward. Generally, you don't have a bunch of extra buttons to use as specific pressure for specific buttons, so a modifier makes more sense.

If you want a different setup, you're free to implement it yourself and open a pull request, but this is the way we chose to do things.

jonasos3:
> I’m not sure this is what was requested. It’s a nice idea and might work for some games, but in the WX version every button is individually configurable.

I agree with you that this was a very useful feature of Wx, and more customizable than any other emulator I've seen. For specific games and use-cases this was a perfect implementation. 

@stenzek If a modifier button is the chosen solution (which I understand, there are probably not many people taking advantage of all the old Wx/LilyPad features), would it be possible to make it work as I explained above? Where it only changes the pressure value _when held down_? This current implementation is appreciated of course, but it doesn't really help with a few specific uses where a "hold-to-activate" would. Sorry for bothering you with this, but for some of us it's a very critical feature. I wish I could PR it myself but I'm not a programmer.

PR #6532, opened 2022-06-28
GS: Don't vmalloc inside fifo_alloc to avoid logic confusion.

refractionpcsx2:
### Description of Changes
Changes the GS memory "fifo_alloc" code to not use vm_alloc if filemapping fails.
Also remove a pointless check from my previous COP2 fix.

### Rationale behind Changes
This causes a logic error, since `m_use_fifo_alloc` is set before entering this function, and can't get unset because it's expecting the pointer to be null in order to do so, and it wasn't. Two things trying to do the same job.

### Suggested Testing Steps
Windows only: load and shut down PCSX2 a bunch of times, make sure it doesn't crash. Keep an eye out for either `Failed to reserve memory. WIN API ERROR` or `Memory mapping failed after %d attempts, aborting. WIN API ERROR` in the log, it's instrumental to this happening.


stenzek (approved):
LGTM

JordanTheToaster (approved):

PR #6533, opened 2022-06-28
GS: Switch to placeholders for VM repeat mapping in Qt

stenzek:
### Description of Changes

An extension of #6532, this fixes the underlying issue that was causing the allocations to fail in the first place.

The software renderer needs a repeat mapping of the GS local memory for more efficient emulation of wrapping. To do this, we create an anonymous/pagefile backed file, and map the same file over 4 times the virtual address space. The problem is, on Windows prior to Win10 1809, you couldn't guarantee that any of that virtual address space was actually free.

So, PCSX2 just yolo'ed it, trying a few times to get a large enough chunk. But because we were only mapping 4MB, but needed 16MB (for the repeats), there's no guarantee that would actually end up in a free 16MB chunk. Hence "randomly" getting allocation errors, and falling back to no wrapping.

That's obviously bad. So we'll do things the modern way, which works similarly to Linux, which can mmap() over the top of existing mappings no dramas. The only downside, is that it requires Win10. But since Qt also requires Win10, and wx is deprecated already, this is fine.

### Rationale behind Changes

Making the software renderer more reliable.

### Suggested Testing Steps

Toggle software renderer a bunch of times and make sure you see no memory allocation errors. I also need to make sure wx still launches on Win8 despite the `_WIN32_WINNT` bump.

### DIFF
--- a/common/RedtapeWindows.h
+++ b/common/RedtapeWindows.h
@@ -24,8 +24,14 @@
 #define NOMINMAX
 #endif
 
-// Win8.1 is our minimum at the moment.
+// Qt build requires Windows 10+, WX Windows 8.1+.
+#ifndef _WIN32_WINNT
+#ifdef PCSX2_CORE
+#define _WIN32_WINNT 0x0A00 // Windows 10
+#else
 #define _WIN32_WINNT 0x0603 // Windows 8.1
+#endif
+#endif
 
 #include <windows.h>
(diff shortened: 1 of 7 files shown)


JordanTheToaster (approved):

stenzek:
Win8.1 works fine after adding a hack a downgrade XAudio (only for wx).

PR #6536, opened 2022-06-28
Qt: Implement more missing features

### LINKED ISSUE #6223: [BUG]: Qt: The "Remove Disc" button does not work.
### Describe the Bug

Launching any disc and attempting to remove said disc from the emulator does nothing.

### Reproduction Steps

Launch any disc, head to the toolbar, press the "Change Disc" button then "Remove Disc."

### Expected Behavior

For the disc to be removed.

### PCSX2 Revision

v1.7.2815

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

stenzek:
### Description of Changes

 - Implement disable window resize option.
 - Implement pause on start option.
 - Implement pause on focus option.
 - Implement start disc (drive selection/auto-default to single drive).
 - Adds the ability to hide the main window when a game is running, aka wx nogui mode.
 - Add global VM validity (useful for other windows).
 - Fixes bug where game properties windows would stay open after closing the main window.

### Rationale behind Changes

Implementing missing features.

Closes #6223.

### Suggested Testing Steps

Test all new features.

### DIFF
--- a/pcsx2-qt/DisplayWidget.cpp
+++ b/pcsx2-qt/DisplayWidget.cpp
@@ -397,22 +397,6 @@ bool DisplayWidget::event(QEvent* event)
 			return true;
 		}
 
-		case QEvent::FocusIn:
-		{
-			QWidget::event(event);
-			emit windowFocusEvent();
-			return true;
-		}
-
-		case QEvent::ActivationChange:
-		{
-			QWidget::event(event);
-			if (isActiveWindow())
-				emit windowFocusEvent();
-
-			return true;
-		}
-
 		default:
 			return QWidget::event(event);
 	}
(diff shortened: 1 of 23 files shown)


Daisouji:
The resizing and pausing working great. 
Just a suggestion for NoGUI, a CL argument would work well w/ it.

jonasos3:
A small issue with this build, not sure if intentional?; before, when closing a game by pressing the X in the window, the whole emulator would exit with that single click. Now, closing the game with the X brings back the main/startup window again.

When rendering the game to a separate window this would be expected, but the X in the main app/window should always close the entire program, right? If one wants to only close the game but retain the main window, that's what the Shut Down button on the toolbar is for.

Daisouji:
> A small issue with this build, not sure if intentional?; before, when closing a game by pressing the X in the window, the whole emulator would exit with that single click. Now, closing the game with the X brings back the main/startup window again.
> 
> When rendering the game to a separate window this would be expected, but the X in the main app/window should always close the entire program, right? If one wants to only close the game but retain the main window, that's what the Shut Down button on the toolbar is for.

1) If you use -batch command  w/ notification of exit off, exiting will yield closing the whole program down. 
2) If you are purely on QT,  and u use those these 2 NoGUI options => exiting will yield back to main window.

jonasos3:
@Daisouji Unfortunately this is a bit above my head.. :) I'm on Qt, and now I can't find any way to make X close the whole program when in-game. It always goes back to main window, even when NOT rendering in a separate one.

PR #6542, opened 2022-06-29
PAD: Adjust existing button pressure when modifier activated

stenzek:
### Description of Changes

Based on the comments in the other PR thread, this seems like a reasonable thing to do.

Also fixes some minor drag and drop issues.

### Rationale behind Changes

Why not.

### Suggested Testing Steps

Test pressure modifier again (it should update existing button pressure when pressed).

Test drag/drop GSdumps.


JordanTheToaster (approved):

jonasos3:
Works as expected, adjusts/releases (back to 100%) the pressure when held/let go. Thanks a lot! 👍

PR #6544, opened 2022-06-29
IOPBios: Make our hostfs HLE safer

F0bes:
### Description of Changes
Instead of letting `filesystem` possibly throw exceptions that crash the emulator, use the overloaded functions that instead take an std::error_code by reference. Logging was also added for when this happens as the IOP HLE error code system is currently meh¹ and is wrong anyways. We just pass EIO (Generic Input / Output Error).

Here is a sample of the logging
image

[1] What's the point of standardizing a filesystem API, only to make the error codes platform dependent? Thanks ISOCpp :/

### Rationale behind Changes
If you were to mkdir a path that is taken by an existing _file_, GHC would throw an exception and crash the emulator.

Try it here:
[hostfs.zip](https://github.com/PCSX2/pcsx2/files/9012446/hostfs.zip)

### Suggested Testing Steps
Use something that utilizes hostfs.

Thanks @fjtrujy for pointing this out ^‿^

### DIFF
--- a/pcsx2/IopBios.cpp
+++ b/pcsx2/IopBios.cpp
@@ -316,6 +316,7 @@ namespace R3000A
 			ghc::filesystem::directory_iterator dirent(path.c_str(), err);
 			if (err)
 				return -IOP_ENOENT; // Should return ENOTDIR if path is a file?
+                                    // Would love to but err's code is platform dependent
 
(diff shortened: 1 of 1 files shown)


fjtrujy:
Thanks for fixing it!

Mrlinkwii (approved):

PR #6551, opened 2022-06-30
3rdparty: Remove ghc::filesystem

stenzek:
### Description of Changes

I had this sitting on the branch which shall not be named and forgot to PR It.

It was only used in DEV9 and IopBios, so seems silly to keep it around. Plus the error code nonsense that fobes noted.

### Rationale behind Changes

Fewer dependencies.

### Suggested Testing Steps

Test DEV9 and IOP HLE functionality (mainly around creating/setting HDD images for the former, and host file access for the latter).


Mrlinkwii (approved):

F0bes (approved):
hostfs seems to work fine

stenzek:
Might want to re-test, somehow I cancelled the runs.

For future reference, I also added some really basic sandboxing to the file paths. Previously, you could do something like `host:..\..\..\windows\pwned.exe`, now that'll get denied. Hopefully nobody relied on that, but it is a security risk.

F0bes:
I built it locally.
Re: sandboxing. Yeah that's fine, it's better than before, even if it breaks compat with some peoples odd setups.

JordanTheToaster (approved):

AKuHAK:
It currently breaks the homebrew host: access:
```
dopen name host: 
IopHLE: Denying access to path outside of ELF directory. Requested path: '', Resolved path: 'D:\[censored]\', ELF directory: 'D:\[censored]\'
```

PR #6555, opened 2022-07-01
CI: Add AppImage builder for Qt

stenzek:
### Description of Changes

What the title says. Also adds SSE4/AVX2 builds for parity with Windows, as well as building with clang with LTO (measurable performance improvement).

Wayland is supported (unlike the linuxdeploy plugin), although apparently some drivers are busted.

Few open questions:
 - Do we need to bother with the libc hack? Personally, I don't think so, since Qt6 requires Ubuntu 20.04, which is what we build on now.
 - Updating.. zsyncs require a fixed url, so that could be tricky.
 - Do we want to drop the clang nopch build now?

### Rationale behind Changes

One step closer to getting rid of wx.

### Suggested Testing Steps

Test on a range of configurations and distributions. So far I've tried Ubuntu 20.04, 22.04, Fedora 36 and Arch without issues.


.github/workflows/scripts/linux/generate-cmake-qt.sh
```diff
@@ -0,0 +1,35 @@
...
+
+echo "Additional CMake Args - ${ADDITIONAL_CMAKE_ARGS}"
+
+# Generate CMake into ./build
+# DISABLE_ADVANCE_SIMD is needed otherwise we end up doing -march=native.
+
+cmake \
+	-B build \
+	-G Ninja \
+	$ADDITIONAL_CMAKE_ARGS \
```
stenzek, line 25: This warning (and the others which are left) are false positives. Not sure if we can mute it specifically so it'll have a sad cross :(

qurious-pixel: You can make codacy happy by appenging `"${ADDITIONAL_CMAKE_ARGS}"` to another value.

``` 
-DCMAKE_BUILD_TYPE=Release "${ADDITIONAL_CMAKE_ARGS}" \
```

That way it accepts the double quotes, and if the variable is empty, it will not fail.

tellowkrinkle: You broke nopch builds last time by doing exactly that

`$ADDITIONAL_CMAKE_ARGS` contains multiple arguments and we need them to be expanded as multiple arguments.  And since it's passed by the CI, we can't make an array.

.github/workflows/scripts/linux/appimage-qt.sh
```diff
@@ -0,0 +1,251 @@
...
+echo "Copying binary and resources..."
+cp -a "$BINDIR/$BINARY" "$BINDIR/resources" "$BINDIR/shaders" "$OUTDIR/usr/bin"
+
+# Don't need old wx locales.
+rm -fr "$OUTDIR/usr/bin/resources/locale"
+
+# Patch RPATH so the binary goes hunting for shared libraries in the AppDir instead of system.
+echo "Patching RPATH in ${BINARY}..."
+patchelf --set-rpath '$ORIGIN/../lib' "$OUTDIR/usr/bin/$BINARY"
+$STRIP "$OUTDIR/usr/bin/$BINARY"
```
tellowkrinkle, line 176: Leave the symbols in PCSX2 in please

Being able to debug the appimage, even if it's just function names, is appreciated

tellowkrinkle (commented):
Codacy is dumb.  I don't think we can turn those lints off without doing it globally, so you'll have to deal with the stupid red x

xTVaser (changes requested):
As the `linux_build.yml` workflow file has been renamed, the release workflow needs to be updated to point to the new file name or the next release won't have linux artifacts https://github.com/PCSX2/pcsx2/blob/master/.github/workflows/release_pipeline.yml

xTVaser:
If you believe the codacy errors are false positives, the underlying linter it is using in this case is ShellCheck and you can ignore them by consulting these docs https://github.com/koalaman/shellcheck/wiki/Directive

stenzek:
Should all be sorted now, including using `~/.config/PCSX2` for `XDG_STD` builds (same as wx appimage).

Doing a test release run on my fork to make sure that works as intended.

.github/workflows/scripts/linux/build-dependencies-qt.sh
```diff
@@ -0,0 +1,123 @@
...
+fce94688ea925782a2879347584991f854630daadba6c52aed6d93e33cd0b19c  qttools-everywhere-src-$QT.tar.xz
+e7b567f6e43ffc5918d4aa825ce1eced66a00cb0a87133b2912ba5c1b2a02190  qtwayland-everywhere-src-$QT.tar.xz
+EOF
+
+curl -L \
+	-O "https://libsdl.org/release/$SDL.tar.gz" \
+	-O "https://download.qt.io/official_releases/qt/${QT%.*}/$QT/submodules/qtbase-everywhere-src-$QT.tar.xz" \
+	-O "https://download.qt.io/official_releases/qt/${QT%.*}/$QT/submodules/qtsvg-everywhere-src-$QT.tar.xz" \
+	-O "https://download.qt.io/official_releases/qt/${QT%.*}/$QT/submodules/qttools-everywhere-src-$QT.tar.xz" \
+	-O "https://download.qt.io/official_releases/qt/${QT%.*}/$QT/submodules/qttranslations-everywhere-src-$QT.tar.xz" \
```
qurious-pixel, line 26: Not hash checking qttranslations. echo "Installing Qt Translations" is outside the "if" statement

stenzek: The ifs can probably go now tbh, I just had them in there when refactoring the script to make it run faster locally

Mrlinkwii (approved):

PR #6559, opened 2022-07-02
Config: Set texture preloading to full by default

JordanTheToaster:
### Description of Changes
Sets texture preloading to Full (Hash Cache) by default.

### Rationale behind Changes
More brrr,

### Suggested Testing Steps
Make sure CI Is happy.


stenzek (approved):
lgtm

coornio:
This was a long time coming, it has more benefits than downsides. If nothing else, we've had more issues with games not using any preloading whatsoever (like Castlevania :P)

DonelBueno:
The only game I know that is slower with this option on is Ridge Racer V. But it's like WAY slower, specially when there are lots of cars on the screen. And it happens with both the "partial" option and the "full" option.

PR #6562, opened 2022-07-03
Qt: Make status bar less confusing

stenzek:
### Description of Changes

Before:
image
After:
image

### Rationale behind Changes

Users get confused what G and V mean.

### Suggested Testing Steps

Check everything shows up as expected.


coornio:
Such a small change, but very welcome indeed. There was plenty of space down there anyway so perhaps you can also add extra info later on, like which savestate slot is currently selected and whatnot!

JordanTheToaster (approved):

PR #6568, opened 2022-07-04
Patches: Update Persona 4 WS Patch

JordanTheToaster:
### Description of Changes
Removes a portion of codes that forced Persona 4 to run at 60 fps. 

### Rationale behind Changes
Forcing Persona 4 to run at 60 fps causes the game to run at double speed making users think the game is broken.

### Suggested Testing Steps
Make sure CI is happy.


Tzerinas:
Regarding Persona WS Patches, do they still have that weird FMV fix on? It would get the FMV to fill up the image, but in turn cut off parts of it. Might be something to remove too.

Though I don't remember if it applies to P4, definitely applied to F3FES.

PR #6578, opened 2022-07-06
QT: more missing descriptions

Mrlinkwii:
### Suggested Testing Steps

make sure QT builds and wording is correct


Blackbird88:
What about moving **Disable Interlace Offset** tooltip to Qt? Right now it still uses the hover one
image

Mrlinkwii:
@Blackbird88  should be fixed now :)

PR #6553, opened 2022-06-30
WIP: GS: Add Pre-Round Sprite hack (Needs rewrite)

refractionpcsx2:
### Description of Changes
This attempts to preproduce hardware behaviour of texel coordinates (0.5 centre), but falls down in a bunch of cases, hence the hack. But does fix a couple of long standing bugs.

### Rationale behind Changes
In some cases, the PS2 will round down coordinates which PCSX2 does not, causing some issues with rendering. A prime example of this is Manhunt which draws a square over an area to hide lights, but uses a single pixel low alpha square in order to do so, and PCSX2 current misses this pixel by 1. This PR makes it pick the right pixel.

### Suggested Testing Steps
Try games that have graphical errors (in software) and see if this resolves them. I've already checked Mercinaries, and Gran Turismo 4 also gets broken by this.

Fixes #257
Resolves the water effect mentioned here #1986

This PR kind of fixes the UI on Dark Cloud but causes some misaligned backgrounds.

Showcase of the two main fixes of this PR:

Beyond Good and Evil:

BGE Before

BGE After

Manhunt:

Manhunt Before

Manhunt After

PR #6584, opened 2022-07-07
QT: set Texture Offsets & Skipdraw Range per game only

### LINKED ISSUE #6181: [Feature Request]: Texture Offsets & Skipdraw Range in Properties of the Game
### Description

In the "Properties" of the game, cannot put inputs in "Textures Offsets & Skipdraw Range".

### Reason

1. For the time being, only the global setting works. A lot of games need a "Texture Offsets" to fix some glitches.
2. I keep changing the global setting every time I launch another game that needs a texture offsets.

### Examples
Screenshot_19

Mrlinkwii:
### Rationale behind Changes

Fixes #6181

Making Texture Offsets & Skipdraw Range less dangerous 
### Suggested Testing Steps

make sure QT compiles and make sure Texture Offsets & Skipdraw Range are per game only  and they work

### DIFF
--- a/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
@@ -290,6 +290,14 @@ GraphicsSettingsWidget::GraphicsSettingsWidget(SettingsDialog* dialog, QWidget* 
 	// only allow disabling readbacks for per-game settings, it's too dangerous
 	m_ui.disableHardwareReadbacks->setEnabled(m_dialog->isPerGameSettings());
 
+	// allow Texture Offset for per-game settings only 
+	m_ui.textureOffsetX->setEnabled(m_dialog->isPerGameSettings());
+	m_ui.textureOffsetY->setEnabled(m_dialog->isPerGameSettings());
+
+	// allow Skipdraw Range for per-game settings only 
+	m_ui.skipDrawStart->setEnabled(m_dialog->isPerGameSettings());
+	m_ui.skipDrawEnd->setEnabled(m_dialog->isPerGameSettings());
+
 	// Display tab
 	{
 


JordanTheToaster (approved):

backgamon:
Now we lost the real time settings :(

mirh:
The user just asked for the thing to work in the game-specific properties. 
Then, to be sure they are two options that have little "general" sense... but ironically I find this location way easier for testing (be it just muscle memory, or even "not having every toggle with an extra state to cycle through").

refractionpcsx2:
Users are a danger to themselves, and setting it globally could break a lot of games, and cause extra headaches for us.

Making it per game makes a lot more sense, for the mild inconvenience of having to go to a different location when you're messing with it.

refractionpcsx2:
That sounds like a you problem ;)

mirh:
"Fearing lots of headaches because dumb users could tinker with setting globally" is not a me problem. 
The same token could be used for everything else too, from interlacing to CRC fixes or texture replacement. 
It doesn't make sense that you have to switch panels back and forth if you want to adjust everything (and again, not because I have an heavy ass, but because somehow the one where zoom controls are available is not the one where texture offsets work).

mirh:
> No, but going in to per game settings, screwing it up completely, rage quitting (apparently), then forgetting you completely screwed it up, is very much a user error, not our problem.

I'm not sure I understood the rage quitting part, anyhow you literally just described your grievance. 
Yes, this is something for the users to handle on their own - still you don't (understandably) want the most clueless ones to come nagging ***you*** for such stupidities. 

But "how much any given one setting could ever be specifically tailored to a game" (which fairly enough these two settings are the most of any other one) is not the same of "how much they are likely to cause problems" and thus be noticed (trivially, for instance it seems to me the resolution option carries a superset of all the nuisances that "texture offsets" could ever entail). 
And on top of this, what you care for here isn't just a function of "erring happening", but you also need the user to be duped far enough that they end up reporting it. 

Clearly this can happen as I myself suggested (even though I could eventually come to a close on my own) but even mistakes do have a logic behind. 
Which arguably is something on the lines of "forgetting about previous choices" or at least "not knowing how settings precedence work".

And what I'm saying is, if you let each "settings view" have different sets of available options, more people will potentially have to hop through both of them. And so you are actually *increasing* the likelihood of users getting the idiot ball. 

> You want it all on one panel? How on earth do you expect to fit all that in? or do you mean all in one place?

Ehrm.. "one setting window" then?
Not sure how I could be clearer. My example was that crop and zoom are greyed out in games' own properties.

mirh:
> That sounds like a you problem ;)
### ***[:)](https://github.com/PCSX2/pcsx2/pull/6553#issuecomment-1832763570)***

refractionpcsx2:
> > That sounds like a you problem ;)
> 
> ### _**[:)](https://github.com/PCSX2/pcsx2/pull/6553#issuecomment-1832763570)**_

What's your point?

mirh:
Because I had UX concerns (which I had spent quite the time to word and frame half decently), and you eventually shot them down as something just kinda airy on my side? And then it also happened to you (and yes, of course it's not a big deal, but the fact that it can be a thing at all is enough to make me wish for improvements). 

Though I guess it was bad taste to bump a PR then, my fault for not having the will to write a proper issue.

refractionpcsx2:
But what happened to me was user error, not a design error. What happened was exactly what *should* happen, your point was not proven here.

And having everything global is still a ridiculous idea, a lot of options are too dangerous to have globally.

Your issue will be denied if you make one on the previously mentioned issue.

PR #6587, opened 2022-07-08
GS: Improve CLUT testing, add test for local->local transfers.

refractionpcsx2:
### Description of Changes
Adds CLUT dirty checks to local->local transfers, and adjusts how it's handled elsewhere.

### Rationale behind Changes
We only flush the CLUT when we know the data has changed, for brr reasons, However if it was a local->local transfer we didn't check it. Also it was pointlessly flushing it if no drawing has happened, so stopped that. And make uploads check the block pointers to make sure they match also instead of just invalidating the CLUT always.

### Suggested Testing Steps
Test games known to have CLUT issues, I guess :P the sega games won't be fixed in HW, but make sure SW is ok.

Fixes Sagashi ni Ikouyo which does local->local transfers in to the CLUT so it needs flushing.

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -710,8 +710,7 @@ __inline void GSState::CheckFlushes()
 			Flush();
 		}
 	}
-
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Sagashi ni Ikouyo

Before:
image

After:
image

PR #6597, opened 2022-07-09
GameDB: Fix glows in Ratatouille

NicknineTheEagle:
### Description of Changes
Enable Auto Flush GS HW hack in Disney-Pixar Ratatouille.

### Rationale behind Changes
Fixes glows from lights showing up through walls in the game.

### Suggested Testing Steps
Bug can be easily reproduced in the very first level by looking at the sun with and without obstructions.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -4880,6 +4880,7 @@ SCKA-20093:
   gsHWFixes:
     halfPixelOffset: 2 # Reduces blooming misalignment.
     wildArmsHack: 1 # Reduces blooming misalignment.
+    autoFlush: 1 # Fixes glows.
 SCKA-20096:
   name: "Barnyard"
   region: "NTSC-K"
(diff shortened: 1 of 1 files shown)


lightningterror (commented):
Can you post a gsdump of the issue?

NicknineTheEagle:
> Can you post a gsdump of the issue?

Here you go.
[ratatouille_glows_bug.zip](https://github.com/PCSX2/pcsx2/files/9088178/ratatouille_glows_bug.zip)

lightningterror:
I wasn't able to reproduce the issue that requires auto flush.

NicknineTheEagle:
> I wasn't able to reproduce the issue that requires auto flush. Maybe the angle is not good when you took the dump.

That is strange, it's reproduced consistently on my end. Did you happen to change GS settings while the sun was obstructed? What happens is that glows are just not updated so if, for example, you switch from SW to HW while the sun is obstructed, the sun glow will permanent disappear instead.

lightningterror:
Does the issue also appear on sw renderer if you disable auto flush?

PR #5673, opened 2022-03-16
GameDB: Ace Combat + Soul Calibur + Metal Gear Solid Series

RedDevilus:
### Description of Changes

Align Sprite:

- Soul Calibur II / III
- Ace Combat Zero / 4 / 5

Mipmap

- Europese Ace Combat Variants

Half-pixel offsets for;

2 Special (Texture):

- Metal Gear Solid 2 (Substance) / 3 (Subsistence)

Roundsprite;

Full:

- Ace Combat Zero / 4 / 5

	

Missing mipmap from the Europese variants for Ace Combat Zero/4/5, Align Sprite for them and Round Sprite. Align Sprite for SoulCalibur series.
### Rationale behind Changes

Less tinkering.
### Suggested Testing Steps

You know the drill.

Issue #1860, opened 2017-03-14
Ace Combat 5 - See through sun in HW renderers


mirh:
[Gs dump](http://www61.zippyshare.com/v/Zn7vCK0h/file.html). 
If I first toggle SW renderer, it then gets fixed.

oldmud0:
In the future, I suggest attaching files via GitHub (just drop your files in the text box) instead of using sketchy file hosting sites.

mirh:
TOO BAD THAT EVERY DAMN TIME I TRY TO ATTACH \*ANYTHING\* IT TELLS ME CONTENT IS NOT SUPPORTED

...

*sob*

FlatOutPS2:
> TOO BAD THAT EVERY DAMN TIME I TRY TO ATTACH *ANYTHING* IT TELLS ME CONTENT IS NOT SUPPORTED

I had that for a while, but it's fine now. I could still upload zip files by renaming them to one of the other supported formats(like pdf).

oldmud0:
Here, I repackaged it into a zip and uploaded it on GitHub.
[AC5sun.zip](https://github.com/PCSX2/pcsx2/files/843189/AC5sun.zip)

lightningterror:
Ah I have that problem as well , that's why I use google drive when I post gs dumps on github.

mirh:
I'll also dupe here the dump from @jupiterbjy that misses "colored halo" around explosions in HW renderer. 
https://drive.google.com/file/d/0BxWgr9SSY0qtYXVGN3JoNUt3Njg/view

(perhaps to mirror on github.. *sighs*)

refractionpcsx2:
Can this be tried again with Auto Flush enabled on #5580 (or if it gets merged before you see this can you retry master)?

coornio:
Tested in 2415 and it's all fixed. I should note, however, that autoflush is not necessary for the sun. It works fine with the hack turned off.

![image](https://user-images.githubusercontent.com/11655762/156919869-6c233c7b-26a5-48a0-9260-cb7967e12c44.png)

mirh:
This is still bugged. 
Not to the same "like the plane wasn't even there" extent, but still.. some good amount of light goes through it. 
And if you switch between the sw CLUT render settings while you are occluding the sun, then rays stop to be a thing until you switch back and forth from the software renderer.

mirh:
[...]

Issue #1547, opened 2016-09-03
Ace Combat 04 (Distant Thunder/Shattered Skies) - Protagonist plane is black in GSdx HW


FlatOutPS2:
-PCSX2 version: 1.5.0-dev-1230 2016-09-03 12:17:59

-PCSX2 options: Defaults, no settings modifications affect the issue

-Plugins used: GSdx, SPU2-X, LilyPad, Gigaherz CDVD, the rest are null plugins

-Plugin settings: All renderers

-Description: The plane of the protagonist is completely black in all hardware renderers. All other planes aren't affected. OpenGL hardware depth, AD and BUA options don't affect the issue.
- Note: Running a build without Texture Cache mostly resolved the issue, but some parts of the plane remained black. So it's at least partially a TC issue.

-How to replicate: Boot the game and view the cutscene, or start a mission(it doesn't show up in hangar).

-Last known version to work: Unknown, probably never worked

-PC specifications: CPU: Intel Core i5 GPU: AMD Radeon HD 6000 OS: Windows 7 64-bit
- GS Dump:
  [AC4.zip](https://github.com/PCSX2/pcsx2/files/453549/AC4.zip)

m4ti140:
The issue seems to disappear partially when running the game with no-interlacing cheatcodes. There's still a black outline around the aircraft but the textures are visible. There's also no issue in hangar/cutscenes anymore.

mirh:
So, you are saying this patch improves it?
```
patch=1,EE,E0050003,extended,0029F418
patch=1,EE,0029F418,extended,00000001
patch=1,EE,D029F420,extended,0000948C
patch=1,EE,0029F420,extended,00000000
patch=1,EE,D029F420,extended,00009070
patch=1,EE,0029F420,extended,00000000
```

mirh:
Indeed, [confirmed](https://youtu.be/H-z4rt3TRGA?t=228) that no interlacing fixes this. 
Can somebody check it?

atomic83GitHub:
No, the black plane issue is TC related, messing around with settings does improve a bit the situation but never fixes it entirely, depending the mission the plane is either black with small colored parts or have collors but massively reduced in quality.

AlessandroVetere:
When upscaling, black vertical stripes appear:

![image](https://user-images.githubusercontent.com/15114320/63966147-ed81ec80-ca9a-11e9-9d27-9b204fb30170.png)

On GSdx debug mode this assertion is false:

[...]

PR #6600, opened 2022-07-09
GameDB: game title and Ace Combat updates

Azekthi:
### Description of Changes
1. Fixed some game titles (Ace Combat, Harry Potter, SpongeBob)
2. Added halfPixelOffset: 3 to Ace Combat 5 and Ace Combat Zero
3. Added mergeSprite: 1 to all three Ace Combat games

### Rationale behind Changes
1. Consistent and accurate gamedb
2. In foggy scenarios, the ground and planes can appear misaligned. Setting halfPixelOffset to 3 is required to fix this.
3. Align Sprite fixes the black lines when upscaling, but there are still "invisible" lines in their place. Merge Sprite fixes this in all three Ace Combat games. This can be seen in gameplay and in menu screens in all three games.

### Suggested Testing Steps
To test halfPixelOffset, play foggy scenarios and test all the options for halfPixelOffset. For AC5, I tested mission 27+ and compared the SOLG's textures. For AC Zero, I tested mission 07 and compared the ground textures.

To test Merge Sprite, simply enable or disable it during gameplay on any of them. For AC5, I did the same test with the SOLG's textures (the rotating arms are very helpful for this). For AC04, the main menu (after pressing start) has its background texture misaligned unless Merge Sprite is enabled.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -876,8 +876,10 @@ SCAJ-20104:
   region: "NTSC-Unk"
   gsHWFixes:
     mipmap: 1
+    halfPixelOffset: 3 # Fixes ghosting in foggy maps.
     roundSprite: 2 # Fixes font and HUD artifacts.
     alignSprite: 1 # Fixes vertical lines.
+    mergeSprite: 1 # Fixes vertical lines.
 SCAJ-20105:
   name: "Armored Core - Nine breaker"
   region: "NTSC-Unk"
@@ -1041,8 +1043,10 @@ SCAJ-20136:
   region: "NTSC-Unk"
   gsHWFixes:
     mipmap: 1
+    halfPixelOffset: 3 # Fixes ghosting in foggy maps.
     roundSprite: 2 # Fixes font and HUD artifacts.
     alignSprite: 1 # Fixes vertical lines.
+    mergeSprite: 1 # Fixes vertical lines.
 SCAJ-20137:
   name: "Musashiden II - Blademaster"
   region: "NTSC-Unk"
@@ -1229,6 +1233,7 @@ SCAJ-20173:
   region: "NTSC-Unk"
   gsHWFixes:
     mipmap: 1
+    halfPixelOffset: 3 # Fixes ghosting in foggy maps.
     roundSprite: 2 # Fixes HUD artifacts.
     alignSprite: 1 # Fixes vertical lines.
     mergeSprite: 1 # Better aligns main menu strips, improving font readability.
@@ -2684,6 +2689,7 @@ SCES-50410:
     mipmap: 1
     roundSprite: 2 # Fixes font and HUD artifacts.
     alignSprite: 1 # Fixes vertical lines.
+    mergeSprite: 1 # Fixes vertical lines.
 SCES-50411:
   name: "Vampire Night"
   region: "PAL-M5"
@@ -3318,8 +3324,10 @@ SCES-52424:
   compat: 5
   gsHWFixes:
     mipmap: 1
+    halfPixelOffset: 3 # Fixes ghosting in foggy maps.
     roundSprite: 2 # Fixes font and HUD artifacts.
     alignSprite: 1 # Fixes vertical lines.
+    mergeSprite: 1 # Fixes vertical lines.
   patches:
     1D54FEA9:
       content: |-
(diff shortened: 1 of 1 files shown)


mirh:
Explosions are a big issue with upscaling (and I would probably add a mention to it in the comments). 
I also [remember](https://old.reddit.com/r/emulation/comments/56wj30/mipmapping_makes_the_ace_combat_games_run_way/d8osrry/) autoflush used to be necessary there?

lightningterror (commented):
Is Align Sprite still needed then?

Azekthi:
@lightningterror 
> Is Align Sprite still needed then?

It is, otherwise the targeting squares on the HUD become thinner rectangles instead of squares. It's hard to notice in screenshots, but:

Without Align Sprite
gs_20220713222324_Ace Combat 5 - The Unsung War_SLUS-20851

With Align Sprite
gs_20220713222319_Ace Combat 5 - The Unsung War_SLUS-20851

Azekthi:
@mirh 
> Explosions are a big issue with upscaling (and I would probably add a mention to it in the comments).
> I also [remember](https://old.reddit.com/r/emulation/comments/56wj30/mipmapping_makes_the_ace_combat_games_run_way/d8osrry/) autoflush used to be necessary there?

I'm not currently seeing any change with autoflush enabled or disabled. What explosions were you referring to?

Azekthi:
@mirh 
> The burst missile explosions. They are blocky AF in hardware mode, and roundSprite=2 breaks them even further.
> On the other hand, you are right.. auto-flush doesn't seem to be doing anything anymore now. Maybe it's related to [#1860 (comment)](https://github.com/PCSX2/pcsx2/issues/1860#issuecomment-1059784635) and [#1547 (comment)](https://github.com/PCSX2/pcsx2/issues/1547#issuecomment-1179761355) (and perhaps I switched the rendering settings too many times now).
Yeah, I figured you were referring to the burst missiles. I had tried enabling and disabling every option and nothing worked for them. If Auto-Flush fixed them at some point in the past, it doesn't now, though I should note that I'm using Vulkan by default.

> ~On the other hand I noticed "disabling safe features" doesn't seem to be doing anything either. And this is so odd though, considering I clearly remember Unscale Point and Line made a huge difference in menus.~ #6628
As mentioned above, I'm using Vulkan, and on there disabling safe features indeed breaks the HUD and menus a bit, so that should remain enabled atm.

This PR is primarily to update some names and apply consistently which hacks are enabled by default for Ace Combat for environment rendering. This isn't a monumental update for these games, unfortunately, just what needs to be set at this time.

mirh:
Yes, sorry, it's just that a combination of ancient memory and unholy bugs led me astray (fun fact: disabling interlace offset may make for "fuller" screenshots without stupid black border lines, but despite its purported description you are actually loosing in overall image clarity). 

Anyhow, I agree on the halfPixelOffset. 
But even putting aside explosions, I'm not totally sold on roundSprite=2. It fixes those little lines around the plane icon, yes, but text in menus gets *fat* enough that it starts to clip around the characters boundary too. 
It's far from unreadable, but I have this little itch it's a bit more difficult to parse (it actually does border unusable, if you were to think to rely on #6553 though)

Azekthi:
Just putting this here somewhere, turns out if you set Blending Accuracy to Minimum, the main menu hover in AC5 gets discolored. Funny thing I just noticed since I was testing AC04 at Minimum.

Also @mirh If the burst missile explosions mean that much to you, I've discovered that you can set the Skipdraw Range to 2 to get rid of the annoying blocks associated with the explosions. It comes with its own visual glitches, most notably hiding the Warning and Score HUD elements every so often, but that might help with investigating what's causing the issue there.

> I'm not totally sold on roundSprite=2.

I didn't set that in this PR. It was already present.

> EDIT: also, while looking at level 5 (Rendezvous) I noticed "half" seems to make for ever so slightly more accurate fog

I don't think that's fog you're seeing. I think those are the shadows of the not-rendered clouds. Without being able to see the clouds, I can't confirm whether Half or Full works best there. I'd rather leave that for now (after all, I'm hoping to get this PR merged on its own merits, not what else needs editing in the gamedb...and I'd appreciate another review from an authorized user at some point...) until we can confirm the better option.

mirh:
They aren't that much of a bother, but half seemed better. Yet, I guess it makes sense not to focus too much on them since they are broken anyway. 

@RedDevilus did ship the present fixes in #5673, probably by just focusing very quickly on the HUD (because no way you wouldn't notice full is quite broken without merge sprite)

p.s. another small place where I found half to make a difference is the lights of the streetlamps of the runaway in the last mission. They are slightly off-center with full.

PR #6602, opened 2022-07-10
GS: Invalidate CLUT by page

refractionpcsx2:
### Description of Changes
Change the CLUT invalidation check to look  by page rather than just by block, since this is really insufficient for CLUT sizes.
Also add check for CLUT invalidation from Z writes, because it *could* happen, weirder things have been done in games.

### Rationale behind Changes
Some games (We Love Katamari) do host->local writes offset inside the CLUT (0x2c00 instead of the CLUT address of 0x2c01), and one block is only 32 colours (in 32bit) anyway, so by block is insufficient.

Write is 32x64 at 0x2c00, each "block contains" 8x8, meaning this spills over in to 0x2c01 and beyond.

### Suggested Testing Steps
Test colours in games, make sure there's no regressions anymore :)

### DIFF
--- a/pcsx2/GS/GSClut.cpp
+++ b/pcsx2/GS/GSClut.cpp
@@ -108,9 +108,11 @@ void GSClut::Invalidate()
 	m_write.dirty = true;
 }
 
+// This checks the whole page! Some games (like We Love Katamari) have the CLUT 1 block ahead of the DBP during a transfer.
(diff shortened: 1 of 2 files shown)


JordanTheToaster (approved):

Mauricio2303-droid:
Good news, when will the menu inside the games without having to leave the same game?

PR #6609, opened 2022-07-11
GS: Fix scanmask interlace offsetting on even numbers

### LINKED ISSUE #6608: [BUG]: Final Fantasy X-2 (PAL) - Bad, interlaced looking appearance with Screen Offsets turned on
### Describe the Bug

This game requires screen offsets to be enabled in order for the gameplay graphics to display at the correct aspect ratio, and for the FMV sequences to display at the correct position. However, with offsets enabled most in-game screens look like they've been badly deinterlaced. The problem appears independent of the interlacing settings though; they do nothing.

There is one PCSX2 setting other than Screen Offsets that has an effect on this issue: Anti-Blur. With Anti-Blur turned off the issues extends to affecting the title screen and most menus rather than just in-game 3D.

There is an in-game setting that mitigates the problem, but you can't set this before starting a game: in the settings panel adjust the screen position up or down one line.

### Reproduction Steps

Either start the game and let the intro play until the first battle, or start the game with Anti-Blur turned off and look at the title screen.

### Expected Behavior

The graphics should display clearly as they do on actual hardware.

### PCSX2 Revision

v1.7.3066

### Operating System

Windows 11

### If Linux - Specify Distro

_No response_

### CPU

Ryzen 3700X

### GPU

RTX 3070

### GS Settings

Software mode. Auto Flush, Mip-Mapping, and Edge Anti-Aliasing on. They don't appear to have any effect on this issue. 

### Emulation Settings

MTVU on. Doesn't appear to change anything if turned off.

### GS Window Screenshots

gs_20220711130429_Final Fantasy X-2 _PAL-E__SLES-51815

### Logs & Dumps

[Final Fantasy X2 Offset.zip](https://github.com/PCSX2/pcsx2/files/9084152/Final.Fantasy.X2.Offset.zip)

refractionpcsx2:
### Description of Changes
Correct the code so if the interlace offset is on an even number, it corrects the offset properly.

### Rationale behind Changes
Before I had it only masking out the lowest bit, which wouldn't have done anything if the offset display had 2 as the lowest nibble.

### Suggested Testing Steps
Test Scanmask games (FFX-2, apparently, MGS3, MGS2, GT4) with and without screen offsets/anti-blur, make sure it's okay. Though I've already checked them :)

Fixes #6608

### DIFF
--- a/pcsx2/GS/Renderers/Common/GSRenderer.cpp
+++ b/pcsx2/GS/Renderers/Common/GSRenderer.cpp
@@ -211,10 +211,13 @@ bool GSRenderer::Merge(int field)
 			// When the displays are offset by 1 we need to adjust for upscale to handle it (reduces bounce in MGS2 when upscaling)
 			interlace_offset += (tex[i]->GetScale().y - 1.0f)  / 2;
 
-			if (!ignore_offset)
-				off.y &= ~1;
+			if (interlace_offset >= 1.0f)
+			{
+				if (!ignore_offset)
+					off.y -= 1;
 
-			display_diff.y &= ~1;
+				display_diff.y -= 1;
+			}
 		}
 
 		// Start of Anti-Blur code.


Silanda:
I've only checked FFX-2 and the intro of MGS3 so far, but it looks good here.

PR #6612, opened 2022-07-11
GameDB: remove texture inside RT from Xenosaga Episode III

MetrosexualGarbodor:
### Rationale behind Changes
When enabled, it broke the title screen.

image

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -1266,7 +1266,6 @@ SCAJ-20179:
   name: "Xenosaga Episode III - Also Sprach Zarathustra [Disc1of2]"
   region: "NTSC-Unk"
   gsHWFixes:
-    textureInsideRT: 1
     autoFlush: 1 # Fixes shadows in cutscenes.
     halfPixelOffset: 1 # Fixes lighting misalignment and reduces ground shadows (probably texture cache issue).
     roundSprite: 2 # Fixes font artifacts.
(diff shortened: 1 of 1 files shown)


F0bes:
Could you squash / drop the commit that deletes the entire GameIndex.yml.

MetrosexualGarbodor:
Done

tellowkrinkle:
Maybe also name the commit something other than "This is a combination of 2 commits."?

MetrosexualGarbodor:
Done

lightningterror (commented):
Commit should be properly named with prefix.

Reonu:
yes this is just nitpicking for the sake of nitpicking, the commit name can be altered at the moment of merge, you should be reviewing the actual change now

Reonu (approved):
i really don't understand why this was added in the first place and should be removed asap given that it produces 0 benefits and more than one bug.

MetrosexualGarbodor:
This was added for some reason 4 months ago, and has been broken by default since 4 months ago. I only noticed because I had the funny idea of checking if the Automatic _Fixes_ were working properly. 
The texture inside RT setting is not even mentioned in this game's wiki page. The only mention I can find relating the 2 things is in PR #3895, but it only says it is "able to fix" an issue with the game, not that it fixes it, and a follow-up comment even shows it doesn't fix anything, so it's honestly baffling that it was added. 

In case you're still on the fence about this change, the _fix_ doesn't only break the title screen: it breaks **all shadows in the game and some lighting instances**. 

image
image


I fixed the commit name. Here's hoping this measly 7 lines of removed code are finally up to PCSX2's high quality coding standards. Please forgive me for not knowing how to use Git to fix something you could've fixed far more easily than I did when merging.

refractionpcsx2:
hey I dunno if you're still around, but can you try enabling Texture in RT (including any other automatic fixes) manually using a build from #8016 ? I'm curious if it's better now with that PR.

refractionpcsx2:
hey @MetrosexualGarbodor I don't know if you're still about, but I was wondering if you could try using Tex in RT on this game again using a build from the #8289 PR?

You will need to manually enable it in the per-game settings, but tex in rt should be a bit more reliable these days, and the main menu looks okay here.

Thanks!

Issue #5244, opened 2022-01-01
[BUG]: FFX FMV Flickering


larwood:
### Describe the Bug

Most but not all FMVs in Final Fantasy X flicker when the Renderer configured is OpenGL.  When the Renderer configured is Software there is no issue.

Once the FMV flicker begins if I quickly toggle the Renderer with F9 from OpenGL -> Software -> OpenGL the issue no longer occurs.

I have not yet watched all 50 FMVs in the The Sphere Theater (Luca Sphere Theater) but the issue does not appear to trigger there.  Perhaps becuase I have to press X to begin the FMV so the transition is different?

I have tested on my Notebook and Desktop with different hardware and the issue is present on both.

There is no difference in behaviour if I load from a Memory Card save or from a Save State.

I have shared logs and a GS dump.

A short 17 second video
https://www.youtube.com/watch?v=11W2vXuybwU

Forum discussion
https://forums.pcsx2.net/Thread-Final-Fantasy-X-Internal-Resolution-Scaling-Incorrect-During-Gameplay-Cutscenes?page=2

### Reproduction Steps

Load the game from either Memory Card save or Save State and wait for transition from in game play to FMV.

The majority but not all FMVs will exhibit the flickering behaviour.

### Expected Behavior

Game play to FMV transition occurs with no flickering.

### PCSX2 Revision

v1.7.2184

### Operating System

Windows 10 (64bit)

### CPU

Intel Core i7 8750H

### GPU

NVIDIA GeForce GTX 1060

### GS Settings

Fresh install no changes from defaults

### Emulation Settings

Fresh install no changes from defaults

### Logs & Dumps

GS dump
https://drive.google.com/file/d/1VsfQnN5RrWuSIebTjE_q79dTrFtGblBz/view?usp=sharing

Logs
https://drive.google.com/file/d/1Jm_py7zY8s99x44UNyBKtvx0GgrNoPws/view?usp=sharing

refractionpcsx2:
I think this has been around for at least a while, but I suspect it will happen randomly depending on what's in VRAM.  Any chance you could check back to when it started happening, if it used to not be an issue?

Of course don't check every build in turn, jump back to 1.7.2000 or something, see if that works, if not, try 1000, then narrow it down etc.

prafullpcsx2:
Looks like you forgot to upload the savestate and memory card save you were talking about in first post.

larwood:
[...]

PR #6659, opened 2022-07-17
GS: Improve target size calcs and remove conservative framebuffer

stenzek:
### Description of Changes

Instead of using a fixed size for all render targets in the hardware renderer, this PR changes it to set the target size based on the FBW that the game configures. Unfortunately, we can't do the same for height, because it's not specified anywhere in the GS (only inferred based on the scissor height).

Using the scissor height alone would be a bit wasteful, because some games configure a stupidly high scissor, but then never draw anything in it. So, we try to be a bit smarter, and use the draw size instead. Naturally, this results in a variable size which changes throughout the frame, so we keep track of it with a FBP+FBW+PSM tuple, which is aged and eventually removed.

This leaves one nasty case which seems to be pretty common (e.g. NFS Carbon, GOW2, Burnout 3). They either do a recursive draw, or a shuffle with the depth buffer, with a scissor and draw height of 896. Because the framebuffer/depth buffer it's sampling from is usually only 448 pixels high, resizing the target to 896 would be kinda silly, because that area will just be rendered black, and waste bandwidth.

So, we try to detect these draws by looking at the gradient in the Y direction, and clamp the target size to the texture size in such cases. This works for everything I've checked so far, and results in a framebuffer size of 512x448 instead of 512x896 as was common before.

As an added bonus, because we're somewhat accurately predicting the target size, we can get rid of the conservative framebuffer option, since targets will only use as much VRAM as they need now. We can probably drop sparse texture too: it was only implemented in OpenGL anyway.

### Rationale behind Changes

Reducing VRAM usage when upscaling, performance increase.

Burnout 3 at 5x on a RX580 goes from 44fps to 85fps or so.

### Suggested Testing Steps

Check GS sensitive games.


refractionpcsx2:
Nearly a clean sweep :) I went through all my dumps and they were all fine, except one.

Test Drive does some sort of huge depth buffer thing, I remember iMine having to fix it when he did the scaling stuff.

Heres a dump:
[testdrivehalfscreen.zip](https://github.com/PCSX2/pcsx2/files/9131209/testdrivehalfscreen.zip)

pcsx2/GameDatabase.cpp
```diff
@@ -542,14 +537,6 @@ u32 GameDatabaseSchema::GameEntry::applyGSHardwareFixes(Pcsx2Config::GSOptions&
 				config.UserHacks_AutoFlush = (value > 0);
 				break;
 
-			case GSHWFixId::ConservativeFramebuffer:
-				config.ConservativeFramebuffer = (value > 0);
-				break;
-
-			case GSHWFixId::CPUFramebufferConversion:
-				config.UserHacks_CPUFBConversion = (value > 0);
-				break;
```
refractionpcsx2, line 551: I presume you didn't mean to remove this? :)

refractionpcsx2:
I know you hate WX, but might be worth cutting it out of there too :) I don't mind not adding new stuff to WX, but if we remove stuff we should really remove the option.

coornio:
madlad at it once again <3

JordanTheToaster (approved):

lightningterror (approved):

AhayriSG:
legendsflickering
https://github.com/PCSX2/pcsx2/issues/5244#issuecomment-1183800737

stenzek:
That GS dump plays back fine for me on this PR, @refractionpcsx2 also checked.

refractionpcsx2:
@AhayriSG did you even try it on this PR before posting that? cos two of us have tested it now and it works fine.

AhayriSG:
> @AhayriSG did you even try it on this PR before posting that? cos two of us have tested it now and it works fine.
yep, build: 9377859
settings;
legendsflickering

JordanTheToaster (approved):

Tyestor:
Keep up the awesome work sten

refractionpcsx2:
[spider2.zip](https://github.com/PCSX2/pcsx2/files/9203169/spider2.zip)

Spider-man 2 has a bar coming out of his shadow (even in native) which isn't there in master

PR #6666, opened 2022-07-18
fix build error when recording is disabled.

MaddTheSane:
### Description of Changes
This makes is so if `DISABLE_RECORDING` is defined, pcsx2/SourceLog.cpp will still build.

### Rationale behind Changes
This change is needed to make PCSX2 to compile when `DISABLE_RECORDING` is defined.

### Suggested Testing Steps
None


refractionpcsx2:
Didn't know this was broken, thanks :)

Issue #6497, opened 2022-06-23
[BUG]: Tales of the Abyss: Line Grid Effect in certain areas within the game


Trixarian:
### Describe the Bug

When upscaling Tales of the Abyss beyond Native resolution, and with Half-Pixel Offset enabled, it causes a line grid effect to appear in several areas within the game. Disabling HPO does solve this issue, but in return results in a ghosting effect around models. The issue also appears if a texture offset is used or when any of the other hacks that have a similar effect, like Align Sprite and Wild Arms Hack, is used. No other arrangement of settings seem to correct this behaviour while HPO is enabled, but notably setting Override Texture Barriers to Force Disabled reduces - and in some instances, eliminates - the effect at the cost of blurring the image.

**Here's an example of the line grid effect in Engeve:**
![gs_20220623075733_Tales of the Abyss_SLUS-21386](https://user-images.githubusercontent.com/983358/175310884-4b8ddeca-84de-4851-8f54-2c796ce4a3b6.jpg)

**Current workarounds:**
The use of the Skipdraw range of 355 and 359 corrects the issue for 90% of the areas affected, but does cause visual bugs within the Desert Oasis and Grand Chokmah. 
Using the HPO setting Normal (Vertex) results in a sharper image, but comes with pixel thick lines of corrupted texture at the top and left side of the screen which requires setting a Zoom of 101% and enabling Screen Offsets to correct. Use of the Anti-Blur setting further sharpens the image. It, however, still requires the Skipdraw range to correct the line grid issue.
Notably the use of the Widescreen patch file results in the the Skipdraw code not working at all. This is due to the use of the No Bloom code within the file, and can be used outside of the Widescreen patch file to reproduce the problem.

**Other known areas that this bug is present:**
* Title Screen
* Cutscene where Tear attacks Van
* Tataroo Valley (Night)
* Engeve
* Kaitzur Navel Port
* Outside Baticul Castle
* Grand Chokmah
* Keterburg
* Keterburg Bay
(See Screenshots below for examples of the above)

### Reproduction Steps

Set any Internal Resolution higher than Native, then enter the affected areas within the game to observe the bug

### Expected Behavior

Preferably a sharp image without the line grid effect in the affected areas, similar to this screenshot using HPO Normal (Vertex) and the Skipdraw range fix:
[...]

PR #6269, opened 2022-05-29
GameDB: Add SkipDraw Range to 'Tales of the Abyss'

Trixarian:
### Description of Changes
Adds the SkipDraw range 355, 359 that fixes the grid effect present in the Title Screen, Engave, Kaitzur Naval Port, Baticul Castle, Keterburg Port and possibly other instances within 'Tales of the Abyss'

### Rationale behind Changes
A recent commit that corrected an issue within Engave eliminated the need for the extreme SkipDraw range (100, 361) called for on the Wiki, but it also broke HPO on the Title Screen resulting in the return of the grid effect it originally corrected. 
This drove me to find a more reasonable range (355, 359) that fixes the same issues, but doesn't come with the flickering, darkening and texture stripping issues that the old range created. It also seems to correct some minor ghosting in the instances that the grid effect would be present in. It does still introduce a visual glitch into the Desert Oasis, but I feel it's an acceptable compromise for what the range corrects - at least until a future commit removes the need for SkipDraw entirely

Before:
Title Screen 01
Baticul Castle 01
Keterburg Bay 01

After:
Title Screen 02
Baticul Castle 02
Keterburg Bay 02

Visual bug introduced into Desert Oasis:
Desert Oasis

### Suggested Testing Steps
Test the game to see if all possible instances of the grid effect has been corrected

PR #6667, opened 2022-07-18
Adds the black grid fix patch for 'Tales of the Abyss' to the GameDB

Trixarian:
Adds the patch found at https://forums.pcsx2.net/Thread-PCSX2-Widescreen-Game-Patches?pid=627849#pid627849 for Tales of the Abyss to the GameDB

### Description of Changes
This adds the patch created by sergx12 to the GameDB. It corrects the black grid effect found in the Titlescreen, Opening Cutscenes, Tataroo Valley (Night), Engeve, Kaitzur Naval Port, Baticul Castle, Grand Chokmah, Keterburg and Keterburg Bay as noted with #6497

### Rationale behind Changes
#6497 has been a long standing issue caused by the use of HPO (or any texture offset) in several areas within the game. Previously the way to correct it was by using a Skipdraw range like those found with #6269 and #6074. This had notable downsides like darkened areas, the introduction of a visual glitch to The Desert Oasis and causing a notable shift of the bloom effect on Native. This patch currently corrects the issue in all known areas without the downsides introduced by the use of Skipdraw.

### Suggested Testing Steps
Test the game to see if any visual glitches are introduced by the use of this patch

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -1194,6 +1194,12 @@ SCAJ-20163:
   region: "NTSC-Unk"
   gsHWFixes:
     halfPixelOffset: 2 # Fixes ghosting.
+  patches:
+    default:
+      content: |-
+        author=sergx12
+        // fixes black grid
+        patch=1,EE,205E570C,extended,00000000
 SCAJ-20164:
   name: "Kingdom Hearts II"
   region: "NTSC-Unk"
@@ -31704,6 +31710,12 @@ SLPM-66897:
   region: "NTSC-J"
   gsHWFixes:
     halfPixelOffset: 2 # Fixes ghosting.
+  patches:
+    default:
+      content: |-
+        author=sergx12
+        // fixes black grid
+        patch=1,EE,205E570C,extended,00000000
 SLPM-66898:
   name: "Spectral Gene [Limited Edition]"
   region: "NTSC-J"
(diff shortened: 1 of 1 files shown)


Mrlinkwii:
dose the patch effect SW mode ?

Trixarian:
> dose the patch effect SW mode ?

I don't believe so. I tested it on Native Resolution and in SW Mode to be sure that it had no negative effect on either

JordanTheToaster:
So if I understand this correctly this is an upscaling patch?

lightningterror (changes requested):
Still this will affect sw renderer, so not a good idea.

DRAGONBLE3API3C3:
lightningterror, does there exist a method to only have this fix apply to hardware mode?

lightningterror:
> > Still this will affect sw renderer, so not a good idea.
> 
> Is there a way to only have this fix apply to hardware mode, then? Surely we have a quarantine-esque command for game fixes, right?

With patches currently no.

DRAGONBLE3API3C3:
> > > Still this will affect sw renderer, so not a good idea.
> > 
> > 
> > Is there a way to only have this fix apply to hardware mode, then? Surely we have a quarantine-esque command for game fixes, right?
> 
> With patches currently no.

Wouldn't that mean that there would then be potentially a large amount of game fixes that can't be automatically applied, then? Maybe I'm getting a bit ahead of myself, but I do think a simple exception system for the game fixes could be a very good thing to implement. Having your emulator be plug and play is a huge boon, in my opinion.

Mrlinkwii:
>Wouldn't that mean that there would then be potentially a large amount of game fixes that can't be automatically applied, then?

@DRAGONBLE3API3C3 
 ok , patches are for core issues  ,  not HW render issues  , patches effect both HW/SW ( even if it dosent look like it ) 

if you want to fix HW renderer issue use a GsHWFix or an upscaling fix ( read the Gameindex  documentation for options  )  if none fix the issue though luck :/

PR #6674, opened 2022-07-19
GameDB: Fix and add Japanese game names, apply upscaling fixes

noigeaR:
### Description of Changes

- Change the name for all releases of Drag-on Dragoon 2 (Japanese name of Drakengard 2). According to [the Japanese Wikipedia page for Drag-on Dragoon 2](https://ja.wikipedia.org/wiki/%E3%83%89%E3%83%A9%E3%83%83%E3%82%B0%E3%82%AA%E3%83%B3%E3%83%89%E3%83%A9%E3%82%B0%E3%83%BC%E3%83%B32_%E5%B0%81%E5%8D%B0%E3%81%AE%E7%B4%85%E3%80%81%E8%83%8C%E5%BE%B3%E3%81%AE%E9%BB%92), redump.org and some other Japanese sources, the correct subtitle is "Fuuin no Aka, Haitoku no Kuro".
The games were also missing two gsHWFixes listed for the US and PAL releases of Drakengard 2.

- Change the name for Simple 2000 Series Vol. 13 to fit the naming of the other games in the series.

- SLPM-62279 was listed as Winning Post 5 - Maximum 2002. This is wrong
SLPM-62279 and SLPM-62280 is a two disc pack.
The correct title for 62279 is GI Jockey 3. Winning Post 5 is actually SLPM-62280. 

### Rationale behind Changes
Improve GameDB

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -1024,10 +1024,13 @@ SCAJ-20131:
   name: "namCollection - Namco 50th Anniversary"
   region: "NTSC-Unk"
 SCAJ-20132:
-  name: "Drag-On Dragon 2 - Fuuin no Kurenai"
+  name: "Drag-on Dragoon 2 - Fuuin no Aka, Haitoku no Kuro"
   region: "NTSC-J"
   clampModes:
     eeClampMode: 2 # Fixes wrong color on some characters and breakable objects.
+  gsHWFixes:
+    halfPixelOffset: 1 # Fixes ghosting characters.
+    mergeSprite: 1 # Align sprite fixes FMVs but not garbage in-game, so needs merge sprite instead.
 SCAJ-20133:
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Just to check, the merge sprite inclusion doesn't break any lighting effects does it? it has a habit of doing that.

noigeaR:
> Just to check, the merge sprite inclusion doesn't break any lighting effects does it? it has a habit of doing that.

I haven't tested it myself, but this fix is currently in GameDB for both Drakengard 2 US and PAL releases and I think US and PAL releases are probably maintained better in the GameDB than the Japanese version. The Japanese release Drag-on Dragoon 2 has a different name, but the game should otherwise be exactly the same as the western versions.

refractionpcsx2:
Ok cool, that's fine then

PR #6675, opened 2022-07-20
Qt: 2 New Theme Options

Daisouji:
### Description of Changes
Added 2 New Options,  A Ruby/Dark Theme (Ruby)  and Another allowing usage of .qss (under the main pcsx2 directory)

**Ruby**

C3
Capture2

**QSS**
C4
c5

### Rationale behind Changes
Thought we needed more color options.

### DIFF
--- a/pcsx2-qt/MainWindow.cpp
+++ b/pcsx2-qt/MainWindow.cpp
@@ -622,7 +622,63 @@ void MainWindow::setStyleFromSettings()
 
 		qApp->setStyleSheet("QToolTip { color: #ffffff; background-color: #2a82da; border: 1px solid white; }");
 	}
-	else
+	else if (theme == "Ruby")
+	{
+		// Custom pallete by Daisouji, Black as main color andd Red as complimentary.
+		// Alternative dark theme.
+		qApp->setStyle(QStyleFactory::create("Fusion"));
+
+		const QColor lighterGray(75, 75, 75);
+		const QColor gray(128, 128, 128);
+		const QColor black(25, 25, 25);
+		const QColor slate(18, 18, 18);
+		const QColor rubyish(172,21,31);
+
+
+		QPalette darkPalette;
+		darkPalette.setColor(QPalette::Window, slate);
+		darkPalette.setColor(QPalette::WindowText, Qt::white);
+		darkPalette.setColor(QPalette::Base, slate.lighter());
+		darkPalette.setColor(QPalette::AlternateBase, slate.lighter());
+		darkPalette.setColor(QPalette::ToolTipBase, slate);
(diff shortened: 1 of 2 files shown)


pcsx2-qt/MainWindow.cpp
```diff
@@ -622,7 +622,63 @@ void MainWindow::setStyleFromSettings()
...
+		qApp->setStyleSheet("QToolTip { color: #ffffff; background-color: #2a82da; border: 1px solid white; }");
+	} 
+	else if (theme == "Custom")
+	{
+		
+		//Additional Theme option than loads .qss from main PCSX2 Directory
+		qApp->setStyle(QStyleFactory::create("Fusion"));
+		
+		QString sheet_content;
+		QFile sheets("./custom.qss");
```
stenzek, line 669: Probably should use a full path here, the current working directory does get changed at runtime (e.g. `QString::fromStdString(Path::Combine(EmuFolders::DataRoot, "custom.qss"))`

pcsx2-qt/MainWindow.cpp
```diff
@@ -622,7 +622,63 @@ void MainWindow::setStyleFromSettings()
...
+		
+		if(sheets.open(QFile::ReadOnly)){
```
stenzek, line 671: Apply clang-format here.

pcsx2-qt/MainWindow.cpp
Daisouji: Sorry I'm kinda iffy on pathing, you are saying concat the filename w/ current-working directory is?

stenzek: No, use DataRoot like I posted above. (DataRoot points to either the documents directory, or the pcsx2 directory if running in portable mode).

refractionpcsx2:
If you could clang format that file, you have a mix of tabs and spaces and same line brackets (we do new line brackets)

lightningterror (changes requested):
Some commits can be squashed, also named properly with prefixes.

Daisouji:
> Some commits can be squashed, also named properly with prefixes.

Sure what are proper naming conventions, I didn't see anything on them.

Mrlinkwii:
QT: <whatever you did > like what you did with the pr title

lightningterror:
> Sure what are proper naming conventions, I didn't see anything on them.

Since you're changing the Qt gui, example would be:
Qt: Add two new themes, Ruby/Dark Theme (Ruby)...
Qt: Clang format filename...

JordanTheToaster (approved):

Mrlinkwii (approved):

PR #6678, opened 2022-07-20
GameDB: addition of missing serials (from H to J) & fixes of titles

rufotheone-fr:
### Description of Changes
Addition of missing serials and fixes of some titles.

### Rationale behind Changes
Consistent and accurate gamedb.

### Suggested Testing Steps
...


rufotheone-fr:
Is there a cache file that I need to delete after an update before rescanning my directories so I can see (and recheck) any change made within GameIndex.yaml?

Mrlinkwii:
@rufotheone-fr  all you need to is just rescan for game and UI will update

PR #6683, opened 2022-07-22
GS-SW: Fix framebuffer looping at 2048 height/width.

refractionpcsx2:
### Description of Changes
Fixes the vertical and horizontal looping of the framebuffer output in the merge circuit in software mode.

### Rationale behind Changes
Before this would just read off the end, look wrong, and crash, this fixes the looped behaviour which we checked on the PS2.

### Suggested Testing Steps
Test games especially with funny offsets (Devil May Cry, Time Crisis multiplayer, WipEout (with screen offsets) for example) make sure they all work okay in software mode, can check hardware too for bonus points.

Fixes the crash on NASCAR '08 in software mode, no other known games benefit from it.

### DIFF
--- a/pcsx2/GS/Renderers/SW/GSRendererSW.cpp
+++ b/pcsx2/GS/Renderers/SW/GSRendererSW.cpp
@@ -35,6 +35,7 @@ GSRendererSW::GSRendererSW(int threads)
 	m_tc = std::make_unique<GSTextureCacheSW>();
 	m_rl = GSRasterizerList::Create<GSDrawScanline>(threads);
 
+	// Max framebuffer size can be 1024x2048 (it will loop at 2048).
 	m_output = (u8*)_aligned_malloc(1024 * 1024 * sizeof(u32), 32);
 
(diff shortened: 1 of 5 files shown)


pcsx2/GS/Renderers/SW/GSRendererSW.cpp
```diff
@@ -143,14 +144,30 @@ GSTexture* GSRendererSW::GetOutput(int i, int& y_offset)
 	if (g_gs_device->ResizeTarget(&m_texture[i], w, h))
 	{
 		static int pitch = 1024 * 4;
-
-		GSVector4i r(0, 0, w, h);
+		int off_x = DISPFB.DBX & 0x7ff;
+		int off_y = DISPFB.DBY & 0x7ff;
```
lightningterror, line 148: These two can be const.

refractionpcsx2:
Here's a comparison of the wrapping behaviour on PCSX2 compared to the PS2 (Software mode only! hardware sucks :( )
wrapcompare

lightningterror (approved):

PR #6687, opened 2022-07-22
GameDB: complete name correction and addition for letter A

rufotheone-fr:
### Description of Changes
Complete check & fix of every title in GameDB + add of missing serial.
Check of titles starting with a "A" (part 1).

### Rationale behind Changes
Consistent and accurate gamedb.

### Suggested Testing Steps
...

### DIFF
--- a/bin/resources/shaders/dx11/tfx.fx
+++ b/bin/resources/shaders/dx11/tfx.fx
@@ -779,7 +779,19 @@ void ps_blend(inout float4 Color, inout float As, float2 pos_xy)
 		if (PS_BLEND_MIX && PS_CLR_HW != 1)
 			C = min(C, 1.0f);
 
-		Color.rgb = (PS_BLEND_A == PS_BLEND_B) ? D : trunc(((A - B) * C) + D);
+		if (PS_BLEND_A == PS_BLEND_B)
+			Color.rgb = D;
+		else if (PS_BLEND_MIX)
+			// In blend_mix, HW adds on some alpha factor * dst.
+			// Truncating here wouldn't quite get the right result because it prevents the <1 bit here from combining with a <1 bit in dst to form a ≥1 amount that pushes over the truncation.
+			// Instead, apply an offset to convert HW's round to a floor.
+			// Since alpha is in 1/128 increments, subtracting (0.5 - 0.5/128 == 127/256) would get us what we want if GPUs blended in full precision.
+			// But they don't.  Details here: https://github.com/PCSX2/pcsx2/pull/6809#issuecomment-1211473399
+			// Based on the scripts at the above link, the ideal choice for Intel GPUs is 126/256, AMD 120/256.  Nvidia is a lost cause.
+			// 124/256 seems like a reasonable compromise, providing the correct answer 99.3% of the time on Intel (vs 99.6% for 126/256), and 97% of the time on AMD (vs 97.4% for 120/256).
+			Color.rgb = ((A - B) * C + D) - (124.0f/256.0f);
+		else
+			Color.rgb = trunc(((A - B) * C) + D);
 
 		if (PS_CLR_HW == 1)
 		{
--- a/bin/resources/shaders/opengl/tfx_fs.glsl
+++ b/bin/resources/shaders/opengl/tfx_fs.glsl
@@ -761,6 +761,15 @@ void ps_blend(inout vec4 Color, inout float As)
 
 #if PS_BLEND_A == PS_BLEND_B
     Color.rgb = D;
+#elif PS_BLEND_MIX
+    // In blend_mix, HW adds on some alpha factor * dst.
+    // Truncating here wouldn't quite get the right result because it prevents the <1 bit here from combining with a <1 bit in dst to form a ≥1 amount that pushes over the truncation.
+    // Instead, apply an offset to convert HW's round to a floor.
+    // Since alpha is in 1/128 increments, subtracting (0.5 - 0.5/128 == 127/256) would get us what we want if GPUs blended in full precision.
+    // But they don't.  Details here: https://github.com/PCSX2/pcsx2/pull/6809#issuecomment-1211473399
+    // Based on the scripts at the above link, the ideal choice for Intel GPUs is 126/256, AMD 120/256.  Nvidia is a lost cause.
+    // 124/256 seems like a reasonable compromise, providing the correct answer 99.3% of the time on Intel (vs 99.6% for 126/256), and 97% of the time on AMD (vs 97.4% for 120/256).
+    Color.rgb = ((A - B) * C + D) - (124.0f/256.0f);
 #else
     Color.rgb = trunc((A - B) * C + D);
 #endif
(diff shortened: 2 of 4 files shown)


refractionpcsx2:
Thanks, however, in future can you name your commits correctly with the GameDB prefix then a brief description of what you're doing? Thanks :)

bin/resources/GameIndex.yaml
```diff
@@ -1540,7 +1546,7 @@ SCCS-40006:
   region: "NTSC-C"
 SCCS-40007:
   name: "Arc the Lad - Seirei no Tasogare"
-  region: "NTSC-C"
+  region: "NTSC-CH"
```
Mrlinkwii, line 1549: this is wrong its NTSC-C ,  https://en.wikipedia.org/wiki/NTSC-C)

rufotheone-fr: Thanks, it wasn't clear because "NTSC-C" & "NTSC-CH" were both in the gamedb :)

Mrlinkwii: if their is any 'NTSC-CH' can you change the to 'NTSC-C' :slightly_smiling_face:

rufotheone-fr:
I'm done with this part.

Check up to "SLES-51821"

Mrlinkwii:
can you squash commits and name the commit correctly :)

rufotheone-fr:
@Mrlinkwii Is this ok like that?

refractionpcsx2:
GameDB: Name correction pass for letter A

That might be better

rufotheone-fr:
I though that I won't have the time to do that this weekend but here it is :)

bin/resources/GameIndex.yaml
```diff
@@ -17165,7 +17207,7 @@ SLES-53866:
   speedHacks:
     MTVUSpeedHack: 0 # Fixes bad graphics due to bad T-Bit handling.
 SLES-53867:
-  name: "Ski Alpin 2006"
+  name: "Alpine Skiing 2006 featuring Bode Miller"
```
Mrlinkwii, line 17210:
```suggestion
  name: "Bode Miller Alpine Skiing"
```

bin/resources/GameIndex.yaml
```diff
@@ -18880,7 +18922,7 @@ SLES-54624:
   name: "Samurai Warriors 2 - Empires"
   region: "PAL-E"
 SLES-54626:
-  name: "An American Tail"
+  name: "American Tail, An"
```
Mrlinkwii, line 18925: no need to change this

bin/resources/GameIndex.yaml
```diff
@@ -13572,12 +13614,12 @@ SLES-52363:
   region: "PAL-G"
 SLES-52366:
   name: "America's 10 Most Wanted"
-  region: "PAL-M5"
+  region: "PAL-UK"
```
Mrlinkwii, line 13617:
```suggestion
  region: "PAL-M5"
```
it was already correct

Mrlinkwii (changes requested):

bin/resources/GameIndex.yaml
Mrlinkwii: http://redump.org/disc/4487/

bin/resources/GameIndex.yaml
rufotheone-fr: No, "Bode Miller Alpine Skiing" is the NTSC-U title.

There is 2 cover titles for SLES-53867:
- "Alpine Skiing 2006 featuring Bode Miller" (for the PAL-E cover)
- "Ski Alpin 2006 mit Bode Miller" (for the PAL-G cover)

bin/resources/GameIndex.yaml
rufotheone-fr: No it isn't. GameIndex asks for the game region, not the languages. So this one is PAL-UK, no matter how many languages it contains... No ?

bin/resources/GameIndex.yaml
rufotheone-fr: Every "The" are dropped to the end of any title for alpha sorting. Why not for "An" (although it doesn't change greatly the sorting)?

bin/resources/GameIndex.yaml
Mrlinkwii: ah  my bad , my mistake then  :+1:

bin/resources/GameIndex.yaml
Mrlinkwii: ah my bad , my mistake then

bin/resources/GameIndex.yaml
rufotheone-fr: @Mrlinkwii @refractionpcsx2 Maybe we should write down some guidelines about titles and regions and insert them in the wiki so we don't loose any extra time about subjects already settled?

refractionpcsx2: No, it's based on the region AND languages.  if it has 5 languages and it's europe, it's PAL-M5, for Multi language (5). same for 3 language it'd be PAL-M3

rufotheone-fr: Sorry to be THAT insistent...
10 Most Wanted (known in the US as "Fugitive Hunter: War on Terror") has 3 versions in PAL area :
* SLES-52367 [America's 10 Most Wanted](http://redump.org/disc/4294/) which is the PAL M5
* SLES-52856 [Most Wanted](http://redump.org/disc/51987/) which was the french version sold in France (and that had M5 too)
* SLES-52366 [America's 10 Most Wanted](http://redump.org/disc/4487/) which was the UK version (it had the BBFC 15 logo on [its cover](http://www.computinghistory.org.uk/userdata/images/large/22/84/product-82284.jpg) and had the M5 too, like the french version)

refractionpcsx2: no, it has 5 languages, so it's M5, I don't care if there's already one M5 version, this is another one.

rufotheone-fr: That's why I talked about some guidelines. I'm fine with setting this one to PAL-M5 but we should write that down somewhere so next contributors know that languages are more important than region.

Could it be possible to have both region and languages in GameIndex.yaml ?
For example, if there's no languages in the "languages" line, the region is the default language.

Example :
`SLES-52366:
  name: "America's 10 Most Wanted"
  region: "PAL-M5"
  languages: E,F,G,I,S`

or

`SLES-52366:
  name: "Akudaikan 3"
  region: "NTSC-J"`
-> no line for language = J

RedDevilus: That is for the future to make it region for real and languages seperately. But GameDB is always a pain.

rufotheone-fr: I'd rather wait for this feature to be real so, because it takes a very long time to check every single title... And I don't want to double that job, once it will be implemented ^_^;

rufotheone-fr:
It should be ok to merge this PR but do the team has any ETA of separate region/languages in the GameIndex.yaml ?

PR #6713, opened 2022-07-25
VU-MTVU: Simulate VU times when in MTVU w/o Instant VU.

refractionpcsx2:
### Description of Changes
Attempt to averagely simulate VU times with Instant VU disabled if MTVU is enabled.

### Rationale behind Changes
Some games such as Juiced can hammer the VU's to try and get as many frames out as possible, you could somewhat repair this with the Cycle Skipping speedhack, but I never liked that. 

### Suggested Testing Steps
Test games with MTVU enabled and Instant VU on and off, especially games that struggle with the VU thread being maxed.

### DIFF
--- a/pcsx2-qt/Settings/SystemSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/SystemSettingsWidget.cpp
@@ -94,17 +94,9 @@ SystemSettingsWidget::SystemSettingsWidget(SettingsDialog* dialog, QWidget* pare
 
 	dialog->registerWidgetHelp(m_ui.fastCDVD, tr("Enable Fast CDVD"), tr("Unchecked"),
 		tr("Fast disc access, less loading times. Check HDLoader compatibility lists for known games that have issues with this."));
-
-	updateVU1InstantState();
-	connect(m_ui.MTVU, &QCheckBox::stateChanged, this, &SystemSettingsWidget::updateVU1InstantState);
 }
 
 SystemSettingsWidget::~SystemSettingsWidget() = default;
-
-void SystemSettingsWidget::updateVU1InstantState()
-{
-	//m_ui.instantVU1->setEnabled(!m_dialog->getEffectiveBoolValue("EmuCore/Speedhacks", "vuThread", false));
-}
 
 int SystemSettingsWidget::getGlobalClampingModeIndex(bool vu) const
 {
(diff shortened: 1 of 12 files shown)


pcsx2/MTVU.cpp
```diff
@@ -458,10 +459,17 @@ void VU_Thread::ExecuteVU(u32 vu_addr, u32 vif_top, u32 vif_itop, u32 fbrst)
 	CommitWritePos();
 	gifUnit.TransferGSPacketData(GIF_TRANS_MTVU, NULL, 0);
 	KickStart();
-	u32 cycles = std::min(Get_vuCycles(), 3000u);
-	cpuRegs.cycle += cycles * EmuConfig.Speedhacks.EECycleSkip;
-	VU0.cycle += cycles * EmuConfig.Speedhacks.EECycleSkip;
+	u32 cycles = std::max(Get_vuCycles(), 4u);
+	u32 skip_cycles = std::min(cycles, 3000u);
+	cpuRegs.cycle += skip_cycles * EmuConfig.Speedhacks.EECycleSkip;
+	VU0.cycle += skip_cycles * EmuConfig.Speedhacks.EECycleSkip;
```
MageKing17, line 465: Is the behavior change to MTVU+Instant VU1 here if `Get_vuCycles()` returns a number less than 4 intentional?

refractionpcsx2: Yes, it's to stop it returning 0, which would break the internal interrupt stuff

MageKing17: So this causes problems with the current code?

refractionpcsx2: It would do had I not done that, yeah.

MageKing17: So it looks like the interrupt is only when Instant VU1 is off, not on, which leads me back to: was it intentional to change the behavior when Instant VU1 is on?

refractionpcsx2: It isn't, but it really won't make any difference, 4 cycles is practically zero

refractionpcsx2: Further to this it only matters of you have cycle skipping enabled

pcsx2-qt/Settings/SystemSettingsWidget.cpp
```diff
@@ -103,7 +103,7 @@ SystemSettingsWidget::~SystemSettingsWidget() = default;
 
 void SystemSettingsWidget::updateVU1InstantState()
 {
-	m_ui.instantVU1->setEnabled(!m_dialog->getEffectiveBoolValue("EmuCore/Speedhacks", "vuThread", false));
+	//m_ui.instantVU1->setEnabled(!m_dialog->getEffectiveBoolValue("EmuCore/Speedhacks", "vuThread", false));
```
lightningterror, line 106: Why not remove the function instead?

refractionpcsx2: I was going to?

CelestePolo:
Hello, I get this improvement on SotC's intro:
pcsx2-SotC-VU
Although the game VPS remains sometimes low (lowest : ~24VPS).

PR #6714, opened 2022-07-26
GameDB: Energy Airforce fixes

JordanTheToaster:
### Description of Changes
Adds autoflush to Energy Airforce Aim Strike.

### Rationale behind Changes
Makes the engines more glow-y.

### Suggested Testing Steps
Make sure CI is happy.


lightningterror:
In the future would be nice to add what these fixes are and what they fix in commit description.

PR #6717, opened 2022-07-26
cheats_ni: Cleanup no interlace patches.

lightningterror:
### Description of Changes

cheats_ni: Cleanup no interlace patches.

Remove widescreen, blur, and other crap patches that shouldn't be here.
### Rationale behind Changes

Self explanatory what patches should be included.


CookiePLMonster:
While those are still checked in the repository as a single file, it's probably a good idea to amend the commit message with the list of affected games.

PR #6719, opened 2022-07-26
Load the complete ROM1 - 4MB

Florin9doi:
### Description of Changes
- Load the complete ROM1 - 4MB
- EROM is part of ROM1, its exact location vary and can't be predicted, so it was removed. It should already be included in ROM1.
- To work is required a BIOS dump in the following format:
  - ROM0 (4 MB) / **ROM1 (2/4 MB)** / ROM2 (0/512 KB)
- This format ~won't work~ is not recommended:
  - BIN (4 MB) / **ROM1 (256/512 KB) / EROM (1.75/3 MB)** / ROM2 (0/512 KB)
- GsCrt modes 0x82 and 0x83 are incorrect. Maybe someone will fix them but until then at least they will no longer crash.


refractionpcsx2:
why is loading a 4mb block so important? what breaks if you don't do that?  The sizes you said don't work are the sizes of those ROMS as reported by the PS2.

Florin9doi:
The issue is that EROM is stored at different locations across models. On FATs it is stored ar 0x40000 (early), 0x80000 (late) relative to ROM1 base. On SLIMs there are multiple copies stored in ROM1, for PAL, USA, JP, and the proper location is determined dynamically. It's hard to guess the proper address for every bios without additional information.

refractionpcsx2:
okay, but we can't decrypt the EROM anyways, so it doesn't really matter too much.

refractionpcsx2:
if the only issue is the storage location, then the better idea would to be to map it correctly like we do with NVM files depending on the BIOS revision, then it's not a problem.

Florin9doi:
It may offer time to users to extract their BIOS in the expected format for when the decryption will be possible 😄

F0bes:
I assume this is the result of a conversation I had in a discord server with someone else.
Completely ignoring this point I brought up then too:

> okay, but we can't decrypt the EROM anyways, so it doesn't really matter too much.

I'm against this as I don't see the point in this. Even if you really need the erom in it's proper place, let's not break compatibility with the old dumper, and now, the new dumper that was just released. You've failed to provide a good reason to do this.

uyjulian:
I think it would be better to keep compatibility with the broken, old layout alongside the new layout.

Florin9doi:
The underdumped BIOSes will continue to work exactly the same how they work today. The EROM was only loaded so far, but it was not even mapped to memory, it was completely unused.

refractionpcsx2:
If that's true, why didn't you open with that? Not being mapped at all is a much more valid reason than "This thing we can't use is in the wrong place"

However it looks like it was mapped to EROM to me?

Florin9doi:
https://github.com/PCSX2/pcsx2/blob/5e87f6d806fe34e361cb1494e5fee893daf06388/pcsx2/x86/iR3000A.cpp#L805
https://github.com/PCSX2/pcsx2/blob/5e87f6d806fe34e361cb1494e5fee893daf06388/pcsx2/x86/ix86-32/iR5900-32.cpp#L616

refractionpcsx2:
Okay I see what you mean.

refractionpcsx2:
well, it would be nice if you provided a solution for all existing bios dumps, if an .erom file is detected then map that to the correct space in memory.  You can check the BIOS version and adjust the mapped location manually.  I don't agree with not supporting existing roms and rom dumpers to force your own ideal on to everyone.

refractionpcsx2:
Woudl you like to try https://github.com/F0bes/biosdrain/pull/10 along with your PR? Also if you could make your commit names somewhat match our standard that would be cool (an appropriate prefix, like BIOS: or something..)

Florin9doi:
Rebased and prefixed. I tested https://github.com/F0bes/biosdrain/pull/10 on a 90k and it extracts a ROM1 file compatible with this PR, having the expected results - both alone and combined with #4274. After these changes, the bios files may be checked against redump database, eventually automatically.

refractionpcsx2:
Alright, thanks :)

PR #6721, opened 2022-07-27
GameDB: Replace Evolution Snowboarding patches.

Goatman13:
### Description of Changes
Instead of skipping videos, patches extend stack in problematic function. So required data for VIF1 is no longer overwritten. Workaround for missing EE data cache emulation.

### Rationale behind Changes
Better solution than just skipping videos.

### Suggested Testing Steps
Test affected games. US/JPN version have "Dolby" PSS video, and later first comics styled cutscene right after starting new game. EU game should have additional "Demo" video right after "Dolby" one. Plus if someone have save game far enough in-game, all sponsor videos (241, etc.) are also PSS ones.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -11831,9 +11831,12 @@ SLES-51392:
   patches:
     C24C7FE3:
       content: |-
-        author=kozarovv
-        // Skip PSS movies, workaround for EE Cache requirement.
-        patch=1,EE,001233e0,word,100000d7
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Is it EE cache causing bad data to VIF1 or the lack of VIF1 FIFO doing it? just out of curiosity.  Cool that you came up with better patches though, so thanks for that :D

refractionpcsx2:
fair enough :) Just wanted to check

PR #6738, opened 2022-07-29
TEST: GS: Use 16bit RGBA float for HDR copies

refractionpcsx2:
### Description of Changes
Use 16bit floats for HDR copies, instead of 32bit.

### Rationale behind Changes
16bit floats are faster to copy on the GS and it *might* be precise enough for us to get away with it, but this is a test to find out if it does cause any problems. This does bring a 40% performance boost in Sly 2 when GPU limited, maybe more, I became CPU limited :(

### Suggested Testing Steps
Try games, make sure colours aren't funky compared to master, note any performance increases.


refractionpcsx2:
Sly 2 at x4 Native

Master:
image

PR:
image

PR #6746, opened 2022-07-31
[BUG] Fix a crash that happens when copying from one card to a folder memory card via BIOS

CakeKing64:
Fixes a crash when attempting to copy from one memory card to a folder memory card via the bios

(And maybe some other problems)

### Description of Changes
This fixes a crash that happens if you attempt to copy from one memory card to a formatted folder memory card

### Rationale behind Changes
I would like that to not happen

### Suggested Testing Steps
1. Have two memory cards (source being either a file or folder, the destination being a folder)
2. Make sure the destination (folder card) is formatted (doesn't seem to crash if it isn't)
3. Copy save data from the source to the destination folder via BIOS
4. Crash (or don't) :)

### P.S
This is probably a god awful fix and pull request and might've been better as a issue but I tried :)


pcsx2/MemoryCardFolder.cpp
```diff
@@ -2238,6 +2238,12 @@ bool FileAccessHelper::CleanMemcardFilenameEndDotOrSpace(char* name, size_t leng
 
 bool MemoryCardFileMetadataReference::GetPath(std::string* fileName) const
 {
+
```
lightningterror, line 2241: You can remove the extra line.

CakeKing64: Will do, wasn't sure about the line spacings lol

CakeKing64:
So after a hellela lot of looking around, found a few things out.

1) Turns out title was a lie, this actually just crashes any time data is saved

2) The cause of the crash was that there was a desktop.ini file (tho any file will cause it) placed in the root of the memcard folder (with the superblock) and because it's on the root yeah it gets nullptr
(pretty sure that's more or less the reason)


Anyways this "fix:tm:" also has the cool effect of making files not show up as corrupted data in the BIOS, guess that's a cool bonus
done by just not calling AddFile on files with nullptr as their parent

Tried out in games, seems to have no effect on saving

Also if the console message sucks I'll redo it

pcsx2/MemoryCardFolder.cpp
```diff
@@ -488,7 +488,7 @@ bool FolderMemoryCard::AddFolder(MemoryCardFileEntry* const dirEntry, const std:
 			if (file.m_isFile)
 			{
 				// don't load files in the root dir if we're filtering; no official software stores files there
-				if (enableFiltering && parent == nullptr)
+				if ((enableFiltering && parent == nullptr) || parent == nullptr)
```
lightningterror, line 491: Why not just check `parent == nullptr`?

CakeKing64: Yeah I didn't think about that :(

CakeKing64:
Also I've left the comment there because not sure about rewriting it or just removing it in it's entirety + not good with that
https://github.com/PCSX2/pcsx2/blob/b3590430c97e8df170df0c00ebf466ce848edbf9/pcsx2/MemoryCardFolder.cpp#L493

pcsx2/MemoryCardFolder.cpp
```diff
@@ -488,7 +488,7 @@ bool FolderMemoryCard::AddFolder(MemoryCardFileEntry* const dirEntry, const std:
 			if (file.m_isFile)
 			{
 				// don't load files in the root dir if we're filtering; no official software stores files there
-				if (enableFiltering && parent == nullptr)
```
RedPanda4552, line 491: I understand _how_ this works (we shouldn't care about filtering or not, loose files should not exist in the folder's root, if they do they are made by homebrew to which I say too bad, do what games do and make a folder). I would _infer_ the original purpose behind the filtering check was that homebrew could run with filtering off and then have loose files in the root, but per my previous sentence I think that's just too bad and homebrew should update to match the standards retail games used, instead of having more jank littered in folder memcard code to make them work.

Conversely, an argument could be made that while Sony's standards were that loose files should not exist on the root, the memcards themselves did not have any control logic to prevent this and it was indeed legal at a hardware level. It was only ever prohibited by SCE policing licensed releases. So you could argue that the folder memcard implementation should also mirror that, instead of just hitting a continue keyword because we're in the root.

I think this is fine for now, for the sake of getting this fixed, but certainly down the line I think this should be revisited and perhaps reworked. This comment isn't something to fix, more like a note to myself or whoever else gets to this, about some considerations for what the logic should be.

pcsx2/MemoryCardFolder.cpp
```diff
@@ -2063,7 +2063,15 @@ void FileAccessHelper::WriteIndex(const std::string& baseFolderName, MemoryCardF
 	pxAssert(entry->IsFile());
 
 	std::string folderName(baseFolderName);
-	parent->GetPath(&folderName);
+	if (parent != nullptr)
+	{
+		parent->GetPath(&folderName);
+	}
+	else
```
RedPanda4552, line 2070: Is baseFolderName always a non-empty string? If the else path is taken then no assignment is made to folderName, so folderName will copy the value from baseFolderName and never be written again. Fine as long as baseFolderName is guaranteed to be (correctly) populated, problematic if its possible that it too could be empty.

RedPanda4552 (approved):
I posted some concerns in comments, but overall I think this is fine to ship. It may not be "perfect" or "technically correct" but I think it is good enough to keep things working without side effects.

pcsx2/MemoryCardFolder.cpp
CakeKing64: It doesn't seem like it can be empty, baseFolderName is gotten from m_folderName which is set when the card is opened.

https://github.com/PCSX2/pcsx2/blob/7d1e7f2b58ea839563310ae1b284ac15c780f17e/pcsx2/MemoryCardFolder.cpp#L168-L173

and there is only one call to FileAccessHelper::WriteIndex
https://github.com/PCSX2/pcsx2/blob/7d1e7f2b58ea839563310ae1b284ac15c780f17e/pcsx2/MemoryCardFolder.cpp#L1369-L1372

PR #6749, opened 2022-07-31
PAD: Fix dualshock lights for more than one controller attached

AKuHAK:
Codacy warnings fixed

### Description of Changes

Currently only first controller correctly handles lights, this pr fixes it


refractionpcsx2:
uhh, that's a hack, it just makes all the sendState.lights values underflow, which is not a desired behaviour.

Go back and try to find the real problem and fix that.

lightningterror:
@AKuHAK Status?

PR #6757, opened 2022-08-01
FindMinMax fixes

tellowkrinkle:
### Description of Changes
- Reduces the number of different FindMinMax instantiations we use by only setting the parameters if it will make a difference
- Fixes handling of lines when `provoking_vertex_first` is set, as well as handling of the final triangle in triangle lists with odd triangle counts
- Fixes handling of FindMinMax sprites' Z and color values

### Rationale behind Changes
- Binary size (saves a whole 32kb, truly amazing)
- Minor correctness things

### Suggested Testing Steps
Make sure nothing broke

### DIFF
--- a/pcsx2/GS/Renderers/Common/GSVertexTrace.cpp
+++ b/pcsx2/GS/Renderers/Common/GSVertexTrace.cpp
@@ -27,9 +27,7 @@ GSVertexTrace::GSVertexTrace(const GSState* state, bool provoking_vertex_first)
 	memset(&m_alpha, 0, sizeof(m_alpha));
 
 	#define InitUpdate3(P, IIP, TME, FST, COLOR) \
-	m_fmm[COLOR][FST][TME][IIP][P] = \
-		provoking_vertex_first ? &GSVertexTrace::FindMinMax<P, IIP, TME, FST, COLOR, true> : \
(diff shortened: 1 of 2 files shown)


pcsx2/GS/Renderers/Common/GSVertexTrace.cpp
```diff
@@ -153,6 +151,20 @@ void GSVertexTrace::Update(const void* vertex, const u32* index, int v_count, in
 	}
 }
 
+template <GS_PRIM_CLASS primclass, u32 iip, u32 tme, u32 fst, u32 color>
+GSVertexTrace::FindMinMaxPtr GSVertexTrace::GetFMM(bool provoking_vertex_first)
+{
+	constexpr bool real_iip = primclass == GS_SPRITE_CLASS ? false : iip;
+	constexpr bool real_fst = tme ? fst : false;
+	bool provoking_vertex_first_class = primclass == GS_LINE_CLASS || primclass == GS_TRIANGLE_CLASS;
+	bool swap = provoking_vertex_first_class && !iip && provoking_vertex_first;
```
lightningterror, line 160: These two can be constants, didn't check the vectors if they are set again so maybe those as well.

PR #6760, opened 2022-08-01
Double quotas fixed

AKuHAK:

### DIFF
--- a/.github/workflows/scripts/linux/appimage-qt.sh
+++ b/.github/workflows/scripts/linux/appimage-qt.sh
@@ -172,7 +172,7 @@ rm -fr "$OUTDIR/usr/bin/resources/locale"
 
 # Patch RPATH so the binary goes hunting for shared libraries in the AppDir instead of system.
 echo "Patching RPATH in ${BINARY}..."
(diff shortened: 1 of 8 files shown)


.github/workflows/scripts/linux/appimage-qt.sh
```diff
@@ -172,7 +172,7 @@ rm -fr "$OUTDIR/usr/bin/resources/locale"
 
 # Patch RPATH so the binary goes hunting for shared libraries in the AppDir instead of system.
 echo "Patching RPATH in ${BINARY}..."
-patchelf --set-rpath '$ORIGIN/../lib' "$OUTDIR/usr/bin/$BINARY"
+patchelf --set-rpath "$ORIGIN/../lib" "$OUTDIR/usr/bin/$BINARY"
```
stenzek, line 175: This is wrong. We **do not** want to expand the variable here, literally `$ORIGIN` is what the ELF's rpath should be set to. Same for all the others.

refractionpcsx2:
Can you at least fill out the description, and please fix your commit name, it should be prefixed with "Build:" at least.

PR #6767, opened 2022-08-02
UI-Qt: Add Bilinear dropdown to SW options

refractionpcsx2:
### Description of Changes
Adds a Texture Filtering (Bilinear) dropdown to the Software renderer settings.

### Rationale behind Changes
This setting affects hardware and software mode, so it needs to always be visible.
Other option is to  move it, but figured I'd keep it with the renderer settings.

### Suggested Testing Steps
Switch between hardware and software mode, flicking it to a different setting, make sure it applies on both.

### DIFF
--- a/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
@@ -250,6 +250,7 @@ GraphicsSettingsWidget::GraphicsSettingsWidget(SettingsDialog* dialog, QWidget* 
 	//////////////////////////////////////////////////////////////////////////
 	// SW Settings
 	//////////////////////////////////////////////////////////////////////////
+	SettingWidgetBinder::BindWidgetToIntSetting(sif, m_ui.swTextureFiltering, "EmuCore/GS", "filter", static_cast<int>(BiFiltering::PS2));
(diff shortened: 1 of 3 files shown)


JordanTheToaster (approved):

pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
```diff
@@ -473,6 +476,16 @@ GraphicsSettingsWidget::GraphicsSettingsWidget(SettingsDialog* dialog, QWidget*
 
 GraphicsSettingsWidget::~GraphicsSettingsWidget() = default;
 
+void GraphicsSettingsWidget::onTextureFilteringChange()
+{
+	m_ui.swTextureFiltering->setCurrentIndex(m_ui.textureFiltering->currentIndex());
```
stenzek, line 481: This will update settings twice (setting the current index on the other combo will fire the signals on it, just like the user selecting the option would).

Easiest way would be to use `QSignalBlocker sb(m_ui.swTextureFiltering);`.

refractionpcsx2: Thanks, I had a feeling you'd know better :P I'll change that in a bit

lightningterror:
Maybe in Display tab would be better.

PR #6771, opened 2022-08-03
GameDB: add Panzer Elite Action upscaling fix

noigeaR:
### Description of Changes
Adds Half Pixel Offset setting of Normal (Vertex) to Panzer Elite Action - Fields of Glory.

### Rationale behind Changes
HPO 1 fixes misaligned bloom when upscaling the game in hw mode.

Hardware renderer (Vulkan) without HPO
Panzer Elite Action - Fields of Glory_SLES-53444_20220801111518

Hardware renderer with HPO applied
Panzer Elite Action - Fields of Glory_SLES-53444_20220801111540

Software renderer for reference
Panzer Elite Action - Fields of Glory_SLES-53444_20220801111522

This does **not** fix the broken 3d models for the player tank and some of the enemies.
The only known fix for this problem is to manually set EE Cycle Rate to 300%
Panzer Elite Action - Fields of Glory_SLES-53444_20220801111552

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -16137,6 +16137,8 @@ SLES-53444:
   region: "PAL-M5"
   gameFixes:
     - VUSyncHack # Partly fixes SPS still needs EE+3.
+  gsHWFixes:
+    halfPixelOffset: 1 # Fixes misaligned bloom.
 SLES-53446:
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Thanks :) Yes the only solution for the textures is 300% overclock on the EE

PR #6772, opened 2022-08-03
MemoryCardFolder: Add write osd notification.

stuken:
### Description of Changes
Adds write osd notifications for folder based memcards.
Moved the logging point to the FolderMemoryCardAggregator class and not in the Save method like the File impl to avoid the recursion and potential message spamming in the folder impl's Save method.

### Rationale behind Changes
This was not implemeted for folder memcard so adding for parity with file based ones.

### Suggested Testing Steps
Might fall over with some kind of funky folder name but shouldn't be any less brittle than file based saves since its all going through the same string handling paths.


RedPanda4552 (approved):
LGTM

PR #6787, opened 2022-08-06
GS-HW: Correct FBMask masks for texture shuffles

### LINKED ISSUE #3338: Incredible Hulk Ultimate Destruction - Broken bloom effect in HW renderer
Theres a fake bloom effect which is broken only in the OpenGL HW renderer even at native resolution.

OpenGL SW:
gsdx_20200429134505

OpenGL HW:
gsdx_20200429134507

DX11 HW:
gsdx_20200430175034

I've tried various combination of settings and hardware hacks, nothing seems to do anything.

refractionpcsx2:
### Description of Changes
Corrects what masks are used during a texture shuffle on the FBMask

### Rationale behind Changes
The masks were errantly paired with the wrong colour channels due to poor naming, when doing a Texture shuffle in 16bit, you select either an GR or AB colour combination to write to, then the FBMask represents what bits of this get written.  if we work out that it's writing to AB and the FBMask (when converted to a 16bit format) is masking the upper 8 bits, we want to mask the alpha and write the blue, but we need to also set the right mask to the channel, else you could end up writing either the incorrect amount of data, or no data at all!

### Suggested Testing Steps
Test games that in hardware (if shuffle games are known, they are a good candidate), I've added temp logging for the case where it might have written the wrong stuff.

Fixes Scarface's weird light ghosting effect. Some other games trigger it, but None seem affected in any obvious visual way.
Fixes #3338

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
@@ -2034,14 +2034,17 @@ void GSRendererHW::EmulateTextureShuffleAndFbmask()
 		// Please bang my head against the wall!
 		// 1/ Reduce the frame mask to a 16 bit format
 		const u32 m = m_context->FRAME.FBMSK & GSLocalMemory::m_psm[m_context->FRAME.PSM].fmsk;
+
+		// fbmask is converted to a 16bit version to represent the 2 32bit channels it's writing to.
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Also fixes Hulk

Master:
image

PR:
image

refractionpcsx2:
And The Incredible Hulk 

Master:
image

PR:
image

Software for reference:
image

lightningterror (approved):

PR #6807, opened 2022-08-08
Qt: Force helpText size to avoid overflow at default window size

BreakPoints:
### Description of Changes

Increase minimum size of helpText to match max so that "Adjust to Host Refresh Rate" doesn't overflow it's help text and create a scrollbar at minimum settings window height.

Master - Settings window minimum (default) height:
OG

Master - Required height to avoid scrollbar:
OG 2

PR - Settings window minimum (default) height:
Fixed

### Rationale behind Changes

Resizing height by 20 pixels is a bit pointless with the amount of unused space at the minimum window size and leads to "Adjust to Host Refresh Rate" having it's help text overflow and create a scrollbar that cannot be interacted with.

### Suggested Testing Steps

Open settings window, make sure it is set to the minimum height and mouse over the "Adjust to Host Refresh Rate" option and observe help text.

### DIFF
--- a/pcsx2-qt/Settings/SettingsDialog.ui
+++ b/pcsx2-qt/Settings/SettingsDialog.ui
@@ -66,7 +66,7 @@
      <property name="minimumSize">
       <size>
        <width>0</width>
-       <height>100</height>
+       <height>120</height>
       </size>
      </property>
      <property name="maximumSize">


BreakPoints:
There are a lot of inconsistencies with resizable elements in Qt, I want to follow this up with more tweaks

PR #6809, opened 2022-08-08
GS:HW: More accurate blend equation for blend mix

### LINKED ISSUE #6755: [BUG]: GS: blend_mix inaccuracies causing color banding
### Describe the Bug

I'm putting issue in per @refractionpcsx2 There is an issue with color banding in some games when any blending mode other than Minimum or Maximum is used. Maximum looks better between the two (higher contrast) but in some games (like Grandia 3 and Shadow Man) have a massive performance impact. Instances that I encountered were:

Fatal Frame - Flashlights and darkness
Fatal Frame 3 - Particularly visible on rounded surfaces (like the main character's butt)
Grandia 3 - Same as Fatal Frame 3
Shadow Man: 2econd Coming - very visible on the blur and fire effects seen on the title screen, this is probably the most obvious case of them all
Final Fantasy XII - Blur effect when Reks is waking up at the beginning of the game

Musashi - Samurai Legend has a similar issue as well, though it can be corrected with blending set to medium on GL or by using Auto Flush.

### Reproduction Steps

Play any of the affected games with blending set to anything between Basic and Full. 

### Expected Behavior

For there not to be color banding, and for it to look similar to either Minimum or Maximum blending modes.

### PCSX2 Revision

v1.7.3146

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

### CPU

Ryzen 9 5900X

### GPU

RTX 3080

### GS Settings

_No response_

### Emulation Settings

_No response_

### GS Window Screenshots

Examples:

Shadow Man - Minimum:

image

Shadow Man - Basic:

image

Shadow Man - Maximum:

image

Fatal Frame 3 - Minimum:

image

Fatal Frame 3 - Basic:

image

Fatal Frame 3 - Maximum:

image

### Logs & Dumps

[Shadow Man - 2econd Coming_SLUS-20413_20220731164949.gs.zst.zip](https://github.com/PCSX2/pcsx2/files/9229693/Shadow.Man.-.2econd.Coming_SLUS-20413_20220731164949.gs.zst.zip)
[Fatal Frame 3 - The Tormented_SLUS-21244_20220731165200.gs.zst.zip](https://github.com/PCSX2/pcsx2/files/9229696/Fatal.Frame.3.-.The.Tormented_SLUS-21244_20220731165200.gs.zst.zip)
[Musashi Samurai Legend_SLUS-20983_20220731154957.zip](https://github.com/PCSX2/pcsx2/files/9229703/Musashi.Samurai.Legend_SLUS-20983_20220731154957.zip)

tellowkrinkle:
### Description of Changes
Takes into account what the hardware blend will be doing when calculating for blend mix

### Rationale behind Changes
Fixes #6755

### Suggested Testing Steps
Test games from #6755
Test SOTC bloom, which breaks on normal HW blend

### DIFF
--- a/bin/resources/shaders/dx11/tfx.fx
+++ b/bin/resources/shaders/dx11/tfx.fx
@@ -779,7 +779,19 @@ void ps_blend(inout float4 Color, inout float As, float2 pos_xy)
 		if (PS_BLEND_MIX && PS_CLR_HW != 1)
 			C = min(C, 1.0f);
 
-		Color.rgb = (PS_BLEND_A == PS_BLEND_B) ? D : trunc(((A - B) * C) + D);
+		if (PS_BLEND_A == PS_BLEND_B)
+			Color.rgb = D;
+		else if (PS_BLEND_MIX)
+			// In blend_mix, HW adds on some alpha factor * dst.
+			// Truncating here wouldn't quite get the right result because it prevents the <1 bit here from combining with a <1 bit in dst to form a ≥1 amount that pushes over the truncation.
+			// Instead, apply an offset to convert HW's round to a floor.
+			// Since alpha is in 1/128 increments, subtracting (0.5 - 0.5/128 == 127/256) would get us what we want if GPUs blended in full precision.
+			// But they don't.  Details here: https://github.com/PCSX2/pcsx2/pull/6809#issuecomment-1211473399
+			// Based on the scripts at the above link, the ideal choice for Intel GPUs is 126/256, AMD 120/256.  Nvidia is a lost cause.
+			// 124/256 seems like a reasonable compromise, providing the correct answer 99.3% of the time on Intel (vs 99.6% for 126/256), and 97% of the time on AMD (vs 97.4% for 120/256).
+			Color.rgb = ((A - B) * C + D) - (124.0f/256.0f);
+		else
+			Color.rgb = trunc(((A - B) * C) + D);
 
 		if (PS_CLR_HW == 1)
 		{
--- a/bin/resources/shaders/opengl/tfx_fs.glsl
+++ b/bin/resources/shaders/opengl/tfx_fs.glsl
@@ -761,6 +761,15 @@ void ps_blend(inout vec4 Color, inout float As)
 
 #if PS_BLEND_A == PS_BLEND_B
     Color.rgb = D;
+#elif PS_BLEND_MIX
+    // In blend_mix, HW adds on some alpha factor * dst.
+    // Truncating here wouldn't quite get the right result because it prevents the <1 bit here from combining with a <1 bit in dst to form a ≥1 amount that pushes over the truncation.
+    // Instead, apply an offset to convert HW's round to a floor.
+    // Since alpha is in 1/128 increments, subtracting (0.5 - 0.5/128 == 127/256) would get us what we want if GPUs blended in full precision.
+    // But they don't.  Details here: https://github.com/PCSX2/pcsx2/pull/6809#issuecomment-1211473399
+    // Based on the scripts at the above link, the ideal choice for Intel GPUs is 126/256, AMD 120/256.  Nvidia is a lost cause.
+    // 124/256 seems like a reasonable compromise, providing the correct answer 99.3% of the time on Intel (vs 99.6% for 126/256), and 97% of the time on AMD (vs 97.4% for 120/256).
+    Color.rgb = ((A - B) * C + D) - (124.0f/256.0f);
 #else
     Color.rgb = trunc((A - B) * C + D);
 #endif
--- a/bin/resources/shaders/vulkan/tfx.glsl
+++ b/bin/resources/shaders/vulkan/tfx.glsl
@@ -1060,6 +1060,15 @@ void ps_blend(inout vec4 Color, inout float As)
 
 		#if PS_BLEND_A == PS_BLEND_B
 				Color.rgb = D;
+		#elif PS_BLEND_MIX
+				// In blend_mix, HW adds on some alpha factor * dst.
+				// Truncating here wouldn't quite get the right result because it prevents the <1 bit here from combining with a <1 bit in dst to form a ≥1 amount that pushes over the truncation.
+				// Instead, apply an offset to convert HW's round to a floor.
+				// Since alpha is in 1/128 increments, subtracting (0.5 - 0.5/128 == 127/256) would get us what we want if GPUs blended in full precision.
+				// But they don't.  Details here: https://github.com/PCSX2/pcsx2/pull/6809#issuecomment-1211473399
+				// Based on the scripts at the above link, the ideal choice for Intel GPUs is 126/256, AMD 120/256.  Nvidia is a lost cause.
+				// 124/256 seems like a reasonable compromise, providing the correct answer 99.3% of the time on Intel (vs 99.6% for 126/256), and 97% of the time on AMD (vs 97.4% for 120/256).
+				Color.rgb = ((A - B) * C + D) - (124.0f/256.0f);
 		#else
 				Color.rgb = trunc((A - B) * C + D);
 		#endif
--- a/pcsx2/GS/Renderers/Metal/tfx.metal
+++ b/pcsx2/GS/Renderers/Metal/tfx.metal
@@ -750,6 +750,15 @@ struct PSMain
 
 			if (PS_BLEND_A == PS_BLEND_B)
 				Color.rgb = D;
+			else if (PS_BLEND_MIX)
+				// In blend_mix, HW adds on some alpha factor * dst.
+				// Truncating here wouldn't quite get the right result because it prevents the <1 bit here from combining with a <1 bit in dst to form a ≥1 amount that pushes over the truncation.
+				// Instead, apply an offset to convert HW's round to a floor.
+				// Since alpha is in 1/128 increments, subtracting (0.5 - 0.5/128 == 127/256) would get us what we want if GPUs blended in full precision.
+				// But they don't.  Details here: https://github.com/PCSX2/pcsx2/pull/6809#issuecomment-1211473399
+				// Based on the scripts at the above link, the ideal choice for Intel GPUs is 126/256, AMD 120/256.  Nvidia is a lost cause.
+				// 124/256 seems like a reasonable compromise, providing the correct answer 99.3% of the time on Intel (vs 99.6% for 126/256), and 97% of the time on AMD (vs 97.4% for 120/256).
+				Color.rgb = ((A - B) * C + D) - (124.f/256.f);
 			else
 				Color.rgb = trunc((A - B) * C + D);
 


refractionpcsx2:
Musashi Samurai Warriors is still banding.
SotC seems underbright.

tellowkrinkle:
Can you send a screenshot?  It looks like this for me:
<details>
<summary>Before</summary>

image
</details>
<details>
<summary>After</summary>

image
</details>
<details>
<summary>Maximum Blending</summary>

image
</details>

tellowkrinkle:
And I don't expect SotC to look any different from Master

The new and old blend mix equations should give the same result for the bloom accumulation on SotC

refractionpcsx2:
sure

Here's master:
image

and the PR:
image

Here's my settings, if it matters
image

tellowkrinkle:
Just in case, can you wipe your shader cache to make sure it's not having issues there?
Also test other renderers to see if it's everywhere or not.

refractionpcsx2:
I always extract PR's in to entirely new folders, the only thing I use from a backup in the WX builds is ini files for folders and default graphics settings, all resources/cache are built from the PR.

Zombeaver:
It doesn't seem to be making a difference in my test cases.

Shadow Man:

Minimum:

image

Basic:

image

Fatal Frame 3:

Minimum:

image

Basic:

image

lightningterror:
Fatal Frame 3 should look darker, there are some slight differences tho.

Minimum:
image

Basic:
image

Maximum:
image

Software:
image

Zombeaver:
I know it's supposed to be darker, I'm just pointing out that there's banding in basic whereas there isn't with minimum, just like with Shadow Man, and all the other cases. With pure hardware (minimum) or pure software (maximum) it doesn't happen. As I said earlier contrast is generally higher with Basic+

tellowkrinkle:
> Edit: humoured you anyways, and it was still banding :(

I want to make sure you're hitting the new shader modifications.  Test the new build on Vulkan.  It should make Musashi a completely black screen.  If you see any image at all in Vulkan, something's very broken.

refractionpcsx2:
Sure, as soon as it's built, I'll give her a spin, I'll do exactly what I do for setting up any new test PR

lightningterror:
Complete black screen here.

tellowkrinkle:
> Complete black screen here.

And did you still have banding on the previous commit?

lightningterror:
Master:
image

PR:
image

refractionpcsx2:
okay, extracted and copied inis *exactly* the same way as the first revision of the PR, aaaaaaaand

image

tellowkrinkle:
Did some testing on how various GPUs round results going through blending.
Looks like Nvidia converts to 8-bit integers between the shader and blending (pretty much what we were doing before with the floor), so it will never be able to do blend_mix well.

AMD and Intel have some error, meaning that the optimal value is slightly different than what I originally put there (120/256 is optimal for AMD, 126/256 for Intel), but we should be able to get generally good results out of those two.

[Results for alpha 10/128 (Musashi's blend) in a spreadsheet](https://docs.google.com/spreadsheets/d/1WLheNwUK9U3UjL38GIExX1TbBNfzLf1ikDpzwnUq-uo/edit?usp=sharing) (X is output color, Y is input dst color, cells are the minimum shader output required to achieve the given output color)
[Results for all alpha values](https://github.com/PCSX2/pcsx2/files/9305437/AlphaBlendResults.zip) (view included README.txt for details)

<details>
<summary>C program to calculate optimal offsets for a given GPU</summary>

```c
#include <stdio.h>
#include <stdlib.h>

struct TestData {
	float data[129][256][256];
};

	if (argc <= 1) {
		fprintf(stderr, "Usage: %s gpuData.bin\n", argv[0]);
		return EXIT_FAILURE;
	}
	FILE* handle = fopen(argv[1], "r");
	if (!handle) {
		fprintf(stderr, "Failed to open %s\n", argv[1]);
		return EXIT_FAILURE;
	}
	struct TestData* file = (struct TestData*)malloc(sizeof(struct TestData));
	if (fread(file, sizeof(*file), 1, handle) != 1) {
		fprintf(stderr, "Failed to read %s\n", argv[1]);
		return EXIT_FAILURE;
	}
	fclose(handle);

	const int total = 129 * 256 * 256;
	int resLo[769] = {};
	int resHi[769] = {};
	for (int alpha = 0; alpha <= 128; alpha++) {
		for (int dst = 0; dst < 256; dst++) {
			for (int src = 0; src < 256; src++) {
				int ps2 = (((src - dst) * alpha) >> 7) + dst;
				float blend_mix = src * alpha * (1.f/128.f);
				float loCutoff = file->data[alpha][dst][ps2] * 255.f;
				float hiCutoff = ps2 == 255 ? 255.f : file->data[alpha][dst][ps2 + 1] * 255.f;
				float loAmt = blend_mix - loCutoff + 1.f;
				float hiAmt = blend_mix - hiCutoff + 1.f;
				int loIdx = loCutoff == 0 || loAmt >= 3.f ? 768 : loAmt < 0.f ? 0 : (int)(loAmt * 256);
				int hiIdx = hiCutoff >= 255.f || hiAmt < 0.f ? 0 : hiAmt >= 3.f ? 768 : (int)(hiAmt * 256);
				resLo[loIdx]++;
				resHi[hiIdx]++;
			}
		}
	}
	for (int i = 768; i > 0; i--) {
		resLo[i - 1] += resLo[i];
		resHi[i - 1] += resHi[i];
	}
	int best = 0;
	int bestVal = resLo[0] - resHi[0];
	for (int i = 0; i <= 768; i++) {
		int inRange = resLo[i] - resHi[i];
		if (inRange >= bestVal) {
			best = i;
			bestVal = inRange;
		}
		printf("Subtracting %4d/256 would be correct for %2d%% of values (%7d/%d)\n", i - 256, inRange * 100 / total, inRange, total);
	}
	printf("For this GPU, the best value to use would be %d/256, which is correct for %2d%% of values (%d/%d)\n", best - 256, bestVal * 100 / total, bestVal, total);
	return 0;
}
```
</details>

Zombeaver:
Hmm, that's interesting. Any ideas for how best to deal with Nvidia in that case?

tellowkrinkle:
Nvidia will have to use higher blend modes to avoid banding
There's nothing we can do to make blend_mix more accurate on Nvidia
Since no matter what we do in the shader, the GPU will do the equivalent of `round(shaderOut)` before running the hardware blend
So the only options we have that don't double-round are full software blend and full hardware blend

On the bright side, a lot of Nvidia GPUs run higher blend modes fairly well.

Zombeaver:
So why is it that with 1.6.0 this issue wasn't present on any blending level? Was it always pure hardware at that point? Because I went back and checked there and I can't replicate the behavior here in any of them at any blending level.

I have an RTX 3080 and basically all of the problem cases that I'm reporting are unplayably slow at maximum, which they require in order for them to display without banding other than with minimum. To any end user, the result of what you have currently is going to be a noticeably worse visual quality in some situations with the same or even higher settings they were using previously. I'm not sure that telling people "Oh you'll need to use maximum to prevent that, which is unplayable even on a 3080" is exactly a great solution.

refractionpcsx2:
1.6 didn't have BLEND_MIX, that's why, it forced software mode, which is what the higher blending level does.

Zombeaver:
Just out of curiosity I went back and did some testing with maximum and at times when performance takes a complete nosedive to around 50% speed it's not even pushing the GPU or CPU to any extreme degree when it comes to utilization. GS is sitting at about 10% and EE and VU are at about 15%. In the case of Shadow Man GS is sitting at about 2% while the game is running anywhere between 30% to 60% speed.

Zombeaver:
> 1.6 didn't have BLEND_MIX, that's why, it forced software blend, which is what the higher blending level does.

So then why can't that just be done here? Why does it suddenly require maximum to take advantage of that, tanking performance horribly? If it was software blend across the board before I don't know why that couldn't be used here. Whatever impact that, on its own, had on performance seems completely negligible.

tellowkrinkle:
Yep, always pure hardware

Which had bloom issues in SotC that were so bad that we had dedicated code in GS to detect the bloom draw and skip it.

At least here it's just some banding.  The sky doesn't look like it's melting or anything.  And it's only really an issue for one generation of one manufacturer's GPUs, since Nvidia's 2000 series should have no trouble with Maximum blending there.  It's just the 3000 series that has trouble.  (And all of AMD, but this PR should fix it for AMD)

refractionpcsx2:
Out of curiosity, I ran Shadowman 2 on my 2080ti to see the performance impact.

Maximum blend = 105fps
Basic = 535fps

So yeah, it's pretty chonky to do that much SW blending lol

Zombeaver:
> it's only really an issue for one generation of one manufacturer's GPUs, since Nvidia's 2000 series should have no trouble with Maximum blending there. It's just the 3000 series that has trouble.

Why are you assuming that this is the only place where it is/will ever be an issue? Doesn't it seem like a safer bet to assume that if it's an issue on 3000 cards that it will continue to be an issue going forward? Calling Nvidia "one manufacturer" in the GPU space seems a bit intentionally obtuse to me. If there is an issue on current hardware, not aging defunct hardware mind you, that strikes me as problem. The fact that it's slowing to a crawl while barely even utilizing the hardware available makes it pretty clear that something is wrong.

> The sky doesn't look like it's melting or anything.

image

I dunno man. I'd actually say that's actually a pretty apt description.

refractionpcsx2:
Dude, what do you expect us to do? Nvidia is fucking us, if you don't like it use minimum blending and it'll act like 1.6 in these games.

Zombeaver:
> The only thing I can suggest, is I don't think this game affected need any fancy blending, so you can play them on minimum blending, which will use hardware and not have the perf impact.

> Nvidia is fucking us, if you don't like it use minimum blending and it'll act like 1.6 in these games.

Well yeah, that's exactly what I've been doing in all these cases because it's effectively my only option.

> Dude, what do you expect us to do?

I don't know man, I'm not a dev. I reported it, because I was asked to report it, because the behavior seemed problematic. So now if the final response to that is to say "nothing that can be done" that's certainly your prerogative. I think it's shortsighted to assume that this is just going to magically resolve itself on GPUs in the future though.

I'm just trying to speak on the behalf of the end user here. That's all I'm trying to stress. The reality is that the average user doesn't know about or care about any of this. The only thing they're going to care about is the fact that situation X looked one way before and now looks worse than it did. This is on default/recommended values, mind you.

refractionpcsx2:
Look, I know you're frustrated because it doesn't do what you want it to do, but honestly this isn't an issue we can resolve.

Blend mix was introduced because hardware is inaccurate on its own and software is slow on its own (as you see in max blending), blend mix made it so most blending can be mixed with both and still be fast and accurate.

Were not going to remove blend mixing when it works so well, that would regress us back to a worse step. Minimum blending in these games will act like 1.6 if, that's your benchmark nvidia will handle that fine.

I'm sorry you're upset at us because nvidia made poor choices, but there's nothing we can do.

Zombeaver:
> Look, I know you're frustrated because it doesn't do what you want it to do, but honestly this isn't an issue we can resolve.
> 
> Blend mix was introduced because hardware is inaccurate on its own and software is slow on its own (as you see in max blending), blend mix made it so most blending can be mixed with both and still be fast and accurate.
> 
> Were not going to remove blend mixing when it works so well, that would regress us back to a worse step. Minimum blending in these games will act like 1.6 if, that's your benchmark nvidia will handle that fine.
> 
> I'm sorry you're upset at us because nvidia made poor choices, but there's nothing we can do.

I'm not upset, I'm really not. If I'm coming across that way, I'm sorry. I'm too old to get bent out of shape about old video games. I'm just trying to do whatever I can to push things in a positive direction. That's really all. If my only option is minimum and the end result is equivalent to 1.6 (which I _think_ it is, but would need to take a closer look at that) then so be it. Though keep in mind that minimum is not the recommended/default value and there will presumably be other users that have 3000 cards, so this may be somewhat counterintuitive depending on the situation.

> Yep, always pure hardware

If this is the case, what was the actual difference in blending steps previously? If there was some granularity in blending accuracy while always being pure hardware would there not be the potential to maintain some degree of that here? i.e. could there be a "more accurate" level of pure hardware above minimum?

refractionpcsx2:
minimum will be as accurate as 1.6 was on those games. the choice you have is use pure hardware (1.6/minimum) or pure software (maximum), there's no real inbetween.

Zombeaver:
> minimum will be as accurate as 1.6 was on those games

As accurate as _what_ in 1.6 though, is what I'm asking. Maximum? There was a delineation of accuracy steps before, when it was pure hardware. So now that there isn't (for pure hardware), is 1.7 minimum the same as 1.6 maximum or something lower? If it really is the equivalent of maximum, then great. I have a list of games that required higher blending accuracy for various elements in 1.6, so I can compare those against minimum in 1.7.

refractionpcsx2:
As accurate as that game was in Basic in 1.6, in these cases they don't make use of any fancy blending (that I know of) so minimum will suffice.  Basic in 1.6 generally used HW blending for most things, where in 1.7 it will use BLEND a lot more, but this game only uses the basic blend equation as far as I'm aware, which is the same in both basic and minimum, and in 1.7 minimum will make it use hardware.

if you have a game that needs basic or higher and has the banding problem, then your only option is maximum blending, sorry.

Zombeaver:
> if you have a game that needs basic or higher and has the banding problem, then your only option is maximum blending, sorry.

That's pretty unfortunate.

And yeah I just confirmed for myself that 1.7 minimum is essentially 1.6 basic. I tested on Rogue Galaxy that normally has gray text in places that's supposed to be white, and requires medium blending. It was gray on minimum.

To be honest I wish the whole hardware/software blending thing could be completely divorced from the rest of the blending accuracy steps. That would would resolve the issue. There are clearly multiple factors in the accuracy steps beyond hardware vs software mix and I wish they weren't tied at the hip like that.

refractionpcsx2:
> To be honest I wish the whole hardware/software blending thing could be completely divorced from the rest of the blending accuracy steps. That would would resolve the issue. There are clearly multiple factors in the accuracy steps beyond hardware vs software mix and I wish they weren't tied at the hip like that.

the hardware/software steps ARE The blending. As I've been trying to explain to you there is only 3 paths, do it in hardware (and it'll be no better than minimum, pretty much), do it in software (like maximum, speed included) or a mix of both, which is what we do.  There is no magical backdoor to get around all this.

It sucks, I get it, I don't like it either (I'm also an nvidia user) but we've just gotta deal, or get an AMD/Intel GPU.

Zombeaver:
> the hardware/software steps ARE The blending.

They didn't used to be though, is what has been established. There was a delineation of accuracy steps _entirely within pure hardware_. And now there's no access to that. All that's there for pure hardware is 1.6 basic. Pure hardware before could be more accurate, as evidenced by my example of Rogue Galaxy. It's not about a magic door, it's just options that do not exist (via the UI at least).

refractionpcsx2:
I feel like you're missing the point. it used to use hardware blending in these cases, the inaccurate but fast method, it doesn't use it anymore, at least not exclusively. the problem now is it's using software and hardware, but Nvidia made the stupid decision to round the shader results before the hardware blend happens, messing the result.

The only thing we could do is reimplement the old hardware blending table separately and do "if blend mix and nvidia, do this shitty blend instead" but the downsides to that are

1. A bunch of work to basically regress our code
2. Blending accuracy will be worse in a majority of cases
3. More code bloat

and 1 and 2 are not things I'd want, personally, as an Nvidia user.

Zombeaver:
> The only thing we could do is reimplement the old hardware blending table separately and do "if blend mix and Nvidia, do this shitty blend instead" but the downsides to that are
> 
> 1. A bunch of work to basically regress our code
> 2. Blending accuracy will be worse in a majority of cases
> 3. More code bloat
> 
> and 1 and 2 are not things I'd want, personally, as an Nvidia user.
I'm not suggesting that / wouldn't want that either. I don't think it would be necessary to overcomplicate things like that anyway.

In 1.6 there were multiple accuracy steps that all sat within the realm of pure hardware. There is only a single pure hardware option now, which is the equivalent of old basic in 1.6. I'm not asking for or suggesting some kind of automated detection and adjustment based on hardware, that's totally overboard. What I'm saying is that there should be a way to access the higher accuracy features used in 1.6 blending while only using pure hardware blending in 1.7. More accurate pure hardware blending existed before, I don't know why it couldn't now. There is obviously more involved in the accuracy steps than _just_ hardware/software mix because otherwise there wouldn't have been any point in having different steps in 1.6, where it was always pure hardware anyway.

From a UI perspective the easiest thing to do would be to just include a single checkbox that said something like "force pure hardware blending" or something and then the accuracy steps functioned just like they used to. Alternately they could just be added as alternate steps to the blending accuracy list, but I don't think that's good from a UI perspective.

refractionpcsx2:
Dude, I jsut said we could redo all the old hardware blend settings, said it would suck for 3 reasons, you said you aren't suggesting that, then proceed to suggest that...

There's one hope for the future for Nvidia cards (900 series onward) and that's ROV, which is kind of a software blend but we have more control, and can hopefully be faster without losing accuracy.

Zombeaver:
I don't think I did but okay. Sorry to bother you, I tried.

PR #6810, opened 2022-08-09
GS-hw: Further improve how we handle blend mix 1.

lightningterror:
### Description of Changes

Compensate slightly for Cd*(As + 1) - Cs*As. Try to compensate a bit with substracting 1 (0.00392) * (Alpha + 1) from Cs. The initial factor we chose is 1 (0.00392) as that is the minimum color Cd can be, then we multiply by alpha to get the minimum blended value it can be.

### Rationale behind Changes

Improves accuracy slightly, helps MGS3 on Direct3D.
### Suggested Testing Steps

Test mgs3 and any other games, logs will trigger if the games use it.

### DIFF
--- a/bin/resources/shaders/dx11/tfx.fx
+++ b/bin/resources/shaders/dx11/tfx.fx
@@ -778,8 +778,9 @@ void ps_blend(inout float4 Color, inout float As, float2 pos_xy)
 
 		// As/Af clamp alpha for Blend mix
 		// We shouldn't clamp blend mix with clr1 as we want alpha higher
+		float C_clamped = C;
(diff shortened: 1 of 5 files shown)


lightningterror:
Some comparisons, note changes will only be visible on Direct3D since Opengl/Vulkan will use sw blending.
Basic blending.

Zombie Hunters:
Master:
image

PR:
image

OpenGL for reference:
image


Haunting Ground 2:
Master:
image

PR:
image

OpenGL for reference:
image


MGS3:
Master:
image

PR:
image

OpenGL for reference:
image

PR #6817, opened 2022-08-10
CDVD: SCMD 03:FD, 03:EF, 2D, 2E, 36, 3E made more accurate

AKuHAK:
### Description of Changes

Currently, some SCMD behavior doesn't consider differences between mechacon revision. This PR aim is to make some SCMDs more accurate.

### Rationale behind Changes

This PR will improve accuracy without affecting performance.

### Suggested Testing Steps

All information was gathered by testing on real hardware on various different mechacon versions. For reference checks, someone can try it on real hardware with an application (sources also available) available here https://www.ps2-home.com/forum/viewtopic.php?f=127&t=8849


refractionpcsx2:
A couple of things to make note of and potentially address.

Firstly instead of checking the mechacon version every time it does an scmd, maybe just check it on cdvdreset, so it only needs to check it once.

Secondly, not everybody will have a .mec file so PCSX2 makes up its own version, but the available functions may not be correct for the BIOS being run, so it might be better to have a table of default versions for MEC generation also.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2247,21 +2249,115 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
...
+									case 0x0E:
+									case 0x0F:
+										memcpy(cdvd.SCMDResult, &MRenewalDate[7], 6);
+										break;
+									default:
+										SetSCMDResultSize(1);
+										cdvd.SCMDResult[0] = 0x80;
+										break;
+								}
+								break;
```
refractionpcsx2, line 2302: Could surely be shortened to just do 

`memcpy(cdvd.SCMDResult, &MRenewalDate[temp_mechaver[2] >> 1], 6);`

You'd just need a special case for 0x6-0x7 and checking it's less than 0x10 (though realistically, will it ever be higher?)

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2247,21 +2249,115 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
...
+									case 0x0C:
+									case 0x0D:
+										memcpy(cdvd.SCMDResult, &MRenewalDate[13], 6);
+										break;
+									default:
+										SetSCMDResultSize(1);
+										cdvd.SCMDResult[0] = 0x80;
+										break;
+								}
+								break;
```
refractionpcsx2, line 2337: as above but +8

pcsx2/CDVD/CDVD_internal.h
```diff
@@ -288,3 +288,21 @@ static u8 biosLangDefaults[8][16] =
...
+	{0x00, 0x03, 0x10, 0x08, 0x22, 0x00,}, // 5.10, 5.11
+	{0x00, 0x03, 0x11, 0x20, 0x20, 0x05,}, // 5.12, 5.13
+	{0x00, 0x04, 0x08, 0x25, 0x02, 0x12,}, // 5.14, 5.15
+	{0x00, 0x04, 0x06, 0x21, 0x09, 0x53,}, // 6.0, 6.1
+	{0x00, 0x04, 0x07, 0x20, 0x09, 0x04,}, // 6.2, 6.3
+	{0x00, 0x04, 0x12, 0x10, 0x01, 0x30,}, // 6.4, 6.5
+	{0x00, 0x05, 0x04, 0x27, 0x09, 0x17,}, // 6.6, 6.7
+	{0x00, 0x06, 0x03, 0x29, 0x06, 0x48,}, // 6.10, 6.11
+	{0x00, 0x07, 0x01, 0x29, 0x03, 0x47,}, // 6.12, 6.13
+};
```
refractionpcsx2, line 308: missing carriage return at the end of file

pcsx2/CDVD/CDVD.cpp
AKuHAK: wow thanks for C tip :)

AKuHAK: yes 0x6, 0x7 is a special mechacon (5.6) that was released in different variations for 2 regions: Mexico and other regions. As Mexico is region 07 it is enough to test region byte in mechacon response.

AKuHAK:
> Firstly instead of checking the mechacon version every time it does an scmd, maybe just check it on cdvdreset, so it only needs to check it once.

ill check that function.



> Secondly, not everybody will have a .mec file so PCSX2 makes up its own version, but the available functions may not be correct for the BIOS being run, so it might be better to have a table of default versions for MEC generation also.

It sounds not very reasonable, cause there is no direct connection between bios version and mechacon version. For example bios version 0170 (FAT 50k) can have any mechacon from the list: 5.0, 5.2, 5.4. The same is correct in other direction: 5.10 mechacon can be in 0170 bios or in 0190 bios. I am not sure if it is possible to generate correct mechacon version on bios version basis.

refractionpcsx2:
> > Firstly instead of checking the mechacon version every time it does an scmd, maybe just check it on cdvdreset, so it only needs to check it once.
> 
> ill check that function.
> 
> > Secondly, not everybody will have a .mec file so PCSX2 makes up its own version, but the available functions may not be correct for the BIOS being run, so it might be better to have a table of default versions for MEC generation also.
> 
> It sounds not very reasonable, cause there is no direct connection between bios version and mechacon version. For example bios version 0170 (FAT 50k) can have any mechacon from the list: 5.0, 5.2, 5.4. The same is correct in other direction: 5.12, 5.13 mechacon can be in 0170 bios or in 0190 bios. I am not sure if it is possible to generate correct mechacon version on bios version basis.

sure, but you could do a safe guess? Or if the BIOS version doesn't matter, then is there any real need to restrict the mechacon at all?

AKuHAK:
> sure, but you could do a safe guess? Or if the BIOS version doesn't matter, then is there any real need to restrict the mechacon at all?

I can try to do that, but this will require a lot of time to check all my charts for that info. Are you sure this is necessary inside this pr?

As for mechacon restriction - this was made more for accuracy, so pcsx2 will behave more like the real machine. As for games, from their side mechacon isn't very involved, cause the game should work with any mechacon and any bios. But still, some games are checking the mechacon version (for an unknown reason), so in theory, more accurate behavior can improve compatibility.

refractionpcsx2:
Well, if you want to do it right, do it as right as you can, which means yes it is necessary on this PR, since that's what you were aiming for, right?

I understand you're making it more like the real machine, I'm fine with that, but there are situations where people don't have a dumped MEC, and we use a predetermined version number, which won't match up with your changes.

pcsx2/CDVD/CDVD.cpp
AKuHAK: 5.8, 5.9, 6.8, 6.9 does not exist. I  added 0x80 response into enumartor

AKuHAK:
PR is now ready for review. I changed the Bios region to match the MagicGate region. Also, realized, that `BiosZoneBytes` has no real use anymore.
Added guessed mechacon version based on bios version if mechacon isnt dumped.

refractionpcsx2:
Alright, I'll look it over at some point.

The Qt interface uses BiosZoneBytes for the BIOS list.

AKuHAK:
> The Qt interface uses BiosZoneBytes for the BIOS list.

Ah, I see. It was implemented in https://github.com/PCSX2/pcsx2/pull/5778. I would like to mention that SCMD 36 does that on real hardware. If you change NVRAM RegionParam first byte then memory is patched and `rom0:ROMVER` will return a different region letter. That basically means that all Deckard units (75k+) share the same rom0, rom1, erom. With SCMD 36 the required areas of memory are patched according to nvram RegionParam area.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2243,7 +2243,34 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
 						memcpy(cdvd.SCMDResult, &temp_mechaver[0], 4);
 						break;
 
-					// case 0x01: // supported by PS3 CECH-A ps2emu
+					case 0x01: // get DSP version (1:3), supported by PS3 CECH-A ps2emu
+						SetSCMDResultSize(3);
+						cdvd.SCMDResult[0] = 0;
+						if ((temp_mechaver[1] == 3) && (temp_mechaver[2] == 9)) // PS3 CECH-A ps2emu special case
```
refractionpcsx2, line 2249: Delete this comment please.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2243,7 +2243,34 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
 						memcpy(cdvd.SCMDResult, &temp_mechaver[0], 4);
 						break;
 
-					// case 0x01: // supported by PS3 CECH-A ps2emu
+					case 0x01: // get DSP version (1:3), supported by PS3 CECH-A ps2emu
```
refractionpcsx2, line 2246: Delete the second part of this comment please.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2226,42 +2238,131 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
...
+							else if ((temp_mechaver[1] == 3) && (temp_mechaver[2] >= 6)) // 0160 with mechacon >= 3.6
+								cdvd.SCMDResult[1] = 3;
+							else
+								cdvd.SCMDResult[1] = 2; // 0160 with mechacon < 3.6
+							cdvd.SCMDResult[2] = 0; // ??
+						}
 						break;
-					case 0x30:
+
+					case 0x30: // supported by PS3 CECH-A ps2emu
```
refractionpcsx2, line 2275: delete this comment please.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2226,42 +2238,131 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
...
 					case 0x44: // write console ID (9:1)
 						SetSCMDResultSize(1);
 						cdvdWriteConsoleID(&cdvd.SCMDParam[1]);
+						cdvd.SCMDResult[0] = 0; // returns 0 on success
 						break;
 
-					case 0x45: // read console ID (1:9)
-						SetSCMDResultSize(9);
-						cdvdReadConsoleID(&cdvd.SCMDResult[1]);
+					case 0x45: // read console ID (1:9), blocked on PS3 CECH-A ps2emu
```
refractionpcsx2, line 2288: delete second part of this comment please.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2226,42 +2238,131 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
...
+						}
+						else
+						{
+							SetSCMDResultSize(9);
+							cdvdReadConsoleID(&cdvd.SCMDResult[1]);
+							cdvd.SCMDResult[0] = 0; // returns 0 on success
+						}
 						break;
 
+					// case 0x48: // supported by PS3 CECH-A ps2emu
```
refractionpcsx2, line 2302: delete this comment please.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2290,7 +2391,7 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
 					cdvd.SCMDResult[0] = cdvdCtrlTrayClose();
 				break;
 
-			case 0x08: // CdReadRTC (0:8)
+			case 0x08: // CdReadRTC (0:8), supported by PS3 CECH-A ps2emu
```
refractionpcsx2, line 2394: delete this comment please.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2324,22 +2425,31 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
 				//memcpy((u8*)&cdvd.RTC, cdvd.Param, 7);
 				break;
 
-			case 0x0A: // sceCdReadNVM (2:3)
-				address = (cdvd.SCMDParam[0] << 8) | cdvd.SCMDParam[1];
-
-				if (address < 512)
+			case 0x0A: // sceCdReadNVM (2:3), blocked on PS3 CECH-A ps2emu
```
refractionpcsx2, line 2428: Delete this comment please.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2365,7 +2476,7 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
 				//			break;
 
 
-			case 0x0F: // sceCdPowerOff (0:1)- Call74 from Xcdvdman
+			case 0x0F: // sceCdPowerOff (0:1)- Call74 from Xcdvdman, supported by PS3 CECH-A ps2emu
```
refractionpcsx2, line 2479: delete the second part of this comment please.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2374,12 +2485,11 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
 #endif
 				break;
 
-			case 0x12: // sceCdReadILinkId (0:9)
+			case 0x12: // sceCdReadILinkId (0:9), supported by PS3 CECH-A ps2emu
```
refractionpcsx2, line 2488: delete the second part of this comment please

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2413,22 +2525,24 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
 				cdvd.SCMDResult[0] = 0;
 				break;
 
-			case 0x17: // CdReadModelNumber (1:9) - from xcdvdman
+			case 0x17: // CdReadModelNumber (1:9) - from xcdvdman, supported by PS3 CECH-A ps2emu
```
refractionpcsx2, line 2528: delete the second part of this comment please.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2596,14 +2772,15 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
 				cdvd.SCMDResult[0] = 0;
 				break;
 
-			case 0x41: // CdReadConfig (0:16)
+			case 0x41: // CdReadConfig (0:16), supported by PS3 CECH-A ps2emu
```
refractionpcsx2, line 2775: delete the second part of this comment please.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2637,7 +2814,7 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
 				cdvd.SCMDResult[0] = 0;
 				break;
 
-			case 0x84: // secrman: __mechacon_auth_0x84
+			case 0x84: // secrman: __mechacon_auth_0x84, supported by PS3 CECH-A ps2emu
```
refractionpcsx2, line 2817: delete the second part of this comment please.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -2656,7 +2833,7 @@ static void cdvdWrite16(u8 rt) // SCOMMAND
 				cdvd.SCMDResult[12] = 0x9b;
 				break;
 
-			case 0x85: // secrman: __mechacon_auth_0x85
+			case 0x85: // secrman: __mechacon_auth_0x85, supported by PS3 CECH-A ps2emu
```
refractionpcsx2, line 2836: delete the second part of this comment please.

AKuHAK: Could you, please, provide more information about why PS3 is restricted to mention?

refractionpcsx2: Because filling our code with "We read how to do this with sony's ps3 emu" is a really fucking dumb idea.

refractionpcsx2:
In fact, I'm sorry, this code is tainted. You've obviously stolen it from Sony. we can't accept this.

AKuHAK:
What? I never stole any code. All data was taken from real hardware by running an elf that is mentioned in first post.

AKuHAK:
That includes ps3, I just run an elf and grab the results that it shows to me.

refractionpcsx2:
Do you *really* think taking stuff from sony code (which you obviously have because you reference it directly) is a good idea to put in our emulator? If you do, you're an absolute moron.

Anyway, you've sullied it by referencing sony's stuff, so I cannot accept this, I'm not getting in to potential legal issues because of your clumsyness.

Florin9doi:
I wonder where all the existing comments and all these sce** function names which were included for 15 years come from ?
https://github.com/PCSX2/pcsx2/blob/6f7890b709d5e3f7f5b824781e493455efc92339/pcsx2/IopModuleNames.cpp#L21

refractionpcsx2:
Open PS2 SDK for one https://github.com/ps2dev/ps2sdk/blob/bb8702287e4ad58094da93d83c76bf178b1921a1/common/include/libcdvd-common.h

anyway, they are just SDK names, they were easy to find even many years ago, but nobody was stupid enough to go "ps3 emu from sony does this" which is basically what you did.

Anyway, conversation over.

AKuHAK:
As regarding ps3, I own CECH-A variant of it. I opened it , take the pictures and realized that there is no mechacon chip. There is only EE chip on the ps2 board, and GS chip, but no Mechacon chip. So I made an assumption that mechacon chip is emulated. @refractionpcsx2 please tell me where is here "sony stolen code"? Then I run mechacon tester - simple app that consist of few code lines, and that just allows to run any scmd with any param and shows the result. Then I checked all SCMDs and their output, and realized that only 8 scmds persist on ps3. I never take any code, my ps3 is running on official firmware, i didnt dig in any sony files for that information. If my comments are breaking some rules, I can remove them.

The same applied to all other submissions in these PR. All the information is taken from actual hardware tests, made with that small app. Anybody can reproduce all the behaviour I made. I didnt own all these hardware, so some tests were made by community, but everything was gathered from tests, and not from "some stolen sony code".

refractionpcsx2:
You don't get it, you basically labelled everything as "this is what the ps3 emu does", which is basically saying "I made sure we're doing the same thing Sony do in their emu" which is basically taking their code, and then you had the nerve to label it so in OUR code. I cannot accept that.

If you've reversed code yourself from the mechacon, that's fine, no problem there, but when you start adding/labelling stuff as the PS3 emu, we have a problem, and I can't trust that what is there you haven't just ripped from the ps3 emu, because you broke my trust with your actions.

AKuHAK:
By PS3 EMU I meant that they dont use real mechacon, but the emulated one. I didnt know that this acronym can lead to so much problems, but ok, its fine. Hope oneday some "trusted" contributor can reproduce at least part of my submissions, i will update the first post with detailed instructions (later) of how to achieve this result.

refractionpcsx2:
@Shizmob that's my problem, he labelled a whole bunch of stuff "ps3 emu" as to imply that's what the PS3 emu does (so looked at sonys code and basically copied it in to ours), That's what I took issue with. I have no problem with reverse engineering the mechacon.

PR #6823, opened 2022-08-12
Config: Default dump compression to LZMA (xy)

JordanTheToaster:
### Description of Changes
Changes default dump compression from uncompressed to LZMA (xy).

### Rationale behind Changes
When asking for a GS dump when providing support the dumps are sometimes so large they cannot be sent over Discord and you have to juggle it around on different platforms the simple answer is to have the default compression set to LZMA to greatly reduce the dump size and make it possible to send most dumps over Discord.

### Suggested Testing Steps
Make sure CI is happy.


BreakPoints:
Quick aside, GitHub doesn't let you attach .gs.xz files. In these cases they would still have to be zipped.

JordanTheToaster:
> Quick aside, GitHub doesn't let you attach .gs.xz files. In these cases they would still have to be zipped.

This was the case with WX as that used xy by default so it's not much of an issue.

Mrlinkwii (approved):

PR #6825, opened 2022-08-12
GS: Improve autoflush detection.

refractionpcsx2:
### Description of Changes
Improves autoflush detection and fixes a bug with splitting Triangle Fan draws.

### Rationale behind Changes
Old method didn't work properly for triangle fans, overall system needed improving to check rect by rect to avoid false hits by games drawing in a L shape.  

Also there was a bug where we save the last few verts if a flush happens, but because triangle fans rely on the first vert to always be there, the xy buffer (used to calculate skips etc) had incorrect information, causing it to falsely skip draws.

### Suggested Testing Steps
Test games, especially those with Autoflush.

Performance is a mixed bag, Jak 2 in a heavy autoflush spot is about 12% slower, Burnout 3 is about 10% faster (in a GS Dump), and some other games vary slightly in terms of number of draw calls.

Edit: Performance loss in Jak 2 has been mitigated (and increased over master) with [GS: Add shortcut for Autoflush on Tri-Strip/Fan](https://github.com/PCSX2/pcsx2/pull/6825/commits/b9ed7b5773a3a53f68d8125db9e1ec84bcf76f8b) .  It *should* be safe enough.

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -2863,6 +2863,10 @@ GSState::PRIM_OVERLAP GSState::PrimitiveOverlap()
 
 __forceinline void GSState::HandleAutoFlush()
 {
+	// Kind of a cheat, making the assumption that 2 consecutive fan/strip triangles won't overlap each other (*should* be safe)
+	if ((m_index.tail & 1) && (PRIM->PRIM == GS_TRIANGLESTRIP || PRIM->PRIM == GS_TRIANGLEFAN))
+		return;
+
 	const u32 frame_mask = GSLocalMemory::m_psm[m_context->TEX0.PSM].fmsk;
 	const bool frame_hit = (m_context->FRAME.Block() == m_context->TEX0.TBP0) && !(m_context->TEST.ATE && m_context->TEST.ATST == 0 && m_context->TEST.AFAIL == 2) && ((m_context->FRAME.FBMSK & frame_mask) != frame_mask);
 	// There's a strange behaviour we need to test on a PS2 here, if the FRAME is a Z format, like Powerdrome something swaps over, and it seems Alpha Fail of "FB Only" writes to the Z.. it's odd.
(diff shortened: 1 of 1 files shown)


Mrlinkwii (approved):

pcsx2/GS/GSState.cpp
```diff
@@ -2952,34 +2982,69 @@ __forceinline void GSState::HandleAutoFlush()
...
-					Flush();
+				// If the draw was 1 line thick, make it larger as rects are exclusive of ends.
+				if (tex_rect.x == tex_rect.z)
+					tex_rect += GSVector4i(0, 0, 1, 0);
+				if (tex_rect.y == tex_rect.w)
+					tex_rect += GSVector4i(0, 0, 0, 1);
+
+				GSVector4i old_tex_rect = GSVector4i(0, 0, 0, 0);
+				GSVector2i offset = GSVector2i(m_context->XYOFFSET.OFX, m_context->XYOFFSET.OFY);
+				GSVector4i scissor = GSVector4i(m_context->scissor.in);
```
lightningterror, line 3009: The two vectors can be const?

refractionpcsx2: yup

pcsx2/GS/GSState.cpp
```diff
@@ -2875,32 +2893,38 @@ __forceinline void GSState::HandleAutoFlush()
 	// So we need to calculate if a page boundary is being crossed for the format it is in and if the same part of the texture being written and read inside the draw.
 	if (PRIM->TME && (frame_hit || zbuf_hit) && GSUtil::HasSharedBits(frame_z_bp, frame_z_psm, m_context->TEX0.TBP0, m_context->TEX0.PSM))
 	{
-		size_t n = 1;
+		int  n = 1;
+		size_t buff[3];
+		size_t head = m_vertex.head;
+		size_t tail = m_vertex.tail;
```
lightningterror, line 2899: These two as well.

lightningterror (approved):

JordanTheToaster (approved):

PR #6836, opened 2022-08-14
Qt: Allow for weird sounds to come out of more speakers

KrossX:
### Description of Changes

The SPU setting `DplDecodingLevel` seemed to be ignored when using the Qt builds. The setting now loads but also ended up adding it to the UI for easy usage/testing. Since it's only used with the 5.1 output, the widget is disabled unless `Expansion: Surround 5.1` is selected.

### Rationale behind Changes

Boredom.
Here, have some music. [TotallyNotRickRoll](https://www.youtube.com/watch?v=6riDJMI-Y8U), [4Real](https://www.youtube.com/watch?v=nRKJBpFFsuI)

### Suggested Testing Steps

Click the setting thing, see the log that it changes accordingly. [LogCode](https://github.com/PCSX2/pcsx2/blob/ea051c6d5fcc2eddbf749664340f283c88d058e3/pcsx2/SPU2/SndOut_Cubeb.cpp#L222)

#EDIT: Template thing.

### DIFF
--- a/pcsx2-qt/Settings/AudioSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/AudioSettingsWidget.cpp
@@ -27,6 +27,7 @@
 static constexpr s32 DEFAULT_INTERPOLATION_MODE = 5;
 static constexpr s32 DEFAULT_SYNCHRONIZATION_MODE = 0;
 static constexpr s32 DEFAULT_EXPANSION_MODE = 0;
+static constexpr s32 DEFAULT_DPL_DECODING_LEVEL = 0;
 static const char* DEFAULT_OUTPUT_MODULE = "cubeb";
 static constexpr s32 DEFAULT_OUTPUT_LATENCY = 100;
 static constexpr s32 DEFAULT_VOLUME = 100;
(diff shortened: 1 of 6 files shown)


F0bes:
Even my speakers too 👉👈🥺

KrossX:
Of course, you can expect sounds like this: [surround spectrogram](https://user-images.githubusercontent.com/2485237/184519300-76699fb4-1f99-4456-938c-18bdb96d41a1.gif)

kenshen112:
Well, part of it is because it'll literally not work without expanded speakers or at least there's not really a point without that. The other is that there's no decoder lib for pro logic so if it's just noise that's probs why

KrossX:
[EqualizerAPO/Surround](https://github.com/Dogway/emulation-random/tree/master/EqualizerAPO/Surround)

These files might be interesting to play with since I don't have a DPLII decoder to hear what the game output is supposed to be. The game's audio might sound better in stereo to begin with after all.

JordanTheToaster (approved):

stenzek (approved):
Code changes LGTM

PR #6841, opened 2022-08-14
Qt: Frame Rate Control - increase maximum

Dreadmoth:
### Description of Changes
Raise maximum values to 999.99, up from [QDoubleSpinBox default maximum of 99.99](https://doc.qt.io/qt-5/qdoublespinbox.html#maximum-prop).

### Rationale behind Changes
Remove an unintentional (?) limitation, parity with wx (but still prevent input of extremely high values).
Values over 99.99 can be useful for high frame rate patches, and for games which don't break or speed up at high FPS.

### Suggested Testing Steps
Test different frame rate control values.

### DIFF
--- a/pcsx2-qt/Settings/AdvancedSystemSettingsWidget.ui
+++ b/pcsx2-qt/Settings/AdvancedSystemSettingsWidget.ui
@@ -146,6 +146,9 @@
         <property name="singleStep">
          <double>0.010000000000000</double>
         </property>
+        <property name="maximum">
+         <double>999.99</double>
+        </property>
        </widget>
       </item>
       <item row="0" column="1">
@@ -155,6 +158,9 @@
         </property>
         <property name="singleStep">
          <double>0.010000000000000</double>
+        </property>
+        <property name="maximum">
+         <double>999.99</double>
         </property>
        </widget>
       </item>


stenzek:
This is fine, but I really don't like the idea... games are definitely going to break, and we don't want to encourage people to mess with it. Although it'll break at 99 too. The intention was to hide the advanced settings panel by default.

stenzek (approved):

lightningterror:
Honestly I rather this wasn't even editable in the gui, and fast forward exists, closing for the reasons said by Sten.

RedDevilus:
And with the framerate percentages being visible for PAL and NTSC and you can pick any custom multiplier, the need is actually removed as unlocking was to make it easier in WX to do for example 120% base framerate for PAL to get 60 FPS equivalent.

Dreadmoth:
> And with the framerate percentages being visible for PAL and NTSC and you can pick any custom multiplier, the need is actually removed as unlocking was to make it easier in WX to do for example 120% base framerate for PAL to get 60 FPS equivalent.

Fast Forward and Base Framerate Adjust / Normal Speed % always result in a game running faster, whereas adjusting the NTSC / PAL framerate can (for the few games that don't speed up anyway due to frame-based timing) make games run at a higher framerate (at the same speed).

Having the setting available in the UI allows adjusting the value without having to reboot a game, which can save a bit of time if it does cause an issue (e.g. FMV audio de-sync).

It's a pretty niche, risky setting, so I understand not wanting to have to deal with support requests resulting from it.
Maybe it could reappear someday in a hidden-by-default menu, preceded by a nice [warning message](https://user-images.githubusercontent.com/8492684/196502505-c88bcc4c-5425-4d23-bfd0-882e26872a44.png).

Margen67:
a

Margen67:
> > And with the framerate percentages being visible for PAL and NTSC and you can pick any custom multiplier, the need is actually removed as unlocking was to make it easier in WX to do for example 120% base framerate for PAL to get 60 FPS equivalent.
> 
> Fast Forward and Base Framerate Adjust / Normal Speed % always result in a game running faster, whereas adjusting the NTSC / PAL framerate can (for the few games that don't speed up anyway due to frame-based timing) make games run at a higher framerate (at the same speed).
> 
> Having the setting available in the UI allows adjusting the value without having to reboot a game, which can save a bit of time if it does cause an issue (e.g. FMV audio de-sync).
> 
> It's a pretty niche, risky setting, so I understand not wanting to have to deal with support requests resulting from it. Maybe it could reappear someday in a hidden-by-default menu, preceded by a nice [warning message](https://user-images.githubusercontent.com/8492684/196502505-c88bcc4c-5425-4d23-bfd0-882e26872a44.png).

I tested some games and there were a few that didn't speed up:
Game | Notes
---- | -----
ATV Offroad Fury
ATV Offroad Fury 2
Futurama
Half-Life | 71 max FPS.
Sitting Ducks | Hangs on startup above 60 FPS. Must be set during gameplay.
Sonic Unleashed | Requires patching.
SpongeBob SquarePants: Battle for Bikini Bottom
Star Wars: Episode III – Revenge of the Sith | Can't exceed 60 FPS.
Test Drive: Off-Road Wide Open
Thrillville | Requires patching to go above 30 FPS. (falsely reports 60 FPS, outputs duplicate frames?)
Thrillville: Off the Rails | Same as Thrillville.

rpcs3 has a similar option (Vblank frequency) and is doing just fine:
image
Cxbx-Reloaded and Xenia also let you uncap the framerate.

EmilianoTalamo:
This should be merged, honestly. PCSX2 has already a plethora of settings that can break your game available for anyone to change (as with any emulator), and I found that this works fine on several games. Now it just has an arbitrary limit imposed on the GUI that can only be changed by manually editing the config file.

Setting this value to 120hz (which is needed to achieve 60fps on NTSC games that use a double-buffered vsync) is as unstable as setting it to 99hz, which is allowed on the GUI.

Maybe change the phrasing of the setting to VBlank something, so it's not assumed that you'll get better framerates for free.

PR #5518, opened 2022-02-12
GS-HW: Use integers for depth conversion shaders

stenzek:
### Description of Changes

This improves the precision when reinterpreting textures to the depth buffer, reducing round-trip error which causes depth fighting in the games listed below.

Sadly, it doesn't improve precision in rendering alone (e.g. Quake 3), since that requires more than 23 bits of precision. For what it's worth, using unrestricted depth range *and* reinterpreting the unsigned 32-bit depth value as a float does fix Quake's rendering, but this will cause problems at the upper end of the range once it hits NaNs/inf, and requires slow depth/disables early-Z (writing gl_FragDepth).

### Rationale behind Changes

Fixing bugs.
Fixes #4051
Fixes #4674
Fixes #2219
### Suggested Testing Steps

Test games with z-fighting, make sure there's no regressions in others which use format conversions.

PR #6847, opened 2022-08-15
GS: Less aggressive rounding on convert shaders

refractionpcsx2:
### Description of Changes
Reduces how much rounding is done on the colours when converted to full range.

### Rationale behind Changes
This was causing lines to appear on the floor in several Shin Megami Tensei games when upscaling (SMT3 and DDS we know of)

### Suggested Testing Steps
Try games mentioned when upscaling, but also check games mentioned in #5518

### DIFF
--- a/bin/resources/shaders/dx11/convert.fx
+++ b/bin/resources/shaders/dx11/convert.fx
@@ -164,8 +164,8 @@ PS_OUTPUT ps_convert_float16_rgb5a1(PS_INPUT input)
 }
 float ps_convert_rgba8_float32(PS_INPUT input) : SV_Depth
 {
-	// Convert a RRGBA texture into a float depth texture
-	uint4 c = uint4(sample_c(input.t) * 255.0f + 0.5f);
+	// Convert a RGBA texture into a float depth texture
+	uint4 c = uint4(sample_c(input.t) * 255.0f + 0.1f);
 	return float(c.r | (c.g << 8) | (c.b << 16) | (c.a << 24)) * exp2(-32.0f);
 }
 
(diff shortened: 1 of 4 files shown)


refractionpcsx2:
Digital Devil Saga:

Before:
image

After:
image


SMT3:
Before:
image

After:
image

tellowkrinkle:
Can you post gsdumps of SMT3 and DDS?

refractionpcsx2:
Sure, here you go :)

[SMT3.zip](https://github.com/PCSX2/pcsx2/files/9339188/SMT3.zip)
[DDS.zip](https://github.com/PCSX2/pcsx2/files/9339191/DDS.zip)

JordanTheToaster (approved):

tellowkrinkle:
image

This isn't the proper solution.
The problem is how bilinear interpolation interpolates across the parts of depth where red rolls over to zero (notice the lines in the depth).
You're just masking the problem by making it so that those bumps always round *down*, which happens to work here where the games want to render something afterwards that goes over the scene.  But what if it was the opposite, where the game wanted to render something that went *under* the scene?  Now you'd be making the problem worse instead.
The proper solution is to manually sample the four neighboring pixels and do the interpolation ourselves *after* converting to depth.

refractionpcsx2:
Okay well, that's beyond my skill level of our TC/hw renderer, so unless somebody else wants to deal with it, this is the best solution we have until somebody does (It doesn't re-break anything that used to be broken).

I will happily mark it as TODO in the shaders.

refractionpcsx2:
Superseded by #6863

PR #6852, opened 2022-08-16
GS: Add constant adjustment in blend mix when reverse subtracting.

refractionpcsx2:
### Description of Changes
Adds the constant for blend mix accuracy when using reverse subtract, as we are going in the opposite direction.
Also disable BLEND_MIX on HDR, since this actually reduces the accuracy.

### Rationale behind Changes
This was causing inaccurate blend results when reverse subtracting, causing things like shadows to break.
HDR mode requires higher accuracy due to clamping in 16bits so we need to disable 

### Suggested Testing Steps
Test stuff which regressed recently.

NOTE: These are all regressions from V1.7.3198 PR #6809, however in some cases they were issues before but the problem was masked/luckily didn't happen.

Fixes shadows in Gregory Horror Show and Pac-Man World 2 and Tomb Raider Angel of Darkness.
Disabling BLEND_MIX on HDR fixes DBZ BT3 shading problem in DX or when barriers are disabled.

Gregory Horror Show:

Before:
image

After:
image

Pac-Man World 2:

Before:
image

After:
image

DBZ BT3:

Before:
image

After:
image

Tomb Raider Angel of Darkness:
Before:
image

After:
image

### DIFF
--- a/bin/resources/shaders/dx11/tfx.fx
+++ b/bin/resources/shaders/dx11/tfx.fx
@@ -776,20 +776,22 @@ void ps_blend(inout float4 Color, inout float As, float2 pos_xy)
 
 		// As/Af clamp alpha for Blend mix
 		// We shouldn't clamp blend mix with clr1 as we want alpha higher
-		if (PS_BLEND_MIX && PS_CLR_HW != 1)
+		if (PS_BLEND_MIX > 0 && PS_CLR_HW != 1)
 			C = min(C, 1.0f);
 
 		if (PS_BLEND_A == PS_BLEND_B)
 			Color.rgb = D;
-		else if (PS_BLEND_MIX)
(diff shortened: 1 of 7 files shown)


lightningterror (changes requested):
Need to change the blend mix 1 case. Change blend mix to 2.

tellowkrinkle:
GSdumps for anyone who wants to test for themselves
[GregoryHorrorShow_BlendMixShadowsNV.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/9349576/GregoryHorrorShow_BlendMixShadowsNV.gs.xz.zip)
[PacManWorld2_BlendMixShadows.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/9349577/PacManWorld2_BlendMixShadows.gs.xz.zip)

tellowkrinkle (approved):

PR #6855, opened 2022-08-16
Qt: Change name of Adjust to Host Refresh Rate

JordanTheToaster:
### Description of Changes
Changes the name of Adjust To Host Refresh Rate to Scale to Host Refresh Rate.

### Rationale behind Changes
The name was inconsistent with the internal name and the description box text.

### Suggested Testing Steps
Make sure CI is happy.

### DIFF
--- a/pcsx2-qt/Settings/EmulationSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/EmulationSettingsWidget.cpp
@@ -65,10 +65,10 @@ EmulationSettingsWidget::EmulationSettingsWidget(SettingsDialog* dialog, QWidget
 	dialog->registerWidgetHelp(m_ui.speedLimiter, tr("Speed Limiter"), tr("checked"),
 		tr("Limits the emulation to the appropriate framerate for the currently running game."));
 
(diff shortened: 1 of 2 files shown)


refractionpcsx2:
Wouldn't something like "Scale" be a better word? "sync" sounds like vsync, which is a different option.

JordanTheToaster:
> Wouldn't something like "Scale" be a better word? "sync" sounds like vsync, which is a different option.

The internal name uses sync but I can adjust the text and description to use scale if it makes it more clear.

refractionpcsx2:
Yeah, I don't care so much about the internal names as only people who know what's going on need to know what it means, but on the UI, I'd like it top be clear and distinct for users, most of which expect us to tell them exactly what stuff does :P

JordanTheToaster:
> Yeah, I don't care so much about the internal names as only people who know what's going on need to know what it means, but on the UI, I'd like it top be clear and distinct for users, most of which expect us to tell them exactly what stuff does :P

Ask and you shall receive.

Mrlinkwii (approved):

PR #6857, opened 2022-08-16
GS: Fix wx replayer readbacks

tellowkrinkle:
### Description of Changes
Makes wx replayer not fail assertions when running dumps with readbacks

### Rationale behind Changes
It's nice when things don't crash with an assertion failure

### Suggested Testing Steps
Verify that this dump doesn't cause any assertion failures [SMT3_DepthConvertIssue.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/9354261/SMT3_DepthConvertIssue.gs.xz.zip)

### DIFF
--- a/pcsx2/gui/Dialogs/GSDumpDialog.cpp
+++ b/pcsx2/gui/Dialogs/GSDumpDialog.cpp
@@ -787,7 +787,7 @@ void Dialogs::GSDumpDialog::ProcessDumpEvent(const GSDumpFile::GSData& event, u8
 		case GSType::ReadFIFO2:
 		{
 			std::unique_ptr<u8[]> arr(new u8[*((int*)event.data) * 16]);
-			GetMTGS().InitAndReadFIFO(arr.get(), *((int*)event.data));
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
I'm sure there was reason it was using GetMTGS, probably because it runs from the EE core on Qt and of course that's where we're ultimately going.

I could be remembering incorrectly, maybe @stenzek can confirm.

tellowkrinkle:
GSDumpDialog is wx code.  It's not used by Qt.

refractionpcsx2:
I know the dialogue is, I wasn't talking about that. I was talking from which thread the dump was replayed from, I suspect on WX it's replayed from the GS thread, where the Qt version replays from the EE thread.

stenzek:
wx doesn't use the EE or GS threads, it uses its own.

refractionpcsx2:
Yeah, I figured there was a thread difference, that's why I was wondering if the GetMTGS() is required for one or the other, don't want stuff being called from the wrong thread, afterall :)

JordanTheToaster (approved):

Mrlinkwii (approved):

PR #6874, opened 2022-08-19
CMake MSVC improvements

tellowkrinkle:
### Description of Changes
- LTO more things
- PDB generation
- More optimization flags

### Rationale behind Changes
This should hopefully get CMake builds to the same level of performance as MSBuild builds

### Suggested Testing Steps
Test CMake builds and see if they're as fast as SSE4 MSBuild builds

### DIFF
--- a/.github/workflows/linux_build_matrix.yml
+++ b/.github/workflows/linux_build_matrix.yml
@@ -15,7 +15,7 @@ jobs:
     with:
       jobName: "with LTO"
       compiler: gcc
-      cmakeflags: "-DLTO_PCSX2_CORE=ON"
+      cmakeflags: "-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON"
       buildAppImage: true
     secrets: inherit
 
--- a/.github/workflows/macos_build.yml
+++ b/.github/workflows/macos_build.yml
@@ -91,7 +91,7 @@ jobs:
       - name: Generate CMake Files
         run: |
           QT_BUILD=$([ "${{ inputs.gui }}" == "Qt" ] && echo "ON" || echo "OFF")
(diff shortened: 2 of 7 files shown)


refractionpcsx2:
Did the Ratchet test

SSE4 158-163
AVX2 160-164
CMake 160-163

:)

JordanTheToaster:
CMake is now 1 to 1 with SSE4 from my testing with Burnout 3 at 8x internal using Vulkan.


fqwegweijpwrhg

JordanTheToaster (approved):

stenzek:
Probably a good one to test here would be the Star Ocean menus - they do a ton of deswizzling and is GS thread bottlenecked when using the hash cache, so exercises that path.

JordanTheToaster:
> Probably a good one to test here would be the Star Ocean menus - they do a ton of deswizzling and is GS thread bottlenecked when using the hash cache, so exercises that path.

One graph coming up.


ehedrjnerdjnhre

refractionpcsx2:
That works out to be < 1%, from some quick math :)

seta-san:
Finally the beginning of collapsing the number of builds and tests for each pr. First multi isa and next the total death of wx. :)

refractionpcsx2:
Because it hasn't been noted, with a specific GS dump, the Cmake performance is about the same as the SSE4 build also in the software renderer, where I would expect the AVX2 to work (that's the whole point, right?)

These were my results in sw mode with 3 threads

SSE4: ~750
AVX2: ~845
CMAKE: ~750

Dump for testing:
[Star_Ocean_3_-_Till_the_End_of_Time__Disc1of2__SLUS-20488_20220819113329.gs.zip](https://github.com/PCSX2/pcsx2/files/9383692/Star_Ocean_3_-_Till_the_End_of_Time__Disc1of2__SLUS-20488_20220819113329.gs.zip)

PR #6899, opened 2022-08-23
CMake MSVC Qt builds

tellowkrinkle:
### Description of Changes
Add a CMake Qt build to the windows CI

### Rationale behind Changes
Make sure CMake builds work with Qt

### Suggested Testing Steps
Make sure the executable actually works (I haven't tested it yet)
Make sure the updater works
Make sure it's about as fast as MSBuild SSE4 builds


JordanTheToaster:
Few things I've noticed when testing the icons for the system console and the exe itself are gone the system console after first boot is empty and doesn't print any info besides CRC and whenever you unfocus the window and refocus it prints out this error:

image

And currently none of the hotkeys seem to work like F4 or tab to speed up the games.

tellowkrinkle:
Should be fixed

JordanTheToaster (approved):
Previously mentioned issues are fixed everything seems to function correctly and performance is in line with SSE4 builds. LGTM.

.github/workflows/windows_build_qt.yml
```diff
@@ -87,12 +86,29 @@ jobs:
...
+            call "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+            cmake --build build --config Release || exit /b
+            copy build\pcsx2-qt\pcsx2-qt.* bin
+            copy build\updater\updater.* bin
+            del bin\*.manifest
+            for %%F in (Qt6Core Qt6Gui Qt6Widgets Qt6Network Qt6Svg Qt6Concurrent) do copy "${{ inputs.qt_dir }}\bin\%%F.dll" bin
+            xcopy /s ${{ inputs.qt_dir }}\plugins\*.dll bin\QtPlugins\
+            for /r bin\QtPlugins\ %%F in (*.dll) do if exist "%%~dpnFd.dll" del "%%~dpnFd.dll"
+            xcopy /s /i ${{ inputs.qt_dir }}\translations bin\translations
+            echo [Paths]> bin\qt.conf
```
stenzek, line 110: IMO it would be better to have this generated in the cmake script. Could do it manually, or with windeployqt (I think I included that in the zip)

.github/workflows/windows_build_qt.yml
```diff
@@ -87,12 +86,29 @@ jobs:
...
+
       - name: Build PCSX2
         shell: cmd
         run: |
           if "${{ inputs.configuration }}"=="CMake" (
-            call "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\${{ steps.cmake.outputs.vcvars }}"
-            cmake --build build --config ${{ steps.cmake.outputs.buildtype }}
+            call "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+            cmake --build build --config Release || exit /b
+            copy build\pcsx2-qt\pcsx2-qt.* bin
```
stenzek, line 103: Would it be worth just changing the output location to bin? (and setting a suffix for debug builds) for consistency with msbuild?

stenzek (commented):
LGTM otherwise

.github/workflows/windows_build_qt.yml
tellowkrinkle: Using windeployqt now

.github/workflows/windows_build_qt.yml
tellowkrinkle: I switched to using a cmake install step to copy to bin, similar to how we do linux builds.  Not perfect, but it's not like we expect people to be building locally with cmake anyways.

PR #6904, opened 2022-08-23
CDVD: Fix integer division on floats and fix logging

refractionpcsx2:
### Description of Changes
Fixes some integer division happening when calculating floats with decimal amounts. Also fix up some logging which was still WX style.

### Rationale behind Changes
More precision is needed when doing these calculations really, causing some CD reads to be off by 3000 cycles. Also the compiler was whining.  The logging fixes are just to make it show CAV/CLV instead of just C.

### Suggested Testing Steps
Not really worth testing, nothing known fixed/broken, unfortunately.

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -744,8 +744,9 @@ static uint cdvdRotationalLatency(CDVD_MODE_TYPE mode)
 	// CAV rotation is constant (minimum speed to maintain exact speed on outer dge
 	if (cdvd.SpindlCtrl & CDVD_SPINDLE_CAV)
 	{
-		float rotationPerSecond = (((mode == MODE_CDROM) ? CD_MIN_ROTATION_X1 : DVD_MIN_ROTATION_X1) * cdvd.Speed) / 60;
-		float msPerRotation = 1000.0f / rotationPerSecond;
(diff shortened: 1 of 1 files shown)


pcsx2/CDVD/CDVD.cpp
```diff
@@ -775,10 +776,9 @@ static uint cdvdRotationalLatency(CDVD_MODE_TYPE mode)
 		}
 		const float sectorSpeed = (((float)(cdvd.SeekToSector - offset) / numSectors) * 0.60f) + 0.40f;
 
-		float rotationPerSecond = (((mode == MODE_CDROM) ? CD_MAX_ROTATION_X1 : DVD_MAX_ROTATION_X1) * cdvd.Speed * sectorSpeed) / 60;
+		float rotationPerSecond = static_cast<float>(((mode == MODE_CDROM) ? CD_MIN_ROTATION_X1 : DVD_MIN_ROTATION_X1) * cdvd.Speed * sectorSpeed) / 60.0f;
 		float msPerRotation = 1000.0f / rotationPerSecond;
```
lightningterror, line 780: Might as well make them constant as well, upper ones too.

kenshen112:
I might have to try it with the cue parser. Given it was having some odd read issues this might fix

refractionpcsx2:
I doubt it, this PR is highly uninteresitng.

lightningterror (approved):

PR #6929, opened 2022-08-27
GS-hw: Improve how we handle DATE selection on fbmask.

lightningterror:
### Description of Changes

GS-hw: Improve how we handle DATE selection on fbmask.
Check if 7th bit is masked, and alpha is still written, adjust date one accordingly since we know alpha will be below 128.
### Rationale behind Changes

Accuracy
Fixes Beyond Good and Evil with Accurate DATE enabled on dx.
This works because we otherwise would run date one on a draw where we shouldn't.
### Suggested Testing Steps

So far I only detected 2 games, Beyond good and evil doesn't want date one, whereas Driver wants it so would've been fine anyway with the option enabled.
[Beyond good and evil sps.zip](https://github.com/PCSX2/pcsx2/files/9438481/Beyond.good.and.evil.sps.zip)
[driver-parallel-lines.zip](https://github.com/PCSX2/pcsx2/files/9438485/driver-parallel-lines.zip)

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
@@ -3415,8 +3415,9 @@ void GSRendererHW::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sourc
 				DATE_BARRIER = true;
 			}
 		}
-		// When Blending is disabled and Edge Anti Aliasing is enabled, the output alpha is Coverage (which we force to 128) so DATE will fail/pass guaranteed on second pass.
-		else if (m_conf.colormask.wa && (m_context->FBA.FBA || IsCoverageAlpha()))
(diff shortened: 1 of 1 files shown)


lightningterror:
Beyond good and evil, maximum blending used for consistency with other renderers.
Master:
image

PR:
image

tellowkrinkle:
DATE_one says "only the first pixel to draw passes"
If alpha is masked, shouldn't that be the equivalent of simple destination alpha ("all pixels pass")?

lightningterror:
It's not a full alpha mask tho, we only check the 7th bit if it's masked, we already check if alpha is fully masked or not, in the case of Beyond good and evil alpha mask is 0x81.

tellowkrinkle:
If the 7th bit is masked, that means it never changes, right?
If so, then wouldn't that mean "no pixel will write an alpha that will fail subsequent pixels"?

lightningterror:
Yeah, I see what you mean, you're right.

PR #6931, opened 2022-08-28
GS: Use accumulation blend when alpha is known to be 1

tellowkrinkle:
### Description of Changes
Enables the accumulation blend flag for more situations when alpha is known to be 1

### Rationale behind Changes
- More hardware blending opportunities for maximum blending
- The new truncation-free blend_mix algorithm doesn't always give the correct values for a (255 - DstColor) color inversion on AMD GPUs.  This is used by some games to invert depth.  Conveniently, such inversions tend to use 1.0 afix alpha (since they're just trying to invert, not scale), which means they get caught by the new detection and pushed off of blend_mix onto the accumulation blend path, working around the issue.

### Suggested Testing Steps
Test the following trace in DX11, DX12, or barrier-free Vulkan on an AMD gpu:
[GodOfWar-MissingFloor.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/9439132/GodOfWar-MissingFloor.gs.xz.zip)

<details>
<summary>Screenshots</summary>

Broken:
image

Working:
image

</details>

Also test your stuff on maximum blending, and see if anything has less barriers now

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
@@ -2504,9 +2504,10 @@ void GSRendererHW::EmulateBlending(bool& DATE_PRIMID, bool& DATE_BARRIER, bool& 
 	const bool alpha_c2_zero = (m_conf.ps.blend_c == 2 && ALPHA.FIX == 0u);
 	const bool alpha_c2_one = (m_conf.ps.blend_c == 2 && ALPHA.FIX == 128u);
 	const bool alpha_c2_high_one = (m_conf.ps.blend_c == 2 && ALPHA.FIX > 128u);
+	const bool alpha_one = alpha_c0_one || alpha_c2_one;
 
 	// Optimize blending equations, must be done before index calculation
-	if ((m_conf.ps.blend_a == m_conf.ps.blend_b) || ((m_conf.ps.blend_b == m_conf.ps.blend_d) && (alpha_c0_one || alpha_c2_one)))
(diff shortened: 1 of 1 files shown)


stenzek:
Looks fine to me on AMD and NVIDIA.

Mrlinkwii (approved):

pcsx2/GS/Renderers/HW/GSRendererHW.cpp
```diff
@@ -2832,6 +2839,23 @@ void GSRendererHW::EmulateBlending(bool& DATE_PRIMID, bool& DATE_BARRIER, bool&
...
+				m_conf.ps.blend_b = 2;
+			if (m_conf.ps.blend_d == 1)
+				m_conf.ps.blend_d = 2;
+
+			if (m_conf.ps.blend_a == 2)
+			{
+				// Accumulation blend is only available in (Cs - 0)*Something + Cd, or with alpha == 1
+				ASSERT(m_conf.ps.blend_d == 2 || alpha_one);
+				// A bit of normalization
+				m_conf.ps.blend_a = m_conf.ps.blend_d;
```
lightningterror, line 2855: You have `m_conf.ps.blend_a = m_conf.ps.blend_b;` a bit below.

tellowkrinkle: Is there a problem with that?

lightningterror: I was thinking maybe it's gonna make this redundant but guess it's fine.

lightningterror:
Fine on intel as well, checked only dx11 tho, checked other dumps on nvidia as well, didn't spot issues.

PR #6938, opened 2022-08-28
Memory Cards: Implement conversion in Qt

RedPanda4552:
### Description of Changes
Implements the memory card conversion option in Qt.

### Rationale behind Changes
Memory card conversion was not implemented in the initial move from wx to Qt.

### Suggested Testing Steps
Convert some memory cards, check that games still see their data on the converted cards.

### DIFF
--- a/pcsx2-qt/Settings/MemoryCardConvertDialog.h
+++ b/pcsx2-qt/Settings/MemoryCardConvertDialog.h
@@ -1,0 +1,78 @@
+/*  PCSX2 - PS2 Emulator for PCs
+ *  Copyright (C) 2002-2022  PCSX2 Dev Team
+ *
+ *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
+ *  of the GNU Lesser General Public License as published by the Free Software Found-
+ *  ation, either version 3 of the License, or (at your option) any later version.
+ *
+ *  PCSX2 is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ *  without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+ *  PURPOSE.  See the GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along with PCSX2.
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <QtWidgets/QDialog>
+
+#include "common/FileSystem.h"
+
+#include "ui_MemoryCardConvertDialog.h"
+
+#include "MemoryCardConvertWorker.h"
+
+#include "pcsx2/MemoryCardFile.h"
+
+class MemoryCardConvertDialog final : public QDialog
+{
+	Q_OBJECT
+
+public:
+	explicit MemoryCardConvertDialog(QWidget* parent, QString selectedCard);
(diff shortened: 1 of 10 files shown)


pcsx2-qt/Settings/MemoryCardConvertDialog.h
```diff
@@ -0,0 +1,52 @@
...
+
+	QString m_selectedCard;
+	AvailableMcdInfo m_srcCardInfo;
+
+	QString m_destCardName;
+	AvailableMcdInfo m_destCardInfo;
+
+	MemoryCardType m_type = MemoryCardType::File;
+	MemoryCardFileType m_fileType = MemoryCardFileType::PS2_8MB;
+};
```
F0bes, line 52: Missing EOF newline

JordanTheToaster:
Doesn't appear to work tried to convert a 32MB card to a folder and Burnout 3 saw it as unformatted.

JordanTheToaster (approved):
All good here now LGTM.

Mrlinkwii (approved):

lightningterror (commented):
Commit can be named properly with prefix.

refractionpcsx2 (changes requested):
Gave it a quick test, and I don't think the conversion from folder to memcard is as good as it could be.

At the moment, if you have too much in your folder, it spams the console (which the user might not even have open) then it says "conversion complete!" and the user might delete the folder thinking everything is peachy, especially if they don't have the console window open.

What I would like to see:
1. When converting to memcard, check if there's enough space for the current game on the card, if not, start a new card, so like `folder_converted_1.mcd` or something.  
2. Maybe make the message to the user a bit more verbose, like "Due to the size of the folder, multiple memcards had to be created" or something.  I dunno, creative freedom :P

RedPanda4552:
@refractionpcsx2 

- Picklist now filters options based on computed space requirements of the card
- Fixed an issue where a card containing more than 512 individual files cannot be opened (by basically changing the default number of files Windows allows a C family program to open from 512 to 1024, equal to that of Unix platforms)
- Added some notes to the UI about it stalling while processing the conversion
- Conversion menu will now refuse to open if folder memcard exceeds 64 MB of data, since no file types can hold more than that atm

JordanTheToaster:
Converted a few of my memory cards including a 32MB card and it seems to read and save fine looks good to me.

Mrlinkwii:
Converted a few of my memory cards including a 64MB memcard  , games seem happy

lightningterror (commented):
Commit names should still be properly named, maybe squash some too.

RedPanda4552:
Squashed.

lightningterror (commented):
`Qt: Add dialog for memory card conversion.`

Will be better for consistency.

lightningterror (approved):

kenshen112:
Shoved my ABSOULTE UNIT cards into the QT converter. Flipped to folder and flipped back to image with no seeming issues. Was able to browse in BIOS and see sizes and order of saves remained the same

refractionpcsx2:
LGTM, nice work.

If you could squash things down and sort out your commit names, we can go ahead and merge :)

refractionpcsx2 (approved):

RedPanda4552:
Squashed and rebased

PR #6960, opened 2022-08-30
GameDB: Spyro NTSC-U bloom fix and add missing PAL demo disc

icup321:
### Description of Changes
Adds HalfPixelOffset bloom fix for NTSC-U version of Spyro - A Hero's Tail and adds a missing PAL variant of the Crash + Spyro demo disc. Thankfully the demo versions of Spyro don't seem to use bloom effects like the retail versions so the HalfPixelOffset fix isn't needed.

### Rationale behind Changes
Fixes some graphics, keeps regional versions in sync with each other.

### Suggested Testing Steps
Check bloom in retail NTSC-U version and check graphics in PAL demo disc.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -14240,6 +14240,8 @@ SLED-52574:
   region: "PAL-E"
   gameFixes:
     - XGKickHack # Fixes bad geometry.
+  gsHWFixes:
+    halfPixelOffset: 2 # Fixes depth lines.
 SLES-52576:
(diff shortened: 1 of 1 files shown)


lightningterror (commented):
Can you squash and name your commit properly with prefix?

Mrlinkwii:
@icup321 can you post a GS dump of whats being fixed?

lightningterror (commented):
Can you squash the commits and name it properly with prefix?

lightningterror:
@icup321 Status?

Issue #6822, opened 2022-08-11
[Feature Request]: (NOT FULFILLED) Smooth scrolling for mouse wheel in the game list and game grid


Rex000:
### Description

Something similar to what modern browsers have.

### Reason

This will be a nice QoL change, and will surely make scrolling through the list/grid super comfortable.

### Examples

.

Rex000:
#6971 did not implement smooth scrolling.

Rex000:
#6971 did not implement smooth scrolling, therefore this request should not be closed.

stenzek:
I don't even know what you're asking for here. You didn't provide any detail.

refractionpcsx2:
My best guess is having a bit of "inertia" like browsers do (since they have the coveted smooth scrolling)

neonoxd:
yes I think Rex000 means scroll steps with ease in/out, my PR didn't implement that exactly. However the issue should be updated because what is described in the "Reason" part is pretty much solved.

stenzek:
If it's the "smooth scrolling" setting similar to what Firefox has, this isn't something which is possible with Qt or native widgets. You'd have to implement it manually, which personally I don't have any interest in doing. PRs welcome of course.

Rex000:
> yes I think Rex000 means scroll steps with ease in/out, my PR didn't implement that exactly. However the issue description should be updated because what is described in the "Reason" part is pretty much solved.

I've updated it. In the reason part I only mentioned the game grid as an example. But yes, your update has made the game grid much easier to scroll now than before.

Rex000:
> If it's the "smooth scrolling" setting similar to what Firefox has, this isn't something which is possible with Qt or native widgets. You'd have to implement it manually, which personally I don't have any interest in doing. PRs welcome of course.

I read somewhere the QTimeLine could do it. I'm not an expert though so I can't say more.

Rex000:
> If it's the "smooth scrolling" setting similar to what Firefox has, this isn't something which is possible with Qt or native widgets. You'd have to implement it manually, which personally I don't have any interest in doing. PRs welcome of course.

I read somewhere that either a timer or alternatively QTimeLine could do it. I'm not an expert though so I can't say more.

Rex000:
> If it's the "smooth scrolling" setting similar to what Firefox has, this isn't something which is possible with Qt or native widgets. You'd have to implement it manually, which personally I don't have any interest in doing. PRs welcome of course.

[...]

PR #6971, opened 2022-08-31
Qt: Scroll GameList by pixels using the scrollwheel

neonoxd:
### Description of Changes
When scrolling the GameList table 1 step was actualy a whole page, now it's scrolling by 15 pixels

### Rationale behind Changes
When scrolling through your Game List you'd get lost on how much you actually scrolled, I propose a different behaviour than how it is currently.
It depends on what you are used to, but I feel like this resembles how you generally scroll content a bit more reasonably
From my testing, now the speed is affected by your OS's mouse settings, so if you change your "Scrolled Lines" settings in Windows, you will notice that you also scroll more here. 

So if you feel like this scrolling speed is too slow you can adjust the OS level

### Suggested Testing Steps
- Test with different mouse drivers that let you set your _scroll distance_
- Test changing the scrolled lines settings in Windows
- Test if behaviour differs on linux

### Note:
I'm not entirely sure if this arbitrary number (15) is the most satisfactory speed but it feels like a good baseline,

<details>
  <summary>Demonstration:</summary>

## Before:
  pcsx2-qtx64_agAIXs8CKO

## After:
  pcsx2-qtx64_dtJ7BJKDfF
</details>

Close https://github.com/PCSX2/pcsx2/issues/6822

### DIFF
--- a/pcsx2-qt/GameList/GameListWidget.cpp
+++ b/pcsx2-qt/GameList/GameListWidget.cpp
@@ -27,6 +27,7 @@
 #include <QtWidgets/QApplication>
 #include <QtWidgets/QHeaderView>
 #include <QtWidgets/QMenu>
+#include <QtWidgets/QScrollBar>
 
 #include "GameListModel.h"
 #include "GameListRefreshThread.h"
@@ -184,6 +185,7 @@ void GameListWidget::initialize()
 	m_list_view->setFrameStyle(QFrame::NoFrame);
 	m_list_view->setSpacing(m_model->getCoverArtSpacing());
 	m_list_view->setVerticalScrollMode(QAbstractItemView::ScrollMode::ScrollPerPixel);
+	m_list_view->verticalScrollBar()->setSingleStep(15);
 
 	updateListFont();
 


Mrlinkwii:
this may fix #6822 , im not sure

lightningterror:
> I'm not entirely sure if this arbitrary number (15) is the most satisfactory speed but it feels like a good baseline.

We can change it in the future if people want a higher number.

Rex000:
Thanks for this change, it certainly got better because of lower stepping. However, it's far from what smooth scrolling is supposed to be. So I don't think it fulfills my request #6822

RedDevilus:
Technically you could also hold left mouse key on the game entries and use it as a scrollwheel.

lightningterror:
> Thanks for this change, it certainly got better because of lower stepping. However, it's far from what smooth scrolling is supposed to be. So I don't think it fulfills my request #6822

It doesn't but it'll have to do for now.

Rex000:
> > Thanks for this change, it certainly got better because of lower stepping. However, it's far from what smooth scrolling is supposed to be. So I don't think it fulfills my request #6822
> 
> It doesn't but it'll have to do for now.

Sure it could, but since it doesn't fulfill the request, the said request should be kept open IMO until someone finds a way to actually implement it. Otherwise it will be forgotten since it's closed.

PR #6986, opened 2022-09-02
microVU: Fix program dumping

stenzek:
### Description of Changes

Invalid format string which caused a crash at runtime.

### Rationale behind Changes

Crash bad.

### Suggested Testing Steps

You can't really test this since it's a compile-time option, but I have tested locally.


JordanTheToaster (approved):
Solid changes increased VPS 197%.

PR #6993, opened 2022-09-05
Better PrimID destination alpha support

### LINKED ISSUE #4480: [BUG] KimiKiss ebkore+ version
**Describe the bug**
Topic bubble rendering is incorrect in D3D11 hardware.

**To Reproduce**
Use hardware renderer in conversation with a girl.

**Expected behavior**
Topics should look like thought bubbles

**GS Settings**
pcsx software
pcsx hardware

**Emulation Settings**
Safe preset with no speedhacks
pcsx2_CoreSettings_Speedhacks

**GS Window Screenshots**
Software mode or OpenGL (correct)
correct rendering
D3D11 mode (incorrect)
d3d11_render

**System Info:**
 -   PCSX2 Revision: dev-1283
 -   OS: Windows 10
 -   CPU: Ryzen 5 3600
 -   GPU: GTX 1660 SUPER

**Logs and Dumps**
[correct rendering.xz.gz](https://github.com/PCSX2/pcsx2/files/6597516/correct.rendering.xz.gz)
[d3d11_render.xz.gz](https://github.com/PCSX2/pcsx2/files/6597517/d3d11_render.xz.gz)

tellowkrinkle:
### Description of Changes
- Fixes an issue where when barriers were unsupported, primid destination alpha would be skipped on draws that would have otherwise gone to barriers if both barriers and primid were supported.
- Adds PrimID destination alpha support to DX11 (finally giving it accurate destination alpha support)
  - Properly fixes #4480
- Switches OpenGL's primid destination alpha algorithm to match other renderers
- Fixes combining primid destination alpha with geometry shaders on Vulkan

Note: DX12's primid destination alpha is still broken, at least on my old AMD-Mac drivers.  Renderdoc was reporting some very weird things and I was very confused (somehow the init and main draws had different geometry, but only if you looked at the visual geometry viewer and not the list of vertex shader output values (geometry shader output values were just weird id numbers)?  Other than that the post-init texture looked like you would expect, aside from the fact that the wireframe view indicated that the main draw wasn't drawing in the same places...).

Note 2: Updates vulkan headers to get access to the barycentric coordinates extension, which we use as a stand-in for a primitive-id support check for MoltenVK.

### Rationale behind Changes
Accurate destination alpha for everyone, even those ancient GPUs that support DX11 feature level 10_0 and don't support enough OGL extensions for our OGL renderer

### Suggested Testing Steps
- Test [kimikiss_dstalpha.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/9487511/kimikiss_dstalpha.gs.xz.zip)
  - Test on Vulkan with barriers forced off
  - Test on DX12 (Didn't work for me, but maybe it will for you)
  - Test on DX11
  - Test on OpenGL with barriers forced off
- Test any other heavy destination alpha games you might have
  - Do some performance comparisons for OpenGL?  Maybe it'll be different?

### DIFF
--- a/bin/resources/shaders/dx11/tfx.fx
+++ b/bin/resources/shaders/dx11/tfx.fx
@@ -102,7 +102,7 @@ struct PS_INPUT
 #else
 	nointerpolation float4 c : COLOR0;
 #endif
-#if PS_DATE > 10 || PS_DATE == 3
+#if PS_DATE >= 1 && PS_DATE <= 3
 	uint primid : SV_PrimitiveID;
 #endif
 };
@@ -110,7 +110,7 @@ struct PS_INPUT
 struct PS_OUTPUT
 {
 #if !PS_NO_COLOR
-#if PS_DATE > 10
+#if PS_DATE == 1 || PS_DATE == 2
 	float c : SV_Target;
 #else
 	float4 c0 : SV_Target0;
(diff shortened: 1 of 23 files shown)


Mrlinkwii (approved):

lightningterror (approved):
Went through a list of dumps that use date, didn't spot any issues besides dx12 being broken, but that's a different issue.

PR #7004, opened 2022-09-07
iR5900: Fix vcalls detection in COP2MicroFinishPass::Run

Goatman13:
### Description of Changes
Fix VCALLMS/VCALLMSR detection in COP2MicroFinishPass::Run.
### Rationale behind Changes
I can't find any affected game, but previous code wasn't able to detect anything (true condition was unreachable, and wrong). 
### Suggested Testing Steps
Maybe test games that need disabled mvuFlagSpeedHack, just in case Soukou Kihei Armodyne still need GameDB entry.

### DIFF
--- a/pcsx2/x86/iR5900Analysis.cpp
+++ b/pcsx2/x86/iR5900Analysis.cpp
@@ -266,7 +266,7 @@ void COP2MicroFinishPass::Run(u32 start, u32 end, EEINST* inst_cache)
 		}
 
 		// Except for VCALLMS/VCALLMSR, that can start a micro, so the next instruction needs to finish it.
-		if (_Funct_ == 070 || _Funct_ == 071)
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
I don't understand? The numbers for VCALLMS/R were correct

stenzek:
It's octal, yes. I just followed the convention, which I presume was due to how the opcode tables were laid out in groups of 8.

https://github.com/PCSX2/pcsx2/blob/65b0511d9e7cd4b3d04cbe42dd2bf70ca6fd3297/pcsx2/x86/microVU_Macro.inl#L634-L643

Goatman13:
Ah, ok then. Sorry for confusion, i really never had any interact with octal numbers in programming. Closing then. :)

refractionpcsx2:
It's okay, they confused me too at first :D

PR #7006, opened 2022-09-07
Qt: Implement settings reset

stenzek:
### Description of Changes

Plus a bunch of associated cleanup.

Also gets rid of the "settings are an incorrect version" popup the first time PCSX2 is started. If the settings ini is corrupted and unloadable, it'll prompt the user to confirm before resetting/overwriting. Safety first ;)

### Rationale behind Changes

Missing functionality in the Qt interface.

### Suggested Testing Steps

Test setting resets in both Qt and big picture ui. Make sure nothing is broken during normal use.


stenzek:
Already merged as part of #7016.

PR #7041, opened 2022-09-12
Misc: Codacy warnings cleanup.

lightningterror:
### Description of Changes

Misc: Codacy warnings cleanup.
### Rationale behind Changes

Less warnings.
### Suggested Testing Steps

Compiles.


pcsx2/windows/VCprojects/IopSif.cpp
```diff
@@ -90,12 +90,6 @@ void FinalizeEERead()
 	CPU_INT(DMAC_SIF0, 16);
 }
 
-s32 DoSIFWrite(u32 iopAvailable)
-{
-	u32 eeAvailable = PrepareEEWrite();
```
Warepire, line 95: This is undefined but probably returning eeAvailable on accident. So the code in sif0DmaRead was probably doing what it was intended to do. Or is the code dead for other reasons?

lightningterror: The whole file isn't used, I'll be purging it.

tellowkrinkle: Why bother making modifications to a file you're about to delete?

Maybe just delete it now

lightningterror: That's what I said?

PR #7043, opened 2022-09-12
Input:SDL: Allow use of axes and buttons outside the normal range

tellowkrinkle:
### Description of Changes
Allow use of buttons outside of normal defined GameController range

### Rationale behind Changes
This might allow DS3s to use analog axes

### Suggested Testing Steps
Test DS3s and see if they can use analog axes

### DIFF
--- a/pcsx2/Frontend/InputManager.cpp
+++ b/pcsx2/Frontend/InputManager.cpp
@@ -275,7 +275,7 @@ std::string InputManager::ConvertInputBindingKeyToString(InputBindingKey key)
 		}
 		else if (key.source_subtype == InputSubclass::PointerAxis)
 		{
-			return fmt::format("Pointer-{}/{}{:c}", u32{key.source_index}, s_pointer_axis_names[key.data], key.negative ? '-' : '+');
+			return fmt::format("Pointer-{}/{}{:c}", u32{key.source_index}, s_pointer_axis_names[key.data], key.modifier == InputModifier::Negate ? '-' : '+');
 		}
 	}
 	else if (key.source_type < InputSourceType::Count && s_input_sources[static_cast<u32>(key.source_type)])
@@ -489,9 +489,9 @@ std::optional<InputBindingKey> InputManager::ParsePointerKey(const std::string_v
 
 			const std::string_view dir_part(sub_binding.substr(std::strlen(s_pointer_axis_names[i])));
 			if (dir_part == "+")
-				key.negative = false;
+				key.modifier = InputModifier::None;
 			else if (dir_part == "-")
-				key.negative = true;
+				key.modifier = InputModifier::Negate;
 			else
 				return std::nullopt;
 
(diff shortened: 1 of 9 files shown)


lucasmr:
Tested it, same behavior. Can't rebind with the controller.

tellowkrinkle:
@lucasmr Okay how about now

lucasmr:
nope

RedPanda4552:
@lucasmr To clarify because I think there's a communication breakdown here:

This PR should allow a DS3 to be bound by a pressure sensitive axis. Your phrasing suggests your issue is a DS3 _not binding at all_. Can you clarify which of the two?

lucasmr:
Sorry, let me be more clear. The UI doesn't respond to the DS3 buttons at all, nothing binds it. I can only bind buttons to keyboard or mouse clicks. It is detected, though. I can alter the rumble strength.

tellowkrinkle:
Ohh
That's not how I was interpreting it
(Side note, I tested with a modified SDL where I mapped my controller's sticks to axes 6, 7, 8, and 9, and it did work there after the second commit)

Does the DS3 show up in the list of connected controllers?
Also, if you use the DS3 with a program that lists all SDL axis and button outputs, can you report what values it gives for the pressure axes?

lucasmr:
```
$ sdl-jstest/build/sdl2-jstest --list
error: failed to read gamecontrollerdb.txt: Invalid RWops
Found 1 joystick(s)

Joystick Name:     'PS3 Controller'
Joystick GUID:     0300afd34c0500006802000000016800
Joystick Number:    0
Number of Axes:    16
Number of Buttons: 15
Number of Hats:     0
Number of Balls:    0
GameControllerConfig:
  Name:    'PS3 Controller'
  Mapping: '0300afd34c0500006802000000016800,*,a:b0,b:b1,back:b4,dpdown:b12,dpleft:b13,dpright:b14,dpup:b11,guide:b5,leftshoulder:b9,leftstick:b7,lefttrigger:a4,leftx:a0,lefty:a1,rightshoulder:b10,rightstick:b8,righttrigger:a5,rightx:a2,righty:a3,start:b6,x:b2,y:b3,crc:d3af,platform:Linux'
  ```

Via sdl2-jstest:
image

The controller, detected on pcsx2:
image

**Comment by mirh** [comment]
Did you remember to use `SDL_JOYSTICK_HIDAPI_PS3=1` this time?

**Comment by tellowkrinkle** [comment]
ohhh
This change only affects Qt builds
wx builds are stuck with no controller remapping, and that probably won't change before we remove wx

**Comment by lucasmr** [comment]
image
That's from the qt build, automatic mapping. Gonna compare now with the upstream qt build...

**Comment by lucasmr** [comment]
Just tested with Burnout, I can confirm the pressure sensitivity worked after manually binding on your branch. The direction is wrong, though: button released was full throttle.

**Comment by tellowkrinkle** [comment]
Manually open up the ini and edit the `-`s to `+`s to invert.  While you're at it, you can remove `SDL-0/DPadUp &` part.

**Comment by lucasmr** [comment]
The button values ranges from:
-32768: released
+32767: fully pressed

There's only a response after the button goes beyond 0, or half the pressure.

File: pcsx2/Frontend/InputManager.cpp [hunk 1]
```diff
@@ -669,11 +669,24 @@ bool InputManager::InvokeEvents(InputBindingKey key, float value, GenericInputBi
...
+				case InputModifier::None:
+					if (value > 0.0f)
+						value_to_pass = value;
+					break;
+				case InputModifier::Negate:
+					if (value < 0.0f)
+						value_to_pass = -value;
+					break;
+				case InputModifier::FullAxis:
+					value_to_pass = value * 0.5f + 0.5f;
```
lucasmr, line 687: I think what you want to do here instead is add 32768 and then divide by two?
Example input: -32768 (button released)
new = (-32768 + 32768)/2

tellowkrinkle:
@lucasmr Can you try the latest build and change `+Axis##` to `FullAxis##`?
(Note: I didn't add support to the auto detection, only manual INI edits)

pcsx2/Frontend/InputManager.cpp
tellowkrinkle: Inputs are [normalized fairly early on](https://github.com/tellowkrinkle/pcsx2/blob/b410447bdab63ed0b54a7745cd420c7ccd939cb6/pcsx2/Frontend/SDLInputSource.cpp#L514) so at that point they're between -1 and 1

Not perfect, since that means we scale values less than 0 slightly differently from values greater than 0 (rather than adjusting the original number to 0-65535 before normalizing) but it should be close enough.

lucasmr:
I managed to test with [padtest.elf](https://www.ps2-home.com/forum/viewtopic.php?t=5483), to avoid non linear translation in games. I can say the pressure correlates very close to sdl2-jstest values!

lucasmr:
Now with the release of [SDL 2.26](https://github.com/libsdl-org/SDL/releases/tag/prerelease-2.25.1), can this work be merged?

PR #7045, opened 2022-09-13
Qt: Make FastCDVD per game only

JordanTheToaster:
### Description of Changes
Makes FastCDVD per game only to prevent users enabling a destructive setting globally.

### Rationale behind Changes
Destructive settings should be per game only or ideally hidden away so there is less chance of users breaking something unintentionally.

### Suggested Testing Steps
Make sure it works per game and that CI passes.

### DIFF
--- a/pcsx2-qt/Settings/SystemSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/SystemSettingsWidget.cpp
@@ -48,6 +48,9 @@ SystemSettingsWidget::SystemSettingsWidget(SettingsDialog* dialog, QWidget* pare
 	SettingWidgetBinder::BindWidgetToIntSetting(sif, m_ui.vuRoundingMode, "EmuCore/CPU", "VU.Roundmode", 3);
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.fastCDVD, "EmuCore/Speedhacks", "fastCDVD", false);
 
+	// Allow for FastCDVD for per-game settings only 
+	m_ui.fastCDVD->setEnabled(m_dialog->isPerGameSettings());
+
 	if (m_dialog->isPerGameSettings())
 	{
 		m_ui.eeCycleRate->insertItem(


mirh:
As I [already](https://github.com/PCSX2/pcsx2/pull/6584#issuecomment-1190230164) argued, this reasoning is backwards and could be paradoxically applied to just about every other setting as well. 

Yes, fastCDVD has potentially "catastrophic" downsides. But being fail-fast is also what makes it so easy to spot something was wrong to begin with. 
Conversely, the more a setting isn't so "all-around encompassing" and the more it can reserve you a pretty bad time down the road when you spot the slightest glitch.

RedDevilus:
Having an option exposed that doesn't have benefits for most games but like 10 or so ( i recall silent hill as example) and can make a mess of others. 

To be fair the naming is deceptive at best it's not like it will increase speeds like loading screens/transitions.

Options should be helpful, not bombard the user and it was always better as a gamefix than a so proclaimed speedhack.

Donel-Bueno:
Is there a list of games this setting benefits?

JordanTheToaster:
> Is there a list of games this setting benefits?

Only 2 so it's pointless to leave it globally as an option when it can break games.

DonelBueno:
@mirh you just have to right click on the game you want to change the setting for, it doesn't make sense to argue against this change. It's not like it will make it super hard to enable.

mirh:
My point in the other thread was actually that per-game-only settings *forces* you to use the "second set" of options panel, which in turn increases the likelihood of forgetting you had set something once (at least if you happen to test games far more often than you play them like me, I believe global settings are far more handier to use). 

But if this hack is not only very game specific, but we already know that 99.98% of times it's also going to be "at best useless".. I already said that this PR is good?

DonelBueno:
> My point in the other thread was actually that per-game-only settings _forces_ you to use the "second set" of options panel, which in turn increases the likelihood of forgetting you had set something once (at least if you happen to test games far more often than you play them like me, I believe global settings are far more handier to use).
> 
> But if this hack is not only very game specific, but we already know that 99.98% of times it's also going to be "at best useless".. I already said that this PR is good?

Makes sense, but you can always check the gamedb file. Also, the emulator warns the user when a gamedb fix is not being applied because the user enabled "manual hardware renderer fixes".

PR #7059, opened 2022-09-17
Revert "BiosTools: Allow BIOS region patching"

F0bes:
This reverts commit a4dcaa7c149c80442e4087ff497561764d04001a.

### Description of Changes
Originally implemented to combat the issue the old BIOS dumper had, where slims would be dumped with the Japan BIOS region.
Now that we officially support https://github.com/F0bes/biosdrain which doesn't have this issue, there isn't a purpose for this feature.

If you are interested in poking around at the BIOS and seeing how it reacts to different regions, you can use the script I provided in the original pull request(#5778). [F0bes/BiosPatcher.py](https://gist.github.com/F0bes/1224ca17a8e71a0a5f93be9ced23b14b)
### Rationale behind Changes
The use case for this is niche, it's simply another thing a user can mistake for something else. 

### Suggested Testing Steps
Make sure the BIOS selection windows in QT and WX are both functional.

### DIFF
--- a/pcsx2-qt/Settings/BIOSSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/BIOSSettingsWidget.cpp
@@ -36,13 +36,9 @@ BIOSSettingsWidget::BIOSSettingsWidget(SettingsDialog* dialog, QWidget* parent)
 	m_ui.setupUi(this);
 
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.fastBoot, "EmuCore", "EnableFastBoot", true);
-	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.patchRegion, "EmuCore", "PatchBios", false);
-	SettingWidgetBinder::BindWidgetToEnumSetting(sif, m_ui.regionComboBox, "EmuCore", "PatchRegion", BiosZoneStrings, BiosZoneBytes, BiosZoneBytes[0]);
 	SettingWidgetBinder::BindWidgetToFolderSetting(sif, m_ui.searchDirectory, m_ui.browseSearchDirectory, m_ui.openSearchDirectory,
 		m_ui.resetSearchDirectory, "Folders", "Bios", Path::Combine(EmuFolders::DataRoot, "bios"));
 
-	dialog->registerWidgetHelp(m_ui.patchRegion, tr("Patch Region"), tr("Unchecked"),
-		tr("Patches the BIOS region byte in ROM. Not recommended unless you really know what you're doing."));
 	dialog->registerWidgetHelp(m_ui.fastBoot, tr("Fast Boot"), tr("Checked"),
 		tr("Patches the BIOS to skip the console's boot animation."));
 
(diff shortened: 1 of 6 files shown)


pcsx2-qt/Settings/BIOSSettingsWidget.cpp
```diff
@@ -36,13 +36,7 @@ BIOSSettingsWidget::BIOSSettingsWidget(SettingsDialog* dialog, QWidget* parent)
 	m_ui.setupUi(this);
 
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.fastBoot, "EmuCore", "EnableFastBoot", true);
-	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.patchRegion, "EmuCore", "PatchBios", false);
-	SettingWidgetBinder::BindWidgetToEnumSetting(sif, m_ui.regionComboBox, "EmuCore", "PatchRegion", BiosZoneStrings, BiosZoneBytes, BiosZoneBytes[0]);
-	SettingWidgetBinder::BindWidgetToFolderSetting(sif, m_ui.searchDirectory, m_ui.browseSearchDirectory, m_ui.openSearchDirectory,
```
stenzek, line 41: Probably want to keep this line, otherwise the directory won't be changeable.

F0bes: That'd be problematic. Fixed.

stenzek (approved):

AKuHAK:
> Originally implemented to combat the issue the old BIOS dumper had, where slims would be dumped with the Japan BIOS region.
> Now that we officially support https://github.com/F0bes/biosdrain which doesn't have this issue, there isn't a purpose for this feature.

@F0bes 
Just would like to mention that this is not an issue but correct behavior. All Deckard consoles (SCPH-75000+) have the Japan region in their bootrom chips (bios chip) - so their romver is always J. These consoles' mechacon has the ability to patch bootrom on the fly in the IOP memory (someone can check the string `EEengEE` inside nvram). First symbol will overwrite ROMVER 4th letter in the IOP RAM. Someone can quickly check this by overwriting this area on any Deckard console with any compatible romver symbol (J, A, E, H, C). Romver will also change once you change the nvram. Old pcsx2 dumper was accessing EE RAM, not IOP RAM, and inside EE RAM bios content is untouched, so it always reported itself as Japan.

F0bes:
> > Originally implemented to combat the issue the old BIOS dumper had, where slims would be dumped with the Japan BIOS region.
> > Now that we officially support https://github.com/F0bes/biosdrain which doesn't have this issue, there isn't a purpose for this feature.
> 
> @F0bes 
> Just would like to mention that this is not an issue but correct behavior. All Deckard consoles (SCPH-75000+) have the Japan region in their bootrom chips (bios chip) - so their romver is always J. These consoles' mechacon has the ability to patch bootrom on the fly in the IOP memory (someone can check the string `EEengEE` inside nvram). First symbol will overwrite ROMVER 4th letter in the IOP RAM. Someone can quickly check this by overwriting this area on any Deckard console with any compatible romver symbol (J, A, E, H, C). Romver will also change once you change the nvram. Old pcsx2 dumper was accessing EE RAM, not IOP RAM, and inside EE RAM bios content is untouched, so it always reported itself as Japan.
> 

Good to know, thanks. Mystery solved 🕵️

PR #7063, opened 2022-09-17
IOPBios: Properly increment host dir iterator

F0bes:
Fixes uLaunchELF and any other homebrew that iterates over directories.

### Description of Changes
We were incrementing the hostdir fd iterator but discarding the new pointer.
This made it so we were pointing at the same file despite the guest calling dread (which is supposed to increment to the next file).

Update:
Also implement different handling for getStat and dread depending on ioman or iomanx.
This was done because their internal structures are different.

### Rationale behind Changes
uLaunchELF host filesystem support was broken. Same with any other homebrew that relies on iterating a directory.

### Suggested Testing Steps
The attached elf reads the filenames of the first 10 files in the host root directory.

[hostfstest.elf.zip](https://github.com/PCSX2/pcsx2/files/9592410/hostfstest.elf.zip)

Please test uLaunchELF as well.

### DIFF
--- a/pcsx2/IopBios.cpp
+++ b/pcsx2/IopBios.cpp
@@ -356,7 +356,7 @@ namespace R3000A
 			StringUtil::Strlcpy(hostcontent->name, dir->FileName, sizeof(hostcontent->name));
 			host_stat(host_path(Path::Combine(basedir, dir->FileName), true), &hostcontent->stat);
 
-			static_cast<void>(std::next(dir)); /* This is for avoid warning of non used return value */
+			dir = std::next(dir);
 			return 1;
 		}
 
--- a/pcsx2/IopBios.cpp
+++ b/pcsx2/IopBios.cpp
@@ -56,6 +56,18 @@ typedef struct
 	unsigned char mtime[8];
 	unsigned int hisize;
 } fio_stat_t;
+typedef struct
+{
+	fio_stat_t _fioStat;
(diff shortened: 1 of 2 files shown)


F0bes:
There is an issue with uLaunchELF v4.40+, it results in a black screen, and is unrelated to this PR.
I've tested with uLaunchELF v3.9 and lower and it seems to work properly.
I've also created homebrew that iterates `host:` with both ioman and iomanx and it works fine.

I'm deeming this ready.

If you want a catalogue of uLaunchELFs, I've used the resource [here](https://sksapps.haldrie.com/ps2launchers.php).

pcsx2/IopBios.cpp
```diff
@@ -989,22 +1061,37 @@ namespace R3000A
...
-			EXPORT_H( 15, dread)
-			EXPORT_H( 16, getStat)
-		END_MODULE
+		// Special case with ioman and iomanX
+		// They are mostly compatible excluding stat structures
+		if(libname == "ioman" || libname == "iomanx")
+		{
+			const bool use_ioman = libname == "ioman";
+			using namespace ioman;
+				switch(index)
```
refractionpcsx2, line 1070: weird indenting

F0bes: I guess it's not wise to rely on clang-format when the code is in a clang-format off region lol.

PR #7064, opened 2022-09-17
Change lba to lsn

kenshen112:
### Description of Changes

Changed names of functions referring to CDVD LBA (Logical Block Address) to instead refer to the more correct term LSN (Logical sector number)

### Rationale behind Changes

Several functions already had lsn in the title where other's did not. This fixes a naming convention that did not match and thus unifies the code base a bit more

### Suggested Testing Steps

C if CI goes brrr

### DIFF
--- a/pcsx2/CDVD/CDVD.h
+++ b/pcsx2/CDVD/CDVD.h
@@ -33,7 +33,7 @@ static __fi s32 msf_to_lsn(u8* Time)
 	return lsn;
 }
 
-static __fi s32 msf_to_lba(u8 m, u8 s, u8 f)
+static __fi s32 msf_to_lsn(u8 m, u8 s, u8 f)
 {
 	u32 lsn;
 	lsn = f;
@@ -56,7 +56,7 @@ static __fi void lsn_to_msf(u8* Time, s32 lsn)
 	Time[2] = itob(f);
 }
 
-static __fi void lba_to_msf(s32 lba, u8* m, u8* s, u8* f)
+static __fi void lsn_to_msf(s32 lba, u8* m, u8* s, u8* f)
 {
 	lba += 150;
 	*m = lba / (60 * 75);
--- a/pcsx2/CDVD/CDVDdiscReader.cpp
+++ b/pcsx2/CDVD/CDVDdiscReader.cpp
@@ -65,7 +65,7 @@ inline void lsn_to_msf(u8* minute, u8* second, u8* frame, u32 lsn)
 // TocStuff
 void cdvdParseTOC()
 {
-	tracks[1].start_lba = 0;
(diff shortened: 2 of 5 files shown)


refractionpcsx2:
Can you fix your commit name? Thanks.

kenshen112:
That good?

refractionpcsx2:
better yeah

MonJamp:
In this line 96 of `CDVDdiscReader.cpp` you're renaming `start_lba` to `start_lsn` and assigning it to `entry.lba`.
I would be a little careful with this change. I remember running into this when I worked on merging the cdvd plugin into the core. 

According to the specs (MMC or MultiMedia Command Set) [`LBA` and `LSN` are synonymous](https://lists.gnu.org/archive/html/libcdio-devel/2012-11/msg00012.html).
However in `libcdio` it's defined as [`LSN = LBA - PREGAP_SECTORS`](https://github.com/xbmc/libcdio/blob/3ffee56e19c18142f07f0e52924c4e7b55d9e11d/lib/driver/sector.c#L51) where [`PREGAP_SECTORS = 150`](https://github.com/xbmc/libcdio/blob/3ffee56e19c18142f07f0e52924c4e7b55d9e11d/include/cdio/sector.h#L92).

In master [150 is added to convert from `LBA` and `LSN`](https://github.com/PCSX2/pcsx2/blob/master/pcsx2/CDVD/CDVD.h#L45-L65).
At the time I was working on this, I was very confused whether or not gigaherz was using `libcdio`'s definition of `LBA` vs `LSN`. 
Looking at the code now, I think `LBA` and `LSN` are used interchangeably throughout the code..

I recommend double checking if this is true and if it is we should rename all references from `LSN` to use `LBA` since that is the more common name. If `LBA` and `LSN` are used differently in the code, then this PR would be making the code misleading.

Sorry to be a pedant but seeing this PR come through I thought we might as well squash this issue now and avoid confusion in the future.

kenshen112:
You're not being pendant at all. According to my understanding, it seems like they're used interchangeably as the math seems almost exactly the same in both instances. As for LSN  vs LBA, I don't really have a preference I just used LSN since that would seem more fitting for DVD's (Sectors on a DVD vs Blocks) but if we want LBA to be the used term that's just fine as well

MonJamp:
I guess `LBA` vs `LSN` doesn't matter, I initially mentioned to change it to `LBA` since it was easier to google for 😛.
Since you already renamed it to `LSN` with this PR, renaming the remaining local variables, parameters, and class members from `LBA` to `LSN` will help clear up the code.

kenshen112:
Superceded by unifiedCdvd which doesn't rename anything yet, but might later

PR #7070, opened 2022-09-19
IOPBios: Implement the different iomanx stat file modes

F0bes:
### Description of Changes
ioman and iomanx have different file mode values.
This caused home-brew that relied on stat() to return directories as files, or files as directories, or something. It was **wrong** nonetheless.

### Rationale behind Changes
Improving iomanx support

### Suggested Testing Steps
Test home-brew that relies on the host filesystem.


fjtrujy (approved):
Glad finally we fixed it!

Mrlinkwii (approved):

JordanTheToaster (approved):

PR #7074, opened 2022-09-20
GameDB: Correct incorrect GTA SA fix

JordanTheToaster:
### Description of Changes
Fixes a mistake someone made adding ee clamping to GTA SA.

### Rationale behind Changes
Some doughnut added vu clamping not ee clamping so the mission "Reunite the family's" would still crash despite being listed as "fixed".

### Suggested Testing Steps
Make sure CI is happy.


Mrlinkwii (approved):

DiegoBrito94:
como que baixa essa poha ?? nada da certo

PR #7080, opened 2022-09-23
Fix build on FreeBSD

martymac:
${pcsx2OSXSources} cannot be built on FreeBSD, so remove them. While here, I've also simplified the logic.

### DIFF
--- a/pcsx2/CMakeLists.txt
+++ b/pcsx2/CMakeLists.txt
@@ -1583,12 +1583,12 @@ endif()
 
 # MacOSX/BSD
 if(UNIX AND NOT Linux)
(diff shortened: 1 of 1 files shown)


pcsx2/CMakeLists.txt
```diff
@@ -1583,8 +1583,14 @@ endif()
 
 # MacOSX/BSD
 if(UNIX AND NOT Linux)
+	if(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD" OR ${CMAKE_SYSTEM_NAME} MATCHES "NetBSD")
```
tellowkrinkle, line 1586: Might be a bit easier to do `if(APPLE)` and assume BSD otherwise

martymac: Right! I've just pushed an update to the patch. Thanks!

PR #7082, opened 2022-09-24
MemoryCardFolder: Use icon osd notification method.

stuken:
### Description of Changes
Tiny update to use the icon keyed osd message method like file memory card uses.

### Rationale behind Changes
The new icon printing method was introduced after the previous logging pr was submitted. This just brings it in line with those newer changes.

### Suggested Testing Steps
A nice little sd card icon appears in the osd messages when mc folder saves occur and nothing goes boom.


stenzek (approved):
lgtm

Mrlinkwii (approved):

PR #7089, opened 2022-09-25
GameDB: Don't search for CRC before the ELF has been loaded

MrCK1:
### Description of Changes
Stops GameDB from trying to search for a CRC when it hasn't been loaded yet from the ELF.

### Rationale behind Changes
Cleanliness

### Suggested Testing Steps
Make sure gamefixes and patches still load properly

### DIFF
--- a/pcsx2/GameDatabase.cpp
+++ b/pcsx2/GameDatabase.cpp
@@ -15,6 +15,7 @@
 
 #include "PrecompiledHeader.h"
 
+#include "R5900.h"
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Hmm, it feels like a bodge, wouldn't it be better to find where it's called findGame and make sure that only happens when there's a game loaded?

lightningterror:
@MrCK1 Status?

lightningterror:
@MrCK1 Status?

PR #7092, opened 2022-09-28
Common track and SubQ

kenshen112:
### Description of Changes

added a common track struct. Adds SubQ read in ps1 mode from disk reader / iso reader

### Rationale behind Changes

To better unify the codebase and add a common track struct for cue, chd, disk reader

### Suggested Testing Steps

See if it builds


pcsx2/CDVD/Ps1CD.cpp
```diff
@@ -365,14 +365,28 @@ void cdrInterrupt()
...
 			SetResultSize(8);
-			cdr.Result[0] = 1;
-			cdr.Result[1] = 1;
-			cdr.Result[2] = cdr.Prev[0];
-			cdr.Result[3] = itob((btoi(cdr.Prev[1])) - 2);
-			cdr.Result[4] = cdr.Prev[2];
-			cdr.Result[5] = cdr.Prev[0];
-			cdr.Result[6] = cdr.Prev[1];
-			cdr.Result[7] = cdr.Prev[2];
+			cdr.subQ = new cdvdSubQ();
```
stenzek, line 368: This is a memory leak. I'd also question the logic here - you should have subq for any sector, including those not in the data/audio tracks (lead-in/lead-out).

kenshen112: Is what I force-pushed any better?

stenzek: Not really. A heap allocation here is completely pointless, when you could just store it on the stack instead.

And what happens if readSubQ does error (e.g. due to using a physical disc and a read error)? The result is left with undefined and/or junk data?

kenshen112: I put it on the stack, and if the main subQ read doesn't work, it'll fake it instead

refractionpcsx2: Wasn't that the entire point of this PR?

kenshen112: Yeah, I just wasn't sure if the original logic for the faked subQ was the best so I tried to ax it. In hindsight, that was a stupid decision :sweat_smile:

kenshen112:
Is there anything else that needs to change?

lightningterror:
It was discussed that this isn't accurate and should be redone differently.

PR #7102, opened 2022-09-29
Qt: Audio backend selection for cubeb (and move in-tree)

### LINKED ISSUE #6693: [BUG]: "Reset Defaults" in Audio settings doesn't do anything
### Describe the Bug

I tried changing latency and timestrech settings but then decided to just use the defaults ones and then found out it doesn't reset for me at all.

### Reproduction Steps

Move audio sliders in the audio settings and then try resetting them and see if it works?

### Expected Behavior

Reset Defaults button to reset settings to their default placement.

### PCSX2 Revision

v1.7.3114

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

stenzek:
### Description of Changes

What the title says.

Also moves cubeb in-tree instead of using the submodules, because it needs the local modification to get the driver names, as well as silencing some annoying log output.

Also fixes up the alsa backend not being enabled for the CI builds.

### Rationale behind Changes

Making options dynamic and available in the UI, instead of hidden in the ini.

Fixes #6693.

### Suggested Testing Steps

Test new setting on Windows and Linux, make sure the correct backend is being used (can check verbose logs, cubeb is quite noisy and should say).


JordanTheToaster (approved):

stenzek:
Tested on Linux with both alsa and pulse backends.

F0bes (approved):
Tested on EndeavourOS (Arch derivative) with pipewire / alsa. Sound works fine.

PR #7104, opened 2022-09-29
GS: Handle higher TH/TW for STQ calculation

refractionpcsx2:
### Description of Changes
Modify the clamps/FixedTEX0 calculations to allow TEX0 to set values higher than 10 (max texture size on the GS)

### Rationale behind Changes
When using STQ, the calculations are done with the TH and TW sizes in mind that were written to the TEX0 register, clamping them to 10 causes the calculation result to be incorrect.

### Suggested Testing Steps
test games (Especially those where textures looked accidentally blown up), make sure nothing is busted. Affects both SW and HW.

TODO: Need to do a hardware test to confirm the clamp behaviour, I suspect it will clamp/wrap at 1024, but we will need to do a hardware test to confirm this.

Fixes the text size in `Simple 2000 Series Vol. 105 - The Maid Fuku to Kikanjuu`, ~~there is a further issue where the 4bit pixels are in the wrong order, however the PS2 also shows this error, so it could be another GS problem with an upload that only happens once, or it's a core problem.~~

Before:
image
image

After:
image
image

### DIFF
--- a/pcsx2/GS/GSDrawingContext.cpp
+++ b/pcsx2/GS/GSDrawingContext.cpp
@@ -120,7 +120,12 @@ GIFRegTEX0 GSDrawingContext::GetSizeFixedTEX0(const GSVector4& st, bool linear, 
 		th = extend(uv.y, th);
 	}
 
-	if (GSConfig.Renderer == GSRendererType::SW && ((int)TEX0.TW != tw || (int)TEX0.TH != th))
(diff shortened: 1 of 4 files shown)


refractionpcsx2:
Update: Turns out the font is just shit, ran the game normally on my PS2 and the result is this

image

Mrlinkwii (approved):

JordanTheToaster (approved):

PR #7109, opened 2022-10-01
UI/GS: Move trilinear option to main renderer tab in wx, also rename the config

lightningterror:
### Description of Changes

Move Trilinear to main renderer tab in wx gu, it's not an user hack on Qt so match the behavior.

Change Trilinear config name, it is no longer an user hack.
### Rationale behind Changes

Consistency.
### Suggested Testing Steps

Trilinear Filtering still works on Qt and Wx.


pcsx2/Config.h
```diff
@@ -549,7 +549,7 @@ struct Pcsx2Config
 		int UserHacks_TCOffsetX{0};
 		int UserHacks_TCOffsetY{0};
 		int UserHacks_CPUSpriteRenderBW{0};
-		TriFiltering UserHacks_TriFilter{TriFiltering::Automatic};
+		TriFiltering TriFilter{TriFiltering::Automatic};
```
stenzek, line 552: While I fundamentally agree that it's no longer a hack, changing the name here doesn't really gain much except breaking everyone's existing configuration.

PR #7131, opened 2022-10-04
GS:HW: Make HDR algorithm more float-precision-friendly

tellowkrinkle:
### Description of Changes
- Use whole numbers instead of 255ths
  - 255ths aren't an even multiple of a power of 2, so they have more precision issues.  When using half precision, some values, e.g. 191/255 and 192/255, start returning wrong results after accumulating just 5 times.
  - With whole numbers from 0-255, it takes 9 accumulations to have any chance of getting a wrong result
- Use range -128 - 127 instead of 0 - 255 for accumulation blends
  - This doubles the number of accumulations we can guarantee perfect results for, bringing it to 16 with half precision
  - If a game has an even distribution of values, we should be able to accumulate infinitely
    - One of our well known HDR-heavy games, Sly, uses exclusively `1` and `255`, which map to `1` and `-1`

### Rationale behind Changes
Hopefully this makes #7132 viable

### Suggested Testing Steps
- Test HDR games, make sure nothing changed
- Especially test OGL and DX, since those are completely untested at the moment


Mrlinkwii (approved):

refractionpcsx2 (approved):

lightningterror (approved):
From the dumps I tested I didn't spot any issues.

PR #7134, opened 2022-10-04
SPU: Allow up to 200% final volume

refractionpcsx2:
### Description of Changes
Extends the volume range allowed to up to 200%.

### Rationale behind Changes
Some games, such as Legacy of Kain - Soul Reaver 2, have tragically low volume levels (I streamed it from my PS2 hardware and it registers around -20dB, practically inaudible!) making it difficult for some people to hear, or anybody who dares attempt to stream the game will have trouble.

### Suggested Testing Steps
Run a game, fiddle with the volume slider.

Remember with great power becomes great responsibility. If you experience clipping in your game after going above 100% (Burnout 3 would be an example where this would happen), dial it back, that's on you :P

### DIFF
--- a/pcsx2-qt/Settings/AudioSettingsWidget.ui
+++ b/pcsx2-qt/Settings/AudioSettingsWidget.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>752</width>
-    <height>430</height>
+    <height>448</height>
    </rect>
   </property>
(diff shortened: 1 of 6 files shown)


marcoluc97:
Games with poor audio levels (what they think we hear):
giphy

Now with Refraction PR:
ezgif com-gif-maker

JordanTheToaster (approved):

PR #7138, opened 2022-10-05
GS-hw: Adjust Shadow of Rome crc hacks.

lightningterror:
### Description of Changes

GS-hw: Adjust Shadow of Rome crc hacks.
Remove some unneeded crc hacks while making leftover upscale only.
### Rationale behind Changes

Purge/adjust crc hacks, accuracy.
### Suggested Testing Steps

Test Shadow of Rome.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSHwHack.cpp
+++ b/pcsx2/GS/Renderers/HW/GSHwHack.cpp
@@ -268,20 +268,10 @@ bool GSC_ShadowofRome(const GSFrameInfo& fi, int& skip)
 {
 	if (skip == 0)
 	{
-		if (fi.FBP && fi.TPSM == PSM_PSMT8H && (fi.FBMSK == 0x00FFFFFF))
+		if (!s_nativeres && fi.TME == 0x0001 && (fi.TBP0 == 0x1300 || fi.TBP0 == 0x0f00) && fi.FBMSK >= 0xFFFFFF)
 		{
-			// Depth issues on all renders, white wall and white duplicate characters.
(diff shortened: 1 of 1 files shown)


lightningterror:
HPO Normal
image


HPO Special:
image

refractionpcsx2:
Normal has fringing:

image


But Special has the cursed capcom lines

image

lightningterror:
So both suck, here's a dump from the demo disc.
[Shadow of Rome [Demo] (NTSC-U)_SLUS-29139_20221005201236.zip](https://github.com/PCSX2/pcsx2/files/9717986/Shadow.of.Rome.Demo.NTSC-U._SLUS-29139_20221005201236.zip)

refractionpcsx2:
yyyyup, it's a pick your poison, until somebody has a genius solution for that edge problem (the effect always starts at the edge then multiplies upward)

refractionpcsx2:
as for the speed hack part of the CRC, the framerate halves without it.

lightningterror:
Hm, then I'll adjust the crc hacks to only work when upscaling, and maybe cut off if some aren't needed.

> as for the speed hack part of the CRC, the framerate halves without it.
How much of a speed impact?

refractionpcsx2:
Testing was done, the speed impact wasn't really enough to matter, unless you're on an absolute potato, in which case, upgrade time.

Mrlinkwii (approved):

lightningterror:
Ok so this is the best outcome, removed unneeded crc hacks, while keeping the leftover upscale only.

PR #7146, opened 2022-10-07
GameDB: Fix ghosting in BlitzTech engine games

icup321:
### Description of Changes
Adds Half-Pixel Offset to BlitzTech engine games from 2006 and onwards like Reservoir Dogs, SpongeBob CFTKK and Atlantis SquarePantis. Also fixes some title formatting issues including an incorrect serial entry (SLES-55030).

### Rationale behind Changes
Fixes ghosting issues in newer BlitzTech games, discovered mainly by looking at the DB entry for Tak and the Guardians of Gross.

### Suggested Testing Steps
Test with and without Half-Pixel Offset and observe bloom/post-processing offset differences.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -5031,8 +5031,11 @@ SCKA-20096:
   gsHWFixes:
     mipmap: 1 # Better characters and environment but has a texture cache issue that makes it worse.
 SCKA-20098:
-  name: "Spongebob Squarepants - Creature from the Krusty Krab"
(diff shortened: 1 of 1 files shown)


IlDucci:
Tested Reservoir Dogs (PAL version, SLES-53775). Hardware mode has a number of additional issues right now on postprocessing effects, with weird lines cropping during the Pause menu and the special bullet festival mode, but there's an area where Half-Pixel Offset actually fixes something.

v1.7.3375:
Reservoir Dogs_SLES-53775_20221007112614

PR:
Reservoir Dogs_SLES-53775_20221007113510

PR #7147, opened 2022-10-07
CI: Switch from libcurl4-gnutls-dev to libcurl4-openssl-dev

stenzek:
### Description of Changes

Hopefully fix the Linux AppImage not finding certificates on Debian.

### Rationale behind Changes

Fix RetroAchievements not working.

### Suggested Testing Steps

Test on Debian and Ubuntu.


stenzek:
Tested on both Ubuntu 20.04 and Debian 11, fixes the issue.

PR #7157, opened 2022-10-08
GS/HW: Only allocate as many mip levels as present

stenzek:
### Description of Changes

This prevents any possibility of sampling undefined or garbage/reused texture data from levels which aren't used by the current draw.

And a tiny bit of VRAM. But nothing really measurable.

Also fixes a bug in Vulkan where a copied texture can get mip-generated in the wrong command buffer when using basic mipmapping.

### Rationale behind Changes

My quest to make mipmapping decent in hardware.

### Suggested Testing Steps

Test mipmap games on basic and full. Off should be unchanged.


refractionpcsx2 (approved):

refractionpcsx2:
Went through a whole bunch of dumps, especially ones I know required mipmapping, no regressions noticed, all looks good :)

Issue #5069, opened 2021-11-26
Meta: Edge Anti-Aliasing Issues (AA1)


RedDevilus:
### Describe the Bug

Test

### Reproduction Steps

to

### Expected Behavior

be replaced

### PCSX2 Revision

a

### Operating System

Windows 10 (32bit)

### If Linux - Specify Distro

a

### CPU

a

### GPU

a

gilderoylockhart01:
Metal Gear Solid 2 (3?) and OSDSYS are also valid tests for that meta. I guess they now should be fine, I will test that later.

RedDevilus:
Doing software mode and switching edge anti-aliasing option is a good way to detect if it's that.

gilderoylockhart01:
That's the case for MGS2. But I would like if someone else that knows the game better can confirm.

gilderoylockhart01:
Ok so, not every games are fine with the hardware render yet, FFX and MGS 2 are still having minor problems with AA1 which is not the case with the software render.

refractionpcsx2:
yes they won't be, because they do AA1 on triangles, something we can't emulate easily. only games which used AA1 lines are fine now.

TransparentBlue:
Parts of the HUD in Ace Combat 4 is either the wrong color (too dark) or blinks a lot depending on whether edge anti-aliasing is turned on in SW, neither being  correct, HW mode is correct however.

gilderoylockhart01:
One title "Trapt", makes abusive use of Edge AA on triangles, to the point where entire areas of the 3d scene are missing on the hardware render (and obviously software without edge AA). Surprising nobody reported it before.

RedDevilus:
https://cdn.knightlab.com/libs/juxtapose/latest/embed/index.html?uid=eb0d1130-0113-11ec-abb7-b9a7ff2ee17c

There is a juxtapose of trapt but it is known.

gilderoylockhart01:
Sure, actually this bug happens a lot in the introduction scene:

Hardware:
![gs_20211218180621](https://user-images.githubusercontent.com/90222976/146649635-dfd71032-0087-434b-a03a-bfa80cdc8615.png)

Software without edge AA:
![Capture d’écran 2021-12-18 180735](https://user-images.githubusercontent.com/90222976/146649642-a5678ebe-f53e-4895-b551-cb650bf01492.png)

Software with edge AA:
[...]

PR #7158, opened 2022-10-08
GS: Correct Alpha min max when alpha coverage is used and minor bugfix for DATE selection.

lightningterror:
### Description of Changes

GS-hw: Add missing stencil buffer check for fba or coverage alpha DATE.
GS: When CoverageAlpha is used set Alpha min max to 128.
### Rationale behind Changes

Consistency, bugfixes.
Improves #5069 #1562
Fixes Trapt black floor on sw and hw renderers.
### Suggested Testing Steps

Testing if alpha min max still works with logs.

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -3590,68 +3590,78 @@ void GSState::CalcAlphaMinMax()
 	if (m_vt.m_alpha.valid)
 		return;
 
-	const GSDrawingContext* context = m_context;
+	int min = 0, max = 0;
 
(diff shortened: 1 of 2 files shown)


refractionpcsx2:
Seems to fix the black floor in Trapt, I guess because of optimisations related to the alpha (even in the SW renderer) it was taking the wrong path.


Berfore:
image

After:
image

refractionpcsx2 (approved):

refractionpcsx2:
Just discovered this also fixes the board in Dragon Quest & Final Fantasy in Itadaki Street, with the alpha problems on the board :)

image

Issue #2795, opened 2019-01-02
Ratchet and Clank 3 - Aquatos graphic issues


Levan7:
PCSX2 1.5.0-20181231230234 - compiled on Jan  1 2019

In the Aquatos mission if you enter select menu and exit it you will get this graphical problems.
This happens on DX11 and opengl

![pcsx2 2019-01-02 14-05-42-243](https://user-images.githubusercontent.com/5770076/50587759-a5251700-0e98-11e9-93d9-32368f122810.jpg)

after entering map menu or any other menu

![pcsx2 2019-01-02 14-05-46-252](https://user-images.githubusercontent.com/5770076/50587777-b66e2380-0e98-11e9-9099-39a58018194e.jpg)

[emuLog.txt](https://github.com/PCSX2/pcsx2/files/2720420/emuLog.txt)

[snaps.zip](https://github.com/PCSX2/pcsx2/files/2720421/snaps.zip)

MrCK1:
Can you also upload a memory card and/or savestate please?

MrCK1:
Thanks for the memcard. Closing in favor of https://github.com/PCSX2/pcsx2/issues/2088

Issue #2088, opened 2017-10-06
Ratchet and Clank series: bug in HW mode.


ghost:
***PCSX2 version:***
Last 1.5 dev

***PCSX2 options:***
All default, Only MTVU is enabled (bug also without)

***Plugins used:***
All default, GSDX OpenGL

***Description of the issue:***
The Ratchet and Clank series have a big issue in OpenGL mode.In some levels (Aquatos in the 3rd, Catacrom 4 in Gladiator), the graphics of the game disappear after visiting a weapon seller or by pausing the game in OpenGL mode

![capture](https://user-images.githubusercontent.com/26351275/31285589-fff52bee-aabb-11e7-8972-cbbf06cd266c.JPG)

***GSDUMP***
https://drive.google.com/open?id=199hYw5E6zETf4XksyztuzMtRa3-6Exlc

***How to reproduce the issue:***
Set GSDX in OpenGL mode, load Ratchet Gladiator, go in the Catacrom 4 level, pause the game and exit the pause menu. Same for Ratchet and Clank 3 in the Aquatos level.

***Last known version to work:***
Don't exist, was broken before.

***PC specifications:***
AMD fx8350  GTX 780Ti  Windows 10 64 Bit

MrCK1:
Try toggling mipmapping? Is automatic gamefixes enabled?

atomic83GitHub:
Toggling mipmapping don't solve the issue, automatic gamefixes and Widescreen patch are disabled.

Note: the DX11 render is not affected and work nice.

MrCK1:
Can you test 1.4.0 and see if the issue still exists?

atomic83GitHub:
Yes, the issue is also on 1.4.0 version.

MrCK1:
Can you go further back to 1.2.1 and 1.0.0?

atomic83GitHub:
1.0.0 don't have a OpenGL render Under GSDX and 1.2.1 crash when pausing.

MrCK1:
Are both SW and HW render modes in OGL affected?

atomic83GitHub:
Yes.

Dreadmoth:
Enabling Fast Texture Invalidation hack or Disable Depth Emulation hack seems to prevent this issue from occurring. It won't fix it if it has already occurred - rebooting the game seems to be necessary in that case.

atomic83GitHub:
Here is the blockdump from the main menu to the bug:
[...]

PR #7163, opened 2022-10-09
GS/TextureCache: Partially invalidate overlapping targets with tex-in-rt

stenzek:
### Description of Changes

Just copying my discord rambling for future reference
> so there's a target at 6a0
> and it does a 128x384 write to 6a0
> which includes the floor texture
> which kills the 6a0 source 
> but then it gets recreated later on
> I think why it's only breaking close up, is because the further ones are using level 1, which will have a different BP, and won't hit the target

The problem here is the condition for invalidating overlapping targets requires that the target be inside the copy rect, as well as having a BW of >2.. neither of which is the case for Jak.

https://github.com/PCSX2/pcsx2/blob/96071e157aaa9b8162fbb0ae39ffa16df541e5e6/pcsx2/GS/Renderers/HW/GSTextureCache.cpp#L984

So, instead, we'll leverage the tex-in-rt routine to find the overlapping area of these targets, and mark them as dirty. Basically, a lighter/hopefully less regression prone version of #5534.

As an aside, we can probably enable full mips and trilinear for Jak now. Although it's not necessary, so maybe not worth gamedb'ing.

### Rationale behind Changes

Fix corrupted textures in Jak games.
image image

image image

### Suggested Testing Steps

Test games with tex-in-rt, make sure nothing broke.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
+++ b/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
@@ -1000,6 +1000,12 @@ void GSTextureCache::InvalidateVideoMem(const GSOffset& off, const GSVector4i& r
 						continue;
 					}
 				}
+				else if (GSConfig.UserHacks_TextureInsideRt && t->Overlaps(bp, bw, psm, rect) && GSUtil::HasCompatibleBits(psm, t->m_TEX0.PSM))
+				{
+					const SurfaceOffset so = ComputeSurfaceOffset(off, r, t);
+					if (so.is_valid)
+						t->m_dirty.push_back(GSDirtyRect(so.b2a_offset, psm, bw));
+				}
 #endif
 			}
 		}


refractionpcsx2:
Also fixes Forever Kingdom (at least as far as the GS dump goes)

image

refractionpcsx2:
Also fixes the R&C pause screen background

image

MarioSonic2987:
The same happens with R&C 2, 3 and Ratchet: Gladiator/Deadlocked. It fixes background in pause menu.
image
image
image

JanLeonard:
> 

Can't test R&C2 right now , but UYA and Deadlocked need DisablePartial invalidation to be disabled (which isn't in gamedb due to knowed issue in those games) in order to catch correct texture to be displayed in background: Example
image

GN876:
I tested Haunting Ground, unlike #5534 this doesn't fix the transparency issues in that game.

Locme:
Tested:
R&C.   Pause screen background fixed. For what it's worth, the fix didn't break ITEMS and Quick Select menus.
R&C2. Same as above. 
R&C Up your arsenal. Same as above.
Secret Agent Clank, Fixed blurring on Clank. 
Shadow of colossus - Breaks graphics.
MGS3 - AFAIK didn't break anything
Prince of Persia- Sands of time  - same as above
RE Outbreak - same as above
Conflict Desert Storm 2 - same as above
. 

**BEFORE:**
image
**AFTER:**
image

Shadow of Colossus with T-I-RT:
image
image

refractionpcsx2:
> Secret Agent Clank, Fixed blurring on Clank. 

are you sure didn't just remove the lighting effect? It looks more like it needs Half Pixel Offset to correct it, and this PR actually just gets rid of the effect (which is bad)

Locme:
> 

You are right. It removes graphics. My bad.

JordanTheToaster (approved):

refractionpcsx2:
GS Dump of Shadow of the Colossus.  it is a little flickery, sorry, but it does show the regression introduced by this PR.

[Shadow of the Colossus_SCUS-97472_20221016214200.gs.zip](https://github.com/PCSX2/pcsx2/files/9795632/Shadow.of.the.Colossus_SCUS-97472_20221016214200.gs.zip)

refractionpcsx2:
IT was merged regardless because Shadow of the Colossus doesn't need Tex in RT, and whatever we do to Tex in RT can cause problems with other games, but this one doesn't matter.

dezraj:
> It was merged regardless because Shadow of the Colossus doesn't need Tex in RT, and whatever we do to Tex in RT can cause problems with other games, but this one doesn't matter.

Thank you @refractionpcsx2 . 👍

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -1000,6 +1000,12 @@ void GSTextureCache::InvalidateVideoMem(const GSOffset& off, const GSVector4i& r
 						continue;
 					}
 				}
+				else if (GSConfig.UserHacks_TextureInsideRt && t->Overlaps(bp, bw, psm, rect) && GSUtil::HasCompatibleBits(psm, t->m_TEX0.PSM))
```
iMineLink, line 1003: I'm pretty sure there's a game (Jak X?) that hits the previous branch but it fails to recognize the overlap correctly, just for reference.

Shoegzer:
This does not fix Ratchet & Clank: UYA menus in testing here:

PIC

Test information:

- OS: Linux Mint 21 / kernel 5.15
- CPU: Intel i7-10870H
- GPU: Nvidia GTX 3070 / driver 520.56
- Commit: 4db49038c8dad7906d0279255d4f1333cce89b8b
- Non-default settings tested: DisablePartial Invalidation: On and Off

JordanTheToaster:
> This does not fix Ratchet & Clank: UYA menus in testing here:
> 
> PIC
> 
> Test information:
> 
> * OS: Linux Mint 21 / kernel 5.15
> * CPU: Intel i7-10870H
> * GPU: Nvidia GTX 3070 / driver 520.56
> * Commit: [4db4903](https://github.com/PCSX2/pcsx2/commit/4db49038c8dad7906d0279255d4f1333cce89b8b)
> * Non-default settings tested: DisablePartial Invalidation: On and Off

Yeah it requires disabling another fix that is needed so the pause menu can't be fixed currently without breaking other parts oft he game.

Shoegzer:
Thanks.  Are you referring to the comment above from @JanLeonard regarding DisablePartial Invalidation?  If so, I tested with this setting on and off but results are the same (not what I expected based on other reported results).  That said, I do see your point that it would not make sense to fix this if it breaks more important aspects of the game, the cure being worse than the disease.

JordanTheToaster:
> Thanks. Are you referring to the comment above from @JanLeonard regarding DisablePartial Invalidation? If so, I tested with this setting on and off but results are the same (not what I expected based on other reported results). That said, I do see your point that it would not make sense to fix this if it breaks more important aspects of the game, the cure being worse than the disease.

You need Tex in RT enabled and Disable Partial Invalidation disabled ideally before booting the game that results in this.

image

Shoegzer:
Thanks. I thought I had tex-in-rt enabled when I ran those last tests but it wasn't for some reason, too much flipping back and forth between settings but that's on me.

Also, I believe the issue you're referring to is [here](https://wiki.pcsx2.net/Ratchet_%26_Clank:_Up_Your_Arsenal), under "Invisible Textures in Aquatos Level".  I tested with tex-in-rt on and DisablePartialInvalidation off, with both OpenGL and Vulkan renderers, but did not observe any invisible textures on the Aquatos level.  I'm guessing there's some issue impacted by this fix that's not mentioned there?

JanLeonard:
The two main issues disabling partial invalidation on Gladiator/UYA are:
#2088
#2795
But i can't reproduce it since a commit that i already found, unfurtanatly i haven't tested whole games since there, so i can't say for sure that is no longer present somewhere

PR #7179, opened 2022-10-12
System: Revamp memory allocation

stenzek:
### Description of Changes

Guest memory is now mapped into a shared memory/file mapping, for use with fastmem.

64-bit and 128-bit arguments are passed by register/value instead of by reference/address.

LDL/LDR/SDL/SDR now use 64-bit GPRs instead of SSE. Because 32bit is gone.

VIF0/VIF1 code region sizes have been reduced from 32/64MB to 8MB, because that's all that they actually commit.

Hardware deficiency exceptions are gone, we check that on startup.

OOM exceptions are gone, you're not going to run out of memory on a 64bit process, and if you run out of swap, well, you're system's going to have slown to a crawl already, and other allocations will fail and crash us anyway.

CPU initialization has been simplified, since it'll never fail now.

### Rationale behind Changes

Simplifying/getting rid of code, preparation for fastmemes.

As an added bonus, PCSX2 uses ~40MB less memory.

### Suggested Testing Steps

Test a range of games, make sure nothing broke. I've checked the ps2autotests elfs for the changed instructions and they're fine, but games might do silly things..


stenzek:
There seems to still be a few games broken which I didn't check. So don't bother wasting any time testing yet, will sort it out soon.

seta-san:
just on a small test in a 2d loading screen saw an unlocked v-fps jump from ~480 to ~630... so there's that...

refractionpcsx2:
link confirmed linux is good.

PR #7188, opened 2022-10-14
Gamedb : add upscaling fixed to Call of duty 3

Mrlinkwii:
### Rationale behind Changes

helps said games 
### Suggested Testing Steps

make sure CI passes


refractionpcsx2:
Before:
image

After:
image

Also worth noting that HPO Normal seems to mess with the sprite decompression and causes corruption bars around the top left edges of the screen/.

PR #7192, opened 2022-10-15
GS: Support fractional upscale

stenzek:
### Description of Changes

This PR adds fractional scaling (non-integer) to the GS renderers. Also removes the 8x cap on upscaling, but you'll need to change it via the ini, not the UI.

Using a non-integer scale likely will break some games. But I'm not aware of any (yet). I definitely wouldn't suggest going below a 0.25x increment.

Still to be decided: there's below-native resolutions in the list currently. It might be a good idea to remove them entirely, but still allow it to be set via the ini.

**NOTE: Fractional upscaling is only supported in Qt, not wx.** The old GS config system for wx does not support float parameters, only int, and I don't feel like adding code which is going to get tossed out.

### Rationale behind Changes

It's a good compromise, especially at higher resolutions, where e.g. your GPU might struggle to do 3x, but you can do 2.5x. 

### Suggested Testing Steps

Test new scaling options, make sure nothing regressed. I haven't checked Metal at all, but the other APIs should be fine.

### DIFF
--- a/bin/resources/shaders/dx11/convert.fx
+++ b/bin/resources/shaders/dx11/convert.fx
@@ -288,9 +288,18 @@ PS_OUTPUT ps_convert_rgba_8i(PS_INPUT input)
 	int txN  = tb.x | (int(input.p.x) & 7);
 	int txH  = tb.x | ((int(input.p.x) + 4) & 7);
 
-	txN *= PS_SCALE_FACTOR;
-	txH *= PS_SCALE_FACTOR;
-	ty  *= PS_SCALE_FACTOR;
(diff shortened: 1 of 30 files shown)


tellowkrinkle (approved):
image
Working fine on Metal

stenzek:
Thanks for checking!

Mrlinkwii (approved):

Nerboruto:
is it possible to have independent horizontal and vertical? ex: horizon 1x vert 2x ...
it might be useful for some pal games that don't have the progressive patch

stenzek:
No. That's just asking for trouble.

JordanTheToaster (approved):

PR #7194, opened 2022-10-15
GS: Add hardware download mode option

stenzek:
### Description of Changes

This PR extends the "disable hardware readbacks" to four levels, instead of two.

- **Enabled/Accurate**: GS downloads will synchronize the EE thread with the GS thread, and will have the best compatibility. Same as the current "disable hardware readbacks" option being **unchecked.**
- **Disable Readbacks**: GS downloads will still synchronize the EE thread with the GS thread, but the GS thread won't synchronize with the GPU. Required for games which swap data in and out of video memory (e.g. FFX and many others).
- **Unsynchronized**: YOLO reads GS local memory on the EE thread. Whatever's in there is what gets returned to the game. This can be sufficient for some, but may usually result in corruption (e.g. FFX intros), and is not deterministic, since it depends how far behind the MTGS is. Upside is you no longer have the EE<->GS thread synchronization, which can be a large speed boost. This is the same as the current "disable hardware readbacks" option being **checked**.
- **Disabled**:  Does not perform any thread synchronization, and zeros out transfer buffers, as opposed to yolo reading out of GS memory. This means that in some cases, instead of getting garbage, you'll get black, which may be preferred.

### Rationale behind Changes

The binary option was insufficient for the range of cases games throw at the console.

### Suggested Testing Steps

Test download heavy games (e.g. KH1 waterfall, GT4 night races/loading screen, etc) with the new options.


lightningterror:
HWDisableReadbacks is still a thing in fullscreenui, and I see some leftovers in tc, and mtgs, is that intended?

PR #7206, opened 2022-10-15
GS:HW: Clean up ps_convert_rgba_8i

tellowkrinkle:
### Description of Changes
Simplifies RGBA to 8i conversion shader

### Rationale behind Changes
Nicer conversion with less texture fetches

### Suggested Testing Steps
Test games that use RGBA to 8i conversion

~~If anyone (especially someone with a slow GPU) wants to uncomment the `#define ONLY_BLUE` and see if it still yields any noticeable speedup in MGS3, that would be good too, it would be nice to get rid of that comment.~~ Testing has indicated that this no longer yields any speedup, comment removed.


lightningterror (approved):
Tested some dumps that hit the shader that I know of, seems good.

Mrlinkwii (approved):

PR #7209, opened 2022-10-16
GS:HW: Fix blend_mix + HDR

tellowkrinkle:
### Description of Changes
Fixes a bug where blend_mix + HDR would end up enabling full barriers

### Rationale behind Changes
None of the render backends expect full barriers and HDR at once, and they mess up in different ways.  Metal and OGL act like there's a single barrier, and Vulkan doesn't properly correct for the newer HDR algorithm's data format, reading full black instead.

Also, blend_mix shouldn't be enabling full barriers

### Suggested Testing Steps
Test [TalesOfDestiny_BlendMixHDR.gs.xz.zip](https://github.com/PCSX2/pcsx2/files/9793545/TalesOfDestiny_BlendMixHDR.gs.xz.zip)

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
@@ -2764,7 +2764,7 @@ void GSRendererHW::EmulateBlending(bool& DATE_PRIMID, bool& DATE_BARRIER, bool& 
 			accumulation_blend = false;
 			blend_mix          = false;
 		}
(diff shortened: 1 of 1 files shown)


lightningterror:
Seems to break characters on dx11/12 on dbz bt3, more like it uncovered an issue.

[DBZ bt3 dark character.zip](https://github.com/PCSX2/pcsx2/files/9795282/DBZ.bt3.dark.character.zip)

PR #7212, opened 2022-10-16
GS: Adjust default renderer for vendors on windows.

lightningterror:
### Description of Changes

GS: Adjust default renderer for vendors on windows.
AMD, feature level 12.0 and higher pick Vulkan by default.
Intel, feature level 12.0 and higher, pick OpenGL by default.
### Rationale behind Changes

Auto pick the best api for user.
For amd/intel case we already tell them to switch so this will be less of a hassle.
### Suggested Testing Steps

Make sure default renderers are picked properly.


slx7R4GDZM:
What about Intel's new Arc GPUs? Aren't they known to run poorly with anything but DX12 or Vulkan?

lightningterror:
That's a good point. We have no data how Arc runs.

JordanTheToaster (approved):

Mrlinkwii (approved):

PR #7228, opened 2022-10-18
GS/Vulkan: Skip first barrier when starting pass

stenzek:
### Description of Changes

When we're the first draw in a render pass, we don't need a barrier. For `one` barriers, we can skip it entirely. For full, we can skip the first barrier.

### Rationale behind Changes

Miniscule performance improvement, maybe.

### Suggested Testing Steps

Check a few blending games, make sure nothing broke. Check for a reduction in barriers - but probably won't be many.


lightningterror (approved):
Checked a handful of dumps, didn't spot issues.

PR #7230, opened 2022-10-18
ImGuiOverlays: Add Texture in RT

JordanTheToaster:
### Description of Changes
Adds Texture in RT to the imgui overlay.

### Rationale behind Changes
Missing settings in overlay is bad.

### Suggested Testing Steps
Make sure CI is happy and the setting appears in the overlay.


Mrlinkwii (approved):
image
LGTM

PR #7233, opened 2022-10-18
Cubeb: Fix logging comparison, causing verbose logging spam

F0bes:
### Description of Changes
When the current logging mode was set to normal (1), verbose(2) logging would (as expected) be greater than the current mode, allowing verbose logging through.

### Rationale behind Changes
The PulseAudio backend does a lot of logging and it was quite the annoyance.

### Suggested Testing Steps
Not Applicable.


kenshen112:
Tested working

Mrlinkwii (approved):

PR #6894, opened 2022-08-22
GS-HW: Expand width on output read if DBX offset

refractionpcsx2:
### Description of Changes
Expands the target read on the merge circuit when there is an offset to the framebuffer. Also update the TBP0 on double buffer change and mis detect, which caused flickering on FMVs (not all FMVs that flicker have this same issue, unfortunately)

### Rationale behind Changes
wasn't accounting for this, causing the right side of the screen to stretch in FMV's. As mentioned in previous PR's, the hardware renderer reads the entire texture, then cuts it up in the merge circuit, so we need to over-read.

Also bad pointer matches are bad.

### Suggested Testing Steps
test FMV's which went stretchy on the right side or flickering.

Effect is seen on (for example)

Chaos Field (Also mentioned in #6813)
DJ Decks ( Fixes #4409 )
Mashed - Fully Loaded
Meiwaku Seijin Panic Maker
Sniper 2 ( Fixes #6813 )
Taiko no Tatsujin: Tatacon de Dodon ga Don
Taiko no Tatsujin: Doki! Shinkyoku Darake no Haru Matsuri
Taiko no Tatsujin: Appare Sandaime
Taiko no Tatsujin: Waku Waku Anime Matsuri
Worms 4 Mayhem
Yamasa Digi World 4 & Yamasa Digi World 4 DX
Yuusei kara no Buttai X ( Fixes #6884 )
Yuusei kara no Buttai X - Episode II

PR #7251, opened 2022-10-20
GS/TextureCache: Use Inside() for checking display target

### LINKED ISSUE #6976: [BUG]: Silent Hill 2 cutscene broken/not rendering (Vulkan, D3D11)
### Describe the Bug

In Silent Hill 2 Greatest Hits (USA) version, a cutscene mid-game does not render: Sound keeps playing, but whenever the pre-rendered/animated parts of the cutscene plays, screen goes black. When I switch the rendering mode to Software, the problem disappears, but present in D3D11 and Vulkan (haven't tried other rendering options) rendering modes.

### Reproduction Steps

1. Set the rendering options according to the images below.

2. Load the game: Silent Hill 2 Greatest Hits (USA)
3. Load the savestate: 
[SLUS-20228 (FE06A030).01.p2s.zip](https://github.com/PCSX2/pcsx2/files/9465449/SLUS-20228.FE06A030.01.p2s.zip)

4. You will see the main character standing in front of a hole. Jump in there by pressing (X) button (Dualshock 4).
5. The cutscene will play.

### Expected Behavior

When a pre-rendered/animated scene plays, the screen will go black, the sound will keep playing. This will happen several times in [that cutscene](https://www.youtube.com/watch?v=9_P-V4JqAFg).

### PCSX2 Revision

v1.7.3269

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

### CPU

Ryzen 3 3100

### GPU

GTX 1650

### GS Window Screenshots

_No response_

### Logs & Dumps

_No response_

stenzek:
### Description of Changes

The existing code was adding however many pages the framebuffer crossed unconditionally, when practically this last row will only be read when the height isn't page-aligned.

> if we had a 64x32 FB with a FBW of 1, that's 1x1 pages, so we'd compute the end as bp + (1 << 5) + ((1 * 1) << 5), which is two pages
> if it's not paged aligned, then yeah, you'd want to go one more "row" of pages down, but not take the full stride (in case the width mod 64 is not zero)

### Rationale behind Changes

Fixes #6976.

### Suggested Testing Steps

Test sensitive games.

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -2046,9 +2046,9 @@ void GSState::Write(const u8* mem, int len)
 	const int page_width = std::max(1, ((w + static_cast<int>(m_env.TRXPOS.DSAX)) / psm.pgs.x));
 	const int page_height = std::max(1, ((h + static_cast<int>(m_env.TRXPOS.DSAY)) / psm.pgs.y));
 	const int pitch = (std::max(1U, blit.DBW) * 64) / psm.pgs.x;
-
+	const u32 end_bp = blit.DBP + ((((page_height % psm.pgs.y) != 0) ? (page_width << 5) : 0) + ((page_height * pitch) << 5));
 	// Try to avoid flushing draws if it doesn't cross paths
-	m_mem.m_clut.InvalidateRange(blit.DBP, blit.DBP + ((page_width << 5) + ((page_height * pitch) << 5)));
+	m_mem.m_clut.InvalidateRange(blit.DBP, end_bp);
 }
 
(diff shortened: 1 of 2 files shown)


refractionpcsx2:
Fixes the regression I made on Jonny Moseley Mad Trix
Silent Hill 2 videos also seem fine :)

Also tried some of the older fixed videos from #6894 and the ones I tested are still fine :)

Grandia Xtreme is still broken, but to be honest we should have it check any textures which are a frame and haven't been written to for 2+ frames and remove them, as I think the problem is misdetection.

JordanTheToaster (approved):

Issue #3805, opened 2020-10-07
Meta: Games that need switch to SW for FMV Gamefix.


RedDevilus:
The current gamefix is a hack and not a proper solution.

Games affected:

- [ ] God of War
- [ ] Silent Hill 2
- [ ] Kingdom Hearts Re: Chain of Memories
- [ ] Silent Hill 3
- [ ] Twisted Black Metal

yxmline:
Gradius V [NTSC-J]
`The second level cutscenes FMV animation flower screen`
Grandia III [NTSC-J]
`FMV animation top flickering and flower screen`
Sakura - Yuki Gekka [NTSC-J]
`FMV animation flicker`
Final Fantasy X-2 - International + Last Mission [NTSC-J]
`The bottom of the animation flickers and the screen is dark`
Red Ninja - End of Honor [NTSC-J]
`FMV animation flicker`
Summon Night Gran-These: Horobi no Ken to Yakusoku no Kishi [NTSC-J]
`FMV animation flicker`

tadanokojin:
Silent Hill 2 and 3 should be sorted:
https://wiki.pcsx2.net/Silent_Hill_2
https://wiki.pcsx2.net/Silent_Hill_3

I'm guessing since the gamefix in the gamedb mentions these two they originally had the gamefix automatically applied but they were removed when they were fixed proper.

KH3, again same deal it's been fixed without the hack:
https://wiki.pcsx2.net/Kingdom_Hearts_Re:_Chain_of_Memories

Twisted metal black I don't see any mention of FMV issues on the wiki but I tested it and it's fine as far as I can tell.

Rule of Rose should be fixed:
https://wiki.pcsx2.net/Rule_of_Rose

I didn't test it but ape escape 2 and 3 should be worked around automatically now via safe features:
https://wiki.pcsx2.net/Ape_Escape_3
https://wiki.pcsx2.net/Ape_Escape_2

kozarovv:
According to wiki: Grandia Xtreme - Enable manual gamefix to switch GSdx to software rendering during FMVs to get rid of flickering video. 

But report is from 1.1.0, so that can be fixed already.

DonelBueno:
Ridge Racer V: HW renderer can't render the opening FMV.

Partially https://github.com/PCSX2/pcsx2/issues/3278

tadanokojin:
@DonelBueno that's in-engine not an FMV

DonelBueno:
[...]

PR #7253, opened 2022-10-20
GS: Only allow recently updated frames for display.

refractionpcsx2:
### Description of Changes
Make sure matched target for output as the display frame isn't too old (likely not been touched for a long time) on initial lookup.

### Rationale behind Changes
Output fames should be updated at least once every 4 frames, and the first lookup on the TC only checks the start address is the same and it ends after that address somewhere, it has no size equality or anything (this is done in later checks), so if something else existed in that space in the texture cache, it can be errantly picked up as a valid texture, causing flashing in some FMV's. This will limit the check to only frames 4 frames old or less. if the frame genuinely hasn't been updated (loading screen etc), it should match on start & end addresses (a later check).

### Suggested Testing Steps
Test movies in games, and make sure things like loading screens do not flicker.

Possibly could fix anything from https://github.com/PCSX2/pcsx2/issues/3805

Fixes FMV's in Grandia Xtreme.

Also slipped in some updates for The Simpsons Game changing the upscaling as HPO normal causes the FMV's to jump around.
Temp logging added for testing.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -6409,8 +6409,6 @@ SCPS-55008:
 SCPS-55010:
   name: "Grandia Xtreme"
   region: "NTSC-J"
-  gameFixes:
-    - SoftwareRendererFMVHack # Flickering in FMV.
 SCPS-55011:
   name: "Dual Hearts"
   region: "NTSC-J"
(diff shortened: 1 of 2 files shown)


refractionpcsx2:
Yep, they won't, they wrap around the end of memory, so it misses badly :P

Mrlinkwii (approved):

PR #7255, opened 2022-10-20
GS: Improve upload TEX overwrite detection

refractionpcsx2:
### Description of Changes
Improves the code which checks if the current upload is going to overwrite the texture in use.

### Rationale behind Changes
Before it checked for start block equality, which isn't always true (as is the case with Tomb Raider - Angel of Darkness)

### Suggested Testing Steps
Test games known to have CLUT problems, and games listed and any which may have software mode problems still (on the offchange it fixes them, already checked mercenaries and Pride FC to no avail).

Fixes Tomb Raider - Angel of Darkness menu (properly)

Master:
image

PR:
image

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -28725,15 +28725,6 @@ SLPM-65797:
   name: "Dragon Quest & Final Fantasy in Itadaki Street"
   region: "NTSC-J"
   compat: 5
-  patches:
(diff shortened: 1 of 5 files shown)


refractionpcsx2:
I ran all my dumps in software mode and there was one improvement

Stuntman - health bar thingy UI (top) becomes visible,

Master:
image

PR:
image

PR #7280, opened 2022-10-25
CDVD: Modify seek/rotational delay behaviour

refractionpcsx2:
### Description of Changes
Moves the rotational delay to the read following a seek, rather than being on the seek itself (not that it did anything)

### Rationale behind Changes
The rotational delay was being skipped due to the delta being 0 on the read, and the seek didn't do any delay either, so it was kinda ignored.

### Suggested Testing Steps
Test CDVD timing sensitive games, see if they're improved/broken.

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -739,7 +739,7 @@ static int cdvdTrayStateDetecting()
 	else
 		return CDVD_TYPE_DETCT; //Detecting any kind of disc existing
 }
-static uint cdvdRotationalLatency(CDVD_MODE_TYPE mode)
+static u32 cdvdRotationalLatency(CDVD_MODE_TYPE mode)
 {
 	// CAV rotation is constant (minimum speed to maintain exact speed on outer dge
 	if (cdvd.SpindlCtrl & CDVD_SPINDLE_CAV)
@@ -1100,8 +1100,9 @@ __fi void cdvdActionInterrupt()
 			cdvdUpdateStatus(CDVD_STATUS_PAUSE);
 			break;
 	}
-	cdvd.Action = cdvdAction_None;
-
+	
+	if(cdvd.Action != cdvdAction_Seek)
+		cdvd.Action = cdvdAction_None;
 	cdvdSetIrq();
 }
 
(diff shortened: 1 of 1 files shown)


fantesykikachu:
this seems very hackish to me, and will almost certainly break PS1 games. can you instead change DVD_SECTORS_PERSECOND which as far as I can tell is only used there.

kenshen112:
This will never interact with ps1 games. Ps1cd.cpp is a seperate file functionally. Pcsx2 is pcsx, it was taken and built into a ps2 emulator. The original CD code still exists as ps1cd and when ps1 games are activated that turns back on switching CDVD off completely. At least until we understand more about the controller and whether it's one or two controllers that are running either mode

fantesykikachu:
My bad, I was looking at the CD_SECTORS_PERSECOND, I always forget some PS2 games came out on CD not DVD. But my point still stands, introducing arbitrary (and undocumented) magic numbers is a bad idea.

refractionpcsx2:
It is a magic number, you're right, but you know what? I tried accurate numbers based on fact and paperwork, the DVD sectors per second etc were all from documented sources, and you know what happened? I fell in to an inverse bell curve of timing problems.  Recently Wisi (one of the reverse engineering guys) got some numbers from a real PS2, now I can't replicate that exactly as I don't have his setup, but my numbers we're about 20% off what he read, hence the "Magic number".

If you have a better idea of why things are reading too fast, then I'm really all ears and happy to listen, but I kinda see this as iterative, bringing it closer to the correct figures, while it might not be correct, it's better than it was.

pcsx2/CDVD/CDVD.cpp
```diff
@@ -1369,17 +1368,21 @@ static uint cdvdStartSeek(uint newsector, CDVD_MODE_TYPE mode)
 	}
 
 	// Only do this on reads, the seek kind of accounts for this and then it reads the sectors after
-	if (delta && !isSeeking)
+	if ((delta || cdvd.Action == cdvdAction_Seek) && !isSeeking)
 	{
-		int rotationalLatency = cdvdRotationalLatency((CDVD_MODE_TYPE)cdvdIsDVD());
-		//DevCon.Warning("%s rotational latency at sector %d is %d cycles", (cdvd.SpindlCtrl & CDVD_SPINDLE_CAV) ? "CAV" : "CLV", cdvd.SeekToSector, rotationalLatency);
+		uint rotationalLatency = cdvdRotationalLatency((CDVD_MODE_TYPE)cdvdIsDVD());
```
lightningterror, line 1373: Can be const and remove the empty row below.

refractionpcsx2: Actually, looks like I accidentally removed the logging, that's why there's a gap, but yeah will const.

PR #7311, opened 2022-11-01
Gamedb: fix 'The Bouncer' name

johanaxel007:
### Description of Changes
Fixes the name of 'The Bouncer' for the three regions.

### Rationale behind Changes
Correct & clean names are nice in the new QT front-end.

### Suggested Testing Steps
Check if name is indeed correct & CI is happy.


refractionpcsx2:
No, the entire indexing system puts , The at the end, else you end up with a million games starting with The and no decent indexing system does that.

PR #7323, opened 2022-11-03
GS: Restrict age of targets for GS Download invalidation.

refractionpcsx2:
### Description of Changes
Restricts the age of a target for download to 30 frames (probably still too much tbh!)

### Rationale behind Changes
While #7283 improved the detection of render targets that were required for download, it also introduced some misdetection because of old targets hanging around in the texture cache. So like the final display target lookups, we can restrict this one to 30 frames or less.  Armored Core - Last Raven was affected by this (targets were 371 frames old!).

### Suggested Testing Steps
Test games with GS downloads (readbacks, in most cases), make sure nothing is broken.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
+++ b/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
@@ -1072,7 +1072,8 @@ void GSTextureCache::InvalidateLocalMem(const GSOffset& off, const GSVector4i& r
 			// Checking for targets that overlap with the requested memory region
 			// instead of just comparing TBPs should fix that.
 			// For example, this fixes Judgement ring rendering in Shadow Hearts 2.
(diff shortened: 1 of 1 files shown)


JordanTheToaster (approved):

lightningterror (approved):

Danke-Boi:
Suggested Testing Steps
Test games with GS downloads (readbacks, in most cases), make sure nothing is broken.

@refractionpcsx2
Would GT4 be a useful test here? It did have some serious issues with readbacks.

refractionpcsx2:
It's worth testing yes, though depends what issues you mean

Danke-Boi:
> It's worth testing yes, though depends what issues you mean

Just massive performance drops the second a readback starts

refractionpcsx2:
oh right, this isn't going to fix that, no, this will only fix graphical issues.

Danke-Boi:
> oh right, this isn't going to fix that, no, this will only fix graphical issues.

Yeah, I checked it, no additional issues anyways. Doesn't improve performance any, like you said.

knight-ryu12 (approved):

PR #7339, opened 2022-11-04
ci/lint: remove GameDB python lint script, replace with a json-schema

xTVaser:
### Description of Changes

Removes the manually created GameDB linter script infavor of a JSON schema

I also noticed the follow regions are either wrong, or not recognized by https://github.com/PCSX2/pcsx2/blob/808fadd0617e0259a6a48a23d83005241bc0a739/pcsx2/Frontend/GameList.cpp#L248

- `NTSC-E`
- `PAL-Unk`
- `PAL`
- `NTSC-M`

### Rationale behind Changes

JSON schema is a simple, declarative way to lint things like this and provides some other useful features:
- code generation
- documentation generation
- editor integrations, (auto-complete / in-editor lints)
- less code to maintain (the script)

### Suggested Testing Steps

CI should pass


RedDevilus:
NTSC-Unk was used a long time (not sure before redump) where the region was an unknown (Asian) NTSC.

refractionpcsx2:
There's probably a few missing.

PR #7344, opened 2022-11-06
GS: On Local->Local transfer wrap around 2048

refractionpcsx2:
### Description of Changes
Wrap transfers around 2048 pixels in either direction.

### Rationale behind Changes
if you had a really high offset with a really big transfer, it could go out of bounds and cause things to explode.

### Suggested Testing Steps
Test games, make sure nothing is busted.

This *might* fix the random crash in Silent Hill Origins (I believe near the end of the game)

### DIFF
--- a/pcsx2/GS/GSState.cpp
+++ b/pcsx2/GS/GSState.cpp
@@ -2193,21 +2193,21 @@ void GSState::Move()
 
 				for (int y = starty; y != endy; y+= y_inc, _sy += y_inc, _dy += y_inc)
 				{
-					auto s = getPAHelper(spo, sx, _sy);
-					auto d = getPAHelper(dpo, dx, _dy);
+					auto s = getPAHelper(spo, 0, _sy);
+					auto d = getPAHelper(dpo, 0, _dy);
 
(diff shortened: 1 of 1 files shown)


pcsx2/GS/GSState.cpp
```diff
@@ -2193,21 +2193,21 @@ void GSState::Move()
...
-					auto d = getPAHelper(dpo, dx, _dy);
+					auto s = getPAHelper(spo, 0, _sy);
+					auto d = getPAHelper(dpo, 0, _dy);
 
 					if (((sx < dx) && ((xpage + page_width) > dx)))
 					{
 						for (int x = w - 1; x >= 0; x--)
 						{
-							pxCopyFn(d, s, x);
+							pxCopyFn(d, s, (dx + x) % 2047, (sx + x) % 2047);
```
hch12907, line 2203: Shouldn't it be `& 2047`?

refractionpcsx2: It should, I thought about that when I went to bed last night, never code at 5am

Issue #6797, opened 2022-08-07
[BUG]: Growlanser 5 & 6 - "Preload Frame Data" HW Hack not working since pcsx2-v1.7.2908


Risae:
### Describe the Bug

Growlanser 5 and 6 suffers from this bug: #4925

It used to be possible to activate the Hardware Hack called `Preload Frame Data` to fix this issue, but since `pcsx2-v1.7.2908` #5349 this does not work anymore. I tested the up-to-date `pcsx2-v1.7.3180-windows-64bit-AVX2-wxWidgets` and the issue still persists.

### Reproduction Steps

Play Growlanser 5 or 6 using `pcsx2-v1.7.2908` and enable the `Preload Frame Data` hack. The issue is not present in `pcsx2-v1.7.2907`.

### Expected Behavior

The hack works as it did in pre `pcsx2-v1.7.2908`.

### PCSX2 Revision

`pcsx2-v1.7.3180-windows-64bit-AVX2-wxWidgets`

### Operating System

Windows 10 (64bit)

### CPU

Ryzen 2700X

### GPU

RTX 2070

### GS Settings

Default settings with Preload Frame Data Hack manually enabled.

### Emulation Settings

Default settings with Preload Frame Data Hack manually enabled.

### GS Window Screenshots

Growlanser 5 US using `pcsx2-v1.7.2907` with `Preload Frame Data` enabled:
![gs_20220807163826_Growlanser - Heritage of War _NTSC-U__SLUS-21571](https://user-images.githubusercontent.com/37260321/183296261-4c031e37-f9b3-4dc6-916e-8cc179e4a516.png)

Growlanser 5 US using `pcsx2-v1.7.2908` with `Preload Frame Data` enabled:
![gs_20220807164027_Growlanser - Heritage of War _NTSC-U__SLUS-21571](https://user-images.githubusercontent.com/37260321/183296390-88495904-2414-4d2d-a42e-1ff574d09f5e.png)

Growlanser 5 US using `pcsx2-v1.7.3180` with `Preload Frame Data` enabled:
![Growlanser - Heritage of War _NTSC-U__SLUS-21571_20220807164242](https://user-images.githubusercontent.com/37260321/183296522-d8044eb0-ddec-4e2d-b67d-7d343c45012c.png)

### Logs & Dumps

GS Dump:
[...]

PR #7349, opened 2022-11-07
GameDB: Disable gsHWFixes for Growlanser 5 & 6

Risae:
### Description of Changes
Disable gsHWFixes for Growlanser 5 and 6 because of #6797

### Rationale behind Changes
gsHWFixes do not fix the background issues anymore.

### Suggested Testing Steps
Make sure CI is happy.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -21165,8 +21165,9 @@ SLES-55214:
 SLES-55215:
   name: "Growlanser - Heritage of War"
   region: "PAL-E"
-  gsHWFixes:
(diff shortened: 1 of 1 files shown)


Mrlinkwii:
wouldn't just be better for Preload Frame Data be looked at , and potentially  fixed rather than removing?

Risae:
> wouldn't just be better for Preload Frame Data be looked at , and potentially fixed rather than removing?

Kindly take a look at the analysis of @iMineLink in the Github issue - As far as i understand (i'm not a programmer) it was just coincidence that the PCSX2 HW Hack `Preload Frame Data` worked in the past. Right now the HW Hack doesn't do anything and i don't believe its a good idea to activate hacks that are kinda useless.

PR #7360, opened 2022-11-09
GameDB: Add CPU CLUT Normal to Lego Starwars 2

JordanTheToaster:
### Description of Changes
Adds CPU CLUT normal to Lego Starwars 2 to fix ghosting.

### Rationale behind Changes
Without:
image

With:
image

### Suggested Testing Steps
Make sure CI is happy.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -18726,6 +18726,8 @@ SLES-54218:
 SLES-54221:
   name: "LEGO Star Wars II - The Original Trilogy"
   region: "PAL-M6"
+  gsHWFixes:
+    cpuCLUTRender: 1 # Fixes ghosting on objects and people.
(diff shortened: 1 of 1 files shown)


Mrlinkwii (approved):

refractionpcsx2:
I suspect this is misdetection, but we can roll with it for now.  But it's sus :D

Bryan1150:
Bionicle Heroes has the exact ghosting problem, would this fix that too?

JordanTheToaster:
> Bionicle Heroes has the exact ghosting problem, would this fix that too?

If you post a GS dump of the issue I can look.

JordanTheToaster:
> Bionicle Heroes has the exact ghosting problem, would this fix that too?

Scratch that it does fix it added it in #7362

Issue #7367, opened 2022-11-10
[BUG]: Inhibit Screensaver not working in fullscreen mode on Linux


davidscholberg:
### Describe the Bug

The inhibit screensaver functionality on Linux (tested on an up to date Fedora system using Gnome X11) only seems to work when a game is not in fullscreen mode. As soon as you fullscreen a game, the screen will blank after the configured time has elapsed. After you close a game, PCSX2 will print a message saying `xdg-screensaver: Window 0x3400016 does not exist`, with varying window IDs.

As a workaround, I can get the functionality to work even in fullscreen mode by manually grabbing the PCSX2 window ID from the output of `wmctrl -l` and then running `xdg-screensaver suspend [ID goes here]`. So I wonder if possibly the wrong window ID is being used.

### Reproduction Steps

1. Set the Inhibit Screensaver setting.
2. Observe that the screen blanks when a game is in fullscreen mode.

### Expected Behavior

I would expect Inhibit Screensaver to work even in fullscreen mode.

### PCSX2 Revision

v1.7.3547 (AVX2 Qt)

### Operating System

Linux (64bit) - Specify Distro Below

### If Linux - Specify Distro

Fedora 36

PR #7375, opened 2022-11-11
Qt: fix inhibit screensaver on Linux (X11)

davidscholberg:
### Description of Changes
This change fixes https://github.com/PCSX2/pcsx2/issues/7367. `xdg-screensaver` appears to only work properly in all cases if you give it the top level window ID of PCSX2.

As far as I can tell, this change is only relevant to the Qt client on Linux with X11.

### Rationale behind Changes
With this change, inhibit screensaver now works with the Qt client on Linux (X11) when a game is fullscreen as well windowed.

### Suggested Testing Steps
Make sure the screensaver is inhibited while a game is both fullscreen and windowed.

### DIFF
--- a/pcsx2/Frontend/CommonHost.cpp
+++ b/pcsx2/Frontend/CommonHost.cpp
@@ -430,8 +430,9 @@ void CommonHost::UpdateInhibitScreensaver(bool inhibit)
 		return;
 
 	WindowInfo wi;
-	if (g_host_display)
-		wi = g_host_display->GetWindowInfo();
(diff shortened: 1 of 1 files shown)


stenzek:
There's already `Host::GetTopLevelWindowInfo()`, would be better to just reuse this and change the inhibt call site, rather than introducing a field into WindowInfo for one function on one platform.

davidscholberg:
@stenzek PR updated.

stenzek (approved):

PR #7376, opened 2022-11-11
Implemented Motion Adaptive Deinterlacing and integrated changes with UI

sideprojectslab:
### Description of Changes
Adds Motion Adaptive Deinterlacing as an additional deinterlacing option (new default).

### Rationale behind Changes
Blend deinterlacing tends to blur objects in motion which isn't very desirable, this new method reduces visual artifacts in games that do not output progressive frames and attempts to provide clear and smooth motion, resulting in a superior experience.

### Suggested Testing Steps
Test games which are generally quite blurry, such as GT4, King's Field VI etc. Also check random games to make sure nothing is broken (confirm it's not broken with "none" deinterlacing before reporting)

### DIFF
--- a/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
@@ -108,7 +108,7 @@ GraphicsSettingsWidget::GraphicsSettingsWidget(SettingsDialog* dialog, QWidget* 
 		sif, m_ui.aspectRatio, "EmuCore/GS", "AspectRatio", Pcsx2Config::GSOptions::AspectRatioNames, AspectRatioType::RAuto4_3_3_2);
 	SettingWidgetBinder::BindWidgetToEnumSetting(sif, m_ui.fmvAspectRatio, "EmuCore/GS", "FMVAspectRatioSwitch",
 		Pcsx2Config::GSOptions::FMVAspectRatioSwitchNames, FMVAspectRatioSwitchType::Off);
-	SettingWidgetBinder::BindWidgetToIntSetting(sif, m_ui.interlacing, "EmuCore/GS", "deinterlace", DEFAULT_INTERLACE_MODE);
+	SettingWidgetBinder::BindWidgetToIntSetting(sif, m_ui.interlacing, "EmuCore/GS", "deinterlace_mode", DEFAULT_INTERLACE_MODE);
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.bilinearFiltering, "EmuCore/GS", "linear_present", true);
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.integerScaling, "EmuCore/GS", "IntegerScaling", false);
 	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.PCRTCOffsets, "EmuCore/GS", "pcrtc_offsets", false);
--- a/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
@@ -75,7 +75,7 @@ static const char* s_anisotropic_filtering_entries[] = {QT_TRANSLATE_NOOP("Graph
 	QT_TRANSLATE_NOOP("GraphicsSettingsWidget", "8x"), QT_TRANSLATE_NOOP("GraphicsSettingsWidget", "16x"), nullptr};
 static const char* s_anisotropic_filtering_values[] = {"0", "2", "4", "8", "16", nullptr};
 
-static constexpr int DEFAULT_INTERLACE_MODE = 7;
+static constexpr int DEFAULT_INTERLACE_MODE = 0;
 static constexpr int DEFAULT_TV_SHADER_MODE = 0;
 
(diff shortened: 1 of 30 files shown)


pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
```diff
@@ -75,7 +75,7 @@ static const char* s_anisotropic_filtering_entries[] = {QT_TRANSLATE_NOOP("Graph
 	QT_TRANSLATE_NOOP("GraphicsSettingsWidget", "8x"), QT_TRANSLATE_NOOP("GraphicsSettingsWidget", "16x"), nullptr};
 static const char* s_anisotropic_filtering_values[] = {"0", "2", "4", "8", "16", nullptr};
 
-static constexpr int DEFAULT_INTERLACE_MODE = 7;
+static constexpr int DEFAULT_INTERLACE_MODE = 0;
```
RedDevilus, line 78: Shouldn't this be different? As far as I know interlacing 0 is none interlacing.
```suggestion
static constexpr int DEFAULT_INTERLACE_MODE = 8;
```

RedDevilus: Ok never mind, You changed the value. My bad.

Was looking at the GameIndexMD:
`        {Automatic Off, WeaveTFF, WeaveBFF, BobTFF, BobBFF, BlendTFF, BlendBFF, AdaptiveTFF, AdaptiveBFF}`

RedDevilus:
If you press F5 which cycles through deinterlace modes:

image

Others do say weave/bob/blending/...

Ziemas:
Probably needs to be added here as well? https://github.com/PCSX2/pcsx2/blob/5321ca913e4145f78eb2fe161869442f71ffb895/pcsx2/Frontend/FullscreenUI.cpp#L2795

RedDevilus:
Yeah, that spot only has 8 values when there should be 9.

refractionpcsx2:
The fields seem backwards for this alpine game compared to blend and weave

[gs_20220409060300_Alpine Racer 3 _PAL-M5__SCES-50887.zip](https://github.com/PCSX2/pcsx2/files/9993921/gs_20220409060300_Alpine.Racer.3._PAL-M5__SCES-50887.zip)

refractionpcsx2:
NTSC BIOS and games when using screen offsets and show overscan will be of a height 486 which divides to 243 and causes the screen to shake (I think I remember you saying the division was a problem?)

sideprojectslab:
> deinterlace options from the gamedb aren't applying, you need to change this line in GameDatabase.cpp
> 
> if (value >= 0 && value <= static_cast(GSInterlaceMode::Automatic))
> 
> to
> 
> if (value >= static_cast(GSInterlaceMode::Automatic) && value < static_cast(GSInterlaceMode::Count))
> 
> That aught to do it :)



> NTSC BIOS and games when using screen offsets and show overscan will be of a height 486 which divides to 243 and causes the screen to shake (I think I remember you saying the division was a problem?). I'd say fudge the overscan value by 1 (it's 243 in there) but that only fixes native, not some of the fractional upscales, and it's not really solving the problem, just avoiding it :P

odd line numbers should not be an issue any more, I will investigate

sideprojectslab:
> 1The fields seem backwards for this alpine game compared to blend and weave, you need BFF to see the picture, rather than TFF like the others
> 
> [gs_20220409060300_Alpine Racer 3 _PAL-M5__SCES-50887.zip](https://github.com/PCSX2/pcsx2/files/9993921/gs_20220409060300_Alpine.Racer.3._PAL-M5__SCES-50887.zip)
> 
> You can tell, cos it's a black screen :) Though I don't know what the best course of action this is, basically the game uses SCANMSK to render odd or even lines every frame, and if you don't get them synced up, it just does a black screen.
> 
> Changing in ps_main4
> 
> if ((optr.y >= 0.0f) && (optr.y < 0.5f) && ((vpos & 1) == field)) to if ((optr.y >= 0.0f) && (optr.y < 0.5f) && ((vpos & 1) != field))
> 
> does resolve this, but it kinda breaks MAD and causes it to start weaving on king's field :( This might be a tough one to resolve unless you can work out a way to modify MAD without killing it. Don't worry if it can't, if alpine breaks, I suspect that's an outlier, I'm not sure how it works with the other two deinterlace settings. But I suspect anything you do might break Kings field, which looks great right now (and scanmsk on mgs3 and 2 are good too right now), maybe King's Field would look better if we flipped the weave shaders too :D

Maybe we have to flip stuff around, indeed the new shader behaves opposite to the old weave shaders. It's just trial and error at this point but at the end of the day it has to do with which index corresponds to which field

lightningterror:
@sideprojectslab Can you rebase the pr?

lightningterror:
Hm, should've checked before this was merged but there seems to be some new warnings introduced
https://github.com/PCSX2/pcsx2/actions/runs/3500364002/jobs/5862952547#step:11:525

KainXVIII:
Dark Cloud 1 ground textures looks really weird with Motion Adaptive Deinterlacing, no-interlace patch is much better for now..

Berylskid:
I tested Armored Core 2 (SLPS-25007) with default settings and found two problems with the menu screen display.

- Background animation flickers.
- Menu items vibrate by highlighted (blue -> orange).

These could not be reproduced with v1.7.3573.

PR #7377, opened 2022-11-11
GS-TC: Invalidate videomem still if surface becomes invalid

refractionpcsx2:
### Description of Changes
Checks if the surface has actually become invalid by the write, if so, delete it.

### Rationale behind Changes
Code was added in #5349 for Destruction Derby Arena to get the correct area of a surface and set the exact parts which needed invalidating, unfortunately it doesn't then check if it's still valid after the write, where it does actually need deleting, which broke Onimusha Warlords, but it was masked by an accidental double flush in the Host->Local write function.

### Suggested Testing Steps
Test foggy stuff in Destruction Derby Arenas (have already but good to double check) and check Onimusha Warlords with texture preloading set to none or partial (there's a bug in full atm :( ) and make sure there's no repeating happening or partially transparent characters.

Test other stuff To I guess

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
+++ b/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
@@ -903,6 +903,10 @@ void GSTextureCache::InvalidateVideoMem(const GSOffset& off, const GSVector4i& r
 				else
 				{
 					// YOLO skipping t->m_TEX0.TBW = bw; It would change the surface offset results...
+					// This code exists because Destruction Derby Arenas uploads a 16x16 CLUT to the same BP as the depth buffer and invalidating the depth is bad (because it's not invalid).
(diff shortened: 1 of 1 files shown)


ps1freak26:
I haven't noticed anything wrong with dda on v1.7.3550 with dx12.

refractionpcsx2:
There isn't anything wrong, but this PR modifies the code which fixed DDA, so it was making sure I haven't broken it ,not that it fixes anything in DDA.

pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -928,7 +932,7 @@ void GSTextureCache::InvalidateVideoMem(const GSOffset& off, const GSVector4i& r
 							r.w
 						);
 					}
-					else
+					if(!ComputeSurfaceOffset(off, r, t).is_valid)
 					{
```
lightningterror, line 936: Space after if.

Issue #7222, opened 2022-10-17
[BUG]: Incorrect PNACH extended behaviour


EdnessP:
### Describe the Bug

The D-codetype in the <tt>extended</tt> pnach mode, which from my understanding is supposed to emulate the behaviour of [PS2RD on a real PS2](https://github.com/mlafeldt/ps2rd/blob/master/Documentation/code_types.txt), isn't being parsed correctly.

While really any scenario would be affected, this is the following patch I've written for the Japanese version of Burnout 3: Takedown (SLPM-65719) that patches the main font with an unused font that is loaded in RAM temporarily
<pre>
D0488874 05001640 // Check if it's loaded (original ptr)
50DD1890 000200C0 // Copy texture data bytes
00495A90 00000000
20488874 00495840 // Overwrite 3rev pointer DD1640->495840
4068A3D0 00020002 // Overwrite fontmap pointers (main, kbd)
00488870 00000000
</pre>

The 1st line is supposed to only allow the next 5 lines to execute if the 16-bit value at 00488874 is 1640 which becomes true after said texture dictionary is loaded, and then those lines copy the data, overwrite pointers, including the one being originally checked just so the PS2 wouldn't be constantly and needlessly copying 128KiB every frame.
**This works without issue on a real PS2**, but on PCSX2 (tried 1.4, 1.6, 1.7.2225, and 1.7.3442) it seems to constantly be copying those 128KiB even long after the game has overwritten the data with other stuff, which then causes corruption like this and in the end - a complete hang.
![pcsx2_2022-10-17-231720](https://user-images.githubusercontent.com/55930127/196288997-0b67094e-6395-4fa0-a76c-1a94a0555460.png)

### Reproduction Steps

In my case, entering in-game on the Japanese version of Burnout 3: Takedown (SLPM-65719) with this PNACH cheat active, but in theory any case where the D-codetype is supposed to block multiple lines from executing if the condition is not met.
<pre>
patch=1,EE,D0488874,extended,05001640
patch=1,EE,50DD1890,extended,000200C0
patch=1,EE,00495A90,extended,00000000
patch=1,EE,20488874,extended,00495840
patch=1,EE,4068A3D0,extended,00020002
patch=1,EE,00488870,extended,00000000
</pre>

### Expected Behavior

[...]

PR #7381, opened 2022-11-12
Patches: Add extended PS2rd support; used by OPL

EdnessP:
### Description of Changes
Adds additional extended codetype functionality that exists in [PS2rd](https://github.com/mlafeldt/ps2rd/blob/master/Documentation/code_types.txt) and is used by Open PS2 Loader, namely:
* Making use of the otherwise unused last 4 digits of the D-codetype data value to also allow for 8-bit and 16-bit multi-line conditionals.
This won't break compatibility with existing D-codes, because PS2rd itself also checks if the lines to skip is 0, and sets it to 1.
* Adds more test conditions 4 - 7 used by PS2rd - NAND, AND, NOR, OR for both D-codetypes and E-codetypes.

PS2rd also has one more C-codetype which is a 32-bit conditional that will block *every* following cheat line if the condition is not met, but at a glance I couldn't see if the function had any way of tracking which line out of how many total it's currently on to know how large to set the SkipCount that would include all the other codes after it.  Not implemented for now.

### Rationale behind Changes
Solves issue #7222 and saves headaches caused by people ~~(me)~~ not being aware that these were only available in PS2rd/OPL, and wondering why cheats and codes I've written work just fine for me on a real PS2, but not for others on PCSX2.

The PS2rd codetypes have stood their ground since 2009, so I'd say it has merit to be supported by PCSX2.  Not to mention OPL uses this too, which is presumably how most people nowadays use these codes on a real PS2 if they don't have access to a cheat disc.

### DIFF
--- a/pcsx2/Patch_Memory.cpp
+++ b/pcsx2/Patch_Memory.cpp
@@ -1,5 +1,5 @@
 /*  PCSX2 - PS2 Emulator for PCs
- *  Copyright (C) 2002-2010  PCSX2 Dev Team
+ *  Copyright (C) 2002-2022  PCSX2 Dev Team
  *
  *  PCSX2 is free software: you can redistribute it and/or modify it under the terms
  *  of the GNU Lesser General Public License as published by the Free Software Found-
@@ -196,7 +196,7 @@ void handle_extended_t(IniPatch *p)
 		}
 		else if ((p->addr & 0xF0000000) == 0x40000000)			// 4aaaaaaa nnnnssss + Another line
 		{
-			IterationCount = ((u32)p->data & 0xFFFF0000) / 0x10000;
+			IterationCount = ((u32)p->data & 0xFFFF0000) >> 16;
 			IterationIncrement = ((u32)p->data & 0x0000FFFF) * 4;
 			PrevCheatAddr = (u32)p->addr & 0x0FFFFFFF;
 			PrevCheatType = 0x4000;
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Thanks for doing that :)

RedDevilus:
image

Network error, so not a coding error.

pcsx2/Patch_Memory.cpp
```diff
@@ -249,41 +249,253 @@ void handle_extended_t(IniPatch *p)
 		}
 		else if (p->addr < 0xE0000000)
 		{
-			if (((u32)p->data & 0xFFFF0000) == 0x00000000)		// Daaaaaaa 0000dddd
+			if (((u32)p->data & 0x00F00000) == 0x00000000)		// Daaaaaaa yy0zdddd
 			{
-				u16 mem = memRead16((u32)p->addr & 0x0FFFFFFF);
-				if (mem != (0x0000FFFF & (u32)p->data))
+				u8 z = ((u32)p->data & 0x000F0000) / 0x10000;
```
lightningterror, line 254: Can be constant, same for the others.

refractionpcsx2: Also make it a shift rather than a divide, there's no need for it to be a divide and it's less clear.

EdnessP: Yeah, I considered doing that as well, but since the whole rest of the file used dividing instead of shifting, I also went with it to keep it consistent.

I'll get around to that hopefully by the weekend when I have some more tree time.

refractionpcsx2: You're welcome to improve the others :D

Tokman5:
https://github.com/PCSX2/pcsx2/blob/f80899117fb161e2417be7d14aea42c2a5b3b557/pcsx2/Patch_Memory.cpp#L269-L270

these can be constant?

EdnessP:
Oops, forgot to, my bad.

EdnessP:
I also got to thinking last night - what real use could the NOR and OR test conditions really have, other than checking if it's a non-zero value or not, at which point one might as well use the regular equality check for 0.

For a bit I even thought that maybe there was something more to it besides just checking if it's a zero or non-zero value, but I then [looked up the PS2rd source which shows that it is in fact just that](https://github.com/mlafeldt/ps2rd/blob/master/ee/engine/engine_asm.S#L465).  Oh well.

PR #7385, opened 2022-11-13
mVU: Clean up range function and improve merging. Slim down cmpProg.

refractionpcsx2:
### Description of Changes
Cleans up the micro program range merging functions and gets rid of some of the junk in the program comparison function. Also only copy parts of the program which are needed for the current microprogram during compilation, which should reduce compilation time slightly.

### Rationale behind Changes
Might be slightly faster in some cases? idk, but it's nicer.

### Suggested Testing Steps
Test games, make sure the VU's don't explode.

### DIFF
--- a/pcsx2/x86/microVU.cpp
+++ b/pcsx2/x86/microVU.cpp
@@ -191,7 +191,8 @@ __ri microProgram* mVUcreateProg(microVU& mVU, int startPC)
 	prog->idx = mVU.prog.total++;
 	prog->ranges = new std::deque<microRange>();
 	prog->startPC = startPC;
-	mVUcacheProg(mVU, *prog); // Cache Micro Program
+	if(doWholeProgCompare)
+		mVUcacheProg(mVU, *prog); // Cache Micro Program
 	double cacheSize = (double)((uptr)mVU.prog.x86end - (uptr)mVU.prog.x86start);
 	double cacheUsed = ((double)((uptr)mVU.prog.x86ptr - (uptr)mVU.prog.x86start)) / (double)_1mb;
 	double cachePerc = ((double)((uptr)mVU.prog.x86ptr - (uptr)mVU.prog.x86start)) / cacheSize * 100;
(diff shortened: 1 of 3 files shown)


rtj503:
Since we were working on that mysterious black triangle glitch in Xenosaga 3 which had something to do with the Vector Unit, I tested that area out on it, and there is slightly different behavior. Setting Rounding Mode to "Positive" still works, but now setting the Clamping Mode to 'Extra + Preserve Sign" also happens to work. Nice!
Evidence:
https://drive.google.com/drive/folders/15XD7Wf5kbMhQn6SQYQWMyWWE336DFBFa?usp=sharing

refractionpcsx2:
I'm not sure what changed that then, because emulation shouldn't have really changed here at all..

I would be very suspicious of that, to be honest. Are you sure it's not just the gamedb auto enabling positive rounding (since it's in there)?

rtj503:
I take it the GameDB settings are under "View" and "Game Properties" right? Nope, "Chop/Zero Default" and "Extra + Preserve Sign" works to solve the black triangle. Maybe I simply forgot to test that option last time? I don't think I did though, could have sworn I checked every option.

refractionpcsx2:
Gamedb settings don't appear in the menus. but the rounding/clamping will override anything you do in the menu, if they exist. you need to look in the Resources folder for the Gameindex.yaml file.

rtj503:
"SCAJ-20179:
  name: "Xenosaga Episode III - Also Sprach Zarathustra [Disc 1 of 2]"
  region: "NTSC-Unk"
  gsHWFixes:
    autoFlush: 1 # Fixes shadows in cutscenes.
    halfPixelOffset: 1 # Fixes lighting misalignment and reduces ground shadows (probably texture cache issue).
    roundSprite: 2 # Fixes font artifacts.
    texturePreloading: 0 # Performs much better with it off.
SCAJ-20180:
  name: "Xenosaga Episode III - Also Sprach Zarathustra [Disc 2 of 2]"
  region: "NTSC-Unk"
  gsHWFixes:
    autoFlush: 1 # Fixes shadows in cutscenes.
    halfPixelOffset: 1 # Fixes lighting misalignment and reduces ground shadows (probably texture cache issue).
    roundSprite: 2 # Fixes font artifacts.
    texturePreloading: 0 # Performs much better with it off."
That's what I got.

refractionpcsx2:
there isn't any then, huh, okay. I still don't know what would have changed that behaviour though.

rtj503:
Lol. Don't worry about it. It's just one black triangle on a ceiling I never even noticed until someone pointed it out here on GitHub. The cost of perfection is infinite. And to be fair, it did fix a problem, rather than cause a glitch to occur, it removed one. We just don't know why.

PR #7387, opened 2022-11-14
GS: More Metal fixes

tellowkrinkle:
### Description of Changes
- Fixes an issue where gpu spinning would randomly be enabled when it wasn't requested or needed (oops)
- Uses `[MTLCommandBuffer presentDrawable:]` instead of `[MTLDrawable present]` when vsync'd or on Ventura
  - Previous use was to work around a bug where `presentDrawable:` acted like vsync was on no matter what.  On the other hand, `present` acted like vsync was off no matter what.  This has been fixed in Ventura, and `presentDrawable:` plays nicer with Xcode GPU traces

### Suggested Testing Steps
Test Metal Renderer

### DIFF
--- a/pcsx2/Frontend/MetalHostDisplay.mm
+++ b/pcsx2/Frontend/MetalHostDisplay.mm
@@ -129,6 +129,13 @@ bool MetalHostDisplay::CreateDevice(const WindowInfo& wi)
 	[m_pass_desc colorAttachments][0].clearColor = MTLClearColorMake(0, 0, 0, 0);
 	[m_pass_desc colorAttachments][0].storeAction = MTLStoreActionStore;
 
+	if (char* env = getenv("MTL_USE_PRESENT_DRAWABLE"))
+		m_use_present_drawable = static_cast<UsePresentDrawable>(atoi(env));
+	else if (@available(macOS 13.0, *))
+		m_use_present_drawable = UsePresentDrawable::Always;
+	else // Before Ventura, presentDrawable acts like vsync is on when windowed
+		m_use_present_drawable = UsePresentDrawable::IfVsync;
+
 	m_capture_start_frame = 0;
 	if (char* env = getenv("MTL_CAPTURE"))
 	{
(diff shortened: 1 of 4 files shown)


pcsx2/Frontend/MetalHostDisplay.mm
```diff
@@ -293,10 +293,20 @@ static void OnMainThread(Fn&& fn)
 	ImGui::Render();
 	dev->RenderImGui(ImGui::GetDrawData());
 	dev->EndRenderPass();
+	bool use_present_drawable = false;
+	if (m_vsync_mode != VsyncMode::Off)
+		use_present_drawable = true;
+	if (@available(macOS 13, *)) // Before Ventura, presentDrawable acts like vsync is on when windowed
```
stenzek, line 299: I'm assuming the overhead from the `@available` bit is low? Not that it matters a ton, given it's called once per frame, but if not, could be worth caching it?

stenzek (approved):
Changes sounds fine otherwise, obviously I can't test it :)

pcsx2/Frontend/MetalHostDisplay.mm
tellowkrinkle: Was a bit more than I expected (~100 instructions)
Cached (and added a env variable override because why not)

Danke-Boi:
What games should be used to test this? I have a few more popular games that I can test.

tellowkrinkle:
> What games should be used to test this? I have a few more popular games that I can test.

Any game, there shouldn't be any noticeable difference (unless you're monitoring power draw, in which case it might be lower)

If you're on pre-Ventura, you can enable vsync and see if frame pacing is better now.  Ventura should see no difference.

Danke-Boi:
> > What games should be used to test this? I have a few more popular games that I can test.
> 
> Any game, there shouldn't be any noticeable difference (unless you're monitoring power draw, in which case it might be lower)
> 
> If you're on pre-Ventura, you can enable vsync and see if frame pacing is better now. Ventura should see no difference.

5 FPS drop (from 30 - 25) during readback on that one scene I used to test the "spin GPU during readback" PR.

tellowkrinkle:
> 5 FPS drop (from 30 - 25) during readback on that one scene I used to test the "spin GPU during readback" PR.

I can't reproduce that at all on my M1
Can you make sure that none of your settings have changed (e.g. MTVU on vs off, which has a ~5fps difference in that scene on my M1)?  Also, I noticed in your video the frame rate changes throughout the scene, which I think was leading to your inconsistent frame times.  If you pause and then take a save state, you can measure while paused, which has a much more consistent frame rate.

PR #7389, opened 2022-11-14
OSD: Decrease delay for messages and toasts

Andryshik345:
### Description of Changes

Decreases delay to five seconds for all OSD messages and toasts.

### Rationale behind Changes

For me, from the side of a regular player, it is not particularly pleasant to see a very long message delay, so I decided to reduce it to five seconds. I do not know why such a long delay was needed at all, because I can read this message several times, especially when you try to screenshots some moment and when you load seivestate with the right moments, these messages continue to hang long and tediously, it's annoying. Five seconds should be enough in that case.

### Suggested Testing Steps

Save/loading savestates messages, changing slots, etc.

Also some article sources to get the idea why long-delay toasts are bad:
https://www.w3.org/TR/UNDERSTANDING-WCAG20/time-limits-pause.html
https://sheribyrnehaber.medium.com/designing-toast-messages-for-accessibility-fb610ac364be

### DIFF
--- a/pcsx2/Frontend/Achievements.cpp
+++ b/pcsx2/Frontend/Achievements.cpp
@@ -549,7 +549,7 @@ void Achievements::UpdateSettings(const Pcsx2Config::AchievementsOptions& old_co
 		else if (!s_challenge_mode && EmuConfig.Achievements.ChallengeMode)
 		{
 			if (HasActiveGame())
-				ImGuiFullscreen::ShowToast(std::string(), "Hardcore mode will be enabled on system reset.", 10.0f);
+				ImGuiFullscreen::ShowToast(std::string(), "Hardcore mode will be enabled on system reset.", 5.0f);
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


Mrlinkwii (approved):

stenzek:
-1 from me. 5 seconds is not long enough to read some of the longer messages, especially for non native English speakers. That's why the delay can be set per message, so the shorter/less important messages don't hang around as long.

If they're so bothering, you can always turn them off.

RedDevilus:
Extra seconds won't make people read it, we used to have 2 seconds on WX. 5-10 seconds is fine depending on the length of messages.

refractionpcsx2:
The longer ones and errors should have a good amount of time to give people time to screenshot (including reading it and fumbling for their keyboard if they're using a controller), so the notifications should be 10-15 seconds really, at least

Andryshik345:
I can agree that you can just turn them off completely, but in case, for example savestates, they are useful to me to see which slot is currently selected and and if state was even saved or not.

PR #7396, opened 2022-11-16
GameDB: Fixes for Finding Nemo, Haven: Call of the King, Kinetica

icup321:
### Description of Changes
Adds HalfPixelOffset and AlignSprite to Finding Nemo and Haven: Call of the King. Also removes a non-existent NTSC-U Haven serial number. Adds AlignSprite to Kinetica as well, uses God of War's engine.

### Rationale behind Changes
Less ghosting and vertical lines when upscaling, and less tinkering for users.

### Suggested Testing Steps
Test said games for ghosting or vertical lines.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -7083,7 +7083,6 @@ SCUS-97160:
 SCUS-97161:
   name: "Kinetica [Demo]"
   region: "NTSC-U"
-  compat: 5
   gsHWFixes:
     roundSprite: 1 # Fixes vertical lines and minor ghosting.
 SCUS-97163:
@@ -9024,7 +9023,6 @@ SLED-51225:
 SLED-51281:
   name: "Haven - Call of the King [Demo]"
   region: "PAL-E"
-  compat: 5
   gsHWFixes:
     halfPixelOffset: 2 # Fixes ghosting.
     alignSprite: 1 # Fixes vertical lines.
@@ -12942,7 +12940,6 @@ SLES-51754:
 SLES-51755:
   name: "Disney-Pixar's Finding Nemo"
   region: "PAL-E"
-  compat: 5
   gsHWFixes:
     halfPixelOffset: 2 # Fixes ghosting.
     alignSprite: 1 # Fixes vertical lines.
@@ -13228,7 +13225,6 @@ SLES-51869:
 SLES-51870:
   name: "Disney-Pixar's Finding Nemo"
   region: "PAL-FI-S"
-  compat: 5
   gsHWFixes:
     halfPixelOffset: 2 # Fixes ghosting.
     alignSprite: 1 # Fixes vertical lines.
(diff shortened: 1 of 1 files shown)


Mrlinkwii:
can you post any gs dumps? that can be checked

bin/resources/GameIndex.yaml
```diff
@@ -39938,9 +39971,6 @@ SLUS-20153:
 SLUS-20154:
   name: "NFL Quarterback Club 2002"
   region: "NTSC-U"
-SLUS-20157:
```
lightningterror, line 39941: Are you sure about this?

icup321: Yes, it's a typo of SLUS-20517, and SLUS-20157 is not used for any other PS2 game afaik.

icup321:
Here's a GS dump for Finding Nemo:
Before fixes:
Disney's Finding Nemo_SLUS-20628_20221116154107
[Disney's Finding Nemo_SLUS-20628_20221116154107.zip](https://github.com/PCSX2/pcsx2/files/10025589/Disney.s.Finding.Nemo_SLUS-20628_20221116154107.zip)
After fixes:
Disney's Finding Nemo_SLUS-20628_20221116153907
[Disney's Finding Nemo_SLUS-20628_20221116153907.zip](https://github.com/PCSX2/pcsx2/files/10025597/Disney.s.Finding.Nemo_SLUS-20628_20221116153907.zip)

JordanTheToaster:
The fixes for finding nemo look good could you remove the compat rating you added.

icup321:
> The fixes for finding nemo look good could you remove the compat rating you added.

Why? That rating was already there for one of the regional versions, I just simply added it to the rest of them for consistency. Plus, it has that same rating listed here: https://wiki.pcsx2.net/Disney/Pixar_Finding_Nemo

RedDevilus:
The rating in the GameDB follows the one on the forum thread here:
https://forums.pcsx2.net/Forum-Public-compatibility-list

The wiki isn't exactly the main source. Besides we need a better reporting system and not just adding some random compatibility ratings even if you are correct. I just don't want it to be like a free for all.

refractionpcsx2:
also you can't assume because one region is playable that all others are too, there has been many situations where an NTSC game was fine but a PAL version was broken.

RedDevilus:
I'm against trusting wiki as gospel or as main source when I add GameDB entries, it needs verification but it's good for reference atleast.

RedDevilus:
Besides, you also added compatability to playable for the demo and other games that are not reported on the pcsx2 compat rating. Not saying it's your fault but compat of wiki is invalid.

icup321:
Fine, but in that case should I apply this to the other games I edited? Should I just remove all of their ratings in that case since they aren't properly "verified" or should I change them to match their forum ratings? (Which only Finding Nemo has a thread)

RedDevilus:
https://pcsx2.net/compat/ you can put the specific serial/game in here and if you don't see it, don't add them basically. And you can't use other serials as baseline

```
SLES-51871:
  name: "Disney-Pixar's Finding Nemo"
  region: "PAL-F-G"
  compat: 5
```
Is already correct, so you can't add it to other Nemo serials

Mrlinkwii:
can you post a gs dump for Haven - Call of the King?

JordanTheToaster (approved):

PR #7419, opened 2022-11-20
GS-TC: Avoid large clears in HW + old frames

### LINKED ISSUE #7371: [BUG]: Fatal Frame 1 camera flash issues
### Describe the Bug

The camera flash will not render a full fade in fade out effect instead will flicker like its flashing twice, stark interlacing lines may appear over the resulting photo,

https://user-images.githubusercontent.com/22896793/201156372-cde458df-1b83-4307-b114-e80c369e7fa2.mp4

### Reproduction Steps

Use any other interlacing method then the current faked blend which will cover the issue up

### Expected Behavior

One solid fade in fade out flash effect

### PCSX2 Revision

Any

### Operating System

### If Linux - Specify Distro

Arch

### CPU

Ryzen 3700x

### GPU

GTX 2070

### GS Settings

Defaults

### Emulation Settings

Defaults

### GS Window Screenshots

_No response_

### Logs & Dumps

_No response_

refractionpcsx2:
### Description of Changes
Adds changes to avoid doing very large clears in hardware, creating huge textures which causes a lot of misdetection.  Also stops incrementing age on frames as they are not being drawn/modified, but also try to avoid using old frames unless nothing better is available.

### Rationale behind Changes
Fatal Frame and Clock Tower 3 do absolutely huge draws just before FMV's to clear out the memory so the surround for the FMV is black, but this causes a really big texture being detected, when it shouldn't exist (since it's a clear).  The other issue was it was keeping an old frame about instead of offsetting inside a large double buffer used for the FMV's, resulting in flashing.

This also affects the camera flickering problem in Fatal frame due to similar issues.

### Suggested Testing Steps
Test games which need a "safe features" enabled (like powerdrome, I think xenosaga, maybe others) to make sure they still clear correctly, but also check loading screens and FMV's for flickering.

Fixes Clock Tower 3 flashing FMV's issue
Fixes #7371 Fatal Frame flickering when the camera flash goes off.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
@@ -1690,6 +1690,40 @@ void GSRendererHW::Draw()
 
 	// The rectangle of the draw
 	m_r = GSVector4i(m_vt.m_min.p.xyxy(m_vt.m_max.p)).rintersect(GSVector4i(context->scissor.in));
+
+
+	if (!GSConfig.UserHacks_DisableSafeFeatures)
+	{
+		// Constant Direct Write without texture/test/blending (aka a GS mem clear)
+		if ((m_vt.m_primclass == GS_SPRITE_CLASS) && !PRIM->TME // Direct write
+			&& (!PRIM->ABE || m_context->ALPHA.IsOpaque()) // No transparency
+			&& (m_context->FRAME.FBMSK == 0) // no color mask
+			&& !m_context->TEST.ATE // no alpha test
+			&& (!m_context->TEST.ZTE || m_context->TEST.ZTST == ZTST_ALWAYS) // no depth test
+			&& (m_vt.m_eq.rgba == 0xFFFF) // constant color write
+			&& m_r.x == 0 && m_r.y == 0) { // Likely full buffer write
+
+			if (OI_GsMemClear() && m_r.w > 1024)
+			{
+				if ((fm & fm_mask) != fm_mask)
+				{
+
+					m_tc->InvalidateVideoMem(context->offset.fb, m_r, true);
+
+					m_tc->InvalidateVideoMemType(GSTextureCache::RenderTarget, context->FRAME.Block());
+				}
+
+				if (zm != 0xffffffff)
+				{
+					m_tc->InvalidateVideoMem(context->offset.zb, m_r, false);
+
+					m_tc->InvalidateVideoMemType(GSTextureCache::DepthStencil, context->ZBUF.Block());
+				}
+				return;
+			}
+		}
+	}
+
 	GSVector2i unscaled_size;
 	const GSVector2i t_size = GetTargetSize(&unscaled_size);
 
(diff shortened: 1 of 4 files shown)


pcsx2/GS/Renderers/HW/GSRendererHW.cpp
```diff
@@ -1685,6 +1685,40 @@ void GSRendererHW::Draw()
...
+			&& !m_context->TEST.ATE // no alpha test
+			&& (!m_context->TEST.ZTE || m_context->TEST.ZTST == ZTST_ALWAYS) // no depth test
+			&& (m_vt.m_eq.rgba == 0xFFFF) // constant color write
+			&& m_r.x == 0 && m_r.y == 0) { // Likely full buffer write
+
+			if (OI_GsMemClear() && m_r.w > 1024)
+			{
+				if ((fm & fm_mask) != fm_mask)
+				{
+
```
lightningterror, line 1705: Extra row.

pcsx2/GS/Renderers/HW/GSRendererHW.cpp
```diff
@@ -1685,6 +1685,40 @@ void GSRendererHW::Draw()
 
 	// The rectangle of the draw
 	m_r = GSVector4i(m_vt.m_min.p.xyxy(m_vt.m_max.p)).rintersect(GSVector4i(context->scissor.in));
+
```
lightningterror, line 1688: Maybe this one too.

refractionpcsx2: That's supposed to be there.

pcsx2/GS/Renderers/HW/GSRendererHW.cpp
```diff
@@ -1690,6 +1690,38 @@ void GSRendererHW::Draw()
...
+	if (!GSConfig.UserHacks_DisableSafeFeatures)
+	{
+		// Constant Direct Write without texture/test/blending (aka a GS mem clear)
+		if ((m_vt.m_primclass == GS_SPRITE_CLASS) && !PRIM->TME // Direct write
+			&& (!PRIM->ABE || m_context->ALPHA.IsOpaque()) // No transparency
+			&& (m_context->FRAME.FBMSK == 0) // no color mask
+			&& !m_context->TEST.ATE // no alpha test
+			&& (!m_context->TEST.ZTE || m_context->TEST.ZTST == ZTST_ALWAYS) // no depth test
+			&& (m_vt.m_eq.rgba == 0xFFFF) // constant color write
+			&& m_r.x == 0 && m_r.y == 0) { // Likely full buffer write
```
lightningterror, line 1703: { bracket should be on new line.

refractionpcsx2: I literally copy/pasted existing code. But I guess I can change it.

lightningterror (approved):

dezraj:
Hello refraction! I don't know if it's related to this fix, but there's a glitch in Dog's Life in which it shows a black bar when it triggers the paw FMV in the stage loading animation, as you see here:
https://drive.google.com/file/d/1I3v1WcddnrujbjAFOKpE6xCZrAGGHKGa/view?usp=sharing

It happens in native resolution, but not in the software renderer. This PR doesn't seem to fix it unfortunately.

Here's a save-state file if you'd like to investigate it: 
[SCES-51248 (531061F2).01.zip](https://github.com/PCSX2/pcsx2/files/10064412/SCES-51248.531061F2.01.zip)

PS: interestingly it doesn't happen the first time you enter the stage, just when changing from one stage/area to another.

refractionpcsx2:
Has that always been there? Or do you know if it happened from a certain point? would be interesting to know!

dezraj:
> Has that always been there? Or do you know if it happened from a certain point? would be interesting to know!

I wouldn't be able to tell precisely, but I think it has been there for quite a while or at least since I started playing the game, which was some months ago.

Thank you for your reply.

PR #7440, opened 2022-11-22
Qt: Change dump compression to Zstandard

JordanTheToaster:
### Description of Changes
Until XZ stops exploding we should default to Zstandard for dump compression.

### Rationale behind Changes
Explosions when trying to get a dump for support is inconvenient. 

### Suggested Testing Steps
Make sure CI is happy.


RedDevilus:
Characterisation of a failed GS dump (with gs.xz) has 1.00 MB filesize. Not sure why it is breaking these last few months but before was mostly fine. Needs research later. -- Note for later.

PR #7441, opened 2022-11-22
BiosTools: improve region codes

AKuHAK:
Region code numbering now matches regions inside CDVD_internal.h Added region assignment for non-standard romver
Renamed HK region to Asia as 'H' covers not only Hong Kong, but also Korea and Taiwan


refractionpcsx2:
I'm sure you need to change the list in 
void BIOSSettingsWidget::listRefreshed(const QVector<BIOSInfo>& items)
which you can find in BIOSSettingsWidget.cpp inside the Qt project.

refractionpcsx2:
Also that does not match the list in 
static u8 biosLangDefaults[8][16] =
which is in CDVD_internal.h as you pointed out.

AKuHAK:
Thanks. Didnt realize that there are so many places that are relying on region order. Ill recheck the code.

refractionpcsx2:
yeah there's a few places :) Thanks

refractionpcsx2:
I'm confused. In some places you've renamed Hong Kong to Asia, and in other places you're calling it Oceania, which isn't even Hong Kong.

Can you check those please. What source do you have for these regions?

AKuHAK:
@refractionpcsx2 I used this list for reference:
https://github.com/PCSX2/pcsx2/blob/b30e93389cc8d145fa7a31144287cfa12f6d73a6/pcsx2/CDVD/CDVD_internal.h#L176
This is a list of mechacon regions, it's hard coded and used by all ps2.
Other region lists are used by the pcsx2 team internally and aren't strictly assigned. This PR syncs mechacon numbering and internal region numbering.

Hong kong was renamed to Asia and everywhere was assigned as region = 4. Oceania is region = 3, but Oceania cannot be detected by Romver check cause Oceania and Europe share the same bios.
Currently, region = 3 (Oceania), region = 5 (Russia), and region = 7 (Mexico) cannot be detected by the Bios version as they share the same bios as other regions (only rom1 is different for them).

I also changed the order - so it is sequential everywhere. I added Oceania handling for future PRs if someone will plan to implement the rom1 check (Oceania and Europe have different rom1 versions).

AKuHAK:
```c
				case 'E': zone = "Europe"; region = 2;  break;
				case 'H': zone = "Asia";   region = 4;  break;
				case 'C': zone = "China";  region = 6;  break;
				case 'T': zone = "T10K";   region = 8;  break;
```
should actually be
```c
				case 'E': zone = "Europe"; region = 2;  break;
				// not implemented // case 'E': zone = "Oceania"; region = 3;  break;
				case 'H': zone = "Asia";   region = 4;  break;
				// not implemented // case 'E': zone = "Russia"; region = 5;  break;
				case 'C': zone = "China";  region = 6;  break;
				// not implemented // case 'A': zone = "Mexico"; region = 7;  break;
				case 'T': zone = "T10K";   region = 8;  break;
```

refractionpcsx2:
Thanks

PR #7444, opened 2022-11-22
GS: Add sharp bilinear for lower resolutions.

### LINKED ISSUE #7183: [Feature Request]: Sharp bilinear interpolation
### Description

Described and shown in detail here, along with some code https://github.com/dolphin-emu/dolphin/pull/9860

Output native resolution > integer scale each axis nearest to (but not smaller than) target resolution (this includes accounting for AR changes) > finally scale down to exact resolution and interpolate image with filtering

### Reason

Results in a lot less blur than immediately filtering the image but also corrects pixel positions/sizes at any resolution or AR.

Essentially this is an alternative for people who like a sharp native res image without artifacts from nearest with uneven scaling, but might not like integer scaling or weird aspect ratios.

"Why not just use nearest and integer scaling?" 
Nearest neighbor can cause squished or stretched pixels across the image when not scaled evenly and is very noticeable in motion. Integer scaling often doesn't fit an image to the display height and requires square pixels to achieve correct scaling, which means non standard AR's for games that use very odd resolutions or multiple resolutions.

Nearest can sorta be "brute forced" by using a high output resolution to have more pixels to sample from (4k displays) but most people are still on 1080p or 1440p displays.

"Why not use bilinear filtering?" 
Causes excessive blurring that only gets worse the lower resolution the source image is.

### Examples

Shown in the linked Dolphin PR and below.

refractionpcsx2:
### Description of Changes
Modifies bilinear filtering to have a sharper appearance at lower resolutions. if the resolution is lower than the window, it will be integer scaled up then bilinear filtered to fit the window.

### Rationale behind Changes
Adds upscaling of the picture past the window size by an integer scale, so downsampling back to the window size with bilinear doesn't destroy information or excessively blur the picture while maintaining pixel uniformity.

### Suggested Testing Steps
test stuff in native, compare to master with different window sizes. compare to nearest filtering also (Bilinear Filter on the Display tab of the graphics settings disabled)

Closes #7183

### DIFF
--- a/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
@@ -104,13 +104,12 @@ GraphicsSettingsWidget::GraphicsSettingsWidget(SettingsDialog* dialog, QWidget* 
 	SettingWidgetBinder::BindWidgetToEnumSetting(sif, m_ui.fmvAspectRatio, "EmuCore/GS", "FMVAspectRatioSwitch",
 		Pcsx2Config::GSOptions::FMVAspectRatioSwitchNames, FMVAspectRatioSwitchType::Off);
 	SettingWidgetBinder::BindWidgetToIntSetting(sif, m_ui.interlacing, "EmuCore/GS", "deinterlace_mode", DEFAULT_INTERLACE_MODE);
-	SettingWidgetBinder::BindWidgetToBoolSetting(sif, m_ui.bilinearFiltering, "EmuCore/GS", "linear_present", true);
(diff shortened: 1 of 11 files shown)


refractionpcsx2:
Here's an example, where a picture (look at the left of the black ring) is distorted in "nearest" scaling to the window size, and bilinear overblurs.

Nearest:
image

Original Bilinear (smooth):
image

New Bilinear (sharp):
image

RedDevilus:
https://imgsli.com/MTM1NDg4

If you want to easily compare, click on the options itself to swap out which image you want to compare.

Mrlinkwii (approved):

PR #7464, opened 2022-11-25
Formatting: return editorconfig

AKuHAK:
reverts #3995

### Description of Changes
Provide `Editor-config` configuration.

### Rationale behind Changes
`Editorconfig` covers more than `clang-format` and is safer. Clang-format only supports C style files (like .c, .cpp, .h, .hpp), while Editorconfig supports any text file.
When using macros like
```
// clang-format off
...
// clang-format on
```
code between these 2 macroses will be completely unchecked, but editor-config still can make simple code cleaning.
Editor-config is safer and doesn't require checking if the code is still in good condition as it is less sophisticated.
All that it does:
- remove trailing tabs and spaces
- converts line endings into one format
- inserts a final newline
- remove non-utf-8 characters

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,30 @@
+# EditorConfig: http://EditorConfig.org
+
+# Top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+
+# Tab indentation
+[*.{cpp,h,js,cmake,mm,inl,in}]
+indent_style = tab
+
+# 1 space
+[*.{ui}]
+indent_style = space
+indent_size = 1
+
+# 2 space
+[*.{js,xml,c,py,asm,cmd,nsi,nsh,tex}]
+indent_style = space
+indent_size = 2
+
+# 4 space
+[*.{sh,pl}]
+indent_style = space
+indent_size = 4


stenzek:
... why?

Any C++ editor worth its salt these days supports clang-format.

AKuHAK:
> Any C++ editor worth its salt these days supports clang-format.

- clang-format does not perform checks on common text files like *.cmake, vscode, *.md files, etc
- when using macro `// clang -format off` (that is pretty common here), part of code becomes unchecked. Editor-config covers those cases

stenzek:
Except we actually want the full formatting from clang-format for consistency.

AKuHAK:
> Except we actually want the full formatting from clang-format for consistency.

Editor-config is used in places where clang-formatting cannot be applied. in all other cases, clang-formatting is still a preferrable solution, but clang-format doesn't cover the whole codespace.

AKuHAK:
This type of linter is very useful when you edit plain text files. On saving it will remove trailing spaces, converts line ending to the one format and remove any accidental non-utf8 characters. Clang-format cannot be used with all type of files, it only covers files that contains some code, but this repository doesnt have only code files but also many other files.
Some files extension can be linted with other linters (.xml file has one linter, *.cmake files has another linter, 8.yml file has 3rd linter), but in case if you dont have set up the linter it will fall back into at least editorconfig. This is something like failsafe linter.
And as I said it is useful for codeblocks where clang-format is disabled by macro `// clang-fromat off`

AKuHAK:
Of course if pcsx2 team thinks that the advantages that I list do not exceed the disadvantage (that it will enlarge cluttering), feel free to close this pr.

refractionpcsx2:
Nope, Clang is fine, not had any problems with that nor have I come across any of the problems you have describe to warrant it.

PR #7474, opened 2022-11-26
GS-TC: Make sure if we have an old frame check bounds of other frame

refractionpcsx2:
### Description of Changes
Makes sure if an old frame has already been perfectly matched and it finds another target that it is inside, make sure either end matches the pointer to properly identify a double buffer.

### Rationale behind Changes
Shadow Hearts 2 was mis detecting false targets between scenes.

### Suggested Testing Steps
test stuff, make sure no new flickering happens.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
+++ b/pcsx2/GS/Renderers/HW/GSTextureCache.cpp
@@ -529,15 +529,15 @@ GSTextureCache::Target* GSTextureCache::LookupTarget(const GIFRegTEX0& TEX0, con
 		// 2nd try ! Try to find a frame at the requested bp -> bp + size is inside of (or equal to)
 		if (!dst)
 		{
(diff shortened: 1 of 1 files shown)


pcsx2/GS/Renderers/HW/GSTextureCache.cpp
```diff
@@ -529,15 +529,15 @@ GSTextureCache::Target* GSTextureCache::LookupTarget(const GIFRegTEX0& TEX0, con
 		// 2nd try ! Try to find a frame at the requested bp -> bp + size is inside of (or equal to)
 		if (!dst)
 		{
+			u32 needed_end = GSLocalMemory::m_psm[TEX0.PSM].info.bn(real_w - 1, real_h - 1, bp, TEX0.TBW);
```
lightningterror, line 532: Can be const.

PR #7481, opened 2022-11-26
RedtapeWindows: Lowercase includes for linux fs

xantares:
Fix build on mingw from linux

### DIFF
--- a/common/RedtapeWindows.h
+++ b/common/RedtapeWindows.h
@@ -28,8 +28,8 @@
 #define _WIN32_WINNT 0x0A00 // Windows 10
 
 #include <windows.h>
-#include <VersionHelpers.h>
-#include <ShTypes.h>
+#include <versionhelpers.h>
+#include <shtypes.h>
 #include <timeapi.h>
 #include <tchar.h>
 


stenzek:
I'll have to check the windows SDK tomorrow but if I remember correctly these files are cased correctly. So this PR may be wrong.

lightningterror:
> I'll have to check the windows SDK tomorrow but if I remember correctly these files are cased correctly. So this PR may be wrong.

Looking at the sdk it's `shtypes.h` and `VersionHelpers.h` (version 10.0.20348.0)

xantares:
that's on windows, where case doesnt matter

stenzek:
So? Lower casing it is wrong.

stenzek:
Just going to make the point that we're not interested in compiling with mingw, because there's no advantages, compared to the cons:

 - the headers are out of date, you'll probably have issues with e.g. fastmem
 - libraries are missing (known issue)
 - qt may or may not build with it, but regardless we don't have prebuilt binaries for mingw, only msvc
 - our CI runs on windows , as do all our developer machines
 - it's not something any of us have set up, so we can't fix any issues
 - the CI doesnt have it, and spending however long every run to install it is silly, so that means we'll never know when it breaks
 - clang-cl likely generates better code, and we're moving to that eventually 

You're welcome to maintain it in a fork, but I don't think we should have incorrect code, for the sake of something that none of us use.

xantares:
I'm not expecting you to maintain a mingw port or use it at all, just pointing out some valid changes that makes it compatible, see how the CI stays green ?

refractionpcsx2:
It compiles fine, checked it locally, so sure, whatever I guess.

PR #7483, opened 2022-11-26
GS-DX11: Fix D3D11_BOX parameters (missing "front")

refractionpcsx2:
### Description of Changes
Fixes D3D11 box parameters for textures.

### Rationale behind Changes
mistake from #7477 which apparently the compiler lets you get away with.

### Suggested Testing Steps
test games on DX11, noticed on burnout.


stenzek (approved):
Code lgtm, thanks for fixing my dumbass mistake

PR #7484, opened 2022-11-26
Misc: Another batch of warnings, mostly codacy.

lightningterror:
### Description of Changes

Misc: Another batch of warnings, mostly codacy.
### Rationale behind Changes

Less warnings.
### Suggested Testing Steps

CI builds, warnings are solved within codacy/ci.

### DIFF
--- a/pcsx2/GS/Renderers/SW/GSRendererSW.h
+++ b/pcsx2/GS/Renderers/SW/GSRendererSW.h
@@ -75,7 +75,7 @@ protected:
 	void Draw() override;
 	void Queue(GSRingHeap::SharedPtr<GSRasterizerData>& item);
 	void Sync(int reason);
-	void ExpandTarget(const GIFRegBITBLTBUF& BITBLTBUF, const GSVector4i& r);
+	void ExpandTarget(const GIFRegBITBLTBUF& BITBLTBUF, const GSVector4i& r) override;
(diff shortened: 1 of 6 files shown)


pcsx2/x86/ix86-32/iR5900-32.cpp
```diff
@@ -575,7 +575,6 @@ static void recAlloc()
 	recROM1 = basepos;
 	basepos += (Ps2MemSize::Rom1 / 4);
 	recROM2 = basepos;
-	basepos += (Ps2MemSize::Rom2 / 4);
```
refractionpcsx2, line 578: You might wanna check this with Tellow, I'm sure we had an arguement on if it should be there or not, I think he got me to put it back in because then you can see if the pointer is pointing to the right place (end of the data), but I forget.

tellowkrinkle: Yeah, not a fan

You could make a `BumpAllocator` struct that holds a pointer with a single method `allocate` that advances the pointer and returns the pre-advance value.  Then codacy won't care that it's modified without use at the end, because it doesn't care as long as you stick enough indirection in there

RedDevilus:
I still can't tell if codacy can't see longer than it's nose, being too paranoid or just too nitpicky for all it's 'supposed' issues.

PR #7492, opened 2022-11-27
input-rec: remove input recording code from wx-related source files

xTVaser:
### Description of Changes

Removes input recording from the wx related source files.  No longer have to worry about maintaining two working versions of the same feature and is one less thing to cleanup when wx is fully deprecated.

The only thing that remains is a deprecation notice on "System > Game Settings > Enable Input Recording"

### Suggested Testing Steps

- does wx still function otherwise
- no unintentional regressions on qt


refractionpcsx2:
`The only thing that remains is a deprecation notice on "System > Game Settings > Enable Input Recording"`

why not just remove it?

xTVaser:
To be proactive about informing users that are still using wxWidgets

refractionpcsx2:
You're making the assumption people read :P

Remember that time I put a message up saying "You are going to be booted in to the bios to configure your language, this will only happen once" and we were flooded with people asking "WHY WON'T MY GAME BOOT, IT SENDS ME TO THE BIOS EVERY TIME!" because they didn't read it, assumed it fucked up and tried again :P

RedDevilus:
I think they had enough time reading for 2 years about a more niche feature :P 
It was going to be a painful procedure no matter how we did it.

Mrlinkwii (approved):

PR #7494, opened 2022-11-28
Use semaphores with a userspace fast path

tellowkrinkle:
### Description of Changes
Adds a semaphore with a userspace fast path and uses it
Improves performance in GS thread limited applications, when MTVU is enabled.  Amount depends on operating system (how slow kernel semaphores were).
In Sly at 1x:
- macOS: +30%
- Windows: +20%
- Linux: +1%

Also fixes high EE thread usage during gsdump playback

### Rationale behind Changes
Faster yay

### Suggested Testing Steps
Test your GS thread limited games (I recommend sly)

### DIFF
--- a/common/Darwin/DarwinSemaphore.cpp
+++ b/common/Darwin/DarwinSemaphore.cpp
@@ -42,16 +42,10 @@
 
 static void MACH_CHECK(kern_return_t mach_retval)
 {
-	switch (mach_retval)
+	if (mach_retval != KERN_SUCCESS)
 	{
-		case KERN_SUCCESS:
-			break;
-		case KERN_ABORTED: // Awoken due reason unrelated to semaphore (e.g. pthread_cancel)
-			pthread_testcancel(); // Unlike sem_wait, mach semaphore ops aren't cancellation points
-			// fallthrough
-		default:
-			fprintf(stderr, "mach error: %s", mach_error_string(mach_retval));
-			assert(mach_retval == KERN_SUCCESS);
+		fprintf(stderr, "mach error: %s", mach_error_string(mach_retval));
+		assert(mach_retval == KERN_SUCCESS);
 	}
 }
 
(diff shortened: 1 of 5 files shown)


Mrlinkwii (approved):

refractionpcsx2:
Naturally did the "Ratchet Test" on my 5900X, aaaand we're back to EE limited :D only just though, they kinda trade places. But it's a nice speed boost!

image

refractionpcsx2:
And NFS carbon, not a bad little boost either :)

image

JordanTheToaster:
DId some testing with DX11 from 1.4 to 1.6 to this PR in Burnout 3.

jkrtkrtmrtfkm

HighBunkerHxH:
Have you tried Tekken Tag Tournament, especially on the Ogre Stage?

refractionpcsx2:
> Have you tried Tekken Tag Tournament, especially on the Ogre Stage?

Sorry, they're fundamentally the same (probably a driver limitation), tested with VUlkan Native x3

image

HighBunkerHxH:
> > Have you tried Tekken Tag Tournament, especially on the Ogre Stage?
> 
> Sorry, they're fundamentally the same (probably a driver limitation), tested with VUlkan Native x3
> 
> image

on DX11 it runs somehow faster

refractionpcsx2:
yes, probably because driver overhead is lower (in some cases)

CookiePLMonster:
If C++20 can be used, then `std::atomic::wait` can be used: https://en.cppreference.com/w/cpp/atomic/atomic/wait

It's implemented in user mode at least on Windows, as per Old New Thing: https://devblogs.microsoft.com/oldnewthing/20160823-00/?p=94145
> Like critical sections, Wait­On­Address does most of its work in user mode, so if you’re lucky, you can avoid kernel transitions.

PR #7497, opened 2022-11-28
Misc: And another batch of warnings fixup.

lightningterror:
### Description of Changes

Misc: And another batch of warnings fixup.
### Rationale behind Changes

Less warnings.
### Suggested Testing Steps

Compiles.

### DIFF
--- a/pcsx2/MemoryCardFile.cpp
+++ b/pcsx2/MemoryCardFile.cpp
@@ -371,7 +371,11 @@ void FileMemoryCard::Open()
 			m_chkaddr = 0x210;
 
 			if (!m_ispsx[slot] && FileSystem::FSeek64(m_file[slot], m_chkaddr, SEEK_SET) == 0)
-				std::fread(&m_chksum[slot], sizeof(m_chksum[slot]), 1, m_file[slot]);
+			{
+				const size_t read_result = std::fread(&m_chksum[slot], sizeof(m_chksum[slot]), 1, m_file[slot]);
+				if (read_result == 0)
+					Host::ReportFormattedErrorAsync("Memory Card", "Error reading memcard.\n");
+			}
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


pcsx2/MemoryCardFile.cpp
```diff
@@ -371,7 +371,11 @@ void FileMemoryCard::Open()
 			m_chkaddr = 0x210;
 
 			if (!m_ispsx[slot] && FileSystem::FSeek64(m_file[slot], m_chkaddr, SEEK_SET) == 0)
-				std::fread(&m_chksum[slot], sizeof(m_chksum[slot]), 1, m_file[slot]);
+			{
+				const size_t read_result = std::fread(&m_chksum[slot], sizeof(m_chksum[slot]), 1, m_file[slot]);
+				if (read_result == 0)
+					Console.Error("(FileMcd) error reading memcard.");
```
F0bes, line 377: Maybe it's advisable to use Host::ReportFormattedErrorAsync() so the failure to read doesn't go unnoticed by the user?

lightningterror: @RedPanda4552 thoughts?

RedPanda4552: Agreed, I would use the function referenced by fobes (same one as just up in the main if block).

pcsx2/MemoryCardFile.cpp
```diff
@@ -503,8 +507,10 @@ s32 FileMemoryCard::Save(uint slot, const u8* src, u32 adr, int size)
 		if (!Seek(mcfp, adr))
 			return 0;
 		m_currentdata.MakeRoomFor(size);
-		std::fread(m_currentdata.GetPtr(), size, 1, mcfp);
 
+		const size_t read_result = std::fread(m_currentdata.GetPtr(), size, 1, mcfp);
+		if (read_result == 0)
+			Console.Error("(FileMcd) error reading memcard.");
```
RedPanda4552, line 513: Ditto

PR #7512, opened 2022-11-30
Qt: Allow to run bilinear filtering when integer scaling is enabled.

### LINKED ISSUE #7180: [BUG]: Integer upscaling disables bilinear filtering, but bilinear filtering can still be used.
### Describe the Bug

Whenever you enable "Integer Upscaling" in the "Graphics" options it will grey out the "Bilinear Filtering" option, however it doesn't turn off BF itself it just greys it out.

Also I would just recommend not greying it out in general because I actually like using IU + BF and I've had no issues with it, I know DuckStation does it for whatever reason as well but I just don't see a point in it.

### Reproduction Steps

- Disable IU (If enabled)
- Toggle BF
- Enable IU

### Expected Behavior

Since DuckStation does the same I assume that it would be trying to disable the option but I would personally just recommend making it so it doesn't grey out the option because like I said above it doesn't cause issues and allows for more accuracy for those who prefer that.

### PCSX2 Revision

v1.7.3411

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

### LINKED ISSUE #7622: [BUG]: Qt wrongly disabling Bilinear Filtering in per-game settings (unset integer scaling)
### Describe the Bug

While it works fine in global settings:
image

It doesn't in per-game settings:
image

### Reproduction Steps

- Go into per-game settings of any game
- Go to graphics settings - rendering tab
- Try to change Bilinear Filtering

### Expected Behavior

It does work correctly if you manually unset global option for integer scaling.

image

### PCSX2 Revision

1.7.3737

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

lightningterror:
### Description of Changes

Qt: Allow to run bilinear filtering when integer scaling is enabled.
### Rationale behind Changes

Fixes #7180
Fixes #7622
User request
### Suggested Testing Steps

Tested already.

### DIFF
--- a/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
+++ b/pcsx2-qt/Settings/GraphicsSettingsWidget.cpp
@@ -127,9 +127,6 @@ GraphicsSettingsWidget::GraphicsSettingsWidget(SettingsDialog* dialog, QWidget* 
 	dialog->registerWidgetHelp(m_ui.noInterlacingPatches, tr("Enable No-Interlacing Patches"), tr("Unchecked"),
 		tr("Automatically loads and applies no-interlacing patches on game start. Can cause issues."));
 
-	connect(m_ui.integerScaling, &QCheckBox::stateChanged, this, &GraphicsSettingsWidget::onIntegerScalingChanged);
-	onIntegerScalingChanged();
-
 	connect(m_ui.fullscreenModes, QOverload<int>::of(&QComboBox::currentIndexChanged), this, &GraphicsSettingsWidget::onFullscreenModeChanged);
 
 	//////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


dmn1n:
There are some instances where horizontal bilinear scaling may be required. For instance at 2x resolution some of the PS2 games run at 1024x896 resolution and 1280x896. What would be the behavior when bilinear filtering is disabled after enabling integer scaling ? is there an option to enable bilinear filtering for horizontal scaling ?

KrossX:
I think that would be incompatible with what integer scaling is trying to do. If you want sharp pixels and still keep the aspect ratio, then I think that's when you would use the new sharp scaling instead ain't it.

dmn1n:
@KrossX I agree, sharp bilinear should scale to any resolution. However I have a display that supports 1600x900@60 hz mode. So it would be better option to do integer scaling for vertical while still allowing bilinear horizontal scaling for widescreen.

lightningterror:
If people want I can remove the restriction from the gui instead and allow bilinear in combination with integer scaling.

ps1freak26:
I'd prefer if integer scaling didn't block the bilinear filtering section if enabled globally and having to disable integer scaling to use bilinear filtering in per game configs even if integer scaling is disabled globally.

RedDevilus (approved):

PR #7524, opened 2022-12-03
GHActions: Update release pipeline for MultiISA

TellowKrinkle:
### Description of Changes
We have multi-isa builds, now we just have to publish them

Do not merge until #7532 has been merged for long enough that most users running auto update have it

### Suggested Testing Steps
Test the release pipeline (@xTVaser)

### DIFF
--- a/.github/workflows/scripts/common/name-artifacts.sh
+++ b/.github/workflows/scripts/common/name-artifacts.sh
@@ -21,7 +21,7 @@ NAME=""
 
 if [ "${OS}" == "macos" ]; then
   NAME="PCSX2-${OS}-${GUI_FRAMEWORK}"
-elif [[ ("${OS}" == "windows" && "$BUILD_SYSTEM" != "cmake") || ("$OS" == "linux" && "$GUI_FRAMEWORK" == "QT") ]]; then
+elif [[ ("${OS}" == "windows" && "$BUILD_SYSTEM" != "cmake") ]]; then
   NAME="PCSX2-${OS}-${GUI_FRAMEWORK}-${ARCH}-${SIMD}"
 else
   NAME="PCSX2-${OS}-${GUI_FRAMEWORK}-${ARCH}"
--- a/.github/workflows/linux_build_matrix.yml
+++ b/.github/workflows/linux_build_matrix.yml
@@ -9,25 +9,12 @@ on:
       - master
 
 jobs:
-  build_linux_qt_sse4:
+  build_linux_qt:
     name: "AppImage"
     uses: ./.github/workflows/linux_build_qt.yml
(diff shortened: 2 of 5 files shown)


.github/workflows/scripts/common/name-artifacts.sh
```diff
@@ -21,7 +21,7 @@ NAME=""
 
 if [ "${OS}" == "macos" ]; then
   NAME="PCSX2-${OS}-${GUI_FRAMEWORK}"
-elif [[ ("${OS}" == "windows" && "$BUILD_SYSTEM" != "cmake") || ("$OS" == "linux" && "$GUI_FRAMEWORK" == "QT") ]]; then
+elif [[ ("${OS}" == "windows" && "$BUILD_SYSTEM" != "cmake") ]]; then
   NAME="PCSX2-${OS}-${GUI_FRAMEWORK}-${ARCH}-${SIMD}"
```
xTVaser, line 25: Does the change no include windows?  It seems like there are still traces of explicitly defining `simd` in a few places:
https://github.com/TellowKrinkle/pcsx2/blob/c357e64dea294952185e58f24a9b2bc7a5d2f1b4/.github/workflows/windows_build_qt.yml#L20-L23
I would have expected this could all be cleaned up.

xTVaser (commented):
Just one point of confusion around why the windows builds don't seem to be fully cleaned up of `simd` traces

I think everything should work fine release-wise (the simd was always an extra tag, not something explicitly checked like an OS). But I can double-check it before this merges on my fork.  One less CI build once again 🥳

.github/workflows/scripts/common/name-artifacts.sh
TellowKrinkle: There aren't any in the release pipeline, but MSBuild builds don't use Multi-ISA, and we'll be keeping at least some of them around to ensure you can continue to use MSBuild to build, even if we don't use it for published releases.

(Windows release pipeline builds were changed to use cmake)

xTVaser (approved):
This looks fine to me, the asset names looks like they shouldn't cause a problem but if they do I'll follow up
image

TellowKrinkle:
@xTVaser Can I get one more test to see if it now builds the classic sse4/avx2 pair for windows in addition to the new ones?  We decided to keep them around for a bit longer since cmake builds have never really been tested on Windows.

RedDevilus:
Should be fine to be merged, I expect most people to have update to recent enough nightly minus some exceptions (unavoidable). If someone else wants to do a final test as well.

refractionpcsx2:
yeah it's been nearly 2 months, should be fine now

PR #7532, opened 2022-12-04
Updater: Support multi-isa switchover

TellowKrinkle:
### Description of Changes
Gives the updater the ability to pick releases that don't perfectly match the current release

### Rationale behind Changes
Required to be able to update from a current SSE4/AVX2 build to a multi-isa build

### Suggested Testing Steps
Point updater at your own testing server and see if it can update
(Windows updater is tested by me)

### DIFF
--- a/pcsx2-qt/AutoUpdaterDialog.cpp
+++ b/pcsx2-qt/AutoUpdaterDialog.cpp
@@ -57,7 +57,9 @@
 #define UPDATE_PLATFORM_STR "Linux"
 #endif
 
-#if _M_SSE >= 0x500
+#if MULTI_ISA_SHARED_COMPILATION
+// #undef UPDATE_ADDITIONAL_TAGS
+#elif _M_SSE >= 0x501
 #define UPDATE_ADDITIONAL_TAGS "AVX2"
 #else
 #define UPDATE_ADDITIONAL_TAGS "SSE4"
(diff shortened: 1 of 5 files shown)


RedDevilus:
I wonder if this would also work with people having renamed their exe to PCSX2.exe (when they should use shortcuts but still)

TellowKrinkle:
> I wonder if this would also work with people having renamed their exe to PCSX2.exe (when they should use shortcuts but still)

It'll rename it back, but it should successfully update and launch the new executable

Do you think I should make it rename the new executable to match the existing one?

RedDevilus:
> > I wonder if this would also work with people having renamed their exe to PCSX2.exe (when they should use shortcuts but still)
> 
> It'll rename it back, but it should successfully update and launch the new executable
> 
> Do you think I should make it rename the new executable to match the existing one?

Hmm perhaps people prefer it being pcsx2.exe or their own style but if it's going to cause issues along the way in the future it is perhaps better that it enforces the current (or future default naming). I haven't really have a preference for either anyhow.

TellowKrinkle:
I don't see why it would cause issues in the future (unless we have other things that care about the executable name)

RedDevilus:
> I don't see why it would cause issues in the future (unless we have other things that care about the executable name)

Perhaps it won't break this year and I probably shouldn't care about a what-if situation in the far future. So first option might be fine. Not sure if we should make it in a log message that it's not the default exe name (maybe too paranoid)

lightningterror:
There will be only one exe very soon anyway so you don't have to worry.

TellowKrinkle:
> There will be only one exe very soon anyway so you don't have to worry.

It doesn't really matter how many exes the user has are as long as we don't ship multiple in our releases

The logic is: Find an exe whose name starts with pcsx2 in the update zip, and rename it to be the same as whatever exe launched the updater.

TellowKrinkle:
Okay all tested

PR #7539, opened 2022-12-04
GS-HW: Handle clears with single page width better + use vert color

refractionpcsx2:
### Description of Changes
Handles single page width clears with a large height as a full buffer clear, since HW cannot handle it. Also uses the colour the game wishes to clear with instead of forcing zero.

### Rationale behind Changes
Superman Returns does single page width clears with a non-zero colour which causes massive textures consuming the entire GS memory to be created, causing bad frames to be selected at the present, resulting in flashing, since we're now more specific on which target we use now.

### Suggested Testing Steps
Test games, make sure nothing is broken, especially games which break if you disable safe features or know of ones which do "double half clear", etc.

Fixes screen flashing on Superman Returns: The Video Game.

### DIFF
--- a/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
+++ b/pcsx2/GS/Renderers/HW/GSRendererHW.cpp
@@ -3627,7 +3627,7 @@ void GSRendererHW::DrawPrims(GSTexture* rt, GSTexture* ds, GSTextureCache::Sourc
 	// Blend
 
 	bool blending_alpha_pass = false;
-	if (!IsOpaque() && rt && m_conf.colormask.wrgba != 0)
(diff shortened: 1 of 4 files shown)


pcsx2/GS/Renderers/HW/GSRendererHW.cpp
```diff
@@ -4292,7 +4294,7 @@ bool GSRendererHW::OI_GsMemClear()
 
 				for (int x = r.left; x < r.right; x++)
 				{
-					*pa.value(x) = 0; // Here the constant color
+					*pa.value(x) = color; // Here the constant color
```
F0bes, line 4297: Comment refers to the colour as a constant, which is no longer the case relative to the guest, as it's taken from RGBAQ.

refractionpcsx2: it is constant, I set it constant above, it doesn't change through the loop :)

PR #7542, opened 2022-12-05
Qt: Add tooltips to memcard checkbox options.

lightningterror:
### Description of Changes

Qt: Add tooltips to memcard checkbox options.
### Rationale behind Changes

Adds missing tooltlips frop wx.
### Suggested Testing Steps

Tooltips work.


stenzek (approved):
Code LGTM, although the note about GH should be checked in the future if it still applies.

PR #7545, opened 2022-12-05
GameDB: A few game fixes and tweaks 

JordanTheToaster:
### Description of Changes
HPO Normal causes really bad glitching on the floor in KH2 round sprite breaks light bloom in Price of Persia WW and WRC Rally Evolved needs round sprite half to fix misaligned text.

### Rationale behind Changes
This is bad it shouldn't be doing this:

image

It should look like this with Round Sprite Full:

image

### Suggested Testing Steps
Happy CI is good yes.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -1198,7 +1198,6 @@ SCAJ-20154:
   region: "NTSC-Unk"
   gsHWFixes:
     autoFlush: 1 # Reduces post-processing misalignment.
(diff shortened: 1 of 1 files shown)


RedDevilus:
Texture cache inquisition vanquished!

refractionpcsx2:
Special texture doesn't work. Look at the roofs

None:
image

HPO Normal:
image

Special:
image

however Round Sprite Full seems close to native:
image

Native for reference:
image

PR #7550, opened 2022-12-07
Qt: Resizes + better tooltips + Savestate warning

RedDevilus:
### Description of Changes

Bigger tooltip window
Controllers button shows up correctly at different DPI percentages 
Rename some options.

image

image

image

### Rationale behind Changes

More sense and more obvious settings or options.
### Suggested Testing Steps

Check the GUI in Qt

### DIFF
--- a/pcsx2-qt/MainWindow.ui
+++ b/pcsx2-qt/MainWindow.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>950</width>
-    <height>640</height>
+    <width>1050</width>
+    <height>666</height>
    </rect>
   </property>
   <property name="acceptDrops">
@@ -22,7 +22,7 @@
     <rect>
      <x>0</x>
      <y>0</y>
-     <width>950</width>
+     <width>1050</width>
      <height>22</height>
     </rect>
    </property>
--- a/pcsx2-qt/Settings/ControllerBindingWidget.ui
+++ b/pcsx2-qt/Settings/ControllerBindingWidget.ui
@@ -29,7 +29,7 @@
    <item>
     <widget class="QGroupBox" name="groupBox">
      <property name="title">
-      <string>Controller Type</string>
+      <string>Virtual Controller Type</string>
      </property>
      <layout class="QHBoxLayout" name="horizontalLayout_3">
       <item>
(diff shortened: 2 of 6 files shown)


Florin9doi:
image

refractionpcsx2:
> image

Can't really do that, unless we diligently kept a list of last versions a specific savestate version worked on, which we don't currently.

refractionpcsx2:
Yeah I agree that's probably the best setup for that dialog.

RedDevilus:
Tracking savestate version suck but hooking up:
image

The link to the site and close are all feasible. but it's just one of many revisions to come either way.

RedDevilus:
Is the PR fine for now or does anyone else have doubt/questions about the suggested changes? I plan to do more smaller PRs either way.

Florin9doi:
> Can't really do that, unless we diligently kept a list of last versions a specific savestate version worked on, which we don't currently, but that sounds like a maintenance annoyance, and people _will_ forget.

This is offtopic, but isn't necessary to determine the latest supported version for a specific savestate if we'd save the emulator version inside savestates together with [SaveVersion](https://github.com/PCSX2/pcsx2/blob/0e73bf1e6d0a7f33d5eed135b4b30787753d7693/pcsx2/SaveState.cpp#L947-L962) and we would suggest same version.

stenzek:
Personally I'd rather just do state migration, like I have in duckstation. The USB PR pulls in my state wrapper class which supports this, we can move the other components over it too eventually.

refractionpcsx2:
> > Can't really do that, unless we diligently kept a list of last versions a specific savestate version worked on, which we don't currently, but that sounds like a maintenance annoyance, and people _will_ forget.
> 
> This is offtopic, but isn't necessary to determine the latest supported version for a specific savestate if we'd save the emulator version inside savestates together with [SaveVersion](https://github.com/PCSX2/pcsx2/blob/0e73bf1e6d0a7f33d5eed135b4b30787753d7693/pcsx2/SaveState.cpp#L947-L962) and we would suggest same version.

I guess, I more figured the user would want the most up to date version before it exploded.  Either way, Sten is right, it would be better to replace our savestate system with his so interversion compatibility isn't quite so deadly, although I worry about new stuff not being in the correct state if the state has changed, but you have more luck than you currently do.

PR #7562, opened 2022-12-08
GameDB: Add placeholders for comments + maintenance

RedDevilus:
### Description of Changes

Some comments were blank (probably still missing some) but in the future we should add everything to real comments.
### Rationale behind Changes

More clean comments
### Suggested Testing Steps

Doesn't affect games, only my OCD.


RedDevilus:
Unwanted 😿

PR #7564, opened 2022-12-10
GS: Add internal resolution screenshot option

stenzek:
### Description of Changes

And you can revert to the old un-aspect-corrected behaviour if you want to for some weird reason...

You also get the option of PNG or JPEG.

Zoom is gone because I needed more space in the page, and it was redundant with crop.

### Rationale behind Changes

Decent looking screenshots.

### Suggested Testing Steps

Test new screenshot options.


JordanTheToaster (approved):

refractionpcsx2:
In aspect ratio correct, if IsAnalog and IsProgressive, you need to double the height, otherwise you get a super squashed picture.

Other than that, LGTM

PR #7565, opened 2022-12-10
GS: Preserve deinterlacing render target

KrossX:
### Description of Changes
Copy interlace result into `m_merge` to avoid keeping it as `m_current`, `m_merge` is reusable at this point... right?

### Rationale behind Changes
I think I only notice it when using weave deinterlacing, but when using shadeboost it would seem as if the effect was applied twice every other line. Since weave uses discard, every other line just shows the previous image which is the result of m_current being used for post processing.

### Suggested Testing Steps
Try shadeboost with low brightness and wave deinterlacing, just the BIOS is good enough test. It would give a sort of scanline effect when captured and an overall darker image than no deinterlace. That odd behavior should be gone with this PR.

I don't know how the other deinterlalcing methods are affected, but if there's a discard involved it's probably affected.... maybe.

### DIFF
--- a/pcsx2/GS/Renderers/Common/GSDevice.cpp
+++ b/pcsx2/GS/Renderers/Common/GSDevice.cpp
@@ -387,13 +387,21 @@ void GSDevice::Interlace(const GSVector2i& ds, int field, int mode, float yoffse
 		default:
 			m_current = m_merge;
 	}
+
+	if ((GSConfig.FXAA || GSConfig.ShadeBoost) && m_current != m_merge)
+	{
+		const GSVector2i s = m_current->GetSize();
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
Can you make it only do the copy if shadeboost or fxaa is enabled? just to save the extra copy.  Bilinear doesn't actually touch the target so we don't need to worry about that part of it.

Best thing would probably be to put the check inside shadowboost or fxaa calls

KrossX:
I think the check is good enough, the commit about clears though seems to have no effect. I checked with renderdoc and there was just one clear either way. Although the clear would happen on size mismatch, so maybe it's one sporadic event. Meh.

PR #7567, opened 2022-12-11
GS-SW: Don't anti-blur frame rect when same source and not anti-bluring

refractionpcsx2:
### Description of Changes
Stops it anti-blurring in software mode when both PCRTC framebuffers are the same source but one is offset, when anti-blur is off.

### Rationale behind Changes
Killing the intended by developer behaviour when you don't want it to is bad.

### Suggested Testing Steps
Test Cold Winter in software, Anti-blur vs not.


MrCK1:
Tested with Cold WInter and a bunch of other games. Looks to be working as expected now

PR #7573, opened 2022-12-12
GS: Compute luminance in shader for FXAA

stenzek:
### Description of Changes

Fixes FXAA being broken after https://github.com/PCSX2/pcsx2/commit/b706c25b68b25fe5e4cc84b5420f225f201ac739.

Will be slightly difference since this is using the BT.709 coefficients instead of CCIR 601 - but it matches the non-gather codepath now.

Also gets rid of an additional copy when FXAA/ShadeBoost is active.

### Rationale behind Changes

Regression fix, ever-so-slight VRAM bandwidth savings.

### Suggested Testing Steps

Test FXAA and also make sure I didn't regress #7565.

### DIFF
--- a/bin/resources/shaders/common/fxaa.fx
+++ b/bin/resources/shaders/common/fxaa.fx
@@ -55,19 +55,22 @@ static constexpr sampler MAIN_SAMPLER(coord::normalized, address::clamp_to_edge,
                              [FXAA CODE SECTION]
 ------------------------------------------------------------------------------*/
 
+// We don't use gather4 for alpha/luminance because it would require an additional
+// pass to compute the values, which would be slower than the extra shader loads.
+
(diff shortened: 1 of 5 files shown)


TellowKrinkle:
Does gather actually give any performance improvement at this point?  I think it's now replacing four loads with six gathers, which doesn't sound like much savings.  Maybe just get rid of it.

stenzek:
Without the early exit, it's 2 loads + 6 gathers, versus 8 loads.. so I'm not certain. Could go either way I guess, but I suspect probably not much of an observable difference.

TellowKrinkle:
Oh for the Metal one, could you mark the functions as `static`?  MSL links C-style, and while I don't expect any other files to be naming their functions `FxaaTexAlpha4`, leaking names into the global namespace isn't great either way.

PR #7590, opened 2022-12-16
microVU: Fix mVUcustomSearch

TellowKrinkle:
### Description of Changes
Apparently NOT doesn't set flags

### Rationale behind Changes
Less broken

### Suggested Testing Steps
Test on GCC, which is apparently the one who happened to enter the function with the zero flag cleared


RedDevilus (approved):
Tested on Zorin VM:
Before WX AppImage:
image

After WX AppImage:
image

Should also fix other GCC like flatpak I guess which is unofficial though.

lightningterror (approved):

PR #7592, opened 2022-12-16
GS: Fix dx11 bad shader/crashing on dx10 gpus.

lightningterror:
### Description of Changes

GS-d3d11: Fix dx11 crashing on dx10 gpus.
CAS requires feature level 11.0.

GS: Move osd log for unsupported cas in Renderer code.
### Rationale behind Changes

Crashes are bad mkay.
### Suggested Testing Steps

Test dx10 gpus.

### DIFF
--- a/pcsx2/GS/Renderers/DX11/GSDevice11.cpp
+++ b/pcsx2/GS/Renderers/DX11/GSDevice11.cpp
@@ -91,6 +91,7 @@ bool GSDevice11::Create()
 	m_dev = static_cast<ID3D11Device*>(g_host_display->GetDevice());
 	m_ctx = static_cast<ID3D11DeviceContext*>(g_host_display->GetContext());
 	level = m_dev->GetFeatureLevel();
+	const bool support_feature_level_11_0 = (level >= D3D_FEATURE_LEVEL_11_0);
 
 	if (!GSConfig.DisableShaderCache)
 	{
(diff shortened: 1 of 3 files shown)


pcsx2/GS/Renderers/DX11/GSDevice11.cpp
```diff
@@ -341,7 +342,8 @@ bool GSDevice11::Create()
 			return false;
 	}
 
-	CreateCASShaders();
+	CreateCASShaders(support_feature_level_11_0);
```
stenzek, line 345: Probably cleaner to just do
```
if (support_feature_level_11_0)
  CreateCASShaders();
else
  m_features.cas_sharpening = false;
```

lightningterror: I did thought about it first but then changed my mind so it matches gl more and all the checks are in one place this way.

stenzek: Tbh I'd suggest changing GL then as well, I wrote it that way before I considered GPUs that wouldn't support it.

RedDevilus (approved):
Old VM is good for something I guess:

[   73.8216] D3D Adapter: VMware SVGA 3D
[   73.8217] Creating a 1050x556 flip-discard windowed swap chain
[   73.9114] D3D11 Graphics Driver Info:
[   73.9116] 	Unknown Feature Level
[   73.9116] 	VID: 0x15AD PID: 0x0405
[   73.9116] 	VMware SVGA 3D
[   73.9116] 	Driver Version: 8.17.3.5
[   73.9194] Feature level 11.0 not supported, CAS is unavailable.

pcsx2/GS/Renderers/DX11/GSDevice11.cpp
lightningterror: That should do it.

stenzek (approved):

PR #7595, opened 2022-12-16
GS/OpenGL: Cleanups and GLES support

stenzek:
### Description of Changes

Gets rid of a bunch of old code which isn't used anymore, and converts the texture upload buffer over to StreamBuffer.

GLES won't be as accurate as desktop GL, because of no clip control, there's a lossy 0..1 -> -1..1 -> 0..1 round trip. This causes issues in lots of games, particularly with depth post-processing effects. So, instead, we just normalize the depth buffer to 2^24 instead of 2^32. Which means games that use very large depth values won't render correctly, but that's better than most not rendering correctly.

Added bonus is PCSX2 works inside VMware VMs now (vmwgfx does not support ARB_clip_control on Windows nor Linux).

### Rationale behind Changes

VMware support mainly.

### Suggested Testing Steps

Test OpenGL renderer on each vendor, make sure it still works. Try some texture upload heavy games, and make sure nothing regressed performance-wise. It actually seems ever-so-slightly faster for me in e.g. snowblind gsdumps.


pcsx2/GS/Renderers/OpenGL/GLLoader.h
```diff
@@ -36,9 +30,12 @@ namespace GLLoader
 	extern bool vendor_id_nvidia;
 	extern bool vendor_id_intel;
 	extern bool mesa_driver;
+	extern bool buggy_pbo;
 	extern bool in_replayer;
 
 	// GL
+  extern bool is_gles;
```
lightningterror, line 37: Formatting.

Mrlinkwii (approved):

RedDevilus:
Just addition for Windows and Linux VMs:

OpenGL - Windows:
image

D3D12 + Vulkan - Windows:
image

Linux: 

To be filled.

PR #7597, opened 2022-12-16
VMManager: Better separate shutdown and close for SPU2 and GS

stenzek:
### Description of Changes

This was quite messy before, SPU2 and GS would tear down more stuff than it needed to when we shut down the VM.

Also fixes the crash on VM shutdown if you started in big picture mode - GSshutdown was being called and killing the device..

### Rationale behind Changes

Less crash and less redundant work.

### Suggested Testing Steps

Make sure sound still works, and verify starting a game/bios from big picture -> shutdown -> no crash.


refractionpcsx2:
big picture -> shutdown crashed :D

PR #7607, opened 2022-12-18
USB: Don't crash and burn if there is no ini file for pcsx2.

arcum42:
### Description of Changes
Prevents DeviceProxy from going backwards from the beginning of an iterator.

### Rationale behind Changes
If there is no ini file, such as on first bootup, Ports[i].DeviceType is initialized to -1, which will cause the emulator to crash when it reaches this point.

### Suggested Testing Steps
Check to make sure this doesn't cause any issues with usb.

### DIFF
--- a/pcsx2/USB/USB.cpp
+++ b/pcsx2/USB/USB.cpp
@@ -74,7 +74,7 @@ OHCIPort& USB::GetOHCIPort(u32 port)
 bool USB::CreateDevice(u32 port)
 {
 	const Pcsx2Config::USBOptions::Port& portcfg = EmuConfig.USB.Ports[port];
-	const DeviceProxy* proxy = RegisterDevice::instance().Device(portcfg.DeviceType);
+	const DeviceProxy* proxy = (portcfg.DeviceType != DEVTYPE_NONE) ? RegisterDevice::instance().Device(portcfg.DeviceType) : nullptr;
 	if (!proxy)
 		return true;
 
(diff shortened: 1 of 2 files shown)


stenzek:
Can you change the call site instead? <https://github.com/PCSX2/pcsx2/blob/master/pcsx2/USB/USB.cpp#L77>

```
const DeviceProxy* proxy = (portcfg.DeviceType != DEVTYPE_NONE) ? RegisterDevice::instance().Device(portcfg.DeviceType) : nullptr;
```

and

<https://github.com/PCSX2/pcsx2/blob/master/pcsx2/USB/USB.cpp#L564>
```
const DeviceProxy* proxy = (device != DEVTYPE_NONE) ? rd.Device(device) : nullptr;
```

arcum42:
@stenzek
Just added in those changes.

stenzek (approved):

lightningterror (commented):
Commits should be properly named with prefixes.

refractionpcsx2:
It's fine, I'll just squash it

arcum42:
@lightningterror @refractionpcsx2 
Did an interactive rebase blaming it on USB.

refractionpcsx2:
Darn it, now I have to wait for the poor CI again :joy:

arcum42:
Well, that's what a silly policy like having to have prefixes on every commit leads to...

PR #7612, opened 2022-12-18
USB: Add missing extra controller bindings for Buzz

stenzek:
### Description of Changes

Adds the missing three controllers to the buzz binding list.

### Rationale behind Changes

They were missing.

### Suggested Testing Steps

Test Buzz games. I don't have any.

### DIFF
--- a/pcsx2/USB/usb-pad/usb-pad.cpp
+++ b/pcsx2/USB/usb-pad/usb-pad.cpp
@@ -932,11 +932,29 @@ namespace usb_pad
 	gsl::span<const InputBindingInfo> BuzzDevice::Bindings(u32 subtype) const
 	{
 		static constexpr const InputBindingInfo bindings[] = {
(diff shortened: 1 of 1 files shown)


Simonxdd:
Even the single controller only works with the wired buzzers, not the wireless ones. The Red Buzzer also doesn't light up at all.

I would hope that an implementation could work as well as [this PCSX2 1.6.0 plugin,](https://github.com/Alby1987/BuzzMultiBuzzersPlugin) which had everything working. 

Huge thanks for taking the time to implement USB in the Qt version though, that was my last gripe with the otherwise fantastic new interface.

stenzek:
Not sure what you would light up with a fully emulated controller. If you have the physical controller, USB passthrough would be the solution (but that's a fair bit of work).

For fun, we probably could change the LED on DS4/DS5..

Simonxdd:
Well the current implementation is very much incomplete (so is the previous one in wx), so I suppose I should look at whether or not I can do this myself (I'm very inexperienced with coding lmao).

Issue #4032, opened 2020-12-21
[Feature Request] Overhaul of the screenrecorder 


RedDevilus:
**Description**
This is not the same as #3663 but more an extension upon it. Making it easier to decide what codec, size to pick for the enduser intended (or as close as possible) result.

Current look:
![image](https://user-images.githubusercontent.com/24227051/102820544-b6dd8d00-43d5-11eb-9591-1f8c8951794a.png)

Not very intuitive in my opinion but does a good job for what it needs to do.

**Reason**
Lowers flooding storage because you can see if it's still recording and as reminder it's recording (especially for uncompressed), maybe set a way to stop recording after a certain time ( a minute max recording) or certain filesize (2 GB). Display in OSD of some kind to display how many seconds/minutes have passed while recording and append the current usage in increments every second, at the end give an overview of the format of video, duration, filesize and the filepath.

**Examples**
OBS, Audacity, internal W10 recorder. Lots of different styles that do show the duration but not the filesizes.

Issue #3663, opened 2020-09-02
[Feature Request] Capture video and audio in 1 combined file.


RedDevilus:
The current implementation:
User activates Video Capture.
GSdx will spit out a video file(extensionless for now, but there is a PR that will add .avi to it)
SPU2-X will spit out a audio file (recorder.wav will be changed to audio_recorder.wav also in same PR)

Future implementation:
Perhaps easiest way is to wait till the plugin system is merged with the core.

RedDevilus:
Partial fix by #3655

RedDevilus:
With the GS-Merge and SPU2-X merge, it's possible to improve this;

Master:
- Video (extensionless or compressed)
- Video (extensionless or compressed) + Audio (wav file)

Future goal:
- Combined Video and Audio file
- Audio file dumping without video

PR #7613, opened 2022-12-18
GS: Add video dumping via ffmpeg

### LINKED ISSUE #5840: Notify the user when recording is not started due to it being disabled
### Describe the Bug

Trying to capture a video of the emulation does nothing. The window which would ask me where to save the file to and other parameters does not appear and there is no OSD message. The only thing that happens are some log messages which only appear tangentially related.

### Reproduction Steps

1. Start any game.
2. Either press F12 or go into the main window and click *Capture* > *Video*, it does not matter which one you use here because the result is the same.
3. Nothing happens.

Every time I try to record I do get the following addition to the log, regardless if I just started up the game or idled hours in it:

```
(gsFrame) Invoking command: Sys_RecordingToggle
[GameDB] Searching for 'slus-21452' in GameDB
[GameDB] Found 'slus-21452' in GameDB
[GameDB] Searching for patch with CRC 'cc96ce93'
[GameDB] No CRC-specific patch or default patch found
(GameDB) Enabled Gamefix: VuAddSub
[GameDB] Enabled GS Hardware Fix: halfPixelOffset to [mode=1]
[GameDB] Enabled GS Hardware Fix: roundSprite to [mode=1]
```

The window title of the emulator window in which the game is running does not change. Usually a "Recording..." would be appended but this is not happening. There is also nothing in the stdout/stderr.

Since I never tried recording before I can not say if this is a regression or not. I also tried with an empty `~/.config/PCSX2` but this changed nothing, so I assume it is not a misconfiguration since these are the default configs. It also does not matter if I am in software mode or not.
Unchecking the "Include audio" option does not change this behavior either.

For reference: this is ffmpeg 5.0 if this is relevant. Nothing gets added to the `snaps` directory either. Capturing a simple screenshot via the menu works perfectly.

### Expected Behavior

A window would open and ask me where to save the file to. After that, the recording starts.

### PCSX2 Revision

v1.7.2581 (As of writing I literally just freshly compiled the new version)

### Operating System

### If Linux - Specify Distro

Arch Linux

### LINKED ISSUE #6082: [BUG]: (VfW) codecs are unlisted as of 3/19/22.
### Describe the Bug

When attempting to use PCSX2's built-in recorder on the latest version, it does not bring up a list of all the VfW codecs installed on the user's computer anymore. I spent some time tracking it down and narrowed the issue down to anything after [v1.7.2485](https://github.com/PCSX2/pcsx2/commit/6977323d88c2d79a3e039c87898cedfb5a7b363b). This was the last version all codecs showed up on. I couldn't test v.1.7.2485-1, 2, or 3 due to those builds not being linked, but it's somewhere in that region. I originally brought it up here before investigating more: https://github.com/PCSX2/pcsx2/pull/4523#issuecomment-1126802093

### Reproduction Steps

Start a game and press F12 to bring up the recording menu, then look through the list of codecs to select. Anything with the (VfW) prefix is missing.
image

### Expected Behavior

On versions before v1.7.2486, bringing up this recording menu would give a list of the (VfW) codecs, allowing me to select the one I normally use.
image

### PCSX2 Revision

v1.7.2746

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

stenzek:
### Description of Changes

What the title says. Gets rid of the old archaic baseclasses too, and linux/mac can actually record to a video file now.

You'll need to bring your own copy of ffmpeg v4.x.x. For testing, you can use https://www.duckstation.org/upload/ffmpeg-dlls.zip (drop in your PCSX2 bin directory), but we'll need to figure out a better distribution method in the future.

TODO in the future once wx is gone:
 - Add audio dumping to container at the same time.
 - Threaded dumping for speed.

### Rationale behind Changes

Adds it to Qt, and like everything else, becomes cross platform in the process.

Closes #5840.
Closes #6082.

### Suggested Testing Steps

Test video dumping. It's in the tools menu.


Tokman5:
when you record a video again, some timer thing doesn't seem to be initialized. I recorded the first video for 20secs and the second video for 20secs, and opened the second one with ffplay, it shows `Duration: 00:00:40.01, start: 19.303000`.
the video itself seems to be recorded correctly.

Tokman5:
my personal request:
- add `always ask save location` option to open a dialog even when video capture is enabled via `Toggle video capture` hotkey

stenzek:
> add always ask save location option to open a dialog even when video capture is enabled via Toggle video capture hotkey

Not going to happen, risks making things very messy/potentially deadlocking when fullscreen. Plus core->UI coupling.

Tokman5:
> Not going to happen, risks making things very messy/potentially deadlocking when fullscreen. Plus core->UI coupling.

ah, ok.

RedDevilus:
Related to:
https://github.com/PCSX2/pcsx2/issues/3663
https://github.com/PCSX2/pcsx2/issues/4032

Spikestuff:
One question about dumping which will blend with the updated TAS tools.
When recording is selected outside of game will it wait for the game/replay to be played to actually initialize from frame 0?

Previous instance of PCSX2 which had dumping didn't allow it to be dumped neatly from the start, so if it does it neatly from 0 then it's honestly a great change.

stenzek:
Not currently, no. It's a tricky one, because the video resolution will change once the game loads, and you can't have multiple resolutions in the one file. We could have a "dump on start" option, but it'd be based on when the ELF loads, plus a few frames, not "power on".

JordanTheToaster (approved):

PR #7617, opened 2022-12-18
USB: Add deadzone for USB wheel steering axis

stuken:
### Description of Changes
Adds a deadzone setting to usb wheel axis handling. Also adds a small bit of feedback to the axis values to ensure that full deflection is registered even at extreme deadzone settings.

### Rationale behind Changes
Pad mapping to a wheel axis can be extremely twitchy. This will make it more useable for pad users that wish to map wheel support to a modern xbox/ps4 style controller or for wheels with no self centering or force feedback.

### Suggested Testing Steps
Tested on GT3 and 4 with each compatible wheel profile. The 900degree wheels will always have too much range to cleanly map to pads but less precise models behave quite nicely and all profiles will register full deflection for all deadzone values. Should be tested against other games supporting wheels. Ensure that center drift is avoided, the game registers the first post input deflection from its internal zero and the game registers the max input at the full stick deflection.

### DIFF
--- a/pcsx2/USB/usb-pad/usb-pad.cpp
+++ b/pcsx2/USB/usb-pad/usb-pad.cpp
@@ -157,6 +157,9 @@ namespace usb_pad
 				{SettingInfo::Type::Integer, "SteeringSmoothing", "Steering Smoothing",
 					"Smooths out changes in steering to the specified percentage per poll. Needed for using keyboards.",
 					"0", "0", "100", "1", "%d%%", nullptr, nullptr, 1.0f},
+				{SettingInfo::Type::Integer, "SteeringDeadzone", "Steering Deadzone",
+					"Steering axis deadzone for pads or non self centering wheels.",
+					"0", "0", "100", "1", "%d%%", nullptr, nullptr, 1.0f},
 			};
 
 			return info;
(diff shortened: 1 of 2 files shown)


pcsx2/USB/usb-pad/usb-pad.cpp
```diff
@@ -439,17 +444,25 @@ namespace usb_pad
 		}
 	}
 
+	s16 PadState::applySteeringAxisDeadzone(float value)
```
stenzek, line 447: Try to keep to the same naming convention as the surrounding functions, i.e. title case / `ApplySteeringAxisDeadzone()`.

pcsx2/USB/usb-pad/usb-pad.h
```diff
@@ -336,6 +338,7 @@ namespace usb_pad
 
 		s16 steering_range = 0;
 		u16 steering_step = 0;
+		long steering_deadzone = 0;
```
stenzek, line 341: This probably should be a `s32`. `long` varies depending on platform.

pcsx2/USB/usb-pad/usb-pad.cpp
```diff
@@ -207,6 +210,8 @@ namespace usb_pad
 				1, std::numeric_limits<u16>::max()));
 		}
 
+		steering_deadzone = (steering_range / 100) * USB::GetConfigInt(si, port, devname, "SteeringDeadzone", 0);
```
stenzek, line 213: Probably should be ```
steering_deadzone = (steering_range * USB::GetConfigInt(si, port, devname, "SteeringDeadzone", 0)) / 100;
```

to preserve as much precision as possible.

stenzek (changes requested):

stenzek (approved):

PR #7633, opened 2022-12-19
USB: Power curve option for wheel steering axis.

stuken:
### Description of Changes
Adds a power curve option to the steering axis.

### Rationale behind Changes
Fully linear steering inputs are quite harsh when mapped to pad sticks. This applies an input^n+1/range^n function to damp smaller inputs at the expense of compressing input at higher values. Not recommended for use with a real wheel but helps smooth input for pad users.

### Suggested Testing Steps
Turning your wheel or stick too much shouldn't blow up things.

### DIFF
--- a/pcsx2/USB/usb-pad/usb-pad.cpp
+++ b/pcsx2/USB/usb-pad/usb-pad.cpp
@@ -62,6 +62,25 @@ namespace usb_pad
 		"",
 		"KONAMI"};
 
+	static int MapSteeringCurveExponentOptionToExponent(const std::string& option)
+	{
+		int exponent = 0;
+		if (option == "Low")
+		{
+			exponent = 1;
+		}
+		else if (option == "Medium")
+		{
+			exponent = 2;
+		}
+		else if (option == "High")
+		{
+			exponent = 3;
+		}
+
+		return exponent;
+	}
+
 	static gsl::span<const InputBindingInfo> GetWheelBindings(PS2WheelTypes wt)
 	{
 		switch (wt)
@@ -153,6 +172,7 @@ namespace usb_pad
 	{
 		if (wt <= WT_GT_FORCE)
 		{
+			static constexpr const char* SteeringCurveExponentOptions[] = {"Off", "Low", "Medium", "High", nullptr};
 			static constexpr const SettingInfo info[] = {
 				{SettingInfo::Type::Integer, "SteeringSmoothing", "Steering Smoothing",
 					"Smooths out changes in steering to the specified percentage per poll. Needed for using keyboards.",
@@ -160,6 +180,9 @@ namespace usb_pad
 				{SettingInfo::Type::Integer, "SteeringDeadzone", "Steering Deadzone",
 					"Steering axis deadzone for pads or non self centering wheels.",
 					"0", "0", "100", "1", "%d%%", nullptr, nullptr, 1.0f},
+				{SettingInfo::Type::StringList, "SteeringCurveExponent", "Steering Damping",
+					"Applies power curve filter to steering axis values. Dampens small inputs.",
+					"Off", nullptr, nullptr, nullptr, nullptr, SteeringCurveExponentOptions}
 			};
 
 			return info;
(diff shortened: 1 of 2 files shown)


pcsx2/USB/usb-pad/usb-pad.cpp
```diff
@@ -62,6 +62,36 @@ namespace usb_pad
...
+		std::vector<std::pair<std::string, std::string>> steeringCurveExponentOptions;
+		steeringCurveExponentOptions.emplace_back(std::make_pair("Off", "Off"));
+		steeringCurveExponentOptions.emplace_back(std::make_pair("Low", "Low"));
+		steeringCurveExponentOptions.emplace_back(std::make_pair("Medium", "Medium"));
+		steeringCurveExponentOptions.emplace_back(std::make_pair("High", "High"));
+
+		return steeringCurveExponentOptions;
+	}
+
+	static int MapSteeringCurveExponentOptionToExponent(std::string option)
```
lightningterror, line 76: Can you take care of the codacy warning?

pcsx2/USB/usb-pad/usb-pad.cpp
```diff
@@ -62,6 +62,36 @@ namespace usb_pad
...
+		std::vector<std::pair<std::string, std::string>> steeringCurveExponentOptions;
+		steeringCurveExponentOptions.emplace_back(std::make_pair("Off", "Off"));
+		steeringCurveExponentOptions.emplace_back(std::make_pair("Low", "Low"));
+		steeringCurveExponentOptions.emplace_back(std::make_pair("Medium", "Medium"));
+		steeringCurveExponentOptions.emplace_back(std::make_pair("High", "High"));
+
+		return steeringCurveExponentOptions;
+	}
+
+	static int MapSteeringCurveExponentOptionToExponent(const std::string option)
```
lightningterror, line 76: `const std::string& option`

pcsx2/USB/usb-pad/usb-pad.cpp
```diff
@@ -160,6 +190,9 @@ namespace usb_pad
 				{SettingInfo::Type::Integer, "SteeringDeadzone", "Steering Deadzone",
 					"Steering axis deadzone for pads or non self centering wheels.",
 					"0", "0", "100", "1", "%d%%", nullptr, nullptr, 1.0f},
+				{SettingInfo::Type::StringList, "SteeringCurveExponent", "Power Curve Modifer",
```
stenzek, line 193: Doesn't need to be a callback, that's intended for things that are truely dynamic (like host device lists). Simply declare an array of `const char*` and use that in the previous field.

stuken: Tried that but if doesnt seem to want to populate the global default setting string. It just shows "[]".

Options are defined with:
`static constexpr const char* SteeringCurveExponentOptions[] = {"Off", "Low", "Medium", "High"};`
Used with:
`{SettingInfo::Type::StringList, "SteeringCurveExponent", "Power Curve Modifer",
					"Applies power curve modifer to axis values. Dampens small user inputs.",
					"Off", nullptr, nullptr, nullptr, nullptr, SteeringCurveExponentOptions}`

stenzek: That would be a bug then, I'll have to sort that out.

stuken: Can confirm that the global setting is recognised so its just the display thats the issue.
Do you want me to update to use the array now or just leave as is and revisit once the fix is in?

stenzek: Yeah, please change it if you don't mind - I'll sort out the inherit thing soon.

stuken: Done.

stenzek (approved):
Code LGTM, untested

MomosuWarosu:
Alright, tested the build on a Thrustmaster T150 set up as a GT Force wheel (200 degree steering.) It works as intended. Set to medium, my inputs are dampened towards the middle and come on strong towards the outer range of steering. Set to off, it behaves as it should on a wheel. Rest assured this does not break real wheels.

I'd just suggest naming the option "Linearity" as that's the expected nomenclature when it comes to this kind of tuning for analog axes on input devices.

Mrlinkwii (approved):

PR #7641, opened 2022-12-21
build: remove flag for skipping portaudio, as we no longer use portau…

arcum42:
…dio.

### Description of Changes
This removes the option in build.sh to skip portaudio.

### Rationale behind Changes
We now no longer use portaudio in pcsx2.

### Suggested Testing Steps
Build pcsx2.


arcum42:
When wx gets removed, we'll want to get rid of the --qt6 option, as well as the --gtk2 option, but that's for later.

stenzek (approved):

PR #7647, opened 2022-12-21
GS: Add fancy downsampling shaders

KrossX:
At some point when internal resolution is high enough compared to the viewing area, bilinear filter stops being useful and aliasing occurs. For the few people that use such high internal resolutions, these supersampling filters might help.

Included filters are 4x rotated grid (4 fetches), quincunx (5 fetches), jittered stochastic (3x3 fetches) [(wiki on patterns)](https://en.wikipedia.org/wiki/Supersampling). I tried other filters like bicubic and lanczos but while those seem to work well when upsampling, it was quite the opposite when downsampling. There's an approximation for gamma correction, which I have yet to compare. To do so, edit the `present_xtra.fx` file and change the define.

Also added an antialiased nearest filter for upsampling that can achieve the sharp bilinear effect. This last filter is mostly for comparing performance between the two methods. Should be used with bilinear (smooth) option.


JordanTheToaster (approved):

DonelBueno:
Tested here and it looks even better than setting your desktop to PCSX2's internal resolution through DSR. Great work!

stenzek:
Any reason why we need this in a separate fx/glsl file, and it can't be combined in the main file?

RedDevilus:
> Any reason why we need this in a separate fx/glsl file, and it can't be combined in the main file?

Bad habit of making forum shaders which were seperated because of plugin system?

KrossX:
> Any reason why we need this in a separate fx/glsl file, and it can't be combined in the main file?

Just a minor convenience thing. I had already lost some changes due to copy pasting files, this helped mitigate that too. It can certainly be combined later.

lightningterror:
I prefer if it was combined now, also needs Metal implementation btw.

KrossX:
Added heavier shader version up to 9x9 as suggested by Jordan. Test performance and quality difference, folliage might be a good thing to test with.

> I prefer if it was combined now, also needs Metal implementation btw.

I'd rather wait until settling what remains and what's gone. As for Metal though... I have no idea how to do that, I'll leave that to last.

HelelSingh:
Hi, how do I download this build? I want to use it for screenshots.

kamfretoz:
You can't download it anymore since all the build artifacts have expired.
@KrossX will have to do a rebase or trigger another check to create another build

image

HelelSingh:
Oh, ok. Do you mind teaching me where to look for those files in the future or in other pull requests?

kamfretoz:
On top of the PR page you should see the "Checks" Tab, click on it and pick your operating system on the left.
Once you're in that page, click on Summary on the top-left.
Scroll down you should see a bunch of files you can download.

KrossX:
Rebased.

For those that try and find this feature useful, please provide feedback on your results with the different methods and take special notice if the jittered versions show any improvement and/or annoyance during gameplay.

ps1freak26:
@HelelSingh You need to run the game at a higher internal res than the display resolution. The SSAA shader doesn't run it at 6x res when it is set to 3x res like a game that would support SSAA in the driver control panel.

HelelSingh:
Thanks, that works. I assumed it would behave just like in the Dolphin emulator where you select a SSAA value from 2x to 8x and then select the Internal Res (like 3x in my case) to which the SSAA downsamples down to. Something similar is also available in DuckStation.

@KrossX will that be possible here in PCSX2?

KrossX:
No idea, the MSAA or SSAA you're thinking of is making use of multisampled framebuffers that need to be resolved and work with all the trickery with format changes and whatnot. Seems not a trivial task and quite the pain in the bum. It is indeed a feature I use with DuckStation as I quite like the result as opposed to higher IR.

This PR is just for people using an absurdely high internal resolution expecting a smoother image but getting a jagged result instead. If you're using 3x IR to sort of match your screen/window resolution, try multiplying it for the SSAA value you would've expected; 6x IR for 2x SSAA, 24x IR for 8x SSAA, etc...

Calinou:
> If you're using 3x IR to sort of match your screen/window resolution, try multiplying it for the SSAA value you would've expected; 6x IR for 2x SSAA,

Slightly off-topic, but I believe 6× IR in this case is 4× SSAA, as the pixel count is quadrupled, not doubled. This should be clarified in the documentation.

KrossX:
Added new shader, this one tries to get the number of samples at runtime. Dunno if there's much of a performance impact from having a known loop. 

If performance is fine compared to the other methods, maybe this is the method to keep.

stenzek:
> Dunno if there's much of a performance impact from having a known loop.

The loop count not being constant means the compiler can't unroll it, so it will have a performance impact. But since it's uniform across all shader invocations, it shouldn't hurt _too_ much (plus, it's only running once per frame).

KrossX:
Alright, now that Metal side seems to be working (thanks Macjestic) this is no longer WIP.

bin/resources/shaders/dx11/present.fx
```diff
@@ -395,4 +395,43 @@ PS_OUTPUT ps_filter_lottes(PS_INPUT input)
...
+{
+	PS_OUTPUT output;
+
+	float2 ratio = (u_source_size / u_target_size) * 0.5;
+	float2 steps = floor(ratio);
+	float3 col = sample_c(input.t).rgb;
+	float div = 1;
+
+	for(float y = 0; y < steps.x; y++)
+		for(float x = 0; x < steps.y; x++) {
```
lightningterror, line 427: Formatting, above one too, same goes for the other shaders.

lightningterror: Brackets and missing space after for.

KrossX: Ooop, one more try then.

KrossX: Done!

Mrlinkwii (approved):

PR #7648, opened 2022-12-21
CDVD: Warn the user when a selected game image is on a removable drive.

F0bes:
### Description of Changes
Adds a warning to `CDVDsys_SetFile` when windows reports the source drive to be removable.
[GetDriveTypeW](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getdrivetypew)

### Rationale behind Changes
People have been experiencing poor performance when running their games off of removable media such as portable hard drives and thumb drives.

### Suggested Testing Steps
Try running an image on a USB stick, or any other removable media, and see if you get the warning. Please test things such as big picture mode.

### DIFF
--- a/pcsx2/CDVD/CDVDcommon.cpp
+++ b/pcsx2/CDVD/CDVDcommon.cpp
@@ -37,6 +37,7 @@
 #include "DebugTools/SymbolMap.h"
 #include "Config.h"
 #include "Host.h"
+#include "IconsFontAwesome5.h"
 
 CDVD_API* CDVD = NULL;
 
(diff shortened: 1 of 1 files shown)


F0bes:
Oh, this is Windows only by the way. I don't think writing 150 lines of C for Linux is worth my time in this instance.

JordanTheToaster (approved):

pcsx2/CDVD/CDVDcommon.cpp
```diff
@@ -303,6 +303,22 @@ static CDVD_SourceType m_CurrentSourceType = CDVD_SourceType::NoDisc;
...
+	if (Path::IsAbsolute(newfile))
+	{
+		const auto& splitPath = Path::SplitNativePath(newfile);
+		// GetDriveType() Requires trailing backslashes
+		const auto root = fmt::format("{}\\", splitPath.at(0));
+
+		const auto driveType = GetDriveType(StringUtil::UTF8StringToWideString(root).c_str());
+		if (driveType == DRIVE_REMOVABLE)
+		{
+			Host::AddOSDMessage("Warning! Game disc location is on a removable drive, performance issues such as jittering and freezing may occur.", 5.0f);
```
stenzek, line 316: Probably best to use `Host::AddIconOSDMessage()` here, that way it doesn't duplicate if you start two games quickly after another (you'll need to include `IconsFontAwesome.h`). Also don't need the double console log, all OSD messages also get written to the log now.

For example:
```
Host::AddIconOSDMessage("RemovableDriveWarning", ICON_FA_EXCLAMATION_TRIANGLE, "Game disc location is on a removable drive, performance issues such as jittering and freezing may occur.", Host::OSD_WARNING_DURATION);
```

Otherwise LGTM.

F0bes: Sweet, I've done that instead.

stenzek (approved):

pcsx2/CDVD/CDVDcommon.cpp
```diff
@@ -303,6 +304,21 @@ static CDVD_SourceType m_CurrentSourceType = CDVD_SourceType::NoDisc;
 
 void CDVDsys_SetFile(CDVD_SourceType srctype, std::string newfile)
 {
+#ifdef WIN32
+	if (Path::IsAbsolute(newfile))
+	{
+		const auto& splitPath = Path::SplitNativePath(newfile);
```
stenzek, line 310: Ah, missed this one - that's returning a new vector, so it shouldn't be stored as a reference (`const auto splitPath` would be fine)... surprised the compiler didn't whinge about it.

PR #7652, opened 2022-12-22
GameDB: Fix empty could not find serial and searching for serial logs.

lightningterror:
### Description of Changes

GameDB: Fix empty could not find serial and searching for serial logs.
Fixes the logs from pasting when the size of the serial is 0 which is invalid.
### Rationale behind Changes

Fixes the logs from pasting when they shouldn't.
### Suggested Testing Steps

Make sure the logs only paste when they actually detect a missing serial in gamedb.

### DIFF
--- a/pcsx2/GameDatabase.cpp
+++ b/pcsx2/GameDatabase.cpp
@@ -753,6 +753,10 @@ const GameDatabaseSchema::GameEntry* GameDatabase::findGame(const std::string_vi
 	GameDatabase::ensureLoaded();
 
 	std::string serialLower = StringUtil::toLower(serial);
+
+	if (serialLower.empty())
+		return nullptr;
+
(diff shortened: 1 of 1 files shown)


pcsx2/GameDatabase.cpp
```diff
@@ -761,6 +761,7 @@ const GameDatabaseSchema::GameEntry* GameDatabase::findGame(const std::string_vi
 		return &gameEntry->second;
 	}
 
-	Console.Error(fmt::format("[GameDB] Could not find '{}' in GameDB", serialLower));
+	if (serialLower.size())
```
stenzek, line 764: Personally I'd use `!serialLower.empty()` for clarity, but same effect.

lightningterror: And will also use an early return, at the top instead, so I don't have to do it for the searching gamedb log too.

PR #7667, opened 2022-12-24
CI: Remove wx builds

stenzek:
### Description of Changes

Before we remove wx, just in case anything e.g. on the site breaks.

### Rationale behind Changes

Easier to revert this if it does break, than 600k lines of wx.

### Suggested Testing Steps

Make sure releases get generated correctly.


Mitezuss:
Why was removed ? for me only work this version

refractionpcsx2:
> Why was removed ? for me only work this version

Then I suggest you upgrade your computer.

PR #7671, opened 2022-12-24
Clean up issues related to removing wx.

arcum42:
### Description of Changes
This gets rid of two now obsolete flags in build.sh.

Also, the variable s_dump_root was still being used in two places, causing compilation to fail if ENABLE_OGL_DEBUG was defined, and it was trying to install a now non-existant file at the end of building.

### Rationale behind Changes
Being able to compile pcsx2 is good, as is not having flags in build.sh that don't do anything.

### Suggested Testing Steps
Check to see if OGL debug actually saves to the right directory, as I just made the same changes that were in the rest of the commit that removed that variable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,9 +56,3 @@ if(ACTUALLY_ENABLE_TESTS)
 	add_subdirectory(tests/ctest)
 endif()
 
-#-------------------------------------------------------------------------------
-
-if(NOT WIN32 AND NOT DISABLE_PCSX2_WRAPPER)
-	# special case to avoid having linux files in windows
-	INSTALL(FILES "${CMAKE_SOURCE_DIR}/linux_various/PCSX2-linux.sh"          DESTINATION "${CMAKE_SOURCE_DIR}/bin")
-endif()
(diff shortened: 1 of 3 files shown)


stenzek:
The freetype stuff in build.sh can go too. As can no-translation.

arcum42:
@stenzek 
Good point. Just took care of both of them. I'll probably be looking for more things to clean up later, too, but I figured I'd better get the compilation errors out of the way first.

stenzek (approved):

arcum42:
Not including it in this PR, since I don't really want to go over to Windows at the moment, but pcsx2/windows/Win32.h can probably be completely removed, though some of the includes in it might need to be added to other files.

The define in it isn't used anywhere, and all the externs are for functions that don't exist. Not sure why that doesn't cause compilation issues...

stenzek:
Yeah, that's a leftover from wx. The resources are split over two files as well, which can be combined.

arcum42:
Yeah, that does look like a bit of a mess left over from legacy code and plugins. afxresmw.h looks like it's probably not needed, too. Though I'll admit I generally avoid anything involving .rc files...

arcum42:
Think I'll just go ahead and commit this, so anyone patching CMakefiles.txt hopefully doesn't have to do it twice...

PR #7688, opened 2022-12-26
Pad: SDL: Assign button names starting with Button0 instead of Button21

Florin9doi:
### Description of Changes
Assign button names starting with Button0 instead of Button21

### Rationale behind Changes
- Easier to track the assigned buttons
- Consistent with DInput

Before: 
image

After:
image

### Suggested Testing Steps
Will need to reassign the impacted buttons to work

### DIFF
--- a/pcsx2/Frontend/SDLInputSource.cpp
+++ b/pcsx2/Frontend/SDLInputSource.cpp
@@ -336,7 +336,7 @@ std::optional<InputBindingKey> SDLInputSource::ParseKeyString(const std::string_
 			if (auto value = StringUtil::FromChars<u32>(binding.substr(6)))
 			{
 				key.source_subtype = InputSubclass::ControllerButton;
-				key.data = *value;
+				key.data = *value + std::size(s_sdl_button_names);
 				return key;
 			}
 		}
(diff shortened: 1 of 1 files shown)


refractionpcsx2:
I'm fine with this, but you do realise you will break *everybodies* controller configs, causing a huge support stress, is it really worth that?

Florin9doi:
It's true that some will have to remap their controllers which aren't included in `game_controller_db`, but these controllers were not usable until recently on QT (#7043) and there were barely any reports. Moreover, without this change these controllers will have to be remapped every time new items are added to `s_sdl_button_names` in the future.

stenzek:
It will affect everyone using wheels/joysticks.

da2ce7:
> I'm fine with this, but you do realise you will break _everybodies_ controller configs, causing a huge support stress because people will assume we've broken pads, is it really worth that?

The controller config should be versioned, so that at least we don't *silently* break controller configurations.

stenzek:
> The controller config should be versioned, so that at least we don't silently break controller configurations.

Sure, if you want to add a ton of complexity to the binding system. You're welcome to do so if you like, but it's definitely not something I'm going to be doing.

PR #7703, opened 2022-12-27
Github: Update issue templates

Florin9doi:
### Rationale behind Changes
Windows 8.1 and macOS 10.13 are no longer supported


refractionpcsx2:
Ah yeah, probably should have updated these, thanks

Issue #7241, opened 2022-10-19
[BUG]: Sly 3: Honor Among Thieves - VU Clamping set to "Extra + Preserve Sign" fixes a broken level but breaks shadow rendering


SlyCooperReloadCoded:
### Describe the Bug

This was brought to my attention very recently by user @Logics4.  If you have "Game Fixes" enabled in the "Advanced System" tab of any release of Sly 3, whether it's NTSC-U, PAL, or NTSC-K, character shadows will still work as expected but environment shadows will act very glitchy, rendering through other buildings, and partially popping in and out of existence.  However, if this option is disabled, both environment and character shadows act like they should.

Short video clip of shadows with "Game Fixes" enabled:

https://user-images.githubusercontent.com/58268914/196756117-8c7f65b3-9346-4aae-a0cb-f001cec7c8f9.mp4

Short video clip of shadows with "Game Fixes" disabled:

https://user-images.githubusercontent.com/58268914/196756208-2eb01821-bfda-413d-b972-a41c2fa12f39.mp4

This has been replicated across my computer and the computer of user, @Logics4.  I'll add both our CPUs and GPUs to issue.

### Reproduction Steps

1.  Edit the custom profile of any release of Sly 3.
2.  Make sure "Game Fixes" is enabled.
3.  Go ingame.

### Expected Behavior

Environment shadows should stay rendered at all times, not clip through buildings, or disappear based on camera angle.

### PCSX2 Revision

v1.7.3461

### Operating System

Windows 10 (64bit)

### CPU

Intel Core i9-10900K (Me), AMD Ryzen 7 5800H (Logics4)

### GPU

NVIDIA RTX 2080 Super (Me), NVIDIA GeForce RTX 3060 (Logics4)

GN876:
I tested this game last week, the camera in the level The Ancestors' Gauntlet is broken without gamefixes. VU clamping needs to be on either extra or extra + preserve sign. In my testing, extra + preserve has better performance. Seems that the rest of the game works fine with VU clamping on normal.

https://forums.pcsx2.net/Thread-Bug-Report-Sly-3-Honor-Among-Thieves-NTSC-U

refractionpcsx2:
why didn't you check that when you added the fixes 17 days ago? https://github.com/PCSX2/pcsx2/pull/7121/files

refractionpcsx2:
[...]

PR #7707, opened 2022-12-27
Config: Add separate VU0/VU1 round/clamp mode options

### LINKED ISSUE #5188: [Feature Request]: Per-component Rounding/Clamping setting.
### Description

Currently, in PCSX2 we have 2 grouped categories, EE/IOP and VUs. When applying a rounding or clamping fix, this will affect all the components inside that group.

This approach is fine for the vast majority of games, but, a very few cases will break that approach.

Kung Fu Panda and Hardware Online Arena are 2 very good exemple of this. They will booth require 2 different settings on each VU. Which means PCSX2 will never be able to play those games properly in it's current form.

The request is pretty straight forward, being able to adjust clamping and rounding on a per-component basis so we can cover every cases.

I guess this is more of a 2.0 release request since I would imagine having QT would massively simplify such a request.

Could also do per address settings for specific parts of a program which require clamping/rounding.

### Reason

Increasing compatibility. In the current state of PCSX2 this can represent one important factor.

### Examples

Official Sony's emulator.

stenzek:
### Description of Changes

As discussed in #7241.

~Untested apart from checking the config loads as expected. Not complete, the big picture UI hasn't been updated, nor have the ints for clamping.~

~Also has no gamedb support yet.~

Note that when testing you'll probably want to disable game fixes (also in the advanced tab), to stop the existing gamedb overriding it (if any). Confirm in the console/log that the correct settings are being applied.

### Rationale behind Changes

#7241.

### Suggested Testing Steps

None at the moment.

Closes #5188

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -1847,6 +1847,8 @@ SCED-50750:
 SCED-50781:
   name: "Destruction Derby Arenas [Beta]"
   region: "PAL-M5"
+  roundModes:
+    vu1RoundMode: 0 # Fixes tyre textures.
   gsHWFixes:
     halfPixelOffset: 2 # Corrects shadow position.
 SCED-50825:
@@ -3026,6 +3028,8 @@ SCES-50781:
   name: "Destruction Derby Arenas [Beta, Promo, & Full Retail]"
   region: "PAL-M6"
   compat: 5
+  roundModes:
+    vu1RoundMode: 0 # Fixes tyre textures.
   gsHWFixes:
     halfPixelOffset: 2 # Corrects shadow position.
 SCES-50791:
@@ -3425,6 +3429,8 @@ SCES-51977:
 SCES-51979:
   name: "Destruction Derby Arenas"
   region: "PAL-M6"
+  roundModes:
+    vu1RoundMode: 0 # Fixes tyre textures.
   gsHWFixes:
     halfPixelOffset: 2 # Corrects shadow position.
 SCES-52004:
@@ -43556,6 +43562,8 @@ SLUS-20855:
   name: "Destruction Derby Arenas"
   region: "NTSC-U"
   compat: 5
+  roundModes:
+    vu1RoundMode: 0 # Fixes tyre textures.
   gsHWFixes:
     halfPixelOffset: 2 # Corrects shadow position.
 SLUS-20856:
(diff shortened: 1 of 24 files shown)


refractionpcsx2:
It looks like Setting VU1 RoundMode to Nearest is enough to fix the wheels in Destruction Derby Arenas without the tyres randomly changing shape, nice! :)

refractionpcsx2:
Could be worth revisting Hitman Contracts, regarding #5719

kakashi314:
Most of the games will only require VU1 clamping/rounding.

refractionpcsx2:
Possibly yes, but for GameDB entries, I would recommend assuming both, unless otherwise required, or that's going to need so much testing to find out what needs what, and I don't think it's really worth the time.  But of course any games which have issues with the current round modes, or couldn't use them because of issues should be revisited.

SlyCooperReloadCoded:
This most likely fix the Sly 3 bug without breaking shadows, just set VU0 clamping to Extra + Preserve Sign, either that or due to the randomness of the bug, I got lucky 30 level loads in a row.  Nah, this is the fix, it's gotta be, and shadows are fine.  Very good!

SlyCooperReloadCoded:
Question, that one Ratchet and Clank mine glove bug, I don't play that game but wasn't that a VU issue?  Could be worth looking into, that is a very long-standing bug.

kakashi314:
> Question, that one Ratchet and Clank mine glove bug, I don't play that game but wasn't that a VU issue? Could be worth looking into, that is a very long-standing bug.

The way i identify Which vector unit is responsible. is to guess whether it's visual (vu1) or game breaking (vu0).

kakashi314:
> Question, that one Ratchet and Clank mine glove bug, I don't play that game but wasn't that a VU issue? Could be worth looking into, that is a very long-standing bug.

Most likely a VU0 issue.

JanLeonard:
> Question, that one Ratchet and Clank mine glove bug, I don't play that game but wasn't that a VU issue? Could be worth looking into, that is a very long-standing bug.

this one are you refferring to? #354

SlyCooperReloadCoded:
That looks like the one.  I see VU mentioned in there.  Yikes it's that old of an issue?  Hopefully this fixes that.

stenzek:
FSUI/gamedb stuff added, good to go.

PR #4541, opened 2021-07-13
GameDB: Add dynamic patching for Ratchet & Clank games

refractionpcsx2:
Removed IPUWait hack as it is no longer required

### Description of Changes
Added dynamic patching system in the recompiler for Ratchet & Clank games toggled by gamefix. Fixes various shadow and SPS problems, improves spawn logic in Q9 station arena challenges, also problems with Ammu-nation in Up Your Arsenal.

### Rationale behind Changes
Ratchet & Clank dynamically allocates memory for the program to run from, so traditional patching is not possible and can cause TLB misses.

### Suggested Testing Steps
Play the Ratchet & Clank games (Aside from Size Matters), places where it went a bit loopy should be okay now.  Turret is still broken, but may be able to add a patch for that before this PR is merged.

PR #7708, opened 2022-12-27
GameDB/Patching: Add dynamic EE JIT patching

### LINKED ISSUE #354: Ratchet & Clank 2 - Megaturret Doesn't Work Correctly

refractionpcsx2:
### Description of Changes
Reimplement dynamic EE JIT patching as was first attempted in #4541, but better and actually generic for use with any game.

### Rationale behind Changes
We needed dynamic patching to fix some bugs with Ratchet & Clank 2, specifically the Mega Turret problem.

### Suggested Testing Steps
Test Ratchet & Clank 2 PAL and see if the Mega Turret works. You can also mess with the GameDB and add your own.

To add your own the format is like this

```
  dynaPatches:
    - pattern:
        - { offset: 0x0, value: 0x0280202d }
        - { offset: 0x4, value: 0x3c0140d0 }
        - { offset: 0x8, value: 0x44810800 }
        - { offset: 0x10, value: 0x46000836 }
      replacement:
        - { offset: 0x4, value: 0x3c013f80 }
 ```

The important ones here are the pattern and replacement. 

The pattern is the sequence of EE instructions it needs to match in order to apply the patch, each code will be tested at an offset from the current program counter being compiles.

The replacement is the code which will be applied as a patch to the EE code, as with the pattern, it will be applied as an offset from the program counter being tested.

You can add multiple patches for different things in the game by duplicating all the code except the `dynaPatches:` line, as - signifies an array entry.

Closes #354 although the real solution is Soft Floats, this'll do for now.

### DIFF
--- a/pcsx2/Patch.cpp
+++ b/pcsx2/Patch.cpp
@@ -33,8 +33,10 @@
 // the only consumer, so it's not made public via Patch.h
 // Applies a single patch line to emulation memory regardless of its "place" value.
 extern void _ApplyPatch(IniPatch* p);
+extern void _ApplyDynaPatch(const DynamicPatch& patch, u32 address);
 
 static std::vector<IniPatch> Patch;
+static std::vector<DynamicPatch> DynaPatch;
 
 struct PatchTextTable
 {
@@ -139,6 +141,7 @@ int LoadPatchesFromString(const std::string& patches)
 void ForgetLoadedPatches()
 {
 	Patch.clear();
+	DynaPatch.clear();
 }
 
 // This routine loads patches from a zip file
@@ -284,5 +287,20 @@ void ApplyLoadedPatches(patch_place_type place)
 	{
 		if (i.placetopatch == place)
 			_ApplyPatch(&i);
+	}
+}
+
+void ApplyDynamicPatches(u32 pc)
+{
+	for (const auto& dynpatch : DynaPatch)
+	{
+		_ApplyDynaPatch(dynpatch, pc);
+	}
+}
+
+void LoadDynamicPatches(const std::vector<DynamicPatch>& patches)
+{
+	for (const DynamicPatch& it : patches){
+		DynaPatch.push_back(it);
 	}
 }
--- a/pcsx2/Patch.h
+++ b/pcsx2/Patch.h
@@ -34,12 +34,12 @@
 // - The 0 cheats - cheats are enabled but nothing found/loaded from the "cheats" folder.
 // - The 6 widescreen patches are 6 pnach-style patch lines loaded either from cheats_ws folder or from cheats_ws.zip
 
-
 #include "common/Pcsx2Defs.h"
 #include "SysForwardDefs.h"
-#include "GameDatabase.h"
 #include <string>
 #include <string_view>
+
+struct IConsoleWriter;
 
 enum patch_cpu_type {
 	NO_CPU,
@@ -97,6 +97,18 @@ struct IniPatch
 	u64 data;
 };
 
+struct DynamicPatchEntry
+{
+	u32 offset;
+	u32 value;
+};
+
+struct DynamicPatch
+{
+	std::vector<DynamicPatchEntry> pattern;
+	std::vector<DynamicPatchEntry> replacement;
+};
+
 namespace PatchFunc
 {
 	PATCHTABLEFUNC author;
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by kakashi314** [comment]
Very good idea! we may also be able to use it as a speedhack. but that's for another day.

File: pcsx2/Patch.cpp [hunk 1]
```diff
@@ -33,8 +33,10 @@
 // the only consumer, so it's not made public via Patch.h
 // Applies a single patch line to emulation memory regardless of its "place" value.
 extern void _ApplyPatch(IniPatch* p);
+extern void _ApplyDynaPatch(DynamicPatch patch, u32 address);
```
stenzek, line 36:
```suggestion
extern void _ApplyDynaPatch(const DynamicPatch& patch, u32 address);
```

pcsx2/Patch.cpp
```diff
@@ -33,8 +33,10 @@
...
 
 static std::vector<IniPatch> Patch;
+static std::vector< DynamicPatch> DynaPatch;
```
stenzek, line 39:
```suggestion
static std::vector<DynamicPatch> DynaPatch;
```

pcsx2/Patch.cpp
```diff
@@ -286,3 +289,18 @@ void ApplyLoadedPatches(patch_place_type place)
 			_ApplyPatch(&i);
 	}
 }
+
+void ApplyDynamicPatches(u32 pc)
+{
+	for (auto dynpatch : DynaPatch)
```
stenzek, line 295:
```suggestion
	for (const auto& dynpatch : DynaPatch)
```

pcsx2/Patch.cpp
```diff
@@ -286,3 +289,18 @@ void ApplyLoadedPatches(patch_place_type place)
...
+	{
+		_ApplyDynaPatch(dynpatch, pc);
+	}
+}
+
+void LoadDynamicPatches(const std::vector<DynamicPatch>& patches)
+{
+	for (auto it = begin(patches); it != end(patches); ++it) {
```
stenzek, line 303:
```suggestion
	for (const DynamicPatch& it : patches)
```

pcsx2/Patch.h
```diff
@@ -34,10 +34,9 @@
 // - The 0 cheats - cheats are enabled but nothing found/loaded from the "cheats" folder.
 // - The 6 widescreen patches are 6 pnach-style patch lines loaded either from cheats_ws folder or from cheats_ws.zip
 
-
```
stenzek, line 37: I think you mentioned this, but is it possible to avoid pulling in the console header by changing the files where Patch.h is included instead?

pcsx2/Patch_Memory.cpp
```diff
@@ -590,6 +590,22 @@ void _ApplyPatch(IniPatch *p)
 	}
 }
 
+void _ApplyDynaPatch(DynamicPatch patch, u32 address)
+{
+	for (auto pattern : patch.pattern)
```
stenzek, line 595:
```suggestion
	for (const auto& pattern : patch.pattern)
```

pcsx2/Patch_Memory.cpp
```diff
@@ -590,6 +590,22 @@ void _ApplyPatch(IniPatch *p)
...
+	{
+		if (*(u32*)PSM(address + pattern.offset) != pattern.value)
+			return;
+	}
+
+	PatchesCon->WriteLn("Applying Dynamic Patch to address 0x%08X", address);
+	// If everything passes, apply the patch.
+	for (auto replacement : patch.replacement)
```
stenzek, line 603:
```suggestion
	for (const auto& replacement : patch.replacement)
```

pcsx2/Patch.h
refractionpcsx2: It was pulling the console stuff from the gamedatabase, but that was a circular include :(

PR #7709, opened 2022-12-27
CDVD: fix compilation if DISABLE_RECORDING is defined.

MaddTheSane:
### Description of Changes
This makes it so that CDVD can compile if the `DISABLE_RECORDING` macro is defined.

### Rationale behind Changes
If `DISABLE_RECORDING` was defined, the header `"Recording/InputRecording.h"` would not be included, but the code still expected `g_InputRecording` to be defined.

### Suggested Testing Steps
None that I can think of.

### DIFF
--- a/pcsx2/CDVD/CDVD.cpp
+++ b/pcsx2/CDVD/CDVD.cpp
@@ -870,6 +870,7 @@ void cdvdReset()
 	cdvd.Action = cdvdAction_None;
 	cdvd.ReadTime = cdvdBlockReadTime(MODE_DVDROM);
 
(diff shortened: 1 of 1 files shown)


stenzek:
We could probably remove that define now. Input recording is in the core, and wx is gone.

stenzek:
https://github.com/PCSX2/pcsx2/commit/f0a7c41b81d8f3ce7888624786a0ea64ba09a689 removed the last instances of `DISABLE_RECORDING`, and it wasn't defined anywhere in the build system anyway.

PR #7716, opened 2022-12-28
Lower minimum GPU requirements

j8r:
### Description of Changes

PCSX2 having gone to several raws of optimizations, it can run now on more modest hardware.
A little i3/i5 PC with an Intel HD 530 is capable of running quite a number of PS2 games in a playable state.
I don't know about equivalent AMD/Nvidia card, Vega graphics can probably be fine.
Cards with a little more Passmark score like the Nvidia GTS 450/GTX 550 Ti and the AMD Radeon HD 5770 remains to be tested.

### Rationale behind Changes

While not ideal to comfortably achieve 100% speed at all times for all games,  people will now know that they can repurpose little PCs to play some PS2 games. 

### Suggested Testing Steps

Try for example God of War 2 or Shadow of the Colossus on a i3 6100T, it will struggle a bit at native at times but other than that (and a few tweaks can help), can be playable, especially less demanding games.
Note: dual-channel makes a BIG difference for integrated graphics.


Mrlinkwii:
not happen

JordanTheToaster:
With how wide the range is I am not in favour of reducing it and honestly wanted it to be raised more.

Mrlinkwii:
> With how wide the range is I am not in favour of reducing it and honestly wanted it to be raised more.

yeah

j8r:
What's the criteria of the minimum requirement for the GPU then? I personally can play most of the games at fullspeed.

JordanTheToaster:
> What's the criteria of the minimum requirement for the GPU then? I personally can play most of the games at fullspeed.

GPUs with new enough features to be consistent with what the program requires now and performance that will be enough for most of the non GPU burning titles.

j8r:
The Vulkan, OpenGL and GLSL version you mean, or because Intel graphics are architectured differently (tiles)? Here the Intel HD 530 have Vulkan 1.3 and OpenGL 4.6, should be fine.

Performance wise, at native, it is enough in most games. It is a minimum requirements, thus not recommended.

JordanTheToaster:
> The Vulkan, OpenGL and GLSL version you mean, or because Intel graphics are architectured differently (tiles)? Here the Intel HD 530 have Vulkan 1.3 and OpenGL 4.6, should be fine.
> 
> Performance wise, at native, it is enough in most games. It is a minimum requirements, thus not recommended.

It's the minimum for a good experience outside of GPU heavy titles recommend is for those games as well as the heavy GPU side of it.

j8r:
Understdood, thanks! I think it fits the bill, but as the PCSX2 developers wish.
What's sure is a GPU less powerful than a GTX 750 Ti can largely play nearly all games, what's alot more limiting in the actual minimum requirements is the CPU.

Ryzen APUs are also surprisingly very good at it, and even for RPCS3 in some circumstances.

j8r:
I'd even say, the minimum CPU requirement could be raised at least at 1800 CPU Passmark.

JordanTheToaster:
> I'd even say, the minimum CPU requirement could be raised at least at 1800 CPU Passmark.

I'll probably do that soon but as for GPU I had an Xe 48EU which was 2300 on passmark and I had a rough time getting 100% steady on games like FFX so 3000 is a far safer bet.

j8r:
I'll try to run that game on my side.
Are you on dual-channel though? The CPU is not bottlenecking I suppose.

Note also I'm on Linux/Mesa, the drivers may be more optimized.

PR #7726, opened 2022-12-29
Readme: Update specifications requirements

JordanTheToaster:
### Description of Changes
Update brings the requirements more in line with modern hardware.

### Rationale behind Changes
More accurate recommendation's are always good.

### Suggested Testing Steps
Make sure the CI is a good boi.


README.md
```diff
@@ -22,17 +22,17 @@ Installers and binaries for both stable and development builds are available fro
...
+| - Windows 10 21H2 (1809 or later) (64-bit) <br/> - Ubuntu 20.04/Debian or newer, Arch Linux, or other distro (64-bit) <br/> - macOS 10.14 | - Supports SSE4.1 <br/> - [PassMark Single Thread Performance](https://www.cpubenchmark.net/singleThread.html) rating near or greater than 1800 <br/> - Two physical cores, with hyperthreading | - Direct3D10 support <br/> - OpenGL 3.x support <br/> - Vulkan 1.1 support <br/> - Metal support <br/> - [PassMark G3D Mark](https://www.videocardbenchmark.net/high_end_gpus.html) rating around 3000 (Geforce MX 450 Radeon RX 560 Intel Arc A380 Steam Deck) <br/> - 2 GB Video Memory | 4 GB |
```
lightningterror, line 25: Was it discussed to raise the str requirements?
I'd also remove the steam deck mention, and leave 750 as it is since it's still supported and people are much more familiar than the mx 450.

JordanTheToaster: Yes we agreed that raising them was fine and SD was added as Tellow pointed out it would be close to that spec anyway but I'll remove it.

j8r: Note this minimum requirements are higher than Yuzu's and RPCS3's (the latter being GTX 400 and HD 5000 series), but their definitions are different: theirs is "For being able to run any games at all, no performance or stability guarantees", and recommended is "For running all playable games with playable performance".

Maybe there could be 3 sections:
- Minimum: bare minimum to run the emulator and the easiest games
- Recommended native: most games are playable at native
- Recommended 1080p: most games are playable at 1080p

JordanTheToaster: The point of the minimum level is what someone should really be using because I'm not going to go to someone and say use a GTX 480 for PCSX2 you can but you really shouldn't be using something that old.

j8r: Well, the minimum requirements are usually not recommended - the latter being another section.
Or maybe here by "minimum", you mean "minimum recommended".

PR #7732, opened 2022-12-29
mVU: Align x86ptr based on AVX2 caps

refractionpcsx2:
### Description of Changes
Aligns the VU JIT buffer to 32byte for AVX2 and 16byte for SSE4, regardless of which version of PCSX2 you use.

### Rationale behind Changes
When using an SSE4 build (which the MultiISA technically will be) it could have aligned to 16bytes (128bits) not 32bytes(256bits) which is required for AVX2, causing crashes when using VMOVAPS (aligned move) since AVX2 instructions can still be used if available.

### Suggested Testing Steps
Already tested, just need to wait for the CI

### DIFF
--- a/pcsx2/PCSX2Base.h
+++ b/pcsx2/PCSX2Base.h
@@ -31,14 +31,6 @@
 	#error PCSX2 requires compiling for at least SSE 4.1
 #endif
 
-// Require 32 bit alignment for vectors for AVX2.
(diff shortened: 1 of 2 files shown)


Mrlinkwii (approved):

SlimDread:
> (256bytes)

you mean bits?

refractionpcsx2:
yes

PR #7748, opened 2022-12-31
UI/AudioSettings: Correct recommended value for overlap

xujibbs:


JordanTheToaster:
You need to fill out the details and name the commit properly but also why? 20ms is fine as it is.

xujibbs:
Overlap maxiumn is 15 ms

> You need to fill out the details and name the commit properly but also why? 20ms is fine as it is.

Max. Overlap is 15 ms, you suggest 20 ms? I think it is typing mistakes.

refractionpcsx2:
Looks like it was a copy paste error from seek window. thanks!

PR #7749, opened 2022-12-31
GameDB: Set preloading to Partial for Tekken 4

JordanTheToaster:
### Description of Changes
Sets texture preloading to partial for Tekken 4 to prevent it from exploding most violently.

### Rationale behind Changes
Performance being bad is bad we should make it good.

### Suggested Testing Steps
Make sure CI is happy.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -3068,6 +3068,7 @@ SCES-50878:
   compat: 5
   gsHWFixes:
     alignSprite: 1
+    texturePreloading: 1 # Performs much better with partial preload.
 SCES-50885:
   name: "Ape Escape 2"
   region: "PAL-E"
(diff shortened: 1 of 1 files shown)


GitHub-Khi:
Did you also test this on the 50Hz/60Hz/Progressive selection screen (i used the PAL version as a test)? Because Partal seems to have a weakness with it.

Multi Frame GS-Dump:
[Tekken 4_SCES-50878_20221231145547.zip](https://github.com/PCSX2/pcsx2/files/10327449/Tekken.4_SCES-50878_20221231145547.zip)

JordanTheToaster:
> Did you also test this on the 50Hz/60Hz/Progressive selection screen (i used the PAL version as a test)? Because Partal seems to have a weakness with it.
> 
> Multi Frame GS-Dump: [Tekken 4_SCES-50878_20221231145547.zip](https://github.com/PCSX2/pcsx2/files/10327449/Tekken.4_SCES-50878_20221231145547.zip)

In game it's slower or just ends up disabling itself regardless and using paltex nukes the framerate so the only solution is partial preloading.

Mrlinkwii (approved):

PR #7760, opened 2023-01-01
x86/microVU: Only update MXCSR when it's different

stenzek:
### Description of Changes

Saves reloading it when the EE and VU0 rounding modes are the same.

### Rationale behind Changes

Minuscule better codegen.

### Suggested Testing Steps

Check a couple of games with gamedb round mode changes, make sure no SPS soup says hello. But should be fine.


refractionpcsx2:
Seems to still work fine :) I tried Burnout 2 with MTVU on and off (since it's VU1) and I tried Spiderwick Chronicles which is VU0, they all seemed to work correctly still, even with manually setting the units involved

PR #7768, opened 2023-01-01
GameDB: Comment all EE/VU fixes, add some missing ones

refractionpcsx2:
### Description of Changes
Makes a bunch of VU gamefixes more specific to the correct VU unit, also add missing fixes for guitar hero games (and corrected some). Only really round modified for the most part, clamping is less problematic to calculation errors.

Also filled in comments for all VU/EE rounding and clamping settings in the DB.

### Rationale behind Changes
Nicer to make things specific to the unit that needs them, less chance of problems.

### Suggested Testing Steps
I guess if you wanna test the games that were modified to make sure they're okay, otherwise, check the CI.

Were missing but now added:
Guitar Hero - Van Halen
Guitar Hero - Smash Hits/Greatest hits
Guitar Hero - Aerosmith on Tour
Simple 2000 Series Vol. 48 - The Taxi - Untenshu ha Kimi da

Can't test:
Sly 3 (bombs/ladders) - GiladN on Discord is testing for me.
Freaky Flyers (bad effects) - not sure it really matters with the patch or not, but leaving alone.

Can't reproduce so left as they were:
Ratatouille (minor lines appearing) Could not see any, maybe only at specific times?
Disney-Pixar WALL-E (minor lines) Could not see any, maybe only at specific times?
Disney-Pixar Toy Story 3 (minor lines) Could not see any, maybe only at specific times?
Star Wars - Bounty Hunter / Jango Fett (no idea, no comment), did manage to work out where the rounding was needed

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -278,7 +278,7 @@ PBPX-95506:
         // 7ACF7E03
         comment=You must enable the EETimingHack when playing the WRC demo to avoids random hangs.
         comment=You must change the EE Clamping mode to Full when playing the Klonoa 2 demo to avoids misplaced objects.
-        comment=You must change the VU Clamping mode to extra when playing the Klonoa 2 demo to fixes reflections issues.
+        comment=You must change the VU0 Clamping mode to extra + sign when playing the Klonoa 2 demo to fixes reflections issues.
 PBPX-95509:
   name: "Linux Release 1.0 Runtime Environment [Disc 1]"
   region: "PAL-E"
@@ -724,7 +724,7 @@ SCAJ-20066:
     mipmap: 2 # Mipmap + trilinear, improves ground textures to match sw renderer.
     trilinearFiltering: 1
     halfPixelOffset: 1 # Fixes weird edge shadows and depth bleed which happens on the edge as well.
-  memcardFilters: #vuClampMode = 2  Text in GT mode works.
+  memcardFilters:
     - "SCAJ-20066"
     - "SCAJ-30006"
     - "SCAJ-30007"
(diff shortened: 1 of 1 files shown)


kakashi314:
We love katamari's sps comes after you finish the level and your "Roll" turns into a planet.

refractionpcsx2:
Any level? Would be good to test that!

kakashi314:
> Any level? Would be good to test that!

Any of them will do.

refractionpcsx2:
Alright, thanks, I'll give it a shot :)

refractionpcsx2:
Are we referring to this line around it?

image

kakashi314:
> Are we referring to this line around it?
> 
> image

That's what i'm referring to yes. as for what the author of the fix referred to ? i do not know.

refractionpcsx2:
I'm not sure that's SPS :/ Unless it's being generated before you get to this screen, because even with gamefixes on it's there (though I did the level without)

kakashi314:
> I'm not sure that's SPS :/ Unless it's being generated before you get to this screen, because even with gamefixes on it's there (though I did the level without)

I think for this game it's best if both VU's have the rounding/clamping on because it's a game with a lot of geometry calculations. as for why they called it an sps ? i agree with you, this is not a spiky polygon syndrome case.

refractionpcsx2:
yes, but I'm trying to work out WHICH VU needs it.

refractionpcsx2:
I just tested from scratch, with gamefixes on, and that glitch still happens, so that's not it at all :(

I can confirm the Extra + sign clamping on VU1 is needed ingame to stop some SPS, but I can't reproduce the rounding, and it's mildly annoying.

refractionpcsx2:
Okay I managed to reproduce it in Katamari Damacy, and I presume they're the same engine since they had the same options, so I could update it.

PR #7769, opened 2023-01-01
x86regs:  iCore.h cleanup.

arcum42:
### Description of Changes
Change x86regs[i].reg to be signed, as -1 is used as unsigned on it. Change some defines to enums.

### Rationale behind Changes
We are testing if it's >=0 and assigning it -1 in at least one place, so it logically should be signed. Those defines are clearly acting as enums, so it's better to actually have them be enums.

### Suggested Testing Steps
See if anything breaks due to now honoring regs being -1 and other changes.


refractionpcsx2:
I expect the correct fix here is to make it s8, because it should be able to be set to -1 (unassigned)

arcum42:
That's a possibility as well. I wonder if anything actually sets it to -1?

refractionpcsx2:
yes, in microVU_IR.h line 969 :)

arcum42:
Alright, so the current behavior is broken, then...

refractionpcsx2:
yeah, I'm guessing it was intended to go negative

arcum42:
Well, changed it to be signed, so the question will now be if anything breaks. :P

refractionpcsx2:
I guess that's what the testers are for :D

refractionpcsx2:
but yeah looking at the existing comment, I assume the original intention was 0 was the "unassigned" value, but I guess that changed along the way :D

arcum42:
I *was* wondering what the difference between unused and unassigned was...

refractionpcsx2:
I don't think there is :P the .reg field is what VI register the x86 reg contains, and that can technically be 0-15, so negative 1 would be a "not a valid reg" value, or unused/unassigned.

arcum42:
I suppose the other question would be the difference between 0/-1 for not in use, and the inuse flag in the same struct.

A lot of these defines should probably be consts or enums, too, but that's probably for another day...

refractionpcsx2:
well all these variables can be different depending on what they're doing, so they can't really be const.

arcum42:
Oh, I was thinking of the defines above the struct. X86TYPE_TEMP, XMMINFO_READLO, and so on. Seems like it'd make sense as an enum of x86types and one of xmminfo, for example.

refractionpcsx2:
oh, yes, that should probably be an ENUM

arcum42:
At the risk of pr creep, changing this to a general cleanup pr, and taking care of the enums.

stenzek:
I can't remember if I ended up using the negative value for temps or not, so to be honest, I'd rather you leave it as-is until I can check all the call sites.

arcum42:
Alright, I'll close this for the moment, then.

stenzek:
There's no temps allocated currently via the VI register allocator, so *technically* -1 isn't needed as a reg value. But, for consistency with the XMMs, I think it's probably a good idea to change it to signed. And in case anyone does allocate temps though the MicroVU allocator, not that those should persist across instructions.

So this should be fine, although I'd remove the comment about not used/unassigned - that's not the case.

refractionpcsx2 (approved):

PR #7771, opened 2023-01-02
CMake: Post-WX cleanup for Linux

stenzek:
### Description of Changes

 - Remove unused build options

 - Disable setcap by default
    Applications should not need to call sudo as part of the build process.

 - Rename XDG_STD to USE_LEGACY_USER_DIRECTORY
    By default, we use ~/.config/PCSX2 now.

 - Default Wayland support to on
    I don't think there's any systems worth supporting that don't have it.

 - Rework "install" logic
    Linux no longer installs to GIT_ROOT/bin, it builds direct to BUILD/bin.
    Saves a file copy, and running make install before running your developer
    build was always silly.

 - Don't require install target to build appimage
    AppImage creator now just adds the entire bin directory as-is.
    Everything needed is in there.

### Rationale behind Changes

Since there's still no Linux maintainer, this stuff annoyed me every time I had to debug an issue on Linux. Plus getting rid of leftover wx junk.

### Suggested Testing Steps

Make sure appimages still work.


stenzek:
Side note: We'll have to update the build instructions on here. Most of it can go, it's really just a matter of doing
```
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release .. # add -DCMAKE_INTERPROCEDURAL_OPTIMIZATION and clang toolchain for extra vroom vroom
make -jN
./bin/pcsx2-qt
```

arcum42:
Linux needs a .desktop file to actually add pcsx2 to the list of installed applications, and the xpm file is the icon it's using. Deleting them is going to cause issues.

ionenwks:
> Linux needs a .desktop file to actually add pcsx2 to the list of installed applications, and the xpm file is the icon it's using. Deleting them is going to cause issues.

Currently there's nothing processing the desktop.in anyway, on Gentoo we just create a generic .desktop file and use `bin/resources/icons/AppIconLarge.png` which is a higher quality icon anyway

stenzek:
> Linux needs a .desktop file to actually add pcsx2 to the list of installed applications, and the xpm file is the icon it's using. Deleting them is going to cause issues.

No, the appimage has its own: https://github.com/PCSX2/pcsx2/blob/master/.github/workflows/scripts/linux/pcsx2-qt.desktop

stenzek:
> Edit: having better means to handle system installs with its own .desktop be nice (aka return of PACKAGE_MODE), but the current linux_various I find it virtually useless, outdated desktop.in using useless env call and missing categories by default, outdated icon

PCSX2-qt is not designed to be installed system wide, `PACKAGE_MODE` is not coming back unless someone wants to take over maintenance for the platform. And I actually mean maintaining it, fixing bugs when they're reported, etc - something which I've had to do for the last few months when I don't even use Linux on the desktop.

ionenwks:
It's ok, as said it'd be "nice" but I don't find it overly needed. I just do a sed so AppRoot know where to find resources in /usr/share, toss the executable in /usr/bin and it works fine.

stenzek:
> It's ok, as said it'd be "nice" but I don't find it overly needed. I just do a sed so AppRoot know where to find resources in /usr/share, toss the executable in /usr/bin and it works fine.

You're free to do that, but keep in mind that your users will receive **no** support from us for modified builds, and get told to download the appimage because their package is broken. We've had so many lately, especially with the crappak, complaining about broken stuff, and it's very tiring.

There's good reason we don't allow system-wide installation - the resources are tied to the binary version, so you can't have multiple versions installed, bisect any issues, try any PR builds, etc, when you're doing a stupid split directory setup. There's really no reason you couldn't just put everything in e.g. `/opt/pcsx2-version`.

ionenwks:
The only alternative is to not ship the package at all, we don't give appimages or even prebuilts in general given Gentoo is a source based distro.

ionenwks:
And /opt is only used as a last resort on Gentoo.

stenzek:
With all respect, that's not our problem. If you don't like the way we designed things, like I said, step up and maintain PCSX2 for Linux (I really don't want to do keep doing it, but nobody else does), or fork and modify. But keep in mind the implications of doing so.

If you want something more concrete, where users won't be expected to try different versions when investigating issues, I'd suggest sticking to the stable builds. The nightly model is not suited to system packaging, it's not like you see Debian/Ubuntu/etc packaging nightly builds either - they stick to proper releases.

ionenwks:
That's fine? We've worked around it and it works fine. Like I said better support for this layout be "nice" but that's about it.

ionenwks:
Also, we do have the stable 1.6.0 build in stable Gentoo. The nightly or latest "live" is a opt-in thing.

Mrlinkwii (approved):

stenzek:
> That's fine? We've worked around it and it works fine.

Doesn't matter. Our stance is we don't support anything which is modified in any way.

Like I said, you're free to use what you like. But you'll need to handle support for your users, if they come to us, like they seem to do instead of contacting their packager, they'll get told to use the non-broken appimage.

kenshen112:
I'm willing to maintain it and tackle most issues if that'll help :wave:

refractionpcsx2:
Are you serious beard, I told you NO literally 5 minutes before you posted.

Stop trying to get around me or I will remove you completely.

stenzek:
And as I've said before I'm fine with that, if you can show you are willing to do anything outside adding stuff for your package first.

PR #7774, opened 2023-01-02
GS: Add GPU Target CLUT

stenzek:
### Description of Changes

An extension to CPU CLUT render, except it's all done on the GPU. So no readbacks or slowdown.

Still has a few things to do:

- [x] Properly handle CSA.
- [x] Fix in OpenGL.
- [x] Test more cases.
- [x] Don't require CPU CLUT enabled for preload heuristics.

### Rationale behind Changes

Readbacks are slow, even without EE<->GS thread sync.

### Suggested Testing Steps

Test games which currently need CPU CLUT, see if any can be changed to GPU.
Test Burnout series and NFS Most Wanted, I switched them to GPU.


DonelBueno:
Test Burnout games here and it works very well, big speed boost when the sun is on the screen, great work!

JordanTheToaster (approved):

PR #7776, opened 2023-01-02
build: Add cmake presets to pcsx2.

arcum42:
### Description of Changes
This adds a number of cmake presets to the project to make compiling easier. These can be used when compiling on the command line, and Visual Studio Code has built in support for them.

Just to note, the existing ways of building still work. This is just an alternative.

Compiling with the clang-devel preset, for example, would be a matter of typing:
```
mkdir build
cd build
cmake --preset clang-devel ..
make
```

Alternatively, if you don't like command lines:
In Visual Studio Code, open the pcsx2 folder. It'll automatically have you choose a preset, create the build folder for you, run the cmake command, and if you hit play at the bottom, it'll compile and open pcsx2.

You can also add your own presets beyond the ones I've added by adding a CMakeUserPresets.json file.

Presets I've added:
"ninja-multi" - Generate multiple ninja build files.

"gcc-debug" - Debug build using make and the default compiler.
"gcc-devel" - Devel build using make and the default compiler.
"gcc-release" - Release build using make and the default compiler.

"gcc-debug-ninja" - Debug build using ninja and the default compiler.
"gcc-devel-ninja" - Devel build using ninja and the default compiler.
"gcc-release-ninja" - Release build using ninja and the default compiler.

"clang-debug"  - Debug build using ninja and clang.
"clang-devel" - Devel build using ninja and clang.
"clang-release" - Clang LTO Release build using ninja. 

The main one that likely needs explanation is ninja-multi. If you use that build, it generates build-*.ninja files for all configurations. You can specify a build file with -f (for example: ninja -f build-Debug.ninja).

### Rationale behind Changes
It's less typing to use a preset, and the Visual Studio Code integration's fairly nice.

Later on, we could build on this. For example, we could move all the warning flags over to presets, and it *is* possible to have windows and mac presets as well. (In fact, I think Microsoft was behind adding this to cmake.)

### Suggested Testing Steps
Build pcsx2 using presets.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -29,16 +29,6 @@
             }
         },
         {
-            "name": "lto",
-            "displayName": "Default Profile using clang/ninja & lto.",
-            "description": "Release lto build using ninja & clang.",
-            "inherits": "clang-base",
-            "cacheVariables": {
-                "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "ON",
-                "CMAKE_BUILD_TYPE": "Release"
-            }
-        },
-        {
             "name": "ninja-multi",
             "displayName": "Ninja Multi Config",
             "description": "Generate multiple ninja build files.",
@@ -126,9 +116,10 @@
         {
             "name": "clang-release",
             "displayName": "Clang Release",
-            "description": "Release build using ninja & clang.",
+            "description": "Release lto build using ninja & clang.",
             "inherits": "clang-base",
             "cacheVariables": {
+                "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "ON",
                 "CMAKE_BUILD_TYPE": "Release"
             }
         }
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -7,85 +7,99 @@
     },
     "configurePresets": [
         {
-            "name": "defaults",
-            "displayName": "Defaults",
-            "description": "Base preset. Only for inheriting from.",
+            "name": "gcc-base",
+            "displayName": "GCC Base",
+            "description": "Base preset for GCC. Only for inheriting from.",
             "hidden": true,
             "binaryDir": "${sourceDir}/build"
         },
         {
-            "name": "clang",
-            "displayName": "Clang",
-            "description": "Base clang preset. Only for inheriting from.",
+            "name": "clang-base",
+            "displayName": "Base",
+            "description": "Base preset for Clang. Only for inheriting from.",
             "hidden": true,
-            "inherits": "defaults",
+            "binaryDir": "${sourceDir}/build",
+            "generator": "Ninja",
             "cacheVariables": {
+                "CMAKE_EXE_LINKER_FLAGS_INIT": "-fuse-ld=lld",
+                "CMAKE_MODULE_LINKER_FLAGS_INIT": "-fuse-ld=lld",
+                "CMAKE_SHARED_LINKER_FLAGS_INIT": "-fuse-ld=lld",
                 "CMAKE_C_COMPILER": "clang",
                 "CMAKE_CXX_COMPILER": "clang++"
+            }
+        },
+        {
+            "name": "lto",
+            "displayName": "Default Profile using clang/ninja & lto.",
+            "description": "Release lto build using ninja & clang.",
+            "inherits": "clang-base",
+            "cacheVariables": {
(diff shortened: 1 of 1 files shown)


CMakePresets.json
```diff
@@ -0,0 +1,125 @@
...
+            }
+        },
+        {
+            "name": "clang-release",
+            "displayName": "Clang Release",
+            "description": "Release build using ninja & clang.",
+            "generator": "Ninja",
+            "inherits": "clang",
+            "cacheVariables": {
+                "CMAKE_BUILD_TYPE": "Release"
```
stenzek, line 121: It's worthwhile enabling LTO here , which is usually `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON`. Although ideally you'd want to use lld as well (`-fuse-ld=lld`), the link times with the gcc linker are pretty terrible.

arcum42: All doable. The questions would be whether I should enable LTO on all the presets, just some, or add new presets for it. Largely just if none of the presets don't have LTO, not using it will be less tested.

With ld, the main thing would be that I don't want to assume people have it installed (which is really the only reason I included non-ninja builds). It is possible to do conditionals, but I'd need something I could check (a cmake variable or environment variable) to see if it's installed, though I suppose I can assume it's installed on the clang builds.

I'd rather do conditionals on ninja, too, really, but I was having trouble finding something I could use, and I figured it was something I could refine later...

arcum42:
I changed the preset names up to give less preference to gcc over clang, and added a "lto" profile, which does a clang lto release build with ninja and lld. (I will note that in my testing, the clang release preset had the fastest compile time, and the lto profile has the slowest, and adding lto is the only difference... )

stenzek:
Right, lto will take longer, but produce faster code.

IMO it shouldn't be separate - Release is LTO'ed in our VS solutions, if you want faster iteration, we usually use devel builds.

arcum42:
I could always bundle it into the release build. I'll admit I usually build development builds anyways. I only added lto to clang, incidentally, as the times I tried with gcc, the build failed.

It also seems like we are really supposed to be doing some checks in cmake before turning it on, like this:

```
include(CheckIPOSupported)
check_ipo_supported(RESULT supported OUTPUT error)
```

But that's more something I might want to look into and do in a separate pr.

arcum42:
Alright, there we go, clang-release is always lto.

arcum42:
Also just wrote myself a very short CMakeUserPresets.json file, but that file is for personal use:
```
{
    "version": 6,
    "cmakeMinimumRequired": {
        "major": 3,
        "minor": 19,
        "patch": 0
    },
    "configurePresets": [
        {
            "name": "gcc-dev",
            "displayName": "GCC DEV",
            "description": "Development build using ninja and gcc.",
            "inherits": "gcc-devel-ninja"
        },
        {
            "name": "dev",
            "displayName": "DEV",
            "description": "Development build using ninja and clang.",
            "inherits": "clang-devel"
        }
    ]
}
```

stenzek:
> It also seems like we are really supposed to be doing some checks in cmake before turning it on, like this:

Given all our target platforms support it, it's probably fine.. otherwise if you *must* call `CheckIPOSupported` first, then it would be impossible to specify on the command line :/

arcum42:
Oh, you could still specify it on the command line. You'd just put the relevant set_target_properties command in an if statement, only executing it if the test passed.

(Though I mostly just care about getting the cmake warnings to go away, honestly. :P )

stenzek:
It only warns on some of the subprojects, presumably because they're targeting older cmake versions (and because they're submodules, we can't change them :().

However, forcing the policy in the main CMake file seems to do the trick:

```cmake
cmake_policy(SET CMP0069 NEW)
set(CMAKE_POLICY_DEFAULT_CMP0069 NEW)
```

arcum42:
Yeah, that's what always sucks about submodules. I'd love to turn off all warnings from them, really, but never got too far.

I'm good with the policy commands, though I suppose that probably should be a separate pr. Seems to work on my side, though.

I have mixed feelings on it, but one idea that does occur to me is that we could move the compiler warnings to a specific preset and take them out of the main cmake file...

arcum42:
Actually, I bet some of the warnings we are turning off don't need to be any more with wx gone.

arcum42:
Unless anyone has any other changes they want to make, I think this is ready to commit, btw. I'll do a separate pr for anything about warning profiles or getting rid of the lto cmake warnings.

PR #7777, opened 2023-01-03
Debugger Breakpoints: (Refactor) Use some reflection and item models

F0bes:
### Description of Changes
Instead of using a QTableWidget, use a QTableView and **cleanly** represent breakpoint / memcheck objects using an Abstract item model.

Also get fancy and use some form of reflection for the BreakPoint and MemCheck objects.

### Rationale behind Changes
I'm a lot happier with how this looks, code is more organized and probably faster.

### Suggested Testing Steps
Mess around with breakpoints. Make sure nothing crashes, corrupts, or summons any unwanted demons.


pcsx2-qt/Debugger/Models/BreakpointModel.cpp
```diff
@@ -0,0 +1,314 @@
...
+		}
+		else if (const auto* mc = std::get_if<MemCheck>(&bp_mc))
+		{
+			switch (index.column())
+			{
+				case BreakpointColumns::TYPE:
+				{
+					QString type("");
+					type += mc->cond & MEMCHECK_READ ? tr("Read") : "";
+					type += ((mc->cond & MEMCHECK_BOTH) == MEMCHECK_BOTH) ? ", " : " ";
```
lightningterror, line 81: Some codacy complaints to take care of, below as well.

F0bes: Resolved.

Mrlinkwii (approved):

Issue #2371, opened 2018-04-06
Meta: Games that require an alternate SPU2 DelayCycles value


MrCK1:
This issue is a list of games that require a modified DelayCycles value in order for sound to work properly. It will be a good reference for the future.

MK:DA and FFBA have been broken since this[ build](https://github.com/PCSX2/pcsx2/commit/f1cffa6f7fd8e85a2d2df35b7c18ca2e4d3606ab#diff-880fcba486dd038bdc85976ae1e9297a), the others here may be the same.

- [ ] Armored Core - https://github.com/PCSX2/pcsx2/issues/2917
- [ ] Fatal Fury Battle Archives Vol. 2 (US:NTSC) - (DelayCycles 0/1)
- [ ] Mortal Kombat: Deadly Alliance (PAL) - not always necessary (DelayCycles 0/1)
- [ ] Unreal Tournament (NTSC-U) - (works with DelayCycles 3 or lower)
- [ ] Higurashi no Naku Koro ni Matsuri Kakera Asobi (#918)

Unconfirmed:
- [ ] Spiderman 2
- [ ] Deus Ex: The Conspiracy (#615)

seta-san:
This issue belongs in here too
https://github.com/PCSX2/pcsx2/issues/268

refractionpcsx2:
Should be all fixed in #3959

PR #7789, opened 2023-01-03
SPU2: Allow loop rewrite of LOOPSTART if block header already passed

### LINKED ISSUE #5991: [BUG]: Pro Biker 2 - Main/Pause Menu Sound Effects Looping Endlessly even when Audio and Music in-game is turned down
### Describe the Bug

Some days ago while testing I found an odd audio bug that makes the game looping the Main and Pause Menus sound effects endlessly when you're either at the Game Mode Main Menu or playing any race, It doesn't stop even If you turn down the Audio and Music Volume in-game to the lowest, tried all available renders and even Software Mode but none of them made a difference at all, also tried to change Audio Synchronizing Mode to "Async Mix" as suggested but the audio bug persisted nonetheless.

### Reproduction Steps

Boot the Game either using Full or Fast Modes and then browse around the Game's Main Menu by going up and down with the Left Stick or D-Pad and you'll eventually notice a lot of sound effects looping pretty loud, like in some kind of fast mode.

### Expected Behavior

Endlessly Loop of Sound Effects from Main and Pause Menus.

### PCSX2 Revision

v1.7.2675

### Operating System

Windows 10 (64bit)

### If Linux - Specify Distro

_No response_

### CPU

i5-9400f

### GPU

GTX 1070

### GS Settings

Default

### Emulation Settings

Default

### GS Window Screenshots

A video which clearly exhibits the bug audio

(https://youtu.be/Oz4_aU89zCI)

### Logs & Dumps

_No response_

refractionpcsx2:
### Description of Changes
Allows writing of loops start if voice processing has already written it.

### Rationale behind Changes
Games do sucky things they shouldn't, waveform writing in the first 4T's has priority over register writes, but if this has already passed then the register write can take effect. Regression from #4186 

### Suggested Testing Steps
Retest stuff on #2371 (already tested Mafia, Megaman X7, Unreal Tournament) and White Van Racer if you're insane enough to own that game.

Fixes White Van Racer
Fixes #5991 (Pro Biker 2)

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -19801,6 +19801,8 @@ SLES-54549:
 SLES-54550:
   name: "White Van Racer"
   region: "PAL-E"
+  gameFixes:
+    - InstantDMAHack # Fixes DMA timeout errors.
 SLES-54551:
   name: "Premier Manager 2006-2007"
   region: "PAL-G"
--- a/pcsx2/SPU2/spu2sys.cpp
+++ b/pcsx2/SPU2/spu2sys.cpp
@@ -1208,7 +1208,7 @@ static void RegWrite_VoiceAddr(u16 value)
 		case 2:
 			{
 				u32* LoopReg;
-				if ((Cycles - thisvoice.PlayCycle) < 4)
+				if ((Cycles - thisvoice.PlayCycle) < 4 && (int)(thisvoice.LoopCycle - thisvoice.PlayCycle) < 0)
 				{
 					LoopReg = &thisvoice.PendingLoopStartA;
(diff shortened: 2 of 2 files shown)


Silanda:
MK Deadly Alliance seems fine. 

It, and all the games in the series, still had problems with occasional sounds being dropped, especially if sounds were played one after another rapidly (spamming directions in the menu, for example). It's a really minor problem and this commit doesn't make it worse. Might even be a bit better, but that's probably a placebo.

refractionpcsx2:
Thanks for testing :)

Mrlinkwii (approved):

JordanTheToaster (approved):

PR #7794, opened 2023-01-05
Build/Qt/GS: Unmute some warnings, & resolve Qt/misc warnings.

arcum42:
### Description of Changes
-Wno-overloaded-virtual & -Wno-deprecated-declarations were there to mute out warnings coming from wx and gtk 2 code, so I removed them.
This lot to a bunch of deprecated Qt warnings that had been muted showing up. Resolved those warnings.
Added -Wno-maybe-uninitialized, as the GS vector code in gcc spams a bunch of warnings (about member m in GSVector8i, IIRC), and I wasn't sure how to resolve them otherwise.
Added cmake policy code for CMP0069 that stenzek suggested earlier.

(There are still some warnings, but a good deal less, and one already has another pr open.)

### Rationale behind Changes
It's generally better to fix warnings, when possible. Deprecated api calls will eventually be removed, so it's best to take care of them. (With wxwidgets, it was using ancient gtk code in the library itself, so there was no way to get rid of the warnings...)

### Suggested Testing Steps
Trigger the code I changed/rewrote, and see if any problems exist. There was a change disc dialog that really needs testing, mouse input in the controller dialog, and some list updating in the usb section of the settings.

I'm setting this to a draft until I'm sure there are no issues in those areas.

### DIFF
--- a/pcsx2-qt/MainWindow.cpp
+++ b/pcsx2-qt/MainWindow.cpp
@@ -521,7 +521,7 @@ void MainWindow::setStyleFromSettings()
 		qApp->setStyleSheet(QString());
 		qApp->setStyle(QStyleFactory::create("Fusion"));
 	}
-else if (theme == "darkfusion")
+	else if (theme == "darkfusion")
 	{
 		// adapted from https://gist.github.com/QuantumCD/6245215
 		qApp->setStyle(QStyleFactory::create("Fusion"));
@@ -742,7 +742,7 @@ else if (theme == "darkfusion")
 		// Custom palette by RedDevilus, Blue as main color and Purple as complimentary.
 		// Alternative dark theme.
 		qApp->setStyle(QStyleFactory::create("Fusion"));
-		
+
 		const QColor blackishblue(50, 25, 70);
 		const QColor darkerPurple(90, 30, 105);
 		const QColor nauticalPurple(110, 30, 125);
@@ -1640,7 +1640,7 @@ void MainWindow::onViewGamePropertiesActionTriggered()
 		auto lock = GameList::GetLock();
 		const GameList::Entry* entry = m_current_elf_override.isEmpty() ?
 										   GameList::GetEntryForPath(m_current_disc_path.toUtf8().constData()) :
-                                           GameList::GetEntryForPath(m_current_elf_override.toUtf8().constData());
+										   GameList::GetEntryForPath(m_current_elf_override.toUtf8().constData());
 		if (entry)
 		{
 			SettingsDialog::openGamePropertiesDialog(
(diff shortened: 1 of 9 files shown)


pcsx2-qt/MainWindow.cpp
```diff
@@ -2867,11 +2867,16 @@ void MainWindow::doDiscChange(CDVD_SourceType source, const QString& path)
 	bool reset_system = false;
 	if (!m_was_disc_change_request)
 	{
-		const int choice = QMessageBox::question(this, tr("Confirm Disc Change"),
-			tr("Do you want to swap discs or boot the new image (via system reset)?"), tr("Swap Disc"), tr("Reset"), tr("Cancel"), 0, 2);
-		if (choice == 2)
-			return;
-		reset_system = (choice != 0);
+		QMessageBox messageBox(QMessageBox::Question, tr("Confirm Disc Change"),
```
stenzek, line 2870: Nitpick: follow the same naming convention as the rest of the function (`snake_case`), using `camelCase` here is just inconsistent.

pcsx2-qt/MainWindow.cpp
```diff
@@ -2867,11 +2867,16 @@ void MainWindow::doDiscChange(CDVD_SourceType source, const QString& path)
...
+			tr("Do you want to swap discs or boot the new image (via system reset)?"));
+		messageBox.addButton(tr("Swap Disc"), QMessageBox::ActionRole);
+		QPushButton *resetButton = messageBox.addButton(tr("Reset"), QMessageBox::ActionRole);
+		QPushButton *cancelButton = messageBox.addButton(QMessageBox::Cancel);
+		messageBox.setDefaultButton(cancelButton);
+		messageBox.exec();
+
+		if (messageBox.clickedButton() == cancelButton) return;
```
stenzek, line 2878: `return` on newline (or just clang-format the function).

stenzek (commented):
Rest seems okay, although probably should squash the commits.

pcsx2-qt/MainWindow.cpp
arcum42: Yeah, I was looking at some Qt documentation as I was writing this, and automatically used the same naming format (and one of the names) as the examples. I can fix that. The irritating thing is that I think the version they deprecated was more straightforward...

pcsx2-qt/MainWindow.cpp
arcum42: Yeah, my normal code formatting was creeping in, and I've had times before where clang format has changed formatting throughout the file I was working on when I only wanted to correct the formatting I was working on. Since this is a new file, that should be safe, though...

stenzek: Yeah, I usually format the whole file for everything in the Qt subdirectory before committing.

arcum42:
Changed up the formatting and variable names. Seems like the formatting did affect a line or two that were out of scope, but someone else has probably modified this code...

stenzek:
That's fine.

arcum42:
Rebased to take care of a conflict with lightningterror's warning fixes. Squashed to one commit.

Mrlinkwii (approved):

arcum42:
Turns out that warning I muted is gcc only, so I moved it to the gcc only section...

lightningterror:
Is this good to go for merge?

arcum42:
Unless anyone's run into any issues with it, it should be fine for merging.

PR #7801, opened 2023-01-05
common: Remove pxUSE_SECURE_MALLOC.

arcum42:
### Description of Changes
Removes the define pxUSE_SECURE_MALLOC, and any code that checks for it.

### Rationale behind Changes
If you try and enable that code, it doesn't compile, as the function being called doesn't exist. Less dead code is better.

### Suggested Testing Steps
Compile!

### DIFF
--- a/common/AlignedMalloc.h
+++ b/common/AlignedMalloc.h
@@ -27,12 +27,6 @@
 #include <malloc.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


Mrlinkwii (approved):

arcum42:
For anyone curious, IndexBoundsAssumeDev turns out to be a define that used to be in Assertions.h, but got deleted back in May.

Issue #7496, opened 2022-11-28
[BUG]: Syphon Filter: The Omega Strain (PAL Version) Frame Rate/Abnormal Speed Issue


LeyoSFO:
### Describe the Bug

The PAL version of the game "Syphon Filter: The Omega Strain" has a frame rate issue, it runs inexplicably fast, faster on PCSX2 than on PS2, and it's not just the frame numbers, the gameplay itself is affected and it is abnormal and certain feats can't be performed because of it., a more accurate description would be that the gameplay movement is "abnormally smooth".

This bug happens randomly upon restarting the emulator, whenever PCSX2 is restarted then it's a game of chance, the game would be either running at normal PAL speed (25 FPS) or abnormally smooth/fast (50 FPS), every time someone restarts, it "might" change between these2 states. The VPS is not affected by this, its value remains 50 in both states.

The NTSC-U & NTSC-K versions of the same game are not affected and work normally, also the NTSC-U/K versions have a fixed frame rate of 30 FPS, which means the PAL version should be 25 FPS, and on PS2 it actually is 25 FPS.

### Reproduction Steps

-Performed numerous restarts to confirm the issue, and it is confirmed, each time i restart, the game speed and frames "might" change or remain the same, and "when" it changes, it's either normal PAL speed (25 FPS) or abnormally smooth/fast (50 FPS)
-No emulator settings were changed between restarts
-PAL games have 2 variations, 25 FPS games & 50 FPS games, i tested a bunch of other PAL games and the 50 ones remained at 50, and the 25 ones remained at 25 after each restart, they don't alternate, so this issue is exclusive to this game only as far as i know
-I tested the same game on an actual PS2 and the issue didn't happen, it ran at normal 25 FPS PAL speed, so the issue is exclusive to this game only, and it happens on the emulator only, and it happens only on the PAL version of this game
-Tested the game on PCSX2 1.5.5 & 1.6 & 1.7, the issue happened in all versions of the emulator
-The testings were done on multiple PCs to make sure it is not a hardware issue

### Expected Behavior

The NTSC version of this game runs at 30 FPS on PS2 and on PCSX2, and the PAL version of this game runs at 25 FPS on PS2, so i "expected" the PAL version to run at 25 FPS on PCSX2 as well at all times, and restarting the emulator should NOT affect the frame rate or game speed.

### PCSX2 Revision

v1.7.3651

### Operating System

Windows 10 (64bit)

### CPU

i5-7300HQ

### GPU

[...]

PR #7806, opened 2023-01-06
GameDB: Syphon Filter: The Omega Strain PAL speed fix patch

LeyoSFO:
### Description of Changes
Note: All the FPS that is gonna be mentioned below is meant to be the in-game/in-level FPS, not the menus nor the cutscenes,

The PAL version of Syphon Filter: The Omega Strain runs at 25 FPS on PS2, but on PCSX2 1.6, 1.7, ..etc (all versions) it runs mostly at 50 FPS and it's abnormally fast which is changing the whole game speed, and on PCSX2 you can reproduce the issue by rebooting the emulator, each time you reboot you might get the game at 25 FPS or at 50 FPS (game of luck) but mostly 50 and sometimes 25, also the NTSC-U and NTSC-K versions of the game run at 30 FPS so the PAL can't be this much faster so the correct speed by A) conclusion and B) testing on a real PS2 should be 25 FPS, this is a patch to fix the issue and make it run always at 25 FPS.

I posted a github issue about this going into more detail and i'll link it here #7496

### Rationale behind Changes
Fixes the game speed and makes it run just like it originally did on PS2.

### Suggested Testing Steps
To reproduce the problem and check the effectiveness of the patch:
-Boot the PAL version of Syphon Filter: The Omega Strain.
-Keep rebooting PCSX2 1.7 you'll notice sometimes the PAL version runs at 50 FPS (abnormal) and sometimes at 25 FPS (correct), but mostly it's 50 FPS so keep rebooting until you get 25 FPS without the patch.
-Compare the game speed on the 25 FPS session (without the patch) to the speed of the game when the submitted 25 FPS is applied, if they're similar then that means the patch is doing its job correctly and is not changing anything in the game incorrectly, all it's doing is restoring its correct speed.
-This is not like applying a 60 FPS patch, it's not an add-on it's a fix, the game inherently runs wrong on the emulator only (it runs correct on PS2) so this is a necessary fix.

### DIFF
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -3470,7 +3470,7 @@ SCES-52033:
         patch=0,EE,003953F8,word,48438000
         patch=0,EE,003735FC,word,4B06521B
         author=YukiXXL
-        //Speed Correction (25 FPS)
+        // Speed Correction (25 FPS)
         patch=1,EE,00175E1C,extended,24020019
 SCES-52042:
   name: "Formula One 2004"
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -3471,7 +3471,7 @@ SCES-52033:
         patch=0,EE,003735FC,word,4B06521B
         author=YukiXXL
         // Speed Correction (25 FPS)
-        patch=1,EE,00175E1C,extended,24020019
+        patch=1,EE,00175E1C,extended,00000019
 SCES-52042:
   name: "Formula One 2004"
   region: "PAL-M6"
--- a/bin/resources/GameIndex.yaml
+++ b/bin/resources/GameIndex.yaml
@@ -3469,6 +3469,9 @@ SCES-52033:
         // Cop2 problems.
         patch=0,EE,003953F8,word,48438000
         patch=0,EE,003735FC,word,4B06521B
+        author=YukiXXL
+        //Speed Correction (25 FPS)
+        patch=1,EE,00175E1C,extended,24020019
 SCES-52042:
   name: "Formula One 2004"
   region: "PAL-M6"


refractionpcsx2:
I have to admit, it seems strange that the game has code to let it run at 50fps, but completely breaks the game in the process... That seems very odd.

Leyo1:
> 

My theory is that the game has 50fps code, but it also has a limiter that limits it to 25 because it was designed to run at 25fps on PS2 but on PCSX2 the limiter is being broken/ignored somehow, point is our reference point is PS2 which is 25, and even on PCSX2 on some reboots you would get the 25fps session sometimes.

The NTSC-U & NTSC-K versions of the same game also have 60fps codes in them but they run at 30fps constantly on PS2 and on PCSX2, they don't break the limiter so all evidence points to 25fps being the correct speed on the PAL version.

bin/resources/GameIndex.yaml
```diff
@@ -3469,6 +3469,9 @@ SCES-52033:
         // Cop2 problems.
         patch=0,EE,003953F8,word,48438000
         patch=0,EE,003735FC,word,4B06521B
+        author=YukiXXL
+        //Speed Correction (25 FPS)
```
refractionpcsx2, line 3473: can you put a space after the // to match the formatting of all other patches? Thanks.
Then we'll be GTG

bin/resources/GameIndex.yaml
```diff
@@ -3469,6 +3469,9 @@ SCES-52033:
...
+        patch=1,EE,00175E1C,extended,24020019
```
refractionpcsx2, line 3474: This patch is only going to modify the first byte, because in extended mode the first nibble is the size. 0 = 8bit 1 = 16bit 2 = 32bit, so only the 0x19 on the end will get written to memory, was that intended?

bin/resources/GameIndex.yaml
LeyoSFO: ok i changed it now, but please check if i actually did it correctly

refractionpcsx2: That's fine, but it was my lesser worry about the PR. If you could correct the other thing and make 100% sure the patch works as intended, if it's working now, I fear it might be placebo.

bin/resources/GameIndex.yaml
refractionpcsx2: alright there was no need to get defensive, my comment was related to the potential that the patch wasn't working correctly, but you explained that it was.

refractionpcsx2: And yes, I want you to change it for clarity's sake, please. The final effect will be identical, yes.

LeyoSFO: > alright there was no need to get defensive, my comment was related to the potential that the patch wasn't working correctly, but you explained that it was.

no no lol i am not defensive at all, i'm just explaining it in more detail because i am afraid to have missed something or that my testing method could be wrong? so i wanted to lay it all out to you so you can confirm that i'm onto something real here, sorry for the confusion, thanks

refractionpcsx2: It sounds fine to me, I don't really know the game so I can't corroborate your findings and results, tbh the game felt fine to me at 50fps, but I don't really know it, as I say.

LeyoSFO: > And yes, I want you to change it for clarity's sake, please. The final effect will be identical, yes.

done

PR #7849, opened 2023-01-09
VMManager: Increase software thread count on multicore (4+) CPUs

RedDevilus:
### Description of Changes

3 is a nice compromise if you have 4 core or more but for some games higher numbers like 8 or even higher would work, however for a lot of games it would perhaps even reduce the performance compared to 3 software rendering threads or in more likelihood have diminishing returns.
### Rationale behind Changes

MTVU was already auto-detected to be enabled if you have 4+ cores (minus the automatic fixes for specific games that dislike MTVU) which makes it better for the user experience not giving a single care and having a better experience than even 1.6 where you needed to enable it by hand or by preset 3. Now this will go hand in hand with MTVU and not leave free performance on the table where possible. Keep in mind if you have 2/4 CPUs like 2 cores with multithreading this won't be activated but might or not help if you do it by hand (too many variables to enable for those type of users and could have a negative impact).
### Suggested Testing Steps

Open up PCSX2 with this Pull Request, make sure you have no inis or old settings and check the Software Rendering Threads:

New Default (if enough cores):
image

Old default (or not enough real cores):
image

### DIFF
--- a/pcsx2/VMManager.cpp
+++ b/pcsx2/VMManager.cpp
@@ -2120,14 +2120,18 @@ static void SetMTVUAndAffinityControlDefault(SettingsInterface& si)
 
 	if (big_cores >= 3)
 	{
-		Console.WriteLn("  So enabling MTVU.");
+		Console.WriteLn("  Enabling MTVU.");
 		si.SetBoolValue("EmuCore/Speedhacks", "vuThread", true);
 	}
 	else
 	{
-		Console.WriteLn("  So disabling MTVU.");
+		Console.WriteLn("  Disabling MTVU.");
 		si.SetBoolValue("EmuCore/Speedhacks", "vuThread", false);
 	}
+
+	const int extra_threads = (big_cores > 3) ? 3 : 2;
+	Console.WriteLn("  Setting Extra Software Rendering Threads to %d.", extra_threads);
+	si.SetIntValue("EmuCore/GS", "extrathreads", extra_threads);
 }
 
 #else


refractionpcsx2:
Get rid of the "So" stuff, because if both of them trigger it'll be like 

"Detecting big cores... So enabling MTVU. So enabling extra threads" which just sounds dumb, better just have
"Detecting big cores...  Enabling MTVU. Enabling extra render threads."

RedDevilus:
Didn't think combining would work, but sure.

F0bes:
Does that really need to be logged in the first place?

refractionpcsx2:
Also you don't want 3 sw threads on a 3 core CPU (I know it's rare...) so I would change that check to be >= 4

RedDevilus:
Test in VM:
2 cores with 2 with MT:
image

4 cores:
image

Seems to work fine.

refractionpcsx2:
a 2 core CPU Should have 0 extra threads.

RedDevilus:
Are there any SSE4 CPUs with 2 cores and no multithreading?

refractionpcsx2:
yup https://en.wikipedia.org/wiki/Penryn_(microprocessor)

TellowKrinkle:
> a 2 core CPU Should have 0 extra threads.

Have you tested this?

The only computer I have with 2c2t is a core duo (SSE2, internally splits SSE instructions into two 64-bit uops), and it tends to run games best with 2 SW renderer threads.  But that might be because it runs the SW renderer so slowly that the CPU and VU threads end up doing almost nothing while they wait for the SW renderer, which might not be the case for a computer with less garbage SSE.

refractionpcsx2:
> > a 2 core CPU Should have 0 extra threads.
> 
> Have you tested this?
> 
> The only computer I have with 2c2t is a core duo (SSE2, internally splits SSE instructions into two 64-bit uops), and it tends to run games best with 2 SW renderer threads. But that might be because it runs the SW renderer so slowly that the CPU and VU threads end up doing almost nothing while they wait for the SW renderer, which might not be the case for a computer with less garbage SSE.

I've later updated my suggested to 1 extra thread for 2 cores.

TellowKrinkle:
I can test on a 2c4t later, but I can almost guarantee that it would prefer 2 threads over 1, which wouldn't match `big_cores - 1`

refractionpcsx2:
alright, then the code will have to be more complicated for those scenarios, which seems a bit OTT to me, but sure.. :P

TellowKrinkle:
`big_cores > 3 ? 3 : 2` doesn't seem especially complicated to me...

refractionpcsx2:
but what about actual dual core CPU's? surely this would be detrimental to their performance.

TellowKrinkle:
The only datapoint I have for 2c2t is a Core Duo T2400
Which preferred 2 threads in the games I tested (though it wasn't able to reach full speed in any of them)

RedDevilus:
For the new default value for software this is probably sane + safe enough, if there are other remarks or not necessary then it's probably fine. We can always review it later if it has a positive gain across the board for users.

And they can always change the value if need be.

Testing in VM
2c4t = 2 extra threads
3c6t = 3 extra threads
…

TransparentBlue:
From my own experience (mostly my own 3800X and helping someone with a 3600 at some point) having more threads than could fit on a single CCD would downgrade performance effectively making some of them behave closer to a 2c/4t or 3c/6t (and technically 1c/2t since there's a couple CPU with 1 core per CCD out there)

RedDevilus:
Atleast SSE2 CPUs won't work anymore, however if you have benchmarks for the lower cores like 2c4t, then sure I'll look at it. I did test it before and 2 was the default for everyone but just expands to 3 if you have enough. I was just too lazt to use CapFrameX.

RedDevilus:
I went ahead and did some higher values with 2ct4 such as 6-8 and they worked fine but that did reduce performance a bit, so I don't think CCD and CCX will matter that much in this case as it's been the default (2 extra threads on top) for easily above 10 years I would say that didn't have the IPC or core count.

The code will just make it 3 for capable PCs and are still in my opinion too safe for users which makes it about perfect for default value. I'll try add benchmarks when I am able to.

Mrlinkwii (approved):

eddlang:
I don't know if newer CPUs are related to the discussion, but Celerons are still 2c2t, and Pentiums up to Skylake.

refractionpcsx2:
We're aware, but due to how poor the performance possibly is, compared to more modern CPU's, it's probably not detrimental to performance.

Issue #2245, opened 2018-01-02
Meta: Games that don't like the Full fpu mode.


atomic83GitHub:
***What is this meta?***

The request was at the origin, to put the EE Clamping mode to full on the safest preset. 
![image](https://user-images.githubusercontent.com/26351275/34483586-c8ee5230-efbf-11e7-9260-ebe4b831e022.png)
But @ramapcsx2 told me that there are some games wich don't support the Full fpu mode so I made a meta to track these.

***Here is an exhaustive list of game affected by the Full fpu mode***

-Gran Turismo 4 (Unfortunately needed to load Demonstration sequences)

-Tourist Trophy (Unfortunately needed to load Demonstration sequences)

Feel free to contribute :).

ramapcsx2:
Hm, are you sure it's just those 2 games? I distinctly remember an about even ratio of breakage and fixes with full fpu mode.
When it was developed, there was only one "full fpu" mode but it soon turned out that even if it was correct, games hated it. That's when we adopted only parts of it and made that the new default.

Full mode is still a good thing, but we need to make sure it doesn't break more than it fixes..

atomic83GitHub:
I tested it in my set of games and none of them (apart GT4 and TR) have an issue with the full mode, maybe a sort of meta could be nice to see if there is some another games wich don't like this.

ramapcsx2:
I don't have the old notes anymore but some tests were really tricky.
One test is a puzzle in Tales of the Abyss, where you couldn't move a certain statue. Nothing short of a full playthrough is going to catch these :p

atomic83GitHub:
Ah ok :).
Anyway, I will still keep this open if there will be enough tests.

MrCK1:
This is one of those things that's hard to predict. I'd say that EE Clamping is at least less sensitive then say Negative VU Rounding which breaks a ton of games and makes Lara Croft fall through the floor (lol).

I'd say it's too late in PCSX2's lifespan to make a huge change like this, the possibility of introducing a regression is too big.

atomic83GitHub:
Ok so based on your opinion, I will close it, I never thought it can introduce some regression.
As a side note, Ratchet and Clank throw a black screen with VU rounding to negative, and if you have the first Ratchet and Clank, can you do a test for me?

Load the gemlick space station first and after, the Rilgar level, I suspect a Timing issue.

mirh:
[...]

PR #7854, opened 2023-01-10
EE: In Clamp=Full manually assemble PS -+FLT_MAX

### LINKED ISSUE #2316: Gran Turismo 3: A-spec - License test B-4 demo fails
***PCSX2 version:***
v1.5.0-dev-2286-g41c49faa8

***PCSX2 options:***
Defaults, no settings changes made a difference.  Tried presets 1 and 2 (Safest and Safe) in Emulation settings, also tried disabling speedhacks.

***Plugins used:***
Tested with various GSdx varieties: 

* GSdx 20180213122600 (MSVC 19.00 SSE4.1/AVX) 1.1.0 [GSdx32-SSE4]
* GSdx 0 (MSVC 18.00, SSE41) 0.1.16 [gsdx32-sse4-r5875]
* GSdx 20180213122600 (MSVC 19.00 SSE2/AVX) 1.1.0 [GSdx32-SSE2]
* GSdx 20180213122600 (MSVC 19.00 AVX2/AVX2) 1.1.0 [GSdx32-AVX2]

PAD: LilyPad (20180213122600) 0.12.1 [LilyPad]
SPU2: SPU2-X 20180213122600 2.0.0 [Spu2-X]
CDVD: cdvdGigaherz 20180213122600 0.11.0 [cdvdGigaherz]
USB: USBnull
FW: FWnull
DEV9: DEV9null

Default settings

During license test B-4, when running the demo, the car deviates from the line and goes off the track during the turn.  This causes the game to enter the license test fail-state and the car stops.  Also tested on actual hardware to confirm that it works fine on there.

Run the demo for license test B-4 (no save needed).

None that I know of, but never tested this before.

***PC specifications:***
CPU: Intel i7 5820k @stock
GPU: AMD Radeon HD 5770 1GB
RAM: 16GB DDR4
OS: Windows 10 64-bit

refractionpcsx2:
### Description of Changes
Modify the handling of overflowed values when EE clamp = full (aka double precision handling), manual assembly of overflow value instead of conversion.

### Rationale behind Changes
I believe the conversion was failing, causing bad clamped values (wrong sign).  This was causing Gran Turismo 4 and Tourist Trophy to crash when doing license tests (possibly GT 3 too).
### Suggested Testing Steps
test GT4 and other games with full clamping, make sure everything works as expected (or at least doesn't crash)
Check games listed in #2245

Might Fix #2316

### DIFF
--- a/pcsx2/x86/iFPUd.cpp
+++ b/pcsx2/x86/iFPUd.cpp
@@ -210,8 +210,12 @@ void ToPS2FPU_Full(int reg, bool flags, int absreg, bool acc, bool addsub)
 	u32* end2 = JMP32(0);
 
 	x86SetJ8(to_overflow);
-	xCVTSD2SS(xRegisterSSE(reg), xRegisterSSE(reg));
-	xOR.PS(xRegisterSSE(reg), ptr[&s_const.pos]); //clamp
+
+	// Move sign bit to bottom 32bit
+	xSHUF.PS(xRegisterSSE(reg), xRegisterSSE(reg), 0x55);
+	// Saturate value for PS FLT_MAX
+	xOR.PS(xRegisterSSE(reg), ptr[&s_const.pos]);
+
 	if (flags && FPU_FLAGS_OVERFLOW)
 		xOR(ptr32[&fpuRegs.fprc[31]], (FPUflagO | FPUflagSO));
 	if (flags && FPU_FLAGS_OVERFLOW && acc)


TellowKrinkle:
If input is positive:
1. Clamp: `0x7fffffffffffffff` (NaN)
2. Subtract exponent: `0x7fefffffffffffff` (+Very Large Number)
3. Convert to float: `0x7f800000` (+Inf)
4. Add exponent: `0x80000000` (-0)

If negative:
1. Clamp: 0xffffffffffffffff` (NaN)
2. Subtract exponent: `0xffefffffffffffff` (-Very Large Number)
3. Convert to float: `ff800000` (-Inf)
4. Add exponent: `0x00000000` (0)

While this might prevent crashes, I can guarantee that it's not correct.
If you really want to go with this, I'd recommend making it more obvious by doing `xPXOR(xRegisterSSE(reg), xRegisterSSE(reg))` instead.

refractionpcsx2:
Well these are the results of the arithmatic test, which the output matches the expected, so idk.... Makes it sound like it's working to me? But yes I suspect it wasn't correct.

so you want me to keep the old behaviour but have a special case for negative by clearing the register?

[arithmetic-expected.txt](https://github.com/PCSX2/pcsx2/files/10387533/arithmetic-expected.txt)
[arithmentic-pcsx2.txt](https://github.com/PCSX2/pcsx2/files/10387534/arithmentic-pcsx2.txt)

TellowKrinkle:
Do you have a run of the test before this commit so I know which values to be looking at?

refractionpcsx2:
tbh that test looked the same before too... which is weird...

TellowKrinkle:
So it's probably just not hitting that branch

refractionpcsx2:
I presumed it would be since it has a lot of FLT_MAX values

TellowKrinkle:
It does seem like it should, but if it doesn't change how those tests run but does change a game, my only conclusion is that those tests aren't testing whatever handling you're changing.

refractionpcsx2:
okay well I've put a replacement solution on discord, if you wanna check it.

refractionpcsx2:
Okay, better solution found, and updated :)

PR #7863, opened 2023-01-12
Misc: Latest round of warning cleanups

stenzek:
### Description of Changes

What the title says.

### Rationale behind Changes

Reducing warning spam. One of these is actually a legit error too (the EE threads one, the object it's referencing is destroyed at the end of the statement).

### Suggested Testing Steps

Make sure build passes.

### DIFF
--- a/common/emitter/cpudetect.cpp
+++ b/common/emitter/cpudetect.cpp
@@ -170,12 +170,6 @@ void x86capabilities::CountCores()
 {
 	Identify();
 
-	s32 regs[4];
-	u32 cmds;
-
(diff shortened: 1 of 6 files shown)


JoS19o30:
I would love for the next update to optimize all pro evolution soccer up to 14 because when a game ends, the fps always drop and also when scoring a goal or an animation arrives, the fps usually drop drastically

RedDevilus:
Please don't make irrelevant comments on Pull Requests. If you require support go to the PCSX2 discord channel.

Issue #7696, opened 2022-12-26
[Feature Request]: Scan Type


buruto:
### Description

Recent versions of PCSX2 have memory search feature including value type, which is currently called just and simply "Type", with options like 1 byte, 2 bytes, 4 bytes, 8 bytes, float, double and even string but unfortunately lacks scan type.

The Scan Type can be useful to compare current values with previous values by boolean operators like "Exact value", "Equal to", "Not equal to", "Less than", "Greater than", "Increased value", "Decreased value", "Value between" and even "Unknown initial value" like Cheat Engine.

Right now I think that the scan type is always hardcoded to "Exact value" and this cannot be changed through the graphical user interface of the "Memory Search" tab of PCSX2 debugger.

### Reason

I need the both "Unknown initial value" and "Decreased value" scan types for all games that render (draw on screen or window) the health of the character in a bar and not number so I don't know the initial value of the health of the character. After "Unknown initial value" I repeat "Decreased value" after I let the character lose health by getting hurt by the enemy until I eventually find the memory address of the health of the character.

Repeating "Decreased value" can also be useful to find the memory address of the money (currency) after buying anything from any shop which decreases this value though the "Exact value" scan type option is the fastest scan type option if the money or currency is rendered (drawn on screen or window) as a number in some screen.

### Examples

Games that render (draw on screen or window) the health of the character in a bar and not number are Devil May Cry, Final Fantasy X (including X-2 as well), God of War, Kingdom Hearts, Resident Evil 4 and etc.

If you find this feature both redundant and superfluous then you can close this as "not planned" but I want to know the reason for the close at least and it should be convincing.

If you accept this feature request then please let me know about this and I would like to have it in the next stable release of PCSX2 after 1.6.0

Thank you so much that Qt finally has a debugger. I know that moving from wxWidgets to Qt was very difficult and not easy and well done that you managed to do this eventually.

If something anything is not understood or clear then don't hesitate to ask me about it.

### Possible implementation for "Unknown initial value"

1. First scan saves the state of the virtual machine (Not a problem since PS2 memory is 32 MB and modern and high end PCs have at least 8 GB memory. Even old and low end PCs with 4 and even 2 GB of RAM will deal with that).
2. Next scan compares the saved state with the memory of the virtual machine and filters out all addresses that don't pass the test.

buruto:
I have done it by myself.

PR #7870, opened 2023-01-13
Debugger Expressions: Add 'target', 'load', and 'store' references

### LINKED ISSUE #7800: [Feature Request]: NoBreakpoints
### Description

NoBreakpoints are set the same way that breakpoints are set: They both are placed on a memory address, a size ranging from 1 to 4 bytes with three types: execute, write or read.

So the only difference between NoBreakpoints and breakpoints is their behaviour.

As you already know when a breakpoint is triggered then PCSX2 pauses but if NoBreakpoint is triggered then PCSX2 won't pause even if any breakpoint was triggered at the same time.

So in fact a breakpoint pauses PCSX2 if and only if it is triggered and no NoBreakpoint is triggered at the same time.

This means that NoBreakpoints can be used to prevent some instructions from pausing PCSX2 or if the user doesn't want that PCSX2 will pause on certain conditions.

### Reason

NoBreakpoints are another possible solution to the same problem that logpoints solve as described in #7750

### Examples

Assuming that #7696 is implemented and after doing "Unknown initial value" and repeating "Increased Value" until finding the memory address of the "clear time" value in Devil May Cry.

Then add a read breakpoint in that "clear time" memory address.

After that the game pauses every frame because there is a `lw` instruction that reads that "clear time" value every frame and because of this the game is unplayable until either the read breakpoint is either removed or disabled or the problematic `lw` instruction is assembled in such a way so it doesn't trigger the read breakpoint anymore and face the consequences of doing that that may be not good as said in #7750.

But if this feature request, that is #7800, will be implemented then simply add an execute NoBreakpoint on that problematic `lw` instruction to prevent that `lw` instruction from pausing PCSX2.

At last finish the current mission and PCSX2 will pause on the `lw register` instruction that reads the "clear time" value before it is drawn on the screen or window because of the same read breakpoint that is still exist and active.

The NoBreakpoint doesn't prevent that `lw register` instruction from pausing PCSX2 because the type of that NoBreakpoint is "Execute" and it is not placed there but on other `lw` instruction that reads the "clear time" every frame and that `lw register` reads the "clear time" value only when completing any mission and before the "clear time" is drawn on the screen or window so the NoBreakpoint is not triggered when that `lw register` instruction is executed.

All what left to do is to assemble that `lw register` instruction to `move register, zero` and then click on the "Resume" button.

Now you see that the clear time is 0:00:00 even though you didn't finish the mission with that time.

Some PCSX2 users may prefer breakpoints with NoBreakpoints more than logpoints if they don't want to mess up with the Logs window of PCSX2 in my opinion.

After all I am talking about an instruction that is executed every frame.

Adding a logpoint that will be triggered by such instructions might "flood" the Logs window very quickly and finding there the specific or particular instruction that user is interested at may be a bit hard.

With both breakpoints and NoBreakpoints the task is easier.

Unfortunately I can't think of another example with the other use case but fortunately I can theoretically both tell and describe what the other use case can be:

Assume that there is a `lw` instruction that reads from two different memory addresses and the user found it somehow:

The first is read by that `lw` instruction every frame while the other is read only when something happens in the game.

The user placed an execute breakpoint on that `lw` instruction and the game pauses every frame because the `lw` instruction reads the first memory address every frame and triggers the execute breakpoint every frame.

Through the debugger the user finds the memory address that `lw` instruction reads from every frame.

Place there a read NoBreakpoint of size 4 bytes, because `lw` always reads a word and a word is always 4 bytes, to prevent that `lw` instruction from pausing PCSX2 when it reads from the first memory address every frame.

Now the user finally can and he or she does what he or she wants to do in the game and causes that `lw` instruction to read from the second memory address.

Now PCSX2 pauses and again through the debugger the user found the memory address that he or she was interested at and the NoBreakpoint didn't prevent that `lw` instruction from pausing PCSX2 because it was not placed on the second memory address that the `lw` instruction reads from even though it's type is "Read".

Instructions that read from more than one different memory addresses can be usually be found inside for loops that traverses or iterates over an array of values to read the entire array.

A read NoBreakpoint can be used if the user doesn't want that PCSX2 will pause on when some elements of any array are read on certain indexes in this generic example.

### Possible Implementation

Let there be two bits which both default to 0: "Break" and "NoBreak"

When any breakpoint is triggered the "Break" bit becomes 1 and when any NoBreakpoint is triggered then the "NoBreak" bit becomes 1.

After drawing a frame:

If the "NoBreak" bit is 1 then the state of the "Break" bit is ignored (don't care) and PCSX2 won't pause.

This means that PCSX2 won't pause even if the "Break" bit is 1.

Else if the "NoBreak" bit is 0 then PCSX2 will pause if and only if the "Break" bit is 1.

In all possible scenarios PCSX2 will pause if and only if both the "NoBreak" bit is 0 and the "Break" bit is 1.

Note that the "NoBreak" bit always takes precedence over the "Break" bit and the "NoBreak" bit is always tested first before the "Break" bit to fulfil user desire to make PCSX2 not pause on certain instructions or conditions that are created by the breakpoints.

And both bits are 0 again before drawing another frame of course.

F0bes:
### Description of Changes
This PR adds three new references for the debugger condition expressions

`target` -> The evaluated target of the load or the store, 0 if the opcode is not a memory access
`load` -> The evaluated target of a load, is 0 if the opcode is not loading
`store` -> The evaluated target of a store, is 0 if the opcode is not storing

----

#### Here is an example

`v1 = 0x12000000`
<table>
<tr>
	<td>
			<b>Breakpoint Instruction:</b>
	<td>
			<b>ld v0,0x1000(v1)</b>
	<td>
			<b>sd v0,0x1000(v1)</b>
	<td>
			<b>nop</b>
<tr>
	<td>
			<code>target</code>
	<td>
			<code>0x12001000</code>
	<td>
			<code>0x12001000</code>
	<td>
			<code>0</code>
<tr>
	<td>
			<code>load</code>
	<td>
			<code>0x12001000</code>
	<td>
			<code>0</code>
	<td>
			<code>0</code>
<tr>
	<td>
			<code>store</code>
	<td>
			<code>0</code>
	<td>
			<code>0x12001000</code>
	<td>
			<code>0</code>
</table>

Some other stuff fixed here:

Fixes the disassembly widget mouse click handler
Fixes the breakpoint dialog not setting the breakpoint address (oops)
Fixes the breakpoint dialog not setting the condition CPU.

### Rationale behind Changes
My solution to #7800, if I understood it correctly.

### Suggested Testing Steps
I've tested it with `ld`, `sd`, `lq`. You can try other memory load / store opcodes.

----
Resolves #7800

### DIFF
--- a/pcsx2-qt/Debugger/BreakpointDialog.cpp
+++ b/pcsx2-qt/Debugger/BreakpointDialog.cpp
@@ -112,11 +112,14 @@ void BreakpointDialog::accept()
 			return;
 		}
 
+		bp->addr = address;
+
 		bp->enabled = m_ui.chkEnable->isChecked();
 
 		if (!m_ui.txtCondition->text().isEmpty())
 		{
 			bp->hasCond = true;
+			bp->cond.debug = m_cpu;
 
 			if (!m_cpu->initExpression(m_ui.txtCondition->text().toLocal8Bit().constData(), expr))
 			{
--- a/pcsx2-qt/Debugger/DisassemblyWidget.cpp
+++ b/pcsx2-qt/Debugger/DisassemblyWidget.cpp
@@ -487,7 +487,7 @@ void DisassemblyWidget::paintEvent(QPaintEvent* event)
 
 void DisassemblyWidget::mousePressEvent(QMouseEvent* event)
 {
-	const u32 selectedAddress = (event->position().y() / m_rowHeight * 4) + m_visibleStart;
+	const u32 selectedAddress = (static_cast<int>(event->position().y()) / m_rowHeight * 4) + m_visibleStart;
 	if (event->buttons() & Qt::LeftButton)
 	{
 		if (event->modifiers() & Qt::ShiftModifier)
@@ -523,7 +523,7 @@ void DisassemblyWidget::mouseDoubleClickEvent(QMouseEvent* event)
 	if (!m_cpu->isAlive())
 		return;
 
-	const u32 selectedAddress = (event->position().y() / m_rowHeight * 4) + m_visibleStart;
+	const u32 selectedAddress = (static_cast<int>(event->position().y()) / m_rowHeight * 4) + m_visibleStart;
 	if (CBreakPoints::IsAddressBreakPoint(m_cpu->getCpuType(), selectedAddress))
 	{
 		Host::RunOnCPUThread([&] { CBreakPoints::RemoveBreakPoint(m_cpu->getCpuType(), selectedAddress); });
--- a/pcsx2/DebugTools/DebugInterface.cpp
+++ b/pcsx2/DebugTools/DebugInterface.cpp
@@ -18,6 +18,7 @@
 #include "DebugInterface.h"
 #include "Memory.h"
 #include "R5900.h"
+#include "R5900OpcodeTables.h"
 #include "Debug.h"
 #include "VU.h"
 #include "GS.h" // Required for gsNonMirroredRead()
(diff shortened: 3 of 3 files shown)


buruto:
And what about #7696? What makes #7696 so difficult?

refractionpcsx2:
Can you not spam every debugger PR with your requests please. If somebody wants to do it, they will get around to it. You're welcome to do it yourself.

buruto:
> Can you not spam every debugger PR with your requests please. If somebody wants to do it, they will get around to it. You're welcome to do it yourself.

I am on it.

F0bes:
> And what about #7696? What makes #7696 so difficult?

What makes it difficult? Spending some of my free time (I **really** don't have much anymore) on a project I volunteer for just to fulfill someone's feature request. Only for them to point at another feature request and go "What about this one???"
You're welcome, by the way :^)

